Issue #10359, opened 2017-03-06
Basic Authentication credentials are not converted to UTF-8 or other charset before usercheck


Manuel1948:
* [x] bug
* [x] enhancement

* CakePHP Version: 3.4.2

### Problem

When using Basic Authentication and UTF-8 as encoding, username and passwords will not be encoded in utf8 instead the encoding is still iso... , so the comparison with the database in utf8-encoding is not possible and reply the following error:

Error: [PDOException] SQLSTATE[IMSSP]: An error occurred translating string for input param 1 to UCS-2

Manuel1948:
PR #10360

cleptric:
PR is open.

PR #10360, opened 2017-03-06
Update BasicAuthenticate To Support Different Encodings Than ISO-8859-1

Manuel1948:
The problem is that basic authentication credentials are encoded in ISO-8859-1, but if you use an other encoding f.e. UTF-8 and the Auth component compares a username or password with special chars it comes to errors.

So I added a check if there is set another valid encoding than ISO-8859-1 and if so, convert it to the actual encoding.


Manuel1948:
Issue #10359

cleptric:
Could please add a test case for that.

src/Auth/BasicAuthenticate.php
```diff
@@ -75,6 +76,11 @@ public function getUser(ServerRequest $request)
     {
         $username = $request->env('PHP_AUTH_USER');
         $pass = $request->env('PHP_AUTH_PW');
+        $encoding = Configure::read('App.encoding');
+        if ($encoding && in_array($encoding, mb_list_encodings()) && $encoding != 'ISO-8859-1') {
```
dereuromark: Please use !==

src/Auth/BasicAuthenticate.php
```diff
@@ -75,6 +76,11 @@ public function getUser(ServerRequest $request)
...
+            $username = mb_convert_encoding($username, $encoding, 'ISO-8859-1');
+            $pass = mb_convert_encoding($pass, $encoding, 'ISO-8859-1');
```
dereuromark: Ideally that string is a constant in that class, so we don't have to use a magic string 3 times (with the chance of typos).

cleptric:
I'm also not sure, if they are always ` ISO-8859-1` encoded. Isn't that depending of our php.ini settings?

src/Auth/BasicAuthenticate.php
Manuel1948: thx

src/Auth/BasicAuthenticate.php
Manuel1948: thx

dereuromark:
Indeed, this sounds like we should get the input encoding from a more dynamic source rather than a hardcoded string.

cleptric:
Which then leads me to the questions if this can't be fixed by just updating your php.ini?

Manuel1948:
Dear cleptric,

can you suggest me a change in php.ini because my default charset is UTF-8 but this not the solution :/

Manuel1948:
http://toscho.de/2009/authentifizierung-php-utf-8/ 

I edited the commits again based on this page.

src/Auth/BasicAuthenticate.php
```diff
@@ -75,6 +77,11 @@ public function getUser(ServerRequest $request)
     {
         $username = $request->env('PHP_AUTH_USER');
         $pass = $request->env('PHP_AUTH_PW');
+        $encoding = Configure::read('App.encoding');
+        if ($encoding && in_array($encoding, mb_list_encodings())) {
+            $username = mb_convert_encoding($username, $encoding, self::ENCODING_SOURCES);
```
dereuromark: Late static binding (static::) helps to allow extensions to work with their own const.

Manuel1948: I'm a bit confused now, should I correct the :: to -> or is it okay?

ADmad: You need to use `::static` instead of `::self`

Manuel1948: Thank you, I'm not that expert as you are

markstory (changes requested):
A test case is required to merge this. It will regress to easily in the future without a test case.

Manuel1948:
Hope someone can write a test case, because I can't :/

Manuel1948:
This problem isn't currently reproducable anymore, looks like something has changed this behavior.

PR #10362, opened 2017-03-06
Update ErrorHandler.php

inoas:
I don't know about the specifics but in general it is app.php now?


inoas:
In general ErrorHandlers class docblock needs an update of those who know its internals

Issue #10361, opened 2017-03-06
EntityTrait::setHidden() with $merge set to TRUE returns wrong results.


alysson-azevedo:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: v3.4.2.
* Platform and Target: PHP 7.0.15-0ubuntu0.16.10.4

When the `_hidden` property is not empty and we call `$entity->setHidden(['some_field'], true)`, the ["merge"](https://api.cakephp.org/3.4/source-class-Cake.Datasource.EntityTrait.html#436) fails.

I don't know about other versions, but in php7, Array + Array only works well with associative arrays.

dereuromark:
You are not using it as intended (see also test cases):

    [
        'some_field' => true,
        'other' => false,
    ]

etc. So it is always an assoc. array.

alysson-azevedo:
But the [docs](https://book.cakephp.org/3.0/en/orm/entities.html#hiding-properties) says:
```
protected $_hidden = ['password'];
```
Should it be change to ['password'=>true] ?

alysson-azevedo:
@dereuromark i could not find the test case, but what you said didn't work.

```
$entity = new Entity(['field1' => 'a', 'field2' => 'b', 'field3' => 'c']);

$entity->setHidden(['field1' => true], true);
$entity->visibleProperties() //returns ["field1", "field2", "field3"]

$entity->setHidden(['field1'], true);
$entity->visibleProperties() //returns [1 => "field2", 2 => "field3"]

$entity->setHidden(['field2' => true], true);
$entity->visibleProperties() //returns [1 => "field2", 2 => "field3"]

$entity->setHidden(['field2'], true);
$entity->visibleProperties() //returns [1 => "field2", 2 => "field3"]
```

dereuromark:
Looks like you are right, according to https://github.com/cakephp/cakephp/blob/master/tests/TestCase/ORM/EntityTest.php#L1029 the syntax of yours should work.

dereuromark:
So the https://github.com/cakephp/cakephp/blob/master/src/Datasource/EntityTrait.php#L436 seems to be the cause of the issue, assoc. array merging is only OK with string key based arrays (the syntax I thought it would use). Apparently the documented syntax is the one in use, and that one collides with the way it gets merged here. :/

markstory:
Closing as a pull request is open now.

PR #10364, opened 2017-03-06
Fix EntityTrait and merging of properties that are not assoc arrays.

dereuromark:
Interesting that there were no test cases for these methods, neither for the merge part nor without.

Resolves https://github.com/cakephp/cakephp/issues/10361

### DIFF
--- a/src/Datasource/EntityTrait.php
+++ b/src/Datasource/EntityTrait.php
@@ -433,7 +433,7 @@ trait EntityTrait
             return $this;
         }
 
-        $this->_hidden += $properties;
+        $this->_hidden = array_merge($this->_hidden, $properties);
 
         return $this;
     }
@@ -482,7 +482,7 @@ trait EntityTrait
             return $this;
         }
 
-        $this->_virtual += $properties;
+        $this->_virtual = array_merge($this->_virtual, $properties);
 
         return $this;
     }
--- a/src/Datasource/EntityTrait.php
+++ b/src/Datasource/EntityTrait.php
@@ -433,7 +433,8 @@ trait EntityTrait
             return $this;
         }
 
-        $this->_hidden = array_merge($this->_hidden, $properties);
+        $properties = array_merge($this->_hidden, $properties);
+        $this->_hidden = array_unique($properties);
 
         return $this;
     }
@@ -482,7 +483,8 @@ trait EntityTrait
             return $this;
         }
 
-        $this->_virtual = array_merge($this->_virtual, $properties);
+        $properties = array_merge($this->_virtual, $properties);
+        $this->_virtual = array_unique($properties);
 
         return $this;
     }
--- a/tests/TestCase/ORM/EntityTest.php
+++ b/tests/TestCase/ORM/EntityTest.php
@@ -1070,6 +1070,7 @@ class EntityTest extends TestCase
         $this->assertSame(['secret', 'name'], $result);
 
         $entity->setHidden(['name'], true);
+        $result = $entity->getHidden();
         $this->assertSame(['secret', 'name'], $result);
     }
 
--- a/tests/TestCase/ORM/EntityTest.php
+++ b/tests/TestCase/ORM/EntityTest.php
@@ -1039,14 +1039,14 @@ class EntityTest extends TestCase
     {
         $data = ['secret' => 'sauce', 'name' => 'mark', 'id' => 1];
         $entity = new Entity($data);
-        $entity->setVirtual(['secret']);
+        $entity->setHidden(['secret']);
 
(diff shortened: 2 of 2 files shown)


src/Datasource/EntityTrait.php
```diff
@@ -482,7 +482,7 @@ public function setVirtual(array $properties, $merge = false)
             return $this;
         }
 
-        $this->_virtual += $properties;
+        $this->_virtual = array_merge($this->_virtual, $properties);
```
markstory: What happens when the same property is set more than once? I don't think we want duplicates.

dereuromark: There is a test for this.

tests/TestCase/ORM/EntityTest.php
```diff
@@ -1060,6 +1103,29 @@ public function testToArrayVirtualProperties()
     }
 
     /**
+     * Tests setting virtual properties with merging.
+     *
+     * @return void
+     */
+    public function testSetVirtualWithMerge()
```
saeideng: And a test for without merge ( for virtual) ?

dereuromark: Those already exist (with their former name without set prefix).
No one added the ones for the new feature however, and it was already broken by design.

dereuromark:
This bugfix should be good to go then.

tests/TestCase/ORM/EntityTest.php
```diff
@@ -1031,6 +1031,49 @@ public function testToArrayHiddenProperties()
...
+        $result = $entity->getHidden();
+        $this->assertSame(['secret'], $result);
+
+        $entity->setHidden(['name'], true);
+
+        $result = $entity->getHidden();
+        $this->assertSame(['secret', 'name'], $result);
+
+        $entity->setHidden(['name'], true);
+        $this->assertSame(['secret', 'name'], $result);
```
saeideng: duplicate assert
and above that  `$entity->setHidden(['name'], true);`  not affect

tests/TestCase/ORM/EntityTest.php
```diff
@@ -1060,6 +1103,29 @@ public function testToArrayVirtualProperties()
...
+        $result = $entity->getVirtual();
+        $this->assertSame(['secret'], $result);
+
+        $entity->setVirtual(['name'], true);
+
+        $result = $entity->getVirtual();
+        $this->assertSame(['secret', 'name'], $result);
+
+        $entity->setVirtual(['name'], true);
+        $this->assertSame(['secret', 'name'], $result);
```
saeideng: duplicate assert
and above that  `$entity->setVirtual(['name'], true);`  not affect

tests/TestCase/ORM/EntityTest.php
dereuromark: Thats in purpose to Test that the Same field doesnt End Up multiple times.

saeideng:
and have we this ( `'name'=>true` )usage ? 
like

` $entity->setHidden(['name'=>true], true);`

tests/TestCase/ORM/EntityTest.php
saeideng: I am not understand your mean but you used $result that created 4 lines before that 
see again

dereuromark:
Not that I know of.

src/Datasource/EntityTrait.php
dereuromark: It seems it doesn't need it, array_merge() here seems to not duplicate, as the test proves.

lorenzo: What test proves that?

dereuromark: The one @saeideng was talking below thinking it was duplicating things.

Calling `$entity->setHidden(['name'], true);` multiple times keeps the array value once in the array, thus this seems to proof that this does not need array_unique() here.

dereuromark: OK, the test was not correct, I see. So we must add array_unique() then.

dereuromark:
I adjusted the code as per fixed unit test.

src/Datasource/EntityTrait.php
saeideng: I no saying about duplicate key/value :)
I just sayed there are duplicated a unit test that @lorenzo was fix it in https://github.com/cakephp/cakephp/pull/10364/commits/bb9eb987b519eef1e4809e970093ab2f00782cdd
Thanks @lorenzo 
Why we need array_unique()?

dereuromark:
Failing CS tests are unrelated (already in master).

PR #10367, opened 2017-03-07
Remove unused namespace

antograssiot:


ADmad:
I was hoping phpstan would report such unneeded `use` statement, guess it doesn't at level 0.

dereuromark:
Did you use my "unused use statement sniff", or did you do all that manually?

antograssiot:
None of those possibilities.
I used php cs fixer

PR #10368, opened 2017-03-07
Remove internal annotations from Cake\Database\Expression classes

chinpei215:
Since being mentioned in the Cookbook, these classes are no longer internal classes.


lorenzo:
Thanks!

PR #10315, opened 2017-02-27
add options of preg_replace limit

gnineh:

PR #10372, opened 2017-03-07
Add preg_replace limit to options for Text::highlight().

dereuromark:
Implements https://github.com/cakephp/cakephp/pull/10315 with a test case.


chinpei215:
@dereuromark I think we should leave the original author's name. You can fetch the pull request, squash it, and add your test case.

dereuromark:
I didn't because of the commit messages in a language I cannot understand.
Maybe that user is fine with it getting into the core this way.

chinpei215:
OK. I'll do it for the original author, and you. Fortunately, I can half-understand what the original author's commit message. So I'll translate it into English when I squash his commit.

PR #10378, opened 2017-03-08
Make error class of div in FormHelper::input configurable

mvdriel:
Make error class of div in FormHelper::input configurable, so it is possible to generate html snippets compatible with Bootstrap 4.

### DIFF
--- a/lib/Cake/View/Helper/FormHelper.php
+++ b/lib/Cake/View/Helper/FormHelper.php
@@ -1068,7 +1068,7 @@ class FormHelper extends AppHelper {
 		if ($type !== 'hidden' && $error !== false) {
 			$errMsg = $this->error($fieldName, $error);
 			if ($errMsg) {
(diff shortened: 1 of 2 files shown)


lib/Cake/View/Helper/FormHelper.php
```diff
@@ -1089,6 +1089,7 @@ public function input($fieldName, $options = array()) {
 		if (!empty($divOptions['tag'])) {
 			$tag = $divOptions['tag'];
 			unset($divOptions['tag']);
+			unset($divOptions['errorClass']);
```
markstory: This can be combined with the `unset()` 1 line above.

markstory:
Related to #2031

lib/Cake/View/Helper/FormHelper.php
mvdriel: Yes, thanks, done.

tersmitten:
Thanks @markstory

PR #10380, opened 2017-03-08
Another set of phpstan reported errors

antograssiot:


ADmad:
How did you find there errors?  They aren't reported at phpstan level 1.

antograssiot:
I add a look to high level locally @ADmad and picked some that were not false positive

ADmad:
Are there too many errors reported at level 2?  If not we could fix them all and bump the level in travis config :)

antograssiot:
level 2 reports 121 errors 😞 
level 4 reports 243 errors

lorenzo:
Good job!

PR #10381, opened 2017-03-08
Add ability to remove columns from schema objects.

markstory:
This is useful when you want to drop a column with 0 downtime/errors. This method can be used in ORM\Table::_initializeSchema() to un-map fields that are going to be dropped soon.

### DIFF
--- a/src/Database/Schema/TableSchema.php
+++ b/src/Database/Schema/TableSchema.php
@@ -346,6 +346,21 @@ class TableSchema
     }
 
     /**
+     * Remove a column from the table schema.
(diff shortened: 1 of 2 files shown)


lorenzo:
This makes sense to me... It would be nice to have it in the docs for both the ORM and migrations so folks understand its purpose

markstory:
@lorenzo I was going to update the docs once this merges.

lorenzo:
I like how simple the solution to this problem is. I've been wondering how to solve it for quite a while now :P

Issue #10397, opened 2017-03-10
3.4.2 Exception on Fatal error misses file name and line number for JSON/serialized responses


inoas:
This is a (multiple allowed):

* [x] bug
* [x] enhancement

* CakePHP Version: 3.4.2
* PHP Version: PHP 7.1.2-4+deb.sury.org~trusty+1

### What you did

- app_local: debug = true.
- Created some syntax error in some table class
- Fired off a JSON request

### What happened

```json
{
  "message": "syntax error, unexpected '$this' (T_VARIABLE)",
  "url": "/admin/flexi-cms/cms-documents/post.json",
  "code": 500
}
```

### What you expected to happen
```json
{
  "message": "syntax error, unexpected '$this' (T_VARIABLE) in /var/www/cakephp/app/plugins/FlexiCms/src/Model/Table/CmsTemplatesTable.php on line 74",
  "url": "/admin/flexi-cms/cms-documents/post.json",
  "code": 500
}
```

Ref: https://github.com/cakephp/cakephp/pull/10363 which fixed the same kind (but obviously not source) of issue for error.log

inoas:
Also https://api.cakephp.org/3.3/source-class-Cake.Error.PHP7ErrorException.html#42 should that not be:

```php
        parent::__construct(
            sprintf('(%s) - %s in file %s on line %s', get_class($error), $message, $this->file, $this->line), $code);
```
... aka include file and line?
Ref: https://api.cakephp.org/3.3/source-class-Cake.Error.FatalErrorException.html#34

And this works with some class variables and inherits a getAttributes custom exception class method... however `PHP7ErrorException` is a decedent of `\Exception`

inoas:
This has been worked around for now in https://github.com/cakephp/cakephp/pull/10400

PR #10399, opened 2017-03-10
3.x ExceptionRenderer should always report filename & line (if debug=true) (PHP 7)

inoas:
Ref: https://github.com/cakephp/cakephp/issues/10397

### Example Before

```json
{
  "message": "syntax error, unexpected '$this' (T_VARIABLE)",
  "url": "/admin/flexi-cms/cms-documents/post.json",
  "code": 500,
}
```

### Example After without patching PHP7Exception (still wrong!)

```json
{
  "message": "syntax error, unexpected '$this' (T_VARIABLE)",
  "url": "/admin/flexi-cms/cms-documents/post.json",
  "code": 500,
  "file": "/var/www/cakephp/app/vendor/cakephp/cakephp/src/Error/BaseErrorHandler.php",
  "line": 169
}
```

### Example After with patching PHP7Exception
```json
{
  "message": "syntax error, unexpected '$this' (T_VARIABLE)",
  "url": "/admin/flexi-cms/cms-documents/patch.json",
  "code": 500,
  "file": "/var/www/cakephp/app/plugins/FlexiCms/src/Model/Table/CmsTemplatesTable.php",
  "line": 169
}
```

#### Reason in BaseErrorHandler.php:
```php
    public function wrapAndHandleException($exception)
    {
        if ($exception instanceof Error) {
            $exception = new PHP7ErrorException($exception); // wraps exception in exception
        }
        $this->handleException($exception);
    }
```

### DIFF
--- a/src/Error/ExceptionRenderer.php
+++ b/src/Error/ExceptionRenderer.php
@@ -179,15 +179,17 @@ class ExceptionRenderer implements ExceptionRendererInterface
             'url' => h($url),
             'error' => $unwrapped,
             'code' => $code,
-            'file' => $exception->getFile(),
-            'line' => $exception->getLine(),
-            '_serialize' => ['message', 'url', 'code', 'file', 'line']
+            '_serialize' => ['message', 'url', 'code']
         ];
         if ($isDebug) {
             $viewVars['trace'] = Debugger::formatTrace($unwrapped->getTrace(), [
                 'format' => 'array',
                 'args' => false
             ]);
+            $viewVars['file'] = $exception->getFile();
+            $viewVars['line'] = $exception->getLine();
+            $viewVars['_serialize'][] = 'file';
+            $viewVars['_serialize'][] = 'line';
         }
         $this->controller->set($viewVars);
 
--- a/src/Error/ExceptionRenderer.php
+++ b/src/Error/ExceptionRenderer.php
@@ -179,7 +179,9 @@ class ExceptionRenderer implements ExceptionRendererInterface
             'url' => h($url),
             'error' => $unwrapped,
             'code' => $code,
-            '_serialize' => ['message', 'url', 'code']
+            'file' => $exception->getFile(),
+            'line' => $exception->getLine(),
+            '_serialize' => ['message', 'url', 'code', 'file', 'line']
         ];
         if ($isDebug) {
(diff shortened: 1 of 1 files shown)


lorenzo:
think those should only be shown if `debug` is enabled

inoas:
@lorenzo good?

inoas:
By the way I think
```
        if ($unwrapped instanceof CakeException && $isDebug) {
            $this->controller->set($unwrapped->getAttributes());
        }
```
... gets filename and line number, however this does not happen for PHP7ErrorException as that has no getAttributes and is not a CakeException.

inoas:
`PHP7ErrorException` also requires a change.

Do you prefer the constructor to match that of `FatalErrorException`

Then it is:
```php
    /**
     * Wraps the passed Error class
     *
     * @param \Error $error the Error object
     * @param int $code Code.
     * @param string|null $file File name.
     * @param int|null $line Line number.
     */
    public function __construct($error, $code = 500, $file = null, $line = null)
    {
        $this->_error = $error;
        $this->message = $error->getMessage();
        if ($code !== null) {
            $this->code = $code; 
        } else {
            $this->code = $error->getCode();
        }
        if ($file !== null) {
            $this->file = $file; 
        } else {
            $this->file = $error->getLine();
        }
        if ($file !== null) {
            $this->error = $error; 
        } else {
            $this->file = $error->getFile();
        }
        parent::__construct(sprintf('(%s) - %s in %s on %s', get_class($error), $this->message, $this->file, $this->line), $this->code);
    }
```

Else it is:
```php
    /**
     * Wraps the passed Error class
     *
     * @param \Error $error the Error object
     */
    public function __construct($error)
    {
        $this->_error = $error;
        $this->message = $error->getMessage();
        $this->code = $error->getCode();
        $this->file = $error->getLine();
        $this->file = $error->getFile();
        parent::__construct(sprintf('(%s) - %s in %s on %s', get_class($error), $this->message, $this->file, $this->line), $code);
    }
```

thinkingmedia:
Does this still work as expected if you don't have xdebug installed. I've found that it adds data to exceptions and `var_dump`.

Also `getFile()` might return null for errors thrown by closure functions or from extensions.

inoas:
New PR open.

PR #10400, opened 2017-03-10
3.x ExceptionRenderer should always report filename & line (if debug=true) (PHP 7)

inoas:
Pass through file name and line number from `PHP7ErrorException` up to `ExceptionRenderer`

Replacement PR for https://github.com/cakephp/cakephp/pull/10399

### DIFF
--- a/src/Error/ExceptionRenderer.php
+++ b/src/Error/ExceptionRenderer.php
@@ -186,8 +186,8 @@ class ExceptionRenderer implements ExceptionRendererInterface
                 'format' => 'array',
                 'args' => false
             ]);
-            $viewVars['file'] = $exception->getFile() ?: "null";
-            $viewVars['line'] = $exception->getLine() ?: "null";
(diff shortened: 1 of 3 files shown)


thinkingmedia:
You shouldn't have to check if debug is true. The exception render doesn't output the error when debug is disabled.

lorenzo:
This looks good to me. Where there any tests to change after this?

thinkingmedia:
Oh I get it now. The original exception is bring wrapped. This looks good to me too. Nice fine.

inoas:
@thinkingmedia wrote:
> You shouldn't have to check if debug is true. The exception render doesn't output the error when debug is disabled.

Assumed the same. Empirically not true.
`$viewVars` would get the line and file serialized and json requests would render those even if debug was set to false.

inoas:
@lorenzo there were, right now it should be okay.
Good to merge?

thinkingmedia:
Looks good to me.

lorenzo:
Thanks @ionas!

PR #10401, opened 2017-03-11
Update URL in Docs and composer.json

tenkoma:
This PR is Docs improvement in cakephp repository.

## 1. Use https as much as possible on the following sites.

If the linked site is always on https, save time for redirect processing.

## 2. Always link API document to latest version

Now that you can redirect to the latest version of API page: cakephp/cakephp-api-docs#82


markstory:
Thanks!

Issue #10398, opened 2017-03-10
Cake 2.x and PHPUnit 4.8


bancer:
This is a (multiple allowed):

* [ ] bug
* [ ] enhancement
* [x] feature-discussion (RFC)

* CakePHP Version: 2.x
* Platform and Target: Any.

It is clear that CakePHP 2.x supports PHPUnit not higher than 3.7.38. My problem is that  one of the projects I work on has many unit tests. Generating the coverage report with xdebug takes hours. According to [this](https://hackernoon.com/generating-code-coverage-with-phpunite-and-phpdbg-4d20347ffb45#.iqfn1oprg) blog generating a coverage report with phpdbg is 14 times faster than with xdebug. PHPUnit 3.7.38 does not support phpdbg but [4.8](https://thephp.cc/news/2015/08/phpunit-4-8-code-coverage-support) does.

The question is how much effort it would take to make CakePHP 2.x compatible with PHPUnit 4.8? Is it achievable at all? What would need to be done?

saeideng:
You can use phpunit4.8 in your project if you need

chinpei215:
> The question is how much effort it would take to make CakePHP 2.x compatible with PHPUnit 4.8?

It is easy. Now I am using PHPUnit 4.8. for my project as I want to to use `assertArraySubset()`. I can send a pull  request to make CakePHP 2.x compatible  with PHPUnit 4.8.

markstory:
@chinpei215 That would be awesome. Previously we pinned the version to 3.7 as I had a number of problems upgrading to earlier versions of PHPUnit 4.x but if you have it working that would be great!

tersmitten:
We also like to upgrade to a more recent version and don't mind putting some work into it. Is there a ticket describing the issues?

tersmitten:
@markstory Congratulations with your baby boy btw

tersmitten:
https://github.com/sebastianbergmann/php-token-stream/issues/65

markstory:
@tersmitten There isn't an existing issue. I remember it having something to do with sessions though.

chinpei215:
All tests passed with PHPUnit 4.8.35: https://travis-ci.org/chinpei215/cakephp/jobs/210079162#L442
I am planning to fix several risky tests and to send a pull request tomorrow.

[...]

PR #10339, opened 2017-03-02
Fix XML output of Cake\Console\HelpFormatter

chinpei215:
- Help text wasn't rendered in `<option>` tag
- Non-existent short name was rendered as '-' unintentionally
- Related test cases had never worked

Note that I didn't change boolean values to be rendered as '0'. Instead, I fixed the test case so that it matches empty strings.

After this commit merged, I would like to port the same changes to the 2.x branch.

PR #10405, opened 2017-03-12
[2.next] Make test suites compatible with PHPUnit 4.8

chinpei215:
- Made test suites compatible with PHPUnit 4.8
- Fixed XML output of HelpFormatter (Backport from #10339. The related test cases don't pass with PHPUnit 4.8. I noticed this issue when I was working to make CakePHP 2.x compatible with PHPUnit 4.8 for my project. Then I checked 3.x branch and noticed that the related test cases had been fixed in a wrong way. That was why I reported #10339).
- Added PHPUnit 4.8 to test matrix
- ~~Added CakeFinalizer class. PHPUnit 4.8 marks test cases as 'risky', if the output buffering level is different before and after the test. CakeFinalizer allows us to adjust the output buffering level, when we want to test some function which changes the output buffering level and throws an exception.~~
- Fixed several risky tests.
    - Fixed CakeTestCase::run() to restore original output buffering level
    - Fixed an undefined variable warning in missing_controller.ctp
    - Fixed ViewTest (backport of 92bd86274b64c31ef6034caa74bb168be041b5e0)
    - Fixed ThemeViewTest

I have not changed `composer.json` as I think that using PHPUnit 4.8 is optional. So, for now, only a test is running with PHPUnit 4.8. However, It may be arguable. I am fine to change `composer.json` and `.travis.yml` to let all test cases run with PHPUnit 4.8.

Refs: #10398

### DIFF
--- a/lib/Cake/Test/Case/View/ThemeViewTest.php
+++ b/lib/Cake/Test/Case/View/ThemeViewTest.php
@@ -198,7 +198,6 @@ class ThemeViewTest extends CakeTestCase {
 /**
  * testMissingView method
  *
- * @expectedException MissingViewException
  * @return void
  */
 	public function testMissingView() {
@@ -211,17 +210,18 @@ class ThemeViewTest extends CakeTestCase {
 		$this->Controller->params['pass'] = array('home');
 
 		$View = new TestTheme2View($this->Controller);
-		ob_start();
-		$View->getViewFileName('does_not_exist');
-		$expected = ob_get_clean();
-		$this->assertRegExp("/PagesController::/", $expected);
-		$this->assertRegExp("/views(\/|\\\)themed(\/|\\\)my_theme(\/|\\\)pages(\/|\\\)does_not_exist.ctp/", $expected);
+
+		try {
+			$View->getViewFileName('does_not_exist');
+			$this->fail('No exception');
+		} catch (MissingViewException $e) {
+			$this->assertContains('Pages' . DS . 'does_not_exist.ctp', $e->getMessage());
+		}
 	}
 
 /**
  * testMissingLayout method
  *
- * @expectedException MissingLayoutException
  * @return void
  */
 	public function testMissingLayout() {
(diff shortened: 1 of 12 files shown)


chinpei215:
I may have to merge 2.x into 2.next and rebase this pull request, to fix an apcu issue: https://github.com/cakephp/cakephp/commit/173f927ef30f2032d14f383e58da4bface5bf130

May I merge 2.x into 2.next?

dereuromark:
Yeah, sure!

lib/Cake/Test/Case/Utility/CakeFinalizerTest.php
```diff
@@ -0,0 +1,50 @@
...
+ */
+class CakeFinalizerTest extends CakeTestCase {
+
+/**
+ * Tests that the given callable function is called by the CakeFinalizer
+ *
+ * @return void
+ */
+	public function testCallable() {
+		$mock = $this->getMock('stdClass', array('finalize'));
```
markstory: We should probably start using the un-deprecated `getMockBuilder()` method for new tests.

lib/Cake/Test/Case/View/ThemeViewTest.php
```diff
@@ -232,11 +228,7 @@ public function testMissingLayout() {
 		$this->Controller->theme = 'my_theme';
 
 		$View = new TestTheme2View($this->Controller);
-		ob_start();
 		$View->getLayoutFileName();
-		$expected = ob_get_clean();
-		$this->assertRegExp("/Missing Layout/", $expected);
```
markstory: Should the exception message have an assertion added?

chinpei215:
Thank you for your comments. Probably, I'll take a look tomorrow.

chinpei215:
Rebased. But I am re-thinking about new CakeFinalizer class. First I thought it was nice idea, but I changed my mind. Now it is being used only in ViewHelperTest. I will remove CakeFinalizer and fix related test cases instead.

chinpei215:
Removed CakeFinalizer and fixed related risky tests. All tests has been passed with both PHPUnit 3.7 and 4.8. By the way, should we bump the PHPUnit version to 4.8? I am fine to change `composer.json` and `.travis.yml`. Then the version constraint in `composer.json` will be `>=3.7 <5.0` for the backward compatibility.

markstory:
Adding a more flexible version constraint on phpunit sounds like a good idea.

Issue #10406, opened 2017-03-12
3.next - Consolidate CookieCollections


markstory:
This is a (multiple allowed):

* [ ] bug
* [ ] enhancement
* [x] feature-discussion (RFC)

CakePHP 3.5.0 will feature a new set of cookie objects that provide an object-orienated interface to building and using HTTP cookies. As of #10328 we have a few different Cookie 'collections'. I'd like to unify those collections with the goal of only having 1 single cookie collection, as I don't think the use-cases we have require 4 classes.

### Use Cases for a CookieCollection

CookieCollections are used in a few different places in CakePHP:

1. In an Http\Client as a cookie jar where state is retained between client requests.
2. Adding cookies to a Controller Response.
3. Reading cookie data from a ServerRequest.

Each use case requires some common methods and some use-case specific methods.

### Generic Methods

* Iteration (IteratorAggregate)
* `has($key)` - Check if a cookie exists
* `get($key)` - Get a cookie
* `remove($key)` - Remove a cookie
* `add($cookie)` - Add a cookie

### Methods specific to Http\Client usage

This is the most complex use-case to afford as there are backwards compatibility concerns. The new methods & backwards compatibility behavior required are:

* `addFromResponse($response, $request)` - Add cookies from a response. A request is needed to infer the path and domain of cookies that have not included it. Currently this method is called `store()`. This alias/signature would be deprecated.
* `addToRequest($request)` - Apply all the cookies matching this request. Currently `Http\Client\CookieCollection` calls this method `get($url)`. This behavior for `get()` would be deprecated and could be shimmed by looking for cookie 'names' that contain the `://` URL sub-pattern. Cookies names containing `://` are very rare, and we can use this as a sniff to enable backwards compatible behavior case libraries that consume `Http\Client\CookieCollection`.

These methods allow a CookieCollection to function as a 'cookie jar' for the client. Covering another use case currently covered by a standalone class. 

### Methods specific to adding cookies to a Controller Response

* `addToResponse($response)` - Add cookies in the collection to a response.

### Methods specific to Reading cookie data from a ServerRequest

* `createFromRequest($request)` - Build a new CookieCollection from a request object.

All of the above methods could be implemented in a single class allowing us to streamline how users have to interact with cookies in CakePHP even further.

### Edits

* Removed `all()`. I don't have any use cases for it.

robertpustulka:
Regarding Http\Client use cases:
- `addFromResponse($response, $request)` this would mutate current collection?
[...]

PR #10408, opened 2017-03-13
3.next - CookieCollection basic manipulation

markstory:
Add the generic/basic collection manipulation methods described in #10406. I'll do the other methods in separate pull requests once we've settled on names/approaches for them.


markstory:
I've made the `add` and `remove` methods work in an immutable way. I feel this is necessary to maintain consistency with `Collection` and also honour the immutability of the request/response that we plan on using cookie collections with. If the collections were not immutable the following could be done:

```php
$cookies = $request->getCookieCollection();
$cookies->add(...);
```

This example would mutate the cookies in the request which is not desirable. Of course this also implies that `Http\Cookie\Cookie` also needs to become an immutable object. Does that sound unreasonable @burzum ?

burzum:
@markstory, yes it does sound OK to me. 👍

PR #10409, opened 2017-03-13
Fix PhpDoc for RouteBuilder::redirect()

tenkoma:
The second argument `$url` of `RouteBuilder::redirect()` accepts a string.


dereuromark:
Thank you!

Issue #10329, opened 2017-03-01
autoLinkUrls not handeling URLs correctly


sevvlor:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.4.2.
* Platform and Target: PHP 7.0.15-0ubuntu0.16.04.2 

### What you did
I tried to autolink various URLs. Most of them worked fine but some failed.

These URLs failed:
`$this->Text->autoLinkUrls('http://dev.local/threads/search?search_string=rhis+is+a+test', ['target' => '_blank', 'rel' => 'nofollow'])`

Output: `<a href="http://dev.local/threads/search?search_string=rhis" target="_blank" rel="nofollow">http://dev.local/threads/search?search_string=rhis</a>+is+a+test`

`$this->Text->autoLinkUrls('http://www.ad.nl/show/giel-beelen-heeft-weinig-moeite-met-rijontzegging~acd8b6ed/', ['target' => '_blank', 'rel' => 'nofollow'])`

Output: `<a href="http://www.ad.nl/show/giel-beelen-heeft-weinig-moeite-met-rijontzegging" target="_blank" rel="nofollow">http://www.ad.nl/show/giel-beelen-heeft-weinig-moeite-met-rijontzegging</a>~acd8b6ed/`

### What you expected to happen
I expected the URLs to be auto linked fully as this happened in the previous versions of CakePHP (3.3.6) even though the URLs aren't fully valid.

Expected output:
`<a href="http://dev.local/threads/search?search_string=rhis+is+a+test" target="_blank" rel="nofollow">http://dev.local/threads/search?search_string=rhis+is+a+test</a>`

`<a href="http://www.ad.nl/show/giel-beelen-heeft-weinig-moeite-met-rijontzegging~acd8b6ed/" target="_blank" rel="nofollow">http://www.ad.nl/show/giel-beelen-heeft-weinig-moeite-met-rijontzegging~acd8b6ed/</a>`

cleptric:
PR is open

PR #10411, opened 2017-03-13
Also covers %

sevvlor:
The issue #10329 is somewhat related. I noticed the same behaviour as with the ~ in urls.


markstory:
Thanks!

PR #10413, opened 2017-03-14
Fix Http\Client\CookieCollection according to RFC6265.

robertpustulka:
Currently `CookieCollection` will not return cookies without leading dot in a domain for subdomain requests.

I had a real life example where cookies with domain `example.com` (no leading dot) would not be sent with requests to `subdomain.example.com`.

I have found out that modern clients should ignore leading dot. Here is a fix for that.

More reading:
- http://bayou.io/draft/cookie.domain.html
- https://tools.ietf.org/html/rfc6265


markstory (approved):
Looks good. I was referencing an older RFC when I built this. :cry:

Issue #10410, opened 2017-03-13
3.4: associations with same name


mirko-pagliai:
This bug occurred recently, unfortunately I had no way to immediately investigate, so I do not know from which version, maybe from 3.4.

I have `Posts` that belong to `PostsCategories` and `Pages` that belong to `PagesCategoriesTable`.

`PostsTable::initialize()`:

       /// ...
        $this->belongsTo('Categories', [
            'foreignKey' => 'category_id',
            'joinType' => 'INNER',
            'className' => 'MeCms.PostsCategories',
        ]);
       /// ...

`PagesTable::initialize()`:

       /// ...
        $this->belongsTo('Categories', [
            'foreignKey' => 'category_id',
            'joinType' => 'INNER',
            'className' => 'MeCms.PagesCategories',
        ]);
       /// ...

So, as you can see, both associations have the same name.  
I already know that this can be confusing, but until recently worked perfectly.

Now what happens?  
I have a cell that lists the post categories (with something like `$this->Posts->Categories->find()`...). Normally, this cell works.
But when the cell is rendered with an action that also uses the page categories (`$this->Pages->Categories->find()`...), this happens:

```
2017-03-13 05:41:12 - Warning (512): Could not render cell - MeCms\Model\Table\PostsTable association "Categories" of type "manyToOne" to "MeCms\Model\Table\PagesCategoriesTable" doesn't match the expected class "MeCms\Model\Table\PostsCategoriesTable". You can't have an association of the same name with a different target "className" option anywhere in your app. [/var/www/quadrifogliopsichiatria.it/vendor/cakephp/cakephp/src/ORM/Association.php, line 420] in [/var/www/quadrifogliopsichiatria.it/vendor/cakephp/cakephp/src/View/Cell.php, line 271]
```

Debugging `$this->Posts->Categories` inside the cell code there are no problems:

```
object(Cake\ORM\Association\BelongsTo) {
	[protected] _validStrategies => [
		(int) 0 => 'join',
		(int) 1 => 'select'
	]
	[protected] _name => 'Categories'
	[protected] _className => 'MeCms.PostsCategories'
	[protected] _bindingKey => null
	[protected] _foreignKey => 'category_id'
	[protected] _conditions => []
	[protected] _dependent => false
	[protected] _cascadeCallbacks => false
	[protected] _sourceTable => object(MeCms\Model\Table\PostsTable) {

		'registryAlias' => 'MeCms.Posts',
		'table' => 'posts',
[...]

PR #7125, opened 2015-07-28
Fix incorrectly hydrated association data

markstory:
When someone typo's an alias name in a `contain` call, they shouldn't get incorrectly hydrated data. Instead we should raise an exception as we don't have a way of easily fixing their mistake.

Refs #7110

PR #10417, opened 2017-03-15
Remove errors when aliases don't match

markstory:
This hard error makes using the `targetTable` option in associations really hard as it always triggers this condition. By reworking how we generate aliases and eagerloader aliases we can enable the `targetTable` option to behave as intended. The foreignKey will need to be set, when used with `targetTable` and a table that doesn't have a matching alias, but I think that is reasonable given how key generation uses the table's alias for other reasons.

I've had to replace target aliases with the association names, so that the selected fields reflect the association's alias and not the table's original alias. I don't think this will cause problems with application code that relies on the target table's alias being used, as aliases always had to match between the association and table instance. One situation that could be problematic with the new code is if a user is using `$table->aliasField()` in query expressions for association finders. Because the table's alias and association alias no longer have be the same, mismatching aliases could end up in the same query.

Refs #7125
Refs #10410

### DIFF
--- a/src/ORM/Association/BelongsTo.php
+++ b/src/ORM/Association/BelongsTo.php
@@ -151,7 +151,7 @@ class BelongsTo extends Association
     protected function _joinCondition($options)
     {
         $conditions = [];
-        $tAlias = $this->getTarget()->getAlias();
+        $tAlias = $this->_name;
         $sAlias = $this->_sourceTable->getAlias();
         $foreignKey = (array)$options['foreignKey'];
         $bindingKey = (array)$this->getBindingKey();
--- a/src/ORM/Association.php
+++ b/src/ORM/Association.php
@@ -1144,7 +1144,7 @@ abstract class Association
         }
 
         if ($fields) {
-            $query->select($query->aliasFields($fields, $target->getAlias()));
+            $query->select($query->aliasFields($fields, $this->_name));
         }
         $query->addDefaultTypes($target);
     }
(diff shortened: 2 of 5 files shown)


src/ORM/Association/BelongsTo.php
```diff
@@ -151,7 +151,7 @@ public function saveAssociated(EntityInterface $entity, array $options = [])
     protected function _joinCondition($options)
     {
         $conditions = [];
-        $tAlias = $this->getTarget()->getAlias();
+        $tAlias = $this->_name;
```
markstory: This lets the join conditions use the association's alias which can now be different than the target table.

lorenzo:
This makes a lot of sense. I realised late after almost being done with the associations logic that it was a mistake having both the Table and the Association hold an alias property. It should have only been in either of them.

Maybe a midterm goal is to move the property use as much as possible to only `Association`?

lorenzo:
Awesome work @markstory

inoas:
> The foreignKey will need to be set, when used with targetTable and a table that doesn't have a matching alias, but I think that is reasonable given how key generation uses the table's alias for other reasons.

Is/Could this be checked? Else: Documented?

markstory:
@inoas It is pretty obvious that foreignKey needs to be set as the generated SQL is invalid without it. I feel that is self-documenting enough.

> Maybe a midterm goal is to move the property use as much as possible to only Association?

@lorenzo Tables will still need an alias But I could see us converting all the alias usage in the association classes to use the association name instead of the target/source table's aliases.

PR #10419, opened 2017-03-15
Remove default options from scriptStart()

### LINKED ISSUE #10416: Inconsistent default 'safe' for `scriptBlock()` and `scriptStart()`
The default was changed in 3.4 for one method, but not the other:

```php
    public function scriptBlock($script, array $options = [])
    {
        $options += ['safe' => false, 'block' => null];
…
```

```php
    public function scriptStart(array $options = [])
    {
        $options += ['safe' => true, 'block' => null];
…
```

I suggest to adapt scriptStart() to ensure consistent behavior with scriptBlock().

ypnos-web:
Stay consistent with defaults in scriptBlock(). Default for 'safe' is now false. Fixes #10416

### DIFF
--- a/src/View/Helper/HtmlHelper.php
+++ b/src/View/Helper/HtmlHelper.php
@@ -609,7 +609,8 @@ class HtmlHelper extends Helper
      *
      * ### Options
      *
-     * - `safe` Whether the code block should contain a CDATA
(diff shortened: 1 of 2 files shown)


ADmad:
Please add a test to avoid regression.

Issue #10418, opened 2017-03-15
Wrong index from find('list')


bancer:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 2.8.9.
* Platform and Target: Linux, Apache, MySQL.

### What you did
The database contains records with ids starting from 0 (not from 1).
```
$params = array(
	"fields" => array("my_column", "id"),
);
$data = $this->find("list", $params);
```

### What happened
The query above returns the list of records but one is wrong - the record that has id 0 in the database has id 100 in $data. The last id in the database is 99.

### What you expected to happen
I would expect id to be 0 in $data.

markstory:
I've added a test for the key=0 case and wasn't able to reproduce the dropped key. Could you help in providing a way to reproduce the problem you're seeing?

bancer:
@markstory I commented in your [commit](https://github.com/cakephp/cakephp/commit/9dbeeaa1fa9e23042192eb1a033b526f4fd69003).

markstory:
I changed the test around, but I'm still not able to get incorrect data. See c5e31e590df0673d60e931932abd4fcf28f8134a

bancer:
The primary key column should have 0 value in ArticleFixture. Notice `'id' => '0'`:
```
public $records = array(
	array('id' => '0', 'user_id' => 1, 'title' => 'Zeroth Article', 'body' => 'Zeroth Article Body', 'published' => 'Y', 'created' => '2007-03-18 10:39:23', 'updated' => '2007-03-18 10:41:31'),
	array('user_id' => 1, 'title' => 'First Article', 'body' => 'First Article Body', 'published' => 'Y', 'created' => '2007-03-18 10:39:23', 'updated' => '2007-03-18 10:41:31'),
	array('user_id' => 3, 'title' => 'Second Article', 'body' => 'Second Article Body', 'published' => 'Y', 'created' => '2007-03-18 10:41:23', 'updated' => '2007-03-18 10:43:31'),
	array('user_id' => 1, 'title' => 'Third Article', 'body' => 'Third Article Body', 'published' => 'Y', 'created' => '2007-03-18 10:43:23', 'updated' => '2007-03-18 10:45:31')
);
```

In the test:
```
$result = $model->find('list', array(
    'fields' => array('title', 'id')
));
```
Notice 'id' (not 'user_id').

markstory:
[...]

PR #10426, opened 2017-03-16
Wrong 0 list index

bancer:
Attempt to reproduce the issue https://github.com/cakephp/cakephp/issues/10418.

### DIFF
--- a/lib/Cake/Model/Datasource/Database/Mysql.php
+++ b/lib/Cake/Model/Datasource/Database/Mysql.php
@@ -895,8 +895,10 @@ class Mysql extends DboSource {
 				$i++;
 			}
 		}
+		$this->_connection->query('SET @OLD_SQL_MODE=@@SQL_MODE, SQL_MODE="NO_AUTO_VALUE_ON_ZERO"');
 		$result = $statement->execute();
 		$statement->closeCursor();
+		$this->_connection->query('SET SQL_MODE=@OLD_SQL_MODE');
 		if ($this->fullDebug) {
 			$this->logQuery($sql, $valuesList);
 		}
--- a/lib/Cake/Test/Fixture/AnotherArticleFixture.php
+++ b/lib/Cake/Test/Fixture/AnotherArticleFixture.php
@@ -42,8 +42,8 @@ class AnotherArticleFixture extends CakeTestFixture {
  */
 	public $records = array(
 		array('id' => '0', 'title' => 'Zeroth Article', 'created' => '2017-03-17 00:48:23', 'updated' => '2017-03-17 00:48:23'),
-		array('title' => 'First Article', 'created' => '2007-03-18 10:39:23', 'updated' => '2007-03-18 10:41:31'),
-		array('title' => 'Second Article', 'created' => '2007-03-18 10:41:23', 'updated' => '2007-03-18 10:43:31'),
-		array('title' => 'Third Article', 'created' => '2007-03-18 10:43:23', 'updated' => '2007-03-18 10:45:31')
+		array('id' => '1', 'title' => 'First Article', 'created' => '2007-03-18 10:39:23', 'updated' => '2007-03-18 10:41:31'),
+		array('id' => '2', 'title' => 'Second Article', 'created' => '2007-03-18 10:41:23', 'updated' => '2007-03-18 10:43:31'),
+		array('id' => '3', 'title' => 'Third Article', 'created' => '2007-03-18 10:43:23', 'updated' => '2007-03-18 10:45:31')
 	);
 }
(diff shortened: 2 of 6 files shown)


lib/Cake/Test/Fixture/BookFixture.php
```diff
@@ -43,6 +43,7 @@ class BookFixture extends CakeTestFixture {
  * @var array
  */
 	public $records = array(
-		array('id' => 1, 'isbn' => '1234567890', 'title' => 'Faust', 'author' => 'Johann Wolfgang von Goethe')
+		array('id' => 1, 'isbn' => '1234567890', 'title' => 'Faust', 'author' => 'Johann Wolfgang von Goethe'),
+		array('id' => 0, 'isbn' => '0000000001', 'title' => 'Bible', 'author' => 'Twelve Apostles'),
```
markstory: Do you need this?

lib/Cake/Test/Fixture/AnotherArticleFixture.php
```diff
@@ -43,6 +43,7 @@ class AnotherArticleFixture extends CakeTestFixture {
 	public $records = array(
 		array('title' => 'First Article', 'created' => '2007-03-18 10:39:23', 'updated' => '2007-03-18 10:41:31'),
 		array('title' => 'Second Article', 'created' => '2007-03-18 10:41:23', 'updated' => '2007-03-18 10:43:31'),
-		array('title' => 'Third Article', 'created' => '2007-03-18 10:43:23', 'updated' => '2007-03-18 10:45:31')
+		array('title' => 'Third Article', 'created' => '2007-03-18 10:43:23', 'updated' => '2007-03-18 10:45:31'),
+		array('id' => '0', 'title' => 'Zeroth Article', 'created' => '2017-03-17 00:48:23', 'updated' => '2017-03-17 00:48:23'),
```
markstory: Setting an explicit ID in the fixture is not going to work for postgres/sqlserver

lib/Cake/Test/Fixture/BookFixture.php
bancer: It seems the problem is at the point of inserting the records. If we switch the positions of these two records we will get a duplicate primary key exception. I suspect that somewhere `!empty($record['id'])` is used instead of `!isset($record['id'])`. I'll try to dig deeper tonight or during the weekend.

markstory: Another option would be to insert the record inside the test and not modifying the fixtures.

lib/Cake/Model/Datasource/Database/Mysql.php
```diff
@@ -895,8 +895,10 @@ public function insertMulti($table, $fields, $values) {
 				$i++;
 			}
 		}
+		$this->_connection->query('SET @OLD_SQL_MODE=@@SQL_MODE, SQL_MODE="NO_AUTO_VALUE_ON_ZERO"');
```
markstory: We can't keep this code here. It will triple query throughput to mysql for what is a very rare use case.

markstory: Perhaps you can find a different way to do the 0 value insert other than in a fixture's default recordset.

saeideng:
what is executed mysql query before and after these change?

saeideng:
is your problem  on inserting  or on  fetching data?

bancer:
@saeideng Inserting. It is described in MySQL manual - https://dev.mysql.com/doc/refman/5.7/en/sql-mode.html#sqlmode_no_auto_value_on_zero.

lib/Cake/Model/Datasource/Database/Mysql.php
bancer: Maybe to use CakeTestFixture::insert()? Are there any examples how to use it?

saeideng: I think you can insert directly by table::save()

markstory: Yes Model::save() would be how I would do it as well.

ravage84: Just that I get it right, doing it in ``Model::save()`` would mean it would be executed on every save, isn't it?

bancer:
```
	public function insert($db) {
		$db->query('SET @OLD_SQL_MODE=@@SQL_MODE, SQL_MODE="NO_AUTO_VALUE_ON_ZERO"');
		$result = parent::insert($db);
		$db->query('SET SQL_MODE=@OLD_SQL_MODE');
		return $result;
	}
```
Looks like overriding insert() in my fixture solves the problem.

markstory: @bancer That might make the fixture not work for postgres though.

saeideng: @ravage84 
Model::save() will use in unit tests

bancer: I doubt Model::save() would work. If you set 'id' to 0 then model save will try to update an existing record but there is no record with id 0 yet.

PR #10427, opened 2017-03-17
TranslateBehavior's method TranslationField now returns standard table alias for default locale

Robertroid:
TranslateBehavior's method TranslationField always returned the alias for the i18n table for the field argument, be the default locale or not.

This fix compares before anything if the current locale is equals to the default locale for the table. If so, returns the default alias. If not, it behaves as always.


markstory (commented):
Looks good to me.

antograssiot (approved):

Issue #10428, opened 2017-03-17
Repo: Use Github's enhanced license meta data


inoas:
Check: https://github.com/google/guetzli/blob/master/LICENSE
vs https://github.com/cakephp/cakephp/blob/master/LICENSE.txt
You can select the license like so: https://help.github.com/articles/licensing-a-repository/#disclaimer

This will add meta information that github will display and is useful for standardized licenses like GPL, MIT etc.

thinkingmedia:
GitHub's license logic is just being dumb. I think the only thing different here is the `.txt` extension.

We should start a public protest and collect signatures to get GitHub to support `LICENSE.txt` files. I mean, we're CakePHP and if we can not impose our own standards onto other people then wheres the fun in that.

Either they start showing meta data for our license file, or we are moving to BitBucket.

dereuromark:
I would drop the .txt extension - and all works just fine :)

ADmad:
Extension doesn't seem to matter e.g. https://github.com/ADmad/cakephp-glide/blob/master/LICENSE.txt

PR #10429, opened 2017-03-17
Rename LICENSE.txt to LICENSE

dereuromark:
Resolves https://github.com/cakephp/cakephp/issues/10428
Recommended approach anyway - also consistent with most plugins and other libraries.


inoas:
Ref: https://github.com/symfony/symfony/blob/master/LICENSE
Ref: https://github.com/rails/rails/blob/master/MIT-LICENSE (LOL they take MIT-LICENSE but not LICENSE.txt - wtf)

ADmad:
Extension doesn't seem to matter, e.g. https://github.com/ADmad/cakephp-glide/blob/master/LICENSE.txt

inoas:
Maybe then the MIT license text itself needs an update?

cleptric:
> In the file name field, type LICENSE or LICENSE.txt (with all caps).

So we should be fine. Don't know why github isn't adding the header to the branch @dereuromark just created.

dereuromark:
OK, maybe its the additional 2nd data row:

    CakePHP(tm) : The Rapid Development PHP Framework (http://cakephp.org)

That messes up their parsing.

inoas:
See my PR here: https://github.com/cakephp/cakephp/pull/10430

However I can see the usefulness of that line.

PR #10430, opened 2017-03-17
LICENSE.txt update to help github detect

inoas:
Taken from https://choosealicense.com/licenses/mit/ which is also the source for github's licensing features.

### DIFF
--- a/LICENSE.txt
+++ b/LICENSE.txt
@@ -1,6 +1,5 @@
-The MIT License (MIT)
+MIT License
 
-CakePHP(tm) : The Rapid Development PHP Framework (http://cakephp.org)
 Copyright (c) 2005-2017, Cake Software Foundation, Inc. (http://cakefoundation.org)
 
 Permission is hereby granted, free of charge, to any person obtaining a copy


inoas:
Maybe we can change:
`Copyright (c) 2005-2017, Cake Software Foundation, Inc. (http://cakefoundation.org)`
to
`Copyright (c) 2005-2017, Cake Software Foundation, Inc. (http://cakefoundation.org); also see http://cakephp.org`

inoas:
Ref: https://github.com/cakephp/cakephp/issues/10428

markstory:
Looks good to me. Any concerns @phpnut ?

cleptric:
@inoas I looked at the branch in your fork. The license meta isn't showing their either.

inoas:
That's true @ https://github.com/inoas/cakephp/blob/patch-3/LICENSE.txt

inoas:
@markstory these are the steps you need to do: https://help.github.com/articles/adding-a-license-to-a-repository/

I assume you can keep your original license text. All you would need to do is adding LICENSE (instead of LICENSE.txt) then chose MIT in the box that will appear on the right side, then remove LICENSE.txt. I assume LICENSE can then be renamed to LICENSE.txt

inoas:
In my local repo it only worked when I removed the link to cakephp.org and changed the first line title to MIT License

https://github.com/inoas/cakephp/blob/master/LICENSE

PR #10433, opened 2017-03-19
3.next - Make most cookie operations immutable

markstory:
Make most of the cookie operations use immutable patterns. This makes cookies work in a similar way to the request & response. This is also important so that we can ensure that the request/response stay actually immutable.

I've not attempted to solve the `expand()`, `flatten()`, and encryption related features as I want to do those separately. I feel that the expand/flatten methods should happen implicitly when complex values are manipulated, and that encryption should be moved into a standalone middleware with a whitelist of cookies it handles.

### DIFF
--- a/src/Http/Cookie/Cookie.php
+++ b/src/Http/Cookie/Cookie.php
@@ -396,6 +396,34 @@ class Cookie implements CookieInterface
     }
 
     /**
+     * Create a new cookie that will virtually never expire.
+     *
+     * @return static
+     */
+    public function withNeverExpire()
+    {
+        $new = clone $this;
+        $new->expiresAt = Chronos::createFromDate(2038, 1, 1)->format('U');
+
+        return $new;
+    }
+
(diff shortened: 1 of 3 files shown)


src/Http/Cookie/Cookie.php
```diff
@@ -321,16 +391,45 @@ public function isHttpOnly()
...
+    public function withNeverExpire()
+    {
+        $new = clone $this;
+        $new->expiresAt = Chronos::createFromDate(2038, 1, 1)->format('U');
+
+        return $new;
+    }
+
+    /**
+     * Create a new cookie that wil deletes the cookie from the browser
```
ankr: will delete / deletes

markstory: Thanks I will get that fixed.

markstory:
I'll get started on the expand/collapse changes next. This was a part of #10406

PR #10435, opened 2017-03-20
"General error: 2006 MySQL server has gone away" bug

vovantune:
"$this->_connection->query('SELECT 1')" returns PDOStatement object with empty result set.


cleptric:
Could you please add a little bit more informations about what this PR should improve/fix.

PR #10438, opened 2017-03-21
update composer

saeideng:
https://github.com/cakephp/cakephp/blob/master/src/Validation/Validation.php#L17
https://github.com/cakephp/cakephp/blob/master/src/Validation/Validation.php#L23


markstory:
Thanks!

PR #10441, opened 2017-03-23
3.next - Implicit cookie value expansion

markstory:
We can transparently expand and flatten encoded data. This makes cookies easier to work with as the no longer have expanded state that developers have to manually check. While this implicit type changing is a departure from the immutable patterns, I think it is a reasonable trade off for more ergonomic objects.

Refs #10406

### DIFF
--- a/src/Http/Cookie/Cookie.php
+++ b/src/Http/Cookie/Cookie.php
@@ -542,36 +542,6 @@ class Cookie implements CookieInterface
     }
 
     /**
-     * Expands a serialized cookie value
-     *
-     * @return $this
-     */
-    public function expand()
-    {
-        if (!$this->isExpanded) {
-            $this->value = $this->_expand($this->value);
-            $this->isExpanded = true;
-        }
-
-        return $this;
-    }
-
-    /**
-     * Serializes the cookie value to a string
-     *
-     * @return $this
-     */
-    public function flatten()
-    {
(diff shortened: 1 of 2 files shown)


src/Http/Cookie/Cookie.php
```diff
@@ -271,6 +275,8 @@ protected function _setValue($value)
     {
         if (is_array($value)) {
             $this->isExpanded = true;
+        } else {
```
antograssiot: maybe replace this else block by `$this->isExpanded = is_array($value);`

src/Http/Cookie/Cookie.php
```diff
@@ -440,9 +446,11 @@ public function withExpired()
      */
     public function check($path)
     {
-        $this->_isExpanded();
```
antograssiot: I might be wrong but we are missing a `$value = $this->value;` before the `if` because `Hash:::check()` expect an array as the first parameter but `$value` could be `null`here.

src/Http/Cookie/Cookie.php
markstory: Good idea.

src/Http/Cookie/Cookie.php
markstory: You're right! We're also missing tests for check method :cry:

PR #10442, opened 2017-03-23
Fix a bug in paginating multiple results

ferama:
Paginating multiple result without specifying a default model didn't work. This pull request fix the bug

### DIFF
--- a/src/View/Helper/PaginatorHelper.php
+++ b/src/View/Helper/PaginatorHelper.php
@@ -451,8 +451,8 @@ class PaginatorHelper extends Helper
         }
         $isSorted = (
             $sortKey === $table . '.' . $field ||
-            $sortKey === $defaultModel . '.' . $key ||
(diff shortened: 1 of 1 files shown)


markstory:
Could we get a testcase ? Or a way to reproduce the problem you're having so we can build a test case?

ferama:
I was trying to follow the instruction into the docs here https://book.cakephp.org/3.0/en/controllers/components/pagination.html#paginating-multiple-queries
 and here https://book.cakephp.org/3.0/en/views/helpers/paginator.html#paginator-helper-multiple 

```
# adding  this
$this->Paginator->sort('title', ['model' => 'Articles']);
# and this
$this->Paginator->sort('tag', ['model' => 'Tags']);
```
doesn't work. PaginatorHelper doesn't use the model option to build the scoped links. 

If I'm not interpreting something wrong, the pull request fix this.

markstory:
Thanks! I can use that to make a test case.

PR #10446, opened 2017-03-25
3.next - Add Client specific methods to CookieCollection

markstory:
Add the methods that will allow the new CookieCollection to be used in Http\Client. This does not cover adding the backwards compatible aliases, those will be done separately.

There is a bunch of duplication with Http\Client\CookieCollection. But it is looking like that class can be entirely removed at the end of this refactoring.

Refs #10406

### DIFF
--- a/src/Http/Cookie/CookieCollection.php
+++ b/src/Http/Cookie/CookieCollection.php
@@ -65,20 +65,22 @@ class CookieCollection implements IteratorAggregate, Countable
     /**
      * Add a cookie and get an updated collection.
      *
+     * Cookie names do not have to be unique in a collection, but
+     * having duplicate cookie names will change how get() behaves.
+     *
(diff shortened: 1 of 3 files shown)


src/Http/Cookie/CookieCollection.php
```diff
@@ -154,4 +173,146 @@ public function getIterator()
     {
         return new ArrayIterator($this->cookies);
     }
+
+    /**
+     * Add cookies that match the path/domain/expiration to the request.
+     *
+     * This allows CookieCollections to be used a 'cookie jar' in an HTTP client
```
antograssiot: To be used _as_ a ?

antograssiot (approved):

src/Http/Cookie/CookieCollection.php
markstory: Thanks!

PR #10451, opened 2017-03-27
3.next - Add client compatibilty methods to CookieCollection

markstory:
Add the methods that make Http\Cooke\CookieCollection compatible with Http\Client. Next, I'll use the new class in the client.

I've also fixed a few issues around how cookies were being stored. Cookies need to be unique per name;domain;path tuple. Without this constraint expiring cookies becomes very challenging. Also it becomes impossible to update a cookie in the collection.

Refs #10406

### DIFF
--- a/src/Http/Cookie/Cookie.php
+++ b/src/Http/Cookie/Cookie.php
@@ -220,7 +220,8 @@ class Cookie implements CookieInterface
      */
     public function getId()
     {
-        return "{$this->name};{$this->domain};{$this->path}";
+        $name = mb_strtolower($this->name);
+        return "{$name};{$this->domain};{$this->path}";
     }
 
     /**
--- a/src/Http/Cookie/Cookie.php
+++ b/src/Http/Cookie/Cookie.php
@@ -212,6 +212,18 @@ class Cookie implements CookieInterface
     }
 
     /**
+     * Get the id for a cookie
+     *
+     * Cookies are unique across name, domain, path tuples.
+     *
+     * @return string
+     */
+    public function getId()
+    {
+        return "{$this->name};{$this->domain};{$this->path}";
+    }
+
+    /**
      * Gets the cookie name
      *
      * @return string
--- a/src/Http/Cookie/Cookie.php
+++ b/src/Http/Cookie/Cookie.php
@@ -221,6 +221,7 @@ class Cookie implements CookieInterface
     public function getId()
     {
         $name = mb_strtolower($this->name);
+
         return "{$name};{$this->domain};{$this->path}";
     }
 
--- a/src/Http/Cookie/CookieCollection.php
+++ b/src/Http/Cookie/CookieCollection.php
@@ -82,7 +82,7 @@ class CookieCollection implements IteratorAggregate, Countable
     /**
      * Get the first cookie by name.
      *
-     * If the provided name matches a URL (matches `#^https?://#`) this method
+     * If the provided name matches a URL (starts with `http:`) this method
      * will assume you want a list of cookies that match that URL. This is
      * backwards compatible behavior that will be removed in 4.0.0
      *
@@ -92,6 +92,9 @@ class CookieCollection implements IteratorAggregate, Countable
      */
     public function get($name)
     {
+        if (substr($name, 0, 4) === 'http') {
+            return $this->getByUrl($name);
+        }
         $key = mb_strtolower($name);
         foreach ($this->cookies as $cookie) {
             if (mb_strtolower($cookie->getName()) === $key) {
@@ -100,6 +103,22 @@ class CookieCollection implements IteratorAggregate, Countable
         }
 
         return null;
+    }
+
+    /**
+     * Backwards compatibility helper for consumers of Client\CookieCollection
+     *
+     * @param string $url The url to get cookies for.
+     * @return array An array of matching cookies.
+     * @deprecated 3.5.0 Will be removed in 4.0.0. Use addToRequest() instead.
+     */
+    protected function getByUrl($url)
+    {
+        $path = parse_url($url, PHP_URL_PATH) ?: '/';
+        $host = parse_url($url, PHP_URL_HOST);
+        $scheme = parse_url($url, PHP_URL_SCHEME);
+
+        return $this->findMatchingCookies($scheme, $host, $path);
     }
 
     /**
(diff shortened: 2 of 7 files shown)


src/Http/Cookie/CookieCollection.php
```diff
@@ -315,4 +361,81 @@ protected function parseSetCookieHeader($values)
...
+     * Store the cookies contained in a response
+     *
+     * This method operates on the collection in a mutable way for backwards
+     * compatibility reasons. This method should not be used and is only
+     * provided for backwards compatibility.
+     *
+     * @param \Cake\Http\Client\Response $response The response to read cookies from
+     * @param string $url The request URL used for default host/path values.
+     * @return void
+     * @deprecated 3.5.0 Will be removed in 4.0.0. Use `addFromResponse()` instead.
```
saeideng: CookieCollection.php **created** and **deprecated** in 3.5.0 before release?

saeideng: CookieCollection.php **created** and **deprecated** in 3.5.0 ,before release?

src/Http/Cookie/CookieCollection.php
```diff
@@ -315,4 +361,81 @@ protected function parseSetCookieHeader($values)
...
+        $this->removeExpiredCookies($host, $path);
+    }
+
+    /**
+     * Get all cookie data as arrays.
+     *
+     * This method should not be used and is only provided for backwards compatibility.
+     *
+     * @return array
+     * @deprecated 3.5.0 Will be removed in 4.0.0
```
saeideng: CookieCollection.php **created** and **deprecated** in 3.5.0 ,before release?

src/Http/Cookie/Cookie.php
```diff
@@ -212,6 +212,18 @@ public function withName($name)
...
     /**
+     * Get the id for a cookie
+     *
+     * Cookies are unique across name, domain, path tuples.
+     *
+     * @return string
+     */
+    public function getId()
+    {
+        return "{$this->name};{$this->domain};{$this->path}";
```
saeideng: needs here  a `mb_strtolower()` ?

src/Http/Cookie/CookieCollection.php
```diff
@@ -103,6 +106,22 @@ public function get($name)
     }
 
     /**
+     * Backwards compatibility helper for consumers of Client\CookieCollection
+     *
+     * @param string $url The url to get cookies for.
+     * @return array An array of matching cookies.
+     * @deprecated 3.5.0 Will be removed in 4.0.0. Use addToRequest() instead.
```
saeideng: **deprecated** even for protected method?
CookieCollection.php **created** and **deprecated** in 3.5.0 ,before release?

src/Http/Cookie/RequestCookies.php
```diff
@@ -37,33 +37,4 @@ public static function createFromRequest(ServerRequestInterface $request, $cooki
...
         return new static($cookies);
     }
-
-    /**
-     * Checks if the collection has a cookie with the given name
-     *
-     * @param string $name Name of the cookie
-     * @return bool
-     */
-    public function has($name)
```
saeideng: probably needs **deprecated** instead of removing !?

src/Http/Cookie/RequestCookies.php
```diff
@@ -37,33 +37,4 @@ public static function createFromRequest(ServerRequestInterface $request, $cooki
...
-    }
-
-    /**
-     * Get a cookie from the collection by name.
-     *
-     * @param string $name Name of the cookie to get
-     * @throws \InvalidArgumentException
-     * @return Cookie
-     */
-    public function get($name)
```
saeideng: probably needs **deprecated** instead of removing !?

saeideng (changes requested):

src/Http/Cookie/CookieCollection.php
markstory: That is correct for this method. I want to replace the `Http\Client\CookieCollection` with this new class. In order to do that, this new class needs to provide the same public methods as the old class. However, I feel that the old methods are a bit clunky in both name and parameters, so I would like to deprecate them.

Another option I considered was to modify `Http\Client\CookieCollection` to extend this new class and leave the backwards compatibility there. I'm still open to that approach if people think it would be easier to understand in the future.

src/Http/Cookie/CookieCollection.php
markstory: Yes, see my comment below.

src/Http/Cookie/RequestCookies.php
markstory: No, because this method was never part of any actual release.

src/Http/Cookie/RequestCookies.php
markstory: No, because this method was never part of any actual release.

markstory: Also both this method and `has()` are implemented in this class's parent. It is worth recognizing that this class will be removed very soon as well.

src/Http/Cookie/CookieCollection.php
saeideng: anyway this method (` protected function getByUrl()`) is with protected scope and no need to this line

markstory: Ok.

src/Http/Cookie/CookieCollection.php
```diff
@@ -65,24 +64,25 @@ public function count()
...
+        $new->cookies[$cookie->getId()] = $cookie;
 
         return $new;
     }
 
     /**
      * Get the first cookie by name.
      *
-     * If the provided name matches a URL (matches `#^https?://#`) this method
+     * If the provided name matches a URL (starts with `http:`) this method
```
antograssiot: the `:`should be there I think

markstory:
I've been thinking about this some more. I'm going to push all the backwards compatibility methods into the `Http\Client\CookieCollection` and extending the new class. I think this will keep the new code cleaner/simpler.

src/Http/Cookie/CookieCollection.php
```diff
@@ -14,6 +14,7 @@
 namespace Cake\Http\Cookie;
 
 use ArrayIterator;
+use Cake\Http\Client\Response as ClientResponse;
```
saeideng: you can remove this line :)

saeideng:
extending class and use of 
` Cake\Http\CookieCollection::$cookies` in `Cake\Http\Client\CookieCollection`
means ,we have some cookie in both Collection ! ? :worried:

markstory:
@saeideng I don't understand your question. There should only be one list of cookies in both collections?

PR #10457, opened 2017-03-29
Getter/Setter/Clear for Query::contain()

burzum:


src/ORM/Query.php
```diff
@@ -377,23 +377,159 @@ public function eagerLoader(EagerLoader $instance = null)
      */
     public function contain($associations = null, $override = false)
```
dereuromark: Should we deprecate the "I do it all" method here then?

src/ORM/Query.php
```diff
@@ -377,23 +377,159 @@ public function eagerLoader(EagerLoader $instance = null)
...
+     * ```
+     * // Use special join conditions for getting an Articles's belongsTo 'authors'
+     * $query->contain([
+     *     'Authors' => [
+     *         'foreignKey' => false,
+     *         'queryBuilder' => function ($q) {
+     *             return $q->where(...); // Add full filtering conditions
+     *         }
+     *     ]
+     * ]);
```
dakota: Should probably update these examples

src/ORM/Query.php
```diff
@@ -377,23 +377,159 @@ public function eagerLoader(EagerLoader $instance = null)
...
-        }
-
         $result = $loader->contain($associations);
         $this->_addAssociationsToTypeMap($this->repository(), $this->getTypeMap(), $result);
 
         return $this;
     }
 
     /**
+     * Clears the contained assocations form the current query.
```
dakota: Typo, 'from'

src/ORM/Query.php
```diff
@@ -793,6 +930,10 @@ public function __clone()
      */
     public function count()
     {
+        return $this->getCount();
+    }
+
+    public function getCount() {
```
burzum: I accidentally committed this part as well. However, do we want to have getCount() all all? I've got the impression that some methods are named on purpose like their function names in the DB systems. So do we want to use getters here as well?

src/ORM/Query.php
burzum: Done.

src/ORM/Query.php
inoas: I think `count()` is also good because here a setter would not make sense. It will also be in line with the sqlish functions you can create via cakephp3 orm: https://github.com/cakephp/cakephp/blob/master/src/ORM/Query.php#L43 <- that would not become `getMin()`

@lorenzo what's your take here?

saeideng: `count()` is good
because we have not `setCount()`

src/ORM/Query.php
```diff
@@ -370,30 +370,167 @@ public function eagerLoader(EagerLoader $instance = null)
...
+     *     ]
+     * ]);
+     * ```
+     *
+     * @param array|string $associations List of table aliases to be queried.
+     * @param bool $override Whether override previous list with the one passed
+     * defaults to merging previous list with the new one.
+     * @return $this
+     */
+    public function setContain($associations, $override = false)
```
markstory: The query builder is one of the few places where set prefixes become a bit clunky. If all the methods like where(), order(), and group() all had set  prefixes I think the interface would be less pleasant to use. However, I think adding get/clear methods makes sense. Is there a way we could deprecate the other modes from `contain()` and add the get/clear methods?

src/ORM/Query.php
markstory: I too would stick with count(). The query builders are supposed to read like sql when possible.

src/ORM/Query.php
```diff
@@ -370,30 +370,167 @@ public function eagerLoader(EagerLoader $instance = null)
...
 
         return $this;
     }
 
     /**
+     * Clears the contained associations from the current query.
+     *
+     * @return $this
+     */
+    public function clearContain()
```
markstory: New methods should have tests.

src/ORM/Query.php
```diff
@@ -370,30 +370,167 @@ public function eagerLoader(EagerLoader $instance = null)
      * If called with an empty first argument and `$override` is set to true, the
      * previous list will be emptied.
      *
+     * @deprecated Use setContain(), getContain() and clearContain() instead.
```
markstory: Should the find() option be updated to use the undeprecated method as well?

lorenzo:
I really don't like this change. Maybe only the `clearContain` method make sense to me. In general, I would not like to change the look and feel of the query builder

dereuromark:
@lorenzo We must at least get rid of the `@return array|$this` part, that is clear indicator this does too much.
If we can split it at least in some setting and getting so chaining is always $this and not suddenly sth else, that would probably be OK then. And it looks like this is aiming to achieve it.

markstory:
@dereuromark @lorenzo  Could the existing method eventually become a setter and the get/reset modes move into separate methods? That should help resolve the return type issues and also keep the main use cases of the present method the same.

saeideng:
I think we can use x()as alias of setX() in 4.x for just setting with return $this
And getX() just for getting
And we can remove setX() in 5.x

ADmad:
> I think we can use x()as alias of setX() in 4.x for just setting with return $this
> And getX() just for getting

This sounds reasonable to me. I have already stated earlier my dislike for `setX()` methods for query. `$query->setWhere()->setContain()->setOrder()` looks pretty ugly to me, plus the editor/IDE gives you decent autocompletion only after typing 3 extra characters.

lorenzo:
Yeah, I'm ok with removing the getter part in `contain` We can just deprecate it and tell people to use `$query->eagerLoader()->...` for getting the info they need

saeideng:
I am :+1: with removing getter code 
Even is good to throw exception/error when user want to use getter mode

But i prefer to having getX() for getter

src/ORM/Query.php
burzum: OK

burzum:
@lorenzo why not keeping getContain()? I know I can use `$query->eagerLoader()->...` but I bet most people don't and I like the convenience of not having to call getEagerLoader().

markstory:
Do you frequently get the current containments? We can have simpler/smaller classes by avoiding 'sugar' methods that don't get used often.

inoas:
IMHO: I'd love to have in general an easy way a.) to introspect/traverse built-query and b.) remove parts of a built-query (maybe).

However that should in no way taint the ease-of-use query-building that cake3 does ❤️  because that is like 95% of use cases.

burzum:
@markstory, two times in my current project. So how do we proceed with this PR?

src/ORM/Query.php
```diff
@@ -394,6 +393,29 @@ public function contain($associations = null, $override = false)
...
+
+        return $this;
+    }
+
+    /**
+     * Gets the contained associations
+     *
+     * @return array
+     */
+    public function getContain()
```
saeideng: needs a test

burzum: I'm not putting that effort in until somebody decides to keep it or to remove it.

burzum:
@markstory so what is needed to get this merged?

By the way, our CI system reports this continuously for lots of queries and the message is correct:

> The expression $query->contain(array('R...paniesTranslations'))); of type array|Cake\ORM\Query adds the type array to the return on line 406 which is incompatible with the return type documented by App\Model\Table\ProjectsTable::findView of type Cake\ORM\Query.

This is pretty annoying and happens for some other occasions and methods as well and should be addressed in the framework where possible IMHO.

src/ORM/Query.php
```diff
@@ -379,12 +379,11 @@ public function contain($associations = null, $override = false)
...
         $loader = $this->getEagerLoader();
         if ($override === true) {
-            $loader->clearContain();
-            $this->_dirty();
+            $this->clearContain();
         }
 
         if ($associations === null) {
-            return $loader->contain();
+            return $this->getEagerLoader()->contain();
```
markstory: You don't need the `getEagerLoader()` call here, this method already has a `$loader` instance available.

PR #10459, opened 2017-03-30
3.next - Update Http\Client to use undeprecated cookie methods.

markstory:
Update the Http\Client and Stream adapter to not use deprecated methods on CookieCollection. I've also fixed a small issue where cookie headers that were set in the request would be dropped.

Refs #10406

### DIFF
--- a/src/Http/Client/Adapter/Stream.php
+++ b/src/Http/Client/Adapter/Stream.php
@@ -147,19 +147,7 @@ class Stream
     {
         $headers = [];
         foreach ($request->getHeaders() as $name => $values) {
-            if ($name !== 'Cookie') {
-                $headers[] = sprintf('%s: %s', $name, implode(", ", $values));
-            }
-        }
-
-        $cookieHeader = $request->getHeaderLine('Cookie');
-        $cookies = [];
-        foreach ($request->cookies() as $name => $value) {
-            $cookies[] = "$name=$value";
-        }
-        $cookieData = implode('; ', $cookies);
-        if ($cookieData || $cookieHeader) {
-            $headers[] = 'Cookie: ' . implode('; ', [$cookieHeader, $cookieData]);
+            $headers[] = sprintf('%s: %s', $name, implode(", ", $values));
         }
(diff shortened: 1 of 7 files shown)


src/Http/Client/Adapter/Stream.php
```diff
@@ -147,15 +147,19 @@ protected function _buildHeaders(Request $request, $options)
...
+        $cookieHeader = $request->getHeaderLine('Cookie');
         $cookies = [];
         foreach ($request->cookies() as $name => $value) {
             $cookies[] = "$name=$value";
         }
-        if ($cookies) {
-            $headers[] = 'Cookie: ' . implode('; ', $cookies);
+        $cookieData = implode('; ', $cookies);
+        if ($cookieData || $cookieHeader) {
+            $headers[] = 'Cookie: ' . implode('; ', [$cookieHeader, $cookieData]);
```
markstory: Just realised I am missing a test​ where a cookie is both in the jar and defined in options. I will add that.

src/Http/Cookie/CookieCollection.php
```diff
@@ -177,18 +177,25 @@ public function getIterator()
      * when this method is called will be applied to the request.
      *
      * @param \Psr\Http\Message\RequestInterface $request The request to update.
+     * @param array $extraCookies Associative array of additional cookies to add into the request. This
+     *   is useful when you have cookie data from outside the collection you want to send.
      * @return \Psr\Http\Message\RequestInterface An updated request.
      */
-    public function addToRequest(RequestInterface $request)
+    public function addToRequest(RequestInterface $request, array $extraCookies = [])
```
saeideng: we can not send object  `new Cookie('api', 'A', null, '/api', 'example.com'))` to $extraCookies
that is true?

markstory: No. I thought building new objects and a new collection would be wasteful when all we need is a key/value list.

lorenzo:
This looks good to me, it is a nice cleanup

PR #10461, opened 2017-03-30
Add web fonts to be treated as binary in .gitattributes

plazareff:
After spending hours before noticing that cakephp's .gitattributes file was overcharging mine, I thought useful to add webfont files patterns to the main distribution :-)

This makes webfonts not being broken by git during checkout/pull.

Hope this helps !

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -28,6 +28,11 @@
 *.gif binary
 *.ico binary
 *.mo binary
+*.ttf	binary
+*.woff 	binary
+*.woff2	binary
+*.eot	binary
+*.otf	binary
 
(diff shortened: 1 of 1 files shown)


markstory:
How was the gitattributes file in `vendor/cakephp/cakephp` effecting your webroot directory?

dereuromark:
Those are already in cakephp/app file, this should suffice
Adding this to core is not relevant for you and sounds wrong to me :-1:

plazareff:
Uh oh, then where is the original .gitattributes file installed at the root of a project ? (via composer)

Issue #8053, opened 2016-01-17
Hash::extract() is transforming values.


dereuromark:
I think Hash::extract() does too much, it should maybe not convert the value itself into an array, but rather return it completely.
This would allow for string casts afterwards, whereas right now it would need information on how the array structure of each transformed object looks like.

Example case: https://github.com/FriendsOfCake/cakephp-csvview/pull/55

We just want the User.created DateTime object from the entity, so we can pass it to _renderRow().
But instead the array of it is returned and we can never transform it again as string without deep knowledge of the structure of this specific array.

Might be a side effect of recent changes, maybe a regression even.

markstory:
I don't see how this is a defect. It sounds like a change in behavior that would break existing applications. There haven't been any changes to Hash since Oct 2015, so the behavior has been consistent for quite some time.

dereuromark:
Is there an alternative to not force the objects to break into arrays upon simple Hash read?
It will blow up in the future with custom objects in the entities, that supply a toString version but don't have the chance to do so.

markstory:
What do you mean by 'force the objects to break into arrays'? Are you getting the array cast of an object instead of the object? Does the object in question implement `toArray` by any chance?

dereuromark:
It is our CakePHP Time object, see the linked issue in CsvView.
There we get an array back with "date" as one key. We would have expected the Time object itself which then can safely output itself as string. The array itself cannot anymore.

markstory:
Odd, I don't really get what is going on in that other issue, but are you attempting to extract a property out of an entity and the property's value is a Time instance?

dereuromark:
And after extracting it is an array, not a Time instance, making it impossible to work on it anymore.

josegonzalez:
Might be because of this [line](https://github.com/FriendsOfCake/cakephp-csvview/blob/master/src/View/CsvView.php#L303).

dereuromark:
@josegonzalez No, it doesnt even run into the isset() check
https://github.com/FriendsOfCake/cakephp-csvview/blob/master/src/View/CsvView.php#L301 already has the array instead of the object.

dereuromark:
The problem is
```
                if (is_object($item) && method_exists($item, 'toArray')) {
                    $item = $item->toArray();
                }
```
This should only be done for iterable items we whitelist, not for all possible objects IMO.
Or at least should only do so if the lookup key goes beyond the current level.
[...]

PR #10464, opened 2017-03-30
WIP Issue/text insert time

steinkel:
While playing with Text::insert method with flattened entities I got to this use case where objects (Time) cannot be cast to (int) by asort here https://github.com/cakephp/cakephp/blob/3.4.0-RC1/src/Utility/Text.php#L198.

Removing this line makes all tests pass, so I guess we need additional unit tests to ensure we need to asort the data array. I'm leaving this PR open in case you could point me to either add additional test cases or it's fine to remove the asort?

Thanks,

### DIFF
--- a/src/Utility/Text.php
+++ b/src/Utility/Text.php
@@ -195,8 +195,6 @@ class Text
             return ($options['clean']) ? static::cleanInsert($str, $options) : $str;
         }
 
-        asort($data);
-
         $dataKeys = array_keys($data);
         $hashKeys = array_map('crc32', $dataKeys);
         $tempData = array_combine($dataKeys, $hashKeys);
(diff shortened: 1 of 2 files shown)


dereuromark:
Sounds like related to https://github.com/cakephp/cakephp/issues/8053 ?

src/Utility/Text.php
```diff
@@ -195,8 +195,6 @@ public static function insert($str, $data, array $options = [])
             return ($options['clean']) ? static::cleanInsert($str, $options) : $str;
         }
 
-        asort($data);
```
markstory: This was added in bf2829a8 which predates us having any test coverage. I suspect it is safe to remove if the existing tests pass.

PR #10466, opened 2017-03-31
3.next - Use CookieCollection in Client responses

markstory:
Use the new CookieCollection objects in Client responses. This allows us to remove one copy of the Cookie parsing code we had.

Refs #10406

### DIFF
--- a/src/Http/Client/Response.php
+++ b/src/Http/Client/Response.php
@@ -13,6 +13,7 @@
  */
 namespace Cake\Http\Client;
 
+use Cake\Http\Cookie\CookieCollection;
 use Psr\Http\Message\ResponseInterface;
 use RuntimeException;
 use Zend\Diactoros\MessageTrait;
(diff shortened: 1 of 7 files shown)


src/Http/Client/Response.php
```diff
@@ -453,11 +434,43 @@ public function getCookie($name)
...
+        }
+        $this->cookies = CookieCollection::createFromHeader($this->getHeader('Set-Cookie'));
+    }
+
+    /**
+     * Property accessor for `$this->cookies`
+     *
+     * @return array Array of Cookie data.
+     */
+    protected function _getCookies()
```
lorenzo: Why have a this method if the public method is supposed to do the same as the public method with the same name? Maybe just make them the same method?

markstory: This is for the magic properties that the response exposes. `$response->cookies` is provided by this method and the _get` prefix is part of the required convention for exposing properties through methods on this class.

ravage84: @markstory May be add ``This is for the magic properties that the response exposes. $response->cookies is provided by this method.`` or something like that as the long description of this method. This way it becomes clear.

PR #10467, opened 2017-04-02
EntityContext: Introspect 'null' to determine use of 'required' attribute

rchavik:
Allow browsers to take advantage of html5 validation

### DIFF
--- a/src/View/Form/EntityContext.php
+++ b/src/View/Form/EntityContext.php
@@ -541,7 +541,7 @@ class EntityContext implements ContextInterface
         $parts = explode('.', $field);
         $table = $this->_getTable($parts);
         $column = (array)$table->getSchema()->column(array_pop($parts));
-        $whitelist = ['length' => null, 'precision' => null];
+        $whitelist = ['length' => null, 'precision' => null, 'null' => null];
 
         return array_intersect_key($column, $whitelist);
     }
--- a/src/View/Helper/FormHelper.php
+++ b/src/View/Helper/FormHelper.php
@@ -1396,6 +1396,10 @@ class FormHelper extends Helper
             $options += ['empty' => false];
         }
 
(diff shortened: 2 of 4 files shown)


ADmad:
I don't really understand what this patch is meant for.

rchavik:
It adds 'required' attribute in form input element when field is not null. that allows the browser html5 validation to be triggered.

saeideng:
When you use require validator,that added automatically

ADmad:
I don't really like adding the required attribute directly based on schema. As @saeideng mentioned it's already added based on your validation rules so you should add proper validation for your fields instead.

markstory:
Moving to 3.5, as changing the default behavior of FormHelper can cause unexpected behavior that shouldn't be part of bugfix releases.

rchavik:
Ok.  I'm ok to close then.

dereuromark:
I like the idea of the database schema (default and null) reflect in the form output defaults.
But as said before the schema-influenced (baked) validation rules then already do that.
It could get weird, if the schema says required and the validation rules say not required, which one takes precedence then?

markstory:
@dereuromark Good point about needing to reflect the validation rules, also shouldn't this kind of code be located in the `EntityContext::isRequired()` with the other logic?

rchavik:
I'll try to rework this based on input.

Thanks

markstory:
Ping

rchavik:
I've not had time to get back to this yet.  As this is not critical, let's close this for now.

I'll reopen if I had a chance to pursue this again.

rchavik:
Branch has been deleted.

I will keep it in my fork:
- https://github.com/rchavik/cakephp/tree/3.4-required-fields This PR
- https://github.com/rchavik/cakephp/tree/3.4-required-fields-new Improvement on this PR based on comments. test not passing.

PR #10470, opened 2017-04-03
3.next - Expose CookieCollection in ServerRequest

markstory:
I've chosen to not use CookieCollection as the internal storage as  keeping the public property in-sync would be hard/impossible. This  approach also avoid eagerly allocating memory for the cookie collection and the cookie objects.

Refs #10406

### DIFF
--- a/src/Http/Cookie/CookieCollection.php
+++ b/src/Http/Cookie/CookieCollection.php
@@ -68,7 +68,7 @@ class CookieCollection implements IteratorAggregate, Countable
     /**
      * Create a new collection from the cookies in a ServerRequest
      *
-     * @param \Psr\Http\Message\ServerRequestInterface $request The array of set-cookie header values.
+     * @param \Psr\Http\Message\ServerRequestInterface $request The request to extract cookie data from
      * @return static
      */
     public static function createFromServerRequest(ServerRequestInterface $request)
--- a/src/Http/Cookie/CookieCollection.php
+++ b/src/Http/Cookie/CookieCollection.php
@@ -68,7 +68,7 @@ class CookieCollection implements IteratorAggregate, Countable
     /**
      * Create a new collection from the cookies in a ServerRequest
      *
-     * @param array $header The array of set-cookie header values.
+     * @param \Psr\Http\Message\ServerRequestInterface $request The array of set-cookie header values.
      * @return static
      */
     public static function createFromServerRequest(ServerRequestInterface $request)
--- a/src/Http/Cookie/CookieCollection.php
+++ b/src/Http/Cookie/CookieCollection.php
@@ -21,6 +21,7 @@ use InvalidArgumentException;
 use IteratorAggregate;
 use Psr\Http\Message\RequestInterface;
 use Psr\Http\Message\ResponseInterface;
+use Psr\Http\Message\ServerRequestInterface;
 
 /**
  * Cookie Collection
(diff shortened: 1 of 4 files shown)


src/Http/Cookie/CookieCollection.php
```diff
@@ -61,7 +62,24 @@ public static function createFromHeader(array $header)
...
+     *
+     * @param \Psr\Http\Message\ServerRequestInterface $request The array of set-cookie header values.
+     * @return static
+     */
+    public static function createFromServerRequest(ServerRequestInterface $request)
+    {
+        $data = $request->getCookieParams();
+        $cookies = [];
+        foreach ($data as $name => $value) {
+            $cookies[] = new Cookie($name, $value);
```
saeideng: This should contain the patch,domain from request

markstory: We have no way of knowing what the cookie's path/domain is on the client though.

src/Http/Cookie/CookieCollection.php
```diff
@@ -61,7 +62,24 @@ public static function createFromHeader(array $header)
...
         $cookies = static::parseSetCookieHeader($header);
 
-        return new CookieCollection($cookies);
+        return new static($cookies);
+    }
+
+    /**
+     * Create a new collection from the cookies in a ServerRequest
+     *
+     * @param \Psr\Http\Message\ServerRequestInterface $request The array of set-cookie header values.
```
saeideng: $request==The array of set-cookie header values. ??

src/Http/Cookie/CookieCollection.php
saeideng: maybe 
$request->host() or $request->base or ...

markstory: If those cookies are then updated and sent back to the client we might corrupt the cookie data.

src/Http/Cookie/CookieCollection.php
markstory: No a just a request instance.

src/Http/Cookie/CookieCollection.php
saeideng: yes i know

markstory: What would be the advantage​of setting path/domain if it can corrupt the cookies in the client?

saeideng: Probably for when it merge with other cookie 
Because cookieCollection will use name;path;domain key
Even that method can be have second arg for include/not include path/domain ,that people able to choose one of that

markstory: How would that extra argument work?

saeideng:
```
 public static function createFromServerRequest(ServerRequestInterface $request,$withPathDomain=false)
    {
        $data = $request->getCookieParams();
        $path=$request->path();
        $domain=$request->domain();
        $cookies = [];
        foreach ($data as $name => $value) {
             if($withPathDomain){
                       $cookies[] = new Cookie($name, $value,$path $domain);
            }else{
                       $cookies[] = new Cookie($name, $value);
             }
            
        }

        return new static($cookies);
     }
```

markstory: How would someone use that from the ServerRequest though?

saeideng:
```
    public function getCookieCollection($withPathInfo=false)
    {
        return CookieCollection::createFromServerRequest($this,$withPathInformation);
    }
```

saeideng: even better this , having new method on cookieCollection for insert/delete  path/domain from all of cooke

saeideng: so, we can 
```
//ServerRequest
 public function getCookieCollection()
    {
        return CookieCollection::createFromServerRequest($this);
    }
```
usage:
```
$this->request->getCookieCollection()->removePathDomain();
$this->request->getCookieCollection()->withPathDomain($this->request);
$this->request->getCookieCollection()->withPathDomain([path=>'....',domain=>'....']);
```

saeideng: public static function pathDomainFromServerRequest(ServerRequestInterface $request)
    {
        $path=$request->path();
        $domain=$request->domain();
        $cookies = [];
        foreach ($this->cookie as $name => $value) {
             //this is pseudo code
                       $cookies[] = new Cookie($name, $value,$path $domain);  
        }

        return new static($cookies);
     }

Issue #10374, opened 2017-03-07
Add Https or Security Middleware


chrisShick:
This is a (multiple allowed):

* [] bug
* [x] enhancement
* [x] feature-discussion (RFC)

* CakePHP Version: 3.4.x and 4.x

I feel that we should have an Https (or Security) middleware as one of the default middlewares. I feel it is definitely something that would benefit the community and could replace the Security component in 4.

markstory:
What would it do? 'Security' is a pretty broad topic.

chrisShick:
Well, it would do what the Security component currently does or if we can just stick with it redirecting to https securely.

cleptric:
> redirecting to https securely.
This should handle the webserver

Oxicode:
Algo así? https://github.com/BePsvPT/secure-headers/blob/master/config/secure-headers.php

cleptric:
Closing as the PR which implemented this was merged.

PR #10474, opened 2017-04-03
Adding a security headers middleware

burzum:
Refs #10374

Tests are coming

### DIFF
--- a/src/Routing/Middleware/SecurityMiddleware.php
+++ b/src/Routing/Middleware/SecurityMiddleware.php
@@ -36,7 +36,7 @@ class SecurityMiddleware
      *
      * Sets the header value for it to 'nosniff'
      *
-     * @link Reference: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Content-Type-Options
+     * @link https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Content-Type-Options
      * @return $this
      */
     public function noSniff()
--- a/src/Routing/Middleware/SecurityMiddleware.php
+++ b/src/Routing/Middleware/SecurityMiddleware.php
@@ -170,11 +170,10 @@ class SecurityMiddleware
     {
         if (!in_array($value, $allowed)) {
             throw new InvalidArgumentException(sprintf(
-                    'Invalid arg `%s`, use one of these: %s',
-                    $value,
-                    implode(', ', $allowed)
-                )
-            );
+                'Invalid arg `%s`, use one of these: %s',
+                $value,
+                implode(', ', $allowed)
+            ));
         }
     }
 
@@ -188,10 +187,11 @@ class SecurityMiddleware
      */
     public function __invoke(ServerRequestInterface $request, ResponseInterface $response, $next)
     {
+        $response = $next($request, $response);
         foreach ($this->headers as $header => $value) {
             $response = $response->withHeader($header, $value);
         }
 
-        return $next($request, $response);
+        return $response;
     }
 }
--- a/src/Routing/Middleware/SecurityMiddleware.php
+++ b/src/Routing/Middleware/SecurityMiddleware.php
@@ -34,10 +34,9 @@ class SecurityMiddleware
     /**
      * X-Content-Type-Options
      *
-     * Reference: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Content-Type-Options
+     * Sets the header value for it to 'nosniff'
      *
-     * Available Value: 'nosniff'
-     *
+     * @link Reference: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Content-Type-Options
      * @return $this
      */
     public function noSniff()
@@ -50,10 +49,9 @@ class SecurityMiddleware
     /**
      * X-Download-Options
      *
-     * Reference: https://msdn.microsoft.com/en-us/library/jj542450(v=vs.85).aspx
+     * Sets the header value for it to 'noopen'
      *
-     * Available Value: 'noopen'
-     *
+     * @link https://msdn.microsoft.com/en-us/library/jj542450(v=vs.85).aspx
      * @return $this
      */
     public function noOpen()
@@ -66,12 +64,9 @@ class SecurityMiddleware
     /**
      * Referrer-Policy
      *
-     * Reference: https://w3c.github.io/webappsec-referrer-policy
-     *
-     * Available Value: 'no-referrer', 'no-referrer-when-downgrade', 'origin', 'origin-when-cross-origin',
-     *                  'same-origin', 'strict-origin', 'strict-origin-when-cross-origin', 'unsafe-url'
-     *
-     * @param string $policy Policy value
+     * @link https://w3c.github.io/webappsec-referrer-policy
+     * @param string $policy Policy value. Available Value: 'no-referrer', 'no-referrer-when-downgrade', 'origin', 'origin-when-cross-origin',
+     *        'same-origin', 'strict-origin', 'strict-origin-when-cross-origin', 'unsafe-url'
      * @return $this
      */
     public function setReferrerPolicy($policy = 'same-origin')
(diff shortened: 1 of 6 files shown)


src/Routing/Middleware/SecurityMiddleware.php
```diff
@@ -0,0 +1,197 @@
...
+     * @return void
+     */
+    protected function checkValues($value, array $allowed)
+    {
+        if (!in_array($value, $allowed)) {
+            throw new InvalidArgumentException(sprintf(
+                'Invalid arg `%s`, use one of these: %s',
+                    $value,
+                    implode(', ', $allowed)
+                )
```
burzum: Fuck spaces... 😡

src/Routing/Middleware/SecurityMiddleware.php
```diff
@@ -0,0 +1,197 @@
...
+     *
+     * @param \Psr\Http\Message\ServerRequestInterface $request The request.
+     * @param \Psr\Http\Message\ResponseInterface $response The response.
+     * @param callable $next Callback to invoke the next middleware.
+     * @return \Psr\Http\Message\ResponseInterface A response
+     */
+    public function __invoke(ServerRequestInterface $request, ResponseInterface $response, $next)
+    {
+        foreach ($this->headers as $header => $value) {
+            $response = $response->withHeader($header, $value);
```
markstory: I think you want to modify the response after calling $next.  Right now headers will be lost if the lower layers create a new instance.

ADmad:
Why do `noOpen()` and `noSniff()` don't have "set" prefix like other methods? Also for methods which don't mutate current instance "with" prefix is used at other places in the framework.

burzum:
@ADmad because you don't set anything, they have a single value, no argument is passed to the method. If you like to rename them, then name them enableNoOpen() or something like that?

ADmad:
Sorry I misread the code, the `setFoo()` methods are just modifying protected variable, so they are fine.

src/Routing/Middleware/SecurityMiddleware.php
```diff
@@ -0,0 +1,195 @@
...
+            'unsafe-url'
+        ];
+
+        $this->checkValues($policy, $available);
+        $this->headers['referrer-policy'] = $policy;
+
+        return $this;
+    }
+
+    /**
```
burzum: I have no idea what's wrong here by looking at the code...

src/Routing/Middleware/SecurityMiddleware.php
```diff
@@ -0,0 +1,197 @@
...
+     *
+     * @throws \InvalidArgumentException Thrown when a value is invalid.
+     * @param string $value
+     * @param array $allowed
+     * @return void
+     */
+    protected function checkValues($value, array $allowed)
+    {
+        if (!in_array($value, $allowed)) {
+            throw new InvalidArgumentException(sprintf(
```
burzum: Nor do I have here. Do really have to put the whole code into one line to make it less readable but to pass stickler?

burzum:
@ADmad 🤣 thanks for showing that the `_` prefix is pretty useful 😆 I'm still against removing it, it's silly even PSR 💩 is saying it's optional and not forbidden.

src/Routing/Middleware/SecurityMiddleware.php
robertpustulka: Maybe put a message into variable.

src/Routing/Middleware/SecurityMiddleware.php
```diff
@@ -0,0 +1,197 @@
...
+     * Security related headers to set
+     *
+     * @var array
+     */
+    protected $headers = [];
+
+    /**
+     * X-Content-Type-Options
+     *
+     * Reference: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Content-Type-Options
```
ADmad: Instead of "Reference:" you can use `@see` tag. Same for other methods.

burzum: Why `@see` and not `@link` for a link?

I'll change the doc blocks when I have time, probably tomorrow.

ADmad: Either works.

tests/TestCase/Routing/Middleware/SecurityMiddlewareTest.php
```diff
@@ -0,0 +1,101 @@
...
+            return $res;
+        };
+
+        $middleware = new SecurityMiddleware();
+        $middleware
+            ->setCrossDomainPolicy()
+            ->setReferrerPolicy()
+            ->setXFrameOptions()
+            ->setXssProtection()
+            ->noOpen()
```
lorenzo: I like a lot more the `noOpen` vs `setNoOpen` Can we do the same for the other methods?

dereuromark: setXssProtection could be enableXssProtection etc.

burzum: @lorenzo I named them intentionally this way because the set* methods have actually values you can set while the other two don't. You just "enable" them. Thought this makes the difference obvious.

ADmad:
> @ADmad :rofl: thanks for showing that the _ prefix is pretty useful :laughing: I'm still against removing it, it's silly even PSR :hankey: is saying it's optional and not forbidden.

I am of the same opinion and keep using it in my personal code :)

dereuromark:
@ADmad We seem to all agree :) Underscores are fine, and are allowed to be used.

src/Routing/Middleware/SecurityMiddleware.php
```diff
@@ -34,10 +34,9 @@ class SecurityMiddleware
     /**
      * X-Content-Type-Options
      *
-     * Reference: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Content-Type-Options
-     *
-     * Available Value: 'nosniff'
+     * Sets the header value for it to 'nosniff'
      *
+     * @link Reference: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Content-Type-Options
```
ADmad: "Reference:  " should be removed.

src/Routing/Middleware/SecurityMiddleware.php
```diff
@@ -0,0 +1,183 @@
...
+namespace Cake\Routing\Middleware;
+
+use InvalidArgumentException;
+use Psr\Http\Message\ResponseInterface;
+use Psr\Http\Message\ServerRequestInterface;
+
+/**
+ * Handles common security headers in a convenient way
+ */
+class SecurityMiddleware
```
markstory: Does the name `SecurityHeaderMiddleware` clarify the scope/purpose more?

robertpustulka: @markstory 👍 at first glance I though this is a port of `SecurityComponent`.

burzum: I've had the exact same discussion with Jose ~5h ago and proposed this name. ;) I just returned from training, will do so tomorrow.

burzum:
Can somebody please restart the Appveyor job? It failed because it couldn't get the Github repo...

robertpustulka:
@burzum would force pushing restart checks?

saeideng:
@burzum you can close/reopen the PR for restart it :)

burzum:
@saeideng or somebody just restarts it or merges it :) Somebody in the team must have access to it.

markstory:
@burzum I can't but I think @lorenzo can.

src/Routing/Middleware/SecurityMiddleware.php
ravage84: Personally, I treat it this way:
``@see`` for class API related references and ``@link`` for anything else I have to link to, e.g. website or ressource PDF.

markstory:
Is `Routing` the right namespace for this? It have anything to do with routing. Perhaps it should go in the `Http` namespace instead?

sukihub:
Hi :) Do you plan adding Content Security Policy headers to this? There's also a nice little library that might be useful https://github.com/paragonie/csp-builder

Also, guys in GitHub and Twitter seem to think, that modifying secure headers dynamically in actions is a good idea:

https://github.com/twitter/secureheaders/blob/master/docs/per_action_configuration.md

> Historically, it has been relatively tedious to make dynamic changes to our policy per endpoint (i.e. we didn’t do it). But, with some recent development by @oreoshake to the Secure Headers library, it is now much easier for us going forward. For example, connections to api.braintreegateway.com only occur on payment related pages. We can now enforce a unique exception to our policy, appending the third-party host only on pages that need to connect to the payment endpoint. Over time we hope to lock down other unique connection endpoints using dynamic CSP policies.

https://githubengineering.com/githubs-csp-journey/

So I was thinking that maybe having component for dynamic changes would be nice.
Or is it more library than Cake's core?

lorenzo:
Maybe open a new ticket @sukihub? So we can discuss it there

sukihub:
Thanks, done! :)

Issue #9474, opened 2016-09-17
(Automatically) Convert empty association properties for `SAVE_REPLACE` compatibility?


ndm2:
This is a (multiple allowed):
- [ ] bug
- [x] enhancement
- [x] feature-discussion (RFC)

---

Let's say we have a form that supports adding/removing `hasMany`/`belongsToMany` associated records that are using the `SAVE_REPLACE` save strategy.

``` php
$this->Form->input('parent_field');
$this->Form->input('association.0.field');
$this->Form->input('association.1.field');
$this->Form->input('association.2.field');
// ...
```

When removing for example `association.2.field`, the saving process would delete the record accordingly. When however removing all of them, so that we'd be left with only

``` php
$this->Form->input('parent_field');
```

then the association property wouldn't be affected anymore at all, and nothing would change, ie the records wouldn't be deleted.

To workaround that one have to do something like adding a hidden field with the name of the association property, and check that before marshalling and replace it with an empty array accordingly, like for example

``` php
$this->Form->input('parent_field');
$this->Form->hidden('association', ['value' => '']);
$this->Form->input('association.0.field');
$this->Form->input('association.1.field');
$this->Form->input('association.2.field');
// ...
```

``` php
public function beforeMarshal(Event $event, \ArrayObject $data, \ArrayObject $options)
{
    if (isset($data['association']) && $data['association'] === '') {
        $data['association'] = [];
    }
}
```

That way all records would be deleted properly when none of the `association.*.field` inputs would be submitted.

I'm wondering if this is something that may fit in the core? Maybe as an optional behavior that can be enabled via the `associated` option?

markstory:
Making an empty string replace the association with an empty list sounds like a reasonable default behavior to me.

lorenzo:
Could this be used in a malicious way? Or could it be a surprise for users to discover that there is a way to delete all associated records from a form?

markstory:
@lorenzo Good point. It could be used for bad purposes.

ndm2:
I was a little worried too at first, hence the suggestion to make this optional, but the more I think about it, the less concerned I am. By default users are currently able to delete all but one record, so I'd say that one record more doesn't really make a difference.

I'd think that if developers use the replace save strategy, then they usually want exactly that to be possible. If they don't want users to be able to remove associated records via a form depending on the current context, then they would have to implement additional measures anyways, like switching to the append strategy, or better yet the other way around, enable the replace strategy only where explicitly required.

[...]

PR #10479, opened 2017-04-04
3.next - Add loose emptiness handling for `HasMany` associations.

ndm2:
Basically copied the functionality and tests from `BelongsToMany` with slight modifications and additions. Also refactored the `saveAssociated` method a little bit, making it easier to test and to read.

I wasn't sure about the rules regarding backwards compatibility for protected methods, if breaking changes were allowed, I would have changed `_unlinkAssociated()` to drop the `$entity` (unused) and `$target` arguments.

Refs #9474

### DIFF
--- a/tests/TestCase/ORM/Association/HasManyTest.php
+++ b/tests/TestCase/ORM/Association/HasManyTest.php
@@ -751,27 +751,22 @@ class HasManyTest extends TestCase
     public function testSaveAssociatedNotEmptyNotIterable()
     {
         $articles = TableRegistry::get('Articles');
-        $assoc = $articles->hasMany('Comments', [
+        $association = $articles->hasMany('Comments', [
             'saveStrategy' => HasMany::SAVE_APPEND
         ]);
 
-        $entity = new Entity(
-            [
-                'id' => 1,
-                'comments' => 'oh noes',
-            ],
-            ['markNew' => true]
-        );
+        $entity = $articles->newEntity();
+        $entity->set('comments', 'oh noes');
 
-        $assoc->saveAssociated($entity);
+        $association->saveAssociated($entity);
     }
 
     /**
-     * Provider for empty values
+     * Data provider for empty values.
      *
      * @return array
      */
-    public function emptyProvider()
+    public function emptySetDataProvider()
     {
         return [
             [''],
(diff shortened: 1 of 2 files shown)


tests/TestCase/ORM/Association/HasManyTest.php
```diff
@@ -740,4 +740,216 @@ public function testLinkUsesSingleTransaction()
...
+            ->setConstructorArgs(['comments', ['sourceTable' => $table]])
+            ->getMock();
+        $assoc->setSaveStrategy(HasMany::SAVE_APPEND);
+
+        $assoc
+            ->expects($this->never())
+            ->method('_unlinkAssociated');
+        $assoc
+            ->expects($this->never())
+            ->method('_saveTarget');
```
markstory: Instead of mocks integrating with the db and checking the db state afterwards has been resulting in easier to read and easier to maintain tests lately.

ndm2: @markstory You mean in this horrendously large beast called the "table test", also know as the unholy slayer of sanity? :wink:

Jokes aside, I could surely switch those to integration tests if that is preferred / considered sufficient. Without mocking, I imagine it could be hard to differentiate between bailing out early on empty sets, and having nothing to unlink/persist.

lorenzo: You could maybe do that by asserting that the set you are passing is not empty

ndm2: Maybe I misunderstood what you're suggesting, but I actually need to test this functionality with empty sets :) I probably overcomplicate things in my head though, I'll give it a try tomorrow and see if it's actually an issue at all.

markstory: @ndm Users of the ORM care about the persistence results not which protected methods are called though. I think the 'integration' tests could continue to exist in this file. I recently change a bunch of association class tests to not use mocks and it greatly simplified the test code.

ndm2: @markstory Of course, my concern was just that in a sitation where the method can return one and the same value for different internal actions, it wouldn't be possible to determine that the result, even though being the expected one (an empty array, for an empty array as input), is the result of internals working as expected or not.

But that's kind of a personal habit, I tend to over-test things, and giving it a try it seems that my concern was unsubstantiated, looks like it's enough to test the possibly existing associated records, and that the input property value equals the output property value :)

I've updated the tests.

markstory (commented):
Looks good to me.

Issue #10445, opened 2017-03-24
CakeSocket cannot connect to TLS 1.2 only servers


RichieB2B:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 2.9.6 83e5ae522e792f8994637938de29bb553384e9c8
* Platform and Target: CentOS 6, rh-php56 (5.6.25)

### What you did
```
$HttpSocket = new HttpSocket($params);
try {
  $response = $HttpSocket->get('https://www.ssllabs.com:10303/');
} catch (Exception $e) {
  print $e->getMessage();  
}
```

### What happened
The `HttpSocket->get()` throws:
```
Unable to perform enableCrypto operation on CakeSocket
```

### What you expected to happen
The `HttpSocket->get()` should complete without problems.

This issue is caused by https://bugs.php.net/bug.php?id=73388 and exists in PHP >= 5.6.7.
The easy fix is to change the definition of [tls_client](https://github.com/cakephp/cakephp/blob/2.x/lib/Cake/Network/CakeSocket.php#L95) to:
```
'tls_client' => STREAM_CRYPTO_METHOD_TLSv1_0_CLIENT | STREAM_CRYPTO_METHOD_TLSv1_1_CLIENT | STREAM_CRYPTO_METHOD_TLSv1_2_CLIENT
```
but this will only work for PHP >= 5.6.0. If support for PHP < 5.6.0 is still important a more elaborate patch is needed.

This issue probably also exists for [tls_server](https://github.com/cakephp/cakephp/blob/2.x/lib/Cake/Network/CakeSocket.php#L99).

markstory:
We need to continue supporting PHP 5.3+ in 2.x. Looks like we'll need to coalesce the various constants in the constructor.

RichieB2B:
PHP < 5.6.0 does not have the individual `STREAM_CRYPTO_METHOD_TLSv1_*_CLIENT` constants. So using `STREAM_CRYPTO_METHOD_TLS_CLIENT` is the best way for those older versions.

markstory:
Closing as a pull request is open now.

PR #10232, opened 2017-02-15
2.x Extend available TLS encrypt methods in CakeSocket

mstramek-CKPL:
Proposed change extend list of available encryption methods in CakeSocket with two new ones: TLSv1_1 and TLSv1_2. Moreover it added the possibility to choose custom TLS version by config option.

PR #10482, opened 2017-04-05
Fix CakeSocket not being able to connect to TLS1.2 only servers

markstory:
Include the changes from #10232 and also fix the issue described in #10445

### DIFF
--- a/lib/Cake/Network/CakeSocket.php
+++ b/lib/Cake/Network/CakeSocket.php
@@ -44,7 +44,8 @@ class CakeSocket {
 		'host' => 'localhost',
 		'protocol' => 'tcp',
 		'port' => 80,
-		'timeout' => 30
+		'timeout' => 30,
+        'cryptoType' => 'tls'
 	);
 
 /**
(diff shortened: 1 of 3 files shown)


lib/Cake/Network/CakeSocket.php
```diff
@@ -116,6 +117,32 @@ class CakeSocket {
  */
 	public function __construct($config = array()) {
 		$this->config = array_merge($this->_baseConfig, $config);
+
+		// These TLS versions are not supported by older PHP versions,
+		// so we have to conditionally set them if they are supported.
+		$conditionalCrypto = array(
+			'tlsv1_1_client' => 'STREAM_CRYPTO_METHOD_TLSv1_1_CLIENT',
```
ravage84: If we set these TLS versions depending on the PHP version, shouldn't we test the list of them conditionally?

lib/Cake/Network/CakeSocket.php
```diff
@@ -116,6 +117,32 @@ class CakeSocket {
  */
 	public function __construct($config = array()) {
 		$this->config = array_merge($this->_baseConfig, $config);
+
+		// These TLS versions are not supported by older PHP versions,
```
ravage84: The whole TLS related code could be moved into a protected method in which the doc blcok could epxlain the situation.

 + smaller method
 + related stuff is extracted and contained
 + in line comment moves to class API  docs

lib/Cake/Network/CakeSocket.php
```diff
@@ -44,7 +44,8 @@ class CakeSocket {
 		'host' => 'localhost',
 		'protocol' => 'tcp',
 		'port' => 80,
-		'timeout' => 30
+		'timeout' => 30,
+		'cryptoType' => 'tls'
```
ravage84: Nitpicking: trailing comma.

ravage84 (commented):
Looks good.

lib/Cake/Network/CakeSocket.php
RichieB2B: This is done by the `if (defined($const))` on line 130. If PHP doesn't know about the constant, it won't get added.

lib/Cake/Network/CakeSocket.php
markstory: good idea

lib/Cake/Network/CakeSocket.php
espaistelematics: Perfect! Works good.

Issue #10486, opened 2017-04-05
CSRF blackhole on Safari


bancer:
This is a (multiple allowed):

* [x] bug
* [x] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 2.9.6.
* Platform and Target: Apache, MySQL.

### What you did
1. Created a form that uses Security component.
2. Added `$this->response->disableCache();` in the controller.
3. Opened the page with that form in Safari.
4. Submitted the form.
5. Clicked browser "Back" icon.
6. Submitted the form again with different data - blackhole because hidden _Token input had the same value as the first time.

### What happened
It seems Safari sometimes ignores headers set by `$this->response->disableCache();` and uses the cache instead of loading the new page after "Back" icon is clicked. 

### What you expected to happen
The page to be reloaded without using cache. Adding `autocomplete="off"` attribute to the hidden token input fixes that.
```
protected function _csrfField() {
	if (empty($this->request->params['_Token'])) {
		return '';
	}
	if (!empty($this->request['_Token']['unlockedFields'])) {
		foreach ((array)$this->request['_Token']['unlockedFields'] as $unlocked) {
			$this->_unlockedFields[] = $unlocked;
		}
	}
	return $this->hidden('_Token.key', array(
		'value' => $this->request->params['_Token']['key'], 'id' => 'Token' . mt_rand(),
		'secure' => static::SECURE_SKIP,
		'autocomplete' => 'off',
	));
}
```

markstory:
Closing as there is a pull request for 3.x, and 2.x was fixed in 5685c031e21756b791202651f1323902d1f324fd.

PR #10487, opened 2017-04-05
3.x - Add autocomplete attribute to CSRF/security token fields.

markstory:
New versions of Safari will overwrite these fields when a user uses the back button. If one-time CSRF tokens are in use the request will be blackholed.

Refs #10486


markstory:
I'm a toolshed and pushed the fixes for 2.x directly to 2.x :cry: See 5685c031e21756b791202651f1323902d1f324fd for the 2.x fixes.

dereuromark:
Wow, thats annoying..

markstory:
Yep..

PR #10488, opened 2017-04-05
Adding controller mapping in resources

chrisShick:
This feature allows one to be able to define a resource that may not have the same name as the controller. It is the same concept as the connect method

      $routes->connect('/Gallery', ['controller' => 'Albums', 'action' => 'add']);

### DIFF
--- a/tests/TestCase/Routing/RouteBuilderTest.php
+++ b/tests/TestCase/Routing/RouteBuilderTest.php
@@ -409,6 +409,19 @@ class RouteBuilderTest extends TestCase
     }
 
     /**
+     * Test connecting resources with a controller
+     *
+     * @return void
+     */
+    public function testResourcesController()
+    {
+        $routes = new RouteBuilder($this->collection, '/api');
+        $routes->resources('Articles', ['controller' => 'Posts']);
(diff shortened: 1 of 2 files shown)


tests/TestCase/Routing/RouteBuilderTest.php
```diff
@@ -409,6 +409,19 @@ public function testResources()
...
+     * Test connecting resources with a controller
+     *
+     * @return void
+     */
+    public function testResourcesController()
+    {
+        $routes = new RouteBuilder($this->collection, '/api');
+        $routes->resources('Articles', ['controller' => 'Posts']);
+        $all = $this->collection->routes();
+        $this->assertEquals('Posts', $all[0]->defaults['controller']);
```
markstory: Should you check the other aspects of the routes, like the action and path?

chrisShick: @markstory Doesn't the testConnectBasic do that for us already?

chrisShick: Also, the testResourcesPrefix function only tests the prefix as well.

PR #10493, opened 2017-04-06
3.next - Integrate CookieCollection into Response

markstory:
Integrate the new Cookies and CookieCollection into the Response. All of the existing cookie related methods are backwards compatible, and `withCookie()` allows new style cookies as a parameter.

I've had to add more compatibility shims into Cookie and relax the expires parameter typing to allow integers. I found this preferable to coercing the existing cookie data into a DateTime just so it can be converted back into an integer.

Refs #10406

### DIFF
--- a/src/Http/Cookie/Cookie.php
+++ b/src/Http/Cookie/Cookie.php
@@ -15,8 +15,8 @@ namespace Cake\Http\Cookie;
 
 use Cake\Chronos\Chronos;
 use Cake\Utility\Hash;
-use DateTimeInterface;
 use DateTimeImmutable;
+use DateTimeInterface;
 use DateTimezone;
 use InvalidArgumentException;
 use RuntimeException;
--- a/src/Http/Cookie/Cookie.php
+++ b/src/Http/Cookie/Cookie.php
@@ -121,7 +121,7 @@ class Cookie implements CookieInterface
      * @link http://php.net/manual/en/function.setcookie.php
      * @param string $name Cookie name
      * @param string|array $value Value of the cookie
-     * @param \DateTimeInterface|null $expiresAt Expiration time and date
+     * @param \DateTimeInterface|int|null $expiresAt Expiration time and date
      * @param string $path Path
      * @param string $domain Domain
      * @param bool $secure Is secure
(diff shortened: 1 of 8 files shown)


src/Http/Response.php
```diff
@@ -462,6 +461,7 @@ public function __construct(array $options = [])
             $this->_contentType = $this->resolveType($options['type']);
         }
         $this->_setContentType();
+        $this->_cookies = new CookieCollection();
```
robertpustulka: Can we inject a collection from `$options`?

src/Http/Response.php
```diff
@@ -1977,30 +1985,44 @@ public function cookie($options = null)
...
-            'domain' => '',
-            'secure' => false,
-            'httpOnly' => false
-        ];
-        $data += $defaults;
-        $data['name'] = $name;
 
         $new = clone $this;
-        $new->_cookies[$name] = $data;
+        $new->_cookies = $new->_cookies->add($cookie);
```
robertpustulka: I think a collection should be cloned as well. Adding a cookie to the new request's collection would also mutate current request's collection.

src/Http/Response.php
markstory: Not right now. Responses don't take cookies in their constructor at this time.

src/Http/Cookie/Cookie.php
```diff
@@ -251,7 +271,7 @@ public function getName()
      */
     protected function validateName($name)
     {
-        if (preg_match("/[=,; \t\r\n\013\014]/", $name)) {
+        if (preg_match("/[=,;\t\r\n\013\014]/", $name)) {
```
ravage84: @markstory do you know of any documented list of invalid characters? If so, we could link to it with a ``@link``.

markstory: The cookie RFCs have the accepted character lists.

ravage84: If I interpreted that correctly:

> cookie-name       = token
> token             = <token, defined in [RFC2616], Section 2.2>

https://tools.ietf.org/html/rfc6265#section-4.1.1

So we could add:
``@link https://tools.ietf.org/html/rfc2616#section-2.2 Rules for naming cookies.``

markstory: done.

Issue #10495, opened 2017-04-06
Fatal error: Trait method getProtocolVersion has not been applied, because there are collisions with other trait methods


mergeweb:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

CakePHP 3.4.1 >
ADDING "zendframework/zend-diactoros": "1.3.10" fixes the issue.

### What you did
Updated Composer

### What happened
After composer update it appears 
https://gist.github.com/mergeweb/44257e8509e6ee67092b142ba768804e
ADDING "zendframework/zend-diactoros": "1.3.10" fixes the issue.

### What you expected to happen
Not a big trace error

mergeweb:
Version 3.4.1
and up.

ADmad:
Problem seems to be cause by some change in diactoros' 1.4.0 release.

markstory:
Looks like zendframework/diactoros 1.4.0 introduced this issue. I will get in contact with those folks tonight.

juniorionut:
This was fast :)

markstory:
Closing as new releases are out.

PR #10496, opened 2017-04-06
Don't doubly include MessageTrait.

markstory:
RequestTrait already includes MessageTrait. Also pin to newer diactoros as there are problems with older versions.

People using CakePHP 3.3.x will need to require diactoros 1.3.10 in their application, as there aren't maintenance releases for 3.3 anymore.

Refs #10495

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -24,7 +24,7 @@
         "cakephp/chronos": "~1.0",
         "aura/intl": "^3.0.0",
         "psr/log": "^1.0",
-        "zendframework/zend-diactoros": "~1.0"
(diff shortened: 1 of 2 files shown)


dereuromark:
@markstory 
We could create a patch release from [3.3.15](https://github.com/cakephp/cakephp/releases/tag/3.3.15) as 3.3.16 with the pinned constraint.
I bet folk with 3.3 still would appreciate that.

markstory:
@dereuromark I can apply this same diff to 3.3

PR #10497, opened 2017-04-06
Make test suite compatible with PHPUnit 5.

### LINKED ISSUE #9973: 2.x PHPunit 3.7.38 doesn't play nicely with PHP 7
This is a (multiple allowed):

* [ ] bug
* [x] enhancement

* CakePHP Version: 2.9.4
* Platform and Target: Centos 7 PHP 7.1.0

### What you did
```php
class MyModel extends Model {
    public function doSomething(array $data): array {
        // ....
    }
}
```

### What happened
Mocking the model with `$this->getMockForModel('MyModel');` results in the following error:

```
Declaration of Mock_MyModel_bea013db::doSomething(array $data = Array) must be compatible with MyModel::doSomething(array $data = Array): array	
```

Not just `getMockForModel` of course, all mocking methods do this.

Model is successfully mocked.

Are there any plans to update PHPunit?

mtricht:
Sadly there are 1200 deprecation warnings for `getMock`, but at least we can use PHP7 features now!

Closes #9973

### DIFF
--- a/lib/Cake/TestSuite/CakeTestCase.php
+++ b/lib/Cake/TestSuite/CakeTestCase.php
@@ -655,7 +655,7 @@ abstract class CakeTestCase extends PHPUnit_Framework_TestCase {
  * @deprecated 3.0.0 This is a compatibility wrapper for 1.x. It will be removed in 3.0.
  * @return void
  */
-	protected function expectException($name = 'Exception', $message = '') {
+	public function expectException($name = 'Exception', $message = '') {
 		$this->setExpectedException($name, $message);
 	}
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,6 +32,9 @@ matrix:
     - php: 7.0
       env: PHPCS=1
 
+    - php: 7.1
+      env: DB=mysql PHPUNIT=5.7.19
+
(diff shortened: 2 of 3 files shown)


markstory:
Do you know if PHPUnit runs the test method when there is a warning? I thought it bailed as soon as the warning condition was hit.

lib/Cake/TestSuite/CakeTestCase.php
```diff
@@ -655,7 +655,7 @@ protected function expectError($expected = false, $message = '') {
  * @deprecated 3.0.0 This is a compatibility wrapper for 1.x. It will be removed in 3.0.
```
inoas: OT: ... removed in 3.0?

swordbeta:
@markstory It executes the test yes.

@inoas looks like it has been removed https://github.com/cakephp/cakephp/blob/3.next/src/TestSuite/TestCase.php

josegonzalez:
This is probably fine for merging, right @markstory ?

markstory:
Sure. I have no time to deal with PHPUnit deprecations in 2.x, perhaps someone else can take it on?

Issue #9978, opened 2017-01-05
3.5: Replace multi-mode method with separate get set methods.


dereuromark:
Before stabilizing the API for 3.4, the following methods are not yet split and still are `$this|...` combinations that need to be separate methods for setting and getting:

Prio:
- [x] ORM Query::contain()
- [x] ORM Query::hydrate()
- [x] ORM Query::autoFields()
- [x] Database Query::valueBinder()

Nice to have:
- [x] QueryTrait::mapReduce()
- [x] QueryTrait::formatResults()
- [ ] TranslateTrait::translation()
- [x] Message::body()
- [x] ValidatorAwareTrait::validator()
- [x] ModelAwareTrait::modelType()
- [x] EntityTrait::hiddenProperties()
- [x] EntityTrait::virtualProperties()
- [x] EntityTrait::errors()
- [x] EntityTrait::source()
- [x] EntityTrait::accessible()
- [x] QueryTrait::eagerLoaded()
- [x] ServerRequest::env()
- [x] LocatorAwareTrait::tableLocator()
- [x] TableRegistry::locator()
- [x] EventDispatcherTrait::eventManager()

Hard to do:
- [x] QueryInterface::repository() / QueryTrait::repository()
- [x] InvalidPropertyInterface::invalid() / EntityTrait::invalid()
- [x] TypedResultInterface::returnType() / TypedResultTrait::returnType()

Already PR opened ( https://github.com/cakephp/cakephp/pull/9769 ):
- InstanceConfigTrait
- StaticConfigTrait

Dangerous $this|null (!) when chaining:
- [x] Query::traverseExpressions()

dereuromark:
- Query::contain() could stay as setter, and Query::getContained() could get?
- Query::valueBinder() into get and set? Or is there something behind the internal ValueBinder class?

inoas:
I'd leave the magic setters/getters (until deprecated). As long as the old interface is there, it will work as expected/learned. So there should be `setContain()`?

chinpei215:
@dereuromark I think you can also add `Database Query::from()` to the list. https://github.com/cakephp/cakephp/pull/10033#discussion_r96113162

cleptric:
What about `EntityTrait ::dirty`?

dereuromark:
Feel free to continue the list of things I might have missed as checklist here to cross of during the next 3.x releases.

burzum:
I've added dirty and made more changes to the code. Need to add some tests to finish this.

[...]

PR #10500, opened 2017-04-10
Split QueryTrait::eagerLoaded() into getter/setter

cleptric:
Continues with #9978

### DIFF
--- a/src/Datasource/QueryTrait.php
+++ b/src/Datasource/QueryTrait.php
@@ -185,7 +185,7 @@ trait QueryTrait
      *
      * @return bool
      */
-    public function getEagerLoaded()
+    public function isEagerLoaded()
     {
         return $this->_eagerLoaded;
     }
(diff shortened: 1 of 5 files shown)


markstory (approved):
Looks good to me.

src/Datasource/QueryTrait.php
```diff
@@ -181,9 +181,33 @@ public function cache($key, $config = 'default')
...
+    public function getEagerLoaded()
+    {
+        return $this->_eagerLoaded;
+    }
+
+    /**
+     * Sets the query instance to be an eager loaded query.
+     *
+     * @param bool $value Whether or not to eager load.
+     * @return $this|\Cake\ORM\Query
```
markstory: Shouldn't this be just `$this`?

cleptric: I copied it from the old method 😇
I'll remove it.

src/Datasource/QueryTrait.php
```diff
@@ -181,9 +181,33 @@ public function cache($key, $config = 'default')
...
+     */
+    public function setEagerLoaded($value)
```
dereuromark: Dont wenn usually use enable.. and isEnabled.. instead for bool values?

cleptric:
I removed the `setEagerLoaded()` method as we have discussed this topic here #10457

PR #10501, opened 2017-04-10
Split ServerRequest::env() into getter/setter

cleptric:
Continues #9978

### DIFF
--- a/src/Http/ServerRequest.php
+++ b/src/Http/ServerRequest.php
@@ -351,7 +351,7 @@ class ServerRequest implements ArrayAccess, ServerRequestInterface
      */
     protected function _processPost($data)
     {
-        $method = $this->env('REQUEST_METHOD');
+        $method = $this->getEnv('REQUEST_METHOD');
         $override = false;
 
         if (in_array($method, ['PUT', 'DELETE', 'PATCH']) &&
@@ -506,12 +506,12 @@ class ServerRequest implements ArrayAccess, ServerRequestInterface
      */
     public function contentType()
     {
-        $type = $this->env('CONTENT_TYPE');
+        $type = $this->getEnv('CONTENT_TYPE');
         if ($type) {
             return $type;
         }
 
-        return $this->env('HTTP_CONTENT_TYPE');
+        return $this->getEnv('HTTP_CONTENT_TYPE');
     }
 
     /**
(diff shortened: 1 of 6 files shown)


src/Http/ServerRequest.php
```diff
@@ -1743,9 +1743,44 @@ public function withProtocolVersion($version)
...
+    }
+
+    /**
+     * Set a value to the request's environment data.
+     *
+     * @param string $key The key you want to write to.
+     * @param string $value Value to set
+     * @return $this
+     */
+    public function setEnv($key, $value)
```
markstory: Given that requests are trending towards immutable patterns, shouldn't any new methods follow that style?

tests/TestCase/Http/ServerRequestTest.php
```diff
@@ -3520,20 +3520,30 @@ public function testWithRequestTarget()
...
+     * @return void
+     */
+    public function testGetEnv()
+    {
+        $request = new ServerRequest();
 
         //Test default null
-        $request->setEnv('HTTP_HOST');
-        $this->assertSame($request->getEnv('HTTP_HOST'), null);
+        $this->assertSame($request->getEnv('Foo'), null);
```
markstory: You should use assertNull().

tests/TestCase/Http/ServerRequestTest.php
```diff
@@ -3543,7 +3543,7 @@ public function testGetEnv()
         $request = new ServerRequest();
 
         //Test default null
-        $this->assertSame($request->getEnv('Foo'), null);
+        $this->assertNull($request->getEnv('Foo'), null);
```
dereuromark: The 2nd argument should then be removed.

tests/TestCase/Http/ServerRequestTest.php
```diff
@@ -3543,7 +3543,7 @@ public function testGetEnv()
...
 
         //Test default set
         $this->assertSame($request->getEnv('Foo', 'Bar'), 'Bar');
```
dereuromark: Those are still using the wrong argument order. "expected, result" is what it should be.

tests/TestCase/Http/ServerRequestTest.php
```diff
@@ -3520,6 +3520,40 @@ public function testWithRequestTarget()
...
+     *
+     * @return void
+     */
+    public function testWithEnv()
+    {
+        $request = new ServerRequest();
+
+        $newRequest = $request->withEnv('HTTP_HOST', 'cakephp.org');
+        $this->assertNotSame($request, $newRequest);
+        $this->assertSame($newRequest->getEnv('HTTP_HOST'), 'cakephp.org');
```
dereuromark: wrong order.

PR #10502, opened 2017-04-10
Split Security::salt() into getter/setter

cleptric:
Continues with #9978


cleptric:
Should `setSalt()` still return the salt?

markstory:
I don't think setSalt() should return the value. Other setter methods return $this or void in general.

PR #10503, opened 2017-04-10
Split ConsoleOutput::outputAs() into getter/setter

cleptric:
Continues with #9978

### DIFF
--- a/src/Console/ConsoleIo.php
+++ b/src/Console/ConsoleIo.php
@@ -266,7 +266,7 @@ class ConsoleIo
      */
     public function outputAs($mode)
     {
-        $this->_out->outputAs($mode);
+        $this->_out->setOutputAs($mode);
     }
 
     /**
--- a/src/Console/ConsoleIo.php
+++ b/src/Console/ConsoleIo.php
@@ -264,6 +264,19 @@ class ConsoleIo
      * @return void
      * @see \Cake\Console\ConsoleOutput::outputAs()
      */
+    public function setOutputAs($mode)
+    {
+        $this->_out->setOutputAs($mode);
+    }
+
+    /**
+     * Change the output mode of the stdout stream
+     *
+     * @deprecated 3.5.0 Use setOutputAs() instead.
+     * @param int $mode The output mode.
+     * @return void
+     * @see \Cake\Console\ConsoleOutput::outputAs()
+     */
     public function outputAs($mode)
     {
         $this->_out->setOutputAs($mode);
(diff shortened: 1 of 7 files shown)


src/Console/ConsoleIo.php
```diff
@@ -266,7 +266,7 @@ public function ask($prompt, $default = null)
      */
     public function outputAs($mode)
```
markstory: Shouldn't this method be marked as deprecated?

cleptric: Yes, we should deprecate this one too.

cleptric: @markstory Just to clearify this. I deprecated `ConsoleOutput::outputAs`. Should I also deprecate `ConsoleIo::outputAs `? `ConsoleIo::outputAs` currently acts as a setter only.

markstory: Oh I misread what ConsoleIo::outputAs() did. It might be nice if the methods had the same name as there would be fewer names to remember later on.

src/Console/ConsoleOutput.php
```diff
@@ -303,8 +303,30 @@ public function styles($style = null, $definition = null)
...
+
+    /**
+     * Set the output type on how formatting tags are treated.
+     *
+     * @param int $type The output type to use. Should be one of the class constants.
+     * @return void
+     */
+    public function setOutputAs($type)
+    {
+        $this->_outputAs = $type;
```
markstory: Should we validate that the parameter is a valid output type?

src/Console/ConsoleOutput.php
```diff
@@ -317,9 +319,14 @@ public function getOutputAs()
      *
      * @param int $type The output type to use. Should be one of the class constants.
      * @return void
+     * @throws InvalidArgumentException in case of a not supported output type.
```
dereuromark: Leading \ missing for FQCN

src/Console/ConsoleOutput.php
```diff
@@ -317,9 +319,14 @@ public function getOutputAs()
...
      */
     public function setOutputAs($type)
     {
+        if (in_array($type, [self::RAW, self::PLAIN, self::COLOR], true) === false) {
```
dereuromark: I always find !in_array() more readable, no need for the strictness of internal php function checks, right?

cleptric:
@markstory Any idea why the `outputAs` method can't be mocked any more?

markstory:
No idea.

PR #10328, opened 2017-03-01
New cookie implementation for the HTTP stack

burzum:
## New prototype implementation for cookies

It needs more tests and I'm not yet happy with the handling of the cookie data and encryption. But I would already like to get feedback on this to continue to develop it into a direction we all agree with.

The code is partially based on https://github.com/hansott/psr7-cookies/ but was changed to fit into our current API and the frameworks current php version. I've had a discussion with Mark Story if we'll directly use it or not. The final decision for the named reasons was to come up with our own implementation. I've also changed the implementation in some aspects, it's not a copy and paste job. I need some advice if we need to give credits and copyright and if yes, how to do it the right way.

Cookies can now be used from wherever the request object and response objects are available.

## Overview of the new implementation

* Uses the request and response objects
* Uses a cookie object to built and modify a cookie
* Generates cookie objects from the request object
* Writes cookies to the response object
* Uses getters and setters
* No longer a component bound controller layer only solution (This change is required to get the cookie authenticator in the new authentication implementation working)
* **TBD**: If possible a BC compatible cookie component that uses the new implementation

## Related issues

* https://github.com/cakephp/authentication/issues/26
* https://github.com/cakephp/cakephp/pull/10203

## History

Former PR from my fork https://github.com/cakephp/cakephp/pull/10203

Created a new PR based on the cake repo and not my fork so other people can contribute to the branch as well without being a member of my fork.

PR #10505, opened 2017-04-11
3.next - Add Cookie->isExpired()

markstory:
Make it easier and less repetitive to check if a cookie is expired or not.

Refs #10406

### DIFF
--- a/src/Http/Cookie/Cookie.php
+++ b/src/Http/Cookie/Cookie.php
@@ -175,23 +175,6 @@ class Cookie implements CookieInterface
     }
 
     /**
-     * Get the timestamp from the expiration time
-     *
-     * Timestamps are strings as large timestamps can overflow MAX_INT
-     * in 32bit systems.
-     *
-     * @return string|null The expiry time as a string timestamp.
-     */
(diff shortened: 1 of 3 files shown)


robertpustulka:
Shouldn't this method go to the interface as well?

dereuromark:
For BC reasons this can usually only happen with the next major.

robertpustulka:
@dereuromark new cookies implementation targets 3.5.0 which hasn't been released yet, am I correct?

dereuromark:
If the whole class is new, then you are right :)

robertpustulka:
@dereuromark it's been introduced here #10328 so we can still change the interface, right?

robertpustulka:
I believe this method has to be added to the interface because of this line: [CookieCollection:106](https://github.com/cakephp/cakephp/blob/a4b663daac64dd8ba78185d234c42de4eb7b8bd2/src/Http/Cookie/CookieCollection.php#L106)

We allow `CookieInterface` to be added to the `$cookies` property. If someone added their own implementation, the `isExpired()` check would fail.

PR #10506, opened 2017-04-11
3.next - Rename controller option to path.

markstory:
After merging the changes from #10488 I started writing the documentation and noticed that most of the other RouteBuilder methods use 'path' and not 'controller' as an option. Given that changing the path was the original intent, I thought that might be a more consistent option name. Furthermore, the previous changes would result in awkward reverse routing scenarios where the routing parameter would not match the controller/model name, and instead match the resource name. e.g. `resources('BlogPosts', ['controller' => 'articles'])` would use `blog_post_id` as the routing parameter instead of `article_id` which does not match the conventions used elsewhere in the framework.

cc @chrisShick

### DIFF
--- a/tests/TestCase/Routing/RouteBuilderTest.php
+++ b/tests/TestCase/Routing/RouteBuilderTest.php
@@ -409,16 +409,25 @@ class RouteBuilderTest extends TestCase
     }
 
     /**
-     * Test connecting resources with a controller
+     * Test connecting resources with a path
      *
      * @return void
      */
-    public function testResourcesController()
+    public function testResourcesPathOption()
     {
         $routes = new RouteBuilder($this->collection, '/api');
-        $routes->resources('Articles', ['controller' => 'Posts']);
+        $routes->resources('Articles', ['path' => 'posts'], function ($routes) {
+            $routes->resources('Comments');
+        });
         $all = $this->collection->routes();
-        $this->assertEquals('Posts', $all[0]->defaults['controller']);
+        $this->assertEquals('Articles', $all[0]->defaults['controller']);
+        $this->assertEquals('/api/posts', $all[0]->template);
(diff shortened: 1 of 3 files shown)


ADmad (approved):

tests/TestCase/Routing/RouteBuilderTest.php
```diff
@@ -409,16 +409,25 @@ public function testResources()
...
+     * Test connecting resources with a path
      *
      * @return void
      */
-    public function testResourcesController()
+    public function testResourcesPathOption()
     {
         $routes = new RouteBuilder($this->collection, '/api');
-        $routes->resources('Articles', ['controller' => 'Posts']);
+        $routes->resources('Articles', ['path' => 'posts'], function ($routes) {
```
saeideng: Needs a  test for "Posts" !?

markstory: Doesn't the assertion on the generated path cover this?

saeideng: I don't know :) I did not test it

chrisShick:
Ouch! Was this happening before with the connect option? Or was this just not as simple of a change as I thought it was?

markstory:
@chrisShick The parameter inflection issue was part of the controller option add. I'm not sure what you're referring to with 'the connect option'

chrisShick:
I meant does this happen in the connect method of the RouteBuilder as well? Or was this issue of me adding the controller option to the resource method incorrectly?

markstory:
@chrisShick It was a side effect of how the new option and existing code interacted. I should have caught this problem before you did any work, so don't feel bad.

chrisShick:
Ahh!! I will look at the new code you did and learn from it, brother.

chrisShick (approved):

PR #10508, opened 2017-04-11
Fix version cakephp 3.4

mehdi-fathi:
change version cakephp in REDAME.md.


ADmad:
We should just remove the version from install command.

PR #10509, opened 2017-04-11
Split ServerRequest::session() into getter/setter

cleptric:
Continues with #9978

### DIFF
--- a/src/Http/ServerRequest.php
+++ b/src/Http/ServerRequest.php
@@ -515,23 +515,6 @@ class ServerRequest implements ArrayAccess, ServerRequestInterface
     }
 
     /**
-     * Update the request with a new session instance.
-     *
-     * Returns an updated request object. This method returns
-     * a *new* request object and does not mutate the request in-place.
-     *
-     * @param \Cake\Network\Session $session the session object to use
-     * @return static
-     */
-    public function withSession(Session $session)
-    {
-        $new = clone $this;
-        $new->session = $session;
-
-        return $new;
-    }
-
-    /**
      * Returns the instance of the Session object for this request
      *
      * @return \Cake\Network\Session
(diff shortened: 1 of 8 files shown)


src/Http/ServerRequest.php
```diff
@@ -515,11 +515,39 @@ public function contentType()
...
+     * Returns an updated request object. This method returns
+     * a *new* request object and does not mutate the request in-place.
+     *
+     * @param \Cake\Network\Session $session the session object to use
+     * @return static
+     */
+    public function withSession(Session $session)
+    {
+        $new = clone $this;
+        $new->session = $session;
```
markstory: Sessions are funny because they are effectively shared global state. I don't think we can or should try to make promises around sessions and request immutability.

tests/TestCase/Http/ServerRequestTest.php
```diff
@@ -3179,6 +3179,35 @@ public function testSession()
...
+        $request = ServerRequestFactory::fromGlobals();
+        $this->assertEquals($session, $request->getSession());
+    }
+
+    /**
+     * Tests setting the session to the request
+     *
+     * @return void
+     */
+    public function testWithtSession()
```
markstory: Typo in the method name.

ADmad:
I don't think the session () method should be changed. As @markstory mentioned having a withSession() is not practical. So since we only have a getter version keeping the method name as session () is good enough.

cleptric:
@ADmad `session()` is currently a mixed setter/getter. I understand the problem with the immutable setter but shouldn't we than rename `withSession()` to `setSession()` and make it behave mutable?

dereuromark:
We have to find a global pattern as concept here. So the API is consistent here across the framework.

We could keep `session()` as the getter alias if that is more convenient, but the setter should definitely be `setSession()` only.

dereuromark:
On a side note:
For me those kind of combined setters are not that big of a deal, because they always return the same type - in this case for both get and set part `@return \Cake\Network\Session`.
This way those are less harmful than the `$this|...` mixed ones.

That's why prioritized them down, we could even keep the setter() one here as the combined one, as it does "only" violate the single responsibility, it doesnt too much interfere with future strictness of typehinting.

ADmad:
@cleptric Yes it's mixed but you almost never have to use `session()` as a setter yourself. The Session instance is already created and set during bootstrapping process. It would be more convenient to use `request->session()` as getter instead of the longer `request->getSession()`.

markstory (approved):
Looks good to me.

PR #10510, opened 2017-04-11
Split Shell::io() into getter/setter

cleptric:
Continues with #9978


dereuromark (approved):
Thanks for helping out here, looks good.

PR #10511, opened 2017-04-11
Split View::x() into getter/setter

cleptric:
Continues with #9978

### DIFF
--- a/src/View/View.php
+++ b/src/View/View.php
@@ -47,7 +47,7 @@ use RuntimeException;
  * ```
  * public function beforeRender(\Cake\Event\Event $event)
  * {
- *      $this->viewBuilder()->theme('SuperHot');
+ *      $this->viewBuilder()->setTheme('SuperHot');
  * }
  * ```
  *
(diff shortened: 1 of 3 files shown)


src/View/View.php
```diff
@@ -393,10 +437,34 @@ public function layoutPath($path = null)
...
+    /**
      * Turns on or off CakePHP's conventional mode of applying layout files.
      * On by default. Setting to off means that layouts will not be
      * automatically applied to rendered templates.
      *
+     * @param bool $autoLayout Boolean to turn on/off.
+     * @return void
+     */
+    public function setAutoLayout($autoLayout)
+    {
```
markstory: Should this be `enableAutoLayout()`? Some of the flag setters follow that convention.

markstory (commented):
Thank you for adding tests for the new methods.

src/View/View.php
cleptric: Yep. I also renamed the getter to `isAutoLayoutEnabled()` as it is done this way in the `ViewBuilder` class.

PR #10512, opened 2017-04-12
Added most commonly used cookie methods to the CookieInterface.

robertpustulka:
Since we allow `CookieInterface` in collections we must ensure methods used on cookies there are implemented.

I've added general cookie related methods to the interface, apart from the ones related to encryption and expanded data.

I'm not sure what to do with `toArray*()` methods. These are here for "historical" reasons. Still used in request/response but I'm not sure we should force users to implement them. Maybe do a `$cookie instanceof Cookie` check where these are used? Or as this is some legacy code - convert cookies to arrays internally in client code and remove these methods from cookie? I think the old code here should adapt to the new one, not otherwise.

Follow-up to the #10505

### DIFF
--- a/src/Http/Cookie/CookieCollection.php
+++ b/src/Http/Cookie/CookieCollection.php
@@ -36,7 +36,7 @@ class CookieCollection implements IteratorAggregate, Countable
     /**
      * Cookie objects
      *
-     * @var Cookie[]
+     * @var \Cake\Http\Cookie\CookieInterface[]
      */
     protected $cookies = [];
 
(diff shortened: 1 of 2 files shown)


markstory:
> convert cookies to arrays internally in client code and remove these methods from cookie? I think the old code here should adapt to the new one, not otherwise.

I put the compatibility methods on the cookie object as the various array formats need different date serializations and I didn't want to spread that logic around. I'm open to moving the logic around if you feel its necessary.

markstory:
Changes look good to me.

PR #10513, opened 2017-04-13
3.next - Add Cookie encryption middleware

markstory:
In order to deprecate the CookieComponent, we need to have an alternative solution. By having middleware that transparently encrypts/decrypts cookies we can deprecate the CookieComponent.

The previous use cases of using different encryption keys for different cookie names, can be afforded by applying the EncryptedCookieMiddleware multiple times.

### TODO

* [x] Complete API docs

### Questions

* Does the whitelist approach make sense?
* Should there be a `*` mode to encrypt all cookies?
* Are there other features of CookieComponent that are missing between the new Cookie objects and this middleware?

Refs #10406


burzum:
The whitelist approach is OK. The * is also OK, I think it's a very rare case somebody encrypts multiple cookies differently? 

Something missing... not sure :)

burzum (approved):

markstory:
Working on the docs for this now.

PR #10514, opened 2017-04-13
Update Plugin.php

Theaxiom:
Variable '$path' is introduced in an outer loop and overridden here.

This is a no-op, however we should aim for best practice.

### DIFF
--- a/src/Core/Plugin.php
+++ b/src/Core/Plugin.php
@@ -230,9 +230,9 @@ class Plugin
                 continue;
             }
             $dir = new DirectoryIterator($path);
-            foreach ($dir as $path) {
-                if ($path->isDir() && !$path->isDot()) {
-                    $plugins[] = $path->getBasename();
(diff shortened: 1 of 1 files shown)


dereuromark:
Yeah, it is not reused, thus no harm.
But I agree, that it would be cleaner (and more future proof in case someone starts using it later on) to use a different var. :+1:

ADmad:
@Theaxiom For future reference please write a proper description for  the change in the commit message, don't just submit with default message GH prefills for you.

Theaxiom:
@ADmad Check the second line of the commit message, is that not verbose enough? I apologize.

ADmad:
The description in second line is nice but the first line is what's commonly shown :)

Theaxiom:
Yeah sorry about that, I am used to using GitLab, GitHub is lacking in that department. I will remember that going forward, thanks for the heads up! :)

PR #10516, opened 2017-04-13
Update Plugin.php

Theaxiom:
Throws a non-annotated/unhandled exception: '\Cake\Core\Exception\MissingPluginException'.

### DIFF
--- a/src/Core/Plugin.php
+++ b/src/Core/Plugin.php
@@ -220,6 +220,7 @@ class Plugin
      *
      * @param array $options Options.
      * @return void
+     * @throws \Cake\Core\Exception\MissingPluginException
      */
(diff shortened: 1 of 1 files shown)


dereuromark:
Hmm, I am not a big fan of trying to manually annotate the exceptions into all the methods in the calling tree. As long as this can't be automated, there is too much human error, especially if those exceptions are removed again at some point and all the calling methods still have the annotation then.

I have a sniffer that only ever annotates the throwing method (automtically) and removes the ones from the calling methods that don't directly throw it, as this is at least somewhat consistent and automatable right now. I would recommend sth similar here.

Theaxiom:
I only suggested it because the other methods in that class are annotated.

PR #10517, opened 2017-04-13
2.x - Fix `ControllerTestCase::testAction()` incompatibility with `App.base`.

ndm2:
When using array URLs with `testAction()`, the generated URL possibly
contains the configured `App.base` path, which needs to be stripped when
set on the request object, as otherwise routes cannot be matched
correctly.

When passing the URL as an option to the `CakeRequest` constructor, the
it will be set as-is, unlike when the URL is being generated by
`CakeRequest::_url()`, which grabs the URL from the environment, and
strips the possible base path.

### DIFF
--- a/lib/Cake/Test/Case/TestSuite/ControllerTestCaseTest.php
+++ b/lib/Cake/Test/Case/TestSuite/ControllerTestCaseTest.php
@@ -624,4 +624,30 @@ class ControllerTestCaseTest extends CakeTestCase {
 		$this->assertEquals($restored, $_POST);
 	}
 
+/**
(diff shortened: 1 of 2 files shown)


ravage84:
@liviascapin could you test your app with this fix, please?

ravage84:
Relates to #4896

liviascapin:
Thanks again @ndm2!

I tested the fix it with my app which was suffering under this bug as described on stackoverflow: http://stackoverflow.com/questions/43340740/cakephp-running-in-subdirectory-creates-problems-with-urls-in-shell-web-and-uni

And it works fine now :)

markstory:
Thanks @ndm2 !

PR #10520, opened 2017-04-14
Move phpstan file in tests

dereuromark:
The root folder is already cluttered with noise files as it is.
You can't even see the badges anymore.

I would like to propose moving the stan file at least inside tests (where it belongs as static analysis part of test CI), to clean up a bit the root.
There is already a https://github.com/cakephp/cakephp/tree/master/tests/PHPStan folder inside anyway.
I did the same with my plugins.

It is nice that phpstan allows to customize/configure the path, I think we should leverage it here.

//cc @ondrejmirtes


markstory:
>  You can't even see the badges anymore.

I'm not sure keeping those visible should be a goal. On my small laptop screen, removing 1 line doesn't make the badges visible :smile:

dereuromark:
We still dont have to clutter it more than necessary.
The stan file inside tests/ makes a lot more sense to me.

Ovsyanka:
I think, that it is not good to store such things it the tests folder because there expected code tests. I mean tests, that actually runs code. In our case it is phpunit tests. Phpstun is more like code style checking.

I dont think that configuration files in the root folder is bad. Because it is standard place for it. And you can put it in some different folder instead tests, like "configs", "style-checking", etc.

But it is only my opinion, i am not insist on that and you may count it or not.

Ovsyanka:
Okay, It looks good now, I only have to to clear up commits history. Unfortunately, I can do it only tomorrow, but you have time to give me mooore crititism =)

Ovsyanka:
It is done.

dereuromark:
But wouldn't you agree, that phpcs and phpstan are part of the "test suite"? And not anything related to actually using code in production? Them being run inside travis suite kind of proves my point :)

Thus to bundle them with the tests folder seems more logical than anything else.

Ovsyanka:
@dereuromark, okay, I more agree with you then not =)

dereuromark:
No big deal either way - I was just trying to improve it a bit.

ravage84 (approved):

PR #10522, opened 2017-04-15
Update code in TimestampBehavior

mehdi-fathi:
Update code in TimestampBehavior for increase readbility.


src/ORM/Behavior/TimestampBehavior.php
```diff
@@ -136,8 +136,9 @@ public function implementedEvents()
      */
     public function timestamp(DateTime $ts = null, $refreshTimestamp = false)
     {
-        if ($ts) {
-            if ($this->_config['refreshTimestamp']) {
+        if ($ts instanceof DateTime) {
```
dereuromark: Should this be DateTimeInterface?

ndm2: I'd say if changing this all, then it should be an explicit not-`null` comparison, as the argument typhint restricts things to either `DateTime` or `null`, ie no other value can get there in the first place.

dereuromark:
What does this PR exactly do/change?
The existing check looked OK to me (it can either be null or an object, right?).

ADmad:
Closing as none of the changes done improve or fix anything.

mehdi-fathi:
we have 
`if ($class instanceof SessionHandlerInterface) {
            return $this->_engine = $class;
        }`
in 

> Cake\Network\Session.php line 245

ndm2:
@mehdi-fathi That is because unlike the `$ts` argument of the `TimestampBehavior::timestamp()` method, the `$class` argument of `Session::engine()` is not type-hinted, as it is also allowed to be a string.

Given that PHP does not allow union type-hints and doesn't support argument based method overloading, this is how things are beig handled for arguments that accept different types. You may want to check [**the PHP manual on Type declarations**](http://php.net/manual/en/functions.arguments.php#functions.arguments.type-declaration) for more information on how type-hinting in PHP works.

PR #10524, opened 2017-04-15
Add phpunit.xml.dist to .gitignore

mehdi-fathi:
add phpunit.xml.dist to .gitignore.


.gitignore
```diff
@@ -5,6 +5,7 @@
 /tags
 /composer.lock
 /phpunit.xml
+/phpunit.xml.dist
```
ADmad: This file is supposed to be present in the repo, not ignored.

src/ORM/Behavior/TimestampBehavior.php
```diff
@@ -136,8 +136,8 @@ public function implementedEvents()
      */
     public function timestamp(DateTime $ts = null, $refreshTimestamp = false)
     {
-        if ($ts) {
-            if ($this->_config['refreshTimestamp']) {
+        if ($ts instanceof DateTime) {
```
ADmad: Why is this change needed? The $ts paramater is typed already type hinted as DateTime so type check again is not necessary.

ADmad:
Duplicate of #10522

Issue #10521, opened 2017-04-14
Minus zero value does not pass notEmpty validation


Asgraf:
### What you did
````php
<?php
App::uses('AppShell', 'Console/Command');
App::uses('Validation', 'Utility');

class MinusZeroTestingShell extends AppShell {
	public  $User;

	function main() {
		$value_to_save1 = 0 * -0.5;
		$value_to_save2 = 0.0 * -1;
		$value_to_save3 = -0.0;

		var_dump($value_to_save1===0.0);
		var_dump($value_to_save2===0.0);
		var_dump($value_to_save3===0.0);

		var_dump((string)$value_to_save1);
		var_dump((string)$value_to_save2);
		var_dump((string)$value_to_save3);

		var_dump(Validation::notBlank($value_to_save1));
		var_dump(Validation::notBlank($value_to_save2));
		var_dump(Validation::notBlank($value_to_save3));
	}
}
````
### What happened
bool(true)
bool(true)
bool(true)
string(2) "-0"
string(2) "-0"
string(1) "0"
bool(false)
bool(false)
bool(true)

### What you expected to happen
bool(true)
bool(true)
bool(true)
string(2) "-0"
string(2) "-0"
string(1) "0"
bool(true)
bool(true)
bool(true)

### Tested on
PHP: 5.6.30
CakePHP: v2.9.7

This problem only occurs on cake 2.x
(on cake 3.x PHP's minus zero correctly passes notBlank validation)

**Comment by Asgraf** [comment]
Please note that writing -0.0 directly gives you positive zero instead negative zero.
Minus operator only change singn on zero values

**Comment by markstory** [comment]
Interesting, the following has both types of -0 being the same

```php
php > var_dump(-0.0);
php shell code:1:
double(-0)

php > var_dump(0.0 * -1);
php shell code:1:
double(-0)

php > var_dump(-0.0 === (0.0 * -1));
php shell code:1:
bool(true)
```

PR #10526, opened 2017-04-16
Fix notBlank() to pass on -0.0

markstory:
Copy the implementation from 3.x as it works with -0.0 already.

Refs #10521

### DIFF
--- a/lib/Cake/Test/Case/Utility/ValidationTest.php
+++ b/lib/Cake/Test/Case/Utility/ValidationTest.php
@@ -148,6 +148,10 @@ class ValidationTest extends CakeTestCase {
  * @return void
  */
 	public function testNotBlank() {
+		$this->assertTrue(Validation::notBlank(0));
+		$this->assertTrue(Validation::notBlank(0.0));
+		$this->assertTrue(Validation::notBlank(-0));
+		$this->assertTrue(Validation::notBlank(-0.0));
 		$this->assertTrue(Validation::notBlank('abcdefg'));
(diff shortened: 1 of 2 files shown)


dereuromark:
Looks good.

lib/Cake/Test/Case/Utility/ValidationTest.php
```diff
@@ -148,6 +148,10 @@ public function tearDown() {
  * @return void
  */
 	public function testNotBlank() {
+		$this->assertTrue(Validation::notBlank(0));
+		$this->assertTrue(Validation::notBlank(0.0));
+		$this->assertTrue(Validation::notBlank(-0));
+		$this->assertTrue(Validation::notBlank(-0.0));
```
Asgraf: @markstory Your test is testing for positive zero not negative one.
There only two ways to get minus zero value:
1) multiply negative float value by zero (-1.23 * 0)
2) multiply 0.0 by any negative value (0.0 * -1)
Minus operator does nothing on zero values so writing -0 or -0.0 gives you positive zero instead of expected negative zero!

markstory: Ok. I'll add another assertion then. It is odd that `var_dump()` shows `-0` 
```php
php > var_dump(-0.0);
php shell code:1:
double(-0)
```

markstory: Looks like the behavior of `-0.0` became correct in PHP7+, but in PHP5 `0.0 * -1` was required. See https://3v4l.org/a0FSJ

PR #10532, opened 2017-04-17
Added some missing annotations, cleaned up some code for best practice

Theaxiom:
In an effort to better understand CakePHP, I went through the entire core today and fixed any glaring issues that I came across. Most of the issues were missing annotations, however there were some minor code changes. I broke up my changes into separate commits for better clarity as well as to allow cherry-picking.

FYI: I avoided the best I could adding missing annotations for deprecated code, wasn't sure it was absolutely necessary.

~~P.S. I am still pushing commits, I am using this PR to help with automatic lint, etc. I will update this PR when I am done.~~

~~EDIT: I am done with this PR now, I did it this way to avoid a ton of PR's. This way, you can pick and choose what changes you want to bring into the core. All changes have been ran through static code analysis and there should not be any problems, but you guys are the experts so I will leave it to you at this point. Thanks!~~

UPDATE: I have split this PR into multiple PR's per request of core team. Please see below, referenced PR's.

PR #10533, opened 2017-04-17
Added some missing annotations

Theaxiom:


lorenzo:
Thanks for this, but without an explanation of why the exceptions are thrown I find the annotations to just add noise, especially because unlike java, PHP does not have checked exceptions so there is little value in annotating the methods without an explanation of what could happen.

dereuromark:
I also prefer Automation via sniffer if we add this to more than the actual throwing method. Or it can become Out of sync and even lying once the method loses the throw.

inoas:
I am with lorenzo. I think the exception annotations are good but only really useful when they have a textual explanation alongside.

Theaxiom:
All of these annotations were confirmed via automation and static code analysis. If you use this plug-in, it can help to ensure annotation integrity:

- https://plugins.jetbrains.com/plugin/7622-php-inspections-ea-extended-
- https://github.com/kalessil/phpinspectionsea

That being said, I understand your reasoning about having the annotations explained. I simply prefer it for IDE auto-completion. Feel free to close this PR.

ADmad:
I too am not thrilled about this additional annotations for exceptions. We should stick to listing only those that are directly throw by a method.

Theaxiom:
@ADmad ok I will remove them

dereuromark:
I have a sniffer we can use to assert exactly that. See Spryker ones, feel free to implemenent for cakephp for future consistency.

PR #10536, opened 2017-04-17
Removing some one-time use variables

Theaxiom:

### DIFF
--- a/src/Auth/BaseAuthenticate.php
+++ b/src/Auth/BaseAuthenticate.php
@@ -156,9 +156,7 @@ abstract class BaseAuthenticate implements EventListenerInterface
             $options['username'] = $username;
         }
 
-        $query = $table->find($finder, $options);
-
-        return $query;
+        return $table->find($finder, $options);
     }
 
     /**
--- a/src/Console/Shell.php
+++ b/src/Console/Shell.php
@@ -528,9 +528,8 @@ class Shell
     public function getOptionParser()
     {
         $name = ($this->plugin ? $this->plugin . '.' : '') . $this->name;
-        $parser = new ConsoleOptionParser($name);
 
-        return $parser;
+        return new ConsoleOptionParser($name);
     }
 
     /**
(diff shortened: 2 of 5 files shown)


ravage84:
The downside of this is that it makes it harder to debug these methods as you have to step out now first before you see the result...

rchavik:
I'm not keen on this.

Wouldn't the compiler/interpreter optimize these anyway?

rchavik:
anyway, this doesn't matter much.

but i was curious:

```php
<?php

function foo() {
    return rand(1, 10);
}

function bar() {
    $c = rand(1, 10);
    return $c;
}
```

`rachman@nucleus $ php7.1 -d opcache.opt_debug_level=0x20000 a.php `

```

$_main: ; (lines=1, args=0, vars=0, tmps=0)
    ; (after optimizer)
    ; /tmp/a.php:1-11
L0:     RETURN int(1)

bar: ; (lines=6, args=0, vars=1, tmps=1)
    ; (after optimizer)
    ; /tmp/a.php:7-10
L0:     INIT_FCALL 2 112 string("rand")
L1:     SEND_VAL int(1) 1
L2:     SEND_VAL int(10) 2
L3:     V1 = DO_ICALL
L4:     ASSIGN CV0($c) V1
L5:     RETURN CV0($c)

foo: ; (lines=5, args=0, vars=0, tmps=1)
    ; (after optimizer)
    ; /tmp/a.php:3-5
L0:     INIT_FCALL 2 112 string("rand")
L1:     SEND_VAL int(1) 1
L2:     SEND_VAL int(10) 2
L3:     V0 = DO_ICALL
L4:     RETURN V0
```

So, the answers seems to be 'no'.

markstory:
@rchavik I know in 2.x we saw decent performance improvements in `Inflector` by doing similar changes that only saved 1 or 2 op codes each. Those methods were called far more often than these ones are though.

Theaxiom:
Wouldn't it be nice if you could be lazy and the compiler would do this stuff for you? :D

PR #10540, opened 2017-04-17
Fixing a couple incrementers and decrementers

Theaxiom:


markstory:
Why use `++$var` over the more conventional `$var++`?

thinkingmedia:
@markstory Shouldn't the linter have caught this?

ADmad:
@thinkingmedia AFAIK we don't have any CS rule in place for that.

dereuromark:
Some  prefer to always use `++$var` to make sure it is changed prior to using it (and not accidentally using the former value). But without a sniffer in place it will be hard to reach a consistency here anyway :)

thinkingmedia:
@derekperkins 

```
   ++$var; // triggers no error.
   +$var; // triggers no error.
   $var++; // triggers no error.
   $var+; // syntax error
```

I think that's the only reason people prefer the post increment. There is nothing wrong with `++$var;` as a line of code, but an accidental delete of a `+` will yield `+$var;` and could go unnoticed by the programmer since there is no errors.

dereuromark:
@thinkingmedia 
The syntax issue sounds rather unrealistic to me, and probably quite obvious.
Whereas the hidden value issue often times is more common and of course more dangerous as well, doing something different without showing it to the developer.

Theaxiom:
@markstory fixed.

ADmad:
Needs rebase :)

Theaxiom:
Done.

ADmad:
Hmm.. All mysql builds are failing, though it seems to have been caused by another PR.

inoas:
I don't think we should have a linter on `$var++` vs `++$var` as they do different things.

However I do think that we should never use ++$var unless we need specifically what ++$var does over $var++ as a code style guideline (same for --).

markstory:
@ADmad I'll get those failures fixed tonight.

PR #10541, opened 2017-04-17
Using short syntax for applied operations

Theaxiom:

### DIFF
--- a/src/I18n/RelativeTimeFormatter.php
+++ b/src/I18n/RelativeTimeFormatter.php
@@ -202,7 +202,7 @@ class RelativeTimeFormatter
 
             if ($months >= 12) {
                 $years = floor($months / 12);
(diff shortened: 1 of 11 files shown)


src/I18n/RelativeTimeFormatter.php
```diff
@@ -231,25 +231,25 @@ protected function _diffData($futureTime, $pastTime, $backwards, $options)
             }
 
             if ($months >= 12) {
-                $years = $years + 1;
-                $months = $months - 12;
+                ++$years;
```
markstory: Can we use `$years++` instead?

Theaxiom: Sure, I will change it.

Theaxiom:
@markstory fixed.

PR #10544, opened 2017-04-17
Binary safe fopen()

Theaxiom:


markstory (changes requested):
Changes look good but there are conflicts now.

ADmad:
Needs rebase. The `b` option should be added only where actually needed. For eg. `.po` files aren't going to have binary content.

Theaxiom:
@ADmad it's a no-op where not needed.

PR #10546, opened 2017-04-17
Use null comparison instead

Theaxiom:

### DIFF
--- a/src/Cache/Engine/FileEngine.php
+++ b/src/Cache/Engine/FileEngine.php
@@ -388,7 +388,7 @@ class FileEngine extends CacheEngine
         if (!$createKey && !$path->isFile()) {
             return false;
         }
-        if (empty($this->_File) || $this->_File->getBasename() !== $key) {
+        if (null === $this->_File || $this->_File->getBasename() !== $key) {
             $exists = file_exists($path->getPathname());
             try {
                 $this->_File = $path->openFile('c+');
--- a/src/Console/Shell.php
+++ b/src/Console/Shell.php
@@ -258,7 +258,7 @@ class Shell
      */
     public function loadTasks()
     {
-        if ($this->tasks === true || empty($this->tasks) || empty($this->Tasks)) {
+        if ($this->tasks === true || empty($this->tasks) || null === $this->Tasks) {
             return true;
         }
         $this->_taskMap = $this->Tasks->normalizeArray((array)$this->tasks);
@@ -767,7 +767,7 @@ class Shell
     {
         $this->_io->err(sprintf('<error>Error:</error> %s', $title));
 
-        if (!empty($message)) {
+        if (null !== $message) {
             $this->_io->err($message);
         }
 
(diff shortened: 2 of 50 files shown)


saeideng:
:-1:  for master target

src/Cache/Engine/FileEngine.php
```diff
@@ -388,7 +388,7 @@ protected function _setKey($key, $createKey = false)
         if (!$createKey && !$path->isFile()) {
             return false;
         }
-        if (empty($this->_File) || $this->_File->getBasename() !== $key) {
+        if (null === $this->_File || $this->_File->getBasename() !== $key) {
```
AD7six: Are there yoda-comparisons anywhere else in the framework?

dereuromark: a) Please dont use YODA, === null always.
b) I agree that this change is not safe for master, as it can break with certain nullish values coming in that so far "just worked" (even if not intended by the code).

Theaxiom:
Most of this exercise for me is for learning, so I do appreciate the feedback on these PR's.

dereuromark:
@Theaxiom we have to be careful not to break BC with cosmetic (non functional) changes. Especially with a (still) non-strict-typed language we must accept certain unclean content being introduced in some ways, and most certainly around the topic "null".

Theaxiom:
Thanks for the clarification, most of these PR's are strictly cosmetic. There are a couple performance enhancements, albeit very few, in some of them.

src/Cache/Engine/FileEngine.php
Theaxiom: What's wrong with YODA? Just curious...

lorenzo: Is just not our style

lorenzo:
I'm preemptively voting against this PR. We already had several cases in the past where changing an `empty` call to something else caused new bugs, even when we thought the change was only cosmetic.

markstory:
Going to close this one as yoda conditions are not used in CakePHP and I share @lorenzo s concern that some of these changes could change behaviour.

src/Cache/Engine/FileEngine.php
inoas: my 2 cents on Yoda:

It creates a longer thought process. You bascially read:
1. Entity null
2. has to equal something
1.+2. entity null has to equal something
3. $this->_File
4. Entity Null has to Equal $this->_File

It is like double negation. You go one extra circle in your through process. You have to read from right to left to circumvent that.

When it is $foo === generic_entity/statement (such as null, false, true) it is much easier than the other way around.

Also for comparison I feel many peeps still "assign" things under the condition of that it can be assigned. "Can Entity Null be assigned to the Equality of some other Entity"? That's a hard way to look at it ... "Can Entity X be assigned to the Equality of Entity Null" is a lot easier for my mind.

Anyway, I don't think this PR got rejected because of YODA style as that can easily be fixed.

dereuromark: As a side note: In the past it was far more valuable to remove the cloaking empty() and isset() checks in favor of !
As !$this->_File behaves exactly the same, but any typo would be recognized immediately - whereas with the cloaking checks typos could lead to dead never-executable code without you knowing it (we had that issue in the past).

PR #10547, opened 2017-04-17
Extending some return types in annotation docblocks

Theaxiom:

### DIFF
--- a/src/Console/ConsoleOptionParser.php
+++ b/src/Console/ConsoleOptionParser.php
@@ -298,7 +298,7 @@ class ConsoleOptionParser
      *
      * @deprecated 3.4.0 Use setCommand()/getCommand() instead.
      * @param string|null $text The text to set, or null if you want to read
-     * @return string|$this If reading, the value of the command. If setting $this will be returned.
+     * @return ConsoleOptionParser|string If reading, the value of the command. If setting $this will be returned.
      */
     public function command($text = null)
     {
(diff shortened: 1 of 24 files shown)


src/Console/ConsoleOptionParser.php
```diff
@@ -298,7 +298,7 @@ public function getCommand()
      *
      * @deprecated 3.4.0 Use setCommand()/getCommand() instead.
      * @param string|null $text The text to set, or null if you want to read
-     * @return string|$this If reading, the value of the command. If setting $this will be returned.
+     * @return ConsoleOptionParser|string If reading, the value of the command. If setting $this will be returned.
```
dereuromark: Please use FQCN in doc blocks: \...\

dereuromark: In these cases $this is correct though afaik - the IDE knows that this is chaining and knows the type implicitly.

dereuromark:
👎 as chaining demands this kind of doc block syntax according to phpdocs.

src/Console/ConsoleOptionParser.php
Theaxiom: Alright, I will update them to FQCN's.

ADmad:
Please don't change any usage of `$this` in param types. It's usage has been decided through enough discussion in the past based on what IDE's expect.

PR #10548, opened 2017-04-17
Removing default null property assignments. Class property default va…

Theaxiom:


src/Cache/Engine/FileEngine.php
```diff
@@ -38,7 +38,7 @@ class FileEngine extends CacheEngine
      *
      * @var \SplFileObject|null
      */
-    protected $_File = null;
+    protected $_File;
```
dereuromark: We used to keep them so far for making it more clear (see doc block), when the value can also be of type null after constructor run.
I agree that we can remove them for all cases where null is not a valid value.

PR #10550, opened 2017-04-17
Ignore debug.log and error.log

cleptric:
It's always tempting to commit `debug.log` and `error.log` when doing a PR.
@dereuromark Can you add the `plugin` thing?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,8 @@
 /phpcs.xml
 /vendor
 *.mo
+debug.log
+error.log
 
 # IDE and editor specific files #
 #################################


ADmad:
Would be nice if we could find the offending test which creates these log files on root.

cleptric:
`ConsoleErrorHandlerTest::testHandleFatalError` adds the `error.log`

cleptric:
`ConsoleErrorHandlerTest::testHandleError` add the `debug.log` in the first place, but a some other test write to this file as well.
`Debug: command 'sample' in plugin 'TestPlugin' was not aliased, conflicts with another shell` and some stuff from `DebuggerTest`

dereuromark (approved):
I think it is fair enough to add those two lines.

Issue #10555, opened 2017-04-18
Issue with routing parameters and testAction after adding PR #10517


liviascapin:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 2.9.7 **after manually adding the fix from https://github.com/cakephp/cakephp/pull/10517**

### What you did
After adding the fix from @ndm2 we continued to write more unit tests and noticed that query parameters do not work anymore. The testAction call:

```php
$viewVars = $this->testAction(
	array(
		'controller' => $this->_controllerName,
		'action' => 'manage_orders',
		'?' => array(
			'show_thank_you' => '1',
		)
	),
	array(
		'method' => 'GET',
		'return' => 'vars',
	)
);
```

### What happened
The following would be the content of `$request` on line 254 (https://github.com/ndm2/cakephp/blob/7d74818d9a0a7cd3bddb49b3192927a8d5a4444c/lib/Cake/TestSuite/ControllerTestCase.php#L254) **before** the fix:
```php
//Without Fix, cli
Mock_CakeRequest_4df83257::__set_state(array(
   '__phpunit_invocationMocker' => NULL,
   'params' => 
  array (
    'plugin' => NULL,
    'controller' => NULL,
    'action' => NULL,
    'named' => 
    array (
    ),
    'pass' => 
    array (
    ),
  ),
   'data' => 
  array (
  ),
   'query' => 
  array (
    'show_thank_you' => '1',
  ),
   'url' => 'Account/manage_orders',
   'base' => 'F%3A%5Cdev%5Cxampp%5Chtdocs%5Cmy_app%5Capp%5CConsole',
   'webroot' => 'F%3A%5Cdev%5Cxampp%5Chtdocs%5Cmy_app%5Capp%5CConsole/',
   'here' => 'F%3A%5Cdev%5Cxampp%5Chtdocs%5Cmy_app%5Capp%5CConsole/Account/manage_orders',
   '_detectors' => 
  array (
    'get' => 
    array (
      'env' => 'REQUEST_METHOD',
      'value' => 'GET',
    ),
    'patch' => 
    array (
      'env' => 'REQUEST_METHOD',
      'value' => 'PATCH',
    ),
    'post' => 
    array (
      'env' => 'REQUEST_METHOD',
      'value' => 'POST',
    ),
    'put' => 
    array (
      'env' => 'REQUEST_METHOD',
      'value' => 'PUT',
    ),
[...]

PR #10557, opened 2017-04-18
2.x - Fix query string data in URL arrays not being passed anymore (#10517 follow-up)

ndm2:
Fix query string data in URL arrays not being passed anymore.

refs #10555, #10517, #5473


markstory:
Nice work @ndm2

PR #10558, opened 2017-04-18
extract controller name logic into another method

steinkel:


src/Http/ControllerFactory.php
```diff
@@ -65,16 +91,7 @@ public function create(ServerRequest $request, Response $response)
...
-        if (!$className) {
-            $this->missingController($request);
-        }
-        $reflection = new ReflectionClass($className);
-        if ($reflection->isAbstract() || $reflection->isInterface()) {
-            $this->missingController($request);
-        }
-
-        return $reflection->newInstance($request, $response, $controller);
+        return App::className($pluginPath . $controller, $namespace, 'Controller');
```
dereuromark: return ... ?: null;

would be cleaner, returning more than null as alternative is a code smell usually.

src/Http/ControllerFactory.php
```diff
@@ -33,14 +33,40 @@ class ControllerFactory
...
+        }
+
+        return $reflection->newInstance($request, $response, $controller);
+    }
+
+    /**
+     * Determine the controller class name based on current request and controller param
+     *
+     * @param \Cake\Http\ServerRequest $request The request to build a controller for.
+     * @param string $controllerName The controller name present in the request params
```
dereuromark: |null missing

steinkel:
@dereuromark thank you for the review

dereuromark:
This looks BC, could also have gone in master. But 3.next sure is also fine.

src/Http/ControllerFactory.php
```diff
@@ -33,14 +33,40 @@ class ControllerFactory
      */
     public function create(ServerRequest $request, Response $response)
     {
-        $pluginPath = $controller = null;
+        $controller = null;
+        if ($request->getParam('controller')) {
+            $controller = $request->getParam('controller');
+        }
```
markstory: Do you need these 3 lines? They are duplicated on 64-66.

markstory (commented):
I'm guessing you have a use case for this being a separate method? What is it?

src/Http/ControllerFactory.php
steinkel: Trying to keep it simple as $controller is used in below https://github.com/cakephp/cakephp/pull/10558/files/4a49f1e8393bb259cf36c4455d1b42f87bdc4e63#diff-4c9ecd90d3740d54e09028ff7be12e1eR49

But makes sense to allow others to use the new getControllerClass method without passing the controller.

steinkel:
The use case for this is: while working on this ticket https://github.com/CakeDC/users/issues/527 I got to a point where I would like to check if the controller is exactly the same as the controller specified by a set of params (a rbac rule). Following the the dispatch process I found this code as part of the controller factory, so the refactor.

src/Http/ControllerFactory.php
markstory: Then could the parameter be required for `getControllerClass`?

steinkel: I added as optional to let others pass only the request.

markstory: I would like to not have the duplicate code though.

steinkel: ok, I'll take a look later to refactor that too, thanks

src/Http/ControllerFactory.php
```diff
@@ -46,22 +42,20 @@ public function create(ServerRequest $request, Response $response)
             $this->missingController($request);
         }
 
-        return $reflection->newInstance($request, $response, $controller);
+        return $reflection->newInstance($request, $response);
```
markstory: This should be safe. The controller uses the request params as a fallback for the name.

tests/TestCase/Http/ControllerFactoryTest.php
```diff
@@ -263,4 +263,23 @@ public function testAbsoluteReferenceFailure()
         ]);
         $this->factory->create($request, $this->response);
     }
+
+    /**
+     * Test building controller name passing no controller name
```
ravage84: Sentence is somewhat confusing, may be changing to "name when passing" helps )If I understood correctly).

steinkel: fixed, thanks

PR #10560, opened 2017-04-19
Added documentation for App.base

liviakuenzli:
As discussed in https://github.com/cakephp/docs/issues/4895 and  the related https://github.com/cakephp/docs/issues/4896.


markstory:
Thanks!

PR #10561, opened 2017-04-19
Turning the CSRF component into a middleware

burzum:
What the title says.

### DIFF
--- a/src/Http/Middleware/CsrfProtectionMiddleware.php
+++ b/src/Http/Middleware/CsrfProtectionMiddleware.php
@@ -159,7 +159,7 @@ class CsrfProtectionMiddleware
      * @param \Cake\Http\Response $response The response to augment
      * @return \Cake\Http\Response Modified response
      */
-    protected function _addTokenCookie($token, $request, $response)
+    protected function _addTokenCookie($token, ServerRequestInterface $request, ResponseInterface $response)
     {
         $expiry = new Time($this->_config['expiry']);
 
--- a/src/Http/Middleware/CsrfProtectionMiddleware.php
+++ b/src/Http/Middleware/CsrfProtectionMiddleware.php
@@ -32,7 +32,7 @@ use Psr\Http\Message\ServerRequestInterface;
  * If the request data is missing or does not match the cookie data,
  * an InvalidCsrfTokenException will be raised.
  *
- * This component integrates with the FormHelper automatically and when
+ * This middleware integrates with the FormHelper automatically and when
  * used together your forms will have CSRF tokens automatically added
  * when `$this->Form->create(...)` is used in a view.
  */
@@ -72,7 +72,7 @@ class CsrfProtectionMiddleware
      */
     public function __construct(array $config = [])
     {
-        $this->_config = $this->_defaultConfig + $config;
+        $this->_config = $config + $this->_defaultConfig;
     }
 
     /**
@@ -88,7 +88,7 @@ class CsrfProtectionMiddleware
         $cookies = $request->getCookieParams();
         $cookieData = null;
         if (isset($cookies[$this->_config['cookieName']])) {
-            $cookieData = isset($cookies[$this->_config['cookieName']]);
+            $cookieData = $cookies[$this->_config['cookieName']];
         }
 
         if (!empty($cookieData)) {
@@ -103,12 +103,12 @@ class CsrfProtectionMiddleware
         }
 
         if ($method === 'GET' && $cookieData === null) {
-            $this->_setCookie($request, $response);
+            $this->_setToken($request, $response);
 
             return $next($request, $response);
         }
 
-        $this->_validateAndUnsetTokenField($request);
+        $request = $this->_validateAndUnsetTokenField($request);
 
         return $next($request, $response);
     }
@@ -119,7 +119,7 @@ class CsrfProtectionMiddleware
      * @param \Cake\Http\ServerRequest $request The request object.
      * @return void
      */
-    protected function _validateAndUnsetTokenField(ServerRequest &$request)
+    protected function _validateAndUnsetTokenField(ServerRequest $request)
     {
         if (in_array($request->getMethod(), ['PUT', 'POST', 'DELETE', 'PATCH']) || $request->getData()) {
             $this->_validateToken($request);
@@ -129,10 +129,12 @@ class CsrfProtectionMiddleware
                 $request = $request->withParsedBody($body);
             }
         }
+
+        return $request;
     }
 
     /**
-     * Set the cookie in the response.
+     * Set the token in the response.
      *
      * Also sets the request->params['_csrfToken'] so the newly minted
      * token is available in the request data.
@@ -141,7 +143,7 @@ class CsrfProtectionMiddleware
      * @param \Cake\Http\Response $response The response object.
      * @return void
      */
-    protected function _setCookie(ServerRequest &$request, Response &$response)
+    protected function _setToken(ServerRequest &$request, Response &$response)
     {
         $expiry = new Time($this->_config['expiry']);
         $value = hash('sha512', Security::randomBytes(16), false);
--- a/src/Http/Middleware/CsrfProtectionMiddleware.php
+++ b/src/Http/Middleware/CsrfProtectionMiddleware.php
@@ -102,7 +102,7 @@ class CsrfProtectionMiddleware
 
         $method = $request->getMethod();
         if ($method === 'GET' && $cookieData === null) {
-            $this->_setToken($request, $response);
+            list($request, $response) = $this->_setToken($request, $response);
 
             return $next($request, $response);
         }
@@ -140,9 +140,9 @@ class CsrfProtectionMiddleware
      *
      * @param \Psr\Http\Message\ServerRequestInterface $request The request object.
      * @param \Psr\Http\Message\ResponseInterface $response The response object.
-     * @return void
+     * @return array
      */
-    protected function _setToken(ServerRequestInterface &$request, ResponseInterface &$response)
+    protected function _setToken(ServerRequestInterface $request, ResponseInterface $response)
     {
         $expiry = new Time($this->_config['expiry']);
         $value = hash('sha512', Security::randomBytes(16), false);
@@ -158,6 +158,8 @@ class CsrfProtectionMiddleware
             'secure' => $this->_config['secure'],
             'httpOnly' => $this->_config['httpOnly'],
         ]);
+
+        return [$request, $response];
     }
 
     /**
--- a/src/Http/Middleware/CsrfProtectionMiddleware.php
+++ b/src/Http/Middleware/CsrfProtectionMiddleware.php
@@ -39,11 +39,11 @@ class CsrfProtectionMiddleware
     /**
      * Default config for the CSRF handling.
      *
-     *  - cookieName = The name of the cookie to send.
-     *  - expiry = How long the CSRF token should last. Defaults to browser session.
-     *  - secure = Whether or not the cookie will be set with the Secure flag. Defaults to false.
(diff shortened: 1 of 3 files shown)


src/Http/Middleware/CsrfProtectionMiddleware.php
```diff
@@ -0,0 +1,184 @@
...
+    protected $_config = [];
+
+    /**
+     * Constructor
+     *
+     * @param array $config Config options
+     */
+    public function __construct(array $config = [])
+    {
+        $this->_config = $this->_defaultConfig + $config;
```
markstory: This merge is backwards.

src/Http/Middleware/CsrfProtectionMiddleware.php
```diff
@@ -0,0 +1,184 @@
...
+     * @param \Psr\Http\Message\ResponseInterface $response The response.
+     * @param callable $next Callback to invoke the next middleware.
+     * @return \Psr\Http\Message\ResponseInterface A response
+     */
+    public function __invoke(ServerRequestInterface &$request, ResponseInterface $response, $next)
+    {
+        $cookies = $request->getCookieParams();
+        $cookieData = null;
+        if (isset($cookies[$this->_config['cookieName']])) {
+            $cookieData = isset($cookies[$this->_config['cookieName']]);
```
markstory: Won't this make the cookieData always true?

src/Http/Middleware/CsrfProtectionMiddleware.php
```diff
@@ -0,0 +1,184 @@
...
+        }
+
+        if (!empty($cookieData)) {
+            $params = $request->getAttribute('params');
+            $params['_csrfToken'] = $cookieData;
+            $request = $request->withAttribute('params', $params);
+        }
+
+        $method = $request->getMethod();
+        if ($method === 'requested') {
```
markstory: I don't think this works.

src/Http/Middleware/CsrfProtectionMiddleware.php
```diff
@@ -0,0 +1,184 @@
...
+        return $next($request, $response);
+    }
+
+    /**
+     * Checks if the request is POST, PUT, DELETE or PATCH and validates the CSRF token
+     *
+     * @param \Cake\Http\ServerRequest $request The request object.
+     * @return void
+     */
+    protected function _validateAndUnsetTokenField(ServerRequest &$request)
```
markstory: Use a return value and not pass by reference​.

src/Http/Middleware/CsrfProtectionMiddleware.php
```diff
@@ -0,0 +1,184 @@
...
+        $request = $request->withAttribute('params', $params);
+
+        $response = $response->withCookie($this->_config['cookieName'], [
+            'value' => $value,
+            'expire' => $expiry->format('U'),
+            'path' => $request->getAttribute('webroot'),
+            'secure' => $this->_config['secure'],
+            'httpOnly' => $this->_config['httpOnly'],
+        ]);
+    }
```
markstory: You're missing a return here.

src/Http/Middleware/CsrfProtectionMiddleware.php
chrisShick: I think this was just a copy&paste error!

src/Http/Middleware/CsrfProtectionMiddleware.php
robertpustulka: Can't you use `InstanceConfigTrait`?

src/Http/Middleware/CsrfProtectionMiddleware.php
```diff
@@ -0,0 +1,184 @@
...
+     * @return void
+     */
+    protected function _setCookie(ServerRequest &$request, Response &$response)
+    {
+        $expiry = new Time($this->_config['expiry']);
+        $value = hash('sha512', Security::randomBytes(16), false);
+
+        $params = $request->getAttribute('params');
+        $params['_csrfToken'] = $value;
+        $request = $request->withAttribute('params', $params);
```
robertpustulka: This part does not "set cookie" but modifies the request, so it probably should be in another method.

src/Http/Middleware/CsrfProtectionMiddleware.php
```diff
@@ -0,0 +1,184 @@
...
+use Cake\I18n\Time;
+use Cake\Network\Exception\InvalidCsrfTokenException;
+use Cake\Utility\Security;
+use Psr\Http\Message\ResponseInterface;
+use Psr\Http\Message\ServerRequestInterface;
+
+/**
+ * Provides CSRF protection & validation.
+ *
+ * This component adds a CSRF token to a cookie. The cookie value is compared to
```
inoas: IMHO: `class CsrfProtectionCookieMiddleware` or `CsrfCookieMiddleware` or then?

src/Http/Middleware/CsrfProtectionMiddleware.php
```diff
@@ -0,0 +1,184 @@
...
+
+    /**
+     * Serve assets if the path matches one.
+     *
+     * @param \Psr\Http\Message\ServerRequestInterface $request The request.
+     * @param \Psr\Http\Message\ResponseInterface $response The response.
+     * @param callable $next Callback to invoke the next middleware.
+     * @return \Psr\Http\Message\ResponseInterface A response
+     */
+    public function __invoke(ServerRequestInterface &$request, ResponseInterface $response, $next)
```
inoas: Is pass by reference on purpose?

src/Http/Middleware/CsrfProtectionMiddleware.php
```diff
@@ -0,0 +1,184 @@
...
+
+    /**
+     * Checks if the request is POST, PUT, DELETE or PATCH and validates the CSRF token
+     *
+     * @param \Cake\Http\ServerRequest $request The request object.
+     * @return void
+     */
+    protected function _validateAndUnsetTokenField(ServerRequest &$request)
+    {
+        if (in_array($request->getMethod(), ['PUT', 'POST', 'DELETE', 'PATCH']) || $request->getData()) {
```
inoas: IMHO always use `in_array($x, $y, true)` (as we always use `===`) unless there is a reason not to.

robertpustulka: There are no falsey ot truey values in an array so it shouldn't be a problem?

src/Http/Middleware/CsrfProtectionMiddleware.php
burzum: Why do we *need* this? Instead of using yet another config array I'm thinking about changing the config to setters.

src/Http/Middleware/CsrfProtectionMiddleware.php
burzum: Yes

src/Http/Middleware/CsrfProtectionMiddleware.php
robertpustulka: @burzum the trait already has getters and setters for config :)

src/Http/Middleware/CsrfProtectionMiddleware.php
burzum: No, because `_setCookie(ServerRequest &$request, Response &$response)`. I've tried to return an array with both objects before but this leads into failing tests because the reference gets lost somewhere. Feel free to try it and remove the `&` and return this array:

```php
        return [
            'request' => $request,
            'response' => $response
        ];
```

src/Http/Middleware/CsrfProtectionMiddleware.php
markstory: I'm not sure we need the config trait. I don't see many scenarios where configuration needs to be modified after the middleware is created.

src/Http/Middleware/CsrfProtectionMiddleware.php
```diff
@@ -0,0 +1,184 @@
...
+     * Set the cookie in the response.
+     *
+     * Also sets the request->params['_csrfToken'] so the newly minted
+     * token is available in the request data.
+     *
+     * @param \Cake\Http\ServerRequest $request The request object.
+     * @param \Cake\Http\Response $response The response object.
+     * @return void
+     */
+    protected function _setCookie(ServerRequest &$request, Response &$response)
```
markstory: Another place we should avoid pass by ref.

tests/TestCase/Http/Middleware/CsrfProtectionMiddlewareTest.php
```diff
@@ -0,0 +1,177 @@
...
+
+    /**
+     * teardown
+     *
+     * @return void
+     */
+    public function tearDown()
+    {
+        parent::tearDown();
+    }
```
markstory: Do you need these no-op methods?

src/Http/Middleware/CsrfProtectionMiddleware.php
burzum: Try to make asserts on the request object without it. No idea how else I could test it.

robertpustulka: Expect method call on a mock?
```php
$request = $this->createMock(RequestInterface::class);
$request->expect('metod')->with($args)->willReturn($request);
```
?

burzum: Got it to work by checking the asserts in the closure thanks to a friendly hint from @markstory

src/Http/Middleware/CsrfProtectionMiddleware.php
burzum: I haven't added it for exactly the reason @markstory mentioned. I don't the a need to add it.

ADmad:
I have pushed commits to
- Update docblocks.
- Remove pass by reference in `_setToken()` method.
- Removed check/support for "requestAction()` requests.

ADmad (changes requested):

src/Http/Middleware/CsrfProtectionMiddleware.php
inoas: I am not saying it matters in this case. In many cases using empty()/isset() instead of === null or array_key_exists do not matter.

I am just saying this should be the default. We should always be strict with types UNLESS we have to be weak with it for whatever reason.

burzum:
@markstory @ADmad @dereuromark @robertpustulka so what is left to get this merged?

src/Http/Middleware/CsrfProtectionMiddleware.php
```diff
@@ -0,0 +1,197 @@
...
+
+    /**
+     * Add a CSRF token to the response cookies.
+     *
+     * @param string $token The token to add.
+     * @param \Psr\Http\Message\ServerRequestInterface $request The to get cookie path data from
+     * @param \Cake\Http\Response $response The response to augment
+     * @return \Cake\Http\Response Modified response
+     */
+    protected function _addTokenCookie($token, $request, $response)
```
ADmad: $request, $response params should perhaps be typehinted as done for other methods?

src/Http/Middleware/CsrfProtectionMiddleware.php
```diff
@@ -159,7 +159,7 @@ protected function _addTokenToRequest($token, ServerRequestInterface $request)
      * @param \Cake\Http\Response $response The response to augment
```
robertpustulka: These should be changed as well.

src/Http/Middleware/CsrfProtectionMiddleware.php
burzum: @markstory any feedback on this?

markstory: All good now. :+1:

markstory (approved):
Looks good to me.

Issue #10115, opened 2017-01-27
Call to protected method of fallback translator Aura\Intl\Translator::getMessage()


JoepRoebroek:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.3.12
* Platform and Target: Nginx, MySQL, Ubuntu 14.04.

### What you did
Updated from 3.3.11 to 3.3.12. 

### What happened
Got the following error on most pages:
```
Fatal error: Uncaught Error: Call to protected method Aura\Intl\Translator::getMessage() from context 'Cake\I18n\Translator' in /var/www/site/vendor/cakephp/cakephp/src/I18n/Translator.php on line 97
```

### What you expected to happen
No error.

### Possible fix

https://github.com/cakephp/cakephp/blob/master/src/I18n/Translator.php#L95 tries to execute protected method on fallback class. `getMessage` should either be replaced by `translate` or the fallback should be moved to the translate function, which is more neat I guess.

markstory:
What version of aura/intl do you have? I don't get the same error with aura/intl 1.1.1. Are you customizing the fallback loader at all?

JoepRoebroek:
Version 1.1.1, installed via composer. Did not customize the fallback loader. But I think it's a bug in their code as well, since `$this->fallback` should adhere to `TranslatorInterface` which does not specify `getMessage` only `translate`.

raul338:
I had the same bug only in production. Somehow it solved by nuking vendor directory and reinstalling via `composer install`

JoepRoebroek:
@raul338 tried your suggestion, but did not solve the problem for me. If I have time this weekend I'll check if I can create a small project that reproduces the problem consistently.

markstory:
@JoepRoebroek I double checked that the tests do hit the fallback loader (they do). I'm not sure why we don't see the protected method errors in our own test harness.

JoepRoebroek:
@markstory ok, probably it's environment specific. I'll try and see if I can figure it out over the weekend. For now, maybe it's good to check php versions, mine is at:
`PHP 7.0.15-1+deb.sury.org~trusty+1`
Forgot to mention this in the original report.

markstory:
Could you try clearing out the translation cache, and possibly deleting and reinstalling aura/intl?

JoepRoebroek:
[...]

PR #10562, opened 2017-04-19
Fix translator package initialization from cache

JoepRoebroek:
This should fix the issue reported by mamchenkov in https://github.com/cakephp/cakephp/issues/10115


lorenzo:
Interesting, thanks for the fix!!

Do you think you can also add a test case for this fix to avoid future regressions?

JoepRoebroek:
@lorenzo Thanks, I've added the regression test.

Issue #10235, opened 2017-02-16
RadioWidget Label attributes should indicate which button is selected.


dwightnetzango:
The Radio widget renders multiple label - inputs for each field.  One for each of the options.

In the RadioWidget->_renderInput method, it identifies if each individual input should be having the active value, however, there is nothing to tell the label that that label is for a an input that is the current value.

This is a very easy, one line fix which allows radio boxes to be rendered as pretty boxes using 
   Button.prototype.toggle inside bootstrap.js 
```
        if (isset($data['val']) && (string)$data['val'] === (string)$radio['value']) {
            $radio['checked'] = true;
            $radio['templateVars']['activeClass'] = "active";// new line
        }

```

dereuromark:
@dwightnetzango Are you able to provide a PR with your suggested enhancements?

markstory:
Closing as there is a pull request open now.

PR #10564, opened 2017-04-20
Radio Labels indicate which button is selected.

dwilbanks:
https://github.com/cakephp/cakephp/issues/10235

### DIFF
--- a/src/View/Widget/RadioWidget.php
+++ b/src/View/Widget/RadioWidget.php
@@ -172,6 +172,7 @@ class RadioWidget implements WidgetInterface
         }
         if (isset($data['val']) && (string)$data['val'] === (string)$radio['value']) {
             $radio['checked'] = true;
+            $radio['templateVars']['activeClass'] = "active";
         }
 
         if (!is_bool($data['label']) && isset($radio['checked']) && $radio['checked']) {
--- a/src/View/Widget/RadioWidget.php
+++ b/src/View/Widget/RadioWidget.php
@@ -172,7 +172,7 @@ class RadioWidget implements WidgetInterface
         }
         if (isset($data['val']) && (string)$data['val'] === (string)$radio['value']) {
             $radio['checked'] = true;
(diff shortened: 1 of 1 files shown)


src/View/Widget/RadioWidget.php
```diff
@@ -172,6 +172,7 @@ protected function _renderInput($val, $text, $data, $context)
         }
         if (isset($data['val']) && (string)$data['val'] === (string)$radio['value']) {
             $radio['checked'] = true;
+            $radio['templateVars']['activeClass'] = "active";
```
markstory: Any chance a test could be added to ensure this templateVar is not lost in future refactorings?

markstory (changes requested):

ADmad:
Don't really like this. Why can't you just use `input:checked` selector?

ADmad:
From your original issue I see you want it mainly to target the labels when using bootstrap. In that case I would prefer this change was done in [friendsofcake/bootstrap-ui](https://github.com/FriendsOfCake/bootstrap-ui) instead.

dwilbanks:
Bootstrap is one example, but, should not be the only.  

The friendsofcake/bootstrap-ui class does not implement the _renderInput method is not being implemented by friendsofcake, they inherit from you.

If the change is implemented in friendsofcake, they will be required to re-implement this single line every time changes are made in cakephp\cakephp.

Don't you think that the label associated with the checked radio item should be allowed to style differently than all others?

ADmad:
Fair point about having to override the whole `_renderInput()` method.

I am just concerned about adding new harcoded classes. But since this change is being made to `templateVars` is shouldn't affect default templates, so I guess it's relatively safe.

src/View/Widget/RadioWidget.php
ADmad: Please use single quotes for consistency.

dwilbanks: oops.

Issue #10565, opened 2017-04-20
Method setData does not exist


infokani:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.4.5.
* Platform and Target: xampp php7.

### What you did
I got an error when I use setData in the controller.

### What happened
.Method setData does not exist

### What you expected to happen
setData method missing

markstory:
Why are you using `setData()`?

infokani:
```
 public function add() {
    ...
              $this->request->setData('field',$value); // is correct?
    ...
```

markstory:
No. Those methods don't exist. However, there is a deprecation doc string suggesting them which is incorrect. You should use `withParam()` or `param()` instead. I'll get the deprecation notice updated as the current documentation is wrong.

infokani:
I found setData method in the Event\Event.php
```
public function setData($key, $value = null)
    {
        if (is_array($key)) {
            $this->_data = $key;
        } else {
            $this->_data[$key] = $value;
        }

        return $this;
    }
```

infokani:
Should I use old $this->request->data['field'] = $value; ?

markstory:
If you want to modify the request data you can use `$this->request->data['field']`, $this->request->data()` or `$this->request->withData()`. The `withData()` method is what I would recommend as it is not deprecated, but the other forms will continue to exist for the remainder of 3.x.

markstory:
Closing as there is a pull request open to fix the incorrect documentation.

PR #10566, opened 2017-04-20
Update deprecation suggestions

markstory:
Don't recommend methods that don't exist, it is confusing and not helpful.

Refs #10565

### DIFF
--- a/src/Http/ServerRequest.php
+++ b/src/Http/ServerRequest.php
@@ -2124,8 +2124,9 @@ class ServerRequest implements ArrayAccess, ServerRequestInterface
      * @param string $name Name of the key being written
      * @param mixed $value The value being written.
      * @return void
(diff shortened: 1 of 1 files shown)


src/Http/ServerRequest.php
```diff
@@ -2124,8 +2124,9 @@ public function offsetGet($name)
      * @param string $name Name of the key being written
      * @param mixed $value The value being written.
      * @return void
-     * @deprecated 3.4.0 The ArrayAccess methods will be removed in 4.0.0. Use setParam(), setData() and setQuery() instead.
+     * @deprecated 3.4.0 The ArrayAccess methods will be removed in 4.0.0. Use withParam() or param() instead.
      */
+
```
dereuromark: this newline looks like it shouldnt have been added.

PR #10567, opened 2017-04-20
Update Response classes to accept custom implementations of CookieInterface.

robertpustulka:
`Cake\Http\Client\Response` and `Cake\Client\Response` would call methods that wasn't a part of `CookieInterface`.

Refs #10406

### DIFF
--- a/src/Http/Client/CookieCollection.php
+++ b/src/Http/Client/CookieCollection.php
@@ -14,6 +14,7 @@
 namespace Cake\Http\Client;
 
 use Cake\Http\Cookie\CookieCollection as BaseCollection;
+use Cake\Http\Cookie\CookieInterface;
 
 /**
  * Container class for cookies used in Http\Client.
@@ -32,7 +33,7 @@ class CookieCollection extends BaseCollection
      * Store the cookies that haven't expired. If a cookie has been expired
      * and is currently stored, it will be removed.
      *
-     * @param \Cake\Http\Client\Response $response The response to read cookies from
+     * @param Response $response The response to read cookies from
      * @param string $url The request URL used for default host/path values.
      * @return void
      */
@@ -78,10 +79,31 @@ class CookieCollection extends BaseCollection
     {
         $out = [];
         foreach ($this->cookies as $cookie) {
-            $out[] = $cookie->toArray();
+            $out[] = $this->convertCookie($cookie);
         }
 
         return $out;
+    }
+
+    /**
+     * Convert the cookie into an array of its properties.
+     *
+     * Primarily useful where backwards compatibility is needed.
+     *
+     * @param \Cake\Http\Cookie\CookieInterface $cookie Cookie object.
+     * @return array
+     */
+    public function convertCookie(CookieInterface $cookie)
+    {
+        return [
+            'name' => $cookie->getName(),
+            'value' => $cookie->getValue(),
+            'path' => $cookie->getPath(),
+            'domain' => $cookie->getDomain(),
+            'secure' => $cookie->isSecure(),
+            'httponly' => $cookie->isHttpOnly(),
+            'expires' => $cookie->getExpiresTimestamp()
+        ];
     }
 }
 
--- a/src/Http/Client/CookieCollection.php
+++ b/src/Http/Client/CookieCollection.php
@@ -70,7 +70,7 @@ class CookieCollection extends BaseCollection
     }
 
     /**
-     * Get all the stored cookies.
+     * Get all the stored cookies as arrays.
      *
      * @return array
      */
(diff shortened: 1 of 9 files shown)


src/Http/Client/Response.php
```diff
@@ -442,7 +444,39 @@ public function getCookieData($name)
...
+     * expects date strings instead of timestamps.
+     *
+     * @param \Cake\Http\Cookie\CookieInterface $cookie Cookie object.
+     * @return array
+     */
+    protected function toArrayClient(CookieInterface $cookie)
+    {
+        if ($cookie instanceof Cookie) {
+            return $cookie->toArrayClient();
+        } elseif ($cookie->getExpiry()) {
```
dereuromark: No Else needed.

src/Http/Client/Response.php
```diff
@@ -442,7 +444,41 @@ public function getCookieData($name)
...
+     * expects date strings instead of timestamps.
+     *
+     * @param \Cake\Http\Cookie\CookieInterface $cookie Cookie object.
+     * @return array
+     */
+    protected function toArrayClient(CookieInterface $cookie)
+    {
+        if ($cookie instanceof Cookie) {
+            return $cookie->toArrayClient();
+        }
```
markstory: Why not remove this method from the `Cookie` object? Having two implementations seems pointless.

src/Http/Response.php
```diff
@@ -2056,13 +2061,45 @@ public function getCookies()
...
+     * where `httponly` is `httpOnly` and `expires` is `expire`
+     *
+     * @param \Cake\Http\Cookie\CookieInterface $cookie Cookie object.
+     * @return array
+     */
+    protected function toArrayResponse(CookieInterface $cookie)
+    {
+        if ($cookie instanceof Cookie) {
+            return $cookie->toArrayResponse();
+        }
```
markstory: Same as above. It looks like the compat methods could be removed from the `Cookie` object and put here instead.

src/Http/Client/Response.php
```diff
@@ -442,7 +444,41 @@ public function getCookieData($name)
...
+
+        if ($cookie->getExpiry()) {
+            $expires = $cookie->getExpiry()->format(Cookie::EXPIRES_FORMAT);
```
markstory: Should there be interface methods for getting the timestamp and formatted time?

robertpustulka: @markstory I'm not sure. The cookie returns `DateTimeInterface` so you can format it anyway you like.

src/Http/Client/Response.php
```diff
@@ -442,7 +444,31 @@ public function getCookieData($name)
...
+    /**
+     * Convert the cookie into an array of its properties.
+     *
+     * This method is compatible with older client code that
+     * expects date strings instead of timestamps.
+     *
+     * @param \Cake\Http\Cookie\CookieInterface $cookie Cookie object.
+     * @return array
+     */
+    public function convertCookie(CookieInterface $cookie)
```
markstory: Does this need to be a public function? It seems like it could be protected. Long term I want the array format of cookies to go away.

src/Http/Response.php
```diff
@@ -2056,13 +2061,35 @@ public function getCookies()
...
     /**
+     * Convert the cookie into an array of its properties.
+     *
+     * This method is compatible with the historical behavior of Cake\Http\Response,
+     * where `httponly` is `httpOnly` and `expires` is `expire`
+     *
+     * @param \Cake\Http\Cookie\CookieInterface $cookie Cookie object.
+     * @return array
+     */
+    public function convertCookie(CookieInterface $cookie)
```
markstory: Does this need to be a public method? It could be a protected instead.

src/Http/Client/Response.php
robertpustulka: I made them public so I could test them easily ;). But right, that doesn't need to be piblic.

markstory: These methods are implicitly tested by the response continuing to work as it did before. :smile:

ravage84: ``convertCookie()``  is a very general name.

Why not call it ``convertToArray()`` or ``ConvertToProerptiesArray()``

> **Convert** the **cookie** into an **array** of its **properties**.

src/Http/Client/CookieCollection.php
```diff
@@ -70,19 +71,40 @@ public function get($url)
...
+
+    /**
+     * Convert the cookie into an array of its properties.
+     *
+     * Primarily useful where backwards compatibility is needed.
+     *
+     * @param \Cake\Http\Cookie\CookieInterface $cookie Cookie object.
+     * @return array
+     */
+    public function convertCookie(CookieInterface $cookie)
```
ravage84: Same as https://github.com/cakephp/cakephp/pull/10567/files#diff-6d806173a45816b0375aaaece3a987a7R461

src/Http/Response.php
ravage84: Same as https://github.com/cakephp/cakephp/pull/10567/files#diff-6d806173a45816b0375aaaece3a987a7R461

PR #10569, opened 2017-04-20
Add Query::valueBinder() getter

cleptric:
Continues with #9978


src/Database/Query.php
```diff
@@ -1801,19 +1801,38 @@ public function traverseExpressions(callable $callback)
...
 
         return $this;
     }
 
     /**
+     * Returns the currently used ValueBinder instance.
+     *
+     * A ValueBinder is responsible for generating query placeholders and temporarily
+     * associate values to those placeholders so that they can be passed correctly
+     * statement object.
```
markstory: Should be 'passed correctly to the statement object'.

ravage84 (approved):

PR #10570, opened 2017-04-20
Split Debugger::outputAs() into getter/setter

cleptric:
Continues with #9978

### DIFF
--- a/src/Error/Debugger.php
+++ b/src/Error/Debugger.php
@@ -651,8 +651,40 @@ class Debugger
     }
 
     /**
+     * Get the output format for Debugger error rendering.
+     *
+     * @return string Returns the current format when getting.
+     */
+    public static function getOutputAs()
+    {
+        $self = Debugger::getInstance();
+
+        return $self->_outputFormat;
+    }
+
+    /**
+     * Set the output format for Debugger error rendering.
+     *
+     * @param string $format The format you want errors to be output as.
+     * @return null
+     * @throws \InvalidArgumentException When choosing a format that doesn't exist.
+     */
+    public static function setOutputAs($format)
+    {
+        $self = Debugger::getInstance();
+
+        if (!isset($self->_templates[$format])) {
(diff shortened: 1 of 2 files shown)


src/Error/Debugger.php
```diff
@@ -651,8 +651,40 @@ protected static function _object($var, $depth, $indent)
...
+        $self = Debugger::getInstance();
+
+        return $self->_outputFormat;
+    }
+
+    /**
+     * Set the output format for Debugger error rendering.
+     *
+     * @param string $format The format you want errors to be output as.
+     * @return null
```
dereuromark: should be void :)

src/Error/Debugger.php
```diff
@@ -651,8 +651,40 @@ protected static function _object($var, $depth, $indent)
...
+    public static function setOutputAs($format)
+    {
+        $self = Debugger::getInstance();
+
+        if (!isset($self->_templates[$format])) {
+            throw new InvalidArgumentException('Invalid Debugger output format.');
+        }
+        $self->_outputFormat = $format;
+
+        return null;
```
dereuromark: this is not needed then.

src/Error/Debugger.php
```diff
@@ -651,8 +651,38 @@ protected static function _object($var, $depth, $indent)
...
     /**
+     * Get the output format for Debugger error rendering.
+     *
+     * @return string Returns the current format when getting.
+     */
+    public static function getOutputAs()
+    {
+        $self = Debugger::getInstance();
+
+        return $self->_outputFormat;
```
markstory: Couldn't this be a one liner?

ravage84: Why not ``getOutputFormat()``?

> Get the **output format** for Debugger error rendering.

cleptric: I'll do a followup. It should be also be `setOutputFormat()` then

PR #10571, opened 2017-04-21
Add a mthod for 1 job and delete duplicate code

mehdi-fathi:
file Session.php has some duplicated code.for example : `$this->_hasSession() && !$this->started()` there is on 3 method.It can be on a method for doing a job.


ADmad:
Thanks but adding an extra method call for such a simple check is not necessary.

PR #10572, opened 2017-04-21
Split Registry/Locator methods into getters and setters.

robertpustulka:
Refs #9978

### DIFF
--- a/src/ORM/TableRegistry.php
+++ b/src/ORM/TableRegistry.php
@@ -69,18 +69,40 @@ class TableRegistry
      *
      * @param \Cake\ORM\Locator\LocatorInterface|null $locator Instance of a locator to use.
      * @return \Cake\ORM\Locator\LocatorInterface
+     * @deprecated 3.5.0 Use getter/setter instead.
      */
     public static function locator(LocatorInterface $locator = null)
     {
         if ($locator) {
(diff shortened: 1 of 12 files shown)


src/ORM/TableRegistry.php
```diff
@@ -69,13 +69,24 @@ class TableRegistry
      *
      * @param \Cake\ORM\Locator\LocatorInterface|null $locator Instance of a locator to use.
      * @return \Cake\ORM\Locator\LocatorInterface
+     * @deprecated 3.5.0 Use getter/setter instead.
```
markstory: Can we name the methods here.

markstory (changes requested):

src/ORM/TableRegistry.php
```diff
@@ -69,13 +69,24 @@ class TableRegistry
...
 
+        return static::getTableLocator();
+    }
+
+    /**
+     * Returns a singleton instance of LocatorInterface implementation.
+     *
+     * @return \Cake\ORM\Locator\LocatorInterface
+     */
+    public static function getTableLocator()
```
saeideng: What not getLocator()

robertpustulka: For consistency with trait method naming.

PR #10573, opened 2017-04-21
Split EventDispatcherTrait::eventManager() into getter/setter.

robertpustulka:
Refs #9978


src/Event/EventDispatcherTrait.php
```diff
@@ -43,19 +43,51 @@
      *
      * @param \Cake\Event\EventManager|null $eventManager the eventManager to set
      * @return \Cake\Event\EventManager
+     * @deprecated 3.5.0 Use getter/setter instead.
```
markstory: Can we name the specific methods here? It makes upgrading easier for folks as they have the correct method in their IDEs

PR #10574, opened 2017-04-21
Allow camelBacked method names for subcommands.

dereuromark:
This is more a feature than a bug, but the expectation sure is that the internal shell "actions" named "commands" work similar to the controller actions regarding their internal mapping.

In controllers it is `'action' => 'myCoolAction'` and does only inflect for the outside (via routing and dashed inflection).
The same expectation should be met on the CLI level when adding subcommands
```php
$parser->addSubcommand('initMyDb', [...]);
```

Those subcommands are basically the methods of this class just as the methods of a controller are its actions.
Whereas from the outside (CLI) you can then either access it via underscored or CamelCase as usual.

In 4.0 we could probably make this a bit stricter then, allowing one internal format (methodName) only.

PS: Is this safe to go into master or is 3.5 better, as people could (wrongly?) use the uninflected/raw camelCase format already in their apps.

### DIFF
--- a/src/Console/ConsoleOptionParser.php
+++ b/src/Console/ConsoleOptionParser.php
@@ -576,6 +576,7 @@ class ConsoleOptionParser
             $command = $name;
             $name = $command->name();
         } else {
+            $name = Inflector::underscore($name);
             $defaults = [
                 'name' => $name,
                 'help' => '',
--- a/src/Console/ConsoleOptionParser.php
+++ b/src/Console/ConsoleOptionParser.php
@@ -665,7 +665,7 @@ class ConsoleOptionParser
      */
     public function parse($argv)
     {
-        $command = isset($argv[0]) ? $argv[0] : null;
+        $command = isset($argv[0]) ? Inflector::underscore($argv[0]) : null;
         if (isset($this->_subcommands[$command])) {
             array_shift($argv);
         }
--- a/src/Console/Shell.php
+++ b/src/Console/Shell.php
@@ -440,7 +440,7 @@ class Shell
      */
     public function runCommand($argv, $autoMethod = false, $extra = [])
     {
-        $command = isset($argv[0]) ? $argv[0] : null;
+        $command = isset($argv[0]) ? Inflector::underscore($argv[0]) : null;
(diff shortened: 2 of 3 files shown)


src/Console/ConsoleOptionParser.php
```diff
@@ -576,6 +576,7 @@ public function addSubcommand($name, array $options = [])
             $command = $name;
             $name = $command->name();
         } else {
+            $name = Inflector::underscore($name);
```
markstory: Will `cake things initMyDb` still work? It looks like that will break which could cause a bunch of havoc for folks with cron and deploy automation.

dereuromark: Yeah, that was the concern I had above. I think I have an idea on how to keep things here rather BC.

dereuromark:
I switched the target to 3.next just to be on the safe side here.
After merging master to 3.next the diff should be clean again then.

markstory:
I've rebased these changes ontop of 3.next.

src/Console/ConsoleOptionParser.php
markstory: Next goofy scenario I would be worried about is. If both `initMyDb` and init_my_db` are proper sub-commands. Right now only one will be callable, but previously both would be. Is that perhaps a scenario that should be rare enough to not worry about? Should we trigger an error/exception when duplicate methods are attached?

dereuromark: I would consider this definitely a misconfig, having both my_method and myMethod as methods + console option config seems nonsense to me.
You either use the underscore or camelCase - and ideally consistently :)

This was part of the reason I think it is better to communicate this in the 3.5 migration guide.

markstory: Ok. Something to take note of in the migration guide then.

dereuromark: Looking into the code a bit more it looks like it is not possible to have multiple methods set up, the method itself always had to be camelCase:

    $method = Inflector::camelize($command);

So this is mainly for making the calling and setup part more consistent.

dereuromark:
Thx, looking good now.

PR #10576, opened 2017-04-24
Adding a getter and setter for cache registry

burzum:


Theaxiom (approved):

src/Cache/Cache.php
```diff
@@ -108,22 +108,45 @@ class Cache
...
+        }
+
+        return static::$_registry;
+    }
+
+    /**
+     * Returns the Cache Registry instance used for creating and using cache adapters.
+     * Also allows for injecting of a new registry instance.
+     *
+     * @param \Cake\Core\ObjectRegistry|null $registry Injectable registry object.
```
cleptric: `null` is obsolete now

src/Cache/Cache.php
```diff
@@ -108,22 +108,45 @@ class Cache
...
+    public static function setRegistry(ObjectRegistry $registry)
+    {
+        static::$_registry = $registry;
+    }
+
+    /**
+     * Returns the Cache Registry instance used for creating and using cache adapters.
      * Also allows for injecting of a new registry instance.
      *
+     * @deprecated Use getRegistry() and setRegistry() instead.
```
cleptric: Could we add the version number?

burzum: OK, but why do you need it? I don't think *any* other deprecated method has the version number included.

cleptric: That's not true. A quick search for `@deprecated` inside the core reveals that basically all methods have a version number attached to the annotation.
Main reason here is to be consistent :)

Theaxiom: If that's the case, we should also raise an issue against any deprecation notices which omit the deprecated version number.

burzum: Yes, I'm not 100% sure but I haven't seen any version in these lines before. Were they added recently?

cleptric: We normally just add them when we deprecate stuff.

@Theaxiom There are just a few methods that missing the version. I guess it's a little bit time consuming to scrape through the history to find the right version. A PR addressing this is of course welcomed at any time 😀

src/Cache/Cache.php
```diff
@@ -108,22 +108,45 @@ class Cache
...
+    {
+        if (!static::$_registry) {
+            static::$_registry = new CacheRegistry();
+        }
+
+        return static::$_registry;
+    }
+
+    /**
+     * Returns the Cache Registry instance used for creating and using cache adapters.
```
markstory: But it doesn't return anything..

markstory (changes requested):
Would be good to have some tests for the new methods.

src/Cache/Cache.php
ravage84: Probably Copy Pasta error.

src/Cache/Cache.php
```diff
@@ -108,22 +108,45 @@ class Cache
...
+    public static function setRegistry(ObjectRegistry $registry)
+    {
+        static::$_registry = $registry;
+    }
+
+    /**
+     * Returns the Cache Registry instance used for creating and using cache adapters.
+     * Also allows for injecting of a new registry instance.
+     *
+     * @deprecated Deprecated since 3.5. Use getRegistry() and setRegistry() instead.
```
ravage84: Move below return.

ravage84 (changes requested):

burzum:
@markstory I've added tests, please review again.

src/Cache/Cache.php
```diff
@@ -165,16 +189,31 @@ protected static function _buildEngine($name)
...
+    /**
+     * Fetch the engine attached to a specific configuration name.
+     *
+     * If the cache engine & configuration are missing an error will be
+     * triggered.
+     *
+     * @param string $config The configuration name you want an engine for.
+     * @return \Cake\Cache\CacheEngine When caching is disabled a null engine will be returned.
+     */
+    public static function getEngine($config)
```
saeideng: we have not setter method for `Engine() ` , so we no need to `getEngine()`

markstory: Agreed.

PR #10577, opened 2017-04-24
Getters and setters for I18n/I18n.php

burzum:


src/I18n/I18n.php
```diff
@@ -206,22 +206,45 @@ public static function config($name, callable $loader)
      * When called with no arguments it will return the currently configure
      * locale as stored in the `intl.default_locale` PHP setting.
      *
+     * @deprecated Deprecated since 3.5. Use setLocale() and getLocale().
```
dereuromark: You dont have to repeat the deprecated part :)
The consistent way across the newer deprecations so far always was:

    @deprecated 3.4.0 Use setTypes()/getTypes() instead.

src/I18n/I18n.php
```diff
@@ -125,16 +126,71 @@ public static function translators()
...
+     * );
+     * ```
+     *
+     * @param string $name The domain of the translation messages.
+     * @param string|null $locale The locale for the translator.
+     * @param callable|null $loader A callback function or callable class responsible for
+     * constructing a translations package instance.
+     * @return void
+     */
+    public static function setTranslator($name = 'default', $locale = null, callable $loader)
```
dereuromark: If loader is not a nullable (optional) argument anymore, maybe the order should be changed here for the setter?

burzum: The problem is it seems that the default name of value `default` is expected. I was looking into changing it to `($name, callable $loader, $locale = null)`.

If we're OK with that I'll do it. BC can be maintained by translator() by passing `default` as it is.

src/I18n/I18n.php
```diff
@@ -125,16 +126,71 @@ public static function translators()
...
+    }
+
+    /**
+     * Configures future translators, this is achieved by passing a callable
+     * as the last argument of this function.
+     *
+     * ### Example:
+     *
+     * ```
+     *  I18n::translator('default', 'fr_FR', function () {
```
dereuromark: they seem outdated now.

src/I18n/I18n.php
```diff
@@ -125,16 +126,71 @@ public static function translators()
...
+     *
+     * You can also use the `Cake\I18n\MessagesFileLoader` class to load a specific
+     * file from a folder. For example for loading a `my_translations.po` file from
+     * the `src/Locale/custom` folder, you would do:
+     *
+     * ```
+     * I18n::translator(
+     *  'default',
+     *  'fr_FR',
+     *  new MessagesFileLoader('my_translations', 'custom', 'po');
```
markstory: This example doesn't match the parameter order in the function signature.

markstory (changes requested):
It would be good to get tests for the newly added methods.

src/I18n/I18n.php
```diff
@@ -263,6 +356,29 @@ public static function defaultFormatter($name = null)
...
+     * @return string The name of the formatter.
+     */
+    public static function getDefaultFormatter()
+    {
+        return static::translators()->defaultFormatter();
+    }
+
+    /**
+     * Sets the name of the default messages formatter to use for future
+     * translator instances. By default the `default` and `sprintf` formatters
```
ravage84: Can be slit up in short and long description.

src/I18n/I18n.php
```diff
@@ -236,10 +315,23 @@ public static function locale($locale = null)
...
      * @return string
      */
     public static function defaultLocale()
     {
+        return static::getDefaultLocale();
+    }
+
+    /**
+     * This returns the default locale before any modifications, i.e.
+     * the value as stored in the `intl.default_locale` PHP setting before
```
ravage84: Can be slit up in short and long description.

src/I18n/I18n.php
```diff
@@ -206,22 +262,45 @@ public static function config($name, callable $loader)
...
 
             return null;
         }
 
+        return self::getLocale();
+    }
+
+    /**
+     * Sets the default locale to use for future translator instances.
+     * This also affects the `intl.default_locale` PHP setting.
```
ravage84: Can be slit up in short and long description.

src/I18n/I18n.php
```diff
@@ -125,16 +126,71 @@ public static function translators()
...
+
+        $loader = static::translators()->setLoaderFallback($name, $loader);
+
+        $packages = static::translators()->getPackages();
+        $packages->set($name, $locale, $loader);
+    }
+
+    /**
+     * Returns an instance of a translator that was configured for the name and passed
+     * locale. If no locale is passed then it takes the value returned by the `getLocale()` method.
```
ravage84: Can be slit up in short and long description.

src/I18n/I18n.php
```diff
@@ -125,16 +126,73 @@ public static function translators()
...
         }
 
+        return self::getTranslator($name, $locale);
+    }
+
+    /**
+     * Sets a translator.
+     *
+     * Configures future translators, this is achieved by passing a callable
+     * as the last argument of this function.
```
markstory: Callable isn't the last argument though.

PR #10582, opened 2017-04-24
Write Some Assertions for Validation Library

sohelrana820:


tests/TestCase/Validation/ValidationTest.php
```diff
@@ -2813,6 +2817,9 @@ public function testContainNonAlphaNumeric()
         $this->assertFalse(Validation::containsNonAlphaNumeric('abcdef##', 3));
         $this->assertFalse(Validation::containsNonAlphaNumeric('abc##def', 3));
         $this->assertFalse(Validation::containsNonAlphaNumeric('ab#cd#ef', 3));
+
+        //CHECK SCALAR
```
markstory: Why are these comments in ALL CAPS?

sohelrana820: No special reason. Should I make it capitalize or lowercase?

dereuromark: Do we need them at all?

sohelrana820: I think so. Cause someone might be tried with Scalar value (CC or AlphaNumeric).

dereuromark: But arent you doing the opposite here? Checking a non-scalar (array) value?
Usually one can make a separate test method with the method name making it clear, also optionally adding a doc block note there.

sohelrana820: So, what do you suggest here?

dereuromark: testContainNonAlphaNumericWithArray() etc for example.

sohelrana820: OK... Please do me a favor. I will do something like that you said. So should I send another pull request after modify it? And If so then what about my current pull request?

dereuromark: You can add commits to your branch. This PR will automatically be updated :)

tests/TestCase/Validation/ValidationTest.php
```diff
@@ -3104,4 +3099,10 @@ public function testImageWidth()
         $this->assertFalse(Validation::imageWidth($upload, '>', 2000));
         $this->assertFalse(Validation::imageWidth($upload, '==', 3000));
     }
+
+    public function testContainNonAlphaNumericAndCCWithScalar()
```
dereuromark: Why scalar? Those are non scalar values - also the 2nd one is not the same as in the previous commit (it now is a scalar).

tests/TestCase/Validation/ValidationTest.php
```diff
@@ -168,6 +168,7 @@ public function testLengthBetween()
 
         $this->assertFalse(Validation::lengthBetween('abcdefg', 1, 6));
         $this->assertFalse(Validation::lengthBetween('ÆΔΩЖÇ', 1, 3));
+        $this->assertFalse(Validation::lengthBetween(1, 1, 3));
```
dereuromark: I feel like this should be true actually - is there maybe a cast missing? Or are we ok with the string input only for it?

tests/TestCase/Validation/ValidationTest.php
```diff
@@ -3097,4 +3099,10 @@ public function testImageWidth()
         $this->assertFalse(Validation::imageWidth($upload, '>', 2000));
         $this->assertFalse(Validation::imageWidth($upload, '==', 3000));
     }
+
+    public function testContainNonAlphaNumericAndCCWithScalar()
+    {
+        $this->assertFalse(Validation::cc(['869972521242198'], ['voyager']));
+        $this->assertFalse(Validation::containsNonAlphaNumeric('ab#cd#ef', 3));
```
markstory: Why is this here vs. with the other tests for this method?

tests/TestCase/Validation/ValidationTest.php
sohelrana820: I will fix it.

tests/TestCase/Validation/ValidationTest.php
sohelrana820: No casting issue. It's just take string input.

tests/TestCase/Validation/ValidationTest.php
sohelrana820: So...What do you suggest?

markstory: These tests could be added to the existing test methods.

sohelrana820: OK... I will modify it.

dereuromark:
I am still not sure any of the tests brings value. They are either covered or completely out of scope (passing an array never works by definition).

markstory:
Thanks @sohelrana820

Issue #10308, opened 2017-02-26
RFC - Middleware 'pipelines' as part of Routing


markstory:
This is a (multiple allowed):

* [ ] bug
* [x] enhancement
* [x] feature-discussion (RFC)

* CakePHP Version: 3.5.0

PSR-7 Middleware makes it simple to compose application logic together as separate layers that are composed together wrapping an application. What is lacking is a way to apply middleware to only specific routes. When this was last brought up @lorenzo  wanted middleware integrated into routing. I originally thought this might be really hard, and it still might be, but here is a rough idea of what the API could be:

* The `Cake\Routing\Middleware\RoutingMiddleware` class would take an `Application` as a parameter. If that parameter is provided, the middleware will invoke the `routing` hook method on the application.
* The `Application::routing()` method would provide an object wrapper for `config/routes.php`  like `Application::bootstrap()` provides for `config/bootstrap.php`. Inside this method routes could be connected, much as they are today.

By default the `routing` hook would simply include `config/routes.php`, but it could be overridden as well.

An example of the routing hook and the routing middleware in action would be:

```php
function routing($routes)
{
    // Register middleware, for use in routing scopes
    $routes->registerMiddleware('auth', new AuthenticationMiddleware(...))
        ->registerMiddleware('cookies', new EncryptedCookieMiddleware(...))
        ->registerMiddleware('csrf', new CsrfMiddleware(...));
        ->registerMiddleware('cors', new CorsMiddleware(...));

    // Create a routing scope (as you do today)
    $routes->scope('/api', function ($routes) {
        // Enable middleware
        $routes->middleware('csrf', 'cors', 'auth');
        $routes->connect('/ping', ['controller' => 'Pings']);
    });

    $routes->scope('/bookmarks', function ($routes) {
        // Enable middleware
        $routes->middleware('csrf', 'cookies', 'auth');
        $routes->connect('/bookmarks/:action', ['controller' => 'Bookmarks']);
    });
}
```

## Detailed Steps

* `RouteBuilder` would be updated to have new methods for registering middleware into the `RouteCollection`.
* After routes are matched, `RoutingMiddleware` could fetch the middleware for a given URL similar to how we do scope matching today. `$next = Router::getMiddlewareForRequest($request, $next);`
* Middleware would be matched by the registered scopes.
* Once found, middleware objects would be added to a new `MiddlewareQueue` object.
[...]

PR #10583, opened 2017-04-25
3.next - Enable scoped middleware to be setup

markstory:
Add the methods that help developers register and configure scoped middleware for their applications. The next task will be to use these new features when dispatching requests.

Refs #10308

### DIFF
--- a/src/Routing/RouteBuilder.php
+++ b/src/Routing/RouteBuilder.php
@@ -714,4 +714,32 @@ class RouteBuilder
         $this->connect('/:controller', ['action' => 'index'], compact('routeClass'));
         $this->connect('/:controller/:action/*', [], compact('routeClass'));
     }
+
+    /**
+     * Register a middleware with the RouteCollection.
+     *
+     * Once middleware has been registered, it can be applied to the current routing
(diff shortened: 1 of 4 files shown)


src/Routing/RouteBuilder.php
```diff
@@ -714,4 +714,38 @@ public function fallbacks($routeClass = null)
...
+    /**
+     * Apply a middleware to the current route scope.
+     *
+     * Requires middleware to be registered via `registerMiddleware()`
+     *
+     * @param string[] ...$names The names of the middleware to apply to the current scope.
+     * @return $this
+     * @see \Cake\Routing\RouteCollection::addMiddlewareToScope()
+     */
+    public function middleware(...$names)
```
cleptric: Maybe `applyMiddleware()` or `enableMiddleware()` to better display what the method does?

markstory: I like `applyMiddleware()`.

cleptric: So mote it be 😀

markstory:
Merging this so I can do the next step.

Issue #10586, opened 2017-04-26
Patching should convert datetime strings


dereuromark:
This is a (multiple allowed):

* [] bug
* [x] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.4.5
* MySQL 5.7

### What you did
A basic JSON export and then import on a different app/DB:
```
                $json = json_decode($content, true);
                ...
                foreach ($json as $row) {
                    unset($row['id']);
                    //$row['created'] is '2017-04-05T17:18:00+00:00',
                    //$row['modified'] is '2017-04-05T17:18:00+00:00',
                    $release = $this->Releases->newEntity($row);

                    $this->Releases->saveOrFail($release);
                }
```

### What happened
```
[PDOException] SQLSTATE[22007]: Invalid datetime format: 
1292 Incorrect datetime value: '2017-04-05T17:18:00+00:00' for column 'created' at row 1
```

After marshalling the JSON ISO format strings are still strings not have not been converted to Time() objects that the ORM can safely handle for saving.

### What you expected to happen
Saving just fine - including the JSON datetime string.

Maybe the patching should check the fields based on the schema, and similar to array=>object for date/time, should also transform ISO strings to Time() objects?
That would avoid manual transformation or unnecessary exceptions on save.

Any takes on where this would best be done?
I can then take a stab at it when I find some time.

ADmad:
The `DateTimeType` db type class already handles converting strings to objects when marshalling, if the datetime string is in the format the type class is configured for. If you haven't turned on locale based parsing then the default format is `Y-m-d H:i:s`.

dereuromark:
I dont think the JSON ISO code counts as localized - it is as valid as Y-m-d H:i:s directly IMO.

dereuromark:
I think prematurely setting tickets on hold is also not too nice - I must say.
You should at least wait for some more feedback before doing so.

Could the _format be an array of acceptable ISO standard formats?
```php
    /**
     * String format to use for DateTime parsing
     *
     * @var string|array
     */
    protected $_format = [
        'Y-m-d H:i:s',
        ...
    ];
```

Refs https://en.wikipedia.org/wiki/ISO_8601

dereuromark:
Another idea:
```php
// Normalize ISO code
if (preg_match('^/[\d]{2}-[\d]{2}-[\d]{2}T[\d]{2}:[\d]{2}:[\d]{2}[+-][\d]{2}:[\d]{2}$/') {
[...]

Issue #10439, opened 2017-03-21
3.x DATETIME(6) support


inoas:
This is a (multiple allowed):

* [x] enhancement
* [x] feature-discussion (RFC)

* CakePHP Version: 3.4.3
* Platform and Target: MariaDB 10.1, PHP 7.1.2

### What you did
```sql
CREATE TABLE `cms_documents` (
  `id` char(36) CHARACTER SET ascii NOT NULL,
  `revision_id` char(36) CHARACTER SET ascii NOT NULL,
  `created` datetime(6) NOT NULL,
  `is_published` tinyint(1) unsigned DEFAULT NULL,
  PRIMARY KEY (`id`,`revision_id`),
  UNIQUE KEY `uniq_create` (`id`,`created`),
  UNIQUE KEY `published_revision` (`is_published`,`id`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8;
```

### What happened
Even if I manually adapt the datetimes with microseconds I still get back a regular datetime.
- E.g. database: 2017-03-21 16:52:49.123123
- Returns: 2017-03-21 16:52:49

### What you expected to happen
- To get back 2017-03-21 16:52:49.123123
- Also to be able to save 2017-03-21 16:52:49.123123 - does Chronos support that anyway?

dakota:
What version as of PHP are you running? Afaik, only 7.1 supports microseconds in Date objects.

inoas:
I am running `PHP 7.1.2-4+deb.sury.org~trusty+1 (cli) (built: Mar  2 2017 13:23:06) ( NTS )`

inoas:
Console:
```php
> var_dump((new \Datetime)->format('Y-m-d\TH:i:s.u'));
string(26) "2017-03-21T17:13:54.093858"
```

In `beforeSave()`
```php
        $entity->set('created', \Cake\I18n\Time::createFromFormat(
            'Y-m-d\TH:i:s.u',
            (new \Datetime)->format('Y-m-d\TH:i:s.u'),
            'America/New_York'
        ));

... but microseconds are not being persisted to the database (instead it saves 2017-03-21T17:13:54.000000)

markstory:
I suspect that the Datetime Type objects are eliding the microseconds when converting Time instances into strings to be inserted into the db.

inoas:
I have checked `DateTimeType:toDatabase()`: `return $value->format($this->_format);`

[...]

PR #10587, opened 2017-04-26
Allow extended ISO datetime format to be marshalled.

dereuromark:
Implements https://github.com/cakephp/cakephp/issues/10586

A first stab at the issue.
I would only recommend adding the single one ISO format `'Y-m-d\TH:i:s.uO'` which CakePHP generates on output in JSON files, so when doing the reverse, import, it can safely and directly convert back into the DB format.

Any better idea on the subject?


src/Database/Type/DateTimeType.php
```diff
@@ -53,9 +53,12 @@ class DateTimeType extends Type implements TypeInterface
     /**
      * String format to use for DateTime parsing
      *
-     * @var string
+     * @var string|array
      */
-    protected $_format = 'Y-m-d H:i:s';
+    protected $_format = [
+        'Y-m-d H:i:s',
+        'Y-m-d\TH:i:s.uO',
```
markstory: This will change the accepted data of every application using CakePHP. Not sure that is something we should be doing in a bugfix release.

dereuromark: It should be fully BC as it only adds on top. But sure, we can also either make it opt-in from app level, or do it in a minor.
Maybe there are also even better options here than this.

markstory: My concern is more that this adds a new valid input form that was not previously accepted. Developers may be surprised when previously invalid data becomes valid. Also concerning is that the new format includes timezone information which will require apps to ensure they are doing timezone conversions.

darxmac: Actually i ran into this issue because we started testing with MySQL 5.7, and the same app/db on MySQL 5.6 this "just works" (posting json from a js frontend to a controller)

dereuromark: Yes that was my Pain Point too.
5.6 works fine, 5.7 crashes :/

markstory: Are you saying that 5.7 fails to insert rows with​ the current datetime format marshalling?

darxmac: MySQL 5.6:

```sql
mysql> insert into `products` set `created` = '2017-05-02T08:30:44+00:00';
Query OK, 1 row affected, 1 warning (0.00 sec)
```

MySQL 5.7:

```sql
mysql> insert into `products` set `created` = '2017-05-02T08:30:44+00:00';
ERROR 1292 (22007): Incorrect datetime value: '2017-05-02T08:30:44+00:00' for column 'created' at row 1
```

markstory: Ok. So having the ISO formats marshalled into datetime objects would likely help with that.

src/Database/Type/DateTimeType.php
```diff
@@ -165,7 +169,7 @@ public function marshal($value)
                 $date = new $class($value);
                 $compare = true;
             }
-            if ($compare && $date && $date->format($this->_format) !== $value) {
+            if ($compare && $date && !$this->_compare($date, $value)) {
```
chinpei215: Do we really need to check whether the date string can be restored here? I think this kind of check should be done by validation. If we could remove this check, people could use `<input type="datetime-local">` fields, that uses another date string formats, such as `Y-m-d\TH:i` or  `Y-m-d\TH:i:s`.
Also, I think we have to change `Validation::datetime()` too. At the moment, it doesn't support for ISO datetime strings.

dereuromark: Do you have an alternative PR for your recommendation? I am open to any other approach that would fix the issue.

chinpei215: If we could get a consensus, I'll do that. Now I have a tricky way to pass the validation in my 2.x project:
```php
public function isDateTime($check) {
    $value = array_values($check)[0];
    $value = implode(' ', explode('T', $value, 2));
    return Validation::datetime($value);
}
```
But it would be nice if Cake supports modern HTML features without this kind of trick.

markstory: This check was originally intended to ensure that an ambiguous input format wasn't used resulting in a different datetime value than was in the request data. Perhaps that isn't useful, an we should instead rely on validation and app logic.

chinpei215: OK. Then I'll send an alternative PR to 3.next when I have time. If it becomes to accept new date formats, the master branch would not be suitable.

darxmac:
what are the chances this could get looked at again ? anything i can do ?

dereuromark:
@darxmac Do you have an idea of implementation that would work better than this one? It seems the compare part here causes some issues. The tests do not pass.

darxmac:
Looks like the format string was off a bit:

https://github.com/cakephp/cakephp/compare/master-iso-datetime...darxmac:master-iso-datetime

src/Database/Type/DateTimeType.php
dereuromark: @chinpei215 I consider this more a bugfix, because so far this kind of ISO date was wrongly not supported.
Are there any reasonable downsides of those "valid" date strings to be not turned down? Just saying "because so far wasnt allowed" is not a valid reason. Those already get exported like this for years, they must also be able get imported the same way.

dereuromark:
Looks like the tests are now green.

src/Database/Type/DateTimeType.php
chinpei215: @dereuromark Changing string variable to array is not backward compatible. If someone uses a sub-class using `_format` property directly, the application would break. So this change must be done in 3.next.

dereuromark: Fair Point.

markstory:
Moved to 3.next as this is a bigger change than I think we should be putting into a bug fix release.

dereuromark:
@lorenzo What do you think?

dereuromark:
bump

darxmac:
anything else i could do to help move this along ?

src/Database/Type/DateTimeType.php
```diff
@@ -206,6 +211,22 @@ public function marshal($value)
     }
 
     /**
+     * @param \Cake\I18n\Time|\DateTime $date DateTime object
+     * @param mixed $value Request data
+     * @return bool
+     */
+    protected function _compare($date, $value)
+    {
+        foreach ((array)$this->_format as $format) {
```
markstory: Do the other subclasses need any changes?

markstory:
@darxmac The earliest this will ship is 3.5.0 which hasn't even started beta/rc releases, so it will be a while even if this is merged today.

src/Database/Type/DateTimeType.php
dereuromark: Not that I could see.

markstory: ok.

src/Database/Type/DateTimeType.php
```diff
@@ -53,9 +53,12 @@ class DateTimeType extends Type implements TypeInterface
     /**
      * String format to use for DateTime parsing
      *
-     * @var string
+     * @var string|array
      */
-    protected $_format = 'Y-m-d H:i:s';
+    protected $_format = [
+        'Y-m-d H:i:s',
+        'Y-m-d\TH:i:sP',
```
inoas: Could we add `Y-m-d\TH:i:s.u'`

That would allow support for `datetime(6)` see: https://github.com/cakephp/cakephp/issues/10439

dereuromark: Sure. You want to make a PR against this branch?
Not sure when I will be able to follow up here.

dereuromark:
We could also already merge this and further enhance with future on top PRs?

PR #10589, opened 2017-04-27
Fix error message for named route matching

berarma:
When a named route doesn't match, the error message contains "%s" instead of the named route. That's my guess.

### DIFF
--- a/src/Routing/RouteCollection.php
+++ b/src/Routing/RouteCollection.php
@@ -298,7 +298,7 @@ class RouteCollection
                 throw new MissingRouteException([
                     'url' => $name,
                     'context' => $context,
-                    'message' => 'A named route was found for "%s", but matching failed.',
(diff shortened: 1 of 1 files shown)


dereuromark (approved):

ADmad:
Fix looks good. Can you please also add a test case for it.

cleptric:
`\Cake\Core\Exception\Exception` does a `vsprintf` if message is passed as array.

```
        throw new \Cake\Routing\Exception\MissingRouteException([
            'url' => 'Foo',
            'context' => 'Context',
            'message' => 'A named route was found for "%s", but matching failed.',
        ]);
```

screen shot 2017-04-27 at 14 17 55

berarma:
Alright @cleptric. It was because of an error in my app that it didn't work like it should. My ErrorController class wasn't properly extending the Cake's class. Sorry for my mistake.

PR #10591, opened 2017-04-27
3.x - Unify conjugation order

ndm2:
Currently `and*()` methods append conditions, while `or*()` methods
prepend them. While this doesn't affect the logic of the statement,
it might be confusing, and it contradicts the description.

Does this need tests?

### DIFF
--- a/src/Database/Query.php
+++ b/src/Database/Query.php
@@ -1986,7 +1986,7 @@ class Query implements ExpressionInterface, IteratorAggregate
         } else {
             $expression = $this->newExpr()
                 ->setConjunction($conjunction)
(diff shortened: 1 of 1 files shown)


markstory:
Seems reasonable to me. What do you think @lorenzo ?

ndm2:
My description might be a little confusing too I think (actually is kinda wrong), as it's not neccesarily the conjunction type of the new conditions, but rather whether the conjunction type of the existing expression is different to the one of the new conditions, ie when `andWhere` follows `orWhere`, the order is being reversed too.

lorenzo:
Looks good to me. Thanks @ndm2!

PR #10597, opened 2017-04-28
Fix PO parsing around context.

dereuromark:
While testing a [MessagesDbParser](https://github.com/dereuromark/cakephp-translate/blob/master/src/I18n/MessagesDbLoader.php) I found out that the core silently skips certain context translations - if they have been defined without context already.
Silently ignoring some of the translations is rather hard to detect.

I found out that the following change always works, so no more silent drops.

### DIFF
--- a/src/I18n/Parser/PoFileParser.php
+++ b/src/I18n/Parser/PoFileParser.php
@@ -142,11 +142,10 @@ class PoFileParser
 
         $translation = stripcslashes($translation);
 
-        if ($context && (!isset($messages[$singular]) || is_array($messages[$singular]))) {
+        if ($context !== null) {
             $messages[$singular]['_context'][$context] = $translation;
-        }
-        if ($context === null) {
-            $messages[$singular] = $translation;
+        } else {
+            $messages[$singular]['_context'][''] = $translation;
         }
 
         if (isset($item['ids']['plural'])) {
@@ -166,10 +165,10 @@ class PoFileParser
             $plurals = array_map('stripcslashes', $plurals);
             $key = stripcslashes($item['ids']['plural']);
 
-            if ($context) {
+            if ($context !== null) {
                 $messages[$key]['_context'][$context] = $plurals;
             } else {
-                $messages[$key] = $plurals;
+                $messages[$key]['_context'][''] = $plurals;
             }
         }
     }
--- a/src/I18n/Translator.php
+++ b/src/I18n/Translator.php
@@ -131,7 +131,11 @@ class Translator implements TranslatorInterface
 
             // No or missing context, fallback to the key/first message
             if ($context === null) {
-                $message = current($message['_context']);
+                if (isset($message['_context'][''])) {
+                    $message = $message['_context'][''];
+                } else {
+                    $message = current($message['_context']);
+                }
             } elseif (!isset($message['_context'][$context])) {
                 $message = $key;
             } elseif (is_string($message['_context'][$context]) &&
--- a/tests/TestCase/I18n/MessagesFileLoaderTest.php
+++ b/tests/TestCase/I18n/MessagesFileLoaderTest.php
@@ -56,12 +56,12 @@ class MessagesFileLoaderTest extends TestCase
         $loader = new MessagesFileLoader('default', 'en');
         $package = $loader();
         $messages = $package->getMessages();
-        $this->assertEquals('Po (translated)', $messages['Plural Rule 1']);
+        $this->assertEquals('Po (translated)', $messages['Plural Rule 1']['_context']['']);
 
         Configure::write('App.paths.locales', [TEST_APP . 'custom_locale' . DS]);
         $loader = new MessagesFileLoader('default', 'en');
         $package = $loader();
         $messages = $package->getMessages();
-        $this->assertEquals('Po (translated) from custom folder', $messages['Plural Rule 1']);
+        $this->assertEquals('Po (translated) from custom folder', $messages['Plural Rule 1']['_context']['']);
     }
 }
--- a/tests/TestCase/I18n/Parser/PoFileParserTest.php
+++ b/tests/TestCase/I18n/Parser/PoFileParserTest.php
@@ -37,7 +37,11 @@ class PoFileParserTest extends TestCase
         $messages = $parser->parse($file);
         $this->assertCount(5, $messages);
         $expected = [
-            'Plural Rule 1' => 'Plural Rule 1 (translated)',
+            'Plural Rule 1' => [
+                '_context' => [
+                    '' => 'Plural Rule 1 (translated)',
+                ],
+            ],
             '%d = 1' => [
                 '_context' => [
                     'This is the context' => 'First Context trasnlation',
@@ -52,14 +56,22 @@ class PoFileParserTest extends TestCase
                     ]
                 ]
             ],
-            '%-5d = 1' => '%-5d = 1 (translated)',
+            '%-5d = 1' => [
+                '_context' => [
+                    '' => '%-5d = 1 (translated)',
+                ],
+            ],
             '%-5d = 0 or > 1' => [
-                0 => '%-5d = 1 (translated)',
-                1 => '',
-                2 => '',
-                3 => '',
-                4 => '%-5d = 0 or > 1 (translated)'
-            ]
+                '_context' => [
+                    '' => [
+                        0 => '%-5d = 1 (translated)',
+                        1 => '',
+                        2 => '',
+                        3 => '',
+                        4 => '%-5d = 0 or > 1 (translated)'
+                    ],
+                ],
+            ],
         ];
         $this->assertEquals($expected, $messages);
     }
@@ -75,7 +87,7 @@ class PoFileParserTest extends TestCase
         $file = APP . 'Locale' . DS . 'en' . DS . 'default.po';
         $messages = $parser->parse($file);
         $this->assertCount(12, $messages);
-        $this->assertTextEquals("v\nsecond line", $messages["valid\nsecond line"]);
+        $this->assertTextEquals("v\nsecond line", $messages["valid\nsecond line"]['_context']['']);
     }
 
     /**
@@ -89,7 +101,7 @@ class PoFileParserTest extends TestCase
         $file = APP . 'Locale' . DS . 'en' . DS . 'default.po';
         $messages = $parser->parse($file);
 
-        $this->assertTextEquals('this is a "quoted string" (translated)', $messages['this is a "quoted string"']);
+        $this->assertTextEquals('this is a "quoted string" (translated)', $messages['this is a "quoted string"']['_context']['']);
     }
 
     /**
@@ -113,7 +125,7 @@ class PoFileParserTest extends TestCase
 
             return $package;
         });
-        $this->assertTextEquals('En cours', $messages['Pending']);
-        $this->assertTextEquals('En resolved', $messages['Resolved']);
+        $this->assertTextEquals('En cours', $messages['Pending']['_context']['']);
+        $this->assertTextEquals('En resolved', $messages['Resolved']['_context']['']);
     }
 }
--- a/tests/TestCase/I18n/Parser/PoFileParserTest.php
+++ b/tests/TestCase/I18n/Parser/PoFileParserTest.php
@@ -130,6 +130,8 @@ class PoFileParserTest extends TestCase
         $this->assertSame('En resolved', $messages['Resolved']['_context']['']);
         $this->assertSame('En resolved - context', $messages['Resolved']['_context']['Pay status']);
 
+        // Confirm actual behavior
+        I18n::locale('en_US');
         $this->assertSame('En cours', __('Pending'));
         $this->assertSame('En cours - context', __x('Pay status', 'Pending'));
         $this->assertSame('En resolved', __('Resolved'));
--- a/tests/TestCase/I18n/Parser/PoFileParserTest.php
+++ b/tests/TestCase/I18n/Parser/PoFileParserTest.php
@@ -125,7 +125,14 @@ class PoFileParserTest extends TestCase
 
             return $package;
         });
-        $this->assertTextEquals('En cours', $messages['Pending']['_context']['']);
-        $this->assertTextEquals('En resolved', $messages['Resolved']['_context']['']);
+        $this->assertSame('En cours', $messages['Pending']['_context']['']);
+        $this->assertSame('En cours - context', $messages['Pending']['_context']['Pay status']);
+        $this->assertSame('En resolved', $messages['Resolved']['_context']['']);
+        $this->assertSame('En resolved - context', $messages['Resolved']['_context']['Pay status']);
+
+        $this->assertSame('En cours', __('Pending'));
+        $this->assertSame('En cours - context', __x('Pay status', 'Pending'));
+        $this->assertSame('En resolved', __('Resolved'));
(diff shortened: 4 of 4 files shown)


markstory:
What kind of impact does this have on translating strings? It might be a good idea to have a test that includes parsing the po file and translating messages.

dereuromark:
According to my tests [here](https://github.com/dereuromark/cakephp-translate/blob/master/tests/TestCase/I18n/MessagesDbLoaderTest.php#L46) it has 0 impact on the translation capabilities, it only allows to keep the same string for context and non-context instead of silently skipping one of them.

Yes, some more additional tests here for the translating might be good.

lorenzo:
According to the diff it seem like combining context + plural or singular may be affected by this change. Can you verify if that is the case?

dereuromark:
What do you mean @lorenzo ?

dereuromark:
One regression I currently see with the code as it is now:
```
msgid "About"
msgstr ""
```
Empty translations do not show up with the original id anymore, but an empty string.

Maybe someone has a better approach to this - or we need to close this.

dereuromark:
Ah, so seems like the regression I thought was because of this PR, but in fact it seems to be from a different change, and now fixed with https://github.com/cakephp/cakephp/pull/10618

markstory:
Are we good to close this then @dereuromark ?

dereuromark:
This aims to fix a different issue, but maybe I still dont understand What Lorenzo meant. What is the downside?

dereuromark:
It now works with both orders: "none and context" + "context and none"
But it only works for de_DE
So if en_US is the default and already set up in bootstrap, it seems you cannot overwrite it.
Is this something inside or outside the scope of this issue?

dereuromark:
You can run the tests I added with the old code base, you will see that the for __()/__x() will not work out.

dereuromark:
@lorenzo Are you OK with the suggested fix that would now allow having the same string for both context and non-context existing side by side?

src/I18n/Parser/PoFileParser.php
```diff
@@ -142,11 +142,10 @@ protected function _addMessage(array &$messages, array $item)
...
         $translation = stripcslashes($translation);
 
-        if ($context && (!isset($messages[$singular]) || is_array($messages[$singular]))) {
+        if ($context !== null) {
             $messages[$singular]['_context'][$context] = $translation;
-        }
-        if ($context === null) {
-            $messages[$singular] = $translation;
+        } else {
+            $messages[$singular]['_context'][''] = $translation;
```
markstory: I wonder what the memory impact of this change will be like on larger applications.

dereuromark: Do you think arrays in modern PHP cannot handle this gracefully? I suspect this to be none, since the additional data is minimal.

markstory: This change adds an extra array and extra string for every translated value. The delta looks like it is 424 bytes per message though. In an application with 10K messages we'd be adding ~4.1MB

```
$ php -r '$a = "testing"; echo memory_get_usage() . "\n\n";'
350064

$ php -r '$a = ["" => "testing"]; echo memory_get_usage() . "\n\n";'
350488
```

dereuromark: Do you think the additional overhead is justified here in terms of functional completeness/expectation? IMO those improvements here trump the overhead - at least the silent dropping of translations sounds rather like a hard to debug and understandable issue otherwise.
Or is there any better way of dealing with it that would make this less lengthening?

ravage84:
> ... I found out that the core silently skips certain context translations - if they have been defined without context already.

I can confirm this behavior in CakePHP 3.4.6. Happened to us just this week.

dereuromark:
@ravage84 Can you also confirm that this fixes it?

ravage84:
Right now, I cannot. But I'm going to forward this to the dev who ran into this. He can have a look at it when he is back on tuesday next week.

@TekkCraft can you please check if this patch fixes the silently ignored translation we found this week?

ravage84:
@dereuromark unfortunately due to the lack of time and since we couldn't require your branch through Composer, we weren't able to verify the fix. Sorry

src/I18n/Parser/PoFileParser.php
burzum: @markstory I would prefer to have the overhead than fucking around with really painful to debug issues like this one, which very likely has the same cause https://github.com/cakephp/cakephp/issues/10730.

dereuromark: That's exactly what I was thinking :)

markstory: @burzum Have you been able to test these changes to see if they fix your issue?

burzum:
I'm not sure if this is related. But your patch doesn't fix this @dereuromark 

`timeAgoDate($date)` is producing `1 day` instead of `2 days` for whatever reason, but *only* when the locale is `en`. It works with fr, de, ca, cn. It also works if I remove the translation for that msgid from the po file. I have no idea whats wrong there. :(

Here is our /src/Locale/en/cake.po file

```
# LANGUAGE translation of CakePHP Application
# Copyright YEAR NAME <EMAIL@ADDRESS>
#
#, fuzzy
msgid ""
msgstr ""
"Project-Id-Version: PROJECT VERSION\n"
"POT-Creation-Date: 2017-06-02 08:39+0000\n"
"PO-Revision-Date: YYYY-mm-DD HH:MM+ZZZZ\n"
"Last-Translator: NAME <EMAIL@ADDRESS>\n"
"Language-Team: LANGUAGE <EMAIL@ADDRESS>\n"
"MIME-Version: 1.0\n"
"Content-Type: text/plain; charset=utf-8\n"
"Content-Transfer-Encoding: 8bit\n"
"Plural-Forms: nplurals=INTEGER; plural=EXPRESSION;\n"

#: Template/Error/error400.ctp:34
#: Template/Error/error500.ctp:35
msgid "Error"
msgstr ""

#: Template/Error/error400.ctp:35
msgid "The requested address %s was not found on this server."
msgstr ""

#: Template/Error/error500.ctp:33
msgid "An Internal Error Has Occurred"
msgstr ""

#: Controller/Component/AuthComponent.php:496
msgid "You are not authorized to access that location."
msgstr ""

#: Controller/Component/CsrfComponent.php:155
msgid "Missing CSRF token cookie"
msgstr ""

#: Controller/Component/CsrfComponent.php:159
msgid "CSRF token mismatch."
msgstr ""

#: Error/ExceptionRenderer.php:258
msgid "Not Found"
msgstr ""

#: Error/ExceptionRenderer.php:260
msgid "An Internal Error Has Occurred."
msgstr ""

#: Http/Response.php:2258
msgid "The requested file contains `..` and will not be read."
msgstr ""

#: Http/Response.php:2269
msgid "The requested file was not found"
msgstr ""

#: I18n/Number.php:90
msgid "{0,number,#,###.##} KB"
msgstr ""

#: I18n/Number.php:92
msgid "{0,number,#,###.##} MB"
msgstr ""

#: I18n/Number.php:94
msgid "{0,number,#,###.##} GB"
msgstr ""

#: I18n/Number.php:96
msgid "{0,number,#,###.##} TB"
msgstr ""

#: I18n/Number.php:88
msgid "{0,number,integer} Byte"
msgid_plural "{0,number,integer} Bytes"
msgstr[0] ""
msgstr[1] ""

#: I18n/RelativeTimeFormatter.php:80
msgid "{0} from now"
msgstr ""

#: I18n/RelativeTimeFormatter.php:80
msgid "{0} ago"
msgstr ""

#: I18n/RelativeTimeFormatter.php:83
msgid "{0} after"
msgstr ""

#: I18n/RelativeTimeFormatter.php:83
msgid "{0} before"
msgstr ""

#: I18n/RelativeTimeFormatter.php:114
msgid "just now"
msgstr ""

#: I18n/RelativeTimeFormatter.php:151
msgid "about a second ago"
msgstr ""

#: I18n/RelativeTimeFormatter.php:152
msgid "about a minute ago"
msgstr ""

#: I18n/RelativeTimeFormatter.php:153
msgid "about an hour ago"
msgstr "an hour ago"

#: I18n/RelativeTimeFormatter.php:154;332
msgid "about a day ago"
msgstr "a day ago"

#: I18n/RelativeTimeFormatter.php:155;333
msgid "about a week ago"
msgstr "a week ago"

#: I18n/RelativeTimeFormatter.php:156;334
msgid "about a month ago"
msgstr "a month ago"

#: I18n/RelativeTimeFormatter.php:157;335
msgid "about a year ago"
msgstr "a year ago"

#: I18n/RelativeTimeFormatter.php:168
msgid "in about a second"
msgstr ""

#: I18n/RelativeTimeFormatter.php:169
msgid "in about a minute"
msgstr ""

#: I18n/RelativeTimeFormatter.php:170
msgid "in about an hour"
msgstr ""

#: I18n/RelativeTimeFormatter.php:171;346
msgid "in about a day"
msgstr ""

#: I18n/RelativeTimeFormatter.php:172;347
msgid "in about a week"
msgstr ""

#: I18n/RelativeTimeFormatter.php:173;348
msgid "in about a month"
msgstr ""

#: I18n/RelativeTimeFormatter.php:174;349
msgid "in about a year"
msgstr ""

#: I18n/RelativeTimeFormatter.php:304
msgid "today"
msgstr ""

#: I18n/RelativeTimeFormatter.php:370
msgid "%s ago"
msgstr "%s ago"

#: I18n/RelativeTimeFormatter.php:371
msgid "on %s"
msgstr ""

#: I18n/RelativeTimeFormatter.php:47;126;316
msgid "{0} year"
msgid_plural "{0} years"
msgstr[0] ""
msgstr[1] ""

#: I18n/RelativeTimeFormatter.php:51;129;319
msgid "{0} month"
msgid_plural "{0} months"
msgstr[0] ""
msgstr[1] ""

#: I18n/RelativeTimeFormatter.php:57;132;322
msgid "{0} week"
msgid_plural "{0} weeks"
msgstr[0] ""
msgstr[1] ""

#: I18n/RelativeTimeFormatter.php:59;135;325
msgid "{0} day"
msgid_plural "{0} days"
msgstr[0] ""
msgstr[1] ""

#: I18n/RelativeTimeFormatter.php:64;138
msgid "{0} hour"
msgid_plural "{0} hours"
msgstr[0] ""
msgstr[1] ""

#: I18n/RelativeTimeFormatter.php:68;141
msgid "{0} minute"
msgid_plural "{0} minutes"
msgstr[0] ""
msgstr[1] ""

#: I18n/RelativeTimeFormatter.php:72;144
msgid "{0} second"
msgid_plural "{0} seconds"
msgstr[0] ""
msgstr[1] ""

#: ORM/RulesChecker.php:57
msgid "This value is already in use"
msgstr ""

#: ORM/RulesChecker.php:104
msgid "This value does not exist"
msgstr ""

#: ORM/RulesChecker.php:128
msgid "The count does not match {0}{1}"
msgstr ""

#: Utility/Text.php:899
msgid "and"
msgstr ""

#: Validation/Validator.php:103
msgid "This field is required"
msgstr ""

#: Validation/Validator.php:104
msgid "This field cannot be left empty"
msgstr ""

#: Validation/Validator.php:1801
msgid "The provided value is invalid"
msgstr ""

#: View/Helper/FormHelper.php:1016
msgid "New %s"
msgstr ""

#: View/Helper/FormHelper.php:1019
msgid "Edit %s"
msgstr ""

#: View/Helper/FormHelper.php:1895
msgid "Submit"
msgstr ""

#: View/Helper/HtmlHelper.php:793
msgid "Home"
msgstr ""

#: View/Widget/DateTimeWidget.php:540
msgid "January"
msgstr ""

#: View/Widget/DateTimeWidget.php:541
msgid "February"
msgstr ""

#: View/Widget/DateTimeWidget.php:542
msgid "March"
msgstr ""

#: View/Widget/DateTimeWidget.php:543
msgid "April"
msgstr ""

#: View/Widget/DateTimeWidget.php:544
msgid "May"
msgstr ""

#: View/Widget/DateTimeWidget.php:545
msgid "June"
msgstr ""

#: View/Widget/DateTimeWidget.php:546
msgid "July"
msgstr ""

#: View/Widget/DateTimeWidget.php:547
msgid "August"
msgstr ""

#: View/Widget/DateTimeWidget.php:548
msgid "September"
msgstr ""

#: View/Widget/DateTimeWidget.php:549
msgid "October"
msgstr ""

#: View/Widget/DateTimeWidget.php:550
msgid "November"
msgstr ""

#: View/Widget/DateTimeWidget.php:551
msgid "December"
msgstr ""
```

burzum:
@markstory this is easy to reproduce. If you put the above pasted po file as `cake.po` into the `en` locale folder of the TestApp and run the I18n tests you'll get this awesome explosion of errors:

image

Houston, we have a problem! 😟

ravage84:
I can confirm this. Just tried @burzum 's file. Got 81 failures. 🔥

markstory:
That file as is has an invalid nplurals header. So it won't work.

markstory:
With that said. The PO parser should emit a warning/error when an invalid plural header has been found. I'll try this file out with a valid plural header when I get a chance.

burzum:
First time I hear / read about this. Is it even in our docs? I know its related to the po files but it should be in the docs in a big warning.

ravage84:
I have setup a small test environment and documented my tests including screenshots in https://github.com/ravage84/cakephp-po-issue-proof/commits/master for my issue. I'm not sure though, if my issue is really the same as yours @dereuromark .

Also tested your patched PoFileParser.php which seems to **not fix** my issue.

burzum:
The issues cause (or one of them) is in this part of the `PoFileParser`:

```php
            } elseif (substr($line, 0, 7) === 'msgstr[') {
                if ($resource === 'C:\xampp\htdocs\cakephp\tests\test_app\TestApp\Locale\en\cake.po') {
                    $size = strpos($line, ']');
                    //debug((int)substr($line, 7, 1));
                    //debug(substr($line, $size + 3, -1));
                    //debug($line);

                    $value = substr($line, $size + 3, -1);
                    if (empty($value)) {
                        $value = 'TEST';
                    }
                }

                $item['translated'][(int)substr($line, 7, 1)] = $value;
            }
```

If $value is empty the string that gets inserted into the map is `''` (empty string). Not sure how to fix it... Because i don't know what happens for languages with more than one plural. I would say the solution is to copy the values from singular and plural if $value is empty.

Another solution would be to document that uses have to "translate" even the po file that comes with the msgids of their native language.

dereuromark:
I agree, my issue is unrelated, even though similar.
Those are two different things to be fixed :)

ravage84:
I have created #10774, so we can work on that separately, if necessary.

burzum:
@markstory there is still no proper fix for my issue? I proposed a fix above but it still has problems. Can we discuss how to resolve the issue?

dereuromark:
@burzum Can you open a PR with your suggested fix?

burzum:
@dereuromark the code above doesn't really fix it but I've created a branch with the po file. https://github.com/cakephp/cakephp/tree/i18n-translation-issue You can just check it out and run the i18n tests to see things blowing up.

markstory:
@burzum There isn't a fix that has an open pull request. I don't really have the time to dedicate to sorting out what is going on here.

src/I18n/Parser/PoFileParser.php
burzum: @markstory the issue still persist for me. Just tried it against the current master a6bf3f04b6499eb3aacb19ffcad9958e3aa8c962

PR #10599, opened 2017-05-01
3.next - Connect Middleware pipelines in RoutingMiddleware

markstory:
Find and apply any scoped middleware from the RoutingMiddleware. With these changes, scoped middleware is now functional, but the proposed changes to `Application` are incomplete, and I'll be doing those next.

I've intentionally not modified the `RoutingFilter` behavior as I feel that scoped middleware is a good reason to migrate to the new `Application` class and PSR7 stack which will become standard in future versions

Refs #10308

### DIFF
--- a/src/Routing/Middleware/RoutingMiddleware.php
+++ b/src/Routing/Middleware/RoutingMiddleware.php
@@ -62,8 +62,7 @@ class RoutingMiddleware
             $runner = new Runner();
 
             return $runner->run($middleware, $request, $response);
-        } else {
-            return $next($request, $response);
         }
+        return $next($request, $response);
     }
 }
--- a/src/Routing/Middleware/RoutingMiddleware.php
+++ b/src/Routing/Middleware/RoutingMiddleware.php
@@ -29,6 +29,11 @@ class RoutingMiddleware
 {
 
     /**
+     * Apply routing and update the request.
+     *
+     * Any route/path specific middleware will be wrapped around $next and then the new middleware stack
+     * will be invoked.
+     *
      * @param \Psr\Http\Message\ServerRequestInterface $request The request.
      * @param \Psr\Http\Message\ResponseInterface $response The response.
      * @param callable $next The next middleware to call.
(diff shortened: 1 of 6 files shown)


src/Routing/Middleware/RoutingMiddleware.php
```diff
@@ -55,7 +56,14 @@ public function __invoke(ServerRequestInterface $request, ResponseInterface $res
                 $response->getHeaders()
             );
         }
+        $middleware = Router::getMatchingMiddleware($request->getUri()->getPath());
```
lorenzo: I wonder if instead of calling this function statically, this class could get the topmost routing collection instance so that it could call the method directly on it.

lorenzo:
This looks nice and clean to me. would be nice if the collection could be passed instead of invoked statically, though :)

src/Routing/Middleware/RoutingMiddleware.php
markstory: Router doesn't have a way to get the collection right now. I could change that though. It might be nice if the middleware could have a route collection passed as a constructor parameter.

markstory:
@lorenzo I have some plans to update `Application` so that we can deprecate static use of `Router` for connecting routes. They involve adding a `routing` method to the `Application`, and having that hook receive a `RouteBuilder`. I think that could be triggered from the `RoutingMiddleware` constructor. Something like.

```php
class Application
{
  public function middleware($middleware)
  {
    // RoutingMiddleware would call `routing()` on the application class here.
    $middleware->add(new RoutingMiddleware($this));

    return $middleware;
  }
}
```

src/Routing/Middleware/RoutingMiddleware.php
```diff
@@ -55,7 +56,14 @@ public function __invoke(ServerRequestInterface $request, ResponseInterface $res
...
+        if ($middleware) {
+            $middleware->add($next);
+            $runner = new Runner();
 
-        return $next($request, $response);
+            return $runner->run($middleware, $request, $response);
+        } else {
```
bcrowe: Probably could do without this `else`, eh?

markstory: Good point.

lorenzo:
> I think that could be triggered from the RoutingMiddleware constructor.

@markstory That makes sense to me :)

src/Routing/Router.php
```diff
@@ -1074,6 +1075,26 @@ public static function routes()
...
+     * @return \Cake\Http\MiddlewareQueue|null Either a queue or null if there are no matching middleware.
+     */
+    public static function getMatchingMiddleware($path)
+    {
+        if (!static::$initialized) {
+            static::_loadRoutes();
+        }
+
+        $middleware = static::$_collection->getMatchingMiddleware($path);
+        if ($middleware) {
```
dereuromark: Small thing: If we invert, the happy case would be the final line in the method, aborting early with a return null then instead.

PR #10611, opened 2017-05-03
Add Mail getter and setter method annotation

Iandenh:
This pull request will add `@method`'s for the Mail getters and setters. 

For the deprecated `@method` there are currently no standard do define it as deprecated. 

phpDocumentor/phpDocumentor2#1604


cleptric:
This looks like a lot of noise to me. I don't use an IDE, so maybe I have a different opinion on this then someone who does.

dereuromark (approved):

inoas:
Is there a plan to add this to all core classes?

dereuromark:
What do you mean? You only need it where magic access happens.

PR #10614, opened 2017-05-03
Fix Hash::extract() to not array cast objects unnecessarily.

dereuromark:
Resolves https://github.com/cakephp/cakephp/issues/8053

Is this acceptable to be announced in migration notes in 3.5?
Should we still add a feature flag of some sort to opt-in the object preserving extract?

The \ArrayAccess objects itself do not change, only the magic ones like Time() now don't get unnecessarily array transformed.

### DIFF
--- a/tests/TestCase/Utility/HashTest.php
+++ b/tests/TestCase/Utility/HashTest.php
@@ -15,6 +15,7 @@
 namespace Cake\Test\TestCase\Utility;
 
 use ArrayObject;
+use Cake\I18n\Time;
 use Cake\ORM\Entity;
 use Cake\TestSuite\TestCase;
 use Cake\Utility\Hash;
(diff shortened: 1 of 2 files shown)


tests/TestCase/Utility/HashTest.php
```diff
@@ -1489,6 +1490,32 @@ public function testExtractUnevenKeys()
...
+                ]),
+                'created' => new Time('2010-01-05'),
+            ],
+        ];
+
+        $result = Hash::extract($data, 'root.created');
+        $this->assertSame([$data['root']['created']], $result);
+
+        $result = Hash::extract($data, 'root.array');
+        $this->assertSame($data['root']['array'], $result);
```
dereuromark: I wonder if the expected behavior here is to return it as [...] as well.

markstory: Are there existing tests for what happens to `ArrayObject` before this change?

dereuromark: I tested it, it was an array before, now stays \ArrayObject if possible.
It should not change code accessing it as array, then of course. So pretty much BC here.

markstory: Unless the function results are put into one of the `array_*` functions. Then its not backwards compatible.

jeremyharris: I also think `+=` doesn't work with ArrayObjects

dereuromark: Ok so we return Array here too

dereuromark:
I wonder if with the current code now it could also be working as master-based bugfix instead of 3.5.
But I am fine either way.

markstory (approved):
Looks good to me.

alexlambson:
If I may reopen this issue, the fix in 0c5c24c93d30a876e31b3b23ee87f96c89c31ef8 doesn't account for any kind of path like: 
`{n}.types.{n}.user_carts`

Currently, I am trying to combine a bunch of carts for a race. Races are linked to user_carts via `Races.Types.UserCarts.` When I run Hash::extract($race, 'types.{n}.user_carts.{n}');`
It returns the correct data, but cast to an array. 

My quick workaround to this issue was done by modifying lines 160:172 is src/Utility/Hash.php

```
foreach ($context[$_key] as $item) {
    **$arrayItem = $item;**
    if (is_object($item) && method_exists($item, 'toArray')) {
        **$arrayItem = $item->toArray();**
    }
    foreach ((array)$arrayItem as $k => $v) {
        if (static::_matchToken($k, $token)) {
            **$next[] = $item[$k];**
        }
    }
}
```

markstory:
> If I may reopen this issue

Pull requests cannot be reopened. Please open a new issue with the problem you're having.

PR #10617, opened 2017-05-04
Added a test case for errors when validation is only for create

davidyell:
So I've been encountering a strange issue today when trying to add a new user record. It continually fails with `id field is required` or similar validation errors. Even with no validation rules set, it still fails.

At first I thought the culprit might have been inverted logic in the `_checkPresence` method. Thought I'd try and write a test to abstract the problem from my project codebase.

Needless to say the testcase and other Validator tests pass just fine.

Just thought I'd submit a test case in case it was any use.

### DIFF
--- a/tests/TestCase/Validation/ValidatorTest.php
+++ b/tests/TestCase/Validation/ValidatorTest.php
@@ -298,6 +298,28 @@ class ValidatorTest extends TestCase
     }
 
     /**
+     * Test that validation on a certain condition generate errors
+     * 
+     * @return void
+     */
+    public function testErrorsWithPresenceRequiredOnCreate()
+    {
+        $validator = new Validator;
+        $validator->requirePresence('id', 'update');
+        $validator->allowEmpty('id', 'create');
+        $validator->requirePresence('title');
+
+        $data = [
+            'title' => 'Example title'
+        ];
+
+        $expected = [];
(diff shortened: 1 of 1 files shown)


davidyell:
Turns out this happens if you do not declare `setPrimaryKey()` inside the table class. I had mistakenly updated my accessor method to `getPrimaryKey()` instead. D'oh

davidyell:
How does adding a single test case reduce coverage?!

thinkingmedia:
@davidyell it's relative to the master branch. So maybe more coverage was added to the master which makes this branch fall behind.

tests/TestCase/Validation/ValidatorTest.php
```diff
@@ -298,6 +298,28 @@ public function testErrorsWithPresenceRequired()
     }
 
     /**
+     * Test that validation on a certain condition generate errors
+     *
+     * @return void
+     */
+    public function testErrorsWithPresenceRequiredOnCreate()
+    {
+        $validator = new Validator;
```
dereuromark: should add ()

davidyell: That would make it inconsistant with the other tests in this file.

markstory: A separate pull request works for me.

Issue #10440, opened 2017-03-22
Default value for __x() with context not returned if non default language translation is empty


Antoniossss:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.4.3

I am not sure if this is desired behavior or not. I have pot and po files generated by Cake's i18n shell. 

I have following translation in non default language `pl` 

```
msgctxt "Navigation menu"
msgid "Search"
msgstr ""
```

And now, behavior is different depending on language beeing default or not.
 `__x("Navigation menu", "Search")`  gives me an empty string in language `pl`. If i switch to any other language, so the default lang would be picked up (default.pot in my case) default value "Search" is returned as I would expect it to be.

Defining translation for example like this
```
msgctxt "Navigation menu"
msgid "Search"
msgstr "test test"
```
will result in returning "test test" for `pl` and "Search" for default language.

markstory:
You've asked the translation value to be `''` so that's what you should get. If you don't want `''` don't make it the translated value.

Antoniossss:
But this is the casewhere translation is missing due to overlook. Besides behavior is inconsistent - for example __("Search") will give me the default value if I ommit the context. 

Up to you what to do about it anyway. Thanks

markstory:
Pot files are never read when translating messages did you mean the po files instead? I might not be fully understanding the issue you are having. Do you have snippets of the po files and translation calls that aren't doing what you expect?

Antoniossss:
I meant `po` files, yes.

And for example, lets have following snippet from `po` file (These are actuall snippets used in my app):
```
#: Template/Pages/home.ctp:29
msgid "home.story.short"
msgstr ""

#: Template/Layout/default.ctp:92;122
#: Template/Pages/home.ctp:14
msgctxt "Navigation menu"
msgid "Search"
msgstr ""
```

and lets have
```

<?= __("home.story.short") ?> // output: home.story.short
<?= __x("Navigation menu", "Search") ?> // output: empty string

```
Whats the difference here? Both translations are described as "empty string" like you have pointed out, but outcome is different.

markstory:
Ok. That behaviour should be more consistent.

markstory:
Returning '' was added incorrectly in #8932

markstory:
pull request up.

Issue #8932, opened 2016-06-03
Empty Translation-String error __x()  


DMO-L:
This is a (multiple allowed):
* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.2.10
* Platform and Target: LAMP

### What you did
Try to translate a string by using the __x() function. 

`<?php $link = $this->Html->link(__x('Homepage','art'),'/linkuri'); ?>`

### Expected Behavior
An Empty string or the msgid from the .po file.

### Actual Behavior
When I use the __x() function for translating, and the msgstr is empty in the Locale/XX/default.po file a "Constructor failed" error appears.

lorenzo:
how does your .po file look like?

DMO-L:
```
#: Template/Pages/home.ctp:23
msgctxt "Homepage"
msgid "art"
msgstr ""
```
When i change it to
 ```
#: Template/Pages/home.ctp:23
msgctxt "Homepage"
msgid "art"
msgstr "art"
```
it works without the error.

micdobro:
I can confirm this is also an issue. Basically the moment I started using ``__x()`` in my code the translation machinery in cake takes down the whole app and nothing works. Apparently not only having just an empty translation, but also the same message with and without context is making the ``.po`` reader collapse.

markstory:
Pull request open now.

PR #10618, opened 2017-05-05
Empty message strings should return the key

markstory:
context based messages should return the key when the translated value is ''. This mirrors the behavior of context-less messages. Returning '' was incorrectly added in #8932

Refs #10440

### DIFF
--- a/src/I18n/Translator.php
+++ b/src/I18n/Translator.php
@@ -119,7 +119,6 @@ class Translator implements TranslatorInterface
     public function translate($key, array $tokensValues = [])
     {
         $message = $this->getMessage($key);
-
         if (!$message) {
             // Fallback to the message key
             $message = $key;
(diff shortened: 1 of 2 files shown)


ravage84 (approved):

dereuromark:
Is it possible that this introduced a regression?
See https://travis-ci.org/dereuromark/cakephp-translate/jobs/234303140
This was green before this merge, and now is red after.

The https://github.com/dereuromark/cakephp-translate/blob/master/tests/TestCase/I18n/MessagesDbLoaderTest.php#L56 test shows it should not transform into context.
Maybe this is also related to the deeper underlying issue of https://github.com/cakephp/cakephp/issues/10597 ?

markstory:
Not sure this specific change would cause that. The checks change here shouldn't be the cases that get hit for a singular message without context.

PR #10621, opened 2017-05-05
Fix some errors reported by phpstan level 2.

ADmad:


src/ORM/Behavior/CounterCacheBehavior.php
```diff
@@ -198,12 +199,12 @@ protected function _processAssociations(Event $event, EntityInterface $entity)
...
      * @param \Cake\Event\Event $event Event instance.
-     * @param \Cake\Datasource\EntityInterface $entity Entity
+     * @param \Cake\ORM\Entity $entity Entity
      * @param \Cake\ORM\Association $assoc The association object
      * @param array $settings The settings for for counter cache for this association
      * @return void
      */
-    protected function _processAssociation(Event $event, EntityInterface $entity, Association $assoc, array $settings)
+    protected function _processAssociation(Event $event, Entity $entity, Association $assoc, array $settings)
     {
```
markstory: Does this method use features not in EntityInterface?

ADmad: The method uses `$entity->extractOriginalChanged()`. This method is not declared in`EntityInterface`, it's only available in `Entity`.

markstory: Ok.

PR #10622, opened 2017-05-05
Added a test for nested array data as a query in the url

davidyell:
I encountered an issue today where trying to form a url such as `example.com/api/v4/packages?sort=name&direction=desc&limit=2&verbose=1&filter[Broadbands.bandwidth]=30&site_id=4` was causing a fatal error with Zend Diactoros' Uri class, as malformed.

Just wanted to check that the Cake HTTP client was url encoding query params, turns out it was, but there was no test for query param arrays. So this change adds a simple test to ensure that nested array data is url encoded properly.


markstory:
Thanks!

PR #10624, opened 2017-05-05
Change ConnectionInterface typehint to Connection in TableSchema.

robertpustulka:
`ConnectionInterface` does not provide a `driver()` method which `*Sql()` methods require.

### DIFF
--- a/src/Database/Schema/Collection.php
+++ b/src/Database/Schema/Collection.php
@@ -15,7 +15,7 @@
 namespace Cake\Database\Schema;
 
 use Cake\Database\Exception;
-use Cake\Datasource\ConnectionInterface;
+use Cake\Database\Connection;
 use PDOException;
 
 /**
@@ -30,7 +30,7 @@ class Collection
     /**
      * Connection object
      *
-     * @var \Cake\Datasource\ConnectionInterface
+     * @var \Cake\Database\Connection
      */
     protected $_connection;
 
@@ -44,9 +44,9 @@ class Collection
     /**
      * Constructor.
      *
-     * @param \Cake\Datasource\ConnectionInterface $connection The connection instance.
+     * @param \Cake\Database\Connection $connection The connection instance.
      */
-    public function __construct(ConnectionInterface $connection)
+    public function __construct(Connection $connection)
     {
         $this->_connection = $connection;
         $this->_dialect = $connection->driver()->schemaDialect();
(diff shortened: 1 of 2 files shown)


ADmad:
:+1: phpstan also complains about this.

dereuromark:
Shouldnt the interface be fixed instead?
Going back to non Interface typehinting looks like a step backwards.

robertpustulka:
I remember "fixing" interfaces in minor versions was an issue somewhere due to the BC.

dereuromark:
Well changing the typehints of public methods is pretty similar :) But yeah, adding methods to interfaces itself is BC breaking, unless one uses Interface aggregation or sth.

markstory:
This might cause compatibility breaks, but I suspect that the extent of this breakage would be much smaller than changing an interface. This is something that would need to be documented in the migration guide though.

ADmad:
Also the chances that someone has extended `TableSchema` are very low, so documenting this change in 3.5's migration guide should be good enough.

ADmad:
Same issue lies in `Cake\Database\Schema\Collection::__construct()`. The `$connection` argument is typehinted as `ConnectionInterface` but has `$connection->driver()` call.

dereuromark:
I noticed the same issues before, as well. But I think not much we can do here right now (and it doesn't hurt anyone for the time being). We should fix all this up to proper Interface typehinting in 4.x.

robertpustulka:
So how should I proceed? Find all occurences of wrong typehints, replace them and update the docs?

ADmad:
I would be in favor of changing the typehint of `Cake\Database\Schema\Collection::__construct()` too. Besides `cakedc/cakephp-oracle-driver` I doubt there is any other userland code out there which extends that class. [OracleConnection](https://github.com/CakeDC/cakephp-oracle-driver/blob/master/src/Database/OracleConnection.php) extends `Collection` but doesn't override the constructor so it wouldn't be affected if the typehint was changed.

lorenzo:
Makes sense to me. I did not like the change when it got changed to use the interface for the same reason it is going back today :)

dereuromark:
@lorenzo Shouldnt the interface have included the driver() in the first place?
Whenever we are not using interfaces but concrete classes, that indicates a clear code smell in the architecture.

robertpustulka:
@dereuromark I think it should but this probably will be revisited in 4.0. Add conneciton methods ti the interface, create DriverInterface (we have only abstract class AFAIR) and fix the typehints against interfaces.

PR #10625, opened 2017-05-05
Fix _validatePost returns true when empty form is submitted

chinpei215:


src/Controller/Component/SecurityComponent.php
```diff
@@ -102,7 +102,7 @@ public function startup(Event $event)
         $controller = $event->getSubject();
         $this->session = $controller->request->getSession();
         $this->_action = $controller->request->getParam('action');
-        $hasData = (bool)$controller->request->getData();
+        $hasData = ($controller->request->getData() || $controller->request->is(['put', 'post', 'delete', 'patch']));
```
chinpei215: I didn't remove `$controller->request->getData()` as it was needed to pass existing tests.

markstory:
Nice work!

Issue #10616, opened 2017-05-04
3.4.x: flash messages on tests


mirko-pagliai:
Probably this is my mistake, but I need some help to figure out what does not work.

I'm writing some tests using `get()` / `post()` methods provided by `IntegrationTestCase` and I'm testing some flash messages.

With one action (and **only for one**) it seems that there is nothing in the session.

My action:

    public function forgotPassword() {
        // ...
        $this->Flash->error(__d('me_cms', 'No account found'));
        // ...
    }

My test:

        $this->post($url, ['email' => 'wrongMail@example.it', 'email_repeat' => 'wrongMail@example.it']);
        dd($this->_requestSession->read());

Result:

```
########## DEBUG ##########
[
        'Flash' => []
]
###########################
```

Some oddities. If I add some debug into the action:

    public function forgotPassword() {
        // ...
        $this->Flash->error(__d('me_cms', 'No account found'));
        debug($this->Flash->_session->read());
        // ...
    }

Result:

```
########## DEBUG ##########
[
        'Flash' => [
                'flash' => [
                        (int) 0 => [
                                'message' => 'No account found',
                                'key' => 'flash',
                                'element' => 'MeTools.Flash/flash',
                                'params' => [
                                        'class' => 'alert-danger'
                                ]
                        ]
                ]
        ]
]
###########################
tests/TestCase/Controller/UsersControllerTest.php (line 389)
########## DEBUG ##########
[
        'Flash' => []
]
###########################
```

Then I thought: ok, there is something that **after** the action is executed, it changes the session.
Yet, if I check the body of the response (`_getBodyAsString()`), the flash message is displayed correctly.

What can be wrong?

markstory:
Not sure what's wrong. I have a few apps that do assertions on flash messages without any problems.

jeremyharris:
I think this is because messages are popped from the session when they are displayed in the layout, which is why they are in the body but not the session. I've run into this before.

markstory:
@jeremyharris Yes! displaying a flash message would totally remove it from the session.

mirko-pagliai:
So, what should I do to assert flash messages?

jeremyharris:
I would like to see a way to assert against messages as well. I think I'll play with some ideas and come up with a PR today.

Issue #10157, opened 2017-02-02
IntegrationTestCase::_sendRequest doesn't handle errors anymore


cleptric:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [x] feature-discussion (RFC)

* CakePHP Version: 3.3.13

### What you did
I wrote some Integration test cases. Some of them trigger a method in my component which throws an `UnauthorizedException`.

### What happened
The exceptions are catched by the `ErrorHandlerMiddleware` and printed out to `stderr`.

### What you expected to happen
As a workaround, I did 
```
Configure::write([
    'Error' => [
        'skipLog' => ['Cake\Network\Exception\UnauthorizedException']
    ],
]);
```
inside my `tests\bootstrap.php`.
I would expect that these exceptions are not printed out while running `phpunit`

markstory:
I think we'd need to remove the ErrorMiddleware from the application, or figure out a way to have the middleware re-raise exceptions in testing. Not sure how to do either of those though.

markstory:
A possible implementation for this came up during @viraj-khatavkar 's talk where he replaced `Error.exceptionRenderer` with a stub that re-throws the exception. This could be packaged up into an `disableErrorMiddleware()` helper function in the IntegrationTestCase to make debugging easier.

ravage84:
Isn't the ErrorMiddleware aware of any state we could check for and act differently? Not that I like the idea...

markstory:
@ravage84 Yes another approach could be to replace the ErrorHandlerMiddleware with a stub that re throws.

cleptric:
PR open 🙂

PR #10626, opened 2017-05-05
Proposal for #10616

jeremyharris:
Refs #10616 

This adds a new method on the IntegrationTestCase called `enableRememberFlashMessages()` that stores flash messages before rendering them and restores them to the session. It's opt-in as users may be relying on the messages being discarded during render.

Naming ideas welcome 😄  I wanted to keep it similar to enabling the tokens.

### DIFF
--- a/src/TestSuite/IntegrationTestCase.php
+++ b/src/TestSuite/IntegrationTestCase.php
@@ -149,6 +149,21 @@ abstract class IntegrationTestCase extends TestCase
     protected $_csrfToken = false;
 
     /**
+     * Boolean flag for whether or not the request should re-store
+     * flash messages
+     *
+     * @var bool
+     */
+    protected $_rememberFlashMessages = false;
+
+    /**
+     * Stored flash messages before render
+     *
+     * @var null|array
+     */
+    protected $_flashMessages;
+
+    /**
      *
      * @var null|string
      */
@@ -187,6 +202,7 @@ abstract class IntegrationTestCase extends TestCase
         $this->_appArgs = null;
         $this->_securityToken = false;
         $this->_csrfToken = false;
+        $this->_rememberFlashMessages = false;
         $this->_useHttpServer = false;
     }
 
@@ -240,6 +256,17 @@ abstract class IntegrationTestCase extends TestCase
     public function enableCsrfToken()
     {
         $this->_csrfToken = true;
+    }
+
+    /**
+     * Calling this method will re-store flash messages into the test session
+     * after being removed by the FlashHelper
+     *
+     * @return void
+     */
+    public function enableRememberFlashMessages()
+    {
+        $this->_rememberFlashMessages = true;
     }
 
     /**
@@ -425,6 +452,9 @@ abstract class IntegrationTestCase extends TestCase
             $request = $this->_buildRequest($url, $method, $data);
             $response = $dispatcher->execute($request);
             $this->_requestSession = $request['session'];
+            if ($this->_rememberFlashMessages) {
+                $this->_requestSession->write('Flash', $this->_flashMessages);
+            }
             $this->_response = $response;
         } catch (\PHPUnit\Exception $e) {
             throw $e;
@@ -466,9 +496,12 @@ abstract class IntegrationTestCase extends TestCase
         }
         $this->_controller = $controller;
         $events = $controller->eventManager();
-        $events->on('View.beforeRender', function ($event, $viewFile) {
+        $events->on('View.beforeRender', function ($event, $viewFile) use ($controller) {
             if (!$this->_viewName) {
                 $this->_viewName = $viewFile;
+            }
+            if ($this->_rememberFlashMessages) {
+                $this->_flashMessages = $controller->request->session()->read('Flash');
             }
         });
         $events->on('View.beforeLayout', function ($event, $viewFile) {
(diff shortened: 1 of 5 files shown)


mirko-pagliai:
Very great work.

As for `useHttpServer()`, a good idea could be `rememberFlashMessages(boolean $enable)`.

If this (leave flash messages in the session during the tests) does not create side effects, I would have it also enabled by default.

src/TestSuite/IntegrationTestCase.php
```diff
@@ -243,6 +259,17 @@ public function enableCsrfToken()
...
 
     /**
+     * Calling this method will re-store flash messages into the test session
+     * after being removed by the FlashHelper
+     *
+     * @return void
+     */
+    public function enableRememberFlashMessages()
+    {
+        $this->_rememberFlashMessages = true;
```
mirko-pagliai: @markstory Can it cause side effects?

jeremyharris: If people rely on the messages being removed it could break their tests, that's the only reason I made it opt in. A user basically re-testing that the flash component removes the message after render would be the only issue I can think of.

jeremyharris:
I'd like to, but I'm afraid there's someone out there who has a test relying on the messages being removed.

ADmad:
We could target this for 3.next instead of master, then potentially breaking tests wouldn't be a biggie. Just need to document it.

src/TestSuite/IntegrationTestCase.php
dereuromark: Yes, people could be not checking on the status code but only on the session content.
That could silently not break the test then, but in fact false-positive show still green even though it now does not redirect anymore and consumes the session on some error page instead.

dereuromark:
Is this really a thing people need by default?
I would expect people to either expect session on redirect or not expect session and instead use the rendered body to assert against.
Mixing this sounds like a bad idea in the first place.

markstory:
> We should by default stick close to the actual results on Integration test cases IMO.

In general I agree, but both flash messages disappearing, and error rendering (#10157) make testing harder than it needs to be.

dereuromark:
The error rendering for sure :) It would be nice to have an easy kind of flag to switch it on/off if needed.
That would help with finding the actual errors.

markstory:
@dereuromark I plan on working on making error responses easier to work with once I'm done with scoped middleware.

src/TestSuite/IntegrationTestCase.php
```diff
@@ -243,6 +259,17 @@ public function enableCsrfToken()
     }
 
     /**
+     * Calling this method will re-store flash messages into the test session
+     * after being removed by the FlashHelper
+     *
+     * @return void
+     */
+    public function enableRememberFlashMessages()
```
ravage84: ``enableRetainFlashMessages()`` or  ``retainFlashMessages()``?

"Remember" is quite human.

jeremyharris: I was also thinking "restore"

jeremyharris:
Any consensus on the name?

- `enableRememberFlashMessages`
- `enableRetainFlashMessages`
- `enableRestoreFlashMessages`

Also should this be targeted against 3.next and enabled by default (and documented)? Or keep it opt-in?

dereuromark:
enableRetainFlashMessages() 👍 
I think it can easily be opt-in enabled globally by extending the IntegrationTestCase to project level and using this one instead. So no need for the core to do that in 3.x.

jeremyharris:
I've renamed the method.

mirko-pagliai:
Great work guys! Thank you!

mirko-pagliai:
Just a moment. Are flash messages properly emptied when a new request is made?
It seems to me not ...
Tomorrow I will invest for good.

jeremyharris:
@mirko-pagliai looks like I did not explicitly reset the $_flashMessages var but it should be reset as it reads from the session and writes to it regardless of what it was (empty flash messages would retain as empty), so it would overwrite it in each request.

I could see a possible bug if you're turning retain "on" for a case that's followed by a case where it's "off" and you assert flash messages in that "off" case that should've failed.

mirko-pagliai:
Good morning, keep in mind that the classic CakePHP add/edit action [is structured so](https://book.cakephp.org/3.0/en/tutorials-and-examples/blog/part-two.html#adding-articles). In other words:

    public function add()
    {
        //...
        
        if ($this->request->is('post')) {
            //...
            
            if ($this->AnEntity->save($article)) {
                $this->Flash->success(__('Success'));
            }
            $this->Flash->error(__('Error.'));
        }
        
        //...
    }


You can see [this test](https://github.com/mirko-pagliai/cakephp/commit/374c0ceb4974caeed993edc0ef0788f18a2fdd4a). This fails with:

```
There was 1 failure:

1) Cake\Test\TestCase\TestSuite\IntegrationTestCaseTest::testFlashAssertionsAfterRender
Session content for "Flash.flash.0.message" differs. 
--- Expected
+++ Actual
@@ @@
-'An error message before the redirect'
+'An error message'

/home/mirko/Libs/cakephp/src/TestSuite/IntegrationTestCase.php:1011
/home/mirko/Libs/cakephp/tests/TestCase/TestSuite/IntegrationTestCaseTest.php:423
```

It seems to me that this happens because after the `enableRetainFlashMessages()` method, flash messages are stored **only** the view is rendered.

Is it possible?

jeremyharris:
@mirko-pagliai I was able to reproduce the issue. I'll work on a PR today.

PR #10628, opened 2017-05-06
Add Validation::hexColor()

chinpei215:
A validation rule for `<input type="color">`.


tests/TestCase/Validation/ValidationTest.php
```diff
@@ -3103,4 +3103,18 @@ public function testImageWidth()
...
+     */
+    public function testColor()
+    {
+        $this->assertTrue(Validation::color('#F01234'));
+        $this->assertTrue(Validation::color('#F56789'));
+        $this->assertTrue(Validation::color('#abcdef'));
+        $this->assertTrue(Validation::color('#ABCDEF'));
+
+        $this->assertFalse(Validation::color('#fff'));
+        $this->assertFalse(Validation::color('ffffff'));
```
markstory: Could we ever get RGB or RGBA values?

chinpei215: At the moment, `input[type=color]` accepts a simple color only. 
https://developers.whatwg.org/states-of-the-type-attribute.html#color-state-(type=color)
But since Validator class doesn't allow people to change the signature without breaking BC, should I now add an additional field to the validation method for future extension? Or should I name the method `simpleColor`?

markstory: `hexColor()` is another suitable name.

chinpei215: Thank you. I would like to choose `hexColor()`. It represents well what this validation does.

markstory: :+1: Then in the future if we add more colour formats, the `color()` method can be used to aggregate the other colour validators.

src/Validation/Validation.php
```diff
@@ -1431,6 +1431,17 @@ public static function isArray($value)
     }
 
     /**
+     * Check that the input value is a 6 digits hex color.
```
ravage84: Why do you only support the ones with 6 digits?

src/Validation/Validation.php
```diff
@@ -1431,6 +1431,17 @@ public static function isArray($value)
...
+     *
+     * @param string|array $check The value to check
+     * @return bool Success
+     */
+    public static function hexColor($check)
```
ravage84: If you only support the ones with 6 digits, I would reflect that in the method name(s).

src/Validation/Validation.php
dereuromark: Thats a valid Point. Both are equally correct.

inoas: The context here was validating html input type=color and I could not find a quick answer if that supports 3 digits or only 6 digits.

As a general purpose rule it should probably support both and have a flag to toggle: `'digits' => ['both', '3', '6']` and then be set to 6 (or both) for input type color.

ravage84: I'm fine for a specific validation for a specific context, but in that case the context should be reflected in the method name .

Alternatively, there could be two methods:

1. a general purpose method, providing @ionas proposed flag parameter
2. a method specific to the input field, which in turn calls the general purpose method with the flag parameter set

chinpei215: @inoas https://developers.whatwg.org/common-microsyntaxes.html#simple-color
> ... the remaining six characters are all ASCII hex digits ...

but I am fine to add an additional option or change the method name if you want.

PR #10630, opened 2017-05-07
Update to codesniffer 3.0

ADmad:


antograssiot (approved):

ADmad:
The test case had to be updated because of [this](https://travis-ci.org/cakephp/cakephp/jobs/229728570#L726). It's quite unintuitive if you ask me that `Folder::findRecursive()` does a case insensitive search and there's no way to make it do a case sensitive search.

PR #10631, opened 2017-05-08
Enhancements to ObjectRegistry.

robertpustulka:
I've added `return $this` for method chaining and implemented `__set()`, `__isset()`, `IteratorAggregate` and `Countable`.

### DIFF
--- a/tests/TestCase/Controller/ComponentRegistryTest.php
+++ b/tests/TestCase/Controller/ComponentRegistryTest.php
@@ -209,6 +209,22 @@ class ComponentRegistryTest extends TestCase
     }
 
     /**
+     * Test __unset.
+     *
+     * @return void
+     */
+    public function testUset()
+    {
+        $eventManager = $this->Components->getController()->getEventManager();
+
+        $result = $this->Components->load('Auth');
+        unset($this->Components->Auth);
+
+        $this->assertFalse(isset($this->Components->Auth), 'Should be gone');
+        $this->assertCount(0, $eventManager->listeners('Controller.startup'));
+    }
+
+    /**
      * Test that unloading a none existing component triggers an error.
      *
      * @expectedException \Cake\Controller\Exception\MissingComponentException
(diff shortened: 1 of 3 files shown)


markstory:
Reasonable thing to add, but you have failing tests, and need tests for the new methods.

markstory (changes requested):
Reasonable thing to add, but you have failing tests, and need tests for the new methods.

lorenzo:
Looks good to me :)

robertpustulka:
@markstory is that OK to test new methods only in `ComponentRegistry`? `ObjectRegistry` is abstract and has never had it's tests.

ADmad:
@robertpustulka I think having tests for `ComponentRegistry` is fine. If you were to add tests for the abstract class you would require phpunit mock which i believe internally would just make a class which extends the abstract class :)

robertpustulka:
@ADmad I usually test abstract classes using mocks. This one doesn't have test, and I don't think it's necessary right now. Maybe in 4.0 :)

tests/TestCase/Controller/ComponentRegistryTest.php
```diff
@@ -209,6 +209,22 @@ public function testUnload()
     }
 
     /**
+     * Test __unset.
+     *
+     * @return void
+     */
+    public function testUset()
```
markstory: Typo in the method name.

markstory (approved):

tests/TestCase/Controller/ComponentRegistryTest.php
robertpustulka: @markstory I think you're referencing the old commit. It's been fixed.

ADmad:
@robertpustulka Does this mean we can use `ObjectRegistry` in authentication plugin instead of custom collection class ? :)

robertpustulka:
@ADmad we already do :)

PR #10634, opened 2017-05-09
Improving the QueryLogger

burzum:
* Adding a filter callback
* Adding a threshold to filter only slow queries


src/Database/Log/QueryLogger.php
```diff
@@ -25,6 +26,34 @@
...
+     *
+     * The $config array takes the following keys:
+     *
+     * - threshold: Threshold in milliseconds to log only query running slower than the given threshold. Default is 0.
+     * - filter: A callable to filter based on the LoggedQuery object.
+     *
+     * @param array $config Config options
+     * @return void
+     */
+    public function __construct(array $config = [])
```
markstory: How will people use these options? Isn't the query logger created behind the scenes?

burzum: No, it has a method to set it, took me some time to find it when I needed it.

```php
// bootstrap.php
ConnectionManager::get('default')->logger(new SlowQueryLogger());
ConnectionManager::get('test')->logger(new SlowQueryLogger());
```

Is what we do in our app. I enable the threshold as needed.

inoas: Could this be documented better? Was it a documentation issue in the first place?

burzum: @ionas add it to the book if you want. The method is in the API docs: https://api.cakephp.org/3.4/class-Cake.Database.Connection.html#_logger

burzum:
Follow up https://github.com/cakephp/cakephp/pull/10638

src/Database/Log/QueryLogger.php
inoas: Took a look but https://github.com/cakephp/docs/edit/3.0/en/core-libraries/logging.rst feels wrong and SlowQueryLogger doesn't exist and I am not familiar with the issue to be suited to create the docs.

Also it appears it is something that should - if possible - be setup in app.php over bootstrap.php?

burzum: @inoas please read:

> Is what we do in our app

This PR is the transfer of our app specific implementation to the core.

lorenzo:
I think the idea is great. After thinking a bit about this, I think that a better solution is having a `SlowQueriesLog` logger class instead. The configuration for that database could choose between the default logger or this one (or maybe both?)

markstory:
@lorenzo Being able to send slow queries to a different log file would be nice.

src/Database/Log/SlowQueryLogger.php
```diff
@@ -0,0 +1,88 @@
...
+ * @since         3.5.0
+ * @license       http://www.opensource.org/licenses/mit-license.php MIT License
+ */
+namespace Cake\Database\Log;
+
+use Cake\Core\InstanceConfigTrait;
+use Cake\Log\Log;
+
+/**
+ * Slow Query logger
```
markstory: It would be good to include a short usage example as most people will not be able to figure out how to use this as is.

robertpustulka:
@markstory @lorenzo How about the ability to configure multiple loggers for queries? One could be configured to log slow queries to `slow_queries` log, other could log all queries somewhere else.

src/Database/Log/SlowQueryLogger.php
burzum: Feel free to add it then. :)

burzum:
@robertpustulka if you want that feature do it in a follow up PR?

robertpustulka:
@burzum this is my suggestion to what @markstory and @lorenzo have been discussing. Configuring one logger to log queries to different logs based on some conditions seems like a bit too much for it to manage.

src/Database/Log/SlowQueryLogger.php
markstory: Features aren't really complete without documentation, as very few people will ever discover or use the feature which reduces its value and utility.

markstory:
Closing as I'm not sure we should be encouraging people to use enable query logging in non-development environments. This code could also live in a plugin very easily.

PR #10638, opened 2017-05-10
Adding getters and setters for logger objects

burzum:
Added getters and setters for two classes.

### DIFF
--- a/src/Database/Connection.php
+++ b/src/Database/Connection.php
@@ -826,11 +826,13 @@ class Connection implements ConnectionInterface
      * Sets a logger
      *
      * @param object $logger Logger object
-     * @return void
+     * @return $this
      */
     public function setLogger($logger)
     {
         $this->_logger = $logger;
+
+        return $this;
     }
 
     /**
--- a/src/Database/Connection.php
+++ b/src/Database/Connection.php
@@ -810,6 +810,8 @@ class Connection implements ConnectionInterface
 
     /**
      * {@inheritDoc}
+     *
+     * @deprecated 3.5.0 Use getLogger() and setLogger() instead.
      */
     public function logger($instance = null)
     {
--- a/src/Database/Connection.php
+++ b/src/Database/Connection.php
@@ -823,7 +823,7 @@ class Connection implements ConnectionInterface
     /**
      * Sets a logger
      *
-     * @param object Logger object
+     * @param object $logger Logger object
      * @return void
      */
     public function setLogger($logger)
--- a/src/Database/Connection.php
+++ b/src/Database/Connection.php
@@ -814,13 +814,35 @@ class Connection implements ConnectionInterface
     public function logger($instance = null)
     {
         if ($instance === null) {
-            if ($this->_logger === null) {
-                $this->_logger = new QueryLogger();
-            }
-
-            return $this->_logger;
+            return $this->getLogger();
         }
-        $this->_logger = $instance;
+
+        $this->setLogger($instance);
+    }
+
+    /**
+     * Sets a logger
+     *
+     * @param object Logger object
+     * @return void
+     */
+    public function setLogger($logger)
+    {
+        $this->_logger = $logger;
+    }
+
+    /**
+     * Gets the logger object
+     *
+     * @return object logger instance
+     */
+    public function getLogger()
+    {
+        if ($this->_logger === null) {
+            $this->_logger = new QueryLogger();
+        }
+
+        return $this->_logger;
     }
 
     /**
--- a/src/Database/Log/LoggingStatement.php
+++ b/src/Database/Log/LoggingStatement.php
@@ -124,7 +124,7 @@ class LoggingStatement extends StatementDecorator
     /**
      * Sets a logger
      *
-     * @param object Logger object
+     * @param object $logger Logger object
      * @return void
      */
     public function setLogger($logger)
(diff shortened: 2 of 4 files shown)


src/Database/Connection.php
```diff
@@ -814,13 +814,35 @@ public function logQueries($enable = null)
     public function logger($instance = null)
     {
         if ($instance === null) {
-            if ($this->_logger === null) {
```
dereuromark: This is Inside the new method. Looks BC.

burzum: Can I write the doc block like this?

```php
    /**
     * {@inheritDoc}
     * @@deprecated 3.5.0 Use getLogger() and setLogger() instead.
     */
```

cleptric: Ahh, overlooked that 😇
@burzum minus one @ and an added blank line after `{@inheritDoc}`, yes 😄

src/Database/Log/LoggingStatement.php
```diff
@@ -115,9 +115,30 @@ public function bindValue($column, $value, $type = 'string')
...
     {
         if ($instance === null) {
-            return $this->_logger;
+            return $this->getLogger();
         }
 
         return $this->_logger = $instance;
     }
+
+    /**
```
burzum: It's there? I'm blind? What exactly is it complaining about?

src/Database/Connection.php
```diff
@@ -810,17 +810,41 @@ public function logQueries($enable = null)
...
+
+        $this->setLogger($instance);
+    }
 
-            return $this->_logger;
+    /**
+     * Sets a logger
+     *
+     * @param object $logger Logger object
+     * @return void
```
robertpustulka: Can you return `$this` for method chaining?

robertpustulka (changes requested):

src/Database/Log/LoggingStatement.php
```diff
@@ -109,15 +109,37 @@ public function bindValue($column, $value, $type = 'string')
...
+     * Sets a logger
+     *
+     * @param object $logger Logger object
+     * @return void
+     */
+    public function setLogger($logger)
+    {
+        $this->_logger = $logger;
```
markstory: Should return $this here.

src/Database/Connection.php
ravage84: @burzum personally, I prefer to have blank line after inheriting, as it makes it more visible that I inherit and what I overwrite/add.

src/Database/Log/LoggingStatement.php
```diff
@@ -109,15 +109,37 @@ public function bindValue($column, $value, $type = 'string')
      * Sets the logger object instance. When called with no arguments
      * it returns the currently setup logger instance
      *
+     * @deprecated 3.5.0 Use getLogger() and setLogger() instead.
```
ravage84: Can we set ``@deprecated`` to the end of the annotations?

burzum: It's on top in other places!?

src/Database/Log/LoggingStatement.php
burzum: Done.

src/Database/Log/LoggingStatement.php
markstory: Or it could not matter :smile:

src/Database/Log/LoggingStatement.php
markstory: I don't see it changed here. I see it changed in `Connection`.

src/Database/Log/LoggingStatement.php
```diff
@@ -109,15 +109,37 @@ public function bindValue($column, $value, $type = 'string')
...
+     */
+    public function setLogger($logger)
+    {
+        $this->_logger = $logger;
+    }
+
+    /**
+     * Gets the logger object
+     *
+     * @return object logger instance
```
robertpustulka: Shouln't that be `@return Cake\Database\Log\QueryLogger` ?

src/Database/Log/LoggingStatement.php
dereuromark: As Long as we dont have a sniffer for it to automate.

burzum: I really don't care, my point is that it is a common PITA that one person tells you to do it this way, then somebody else comes in tells you the opposite. So I would like to get phpcs to detect it and enforce you to do it right or at least a wiki page for reference to look it up.

dereuromark: I agree, before there are no clear sniffers/guidelines it is not worth it.
And I agree, the order it currently is is fine.
```
@other
@param
@throws
@return
```
usually

ravage84: I agree, we should have a) section about the order of the doc block annotations in our [coding standards page in the CookBook](https://book.cakephp.org/3.0/en/contributing/cakephp-coding-conventions.html#commenting-code) and b) a sniff that can bug people, so I don't need to :trollface: 

@dereuromark I just searched for ``@deprecated`` in the 3.x code and my findings (samples) disagree with your order.

It seems the following is used most often (which I also remember we used normally):

````
@param
@return
@throws
@other
````

[Search on GitHub repo for "deprecated"](https://github.com/cakephp/cakephp/search?p=3&q=deprecated&type=&utf8=%E2%9C%93).

But we _could_ use the yet to be written sniff to tell us which order results in more violations in the current code and we then use the one that results in the least violations.
Though, I still prefer the ordering above, as it makes more sense to me (order of importance & of usage).

src/Database/Log/LoggingStatement.php
markstory: I'll fix these type errors up post merge.

Issue #10632, opened 2017-05-08
Deprecate RepositoryInterface::alias in favor of get/set


steinkel:
This is a (multiple allowed):

* [ ] bug
* [x] enhancement
* [x] feature-discussion (RFC)

* CakePHP Version: dev-master

### What you expected to happen

Following the changes done in \Cake\ORM\Table to mark alias method as deprecated, it could make sense to deprecate https://github.com/cakephp/cakephp/blob/master/src/Datasource/RepositoryInterface.php#L30 and require get/set methods?

markstory:
Sounds like a good idea. We can't add new methods to interface classes as doing so breaks compatibility. But we can deprecate the existing method and add doc blocks for the new methods.

cleptric:
PR is open

PR #10639, opened 2017-05-10
Add deprecation annotation to RepositoryInterface::alias

cleptric:
See #10632. As Table::alias was deprecated in 3.4, I added 3.4 to the interface as well. Is this correct?

### DIFF
--- a/src/Datasource/RepositoryInterface.php
+++ b/src/Datasource/RepositoryInterface.php
@@ -24,6 +24,7 @@ interface RepositoryInterface
     /**
      * Returns the table alias or sets a new one
      *
+     * @deprecated 3.4.0 Use setAlias()/getAlias() instead.
      * @param string|null $alias the new table alias
      * @return string
      */
--- a/src/Datasource/RepositoryInterface.php
+++ b/src/Datasource/RepositoryInterface.php
@@ -24,7 +24,6 @@ interface RepositoryInterface
     /**
      * Returns the table alias or sets a new one
      *
-     * @deprecated 3.4.0 Use setAlias()/getAlias() instead.
      * @param string|null $alias the new table alias
      * @return string
      */
--- a/src/Controller/Component/PaginatorComponent.php
+++ b/src/Controller/Component/PaginatorComponent.php
@@ -171,7 +171,7 @@ class PaginatorComponent extends Component
             $object = $query->repository();
         }
 
-        $alias = $object->alias();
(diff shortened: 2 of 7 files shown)


robertpustulka:
There are no setAlias/getAlias methods on the interface so I'm not sure about this change.

cleptric:
@robertpustulka Of course not :) But I understand @markstory over in the linked issue as we should add the annotation here, too.

robertpustulka:
We should add deprecation to all deprecated methods across several interfaces as well.

How would this behave in custom classes that implement these interfaces? IDEs will probably ignore the annotation if the docblock is overwritten?

src/Datasource/RepositoryInterface.php
```diff
@@ -24,6 +24,7 @@
     /**
      * Returns the table alias or sets a new one
      *
+     * @deprecated 3.4.0 Use setAlias()/getAlias() instead.
```
markstory: We should add a doc string annotation for the new methods. And implement those methods where possible.

cleptric: I'm sorry, but I don't know exactly what to do here regarding the doc annotation 😥

cleptric:
I updated all the places where `RepositoryInterface::alias` was used. I had to change some tests inside `PaginatorComponentTest`, as the mock would be created with the old method.

markstory:
Thank you for updating the tests.

robertpustulka:
@cleptric  why has the annotation has been removed? Now the PaginatorComponent will break for custom implementations of `RepositoryInterface`. There's no indication that the `alias()` method has been deprecated in the interface and what to use instead.

We should probably do what @markstory suggested: add annotations for new methods to the docblock and deprecate `alias()`  at least.

cleptric:
> Now the PaginatorComponent will break 
`alias()` wasn't removed, so why should something break now?
> add annotations for new methods to the docblock and deprecate alias() at least.
Yes, but I still don't know exactly what this means. I guess my question on @markstory's review was overlooked.
> Also the PaginatorComponentTest should use RepositoryInterface mocks...
Yes, good idea 👍

robertpustulka:
> alias() wasn't removed, so why should something break now?

But you removed it from the component and replaced with `getAlias()` :). Now every custom implmenentation of `RepositoryInterface` which does not have `getAlias()` method implemented will break in pagination.

Issue #8155, opened 2016-02-01
3.3: AppView needs to be non-optional


dereuromark:
We already ship with an https://github.com/cakephp/app/blob/master/src/View/AppView.php
but it seems we do not require it.
This being optional is a major problem with plugin development and the current way of documenting/recommending AppView for loading helpers.

Refs https://github.com/FriendsOfCake/CakePdf/pull/146
### tl;dr

Either we revert the current recommendation and declare helpers in controller again, or we need to make AppView always existent, so we can rely on it and extend it to leverage the helper config there for on-top View classes that need this information.
And all plugins should extend AppView (not View) then, if they need the helper information (which they usually do).

ADmad:
Making `AppView` required in 3.3 sounds good to me. We missed the boat for 3.2.

jhli:
Could we not patch the app and add a view loader in appcontroööer that defaults to appview or do I miss the point?

ADmad:
> ..or do I miss the point?

You do. Problem is not class loading but class inheritance.

renan:
This will kind of force applications to *always* use `App` namespace. I don't think that is a desired side-effect.

dereuromark:
@renan Don't we have that constraint with other classes as well?
Whats a solution?

markstory:
I don't think requiring AppView is that great of an idea. Custom namespaces will become basically impossible if we let plugins start making assumptions about the shape of the app classes.

I also think solving these problems through inheritance might not be the best design. How else could we solve the underlying problem?

ADmad:
Hmm I overlooked the namespacing issue. I guess the alternative would be promoting use of composition instead of inheritance in custom views.

burzum:
Couldn't we fire an event like `View.initialize`?

dereuromark:
I think that kind of event would be fine, too.
But we need to make sure that the helpers for example are passed in with ease.

dereuromark:
So what's the verdict? What kind of events are desired instead of the AppView itself? :)

burzum:
I'll implement them if we're OK to do so.

markstory:
We could start with a single event `View.initialize`.

cleptric:
As the `View.initialize` event hasn't archived the goal of this Issue, I reopen this to discuss further actions needed.

dereuromark:
Any ideas here? Otherwise we can also close this as "not possible to solve".

PR #10640, opened 2017-05-10
Add View.initialize event

cleptric:
Dispatch a new `View.initialize` event when a new View instance was created as discussed in #8155
I'm under the impression I missed sth here, so RFC :)

### DIFF
--- a/src/View/View.php
+++ b/src/View/View.php
@@ -347,6 +347,7 @@ class View implements EventDispatcherInterface
         }
         $this->Blocks = new ViewBlock();
         $this->initialize();
+        $this->dispatchEvent('View.initialize');
         $this->loadHelpers();
     }
 
--- a/src/View/View.php
+++ b/src/View/View.php
@@ -315,6 +315,7 @@ class View implements EventDispatcherInterface
      * @param \Cake\Event\EventManager|null $eventManager Event manager instance.
      * @param array $viewOptions View options. See View::$_passedVars for list of
      *   options which get set as class properties.
+     * @triggers View.initialize $this
      */
     public function __construct(
         ServerRequest $request = null,
--- a/tests/TestCase/View/ViewTest.php
+++ b/tests/TestCase/View/ViewTest.php
@@ -21,6 +21,7 @@ use Cake\Core\Configure;
 use Cake\Core\Plugin;
 use Cake\Event\Event;
 use Cake\Event\EventListenerInterface;
+use Cake\Event\EventManager;
 use Cake\Http\ServerRequest;
 use Cake\Routing\Router;
 use Cake\TestSuite\TestCase;
@@ -2089,5 +2090,23 @@ TEXT;
         $View->viewPath = 'mypath';
         $this->assertEquals('mypath', $View->templatePath());
         $this->assertEquals('mypath', $View->templatePath);
+    }
+
+    /**
+     * Test that creating a view fires the initialize event.
+     *
+     * @return void
+     */
+    public function testInitializeEvent()
+    {
+        $count = 0;
+        $callback = function (Event $event) use (&$count) {
+            $count++;
+        };
+        EventManager::instance()->on('View.initialize', $callback);
+        $view = new TestView();
+
+        $this->assertEquals(1, $count, 'Callback should be called');
+        EventManager::instance()->off('View.initialize', $callback);
     }
 }


robertpustulka:
Looks OK to me.

markstory:
Looks good to me, just needs some documentation added.

chinpei215:
Looks good to me. But  I think we have to remove AppView from our skeleton app. Otherwise, people don't stop loading helpers in the `initialize()` method. However, plugin authors cannot rely on the class. It would mean that @dereuromark cannot close https://github.com/FriendsOfCake/CakePdf/pull/146 even after this PR merged.

markstory:
How would removing AppView work? We would still need to maintain backwards compatibility, and where  would developers load helpera?

chinpei215:
I thought we were going to change how to load helpers. If `View.initialize` isn't the event to load helpers, why do we add this event?
As for backwards compatibility, I think we can remove the class any time, as it is in the skeleton app. Even if we remove the class, it wouldn't break user applications.

ADmad:
@chinpei215 Even with the event fired you still need to setup a listener for that event. Removing AppView would mean forcing users to create a listener themselves. I don't see how that's a good thing.

chinpei215:
Then what is solved by this PR? Even now, people can add some helper for all views if they want:
```php
// In some controller
public function beforeRender(Event $event)
{
    $this->viewBuilder()->setHelpers(['Tools.AuthUser']);
    parent::beforeRender($event);
}
```

Also, I noticed an issue introduced by `View.initialize`. If people add some non-existent helper in the event, `ExceptionRenderer` fails to render the `MissingHelperException`, as the non-existent helper is added again after it has been removed by the [_outputMessageSafe](https://github.com/cakephp/cakephp/blob/7bd33ddc3ddabf80f5769ae38dd8c5361573f5fa/src/Error/ExceptionRenderer.php#L359) method. So now I think `Controller.beforeRender` is more suitable place to load helpers.

How does this new event help people? If it is for very rare use cases, dispatching the event for each request may be not a good idea

robertpustulka:
This event could be helpful for DI.

```php
EventManager::instance()->on('View.initialize', function($e, $v) use ($foo) {
    if ($v instanceof FooAwareInterface) {
        $v->setFoo($foo);
    }
});

chinpei215:
Thank you. If it is helpful then I am OK. I think people can inject any dependencies to views via helpers even now, but they would have to write little longer code than the example.

robertpustulka:
@chinpei215 I'm not sure I'd do it using events but I think that allowing to hook into creation of objects using events can be helpful since we don't provide any DIC.

cleptric:
@chinpei215 Really good catch 👍

chinpei215:
I understand. Thank you, everyone.

cleptric:
How should we proceed now? The original goal of creating a new way of loading helpers isn't fulfilled. Do we want to keep that event or should I close this PR?

robertpustulka:
The listeners for this event have to be added in a global manager. The event is fired in a constructor so there's no other way to hook into it. It's not very elegant solution to load helpers in a global context.

cleptric:
@robertpustulka I wouldn't recommend loading helpers in this event regarding the point @chinpei215 pointed out. Originally I was inspired by the way `Table` dispatches this event.
Do you maybe have a better idea? Fell free to add it here 😀

robertpustulka:
@cleptric the event itself is not bad thing. But not for loading helpers :)

cleptric:
Ok, so we have to discuss another way to archive the load helpers thing then.

robertpustulka:
@cleptric what's bad about custom `View` classes like `AppView` etc or loading helpers at runtime?

cleptric:
The goal of this PR was to find a way how plugins can reliable load helpers. First idea was to make the `AppView` none optional. As this seemed odd, the idea was to do it with a new event.
It turns out that this isn't the best solution either.

robertpustulka:
@cleptric I'm not sure the idea that plugins can hack views and load some helpers is right. A developer should decide which helpers to load, where and when.

dereuromark:
@robertpustulka If they provide their own controllers and view, they **must** be able to do that, this is not hacking, but provoding basic functionality for that plugin to work when loaded incl. routing.

robertpustulka:
@dereuromark so plugins can load helpers inside `initialize()` method of the controllers/views they provide.

dereuromark:
If we deprecate the controller injection here for helpers, then it must be the AppView, this is why I said in the linked issue that this class should be considered a must have for projects in order for plugins to rely on it :) The other way would be the said extension point via hook or event.

robertpustulka:
@dereuromark are you saying that the plugins should depend on `AppView`? Still, while developing a plugin you can't be sure how the client app is structured. There may be some other base controllers and views other controllers depend on. Plugin controllers and views should be standalone and independent of app structure.

If you want to provide tools with a plugin let them be decorators like components, helpers and behaviors, not concrete controllers, views and model. IMO plugins should be developed with composition over inheritance principle in mind.

dereuromark:
So far we always require AppController - why not the same for AppView if it takes over some basics from the controller layer that is more relevant for view layer? : )

I dont care either way, I just think that a basic AppView could make sense to require.
Whereas decoration sure could be also a nice future-way of doing things.

robertpustulka:
Ok then I'm :+1: for including `AppView` for consistency.

cleptric:
I guess we are loosing the scope a little bit 😀 

>**markstory**: I don't think requiring AppView is that great of an idea. Custom namespaces will become basically impossible if we let plugins start making assumptions about the shape of the app classes.
I also think solving these problems through inheritance might not be the best design. How else could we solve the underlying problem?

dereuromark:
@cleptric Not necessarily. Inside the current version there is no custom namespace anyway, since everything requires the App namespace for AppController afaik.
Thus it would be perfectly fine to also require the view class - unless we are resolving this in the future - which I didnt know there was a plan to do so.

robertpustulka:
@cleptric `AppView` is OK for a place to configure helpers in your app. Still not good for plugins to hook into, so I agree with @markstory. You shouldn't guess how a client app is designed while developing a plugin.

The other think I haven't thought about is custom namespaces. Plugin view extending `\App\View\AppView` would break if your root namespace is `\MyApp` for example.

inoas:
I would love to see a better, non-configuration-hell-way than inheritance for App* classes, that works with sophisticated apps but also allows XP and has linear/non-event flow by default and allows newbies to get on board and understand quickly.

markstory:
Has anyone looked at how other frameworks approach this problem? What doesn't work with the `beforeRender` solution pointed out earlier by @chinpei215 ?  The issues with outputMessageSafe coild be addressed by using a view class that doesn't trigger this new event.

markstory:
@dereuromark when we added namespaces the community wanted the ability to control that namespace, so because of that we have to maintain compatibility for custom namespaces.

cleptric:
Can I close this? Or is there any need for this new event?

dereuromark:
I dont think the issue is resolved yet.

PR #10642, opened 2017-05-11
Added the ability to serialize Table objects.

robertpustulka:
The connection name is stored for serialization and reinstantated on wakeup.

I'm not sure if this is done right so the suggestions are welcome.

Targetting this for next 3.4.7 because for me this has always been a bug that I couldn't serialize tables.
This also shouldn't break anything Let's see if tests pass on all adapters.

### DIFF
--- a/src/ORM/Table.php
+++ b/src/ORM/Table.php
@@ -20,6 +20,7 @@ use Cake\Core\App;
 use Cake\Database\Schema\TableSchema;
 use Cake\Database\Type;
 use Cake\Datasource\ConnectionInterface;
+use Cake\Datasource\ConnectionManager;
 use Cake\Datasource\EntityInterface;
 use Cake\Datasource\Exception\InvalidPrimaryKeyException;
 use Cake\Datasource\RepositoryInterface;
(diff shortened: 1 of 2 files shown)


robertpustulka:
Whats the issue with travis?

cleptric:
`Could not authenticate against github.com` 😞

dereuromark:
Master is broken since https://travis-ci.org/cakephp/cakephp/builds/230162397

robertpustulka:
Is it going to be fixed? :)

lorenzo:
What's the use case for this?

robertpustulka:
I'm storing some objects in a queue. Some of them have tables as dependencies. I know I could refactor my code to reload the table on wakeup but this would mean I'd also have to handle table locator etc...

In the long term fixing the Table seems like a better idea. Also I noticed some issues with DebugKit and  serializing PDO instances some time ago. I'm not sure if that's related though.

dereuromark:
Tables are considered state less classes IMO (vs Entities), thus serializing them and stuff seems more like an anti-pattern to me.
Instead the stateful data should be in the entity and the table should work on those (and read out the data it needs).

robertpustulka:
@dereuromark you're right probably. I have a kind of a stateful object with a table dependency. I'll refactor the code eventually but I also think that including this fix wouldn't hurt :)

lorenzo:
@robertpustulka serialising for a queue sounds weird and error prone. But regardless of that, I think this could be leveraged to bootstrap the ORM faster after a second request. Maybe we could cache the tables definitions so they don't have to be loaded again?

robertpustulka:
@lorenzo Interesting idea. Would that caching take place in TableRegistry or inside table locators?

chinpei215:
Sorry if I am mistaken, but I think it is difficult to serialize tables correctly. For example, the following expression must return `true`, even after unserialization:
```php
$this->Bookmarks === $this->Tags->Bookmarks->getTarget();
```

Serializing the event manager also would be complicated. Closures may be attached to it. `Model.initialize` events would never be called.

robertpustulka:
@chinpei215 you're right. Serialization of table objects just doesn't make sense.

I'm closing thsi PR.

PR #10644, opened 2017-05-11
Avoid logging schema queries based on Configure parameter

siran:
It's many time an inconvenience to have Schema Queries on DebugKit's SqlLog

This is a tiny fix to avoid logging those queries based on a App.logSchemaQueries Configure parameter.

### DIFF
--- a/src/ORM/Table.php
+++ b/src/ORM/Table.php
@@ -17,9 +17,11 @@ namespace Cake\ORM;
 use ArrayObject;
 use BadMethodCallException;
 use Cake\Core\App;
+use Cake\Core\Configure;
 use Cake\Database\Schema\TableSchema;
 use Cake\Database\Type;
 use Cake\Datasource\ConnectionInterface;
+use Cake\Datasource\ConnectionManager;
 use Cake\Datasource\EntityInterface;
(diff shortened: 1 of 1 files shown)


siran:
I think my commit didn't introduce that error... Or I don't even understand why or what happened.

robertpustulka:
Is this supposed to work only on default connection?

siran:
you are right, it should be generalized to the 'active' connection... if the change is agreed upon, it'd be easy to update

src/ORM/Table.php
```diff
@@ -524,13 +527,19 @@ public function connection(ConnectionInterface $connection = null)
      */
     public function getSchema()
     {
+        $logQueries = ConnectionManager::get('default')->logQueries();
```
markstory: Many applications have more than just one connection. Couldn't this schema logging be done at a lower level? Perhaps in the SchemaCollection instead?

robertpustulka:
I also think that should be configured using QueryLogger not in some global config key.

burzum:
This could be done using the query logger from this PR https://github.com/cakephp/cakephp/pull/10634 that supports a filter callback in which you could simply filter out the schema queries.

markstory:
Moved to a DebugKit issue as I think this is better solved there.

PR #10646, opened 2017-05-12
3.next - Add application routes hook

markstory:
This finishes off the bulk of the work for #10308. I'm expanding the scope of the `Application` class to be a vessel for configuration/setup logic. This change also enables us to deprecate most of the static methods on `Router` (in a subsequent pull request) which will let us significantly reduce the static API surface area in CakePHP.

## TODO

* [x] Test these changes against an 3.2 style app to ensure backwards compatibility for route reloading.
* [ ] Should `RouteBuilder` have an interface added?

### DIFF
--- a/src/Http/ServerRequestFactory.php
+++ b/src/Http/ServerRequestFactory.php
@@ -100,6 +100,10 @@ abstract class ServerRequestFactory extends BaseFactory
             $uri = static::updatePath($base, $uri);
         }
 
+        if (!$uri->getHost()) {
+            $uri = $uri->withHost('localhost');
+        }
+
         // Splat on some extra attributes to save
         // some method calls.
         $uri->base = $base;
--- a/src/Routing/Middleware/RoutingMiddleware.php
+++ b/src/Routing/Middleware/RoutingMiddleware.php
@@ -14,9 +14,11 @@
  */
 namespace Cake\Routing\Middleware;
 
+use Cake\Http\BaseApplication;
 use Cake\Http\MiddlewareQueue;
 use Cake\Http\Runner;
 use Cake\Routing\Exception\RedirectException;
+use Cake\Routing\RouteBuilder;
 use Cake\Routing\Router;
 use Psr\Http\Message\ResponseInterface;
 use Psr\Http\Message\ServerRequestInterface;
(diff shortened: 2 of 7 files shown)


src/Routing/Router.php
```diff
@@ -926,6 +911,27 @@ public static function parseNamedParams(ServerRequest $request, array $options =
     }
 
     /**
+     * Create a RouteBuilder for the provided path.
+     *
+     * @param string $path The path to set the builder to.
+     * @param array $options The options for the builder
+     * @return \Cake\Routing\RouteBuilder
+     */
+    public static function getRouteBuilder($path, array $options = [])
```
lorenzo: Can we rename this to `createRouteBuilder` or `newRouteBuilder`, so that it does not give the impression it is the same one every time?

tests/TestCase/Routing/RouterTest.php
```diff
@@ -2035,7 +2037,6 @@ public function testGenerationWithSslOption()
         Router::connect('/:controller/:action/*');
 
         $request = new ServerRequest();
-        $request->env('HTTP_HOST', 'localhost');
```
lorenzo: Nice!

lorenzo:
This looks great!

src/Routing/Middleware/RoutingMiddleware.php
```diff
@@ -27,12 +30,39 @@
...
 class RoutingMiddleware
 {
+    /**
+     * Constructor
+     *
+     * @param \Cake\Http\BaseApplication $app The application instance that routes are defined on.
+     */
+    public function __construct(BaseApplication $app = null)
+    {
+        $this->app = $app;
```
ADmad: The property should be declared, makes phpstan sad without it.

ADmad (changes requested):

src/Routing/Router.php
markstory: Sure.

src/Routing/Middleware/RoutingMiddleware.php
markstory: Whoops.

src/Http/ServerRequestFactory.php
```diff
@@ -100,6 +100,10 @@ public static function marshalUriFromServer(array $server, array $headers)
             $uri = static::updatePath($base, $uri);
         }
 
+        if (!$uri->getHost()) {
+            $uri = $uri->withHost('localhost');
```
robertpustulka: :+1:

Issue #9824, opened 2016-12-01
Add support for exception chaining


chinpei215:
This is a (multiple allowed):

* [ ] bug
* [x] enhancement
* [x] feature-discussion (RFC)

* CakePHP Version: 3.3.9 / 2.9.3

### What you did

Chained exceptions:
```php
try {
    throw new \Exception('Foo');
} catch (\Exception $e) {
    throw new \Exception('Bar', 0, $e);
}
```

### What happened

The chained exceptions were not rendered/logged:
```
2016-12-01 18:27:49 Error: [Exception] Bar
Request URL: /somethings/
Client IP: 127.0.0.1
Stack Trace:
#0 [internal function]: App\Controller\SomethingsController->index()
#1 C:\var\www\cakephp3.dev\vendor\cakephp\cakephp\src\Controller\Controller.php(435): call_user_func_array(Array, Array)
#2 C:\var\www\cakephp3.dev\vendor\cakephp\cakephp\src\Http\ActionDispatcher.php(122): Cake\Controller\Controller->invokeAction()
#3 C:\var\www\cakephp3.dev\vendor\cakephp\cakephp\src\Http\ActionDispatcher.php(96): Cake\Http\ActionDispatcher->_invoke(Object(App\Controller\SomethingsController))
#4 C:\var\www\cakephp3.dev\vendor\cakephp\cakephp\src\Routing\Dispatcher.php(60): Cake\Http\ActionDispatcher->dispatch(Object(Cake\Network\Request), Object(Cake\Network\Response))
#5 C:\var\www\cakephp3.dev\webroot\index.php(37): Cake\Routing\Dispatcher->dispatch(Object(Cake\Network\Request), Object(Cake\Network\Response))
#6 {main}
```

### What you expected to happen

The chained exceptions are rendered/logged, like the PHP built-in exception handler:
```
Fatal error: Uncaught exception 'Exception' with message 'Foo' in C:\var\sandbox\test.php:3
Stack trace:
#0 {main}

Next exception 'Exception' with message 'Bar' in C:\var\sandbox\test.php:5
Stack trace:
#0 {main}
  thrown in C:\var\sandbox\test.php on line 5
```

I would like to add the new third argument `$previous` to all existing core exception classes, and to fix `ErrorHandler` and several core view templates (such as `exception_stack_trace.ctp`) to render chained exceptions correctly.

Now I am targeting 3.4.0 and 2.10.0 (Next minor versions).

[...]

PR #10649, opened 2017-05-12
Add the $previous argument to all exceptions

chinpei215:
This pull request is the first step to support exception chaining (#9824).

While adding an additional argument to the constructor is usually quite backward compatible, but I would like to target 3.next branch, as this is a little large change.

This is a work in progress, as I am not sure how I should write related test cases. I was going to write test cases for each effected exception class, like the following:
```php
class MissingActionExceptionTest extends TestCase
{

    /**
     * Test the getPrevious() method works.
     *
     * @return void
     */
    public function testGetPrevious()
    {
        $previous = new Exception();
        $exception = new MissingActionException('message', -1, $previous);
        $this->assertSame('message', $exception->getMessage());
        $this->assertSame(-1, $exception->getCode());
        $this->assertSame($previous, $exception->getPrevious());
    }

    /**
     * Test the getPrevious() method works for PHP7 Error
     */
    public function testGetPreviousError() {
        $this->skipIf(version_compare(PHP_VERSION, '7.0.0', '<'));

        $previous = new Error();
        $exception = new MissingActionException('message', -1, $previous);
        $this->assertSame('message', $exception->getMessage());
        $this->assertSame(-1, $exception->getCode());
        $this->assertSame($previous, $exception->getPrevious());
    }
}
```

But should I create something like a `ExceptionsTest` and test all exceptions there instead?

### DIFF
--- a/tests/TestCase/ExceptionsTest.php
+++ b/tests/TestCase/ExceptionsTest.php
@@ -1,0 +1,185 @@
+<?php
+/**
+ * ExceptionsTest file
+ *
+ * CakePHP(tm) : Rapid Development Framework (https://cakephp.org)
+ * Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
+ *
+ * Licensed under The MIT License
+ * For full copyright and license information, please see the LICENSE.txt
+ * Redistributions of files must retain the above copyright notice
+ *
+ * @copyright     Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
+ * @link          https://cakephp.org CakePHP(tm) Project
+ * @since         3.5.0
+ * @license       https://opensource.org/licenses/mit-license.php MIT License
+ */
+namespace Cake\Test\TestCase;
+
+use Cake\Error\FatalErrorException;
+use Cake\Error\PHP7ErrorException;
+use Cake\ORM\Entity;
+use Cake\ORM\Exception\PersistenceFailedException;
+use Cake\TestSuite\TestCase;
+use Error;
+use Exception;
+
+class ExceptionsTest extends TestCase
+{
+
+    /**
+     * Tests simple exceptions work.
+     *
+     * @dataProvider exceptionProvider
+     * @param $class The exception class name
+     * @param $defaultCode The default exception code
+     * @return void
+     */
+    public function testSimpleException($class, $defaultCode)
+    {
+        $previous = new Exception();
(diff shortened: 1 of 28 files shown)


dereuromark:
Thanks for doing this!

chinpei215:
Thank you.

I forgot that I had one more question. Should we *overload* the constructor of the exceptions like the follwoing? 
```php
public function __construct($message, $code, $previous);
public function __construct($message, $previous);
```
If it is a sub class of the `HttpException`, the following code doesn't make sense:
```php
catch (Exception $e) {
    throw new NotFoundException('message', 404, $e);
}
```

Or should we create a static factory for the purpose instead?
```php
public static function withPrevious($message, $previous);
```

markstory:
> But should I create something like a ExceptionsTest and test all exceptions there instead?

You could create a test case, that uses a data provider to ensure all the known exception classes implement the correct 'interface'.

> I forgot that I had one more question. Should we overload the constructor of the exceptions like the follwoing?

That sounds like it will add complexity that we could avoid. Instead of overloading could use use `null` as a way to avoid duplication around status codes?

```php
catch (Exception $e) {
    throw new NotFoundException('message', null, $e);
}
```

chinpei215:
Sorry, I totally forgot about this PR... Now I added test cases.

chinpei215:
Additional notable changes:
- Add _defaultCode property to Cake\Core\Exception\Exception. This allows people to pass $previous argument without $code.
- Now InvalidPrimaryKeyException, RecordNotFoundException, SocketException, RedirectException and XmlException are subclass of Cake\Core\Exception\Exception.

lorenzo:
Thanks for doing this @chinpei215 !

tests/TestCase/ExceptionsTest.php
```diff
@@ -0,0 +1,185 @@
...
+
+    /**
+     * Provides pairs of exception name and default code.
+     *
+     * @return array
+     */
+    public function exceptionProvider()
+    {
+        return [
+            ['Cake\Console\Exception\ConsoleException', 500],
```
markstory: Couldn't the default code be read from the class using reflection?

chinpei215: It could. But then, test cases will pass even if unexpected code is defined in the class.

markstory: Good point.

PR #10656, opened 2017-05-14
Split EntityTrait::invalid into getter and setter

cleptric:
Continues with #9978

### DIFF
--- a/src/Datasource/EntityTrait.php
+++ b/src/Datasource/EntityTrait.php
@@ -1036,12 +1036,74 @@ trait EntityTrait
     }
 
     /**
+     * Get a list of invalid fields and their data for errors upon validation/patching
+     *
+     * @return array
+     */
+    public function getInvalid()
+    {
+        return $this->_invalid;
+    }
+
+    /**
+     * Get a single value of an invalid field. Returns null if not set.
+     *
+     * @param string $field The name of the field.
+     * @return mixed
+     */
+    public function getInvalidField($field)
+    {
+        $value = isset($this->_invalid[$field]) ? $this->_invalid[$field] : null;
+
+        return $value;
+    }
+
+    /**
+     * Set fields as invalid and not patchable into the entity.
+     *
+     * This is useful for batch operations when one needs to get the original value for an error message after patching.
+     * This value could not be patched into the entity and is simply copied into the _invalid property for debugging purposes
+     * or to be able to log it away.
+     *
+     * @param array $fields The values to set.
+     * @param bool $overwrite Whether or not to overwrite pre-existing values for $field.
+     * @return $this
+     */
+    public function setInvalid(array $fields, $overwrite = false)
+    {
+        foreach ($fields as $field => $value) {
+            if ($overwrite === true) {
+                $this->_invalid[$field] = $value;
+                continue;
+            }
+            $this->_invalid += [$field => $value];
+        }
+
+        return $this;
+    }
+
+    /**
+     * Sets a field as invalid and not patchable into the entity.
+     *
+     * @param string $field The value to set.
+     * @param mixed $value The invalid value to be set for $field.
+     * @return $this
+     */
+    public function setInvalidField($field, $value)
+    {
+        $this->_invalid[$field] = $value;
+
+        return $this;
+    }
+
+    /**
      * Sets a field as invalid and not patchable into the entity.
      *
      * This is useful for batch operations when one needs to get the original value for an error message after patching.
      * This value could not be patched into the entity and is simply copied into the _invalid property for debugging purposes
      * or to be able to log it away.
      *
+     * @deprecated 3.5 Use getInvalid()/getInvalidField()/setInvalid() instead.
      * @param string|array|null $field The field to get invalid value for, or the value to set.
      * @param mixed|null $value The invalid value to be set for $field.
      * @param bool $overwrite Whether or not to overwrite pre-existing values for $field.
--- a/src/Datasource/InvalidPropertyInterface.php
+++ b/src/Datasource/InvalidPropertyInterface.php
@@ -17,6 +17,11 @@ namespace Cake\Datasource;
 /**
  * Describes the methods that any class representing a data storage should
  * comply with.
+ *
+ * @method array getInvalid()
+ * @method mixed getInvalidField($field)
+ * @method $this setInvalid($field, $value = null, $overwrite = false)
+ * @method $this setInvalidField($field, $value = null, $overwrite = false)
  */
 interface InvalidPropertyInterface
 {
@@ -31,6 +36,7 @@ interface InvalidPropertyInterface
      * @param mixed|null $value The invalid value to be set for $field.
      * @param bool $overwrite Whether or not to overwrite pre-existing values for $field.
      * @return $this|mixed
+     * @deprecated 3.5.0 Use getInvalid()/getInvalidField() and setInvalid()/setInvalidField() instead.
      */
     public function invalid($field = null, $value = null, $overwrite = false);
 }
--- a/tests/TestCase/ORM/EntityTest.php
+++ b/tests/TestCase/ORM/EntityTest.php
@@ -1573,4 +1573,61 @@ class EntityTest extends TestCase
         $this->assertTrue($cloned->dirty('a'));
         $this->assertTrue($cloned->dirty('b'));
     }
+
+    /**
+     * Tests getInvalid and setInvalid
+     *
+     * @return void
+     */
+    public function testGetSetInvalid()
+    {
+        $entity = new Entity();
+        $return = $entity->setInvalid([
+            'title' => 'albert',
+            'body' => 'einstein'
+        ]);
+        $this->assertSame($entity, $return);
+        $this->assertSame([
+            'title' => 'albert',
+            'body' => 'einstein'
+        ], $entity->getInvalid());
+
+        $overwrite = $entity->setInvalid([
+            'title' => 'nikola',
+            'body' => 'tesla'
+        ], true);
+        $this->assertSame($entity, $overwrite);
+        $this->assertSame([
+            'title' => 'nikola',
+            'body' => 'tesla'
+        ], $entity->getInvalid());
+    }
+
+    /**
+     * Tests getInvalidField
+     *
+     * @return void
+     */
+    public function testGetSetInvalidField()
+    {
+        $entity = new Entity();
+        $return = $entity->setInvalidField('title', 'albert');
+        $this->assertSame($entity, $return);
+        $this->assertSame('albert', $entity->getInvalidField('title'));
+
+        $overwrite = $entity->setInvalidField('title', 'nikola');
+        $this->assertSame($entity, $overwrite);
+        $this->assertSame('nikola', $entity->getInvalidField('title'));
+    }
+
+    /**
+     * Tests getInvalidFieldNull
+     *
+     * @return void
+     */
+    public function testGetInvalidFieldNull()
+    {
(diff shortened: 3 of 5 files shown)


robertpustulka:
Can we also start adding new methods to the interfaces via annotations?

Ie:
```php
/**
 * ...
 * @method $this setFoo(Foo $foo)
 * @method Foo getFoo()
 */
interface BarInterface {
    /**
     * @deprecated Use getFoo/setFoo
     */
    public function foo(Foo $foo = null);
}
```

src/Datasource/EntityTrait.php
```diff
@@ -1036,12 +1036,58 @@ protected function _readError($object, $path = null)
...
+     * This is useful for batch operations when one needs to get the original value for an error message after patching.
+     * This value could not be patched into the entity and is simply copied into the _invalid property for debugging purposes
+     * or to be able to log it away.
+     *
+     * @param string|array $field The value to set.
+     * @param mixed|null $value The invalid value to be set for $field.
+     * @param bool $overwrite Whether or not to overwrite pre-existing values for $field.
+     * @return $this|mixed
+     */
+    public function setInvalid($field, $value = null, $overwrite = false)
```
robertpustulka: How about splitting this method into two: one accepting an array and another accepting `$field` and `$value`.

src/Datasource/EntityTrait.php
```diff
@@ -1036,12 +1036,58 @@ protected function _readError($object, $path = null)
...
+    {
+        if (is_string($field) && $value === null) {
+            $value = isset($this->_invalid[$field]) ? $this->_invalid[$field] : null;
```
robertpustulka: This is actually a getter.

src/Datasource/InvalidPropertyInterface.php
```diff
@@ -17,6 +17,9 @@
 /**
  * Describes the methods that any class representing a data storage should
  * comply with.
+ *
+ * @method \Cake\Datasource\EntityTrait getInvalid()
```
robertpustulka: Should return `$this`. You can't return traits AFAIR.

src/Datasource/EntityTrait.php
```diff
@@ -1036,12 +1036,58 @@ protected function _readError($object, $path = null)
...
+
+            return $value;
```
robertpustulka: This looks like a getter.

src/Datasource/EntityTrait.php
```diff
@@ -1036,12 +1036,58 @@ protected function _readError($object, $path = null)
     }
 
     /**
+     * Get a list of invalid fields and their data for errors upon validation/patching
+     *
+     * @return array
+     */
+    public function getInvalid()
```
robertpustulka: Maybe this also should be split into `getInvalidFields(): array` and `getInvalidField($field): mixed`?

robertpustulka (changes requested):
The setter needs to be fixed.

src/Datasource/EntityTrait.php
cleptric: Does the EntityTrait really needs another method? 😄

src/Datasource/EntityTrait.php
cleptric: I just kept the behavior of the old method.

src/Datasource/EntityTrait.php
robertpustulka: @cleptric changing the method behavior based on argument type can be confusing. Also this would make a better transition to PHP7:
```php
public function setInvalid(array $fields): self;
public function setInvalidField(string $field, string $value): self
```

src/Datasource/InvalidPropertyInterface.php
cleptric: Do you have some docs for me on how to create this annotations correctly? I never used an IDE so I have no clue what is needed here.

robertpustulka: @cleptric you just type hint return types as regular.

```php
/**
 * @param string $foo
 * @return \Bar
 */
public function foo($bar);
```
Is same as:
```php
/**
 * @method \Bar foo(string $bar)
 */
```
See: https://phpdoc.org/docs/latest/references/phpdoc/tags/method.html

src/Datasource/EntityTrait.php
robertpustulka: Actually I'd make the second method name shorter:
```php
public function setInvalidFields(array $fields): self;
public function setInvalid(string $field, string $value): self
```

src/Datasource/InvalidPropertyInterface.php
cleptric: Ahh, I see :)
I mixed that up with this forwarding to classes that implement that method.
Should be `array` then.

src/Datasource/EntityTrait.php
```diff
@@ -1036,12 +1036,65 @@ protected function _readError($object, $path = null)
...
+    {
+        return $this->_invalid;
+    }
+
+    /**
+     * Get a single value of an invalid field. Returns null if not set.
+     *
+     * @param string $field The name of the field.
+     * @return mixed
```
dereuromark: Can this be more concrete?

cleptric: `setInvalid()/invalid()` currently allows you to set whatever value you want.

src/Datasource/EntityTrait.php
```diff
@@ -1046,6 +1046,19 @@ public function getInvalid()
     }
 
     /**
+     * Get a single value of an invalid field. Returns null if not set.
+     *
+     * @param string $field The name of the field.
+     * @return mixed
+     */
+    public function getInvalidField($field)
```
robertpustulka: @cleptric could you split a setter in the same fashion?

cleptric:
I split the setter into two separate methods. I'm a little bit confused about the fact that in the old `invalid` method, the `$value` parameter was always overwritten by `foreach ($field as $f => $value) {...}`. Did I miss something?

src/Datasource/EntityTrait.php
```diff
@@ -1059,29 +1059,48 @@ public function getInvalidField($field)
...
-        if (!is_array($field)) {
-            $field = [$field => $value];
-        }
-
-        foreach ($field as $f => $value) {
-            if ($overwrite) {
-                $this->_invalid[$f] = $value;
+        foreach ($fields as $field => $value) {
+            if ($overwrite === true) {
+                $this->_invalid[$field] = $overwriteValue;
```
robertpustulka: Why this? The value geta overwritten by an array value, not some other one.

src/Datasource/EntityTrait.php
```diff
@@ -1036,12 +1036,84 @@ protected function _readError($object, $path = null)
...
+     * or to be able to log it away.
+     *
+     * @param string $field The value to set.
+     * @param mixed|null $value The invalid value to be set for $field.
+     * @param bool $overwrite Whether or not to overwrite pre-existing values for $field.
+     * @return $this
+     */
+    public function setInvalidField($field, $value = null, $overwrite = false)
+    {
+        if ($overwrite === true) {
```
robertpustulka: I think that this method should always overwrite the value as this is what you generally want to do if you call a setter.

dereuromark: There is then no 2nd argument needed here. That should stay with the old deprecated method.

src/Datasource/EntityTrait.php
cleptric: See my comment above. The old parameter doesn't makes sense IMO.

src/Datasource/EntityTrait.php
cleptric: Do you mean the third? `$overwrite`?

src/Datasource/EntityTrait.php
```diff
@@ -1066,17 +1066,11 @@ public function getInvalidField($field)
...
      */
-    public function setInvalid(array $fields, $overwriteValue = null, $overwrite = false)
+    public function setInvalid(array $fields)
     {
         foreach ($fields as $field => $value) {
-            if ($overwrite === true) {
-                $this->_invalid[$field] = $overwriteValue;
-                continue;
-            }
             $this->_invalid += [$field => $value];
```
robertpustulka: Looks like you removed the overwrite feature entirely.

I think there should be overwrite support here disabled by default as before.

src/Datasource/EntityTrait.php
```diff
@@ -1091,17 +1085,12 @@ public function setInvalid(array $fields, $overwriteValue = null, $overwrite = f
...
      */
-    public function setInvalidField($field, $value = null, $overwrite = false)
+    public function setInvalidField($field, $value)
     {
-        if ($overwrite === true) {
-            $this->_invalid[$field] = $value;
-        } else {
-            $this->_invalid += [$field => $value];
-        }
+        $this->_invalid += [$field => $value];
```
robertpustulka: Same as `setInvalid()`. Now there's no way to overwrite a value at all.

This though should always overwrite the value. If you call a setter explicitly for a given field, you expect the value is set regardless.

robertpustulka:
IMO `setInvalidField()` should always overwrite the value.

On the other hand `setInvalid()` setter should support the `$overwrite` parameter.

Eg:
1. My current invalid state is: `['foo' => 'foo]`. 
2. I call `$e->setInvalid(['foo' => 'bar', 'bar' => 'baz'])` (no overwrite). 
3. I expect the state would be `['foo' => 'foo, 'bar' => 'baz']` afterwards.

tests/TestCase/ORM/EntityTest.php
```diff
@@ -1591,6 +1591,16 @@ public function testGetSetInvalid()
             'title' => 'albert',
             'body' => 'einstein'
         ], $entity->getInvalid());
+
+        $overwrite = $entity->setInvalid([
+            'title' => 'nikola',
+            'body' => 'tesla'
+        ], true);
```
robertpustulka: An assertion for `$overwrite === false` with the set state would be nice to make sure the state doesn't get overwritten.

src/Datasource/EntityTrait.php
```diff
@@ -1036,12 +1036,78 @@ protected function _readError($object, $path = null)
...
+
+        return $this;
+    }
+
+    /**
+     * Sets a field as invalid and not patchable into the entity.
+     *
+     * This is useful for batch operations when one needs to get the original value for an error message after patching.
+     * This value could not be patched into the entity and is simply copied into the _invalid property for debugging purposes
+     * or to be able to log it away.
```
markstory: This doc block doesn't really make sense for the single setter.

tests/TestCase/ORM/EntityTest.php
cleptric: I'm not sure what you mean by that 😅

robertpustulka: See my comment: https://github.com/cakephp/cakephp/pull/10656#issuecomment-304821099

cleptric: I already changed this. Was asking about the testcase 😀

robertpustulka: Yeah, but we need a test that it works as expected ie does not overwrite current state 😀

robertpustulka: @cleptric I've added an assertion for that (6f4f7ce) :smile:

robertpustulka (approved):

cleptric:
@robertpustulka Thank you for adding the tests. Builds are now also passing after a restarting them 😄

Issue #10651, opened 2017-05-12
3.5/4.0 Query Builder: Make contain() have a similar interface to matching(), leftJoinWith(), innerJoinWith()


inoas:
This is a (multiple allowed):

* [x] enhancement
* [x] feature-discussion (RFC)

* CakePHP Version: 3.4

### What you did
I am often using closures on contain, matching, innerJoinWith, leftJoinWith.

### What happened
I am often typing the interface wrong because:

- `contain(['AssocName' => function($q) {}])`
vs
- `matching('AssocName',  function($q) {})` et al.

### What you expected to happen
The interfaces to be the same.

Could this be changed for 4.0?

markstory:
Do we need to break compatibility to make both methods work similarly?

inoas:
From the outside it looks like there are two options that could be BC:
- allow `contain()` to take a string + optional closure in addition to the array syntax.
- allow all the other methods instead to take an array in addition to string + optional closure

markstory:
> allow contain() to take a string + optional closure in addition to the array syntax.

This sounds pretty reasonable given the current signature of `contain()`. Did you want to put a pull request together for that?

jeremyharris:
If you cannot @inoas I would be happy to.

inoas:
@jeremyharris so I am not sure if I can do it this week so in case you are faster, go ahead. Before attempting myself I will query/pm you on IRC/slack - so feel free to start any time!

saeideng:
`contain()` receive multi key/value  but `matching()` no

cleptric:
Closing this as we have an open PR

PR #10660, opened 2017-05-15
Added second signature for contain that matches the signature for mat…

jeremyharris:
Refs https://github.com/cakephp/cakephp/issues/10651

Adds a new signature for `contain()` that matches the signature of `matching()`:

```php
$query->contain('Articles', function ($q) {
    return $q->where(['Articles.published' => true]);
});
```

### DIFF
--- a/src/ORM/EagerLoader.php
+++ b/src/ORM/EagerLoader.php
@@ -126,12 +126,21 @@ class EagerLoader
      * @param array|string $associations list of table aliases to be queried.
      * When this method is called multiple times it will merge previous list with
      * the new one.
+     * @param callable|null $queryBuilder The query builder callable
      * @return array Containments.
      */
-    public function contain($associations = [])
+    public function contain($associations = [], callable $queryBuilder = null)
     {
         if (empty($associations)) {
             return $this->_containments;
+        }
+
+        if ($queryBuilder) {
+            $associations = [
+                $associations => [
+                    'queryBuilder' => $queryBuilder
+                ]
+            ];
         }
 
         $associations = (array)$associations;
--- a/src/ORM/EagerLoader.php
+++ b/src/ORM/EagerLoader.php
@@ -128,6 +128,7 @@ class EagerLoader
      * the new one.
      * @param callable|null $queryBuilder The query builder callable
      * @return array Containments.
+     * @throws \InvalidArgumentException When using $queryBuilder with an array of $associations
      */
     public function contain($associations = [], callable $queryBuilder = null)
     {
@@ -136,6 +137,12 @@ class EagerLoader
         }
 
         if ($queryBuilder) {
+            if (!is_string($associations)) {
+                throw new InvalidArgumentException(
+                    sprintf('Cannot set containments. To use $queryBuilder, $associations must be a string')
+                );
+            }
+
             $associations = [
                 $associations => [
                     'queryBuilder' => $queryBuilder
--- a/src/ORM/Query.php
+++ b/src/ORM/Query.php
@@ -352,14 +352,21 @@ class Query extends DatabaseQuery implements JsonSerializable, QueryInterface
      * Failing to do so will trigger exceptions.
      *
      * ```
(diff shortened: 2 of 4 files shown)


inoas:
Thanks for taking care ❤️ 

One question that remains (for me):
- Should the array syntax of contain be deprecated for better typehinting/stricter interfaces in 4.0
- If not: Should innerJoinWith/leftJoinWith/matching/notMatching also receive the array syntax to be on par completely?

src/ORM/Query.php
```diff
@@ -386,7 +395,12 @@ public function contain($associations = null, $override = false)
             return $loader->contain();
         }
 
-        $result = $loader->contain($associations);
```
inoas: Seems you are going for the new syntax internally already

jeremyharris: I'm not sure what you mean :)

The EagerLoader is what does the work so it needs to have the new signature first, then Query::contain needs to just support it.

inoas: Thanks for clarification!

jeremyharris:
I don't know that we can or want to deprecate the array syntax as it would make migrating very difficult and unfriendly:

```
$query
->contain('Articles')
->contain('Tags')
->contain('Profiles');
``

Some people contain *tons* of associations, I've seen 15+. Migrating to a method call per-contain would be a burden.

As for adding the array signature to matching()/etc. I don't think that's a great idea to support so many signatures as it's confusing for the user. Also, I don't know that people have so many matching() calls in a single query that an array syntax would be useful.

**Comment by inoas** [comment]
I see the point. What about adding methods that take arrays, loop, apply and call aka:
* containList() - in 4.x contain with array syntax would be required to be replaced by containList()
* matchingList() - allows to use array syntax as contain does as of 3.x
* notMatchingList() - allows to use array syntax as contain does as of 3.x
* innerJoinWithList() - allows to use array syntax as contain does as of 3.x

**Comment by cleptric** [comment]
We already had some discussions about changing the look and feel of the Query Builder.
Removing arrays from contain will be a huge BC break, as far as I know from my apps.
Doing stuff like this PR is fine, removing anything is not so fine 😢

**Comment by jeremyharris** [comment]
I'm down for writing the List methods as separate PRs.

Aside from that, anything else this needs?

**Comment by markstory** [comment]
> I'm down for writing the List methods as separate PRs.

I don't think we _need_ list methods. The non list methods will likely be simpler to read.

File: src/ORM/EagerLoader.php [hunk 2]
```diff
@@ -126,14 +126,23 @@ class EagerLoader
...
-    public function contain($associations = [])
+    public function contain($associations = [], callable $queryBuilder = null)
     {
         if (empty($associations)) {
             return $this->_containments;
         }
 
+        if ($queryBuilder) {
+            $associations = [
+                $associations => [
```
markstory: What happens when the parameter is an array? Someone will goof that up. I think you want line 138 to be `if ($queryBuilder && is_string($associations)) {`

inoas:
We do not need the list methods right now, but what about 4.x and php 7.1 The list methods could proxy the single methods and thus we have very strict typed interfaces - not a good thing?

markstory:
@inoas Why though? Calling the single methods multiple times works just as well.

inoas:
@markstory so are you okay removing array signature support from contain() in 4.x?

I'd really like to see: strict signatures for contain/matching/innerJoinWith/leftJoinWith/notMatching in 4.x - However especially for all those methods, both, dotnotation and array nesting, can be pretty useful. Latter would break the signatures strictness.

lorenzo:
Looks good to me!

markstory:
@inoas I think the contain(array) is used frequently enough that we might want to have `containMany()`

jeremyharris:
I would be happy to submit a PR for `containMany()` if it's generally wanted, and "deprecate" the array usage in `contain()`. I think at that point it would warrant a documentation change to push users to using the new signature for `contain()` and `containMany()` where appropriate.

PR #10667, opened 2017-05-17
Update code in Log.php

mehdi-fathi:
add `__FUNCTION__` to duplicate code name of function 

```
public static function emergency($message, $context = [])
    {
        return static::write('emergency', $message, $context);
    }
```


markstory:
Thanks!

PR #10668, opened 2017-05-18
RFC - 3.next Abstract schema type constants

markstory:
Add constants that allow us to reduce the chances of error when referring to abstract schema types. I know I've made a few confusing typos when building schema reflection and having constants would have made work simpler.

If people think this is a good idea, then I can update the other drivers to use the constants as well.

### DIFF
--- a/src/Database/Schema/MysqlSchema.php
+++ b/src/Database/Schema/MysqlSchema.php
@@ -14,6 +14,7 @@
  */
 namespace Cake\Database\Schema;
 
+use Cake\Database\Schema\TableSchema;
 use Cake\Database\Exception;
 
 /**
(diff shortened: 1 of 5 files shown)


src/Database/Schema/MysqlSchema.php
```diff
@@ -361,26 +362,41 @@ public function columnSql(TableSchema $schema, $name)
...
         }
 
-        $hasUnsigned = ['float', 'decimal', 'tinyinteger', 'smallinteger', 'integer', 'biginteger'];
+        $hasUnsigned = [
+            TableSchema::TYPE_TINYINTEGER,
+            TableSchema::TYPE_SMALLINTEGER,
+            TableSchema::TYPE_INTEGER,
+            TableSchema::TYPE_BIGINTEGER,
+            TableSchema::TYPE_FLOAT,
+            TableSchema::TYPE_DECIMAL
```
ravage84: 👍 to the improved order.

src/Database/Schema/MysqlSchema.php
```diff
@@ -361,26 +362,41 @@ public function columnSql(TableSchema $schema, $name)
                     break;
             }
         }
-        $hasLength = ['integer', 'smallinteger', 'tinyinteger', 'string'];
+        $hasLength = [
```
ravage84: May be we re-order them like https://github.com/cakephp/cakephp/pull/10668/files#diff-e7a8f6f7caf1af99989fbae47509a4f8R382

```` php
        $hasLength = [
            TableSchema::TYPE_TINYINTEGER,
            TableSchema::TYPE_SMALLINTEGER,
            TableSchema::TYPE_INTEGER,
            TableSchema::TYPE_STRING
        ];
````

Or does it need to be ordered that way?

File: src/Database/Schema/MysqlSchema.php [hunk 3]
```diff
@@ -361,26 +362,41 @@ public function columnSql(TableSchema $schema, $name)
...
+        ];
         if (in_array($data['type'], $hasUnsigned, true) &&
             isset($data['unsigned']) && $data['unsigned'] === true
         ) {
             $out .= ' UNSIGNED';
         }
 
-        $hasCollate = ['text', 'string'];
+        $hasCollate = [
+            TableSchema::TYPE_TEXT,
```
ravage84: order: string, text?

src/Database/Schema/TableSchema.php
```diff
@@ -249,6 +249,118 @@ class TableSchema
...
+
+    /**
+     * Timestamp column type
+     *
+     * @var string
+     */
+    const TYPE_TIMESTAMP = 'timestamp';
+
+    /**
+     * json column type
```
ravage84: JSON

ravage84 (commented):
Some nitpicking, but love the general idea 👍

robertpustulka:
Can we create `TableSchemaInterface` and these constants there? By the way `TableSchema` could be refactored to use getters/setters and those new methods added to the interface as well.

markstory:
I'm up for adding an interface, but I don't see any methods that combine get/set that still need to be split up.

robertpustulka:
> could be refactored to use getters/setters

\* If there are any :)

robertpustulka:
@markstory There's `columnType()`

PR #10669, opened 2017-05-18
Allow more IDE detectability. Fix invalid doc blocks.

dereuromark:
So far 
```
$result = $this->Bundles->find()->where(['name' => $bundle['name']])->first();
```
always needed a manual inline doc block, as first() said "mixed".
With this it will be `EntityInterface|null`, as expected by Query(Trait).
It then also more clearly states that this could return null :)

Also fixed up a few other lies.

### DIFF
--- a/src/Datasource/QueryTrait.php
+++ b/src/Datasource/QueryTrait.php
@@ -393,7 +393,7 @@ trait QueryTrait
      * $singleUser = $query->select(['id', 'username'])->first();
      * ```
      *
-     * @return \Cake\Datasource\EntityInterface|null The first result from the ResultSet.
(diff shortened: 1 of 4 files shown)


src/Datasource/QueryTrait.php
```diff
@@ -393,7 +393,7 @@ public function formatResults(callable $formatter = null, $mode = 0)
      * $singleUser = $query->select(['id', 'username'])->first();
      * ```
      *
-     * @return mixed the first result from the ResultSet
+     * @return \Cake\Datasource\EntityInterface|null The first result from the ResultSet.
```
ndm2: It could also be an array in case hydration is disabled.

dereuromark: That is true. Thx.

PR #10673, opened 2017-05-18
[3.x] Give the Composer JSONs some Love

ravage84:
Since the sub packages can be used separately, they should be setup correctly.

Gave all composer.json files an overhaul.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -32,10 +32,10 @@
         "php": ">=5.6.0",
         "ext-intl": "*",
         "ext-mbstring": "*",
-        "cakephp/chronos": "~1.0",
+        "cakephp/chronos": "^1.0.0",
         "aura/intl": "^3.0.0",
-        "psr/log": "^1.0",
-        "zendframework/zend-diactoros": "^1.4"
+        "psr/log": "^1.0.0",
+        "zendframework/zend-diactoros": "^1.4.0"
     },
     "suggest": {
         "ext-openssl": "To use Security::encrypt() or have secure CSRF token generation.",
--- a/composer.json
+++ b/composer.json
@@ -3,7 +3,16 @@
     "description": "The CakePHP framework",
     "type": "library",
     "keywords": [
-        "framework"
+        "framework",
+        "mvc",
+        "rapid-development",
+        "conventions over configuration",
+        "dry",
+        "orm",
+        "form",
+        "validation",
+        "rest-api",
+        "psr-7"
     ],
     "homepage": "https://cakephp.org",
     "license": "MIT",
--- a/composer.json
+++ b/composer.json
@@ -11,7 +11,6 @@
         "orm",
         "form",
         "validation",
-        "rest-api",
         "psr-7"
     ],
     "homepage": "https://cakephp.org",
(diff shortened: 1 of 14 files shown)


composer.json
```diff
@@ -2,7 +2,18 @@
...
+    "keywords": [
+        "framework",
+        "mvc",
+        "rapid-development",
+        "conventions over configuration",
+        "dry",
+        "orm",
+        "form",
+        "validation",
+        "rest-api",
```
dereuromark: I am not sure those terms should be in here -those are better used with the respective plugins that actually provide it fully.

src/I18n/composer.json
```diff
@@ -1,26 +1,48 @@
 {
     "name": "cakephp/i18n",
     "description": "CakePHP Internationalization library with support for messages translation and dates and numbers localization",
+    "type": "library",
+    "keywords": [
+        "cakephp",
+        "i18n",
+        "internationalisation",
+        "internationalization",
+        "localisation",
```
dereuromark: We should only use the US english (norm) term IMO - as we do with behavior and color etc.

src/Validation/composer.json
```diff
@@ -1,21 +1,35 @@
 {
     "name": "cakephp/validation",
     "description": "CakePHP Validation library",
+    "type": "library",
+    "keywords": [
+        "cakephp",
+        "validation",
+        "data validation"
```
dereuromark: Duplicate, validation itself should suffice, should it not?

composer.json
ravage84: I took some of the keywords from the GH repo https://github.com/cakephp/cakephp/

src/I18n/composer.json
ravage84: For a i18n package I thought it would make sense to be more inclusive 😼

composer.json
dereuromark: That kind of implies then that those are also not justified there, and shouldn't have found their way into the main repo :)

src/Validation/composer.json
ravage84: It's kind of SEO, I guess.

This way, it would be shown on this list, too:
https://packagist.org/search/?tags=data%20validation

Otherwise, not.

composer.json
ravage84: The framework is the sum of those sub packages (well more than just that) and thus both can contain the keywords that fit in what they provide.

src/Validation/composer.json
dereuromark: ;) all right.

composer.json
dereuromark: They don't provide a rest-api. you can build one with it (using the plugin that actually has those keywords listed), but so can you build a "youtube reloaded", a "math forum for geeks" or a "dating portal" :) None of those are candidates for keywords, though.

inoas: The feature list is good for exposure...

inoas (approved):

composer.json
dereuromark: Overdoing this is way worse then underdoing it IMO :)

inoas: More exposure / meta tags are always better unless it is miss-leading, which is not, in this case...

dereuromark (approved):

Issue #10665, opened 2017-05-17
BelongsToMany::link has defects


slince:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.4.6

### What you did

- `Users` ManyToMany `Roles`
-  set `saveStrategy` = 'append'
-  set (user_id, role_id) as primary key for join table `users_roles`
- code
```php
        $user1 = $this->Users->get(1);
        $user2 = $this->Users->get(2);

        $role1 = $this->loadModel('Roles')->get(1);
        $role2 = $this->loadModel('Roles')->get(2);

        $result1 = $this->Users->Roles->link($user1, [$role1, $role2]);
        $result2 = $this->Users->Roles->link($user2, [$role1]);
        debug($result1);
        debug($result2);
```
### What happened
Output true; but the join table has only two records;

| user_id | role_id|
| --- | --- |
| 1 | 1 |
| 1 | 2 |

If add field `id` and set `id`  as pk, the result is:

| id | user_id | role_id |
| --- | --- | ---|
|  1  | 2 | 1 |
|  2   |  1 | 2 |

markstory:
I will take a look.

markstory:
Pull request open now.

fquffio:
Just to let you know, commit b94a3c7 actually breaks things in a scenario where a single junction table is used for several associations, and a "discriminator" column is in the primary key and is used to tell one association from the others.

For instance, if one has a junction table with three columns:
 - `user_id`: foreign key to users table
 - `place_id`: foreign key to places table
 - `type`: one of `works_at`, `lives_in`, `usually_haves_breakfast_at`

In this case, you must manually set all these three columns in order for things to work, otherwise the whole primary key will be unset, thus `type` will become null. Before 3.4.8, just setting `type` did work, since the association took care of setting `user_id` and `place_id` correctly without touching `type`.

Since this is quite an uncommon scenario I'm not sure if it should be considered an issue, as fixing it would likely add complexity not needed by most users, but maybe somebody else is having an headache with the same problem. 😉

PR #10679, opened 2017-05-21
Fix multiple link() operations not persisting correctly

markstory:
When an entity is used in multiple link operations, the joint record would be reused, resulting in incorrectly persisted state.

I've also improve persistence related error messages a bit as I didn't grok them on first read.

Refs #10665

### DIFF
--- a/src/ORM/Association/BelongsToMany.php
+++ b/src/ORM/Association/BelongsToMany.php
@@ -1282,13 +1282,13 @@ class BelongsToMany extends Association
     protected function _checkPersistenceStatus($sourceEntity, array $targetEntities)
     {
         if ($sourceEntity->isNew()) {
-            $error = 'Source entity needs to be persisted before proceeding';
+            $error = 'Source entity needs to be persisted before links can be created or removed.';
             throw new InvalidArgumentException($error);
         }
 
         foreach ($targetEntities as $entity) {
             if ($entity->isNew()) {
-                $error = 'Cannot link not persisted entities';
+                $error = 'Cannot link entities that have not been persisted yet.';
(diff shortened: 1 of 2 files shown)


cleptric (approved):

src/ORM/Association/BelongsToMany.php
```diff
@@ -797,14 +797,19 @@ protected function _saveLinks(EntityInterface $sourceEntity, $targetEntities, $o
             $sourceKeys = array_combine($foreignKey, $sourceEntity->extract($bindingKey));
             $targetKeys = array_combine($assocForeignKey, $e->extract($targetPrimaryKey));
 
-            if ($sourceKeys !== $joint->extract($foreignKey)) {
-                $joint->set($sourceKeys, ['guard' => false]);
+            $changedKeys = (
+                $sourceKeys !== $joint->extract($foreignKey) ||
+                $targetKeys !== $joint->extract($assocForeignKey)
+            );
+            // Keys were changed, the junction table record _could_ be
```
ravage84: I wonder if we need so much explanation here, do we need to do something else or do it in another way, may be?

It _could_ be a sign of code smell or a general design problem...

cleptric: IMO this should be added way more often into the code. Sometimes you really have to dig through the code to understand things.

ravage84: Don't get me wrong, I'm all for inline comments. This was actually one thing I disliked about CakePHP 2.x.

Just thought (generally), if you need a lot explanation, it _could_ be a sign - whatever it means 😼

PR #10681, opened 2017-05-22
Class Connection true/false vs docblock

inoas:
However it returns true or throws an exception.
Question is also why $this->_driver->connect() does not also return a boolean...?


inoas:
Maybe better would be to remove the docblock as it can never return false?

markstory:
Removing the return false bit sounds like a good idea.

inoas:
I took a look at what is being called and the docblock was simply lying, AFAIU.

PR #10687, opened 2017-05-23
Fixed overwriting existing flash messages if none were retained

jeremyharris:
This fixes `enableRetainFlashMessages()` from overwriting the flash messages if no messages were retained in the first place, which fixes an issue (https://github.com/cakephp/cakephp/pull/10626#issuecomment-303332243) where early response returns had their messages overwritten.

ping @mirko-pagliai


mirko-pagliai:
Great work. Thanks again!

PR #10691, opened 2017-05-24
[RFC] Add TableSchemaInterface

robertpustulka:
Following https://github.com/cakephp/cakephp/pull/10668#issuecomment-302897784

The idea here is to add an interface for `TableSchema` with commonly used methods. I'm wondering if we should add all methods to the interface or are there any specific ones?

There are some Connection related SQL methods. I'm wondering maybe these could be placed in another interface?

Also there are some legacy getter methods without the `get*` prefix. Should we add the "correct" ones to the interface for consistency (getFoo/setFoo) and deprecate the old ones?

I've started with splitting another combined getter/setter for colum type and adding `hasColumn()` method.


markstory:
New methods look good. Adding an interface for the methods that we rely on internally would be good as well. I don't know if many plugins would implement the SQLy methods though.

src/Database/Schema/TableSchema.php
```diff
@@ -990,7 +1032,7 @@ public function truncateSql(Connection $connection)
      * Generate the SQL statements to add the constraints to the table
      *
      * @param \Cake\Database\Connection $connection The connection to generate SQL for.
-     * @return array SQL to drop a table.
+     * @return array SQL to add the constraints.
```
thinkingmedia: thank you :)

robertpustulka:
@markstory We could have something like `SqlGeneratorInterface` apart from `TableSchemaInterface`. Core `TableSchema` would implement both of them, but plugins could opt-out from generating sql.

markstory:
@robertpustulka That could work well. Perhaps we can use cakephp/elasticsearch as the second implementation of the `TableSchemaInterface`?

robertpustulka:
I've added a `SchemaInterface` to `Cake\Datasource` namespace. Unfortunetally there's already a `TableSchemaInterface` which actually looks more like a table schema decorator of some sort. It's used in fixtures. I'm wondering if we could rename it so the `TableSchemaInterface` name would be free in 4.0.

lorenzo:
> There are some Connection related SQL methods. I'm wondering maybe these could be placed in another interface?

This makes sense to me

src/Database/Schema/TableSchema.php
```diff
@@ -493,9 +382,21 @@ public function columns()
...
     public function column($name)
     {
+        return $this->getColumn($name);
+    }
+
+    /**
+     * Get column data in the table.
+     *
+     * @param string $name The column name.
+     * @return array|null Column data or null.
```
markstory: Couldn't these be `{@inheritDoc}` now?

robertpustulka:
I've split the interface into the one with general datasource methods and database specific methods. NoSQL datasources don't need to implement some of the methods.

I've also deprecated datasource `TableSchemaInterface` in favour of database `TableSchemaAwareInterface` which declares a getter and a setter for table schema.

lorenzo:
Changes look great!

src/Database/Schema/TableSchemaInterface.php
```diff
@@ -0,0 +1,244 @@
...
+ * @since         3.5.0
+ * @license       http://www.opensource.org/licenses/mit-license.php MIT License
+ */
+namespace Cake\Database\Schema;
+
+use Cake\Datasource\SchemaInterface;
+
+/**
+ *
+ * @author Robert Pustułka <r.pustulka@robotusers.com>
```
markstory: Could this be removed?

markstory (changes requested):
Author tags need to be removed.

src/Database/Schema/TableSchemaInterface.php
robertpustulka: OK, I missed that - it's been added by the IDE.

markstory (approved):

Issue #10689, opened 2017-05-24
Improve ergonomics of defining routes


markstory:
Separated from #10643 to be the work that can fit into the 3.5/3.6 timeline. These changes  intend to make the ergonomics of defining routes simpler:

* [x] Addition of HTTP method specific shorthands $routes->post(), $routes->put() etc.
* [x] Fluent interface methods for existing 'options'. The existing options that could have fluent setters defined are name(), method(), host(), requirements()
* [x] Add fluent helpers that were missed the first time. `pass()`, `persist()`, `multibytePattern()`
* [x] Make loading plugin routes a feature of the RouteBuilder. eg. `$routes->loadPlugin($name)`.
* [x] Compact defintion of Articles::view pairs. This is just a sugared version of ['controller' => 'Articles', 'action' => 'view']. It has no option/features for specifying more than the controller/action.
* [x] Enable `{var}` placeholders. This format will replace `:var` style placeholders potentially in 5.x as it is simpler/faster to parse out and doesn't suffer from mid word placement issues like `:var` does. See #9188 for an example route.

jeremyharris:
Copied from slack:

> I was wondering about the shorthand verb methods. Would they cause a 404 or a 405 if you sent a GET to a route that only had a DELETE. My initial assumption would be 404 as a get route doesn’t exist, although this is different behavior than most current applications which throw method not allowed within the action.

Perhaps this is just a matter of documentation.

inoas:
It really depends on security, I think. Github for instance shows 404 for projects you have no access to instead of 403 or 401.

The more pure response would be to clearly state what's wrong within HTTP status code speak: The method. The obfuscating/"secure" response would be to just throw 404.

And well, depending on the context / app - imho - you would want one or the other.
=>
* make it configurable?
* pick a sane default?

markstory:
Presently a mismatch on method will result in a 404 as the URL may not match any defined routes.

rchavik:
Is there anything that can be done for `Html::link()`?

Setting route is a one time thing, but `link()` usage is far more numerous.

ravage84:
I'm not sold on this ``Articles::view`` thing. Is that something that is asked a lot?

Also, agree with @rchavik.

jeremyharris:
I was worried this would be expected if we add a short reference for controller/action.

In regards to links, they often have much more information for building the route. How would you handle this:

```php
$this->Html->link('Title', [
    'controller' => 'Users',
    'action' => 'view',
[...]

PR #10692, opened 2017-05-25
3.5 - Add fluent setters to Route

markstory:
Implement the fluent setter features for Route described in #10689. I've opted to use `set` prefixes on the methods to better match the conventions we are headed towards with other parts of the framework and because `extensions()` is already deprecated in favour of `setExtensions()` and I did not want two method styles in the same class.

### DIFF
--- a/src/Routing/Route/Route.php
+++ b/src/Routing/Route/Route.php
@@ -182,6 +182,19 @@ class Route
     }
 
     /**
+     * Set pattern requirements for routing parameters
+     *
+     * @param array $patterns The patterns to apply to routing elements
+     * @return $this
+     */
+    public function setRequirements(array $patterns)
+    {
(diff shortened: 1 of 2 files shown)


src/Routing/Route/Route.php
```diff
@@ -152,6 +161,53 @@ public function getExtensions()
...
+        return $this;
+    }
+
+    /**
+     * Set pattern requirements for routing parameters
+     *
+     * @param array $patterns The patterns to apply to routing elements
+     * @return $this
+     */
+    public function setRequirements(array $patterns)
```
ADmad: Could we have a shorter method name, `setPatterns()` perhaps?

markstory: Done.

PR #10698, opened 2017-05-26
Cake 2.x - Some fix into Paginator component for order / sort classic sintax

lucasferreira:
Hi folks,

I've noted that classic array "order" aren't work in pagination (CakePHP 2.x):

```php
$this->Paginator->settings['Post'] = array(
  'order' => array('Post.created DESC', 'Post.id ASC') // <--- this doesn't works
);

$items = $this->Paginator->paginate('Post');
```

In my test only this kind of syntax are working:

```php
'order' => array('Post.created' => 'DESC', 'Post.id' => 'ASC') // this works
'order' => 'Post.created DESC, Post.id ASC' // with string also works

'order' => array('Post.created DESC', 'Post.id ASC') // <--- this doesn't works but with this PR will ;)
```

So this pull request has a fix to support all kind of order by syntax, link find() model method does.

Thanks ;)

### DIFF
--- a/lib/Cake/Test/Case/Controller/Component/PaginatorComponentTest.php
+++ b/lib/Cake/Test/Case/Controller/Component/PaginatorComponentTest.php
@@ -320,11 +320,19 @@ class PaginatorComponentTest extends CakeTestCase {
 
 		$Controller->PaginatorControllerPost->order = null;
 
+		/* ORDER ARRAY FIELD => SORT MODE */
 		$Controller->Paginator->settings = array(
 			'order' => array('PaginatorControllerComment.id' => 'ASC')
 		);
 		$results = Hash::extract($Controller->Paginator->paginate('PaginatorControllerComment'), '{n}.PaginatorControllerComment.id');
 		$this->assertEquals(array(1, 2, 3, 4, 5, 6), $results);
+
+		/* ORDER ARRAY "FIELD SORT" MODE */
+		$Controller->Paginator->settings = array(
+			'order' => array('PaginatorControllerComment.id DESC')
+		);
+		$results = Hash::extract($Controller->Paginator->paginate('PaginatorControllerComment'), '{n}.PaginatorControllerComment.id');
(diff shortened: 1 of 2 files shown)


markstory:
Could we get tests to cover the new functionality that is being added?

lucasferreira:
Shure @markstory 

I've added to this PR some simple tests to ensure this _classic array order_ behavior.

What do you think now?

markstory:
Looking​ good. I will look into why the tests on 2.next are failing.

lucasferreira:
I don't know if it helps, but here all tests passed:

pg_tests_ok

markstory:
Thanks. But we need tests passing in our CI system before we merge.

lib/Cake/Test/Case/Controller/Component/PaginatorComponentTest.php
```diff
@@ -629,6 +638,18 @@ public function testPaginateOrderModelDefault() {
 			'PaginatorControllerPost.created' => 'asc'
 		);
 		$this->assertEquals($expected, $result['order']);
+
+		/* CLASSIC */
```
ravage84: If we add comments, they should be a bit more expressive/verbose. Otherweise they aren't much of help.

ravage84 (changes requested):

lib/Cake/Test/Case/Controller/Component/PaginatorComponentTest.php
lucasferreira: Hi @ravage84, can I remove the comments?

markstory: > can I remove the comments?

That would be great.

lucasferreira: Done ;)

dereuromark:
Looks like unrelated fails. I restarted the build.

Issue #10697, opened 2017-05-26
Malformed SQL still Generated in Pagination with complex Matching() added


eaxexe:
This is a:

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.4.7
* Platform and Target: Apache2, PHP v. 5.6.2, MySQL v. 5.5.38 

This issue was reported before #10633, but still exists after #10657 was merged.

I have set up a repository that reproduces this error: [https://github.com/eaxexe/matching_test](https://github.com/eaxexe/matching_test)

It contains both a test `tests/TestCase/Model/Table/ArticlesTableTest.php` and the error can be seen by visiting matching_test/articles in the browser.

### What I did

I set up a beforeFind on my ArticlesTable that adds a few matching() queries to the find Query, with one having a matching() as a sub-query.

**Relevant code from ArticlesController.php:**
```
public function index()
    {
        $articles = $this->paginate($this->Articles);

        $this->set(compact('articles'));
        $this->set('_serialize', ['articles']);
    }
```

**ArticlesTable.php**
```
...
class ArticlesTable extends Table
{
    public function initialize(array $config)
    {
        parent::initialize($config);

        $this->setTable('articles');
        $this->setDisplayField('title');
        $this->setPrimaryKey('id');

        $this->addBehavior('Timestamp');

        $this->belongsToMany('Icons', [
            'foreignKey' => 'article_id',
            'targetForeignKey' => 'icon_id',
            'joinTable' => 'articles_icons'
        ]);
        $this->belongsToMany('Tags', [
            'foreignKey' => 'article_id',
            'targetForeignKey' => 'tag_id',
            'joinTable' => 'articles_tags'
        ]);
    }

    public function validationDefault(Validator $validator)
    {
        $validator
            ->uuid('id')
            ->allowEmpty('id', 'create');

        $validator
            ->requirePresence('title', 'create')
            ->notEmpty('title');

        $validator
            ->requirePresence('content', 'create')
            ->notEmpty('content');

        return $validator;
    }

    public function beforeFind(Event $event, Query $query) {
		$query->matching(
			'Tags', function(Query $q) {
				return $q
					->where(['Tags.name' => 'Good'])
					->matching('TagOptions', function(Query $i) {
						return $i->where(['TagOptions.name' => 'display_tag']);
					});
				})
				->matching('Icons', function(Query $q) {
					return $q->where(['Icons.icon' => 'sad']);
				});

[...]

PR #10699, opened 2017-05-28
Clone queries during pagination.

markstory:
While I don't think the SQL errors coming from EagerLoader re-use can be fixed without invasive changes to `EagerLoader`, we can treat the symptom of PaginatorComponent triggering this scenario by pre-emptively cloning the query.

Refs #10697

### DIFF
--- a/src/Controller/Component/PaginatorComponent.php
+++ b/src/Controller/Component/PaginatorComponent.php
@@ -187,9 +187,10 @@ class PaginatorComponent extends Component
             $query->applyOptions($options);
         }
 
+        $cleanQuery = clone $query;
         $results = $query->all();
         $numResults = count($results);
-        $count = $numResults ? $query->count() : 0;
+        $count = $numResults ? $cleanQuery->count() : 0;
 
         $defaults = $this->getDefaults($alias, $settings);
         unset($defaults[0]);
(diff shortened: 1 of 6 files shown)


src/Controller/Component/PaginatorComponent.php
```diff
@@ -187,9 +187,10 @@ public function paginate($object, array $settings = [])
             $query->applyOptions($options);
         }
 
+        $cleanQuery = clone $query;
         $results = $query->all();
         $numResults = count($results);
-        $count = $numResults ? $query->count() : 0;
+        $count = $numResults ? $cleanQuery->count() : 0;
```
lorenzo: count() is supposed to do the cloning. Should we fix it there? Or maybe remove it?

markstory: The issue is that after all() is called the EagerLoader retains state, as the queryBuilder callbacks are modified/wrapped during the first run.

markstory: I had considered changing the EagerLoader to separate 'runtime' state from the configuration state. It seems like a riskier job that might not be doable.

lorenzo: I think that is worth pursuing. In the meantime this is a good palliative.

markstory: I'll log a ticket for 3.5/3.6 to separate the runtime and configuration state. Hopefully we can get to it then.

PR #10700, opened 2017-05-28
3.next - HTTP method route creation helpers

markstory:
Add methods to RouteBuilder that make connecting routes for a single HTTP method simpler.

Refs #10689

### DIFF
--- a/src/Routing/RouteBuilder.php
+++ b/src/Routing/RouteBuilder.php
@@ -393,6 +393,131 @@ class RouteBuilder
     }
 
     /**
+     * Create a route that only responds to GET requests.
+     *
+     * @param string $template The URL template to use.
+     * @param array $target An array describing the target route parameters. These parameters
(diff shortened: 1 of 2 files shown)


lorenzo (commented):
I feel like the new methods should have an optional $options arguments, for stuff like setting regular expressions for the capture arguments.

markstory:
@lorenzo From the previous discussions people were leaning towards using the fluent interface methods added to Route. I am open to adding an options parameter like connect() has if it will be used.

lorenzo:
Do we expose methods for setting each of the options separately? In my view it is still a good idea to have it, even though we can encourage to use the fluent methods

markstory:
@lorenzo Not yet we don't but all the options will have methods in the next week or so.

markstory:
@lorenzo If it is ok with you, I'd like to hold off on adding the options parameter right now. Lets require use of the fluent interface and see how people respond. It will be a quick change for a future release.

PR #10701, opened 2017-05-28
Add small/tiny integer type mappings.

markstory:
The small and tiny integer types were not mapped in `Type`. This fixes errors that would be created when small/tiny integer fields were used as conditions in queries.

### DIFF
--- a/src/Database/Type.php
+++ b/src/Database/Type.php
@@ -32,6 +32,9 @@ class Type implements TypeInterface
      * @var array
      */
     protected static $_types = [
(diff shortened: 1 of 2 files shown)


lorenzo (commented):
Is the mapping complete though? When using it in fixtures, does it create the right type?

markstory:
@lorenzo Yes, both schema reflection and schema generation were correct. This part was forgotten though. I can update the DataTypeFixture to include these new types for additional coverage.

PR #10704, opened 2017-05-29
HTTP Client cookies add/remove methods.

robertpustulka:
Cookie collections are immutable in 3.5 and the setter allows for modifying client cookies on runtime.

### DIFF
--- a/src/Http/Client.php
+++ b/src/Http/Client.php
@@ -18,6 +18,7 @@ use Cake\Core\Exception\Exception;
 use Cake\Core\InstanceConfigTrait;
 use Cake\Http\Client\CookieCollection;
 use Cake\Http\Client\Request;
+use Cake\Http\Cookie\CookieInterface;
 use Cake\Utility\Hash;
 
 /**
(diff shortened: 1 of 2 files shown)


src/Http/Client.php
```diff
@@ -168,23 +169,45 @@ public function __construct($config = [])
...
+        return $this->_cookies;
+    }
+
+    /**
+     * Sets the cookie collection.
+     *
+     * @param \Cake\Http\Cookie\CookieCollection $cookies Cookie collection to be set.
+     * @return $this
+     */
+    public function setCookies(CookieCollection $cookies)
```
markstory: Why do you need to do this?

robertpustulka: Is there any other way to add a cookie to the client? $client->cookies()->add() will return a new collection AFAIR.

markstory: You can do it at construction time, or through the cookie option on requests. If you want to add a cookie to the client, lets expose that method then :smile:

robertpustulka: @markstory OK then. Should we expose delete method as well?

markstory: Do you find yourself often dropping cookies from a client? I've not needed to do that before personally.

robertpustulka: Maybe if you want to quickly drop session/logout. I have never used that either, but if we have add method, why not remove?

markstory: Keeping the class small and focuses on features they need is the only reason I can think of.

robertpustulka: @markstory I'll remove that method now but we can revert it if someone needs that.

PR #10705, opened 2017-05-29
Fix scope log format

mehdi-fathi:
Scope log wasn't in function _format() in BaseLog.php.Actually it just gives $context to that method.I move log message to scope if you declare a scope.

### DIFF
--- a/src/Routing/RouteBuilder.php
+++ b/src/Routing/RouteBuilder.php
@@ -587,13 +587,24 @@ class RouteBuilder
      * This method creates a scoped route collection that includes
      * relevant prefix information.
      *
-     * The path parameter is used to generate the routing parameter name.
+     * The $name parameter is used to generate the routing parameter name.
      * For example a path of `admin` would result in `'prefix' => 'admin'` being
      * applied to all connected routes.
      *
      * You can re-open a prefix as many times as necessary, as well as nest prefixes.
      * Nested prefixes will result in prefix values like `admin/api` which translates
      * to the `Controller\Admin\Api\` namespace.
+     *
+     * If you need to have prefix with dots, eg: '/api/v1.0', use 'path' key
+     * for $params argument:
+     *
+     * ```
+     * $route->prefix('api', function($route) {
+     *     $route->prefix('v10', ['path' => '/v1.0'], function($route) {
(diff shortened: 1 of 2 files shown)


src/Log/Engine/BaseLog.php
```diff
@@ -89,20 +89,31 @@ public function scopes()
      */
     protected function _format($data, array $context = [])
     {
-        if (is_string($data)) {
+        $result = '';
+        if (isset($context['scope']) && is_array($context['scope'])) {
+            foreach ($context['scope'] as $c) :
+                $result['scope'][$c] = $data;
+            endforeach;
```
markstory: Please use braces in PHP code, and not alternative control structures.

src/Log/Engine/BaseLog.php
```diff
@@ -89,20 +89,31 @@ public function scopes()
      */
     protected function _format($data, array $context = [])
     {
-        if (is_string($data)) {
+        $result = '';
+        if (isset($context['scope']) && is_array($context['scope'])) {
+            foreach ($context['scope'] as $c) :
+                $result['scope'][$c] = $data;
```
markstory: $result is a string here. You're relying on implict array casting which looks incorrect.

src/Log/Engine/BaseLog.php
```diff
@@ -89,20 +89,31 @@ public function scopes()
...
+            endforeach;
+        }
+        if (is_string($data) && !is_array($result)) {
             return $data;
+        } elseif (is_string($data) && is_array($result)) {
+            return print_r($result);
```
markstory: This is going to dump to stdout. That doesn't seem right.

src/Log/Engine/BaseLog.php
```diff
@@ -89,20 +89,31 @@ public function scopes()
...
+                $data = $result;
+            }
             return json_encode($data);
         }
-
-        return print_r($data, true);
+        if (is_array($result)) {
+            return print_r($result, true);
+        } else {
+            return print_r($data, true);
```
markstory: More print_r used here.

markstory (changes requested):
The print_r usage looks like a mistake.

PR #10706, opened 2017-05-30
3.next - Add remaining fluent methods for routes.

markstory:
I've made the multibyte option an implicit setting. I feel this makes routing simpler to use as we can infer from the patterns used whether or not multiple patterns should be used. Having an additional method call required makes Route clunkier to use.

Refs #10689

### DIFF
--- a/src/Routing/Route/Route.php
+++ b/src/Routing/Route/Route.php
@@ -184,11 +184,18 @@ class Route
     /**
      * Set regexp patterns for routing parameters
      *
+     * If any of your patterns contain multibyte values, the `multibytePattern`
+     * mode will be enabled.
+     *
      * @param array $patterns The patterns to apply to routing elements
      * @return $this
      */
     public function setPatterns(array $patterns)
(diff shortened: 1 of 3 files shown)


src/Routing/Route/Route.php
```diff
@@ -184,11 +184,18 @@ public function setMethods(array $methods)
     /**
      * Set regexp patterns for routing parameters
      *
+     * If any of your patterns contain multibyte values, the `multibytePattern`
+     * mode will be enabled.
+     *
      * @param array $patterns The patterns to apply to routing elements
      * @return $this
      */
     public function setPatterns(array $patterns)
```
ADmad: Since other methods like setPass(), setPersist() which also take array argument are singular, wouldn't it be better to name this method too setPattern() for consistency?

markstory: setPerists() and setPasses() sounded silly, which is why they are 'singular'.

PR #10707, opened 2017-05-30
Split ValidatorAwareTrait::validator() into getter/setter.

robertpustulka:
Refs #9978

I've also added an interface with definitions for new methods. Could be useful to determine if an object supports validators.

### DIFF
--- a/src/ORM/Table.php
+++ b/src/ORM/Table.php
@@ -37,6 +37,7 @@ use Cake\ORM\Exception\PersistenceFailedException;
 use Cake\ORM\Exception\RolledbackTransactionException;
 use Cake\ORM\Rule\IsUnique;
 use Cake\Utility\Inflector;
+use Cake\Validation\ValidatorAwareInterface;
 use Cake\Validation\ValidatorAwareTrait;
 use InvalidArgumentException;
 use RuntimeException;
@@ -123,19 +124,12 @@ use RuntimeException;
  *
  * @see \Cake\Event\EventManager for reference on the events system.
  */
-class Table implements RepositoryInterface, EventListenerInterface, EventDispatcherInterface
+class Table implements RepositoryInterface, EventListenerInterface, EventDispatcherInterface, ValidatorAwareInterface
 {
 
     use EventDispatcherTrait;
     use RulesAwareTrait;
     use ValidatorAwareTrait;
-
-    /**
-     * Name of default validation set.
-     *
-     * @var string
-     */
-    const DEFAULT_VALIDATOR = 'default';
 
     /**
      * The alias this object is assigned to validators as.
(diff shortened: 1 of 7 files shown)


src/ORM/Table.php
```diff
@@ -123,26 +124,26 @@
...
+    const VALIDATOR_PROVIDER_NAME = 'table';
 
     /**
-     * The alias this object is assigned to validators as.
+     * The name of the event dispatched when a validator has been built.
      *
      * @var string
      */
-    const VALIDATOR_PROVIDER_NAME = 'table';
+    const BUILD_VALIDATOR_EVENT = 'Model.buildValidator';
```
markstory: Is this constant used anywhere?

src/Validation/ValidatorAwareTrait.php
```diff
@@ -93,31 +93,139 @@
...
      * @return \Cake\Validation\Validator
      * @throws \RuntimeException
+     * @deprecated 3.5.0 Use getValidator/setValidator instead.
      */
     public function validator($name = null, Validator $validator = null)
     {
+        if ($validator !== null) {
+            if ($name === null) {
+                $name = self::DEFAULT_VALIDATOR;
+            }
```
markstory: This could be

```php
$name = $name ?: self::DEFAULT_VALIDATOR
```
🏌️

src/Validation/ValidatorAwareTrait.php
```diff
@@ -93,31 +93,139 @@
...
+     * class constant.
+     *
+     * @param string|null $name The name of the validation set to return.
+     * @return \Cake\Validation\Validator
+     */
+    public function getValidator($name = null)
+    {
         if ($name === null) {
             $name = self::DEFAULT_VALIDATOR;
         }
```
markstory: Another place you could use the short ternary to save some lines 🏌️

src/Validation/ValidatorAwareTrait.php
```diff
@@ -93,31 +93,139 @@
...
+     *
+     * This method is used only to build a new validator and it does not store
+     * it in your object. If you want to build and reuse validators,
+     * use getValidator() method instead.
+     *
+     * @param string $name The name of the validation set to create.
+     * @return \Cake\Validation\Validator
+     * @throws \RuntimeException
+     */
+    public function createValidator($name)
```
markstory: Does this method need to be public?

src/ORM/Table.php
robertpustulka: Here in a trait: https://github.com/cakephp/cakephp/pull/10707/files#diff-5ea28017bc64b94c8fdbe87e3b422986R178

I wanted this trait to be used outside model scope.

src/Validation/ValidatorAwareTrait.php
robertpustulka: Right, better be it protected.

ADmad:
Using the constant `BUILD_VALIDATOR_EVENT` in `ValidatorAwareTrait::createValidator()` is not backwards compatible. It's causing error in `muffin/webservice` plugin, see UseMuffin/Webservice#46.

Issue #10581, opened 2017-04-24
Inconsistent behaviour of Entity::has()


burzum:
This is a (multiple allowed):

* [x] bug
* [x] enhancement
* [x] feature-discussion (RFC)

* CakePHP Version: 3.next and previous

### What you did
Called[ Entity::has()](https://github.com/cakephp/cakephp/blob/3.next/src/Datasource/EntityTrait.php#L339) on an empty hasMany or belongsToMany property.

### What happened
Returns true

### What you expected to happen
Expected false

---

I know you can have a long discussion about if has() means that the property is expected to be empty or just not null. And I'm pretty sure the opinions in the team will be different.

I personally would prefer to get false back when a hasMany / belongsToMany property is an empty array. I could accept that this is not the case if the entity would provide another method to check if a property, that is hasMany / belongsTomany, is empty or not. Maybe notEmpty()?

However, the current implementation, especially to devs not used to it is...

![image](https://cloud.githubusercontent.com/assets/162789/25352921/e5b6606e-292d-11e7-84d0-b2983d19ba26.png)

markstory:
The entity you were checking did *have* a property, so the return value makes sense to me. How would an entity determine whether or not the property being checked is an association property or something that should contain a scalar. This also extends into other types like JSON column values. How should `has()` treat those?

It sounds like we might want another methods for checking if a property is not 'empty'.

dereuromark:
maybe has() should be hasDefined() - and then has() can be non nullish value for real.

cleptric:
Adding a new method for checking if the property is not empty sounds responsible to me, so 👍

dereuromark:
notEmpty()? could be confusing as this should probably return true for 0 :)
hasValue()?

burzum:
Yea, good question, I started writing code but came across the 0 "issue" as well. :) How do we want to threat it?

cleptric:
It should not return `true`for `[]`, `null` and `''` I guess.
If we can agree on that `0` is a a value, `hasValue()` sounds good to me.

burzum:
I don't like `value`, because an empty array is a value and '' is basically a value as well.

How about isEmpty()? I think that most of the time the state you want to check against.

inoas:
> maybe has() should be hasDefined() - and then has() can be non nullish value for real.

:+1:

hasPropertyDefined() = property_exists()
[...]

PR #10709, opened 2017-05-30
Adding methods to the entity to check for values

burzum:
See https://github.com/cakephp/cakephp/issues/10581

### DIFF
--- a/src/Datasource/EntityTrait.php
+++ b/src/Datasource/EntityTrait.php
@@ -344,6 +344,44 @@ trait EntityTrait
      *
      * ```
      * $entity = new Entity(['id' => 1, 'name' => null]);
+     * $entity->hasProperty('id'); // true
+     * $entity->hasProperty('name'); // false
+     * $entity->hasProperty('last_name'); // false
+     * ```
+     *
+     * You can check multiple properties by passing an array:
+     *
+     * ```
+     * $entity->hasProperty(['name', 'last_name']);
+     * ```
+     *
+     * All properties must not be null to get a truthy result.
+     *
+     * When checking multiple properties. All properties must not be null
+     * in order for true to be returned.
+     *
+     * @param string|array $property The property or properties to check.
+     * @return bool
+     */
+    public function hasProperty($property)
+    {
+        foreach ((array)$property as $prop) {
+            if ($this->get($prop) === null) {
+                return false;
+            }
+        }
+
+        return true;
+    }
+
+    /**
+     * Returns whether this entity contains a property named $property
+     * that contains a non-null value.
+     *
+     * ### Example:
+     *
+     * ```
+     * $entity = new Entity(['id' => 1, 'name' => null]);
      * $entity->has('id'); // true
      * $entity->has('name'); // false
      * $entity->has('last_name'); // false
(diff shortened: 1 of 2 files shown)


src/Datasource/EntityTrait.php
```diff
@@ -375,6 +375,87 @@ public function has($property)
...
+     *
+     * All properties must not be null to get a truthy result.
+     *
+     * When checking multiple properties. All properties must not be null
+     * in order for true to be returned.
+     *
+     * @param string|array $property The property or properties to check.
+     * @return bool
+     */
+    public function has($property)
```
robertpustulka: Is this method deprecated?

burzum: I've asked in the issue if we want to deprecate it or keep it as shortcut.

robertpustulka: @burzum I think that there should be only one method for this.

burzum: @robertpustulka fine for me, let's wait until somenbody else comments on this or can you alone make this decision?

robertpustulka: @burzum let's hear others :)

inoas: I think there should either be just has() or hasProperty(), not both of them as they do the same (aka deprecation may be required).

I do not care much which one should be in, but suspect that for newbies latter is better as it clearly tells what it does.

tests/TestCase/ORM/EntityTest.php
```diff
@@ -1573,4 +1573,68 @@ public function testIsDirtyFromClone()
...
+            'string' => 'string',
+            'emptyString' => '',
+            'intZero' => 0,
+            'intNotZero' => 1,
+            'floatZero' => 0.0,
+            'floatNonZero' => 1.5,
+            'null' => null
+        ]);
+
+        $this->assertFalse($entity->isEmpty('array'));
```
inoas: Maybe group both - the list above and the asset list below - by their return value (true/false)

src/Datasource/EntityTrait.php
markstory: 👍 to only having one way to do this. What do we gain from the rename?

src/Datasource/EntityTrait.php
```diff
@@ -375,6 +375,87 @@ public function has($property)
...
+    {
+        return $this->hasProperty($property);
+    }
+
+    /**
+     * Checks that a property is empty
+     *
+     * This is not working like the built in empty() of php. The method will
+     * return true for:
```
markstory: Should be PHP

src/Datasource/EntityTrait.php
```diff
@@ -375,6 +375,87 @@ public function has($property)
...
+        return false;
+    }
+
+    /**
+     * Checks tha a property has a value.
+     *
+     * This method will return true for
+     *
+     * - 'foo' (non empty string)
+     * - ['foo', 'bar']
```
markstory: Could use 'array with one or more elements' to be clearer.

src/Datasource/EntityTrait.php
burzum: > What do we gain from the rename?

Clarity. ;) It's longer but much more clear. The current name and it's behavior is a trap. I would prefer hasProperty() over has(). Tell me your decision and I'll do it.

robertpustulka: Currently `has()` acts more like `isset()`. It will return false if a property has been explicitly set to `null`. Tthere's no distinction between `array_key_exists($this->_properties['foo'])` and `isset($this->_properties['foo'])`.

Do we need some way of checking if property exists in `$_properties`?

inoas: I suggested `hasPropertyDefined()` and `hasAssociation()` (vs hasOne/belongsTo) /  `hasAssociations()`.... (vs hasMany)

burzum: There are no associations, it's just data. The entity is not aware of any association. Also checking if a property is defined is something else than checking if a field contains a value.

inoas: Properties can have entities as values which closely mimics associations: `hasEntity()` and `hasEntities()` then.

Basically there should be:
- a way to check if a key property exists
- a way to check if a property is set to a non-nullish value
- a way to check if a property `is_a()` given Entity X or is a list with m...n items (or a closure for the count check)

Example:
```php
// property must exist but can be nullish
$entity->hasProperty('locale_code_user_preference');

// property must exist and cannot be nullish
$entity->hasValue('title');

// property must exist and must be is_a/subclass of Profile
$entity->hasEntity('profile', \App\Model\Profile::classname); 

// property must exist and returns true if the list contains 1 to 20 Comment entities. This 3rd param could also be a closure.
$entity->hasEntities('comments', \App\Model\Comment::classname, ['min' => 1, 'max' => '20']);

// ... or more general purpose, if preferred
$entity->hasObject('created', \Datetime::classname);
$entity->hasValue('created', \Datetime::classname);
```

For this PR then IMHO `hasValue()` + `hasProperty()` + deprecation of `has()` is good

src/Datasource/EntityTrait.php
```diff
@@ -375,6 +375,87 @@ public function has($property)
...
+     *
+     * @param string $property The property to check.
+     * @return bool
+     */
+    public function isEmpty($property)
+    {
+        $value = $this->get($property);
+        if ($value === null
+            || (is_array($value) && empty($value)
+            || (is_string($value) && empty($value)))
```
markstory: empty() on string values will fail on `'0'`. I think you want `strlen()` instead.

src/Datasource/EntityTrait.php
markstory: I decided to keep `has` over the rename. I don't think the rename adds substantial value, and it would be yet another breaking change we'd be adding in for 4.0.

PR #10710, opened 2017-06-01
Improved parsing of AngularJS URLS.

sevvlor:
The AngularJS URLS may contain `;` and `!`, they broke (parsed partially).

I have added the two characters to the regex and added the unittests. Though I really think that I (or someone else) should look into an more permanent solution to this.

See issue #10329

### DIFF
--- a/src/View/Helper/TextHelper.php
+++ b/src/View/Helper/TextHelper.php
@@ -123,7 +123,7 @@ class TextHelper extends Helper
                     (?<left>[\[<(]) # left paren,brace
                     (?>
                         # Lax match URL
-                        (?<url>(?:https?|ftp|nntp):\/\/[\p{L}0-9.\-_:]+(?:[\/?][\p{L}0-9.\-_:\/?=&>\[\]()#@\+~!;%]+)?)
+                        (?<url>(?:https?|ftp|nntp):\/\/[\p{L}0-9.\-_:]+(?:[\/?][\p{L}0-9.\-_:\/?=&>\[\()#@\+~!;,%]+[^-_:?=&>\[\()#@\+~!;,.%\s])?)
                         (?<right>[\])>]) # right paren,brace
                     )
                 )
--- a/src/View/Helper/TextHelper.php
+++ b/src/View/Helper/TextHelper.php
@@ -123,7 +123,7 @@ class TextHelper extends Helper
                     (?<left>[\[<(]) # left paren,brace
                     (?>
                         # Lax match URL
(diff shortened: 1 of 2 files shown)


dereuromark:
Could this break URLs that so far worked when ending inside a sentence?

    Check out www.my-example.org!

The `!` so far wasnt part of the URL afaik, same for `;`

sevvlor:
@dereuromark Correct. I couldn't run the unittest myself. I'll fix it later today and update the PR.

sevvlor:
I might have a solution. Currently the URL parser has this regex:
```regex
(?<url>(?:https?|ftp|nntp):\/\/[\p{L}0-9.\-_:]+(?:[\/?][\p{L}0-9.\-_:\/?=&>\[\]()#@\+~%]+)?)
```
The problem with this is that it also parses `https://domain.name/path/to/file!` rather than `https://domain.name/path/to/file`. In other words it also captures the `!` (same goes for other characters). 

I do have a fix. This regex ignores the character  `-_:?=&>\[\()#@\+~!;,.%\s` if it's the trailing character to an URL.

```regex
(?<url>(?:https?|ftp|nntp):\/\/[\p{L}0-9.\-_:]+(?:[\/?][\p{L}0-9.\-_:\/?=&>\[\()#@\+~!;,%]+[^-_:?=&>\[\()#@\+~!;,.%\s])?)
```
This fixes the the problem with the `;` and `!` too. However, I think that this is not a perfect solution as the regex gets more and more complicated and URLs still might break.

[See regex101](https://regex101.com/r/1pI6D7/3)

Working URLS:
```
This is a test text with URL ;http://www.cakephp.org/page/4; semi-colon
This is a test text with URL ;http://www.cakephp.org/pag!e/4; semi-colon
This is a test text with URL ;http://www.cakephp.org/page/4;/asdhfasd semi-colon
This is a test text with URL ;http://www.cakephp.org/pag!!!e/4;/askdhasd! semi-colon
This is a test text with URL ;http://www.cakephp.org/page/4! semi-colon
This is a test text with URL ;http://www.cakephp.org/page/4.asdasd. semi-colon
This is a test text with URL ;http://www.cakephp.org/page/4,pasdfsd, semi-colon
This is a test text with URL ;http://www.cakephp.org/#!page/4,pasdfsd, semi-colon
```

dereuromark:
Maybe we don't have to, it was just a small question.
It could very well also be a valid character of the URL, I imagine.
So there is no perfect solution here - I guess either way is fine moving forward.

markstory:
Changes look good but there is a regression with linking to fragments.

PR #10711, opened 2017-06-02
3.next - Add RouteBuilder::loadPlugin.

markstory:
Deprecate the static method and provide a route builder method that allows routes to be loaded inside scopes.

Refs #10689

### DIFF
--- a/src/Routing/RouteBuilder.php
+++ b/src/Routing/RouteBuilder.php
@@ -16,6 +16,8 @@ namespace Cake\Routing;
 
 use BadMethodCallException;
 use Cake\Core\App;
+use Cake\Core\Exception\MissingPluginException;
+use Cake\Core\Plugin;
 use Cake\Routing\Route\Route;
 use Cake\Utility\Inflector;
 use InvalidArgumentException;
@@ -516,6 +518,35 @@ class RouteBuilder
         $this->_collection->add($route, $options);
 
         return $route;
+    }
+
+    /**
+     * Load routes from a plugin.
+     *
+     * The routes file will have a local variable named `$routes` made available which contains
+     * the current RouteBuilder instance.
+     *
+     * @param string $name The plugin name
+     * @param string $file The routes file to load. Defaults to `routes.php`
+     * @return void
+     * @throws \Cake\Core\Exception\MissingPluginException When the plugin has not been loaded.
+     * @throws \InvalidArgumentException When the plugin does not have a routes file.
+     */
+    public function loadPlugin($name, $file = 'routes.php')
+    {
+        if (!Plugin::loaded($name)) {
+            throw new MissingPluginException(['plugin' => $name]);
+        }
+        $path = Plugin::configPath($name) . DIRECTORY_SEPARATOR . $file;
+        if (!file_exists($path)) {
+            throw new InvalidArgumentException(sprintf(
+                'Cannot load routes for the plugin named %s. The %s file does not exist.',
+                $name,
+                $path
+            ));
+        }
+        $routes = $this;
+        include $path;
     }
 
     /**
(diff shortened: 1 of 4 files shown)


src/Routing/RouteBuilder.php
```diff
@@ -519,6 +521,35 @@ protected function _methodRoute($method, $template, $target, $name)
     }
 
     /**
+     * Load routes from a plugin.
+     *
+     * The routes file will have a local variable named `$routes` made available which contains
```
ravage84: will -> must?

src/Routing/RouteBuilder.php
```diff
@@ -519,6 +521,35 @@ protected function _methodRoute($method, $template, $target, $name)
...
+     * the current RouteBuilder instance.
+     *
+     * @param string $name The plugin name
+     * @param string $file The routes file to load. Defaults to `routes.php`
```
ravage84: 👍

src/Routing/RouteBuilder.php
```diff
@@ -519,6 +521,35 @@ protected function _methodRoute($method, $template, $target, $name)
...
+     * @return void
+     * @throws \Cake\Core\Exception\MissingPluginException When the plugin has not been loaded.
+     * @throws \InvalidArgumentException When the plugin does not have a routes file.
+     */
+    public function loadPlugin($name, $file = 'routes.php')
+    {
+        if (!Plugin::loaded($name)) {
+            throw new MissingPluginException(['plugin' => $name]);
+        }
+        $path = Plugin::configPath($name) . DIRECTORY_SEPARATOR . $file;
```
ravage84: I'd insert a blank line before for readability.

src/Routing/RouteBuilder.php
```diff
@@ -519,6 +521,35 @@ protected function _methodRoute($method, $template, $target, $name)
...
+        if (!file_exists($path)) {
+            throw new InvalidArgumentException(sprintf(
+                'Cannot load routes for the plugin named %s. The %s file does not exist.',
+                $name,
+                $path
+            ));
+        }
+        $routes = $this;
```
ravage84: I'd insert a blank line before for readability.

ravage84:
@markstory can you elaborate what the implications are for the future app setup?

src/Routing/RouteBuilder.php
markstory: I was trying to say the routes file will have access to `$routes` as a local variable.

markstory:
@ravage84 I would like to see `Plugin::routes()` go away, and instead plugins either use events to automagically add their routes, or the installation of plugins requires adding `$routes->loadPlugin(...)` in the right place.

src/Routing/RouteBuilder.php
ravage84: Yeah, I see now. Never mind.

ravage84:
@markstory I see. I just "fear" the installation of plugins or better said the configuration of plugin routes could become more complicated because of this...

markstory:
@ravage84 Fully removing `Plugin::routes()` and only having this method would make plugin installation more complex, which is not my goal. I want to make routing easier to understand and simpler to change. I found that `Plugin::routes()` is quite opaque and makes routing harder to understand and change.

Perhaps `Plugin::routes()` can live alongside this new method as a simple way to splat all the routes at once, while this new method gives people finer grained control over routes.

lorenzo:
@markstory what happens if you use both `loadPlugin` and `Plugin::routes()`. Should this be an error condition?

markstory:
@lorenzo The routes would be loaded twice. Which could result in incorrect URLs being generated. While I like to use `include_once` for routes to prevent that, but that complicates testing as we need to reload routes quite often.

If the routes are named, they will hit an error, but un-named routes are harder to catch.

Issue #7276, opened 2015-08-21
pagination in cells


arilia:
being "small mini-controllers" I expected I could use components in Cells. 

I understand that components are attached to actual controllers and cells are not controllers. But It would be very useful to have at least pagination. Is it possible? Are there technical problems to solve or it is a deliberate choice?

dereuromark:
Refs https://github.com/cakephp/cakephp/issues/3660
If the PaginatorComponent logic was extracted into a more common lib class it might very well possible to also make it available for cells and other use cases.

lorenzo:
@arilia How would pagination work in a cell? It requires reading variables from the query string, wouldn't it clash with the same variables used in the pagination for the main view?

ionas:
Paginator should take some (optional) parameters to encapsulate one pagination from another (arrayish params). I have done that for 2.x and imho it should come out of the box for bake already.

arilia:
@lorenzo the optimal would be being able to paginate more than one controller at a time as described by ionas. But I think it's up to me avoid conflicts.

Even now it's possible to call paginate() many time in the same action

lorenzo:
Closing as duplicate of https://github.com/cakephp/cakephp/issues/1731

PR #6356, opened 2015-04-17
RFC: Refactor/paginator

burzum:
There is no reason why the pagination component has  to be coupled to the component / controller layer. If you want to use pagination inside a shell or a service layer you'll have two choices:
- Ignore MVC and import the paginator component in a place where it doesn't belong to technically
- Re-invent the wheel and implement your own pagination

This change moved the whole pagination logic into a new class.

There is one thing to discuss: I couldn't find a better way to implement the pagination based on the request than extending the new class by another class and injecting the request there. Ideas how to do this better are welcome! The paginator component test passes for me with my current changes.

PR #10718, opened 2017-06-04
Add standalone Paginator class.

ADmad:
Refs #6356, #7276. This should allow paginating in cells :)

I'll update the docblocks once code changes are approved.


src/Controller/Component/PaginatorComponent.php
```diff
@@ -56,6 +55,23 @@ class PaginatorComponent extends Component
...
+     * @var \Cake\ORM\Paginator
+     */
+    protected $_paginator;
+
+    /**
+     * {@inheritDoc}
+     */
+    public function __construct(ComponentRegistry $registry, array $config = [])
+    {
+        $this->_paginator = new Paginator();
```
ndm2: Could this maybe be injectable, for easy integration of custom/extended paginators?

ndm2:
I like the direction this is going :+1: , but it's a kind of a hard break for a patch release. I have for example at least two apps with extended paginator components that would break (a little bit), as they rely on overridden `getDefaults()` and `validateSort()` methods.

dereuromark:
3.5 should be fine then?

src/Controller/Component/PaginatorComponent.php
markstory: Or for plugins like elasticsearch to provide implementations of.

markstory (changes requested):
Needs some tests, and this is a big enough change that it should probably target 3.next

src/ORM/Paginator.php
```diff
@@ -0,0 +1,455 @@
...
+ * This component is used to handle automatic model data pagination. The primary way to use this
+ * component is to call the paginate() method. There is a convenience wrapper on Controller as well.
+ *
+ * ### Configuring pagination
+ *
+ * You configure pagination when calling paginate(). See that method for more details.
+ *
+ * @link http://book.cakephp.org/3.0/en/controllers/components/pagination.html
+ */
+class Paginator
```
robertpustulka: How about adding a `PaginatorInterface` so plugins can provide custom compatible implementations?

src/ORM/Paginator.php
```diff
@@ -0,0 +1,455 @@
...
+ * @link          http://cakephp.org CakePHP(tm) Project
+ * @since         3.4.8
+ * @license       http://www.opensource.org/licenses/mit-license.php MIT License
+ */
+namespace Cake\ORM;
+
+use Cake\Core\InstanceConfigTrait;
+use Cake\Datasource\QueryInterface;
+use Cake\Datasource\RepositoryInterface;
+use Cake\Network\Exception\NotFoundException;
```
lorenzo: Not cool having the Network package as a dependency for the ORM now.

lorenzo: Maybe create a new exception class that also returns a 404 code. This is a good opportunity to also make a nice error page.

lorenzo (commented):
I was attempted to do something similar but got desilusioned with the result, which was similar to this. 

My only gripe with the change is that we have moved all the request params and cleanup logic to his new class, and the logic will not be shared by other custom paginators that will still need to copy and paste from this.

I think it is a good first step to get things moving, though.

src/ORM/Paginator.php
ADmad: The name `PageOutOfBounds` for the exception sound good?

src/ORM/Paginator.php
```diff
@@ -0,0 +1,455 @@
...
+            $field = $key;
+            $alias = $tableAlias;
+
+            if (strpos($key, '.') !== false) {
+                list($alias, $field) = explode('.', $key);
+            }
+            $correctAlias = ($tableAlias === $alias);
+
+            if ($correctAlias && $whitelisted) {
+                // Disambiguate fields in schema. As id is quite common.
```
inoas: While I am personally a fan of comments like this, the core is void of them, isn't it?

src/ORM/Paginator.php
```diff
@@ -0,0 +1,455 @@
...
+     *
+     * Each of the above queries will use different query string parameter sets
+     * for pagination data. An example URL paginating both results would be:
+     *
+     * ```
+     * /dashboard?articles[page]=1&tags[page]=2
+     * ```
+     *
+     * @param \Cake\Datasource\RepositoryInterface|\Cake\Datasource\QueryInterface $object The table or query to paginate.
+     * @param array $params Request params
```
inoas: .

src/ORM/Paginator.php
dereuromark: Even just OutOfBounds would do it IMO.

src/ORM/Paginator.php
```diff
@@ -0,0 +1,455 @@
...
+     * @param array $settings The settings/configuration used for pagination.
+     * @return \Cake\Datasource\ResultSetInterface Query results
+     * @throws \Cake\Network\Exception\NotFoundException
+     */
+    public function paginate($object, array $params = null, array $settings = [])
+    {
+        $query = null;
+        if ($object instanceof QueryInterface) {
+            $query = $object;
+            $object = $query->repository();
```
inoas: Shouldn't `/* @var \Cake\Datasource\RepositoryInterface $object */` be here instead of down there?

dereuromark:
@lorenzo Would an abstract class help that could share the common code for both core and custom extension points?

src/ORM/Paginator.php
```diff
@@ -0,0 +1,455 @@
...
+            throw new NotFoundException();
+        }
+
+        return $results;
+    }
+
+    /**
+     * Extracts the finder name and options out of the provided pagination options
+     *
+     * @param array $options the pagination options
```
inoas: .

src/ORM/Paginator.php
```diff
@@ -0,0 +1,455 @@
...
+     * @return array An array containing in the first position the finder name and
+     * in the second the options to be passed to it
```
inoas: .

src/ORM/Paginator.php
```diff
@@ -0,0 +1,455 @@
...
+
+        if ($requestedPage > $page) {
+            throw new NotFoundException();
+        }
+
+        return $results;
+    }
+
+    /**
+     * Extracts the finder name and options out of the provided pagination options
```
inoas: .

src/ORM/Paginator.php
```diff
@@ -0,0 +1,455 @@
...
+     * Pulls settings together from the following places:
+     *
+     * - General pagination settings
+     * - Model specific settings.
+     * - Request parameters
+     *
+     * The result of this method is the aggregate of all the option sets combined together. You can change
+     * config value `whitelist` to modify which options/values can be set using request parameters.
+     *
+     * @param string $alias Model alias being paginated, if the general settings has a key with this value
```
inoas: .

src/ORM/Paginator.php
```diff
@@ -0,0 +1,455 @@
...
+
+        return $options;
+    }
+
+    /**
+     * Prefixes the field with the table alias if possible.
+     *
+     * @param \Cake\Datasource\RepositoryInterface $object Repository object.
+     * @param array $order Order array.
+     * @param bool $whitelisted Whether or not the field was whitelisted
```
inoas: .

src/ORM/Paginator.php
```diff
@@ -0,0 +1,455 @@
...
+        }
+
+        return $tableOrder;
+    }
+
+    /**
+     * Check the limit parameter and ensure it's within the maxLimit bounds.
+     *
+     * @param array $options An array of options with a limit key to be checked.
+     * @return array An array of options for pagination
```
inoas: .

src/ORM/Paginator.php
dereuromark: Or/Also an AbstractPaginator one holding the actual "common code" to keep things DRY when actually using custom implementations.

src/ORM/Paginator.php
```diff
@@ -0,0 +1,455 @@
...
+     * for pagination data. An example URL paginating both results would be:
+     *
+     * ```
+     * /dashboard?articles[page]=1&tags[page]=2
+     * ```
+     *
+     * @param \Cake\Datasource\RepositoryInterface|\Cake\Datasource\QueryInterface $object The table or query to paginate.
+     * @param array $params Request params
+     * @param array $settings The settings/configuration used for pagination.
+     * @return \Cake\Datasource\ResultSetInterface Query results
```
inoas: .

src/ORM/Paginator.php
```diff
@@ -0,0 +1,455 @@
...
+ * CakePHP(tm) : Rapid Development Framework (http://cakephp.org)
+ * Copyright (c) Cake Software Foundation, Inc. (http://cakefoundation.org)
+ *
+ * Licensed under The MIT License
+ * For full copyright and license information, please see the LICENSE.txt
+ * Redistributions of files must retain the above copyright notice.
+ *
+ * @copyright     Copyright (c) Cake Software Foundation, Inc. (http://cakefoundation.org)
+ * @link          http://cakephp.org CakePHP(tm) Project
+ * @since         3.4.8
```
inoas: 3.5 for a big and possibly breaking change?

src/ORM/Paginator.php
robertpustulka: Abstract class for shared code + interface for specific implementations sounds good.

src/ORM/Paginator.php
dereuromark: It would be better to write and submit a sniffer for this instead of adding noise comments for this very trivial thing 10+ times in a PR, cloaking the more relevant issues at hand.

src/ORM/Paginator.php
robertpustulka: IMO this method should accept only `QueryInterface` objects.

src/ORM/Paginator.php
robertpustulka: I couldn't figure out what all these dot comments meant... Wouldn't it be clearer just to write "Missing dots at the end of the sentence" once?

inoas: Sorry about that.

lorenzo:
> @lorenzo Would an abstract class help that could share the common code for both core and custom extension points?

Probably not, i was trying to do something with Abstract classes, but reality is that custom paginates actually share little. That's why I abandoned implementing it on my own.

I would like to see the param validation and cleanup perhaps as different objects that can be composed

ADmad:
Rebased and re-targeted to 3.next.

@lorenzo All the `Paginator` methods that do options juggling now take required data as arguments so they can be easily moved in a separate class. I created new exception class as suggested.

@robertpustulka Having an interface sounds good. I believe the interface would just have `paginate()` and `getPagingParams()` methods. Your suggestion of making `paginate()` only accept `QueryInterface` object also sounds good. I'll look into that.

@markstory Currently I am just relying on paginator component tests to ensure BC. I will add direct tests for the `Paginator` class too when code settles down.

@ionas Thanks for the comment spam.

src/Controller/Component/PaginatorComponent.php
```diff
@@ -275,155 +224,122 @@ protected function _extractFinder($options)
...
-        unset($options['sort'], $options['direction']);
+        $request = $this->_registry->getController()->request;
 
-        if (empty($options['order'])) {
-            $options['order'] = [];
-        }
-        if (!is_array($options['order'])) {
-            return $options;
+        if (!$request->getParam('paging')) {
+            $request->params['paging'] = [];
```
saeideng: used deprecated params ?

tests/TestCase/Controller/Component/PaginatorComponentTest.php
```diff
@@ -70,8 +78,8 @@ public function setUp()
         $this->request = new ServerRequest('controller_posts/index');
         $this->request->params['pass'] = [];
```
saeideng: and here

src/Controller/Component/PaginatorComponent.php
```diff
@@ -275,155 +224,122 @@ protected function _extractFinder($options)
...
         }
 
-        $inWhitelist = false;
-        if (isset($options['sortWhitelist'])) {
-            $field = key($options['order']);
-            $inWhitelist = in_array($field, $options['sortWhitelist'], true);
-            if (!$inWhitelist) {
-                $options['order'] = [];
+        $request->params['paging'] = $this->_paginator->getPagingParams()
```
saeideng: and here

ADmad:
@robertpustulka I looked into making `Paginator` accept only `QueryInterface` only but the fact that you can set custom finders in paginator settings really complicates things. The custom finder is called after the options merging is done. Also even if you have finder configured in setting if you pass a query to `PaginatorComponent::paginate()` then the finder won't be called.

robertpustulka:
@ADmad can't you call custom finders on a query object?

ADmad:
@robertpustulka Not directly, would have to get the repository instance from the query and call finder on that.

robertpustulka:
@ADmad wouldn't that work?
$query->find('foo')->find('bar')

Yoy can call configured finders on a query itself.

ADmad:
It would but as I mentioned earlier if a `QueryInterface` instance is directly passed to `PaginatorComponent::paginate()` then finders are not used, the options are directly applied to the query. So if the new `Paginator::paginator()` only takes a `QueryInterface` instance we know need a new option to indicate whether to use finder or not, that's sounds icky to me.

ADmad:
I have added `PaginatorInterface`.

I think the paginator class and the interface belong under `Datasource` instead of `ORM` package.

robertpustulka:
What about dropping support for finders in Paginator opions and leave it only in a component for BC? In that case a fully composed would be required for a paginator.

ADmad:
@robertpustulka In that case the component won't be able ti simply proxy the paginate call to the new paginator since currently the options merging is done before calling custom finder.

ADmad:
@markstory Tests added for `Paginator` class.

I think I have completed what I set out to achieve initially. Would someone like to follow up on @robertpustulka's suggestion of making paginator only accept `QueryInterface` and @lorenzo's suggestion of having the request params and options juggling in a separate class for easy composition?

ADmad:
@robertpustulka Making `PaginatorInterface::paginate()` only accept `QueryInterface` instance would be limiting for custom paginators. For e.g. the elastic search plugin's `Query` class doesn't implement `QueryInterface`. So not sure if it's worth pursuing.

src/Controller/Component/PaginatorComponent.php
```diff
@@ -275,155 +225,119 @@ protected function _extractFinder($options)
...
+     * Set paginator instance.
      *
-     * @param string $alias Model name to get settings for.
-     * @param array $settings The settings which is used for combining.
-     * @return array An array of pagination settings for a model, or the general settings.
+     * @param \Cake\Datasource\Paginator $paginator Paginator instance.
+     * @return self
      */
-    public function getDefaults($alias, $settings)
+    public function setPaginator(Paginator $paginator)
```
markstory: Shouldn't you hint to the interface?

src/Controller/Component/PaginatorComponent.php
```diff
@@ -275,155 +225,119 @@ protected function _extractFinder($options)
...
-                $direction = 'asc';
-            }
-            $options['order'] = [$options['sort'] => $direction];
-        }
-        unset($options['sort'], $options['direction']);
-
-        if (empty($options['order'])) {
-            $options['order'] = [];
+        $return = $this->_paginator->config($key, $value, $merge);
+        if ($return instanceof Paginator) {
```
markstory: interface again?

src/Controller/Component/PaginatorComponent.php
ADmad: I should, forgot to update after I added the interface.

markstory:
I think these changes might let us solve pagination in the elastic-search plugin as well. As the ES plugin will only need to implement this new interface :+1:

markstory:
Looks good. There are conflicts now though :cry:

markstory (approved):

ADmad:
@markstory Conflict fixed.

PR #10719, opened 2017-06-05
Deprecate RouteCollection::extensions()

markstory:
Another combined get/set method.

Refs #9978


tests/TestCase/Routing/RouteCollectionTest.php
```diff
@@ -593,9 +593,10 @@ public function testAddingDuplicateNamedRoutes()
     }
 
     /**
-     * Test basic get/set of extensions.
+     * Test get/set combined method.
```
ravage84: Test the combined get/set extension method.

ravage84 (approved):
Looks good apart from the test method doc block

PR #10720, opened 2017-06-05
Fixed a regression in ValidatorAwareTrait

robertpustulka:
#10707 introduced a regression where validators defined in behaviors were no longer supported.

There's a fix and a test case to ensure that behavior validation methods are fired.


lorenzo:
Good job!

PR #10726, opened 2017-06-06
[RFC] 3.next Cache fallback

jeremyharris:
Missing or broken cache engines can cause applications to fail, and in my opinion, cache is *optional* and shouldn't break an app. Since engines are lazy loaded, it's difficult to predict that an engine will fail by catching the exception unless you build all engines during bootstrap, defeating the lazy loading.

This PR adds the ability to define fallback engines within cache config. An example might be falling back to the FileEngine if Redis fails to connect:

```php
Cache::setConfig('tests', [
    'engine' => 'Redis',
    'host' => 'badhost',
    'fallback' => 'tests_fallback'
]);
Cache::setConfig('tests_fallback', [
    'engine' => 'File',
    'path' => TMP,
    'prefix' => 'test_',
]);
```

A developer could configure fallbacks all the way to the NullEngine in order to prevent their app from ever throwing an exception due to missing cache.

I'd like to discuss potential pitfalls (especially regarding groups, as I do not have much experience with them) of this change. Also, hiding errors isn't a good thing, so ideas on logging those errors are welcome.

### DIFF
--- a/src/Cache/Cache.php
+++ b/src/Cache/Cache.php
@@ -175,10 +175,10 @@ class Cache
             if (!array_key_exists('fallback', $config)) {
                 throw $e;
             }
-            $fallbackEngine = static::engine($config['fallback']);
-            static::getRegistry()->set($name, $fallbackEngine);
-
-            return;
+            $fallbackEngine = clone static::engine($config['fallback']);
+            $copyableConfig = ['groups' => null, 'prefix' => null];
+            $fallbackEngine->init(array_intersect_key($config, $copyableConfig));
+            $registry->set($name, $fallbackEngine);
         }
 
         if ($config['className'] instanceof CacheEngine) {
--- a/src/Cache/Cache.php
+++ b/src/Cache/Cache.php
@@ -178,6 +178,13 @@ class Cache
 
                 return;
             }
+
+            if ($config['fallback'] === $name) {
+                throw new InvalidArgumentException(
+                    sprintf('"%s" cache configuration cannot fallback to itself.', $name)
+                );
+            }
+
             $fallbackEngine = clone static::engine($config['fallback']);
             $copyableConfig = ['groups' => null, 'prefix' => null];
             $fallbackEngine->init(array_intersect_key($config, $copyableConfig));
--- a/src/Cache/Cache.php
+++ b/src/Cache/Cache.php
@@ -186,8 +186,11 @@ class Cache
             }
 
             $fallbackEngine = clone static::engine($config['fallback']);
-            $copyableConfig = ['groups' => null, 'prefix' => null];
-            $fallbackEngine->init(array_intersect_key($config, $copyableConfig));
+            $newConfig = $config + ['groups' => [], 'prefix' => null];
+            $fallbackEngine->setConfig('groups', $newConfig['groups'], false);
+            if ($newConfig['prefix']) {
+                $fallbackEngine->setConfig('prefix', $newConfig['prefix'], false);
+            }
             $registry->set($name, $fallbackEngine);
         }
 
--- a/src/Cache/Cache.php
+++ b/src/Cache/Cache.php
@@ -175,6 +175,7 @@ class Cache
             if (!array_key_exists('fallback', $config)) {
                 $registry->set($name, new NullEngine());
                 trigger_error($e->getMessage(), E_USER_WARNING);
+
                 return;
             }
             $fallbackEngine = clone static::engine($config['fallback']);
--- a/src/Cache/Cache.php
+++ b/src/Cache/Cache.php
@@ -173,7 +173,9 @@ class Cache
             $registry->load($name, $config);
         } catch (RuntimeException $e) {
             if (!array_key_exists('fallback', $config)) {
-                throw $e;
+                $registry->set($name, new NullEngine());
+                trigger_error($e->getMessage(), E_USER_WARNING);
+                return;
             }
             $fallbackEngine = clone static::engine($config['fallback']);
             $copyableConfig = ['groups' => null, 'prefix' => null];
--- a/src/Cache/Cache.php
+++ b/src/Cache/Cache.php
@@ -168,7 +168,18 @@ class Cache
         }
 
         $config = static::$_config[$name];
-        $registry->load($name, $config);
+
+        try {
+            $registry->load($name, $config);
+        } catch (RuntimeException $e) {
+            if (!array_key_exists('fallback', $config)) {
+                throw $e;
+            }
+            $fallbackEngine = static::engine($config['fallback']);
+            static::getRegistry()->set($name, $fallbackEngine);
+
+            return;
+        }
 
         if ($config['className'] instanceof CacheEngine) {
             $config = $config['className']->getConfig();
--- a/src/Cache/Engine/FileEngine.php
+++ b/src/Cache/Engine/FileEngine.php
@@ -427,7 +427,8 @@ class FileEngine extends CacheEngine
             //@codingStandardsIgnoreEnd
         }
 
-        if ($success && $this->_init && !($dir->isDir() && $dir->isWritable())) {
+        $isWritableDir = ($dir->isDir() && $dir->isWritable());
+        if (!$success || ($this->_init && !$isWritableDir)) {
             $this->_init = false;
             trigger_error(sprintf(
                 '%s is not writable',
--- a/src/Cache/Engine/FileEngine.php
+++ b/src/Cache/Engine/FileEngine.php
@@ -420,21 +420,22 @@ class FileEngine extends CacheEngine
     {
         $dir = new SplFileInfo($this->_config['path']);
         $path = $dir->getPathname();
+        $success = true;
         if (!is_dir($path)) {
-            mkdir($path, 0775, true);
+            //@codingStandardsIgnoreStart
+            $success = @mkdir($path, 0775, true);
+            //@codingStandardsIgnoreEnd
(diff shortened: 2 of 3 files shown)


markstory:
> Also, hiding errors isn't a good thing, so ideas on logging those errors are welcome.

Could `Cache` log errors when a fallback is used? That seems like an error condition to me.

src/Cache/Cache.php
```diff
@@ -168,7 +168,18 @@ protected static function _buildEngine($name)
         }
 
         $config = static::$_config[$name];
-        $registry->load($name, $config);
+
+        try {
+            $registry->load($name, $config);
+        } catch (RuntimeException $e) {
+            if (!array_key_exists('fallback', $config)) {
```
markstory: Logging could go here.

src/Cache/Cache.php
```diff
@@ -168,7 +168,18 @@ protected static function _buildEngine($name)
...
+                throw $e;
+            }
+            $fallbackEngine = static::engine($config['fallback']);
```
markstory: The groups from the original engine should probably be applied to the fallback as well. You might want to use the fallback configuration as a 'template' for the actual fallback engine and use it to replace the missing/broken engine.

src/Cache/Engine/FileEngine.php
```diff
@@ -420,21 +420,22 @@ protected function _active()
...
+            //@codingStandardsIgnoreEnd
         }
 
-        if ($this->_init && !($dir->isDir() && $dir->isWritable())) {
+        if ($success && $this->_init && !($dir->isDir() && $dir->isWritable())) {
             $this->_init = false;
             trigger_error(sprintf(
                 '%s is not writable',
                 $this->_config['path']
             ), E_USER_WARNING);
```
markstory: I wonder if this should be an error log instead of triggered error.

src/Cache/Cache.php
jeremyharris: That means https://github.com/cakephp/cache would require `cakephp/log`, yeah? I'm not sure how the splits are managed, so not sure what I would need to do to allow it.

src/Cache/Engine/FileEngine.php
jeremyharris: Yeah or an exception to be caught. I'm wondering if these `trigger_error()` were leftovers from past versions.

jeremyharris:
I probably need a better test case because appveyor is failing. Additional test case ideas are encouraged to feel good about this change (e.g. one for groups, which I will write if people think this is a good idea).

src/Cache/Cache.php
ADmad: Depending on how long the cache engine is down and how many requests are received in the mean time, is there a risk of potentially filling up the logs quite quickly by logging here?

jeremyharris: I don't think it would log anymore than a normal exception due to a failed init would log, since building only happens once per request and the "bad" object is replaced in the registry.

markstory: @ADmad That could happen, I think silently dropping the failure is not good though.

src/Cache/Engine/FileEngine.php
markstory: Triggering an error is a way for us to avoid coupling to cakephp/log, which is probably another reason for why it is this way now.

src/Cache/Cache.php
ADmad: Yeah, logging should happen. Only thing​ it adds dependency on log package as @jeremeyharris mentioned.

markstory: We could use `trigger_error` and rely on a logging being hooked up to an error logger / SAPI logger.

ADmad: Sounds good.

lorenzo: can we make all engines have a fallback? Just set the fallback to the `Null` engine, this way we avoid the check

jeremyharris: @lorenzo I was considering that, however it changes the default behavior so I hesitated. Do you think that's a big deal or not?

jeremyharris:
Okay, finally got groups working with fallbacks properly. (I had to modify existing tests as they were relying on a lot of Cache state that broke.)

src/Cache/Cache.php
lorenzo: I think it is a good default... Not sure what the others think

markstory: I'm up for that. Caching shouldn't blow an app up if its not hooked up.

tests/TestCase/Cache/CacheTest.php
```diff
@@ -332,12 +439,18 @@ public function testGroupConfigs()
...
+        $this->assertContains('latest', $result['posts']);
+        $this->assertContains('page', $result['posts']);
+
+        $this->assertArrayHasKey('comments', $result);
+        $this->assertContains('latest', $result['comments']);
+        $this->assertNotContains('page', $result['comments']);
+
+        $this->assertArrayHasKey('archive', $result);
+        $this->assertContains('page', $result['archive']);
+        $this->assertNotContains('latest', $result['archive']);
```
markstory: This looks good :+1:

src/Cache/Cache.php
jeremyharris: I'll go ahead and implement the null engine as the default fallback 👍

src/Cache/Engine/FileEngine.php
ravage84: @markstory it might be helpful if we document this way in our docs.

src/Cache/Cache.php
```diff
@@ -168,7 +168,21 @@ protected static function _buildEngine($name)
...
+        } catch (RuntimeException $e) {
+            if (!array_key_exists('fallback', $config)) {
+                $registry->set($name, new NullEngine());
+                trigger_error($e->getMessage(), E_USER_WARNING);
+
+                return;
+            }
+            $fallbackEngine = clone static::engine($config['fallback']);
+            $copyableConfig = ['groups' => null, 'prefix' => null];
+            $fallbackEngine->init(array_intersect_key($config, $copyableConfig));
```
markstory: If the groups are only passed into the init() will the state in `Cache::$_groups` be correct afterwords?

jeremyharris:
Anything else anyone would like to see in this? I've implemented falling back to NullEngine (and triggering an error) if no fallback is found or all the fallback configs fail.

Not sure why but the builds are broken and need to be reset.

src/Cache/Cache.php
jeremyharris: There are essentially two configs going on here, the static one in the cache frontend `Cache::$_groups` and what the engine needs (group, prefix) to create the proper prefixes to actually find the file. 

`Cache::$_groups` needs to know that even though "test" fails, if you reference a group it had it will still write (using the fallback). The fallback engine then needs to use those group configs to construct the proper prefixes.

This dual config caused me a bit of confusion; perhaps I need more tests for us to be totally confident about them?

markstory: Ok. More tests covering group writes/clears might be good to have in general if we can avoid coupling to the guts of the cache frontend.

jeremyharris: There are a couple currently:

- `testCacheFallbackWithGroups`: tests original config is still associated with its groups, fallback config is associated with only its original groups
- `testCacheFallbackIntegration`: tests clearing group of original config causes reading from that config's fallback fails properly (when reading from original config)

I will add some more that write/clear groups both in the original and the fallback to make sure they don't get confused somehow.

src/Cache/Engine/FileEngine.php
lorenzo: @jeremyharris Do you mind also changing the error message here? It would be better if it says something like `%s is not writable. When trying to store cache data for config %s`

jeremyharris: @lorenzo I sure can!

jeremyharris: @lorenzo actually, I sure can't! Engines are not aware of their cache config name.

src/Cache/Cache.php
```diff
@@ -178,9 +178,19 @@ protected static function _buildEngine($name)
 
                 return;
             }
+
+            if ($config['fallback'] === $name) {
+                throw new InvalidArgumentException(
+                    sprintf('"%s" cache configuration cannot fallback to itself.', $name)
+                );
```
markstory: :+1:

markstory:
@jeremyharris Is this ready to be merged? If so would you be able to update the docs for this feature?

jeremyharris:
@markstory yes and yes. I'll have a docs PR up soon

AD7six:
Hmm I know this is long merged but I have some concerns here.

this is before this PR, applied to my use case:
* app cache is broken
* nothing works
* damage limited to php app

this is now, applied to my use case:
* app cache is broken
* app runs with no cache
* db and upstream service load explodes
* damage affects everything using db+upstream services

I'm not sure how to achieve achieve the previous in-your-face behavior of an exception without cheating to trigger `if ($config['fallback'] === $name) {`

Note that this PR only catches bad config - it doesn't afaics handle a redis read/write fail which would still result in an uncaught RedisException.

markstory:
@AD7six Would having a cache engine implementation that raises exceptions on every operation help? It could act as a `FailEngine` and act as a circuit breaker for cache fallbacks.

AD7six:
@markstory I missed this comment, could do - I think the above PR is a simpler change though

PR #10731, opened 2017-06-07
Consistent unicode logging

jzdunek:
The current behaviour of CakePHP regarding the logging of unicode characters is inconsistent. Strings and arrays containing unicode characters will be logged as such, keeping the unicode characters readable. Entity properties and objects implementing JsonSerializable with unicode characters will be logged with all unicode characters escaped as \uXXXX, making them very hard to read and to search for.

As CakePHP is unicode aware in general, I'd like to suggest implementing a consistent logging of unicode characters over all types of parameters. Unicode characters should always be logged as such, keeping them readable and searchable.

This can be achieved by using `JSON_UNESCAPED_UNICODE` with `json_encode()`. This applies to `BaseLog::_format()` and `EntityTrait::__toString()` as those are the functions participating in logging message and objects.

### DIFF
--- a/src/Datasource/EntityTrait.php
+++ b/src/Datasource/EntityTrait.php
@@ -1231,7 +1231,7 @@ trait EntityTrait
      */
     public function __toString()
     {
-        return json_encode($this, JSON_PRETTY_PRINT);
+        return json_encode($this, JSON_PRETTY_PRINT | JSON_UNESCAPED_UNICODE);
     }
 
     /**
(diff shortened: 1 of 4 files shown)


jzdunek:
AppVeyor CI fails because I'm using a PHP 7 feature (anonymous classes) in the new test case `BaseLogTest`. If this is a problem please tell me and I'll find a way around it...

cleptric:
@jzdunek CakePHP is supporting >= php 5.6, so you unfortunately can't use any exclusive php7 features.

jzdunek:
Closing the pull request. I'll create a new pull request that supports >= PHP 5.6.

cleptric:
@jzdunek If you want, you could also change the code and squash your commits in this PR 🙂

jzdunek:
@cleptric Thanks for the hint - I'll try to figure how **that** works some time later :wink:

PR #10732, opened 2017-06-07
Added support for cookie Max-Age.

robertpustulka:
Most modern browsers support `Max-Age` directive. 

`Max-Age` should have a precedence over `Expires` if both of these are present (ie. https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie).


markstory:
Looks good. Should we also support `Max-Age` in the cookie objects?

robertpustulka:
@markstory Im not sure. If you want your cookie to expire in 3600 seconds you can set expires to `new Chronos('1 hour)`. It's basically the same thing.

markstory:
@robertpustulka Good point. We can always add more methods later :smile:

PR #10735, opened 2017-06-07
Fix type static string and make const

mehdi-fathi:
Fix type static string and make const and change name of var for make better code.

### DIFF
--- a/src/View/View.php
+++ b/src/View/View.php
@@ -197,7 +197,7 @@ class View implements EventDispatcherInterface
      *
      * @var array
      */
-    public $uuIds = [];
+    public $uuids = [];
 
     /**
      * An instance of a \Cake\Http\ServerRequest object that contains information about the current request.
@@ -299,6 +299,13 @@ class View implements EventDispatcherInterface
      * @var string
      */
     const TYPE_ELEMENT = 'element';
+
+    /**
+     * Constant for name of view file 'Element'
+     *
+     * @var string
+     */
+    const NAME_ELEMENT = 'Element';
 
     /**
      * Constant for view file type 'layout'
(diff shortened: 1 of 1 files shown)


src/View/View.php
```diff
@@ -197,7 +197,7 @@ class View implements EventDispatcherInterface
      *
      * @var array
      */
-    public $uuids = [];
+    public $uuIds = [];
```
markstory: This is a breaking change. You cannot rename public properties.

src/View/View.php
```diff
@@ -830,7 +837,7 @@ public function extend($name)
                     if (!$parent) {
                         list($plugin, $name) = $this->pluginSplit($name);
                         $paths = $this->_paths($plugin);
-                        $defaultPath = $paths[0] . 'Element' . DIRECTORY_SEPARATOR;
+                        $defaultPath = $paths[0] . ucfirst(static::TYPE_ELEMENT) . DIRECTORY_SEPARATOR;
```
markstory: Adding a function call adds overhead. Why not add a constant with the appropriate value?

markstory: Perhaps `NAME_ELEMENT`?

markstory (changes requested):
You have breaking changes right now. The public property change needs to be removed for this to be mergeable.

src/View/View.php
mehdi-fathi: Yes NAME_ELEMENT is better than

src/View/View.php
```diff
@@ -1,16 +1,16 @@
 <?php
 /**
- * CakePHP(tm) : Rapid Development Framework (http://cakephp.org)
- * Copyright (c) Cake Software Foundation, Inc. (http://cakefoundation.org)
+ * CakePHP(tm) : Rapid Development Framework (https://cakephp.org)
```
ADmad: Undo these http to https changes. This header needs to stay consistent across all files.

You can make a separate PR to change it for all files if you want.

ADmad (changes requested):

src/View/View.php
mehdi-fathi: Yes i think a separate commit is good for this work.Because of almost all of file's url cakephp.org has this problem.

markstory:
Do you have a scenario where you needed to change the constants?

mehdi-fathi:
@markstory Actually i don't have.But it better than former code.When you use a string can will be use more than one time.Why it be not a constants.

markstory:
I wasn't arguing for not having constants. Just wondering what your usecase/motivation was to change this :smile:

markstory (approved):

mehdi-fathi:
Nothing.just a name of element constant for access another place.maybe there is no need.

PR #10736, opened 2017-06-07
Consistent unicode logging

jzdunek:
The current behaviour of CakePHP regarding the logging of unicode characters is inconsistent. Strings and arrays containing unicode characters will be logged as such, keeping the unicode characters readable. Entity properties and objects implementing JsonSerializable with unicode characters will be logged with all unicode characters escaped as \uXXXX, making them very hard to read and to search for.

As CakePHP is unicode aware in general, I'd like to suggest implementing a consistent logging of unicode characters over all types of parameters. Unicode characters should always be logged as such, keeping them readable and searchable.

This can be achieved by using `JSON_UNESCAPED_UNICODE` with `json_encode()`. This applies to `BaseLog::_format()` and `EntityTrait::__toString()` as those are the functions participating in logging message and objects.

### DIFF
--- a/src/Datasource/EntityTrait.php
+++ b/src/Datasource/EntityTrait.php
@@ -1231,7 +1231,7 @@ trait EntityTrait
      */
     public function __toString()
     {
-        return json_encode($this, JSON_PRETTY_PRINT | JSON_UNESCAPED_UNICODE);
+        return json_encode($this, JSON_PRETTY_PRINT);
     }
 
     /**
--- a/src/Datasource/EntityTrait.php
+++ b/src/Datasource/EntityTrait.php
@@ -1231,7 +1231,7 @@ trait EntityTrait
      */
     public function __toString()
     {
-        return json_encode($this, JSON_PRETTY_PRINT);
+        return json_encode($this, JSON_PRETTY_PRINT | JSON_UNESCAPED_UNICODE);
     }
 
     /**
--- a/src/Log/Engine/BaseLog.php
+++ b/src/Log/Engine/BaseLog.php
@@ -15,6 +15,7 @@
 namespace Cake\Log\Engine;
 
 use Cake\Core\InstanceConfigTrait;
+use Cake\Datasource\EntityInterface;
 use JsonSerializable;
 use Psr\Log\AbstractLogger;
 
@@ -94,6 +95,10 @@ abstract class BaseLog extends AbstractLogger
         }
 
         $object = is_object($data);
+
+        if ($object && $data instanceof EntityInterface) {
+            return json_encode($data, JSON_UNESCAPED_UNICODE | JSON_PRETTY_PRINT);
+        }
 
         if ($object && method_exists($data, '__toString')) {
             return (string)$data;
--- a/src/Log/Engine/BaseLog.php
+++ b/src/Log/Engine/BaseLog.php
@@ -100,7 +100,7 @@ abstract class BaseLog extends AbstractLogger
         }
 
         if ($object && $data instanceof JsonSerializable) {
-            return json_encode($data);
+            return json_encode($data, JSON_UNESCAPED_UNICODE);
         }
 
         return print_r($data, true);
--- a/tests/TestCase/Log/Engine/BaseLogTest.php
+++ b/tests/TestCase/Log/Engine/BaseLogTest.php
@@ -95,7 +95,9 @@ class BaseLogTest extends TestCase
      */
     public function testLogUnicodeObjectToString()
     {
-        $stub = $this->createMock(Entity::class);
+        $stub = $this->getMockBuilder(\stdClass::class)
+            ->setMethods(['__toString'])
+            ->getMock();
         $stub->method('__toString')
             ->willReturn(implode($this->testData));
 
@@ -115,6 +117,18 @@ class BaseLogTest extends TestCase
             ->willReturn($this->testData);
 
         $logged = $this->logger->log(LogLevel::INFO, $stub);
+
+        $this->assertUnescapedUnicode($this->testData, $logged);
+    }
+
+    /**
+     * Tests the logging output of an entity with property value that contains unicode characters.
+     */
+    public function testLogUnicodeEntity()
+    {
+        $entity = new Entity(['foo' => implode($this->testData)]);
+
+        $logged = $this->logger->log(LogLevel::INFO, $entity);
 
         $this->assertUnescapedUnicode($this->testData, $logged);
     }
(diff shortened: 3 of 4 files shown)


src/Datasource/EntityTrait.php
```diff
@@ -1231,7 +1231,7 @@ public function source($alias = null)
      */
     public function __toString()
     {
-        return json_encode($this, JSON_PRETTY_PRINT);
+        return json_encode($this, JSON_PRETTY_PRINT | JSON_UNESCAPED_UNICODE);
```
markstory: This could change/break existing applications that are relying on string casts to be usable in javascript code. I'm not sure that unescaped unicode will work in all javascript engines.

src/Log/Engine/BaseLog.php
```diff
@@ -100,7 +100,7 @@ protected function _format($data, array $context = [])
         }
 
         if ($object && $data instanceof JsonSerializable) {
-            return json_encode($data);
+            return json_encode($data, JSON_UNESCAPED_UNICODE);
```
markstory: :+1: this looks reasonable.

tests/TestCase/Log/Engine/BaseLogTest.php
```diff
@@ -0,0 +1,121 @@
...
+
+    /**
+     * Tests the logging output of an object implementing __toString().
+     * Note: __toString() will return a single string containing unicode characters.
+     */
+    public function testLogUnicodeObjectToString()
+    {
+        $stub = $this->createMock(Entity::class);
+        $stub->method('__toString')
+            ->willReturn(implode($this->testData));
```
markstory: You don't need a mock here. Just use an entity.

tests/TestCase/Log/Engine/BaseLogTest.php
```diff
@@ -0,0 +1,121 @@
...
+use Cake\ORM\Entity;
+use Cake\TestSuite\TestCase;
+use Psr\Log\LogLevel;
+
+/**
+ * Class BaseLogImpl
+ * Implementation of abstract class {@see Cake\Log\Engine\BaseLog},
+ * required by test case {@see Cake\Test\TestCase\Log\Engine\BaseLogTest}.
+ */
+class BaseLogImpl extends BaseLog
```
markstory: Isn't it more important that each of the implementations of BaseLog have the unicode handling behavior? Seems like the tests you have here should be a trait that is included in the test cases for each implementation.

src/Datasource/EntityTrait.php
jzdunek: @markstory D'oh! Haven't though of that use case...

What do you think about using a protected field `_json_encode_options` defaulting to `JSON_PRETTY_PRINT`? So everone wanting unescaped unicode could set the options accordingly in his/her own entities.

tests/TestCase/Log/Engine/BaseLogTest.php
jzdunek: @markstory Fair point. On the other hand the inconsistent handling of unicode is implemented in `BaseLog` and not in the sub classes. Therefore I decided for this pragmatic approach for testing the behaviour.

Perhabs it is a good idea to extract the `_format()` function into some kind of LogFormatter class that could be tested independently.

But for now I hope that this hands-on approach for testing the change is good enough.

src/Datasource/EntityTrait.php
markstory: Couldn't we re-jig logging to json format entities instead of string casting them? Then we wouldn't need any entity changes to afford this use case.

markstory: That does take into account scenarios where people are sprintf() or concatting entities into log messages though.

tests/TestCase/Log/Engine/BaseLogTest.php
markstory: > Fair point. On the other hand the inconsistent handling of unicode is implemented in BaseLog and not in the sub classes. 

For now. In the future if BaseLog was refactored, or the implementations override the parent behavior we'd have accidental regressions.

src/Datasource/EntityTrait.php
jzdunek: Do you think about something like this in `BaseLog::_format()`?
```
        $object = is_object($data);

        if ($object && $data instanceof EntityTrait) {
            return json_encode($data, JSON_UNESCAPED_UNICODE);
        }

        if ($object && method_exists($data, '__toString')) {
            return (string)$data;
        }

        if ($object && $data instanceof JsonSerializable) {
            return json_encode($data, JSON_UNESCAPED_UNICODE);
        }
```
That would leave the logging scenarios sprintf() and concatting entities into log messages unchanged (escaping unicode). Is that your intention here?

markstory: You could combine the 1st and 3rd conditions. But otherwise that change looks good to me.

jzdunek: Update:
```
       $object = is_object($data);

        if ($object && ($data instanceof EntityTrait || $data instanceof JsonSerializable)) {
            return json_encode($data, JSON_UNESCAPED_UNICODE);
        }

        if ($object && method_exists($data, '__toString')) {
            return (string)$data;
        }
```

This changes the logging behaviour for everyone who does something like `$logger->log('debug', $entity)` with a custom logging format implemented by overwriting `EntityTrait::__toString()`. Don't you think that might take one or the other by surprise?

markstory: I think it that is a safer behaviour change than modifying how json serialization works by default.

src/Log/Engine/BaseLog.php
```diff
@@ -95,12 +96,16 @@ protected function _format($data, array $context = [])
...
+            return json_encode($data, JSON_UNESCAPED_UNICODE | JSON_PRETTY_PRINT);
+        }
+
         if ($object && method_exists($data, '__toString')) {
             return (string)$data;
         }
 
         if ($object && $data instanceof JsonSerializable) {
-            return json_encode($data);
+            return json_encode($data, JSON_UNESCAPED_UNICODE);
```
jzdunek: I decided to treat `EntityInterface` separate from `JsonSerializable` because I wanted to keep the `JSON_PRETTY_PRINT` used in `EntityTrait::__toString()`. Doing so for `JsonSerializable` as well would have changed even more of the current logging (including breaking `FileLogTest`).

src/Log/Engine/BaseLog.php
```diff
@@ -95,12 +96,16 @@ protected function _format($data, array $context = [])
 
         $object = is_object($data);
```
ravage84: Nitpicking: I would call it ``$isObject`` as it increased readability of the conditions below.

Issue #10734, opened 2017-06-07
Clause may we check if key exist before return


Aerue:
This is a (multiple allowed):

* [x] bug
* [x] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Application Skeleton Version: b1aaa094894c020cd49d093bf5d6111fe26f5c45
* Platform and Target:
    * Linux 3.16.0-4-amd64 #1 SMP Debian 3.16.39-1+deb8u2 (2017-03-07) x86_64 GNU/Linux
    * Debian 8.7
    * PHP 7.0.17-1~dotdeb+8.1 (cli) ( NTS )
    * nginx version: nginx/1.11.13
    * mysql  Ver 15.1 Distrib 10.0.30-MariaDB, for debian-linux-gnu (x86_64) using readline 5.2

### What you did

I make a Query who is cached, this is my query

```
self::$__pagebuilder = $Pagebuilders
            ->getAllPagebuildersActivated()
            ->matching('Templates', function($q) {
                return $q
                    ->where([
                        'Templates.template_id' => Template::TEMPLATE_NEWS_LIST,
                        'Templates.is_activated' => true
                    ])
                    ->order('Templates.id ASC')
                    ->limit(1);
            });
```

Into my Table i manage the Query and the cache

```
public function getAllPagebuildersActivated($type = 'all')
    {
        return $this
            ->find($type, [
                'conditions' => [
                    'Pagebuilders.is_activated' => true,
                    'Pagebuilders.type_id IN' => [Pagebuilder::PAGEBUILDER_TYPE_HOME, Pagebuilder::PAGEBUILDER_TYPE_PAGE]
                ]
            ])
            ->find('translations')
            ->cache(function($q) use ($type) {
                /** @var Query $q */
                $matchingQuery = $q->clause('matching');

                debug($matchingQuery);
                die();

                return 'get-all-pagebuilders-activated-' . strtolower($type) . '-' . md5(serialize($q->clause('where')));
            }, 'infinite');
    }
```

### What happened
`Notice (8): Undefined index: matching [CORE/src/Database/Query.php, line 1701]`

### What you expected to happen
The clause method of Query class should check if the key exist before return parts data.

Query.php method clause should be rework from

```
public function clause($name)
{
        return $this->_parts[$name];
}
```

to

```
public function clause($name)
{
        if (array_key_exists($name, $this->_parts)) {
            return $this->_parts[$name];
        }
        return null;
}
```

[...]

PR #10737, opened 2017-06-07
Don't emit errors when clauses are undefined.

markstory:
Refs #10734

### DIFF
--- a/src/Database/Query.php
+++ b/src/Database/Query.php
@@ -1698,7 +1698,7 @@ class Query implements ExpressionInterface, IteratorAggregate
      */
     public function clause($name)
     {
-        return $this->_parts[$name];
+        return isset($this->_parts[$name]) ? $this->_parts[$name] : null;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


lorenzo (commented):
Shouldn't it be an error though?

markstory:
@lorenzo The issue reporter thought it shouldn't be an error. If we're going to have an error it shouldn't be an `E_NOTICE`. I'm ok with raising a `RuntimeException` though.

lorenzo:
I agree it should not be a `E_NOTICE`. Looking again at the original error report, it would have been even more difficult for him to realise that `matching` is not a query clause, had we returned null instead of reporting the error.

Converting this to an exception sounds like a better idea.

markstory:
@lorenzo I added an error with what I think is a good error message.

markstory:
@lorenzo I added an error with what I think is a good error message.

lorenzo:
Looks great!

PR #10738, opened 2017-06-07
Make RouteBuilder::connect() return route instances.

markstory:
The utility of the fluent setters on Route are limited if connect() does not also return the route instances.


lorenzo:
Nice!

Issue #10318, opened 2017-02-27
Unnecessary joins when using deleteAll()


bancer:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [x] feature-discussion (RFC)

* CakePHP Version: 2.8.9
* Platform and Target: Linux, MySQL, Apache.

### What you did
```
$this->recursive = -1;
$this->deleteAll($conditions);
```

### What happened
The records were deleted but while deleting SELECT query was generated with unnecessary joins. I discovered that that is caused by hard coded `'recursive' => 0` in Model::deleteAll(). Those extra joins are not necessary and may affect the performance.

### What you expected to happen
I would expect that the recursive value set before the call to deleteAll() would be respected.

dereuromark:
It is a well known issue that in 2.x a lot of more joins are done for updateAll() and deleteAll() than necessary usually. It is partially also because it is harder to inspect the conditions and query and thus by default assumes those joins are necessary.

I recommend using updateAllJoinless() and deleteAllJoinless() from [Shim plugin](https://github.com/dereuromark/cakephp-shim/blob/2.x/docs/Model.md#updatealldeleteall) here, since those then flawlessly upgrade itself to 3.x updateAll() and deleteAll(). They completely reset the joins internally.

markstory:
We could take the min of `$this->recursive` and `0`. That would make it easier for developers to turn off the joins as needed.

dereuromark:
Based on the effect it can have on existing apps relying on the current behavior this would be safer to go into 2.10 then.

markstory:
Closing as a pull request is open.

sebastienbarre:
@bancer @dereuromark @markstory Oh man this *definitely* needs to be added to the migration guide, this change is wreaking havoc in our application.

Our `AppModel` has `public $recursive = -1;` so that we are in better control of the joins, but many of our `deleteAll()` are now failing because `$conditions` has conditions on *associated* models, which were brought in automatically by the previous implementation, where `'recursive' => 0`. 

While your fix is "the right thing to do", there was an assumption that `deleteAll()` was bringing associations in. Fixing this is not going to be elegant, something like:
```
    $old_recursive = $this->recursive;
    $this->recursive = 0;
    $this->deleteAll($conditions, true, true); // run the callbacks()
    $this->recursive = $old_recursive;
```

Just FYI :)

markstory:
[...]

PR #10739, opened 2017-06-07
Takes into account the current recursive value in deleteAll().

bancer:
Resolves https://github.com/cakephp/cakephp/issues/10318.


markstory:
Is there a test that could be added for this?

bancer:
There are tests in https://github.com/cakephp/cakephp/blob/2.x/lib/Cake/Test/Case/Model/ModelDeleteTest.php#L285. This change improves the performance but has no effect on the result of deleteAll().

Issue #10741, opened 2017-06-08
Discrepancy in Model::id when using ClassRegistry::init versus $this->loadModel


tersmitten:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 2.9.9
* Platform and Target: Apache, Mysql, PHP7

#### Code

```php
		$SubscriptionPeriod = ClassRegistry::init('SubscriptionPeriod', true);
		debug($SubscriptionPeriod->id);
		debug($SubscriptionPeriod->field('MIN(number_of_months)'));
		exit;
```

#### Output

```sh
/app/Controller/AppController.php (line 120)
false
/app/Controller/AppController.php (line 121)
'6'
```

#### Code

```php
		$this->loadModel('SubscriptionPeriod');
		debug($this->SubscriptionPeriod->id);
		debug($this->SubscriptionPeriod->field('MIN(number_of_months)'));
		exit;
```

#### Output

```sh
/app/Controller/AppController.php (line 125)
null
/app/Controller/AppController.php (line 126)
null
```

### What you expected to happen

So what I expected to happen is that both `Model::field` calls would result in returning `'6'`, but that's not the cause. It seems that `Model::field` behaves differently based on the value of `Model::id`.

Is this to be expected?

markstory:
> It seems that Model::field behaves differently based on the value of Model::id

Yes that is intentional behavior.

The second argument of `loadModel()` is the `$id` you want to initialize the model with. The default value is null. If you want it to be not-null you must provide the desired value. `ClassRegistry::init()` doesn't support passing an id, so I suspect that the `id` value is retained state from somewhere else in your application, or a model constructor.

tersmitten:
[This line](https://api.cakephp.org/2.9/source-class-Model.html#1656) is what's causing the strange behavior. I would expect it to behave the same for `null` and `false`.

dereuromark:
You could make a PR to include also false as check, that wound sound reasonable BC and patchy.

tersmitten:
I'll put something together

tersmitten:
Done

PR #10742, opened 2017-06-08
Fix discrepancy in Model::field when Model::id is null

tersmitten:
See #10741


dereuromark:
false and null are usually the same in those contexts, so this seems like a reasonable fix to me 👍

markstory:
Nice find!

Issue #10724, opened 2017-06-06
Security.cipherSeed is rounded to PHP_INT_MAX in 2.8 and existing cookies are broken


ikuwow:
This is a:

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 2.8.x
  * after this commit: https://github.com/cakephp/cakephp/commit/6a68032e0b6d6115e0c47ee5615a8ba82245561d
* Platform and Target: php5.6.30, apache 2.4

### What you did
I updated cakephp 2.7 to 2.8.

### What happened
All cookies broke (which already exists on client browsers)

### What you expected to happen

Security.cipherSeed is explicitly cast as int (for removing warning in php7):
https://github.com/cakephp/cakephp/commit/6a68032e0b6d6115e0c47ee5615a8ba82245561d

cipherSheed seems to be assumed to be larger than PHP_INT_MAX:
https://github.com/cakephp/cakephp/blob/2.x/app/Config/core.php#L240
（My project also uses large cipherSeed like this）

So after this change all cookies that is encrypted before by `Security` utility cannot be decoded correctly.

---

How can i update to 2.8 without breaking existing cookies?

markstory:
> How can i update to 2.8 without breaking existing cookies?

Sounds like you can't because your seed was too big.

ikuwow:
Oh... I see. Thank you for replying.

I think the TODOs are:

* Change Security.cipherSeed samples to a valid integer
  * in documents
  * in tests
  * in app skeleton (/app/Config/core.php)
* Add warnings of this problem in 2.8 migration guide

markstory:
@ikuwow Any chance you'd be willing to help with those tasks?

ikuwow:
@markstory Ok, i'll do these in a few days or a week 👍

ikuwow:
I found a workaround. It seems strange but it works. #10749

markstory:
Great work @ikuwow. Closing this as there is a pull request now.

PR #10749, opened 2017-06-10
Fix broken cookie issue

ikuwow:
#10724

This change makes Security::cipher() encoding and decoding same as 2.7 and below.


markstory:
I'm looking into why the build failed. I don't think it should be caused by your changes.

Issue #10750, opened 2017-06-10
i18nFormat support for ISO8601 timestamps


juliancarrivick:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.4.7
* Platform and Target: PHP 7

### What you did
I created a `Time` Object using the ISO 8601 format and with a `Z` as the timezone, then attempted to call `i18nFormat` on it.
i.e.
``` php
$time = new Time('2014-01-01T00:00:00Z');
$time->i18nFormat();
```

### What happened
An error is thrown: `Error: Call to a member function format() on null`

This is due to the timezone `Z` is passed directly to `datefmt_create()` in `DateFormatTrait->_formatObject()` which is not recognised as a valid timezone by `datefmt_create()` which then returns `false` which cannot be formatted resulting in the error:

### What you expected to happen
I expected the date to be formatted in the exact same way as if I had created it with an equivalent timezone such as `UTC` or `+00:00`.

juliancarrivick:
I have written a fix and test cases for this bug, and will create a pull request shortly.

PR #10751, opened 2017-06-10
issue-10750: Fix ISO8601 i18n formatting

juliancarrivick:
Addresses issue #10750.

When a Time or a Date is created using an ISO8601 UTC timezone, attempting to execute i18nFormat() on it results in an error and the timestamp not being formatted. This patch fixes that and adds the requisite tests.


markstory:
Thank you :+1:

Issue #10716, opened 2017-06-03
RFC - Console CommandRunner to integrate Application into CLI contexts


markstory:
The `Application` class added in 3.3 is not currently used in the CLI context. This will prevent `Application` from being the canonical way to bootstrap an application, load configuration, define routing and more. To remedy this the Console entry point needs to start using `Application`.

We have also had a number of requests to allow console 'routing' and allow creating standalone CLI tools more easily.

## Application::console hook

By adding a hook method to `Application` we can enable applications to declare their CLI commands or continue using the auto-discovery we have today. The `Application::console()` hook would be called after bootstrapping is complete, and is passed a `CommandCollection` that allows commands to be added and aliases to be set

```php
// In App\Application.

public function console(CommandCollection $collection)
{
   $collection->add('bake', BakeShell::class);
   $collection->add('asset_compress' AssetCompressShell::class);

   // Auto-discover commands in the app/plugin (current behavior)
   $collection->autoDiscover();

   return $collection;
}
```

In addition to the hook method a `Console.buildCommands` event would be emitted to enable plugins to register commands through the global event listener. This mirrors the `Server.buildMiddleware` event for the HTTP server.

Each added command defines its option parser and sub-commands as they do today. The additional console hook lets users define the commands they want loaded, with short aliases, and load commands from plugins. It also makes it easier to split the console libraries into a standalone library.

The `CommandCollection` class would provide the following methods:

- `addCommand($name, $command)` Add a command to the collection.
- `removeCommand($name)` Remove a command.
- `hasCommand($name)` Check if a command has been set.
- `autoDiscover()` Search the application, and plugins for commands.
- IteratorInterface for iterating attached commands.
- Other methods?

The `CommandCollection` will be initialized with two commands. `help/-h/--help` and `-v/--version` will map to commands that provide their current behavior. Users will be able to replace/remove these commands if they wish.

## Deprecate and replace ShellDispatcher

`ShellDispatcher` is the current entry point into the console environment. To preserve full backwards compatibility, `ShellDispatcher` would be left unchanged. A new class `Cake\Console\CommandRunner` would be added. This new class would be responsible for:

1. Accepting an `Application` instance and `$argv`.
2. Use `Application::bootstrap()` to bootstrap the application
3. Use `Application::routes()` to load the HTTP routing so URL generation works.
4. Use `Application::console()`` to load the console routing.
5. Use the created CommandCollection to dispatch the CLI command.

A sketch of the new CLI entrypoint is:

[...]

PR #10753, opened 2017-06-10
3.next - Start adding CommandCollection

markstory:
This is  part of #10716, and adds a very basic `CommandCollection` without the auto-discovery features. The class is very basic right now, but it is a start :smile:

### DIFF
--- a/src/Console/CommandCollection.php
+++ b/src/Console/CommandCollection.php
@@ -1,14 +1,45 @@
 <?php
+/**
+ * CakePHP(tm) : Rapid Development Framework (http://cakephp.org)
+ * Copyright (c) Cake Software Foundation, Inc. (http://cakefoundation.org)
+ *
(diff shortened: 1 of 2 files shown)


src/Console/CommandCollection.php
```diff
@@ -0,0 +1,132 @@
...
+     *
+     * @param string $name The name of the command you want to map.
+     * @param string|\Cake\Console\Shell $command The command to map.
+     * @return $this
+     */
+    public function add($name, $command)
+    {
+        // Once we have a new Command class this should check
+        // against that interface.
+        if (!is_subclass_of($command, 'Cake\Console\Shell')) {
```
dereuromark: Should we also use Shell::class here?

markstory: Good idea.

lorenzo:
This looks good to me

PR #10755, opened 2017-06-10
2.x Use HTTPs URLs

ravage84:
Inspired by the changes and the discussion at https://github.com/cakephp/cakephp/pull/10735/files#diff-cfb68b778e2105931c37743c65812a51R3

I thought I would replace all(*) HTTP URLs with HTTPS ones for 2.x, where available.

(*) left _some_ test related URLs unchanged.

Changes for 3.x in #10756


markstory:
Thanks for doing this in both branches @ravage84

PR #10757, opened 2017-06-11
Make init() chainable and improve docs.

markstory:
I bumped into this during the workshops and found it annoying that `init()` couldn't be chained as I wanted to do `helper('Progress')->init(...)->output(...)` and couldn't.

### DIFF
--- a/src/Shell/Helper/ProgressHelper.php
+++ b/src/Shell/Helper/ProgressHelper.php
@@ -19,6 +19,17 @@ use RuntimeException;
 
 /**
  * Create a progress bar using a supplied callback.
+ *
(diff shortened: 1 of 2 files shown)


jeremyharris:
Should the other methods be chainable too? So you can do:

```
$progress = helper('Progress')
    ->init()
    ->output();

foreach (...) {
    $progress
        ->increment(10)
        ->draw();
}
```

lorenzo:
yes! This has been annoying for me as well.

markstory:
@jeremyharris Yes, increment() and draw() could both be chainable too.

PR #10758, opened 2017-06-11
Ability to see SQL generated statement with parameters substituted 3.next

brettl:
When running the ->sql() function on an ORM call allow the ability to see the sql statement with the parameters in the place holders.

1. Currently when calling sql  on the ORM you will see example "where id = :c1"
2. allow the ability to call an example ->sql(1) or ->sql()->substituted(); and the resulting sql will now be "where id = 1" 

this would allow the ability to take your sql being generated and run in a sql viewer /editor and run your command and possibly see if you are producing the incorrect data or if you are generating the incorrectsql statement


markstory:
Hy @brettl I think you're using the wrong head branch. Right now you're trying to merge `3.next` into `master` on the CakePHP repo. Did you mean to try and merge a branch from your fork into cakephp:master?

Issue #10752, opened 2017-06-10
Improve error messages


ravage84:
Let us talk about improving the error messages as @lorenzo recommended in his CakeFest 2017 keynote.

@lorenzo could you share your slides about improving the error messages here, so we can base the discussion on that? Thank you for the key note & idea!

burzum:
I'm all in for this and refactored a lot of messages in the past. I personally hate nothing more as pointless error messages because they aren't any help for my work. I don't know if it was that worse but here is an example of a pretty useless message: `Could not load association` it's pretty easy to improve that: `Could not load hasMany Association Foo in Some\Table\Class. Did you forget to declare it?`.

I remember that I've at least improved three exception messages that had very limited information. One was related to assocs, another to failing loading fixtures and another on the console AFAIR.

A good error message / exception message contains:

* What caused it?
* Where did it happen?
* What are the involved actors / objects
* If possible to answer: Why did it happen, give some hints

ericadeefox:
Bouncing off of what @burzum said, error messages that check for common syntax/convention errors would be really helpful too, e.g. when `usersController` isn't found, it checks to see if that first character is a capitalized letter--or checks for "sC".

ravage84:
> If possible to answer: Why did it happen, give some hints

I like to insert a) the actual values and b) the range of valid values, value limit or expected list of values.

From Viraj's talk at CakeFest 2017 about testing:

Instead of writing "Status code not between 200 and 204", it should read "Returned status code of 500 is not between 200 and 204".

Though in this particular case it wouldn't hint at the fact that the controller wasn't there, yet.

burzum:
> when usersController isn't found, it checks to see if that first character is a capitalized letter--or checks for "sC".

I think this goes a little to far. 🙂 Also it would pass on windows, it's not case sensitive.

johmc:
(y)

ADmad:
@johmc Unless you have something substantial to add to the discussion you can just use the "Add your reaction" feature, instead of posting a new comment feature :slightly_smiling_face:

johmc:
Thanks! Just testing out hithub :) Will keep that in mind. :)

PR #10760, opened 2017-06-11
Improve MissingController error page. 

markstory:
Don't give people false directions when controllers are missing. Suggest the proper class and filename instead of the lowercased version.

I was inspired to fix this after @lorenzo's talk yesterday.


markstory:
Related to #10752

PR #10762, opened 2017-06-12
Make test method's request

mehdi-fathi:
Make test method's request and fix some of problems in Request.php


src/Http/Client/Request.php
```diff
@@ -253,5 +253,5 @@ public function body($body = null)
     }
 }
 
-// @deprecated Add backwards compat alias.
+// @deprecated Add backwards compact alias.
```
robertpustulka: I think it's short for backwards compat*ibility*

tests/TestCase/Http/Client/RequestTest.php
```diff
@@ -292,14 +341,14 @@ public function testCookie()
     public function testVersion()
     {
         $request = new Request();
-        $result = $request->version('1.0');
+        $request->version('1.0');
         $this->assertSame($request, $request, 'Should return self');
```
robertpustulka: This should be fixed to assert against `$result`

src/Http/Client/Request.php
markstory: It is.

src/Http/Client/Request.php
```diff
@@ -9,7 +9,7 @@
  * @copyright     Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
  * @link          https://cakephp.org CakePHP(tm) Project
  * @since         3.0.0
- * @license       https://opensource.org/licenses/mit-license.php MIT License
+ * @license       https://www.opensource.org/licenses/mit-license.php MIT License
```
markstory: You don't need the www on the https URLs.

tests/TestCase/Http/Client/RequestTest.php
```diff
@@ -37,7 +48,44 @@ public function testConstructorStringData()
...
+        $headers = [
+            'Content-Type' => 'application/json',
+            'Authorization' => 'Bearer valid-token',
+        ];
+        $data = ['a' => 'b', 'c' => 'd'];
+        $request = new Request('http://example.com', 'POST', $headers, json_encode($data));
+        $this->assertContains($request->getMethod(), $this->methods);
+        $this->assertEquals('http://example.com', $request->url());
+        $this->assertEquals('application/json', $request->getHeaderLine('Content-Type'));
+        $this->assertEquals(json_encode($data), $request->body());
```
markstory: This looks like a test that could benefit from a [dataprovider](https://phpunit.de/manual/current/en/writing-tests-for-phpunit.html#writing-tests-for-phpunit.data-providers).

markstory (changes requested):

tests/TestCase/Http/Client/RequestTest.php
mehdi-fathi: Yes exacly

tests/TestCase/Http/Client/RequestTest.php
```diff
@@ -37,10 +48,44 @@ public function testConstructorStringData()
...
+     * @param array|string|null $data The request body to use.
+     * @param string $method The HTTP method to use.
+     *
+     * @dataProvider additionProvider
+     */
+    public function testAdd(array $headers, $data, $method)
+    {
+        $request = new Request('http://example.com', $method, $headers, json_encode($data));
+
+        $this->assertContains($request->getMethod(), $this->methods);
```
markstory: Wouldn't it be more accurate to check that the method matches exactly? Right now the test would continue to pass even if the request method was transposed to another valid method that wasn't requested.

mehdi-fathi: It make no difference.but you are right assertcontains isn't suitable here.

src/Http/Client/Request.php
ravage84: I would recommend to write out the word compatibility in the future.

Issue #9558, opened 2016-10-03
TranslateBehavior - alter join type or make it configurable


bancer:
This is a (multiple allowed):
- [ ] bug
- [x] enhancement
- [x] feature-discussion (RFC)
- CakePHP Version: 2.8.9.
- Platform and Target: Linux, Apache, MySQL.
### What you did

When TranslateBehavior is attached to a model then INNER joins are used. This leads to empty results when the model record exists but the related translation is absent. 
### Expected Behavior

When there is no record in i18n table that is related to the current model record then the result from find() should contain the model fields without translatable column. So either INNER join should be changed to LEFT join or the join type should be configurable.
### Actual Behavior

The result is empty when there is no record in i18n table but there is a record in my_models table.

dereuromark:
I am not sure why the INNER is the default. I agree, that in certain cases the LEFT appears to make more sense.
Let's wait on feedback from others - and maybe someone finds the time to draft a PR with the suggested improvements.

markstory:
Closing as a pull request is open now.

PR #10764, opened 2017-06-12
left join support in TranslateBehavior

bancer:
Resolves https://github.com/cakephp/cakephp/issues/9558

### DIFF
--- a/lib/Cake/Model/Behavior/TranslateBehavior.php
+++ b/lib/Cake/Model/Behavior/TranslateBehavior.php
@@ -60,6 +60,25 @@ class TranslateBehavior extends ModelBehavior {
  * $config could be empty - and translations configured dynamically by
  * bindTranslation() method
  *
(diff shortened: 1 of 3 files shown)


lib/Cake/Model/Behavior/TranslateBehavior.php
```diff
@@ -74,9 +93,16 @@ public function setup(Model $Model, $config = array()) {
 			return false;
 		}
 
-		$this->settings[$Model->alias] = array();
+		$this->settings[$Model->alias] = array(
+			'fields' => array(),
+			'join_type' => 'INNER',
```
markstory: Wouldn't this be 'joinType' to follow the naming conventions of other array keys?

markstory: Changing the structure of public properties will break compatibility for someone. Is there another way to store this that doesn't require potentially breaking user land code?

Issue #10763, opened 2017-06-12
CakeEmail Comma in From address


vanbe:
* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 2.9.4
* Platform and Target: CakeEmail -> SMTP Transport

### What you did
I'm trying to send and email with a comma in the from_name.

### What happened
The sent email is malformed. The MIME enveloppe is not well structured because the comma is not escaped.

Received email header excerpt:
```
From: "Website, =?UTF-8?B?c3DDqWNpYWxpc3RlIFZPTyI=?= <test@domain.com>
To: test@domain2.com
X-Mailer: CakePHP Email

```
### What you expected to happen
The header should have an extra " :

`From: "Website, =?UTF-8?B?c3DDqWNpYWxpc3RlIFZPTw==?=" <test@domain.com>
`
Otherwise some email readers / gmail, etc are considering that there are multiple From emails due to the non-escaped character.

markstory:
Thanks for reporting this. I'll see if I can reproduce it.

markstory:
Pull request is up now.

PR #10766, opened 2017-06-13
Fix encoding of addreses contain comma & unicode

markstory:
Email addresses that contain both unicode and commas will not be correctly encoded by mime_encode_header if the comma precedes the unicode. In this scenario we have to quote the encoded address.

Refs #10763

### DIFF
--- a/lib/Cake/Test/Case/Network/Email/CakeEmailTest.php
+++ b/lib/Cake/Test/Case/Network/Email/CakeEmailTest.php
@@ -450,10 +450,6 @@ class CakeEmailTest extends CakeTestCase {
 		$expected = array('CakePHP <cake@cakephp.org>', 'Cake <php@cakephp.org>');
 		$this->assertSame($expected, $result);
 
-		$result = $this->CakeEmail->formatAddress(array('me@example.com' => 'Last, First'));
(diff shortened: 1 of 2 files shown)


lib/Cake/Test/Case/Network/Email/CakeEmailTest.php
```diff
@@ -472,6 +468,33 @@ public function testFormatAddress() {
 	}
 
 /**
+ * Test that addresses are quoted correctly when they contain unicode and
+ * commas
+ *
+ * @return void
+ */
+	public function testFormatAddressEncodeAndEscape()
+	{
```
ravage84: @markstory it's CakePHP 2.x, not 3.x :trollface:  refs build error https://travis-ci.org/cakephp/cakephp/jobs/242248221#L246

ravage84 (changes requested):
Looks good apart from the CS violation.

lib/Cake/Test/Case/Network/Email/CakeEmailTest.php
markstory: Thanks for pointing that out.

PR #10768, opened 2017-06-13
Remove extract() from ServerRequestFactory::getBase()

slywalker:
By using `extract()`, I propose a modification because it causes poor readability as well as side effects (overwriting `$uri`, `$server`).


ADmad (approved):

markstory:
Thanks @slywalker :clap:

PR #10769, opened 2017-06-13
Add options and head methods to IntegrationTestCase

Iandenh:
As of #10700 there is method to simply add an OPTIONS route. I thought is was handy to make an options method in IntegrationTestCase.

### DIFF
--- a/src/TestSuite/IntegrationTestCase.php
+++ b/src/TestSuite/IntegrationTestCase.php
@@ -450,7 +450,7 @@ abstract class IntegrationTestCase extends TestCase
     }
 
     /**
-     * Performs an OPTION request using the current request data.
+     * Performs an OPTIONS request using the current request data.
      *
      * The response of the dispatched request will be stored as
      * a property. You can use various assert methods to check the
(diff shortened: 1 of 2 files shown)


markstory:
Any chance we could get a test to cover these methods existing?

Iandenh:
I can add a test like:
```php
     /**
     * Test sending head requests.
     *
     * @return void
     */
    public function testHead()
    {
        $this->assertNull($this->_response);

        $this->head('/request_action/test_request_action');
        $this->assertNotEmpty($this->_response);
        $this->assertInstanceOf('Cake\Http\Response', $this->_response);
        $this->assertEquals('This is a test', $this->_response->body()); //this line is agains the HTTP HEAD specification and should be changed to assertEmpty/assertNull

        $this->_response = null;
        $this->head('/head/request_action/test_request_action');
        $this->assertResponseSuccess();
    }
```

But the problem is that a HEAD and OPTIONS request MUST NOT contain a body response. 
https://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html#sec9.4
Right now the same body response as with a GET request is send. Is that a bug in Cakephp or should that be solved by the user?

markstory:
> Right now the same body response as with a GET request is send. Is that a bug in Cakephp or should that be solved by the user?

I believe that should be up to the application developer.

markstory:
Thanks!

PR #10770, opened 2017-06-13
Fixes warning für "disabled" option in FormHelper.php

alexdd55:
removes warning when key "disabled" is delivered, but array is empty

having the key disabled and filling it, is in my opinion a common procedure, so it is possible, that the the array is set, but empty.

### DIFF
--- a/src/View/Helper/FormHelper.php
+++ b/src/View/Helper/FormHelper.php
@@ -2578,7 +2578,7 @@ class FormHelper extends Helper
             $options = $this->addClass($options, $this->_config['errorClass']);
         }
         $isDisabled = false;
-        if (isset($options['disabled'])) {
+        if (isset($options['disabled']) && !empty($options['disabled'])) {
             $isDisabled = (
                 $options['disabled'] === true ||
                 $options['disabled'] === 'disabled' ||


markstory:
How does one trigger the error you were having? I'd like to use that scenario to create a regression test.

alexdd55:
i have basically this code as a result of some data manipulation after fetching data from DB.
i created 
`
$disabled = []
if($foo == $bar) {
// add data (eg. 1) to array $disabled otherwise the array $disabled stays empty
}

$data = [
	[
		'value' => 1,
		'text' => 'radio1'
	],
	[
		'value' => 2,
		'text' => 'radio2'
	],
	[
		'value' => 3,
		'text' => 'radio3'
	]
];
$attr => [
	'escape' => false,
	'label' => 'radiolabel',
	'disabled' => $disabled
];

echo $this->Form->radio('foo', $data, $attr);`

markstory:
Ok and what is in `$disabled`?

alexdd55:
It’s either an empty array, or ids of the radio buttons that are disabled, which worked fine, besides the warning that appeared.

When I look at the docs, the similar behaviour seems to be achieved with select fields (never tried that).

alexdd55:
let me add what i wanted to do.. hopefully this is helpful 
i want do show a number radiobuttons, and disable the ones that are not available anymore.
works smoothly, besides the warning message.

markstory:
I'm not sure this patch will fix the scenarios where `$attrs['disabled']` is not empty. I'm still getting warnings locally when disabled is not empty.

markstory:
Closing as I've built upon these changes and covered more cases.

Issue #10765, opened 2017-06-12
Cannot insert zero using Fixture for testing


dangerdan:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.4.7
* Platform and Target: nginx, mariadb

### What you did
I'm using a fixture to insert values into a status table that has an id and label column. This is failing due to the primary index on id, even though the values are all different.

It appears to be the '0' for pending that is causing the issue:

    public $fields = [
        'id' => ['type' => 'integer', 'autoIncrement' => false, 'length' => 11, 'unsigned' => false,'signed' => true, 'null' => false, 'default' => null, 'comment' => '', 'precision' => 10],
        'label' => ['type' => 'string', 'length' => 15, 'null' => true, 'default' => '', 'comment' => '', 'precision' => null, 'fixed' => null],
        '_constraints' => [
           'primary' => ['type' => 'primary', 'columns' => ['id'], 'length' => []],
        ]
    ];
    
    public $records = [
        [
            'id' => -1,
            'label' => 'Cancelled'
        ],
        [
            'id' => 0,
            'label' => 'Pending'
        ],
        [
            'id' => 1,
            'label' => 'Confirmed'
        ]
    ];

### What happened

Exception: Unable to insert fixture "App\Test\Fixture\StatusesFixture" in "App\Test\TestCase\Controller\BusinessControllerTest" test case: 
SQLSTATE[23000]: Integrity constraint violation: 1062 Duplicate entry '1' for key 'PRIMARY' in [/application/vendor/cakephp/cakephp/src/TestSuite/Fixture/FixtureManager.php, line 337]

### What you expected to happen

The test table should be created with a '0' as the id for the Pending status.

jeremyharris:
According to the error, it's the third fixture that's the problem. You use 1 again as an id.

markstory:
I will see if I can reproduce this on my side.

markstory:
Looks like this is caused by Cake generating the incorrect schema based on the fixture data. The schema I'm getting locally is:

```sql
CREATE TABLE `statuses` (
`id` INTEGER(11) NOT NULL AUTO_INCREMENT,
`label` VARCHAR(15) DEFAULT '',
PRIMARY KEY (`id`)
)
```

which is not correct as `id` should not be `autoincrement`. This sounds very similar to #10045 which is also about fixtures forcing auto increment on tables in MySQL.

Issue #10045, opened 2017-01-16
Fixtures force autoincrement on ID column


thinkingmedia:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.3.11
* Platform and Target: Apache2, MySQL, PHP 7.1, Bash on Windows 10

### What you did

Created a fixture that imports the schema.

### What happened

Here is the production table named `ratings`. Notice that by design it does not have an auto-increment on the primary `ID` column. This table stores film ratings and I use the numeric age as the primary key.

```
CREATE TABLE `ratings` (
  `id` int(11) NOT NULL,
  `key` char(5) COLLATE utf8mb4_unicode_ci NOT NULL,
  `title` varchar(80) COLLATE utf8mb4_unicode_ci NOT NULL,
  `created` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP,
  `updated` datetime NOT NULL,
  PRIMARY KEY (`id`),
  UNIQUE KEY `id` (`id`),
  KEY `created` (`created`),
  KEY `updated` (`updated`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_unicode_ci
```

When I created a fixture for `ratings` and used the schema import feature. The fixture manager created a table with an auto-increment `ID`. This creates problems for my code.

Here's the fixture table that was created.

```
CREATE TABLE `ratings` (
  `id` int(11) NOT NULL AUTO_INCREMENT,
  `key` char(5) COLLATE utf8mb4_unicode_ci NOT NULL,
  `title` varchar(80) COLLATE utf8mb4_unicode_ci NOT NULL,
  `created` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP,
  `updated` datetime NOT NULL,
  PRIMARY KEY (`id`),
  UNIQUE KEY `id` (`id`),
  KEY `created` (`created`),
  KEY `updated` (`updated`)
) ENGINE=InnoDB AUTO_INCREMENT=20 DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_unicode_ci
```

Ignore the `AUTO_INCREMENT=20` as I dumped the create statement as the tests ran.

### What you expected to happen

Should not create an auto-increment column unless the source table has that attribute.

markstory:
[...]

PR #10772, opened 2017-06-14
Don't add auto increment to columns that turn it off.

markstory:
When autoIncrement has been set to a specific value we shouldn't overwrite the user's request. This allows integer keys to exist without autoincrement being set.

Refs #10765
Refs #10045


dereuromark:
That's gonna make a few people very happy :)

PR #10777, opened 2017-06-15
Add disableErrorHandlerMiddleware method

### LINKED ISSUE #10157: IntegrationTestCase::_sendRequest doesn't handle errors anymore

cleptric:
Fixes #10157

### DIFF
--- a/src/TestSuite/IntegrationTestCase.php
+++ b/src/TestSuite/IntegrationTestCase.php
@@ -1053,8 +1053,9 @@ abstract class IntegrationTestCase extends TestCase
     /**
      * Disable the error handler middleware.
      *
-     * By using this function, exceptions are no longer caught by ErrorHandlerMiddleware
-     * and are instead re-thrown by the TestExceptionRenderer. This results in a more clearer error message.
+     * By using this function, exceptions are no longer caught by the ErrorHandlerMiddleware
+     * and are instead re-thrown by the TestExceptionRenderer. This can be helpful
+     * when trying to diagnose/debug unexpected failures in test cases.
      *
      * @return void
      */
--- a/src/TestSuite/IntegrationTestCase.php
+++ b/src/TestSuite/IntegrationTestCase.php
@@ -1053,8 +1053,8 @@ abstract class IntegrationTestCase extends TestCase
     /**
      * Disable the error handler middleware.
      *
-     * By using this function, the exception gets no longer catched by the ErrorHandlerMiddleware
-     * and is instead just rethrown by the TestExceptionRenderer. This results in a more clearer error message.
+     * By using this function, exceptions are no longer caught by ErrorHandlerMiddleware
+     * and are instead re-thrown by the TestExceptionRenderer. This results in a more clearer error message.
      *
      * @return void
      */
--- a/src/TestSuite/IntegrationTestCase.php
+++ b/src/TestSuite/IntegrationTestCase.php
@@ -24,6 +24,7 @@ use Cake\Core\Configure;
 use Cake\Database\Exception as DatabaseException;
 use Cake\Network\Session;
 use Cake\Routing\Router;
+use Cake\TestSuite\Stub\TestExceptionRenderer;
 use Cake\Utility\CookieCryptTrait;
 use Cake\Utility\Hash;
 use Cake\Utility\Security;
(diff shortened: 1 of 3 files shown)


viraj-khatavkar:
@cleptric I appreciate your efforts for the PR. Throwing a stub at a global level wouldn't take care of this problem. Following is one use-case which we need to tackle before throwing this directly:

**Use-case**

I want to assert that a 404 page is thrown and the content on the page is *Page not found* If we throw the exception directly, the framework wouldn't catch the error and show the 404 page which I want to assert. 

Throwing exception works only if we want to test against positive results, and debugging purposes. I don't intend to stub the exception handler for all my tests, but only where I need help in debugging. In the case of asserting against an exception or a scenario like 404-page assertion, the test would break even though the assertion might have been correct in the case the error handler worked properly.

src/TestSuite/IntegrationTestCase.php
```diff
@@ -1050,6 +1051,19 @@ public function assertCookieNotSet($cookie, $message = '')
     }
 
     /**
+     * Disable the error handler middleware.
+     *
+     * By using this function, the exception gets no longer catched by the ErrorHandlerMiddleware
```
markstory: 'exceptions are no longer caught by ErrorHandlerMiddleware'

src/TestSuite/IntegrationTestCase.php
```diff
@@ -1050,6 +1051,19 @@ public function assertCookieNotSet($cookie, $message = '')
...
+     * and is instead just rethrown by the TestExceptionRenderer. This results in a more clearer error message.
```
markstory: 'and are instead re-thrown'

tests/TestCase/TestSuite/IntegrationTestCaseTest.php
```diff
@@ -921,4 +921,18 @@ public function testAssertFileNoFile()
...
+     *
+     * @expectedException \Cake\Routing\Exception\MissingRouteException
+     * @expectedExceptionMessage A route matching "/foo" could not be found.
+     * @return void
+     */
+    public function testDisableErrorHandlerMiddleware()
+    {
+        $this->disableErrorHandlerMiddleware();
+        $this->get('/foo');
+        $this->assertResponseOk();
```
markstory: This line won't be reached will it?

cleptric: True 😄

cleptric:
@viraj-khatavkar Maybe our use case differs a little bit:

I have an 3.3 app, where I wrote some integration tests which triggers exceptions.
The exceptions are just printed out to the cli, add a lot of noise to the test logs and let the tests pass in the end.

Now, I'm able to add `$this->disableErrorHandlerMiddleware()` explicitly to this test cases and add proper `@expectedException` annotations to this tests.

In my opinion, the new method shouldn't be used in a global context and instead, just used for debugging or granular in some tests that need this behaviour.

src/TestSuite/Stub/TestExceptionRenderer.php
```diff
@@ -0,0 +1,36 @@
...
+ * @since         3.5.0
+ * @license       https://opensource.org/licenses/mit-license.php MIT License
+ */
+namespace Cake\TestSuite\Stub;
+
+use Exception;
+
+/**
+ * Test Exception Renderer.
+ */
```
ravage84: It might make sense to explain a little why this exists. Something like https://github.com/cakephp/cakephp/pull/10777/files#diff-30ad814a76cb9a0911ed10f377ace431R1056

src/TestSuite/Stub/TestExceptionRenderer.php
```diff
@@ -0,0 +1,36 @@
...
+class TestExceptionRenderer
+{
+
+    /**
+     * Constructor
```
ravage84: "Simply rethrows the given exception"

tests/TestCase/TestSuite/IntegrationTestCaseTest.php
```diff
@@ -921,4 +921,17 @@ public function testAssertFileNoFile()
         $this->get('/posts/get');
         $this->assertFileResponse('foo');
     }
+
+    /**
+     * undocumented function
```
ravage84: ?

ravage84 (commented):
Looks good!

tests/TestCase/TestSuite/IntegrationTestCaseTest.php
cleptric: Ops 😅

Issue #10778, opened 2017-06-15
RFC: Router scope cache


steinkel:
This is a (multiple allowed):

* [] bug
* [x] enhancement
* [x] feature-discussion (RFC)

* CakePHP Version: 3.next
* Platform and Target: 3.next

In projects where several routes are connected, sometimes with routes coming from the database or external files, there is a possible optimization caching the generated RouteCollection.

### Interface

The interface for this cache could be located in the scope params and define the scoped route like

    Router::scope('/', ['_cache' => 'default'], function (RouteBuilder $routes) {...});

The `_cache` param defines the cache configuration to use, and the cache key will be the `path` of the scope.

### Test config and results

This is the configuration I've used for testing performance locally (using default File based cache)

```
$start = microtime(true);
Router::scope('/', ['_cache' => 'default'], function (RouteBuilder $routes) {
    $routes->connect('/', ['controller' => 'Pages', 'action' => 'display', 'home']);
    $routes->connect('/pages/*', ['controller' => 'Pages', 'action' => 'display']);
    $range = range(1, 4000);
    foreach ($range as $i) {
        $routes->connect('/pages/' . $i, ['controller' => 'Pages', 'action' => 'display']);
    }
    $routes->fallbacks(DashedRoute::class);
});
$stop = microtime(true);
Cake\Log\Log::debug($stop - $start);
```

**Times with/without cache were: 
0.00463 vs 0.00276 for 50 connected routes
1.45s vs 0.015 for 4000 connected routes**

### Implementation notes:

* We could think of another implementation Router::cache() to define global caching of the RouteCollection (it's static anyway) but I think it's better to let the user define which scopes should be cached.
* If 2 or more scopes are cached, the per scope cache will store previous cached router collection twice as the RouteCollection object is static, this could be a possible improvement in a future PR if this one is looking good.
* This feature is BC with existing 3.x versions

### Todo:

* Add new option to documentation
* Add note about using `bin/cake routes` to build the routes cache as a deploy step

davidyell:
Sounds like a good idea to me. Would this cache be clearable using a shell? Would it respect the configured `Cache.diabled` configuration flag?

markstory:
Closing as there is a pull request open now.

steinkel:
Here's another round on the route-caching topic. I've been testing the middleware based routing cache and this would remove some of the issues with the scope based caching, getting all the routeCollection serialized and reused between requests. The new CacheRoutingMiddleware class would look like this:

```
[...]

PR #10779, opened 2017-06-15
add router scope cache

steinkel:
Related to #10778 it's a first implementation for your review & comments. I've tested it locally and found interesting improvements in performance and would be interested in routing use cases that could break the caching feature.

Thanks,

### DIFF
--- a/src/Routing/Router.php
+++ b/src/Routing/Router.php
@@ -1007,15 +1007,10 @@ class Router
             $options = $params;
             $cache = (string)Hash::get($params, '_cache');
             unset($params['routeClass'], $params['extensions'], $params['_cache']);
-            if (!empty($cache) && static::_isCacheSerialize($cache)) {
+            if ($cache) {
                 static::_rememberScope($path, $params, $options, $callback, $cache);
 
                 return;
-            } elseif (!empty($cache) && !static::_isCacheSerialize($cache)) {
-                $msg = 'Router scope "%s" was NOT cached. The scope cache is enabled, but "%s" cache engine' .
-                    'does not allow serialization. ' .
-                    'Use a valid engine, and ensure "serialize" is enabled in the cache configuration';
-                Log::warning(sprintf($msg, $path, $cache));
             }
         }
 
@@ -1037,15 +1032,42 @@ class Router
      */
     protected static function _rememberScope($path, $params, $options, $callback, $cache)
     {
-        static::$_collection = Cache::remember(
-            static::$_cachePrefix . $path,
-            function () use ($path, $params, $options, $callback) {
-                static::_buildScope($path, $params, $options, $callback);
+        if (static::_checkSerializableCache($path, $cache)) {
+            static::$_collection = Cache::remember(
+                static::$_cachePrefix . $path,
+                function () use ($path, $params, $options, $callback) {
+                    static::_buildScope($path, $params, $options, $callback);
 
-                return static::$_collection;
-            },
-            $cache
-        );
+                    return static::$_collection;
+                },
+                $cache
+            );
+
+            return;
+        }
+
+        static::_buildScope($path, $params, $options, $callback);
+    }
+
+    /**
+     * Check if the cache config is serializable, log an error message if it's not
+     *
+     * @param string $path The path prefix for the scope. This path will be prepended
+     *   to all routes connected in the scoped collection.
+     * @param string $cache The Cache configuration name to use, a $_cachePrefix will be added to the key
+     * @return bool
+     */
+    protected static function _checkSerializableCache($path, $cache)
+    {
+        $isCacheSerializable = static::_isCacheSerialize($cache);
+        if (!$isCacheSerializable) {
+            $msg = 'Router scope "%s" was NOT cached. The scope cache is enabled, but "%s" cache engine ' .
+                'does not allow serialization. ' .
+                'Use a valid engine, and ensure "serialize" is enabled in the cache configuration';
+            Log::warning(sprintf($msg, $path, $cache));
+        }
+
+        return $isCacheSerializable;
     }
 
     /**
--- a/src/Routing/Router.php
+++ b/src/Routing/Router.php
@@ -164,6 +164,13 @@ class Router
     protected static $_defaultExtensions = [];
 
     /**
+     * Default prefix used for scope cache keys
+     *
+     * @var string
+     */
+    protected static $_cachePrefix = 'routes';
+
+    /**
      * Get or set default route class.
      *
      * @param string|null $routeClass Class name.
@@ -986,7 +993,7 @@ class Router
             if (!empty($cache)) {
                 unset($params['_cache']);
                 static::$_collection = Cache::remember(
-                    $path,
+                    static::$_cachePrefix . $path,
                     function () use ($path, $params, $options, $callback) {
                         self::_buildScope($path, $params, $options, $callback);
 
--- a/src/Routing/Router.php
+++ b/src/Routing/Router.php
@@ -12,11 +12,14 @@
  * @since         0.2.9
  * @license       https://opensource.org/licenses/mit-license.php MIT License
  */
+
 namespace Cake\Routing;
 
+use Cake\Cache\Cache;
 use Cake\Core\Configure;
 use Cake\Http\MiddlewareQueue;
 use Cake\Http\ServerRequest;
+use Cake\Utility\Hash;
 use Cake\Utility\Inflector;
 use InvalidArgumentException;
 use Psr\Http\Message\ServerRequestInterface;
@@ -979,7 +982,39 @@ class Router
         if (is_array($params)) {
             $options = $params;
             unset($params['routeClass'], $params['extensions']);
+            $cache = (string)Hash::get($params, '_cache');
+            if (!empty($cache)) {
+                unset($params['_cache']);
+                static::$_collection = Cache::remember(
+                    $path,
+                    function () use ($path, $params, $options, $callback) {
+                        self::_buildScope($path, $params, $options, $callback);
+
+                        return static::$_collection;
+                    },
+                    $cache
+                );
+
+                return;
+            }
         }
+
+        static::_buildScope($path, $params, $options, $callback);
+    }
+
+    /**
+     * @param string $path The path prefix for the scope. This path will be prepended
+     *   to all routes connected in the scoped collection.
+     * @param array|callable $params An array of routing defaults to add to each connected route.
+     *   If you have no parameters, this argument can be a callable.
+     * @param array $options An array of routing defaults to add to each connected route after
+     *   routeClass, extensions and _cache are unset
+     * @param callable|null $callback The callback to invoke with the scoped collection.
+     * @throws \InvalidArgumentException When an invalid callable is provided.
+     * @return void
+     */
+    protected static function _buildScope($path, $params = [], $options = [], $callback = null)
+    {
         $builder = static::createRouteBuilder('/', $options);
         $builder->scope($path, $params, $callback);
     }
--- a/src/Routing/Router.php
+++ b/src/Routing/Router.php
@@ -1059,7 +1059,7 @@ class Router
      */
     protected static function _checkSerializableCache($path, $cache)
     {
-        $isCacheSerializable = static::_isCacheSerialize($cache);
+        $isCacheSerializable = static::_isCacheSerializable($cache);
         if (!$isCacheSerializable) {
             $msg = 'Router scope "%s" was NOT cached. The scope cache is enabled, but "%s" cache engine ' .
                 'does not allow serialization. ' .
@@ -1095,7 +1095,7 @@ class Router
      * @param string $cache Cache configuration
      * @return bool
      */
-    protected static function _isCacheSerialize($cache)
+    protected static function _isCacheSerializable($cache)
     {
         $config = Cache::engine($cache)->getConfig();
 
--- a/src/Routing/Router.php
+++ b/src/Routing/Router.php
@@ -1012,10 +1012,10 @@ class Router
 
                 return;
             } elseif (!empty($cache) && !static::_isCacheSerialize($cache)) {
-                $msg = 'Scoped routes were NOT cached. The scope cache is enabled, but "%s" cache engine' .
+                $msg = 'Router scope "%s" was NOT cached. The scope cache is enabled, but "%s" cache engine' .
                     'does not allow serialization. ' .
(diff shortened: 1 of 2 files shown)


src/Routing/Router.php
```diff
@@ -979,7 +982,39 @@ public static function scope($path, $params = [], $callback = null)
         if (is_array($params)) {
             $options = $params;
             unset($params['routeClass'], $params['extensions']);
+            $cache = (string)Hash::get($params, '_cache');
+            if (!empty($cache)) {
+                unset($params['_cache']);
+                static::$_collection = Cache::remember(
+                    $path,
```
markstory: Should these keys be prefixed with `routes` to avoid potential conflicts with other cached data?

src/Routing/Router.php
```diff
@@ -979,7 +982,39 @@ public static function scope($path, $params = [], $callback = null)
...
+                    function () use ($path, $params, $options, $callback) {
+                        self::_buildScope($path, $params, $options, $callback);
+
+                        return static::$_collection;
```
markstory: How would this work if the cache configuration had `serialize => false`?

src/Routing/Router.php
steinkel: yes, that's a good idea

src/Routing/Router.php
steinkel: I'll check if serialize is disable and handle serialization here before saving into cache. Thanks for raising the flag.

src/Routing/Router.php
steinkel: fixed

src/Routing/Router.php
steinkel: I'm now checking for serializable, if not, I'm logging a warning message. I don't think we should handle the serialization in the Router as I was about to do. Throwing an exception is an option too, but maybe too hard and could break apps when going live badly if the cache is not correctly configured...

src/Routing/Router.php
```diff
@@ -951,6 +962,14 @@ public static function createRouteBuilder($path, array $options = [])
      * - `extensions` The extensions to enable in this scope. Defaults to the globally
      *   enabled extensions set with `Router::extensions()`
      *
+     * ### Cache
+     *
+     * You can enable cache to reduce the time spent building the scoped routes.
```
ravage84: Either "the cache" or just "caching".

ravage84: "time needed to build"?

src/Routing/Router.php
```diff
@@ -951,6 +962,14 @@ public static function createRouteBuilder($path, array $options = [])
...
+     * The `_cache` param will allow you to define the cache configuration to store the routes.
```
ravage84: "the scoped routes", if I understand correctly.

src/Routing/Router.php
```diff
@@ -978,13 +1005,82 @@ public static function scope($path, $params = [], $callback = null)
...
             $options = $params;
-            unset($params['routeClass'], $params['extensions']);
+            $cache = (string)Hash::get($params, '_cache');
+            unset($params['routeClass'], $params['extensions'], $params['_cache']);
+            if (!empty($cache) && static::_isCacheSerialize($cache)) {
+                static::_rememberScope($path, $params, $options, $callback, $cache);
+
+                return;
+            } elseif (!empty($cache) && !static::_isCacheSerialize($cache)) {
+                $msg = 'Scope Routes were NOT cached. The scope cache is enabled, but "%s" cache engine' .
```
ravage84: Not "Scoped Routes"?

ravage84 (commented):
Looks good. Thanks.

steinkel:
@ravage84 thanks for the review, docs fixed

src/Routing/Router.php
```diff
@@ -12,11 +12,15 @@
  * @since         0.2.9
  * @license       https://opensource.org/licenses/mit-license.php MIT License
  */
+
```
bcrowe: Extra newline

steinkel: fixed, it's weird that the cs check didn't catch this one...

src/Routing/Router.php
```diff
@@ -978,13 +1005,82 @@ public static function scope($path, $params = [], $callback = null)
...
             $options = $params;
-            unset($params['routeClass'], $params['extensions']);
+            $cache = (string)Hash::get($params, '_cache');
+            unset($params['routeClass'], $params['extensions'], $params['_cache']);
+            if (!empty($cache) && static::_isCacheSerialize($cache)) {
+                static::_rememberScope($path, $params, $options, $callback, $cache);
+
+                return;
+            } elseif (!empty($cache) && !static::_isCacheSerialize($cache)) {
+                $msg = 'Scoped routes were NOT cached. The scope cache is enabled, but "%s" cache engine' .
```
lorenzo: You should mention what scope could not be serialized

tests/TestCase/Routing/RouterTest.php
```diff
@@ -3087,6 +3090,103 @@ public function testScope()
...
+            'prefix' => 'myapp_cake_someprefix_' . $tmpName,
+            'path' => CACHE . 'persistent/',
+            'serialize' => false,
+        ]);
+
+        Router::scope('/path', ['param' => 'value', '_cache' => 'routes'], function ($routes) {
+            $this->assertInstanceOf('Cake\Routing\RouteBuilder', $routes);
+            $this->assertEquals('/path', $routes->path());
+            $this->assertEquals(['param' => 'value'], $routes->params());
+            $this->assertEquals('', $routes->namePrefix());
```
dereuromark: Can we use assertSame()? Equals would also return true for 0 or other falsy content here.

src/Routing/Router.php
```diff
@@ -978,13 +1005,82 @@ public static function scope($path, $params = [], $callback = null)
...
         if (is_array($params)) {
             $options = $params;
-            unset($params['routeClass'], $params['extensions']);
+            $cache = (string)Hash::get($params, '_cache');
+            unset($params['routeClass'], $params['extensions'], $params['_cache']);
+            if (!empty($cache) && static::_isCacheSerialize($cache)) {
+                static::_rememberScope($path, $params, $options, $callback, $cache);
+
+                return;
+            } elseif (!empty($cache) && !static::_isCacheSerialize($cache)) {
```
dereuromark: no else required after return early.

src/Routing/Router.php
```diff
@@ -978,13 +1005,82 @@ public static function scope($path, $params = [], $callback = null)
         $options = [];
         if (is_array($params)) {
             $options = $params;
-            unset($params['routeClass'], $params['extensions']);
+            $cache = (string)Hash::get($params, '_cache');
```
lorenzo: It would be more readable and potentially faster to do `$cache = $params['_cache'] ?? null`

src/Routing/Router.php
```diff
@@ -978,13 +1005,82 @@ public static function scope($path, $params = [], $callback = null)
...
+            unset($params['routeClass'], $params['extensions'], $params['_cache']);
+            if (!empty($cache) && static::_isCacheSerialize($cache)) {
```
lorenzo: This would be simpler if the check for the cache is done in `rememberScope`
The end of the function would be this line, having something like `return static::rememberScope(...)`

src/Routing/Router.php
steinkel: I agree, but ?? is not available in PHP 5.6

src/Routing/Router.php
steinkel: fixed

tests/TestCase/Routing/RouterTest.php
steinkel: fixed

src/Routing/Router.php
steinkel: fixed

src/Routing/Router.php
steinkel: code fixed to make it simpler (hopefully) to understand. I can't return at that point because the scope function returns void and I would like to keep it BC

src/Routing/Router.php
```diff
@@ -978,13 +1005,104 @@ public static function scope($path, $params = [], $callback = null)
...
         $builder->scope($path, $params, $callback);
     }
 
     /**
+     * Require the cache engine is serializable
+     *
+     * @param string $cache Cache configuration
+     * @return bool
+     */
+    protected static function _isCacheSerialize($cache)
```
bcrowe: Should this method be named `isCacheSerializable` instead?

src/Routing/Router.php
lorenzo: True, forgot about this

src/Routing/Router.php
```diff
@@ -1037,15 +1032,42 @@ public static function scope($path, $params = [], $callback = null)
...
+     * @param string $path The path prefix for the scope. This path will be prepended
+     *   to all routes connected in the scoped collection.
+     * @param string $cache The Cache configuration name to use, a $_cachePrefix will be added to the key
+     * @return bool
+     */
+    protected static function _checkSerializableCache($path, $cache)
+    {
+        $isCacheSerializable = static::_isCacheSerialize($cache);
+        if (!$isCacheSerializable) {
+            $msg = 'Router scope "%s" was NOT cached. The scope cache is enabled, but "%s" cache engine ' .
```
lorenzo: Great message!

src/Routing/Router.php
steinkel: fixed

steinkel:
hmm the build is failing for php5.6 
`Fatal error: Declaration of Cake\TestSuite\Fixture\FixtureInjector::addFailure() must be compatible with PHPUnit_Framework_TestListener::addFailure(PHPUnit_Framework_Test $test, PHPUnit_Framework_AssertionFailedError $e, $time) in /home/travis/build/cakephp/cakephp/src/TestSuite/Fixture/FixtureInjector.php on line 45`

but looks like related to another issue in FixtureInjector

steinkel:
Answering to @davidyell here
`Would this cache be clearable using a shell?

I guess it'd also respect the configured Cache.diabled configuration flag?`

Yes to both, as we are using internally `Cache::remember()` it will have these features as any other cache configuration. You'll be able to use `bin/cake cache clear routes` to clear the routes cache, and possibly a good idea to call `bin/cake routes` as part of the deployment too to warm up the routes cache.

cleptric:
@steinkel This was fixed in master by @jeremyharris. See https://github.com/cakephp/cakephp/pull/10821

markstory:
Shouldn't these caching features be implemented on `RouteBuilder` instead of Router? I would prefer us to steer people away from the static methods and only offering caching on the static method does the opposite of that.

steinkel:
@markstory this could also improve the caching of duplicated contents, let me play with this idea a bit and provide a patch

steinkel:
They are all operating on the same RouteCollection object, so moving the cache logic to RouteBuilder will have no effect on caching existing cached data (in another scope) unless we would have a RouteCollection per scope. Anyway I'll move the cache logic to RouteBuilder instead of Router.

markstory:
Adding more route collections will be a bigger issue. But the shared collection does present an interesting challenge as it creates an opportunity for the cache option to not match what people are expecting if its used on nested scopes, or multiple times in a routes list. As each scopes' cache file will contain all the routes that preceded it.

steinkel:
Yes, containing already cached routes could be an issue, so I was playing with the concept of isolated RouteCollection objects getting merged after each scope is built so each scope could be cached independently. I'll continue this path a little bit more to see where it goes...

steinkel:
After some more research&play with the concept of the isolated RouteCollection objects per scope, although this is possible (I got 6 unit tests failing and stopped there) here are my findings for discussion:
* Creating a new RouteCollection and passing it to the builder is possible, but requires a `merge` method to allow including the routes after the add logic.
* The add route logic also deals with other properties that are preserved, affecting new routes added, this is hard to maintain across the newly created objects.
* The `merge` method adds time to the route processing, so to ALL applications, so if the are not using cache, they could experience slower route processing, for example: detecting duplicated routes and throwing the expected exception.
* There are specific use cases that could be hard to understand/debug, for example caching a parent scope, and having a nested scope generating dynamic routes.

At this point, I don't feel using isolated RouteCollection would work without refactoring a lot of code, out of the scope here. I'll add a RFC for discussion related to isolated RouteCollection per scopes maybe for 4.x

About caching the whole static RouteCollection, I still see benefit for performance, and **I would ask for some feedback about the best way to proceed between**:
1. Adding constraints to the way the cache can be used, for example, detecting ALL the cached scopes must be declared at the beginning of the route configuration, and cached scopes cannot have nested scopes, preventing the weird use cases. We can detect both conditions and raise exceptions only for cached scopes (BC). In this approach we would be caching duplicated routes if caching multiple scopes, but the optimization in time justifies IMO the extra space used, as the user is already trading space vs time using the cache.

2. Dropping this PR here and moving the feature to a plugin, where we can play with this approach in a safer environment.

What do you think?

markstory:
> Adding constraints to the way the cache can be used, for example, detecting ALL the cached scopes must be declared at the beginning of the route configuration, and cached scopes cannot have nested scopes, preventing the weird use cases. 

Could this be a flag that is set on the top-level builder. We can have some confidence around which builders are top-level and which are not.

Another option could be to move caching to the `RoutingMiddleware`. That would enable us to cache the entire route collection as a single cache entry.

steinkel:
about caching from the `RoutingMiddleware` the problem I see there is we only have access to the Router class, forcing us to either expose the collection and let the middleware manage the cache or add a cache method in the router class. Both options look bad based on our previous discussion...

markstory:
Hrm. could the builder expose a 'cache' method? That method could open a block that is cached. We'd be able to track whether or not it was currently in use getting around the nested cache block issues, and potentially prevent the method from being called multiple times as well.

```php
$routes->cache(function ($routes) {
  // Connect routes here.
});
```

dereuromark:
Is this still interesting?

burzum:
@markstory any news on this topic? :)

dereuromark:
We could already try some performance testing with this version and see the difference here, how relevant a performance boost this will give us.

steinkel:
I'll re-visit this PR in the following days and try to provide a better approach

markstory:
@burzum No news. I think adding cache options to `scope()` is going to contain a number of edge cases as @steinkel has proven with this. I think we should look at adding caching higher up in the router stack so we can cache all the routes for an application. Perhaps the RoutingMiddleware, or somewhere during application bootstrapping could be leveraged.

I'm going to close this as it cannot be merged, and I'm not comfortable with the edge-cases that caching individual scopes creates.

PR #10781, opened 2017-06-15
Fix Correspond to RedisCluster from RedisEngine::clear()

h-moriya:
Redis cluster will fail with Multiple Keys

redis-cli : 
```
> del kye1 kye2
> (error) CROSSSLOT Keys in request don't hash to the same slot
```

Execution of the `RedisEngine::clear()` makes it true, but the key is not deleted

### DIFF
--- a/src/Cache/Engine/RedisEngine.php
+++ b/src/Cache/Engine/RedisEngine.php
@@ -218,7 +218,12 @@ class RedisEngine extends CacheEngine
         }
         $keys = $this->_Redis->getKeys($this->_config['prefix'] . '*');
 
-        return !in_array(false, $this->deleteMany($keys));
+        $result = [];
+        foreach ($keys as $key) {
+            $result[] = $this->_Redis->delete($key) > 0;
+        }
+
+        return !in_array(false, $result);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


markstory (changes requested):
The test suite is failing. Could you fix the failing test?

h-moriya:
Sorry, Fixed the mistake

src/Cache/Engine/RedisEngine.php
```diff
@@ -218,7 +218,12 @@ public function clear($check)
         }
         $keys = $this->_Redis->getKeys($this->_config['prefix'] . '*');
 
-        return !in_array(false, $this->deleteMany($keys));
+        $result = [];
+        foreach ($keys as $key) {
+            $result[] = $this->_Redis->delete($key) > 0;
```
markstory: It looks like `delete()` takes an [array of keys](https://github.com/phpredis/phpredis#del-delete). Why are you calling delete N times?

h-moriya: because `delete()` is to create the Multiple Keys command

e.g.

---

php
```
$keys = ['KEY1', 'KEY2', 'KEY3'];
$this->_Redis->delete($keys)
```
redis-cli
```
del KEY1 KEY2 KEY3
```
---
php
```
$keys = ['KEY1', 'KEY2', 'KEY3'];
foreach ($keys as $key) {
    $this->_Redis->delete($key);
}
```
redis-cli
```
del KEY1
del KEY2
del KEY3
```

markstory: Ok. Which is the source of the CROSSLOT issue you're facing.

markstory (approved):

PR #10783, opened 2017-06-15
Improve readability

ravage84:
Since we are not writing back into the config array, we do not need to work on the $config variable the whole time.


markstory:
Thanks!

PR #10784, opened 2017-06-15
Add a setAppNamespace() convenience function to replace literal calls

ravage84:
I noticed we are doing this quite often. Thought we could replace it with a convencience function. No big deal, though.


markstory (commented):
Seems reasonable to me.

Issue #10743, opened 2017-06-08
RedisEngine Cache::increment / Cache::decrement do not honor duration


daywalker313:
This is a (multiple allowed):
* [x] bug

Version: master branch & probably others

Cache::increment / Cache::decrement used with Redis does not honor the duration settings of the configuration, instead the ttl is -1 (infinite).

--- 

Example:
Cache::increment('testkey_incr', 1, '2_minutes_duration');
Cache::write('testkey_write', 1, '2_minutes_duration');

---

Redis-CLI data:
tux@tux:~$ redis-cli
127.0.0.1:6379> GET cake_testkey_incr
"1"
127.0.0.1:6379> GET cake_testkey_write
"1"
127.0.0.1:6379> TTL cake_testkey_incr
(integer) -1
127.0.0.1:6379> TTL cake_testkey_write
(integer) 97

---

Reason:
In RedisEngine.php, incrBy/decrBy is used, which does not set / update the key ttl:
return (int)$this->_Redis->incrBy($key, $offset);

---

Expected behaviour:
The configuration's duration should be honored (either by resetting or initially setting the ttl).
I didn't find the current behaviour documented anywhere, it should definitely get a huge remark in the documentation as this seems inconsistent.

---

Solution:
Before returning the result, use phpredis setTimeout function to set the duration / ttl
$redis->setTimeout($key, $ttl);

---

PS: Looking at the class I'm also wondering why in L244/245 the decision was made to issue 2 write calls for the ttl setting instead of using "setTimeout", which is presumably a faster and cleaner way to set the ttl?

markstory:
I suspect that we don't set the TTL on increment() as neither the apcu, wincache or redis clients support settings a TTL when incrementing a counter. Because of this the cache backend interface doesn't include a TTL parameter.

I've replaced the setex call with settimeout as you pointed out - Thank you :+1:

daywalker313:
But wouldn't the same make sense for increment / decrement as well?
Just because it isn't implemented for other engines doesn't make it right, especially if the documentation says otherwise. 
If "duration" only works for 50% of the function calls, it should either be renamed or discarded.

Increment / decrement is often used for concurrency locks due to its atomic nature (I think that was even proposed somewhere in the book). 
But if duration doesn't work, it's not possible to configure a simple, automatic deadlock evasion.

My proposal stands:
A setTimeout call after each successful increment/decrement to reset the ttl for that key. That would be consistent with the add/write behaviour.
[...]

PR #10786, opened 2017-06-16
[RedisEngine] Respect duration by using setTimeout for incr & decr

ghost:
This refers to issue #10743 and fixes infinite TTL for keys created with increment/decrement (in Redis only).


markstory:
Thanks. I can take care of the other atomic cache backends.

markstory:
New pull request open with test cases in #10796

PR #10791, opened 2017-06-16
Fix disabled attribute calculation.

markstory:
Take the work done in #10770 and account for the complex options and non-empty disabled option sets.

Refs #10770

### DIFF
--- a/src/View/Helper/FormHelper.php
+++ b/src/View/Helper/FormHelper.php
@@ -2577,17 +2577,7 @@ class FormHelper extends Helper
         if ($context->hasError($field)) {
             $options = $this->addClass($options, $this->_config['errorClass']);
         }
-        $isDisabled = false;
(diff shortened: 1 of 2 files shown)


dereuromark:
The protected method wraps this nicely 👍

saeideng:
also please update doc block 
now is
```
      * - `disabled` - mixed - Either a boolean indicating disabled state, or the string in
     *   a numerically indexed value.
```

dereuromark:
@saeideng : bool|string then

saeideng:
bool ,string and array @dereuromark

dereuromark:
My point was more about not using mixed ever ;) but listing the concrete types if possible.

markstory:
@saeideng I updated the doc block. I didn't add type information as IDE's and other tools can't read that deeply into doc blocks.

alexdd55:
+1

PR #10796, opened 2017-06-19
Make Redis::increment/decrement() set a TTL

markstory:
Continuing the work done in #10786 and adding tests. I've not modified the Memcached driver as memcached values will have a TTL set when the key is initially created.

I am targeting `master` with these changes as it is reasonable to expect that counters would work consistently across Redis and Memcached implementations (which they didn't) and these changes are normalizing behavior across backends.


lorenzo (commented):
Looks good to me :)

Issue #10795, opened 2017-06-17
Fixtures from plugin and subfolder


mirko-pagliai:
I'm trying to load a fixture that is in a subdirectory of a plugin. My plugin is `MysqlBackup` and the subfolder is `Sqlite` (uppercase first).

Reading from [CookBook](https://book.cakephp.org/3.0/en/development/testing.html#loading-fixtures-in-your-test-cases), I would expect the correct syntax to be:

        plugin.mysql_backup.sqlite/articles

But running phpunit, I get this:

> PHP Fatal error:  Uncaught UnexpectedValueException: Referenced fixture class "MysqlBackup\Test\Fixture\Sqlite/articlesFixture" not found.

Note: incorrect slash and class name with lowercase initial.  
From here it seems that the subfolder should have the uppercase letter: in my case this is so, but the CookBook provides examples (`app.blog/articles`) with the lowercase letter (in other words: how should I do if the subfolder was `sqlite`?).

So the only way I have to load my fixture is this syntax:

        plugin.mysql_backup.sqlite\Articles

And this works.

This is a bit strange, does not it seem to you?

ADmad:
Does using ` 'plugin.MysqlBackup.Sqlite/Articles' work?

mirko-pagliai:
@ADmad no:

> Uncaught UnexpectedValueException: Referenced fixture class "MysqlBackup\Test\Fixture\Sqlite/ArticlesFixture" not found

markstory:
I can take a look what you are describing should work.

mirko-pagliai:
Another thing: fixture is not loaded anyway.  
I mean. Until now I used `$autoFixtures` as `false`, I was getting errors by just specifying the class of fixtures.

In other words, this is the only syntax that does not give me any errors:

    public $autoFixtures = false;

    public $fixtures = [
        'plugin.mysql_backup.sqlite\Articles',
    ];

If I try to load fixtures:

        $this->loadFixtures('Articles');

I get: `UnexpectedValueException: Referenced fixture class Articles not found`

markstory:
Closing as a pull request is open now.

PR #10799, opened 2017-06-20
Fix plugin fixtures in sub-directories not resolving.

markstory:
We need to flip / into \ to correctly generate the namespace for fixtures in subdirectories.

Refs #10795


josegonzalez (approved):
Thought we normally don't have comments in core code...

markstory:
@josegonzalez I found this code a bit confusing as its monkeying around with strings and has little context, so I thought it was a reasonable spot to add a comment. I'm happy to remove it though.

josegonzalez:
It's all good, figured thats why you added the comments :)

PR #10802, opened 2017-06-20
Fix some issues that came up while using strict_types

cleptric:
If fixed some issues while using `strict_types`. Most of them caused by passing `null` to internal php functions.

### DIFF
--- a/src/Core/functions.php
+++ b/src/Core/functions.php
@@ -51,7 +51,7 @@ if (!function_exists('h')) {
             } else {
                 $text = '(object)' . get_class($text);
             }
-        } elseif (is_bool($text)) {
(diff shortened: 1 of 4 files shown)


src/Core/functions.php
```diff
@@ -199,7 +199,7 @@ function env($key, $default = null)
                 return (!empty($_SERVER['HTTPS']) && $_SERVER['HTTPS'] !== 'off');
             }
 
-            return (strpos(env('SCRIPT_URI'), 'https://') === 0);
+            return (strpos((string)env('SCRIPT_URI'), 'https://') === 0);
```
dereuromark: Doesnt env() always return a string?

cleptric: Nope. You can pass a default value as second parameter bit this defaults to null.

PR #10804, opened 2017-06-21
3.next - Start CommandRunner

markstory:
This gets the new `CommandRunner` roughly in place. This class will eventually replace `ShellDispatcher` as the way shell commands are run. It integrates the `Application` with the console environment, and through a new hook method applications are able to selectively enable console commands, or rename commands.

This is by no means complete, but it will unblock work around `ConsoleIntegrationTest` and enable smaller pull requests for future work.

Refs #10716

### DIFF
--- a/src/Console/CommandRunner.php
+++ b/src/Console/CommandRunner.php
@@ -1,0 +1,86 @@
+<?php
+/**
+ * CakePHP(tm) : Rapid Development Framework (http://cakephp.org)
+ * Copyright (c) Cake Software Foundation, Inc. (http://cakefoundation.org)
+ *
+ * Licensed under The MIT License
+ * For full copyright and license information, please see the LICENSE.txt
+ * Redistributions of files must retain the above copyright notice.
+ *
+ * @copyright     Copyright (c) Cake Software Foundation, Inc. (http://cakefoundation.org)
+ * @link          http://cakephp.org CakePHP(tm) Project
+ * @since         3.5.0
+ * @license       http://www.opensource.org/licenses/mit-license.php MIT License
+ */
+namespace Cake\Console;
+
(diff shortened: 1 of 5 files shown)


HavokInspiration:
Really liking the fact that we can change the `root` command !
And 👍 on the error messages.

src/Console/CommandRunner.php
```diff
@@ -0,0 +1,123 @@
...
+{
+    /**
+     * @var \Cake\Http\BaseApplication
+     */
+    protected $app;
+
+    /**
+     * @var string
+     */
+    protected $root;
```
HavokInspiration: Maybe these properties could have more detailed docblocks ?

lorenzo:
Looks good to me :)

src/Console/CommandRunner.php
markstory: Sure thing.

tests/TestCase/Shell/CompletionShellTest.php
```diff
@@ -258,7 +258,7 @@ public function testSubCommandsAppDuplicatePluginNoDot()
         $this->Shell->runCommand(['subcommands', 'sample']);
         $output = $this->out->output;
 
-        $expected = "derp load sample\n";
+        $expected = "derp load returnValue sample withAbort\n";
```
jeremyharris: You might want to exchange this with some `$this->assertContains()` assertions instead, to prevent future breaks.

(If not, no big deal, it'll be changed in the future cli integration test PR as I ran into the same issue.)

markstory: I'd rather the tests be a little fragile in order to get more correctness.

PR #10807, opened 2017-06-21
Fix usage of contain finder options for hasMany relations

### LINKED ISSUE #10223: Contain usage of custom finder not working
This is a (multiple allowed):

* [ ] bug
* [x] enhancement

* CakePHP Version: 3.4.0
* Platform and Target: Apache 2, PHP 7.1.1, MySQL 5.7

### What you did

```
        $jobs = $this->Jobs->find('active')
            ->contain(['JobLogs' => [
                'strategy' => 'select',
                'finder' => ['orphan' => ['node_id' => $nc->node_id()]]
            ]]);
```

Jobs is returned with all JobLogs records.

### What happened

Confirmed that finder is not called.

```
    public function findOrphan(Query $query, array $options = [])
    {
        dd('findOrphan');
    }
```

Finder for contain should be used instead of find "All".

cleptric:
This fixes #10223 as the option didn't work anymore for `hasMany` relations.
I didn't created a new trait for `_extractFinder` and instead copied `Association::_extractFinder()`.

@lorenzo Thank you for your help 😄

### DIFF
--- a/src/ORM/Association/Loader/SelectLoader.php
+++ b/src/ORM/Association/Loader/SelectLoader.php
@@ -163,7 +163,14 @@ class SelectLoader
             $options['fields'] = [];
         }
 
-        $fetchQuery = $finder()
+        if (isset($options['finder'])) {
+            list($finderName, $opts) = $this->_extractFinder($options['finder']);
+            $query = $finder()->find($finderName, $opts);
+        } else {
+            $query = $finder();
+        }
+
+        $fetchQuery = $query
             ->select($options['fields'])
             ->where($options['conditions'])
             ->eagerLoaded(true)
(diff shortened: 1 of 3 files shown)


tests/TestCase/ORM/QueryTest.php
```diff
@@ -2817,6 +2817,108 @@ public function testContainFinderCanSpecifyOptions()
...
+                    'finder' => [
+                        'published' => [
+                            'title' => 'Foo'
+                        ]
+                    ]
+                ]
+            ]);
+
+        $this->assertSame(2, count($resultWithArticles->first()->articles));
+        $this->assertSame(2, count($resultWithArticlesArray->first()->articles));
```
markstory: You can use`assertCount()` to save some typing and get better failure messages.

src/ORM/Association/Loader/SelectLoader.php
```diff
@@ -163,7 +163,14 @@ protected function _buildQuery($options)
             $options['fields'] = [];
         }
 
-        $fetchQuery = $finder()
+        if (isset($options['finder'])) {
+            list($finderName, $opts) = $this->_extractFinder($options['finder']);
+            $query = $finder()->find($finderName, $opts);
+        } else {
+            $query = $finder();
+        }
```
lorenzo: Another way to write this without the `else`

```
$query = $finder();
if (...) {
 ...
 $query = $query->find(..)
} // no else
```

src/ORM/Association/Loader/SelectLoader.php
```diff
@@ -194,6 +201,33 @@ protected function _buildQuery($options)
...
+     * The following will call the finder 'translations' with the value of the finder as its options:
+     * $query->contain(['Comments' => ['finder' => ['translations']]]);
+     * $query->contain(['Comments' => ['finder' => ['translations' => []]]]);
+     * $query->contain(['Comments' => ['finder' => ['translations' => ['locales' => ['en_US']]]]]);
+     *
+     * @param string|array $finderData The finder name or an array having the name as key
+     * and options as value.
+     * @return array
+     */
+    protected function _extractFinder($finderData)
```
robertpustulka: Is it possible to support multiple finders, eg `['finder' => ['translations' => [], 'foo', 'bar']`?

Is is supported anywhere else?

lorenzo: it is not possible and not supported anywhere else.

ADmad:
I made updates as per the feeback.

Issue #10800, opened 2017-06-20
ServerRequest::getRequestTarget() does not return query string when using IntegrationTestCase


ravage84:
According to the [API docs of ServerRequest::getRequestTarget()](https://api.cakephp.org/3.4/class-Cake.Http.ServerRequest.html#_getRequestTarget), the function should also return the query string.

> By default this will return the application relative path without base directory, and the **query string** defined in the SERVER environment.

I was just testing a middleware which checks the request target for some given strings.
While my middleware worked fine when I checked through the browser, the related integration test using **IntegrationTestCase** failed.

I found out, that **ServerRequest::getRequestTarget()** returns the query string just fine normally, but does not when using **IntegrationTestCase**.

URL = /servlet/com.piso.xyz.ecom.gui.DispatcherServlet?function=ReturnxyzContent&amp;language=de&amp;agency=AGENCY

Normal Request through Browser
$path = /servlet/com.piso.xyz.ecom.gui.DispatcherServlet?function=ReturnxyzContent&amp;language=de&amp;agency=AGENCY
$request instanceOf \Cake\Http\ServerRequest
$request->uri->path = /servlet/com.piso.xyz.ecom.gui.DispatcherServlet
$request->uri->query = function=ReturnxyzContent&amp;language=de&amp;agency=AGENCY

Integraton Test
$path = /servlet/com.piso.xyz.ecom.gui.DispatcherServlet
$request instanceOf \Cake\Http\ServerRequest
$request->uri->path = /servlet/com.piso.xyz.ecom.gui.DispatcherServlet
$request->uri->query = _empty_

I traced it down:
IntegrationTestCase::_url()
- $parameter & $query are OK

IntegrationTestCase::_buildRequest()
- $tokenUrl = /servlet/com.piso.xyz.ecom.gui.DispatcherServlet?function=ReturnxyzContent&amp%3Blanguage=de&amp%3Bagency=AGENCY
- $url = /servlet/com.piso.xyz.ecom.gui.DispatcherServlet

It seems the changes in https://github.com/cakephp/cakephp/commit/d59ac1bcc2b7b8ceb56f94dd6c95eead7a712108 from https://github.com/cakephp/cakephp/pull/6481 changed this behavior, though at the time focusing on fixing a similar but reversed problem with the legacy ``$request::here()``.

markstory:
How could someone reproduce building a request that has the incorrect request target data?

ravage84:
Sorry, thought it would be clear:

[...]

PR #10808, opened 2017-06-22
Fix incorrect PSR7 request target.

markstory:
The IntegrationTestCase was not setting the correct server keys to set the URI state correctly for PSR7 methods.

Refs #10800


ravage84 (approved):
Many thanks, @markstory for the fast fix. This resolves my issue. Could test it successfully with myapp test suite.

markstory:
Thank you for verifying the fix @ravage84

PR #10811, opened 2017-06-22
Improved collections performance in common use cases

lorenzo:
This improves the performance of most collection methods sometimes
by making it 100% faster compared to the previous implementation.

The improvement is not for free, with this change we are giving up
the lazyness feature (not iterate unless requested). The giving
up of the feature only happens when the collection has been initialized
with an array, as oposed to being initialized with another iterator or
generator; and this is the reson I consider this change a safe one.

For the curious, the improvement comes from the (sad) fact that calling
functions in php is extremenly expensive, specially when using iterators
since each iteration will call at least 4 functions (valid, next, current, key).

This becomes even worse as `IteratorIterator` does not have any optimizations,
so for each wrapped iterator, the number of functions is multiplied by 2
for each iteration.

The change proposed here will unwrap nested iterators as much as possible
to avoid the function call explosion, and in some (safe) cases, will
iterator the collection immediately as an array before wrapping it again
in an iterator.

I was inspired by Haskell when implementing this, as the language is lazy
by default, but the compiler optimizes the cases where code is safe to be
called strictly. Thats is called strictness analysis.

### DIFF
--- a/src/Collection/Iterator/ExtractIterator.php
+++ b/src/Collection/Iterator/ExtractIterator.php
@@ -75,7 +75,7 @@ class ExtractIterator extends Collection
     /**
      * {@inheritDoc}
      *
-     * We perform here some stricness analysys so that the
+     * We perform here some strictness analysis so that the
      * iterator logic is bypassed entirely.
      *
      * @return \Iterator
--- a/src/Collection/Iterator/ExtractIterator.php
+++ b/src/Collection/Iterator/ExtractIterator.php
@@ -14,7 +14,9 @@
  */
 namespace Cake\Collection\Iterator;
 
+use ArrayIterator;
 use Cake\Collection\Collection;
+use Cake\Collection\CollectionInterface;
 
 /**
  * Creates an iterator from another iterator that extract the requested column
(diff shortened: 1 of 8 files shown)


src/Collection/Iterator/ExtractIterator.php
```diff
@@ -69,4 +71,37 @@ public function current()
...
+            return $this;
+        }
+
+        // ArrayIterator can be traversed strictly.
+        // Let's do that for performance gains
+
+        $callback = $this->_extractor;
+        $res = [];
+
+        foreach ($iterator->getArrayCopy() as $k => $v) {
```
markstory: I'm guessing that getting an array copy is part of the optimization windfall as you skip all the function calls to the Iterator interface methods?

src/Collection/Iterator/FilterIterator.php
```diff
@@ -37,9 +46,50 @@ class FilterIterator extends Collection
...
+            return $filter;
+        }
+
+        // ArrayIterator can be traversed strictly.
+        // Let's do that for performance gains
+
+        $callback = $this->_callback;
+        $res = [];
+
+        foreach ($iterator as $k => $v) {
```
markstory: Should you be using `getArrayCopy()` here?

src/Collection/Iterator/StoppableIterator.php
```diff
@@ -78,4 +80,41 @@ public function valid()
...
+        if (!$iterator instanceof ArrayIterator) {
+            return $this;
+        }
+
+        // ArrayIterator can be traversed strictly.
+        // Let's do that for performance gains
+
+        $callback = $this->_condition;
+        $res = [];
+        $stop = false;
```
markstory: You never use this variable.

markstory (commented):
Nice work. Great job cleaning up the tests.

src/Collection/Iterator/ExtractIterator.php
lorenzo: That's right

src/Collection/Iterator/FilterIterator.php
lorenzo: No, since the iterator itself is an argument for the callback. The callback can inspect the iterator state and if we for each the array, the the iteration would not be changing as we go further.

Using the array would have been a lot faster, but it also breaks a feature we actually use.

markstory: I thought that the iterator state might be important, but wanted to ask to make sure I understood the code.

ravage84:
@lorenzo how can we make sure that future changes do not break these pptimisations?
Let's say in a year somebody got a good idea but that breaks these optimisations. How can we make sure we detect that without keeping your PR here in the back of our heads?

May be we need even more concrete hints as https://github.com/cakephp/cakephp/pull/10811/files#diff-f952390778b4008d8a958cbabddf3d62R94

lorenzo:
@ravage84 we could have a benchmarking test suite, but I'm not up for the task myself

PR #10815, opened 2017-06-23
3.next - Update dirty method for get/set

albertcansado:


cleptric:
Thank you 
Looks like we missed that when deprecating the mixed mode method 😃 
I don't know if it's a good idea to remove all tests for `dirty()`.
I handeld that by just adding new tests for the new methods and let the old tests untouched.

ADmad:
Yeah test for old methods should be kept.

markstory:
@ADmad The entity tests for dirty weren't changed from what I could see.

PR #10816, opened 2017-06-23
3.next RFC ConsoleIntegrationTestCase

jeremyharris:
This is the first attempt at creating a new `ConsoleIntegrationTestCase` class that allows users to create simple integration tests against shells.

```php
public function testRoutesShell()
{
    $this->exec('routes check /users/index');
    $this->assertExitCode(Shell::CODE_SUCCESS);
}
```

The `exec()` method also allows an array of "answers" to be fed to an interactive shell:

```php
public function testRoutesShell()
{
    $this->exec('my_plugin.ama', [
        'yes', // answer "yes" to first input request
        'cats', // answer "cats" to second
    ]);
}
```

The current assertions are:

- assertExitCode: assert that the shell exited with the expected code
- assertOutputContains: assert that stdout contains expected string
- assertErrorContains: assert that stderr contains expected string

Open questions:

1. What other assertion methods, if any, would be helpful?
2. <del>Parsing argv from a string is hard because of quoting options. The current solution execs a basic script that just captures what PHP actually receives in `$argv`, which is obviously slower than a regexp solution. Any regexp experts out there willing to help with parsing? `ConsoleIntegrationTestCaseTest::testCommandStringToArgs()` contains a good example of what we expect from a string. </del>


src/TestSuite/ConsoleIntegrationTestCase.php
```diff
@@ -0,0 +1,191 @@
...
+
+    /**
+     * Asserts shell exited with the expected code
+     *
+     * @param int $expected Expected exit code
+     * @return void
+     */
+    public function assertExitCode($expected)
+    {
+        $this->assertSame($this->_exitCode, $expected);
```
markstory: You should provide a nice failure message here.

src/TestSuite/ConsoleIntegrationTestCase.php
```diff
@@ -0,0 +1,191 @@
...
+    }
+
+    /**
+     * Asserts `stdout` contains expected output
+     *
+     * @param string $expected Expected output
+     * @return void
+     */
+    public function assertOutputContains($expected)
```
markstory: This should take a message, so people can write informative test failures.

tests/argv.php
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/php -q
```
markstory: What about windows? Will this work there?

jeremyharris: Nope. Good point, I totally spaced out here. 

Before trying to come up with something dumb/clever I'm going to wait to see if anyone has a better idea on how to solve this problem.

jeremyharris:
I tried more regexping but still couldn't get it (I got close once a while ago). So, I decided to take a look at how C parses `argv` and it turns out it's just good old fashioned loop programming. Inspired by that, I wrote a small parser in about 5 minutes as opposed to the hours of regexp attempts 😄

src/TestSuite/ConsoleIntegrationTestCase.php
```diff
@@ -172,19 +181,35 @@ protected function _makeRunner()
...
+                $prevChar = substr($command, $i - 1, 1);
+                if ($prevChar === '\\') {
+                    $arg[strlen($arg) - 1] = $char;
+                }
+                $inQuote = $inQuote ? false : true;
+                continue;
+            }
+            $arg .= $char;
+        }
+        $argv[] = $arg;
```
markstory: This will work well across platforms.

jeremyharris:
I worked on the parser and tuned it up a bit, now works properly. Any recommendations for additional test cases are welcome :)

ravage84:
How about ``assertOutputRegExp()``? Analogous with https://api.cakephp.org/3.4/class-Cake.TestSuite.IntegrationTestCase.html#_assertResponseRegExp

markstory:
@jeremyharris Would you be able to update the testing chapter in the book on how to use this new feature?

jeremyharris:
@markstory absolutely

PR #10818, opened 2017-06-23
Fix last line excerpt not retrievable.

ndm2:
The line number needs to be decremented before testing for existence,
otherwise the last line number would be out of bounds.


markstory:
Thanks!

inoas:
Thank You!

PR #10819, opened 2017-06-24
Add version shell and --version

markstory:
By having aliases and standalone shells we avoid having 'command' logic in the command runner. It also means that application developers can replace the default --version behavior if they want to.

Part of #10716

### DIFF
--- a/src/Console/CommandRunner.php
+++ b/src/Console/CommandRunner.php
@@ -64,11 +64,17 @@ class CommandRunner
     }
 
     /**
-     * Replace the alias map for a runner.
+     * Replace the entire alias map for a runner.
      *
      * Aliases allow you to define alternate names for commands
      * in the collection. This can be useful to add top level switches
      * like `--version` or `-h`
+     *
+     * ### Usage
+     *
+     * ```
+     * $runner->setAliases(['--version' => 'version']);
+     * ```
      *
      * @param array $aliases The map of aliases to replace.
      * @return $this
--- a/src/Console/CommandRunner.php
+++ b/src/Console/CommandRunner.php
@@ -19,6 +19,7 @@ use Cake\Console\ConsoleIo;
 use Cake\Console\Exception\StopException;
 use Cake\Console\Shell;
 use Cake\Http\BaseApplication;
+use Cake\Shell\VersionShell;
 use RuntimeException;
 
 /**
@@ -41,6 +42,13 @@ class CommandRunner
     protected $root;
 
     /**
+     * Alias mappings.
+     *
+     * @var array
+     */
+    protected $aliases = [];
+
+    /**
      * Constructor
      *
      * @param \Cake\Http\BaseApplication $app The application to run CLI commands for.
@@ -50,6 +58,26 @@ class CommandRunner
     {
         $this->app = $app;
         $this->root = $root;
+        $this->aliases = [
+            '--version' => 'version'
+        ];
+    }
+
+    /**
+     * Replace the alias map for a runner.
+     *
+     * Aliases allow you to define alternate names for commands
+     * in the collection. This can be useful to add top level switches
+     * like `--version` or `-h`
+     *
+     * @param array $aliases The map of aliases to replace.
+     * @return $this
+     */
+    public function setAliases(array $aliases)
+    {
+        $this->aliases = $aliases;
+
+        return $this;
     }
 
     /**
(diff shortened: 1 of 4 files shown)


src/Console/CommandRunner.php
```diff
@@ -50,6 +58,26 @@ public function __construct(BaseApplication $app, $root = 'cake')
...
+     * Aliases allow you to define alternate names for commands
+     * in the collection. This can be useful to add top level switches
+     * like `--version` or `-h`
+     *
+     * @param array $aliases The map of aliases to replace.
+     * @return $this
+     */
+    public function setAliases(array $aliases)
+    {
+        $this->aliases = $aliases;
```
saeideng: Set whole aliases ?
Maybe need to loop

markstory: I considered an array merge. But then we need another method to remove aliases.

saeideng: OK
but please change method name to `replaceAliases()`

tests/TestCase/Shell/CompletionShellTest.php
```diff
@@ -116,7 +116,7 @@ public function testCommands()
         $output = $this->out->output;
 
         $expected = 'TestPlugin.example TestPlugin.sample TestPluginTwo.example unique welcome ' .
-            "cache i18n orm_cache plugin routes server i18m sample testing_dispatch\n";
+            "cache i18n orm_cache plugin routes server version i18m sample testing_dispatch\n";
```
ADmad: "i18m" ?

src/Console/CommandRunner.php
markstory: Many of our other setters 'replace' and don't use replace in the name. Why should this one be different?

tests/TestCase/Shell/CompletionShellTest.php
markstory: Yes its an intentional typo in the classname. Not sure from whence it came, but its definitely there.

src/Console/CommandRunner.php
ravage84: Instead of renaming this method to ``replaceAliases()`` we could have ``addAlias()`` /  ``addAliases()`` and ``removeAlias()`` / ``removeAliases()``. This way, it could be clearer that ``setAliases()`` replaces all aliases previously set.

ravage84:
> It also means that application developers can replace the default --version behavior if they want to.

I like that, cause I might want to display the application version instead of CakePHP's. That would be easily possible, right @markstory ?

src/Console/CommandRunner.php
```diff
@@ -41,6 +42,13 @@ class CommandRunner
     protected $root;
 
     /**
+     * Alias mappings.
```
ravage84: It might help to understand how the mapping should be set, if we show an example like https://github.com/cakephp/cakephp/pull/10819/files#diff-5746a5b6845a932f92018ff514ae2000R62

src/Console/CommandRunner.php
```diff
@@ -50,6 +58,26 @@ public function __construct(BaseApplication $app, $root = 'cake')
     {
         $this->app = $app;
         $this->root = $root;
+        $this->aliases = [
+            '--version' => 'version'
+        ];
+    }
+
+    /**
+     * Replace the alias map for a runner.
```
ravage84: "the whole alias map"?

ravage84 (commented):
Looks good!

tests/TestCase/Shell/CompletionShellTest.php
HavokInspiration: I think I'm the one who added this, but it was a while ago (maybe years ?).

If I remember correctly it was for a test case linked to the `CompletionShell`. However, the details elude me now...

src/Console/CommandRunner.php
markstory: That is a quite a few methods for something I think will be used very seldomly. I would prefer to start out with the smallest set of methods possible and expand the API if people find the features we offer lacking.

markstory:
> I like that, cause I might want to display the application version instead of CakePHP's. That would be easily possible, right 

Correct.

lorenzo:
This looks good to me :)

src/Console/CommandRunner.php
saeideng: Because you can not change this behavior in future if people want it for BC resone(s)

markstory: I don't see why we would have to change it. As @ravage84 mentioned we could add methods like `addAliases()` and `removeAlias()` if we need to add/remove into the existing set of aliases.

PR #10830, opened 2017-06-27
Fix error when default value is reported as CURRENT_TIMESTAMP() by mysql - Fixes #10829

### LINKED ISSUE #10829: Creating fixtures fails if mysql reports current_timestamp() with parenthesis
This is a (multiple allowed):

* [x] bug
* [ ] enhancement

* CakePHP Version: 2.9.9.
* Platform and Target: MacOS, mysqld Ver 10.2.6-MariaDB for osx10.12 on x86_64 (Homebrew).

### What you did
I have a table that has a field called updated_at that was created as follows: ``` `updated_at` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP```
However my db version reports it back as ``` `updated_at` timestamp NOT NULL DEFAULT current_timestamp() ON UPDATE current_timestamp()```.
The bug occurs when trying to run tests that create fixtures for the affected table.

### What happened
An error is displayed and tests won't run.
```
2017-06-27 11:51:16 Error: Fixture creation for "ad_campaigns" failed "SQLSTATE[42000]: Syntax error or access violation: 1067 Invalid default value for 'updated_at'"
PHP Warning:  Fixture creation for "ad_campaigns" failed "SQLSTATE[42000]: Syntax error or access violation: 1067 Invalid default value for 'updated_at'" in /Users/kurre/mcn/nativeflow/Vendor/cakephp/cakephp/lib/Cake/TestSuite/Fixture/CakeTestFixture.php on line 244
```
Looking at the input to DboSource->_execute() I see the column being described as ``` `updated_at` timestamp DEFAULT 'current_timestamp()' NOT NULL``` which is obviously not correct.

I expect tests to run and no error to be raised.

I am opening a PR with a fix.

kurrestahlberg:
My database version (mysqld  Ver 10.2.6-MariaDB for osx10.12 on x86_64 (Homebrew)) reports the default value current_timestamp with parenthesis (like so: ``` `updated_at` timestamp NOT NULL DEFAULT current_timestamp() ON UPDATE current_timestamp()```) which confuses the CakePHP mysql driver to think it is a string and that leads to the following error when running tests: 
```
2017-06-27 11:51:16 Error: Fixture creation for "ad_campaigns" failed "SQLSTATE[42000]: Syntax error or access violation: 1067 Invalid default value for 'updated_at'"
PHP Warning:  Fixture creation for "ad_campaigns" failed "SQLSTATE[42000]: Syntax error or access violation: 1067 Invalid default value for 'updated_at'" in /Users/kurre/mcn/nativeflow/Vendor/cakephp/cakephp/lib/Cake/TestSuite/Fixture/CakeTestFixture.php on line 244
```
This patch will allow for the parenthesis and that fixes my issue (#10829)

I am unsure how to write a test for this as it seems to be database version specific issue.

### DIFF
--- a/lib/Cake/Model/Datasource/Database/Mysql.php
+++ b/lib/Cake/Model/Datasource/Database/Mysql.php
@@ -352,7 +352,8 @@ class Mysql extends DboSource {
 			if (in_array($fields[$column->Field]['type'], $this->fieldParameters['unsigned']['types'], true)) {
 				$fields[$column->Field]['unsigned'] = $this->_unsigned($column->Type);
 			}
-			if (in_array($fields[$column->Field]['type'], array('timestamp', 'datetime')) && strtoupper($column->Default) === 'CURRENT_TIMESTAMP') {
+			if (in_array($fields[$column->Field]['type'], array('timestamp', 'datetime')) &&
+					in_array(strtoupper($column->Default), array('CURRENT_TIMESTAMP', 'CURRENT_TIMESTAMP()'))) {
 				$fields[$column->Field]['default'] = null;
 			}
 			if (!empty($column->Key) && isset($this->index[$column->Key])) {


markstory (changes requested):
Any chance we could get an updated unit test covering this scenario?

markstory:
If you need help with the tests let me know. I'm happy to help out with them :smile:

kurrestahlberg:
Yeah, if you could help that would be great. I've looked at the tests that currently are there in the 2.x branch and it tests only that the default value is ignored, but what the default is reported as cannot be IMO changed.

Is there a way to custom build (or alter) whatever describe()'s input without hacking Model?

markstory:
I took a look and you're right that testing this would only be possible by using mocks to muck around with the query results, which I'd rather not have.

PR #10834, opened 2017-06-28
3.next - Add Help shell

markstory:
This shell is a replacement for CommandListShell. It uses the provided CommandCollection to list the commands instead of scanning the application. This will allow it to play nicer with standalone applications.

I've not yet fixed the hardcoded `cake` in the help output. But I'll be doing that in my next pull request for the new console libraries.

Refs #10716

### DIFF
--- a/src/Console/CommandRunner.php
+++ b/src/Console/CommandRunner.php
@@ -19,6 +19,7 @@ use Cake\Console\ConsoleIo;
 use Cake\Console\Exception\StopException;
 use Cake\Console\Shell;
 use Cake\Http\BaseApplication;
+use Cake\Shell\HelpShell;
 use Cake\Shell\VersionShell;
 use RuntimeException;
 
@@ -59,7 +60,9 @@ class CommandRunner
         $this->app = $app;
         $this->root = $root;
         $this->aliases = [
-            '--version' => 'version'
+            '--version' => 'version',
+            '--help' => 'help',
+            '-h' => 'help',
         ];
     }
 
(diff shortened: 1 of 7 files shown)


src/Console/CommandRunner.php
```diff
@@ -157,11 +161,15 @@ protected function getShell(ConsoleIo $io, CommandCollection $commands, $name)
...
         }
-        $classOrInstance = $commands->get($name);
-        if (is_string($classOrInstance)) {
-            return new $classOrInstance($io);
+        $instance = $commands->get($name);
+        if (is_string($instance)) {
+            $instance = new $instance($io);
+        }
+        // TODO Should this be an interface?
+        if (method_exists($instance, 'setCommandCollection')) {
```
markstory: I'm contemplating doing this at the same time as propagating the root command name. Right now we have a few places we hard code `cake` when it should really reflect the root command name given to the `CommandRunner`.

lorenzo: keeping the root command name everywhere would be nice :)

dereuromark: Interface would sure be a cleaner way :+1:

markstory: Ok. I'll do both of those in a subsequent pull request :+1:

lorenzo:
Good job @markstory

src/Shell/CommandListShell.php
```diff
@@ -145,7 +146,7 @@ public function getOptionParser()
             'help' => 'Get the listing as XML.',
             'boolean' => true
         ])->addOption('version', [
-            'help' => 'Prints the currently installed version of CakePHP.',
+            'help' => 'Prints the currently installed version of CakePHP. (deprecated - use `cake --version` instead)',
```
ravage84: I do not understand. If you want to see the help, you have to use `cake --version` instead? That does not make much sense to me.

markstory: I was trying to document that `cake command_list --version` is deprecated.

PR #10837, opened 2017-06-28
Add a method to remove certain middleware from a scope

cleptric:
This improves the scoped middleware feature #10308.
The newly added `removeMiddleware` method let you remove certain middleware from a given scope.

My use case is based on the new auth plugin. I wanted to apply the session authenticator to the `/` scope, but not to `/api`, as this should only use a stateless authentication.

```
Router::scope('/', function (RouteBuilder $routes) {
    ....
    $routes->registerMiddleware('authentication', new AuthenticationMiddleware($service));
    $routes->scope('/', function ($routes) {
        $routes->applyMiddleware('authentication');
    });
    $routes->scope('/api', function ($routes) {
        $routes->removeMiddleware('authentication');
    });
    ...
});
```

### DIFF
--- a/src/Routing/RouteBuilder.php
+++ b/src/Routing/RouteBuilder.php
@@ -924,4 +924,20 @@ class RouteBuilder
 
         return $this;
     }
+
+    /**
+     * Remove a middleware from the current route scope.
+     *
+     * Requires middleware to be registered via `registerMiddleware()`
+     *
+     * @param string[] ...$names The names of the middleware to apply to the current scope.
+     * @return $this
+     * @see \Cake\Routing\RouteCollection::addMiddlewareToScope()
+     */
+    public function removeMiddleware(...$names)
+    {
+        $this->_collection->removeMiddleware($this->_path, $names);
+
+        return $this;
+    }
 }
--- a/src/Routing/RouteCollection.php
+++ b/src/Routing/RouteCollection.php
@@ -32,6 +32,20 @@ class RouteCollection
 {
 
     /**
+     * Indicates to apply a middleware to the current route scope.
+     *
+     * @var string
+     */
+    const APPLY = 'apply';
+
+    /**
+     * Indicates to remove a middleware to the current route scope.
+     *
+     * @var string
+     */
+    const REMOVE = 'remove';
+
+    /**
      * The routes connected to this collection.
      *
      * @var array
@@ -72,6 +86,13 @@ class RouteCollection
      * @var array
      */
     protected $_middlewarePaths = [];
+
+    /**
+     * A map of paths and the list of middleware that should be removed.
+     *
+     * @var string
+     */
+    protected $_ignoreMiddlewarePaths = [];
 
     /**
      * Route extensions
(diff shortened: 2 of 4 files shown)


lorenzo:
@cleptric you example looks super confusing to anyone reading a large piece of code.

First you state that anything inside the block should use `authentication` but suddenly that is not the case?

I would say this feature is confusing enough to not even support it. This can be done by re-opening the scope and apply different middleware as desired. That's how other frameworks like elixir do it.

cleptric:
After consulting @lorenzo and trying out different stuff I'm still under the impression that it's not possible to remove a middleware when it was registered and applied to a stronger scope before.

I looked into `RoutingMiddleware::__invoke` again, and it calls `Router::getRouteCollection()->getMatchingMiddleware($request->getUri()->getPath());` once per request.
As this is feature is altering the same route collection, you could also change the example code above to:
```
Router::scope('/', function (RouteBuilder $routes) {
    ...
    $routes->registerMiddleware('authentication', new AuthenticationMiddleware($service));
    $routes->scope('/', function ($routes) {
        $routes->applyMiddleware('authentication');
    });
    ...
});

Router::scope('/api', function (RouteBuilder $routes) {
    ....
    $routes->scope('/', function ($routes) {
        $routes->removeMiddleware('authentication');
    });
    ....
});
```
which maybe is less confusing 🙂

markstory:
> This can be done by re-opening the scope and apply different middleware as desired. That's how other frameworks like elixir do it.

We can't easily provide middleware replacement. Right now the model is that middleware is registered to a name and then applied. Perhaps that is the wrong model though. What if there was no registration, and instead `applyMiddleware()` worked like `applyMiddleware('auth', ...)`? This would let us directly inject instances in each scope and override middleware defined in parent scopes by redefining them in the child scopes.

lorenzo:
@markstory I think it is not the idea model. The middleware should not be attached to a scope name, but to a scope builder.

Overriding and removing is not ideal because it is not declarative, and routes are declarative. I would not like to remember where the state is kept and changed.

markstory:
@lorenzo Registered middleware are not attached to individual scopes. There is only one keyspace for all middleware presently. In my last comment I was proposing something like

``php
Router::scope('/', function ($routes) {
  $routes->applyMiddleware('auth', new AuthMiddleware(/* session config */));

  $routes->scope('/api', function ($routes) {
    $routes->applyMiddleware('auth', new AuthMiddleware(/* token config */));
  });
});
```

When finding applicable middleware for a request, the scopes are traversed from shortest to longest. Scopes that have longer URLs (more specific) would be able to replace middleware by re-using the name.

This would allow us to not have a `registerMiddleware()` method and not need `removeMiddleware()` at all.

lorenzo:
@markstory that's sound like a possible solution, but I think it leaves out things like different middleware for the same scope but different http verbs, for example

markstory:
@lorenzo Yes. We'd have to rely on middleware to do HTTP verb checks themselves.

cleptric:
@markstory I'll give your proposal a try and update that PR accordingly. Could the updated `applyMiddleware` method have an optional second parameter to overwrite and empty an existing name?
I still thing there is a need to remove certain middleware completely and not just change the config.

markstory:
You could replace the problematic middleware with a no-op handler. When do you foresee needing to entirely remove middleware?

cleptric:
Passing a no-op handler sounds like a good idea too 🙂

cleptric:
A new and probably better way to solve this is #10909

Issue #10836, opened 2017-06-28
QueryBuilder - lastInsertId() returns "0"


odan:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.4.7
* Platform and Target: MariaDB (5.5.5-10.1.21), PHP Version 7.1.1, Apache, Windows 7

### What you did

Trying to insert a new record and get the lastInsertId (AUTO_INCREMENT).

### What happened
The result of lastInsertId() is always "0".

Example code:

```sql
CREATE TABLE `users` (
  `id` INT NOT NULL AUTO_INCREMENT,
  `username` VARCHAR (255),
  PRIMARY KEY (`id`)
) ENGINE = INNODB CHARSET = utf8 COLLATE = utf8_unicode_ci ;
```

```php
// composer require cakephp/database
require_once __DIR__ . '/vendor/autoload.php';

// Database settings
$config = array(
    'driver' => 'mysql',
    'host' => '127.0.0.1',
    'port' => '3306',
    'username' => 'root',
    'password' => '',
    'database' => 'test',
    'charset' => 'utf8',
    'collation' => 'utf8_unicode_ci',
    'prefix' => '',
);

$driver = new \Cake\Database\Driver\Mysql($config);
$db = new \Cake\Database\Connection(['driver' => $driver]);

$row = [
    'username' => 'max'
];

$result = $db->insert('users', $row);

$newId = $result->lastInsertId();
echo $newId;  // value is "0"
```

If I try the same with a native PDO query it works, but not with the CakePHP QueryBuilder.

```php
$pdo = $db->getDriver()->connection();
$stmt = $pdo->query("SELECT LAST_INSERT_ID()");
$lastId = $stmt->fetchColumn(); // This works. Returns 1 or more.
```

### What you expected to happen
Returns the ID of the last inserted row or sequence value.

lorenzo:
@odan Try with the query builder instead of working directly with the connection instead while this is being investigated.

lorenzo:
@odan Here you have an example of using the query builder for inserting https://github.com/cakephp/cakephp/blob/master/tests/TestCase/Database/QueryTest.php#L4125-L4136

markstory:
What does `$pdo->lastInsertId('users')` return in your PDO example. That is what CakePHP  is doing under the hood.

odan:
Hi @lorenzo and @markstory 

[...]

PR #10839, opened 2017-06-29
Fixed Xdebug issue in combination with isConnected() #10836

odan:
Xdebug uses PHP's var_dump() function for displaying variables.
https://xdebug.org/docs/display

Setting a breakpoint at or before a line with a lastInsertId() method call the __debugInfo() method of the Driver object was magically called by Xdebug and PHP.

Within the __debugInfo() a new SELECT 1 statement was called and cleared the lastInsertId of the connection. Instead of a new SELECT we only check the state of the connection object now.

You can find more details here: #10836 (comment)

### DIFF
--- a/src/Database/Driver.php
+++ b/src/Database/Driver.php
@@ -386,7 +386,7 @@ abstract class Driver
     public function __debugInfo()
     {
         return [
-            'connected' => $this->isConnected()
+            'connected' => $this->_connection !== null
         ];
     }
 }


lorenzo:
What was the issue?

odan:
You can find the more details here: https://github.com/cakephp/cakephp/issues/10836#issuecomment-311912675

jeremyharris:
Should there be a comment above this change to reference the issue so it doesn't accidentally get reversed?

ADmad:
@jeremyharris The commit message does include the ticket number so if you check file's history or blame you would have the reference to the issue :)

PR #10842, opened 2017-06-30
Add root command propagation and CommandCollectionInterface

markstory:
Add root command propagation and an interface for shells to hint that they need the command collection in the new runner.

Presently the `HelpShell` will not work for applications using `ShellDispatcher`. I'm thinking of putting that requirement into the error messages HelpShell creates.

### DIFF
--- a/tests/TestCase/Console/CommandRunnerTest.php
+++ b/tests/TestCase/Console/CommandRunnerTest.php
@@ -241,6 +241,30 @@ class CommandRunnerTest extends TestCase
         $this->assertSame(99, $result);
     }
 
+    /**
+     * Ensure that the root command name propagates to shell help
+     *
+     * @return void
+     */
+    public function testRunRootNamePropagates()
+    {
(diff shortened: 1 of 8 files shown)


tests/TestCase/Console/CommandRunnerTest.php
```diff
@@ -241,6 +241,30 @@ public function testRunValidCommandReturnInteger()
...
+        $commands = new CommandCollection(['sample' => SampleShell::class]);
+        $app->method('console')->will($this->returnValue($commands));
+
+        $output = new ConsoleOutput();
+
+        $runner = new CommandRunner($app, 'widget');
+        $runner->run(['widget', 'sample', '-h'], $this->getMockIo($output));
+        $result = implode("\n", $output->messages());
+        $this->assertContains('widget sample [-h]', $result);
+        $this->assertNotContains('cake sample [-h]', $result);
```
lorenzo: 👏

lorenzo:
Looks great!

Issue #10658, opened 2017-05-15
Different dirty behaviour between newEntity($data) and patchEntity($entity, $data)


Slessi:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.4.5
* Platform and Target: PHP 7.1.4, Apache 2.4.25, MySQL 5.7.16

### What you did
I save a model `NocForm` which has one `Document` association
```php
// In my controller
private function saveForm($id)
{
    $this->loadModel('NocForms');

    $data = $this->request->data;

    if (empty($id)) {
        $form = $this->NocForms->newEntity($data);
    } else {
        $form = $this->NocForms->getForm($id);
        $form = $this->NocForms->patchEntity($form, $data);
    }

    if ($form->errors() || !$this->NocForms->save($form)) {
        $this->addDangerMessage('There was a problem saving your form');
        return;
    }
}
```

Inside `$data` I have a bunch of `NocForm` fields, and also this `tmp_file` field which comes from a file input. Irrespective of whether I pass a file to my file input or not, `tmp_file` will be set. The contents will change accordingly when a file is included.
> *Note*: If relevant, the file input `name` is `document[tmp_file]`
```json
{
    "document": {
        "tmp_file": {
            "tmp_name": "",
            "error": 4,
            "name": "",
            "type": "",
            "size": 0
        }
    }
}
```

Inside `DocumentsTable` I have the following events defined. `beforeMarshal` checks for the `tmp_file` property and if a file is actually contained in it, and if found it extracts the information relevant to me. If no file is found to be attached I simply unset the field.
```php
public function beforeMarshal(Event $event, ArrayObject $data, ArrayObject $options)
{
    if (!empty($data['tmp_file'])) {
        $file = $data['tmp_file'];

        if (!empty($file['tmp_name'])) {
            $data['extension'] = pathinfo($file['name'], PATHINFO_EXTENSION);
            $data['content_type'] = mime_content_type($file['tmp_name']);
            $data['file_size'] = filesize($file['tmp_name']);
            $data['date_uploaded'] = new \Cake\I18n\Time();
            $data['file_path'] = md5(uniqid()).'.'.$data['extension'];
        } else {
            unset($data['tmp_file']);
        }
    }
}
```

`beforeSave` will check if we are saving a new file, and uploads it to our bucket if so.
```php
public function beforeSave(Event $event, EntityInterface $entity, ArrayObject $options)
[...]

PR #10844, opened 2017-07-01
Make one()/many() more consistent with empty associations

markstory:
When an association is cleaned out by a beforeMarshal hook, the entity property should not be marked as dirty as the related entity has no properties to persist.

Refs #10658

### DIFF
--- a/src/ORM/Marshaller.php
+++ b/src/ORM/Marshaller.php
@@ -595,11 +595,12 @@ class Marshaller
         }
 
         foreach ((array)$options['fields'] as $field) {
-            if (array_key_exists($field, $properties)) {
-                $entity->set($field, $properties[$field]);
-                if ($properties[$field] instanceof EntityInterface) {
-                    $entity->setDirty($field, $properties[$field]->isDirty());
-                }
+            if (!array_key_exists($field, $properties)) {
+                continue;
+            }
+            $entity->set($field, $properties[$field]);
+            if ($properties[$field] instanceof EntityInterface) {
+                $entity->setDirty($field, $properties[$field]->isDirty());
             }
         }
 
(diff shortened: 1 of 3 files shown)


lorenzo:
This is a big change in the behaviour of the marshaller. Better not target master?

src/ORM/Marshaller.php
```diff
@@ -211,6 +211,14 @@ public function one(array $data, array $options = [])
             $entity->set($properties);
         }
 
+        // Don't flag clean association entities as
+        // dirty so we don't persist empty records.
+        foreach ($properties as $field => $value) {
+            if ($value instanceof EntityInterface) {
+                $entity->dirty($field, $value->dirty());
```
lorenzo: This is where I think the big change is... a good one, though.

Does it make more sense to only traverse the properties that we know are from associations and that are allowed in the associations whitelist?

markstory: I'm a little worried about this change as well. I contemplated changing `merge()` so that it marked the association property as dirty if the property did not exist before but had a record marshalled. I think having merge() and one() be consistent is what I think should change.

The list of properties here are the ones that have already been set into the entity on line 211, so it should be less risky to propagate the dirty state.

markstory:
@lorenzo Moved to 3.next.

src/ORM/Marshaller.php
lorenzo: Agree, having the behaviour of both functions work consistently is desirable

lorenzo:
@markstory there's a conflict now

src/ORM/Marshaller.php
saeideng: i think `dirty()` is deprecated

markstory: This was updated during the rebase.

PR #10846, opened 2017-07-02
3.next Add Console.buildCommands event

markstory:
Add an event that provides the console analogy to `Server.buildMiddleware`. This event will allow auto-wiring plugins to inject their commands.

Refs #10716


ravage84 (commented):
Looks good

PR #10848, opened 2017-07-03
Add compatibility to HelpShell for ShellDispatcher apps.

markstory:
Not everyone will upgrade right away, and HelpShell shouldn't brick until they do update.

### DIFF
--- a/src/Shell/HelpShell.php
+++ b/src/Shell/HelpShell.php
@@ -19,6 +19,8 @@ use Cake\Console\CommandCollection;
 use Cake\Console\CommandCollectionAwareInterface;
 use Cake\Console\ConsoleOutput;
 use Cake\Console\Shell;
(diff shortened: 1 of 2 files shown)


src/Shell/HelpShell.php
```diff
@@ -85,6 +85,32 @@ public function main()
...
+     * @return array
+     */
+    protected function getCommands()
+    {
+        $task = new CommandTask($this->getIo());
+        $nested = $task->getShellList();
+        $out = [];
+        foreach ($nested as $section => $commands) {
+            $prefix = '';
+            if ($section !== 'CORE' && $section !== 'app') {
```
dereuromark: Should `app` also be `APP` for consistency?

markstory: Not without possibly causing other issues with other consumers of the CommandTask.

PR #10851, opened 2017-07-04
Add an interface for the Application.

markstory:
I wasn't too happy about the coupling between Console and Http, and this new interface lets us not have that. Instead Http and Console will depend on Core (which they already do). I also felt a bit icky adding more typehints against concreteish implementations, as they have caused pain in the past.

I contemplated adding a ConsoleApplication and HttpApplication interface to  separate the `middleware` and `console` hooks, but thought that might be a bit overkill.

I'm completely open to scrapping this, or adding the additional interfaces based on what you folks think.


robertpustulka:
I'd really like to see the seperate Console/Http app interfaces. I mentioned something like that on slack AFAIR some time ago. It could help develop console-only applications and sepere the console package out of the core.

markstory:
@robertpustulka Would you put the console/http application interfaces in the http & console packages, or in core?

robertpustulka:
@markstory maybe in the core. Standalone console package would depend on core anyway. And we'd keep application-related interfaces and classes in one place. What about `Cake\Core\Application` namespace for that?

markstory:
@robertpustulka I've updated the interfaces now.

lorenzo (approved):
I like how his turned out!

PR #10856, opened 2017-07-04
Improve exception message for Email::_validateEmail

cleptric:
This improves the exception message, thrown when the `Email` class calls `_validateEmail`.
The current message `Invalid email: ""` is IMO not helpful at all.
This adds a context to tell which of the properties is not valid and also handles empty passed properties.

Inspired by @lorenzo


markstory:
Nice work!

Issue #8933, opened 2016-06-03
Pagination 'NOT' notation fails with i18n


bancer:
This is a (multiple allowed):
- [x] bug
- [ ] enhancement
- [ ] feature-discussion (RFC)
- CakePHP Version: 2.8.4.
- Platform and Target: Linux, Apache, MySQL.
### What you did

In the controller:

```
$this->Paginator->settings['conditions'] = array(
    'NOT' => array(
        'I18n__title.content' => ''
    ),
);
```
### Expected Behavior

Properly paginated page
### Actual Behavior

Error: SQLSTATE[42S22]: Column not found: 1054 Unknown column 'I18n__title.content' in 'where clause'
### Solution

I fixed it with:

```
$this->Paginator->settings['conditions'] = array(
    'I18n__title.content !=' => ''
);
```

but it would be good if array notation with 'NOT' would not fail.

lorenzo:
Can you try this? 

```php
NOT' => array(
        array('I18n__title.content' => '')
    ),
```

I think that is the correct usage for `NOT`

bancer:
```
$this->Paginator->settings['conditions'] = array(
    'NOT' => array(
        array(
            'I18n__title.content' => ''
        )
    ),
);
``` 
generates the same SQL error.

markstory:
Closing as a pull request is open now.

bancer:
The problem seems to exist only when limit is set.

PR #10857, opened 2017-07-04
Pagination 'NOT' notation fails with i18n

bancer:
Some unit tests for https://github.com/cakephp/cakephp/issues/8933


markstory:
The tests are all passing. Is that expected?

bancer:
I was not able to reproduce the problem in CakePHP. I will try a bit later in another project and come back here with the results.

markstory:
Ok. I'll merge this and if you can reproduce let us know 😄

Issue #10853, opened 2017-07-04
Empty form values for fields with array-like values


Asgraf:
This is a (multiple allowed):
[x] bug
CakePHP Version: 3.4.9

### What you did
I have created custom JsonType that maps database field with json string to an object (JsonObject class implements JsonSerializable, IteratorAggregate, ArrayAccess,Countable interfaces)

Then i have created following test form:
````PHP
//Users/edit.ctp
echo $this->Form->create($user);
debug($user['metadata']->toArray());

debug($user['metadata']['aff_id']);
debug($this->Form->getSourceValue('Users.metadata.aff_id'));

echo $this->Form->control('Users.metadata.aff_id',['type'=>'number']);
echo $this->Form->end();
````

### What happened
```
[
	'activation_token' => '2r3y0MStLinhS14Ax45Z5BY4x251Far6gwii0fPb',
	'expiry_notifications_active' => '1',
	'bnd_notifications_active' => '1',
	'developer_notifications_active' => '1',
	'aff_id' => '1',
	'language' => 'pl',
	'timezone' => 'Europe/Warsaw',
	'default_currency' => '1'
]

'1'
null

(input form with empty value)
```
### What you expected to happen
```
[
	'activation_token' => '2r3y0MStLinhS14Ax45Z5BY4x251Far6gwii0fPb',
	'expiry_notifications_active' => '1',
	'bnd_notifications_active' => '1',
	'developer_notifications_active' => '1',
	'aff_id' => '1',
	'language' => 'pl',
	'timezone' => 'Europe/Warsaw',
	'default_currency' => '1'
]

'1'
'1'

(input form with '1' value)
```
### The  problem
https://github.com/cakephp/cakephp/blob/master/src/View/Form/EntityContext.php#L262
When the entity field contains plain php array EntityContext->val() returns correct value, but in case of array-like object null is returned instead.

markstory:
How would you recommend we traverse into arbitrary userland objects? Would using `ArrayAccess` in that specific case work?

Asgraf:
ArrayAccess should work in my case.
My JsonObject class is intended to be used with Cake\Utility\Hash::get() and Cake\Utility\Hash::extract()

markstory:
Pull request open now.

PR #10859, opened 2017-07-05
Enable EntityContext to read from ArrayAccess properties

markstory:
Traverse into ArrayAccess properties as custom type mappings should be readable if they quack like arrays.

Refs #10853


src/View/Form/EntityContext.php
```diff
@@ -259,7 +260,7 @@ public function val($field, $options = [])
 
             return $this->_schemaDefault($part, $entity);
         }
-        if (is_array($entity)) {
+        if (is_array($entity) || $entity instanceof ArrayAccess) {
             $key = array_pop($parts);
```
Asgraf: Please note that array_pop will not work with ArrayAccess
https://stackoverflow.com/questions/8788002/arrayaccess-and-native-php-array-functions

Issue #10858, opened 2017-07-04
ConsoleIo overwrite failed if output is less than previous overvrite


yark:
This is a:

* [x] bug

* CakePHP Version: 3.4.9.
* Platform and Target: Mac os x 10.11 console.

### What you did
```
<?php
namespace App\Shell;

use Cake\Console\Shell;

class TmertestShell extends Shell
{
    public function testTimer($sec = 10)
    {
        $this->out($sec, 0);
        for ($i = $sec; $i > 0; $i--) {
            $this->_io->overwrite($i, 0);
            sleep(1);
        }
        $this->out();
    }
}

```

```$ ./bin/cake timertest testTimer```

### What happened
If length of `message` param of `_io->overwrite` ($i) less than last output it unable to overwrite next output.
$sec = 10
```
10
9
98
97
...
91
```

$sec = 101
```
101
100
99
9998
9997
...
9912
9911
9910
999
9998
...
9991
```

*) I was showing each iteration in new line, but it actually happens in single line

### What you expected to happen
I expect correct overwrite
```
101
100
99
98
...
11
10
9
8
...
1
```

I can fix it providing third argument length

```
    public function testTimer($sec = 10)
    {
        $this->out($sec, 0);
        $ow_length = strlen((string)$sec);
        for ($i = $sec; $i > 0; $i--) {
            $this->_io->overwrite($i, 0, $ow_length);
            sleep(1);
        }
        $this->out();
    }
```

but in this case it will output extra space after message.

markstory:
I'll see if I can reproduce this locally.

markstory:
Closing as a pull request is open.

PR #10860, opened 2017-07-05
Fix overwrite() not replacing longer content.

markstory:
When overwriting content with shorter content, we need to pad out to the longest content that was recently output. This allows countdown timers to be built properly.

Refs #10858

### DIFF
--- a/src/Console/ConsoleIo.php
+++ b/src/Console/ConsoleIo.php
@@ -209,9 +209,7 @@ class ConsoleIo
         // Store length of content + fill so if the new content
         // is shorter than the old content the next overwrite
         // will work.
-        if ($fill > 0) {
-            $this->_lastWritten = $newBytes + $fill;
-        }
+        $this->_lastWritten = $newBytes + $fill;
     }
 
     /**
--- a/src/Console/ConsoleIo.php
+++ b/src/Console/ConsoleIo.php
@@ -205,6 +205,13 @@ class ConsoleIo
         if ($newlines) {
             $this->out($this->nl($newlines), 0);
         }
+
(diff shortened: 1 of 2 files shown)


src/Console/ConsoleIo.php
```diff
@@ -205,6 +205,13 @@ public function overwrite($message, $newlines = 1, $size = null)
         if ($newlines) {
             $this->out($this->nl($newlines), 0);
         }
+
+        // Store length of content + fill so if the new content
+        // is shorter than the old content the next overwrite
+        // will work.
+        if ($fill > 0) {
```
ravage84: Couldn't you just add the value of ``$fill``? If it's zero, it would still add up fine 😼

tests/TestCase/Console/ConsoleIoTest.php
```diff
@@ -333,6 +333,57 @@ public function testOverwrite()
     }
 
     /**
+     * Test overwriting content with shorter content
+     *
+     * @return void
+     */
+    public function testOverwriteShorterContent()
+    {
+        $number = strlen('12345');
```
ravage84: Shouldn't this be ``$length`` for better understanding?

ravage84 (approved):
Looks good apart form the naming

src/Console/ConsoleIo.php
markstory: Good point.

lorenzo (approved):
Looks good, but tests are failing :(

markstory:
@lorenzo Test failure was a just one travis segment that fell down. I re-ran it and it passed.

PR #10863, opened 2017-07-05
Pagination 'NOT' notation fails with i18n

bancer:
See https://github.com/cakephp/cakephp/issues/8933.

### DIFF
--- a/lib/Cake/Model/Behavior/TranslateBehavior.php
+++ b/lib/Cake/Model/Behavior/TranslateBehavior.php
@@ -216,7 +216,7 @@ class TranslateBehavior extends ModelBehavior {
  *
  * @param Model $Model The model being worked on.
  * @param array $query The query array to append joins to.
- * @param string $addFields The fields being joined.
+ * @param array $addFields The fields being joined.
  * @return array The modified query
  */
 	protected function _addAllJoins(Model $Model, $query, $addFields) {
--- a/lib/Cake/Model/Behavior/TranslateBehavior.php
+++ b/lib/Cake/Model/Behavior/TranslateBehavior.php
@@ -245,11 +245,26 @@ class TranslateBehavior extends ModelBehavior {
  * @return array The list of translated fields that are in the conditions.
  */
 	protected function _checkConditions(Model $Model, $query) {
-		$conditionFields = array();
 		if (empty($query['conditions']) || (!empty($query['conditions']) && !is_array($query['conditions']))) {
-			return $conditionFields;
+			return array();
 		}
-		foreach ($query['conditions'] as $col => $val) {
+		return $this->_getConditionFields($Model, $query['conditions']);
+	}
+
+	/**
+	 * Extracts condition field names recursively.
+	 *
+	 * @param Model $Model The model being read.
+	 * @param array $conditions The conditions array.
+	 * @return array The list of condition fields.
+	 */
+	protected function _getConditionFields(Model $Model, $conditions) {
+		$conditionFields = array();
+		foreach ($conditions as $col => $val) {
+			if (is_array($val)) {
+				$subConditionFields = $this->_getConditionFields($Model, $val);
+				$conditionFields = array_merge($conditionFields, $subConditionFields);
+			}
 			foreach ($this->settings[$Model->alias] as $field => $assoc) {
 				if (is_numeric($field)) {
 					$field = $assoc;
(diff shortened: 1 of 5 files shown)


bancer:
It seems there is inconsistency in how `buildQuery('all'` and `buildQuery('count'` generates joins and handles 'I18n__title.content' field. 'all' uses I18n__title alias but 'count' uses TranslateArticleModel alias. Any suggestions how that should be fixed?

markstory:
Have you tried tracking down where the joins are being incorrectly generated in `TranslateBehavior::beforeFind()`? I see some count related logic there that might shed some light onto why count and all finds behave differently.

bancer:
Not yet but I will do.

bancer:
What should be the expected result of TranslateBehavior::beforeFind()? When I18n__title alias is supplied in the input should it be converted to TranslateArticleModel in the output?

bancer:
It seems that when TranslateBehavior::beforeFind() is called for 'all' query then _addJoin() method is called in order to generate the joins but when it is called for 'count' query then _addJoin() is not called and the join is constructed right in beforeFind().

bancer:
It seems TranslateBehavior::_checkConditions() does not process I18n__title properly.

lib/Cake/Test/Case/Model/Behavior/TranslateBehaviorTest.php
```diff
@@ -1478,7 +1478,7 @@ public function testBeforeFindAllI18nConditions() {
 					'table' => (object)array(
 						'tablePrefix' => '',
 						'table' => 'article_i18n',
-						'schemaName' => 'test',
+						'schemaName' => 'cakephp_test',
```
markstory: Why do these need to change?

lib/Cake/Model/Behavior/TranslateBehavior.php
```diff
@@ -191,24 +208,31 @@ public function beforeFind(Model $Model, $query) {
...
+		return $addFields;
+	}
 
-		$this->runtime[$Model->alias]['virtualFields'] = $Model->virtualFields;
+/**
+ * Appends all necessary joins for translated fields.
+ *
+ * @param Model $Model The model being worked on.
+ * @param array $query The query array to append joins to.
+ * @param string $addFields The fields being joined.
```
markstory: How can this parameter be a string, if it is being looped over?

lib/Cake/Test/Case/Model/Behavior/TranslateBehaviorTest.php
```diff
@@ -1443,4 +1443,164 @@ public function testNoExtraRowsForAssociatedTranslations() {
...
+	}
+
+	public function testBeforeFindCountI18nConditions() {
+		$this->loadFixtures('TranslateArticle', 'TranslatedArticle', 'User');
+		$TestModel = new TranslatedArticle();
+		$TestModel->cacheQueries = false;
+		$TestModel->locale = 'eng';
+		$expected = array(
+			'conditions' => array(
+				'NOT' => array('I18n__title.content' => ''),
```
markstory: Shouldn't this be `TranslateArticle__title.content`?

bancer: Perhaps you meant `TranslateArticleModel.title`. That could be this way I guess but the conditions array is not changed in TranslateBehavior. It is returned unaltered - as supplied in $query parameter. I didn't want to change much in TranslateBehavior.

markstory: Yes I meant TranslateArticleModel. Shouldn't the query conditions use aliases that exist in the current model.

bancer: I am not sure about that. TranslateArticleModel alias exists here only when `find('count'` is called. It does not exist here when `find('all'` is called. The join array with TranslateArticleModel is a bit suspicious because the join condition does not have `TranslateArticleModel.title` column. So it is not fully clear what records are joined by that when a model has more than one translatable fields.

Altering the query conditions only in `find('count'` will introduce more inconsistencies to the code comparing it with `find('all'`. That does not seems to be the right way to go. Altering the query conditions in both `find('count'` and  `find('all'` will require more changes because in that case the joins in `find('all'` will need to be altered to make TranslateArticleModel available. Perhaps it would be better to open another issue to make the behavior of `find('count'` and `find('all'` consistent using the same joins and altering conditions in the same way when necessary.

markstory: My bad. I misread the code last time.

bancer:
The build for PHP 5.3 failed before the tests started.

PR #10864, opened 2017-07-05
Stop warnings when using count in QueryCompiler in PHP 7.2

### LINKED ISSUE #10862: PHP 7.2 has changed count's behavior causing problems with QueryCompiler
This is a (multiple allowed):

* [x] bug
* [ ] enhancement

* CakePHP Version: 3.4.8
* Platform and Target: Windows 7 (but it will affect all platforms running PHP 7.2)

PHP 7.2 has changed the behavior of count, it emits a warning if the object to be counted is not an array, or it does not implement the Countable interface:

https://wiki.php.net/rfc/counting_non_countables

As a result the _sqlCompiler created function in the QueryBuilder now emits many warnings.  This line is the problem:

https://github.com/cakephp/cakephp/blob/master/src/Database/QueryCompiler.php#L128

The `$parts` variable can be many different things, including non-arrays and objects not implementing the Countable interface (for example instances of Cake\Database\Expression\ValuesExpression), and many different warnings will result during normal operations.

I presume that in future versions of PHP the old count behavior of returning 1 for non-arrays and non-Countable objects will be removed, so this should be fixed for that reason as well.

MCF:
- PHP 7.2 has changed the count function to emit warnings when the object
  being "counted" is not an array or does not implement the Countable
  interface.  The QueryCompiler was using count for objects that did not
  fall into those two categories.  This fixes that problem while
  maintaining the same logic (returning early when $parts is not set, is
  null, or has a count of 0).

Fixes https://github.com/cakephp/cakephp/issues/10862

### DIFF
--- a/src/Database/QueryCompiler.php
+++ b/src/Database/QueryCompiler.php
@@ -125,7 +125,9 @@ class QueryCompiler
     protected function _sqlCompiler(&$sql, $query, $generator)
     {
         return function ($parts, $name) use (&$sql, $query, $generator) {
-            if (!count($parts)) {
+            if (!isset($parts) ||
+                ((is_array($parts) || $parts instanceof \Countable) && !count($parts))
+            ) {
                 return;
             }
             if ($parts instanceof ExpressionInterface) {


src/Database/QueryCompiler.php
```diff
@@ -125,7 +125,9 @@ public function compile(Query $query, ValueBinder $generator)
     protected function _sqlCompiler(&$sql, $query, $generator)
     {
         return function ($parts, $name) use (&$sql, $query, $generator) {
-            if (!count($parts)) {
+            if (!isset($parts) ||
```
ADmad: Wouldn't just checking `if (empty($parts))` work?

MCF: Not for the full set of conditionals (if that was the question?).
Objects with Countable interfaces and count() == 0 are not considered "empty" by the empty function.

Also I am mimicing the old conditional's (possible) reliance on count returning 1 when an object was not an array or Countable.  In other words I _know_ that this new conditional acts exactly the same as `!count($parts)` but without the warning emitted.

MCF: As a follow on, $parts can be an integer (I've dumped out the type of parts so this is not a hypothetical,  sometimes an integer is passed in).  So the following case shows why empty would be a change from current behavior:

```PHP
$parts = 0;
!count($parts)   // == 1 or effectively true
empty($parts)    // == false
```

Issue #10855, opened 2017-07-04
Entity protected fields interferes with Digest Authentication


jigzat:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.3.15
* Platform and Target: Mac OS X Yosemite With Server App MySQL.

### What you did
I was trying to implement digest authentication and I have the Users table with the password field already being used for JWT so I added a second field called digest. Also inside the User Entity I added the field 'digest' to the Hidden Fields in order to hide this data from the API request.

```
protected $_hidden = [
        'password','mydigest'
    ];
```

### What happened
When I try to get into the site I receive an error that points me to the line 115 of the DigestAuthenticate.php file . This happens even with the default 'password' setting.

```
$field = $this->_config['fields']['password'];
$password = $user[$field];
```

### What you expected to happen
DigestAuthenticate relies on BaseAuthenticate that gets the user from the database but with the hidden fields stripped. In my case I had to replace 

`$user = $this->_findUser($digest['username']);`

with 
```
$users = TableRegistry::get('Users');
        $user = $users->find('all')
    	->where(['username' => $digest['username']])
    	->first();
```

Not sure if this is something desirable, of course I had to manually strip both 'password' and 'mydigest' from the user object but this brakes the settings pattern.

P.S. Remember, an issue is not the place to ask questions. You can use [Stack Overflow](https://stackoverflow.com/questions/tagged/cakephp)
for that or join the #cakephp channel on irc.freenode.net, where we will be more
than happy to help answer your questions.

Before you open an issue, please check if a similar issue already exists or has been closed before.

markstory:
Hidden fields are not elided from the database query results, and should be accessible to DigestAuthentication. Have you tried removing `mydigest` from the hidden fields? Does that resolve the problem?

jigzat:
Yes that was the first thing that I tried, but since the password field is being hidden by default it kinda creeped me out to remove it, since it's basically a second password. Of course DigestAuthenticate unsets the field but in all other parts the user object will have it. Not trying to pretend to be an expert here.

markstory:
@jigzat Even with the digest hash hidden, CakePHP should still be able to read the field. The hash check is done [here](https://github.com/cakephp/cakephp/blob/master/src/Auth/DigestAuthenticate.php#L120-L123) and the array access methods on entities do not 'protect' direct property reads.

ADmad:
[...]

PR #10867, opened 2017-07-06
Enable DigestAuth to work when digest key is hidden

markstory:
When the digest key is a hidden field it will not be included in the results of _findUser() which makes it impossible for users to authenticate. In the case where there was no password comparison done, and the passwordfield is a hidden field, removing the passwordField from the hidden list allows digest auth to work. Digest auth subsequently removes the password field after checking the key.

Refs #10855

### DIFF
--- a/src/Auth/BaseAuthenticate.php
+++ b/src/Auth/BaseAuthenticate.php
@@ -110,15 +110,22 @@ abstract class BaseAuthenticate implements EventListenerInterface
             return false;
         }
 
+        $passwordField = $this->_config['fields']['password'];
         if ($password !== null) {
             $hasher = $this->passwordHasher();
-            $hashedPassword = $result->get($this->_config['fields']['password']);
+            $hashedPassword = $result->get($passwordField);
             if (!$hasher->check($password, $hashedPassword)) {
                 return false;
             }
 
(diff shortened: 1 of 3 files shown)


src/Datasource/EntityTrait.php
```diff
@@ -420,7 +420,7 @@ public function hiddenProperties($properties = null)
     /**
      * Sets hidden properties.
      *
-     * @param array $properties An array of properties to treat as virtual.
+     * @param array $properties An array of properties to hide from array exports.
```
ravage84: Much better!

src/Auth/BaseAuthenticate.php
```diff
@@ -110,15 +110,22 @@ protected function _findUser($username, $password = null)
...
             if (!$hasher->check($password, $hashedPassword)) {
                 return false;
             }
 
             $this->_needsPasswordRehash = $hasher->needsRehash($hashedPassword);
-            $result->unsetProperty($this->_config['fields']['password']);
+            $result->unsetProperty($passwordField);
+        }
+        if ($password === null && in_array($passwordField, $result->getHidden())) {
+            $hidden = $result->getHidden();
```
ravage84: Couldn't this be moved above the if clause? This would result in ``getHidden()`` being called only once.

ravage84 (approved):

Issue #10813, opened 2017-06-23
Wrong count query results / Containment state of clean query copies only reset partially


ndm2:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.4.8

---

So there's two problems here

1. `\Cake\ORM\Query::cleanCopy()` partially resets containments
2. Count queries can return wrong results

I stumbled over this via **https://stackoverflow.com/questions/44366686/retrieve-no-translations-for-a-single-query-in-cakephp-3**

### The cause of the problems

Count queries are using clean query copies, and the docblock of `\Cake\ORM\Query::cleanCopy()` claims that the method would reset the containment state, but before **#10657**, [**there would be no changes to the containments at all**](https://github.com/cakephp/cakephp/blob/3.4.6/src/ORM/Query.php#L756).

The changes introduced with **#10657** will "reset" the containments that have been set in the `beforeFind` event, as [**the cloned eager loader where these containments have been applied**](https://github.com/cakephp/cakephp/blob/3.4.8/src/ORM/Query.php#L784), will be overwritten with [**a new eager loader cloned from the original query**](https://github.com/cakephp/cakephp/blob/3.4.8/src/ORM/Query.php#L768), which holds the containments that have been set on the query directly.

### Expected behavior

Given the description of `cleanCopy()`, one would expect that all containments are being removed. However, since clean copies are being used for counting, removing containments can cause wrong results, as containments might apply joins that can have an effect on the included rows, ie one would expect the containments to be kept.

Here's two tests to illustrate the behavior changes:

```php
public function testCleanCopyContain()
{
	$table = TableRegistry::get('Articles');
	$table->hasMany('Comments');
	$table->belongsTo('Authors');

	$table->eventManager()
		->on('Model.beforeFind', function (Event $event, $query) {
			$query->contain(['Authors']);
		});

	$query = $table
		->find()
		->contain(['Comments']);

	$copy = $query->cleanCopy();

	$expected = [
		'Comments' => [],
		'Authors' => []
	];
	$this->assertEquals($expected, $copy->getEagerLoader()->contain());
}

public function testCountWithInnerJoinContain()
{
	$table = TableRegistry::get('Articles');
	$table->hasMany('Comments');
	$table->belongsTo('Authors')->setJoinType('INNER');

	$result = $table->save($table->newEntity([
		'author_id' => null,
		'title' => 'title',
		'body' => 'body',
[...]

PR #10868, opened 2017-07-06
Reset eager loaders before triggering before find.

markstory:
In #10657 I made a change to reset eager loaders when creating a clone to fix state being shared across query clones. Because of method ordering this resulted in the eager loader being reset *after* the beforeFind was triggered, which dropped any contained associations added during that event. Re-ordering the method calls allows contained associations added during the beforeFind to be retained in the count() query.

Refs #10813


lorenzo:
Order based dependencies are the worst, glad you could find a way to solve the issue :)

PR #10871, opened 2017-07-06
Change function signature to be less confusing

jonathan-neugber:
In the documentation https://book.cakephp.org/3.0/en/orm/table-objects.html#beforesave it says:
```
The Model.beforeSave event is fired before each entity is saved. Stopping this event will abort the save operation. When the event is stopped the result of the event will be returned.
```
So it was confusing that the name of the function to stop the event is called `stopPropagation` as this sounds like the event is still valid but just further events are executed and **not** stopped.

I propose this simple change as it would make it more clear that the function actually stops the event.

### DIFF
--- a/src/Event/Event.php
+++ b/src/Event/Event.php
@@ -167,8 +167,19 @@ class Event
      * Stops the event from being used anymore
      *
      * @return void
+     * @deprecated 3.4.10 use stop() instead.
      */
     public function stopPropagation()
+    {
+        $this->_stopped = true;
+    }
+
+    /**
+     * Stops the event from being used anymore
+     *
+     * @return void
+     */
+    public function stop()
     {
         $this->_stopped = true;
     }
--- a/tests/TestCase/Event/EventTest.php
+++ b/tests/TestCase/Event/EventTest.php
@@ -73,6 +73,20 @@ class EventTest extends TestCase
     }
 
     /**
+     * Tests the event stopping property
+     *
+     * @return void
+     * @triggers fake.event
+     */
+    public function testStop()
+    {
+        $event = new Event('fake.event');
+        $this->assertFalse($event->isStopped());
+        $event->stop();
+        $this->assertTrue($event->isStopped());
+    }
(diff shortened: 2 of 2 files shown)


ionas:
If the Event behaves in any way similar to say JS/jQuery events `stopPropagation` is then a good name. It stops the event from further triggering anything. However it does not stop the process that triggered the event AFAIU. So `stop()` feels ambiguous to me.

dereuromark:
I agree with @ionas 
The actual confusing for 99% of all developers will actually come if this change lands.
The conventional term here is the current one, across all languages and toolings :)

See
- https://wiki.selfhtml.org/wiki/JavaScript/DOM/Event/stopPropagation
- https://api.jquery.com/event.stoppropagation/
- https://developer.mozilla.org/en-US/docs/Web/API/Event/stopPropagation
- http://docs.giderosmobile.com/reference/gideros/Event/stopPropagation
- https://api.drupal.org/api/drupal/vendor!symfony!event-dispatcher!Event.php/function/Event%3A%3AstopPropagation/8.2.x
- ... (google)

dakota:
At first I thought this might be a good change, but after some research, I agree with @dereuromark. The convention is `stopPropagation`, and we should probably follow that convention.

jonathan-neugber:
It just feels confusing for me as if it were like jQuery `stopPropagation` would only stop the propagation and not cause the event to 'fail' in the `beforeSave` context.

ionas:
@jonathan-neugber I can feel your confusion here but the thing that you should think/PR about is beforeSave() not the basic event system.

ADmad:
In CakePHP's context stopping propagation means preventing the next handler for the event being called. The method name is fine. If there is confusion then the description related to beforeSave could be improved

burzum:
I agree with @ADmad, let's improve the docs.

markstory:
Thank you for taking the time to put these changes together. I am not sure breaking convention with other event systems that developers use.

Issue #10870, opened 2017-07-06
Remove "Welcome to Cake..." banner for cake console commands.


notoriousturtle:
* [x] enhancement

* CakePHP Version: 3.3.7
* Platform and Target: 7.0.18-0ubuntu0.16.04.1

### What you did
Ran `bin/cake`

### What happened

Output:
```
Welcome to CakePHP v3.3.7 Console
---------------------------------------------------------------
App : src
Path: /var/www/vhosts/xxxxx/htdocs/src/
PHP : 7.0.18-0ubuntu0.16.04.1
---------------------------------------------------------------
```

This banner makes working with the console a bit dirty, as it is output with each command.

### What you expected to happen
Can we save this output for `bin/cake version` output? Or for a special `info` command option?

slince:
I agree with this. The description should be removed.

slince:
The shell library of cake is too tasteless. I think the it should be re-implemented, see  [symfony/console](https://github.com/symfony/console);

josegonzalez:
@notoriousturtle seems reasonable to me, though I think you can currently override a method to remove this from your own shells.

> The shell library of cake is too tasteless. I think the it should be re-implemented, see symfony/console;

@slince This sort of feedback isn't very useful as there isn't a list of changes/enhancements that we can act upon. Do you have specific feedback that we can work with?

robertpustulka:
@slince see #10716

slince:
@josegonzalez 

- Support for custom route for the command.
The command like  './cake foo_bar_baz foo_bar' lacks beauty.

- Increase flexibility
The library helped the developer do too much
```
$this->createFile('bower.json', $stuff);
```
I cannot replace the default  prompt if the target file exists. Perhaps this right should be given to the developer

- Remove the prefix "_" 
```
$this->_io->out('foo');
```


- If I execute a nonexistent command, the output will make me wonder.

```bash
>cake user_binding non_exists

Welcome to CakePHP v3.4.9 Console
---------------------------------------------------------------
App : src
Path: C:\www\slshop\src\
PHP : 5.6.19
---------------------------------------------------------------
Usage:
cake user_binding [-h] [-q] [-v]

Options:

--help, -h     Display this help.
--quiet, -q    Enable quiet output.
--verbose, -v  Enable verbose output.
```
You know, I need an accurate tip. like git.
```
>git commmit
git: 'commmit' is not a git command. See 'git --help'.

Did you mean this?
        commit
```

[...]

PR #10875, opened 2017-07-07
Fix php version 7 in console

mehdi-fathi:
fix issue #10870 for php version 7.


markstory:
Based on the discussion wouldn't it be better to remove the welcome banner entirely?

src/Console/Shell.php
```diff
@@ -84,6 +84,13 @@ class Shell
     public $OptionParser;
 
     /**
+     * A var for maintain phpversion()
+     *
+     * @var show php version
+     */
+    public $phpVersion;
```
markstory: We are trying to not add any new public properties. Can this be protected.

dereuromark: Why is this needed in the first place if the scope is only this welcome method?

src/Console/Shell.php
```diff
@@ -242,12 +249,14 @@ public function startup()
      */
     protected function _welcome()
     {
+        $phpInfo = explode('~', phpversion());
+        $this->phpVersion = (isset($phpInfo[0]) ? $phpInfo[0] : phpversion());
```
dereuromark: The isset is not necessary as it will always have a 0 element after an explode() of any content input.

src/Console/Shell.php
mehdi-fathi: @dereuromark Maybe this uses in another method later.

src/Console/Shell.php
inoas: I would return the full php version as `phpversion()` returns it anyway. And I do not see the reason to use a variable then.

src/Console/Shell.php
dereuromark: We dont code for maybes usually :)

src/Console/Shell.php
mehdi-fathi: @inoas you can see bin/cake in php 7.x.You should reed #10870

dereuromark: The consensus was to fix it the proper way by only displaying it alltogether for the mentioned main page and --version --help things. 
This here only fixes a small symptom, so let's please focus on the actual way to solve it right.

robertpustulka:
I don't see how this PR fixes #10870.

src/Console/Shell.php
markstory: I agree, given the usage of this variable it does not require a property at all.

dereuromark:
Resolved with https://github.com/cakephp/cakephp/pull/10879

Issue #10820, opened 2017-06-24
HTTP client redirect following doesn't include cookies from intermediary responses


thumbtech:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.4.7
* Platform and Target: Console (Windows 2012)

### What you did
I am using HTTP client to post credentials to an external website login form which requires several cookies & redirects (2 redirects, to be exact) to process ...

Example A (standalone PHP Script):

```php
$ch = curl_init();
curl_setopt($ch, CURLOPT_URL, $url);
curl_setopt($ch, CURLOPT_RETURNTRANSFER, 1);
curl_setopt($ch, CURLOPT_COOKIEFILE, 'cookie.txt'); 
curl_setopt($ch, CURLOPT_COOKIEJAR, 'cookie.txt');
curl_setopt($ch, CURLOPT_FOLLOWLOCATION, 1);
curl_setopt($ch, CURLOPT_POST, 1);
curl_setopt($ch, CURLOPT_POSTFIELDS, $post_data);
$response = curl_exec($ch);
```

Example B (CakePHP 3.4 shell):

```php
$http = new Client();
$response = $http->post($url, $post_data, ['redirect' => 99]);
```
Example C (same CakePHP 3.4 shell as B):

```php
$http = new Client();
$response = $http->post($url, $post_data);
while ($response->isRedirect()) {
    $location = $response->getHeader('Location');
    $url = $location[0]; // SEE CAVEAT BELOW
    $response = $http->get($url);
}
```

### What happened
Examples A & C return the content expected after log in. Example B "fails" (status 200, but external webserver just returns the login form indicating failed login attempt).

### What you expected to happen
I would expect all three examples to be functionally equivalent.

I attempted to debug myself but didn't get far ... couldn't even locate where the redirects are actually called for re-requesting!

CAVEAT: Example C simplifies the step where a redirect may be relative (which in my case they are) and I "expand" the previous url for this case. Such a case is (apparently) handled internally by lib_curl (Ex. A) and should be also by the HTTP class (Ex. B), but I couldn't check the logic there since I couldn't even locate the code :-) ... not sure if that is even the issue!

ndm2:
You're probably looking for the `Client` class' `redirect` option.

> Config options
> 
> You can set the following options when creating a client:
> 
> ...
> redirect - Number of redirects to follow. Defaults to false

[...]

PR #10876, opened 2017-07-07
Move redirects handling to the Client.

robertpustulka:
Fix for #10820

This would need a slightly different implementation for 3.5.


markstory (commented):
Looks good. What would need to change for 3.5?

robertpustulka:
@markstory the way the cookies are added to the response has changed.

Issue #10865, opened 2017-07-05
PHP 7.2 count warnings in \View\Form\EntityContext


MCF:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.4.8
* Platform and Target: PHP 7.2

Warnings emitted by count in PHP 7.2 shown below

```
Warning (2): count() [<a href='https://secure.php.net/function.count'>function.count</a>]: Parameter must be an array or an object that implements Countable [CORE\src\View\Form\EntityContext.php, line 485]
Code Context
$parts = object(Cake\ORM\Entity) {

	'[new]' => true,
	'[accessible]' => [
		'*' => true
	],
	'[dirty]' => [],
	'[original]' => [],
	'[virtual]' => [],
	'[errors]' => [],
	'[invalid]' => [],
	'[repository]' => 'HeadTypes'

}
$rootFallback = true
count - [internal], line ??
Cake\View\Form\EntityContext::_getTable() - CORE\src\View\Form\EntityContext.php, line 485
Cake\View\Form\EntityContext::_schemaDefault() - CORE\src\View\Form\EntityContext.php, line 280
Cake\View\Form\EntityContext::val() - CORE\src\View\Form\EntityContext.php, line 260
Cake\View\Helper\FormHelper::getSourceValue() - CORE\src\View\Helper\FormHelper.php, line 2843
Cake\View\Helper\FormHelper::_initInputField() - CORE\src\View\Helper\FormHelper.php, line 2571
Cake\View\Helper\FormHelper::__call() - CORE\src\View\Helper\FormHelper.php, line 1629
Cake\View\Helper\FormHelper::_getInput() - CORE\src\View\Helper\FormHelper.php, line 1242
Cake\View\Helper\FormHelper::control() - CORE\src\View\Helper\FormHelper.php, line 1125
...
```

ravage84:
Isn't this related to or at least similar to #10862 ?

markstory:
Yes it is.

PR #10878, opened 2017-07-07
Fix conditional using count to not emit warning in PHP 7.2

MCF:
- In this case it appears that only arrays with a size greater than 1 need
  to be dealt with by the main part of the _getTable function.  Every
  other kind of object should return $this->_tables[$this->_rootName].

Fixes https://github.com/cakephp/cakephp/issues/10865


markstory:
Thanks :clap:

PR #10879, opened 2017-07-08
Do not spam users with welcome message in every commands

### LINKED ISSUE #10870: Remove "Welcome to Cake..." banner for cake console commands.

rchavik:
This is a alternative of #10875

Closes #10870

### DIFF
--- a/src/Shell/CommandListShell.php
+++ b/src/Shell/CommandListShell.php
@@ -38,6 +38,22 @@ class CommandListShell extends Shell
     public $tasks = ['Command'];
 
     /**
+     * Displays a header for the shell
+     *
+     * @return void
+     */
+    protected function _welcome()
+    {
+        $this->out();
+        $this->out(sprintf('<info>Welcome to CakePHP %s Console</info>', 'v' . Configure::version()));
+        $this->hr();
+        $this->out(sprintf('App : %s', APP_DIR));
+        $this->out(sprintf('Path: %s', APP));
+        $this->out(sprintf('PHP : %s', phpversion()));
+        $this->hr();
+    }
+
+    /**
      * startup
      *
      * @return void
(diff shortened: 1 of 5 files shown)


lorenzo:
This cannot target master. I use the welcome protected method to show relevant information from my environment.

I would suggest instead that the check for empty args (or equivalent condition) to be done in the _welcome() method directly. So that anyone overriding it still gets the same behavior as before.

markstory:
Another option could be to leave an empty _welcome() method on Shell, and retain the output on the CommandList shell.

rchavik:
Implemented mark's suggestion and retargeted to 3.next.

markstory (approved):
Looks good to me

lorenzo:
Thanks @rchavik

src/Shell/CommandListShell.php
```diff
@@ -38,6 +38,22 @@ class CommandListShell extends Shell
     public $tasks = ['Command'];
 
     /**
+     * Displays a header for the shell
+     *
+     * @return void
+     */
+    protected function _welcome()
```
saeideng: how this works?

dereuromark: It overwrites the parent (empty) method.

saeideng: thanks

PR #10880, opened 2017-07-08
Sync method signature with parent class method.

ADmad:
Hopefully this should fix [this](https://travis-ci.org/cakephp/cakephp/jobs/251392632#L613) error on PHP 7.2 (nightly) without causing problems for other versions of phpunit / PHP.


ADmad:
:angry: https://travis-ci.org/cakephp/cakephp/jobs/251402498#L763

markstory:
Merging as the tests at least run now, which is good progress.

PR #10886, opened 2017-07-08
Remove root command comparisons.

markstory:
This has not worked out well. Because the cake.php file can be anywhere on the file system, and wrapped by bash, or powershell it often has a name that there is no point in validating, as we'd need to compare against argv[0] which removes any value validation could provide.

Refs #10716


lorenzo:
Good idea

PR #10887, opened 2017-07-09
Implementing an URL object

burzum:


src/Routing/Url.php
```diff
@@ -0,0 +1,178 @@
...
+ *
+ * @copyright     Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
+ * @link          https://cakephp.org CakePHP(tm) Project
+ * @since         3.5.0
+ * @license       https://opensource.org/licenses/mit-license.php MIT License
+ */
+namespace Cake\Routing;
+
+/**
+ * URL Object
```
markstory: Some context on how/when to use this object would be good given that this could be something developers interact with very often.

src/Routing/Url.php
```diff
@@ -0,0 +1,178 @@
...
+
+    /**
+     * Sets a query param
+     *
+     * @param string $key Query param name
+     * @param string|int|float $value Value
+     * @return $this
+     */
+    public function setQuery($key, $value)
+    {
```
markstory: How do custom routing parameters get set?

src/Routing/Url.php
```diff
@@ -0,0 +1,178 @@
...
+ * @link          https://cakephp.org CakePHP(tm) Project
+ * @since         3.5.0
+ * @license       https://opensource.org/licenses/mit-license.php MIT License
+ */
+namespace Cake\Routing;
+
+/**
+ * URL Object
+ */
+class Url
```
robertpustulka: Maybe this should implement `ArrayAccess`?

src/Routing/Url.php
ADmad: I guess a new method `setParam($key, $value)` would need to be added.

ADmad:
Do we need `with` prefixed immutable version of all these methods? :trollface:

src/Routing/Url.php
```diff
@@ -175,4 +181,59 @@ public function __toString()
...
+    /**
+     * Offset to unset
+     *
+     * @link http://php.net/manual/en/arrayaccess.offsetunset.php
+     * @param mixed $offset The offset to unset.
+     * @return void
+     */
+    public function offsetUnset($offset)
+    {
+        unset($offset);
```
robertpustulka: Should be  `unset($this->url[$offset]);`

src/Routing/Url.php
burzum: There is already setPass()? I could rename it?

markstory: But setPass() only sets 'passed' arguments. I'm thinking about custom routing params like `/blog/:title`. How would someone set `title`?

burzum: Ahh... I totally forgot them... 🙄

burzum: Done.

burzum:
@markstory how about setting plugin and prefix by default to null and false? This has been a common annoying thing in the past when declaring URLs in navigation items. When working in a plugin environment and you haven't set it, you automatically ended up with a link to the plugin.

dereuromark:
This change in behavior could be quite the heavy BC break for some.. While I also found it annoying I learned to live with it. And use my own convenience methods here instead.
If changed, this could break quite a few existing plugins relying on the current behavior however.

burzum:
@dereuromark how would it break if you don't explicitly use the URL object? :)

src/Routing/Url.php
```diff
@@ -0,0 +1,271 @@
...
+        return $this;
+    }
+
+    /**
+     * Sets values to be passed in the URL
+     *
+     * @param string $pass Args to be passed to a controller action
+     * @return $this
+     */
+    public function setPass($pass)
```
robertpustulka: Should this be named `addPass`? `setPass` suggests that the old value would be overwritten.

src/Routing/Url.php
```diff
@@ -0,0 +1,271 @@
...
+        return $this;
+    }
+
+    /**
+     * Absolute URL
+     *
+     * @param bool $absolute Generate an absolute URL or not
+     * @return $this
+     */
+    public function absolute($absolute = true)
```
robertpustulka: `setAbsolute` ?

dereuromark: Agreed, because we also need getters here usually.

src/Routing/Url.php
```diff
@@ -0,0 +1,272 @@
...
+        return $this;
+    }
+
+    /**
+     * Sets values to be passed in the URL
+     *
+     * @param string $pass Args to be passed to a controller action
+     * @return $this
+     */
+    public function addPass($pass)
```
ravage84: ``addPassedValues()``?

src/Routing/Url.php
```diff
@@ -0,0 +1,272 @@
...
+    {
+        foreach ($params as $key => $value) {
+            $this->setParam($key, $value);
+        }
+
+        return $this;
+    }
+
+    /**
+     * Sets a query param
```
ravage84: param => parameter. Can we generally write out such wods? I guess that wouldn't hurt.

src/Routing/Url.php
```diff
@@ -0,0 +1,272 @@
...
+     *
+     * @param string $key Query param name
+     * @param string|int|float $value Value
```
ravage84: Query parameter value.

src/Routing/Url.php
```diff
@@ -0,0 +1,272 @@
...
+            $this->url['?'] = [];
+        }
+
+        $this->url['?'] = array_merge($this->url['?'], $params);
+
+        return $this;
+    }
+
+    /**
+     * Absolute URL
```
ravage84: [Set w|W]hether to generate an absolute URL or not.

src/Routing/Url.php
```diff
@@ -0,0 +1,272 @@
...
+     */
+    public function setAbsolute($absolute = true)
+    {
+        $this->absolute = (bool)$absolute;
+
+        return $this;
+    }
+
+    /**
+     * To string
```
ravage84: Get URL as string.

src/Routing/Url.php
```diff
@@ -0,0 +1,272 @@
...
+     *
+     * @return string;
+     */
+    public function toString()
+    {
+        return Router::url($this->url, $this->absolute);
+    }
+
+    /**
+     * To array
```
ravage84: Get URL as array.

src/Routing/Url.php
```diff
@@ -0,0 +1,272 @@
...
+     *
+     * @return array
+     */
+    public function toArray()
+    {
+        return $this->url;
+    }
+
+    /**
+     * To string
```
ravage84: "To String auto convertion" or "To String Magic Method"?

src/Routing/Url.php
```diff
@@ -0,0 +1,272 @@
...
+     *
+     * @return string String URL
```
ravage84: URL as string.

src/Routing/Url.php
```diff
@@ -0,0 +1,272 @@
...
+     */
+    public function toString()
+    {
+        return Router::url($this->url, $this->absolute);
+    }
+
+    /**
+     * To array
+     *
+     * @return array
```
ravage84: URL as array.

src/Routing/Url.php
```diff
@@ -0,0 +1,272 @@
...
+    {
+        $this->absolute = (bool)$absolute;
+
+        return $this;
+    }
+
+    /**
+     * To string
+     *
+     * @return string;
```
ravage84: URL as string.

src/Routing/Url.php
```diff
@@ -0,0 +1,272 @@
...
+     */
+    public function __toString()
+    {
+        return $this->toString();
+    }
+
+    /**
+     * Whether a offset exists
+     *
+     * @link http://php.net/manual/en/arrayaccess.offsetexists.php
```
ravage84: May be after the ``@return``?

src/Routing/Url.php
```diff
@@ -0,0 +1,272 @@
...
+            return $this->url[$offset];
+        }
+
+        return null;
+    }
+
+    /**
+     * Offset to set
+     *
+     * @link http://php.net/manual/en/arrayaccess.offsetset.php
```
ravage84: May be after the ``@return``?

src/Routing/Url.php
```diff
@@ -0,0 +1,272 @@
...
+     */
+    public function offsetSet($offset, $value)
+    {
+        $this->url[$offset] = $value;
+    }
+
+    /**
+     * Offset to unset
+     *
+     * @link http://php.net/manual/en/arrayaccess.offsetunset.php
```
ravage84: May be after the ``@return``?

ravage84 (commented):
Nice! Thanks for taking care of it!

src/Routing/Url.php
inoas: What should offset be for?

src/Console/CommandRunner.php
```diff
@@ -20,6 +20,7 @@
 use Cake\Console\Exception\StopException;
 use Cake\Console\Shell;
 use Cake\Core\ConsoleApplicationInterface;
+use Cake\Core\HttpApplicationInterface;
```
burzum: No but I'll take care of it as well.

src/Routing/Url.php
burzum: Do we have a convention for that? I usually put things like that before params and returns.

src/Routing/Url.php
```diff
@@ -0,0 +1,272 @@
...
+    {
+        $this->url['action'] = $action;
+
+        return $this;
+    }
+
+    /**
+     * Sets the controller
+     *
+     * @param string $controller Controller class name without the controller suffix
```
burzum: What if you have a case like `AccountingControllerController`? ;)

inoas:
What about separating cake objects that store state into two categories:
1.) read/write objects, like say: request/response or entities
2.) builders, objects mostly used to set multiple values and then be consumed, like say: query builder or routing/url options.

Then for 1. we can use `get*()` and `set*()` and where for 2. we can use `get*()` for getting (if at all feasible) and the setters would just be the method names.

This would reduce verbosity but still allow IDE completion and strict method signatures and return types, wouldn't it? What would the drawbacks be?

burzum:
@markstory is there something left before this can be merged?

src/Routing/Url.php
```diff
@@ -0,0 +1,272 @@
...
+    }
+
+    /**
+     * Sets a query parameter
+     *
+     * @param string $key Query Parameter name
+     * @param string|int|float $value Parameter value
+     * @return $this
+     */
+    public function setQuery($key, $value)
```
ADmad: This method doesn't seem to have test coverage.

src/Routing/Url.php
ADmad: This method doesn't seem to have test coverage.

ADmad:
I kinda like @ionas' suggestion. For builders which are primarily going to be used to set vars (this one doesn't even have any getter methods) adding "set" prefix to the setters adds unnecessary verbosity.

robertpustulka:
@burzum isn't the merge with 3.next messed up here?

src/Console/CommandRunner.php
```diff
@@ -14,11 +14,7 @@
  */
 namespace Cake\Console;
 
-use Cake\Console\CommandCollection;
-use Cake\Console\CommandCollectionAwareInterface;
-use Cake\Console\ConsoleIo;
 use Cake\Console\Exception\StopException;
-use Cake\Console\Shell;
```
robertpustulka: Why are those removed?

chinpei215:
Replaced by #11036.

PR #10890, opened 2017-07-10
Added configurable default value to Configure::read()

h-moriya:
default value to `Configure::read()`

`Configure::read('key','none');` = `Configure::read('key') ?? 'none';` 

It works the same as `env ()`


dereuromark:
Sounds reasonable given the other reader methods also have this default argument.

cleptric:
What use case do you see for this feature? Do you end up missing a config option on a given case?

dereuromark:
For BC reasons this could however go into 3.5 better
As people could have extended the Configure class and call e.g. MyConfigure::read() which would then throw notices due to the signature change.
Very theoretical, though.

markstory:
Moved to `3.next`

h-moriya:
@cleptric 
I think that use it in this case

https://github.com/CakeDC/ratings/blob/31136942e4dab7563b64ad3c18b96db13aaaabae/Model/Rating.php#L41
https://github.com/multidots/cakephp-rest-api/blob/867bcc5687310cf3424a03de8c11618b0cc45b15/src/Controller/AppController.php#L68
https://github.com/uzyn/cakephp-opauth/blob/2ca4dc8f887a5ea09b2536b7d1630f4db7f123b9/Controller/OpauthAppController.php#L110

VERSION.txt
```diff
@@ -16,4 +16,4 @@
 // @license       https://opensource.org/licenses/mit-license.php MIT License
 // +--------------------------------------------------------------------------------------------+ //
 ////////////////////////////////////////////////////////////////////////////////////////////////////
-3.4.9
+3.4.10
```
ravage84: You don't need to do that. This is taken careof when it get's released.

ravage84 (commented):
Lovely, that was on my TODO list for so long. Thanks for taking care of.

VERSION.txt
dereuromark: this is because of branch change. can be disregarded.

markstory:
Thanks @h-moriya :clap:

PR #10892, opened 2017-07-10
Fix wrong variable assignment

cleptric:
This fixes an issue when using the newly added fluent route methods.
The`_method` was assigned to the wrong variable.

This PR lacks a fix for the case when no method could be gathered out of the request. That's why two test cases are failing. I wasn't sure what to do here.


markstory:
@cleptric Fixed up the tests. We were missing parameters before.

Issue #10873, opened 2017-07-06
Marshal does not detect custom date format when locale parser is enabled


micdobro:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.4.5
* Platform and Target: NGINX, MariaDB, PHP 7.0.15

### What you did
`intl.default_locale` is set to `da_DK`. 

In `bootstrap.php` I have the following: 

```php 

Type::build('time')
    ->useImmutable()
    ->useLocaleParser()
    ->setLocaleFormat('HH.mm'); 
```

That expects times in the format of: `HH.mm` - for example `03.20`. Separator etc. adheres do Danish locale as described in ICU docs. 

### What happened
When using this shortened version, i.e. without seconds (`03.20`), then the `marshal()` method in class `Cake\Database\Type\DateTimeType` returns a string instead of `FrozenTime` object, as one of the tests is raising an exception (and the whole block of code depends on it). 

It's precisely this test that raises the exception: 

```php 
 elseif (is_numeric($value)) {
    $date = new $class('@' . $value);
}
```

It happens to be executed *before* this test: 
```php
elseif (is_string($value) && $this->_useLocaleParser) {
  return $this->_parseValue($value);
}
```

Input `03.20.00` is recognized correctly, as it does not evaluate to `true` on the `is_numeric` test. 

### What you expected to happen
My expectation would be: when Cake is instructed to use a locale-aware parser, and I also say to it: expect the input of the form `HH.mm` (according to ICU constants) - then I want in return a correct time-object instance, that is then correctly serialized into the DB column of type `time`. 
At this point `03.20` is saved into db as `00:00:03`. 

I have at this point reversed the order of tests, i.e. I put the test that uses `$this->_useLocaleParser` flag in the `else if` statement *before* the `is_numeric()` test - and then everything works. I cannot see how this could fail in a broader context - hence a bug report to trigger a discussion.

markstory:
Pull request up now.

PR #10895, opened 2017-07-11
Make timestamp checks less permissive

markstory:
Some locale based formats look like numeric values. If locale parsing is enabled we need to enter that case instead of the numeric one. By restricting the valid format of timestamps we can do that.

Refs #10873

### DIFF
--- a/src/Database/Type/DateTimeType.php
+++ b/src/Database/Type/DateTimeType.php
@@ -157,11 +157,12 @@ class DateTimeType extends Type implements TypeInterface
             if ($value === '' || $value === null || $value === false || $value === true) {
                 return null;
             }
+            $isString = is_string($value);
             if (ctype_digit($value)) {
                 $date = new $class('@' . $value);
-            } elseif (is_string($value) && $this->_useLocaleParser) {
+            } elseif ($isString && $this->_useLocaleParser) {
                 return $this->_parseValue($value);
-            } elseif (is_string($value)) {
+            } elseif ($isString) {
                 $date = new $class($value);
                 $compare = true;
             }
--- a/src/Database/Type/DateTimeType.php
+++ b/src/Database/Type/DateTimeType.php
@@ -157,7 +157,7 @@ class DateTimeType extends Type implements TypeInterface
             if ($value === '' || $value === null || $value === false || $value === true) {
                 return null;
             }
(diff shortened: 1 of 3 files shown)


lorenzo:
Makes the tests sad, though :(

markstory:
I bet there is no da_DK locale in travis.

src/Database/Type/DateTimeType.php
```diff
@@ -157,7 +157,7 @@ public function marshal($value)
             if ($value === '' || $value === null || $value === false || $value === true) {
                 return null;
             }
-            if (is_numeric($value)) {
+            if (ctype_digit($value)) {
```
saeideng: `ctype_digit` is safe?
i think no
see
```
<?php
$value='0392387900';
var_dump(ctype_digit($value));
var_dump(is_numeric($value));
var_dump(new \DateTime('@' .$value));
$value=0392387900;
var_dump(ctype_digit($value));
var_dump(is_numeric($value));
var_dump(new \DateTime('@' .$value));
```

markstory: Our tests don't complain about any type handling issues, so I don't know what problem you're seeing.

saeideng: Add $value=0392387900 to test case
This value is valid but can not be accept by ctype_digit()

saeideng: `Ctype_digit()` for `$value<48` returned `false`

saeideng: Means this change is not BC

markstory: Marshalling is generally handling string data though. As people are converting POST parameters into entities.

saeideng: Generally yes but not for all time
I worried about especial usage
`$calculate=...`
`$this->Models->newEntity(['data`=>$calculate])`

saeideng: I think we should convert to string before convert to date

markstory: Ok. Do you have a pull request in mind?

saeideng: no ,but I'll try to that

PR #10897, opened 2017-07-11
Fix for overwriting with longer content.

robertpustulka:
Fix a regression introduced in #10860 where content would not be replaced with longer content.


markstory:
Thanks :tada:

PR #10899, opened 2017-07-11
Some tests refactored. Skipped non-compartible tests in MySQL ONLY_FULL_GROUP_BY mode.

bancer:


markstory:
We don't run against MySQL 5.7 in our CI environments, is this to fix test failures for you locally?

bancer:
Yes, locally and on production applications. Somebody else may be using 5.7 too.

ravage84:
@bancer thanks for taking care on the aging 2.x branch! ❤️

PR #10902, opened 2017-07-12
makes the test more stable

bancer:
Added 'order' to HABTM association in the test makes it more stable and passing in MySQL 5.7.


markstory:
Thanks!

Issue #10901, opened 2017-07-12
Incorrect WHERE clause with array of string(s)


tersmitten:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

----

* CakePHP Version: 2.9.9
* Platform and Target: PHP 7.1.6-2~ubuntu14.04.1+deb.sury.org+1, MySQL 5.6.36-82.0

----

### What you did
```php
<?php
$conditions = ['school_id' => 12501, 'role_id' => 2, 'key' => ['939', '944']];
$fields = ['user_id', 'user_id'];
$userIds = $this->User->UserEdexmlKey->find('list', compact('conditions', 'fields'));

$conditions = ['school_id' => 12501, 'role_id' => 2, 'key' => ['939']];
$fields = ['user_id', 'user_id'];
$userIds = $this->User->UserEdexmlKey->find('list', compact('conditions', 'fields'));
```

### What happened
Second `find` resulted in an exception:

```
You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'key = ('939')' at line 1
```

```sql
SELECT `UserEdexmlKey`.`user_id` FROM `example_com_app`.`user_edexml_keys` AS `UserEdexmlKey`   WHERE `school_id` = 12501 AND `role_id` = 2 AND key = ('939')
```

### What you expected to happen
Both `find`s to work

dereuromark:
You need to use IN manually for arrays:

    'key IN' => ['939', '944']

This is documented in the book.

tersmitten:
I thought so, but `'key IN' => ['939', '944']` does fix the problem in 2.x

dereuromark:
Also for single array items? `'key IN' => ['939']`

tersmitten:
Yes

dereuromark:
OK, so small bugfix, maybe you can provide a PR here to fix it for the former non IN version here.

tersmitten:
I'll give it a try

tersmitten:
https://github.com/cakephp/cakephp/blob/2.x/lib/Cake/Model/Datasource/DboSource.php#L2810 

Seems to be the problem

dereuromark:
Yeah, I don't get why the count() == 1 check has to be ;) really. At least in the current state it doesn't make much sense.

markstory:
These commits look relevant:

* https://github.com/cakephp/cakephp/commit/9e6120c86ab
[...]

PR #10903, opened 2017-07-12
WIP - Fix incorrect WHERE clause with array of string(s)

tersmitten:
See #10901

### DIFF
--- a/lib/Cake/Test/Case/Model/Datasource/Database/MysqlTest.php
+++ b/lib/Cake/Test/Case/Model/Datasource/Database/MysqlTest.php
@@ -2098,6 +2098,10 @@ SQL;
 		$expected = " WHERE `Post`.`title` = '1.1'";
 		$this->assertEquals($expected, $result);
 
+		$result = $this->Dbo->conditions(array('Post.title' => array('foo', 'bar')));
+		$expected = " WHERE `Post`.`title` IN ('foo', 'bar')";
+		$this->assertEquals($expected, $result);
+
 		$result = $this->Dbo->conditions("User.id != 0 AND User.user LIKE '%arr%'");
 		$expected = " WHERE `User`.`id` != 0 AND `User`.`user` LIKE '%arr%'";
 		$this->assertEquals($expected, $result);
@@ -2406,6 +2410,10 @@ SQL;
 		$expected = " WHERE `score` IN (1, 2, 10)";
 		$this->assertEquals($expected, $result);
 
+		$result = $this->Dbo->conditions(array('score' => array('1', '2', '10')));
+		$expected = " WHERE `score` IN ('1', '2', '10')";
+		$this->assertEquals($expected, $result);
+
 		$result = $this->Dbo->conditions(array('score' => array()));
 		$expected = " WHERE `score` IS NULL";
 		$this->assertEquals($expected, $result);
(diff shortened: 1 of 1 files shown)


saeideng:
> Fix incorrect

how tests can be to fix the problem ?

tersmitten:
> how tests can be to fix the problem ?

They can't

tersmitten:
But I'm figuring out if and how I can fix this issue :-)

lib/Cake/Test/Case/Model/Datasource/Database/MysqlTest.php
```diff
@@ -2625,10 +2633,20 @@ public function testArrayConditionsOneElement() {
 		$expected = " WHERE id = (1)";
 		$this->assertEquals($expected, $result);
 
+		$conditions = array('key' => array('939'));
+		$result = $this->Dbo->conditions($conditions);
+		$expected = " WHERE `key` IN ('939')";
```
markstory: This could also conceivably be converted into `WHERE key = '939'`. That might make the solution more complex though.

tersmitten: Do you have any idea in what cases `WHERE <X> = (<Y>)` would make sense (or is optimal)?

lorenzo: It used to matter back when we did it in ~2007, because of old mysql versions, but it is not a problem anymore.

tersmitten: But do we need the `()` around the value (e.g. are there special kind of values that require it)?

markstory: Not to my knowledge there aren't.

tersmitten:
```php
<?php
$data = $this->_quoteFields($key) . ' = (';
if ($quoteValues) {
	if ($Model !== null) {
		$columnType = $Model->getColumnType($key);
	}
	$data .= implode(', ', $this->value($value, $columnType));
}
$data .= ')';
```

tersmitten: Some questions:

* Is `$quoteValues` ever `false`?
* When does `$this->value` return an `array`?

chinpei215: @tersmitten Sorry, I didn't notice your pull request.

> Is $quoteValues ever false?

It looks like a bug. The `if` check should be removed. The `$quoteValues` argument seems to be no longer used as of [this commit](https://github.com/cakephp/cakephp/commit/39feb3f7c1675f73c15db885ef4cc571428a9520#diff-d61b91ae3172c2a361c5c376c892c195), 

> When does $this->value return an array?

For example, [this call](https://github.com/cakephp/cakephp/blob/2032fef772abcf6d827ba897e620de2b0253871a/lib/Cake/Model/Datasource/DboSource.php#L2904) returns an array. But I think it is a bad idea. No one would not expect that the method returning array. However, since it is a public method, we cannot change the behavior. Instead, having a new `DboSource::val()` not returning array would be an idea. Actually, I was planning to add this method to the core in the past.

chinpei215: Adding a new config option to DboSource class may be better.

dereuromark:
ping

This sounds like it should be merged as part of a bugfix release soon?

tersmitten:
That would be nice, but we found a reasonable workaround and there's no fix yet :-(

dereuromark:
@chinpei215 Do you have an idea for a fix here?

chinpei215:
@dereuromark I have no idea for now.

dereuromark:
Closing then for lack of interest in getting this into a maintenance release of 2.x.

PR #10909, opened 2017-07-14
Improve scoped middleware

cleptric:
This is a successor of #10837 

This changes how middleware and scopes work, as a middleware is now only applied to their closets matching scope.

```
Router::scope('/', function($routes) {
    $routes->applyMiddleware('one', 'two');
    $routes->scope('/api', function ($routes) {
        $routes->applyMiddleware('api');
        $routes->get('/ping');
    });
});
```

> In the above example, the /api/ping route would only have the 'api' middleware applied. The 'one', 'two' middleware would not be applied. This behavior allows us to remove/redefine middleware in nested scopes. Because middleware would need to be defined in nested scopes, we should also add the concept of 'middleware groups'.

Middleware groups are also added. This allows to group a bunch of **registered** middleware together and use them.
If there are any overlaps between middleware names and middleware group names, a `RuntimeException` is thrown.

```
Router::scope('/', function ($routes) {
    $routes->registerMiddleware('auth', new AuthMiddleware());
    $routes->registerMiddleware('auth.api', new AuthMiddleware());
    $routes->registerMiddleware('cookie', new CookieMiddleware());

    $routes->middlewareGroup('web', ['cookie', 'auth']);
    $routes->middlewareGroup('api', ['cookie', 'auth.api']);

    $routes->applyMiddleware('web');

    $routes->scope('/api', function ($routes) {
        // Apply the middleware in the api group. Web will not be applied.
        $routes->applyMiddleware('api');
    });
});
```


src/Routing/RouteBuilder.php
```diff
@@ -925,4 +925,18 @@ public function applyMiddleware(...$names)
 
         return $this;
     }
+
+    /**
+     * Apply a set of middleware to a group
```
markstory: Doesn't this define a new group of middleware?

src/Routing/RouteCollection.php
```diff
@@ -425,6 +432,47 @@ public function registerMiddleware($name, callable $middleware)
     }
 
     /**
+     * Add middleware to a middleware group
+     *
+     * @param string $name Name of the middleware group
+     * @param array $middlewareNames Names of the middleware
+     * @return $this
+     */
+    public function middlewareGroup($name, $middlewareNames)
```
markstory: Should this have the `array` typehint?

src/Routing/RouteCollection.php
```diff
@@ -425,6 +432,47 @@ public function registerMiddleware($name, callable $middleware)
...
+    {
+        if ($this->hasMiddleware($name)) {
+            $message = "Cannot add middleware group '$name'. A middleware by this name has already been registered.";
+            throw new RuntimeException($message);
+        }
+        if ($this->hasMiddlewareGroup($name)) {
```
markstory: Should it be possible to overwrite a middleware group? Right now once defined middleware groups are immutable.

src/Routing/RouteCollection.php
```diff
@@ -465,30 +513,39 @@ public function applyMiddleware($path, array $middleware)
...
+        $keys = array_map('strlen', array_keys($paths));
+        array_multisort($keys, SORT_DESC, $paths);
+
+        foreach ($paths as $pattern => &$middleware) {
+            $matching = [];
+
+            foreach ($middleware as $key => $name) {
+                if ($this->hasMiddlewareGroup($name)) {
+                    unset($middleware[$key]);
+                    $middleware = array_merge($middleware, $this->_middlewareGroups[$name]);
```
markstory: Won't this mutate the `$paths` variable?

lorenzo:
I'm still not convinced this the ideal way of using middleware.

In you first example I would have expected both middleware stacks to be applied.

I would say that the least surprising way of registering non overlapping middleware is by opening separate scope blocks.

Middleware applied inside the same block should always be applied.

ADmad:
>  The 'one', 'two' middleware would not be applied.

As @lorenzo said this behavior causes ambiguity. Some people might expect this behavior and some might expect them to be applied to nested scope too.

robertpustulka:
@lorenzo @ADmad maybe there should be some `$overwrite` option in `applyMiddleware()`? Or another method like `replaceMiddleware()`? So it would be possible to stack middleware (middleware groups) or replace them?

ADmad:
Adding options doesn't improve intuitivenss. Why not just follow the easiest option @lorenzo suggested, open separate scope blocks?

src/Routing/RouteCollection.php
cleptric: Yes, let's make that possible.

markstory:
@lorenzo This new approach is similar to how phoenix handles their middleware. Do you have an example of what separate scope blocks would look like? Internally if a scope is reopened multiple times, it is the same scope.

If scopes inherit middleware from their enclosing scope, how would you go about having 'auth' applied to / and a different configuration of auth applied to /api? This approach makes that trivial to do. It is also worth nothing that if an inner scope defines no middleware, the enclosing scopes middleware will be used.

lorenzo:
@markstory  This is an example of one phoenix application I made

```elixir
    scope "/friends" do
      get "/", FriendController, :index
      get "/show/:id", FriendController, :show
    end


    scope "/mates" do
      pipe_through :private

      get "/new", FriendController, :new
      put "/new", FriendController, :create
      get "/new/id", FriendController, :new_id
      post "/new/id", FriendController, :upload_id
      post "/new/terms", FriendController, :create_terms
      put "/new/terms", FriendController, :create_terms
    end
```

As you can see I opened the same scope twice, I first did it for public actions, and then for actions that require being authenticated.

markstory:
Ok. I misunderstood their documentation then. The problem that started this work and the previous pull request was:

* Scope of `/` has session based authentication applied to it.
* Scope of `/api` has token based authentication applied.

The original implementation of scoped middleware made this impossible, because middleware was inherited through the scope paths. I don't think binding middleware to URL scopes is going to be a workable model because how inheritance is expected to work.
 
Instead I think a model where RouteBuilder passes the desired middleware onto each connected route will work better. This model will enable reopened scopes to have completely different sets if middleware applied, even though they resolve to the same URL, as the middleware names are stored on each route.

I can update this pull request with that approach later today/tonight.

markstory:
@cleptric @lorenzo I've updated the code to make middleware attached to each route. This enables nested scopes to inherit middleware from their parents. Reopened scopes will maintain separate middleware stacks similiar to the phoenix example. With these changes the following a pattern for having different configurations of auth/other middleware

```php
// Assume web is a group that contains auth, cookies, and csrf.
// The api group contains auth.api, and rate limiting
$routes->scope('/', function($routes) {
  $routes->applyMiddleware('web');
  $routes->get('/login', ...);
});

$routes->scope('/api', function ($routes) {
  $routes->applyMiddleware('api');
  $routes->get('/ping', ...);
});
```

cleptric:
Works for me 🙂

lorenzo:
Great job guys!

PR #10910, opened 2017-07-14
Update cake.php

EOM:
Hi, 
Now feature in cake Console, add o set ENVIRONMENT in CLI, it's possible..?

Example use CRON limit use:
/usr/bin/php /var/www/clients/client508/web1676/dev/app/Console/cake.php get_num_rest APP_ENV=pro

### DIFF
--- a/app/Console/cake.php
+++ b/app/Console/cake.php
@@ -44,4 +44,16 @@ if (!include $dispatcher) {
 }
 unset($dispatcher);
 
+if(function_exists('putenv')){
+    // Loop args
+    foreach ($argv as $k => $v) {
+        // Is args ENVIRONMENT in CLI, FOO_ENV=BLABLA..?
+        if(stripos($v,'_ENV=')!==false){
+            // Set add ENVIRONMENT
+            putenv($v);
+	    unset($argv[$k]);
+        }
+    }
+}
+
 return ShellDispatcher::run($argv);


markstory:
Any reason you aren't using your shell's ability to set env vars? I do something like

```
FOO=stuff Console/cake do_task
```

And it works quite well.

EOM:
Hi Markstory,
The problem prevails when there are cron's that are very limited to run.
For example in ISPConfig in VPS does not allow more than just run URL or a simple command, and does not allow chaining "&&" command or use bash ".sh".
Example was the case of my problem in an ISPConfig, where I could execute just something like this.

/usr/bin/php /var/www/clients/dev/app/Console/cake.php get_num_rest APP_ENV=pro DEBUG_ENV=0

Not suport ;(
APP_ENV=pro DEBUG_ENV=0 ./Console/cake get_num_rest bla bla ....
export APP_ENV=pro && export DEBUG_ENV=0 && ./Console/cake get_num_rest bla bla...

Markstory, do not you think that's a good idea, to have something inside the code that supports that ...?
Inside the cake.php code or inside the class ShellDispatcher.

To have more flexibility in more hosting environments ..? ;)

markstory:
> For example in ISPConfig in VPS does not allow more than just run URL or a simple command, and does not allow chaining "&&" command or use bash ".sh".

In these environments you could use wrapping shell scripts. I'm reluctant to add features to the framework that duplicate features that are generally operating system features. The CLI library also has the ability to set options/arguments which can be used to change application behavior.

EOM:
Ahh very good. I did not know that I could extend and modify the behavior so I can load parameters before I load bootsrap. Because I try u can not and in the cake manual I did not find anything.

rchavik:
Closing as workaround exists.

Issue #10900, opened 2017-07-12
.po files with empty translations cause wrong output


burzum:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.4
* Platform and Target: Windows / Linux

### What you did

Generated the translations of the core strings for our app. Some of the translations are left empty in the po file.

### What happened

Ended up with empty translation strings.

I've added the po file to the core tests `tests/test_app/TestApp/Locale/en/cake.po` and ran the tests. 

You can checkout this branch to reproduce the issue: https://github.com/cakephp/cakephp/commits/i18n-translation-issue

I've already spent some time trying to find a proper fix but without success.

![image](https://user-images.githubusercontent.com/162789/28115284-840b3f3c-6704-11e7-8eaa-231d05c781e6.png)

### What you expected to happen

Falling back to the provided translation string of the __() method as default instead of ending up with empty string.

markstory:
Is this the same or different than #10774?

burzum:
@markstory it's not related to the context I think and it still fails, even with these changes.

markstory:
I'll see if I can reproduce this in the existing test suite.

markstory:
Pull request up now.

PR #10911, opened 2017-07-15
Fix plural messages not falling back to key.

markstory:
Plural messages without a translated value did not correctly fallback to the message key as singular messages do. This caused undesirable behavior when working with plural messages.

I've shifted the plural form resolution into the Translator. Now that we have a custom translator class this logic is better suited here instead of having duplicate logic in each message formatter. Interestingly our formatters had divergent logic for handling plurals which could cause some tedious to fix issues for app developers.

I've not unset the `_singular` key in the Translator as userland formatters could be relying on those keys being passed in.

Refs #10900


src/I18n/Translator.php
```diff
@@ -156,10 +140,58 @@ public function translate($key, array $tokensValues = [])
...
+                return $message['_context'][''];
+            }
+
+            return current($message['_context']);
+        }
+        if (!isset($message['_context'][$context])) {
+            return $key;
+        }
+        if (is_string($message['_context'][$context]) &&
+            strlen($message['_context'][$context]) === 0
```
lorenzo: This can be written as

   if ($message['_context'][$context] === '')

markstory: Good catch!

PR #10912, opened 2017-07-15
adapter for the getMock() depricated in phpunit

bancer:
Related to https://github.com/cakephp/cakephp/pull/10497.


bancer:
Related to https://github.com/cakephp/cakephp/pull/10497.

bancer:
There are no warnings anymore when PHPUnit 5.7.19 is used.

It would be good to merge 2.next to 2.x I suppose.

lib/Cake/TestSuite/CakeTestCase.php
```diff
@@ -716,6 +716,78 @@ protected function skipUnless($condition, $message = '') {
...
+			$callOriginalConstructor = true, $callOriginalClone = true,
+			$callAutoload = true, $cloneArguments = false,
+			$callOriginalMethods = false, $proxyTarget = null) {
+		$phpUnitVersion = PHPUnit_Runner_Version::id();
+		if ($phpUnitVersion < '5.7.0') {
+			return parent::getMock($originalClassName, $methods, $arguments,
+					$mockClassName, $callOriginalConstructor, $callOriginalClone,
+					$callAutoload, $cloneArguments, $callOriginalMethods, $proxyTarget);
+		}
+		$MockBuilder = $this->getMockBuilder($originalClassName);
```
ravage84: Add a white line before, for readability.

lib/Cake/TestSuite/CakeTestCase.php
```diff
@@ -716,6 +716,78 @@ protected function skipUnless($condition, $message = '') {
...
+		if ($cloneArguments) {
+			throw new InvalidArgumentException('$cloneArguments parameter is not supported');
+		}
+		if ($callOriginalMethods) {
+			throw new InvalidArgumentException('$callOriginalMethods parameter is not supported');
+		}
+		if ($proxyTarget !== null) {
+			throw new InvalidArgumentException('$proxyTarget parameter is not supported');
+		}
+		return $MockBuilder->getMock();
```
ravage84: Add a white line before, for readability.

ravage84 (commented):
Lovely! Is there any way, we could have a (simple) test?

lib/Cake/TestSuite/CakeTestCase.php
bancer: It is not clear how that improves readability. It makes the function longer and the user possibly will need to scroll more. Empty lines are the [code smell](http://www.yegor256.com/2014/11/03/empty-line-code-smell.html) I think.

bancer:
> Lovely! Is there any way, we could have a (simple) test?

This [inspection](https://travis-ci.org/cakephp/cakephp/jobs/254151338) runs with `PHPUNIT=5.7.19`. Is it what you meant?

lib/Cake/TestSuite/CakeTestCase.php
ravage84: I disagree with this statement. Just removing or omitting white lines which would visually group actual "parts" of the method does not remove any code smell, if there was one.

Sometimes a method does have parts, most often when the code cannot be extracted further into separate methods simply by the nature of the code.

And in case of ``getMock()`` here, we do have at least two of such parts:

1. part: Branching logic with check for version and return early.
2. part: The main mock building
3. part (potential): checking for invalid arguments and throwing exceptions, this could be done early, though I understand the current order.
4. part (arguable): the return statement (in CakePHP 3 we adpted a blank line before almost all return statements. While I personally do not agree with that in all those cases, it lead to mostly more readable code IMHO).

Disclaimer: I'm the co-developer or [PHPMD](https://phpmd.org/), that's why I care so much about messes in code and code smells.

ravage84:
@bancer yeah, I guess this coverage is fine.

lib/Cake/TestSuite/CakeTestCase.php
bancer: > An empty line, used as a separator of instructions in an object method, is a code smell. Why? In short, because a method should not contain "parts." A method should always do one thing and its functional decomposition should be done by language constructs (for example, new methods), and never by empty lines.

`part: The main mock building` can be easily extracted to another method. 

CakePHP 3 uses a different code style than CakePHP 2. Empty lines are not mentioned in https://book.cakephp.org/2.0/en/contributing/cakephp-coding-conventions.html but usually there are no empty lines above return statements in CakePHP 2 code.

lib/Cake/TestSuite/CakeTestCase.php
```diff
@@ -716,6 +716,112 @@ protected function skipUnless($condition, $message = '') {
...
+ * @param string $mockClassName The fourth (optional) parameter can be used to
+ *   specify a class name for the generated test double class.
+ * @param bool $callOriginalConstructor The fifth (optional) parameter can be
+ *   used to disable the call to the original class' constructor.
+ * @param bool $callOriginalClone The sixth (optional) parameter can be used
+ *   to disable the call to the original class' clone constructor.
+ * @param bool $callAutoload The seventh (optional) parameter can be used to
+ *   disable __autoload() during the generation of the test double class.
+ * @return object
+ * @deprecated Use `getMockBuilder()` or `createMock()` in new unit tests.
```
bancer: I am a bit unsure whether this method really should be deprecated. It does not call any deprecated methods but has too many parameters that is a bad design that comes from `getMock`.

markstory: With only one caller could this method be inlined there?

bancer: Sorry, I do not understand the question. Could you rephrase it?

markstory:
All tests builds are failing in travis. Looks like it is related to these changes as well.

app/Config/bootstrap.php
```diff
@@ -22,6 +22,17 @@
  * @license       https://opensource.org/licenses/mit-license.php MIT License
  */
 
+/**
+ * Load Composer autoload.
+ *
+ * Remove and re-prepend CakePHP's autoloader as Composer thinks it is the most important.
+ * @see http://goo.gl/kKVJO7
```
saeideng: https://github.com/composer/composer/commit/c80cb76

bancer: @saeideng Do you want to say that this should be fixed in composer?

saeideng: i say use `https://github.com/composer/composer/commit/c80cb76` instead of short/indirect url `http://goo.gl/kKVJO7`

bancer:
I am a bit confused why the builds fail. Is there caching problem?

markstory:
I think travis has moved our build to trusty which doesn't offer php 5.3. I am working on getting builds fixed for 2.x and will merge this after that.

tersmitten:
@bancer @markstory Do you have an example of a working `phpunit.xml`?

This is what I have now in my `app` directory. However this does not allow code coverage reports to be generated (as `blacklist` is no longer supported)

```
<phpunit colors="false">
	<filter>
		<blacklist>
			<directory suffix=".php">Plugin</directory>
			<directory suffix=".php">Test</directory>
			<directory suffix=".php">Vendor</directory>
		</blacklist>
	</filter>
</phpunit>
```

When I try to change my config to use `whitelist`, our build jobs crash.

```
 <phpunit colors="false">
 	<filter>
		<whitelist processUncoveredFilesFromWhitelist="true">
			<directory suffix=".php">./</directory>
			<exclude>
				<directory suffix=".php">./Plugin</directory>
				<directory suffix=".php">./Test</directory>
				<directory suffix=".php">./Vendor</directory>
			</exclude>
		</whitelist>
 	</filter>
 </phpunit>
```

tersmitten:
```xml
<phpunit colors="false">
	<filter>
		<whitelist>
			<directory suffix=".php">./</directory>
			<exclude>
				<directory suffix=".php">./Config/Schema</directory>
				<directory suffix=".php">./Plugin</directory>
				<directory suffix=".php">./Test</directory>
				<directory suffix=".php">./Vendor</directory>
				<directory suffix=".php">./webroot</directory>
				<file>./Config/acl.ini.php</file>
				<file>./Config/acl.php</file>
				<file>./index.php</file>
			</exclude>
		</whitelist>
	</filter>
</phpunit>
```

Seems to work stable

PR #10922, opened 2017-07-18
Add avg and median methods to collection

### LINKED ISSUE #10920: Add avg and median methods to collection
* [ ] bug
* [x] enhancement

* CakePHP Version: 3.5.0

### What you did
Add methods `avg` and `median` to `Collection`.

#### Avg
```php
$items = [1, 2, 3];
$collection = new Collection($items);
$this->assertEquals(2, $collection->avg());

$items = [['foo' => 1], ['foo' => 2], ['foo' => 3]];
$collection = new Collection($items);
$this->assertEquals(2, $collection->avg('foo'));
```
#### Median
```php
$items = [5, 2, 4];
$collection = new Collection($items);
$this->assertEquals(4, $collection->median());

$items = [1, 2, 3, 4];
$collection = new Collection($items);
$this->assertEquals(2.5, $collection->median());

$items = [
    ['invoice' => ['total' => 400]],
    ['invoice' => ['total' => 500]],
    ['invoice' => ['total' => 200]],
    ['invoice' => ['total' => 100]],
    ['invoice' => ['total' => 333]]
];
$this->assertEquals(333, (new Collection($items))->median('invoice.total'));
```

Iandenh:
Resolves #10920

### DIFF
--- a/src/Collection/CollectionTrait.php
+++ b/src/Collection/CollectionTrait.php
@@ -226,7 +226,7 @@ trait CollectionTrait
         if ($matcher != null) {
             $elements = $elements->extract($matcher);
         }
-        $values = $elements->toArray();
+        $values = $elements->toList();
         sort($values);
 
         if ($count % 2) {
--- a/src/Collection/CollectionTrait.php
+++ b/src/Collection/CollectionTrait.php
@@ -193,6 +193,54 @@ trait CollectionTrait
     /**
      * {@inheritDoc}
      */
+    public function avg($matcher = null)
+    {
+        $iterator = $this->unwrap();
+        $count = $iterator instanceof Countable ?
+            count($iterator) :
+            iterator_count($iterator);
+
+        if ($count === 0) {
+            return null;
+        }
+
+        return $this->sumOf($matcher) / $count;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function median($matcher = null)
+    {
+        $iterator = $this->unwrap();
+        $count = $iterator instanceof Countable ?
+            count($iterator) :
+            iterator_count($iterator);
+
+        if ($count === 0) {
+            return null;
+        }
+
+        $middle = (int)($count / 2);
+        $elements = $this;
+        if ($matcher != null) {
+            $elements = $elements->extract($matcher);
+        }
+        $values = $elements->toArray();
+        sort($values);
+
+        if ($count % 2) {
+            return $values[$middle];
+        }
+
+        return (new static([
+            $values[$middle - 1], $values[$middle],
+        ]))->avg();
+    }
+
+    /**
+     * {@inheritDoc}
+     */
     public function sortBy($callback, $dir = SORT_DESC, $type = SORT_NUMERIC)
     {
         return new SortIterator($this->unwrap(), $callback, $dir, $type);
--- a/src/Collection/CollectionTrait.php
+++ b/src/Collection/CollectionTrait.php
@@ -202,7 +202,7 @@ trait CollectionTrait
         $result = $result
             ->reduce(function ($acc, $current) {
                 list($count, $sum) = $acc;
-                
+
                 return [$count + 1, $sum + $current];
             }, [0, 0]);
 
--- a/src/Collection/CollectionTrait.php
+++ b/src/Collection/CollectionTrait.php
@@ -195,16 +195,22 @@ trait CollectionTrait
      */
     public function avg($matcher = null)
     {
-        $iterator = $this->unwrap();
-        $count = $iterator instanceof Countable ?
-            count($iterator) :
-            iterator_count($iterator);
+        $result = $this;
+        if ($matcher != null) {
+            $result = $result->extract($matcher);
+        }
+        $result = $result
+            ->reduce(function ($acc, $current) {
+                list($count, $sum) = $acc;
+                
+                return [$count + 1, $sum + $current];
+            }, [0, 0]);
 
-        if ($count === 0) {
+        if ($result[0] === 0) {
             return null;
         }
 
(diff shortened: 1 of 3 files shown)


src/Collection/CollectionTrait.php
```diff
@@ -193,6 +193,54 @@ public function min($callback, $type = SORT_NUMERIC)
...
+        if ($count === 0) {
+            return null;
+        }
+
+        $middle = (int)($count / 2);
+        $elements = $this;
+        if ($matcher != null) {
+            $elements = $elements->extract($matcher);
+        }
+        $values = $elements->toArray();
```
markstory: You should use `toList()` here as the extracted data could have non-unique keys.

src/Collection/CollectionTrait.php
```diff
@@ -193,6 +193,54 @@ public function min($callback, $type = SORT_NUMERIC)
...
+        $iterator = $this->unwrap();
+        $count = $iterator instanceof Countable ?
+            count($iterator) :
+            iterator_count($iterator);
+
+        if ($count === 0) {
+            return null;
+        }
+
+        return $this->sumOf($matcher) / $count;
```
lorenzo: I would say that the only advantage of having this method in the core is so that really large iterators (and specially generators) can be used without consuming all the memory or throwing an exception.

This implementation will, for example, throw an exception if the internal iterator in the collection is a generators (a function calling `yield`)

a better implementation would calculate the result on the fly as it iterates on the result. This is one version:

```php
$result =
$this
    ->extract($matcher)
     ->reduce(function($acc, $current) {
         list($count, $sum) = $acc;
         return [$count + 1, $sum + $current]
    }, [0, 0]);
return $result[1] / $result[0];
```

src/Collection/CollectionTrait.php
```diff
@@ -193,6 +193,54 @@ public function min($callback, $type = SORT_NUMERIC)
...
+            return null;
+        }
+
+        $middle = (int)($count / 2);
+        $elements = $this;
+        if ($matcher != null) {
+            $elements = $elements->extract($matcher);
+        }
+        $values = $elements->toList();
+        sort($values);
```
lorenzo: Better to do the count after sorting, so you would be counting array elements instead of an iterator, which is faster, and also will prevent throwing exceptions when using generators.

tests/TestCase/Collection/CollectionTest.php
```diff
@@ -60,6 +60,60 @@ public function testArrayIsWrapped()
     }
 
     /**
+     * Tests that it is possible to convert an array into a collection
+     *
+     * @return void
+     */
+    public function testAvg()
+    {
+        $items = [1, 2, 3];
```
lorenzo: Take a look at the other tests in this file. Use a `@dataProvider` so you can automatically test your function when both passing an array and passing a generator.

lorenzo (changes requested):

lorenzo (commented):
Great job, this is looking great!

One final thing, can you add a couple tests to test that te new methods return null when the collection is empty?

Iandenh:
Thanks

Not sure what you mean with empty collection test. I have already added these test https://github.com/cakephp/cakephp/pull/10922/files#diff-66f2154ab9fca71911f66b7706f66fc7R88 both for `avg()` and `median()`. If there are more tests needed, I'm happy to add them.

robertpustulka:
Can we add new methods to the interface and still keep BC?

src/Collection/CollectionTrait.php
```diff
@@ -193,6 +193,57 @@ public function min($callback, $type = SORT_NUMERIC)
...
+
+        $middle = (int)($count / 2);
+
+        if ($count % 2) {
+            return $values[$middle];
+        }
+
+        return (new static([
+            $values[$middle - 1], $values[$middle],
+        ]))->avg();
```
robertpustulka: Wouldn't it be more simple to do just `return ($values[$middle - 1] + $values[$middle]) / 2;` ?

lorenzo:
I'm inclined to say that we should add it to the interface, what are your thoughts about it @markstory ?

lorenzo (approved):

robertpustulka:
I'm just asking. I think that's a very low possibility that somebody implemented the interface on their own instead of using the trait.

tests/TestCase/Collection/CollectionTest.php
```diff
@@ -85,15 +85,23 @@ public function testAvg($items)
...
 
     /**
+     * Tests the avg method when on an empty collection
+     *
+     * @return void
+     */
+    public function testAvgWithEmptyCollection()
+    {
+        $collection = new Collection([]);
+        $this->assertNull($collection->avg());
```
markstory: Should this be `0`?

Iandenh: According to https://math.stackexchange.com/questions/909395/what-is-the-average-of-no-numbers the average without numbers doesn't exist.

But I can add it.

robertpustulka: How other implementations handle those cases?

markstory: Null it is then 😁

robertpustulka (commented):
Should we have a check for elements whether they are numeric?

markstory:
Won't the various math operations fail giving the developer feedback?

robertpustulka:
@markstory 

In PHP 5.6:
```php
$a = 'foo' + 'bar';
echo $e; //gives 0, no warning
```

In PHP 7.1:
```php
$a = 'foo' + 'bar';
PHP Warning:  A non-numeric value encountered in php shell code on line 1
```

Issue #10874, opened 2017-07-07
SQLite mode or mask


Erwane:
This is a (multiple allowed):

* [ ] bug
* [x] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.4.x
* Platform and Target: All.

### What you did
using SQLite databases

### What happened
DebugKit SQLite file is in 0644 mode and can't be writed by my group

### What you expected to happen
Bases PHP function and method (sqlite_open or SQLite3::_construct()) have a "mode" option to fix the database file mode

PDO can't do that.

I propose a new option in ConnectionManager config for "SQLite" called "mask".
In Cake\Database\Driver\Sqlite::connect(), test if file exist first, if not, flag it and chmod the created file with mask if $options['mask'] is set

markstory:
Sounds like a great idea. Any chance you'd be willing to open a pull request with the changes you think should be done?

Erwane:
i'll try to do that, i already have another issue to PR too ;)

Erwane:
Starting dev

markstory:
Closing as a pull request is open now.

PR #10924, opened 2017-07-19
10874 sqlite mask

Erwane:
issue #10874

new "mask" option for sqlite database driver.

Driver\Sqlite chmod the database file with mask on creation only.


markstory:
Any way we could get a test or the doc blocks updated?

Erwane:
```php
chmod($config['database'], $config['mask']);
```
this line fail because Mock don't create the db file.
I don't know how to fix that in test :(

use @chmod to avoid message ?

markstory (commented):
Looking good

PR #10925, opened 2017-07-19
Fix regression with arrays being passed for IO output.

dereuromark:
I had an app that still had the syntax

```php
$this->Shell->err(['Just', 'a', 'test']);
```

which now breaks due to the markup added.
This fix resolves this regression and restores the untested but documented feature of being able to pass arrays instead of a string as convenience wrapper input.

### DIFF
--- a/tests/TestCase/Console/ShellTest.php
+++ b/tests/TestCase/Console/ShellTest.php
@@ -310,6 +310,20 @@ class ShellTest extends TestCase
     }
 
     /**
+     * testErr method with array
+     *
+     * @return void
+     */
+    public function testErrArray()
+    {
+        $this->io->expects($this->once())
+            ->method('err')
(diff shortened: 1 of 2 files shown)


tests/TestCase/Console/ShellTest.php
```diff
@@ -310,6 +310,20 @@ public function testErr()
     }
 
     /**
+     * testErr method with array
+     *
+     * @return void
+     */
+    public function testErrArray()
```
markstory: Should we have tests for the other methods as well?

dereuromark: We could. Not sure if required, though.

markstory: We didn't have them before and there was a regression.

dereuromark: Done :)

PR #10926, opened 2017-07-19
Throw a 504 when upstream request times out

ogrrd:
Rather than throwing a more general Exception, using HttpException to throw a 504 seems to fit the description well for a 504: "The server was acting as a gateway or proxy and did not receive a timely response from the upstream server."


markstory:
Seems reasonable. Old catch blocks will continue working because the new exception is a subclass of the old one.

PR #10927, opened 2017-07-19
3.next add new paramater to testing helper

viraj-khatavkar:
Currently, the `assertResponseContains` method accepts only two parameters for assertion against the content and message.

There have been some use-cases while testing where I needed to ignore the case for asserting the content on pages where the content has been capitalized by CSS. `PHPUnit` gives support for ignoring the content case with a `ignoreCase` parameter. I have added support for the same in this test method.

### DIFF
--- a/src/TestSuite/IntegrationTestCase.php
+++ b/src/TestSuite/IntegrationTestCase.php
@@ -913,14 +913,15 @@ abstract class IntegrationTestCase extends TestCase
      *
      * @param string $content The content to check for.
      * @param string $message The failure message that will be appended to the generated message.
+     * @param bool   $ignoreCase A flag to check whether we should ignore case or not.
      * @return void
      */
-    public function assertResponseContains($content, $message = '')
+    public function assertResponseContains($content, $ignoreCase = false, $message = '')
     {
         if (!$this->_response) {
             $this->fail('No response set, cannot assert content. ' . $message);
         }
-        $this->assertContains($content, $this->_getBodyAsString(), $message);
+        $this->assertContains($content, $this->_getBodyAsString(), $message, $ignoreCase);
     }
 
     /**
--- a/src/TestSuite/IntegrationTestCase.php
+++ b/src/TestSuite/IntegrationTestCase.php
@@ -916,7 +916,7 @@ abstract class IntegrationTestCase extends TestCase
      * @param bool   $ignoreCase A flag to check whether we should ignore case or not.
      * @return void
      */
-    public function assertResponseContains($content, $ignoreCase = false, $message = '')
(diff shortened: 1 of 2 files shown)


src/TestSuite/IntegrationTestCase.php
```diff
@@ -913,14 +913,15 @@ public function assertResponseEquals($content, $message = '')
      *
      * @param string $content The content to check for.
      * @param string $message The failure message that will be appended to the generated message.
+     * @param bool   $ignoreCase A flag to check whether we should ignore case or not.
      * @return void
      */
-    public function assertResponseContains($content, $message = '')
+    public function assertResponseContains($content, $ignoreCase = false, $message = '')
     {
```
markstory: This isn't backwards compatible right now. You need to make ignoreCase the last parameter.

markstory (changes requested):

src/TestSuite/IntegrationTestCase.php
```diff
@@ -882,14 +913,15 @@ public function assertResponseEquals($content, $message = '')
      *
      * @param string $content The content to check for.
      * @param string $message The failure message that will be appended to the generated message.
+     * @param bool   $ignoreCase A flag to check whether we should ignore case or not.
```
viraj-khatavkar: It matches, is it because of the space?

src/TestSuite/IntegrationTestCase.php
viraj-khatavkar: Then I would have to always specify the message parameter also. But I agree that we need it to be backward compatible. I'll amend it.

markstory: Another option would be to parameter sniff based on types, but that gets awkward if more parameters are added later.

viraj-khatavkar: Let's keep it as the last parameter for now. Would I propose making another pull request with `ignoreCase` as the second parameter for the 4.0 branch?

src/TestSuite/IntegrationTestCase.php
HavokInspiration: @viraj-khatavkar I think so. We never align docblocks parameters as they can become quickly out of sync.

viraj-khatavkar:
@markstory Do I need to make any other amends, or is this fine?

src/TestSuite/IntegrationTestCase.php
viraj-khatavkar: @HavokInspiration Fair enough!

saeideng:
tests on php 7.0 failed

markstory (approved):

PR #10929, opened 2017-07-20
Support non-callable HttpApplicationInterface implementations.

robertpustulka:
Custom Application classes are not required to implement `__invoke()` method
so they shouldn't be treated as a middleware in such case.

### DIFF
--- a/tests/test_app/TestApp/Http/CustomApplication.php
+++ b/tests/test_app/TestApp/Http/CustomApplication.php
@@ -1,0 +1,21 @@
+<?php
+namespace TestApp\Http;
+
+use Cake\Core\HttpApplicationInterface;
+
+class CustomApplication implements HttpApplicationInterface
+{
+
+    public function bootstrap()
+    {
+    }
+
+    public function middleware($middleware)
+    {
+        return $middleware;
+    }
+
+    public function routes($routes)
+    {
+    }
+}
--- a/src/Http/Server.php
+++ b/src/Http/Server.php
@@ -77,7 +77,9 @@ class Server
             throw new RuntimeException('The application `middleware` method did not return a middleware queue.');
         }
         $this->dispatchEvent('Server.buildMiddleware', ['middleware' => $middleware]);
-        $middleware->add($this->app);
+        if (is_callable($this->app)) {
+            $middleware->add($this->app);
+        }
         $response = $this->runner->run($middleware, $request, $response);
 
         if (!($response instanceof ResponseInterface)) {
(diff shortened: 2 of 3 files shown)


tests/test_app/TestApp/Http/CustomApplication.php
```diff
@@ -0,0 +1,32 @@
...
+ */
+
+namespace TestApp\Http;
+
+use Cake\Core\HttpApplicationInterface;
+
+/**
+ * Description of CustomApplication
+ *
+ * @author Robert Pustułka <r.pustulka@robotusers.com>
```
saeideng: ??

tests/test_app/TestApp/Http/CustomApplication.php
```diff
@@ -0,0 +1,32 @@
+<?php
+/*
+ * To change this license header, choose License Headers in Project Properties.
+ * To change this template file, choose Tools | Templates
+ * and open the template in the editor.
+ */
```
saeideng: what is these??

tests/test_app/TestApp/Http/CustomApplication.php
ADmad: This needs to be removed and the license docblock at top needs to be updated.

saeideng (commented):
what you need to `non-callable` ?

tests/test_app/TestApp/Http/CustomApplication.php
robertpustulka: IDE :/

robertpustulka:
@saeideng 
`Error: Function name must be a string` error would be triggered if someone used their own implementation of `HttpApplicationInterface`.

And I don't think we need to force people to implement their custom app classes as a middleware.

markstory:
> And I don't think we need to force people to implement their custom app classes as a middleware.

If the app is not middleware no controllers will be invoked. Being able to enter the controllers is the why the Application is invokable.

robertpustulka:
@markstory what about controller-less apps? Like micro-framework implementations.

markstory:
@robertpustulka I don't think it is possible to make those just yet, at least we don't make it easy.

robertpustulka:
@markstory you think we should add `__invoke($request, $response, $next)` to the `HttpApplicationInterface` instead?

markstory:
@robertpustulka I think so. At least that will make all the methods consumers rely on part of the interface.

robertpustulka:
Closing in favor of #10945

PR #10931, opened 2017-07-21
Added the wrapper for creating shell instances.

robertpustulka:
Make it easier to extend this class for DIC support.

### DIFF
--- a/src/Console/CommandRunner.php
+++ b/src/Console/CommandRunner.php
@@ -155,7 +155,7 @@ class CommandRunner
     /**
      * Get the shell instance for a given command name
      *
-     * @param \Cake\Console\ConsoleIo $io The io wrapper for the created shell class.
+     * @param \Cake\Console\ConsoleIo $io The IO wrapper for the created shell class.
      * @param \Cake\Console\CommandCollection $commands The command collection to find the shell in.
      * @param string $name The command name to find
      * @return \Cake\Console\Shell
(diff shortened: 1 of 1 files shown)


dereuromark (approved):

src/Console/CommandRunner.php
```diff
@@ -182,4 +182,16 @@ protected function getShell(ConsoleIo $io, CommandCollection $commands, $name)
 
         return $instance;
     }
+
+    /**
+     * The wrapper for creating shell instances.
+     *
+     * @param string $className Shell class name.
+     * @param \Cake\Console\ConsoleIo $io The io wrapper for the created shell class.
+     * @return \Cake\Console\Shell
```
dereuromark: Would be nice to some day have more interfaces for these sort of things :)

src/Console/CommandRunner.php
```diff
@@ -182,4 +182,16 @@ protected function getShell(ConsoleIo $io, CommandCollection $commands, $name)
 
         return $instance;
     }
+
+    /**
+     * The wrapper for creating shell instances.
+     *
+     * @param string $className Shell class name.
+     * @param \Cake\Console\ConsoleIo $io The io wrapper for the created shell class.
```
dereuromark: io => IO maybe

src/Console/CommandRunner.php
dereuromark: Not sure if I like that, though. Inheritance of some base logic can be useful. But out of scope of this discussion here now :)

PR #10932, opened 2017-07-21
Added optional ActionDispatcher dependency to BaseApplication.

robertpustulka:
Make it easier to inject custom `ActionDispatcher` and `ControllerFactory`.

### DIFF
--- a/src/Http/BaseApplication.php
+++ b/src/Http/BaseApplication.php
@@ -16,6 +16,7 @@ namespace Cake\Http;
 
 use Cake\Core\ConsoleApplicationInterface;
 use Cake\Core\HttpApplicationInterface;
+use Cake\Http\ActionDispatcher;
 use Cake\Routing\DispatcherFactory;
 use Psr\Http\Message\ResponseInterface;
 use Psr\Http\Message\ServerRequestInterface;
@@ -36,13 +37,20 @@ abstract class BaseApplication implements ConsoleApplicationInterface, HttpAppli
     protected $configDir;
 
     /**
+     * @var \Cake\Http\ActionDispatcher|null
+     */
+    protected $dispatcher;
+
+    /**
      * Constructor
      *
      * @param string $configDir The directory the bootstrap configuration is held in.
+     * @param \Cake\Http\ActionDispatcher|null $dispatcher Optional action dispatcher.
      */
-    public function __construct($configDir)
+    public function __construct($configDir, ActionDispatcher $dispatcher = null)
     {
         $this->configDir = $configDir;
+        $this->dispatcher = $dispatcher;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


src/Http/BaseApplication.php
```diff
@@ -36,13 +37,20 @@
...
     /**
+     * @var \Cake\Http\ActionDispatcher
+     */
+    protected $dispatcher;
+
+    /**
      * Constructor
      *
      * @param string $configDir The directory the bootstrap configuration is held in.
+     * @param \Cake\Http\ActionDispatcher $dispatcher Optional action dispatcher.
```
dereuromark: |null missing

src/Http/BaseApplication.php
```diff
@@ -36,13 +37,20 @@
     protected $configDir;
 
     /**
+     * @var \Cake\Http\ActionDispatcher
```
dereuromark: Same here apparently

src/Http/BaseApplication.php
```diff
@@ -110,6 +118,10 @@ public function __invoke(ServerRequestInterface $request, ResponseInterface $res
      */
     protected function getDispatcher()
     {
-        return new ActionDispatcher(null, null, DispatcherFactory::filters());
+        if ($this->dispatcher === null) {
```
markstory: Shouldn't this be done in the constructor? Otherwise the Application is 'invalid' until this method is called to create a side-effect.

robertpustulka: This method is called while the class is being invoked. The dispatcher wasn't created in a constructor before. What's different now?

markstory: Previously the Application retained no state. Each call to getDispatcher() shared no state with other calls. Now userland code could use the added property and hit run-time errrors.

robertpustulka: Fair point.

robertpustulka: @markstory 

On the other thought, I don't think we can instantiate `ActionDispatcher` in a constructor because of the `DispatcherFactory::filters()` dependency. FIlters are usually added in a `bootstrap.php` file so after the application object has been created, am I correct?

markstory: Yes. Most plugins/apps add their dispatch filters in the bootstrap process, which is after the application is created.

How would injecting the dependency work if building in the constructor is not possible?

robertpustulka: @markstory in that case it wouldn't if we wanted to support dispatcher filters while injecting a dispatcher on app creation.

I'll hold with this one and try a different middleware-based approach as discussed on slack :)

robertpustulka:
I need to reconsider my approach here. Closing this for now.

PR #10933, opened 2017-07-21
Deprecate TableRegistry static API.

robertpustulka:
There were some complaints about CakePHP relying too much on static classes and `TableRegistry` was one of the examples. Actually `TableRegistry` is just a static API wrapper for `TableLocator`.

The best practice is to use `LocatorAwareTrait` in classes that need access to the table "registry".

For example:
```php
//deprecated call
$users = TableRegistry::get('Users');

//use table locator instead
use \Cake\ORM\Locator\LocatorAwareTrait;
...
$locator = $this->getTableLocator();
$users = $locator->get('Users');

//or in global context:
$locator = TableRegistry::getTableLocator();
$users = $locator->get('Users');
```

`TableRegistry::getTableLocator()` needs to stay for a while so the apps relying on a global locator wouldn't break.

Nothing changes for classes that use `ModelAwareTrait` (controllers, shells, etc):
```php
$this->loadModel('Users');
$this->Users->get(1);
```


lorenzo:
Don't we end up with the same implementation by making `TableLocator` a singleton again?

robertpustulka:
`TableLocator` has always been a singleton by default.
You can swap locator instances locally though.

lorenzo:
well, it is nice that we are getting rid of the unnecessary indirection

saeideng:
I am ok by change way
But not by this
:-1: 
I think we can use shorter naming 
This is very long and not good
Please open issue before implement it

robertpustulka:
@saeideng We can't change the naming without breaking the BC. Table locator related code has been around for a while and is actually the preferred way of locating tables (contrary to calling `TableRegistry` statically).

dereuromark:
@saeideng It is pretty much the same name length, there really is no overhead here either way.

ADmad:
@saeideng At least until 4.0 can keep using `$this->tableLocator()`. After that we'll just have to bite the bullet and get used to typing extra "get", "set" for lot of methods. :slightly_smiling_face:

saeideng:
I looking for very simple way like
`$this->getTable('post')`

robertpustulka:
You can create your own trait with a wrapper method.

I'm afraid we can't add getTable() to LocatorAwareTrait since it's a very common method name and could break userland code.

saeideng:
Or
`Cake\orm\table:get('posts')`

robertpustulka:
@saeideng Static calls is what we're trying to avoid here.

saeideng:
I class name is for v4 :smiley:

src/Console/Shell.php
```diff
@@ -183,7 +183,7 @@ public function __construct(ConsoleIo $io = null)
         }
         $this->_io = $io ?: new ConsoleIo();
 
-        $locator = $this->getTableLocator() ? : 'Cake\ORM\TableRegistry';
+        $locator = $this->getTableLocator() ? : TableLocator::getInstance();
```
AlaaAttya: why not using empty statements instead use this

```
$locator = !$this->getTableLocator() ?? TableLocator::getInstance();
```

inoas:
> @saeideng We can't change the naming without breaking the BC. Table locator related code has been around for a while and is actually the preferred way of locating tables (contrary to calling TableRegistry statically).

We can actually with class alias and deprecation.
Why not `getTable()`? within the trait?

I don't think there should be much of a Problem. If you extended core classes in the past you would continue using `TableRegistry::get()`.

markstory:
I confused about why we would add a new trait. What does it do better than ModelAwareTrait does? It appears to do very similar work.

src/Console/Shell.php
markstory: We can't use ?? because of php5 compatibility.

src/ORM/Locator/TableLocator.php
```diff
@@ -299,4 +306,29 @@ public function remove($alias)
             $this->_fallbacked[$alias]
         );
     }
+
+    /**
+     * Returns a singleton instance of LocatorInterface implementation.
+     *
+     * @return \Cake\ORM\Locator\LocatorInterface
+     */
+    public static function getInstance()
```
markstory: The singleton doesn't remove static class usage it just moves it. This class would now have a mixture of static and non-static state, which doesn't seem great.

robertpustulka: This PR aims to reduce the number of static calls in userland code, to discourage from using `TableRegistry` and move to the underlying implementation of table locators.

The static call is done only once when the `getTableLocator()` is called from an objecy using the trait.

markstory: We have ModelAwareTrait already though. How is this better than that?

robertpustulka: Note that I haven't added any new trait. Just reused the existing one.

`ModelAwareTrait` is used to work with *models* in general while there were two ways to work with ORM while developing plugins for example. Using static classes shouldn't be the preferred one.

robertpustulka:
@inoas, @markstory 

Note that `LocatorAwareTrait` is not a new trait. It has been in the core since 3.1.0 and used throughout the core classes as a replacement for `TableRegistry` "static" class. I think we should have deprecated static calls back then though. This PR does just this.

We also can't change trait method definitions here without breaking the BC. Adding a new method with such a common name as `getTable()` to the trait that is used in base `Controller` and `Shell` classes can cause problems with existing applications.

robertpustulka:
Is there something left to be done here?

alexsmonte:
One of the great advantages of cakephp is precisely its easy writing, this lowers the learning curve.
I understand the question of static methods however it is necessary to have something simple, as it always was in cakephp.

robertpustulka:
Is calling `TableRegistry::get('Foo')` really much simplier than `$this->getTableLocator()->get('Foo')`?

saeideng:
Yes
Because you need to call trait namespace

robertpustulka:
@saeideng don't you need to use `Cake\ORM\TableRegistry` on top of your class as well?

saeideng:
see usage 
1 )linear
//TableRegistry
```
class A{
\Cake\ORM\TableRegistry:get()
}
```
//LocatorAwareTrait
```
class A{
use \Cake\ORM\Locator\LocatorAwareTrait;
$this->getTableLocator()->get('Foo');
}
```
2 )
//TableRegistry
```
use Cake\ORM\TableRegistry;
class A{
TableRegistry:get()
}
```
//LocatorAwareTrait
```
use Cake\ORM\Locator\LocatorAwareTrait;
class A{
use LocatorAwareTrait;
$this->getTableLocator()->get('Foo');
}

```

robertpustulka:
@saeideng looks like for the cost of 1 line you've saved your code from a hard dependency and a global :smiley:

inoas:
Well the point stands nevertheless. The question is how bad that dependency is when you can easily(!) just use another use statement to swap out `TableRegistry`.

dereuromark:
This looks good.
Whoever wants can still use static access as outlined from here on :)
So this is for now mainly core relevant and helps to overcome the CakePHP stigma of too much static calls around locators and factory access.

robertpustulka:
@inoas to change the dependency of a static class you need to change your class. Thanks to the trait you can swap dependencies using a setter externally. Or mock it easily as well.

robertpustulka:
I'm moving it to the 3.6.0

dereuromark:
I just made TableRegistry::get() calls [IDE aware](https://github.com/dereuromark/cakephp-ide-helper/blob/master/docs/Generator.md). Which works quite well on project level for now at least.
I wonder how this would look like with the dynamic approach here. Any ideas to get the right annotated object typehinted here?

robertpustulka:
I don't use Phpstorm, so I'm not sure how it would work. But since 3.1 `TableRegistry` is just a static wrapper for a `TableLocator`. Probably would work the same. The API for `TableLocator` is similar, just no static methods.

ndm2:
@dereuromark  Given that in PHP dynamic and static methods with the same name cannot exist, dynamic methods are annotated just like static methods:

```php
<?php
namespace PHPSTORM_META;

override(
    \Cake\Datasource\ModelAwareTrait::loadModel(0),
    map([
        "Alias" => \App\Model\Table\ClassName::class,
        //...
    ])
);

override(
    \Cake\ORM\Locator\LocatorInterface::get(0),
    map([
        "Alias" => \App\Model\Table\ClassName::class,
        //...
    ])
);

override(
    \Cake\ORM\TableRegistry::get(0),
    map([
        "Alias" => \App\Model\Table\ClassName::class,
        //...
    ])
);
```

dereuromark:
So are we good to continue and merge? All issues resolved?

robertpustulka:
I don't see any issues. The locators have been around since 3.1.0 and TableRegistry should have been deprecated back then IMO.

saeideng:
No :)

inoas:
Again for internal use this is all great.

For userland code I like concise. If all you do in 99% of all cases is getting a table object from via getTableLocator->get then why does the developer has to go through all boilerplate?

Similar to https://github.com/cakephp/cakephp/blob/master/src/Datasource/ModelAwareTrait.php
We can have TableAwareTrait wich would work almost the same as ModelAwareTrait.

```
$this->loadTable('Comments') // In 3.x and 4.x instantiates $this->Comments CommentsTable object
$this->loadTables(['Comments', 'Users']);
$this->initTable('Comments') // Returns CommentsTable object, loadModel() in Controller domain could do the same.
```

I am okay with this PR. I just don't think that it improves things from an implementor's perspective much and that we might have duplicated features before 4.x if we want to improve the implementor/developer experience.

robertpustulka:
We probably could add `loadTable` to `LocatorAwareTrait` as a shorthand method for `$this->getTableLocator->get()`.

inoas:
We should not do that :fearful:: `loadTable()` should behave in a similar way `loadModel()` does. and `LocatorAwareTrait::get()` does not.

However, could we add `initTable()` to `LocatorAwareTrait` as a shorthand method for `$this->getTableLocator->get()` and add LocatorAwareTrait to `class Table`?

robertpustulka:
How is `initTable` better than `loadTable`?

inoas:
`initTable()` would be `getTable()` doing two things:
- circumventing your fear or method name collisions that you raised here https://github.com/cakephp/cakephp/pull/10933#issuecomment-317339363
- `init` vs `get`: Making clear that you don't just "get" something, but that you also create an object; `newTable()` also came to my mind and `createTable()` or `buildTable()` but the latter two can be miss understand easily. and `newTable()` sounds like you want to create a new schema.

saeideng:
anyway we need short naming for getting `table`
I strongly suggestion is 
` Cake\orm\table::get('posts')`
this is very short name
within `table::get` we can use `\Cake\ORM\Locator\LocatorAwareTrait;` or other class
`table::get` is just a wrapper 
```
class table {
 function get($name){
    return TableLocator::getInstance()->get($name);  
}
}
```
this is for static way

robertpustulka:
@inoas Locator `get()` method will create an object if a table isn't found in the registry.

@saeideng This PR aims to reduce static calls and ideally drop global table registry in 4.0. 
The method needs to operate in scope of an object.

inoas:
@robertpustulka sure it will, but it will return the object where `loadModel()` will mutate `$this`

Here is some pseudo-code to make it clear, if it isn't by now:

```php
// this->getTableLocator->get($name) or TableLocator::getInstance()->get($name)
class Table {
    public function initTable($name) {
        return TableLocator::getInstance()->get($name);  
    }
    public function loadTable($name) {
        $this->{$name} = TableLocator::getInstance()->get($name);
    }
    public function loadTables(array $names) {
        $tableLocator = TableLocator::getInstance();
        foreach ($names as $name) {
            $this->{$name} = $tableLocator->get($name);
        }
    }
}
```

robertpustulka:
@inoas I can see your point.

Although I'm not sure if the `loadTable()` or `initTable()` methods are necessary. There's already `ModelAwareTrait::loadModel()` method which will load a table as a default model.

inoas:
`loadModel()` however can load non-table repositories. `loadTable()` would not.

robertpustulka:
> loadModel() however can load non-table repositories

Yes, if you tell it to do so.

robertpustulka:
I'm not convinced that a `loadTable` method is necessary. The `loadModel` will load a table repository by default.

Is there anything more to do here?

lorenzo:
> Is there anything more to do here?

Again, I'm not sold to the idea of making TableLocator a singleton.

I would prefer the singleton be kept in TableRegistry for now and keep the rest of the changes as you have them. If we decide in the next release to finally have a singleton in TableLocator, then we can do that freely.

robertpustulka:
>  I'm not sold to the idea of making TableLocator a singleton.

This is not the idea of this PR. The singleton has been moved from the TableRegistry to the TableLocator just for BC reasons (so the entire TableRegistry class could be deprecated) and I'd like to remove it in 4.x.

saeideng:
> I would prefer the singleton be kept in TableRegistry for now 

:+1:  :ok_hand:

robertpustulka:
@lorenzo 
 > I would prefer the singleton be kept in TableRegistry for now

You mean keep only `TableRegistry::getTableLocator()` method, and leave static proxy calls deprecated, or keep all `TableRegistry` methods "active"?

lorenzo:
keep only `TableRegistry::getTableLocator()` method, and leave static proxy calls deprecated

saeideng:
And `tableRegistry::get()`

robertpustulka:
OK, makes sense. This way we can safely remove the singleton locator in 4.x if necessary without the need for deprecating any new static methods in `TableLocator` first.

robertpustulka:
@lorenzo how about it now?

jimgwhit:
Read this

> Facades provide a "static" interface to classes that are available in the application's service container. Laravel ships with many facades which provide access to almost all of Laravel's features. Laravel facades serve as "static proxies" to underlying classes in the service container, providing the benefit of a terse, expressive syntax while maintaining more testability and flexibility than traditional static methods.

Noting

> providing the benefit of a terse, expressive syntax 

Just do like that.  

I like as example
```
DB::whatever
```
better than 15 arrows $this->whatever->another->

robertpustulka:
„Facade” is just a fancy word for „global”.
Laziness shouldn’t justify bad design.

But this is just my opinion.

jimgwhit:
In php I can
```
$Dog = new \app\models\Dog();
$dogs = $Dog->getDogs($pages->offset(), $pages->Perpage(), $dogsearch);
```
or with __callStatic I can make a class like
```
<?php

namespace App\Helpers;

class DBS
{

    public static function __callStatic($method, $params)
    {
        $trace = debug_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS, 2);

        if (isset($trace[1]['class'])) {
            $class = $trace[1]['class'];
            $slh = strrpos($class, "\\") + 1;
        } else {
            return null;
        }
        $model = "\App\Model\SMVC" . "\\" . substr($class, $slh, -11) . "Model";
        $instance = new $model();

        return call_user_func_array(array($instance, $method), $params);
    }


}
```
then I can in controller 
```
use App\Helpers\DBS;   //at top
// and in method
$dogs = DBS::getDogs($pages->offset(), $pages->Perpage(), $dogsearch);
```
So that is not global, the __callStatic method still 
```
$instance = new $model();
```
in the background.  Bottom line Myself and a lot of other folks like
```
::
```
over
```
->
```
Furthermore __callStatic is a php thing.
But in no way is it global.

robertpustulka:
This is not how the TableRegistry/Locator works so the example isn't valid in this discussion.

TableRegistry uses a global locator instance which is accessible from any place in the application. Changing its state in one place affects another.

inoas:
So if it is not global, why is the example invalid?

jimgwhit:
My example was just a general php example, not unique to cakephp.  But the point is php has  __callStatic. And a lot of people like calling things with the double colon instead of a bunch of arrows.  
No not necessary, but looks so much nicer.  
And there is nothing wrong with static methods anyway.  But using __callStatic in my example isn't calling a static method, it's calling a non-static method using what I think is the pretty
```
::
```
My example was just to show that even in cakephp, you can make pretty calls.

robertpustulka:
There's no state preserved in static scope of @jimgwhit class example, so this is not global in mu understanding. This looks something like a convinience factory method and I'm okay with that. 

But this is not the same as using static variables. What's different between: `SomeClass:$var = 'foo';` and `global $var = 'foo'`?

davidyell:
I'm not in favour of the introduction of more magic methods in use of `__callStatic` nor do I think the style of code should define the unlying implementation.

Otherwise I don't see any issues with the PR. I like @inoas idea of the functional separation of the methods though. As it is important to remember that `loadModel` does more than just load a table instance. I really like the idea of being able to pass an array of table aliases to load.

robertpustulka:
> I really like the idea of being able to pass an array of table aliases to load.

@davidyell Good idea! But this and seperation of `loadModel` is out of the scope of this PR I'm afraid. I'd  like to see that implemented though. Sometimes I miss the old school `$uses` variable :)

ADmad:
> Sometimes I miss the old school $uses variable :)

Maybe we can modify `Table::__get()` to allow auto loading any model instead of just the one set as modelClass :)

inoas:
> Maybe we can modify Table::__get() to allow auto loading any model instead of just the one set as modelClass :)

:trollface: ?

inoas:
> @davidyell Good idea! But this and seperation of loadModel is out of the scope of this PR I'm afraid. I'd like to see that implemented though. Sometimes I miss the old school $uses variable :)

Well the scope is to deprecate and you say:

```php
// new way:
use \Cake\ORM\Locator\LocatorAwareTrait;
// ...
$users = $this->getTableLocator()->get('Users');
```

Where I'd suggest:

```php
// Via some Trait pre-attached to Table, Behavior, Controller, Component
$users = $this->initTable('Users');
```
... which is a lot more concise and I still don't know why/where it would hurt.

The same trait could then supply `loadTable()` and `loadTables()` which would attach the instances to `$this` like `loadModel()` does.

Not good?

robertpustulka:
I’m sorry, i wrote „new way” but this is a preferred way since 3.1. So this PR does not add anything new. It deprecates the static API what should have been done in 3.1

robertpustulka:
@inoas I updated the description.

I can see the point of having convenience methods, but a "quick access" to the model is already possible using `ModelAwareTrait` and `loadModel()` method. We can discuss the point of having another similar method (in `LocatorAwareTrait` or somewhere else) in another PR as this one is already a bit too large.

I also agree that I may have not made the intention of this PR clear enough. I hope it is a bit clearer now.

dereuromark:
O_o lots of conflicts now.
We should decide on proceeding here and then a final resolve + merge IMO.

robertpustulka:
@dereuromark Yeah, I intend to fix the conflicts after the changes are approved by @lorenzo

dereuromark:
@lorenzo ping

lorenzo:
How easy is to create a table locator? I worry about the ability to use the ORM in a simple php script or repl

robertpustulka:
@lorenzo 

Use `LocatorAwareTrait` and access it using `getTableLocator()` method.
Or just `new TableLocator()`.

robertpustulka:
@lorenzo ping :)

dereuromark:
It seems this is a rather controversial topic.
How about a compromise: Moving forward, but not adding deprecationWarning() into TableRegistry::get().
This way, people do not have to use the non-static approach here, but core can use it, and so can anyone who wants to.

lorenzo:
It's definitely going to be deprecation hell for people migrating to cake 3.6.

@robertpustulka can you change the readme for the ORM package so that it shows you now need to instantiate a table locator?

dereuromark:
Yeah, since it really just delegates it through, I vote for keeping both around.
No harm, no additional baggage really IMO.

robertpustulka:
@lorenzo readme is updated in this PR as well: https://github.com/cakephp/cakephp/pull/10933/files#diff-be6e833da32246111b12773eea161b28

I'm leaning towards @dereuromark opinion on this. I could remove warnings but keep deprecation annotations.

Another idea is to keep static API as it is but update the core to never use TableRegistry static methods and update docs to show how to use locators instead (this is already done here more or less). This way we could keep the static methods for people that use them (IDEs wouldn't mark methods as deprecated) but encourage new people to use locators instead (by never showing that something like `TableRegistry` exists). The static API could be depracated in 4.x and completely removed in 5.x then.

lorenzo:
fat fingers, sorry

lorenzo:
@robertpustulka I meant that I wanted it to show first instantiating `new TableLocator` and then explain you can add the trait to classes.

robertpustulka:
I fixed conflicts.

Also removed deprecation warnings for TableRegistry methods. I left annotations so it's clear that the TableRegistry use should be avoided.

lorenzo (approved):

src/Console/Shell.php
```diff
@@ -175,18 +176,19 @@ class Shell
...
      */
-    public function __construct(ConsoleIo $io = null)
+    public function __construct(ConsoleIo $io = null, LocatorInterface $locator = null)
     {
         if (!$this->name) {
             list(, $class) = namespaceSplit(get_class($this));
             $this->name = str_replace(['Shell', 'Task'], '', $class);
         }
         $this->_io = $io ?: new ConsoleIo();
+        $this->_tableLocator = $locator;
```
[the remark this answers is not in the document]
robertpustulka: The locator will fallback to the global one if is set to null.

robertpustulka:
Can this be merged now?

lorenzo:
Yes

markstory:
@robertpustulka Can you update the 3.6 migration guide for these changes?

robertpustulka:
Is that ok? https://github.com/cakephp/docs/commit/d5a4362a9a7f3d5a9336279bcebe3c2efbafe04d

markstory:
Yes. We should also start thinking about how we'd update code examples for when `TableRegistry::get()` is eventually removed.

robertpustulka:
Static get calls have already been removed from the code. All tests now use a trait for accessing the locator.

ADmad:
@robertpustulka I think he means the code examples in the manual.

davidyell:
Yes, I would like to see a complete code implementation example somewhere with this new api.

mbryja:
In my opinion deprecation of the TableRegistry::get() static call was a bad idea.

First of all, using LocatorAwareTrait is wrong on many levels. Most important is that by using traits in such way breaks Single Responsibility and Separation of Concerns principles.

I prefer to use delegation with combination of flyweight/singleton approach. I know that it is done by the LocatorAwareTrait but the only problem is that it exposes the (get/set)TableLocator methods that can be used incorrectly.

In other words if I have following facade:

    class Fruits {
        use \Cake\ORM\Locator\LocatorAwareTrait;

        public function getApples() { ... }
        public function getOranges() { ... }
        ...
    }

    $fruits = new Fruits();

I don't want to be able to call $fruits->getTableLocator()->get('table') outside of the scope of Fruits.

The other thing you need to consider when you make such changes is the adaptation of the framework. Doing TableRegistry::getTableLocator()->get('table') every time I need to access the model is not the best thing if I have multiple modules in my application that move beyond simple layered architecture.
Having flyweight/singleton class like TableRegistry with property get to access desired model just makes the development more straight forward and life easier.

Ideally I would just like to call TR::get('table') although that breaks the Cake's coding standards.

PR #10935, opened 2017-07-22
Add failing test for Http/Client/Response

ADmad:
`$response->getBody()->getContents()` returns empty string instead of same output as `$response->body()`.

I was trying use the response instance created by `HttpClient()` with an external lib which expects a `Psr\Http\Message\ResponseInterface` instance.

### DIFF
--- a/src/Http/Client/Response.php
+++ b/src/Http/Client/Response.php
@@ -151,6 +151,7 @@ class Response extends Message implements ResponseInterface
         }
         $stream = new Stream('php://memory', 'wb+');
         $stream->write($body);
+        $stream->rewind();
         $this->stream = $stream;
     }
 
--- a/tests/TestCase/Http/Client/ResponseTest.php
+++ b/tests/TestCase/Http/Client/ResponseTest.php
@@ -102,10 +102,12 @@ class ResponseTest extends TestCase
         $encoded = json_encode($data);
 
         $response = new Response([], $encoded);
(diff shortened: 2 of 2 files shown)


dakota (commented):
Since the response body is a stream (that has already been read). You need to first rewind the stream. Alternatively, and probably easier is to typecast it to a string, the PSR7 spec says that it must automatically rewind itself then.

ADmad:
@dakota I moved the call `$response->getBody()->getContents()` up so that it's the first statement after response instance if return. It still fails. 

For reference I am trying to update the Cake adapter for [HttpPlug](http://httplug.io/) to use Cake 3.4. [These](https://github.com/ADmad/cakephp-adapter/commit/e5ba39ea0585f82cd821b20023824fc27abd3bb8) are the changes and these are it's [failing tests](https://travis-ci.org/ADmad/cakephp-adapter/jobs/256312997).

It I uncomment [this](https://github.com/ADmad/cakephp-adapter/commit/e5ba39ea0585f82cd821b20023824fc27abd3bb8#diff-ff73e042e738204c6da009e2ed19f783R62) which generates a new guzzle psr7 response instance then the tests pass.

ADmad:
Rewinding the stream [here](https://github.com/cakephp/cakephp/blob/master/src/Http/Client/Response.php#L154) fixes the issue.

markstory:
Looks good :+1:

PR #10941, opened 2017-07-24
Shell - Improve output when running unknown subcommand

HavokInspiration:
Currently, if you try to run an unknown subcommand, it just outputs the current command help, which is rather confusing and not very dev-friendly.
So in light to recent efforts made to improve guidance when a user mistake is made and improve error messages, this change aims to improve that by displaying a real error message and some directions.
For instance, if I try to run `bin/cake cache cle`, I'll get this instead of the entire output of `bin/cake cache --help` :

```sh
bin/cake cache cle
```
help-unknown-subcommand

The guess is made with the `levenshtein` method. It simply takes the one with the better score. If multiple have the same score, the first one is used.

This is just a proposition. It it is accepted, maybe this can also be done with options, arguments and even root commands.

### DIFF
--- a/src/Console/ConsoleOptionParser.php
+++ b/src/Console/ConsoleOptionParser.php
@@ -753,7 +753,7 @@ class ConsoleOptionParser
      * wanted to say. Output a list of available subcommands as well.
      *
      * @param string $command Unknown command name trying to be dispatched.
-     * @return array
+     * @return string The message to be displayed in the console.
      */
     protected function getUnknownCommandOutput($command)
     {
--- a/src/Console/ConsoleOptionParser.php
+++ b/src/Console/ConsoleOptionParser.php
@@ -775,6 +775,7 @@ class ConsoleOptionParser
             $out[] = '';
         }
         $out[] = sprintf('Available subcommands for the `%s` command are : ', $rootCommand);
+        $out[] = '';
         foreach ($subcommands as $subcommand) {
             $out[] = ' - ' . $subcommand;
         }
@@ -796,7 +797,7 @@ class ConsoleOptionParser
             $score = levenshtein($command, $subcommand);
 
             if (!isset($bestScore) || $score < $bestScore) {
-                $bestScore = levenshtein($command, $subcommand);
+                $bestScore = $score;
                 $bestGuess = $subcommand;
             }
         }
--- a/src/Console/ConsoleOptionParser.php
+++ b/src/Console/ConsoleOptionParser.php
@@ -721,6 +721,16 @@ class ConsoleOptionParser
      */
     public function help($subcommand = null, $format = 'text', $width = 72)
     {
+        if ($subcommand === null) {
+            $formatter = new HelpFormatter($this);
+
+            if ($format === 'xml') {
+                return $formatter->xml();
+            }
+
+            return $formatter->text($width);
+        }
+
         if (isset($this->_subcommands[$subcommand])) {
             $command = $this->_subcommands[$subcommand];
             $subparser = $command->parser();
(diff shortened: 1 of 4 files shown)


tests/TestCase/Console/ShellTest.php
```diff
@@ -161,792 +161,6 @@ public function testConstruct()
...
-
-        $this->Shell->quiet('Just a test');
-    }
-
-    /**
-     * testOut method
-     *
-     * @return void
-     */
-    public function testOut()
```
dereuromark: Why are so many tests removed?

src/Console/Shell.php
```diff
@@ -472,12 +472,64 @@ public function runCommand($argv, $autoMethod = false, $extra = [])
...
     }
 
     /**
+     * Output a message in the console stating that the command can not be found and tries to guess what the user
+     * wanted to say. Output a list of available subcommands as well.
+     *
+     * @param string $command Unknown command name trying to be dispatched.
+     * @return void
+     */
+    protected function outputUnknownCommand($command)
```
markstory: This feels awkward on Shell. Shell should be a starting point for userland classes. This feature feels like something that the ConsoleOptionParser or HelpFormatter should be doing.

tests/TestCase/Console/ShellTest.php
markstory: I second that question.

markstory (commented):
I like the idea, and the error message is far more helpful. I'm not confident this should be a method on `Shell` though.

tests/TestCase/Console/ShellTest.php
HavokInspiration: What on earth did I do ?!
I'll fix it.

src/Console/Shell.php
HavokInspiration: > Shell should be a starting point for userland classes.
Fair enough.

I can put the method in `ConsoleOptionParser` to build the message to be displayed and just have a call to `out()` in `Shell`.

markstory: I think the ConsoleOptionParser has either direct or indirect access to the ConsoleIo object.

HavokInspiration: I checked and it does not seem to.
The previous behavior (outputting the full help of the command) that this PR replace did this : 

```
$this->out($this->OptionParser->help($command));
```
https://github.com/cakephp/cakephp/pull/10941/files#diff-d606d81f2aa3b802daa770d43e96be2fL475

markstory: Ok. If `help()` returned the unknown command text wouldn't it all still work?

HavokInspiration: That would keep the API cleaner indeed. 
There would three cases in `help()` then :

- known `subcommand` : returns the help for the subcommand
- unknown `subcommand` : returns the new output this PR aims to implement
- `null` subcommand passed (first argument of `ConsoleOptionParser::help()`) : display full command help

markstory: I think having the optionparser handle those 3 cases is reasonable. If the logic gets too complex we can factor out the big chunks into separate classes.

HavokInspiration:
As discussed with @markstory, I pushed a few commits which transfers the responsibility of building the output message in case of an unknown subcommand to the `ConsoleOptionParser::help()` method.

src/Console/ConsoleOptionParser.php
```diff
@@ -735,13 +745,63 @@ public function help($subcommand = null, $format = 'text', $width = 72)
...
+        );
+        $out[] = '';
+        if ($bestGuess !== null) {
+            $out[] = sprintf('Did you mean : `%s %s` ?', $rootCommand, $bestGuess);
+            $out[] = '';
         }
-        if ($format === 'xml') {
-            return $formatter->xml();
+        $out[] = sprintf('Available subcommands for the `%s` command are : ', $rootCommand);
+        foreach ($subcommands as $subcommand) {
```
markstory: Should there be a newline before the list of available commands?

src/Console/ConsoleOptionParser.php
```diff
@@ -735,13 +745,63 @@ public function help($subcommand = null, $format = 'text', $width = 72)
...
+     * @return string|null The closest name to the command submitted by the user.
+     */
+    protected function findClosestCommand($command, $subcommands)
+    {
+        $bestGuess = null;
+        foreach ($subcommands as $subcommand) {
+            $score = levenshtein($command, $subcommand);
+
+            if (!isset($bestScore) || $score < $bestScore) {
+                $bestScore = levenshtein($command, $subcommand);
```
markstory: Couldn't you do `$bestScore = $score`?

HavokInspiration:
@markstory I addressed your comments. Thanks for the review.

markstory:
Looks great. I wonder if we shouldn't do something similar for missing commands and missing options?

HavokInspiration:
@markstory I was planning on doing the same thing for options and commands in two separate PR. Smaller chunks is easier for me to manage my time.

HavokInspiration:
Closing. After discussing with @markstory, we decided to move this to 3.next. I'll open another PR soon.

HavokInspiration:
Replaced in favor of #10974

PR #10942, opened 2017-07-24
Add a getter for Response::charset

cleptric:
I also smuggeld some typo fixes into this PR

### DIFF
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -4,8 +4,6 @@ parameters:
     ignoreErrors:
         - '#Function wincache_ucache_[a-zA-Z0-9_]+ not found#'
         - '#Function xcache_[a-zA-Z0-9_]+ not found#'
-        - '#Function mcrypt_[a-zA-Z0-9_]+ not found#'
-        - '#Function apcu_[a-zA-Z0-9_]+ not found#'
         - '#Function fastcgi_[a-zA-Z0-9_]+ not found#'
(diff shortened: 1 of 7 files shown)


src/Http/Response.php
```diff
@@ -1153,7 +1153,7 @@ public function mapType($ctype)
      *
      * @param string|null $charset Character set string.
      * @return string Current charset
-     * @deprecated 3.4.0 Use withCharset() instead.
+     * @deprecated 3.5.0 Use getCharset()/withCharset() instead.
```
cleptric: Should this be `3.4.0/3.5.0` or is it fine this way?

dereuromark: It can only accept one number, we either chose the lowest or highest.

markstory:
Looks good. I will get the phpstan errrors fixed up.

phpstan.neon
```diff
@@ -4,6 +4,8 @@ parameters:
     ignoreErrors:
         - '#Function wincache_ucache_[a-zA-Z0-9_]+ not found#'
         - '#Function xcache_[a-zA-Z0-9_]+ not found#'
+        - '#Function mcrypt_[a-zA-Z0-9_]+ not found#'
```
ADmad: You need to revert this extra ignores you added. They are causing the travis build to fail since these methods are actually available there.

ADmad (changes requested):

cleptric:
Phpstan is happy now 😀

PR #10943, opened 2017-07-24
Split RouteBuilder::routeClass() and RouteBuilder::extensions()

chinpei215:
Refs #9978


markstory:
Looks good. This is another builder where the get method are almost never used but the consistency and ability to add typehinting makes this good.

PR #10947, opened 2017-07-25
Fix pagination ordering on calculated columns on SQL Server.

### LINKED ISSUE #10944: Bug or (docs are wrong) for SQL Server and ordering by synthetic columns
This is a (multiple allowed):

* [x] bug
* [ ] enhancement

PHP 7.2
CakePHP 3.4.11
Windows 7
SQL Server 2008

I'm trying to paginate a query result set that has a synthetic (aka calculated) column.  But the ordering will not work.  AFAIK I am following the proper steps, but it appears that SQL Server will not accept ordering by synthetic columns with the ordering strategy that Cake and it's ORM + Query builder is using.

The code looks soomething like the following.

Model has a custom finder:

```PHP
public function findWithCounts(Query $query, array $options)
{
    return $query
        ->select(['number_applied' => $query->func()->count('DpsComponents.id')])
}
```

The pagination configuration is specifying ```number_applied``` in the whitelist.  But sorting by that column gives the following SQL:

```SQL
FROM 
  (
    SELECT 
      (
        COUNT(DpsComponents.id)
      ) AS [number_applied], 
      (
        ROW_NUMBER() OVER (
          ORDER BY 
             number_applied asc
        )
      ) AS [_cake_page_rownum_] 
    FROM 
      tbl_dps Dps 
      LEFT JOIN tbl_dps_components DpsComponents ON Dps.id = (DpsComponents.dp_id) 
  ) _cake_paging_ 
WHERE 
  _cake_paging_._cake_page_rownum_ <= 25
```

I've chopped the actual code down to show the relevant parts - apologies if there are unrelated syntax errors as a result.  The actual example is quite large.

The core of the problem appears to be that SQL server will not accept aliases in the  ```ROW_NUMBER() OVER``` clause as outlined here:

https://stackoverflow.com/questions/7297609/cannot-use-alias-in-row-number-over-in-sql-server

One possible solution to get the SQL to run is to change the ROW_NUMBER clause to this:

```SQL
      (
        ROW_NUMBER() OVER (
          ORDER BY 
             COUNT(DpsComponents.id) asc
        )
      ) AS [_cake_page_rownum_] 
```

I have not tried this workaround with Pagination sort links because I have other synthetic columms that are the result of much more complicated calculations and am doubtfull I will get them all to work.

So I guess I had a couple of questions:

1. Am I doing something wrong with my approach?
1. If no, is this a CakePHP bug with the SQL server integration?
1. If it is a bug are there any workarounds?  Back to the messy custom pagination I was using with CakePHP 2 ... (is that still possible with CakePHP3?).  Or is it possible to get Cake pagination to use the calculation in the order by clause rather than the alias?

MCF:
On older versions of SQL Server, the pagination is accomplished using a
ROW_NUMBER() OVER clause.  Unfortunately, OVER does not support the use of
column aliases.  But for calculated columns the only practical way to
specify their use in ordering is with their alias, this currently causes a
SQL error.

This fix will substitute the calculation's SQL syntax in place of the
column alias before generating the OVER clause for calculated columns,
fixing the bug.

An initial attempt to fix https://github.com/cakephp/cakephp/issues/10944

### DIFF
--- a/src/Database/Dialect/SqlserverDialectTrait.php
+++ b/src/Database/Dialect/SqlserverDialectTrait.php
@@ -106,29 +106,29 @@ trait SqlserverDialectTrait
         $field = '_cake_paging_._cake_page_rownum_';
 
         if ($original->clause('order')) {
-            $order = clone $original->clause('order');
+            // SQL server does not support column aliases in OVER clauses.  But
+            // the only practical way to specify the use of calculated columns
+            // is with their alias.  So substitute the select SQL in place of
+            // any column aliases for those entries in the order clause.
+            $select = $original->clause('select');
+            $order = new OrderByExpression();
+            $original
+                ->clause('order')
+                ->iterateParts(function ($direction, $orderBy) use ($select, $order) {
+                    $key = $orderBy;
+                    if (isset($select[$orderBy]) &&
+                        $select[$orderBy] instanceof ExpressionInterface
+                    ) {
+                        $key = $select[$orderBy]->sql(new ValueBinder());
+                    }
+                    $order->add([$key => $direction]);
+
+                    // Leave original order clause unchanged.
+                    return $orderBy;
+                });
         } else {
             $order = new OrderByExpression('(SELECT NULL)');
         }
-
-        // SQL server does not support column aliases in OVER clauses.  But for
-        // calculated columns the alias is the only practical identifier to use
-        // when specifying the order.  So if a column alias is specified in the
-        // order clause, and the value of that alias is an expression, change
-        // the alias into what it represents by setting the clause's key to be
-        // the SQL representation of its value.  The UnaryExpression creation
-        // below will then do the right thing and use the calculation in the
-        // ROW_NUMBER() OVER clause instead of the alias.
-        $select = $original->clause('select');
-        $order->iterateParts(function ($direction, &$orderBy) use ($select) {
-            if (isset($select[$orderBy])) {
-                if ($select[$orderBy] instanceof ExpressionInterface) {
-                    $orderBy = $select[$orderBy]->sql(new ValueBinder());
-                }
-            }
-
-            return $direction;
-        });
 
         $query = clone $original;
         $query->select([
--- a/src/Database/Dialect/SqlserverDialectTrait.php
+++ b/src/Database/Dialect/SqlserverDialectTrait.php
@@ -14,6 +14,7 @@
  */
 namespace Cake\Database\Dialect;
 
+use Cake\Database\ExpressionInterface;
 use Cake\Database\Expression\FunctionExpression;
 use Cake\Database\Expression\OrderByExpression;
 use Cake\Database\Expression\UnaryExpression;
@@ -21,6 +22,7 @@ use Cake\Database\Query;
 use Cake\Database\Schema\SqlserverSchema;
 use Cake\Database\SqlDialectTrait;
 use Cake\Database\SqlserverCompiler;
+use Cake\Database\ValueBinder;
 use PDO;
 
 /**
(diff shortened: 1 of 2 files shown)


src/Database/Dialect/SqlserverDialectTrait.php
```diff
@@ -102,7 +104,31 @@ public function _version()
...
+        // SQL server does not support column aliases in OVER clauses.  But for
+        // calculated columns the alias is the only practical identifier to use
+        // when specifying the order.  So if a column alias is specified in the
+        // order clause, and the value of that alias is an expression, change
+        // the alias into what it represents by setting the clause's key to be
+        // the SQL representation of its value.  The UnaryExpression creation
+        // below will then do the right thing and use the calculation in the
+        // ROW_NUMBER() OVER clause instead of the alias.
+        $select = $original->clause('select');
+        $order->iterateParts(function ($direction, &$orderBy) use ($select) {
```
markstory: Instead of pass by reference, could we build a new order clause?

src/Database/Dialect/SqlserverDialectTrait.php
```diff
@@ -102,7 +104,31 @@ public function _version()
...
+            if (isset($select[$orderBy])) {
+                if ($select[$orderBy] instanceof ExpressionInterface) {
+                    $orderBy = $select[$orderBy]->sql(new ValueBinder());
```
markstory: This will likely have problems if the expression has bound parameters.

markstory:
Would it be possible to have a test added that exercises this code? We run all our tests against SQLServer2008 via AppVeyor.

src/Database/Dialect/SqlserverDialectTrait.php
MCF: Yes, I assumed that would be an issue.  But I don't really know how to get the bound parameters within the Trait.  Any suggestions?

MCF:
Looking in appveyor.yml all the references to SQL Server version number seem to indicate 2012.  Is there another setup for 2008?

src/Database/Dialect/SqlserverDialectTrait.php
MCF: My knowledge of the QueryExpression object and it's descendants is limited.  Here is how I would change to creating a new order clause (IMHO it is clunkier than the pass by reference).  Thoughts or improvements before I implment it?

```PHP
        if ($original->clause('order')) {
            $select = $original->clause('select');
            $order = new OrderByExpression();
            $original
                ->clause('order')
                ->iterateParts(function ($direction, $orderBy) use ($select, $order) {
                    $key = $orderBy;
                    if (isset($select[$orderBy])) {
                        if ($select[$orderBy] instanceof ExpressionInterface) {
                            $key = $select[$orderBy]->sql(new ValueBinder());
                        }
                    }
                    $order->add([$key => $direction]);

                    return $orderBy;
                });
        } else {
            $order = new OrderByExpression('(SELECT NULL)');
        }
```

MCF: The problem is that I need the keys from QueryExpression's _conditions, otherwise I could use traverse().

src/Database/Dialect/SqlserverDialectTrait.php
markstory: Why do you need the bound parameters? Does attaching the expression to the orderby clause not work?

markstory:
@MCF then my memory fails me. There is only the one appveyor configuration file.

antograssiot:
@markstory we removed the build against 2008 a long time ago as it was repetively failling with tumeout after 30 minutes IIRC

src/Database/Dialect/SqlserverDialectTrait.php
MCF: My fix works fine.  The QueryExpression stuff has (seemingly) many variations possible.  I was just agreeing with you that if bound parameters are used in an order by clause it will (still) not work.

I'll admit that I'm still figuring out the CakePHP 3 way of doing things with queries and it is far more featured the version 2.  I'm not really able to test all possible inputs yet because I don't know what those are yet.  This pull request fixes the original issue I logged just fine though.

MCF:
I've change things to avoid pass by reference and created a new ```OrderByExpression``` instead.

lorenzo:
Is the limitation also present in SQLServer 2012?

MCF:
I don't have SQLServer 2012 to test against to easily answer that question.  I suspect that the problem may also occur with SQLServer 2012 and up as _transformDistinct also uses the ROW_NUMBER() OVER technique to order the query.  But it is hard for me to test easily without a newer version of SQLServer.

MCF:
I suppose we could add a test that selects a caculated column and then orders by it?  the Appveyor build should then throw the error if it will fail with SQLServer 2012.

MCF:
The test I just added seems to run fine with SQL Server 2012, so I don't think the same issue is present with 2012.

lorenzo:
This looks good to me

lorenzo:
Thanks a lot @MCF

src/Database/Dialect/SqlserverDialectTrait.php
```diff
@@ -102,7 +104,31 @@ public function _version()
...
+            $select = $original->clause('select');
+            $order = new OrderByExpression();
+            $original
+                ->clause('order')
+                ->iterateParts(function ($direction, $orderBy) use ($select, $order) {
+                    $key = $orderBy;
+                    if (isset($select[$orderBy]) &&
+                        $select[$orderBy] instanceof ExpressionInterface
+                    ) {
+                        $key = $select[$orderBy]->sql(new ValueBinder());
```
markstory: I'm not sure this is going to work all the time, as you'll drop bound parameters. That scenario didn't work in the past either, so I guess this is incremental progress.

MCF: @markstory - if you can give me an example of a select statement using a calculated value and bound parameters I'd be happy to see if I can fix it for that case.  I'm still struggling to understand all the details of how to build SQL via the (new to me) V3 ORM stuff.  There are so may variations in syntax and approach that it can be hard to sort out the details.

markstory: Something like 

```php
$query->select(['name' => $query->func()->concat('bob', 'jill')])
```

Issue #10946, opened 2017-07-25
[RFC] Getters/setters for Controller request and response objects.


robertpustulka:
This is a (multiple allowed):

* [ ] bug
* [ ] enhancement
* [x] feature-discussion (RFC)

* CakePHP Version: 3.5.0

Controller `$this->request` and `$this->response` properties are still public.

I'm wondering if we should deprecate the direct access and add getters and setters instead?

robertpustulka:
BTW there are number of public properties stil in use. Like `$this->name` and  `$this->plugin`  which could be retrieved from request params.

inoas:
Agreed, either we should only have public properties or only getters/setters, never both!

saeideng:
your means is `$this->request` Or `$this->request->methodName` ?

markstory:
We could make the properties protected as well 😄

robertpustulka:
But for BC these should be accessible using `__get` and `__set`?

inoas:
Just deprecate them now and remove them in 4.x without adding more code @ `__get`?

markstory:
Removing these properties is going to have wide sweeping consequences as these are the most used properties on controllers, and changing them will impact all controllers and components. It will also introduce inconsistencies with `View`. Perhaps we need to think about these properties more holistically?

robertpustulka:
I'm afraid we can't make `$request` and `$response` protected and use `__get`/`__set` instead because of `Indirect modification of overloaded property Controller::$request has no effect` error.

Can we mark those deprecated in 3.5 and change the visibilty in 4.0?

chinpei215:
I think you can avoid the error if you change `__get()` to return reference. I was planning to change `__get()` in 2.x, but the last bus had already gone 😭

robertpustulka:
Thanks, I'll try.

saeideng:
we can validate that on `__set()`

robertpustulka:
PR is open.

Issue #10930, opened 2017-07-20
4.0+ Reduce setter/getter Verbosity


inoas:
This is a (multiple allowed):

* [x] enhancement
* [x] feature-discussion (RFC)
* CakePHP Version: 3.4.x

### What you did

- Used the query builder
- used request/response objects

### What happened

Realized how nice it is to have a concise interface over verbose setters/getters.

### What you expected to happen

I was hoping there would be a distinction of stateful objects in core cakephp grouped into following categories in future:

1. builders
2. vanilla read/write objects
3. readers and immutable objects

#### Builders

Builders could drop the 'set' prefix from all setters and/or would not get them for 4.x.

**Aka the ORM's query builder is a good example**. It is 99% about using setters, then at some point firing off the query.

When using builders without the setter prefix we can get very close to the visual size of array options, too. Example:

```php
function o($filter) { return OptionsFactory::create($filter); } // somewhere in bootstrap or so

$options = ['controller' => 'Articles', 'action' => 'view', 5, 'q' => ['show' => 'recent']];
// vs
$options = o('url')->controller('Articles')->action('view')->param(5)->q(['show' => 'recent']);
// __call() can be used as a generic setter or the setters can be declared functions to help with IDE support
```

#### Readers / Immutable Objects

Say working with request/response here there is no need to use the `get` prefix. To modify an object that is mostly being read one simply uses `setFoo()` or to copy on change a immuntable object `withFoo()`.

#### Vanilla Read/Write Objects

Entities come to mind... their properties are often set and get and thus both prefixes make sense.

Ref: https://github.com/cakephp/cakephp/pull/10887#issuecomment-316465007
Ref: https://github.com/cakephp/cakephp/issues/9978#issuecomment-323011610

davidyell:
I much prefer the verbosity as it makes my code easier to read and thus reason about without having to go and read method source code.

markstory:
Don't builders add more complexity over vanilla objects? There would be 2 classes to learn, use update and write tests for.

inoas:
I didn't mean to create two classes; just to categorize all classes that are instantiated as stateful objects into 2 or 3 groups.

Objects similar to the query builder would then not get setter prefixes (but 'could' have getter methods). Etc.

davidyell:
I'd be worried about the provision of a consistent api interface with the classes if things started getting mixed up. It's quite possible that users could try using an object in a way it was not intended for and get confused.

dereuromark:
[...]

PR #10949, opened 2017-07-26
Deprecate public Controller properties

robertpustulka:
Refs #10946

I'm afraid it's hard to change the visibility of `$request` and `$response` properties now, so I kept them public but added deprecation annotations to them.

`$name`, `$plugin`, `$autoRender` properties are now protected and accessed via magic methods for BC.

It was also hard to refactor the code to use request controller name and plugin params as those not always share the same state, at least in tests. So these are still controller properties. We can get back to this in 4.0 though.

I added deprecation annotations to `$components` and `$helpers` as well because those are deprecated since the 3.0 I believe.

After this is reviewed and merged, I'll update the core to use new methods.

### DIFF
--- a/src/Controller/Controller.php
+++ b/src/Controller/Controller.php
@@ -213,9 +213,8 @@ class Controller implements EventListenerInterface, EventDispatcherInterface
      * Automatically set to the name of a plugin.
      *
      * @var string
-     * @deprecated 3.6.0 Use `$this->request->getParam('plugin')` instead.
      */
-    public $plugin;
+    protected $plugin;
 
     /**
      * Holds all passed params.
@@ -348,7 +347,7 @@ class Controller implements EventListenerInterface, EventDispatcherInterface
         ];
         if (isset($deprecated[$name])) {
             $method = $deprecated[$name];
-            deprecationWarning(sprintf('Controller::$%s is deprecated. Use $this->%s() instead.', $name, $method));
+//            deprecationWarning(sprintf('Controller::$%s is deprecated. Use $this->%s instead.', $name, $deprecated[$name]['method']));
 
             return $this->{$method}();
         }
@@ -391,13 +390,14 @@ class Controller implements EventListenerInterface, EventDispatcherInterface
         ];
         if (isset($deprecated[$name])) {
             $method = $deprecated[$name];
-            deprecationWarning(sprintf('Controller::$%s is deprecated. Use $this->%s() instead.', $name, $method));
+//            deprecationWarning(sprintf('Controller::$%s is deprecated. Use $this->%s() instead.', $name, $method));
             $this->{$method}($value);
 
             return;
         }
         if ($name === 'autoRender') {
             $value ? $this->enableAutoRender() : $this->disableAutoRender();
+//            deprecationWarning(sprintf('Controller::$%s is deprecated. Use $this->enableAutoRender/disableAutoRender() instead.', $name));
 
             return;
         }
@@ -442,6 +442,31 @@ class Controller implements EventListenerInterface, EventDispatcherInterface
     public function setName($name)
     {
         $this->name = $name;
+
+        return $this;
+    }
+
+    /**
+     * Returns the plugin name.
+     *
+     * @return string
+     * @since 3.6.0
+     */
+    public function getPlugin()
+    {
+        return $this->plugin;
+    }
+
+    /**
+     * Sets the plugin name.
+     *
+     * @param string $name Plugin name.
+     * @return $this
+     * @since 3.6.0
+     */
+    public function setPlugin($name)
+    {
+        $this->plugin = $name;
 
         return $this;
     }
@@ -500,9 +525,7 @@ class Controller implements EventListenerInterface, EventDispatcherInterface
      * which must also be updated here. The properties that get set are:
      *
      * - $this->request - To the $request parameter
-     * - $this->plugin - To the $request->params['plugin']
      * - $this->passedArgs - Same as $request->params['pass]
-     * - View::$plugin - $this->plugin
      *
      * @param \Cake\Http\ServerRequest $request Request instance.
      * @return $this
--- a/src/Controller/Controller.php
+++ b/src/Controller/Controller.php
@@ -163,7 +163,6 @@ class Controller implements EventListenerInterface, EventDispatcherInterface
      * after action logic.
      *
      * @var bool
-     * @deprecated 3.6.0 Use enableAutoRender()/disableAutoRender() and isAutoRenderEnabled() instead.
      */
     protected $autoRender = true;
 
--- a/src/Controller/Controller.php
+++ b/src/Controller/Controller.php
@@ -125,6 +125,7 @@ class Controller implements EventListenerInterface, EventDispatcherInterface
      *
      * @var \Cake\Http\ServerRequest
      * @link https://book.cakephp.org/3.0/en/controllers/request-response.html#request
+     * @deprecated 3.5.0 Use getRequest()/setRequest instead.
      */
     public $request;
 
@@ -133,6 +134,7 @@ class Controller implements EventListenerInterface, EventDispatcherInterface
      *
      * @var \Cake\Http\Response
      * @link https://book.cakephp.org/3.0/en/controllers/request-response.html#response
+     * @deprecated 3.5.0 Use getResponse()/setResponse instead.
      */
     public $response;
 
@@ -451,6 +453,16 @@ class Controller implements EventListenerInterface, EventDispatcherInterface
     }
 
     /**
+     * Gets the request instance.
+     *
+     * @return \Cake\Http\ServerRequest
+     */
+    public function getRequest()
+    {
+        return $this->request;
+    }
+
+    /**
      * Sets the request objects and configures a number of controller properties
      * based on the contents of the request. Controller acts as a proxy for certain View variables
      * which must also be updated here. The properties that get set are:
@@ -461,7 +473,7 @@ class Controller implements EventListenerInterface, EventDispatcherInterface
      * - View::$plugin - $this->plugin
      *
      * @param \Cake\Http\ServerRequest $request Request instance.
-     * @return void
+     * @return $this
      */
     public function setRequest(ServerRequest $request)
     {
@@ -471,6 +483,31 @@ class Controller implements EventListenerInterface, EventDispatcherInterface
         if ($request->getParam('pass')) {
             $this->passedArgs = $request->getParam('pass');
         }
+
+        return $this;
+    }
+
+    /**
+     * Gets the response instance.
+     *
+     * @return \Cake\Http\Response
+     */
+    public function getResponse()
+    {
+        return $this->response;
+    }
+
+    /**
+     * Sets the response instance.
+     *
+     * @param \Cake\Http\Response $response Response instance.
+     * @return $this
+     */
+    public function setResponse(Response $response)
+    {
+        $this->response = $response;
+
+        return $this;
     }
 
     /**
--- a/src/Controller/Controller.php
+++ b/src/Controller/Controller.php
@@ -347,7 +347,7 @@ class Controller implements EventListenerInterface, EventDispatcherInterface
         ];
         if (isset($deprecated[$name])) {
             $method = $deprecated[$name];
-//            deprecationWarning(sprintf('Controller::$%s is deprecated. Use $this->%s instead.', $name, $deprecated[$name]['method']));
+            deprecationWarning(sprintf('Controller::$%s is deprecated. Use $this->%s instead.', $name, $method));
 
             return $this->{$method}();
         }
@@ -390,14 +390,14 @@ class Controller implements EventListenerInterface, EventDispatcherInterface
         ];
         if (isset($deprecated[$name])) {
             $method = $deprecated[$name];
-//            deprecationWarning(sprintf('Controller::$%s is deprecated. Use $this->%s() instead.', $name, $method));
+            deprecationWarning(sprintf('Controller::$%s is deprecated. Use $this->%s() instead.', $name, $method));
             $this->{$method}($value);
 
             return;
         }
         if ($name === 'autoRender') {
             $value ? $this->enableAutoRender() : $this->disableAutoRender();
-//            deprecationWarning(sprintf('Controller::$%s is deprecated. Use $this->enableAutoRender/disableAutoRender() instead.', $name));
+            deprecationWarning(sprintf('Controller::$%s is deprecated. Use $this->enableAutoRender/disableAutoRender() instead.', $name));
 
             return;
         }
--- a/src/Controller/Controller.php
+++ b/src/Controller/Controller.php
@@ -212,8 +212,9 @@ class Controller implements EventListenerInterface, EventDispatcherInterface
      * Automatically set to the name of a plugin.
      *
      * @var string
+     * @deprecated 3.5.0 Use `$this->request->getParam('plugin')` instead.
      */
-    protected $plugin;
+    public $plugin;
 
     /**
      * Holds all passed params.
@@ -252,14 +253,15 @@ class Controller implements EventListenerInterface, EventDispatcherInterface
         }
 
         $this->setRequest($request !== null ? $request : new ServerRequest());
-        $this->response = $response !== null ? $response : new Response();
+        $this->setResponse($response !== null ? $response : new Response());
 
         if ($eventManager !== null) {
             $this->setEventManager($eventManager);
         }
 
         $this->modelFactory('Table', [$this->getTableLocator(), 'get']);
-        $modelClass = ($this->plugin ? $this->plugin . '.' : '') . $this->name;
+        $plugin = $this->request->getParam('plugin');
+        $modelClass = ($plugin ? $plugin . '.' : '') . $this->name;
         $this->_setModelClass($modelClass);
 
         if ($components !== null) {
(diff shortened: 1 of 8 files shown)


src/Controller/Controller.php
```diff
@@ -159,7 +163,7 @@ class Controller implements EventListenerInterface, EventDispatcherInterface
      *
      * @var bool
      */
-    public $autoRender = true;
+    protected $autoRender = true;
```
chinpei215: I think this change is meaningless. Because when I touch this property, I am in the action of the controller usually. So I can continue to use this property without any warnings even if I am using some great IDE. Probably, it's the time to allow us to use `private` modifiers. To be honest, I think clean APIs wouldn't prefer setters/getters for public/protected properties or getters for private mutable objects.

inoas:
I'd either have also added $request and $response to the magic getters/setters or (preferably) not used magic getters/setters at all but just mark the properties with a deprecation warning that their visibility will change from public to private in 4.0 and that you should use the appropriate getter/setter methods.

robertpustulka:
> I'd either have also added $request and $response to the magic getters/setters

I tried, but many of the tests just failed. I'm afraid it's not that easy to do this.


> Because when I touch this property, I am in the action of the controller usually.

For me that is a reason this property shouldn't be public.

src/Controller/Controller.php
```diff
@@ -124,6 +126,7 @@ class Controller implements EventListenerInterface, EventDispatcherInterface
      *
      * @var \Cake\Http\ServerRequest
      * @link https://book.cakephp.org/3.0/en/controllers/request-response.html#request
+     * @deprecated 3.5.0 Use getRequest()/setRequest instead.
```
inoas: This is one of the places where I would strongly prefer `request()` for getting as that is the 99% case and everything else is annoying verbosity. See https://github.com/cakephp/cakephp/issues/10930.

I am not even sure if the request object should be changed? Should it? Or should it merely be cloned via `withFoo()`?

src/Controller/Controller.php
ADmad: This change is not BC and actually not meaningless. There could be some doing `$this->getController()->autoRender = false` in a component.

ADmad:
Changing public properties to protected in not backwards compatible, please revert it. I showed an example above how it can create problems.

robertpustulka:
@ADmad There are magic `__get` and `__set` updated to handle this. Please look at the diff :)

src/Controller/Controller.php
robertpustulka: @ionas I prefer the consistency. All getters here have a prefix, so this one should also have it. BTW `setRequest` has been around for the while and is used to set the state of the controller.

ADmad:
Sorry missed that :)

src/Controller/Controller.php
```diff
@@ -390,6 +418,85 @@ public function __set($name, $value)
...
+        return $this->autoRender;
+    }
+
+    /**
+     * Enable or disbale automatic action rendering.
+     *
+     * @param bool $autoRender Flag.
+     * @return $this
+     */
+    public function enableAutoRender($autoRender = true)
```
ADmad: If this method has `enable` prefix then it should only enable auto rendering. Currently doing `enableAutoRender(false)` would disable auto rendering which is not good for readability.

ADmad (changes requested):

chinpei215:
What do you think about benefits of setters/getters? In my opinion, setters ensure to set unexpected values to the property. But if it is a `protected` property, I can write `$this->autoRender = 0;` in my controller for example. So this kind of setters/getters don't have any advantages actually.

src/Controller/Controller.php
robertpustulka: I was following the examples in a query builder, like `enableAutoFields` and `enableHydration` which accept `false`.

robertpustulka:
@chinpei215 Right, I can add a typecheck to the `enableAutoRender` since we're not on PHP 7 yet.

chinpei215:
Even if we are on PHP7, properties will be weakly typed. Are you planning to change the modifier to `private` in 4.0? Then I am ok. But if you are not planning that, I am not in favor of this change, as people would continue to write `$this->autoRender = true;`.

src/Controller/Controller.php
ADmad: So either the method should be name `setAutoRender()` or two separate methods `enableAutoRender()` and `disableAutoRender()` should be added which don't take any argument.

robertpustulka:
@chinpei215 I can't recall if there has been any discussion or decision made on `private` properties. But I also think that some of the properties could become private.

chinpei215:
OK. Thank you for your reply :)

src/Controller/Controller.php
robertpustulka: @ADmad are you suggesting that we should do the same in a query builder? That obviously makes more sense then calling `enableSomething(false)`

markstory: I've found `enableThing(false)` to be clear enough, but I might be an outlier there.

src/Controller/Controller.php
```diff
@@ -410,6 +517,31 @@ public function setRequest(ServerRequest $request)
...
+
+    /**
+     * Sets the response instance.
+     *
+     * @param \Cake\Http\Response $response Response instance.
+     * @return $this
+     */
+    public function setResponse(Response $response)
+    {
+        $this->response = $response;
```
markstory: Because responses and requests are immutable I'm wondering if we should start thinking of ways to make working with responses less clunky. Using setResponse() is going to get annoying.

src/Controller/Controller.php
robertpustulka: Subclasses could call `$this->enable/disableAutoRender()` in their initialize methods.

src/Controller/Controller.php
```diff
@@ -206,7 +212,7 @@ class Controller implements EventListenerInterface, EventDispatcherInterface
      *
      * @var string
      */
-    public $plugin;
+    protected $plugin;
```
robertpustulka: We could but I've seen some tests that set the plugin property. I'm not sure how this would work with immutable requests but this could be a matter of refactoring tests.

src/Controller/Controller.php
robertpustulka: Are there many cases where a request is modified in a controller scope?

src/Controller/Controller.php
ADmad: For me personally even `something(false)` is good enough :)

src/Controller/Controller.php
markstory: They could. But making just _this_ property private would introduce inconsistencies with other parts of the framework and break compatibility.

src/Controller/Controller.php
markstory: We can start by refactoring our tests.

src/Controller/Controller.php
markstory: I'm sure there are. People do things like mutate request data quite often.

robertpustulka: Right, I forgot about data. 

In controller methods it will be still possible to do `$this->request = $this->request->withSomething()`

saeideng:
I think 
1)we need change scope of `request` and `response` to protected
2) un-deprecate   accessing `this->request` and `$this->response`
3)set those by `__set()` and validate if received object is `serverRequest` and `response` Object or no
4)run a error if received object is invalid 
5)get by `__get()` 

so we able to use 
`$this->request = $this->request->withSomething()`
`$this->response  = $this->response->withSomething()`
safety

chinpei215:
No. When you assign a value to a protected property from a derived class, `__set()` will not be called. So people can write `$this->request = 1;` unsafely. Please read the earlier discussion about `private` modifier.

saeideng:
what happen if we change `public $request` to `protected $_request` or `protected $_other_request_property`

saeideng:
i means
```
     public function setRequest(ServerRequest $request)
     {
         $this->request = $request;
         $this->plugin = $request->getParam('plugin') ?: null;
 
         if ($request->getParam('pass')) {
             $this->passedArgs = $request->getParam('pass');
         }

        return $this;
    }
```
```
         $deprecated = [
            'request' => 'setRequest',
            'response' => 'setResponse'
        ];
        if (isset($deprecated[$name])) {
            $method = $deprecated[$name];

            return $this->{$method}();
        }
```

saeideng:
you can not pass `1` to `public function setRequest(ServerRequest $request)`
I misunderstood ?

chinpei215:
> I misunderstood ?

Yes. First, people can access the protected property directly: `$this->_request = 1;`.
Second, if you declare `public $request = 1;` in a derived class, the property is evaluated as `1` in any methods of the parent class **if `private $request` is not declared** . So, for example, `$this->referer()` will cause fatal error because it calls `$this->request->referer()` internally.

saeideng:
I forgot we are in subcluss of cor controller :smiley:

robertpustulka:
I've added the `disableAutoRender()` method and marked the property deprecated. It may convince users to using new  methods in their actions instead.
I've also removed get/setPlugin methods, deprecated the property in favor of the requets param.

I'm going to remove the deprecated properties and method calls in another PR once this is merged.

robertpustulka:
Is there a need for deprecation warnings for the properties in __get and __set? I mean something like with the deprecated view builder properties.

cleptric:
@robertpustulka I would add them after 3.5 shipped.

src/Controller/Controller.php
```diff
@@ -251,14 +252,15 @@ public function __construct(ServerRequest $request = null, Response $response =
         }
 
         $this->setRequest($request !== null ? $request : new ServerRequest());
-        $this->response = $response !== null ? $response : new Response();
+        $this->setResponse($response !== null ? $response : new Response());
```
markstory: This could use `?:` now I think.

tests/TestCase/Controller/ControllerTest.php
```diff
@@ -1145,15 +1131,57 @@ public function testAutoRender()
...
+        $controller = new AnotherTestController();
+        $message = false;
+
+        set_error_handler(function ($errno, $errstr) use (&$message) {
+            $message = ($errno === E_USER_DEPRECATED ? $errstr : false);
+        });
+
+        try {
+            $controller->$property = $value;
+//
```
markstory: Looks like a mistake.

ADmad (approved):

markstory:
I am going to hold off merging this until 3.6 development starts. 3.5 is already a big release.

dereuromark:
Looks like this is now pretty much good to go.

src/Controller/Controller.php
markstory: @inoas You wouldn't need to use the methods *inside* the controller actions. In that context you could use the protected properties.

src/Controller/Controller.php
```diff
@@ -206,6 +213,7 @@ class Controller implements EventListenerInterface, EventDispatcherInterface
      * Automatically set to the name of a plugin.
      *
      * @var string
+     * @deprecated 3.5.0 Use `$this->request->getParam('plugin')` instead.
```
markstory: These numbers should be `3.6.0` now.

robertpustulka:
I've updated the numbers, added `@since` tags and fixed some tests to use getters and setters.

src/Controller/Controller.php
```diff
@@ -159,8 +163,9 @@ class Controller implements EventListenerInterface, EventDispatcherInterface
      * after action logic.
      *
      * @var bool
+     * @deprecated 3.6.0 Use enableAutoRender()/disableAutoRender() and isAutoRenderEnabled() instead.
      */
-    public $autoRender = true;
+    protected $autoRender = true;
```
dereuromark: Will this be a BC issue from components setting this on the controller?

robertpustulka: With this it should be ok:
https://github.com/robertpustulka/cakephp/blob/e2d894e65b7731b971a8ded78d42971bb32b7c27/src/Controller/Controller.php#L403-L407

tests/TestCase/Controller/ControllerTest.php
```diff
@@ -1086,7 +1086,121 @@ public function testBeforeRenderViewVariables()
...
+     * @return void
+     * @dataProvider deprecatedControllerPropertyProvider
+     */
+    public function testDeprecatedControllerProperty($property, $getter, $setter, $value)
+    {
+        $controller = new AnotherTestController();
+        $message = false;
+
+        set_error_handler(function ($errno, $errstr) use (&$message) {
+            $message = ($errno === E_USER_DEPRECATED ? $errstr : false);
```
saeideng: `$message` not used 
see my recent PR

robertpustulka: I'll wait with updating tests until #11231 is merged if that's what you mean.

saeideng: yes 
I'll update it tonight

robertpustulka:
The branch is up to date with `3.next`.

I added `plugin` getter and setter to the controller. Some tests seem to expect that the state of controller plugin and request plugin parameters may differ and I didn't want to break that behavior by delegating controller plugin property to the request entirely.

This could be refactored in 4.x so the `name` and `plugin` parameters would be accessed only via request.

tests/TestCase/Controller/ControllerTest.php
```diff
@@ -1115,6 +1115,166 @@ public function testBeforeRenderViewVariables()
...
+            $controller->{$setter}($value);
+            $result = $controller->$property;
+    }
+
+    /**
+     * Data provider for testing deprecated view properties
+     *
+     * @return array
+     */
+    public function deprecatedControllerPropertyProvider()
```
saeideng: if structure of this and `deprecatedViewPropertyProvider` is same
you can reduce test by adding content of this to `deprecatedViewPropertyProvider` 
so you can remove  bellow test IMO
```php
testDeprecatedControllerPropertyGetterMessage
testDeprecatedControllerPropertySetterMessage
testDeprecatedControllerProperty
```
also `expectedExceptionMessageRegExp /^Controller::\$\w+ is deprecated(.*)/` is same for 9 deprecated methods and this RegEx just check first of message (until `'Controller::$%s is deprecated.` )not all

saeideng: we no need to check all `'Controller::$%s is deprecated. Use $this->%s instead.'` message IMO

robertpustulka: Deprecated view properties point to the `$controller->viewBuilder()->method()` so this is a bit different.

saeideng: OK :+1:  Your are right 
I forget the `->viewBuilder()`

robertpustulka:
Can it be merged? I'd like to do the similar with `View` and `Cell` classes.

tests/TestCase/Controller/ControllerTest.php
```diff
@@ -1115,6 +1115,166 @@ public function testBeforeRenderViewVariables()
...
+     * @param $setter Setter name
+     * @param mixed $value Value to be set
+     * @return void
+     * @expectedException PHPUnit\Framework\Error\Deprecated
+     * @expectedExceptionMessageRegExp /^Controller::\$\w+ is deprecated(.*)/
+     * @dataProvider deprecatedControllerPropertyProvider
+     */
+    public function testDeprecatedControllerPropertySetterMessage($property, $getter, $setter, $value)
+    {
+        error_reporting(E_ALL);
```
saeideng: now
you should use `withErrorReporting()` for wrapping   `error_reporting()`
and `deprecated()` for testing deprecated methods

in here you must use `deprecated()`

src/Controller/Controller.php
```diff
@@ -386,17 +422,113 @@ public function __set($name, $value)
...
+        return $this;
+    }
+
+    /**
+     * Returns the plugin name.
+     *
+     * @return string|null
+     * @since 3.6.0
+     */
+    public function getPlugin()
```
ADmad: I don't think we should `getPlugin()`, `setPlugin()`. Just make people use `$this->request->getParam('plugin')` instead.

robertpustulka: I tried this but some tests were failing. It looked to me that there were some cases where different state od request plugin param and controller plugin name was expected.

ADmad: Okay, that's sad :disappointed:.

tests/TestCase/Controller/ControllerTest.php
robertpustulka: @saeideng tests have been updated

saeideng: nice :+1:

robertpustulka:
Can we merge this? I had to resolve conflicts like 3rd or 4th time already...

tests/TestCase/Controller/ControllerTest.php
```diff
@@ -1085,6 +1085,169 @@ public function testBeforeRenderViewVariables()
...
+     * @return void
+     * @expectedException PHPUnit\Framework\Error\Deprecated
+     * @expectedExceptionMessageRegExp /^Controller::\$\w+ is deprecated(.*)/
+     * @dataProvider deprecatedControllerPropertyProvider
+     */
+    public function testDeprecatedControllerPropertyGetterMessage($property, $getter, $setter, $value)
+    {
+        $controller = new AnotherTestController();
+        $controller->{$setter}($value);
+        $this->withErrorReporting(E_ALL, function () use ($controller, $property) {
```
markstory: The default error level is `E_ALL` why do you need to increase it here?

robertpustulka: Must be some out of date code. I'll fix that.

saeideng: default yes 
but we can change default in future if needed 
and in this case test must return  error/warning , so `E_ALL` is true IMO

robertpustulka: @markstory I think that @saeideng has a point here. What is your opinion on this?

markstory: I would prefer to keep our test suite at `E_ALL` and disable deprecation warnings as needed via `deprecated()`.

saeideng: so we no need to assert deprecation warnings in tests like other place

PR #10951, opened 2017-07-26
DefaultPasswordHasher with changed options enables needsRehash

jorisvaesen:
When changing the cost of bcrypt, the `needRehash` function should return `true`.


robertpustulka:
This needs to be patched in [Authentication plugin](https://github.com/cakephp/authentication/blob/master/src/PasswordHasher/DefaultPasswordHasher.php) as well i suppose.

ravage84:
@robertpustulka it seems it's done there already: https://github.com/cakephp/authentication/blob/master/src/PasswordHasher/DefaultPasswordHasher.php#L75

PR #10952, opened 2017-07-26
Add Validation::isScalar()

chinpei215:


tests/TestCase/Validation/ValidationTest.php
```diff
@@ -2879,6 +2879,18 @@ public function testIsArray()
...
 
     /**
+     * Test isScalar
+     *
+     * @return void
+     */
+    public function testIsScalar()
+    {
+        $this->assertTrue(Validation::isScalar(1));
+        $this->assertTrue(Validation::isScalar(''));
```
inoas: + `$this->assertTrue(Validation::isScalar(true));`

chinpei215: I am lazy 😛

src/Validation/Validator.php
```diff
@@ -1619,6 +1619,24 @@ public function isArray($field, $message = null, $when = null)
     }
 
     /**
+     * Add a validation rule to ensure that a field contains a scalar.
```
dereuromark: Should we add a small note that it means? That it would not be an object or array?

src/Validation/Validation.php
```diff
@@ -1424,6 +1424,17 @@ public static function isArray($value)
...
 
     /**
+     * Check that the input value is a scalar.
+     *
+     * @param mixed $value The value to check
+     * @return bool
+     */
+    public static function isScalar($value)
+    {
+        return is_scalar($value);
```
dereuromark: Please add a test for NULL, afaik NULL needs special handling as it is not "scalar" according to this check, but in our scope it should be OK to pass IMO.

chinpei215: I am find to add tests for nulls. But in my opinion, this method should not pass null, because null is not a scalar, and if I remember correctly, null values are not passed to validation methods and handled specially in Validator class.

src/Validation/Validator.php
chinpei215: I think scalar is scalar in PHP, and there is no ambiguity.

src/Validation/Validation.php
chinpei215: To be clear, I am talking about a case of `allowEmpty(true)`.

inoas: Imho it should NOT behave different than `is_scalar()`; But we can have `isScalarOrNull()` or `isScalarOrNullish()` (latter accepts '' and null)?

chinpei215: We don't have `isIntegerOrNull()`, and `isInteger()` doesn't pass null values. So I don't think we need to have `isScalarOrNull()`.

src/Validation/Validator.php
dereuromark: Validation rules are still high level and for a user just seeing that rule it might still be valuable not having to check the php internal manuals maybe :)

tests/TestCase/Validation/ValidationTest.php
inoas: but after the discussion below: `$this->assertFalse(Validation::isScalar(null));`... maybe ;)

src/Validation/Validator.php
chinpei215: OK, I can add a note.

tests/TestCase/Validation/ValidationTest.php
chinpei215: Probably I would add the test too.

chinpei215:
Oh I didn't notice that the proxy method for `Validation::isInteger()` is `Validator::integer()`. Should the proxy method for `Validation::isScalar()` be `Validator::scalar()`?

chinpei215:
What do you think about `Validation::isScalar()`? Most methods in Validation class doesn't have `is` prefix. Would `Validation::scalar()` and `Validator::scalar()` be better?

dereuromark:
wasnt the issue that integer is a reserved keyword?

chinpei215:
Though `array` is a reserved keyword, but `integer` and `scalar` are not. So we can name `Validator::integer()`. And in PHP7, we can use `array` for a method name without using `__call()` hack. So I think it is unecessary to care about reserved keywords if it is not reserved at this time.

saeideng:
`isScalar` is good naming

markstory:
:+1: for `isScalar()` as a method name for the validator.

chinpei215:
OK. Then I would rename `Validator::isScalar()` only.

PR #10955, opened 2017-07-27
Fix empty translations

### LINKED ISSUE #10954: Empty translations show as empty strings
This is a (multiple allowed):

* [x] bug
* [ ] enhancement

* CakePHP Version: 3.4.11

Strings with empty translations show as empty strings. It seems to be caused by PR #10597 since it adds a context parameter to all translations and the Translator class doesn't test for empty translations when there's a context parameter in the translation message.

I'm adding a PR with a test in short.

While doing this I realized there was an error in another test that checks empty translations with plurals although the test is succesful with or without the fix anyway.

While investigating this problem I've realized it's a waste of memory storing all those empty translations as arrays with empty contexts. I think the PO parser should discard those translations to save memory.

Also, while gettext tools ignore fuzzy translations, the PO parser does not. This can cause problems since wrong/incomplete translations may appear where they aren't supposed to. I've found a MO parser and I think using it should solve all these problems but I don't know how to select it or if that's discouraged for some reason.

berarma:
With the latest changes in the handling of contexts, empty translations had stopped working and returned an empty string. I've added a test and had a try at fixing it.

I don't think it's an efficient way to handle it. I would rather discard all empty translations when parsing the PO file or parse the MO file instead because the gettext formatter already discards them (and also fuzzy translations).

Still this is an easy fix for the regression.

Issue #10954.

### DIFF
--- a/src/I18n/Translator.php
+++ b/src/I18n/Translator.php
@@ -174,7 +174,7 @@ class Translator implements TranslatorInterface
         // No or missing context, fallback to the key/first message
         if ($context === null) {
             if (isset($message['_context'][''])) {
-                return $message['_context'][''];
+                return $message['_context'][''] === '' ? $key : $message['_context'][''];
             }
 
             return current($message['_context']);
--- a/tests/TestCase/I18n/I18nTest.php
+++ b/tests/TestCase/I18n/I18nTest.php
@@ -844,4 +844,16 @@ class I18nTest extends TestCase
         $this->assertEquals('Le moo', $translator->translate('Cow'));
         $this->assertEquals('Le bark', $translator->translate('Dog'));
     }
+
+    /**
+     * Tests the __() function on empty translations
+     *
+     * @return void
+     */
+    public function testEmptyTranslationString()
+    {
+        I18n::defaultFormatter('sprintf');
+        $result = __('No translation needed');
+        $this->assertEquals('No translation needed', $result);
+    }
 }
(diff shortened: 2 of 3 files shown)


dereuromark:
nice job!
thanks for taking care of this.

src/I18n/Translator.php
```diff
@@ -174,6 +174,9 @@ protected function resolveContext($key, $message, array $vars)
         // No or missing context, fallback to the key/first message
         if ($context === null) {
             if (isset($message['_context'][''])) {
+                if ($message['_context'][''] === '') {
```
lorenzo: Can you write this as

```php
if (context === null && isset($message['context'][''])) {
    return $message['_context'][''] === '' ? $key : $message['_context'][''];
}
```

It is starting to get confusing with all the nested `if` cases

lorenzo (changes requested):

src/I18n/Translator.php
berarma: Agree. That ``if``` has 3 ```return``` cases. I've changed it slightly differently.

src/I18n/Translator.php
```diff
@@ -174,9 +174,8 @@ protected function resolveContext($key, $message, array $vars)
         // No or missing context, fallback to the key/first message
         if ($context === null) {
             if (isset($message['_context'][''])) {
-                return $message['_context'][''];
+                return $message['_context'][''] === '' ? $key : $message['_context'][''];
             }
-
```
dereuromark: to fix cs you must revert this change. newline must stay.

berarma: Ugh, sorry, I didn't know.

markstory:
Thanks!

Issue #10128, opened 2017-01-30
RFC - Internal Usage of deprecated methods


markstory:
A big part of 3.4 has been to deprecate the combined getter/setter methods and replace those with separate `getX`, `setX` methods. The benefits of these new methods are:

1.  We will be able to more easily add strict typing in the future as each method has fewer possible return values.
2. Methods are simpler, have fewer parameters and modes which in theory leads to fewer issues due to reduced complexity.
3. IDEs can provide better auto-completion support.

While talking with @chinpei215 he brought up the fact that we should fully migrate internal usage of CakePHP to use the new un-deprecated methods *before* releasing 3.4.0. Doing this internal migration now would help ensure that any issues with the new methods are found earlier, and that users would only have to check their deprecated method usage once when they first upgrade, instead of at each release in 3.4.x

### Benefits of doing the work now

1. We can have simpler upgrade guidelines. We can recommend that users must update any deprecated methods that they have overloaded to use the new methods instead. Doing so will ensure that their code continues to work.
2. We can save work in the future as we won't have a large block of work when deprecations become runtime errors (tentatively planned for 3.6.0)

### Drawbacks of doing the work now

1. Will delay the release of 3.4.0 by a week or two.
2. We'll need to stick to the immediate internal upgrades for all future deprecations.

I'm looking to get some consensus on whether we want to finish the usage of undeprecated methods now, or in future 3.4.x releases.

#### Related issues

* #10086 
* #9978

chinpei215:
I vote we do the work now. Also I lend some hands.

markstory:
@chinpei215 I've started with `config()`, I'll have the pull request up tonight.

antograssiot:
I too think that we should upgrade them before the release. The reason is that we don't add tests for the methods and just implicitly rely on the old existing method tests to still pass but I remember that I found a/some issue with the new methods while upgrading some core code.

I can also help, what else do we have  except the configuration you are working on ?

On a side note, I like the idea of having a last 3.x serie release that will start trigger errors but I definitely prefer that we would have throw USER_DEPRECATED errors all the way like we did in 3.1 with the viewBuilder and add again a note in the migration guide to tell users how to hide those.  It would then allow us to provide a deprecated errors section in debug_kit for example and help users to upgrade their app gradually. I understand that this is too late for 3.4 though but that's something we can stick with for the 4.x serie...

ADmad:
[...]

PR #10956, opened 2017-07-27
Remove deprecated method calls

chinpei215:
Since we have already deprecated a large number of methods, I think we have to do the same work as #10128 again.


markstory:
Thanks!

PR #10957, opened 2017-07-27
Fixed bug causing requests with queries to be invalidated

jeremyharris:
The SecurityComponent would fail at _validatePost because the query
arguments were not encoded when the tokens were generated in the
IntegrationTestCase

I found this regression when one of my test cases failed during an upgrade.

edit: as I recall this was an issue brought up in slack that I couldn't reproduce because my app didn't have the latest version. I never did delve into it until it affected me, sorry!


markstory:
Nice find!

PR #10958, opened 2017-07-27
Deprecation cleanup

cleptric:
Update more deprecated method calls.

### DIFF
--- a/src/Controller/Component/PaginatorComponent.php
+++ b/src/Controller/Component/PaginatorComponent.php
@@ -282,7 +282,7 @@ class PaginatorComponent extends Component
      */
     public function config($key = null, $value = null, $merge = true)
     {
-        $return = $this->_paginator->config($key, $value, $merge);
+        $return = $this->_paginator->setConfig($key, $value, $merge);
         if ($return instanceof PaginatorInterface) {
             $return = $this;
         }
--- a/src/Controller/Component/PaginatorComponent.php
+++ b/src/Controller/Component/PaginatorComponent.php
@@ -282,7 +282,7 @@ class PaginatorComponent extends Component
      */
     public function config($key = null, $value = null, $merge = true)
     {
-        $return = $this->_paginator->setConfig($key, $value, $merge);
+        $return = $this->_paginator->config($key, $value, $merge);
         if ($return instanceof PaginatorInterface) {
             $return = $this;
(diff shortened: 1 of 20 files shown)


src/Controller/Component/PaginatorComponent.php
```diff
@@ -282,7 +282,7 @@ protected function _setPagingParams()
      */
     public function config($key = null, $value = null, $merge = true)
     {
-        $return = $this->_paginator->config($key, $value, $merge);
+        $return = $this->_paginator->setConfig($key, $value, $merge);
```
chinpei215: This change is incorrect, as `config()` is really used as a combined setter/getter here. See the following code in the method. If we want to remove this call, probably the fix will be like this:
```php
public function config($key = null, $value = null, $merge = true)
{
    if (is_array($key) || func_num_args() >= 2) {
        $this->_paginator->setConfig($key, $value, $merge);

        return $this;
    }

    return $this->_paginator->getConfig($key);
}
```
Though I touched this method a little bit in my PR (#10959), feel free to fix it. I will rebase my commit then.

src/ORM/ResultSet.php
```diff
@@ -185,7 +185,6 @@ public function __construct($query, $statement)
         $this->_useBuffering = $query->isBufferedResultsEnabled();
         $this->_defaultAlias = $this->_defaultTable->getAlias();
         $this->_calculateColumnMap($query);
-        $this->_calculateTypeMap();
```
markstory: Why is this being removed?

cleptric:
```
    /**
     * Creates a map of Type converter classes for each of the columns that should
     * be fetched by this object.
     *
     * @deprecated 3.2.0 Not used anymore. Type casting is done at the statement level
     * @return void
     */
    protected function _calculateTypeMap()
    {
    }
```

src/Controller/Component/PaginatorComponent.php
cleptric: You're absolutely right. Thanks 👍 I'll revert it for the mean time.

src/ORM/ResultSet.php
markstory: Ok.

markstory:
Nice work!

PR #10960, opened 2017-07-27
Loading all fixtures if no args are passed to loadFixtures

### LINKED ISSUE #10831: Add TestCase::loadAllFixtures() method
Suppose I'm using `$autoFixtures` as `false`, simply because I do not need fixtures to be loaded for all my class tests.

So I will use the `loadFixtures()` method whenever I need it, but often I manually load the same fixtures manually, which are all I have declared with the `$fixtures` property.

Something like this:

    class DriverTest extends TestCase
    {
        public $autoFixtures = false;

        public $fixtures = ['core.articles', 'core.comments'];

        public function testFirstMethod()
        {
            $this->loadFixtures('Articles', 'Comments');

            //... test
        }

        public function testSecondMethod()
        {
            $this->loadFixtures('Articles', 'Comments');

            //... test
        }
    }

So why not add a `loadAllFixtures()` method?

For now, because the `$this->fixtureManager->_fixtureMap` property is protected, I have this method:

    protected function loadAllFixtures()
    {
        $reflection = new \ReflectionClass($this->fixtureManager);
        $property = $reflection->getProperty('_fixtureMap');
        $property->setAccessible(true);
        $fixtures = $property->getValue($this->fixtureManager);
        $fixtures = array_keys($fixtures);

        call_user_func_array([$this, 'loadFixtures'], $fixtures);
    }

And this works. So I think it's enough to add a `getFixtureMap()` method and then the `loadAllFixtures()` method.

What do you think? Can I do a PR?

jeremyharris:
Closes #10831 

I based this on `3.next` as it's a behavior change.

### DIFF
--- a/src/TestSuite/TestCase.php
+++ b/src/TestSuite/TestCase.php
@@ -127,6 +127,7 @@ abstract class TestCase extends BaseTestCase
      * Chooses which fixtures to load for a given test
      *
      * Each parameter is a model name that corresponds to a fixture, i.e. 'Posts', 'Authors', etc.
+     * Passing no parameters will cause all fixtures on the test case to load.
      *
      * @return void
      * @see \Cake\TestSuite\TestCase::$autoFixtures
--- a/src/TestSuite/TestCase.php
+++ b/src/TestSuite/TestCase.php
@@ -141,6 +141,13 @@ abstract class TestCase extends BaseTestCase
         foreach ($args as $class) {
             $this->fixtureManager->loadSingle($class, null, $this->dropTables);
         }
+
+        if (empty($args)) {
+            $autoFixtures = $this->autoFixtures;
+            $this->autoFixtures = true;
+            $this->fixtureManager->load($this);
+            $this->autoFixtures = $autoFixtures;
+        }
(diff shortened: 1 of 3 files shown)


tests/Fixture/FixturizedTestCase.php
```diff
@@ -2,6 +2,7 @@
 namespace Cake\Test\Fixture;
 
 use Cake\TestSuite\TestCase;
+use Cake\ORM\TableRegistry;
```
jeremyharris: I can't spel

src/TestSuite/TestCase.php
```diff
@@ -141,6 +141,13 @@ public function loadFixtures()
         foreach ($args as $class) {
             $this->fixtureManager->loadSingle($class, null, $this->dropTables);
         }
+
+        if (empty($args)) {
+            $autoFixtures = $this->autoFixtures;
```
markstory: The doc block should be updated to include this usage.

jeremyharris: Done.

PR #10961, opened 2017-07-28
Output the command list when no command is given.

markstory:
Make `bin/cake` in CommandRunner work similar to how it works with ShellDispatcher. Output the help list when no command is given.

cc @robertpustulka @cleptric


HavokInspiration (approved):
Looks good. I restarted the failing build.

cleptric:
Looks good 🙂

PR #10968, opened 2017-07-31
Update Mysql.php

hekep:
CakePhp Mysql datasource could provide more debug information, when 'SQLSTATE[HY000]: General error: 1205 Lock wait timeout exceeded; try restarting transaction' occurs.

https://stackoverflow.com/questions/6000336/how-to-debug-lock-wait-timeout-exceeded-on-mysql

### DIFF
--- a/lib/Cake/Model/Datasource/Database/Mysql.php
+++ b/lib/Cake/Model/Datasource/Database/Mysql.php
@@ -245,6 +245,54 @@ class Mysql extends DboSource {
 		return $tables;
 	}
 
+	
+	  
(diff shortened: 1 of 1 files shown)


markstory (changes requested):
I'm not clear on what you're trying to solve here. The lack of tests would also need to be addressed as well.

markstory:
Closing as I am not comfortable merging this in. I feel it is not at the right layer and the lack of tests make this a feature I don't feel comfortable attempting to maintain in the future.

Issue #10964, opened 2017-07-30
Issues when loading MO/PO translation files


berarma:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.4.11

I have some projects where I use empty translation files.  My PO files my have some possible translations flagged fuzzy while the compliled MO files will be empty.

CakePHP loads the MO files then checks they're not empty or the PO file is loaded instead. This has 2 immediate consequences:
- it uses quite some memory due to lots of arrays used for empty translations.
- fuzzy translations become visible.

All of this wouldn't happen if there was at least one translation in the MO file. One translation up and the behavior changes significantly.

Is this the intended way to do it? If not I'm willing to work on it.

I would propose:
- Not ignoring empty MO files. Maybe check timestamps with PO files instead.
- Don't load empty translations in the PO file parser.
- Honoring the fuzzy flag in PO files so it's the same than loading corresponding MO files.

lorenzo:
It makes sense to not load the `po`, regardless of the contents of the `mo`.  Would you like to make that change? I really it should just check whether or not the `mo` file exists, if it does then it should not load the `po` at all

berarma:
I'll make a PR soon.

markstory:
Closing as a pull request is open now.

PR #10972, opened 2017-07-31
Don't ignore empty MO translation files

berarma:
Issue #10964.


src/I18n/MessagesFileLoader.php
```diff
@@ -126,7 +125,7 @@ public function __invoke()
         }
 
         if (!$file) {
-            return $package;
+            return false;
```
markstory: Will consumers of this class be ok with getting `false`? I think that's why the old code returned an empty `Package` instance.

tests/TestCase/I18n/MessagesFileLoaderTest.php
```diff
@@ -64,4 +64,15 @@ public function testCustomLocalePath()
...
+
+    /**
+     * Test reading MO files
+     * @return void
+     */
+    public function testMoFiles()
+    {
+        $loader = new MessagesFileLoader('empty', 'es', 'mo');
+        $package = $loader();
+        $this->assertNotNull($package);
```
markstory: `false` is not null too. Are there any other assertions we can make about this package object?

src/I18n/MessagesFileLoader.php
berarma: If that's true then I don't know how to deal with non-existing files. It's perfectly fine for a package to be empty.

tests/TestCase/I18n/MessagesFileLoaderTest.php
berarma: My bad. I should check for false.

src/I18n/MessagesFileLoader.php
markstory: Is the empty package causing the cascade to the PO file?

berarma: Yes. I need some way to differentiate when the file doesn't exists or doesn't have any translations.

berarma:
I've changed the assertions in the tests and updated the docblock for the modified method.

It doesn't seem the ```MessagesFileLoader``` class is used anywhere else than the ```TranslationRegistry``` class.

The ```ChainMessagesLoader``` class that invokes the class already does a fuzzy check for a false return value, thus I thought it would OK to use false as return value when the file doesn't exist. If a Package should always be expected then that check would be useless.

I could change the return value to null if that makes more sense.

markstory:
I took a look and you're right about the `!$package` check. Given that exists, returning false is reasonable.

PR #10974, opened 2017-08-01
Shell : improve unknown tokens error messages

HavokInspiration:
Replacement for PR #10941.

The aim remains the same : providing the user a more useful feedback by suggesting him the closest subcommand or option name to what he typed.
This PR also provides help for an unknown option or an unknown short option (with no "guess" for the latter for obvious reasons).

Compared to the previous pass, the error messages are now all targeting `stderr`.
I also improved the method in charge of guessing the user intent with some regex patterns before using the `levenshtein()` function. This provides better suggestions based on what the user typed and what tokens are available :
- It will first try to match tokens beginning with what the user typed (`boots` would match an option `bootstrap` before anything else). First match is returned.
- If nothing is found with the above logic, it will try to match tokens containing what the user typed (`boots` would match an option `clear-bootstrap`). First match is returned.
- the `levenshtein()` function is the last resort to suggest the best guess.

Outputs will be :

## Unknown command

```
$ bin/cake plugin lo
```
```
Unable to find the `plugin lo` subcommand. See `bin/cake plugin --help`.

Did you mean : `plugin load` ?

Available subcommands for the `plugin` command are : 

 - assets
 - load
 - loaded
 - unload
```

## Unknown option

```
$ bin/cake plugin load test --loa
```
``` 
Error: Unknown option `loa`.

Did you mean `autoload` ?

Available options are :

 - autoload
 - bootstrap
 - cli
 - help
 - quiet
 - routes
 - verbose
```

## Unknown short option

```
$ bin/cake plugin load test -a
```
```
Error: Unknown short option `a`

Available short options are :

 - `b` (short for `--bootstrap`)
 - `h` (short for `--help`)
 - `q` (short for `--quiet`)
 - `r` (short for `--routes`)
 - `v` (short for `--verbose`)
```

### DIFF
--- a/src/Console/ConsoleOptionParser.php
+++ b/src/Console/ConsoleOptionParser.php
@@ -879,13 +879,13 @@ class ConsoleOptionParser
     {
         $bestGuess = null;
         foreach ($haystack as $item) {
-            if (preg_match('/^' . $needle . '/', $item, $matches)) {
+            if (preg_match('/^' . $needle . '/', $item)) {
                 return $item;
             }
         }
 
         foreach ($haystack as $item) {
-            if (preg_match('/' . $needle . '/', $item, $matches)) {
+            if (preg_match('/' . $needle . '/', $item)) {
                 return $item;
             }
 
--- a/src/Console/ConsoleOptionParser.php
+++ b/src/Console/ConsoleOptionParser.php
@@ -731,6 +731,18 @@ class ConsoleOptionParser
      */
     public function help($subcommand = null, $format = 'text', $width = 72)
     {
+        if ($subcommand === null) {
+            $formatter = new HelpFormatter($this);
+            $formatter->setAlias($this->rootName);
+
+            if ($format === 'text') {
+                return $formatter->text($width);
+            }
+            if ($format === 'xml') {
+                return $formatter->xml();
+            }
+        }
+
         if (isset($this->_subcommands[$subcommand])) {
             $command = $this->_subcommands[$subcommand];
             $subparser = $command->parser();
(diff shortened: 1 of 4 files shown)


src/Console/Shell.php
```diff
@@ -549,6 +548,8 @@ protected function _displayHelp($command)
             $this->_welcome();
         }
 
+        $subcommands = $this->OptionParser->subcommands();
+        $command = isset($subcommands[$command]) ? $command : null;
```
HavokInspiration: This is necessary because unknown command will now trigger the new error message.
However, when you type `bin/cake plugin -h`, `-h` will be considered as a subcommand (but the parser will notice that the help is required, which is why we end up in the `_displayHelp()` method in the first place).

This change allows the `ConsoleOptionParser::help()` to display the good help message : if the subcommand exists, we'll display the subcommand's help, if not, we display the command general help and not the new error message.

HavokInspiration:
I added an example output in the original message

src/Console/ConsoleOptionParser.php
```diff
@@ -783,6 +787,125 @@ public function setRootName($name)
...
+     * @param string $option Unknown option name trying to be used.
+     * @return string The message to be displayed in the console.
+     */
+    protected function getOptionError($option)
+    {
+        $availableOptions = array_keys($this->_options);
+        $bestGuess = $this->findClosestItem($option, $availableOptions);
+        $out = [];
+        $out[] = sprintf('Unknown option `%s`.', $option);
+        $out[] = '';
```
markstory: You could initialize `$out` as an array with 2 elements in it.

src/Console/ConsoleOptionParser.php
```diff
@@ -783,6 +787,125 @@ public function setRootName($name)
...
+
+        foreach ($haystack as $item) {
+            if (preg_match('/' . $needle . '/', $item)) {
+                return $item;
+            }
+
+            $score = levenshtein($needle, $item);
+
+            if (!isset($bestScore) || $score < $bestScore) {
+                $bestScore = levenshtein($needle, $item);
```
markstory: You could do `$bestScore = $score` here.

HavokInspiration: It's not as if you already told me in the previous PR :smile:

HavokInspiration:
@markstory I took care of your comments

markstory:
Great work @HavokInspiration

Issue #10967, opened 2017-07-31
Ability to use OrmCache build/clear in non-shell contexts


josegonzalez:
This is a (multiple allowed):

* [x] enhancement

* CakePHP Version: 3.6

### What you did

Tried importing OrmCache shell functions to run programmatically.

### What happened

Unable to call OrmCache shell functions natively creating a shell instance.

### What you expected to happen

Ability to import OrmCache shell functions via a library or similar.

### Why is this useful?

You might wish to clear orm cache and rebuild it for use within a migration.

robertpustulka:
https://github.com/cakephp/cakephp/issues/10202#issuecomment-299044069 here we go again.

This comment and discussion should be seperated into it's own issue @dereuromark

markstory:
@robertpustulka Could we not refactor the shell logic out as people have a need to reuse it elsewhere?  While there is probably a larger discussion to be had around if we want to provide framework hooks for these 'service' classes. I think we could start small with individual shells for which there is a reuse case already known.

robertpustulka:
@markstory right, we should refactor the logic. But should we settle on a pattern for this first that we don't have to refactor the logic twice?

markstory:
@robertpustulka I'm not sure there is going to be a pattern to build though.

burzum:
@robertpustulka we could use events or have something like a ConsoleAware trait that implement a getter and setter for the console i/o objects?

markstory:
> We could use an event there but I think that's to heavy. Another option would be to implement isClearing() and isBuilding() and getCurrentTable() ...

I like this approach it enables integration in various contexts as it is unaware of the output environment (cli/http).

burzum:
@markstory I've started the work here https://github.com/cakephp/cakephp/tree/3.next-orm-cache It's basically done just the verbose message output is missing.

markstory:
Looks good. Perhaps `build()` can return the list of tables that had their cache rebuilt?

burzum:
@markstory @josegonzalez can we close this as the PR has been merged? Does it fulfil all your requirements?

markstory:
I think we're good here.

PR #10977, opened 2017-08-01
Decoupling the cache logic from the OrmCacheShell

burzum:
Refs #10967


src/ORM/OrmCache.php
```diff
@@ -0,0 +1,115 @@
...
+ * versions of your application are deployed, or when migrations
+ * requiring updated metadata are required.
+ */
+class OrmCache
+{
+
+    /**
+     * Schema
+     *
+     * @var \Cake\Datasource\SchemaInterface
```
ADmad: Shouldn't the type be same as those returned by `getSchema()` as it's return value is assigned to this property?

src/ORM/OrmCache.php
```diff
@@ -0,0 +1,115 @@
...
+        return $tables;
+    }
+
+    /**
+     * Helper method to get the schema collection.
+     *
+     * @param string $connection Connection name to get the schema for
+     * @return false|\Cake\Database\Schema\Collection|\Cake\Database\Schema\CachedCollection
+     */
+    public function getSchema($connection)
```
ADmad: Would be nice if the method also took `\Cake\Database\Connection` instance directly.

ADmad (changes requested):
Given that this class deals with only caching schema metadata wouldn't it be better to have it under `Database` instead?  It has nothing to do with ORM features and current name and location can be misleading.

`Cake\Database\SchemaCache.php` would be a better name and location for this IMO.

src/ORM/OrmCache.php
burzum: True.

src/ORM/OrmCache.php
```diff
@@ -90,17 +92,31 @@ public function clear($name = null)
...
-     * @param string $connection Connection name to get the schema for
-     * @return false|\Cake\Database\Schema\Collection|\Cake\Database\Schema\CachedCollection
+     * @param string|\Cake\Datasource\ConnectionInterface $connection Connection name to get the schema for or a connection instance
+     * @return \Cake\Database\Schema\Collection|\Cake\Database\Schema\CachedCollection
      */
     public function getSchema($connection)
     {
-        /* @var \Cake\Database\Connection $source */
-        $source = ConnectionManager::get($connection);
+        if (is_string($connection)) {
```
markstory: Do we need to support strings and instances?

burzum: @markstory @ADmad wanted it for some reason... Can we in the future find a process that somehow makes sure things aren't done and then removed? :(

ADmad: I don't mind if the method only took connection instance. That would allow typehinting the argument and removing checks. Since this class is not going to be used often in userland code I don't think having to do `new SchemaCache(ConnectionManager::get('default'))` would be too much of a inconvenience.

We could also make the constructor argument optional and use "default" connection if no argument is passed, as that's what's gonna be used in most cases anyway.

src/Shell/OrmCacheShell.php
```diff
@@ -26,118 +22,10 @@
  * can prevent thundering herd effects on the metadata cache when new
  * versions of your application are deployed, or when migrations
  * requiring updated metadata are required.
+ *
+ * @deprecated Use \Cake\Shell\SchemaCacheShell instead
```
saeideng: from 3.5??

src/Shell/SchemaCacheShell.php
```diff
@@ -0,0 +1,119 @@
...
+ */
+class SchemaCacheShell extends Shell
+{
+
+    /**
+     * Schema Cache
+     *
+     * @var \Cake\Database\SchemaCache
+     */
+    protected $_schemaCache;
```
dereuromark: this seems unused

src/ORM/OrmCache.php
markstory: I agree with @ADmad. This is not a class that many developers will need to interact with. It doesn't need to support a variety of input types.

ADmad:
The docblock and typehint for constructor should also be updated isn't?  And the unused `use` for `ConnectionManager` needs to be removed.

ADmad:
Tests failing on PHP 5.6

Issue #10975, opened 2017-08-01
Remove cookie from Response object


jorisvaesen:
This is a (multiple allowed):

* [ ] bug
* [x] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.5.0-RC1

### What you did
I wanted to delete a cookie using the new CookieCollection.

### What happened
```
$cookies = $this->request->getCookieCollection();
if ($cookies->has('remember_me')) {
    $cookies = $cookies->remove('remember_me');
}
$this->response = $this->response->withCookieCollection($cookies);
```

Although the method exists in the ServerRequest object, the  `withCookieCollection()` method doesn't seem to exist in the Response object yet.

### What you expected to happen
The response with the modified cookie collection

markstory:
It might be easier to use if the method was `$response->withoutCookie('remember_me');`

jorisvaesen:
Indeed pretty straight forward.

cleptric:
Thanks for taking care of this @jorisvaesen 🙂 
Closing as we have a PR.

PR #10981, opened 2017-08-02
Response withExpiredCookie

jorisvaesen:
Refs #10975

### DIFF
--- a/src/Http/Cookie/Cookie.php
+++ b/src/Http/Cookie/Cookie.php
@@ -510,7 +510,7 @@ class Cookie implements CookieInterface
     public function withExpired()
     {
         $new = clone $this;
-        $new->expiresAt = Chronos::parse('-1 year');
+        $new->expiresAt = Chronos::createFromTimestamp(1);
 
         return $new;
     }
--- a/src/Http/Response.php
+++ b/src/Http/Response.php
@@ -19,7 +19,6 @@ use Cake\Filesystem\File;
 use Cake\Http\Cookie\Cookie;
 use Cake\Http\Cookie\CookieCollection;
 use Cake\Http\Cookie\CookieInterface;
-use Cake\I18n\Time;
 use Cake\Log\Log;
 use Cake\Network\CorsBuilder;
 use Cake\Network\Exception\NotFoundException;
@@ -2084,7 +2083,7 @@ class Response implements ResponseInterface
             $cookie = new Cookie(
                 $name,
                 '',
-                new Time(1),
+                DateTime::createFromFormat('U', 1),
                 $options['path'],
                 $options['domain'],
                 $options['secure'],
--- a/src/Http/Response.php
+++ b/src/Http/Response.php
@@ -19,6 +19,7 @@ use Cake\Filesystem\File;
 use Cake\Http\Cookie\Cookie;
 use Cake\Http\Cookie\CookieCollection;
 use Cake\Http\Cookie\CookieInterface;
+use Cake\I18n\Time;
 use Cake\Log\Log;
 use Cake\Network\CorsBuilder;
 use Cake\Network\Exception\NotFoundException;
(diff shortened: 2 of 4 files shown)


src/Http/Response.php
```diff
@@ -2041,6 +2042,66 @@ public function withCookie($name, $data = '')
...
+     * $response = $response->withoutCookie('remember_me');
+     *
+     * // customize cookie attributes
+     * $response = $response->withoutCookie('remember_me', ['path' => '/login']);
+     *
+     * // add a cookie object
+     * $response = $response->withoutCookie(new Cookie('remember_me', 'deleted'));
+     * ```
+     *
+     * @param string|\Cake\Http\Cookie\Cookie $name The name of the cookie to expire, or a cookie object
```
robertpustulka: Should be `string|\Cake\Http\Cookie\CookieInterface`.

src/Http/Response.php
```diff
@@ -2041,6 +2042,66 @@ public function withCookie($name, $data = '')
...
+     * @param array $options An array of cookie options.
+     * @return static
+     */
+    public function withoutCookie($name, $options = [])
+    {
+        if ($name instanceof Cookie) {
```
robertpustulka: `$name instanceof CookieInterface`

src/Http/Response.php
```diff
@@ -2041,6 +2042,66 @@ public function withCookie($name, $data = '')
...
+     *
+     * // add a cookie object
+     * $response = $response->withoutCookie(new Cookie('remember_me', 'deleted'));
+     * ```
+     *
+     * @param string|\Cake\Http\Cookie\Cookie $name The name of the cookie to expire, or a cookie object
+     * @param array $options An array of cookie options.
+     * @return static
+     */
+    public function withoutCookie($name, $options = [])
```
robertpustulka: `array` type hint would be nice.

markstory: Is this method 'removing' a cookie from the response that is in flight or adding an expired cookie? Right now `withoutCookie()` will add a cookie which is a bit counter-intuitive.

src/Http/Response.php
```diff
@@ -2041,6 +2042,66 @@ public function withCookie($name, $data = '')
...
+        if ($name instanceof Cookie) {
+            $cookie = $name->withExpiry(new Time(1));
+        } else {
+            $options += [
+                'value' => '',
+                'path' => '/',
+                'domain' => '',
+                'secure' => false,
+                'httpOnly' => false
+            ];
```
markstory: If the cookie exists in the collection, would it make sense to read the options out of that cookie object?

tests/TestCase/Http/ResponseTest.php
```diff
@@ -1491,6 +1491,53 @@ public function testWithCookieObject()
...
+            'secure' => true,
+            'httpOnly' => true,
+            'expire' => (string)strtotime('+14 days'),
+        ];
+
+        $response = new Response();
+        $response = $response->withCookie('testing', $options);
+        $this->assertEquals($options, $response->getCookie('testing'));
+
+        $new = $response->withoutCookie('testing', $options);
```
markstory: expire as an option to this method should probably trigger an error, as should 'value'.  At the very least those options should be ignored. The test doesn't cover the value being ignored.

src/Http/Response.php
```diff
@@ -2041,6 +2042,66 @@ public function withCookie($name, $data = '')
     }
 
     /**
+     * Create a new response with an expired cookie set.
+     *
+     * ### Options
+     *
+     * - `name`: The Cookie name
+     * - `value`: Value of the cookie
```
markstory: Value shouldn't be an option. We can hardcode the value to ''.

src/Http/Response.php
jorisvaesen: @markstory Removing means calling `remove` on the CookieCollection and reapply it.
Nonetheless, I agree that this naming isn't very convenient but the CookieComponent is working the same where `delete` has the same functionality. How about `withExpiredCookie()`?

markstory: I like `withExpiredCookie()`. It better describes what this method is doing.

src/Http/Response.php
```diff
@@ -19,6 +19,7 @@
 use Cake\Http\Cookie\Cookie;
 use Cake\Http\Cookie\CookieCollection;
 use Cake\Http\Cookie\CookieInterface;
+use Cake\I18n\Time;
```
robertpustulka: I'm wondering if there's a need to depend on `I18n`?

src/Http/Cookie/Cookie.php
```diff
@@ -510,7 +510,7 @@ public function withNeverExpire()
     public function withExpired()
     {
         $new = clone $this;
-        $new->expiresAt = Chronos::parse('-1 year');
+        $new->expiresAt = Chronos::createFromTimestamp(1);
```
markstory: Why? -1 year is good enough to expire cookies.

src/Http/Response.php
```diff
@@ -2058,26 +2056,25 @@ public function withCookie($name, $data = '')
...
      * @param array $options An array of cookie options.
      * @return static
      */
-    public function withoutCookie($name, $options = [])
+    public function withExpiredCookie($name, $options = [])
     {
-        if ($name instanceof Cookie) {
-            $cookie = $name->withExpiry(new Time(1));
+        if ($name instanceof CookieInterface) {
+            $cookie = $name->withExpired();
```
markstory: :+1:

src/Http/Response.php
```diff
@@ -2058,26 +2056,25 @@ public function withCookie($name, $data = '')
...
         } else {
             $options += [
-                'value' => '',
                 'path' => '/',
```
markstory: Should this default to `$this->getAttribute('webroot')`?

src/Http/Cookie/Cookie.php
jorisvaesen: For me, it's not a worse option than `-1 year` (and looks like the safest and most consistent) and is now the same as how the CAKEPHP-cookie gets expired.

src/Http/Response.php
markstory: Ignore this. My suggestion is not possible.

Issue #10979, opened 2017-08-02
Slow DboSource::truncate


bancer:
This is a (multiple allowed):

* [ ] bug
* [x] enhancement
* [x] feature-discussion (RFC)

* CakePHP Version: 2.9.9.
* Platform and Target: MySQL, Apache, Linux.

### What you did
Run unit tests for a single controller with 73 fixtures and 35 tests using xdebug to generate cachegrind file.

### What happened
Cachegrind reports that `DboSource::truncate` method is quite slow using 28% of tests time.

![screenshot from 2017-08-02 11 27 59](https://user-images.githubusercontent.com/3830106/28867414-5ad8e150-7776-11e7-9825-7e44e6613241.png)

### What you expected to happen
I would expect that truncating tables would not be so slow. Is there a good explanation of that or a way to improve the performance of truncating tables?

dereuromark:
First of all: truncating like this is usually more a debugging tool and not something for productive use.
It is a good way of resetting some local things, but for production, one would usually use the DB directly :)

Second: Did you do this with debug 0 and cache enabled?
Because otherwise your results are void. Only when using production cache for ORM you can actually the the relevant things going on :)
And even then this method really is not relevant IMO (and also always O:1 and a non loop operation which makes it insignificant).

bancer:
Yes, this is not for production. The problem is that unit tests are quite slow. Using 1/4 of unit tests time to truncate tables looks odd.

dereuromark:
Then don't truncate like this :) Use SQL command for it directly.
But I would still consider this a non-issue.

bancer:
Well, it's not our code that calls `truncate()`. It's called by CakePHP in CakeFixtureManager.

AD7six:
@bancer I think you're going to find it'ss not that `truncate` is slow, it's that calling it 7829 times simply adds up. I doubt making truncate faster (even if that were possible) would make much difference.

Therefore:

 * Is it real-world slow, or just the % that caught your attention? how many ms is 27.37%?
 * Is truncate being called unnecessarily by fixture logic?
 * Are fixtures being over used?

Probably worth mentioning that (over)reliance on fixtures leads to unit tests that aren't actually unit tests.

bancer:
> Is it real-world slow, or just the % that caught your attention? how many ms is 27.37%?

[...]

PR #10985, opened 2017-08-02
Improved performance of CakeFixtureManager

bancer:
Improved performance of CakeFixtureManager. See https://github.com/cakephp/cakephp/issues/10979


lib/Cake/Test/Case/Model/Datasource/Database/MysqlTest.php
```diff
@@ -3701,6 +3704,7 @@ public function testFieldsWithComplexVirtualFields() {
  * @return void
  */
 	public function testExecute() {
+		$this->loadFixtures('Article');
```
markstory: Why do these tests methods need to load fixtures now?

lib/Cake/TestSuite/Fixture/CakeFixtureManager.php
```diff
@@ -229,32 +231,43 @@ public function load(CakeTestCase $test) {
...
+		$dropDataSources = array();
 		$fixtures = !empty($test->fixtures) ? $test->fixtures : array();
 		foreach (array_reverse($fixtures) as $f) {
 			if (isset($this->_loaded[$f])) {
 				$fixture = $this->_loaded[$f];
 				if (!empty($fixture->created)) {
 					foreach ($fixture->created as $ds) {
 						$db = ConnectionManager::getDataSource($ds);
-						$fixture->truncate($db);
+						if ($test->dropTables) {
```
markstory: Is there a way we can avoid having 5 levels of indentation?

markstory (commented):
The number of tests that needed to change make me concerned that this isn't a completely safe change, and that it will result in breaking other apps.

lib/Cake/TestSuite/Fixture/CakeFixtureManager.php
bancer: Yes, I can flatten that code.

lib/Cake/Test/Case/Model/Datasource/Database/MysqlTest.php
bancer: The value of `$autoFixtures` in MysqlTest is set to `false`. According to the [documentation](https://api.cakephp.org/2.9/class-CakeTestCase.html#$autoFixtures) the fixtures must be loaded manually here:
```
/**
  * By default, all fixtures attached to this class will be truncated and reloaded after each test.
  * Set this to false to handle manually
  *
  * @var array
  */
public $autoFixtures = true;
```
So the missing call to `loadFixtures` here is actually a bug. I believe that the code worked without loading fixtures manually because the fixtures were not properly dropped in `CakeFixtureManager::unload` as it is promissed in the [documentation](https://api.cakephp.org/2.9/class-CakeTestCase.html#$dropTables):
```
/**
  * Control table create/drops on each test method.
  *
  * Set this to false to avoid tables to be dropped if they already exist
  * between each test method. Tables will still be dropped at the
  * end of each test runner execution.
  *
  * @var bool
  */
public $dropTables = true;
```

ravage84:
I agree with @markstory but I would prefer if we could fix the issue reported by @bancer nonetheless.

bancer:
I would need help of Posgres expert. Does anybody have an idea how to fix this error in the failed build?
```
PHP Fatal error:  Uncaught exception 'PDOException' with message 'SQLSTATE[3B001]: Invalid savepoint specification: 7 ERROR:  no such savepoint' in /home/travis/build/cakephp/cakephp/lib/Cake/Model/Datasource/DboSource.php:2489
Stack trace:
#0 /home/travis/build/cakephp/cakephp/lib/Cake/Model/Datasource/DboSource.php(2489): PDO->exec('ROLLBACK TO SAV...')
#1 /home/travis/build/cakephp/cakephp/lib/Cake/Model/Datasource/DboSource.php(2471): DboSource->_rollbackNested()
#2 /home/travis/build/cakephp/cakephp/lib/Cake/Model/Datasource/DataSource.php(430): DboSource->rollback()
#3 /home/travis/build/cakephp/cakephp/lib/Cake/Model/Datasource/DboSource.php(3721): DataSource->__destruct()
#4 [internal function]: DboSource->__destruct()
#5 {main}
  thrown in /home/travis/build/cakephp/cakephp/lib/Cake/Model/Datasource/DboSource.php on line 2489
```
screenshot from 2017-08-04 00-07-34

lorenzo:
@bancer have you tried a solution like this? https://github.com/lorenzo/cakephp-fixturize

I found that approach to be a lot faster than Normal cake fixtures, with no need to change how the cor works internally.

lorenzo:
You could also use something like this https://github.com/FriendsOfCake/fixturize/blob/master/src/TestSuite/Fixture/ChecksumTestFixture.php

The fixture class will check whether the table was changed at all at the end of the test, saving you the need to truncate it most of the times

bancer:
@lorenzo https://github.com/lorenzo/cakephp-fixturize seems to require changes to the existing fixtures. Not a very good option for us. https://github.com/FriendsOfCake/fixturize looks much better but seems to be for CakePHP 3. Would it be possible to publish another version for CakePHP 2?

lorenzo:
@bancer that file would work with very minor changes in Cake 2, starting with removing the connection annotations. Try doing that and I can help you publish it in my plugin if you open a pull request

markstory:
What kind of performance increase are you seeing with these changes? I ask because it is a somewhat risky change, so the benefit should be reasonably high.

dereuromark:
I agree, this looks very dangerous and seems to break basically every test (requires $this->loadFixtures() added). The benefit here, especially for test only scope, seems rather small, and compared to the fallout not acceptable it seems.
We should probably close here.

markstory:
Closing as I'm too gun shy to add what looks like a risky change this late in 2.x

Issue #10948, opened 2017-07-26
Table object is 'false' when baking underscore separated table


GusAntoniassi:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.4.11
* Platform and Target: 
Apache/2.4.26 (Win32) OpenSSL/1.0.2l PHP/7.1.7
Windows NT 10.0 build 15063 (Windows 10) i586
libmysql - mysqlnd 5.0.12-dev - 20150407
Running under XAMPP v3.2.2

### What you did

1. Started a fresh project with `composer self-update && composer create-project --prefer-dist cakephp/app {project}`
2. Created the following test database:
```
CREATE TABLE `underscore_tests` (
  `id` int(11) NOT NULL,
  `col` varchar(255) NOT NULL
) ENGINE=InnoDB DEFAULT CHARSET=utf8;
ALTER TABLE `underscore_tests`
  ADD PRIMARY KEY (`id`);
ALTER TABLE `underscore_tests`
  MODIFY `id` int(11) NOT NULL AUTO_INCREMENT;
```
3. Updated `app.php` with database configuration data
4. Ran bake with `cake bake all underscore_tests`

### What happened
On `UnderscoreTestsController.php`, object `$this->UnderscoreTests` has the value of `false`. Trying to access `localhost/{project}/underscoretests` gives the error "Unable to locate an object compatible with paginate.", likely because of the query `$this->paginate($this->UnderscoreTests);` with `$this->UnderscoreTests` having the value of `false`.
Trying to access the object `$this->Underscoretests` works fine and debugging it returns the following:
```
object(App\Model\Table\UnderscoreTestsTable) {

	'registryAlias' => 'Underscoretests',
	'table' => 'underscore_tests',
	'alias' => 'Underscoretests',
	'entityClass' => 'App\Model\Entity\UnderscoreTest',
	'associations' => [],
	'behaviors' => [],
	'defaultConnection' => 'default',
	'connectionName' => 'default'

}
```
I am not sure if this is a problem with CakePHP's core or the Bake utility. Running `cake bake all` on console returns me:
```
Welcome to CakePHP v3.4.11 Console
---------------------------------------------------------------
App : src
Path: C:\xampp\htdocs\cake-bug\cake-bug\src\
PHP : 7.1.7
---------------------------------------------------------------
Bake All
---------------------------------------------------------------
Possible model names based on your database:
- underscore_tests
[...]

PR #10987, opened 2017-08-03
Add more clear error messaging for incorrect URL inflection

markstory:
Update the missing controller error page in an attempt to make it clearer when a developer is using the wrong inflection type in their URLs.

screen shot 2017-08-02 at 21 54 39

Refs #10948


HavokInspiration (approved):
Always +1 on better error messages !

PR #10991, opened 2017-08-03
New Rule : IsNotReferencedBy

half2me:
For associations using `ON DELETE RESTRICT` for foreign keys, this Rule checks for any references and does not allow deleting if there is a row in the specified table referencing the entity.


tests/TestCase/ORM/RulesCheckerIntegrationTest.php
```diff
@@ -490,6 +490,28 @@ public function testIsUniqueMultipleFieldsOneIsNull()
...
+    {
+        $articlesTable = TableRegistry::get('Articles');
+        $articlesTable->belongsTo('Authors');
+        $authorsTable = TableRegistry::get('Authors');
+
+        $author = $authorsTable->get(1);
+        $authorWithoutArticles = $authorsTable->get(2);
+
+        $rules = $authorsTable->rulesChecker();
+        $rules->addDelete($rules->IsNotReferencedBy('Articles'));
```
markstory: I don't see where this method is defined.

half2me: @markstory I would request some help with that. Is it not enough to create the Rule Class?

inoas: You need to add a proxy method to `\Cake\ORM\RulesChecker`;

inoas: You need to add a proxy method to `\Cake\ORM\RulesChecker` such as `public function isNotReferencedBy(array $tableOrAlias?, $message = null)`.

Also note that most (or all all) of the rules work on fields, so I would like to see that you can explicitly specify bindingKey and foreignKey (as array or array|string):

`public function isNotReferencedBy(array $tableOrAlias?, $message = null, $options)`

half2me: @inoas its not enough to auto-detect the binding fields?

inoas: @half2me first make it work with that proxy method - that's probably mandatory to merge (I can't speak for the core).

I do really like configurable binding and foreign keys. They can get you around all the complex situations. It is not that you would need to declare them, only have the option, to. If you don't they are inferred by the association like you already do in your code.

markstory: @half2me You would need to add a method to ORM\RulesChecker. You can use the isUnique() method there as an example.

half2me: @ionas I agree, it would add more flexibility. I will implement those then.

half2me: @markstory just a quick question here, it seems to me that with each Rule added, we need to add a method to the RulesChecker. (Seems like this might be boilerplate code) Would it be preferable to modify the internals of RulesChecker to automatically load rules from the `Cake\ORM\Rule` namespace? It could be implemented easily as a single magic method.

markstory: Magic methods make documentation and typehinting a bit harder. We could factor common code into methods if there is enough of it.

src/ORM/RulesChecker.php
```diff
@@ -113,6 +114,26 @@ public function existsIn($field, $table, $message = null)
...
+     * ```
+     * $rules->add($rules->isNotReferencedBy('Articles'));
+     * ```
+     *
+     * @param string $table The table name to check for references
+     * @return callable
+     */
+    public function isNotReferencedBy($table)
+    {
+        return $this->_addError(new IsNotReferencedBy($table), '_isNotReferencedBy', compact('errorField', 'message'));
```
markstory: Won't message and errorField always be null here?

half2me: @markstory yeah, I just copied the code from the other rules, and wasn't quite sure what this part did, and how it did it

src/ORM/RulesChecker.php
```diff
@@ -113,6 +114,26 @@ public function existsIn($field, $table, $message = null)
...
+     * ### Example:
+     *
+     * ```
+     * $rules->add($rules->isNotReferencedBy('Articles'));
+     * ```
+     *
+     * @param string $table The table name to check for references
+     * @return callable
+     */
+    public function isNotReferencedBy($table)
```
robertpustulka: Can this be renamed to `$association` since the check is made against the association?

src/ORM/Rule/IsNotReferencedBy.php
```diff
@@ -0,0 +1,63 @@
...
+     * @var array
+     */
+    protected $_repository;
+
+    /**
+     * Constructor.
+     *
+     * @param string $repository The association name for the repository which will be checked.
+     */
+    public function __construct($repository)
```
robertpustulka: Core [rules support](https://github.com/cakephp/cakephp/blob/master/src/ORM/Rule/ExistsIn.php#L80-L94) `Association` and `Table` objects as well in such cases, not only association name strings.

src/ORM/RulesChecker.php
markstory: Perhaps errorField and message should be parameters/keys in an options array?

tests/TestCase/ORM/RulesCheckerIntegrationTest.php
half2me: @markstory of course I understand. Just seemed to me, like there was a lot of redundant code, which would grow with each Rule added. (Doesn't seem to me though, like anyone is planning on adding extra rules anytime soon :) )

src/ORM/RulesChecker.php
half2me: @robertpustulka I was using `public function existsIn($field, $table, $message = null)` as a base for mine. There it is called a table, and what `isNotReferencedBy` does, is very similar, just in reverse. I could rename it, but then wouldn't this need renaming too?

tests/TestCase/ORM/RulesCheckerIntegrationTest.php
markstory: We accrue them pretty slowly. After this one, it could be worthwhile to see if we can find a pattern, extract it and abstract it.

src/ORM/Rule/IsNotReferencedBy.php
```diff
@@ -0,0 +1,82 @@
...
+     *
+     * @param \Cake\Datasource\EntityInterface $entity The entity to check against
+     * @param array $options Options passed to the check,
+     * where the `repository` key is required.
+     * @throws \RuntimeException When the rule refers to an undefined association.
+     * @return bool
+     */
+    public function __invoke(EntityInterface $entity, array $options)
+    {
+        if (is_string($options))
```
markstory: Conditions need braces.

src/ORM/RulesChecker.php
```diff
@@ -113,6 +114,47 @@ public function existsIn($field, $table, $message = null)
...
+            if ($this->_useI18n) {
+                $message = __d('cake', 'This field is referenced by another entity');
+            } else {
+                $message = 'This field is referenced by another entity';
+            }
+        }
+
+        $errorField = is_string($field) ? $field : current($field);
+
+        return $this->_addError(new IsNotReferencedBy($field, $table, $options), '_isNotReferencedBy', compact('errorField', 'message'));
```
markstory: Any way we could get tests for setting custom messages & error fields?

markstory (changes requested):

src/ORM/RulesChecker.php
half2me: @markstory yes I'm not finished with this PR yet, just didn't have time to finish up yet.

dereuromark:
@half2me Are you still interested in finishing this PR?

half2me:
@dereuromark yes, unfortunately haven't had much time to work on this, but I think this is something which would come in handy. If all that is missing are unit tests, I might ask for some help though...

dereuromark:
@half2me 3.7 is soon to be released as stable. I think now would be time to finish this up if you want to land this into that release.

half2me:
> @half2me 3.7 is soon to be released as stable. I think now would be time to finish this up if you want to land this into that release.

I'll try to finish it up soon, would be nice to have it included in 3.7.

src/ORM/Rule/IsNotReferencedBy.php
```diff
@@ -0,0 +1,83 @@
...
+     *
+     * @param \Cake\Datasource\EntityInterface $entity The entity to check against
+     * @param array $options Options passed to the check,
+     * where the `repository` key is required.
+     * @throws \RuntimeException When the rule refers to an undefined association.
+     * @return bool
+     */
+    public function __invoke(EntityInterface $entity, array $options)
+    {
+        if (is_string($options)) {
```
ADmad: How can options be a string when the argument is typehinted as array?

src/ORM/Rule/IsNotReferencedBy.php
```diff
@@ -0,0 +1,83 @@
...
+            $assoc = $options['repository']->association($this->_repository);
+        }
+
+        if (!$assoc) {
```
ADmad: This will throw an error of $assoc is not set since the variable is not initialized.

src/ORM/Rule/IsNotReferencedBy.php
half2me: @ADmad do you think the `fields` option should be optional? It could default to the field(s) used in the foreign key.

half2me:
> ping @half2me You still want to finish this into 3.8?

Yes

dereuromark:
We will close this if no further activity is happening here. After 2 years maybe this isn't that useful.

half2me:
> We will close this if no further activity is happening here. After 2 years maybe this isn't that useful.

I think I'll just put it into a plugin. Thanks though

PR #10994, opened 2017-08-03
Address CakeEmail regression when data is defined with no mimetype

watermark:
Prior to 2.10.x, CakeEmail allowed adding an attachment using only "data" with no "mimtype" defined (which defaulted the mimetype to "application/octet-stream").  Starting in 2.10.0, a "Undefined index: file" is triggered when an attachment is added using "data" with no "mimetype".


markstory:
There are a few [phpcs](https://travis-ci.org/cakephp/cakephp/jobs/260781844) errors.

markstory (approved):
Looks good. Once the build is passing I'll get this merged.

markstory:
Thanks!

PR #10997, opened 2017-08-04
Attempt to get the session tests passing in PHP7.2

markstory:
In PHP7.2 ini_set() cannot modify session configuration once headers have been 'sent'. In the CLI SAPI, headers are 'sent' as soon as STDOUT gets written to. By running most of the session tests in separate processes we can get around most of the limitations. But I had to change a few of the methods to suppress errors in every other test.

This doesn't _fix_ the PHP7.2 builds but gets us a bunch closer.

### DIFF
--- a/src/Network/Session.php
+++ b/src/Network/Session.php
@@ -108,7 +108,9 @@ class Session
             $sessionConfig['ini']['session.name'] = $sessionConfig['cookie'];
         }
 
-        if (!empty($sessionConfig['handler'])) {
+        // In PHP7.1.0+ session.save_handler can't be set to user by the user.
+        // https://github.com/php/php-src/blob/master/ext/session/session.c#L559
+        if (!empty($sessionConfig['handler']) && version_compare(PHP_VERSION, '7.1.0', '<=')) {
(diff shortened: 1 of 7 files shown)


ADmad:
:clap:

markstory:
Only 24 errors left. I will do another chunk tonight.

jeremyharris:
Nice work!

src/Network/Session.php
```diff
@@ -108,7 +108,9 @@ public static function create($sessionConfig = [])
             $sessionConfig['ini']['session.name'] = $sessionConfig['cookie'];
         }
 
-        if (!empty($sessionConfig['handler'])) {
+        // In PHP7.1.0+ session.save_handler can't be set to user by the user.
+        // https://github.com/php/php-src/blob/master/ext/session/session.c#L559
+        if (!empty($sessionConfig['handler']) && version_compare(PHP_VERSION, '7.1.0', '<=')) {
```
inoas: So it really/also fixes tests passing on `^7.1`?

markstory: No in 7.1 this simply did nothing from what I could figure out of the php source code.

PR #10998, opened 2017-08-05
Add support for default value to InstanceConfigTrait::getConfig().

ADmad:
`InstanceConfigTrait::getConfig()` was missing the default value convenience which has been added to various other getters like `Hash::get()`, `Configure::read()` etc.


markstory:
:+1: for consistency

PR #11000, opened 2017-08-05
Improved performance of CakeFixtureManager.

bancer:
See https://github.com/cakephp/cakephp/issues/10979


markstory:
Does this replace #10985?

bancer:
No. This is a partial performance improvement from https://github.com/cakephp/cakephp/pull/10985. It truncates tables only when it is really necessary by skipping the truncation right after the tables were created. It will give a little but still performance improvement.

Issue #11003, opened 2017-08-06
ContextArray doesn't handle nesting very well


lilHermit:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.4.x

When using ContextArray for modelless forms nested element are not supported well. This is the case of schema, defaults and required.

For example, content array requires you add the extra array which means setting defaults/required/schema needs to be done on all elements

```php
        $schema = [
            'schema' => [
                'users' => [
                    ['age' => ['type' => 'integer']]
                    ]
            ],
            'required' => [
                'users' => [[
                    'age' => true,
                ]],
            ],
            'defaults' => [
                'users' => [
                   ['age'=>'21']
                ],
            ]
        ];

$content = new ArrayContext(Request::createFromGlobals(), $schema);
debug($content->type('users.0.age'));
debug($content->isRequired('users.0.age'));
debug($content->val('users.0.age'));

// outputs
// 'integer'
// true
// '21'
```
but for `users.1.age` you would need the following schema

```php
$schema = [
            'schema' => [
                'users' => [
                    ['age' => ['type' => 'integer']],
                    ['age' => ['type' => 'integer']]
                    ]
            ],
            'required' => [
                'users' => [
                     [ 'age' => true],
                     [ 'age' => true],
                 ],
            ],
            'defaults' => [
                'users' => [
                   ['age'=>'21'],
                   ['age'=>'21'],
                ],
            ]
        ];
```

I propose that contentArray methods `type`, `isRequired` and `val` either remove the `.0`, .`1` or fall back to removing it. That way you could use schema like follows

```php
$schema = [
            'schema' => [
                'users' => [ 
                    'age' => ['type' => 'integer']
                  ]
            ],
            'required' => [
                'users' => [
                      'age' => true
                 ],
            ],
            'defaults' => [
                'users' => [
                   'age'=>'21'
                ],
            ]
        ];
```

ADmad:
Sounds reasonable to me if your proposal can be implemented in a backwards compatible way.

lilHermit:
@ADmad Yeah we can either try non-nested first and fallback to the current implementation or if currently implementation fails then fallback to non-nested. I'll create a PR

markstory:
Closing as a pull request is open.

PR #11004, opened 2017-08-06
Added support for numeric nested fields type, isRequired, val and att…

lilHermit:
…ributes

    Updated tests too
See #11003

### DIFF
--- a/src/View/Form/ArrayContext.php
+++ b/src/View/Form/ArrayContext.php
@@ -180,9 +180,9 @@ class ArrayContext implements ContextInterface
         // Using Hash::check here incase the default value is actually null
         if (Hash::check($this->_context['defaults'], $field)) {
             return Hash::get($this->_context['defaults'], $field);
-        } else {
-            return Hash::get($this->_context['defaults'], $this->stripNesting($field));
         }
+
+        return Hash::get($this->_context['defaults'], $this->stripNesting($field));
     }
 
     /**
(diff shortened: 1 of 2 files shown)


src/View/Form/ArrayContext.php
```diff
@@ -177,7 +177,12 @@ public function val($field, $options = [])
             return null;
         }
 
-        return Hash::get($this->_context['defaults'], $field);
+        // Using Hash::check here incase the default value is actually null
+        if (Hash::check($this->_context['defaults'], $field)) {
+            return Hash::get($this->_context['defaults'], $field);
+        } else {
```
ADmad: `else` is not required since there is a `return` in `if` block.

src/View/Form/ArrayContext.php
```diff
@@ -273,4 +288,17 @@ public function error($field)
 
         return Hash::get($this->_context['errors'], $field);
     }
+
+    /**
+     * Strips out any numeric nesting like users.0.age
+     *
```
ADmad: Please add an example of input and output for easier understanding.

ADmad (changes requested):

lilHermit:
Do you think the following it more efficient?!

```php
protected function stripNesting($field)
{
    return preg_replace('/\.\d/','',$field);
}
```

markstory:
@lilHermit It does look much simpler.

lilHermit:
@markstory definitely simpler but my concern was (and the reason I used implode/explode) keys starting with numeric will break for example `users.0.9tags` becomes `userstags`

I've have now tweaked the regex to fix this

PR #11006, opened 2017-08-06
Add ContextFactory class.

ADmad:
This offloads the code for generating context instance based
on data passed to FormHelper::create() into separate class.

### DIFF
--- a/src/View/Form/ContextFactory.php
+++ b/src/View/Form/ContextFactory.php
@@ -31,7 +31,7 @@ class ContextFactory
      *
      * @var array
      */
-    protected $contextProviders = [];
+    protected $providers = [];
 
     /**
      * Constructor.
@@ -111,8 +111,8 @@ class ContextFactory
      */
     public function addProvider($type, callable $check)
     {
-        $this->contextProviders = [$type => ['type' => $type, 'callable' => $check]]
-            + $this->contextProviders;
+        $this->providers = [$type => ['type' => $type, 'callable' => $check]]
+            + $this->providers;
     }
 
     /**
@@ -130,7 +130,7 @@ class ContextFactory
     {
         $data += ['entity' => null];
 
-        foreach ($this->contextProviders as $provider) {
+        foreach ($this->providers as $provider) {
             $check = $provider['callable'];
             $context = $check($request, $data);
             if ($context) {
--- a/src/View/Form/ContextFactory.php
+++ b/src/View/Form/ContextFactory.php
@@ -38,17 +38,60 @@ class ContextFactory
      *
      * @param array $providers Array of provider callables. Each element should
      *   be of form `['type' => 'a-string', 'callable' => ..]`
-     * @param bool $addDefaults Whether default providers should be added.
      */
-    public function __construct(array $providers = [], $addDefaults = true)
+    public function __construct(array $providers = [])
     {
-        if ($addDefaults) {
-            $this->addDefaultProviders();
-        }
-
         foreach ($providers as $provider) {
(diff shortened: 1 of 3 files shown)


src/View/Form/ContextFactory.php
```diff
@@ -0,0 +1,143 @@
...
+    protected $contextProviders = [];
+
+    /**
+     * Constructor.
+     *
+     * @param array $providers Array of provider callables. Each element should
+     *   be of form `['type' => 'a-string', 'callable' => ..]`
+     * @param bool $addDefaults Whether default providers should be added.
+     */
+    public function __construct(array $providers = [], $addDefaults = true)
```
markstory: Should adding the defaults be done in a static factory method?

src/View/Form/ContextFactory.php
```diff
@@ -0,0 +1,143 @@
...
+            if ($context) {
+                break;
+            }
+        }
+        if (!isset($context)) {
+            $context = new NullContext($request, $data);
+        }
+        if (!($context instanceof ContextInterface)) {
+            throw new RuntimeException(
+                'Context objects must implement Cake\View\Form\ContextInterface'
```
markstory: Should this say what we got instead of the correct type?

src/View/Form/ContextFactory.php
ADmad: You mean the `addDefaultProviders()` method should be made static?

src/View/Form/ContextFactory.php
ADmad: Okay, will update error message to mention type received too.

src/View/Form/ContextFactory.php
markstory: More that `ContextFactory` could have a method like `ContextFactory::createWithDefaults()`.

ADmad:
Updated as per feedback.

antograssiot (approved):

src/View/Form/ContextFactory.php
```diff
@@ -98,46 +141,13 @@ public function get(ServerRequest $request, array $data = [])
             $context = new NullContext($request, $data);
         }
         if (!($context instanceof ContextInterface)) {
-            throw new RuntimeException(
-                'Context objects must implement Cake\View\Form\ContextInterface'
-            );
+            throw new RuntimeException(sprintf(
+                'Context providers must return object implementing %s. Got "%s" instead.',
+                'Cake\View\Form\ContextInterface',
```
markstory: Should this be `ContextInterface::class`?

ADmad:
@markstory You are probably looking at outdated diff. It's already `ContextInterface::class`.

src/View/Form/ContextFactory.php
```diff
@@ -0,0 +1,153 @@
...
+     * data providers that come from outside CakePHP. For example
+     * if you wanted to use an alternative ORM like Doctrine you could
+     * create and connect a new context class to allow FormHelper to
+     * read metadata from doctrine.
+     *
+     * @param string $type The type of context. This key
+     *   can be used to overwrite existing providers.
+     * @param callable $check A callable that returns an object
+     *   when the form context is the correct type.
+     * @return void
```
markstory: Should this method return `$this` to make it easier to add multiple providers?

inoas: Imho generally `add*()` and `set*()` should unless we use return types instead of exceptions for results of the "adding" or "setting".

ADmad:
@markstory Done

PR #11008, opened 2017-08-07
More PHP7.2 Fixes

markstory:
Another batch of PHP7.2 fixes. We should be pretty close after this. In my local docker environment, I only had a few failures related to filesystem permissions issues, which could be caused by docker-machine.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ php:
   - 7.0
   - 5.6
   - 7.1
-  - nightly
+  - 7.2
 
 dist: trusty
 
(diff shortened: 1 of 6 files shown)


antograssiot (approved):

ADmad:
Only this weird error for `RedisEngine` seems to remain https://travis-ci.org/cakephp/cakephp/jobs/261689574#L704

markstory:
@ADmad Those redis failures are super strange. In all cases the variables *do* exist. Not sure why PHP7.2 thinks they don't.

PR #11009, opened 2017-08-07
Run tests against 7.2 not 7.3.0-dev

antograssiot:
travis nigthly is using 7.3.0-dev
see https://travis-ci.org/cakephp/cakephp/jobs/261692749#L509 for example

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ php:
   - 7.0
   - 5.6
   - 7.1
-  - nightly
+  - 7.2
 
 dist: trusty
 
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -4,7 +4,7 @@ php:
   - 7.0
   - 5.6
   - 7.1
-  - nightly
+  - 7.2
```
HavokInspiration: Shouldn't we keep `nightly` alongside with 7.2 ?

ADmad: @HavokInspiration It's too early to start worrying about 7.3 just yet :)

ADmad:
I cherry picked your commit onto branch `php72-fixes`

.travis.yml
HavokInspiration: As you folks see fit.

Issue #11012, opened 2017-08-07
EntityTrait::setDirty($property, false) should return $this


alysson-azevedo:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.4.12.

`EntityTrait::setDirty($property, false)` returns `false` instead `$this`.
https://github.com/cakephp/cakephp/blob/master/src/Datasource/EntityTrait.php#L756

I'm creating this issue because i don't have time to write a PR with tests.

markstory:
Pull request up now.

PR #11014, opened 2017-08-08
Fix return value of setEntity()

markstory:
It should return $this as documented. Expand test coverage as there weren't any explicit tests for the new undeprecated methods.

Refs #11012


antograssiot (approved):

inoas:
hmhm... I would have put this into 3.5.

markstory:
@inoas I don't want to set that precedent. The original implementation was wrong, and fixing it is a 'bugfix' level fix. We fix lots of bugs that people could relying on.

inoas:
@markstory point taken.

Issue #11002, opened 2017-08-06
hiddenField for radio input not set correctly


jaykay-design:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 2.9.7 (also present in 2.10.0)
* Platform and Target: All

### What you did
echo $this->Form->input('some_name', array('type' => 'radio', 'options' => $options, 'hiddenField' => 'some value');

### What happened
The hidden field is not rendered with the value "some value" as documented here: https://book.cakephp.org/2.0/en/core-libraries/helpers/form.html#options-for-select-checkbox-and-radio-inputs

### What you expected to happen
Have the hidden field's value attribute set with the provided value.

Change line 1672+ in FormHelper (2.10.0) from:
```
		if ($hiddenField) {
			if (!isset($value) || $value === '') {
				$hidden = $this->hidden($fieldName, array(
					'form' => isset($attributes['form']) ? $attributes['form'] : null,
					'id' => $attributes['id'] . '_',
					'value' => '',
					'name' => $attributes['name']
				));
			}
		}
```
to
```
		if ($hiddenField) {
			if (!isset($value) || $value === '') {
				$hidden = $this->hidden($fieldName, array(
					'form' => isset($attributes['form']) ? $attributes['form'] : null,
					'id' => $attributes['id'] . '_',
					'value' => $hiddenField === true ? '' : $hiddenField,
					'name' => $attributes['name']
				));
			}
		}
```

markstory:
Pull request up now.

PR #11021, opened 2017-08-09
Fix hiddenField option not working for radio (2.x)

markstory:
The hiddenField option was not working as documented for radio buttons. Instead of using the provided value, the hidden input's value was hardcoded to ''

Refs #11002


dereuromark:
looks good

Issue #11019, opened 2017-08-08
StaticConfigTrait::parseDsn() returns string if parse_url() fails.


sitedyno:
This is a (multiple allowed):

* [ ] bug
* [ ] enhancement
* [x] feature-discussion (RFC)

* CakePHP Version: 3.5.0-RC1
* Platform and Target: Ubuntu 16.04, Apache, MySQL

### What you did
I used a database password that has a `#` in it. Similar to `mysql://user:#word@localhost/my_db`

### What happened
I can access the database from the command line just fine, but the app throws:
```
Warning (2): Illegal string offset 'className' [CORE/src/Datasource/ConnectionManager.php, line 117]
```
This is because `parse_url()` will not parse a URL with a `#` in it, at least in the password section. This results in `StaticConfigTrait::parseDsn()` returning a string instead of an array and `ConnectionManager` expects an array.

### What you expected to happen
It would have been really nice to have been told the DSN couldn't be parsed by `parse_url()`, then I would have looked into that instead of digging through the core to see what's going on. `parse_url()` specifically mentions that it's not meant for dealing with URIs:

    Note:

    This function is intended specifically for the purpose of parsing URLs and not URIs. However, to comply with PHP's backwards compatibility requirements it makes an exception for the file:// scheme where triple slashes (file:///...) are allowed. For any other scheme this is invalid.

The `#` in password does work if configured in app.php.

Not being able to use `#` in a db password with environment variables is also a limitation on security. A small limitation, but a limitation nonetheless.

markstory:
Pull request up now.

PR #11024, opened 2017-08-10
Replace parse_url() with a regex when parsing DSN strings.

markstory:
parse_url() fails on passwords that contain # and other special URL characters. I've re-implemented a DSN parser with a regular expression. It handles all the DSN flavours in the core tests, and should be transparent to userland code.

Refs #11019


tests/TestCase/Core/StaticConfigTraitTest.php
```diff
@@ -17,53 +17,6 @@
 use Cake\TestSuite\TestCase;
 
 /**
- * TestConnectionManagerStaticConfig
```
josegonzalez: Why the test move?

markstory: Moving the tests let me remove the stub class which had code duplicated from `ConnectionManager`.

josegonzalez: Fair enough.

tests/TestCase/Datasource/ConnectionManagerTest.php
```diff
@@ -285,6 +285,165 @@ public function testParseDsn()
...
+        $expected = [
+            'className' => 'Cake\Database\Connection',
+            'database' => 'database',
+            'driver' => 'Cake\Database\Driver\Mysql',
+            'host' => 'localhost',
+            'password' => 'pas#][{}$%20@!',
+            'port' => 3306,
+            'scheme' => 'mysql',
+            'username' => 'user',
+            'log' => 1,
```
lorenzo: Should this not be converted to Boolean?

AD7six (approved):
Changes as they are look fine

I think it would be a good idea to explicitly test parseDsn https://github.com/cakephp/cakephp/blob/master/src/Core/StaticConfigTrait.php#L242

The existing tests only cover that indirectly.

Opinion on using data providers? If I were to write this test right-now, each test with multiple asserts in it would be converted to one test with one assert, and a dataprovider for the input/expected/fail message.

Happy to action these test-style comments as a follow up.

src/Core/StaticConfigTrait.php
```diff
@@ -249,18 +249,23 @@ public static function parseDsn($dsn)
...
+        $pattern = '/^(?P<scheme>[\w\\\\]+):\/\/((?P<user>.*?)(:(?P<password>.*?))?@)?' .
+            '((?P<host>[.\w\\\\]+)(:(?P<port>\d+))?)?' .
+            '(?P<path>\/[^?]*)?(\?(?P<query>.*))?$/';
+        preg_match($pattern, $dsn, $parsed);
 
-        $parsed = parse_url($dsn);
-        if ($parsed === false) {
-            return $dsn;
+        if (empty($parsed)) {
+            return false;
```
sitedyno: What's the use case for returning a boolean here? The return line in the doc-block specifies only array. Since this is ?only? used for configuration wouldn't you expect some kind of error if the DSN you've passed can't be parsed?

tests/TestCase/Datasource/ConnectionManagerTest.php
markstory: I think only 'true' is cast to a bool.

src/Core/StaticConfigTrait.php
markstory: That is a good point. I will update the code to raise an error.

markstory:
@AD7six I agree about the data provider approach. I will update the tests.

markstory:
Updated.

src/Core/StaticConfigTrait.php
```diff
@@ -249,18 +249,23 @@ public static function parseDsn($dsn)
             throw new InvalidArgumentException('Only strings can be passed to parseDsn');
         }
 
-        $scheme = '';
-        if (preg_match("/^([\w\\\]+)/", $dsn, $matches)) {
-            $scheme = $matches[1];
-            $dsn = preg_replace("/^([\w\\\]+)/", 'file', $dsn);
-        }
+        $pattern = '/^(?P<scheme>[\w\\\\]+):\/\/((?P<user>.*?)(:(?P<password>.*?))?@)?' .
+            '((?P<host>[.\w\\\\]+)(:(?P<port>\d+))?)?' .
```
chinpei215: Since `\w` doesn't match `-`, the regular expression for the host part (and the schema part?) should be more relaxed.

src/Core/StaticConfigTrait.php
```diff
@@ -249,18 +249,23 @@ public static function parseDsn($dsn)
...
+            '(?P<path>\/[^?]*)?(\?(?P<query>.*))?$/';
+        preg_match($pattern, $dsn, $parsed);
 
-        $parsed = parse_url($dsn);
-        if ($parsed === false) {
-            return $dsn;
+        if (empty($parsed)) {
```
chinpei215: `(!$parsed)` would be better than `empty($parsed)`.

markstory:
@chinpei215 Thanks for the feedback. Updated now.

chinpei215:
Thanks. Looks good to me.

garas:
Before this update, `redis:?prefix=myapp:_cake_core_:` used to work.

I had to change it to `redis://?prefix=myapp:_cake_core_:` to work again.

josegonzalez:
Yeah, `redis:` isn't a proper scheme, so that makes sense. `parse_url` can be a bit more tolerant of input, which I don't think is the best way forward here.

If you feel otherwise, feel free to open an issue to relax the input validation for schemes.

Issue #10992, opened 2017-08-03
Query::deselect() method


berrygoudswaard:
This is a (multiple allowed):

* [ ] bug
* [ ] enhancement
* [x] feature-discussion (RFC)

* CakePHP Version: > 3.x

### Problem
Sometimes I want the select all columns from a table, except for a few. Right now I am using the Query::select() method to select all the columns I want, but it would be convenient to have a Query::deselect()

### Proposal
By default CakePHP uses the autoField trigger to select all columns.
When a developer calls the Query::select, it will that disable autoField and select only the fields that are passed to the Query::select() method.

My idea is to implement a Query::deselect, that will use the autoFields functionality but removes a few columns from the columns-list.

The following example will select all columns of the articles table, except 'modified' and 'created'
```php
$query = $articles
    ->find()
    ->deselect(['modified', 'created'])
    ->where(['id !=' => 1])
    ->order(['created' => 'DESC']);
```

I could create a PR for this, but I would like to know if this functionality will be accepted in the CakePHP core.

lorenzo:
I'm not sure that it is a method that naturally belongs in the `Query` object, but it can maybe be a helper function.

for example this should work:

```php
$query = $articles
    ->find()
    ->select(array_diff($articles->schema()->columns(), ['created', 'modified']))
    ->where(['id !=' => 1])
    ->order(['created' => 'DESC']);
```

We could make this a helper function

```php
function allExcept($table, $fields) {
   return array_diff($table->schema()->columns(), $fields);
}
```

And then you could call

```php
$query = $articles
    ->find()
    ->select(Db\selectAllExcept($articles, ['created', 'modified']))
    ->where(['id !=' => 1])
    ->order(['created' => 'DESC']);
```

burzum:
@lorenzo can't this simply be `->selectAllExcept($articles, ['created', 'modified']))` and `selectAllExcept(['created', 'modified'])`?

lorenzo:
I'm not comfortable putting it in the `Query` object because then it will need to interact with all the state it has internally:

* What happens if the columns are already selected?
* How does the interact with `autoFields`?

berrygoudswaard:
Autofields is disabled, when you use Query::select(), so I can imagine it does the same with Query::selectAllExcept()

I think that Query::selectAllExcept() could also have a $overwrite parameter (defaults to true). This way the developer can decide what happens if a column is already selected.

Signature would be something like:
```php
[...]

PR #11025, opened 2017-08-10
Add selectAllExcept method

berrygoudswaard:
Added Query::selectAllExcept(), which makes it possible to select all the fields in a schema, with the exception of the one passed to the method. 

Also see issue #10992

### DIFF
--- a/src/ORM/Query.php
+++ b/src/ORM/Query.php
@@ -1277,8 +1277,6 @@ class Query extends DatabaseQuery implements JsonSerializable, QueryInterface
     }
 
     /**
-     * {@inheritDoc}
-     *
      * Selects all fields in the schema with the exception of the fields you pass
      * in the $excludedFields parameter.
      *
@@ -1293,6 +1291,6 @@ class Query extends DatabaseQuery implements JsonSerializable, QueryInterface
         $aliasedExcluded = $this->aliasFields($excludedFields, $this->repository()->alias());
 
         $fields = array_diff($aliased, $aliasedExcluded);
-        return parent::select($fields, $overwrite);
+        return $this->select($fields, $overwrite);
     }
 }
--- a/src/ORM/Query.php
+++ b/src/ORM/Query.php
@@ -1275,4 +1275,24 @@ class Query extends DatabaseQuery implements JsonSerializable, QueryInterface
 
         return $result;
     }
+
+    /**
+     * {@inheritDoc}
+     *
+     * Selects all fields in the schema with the exception of the fields you pass
+     * in the $excludedFields parameter.
+     *
+     * @param array $excludedFields fields you don't want to select
+     * @param bool $overwrite whether to reset fields with passed list or not
+     * @return $this
+     */
+    public function selectAllExcept($excludedFields = [], $overwrite = false)
+    {
+        $allFields = $this->repository()->getSchema()->columns();
+        $aliased = $this->aliasFields($allFields, $this->repository()->alias());
+        $aliasedExcluded = $this->aliasFields($excludedFields, $this->repository()->alias());
+
+        $fields = array_diff($aliased, $aliasedExcluded);
+        return parent::select($fields, $overwrite);
+    }
 }
--- a/src/ORM/Query.php
+++ b/src/ORM/Query.php
@@ -1284,7 +1284,7 @@ class Query extends DatabaseQuery implements JsonSerializable, QueryInterface
      * @param bool $overwrite whether to reset fields with passed list or not
      * @return $this
      */
-    public function selectAllExcept($excludedFields = [], $overwrite = false)
+    public function selectAllExcept(array $excludedFields = [], $overwrite = false)
     {
         $allFields = $this->repository()->getSchema()->columns();
         $aliased = $this->aliasFields($allFields, $this->repository()->alias());
--- a/tests/TestCase/ORM/QueryTest.php
+++ b/tests/TestCase/ORM/QueryTest.php
@@ -3466,4 +3466,28 @@ class QueryTest extends TestCase
 
         $this->assertEquals(1, $query->__debugInfo()['decorators'], 'Only one typecaster should exist');
     }
+
+    /**
+     * Tests that selectAllExcept() selects all columns, except
+     * the one that are passed.
+     *
+     * @return void
+     */
+    public function testSelectAllExcept()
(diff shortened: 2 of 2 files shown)


src/ORM/Query.php
```diff
@@ -1275,4 +1275,24 @@ protected function _decorateResults($result)
 
         return $result;
     }
+
+    /**
+     * {@inheritDoc}
```
robertpustulka: Are there any docs to inherit?

src/ORM/Query.php
```diff
@@ -1275,4 +1275,24 @@ protected function _decorateResults($result)
...
+     * @return $this
+     */
+    public function selectAllExcept($excludedFields = [], $overwrite = false)
+    {
+        $allFields = $this->repository()->getSchema()->columns();
+        $aliased = $this->aliasFields($allFields, $this->repository()->alias());
+        $aliasedExcluded = $this->aliasFields($excludedFields, $this->repository()->alias());
+
+        $fields = array_diff($aliased, $aliasedExcluded);
+        return parent::select($fields, $overwrite);
```
dereuromark: Shouldnt this be

    $this->select()

?

src/ORM/Query.php
berrygoudswaard: Nope, you are right.

src/ORM/Query.php
berrygoudswaard: You are absolutely right, I have changed it.

src/ORM/Query.php
```diff
@@ -1275,4 +1275,22 @@ protected function _decorateResults($result)
...
+
+    /**
+     * Selects all fields in the schema with the exception of the fields you pass
+     * in the $excludedFields parameter.
+     *
+     * @param array $excludedFields fields you don't want to select
+     * @param bool $overwrite whether to reset fields with passed list or not
+     * @return $this
+     */
+    public function selectAllExcept($excludedFields = [], $overwrite = false)
```
dereuromark: Should $excludedFields be array typehinted?

berrygoudswaard: Yes, thank you for your feedback

tests/TestCase/ORM/QueryTest.php
```diff
@@ -3466,4 +3466,28 @@ public function testAllNoDuplicateTypeCasting()
...
+        $selectedFields = $result->clause('select');
+        $expected = [
+            'articles__id' => 'articles.id',
+            'articles__author_id' => 'articles.author_id',
+            'articles__title' => 'articles.title',
+            'articles__published' =>'articles.published'
+        ];
+
+        $this->assertEquals($expected, $selectedFields);
+    }
```
lorenzo: Missing a few tests showing how this interacts with `autoFields` and `contain`

src/ORM/Query.php
```diff
@@ -1275,4 +1275,22 @@ protected function _decorateResults($result)
...
+
+    /**
+     * Selects all fields in the schema with the exception of the fields you pass
+     * in the $excludedFields parameter.
+     *
+     * @param array $excludedFields fields you don't want to select
+     * @param bool $overwrite whether to reset fields with passed list or not
+     * @return $this
+     */
+    public function selectAllExcept(array $excludedFields = [], $overwrite = false)
```
lorenzo: This should receive an instance of `Table` `Association` or `TableSchema` as first argument

src/ORM/Query.php
```diff
@@ -1275,4 +1275,22 @@ protected function _decorateResults($result)
...
+    {
+        $allFields = $this->repository()->getSchema()->columns();
```
lorenzo: Use the passed argument instead of getting the repository here.

src/ORM/Query.php
```diff
@@ -1275,4 +1275,22 @@ protected function _decorateResults($result)
...
+        $aliased = $this->aliasFields($allFields, $this->repository()->alias());
```
lorenzo: Why not do the `array_diff` before?

lorenzo (changes requested):

src/ORM/Query.php
berrygoudswaard: I need the aliased fields otherwise it won't work if I pass aliased fields to as $excludedFields

src/ORM/Query.php
lorenzo: Another thing, there should be no default for $excludeFielda

berrygoudswaard: @lorenzo If I change the first param to an Association or TableSchema, it's no longer possible to do something like:

```php
$this->Articles->find()
   ->selectAllExcept(['modified', 'created']);
```

Then it becomes something like:

```php
$this->Articles->find()
   ->selectAllExcept($this->Articles, ['modified', 'created']);
```
In this case the Table parameter is redundant. 

Maybe we should not make the Table/Association the first param, but an optional parameter.

```php
$this->Articles->find()
   ->selectAllExcept(['modified', 'created'], $this->Articles);
```

src/ORM/Query.php
berrygoudswaard: See response above......

If the developer passes an Association or Table as second parameter, we use that param, otherwise we fallback on ```$this->respository()```

tests/TestCase/ORM/QueryTest.php
```diff
@@ -3466,4 +3466,28 @@ public function testAllNoDuplicateTypeCasting()
...
+     * the one that are passed.
+     *
+     * @return void
+     */
+    public function testSelectAllExcept()
+    {
+        $table = TableRegistry::get('articles');
+        $result = $table
+            ->find()
+            ->selectAllExcept(['body']);
```
markstory: Would be good to have a test covering how this method is expected to work when it is called multiple times.

src/ORM/Query.php
lorenzo: Yes, I understood that, but I think the fallback is not a great idea. If you call this method after calling join() or contain()!it would read as you are selecting all from all tables. But if we make the first param the table object there can be no confusion

markstory:
Moving to 3.6 as I think this is a big enough change to warrant more soak time.

dereuromark:
@berrygoudswaard Are you interested in finishing this up?

markstory:
Closing in favour of #11784

Issue #7499, opened 2015-10-01
Intermittent crashes when connecting to SQL server


MCF:
I am seeing intermittent crashes when connecting to SQL server.  There are no PHP errors thrown, the sqlsrv driver seems to be crashing.  Here is the information from the Windows event logs:

```
Faulting application name: httpd.exe, version: 2.4.16.0, time stamp: 0x55a22a64
Faulting module name: php_pdo_sqlsrv_56_ts.dll, version: 3.2.0.0, time stamp: 0x54f4b646
Exception code: 0xc0000005
Fault offset: 0x00014926
Faulting process id: 0x854
Faulting application start time: 0x01d0fc91fac8f4fc
Faulting application path: C:\Apache24\bin\httpd.exe
Faulting module path: C:\PHP\ext\php_pdo_sqlsrv_56_ts.dll
Report Id: 3f966aa0-6885-11e5-be5e-180373b9ebbe
```

I am using:

Windows 7
PHP 5.6.13
Apache 2.4.16
CakePHP 2.7.3
Microsoft's sqlsrv pdo driver version 3.2

The crashes happen when making a connection to the database.  More precisely At this line:

https://github.com/cakephp/cakephp/blob/2.7.3/lib/Cake/Model/Datasource/Database/Sqlserver.php#L135

As mentioned previously the problem is intermittent and highly variable.  It is not easy to reproduce as a result.

The crash information does seem similar to this older issue: https://github.com/cakephp/cakephp/issues/2003

As suggested in that older issue I've set the database connection to `persistent => false` in an attempt to solve the problem.  So far it seems to have fixed it, although I've only spent an hour testing the legacy application I am upgrading to 2.7.3.  I will continue to test and report back in a few days.

I have not looked in detail at what the persistent flag is affecting with respect to the sqlsrv connection.  Any comments on what this is doing (or if it is actually doing anything) are welcome.

markstory:
If the sqlsrv driver code ia failing, there is not much we can really do.

ravage84:
Are you trying to create a ``'persistent'`` connection?
If so change that to ``false``.

Have you tried to use the non-thread safe DLL instead?

ravage84:
May be it has to do with the fact that the sqlsrv driver relies on connection pooling:

[...]

PR #11028, opened 2017-08-11
Remove the 'persistent' config option for Sqlserver connections

MCF:
- The PDO::ATTR_PERSISTENT connection configuration option is not allowed
  by the SQL Server PHP driver.  In newer versions of the driver an
  exception is thrown if that option is set.  In older versions of the
  driver it triggered a heap corruption bug which usually caused
  intermittent crashes.

This is a fix for an issue reported several years ago: https://github.com/cakephp/cakephp/issues/7499

At the time a note was added to the documentation to avoid setting persistent to true with SQL Server.  Subsequently the PHP SQL Server driver maintainers have made changes to the driver (as of version 4.1.8) to throw an exception if persistent is set to true.  The persistent option should be ignored for SQL Server connections in all cases.

Please see the PHP SQL Server driver issue for more information: https://github.com/Microsoft/msphpsql/issues/65

### DIFF
--- a/src/Database/Driver/Sqlserver.php
+++ b/src/Database/Driver/Sqlserver.php
@@ -35,7 +35,6 @@ class Sqlserver extends Driver
      * @var array
      */
     protected $_baseConfig = [
-        'persistent' => false,
         'host' => 'localhost\SQLEXPRESS',
         'username' => '',
         'password' => '',
@@ -54,8 +53,14 @@ class Sqlserver extends Driver
     ];
 
     /**
-     * Establishes a connection to the database server
+     * Establishes a connection to the database server.
      *
+     * Please note that the PDO::ATTR_PERSISTENT attribute is not supported by
+     * the SQL Server PHP PDO drivers.  As a result you cannot use the
+     * persistent config option when connecting to a SQL Server  (for more
+     * information see: https://github.com/Microsoft/msphpsql/issues/65).
+     *
+     * @throws \InvalidArgumentException if an unsupported setting is in the driver config
      * @return bool true on success
      */
     public function connect()
@@ -64,8 +69,12 @@ class Sqlserver extends Driver
             return true;
         }
         $config = $this->_config;
+
+        if (isset($config['persistent']) && $config['persistent']) {
+            throw new \InvalidArgumentException('Config setting "persistent" cannot be set to true, as the Sqlserver PDO driver does not support PDO::ATTR_PERSISTENT');
+        }
+
         $config['flags'] += [
-            PDO::ATTR_PERSISTENT => $config['persistent'],
             PDO::ATTR_EMULATE_PREPARES => false,
             PDO::ATTR_ERRMODE => PDO::ERRMODE_EXCEPTION
         ];
--- a/tests/TestCase/Database/Driver/SqlserverTest.php
+++ b/tests/TestCase/Database/Driver/SqlserverTest.php
@@ -103,7 +103,6 @@ class SqlserverTest extends TestCase
     {
         $this->skipIf($this->missingExtension, 'pdo_sqlsrv is not installed.');
         $config = [
-            'persistent' => false,
             'host' => 'foo',
             'username' => 'Administrator',
             'password' => 'blablabla',
@@ -121,7 +120,6 @@ class SqlserverTest extends TestCase
 
         $expected = $config;
         $expected['flags'] += [
-            PDO::ATTR_PERSISTENT => false,
             PDO::ATTR_EMULATE_PREPARES => false,
             PDO::ATTR_ERRMODE => PDO::ERRMODE_EXCEPTION,
             PDO::SQLSRV_ATTR_ENCODING => 'a-language'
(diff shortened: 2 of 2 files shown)


MCF:
If there is a desire to be complete with this fix it could be backported to CakePHP 2, and the original documentation fix should be adjusted to indicate that the option is ignored for SQL Server.

ADmad:
I don't think silently ignoring an option is a good idea. It would br better to throw an exception instead with proper message.

MCF:
Current behavior when using versions 4.1.8 of the PHP SQL Server drivers is that an exception _is_ thrown.  But it is thrown by the PHP driver.  For older versions of the driver (including everyone still on PHP 5.6 and older) a crashing bug is triggered.

I don't have a problem throwing an exception instead of just ignoring it, but I don't have strong feeling either way.

Just add it to the connection method in Sqlserver.php ?

MCF:
Also should I throw the exception if the 'persistent' option is present in the config?  Or just if it is present and set to true?  If the former then users will be forced to adjust their config once this change lands, if the later (or my original fix) then no change is required to their config.

ADmad:
Throw an exception only if `persist` is set to `true`.

src/Database/Exception/InvalidSettingException.php
```diff
@@ -0,0 +1,26 @@
...
+ * CakePHP(tm) : Rapid Development Framework (https://cakephp.org)
+ * Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
+ *
+ * Licensed under The MIT License
+ * For full copyright and license information, please see the LICENSE.txt
+ * Redistributions of files must retain the above copyright notice.
+ *
+ * @copyright     Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
+ * @link          https://cakephp.org CakePHP(tm) Project
+ * @since         3.0.0
```
markstory: This should reflect the actual version number.

src/Database/Driver/Sqlserver.php
```diff
@@ -64,8 +65,14 @@ public function connect()
             return true;
         }
         $config = $this->_config;
+
+        // PDO::ATTR_PERSISTENT is not allowed by SQL Server PHP drivers.
+        // See: https://github.com/Microsoft/msphpsql/issues/65
+        if (isset($config['persistent']) && $config['persistent']) {
+            throw new InvalidSettingException(['setting' => 'persistent']);
```
markstory: Not sure we need an exception class for this. You could use InvalidArgumentException instead.

src/Database/Driver/Sqlserver.php
```diff
@@ -64,8 +64,14 @@ public function connect()
             return true;
         }
         $config = $this->_config;
+
+        // PDO::ATTR_PERSISTENT is not allowed by SQL Server PHP drivers.
+        // See: https://github.com/Microsoft/msphpsql/issues/65
```
ravage84: You could move these two lines above and add an ``@link`` to the doc block to increase visibilty in the API docs for this issue.

src/Database/Driver/Sqlserver.php
```diff
@@ -64,8 +64,14 @@ public function connect()
...
+        if (isset($config['persistent']) && $config['persistent']) {
+            throw new \InvalidArgumentException('Sqlserver driver does not allow config setting: persistent');
```
ravage84: Why don't we use a message that actually explains the problem, like ``Sqlserver does not support perstistent connections set through config settings persistent``. Or even better we state: ``The SQL Server PHP driver does not support persistent connections. Please set config setting "persistent" to "false".``.

tests/TestCase/Database/Driver/SqlserverTest.php
```diff
@@ -160,6 +158,31 @@ public function testConnectionConfigCustom()
...
+     *
+     * @expectedException \InvalidArgumentException
+     * @expectedExceptionMessage Sqlserver driver does not allow config setting: persistent
+     * @return void
+     */
+    public function testSettingPersistentException()
+    {
+        $this->skipIf($this->missingExtension, 'pdo_sqlsrv is not installed.');
+        $config = [
+            'persistent' => true,
```
ravage84: Is there a test which covers if ``'persistent' => false,``?

ravage84 (changes requested):
Looks good, I think we still can improve this.

ravage84:
@MCF I would love to see this backported to 2.x. This could save us from some support issues...

src/Database/Driver/Sqlserver.php
MCF: I don't think we should tell anyone to set it to false becaue it may give the impression that we are telling the underlying driver to do something related to persistence of connections.  We are not.

Connection persistence can only be manipulated through the configuration of the underlying ODBC layer, which is not exposed through the PHP PDO drivers.

tests/TestCase/Database/Driver/SqlserverTest.php
MCF: Since the 'persistent' option is ignored completely I think this test case would have limited utility.  We don't test that the connection still works with any other unused options (for example option name mispellings).  I can add a test but it feels like overkill.

MCF:
I can backport, but will wait until we are all happy with this pull request first.

src/Database/Driver/Sqlserver.php
ravage84: What about? ``Enabling persistent connections in Sqlserver by setting the config key "persistent" to "true" is not supported, as the SQL Server PHP driver does not support persistent connections``.

My goal is to make the error messag as clear as possible, as discussed in https://github.com/cakephp/cakephp/issues/10752

MCF:
FWIW here is the documentation for the SQL Server PHP drivers and connection pooling:

https://docs.microsoft.com/en-us/sql/connect/php/connection-pooling-microsoft-drivers-for-php-for-sql-server

Connection pooling is the functional equivalent to persistent connections, but it cannot be configured via the PHP PDO driver for SQL Server.  It is configured differently depending on the operating system, Windows gets it by default, *nix OSes need to configure it through a config file and a driver restart.

src/Database/Driver/Sqlserver.php
MCF: I adjusted the error message early today.  Is the new one O.K.?

Your newest proposal is similar to what I changed it to but your suggestion is not quite factually correct.  You may or may or may not have "persistent" connections through the SQL Server PDO drivers.  IMHO telling users they can't have it because of the PDO driver is misleading.  The functional equivalent of persistent connections is connection pooling which is controlled outside the PDO drivers.  I've added a link to some documentation below.

Or perhaps you would be happier with the following message which covers all of the bases?

_Enabling persistent connections in Sqlserver by setting the config key "persistent" to "true" is not supported, as connection persistence is not controlled through the SQL Server PDO PHP driver.   Instead the PDO PHP driver relies on ODBC connection pooling, which is configured at the ODBC driver level._

markstory:
Thanks!

Issue #11027, opened 2017-08-10
memcached.use_sasl is no longer a thing in memcached extension version 3


dzuelke:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: master
* Platform and Target: PHP 7.1.8 with `memcached` extension 3.0.3

### What you did

Called `MemcachedEngine::init()` on PHP with `memcached` extension, built with SASL support.

### What happened

An exception got thrown, because the `memcached.use_sasl` INI setting no longer exists in version 3 of the extension (released in February 2017). See https://pecl.php.net/package-info.php?package=memcached&version=3.0.0

```
2017-08-09 18:48:04 Error: [InvalidArgumentException] Memcached extension is not build with SASL support
Request URL: …
Stack Trace:
#0 vendor/cakephp/cakephp/src/Cache/CacheRegistry.php(90): Cake\Cache\Engine\MemcachedEngine->init(Array)
…
```

### What you expected to happen

Exception should not be thrown for `memcached` extension 3.0.0 and later.

markstory:
I'll take a look.

markstory:
Pull request up now.

PR #11030, opened 2017-08-11
Remove ini flag for memcached.sasl

markstory:
This ini flag doesn't exist in newer versions of memcached. Instead we should use the existence of the method to see if memcached was compiled with sasl support.

Refs #11027

### DIFF
--- a/src/Cache/Engine/MemcachedEngine.php
+++ b/src/Cache/Engine/MemcachedEngine.php
@@ -153,7 +153,7 @@ class MemcachedEngine extends CacheEngine
 
         $servers = [];
         foreach ($this->_config['servers'] as $server) {
-            $servers[] = $this->_parseServerString($server);
+            $servers[] = $this->parseServerString($server);
         }
 
         if (!$this->_Memcached->addServers($servers)) {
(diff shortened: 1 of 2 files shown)


dzuelke:
Is the method only available if SASL support is on? I can't check the source right now.

dzuelke:
I checked. The answer is yes :) but for 2.0 of the extension, the INI setting is still needed. Probably doesn't matter though.

src/Cache/Engine/MemcachedEngine.php
```diff
@@ -268,6 +268,29 @@ protected function _parseServerString($server)
     }
 
     /**
+     * Backwards compatible alias of parseServerString
+     *
+     * @param string $server The server address string.
+     * @return array Array containing host, port
+     * @deprecated 3.4.13 Will be removed in 4.0.0
+     */
+    protected function _parseServerString($server)
```
saeideng: Backwards compatible for protected/private function??

markstory:
I don't think we need to continue checking the ini flag. I can update the doc blocks to mention that this flag is required for the 2.x extension.

PR #11031, opened 2017-08-11
Bump up phpstan to level 2.

ADmad:
Any help appreciated in dealing with the remaining reported errors. Though it seems most will have to be added to ignore list.


dereuromark:
Can we annotate those missing methods on the interface or somewhere?
Will phpstan accept that instead of having to add those to the ignore list?

ADmad:
I had tried adding annotations for the missing methods to the interface but IIRC it doesn't work.

ADmad:
I checked again and adding `@method` annotation to interfaces for missing methods does make phpstan happy.

src/Collection/CollectionTrait.php
```diff
@@ -702,7 +702,7 @@ public function isEmpty()
     public function unwrap()
     {
         $iterator = $this;
-        while (get_class($iterator) === 'Cake\Collection\Collection') {
+        while ($iterator instanceof Collection) {
```
markstory: I think this was checking for the specific class as there are a few subclasses that have to be treated differently.

ADmad: Okay, I'll revert this and ignore the error instead.

inoas: Can we have an inline comment about such decisions that are not apparent from the code bae itself?

ADmad:
Finally all green!

src/ORM/Association/BelongsToMany.php
```diff
@@ -510,7 +510,7 @@ protected function _appendNotMatching($query, $options)
         $subquery = $this->_appendJunctionJoin($subquery, $conditions);
 
         $query
-            ->andWhere(function ($exp) use ($subquery, $conds) {
+            ->where(function ($exp) use ($subquery, $conds) {
```
lorenzo: This was actually on purpose, because of the derp I made with `$query->orWhere()` leaving the query in `OR` mode. To force it out of that mode, I used `andWhere`

ADmad: Okay, I'll revert this change and add method annotation to `QueryInterface` instead. In a previous PR @markstory had suggested changing the method call https://github.com/cakephp/cakephp/pull/11063#discussion_r134104138. I'll revert that too :)

ADmad:
@lorenzo Done

src/Database/Statement/StatementDecorator.php
```diff
@@ -39,7 +39,7 @@ class StatementDecorator implements StatementInterface, Countable, IteratorAggre
      * Statement instance implementation, such as PDOStatement
      * or any other custom implementation.
      *
-     * @var \Cake\Database\StatementInterface
+     * @var \PDOStatement
```
lorenzo: It should maybe be `\PDOStatement|\Cake\Database\StatementInterface`

dereuromark: We should always keep the interface in `|` appended IMO for all cases where we "are forced" to go back to the concrete one.

lorenzo (approved):
great job!

src/ORM/Association/BelongsToMany.php
saeideng: Where() and andWhere() is equal exactly

ondrejmirtes:
Really happy to see this! I hope PHPStan serves you well. 👍

PR #11032, opened 2017-08-12
Add warning log when a connection with transaction is going to be closed

motooka:
If a developer writes a code like the one described below, it's so hard to debug. This warning log may help them.

```
$conn = ConnectionManager::get('my_connection');
try {
	$conn->begin();
	// some insert/update/delete operations
	if ($someCondition) {
		// bug : missing $conn->commit();
		return;
	}
	// some insert/update/delete operations
	$conn->commit();
}
catch(Exception $e) {
	$conn->rollback();
}
```


tests/TestCase/Database/ConnectionTest.php
```diff
@@ -488,6 +488,25 @@ public function testSimpleTransactions()
...
+    {
+        $driver = $this->getMockFormDriver();
+        $connection = $this->getMockBuilder('\Cake\Database\Connection')
+            ->setMethods(['connect'])
+            ->setConstructorArgs([['driver' => $driver]])
+            ->getMock();
+        $connection->begin();
+        $this->assertTrue($connection->inTransaction());
+
+        // a warning log will be generated by __destruct of Connection.
```
markstory: But there is no test for the log message. Right now this test doesn't seem to be adding any additional coverage that we don't already have.

motooka: Thanks, I believed that testing the log messages is impossible, but it was wrong.

src/Database/Connection.php
```diff
@@ -127,6 +128,9 @@ public function __construct($config)
      */
     public function __destruct()
     {
+        if ($this->_transactionStarted) {
```
lorenzo: This need to check that the Log class actually exists, as it is possible to use the Database package without Log

lorenzo (changes requested):

src/Database/Connection.php
motooka: Sure, I'm going to add a condition. Thanks.

ravage84:
Could we prevent such bugs with a hint in the docs, may be?

markstory (approved):
Looks good to me. What do you think @lorenzo ?

lorenzo:
Looks good to me as well, currently on mobile, but it seems like Travis did not pass for this change :(

markstory:
There are now conflicts that need to be resolved before this can be merged.

src/Database/Connection.php
```diff
@@ -121,6 +122,19 @@ public function __construct($config)
...
     /**
+     * Destructor
+     *
+     * Disconnects the driver to release the connection.
+     */
+    public function __destruct()
+    {
+        if ($this->_transactionStarted && class_exists('Cake\Log\Log')) {
+            Log::warning('The connection is going to be closed but the transaction is still active.');
+        }
```
markstory: Recently this destructor was removed. Wouldn't that obviate the need for this change?

lorenzo: No, it would still be the same

src/Database/Connection.php
```diff
@@ -121,6 +122,19 @@ public function __construct($config)
...
+        unset($this->_driver);
```
dereuromark: Should we keep this line out of it then?

markstory: Yeah, we just removed that code.

chinpei215: I think it would be better to explain more detail about what had happened.
@motooka We have removed this `unset()` call in #11049. However, you have reverted the change in your merge commit accidentally. I cannot see any new changes in your merge commit. Could you please *merge*, not overwrite.
Also, I would suggest to use `git rebase` instead of `git merge` to solve conflicts. See [this help](https://book.cakephp.org/3.0/en/contributing/code.html#submitting-a-pull-request) for contributing code.

PR #11036, opened 2017-08-13
Adding a URL object

burzum:
Adding an object to build URLs the OO way

### DIFF
--- a/src/Routing/Url.php
+++ b/src/Routing/Url.php
@@ -182,6 +182,7 @@ class Url implements ArrayAccess
     public function setAbsolute($absolute = true)
     {
         $this->absolute = (bool)$absolute;
+        $this->url['_full'] = $absolute;
 
         return $this;
     }
--- a/src/Routing/Url.php
+++ b/src/Routing/Url.php
@@ -220,6 +220,13 @@ class Url implements ArrayAccess
      */
     public function setPort($port)
     {
+        if ($port > 65535 || $port <= 0) {
+            throw new InvalidArgumentException(sprintf(
+                'Port `%s` is outside of the specified range of 1 to 65535',
+                $port
+            ));
+        }
+
         $this->url['_port'] = $port;
 
         return $this;
--- a/src/Routing/Url.php
+++ b/src/Routing/Url.php
@@ -1,0 +1,272 @@
+<?php
+/**
+ * CakePHP(tm) : Rapid Development Framework (https://cakephp.org)
+ * Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
+ *
+ * Licensed under The MIT License
+ * For full copyright and license information, please see the LICENSE.txt
+ * Redistributions of files must retain the above copyright notice.
+ *
+ * @copyright     Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
+ * @link          https://cakephp.org CakePHP(tm) Project
+ * @since         3.5.0
+ * @license       https://opensource.org/licenses/mit-license.php MIT License
+ */
+namespace Cake\Routing;
+
+use ArrayAccess;
+use InvalidArgumentException;
+
+/**
+ * URL Object
+ *
+ * Use this object whenever you need to generate a URL for your app.
+ *
+ * This is not thought for external URLs.
+ */
+class Url implements ArrayAccess
+{
+
+    /**
+     * Array representation of the URL
+     *
+     * @var array
+     */
+    protected $url = [];
+
+    /**
+     * Absolute URL
+     *
+     * @var bool
+     */
+    protected $absolute;
+
+    /**
+     * Sets the action
+     *
+     * @param string $action Controller action name
+     * @return $this
+     */
+    public function setAction($action)
+    {
+        $this->url['action'] = $action;
+
+        return $this;
+    }
+
+    /**
+     * Sets the controller
+     *
+     * @param string $controller Controller class name without the controller suffix
+     * @return $this
+     */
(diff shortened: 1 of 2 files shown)


markstory:
Should we have integration tests that use this new object in places where url arrays are used? I suspect there are likely a few rough spots still to be found.

chinpei215:
@burzum @markstory Do you have any opinions from view from the view point of PSR-7? Since PSR-7 contains a Uri class, we would end up having two different url classes which cannot communicate each other. But if our new url class were compatible with PSR-7 interface, people could do like the following:
```php
$uri = $this->getRequest()->getUri()->withPrefix('admin');
```

burzum:
@chinpei215 this serves a different purpose. Have you actually looked at the UriInterface?

https://github.com/php-fig/http-message/blob/master/src/UriInterface.php

The URL object here will be feed to the router, it's just like a config object and the router will generate the URL. What you suggest would imply to implement the actual URL building in the object itself?

chinpei215:
> this serves a different purpose. Have you actually looked at the UriInterface?

Yes, but I don't think PSR-7 serves a different purpose. To me,

```php
(new Url)
    ->setController('users')
    ->setAction('index')
    ->setPass(1)
    ->setQuery('foo', 'bar');
```
seems not so different from:
```php
(new Uri)
    ->withPath('/users/index/1')
    ->withQuery('foo=bar');
```
A difference is that the former can handle cake specific url parameters, but the latter cannot. But if we have our own Uri class which implements the PSR-7 interface, the former doesn't have any benefit to build urls:
```php
(new Uri)
    ->withController('users')
    ->withAction('index')
    ->withPass(1)
    ->withQueryParam('foo', 'bar');
```
An issue I can see is that it might not be immutable strictly, as people can change the router settings after the Uri object has been constructed.

> What you suggest would imply to implement the actual URL building in the object itself?

Yes, it would. Even now, the new Url class has the toString() method which wraps Router::url().

burzum:
@chinpei215 it is defferent.

The closest that `->withPath('/users/index/1')` matches is `setPass()`, but this would only match the `1`. Also by passing the path directly you would have to either parse it or lose the routing feature. The routing requires the controller and action plus optionally prefix and other things.

Also I'm not a huge fan of this notation: `->withQuery('foo=bar');` It's easy to make mistake this way, especially if the string becomes longer, thats why we have `setQueryParam($key, $value)`.

> Yes, it would. Even now, the new Url class has the toString() method which wraps Router::url().

Well, this would require a lot of refactoring work on the whole routing code I think. I've already seen cases in the router code that are marked for deprecation in 4.0. I doubt we want to make huge changes to the 3.x router right now?

chinpei215:
I still cannot see the difference well. I know withPath() would work differently from setPass(). But what about withPass()? (By the way, I didn't notice your Url class doesn't have setPass() actually).

> Well, this would require a lot of refactoring work on the whole routing code I think.

Maybe. But if people want this feature in the future, then we would have deprecate your Url class. So I think it is worth discussing it before introducing new Url class. 

If I remember correctly, people can use the `_host` or `_scheme` or `#` option in a url array. In your implementation, people would have to write the following:
```php
$host = $url->getParam('_host');
$scheme = $url->getParam('_scheme');
$fragment = $url->getParam('#');
```
It is hard to say that this is a good interface. Though you can add getHost()/getScheme(), setHost()/setScheme() and getFragment()/setFragment() to your Url class, but then it would be more similar the PSR-7 UriInterface than now.

inoas:
As this class relates to CakePHP constructs such as controller/action/prefix it should not be named URL as it is not strictly an URL. Maybe "Route" or "CakeUri" ?

ravage84:
@ionas  The full class name is ``Cake\Routing\Url ``. For me, it contains everything it needs. 1. It's part of CakePHP, 2. it's part of routing, not some general HTTP namespace, 3. it's about URLs.

burzum:
@markstory what needs to be done to get this finished?

src/Routing/Url.php
```diff
@@ -0,0 +1,332 @@
...
+ * CakePHP(tm) : Rapid Development Framework (https://cakephp.org)
+ * Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
+ *
+ * Licensed under The MIT License
+ * For full copyright and license information, please see the LICENSE.txt
+ * Redistributions of files must retain the above copyright notice.
+ *
+ * @copyright     Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
+ * @link          https://cakephp.org CakePHP(tm) Project
+ * @since         3.5.0
```
robertpustulka: This needs a version bump.
`@since 3.6.0`

tests/TestCase/Routing/UrlTest.php
```diff
@@ -0,0 +1,233 @@
...
+/**
+ * CakePHP(tm) : Rapid Development Framework (https://cakephp.org)
+ * Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
+ * Licensed under The MIT License
+ * For full copyright and license information, please see the LICENSE.txt
+ * Redistributions of files must retain the above copyright notice.
+ *
+ * @copyright     Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
+ * @link          https://cakephp.org CakePHP(tm) Project
+ * @since         3.5
```
robertpustulka: Same here: `@since 3.6.0`

dereuromark:
Seems like this is not going to happen.
Maybe making this a plugin for now as opt-in would help?
Closing then as it wont fit into 3.7.

PR #11046, opened 2017-08-15
Make internal driver usage consistent.

dereuromark:
Edit: Adjusting internal use of property.

The Connection typehints are necessary because the used methods are not in the interface (not clean!),
so we need to typehint the concretes are until we can clean this up in 4.x.

### DIFF
--- a/src/Database/Query.php
+++ b/src/Database/Query.php
@@ -211,10 +211,10 @@ class Query implements ExpressionInterface, IteratorAggregate
     public function execute()
     {
         $statement = $this->_connection->run($this);
-        $driver = $this->_connection->driver();
         $typeMap = $this->getSelectTypeMap();
 
         if ($typeMap->toArray() && $this->_typeCastAttached === false) {
+            $driver = $this->_connection->getDriver();
             $this->decorateResults(new FieldTypeConverter($typeMap, $driver));
             $this->_typeCastAttached = true;
         }
--- a/src/Database/Connection.php
+++ b/src/Database/Connection.php
@@ -294,7 +294,7 @@ class Connection implements ConnectionInterface
      */
     public function compileQuery(Query $query, ValueBinder $generator)
     {
-        return $this->getDriver()->compileQuery($query, $generator)[1];
+        return $this->_driver->compileQuery($query, $generator)[1];
     }
 
     /**
@@ -843,7 +843,7 @@ class Connection implements ConnectionInterface
      */
     protected function _newLogger(StatementInterface $statement)
     {
-        $log = new LoggingStatement($statement, $this->getDriver());
+        $log = new LoggingStatement($statement, $this->_driver);
         $log->logger($this->logger());
 
         return $log;
--- a/src/Database/Connection.php
+++ b/src/Database/Connection.php
@@ -294,7 +294,7 @@ class Connection implements ConnectionInterface
      */
     public function compileQuery(Query $query, ValueBinder $generator)
     {
-        return $this->_driver->compileQuery($query, $generator)[1];
+        return $this->getDriver()->compileQuery($query, $generator)[1];
     }
 
     /**


markstory:
Changes look good. Should we consider eagerly building the driver instance instead of lazily fetching it each time? @lorenzo Are you concerned about the performance impact of the additional method call in `prepare()`?

chinpei215:
@dereuromark Sorry, but could you please explain why this change solves the issue? I think Connection doesn't lazily load the driver.
https://github.com/cakephp/cakephp/blob/3a6bd7583204e3619b8bb98b385cc1b27d2724eb/src/Database/Connection.php#L186-L189

Also, as I said on our Slack, `Undefined property: Cake\Database\Connection::$_driver` error doesn't mean the properties is null. If so, the error message should be `Call to a member function prepare() on null` or something. So I am guessing that the destructor has been called for unknown reasons.
https://github.com/cakephp/cakephp/blob/3a6bd7583204e3619b8bb98b385cc1b27d2724eb/src/Database/Connection.php#L128-L131

By the way, regarding the data types, I have already changed them in the 3.5 branch.
https://github.com/cakephp/cakephp/blob/020915ecaf620f4d09360388f4ea28298b249769/src/Database/Query.php#L40

dereuromark:
@chinpei215 That is exactly the scenario.
Either we need to stop unsetting on destruct (doesnt PHP handle it well internally already?) or we need to make sure those dependencies can be lazyloaded again properly. Because it is possible that somewhere the destruct is called, and then later the code is used again.

Also please note that currently some methods are using the method, some dont. This feels weird a bit, too.

dereuromark:
@chinpei215 I tested some more and you are right, the main issue is the destruct that makes it impossible to use the current code in e.g. __destruct() of a Shell to log sth in DB etc.
For now I removed my __destruct() code in shell and it works. And I made the core code consistent.

I reverted the method wrapper because the fix wasnt correct. It would only work if getDriver() would check for null and reloading sth if so. Just using the wrapper does not help at all.
I wonder what could be done here.

src/Database/Query.php
```diff
@@ -211,10 +211,10 @@ public function connection($connection = null)
     public function execute()
     {
         $statement = $this->_connection->run($this);
-        $driver = $this->_connection->driver();
         $typeMap = $this->getSelectTypeMap();
 
         if ($typeMap->toArray() && $this->_typeCastAttached === false) {
+            $driver = $this->_connection->getDriver();
```
chinpei215: Seems still inconsistent to me. Shouldn't it be `_driver`?

lorenzo:
I think we can safely remove the `__destruct` method in `Connection`

dereuromark:
@lorenzo I think that would solve a few issues. If it doesnt impact any performance it sounds like a win win.

src/Database/Query.php
```diff
@@ -37,7 +37,7 @@ class Query implements ExpressionInterface, IteratorAggregate
     /**
      * Connection instance to be used to execute this query.
      *
-     * @var \Cake\Datasource\ConnectionInterface
+     * @var \Cake\Datasource\ConnectionInterface|\Cake\Database\Connection
```
chinpei215: Since I have already changed it in 3.5, this change will cause conflict on the next Friday :(
Also, `\Cake\Database\Query` should not accept `\Cake\Datasource\ConnectionInterface`. As you said, it wouldn't work if it is not an instance of `\Cake\Database\Connection`. So I simply changed `\Cake\Datasource\ConnectionInterface` to `Cake\Database\Connection` in my commit.

chinpei215:
>  I think that would solve a few issues combined with a true lazy loader approach. 

No, we cannot do it properly. Because the `driver` option accepts an  instance of `Driver`. In that case, you cannot lazily load again the driver.

src/Database/Query.php
dereuromark: Why? This is not protected internal access, this is public access from the outside.

src/Database/Query.php
dereuromark: Then merge yours first, I can resolve the conflicts.

src/Database/Query.php
chinpei215: Oh sorry, I misread it 🙇

src/Database/Query.php
chinpei215: This is my commits: #10963. Please do it if you want to fix it in 3.4.

dereuromark: Given the fact that 3.5 is basically there now, I reverted mine :) Thank you for doing the fixes!

chinpei215: Thank you.

markstory:
Code looks good but builds are very sad.

dereuromark:
Interesting, why would that break the tests? Does the getDriver() method do magically more than it looks like?

chinpei215:
@dereuromark Probably, there are mock connections overriding `getDeriver()` to return a mock driver.

PR #11047, opened 2017-08-15
phpdoc $names param type for applyMiddleware() function

jtraulle:
Phpdoc `...$names` param tag should be of type **string**, not **array of strings** for `applyMiddleware()` function because of the splat operator.
When type is `string[]`, [phan](/etsy/phan) complains about type mismatch when running static analysis.


markstory:
Thanks!

jtraulle:
You're welcome 😉

Issue #11044, opened 2017-08-15
Execution order bug in Query


dereuromark:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.4.latest

### What you did
```php
$this->deleteAll(['pid' => $pid]);
```

### What happened
```
PHP Notice:  Undefined property: Cake\Database\Connection::$_driver in /home/travis/build/dereuromark/cakephp-queue/vendor/cakephp/cakephp/src/Database/Connection.php on line 256
PHP Stack trace:
PHP   1. {main}() /home/travis/build/dereuromark/cakephp-queue/vendor/phpunit/phpunit/phpunit:0
PHP   2. PHPUnit\TextUI\Command::main() /home/travis/build/dereuromark/cakephp-queue/vendor/phpunit/phpunit/phpunit:53
PHP   3. PHPUnit\TextUI\Command->run() /home/travis/build/dereuromark/cakephp-queue/vendor/phpunit/phpunit/src/TextUI/Command.php:141
PHP   4. PHPUnit\TextUI\TestRunner->doRun() /home/travis/build/dereuromark/cakephp-queue/vendor/phpunit/phpunit/src/TextUI/Command.php:212
PHP   5. Mock_QueueShell_d67fbc62->__destruct() /home/travis/build/dereuromark/cakephp-queue/src/Shell/QueueShell.php:0
PHP   6. Queue\Model\Table\QueueProcessesTable->remove() /home/travis/build/dereuromark/cakephp-queue/src/Shell/QueueShell.php:496
PHP   7. Queue\Model\Table\QueueProcessesTable->deleteAll() /home/travis/build/dereuromark/cakephp-queue/src/Model/Table/QueueProcessesTable.php:101
PHP   8. Cake\ORM\Query->execute() /home/travis/build/dereuromark/cakephp-queue/vendor/cakephp/cakephp/src/ORM/Table.php:1595
PHP   9. Cake\Database\Connection->run() /home/travis/build/dereuromark/cakephp-queue/vendor/cakephp/cakephp/src/Database/Query.php:213
PHP  10. Cake\Database\Connection->prepare() /home/travis/build/dereuromark/cakephp-queue/vendor/cakephp/cakephp/src/Database/Connection.php:309
[...]

PR #11049, opened 2017-08-16
Removing the destructor to prevent notices on reuse.

dereuromark:
Fix for #11044 after more discussion in https://github.com/cakephp/cakephp/pull/11046
The alternative would be to set it to `null` only, not unsetting.

Lets see what the tests say.


dereuromark:
The build times on travis seem to be even faster than on master.

lorenzo:
Looks good!

PR #11050, opened 2017-08-16
Fix code smell

dereuromark:
We should strict check on a pure integer.

### DIFF
--- a/src/Controller/Component/PaginatorComponent.php
+++ b/src/Controller/Component/PaginatorComponent.php
@@ -204,7 +204,7 @@ class PaginatorComponent extends Component
 
         $order = (array)$options['order'];
         $sortDefault = $directionDefault = false;
-        if (!empty($defaults['order']) && count($defaults['order']) == 1) {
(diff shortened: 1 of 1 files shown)


src/Controller/Component/PaginatorComponent.php
```diff
@@ -204,7 +204,7 @@ public function paginate($object, array $settings = [])
 
         $order = (array)$options['order'];
         $sortDefault = $directionDefault = false;
-        if (!empty($defaults['order']) && count($defaults['order']) == 1) {
+        if (!empty($defaults['order']) && count($defaults['order']) === 1) {
```
saeideng: why not   `count($order) === 1` ??

dereuromark: $options['order'] != defaults['order'] maybe?

saeideng: I'm sorry 
`defaults` is not equal to `options`

PR #11053, opened 2017-08-16
Always do count for paginator results in params

dereuromark:
Always do a count on the pagination results in order to have that available when catching the NotFoundException to e.g. redirect to last page.
This was a quite a few times already requested feature that would easily be possible.

Is there a way maybe to only run the count() and fake the collection as empty one to avoid the 2nd query?
If so, it would invert the status quo resulting in the same output as this PR.

With the change here you now then could do sth like
```php
    public function paginate($object, array $settings = [])
    {
        try {
            $resultSet = parent::paginate($object, $settings);
        } catch (NotFoundException $exception) {
            $query = null;
            if ($object instanceof QueryInterface) {
                $query = $object;
                $object = $query->repository();
            }
            $alias = $object->alias();
            $lastPage = $this->request->params['paging'][$alias]['pageCount'] > 1 ? $this->request->params['paging'][$alias]['pageCount'] : null;

            $response = $this->getController()->redirect(['?' => ['page' => $lastPage] + $this->request->getQuery()]);

            // To be please PHPCS and tests, cannot be reached in production.
            if (PHP_SAPI === 'cli') {
                throw new NotFoundException('Redirect to ' . $response->getHeaderLine('Location') . ' for non-CLI.');
            } else {
                $response->send();
            }

            exit();
        }

        return $resultSet;
    }
```
for all your paginations at once.

But without the total count it will not be possible to calculate the correct page reusing the query inside the parent.

### DIFF
--- a/src/Controller/Component/PaginatorComponent.php
+++ b/src/Controller/Component/PaginatorComponent.php
@@ -190,7 +190,7 @@ class PaginatorComponent extends Component
         $cleanQuery = clone $query;
         $results = $query->all();
         $numResults = count($results);
-        $count = $numResults ? $cleanQuery->count() : 0;
+        $count = $cleanQuery->count();
 
         $defaults = $this->getDefaults($alias, $settings);
         unset($defaults[0]);
--- a/tests/TestCase/Controller/Component/PaginatorComponentTest.php
+++ b/tests/TestCase/Controller/Component/PaginatorComponentTest.php
@@ -673,7 +673,6 @@ class PaginatorComponentTest extends TestCase
             $this->Paginator->paginate($table);
             $this->fail('No exception raised');
         } catch (NotFoundException $e) {
-
             $this->assertEquals(
                 3,
                 $this->request->params['paging']['PaginatorPosts']['pageCount'],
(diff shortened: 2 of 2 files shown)


dereuromark:
To clarify:

### requesting page=12 (one page too much)

Before (problematic):
```
		'finder' => 'all',
		'page' => (int) 1,
		'current' => (int) 0,
		'count' => (int) 0,
		'perPage' => (int) 20,
		'prevPage' => false,
		'nextPage' => false,
		'pageCount' => (int) 0,
		'sort' => 'Modules.name',
		'direction' => 'asc',
		'limit' => null,
		'sortDefault' => 'name',
		'directionDefault' => 'ASC',
		'scope' => null
```

After:
```
		'finder' => 'all',
		'page' => (int) 11,
		'current' => (int) 0,
		'count' => (int) 216,
		'perPage' => (int) 20,
		'prevPage' => true,
		'nextPage' => false,
		'pageCount' => (int) 11,
		'sort' => 'Modules.name',
		'direction' => 'asc',
		'limit' => null,
		'sortDefault' => 'name',
		'directionDefault' => 'ASC',
		'scope' => null
```

lorenzo (approved):
This looks good to meo

markstory (commented):
Looks good. This feels like something that should have a test though.

dereuromark:
You are absolutely right, I added a test case.

ADmad:
Since 3.5 stable release is eminent I would suggest targeting this and #11054 to 3.next instead of master. Otherwise I will be unnecessary hassle dealing with merge conflicts since in 3.5 the guts of paging have been moved to `Paginator` class.

Issue #11052, opened 2017-08-16
Paginator page count bug


dereuromark:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.4 latest

### What you did
Filtering and pagination shows 0 results.
The params though show

		'finder' => 'all',
		'page' => (int) 1,
		'current' => (int) 0,
		'count' => (int) 0,
		'perPage' => (int) 20,
		'prevPage' => false,
		'nextPage' => false,
		'pageCount' => (int) 0, // Should be also 1 as page
		'sort' => 'Modules.name',
		'direction' => 'asc',
		'limit' => null,
		'sortDefault' => 'name',
		'directionDefault' => 'ASC',
		'scope' => null

### What you expected to happen
```php
$page = max(min($page, $pageCount), 1);
```
most likely needs a max(x, 1) for pageCount as well.

Or is the pageCount here acceptable as "we show first page, but it is not a real page"?

lorenzo:
How is this a problem?

dereuromark:
When catching the exception and relying on the results of params, you need to know the total number of pages sometimes.
There is a 2nd issue here, the count query not run, but thats a different issue.
It was just topic in the channel earlier.

dereuromark:
PR open.

PR #11054, opened 2017-08-16
Fix lower bound of total pages

dereuromark:
This seems to be the fix for https://github.com/cakephp/cakephp/issues/11052

Before (problematic):
```
		'page' => (int) 1,
		'pageCount' => (int) 0,
		'count' => (int) 0,
		...
```

After:
```
		'page' => (int) 1,
		'pageCount' => (int) 1,
		'count' => (int) 0,
		...
```

If you want to easily read the total number of pages and we currently display page 1, you would probably expect the pageCount to also be 1, and not having to do additional operations here based on another field.
If you need to know the results count in total, you can always check count 0.

### DIFF
--- a/src/Controller/Component/PaginatorComponent.php
+++ b/src/Controller/Component/PaginatorComponent.php
@@ -197,9 +197,9 @@ class PaginatorComponent extends Component
 
         $page = $options['page'];
         $limit = $options['limit'];
-        $pageCount = (int)ceil($count / $limit);
+        $pageCount = max((int)ceil($count / $limit), 1);
         $requestedPage = $page;
-        $page = max(min($page, $pageCount), 1);
+        $page = min($page, $pageCount);
         $request = $this->_registry->getController()->request;
 
         $order = (array)$options['order'];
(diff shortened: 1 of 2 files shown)


lorenzo (commented):
I'm not sure saying that there is always one page is the right thing to do. It is a valid interpretation, though: there is one page, with no results. 

But that's more like a mathematical definition that I'm sure would surprise many

dereuromark:
Isnt pageCount our "max page value" whereas page is the "min page value"?
max shouldnt be below min.

> there is one page, with no results.

That's why you for this check "count" key IMO, not pages.

markstory:
If you check the `count`  key why would this change need to be made?

dereuromark:
Yes. My point was about others checking it.
The above is needed to always know the current max page size per request without the need to calculate.

dereuromark:
$pageCount = $params['pageCount'] ?: 1;

would sure work, but shouldnt be necessary maybe.

dereuromark:
I added a test to avoid regressions.

ravage84:
I spontaneously asked to of my colleagues. Both agreed quite fast with what @dereuromark wants to express.

I think it's a kind of philosophical question, what this ``pageCount`` means, but since we are in a context of a "pagination" and pages are the main thing of a pagination, I think it does not make much sense to express that there are no pages in that set of pages. Especially if you express at the same time, that you are currently on page one. 😼  

It's like saying this is a book that has no pages in it, as there are no words to be found, but you are curently on page one.... 😆

PR #11056, opened 2017-08-16
Sqlserver DboSource does not allow the persistent option to be true.

MCF:
This is backported from CakePHP 3.  The SQL Server PHP PDO driver does not
support the PDO::ATTR_PERSISTENT attribute.  So throw an exception if the
'persistent' option is set in the database config.  Also removes that
option from the Sqlserver base config.

Backport of: https://github.com/cakephp/cakephp/pull/11028 for CakePHP 2


markstory:
Thanks!

PR #11057, opened 2017-08-17
Passing paging params to PageOutOfBoundException.

ADmad:
This for e.g. would allow someone to have use a custom exception render
and redirect to last available page if PageOutOfBoundException is thrown.

### DIFF
--- a/src/Datasource/Exception/PageOutOfBoundsException.php
+++ b/src/Datasource/Exception/PageOutOfBoundsException.php
@@ -22,12 +22,12 @@ class PageOutOfBoundsException extends Exception
     /**
      * {@inheritDoc}
      */
-    protected $_messageTemplate = 'Page number "%s" could not be found.';
+    protected $_messageTemplate = 'Page number %s could not be found.';
 
     /**
      * Constructor
      *
-     * @param string|null $message The error message.
+     * @param array|null $message Paging info.
      * @param int $code The code of the error, is also the HTTP status code for the error.
      * @param \Exception|null $previous The previous exception.
      */
--- a/src/Datasource/Paginator.php
+++ b/src/Datasource/Paginator.php
@@ -220,7 +220,10 @@ class Paginator implements PaginatorInterface
         $this->_pagingParams = [$alias => $paging];
 
         if ($requestedPage > $page) {
-            throw new PageOutOfBoundsException([$requestedPage]);
+            throw new PageOutOfBoundsException([
+                'requestedPage' => $requestedPage,
+                'pagingParams' => $this->_pagingParams
+            ]);
         }
 
         return $results;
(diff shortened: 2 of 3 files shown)


dereuromark:
That definitely makes things easier :)

src/Datasource/Exception/PageOutOfBoundsException.php
```diff
@@ -22,12 +22,12 @@ class PageOutOfBoundsException extends Exception
...
      * {@inheritDoc}
      */
-    protected $_messageTemplate = 'Page number "%s" could not be found.';
+    protected $_messageTemplate = 'Page number %s could not be found.';
 
     /**
      * Constructor
      *
-     * @param string|null $message The error message.
+     * @param array|null $message Paging info.
```
dereuromark: Is it allowed for some exceptions to have an array here instead of string as argument?
Couldnt this break some handlers that do not expect this?

ADmad: My test case wouldn't pass if it was not allowed :)  `Cake\Core\Exception` allows `$message` to be string or array.

lorenzo:
I like this :)

markstory:
Does this replace #11053 or #11054?

dereuromark:
No, those are all 3 different issues.

markstory:
But all three are working to solve the same problem which is making application code easier to write when someone requests an out of bounds page?

dereuromark:
not sand but related

src/Datasource/Paginator.php
```diff
@@ -220,7 +220,10 @@ public function paginate($object, array $params = [], array $settings = [])
         $this->_pagingParams = [$alias => $paging];
 
         if ($requestedPage > $page) {
-            throw new PageOutOfBoundsException([$requestedPage]);
+            throw new PageOutOfBoundsException([
+                'requestedPage' => $requestedPage,
+                'pagingParams' => $this->_pagingParams
```
dereuromark: We should either add the alias here too, or narrow down the paging params to only the subset of this alias IMO.
Otherwise it can be difficult to get to the needed params here.

ADmad: `$this->_pagingParams` already has the alias at this point due to line 220.

dereuromark: Yes, but could there be multiple keys in here? How would you then know to extract the correct alias etc?

ADmad: Given that `$this->_pagingParams` is reset for every call to `paginate()` I don't see how it could have multiple aliases/keys.

dereuromark: Ok, so foreach over it should reveal it then. Fair enough.

PR #11058, opened 2017-08-17
[code style] remove unused namespaces

TomasVotruba:


dereuromark:
Isnt there a sniff that could be included to auto-fix this?
Otherwise one should port [this one](https://github.com/spryker/code-sniffer/blob/master/Spryker/Sniffs/Namespaces/UnusedUseStatementSniff.php) to CakePHP, as this should not have to be done manually IMO.

TomasVotruba:
I hoped you'd ask :+1: I used `NoUnusedImportsFixer`

image

dereuromark:
Well, we dont use that fixer tool, it is completely different to the phpcs sniffer tool :)
We need a sniff here with autofixing.

TomasVotruba:
> Well, we dont use that fixer tool

I know, that's why I share it.

> it is completely different to the phpcs sniffer tool :)

In what way? I use them both with no difference.

dereuromark:
Because that's the tool we are using: https://github.com/cakephp/cakephp/blob/master/composer.json#L94-L95
We successfully use the above linked sniffer for our code base for many months. Might be worth porting.

TomasVotruba:
I just see duplicating open-source projects as a waste of time that could be used on creating new features, that's all.

Feel free to use tool you know of course.

dereuromark:
phpcsniffer has a few advantages over the fixer one, we do not have to get into details here now.
Sure both are doing their job, but switching our tooling because of this also doesnt seem wise maybe.
So porting a simple sniff seems way less a waste of time in my book.

TomasVotruba:
> but switching our tooling because of this also doesnt seem wise maybe.

I definitelly agree. I talked about switching them? Not at all.

**Both** have their advantages, so I use them **both**.

TomasVotruba:
:+1: Thanks

gmponos:
Although the having a sniff is the best thing to use here few of you who use PHPStorm you could use the option from PHPStorm which optimizes imports. This way it removes also unused `uses`

TomasVotruba:
I try to think of program independent solutions. 
I mean we could also run Docker container that would run PHPStorm on Travis CI. That *would* work, I would not be able to sell it on others because I would not believe it.

*Anything that could be automated, should be*. For the best, this issue should never appear again till we die :)

gmponos:
I am not saying that this is the solution that you should apply. I am just saying that this thing also exists.

PR #11060, opened 2017-08-19
Fix string handling of Hash::remove() and Hash::_simpleOp('remove')

chinpei215:
At the momemnt,
```php
$array = [
    'foo' => 'string',
];
Hash::remove($array, 'foo.bar');
```
Will cause a fatal error.

```php
$array = [
    'foo' => [
        0 => 'a',
        1 => 'b',
    ],
];
Hash::remove($array, 'foo[1=b].0');
```
Will return an empty array unexpectedly.

Also, I removed a meaningless code. 
https://github.com/cakephp/cakephp/blob/16903f9d60c6e00443637a6f2dd9a77af649269c/src/Utility/Hash.php#L353-L357
Decimal keys need not to be converted to integers explicitly when using operator `[]`.

### DIFF
--- a/src/Utility/Hash.php
+++ b/src/Utility/Hash.php
@@ -365,7 +365,9 @@ class Hash
                 }
             } elseif ($op === 'remove') {
                 if ($i === $last) {
-                    unset($_list[$key]);
+                    if (!is_string($_list)) {
+                        unset($_list[$key]);
+                    }
 
                     return $data;
                 }
@@ -414,7 +416,7 @@ class Hash
             if ($match && is_array($v)) {
                 if ($conditions) {
                     if (static::_matches($v, $conditions)) {
-                        if ($nextPath) {
+                        if ($nextPath !== '') {
                             $data[$k] = static::remove($v, $nextPath);
                         } else {
                             unset($data[$k]);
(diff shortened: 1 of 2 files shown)


src/Utility/Hash.php
```diff
@@ -370,7 +365,9 @@ protected static function _simpleOp($op, $data, $path, $values = null)
                 }
             } elseif ($op === 'remove') {
                 if ($i === $last) {
-                    unset($_list[$key]);
+                    if (!is_string($_list)) {
```
markstory: Should we check for other scalar values? Or ensure that `$_list` is an array before unsetting from it?

chinpei215: I thought other scalar values can be unset safely and nothing happens, but using `is_array()` would be better, as `_simpleOp('insert')` uses the function [here](https://github.com/chinpei215/cakephp/blob/ea5f05b46136803703b946504bc5d66901f6b36a/src/Utility/Hash.php#L363-L365). And, I could move this check to the first statement of the `elseif ($op === 'remove')` block and return `$data` there. Thanks.

chinpei215: Hmm, but taking the performance into consideration, doing the check at the current place may be better. I'll change `!is_string()` to `is_array()`.

PR #11062, opened 2017-08-19
Fix up behavior callbacks for clean inheritance and consistency.

dereuromark:
Refs https://github.com/cakephp/docs/issues/5189 and the IDE warning issued here.

Makes them consistent with https://book.cakephp.org/3.0/en/orm/table-objects.html and allows clean inheritence without the "Declaration" notice.


chinpei215:
This change is not so backward compatible, as it would report E_STRICT level errors if people have overridden those methods. It would be better to target 3.6. By the way, why does IDE warn about those declarations? Does it dislike passing more arguments than the declaration?

dereuromark:
The `@mixin` annotations cause this because they then treat them on inheritance level.
3.6 sounds fine to me.

README.md
```diff
@@ -38,7 +38,7 @@ recommend using the [app skeleton](https://github.com/cakephp/app) as
 a starting point. For existing applications you can run the following:
 
 ``` bash
-$ composer require cakephp/cakephp:"~3.4"
+$ composer require cakephp/cakephp:"~3.5"
```
saeideng: 3.next is 3.6 not 3.5

dereuromark: those are not my changes

markstory: That's a change I made on master last night.

VERSION.txt
```diff
@@ -16,4 +16,4 @@
 // @license       https://opensource.org/licenses/mit-license.php MIT License
 // +--------------------------------------------------------------------------------------------+ //
 ////////////////////////////////////////////////////////////////////////////////////////////////////
-3.5.0-RC2
+3.5.0
```
markstory: Odd. This should be 3.6.0-dev on `3.next`.

src/ORM/Behavior/CounterCacheBehavior.php
```diff
@@ -114,7 +115,7 @@ class CounterCacheBehavior extends Behavior
      * @param \ArrayObject $options The options for the query
      * @return void
      */
-    public function beforeSave(Event $event, EntityInterface $entity, $options)
+    public function beforeSave(Event $event, EntityInterface $entity, ArrayObject $options)
```
robertpustulka: Can we hint against the `ArrayAccess` interface instead of a concrete implementation?

dereuromark: probably in 4.0 then.

robertpustulka: Why?

dereuromark: for consistency across all callbacks i guess.

robertpustulka: Sorry, I was convinced we were type hinting against `ArrayAccess` everywhere else.

dereuromark:
So is this good to go then?

PR #11063, opened 2017-08-19
Add annotations for missing methods to interfaces.

### LINKED ISSUE #11051: Core methods start using classes instead of interfaces.
This is a (multiple allowed):

* [ ] bug
* [ ] enhancement

* CakePHP Version: 3.5-RC2

### What you did
Initially when cakephp 3.x branch was started we had discussion and agreed to use interfaces over classes in case we have alternative to use interfaces.
E.g. we used EntityInterface instead of Entity class to provide way to create app level Entity classes implementing same interface with additional functionallity.

### What happened
But in 3.5 branch i see this commit 68d2fcb83fbd1612f10c00b2ba14e44c8cfeeafb (Fix some errors reported by phpstan level 2.) that revert some methods declaration to use Entity instead of interface and some other interfaces replaced with classes too.

I need confirmation everyone fine to revert to interfaces back.

ADmad:
This prevents static analyzers like phpstan complaining when an argument is
typehinted as interface and a method used on object is not present in the interface.

Closes #11051

### DIFF
--- a/src/Datasource/QueryInterface.php
+++ b/src/Datasource/QueryInterface.php
@@ -18,7 +18,7 @@ namespace Cake\Datasource;
 /**
  * The basis for every query object
  *
- * @package Cake\Datasource
+ * @method $this andWhere($conditions, $types = [])
  */
 interface QueryInterface
 {
--- a/src/Datasource/QueryInterface.php
+++ b/src/Datasource/QueryInterface.php
@@ -18,7 +18,6 @@ namespace Cake\Datasource;
 /**
  * The basis for every query object
  *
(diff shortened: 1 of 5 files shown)


src/Datasource/QueryInterface.php
```diff
@@ -18,7 +18,7 @@
 /**
  * The basis for every query object
  *
- * @package Cake\Datasource
+ * @method $this andWhere($conditions, $types = [])
```
markstory: We could change the callers to use `where()` instead of expanding the interface here.

Issue #11045, opened 2017-08-15
[RFC] Collection should not extend IteratorIterator


chinpei215:
This is a (multiple allowed):

* [ ] bug
* [ ] enhancement
* [x] feature-discussion (RFC)

* CakePHP Version: 3.4.12
* Platform and Target: Non-related

### What you did

```php
$collection = collection([3,2,1]);
$collection->asort();
debug( $collection->toArray() );
```

### What happened

```php
[
	(int) 2 => (int) 1,
	(int) 1 => (int) 2,
	(int) 0 => (int) 3
]
```

### What you expected to happen

Methods on the inner ArrayIterator should not be called.

### Why this issue happens

Because `Collection` is a sub-class of `IteratorIterator`, and [PHP manual](http://php.net/manual/en/class.iteratoriterator.php) says:
> This class permits access to methods of the inner iterator via the __call magic method.

So Collection allows people to call methods on the inner Traversable objects. If it is an ArrayIterator, then people can call `offsetSet()`, `offsetUnset()` or a number of sorting methods. This fact breaks immutability of the Collection class. It would be better to not extend `IteratorIterator` if possible.

Related to #7565

markstory:
Could this change be done without breaking compatibility? Could we overload `__call` to raise exceptions as another way to ensure encapsulation doesn't get broken?

chinpei215:
> Could this change be done without breaking compatibility?

Strictly speaking, no. If people are calling some methods on the inner iterator, like `count()`, their code will break. Or if they have `if ($collection instanceof IteratorIterator)' check for some reasons, collections never run into the block. 

But there might be some technical reasons. @lorenzo Any thoughts? Wouldn't it work if `Collection` implemented `IteratorAggregate` and returned `IteratorIterator` in the `getOuterIterator` method?

> Could we overload __call to raise exceptions as another way to ensure encapsulation doesn't get broken?

Maybe, we could. I could iterate a `DatePeriod` object without any problems even if the `__call` method has been disabled by overriding. So I don't know why `IteratorIterator` permits access to methods of the inner iterator.

lorenzo:
Trust me, I have tried hundreds of times to implement it without IteratorIterator, it is. It is near to impossible considering all the features I wanted to support initially.

Last time I tried was a month ago, and I had to throw away the code when I kept hitting problems :(

lorenzo:
We could implement __call() and throw an exception there, maybe that would not break the world and fix your issue?

[...]

PR #11064, opened 2017-08-19
[3.next] Disable access to methods of inner iterator of Collection

chinpei215:
Refs #11045


markstory (approved):
Looks good to me.

chinpei215:
Thanks. @lorenzo Any thoughts?

lorenzo:
Thanks!

ravage84:
This should probably be included in the migration guide, isn't it?

chinpei215:
Yes, it should.

PR #11068, opened 2017-08-20
3.x Docblock CacheSession $id is string|int

inoas:
Ref: https://github.com/cakephp/app/issues/535#issuecomment-323557234


markstory:
Thanks!

Issue #11023, opened 2017-08-09
[RFC] Support for connection switching


makallio85:
This is a (multiple allowed):

* [ ] bug
* [ ] enhancement
* [X] feature-discussion (RFC)

CakePHP4?

I would really like to see CakePHP ORM to support connection switching in following scenario.

I have application that is built in concept "customer per database". Meaning, that I have n * databases for each customer and n * some kind of master databases.

I am aware of altering connection details and using one static connection name for each customer database but problem occurs, when i want to use same table objects in customer and master databases.

Example

I have table classes InvoicesTable and InvoiceRowsTable. Each customer can have its own invoices as do I as service provider. Design is equal in both levels. Customers are sending invoices to their customers and I am sending invoices to my customers, that are using this application.

Problem occurs, when I want to do something like this:

```
InvoicesTable extends Table
{
    $this->hasMany('InvoiceRows');
}

InvoiceRowsTable extends Table
{
   $this->belongsTo('Invoices');
}

$tableInShard = TableRegistry::get('Invoices', ['connection' => ConnectionManager::get('shard')]);
$tableInShard->save($table->newEntity(['invoice_rows' => [['amount' => 100]]]), ['associated' => 'InvoiceRows']);

$tableInMaster = TableRegistry::get('Invoices', ['connection' => ConnectionManager::get('master')]);
$tableInShard->save($table->newEntity(['invoice_rows' => [['amount' => 100]]]), ['associated' => 'InvoiceRows']);
```
Because table is already configured in table registry, currently I have to iterate over table registry and replace connection for every table (that is used, even associations) every time Table in instatiated.  This requires small helper class.

robertpustulka:
Maybe all you need is different table aliases for different connections?

```php
$tableInShard = TableRegistry::get('ShardInvoices', [
  'className' => 'Invoices', 
  'connection' => ConnectionManager::get('shard')
]);
$tableInMaster = TableRegistry::get('MasterInvoices', [
  'className' => 'Invoices', 
  'connection' => ConnectionManager::get('master')
]);
```

Your associations could also have connection name prefixes if necessary.

```php
InvoicesTable extends Table
{
    $name = ucfirst($this->_connection->configName()) . 'InvoiceRows';
    $this->hasMany($name, [
        'className' => 'InvoiceRows'
    ]);
}
```

You could build a behavior that would handle the correct naming of associations. Or maybe you need to implement your own table locator, would assign correct connections.

makallio85:
[...]

PR #11072, opened 2017-08-21
Make Table aware of a table locator.

robertpustulka:
Following https://github.com/cakephp/cakephp/issues/11023#issuecomment-321171559

`TableLocator` instance passes itself to the table objects it creates. Then the locator is being passed to the associations created by this table object.

This allows to isolate tables created with a particular table locator making independent of a "global" table locator. 

This also makes sure that associations created by tables originating from different locators do not collide making it possible to isolate parts of the system where groups of tables should behave differently in some scenarios, ie share different connections.

For example:

```php
class OrdersTable extends Table
{
    public function initialize()
    {
        $this->hasMany('Articles');
    }
}

class ConnectionAwareLocator extends TableLocator
{
    public function __construct(ConnectionInterface $connection)
    {
        $this->connection = $connection;
    }

    public function get($alias, array $options = [])
    {
        $table = parent::get($alias, $options);
        return $table->setConnection($this->connection);
    }
}

$fooLocator = new ConnectionAwareLocator(ConnectionManager::get('foo'));
$table = $fooLocator->get('Orders');
//this table and it's associations share the "foo" connection


$barLocator = new ConnectionAwareLocator(ConnectionManager::get('bar'));
$table = $barLocator->get('Orders');
//this table and it's associations share the "bar" connection
```

### DIFF
--- a/src/ORM/Locator/TableLocator.php
+++ b/src/ORM/Locator/TableLocator.php
@@ -211,6 +211,9 @@ class TableLocator implements LocatorInterface
             }
             $options['connection'] = ConnectionManager::get($connectionName);
         }
+        if (empty($options['tableLocator'])) {
+            $options['tableLocator'] = $this;
+        }
 
         $options['registryAlias'] = $alias;
         $this->_instances[$alias] = $this->_create($options);
--- a/src/ORM/Table.php
+++ b/src/ORM/Table.php
@@ -35,6 +35,7 @@ use Cake\ORM\Association\HasOne;
 use Cake\ORM\Exception\MissingEntityException;
 use Cake\ORM\Exception\PersistenceFailedException;
 use Cake\ORM\Exception\RolledbackTransactionException;
+use Cake\ORM\Locator\LocatorAwareTrait;
 use Cake\ORM\Rule\IsUnique;
 use Cake\Utility\Inflector;
 use Cake\Validation\ValidatorAwareInterface;
@@ -128,6 +129,7 @@ class Table implements RepositoryInterface, EventListenerInterface, EventDispatc
 {
 
     use EventDispatcherTrait;
+    use LocatorAwareTrait;
     use RulesAwareTrait;
     use ValidatorAwareTrait;
 
@@ -282,6 +284,9 @@ class Table implements RepositoryInterface, EventListenerInterface, EventDispatc
                     $this->setValidator($name, $validator);
                 }
             }
+        }
+        if (!empty($config['tableLocator'])) {
+            $this->setTableLocator($config['tableLocator']);
         }
         $this->_eventManager = $eventManager ?: new EventManager();
         $this->_behaviors = $behaviors ?: new BehaviorRegistry();
--- a/src/ORM/Table.php
+++ b/src/ORM/Table.php
@@ -971,7 +971,10 @@ class Table implements RepositoryInterface, EventListenerInterface, EventDispatc
      */
     public function belongsTo($associated, array $options = [])
     {
-        $options += ['sourceTable' => $this];
+        $options += [
+            'sourceTable' => $this,
+            'tableLocator' => $this->getTableLocator()
+        ];
         $association = new BelongsTo($associated, $options);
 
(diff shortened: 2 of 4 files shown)


lorenzo:
I'm more inclined to say that the `Associations` class is what should be made aware of the locator, and  avoid having the locator be stored in the table class directly.

robertpustulka:
@lorenzo `Association` class is already aware of the locator. 
https://github.com/cakephp/cakephp/blob/master/src/ORM/Association.php#L38

But it falls back to the "global" locator if not set explicitly. In order to achieve "scoped", isolated table you'd need to manually attach the locator to each association. By having the table aware of the locator we can make it easier to use the original locator across associations.

lorenzo:
we can do exactly what you are doing here, but instead of storing the locator in the table, we just pass it to the associations.

I'm not sure there is an actual need for storing the locator in the table

lorenzo:
If you create a table with a particular locator, then that should be the locator stored in the associations class

robertpustulka:
@lorenzo how can we pass the locator over to associations without storing it in the table first?

lorenzo:
You can pass the locator to the constructor, and use it to create the Associations object, without storing it.

robertpustulka:
Ah ok, you're talking about `AssociationCollection` object.

Are you suggesting that the locator should be attached to the association in `AssociationCollection::add()` method?

lorenzo:
Yes, sorry for the confusion.

> Are you suggesting that the locator should be attached to the association in AssociationCollection::add() method?

That is sort of what I had in mind

robertpustulka:
After some thought I think that moving the locator to the collection might raise some issues.

 `AssociationCollection::add()` just adds the instantated object to the collection. If we wanted to set the locator there, it would overwrite the custom locator added as an option in eq `Table::hasMany()` factory method. It would be hard to detect there if a custom locator has been already attached.

```php
//Table
public function initialize()
{
    $this->hasMany('Foos', [
        'tableLocator' -> new TableLocator
    ]);
}

//AssociationCollection
public function add($alias, Association $association)
{
    list(, $alias) = pluginSplit($alias);
    $association->setTableLocator($this->getTableLocator());
        //overwrites a locator injected in `Table::initialize()`

    return $this->_items[strtolower($alias)] = $association;
}
````


Also sometimes behaviors (such as `TranslateBehavior`) use table locators. It would be easier to attach the custom locator from the table.

```php
public function initialize()
{
    $this->addBehavior('Translate', [
        'tableLocator' -> $this->getTableLocator()
    ]);
}
```

It would also make possible to modify the behavior to use the locator from a table it is attached to.

makallio85:
Is this going to be included in 3.6.0, this is really something, that I would like to have :D

lorenzo:
@robertpustulka I see what you mean now. Still I'm not fond at all of having the locator in each table, it feels like the wrong design:

In what occasions might a Table need a locator? Mainly when it's collaborators need to deal with associations... For example when a behaviour needs to crate a new association, or when a validator needs to check on a related table, or when the Associations need to load the associated table.

It is never the table that needs the locator by itself, it's always the collaborators. In that sense, it is obvious that the easiest place the locator is in the table itself, but I think this give the wrong signal to developers, that will likely be using it inside of the table itself.

robertpustulka:
@lorenzo I'll try to figure out how to use the locator from collection and still allow for overwriting the default one.

markstory:
@makallio85 I am curious, what about the current registry causes you trouble?

robertpustulka:
There's an alternative approach: #11172

makallio85:
@markstory I'll write small example shortly

makallio85:
@markstory Here you can find example project and methods demonstrating the actual problem I am having

https://github.com/makallio85/connection-example/blob/master/src/Controller/AnimalsController.php

It would so great to have some way to distribute connection to associations as well. But it is not that straightforward. It should be also some how configurable, to what associations to distribute, as it is totally possible that some associations have static connection in project and it should not be changed.

As you can see, my example of deleting data can cause quite nasty bugs.

PR #11073, opened 2017-08-21
Fix notMatching for BelongsToMany with composite keys

chinpei215:
Fixed the following error:
```
SQLSTATE[HY000]: General error: 1 sub-select returns 2 columns - expected 1
```

By the way, Sqlserver and Sqlite (< 3.15.0) don't support row values expression like `(1,2) = (1,2)`, it will raise another error anyway. Perhaps, we could use `NOT EXISTS` instead of `NOT IN`?

### DIFF
--- a/src/ORM/Association/BelongsToMany.php
+++ b/src/ORM/Association/BelongsToMany.php
@@ -513,7 +513,7 @@ class BelongsToMany extends Association
             ->andWhere(function ($exp) use ($subquery, $conds) {
                 $identifiers = [];
                 foreach (array_keys($conds) as $field) {
-                    $identifiers = new IdentifierExpression($field);
+                    $identifiers[] = new IdentifierExpression($field);
                 }
                 $identifiers = $subquery->newExpr()->add($identifiers)->setConjunction(',');
                 $nullExp = clone $exp;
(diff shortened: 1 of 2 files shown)


src/ORM/Association/BelongsToMany.php
```diff
@@ -513,7 +513,7 @@ protected function _appendNotMatching($query, $options)
             ->andWhere(function ($exp) use ($subquery, $conds) {
                 $identifiers = [];
                 foreach (array_keys($conds) as $field) {
-                    $identifiers = new IdentifierExpression($field);
+                    $identifiers[] = new IdentifierExpression($field);
```
lorenzo: wow, I made a sad mistake here

lorenzo:
> By the way, Sqlserver and Sqlite (< 3.15.0) don't support row values expression like (1,2) = (1,2)

I created the `TupleComparison` class for that reason

chinpei215:
> I created the TupleComparison class for that reason

Oh good to hear. So if the TupleComparison could emulate `(1,2) IN (SELECT a, b FROM ...)`, people could execute the query for any databases.

lorenzo:
@chinpei215 I think this is good for merging now. Do you want to try using `TupleComparison` in another pull request?

chinpei215:
Thank you. 

> Do you want to try using TupleComparison in another pull request?

I am not planning to do it at this point.

PR #11075, opened 2017-08-22
3.6 - Add a deprecation warning helper.

markstory:
This will make adding framework deprecation warnings we plan to add in 3.6.x simpler as they will be easier to write.

### DIFF
--- a/src/Core/functions.php
+++ b/src/Core/functions.php
@@ -248,3 +248,16 @@ if (!function_exists('env')) {
     }
 
 }
+
+if (!function_exists('deprecationWarning')) {
+    /**
+     * Helper method for outputting deprecation warnings
+     *
+     * @param string $message The message to output as a deprecation warning.
+     * @return void
+     */
+    function deprecationWarning($message)
+    {
+        trigger_error($message, E_USER_DEPRECATED);
+    }
+}
--- a/src/Core/functions.php
+++ b/src/Core/functions.php
@@ -263,6 +263,7 @@ if (!function_exists('deprecationWarning')) {
         if (!(error_reporting() & E_USER_DEPRECATED)) {
             return;
         }
+
         $trace = debug_backtrace();
         if (isset($trace[$stackFrame])) {
             $frame = $trace[$stackFrame];
--- a/src/Core/functions.php
+++ b/src/Core/functions.php
@@ -260,7 +260,7 @@ if (!function_exists('deprecationWarning')) {
      */
     function deprecationWarning($message, $stackFrame = 2)
     {
-        if (!Configure::read('debug')) {
+        if (!(error_reporting() & E_USER_DEPRECATED)) {
             return;
         }
         $trace = debug_backtrace();
--- a/src/Core/functions.php
+++ b/src/Core/functions.php
@@ -254,10 +254,28 @@ if (!function_exists('deprecationWarning')) {
      * Helper method for outputting deprecation warnings
      *
      * @param string $message The message to output as a deprecation warning.
+     * @param int $stackFrame The stack frame to include in the error. Defaults to 2
+     *   as that should point to application/plugin code.
      * @return void
      */
-    function deprecationWarning($message)
+    function deprecationWarning($message, $stackFrame = 2)
     {
+        if (!Configure::read('debug')) {
+            return;
+        }
+        $trace = debug_backtrace();
+        if (isset($trace[$stackFrame])) {
+            $frame = $trace[$stackFrame];
+            $frame += ['file' => '[internal]', 'line' => '??'];
+
+            $message = sprintf(
+                '%s - %s, line: %s',
+                $message,
+                $frame['file'],
+                $frame['line']
+            );
+        }
+
         trigger_error($message, E_USER_DEPRECATED);
     }
 }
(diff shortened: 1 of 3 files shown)


chinpei215:
I think the function would report unhelpful line number and file name.

inoas:
```php
<?php

    function deprecationWarning($message)
    {
        trigger_error($message, E_USER_DEPRECATED);
    }

deprecationWarning('here');
```

PHP Deprecated:  here in /Users/leo/Downloads/test-error.php on line 5

Deprecated: here in /Users/leo/Downloads/test-error.php on line 5

markstory:
@chinpei215 The HTML error output will include a backtrace. Would you want to see the backtrace in the deprecation warning message as well?

dereuromark:
the log entries would be useful to have the right location and line number.

markstory:
@dereuromark Logging is something that the error handler already does if its configured correctly.

chinpei215:
@markstory But people would have to click links to show stacktraces when they encountered errors.
deprecations

markstory:
@chinpei215 Fair, but they would have to do that anyways as the raw `trigger_error` calls would point to code inside cakephp not their application. We could do what @dereuromark suggested and 'jump up' two stack frames to hopefully get the right line.

chinpei215:
@markstory ah sorry you are right. and it might be nice if we could do that.

markstory:
I think generating a backtrace 2 levels up is reasonable. However, I would want to guard that kind of logic with a debug check. Does that seem reasonable?

chinpei215:
Sounds reasonable to me.

markstory:
Error messages now have file/lines displayed.

gmponos:
IMHO I see more trouble caused rather than value added.

markstory:
@gmponos If we use trigger_error() without a wrapper, the errors will not be able to indicate what application code is using the deprecated functions.

gmponos:
> If we use trigger_error() without a wrapper, the errors will not be able to indicate what application code is using the deprecated functions.

If that's the goal then fine. I maybe rushed judging the goal of this PR because of the initial commit where the only thing I saw was a "proxy" function...

But still your comment raises some more questions for me and I am sorry for that... maybe because I don't follow all discussions.

I believe you are using an ErrorHandler on a CakePHP application isn't this possible to be handled there? Hope I didn't misunderstood and sorry for any trouble.

markstory:
> I believe you are using an ErrorHandler/Debugger on a CakePHP application isn't this possible to be handled there?

We are using an error handler, but that error handler relies on errors coming out of `trigger_error()`.  While the current ErrorHandler does show a stack trace in the HTML output, the CLI output does not include stack trace information. Having to click into the stack trace for HTML messages, and having not hints at all in CLI contexts are some of the reasons I think this function could be useful to people upgrading.

src/Core/functions.php
```diff
@@ -254,10 +254,28 @@ function env($key, $default = null)
...
      *
      * @param string $message The message to output as a deprecation warning.
+     * @param int $stackFrame The stack frame to include in the error. Defaults to 2
+     *   as that should point to application/plugin code.
      * @return void
      */
-    function deprecationWarning($message)
+    function deprecationWarning($message, $stackFrame = 2)
     {
+        if (!Configure::read('debug')) {
```
dereuromark: If we do this, then we wont be able to find any left overs in production.

markstory: Yes, but I'm not comfortable adding potentially 100s of calls to `debug_backtrace()` to people's applications in production mode. Would it be better to check the error level here instead?

dereuromark: Yeah, I think so.

inoas: I think we should have a separate flag for this. Something along `log3xDeprecations` (default true if not set), which should go into a dedicated log file. That way it is easy to spot in production and easy to disable if you just want to run your app and not update it.

src/Core/functions.php
```diff
@@ -248,3 +248,34 @@ function env($key, $default = null)
...
+     * @param int $stackFrame The stack frame to include in the error. Defaults to 2
+     *   as that should point to application/plugin code.
+     * @return void
+     */
+    function deprecationWarning($message, $stackFrame = 2)
+    {
+        if (!(error_reporting() & E_USER_DEPRECATED)) {
+            return;
+        }
+        $trace = debug_backtrace();
```
ravage84: May be a white line above

tests/TestCase/Core/FunctionsTest.php
```diff
@@ -42,4 +42,39 @@ public function testEnv()
         $this->assertEquals('0', env('ZERO'));
         $this->assertEquals('0', env('ZERO', '1'));
     }
+
+    /**
+     * Test error messages coming out when debug is on
```
ravage84: "Test error messages coming out when debug is on, manually setting the stack frame"

tests/TestCase/Core/FunctionsTest.php
```diff
@@ -42,4 +42,39 @@ public function testEnv()
...
+     * @expectedExceptionMessage This is going away - [internal], line: ??
+     */
+    public function testDeprecationWarningEnabled()
+    {
+        error_reporting(E_ALL);
+        deprecationWarning('This is going away', 1);
+    }
+
+    /**
+     * Test error messages coming out when debug is on
```
ravage84: "Test error messages coming out when debug is on, not setting the stack frame manually"

ravage84 (commented):
Beautiful!

PR #11076, opened 2017-08-22
Fix Collection throwing an exception when Traversable given.

chinpei215:
Since `Collection::unwrap()` may return `Traversable`, the return value cannot be passed to `LimitIterator` directly. Also, I changed the `first()` method to not use `take()` for two reasons. The first  reason is that it could improve the performance. The second reason is that the `last()` method no longer uses `take()` because generators cannot be passed to `iterator_count()`.


lorenzo:
`unwrap` should no return a `Traversable` I would say the problem was there, instead of the code you changed

lorenzo:
The changes look good, though

lorenzo:
Great job @chinpei215, I think this is a really good change

chinpei215:
@lorenzo Thank you!

PR #11078, opened 2017-08-23
Revert "Replace parse_url() with a regex.", and add fallback code to parseDsn()

chinpei215:
Reported in the japanese channel on the Slack.

As of 3.4.13 / 3.5.0, parseDsn() no longer accepts empty passwords:
```php
mysql://user:@localhost:3306/database
```

Since it is hard to distinguish 'empty string' from unmatch in the regular expression, I reverted StaticConfigTrait to the previous code before the change in 519b0ce66f3c43206ed9b8e026a928ef2339d3ff. Instead, I added fallback code to parseDsn() and more test cases.

### DIFF
--- a/src/Core/StaticConfigTrait.php
+++ b/src/Core/StaticConfigTrait.php
@@ -237,7 +237,7 @@ trait StaticConfigTrait
      *
      * @param string $dsn The DSN string to convert to a configuration array
      * @return array The configuration array to be stored after parsing the DSN
-     * @throws \InvalidArgumentException If not passed a string, or passed an invalid string
+     * @throws \InvalidArgumentException If not passed a string
      */
     public static function parseDsn($dsn)
     {
@@ -249,23 +249,18 @@ trait StaticConfigTrait
             throw new InvalidArgumentException('Only strings can be passed to parseDsn');
         }
 
-        $pattern = '/^(?P<scheme>[\w\\\\]+):\/\/((?P<username>.*?)(:(?P<password>.*?))?@)?' .
-            '((?P<host>[^?#\/:@]+)(:(?P<port>\d+))?)?' .
-            '(?P<path>\/[^?#]*)?(\?(?P<query>[^#]*))?(#(?P<fragment>.*))?$/';
-        preg_match($pattern, $dsn, $parsed);
-
-        if (!$parsed) {
-            throw new InvalidArgumentException("The DSN string '{$dsn}' could not be parsed.");
-        }
-        foreach ($parsed as $k => $v) {
-            if (is_int($k)) {
-                unset($parsed[$k]);
-            }
-            if ($v === '') {
-                unset($parsed[$k]);
-            }
+        $scheme = '';
+        if (preg_match("/^([\w\\\]+)/", $dsn, $matches)) {
+            $scheme = $matches[1];
+            $dsn = preg_replace("/^([\w\\\]+)/", 'file', $dsn);
         }
 
+        $parsed = parse_url($dsn);
+        if ($parsed === false) {
+            return $dsn;
+        }
+
+        $parsed['scheme'] = $scheme;
         $query = '';
 
         if (isset($parsed['query'])) {
@@ -285,6 +280,15 @@ trait StaticConfigTrait
             }
         }
 
+        if (isset($parsed['user'])) {
+            $parsed['username'] = $parsed['user'];
+        }
+
+        if (isset($parsed['pass'])) {
+            $parsed['password'] = $parsed['pass'];
+        }
+
+        unset($parsed['pass'], $parsed['user']);
         $parsed = $queryArgs + $parsed;
 
         if (empty($parsed['className'])) {
--- a/src/Core/StaticConfigTrait.php
+++ b/src/Core/StaticConfigTrait.php
@@ -237,7 +237,7 @@ trait StaticConfigTrait
      *
      * @param string $dsn The DSN string to convert to a configuration array
      * @return array The configuration array to be stored after parsing the DSN
-     * @throws \InvalidArgumentException If not passed a string
+     * @throws \InvalidArgumentException If not passed a string, or passed an invalid string
      */
     public static function parseDsn($dsn)
     {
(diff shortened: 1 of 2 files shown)


src/Core/StaticConfigTrait.php
```diff
@@ -249,23 +249,53 @@ public static function parseDsn($dsn)
...
         }
-        foreach ($parsed as $k => $v) {
-            if (is_int($k)) {
-                unset($parsed[$k]);
+
+        $parsed = parse_url('file://' . $noScheme);
+        if ($parsed === false) {
+            // Failed to parse.
+            // Try to encode the userinfo if the DSN string contains '@' marks.
+            for ($p = 0; ($p = strpos($noScheme, '@', $p)) !== false; ++$p) { // @codingStandardsIgnoreLine
```
chinpei215: I added this notation to avoid 'Avoid function calls in a FOR loop test part' warning.

chinpei215:
Actual diff of StaticConfigTrait.php:
```diff
diff --git a/src/Core/StaticConfigTrait.php b/src/Core/StaticConfigTrait.php
index c627b55..afab4e5 100644
--- a/src/Core/StaticConfigTrait.php
+++ b/src/Core/StaticConfigTrait.php
@@ -237,7 +237,7 @@ trait StaticConfigTrait
      *
      * @param string $dsn The DSN string to convert to a configuration array
      * @return array The configuration array to be stored after parsing the DSN
-     * @throws \InvalidArgumentException If not passed a string
+     * @throws \InvalidArgumentException If not passed a string, or passed an invalid string
      */
     public static function parseDsn($dsn)
     {
@@ -249,15 +249,50 @@ trait StaticConfigTrait
             throw new InvalidArgumentException('Only strings can be passed to parseDsn');
         }
 
-        $scheme = '';
-        if (preg_match("/^([\w\\\]+)/", $dsn, $matches)) {
-            $scheme = $matches[1];
-            $dsn = preg_replace("/^([\w\\\]+)/", 'file', $dsn);
+        $scheme = $noScheme = '';
+        if (preg_match('{^(?<scheme>[\w\\\]+)(:(//)?)(?P<noScheme>(.*))}', $dsn, $matches)) {
+            $scheme = $matches['scheme'];
+            $noScheme = $matches['noScheme'];
         }
 
-        $parsed = parse_url($dsn);
+        $parsed = parse_url('file://' . $noScheme);
         if ($parsed === false) {
-            return $dsn;
+            // Failed to parse.
+            // Try to encode the userinfo if the DSN string contains '@' marks.
+            for ($p = 0; ($p = strpos($noScheme, '@', $p)) !== false; ++$p) {
+                $user = substr($noScheme, 0, $p);
+                $rest = substr($noScheme, $p + 1);
+
+                $withoutUser = parse_url('file://' . $rest);
+                if ($withoutUser === false) {
+                    continue;
+                }
+
+                $user = implode(':', explode('%3A', urlencode($user), 2));
+                $withUser = parse_url('file://' . $user . '@' . $rest);
+                if (!isset($withUser['user'])) {
+                    continue;
+                }
+
+                // Ensure other url components match the result of without the userinfo.
+                $check = $withUser;
+                unset($check['user'], $check['pass']);
+                if ($check !== $withoutUser) {
+                    continue;
+                }
+
+                $withUser['user'] = urldecode($withUser['user']);
+                if (isset($withUser['pass'])) {
+                    $withUser['pass'] = urldecode($withUser['pass']);
+                }
+
+                $parsed = $withUser;
+                break;
+            }
+
+            if (!$parsed) {
+                throw new InvalidArgumentException("The DSN string '{$dsn}' could not be parsed.");
+            }
         }
 
         $parsed['scheme'] = $scheme;
```

tests/TestCase/Datasource/ConnectionManagerTest.php
```diff
@@ -306,6 +306,31 @@ public function dsnProvider()
                     'log' => '1'
                 ]
             ],
+            'no password' => [
+                'mysql://user@localhost:3306/database',
```
markstory: Is the password supposed to be '' here or `null`? Don't both turn into `''` when sent to the driver?

src/Core/StaticConfigTrait.php
markstory: Would the code remain simpler if we fall back to the regex if `parse_url()` fails? Wouldn't that let us cover the same scenarios but with less code?

tests/TestCase/Datasource/ConnectionManagerTest.php
chinpei215: No. If some password has been already set in the configuration, this DSN doesn't overwrite it.

src/Core/StaticConfigTrait.php
chinpei215: It is difficult for me to write a regex pattern which emulates how parse_url() works properly.

tests/TestCase/Datasource/ConnectionManagerTest.php
markstory: interesting use case

src/Core/StaticConfigTrait.php
markstory: But wouldn't `parse_url()` cover most use cases/scenarios has it historically has? The regex would only be used if the user/password contained `#` or `?` which cause `parse_url()` to behave incorrectly.

tests/TestCase/Datasource/ConnectionManagerTest.php
chinpei215: I misunderstood the behavior a little bit. The related code is:
https://github.com/cakephp/cakephp/blob/8a91fca6b147ebc0eb39c77ef330e357207dfd5a/src/Core/StaticConfigTrait.php#L96-L100

So, probably, people could use a config array like this:
```php
[
    'password' => env('DATABASE_PASSWORD'),
    'url' => 'mysql://user@localhost:3306/database',
]
```

src/Core/StaticConfigTrait.php
chinpei215: I would have to apologize to you. In the first place, the fallback code might be unnecessary. Even if we use `preg_match()`, we could distinguish empty strings from non-specified, by capturing `@`, `:`, `/`, `?` and `#` followed by user, password/port, path, query and fragment. And, since the host part cannot be an empty string, if it is a empty string in the result, then it would mean that the host part is not specified.
So we could enhance the current code without reverting it. We could check whether those characters were captured or not before unsetting empty url components [here](https://github.com/cakephp/cakephp/blob/8a91fca6b147ebc0eb39c77ef330e357207dfd5a/src/Core/StaticConfigTrait.php#L264-L266). The big advantage of this approach is that we could return consistent results in both PHP5 and PHP7. Would you prefer this approach over the current approach in this pull request?

markstory: If it results in simpler/less code without causing incorrect results I'm all for it.

chinpei215: OK. I'll open another pull request to fix this issue.

markstory:
Closing in favour of #11090

PR #11083, opened 2017-08-23
Change docblock to using getters

davidyell:
This docblock shows the use of setter methods, but doesn't show the getter methods, and instead uses the "old school" direct class property approach. Just changed that to show getters instead for consistency.

### DIFF
--- a/src/Datasource/EntityTrait.php
+++ b/src/Datasource/EntityTrait.php
@@ -188,8 +188,8 @@ trait EntityTrait
      *
      * ```
      * $entity->set(['name' => 'andrew', 'id' => 1]);
-     * echo $entity->name // prints andrew
-     * echo $entity->id // prints 1
+     * echo $entity->get('name') // prints andrew
+     * echo $entity->get('id') // prints 1
      * ```
      *
      * Some times it is handy to bypass setter functions in this entity when assigning


dereuromark:
that's not old-school, that's actually inferior. a getter with magic strings is worse imo.

inoas:
The entities are meant to be low-weight stores as far as I know. Directly accessing their properties can be much more speedy AFAIU.

dereuromark:
I care more about correctness, refactor-ability and IDE comparability.

    $entity->get('my_field');

has no IDE support, no existence check etc.

Far better would be to have const in the entity and use those:

    $entity->get($entity::MY_FIELD);

If you provide real methods that would even be better:

    $entity->getMyField();

But until then using the properties at least make it visible in the IDE if those exist (due to the annotations in the entity (which are kept up to date using ide-helper plugin).

davidyell:
>Directly accessing their properties can be much more speedy
I can hook the `get` method very easily without having to find and replace in my whole project.

> that's actually inferior. a getter with magic strings is worse imo.
So is it inferior in your opinion, but not necessarily worse.

> IDE comparability
What tooling you choose is up to you, it's irrelevant to the framework.

> no existence check
[Incorrect](https://github.com/cakephp/cakephp/blob/master/src/Datasource/EntityTrait.php#L281-L283).

> which are kept up to date using ide-helper plugin
Depending on your own plugin is irrelevant to the framework.

dereuromark:
Best practice is opinionated - and derived from some good reasons to do so :)
Anyway, no big deal either way.

davidyell:
If that's the case, then I'll leave it to the core team 😄

markstory:
I thought we generated doc blocks to make IDE's understand the property `__get` behavior?

dereuromark:
Exactly :) But that only works with the properties - not the magic strings and get() usage directly.

markstory:
Closing as the magic properties have documentation hints generated by bake, and can be autocompleted by IDE's more easily than `get(string)` can be.

PR #11087, opened 2017-08-24
Fix response headers set in templates not be sent.

### LINKED ISSUE #11084: Manipulating Response in View Classes is not possible
This is a (multiple allowed):

* [x] bug
* [ ] enhancement

* CakePHP Version: 3.5
* Platform and Target: XAMP on Linux, CakePHP 3.5.0

### What you did
Trying to manipulate the Response object with the new 'with...' functions in a View class.
(e.g. AjaxView)

### What happened
Changed in a View class (e.g. default App\AjaxView)  the $this->response object.
e.g.
$this->response = $this->response->withStatus(401) will not update the Response object in the Controller class

The response object to be changed

### Explanations
The with...-functions return a clone of the Response object which has been passed from Controller class.
Setting $this->response in a View class will change the property $this->response in the View class but will not update the Response object in the Controller class.

I believe it is because the object is cloned and is therefore a copy of the original one.
So when you step up from View back to Controller $this->response in the Controller remain unchanged.

The only way to change the Response is to do it directly in the controller, before rendering.
Changing it inside View classes won't work with the 'with....-functions'

Don't know if this is intention and I'm missing something or don't understand this fully.

If one want's to update the Controller then $this->response inside the View class has to be transfered back into the Controller class, e.g. in Controller->render
```
623 $this->View = $this->createView();
624 $rendered = $this->View->render($view, $layout);
625 $this->response = $this->View->response;
626 $this->response->body($rendered);
```

markstory:
When response properties are set in the view templates, they should show up in the controller's response as well.

Fixes #11084

### DIFF
--- a/src/Controller/Controller.php
+++ b/src/Controller/Controller.php
@@ -622,9 +622,9 @@ class Controller implements EventListenerInterface, EventDispatcherInterface
 
         $this->View = $this->createView();
         $contents = $this->View->render($view, $layout);
-        $this->response = $this->View->response;
+        $this->response = $this->View->response->withStringBody($contents);
 
-        return $this->response->withStringBody($contents);
+        return $this->response;
     }
 
     /**
--- a/src/Controller/Controller.php
+++ b/src/Controller/Controller.php
@@ -621,9 +621,10 @@ class Controller implements EventListenerInterface, EventDispatcherInterface
         }
 
         $this->View = $this->createView();
-        $this->response->body($this->View->render($view, $layout));
+        $contents = $this->View->render($view, $layout);
+        $this->response = $this->View->response;
 
-        return $this->response;
+        return $this->response->withStringBody($contents);
     }
 
     /**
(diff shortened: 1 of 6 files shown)


lorenzo:
Seems like this breaks a bunch of tests. The changes look fine to me though

vonboth:
Also affects 3.4.X ...

markstory:
@vonboth 3.4 is only getting security patches going foward.

nrother:
Am I right that this disabled the possibility to set the response body in the view, instead of returning it? `$response->withStringBody()` looks like it would always overwrite the body, so if `View::render()`return `null`the body is discarded.

This seems to affect the possibility to use a `CallbackStream` as a response for streaming responses. Or is there another possibility for that?

Documentations for the return value of `View::render` says "Rendered content or null if content already rendered and returned earlier." btw.

markstory:
> This seems to affect the possibility to use a CallbackStream as a response for streaming responses. Or is there another possibility for that?

This should still be possible if you don't render a view as well.

PR #11090, opened 2017-08-24
Fix parseDsn() ignoring empty strings

chinpei215:
Another approach compatible with #11078. Uses more named sub-patterns prefixed '_' to distinguish empty strings from non-specified.


markstory (approved):
Looks good!

chinpei215:
Thank you.

PR #11092, opened 2017-08-24
Escape urls to avoid xss

ceeram:
Escape urls to avoid xss

### DIFF
--- a/src/View/Helper/UrlHelper.php
+++ b/src/View/Helper/UrlHelper.php
@@ -144,7 +144,7 @@ class UrlHelper extends Helper
             return $this->build($path, !empty($options['fullBase']));
         }
         if (strpos($path, '://') !== false || preg_match('/^[a-z]+:/i', $path)) {
-            return $path;
+            return ltrim($this->build($path), '/');
         }
         if (!array_key_exists('plugin', $options) || $options['plugin'] !== false) {
             list($plugin, $path) = $this->_View->pluginSplit($path, false);
--- a/src/View/StringTemplate.php
+++ b/src/View/StringTemplate.php
@@ -315,6 +315,9 @@ class StringTemplate
         }
         $truthy = [1, '1', true, 'true', $key];
         $isMinimized = isset($this->_compactAttributes[$key]);
+        if (!preg_match('/\A(\w|[.-])+\z/', $key)) {
+            $key = h($key);
+        }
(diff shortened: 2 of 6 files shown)


src/View/StringTemplate.php
```diff
@@ -315,6 +315,9 @@ protected function _formatAttribute($key, $value, $escape = true)
         }
         $truthy = [1, '1', true, 'true', $key];
         $isMinimized = isset($this->_compactAttributes[$key]);
+        if(!preg_match_all('/\A(\w|[.-])+\z/', $key)) {
```
markstory: Pretty sure this is supposed to be `if (` with a space.

markstory: Do you need preg_match_all or would preg_match work?

ceeram: not sure, im no regex hero, if you think `preg_match` would do, i can change that

ceeram:
fixed cs issue and replaced `preg_match_all` with `preg_match`

dereuromark:
This causes a major regression in all HtmlHelper generated img URLs that contain query strings:
https://travis-ci.org/dereuromark/cakephp-geo/builds/270748293?utm_source=github_status&utm_medium=notification
Those URLs have now double encoded query string params:

    Karte

We need a 3.5.2 to fix this before too many people update and have their apps completely broken.
Maybe also issue a warning in the 3.5.1 release notes not to upgrade to this and skip to the next patch release instead.

markstory:
I think it is relevant to mention that your plugin was HTML encoding URLs before passing them to assetUrl methods in UrlHelper, and that is the source of double encoding.

dereuromark:
Indeed. It was properly handling it on its own. I guess people doing similar handling will have to fix their code now to adhere to the new security standard of this patch release.

src/View/Helper/UrlHelper.php
```diff
@@ -144,7 +144,7 @@ public function assetUrl($path, array $options = [])
             return $this->build($path, !empty($options['fullBase']));
         }
         if (strpos($path, '://') !== false || preg_match('/^[a-z]+:/i', $path)) {
-            return $path;
+            return ltrim($this->build($path), '/');
```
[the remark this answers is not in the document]
lorenzo: Please open an issue, don't comment on old commits

PR #11093, opened 2017-08-24
HTTP_X_FORWARDED_FOR can be spoofed, proxies append to the list, so u…

ceeram:
A proxy will append the client’s real IP address to the HTTP header, so use last ip instead of first ip, which could be manipulated by an attacker


markstory (approved):
Looks good to me.

PR #10034, opened 2017-01-14
Add support for the parseHuge option.

markstory:
Sometimes people need to load huge XML files. Add an option to enable people to enable this flag.

Refs #10031

PR #11094, opened 2017-08-24
Change default for parseHuge to false to avoid OOM on xml expansion

ceeram:
The option was added to allow for enabling (https://github.com/cakephp/cakephp/pull/10034) parseHuge as the docblock also decribes, having this enabled though by default is not recommended.

### DIFF
--- a/src/Utility/Xml.php
+++ b/src/Utility/Xml.php
@@ -107,7 +107,7 @@ class Xml
             'return' => 'simplexml',
             'loadEntities' => false,
             'readFile' => true,
-            'parseHuge' => true,
+            'parseHuge' => false,
         ];
         $options += $defaults;
 


inoas:
Doesn't look BC to me, is it?

ceeram:
Correct, in most cases people won't notice the changed default. The
docblock already stated that you had to enable parseHuge, but the code
defaulted to true. This default behavior is an insecure one, so it is now
changed to false.

2017-08-28 3:42 GMT+02:00 inoas <notifications@github.com>:

ravage84:
@markstory This should be shown very prominently in the release notes, then. So pepole don't trip over it.

ceeram:
I doubt if many users will trip. First off, Most users don't use Xml, those that do use it would only need parseHuge in edge cases. The docs already stated to set parseHuge option to true, even though the code itself didnt need it, because of the default. So most likely for those edge cases people are passing the parseHuge option as true already.

ravage84:
@ceeram thanks for the clarification.

PR #11095, opened 2017-08-25
Ensure input passwords will be hashed even when a user doesn't exist

ceeram:
While the docblock states `Input passwords will be hashed even when a user doesn't exist` in fact it does not. This change ensures passwords are actually hashed even when no user was found to help mitigate timing-based user enumeration attacks.


lorenzo:
Looks good, thanks!

PR #11099, opened 2017-08-25
Do not fallback to insecure randomBytes, should be called explicitly

ceeram:
Throw exception instead of trigger_error. Exceptions can be avoided by calling `insecureRandomBytes()` instead of `randomBytes()` in applications.
This is however a BC breaking change, might not fit in a bugfix release.

### DIFF
--- a/src/Utility/Security.php
+++ b/src/Utility/Security.php
@@ -17,6 +17,7 @@ namespace Cake\Utility;
 use Cake\Utility\Crypto\Mcrypt;
 use Cake\Utility\Crypto\OpenSsl;
 use InvalidArgumentException;
+use RuntimeException;
 
 /**
  * Security Library contains utility methods related to security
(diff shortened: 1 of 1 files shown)


lorenzo:
> Exceptions can be avoided by calling insecureRandomBytes() instead of randomBytes() in applications

Well, the core uses `randomBytes` in many places. That may not be a possible for many, but I think it is the most secure option.

ceeram:
hmm right, if core uses this, then we cannot merge it, unless we try catch everywhere and fallback to insecure, but that defeats the purpose of this PR. Guess we cannot enforce secure until php7 which has random_bytes

markstory:
> Guess we cannot enforce secure until php7 which has random_bytes

We can because ext/openssl is still an option and there are [polyfill libraries](https://packagist.org/packages/paragonie/random_compat) that we could 'suggest' via composer.

markstory:
Moving to 3.6 so we don't surprise people with new exceptions.

PR #11100, opened 2017-08-25
Do not serve hidden files or files in hidden folders

ceeram:
Ignore hidden files and files in hidden folders within AssetMiddleware


inoas:
Thanks!

PR #11101, opened 2017-08-25
Use hmac for token hashes to avoid collisions

ceeram:
Use hmac for token hashes to avoid collisions,
Also includes changes of #11151 now, making token hash user specific to avoid privilege escalation


markstory:
We'll have to mention in the release notes that security component tokens changed, and that existing tokens will become invalid.

ceeram:
Additional PR created targeting this PRs branch see #11151

markstory:
Moving to 3.6 as the risk of breaking user form submissions is high enough.

PR #11102, opened 2017-08-25
Replace md5 with hmac hashes 

ceeram:
Replace md5 with hmac hashes in autolinking to provide unpredictability and resistance to collisions

### DIFF
--- a/src/View/Helper/TextHelper.php
+++ b/src/View/Helper/TextHelper.php
@@ -16,6 +16,7 @@ namespace Cake\View\Helper;
 
 use Cake\Core\App;
 use Cake\Core\Exception\Exception;
+use Cake\Utility\Security;
 use Cake\View\Helper;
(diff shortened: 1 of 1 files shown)


src/View/Helper/TextHelper.php
```diff
@@ -167,7 +168,7 @@ protected function _insertPlaceHolder($matches)
         if (isset($matches['url_bare'])) {
             $match = $matches['url_bare'];
         }
-        $key = md5($match);
+        $key = hash_hmac('sha1', $match, Security::getSalt());
```
markstory: Would using a salted sha1 or sha256 work here?

ceeram: it would, need to change that?

markstory: No. I don't forsee this becoming a performance issue.

PR #11103, opened 2017-08-25
Use hmac for digest nonces

ceeram:
Use hmac for digest nonces

### DIFF
--- a/src/Auth/DigestAuthenticate.php
+++ b/src/Auth/DigestAuthenticate.php
@@ -17,6 +17,7 @@ namespace Cake\Auth;
 use Cake\Controller\ComponentRegistry;
 use Cake\Core\Configure;
 use Cake\Http\ServerRequest;
+use Cake\Utility\Security;
 
 /**
  * Digest Authentication adapter for AuthComponent.
@@ -250,7 +251,7 @@ class DigestAuthenticate extends BasicAuthenticate
     protected function generateNonce()
     {
         $expiryTime = microtime(true) + $this->getConfig('nonceLifetime');
-        $signatureValue = md5($expiryTime . ':' . $this->getConfig('secret'));
+        $signatureValue = hash_hmac('sha256', $expiryTime . ':' . $this->getConfig('secret'), Security::getSalt());
         $nonceValue = $expiryTime . ':' . $signatureValue;
 
         return base64_encode($nonceValue);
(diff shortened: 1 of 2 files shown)


src/Auth/DigestAuthenticate.php
```diff
@@ -250,7 +251,7 @@ public function loginHeaders(ServerRequest $request)
     protected function generateNonce()
     {
         $expiryTime = microtime(true) + $this->getConfig('nonceLifetime');
-        $signatureValue = md5($expiryTime . ':' . $this->getConfig('secret'));
+        $signatureValue = hash_hmac('sha256', $expiryTime . ':' . $this->getConfig('secret'), Security::getSalt());
```
markstory: Don't you have to change the validation of the nonce as well?

ceeram: yes it should, will take care of that

markstory: There is also a generateNonce method in the test case that is still using md5() as well.

ceeram: updated both

tests/TestCase/Auth/DigestAuthenticateTest.php
```diff
@@ -515,7 +515,7 @@ protected function generateNonce($secret = null, $expires = 300, $time = null)
         $secret = $secret ?: Configure::read('Security.salt');
         $time = $time ?: microtime(true);
         $expiryTime = $time + $expires;
-        $signatureValue = md5($expiryTime . ':' . $secret);
+        $signatureValue = hash_hmac('sha1', $expiryTime . ':' . $secret, $secret);
```
timalive: Is this just me, or shouldn't this be sha256?  seems to be causing some trouble for me.

src/Auth/DigestAuthenticate.php
```diff
@@ -276,7 +277,8 @@ protected function validNonce($nonce)
         if ($expires < microtime(true)) {
             return false;
         }
+        $check = hash_hmac('sha1', $expires . ':' . $this->getConfig('secret'), $this->getConfig('secret'));
```
timalive: Is this just me, or shouldn't this be sha256? seems to be causing some trouble for me.

tests/TestCase/Auth/DigestAuthenticateTest.php
markstory: Are you suggesting sha256 for the stronger algorithm?

timalive: no, due to inconsistency with the validNonce function

markstory: I totally missed that :blush:. I'll get it fixed.

timalive: great! Thanks. Had me pulling my hair out.

PR #11105, opened 2017-08-25
Routes shell test updates

markstory:
This new test case class lets us use less code to test shell commands. I've added a new assertion as it made updating the existing tests easier.

cc @jeremyharris

### DIFF
--- a/src/TestSuite/ConsoleIntegrationTestCase.php
+++ b/src/TestSuite/ConsoleIntegrationTestCase.php
@@ -139,6 +139,18 @@ class ConsoleIntegrationTestCase extends TestCase
     }
 
     /**
+     * Asserts that `stdout` is empty
+     *
+     * @param string $message The message to output when the assertion fails.
+     * @return void
+     */
+    public function assertOutputEmpty($message = 'stdout was not empty')
+    {
+        $output = implode(PHP_EOL, $this->_out->messages());
+        $this->assertSame('', $output, $message);
+    }
+
+    /**
      * Asserts `stdout` contains expected output
      *
      * @param string $expected Expected output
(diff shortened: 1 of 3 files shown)


tests/TestCase/Shell/RoutesShellTest.php
```diff
@@ -58,7 +46,22 @@ public function tearDown()
...
+     * @return void
+     */
+    protected function assertOutputContainsRow(array $row)
+    {
+        $row = array_map(function ($cell) {
+            return preg_quote($cell, '/');
+        }, $row);
+        $cells = implode('\s+\|\s+', $row);
+        $pattern = '/' . $cells . '/';
+        $this->assertOutputRegexp($pattern);
```
jeremyharris: I think this is a pretty useful assertion - do you think it should be moved to the test case? Also, should be named `assertOutputContainsRows` (plural) as it checks for multiple rows.

src/TestSuite/ConsoleIntegrationTestCase.php
```diff
@@ -191,6 +191,18 @@ public function assertErrorRegExp($pattern, $message = '')
...
+     * Asserts that `stderr` is empty
+     *
+     * @param string $message The message to output when the assertion fails.
+     * @return void
+     */
+    public function assertErrorEmpty($message = 'stderr was not empty')
+    {
+        $output = implode(PHP_EOL, $this->_err->messages());
+        $this->assertSame('', $output, $message);
+    }
```
jeremyharris: I imagine people might expect `assertOutputEmpty` too :)

jeremyharris (commented):
Sweet! Looks good.

Do you think it'd be useful to add `$this->assertExitCode(Shell::CODE_SUCCESS);` to each of them as you expect they're successful? Except the ones that fail, of course.

src/TestSuite/ConsoleIntegrationTestCase.php
markstory: Ok I can add that too.

tests/TestCase/Shell/RoutesShellTest.php
markstory: I can move it up.

markstory: Won't a plural method name imply it can check for multiple rows? Right now I think it will only find one row.

PR #11109, opened 2017-08-27
Add Form::setErrors().

### LINKED ISSUE #10971: Add a setErrors() method to the Form class
This is a (multiple allowed):

* [x] enhancement

* CakePHP Version: 3.x

### What you did

Tried to set form errors on a form using `setErrors`

### What happened

Missing method exception.

 It works.

### Why is this useful?

If you are attempting to persist errors between redirects, you can very easily apply a set of errors to an entity, but cannot do so to a Form instance without a custom setErrors method. There is an `errors` method which only returns it, unlike in the `EntityInterface` which can do setting and getting of an error or set of errors.

ADmad:
Closes #10971

### DIFF
--- a/src/Form/Form.php
+++ b/src/Form/Form.php
@@ -159,6 +159,28 @@ class Form
     }
 
     /**
+     * Set the errors in the form.
+     *
+     * ```
+     * $errors = [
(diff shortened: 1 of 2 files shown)


markstory (approved):
Looks good to me.

Issue #11106, opened 2017-08-26
ValidatorAwareTrait shouldn't use self::BUILD_VALIDATOR_EVENT


lilHermit:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.5.0

### What you did
Used an Object with uses the ValidatorAwareTrait for example ([Type](https://github.com/cakephp/elastic-search/blob/master/src/Type.php))

### What happened
Get Error "Undefined class constant 'BUILD_VALIDATOR_EVENT'"

### What you expected to happen
No error ;-)

ValidatorAwareTrait should use `Table::BUILD_VALIDATOR_EVENT` or the const should be in ValidatorAwareTrait and Table references `ValidatorAwareTrait::BUILD_VALIDATOR_EVENT`. Another alternative is put the const in `ValidationAwareInterface`. Then ([Type](https://github.com/cakephp/elastic-search/blob/master/src/Type.php)) can `implement ValidatorAwareInterface` this also means the duplicated `DEFAULT_VALIDATOR` can be removed. 

I'll create a pull request once it's decided which way round it should be moved to.

Discuss ;-)

lilHermit:
My vote is for `BUILD_VALIDATOR_EVENT` to be in `ValidatorAwareTrait` as it isn't used anywhere else

markstory:
Traits cannot define constants. The interface would work though.

lilHermit:
@markstory ofc! I'll create a PR and one for `cakephp/elastic-search`

robertpustulka:
What would be the name of a default event defined in `ValidationAwareInterface::BUILD_VALIDATOR_EVENT`? 

Also https://github.com/cakephp/cakephp/blob/master/src/Validation/ValidatorAwareTrait.php#L174 needs to be changed to use `static::BUILD_VALIDATOR_EVENT` if we want to allow to override the constant.

burzum:
@lilHermit it's already fixed in the ES plugin. I had the problem there some time ago https://github.com/cakephp/elastic-search/pull/132/files

lilHermit:
Closed as PR created

PR #11113, opened 2017-08-28
Fix ValidationAwareTrait relying on Table const

lilHermit:
as discussed in #11106

### DIFF
--- a/src/Validation/ValidatorAwareInterface.php
+++ b/src/Validation/ValidatorAwareInterface.php
@@ -25,7 +25,7 @@ interface ValidatorAwareInterface
      *
      * @var string
      */
-    const BUILD_VALIDATOR_EVENT = 'Event.buildValidator';
+    const BUILD_VALIDATOR_EVENT = 'Validation.buildValidator';
 
     /**
      * Name of default validation set.
--- a/src/Validation/ValidatorAwareInterface.php
+++ b/src/Validation/ValidatorAwareInterface.php
@@ -25,7 +25,7 @@ interface ValidatorAwareInterface
      *
      * @var string
      */
-    const BUILD_VALIDATOR_EVENT = 'Model.buildValidator';
+    const BUILD_VALIDATOR_EVENT = 'Event.buildValidator';
 
     /**
      * Name of default validation set.
@@ -33,6 +33,13 @@ interface ValidatorAwareInterface
      * @var string
      */
     const DEFAULT_VALIDATOR = 'default';
+
+    /**
+     * The alias this object is assigned to validators as.
+     *
+     * @var string
+     */
+    const VALIDATOR_PROVIDER_NAME = 'collection';
 
     /**
      * Returns the validation rules tagged with $name.
--- a/src/Validation/ValidatorAwareInterface.php
+++ b/src/Validation/ValidatorAwareInterface.php
@@ -21,6 +21,13 @@ interface ValidatorAwareInterface
 {
 
     /**
+     * The name of the event dispatched when a validator has been built.
+     *
+     * @var string
+     */
+    const BUILD_VALIDATOR_EVENT = 'Model.buildValidator';
+
+    /**
      * Name of default validation set.
      *
      * @var string
--- a/src/ORM/Table.php
+++ b/src/ORM/Table.php
@@ -139,6 +139,13 @@ class Table implements RepositoryInterface, EventListenerInterface, EventDispatc
     const VALIDATOR_PROVIDER_NAME = 'table';
 
     /**
+     * The name of the event dispatched when a validator has been built.
+     *
+     * @var string
+     */
+    const BUILD_VALIDATOR_EVENT = 'Model.buildValidator';
+
+    /**
      * The rules class name that is used.
      *
      * @var string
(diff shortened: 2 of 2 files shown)


src/Validation/ValidatorAwareInterface.php
```diff
@@ -21,6 +21,13 @@
 {
 
     /**
+     * The name of the event dispatched when a validator has been built.
+     *
+     * @var string
+     */
+    const BUILD_VALIDATOR_EVENT = 'Model.buildValidator';
```
robertpustulka: I don't think this can be `Model.buildValidator` as the trait can be used outside model scope.

dereuromark: But is it then still BC if you change the value?

robertpustulka: You can keep the old value in `Table` class and add some generic name in the interface. The constant will get overridden in a class.

lilHermit: What is the actual purpose of the const? It's only ever referenced by the Trait!

dereuromark: That you can refer to it via class const instead of relying on magic strings.
This helps to avoid mistakes in spelling and eases migration/updates as the IDE knows about usage and what to rename.

robertpustulka: You can as well use the trait in your class and define different event name using the const instead of rewriting the code.

Also [this line](https://github.com/cakephp/cakephp/blob/master/src/Validation/ValidatorAwareTrait.php#L174) needs to be changed to allow for late static binding.

lilHermit: ok I'll let you guys decide where it goes but currently it breaks implementations where `Table` is not used. For example ([Type](https://github.com/cakephp/elastic-search/blob/master/src/Type.php))

lilHermit: So what you're suggesting is (1) leave in `Table` (2) still have in `ValidatorAwareInterface` but change the value to something generic like `Event.buildValidator` and other non-`Table` implementations carry on as suggested and implement `ValidatorAwareInterface` or can override?

robertpustulka: OK, so it needs to be fixed there. The class defines some validation related constants, but not the one that has been added recently.

Maybe the remaining constants also should be moved to the interface?

robertpustulka: Yes, that's what I'm suggesting. Have default values in an interface and allow those to be overriden in classes.

lilHermit: Yeah I agree all the validation related consts should be moved to the `ValidatorAwareInterface`

`BUILD_VALIDATOR_EVENT`, `VALIDATOR_PROVIDER_NAME` and `DEFAULT_VALIDATOR`

src/Validation/ValidatorAwareInterface.php
```diff
@@ -25,7 +25,7 @@
      *
      * @var string
      */
-    const BUILD_VALIDATOR_EVENT = 'Model.buildValidator';
+    const BUILD_VALIDATOR_EVENT = 'Event.buildValidator';
```
markstory: `Validation.buildValidator` might be a better name.

markstory:
Builds are failing with

```
PHP Fatal error:  Cannot inherit previously-inherited or override constant BUILD_VALIDATOR_EVENT from interface Cake\Validation\ValidatorAwareInterface in /home/travis/build/cakephp/cakephp/src/ORM/Table.php on line 127
```

robertpustulka:
Looks like interface constants cannot be overridden:

> It's possible for interfaces to have constants. Interface constants works exactly like class constants except they cannot be overridden by a class/interface that inherits them.
http://php.net/manual/en/language.oop5.interfaces.php

Sorry for the invalid suggestion, I wasn't aware of that.

How can we "fix" this and still allow for custom event names?

lilHermit:
Remove the constants and change to interface methods which return the string value?!? 

```
public function getBuildValidatorEvent();

public function getDefaultValidator();

public function getValidatorProviderName();
```

robertpustulka:
@lilHermit we can't add new methods to the interface I'm afraid due to the BC reasons.

Maybe we could check if a class constant is defined and fallback to the default value if it does not.

```php
$event = defined(static::BUILD_VALIDATOR_EVENT) ? static::BUILD_VALIDATOR_EVENT : 'Validation.buildValidator`;
```

lilHermit:
@robertpustulka 

Certainly an option, are we going to do the same for `DEFAULT_VALIDATOR` and `VALIDATOR_PROVIDER_NAME` where possible?

markstory:
The trait can use constants, and one of the requirements of  implementing the trait is that the implementer must provide those constants. We've had a few traits in the past that required methods to be implemented by the consuming class.

robertpustulka:
Maybe we just need to update the trait's docblock. The information about `BUILD_VALIDATOR_EVENT` [is missing there](https://github.com/lilHermit/cakephp-cakephp/blob/0b5155e46c10fc770f86b0972bf2adba095cf4e2/src/Validation/ValidatorAwareTrait.php#L28-L32).

markstory:
Additional constant documented in 61e19f3ba779d90bc84353a8c75fe488f05aa05c

PR #11115, opened 2017-08-28
Adding a CSP Middleware

burzum:
This adds a first iteration of a CSP middleware based on https://github.com/paragonie/csp-builder

Be aware that this feature *requires* php 7. This has been discussed with @markstory before:

> [21:03] 
> Hmm the main problem I have is anyway the php7 requirement.
> [21:03] 
> Can we use it if we just suggest it?
> [21:06] 
> markstory I guess we could

Appveyor will fail on setup because it only features php 5.6. So ignore that CI service.


src/Http/Middleware/CspMiddleWare.php
```diff
@@ -0,0 +1,80 @@
...
+     * CSP Builder
+     *
+     * @var \ParagonIE\CSPBuilder\
+     */
+    protected $csp;
+
+    /**
+     * Constructor
+     *
+     * @param null|\ParagonIE\CSPBuilder\
```
ADmad: Shouldn't this include `array` too?

tests/TestCase/Http/Middleware/CspMiddlewareTest.php
```diff
@@ -0,0 +1,68 @@
...
+     * @return void
+     */
+    public function testInvoke() {
+        $request = new ServerRequest();
+        $response = new Response();
+        $callable = function($request, $response) {
+            return $response;
+        };
+
+        $middleware = new CspMiddleware([
```
ADmad: Need a test passing `\ParagonIE\CSPBuilder\` instance too to the constructor.

src/Http/Middleware/CspMiddleWare.php
burzum: Yes, I've had the idea of supporting an array after I wrote that. ;)

tests/TestCase/Http/Middleware/CspMiddlewareTest.php
burzum: Yes, I'll wait with further changes until I've got more feedback if the approach of the implementation will be accepted.

composer.json
```diff
@@ -38,11 +38,13 @@
...
         "ext-openssl": "To use Security::encrypt() or have secure CSRF token generation.",
-        "lib-ICU": "The intl PHP library, to use Text::transliterate() or Text::slug()"
+        "lib-ICU": "The intl PHP library, to use Text::transliterate() or Text::slug()",
+        "paragonie/csp-builder": "CSP builder, to use the CSP Middleware"
     },
     "require-dev": {
         "phpunit/phpunit": "^5.7|^6.0",
-        "cakephp/cakephp-codesniffer": "^3.0"
+        "cakephp/cakephp-codesniffer": "^3.0",
+        "paragonie/csp-builder": "^2.1"
```
markstory: Won't this break builds on php5 going forwards? Could we conditionally include this dependency in travis instead of breaking the php5 builds?

src/Http/Middleware/CspMiddleWare.php
```diff
@@ -0,0 +1,80 @@
...
+     */
+    public function __construct($csp = null)
+    {
+        if (!is_array($csp)) {
+            $cspConfig = (array)Configure::read('App.CSP');
```
markstory: Why would this be an application config value? Do you see the CSP header varying by environment?

composer.json
burzum: We can do that. Any idea how? I'm not really good with the scripts for Travis and don't want to produce 10 or 20 trial and error commits just to get it working.

src/Http/Middleware/CspMiddleWare.php
ADmad: In general too let's not make new classes directly read from `Configure::read()` if possible. Even if someone is saving the setting in `Configure` they can easily do `new CspMiddleware(Configure::read('App.CSP'))`.

burzum: It's not creating a new class based on it, it's just passing the configuration to the new instance.

ADmad: There's no problem using `Configure` to store configure. But there's no need to use `Configure::read()` inside the class constructor where you can easily read and pass the config array as argument.

For datasouces also we do the [same](https://github.com/cakephp/app/blob/master/config/bootstrap.php#L146). Read the config array and pass it to `ConnectionManager::setConfig()`.

ADmad:
I have made a couple of fixes and update travis config to load `paragonie/csp-builder` only for PHP 7.x.

src/Http/Middleware/CspMiddleWare.php
burzum: Ah OK, I get your point now. But the problem is then, even if we provide defaults, the user would still have to be aware of them and manually pass them. I was looking for a convenient way to make CSP work out of the box with some good default settings.

src/Http/Middleware/CspMiddleware.php
```diff
@@ -36,7 +36,7 @@ class CspMiddleware
     /**
      * Constructor
      *
-     * @param null|\ParagonIE\CSPBuilder\
+     * @param null|\ParagonIE\CSPBuilder\ $csp CSP object or config array
```
dereuromark: The trailing \ is too much

src/Http/Middleware/CspMiddleWare.php
markstory: We can make it work out of the box by updating the application skeleton as needed.

burzum:
@markstory @ADmad what is left here? The two tests that fail are failing because of something related to the connection interface which I haven't even touched at all in this PR.

cleptric:
@burzum I fixed this in another PR. So just rebase 3.next into your branch or someone else could just merge it 🙂

PR #11117, opened 2017-08-29
Update documentation to be clearer on how marshalling works

markstory:
Document that setters will not be triggered, and fields will not be marked dirty if the value is an identical scalar and an entity is updated.

### DIFF
--- a/src/ORM/Table.php
+++ b/src/ORM/Table.php
@@ -2490,6 +2490,11 @@ class Table implements RepositoryInterface, EventListenerInterface, EventDispatc
      *
      * You can use the `Model.beforeMarshal` event to modify request data
      * before it is converted into entities.
+     *
+     * When patching scalar values (null/booleans/string/integer/float), if the property
+     * presently has an identical value, the setter will not be called, and the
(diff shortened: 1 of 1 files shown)


src/ORM/Table.php
```diff
@@ -2490,6 +2490,11 @@ public function newEntities(array $data, array $options = [])
      *
      * You can use the `Model.beforeMarshal` event to modify request data
      * before it is converted into entities.
+     *
+     * When patching scalar values (null/booleans/string/integer/float), if the property
+     * presently has an identical value, the setter will not be called, and the
+     * property will not be marked as dirty. This is an optimization to prevent unnecessary field
+     * updates when persisting entities.
```
inoas: ... maybe add something like:
"If you want to run setters, either run them manually or mark fields dirty explicitly, use `EntityInferface::setDirty()`"?

markstory: Why? I was aiming to document that on no-op updates properties are not dirtied.

inoas: Because custom setters could do more than a no-op. But yes, well designed a single field no-change should have no side effects (similar to idempotency) - so pls ignore my comment.

PR #11121, opened 2017-08-30
Fix CookieCollection::parseSetCookieHeader when the cookie has proper…

supermanner:
When the Cookie has 「version」 property, `CookieCollecion::parseSetCookieHeader` occured Undefined index: version.

I used `Cake\Http\Client`, so the Cookie of the returned Response has value version property....
(in case of in-App Purchase API)

### DIFF
--- a/tests/TestCase/Http/Cookie/CookieCollectionTest.php
+++ b/tests/TestCase/Http/Cookie/CookieCollectionTest.php
@@ -436,7 +436,7 @@ class CookieCollectionTest extends TestCase
         $header = [
             'http=name; HttpOnly; Secure;',
             'expires=expiring; Expires=Wed, 15-Jun-2022 10:22:22; Path=/api; HttpOnly; Secure;',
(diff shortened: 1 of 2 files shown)


tests/TestCase/Http/Cookie/CookieCollectionTest.php
```diff
@@ -436,7 +436,7 @@ public function testCreateFromHeader()
         $header = [
             'http=name; HttpOnly; Secure;',
             'expires=expiring; Expires=Wed, 15-Jun-2022 10:22:22; Path=/api; HttpOnly; Secure;',
-            'expired=expired; Expires=Wed, 15-Jun-2015 10:22:22;',
+            'expired=expired; version=1; Expires=Wed, 15-Jun-2015 10:22:22;',
```
markstory: Should the test ensure that this attribute isn't in the Cookie object that is created?

supermanner: Sorry!!
I think so.

markstory:
Thank you!

PR #11124, opened 2017-08-30
Stacking messages in SessionComponent::setFlash

jeremyharris:
In my last attempt to fix flash stacking being BC with the SessionHelper, I accidentally broke compatibility between `SessionComponent` and the `SessionHelper` :(

This PR fixes compatibility between those two deprecated components - all the different components and helpers should be using the same format now.

### DIFF
--- a/lib/Cake/Controller/Component/SessionComponent.php
+++ b/lib/Cake/Controller/Component/SessionComponent.php
@@ -137,11 +137,11 @@ class SessionComponent extends Component {
  */
 	public function setFlash($message, $element = 'default', $params = array(), $key = 'flash') {
 		$messages = (array)CakeSession::read('Message.' . $key);
-		$messages[] = [
+		$messages[] = array(
 			'message' => $message,
 			'element' => $element,
 			'params' => $params,
-		];
+		);
 		CakeSession::write('Message.' . $key, $messages);
 	}
 
--- a/lib/Cake/Controller/Component/SessionComponent.php
+++ b/lib/Cake/Controller/Component/SessionComponent.php
@@ -136,7 +136,13 @@ class SessionComponent extends Component {
  * @deprecated 3.0.0 Since 2.7, use the FlashComponent instead.
  */
 	public function setFlash($message, $element = 'default', $params = array(), $key = 'flash') {
-		CakeSession::write('Message.' . $key, compact('message', 'element', 'params'));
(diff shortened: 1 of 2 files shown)


lib/Cake/Test/Case/Controller/Component/SessionComponentTest.php
```diff
@@ -247,16 +247,13 @@ public function testSessionFlash() {
...
 
 		$Session->setFlash('This is a test message', 'test', array('name' => 'Joel Moss'));
-		$this->assertEquals(array('message' => 'This is a test message', 'element' => 'test', 'params' => array('name' => 'Joel Moss')), $Session->read('Message.flash'));
+		$this->assertEquals(array('message' => 'This is a test message', 'element' => 'test', 'params' => array('name' => 'Joel Moss')), $Session->read('Message.flash.1'));
 
 		$Session->setFlash('This is a test message', 'default', array(), 'myFlash');
-		$this->assertEquals(array('message' => 'This is a test message', 'element' => 'default', 'params' => array()), $Session->read('Message.myFlash'));
-
-		$Session->setFlash('This is a test message', 'non_existing_layout');
-		$this->assertEquals(array('message' => 'This is a test message', 'element' => 'default', 'params' => array()), $Session->read('Message.myFlash'));
```
jeremyharris: I deleted this last test as it was just a duplicate of the first assertion in this case.

lib/Cake/Controller/Component/SessionComponent.php
```diff
@@ -136,7 +136,13 @@ public function error() {
  * @deprecated 3.0.0 Since 2.7, use the FlashComponent instead.
  */
 	public function setFlash($message, $element = 'default', $params = array(), $key = 'flash') {
-		CakeSession::write('Message.' . $key, compact('message', 'element', 'params'));
+		$messages = (array)CakeSession::read('Message.' . $key);
+		$messages[] = [
```
markstory: Short arrays are not compatible with PHP 5.3

markstory (changes requested):

lib/Cake/Controller/Component/SessionComponent.php
jeremyharris: Thanks for catching that.

markstory: It was travis, I just read the build logs.

Issue #11089, opened 2017-08-24
Translation fallback to msgid_plural not working


burzum:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.4.13

### What you did

Ran tests of a helper that uses translations.

The code uses this line for translations `'name' => __d('wa', 'Projects'),` Please note that we're using the *singular* translation here and pay attention to the paste of our po file.

### What happened

`'name' => __d('wa', 'Projects'),` doesn't return a non empty string.

### What you expected to happen

Having plural translations falling back to the msgstr. I think this is somehow related to the order in the po file or something. Because we have `msgid "Projects"` and the plural `msgid "Project"`. See the pasted part of our po file below.

I *guess* this is probably related to the way the translated strings are cached inside the translator object.

Doesn't work, empty output.

```
msgid "Projects"
msgstr ""

#: Model/Search.php:95
msgid "Daily News"
msgstr ""

#: Model/Constants/AwardTypes.php:19
msgid "Personal"
msgstr ""

#: Model/Constants/AwardTypes.php:20
#: Template/Element/Competitions/form.ctp:50;52
#: Template/Element/Profiles/panel.ctp:92
msgid "Project"
msgid_plural "Projects"
msgstr[0] ""
msgstr[1] ""
```

Works:

```
msgid "Projects"
msgstr ""

#: Model/Search.php:95
msgid "Daily News"
msgstr ""

#: Model/Constants/AwardTypes.php:19
msgid "Personal"
msgstr ""

#: Model/Constants/AwardTypes.php:20
#: Template/Element/Competitions/form.ctp:50;52
#: Template/Element/Profiles/panel.ctp:92
msgid "Project"
msgid_plural "Projects"
msgstr[0] "Projects"
msgstr[1] "Project"
```

markstory:
I'm pretty sure this was fixed in #10911. What is different about this scenario?

burzum:
I'm at **3.4.13** and it is not working properly in this case. Like I said, I suspect that it is related to the caching or something a collision between the singular and plural or something like that maybe inside the cached translations? I'm not sure, the code around that is not really the most nice to read and understand part of the core. I haven't had more time for this yet.

markstory:
Ok. You're also in a domain file which might complicate the situation. What makes you think it is caching? Have you tried disabling caching?

burzum:
[...]

PR #11127, opened 2017-08-31
Fix plural translations falling back when they shouldn't

markstory:
When a messageid is defined as a plural key, and *then* a singular one in the same context, the singular only message id should be ignored as message selection methods will choose the first value from the plural values. The plural translations obviously cannot read multiple values from the singular data, causing missed messages.

I've targetted `master` with this change as I think it is a reasonably small enough change that should fix more applications than it could break.

Refs #11089


dereuromark:
I restarted the one failing travis build, looks all good here.

PR #11130, opened 2017-08-31
Update docblocks.

ADmad:


src/View/View.php
```diff
@@ -497,7 +497,7 @@ public function autoLayout($autoLayout = null)
     /**
      * Get the current view theme.
      *
-     * @return string
+     * @return string|null
```
dereuromark: Can the setters then also accept null? E.g. `setTheme($theme)`

ADmad: Yes it can.

dereuromark: We should also point that out. One more advantage of separating the concerns in separate methods :)

ADmad: I updated docblock of `setTheme()` too.

ADmad:
I also noticed the `set*` methods of `View` class don't return `$this` like other setters throughout the framework. Should I update those return types?

markstory:
@ADmad Yes.

PR #11133, opened 2017-08-31
Fix #11132 - trim warning in Cake\View\Helper::addClass

### LINKED ISSUE #11132: trim warning in Cake\View\Helper::addClass when $options['class'] is an array
This is a (multiple allowed):

* [x] bug
* [ ] enhancement

* CakePHP Version: 3.4.13.
* Platform and Target: nginx/1.13.4 , MariaDB/10.1.26, php7.0-fpm/7.0.19-1
### What you did
Passing an array of class attributes to a Form field results in a warning in `\Cake\View\Helper.php:183` when the field has an error state, where `Helper::addClass(array $options = [], $class = null, $key = 'class')` is not designed to handle an array of class attributes in the `$option` parameter.

The following results in a trim warning when a modelless form is validated in the controller method.
```
/* template.ctp */
<?= $this->Form->password('new_password', ['class' => ['form-control']]) ?>

/* Controller.php */
$modellessForm->validate($this->request->getData())
```
### What happened
Warning displayed where a parameter passed to trim `trim` was array.

Internally, element gets added to array of classes.
Externally, no trim warning.

huwcbjones:
Fixes #11132 .

Adds support for the `class` element of an `$options` array to also be an array.

### DIFF
--- a/src/View/Helper.php
+++ b/src/View/Helper.php
@@ -197,9 +197,9 @@ class Helper implements EventListenerInterface
      */
     public function addClass(array $options = [], $class = null, $key = 'class')
     {
-        if (is_array($options[$key])) {
+        if (isset($options[$key]) && is_array($options[$key])) {
             $options[$key][] = $class;
-        } else if (isset($options[$key]) && trim($options[$key])) {
+        } elseif (isset($options[$key]) && trim($options[$key])) {
             $options[$key] .= ' ' . $class;
         } else {
             $options[$key] = $class;
--- a/src/View/Helper.php
+++ b/src/View/Helper.php
@@ -197,7 +197,9 @@ class Helper implements EventListenerInterface
      */
     public function addClass(array $options = [], $class = null, $key = 'class')
     {
-        if (isset($options[$key]) && trim($options[$key])) {
+        if (is_array($options[$key])) {
+            $options[$key][] = $class;
+        } else if (isset($options[$key]) && trim($options[$key])) {
             $options[$key] .= ' ' . $class;
         } else {
             $options[$key] = $class;
--- a/tests/TestCase/View/HelperTest.php
+++ b/tests/TestCase/View/HelperTest.php
@@ -188,4 +188,75 @@ class HelperTest extends TestCase
         $result = $Helper->__debugInfo();
         $this->assertEquals($expected, $result);
     }
+
+    /**
+     * Test addClass() with 'class'=>array
+     */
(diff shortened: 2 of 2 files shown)


src/View/Helper.php
```diff
@@ -197,7 +197,9 @@ protected function _confirm($message, $okCode, $cancelCode = '', $options = [])
      */
     public function addClass(array $options = [], $class = null, $key = 'class')
     {
-        if (isset($options[$key]) && trim($options[$key])) {
+        if (is_array($options[$key])) {
```
markstory: I think you need an isset() here as well.

markstory (changes requested):
Having a test case for this case is needed. If you need help with that let me know and I can add a test.

dereuromark:
This looks like sth that might need a test case to prevent regressions in the future.

src/View/Helper.php
```diff
@@ -197,7 +197,9 @@ protected function _confirm($message, $okCode, $cancelCode = '', $options = [])
...
+            $options[$key][] = $class;
+        } else if (isset($options[$key]) && trim($options[$key])) {
```
ADmad: `elseif`

src/View/Helper.php
huwcbjones: Yeah, it was a quick patch - I'll have a look at writing a test.

Thank you for offering help - if I get stuck, I'll ask! :)

src/View/Helper.php
dereuromark: elseif - see cs.

src/View/Helper.php
markstory: Thanks.

tests/TestCase/View/HelperTest.php
```diff
@@ -188,4 +188,75 @@ public function testDebugInfo()
...
         $this->assertEquals($expected, $result);
     }
+
+    /**
+     * Test addClass() with 'class'=>array
+     */
+    public function testAddClassArray()
+    {
+        $Helper = new TestHelper($this->View);
+        $input = array('class' => [
```
ADmad: Use short array syntax everywhere please.

ADmad (changes requested):

tests/TestCase/View/HelperTest.php
huwcbjones: Fixed - thanks

PR #11134, opened 2017-08-31
Make View's setter methods return $this instead of void.

ADmad:


markstory (commented):
Given that we're changing the public API we should probably have tests.

PR #11139, opened 2017-09-02
Add more context on formhelper templates.

markstory:
Add more context/usage for each template string that FormHelper uses. I chose to update the API docs that the book links to as I didn't want to have to duplicate the default template strings into the book as well.

Refs cakephp/docs#5232


lorenzo:
Nice!

PR #11141, opened 2017-09-02
Fix #11140 - Added support for `fullBase` option in `::link()`.

### LINKED ISSUE #11140: HtmlHelper::link fullBase option
This is a (multiple allowed):

* [x] bug
* [x] enhancement

* CakePHP Version: CakePHP Version: 3.4.13.
* Platform and Target: nginx/1.13.4 , MariaDB/10.1.26, php7.0-fpm/7.0.19-1

### What you did
Used `$options['fullBase' => true]` in `HtmlHelper::link()`.
Output linkdid not contain full base.
Method options are inconsistent as `fullBase` is accepted as an option in the rest of `HtmlHelper`/`UrlHelper` options.

### What happened
Output link did not contain full base.

Output link contains full base.

huwcbjones:
Fixes #11140 
Includes test case for `fullBase`.

### DIFF
--- a/src/View/Helper/HtmlHelper.php
+++ b/src/View/Helper/HtmlHelper.php
@@ -346,7 +346,8 @@ class HtmlHelper extends Helper
     {
         $escapeTitle = true;
         if ($url !== null) {
-            $url = $this->Url->build($url);
+            $url = $this->Url->build($url, $options);
+            $options = array_diff_key($options, ['fullBase' => null]);
         } else {
             $url = $this->Url->build($title);
             $title = htmlspecialchars_decode($url, ENT_QUOTES);
--- a/src/View/Helper/HtmlHelper.php
+++ b/src/View/Helper/HtmlHelper.php
@@ -347,7 +347,7 @@ class HtmlHelper extends Helper
         $escapeTitle = true;
         if ($url !== null) {
             $url = $this->Url->build($url, $options);
-            $options = array_diff_key($options, ['fullBase' => null]);
+            unset($options['fullBase']);
         } else {
             $url = $this->Url->build($title);
             $title = htmlspecialchars_decode($url, ENT_QUOTES);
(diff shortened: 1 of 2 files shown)


src/View/Helper/HtmlHelper.php
```diff
@@ -346,7 +346,8 @@ public function link($title, $url = null, array $options = [])
     {
         $escapeTitle = true;
         if ($url !== null) {
-            $url = $this->Url->build($url);
+            $url = $this->Url->build($url, $options);
+            $options = array_diff_key($options, ['fullBase' => null]);
```
markstory: Could this be `unset($options['fullBase'])` as well?

huwcbjones: Yes, it could be - I looked at what was implemented in `::image()` and copied that.

huwcbjones: After quickly benchmarking it `unset` is about 80% quicker than `array_diff` in this case - didn't see that one coming :smirk:
Will change accordingly!

src/View/Helper/HtmlHelper.php
```diff
@@ -346,7 +346,8 @@ public function link($title, $url = null, array $options = [])
     {
         $escapeTitle = true;
         if ($url !== null) {
-            $url = $this->Url->build($url);
+            $url = $this->Url->build($url, $options);
```
markstory: We should probably whitelist options going into the `build()` right now all the attributes for the link go in which will make it hard for us to add options to build() in the future.

huwcbjones: Is that something for the 4.x.x release, or should we start doing it now?
And if now, what options are going to be whitelisted?

markstory: I don't know when options will be added, if ever. As for a whitelist we could start with `fullBase`.

markstory: I changed my mind. The options for UrlHelper::build() have not changed in a very long time. We can deal with the changes later one.

PR #11149, opened 2017-09-04
Logging email fails array to string conversion

AquilaeNox:
Array to string conversion fails at class Email.php line 2087.
The _logDelivery function under Email.php is trying to convert an array to a string and it's failing as expected.

### DIFF
--- a/src/Mailer/Email.php
+++ b/src/Mailer/Email.php
@@ -2061,7 +2061,6 @@ class Email implements JsonSerializable, Serializable
         return $contents;
     }
 
-    
     /**
      * Log the email message delivery.
      *
--- a/src/Mailer/Email.php
+++ b/src/Mailer/Email.php
@@ -2101,6 +2101,7 @@ class Email implements JsonSerializable, Serializable
         return is_array($key) ? implode(';', $key) : (string) $key;
     }
 
+    
     /**
      * Static method to fast create an instance of \Cake\Mailer\Email
      *
--- a/src/Mailer/Email.php
+++ b/src/Mailer/Email.php
@@ -2060,7 +2060,8 @@ class Email implements JsonSerializable, Serializable
 
         return $contents;
     }
-
+    
+    
     /**
      * Log the email message delivery.
      *
@@ -2084,9 +2085,19 @@ class Email implements JsonSerializable, Serializable
         }
         Log::write(
             $config['level'],
-            PHP_EOL . $contents['headers'] . PHP_EOL . PHP_EOL . $contents['message'],
+            PHP_EOL . $this->_isKeyArray($contents['headers']) . PHP_EOL . PHP_EOL . $this->_isKeyArray($contents['message']),
             $config['scope']
         );
+    }
+    
+    
+    /**
+     * @param null $key
+     * @return null|string
+     */
+    private function _isKeyArray($key = null)
+    {
+        return is_array($key) ? implode(';', $key) : $key;
     }
 
     /**
--- a/src/Mailer/Email.php
+++ b/src/Mailer/Email.php
@@ -2089,7 +2089,8 @@ class Email implements JsonSerializable, Serializable
             $config['scope']
         );
     }
-
+    
+    
     /**
      * Converts an array to string
      * @param null $key
--- a/src/Mailer/Email.php
+++ b/src/Mailer/Email.php
@@ -2092,13 +2092,13 @@ class Email implements JsonSerializable, Serializable
     
     
     /**
-     * Converts $param to string if array given
+     * Converts $array to string if array given
      * @param null $key
      * @return string
      */
-    protected function flatten($param = null)
+    protected function flatten($array = null)
     {
-        return is_array($param) ? implode(';', $param) : (string) $param;
+        return is_array($array) ? implode(';', $array) : (string)$array;
     }
 
     
--- a/src/Mailer/Email.php
+++ b/src/Mailer/Email.php
@@ -2092,12 +2092,12 @@ class Email implements JsonSerializable, Serializable
     /**
      * Converts given value to string
      *
-     * @param string|array $value
+     * @param string|array $value The value to convert
      * @return string
      */
     protected function flatten($value)
     {
-        return is_array($value) ? implode(';', $value) : $value;
+        return is_array($value) ? implode(';', $value) : (string)$value;
     }
 
     /**
--- a/src/Mailer/Email.php
+++ b/src/Mailer/Email.php
@@ -2092,11 +2092,11 @@ class Email implements JsonSerializable, Serializable
     
     
     /**
-     * Converts $array to string if array given
-     * @param null $key
+     * Converts array to string
+     * @param array $array
      * @return string
      */
-    protected function flatten($array = null)
+    protected function flatten($array = [])
     {
         return is_array($array) ? implode(';', $array) : (string)$array;
     }
--- a/src/Mailer/Email.php
+++ b/src/Mailer/Email.php
@@ -2089,19 +2089,18 @@ class Email implements JsonSerializable, Serializable
             $config['scope']
         );
     }
-    
-    
+
     /**
      * Converts given value to string
-     * @param $value
+     *
+     * @param string|array $value
      * @return string
      */
     protected function flatten($value)
     {
-        return is_array($value) ? implode(';', $value) : (string)$value;
+        return is_array($value) ? implode(';', $value) : $value;
     }
 
-    
     /**
      * Static method to fast create an instance of \Cake\Mailer\Email
      *
--- a/src/Mailer/Email.php
+++ b/src/Mailer/Email.php
@@ -2092,13 +2092,13 @@ class Email implements JsonSerializable, Serializable
     
     
     /**
-     * Converts array to string
-     * @param array $array
+     * Converts given value to string
+     * @param $value
      * @return string
      */
-    protected function flatten($array = [])
+    protected function flatten($value)
     {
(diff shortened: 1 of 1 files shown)


src/Mailer/Email.php
```diff
@@ -2084,10 +2085,20 @@ protected function _logDelivery($contents)
...
             $config['scope']
         );
     }
+    
+    
+    /**
+     * @param null $key
+     * @return null|string
+     */
+    private function _isKeyArray($key = null)
```
robertpustulka: `_is*` prefix suggests that the method returns `bool`. 
What about `_flatten` for a  method name?

markstory:
Transports should be returning strings for the header values. Which transport is giving you an array?

src/Mailer/Email.php
```diff
@@ -2060,7 +2060,8 @@ public function send($content = null)
 
         return $contents;
     }
-
+
```
dereuromark: Whitespace issue, needs to be reverted

src/Mailer/Email.php
```diff
@@ -2084,10 +2085,20 @@ protected function _logDelivery($contents)
...
             $config['level'],
-            PHP_EOL . $contents['headers'] . PHP_EOL . PHP_EOL . $contents['message'],
+            PHP_EOL . $this->_isKeyArray($contents['headers']) . PHP_EOL . PHP_EOL . $this->_isKeyArray($contents['message']),
             $config['scope']
         );
     }
+    
+    
+    /**
+     * @param null $key
```
dereuromark: null is not possible :) it is string|null

src/Mailer/Email.php
dereuromark: And never use private, use protected, for these cases.

src/Mailer/Email.php
```diff
@@ -2084,10 +2085,20 @@ protected function _logDelivery($contents)
...
+     * @return null|string
```
dereuromark: If you cast to string in your ternary we can just use string here:

    ... : $key

src/Mailer/Email.php
obinoob: Why not private? I'm not sure if that method will ever be used again :)
And yes it actually suggests a boolean didn't realised that at the time.

src/Mailer/Email.php
obinoob: You mean I could only add a line instead of removing and add two lines?

dereuromark: You also added whitespace characters here. Best to make sure modified lines are trimmed "right".

src/Mailer/Email.php
dereuromark: > I'm not sure if that method will ever be used again :)

Exactly, thus from a a framework perspective allowing.

src/Mailer/Email.php
obinoob: good point it makes sense to ensure that it will return a string

src/Mailer/Email.php
```diff
@@ -2085,19 +2085,19 @@ protected function _logDelivery($contents)
...
+     * Converts an array to string
      * @param null $key
-     * @return null|string
+     * @return string
      */
-    private function _isKeyArray($key = null)
+    protected function flatten($key = null)
     {
-        return is_array($key) ? implode(';', $key) : $key;
+        return is_array($key) ? implode(';', $key) : (string) $key;
```
dereuromark: Casts dont have a whitespace afterwards. See CS report.

src/Mailer/Email.php
dereuromark: Actually, it looks more like array|string

src/Mailer/Email.php
```diff
@@ -2084,11 +2085,23 @@ protected function _logDelivery($contents)
...
         );
     }
+    
+    
+    /**
+     * Converts $array to string if array given
+     * @param null $key
+     * @return string
+     */
+    protected function flatten($array = null)
```
dereuromark: This must not be nullable.
Also, the param annotation is still invalid (null), must be array|string.

src/Mailer/Email.php
```diff
@@ -2092,11 +2092,11 @@ protected function _logDelivery($contents)
...
     
     /**
-     * Converts $array to string if array given
-     * @param null $key
+     * Converts array to string
+     * @param array $array
      * @return string
      */
-    protected function flatten($array = null)
+    protected function flatten($array = [])
```
dereuromark: What I meant is: it must not have any defaults.

src/Mailer/Email.php
```diff
@@ -2092,11 +2092,11 @@ protected function _logDelivery($contents)
     
     
     /**
-     * Converts $array to string if array given
-     * @param null $key
+     * Converts array to string
+     * @param array $array
```
dereuromark: There is still no string annotation. I also wouldnt have used $array as variable name, it is a lie if it can also be a string.
$content is maybe more appropriate.

saeideng:
before sending your change check that by `phpcs `
running .. `vendor/bin/phpcs` in console

src/Mailer/Email.php
obinoob: $content is perfect.
why is phpcs failing is it due to some sort of hidden character, space or break?

src/Mailer/Email.php
obinoob: It would make sense if it was optional i guess... what a learning process!

src/Mailer/Email.php
obinoob: I guess that is failing due to PSR right?

src/Mailer/Email.php
```diff
@@ -2061,6 +2061,7 @@ public function send($content = null)
         return $contents;
     }
 
+
```
dereuromark: I fixed up the code. This you need also to revert, please make sure you do not introduce noise whitespace.

obinoob: Ok thank you for all the input

markstory:
Which transport is giving you arrays?

obinoob:
SMTP is giving me an array

src/Mailer/Email.php
```diff
@@ -2092,12 +2092,12 @@ protected function _logDelivery($contents)
...
      * Converts given value to string
      *
-     * @param string|array $value
+     * @param string|array $value The value to convert
      * @return string
      */
     protected function flatten($value)
     {
-        return is_array($value) ? implode(';', $value) : $value;
+        return is_array($value) ? implode(';', $value) : (string)$value;
```
dereuromark: I removed the cast since according to doc block there is no cast necessary for always string input.

ndm2:
If the headers were an array, wouldn't that already fail in the transport, given that the very same value is being treated as a string even before the Email class will receive anything in return?

src/Mailer/Email.php
obinoob: If value isn't an array might be very well a string... so casting will only add cost I guess ;) I also think that would be nice to have a class of utilities?!? This method could be static... Utilities::flatten($array);

markstory:
Its odd that you're getting an array from SmtpTransport. The code https://github.com/cakephp/cakephp/blob/master/src/Mailer/Transport/SmtpTransport.php#L382-L388 looks like it should be making strings. Instead of adding casting to the logging, would it be better to update the SmtpTransport?

obinoob:
I've no doubts that I am receiving an array, I've debugged/dumped $contents['headers'] and $contents['message'], both are arrays.

I'm using $email->setEmailFormat('html')

I'm terrible sorry but I don't know what's the proper solution! I have no time to study the Mailer API. Thank you all as always...

markstory:
@obinoob Ok. I'll take a look at how the various transports handle their return value.

markstory:
I wasn't able to reproduce the issue you're trying to fix here. I used the stock SmtpTransport hooked up to one of my google accounts, and got headers back as a string. My Email code looked like:

```php
        $email = new Email('default');
        $res = $email->to(['mark@example.com' => 'testing'])
            ->setSubject('oh hi mark')
            ->setEmailFormat('both')
            ->setTemplate('things')
            ->send();
```

My `default` email profile had logging enabled, and I wasn't able to observe the error you're having. How could one reproduce what you're seeing?

obinoob:
I'm not using nothing too fancy here... but eventually I can be doing something wrong?!?

```
$email = new Email('default');
$email->setEmailFormat('html');
$email->setTemplate('invoice');
$email->set(compact('invoice', $this->_invoice));
$email->setTo($data->to);
$email->setCc($data->cc);
$email->setSubject('your invoice');
$email->addAttachments($this->_invoice);
```

markstory:
That looks similar to what I have. Are you sure you're using the stock smtptransport? If you put a debug in it before the send() method returns what do you get?

obinoob:
I'm sure I haven't extended or hacked any class related with the core.
You mean debug($email) ? If so, I get an object filled with a lot of data due to the objects I'm passing to the view ;) what should be expected at this stage?

markstory:
Sorry, I meant at this line. https://github.com/cakephp/cakephp/blob/3f9caaaf1227e00df9921822536b6bba1c5b4722/src/Mailer/Transport/SmtpTransport.php#L175 just before this line `$this_content` should contain 'headers' and 'message' as a string. This https://github.com/cakephp/cakephp/blob/3f9caaaf1227e00df9921822536b6bba1c5b4722/src/Mailer/Transport/SmtpTransport.php#L388 collapses the headers and message into a string (at least it should).

markstory:
While. I am not able to reproduce this, there could be userland transports that do not return string data and these changes help protect against that.

PR #11151, opened 2017-09-04
Add user specifc data to token hash to avoid privilige escalation

ceeram:
Add user specifc data to token hash to avoid privilige escalation

Targeted against token-hmac branch of #11101

### DIFF
--- a/src/View/Helper/SecureFieldTokenTrait.php
+++ b/src/View/Helper/SecureFieldTokenTrait.php
@@ -55,7 +55,8 @@ trait SecureFieldTokenTrait
         $hashParts = [
             $url,
             serialize($fields),
-            $unlocked
+            $unlocked,
+            session_id(),
         ];
         $fields = hash_hmac('sha1', implode('', $hashParts), Security::getSalt());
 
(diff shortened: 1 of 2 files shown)


markstory (approved):
Looks good to me.

src/View/Helper/SecureFieldTokenTrait.php
```diff
@@ -55,7 +55,8 @@ protected function _buildFieldToken($url, $fields, $unlockedFields = [])
         $hashParts = [
             $url,
             serialize($fields),
-            $unlocked
+            $unlocked,
+            session_id(),
```
inoas: Was Security Component requireing native PHP sessions before? If not, could the session id be fetched via https://api.cakephp.org/3.5/class-Cake.Network.Session.html#_id for more flexibility?

rchavik: it's the same thing:

https://api.cakephp.org/3.5/source-class-Cake.Network.Session.html#456

ravage84: I would have preferred a wrapped call, too...

ravage84:
Pardon me ignorance, but why was there no changes to the unit tests?

markstory:
Oh there were after I merged this, as the tests got super sad.

PR #11153, opened 2017-09-05
3.x - Fix error output line highlighting off by one.

ndm2:
The debugger highlights the line before the actual line of error.


markstory:
Thanks!

Issue #11148, opened 2017-09-04
SQL Injection After Order clause


millancore:
* [x] bug

* CakePHP Version: 3.1 to 3.5
* Platform and Target: ORM

### What you did
It is possible to execute SQL sentences after "order" clause
Examples: 

```php
   $query = $this->Assets->find('all',array(
           'contain' => array(
              'Attributes' => array(
                 'sort' => array(
                    'Attributes.name' => "DESC; UPDATE assets SET name = 'inject' WHERE id = 1"
                 )
              )
          ),
          'order' => array(
              'name' => "DESC; UPDATE assets SET name = 'inject' WHERE id = 1"
          )
        ));
```

### What happened
SQL UPDATE executed!

### What you expected to happen
I expected that it doesn't possible!

markstory:
You are inserting snippets of SQL into the QUERY builder, why would you expect different results?

millancore:
I am not sure if this way is correct, I created an API that extends ORM and  I make SQL injection test this happened, at this moment I sanitize all data but I don't know this is a bug or not.

markstory:
You should be using a whitelist for your order fields and directions. We have to support sql snippets in order by clauses to provide support for advanced ordering in postgres and SQLServer. I can update the API docs to make it clearer that order fields/directions are treated as raw data.

millancore:
I'll keep that in mind. Thanks

lorenzo:
We could maybe deprecate that and force people to use `QueryExpression`?

markstory:
@lorenzo It looks like he's using the 2.x compatible `find('type, $options)` call type. I'm not sure we can require query expressions there.

In general I'm :+1: for requiring `QueryExpression` to be the vehicle that raw SQL needs. The extra object prevents accidents as user supplied data will have a very hard time accidentally ending up in an expression object.

inoas:
What about dropping the 2.x compatible way in 4.x then?

burzum:
I agree with @inoas, we should already flag this as deprecated in the next release.

markstory:
@inoas Makes sense to me. Care to put together a pull request for it?

markstory:
Closing as the documentation aspects of this have been fixed.

PR #11155, opened 2017-09-06
3.next Deprecate 2.x-style-finder options

inoas:
Ref: https://github.com/cakephp/cakephp/issues/11148

### DIFF
--- a/src/Datasource/QueryInterface.php
+++ b/src/Datasource/QueryInterface.php
@@ -87,7 +87,7 @@ interface QueryInterface
      * ]);
      * ```
      *
-     * Is equivalent to:
+     * Note: 2.x-style finder are deprecated in 3.6 and will be removed in 4.0. Use the query builder instead:
      *
      * ```
      *  $query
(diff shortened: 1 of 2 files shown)


inoas:
Seems a ton of unit tests are using the old style finder options - right?
https://ci.appveyor.com/project/cakephp/cakephp/build/1.0.8421

lorenzo:
This may affect the Table::get() function and I'm not sure that is intended

saeideng:
I think this is not useful 
because travis says 
```
2.x-style finder options will be removed in 4.0, 
but you supplied following 2.x-style finder options: 
conditions - /home/travis/build/cakephp/cakephp/src/ORM/Table.php, line: 1181
```

`deprecationWarning` says enduser using this line ,that is not useful to member

this method should say for example 
```
2.x-style finder options will be removed in 4.0, 
but you supplied following 2.x-style finder options: 
conditions - App/src/ArticlesController.php, line: 10
``` 

so I think you should deprecated some of option in 
 https://github.com/cakephp/cakephp/blob/d4eadb8a76363cf030196586088c2e285a21570e/src/ORM/Table.php#L1176

inoas:
> This may affect the Table::get() function and I'm not sure that is intended

- Can't we recommend `find()->first()` over `get()`?
- Which keywords of the 2.x-style options would you want to keep usable?

@saeideng I'll take a look!

inoas:
Followup #11157

PR #11157, opened 2017-09-06
Deprecate 2.x-finder option-style via Table::find() deprecationWarning

inoas:
Ref: https://github.com/cakephp/cakephp/issues/11148#issuecomment-327344580

### DIFF
--- a/src/ORM/Table.php
+++ b/src/ORM/Table.php
@@ -1176,16 +1176,16 @@ class Table implements RepositoryInterface, EventListenerInterface, EventDispatc
     public function find($type = 'all', $options = [])
     {
         $deprecatedKeys = [
-            'fields',
-            'conditions',
-            'join',
-            'order',
-            'limit',
-            'offset',
-            'group',
-            'having',
-            'contain',
-            'page',
+            // 'fields', // 24 error warnings
+            // 'conditions', // 77 errors warnings, core seems to rely on thi...
+            'join', // 0 error warnings
+            // 'order', // 19 error warnings
+            // 'limit', // 21 error warnings
+            'offset', // 0 error warnings
+            // 'group', // 1 error warning in cakephp/src/Controller/Component/PaginatorComponent.php, line: 201
+            'having', // 0 error warnings
+            // 'contain', // 83 error warnings, core seems to rely on this...
+            // 'page', // 17 error warnings
         ];
 
         $deprecatedKeyWarns = [];
--- a/src/ORM/Table.php
+++ b/src/ORM/Table.php
@@ -1197,7 +1197,8 @@ class Table implements RepositoryInterface, EventListenerInterface, EventDispatc
         if ($deprecatedKeyWarns !== []) {
             deprecationWarning(
                 '2.x-style finder options will be removed in 4.0, but you supplied following 2.x-style finder options: '
-                    . join(',', $deprecatedKeyWarns), 3);
+                . join(',', $deprecatedKeyWarns)
+            , 3);
         }
 
         $query = $this->query();
--- a/src/ORM/Table.php
+++ b/src/ORM/Table.php
@@ -1175,6 +1175,31 @@ class Table implements RepositoryInterface, EventListenerInterface, EventDispatc
      */
     public function find($type = 'all', $options = [])
     {
+        $deprecated2xKeys = [
+            'fields',
+            'conditions',
+            'join',
+            'order',
+            'limit',
+            'offset',
+            'group',
+            'having',
+            'contain',
+            'page',
+        ];
+
+        $possiblyDeprecated2xKeys = [];
+        foreach ($deprecated2xKeys as $key) {
+            if (array_key_exists($key, $options)) {
+                $possiblyDeprecated2xKeys[] = $option;
+            }
+        }
+        if ($possiblyDeprecated2xKeys !== []) {
+            deprecationWarning(
+                '2.x-style finder options will be removed in 4.0, but you supplied following 2.x-style finder options: '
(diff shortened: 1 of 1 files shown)


saeideng:
this is a bit better than https://github.com/cakephp/cakephp/pull/11155 IMO

ADmad:
As per #11148 the problem is only with "order" key, so why do we need to deprecate other options?

inoas:
Well the idea in https://github.com/cakephp/cakephp/issues/11148#issuecomment-327330109 was to deprecate them entirely and I think we can benefit from having one style to build queries instead of two.

However this /can/ also be done later during 3.6 *or* in 4.0 towards 5.0.

So is the consensus: You want me to just deprecate the order key and try hard so that the unit tests run again?

src/ORM/Table.php
```diff
@@ -1175,6 +1175,32 @@ public function belongsToMany($associated, array $options = [])
...
+            // 'order', // 19 error warnings
+            // 'limit', // 21 error warnings
+            'offset', // 0 error warnings
+            // 'group', // 1 error warning in cakephp/src/Controller/Component/PaginatorComponent.php, line: 201
+            'having', // 0 error warnings
+            // 'contain', // 83 error warnings, core seems to rely on this...
+            // 'page', // 17 error warnings
+        ];
+
+        $deprecatedKeyWarns = [];
```
markstory: Could be 'deprecatedKeys' 🚲 🏠

src/ORM/Table.php
```diff
@@ -1175,6 +1175,32 @@ public function belongsToMany($associated, array $options = [])
...
+            // 'fields', // 24 error warnings
+            // 'conditions', // 77 errors warnings, core seems to rely on thi...
+            'join', // 0 error warnings
+            // 'order', // 19 error warnings
+            // 'limit', // 21 error warnings
+            'offset', // 0 error warnings
+            // 'group', // 1 error warning in cakephp/src/Controller/Component/PaginatorComponent.php, line: 201
+            'having', // 0 error warnings
+            // 'contain', // 83 error warnings, core seems to rely on this...
+            // 'page', // 17 error warnings
```
markstory: If we want to remove the 2.x style find() we should go through the code and replace this usage with query builder calls. Partially deprecating the 2.x style find() doesn't make sense to me.

saeideng: Can we convert array to method ?or  convert 2x style to 3x internally?

saeideng: @markstory 
Cc @lorenzo

src/ORM/Table.php
inoas: Not really as that variable is being used above for the list to compare against ;-)

src/ORM/Table.php
inoas: @markstory while I agree that seems to be a major undertaking. I can go one by one key and see how far I can push it starting with the keys that only throw some error warnings if the consensus/decision is to deprecate the whole 2.x style finders including the way `Table::get()` works. Ref: https://book.cakephp.org/3.0/en/orm/retrieving-data-and-resultsets.html#getting-a-single-entity-by-primary-key

inoas: What about making `get()` chain to the query object similar to `firstOrFail()`?

Current 3.x Style `get()`:
```php
$article = $this->Articles->get($id, [
    'finder' => 'translations',
    'contain' => ['Comments' => 'Users'],
    'cache' => 'custom','key' => 'mykey']);
```

Proposed 4.x style `get()`:
```php
$article = $this->Articles
    ->find('translations')
    ->contain('Comments.Users')
    ->get([$id], ['cache' => 'custom', 'key' => 'mykey']);
```

markstory: That could work. I think get() as a finder method could have strange interactions with finders that set conditions. The `get()` method could convert those specific options into query builder method calls too.

robertpustulka: `get()` could just accept a query builder as an option or a second argument.

```php
$this->Articles->get($id, [
    'foo' => 'bar'
    'query' => function($q) {
        return $q->contain('Comments');
    }
]);
```
or:
```php
$this->Articles->get($id,  function($q) {
    return $q->applyOptions([
        'foo' => 'bar'
    ])->contain('Comments');
});
```

saeideng: I like the 
```
->get($id,function($q){
return $q->order(...)
}
```

saeideng: >If we want to remove the 2.x style find() we should go through the code and replace this usage with query builder calls

I see the core ,internally all of these key will convert to method 
https://github.com/cakephp/cakephp/blob/0e493afa03d656ca7dcbfe63056ce167a5d5fb3d/src/ORM/Query.php#L756

so 
```

Comment (saeideng) [reply src/ORM/Table.php]: if you say replace all usage of array style(within core) to method that is another thing

**Comment by dereuromark** [comment]
Sounds like just going for the actual issue - the order key - might be best to land this and get it merged rather sooner than later.

**Comment by chinpei215** [comment]
To be honest, I am not in favor of deprecating 2.x way. Why not allowing only ASC and DESC when associated array given?
```php
$query->find('all', [
    'order' => [
        'field1' => 'ASC', // OK
        'field2' => 'ASC NULLS FIRST', // Deprecated
        'field3 ASC NULLS FIRST', // OK
    ]
]
```

dereuromark:
Sounds like a reasonable compromise that will allow everything to work as designed.

chinpei215:
OK. Then I will send the alternative pull request tomorrow.

markstory:
Closing as #11585 was merged and there doesn't seem to be strong support to remove the 2.x style finder entirely just yet.

PR #11161, opened 2017-09-06
Fix integration tests when named routes are used

### LINKED ISSUE #11160: Integration test fails with named routes using cakephp/app 3.5
This is a (multiple allowed):

* [x] bug
* [ ] enhancement

* CakePHP Version: 3.5.1
* Platform and Target: Apache 2.4, MySQL 5.7. Integration tests

### What you did
Doing an Integration test in an app (updated to cakephp/app 3.5) that have named routes. The middleware stack was used.

### What happened
The test fails throwing a `DuplicateNamedRouteException` but there aren't duplicate in `_name` option.

Test should be ok.

batopa:
This PR fix #11160 

The issue here is that in `Application` class of cakephp/app 3.5 the `RoutingMiddlawere` is initialized using the `Application` class and `RoutingMiddleware::loadRoutes()` try to load the route rules again also if they are already loaded in other place as in `Router::url()` called in `IntegrationTestCase::_url()`.

I tried to reproduce that case in unit test. I hope it's ok.

### DIFF
--- a/src/Routing/Middleware/RoutingMiddleware.php
+++ b/src/Routing/Middleware/RoutingMiddleware.php
@@ -48,7 +48,7 @@ class RoutingMiddleware
     }
 
     /**
-     * Trigger the application's routes() hook if the application exists.
+     * Trigger the application's routes() hook if the application exists and Router isn't initialized.
      *
      * If the middleware is created without an Application, routes will be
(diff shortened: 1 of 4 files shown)


markstory:
Thanks!

PR #11163, opened 2017-09-07
Support protocol relative URL in HTTP Client.

robertpustulka:
Some servers set `//example.com` as a `Location` header. `buildUrl()` needs to support that in order to build absolute urls.

### DIFF
--- a/src/Http/Client.php
+++ b/src/Http/Client.php
@@ -448,15 +448,20 @@ class Client
             $url .= $q;
             $url .= is_string($query) ? $query : http_build_query($query);
         }
-        if (preg_match('#^https?://#', $url)) {
-            return $url;
-        }
         $defaults = [
             'host' => null,
             'port' => null,
             'scheme' => 'http',
         ];
         $options += $defaults;
+
+        if (preg_match('#^//#', $url)) {
+            $url = $options['scheme'] . ':' . $url;
+        }
+        if (preg_match('#^https?://#', $url)) {
+            return $url;
+        }
+
         $defaultPorts = [
             'http' => 80,
             'https' => 443
--- a/src/Http/Client.php
+++ b/src/Http/Client.php
@@ -402,7 +402,8 @@ class Client
                 $locationUrl = $this->buildUrl($location, [], [
                     'host' => $url->getHost(),
                     'port' => $url->getPort(),
-                    'scheme' => $url->getScheme()
+                    'scheme' => $url->getScheme(),
+                    'protocolRelative' => true
                 ]);
 
                 $request = $request->withUri(new Uri($locationUrl));
@@ -452,10 +453,11 @@ class Client
             'host' => null,
             'port' => null,
             'scheme' => 'http',
+            'protocolRelative' => false
         ];
         $options += $defaults;
 
-        if (preg_match('#^//#', $url)) {
+        if ($options['protocolRelative'] && preg_match('#^//#', $url)) {
             $url = $options['scheme'] . ':' . $url;
         }
         if (preg_match('#^https?://#', $url)) {
--- a/tests/TestCase/Http/ClientTest.php
+++ b/tests/TestCase/Http/ClientTest.php
@@ -139,12 +139,24 @@ class ClientTest extends TestCase
             ],
             [
                 'http://example.com/test.html',
+                '//test.html',
+                [],
+                [
+                    'scheme' => 'http',
+                    'host' => 'example.com',
+                    'protocolRelative' => false
+                ],
+                'url with a double slash',
+            ],
+            [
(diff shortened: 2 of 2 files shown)


dereuromark (approved):

AD7six:
> Some servers set //example.com as a Location header.

Can you explain the actual use case this is for please. 

Do you mean the reason for this change is to follow redirects like so:

```
$ curl -I http://example.com/not/here.html
HTTP/1.1 301 Moved Permanently
Location: //example.com/but/here.html
```

?

Feels like there is significant scope for "upgrading broke my app" with this as `buildUrl('//some/path')` will change behavior.

robertpustulka:
> Do you mean the reason for this change is to follow redirects like so:

Yes.

To me it looks like a bug that those URL weren't supported.
Locations like `//example.com` were redirected to `http://example.com/example.com`.

AD7six:
In the scope of a location header the change makes sense, but it applies much broader than that

I think this is a problem:
```
-> git diff
diff --git tests/TestCase/Http/ClientTest.php tests/TestCase/Http/ClientTest.php
index d8c379e..b9b4d52 100644
--- tests/TestCase/Http/ClientTest.php
+++ tests/TestCase/Http/ClientTest.php
@@ -146,6 +146,15 @@ class ClientTest extends TestCase
                 ],
                 'url without a scheme',
             ],
+            [
+                'http://example.com//some/file',
+                '//some/file',
+                [],
+                [
+                    'scheme' => 'http'
+                ],
+                'Borked but valid path should not be understood as a protocol-relative url',
+            ],
         ];
     }
 
vagrant @ sg7 [ /tmp/cakephp ] (schemeless-url-fix)
-> phpunit tests/TestCase/Http/ClientTest.php
PHPUnit 6.3.0 by Sebastian Bergmann and contributors.

............F..........................                           39 / 39 (100%)

Time: 354 ms, Memory: 12.00MB

There was 1 failure:

1) Cake\Test\TestCase\Http\ClientTest::testBuildUrl with data set #11 ('http://example.com//some/file', '//some/file', array(), array('http'), 'Borked but valid path')
--- Expected
+++ Actual
@@ @@
-'http://example.com//some/file'
+'http://some/file'

/tmp/cakephp/tests/TestCase/Http/ClientTest.php:169
```

I could be wrong, but I feel the right place for fixing that problem is https://github.com/cakephp/cakephp/blob/master/src/Http/Client.php#L401-L402

For comparison, curl doesn't understand protocol-relative urls:

```
$ curl -IL //example.com
curl: (3) <url> malformed
```

unless they are in a location header:
```
$ curl -IL http://cake.dev/foo
HTTP/1.1 302 Moved Temporarily
Location: //example.com

HTTP/1.1 200 OK
...
```

Note also that there is no need to use preg_match to test if a string matches one permutation. `(substr($url, 0, 2) === '//')` would suffice.

robertpustulka:
`//example.com` syntax is used in a scope where a protocol is known, this is a "Protocol relative URL".

Maybe the `buildUrl` method should append the scheme only if a protocol has been explicitly passed as an option?

```php
if (isset($options['scheme'] || (substr($url, 0, 2) === '//')) {
    $url = $options['scheme'] . ':' . $url;
}
...
$options += $defaults;
```

markstory:
@AD7six `//some/file` doesn't seem like valid path to me. It is ambiguous as `//host/path` style URLs are used quite often to load CSS/JS assets, and `//` should likely be interpreted the same in our Client as it would in a browser. 

[rfc 7231](https://tools.ietf.org/html/rfc7231#page-68) states that `Location` should be a [Relative Reference](https://tools.ietf.org/html/rfc3986#section-4.2) which indicates that the first segment after `//` is the authority or domain.

AD7six:
@markstory my concern/point is that a request like this https://cakephp.org//css/cake.css is a valid response on most webservers, and is affected by the modified code. It's easy to write code which would make a request like that:

```
$someVariable = '/css/cake.css';
...
$path = '/';
...
$path .= $someVariable;
$result = (new Client())->get($path, [], ['host' => 'cakephp.org', 'scheme' => 'https']);
```

The behavior of this request is going to significantly change when this PR is merged. Hoping that's understood - I have no problem with the intended scope of the PR (handling protocol relative redirects) as I previously mentioned.

markstory:
That's a good point @AD7six I hadn't considered that. @robertpustulka Is there a way your change could be scoped closer to where Location headers are followed? Perhaps the scheme should only be applied if the scheme option was set as you mentioned earlier.

robertpustulka:
Protocol relative fix is now opt-in. I added a `protocolRelative` option which is set to true in redirects.

markstory:
Nice work!

PR #11164, opened 2017-09-07
Require that cookies added to the Client have a path and a domain.

robertpustulka:
Cookies added to the HTTP client should have a domain and a path set.

`CookieCollection::findMatchingCookies()` will throw a warning when a cookie without a domain is encountered:

`Warning (2): strpos(): Empty needle in [/vendor/cakephp/cakephp/src/Http/Cookie/CookieCollection.php, line 252]`

### DIFF
--- a/src/Http/Client.php
+++ b/src/Http/Client.php
@@ -20,6 +20,7 @@ use Cake\Http\Client\CookieCollection;
 use Cake\Http\Client\Request;
 use Cake\Http\Cookie\CookieInterface;
 use Cake\Utility\Hash;
+use InvalidArgumentException;
 use Zend\Diactoros\Uri;
 
 /**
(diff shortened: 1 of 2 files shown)


src/Http/Client.php
```diff
@@ -192,6 +193,9 @@ public function cookies()
      */
     public function addCookie(CookieInterface $cookie)
     {
+        if (!$cookie->getDomain() || !$cookie->getPath()) {
+            throw new InvalidArgumentException('Cookie must have a domain and a path set.');
+        }
```
markstory: This might caused otherwise valid cookies to be rejected. For example

```
Set-Cookie: foo=bar
```

Is a perfectly legal cookie header, which would now be 'invalid' breaking client requests to servers that make cookies like this. Could we instead fix the needle check?

markstory: Ignore me. When cookie values are parsed we apply the request path/domain if the cookie does not specify them.

robertpustulka: This is a HTTP Client cookies collection. No response cookies involved.

markstory: Good point. I missed that.

PR #11166, opened 2017-09-07
Add propertyExists() and getOrFail() to EntityTrait

markusramsak:
Added two convenience methods which I use in my projects and find very useful.
I hope you find them useful too.

### DIFF
--- a/src/Datasource/EntityTrait.php
+++ b/src/Datasource/EntityTrait.php
@@ -807,13 +807,22 @@ trait EntityTrait
      * Check if property exists in entity even if it is null
      *
      * @param string $property The field to check for existence.
+     * @param bool $alsoCheckForVirtual Check virtual property too?
      * @return bool
      */
-    public function propertyExists($property)
+    public function propertyExists($property, $alsoCheckForVirtual = false)
     {
         $hasProperty = ($this->has($property) || $this->isDirty($property));
-        if (!$hasProperty && array_key_exists($property, $this->_properties)) {
-            return true;
+        if (!$hasProperty) {
+            if (array_key_exists($property, $this->_properties)) {
+                return true;
+            }
+            if ($alsoCheckForVirtual) {
+                $method = static::_accessor($property, 'get');
+                if ($method) {
+                    return true;
+                }
+            }
         }
 
         return $hasProperty;
--- a/src/Datasource/EntityTrait.php
+++ b/src/Datasource/EntityTrait.php
@@ -804,26 +804,23 @@ trait EntityTrait
     }
 
     /**
-     * Check if real or virtual property exists in entity even if it is null
+     * Check if real or virtual property for existence
      *
      * @param string $property The field to check for existence.
      * @return bool
      */
     public function propertyExists($property)
     {
-        $hasProperty = ($this->has($property) || $this->isDirty($property));
-        if (!$hasProperty) {
-            if (array_key_exists($property, $this->_properties)) {
-                return true;
-            }
-
-            $method = static::_accessor($property, 'get');
-            if ($method) {
-                return true;
-            }
+        if (array_key_exists($property, $this->_properties)) {
+            return true;
         }
 
-        return $hasProperty;
+        $method = static::_accessor($property, 'get');
+        if ($method) {
+            return true;
+        }
+
+        return false;
     }
 
     /**
--- a/src/Datasource/EntityTrait.php
+++ b/src/Datasource/EntityTrait.php
@@ -791,12 +791,13 @@ trait EntityTrait
      *
      * @param string $property The field to get value for.
      * @return mixed
+     * @throws \InvalidArgumentException if the property name doesn't exist
      */
     public function getOrFail($property)
     {
         $value = $this->get($property);
         if (!$this->propertyExists($property) && $value === null) {
-            throw new \UnexpectedValueException('Property "' . $property . '" not found.');
+            throw new \InvalidArgumentException('Property "' . $property . '" not found.');
         }
 
         return $value;
--- a/src/Datasource/EntityTrait.php
+++ b/src/Datasource/EntityTrait.php
@@ -796,7 +796,7 @@ trait EntityTrait
     {
         $value = $this->get($property);
         if (!$this->propertyExists($property) && $value === null) {
-            throw new \BadMethodCallException('Property "' . $property . '" not found.');
+            throw new \UnexpectedValueException('Property "' . $property . '" not found.');
         }
 
         return $value;
--- a/src/Datasource/EntityTrait.php
+++ b/src/Datasource/EntityTrait.php
@@ -785,11 +785,11 @@ trait EntityTrait
     {
         return array_keys($this->_dirty);
     }
-    
+
     /**
      * return value if property exists otherwise throw Exception
      *
-     * @param string $property
+     * @param string $property the field to get value for
      * @return mixed
      */
     public function getOrFail($property)
@@ -805,7 +805,7 @@ trait EntityTrait
     /**
      * Check if property exists in entity even if it is null
      *
-     * @param string $property
+     * @param string $property the field to check for existence
      * @return bool
      */
     public function propertyExists($property)
--- a/src/Datasource/EntityTrait.php
+++ b/src/Datasource/EntityTrait.php
@@ -793,7 +793,7 @@ trait EntityTrait
      * @return mixed
      * @throws \InvalidArgumentException if the property name doesn't exist
      */
-    public function getOrFail($property)
+    public function &getOrFail($property)
     {
         $value = $this->get($property);
         if (!$this->propertyExists($property) && $value === null) {
--- a/src/Datasource/EntityTrait.php
+++ b/src/Datasource/EntityTrait.php
@@ -797,7 +797,7 @@ trait EntityTrait
     {
         $value = $this->get($property);
         if ($value === null && !$this->propertyExists($property)) {
-            throw new \InvalidArgumentException('Property "' . $property . '" not found.');
+            throw new InvalidArgumentException('Property "' . $property . '" not found.');
         }
 
         return $value;
--- a/src/Datasource/EntityTrait.php
+++ b/src/Datasource/EntityTrait.php
@@ -764,7 +764,7 @@ trait EntityTrait
     /**
      * Checks if the entity is dirty or if a single property of it is dirty.
      *
-     * @param string $property The field to check the status for.
+     * @param string $property the field to check the status for
      * @return bool Whether the property was changed or not
      */
     public function isDirty($property = null)
--- a/src/Datasource/EntityTrait.php
+++ b/src/Datasource/EntityTrait.php
@@ -804,24 +804,22 @@ trait EntityTrait
     }
 
     /**
-     * Check if property exists in entity even if it is null
+     * Check if real or virtual property exists in entity even if it is null
      *
      * @param string $property The field to check for existence.
-     * @param bool $alsoCheckForVirtual Check virtual property too?
      * @return bool
      */
-    public function propertyExists($property, $alsoCheckForVirtual = false)
+    public function propertyExists($property)
     {
         $hasProperty = ($this->has($property) || $this->isDirty($property));
         if (!$hasProperty) {
             if (array_key_exists($property, $this->_properties)) {
                 return true;
             }
-            if ($alsoCheckForVirtual) {
-                $method = static::_accessor($property, 'get');
-                if ($method) {
-                    return true;
-                }
+
+            $method = static::_accessor($property, 'get');
+            if ($method) {
+                return true;
             }
         }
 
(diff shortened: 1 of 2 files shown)


src/Datasource/EntityTrait.php
```diff
@@ -785,6 +785,38 @@ public function getDirty()
...
+
+    /**
+     * Check if property exists in entity even if it is null
+     *
+     * @param string $property
+     * @return bool
+     */
+    public function propertyExists($property)
+    {
+        $hasProperty = ($this->has($property) OR $this->isDirty($property));
```
davidyell: I'm not sure how I feel about mixing `&&` and `OR` instead of `||`.

src/Datasource/EntityTrait.php
```diff
@@ -785,6 +785,38 @@ public function getDirty()
...
+        if (!$hasProperty && array_key_exists($property, $this->_properties)) {
```
davidyell: Does this work for entity mutator methods?

src/Datasource/EntityTrait.php
dereuromark: The code sniffer is: https://travis-ci.org/cakephp/cakephp/jobs/272836802

src/Datasource/EntityTrait.php
```diff
@@ -787,6 +787,38 @@ public function getDirty()
...
+    /**
+     * Check if property exists in entity even if it is null
+     *
+     * @param string $property the field to check for existence
+     * @return bool
+     */
+    public function propertyExists($property)
+    {
+        $hasProperty = ($this->has($property) || $this->isDirty($property));
+        if (!$hasProperty && array_key_exists($property, $this->_properties)) {
```
inoas: What about virtual properties? `method_exists()`?

src/Datasource/EntityTrait.php
```diff
@@ -785,11 +785,11 @@ public function getDirty()
     {
         return array_keys($this->_dirty);
     }
-    
+
     /**
      * return value if property exists otherwise throw Exception
      *
-     * @param string $property
+     * @param string $property the field to get value for
```
dereuromark: Sentences start with capital letter and end with fullstop.

    The field to get value for.

src/Datasource/EntityTrait.php
mdlouhy: I did that. If the virtual property returns "null" it is considered as not existing. I think that covers all cases. I just thought that a virtual property which returns null makes no sense.

src/Datasource/EntityTrait.php
mdlouhy: I am sorry, I missed the most important part 👍  :-P

src/Datasource/EntityTrait.php
inoas: I think it depends on how you use it and propertyExists does not tell you if it checks for virtual properties or not.

`public function propertyExists($property, $alsoCheckForVirtual = true)`
... would and depending on the core teams preferences the default could also be false.

src/Datasource/EntityTrait.php
```diff
@@ -787,6 +787,38 @@ public function getDirty()
...
+     * return value if property exists otherwise throw Exception
+     *
+     * @param string $property The field to get value for.
+     * @return mixed
+     */
+    public function getOrFail($property)
+    {
+        $value = $this->get($property);
+        if (!$this->propertyExists($property) && $value === null) {
+            throw new \BadMethodCallException('Property "' . $property . '" not found.');
```
inoas: This is not a good exception. BadMethodCallException: `Exception thrown if a callback refers to an undefined method or if some arguments are missing. `
http://php.net/manual/de/class.badmethodcallexception.php

This is what you want - I think: http://php.net/manual/de/class.unexpectedvalueexception.php

src/Datasource/EntityTrait.php
```diff
@@ -787,6 +787,38 @@ public function getDirty()
...
     /**
+     * return value if property exists otherwise throw Exception
+     *
+     * @param string $property The field to get value for.
+     * @return mixed
+     */
+    public function getOrFail($property)
+    {
+        $value = $this->get($property);
+        if (!$this->propertyExists($property) && $value === null) {
```
inoas: As this misses checks for virtual properties this is incomplete. I would expect `getOrFail('full_name')` to work.

inoas (changes requested):

src/Datasource/EntityTrait.php
mdlouhy: you are right ;-)

src/Datasource/EntityTrait.php
mdlouhy: You are right. Feel free to update it ;-)

robertpustulka:
Is the `propertyExists()` part of this PR related to #10709 ?

mdlouhy:
@robertpustulka I think #10709 is little more than this PR but maybe they can be merged ... I wanted new methods because of backwards compatibility.

src/Datasource/EntityTrait.php
```diff
@@ -787,6 +787,43 @@ public function getDirty()
...
+        return $value;
+    }
+
+    /**
+     * Check if real or virtual property for existence
+     *
+     * @param string $property The field to check for existence.
+     * @return bool
+     */
+    public function propertyExists($property)
```
markstory: How is this better than `isset($thing->property)` which will check virtual and concrete properties?

markstory (changes requested):
We require tests for any behavior.

src/Datasource/EntityTrait.php
mdlouhy: isset() returns true if the value is null - but property could exist having the value NULL

mdlouhy:
I am happy if anyone could write tests - I don't know how you use PHPUnit ... I've never done that before :(

robertpustulka:
@mdlouhy please look at how other Entity [tests are implemented](https://github.com/cakephp/cakephp/blob/master/tests/TestCase/ORM/EntityTest.php).

There's also a chapter about testing [in the cookbook](https://book.cakephp.org/3.0/en/development/testing.html).

mdlouhy:
Thank you for your help. I will try to write Unit tests. :)

src/Datasource/EntityTrait.php
mdlouhy: Checking virtual properties is done now.

src/Datasource/EntityTrait.php
```diff
@@ -787,6 +787,43 @@ public function getDirty()
...
+     * return value if property exists otherwise throw Exception
+     *
+     * @param string $property The field to get value for.
+     * @return mixed
+     * @throws \InvalidArgumentException if the property name doesn't exist
+     */
+    public function &getOrFail($property)
+    {
+        $value = $this->get($property);
+        if (!$this->propertyExists($property) && $value === null) {
```
dereuromark: If you invert these two, it would be a bit more performent in same cases. Nano-opt. :)

src/Datasource/EntityTrait.php
robertpustulka: @mdlouhy Are you sure that `isset()` returns true for `null`?
http://php.net/manual/en/function.isset.php

inoas: isset() will return false, if the property returns null (and probably? if the method/virtual property returns null), *but* we need to return true if the property or method (virtual property) exists, don't we?

src/Datasource/EntityTrait.php
```diff
@@ -787,6 +787,43 @@ public function getDirty()
...
+
+    /**
+     * Check if real or virtual property for existence
+     *
+     * @param string $property The field to check for existence.
+     * @return bool
+     */
+    public function propertyExists($property)
+    {
+        if (array_key_exists($property, $this->_properties)) {
```
inoas: Are you certain that `array_key_exists` on the private `_properties` "property" vs `property_exists` is better?
Is the core as well?

src/Datasource/EntityTrait.php
mdlouhy: @robertpustulka your are right, of course - I mistyped ...
I meant: isset() returns FALSE if the property value is null - but I want to check if the property exists, so it should return true when value is set but null.
($entity->property = null; -> property exists)
It should ONLY return false if it really doesn't exist.
The reason is very simple. I have in my use cases some columns which can be null - for example a datetime column "deleted_date". In den beforeSave or afterSave I want to check if the value was changed to or is NULL without checking "isDirty()". In my routine I do or don't do some actions depending on the value available in the entity and is NULL.
I hope this make my intentions a little bit clearer.

src/Datasource/EntityTrait.php
mdlouhy: Actually no I am not sure but I think that this piece of code is very simple to read and therefore okay. I am open to a better solution ;-)

robertpustulka: @inoas how would you check if a key exists in an array using `property_exists`?

inoas: @robertpustulka I am just worried if someone sets up a real property it is not getting returned by this despite accessible via `$entity->myCustomProperty`.

src/Datasource/EntityTrait.php
mdlouhy: Oh I missed this one - usually I take care of the order. Thanks for your feedback ;-)

src/Datasource/EntityTrait.php
```diff
@@ -796,7 +796,7 @@ public function getDirty()
     public function &getOrFail($property)
     {
         $value = $this->get($property);
-        if (!$this->propertyExists($property) && $value === null) {
+        if ($value === null && !$this->propertyExists($property)) {
             throw new \InvalidArgumentException('Property "' . $property . '" not found.');
```
inoas: `\InvalidArgumentException` <= Should IMHO be a use statement at the top then call it without the `\`

dereuromark: Just include my sniffer :) To auto-fix those issues. Problem solved more cleanly.
But sure, one can also manually adjust all those in the code base.

src/Datasource/EntityTrait.php
robertpustulka: @inoas would a real property still be a property from the ORM perspective?

src/Datasource/EntityTrait.php
mdlouhy: You are right. I just saw and changed that. But in comments I should still use the absolute path, right?

dereuromark: Yes, comments are always full FQCN.

dereuromark: You couldnt, but CakePHP community could have :)

src/Datasource/EntityTrait.php
inoas: @robertpustulka should it be? Could it be? It would if both $entity->real_property and $entity->get('real_property') / $entity->set('real_property', $value) would allow access, right?

```php
        logdebug($cmsDocument->foo_bar_quux); // returns value
        logdebug($cmsDocument->get('foo_bar_quux')); // does not return value
```

So right now this is no requirement as it is not transparent to the ORM.

jeremyharris:
@mdlouhy if you struggle trying to write the tests, feel free to ping me :) I'd suggest giving it a try, this is a good first one to learn on!

src/Datasource/EntityTrait.php
```diff
@@ -787,6 +787,43 @@ public function getDirty()
...
+     * @return bool
+     */
+    public function propertyExists($property)
+    {
+        if (array_key_exists($property, $this->_properties)) {
+            return true;
+        }
+
+        $method = static::_accessor($property, 'get');
+        if ($method) {
```
markstory: This could be `return (bool)$method`

mdlouhy: I did it.

src/Datasource/EntityTrait.php
ADmad: > ..without checking "isDirty()..

Why not? Checking whether an entity property has been changed is precisely that method is for.

mdlouhy: @ADmad isDirty feels wrong for me in that use case. I want to check if a property exists and I don't want to care if it was marked as dirty implicitly or explicitly. It feels better for me that way. I think that the code is easier to read that way.

ADmad: Can you show an e.g. of how want you want would be achieved using isDirty() vs the new method you are adding?

lorenzo:
@mdlouhy can you explain what the use case is for this addition?

mdlouhy:
I just want to check a property for existence sometimes. Sometimes I want to recompile or reprocess some texts. I don't want to depend on the isDirty If the the value is null it should be recomplied.

lorenzo:
@mdlouhy under what circumstances is `get()` not good enough for recomputing the values you need? In your case what's the difference between a null value and a property not being set previously?

mdlouhy:
I am sorry for not replying. The more I try to answer your question, the more I think these two methods could be easily replaced by existing core methods. I think isDirty() in combination with get() provides similar functionality. At the moment I don‘t use these two methods anymore cause I‘ve rewritten my code a lot the last weeks. If someone else finds these methods useful then I am okay with adding them to the core but I don‘t think anymore that they are necessary ...

PR #11167, opened 2017-09-07
Fix adding request cookies for urls without a path.

robertpustulka:
Cookies with root `/` path should be also added to urls without a path (like `http://example.com`).


markstory:
Nice find!

PR #11171, opened 2017-09-11
Mute E_DEPRECATED warnings during tests and add depreciation warnings

cleptric:
First batch of added deprecation warnings. How atomic should we do this?
Also, which test should be updated?
For example `EntityContextTest::testValSchemaDefault` calls `$table->schema()->column('title');`.
Should this become `getColumn()`?
We also have a whole test file for `Cake\Database\Schema\Table`

### DIFF
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -11,7 +11,8 @@
     <php>
         <ini name="memory_limit" value="-1"/>
         <ini name="apc.enable_cli" value="1"/>
-        <ini name="error_reporting" value="E_ALL ~ &amp;E_USER_DEPRECATED"/>
+        <!-- E_ALL & ~E_USER_DEPRECATED -->
+        <ini name="error_reporting" value="16383"/>
     </php>
 
     <testsuites>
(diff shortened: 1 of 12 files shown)


phpunit.xml.dist
```diff
@@ -11,6 +11,7 @@
     <php>
         <ini name="memory_limit" value="-1"/>
         <ini name="apc.enable_cli" value="1"/>
+        <ini name="error_reporting" value="E_ALL ~ &amp;E_USER_DEPRECATED"/>
```
markstory: I'm not sure this will work. I think you might need to use `16383` with a comment on where that value came from.

markstory (commented):
Thanks for getting this started!

markstory:
@cleptric I fixed up the constant value and fixed the deprecation warnings I could easily find.

antograssiot:
I also removed deprecations and add a specific error_reporting call has it doesn't seems to works bu using PHPUnit php ini section

cleptric:
Thank you @markstory and @antograssiot 
So I guess the way to go is to update all deprecated method calls to use the new methods.

markstory:
>  So I guess the way to go is to update all deprecated method calls to use the new methods.

I think updating our internal usage of deprecated methods will make the transition to 4.x easier for us in the future.

cleptric:
Besides the phpstan issues (not related to the changes here) I think we good now 🙂

PR #11172, opened 2017-09-12
Make AssociationCollection aware of a table locator.

robertpustulka:
Alternative for #11072

Following https://github.com/cakephp/cakephp/issues/11023#issuecomment-321171559

-----

`TableLocator` instance passes itself to the `AssociationCollection` object. Then the locator is being passed to the created associations. The association creation has been delegated to new `AssociationCollection::load()` method.

This allows to isolate tables created with a particular table locator making independent of a "global" table locator. 

This also makes sure that associations created by tables originating from different locators do not collide making it possible to isolate parts of the system where groups of tables should behave differently in some scenarios, ie share different connections.

For example:

```php
class OrdersTable extends Table
{
    public function initialize()
    {
        $this->hasMany('Articles');
    }
}

class ConnectionAwareLocator extends TableLocator
{
    public function __construct(ConnectionInterface $connection)
    {
        $this->connection = $connection;
    }

    public function get($alias, array $options = [])
    {
        $table = parent::get($alias, $options);
        return $table->setConnection($this->connection);
    }
}

$fooLocator = new ConnectionAwareLocator(ConnectionManager::get('foo'));
$table = $fooLocator->get('Orders');
//this table and it's associations share the "foo" connection


$barLocator = new ConnectionAwareLocator(ConnectionManager::get('bar'));
$table = $barLocator->get('Orders');
//this table and it's associations share the "bar" connection
```

### DIFF
--- a/src/ORM/AssociationCollection.php
+++ b/src/ORM/AssociationCollection.php
@@ -16,6 +16,7 @@ namespace Cake\ORM;
 
 use ArrayIterator;
 use Cake\Datasource\EntityInterface;
+use Cake\ORM\Locator\LocatorAwareTrait;
 use InvalidArgumentException;
 use IteratorAggregate;
 
@@ -29,6 +30,7 @@ class AssociationCollection implements IteratorAggregate
 {
 
     use AssociationsNormalizerTrait;
+    use LocatorAwareTrait;
 
     /**
      * Stored associations
--- a/src/ORM/AssociationCollection.php
+++ b/src/ORM/AssociationCollection.php
@@ -57,6 +57,30 @@ class AssociationCollection implements IteratorAggregate
     }
 
     /**
+     * Creates and adds the Association object to this collection.
+     *
+     * @param string $className The name of association class.
+     * @param string $associated The alias for the target table.
+     * @param array $options List of options to configure the association definition.
+     * @return \Cake\ORM\Association
+     * @throws InvalidArgumentException
+     */
+    public function load($className, $associated, array $options = [])
+    {
+        $options += [
+            'tableLocator' => $this->getTableLocator()
+        ];
+
+        $association = new $className($associated, $options);
+        if (!$association instanceof Association) {
+            $message = sprintf('The association must extend `%s` class, `%s` given.', Association::class, get_class($association));
+            throw new InvalidArgumentException($message);
+        }
+
+        return $this->add($association->getName(), $association);
+    }
+
+    /**
      * Fetch an attached association by name.
      *
      * @param string $alias The association alias to get.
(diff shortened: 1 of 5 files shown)


src/ORM/AssociationCollection.php
```diff
@@ -29,6 +30,7 @@ class AssociationCollection implements IteratorAggregate
 {
 
     use AssociationsNormalizerTrait;
+    use LocatorAwareTrait;
```
lorenzo: I imagined that you would have added the locator as an argument to the constructor and pass it there when building the object.

It is ok to have this trait, but better to have it in the constructor as well.

src/ORM/Locator/TableLocator.php
```diff
@@ -211,6 +212,11 @@ public function get($alias, array $options = [])
             }
             $options['connection'] = ConnectionManager::get($connectionName);
         }
+        if (empty($options['associations'])) {
+            $associations = new AssociationCollection();
```
lorenzo: It would be better to pass the locator as an argument here

src/ORM/Table.php
```diff
@@ -967,9 +967,8 @@ public function addAssociations(array $params)
     public function belongsTo($associated, array $options = [])
     {
         $options += ['sourceTable' => $this];
-        $association = new BelongsTo($associated, $options);
 
-        return $this->_associations->add($association->getName(), $association);
+        return $this->_associations->load(BelongsTo::class, $associated, $options);
```
lorenzo: I like this change!

src/ORM/AssociationCollection.php
```diff
@@ -55,6 +57,30 @@ public function add($alias, Association $association)
...
+     * @return \Cake\ORM\Association
+     * @throws InvalidArgumentException
+     */
+    public function load($className, $associated, array $options = [])
+    {
+        $options += [
+            'tableLocator' => $this->getTableLocator()
+        ];
+
+        $association = new $className($associated, $options);
```
lorenzo: Now that you pass the locator here, it is trivial to let the tables pass the locator when building the `AssociationCollection`

lorenzo (changes requested):

src/ORM/AssociationCollection.php
robertpustulka: Good idea :+1:

robertpustulka:
I added the locator as an optional argument to the constructor.

lorenzo (approved):

lorenzo:
This looks great @robertpustulka

makallio85:
How do we are supposed to handle situation, where you have _some_ associations that should not implement connection what was provided by table locator? I mean I personally have scenario, where I need to distribute main table connection to all associations but 1.

robertpustulka:
@makallio85 You can overwrite the locator passed to the association:

```php
public funciton initialize()
{
    $this->hasMany('Foos'); //this will use the same locator as the parent table
    $this->hasMany('Bars', [ // this will use the provided locator.
        'tableLocator' => $anotherLocator
    ]);
}
```

makallio85:
@robertpustulka Damn right you are. Thanks. This made my day :D

PR #11176, opened 2017-09-13
Better document unsafe inputs in the QueryBuilder.

markstory:
Better document which methods are not safe for untrusted data.

Refs #11148

### DIFF
--- a/src/Database/Query.php
+++ b/src/Database/Query.php
@@ -315,7 +315,7 @@ class Query implements ExpressionInterface, IteratorAggregate
     }
 
     /**
-     * Adds new fields to be returned by a SELECT statement when this query is
+     * Adds new fields to be returned by a `SELECT` statement when this query is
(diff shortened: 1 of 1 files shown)


src/Database/Query.php
```diff
@@ -1076,6 +1082,9 @@ public function order($fields, $overwrite = false)
      * This method allows you to set complex expressions
      * as order conditions unlike order()
      *
+     * Order fields are not suitable for use with user supplied data as they are
```
saeideng: plain text or text in note box(warning,..) for these notes?

lorenzo: Is it possible to make a box here?

markstory: I don't think so unless we start embedding HTML in docblocks.

saeideng: oh, I forgot  this repo is for `core` not `docs` ,i'm sorry :)

PR #11177, opened 2017-09-13
Update the typehint to Connection instead of ConnectionInterface

cleptric:
The methods `cacheMetadata()` and `getSchemaCollection()` are not part of the `ConnectionInterface`

### DIFF
--- a/src/Database/SchemaCache.php
+++ b/src/Database/SchemaCache.php
@@ -15,7 +15,7 @@
 namespace Cake\Database;
 
 use Cake\Cache\Cache;
-use Cake\Datasource\ConnectionInterface;
+use Cake\Database\Connection;
 use Cake\Datasource\ConnectionManager;
 use InvalidArgumentException;
 use RuntimeException;
@@ -93,10 +93,10 @@ class SchemaCache
     /**
      * Helper method to get the schema collection.
      *
-     * @param \Cake\Datasource\ConnectionInterface $connection Connection object
+     * @param \Cake\Database\Connection $connection Connection object
      * @return \Cake\Database\Schema\Collection|\Cake\Database\Schema\CachedCollection
      */
-    public function getSchema(ConnectionInterface $connection)
+    public function getSchema(Connection $connection)
     {
         if (!method_exists($connection, 'schemaCollection')) {
             throw new RuntimeException('The given connection object is not compatible with schema caching, as it does not implement a "schemaCollection()" method.');


ADmad:
Wonder why phpstan didn't report this.

dereuromark:
sad :/
I cant wait for 4.0 to clean this up. Way too much concrete classes used all over the place for method typehinting.

src/Database/SchemaCache.php
```diff
@@ -93,10 +93,10 @@ public function clear($name = null)
...
      * Helper method to get the schema collection.
      *
-     * @param \Cake\Datasource\ConnectionInterface $connection Connection object
+     * @param \Cake\Database\Connection $connection Connection object
      * @return \Cake\Database\Schema\Collection|\Cake\Database\Schema\CachedCollection
      */
-    public function getSchema(ConnectionInterface $connection)
+    public function getSchema(Connection $connection)
     {
         if (!method_exists($connection, 'schemaCollection')) {
```
dereuromark: Can we add an interface here soon, e.g. SchemaCollectionInterface? Then we have automatic "instance of" check here.

dereuromark: Actually, I would then change the typehint to this interface in the next minor and let the Connection object implement this.
That would solve the issue here cleaner IMO. And it should be as BC as this change, as one must use the concrete Connection object so far anyway.

cleptric:
@ADmad It did here: https://github.com/cakephp/cakephp/pull/11171

src/Database/SchemaCache.php
robertpustulka: See this part
https://github.com/cakephp/cakephp/blob/62b363dd70ea6e27ab2b94d91e9e2f914e5551de/src/Database/SchemaCache.php#L99-L110

There's a check for `schemaCollection` method, but the call is on `getSchemaCollection`. Should this be fixed as well? We could also add checks for `cacheMetadata` and leave the interface in a hint. But I'm not sure if PHPStan would catch the *if method exists* check.

dereuromark: Calls like method_exists are not part of constract specification :) Pretty sure hat's why phpstan ignores them.
It probably should be fixed, yeah. We just have to be careful to fix it in a way that it doesnt break too much going to the next minor release.

dereuromark:
@cleptric Will you provide a follow up PR regarding the discussion above?

PR #11179, opened 2017-09-13
Display shell help message with `out` instead of `err` when missing method

littleylv:
When i run `bin/cake orm_cache` (without any methods), it shows error message (red font color):

```
<info>Usage:</info>
cake orm_cache [subcommand] [-c default] [-h] [-q] [-v] [<name>]

<info>Subcommands:</info>

build  Build all metadata caches for the connection. If a table name is
       provided, only that table will be cached.
clear  Clear all metadata caches for the connection. If a table name is
       provided, only that table will be removed.

To see help on a subcommand use <info>`cake orm_cache [subcommand] --help`</info>

<info>Options:</info>

--connection, -c  The connection to build/clear metadata cache data for.
                  <comment>(default: default)</comment>
--help, -h        Display this help.
--quiet, -q       Enable quiet output.
--verbose, -v     Enable verbose output.

<info>Arguments:</info>

name  A specific table you want to clear/refresh cached data for.
      <comment>(optional)</comment>
```

### DIFF
--- a/src/Console/Shell.php
+++ b/src/Console/Shell.php
@@ -508,7 +508,7 @@ class Shell
             return $this->main(...$this->args);
         }
 
-        $this->err($this->OptionParser->help($command));
+        $this->out($this->OptionParser->help($command));
 
         return false;
     }
--- a/src/Console/Shell.php
+++ b/src/Console/Shell.php
@@ -508,6 +508,7 @@ class Shell
             return $this->main(...$this->args);
         }
 
+        $this->err('No subcommand provided. Choose one of the available subcommands.', 2);
         $this->out($this->OptionParser->help($command));
 
         return false;
--- a/tests/TestCase/Console/ShellTest.php
+++ b/tests/TestCase/Console/ShellTest.php
@@ -1049,7 +1049,7 @@ TEXT;
     public function testRunCommandBaseClassMethod()
     {
         $shell = $this->getMockBuilder('Cake\Console\Shell')
-            ->setMethods(['startup', 'getOptionParser', 'err', 'hr'])
+            ->setMethods(['startup', 'getOptionParser', 'out', 'hr'])
             ->disableOriginalConstructor()
             ->getMock();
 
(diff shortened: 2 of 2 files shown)


dereuromark:
I discovered the same the last days.
Might also be intentional, as without arguments the command is no longer valid without an explicit default main command method.
I guess that one always has to run `bin/cake my_command -h` with help to get a help output.
Without it it seems to be considered incomplete now.

dereuromark (approved):

littleylv:
Or should we show some error message before output help: 
```php
$this->err('foo');
$this->out($this->OptionParser->help($command));
```
see: [CommandRunner.php#L166-L169](https://github.com/cakephp/cakephp/blob/master/src/Console/CommandRunner.php#L166-L169)

dereuromark:
Looks like that would make sense. We should at least make sure we keep the exit code of 1 here.

markstory:
Thanks!

thaJeztah:
I don't think this change is correct; when usage (help) output is printed due to missing arguments on Linux, that output should be send to stdErr to prevent the output from being piped to other commands when redirecting output.

I did a quick search to provide more information, and found this; https://www.jstorimer.com/blogs/workingwithcode/7766119-when-to-use-stderr-instead-of-stdout, but there should be other resources explaining this

markstory:
@thaJeztah This change was to restore behaviour that has existed for quite some time. The reasons you cited were part of the reason I has used stderr. But those changes have caused people grief.

thaJeztah:
@markstory thanks for clarifying - my eye fell on this PR's title, and I know there's sometimes confusion around this, so thought to add the information above, but understand the reason to change if it broke people 👍

rchavik:
I too disagree with this change.

I think a better solution is:

```diff
diff --git a/src/Console/Shell.php b/src/Console/Shell.php
index d513038..8bd52aa 100644
--- a/src/Console/Shell.php
+++ b/src/Console/Shell.php
@@ -508,7 +508,7 @@ class Shell
             return $this->main(...$this->args);
         }
 
-        $this->err($this->OptionParser->help($command));
+        $this->_io->err($this->OptionParser->help($command));
 
         return false;
     }
```
This also correctly handles the `<info />` tags as oppose to using the `err()` or `out()` wrapper

dereuromark:
@rchavik You can make a follow up PR :)

littleylv:
@rchavik  👍

Issue #10530, opened 2017-04-16
Possible race-condition while checking `is_dir`


Theaxiom:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: bfa6e091bc5da1c573acf3ead0bb4c25b7298caa

### What happened
https://github.com/cakephp/cakephp/blob/master/src/Cache/Engine/FileEngine.php#L383
```
        if (!is_dir($dir)) {
            mkdir($dir, 0775, true);
        }
```
As you can see, there is a possibility of a race-condition while checking `is_dir`. This nomenclature exists in other places in the core as well.

### What you expected to happen
It would be better to do something like the following:
`if (!@mkdir($dir, 0775, true) && !is_dir($dir)) { throw... }`

markstory:
If you're in a high-concurrency environment you should avoid using FileCache, and if that's not possible you should create the directories ahead of time. Multiple operations on the file system are very hard to make atomic in PHP.

Theaxiom:
I agree with you, I was just going through the code and decided to raise this up. Thanks!

inoas:
Wouldn't that change still help?
What should be used instead?
Should there be a note in the docs?

Theaxiom:
Here's an example of where it is done correctly: https://github.com/cakephp/cakephp/blob/master/src/Shell/Task/AssetsTask.php#L170

markstory:
Pull request up now.

PR #11182, opened 2017-09-14
Use more atomic options when creating directories.

markstory:
Using `is_dir` and then `mkdir` creates an opportunity for a concurrent process to create the directory at the same time resulting in `mkdir()` emitting an error. We can always create the directory and silence the warning to avoid that issue.

Refs #10530


inoas:
This is an interesting practice utilizing '@' to make things atomic. Thanks for doing this and also for making us/me aware.

PR #11190, opened 2017-09-14
Fix: Avoid displaying help messages in red

rchavik:
Fix some issues I found when running core tests in `APP/vendor/cakephp/cakephp`

See #11179

### DIFF
--- a/src/Console/Shell.php
+++ b/src/Console/Shell.php
@@ -509,7 +509,7 @@ class Shell
         }
 
         $this->err('No subcommand provided. Choose one of the available subcommands.', 2);
-        $this->err($this->OptionParser->help($command));
+        $this->_io->err($this->OptionParser->help($command));
 
         return false;
     }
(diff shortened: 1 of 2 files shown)


src/Console/Shell.php
```diff
@@ -508,8 +508,7 @@ public function runCommand($argv, $autoMethod = false, $extra = [])
             return $this->main(...$this->args);
         }
 
-        $this->err('No subcommand provided. Choose one of the available subcommands.', 2);
```
dereuromark: This should be in sync with the new command runner, thats why this line should probably not be removed, right?

rchavik: I dunno. I just reverted the commit, so it wasn't there before 11179

markstory: I think we should keep this as well.

rchavik:
I've made the adjustment to keep the one line.

thaJeztah:
Thanks!

PR #11194, opened 2017-09-15
fixed basic.php miss PHPDocs

lustyrain:
The return value is incorrect


lib/Cake/basics.php
```diff
@@ -196,7 +196,7 @@ function sortByKey(&$array, $sortBy, $order = 'asc', $type = SORT_NUMERIC) {
  *    implement a `__toString` method. Otherwise the class name will be used.
  * @param bool $double Encode existing html entities
  * @param string $charset Character set to use when escaping. Defaults to config value in 'App.encoding' or 'UTF-8'
- * @return string Wrapped text
+ * @return string|array|object Wrapped text, Wrapped Array or Wrapped Object
```
dereuromark: The description could stay as it was.

markstory: The updated text is fine too.

PR #11198, opened 2017-09-15
Enable overriding theme/plugin templates in APP

rchavik:
Example use case:

  You buy a premium theme called Foo and installed in `APP/plugins/Foo`
  and it has the following templates:
    `APP/plugins/Foo/src/Template/Layout/default.ctp`

  You can now override its layout by copying and modifying it in:
    `APP/src/Template/Plugin/Foo/Layout/default.ctp`

### DIFF
--- a/src/View/View.php
+++ b/src/View/View.php
@@ -1459,8 +1459,8 @@ class View implements EventDispatcherInterface
             array_unshift($themePaths, APP . 'Template' . DIRECTORY_SEPARATOR . 'Plugin' . DIRECTORY_SEPARATOR . $this->theme . DIRECTORY_SEPARATOR);
 
             if ($plugin) {
-                for ($i = 0, $count = count($themePaths); $i < $count; $i++) {
-                    array_unshift($themePaths, $themePaths[$i] . 'Plugin' . DIRECTORY_SEPARATOR . $plugin . DIRECTORY_SEPARATOR);
+                foreach (array_reverse($themePaths) as $path) {
+                    array_unshift($themePaths, $path . 'Plugin' . DIRECTORY_SEPARATOR . $plugin . DIRECTORY_SEPARATOR);
                 }
             }
         }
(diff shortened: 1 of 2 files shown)


saeideng:
and when you use 
`APP/src/Template/Plugin/Foo/Layout/default.ctp`
this can be override by (again ) 
`APP/plugins/Foo/src/Template/Plugin/Foo/Layout/default.ctp`

so I'm :-1:

markstory:
Could you not override plugin layouts before?

rchavik:
> Could you not override plugin layouts before?

No. The use case I described didn't work.

I think right now you cannot override a theme's layout and elements from app.

markstory:
You're right that themes take precedence over app templates right now.  Are you concerned about the extra filesystem hits this will incur?  I don't see this being used frequently based on us not having it for so long, so I wonder if this worth the complexity and possible performance costs.

rchavik:
> Are you concerned about the extra filesystem hits this will incur? I don't see this being used frequently based on us not having it for so long, so I wonder if this worth the complexity and possible performance costs.

I can use a modified cake, but would prefer to have this built-in as this feature has been available since 1.x days.

However, f94d8581c8d80e60098d112feee54a030520b1b4 needs to be merged in as it's a bug.

saeideng:
I like to having 
```
controller
$this->viewBuilder()->theme('premium','custom');
//
//usage: 

APP/plugins/premium/src/Template/Layout/default.ctp
//override by
APP/plugins/custom/src/Template/Layout/default.ctp
```

markstory:
@rchavik Would you be able to make a pull request with just that change in it?

rchavik:
I could.

But I thought I could have workaround this by having the theme path configured `config/app.php`, turns out I can't.  Do you have alternative workaround.

I'll prep another PR anyway.

rchavik:
> I like to having 
>
> ```
> //controller
> $this->viewBuilder()->theme('premium','custom'); 
> //usage: 
> APP/plugins/premium/src/Template/Layout/default.ctp
> //override by 
> APP/plugins/custom/src/Template/Layout/default.ctp
> ```

Having two themes active at the same time doesn't make sense to me.

rchavik:
Closing as #11210 is up

PR #11200, opened 2017-09-15
Get default locale before setting a new locale

### LINKED ISSUE #11199: I18n::setLocale doesn't behave the same as in 3.4 and overwrites default locale
This is a (multiple allowed):

* [x] bug (?)

* CakePHP Version: 3.5.2
* Platform and Target: -

### What I did - What happened - And why it happened
I wrote some kind of I18n middleware for one of our projects and the tests I wrote for that middleware failed after I updated from CakePHP 3.4.x to 3.5.x (3.5.2 to be precise, but I don't think it matters). The tests fail because the ``I18n::locale()`` function behaves slightly different now.

The relevant changes can be seen in this commit:
https://github.com/cakephp/cakephp/commit/de23e33c44ca7a2cc3f7636b303f48966be222a9#diff-f93c69a473b7c4e871355ef04e668140L214

In the "old" version (3.4), when calling ``I18n::locale()`` we first get the default locale, before doing anything else, setting the variable ``$_defaultLocale`` to whatever the default was (en_US in my case because I didn't set any other default). Then we can do whatever we want (e.g. set a new locale) but ``$_defaultLocale`` will always be the very first locale (en_US in my case) and ``getDefaultLocale()`` will always return exactly that. That is also what the documentation of ``getDefaultLocale()`` says: "This returns the default locale before any modifications" https://github.com/cakephp/cakephp/blob/master/src/I18n/I18n.php#L332

But in 3.5, we don't call ``getDefaultLocale()`` (or ``defaultLocale()``) anymore when setting a new locale (it doesn't matter if you call ``setLocale()`` directly or if you call ``locale()``), meaning the 'default locale' gets lost. Since ``setLocale()`` overwrites the default locale on "php level" in intl https://github.com/cakephp/cakephp/blob/master/src/I18n/I18n.php#L292 and ``getDefaultLocale()`` reads the default locale from intl https://github.com/cakephp/cakephp/blob/master/src/I18n/I18n.php#L341, when we first use ``getDefaultLocale()``, ``getLocale()`` or ``locale()`` without parameter, the value that, at this point, has last been set will be set as default locale and it will stay that value from then on.

liviakuenzli:
This should fix #11199

### DIFF
--- a/src/I18n/I18n.php
+++ b/src/I18n/I18n.php
@@ -289,6 +289,7 @@ class I18n
      */
     public static function setLocale($locale)
     {
+        static::getDefaultLocale();
         Locale::setDefault($locale);
         if (isset(static::$_collection)) {
             static::translators()->setLocale($locale);
--- a/tests/TestCase/I18n/I18nTest.php
+++ b/tests/TestCase/I18n/I18nTest.php
@@ -19,6 +19,7 @@ use Cake\Cache\Cache;
 use Cake\Core\Plugin;
 use Cake\I18n\I18n;
 use Cake\TestSuite\TestCase;
+use Locale;
 
 /**
  * I18nTest class
@@ -41,7 +42,7 @@ class I18nTest extends TestCase
     public function setUp()
     {
         parent::setUp();
-        $this->locale = I18n::getLocale();
+        $this->locale = Locale::getDefault() ?: I18n::DEFAULT_LOCALE;
     }
 
     /**
@@ -57,6 +58,18 @@ class I18nTest extends TestCase
         I18n::setLocale($this->locale);
         Plugin::unload();
         Cache::clear(false, '_cake_core_');
+    }
+    
+    /**
+     * Tests that the default locale is set correctly
+     *
+     * @return void
+     */
+    public function testDefaultLocale() {
+        $newLocale = 'de_DE';
+        I18n::setLocale($newLocale);
+        $this->assertEquals($newLocale, I18n::getLocale());
+        $this->assertEquals($this->locale, I18n::getDefaultLocale());
     }
 
     /**


dereuromark:
Sounds reasonable.

markstory:
Is there anyway we could test for this?

liviascapin:
I'm not sure how, since we call ``I18n::getLocale()`` in the setup of the test: https://github.com/cakephp/cakephp/blob/master/tests/TestCase/I18n/I18nTest.php#L44 which sets the default class variable before we can actually test if ``setLocale()`` overwrites it. So we can not test this correctly with the current setup and I'm unsure how to proceed.

But the general idea would be to read the 'default' without setting the class variable, then setting another locale (using ``setLocale()``) and  then asserting that ``getLocale()`` returns the set locale, while ``getDefaultLocale`` returns the default we extracted earlier.

I imagine the test to go something like this:
```php
public function testDefaultLocale() {
    $default = Locale::getDefault() ?: I18n::DEFAULT_LOCALE;
    $newLocale = 'de_DE';
    I18n::setLocale($newLocale);
    $this->assertEquals($newLocale, I18n::getLocale());
    $this->assertEquals($default, I18n::getDefaultLocale());
}
```

What do you think? And how can we integrate that in the current test setup?

markstory:
@liviascapin I think that test looks good. We could also read the class properties with phpunit as well.

liviascapin:
Thanks! :)
But the test won't work properly as long as we call ``I18n::getLocale()`` in the ``setUp()`` method (https://github.com/cakephp/cakephp/blob/master/tests/TestCase/I18n/I18nTest.php#L44). It will probably work fine, but it will not be able to determine if setLocale() overwrites the default or not (and that is the point of all of this). What can we do about that?

markstory:
You should be able to use PHPUnit or Reflection APIs to read the value of the protected property.

liviascapin:
Ah I think I get it now :)

I tried to write the test, but I still have a problem with it and I feel kind of stuck. I tried to ensure that the test fails in 3.5.2 and doesn't fail with this fix. However, if I run all the tests, this test passes in both cases. If I only run this specific test, it fails as expected (in 3.5.2). I guess it is because other tests (such as the FloatTypeTest or the NumberTest) use ``getLocale()`` too...

markstory:
Yeah. I think we're going to have a really hard time testing this. I'll merge as the tests are still useful to have.

PR #11203, opened 2017-09-16
Display sorted command list

rchavik:
Better UX

### DIFF
--- a/src/Shell/HelpShell.php
+++ b/src/Shell/HelpShell.php
@@ -73,15 +73,19 @@ class HelpShell extends Shell implements CommandCollectionAwareInterface
         }
 
         if (!$this->commands) {
-            $this->commands = $this->getCommands();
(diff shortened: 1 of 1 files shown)


ADmad:
phpstan is unhappy https://travis-ci.org/cakephp/cakephp/jobs/276148218#L836

rchavik:
What is it talking about?

ADmad:
Problem is due to line 76 `$this->commands = $this->getCommands();`.  `getCommands()` returns an array and you can use `getIterator()` on an array. So the code needs to account for the fact that `$this->commands` can be an array or `\Cake\Console\CommandCollection` instance. Also the typehint of `$commands` property needs to be updated.

ADmad:
Also there is currently no test coverage for line `76` or the `getCommands()` method.

rchavik:
how about now, phpstan seems happy.

Issue #11208, opened 2017-09-18
Issues with the new Cookie management


stefanozoffoli:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.5.2
* Platform and Target: Apache/2.4.12 (Win32) OpenSSL/1.0.1l PHP/5.6.8

### What you did
I tried to port my previous "remember me" cookie implementation (based on Cookie Component) to the new Cookie management.

My codebase is pretty simple:
```php
$this->response = $this->response->withCookie(
    (new Cookie('urmc'))
        ->withValue([
            'user_id' => $user_id,
            'token' => $token,
        ])
        ->withExpiry(new DateTime('+1 month'))
        ->withHttpOnly(true)
);

return $this->redirect('/');
```

I had 3 problems:

### Cookie with redirect
If I set the cookie before a redirect, it doesn't set any cookie.
I tried to debug `ResponseEmitter.php`:  in the function `emitHeaders()` it seems to resets the value of the variable `$cookies` right after setting a header of type `'Location: ...'.`
For debug purpose I tried to temporarily remove the redirect like this:
```php
foreach ($response->getHeaders() as $name => $values) {
    if (strtolower($name) === 'set-cookie') {
        $cookies = array_merge($cookies, $values);
        continue;
    }
    if (strtolower($name) === 'location') {
        continue;
    }
    ...
```
and then `$cookies` correctly had the cookie i setted.

### Cookie with value of type array
After I removed the redirect, it gave me another error:
```
setcookie() expects parameter 2 to be string, array given
```
This is because it calls `Response->convertCookieToArray()` that doesn't convert values of type `array` to `string`.
But I saw that `Cookie()` object can handle values of type `array` by using an internal variable `$isExpanded`... is this correct?
I expected that I could set a cookie with value of type `array` without the need to explicitly call `json_encode()`

### Encrypted Cookie
I don't know if I'm using it incorrectly, but `EncryptedCookieMiddleware` only decodes my cookie and doesn't encode it.
This is my code (in `routes.php`):
```php
Router::scope('/', function (RouteBuilder $routes) {
    ...

    $routes->registerMiddleware('csrf', new CsrfProtectionMiddleware());
    $routes->registerMiddleware('cookies', new EncryptedCookieMiddleware(['urmc'], Configure::read('Security.cookieKey')));

    $routes->applyMiddleware('csrf', 'cookies');

    ...

[...]

PR #11209, opened 2017-09-19
Make array value cookies able to be sent.

markstory:
Add a new method to get a cookie value as a string. I considered changing how `getValue()` works but thought that had bigger compatibility implications. Instead I've modified the interface/implementation. Normally I would shy away from changing interfaces, but this is a very new class and the likelihood of there be user-space implementations is quite low.

Refs #11208

### DIFF
--- a/src/Http/Cookie/Cookie.php
+++ b/src/Http/Cookie/Cookie.php
@@ -246,6 +246,18 @@ class Cookie implements CookieInterface
     /**
      * {@inheritDoc}
      */
+    public function getStringValue()
+    {
+        if ($this->isExpanded) {
+            return $this->_flatten($this->value);
+        }
+
+        return $this->value;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
     public function withValue($value)
     {
         $new = clone $this;
--- a/src/Http/Cookie/Cookie.php
+++ b/src/Http/Cookie/Cookie.php
@@ -185,10 +185,7 @@ class Cookie implements CookieInterface
     }
 
     /**
-     * Create a cookie with an updated name
-     *
-     * @param string $name Name of the cookie
-     * @return static
+     * {@inheritDoc}
      */
     public function withName($name)
     {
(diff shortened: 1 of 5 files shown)


src/Http/Cookie/CookieInterface.php
```diff
@@ -49,6 +49,15 @@ public function getName();
     public function getValue();
 
     /**
+     * Gets the cookie value as a string.
+     *
+     * This will collapse any complex data in the cookie with json_encode()
+     *
+     * @return string
+     */
+    public function getStringValue();
```
jeremyharris: Any reason not to use `__toString`?

src/Http/Cookie/Cookie.php
```diff
@@ -245,20 +236,27 @@ protected function validateName($name)
...
 
     /**
-     * Gets the cookie value
-     *
-     * @return string|array
+     * {@inheritDoc}
      */
     public function getValue()
     {
         return $this->value;
```
stefanozoffoli: I applied your fix and it correctly save the cookie now.
However when I read the cookie value, it continues to be in json format.

Can be correct to modify here like this?
```php
return $this->_expand($this->value);
```

But also when using `ServerRequest->getCookie($cookie_name);` it returns the cookie value in json format and it doesn't call `Cookie->getValue()`.

stefanozoffoli: By setting
```php
return $this->_expand($this->value);
```
There are warnings in the `EncryptedCookieMiddleware` so it's more complicated than this :sweat_smile:

src/Http/Cookie/CookieInterface.php
markstory: Cookies have more properties than just the value. Wouldn't the string value of a cookie include the Secure, HttpOnly, Path and Domain bits?

src/Http/Cookie/Cookie.php
markstory: The value of the cookie is the JSON data though when you're reading from a request. The cookie middleware adds more behavior to cookies though. I wouldn't recommend people store JSON data in plaintext cookies as its full of 💣

src/Http/Cookie/CookieInterface.php
jeremyharris: Yep, you're absolutely right! Disregard :)

PR #11210, opened 2017-09-19
Fix: Duplicate path in $themePaths

rchavik:
Consider:

```
$themePaths = [
    'Theme1',
    'Theme2',
];

array_unshift($themePaths, $themePaths[$i] . 'Plugin' . DS . $plugin . DS);
```

After first iteration, a new element is inserted as first element
so at 2nd iteration, Theme1 is no longer at key 0, but pushed to key 1,
Hence it will do another one with Theme1

Now we can correctly override plugin layouts and elements from app and themes
template files again.

Fixed with a little help from admad, ceeram and robertpustulka

This is a stripped down version of #11198

### DIFF
--- a/src/View/View.php
+++ b/src/View/View.php
@@ -1458,8 +1458,8 @@ class View implements EventDispatcherInterface
             $themePaths = App::path(static::NAME_TEMPLATE, Inflector::camelize($this->theme));
 
             if ($plugin) {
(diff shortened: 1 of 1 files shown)


markstory:
Any tests that we could add?

dereuromark:
@rchavik ping

PR #11211, opened 2017-09-19
False positive URL validation (Test case only)

real34:
with Kibana urls for instance

Kibana urls such as this one are not valid whereas they seem to be valid regarding the specs (valid in an `input type=url` for instance). The failure seems to be related to the presence of the trailing `?_g=()` part.

The sample url in this failing test can be found from the "Share" tab of a dashboard in the latest Kibana version.

kibana-report-template-dashboard-link

The PR only contains the failing test case for now, I will try to look at the regex and update the PR if nobody beats me to it!

### DIFF
--- a/src/Validation/Validation.php
+++ b/src/Validation/Validation.php
@@ -899,7 +899,7 @@ class Validation
      *   with an optional port number
      * - an optional valid path
      * - an optional query string (get parameters)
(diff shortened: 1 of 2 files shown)


real34:
Thank you for your work on this!

Issue #11137, opened 2017-09-01
RFC - Command Classes (replacing Shell / Tasks)


markstory:
The current Shell/Tasks classes have a number of rough spots which are mostly due to features being incrementally added over the years. A short list of the problems I see are as follows:

* Public properties for `args` and `params`. These simple arrays have been partially wrapped with methods, but they are still a bit clunky to use.
* Subcommands can be tasks or shell methods. Having both options makes it hard to give people clear directions on which approach they should use.
* The API of Shell is confusing. Feature both methods like `out()` `err()` and `abort()` as well as user land code. Shell has too many concerns.

## Focused Commands

If shells were stripped down their goals are to provide a series of 'commands'. Each command is effectively a callable that can be invoked from the CLI. The callable could have the input/output wrappers passed in removing the need for most state in a command.

Sub-commands can be created by mounting commands in a nested fashion. For example `bake test` could be a name that points to `Bake\Console\Console\TestCommand` while `bake controller` could be a name that resolves to `Bake\Console\Console\ControllerCommand`. Each command can focus on only the task they need to accomplish.

## Code Reuse in Commands

If each command is single purpose, we'll need a good way to share logic between commands. We've had good success so far with Traits being utilized to allow access to models, logging and mailers. We can leverage those traits in Commands to cover most of the need for shared logic. In addition to traits, we can improve ShellHelpers to enable more code re-use.

Shared option parser settings is another place where I can see logic needing to be shared. This can be addressed by enabling option parsers to be standalone classes like we did with validators.

## Command Life-Cycle

Commands would have the following life-cycle:

1. `CommandRunner` locates a command based on the CLI arguments.
2. `CommandRunner` constructs the command. The `CommandCollection` would be injected if necessary.
3. The `Command.initialize` event would be triggered. Calling the command's `initialize()` hook if defined.
4. The command's OptionParser is created with `getOptionParser()`
5. CLI options are parsed. An error will be output if required options/arguments are missing.
6. Help might be displayed if requested.
7. The command's `execute()` method would be called receiving the `InputArguments` and `ConsoleIo` objects as parameters.
8. The return value of the command would be converted into an exit code, and passed out to the CLI environment.

## Arguments class

The `Arguments` class provides an interface for interacting with the parameters and passed arguments in a command. It is passed to commands after the CLI arguments have been parsed, and would feature methods like:

```php
// Get all parameters
$input->getParams();

// Get the parameter. No default option as defaults are defined in the option parser.
[...]

PR #11216, opened 2017-09-20
3.next- Start implementing the Arguments class.

markstory:
Build out basic positional argument methods. I've skipped methods to fetch positional arguments by name, as I don't think it makes sense now. The whole point of positional arguments is that they are _positional_.

Refs #11137

### DIFF
--- a/src/Console/Arguments.php
+++ b/src/Console/Arguments.php
@@ -1,0 +1,84 @@
+<?php
+/**
+ * CakePHP(tm) : Rapid Development Framework (https://cakephp.org)
+ * Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
+ *
+ * Licensed under The MIT License
+ * For full copyright and license information, please see the LICENSE.txt
+ * Redistributions of files must retain the above copyright notice.
+ *
+ * @copyright     Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
+ * @link          https://cakephp.org CakePHP(tm) Project
+ * @since         3.6.0
+ * @license       https://opensource.org/licenses/mit-license.php MIT License
+ */
+namespace Cake\Console;
+
+/**
+ * Provides an interface for interacting with
+ * a command's options and arguments.
+ */
+class Arguments
+{
+    /**
+     * Positional arguments.
+     *
+     * @var array
+     */
+    protected $args;
+
+    /**
+     * Named options
+     *
+     * @var array
+     */
+    protected $options;
+
+    /**
+     * Constructor
+     *
+     * @param array $args Positional arguments
+     * @param array $options Named arguments
+     */
+    public function __construct(array $args, array $options)
+    {
+        $this->args = $args;
+        $this->options = $options;
+    }
+
+    /**
+     * Get all positional arguments.
+     *
+     * @return array
+     */
+    public function getArguments()
+    {
+        return $this->args;
+    }
+
+    /**
+     * Get positional arguments by index.
+     *
+     * @param int $index The argument index to access.
+     * @return string|null The argument value or null
+     */
+    public function getArgument($index)
+    {
+        if (isset($this->args[$index])) {
+            return $this->args[$index];
+        }
+
+        return null;
+    }
+
+    /**
+     * Check if a positional argument exists
+     *
+     * @param int $index The argument index to check.
+     * @return bool
+     */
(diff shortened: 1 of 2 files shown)


src/Console/Arguments.php
```diff
@@ -0,0 +1,84 @@
...
+     */
+    protected $options;
+
+    /**
+     * Constructor
+     *
+     * @param array $args Positional arguments
+     * @param array $options Named arguments
+     */
+    public function __construct(array $args, array $options)
```
lorenzo: Should it maybe have the ability to name arguments? like being able to address the 3rd argument by name.
I've found this to be very useful in other frameworks.

src/Console/Arguments.php
```diff
@@ -0,0 +1,84 @@
...
+/**
+ * Provides an interface for interacting with
+ * a command's options and arguments.
+ */
+class Arguments
+{
+    /**
+     * Positional arguments.
+     *
+     * @var array
```
dereuromark: Are those actuall `string[]`?

src/Console/Arguments.php
```diff
@@ -0,0 +1,84 @@
...
+    public function __construct(array $args, array $options)
+    {
+        $this->args = $args;
+        $this->options = $options;
+    }
+
+    /**
+     * Get all positional arguments.
+     *
+     * @return array
```
dereuromark: Same here then.

src/Console/Arguments.php
dereuromark: I would then use a similar approach of naming as I mentioned in the main ticket:

    getArgumentAt(index)
    hasArgumentAt(index)

and the new

    hasArgument(string)

maybe then.

markstory: @lorenzo Are you thinking of a map between names and indexes?

lorenzo: Yes, exactly

markstory: Ok. I was thinking of skipping that. But if you think it will be valuable to have. I'll add it in.

markstory: I've renamed the methods to include 'At'. I'll add the name based methods in a subsequent pull request.

markstory:
@lorenzo @dereuromark I've added named argument support.

tests/TestCase/Console/ArgumentsTest.php
```diff
@@ -57,9 +56,43 @@ public function testGetArgumentAt()
...
+
+    /**
+     * get arguments by name
+     *
+     * @return void
+     */
+    public function testGetArgument()
+    {
+        $values = ['big', 'brown', 'bear'];
+        $names = ['size' => 0, 'color' => 1, 'species' => 2, 'odd' => 3];
```
saeideng: how happen if  `$name` set like bellow 
```php
$names = ['size', 'color', 'species', 'odd'];
//or 
$names = [0 => 'size', 1 => 'color', 2 => 'species',3 =>  'odd'];
```

src/Console/Arguments.php
```diff
@@ -0,0 +1,125 @@
...
+     * @param int $index The argument index to access.
+     * @return string|null The argument value or null
+     */
+    public function getArgumentAt($index)
+    {
+        if ($this->hasArgumentAt($index)) {
+            return $this->args[$index];
+        }
+
+        return null;
```
robertpustulka: Maybe the exception should be thrown if an argument is not present?

src/Console/Arguments.php
```diff
@@ -0,0 +1,125 @@
...
+    /**
+     * Check if a positional argument exists by name
+     *
+     * @param string $name The argument name to check.
+     * @return string|null
+     */
+    public function getArgument($name)
+    {
+        if (!isset($this->argNames[$name])) {
+            return null;
```
robertpustulka: Same here.

src/Console/Arguments.php
markstory: Is an argument missing exceptional? The CLI parser should throw errors when required arguments are missing.

tests/TestCase/Console/ArgumentsTest.php
markstory: @saeideng We'd have to flip that data so we can use `isset()`, or switch to using `array_search()`. Using `array_search()` might be simpler for callers now that you mention it.

PR #11217, opened 2017-09-20
Return empty string when View instance is cast to string.

ADmad:
Refs WyriHaximus/TwigView#60

### DIFF
--- a/src/View/View.php
+++ b/src/View/View.php
@@ -1551,4 +1551,14 @@ class View implements EventDispatcherInterface
 
         return $element;
     }
+
+    /**
+     * Return empty string when View instance is cast to string.
+     *
+     * @return string
+     */
+    public function __toString()
+    {
+        return '';
+    }
 }
--- a/tests/TestCase/View/ViewTest.php
+++ b/tests/TestCase/View/ViewTest.php
@@ -2115,4 +2115,10 @@ TEXT;
         $this->assertEquals('mypath', $View->templatePath());
         $this->assertEquals('mypath', $View->templatePath);
     }
+
+    public function testToString()
+    {
+        $result = "{$this->View}";
+        $this->assertSame('', $result);
+    }
 }


src/View/View.php
```diff
@@ -1551,4 +1551,14 @@ protected function _renderElement($file, $data, $options)
 
         return $element;
     }
+
+    /**
+     * Return empty string when View instance is cast to string.
```
dereuromark: is the participle form "casted"?

ADmad: > The verb cast is conventionally uninflected in the past tense and as a past participle.

http://grammarist.com/usage/casted/

inoas:
This will potentially leave authors of affected plugins in the dark as say `<h1>{{ _view.assign('title', 'What is BC') }}</h1>` will continue to run but leave an empty `<h1/>`. So I'd prefer to:
- either revert the change in CakePHP as Mark Story suggested earlier
- or implement this change (and some more) in plugins like `TwigView` - I can take care of that for TwigView.

dereuromark:
Shouldn't [TwigView](https://github.com/WyriHaximus/TwigView/blob/master/src/View/TwigView.php) implement this fix instead? Then the core doesn't need any "workaround fixes" here for an issue that only really can come up in such a Twig specific "always read return values" scenario.

ADmad:
Strictly speaking changing return type is BC break and since the "fix" is so simple I think we should do it. I don't think this change will affect any normal usage nor say dumping view object through `debug()`.

dereuromark:
Changing void by its very definition of it ("do not ever check and use return type here") cannot be a BC break, even in the strictest terms.
Thus this really should only be done in TwigView which violated this contract - and feels the issue as probably only view class ever out there :)

src/View/View.php
```diff
@@ -1551,4 +1551,14 @@ protected function _renderElement($file, $data, $options)
...
+     *
+     * @return string
+     */
+    public function __toString()
+    {
+        return '';
```
dereuromark: Intuitively, I would expect it to internally generate() here really. Rendering the view as is as a whole. This quick-fix for a plugin that did it wrong in the first place sounds misplaced at this point.

ADmad: You probably mean render(). I agree it does seem more intuitive, that's what we do for cells too.

dereuromark:
With https://github.com/WyriHaximus/TwigView/pull/61 now it seems this PR has become not necessary anymore. I vote for closing.

markstory:
With TwigView updated we could close this.

saeideng:
a question in here ,`view::__toString` just used in this plugin ?

inoas:
@saeideng the problem is IMHO unlikely to exist but could exist in other plugins that assumed to void returns, those plugins should IMHO implement the same fix.

PR #11218, opened 2017-09-20
Update tests and add deprecation warnings for mailing classes

cleptric:
All the stuff around the Mailer looked like a nice package, so I deprecated all together.


markstory:
Awesome work @cleptric

inoas:
Thanks!

PR #11219, opened 2017-09-20
Improve IDE compatibility

dereuromark:
Less yellow in the IDE, more typehinting now included :)

### DIFF
--- a/src/Controller/Component/AuthComponent.php
+++ b/src/Controller/Component/AuthComponent.php
@@ -877,7 +877,7 @@ class AuthComponent extends Component
      *
      * @param \Cake\Auth\Storage\StorageInterface|null $storage Sets provided
      *   object as storage or if null returns configured storage object.
-     * @return \Cake\Auth\Storage\StorageInterface|null
+     * @return \Cake\Auth\Storage\StorageInterface|\Cake\Core\InstanceConfigTrait|null
      */
     public function storage(StorageInterface $storage = null)
     {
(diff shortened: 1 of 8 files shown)


src/Controller/Component/AuthComponent.php
```diff
@@ -877,7 +877,7 @@ public function constructAuthenticate()
      *
      * @param \Cake\Auth\Storage\StorageInterface|null $storage Sets provided
      *   object as storage or if null returns configured storage object.
-     * @return \Cake\Auth\Storage\StorageInterface|null
+     * @return \Cake\Auth\Storage\StorageInterface|\Cake\Core\InstanceConfigTrait|null
```
dereuromark: Unfortunately, this is needed for getConfig() and setConfig calls on this storage() to be typehinted and found as valid.

src/Controller/Component/SecurityComponent.php
```diff
@@ -510,12 +511,12 @@ protected function _debugPostTokenNotMatching(Controller $controller, $hashParts
             'Missing field \'%s\' in POST data'
         );
         $expectedUnlockedFields = Hash::get($expectedParts, 2);
-        $dataUnlockedFields = Hash::get($hashParts, 2) ?: [];
+        $dataUnlockedFields = Hash::get($hashParts, 2) ?: null;
```
markstory: Why are you defaulting to `null` and then casting?

dereuromark: Warning about string Operation on array in next line. Switching types on the same var here.

markstory: Ok. I misread explode as implode derp.

PR #11222, opened 2017-09-21
Bug in CakeEmail matching pattern

IonutBurlacu:


dereuromark:
We should have a test case with such an example domain to prevent regressions.

gemal:
can this be merged?

dereuromark:
I mentioned above that a test case is critical for this to be mergable.

gemal:
ok. sorry. we are doing the test case now

dereuromark (approved):

IonutBurlacu:
Not sure why the tests failed but it seems that it's not caused by my changes.

dereuromark:
Yes, unrelated. I re-triggered it.

thaJeztah:
Are underscores actually allowed in domain names?

dereuromark:
I was reading https://stackoverflow.com/questions/2180465/can-domain-name-subdomains-have-an-underscore-in-it

markstory:
I'll get this merged and ported to 3.x

PR #11225, opened 2017-09-21
Rename AssociationCollection::type() to AssociationCollection::getByType()

### LINKED ISSUE #11220: RFC: Rename AssociationCollection::type() to AssociationCollection::getByType()
The proposed method name is lot for intuitive and better conveys what the method does. It would also be consistent with existing `AssociationCollection::getByProperty()` method.

ADmad:
Closes #11220

### DIFF
--- a/src/ORM/AssociationCollection.php
+++ b/src/ORM/AssociationCollection.php
@@ -114,8 +114,22 @@ class AssociationCollection implements IteratorAggregate
      * @param string|array $class The type of associations you want.
      *   For example 'BelongsTo' or array like ['BelongsTo', 'HasOne']
      * @return array An array of Association objects.
+     * @deprecated 3.6.0 Use getByType() instead.
      */
     public function type($class)
+    {
+        return $this->getByType($class);
+    }
+
+    /**
+     * Get an array of associations matching a specific type.
+     *
+     * @param string|array $class The type of associations you want.
+     *   For example 'BelongsTo' or array like ['BelongsTo', 'HasOne']
+     * @return array An array of Association objects.
+     * @since 3.5.3
+     */
+    public function getByType($class)
     {
         $class = array_map('strtolower', (array)$class);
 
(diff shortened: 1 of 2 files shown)


src/ORM/AssociationCollection.php
```diff
@@ -114,9 +114,23 @@ public function keys()
      * @param string|array $class The type of associations you want.
      *   For example 'BelongsTo' or array like ['BelongsTo', 'HasOne']
      * @return array An array of Association objects.
+     * @deprecated 3.6.0 Use getByType() instead.
```
dereuromark: Isn't that 3.5.3 then as well?

ADmad: The phpdoc [docs](https://docs.phpdoc.org/references/phpdoc/tags/deprecated.html) say:
> This tag MAY also contain a version number up till which it is guaranteed to be included in the software. Starting with the given version, the function will be removed or may be removed without further notice.

The method is gonna exist in 3.6 so wouldn't 3.6.0 be the right version?

dereuromark: It will also exist in any 3.7 and alike. We cannot say anything about the future except for it will stay for this major release.
I think we always put the version number it got deprecated, since that is the most interesting information. We probably need to know at what point to look for in the migration guides and the rest of the code to use instead of it.

src/ORM/AssociationCollection.php
```diff
@@ -114,9 +114,23 @@ public function keys()
...
+
+    /**
+     * Get an array of associations matching a specific type.
+     *
+     * @param string|array $class The type of associations you want.
+     *   For example 'BelongsTo' or array like ['BelongsTo', 'HasOne']
+     * @return array An array of Association objects.
+     * @since 3.5.3
+     */
+    public function getByType($class)
```
markstory: We should update tests to use the new method to save future us work.

PR #11227, opened 2017-09-23
3.next Arguments option methods.

markstory:
Add methods for handling options.

Refs #11137


dereuromark:
Do we support arrays for arguments or options? if so, how would that work?

lorenzo:
Looks good to me :)

markstory:
@dereuromark We do have array values from options as there are options that can be defined multiple times.

PR #11228, opened 2017-09-23
Fix up doc blocks

dereuromark:
Small mixup of array annotations fixed up.

Follows https://github.com/cakephp/cakephp/pull/11216


markstory:
Thanks!

PR #11231, opened 2017-09-24
update to use deprecationWarning method+tests

saeideng:


tests/TestCase/Controller/ControllerTest.php
```diff
@@ -258,6 +264,27 @@ public function tearDown()
...
+        error_reporting(self::$error_level);
+    }
+
+    /**
+     * setUpBeforeClass
+     *
+     * @return void
+     */
+    public static function setUpBeforeClass()
+    {
```
saeideng: I'm not sure needs `parent::setUpBeforeClass();` here or no
because we no never use this in cakephp core

tests/TestCase/Controller/ControllerTest.php
```diff
@@ -237,6 +237,12 @@ class ControllerTest extends TestCase
     ];
 
     /**
+     * error level property
+     *
+     */
+    private static $error_level;
```
markstory: This should be `errorLevel`.

tests/TestCase/Controller/ControllerTest.php
```diff
@@ -1097,24 +1126,50 @@ public function testBeforeRenderViewVariables()
...
-            $this->assertSame(sprintf('Controller::$%s is deprecated. Use $this->viewBuilder()->%s() instead.', $property, $getter), $message);
-
             $this->assertSame($value, $controller->viewBuilder()->{$getter}());
-        } finally {
-            restore_error_handler();
-        }
+    }
+
+    /**
+     * Tests deprecated view propertiyes message
```
markstory: Typo in properties.

tests/TestCase/Controller/ControllerTest.php
```diff
@@ -1097,24 +1126,50 @@ public function testBeforeRenderViewVariables()
...
+     * @param $property Deprecated property name
+     * @param $getter Getter name
+     * @param $setter Setter name
+     * @param mixed $value Value to be set
+     * @return void
+     * @expectedException PHPUnit\Framework\Error\Deprecated
+     * @expectedExceptionMessageRegExp /^Controller::\$\w+ is deprecated(.*)/
+     * @dataProvider deprecatedViewPropertyProvider
+     */
+    public function testDeprecatedViewPropertySetterMessage($property, $getter, $setter, $value)
```
markstory: You don't use getter ever. Perhaps it should be removed?

saeideng: Yes,but why this? because I use exist/and same data provider for 3 test

saeideng: by using same data(provider) we can ensure all test is ok when we want add `viewVarsTrait::viewClass` to this data provider future

saeideng: I see again 
there are 3 test with bellow usage
```php
1) ($property, $getter, null, $value)
2) ($property, null, null, $value)
3) ($property, null, $setter, $value)
```
we can remove some of args just by changing order of args

markstory: Probably not worth the effort.

markstory:
Thanks!

Issue #11230, opened 2017-09-24
Using find with ->page(0) results in inadequate error message


jeffreyroberts:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.4.13
* Platform and Target: OS X local setup

### What you did
![screen shot 2017-09-24 at 9 50 47 am](https://user-images.githubusercontent.com/2320614/30783291-625a3980-a10e-11e7-9e80-9610ba274654.png)

### What happened
     Exception: SQLSTATE[42000]: Syntax error or access violation: 1064 You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near '-1000' at line 1 in [/Users/jlroberts/Projects/JeffreyLRobertsCom/Dashbored/vendor/cakephp/cakephp/src/Database/Statement/MysqlStatement.php, line 39]

and if I put the ->page before the ->limit I get another error message

    Exception: SQLSTATE[42000]: Syntax error or access violation: 1064 You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near '-25' at line 1 in [/Users/jlroberts/Projects/JeffreyLRobertsCom/Dashbored/vendor/cakephp/cakephp/src/Database/Statement/MysqlStatement.php, line 39]

### What you expected to happen

I expected the first page of results to be returned, thinking it was a zero based index.

saeideng:
Pages should start at 1.
https://github.com/cakephp/cakephp/blob/8dea5d8c6a5cb01e9e0e52948c8700f55a7c4d70/src/Database/Query.php#L1256

markstory:
We can raise an error for pages < 1.

markstory:
Pull request merged now (#11233)

PR #11233, opened 2017-09-25
raise an error for pages < 1. #11230

saeideng:


tests/TestCase/Database/QueryTest.php
```diff
@@ -2207,6 +2207,21 @@ public function testSelectOffset()
...
+     * @expectedException \InvalidArgumentException
+     * @expectedExceptionMessage Pages should start at 1.
+     * @return void
+     */
+    public function testPageShouldStartAtOne()
+    {
+        $this->loadFixtures('Comments');
+        $query = new Query($this->connection);
+        $result = $query->from('comments')
+                        ->page(0);
```
dereuromark: Please no inline indentation.

src/Database/Query.php
```diff
@@ -1259,9 +1259,14 @@ public function orHaving($conditions, $types = [])
      * @param int|null $limit The number of rows you want in the page. If null
      *  the current limit clause will be used.
      * @return $this
+     * @throws \InvalidArgumentException If page number < 1.
      */
     public function page($num, $limit = null)
     {
+        if ($num < 1) {
+            $msg = 'Pages should start at 1.';
```
ADmad: Why use a variable instead of directly using the string on next line?

tests/TestCase/Database/QueryTest.php
saeideng: phpcs is happy

src/Database/Query.php
saeideng: if important I can change it

tests/TestCase/Database/QueryTest.php
dereuromark: "happy" doesnt mean it finds all issues :) Some issues are no properly written sniffs for yet.

src/Database/Query.php
dereuromark: The extra line is just pointless here.

saeideng:
fixed

src/Database/Query.php
```diff
@@ -1259,9 +1259,13 @@ public function orHaving($conditions, $types = [])
      * @param int|null $limit The number of rows you want in the page. If null
      *  the current limit clause will be used.
      * @return $this
+     * @throws \InvalidArgumentException If page number < 1.
      */
     public function page($num, $limit = null)
     {
+        if ($num < 1) {
+            throw new InvalidArgumentException('Pages should start at 1.');
```
dereuromark: should or must?

saeideng: I copy this from docs

saeideng: also maybe we want to change 
`$offset = ($num - 1) * $limit;` to `$offset = ($num ) * $limit;` in future

dereuromark: What I am getting at: With your added exception it sounds more logical to make this a "must" here, as should implies it work also work with 0 etc.

ADmad (approved):

dereuromark (approved):

markstory:
Thanks!

PR #11234, opened 2017-09-25
Fix #11229 : Change Email Pattern to match RFC3696 and RFC5321

### LINKED ISSUE #11229: Email pattern does not match RFC3696
This is a (multiple allowed):

* [x] bug
* [ ] enhancement

* CakePHP Version: 3.5.1

Constante [EMAIL_PATTERN](https://github.com/cakephp/cakephp/blob/master/src/Mailer/Email.php#L90) does not match **double-quote examples** given in [RFC3696](https://tools.ietf.org/html/rfc3696) and **Literal IPV6** in domain name [RFC2732](https://tools.ietf.org/html/rfc2732)

RFC3696 - page 5:
>    In addition to quoting using the backslash character, conventional
>    double-quote characters may be used to surround strings.  For example
> 
>       "Abc@def"@example.com
> 
>       "Fred Bloggs"@example.com

And theses [examples ](https://en.wikipedia.org/wiki/Email_address#Examples) too:

    "very.unusual.@.unusual.com"@example.com
    "very.(),:;<>[]\".VERY.\"very@\\ \"very\".unusual"@strange.example.com
    "()<>[]:,;@\\\"!#$%&'-/=?^_`{}| ~.a"@example.org
    " "@example.org (space between the quotes)
    user@[2001:DB8::1]

Wich contain an example with litteral IPV6 address in the URL - [RFC2732](https://tools.ietf.org/html/rfc2732)

This pattern work for double quote:
```
^(((?:[\p{L}0-9.!#$%&\'*+\/=?^_`{|}~-]+)*|(".*"))@[\p{L}0-9-_.]+)$
```
And this one if we add IPV6 support:
```
^(((?:[\p{L}0-9.!#$%&\'*+\/=?^_`{|}~-]+)*|(".*"))@([\p{L}0-9-_.]+|\[[0-9A-F:]+\]))$
```

Benfarhat:
Fix #11229

### DIFF
--- a/src/Mailer/Email.php
+++ b/src/Mailer/Email.php
@@ -87,7 +87,7 @@ class Email implements JsonSerializable, Serializable
      *
      * @var string
      */
-    const EMAIL_PATTERN = '/^((?:[\p{L}0-9.!#$%&\'*+\/=?^_`{|}~-]+)*@[\p{L}0-9-._]+)$/ui';
+    const EMAIL_PATTERN = '/^(((?:[\p{L}0-9.!#$%&\'*+\/=?^_`{|}~-]+)*|(".*"))@([\p{L}0-9-_.]+|\[[0-9A-F:]+\]))$/ui';
 
     /**
      * Recipient of the email
--- a/src/Mailer/Email.php
+++ b/src/Mailer/Email.php
@@ -87,7 +87,7 @@ class Email implements JsonSerializable, Serializable
      *
      * @var string
      */
-    const EMAIL_PATTERN = '/^(((?:[\p{L}0-9.!#$%&\'*+\/=?^_`{|}~-]+)*|(".*"))@([\p{L}0-9-_.]+|\[[0-9A-F:]+\]))$/ui';
+    const EMAIL_PATTERN = '/^(((?:[\p{L}0-9.!#$%&\'*+\/=?^_`{|}~-]+)*|(".*"))@([\p{L}0-9-_.]+|\[[0-9A-F:.]+\]))$/ui';
 
     /**
      * Recipient of the email
--- a/tests/TestCase/Mailer/EmailTest.php
+++ b/tests/TestCase/Mailer/EmailTest.php
@@ -293,11 +293,41 @@ class EmailTest extends TestCase
      */
     public function testToDoubleQuotedLocalPart()
     {
-        $result = $this->Email->to('"cake@p h p"@cake_php.org');
-        $expected = ['"cake@p h p"@cake_php.org' => '"cake@p h p"@cake_php.org'];
+        $list = [
+            '"cake@php"@cake_php.org' => '"cake@php"',
+            '"cake php"@cake_php.org' => '"cake php"'
+        ];
+        $this->Email->to($list);
+        $expected = [
+            '"cake@php"@cake_php.org' => '"cake@php"',
+            '"cake php"@cake_php.org' => '"cake php"'
+        ];
         $this->assertSame($expected, $this->Email->to());
-        $this->assertSame($this->Email, $result);
     }
+
+    /**
+     * test to address with IPV4 and IPV6 in domain - RFC5321
+     *
+     * @return void
+     */
+    public function testToLiteralIpv4Ipv6Domain()
+    {
+        $list = [
+            'cakephp@[8.8.8.8]',
+            'cakephp@[2001:0DB8:0:CD30:123:4567:89AB:CDEF]',
+            'cakephp@[2001:DB8::8:800:200C:417A]',
+            'cakephp@[::13.1.68.3]'
+        ];
+        $this->Email->to($list);
+        $expected = [
+            'cakephp@[8.8.8.8]',
+            'cakephp@[2001:0DB8:0:CD30:123:4567:89AB:CDEF]',
+            'cakephp@[2001:DB8::8:800:200C:417A]',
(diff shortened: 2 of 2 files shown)


src/Mailer/Email.php
```diff
@@ -87,7 +87,7 @@ class Email implements JsonSerializable, Serializable
      *
      * @var string
      */
-    const EMAIL_PATTERN = '/^((?:[\p{L}0-9.!#$%&\'*+\/=?^_`{|}~-]+)*@[\p{L}0-9-._]+)$/ui';
+    const EMAIL_PATTERN = '/^(((?:[\p{L}0-9.!#$%&\'*+\/=?^_`{|}~-]+)*|(".*"))@([\p{L}0-9-_.]+|\[[0-9A-F:.]+\]))$/ui';
```
dereuromark: Is the `.*` really true?
And wouldnt this require also weakening the Utility validation for emails? Otherwise they might not end up here to be used anyway.

If so, I am a bit concerned about security issues here, if you basically now allow any email of "SQLInjection..." to be passed, when those then go into the ORM layer.

ADmad: I share @dereuromark's concern.  If we are okay with such lax validation why not simply use `filter_var()` instead of maintaining our own regex?

markstory: The `.*` is within a quoted string. This is for emails like `"a big bunch of silly"@example.com`

dereuromark: @markstory Yeah but you could add `'` or other chars, that could together with custom SQL then make the injection happen.

Benfarhat: @ADmad  FILTER_VALIDATE_EMAIL doesn't pass RFC 3696 (for double-quote local-part) nor RFC 5321 (For IP instead of domain)
```
<?php
var_dump(filter_var('cake@example.com', FILTER_VALIDATE_EMAIL));
var_dump(filter_var('"hello@world"@example.com', FILTER_VALIDATE_EMAIL));
var_dump(filter_var('"hello world"@example.com', FILTER_VALIDATE_EMAIL));
var_dump(filter_var('cakephp@192.168.1.1', FILTER_VALIDATE_EMAIL));
var_dump(filter_var('cakephp@[2001::1]]', FILTER_VALIDATE_EMAIL));
var_dump(filter_var('cakephp@ex_ample.com', FILTER_VALIDATE_EMAIL));
var_dump(filter_var('"\'OR 1=1--@gmail.com"@example.com', FILTER_VALIDATE_EMAIL));
```
Result:
```
string(16) "cake@example.com"
string(25) ""hello@world"@example.com"
bool(false)
bool(false)
bool(false)
bool(false)
bool(false)
```

ADmad: How often are such email ids actually used in practice? Pretty sure no popular email provider allows such ids.

Personally I would rather validate based on common usage instead of technical conformity.

ADmad: I realized that this change is for the `Email` class not email validator. So ensuring that `Email` accepts all technically valid addresses makes sense. In most cases the email id received from user input or fetched from db would have been already validated.

Benfarhat: @markstory @ADmad @dereuromark  The following pattern prevent from using `'` (for SQL injection but using prepared stamements and ORM is supposed to do the stuff) and `"` which avoid adding extra parameters in the double-quoted local-part ([vulnerabilities](https://cve.mitre.org/cgi-bin/cvekey.cgi?keyword=rfc+3696) CVE-2016-10033 and CVE-2016-10034)
```
const EMAIL_PATTERN = '/^(((?:[\p{L}0-9.!#$%&\'*+\/=?^_`{|}~-]+)*|("[^"\']+"))@([\p{L}0-9-_.]+|\[[0-9A-F:.]+\]))$/ui';
```
Applying this pattern give me these results:
```
    test@example.com -> valid
    cakephp.com -> invalid
    "cake@php"@example.com -> valid
    "cake php"@example.com -> valid
    "cake \"php"@example.com -> invalid
    "cake ' OR 1=1php"@example.com -> invalid
    cake@example.com -> valid
    "hello@world"@example.com -> valid
    "hello world"@example.com -> valid
    "he'llo@world"@example.com -> invalid
    cakephp@192.168.1.1 -> valid
    cakephp@[2001::1] -> valid
    cakephp@ex_ample.com -> valid
    "'OR 1=1--@gmail.com"@example.com -> invalid
```

dereuromark: Would it make sense to do sth similar like
https://github.com/symfony/validator/blob/master/Constraints/EmailValidator.php#L61
?
This way the exact and everchanging specifics are outsourced, and can be reqire-dev for tests and can be manually required on project side if desired/needed, otherwise just a basic check is done:

    preg_match('/^.+\@\S+\.\S+$/', $value)

markstory: I remembered that we also allow email address patterns to be swapped at runtime using `$email->setEmailPattern()` which will allow people to use patterns that allow more exotic email addresses.

josegonzalez:
I think we should close this PR. While the change is reasonable, the emails this allows users to submit are almost always going to be typos. If this actually affects users of the framework, they should be encouraged to use `$email->setEmailPattern()` - and can even subclass the Email class to do so - in order to grant them the desired effect.

markstory:
@josegonzalez Agreed.

PR #11236, opened 2017-09-25
Improve doc blocks for IDE understanding.

dereuromark:
Turns out we need mixin annotations when accessing the properties _config directly etc.
The methods work without it, but we still have a lot of direct property access.
This makes it less "red" in IDEs.

Also fixed a few other missing interface annotations. The alternative would be to typehint the closure signatures, but I wasn't sure if that was a performance issue in loops. This way it doesnt cost anything but a single line more.

And by properly annotating everything the IDE then found out that aliasField() has the wrong type annotated :) So that is now also fixed for good.

Note:
When we add methods to classes, but due to BC not to the interface, we must at least add the `@method` annotation in the interface doc block.

### DIFF
--- a/src/ORM/Behavior/CounterCacheBehavior.php
+++ b/src/ORM/Behavior/CounterCacheBehavior.php
@@ -227,7 +227,7 @@ class CounterCacheBehavior extends Behavior
                 continue;
             }
 
-            if (!is_string($config) && is_callable($config)) {
+            if (is_callable($config)) {
                 $count = $config($event, $entity, $this->_table, false);
             } else {
                 $count = $this->_getCount($config, $countConditions);
@@ -236,7 +236,7 @@ class CounterCacheBehavior extends Behavior
             $assoc->getTarget()->updateAll([$field => $count], $updateConditions);
 
             if (isset($updateOriginalConditions)) {
-                if (!is_string($config) && is_callable($config)) {
+                if (is_callable($config)) {
                     $count = $config($event, $entity, $this->_table, true);
                 } else {
                     $count = $this->_getCount($config, $countOriginalConditions);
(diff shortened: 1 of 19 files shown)


src/ORM/Behavior/CounterCacheBehavior.php
```diff
@@ -227,7 +227,7 @@ protected function _processAssociation(Event $event, EntityInterface $entity, As
                 continue;
             }
 
-            if (!is_string($config) && is_callable($config)) {
+            if (is_callable($config)) {
```
markstory: This string check was to prevent global functions being used as callbacks.

dereuromark: I see. That makes sense, I can revert. It was just weird that the method never allowed for any string input. The IDE complained here.

dereuromark: I fixed it with an explicit exception inside the if block. I think that is cleaner.

src/ORM/Behavior/CounterCacheBehavior.php
```diff
@@ -236,7 +241,10 @@ protected function _processAssociation(Event $event, EntityInterface $entity, As
             $assoc->getTarget()->updateAll([$field => $count], $updateConditions);
 
             if (isset($updateOriginalConditions)) {
-                if (!is_string($config) && is_callable($config)) {
+                if (is_callable($config)) {
+                    if (is_string($config)) {
+                        throw new RuntimeException('You must not use a string as callable.');
+                    }
```
markstory: This is a behavior change in that a callable string will now raise an exception. Previously it would be ignored and the default implementation of `_getCount()` would be used. Not sure this will be relevant though.

dereuromark: that method has a array typehint. afaik that would blow up either way.

PR #11238, opened 2017-09-26
3.next - Start command class

markstory:
This is a modest start to the new Command class proposed in #11137. This gets the option parser integration defined as well as creating a hook point for the `CommandCollection` to set the command name so it can be used to create an option parser.

I'm proposing that we use the validator model to build the option parser. Instead of having to call `parent` as we do with `Shell`, developers will implement the `buildOptionParser` method and define their option parser.

### DIFF
--- a/src/Console/Command.php
+++ b/src/Console/Command.php
@@ -28,6 +28,13 @@ class Command
     use ModelAwareTrait;
 
     /**
+     * The name of this command. Inflected from the class name.
+     *
+     * @var string
+     */
+    protected $name;
+
+    /**
      * Constructor
      *
      * By default CakePHP will construct command objects when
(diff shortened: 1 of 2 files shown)


src/Console/Command.php
```diff
@@ -0,0 +1,121 @@
...
+
+    /**
+     * Constructor
+     *
+     * By default CakePHP will construct command objects when
+     * building the CommandCollection for your application.
+     */
+    public function __construct()
+    {
+        $locator = $this->getTableLocator() ? : 'Cake\ORM\TableRegistry';
```
robertpustulka: `getTableLocator()` will return the global locator if none has been set, so it's not necessary.

src/Console/Command.php
```diff
@@ -0,0 +1,121 @@
...
+        $this->modelFactory('Table', [$locator, 'get']);
```
robertpustulka: This could be:
```php
$this->modelFactory('Table', function($alias){
    return $this->getTableLocator()->get($alias);
});
```
This way the factory will use the new locator if it was later injected using  a setter.

src/Console/Command.php
markstory: Thanks. I'll fix that.

src/Console/Command.php
markstory: Good idea.

src/Console/Command.php
```diff
@@ -43,8 +43,10 @@ class Command
      */
     public function __construct()
     {
-        $locator = $this->getTableLocator() ? : 'Cake\ORM\TableRegistry';
-        $this->modelFactory('Table', [$locator, 'get']);
+        $locator = $this->getTableLocator();
```
robertpustulka: This line is no longer necessary.

PR #11244, opened 2017-09-27
added deprecated helper

saeideng:
https://github.com/cakephp/cakephp/pull/11241#issuecomment-332301887

### DIFF
--- a/src/TestSuite/TestCase.php
+++ b/src/TestSuite/TestCase.php
@@ -88,6 +88,20 @@ abstract class TestCase extends BaseTestCase
     }
 
     /**
+     * Helper method for check deprecation methods
+     *
+     * @param callable $callable callable function that will receive asserts
+     * @return void
+     */
+    public function deprecated($callable)
+    {
+        $errorLevel = error_reporting();
+        error_reporting(E_ALL ^ E_USER_DEPRECATED);
+            $callable();
+        error_reporting($errorLevel);
+    }
+
+    /**
      * Setup the test case, backup the static object values so they can be restored.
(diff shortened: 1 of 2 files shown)


src/TestSuite/TestCase.php
```diff
@@ -88,6 +88,20 @@ public function skipIf($shouldSkip, $message = '')
...
+     * Helper method for check deprecation methods
+     *
+     * @param callable $callable callable function that will receive asserts
+     * @return void
+     */
+    public function deprecated($callable)
+    {
+        $errorLevel = error_reporting();
+        error_reporting(E_ALL ^ E_USER_DEPRECATED);
+            $callable();
```
markstory: The indentation here looks off.

markstory (commented):
I like this approach. It helps contain the deprecated feature tests and will make them easier to remove later on.

saeideng:
even we can assert exception message
But there is a bit problem
line and file path of exception message return 
`....src/.... TestSuite/testcase.php line 100 `
So we can not assert this line/file path
But I think we no needs to assert to message
That is  a bit hard to implement
But usage :

```php
This->deprecated(function(){
... 
},[
'Message'=>....
'MessageRegex'=>...
])
```
But Im not sure line after that can be works or no
I need a test

saeideng:
Lines after 'deprecated()' is works correctly now (without assert message)

src/TestSuite/TestCase.php
saeideng: fixed

saeideng:
you can merge this if no need to assert message
we can add extra assert in future if necessary

Issue #11245, opened 2017-09-27
RFC 3.6: Deprecate property access on BehaviorRegistry


dereuromark:
This is a (multiple allowed):

* [ ] bug
* [x] enhancement
* [x] feature-discussion (RFC)

### What you did

Currently people use
```php
$searchManager = $this->behaviors()->Search->searchManager();
// or
$searchManager = $this->behaviors()->get('Search')->searchManager();
```
The latter is bad, because it cannot be auto-discovered by the IDE (I didnt find any way), magic string issues on top.
The former falls in line of direct (public) property access, which is somewhat discouraged now.

I propose to deprecate at least the property one in favor of a consistent method access to behaviors:
```php
$searchManager = $this->behaviors()->search()->searchManager();
```

//EDIT now
```php
$searchManager = $this->behaviors()->getSearch()->searchManager();
```

It can, with some effort, made auto-discoverable for IDEs, thus providing typehinting, clickability, etc.
We can keep the magic string getter if needed as 2nd solution, but usually it is better to not have too many ways to do the same thing IMO.

//cc @davidyell

dereuromark:
If anyone is interested, this generated (by IdeHelper for example) file enables IDE support for the new syntax:

```
namespace Cake\ORM;

exit('Only for code completion');

abstract class BehaviorRegistry extends \Cake\Core\ObjectRegistry {

	/**
	 * Search.Search behavior
	 *
	 * @var \Search\Model\Behavior\SearchBehavior
	 */
	public $Search;

	/**
	 * @return \Search\Model\Behavior\SearchBehavior
	 */
	public function search() {
		return $this->Search;
	}

}
```

ADmad:
I don't like this idea. Accessing behavior instances as function feels wrong. Also there's potential of clashing with actual functions. For eg. since there is a function `ObjectRegistry::reset()` you wouldn't be able to accessing a behavior named `Reset` through your proposal.

dereuromark:
We are also accessing behaviors() and other classes this way, perfectly fine IMO.
The class issue sure is sth that needs to be considered, though.

dereuromark:
Actually, by suffixing the type it would work fine:

    ->searchBehavior()->

etc
This would also make it more speaking.

ADmad:
`->behaviors()->searchBehavior()` is unnecessarily redundant. IDE auto completing is nice to have but we shouldn't be uglifying code just for that purpose. Not everyone uses IDEs.

[...]

PR #11246, opened 2017-09-27
Use method instead of public property.

dereuromark:
Implements https://github.com/cakephp/cakephp/issues/11245 on all object registry levels.

//EDIT: As mentioned we need suffixing to avoid collisions with existing method names, adjusting the examples:

E.g.
```php
$searchManager = $this->behaviors()->Search->searchManager();
// becomes
$searchManager = $this->behaviors()->getSearch()->searchManager();
```

Deprecates property access as minimal changeset so far.
If acceptable, I can finalize the tests here.

Since it is auto-competable via IDE you actually have to type less, especially no magic strings here.

You can also continue to use the internally used
```php
$searchManager = $this->behaviors()->get('Search')->searchManager();
```
if you want to, but for this you always need a manual inline doc block to annotate the last chaining element it for IDEs. It also wont work for any element before in the chain here then.

### DIFF
--- a/src/ORM/BehaviorRegistry.php
+++ b/src/ORM/BehaviorRegistry.php
@@ -67,6 +67,20 @@ class BehaviorRegistry extends ObjectRegistry implements EventDispatcherInterfac
     }
 
     /**
+     * Get loaded behavior instance.
+     *
+     * @param string $name Name of behavior.
+     * @return \Cake\ORM\Behavior|object|null
+     */
+    public function __call($name, $params)
+    {
+        $name = substr($name, 0, -8);
+        $name = ucfirst($name);
+
+        return $this->get($name);
+    }
+
+    /**
      * Attaches a table instance to this registry.
      *
      * @param \Cake\ORM\Table $table The table this registry is attached to.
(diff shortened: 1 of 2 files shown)


src/ORM/BehaviorRegistry.php
```diff
@@ -67,6 +67,19 @@ public function __construct($table = null)
...
+     * Get loaded behavior instance.
+     *
+     * @param string $name Name of behavior.
+     * @return \Cake\ORM\Behavior|object|null
+     */
+    public function __call($name, $params)
+    {
+        $name = substr($name, 3);
+
+        return $this->get($name);
```
markstory: This seems like it would limit our options in the future. Shouldn't we check the prefix and throw an error if it is not get?

dereuromark: Sure a possiblity.

dereuromark:
In light of https://github.com/dereuromark/cakephp-ide-helper/pull/56 - and that both magic property and magic method access work fine here in IDEs - I will close for now. No immediate problem to be solved.

PR #11255, opened 2017-09-29
Fix doc blocks around Entity.

dereuromark:
The IDE shows it yellow in Table class where it is accessed via array type.

### DIFF
--- a/src/Datasource/EntityTrait.php
+++ b/src/Datasource/EntityTrait.php
@@ -151,7 +151,7 @@ trait EntityTrait
      * Returns whether this entity contains a property named $property
      * regardless of if it is empty.
      *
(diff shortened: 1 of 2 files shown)


src/Datasource/EntityTrait.php
```diff
@@ -151,7 +151,7 @@ public function __set($property, $value)
      * Returns whether this entity contains a property named $property
      * regardless of if it is empty.
      *
-     * @param string $property The property to check.
+     * @param string|array $property The property to check.
```
ADmad: Magic `__isset()` only takes a string.

dereuromark: Oh, you are right, thanks.

PR #11256, opened 2017-09-30
3.next - Implement basic Command dispatching

markstory:
Implement the basic code required for commands to be callable.

* Enable commands to be added to the collection
* Add basic construction of commands.
* Add help and output levels.

In the next set of changes I'll focus on file creation, and formatted text.

Refs #11137

### DIFF
--- a/src/Console/Command.php
+++ b/src/Console/Command.php
@@ -18,8 +18,8 @@ use Cake\Console\Exception\ConsoleException;
 use Cake\Datasource\ModelAwareTrait;
 use Cake\Log\LogTrait;
 use Cake\ORM\Locator\LocatorAwareTrait;
-use RuntimeException;
 use InvalidArgumentException;
+use RuntimeException;
 
 /**
  * Base class for console commands.
@@ -150,7 +150,8 @@ class Command
     /**
      * Run the command.
      *
-     * @param array $argv
+     * @param array $argv Arguments from the CLI environment.
+     * @param \Cake\Console\ConsoleIo $io The console io
      * @return int|null Exit code or null for success.
      */
     public function run(array $argv, ConsoleIo $io)
@@ -175,6 +176,7 @@ class Command
         if ($args->getOption('help')) {
             return $this->displayHelp($parser, $args, $io);
         }
+
         return $this->execute($args, $io);
     }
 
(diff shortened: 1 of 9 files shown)


src/Console/CommandRunner.php
```diff
@@ -200,6 +240,11 @@ protected function getShell(ConsoleIo $io, CommandCollection $commands, $name)
      */
     protected function createShell($className, ConsoleIo $io)
     {
-        return new $className($io);
+        if (is_subclass_of($className, Shell::class)) {
+            return new $className($io);
+        }
+
+        // Command class
+        return new $className();
```
dereuromark: Is the doc block `@return \Cake\Console\Shell` then correct?

src/Console/Command.php
```diff
@@ -52,12 +68,20 @@ public function __construct()
...
+     * Required to have at least one space in the name so that the root
+     * command can be calculated.
      *
      * @param string $name The name the command uses in the collection.
      * @return $this;
+     * @throws \InvalidArgumentException
      */
     public function setName($name)
     {
+        if (strpos($name, ' ') === false) {
```
robertpustulka: This will allow space to be a first character. Is that something we should worry about?

src/Console/CommandRunner.php
```diff
@@ -159,10 +161,39 @@ public function run(array $argv, ConsoleIo $io = null)
...
-     * @return \Cake\Console\Shell
+     * @return \Cake\Console\Shell|\Cake\Console\Command
      */
     protected function getShell(ConsoleIo $io, CommandCollection $commands, $name)
     {
+        $instance = $commands->get($name);
+        if (is_string($instance)) {
+            $instance = $this->createShell($instance, $io);
+        }
+        if (method_exists($instance, 'setRootName')) {
```
robertpustulka: Maybe an instance check would be better here?

```php
if ($instance instanceof Shell) {
     $instance->setRootName($this->root);
}
if ($instance instanceof Command) {
     $instance->setName("{$this->root} {$name}");
}
```

These are only two classes that the shell/command can extend.

src/Console/CommandRunner.php
markstory: Fixed now.

src/Console/CommandRunner.php
markstory: I can do that.

src/Console/Command.php
markstory: Yes. I'll fix that.

PR #11259, opened 2017-10-02
Fix: void unreturn

lustyrain:
Hi.

Is not return unnecessary?


dereuromark:
Technically it is not even allowed :) Void return methods must not return something.
Nice find!

inoas:
Could this be caught by some linters/checkers in future?

dereuromark:
I doubt that in 2.x - any phpstan might not work due to the lack of namespaces.

markstory:
Thanks!

PR #11263, opened 2017-10-03
Fix: phpDocs miss

lustyrain:
Hi.
 
This time, I corrected the point I made a mistake in the place to understand.
The change is many, but please check it.
 

## did
- void unReturn
- miss return void
- add return type
- type miss typing
- add param type and return type
  - string → string|array
- change ClassName

### DIFF
--- a/lib/Cake/Console/ConsoleErrorHandler.php
+++ b/lib/Cake/Console/ConsoleErrorHandler.php
@@ -60,7 +60,7 @@ class ConsoleErrorHandler {
 		));
 		$code = $exception->getCode();
 		$code = ($code && is_int($code)) ? $code : 1;
-		return $this->_stop($code);
+		$this->_stop($code);
 	}
 
 /**
@@ -88,7 +88,7 @@ class ConsoleErrorHandler {
 		}
 
 		if ($log === LOG_ERR) {
-			return $this->_stop(1);
+			$this->_stop(1);
 		}
 	}
 
--- a/lib/Cake/Console/ConsoleErrorHandler.php
+++ b/lib/Cake/Console/ConsoleErrorHandler.php
@@ -60,7 +60,7 @@ class ConsoleErrorHandler {
 		));
 		$code = $exception->getCode();
 		$code = ($code && is_int($code)) ? $code : 1;
-		$this->_stop($code);
+		return $this->_stop($code);
 	}
 
 /**
--- a/lib/Cake/Console/ConsoleErrorHandler.php
+++ b/lib/Cake/Console/ConsoleErrorHandler.php
@@ -88,7 +88,7 @@ class ConsoleErrorHandler {
 		}
 
 		if ($log === LOG_ERR) {
-			$this->_stop(1);
+			return $this->_stop(1);
 		}
 	}
 
--- a/lib/Cake/Event/CakeEventManager.php
+++ b/lib/Cake/Event/CakeEventManager.php
@@ -167,7 +167,7 @@ class CakeEventManager {
 	public function detach($callable, $eventKey = null) {
 		if ($callable instanceof CakeEventListener) {
 			$this->_detachSubscriber($callable, $eventKey);
-			return ;
+			return;
 		}
 		if (empty($eventKey)) {
 			foreach (array_keys($this->_listeners) as $eventKey) {
--- a/lib/Cake/Event/CakeEventManager.php
+++ b/lib/Cake/Event/CakeEventManager.php
@@ -166,7 +166,8 @@ class CakeEventManager {
  */
 	public function detach($callable, $eventKey = null) {
 		if ($callable instanceof CakeEventListener) {
-			return $this->_detachSubscriber($callable, $eventKey);
+			$this->_detachSubscriber($callable, $eventKey);
+			return ;
 		}
 		if (empty($eventKey)) {
 			foreach (array_keys($this->_listeners) as $eventKey) {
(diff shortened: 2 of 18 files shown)


lib/Cake/Utility/File.php
```diff
@@ -581,10 +581,10 @@ public function mime() {
  */
 	public function clearStatCache($all = false) {
 		if ($all === false && version_compare(PHP_VERSION, '5.3.0') >= 0) {
-			return clearstatcache(true, $this->path);
+			clearstatcache(true, $this->path);
```
chinpei215: This change would make `clearstatcache()` be called twice.

lib/Cake/Console/ConsoleErrorHandler.php
```diff
@@ -60,7 +60,7 @@ public function handleException($exception) {
 		));
 		$code = $exception->getCode();
 		$code = ($code && is_int($code)) ? $code : 1;
-		return $this->_stop($code);
+		$this->_stop($code);
```
chinpei215: I think we should keep the current behavior as it is, when we found this kind of code. The current behavior allow people to return any value by overriding `_stop()`. So this change might break exiting applications. Instead, it would be better to change phpDoc blocks only.

markstory: These returns are needed for mocks to work properly.

dereuromark: Yes. Must be reverted and the doc blocks fixed instead.

lib/Cake/Utility/File.php
```diff
@@ -581,10 +581,11 @@ public function mime() {
...
+			return ;
```
dereuromark: There is a space too much here before `;`.

lustyrain: Sorry.
I mistook a rule with another project.

lib/Cake/Console/ConsoleErrorHandler.php
lustyrain: Reverted. 
I do not know the optimal answer for doc blocks :,-(

lib/Cake/Console/ConsoleErrorHandler.php
```diff
@@ -88,7 +88,7 @@ public function handleError($code, $description, $file = null, $line = null, $co
 		}
 
 		if ($log === LOG_ERR) {
-			return $this->_stop(1);
```
saeideng: and this??

lib/Cake/Event/CakeEventManager.php
```diff
@@ -166,7 +166,8 @@ protected function _extractCallable($function, $object) {
  */
 	public function detach($callable, $eventKey = null) {
 		if ($callable instanceof CakeEventListener) {
-			return $this->_detachSubscriber($callable, $eventKey);
+			$this->_detachSubscriber($callable, $eventKey);
+			return ;
```
saeideng: phpcs 
> Space found before semicolon; expected "return;" but found "return ;"

lib/Cake/Console/ConsoleErrorHandler.php
lustyrain: thank you. restored return.

lib/Cake/Event/CakeEventManager.php
lustyrain: thank you. fix

lib/Cake/Event/CakeEventManager.php
```diff
@@ -166,7 +166,8 @@ protected function _extractCallable($function, $object) {
  */
 	public function detach($callable, $eventKey = null) {
 		if ($callable instanceof CakeEventListener) {
-			return $this->_detachSubscriber($callable, $eventKey);
+			$this->_detachSubscriber($callable, $eventKey);
+			return;
```
chinpei215: I think this change also should be reverted. Currently, this method allows people to return any types by overriding `_detachSubscriber()`.

dereuromark: Is Event::detach() being checked somewhere? Because this method here also should return void and not be checked for return values ideally.

ADmad:
It would be best to stick to just docblock updates and not make code changes unless there's an actual bug to fix. Such code changes like changing return values have more potential to do harm and then good.

ADmad:
@lustyrain Please revert all code changes unless you can also provide test cases / examples code showing a bug.

lib/Cake/Utility/File.php
dereuromark: This specific code change can be kept now IMO.

lib/Cake/Event/CakeEventManager.php
chinpei215: Ideally, you are right. But people might write such code. For example, talking about Table::find() in 3.x, you could override it to return `int` to support `find('count')`, despite of that the phpDoc says it `@return \Cake\ORM\Query`.

lustyrain:
@ADmad I got it. Revert all code changes. I will fix the wrong docblock.

dereuromark (approved):

lib/Cake/Event/CakeEventManager.php
lustyrain: I got it. I got smarter. reverted code.

lib/Cake/Utility/File.php
lustyrain: Me too.  but, This time, modify the docblock.

Issue #11267, opened 2017-10-04
Add support for default value in Form schema.


robertpustulka:
This is a (multiple allowed):

* [x] bug
* [x] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.5.3

### What you did
Created a form class

```php
class MyForm extends Form
{
    protected function _buildSchema(Schema $schema): Schema
    {
        return $schema
            ->addField('field', [
                'type' => 'string',
                'default' => 'value'
            ]);
    }
}
```

### What happened
Default schema value gets ignored by `FormContext` even if the `schemaDefault` option is set to `true` (which is by default)

### What you expected to happen
The default schema value should be used if no value is found in request data.

PR #11269, opened 2017-10-04
Form schema default

ADmad:
Refs #11267

### DIFF
--- a/src/View/Form/FormContext.php
+++ b/src/View/Form/FormContext.php
@@ -94,7 +94,28 @@ class FormContext implements ContextInterface
             return $val;
         }
 
-        return $options['default'];
+        if ($options['default'] !== null || !$options['schemaDefault']) {
+            return $options['default'];
+        }
+
+        return $this->_schemaDefault($field);
+    }
+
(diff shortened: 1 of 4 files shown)


robertpustulka (approved):

src/View/Form/FormContext.php
```diff
@@ -94,7 +94,28 @@ public function val($field, $options = [])
...
+     *
+     * @param string $field Field name.
+
+     * @return mixed
+     */
+    protected function _schemaDefault($field)
+    {
+        $field = $this->_form->schema()->field($field);
+        if ($field) {
+            return $field['default'];
```
markstory: Should we isset() check on `$field['default']` instead of inferring the key based on the field existing.

robertpustulka: `default` key should always exist:

https://github.com/cakephp/cakephp/blob/ba8dfb1a9b284c335970a41ff9ff254e3189717c/src/Form/Schema.php#L71

markstory: :+1:

PR #11270, opened 2017-10-04
Fix Credit card number pattern(JCB) is wrong

tenkoma:
My colleague found a bug that valid JCB card number can't be verified with validation.
The explanation is as follows.

> JCB: `^(?:2131|1800|35\d{3})\d{11}$` JCB cards beginning with 2131 or 1800 have 15 digits. JCB cards beginning with 35 have 16 digits.
> https://www.regular-expressions.info/creditcard.html

Other frameworks are as follows.

## symfony validator

https://github.com/symfony/validator/blob/master/Constraints/CardSchemeValidator.php#L61

## zend-validator (First 4 digit list)

https://github.com/zendframework/zend-validator/blob/010084ddbd33299bf51ea6f0e07f8f4e8bd832a8/src/CreditCard.php#L111

If merged, it will also backport to 2.x branch.


dereuromark:
Sounds good :+1:

tenkoma:
@dereuromark Ummm...🤔
I can't find the card number verification function https://github.com/cakephp/localized .

dereuromark:
My bad, the cc() function was never ported to localized, as this is not really localized probably.

chinpei215:
@tenkoma I have a question. Would your colleague really have encountered the actual JCB card number beginning with 2131? And are you sure about that JCB card numbers beginning with 35 only?

I took a look at the history of Wikipedia. The prefix 2131 was added to the card numbering list by an IP user on [01:31 30 April 2005](https://en.wikipedia.org/w/index.php?title=Credit_card&diff=prev&oldid=13001004) originally. The user also changed 35 to 3 at the same time, for some reasons. Probably, we would referred to this information to implement `cc()`. Therefore, our regex would have become a little bit rough.

The prefix 2100 in the `cc()` method would be a typo. So I don't mind if we fix it and port the change to 2.x. But if we believe the IP user, I am not sure whether we should change `3\\d{4}` to `35\\d{3}`.

chinpei215:
I finally found an official source. See https://www.jcb.co.jp/point/use/transfer/premo-charge.html (Japanese). According to the page, JCB premo card number begins with 3337. So we should revert the change about `3\\d{4}` at least.

Also I think it is not always a good idea to referrer to other frameworks when we solve Japanese-specific issues, as CakePHP is the most popular framework in Japan.

Other related documents:
https://stackoverflow.com/questions/35565991/reg-ex-for-jcb-credit-card-validation
https://en.wikipedia.org/wiki/Talk:Payment_card_number#JCB_range_too_small

markstory:
@chinpei215 or @tenkoma could one of you make the necessary corrections for the jcb pattern? I don't know those card numbers at all.

chinpei215:
We might have referred to [ccsrch](https://github.com/adamcaudill/ccsrch) to implement `cc()`. I downloaded the oldest version of ccsrch released on 15 July 2005. I quote the related code:

```c
/*
 * Logic Assumptions 
 * 1. Cards can be a minimum of 13 numbers and up to 16 numbers 
 * 2. Card numbers must be contiguous.  The only characters ignored
 *    when processing the files are carriage returns, new line feeds, and nulls.
 * 3. Card prefixes are validated to public info. These sites include
 *    http://javascript.internet.com/forms/val-credit-card.html
 *    http://www.eriandanna.com/Business/CreditCards/credit_card_authorization.htm 
 * 4. Files are treated as raw binary objects 
 * 5. Compressed or encoded files are not uncompressed or decoded 
 * 6. Solo and Switch cards are not processed
 * 
 */

...

void 
check_jcb_16(int *num, long offset)
{
  char            num2buf[5];
  int             vnum = 0;

  memset(&num2buf, '\0', 2);
  snprintf(num2buf, 5, "%d%d%d%d\0", num[0], num[1], num[2], num[3]);
  vnum = atoi(num2buf);
  /* http://javascript.internet.com/forms/val-credit-card.html */
  if ((vnum == 3088) || (vnum == 3096) || (vnum == 3112) || (vnum == 3158) || (vnum == 3337) || (vnum == 3528) || (vnum == 3529))
    print_result("JCB", num, 16, offset);
}

...

void 
check_jcb_15(int *num, long offset)
{
  char            num2buf[5];
  int             vnum = 0;

  memset(&num2buf, '\0', 5);
  snprintf(num2buf, 5, "%d%d%d%d\0", num[0], num[1], num[2], num[3]);
  vnum = atoi(num2buf);
  if ((vnum == 2131) || (vnum == 1800) || (vnum == 3528) || (vnum == 3529))
    print_result("JCB", num, 15, offset);
}
```

So probably, in the past, there were some sources about those JCB prefixes, and 2100 in our code would be a typo. However, test cases for 3088, 3112 and so on in our code were intended, and 3337 is still used for JCB premo card at least. It is not a credit card though.

If we want to validate available JCB credit cards only, I think we could change `3\\d{4}` to `35\\d{3}`. But, in this case, I think 2131 and 1800 also should be removed, as I don't think those numbers are still available. Those have been already removed from Wikipedia for example, as of [20:25, 4 March 2009](https://en.wikipedia.org/w/index.php?title=Payment_card_number&diff=prev&oldid=275000515). According to [the latest version of the page](https://en.wikipedia.org/wiki/Payment_card_number#Issuer_identification_number_.28IIN.29
), the current available range is 3528–3589. There is an reliable source.

Anyway, I think the change of `3\\d{4}` is more than a bug fix. It should not be applied into bugfix releases. The change of 2100 could be applied into bugfix releases though.

@tenkoma What is an actual problem your colleague found? Did he/she encounter an available card number beginning with 2131? Or did he/she want to restrict the range of JCB card number?

chinpei215:
We discussed this issue in the japanese channel on our Slack. @tenkoma will send another pull request when he has time.

PR #11278, opened 2017-10-05
Add notSameAs validation rule.

robertpustulka:
This is the opposite to `sameAs` validation rule. The rule will pass only when field values are different.

I added `Validation::compareFields()` method that allows to check if the comparison result matches the expected boolean value. ~~This is a replacement for `compareWith()` method which is now deprecated.~~

Any suggestions for a better implementation are welcome.

### DIFF
--- a/src/Validation/Validation.php
+++ b/src/Validation/Validation.php
@@ -316,7 +316,7 @@ class Validation
     /**
      * Compare one field to another.
      *
-     * Return true if the comparison match the expected result.
+     * Return true if the comparison matches the expected result.
      *
      * @param mixed $check The value to find in $field.
      * @param string $field The field to check $check against. This field must be present in $context.
--- a/src/Validation/Validation.php
+++ b/src/Validation/Validation.php
@@ -306,14 +306,34 @@ class Validation
      * @param string $field The field to check $check against. This field must be present in $context.
      * @param array $context The validation context.
      * @return bool
+     * @deprecated 3.6.0 Use compareFields() instead.
      */
     public static function compareWith($check, $field, $context)
+    {
+        return self::compareFields($check, $field, true, $context);
+    }
+
+    /**
+     * Compare one field to another.
+     *
+     * Return true if the comparison match the expected result.
+     *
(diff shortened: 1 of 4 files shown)


saeideng:
I no understand why you flagged `compareWith` to deprecated !!?

robertpustulka:
@saeideng the new method does nearly exactly the same thing so there's no need to keep the two of them.

saeideng:
```php
compareWith($check, $field, $context)
compareFields($check, $field, $result, $context)
```
I am :-1:  with deprecating `compareWith()`
you can deprecated this when you have method like
```php
compareFields($check, $field, $context,$result=true)
```

even in this case I prefer to add `result` arg to `compareWith()` if possible 
```php
compareWith($check, $field, $context,$result=true)
```

chinpei215:
@saeideng Sorry if I am wrong, but your suggestion might not work. Since the `$context` parameter of validation methods are handled magically, I think it must be placed at the end of arguments. See [RulesProvider](https://github.com/cakephp/cakephp/blob/master/src/Validation/RulesProvider.php).

src/Validation/Validation.php
```diff
@@ -306,14 +306,34 @@ public static function comparison($check1, $operator, $check2)
      * @param string $field The field to check $check against. This field must be present in $context.
      * @param array $context The validation context.
      * @return bool
+     * @deprecated 3.6.0 Use compareFields() instead.
```
markstory: Not sure this has to be deprecated. The new function is slightly more awkward to use because of the additional arguments.

chinpei215:
Whouldn't be there any ways to implement a feature like logicalNot() in PHPUnit?

robertpustulka:
@chinpei215 something like a wrapper for another rule?

chinpei215:
@robertpustulka Yes. I thought it might be useful if we want to implement Validation::notEqualTo() or something.

markstory (approved):
Looks good to me.

robertpustulka:
Thanks

PR #11279, opened 2017-10-05
3.x Make ApcuEngine more atomic & reliant

inoas:
Currently trying to make ApcEngine work I was wondering why it had so many non atomic operations and no exceptions?

### DIFF
--- a/src/Cache/Engine/ApcEngine.php
+++ b/src/Cache/Engine/ApcEngine.php
@@ -140,7 +140,7 @@ class ApcEngine extends CacheEngine
      *    from APC as they expired. This flag is really only used by FileEngine.
      * @return bool True Returns true.
      * @throws RuntimeException If cache info cannot be retrieved from APCu cache.
-     * @throws RuntimeException If cache a stored key could not be removed from APCu cache. 
+     * @throws RuntimeException If cache a stored key could not be removed from APCu cache.
      */
     public function clear($check)
     {
@@ -163,7 +163,8 @@ class ApcEngine extends CacheEngine
             if (strpos($key['info'], $this->_config['prefix']) === 0) {
                 if (apcu_delete($key['info']) === false) {
                     throw new RuntimeException(
-                        sprintf('Failed to remove stored key "%s" from APCu cache.', $key['info']));
+                        sprintf('Failed to remove stored key "%s" from APCu cache.', $key['info'])
+                    );
                 }
             }
         }
@@ -225,7 +226,8 @@ class ApcEngine extends CacheEngine
                     $value = 1;
                     if (apcu_store($group, $value) === false) {
                         throw new RuntimeException(
-                            sprintf('Failed to store key "%s" with value "%s" into APCu cache.', $group, $value));
+                            sprintf('Failed to store key "%s" with value "%s" into APCu cache.', $group, $value)
+                        );
                     }
                     $groups[$group] = $value;
                 }
--- a/src/Cache/Engine/ApcEngine.php
+++ b/src/Cache/Engine/ApcEngine.php
@@ -66,9 +66,8 @@ class ApcEngine extends CacheEngine
         if ($duration) {
             $expires = time() + $duration;
         }
-        apcu_store($key . '_expires', $expires, $duration);
 
-        return apcu_store($key, $value, $duration);
+        return apcu_store($key . '_expires', $expires, $duration) === true && apcu_store($key, $value, $duration) === true;
     }
 
     /**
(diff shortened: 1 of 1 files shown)


src/Cache/Engine/ApcEngine.php
```diff
@@ -66,9 +66,8 @@ public function write($key, $value)
         if ($duration) {
             $expires = time() + $duration;
         }
-        apcu_store($key . '_expires', $expires, $duration);
 
-        return apcu_store($key, $value, $duration);
+        return apcu_store($key . '_expires', $expires, $duration) === true && apcu_store($key, $value, $duration) === true;
```
markstory: Does this all need to be on one line?

src/Cache/Engine/ApcEngine.php
```diff
@@ -149,13 +148,13 @@ public function clear($check)
                 '/^' . preg_quote($this->_config['prefix'], '/') . '/',
                 APC_ITER_NONE
             );
-            apcu_delete($iterator);
 
-            return true;
+            return apcu_delete($iterator) === true;
         }
-        $cache = apcu_cache_info();
+        $cache = apcu_cache_info();// Shouldn't this throw an exception if false?
```
markstory: I agree that this should probably be an exception.

src/Cache/Engine/ApcEngine.php
```diff
@@ -181,9 +180,8 @@ public function add($key, $value)
         if ($duration) {
             $expires = time() + $duration;
         }
-        apcu_add($key . '_expires', $expires, $duration);
 
-        return apcu_add($key, $value, $duration);
+        return apcu_add($key, $value, $duration) === true && apcu_add($key . '_expires', $expires, $duration);
```
markstory: I think the return checking is a good idea. Not a fan of the really long line.

inoas: Well the long line was just cause I did it quickly through github.

inoas:
I wanted to see if what I was proposing was welcomed first.

Personal Note: I did the changes through github. I since discovered that the apcu problems with apcu not running on cli where (probably) because it makes no sense technically (no shared php fpm @hmic).

I was getting errors that they key simply was not being able to be written - but only for some keys, and thought it would help to improve the atomicness/strictness of the engine.

src/Cache/Engine/ApcEngine.php
```diff
@@ -16,9 +16,10 @@
 
 use APCUIterator;
 use Cake\Cache\CacheEngine;
+use RuntimeException;
 
 /**
- * APC storage engine for cache
+ * APCu storage engine for cache
  */
 class ApcEngine extends CacheEngine
```
inoas: Should this not be named ApcuEngine, then have a class_alias from ApcEngine?

src/Cache/Engine/ApcEngine.php
```diff
@@ -149,14 +169,20 @@ public function clear($check)
                 '/^' . preg_quote($this->_config['prefix'], '/') . '/',
                 APC_ITER_NONE
             );
-            apcu_delete($iterator);
 
-            return true;
+            return apcu_delete($iterator) === true;
```
inoas: This should throw an exception, too, instead of returning false?
Reason being: the return type for clear was always `bool true` and now it can suddenly be `bool false` and people would probably not check if clearing was successful. An exception would tell them that it wasn't?

src/Cache/Engine/ApcEngine.php
markstory: Could do that, as APC is kind of done now.

src/Cache/Engine/ApcEngine.php
markstory: We don't throw exceptions from other cache engines when deletes fail though.

inoas: And that's good behavior?

src/Cache/Engine/ApcEngine.php
inoas: I can do that in a separate PR after this is done.

src/Cache/Engine/ApcEngine.php
inoas: We could also make this configurable for now and switch around the defaults to 'throw exception' in 4.x?

markstory: Generally we don't want caching operations to destroy an inflight request. Caching should be used as an optimization or performance improvement not as critical path code that needs to be 100% reliable. At least that is how I approach cache usage.

inoas: Should we throw a warning then, instead? Maybe for all the exceptions I was adding here?
Make the throwing of the warning configurable (true by default)?

markstory: Using trigger_error() might be a good option here, as that will generally end up in the logs.

inoas:
Closing as work continues in #11290

PR #11281, opened 2017-10-06
3.next - Convert help & version to commands

markstory:
Convert the Help/VersionShells into Commands. I've removed the old classes, which might be a backwards incompatible change depending on what you consider the 'API' of the console libraries is. I'm OK with reverting the deletions if folks don't feel comfortable with that part of the change.


markstory:
Part of #11137

Issue #11280, opened 2017-10-06
[2.x] CookieComponent problems


chinpei215:
This is a bug.

* CakePHP Version: 2.10.3
* Platform and Target: Non-related

This ticket contains multiple problems of 2.x CookieComponent. I am planning to send multiple pull requests to fix them step by step.

### Problems

* [x] Problem 1 - The write() method throws fatal error when replacing scalar with array
* [ ] Problem 2 - The read() method restores original values if cookies have been completely deleted
* [x] Problem 3 - Deleting deep entry doesn't work (As of 2.4.3)
* [ ] Problem 4 - Writing cookie with expiration time for specific entry doesn't work (As of 2.4.3)

### TestsController

```php
class TestsController extends AppController
{
	public $autoRender = false;

	public function read() {
		debug( $this->Cookie->read() );
	}

	public function write() {
		$this->Cookie->write('foo.bar', 'value');
	}

	public function deleteDeepEntry() {
		$this->Cookie->delete('foo.bar');
	}

	public function replaceScalarWithArray() {
		$this->Cookie->write('foo', 1);
		$this->Cookie->write('foo.bar', 2);
	}

	public function readAfterDelete() {
		$this->Cookie->delete('foo');
		debug( $this->Cookie->read() );
	}

	public function writeWithExpirationTime() {
		$this->Cookie->write('foo.baz', 'value', true, '5 second');
	}
}
```

----

### Problem 1

The write() method throws fatal error when replacing scalar with array.

#### Test

1. Access /tests/replaceScalarWithArray

#### Expected

Works.

#### Actual

```
Argument 1 passed to Hash::insert() must be of the type array, integer given, called in ...
```

----

### Problem 2

The read() method restores original values if cookies have been completely deleted.

#### Test

1. Access /tests/write
2. Access /tests/readAfterDelete

#### Expected

```
array()
```

#### Actual

```
array(
	'foo' => array(
		'bar' => 'value'
	)
)
```

----

### Problem 3

Deleting deep entry doesn't work (As of 2.4.3).

#### Test

1. Access /tests/write
2. Access /tests/deleteDeepEntry
3. Access /tests/read

#### Expected

```
array()
```

#### Actual

```
array(
	'foo' => array(
		'bar' => 'value'
	)
)
```

----

### Problem 4

Writing cookie with expiration time for specific entry doesn't work (As of 2.4.3).

#### Test

1. Access /tests/write
2. Access /tests/writeWithExpirationTime
[...]

PR #11283, opened 2017-10-06
[2.x] Fix fatal error thrown when replacing scalar with array

chinpei215:
I simplified the code on the way to fixing the problem.

Refs #11280 (Problem 1)


markstory:
Nice work :clap:

PR #11284, opened 2017-10-06
msSQL - also handle offset as string

kolorafa:
When doing pagination you could get offset not as a int(eg. 10) but string(eg. "10") and it will not paginate at all.

For example DataTables plugin pass offset from params and all params from http request are strings or numbers wrapped in strings.

Adding ctype_digit($offset) will also check the case.
ctype_digit will only allow string cotaining pure numbers.


markstory (commented):
Any chance you could update the test cases for this?

Issue #11253, opened 2017-09-28
Security::hash() should throw an exception if algo not found


inoas:
This is a (multiple allowed):

* [x] bug
* [x] enhancement

* CakePHP Version: 3.4.2

### What you did
`$hash = Security::hash('foobar', 'sha512a');`

`sha512a` doesn't exist, only `sha512` does (on my system).

### What happened
Warning

### What you expected to happen
Exception

Reason: This should break your application not just throw a warning for better security.
Result was that $hash became `false`. Labeled as bug just cause I consider this bad practice for security.

I could probably create a PR if I can make the test cases catch the exception.
This should probably happen for 4.x as it is non-BC?

dereuromark:
In 3.6 the exception sure makes sense. For 3.5 we could think about triggering a warning instead.

markstory:
Adding exceptions is acceptable as far as backwards compatibility is concerned. Would you be able to put a pull request together @inoas ?

inoas:
I'll take a look this weekend / next week, I suppose.

inoas:
Closing as PR is open.

PR #11285, opened 2017-10-06
#11253 Check hash algorithm availability

burzum:
See #11253

### DIFF
--- a/src/Utility/Security.php
+++ b/src/Utility/Security.php
@@ -51,7 +51,7 @@ class Security
      * Create a hash from string using given method.
      *
      * @param string $string String to hash
-     * @param string|null $type Hashing algo to use (i.e. sha1, sha256 etc.).
+     * @param string|null $algorithm Hashing algo to use (i.e. sha1, sha256 etc.).
      *   Can be any valid algo included in list returned by hash_algos().
      *   If no value is passed the type specified by `Security::$hashType` is used.
      * @param mixed $salt If true, automatically prepends the application's salt
@@ -59,19 +59,19 @@ class Security
      * @return string Hash
      * @link https://book.cakephp.org/3.0/en/core-libraries/security.html#hashing-data
      */
-    public static function hash($string, $type = null, $salt = false)
+    public static function hash($string, $algorithm = null, $salt = false)
     {
-        if (empty($type)) {
-            $type = static::$hashType;
+        if (empty($algorithm)) {
+            $algorithm = static::$hashType;
         }
-        $type = strtolower($type);
+        $algorithm = strtolower($algorithm);
 
-        $hashAlgos = hash_algos();
-        if (!in_array($type, $hashAlgos)) {
+        $availableAlgorithms = hash_algos();
+        if (!in_array($algorithm, $availableAlgorithms)) {
             throw new RuntimeException(sprintf(
                 'The hash type `%s` was not found. Available algorithms are: %s',
-                $type,
-                implode(', ', $hashAlgos)
+                $algorithm,
+                implode(', ', $availableAlgorithms)
             ));
         }
 
@@ -82,7 +82,7 @@ class Security
             $string = $salt . $string;
         }
 
-        return hash($type, $string);
+        return hash($algorithm, $string);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


inoas:
Thanks for taking care. I totally forgot about it.

src/Utility/Security.php
```diff
@@ -66,6 +66,15 @@ public static function hash($string, $type = null, $salt = false)
         }
         $type = strtolower($type);
 
+        $hashAlgos = hash_algos();
+        if (!in_array($type, $hashAlgos)) {
+            throw new RuntimeException(sprintf(
+                'The hash type `%s` was not found. Available algorithms are: %s',
```
inoas: instead of type:
`hash algorithm`
`hash algo`
The function variable could also be renamed to `$algo`.

tests/TestCase/Utility/SecurityTest.php
```diff
@@ -70,6 +70,20 @@ public function testHash()
...
     /**
+     * testInvalidHashTypeException
+     *
+     * @return void
+     */
+    public function testInvalidHashTypeException() {
+        try {
+            Security::hash('test', 'doesnotexist', false);
+            $this->fail('Expected \RuntimeException');
+        } catch (\RuntimeException $e) {
```
inoas: Is it common to do FQNS in tests?

tests/TestCase/Utility/SecurityTest.php
```diff
@@ -70,6 +70,20 @@ public function testHash()
...
+            $this->assertTextContains('The hash type `doesnotexist` was not found. Available algorithms are:', $e->getMessage());
```
inoas: s/type/algo

robertpustulka: Can it be achieved using `@expectedException` and `@expectedExceptionMessage` annotations?

tests/TestCase/Utility/SecurityTest.php
dereuromark: I still got a sniffer to auto fix this globally :)

inoas: Why is it not part of the cakephp tests running on GH :)?

dereuromark: Because no one ported it over yet.

tests/TestCase/Utility/SecurityTest.php
burzum: @robertpustulka can you do a regex or a partial match for the assert of the message?

tests/TestCase/Utility/SecurityTest.php
```diff
@@ -70,6 +71,21 @@ public function testHash()
...
+     *
+     * @return void
+     */
+    public function testInvalidHashTypeException()
+    {
+        try {
+            Security::hash('test', 'doesnotexist', false);
+            $this->fail('Expected \RuntimeException');
+        } catch (RuntimeException $e) {
+            $this->assertTextContains('The hash type `doesnotexist` was not found. Available algorithms are:', $e->getMessage());
```
markstory: You could use `@expectedException` and `@expectedExceptionMessage` here to save some lines of code 🏌️

tests/TestCase/Utility/SecurityTest.php
robertpustulka: @burzum yes, you can:

https://phpunit.de/manual/current/en/appendixes.annotations.html#appendixes.annotations.expectedExceptionMessageRegExp

tests/TestCase/Utility/SecurityTest.php
burzum: @markstory does `@expectedExceptionMessage` it work with changing messages? The message contains a list of available algorithms, depending on the system the test runs on it will be different...

src/Utility/Security.php
```diff
@@ -51,20 +51,29 @@ class Security
...
     {
-        if (empty($type)) {
-            $type = static::$hashType;
+        if (empty($algorithm)) {
+            $algorithm = static::$hashType;
+        }
+        $algorithm = strtolower($algorithm);
+
+        $availableAlgorithms = hash_algos();
+        if (!in_array($algorithm, $availableAlgorithms)) {
```
inoas: I always use true as in_arrays last arg unless required to not to (same as always === unless required to not to)

inoas (approved):

Issue #11272, opened 2017-10-04
3.6: Do not silently load invalid/inexistent tasks


dereuromark:
This is a (multiple allowed):

* [x] bug
* [x] enhancement
* [ ] feature-discussion (RFC)

```php
    public $tasks = [
        'Bake.SuperInvalid',
        ...
    ];
```

Refs https://github.com/cakephp/bake/pull/362
We should at least trigger an error, better yet throw a clear exception if an invalid task (or typo) is loaded.

PR #11286, opened 2017-10-06
#11272 Throw an exception early when invalid task is loaded

burzum:
See #11272

### DIFF
--- a/src/Console/Shell.php
+++ b/src/Console/Shell.php
@@ -198,8 +198,6 @@ class Shell
         if (isset($this->modelClass)) {
             $this->loadModel();
         }
-
-        $this->loadTasks();
     }
 
     /**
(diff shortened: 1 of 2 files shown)


dereuromark (approved):
This sure helps console development! thank you.

src/Console/Shell.php
```diff
@@ -197,6 +198,8 @@ public function __construct(ConsoleIo $io = null)
         if (isset($this->modelClass)) {
             $this->loadModel();
         }
+
+        $this->loadTasks();
```
markstory: Why does this need to be done earlier? Wouldn't doing this in `initialize()` still be early enough?

dereuromark: I guess it would be.

burzum: I haven't realized it's already been called there. Change the code. I've also refactored the added code a little and put it in it's own method to allow easier overriding and reduce method complexity.

markstory:
Nice work @burzum

PR #11290, opened 2017-10-07
3.x APCu Cache Engine Improvements

inoas:
This super-cedes https://github.com/cakephp/cakephp/pull/11279


inoas:
Hmm... warn() will not report the correct file etc... right?

src/Cache/CacheEngine.php
```diff
@@ -276,4 +277,18 @@ protected function _key($key)
...
     }
+
+    /**
+     * Cache Engines may trigger warnings if they encounter failures during operation,
+     * if option warningOnFailures is set to true.
+     *
+     * @param string $message The warning message.
+     * @return void
+     */
+    private function warn($message)
```
ADmad: We don't use privates.

ADmad (changes requested):

src/Cache/Engine/ApcuEngine.php
```diff
@@ -0,0 +1,281 @@
...
+ * CakePHP(tm) : Rapid Development Framework (https://cakephp.org)
+ * Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
+ *
+ * Licensed under The MIT License
+ * For full copyright and license information, please see the LICENSE.txt
+ * Redistributions of files must retain the above copyright notice.
+ *
+ * @copyright     Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
+ * @link          https://cakephp.org CakePHP(tm) Project
+ * @since         1.2.0
```
markstory: Needs to be updated

inoas:
What can/should I do about tests failing because they try to read non-existing apcu keys?
https://travis-ci.org/cakephp/cakephp/jobs/284853635

inoas:
Is this a good solution? https://stackoverflow.com/questions/32884788/phpunit-assert-e-user-warning-is-generated-as-expected-without-ending-a-test

saeideng:
I think you should use `git mv ... `  to moving a file in git

inoas:
For the diffs on ApcuEngine vs ApcEngine you can take a look here: https://github.com/cakephp/cakephp/pull/11290/commits/944d57b4c583fee5af1737bcbc02a223e028ff25

saeideng:
maybe needs `3.next` branch 

`src/Cache/Engine/ApcEngine_CLASS_ALIAS.php`

src/Cache/CacheEngine.php
```diff
@@ -276,4 +276,30 @@ protected function _key($key)
...
+    /**
+     * Cache Engines may trigger warnings if they encounter failures during operation,
+     * if option warningOnFailures is set to true.
+     *
+     * @param string $message The warning message.
+     * @return void
+     */
+    protected function warn($message)
+    {
+        if ($this->config('disableWarningsOnFailures') !== true) {
```
dereuromark: Shouldnt you rather bail early to avoid nesting?

PR #11291, opened 2017-10-07
Fix $previous becoming null only

chinpei215:
Also fixed $code becoming 500 when null given.


ADmad:
Change looks good but test failing.

markstory:
@ADmad that failing test was fixed on master in 06f3a0077206dd91da3a2f17af66ec5fca8ffac1

PR #11292, opened 2017-10-07
[3.next] Use exception chaining

chinpei215:
This pull request is the second step to support exception chaining (#9824).

Most exceptions thrown in catch block will wrap the caught exception. Since our build-in error renderer doesn't support exception chaining, you cannot see any chained exceptions yet. The final step will be enhancement of it.

### DIFF
--- a/src/Cache/Cache.php
+++ b/src/Cache/Cache.php
@@ -181,9 +181,7 @@ class Cache
             }
 
             if ($config['fallback'] === $name) {
-                throw new InvalidArgumentException(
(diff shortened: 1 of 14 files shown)


burzum:
Awesome! I hope this fixes the issues we have sometimes. We have occasionally the case that an exception is reported but when you then comment the cause another one pop up that is the *real* cause.

markstory:
Nice work @chinpei215 !

chinpei215:
Thank you.
@burzum Please wait a little until new error renderer comes 🙇

Issue #8368, opened 2016-02-27
Url validation returns false when emoji is used in url 


dashrat:
This is a (multiple allowed):
- [x] enhancement
- CakePHP Version: I am using 3.0.8 but I checked github for latest implementation in Validation.php and validation implementation is still same. So I think its not updated yet to accept emoji in valid url.
## In controller I am using validator as : 

```
$validator = new Validator();
            $validator
                ->requirePresence('old_url')
                ->add('old_url', 'validFormat', [
                    'rule' => 'url',
                    'message' => 'Url must be valid'
                ]);
            $errors = $validator->errors($this->request->data());
            if (!empty($errors)) {
                echo json_encode(['result' => 1, "status" => 0, "message" => "Please enter a valid url. Ex: http://example.com " ], true);
                exit;
            }
// more code
```

---
### Expected Behavior

http://hashtrax.com/👹 should be a valid url in validator
### Actual Behavior

error is being returned when emoji added in url and try to validate

markstory:
Wouldn't the emoji be URL encoded, and then be able to pass validation?

ndm2:
@markstory Not necessarily, that would only be realiably true for the current request URI. Data retrieved via GET/POST, or generally from external sources, doesn't necessarily have to be encoded.

markstory:
Closing as a pull request is open now.

PR #11293, opened 2017-10-07
Allow Emoji Characters in the URL

AntonNguyen:
Emoji characters are a valid in a URL, so the validator should be updated to allow this edge case to pass.

Fun Emoji/Browser facts
1. Browsers utilize something called `punycoder` to convert something like `🍕` to `xn--vi8h`
2. At the moment, the only ccTLD that supports emojis in the domain name, is `.ws`!

refs #8368

### DIFF
--- a/src/Validation/Validation.php
+++ b/src/Validation/Validation.php
@@ -908,7 +908,9 @@ class Validation
     public static function url($check, $strict = false)
     {
         static::_populateIp();
-        $alpha = '0-9\p{L}\p{N}';
+
+        $emoji = '\x{1F300}-\x{1F6FF}';
+        $alpha = '0-9(\p{L}\p{N}' . $emoji;
         $hex = '(%[0-9a-f]{2})';
         $subDelimiters = preg_quote('/!"$&\'()*+,-.@_:;=~[]', '/');
         $path = '([' . $subDelimiters . $alpha . ']|' . $hex . ')';
--- a/src/Validation/Validation.php
+++ b/src/Validation/Validation.php
@@ -909,7 +909,7 @@ class Validation
     {
         static::_populateIp();
 
-        $emoji = '\x{1F300}-\x{1F6FF}';
+        $emoji = '\x{1F190}-\x{1F9EF}';
         $alpha = '0-9(\p{L}\p{N}' . $emoji;
         $hex = '(%[0-9a-f]{2})';
         $subDelimiters = preg_quote('/!"$&\'()*+,-.@_:;=~[]', '/');
--- a/src/Validation/Validation.php
+++ b/src/Validation/Validation.php
@@ -904,6 +904,7 @@ class Validation
      * @param string $check Value to check
      * @param bool $strict Require URL to be prefixed by a valid scheme (one of http(s)/ftp(s)/file/news/gopher)
      * @return bool Success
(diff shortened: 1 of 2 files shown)


burzum:
Looks good. But could we add a `@link` doc block to some RFC or spec for the emoji stuff? I haven't been aware that there is something like that and also not that you're allowed to use them in URLs.

src/Validation/Validation.php
```diff
@@ -908,7 +908,9 @@ public static function range($check, $lower = null, $upper = null)
     public static function url($check, $strict = false)
     {
         static::_populateIp();
-        $alpha = '0-9\p{L}\p{N}';
+
+        $emoji = '\x{1F300}-\x{1F6FF}';
```
markstory: Should we go higher than U+1F6FF? The [emoji table on wikipedia](https://en.wikipedia.org/wiki/Emoji) has 🧀 at U+1F9C0. The points above 1F6FF are a bit sparse but there are a few good ones up there.

AntonNguyen: Absolutely!

src/Validation/Validation.php
```diff
@@ -904,11 +904,14 @@ public static function range($check, $lower = null, $upper = null)
...
      * @return bool Success
+     * @link https://tools.ietf.org/html/rfc3986
      */
     public static function url($check, $strict = false)
     {
         static::_populateIp();
-        $alpha = '0-9\p{L}\p{N}';
+
+        $emoji = '\x{1F190}-\x{1F9EF}';
+        $alpha = '0-9(\p{L}\p{N}' . $emoji;
```
ADmad: Is the addition of `(` in the regex intentional? It wasn't there earlier.

saeideng:
maybe `urlWithEmoji()`  instead of `url()`

markstory:
@saeideng Why a separate method?

saeideng:
maybe someone unlike the emoji in Url

markstory:
Someone not liking emoji in urls doesn't make those URLs less valid though.

src/Validation/Validation.php
AntonNguyen: Nice catch!

inoas:
It would be great if the validator allowed validation of:
- asciiOnlyUrl
- utf8UrlNoEmoji
- utf8Url

... or similar.

markstory:
@inoas I don't think we need to meet all those use cases in the validators that CakePHP provides. If people need to support a subset of valid URLs they can create their own patterns.

inoas:
Well it may be a valid URL but that doesn't mean it is a good URL and the validator before allowing emojis > the validator after allowing them (also it is not BC technically).

dereuromark:
I am with @markstory on this one:
If the current http specs allow for these URLs, and those are technically valid/correct, then this warrants a fix in a patch release.

If you don't like them, feel free to either use your own validation rule for URLs or simply add a validation rule that sniffs for the '\x{1F190}-\x{1F9EF}' 4 byte range and disallows their use. I don't think this is needed or justified, however, in the majority of use cases.

src/Validation/Validation.php
markstory: I can fix that before merging.

markstory:
Thanks!

Issue #10432, opened 2017-03-18
Add UUID support for binary(16)


bluesuiter:
This is a (multiple allowed):

* [] bug
* [ ] enhancement
* [X ] feature-discussion (RFC)

Hello All,

### Why the binary UUID support is not added to 3.x. 

Even though it is better than char(36) performance wise, why it was dropped out from cakePHP.

Thank You,

dereuromark:
This is a duplicte of https://github.com/cakephp/cakephp/issues/9502

bluesuiter:
Do we have support for binary uuid for current version or not. Because I tried to generate UUID with binary(16) and binary(36) also but it was all in vain.
Although char(36) is working fine.

Thank You,

dereuromark:
Please also see https://github.com/cakephp/cakephp/issues/7013

The main reason, as stated in the first linked ticket, is really just that no one took the time to do it.
Anyone in need and interested in getting this supported can help out and provide a PR or basic ground works here in forms of custom Type classes (see mine linked in the 2nd ticket as a very first basic draft).

markstory:
We haven't added support for binary(16) as there is support fot native uuid types in postgres and the uuid type maps to char(36) in mysql. I don't know how we could preserve schema types if both char(36) and binary(16) are both interpreted as 'uuid' types.

lorenzo:
If you are interested in implementing it, you can do something like this:

https://github.com/cakephp/cakephp/blob/8c94ea345d6083962266efffde7fdeb1e03b13ef/tests/TestCase/Database/ExpressionTypeCastingIntegrationTest.php

The only thing missing in that implementation is doing the bin packing and unpacking in the type class.

inoas:
Nane it type uuidbinary or binaryuuid @markstory?

markstory:
@inoas `binaryuuid` sounds fine. Just needs implementation and tests :smile:

bluesuiter:
@markstory So that means if I use binary(16) as data type with Postgres, I will get UUID with CakePHP 3.4.3 without adding any third party code and special data handling code?

markstory:
@bluesuiter Only after the `binaryuuid` is created. Presently you won't get the behavior you're describing as the new type hasn't been created.

bluesuiter:
@markstory Hello Mark,
Is `binaryuuid` become the part of CakePHP now?

inoas:
I haven't found time so did no one else, so far, it seems.

[...]

PR #11297, opened 2017-10-07
[RFC] Added database support for the binary-uuid type

AntonNguyen:
I haven't worked with binary uuids before, so it'd be great if I could get some extra eyes on this. I was just following the thread in #10432, and decided to take a stab at it. 😄 

refs #10432

### DIFF
--- a/src/Database/Type/BinaryUuidType.php
+++ b/src/Database/Type/BinaryUuidType.php
@@ -67,6 +67,7 @@ class BinaryUuidType extends Type implements TypeInterface
         if (is_string($value)) {
             return $this->convertStringToBinaryUuid($value);
         }
+
         return $value;
     }
 
--- a/src/Database/Type/BinaryUuidType.php
+++ b/src/Database/Type/BinaryUuidType.php
@@ -9,7 +9,7 @@
  *
  * @copyright     Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
  * @link          https://cakephp.org CakePHP(tm) Project
- * @since         3.0.0
+ * @since         3.6.0
  * @license       https://opensource.org/licenses/mit-license.php MIT License
  */
 namespace Cake\Database\Type;
--- a/src/Database/Type/BinaryUuidType.php
+++ b/src/Database/Type/BinaryUuidType.php
@@ -64,6 +64,9 @@ class BinaryUuidType extends Type implements TypeInterface
      */
     public function toDatabase($value, Driver $driver)
     {
+        if (is_string($value)) {
+            return $this->convertStringToBinaryUuid($value);
+        }
         return $value;
     }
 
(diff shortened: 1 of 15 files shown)


src/Database/Type/BinaryUuidType.php
```diff
@@ -0,0 +1,165 @@
...
+ * CakePHP(tm) : Rapid Development Framework (https://cakephp.org)
+ * Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
+ *
+ * Licensed under The MIT License
+ * For full copyright and license information, please see the LICENSE.txt
+ * Redistributions of files must retain the above copyright notice.
+ *
+ * @copyright     Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
+ * @link          https://cakephp.org CakePHP(tm) Project
+ * @since         3.0.0
```
markstory: Should be 3.6.0

tests/TestCase/Database/Type/BinaryUuidType.php
```diff
@@ -0,0 +1,113 @@
...
+     */
+    public function testToDatabase()
+    {
+        $value = 'some data';
+        $result = $this->type->toDatabase($value, $this->driver);
+        $this->assertEquals($value, $result);
+
+        $fh = fopen(__FILE__, 'r');
+        $result = $this->type->toDatabase($fh, $this->driver);
+        $this->assertSame($fh, $result);
```
markstory: Should there be a test for a string uuid coming back as bytes?

markstory (commented):
Thanks for taking this on. A few people should be quite happy🎉

markstory:
Updated with schema reflection for each db driver and integrated the new type into the Type mappings. In MySQL, SQLServer and SQLite `BINARY(16)` will now map to `binaryuuid`. Postgres does not support binary columns with lengths so binaryuuid maps to the `UUID` datatype there.

markstory:
Thanks @AntonNguyen :clap:

dereuromark:
@AntonNguyen Do you agree on the following simpifications and fixes? https://github.com/dereuromark/cakephp-shim/pull/11 We could make a follow up PR for 3.6 then.

dereuromark:
Also, when adding invalid UUIDs like `'test'` it will save them as `0e000000000000000000000000000000
` instead of failing, and also a lot of

    Warning (2): pack(): Type H: illegal hex digit t [ROOT/.../BinaryUuidType.php, ...]

etc
But maybe this should be part of the validation/patching part and is out of scope of the type class then?

markstory:
> But maybe this should be part of the validation/patching part and is out of scope of the type class then?

Yes that seems like a validation concern to me.

lorenzo:
I just saw this... would have preferred a sorted binary uuid implementation, like the one we have in some of the tests.

I guess that could be another custom type :)

dereuromark:
Do you mean https://github.com/alizain/ulid or https://github.com/mbbender/doctrine-ordered-uuid-generator/blob/master/src/ORM/Id/OrderedGuidGenerator.php @lorenzo  ?

markstory:
@lorenzo We can still change this implementation as it is in the unreleased 3.next branch.

dereuromark:
Is there any interest to finish this for 3.x or 4.x?

PR #11298, opened 2017-10-08
Revert partial of #11270 (Fix Credit card number pattern(JCB) is wrong)

tenkoma:
In the previous pull request (#11270), I made a proposal to fix the pattern of JCB credit card with reference to the code of another framework, but that was not good.
In this pull request, rewind from the previous pull request except for typo fix.


chinpei215 (approved):
Looks good to me.

Issue #11277, opened 2017-10-05
RFC : EntityRouting


Kareylo:
This is a (multiple allowed):

* [ ] bug
* [x] enhancement
* [x] feature-discussion (RFC)

As I tried latest releases of Ruby on Rails, I found a really usefull feature with routing : EntityRouting. The main goal is to have the ability to name routes - as we can do with `'_name' => 'myrouteName'` in Route options and use an entity to put on the back side all the parameters we need, and, later, we can change what we need in the route without having the need to change all routes in our views.

For example, we've got this, right now :

```php
// Routing
$routes->connect('/view/:id', ['controller' => 'Posts', 'action' => 'view'], ['pass' => ['id'], 'id' => '\d+']);

// In View
$this->Html->link('View Post', ['controller' => 'Posts', 'action' => 'view', 'id' => $post->id]);
```

If in later days we change the route like this :

```php
$routes->connect('/view/:slug-:id', ['controller' => 'Posts', 'action' => 'view'], ['pass' => ['slug', 'id'], 'slug' => '[a-z\-]+', 'id' => '\d+']);
```
We'll need to change all the routes in the app

### How can we enhance this ?

The main goal is to forgot what we need to pass to the Router::url with the entity directly with something like the following example without having to really change the route side, just the need to give a name to this route :
```php
// Routing
$routes->connect('/view/:slug-:id', ['controller' => 'Posts', 'action' => 'view'], ['pass' => ['slug', 'id'], 'slug' => '[a-z\-]+', 'id' => '\d+', '_name' => 'postsView']);

// In View
$this->Html->link('View Post', ['_name' => 'postsView', '_entity' => $post]);
```

So, I searched for someone who have writing something like that and found this [Blog Post](http://ad7six.com/blog/2014/09/16/cakephp-entity-routing/) 

I know it's not really a needed feature, but still a good one.

Have you got any idea on this feature ?
Maybe could be improved with #10643 ?

saeideng:
>[Blog Post](http://ad7six.com/blog/2014/09/16/cakephp-entity-routing/)

adding `'routeClass' => 'EntityRoute'` is good idea

markstory:
Seems like a neat idea with a pretty simple solution from @AD7six  blog.

Kareylo:
I think just giving the entity should be really nice too, but it's more work on routing.

```php
$this->Html->link('View Post', ['_name' => 'postsView', $post]);
```

[...]

PR #11300, opened 2017-10-08
#11277 Adding an entity route class

burzum:
See #11277


src/Routing/Route/EntityRoute.php
```diff
@@ -0,0 +1,69 @@
...
+     *   Contains information such as the current host, scheme, port, and base
+     *   directory.
+     * @return bool|string Either false or a string URL.
+     */
+    public function match(array $url, array $context = [])
+    {
+        if (isset($url['_entity'])) {
+            $this->_checkEntity($url['_entity']);
+
+            $entity = $url['_entity'];
```
saeideng: maybe 
```php
$entity = $url['_entity'];
$this->_checkEntity($entity);
```

src/Routing/Route/EntityRoute.php
```diff
@@ -0,0 +1,69 @@
...
+ * @since         3.6.0
+ * @license       https://opensource.org/licenses/mit-license.php MIT License
+ */
+namespace Cake\Routing\Route;
+
+use Cake\Datasource\EntityInterface;
+use RuntimeException;
+
+/**
+ * Matches entities to routes
```
markstory: We might want to explain when/why someone would want to use this class.

src/Routing/Route/EntityRoute.php
```diff
@@ -0,0 +1,69 @@
...
+    /**
+     * Checks that we really deal with an entity object
+     *
+     * @throws \RuntimeException
+     * @param mixed $entity Entity value from the URL options
+     * @return void
+     */
+    protected function _checkEntity($entity)
+    {
+        if (!$entity instanceof EntityInterface) {
```
markstory: I wonder if we should generalize this further and accept any array or ArrayAccess instance?

saeideng: Yes we should accept array type

burzum: Done.

src/Routing/Route/EntityRoute.php
```diff
@@ -0,0 +1,74 @@
...
+        }
+
+        return parent::match($url, $context);
+    }
+
+    /**
+     * Checks that we really deal with an entity object
+     *
+     * @throws \RuntimeException
+     * @param mixed $entity Entity value from the URL options
```
dereuromark: \ArrayAccess|array ?

src/Routing/Route/EntityRoute.php
burzum: Done

src/Routing/Route/EntityRoute.php
burzum: Hm why? It technically takes any, the method checks for the correct type and object.

dereuromark: That is not how we ever did things :) We typehint the expected possible input, not the worst possible, otherwise we would have mixed on every protected method..

ADmad: Well the docblock is supposed to document valid types so it should be `\ArrayAccess|array` as @dereuromark mention.

src/Routing/Route/EntityRoute.php
```diff
@@ -0,0 +1,74 @@
...
+     * @return void
+     */
+    protected function _checkEntity($entity)
+    {
+        if (!$entity instanceof ArrayAccess && !is_array($entity)) {
+            throw new RuntimeException(sprintf(
+                'Route `%s` expects the URL option `_entity` to be `%s`, but `%s` passed.',
```
ADmad: This should now be changed to `'Route `%s` expects the URL option `_entity` to be an array or object implementing \ArrayAccess, but `%s` passed.' and method args updated accordingly.

src/Routing/Route/EntityRoute.php
```diff
@@ -0,0 +1,74 @@
...
+ *
+ * @copyright     Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
+ * @link          https://cakephp.org CakePHP(tm) Project
+ * @since         3.6.0
+ * @license       https://opensource.org/licenses/mit-license.php MIT License
+ */
+namespace Cake\Routing\Route;
+
+use ArrayAccess;
+use Cake\Datasource\EntityInterface;
```
ADmad: This is no longer required.

src/Routing/Route/EntityRoute.php
```diff
@@ -0,0 +1,72 @@
...
+     * @param \ArrayAccess|array $entity Entity value from the URL options
+     * @return void
+     */
+    protected function _checkEntity($entity)
+    {
+        if (!$entity instanceof ArrayAccess && !is_array($entity)) {
+            throw new RuntimeException(sprintf(
+                'Route `%s` expects the URL option `_entity` to be an array or object implementing \ArrayAccess, but `%s` passed.',
+                $this->template,
+                is_object($entity) ? get_class($entity) : gettype($entity)
```
ADmad: This ternary is now used quite often throughout our code base. Perhaps we should wrap it in a helper function. (Can be done in a separate PR, just wanted to point it out.)

dereuromark: This is so low level that it is probably not worth wrapping.

burzum: I found six use cases and there will be probably more in the future if we keep up with writing meaningful and informative exception messages.

markstory: Nice error message 👏

src/Routing/Route/EntityRoute.php
```diff
@@ -0,0 +1,72 @@
...
+    public function match(array $url, array $context = [])
+    {
+        if (isset($url['_entity'])) {
+            $this->_checkEntity($url['_entity']);
+
+            $entity = $url['_entity'];
+            preg_match_all('@:(\w+)@', $this->template, $matches);
+
+            foreach ($matches[1] as $field) {
+                $url[$field] = $entity[$field];
```
Kareylo: What if we want to pass a association key like `category_id`in route ?
Maybe something like this : 
```php
$url[$field] = $url['_entity'][\Cake\Utility\Inflector::delimit($field)];
```

ADmad: `Inflector::delimit()` converts a CamelCased string to underscored, which field name already are by convention. So why use `delimit()`?

Kareylo: Route could be like this :
```php
$routes->connect('/posts/:categoryId/:slug-:id', ['controller' => 'Posts', 'action' => 'category']);
```
If not use `delimit()`, we need to catch the underscore in the regex

Kareylo: Oh. My bad, `\w` catch `_` too.

burzum:
@markstory is there something left to do in this ticket?

src/Routing/Route/EntityRoute.php
markstory: Could we use the new `getTypeName()` function here?

tests/TestCase/Routing/Route/EntityRouteTest.php
```diff
@@ -0,0 +1,108 @@
...
+        $entity = new Article([
+            'category_id' => 2,
+            'slug' => 'article-slug'
+        ]);
+
+        $route = $route = new EntityRoute(
+            '/articles/:category_id/:slug',
+            [
+                '_name' => 'articlesView',
+                '_entity' => $entity,
```
markstory: Why is entity being passed in the constructor/route defaults?

tests/TestCase/Routing/Route/EntityRouteTest.php
```diff
@@ -0,0 +1,108 @@
...
+                'controller' => 'articles',
+                'action' => 'view'
+            ]
+        );
+
+        $result = $route->match([
+            '_entity' => $entity,
+            '_name' => 'articlesView',
+            'controller' => 'articles',
+            'action' => 'view'
```
markstory: You don't need controller/action & _name.

tests/TestCase/Routing/Route/EntityRouteTest.php
```diff
@@ -0,0 +1,108 @@
...
+        $entity = [
+            'category_id' => 2,
+            'slug' => 'article-slug'
+        ];
+
+        $route = $route = new EntityRoute(
+            '/articles/:category_id/:slug',
+            [
+                '_name' => 'articlesView',
+                '_entity' => $entity,
```
markstory: I don't think _name and _entity need to be here.

tests/TestCase/Routing/Route/EntityRouteTest.php
```diff
@@ -71,18 +66,14 @@ public function testMatchingArray()
             '/articles/:category_id/:slug',
             [
                 '_name' => 'articlesView',
-                '_entity' => $entity,
-                'controller' => 'articles',
-                'action' => 'view'
+                '_entity' => $entity
```
markstory: The route constructor should get controller/action but no entity key.

PR #11302, opened 2017-10-08
Fixed SqlserverSchema return colum length double the actual size

### LINKED ISSUE #11295: SqlserverSchema return colum length double the actual size
This is a (multiple allowed):

* [x] bug
* [ ] enhancement

* CakePHP Version: 3.5.3.
* Platform and Target: Windows, MS SQL Server 2012.

### What you did
I've created a fixture with a string field having a limit of 50 characters. 
```php
// extra columns were truncated
class BakeArticlesFixture extends TestFixture
{
    public $fields = [
        'title' => ['type' => 'string', 'length' => 50, 'null' => false],
    ];
}
```

### What happened
During unit-testing, the table's object reported, that the "title" field has a length of 100 characters.
```php
$schema = $table->getSchema();
```

The length of the field to be reported as 50 characters.

[AppVeyor's build log](https://ci.appveyor.com/project/cakephp/bake/build/1.0.594#L285)
Affects cakephp/bake/pull/335

dmromanov:
I don't have a real MS SQL instance, so I've tested it in SQL Fiddle http://sqlfiddle.com/#!6/dffef/1 . Seems to be working.

List of data types supporting Unicode in MS SQL was taken from the following page: https://docs.microsoft.com/en-us/sql/t-sql/data-types/data-types-transact-sql#unicode-character-strings 

Closes  #11295


markstory (approved):
Looks good.

dmromanov:
I wonder if it was better to do the division operation in _convertColumn() method?

markstory:
> I wonder if it was better to do the division operation in _convertColumn() method?

I think doing it in the database server is fine. Pushing simple operations down to SQLServer lets the code be simpler/faster as I would assume sqlserver will be faster than PHP.

markstory:
The [appveyor build](https://ci.appveyor.com/project/cakephp/cakephp/build/1.0.8727) is showing something funny going on with column type mappings.

dmromanov:
I have difficulties debugging this, so it will take some time.

dmromanov:
Thanks!

PR #11304, opened 2017-10-08
Add abort() to Command.

markstory:
Messages can be handled in the command via the `$io` object. I thought it was a bit silly to pass in ConsoleIo so that a message could be printed using it.

Part of #11137

### DIFF
--- a/src/Console/Command.php
+++ b/src/Console/Command.php
@@ -15,6 +15,7 @@
 namespace Cake\Console;
 
 use Cake\Console\Exception\ConsoleException;
+use Cake\Console\Exception\StopException;
 use Cake\Datasource\ModelAwareTrait;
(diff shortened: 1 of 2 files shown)


src/Console/Command.php
```diff
@@ -233,4 +234,16 @@ public function execute(Arguments $args, ConsoleIo $io)
     {
         return null;
     }
+
+    /**
+     * Halt the the current process with a StopException.
+     *
+     * @param int $code The exit code to use.
+     * @throws \Cake\Console\Exception\ConsoleException
```
robertpustulka: `@throws \Cake\Console\Exception\StopException`

markstory: 😊

PR #11305, opened 2017-10-08
Adding getTypeName() to functions.php

burzum:
Suggested by @ADmad [here](https://github.com/cakephp/cakephp/pull/11300#discussion_r143363472). We have right now already 5 cases of this in our code and there will be probably more in the future if we stick to the progress on improving exception messages with meaningful information.

### DIFF
--- a/src/Core/functions.php
+++ b/src/Core/functions.php
@@ -288,7 +288,7 @@ if (!function_exists('getVarType')) {
      * @param mixed $var Variable to check
      * @return string Returns the class name or variable type
      */
-    function getVarType($var)
+    function getTypeName($var)
     {
         return is_object($var) ? get_class($var) : gettype($var);
     }
--- a/src/Core/functions.php
+++ b/src/Core/functions.php
@@ -281,7 +281,7 @@ if (!function_exists('deprecationWarning')) {
     }
 }
 
-if (!function_exists('getVarType')) {
+if (!function_exists('getTypeName')) {
     /**
      * Returns the objects class or var type of it's not an object
      *
--- a/src/Core/functions.php
+++ b/src/Core/functions.php
@@ -280,3 +280,16 @@ if (!function_exists('deprecationWarning')) {
         trigger_error($message, E_USER_DEPRECATED);
     }
 }
+
+if (!function_exists('getVarType')) {
+    /**
+     * Returns the objects class or var type of it's not an object
+     *
+     * @param mixed $var Variable to check
+     * @return string Returns the class name or variable type
+     */
+    function getVarType($var)
+    {
+        return is_object($var) ? get_class($var) : gettype($var);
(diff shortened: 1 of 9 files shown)


ADmad:
Could be useful in app/plugin code too :)

src/basics.php
```diff
@@ -155,3 +155,16 @@ function loadPHPUnitAliases()
         require_once dirname(__DIR__) . DS . 'tests' . DS . 'phpunit_aliases.php';
     }
 }
+
+if (!function_exists('get_var_type')) {
```
dereuromark: Since when are we using underscore as convention for method names? getVarType() would be the consistent naming IMO.

dereuromark: The "var" sounds odd here... getTypeOrClassName() maybe?

dereuromark:
I would definitely not abbreviate the "var" => "variable".
But I am sure it is not even needed if we can come up with a better name.

inoas:
just getType... tbh.

inoas:
http://php.net/manual/de/function.gettype.php#104224 do we cover all of these? Including null?

dereuromark:
@ionas Isn't this used inside the Debugger class? And shouldn't this then rather be bound to that scope (as class method `Debugger::getType()`), as this is mainly a debugging convenience method?

dereuromark:
You also have to consider if all subsplits will still work with this method being added to the main functions file.
So far those might have worked without it, and the dependencies should be double-checked.

src/basics.php
```diff
@@ -155,3 +155,16 @@ function loadPHPUnitAliases()
         require_once dirname(__DIR__) . DS . 'tests' . DS . 'phpunit_aliases.php';
     }
 }
+
+if (!function_exists('getVarType')) {
```
markstory: If we intend to use this in classes that end up in sub split packages,this function should be in src/Core/functions.php instead.

markstory (commented):
I like the idea and implementation, but the function will likely need to be moved.

ADmad:
@markstory I have moved the function to `functions.php`.

ADmad (approved):

dereuromark:
The "var" name thing is still wrong IMO. We never add var to anything, and of course the input is usually the object the method operates on.
Best to remove it and stick to `getType()` or `getTypeOrClassName()` then only.

ADmad:
getTypeName() works for me too.

src/Core/functions.php
```diff
@@ -280,3 +280,16 @@ function deprecationWarning($message, $stackFrame = 2)
         trigger_error($message, E_USER_DEPRECATED);
     }
 }
+
+if (!function_exists('getVarType')) {
```
markstory: This needs to reflect the rename as well.

PR #11308, opened 2017-10-09
Remove unneeded "use" statements.

ADmad:


dereuromark:
What sniff did you run?

ADmad:
https://github.com/slevomat/coding-standard/blob/master/SlevomatCodingStandard/Sniffs/Namespaces/UnusedUsesSniff.php  :)

dereuromark:
Ah, I thought maybe https://github.com/spryker/code-sniffer/blob/master/Spryker/Sniffs/Namespaces/UnusedUseStatementSniff.php
If that one is working fine - and maybe even better than my linked one - then maybe we can port it into the core?

ADmad:
Would be great if you could port it to CakePHP standard.

Issue #11251, opened 2017-09-28
Updating from 3.5.2 => 3.5.3 triggers error with FileEngine.php(384): mkdir()


makallio85:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.5.3
* Platform and Target: Windows 7, PHP 7.0.0

### What you did
Updated from 3.5.2 to 3.5.3

### What happened
Following error was triggered

```
2017-09-28 15:30:24 Error: [ErrorException] mkdir() [<a href='https://secure.php.net/function.mkdir'>function.mkdir</a>]: File exists in C:\xampp\htdocs\vuokravalvonta\vendor\cakephp\cakephp\src\Cache\Engine\FileEngine.php:384.
Request URL: /rental-agreements/add/2
Referer URL: http://localhost/vuokravalvonta/rental-properties/view/2
Stack Trace:
#0 [internal function]: App\Error\AppErrorHandler->handleError(2, 'mkdir() [<a hre...', 'C:\\xampp\\htdocs...', 384, Array)
#1 C:\xampp\htdocs\vuokravalvonta\vendor\cakephp\cakephp\src\Cache\Engine\FileEngine.php(384): mkdir('C:\\xampp\\htdocs...', 509, true)
#2 C:\xampp\htdocs\vuokravalvonta\vendor\cakephp\cakephp\src\Cache\Engine\FileEngine.php(184): Cake\Cache\Engine\FileEngine->_setKey('myapp_cake_core...')
#3 C:\xampp\htdocs\vuokravalvonta\vendor\cakephp\debug_kit\src\Cache\Engine\DebugEngine.php(149): Cake\Cache\Engine\FileEngine->read('myapp_cake_core...')
#4 C:\xampp\htdocs\vuokravalvonta\vendor\cakephp\cakephp\src\I18n\TranslatorRegistry.php(148): DebugKit\Cache\Engine\DebugEngine->read('translations.de...')
#5 C:\xampp\htdocs\vuokravalvonta\vendor\cakephp\cakephp\src\I18n\I18n.php(204): Cake\I18n\TranslatorRegistry->get('debug_kit')
#6 C:\xampp\htdocs\vuokravalvonta\vendor\cakephp\cakephp\src\I18n\functions.php(88): Cake\I18n\I18n::getTranslator('debug_kit')
[...]

PR #11310, opened 2017-10-09
Error supressing is causing problems with asset files inside plugin

makallio85:
Fixed bug with missing file inside plugin

This relates to this issue https://github.com/cakephp/cakephp/issues/11251 Same environment.

First this was faced with dynamic asset files when using HtmlHelper->script(), but I testsed it and seems like it happens always if file does not exists.

### DIFF
--- a/src/View/Helper/UrlHelper.php
+++ b/src/View/Helper/UrlHelper.php
@@ -210,7 +210,7 @@ class UrlHelper extends Helper
             $webrootPath = WWW_ROOT . str_replace('/', DIRECTORY_SEPARATOR, $filepath);
             if (file_exists($webrootPath)) {
                 //@codingStandardsIgnoreStart
-                return $path . '?' . @filemtime($webrootPath);
+                return $path . '?' . filemtime($webrootPath);
                 //@codingStandardsIgnoreEnd
             }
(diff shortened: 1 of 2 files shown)


makallio85:
I was just wondering, is the correct fix to remove ? from return value of assetTimestamp() method with missing files or keep it?

src/View/Helper/UrlHelper.php
```diff
@@ -218,9 +218,11 @@ public function assetTimestamp($path)
             if (Plugin::loaded($plugin)) {
                 unset($segments[0]);
                 $pluginPath = Plugin::path($plugin) . 'webroot' . DIRECTORY_SEPARATOR . implode(DIRECTORY_SEPARATOR, $segments);
-                //@codingStandardsIgnoreStart
-                return $path . '?' . @filemtime($pluginPath);
-                //@codingStandardsIgnoreEnd
+                if (file_exists($pluginPath)) {
+                    //@codingStandardsIgnoreStart
+                    return $path . '?' . @filemtime($pluginPath);
```
markstory: If the file exists, you don't need to suppress errors when getting its modified time.

markstory (changes requested):

markstory (approved):

PR #11313, opened 2017-10-09
3.next ApcuEngine: More atomic & verbose on failures

inoas:
- Supercedes Ref: https://github.com/cakephp/cakephp/pull/11290
- Rebased on 3.next as requested here: https://github.com/cakephp/cakephp/pull/11290#issuecomment-335042913
- moved `warn()` method to a basic function for re-use (could also be used by deprecationWarning()?)


src/Cache/CacheEngine.php
```diff
@@ -276,4 +277,20 @@ protected function _key($key)
...
+    /**
+     * Cache Engines may trigger warnings if they encounter failures during operation,
+     * if option warnOnFailures is set to true.
+     *
+     * @param string $message The warning message.
+     * @return void
+     */
+    protected function warn($message)
+    {
+        if (in_array($this->config('warnOnFailures'), [null, false], true)) {
```
markstory: Can you use `getConfig()`? `config()` is deprecated.

markstory: This could also be
```php
if ($this->getConfig('warnOnFailure') !== true) {
  return;
}
```

src/Cache/Engine/ApcEngine_CLASS_ALIAS_RENAME_BEFORE_MERGE.php
```diff
@@ -0,0 +1,18 @@
...
+ *
+ * @copyright     Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
+ * @link          https://cakephp.org CakePHP(tm) Project
+ * @since         1.2.0
+ * @license       https://opensource.org/licenses/mit-license.php MIT License
+ */
+namespace Cake\Cache\Engine;
+
+// @deprecated Add backwards compat alias.
+class_alias('Cake\Cache\Engine\ApcuEngine', 'Cake\Cache\Engine\ApcEngine');
```
markstory: I don't think this file is supposed to still be here.

src/basics.php
```diff
@@ -62,6 +62,31 @@ function debug($var, $showHtml = null, $showFrom = true)
...
 }
 
+if (!function_exists('warn')) {
+    /**
+     * Throws a warning.
+     *
+     * @param string $message The warning message.
+     * @return void
+     */
+    function warn($message)
```
markstory: This function would need to be in `src/Core/functions.php` as the cache package is subsplit out, and cakephp/core does not include this file.

src/Cache/Engine/ApcEngine_CLASS_ALIAS_RENAME_BEFORE_MERGE.php
inoas: Sure? If peeos specify ApcEngine then this alias will kick in?

Should be named ‘ApcEngine.php’ but then the guthub diff of ApcEngine>ApcuEngine ain’t work for review purposes.

inoas: Should this be 'apcu' so it loads ApcuEngine?
https://github.com/cakephp/cakephp/pull/11313/commits/ac0fb72059696420c2231bd66bfd33094c128ad4#diff-4d9f8cefbaed0af0a66e03a6206f6e3bR79

ADmad: @inoas Did you notice the filename? `src/Cache/Engine/ApcEngine_CLASS_ALIAS_RENAME_BEFORE_MERGE.php`

inoas: Yes I did @saeideng really wanted to keep the diff view on github alive. The moment I rename that file to `ApcEngine.php` that moment the github diff rendering from ApcEngine.php <-> ApcuEngine.php will break and it will show ApcuEngine.php as a new file.

The file exists because I think we should add a class alias until 4.x hits in case people used ApcEngine directly for BC reasons.

ADmad: The rename was unnecessary, brings no benefit. We can rename it in 4.0.

inoas: Well I asked before...
https://github.com/cakephp/cakephp/pull/11279#discussion_r143287747
But sure I can rename back everything.
I am not sure that it hurts in any way now.

src/Cache/Engine/ApcuEngine.php
```diff
@@ -77,18 +76,29 @@ public function write($key, $value)
...
     public function read($key)
     {
         $key = $this->_key($key);
 
         $time = time();
-        $cachetime = (int)apcu_fetch($key . '_expires');
+        $success = false;
+        $cachetime = (int)apcu_fetch($key . '_expires', $success);
+        if ($success === false) {
+            $this->warn(sprintf('Failed to fetch key "%s" from APCu cache.', $key . '_expires'));
```
ADmad: Warning on read failure don't make sense to me. Cache misses are bound to occur, they are not an error case.

hmic: 👍

inoas: I can easily change this. The idea is that you know when your Cache does not work for whatever reason. Is there a better way?

src/Cache/Engine/ApcEngine_CLASS_ALIAS_RENAME_BEFORE_MERGE.php
ADmad: > The moment I rename that file to ApcEngine.php that moment the github diff rendering from ApcEngine.php <-> ApcuEngine.php will break and it will show ApcuEngine.php as a new file.

That's expected, can't do anything about it. We are definitely not going to keep filename `ApcEngine_CLASS_ALIAS_RENAME_BEFORE_MERGE.php`.

src/Cache/Engine/ApcuEngine.php
hmic: Monitor your cache and application performance to know it is working.
For the app using the cache, this is bullshit!
You don't need no warning on any of these operations.

src/Cache/CacheEngine.php
```diff
@@ -45,7 +45,8 @@
         'duration' => 3600,
         'groups' => [],
         'prefix' => 'cake_',
-        'probability' => 100
+        'probability' => 100,
+        'warnOnFailures' => true,
```
ADmad: IMO this should be false by default. For e.g. most people won't care about failures when calling `clear()`.

src/Cache/Engine/ApcEngine_CLASS_ALIAS_RENAME_BEFORE_MERGE.php
inoas: Obviously I am going to change that class name before merge :trollface: - the rename is there just to make reviewing better!

src/Cache/CacheEngine.php
inoas: For now I opted to keep this on by default but I removed almost all warnings, except those where writes cache fail which is not supposed to happen, right?

inoas:
I think the only remaining error on travis is unrelated, right?

src/Cache/Engine/ApcEngine_CLASS_ALIAS_RENAME_BEFORE_MERGE.php
saeideng: Inoas
I mean about keeping old file history into new file
Not  old file seperatly
So remove certainly 
`e_CLASS_ALIAS_RENAME_BEFORE_MERGE`

src/Cache/Engine/ApcuEngine.php
```diff
@@ -181,9 +202,9 @@ public function add($key, $value)
         if ($duration) {
             $expires = time() + $duration;
         }
-        apcu_add($key . '_expires', $expires, $duration);
 
-        return apcu_add($key, $value, $duration);
+        return apcu_add($key, $value, $duration) === true
```
ADmad: Why not use array syntax here too as you have done for `apc_store()`?

src/Cache/Engine/ApcuEngine.php
markstory: Reads and deletes can and will fail as data expires. Writes failing is more alarming though.

src/Cache/Engine/ApcuEngine.php
inoas: * Slipped through that _add is actually a write
* Didn't know it had support, for that, too (only discovered myself)
* Will fix

src/Cache/Engine/ApcEngine_CLASS_ALIAS_RENAME_BEFORE_MERGE.php
inoas: So and then we don't have an alias file at all :)?
BC?
Before merge I would rename this file to ApcEngine.php.
* [ ] Good?
* [ ] Bad?

src/Cache/Engine/ApcuEngine.php
inoas: After digging a bit into this I think we should not raise warnings for failings of `apc_add`.
`apc_add` is meant to add-or-ignore. If it cannot because the key already exists, the array signature form of `apcu_add` will return those keys it could not insert because they already exist as failures.

That is probably not something we want to have appear in logs anywhere, right?

src/Cache/Engine/ApcEngine_CLASS_ALIAS_RENAME_BEFORE_MERGE.php
saeideng: >So and then we don't have an alias file at all :)?

no

>BC?

no related to BC :)

> Before merge I would rename this file to ApcEngine.php.

rename now
tracing change is good for new file not for old file that you remove all lines

saeideng: GITHUB show new file correctly

inoas: Okay you don't believe me. I will rename the file and the diff-view on github will not work anymore :)

saeideng: I now understand your mean :)

inoas:
The failing travis test seems unrelated to this PR.
Anything else to do?

ADmad:
Squash the commits :)

ADmad:
The CS failure has been taken care of in #11318

saeideng:
@ADmad 
CS fixed https://github.com/cakephp/cakephp/pull/11241
you can fix that again in this PR and #11318

ADmad:
@saeideng We don't need to fix again what's already fixed on master :)

src/Core/functions.php
```diff
@@ -249,6 +249,31 @@ function env($key, $default = null)
...
 }
 
+if (!function_exists('warn')) {
+    /**
+     * Throws a warning.
+     *
+     * @param string $message The warning message.
+     * @return void
+     */
+    function warn($message)
```
inoas: Following should I change this to `warning()`?
https://github.com/cakephp/cakephp/pull/11319

markstory: That would be great.

inoas:
codecov issues seem not related: https://codecov.io/gh/cakephp/cakephp/compare/53014489d8a9381887090d81bb38961acb990dc0...17960b6a5e4d98c6f81f709178e9fdfe5dc5e681/changes

https://codecov.io/gh/cakephp/cakephp/compare/53014489d8a9381887090d81bb38961acb990dc0...17960b6a5e4d98c6f81f709178e9fdfe5dc5e681/changes

inoas:
Travis issue seems not related to this PR: https://travis-ci.org/cakephp/cakephp/builds/288059789?utm_source=github_status&utm_medium=notification

src/Core/functions.php
```diff
@@ -249,6 +249,31 @@ function env($key, $default = null)
 
 }
 
+if (!function_exists('warning')) {
```
ravage84: ``warning()`` seems quite general. There is a good chance of a name clash with user land code. How about ``throwWarning()`` instead?

hmic: Why does this need to exist in the global space anyways?
Additionally, "throw" is not quite what it is supposed to do, logWarning or maybe processWarning or even cacheWarning, as it could live in the cache namespace anyways and be included - maybe a trait - from the cache engines only, instead of poluting global space :P

inoas: http://php.net/manual/en/function.trigger-error.php
- `triggerWarning()`
- `generateWarning()`

I can for sure move this to `class CacheEngine` if any of the core members prefer that as well.

ravage84: Yeah, sorry, I agree with @hmic on the name. I just converted what the doc block stated into its name. 😼 
``triggerWarning()`` sounds good to me.

inoas:
- So, should `triggerWarning()` be moved out of global context into CacheEngine or a trait?
- Anything else to do?

Issue #11289, opened 2017-10-06
Multi Tree Deletes Nodes outside the Scope


jonmc1:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.5.3

### What you did
In a Multi Tree, multiple tree structures in the same db table, delete a tree node that is not a leaf

In a Table:
$this->addBehavior('Tree');
...
$this->behaviors()->Tree->config('scope', ['field_name' => $fieldValue]);
$node = $this->get($id);
$this->delete($node);

### What happened
Nodes were deleted, by 'lft' field, for all trees in the db table

### What you expected to happen
Only delete nodes from the tree that 'scope' is set to

markstory:
Closing as a pull request is open now.

PR #11314, opened 2017-10-09
Fixed issue Multi-Tree Deleting Nodes outside the Scope

jonmc1:
GitHub Issue #11289

### DIFF
--- a/src/ORM/Behavior/TreeBehavior.php
+++ b/src/ORM/Behavior/TreeBehavior.php
@@ -224,12 +224,16 @@ class TreeBehavior extends Behavior
         $diff = $right - $left + 1;
 
         if ($diff > 2) {
-            $this->_table->deleteAll(function ($exp) use ($config, $left, $right) {
-                /* @var \Cake\Database\Expression\QueryExpression $exp */
-                return $exp
-                    ->gte($config['leftField'], $left + 1)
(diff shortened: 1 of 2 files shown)


src/ORM/Behavior/TreeBehavior.php
```diff
@@ -224,12 +224,16 @@ public function beforeDelete(Event $event, EntityInterface $entity)
...
+            /* @var \Cake\Database\Expression\QueryExpression $expression */
+            $expression = $this->_scope($this->_table->find())
+                ->where(function ($exp) use ($config, $left, $right) {
+                    return $exp
+                        ->gte($config['leftField'], $left + 1)
+                        ->lte($config['leftField'], $right - 1);
+                })
+                ->clause('where');
+
+            $this->_table->deleteAll($expression);
```
markstory: Instead of picking off the where clause you might want to use a subquery. The default find method _could_ add conditions from associations which will fail with this approach.

markstory: I pushed up a commit with the changes I was proposing and a test case for this issue.

ADmad (approved):

ADmad (changes requested):
Tests failing

markstory:
Oh MySQL. You so silly. I'll get that fixed up.

PR #11317, opened 2017-10-10
Update deprecated method calls in tests/bootsrap.php

o0h:
updated deprecated method `config()` call to `setConfig()` for Cache, ConnectionManager, Log classes.


markstory:
Thank you!

PR #11319, opened 2017-10-12
3.next - ConsoleIo helper methods

markstory:
These methods will allow Commands to have the same output interfaces as shells have today. One change is that `warn` has been renamed to `warning`. This removes a short form that to remember, and makes output consistent with logging which also uses `warning`.

Refs #11137


ravage84 (approved):
👍

PR #11328, opened 2017-10-14
Use utf8mb4 instead uf utf8mb3 (= utf8)

inoas:
- [x] Enhancement
- [x] Bug (inserting valid utf8 such as 😃 will break your data using mysql/cakephp and default utf8 settings set by the driver and recommended app.php)

Refs:
* issue with more info: https://github.com/cakephp/cakephp/issues/11324#issuecomment-336672560
* related app skeleton PR https://github.com/cakephp/app/pull/554

Issue #11324, opened 2017-10-13
3.x/4.x Change MySQL encoding to UTF-8 Full Unicode (utf8mb4)


inoas:
This is a (multiple allowed):

* [x] feature-discussion (RFC)

* CakePHP Version: 3.4

### What you did

Converting my app from mariadb-utf8 to mariadb-utf8mb4

### What happened

Either I got no query results at all (when I forgot to push real utf8), or this:
```
Illegal mix of collations (utf8mb4_bin,IMPLICIT) and (utf8_general_ci,COERCIBLE) for operation '='
```

### What you expected to happen

This to work out of the box.

- app.php datasource encoding should probably change to utf8mb4
- MySql Driver baseConfig encoding should probably change to utf8mb4
- utf8 3 byte access will/should still work like before - **in addition** true utf8 (4byte) support will work without data being truncated before hitting the table collation already (without changing the table and/or column from utf8 to utf8mb4 - 4 byte code points will still fail)

markstory:
I am guessing you didn't set the encoding in your connection configuration?

inoas:
Well I did after I discovered an issue in sequel pro (does not auto detect utf8mb4).

I support setting it as default in cakephp/app and cakephp/cakephp; for posrgres it would map to utf8 as thats 4byte by defUlt as far as I know...

inoas:
Refs:
- https://medium.com/@adamhooper/in-mysql-never-use-utf8-use-utf8mb4-11761243e434
- https://mathiasbynens.be/notes/mysql-utf8mb4
- http://www.phptherightway.com/#utf-8-at-the-database-level

inoas:
Closing as PRs are open.

PR #11329, opened 2017-10-15
MySQL driver+test => utf8mb3 -> utf8mb4

inoas:
- [x] Enhancement
- [x] Bug (inserting valid utf8 such as 😃 will break your data using mysql/cakephp and default utf8 settings set by the driver and recommended app.php)

Refs:
* issue with more info: https://github.com/cakephp/cakephp/issues/11324#issuecomment-336672560
* related app skeleton PR https://github.com/cakephp/app/pull/554
* super-cedes https://github.com/cakephp/cakephp/pull/11328 (fixes unit tests)


inoas:
Is it a good idea to add a unit test that tests this though a MySQL connection?:

```sql
DROP TABLE IF EXISTS `utf8mb4test`;
CREATE TABLE `utf8mb4test` (
  `field` TEXT CHARACTER SET utf8mb4 NOT NULL
) ENGINE=INNODB;

SET NAMES utf8;
INSERT INTO utf8mb4test SET field = '😃'; # [ERROR IN QUERY 2] Incorrect STRING VALUE: '\xF0\x9F\x98\x83' FOR COLUMN 'field' AT ROW 1

SET NAMES utf8mb4;
INSERT INTO utf8mb4test SET field = '😃'; # Should work
```

If yes, where would it go and how should I do it? Is pure SQL against the driver, including the create table statement, ok?

inoas:
I'd need a little bit of assistance with the unit tests. I don't know how to select the test database...

chinpei215:
The `cake` database seems to be used for Mock Connection. So I think you cannot connect the database actually. You may have to choose one database in the following list:
https://github.com/cakephp/cakephp/blob/39d7fa96eaeaafb89f214f462878b669a832f9fa/.travis.yml#L49-L51

inoas:
Hm... locally I get a PDOException.

chinpei215:
@inoas What sql_mode are you using on your database server?

inoas:
Do you mean this?
```sql
# In
SELECT @@SQL_MODE
# Out
STRICT_TRANS_TABLES,ERROR_FOR_DIVISION_BY_ZERO,NO_AUTO_CREATE_USER,NO_ENGINE_SUBSTITUTION
```

inoas:
Let me know if this is any good, then I will squash all commits.

tests/TestCase/Database/Driver/MysqlTest.php
```diff
@@ -163,4 +163,138 @@ public function testCommitTransactionAutoConnect()
...
+
+        $connection = new \Cake\Database\Connection([
+            'driver' => '\Cake\Database\Driver\Mysql',
+            'persistent' => true,
+            'host' => 'localhost',
+            'username' => 'root',
+            'password' => '',
+            'database' => 'cakephp_test',
+            'port' => '3306',
+            // Try to insert utf8 4-byte symbol through utf8 3-byte connection character set
```
markstory: What is this test testing? It looks like it is testing MySQL, I'm not sure we need to do that. The tests above ensure that the DSN sent to PDO is correctly built which is enough for me.

inoas: Then I assume correctly that none of the new tests are required for this PR to become mergable, correct?

markstory: I don't think this test is necessary at least.

inoas: Well the DSN is tested above by another unit. I will remove all the other tests.

PR #11332, opened 2017-10-16
[2.x] Fix CookieComponent::delete() not working for deep children

chinpei215:
Refs #11280 (Problem 3)

Now [delete('foo.bar')](https://book.cakephp.org/2.0/en/core-libraries/components/cookie.html#CookieComponent::delete) works again as documented in the Cookbook.
Also, I simplified the read() method, on the way to fix a minor issue. Previously, `read('foo.0')` returned incorrect result.

### DIFF
--- a/lib/Cake/Controller/Component/CookieComponent.php
+++ b/lib/Cake/Controller/Component/CookieComponent.php
@@ -316,20 +316,16 @@ class CookieComponent extends Component {
 			$this->read();
 		}
 		if (strpos($key, '.') === false) {
-			if (isset($this->_values[$this->name][$key]) && is_array($this->_values[$this->name][$key])) {
-				foreach ($this->_values[$this->name][$key] as $idx => $val) {
-					$this->_delete("[$key][$idx]");
-				}
-			}
-			$this->_delete("[$key]");
 			unset($this->_values[$this->name][$key]);
-			return;
+			$this->_delete('[' . $key . ']');
+		} else {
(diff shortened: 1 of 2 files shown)


lib/Cake/Controller/Component/CookieComponent.php
```diff
@@ -316,20 +301,16 @@ public function delete($key) {
...
 			unset($this->_values[$this->name][$key]);
-			return;
-		}
-		$names = explode('.', $key, 2);
-		if (isset($this->_values[$this->name][$names[0]]) && is_array($this->_values[$this->name][$names[0]])) {
-			$this->_values[$this->name][$names[0]] = Hash::remove($this->_values[$this->name][$names[0]], $names[1]);
+			$this->_delete('[' . $key . ']');
+		} else {
+			$this->_values[$this->name] = Hash::remove((array)$this->_values[$this->name], $key);
+			list($key) = explode('.', $key, 2);
```
markstory: Why explode 2 segments if you're only going to use the first one?

chinpei215: It is faster than `explode('.', $key)`. Or would you prefer to use `strstr($key, '.', true)`? It is much faster though.

markstory: I hadn't considered the performance reason.

chinpei215:
Thank you.

PR #11335, opened 2017-10-17
Implement response::getType()

saeideng:
https://github.com/cakephp/app/pull/557

### DIFF
--- a/src/Http/Response.php
+++ b/src/Http/Response.php
@@ -1069,7 +1069,7 @@ class Response implements ResponseInterface
     }
 
     /**
-     * Retruns the current content type.
+     * Returns the current content type.
      *
      * @return string
      */
@@ -1177,7 +1177,7 @@ class Response implements ResponseInterface
     }
 
     /**
-     * Retruns the current charset.
+     * Returns the current charset.
      *
      * @return string
      */
--- a/src/Http/Response.php
+++ b/src/Http/Response.php
@@ -1046,14 +1046,14 @@ class Response implements ResponseInterface
     public function type($contentType = null)
     {
         if ($contentType === null) {
-            return $this->_contentType;
+            return $this->getType();
         }
         if (is_array($contentType)) {
             foreach ($contentType as $type => $definition) {
                 $this->_mimeTypes[$type] = $definition;
             }
 
-            return $this->_contentType;
+            return $this->getType();
         }
         if (isset($this->_mimeTypes[$contentType])) {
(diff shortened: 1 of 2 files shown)


src/Http/Response.php
```diff
@@ -1069,6 +1069,16 @@ public function type($contentType = null)
     }
 
     /**
+     * Retruns the current content type.
```
ADmad: Typo "Retruns"

saeideng: solved

src/Http/Response.php
```diff
@@ -1069,6 +1069,16 @@ public function type($contentType = null)
     }
 
     /**
+     * Returns the current content type.
+     *
+     * @return string
+     */
+    public function getType()
```
markstory: Should we update the deprecation notice for `Response::type()`

saeideng: I do it in new PR after merging this 
+
implementing `public function withMimeType(array $mimeTypes)`

saeideng: also https://github.com/cakephp/cakephp/pull/11336#issuecomment-337341305

markstory: If someone wants to use a content type with no short alias, they should give the full content type. I don't know if we need to allow the short type aliases to be modified.

tests/TestCase/Http/ResponseTest.php
```diff
@@ -214,6 +214,19 @@ public function testType()
...
+     * Tests the getType method
+     *
+     * @return void
+     */
+    public function testGetType()
+    {
+        $response = new Response();
+        $this->assertEquals('text/html', $response->getType());
+        $response->type('pdf');
+        $this->assertEquals('application/pdf', $response->getType());
```
markstory: Could we get a test for withType() and getType() being used together?

src/Http/Response.php
saeideng: Anyway we need a method for adding new mimetypes before full deprecating `type()`

ADmad: @saeideng You already can using `withHeader('Content-Type', 'foo/bar')` :)

saeideng: >withHeader('Content-Type', 'foo/bar')

@ADmad 
so we no longer need to new method( `withMimeType`) :)

ADmad: We could add that convenience too if people think it's worth saving typing of `'Content-Type'` :)

markstory:
Thanks @saeideng !

PR #11336, opened 2017-10-17
Add deprecated tag to Response::type()

ADmad:
Refs #11335


saeideng:
@ADmad needs a `setMimeType()`/`withMimeType()` before deprecating IMO
for bellow block 
```php
        if (is_array($contentType)) {
            foreach ($contentType as $type => $definition) {
                $this->_mimeTypes[$type] = $definition;
            }

            return $this->getType();
        }
```

ADmad:
Well then nvm, for now :P

PR #11339, opened 2017-10-18
Add deprecated tag to Response::type()

ADmad:
Refs #11335

### DIFF
--- a/src/Http/Response.php
+++ b/src/Http/Response.php
@@ -1041,7 +1041,8 @@ class Response implements ResponseInterface
      * ```
      *
      * @param string|null $contentType Content type key.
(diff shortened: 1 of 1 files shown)


src/Http/Response.php
```diff
@@ -1041,7 +1041,8 @@ public function httpCodes($code = null)
      * ```
      *
      * @param string|null $contentType Content type key.
-     * @return mixed Current content type or false if supplied an invalid content type
+     * @return mixed Current content type or false if supplied an invalid content type.
+     * @deprecated 3.5.5 Use getType() or withType() instead.
```
saeideng: and `withHeader()` maybe

saeideng (approved):

PR #11344, opened 2017-10-19
Add custom message to response code assertions

jadedcore:
It would be nice to be able to pass in a custom message to the response code convenience methods.  Especially since it is already supported in the protected method _assertStatus.

### DIFF
--- a/src/TestSuite/IntegrationTestCase.php
+++ b/src/TestSuite/IntegrationTestCase.php
@@ -763,11 +763,11 @@ abstract class IntegrationTestCase extends TestCase
     public function assertResponseCode($code, $message = null)
     {
         $actual = $this->_response->getStatusCode();
-        
+
         if (empty($message)) {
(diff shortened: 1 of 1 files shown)


src/TestSuite/IntegrationTestCase.php
```diff
@@ -702,39 +702,51 @@ public function viewVariable($name)
      *
      * @return void
      */
-    public function assertResponseOk()
+    public function assertResponseOk($message)
```
markstory: These new parameters have to be optional parameters for backwards compatibility. `$message = null` would be a good default.

markstory (changes requested):

markstory (approved):

markstory:
@jadedcore Looking good. Only a few lint errors left and this is good to go.

markstory:
Thanks!

Issue #11311, opened 2017-10-09
columnType JSON > json_encode options


andruhapx:
This is a (multiple allowed):

* [ ] bug
* [x] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.5.3

JsonType should have settings to set json_encode options.
In my case I want to use JSON_UNESCAPED_UNICODE, but only one way to do this - override class.

And also maybe it will be good idea to allow seting $assoc parameter of json_decode

dereuromark:
Are you able to provide a PR here for either 3.5 or 3.6?

andruhapx:
No, because TableSchema::setColumnType has no config parameter
and TypeInterface didnt use InstanceConfigTrait. 
I can work only with SVN, but not GIT.

Creating PR for this will be too big for me ;)

robertpustulka:
@andruhapx

> Creating PR for this will be too big for me ;)

You never know until you try it :)

saeideng:
I'll Open a PR for `json_encode:$options`  unless someone open before me
why we should having bellow options? is necessary ?

```php
 //json_encode:
$depth`
//json_decode 
$assoc 
 $depth 
$options
```

markstory:
>  In my case I want to use JSON_UNESCAPED_UNICODE, but only one way to do this - override class.

What is the problem with that?

andruhapx:
No problem, but type settings, like its already done for components, will be much close to cakephp styles. For any core and custom types

davidyell:
Could this be solved by a custom data type class?

markstory:
> Could this be solved by a custom data type class?

Of course it could.

saeideng:
@markstory 
let me know if needs to implement new method for `json_encode:$options` :)
and what is suggested method name for this

PR #11345, opened 2017-10-20
added json_encode option into jsonType

saeideng:
my suggested sample code to : https://github.com/cakephp/cakephp/issues/11311
```php
Type::build('json')
    ->encode(function($value){
             return json_encode($value, JSON_HEX_TAG,  100);
           });

```
feel free to accept or close :)

### DIFF
--- a/src/Database/Type/JsonType.php
+++ b/src/Database/Type/JsonType.php
@@ -38,6 +38,13 @@ class JsonType extends Type implements TypeInterface
     protected $_name;
 
     /**
+     * json_encode callback.
+     *
+     * @var string|null
+     */
+    protected $_encodeCallback = null;
+
+    /**
      * Constructor.
      *
      * (This method is declared here again so that the inheritance from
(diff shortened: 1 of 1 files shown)


davidyell:
This could do with a test case.

saeideng:
yes after accepting by core team

ADmad:
A callback seems overkill to me. I would just allow setting encoding options. Changing depth seems like a very uncommon use case and for that one should just use a custom type class.

saeideng:
And what is your suggest for json_decode()

markstory:
> And what is your suggest for json_decode()?

I still think we shouldn't be adding methods for this and instead get developers who need custom JSON handling to replace the JSON type.

ADmad:
I guess it's `..or close` then :stuck_out_tongue:

saeideng:
>I still think we shouldn't be adding methods for this (json_decode)
>I guess it's ..or close then 

do you agree with `setEncodeOption()` ?
if no ,please close the PR
(I don't understand that  you agree with `json_encode option` or no)

markstory:
@saeideng I don't think providing a callback to entirely replace the encoding process is the right approach. I could see having `setEncodeOptions(JSON_UNESCAPED_UNICODE | JSON_HEX_TAG)` and having a `setDecodeOptions()` method to allow control over the decoding.

I'm not confident that needing to customize how JSON is handled is a common enough requirement that we _should_ do this. Adding more methods for what seem like infrequent scenarios adds complexity to CakePHP. Replacing the Type mapping gives people complete control over json types and requires no work from us.

davidyell:
I agree, this is exactly what custom data types are for.

saeideng:
thanks @markstory  
closed

Issue #11342, opened 2017-10-19
Email validation gethostbynamel is used wrong


jeroenbos22:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 2.9.x / 2.10.x

### What you did
I validated email-adresses with deep validation. On our production-server it says every email-adress is valid. Tested with gmail.com and gmail.cok (filled in by user). Both tested true.

### What happened
When checking the Email validation done by cake I see the use of gethostbynamel (which is provided with the hostname as argument). If the output is an array it returns true. If you check the php documentation you need to provide an dot on the and otherwise it will check it as a local domain, and thats exectly whats happening because it will return an array with the servers ip when providing gmail.cok. Adding the dot after the hostname fixes the issue: 
return is_array(gethostbynamel($regs[1].'.'));

markstory:
Close as a pull request is open now.

markstory:
2.x fixed in #11371

PR #11356, opened 2017-10-23
Force email domain lookups to work in fallback case.

markstory:
In fallback scenarios, Validation::email() would resolve all domains to
localhost because of a missing `.`.

There are no additional tests as Travis and most PHP installations come with `checkdnsrr` or `getmxrr`.

Refs #11342

### DIFF
--- a/src/Validation/Validation.php
+++ b/src/Validation/Validation.php
@@ -656,7 +656,7 @@ class Validation
                 return true;
             }
 
-            return is_array(gethostbynamel($regs[1]));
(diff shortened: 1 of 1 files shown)


jeroenbos22:
@markstory I see this is not fixed in 2.x version, only milestone here 3.5.5. Can you also push this fix in the 2.x version? Thanks in advance!

markstory:
@jeroenbos22 I'll get around to it sometime soon.

PR #11358, opened 2017-10-23
Move Router::$initialized from RoutingMiddleware to BaseApplication

### LINKED ISSUE #11357: Routes are initialized before application hook in tests.
This is a (multiple allowed):

* [x] bug
* [ ] enhancement

* CakePHP Version: 3.5.4

### What you did
Set up routes using application hook.

```php
    public function routes($routes)
    {
            $routes->connect('/', ['controller' => 'Pages', 'action' => 'display', 'home']);
            $routes->fallbacks(DashedRoute::class);
        });
    }
```

And run tests:
```php
    public function testGet()
    {        
        $this->get('/');
        $this->assertResponseOk();
    }
```

### What happened
```
1) App\Test\TestCase\Controller\PagesControllerTest::testGet
Status code is not between 200 and 204
Failed asserting that 404 is equal to 204 or is less than 204.
```

Tests should pass.

### What do you suspect has happend
`Routes::$initialized` has been set ([here?](https://github.com/cakephp/cakephp/blob/a073f4336084a6090fc5ece6401299c649305cc2/src/Routing/Router.php#L586)) before the app hook had a chance to be fired.

robertpustulka:
An attempt to fix #11357

This fix will not prevent routes from `routes.php` to load even if the application routes have been loaded. At least in tests it seems. 

I'm not sure if this is a BC breaking change, but since nobody reported the issue before, it seems not many people moved to the app routes yet anyway.

### DIFF
--- a/src/Http/BaseApplication.php
+++ b/src/Http/BaseApplication.php
@@ -17,6 +17,7 @@ namespace Cake\Http;
 use Cake\Core\ConsoleApplicationInterface;
 use Cake\Core\HttpApplicationInterface;
 use Cake\Routing\DispatcherFactory;
+use Cake\Routing\Router;
 use Psr\Http\Message\ResponseInterface;
 use Psr\Http\Message\ServerRequestInterface;
 
@@ -69,7 +70,11 @@ abstract class BaseApplication implements ConsoleApplicationInterface, HttpAppli
      */
     public function routes($routes)
     {
-        require $this->configDir . '/routes.php';
+        if (!Router::$initialized) {
+            require $this->configDir . '/routes.php';
+            // Prevent routes from being loaded again
+            Router::$initialized = true;
+        }
     }
 
     /**
--- a/src/Routing/Middleware/RoutingMiddleware.php
+++ b/src/Routing/Middleware/RoutingMiddleware.php
@@ -56,11 +56,9 @@ class RoutingMiddleware
      */
     protected function loadRoutes()
     {
-        if ($this->app && !Router::$initialized) {
+        if ($this->app) {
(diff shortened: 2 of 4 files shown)


src/Http/BaseApplication.php
```diff
@@ -69,7 +70,11 @@ public function bootstrap()
      */
     public function routes($routes)
     {
-        require $this->configDir . '/routes.php';
+        if (!Router::$initialized) {
+            require $this->configDir . '/routes.php';
+            // Prevent routes from being loaded again
+            Router::$initialized = true;
```
markstory: Won't this mean that user overrides will need to set this flag as well?

src/Routing/Middleware/RoutingMiddleware.php
```diff
@@ -56,11 +56,9 @@ public function __construct(BaseApplication $app = null)
      */
     protected function loadRoutes()
     {
-        if ($this->app && !Router::$initialized) {
+        if ($this->app) {
             $builder = Router::createRouteBuilder('/');
             $this->app->routes($builder);
-            // Prevent routes from being loaded again
-            Router::$initialized = true;
```
markstory: Could this be done in the test harness?

src/Http/BaseApplication.php
robertpustulka: If they load the routes file then probably yes. If they load their own routes then it shouldn’t be necessary. 

Shouldn’t the flag be deprecated as well since there will be no `routes.php` in 4.0?

markstory: It should. The `Router::_loadRoutes()` is already deprecated.

Issue #11361, opened 2017-10-23
Allow conditions or callable to be passed to Association::setDependent() ?


kevinquinnyo:
This is a (multiple allowed):

* [ ] bug
* [x] enhancement
* [x] feature-discussion (RFC)

* CakePHP Version: eb609aff20945f8bc86cfb1fd3c50e6c62a502b0 (3.5.4)
* Platform and Target: n/a

### What you did
Tried to find a way to pass conditions/callable to a table's dependent association so that cascading deletes will occur or not occur based on a specific condition.

### What happened
I don't think this option is supported.

### What you expected to happen
I thought it would be nice if the feature existed.

Is there a feature like this already that I just don't see?

Basically I'm imagining something like this would be nice when setting up associations:
```php

# In say, Articles Table

$this->hasMany('Comments')
    ->setDependent(function ($comment) {
        return $comment->is_archived === false;
    });

# or maybe modifying the query in cascadeDelete would be better?
$this->hasMany('Comments')
    ->setDependent(function ($query) {
        return $query->where(['Comments.is_archived' => false]);
    });

# perhaps allow it to simply append to the $conditions array created in cascadeDelete?
$this->hasMany('Comments')
    ->setDependent(['Comments.is_archived' => false]);
```

I don't know how useful this would be to others, but unless I'm missing something obvious, if you want to automatically delete dependent records conditionally, you have to remove `->setDependent(true)` and handle this yourself in the `afterDelete` callback for instance, so this would be a handy shortcut I think.

kevinquinnyo:
by the way I realize my specific my example is kind of weird, but I just wanted to stick with the default Articles and Comments that are used as examples in the docs, etc

kevinquinnyo:
Just an additional thought -- since setDependent expects boolean (or null) currently, to avoid the headache of adding type checks inside the method (`if (is_callable($dependent))`, `if ($dependent === null)`, etc) and maybe for more flexibility it might be better to allow an `$options = []` to be passed to setDependent for this purpose?

markstory:
I don't know how the query modification approach would work. But having setDependent() calculate state based on the entity state could be reasonable.

kevinquinnyo:
@markstory No rush on this obviously. I just had some time at the end of the day here and thought I would give this a shot. Can you think of anything I'm overlooking with the minor changes in the PR?

I can write a test for it too but haven't yet.

PR #11363, opened 2017-10-23
Allow setDependent to accept an entity callback

kevinquinnyo:
This just allows setDependent to accept an optional callback (instead of just true/false) that is expected to return boolean and acts on the entity being deleted.

See issue #11361

I think this needs a test which I didn't write yet. Thoughts?

### DIFF
--- a/src/ORM/Association/BelongsToMany.php
+++ b/src/ORM/Association/BelongsToMany.php
@@ -586,9 +586,16 @@ class BelongsToMany extends Association
      */
     public function cascadeDelete(EntityInterface $entity, array $options = [])
     {
-        if (!$this->getDependent()) {
+        $dependent = $association->getDependent();
+
+        if (is_callable($dependent)) {
+            $dependent = (bool)$dependent($entity);
+        }
+
+        if ($dependent === false) {
             return true;
         }
+
         $foreignKey = (array)$this->getForeignKey();
         $bindingKey = (array)$this->getBindingKey();
         $conditions = [];
--- a/src/ORM/Association/BelongsToMany.php
+++ b/src/ORM/Association/BelongsToMany.php
@@ -586,7 +586,7 @@ class BelongsToMany extends Association
      */
     public function cascadeDelete(EntityInterface $entity, array $options = [])
     {
-        $dependent = $association->getDependent();
+        $dependent = $this->getDependent();
 
         if (is_callable($dependent)) {
             $dependent = (bool)$dependent($entity);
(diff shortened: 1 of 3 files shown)


kevinquinnyo:
Also, since I duplicated code in DependentDeleteHelper and in BelongsToMany I wonder if it would be better to have that in its own method somwehere

tests/TestCase/ORM/Association/BelongsToManyTest.php
```diff
@@ -346,6 +346,74 @@ public function testCascadeDeleteDependent()
...
+    {
+        $articleTag = $this->getMockBuilder('Cake\ORM\Table')
+            ->setMethods(['deleteAll'])
+            ->getMock();
+        $config = [
+            'sourceTable' => $this->article,
+            'targetTable' => $this->tag,
+            'sort' => ['id' => 'ASC'],
+            'dependent' => function ($tag) {
+                return true;
```
markstory: Would be good to have an assertion that the `$tag` argument is the entity.

src/ORM/Association/BelongsToMany.php
```diff
@@ -586,9 +586,16 @@ public function eagerLoader(array $options)
      */
     public function cascadeDelete(EntityInterface $entity, array $options = [])
     {
-        if (!$this->getDependent()) {
+        $dependent = $this->getDependent();
```
markstory: Doesn't this change the return types of `getDependent()` and `setDependent()` as new types can be provided/fetched?

tests/TestCase/ORM/Association/BelongsToManyTest.php
kevinquinnyo: Just had a chance to revisit this. I don't think it's done correctly because the entity that I'm actually passing to the callable is the entity that *started* the cascading delete.  But then it makes me wonder which is more useful? A callable on the target entities (that are dependent) or on the source entity?

I honestly ended up not needing this feature for what originally sparked my interest in it.  Do you think it's worth pursuing? I'm happy to do it if you think it would be a good addition, but I'm not sure the best way to proceed.

With deletes, we definitely want zero ambiguity as to what a callable is using as it's argument or bad things could happen.

markstory: If we don't have a solid set of use cases, I would rather not add features. We can revisit this when we better understand the use cases.

kevinquinnyo: Agreed

Issue #11362, opened 2017-10-23
CommandScanner should not duplicate plugin commands


markstory:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.5.x

### What you did

Commands discovered by `CommandScanner` are duplicated for plugin shells/commands as both the short name and long name are displayed. Eg.

```
Available Commands:

- bake
- bake.bake
- benchmark
- cache
- completion
- console
- debug_kit.benchmark
- debug_kit.whitespace
- help
- i18n
- migrations
- migrations.migrations
- orm_cache
- plugin
- routes
- server
- version
- whitespace
```

### What you expected to happen

Only one of each command name, which preference to the shorter aliases.

markstory:
Closing as a pull request is open now.

PR #11364, opened 2017-10-24
Don't duplicate command names in help listing.

markstory:
Only show the shortest name that resolves to an individual classname. This simplifies the help output and makes it similar to what we had in the ShellDispatcher implementation.

Refs #11362

### DIFF
--- a/src/Command/HelpCommand.php
+++ b/src/Command/HelpCommand.php
@@ -21,7 +21,6 @@ use Cake\Console\CommandCollectionAwareInterface;
 use Cake\Console\ConsoleIo;
 use Cake\Console\ConsoleOptionParser;
 use Cake\Console\ConsoleOutput;
-use Cake\Shell\Task\CommandTask;
 use Cake\Utility\Inflector;
 use SimpleXmlElement;
 
@@ -64,13 +63,8 @@ class HelpCommand extends Command implements CommandCollectionAwareInterface
             $io->out('<info>Available Commands:</info>', 2);
         }
 
-        if (!$this->commands) {
-            $this->commands = new CommandCollection($this->getCommands($io));
-        }
-
         $commands = $this->commands->getIterator();
         $commands->ksort();
-        $commands = new CommandCollection((array)$commands);
 
         if ($args->getOption('xml')) {
             $this->asXml($io, $commands);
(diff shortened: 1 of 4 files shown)


dereuromark:
Are there any plans to get the old plugin-sorted overview back? Which was a bit more compressed.

```
[AssetCompress] asset_compress

[Bake] bake

[DatabaseLog] database_log

[DebugKit] benchmark, whitespace

[IdeHelper] annotations, code_completion, phpstorm

[Migrations] migrations

[Queue] queue

[Setup] clear, current_config, db_maintenance, db_migration, indent, mailmap, maintenance_mode, reset, superfluous_whitespace, test_cli, user

[Tools] inflect

[CORE] cache, help, i18n, orm_cache, plugin, routes, server, version

[app] cleanups_importer, console, documentation, downloads, fixture_check, generator, release, test_mail
```
etc

dereuromark:
Could we have grouping with verbose flag or in some other opt-in way maybe?
Is there any plan on namespacing, so you can group by custom prefixes, as Symfony commands do: "foo:bar:baz" => "f:b:b" would also work then, or any non-conflicting unique beginning of a command.

markstory:
> Is there any plan on namespacing, so you can group by custom prefixes, as Symfony commands do: foo:bar:baz => f:b:b would also work then, or any non-conflicting unique beginning of a command.

Yes. It is on the longer list of things I want to do with the new command structure. In the ShellDispatcher implementation the full scope of all available commands is never known so doing prefix matching is challenging. The command names are opaque to `CommandCollection` but `:` and `.` separated segments could be supported.

markstory:
I'll get the builds fixed tonight, I forgot to add a stub class.

markstory:
> Could we have grouping with verbose flag or in some other opt-in way maybe?

Are you thinking something like `cake help --grouped`?

markstory:
I'm happy to discuss the value of having a grouped output in another ticket. I'm not sure I understand how it is better than the copy & pastable command listing we have today.

dereuromark:
With this released shouldnt it be solved with 3.5.5?
```
Available Commands:

- bake
- bake.bake
- benchmark
- cache
- completion
- console
- debug_kit.benchmark
- debug_kit.whitespace
- help
- i18n
- migrations
- migrations.migrations
- orm_cache
```

dereuromark:
I would also assume the debug_kit prefixes etc not be shown unless necessary.
because in fact "bin/cake benchmark" etc works just fine.

markstory:
@dereuromark This was merged into 3.next as it required changes for both commands and shells, and I was lazy.

Issue #11367, opened 2017-10-24
joins result null values for all field if the condition is in array key-value style


rrd108:
This is a (multiple allowed):

* [x] bug
* [x] enhancement

* CakePHP Version: 3.4.13
* Platform and Target: Apache/2.4.25 (Ubuntu), PHP Version 7.0.22-0ubuntu0.17.04.1, MySQL 5.7

### What you did
I have created a custom finder with a right join. I defined the second parameter for the right join in the array syntax as anywhere else.
`$this->find()->rightJoin(
            ['latest' => $latest],
            [
                'Histories.company_grant_id' => 'latest.cg_id',
                'Histories.created' => 'latest.latest'
            ]
        );`

### What happened
The generated SQL is ok, running it in the MySQL console returns the expected results. After executing the query I got the excepted number of entities, but all filed values are null.

### What you expected to happen
I expexted to get the entities with the values from the SQL.

Solution: do not use the array key-value syntax for the condition, but use equals instead.
`$this->find()->rightJoin(
            ['latest' => $latest],
            [
                'Histories.company_grant_id = latest.cg_id',
                'Histories.created = latest.latest'
            ]
        );`

If it is like that for some purpose it should be more clearly stated in the cookbook. This kind of behaviour is hard to debug, as the generated SQL is OK, but the results are not.

markstory:
Are you sure your query is being generated correctly? An array style join condition generally results in the values being set as bound parameters which in your case would result in something like

```sql
RIGHT JOIN () AS latest ON `Histories.company_grant_id` = 'latest.cg_id' AND `Histories.created` = 'latest.latest' 
```
Which is not what I think you're hoping to get.

Instead you might want to try making the values of your conditions into `IdentifierExpression` objects.

```php
use Cake\Database\Expression\IdentifierExpression;

[
  'Histories.company_grant_id' => new IdentifierExpression('latest.cg_id'),
  'Histories.created' => new IdentifierExpression('latest.latest')
] 
```

rrd108:
Well, I was sure that I got the desired SQL, but it seems you are right.
My query object is this:

```php
object(Cake\ORM\Query) {
	'(help)' => 'This is a Query object, to get the results execute or iterate it.',
[...]

PR #11368, opened 2017-10-24
test case for #11367

rrd108:
Refs https://github.com/cakephp/cakephp/issues/11367

### DIFF
--- a/tests/TestCase/Database/QueryTest.php
+++ b/tests/TestCase/Database/QueryTest.php
@@ -19,6 +19,8 @@ use Cake\Database\Query;
 use Cake\Database\StatementInterface;
 use Cake\Database\TypeMap;
 use Cake\Datasource\ConnectionManager;
+use Cake\ORM\Table;
+use Cake\ORM\TableRegistry;
 use Cake\TestSuite\TestCase;
 
 /**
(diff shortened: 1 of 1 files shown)


tests/TestCase/Database/QueryTest.php
```diff
@@ -403,6 +405,37 @@ public function testSelectRightJoin()
...
+            'connection' => $this->connection
+        ]);
+
+        $latest = $query->select([
+                'u_id' => 'Comments.user_id',
+                'latest' => $query->func()->max('Comments.created')
+            ])->from(['Comments' => 'comments'])
+            ->group('Comments.user_id');
+
+        $actual = $comments->find()->rightJoin(
```
markstory: Why is the right join relevant?

rrd108: In my original code I had a right join. I just kept it in the test.

markstory: Ok. I've posted a comment on #11367 as I think you use case is not generating the correct SQL for the join conditions as right side values are treated as bound parameters.

ADmad:
Closing as the provided test case isn't support to work.

PR #11370, opened 2017-10-25
Add Deprecation warnings to a few packages

markstory:
Add deprecation warnings to a few packages. I've updated the tests as well to leverage the `deprecated()` helper.

### DIFF
--- a/tests/TestCase/Error/DebuggerTest.php
+++ b/tests/TestCase/Error/DebuggerTest.php
@@ -155,7 +155,6 @@ class DebuggerTest extends TestCase
     /**
      * Test that setOutputFormat works.
      *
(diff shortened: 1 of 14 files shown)


tests/TestCase/Error/DebuggerTest.php
```diff
@@ -141,23 +141,41 @@ public function testExcerpt()
...
+        $this->deprecated(function () {
+            Debugger::outputAs('html');
+            $this->assertEquals('html', Debugger::outputAs());
+        });
+    }
+
+    /**
+     * Test that setOutputFormat works.
+     *
+     * @group deprecated
```
saeideng: unnecessary flag

PR #11371, opened 2017-10-25
Force email domain lookups to work in fallback case.

Milanzor:
Exactly the same fix as #11356, but for the Cake 2.x branch.


markstory:
Thank you for taking care of this. :clap:

PR #11369, opened 2017-10-24
Access table associations using the dot syntax

sdustinh:
Reference: #11366

Issue #11366, opened 2017-10-24
Table::association() dot format support


sdustinh:
This is a (multiple allowed):

* [ ] bug
* [x] enhancement
* [ ] feature-discussion (RFC)

```bash
sdustinh@sdustinh:~/github/app$ php --version
PHP 7.0.22-0ubuntu0.16.04.1 (cli) ( NTS )
Copyright (c) 1997-2017 The PHP Group
Zend Engine v3.0.0, Copyright (c) 1998-2017 Zend Technologies
    with Zend OPcache v7.0.22-0ubuntu0.16.04.1, Copyright (c) 1999-2017, by Zend Technologies
    with Xdebug v2.4.0, Copyright (c) 2002-2016, by Derick Rethans
sdustinh@sdustinh:~/github/app$ mysql --version
mysql  Ver 14.14 Distrib 5.7.20, for Linux (x86_64) using  EditLine wrapper
sdustinh@sdustinh:~/github/app$ composer show cakephp/*
cakephp/bake                1.4.2  Bake plugin for CakePHP 3
cakephp/cakephp             3.5.2  The CakePHP framework
cakephp/cakephp-codesniffer 3.0.1  CakePHP CodeSniffer Standards
cakephp/chronos             1.1.2  A simple API extension for DateTime.
cakephp/debug_kit           3.11.1 CakePHP Debug Kit
cakephp/elastic-search      0.3.5  An Elastic Search datasource and data mapper for CakePHP 3.0
cakephp/migrations          1.7.1  Database Migration plugin for CakePHP 3.0 based on Phinx
cakephp/plugin-installer    1.0.0  A composer installer for CakePHP 3.0+ plugins.
```

### What you did
```php
public function someMethodOnTable()
{
    $association = $this->association('Articles')->association('Comments')->association('Users');
}
```
Given how easy the dot syntax makes everything, it would be handy to use it in the association accessor.

```php
public function someMethodOnTable()
{
    $association = $this->association('Articles.Comments.Users');
}
```

ADmad:
Why not just use `$association = $this->Articles->Comments->Users;` ?

sdustinh:
I've been trying to move away from accessing objects as properties (from magic methods in general) and instead preferring to use direct method calls for access. Using the magic methods has caused me a headache more than a few times. For example, if I'm using the `association()` method I have the option to add some additional logic if I wanted:

```php
public function association($name)
{
[...]

PR #11375, opened 2017-10-25
Access table associations using the dot syntax

sdustinh:
Reference: #11366 #11369

### DIFF
--- a/src/ORM/Table.php
+++ b/src/ORM/Table.php
@@ -880,6 +880,8 @@ class Table implements RepositoryInterface, EventListenerInterface, EventDispatc
      */
     public function association($name)
     {
+        deprecationWarning('Use Table::getAssociation() instead.');
+
         return $this->getAssociation($name);
     }
 
--- a/src/ORM/Table.php
+++ b/src/ORM/Table.php
@@ -897,18 +897,18 @@ class Table implements RepositoryInterface, EventListenerInterface, EventDispatc
      */
     public function getAssociation($name)
     {
-        if (strpos($name, '.') !== false) {
-            list($name, $next) = array_pad(explode('.', $name, 2), 2, null);
-            $result = $this->_associations->get($name);
-
-            if ($result !== null && $next !== null) {
-                $result = $result->getTarget()->association($next);
-            }
-
-            return $result;
+        if (strpos($name, '.') === false) {
+            return $this->_associations->get($name);
         }
 
-        return $this->_associations->get($name);
+        list($name, $next) = array_pad(explode('.', $name, 2), 2, null);
+        $result = $this->_associations->get($name);
+
+        if ($result !== null && $next !== null) {
+            $result = $result->getTarget()->association($next);
+        }
+
+        return $result;
     }
 
     /**
--- a/src/ORM/Table.php
+++ b/src/ORM/Table.php
@@ -872,14 +872,27 @@ class Table implements RepositoryInterface, EventListenerInterface, EventDispatc
     }
 
     /**
-     * Returns an association object configured for the specified alias if any
+     * Returns an association object configured for the specified alias if any.
+     *
+     * The name argument also supports dot syntax to access deeper associations.
+     *
+     * ```
+     * $users = $this->association('Articles.Comments.Users');
+     * ```
      *
      * @param string $name the alias used for the association.
      * @return \Cake\ORM\Association|null Either the association or null.
      */
     public function association($name)
     {
-        return $this->_associations->get($name);
+        list($name, $next) = array_pad(explode('.', $name, 2), 2, null);
+        $result = $this->_associations->get($name);
+
+        if ($result !== null && $next !== null) {
+            $result = $result->getTarget()->association($next);
+        }
+
+        return $result;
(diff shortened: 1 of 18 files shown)


ADmad:
Instead of modifying `associations()` wouldn't it be better to add a new `getAssociations()` method which supports the new dot syntax and deprecate the former? Two birds with one stone.

sdustinh:
:+1: That would be awesome. I'll add it.

src/ORM/Table.php
```diff
@@ -872,14 +872,27 @@ public function hasBehavior($name)
...
+     * $users = $this->association('Articles.Comments.Users');
+     * ```
      *
      * @param string $name the alias used for the association.
      * @return \Cake\ORM\Association|null Either the association or null.
      */
     public function association($name)
     {
-        return $this->_associations->get($name);
+        list($name, $next) = array_pad(explode('.', $name, 2), 2, null);
```
ADmad: Optimization should be added for checking whether name contains `.` and return early from the method.

ADmad (changes requested):

dereuromark:
But singular association() => getAssociation() afaik.

ADmad:
> But singular association() => getAssociation() afaik.

Yeah, sorry I typoed in previous comment.

tests/TestCase/ORM/TableTest.php
```diff
@@ -511,6 +511,30 @@ function (Event $event, $query, $options) use ($expected) {
...
+        $members = TableRegistry::get('Members');
+        $groupsMembers = TableRegistry::get('GroupsMembers');
+
+        $groups->belongsToMany('Members');
+        $groups->hasMany('GroupsMembers');
+        $groupsMembers->belongsTo('Members');
+        $members->belongsToMany('Groups');
+
+        $association = $groups->association('GroupsMembers.Members.Groups');
+        $this->assertInstanceOf('Cake\ORM\Association\BelongsToMany', $association);
```
markstory: You could use `BelongsToMany::class` here 🏌️

sdustinh:
@ADmad This should be good to go unless you guys see anything you want changed.

src/ORM/Table.php
```diff
@@ -874,11 +874,40 @@ public function hasBehavior($name)
...
+     * ```
+     * $users = $this->getAssociation('Articles.Comments.Users');
+     * ```
+     *
+     * @param string $name the alias used for the association.
+     * @return \Cake\ORM\Association|null Either the association or null.
+     */
+    public function getAssociation($name)
+    {
+        if (strpos($name, '.') !== false) {
```
ADmad: We normally follow "return early" paradigm so this should instead check for false and return the association.

sdustinh: No problem. I'll take care of it now.

src/ORM/Table.php
```diff
@@ -874,12 +874,41 @@ public function hasBehavior($name)
...
+    {
+        if (strpos($name, '.') === false) {
+            return $this->_associations->get($name);
+        }
+
+        list($name, $next) = array_pad(explode('.', $name, 2), 2, null);
+        $result = $this->_associations->get($name);
+
+        if ($result !== null && $next !== null) {
+            $result = $result->getTarget()->association($next);
```
ADmad: You are using deprecated `association()` method :smile:

sdustinh: Haha, jesus.

src/ORM/Table.php
```diff
@@ -874,12 +874,41 @@ public function hasBehavior($name)
...
      * Returns an association object configured for the specified alias if any
      *
+     * @deprecated 3.6.0 Use getAssociation() instead.
      * @param string $name the alias used for the association.
      * @return \Cake\ORM\Association|null Either the association or null.
      */
     public function association($name)
     {
-        return $this->_associations->get($name);
+        return $this->getAssociation($name);
```
markstory: Could we get a `deprecationWarning()` here? It'll save some time in the future.

sdustinh:
@markstory Added a deprecation warning. I wasn't sure if you guy were adding tests for the deprecation warning or not. If you want that, I would be happy to add that as well.

markstory:
>  I wasn't sure if you guy were adding tests for the deprecation warning or not. If you want that, I would be happy to add that as well.

We do have tests for deprecated methods, but we're also in the process of updating all the internal usage of deprecated methods so that they are unused.

sdustinh:
@markstory Would you like me to replace all instances of `Table::association()` with `Table::getAssociation()` in this PR as well? If so, should this include those used in the tests?

markstory:
@sdustinh That would be ideal. All uses in tests should be updated except one that is a test for the association() method to help prevent regressions until 4.0

sdustinh:
@markstory I replaced all instances of `Table::association()` with `Table::getAssociation()` in both `src/` and `tests/`. I wasn't able to find a test specific for `Table::association()`. For now, I'll just push up those changes for review and if there is a test in there that I missed I'll patch that after.

markstory:
Looking good :+1:

PR #11376, opened 2017-10-25
Use string instance of constant for backwards compatibility.

ADmad:
Refs UseMuffin/Webservice#46


robertpustulka:
This breaks any custom events defined as a const.

Issue #11372, opened 2017-10-25
Return schema as public 


AndreLZGava:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: EXACT RELEASE VERSION OR COMMIT HASH, HERE.
* Platform and Target: YOUR WEB-SERVER, DATABASE AND OTHER RELEVANT INFO AND HOW THE REQUEST IS BEING MADE, HERE.

### What you did
I need to verify wich column is on a table dinamicaly. I tried $this->schema and $this->getSchema none of them allow me to manipulate what is inside,

### What happened
this returns allways that cannot access protected property

### What you expected to happen
I was expecting to have a method that allows me manipulate schema.

P.S. Remember, an issue is not the place to ask questions. You can use [Stack Overflow]
 (nobody anwseared )(https://stackoverflow.com/questions/tagged/cakephp)
for that or join the #cakephp channel on irc.freenode.net, where we will be more
than happy to help answer your questions.

Before you open an issue, please check if a similar issue already exists or has been closed before.

ADmad:
`schema` and `getSchema` are methods not properties which return schema instance. You can check if a column exists in schema using `in_array($column, $this->getSchema()->columns())`.

AndreLZGava:
Thank you very very very much

ADmad:
In future please use one of the support channels and don't post help questions on this issue tracker.

dereuromark:
IMO columns() should be get... prefixed and also named getColumnNames() to avoid this semantic issue with it - columns() vs. getColumn().

inoas:
Not sure if the getter/setter prefixing is so beneficial here but a Names suffix 👍

PR #11377, opened 2017-10-25
Allow debug() to provide current fields instead of all content.

dereuromark:
A toArray() and maybe even __debugInfo() could make sense here.

Refs https://github.com/cakephp/cakephp/issues/11372

Should the static stuff also be in there? Or just the non static attributes?

### DIFF
--- a/src/Database/Schema/TableSchema.php
+++ b/src/Database/Schema/TableSchema.php
@@ -834,6 +834,33 @@ class TableSchema implements TableSchemaInterface, SqlGeneratorInterface
 
         return $dialect->dropConstraintSql($this);
     }
+
+    /**
+     * Returns an array of the table schema.
+     *
+     * @return array
+     */
+    public function toArray()
+    {
+        return [
+            'columns' => $this->_columns,
+            'indexes' => $this->_indexes,
+            'constraints' => $this->_constraints,
+            'options' => $this->_options,
+            'typeMap' => $this->_typeMap,
+            'temporary' => $this->_temporary,
+        ];
+    }
+
+    /**
+     * Returns an array of the table schema.
+     *
+     * @return array
+     */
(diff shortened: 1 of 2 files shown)


src/Database/Schema/TableSchema.php
```diff
@@ -834,6 +834,33 @@ public function dropConstraintSql(Connection $connection)
...
+
+    /**
+     * Returns an array of the table schema.
+     *
+     * @return array
+     */
+    public function toArray()
+    {
+        return [
+            'columns' => $this->_columns,
```
markstory: Shouldn't this output include the table's name?

tests/TestCase/Database/Schema/TableTest.php
```diff
@@ -663,6 +663,48 @@ public function testTemporary()
...
+        $result = $table->__debugInfo();
+        $expected = [
+            'columns' => [],
+            'indexes' => [],
+            'constraints' => [],
+            'options' => [],
+            'typeMap' => [],
+            'temporary' => true,
+        ];
+        $this->assertSame($expected, $result);
```
markstory: Might be worth having a test for a table with columns.

dereuromark: Since this just maps the internal property I thought checking the keys suffices, the actual table and columns are tested with the designated setters and getters I would think.

src/Database/Schema/TableSchema.php
dereuromark: $this->myTable->getSchema()->toArray()

In this case I know the table from the myTable property, right?

markstory: In that scenario yes, but schema objects could be used in other contexts and knowing the table name would be useful.

PR #11382, opened 2017-10-28
 3.next - Add deprecation warnings to the Database package

markstory:
Add deprecation warnings to classes in the Database package. 

@lorenzo I was not able to add warnings to `Cake\Database\Query::orWhere()` as I was not able to easily replace it with `Query::where()` as we had originally thought. The problematic uses were in the [Marshaller](https://github.com/cakephp/cakephp/blob/database-deprecations/src/ORM/Marshaller.php#L689) and in [TupleComparisonTranslatorTrait](https://github.com/cakephp/cakephp/blob/b27ebbd6a4e9b6062bd45e13a439b7ab7f29d0af/src/Database/Dialect/TupleComparisonTranslatorTrait.php#L82). Any ideas on how that usage could be updated?

### DIFF
--- a/src/Database/Connection.php
+++ b/src/Database/Connection.php
@@ -206,6 +206,7 @@ class Connection implements ConnectionInterface
      */
     public function driver($driver = null, $config = [])
     {
+        deprecationWarning('Connection::driver() is deprecated. Use Connection::setDriver()/getDriver() instead.');
         if ($driver !== null) {
             $this->setDriver($driver, $config);
         }
@@ -380,6 +381,10 @@ class Connection implements ConnectionInterface
      */
     public function schemaCollection(SchemaCollection $collection = null)
     {
+        deprecationWarning(
+            'Connection::schemaCollection() is deprecated. ' .
+            'Use Connection::setSchemaCollection()/getSchemaCollection() instead.'
+        );
         if ($collection !== null) {
             $this->setSchemaCollection($collection);
         }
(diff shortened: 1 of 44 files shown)


inoas:
Can we atleast rename orWhere() to something like appendOrWhere()? or chainOrWhere()? or addOrWhere() to make sure newbies in 4.x do not fall for the confusion/trap orWhere can be?

Would we also remove andWhere() when orWhere() really get's removed?

markstory:
I am open to removing both methods if we can create an easy to use replacement. While the `orWhere` method is not perfect it doesn't have a simple replacement when incrementally building conditions right now.

markstory:
Refs #11385

saeideng:
i guess you test this by postgres DB :)
needs updates Mysql drivers 
do you sure current usage of `orWhere ` on `Marshaller `  and `TupleComparisonTranslatorTrait` are true ?
```php
///result of `orWhere(...)->orWhere(...)->orWhere(...)->orWhere(...)->..`
WHERE 
  (
    (
      Tests.id = 1 
      AND Tests.title = 'test'
    ) 
    OR Tests.id = 2 
    OR Tests.title = 'test' 
    OR Tests.id = 3 
    OR Tests.title = 'test' 
    OR Tests.id = 4 
    OR Tests.title = 'test'
  )

```

markstory:
@saeideng I didn't specifically look at the generated sql. But changing the existing `orWhere()` to where() resulted in test failures.

markstory:
I'll come back to the `orWhere()` issue once other deprecations have had their runtime warnings added.

Issue #11351, opened 2017-10-21
addFile to FormData don't add file to parts . . .


kingofnull:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: v3.4.13.
* Platform and Target: NGINX,  Ubuntu 16.4, PHP 5.6.

### What you did
Try to send file using http client.

### What happened
`addFile` didn't append to file to parts.

### What you expected to happen
Add file to parts.

Solution :
in file `/vendor/cakephp/cakephp/src/Http/Client/FormData.php`  line:172,function of `addFile`  add ` $this->add($part);` before return. the final function code:

```
public function addFile($name, $value)
    {
        $this->_hasFile = true;

        $filename = false;
        $contentType = 'application/octet-stream';
        if (is_resource($value)) {
            $content = stream_get_contents($value);
            if (stream_is_local($value)) {
                $finfo = new finfo(FILEINFO_MIME);
                $metadata = stream_get_meta_data($value);
                $contentType = $finfo->file($metadata['uri']);
                $filename = basename($metadata['uri']);
            }
        } else {
            $finfo = new finfo(FILEINFO_MIME);
            $value = substr($value, 1);
            $filename = basename($value);
            $content = file_get_contents($value);
            $contentType = $finfo->file($value);
        }
        $part = $this->newPart($name, $content);
        $part->type($contentType);
        if ($filename) {
            $part->filename($filename);
        }
        
        $this->add($part);
        return $part;
    }public function addFile($name, $value)
    {
        $this->_hasFile = true;

        $filename = false;
        $contentType = 'application/octet-stream';
        if (is_resource($value)) {
            $content = stream_get_contents($value);
            if (stream_is_local($value)) {
                $finfo = new finfo(FILEINFO_MIME);
                $metadata = stream_get_meta_data($value);
                $contentType = $finfo->file($metadata['uri']);
                $filename = basename($metadata['uri']);
            }
        } else {
            $finfo = new finfo(FILEINFO_MIME);
            $value = substr($value, 1);
            $filename = basename($value);
            $content = file_get_contents($value);
            $contentType = $finfo->file($value);
        }
        $part = $this->newPart($name, $content);
        $part->type($contentType);
        if ($filename) {
            $part->filename($filename);
        }
        
        $this->add($part);
        return $part;
    }
```

I fork cakephp.next but compare failed and cant apply a push request. anyway there is it:
[...]

PR #11383, opened 2017-10-28
Fix addFile bug

kingofnull:
addFile should add file to send parts but don't do it so I add the missing code.

#11351


markstory:
Thank you for the pull request but this was fixed in #11352

PR #11384, opened 2017-10-28
3.next - add deprecation warnings for Event & Filesystem

markstory:
Another batch of deprecation warnings added. I've updated the test cases to use the new methods unless there were existing tests for the new features already.


markstory:
Refs #11385

Issue #11385, opened 2017-10-28
3.6 - Add deprecation warnings across all packages


markstory:
We have a number of documentation only deprecations across CakePHP. Update these documentation warnings to be runtime errors using `deprecationWarning()`. Having runtime warnings will make it easier for application/plugin developers to know if they are ready for 4.0.0

### Packages to Update

* [x] Auth
* [x] Cache
* [x] Collection
* [x] Console
* [x] Core
* [x] Database
* [x] Event
* [x] Filesystem
* [x] Form
* [x] Http
* [x] I18n
* [x] Log
* [x] Mailer
* [x] Network
* [x] ORM
* [x] Routing
* [x] Shell
* [x] Template
* [x] TestSuite
* [x] Utility
* [x] Validation
* [x] View
* [x] Add warnings for class moves across the various packages.
* [x] Move http exceptions to Http package.
* [x] Revisit `orWhere()` usage in  ORM\Marshaller and in Database\TupleComparisonTranslatorTrait.
* [x] Deprecate ServerRequest::addParams() and ServerRequest::addPaths()
* [x] Add warnings for public properties in `ServerRequest`
* [x] Deprecate ServerRequest public properties.

josegonzalez:
For the shell deprecations, we actually deprecated the whole shell class. I don't think we want to add `deprecationWarning` to the top of the class as any inclusion will trigger it. Leaving it alone sounds fine, right?

josegonzalez:
I've marked `Shell` as complete as it's done in 3.next at least.

markstory:
> For the shell deprecations, we actually deprecated the whole shell class. I don't think we want to add deprecationWarning to the top of the class as any inclusion will trigger it. Leaving it alone sounds fine, right?

`Shell` is special in that it isn't going away until 5.0. I've not added warnings to the class moves yet, as I wasn't sure that was necessary. Now that you bring it up, old class locations should likely emit warnings too.

markstory:
Closing as to the best of my knowledge all the currently deprecated features have runtime warnings :tada:

HJerem:
@markstory 
Deprecated warnings are not displayed when in debug mode :
```php
        if (Configure::read('debug')) {
            $this->addPlugin(\DebugKit\Plugin::class);
        }
```

If I comment this out, deprecated warnings are displayed.

markstory:
> Is there an option to "not catch them" ? Because otherwise we need to check this panel everytime.

[...]

PR #11386, opened 2017-10-29
3.next - Add deprecation warnings to the Validation package

josegonzalez:
- Added missing cakephp/core dependency so that the deprecationWarning() is pulled in
- Added missing docblock deprecation message to Validation::blank()

Refs #11385


markstory:
Nice work!

PR #11387, opened 2017-10-29
3.next - Add deprecation warnings to the Utility package

josegonzalez:
- There are still a ton of Inflector::slug() tests. Should these be moved/consolidated to Text?
- OpenSsl does not support the rijndael method, so I've just set it as "to be removed".

Refs #11385

### DIFF
--- a/src/Utility/Inflector.php
+++ b/src/Utility/Inflector.php
@@ -752,7 +752,7 @@ class Inflector
     {
         deprecationWarning(
             'Inflector::slug() is deprecated. ' .
-            'Use Text::getSalt()/setSalt() instead.'
+            'Use Text::slug() instead.'
         );
         $quotedReplacement = preg_quote($replacement, '/');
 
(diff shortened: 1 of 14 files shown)


src/Utility/Inflector.php
```diff
@@ -750,6 +750,10 @@ public static function variable($string)
      */
     public static function slug($string, $replacement = '-')
     {
+        deprecationWarning(
+            'Inflector::slug() is deprecated. ' .
+            'Use Text::getSalt()/setSalt() instead.'
```
markstory: Shouldn't this be `Text::slug()`?

markstory (commented):
If there aren't tests for Text::slug() we should add some. The existing tests can be wrapped with `$this->deprecated()` to change the error level and allow the tests to pass.

src/Utility/Inflector.php
josegonzalez: No, you definitely want to get a salt and not a slug.

PR #11396, opened 2017-11-01
Fix collections when iterating over certain classes

dakota:
When a class that extends the base ArrayIterator class are iterated
over, the collection class would replace the class with a basic array or
ArrayIterator, breaking any custom logic in the original class.

This fixes the issue by checking if the class is an ArrayIterator,
rather than if it is simply a instance of ArrayIterator.

### DIFF
--- a/src/Collection/CollectionTrait.php
+++ b/src/Collection/CollectionTrait.php
@@ -819,7 +819,7 @@ trait CollectionTrait
     {
         $iterator = $this->unwrap();
 
-        if (get_class($iterator) === ArrayIterator::class) {
+        if (get_class($iterator) === ArrayIterator::class && $iterator instanceof ArrayIterator) {
             $iterator = $iterator->getArrayCopy();
         }
 
--- a/src/Collection/CollectionTrait.php
+++ b/src/Collection/CollectionTrait.php
@@ -819,7 +819,7 @@ trait CollectionTrait
     {
         $iterator = $this->unwrap();
 
-        if (get_class($iterator) === ArrayIterator::class && $iterator instanceof ArrayIterator) {
+        if (get_class($iterator) === ArrayIterator::class) {
             $iterator = $iterator->getArrayCopy();
         }
 
--- a/src/Collection/CollectionTrait.php
+++ b/src/Collection/CollectionTrait.php
@@ -512,7 +512,7 @@ trait CollectionTrait
     public function toArray($preserveKeys = true)
     {
         $iterator = $this->unwrap();
-        if (get_class($iterator) === ArrayIterator::class) {
+        if ($iterator instanceof ArrayIterator) {
             $items = $iterator->getArrayCopy();
 
             return $preserveKeys ? $items : array_values($items);
--- a/src/Collection/CollectionTrait.php
+++ b/src/Collection/CollectionTrait.php
@@ -512,7 +512,7 @@ trait CollectionTrait
     public function toArray($preserveKeys = true)
     {
         $iterator = $this->unwrap();
-        if ($iterator instanceof ArrayIterator) {
+        if (get_class($iterator) === ArrayIterator::class) {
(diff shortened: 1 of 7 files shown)


ADmad:
phpstan seems to be unhappy with your changes :)

src/Collection/CollectionTrait.php
```diff
@@ -512,7 +512,7 @@ public function insert($path, $values)
     public function toArray($preserveKeys = true)
     {
         $iterator = $this->unwrap();
-        if ($iterator instanceof ArrayIterator) {
+        if (get_class($iterator) === ArrayIterator::class) {
```
lorenzo: I think this case is safe to keep as it was before. We are about to return the array values anyway

markstory (commented):
Any chance we could get a test to help reproduce this?

dakota:
@markstory Test added

@ADmad Very silly change added to make PHPStan happy

src/Collection/CollectionTrait.php
```diff
@@ -819,7 +819,7 @@ protected function optimizeUnwrap()
     {
         $iterator = $this->unwrap();
 
-        if ($iterator instanceof ArrayIterator) {
+        if (get_class($iterator) === ArrayIterator::class && $iterator instanceof ArrayIterator) {
```
lorenzo: Isn't this redundant?

dakota: Very redundant, but otherwise PHPStan fails with a `Call to an undefined method Traversable::getArrayCopy()` error.

ADmad: We can just ignore the error in phpstan.neon instead of adding such redundant check.

ADmad: This looks like probably an error in phpstan and should be reported.

dakota:
@lorenzo Ready to go :D

lorenzo:
Sweet!

tests/TestCase/Collection/CollectionTest.php
```diff
@@ -2512,4 +2534,30 @@ protected function datePeriod($start, $end)
...
+        $newIterator = $iterator
+            ->filter(function ($item) {
+                return $item < 5;
+            })
+            ->reject(function ($item) {
+                return $item > 2;
+            });
+
+        $this->assertTrue(method_exists($newIterator, 'checkValues'), 'Our method has gone missing!');
+        $this->assertTrue($newIterator->checkValues());
```
chinpei215: @lorenzo Is this a correct use of the Collection class? Personally, I don't think so. Because, for example, if the method of inner iterator is `getPrimeNumbers()`, people would expect that it returns array(1), as 2, 3, 5 and 7 have been already filtered by `filter()` and `reject()` calls. However it would returns array(1, 2, 3, 5, 7) actually, as Collection doesn't remove any elements in the original iterator. I thought that was why we blocked calling `ArrayIterator::count()` via the Collection class.

Refs  #11440

lorenzo: It is another method, though. Not the results you get when iterating, so it could be argued that this is vivo the correct and at the same time not.

chinpei215: If it's a feature I think it would be needed to documented in the Collection page of our cookbook explicitly. Also, I would have to revert my PR #11045, as `method_exists($collection, 'cuustomMethod')` no longer works. `IteretorIterator` seems to handle `method_exists()` specially if the `__call()` method is not overridden, and we would not be able to emulate the behavior by any PHP code.

lorenzo: Let's not support that feature, then

chinpei215: OK. Then I will send a patch to fix the problematic test case.

Issue #11399, opened 2017-11-02
HTTP Client causes circular loop when auth config specified on Client


timalive:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.5.4
* Platform and Target: 
Request made from Chrome to a local apache2 webserver.  Http Client in question accessing another CakePHP 2 application with Digest auth enabled.

### What you did and What Happened
When a request is made with Digest Auth using Http Client, where settings are stored in the client (and not passed on the get function) and do not specify the domain, nonce and other Digest configs.
The Digest adapter then hits the server with no authentication to get the Digest configuration ([the _getServerInfo function of the Digest adapter](https://github.com/cakephp/cakephp/blob/master/src/Http/Client/Auth/Digest.php#L81))
Due to the fact that config is merged from the Client, this subsequent hit is also done using an incomplete Digest configuration, which again triggers a hit to the server to get Digest configuration and so on.

I will submit a pull request with a suggested fix.

timalive:
see pull request: https://github.com/cakephp/cakephp/pull/11400

markstory:
Closing as there is a pull request open.

PR #11400, opened 2017-11-02
Update Digest.php to fix issue 11399

timalive:
This seeks to fix issue [11399](https://github.com/cakephp/cakephp/issues/11399)

### DIFF
--- a/src/Http/Client/Auth/Digest.php
+++ b/src/Http/Client/Auth/Digest.php
@@ -83,7 +83,7 @@ class Digest
         $response = $this->_client->get(
             $request->getUri(),
             [],
-            ['auth' => ['type' => 'basic']]
+            ['auth' => ['type' => null]]
         );
 
         if (!$response->getHeader('WWW-Authenticate')) {
--- a/src/Http/Client/Auth/Digest.php
+++ b/src/Http/Client/Auth/Digest.php
@@ -83,7 +83,7 @@ class Digest
         $response = $this->_client->get(
             $request->getUri(),
             [],
-            ['auth' => []]
+            ['auth' => ['type'=>'basic']]
         );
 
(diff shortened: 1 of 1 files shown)


src/Http/Client/Auth/Digest.php
```diff
@@ -83,7 +83,7 @@ protected function _getServerInfo(Request $request, $credentials)
         $response = $this->_client->get(
             $request->getUri(),
             [],
-            ['auth' => []]
+            ['auth' => ['type' => 'basic']]
```
markstory: Why do you need to switch to basic auth in order to get the digest challenge information?

timalive: Digest requires an initial hit to the server if the realm, nonce etc are not known.  This doesn't have to be a digest hit - in fact any unauthenticated hit will give this in formation in the response headers (as I undestand it at least).  If it is left as digest, then that initial server hit triggers another initial server hit... which keeps going forever.  It could be left as digest if we could somehow break this loop.  Given the fact that "blank" auth (as was used here) defaults to basic, this simply ensures the code works as originally intended.

markstory: OK. Does setting `auth => false` disable the authentication headers entirely? My concern is trying to negotiate as basic auth and the server not responding with challenge headers.

markstory: Might need to do `auth => null` instead. `false` looks like it will result in errors.

timalive: ok let me check. sorry for the delay

timalive: OK.  I can get it to work on auth => null.  Will update the pull request.

markstory:
Thanks!

PR #11402, opened 2017-11-02
Adding deprecation warnings to the view package

burzum:
Refs #11385

### DIFF
--- a/src/View/Helper/HtmlHelper.php
+++ b/src/View/Helper/HtmlHelper.php
@@ -158,6 +158,11 @@ class HtmlHelper extends Helper
      */
     public function addCrumb($name, $link = null, array $options = [])
     {
+        deprecationWarning(
+            'HtmlHelper::addCrumb() is deprecated. ' .
+            'Use the BreadcrumbsHelper instead.'
+        );
+
         $this->_crumbs[] = [$name, $link, $options];
 
         return $this;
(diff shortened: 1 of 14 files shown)


src/View/Helper/HtmlHelper.php
```diff
@@ -784,6 +799,11 @@ public function getCrumbList(array $options = [], $startText = false)
      */
     protected function _prepareCrumbs($startText, $escape = true)
     {
+        deprecationWarning(
+            'HtmlHelper::_prepareCrumbs() is deprecated. ' .
```
saeideng: even for `protected` methods ?

burzum: Well, you can call them? @markstory has not told me any restrictions and I don't think it hurts to have it in one more place?

inoas: Makes sense to deprecate those as well, as long as it is not `final private` ;-)...

saeideng: >Well, you can call them when extending the class
OK :+1:

markstory:
Tests will need to be updated to either not use the deprecated methods or use the `deprecated()` test helper to change the error reporting level.

src/View/Helper/HtmlHelper.php
markstory: We should warn on protected methods as well. They will be removed and break code in the future.

markstory:
I think I got most of the tests fixed up.

ADmad:
Few CS errors need ti be fixed.

PR #11404, opened 2017-11-03
- DigestAuthenticate modification for cakephp 2.X

ynaderi:
In [password](https://github.com/cakephp/cakephp/blob/24e4acf9a38dc6a3a96fa5bebf9939988e715d77/lib/Cake/Controller/Component/Auth/DigestAuthenticate.php#L203) function realm is the third parameter but in [documentation](https://github.com/cakephp/cakephp/blob/24e4acf9a38dc6a3a96fa5bebf9939988e715d77/lib/Cake/Controller/Component/Auth/DigestAuthenticate.php#L47) is the seconde


markstory:
Typo fixed in f788c90b3cf57317da46fbb0fb9e8721ffbec9ed

PR #11405, opened 2017-11-04
Moved creation of HttpClient's response object into a separate method.

dmromanov:
Moved creation of HttpClient's response object into a separate method to allow easier extension.


markstory:
Thanks!

PR #11406, opened 2017-11-05
3.next - Add deprecations for core package.

markstory:
Add runtime warnings to deprecated methods in Core. `InstanceConfigTrait` and `StaticConfigTrait` had a ton of use so this is a fairly big pull request :cry:

Refs #11385

### DIFF
--- a/src/Core/Configure/Engine/PhpConfig.php
+++ b/src/Core/Configure/Engine/PhpConfig.php
@@ -93,6 +93,7 @@ class PhpConfig implements ConfigEngineInterface
         if (!isset($config)) {
             throw new Exception(sprintf('Config file "%s" did not return an array', $key . '.php'));
         }
+        deprecationWarning('PHP configuration files should not set `$config. Instead return an array.');
 
         return $config;
     }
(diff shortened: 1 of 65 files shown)


src/Core/Configure/Engine/PhpConfig.php
```diff
@@ -93,6 +93,7 @@ public function read($key)
         if (!isset($config)) {
             throw new Exception(sprintf('Config file "%s" did not return an array', $key . '.php'));
         }
+        deprecationWarning('PHP configuration files should not set `$config. Instead return an array.');
```
saeideng:
```php
if(!isset($config)){
   throw new Exception....
}else{
    throw new Warning......
}
```
what this say !?

markstory: The code path for a file returning an array is on line 89 - just outside if this diff.

tests/TestCase/Cache/Engine/XcacheEngineTest.php
```diff
@@ -75,7 +75,7 @@ public function tearDown()
      */
     public function testConfig()
     {
-        $config = Cache::engine('xcache')->config();
+        $config = Cache::getConfig('xcache')->getConfig();
```
saeideng: `Cache::engine()` did not deprecate 
I think your want to use 
`Cache::engine('xcache')->getConfig();`

markstory: Derp. You're right this is a mistake.

PR #11408, opened 2017-11-06
Make sure that CommandScanner scans proper classes.

robertpustulka:
Matching e.g. symfony commands under `Command` namespace shouldn't be detected by the `CommandScanner`.


markstory:
Thanks!

PR #11409, opened 2017-11-06
Fix deprecation warnings for Event class.

robertpustulka:
`$` needs an escape character.

### DIFF
--- a/src/Event/Event.php
+++ b/src/Event/Event.php
@@ -94,7 +94,7 @@ class Event
     {
         $method = 'get' . ucfirst($attribute);
         deprecationWarning(
-            "Event::${$attribute} is deprecated. " .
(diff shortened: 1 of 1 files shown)


src/Event/Event.php
```diff
@@ -120,7 +120,7 @@ public function __set($attribute, $value)
     {
         $method = 'set' . ucfirst($attribute);
         deprecationWarning(
-            "Event::${$attribute} is deprecated. " .
+            "Event::\${$attribute} is deprecated. " .
```
dereuromark: I would be more the friend of `sprintf('%s is deprecated.', 'Event::' . $attribute)` etc to clearly separate code and text parts. And also to avoid the escaping and  {} overhead then.

PR #11410, opened 2017-11-06
Avoid double parenthesis when nesting functions and query expressions

raul338:
For current behavior/examples see my post: [COUNT DISTINCT using Query->newExpr()?](http://discourse.cakephp.org/t/count-distinct-using-query-newexpr/3442)
In short: the following code generates an invalid SQL (at least for mysql)
```php
return $query->select([
    'required_approved' => $query->func()->count(
        $query->func()->DISTINCT(
            [
                $query->newExpr()->addCase(
                    (clone $baseExpr)
                        ->add($query->newExpr()->eq('DocumentTypes.required', 1))
                        ->add($query->newExpr()->eq('Documents.approved', 2)),
                    new IdentifierExpression('DocumentTypes.id')
                )
            ],
            ['literal'],
            ['integer']
        )
    ),
    // ...
);
```
```sql
SELECT (
    COUNT((  -- double parens breaking sql
        DISTINCT((
            CASE WHEN (
              DocumentTypes.entity_id = 1
              AND (DocumentTypes.id) IS NOT NULL
              AND IF(
                Providers.relation = 1, DocumentTypes.applyDep = 1,
                DocumentTypes.applyAuto = 1
              )
              AND DocumentTypes.required = 1
              AND Documents.approved = 1
            ) THEN DocumentTypes.id END
        ))
    ))
) AS `required_approved`,
-- ....
FROM ....
```

Then I thought that the double parenthesis are being generated somewhere, I found QueryExpression adds when it has more than one condition, and FunctionExpressions adds again even if QueryExpression is complex, resulting in two parenthesis

This PR attempts to fix that, I didn't know if split in two commits (first add a test and then the fix)

### DIFF
--- a/src/Database/Expression/FunctionExpression.php
+++ b/src/Database/Expression/FunctionExpression.php
@@ -168,7 +168,7 @@ class FunctionExpression extends QueryExpression implements TypedResultInterface
         $parts = [];
         foreach ($this->_conditions as $condition) {
             if ($condition instanceof ExpressionInterface) {
-                $condition = sprintf('(%s)', $condition->sql($generator));
+                $condition = sprintf('%s', $condition->sql($generator));
             } elseif (is_array($condition)) {
                 $p = $generator->placeholder('param');
                 $generator->bind($p, $condition['value'], $condition['type']);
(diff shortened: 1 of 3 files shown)


markstory:
Changes look reasonable, the tests are failing presently though.

lorenzo:
I'm surprised that this didn't break more stuff. Thanks for fixing it!

PR #11412, opened 2017-11-07
Allow registering middleware as string with RouteCollection.

ADmad:
Given that `MiddlewareQueue` accepts string names for middlewares, I don't see any reason for preventing use of strings when registering middlewares with `RouteBuilder`/`RouteCollection`.

### DIFF
--- a/src/Routing/RouteCollection.php
+++ b/src/Routing/RouteCollection.php
@@ -418,14 +418,11 @@ class RouteCollection
      * scope or any child scopes that share the same RouteCollection.
      *
      * @param string $name The name of the middleware. Used when applying middleware to a scope.
-     * @param callable $middleware The middleware object to register.
+     * @param callable|string $middleware The middleware callable or class name to register.
      * @return $this
      */
-    public function registerMiddleware($name, callable $middleware)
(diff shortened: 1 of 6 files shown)


src/Routing/RouteCollection.php
```diff
@@ -418,14 +418,11 @@ public function setExtensions(array $extensions, $merge = true)
...
-     * @param callable $middleware The middleware object to register.
+     * @param callable|string $middleware The middleware callable or class name to register.
      * @return $this
      */
-    public function registerMiddleware($name, callable $middleware)
+    public function registerMiddleware($name, $middleware)
     {
-        if (is_string($middleware)) {
-            throw new RuntimeException("The '$name' middleware is not a callable object.");
-        }
```
markstory: Should we check that the argument is a non-string callable or a valid class name and error out early?

ADmad: I don't think that's necessary.  MiddlewareQueue also accepts middleware name like `Foo` which it expands to class name using App::className() when the queue is run.

markstory: ok

PR #11414, opened 2017-11-07
Adding deprecation warning for the Http package

burzum:
Refs #11385


markstory:
I will work on getting the tests updated over the next few days as there are quite a few uses.

burzum:
@markstory please take a look at this commit https://github.com/cakephp/cakephp/pull/11414/commits/3f1d4dac10358f6b87edd9038b1789da309e8554 I had a short discussion with @lorenzo about this. If you want to do it different feel free to change it.

src/Http/ServerRequest.php
```diff
@@ -625,6 +635,11 @@ public function __call($name, $params)
      */
     public function __get($name)
     {
+        deprecationWarning(
+            'Accessing routing parameters through __get will removed in 4.0.0. ' .
```
ADmad: A message like `"Accessing '{$name}' as property has been deprecated and will be removed in 4.0.0"` might be more easily understandable to novice users.

markstory:
> please take a look at this commit 3f1d4da I had a short discussion with @lorenzo about this. If you want to do it different feel free to change it.

I think what you have is reasonable for that given the constraints we're facing.

tests/TestCase/Controller/Component/RequestHandlerComponentTest.php
```diff
@@ -542,37 +553,44 @@ public function testStartupProcessData()
...
     /**
      * Test mapping a new type and having startup process it.
      *
+     * @group deprecated
      * @return void
      * @triggers Controller.startup $this->Controller
      */
     public function testStartupCustomTypeProcess()
     {
-        $restore = error_reporting(E_ALL & ~E_USER_DEPRECATED);
```
saeideng: please re-base 
I removed this line in other PR
`testStartupCustomTypeProcess` should not flag `@group deprecated`

markstory: Thanks!

tests/test_app/TestApp/Controller/TestsAppsController.php
```diff
@@ -62,9 +60,7 @@ public function redirect_to_permanent()
 
     public function set_type()
     {
-        $this->response->type('json');
-
-        return $this->response;
+        return $this->response->withType('json');
```
saeideng: is this true?
```php
       $this->response->type('json');

        return $this->response;
```
will change origin `response` and returns (mutable )
```php
$this->response->withType('json');
```
will no change origin `response` and returns (immutable )


I mean

```php
// value of $this->response
$responseResult=set_type()

$this->response == $responseResult //before
$this->response != $responseResult //now
```

I know this is just a testCase

saeideng:
` Files changed 66 ` until now :clap:

tests/test_app/TestApp/Controller/TestsAppsController.php
markstory: You're right this doesn't modify the controller's `response` property. In this situation that isn't relevant as a controller action returning a response overrides the property.

markstory:
@saeideng Almost below 100 failing tests as well.

ADmad:
:white_check_mark:  :tada:

Issue #9023, opened 2016-06-22
Add validation features for validating non associative array


doshu:
This is a (multiple allowed):
- [ ] bug
- [ ] enhancement
- [x] feature-discussion (RFC)
- CakePHP version: 3.*

i found this issue in the following situation.
i have created a modeless form. the data from the form must have this schema.

```
[
 'tel' => '000000',
 'type' => [
    'foo',
   'lol',
    'asd'
 ]
]
```

the form is created dynamically and each type entry is a text field, so every input is created with the name = 'type.0' , 'type.1', ecc

i need that if the field for the type has been created by the user, validation force the user to fill it and check the maximum size.

validation now, cannot check every index of the array as a single field if i don't set a key.

validation should be allow to set a nested validator and set a rule like ->add('.', 'rulename', .....)
so the error message will be set for each different input

markstory:
Couldn't you write a custom validator for the `type` field and validate each item in the list inside your function?

doshu:
yes but if i write a custom validator, i cannot set a validation error for
each element of the erray and it will not work with the form helper
Il 23 Giu 2016 04:02, "Mark Story" <notifications@github.com> ha scritto:

burzum:
@doshu I don't think this very specific use case belongs to the core but I can see the problem with the error messages per nested field.

Maybe you can use the nested validation functionality to get this done? See http://book.cakephp.org/3.0/en/core-libraries/validation.html#nesting-validators

doshu:
nested validation cannot be setted for each element of an array. the nested
array must have a string key that identify the field.
a single array with elements with numeric keys must not nested validated

2016-06-30 12:02 GMT+02:00 Florian Krämer <notifications@github.com>:

markstory:
The `isArray` validator proposed in #9057 seems related to this. We could look at having another rule to apply a rule to every element in a list value. Something like:

```php
$validator->allElementsMust(['lengthBetween', 1, 5], 'All things must be between 1-5', $options);
```

thinkingmedia:
@markstory 

We will get stuck having to duplicate validation rules just for arrays.

What if we add support for `Hash::extract` as part of the fieldname, and then iterate the rule over what was extracted. This change only kicks in if `Validator` sees that a rule's field name looks like an extraction.

The usage would look like this.

```
  $validatior->integer('field.{n}.Foo.Bar');
  $validatior->requirePresence('field.{n}.Foo.Bar', 'create');
[...]

PR #11416, opened 2017-11-09
allow validation of fields nested inside array

lyziane:
This is for issue #9023

### DIFF
--- a/src/Validation/Validator.php
+++ b/src/Validation/Validator.php
@@ -16,6 +16,7 @@ namespace Cake\Validation;
 
 use ArrayAccess;
 use ArrayIterator;
+use Cake\Utility\Hash;
 use Countable;
 use InvalidArgumentException;
 use IteratorAggregate;
--- a/src/Validation/Validator.php
+++ b/src/Validation/Validator.php
@@ -104,6 +104,8 @@ class Validator implements ArrayAccess, IteratorAggregate, Countable
     {
         $errors = [];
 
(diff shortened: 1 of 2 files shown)


src/View/Form/EntityContext.php
```diff
@@ -584,6 +584,11 @@ public function error($field)
             return $entity->errors(array_pop($parts));
         }
 
+        if (is_array($entity)) {
+            $entity = $this->entity(array_shift($parts));
+            return $entity->errors($field);
```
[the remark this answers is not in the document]
lyziane: line 584 didn't have a blank line before so I just copied the style

src/Validation/Validator.php
```diff
@@ -104,6 +105,8 @@ public function errors(array $data, $newRecord = true)
     {
         $errors = [];
 
+        $data = Hash::flatten($data);
```
markstory: How do you see this interacting with the existing [nested validator](https://book.cakephp.org/3.0/en/core-libraries/validation.html#nesting-validators) features?

markstory (changes requested):
New features require tests to be merged. It also looks like you have caused existing tests to fail.

lyziane:
I was happy a little too fast. It worked in my project...and then more issues. This is no good...

src/Validation/Validator.php
lyziane: What I wanted is pretty much to store a modelless form inside my entity.

PR #11422, opened 2017-11-11
Use adventage of newer PHPUnit sytax

keradus:
Brought to you by [PHP CS Fixer](https://github.com/FriendsOfPHP/PHP-CS-Fixer)

### DIFF
--- a/tests/TestCase/Auth/BasicAuthenticateTest.php
+++ b/tests/TestCase/Auth/BasicAuthenticateTest.php
@@ -214,12 +214,12 @@ class BasicAuthenticateTest extends TestCase
     /**
      * test scope failure.
      *
-     * @expectedException \Cake\Network\Exception\UnauthorizedException
-     * @expectedExceptionCode 401
      * @return void
      */
     public function testAuthenticateFailReChallenge()
     {
+        $this->expectException(\Cake\Network\Exception\UnauthorizedException::class);
+        $this->expectExceptionCode(401);
         $this->auth->config('scope.username', 'nate');
         $request = new ServerRequest([
             'url' => 'posts/index',
--- a/tests/TestCase/Auth/ControllerAuthorizeTest.php
+++ b/tests/TestCase/Auth/ControllerAuthorizeTest.php
@@ -49,11 +49,11 @@ class ControllerAuthorizeTest extends TestCase
     }
 
     /**
-     * @expectedException \Cake\Core\Exception\Exception
      * @return void
      */
     public function testControllerErrorOnMissingMethod()
     {
+        $this->expectException(\Cake\Core\Exception\Exception::class);
         $this->auth->controller(new Controller());
     }
 
--- a/tests/TestCase/Auth/DigestAuthenticateTest.php
+++ b/tests/TestCase/Auth/DigestAuthenticateTest.php
@@ -107,12 +107,12 @@ class DigestAuthenticateTest extends TestCase
     /**
      * test the authenticate method
      *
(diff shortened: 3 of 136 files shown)


tests/TestCase/Auth/BasicAuthenticateTest.php
```diff
@@ -214,12 +214,12 @@ public function testAuthenticateSuccess()
...
      * test scope failure.
      *
-     * @expectedException \Cake\Network\Exception\UnauthorizedException
-     * @expectedExceptionCode 401
      * @return void
      */
     public function testAuthenticateFailReChallenge()
     {
+        $this->expectException(\Cake\Network\Exception\UnauthorizedException::class);
+        $this->expectExceptionCode(401);
```
markstory: Why is this notation an improvement? The annotation style is not deprecated and this type of change has the opportunity to create merge conflicts with the `3.next` branch.

keradus: it is best practice and suggested way of expecting expectations since almost 2 years
(TLDR: code over annotation for better detection&refactoring, method call could be put directly before code that is expected to raise exception, not to whole codeblock, so better control of what is raising the exception)

keradus: while merging into `3.next`, you could keep `3.next` version of code on each conflict and we could re-apply auto-changes in separated PR for `3.next`. that's easy ;)

dereuromark: > method call could be put directly before code that is expected to raise exception

Thats when I would start putting it in code, and that needs to be done manually.

keradus: that part of the benefit need to be achieved manually, yes.
having real class usage in code instead of text (annotation) is still benefit.

tests/TestCase/Cache/Engine/FileEngineTest.php
```diff
@@ -418,7 +418,7 @@ public function testPathDoesNotExist()
         ]);
 
         Cache::read('Test', 'file_test');
-        $this->assertTrue(file_exists($dir), 'Dir should exist.');
+        $this->assertFileExists($dir, 'Dir should exist.');
```
dereuromark: I would extract these changes as they are without discussion.

tests/TestCase/Auth/BasicAuthenticateTest.php
inoas: Maybe this could be done in/for the 3.next branch instead.

tests/TestCase/Cache/Engine/FileEngineTest.php
keradus: done

tests/TestCase/Auth/BasicAuthenticateTest.php
keradus: I sent this PR to default branch, if I shall change target branch, let me know ;)

markstory: I will get this merged into master/3.next over the next few days. Any conflicts will be a one time cost.

keradus: i'm not fully get it. it's already targeting master. shall I change PR target branch to sth ? (then I could solve the conflicts saving your time!)

markstory: What you have now is fine. I can take care of any conflicts that come up when I merge master -> 3.next.

PR #11433, opened 2017-11-16
PHPUnit - use expectExpectation in last cases

keradus:
continuation of #11422 

```
ker@dus:~/github/cakephp$ docker run -it --rm --name my-running-script -v "$PWD":/usr/src/myapp -w /usr/src/myapp jakzal/php-intl:7.1-60.1 php vendor/bin/phpunit tests/TestCase/View/ViewTest.php
PHPUnit 6.4.4 by Sebastian Bergmann and contributors.

................................................................. 65 / 89 ( 73%)
........................                                          89 / 89 (100%)

Time: 214 ms, Memory: 16.00MB

OK (89 tests, 189 assertions)
```

### DIFF
--- a/tests/TestCase/View/ViewTest.php
+++ b/tests/TestCase/View/ViewTest.php
@@ -939,7 +939,7 @@ class ViewTest extends TestCase
     public function testElementNonExistent()
     {
         $this->expectException(\Cake\View\Exception\MissingElementException::class);
-        $this->expectExceptionMessageRegExp('#^Element file \"Element[\\|/]non_existent_element\.ctp\" is missing\.$#');
+        $this->expectExceptionMessageRegExp('#^Element file "Element[\\\\/]non_existent_element\.ctp" is missing\.$#');
 
         $this->View->element('non_existent_element');
     }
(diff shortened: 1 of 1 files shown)


markstory:
Looks good but tests are failing on windows. Perhaps you need more `\` ?

keradus:
escape inside strings shall work the same under windows, won't they?
i need to find nice way to debug windows...

tests/TestCase/View/ViewTest.php
```diff
@@ -934,24 +934,26 @@ public function testPrefixElement()
...
      * Test loading non-existent view element
      *
-     * @expectedException \Cake\View\Exception\MissingElementException
-     * @expectedExceptionMessageRegExp $Element file \"Element[\\|/]non_existent_element\.ctp\" is missing$
      * @return void
      */
     public function testElementNonExistent()
     {
+        $this->expectException(\Cake\View\Exception\MissingElementException::class);
+        $this->expectExceptionMessageRegExp('#^Element file \"Element[\\|/]non_existent_element\.ctp\" is missing\.$#');
```
markstory: You might need `\\\` or `\\\\` instead of `\\` here.

saeideng:
another benefit of annotation style is you can use  RegExp with less escape  :)

keradus:
we did it everywhere out of the box ;) only this one is failing without extra escaping, and I still wouldn't say that it is escaping which is the issue. especially if it works on one CI, but not another.
TL;DR - disagree until proven

markstory:
Thanks @keradus

PR #11440, opened 2017-11-19
Fix access to inner iterator methods in some cases.

markstory:
A fix was made to 3.5 that allows subclasses of ArrayIterator to have userland methods defined on collections via IteratorIterator's __call features. However a conflicting change was made in 3.next to block access to the in-place mutation methods available on ArrayIterator. This set of changes attempts to compromise between those two goals.

This also fixes the 3.next build which is currently broken.

Refs #11045
Refs #11396


markstory:
Merging this so I can get other builds moving. I'm more than happy to revisit this if people aren't happy with the changes here.

PR #11441, opened 2017-11-19
Add missing core dependencies and bump versions.

markstory:
Most packages contain deprecation warnings now, which mean they rely on cakephp/core. They will also need at 3.6.0 as that will be the first version with deprecationWarning() in it.

### DIFF
--- a/src/Event/composer.json
+++ b/src/Event/composer.json
@@ -23,7 +23,8 @@
         "source": "https://github.com/cakephp/event"
     },
     "require": {
-        "php": ">=5.6.0"
+        "php": ">=5.6.0",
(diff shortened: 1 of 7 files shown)


lorenzo:
Wouldn't it be a lot better to just duplicate the function instead of making everything depend on a package just for one function?

The function is transient anyways.

markstory:
> The function is transient anyways.

I'm not sure the function will be transient. After slogging through the deprecations for 3.6 I think we should be adding runtime deprecations in each minor release as features are deprecated. Doing it all in one pass is taking a ton of time and will create a large amount of work for developers when they upgrade.

PR #11443, opened 2017-11-20
Docblock tweaks

davidyell:
Just a few minor tweaks

### DIFF
--- a/src/Http/Client.php
+++ b/src/Http/Client.php
@@ -28,7 +28,7 @@ use Zend\Diactoros\Uri;
  *
  * ### Scoped clients
  *
- * If you're doing multiple requests to the same hostname its often convenient
+ * If you're doing multiple requests to the same hostname it's often convenient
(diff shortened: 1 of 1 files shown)


davidyell:
Thanks @dereuromark 👍

src/Http/Client.php
```diff
@@ -442,7 +442,7 @@ protected function _sendRequest(Request $request, $options)
      * @param string $url Either a full URL or just the path.
      * @param string|array $query The query data for the URL.
      * @param array $options The config options stored with Client::config()
-     * @return string A complete url with scheme, port, host, path.
+     * @return string A complete url with scheme, port, host, and path.
```
ADmad: @davidyell Does the Queen's English use comma before "and"? :)

davidyell: https://en.oxforddictionaries.com/explore/what-is-the-oxford-comma

ADmad: Thanks for the reference.

Issue #11444, opened 2017-11-20
Bug in Marshaller::_mergeAssociation()


dereuromark:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

latest cakephp master

after looking into https://github.com/cakephp/cakephp/pull/11439 I found out that

    protected function _mergeAssociation($original, $assoc, $value, $options)

expects \Cake\Datasource\EntityInterface as $original, but then passes this on to

    public function mergeMany($entities, array $data, array $options = [])

via `return $marshaller->mergeMany($original, $value, (array)$options);`
which expect an array of entities and directly

    foreach ($entities as $entity) {}

over them.

PHPStorm marks this as code smell / error of course, and it seems this is indeed some bug, is it not?

PS: Same for _mergeJoinData()

markstory:
I think your pull request solves the issues with the doc block.

PR #11445, opened 2017-11-20
Follow up on #11439 regarding entity phpdoc typehinting.

dereuromark:
Refs https://github.com/cakephp/cakephp/issues/11444

But open issue still is _mergeAssociation()
Is $original here in fact maybe

    @param \Cake\Datasource\EntityInterface|\Cake\Datasource\EntityInterface[] $original The original entity

maybe?


markstory:
I think you're right that the `$original` parameter is one or many entities.

dereuromark:
My last commit should now fix the linked ticket and issue.

PR #11447, opened 2017-11-20
Fix more doc blocks.

dereuromark:
And one CS issue - Oh how I hate spaces..

### DIFF
--- a/src/ORM/Association/Loader/SelectWithPivotLoader.php
+++ b/src/ORM/Association/Loader/SelectWithPivotLoader.php
@@ -48,7 +48,7 @@ class SelectWithPivotLoader extends SelectLoader
     /**
      * Custom conditions for the junction association
      *
-     * @var mixed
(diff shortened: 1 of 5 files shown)


src/ORM/Association/Loader/SelectWithPivotLoader.php
```diff
@@ -111,7 +111,7 @@ protected function _buildQuery($options)
         }
 
         $query
-           ->where($this->junctionConditions)
+            ->where($this->junctionConditions)
```
dereuromark: Shouldn't this be detected by some CS tool? I mean 3 spaces seem really weird to accept, if we have tools running here in Travis.

ADmad: That is strange, it should have been reported by phpcs.

dereuromark: It probably has an issue with multiline.

PR #11451, opened 2017-11-21
More phpstan related fixes.

dereuromark:
PHPSTAN level 3

    [ERROR] Found 62 errors    

=>

    [ERROR] Found 25 errors       

:-)

We should also start inline deprecations for specific (and really bad) default values like bool where in future API it will actually have to be null (as per strict php7.1+).
This way people can already be a bit more aware and maybe adjust already the code to reflect this.

### DIFF
--- a/src/Datasource/QueryTrait.php
+++ b/src/Datasource/QueryTrait.php
@@ -87,8 +87,8 @@ trait QueryTrait
      * When called with a Table argument, the default table object will be set
      * and this query object will be returned for chaining.
      *
-     * @param \Cake\Datasource\RepositoryInterface|null $table The default table object to use
-     * @return \Cake\Datasource\RepositoryInterface|$this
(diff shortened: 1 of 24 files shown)


src/Datasource/QueryTrait.php
```diff
@@ -87,8 +87,8 @@
      * When called with a Table argument, the default table object will be set
      * and this query object will be returned for chaining.
      *
-     * @param \Cake\Datasource\RepositoryInterface|null $table The default table object to use
-     * @return \Cake\Datasource\RepositoryInterface|$this
+     * @param \Cake\Datasource\RepositoryInterface|\Cake\ORM\Table|null $table The default table object to use
```
ADmad: The param is type hinted as RepositoryInterface and Table implements that interface, so why is explicitly specifying Table necessary?

dereuromark: Because it is used in other places with the concrete one (code smell that cant be fixed until the next major).

PR #11454, opened 2017-11-22
Digest secret is not defaulting to Salt

timalive:
Digest secret is returning null. Because we use Salt and Secret interchangeably in the code, this causes chaos.  The reason for this is that ```Configure::read('Security.salt')``` returns null and not the Salt.


markstory:
Looks like we're still incorrectly using `Security::getSalt()` in one place. I'm going to merge these changes with those in #11453 and get tests passing again.

timalive:
ok, yes my commit was rather hasty ... This will help!

markstory:
Changes merged into #11453 now.

PR #11456, opened 2017-11-23
Add option to make `_validAgentAndTime` 3.x compatible

### LINKED ISSUE #10563: CakePHP 2.x // 3.x Session.time inconsistent
This is a (multiple allowed):

* [x] bug
* [x] enhancement

* CakePHP Version: 3.4.1
* Platform and Target: nginx / apache, mysql

### What you did
Tried to migrate old CakePHP 2 project step by step to version 3.x, i put a nginx proxy before both versions and route everything to the old version. If a new controller is able to handle the user needs i route with nginx reverse proxy /converted_controller_name to the CakePHP 3 project.

### What happened
Session closed, coming back from CakePHP 3 back to CakePHP 2 (browser back button or existing link) because of time expired.

#### CakePHP 2 Session Validation

// lib/Cake/Model/Datasource/CakeSession.php

	protected static function _validAgentAndTime() {
		$userAgent = static::read('Config.userAgent');
		$time = static::read('Config.time');
		$validAgent = (
			Configure::read('Session.checkAgent') === false ||
			isset($userAgent) && static::$_userAgent === $userAgent
		);
		return ($validAgent && static::$time <= $time);
	}

#### CakePHP 3 Session Timeout Check

// src/Network/Session.php

    protected function _timedOut()
    {
        $time = $this->read('Config.time');
        $result = false;

        $checkTime = $time !== null && $this->_lifetime > 0;
        if ($checkTime && (time() - $time > $this->_lifetime)) {
            $result = true;
        }

        $this->write('Config.time', time());

        return $result;
    }

EXPLAIN WHAT IS TO BE EXPECTED, HERE.

The logout is caused because CakePHP 2 and 3 are using Config.time in a different way, once with the last time the session was written and once the time when the session expires. Would be helpful to ether update the name to something different or use it in the same way. Otherwise its hard to share a session between two versions of the same framework.

tersmitten:
fixes #10563

### DIFF
--- a/lib/Cake/Model/Datasource/CakeSession.php
+++ b/lib/Cake/Model/Datasource/CakeSession.php
@@ -135,6 +135,13 @@ class CakeSession {
 	protected static $_cookieName = null;
 
 /**
+ * Whether or not to make `_validAgentAndTime` 3.x compatible.
+ *
+ * @var bool
+ */
+	protected static $_useForwardsCompatibleTimeout = false;
+
+/**
  * Whether this session is running under a CLI environment
  *
  * @var bool
@@ -360,6 +367,9 @@ class CakeSession {
 	protected static function _validAgentAndTime() {
 		$userAgent = static::read('Config.userAgent');
 		$time = static::read('Config.time');
+		if (static::$_useForwardsCompatibleTimeout) {
+			$time += (Configure::read('Session.timeout') * 60);
+		}
 		$validAgent = (
 			Configure::read('Session.checkAgent') === false ||
 			isset($userAgent) && static::$_userAgent === $userAgent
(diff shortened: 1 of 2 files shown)


dereuromark (approved):
Looks good to me (and BC).

lib/Cake/Model/Datasource/CakeSession.php
```diff
@@ -527,6 +537,10 @@ protected static function _configureSession() {
 		if (isset($sessionConfig['timeout']) && !isset($sessionConfig['cookieTimeout'])) {
 			$sessionConfig['cookieTimeout'] = $sessionConfig['timeout'];
 		}
+		if (isset($sessionConfig['useForwardsCompatibleTimeout']) && $sessionConfig['useForwardsCompatibleTimeout']) {
+				static::$_useForwardsCompatibleTimeout = true;
```
markstory: This looks like it has an extra layer of indentation.

tersmitten: I'll have a look

tersmitten: You were right, you hawk :-)

tersmitten:
Thanks

PR #11457, opened 2017-11-23
Fix map() resetting cached object allowing build() to use the updated object

dereuromark:
map() does not reset internal object cache for this key, and there is no way to tell build() to make a fresh one.

As Travis around https://github.com/dereuromark/cakephp-ide-helper/pull/64/commits/f3128fed1c4f99afe9c7f3b74b0b6c5c85d27c88 shows, mapping a different type after building has already been done still uses the old object.
Using set() and an actual instance (`Type::set('uuid', new UuidType());`) works, but I think resetting here is the expected and intuitive behavior after remapping.

With this fix `Type::map('uuid', UuidType::class);` etc works.

### DIFF
--- a/src/Database/Type.php
+++ b/src/Database/Type.php
@@ -162,7 +162,9 @@ class Type implements TypeInterface
         if ($className === null) {
             return isset(self::$_types[$type]) ? self::$_types[$type] : null;
         }
+
         self::$_types[$type] = $className;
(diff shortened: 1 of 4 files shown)


markstory:
Is there a way we could capture the changed behavior with a test case?

dereuromark:
I added some tests, and cleaned up the smell about the inline class in the process :)

PR #11460, opened 2017-11-24
CounterCacheBehavior: Allow callback to refuse update

ypnos-web:
In some circumstances (see below) it is desirable to allow the refusal of updating a counter field. This is easy to do with a callback function, as the function may just return `null` instead of a count. A simple patch that you might like to accept.

My Motivation:
Currently, the CounterCacheBehavior only works if there is no condition on the association. A counter example for a Comments table:

```php
$this->belongsTo('Articles')
    ->setConditions(['belongs_to' => 'articles'])
    ->setForeignKey('foreign_id');
$this->addBehavior('CounterCache', [
    'Articles' => [ 'comment_count' => ['conditions' => ['belongs_to' => 'articles']]]
```
This works for all Comment entities that are associated with an Article. Comments that are not associated with an Article will still lead the behavior to try an update on the Articles table. With the patch, a custom callback can be used instead:
```php
$this->addBehavior('CounterCache', [
    'Articles' => [ 'comment_count' => function ($event, $entity, $table) {
        if ($entity->belongs_to != 'articles')
            return null;
        …
    ]]);
```

I am happy to provide corresponding tests if you would like to accept this pull request.

### DIFF
--- a/src/Error/Debugger.php
+++ b/src/Error/Debugger.php
@@ -766,7 +766,7 @@ class Debugger
     /**
      * Takes a processed array of data from an error and displays it in the chosen format.
      *
-     * @param string $data Data to output.
+     * @param array $data Data to output.
      * @return void
      */
     public function outputError($data)
--- a/src/Error/ErrorHandler.php
+++ b/src/Error/ErrorHandler.php
@@ -160,11 +160,11 @@ class ErrorHandler extends BaseErrorHandler
     }
 
     /**
-     * Log both PHP5 and PHP7 errors.
+     * Logs both PHP5 and PHP7 errors.
      *
      * The PHP5 part will be removed with 4.0.
      *
-     * @param \Throwable|\Exception $exception
+     * @param \Throwable|\Exception $exception Exception.
      *
      * @return void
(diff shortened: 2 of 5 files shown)


markstory:
Isn't this a breaking change? You are changing the semantics of what null does.

ypnos-web:
This is a valid point, in the current implementation it is possible to set the database field to NULL by returning null.

Do you think we should find a more complex solution to address the needed functionality, or would you say there is no demand in extending the Behavior for something like this?

markstory:
Using false or raising an exception might be a safer way to handle aborts. False would mirror aborts in event listeners, and an exception would be very safe from a compatibility perspective.

ypnos-web:
This is a solution using `false`. What do you think?

markstory:
Looks good outside of a few phpcs erorrs.

ypnos-web:
Yeah that was some stuff my IDE was doing.

markstory:
Rebased and merged into `3.next` in 66633f741d1b2e1f720e67e10b7bf589a732e878

Issue #11455, opened 2017-11-22
ignored exist field when using "EntityRoute"


saeideng:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

https://github.com/cakephp/cakephp/pull/11300
in route class
```php
		$url['lang'] = 'en';
		if (isset($url['_entity'])) {

				$entity = $url['_entity'];
				preg_match_all('@:(\w+)@', $this->template, $matches);

				foreach($matches[1] as $field) {
						$url[$field] = $entity[$field];
				}
		}
		debug($url);
```
my entity ( ` $url['_entity']` )  does not any field named `'lang'` 
`debug($url); ` returns 
```php
//
'lang`=>null
//
```
so you can not use 
```php
Router::url([ 'lang' => 'en', '_name' => 'articles:view', '_entity' => $article]);
```

saeideng:
I fixed this by a `if` in my app
```php
				foreach($matches[1] as $field) {
						if(!isset($url[$field])){
							$url[$field] = $entity[$field];
						}

				}
```

saeideng:
but above solution is not true for general usage

markstory:
I think only extracting fields that aren't already defined in the route array makes sense. It gives a way to explicitly override the entity data as well.

lorenzo:
We would need to distinguish between stuff that was passed manually and fields that are present because of the `$this->defaults` property in the Route

markstory:
@lorenzo Does the following precedence make sense:

* URL key
* Entity property
* Route defaults

lorenzo:
It does

PR #11461, opened 2017-11-25
Allow url data to overwrite Entity data.

markstory:
When a routing parameter is defined in the URL array it should overwrite the entity data. This gives us a way to explicitly overwrite the entity data if necessary.

Right now there isn't a way to define fallback parameters but that is on the roadmap for 3.6.0

Refs #11455

### DIFF
--- a/tests/TestCase/Routing/Route/EntityRouteTest.php
+++ b/tests/TestCase/Routing/Route/EntityRouteTest.php
@@ -24,6 +24,34 @@ use TestApp\Model\Entity\Article;
 class EntityRouteTest extends TestCase
 {
     /**
+     * test that route keys take precedence to object properties.
+     *
+     * @return void
+     */
+    public function testMatchRouteKeyPrecedence()
+    {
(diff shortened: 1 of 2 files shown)


tests/TestCase/Routing/Route/EntityRouteTest.php
```diff
@@ -23,6 +23,34 @@
...
+            ]
+        );
+
+        $result = $route->match([
+            'slug' => 'other-slug',
+            '_entity' => $entity,
+            '_name' => 'articlesView'
+        ]);
+
+        $this->assertEquals('/articles/2/other-slug', $result);
```
saeideng: needs another test for `Route defaults`

markstory: Route defaults currently cannot overlap routing keys. That feature will be implemented late in 3.6 though.

lorenzo:
Looks good!

Issue #9043, opened 2016-06-28
What are we going to do about PHP7 "throwable"


thinkingmedia:
This is a (multiple allowed):
- [ ] bug
- [ ] enhancement
- [x] feature-discussion (RFC)
- CakePHP Version: 3.6.10
- Platform and Target: Apache 2.4, PHP 7.0
### What you did

Keep running into cases where CakePHP is not handling PHP 7's new "Error" and "Throwable" exceptions.
### Expected Behavior

Cells trigger a server 500 if there is a `Error` thrown during render.
Transactions not commit and leak if there is an `Error` class thrown.
Basically, anywhere there is a `catch(Exception $ex)` the code does not work as expected in PHP 7.
### Actual Behavior

PHP 7 introduced a new error type `Error`, and gave `Error` and `Exception` a common interface `Throwable`.

All `try/catch` methods that do a broad catch all `catch(Exception $e)` should be doing `catch(\Throwable $e)` in PHP 7.

I don't know how to address this issue in a backward compatible way.

Maybe use a global function that wraps a closure in the proper try/catch type for the current PHP version?

As this been discussed anywhere else by the Cake team?

thinkingmedia:
Would something like this work without issues in PHP 5.x?

I found this while googling, but wouldn't Throwable be a undeclared type in PHP 5?

```
try {
    // Code that may throw an Exception or Error.
} catch (Throwable $t) {
    // Executed only in PHP 7, will not match in PHP 5.x
} catch (Exception $e) {
    // Executed only in PHP 5.x, will not be reached in PHP 7
}
```

thinkingmedia:
I think I can make a PR to fix these. I found 27 try/catch pairs that need to be fixed.

markstory:
> I found this while googling, but wouldn't Throwable be a undeclared type in PHP 5?

PHP is funny in that it won't complain about missing classes used in typehints/catch blocks. It will however try to autoload them.

Instead of duplicating all the `catch` blocks, could we conditionally define `Throwable`? Something like:

```php
if (!class_exists('Throwable', false)) {
  class Throwable extends Exception
  {
  }
}
```

That should enable us to use `\Throwable` throughout CakePHP in a transparent to PHP7 way.

thinkingmedia:
It's an interface, but still. I think you're onto something. I might close this and try again.

markstory:
[...]

Issue #11323, opened 2017-10-13
Custom ErrorHandlers are not used by ErrorHandlerMiddleware


chinpei215:
This is a (multiple allowed):

* [ ] bug
* [ ] enhancement
* [x] feature-discussion (RFC)

* CakePHP Version: 3.5.3
* Platform and Target: PHP7

### What you did

Declare the following controller:
```php
class TestsController extends AppController
{
    public function exception() {
        throw new \Exception();
    }

    public function error() {
        throw new \Error();
    }
}
```
Access `/tests/exception` and `/tests/error`.

### What happened

While Exceptions are handled by `ErrorHandlerMiddleware::handleException()`, PHP7 Errors are handled by `ErrorHandler::handleException()`.

### What you expected to happen

Both Exceptions and PHP7 Errors are handled by `ErrorHandlerMiddleware::handleException()` if `ErrorHandlerMiddleware` is enabled.

### Related code

https://github.com/cakephp/cakephp/blob/35f70ad2c9843b1c42afba072615ce9141d3bf45/src/Error/Middleware/ErrorHandlerMiddleware.php#L92-L99

### Related issues

#8997, #9500, #9931

robertpustulka:
Would something like that work both in PHP 5 and 7?

```php
         try { 
             return $next($request, $response); 
         } catch (Exception $e) { 
             return $this->handleException($e, $request, $response); 
         } catch (Throwable $e) { 
             return $this->handleException($e, $request, $response); 
         }  
```

chinpei215:
Yes. And if there are no problems, I would like to add the same catch blocks to [Connection::transactional()](https://github.com/cakephp/cakephp/blob/7c988031eb5b5ec966a85b4a0ba0d3c44f5d7259/src/Database/Connection.php#L680-L685) too. Otherwise, if people use DatabaseLog engine, the logs are not logged if PHP7 errors happen in transactions.

markstory:
Related to #8997 and #9500

chinpei215:
By the way, what do you think about using the `ErrorHandler` in the `ErrorHandlerMiddleware`? Right now I can see kinds of duplicate code in the two classes:
https://github.com/cakephp/cakephp/blob/668c0873986adeec2187aaf8f798d1d62306e61a/src/Error/BaseErrorHandler.php#L350
[...]

PR #11462, opened 2017-11-25
Fix PHP7 Throwable to be caught by low level try catch.

dereuromark:
With PHP7 support we should probably consider fixing the low level error handling in terms of Throwable.
Refs https://github.com/cakephp/cakephp/issues/9043#issuecomment-346676574
On PHP5 that part will just be ignored.

http://php.net/manual/de/language.errors.php7.php#119652 and https://trowski.com/2015/06/24/throwable-exceptions-and-errors-in-php7/ document this guideline if one must support PHP5 code.
In Spryker we already use it this way for this reason.

We should also check what other places might need this and can be considered "low level" enough.
Of course we do not need to add this where we know a Throwable kind of exception cannot be thrown or should not be caught (high level functionality).

### DIFF
--- a/src/Error/Middleware/ErrorHandlerMiddleware.php
+++ b/src/Error/Middleware/ErrorHandlerMiddleware.php
@@ -21,6 +21,7 @@ use Cake\Core\InstanceConfigTrait;
 use Cake\Error\ExceptionRenderer;
 use Cake\Log\Log;
 use Exception;
+use Throwable;
 
 /**
  * Error handling middleware.
@@ -93,8 +94,10 @@ class ErrorHandlerMiddleware
     {
         try {
             return $next($request, $response);
-        } catch (Exception $e) {
-            return $this->handleException($e, $request, $response);
+        } catch (Throwable $exception) {
+            return $this->handleException($exception, $request, $response);
+        } catch (Exception $exception) {
+            return $this->handleException($exception, $request, $response);
         }
     }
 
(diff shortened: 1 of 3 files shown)


src/Error/Middleware/ErrorHandlerMiddleware.php
```diff
@@ -114,18 +117,32 @@ public function handleException($exception, $request, $response)
...
+    /**
+     * @param \Psr\Http\Message\ResponseInterface $response The response
+     *
+     * @return \Psr\Http\Message\ResponseInterface A response
+     */
+    protected function handleInternalError($response)
+    {
+        $body = $response->getBody();
+        $body->write('An Internal Server Error Occurred');
+        $response = $response->withStatus(500)
```
ADmad: Nitpick: Could directly return here instead of assigning var.

dereuromark: Sometimes it is just easier to debug immutable objects with a debugger when you can set a break point before returning. But I will change it then.

markstory (approved):
Looks good to me.

chinpei215:
I think it would be better to merge this PR into 3.next. If users have custom ErrorHandler to handle PHP7 errors, it would break after the next 3.5.x. So we had a discussion about how to pass ErrorHandler to ErrorHandlerMiddleware in #11323.

dereuromark:
It cant be worse than not catching them at all. This is a bugfix that was needed IMO.

chinpei215:
Why 'at all'? In 3.5.6 or earlier, ErrorHandler would catch PHP7Errors and log the error, as wrapAndHandleException() is registered for uncaught exceptions.
https://github.com/cakephp/cakephp/blob/668c0873986adeec2187aaf8f798d1d62306e61a/src/Error/BaseErrorHandler.php#L76
https://github.com/cakephp/cakephp/blob/668c0873986adeec2187aaf8f798d1d62306e61a/src/Error/BaseErrorHandler.php#L166-L172
Additionally, the log formats are little bit different between ErrorHandler and ErrorHandlerMiddleware. For example, ErrorHandlerMiddleware's one doesn't contain IP addresses. So I think this is more than a bugfix. Also, I think it is unnecessary to change ErrorHandler::_displayException() as Errors are not passed to the method actually.

dereuromark:
_displayException() then could probably be relaxed, I agree.

dereuromark:
The logging was actually the problem, since it broke any non-file-logging (e.g. DatabaseLog plugin).
So this is a bugfix to properly be able to switch this at all to a different logger :)
Especially important if file system (default logging) is not possible, e.g. on aws and other such cloud instances.

chinpei215:
I cannot see the actual problem yet. Why does DatabaseLog not work in the current logging system? I thought ErrorHandler logs PHP7 Errors and fatal errors with DatabaseLog engine, instead of ErrorHandlerMiddleware.
(I know Connection::transactional() breaks when PHP7 Errors happens, and in such cases, DatabaseLog would not work properly because of rollback.)

PR #11463, opened 2017-11-25
Fix CS around use statements.

dereuromark:
no fqcn here inside code.


src/Database/Connection.php
```diff
@@ -223,7 +223,7 @@ public function connect()
     {
         try {
             return $this->_driver->connect();
-        } catch (\Exception $e) {
+        } catch (Exception $e) {
```
ADmad: Shouldn't a `use Exception` be added at top of file?

ADmad: I added it.

dereuromark: Weird that the Ide didnt. Thank you.

PR #11464, opened 2017-11-25
3.next - Move Http Exceptions

markstory:
This will assist in making the Http package a subsplit, and also puts the exceptions in the same package as where they are frequently used.

I'll update the migration guide once this merges.

Refs #11385


saeideng:
just remains cake\network\session\*

markstory:
@saeideng Session and cors builder will move soon as well

PR #11469, opened 2017-11-27
fixes #11468 sending user data on basic auth in API environment

### LINKED ISSUE #11468: Cake 2.x AuthComponent does not return user information when basic auth headers are sent
This is a (multiple allowed):

* [x] bug
* [ ] enhancement

* CakePHP Version: 2.10.5
* Platform and Target: Windows, Apache, MySQL

### What you did
Trying to get the user information using basic auth in API environment.

### What happened
Because the method was set as "allowed", AuthComponent would not return the user information.
`$this->Auth->user()` returns `null`, even though basic auth headers are sent.

I was expecting to be able to retrieve the user information regardless of whether or not the endpoint has authentication as mandatory.

-----
I have seen a similar issue reported here: https://github.com/cakephp/cakephp/issues/9070
But it does not seem like the pull request which resulted from it fixed this particular problem.

ghost:
Calling the protected method `$this->_getUser()` should return the requested user whenever it is available.

### DIFF
--- a/lib/Cake/Controller/Component/AuthComponent.php
+++ b/lib/Cake/Controller/Component/AuthComponent.php
@@ -298,6 +298,7 @@ class AuthComponent extends Component {
 		}
 
 		if ($this->_isAllowed($controller)) {
+			$this->_getUser();
 			return true;
 		}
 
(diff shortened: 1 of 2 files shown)


markstory:
Any way we could get a test to help ensure this stays fixed?

markstory:
Refs #11468

ADmad:
I don't think calling `_getUser()` for publicly accessible actions is a good idea. It's a significant change in behavior which can lead to unexpected results in existing apps.

For e.g. when submitting a user registration form the user would get automatically logged in due to presence of username and password fields in POST data.

markstory:
Thanks!

PR #7436, opened 2015-09-22
Only array-wrap 'order' if it's not already an array.

dereuromark:
Follow up on https://github.com/cakephp/cakephp/pull/7146

The changed tests show that it works as expected.

PR #11472, opened 2017-11-28
[2.x] Fix 'order' not working with a single expressions

chinpei215:
Prior to 2.8.0, we could pass a single expressions to the 'order' option of the `Model::find()`. But in the latest 2.x series, those are converted into arrays, and DboSource will generate incorrect ORDER clauses like the following:
```mysql
ORDER `type` expression, `value` foo 
```

Relates to #7436


dereuromark (approved):
Thx for looking into this regression.

Issue #11467, opened 2017-11-27
UrlHelper can't match with any defined routes if missing _method option


eryw:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.5.6
* Platform and Target: PHP built-in webserver

### What you did
In `routes.php` I declared a route like this:
```php
$routes->connect('/login', ['controller' => 'Users', 'action' => 'login'], ['_name' => 'login'])->setMethods(['GET', 'POST']);
```

In the template file, I uses `UrlHelper` like this:
```php
$this->Url->build(['_name' => 'login'])
```

### What happened
I get `Cake\Routing\Exception\MissingRouteException`
and an error
`Warning (2): strtoupper() expects parameter 1 to be string, array given [CORE/src/Routing/Route/Route.php, line 744]`

### What you expected to happen
It should match with the route that I have declared in `routes.php` and return `/login` string.
The error will gone if I'm also add `_method` option like this:
```php
$this->Url->build(['_name' => 'login', '_method' => 'GET'])
```
I think it should not matter what http method will be used the url to be generated is `/login`

saeideng:
do you use custom routeClass ?

saeideng:
`$url['_method'] ` changed in someplace like routeClass

markstory:
How could someone reproduce this? I've not seen `$url['_method']` ever be an array.

eryw:
@saeideng I don't use custom route class. I use built-in `DashedRoute` class.
@markstory `_method` become an array because:
- I call this method `->setMethods(['GET', 'POST'])` in `routes.php`. It will set default method to an array `['GET', 'POST']`.
- I don't provide the `_method` option when calling `Url::build()`. I don't provide it because I use named route, that what I expect is only need to pass the `_name` option.
- At this line, https://github.com/cakephp/cakephp/blob/3.5.6/src/Routing/RouteCollection.php#L316 will use default `_method` which is an array.

saeideng:
confirm 
```php
//routes.php
$routes->connect('/login', ['controller' => 'Users', 'action' => 'login'], ['_name' => 'login'])->setMethods(['GET', 'POST']);

//view
$this->Url->build(['_name' => 'login'])

saeideng:
I can care to this

markstory:
Thanks @saeideng

markstory:
Closing as a pull request is open now.

[...]

PR #11474, opened 2017-11-28
accept array of "_method"s  on route::match()

saeideng:
array of _method
`"_method" => ['PUT ','POST']`  
comes from
https://github.com/cakephp/cakephp/blob/c159efc166a7936ef96f8e9e22dd7543d4e9ce61/src/Routing/RouteCollection.php#L316
when someone set 
```php
$routes->connect('/login', ['controller' => 'Users', 'action' => 'login'], ['_name' => 'login'])->setMethods(['PUT', 'POST']);
or
$routes->connect('/login', ['controller' => 'Users', 'action' => 'login', '_method' => ['PUT', 'POST']], ['_name' => 'login'])
```
`$url + $route->defaults` just merge when someone set `_name` now 
but probably we need to this for non-named route (in other PR)

Refs : https://github.com/cakephp/cakephp/issues/11467

### DIFF
--- a/src/Routing/Route/Route.php
+++ b/src/Routing/Route/Route.php
@@ -741,11 +741,14 @@ class Route
         if (empty($url['_method'])) {
             return false;
         }
-        if (!in_array(strtoupper($url['_method']), (array)$this->defaults['_method'])) {
-            return false;
+        $methods = array_map('strtoupper', (array)$url['_method']);
+        foreach ($methods as $value) {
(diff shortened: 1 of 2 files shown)


markstory:
Why would we need to merge the defaults for routes that are matched without names? By merging the default parameters we could cause false matches.

saeideng:
Ok

saeideng:
For this PR i think we can even remove default[_name] key before sending/meging into match()

PR #11475, opened 2017-11-29
Split ValidationSet setters.

robertpustulka:
Deprecate `ValidationSet::isPresenceRequired()` and `ValidationSet::isEmptyAllowed()` as setters.

The method name for a getter is fine, so only the parameters are deprecated. I'm not sure this is a good way to do it though.

### DIFF
--- a/tests/TestCase/Validation/ValidationSetTest.php
+++ b/tests/TestCase/Validation/ValidationSetTest.php
@@ -19,6 +19,7 @@ namespace Cake\Test\TestCase\Validation;
 use Cake\TestSuite\TestCase;
 use Cake\Validation\ValidationRule;
 use Cake\Validation\ValidationSet;
+use PHPUnit\Framework\Error\Deprecated;
 
 /**
  * ValidationSetTest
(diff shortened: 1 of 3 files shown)


markstory:
I think the approach is pretty good and results in a similar end state to what we have with other split methods. It would be nice to keep a test for the deprecated behaviour though.

robertpustulka:
@markstory tests added

tests/TestCase/Validation/ValidationSetTest.php
```diff
@@ -191,20 +192,129 @@ public function testCount()
...
+    /**
+     * Test isPresenceRequired deprecated setter
+     *
+     * @group deprecated
+     * @return void
+     */
+    public function testRequirePresenceDeprecated()
+    {
+        $this->deprecated(function () {
+            $set = (new ValidationSet);
```
dereuromark: Why these bracing styles? Shouldnt it be `= new ValidationSet()`?

robertpustulka: Copy-pasta :)

dereuromark: Sounds like a sniff is missing for consistent newable :)

markstory:
Looks great! I will get this merged after I get the migration guide updated.

PR #11478, opened 2017-11-29
Bump up to phpstan 0.9

ADmad:


ADmad:
@ondrejmirtes Can you help fix this error https://travis-ci.org/cakephp/cakephp/jobs/308996496#L798 ? :slightly_smiling_face:

src/ORM/Query.php
```diff
@@ -424,7 +424,7 @@ public function clearContain()
      * Used to recursively add contained association column types to
      * the query.
      *
-     * @param \Cake\ORM\Table|\Cake\Datasource\RepositoryInterface $table The table instance to pluck associations from.
```
dereuromark: The interface should stay for where it is used (instead of concrete one). Both are in use currently.

src/TestSuite/Fixture/TestFixture.php
```diff
@@ -264,8 +264,8 @@ protected function _schemaFromReflection()
     /**
      * Gets/Sets the TableSchema instance used by this fixture.
      *
-     * @param \Cake\Database\Schema\TableSchema|\Cake\Database\Schema\TableSchemaInterface|null $schema The table to set.
-     * @return \Cake\Database\Schema\TableSchema|\Cake\Database\Schema\TableSchemaInterface|null
+     * @param \Cake\Database\Schema\TableSchema|null $schema The table to set.
```
dereuromark: Same here!

ondrejmirtes:
Hi, I tested PHPStan 0.9 on CakePHP so I'm aware of the problems 😊 I've wanted to send you a PR with the fixes but you were faster 😊 I think you can fix it like this:

```diff
diff --git a/tests/PHPStan/AssociationTableMixinClassReflectionExtension.php b/tests/PHPStan/AssociationTableMixinClassReflectionExtension.php
index bc2282b9d2..627f292b94 100644
--- a/tests/PHPStan/AssociationTableMixinClassReflectionExtension.php
+++ b/tests/PHPStan/AssociationTableMixinClassReflectionExtension.php
@@ -57,7 +57,7 @@ class AssociationTableMixinClassReflectionExtension implements PropertiesClassRe
      */
     public function getMethod(ClassReflection $classReflection, string $methodName): MethodReflection
     {
-        return $this->getTableReflection()->getMethod($methodName);
+        return $this->getTableReflection()->getNativeMethod($methodName);
     }
 
     /**
@@ -81,6 +81,6 @@ class AssociationTableMixinClassReflectionExtension implements PropertiesClassRe
      */
     public function getProperty(ClassReflection $classReflection, string $propertyName): PropertyReflection
     {
-        return $this->getTableReflection()->getProperty($propertyName);
+        return $this->getTableReflection()->getNativeProperty($propertyName);
     }
 }
```

src/Utility/Xml.php
```diff
@@ -97,7 +97,7 @@ class Xml
      * If using array as input, you can pass `options` from Xml::fromArray.
      *
      * @param string|array $input XML string, a path to a file, a URL or an array
-     * @param string|array $options The options to use
+     * @param array $options The options to use
```
dereuromark: Invalid change, input is still possibly string, see code.

ondrejmirtes:
Also, I'm not sure what to do about `Class PHPUnit_Runner_Version not found and could not be autoloaded.`. Where does the class come from? We should probably fix it by defining a fake one in your PHPStan's bootstrap.

src/ORM/Query.php
ADmad: The method has call to `$table->association()` and `RepositoryInterface` doesn't have an `association()` method, so how is it valid?

src/TestSuite/Fixture/TestFixture.php
ADmad: Then the docblock should be updated in future when the typhint is changed.

ADmad:
> Also, I'm not sure what to do about Class PHPUnit_Runner_Version not found and could not be autoloaded.. Where does the class come from? We should probably fix it by defining a fake one in your PHPStan's bootstrap.

@ondrejmirtes We have some class aliasing hackery to support phpunit 5.7 and 6, probably that's what causing this.

ADmad:
@ondrejmirtes For e.g. https://github.com/cakephp/cakephp/blob/master/src/TestSuite/Constraint/EventFired.php#L4

I wonder if using `class_exists()` with 2nd param `false` would fix the issue while keeping the required functionality working.

ADmad:
We have cases like:
```php
if (method_exists($foo, 'someMethod')) {
    $foo->someMethod();
}
```
for which phpstan shows error like `Call to an undefined method object::someMethod()`. What's the proper method to deal with that?

src/TestSuite/Fixture/TestFixture.php
dereuromark: Fair enough then..

src/ORM/Query.php
dereuromark: OK, the classes that use the interface approach will then all be "red". Not sure what the worse code is then :/ I guess we have to live with it either way we chose that something will "fail".

markstory: >  Bummer that we cannot do faster majors to actually clean up those smells.

I hope that the 4->5 journey isn't filled with as many deprecations and method renames as well. We've churned huge parts of the framework API in order to get stricter typehinting. I hope to not have to do that again :smil:

dereuromark: > churned huge parts of the framework API 

The main issue I have is the wrong interface usage everywhere and that you cant fix it BC, thus we carry this baggage always for years till it can be resolved.
But sure, API regarding methods and their interface definition is also a problem for such long release cycles.

stmeyer:
@ADmad @ondrejmirtes 
If I add
```
parameters:
	autoload_files:
		- %rootDir%/../../cakephp/cakephp/src/TestSuite/IntegrationTestCase.php
```
to my local app the error `Class PHPUnit_Runner_Version not found and could not be autoloaded.` is gone. Maybe this helps :-)

ADmad:
@stmeyer Added, thanks.

ADmad:
@ondrejmirtes Are inline `@var` tags not supported anymore? For e.g. check

https://github.com/cakephp/cakephp/blob/4b10e9c559ad0be317209642173938279f3ecee2/src/Controller/Component/SecurityComponent.php#L102-L103

In spite of the `@var` tag for `$controller` variable, phpstan throws error `Access to an undefined property object::$request` because the return type of `Event::getSubject()` is declared as `object`.

ondrejmirtes:
@ADmad Hi, the only supported format is `/** @var Type $controller */`. You're missing a `*` in `/*`.

ADmad:
@ondrejmirtes Just `/* ` used to work in phpstan 0.8 :slightly_smiling_face:. Our coding standards actually prevents using `/** ` for inline declarations so have a dilemma.

ondrejmirtes:
@ADmad I might add support in a later version of PHPStan (should be really easy), but please check - https://docs.phpdoc.org/guides/docblocks.html - there's no word about that docblocks can be in `/*`, so you should probably reconsider your coding standard to be more correct :)

ADmad:
Yeah upgrading our coding standard would probably be the best. Just that it means more work to get using phpstan 0.9 :stuck_out_tongue:

ondrejmirtes:
As I said, I'm willing to return the compatibility for `/*`, shouldn't be hard, but it will take a few days or weeks :)

dereuromark:
@ondrejmirtes I'd rather fix the sniffer to use proper format :) That sniffer exists and could easily be ported and enabled.

ADmad:
@ondrejmirtes Inline `@var` isn't supported for interfaces and traits? 

Getting error `Call to an undefined method object::toArray()` is spite of having https://github.com/cakephp/cakephp/pull/11478/files#diff-8b0f4da31563194a4b4f024bb4646c56R163 

and `Call to an undefined method DateTimeInterface::i18nFormat()` for https://github.com/cakephp/cakephp/pull/11478/files#diff-1181b84458a7833c6bf32f5e38b28a53R118

ondrejmirtes:
@ADmad `@var` comment is only supported above an assignment statement:

```php
/** @var Foo $var */
$var = getFoo();
```

That has been true since the beginning, even in pre-0.9 times 😊 It's also supported for iterated item above a foreach:

```php
/** @var Foo $foo */
foreach ($foos as $foo) {

}
```

BTW it's really weird to say that some piece of code is a trait (an implementation detail) - an interface (a contract) would be better. You shouldn't force someone to use a trait, but to implement an interface is fine and preferred.

ondrejmirtes:
And about the first example:

```php
/** @var \Cake\Datasource\EntityInterface $item */
$item = $item->toArray();
```

This does not work because the statement changes `$item` below the assignment.

ADmad:
@ondrejmirtes Thank you for your clarifications. Much appreciated.

src/I18n/RelativeTimeFormatter.php
```diff
@@ -115,7 +115,6 @@ public function timeAgoInWords(DateTimeInterface $time, array $options = [])
         }
 
         if ($diff > abs($now - (new FrozenTime($options['end']))->format('U'))) {
-            /** @var \Cake\I18n\DateFormatTrait $time */
```
dereuromark: We shouldnt remove them, but then assign them, otherwise the method is not followable or clickable.
Also IDE cannot detect for us if the argument count or type is OK here.

ADmad: This inline docComment doesn't exist on master either, it was added on this branch itself to try to make phpstan happy.

dereuromark: I see.
Maybe it was added somewhere else then in a similar way.
Well, we could still keep it here too :)

ADmad:
@ondrejmirtes Sorry to bug you again but can you please help fix these last 3 remaining errors https://travis-ci.org/cakephp/cakephp/jobs/315337836#L685 ?

The code is perfectly valid since all calls are wrapped in `method_exists` checks. I can't simply ignore the error as the messages are too generic and could hide actual problems in future.

ondrejmirtes:
@ADmad I'm not able to fix them right now - I think it's better to ignore them for now and start using the latest PHPStan rather than postponing integration and miss some errors because of that? I will fix it in some new version in the upcoming weeks but not right now...

dereuromark:
@ondrejmirtes On a related note: I saw some "clone" code also not respecting the types then, failing in phpstan. I would expect a clone to completely pass over the typehinting to the new variable.

ondrejmirtes:
@dereuromark I'll look into that.

src/View/Form/EntityContext.php
```diff
@@ -147,7 +148,7 @@ protected function _prepare()
             $table = TableRegistry::get($table);
         }
 
-        if (!is_object($table)) {
+        if (!($table instanceof RepositoryInterface)) {
```
dereuromark: Could this be a BC break for some apps?

dereuromark (approved):
Nice work!

src/View/Form/EntityContext.php
ADmad: The object comes from `TableRegistry::get()` and later there's a call `$table->getAlias()`, so extremely unlikely it's not going to be an `RepositoryInterface` instance.

ADmad:
For the phpcs build failure our codesniffer needs to be updated.

ADmad:
Actually, I mucked something up inside `Marshaller`.

dereuromark:
Let's fix the sniffer first so the builds will stay green.

ADmad:
@dereuromark The 3 inline docComment phpcs is complaining about are inside closures. I don't think keeping them is going to help IDEs so I am thinking of just removing them. They were most likely only added earlier to make phpstan happy.

dereuromark:
@ADmad They are needed to follow the method, but they probably can also be outside of it, as the variable is passed in 1:1. So whereever assigned would suffice afaik.

ADmad:
Reverted and made a new release for our codesniffer.

ADmad:
Now :pray: for :white_check_mark:

ADmad:
:tada:

PR #11479, opened 2017-11-29
Clarify migration path to 3.x

dereuromark:
saveField() uses $this->id on the model itself which is discouraged.

What do you think about a more clear migration path for those still on the last 2.x version?
Many always ask how to do that, and still use this a lot in 2.x code, even though they might very well be upgrading that code.
It would clarify to do the same we do forward also in a backport way and more clearly express certain API to be not useful to be used for new code.

There are few more we could clarify.


markstory:
> What do you think about a more clear migration path for those still on the last 2.x version?

If you think documentation updates would help, I think its worth doing.

Issue #11481, opened 2017-11-30
Throw a warning if a cookie exceeds the recommended size


burzum:
This is a (multiple allowed):

* [ ] bug
* [x] enhancement
* [ ] feature-discussion (RFC)

As discussed yesterday with @markstory we might want to check the size of the cookies.

See https://stackoverflow.com/questions/640938/what-is-the-maximum-size-of-a-web-browsers-cookies-key

This piece of code in the CookieCollection::addToRequest() should do it:

```php
        foreach ($cookies as $key => $value) {
            $cookie = sprintf("%s=%s", rawurlencode($key), rawurlencode($value));
            $size = mb_strlen($cookie);
            if ($size > 4096) {
                triggerWarning(sprintf(
                    'The cookie `%s` exceeds the recommended maximum cookie length of 4096 bytes.',
                    $cookie['name']
                ));
            }
            $cookiePairs[] = $cookie;
        }
```

burzum:
Closing this PR is open.

PR #11482, opened 2017-11-30
Implement a warning when a cookie exceeds the recommended size

burzum:
Refs #11481

https://stackoverflow.com/questions/640938/what-is-the-maximum-size-of-a-web-browsers-cookies-key

### DIFF
--- a/src/Http/Cookie/CookieCollection.php
+++ b/src/Http/Cookie/CookieCollection.php
@@ -223,8 +223,17 @@ class CookieCollection implements IteratorAggregate, Countable
         $cookies = array_merge($cookies, $extraCookies);
         $cookiePairs = [];
         foreach ($cookies as $key => $value) {
-            $cookiePairs[] = sprintf("%s=%s", rawurlencode($key), rawurlencode($value));
+            $cookie = sprintf("%s=%s", rawurlencode($key), rawurlencode($value));
+            $size = mb_strlen($cookie);
+            if ($size > 4096) {
+                triggerWarning(sprintf(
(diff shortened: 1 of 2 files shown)


tests/TestCase/Http/Cookie/CookieCollectionTest.php
```diff
@@ -373,6 +374,23 @@ public function testAddToRequestNoCookies()
...
+    /**
+     * Testing the cookie size limit warning
+     *
+     * @expectedException \PHPUnit\Framework\Error\Warning
+     * @expectedExceptionMessage The cookie `default` exceeds the recommended maximum cookie length of 4096 bytes.
+     * @return void
+     */
+    public function testCookieSizeWarning()
+    {
+        $string = Security::insecureRandomBytes(9000);
```
dereuromark: Wouldn't 5000 suffice then? But fine either way.

src/Http/Cookie/CookieCollection.php
```diff
@@ -223,8 +223,17 @@ public function addToRequest(RequestInterface $request, array $extraCookies = []
         $cookies = array_merge($cookies, $extraCookies);
         $cookiePairs = [];
         foreach ($cookies as $key => $value) {
-            $cookiePairs[] = sprintf("%s=%s", rawurlencode($key), rawurlencode($value));
+            $cookie = sprintf("%s=%s", rawurlencode($key), rawurlencode($value));
+            $size = mb_strlen($cookie);
```
markstory: I think you want `strlen()` here, as you need to count bytes not characters.

PR #11483, opened 2017-11-30
Add confirm message for form buttons.

dereuromark:
Inside normal HTML using postLink() and confirm messages work fine.
But sometimes, working with a small actual form, you also would like to easily have confirm messages to confirm the form execution with "Are you sure..."?
```php
<?= $this->Form->button(__('Finalize'), ['confirm' => 'Make sure ... before you continue.'] ?>
<?= $this->Form->end() ?>
```

Also works for postButton() then, of course, as standalone forms:
```php
<?= $this->Form->postButton('Click', ['action' => 'foo', 'bar'], ['confirm' => 'Yeah?']) ?>
```

Not sure about the "old" `Form->input()` though..

### DIFF
--- a/src/View/Helper/FormHelper.php
+++ b/src/View/Helper/FormHelper.php
@@ -1730,6 +1730,7 @@ class FormHelper extends Helper
      * ### Options:
      *
      * - `escape` - HTML entity encode the $title of the button. Defaults to false.
+     * - `confirm` - Confirm message to show. Form execution will only continue if confirmed then.
      *
      * @param string $title The button's caption. Not automatically HTML encoded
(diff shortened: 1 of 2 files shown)


ADmad:
Not really thrilled about adding more support for inline JS :(

dereuromark:
@ADmad As long as we do not have a good concept to move away from the already existing functionality, completing it is a must-have IMO :)

markstory:
I share @ADmad's thoughts on inline javascript, but also recognize we don't have a better solution right now. Hopefully in a future release we can replace the inline JS with event bindings.

PR #11484, opened 2017-11-30
Add Table::addBehaviors() method.

sdustinh:
This method mirrors Table::addAssociations() and lets you add many behaviors at the same time.

### DIFF
--- a/src/ORM/Table.php
+++ b/src/ORM/Table.php
@@ -861,6 +861,38 @@ class Table implements RepositoryInterface, EventListenerInterface, EventDispatc
     }
 
     /**
+     * Add Many Behaviors
+     *
+     * Adds an array of behaviors to the table's behavior collection.
+     *
+     * Example:
+     *
+     * ```
+     * $this->addBehaviors([
+     *      'Timestamp',
+     *      'Tree' => ['level' => 'level'],
+     * ]);
+     * ```
+     *
+     * @param array $behaviors All of the behaviors to load.
+     * @return $this
+     * @throws \RuntimeException If a behavior is being reloaded.
+     */
+    public function addBehaviors(array $behaviors)
+    {
+        foreach ($behaviors as $name => $options) {
+            if (is_int($name) === true) {
+                $name = $options;
+                $options = [];
+            }
+
+            $this->addBehavior($name, $options);
+        }
+
+        return $this;
+    }
+
+    /**
      * Removes a behavior from this table's behavior registry.
      *
      * Example:
--- a/src/ORM/Table.php
+++ b/src/ORM/Table.php
@@ -861,8 +861,6 @@ class Table implements RepositoryInterface, EventListenerInterface, EventDispatc
     }
 
     /**
-     * Add Many Behaviors
-     *
      * Adds an array of behaviors to the table's behavior collection.
      *
      * Example:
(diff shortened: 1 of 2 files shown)


src/ORM/Table.php
```diff
@@ -860,6 +860,38 @@ public function addBehavior($name, array $options = [])
...
+     * ```
+     *
+     * @param array $behaviors All of the behaviors to load.
+     * @return $this
+     * @throws \RuntimeException If a behavior is being reloaded.
+     */
+    public function addBehaviors(array $behaviors)
+    {
+        foreach ($behaviors as $name => $options) {
+            if (is_int($name) === true) {
```
dereuromark: `if (is_int(...))` could suffice (the return value is 100% bool always).

src/ORM/Table.php
```diff
@@ -860,6 +860,38 @@ public function addBehavior($name, array $options = [])
         return $this;
     }
 
+    /**
+     * Add Many Behaviors
```
dereuromark: This is redundant with the next sentence.

dereuromark:
Fair enough :)

Even though chaining is also not too much more:

    $this->addBehavior('One', [...])
        ->addBehavior('Two', [...])
        ->addBehavior('Three', [...])

The benefit here is full IDE typehinting and auto-complete of it (at least using latest [IdeHelper](https://github.com/dereuromark/cakephp-ide-helper/) plugin).

src/ORM/Table.php
sdustinh: Affirm. I'll update. It's just habit. I try to use strict comparison as much as possible.

saeideng:
why you need `addBehaviors` when you can 
```php

**Comment by dereuromark** [comment]
@saeideng It seems to be a matter of choice here probably. No big deal to have it probably.
Also easier for plugins like Shim migrate old $behaviors property etc maybe instead of looping.

**Comment by sdustinh** [comment]
It's completely a preference addition. For me the big benefit is in readability:

```php
public function initialize(array $config)
{
    $this->addAssociations([
        'belongsTo' => [
            'One' => [
                'foreignKey' => 'one_id',
                'propertyName' => 'first',
            ],
            'Two' => [
                'foreignKey' => 'two_id',
                'propertyName' => 'second',
            ],
        ],
    ]);
    $this->addBehaviors([
        'Sluggable',
        'Timestamp',
        'NoUpdateFields' => [
            'fields' => ['one_id', 'two_id'],
        ],
        'Tree' => ['level' => 'level'],
    ]);

    // vs...
    $this->belongsTo('One')
        ->setForeignKey('one_id')
        ->setProperty('first');
    $this->belongsTo('Two')
        ->setForeignKey('two_id')
        ->setProperty('second');
    $this->addBehavior('Sluggable');
    $this->addBehavior('Timestamp');
    $this->addBehavior('NoUpdateFields', [
        'fields' => ['one_id', 'two_id'],
    ]);
    $this->addBehavior('Tree', [
        'level' => 'level',
    ]);
}
```
The tables that I have with a handful of behaviors and associations creates a huge wall of text. The array format is easier to read. Plus I can do something like this:
```php
public function initialize(array $config)
{
    $this->addAssociations(self::ASSOCIATIONS);
    $this->addBehaviors(self::BEHAVIORS);
}
```

dereuromark (approved):

PR #11487, opened 2017-12-01
3.next- Deprecate Query::orWhere().

markstory:
I made time to circle back on this as I'd really like to not have this method in the future. It is easy to replace after I thought through the problem a bit more.

Refs #11385

### DIFF
--- a/src/Database/Dialect/TupleComparisonTranslatorTrait.php
+++ b/src/Database/Dialect/TupleComparisonTranslatorTrait.php
@@ -78,15 +78,15 @@ trait TupleComparisonTranslatorTrait
             $value = [$value];
         }
 
+        $conditions = ['OR' => []];
         foreach ($value as $tuple) {
-            $surrogate->orWhere(function ($exp) use ($fields, $tuple) {
-                foreach (array_values($tuple) as $i => $value) {
-                    $exp->add([$fields[$i] => $value]);
-                }
-
-                return $exp;
-            });
+            $item = [];
+            foreach (array_values($tuple) as $i => $value) {
+                $item[] = [$fields[$i] => $value];
+            }
+            $conditions['OR'][] = $item;
         }
+        $surrogate->where($conditions);
 
         $expression->setField($true);
         $expression->setValue($surrogate);
--- a/src/Database/Query.php
+++ b/src/Database/Query.php
@@ -811,6 +811,10 @@ class Query implements ExpressionInterface, IteratorAggregate
      *
      * `$query->where(['OR' => [['published' => false], ['published' => true]])`
      *
+     * Would result in:
+     *
+     * `WHERE (published = false) OR (published = true)`
+     *
      * Keep in mind that every time you call where() with the third param set to false
      * (default), it will join the passed conditions to the previous stored list using
      * the `AND` operator. Also, using the same array key twice in consecutive calls to
@@ -1008,6 +1012,10 @@ class Query implements ExpressionInterface, IteratorAggregate
      */
     public function orWhere($conditions, $types = [])
     {
+        deprecationWarning(
+            'Query::orWhere() is deprecated as it creates hard to predict SQL based on the ' .
+            'current query state. Use `Query::where()` instead.'
+        );
         $this->_conjugate('where', $conditions, 'OR', $types);
 
         return $this;
--- a/src/Database/README.md
+++ b/src/Database/README.md
@@ -240,9 +240,6 @@ $query->where(['id >' => 1, 'title' => 'My title']);
 It is possible to generate `OR` conditions as well
 
 ```php
-$query->where(['id >' => 1])->orWhere(['title' => 'My Title']);
-
-// Equivalent to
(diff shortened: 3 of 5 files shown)


jeremyharris:
It would be nice to see an option suggested in the deprecation notes. I tried replacing a simple orWhere with a typical "or" but it required writing a bunch of array conditions reminding me of cake 2. I'll try and post an example but I'm on my phone atm.

To be clear, I agree that the original problem should be fixed, but there should be a clear path towards migration.

jeremyharris:
Okay, so before I write some docs, I want to make sure I'm doing this right.

This is a finder that finds a User's ContactTypes. If they are not a "pro" subscriber, they only get their built in tools. The "or" comes in when they are editing a Contact (Contacts belongsTo ContactTypes) that had a non-built tool. This happens when the user was at one time pro but is no longer -- I want the ContactTypes dropdown to still have the originally selected item. 

Not sure all that explanation is necessary, but here we go:

*before*
```
public function findForUser(Query $query, array $options)
    {
        $options += [
            'contact' => null,
            'user' => null,
        ];

        if (!($options['user'] instanceof User)) {
            throw new RuntimeException('Cannot get contact types list without defining "user" in options.');
        }

        $query->where([
            'user_id' => $options['user']->id
        ]);

        if (!$options['user']->is('pro')) {
            $query->where([
                'is_built_in' => true
            ]);
        }

        if ($options['contact'] instanceof Contact && !$options['contact']->isNew()) {
            $query->orWhere([
                'id' => $options['contact']->contact_type_id,
            ]);
        }

        return $query;
    }
```

*after* 
```
public function findForUser(Query $query, array $options)
    {
        $options += [
            'contact' => null,
            'user' => null,
        ];

        if (!($options['user'] instanceof User)) {
            throw new RuntimeException('Cannot get contact types list without defining "user" in options.');
        }

        $conditions = [
            'user_id' => $options['user']->id,
        ];

        if (!$options['user']->is('pro')) {
            $conditions['is_built_in'] = true;
        }

        if ($options['contact'] instanceof Contact && !$options['contact']->isNew()) {
            $conditions = [
                'or' => [
                    $conditions,
                    'id' => $options['contact']->contact_type_id,
                ]
            ];
        }

        return $query->where($conditions);
    }
```

As you can see the after is harder to read and understand, at least for me. I tried using the `or_` expression without success. Is this what a query like this should look like now? I don't want to write docs if I'm doing it wrong 😄

markstory:
That is one way to rewrite that query conditions. I think the following may also work:

```php
public function findForUser(Query $query, array $options)
{
    $options += [
        'contact' => null,
        'user' => null,
    ];

    if (!($options['user'] instanceof User)) {
        throw new RuntimeException('Cannot get contact types list without defining "user" in options.');
    }

    $query->where(function ($exp, $query) use ($options) {
        $basic = $exp->and_(['user_id' => $options['user']->id]);
        if (!$options['user']->is('pro')) {
            $basic->add(['is_built_in' => true]);
        }

        if ($options['contact'] instanceof Contact && !$options['contact']->isNew()) {
            return $exp->or_([
                $basic,
                'id' => $options['contact']->contact_type_id
            ]);
        }

        return $basic;
    });

    return $query;
}
```

I've not run this code, but it may cover your scenario.

jeremyharris:
Okay, that works as well. Looks pretty much like the conditions array but with an array of expressions instead of the 'or' key. Hmm.

Which would you prefer to see in the docs, the 'or' key or the array of expressions?

markstory:
I think we should show the expression based example. It gives better IDE support and shows more of the query builder features.

inoas:
Another benefit of the _or() expression is that with the 'OR' key way you can run into problems with duplicate hash-keys replacing themselves if you iterate different conditions but on the right side but then same "identifier" on the left side.

arilia:
I don't want to resurrect orWhere (I never actually liked it) but there is a situation where I can't see how I can replace it: when I want to add an OR condition to a custom finder i.e.

    $query = $this->Users->find('veryComplexFinder')
        ->orWhere(['role' => 'admin']);

in this way I used to get all the 'veryComplexFinder' users OR the admins but now of course I get the warning

I can't find a way to put the conditions set by the finder in OR with the new condition

xaphalanx:
In cases where you don't know all your where "or" conditions up front, there doesn't appear to be a replacement. I have a report builder where that is the case.  Multiple custom finders is another.

inoas:
What should happen instead is having aliases/identifiers for expressions and then having `addClause()` or similar so you could add clauses to say not, or or blocks.

arilia:
that's how I did. Guess there is a simpler method


    $query->where(function($exp, $q) use ($options) {
            $exp = $exp
                ->or_([
                    'role' => 'admin',
                    $q->find('veryComplexFinder', $options)->clause('where')
                ]);
            return $exp;
        });

this works in my case because i know I'm applying my veryComplexFinder, but there is still the case where I have to append the or without knowing what finders have been applied before

Issue #11485, opened 2017-11-30
Unable to send email with DKIM signature


LifeOrYou:
* [x] bug
* [x] enhancement
* [ ] feature-discussion (RFC)

Cakephp version : 3.4.13

### What you did
I want to send a email with DKIM signature but my email client can't verify this signature (body hash did not verify).

### What you expected to happen
For solve it, i must force the content-transfer-encoding to `quoted-printable` not `8bit` or `7bit`.
But i must override cakephp source code for set content-transfer-encoding to `quoted-printable`.
```
protected function _getContentTransferEncoding()
    {
        return 'quoted-printable';
        // Cakephp source code
        $charset = strtoupper($this->charset);
        if (in_array($charset, $this->_charset8bit)) {
            return '8bit';
        }

        return '7bit';
    }
```

How i can set the transfer encoding whitout override cakephp source code ?

markstory:
> How i can set the transfer encoding whitout override cakephp source code ?

You'd have to extend the Email interface to expose methods to allow the transfer encoding to be changed.

markstory:
Related to #10789

markstory:
In #10789 there was a suggestion of adding a `setEncoding()` method. No one followed that up with a pull request. If you need this feature @LifeOrYou a pull request would be merged for that kind of thing.

markstory:
Closing as #11488 was merged.

PR #11488, opened 2017-12-01
Add email config for transferEncoding #11485

maratth:
Add getter/setter for transferEncoding and modify the `_getContentTranferEncoding` for use transferEncoding if is set.

Write test in `EmailTest` for check it.

### DIFF
--- a/src/Mailer/Email.php
+++ b/src/Mailer/Email.php
@@ -247,7 +247,7 @@ class Email implements JsonSerializable, Serializable
      *
      * @var string|null
      */
-    public $transferEncoding;
+    protected $transferEncoding;
 
     /**
      * The application wide charset, used to encode headers and body
--- a/src/Mailer/Email.php
+++ b/src/Mailer/Email.php
@@ -250,6 +250,19 @@ class Email implements JsonSerializable, Serializable
     protected $transferEncoding;
 
     /**
+     * Available encoding to be set for transfer.
+     *
+     * @var array
+     */
+    protected $_transferEncodingAvailable = [
+        '7bit',
+        '8bit',
+        'base64',
+        'binary',
+        'quoted-printable'
+    ];
+
+    /**
      * The application wide charset, used to encode headers and body
      *
      * @var string|null
@@ -843,6 +856,15 @@ class Email implements JsonSerializable, Serializable
      */
     public function setTransferEncoding($encoding)
     {
+        $encoding = strtolower($encoding);
+        if (!in_array($encoding, $this->_transferEncodingAvailable)) {
+            throw new InvalidArgumentException(
+                sprintf(
+                    'Transfer encoding not available. Can be : %s.',
+                    implode(', ', $this->_transferEncodingAvailable)
+                )
+            );
+        }
         $this->transferEncoding = $encoding;
 
         return $this;
--- a/src/Mailer/Email.php
+++ b/src/Mailer/Email.php
@@ -836,9 +836,9 @@ class Email implements JsonSerializable, Serializable
     }
 
     /**
-     * TransportCharset setter.
+     * TransferEncoding setter.
      *
-     * @param string|null $encoding Character set.
+     * @param string|null $encoding Encoding set.
      * @return $this
      */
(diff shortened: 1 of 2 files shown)


tests/TestCase/Mailer/EmailTest.php
```diff
@@ -2487,6 +2497,23 @@ public function testHeaderCharset()
...
 
+    /**
+     * Test transferEncoding
+     *
+     * @return void
+     */
+    public function testTransferEncoding(){
+        // Test new transfer encoding
+        $this->Email->setTransferEncoding('quoted-printable');
+        $this->assertSame($this->Email->getTransferEncoding(), 'quoted-printable');
```
dereuromark: Order is wrong, should be: expected, result.

src/Mailer/Email.php
```diff
@@ -241,6 +241,14 @@ class Email implements JsonSerializable, Serializable
      */
     public $headerCharset;
 
+    /**
+     * The email transfer encoding used.
+     * If null, the $charset property is used for determined the transfer encoding.
+     *
+     * @var string|null
+     */
+    public $transferEncoding;
```
robertpustulka: Does it make sense to add another public properties and a pair of getter/setter for that?

LifeOrYou: I think no, but i follow what was done. 
You can see the charset or headerCharset properties.

robertpustulka: Those are public for BC reasons I suppose. I think we can make the new property protected.

tests/TestCase/Mailer/EmailTest.php
LifeOrYou: Hum, sorry i fix this in #a006420
But i see, the tests above have the same issue.

src/Mailer/Email.php
ADmad: Yeah the property should be protected.

LifeOrYou: Ok i fix it

src/Mailer/Email.php
```diff
@@ -827,6 +835,29 @@ public function headerCharset($charset = null)
...
 
+    /**
+     * TransferEncoding setter.
+     *
+     * @param string|null $encoding Encoding set.
+     * @return $this
+     */
+    public function setTransferEncoding($encoding)
+    {
+        $this->transferEncoding = $encoding;
```
markstory: Should we validate that for a valid/known transfer encoding?

LifeOrYou: I don't know all content transfer encoding. But I found [this](https://www.w3.org/Protocols/rfc1341/5_Content-Transfer-Encoding.html). 
I understand that there is `BASE64 / QUOTED-PRINTABLE / 8BIT / 7BIT / BINARY / x-token` but I'm not sure, I'm not a expert.

If you think is good, I can implement it.

markstory: I think checking the standard types (not `x-token`) is a good place to start.

LifeOrYou:
Is possible add this pull request for cakephp v3.4.x or is only for 3.5.x ?

markstory:
> Is possible add this pull request for cakephp v3.4.x or is only for 3.5.x ?

We're only doing security fixes for 3.4 at this point in time. The upgrade from 3.4 to 3.5 should be pretty painless if you need this in your app.

src/Mailer/Email.php
```diff
@@ -827,6 +848,38 @@ public function headerCharset($charset = null)
         return $this->headerCharset;
     }
 
+    /**
+     * TransferEncoding setter.
+     *
+     * @param string|null $encoding Encoding set.
+     * @return $this
+     */
+    public function setTransferEncoding($encoding)
```
markstory: Should we be handling encoding the message based on the indicated transfer encoding? Right now someone could set the transfer encoding to base64, but the email message will still be in 8bit (generally).

LifeOrYou: I think than that two differents parts. I don't know if have a link between transfer encoding and message encoding.

markstory: Ok.

PR #11491, opened 2017-12-02
Add DebugInfo and deprecate bad object usage.

dereuromark:
`__debugInfo()` doest help with the `dd(Type::buildAll())` bug, but it does output only the relevant info when doing dd($type).

Also tweaked performance by caching the object, and fixed up doc blocks.
The _types also can already contain objects instead of just class names.
map() itself cannot take objects unless it is an array of it.

In 3.6 we should also split that super-method into getMap and setMap parts, it does way too much and that pretty badly.

### DIFF
--- a/src/Database/Type.php
+++ b/src/Database/Type.php
@@ -313,8 +313,6 @@ class Type implements TypeInterface
     {
         return [
             'name' => $this->_name,
-            'types' => static::$_types,
-            'builtTypes' => static::$_builtTypes,
         ];
     }
 }
--- a/src/Database/Type.php
+++ b/src/Database/Type.php
@@ -106,9 +106,7 @@ class Type implements TypeInterface
             throw new InvalidArgumentException(sprintf('Unknown type "%s"', $name));
         }
         if (is_string(static::$_types[$name])) {
-            static::$_types[$name] = new static::$_types[$name]($name);
-
-            return static::$_builtTypes[$name] = static::$_types[$name];
+            return static::$_builtTypes[$name] = new static::$_types[$name]($name);
         }
 
         return static::$_builtTypes[$name] = static::$_types[$name];
@@ -145,6 +143,9 @@ class Type implements TypeInterface
      * Registers a new type identifier and maps it to a fully namespaced classname,
      * If called with no arguments it will return current types map array
      * If $className is omitted it will return mapped class for $type
+     *
+     * Deprecated: The usage of $type as \Cake\Database\Type[] is deprecated. Please always use string[] if you pass an array
+     * as first argument.
      *
      * @param string|string[]|\Cake\Database\Type[]|null $type If string name of type to map, if array list of arrays to be mapped
      * @param string|\Cake\Database\Type|null $className The classname or object instance of it to register.
--- a/src/Database/Type.php
+++ b/src/Database/Type.php
@@ -107,6 +107,7 @@ class Type implements TypeInterface
         }
         if (is_string(static::$_types[$name])) {
             static::$_types[$name] = new static::$_types[$name]($name);
+
             return static::$_builtTypes[$name] = static::$_types[$name];
         }
 
--- a/src/Database/Type.php
+++ b/src/Database/Type.php
@@ -106,7 +106,8 @@ class Type implements TypeInterface
             throw new InvalidArgumentException(sprintf('Unknown type "%s"', $name));
         }
         if (is_string(static::$_types[$name])) {
-            return static::$_builtTypes[$name] = new static::$_types[$name]($name);
+            static::$_types[$name] = new static::$_types[$name]($name);
+            return static::$_builtTypes[$name] = static::$_types[$name];
         }
 
         return static::$_builtTypes[$name] = static::$_types[$name];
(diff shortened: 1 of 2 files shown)


tests/TestCase/Database/TypeTest.php
```diff
@@ -251,4 +254,33 @@ public function testSet()
...
+    {
+        $type = new Type('foo');
+        Type::clear();
+        Type::map('bool', BoolType::class);
+        Type::map('int', IntegerType::class);
+        $uuidType = new UuidType('uuid');
+        Type::map('uuid', $uuidType);
+        Type::build('bool');
+
+        $result = $type->__debugInfo();
```
markstory: Why does each type instance need access to the type registry in its debug output?

dereuromark: Currently it mainly also contained

	[protected] _basicTypes => [
		'string' => [
			'callback' => [
				(int) 0 => '\Cake\Database\Type',
				(int) 1 => 'strval'
			]
		],
		'text' => [
			'callback' => [
				(int) 0 => '\Cake\Database\Type',
				(int) 1 => 'strval'
			]
		],
		'boolean' => [
			'callback' => [
				(int) 0 => '\Cake\Database\Type',
				(int) 1 => 'boolval'
			],
			'pdo' => (int) 5
		]
	]

So this cuts it down a bit.
But if you think that should also be omitted, we can further remove.

dereuromark:
Are those failing tests related?

src/Database/Type.php
```diff
@@ -106,21 +106,23 @@ public static function build($name)
             throw new InvalidArgumentException(sprintf('Unknown type "%s"', $name));
         }
         if (is_string(static::$_types[$name])) {
-            return static::$_builtTypes[$name] = new static::$_types[$name]($name);
+            static::$_types[$name] = new static::$_types[$name]($name);
```
saeideng: $_types[$name] just have  string values 
do you want to save object in here ?

dereuromark: It always expects that it can contain objects. See map() with array as first arg.

saeideng: I see
Not expect
Arrays of string `string[]` not arrays of objects

dereuromark: We can also deprecate the \Cake\Database\Type[] part of map() as array setter and force the _types to always be of type string[] in future versions.
I am fine either way. But as of currently, the above tweak seems to re-use an object if possible instead of re-new it. More like nano-opt probably.

saeideng: That use once per key
See line 102 an 114
I pretty sure that use just once and not use again
Except after clear()

saeideng: So type of string is true

dereuromark: True, we could sure deprecate the object assignment for map($arrayOfTypes) then.

tests/TestCase/Database/TypeTest.php
markstory: I don't think the instances need to expose that static information. We'll be splitting up the static and instance sides of Types in 4.0

src/Database/Type.php
saeideng: 1) I test your branch on local 
```php
vendor/bin/phpunit tests/TestCase/Database/Type
PHPUnit 5.7.21 by Sebastian Bergmann and contributors.

...............................................................  63 / 125 ( 50%)
.......................................................S......  125 / 125 (100%)

Time: 204 ms, Memory: 11.25MB

OK, but incomplete, skipped, or risky tests!
Tests: 125, Assertions: 279, Skipped: 1.
```
2) I not find any usage of object for type() in tests

saeideng: 1) I test your branch on local
`vendor/bin/phpunit tests/TestCase/Database`
all test :+1: 
2) I not find any usage of object for type() in tests

dereuromark: I didnt say there are tests for it, thats also why I added them.
I said that the code itself already expects them to be possibly of type object. So in light of that this only makes it official and obvious what the code always has already been checking for.

saeideng:
I found this 
https://codecov.io/gh/cakephp/cakephp/src/master/src/Database/Type.php#L152...167

src/Database/Type.php
saeideng: @dereuromark  
:+1:  you are right
But I believe is irrational to accept object on `$_types`, because we have `$_builtTypes` for that

dereuromark: That also concerned me, yeah. We should really deprecate that part and remove it in 4.0.

tests/TestCase/Database/TypeTest.php
```diff
@@ -159,6 +162,23 @@ public function testReMapAndBuild()
...
+    public function testMapAndBuildWithObjects()
+    {
+        $map = Type::map();
+        Type::clear();
+
+        $uuidType = new UuidType('uuid');
+        Type::map('uuid', $uuidType);
+
+        $this->assertSame($uuidType, Type::build('uuid'));
+        Type::map($map);
```
saeideng: unnecessary  (setup() and tearDown() ... )
but I have another concern

dereuromark: Thats the only way to reset..

saeideng: my concern is about `$_builtTypes`
you want to reset `$_types` by  `map($map)`  but `$_builtTypes` remains 
and in next usage that use old `$_builtTypes` :)

dereuromark: Are you sure?
The code reads

    static::$_types = [];
    static::$_builtTypes = [];

for clear()

dereuromark: You are right, I now removed the caching and deprecated the object usage here.
Should be good now.

saeideng: see map and build 
```php
    public static function map($type = null, $className = null)
    {
        if ($type === null) {
            return static::$_types;
        }
        if (is_array($type)) {
            static::$_types = $type;

            return null;
        }

/////////
    public static function build($name)
    {
        if (isset(static::$_builtTypes[$name])) {
            return static::$_builtTypes[$name];
        }
```
if you need to reset `$_builtTypes`
try 
```php
$map = Type::map();
Type::clear();
Type::map($map);
```
in the setup and teardown()

PR #11496, opened 2017-12-02
Add BehaviorRegistry::className().

ADmad:
This helps in behavior class resolution since core behaviors are
under ORM/Behavior while in app they are under Model/Behavior.

### DIFF
--- a/src/ORM/BehaviorRegistry.php
+++ b/src/ORM/BehaviorRegistry.php
@@ -85,7 +85,7 @@ class BehaviorRegistry extends ObjectRegistry implements EventDispatcherInterfac
      * Resolve a behavior classname.
      *
      * @param string $class Partial classname to resolve.
-     * @return string|false Either the correct classname or false.
+     * @return string|null Either the correct classname or null.
      * @since 3.5.7
      */
     public static function className($class)
@@ -95,7 +95,7 @@ class BehaviorRegistry extends ObjectRegistry implements EventDispatcherInterfac
             $result = App::className($class, 'ORM/Behavior', 'Behavior');
         }
 
-        return $result;
+        return $result ?: null;
     }
 
     /**
@@ -108,7 +108,7 @@ class BehaviorRegistry extends ObjectRegistry implements EventDispatcherInterfac
      */
     protected function _resolveClassName($class)
     {
-        return static::className($class);
+        return static::className($class) ?: false;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


dereuromark:
Can we please have 4.0 use also Model/Behavior for consistency? :)
That would solve the issue in a better way.

For 3.x :+1: looks good.

src/ORM/BehaviorRegistry.php
```diff
@@ -84,12 +84,11 @@ public function setTable(Table $table)
...
      * Resolve a behavior classname.
      *
-     * Part of the template method for Cake\Core\ObjectRegistry::load()
-     *
      * @param string $class Partial classname to resolve.
      * @return string|false Either the correct classname or false.
+     * @since 3.5.7
      */
-    protected function _resolveClassName($class)
+    public static function className($class)
```
dereuromark: Can we have this method return string|null instead?
The former method sure must stay BC, but this new one could get rid of the false code smell IMO.

src/ORM/BehaviorRegistry.php
```diff
@@ -99,6 +98,19 @@ protected function _resolveClassName($class)
...
 
+    /**
+     * Resolve a behavior classname.
+     *
+     * Part of the template method for Cake\Core\ObjectRegistry::load()
+     *
+     * @param string $class Partial classname to resolve.
+     * @return string|false Either the correct classname or false.
+     */
+    protected function _resolveClassName($class)
```
saeideng: how we can find this when porting to 4.x for removing

ADmad: @saeideng This won't be removed in 4.0 since it's an inherited abstract method `ObjectRegistry::_resolveClassName()`, implemented by all child classes.

src/ORM/BehaviorRegistry.php
ADmad: @dereuromark For now maintaining consistency with `App::className()` is more important so that this method can be easily swapped where required. In future both can be changed to return null.

dereuromark: Well the basic usage will still be the same: if () and empty() will be the same, but future strictness in nullable regards is easier if the method would already have a proper API.
But no big deal right now either way.

markstory:
> Can we please have 4.0 use also Model/Behavior for consistency? :)

How would that work with cakephp/orm being a subsplit? That is the primary reason we have behaviors in orm in the framework. We had decided that adding an ORM package to applications didn't really make much sense as applications don't really contain ORMs.

src/ORM/BehaviorRegistry.php
markstory: This new method could use the string|null pattern. As people switch over they would be able to update their checks as well.

ADmad:
I have updated the method to return `null` on failure.

dereuromark (approved):

PR #11497, opened 2017-12-02
Simplify travis matrix

dereuromark:
At this point it should suffice to only run phpstan once, in php7.1


markstory:
Looks like phpcsis failing because of mcrypt function calls.

PR #11480, opened 2017-11-30
Refactoring tests

carusogabriel:
I've refactored several tests with [`assertFileNotExists`](https://phpunit.de/manual/current/pt_br/appendixes.assertions.html#appendixes.assertions.assertFileExists), [`assertFalse`](https://phpunit.de/manual/current/pt_br/appendixes.assertions.html#appendixes.assertions.assertFalse), [`assertCount`](https://phpunit.de/manual/current/pt_br/appendixes.assertions.html#appendixes.assertions.assertCount), [`assertEmpty`](https://phpunit.de/manual/current/pt_br/appendixes.assertions.html#appendixes.assertions.assertEmpty)   and [`assertNull`](https://phpunit.de/manual/current/pt_br/appendixes.assertions.html#appendixes.assertions.assertNull) methods.

PR #11498, opened 2017-12-03
Refactoring tests

carusogabriel:
Continuing to refactory test, like #11480, I've used:
- `assertNotInternalType` instead of `is_*` methods;
- `assertNull` instead of strict comparisons with `null` keyword;
- `assertTrue` instead of strict comparisons with `true` keyword;
- `assertFalse` and `assertNotFalse` instead of strict comparisons with `false` keyword;
- `assertGreaterThanOrEqual` for matematical comparisions;
- `assertStringEqualsFile` when comparing a `string` with a `file`;
- `assertSame` and `assertNotSame` instead of strict comparisons;
- `assertContains` and `assertNotContains` instead of `in_array` method.


markstory:
Thank you!

PR #11503, opened 2017-12-04
Fix Request::referer(true) returning scheme-relative URLs

chinpei215:


tests/TestCase/Http/ServerRequestTest.php
```diff
@@ -731,6 +734,10 @@ public function testReferer()
         $result = $request->referer(true);
         $this->assertSame('/some/path', $result);
 
+        $request->env('HTTP_REFERER', Configure::read('App.fullBaseUrl') . '///cakephp.org/');
```
dereuromark: Whats the value here?
I would expect a test like `//my-domain/path` as test to test the above code change.

chinpei215: Sorry if I am wrong, but the problem is at line 578 in ServerRequest.php:
```php
$ref = substr($ref, strlen($base));
```
This removes `App.fullBaseUrl` (e.g. `http://localhost`) and `ServerRequest::$webroot` (e.g. `/`) from the `HTTP_REFERER`. So I tested `http://localhost` + `/` + `//cakephp.org/` here.

Issue #11501, opened 2017-12-04
Add $_validatorClass to Cake\Form\Form


andruhapx:
This is a (multiple allowed):

* [x] bug
* [x] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.5.5

Is it possible to add new field "protected $_validatorClass = '\Cake\Validation\Validator';" to Cake\Form\Form?
Because now Form does not use ValidatorAwareTrait, and validator class is hardcoded:
$validator = $this->_buildValidator(new Validator());

dereuromark:
You can make a PR here to request your enhancement. Sounds like a good idea.

ADmad:
Closing as related PR #11505 is open.

PR #11505, opened 2017-12-04
Form validator

ADmad:
Implements https://github.com/cakephp/cakephp/issues/11501

### DIFF
--- a/src/Form/Form.php
+++ b/src/Form/Form.php
@@ -14,7 +14,14 @@
  */
 namespace Cake\Form;
 
+use Cake\Event\Event;
+use Cake\Event\EventDispatcherInterface;
+use Cake\Event\EventDispatcherTrait;
+use Cake\Event\EventListenerInterface;
+use Cake\Event\EventManager;
 use Cake\Validation\Validator;
+use Cake\Validation\ValidatorAwareInterface;
+use Cake\Validation\ValidatorAwareTrait;
 
 /**
  * Form abstraction used to create forms not tied to ORM backed models,
@@ -33,8 +40,25 @@ use Cake\Validation\Validator;
  *
  * Forms are conventionally placed in the `App\Form` namespace.
  */
-class Form
+class Form implements EventListenerInterface, EventDispatcherInterface, ValidatorAwareInterface
 {
+
+    use EventDispatcherTrait;
+    use ValidatorAwareTrait;
+
+    /**
+     * The alias this object is assigned to validators as.
+     *
+     * @var string
+     */
+    const VALIDATOR_PROVIDER_NAME = 'form';
+
+    /**
+     * The name of the event dispatched when a validator has been built.
+     *
+     * @var string
+     */
+    const BUILD_VALIDATOR_EVENT = 'Form.buildValidator';
 
     /**
      * The schema used by this form.
@@ -58,11 +82,29 @@ class Form
     protected $_validator;
 
     /**
-     * Validator class.
-     *
-     * @var string
+     * Constructor
      */
-    protected $_validatorClass = 'Cake\Validation\Validator';
+    public function __construct()
+    {
+        $this->_eventManager = new EventManager();
+        $this->_eventManager->on($this);
+    }
+
+    /**
+     * Get the Form callbacks this form is interested in.
+     *
+     * The conventional method map is:
+     *
+     * - Form.buildValidator => buildValidator
+     *
+     * @return array
+     */
+    public function implementedEvents()
+    {
+        return [
+            'Form.buildValidator' => 'buildValidator',
+        ];
+    }
 
     /**
      * Get/Set the schema for this form.
@@ -129,37 +171,6 @@ class Form
     }
 
     /**
-     * Get the validator for this form.
-     *
-     * This method will call `_buildValidator()` when the validator
-     * is first built. This hook method lets you configure the
-     * validator or load a pre-defined one.
-     *
-     * @return \Cake\Validation\Validator the validator instance.
-     */
-    public function getValidator()
-    {
-        if (empty($this->_validator)) {
-            $this->_validator = $this->_buildValidator(new $this->_validatorClass);
-        }
-
-        return $this->_validator;
-    }
-
-    /**
-     * Set a custom validator instance.
-     *
-     * @param \Cake\Validation\Validator $validator Validator object to be set.
-     * @return $this
-     */
-    public function setValidator(Validator $validator)
-    {
-        $this->_validator = $validator;
-
-        return $this;
-    }
-
-    /**
      * A hook method intended to be implemented by subclasses.
      *
      * You can use this method to define the validator using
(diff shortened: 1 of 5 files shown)


src/Form/Form.php
```diff
@@ -57,6 +57,13 @@ class Form
      */
     protected $_validator;
 
+    /**
+     * Validator class.
+     *
+     * @var string
+     */
+    protected $_validatorClass = '\Cake\Validation\Validator';
```
dereuromark: The \ prefix should be removed

dereuromark: Or can we even use ::class syntax maybe?

src/Form/Form.php
```diff
@@ -139,7 +182,7 @@ protected function _buildValidator(Validator $validator)
      */
     public function validate(array $data)
     {
-        $validator = $this->validator();
+        $validator = $this->getValidator();
```
dereuromark: Are we allowed to use the non-deprecated new one inside 3.x?
I mean if someone extended validator(), would that be an issue because it is not called anymore?

ADmad: There won't be any issue.

dereuromark: What do you mean?
I just remember that we had some issues with inheritance here, because people implemented the old method, and the core itself now only calls new method, thus the extended method and its functionality is not used/called anymore.
That sounds like a BC break.

ADmad: I am pretty sure we don't provide any guarantees about overridden methods in minor releases. For e.g. when `Form::input()` was deprecated in favor of `Form::control()`, bootstrap-ui plugin's `FormHelper` which had overridden `Form::input()` had to be updated.

ADmad: We only guarantee backwards compatibility with public API, not the internals.

dereuromark: What is the definition of public api?
Because `public function validator(Validator ...)` sounds rather public, and might very well be called from the outside as well.

tests/TestCase/Form/FormTest.php
```diff
@@ -45,18 +45,35 @@ public function testSchema()
      * Test validator()
      *
      * @return void
+     * @deprecated
```
saeideng: `* @group deprecated` ?

src/Form/Form.php
```diff
@@ -116,6 +128,37 @@ public function validator(Validator $validator = null)
...
+    /**
+     * Get the validator for this form.
+     *
+     * This method will call `_buildValidator()` when the validator
+     * is first built. This hook method lets you configure the
+     * validator or load a pre-defined one.
+     *
+     * @return \Cake\Validation\Validator the validator instance.
+     */
+    public function getValidator()
```
markstory: Should Form use the `ValidatorAwareTrait`? That would let it implement the same interface as the ORM and open up the option for multiple validators if people need them.

ADmad: @markstory I thought about that but `Form::validator()` and `ValidatorAwareTrait::validator()` don't have the same arguments, so I don't think PHP will be happy if I make `Form` use `ValidatorAwareTrait`.

src/Form/Form.php
ADmad: `ValidatorAwareTrait::validator()` also uses it's new `getValidator()` and `setValidator()` internally. I am sure there are other such examples where the deprecated methods use the newly added methods internally.

dereuromark: I see. So we have a more global BC problem here, strictly speaking. But maybe we can ignore it.

src/Form/Form.php
ADmad: But perhaps at least I can make the new get/set methods have the same params as the trait so that in 4.x we can get rid of the method and have `Form` use `ValidatorAwareTrait`.

markstory: You could alias the trait method to another name and make it protected?

tests/TestCase/View/Form/FormContextTest.php
```diff
@@ -138,18 +138,20 @@ public function testValDefault()
...
-        $this->assertTrue($context->isRequired('email'));
-        $this->assertFalse($context->isRequired('body'));
-        $this->assertFalse($context->isRequired('Prefix.body'));
+            $context = new FormContext($this->request, [
+                'entity' => $form
+            ]);
+            $this->assertTrue($context->isRequired('name'));
+            $this->assertTrue($context->isRequired('email'));
+            $this->assertFalse($context->isRequired('body'));
+            $this->assertFalse($context->isRequired('Prefix.body'));
```
markstory: This test should be updated to use the non-deprecated features.

tests/TestCase/View/Form/FormContextTest.php
```diff
@@ -206,46 +208,48 @@ public function testAttributes()
...
+                ->add('email', 'format', ['rule' => 'email'])
+                ->add('name', 'length', ['rule' => ['minLength', 10]])
+                ->addNested('pass', $nestedValidator);
+            $form->validate([
+                'email' => 'derp',
+                'name' => 'derp',
+                'pass' => [
+                    'password' => 'short',
+                    'confirm' => 'long enough',
+                ],
```
markstory: This test should be updated to use the non-deprecated features. I'm planning to delete all the tests using `deprecated()` when we move to 4.x development.

src/Form/Form.php
```diff
@@ -33,9 +40,26 @@
  *
  * Forms are conventionally placed in the `App\Form` namespace.
  */
-class Form
+class Form implements EventListenerInterface, EventDispatcherInterface, ValidatorAwareInterface
```
saeideng: using `Event` needs to update https://github.com/cakephp/cakephp/blob/3.next/src/Form/composer.json

src/Form/Form.php
```diff
@@ -57,6 +81,31 @@ class Form
      */
     protected $_validator;
 
+    /**
+     * Constructor
+     */
+    public function __construct()
+    {
+        $this->_eventManager = new EventManager();
```
robertpustulka: Can this be injectable?

src/Form/Form.php
```diff
@@ -125,12 +181,25 @@ public function validator(Validator $validator = null)
...
 
+    /**
+     * Callback method for Form.buildValidator event.
+     *
+     * @param \Cake\Event\Event $event The Form.buildValidator event instance.
+     * @param \Cake\Validation\Validator $validator The validator to customize.
+     * @param string $name Validator name
+     * @return void
+     */
+    public function buildValidator(Event $event, Validator $validator, $name)
```
robertpustulka: I like this change :+1:

src/Form/Form.php
ADmad: @saeideng Good catch.

src/Form/Form.php
ADmad: @robertpustulka Done

src/Form/composer.json
```diff
@@ -22,7 +22,7 @@
     },
     "require": {
         "php": ">=5.6.0",
-        "cakephp/event": "^3.0.0",
+        "cakephp/event": "^3.5.0",
         "cakephp/validation": "^3.0.0"
```
robertpustulka: This needs to be updated as well.
https://github.com/cakephp/cakephp/blob/master/src/Validation/ValidatorAwareInterface.php#L12

Issue #10730, opened 2017-06-07
__d() doesn't take the correct msgstr depending on the position of the msgstr inside the .po file


burzum:
Copy pasting the internal ticket from my workmate @antoniosergio-moreira here. @netstyler

---

### Original Problem
In the website's - www.german-architects.com -  main navigation and in the Profiles navigation, the translation for the word "Projects" is only displaying "Projekt" and not "Projekte" (plural)

![bildschirmfoto-2017-06-06-um-15 22 17](https://cloud.githubusercontent.com/assets/13075805/26831430/8cba086a-4acc-11e7-817b-cb191fc174ea.jpg)

### What we did at first
We checked the .po file for the word "Projects" and did not find anything. That was issue n.1. Issue n.2 is a bit trickier. 

in `src/View/Helper/ProfileHelper.php`, line 75

```
[
	'name' => __d('wa', 'Projects'),
	'link' => ['controller' => 'Profiles', 'action' => 'view', $profile->slug, 'projects'],
	'editLink' => ['controller' => 'Profiles', 'action' => 'edit', 'projects'],
	'counterField' => 'project_count',
],
```

We added this word - "Projects" -  as a msgstr to the `src/Locale/de/wa.po`file. 
At first, we added it in lines 297-298. Note that in lines 306-309 there is the singular/plural translation for Project/Projects. Having done this and cleared the cache, we note that the translations are not working in the website. 

**Addind the msgstr in lines 297-298**
![screen shot 2017-06-07 at 10 16 00](https://user-images.githubusercontent.com/8680933/26868754-5ab27502-4b6a-11e7-9a4f-259531c85b78.png)
**In the website, translations are not accurate**
![screen shot 2017-06-07 at 10 17 38](https://user-images.githubusercontent.com/8680933/26868849-9c3dc5e4-4b6a-11e7-9ab4-656f3400c4e7.png)

### What we did afterwards
Afterwards, we moved the msgstrs to the bottom of the `src/Locale/de/wa.po` file (lines 1934-1935). After clearing cache once again, we note that the translations in the frontend are now correct. 
**Addind the msgstr in lines 1935-1936**
![screen shot 2017-06-07 at 10 29 59](https://user-images.githubusercontent.com/8680933/26869329-5679baac-4b6c-11e7-80f2-6f36e1733193.png)
**In the website, translations are accurate**
[...]

PR #11508, opened 2017-12-05
Fixing plural/singular issues in the po parser

burzum:
This will fix the case described here: https://github.com/cakephp/cakephp/issues/10730#issuecomment-349286149

### DIFF
--- a/src/I18n/Translator.php
+++ b/src/I18n/Translator.php
@@ -22,6 +22,8 @@ use Aura\Intl\Translator as BaseTranslator;
 class Translator extends BaseTranslator
 {
 
+    const PLURAL_PREFIX = 'p:';
+
     /**
      * Translates the message formatting any placeholders
      *
@@ -33,7 +35,18 @@ class Translator extends BaseTranslator
      */
     public function translate($key, array $tokensValues = [])
     {
-        $message = $this->getMessage($key);
+        if (isset($tokensValues['_count'])) {
+            $message = $this->getMessage(static::PLURAL_PREFIX . $key);
+            if (!$message) {
+                $message = $this->getMessage($key);
+            }
+        } else {
+            $message = $this->getMessage($key);
+            if (!$message) {
+                $message = $this->getMessage(static::PLURAL_PREFIX . $key);
+            }
+        }
+
         if (!$message) {
             // Fallback to the message key
             $message = $key;
--- a/tests/TestCase/I18n/Parser/PoFileParserTest.php
+++ b/tests/TestCase/I18n/Parser/PoFileParserTest.php
@@ -77,8 +77,8 @@ class PoFileParserTest extends TestCase
             'p:%d = 0 or > 1' => [
                 '_context' => [
                     'Another Context' => [
-                        (int)0 => '%d = 1 (translated)',
-                        (int)1 => '%d = 0 or > 1 (translated)'
+                        0 => '%d = 1 (translated)',
+                        1 => '%d = 0 or > 1 (translated)'
                     ]
                 ]
             ],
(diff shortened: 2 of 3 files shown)


tests/TestCase/I18n/Parser/PoFileParserTest.php
```diff
@@ -65,39 +65,39 @@ public function testParse()
         $expected = [
             'Plural Rule 1' => [
                 '_context' => [
-                    '' => 'Plural Rule 1 (translated)',
-                ],
```
dereuromark: You shouldnt revert the trailing commas IMO, just too much diff changes then that are none.

tests/TestCase/I18n/Parser/PoFileParserTest.php
```diff
@@ -65,39 +65,39 @@ public function testParse()
...
                     'Another Context' => '%d = 1 (translated)'
                 ]
             ],
-            '%d = 0 or > 1' => [
+            'p:%d = 0 or > 1' => [
                 '_context' => [
                     'Another Context' => [
-                        0 => '%d = 1 (translated)',
-                        1 => '%d = 0 or > 1 (translated)'
+                        (int)0 => '%d = 1 (translated)',
```
dereuromark: I dont think you need the (int) cast here on the keys.

src/I18n/Translator.php
```diff
@@ -22,6 +22,8 @@
 class Translator extends BaseTranslator
 {
 
+    const PLURAL_PREFIX = 'p:';
```
markstory: Could we have one copy of the constant?

src/I18n/Translator.php
```diff
@@ -33,7 +35,18 @@ class Translator extends BaseTranslator
      */
     public function translate($key, array $tokensValues = [])
     {
-        $message = $this->getMessage($key);
+        if (isset($tokensValues['_count'])) {
+            $message = $this->getMessage(static::PLURAL_PREFIX . $key);
+            if (!$message) {
+                $message = $this->getMessage($key);
+            }
```
markstory: We'll need this for backwards compatibilty and userland defined message loaders/parsers.

tests/TestCase/I18n/Parser/PoFileParserTest.php
burzum: Seriously? All the other lines haven't had one I think and AFAIK best practice by CS is to not have it there?

src/I18n/Translator.php
burzum: What exactly do you mean? Do I need to change something else here?

tests/TestCase/I18n/Parser/PoFileParserTest.php
dereuromark: No best practice (what the sniffer should be enforcing) is to always have the multiline trailing commas, to keep diffs low exactly for the reasons shown in this PR: Too much unrelated noise around the few lines of actual change.

burzum: They aren't used everywhere either here! https://github.com/cakephp/cakephp/blob/master/tests/TestCase/I18n/Parser/PoFileParserTest.php#L59-L102 So either remove them all or add them all!? However both ways will cause "noise".

dereuromark: I guess in that case you can ignore my comment.

burzum: We should agree on something and putt it into the coding standard? What do you think?

dereuromark: Yeah, but can then be unrelated to this PR.

src/I18n/Translator.php
markstory: No. I was just commenting on it for future us.

PR #11512, opened 2017-12-07
3.next - Add deprecation warnings for public request properties

markstory:
Having public properties on the request is not something we can have in a PSR7 style request object. I've added deprecation warnings to these properties by using magic methods. This may result in small issues, but we don't have many other tools to deprecate public properties.

Unfortunately it is another very large pull request :cry:. This should be the last one though.

Refs #11385

### DIFF
--- a/src/Routing/Router.php
+++ b/src/Routing/Router.php
@@ -620,8 +620,8 @@ class Router
         // In 4.x this should be replaced with state injected via setRequestContext
         $request = static::getRequest(true);
         if ($request) {
-            $params = $request->params;
-            $here = $request->here;
+            $params = $request->getAttribute('params');
+            $here = $request->getRequestTarget();
             $base = $request->getAttribute('base');
         } else {
             $base = Configure::read('App.base');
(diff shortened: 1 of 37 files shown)


src/Routing/Router.php
```diff
@@ -902,17 +902,15 @@ public static function parseNamedParams(ServerRequest $request, array $options =
             '2.x backwards compatible named parameter support will be removed in 4.0'
         );
         $options += ['separator' => ':'];
-        if (empty($request->params['pass'])) {
-            $request->params['named'] = [];
-
-            return $request;
+        if (!$request->getParam('pass')) {
+            return $request->withParam('named', []);
```
saeideng:
```‍‍‍php
$request->params['named'] = [];
return $request;
```
is not equal to 
```php
return $request->withParam('named', []);
```
i mean 
```php
return $request=$request->withParam('named', []);
```

src/Routing/Router.php
```diff
@@ -933,9 +931,7 @@ public static function parseNamedParams(ServerRequest $request, array $options =
             }
             $named = array_merge_recursive($named, [$key => $value]);
         }
-        $request->params['named'] = $named;
-
-        return $request;
+        return $request->withParam('named', $named);
```
saeideng: and here

src/Routing/Router.php
markstory: How are they not equivalent? In both cases the mutated request is returned.

markstory: The difference is the request is not mutated in place. I've updated the tests to use the return value and will update the migration guide as well.

Issue #10877, opened 2017-07-07
3.4: Timezone wrong for Widgets


dereuromark:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.4

### What you did
Normal recommended bootstrap of
```php
// Internal UTC
date_default_timezone_set('UTC');

// Localized DE
ini_set('intl.default_locale', 'de_DE');

Type::build('time')->useLocaleParser();
Type::build('date')->useLocaleParser();
Type::build('datetime')->useLocaleParser();
```

And a CRUD for events - and use default dropdown selects for date time via DateTimeWidget.

### What happened
The added event shows up in the DB as inputted (UTC), but should in fact be stored in UTC converted from local input. It is now 2 hours off.
The output via TimeHelper and `outputTimezone` option is properly converted from the UTC to GMT+2 value in view/index (even though now two hours off).
But this part of timezone adjustment for output is missing for add/edit and thus shows up wrong here.

### What you expected to happen
The FormHelper and the Widgets must to the same transformation of +2 and -2 (for berlin anyway) just as the TimeHelper is doing. Otherwise localized times are stored and read wrong for Forms.

This fixes the issue for display of forms:
```php
protected function _deconstructDate($value, $options)
    {
        ...

        $value = $value->timezone('Europe/Berlin');
```
But one would also need to fix the input convertion, otherwise it will always add +2 more...

markstory:
Are you storing your datetime into a datetime column or a timestamp column? And is your database connection setup to use UTC as well?

dereuromark:
It is a datetime one.
I can see it is set to SYSTEM by default.

dereuromark:
Can we set a connection based timezone from Cake here for the (mysql) connection?

markstory:
datetime columns in mysql do not store timezone information. I asked because timestamp columns do timezone conversions when reading data. 

If you are using the selectbox widget for selecting your datetime values, you don't need to enable `useLocaleParser()` as that mode is only used when marshaling string values. This is also why your dates are off by 2 hours. Additionally, locale parsing does not set/change the timezone to the locale timezone as that is not possible for most locales.

It sounds like you are missing a timezone conversion as CakePHP can't do it automatically as it doesn't know the user's timezone.

dereuromark:
Well, even if I deactivate the locale parsing, the forms still display the utc time directly displayed.
[...]

PR #11517, opened 2017-12-07
Add basic timezone support for form datetime input.

dereuromark:
Basic stab at https://github.com/cakephp/cakephp/issues/10877

Currently all forms are broken if you follow recommended UTC for DB and localized forms (e.g. UTC+1 Berlin time).
The time in forms is always behind upon transformation to output, and when you save it, it becomes even more behind as it converts it wrong a 2nd time.
Using a defaultOutputTimezone config (and optionally even a timezone dropdown maybe) you can have localized forms and keep the expected timezoned datetime.

If there is only one timezone (output), one can just use config here:
```php
	'App' => [
		'defaultLocale' => env('APP_DEFAULT_LOCALE', 'de_DE'),
		'defaultOutputTimezone' => env('APP_DEFAULT_OUTPUT_TIMEZONE', 'Europe/Berlin'),
```

I kindly ask for help to finalize this for 3.6.

### DIFF
--- a/src/Database/Type/DateTimeType.php
+++ b/src/Database/Type/DateTimeType.php
@@ -243,7 +243,8 @@ class DateTimeType extends Type implements TypeInterface
      *
      * @return bool
      */
-    protected function _isEmpty($value) {
+    protected function _isEmpty($value)
+    {
         unset($value['timezone']);
 
         return implode('', $value) === '';
--- a/src/Database/Type/DateTimeType.php
+++ b/src/Database/Type/DateTimeType.php
@@ -14,6 +14,7 @@
  */
 namespace Cake\Database\Type;
 
+use Cake\Core\Configure;
 use Cake\Database\Driver;
 use Cake\Database\Type;
 use Cake\Database\TypeInterface;
@@ -181,7 +182,7 @@ class DateTimeType extends Type implements TypeInterface
             return $value;
         }
 
-        if (is_array($value) && implode('', $value) === '') {
+        if (is_array($value) && $this->_isEmpty($value)) {
             return null;
         }
         $value += ['hour' => 0, 'minute' => 0, 'second' => 0];
@@ -206,9 +207,17 @@ class DateTimeType extends Type implements TypeInterface
             $value['minute'],
             $value['second']
         );
-        $tz = isset($value['timezone']) ? $value['timezone'] : null;
+        $tz = isset($value['timezone']) ? $value['timezone'] : Configure::read('App.defaultOutputTimezone');
 
-        return new $class($format, $tz);
+        /* @var \Cake\I18n\Time $time */
+        $time = new $class($format, $tz);
+
+        $internalTimezone = date_default_timezone_get();
+        if ($tz && $tz !== $internalTimezone) {
+            $time = $time->timezone($internalTimezone);
+        }
+
+        return $time;
     }
 
     /**
@@ -225,6 +234,19 @@ class DateTimeType extends Type implements TypeInterface
         }
 
         return false;
+    }
+
+    /**
+     * Returns true if the array can be considered an empty date time value.
+     *
+     * @param array $value Date time array.
+     *
+     * @return bool
+     */
+    protected function _isEmpty($value) {
+        unset($value['timezone']);
+
+        return implode('', $value) === '';
     }
 
     /**
--- a/src/View/Widget/DateTimeWidget.php
+++ b/src/View/Widget/DateTimeWidget.php
@@ -524,7 +524,7 @@ class DateTimeWidget implements WidgetInterface
     /**
      * Generates a meridian select
      *
-     * @param array $options The options to generate a meridian select with.
+     * @param array $options The options to generate a meridian select.
      * @param \Cake\View\Form\ContextInterface $context The current form context.
      * @return string
      */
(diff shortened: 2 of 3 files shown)


saeideng:
looks this is only for showing , not converting request data from user

dereuromark:
Both, it converts correctly for display in the input form (dropdowns by default), and it converts back before validating/saving.
Not yet sure about any edge cases yet though.
But seems BC so far, if you do not use any timezoning here.

saeideng:
do you think we can add calender in future ?(non-Gregorian)

saeideng:
how happen if someone remove hidden timezone field ?

src/View/Widget/DateTimeWidget.php
```diff
@@ -93,6 +94,7 @@ public function __construct(StringTemplate $templates, SelectBoxWidget $selectBo
      * - `second` - Set to true to enable the seconds input. Defaults to false.
      * - `meridian` - Set to true to enable the meridian input. Defaults to false.
      *   The meridian will be enabled automatically if you choose a 12 hour format.
+     * - `timezone` - A hidden field for timezone handling when working with UTC internally.
```
markstory: Not just UTC really. Any timezone could be stored in the hidden input.

src/View/Widget/DateTimeWidget.php
```diff
@@ -528,6 +540,26 @@ protected function _meridianSelect($options, $context)
...
+    {
+        $options += [
+            'name' => '',
+            'val' => null,
+            'style' => ['display: none'],
+            'options' => [$options['val'] => $options['val']],
+            'templateVars' => [],
+        ];
+
+        return $this->_select->render($options, $context);
```
markstory: Why are we making a select and not a hidden input? Seems like a hidden input is what we want here.

src/Database/Type/DateTimeType.php
```diff
@@ -206,9 +207,17 @@ public function marshal($value)
...
+        $tz = isset($value['timezone']) ? $value['timezone'] : Configure::read('App.defaultOutputTimezone');
 
-        return new $class($format, $tz);
+        /* @var \Cake\I18n\Time $time */
+        $time = new $class($format, $tz);
+
+        $internalTimezone = date_default_timezone_get();
+        if ($tz && $tz !== $internalTimezone) {
+            $time = $time->timezone($internalTimezone);
+        }
```
markstory: Does the time type need this kind of change as well?

src/View/Widget/DateTimeWidget.php
```diff
@@ -211,25 +214,30 @@ protected function _normalizeData($data)
...
         try {
             if (is_string($value) && !is_numeric($value)) {
-                $date = new DateTime($value);
+                $date = new DateTime($value, $timezone);
             } elseif (is_bool($value)) {
-                $date = new DateTime();
+                $date = new DateTime('now', $timezone);
             } elseif (is_int($value) || is_numeric($value)) {
-                $date = new DateTime('@' . $value);
+                $date = new DateTime('@' . $value, $timezone);
```
ADmad: Passing `$timezone` here is redundant. From the PHP manual:
> The $timezone parameter and the current timezone are ignored when the $time parameter either is a UNIX timestamp (e.g. @946684800) or specifies a timezone (e.g. 2010-01-28T15:00:00+02:00).

src/Database/Type/DateTimeType.php
```diff
@@ -206,9 +207,17 @@ public function marshal($value)
...
-        $tz = isset($value['timezone']) ? $value['timezone'] : null;
+        $tz = isset($value['timezone']) ? $value['timezone'] : Configure::read('App.defaultOutputTimezone');
 
-        return new $class($format, $tz);
+        /* @var \Cake\I18n\Time $time */
+        $time = new $class($format, $tz);
+
+        $internalTimezone = date_default_timezone_get();
+        if ($tz && $tz !== $internalTimezone) {
+            $time = $time->timezone($internalTimezone);
```
ADmad: Doesn't this change mean the timezone should be again set to default timezone before converting the object to string in `toDatabase()` method?

dereuromark: My tests seemed to yield the right conversion, so maybe we do not? But I am happy to be proven wrong.

src/Database/Type/DateTimeType.php
dereuromark: I would think so, so far I only checked datetime. If anyone has the chance to help with other types, that would be more than welcome.

src/View/Widget/DateTimeWidget.php
dereuromark: I agree. How would that look like here? Using a widget in a widget?

src/View/Widget/DateTimeWidget.php
dereuromark: The point is that you store anything but UTC usually :) The UTC is the server timezone for DB storing it normalized, and the timezone is either passed in the array from user input (session data?) or coming from config by default (e.g. german based website using global gmt+1 output).
So the idea is UTC => GMT+1 for display => UTC for form handling.
Thus "working with UTC internally" - which seems to be the best practice when dealing with localized datetime across multiple timezones.

markstory: I agree UTC is what we should recommend, but people will do all sorts of interesting things.

src/View/Widget/DateTimeWidget.php
markstory: That's what `$this->_select` is as well. The WidgetRegistry stitches various widget dependencies together.

dereuromark:
Anyone able to help out, getting this ready for 3.7 maybe?
We could also look into https://github.com/JeanValJeann/cakephp-timezone and what they did.

Issue #11514, opened 2017-12-07
Argument 1 passed to Cake\Error\ExceptionRenderer::__construct() must be an instance of Exception


psantos10:
This is a (multiple allowed):

* [x] bug

* CakePHP Version: 3.5.7
* Platform and Target: PHP 7.1.11; MariaDB 10.2.11

### What you did

I write this code on the controller:

```
$categories = TableRegistry::get('Categories');
$category = $categories->find('all');
```

### What happened

Gives me this error:

> Argument 1 passed to Cake\Error\ExceptionRenderer::__construct() must be an instance of Exception, instance of Error given, called in /home/psantos/Workspace/my_app/vendor/cakephp/cakephp/src/Error/Middleware/ErrorHandlerMiddleware.php on line 167

### What you expected to happen

I was expecting to get a more descriptive error info. no matter what change I do, I stil getting the same error.

dereuromark:
We need to remove the typehint @ https://github.com/cakephp/cakephp/blob/master/src/Error/ExceptionRenderer.php#L88 for PHP7+.
Can you provide a patch PR maybe?

chinpei215:
I don't think we can remove typehints in a bugfix release. I think this issue is caused by a recent patch for the ErrorHandlerMiddleware. We could revert it or enhance it to use PHP7ErrorException.

dereuromark:
Typehints should be able to be removed if they fix these issues of PHP7 and broken code. But if you have an idea for enhancement, go for it.

chinpei215:
It would be better to revert it, as I said in #11462. I am not sure what bug is solved this PR.
However, if it fixed  some bugs, then we mat need to fix this new bug without breaking BC. 
Note that I am not saying the PR is not good. It is really good. but we can do it in 3.6. then  we could change the problematic typehints too.

dereuromark:
IMO the bigger issue is some php7 errors to not being logged, the typehint issue is minor in comparison.

chinpei215:
As I said in the PR, I think PHP7 errors are logged correctly by the legacy ErrorHandler class even in 3.5.6 or earlier. But if you could show me some edge case that those are not logged, then I could understand what was a problem.

dereuromark:
Custom error handler, see e.g. DatabaseLog plugin.

chinpei215:
Do you have some code snippet that causes the problem?

ADmad:
Perhaps this patch might fix the reported issue.

```php
[...]

PR #11518, opened 2017-12-07
Fix PHP7 error handler compatibility.

dereuromark:
Possible solution of https://github.com/cakephp/cakephp/issues/11514
We convert it back to the PHP5 exception where needed as the constructor for ExceptionRenderer exceptions Exception which is not compatible with PHP7.
Removing the typehint here can only be done in a non-patch version as the eco system could implement them still as such.

### DIFF
--- a/src/Error/Middleware/ErrorHandlerMiddleware.php
+++ b/src/Error/Middleware/ErrorHandlerMiddleware.php
@@ -191,10 +191,13 @@ class ErrorHandlerMiddleware
             return;
         }
 
-        $skipLog = $this->getConfig('skipLog');
-        if ($skipLog) {
-            foreach ((array)$skipLog as $class) {
-                if ($exception instanceof $class) {
+        $unwrapped = $exception instanceof PHP7ErrorException ?
+            $exception->getError() :
+            $exception;
+
+        if ($this->getConfig('skipLog')) {
+            foreach ((array)$this->getConfig('skipLog') as $class) {
+                if ($unwrapped instanceof $class) {
                     return;
                 }
             }
--- a/src/Error/Middleware/ErrorHandlerMiddleware.php
+++ b/src/Error/Middleware/ErrorHandlerMiddleware.php
@@ -19,7 +19,9 @@ use Cake\Core\Configure;
 use Cake\Core\Exception\Exception as CakeException;
 use Cake\Core\InstanceConfigTrait;
 use Cake\Error\ExceptionRenderer;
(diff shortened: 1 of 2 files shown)


markstory:
Any chance we could add a test for the scenario this is handling?

ADmad:
I added a test.

dereuromark:
Thank you for helping with the test case :)

src/Error/Middleware/ErrorHandlerMiddleware.php
```diff
@@ -184,10 +191,13 @@ protected function logException($request, $exception)
...
-        if ($skipLog) {
-            foreach ((array)$skipLog as $class) {
-                if ($exception instanceof $class) {
+        $unwrapped = $exception instanceof PHP7ErrorException ?
+            $exception->getError() :
+            $exception;
+
+        if ($this->getConfig('skipLog')) {
+            foreach ((array)$this->getConfig('skipLog') as $class) {
+                if ($unwrapped instanceof $class) {
```
markstory: Do tests need to be updated for logging fixes?

dereuromark: This is basically just a copy and paste of the old error handler. But  could sure be added.
Would probably be better to follow @saeideng s advice and re-use more common code instead of duplicating in the long run. But for now out of scope of the bugfix.

ADmad: Actually this change isn't necessary at all. `logException()` never get's `Error` instance wrapped inside `PHP7ErrorException`. The wrapping is only done inside `getRenderer()` when passing the exception instance to `ExceptionRender`'s constructor.

dereuromark: Are you 100% sure? I am also referring to any outside or custom handler extending the middleware?
With the above change I just wanted to make double sure nothing goes wrong if it ever would get such an exception passed.

ADmad: The unwrapping is surely not needed with current state of middleware code. If some extending middleware needs it (I don't see why) then it's up to them to update their code as required.

dereuromark: I think you are mistaken. The incoming one can very well be PHP7+. The wrapping currently only happens inside getRenderer() and thus does not affect the variable inside handleException().
As such you should revert as that code piece is very well required and should stay both for the class itself as for any extension to function properly.

PR #11519, opened 2017-12-07
Fix TimeHelper.php

savedario:
TimeHelper::format() arguments $date and $format had swapped positions.

It seems fixed already in the 3.x branch.


dereuromark:
I am not sure we can fix that in a patch release in a BC way. Most code will expect this order for the helper.
So we might have to document it better for now, not much we can do.

savedario:
Of course...
I'm not familiar with the process.
Who's changing the documentation etc... ?

markstory:
This is a breaking change for 2.x. We've updated 3.x because it was a major release and we could break compatibility.

Thank you for taking the time to fix this but we can't introduce these kinds of changes to 2.x.

PR #11521, opened 2017-12-09
Add to `203` and `204` allowed status code at `Cake\Http\Client\Response::isOk()`

gotoeveryone:
At the moment, `Cake\Http\Client\Response::isOk()` allow to status code between `200` and `202`.
But `Cake\TestSuite\IntegrationTestCase::assertResponseOk()` allow to status code between `200` and `204`.

For example, Status code `204` using to succeed response at `DELETE` method by API call.

#### Call to `http://example.com` with delete method.

- Use `Cake\Http\Client`

```php
// use Cake\Http\Client;
$http = new Client();

// Status code is `204`
$response = $http->delete('http://example.com', $data, $headers);
$body = $response->getBody();
$this->response = $this->response->withStatus($response->getStatusCode());

// `204` is false
if ($response->isOk()) {
    return 'OK';
}
return 'NG';
```

- Use `Cake\TestSuite\IntegrationTestCase`

```php
// TestCase
$this->enableCsrfToken();
$this->delete('http://example.com', $data);
$this->assertResponseOk(); // `204` is OK
```

### DIFF
--- a/src/Http/Client/Message.php
+++ b/src/Http/Client/Message.php
@@ -46,6 +46,20 @@ class Message
     const STATUS_ACCEPTED = 202;
 
     /**
+     * HTTP 203 code
+     *
+     * @var int
+     */
+    const STATUS_NON_AUTHORITATIVE_INFORMATION = 203;
+
+    /**
+     * HTTP 204 code
+     *
+     * @var int
+     */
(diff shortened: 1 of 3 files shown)


markstory:
Thank you!

PR #11522, opened 2017-12-09
Refactoring tests

carusogabriel:
Continuing to improve tests, like #11480 and #11498, this time I've used:
- `assertArrayHasKey` and `assertArrayHasKey` instead of `isset` function;
- `assertContains` instead of `strpos` function;
- `assertInstanceOf` instead of `instanceof` operator;
- `assertObjectNotHasAttribute` instead of `property_exists` function;


markstory:
Thanks again!

PR #11526, opened 2017-12-11
Make postConditions() less permissive.

markstory:
We were notified by `ooooooo_q` that postConditions() is vulnerable to SQL injection if used without SecurityComponent tampering prevention.

This change attempts to make postConditions() safer by exploding in unsafe scenarios.


chinpei215:
Looks good to me. I also think a permitted list is better than a ban list.

Issue #11520, opened 2017-12-08
Make it easier to override timestamp generation when using UrlHelper::assetUrl()


JoshuaLuckers:
This is a (multiple allowed):

* [ ] bug
* [x] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.5.6

### Enhancement details
I've set Asset.timestamp to "force" in app.php. When using UrlHelper to generate a URLs for asset files it adds a timestamp like it's intended to do. In certain situations I don't want to add a timestamp to the generated URL.

### What you did
When generating a URL for an asset file the method assetTimestamp is called. It checks the 'Asset.timestamp' configuration value to check if a timestamp should be added. 
When I don't want a timestamp for a certain asset URL I now do this:

```php
$orginalAssetTimestampConfiguration = Configure::read('Asset.timestamp');
Configure::write('Asset.timestamp',false);

$assetUrl = $this->Url->assetUrl('asset.ext');

Configure::write('Asset.timestamp',$orginalAssetTimestampConfiguration);
```

### Enhancement suggestion
Consider adding a parameter to the assetUrl function so we can overrule the 'Asset.timestamp' value set in the configuration file.

ADmad:
You can make a PR to add `assetTimestamp` config to `$options` argument and if set to `false` skip calling `assetTimestamp()` method.

JoshuaLuckers:
Ok! When I have some time this weekend I will create a pull request.

PR #11532, opened 2017-12-11
Possibilty to override the 'Asset.timestamp' value in Configure when generating assetUrls

JoshuaLuckers:
See issue #11520

### DIFF
--- a/src/View/Helper/UrlHelper.php
+++ b/src/View/Helper/UrlHelper.php
@@ -71,7 +71,7 @@ class UrlHelper extends Helper
      *   `fullBase` Return full URL with domain name
      *   `pathPrefix` Path prefix for relative URLs
      *   `plugin` False value will prevent parsing path as a plugin
-     *   `assetTimestamp` Overrides the value of `Asset.timestamp` in Configure.
+     *   `timestamp` Overrides the value of `Asset.timestamp` in Configure.
      *        Set to false to skip timestamp generation.
      *        Set to true to apply timestamps when debug is true. Set to 'force' to always
      *        enable timestamping regardless of debug value.
@@ -96,7 +96,7 @@ class UrlHelper extends Helper
      *   `pathPrefix` Path prefix for relative URLs
      *   `ext` Asset extension to append
      *   `plugin` False value will prevent parsing path as a plugin
-     *   `assetTimestamp` Overrides the value of `Asset.timestamp` in Configure.
+     *   `timestamp` Overrides the value of `Asset.timestamp` in Configure.
      *        Set to false to skip timestamp generation.
      *        Set to true to apply timestamps when debug is true. Set to 'force' to always
      *        enable timestamping regardless of debug value.
@@ -122,7 +122,7 @@ class UrlHelper extends Helper
      *   `pathPrefix` Path prefix for relative URLs
      *   `ext` Asset extension to append
      *   `plugin` False value will prevent parsing path as a plugin
-     *   `assetTimestamp` Overrides the value of `Asset.timestamp` in Configure.
+     *   `timestamp` Overrides the value of `Asset.timestamp` in Configure.
      *        Set to false to skip timestamp generation.
      *        Set to true to apply timestamps when debug is true. Set to 'force' to always
      *        enable timestamping regardless of debug value.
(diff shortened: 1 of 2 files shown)


src/View/Helper/UrlHelper.php
```diff
@@ -195,11 +218,14 @@ protected function _encodeUrl($url)
      * a timestamp will be added.
      *
      * @param string $path The file path to timestamp, the path must be inside WWW_ROOT
+     * @param bool|string $stamp If set will overrule the value of `Asset.timestamp` in Configure.
      * @return string Path with a timestamp added, or not.
      */
-    public function assetTimestamp($path)
+    public function assetTimestamp($path, $stamp = null)
```
ADmad: We can't change the method arguments in a point/bugfix release.

ADmad: You'll need to target your changes to `3.next` branch and they can be included in `3.6.0`.

JoshuaLuckers: I guess I did something wrong. Is it better to close this pull request and make a new one?

markstory: You can rebase your changes onto 3.next and the diff/commit list will be correct.

JoshuaLuckers: Thanks for pointing me to the right direction.

src/View/Helper/UrlHelper.php
```diff
@@ -71,6 +71,10 @@ public function build($url = null, $options = false)
      *   `fullBase` Return full URL with domain name
      *   `pathPrefix` Path prefix for relative URLs
      *   `plugin` False value will prevent parsing path as a plugin
+     *   `assetTimestamp` Overrides the value of `Asset.timestamp` in Configure.
```
ADmad: Given that all methods are for assets perhaps a shorter name `timestamp` would be enough.

src/View/Helper/UrlHelper.php
ADmad: Change the argument name `$stamp` to `$timestamp` for consistency?

JoshuaLuckers: I chose `$stamp` as argument name to keep it consistent with the variable used in the function.

dereuromark: You can easily fix up all of them to "timestamp".

ADmad: @JoshuaLuckers Internal variables don't show up in API docs but methods arguments and options do :)

src/View/Helper/UrlHelper.php
```diff
@@ -195,12 +218,15 @@ protected function _encodeUrl($url)
...
      * @param string $path The file path to timestamp, the path must be inside WWW_ROOT
+     * @param bool|string $timestamp If set will overrule the value of `Asset.timestamp` in Configure.
      * @return string Path with a timestamp added, or not.
      */
-    public function assetTimestamp($path)
+    public function assetTimestamp($path, $timestamp = null)
     {
-        $stamp = Configure::read('Asset.timestamp');
-        $timestampEnabled = $stamp === 'force' || ($stamp === true && Configure::read('debug'));
+        if (is_null($timestamp)) {
```
markstory: We generally use `$timestamp === null` as it doesn't use a function call.

tests/TestCase/View/Helper/UrlHelperTest.php
```diff
@@ -254,6 +254,16 @@ public function testAssetUrlTimestampForce()
...
 
+    public function testAssetTimestampConfigureOverride()
+    {
+        $this->Helper->webroot = '';
+        Configure::write('Asset.timestamp', 'force');
+        $timestamp = false;
+
+        $result = $this->Helper->assetTimestamp(Configure::read('App.cssBaseUrl') . 'cake.generic.css', $timestamp);
+        $this->assertEquals(Configure::read('App.cssBaseUrl') . 'cake.generic.css', $result);
+    }
```
markstory: We should also have tests for the new option in each of the public methods as that is new API surface area we're adding.

JoshuaLuckers: Fair enough! I will add those later this week.

markstory: Awesome!

PR #11539, opened 2017-12-13
bring .gitignore in line with cakephp/app's

inoas:
Ref: https://github.com/cakephp/app/blob/master/.gitignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,11 +11,6 @@
 debug.log
 error.log
 
-# IDE and editor specific files #
-#################################
-/nbproject
-.idea
-
 # OS generated files #
 ######################
 .DS_Store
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -26,3 +21,25 @@ error.log
...
+# Eclipse
+.settings/*
+# JetBrains, aka PHPStorm, IntelliJ IDEA
+.idea/*
+# NetBeans
+nbproject/*
+# Visual Studio Code
+.vscode
+# Sass preprocessor
+.sass-cache/
```
markstory: There is no sass in CakePHP.

ADmad: Core does does not have any sass files.

Issue #11524, opened 2017-12-10
4.0: Use same syntax for 18n and Cake sprintf functionality


dereuromark:
This is a (multiple allowed):

* [ ] bug
* [ ] enhancement
* [x] feature-discussion (RFC)

See https://github.com/cakephp/localized/pull/181#pullrequestreview-80384815

It is currently very annoying when working with translations that you cannot be sure if those are %s or {...}.
They would also currently have to be duplicated and translated twice.

### Option A
Use old sprintf() %s syntax again also for __() by default. With a bit of configuration that should be easily doable.

### Option B
Invent a Cake/I18n sprintf() that can work on the same {0...n} syntax as i18n __() can to consolidate this and make translating easier and more DRY.
Maybe a method `I18n::insert()` or sth could do the trick and can then be used in such cases instead of vanilla PHP.

Since the numbering has the advantage of re-ordering (which is probably necessary in same languages), I guess we would have to go with B.

markstory:
I would prefer us to use the `{0}` style placeholders, as the ICU format gives far more options when plural messages are involved.

ADmad:
I agree with @markstory, it would be unwise to limit usability but not using the ICU format for translation messages.

lorenzo:
I also agree with @markatory

dereuromark:
PR is open.

PR #11540, opened 2017-12-13
Provide consistent formatting besides the translating+formatting.

dereuromark:
This resolves a painpoint in terms of i18n localization and un-DRY strings.
See https://github.com/cakephp/localized/pull/181#pullrequestreview-80384815
Resolves https://github.com/cakephp/cakephp/issues/11524

Usage:
```php
$message = 'List {0}';
$args = 'Posts';
\Cake\I18n\I18n::getTranslator()->format($message, $args);
```
But a cleaner way would be to use the I18n wrapper directly:
```php
$message = 'List {0} of {1}';
$args = ['Posts', 'today'];
\Cake\I18n\I18n::format($message, $args)
```

So the core FormHelper and other classes can then be fixed from

    $actionName = __d('cake', 'New %s');
    $legend = sprintf($actionName, $modelName);

to

    $actionName = __d('cake', 'New {0}');
    $legend = I18n::format($actionName, $modelName);

and reuse the same strings

A small migration guide can be added to clarify the translations.
I will also add tests if this looks all good.

One could even completely use old %s syntax by translating `New {0}` to `New %s` and configurating i18n properly, so the internal formatting would use the sprintf() one then.

### DIFF
--- a/src/I18n/Formatter/FormatterInterface.php
+++ b/src/I18n/Formatter/FormatterInterface.php
@@ -1,0 +1,27 @@
+<?php
+/**
+ * CakePHP(tm) : Rapid Development Framework (https://cakephp.org)
+ * Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
+ *
+ * Licensed under The MIT License
+ * For full copyright and license information, please see the LICENSE.txt
+ * Redistributions of files must retain the above copyright notice.
+ *
+ * @copyright     Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
+ * @link          https://cakephp.org CakePHP(tm) Project
+ * @since         3.6.0
+ * @license       https://opensource.org/licenses/mit-license.php MIT License
+ */
+namespace Cake\I18n\Formatter;
+
+interface FormatterInterface
+{
+    /**
+     * Formats message
+     *
+     * @param string $message
+     * @param string|array $args
+     * @return string
+     */
+    public function format($message, $args);
+}
--- a/src/I18n/Formatter/FormatterInterface.php
+++ b/src/I18n/Formatter/FormatterInterface.php
@@ -19,8 +19,8 @@ interface FormatterInterface
     /**
      * Formats message
      *
-     * @param string $message
-     * @param string|array $args
+     * @param string $message The message content.
+     * @param string|array $args Argument or multiple arguments.
      * @return string
      */
     public function format($message, $args);
--- a/src/I18n/I18n.php
+++ b/src/I18n/I18n.php
@@ -17,9 +17,11 @@ namespace Cake\I18n;
 use Aura\Intl\FormatterLocator;
 use Aura\Intl\PackageLocator;
 use Cake\Cache\Cache;
+use Cake\I18n\Formatter\FormatterInterface;
 use Cake\I18n\Formatter\IcuFormatter;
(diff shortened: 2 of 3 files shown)


ADmad:
So are you proposing a new formatter interface to swap placeholders for different types of formatters? If so that seems odd to me.

Why can't we just change
```php
$actionName = __d('cake', 'New %s');
$legend = sprintf($actionName, $modelName);
```
to
```php
$legend = __d('cake', 'New {0}', $modelName);
```
?

We already have formatter classes under `Cake\I18n\Formatter` which implement `Aura\Intl\FormatterInterface`, so adding a new formatter interface at another level of abstraction is gonna complicate things.

ADmad:
By default `I18n` uses `IcuFormatter` so it seems an oversight (or intentionally left so for BC) the message strings in `FormHelper` use the `sprintf` placeholders.

So we either leave it this way until 4.0 or make the change I showed above in 3.6. There's no need for a new interface.

dereuromark:
I am not quite sold on that one. You are omitting one key factor: Parsability of the translation PO strings by an extractor tool.
If you can solve the following FormHelper code while doing so, I am all for it:
```php
            $actionName = __d('cake', 'New %s');
            $isCreate = $context->isCreate();
            if (!$isCreate) {
                $actionName = __d('cake', 'Edit %s');
            }
            $modelName = Inflector::humanize(Inflector::singularize($this->request->getParam('controller')));
            $legend = sprintf($actionName, $modelName);
```

But in some cases we need the internal (projected) formatter to run post translate, and for that this BC way on top would allow that, until a 4.0 can find a hopefully interfacy and more clean way then.

dereuromark:
OK, we could make the more a bit different:

```php
        $isCreate = $context->isCreate();
        $modelName = Inflector::humanize(Inflector::singularize($this->request->getParam('controller')));
        if (!$isCreate) {
            $legend = __d('cake', 'Edit {0}', $modelName);
        } else {
            $legend = __d('cake', 'New {0}', $modelName);
        }
```
But this function could still be useful probably. And make things easier for more complex cases where you wouldnt want to repeat this multiple times (or can't because it comes from dynamic content or DB).

markstory:
> this function could still be useful probably. 

I think we should understand the problems/scenarios this solves/makes better more before adding it. Right now the use cases for this method are largely covered by existing methods.

lorenzo:
> By default I18n uses IcuFormatter so it seems an oversight (or intentionally left so for BC) that the message strings in FormHelper use the sprintf placeholders.

It was left that way for BC

ADmad:
> It was left that way for BC

So we just need to decide whether to change it to use ICU placeholder in 3.6 or 4.0. I would suggest 4.0 but don't mind either way.

markstory:
I think adapting core messages to use ICU formatting is something that we could do during 4.x. My concern with doing it for 4.0 is that 4.0 is already rife with deprecations and we're aiming for a less painful upgrade process than 2.x to 3.x was.

dereuromark:
I still think we are then missing something for dynamically generated translations (not parseable via extract tool...). But I am fine with closing.

We should then fix it in 3.6 and migration guide addition, rather than having 4.0 blow up more, I agree.
This is something that currently creates more mistakes from having this inconsistency than quickly cleaning this up in a minor. Especially when you upgrade from 3.x and replace all i18n %s with {} only to find out that there are unfortunate exceptions. I can open a replacement PR then soon.

PR #11542, opened 2017-12-15
fixed hasPage(), params()

thienthansao29:
if current page have not paging ( one page ) , hasPage() return false


markstory:
If there are no paging parameters why are you using PaginatorHelper? Also there is 1 page, as you are likely looking at the first (and only) page.

markstory:
Closing as this breaks existing behaviour/tests and I am not clear on what problem is being solved.

PR #11549, opened 2017-12-16
Fix dispatched event info display in debug.

ADmad:
"foreach" can't be used for event list since it's not traversable.

### DIFF
--- a/tests/TestCase/Event/EventManagerTest.php
+++ b/tests/TestCase/Event/EventManagerTest.php
@@ -819,7 +819,7 @@ class EventManagerTest extends TestCase
         );
 
         $eventManager->setEventList(new EventList());
-        $eventManager->addEventToList(new Event('Foo'));
+        $eventManager->addEventToList(new Event('Foo', $this));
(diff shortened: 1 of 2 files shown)


tests/TestCase/Event/EventManagerTest.php
```diff
@@ -811,13 +811,30 @@ public function testDebugInfo()
             [
                 '_listeners' => [],
                 '_isGlobal' => false,
-                '_eventList' => null,
                 '_trackEvents' => false,
                 '_generalManager' => '(object) EventManager',
+                '_dispatchedEvents' => null,
```
markstory: Why the key rename?

ADmad: This is the key `__debugInfo()` meant to set when event list is not empty but never got set due to buggy code. I can change it to `_eventList` if you like.

markstory: No this is fine, I was just confused.

PR #11550, opened 2017-12-16
Fix few errors reported by phpstan level 3

ADmad:


src/TestSuite/TestCase.php
```diff
@@ -641,7 +641,7 @@ protected function skipUnless($condition, $message = '')
      * @param array $methods The list of methods to mock
      * @param array $options The config data for the mock's constructor.
      * @throws \Cake\ORM\Exception\MissingTableClassException
-     * @return \Cake\ORM\Table|\PHPUnit_Framework_MockObject_MockObject
```
dereuromark: Wont that lose IDE typehinting?

ADmad: Fixed

dereuromark: I dont think any IDE understands this new strange concept.

ADmad: Then submit an issue for your IDE :)  While new it's not strange, it's `intersection` instead of commonly used `union` type, explained by @ondrejmirtes [here](https://medium.com/@ondrejmirtes/union-types-vs-intersection-types-fd44a8eacbb).

ondrejmirtes: Well, you don't need to change this because of PHPStan, there's a hardcoded hack so that a union with `PHPUnit_Framework_MockObject_MockObject` is also interpreted as an intersection, because it's a common situation.

dereuromark: Doesnt matter if it is not supported by any IDE at all.
Breaks usability.. but well, guess the alternative is worse.

ADmad:
This is good to go, will send more in another batch.

Issue #10746, opened 2017-06-09
ORM issue in case of combining joinWith and Join.


skie:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.4.* (tested on master branch)

Here is the testcase that demonstrate the issue

```php
    public function testCombinedJoinWithAndJoins()
    {
        \Cake\Core\Plugin::load('TestPlugin');
        $Comments = TableRegistry::get('TestPlugin.Comments');
        $Comments->belongsTo('Articles', [
            'className' => 'Articles',
            'foreignKey' => 'article_id'
        ]); 		

        $Articles = TableRegistry::get('Articles');
        $Articles->hasMany('Comments', ['className' => 'Comments']);

         // this is working query 
        $items = $Comments->find('all')
            ->innerJoin(['Articles' => 'articles'], [
                'Comments.article_id = Articles.id'
            ])
            ->innerJoin(['Authors' => 'authors'], [
                'Articles.author_id = Authors.id'
            ])
            ->all()
            ->toArray();
        $this->assertNotEmpty($items);

        // now trying to perform the same with both JoinWith and Join
        $items = $Comments->find('all')
            ->innerJoinWith('Articles')
            ->innerJoin(['Authors' => 'authors'], [
                'Articles.author_id = Authors.id'
            ])
            ->all()
            ->toArray();
        $this->assertNotEmpty($items);		
    }
```

### What happened
When we use two joins we getting query
```sql
SELECT Comments.* FROM comments Comments
INNER JOIN articles Articles ON Comments.article_id = Articles.id
INNER JOIN authors Authors ON Articles.author_id = Authors.id
```
(I shortend select fields list to focus on real issue).
So in this case we have inner joins in correct order `articles, authors` - same we have in ORM builder.

When we combining the joinWith and join we getting sql 
```sql 
SELECT Comments.* FROM comments Comments
INNER JOIN authors Authors ON Articles.author_id = Authors.id
INNER JOIN articles Articles ON Articles.id = (Comments.article_id)
```

And in this case we have inner joins in other order `authors, articles` - same we have in ORM builder.

Seems it is caused by fact that joins and joinWith stored separately from each others.

### What you expected to happen
Order should be definetely keeped to allow ORM users combine both join and joinWith.

markstory:
[...]

PR #11552, opened 2017-12-16
Adds the possibility of using innerJoin after contain

DiogoRMax:
Example of use:

$Users->find()->contain(['Contacts'])->innerJoin(['CT' => 'contact_types'], ['CT.id = Contacts.contact_type_id']);

### DIFF
--- a/src/Database/Query.php
+++ b/src/Database/Query.php
@@ -12,6 +12,7 @@
  * @since         3.0.0
  * @license       https://opensource.org/licenses/mit-license.php MIT License
  */
+
 namespace Cake\Database;
 
 use Cake\Database\Expression\OrderByExpression;
@@ -41,6 +42,13 @@ class Query implements ExpressionInterface, IteratorAggregate
      * @var \Cake\Database\Connection
      */
     protected $_connection;
+
+    /**
+     * Enable smart ordering of JOIN.
+     *
+     * @var bool
+     */
+    protected $_smartJoin = false;
 
     /**
      * Type of this query (select, insert, update, delete).
@@ -1927,6 +1935,31 @@ class Query implements ExpressionInterface, IteratorAggregate
     }
 
     /**
+     * Toggle smart join.
+     *
+     * If set to true, enables smart sorting in query compilation.
+     *
+     * @param bool $enable Use a boolean to set the smart join mode.
+     * @return $this
+     */
+    public function enableSmartJoin($enable = true)
+    {
+        $this->_smartJoin = (bool)$enable;
+
+        return $this;
+    }
+
+    /**
+     * Returns the current hydration mode.
+     *
+     * @return bool
+     */
+    public function isSmartJoin()
+    {
+        return $this->_smartJoin;
+    }
+
+    /**
      * Enable/Disable buffered results.
      *
      * When enabled the results returned by this Query will be
@@ -1965,6 +1998,7 @@ class Query implements ExpressionInterface, IteratorAggregate
 
         return $this;
     }
+
     /**
      * Gets the TypeMap class where the types for each of the fields in the
      * select clause are stored.
--- a/src/Database/QueryCompiler.php
+++ b/src/Database/QueryCompiler.php
@@ -12,6 +12,7 @@
  * @since         3.0.0
  * @license       https://opensource.org/licenses/mit-license.php MIT License
  */
+
 namespace Cake\Database;
 
 use Cake\Database\Expression\QueryExpression;
@@ -223,6 +224,9 @@ class QueryCompiler
      */
     protected function _buildJoinPart($parts, $query, $generator)
     {
+        if ($query->isSmartJoin())
+            return $this->_buildJoinPartSmartly($parts, $query, $generator);
+
         $joins = '';
         foreach ($parts as $join) {
             $subquery = $join['table'] instanceof Query || $join['table'] instanceof QueryExpression;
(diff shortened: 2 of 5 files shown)


inoas:
Most important would be unit tests that proof that this did not work before and that proof that it works afterwards I suppose as changes to the ORM can break real world apps quite easily.

DiogoRMax:
Excuses I do not know how to do these unit tests.

However the sql generated when using the current ORM is like:

> SELECT * 
> FROM users Users
> 	LEFT JOIN contact_types CT ON (
> 		CT.id = Contacts.contact_type_id
> 	)
> 	LEFT JOIN contacts Contacts ON (
> 		Contacts.user_id = (Users.id)
> 	)

What causes error

With the change it puts the innerJoin afterwards like this:

> SELECT * 
> FROM users Users
> 	LEFT JOIN contacts Contacts ON (
> 		Contacts.user_id = (Users.id)
> 	)
> 	LEFT JOIN contact_types CT ON (
> 		CT.id = Contacts.contact_type_id
> 	)


If you can help me with the process, I'm grateful ....

markstory:
Your changes [cause existing tests to fail](https://travis-ci.org/cakephp/cakephp/jobs/317401087) so there is likely a regression against existing behavior being introduced.

src/ORM/Query.php
```diff
@@ -1059,12 +1059,20 @@ protected function _transformQuery()
...
         if (empty($this->_parts['from'])) {
             $this->from([$this->_repository->getAlias() => $this->_repository->table()]);
         }
         $this->_addDefaultFields();
         $this->getEagerLoader()->attachAssociations($this, $this->_repository, !$this->_hasFields);
         $this->_addDefaultSelectTypes();
+
+        $i = count($this->_parts['join']);
+        foreach ($joins as $a => $j) {
+            $this->_parts['join'][is_string($a) ? $a : $i++] = $j;
```
markstory: Couldn't this result in joins created by the eagerloader being overwritten?

DiogoRMax: The problem is in the order in which JOINS are created.

The eagerloader is processed last.
Indepedent order to call the method.

DiogoRMax: The problem is in the order in which JOINS are created.

The eagerloader is processed last.
Indepedent order to call the method.

DiogoRMax:
The problem is in the order in which JOINS are created.

The eagerloader is processed last.
Indepedent order to call the method.

DiogoRMax:
Is there no way that the call order can be respected?

DiogoRMax:
I believe adding the respect to the invocation order of the method makes the ORM much more powerful and versatile.

But I also agree that it may end up breaking some apps.

The alternative would be to create a smart computer, which takes into account the dependency of JOINS.

If this change can be accepted, I can create the code.

inoas:
Are you certain the order of non-nested LEFT-JOINS matters as shown in your examples? As far as I know it does not.

lorenzo:
Have you tried using `matching()` or `innerJoinWith()`? I believe either of those will solve your problem

DiogoRMax:
Sorry for the inconvenience, I'm new to this business.

DiogoRMax:
How can errors occur in the tests, and the code added, should not change the flow of the program in any way?

src/Database/Query.php
```diff
@@ -1926,6 +1934,31 @@ public function isBufferedResultsEnabled()
...
+     */
+    public function enableSmartJoin($enable = true)
+    {
+        $this->_smartJoin = (bool)$enable;
+
+        return $this;
+    }
+
+    /**
+     * Returns the current hydration mode.
```
markstory: Documentation needs to be updated.

src/Database/QueryCompiler.php
```diff
@@ -250,6 +255,71 @@ protected function _buildJoinPart($parts, $query, $generator)
...
+            $alias[trim($join['alias'], '`')] = $join['alias'];
+            $alias[trim($join['table'], '`')] = $join['alias'];
+        }
+
+        $part = '';
+        foreach ($order as $alias) {
+            $part .= $joins[$alias];
+        }
+
+        return $part;
```
markstory: This is a significant amount of code without any tests. Adding new modes increases complexity quite a bit as this mode can be combined with every other possible state. Is there a way to accomplish what you are trying to do without additional modes?

DiogoRMax: I can not imagine any solution that will not break applications already deployed in the real world.

My point is:

$query->innerJoin(A...)->innerJoin(B...)
does not generate the same SQL as:
$query->innerJoin(B...)->innerJoin(A...)

But ...

$query->contain(A...)->innerJoin(B...)
$query->innerJoin(B...)->contain(A...)
Generates the same SQL, always leaving the JOIN part of the 'contain' at the end.

DiogoRMax: Has anyone ever had this difficulty?

The only way I can see not to break the implementations already made, is to create an intelligent ordering system for the JOIN part.

DiogoRMax: Or maybe create some parameter that holds the order of the containsJOINs

markstory: Have you tried replacing `innerJoin` with `innerJoinWith` as @lorenzo suggested earlier? `innerJoinWith` uses the eagerloader so it should preserve join ordering.

> Has anyone ever had this difficulty?

Yes a very similar issue was reported in #10746 but there are a few solutions to that problem.

DiogoRMax:
Thank you...
But my project does not understand the use of innerJoinWith, and thinking about it is a particularity that does not have to be included in Cakephp.

I have solved my problems with other solutions, thanks for the attention ...

PR #11554, opened 2017-12-16
Add missing EventInterface to ease future typehinting using interfaces.

dereuromark:
We should have added the interface a long type ago instead of forcing to typehint against a concrete class.
It would also have less side-effects, e.g. clashing with Event class as entity (since we unfortunately dont use suffixing here), if we start to allow users having their callbacks switch from
```php
beforeSave(Event $event, EntityInterface $entity, ArrayObject $options)
```
to
```php
beforeSave(EventInterface $event, EntityInterface $entity, ArrayObject $options)
```

Notes:
* We should be fairly certain that the API in its current form (getter/setters) is now stable and will not change again. This will ease 4.0 migration then, no additional refactor necessary, and the interface stays the same.

### DIFF
--- a/src/Event/Event.php
+++ b/src/Event/Event.php
@@ -14,17 +14,7 @@
  */
 namespace Cake\Event;
 
-/**
- * Represents the transport class of events across the system. It receives a name, subject and an optional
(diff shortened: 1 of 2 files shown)


markstory:
Great idea!

PR #11555, opened 2017-12-16
Consolidate placeholders to new syntax. Eases 4.x migration.

dereuromark:
Follow up on https://github.com/cakephp/cakephp/pull/11540
Resolves https://github.com/cakephp/cakephp/issues/11524

Docs: https://github.com/cakephp/docs/pull/5478

Consolidating translation strings.

This eases migration for 4.0 as this can be cleaned up in 3.6 with migration guide.
Also less confusion for those upgrading from 2.x and str-replacing all placeholders to discover that some do not function as expected.

### DIFF
--- a/src/View/Helper/FormHelper.php
+++ b/src/View/Helper/FormHelper.php
@@ -1046,13 +1046,13 @@ class FormHelper extends Helper
         }
 
         if ($legend === true) {
-            $actionName = __d('cake', 'New %s');
             $isCreate = $context->isCreate();
-            if (!$isCreate) {
-                $actionName = __d('cake', 'Edit %s');
-            }
             $modelName = Inflector::humanize(Inflector::singularize($this->request->getParam('controller')));
-            $legend = sprintf($actionName, $modelName);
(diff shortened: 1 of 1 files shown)


inoas:
IMHO we should however keep a list of the %s strings in some deprecated dummy php file just so that i18n extract can pick them up when updating cakephp/localized so that we can for some time still support 3.0 < 3.6?

dereuromark:
Sure, also cakephp/localized can still contain them as dummy content.

inoas:
So can you add `src/View/Helper/FormHelper_deprecated_translations.php` to this PR
```php
<?php
// @deprecated since 3.6, will be removed in 4.0, use {0} syntax instead.
__d('cake', 'New %s');
__d('cake', 'Edit %s');
```

dereuromark:
No, in this PR it makes no sense, once you cross to 3.6 and applied the migration you need to handle this old one on your own.
This is only required if you use an older CakePHP version and here localized plugin can handle this extra ones.

Issue #11545, opened 2017-12-15
Flash Element for Success and Error not working properly


rafaeldotjs:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.5.3
* Platform and Target: Wamp Server x64

### What you did
Created a prefix /admin/, website is inside a subfolder /new/, I have this in my controller: 

`$this->Flash->success(__('The service has been saved.'));`

I placed a Flash/success element in my index action:
`<?= $this->element('Flash/success'); ?>`

the element file, in Template/Element/Flash/success.ctp :
```
<?php
if (!isset($params['escape']) || $params['escape'] !== false) {
    $message = h($message);
}
?>
<div class="message success" onclick="this.classList.add('hidden')"><?= $message ?></div>
```
### What happened
The flash message doesn't print, it gives an error because the $message var is undefined. The same happened to the error element.

> Notice (8): Undefined variable: message [APP/Template\Element\Flash\success.ctp, line 3]

### What you expected to happen
I expected it to print the Success/Error Flash message.

ADmad:
You are supposed to use the Flash helper to render flash messages, `$this->Flash->render()`.

rafaeldotjs:
I understand that. But since I have $this->Flash->success('message'), and in the Flash component it gives it the option "Flash/success", shouldn't the default code of Element/Flash/success.ctp print the success messages? I might not be getting something here, but why does the Flash component points to an element file, and the element file does not print the message? If it is not a bug or incomplete feature, what is it? I asked about this in cakephp slack, didn't find an answer. I searched google, didn't find a solution.

If I use $this->Flash->render() in Flash/success.ctp, it still does not work, unless I hack the Flash component, like I did.

Can we talk a little longer before closing it? If indeed this is not a bug, can you explain clearly the solution? Because bug or not, it is definitely not clear how to proceed. Am I expecting something of the Flash/success.ctp that it is not made to do?

Thank you

ADmad:
> But since I have $this->Flash->success('message'), and in the Flash component it gives it the option "success" as element, shouldn't the default code of Element/Flash/success.ctp print the success messages?

No it won't, since the component only sets the data to session. The flash elements needs required template variables to be set based on the session data, which is done by the FlashHelper. So you can't just set message through component and expect the message to show up properly by directly including the flash element.

This is not a help forum. The ticket tracker is reserved for possible
[...]

PR #11557, opened 2017-12-18
Use term "template" instead of "view".

ADmad:
Refs #11545


inoas:
Can we generally stick to replacing "view" with "view template" in docs and "view" with "template" in code to make it easy for people migrating to pick up our new nomenclature?

PR #11564, opened 2017-12-21
Implement Plugin Classes

burzum:
* Plugins have a class that describes bootstrapping and routes
* Plugin objects are attached to a plugin registry in Application.php

This is still WIP but I opened the PR already to get some feedback before going on with this.


src/Core/PluginRegistry.php
```diff
@@ -0,0 +1,172 @@
...
+    /**
+     * Define the console commands for an application.
+     *
+     * @param \Cake\Console\CommandCollection $commands The CommandCollection to add commands into.
+     * @return \Cake\Console\CommandCollection The updated collection.
+     */
+    public function console($commands)
+    {
+        foreach ($this as $plugin) {
+            if ($plugin instanceof ConsoleApplicationInterface) {
```
ADmad: Since you have `PluginApp implements ConsoleApplicationInterface` doesn't this check become redundant? IMO the base class should not implement `ConsoleApplicationInterface`, only plugins which provide console commands should implement that interface.

src/Core/PluginApp.php
```diff
@@ -0,0 +1,144 @@
...
+     */
+    public function initialize()
+    {
+
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function disableRoutes($disabled)
```
ADmad: It would be better to have 2 separate methods `disableRoutes()` and `enableRoutes()`.

src/Http/BaseApplication.php
```diff
@@ -117,4 +168,14 @@ protected function getDispatcher()
     {
         return new ActionDispatcher(null, null, DispatcherFactory::filters());
     }
+
+    /**
+     * Plugins
+     *
+     * @return \Cake\Core\PluginRegistry
+     */
+    public function pluginRegistry()
```
ADmad: `getPluginRegistry()`? :)

tests/TestCase/Core/PluginRegistryTest.php
```diff
@@ -0,0 +1,68 @@
...
+        //$this->assertInstanceOf(TestPlugin::class, $result);
+
+        $result = $registry->load('Company\TestPluginFive');
+        //$this->assertInstanceOf(TestPlugin::class, $result);
+        debug($result);
+        */
+    }
+
+    /**
+     * @expectedException \Error
```
ADmad: This won't work on PHP 5.6 since it does not have an `Error` class.
https://travis-ci.org/cakephp/cakephp/jobs/319198618#L917

src/Core/PluginRegistry.php
```diff
@@ -0,0 +1,172 @@
...
+    /**
+     * Define the HTTP middleware layers for an application.
+     *
+     * @param \Cake\Http\MiddlewareQueue $middleware The middleware queue to set in your App Class
+     * @return \Cake\Http\MiddlewareQueue
+     */
+    public function middleware($middleware)
+    {
+        foreach ($this as $plugin) {
+            $middleware->add($plugin);
```
markstory: Is the thinking here that plugins will *be* middleware? Or that plugins can *add* middleware. I think we want plugins to *add* middleware, which would look like

```php
foreach ($this as $plugin) {
  if ($plugin instanceof HttpApplicationInterface) {
    $middleware = $plugin->middleware($middleware);
  }
}
return $middleware
```

src/Core/PluginRegistry.php
```diff
@@ -0,0 +1,172 @@
...
+     *
+     * @param \Psr\Http\Message\ServerRequestInterface $request The request
+     * @param \Psr\Http\Message\ResponseInterface $response The response
+     * @param callable $next The next middleware
+     * @return \Psr\Http\Message\ResponseInterface
+     */
+    public function __invoke(ServerRequestInterface $request, ResponseInterface $response, $next)
+    {
+        foreach ($this as $plugin) {
+            $response = $plugin($request, $response, $next);
```
markstory: This implementation is at odds with the `middleware()` method. I'm not convinced that plugin classes should *be* middleware yet.

src/Http/BaseApplication.php
```diff
@@ -73,6 +120,8 @@ public function routes($routes)
         if (!Router::$initialized) {
             require $this->configDir . '/routes.php';
             // Prevent routes from being loaded again
+
+            $this->pluginRegistry()->routes($routes);
```
markstory: I would prefer to see us load routes in the application's routes file. This allows the app developer the ability to add additional paths/parameters via scopes. In 3.5 we added `RouteBuilder::loadPlugin()` to load routes from a single plugin. If the route builder had access to the plugin registry, or application instance we'd be able to leverage that existing interface.  With that said we may also need a `$routes->loadPlugins()` to cover the case where people just don't want to customize routes and just load them all.

src/Core/PluginRegistry.php
burzum: Correct @robertpustulka

src/Http/BaseApplication.php
burzum: Any suggestion how to have a single plugin registry and share it between the router and the app without having a global DI container?

src/Core/PluginRegistry.php
burzum: OK, but the method must exists because of the interfaces.

thinkingmedia:
It's always great to see people working on new features, but PR's make it difficult to have a discussion. You can keep this PR up (absolutely nothing wrong with that), but if you want *feedback* it's better to create an issue and link it to this PR.

When you open the issue. Besure to explain how your changes make life easier for your fellow Cake developers, or explain what kind of problem you're solving.

src/Core/PluginRegistry.php
ADmad: Gotcha, my bad.

src/Http/BaseApplication.php
markstory: We could hide the PluginRegistry inside `Plugin`. That would let us make the new/old ways to work with plugins easier to deal with.

> I understand your arguments but it don't think this functionality should exist in two places?

I don't think we should plan to have the same functionality in two places as a permanent solution. Doing so requires us to document when to use each solution and ensure that both of them continue working.

> Also it's already loading the routes if the plugin is configured to load them? I don't see the issue except that we would lose a previously added methods feature?

The routebuilder methods make it easier to customize plugin routes with additional scopes/paths. For example

```php
$routes->scope('/tools', function ($routes) {
  $routes->loadPlugin('DebugKit');
});
```
Is not possible with the current implementation you have. This approach prevents plugin routes to be placed above routes defined in the application's route list. Placing frequently used routes higher in the file helps improve parse performance as fewer routes are checked on incoming requests.

src/Core/PluginInterface.php
```diff
@@ -30,18 +30,30 @@ public function bootstrap();
...
      */
-    public function disableRoutes($disabled);
+    public function disableRoutes();
+
+    /**
+     * Enables route loading for the plugin
+     *
+     * @return $this
+     */
+    public function enableRoutes();
```
markstory: I like these method names :clap:

src/Http/BaseApplication.php
robertpustulka: From the lack of a better designed solution making `Plugin` a “facade” could work for the time being. We can rethink that in the future (4.x/5.x) when we have a DIC built in.

burzum: >  In 3.5 we added RouteBuilder::loadPlugin() to load routes from a single plugin. 

Can't we check in the method if the first passed was a plugin registry object and then use the 2nd args as string for the name?

> $routes->loadPlugins()

Could take the plugin registry as arg as well.

src/Core/PluginRegistry.php
```diff
@@ -0,0 +1,194 @@
...
+    {
+        $plugin = parent::load($objectName, $config);
+
+        if (class_exists('\Cake\Event\Event')) {
+            $event = new Event('Plugin.loaded', $this, [
+                'name' => $objectName,
+                'plugin' => $plugin,
+                'config' => $config
+            ]);
+            EventManager::instance()->dispatch($event);
```
ADmad: Once #11692 is done the application event manager instance could be injected into plugin registry instead of fetching the global instance through static method.

burzum: This might become a little off topic but by looking at this PR and the use case(s), I think we begin to use it like a DI container? Wouldn't it be better to wait for that?

markstory: I think we can separate off events from this. #11692 will at solve several event manager related requirements.

markstory:
I've pushed up a bunch of changes that make this code more inline with what was documented in the related issue. I've not fully implemented the `addPlugin()` method yet, but will be doing that shortly.  What is 'done' is the plugin collection (instead of a registry) and the plugin hooks being called. Still to be done are:

* [ ] Actually loading plugins.
* [ ] Options and methods for the other plugin hooks
* [ ] Injecting the plugin collection in to `Plugin` for backwards compatibility
* [ ] Connecting the plugin collection into the route builder so that `RouteBuilder::loadPlugin()` works as it should.

src/Core/PluginApp.php
```diff
@@ -0,0 +1,236 @@
...
+    protected $consoleEnabled = true;
+
+    /**
+     * Constructor
+     *
+     * @param array $options Options
+     */
+    public function __construct(array $options = [])
+    {
+        foreach (['bootstrap', 'routes', 'console', 'middleware'] as $key) {
```
robertpustulka: Can use [`VALID_HOOKS`](https://github.com/cakephp/cakephp/pull/11564/files#diff-56af3672217e52636f93d3273b3daa1fR38) const here

markstory: Good idea. Fixed.

src/Core/PluginInterface.php
```diff
@@ -0,0 +1,141 @@
...
+    public function getClassPath();
+
+    /**
+     * Load all the application configuration and bootstrap logic.
+     *
+     * Override this method to add additional bootstrap logic for your application.
+     *
+     * @return void
+     */
+    public function bootstrap();
```
robertpustulka: Should this interface define other hooks (`console` etc) as well? 

Or maybe it should be split into `ConsolePlugin` and `HttpPlugin` interfaces? I'm wondering if we should keep different interfaces for plugins and regular app. App is nicely  defined through interfaces (console, http, upcoming events) but here we just throw everything into one bucket.

src/Http/BaseApplication.php
```diff
@@ -52,6 +68,63 @@ public function __construct($configDir)
      */
     abstract public function middleware($middleware);
 
+    /**
+     * {@inheritDoc}
+     */
+    public function pluginMiddleware($middleware)
+    {
+        foreach ($this->plugins->with('middleware') as $plugin) {
+            $middleware = $plugin->middleware($middleware);
```
robertpustulka: This assumes that the `middleware()` method is implemented by the plugin app class, but the interface does not define it. This applies to other hooks as well.
Either we should add hook methods into the `PluginInterface` or check for methods existence inside `with()` method.

src/Core/PluginCollection.php
```diff
@@ -0,0 +1,151 @@
...
+class PluginCollection implements IteratorAggregate, Countable
+{
+    /**
+     * Plugin list
+     *
+     * @var array
+     */
+    protected $plugins = [];
+
+    const VALID_HOOKS = ['routes', 'bootstrap', 'console', 'middleware'];
```
robertpustulka: Should we add another valid hook into the array once #11692 is merged?

robertpustulka (commented):
I'm not sure we should mix http with console in one `PluginInterface`. This seems inconsistent with the design of current application interfaces.

robertpustulka:
Would it make sense to support any kind of hook by the plugin app? Instead of having concrete `enableRoutes()` etc methods we could have `enable($hook)`, `isEnabled($hook)` etc similar to the `with($hook)`. This way the interface wouldn’t mix the CLI and web context and allow plugins to hook into authentication for example.

src/Core/PluginInterface.php
markstory: What benefits are there in having 3 interfaces for plugins? With the application interfaces it enabled standalone console applications to not concern themselves with http. This may also have been a side-effect of the `BaseApplication` being in the Http namespace. With plugins the base class is in Core along with the interface.

src/Http/BaseApplication.php
markstory: middleware missing from the interface is a mistake. I'll fix that up.

src/Core/PluginCollection.php
markstory: Yes

robertpustulka:
Recent changes look good :+1: 

There are number of PHPStan reported issues though.

markstory:
Thank you for fixing the phpstan errors @robertpustulka 👏

markstory:
@robertpustulka I've added in an `events` hook for plugins now.

markstory:
> EventApplicationInterface returns the manager instance. Do we want to do the same here?

We could, but then callers would need to overwrite the manager on themselves and the application to ensure that state is consistent. Overwriting event managers also runs the risk of dropping listeners in the case a plugin accidentally creates a new manager. Should we remove the overwriting from the application events hook?

robertpustulka:
> Overwriting event managers also runs the risk of dropping listeners in the case a plugin accidentally creates a new manager.

The same can happen with routes or middleware as the hooks must return objects. I think that behavior of all hooks that modify an object should be similar.

src/Http/BaseApplication.php
```diff
@@ -61,6 +78,75 @@ public function __construct($configDir, EventManagerInterface $eventManager = nu
      */
     abstract public function middleware($middleware);
 
+    /**
+     * {@inheritDoc}
+     */
+    public function pluginEvents()
```
robertpustulka: I think that this method should behave similar to the `pluginMiddleware` and `pluginRoutes` ie accept an object and return the modified one.

The risk of dropping changes if a new instance is returned by mistake is the same as with the middleware and routes.

dereuromark: Immutable approach of always using return value to continue sounds indeed more consistent.

markstory: Ok. I will get this updated. Do you also think all the new hooks should have typehints added where we have existing interfaces?

src/Core/PluginInterface.php
```diff
@@ -0,0 +1,117 @@
...
+     */
+    public function middleware($middleware);
+
+    /**
+     * Add routes for the plugin.
+     *
+     * @param \Cake\Routing\RouteBuilder $routes The route builder to update.
+     * @return \Cake\Routing\RouteBuilder
+     */
+    public function routes($routes);
```
dereuromark: Could we use some interface of RouteBuilder here as typehint?

src/Core/PluginInterface.php
```diff
@@ -0,0 +1,117 @@
...
+
+    /**
+     * Add events for the plugin.
+     *
+     * @param \Cake\Event\EventManager $events The application's event manager
+     * @return void
+     */
+    public function events($events);
```
dereuromark: Could we use some interface of EventManager here as typehint?

src/Core/PluginInterface.php
markstory: An interface for RouteBuilder seems wasteful to me as it is not something we expect/want users to reimplement.

src/Http/BaseApplication.php
markstory: I've added typehints for `EventManagerInterface` as it was the only existing interface.

markstory:
What do folks think about the name `PluginApp`? Is `ApplicationPlugin` a better name?

What about the informal convention of having `src/Plugin.php` in each plugin. Should that instead be `src/Debugkit.php` where 'DebugKit' is the plugin's name?

rchavik:
@markstory what about namespaced plugin?

burzum:
@markstory if we move the whole plugin related classes into `Cake\Plugin`? We can have a `Plugin` class there as well then and deprecate the one in `Cake\Core`?

Let's use `ApplicationPlugin`. And I prefer to have a file `Plugin` in each plugin, makes it easy to remember and spot and also tells you what it is about.

src/Http/BaseApplication.php
robertpustulka: @markstory `pluginEvents()` hasn't been updated with type hints.

src/Core/PluginInterface.php
dereuromark: OK

saeideng:
>having src/Plugin.php in each plugin

src/Http/BaseApplication.php
```diff
@@ -61,6 +78,79 @@ public function __construct($configDir, EventManagerInterface $eventManager = nu
...
+    /**
+     * Create a plugin instance from a classname and configuration
+     *
+     * @param string $name The plugin classname
+     * @param array $config Configuration options for the plugin
+     * @return \Cake\Core\PluginInterface
+     */
+    public function makePlugin($name, array $config)
+    {
+        if (!class_exists($name)) {
```
ADmad: Could just check if name does not have backlash instead of class_exists(), similar to what App::className() does.

src/Http/BaseApplication.php
```diff
@@ -61,6 +78,79 @@ public function __construct($configDir, EventManagerInterface $eventManager = nu
...
+            $name = str_replace('/', '\\', $name) . '\\' . 'Plugin';
+        }
+        if (!class_exists($name)) {
```
ADmad: Wouldn't PHP already generate a good enough error message if you try to instantiate a non existent class below?

markstory: Could, I thought this error was better though.

src/Http/BaseApplication.php
robertpustulka: > It also doesn't take parameters or return a value either though.

`pluginRoutes` and `pluginMiddleware` take an argument and return the modified object. I don't like that `pluginEvents` is different. Either we remove that from routes and middleware, or add to events.

> Perhaps the Server and CommandRunner should drop their event managers and instead proxy the application manager?
In this case the application would need to support events. This is optional now, as http and console application interfaces don't extend `EventDispatcherInterface`.

robertpustulka: > It also doesn't take parameters or return a value either though.

`pluginRoutes` and `pluginMiddleware` take an argument and return the modified object. I don't like that `pluginEvents` is different. Either we remove that from routes and middleware, or add to events.

Also having `pluginEvents` with no arguments assumes that application that implements `PluginApplicationInterface` must have it's own event manager. Executing that method on an "eventless" application instance would take no effect. The idea of application "hooks" in my opinion is to take an object, modify it and return the modified instance.

> Perhaps the Server and CommandRunner should drop their event managers and instead proxy the application manager?

In this case the application would always need to support events. This is optional now, as http and console application interfaces don't extend `EventDispatcherInterface`.

robertpustulka: I feel like I need to explain my point a bit more.

In "isolation" application does not dispatch events by default (it doesn't have to implement the `EventDispatcherInterface`), the Server and CommandRunner do. So in this scenario event hooks should be executed on Server's and CommandRunner's event manager. So both `events` and `pluginEvents` must take an argument in order to modify the server/runner's event manager. 
If application does implement `EventDispatcherInterface` it's event manager is injected into the server/runner in `setApp` setter. The process of applying hooks on an event manager is still the same, but the subject (event manager instance) is different now (app's event manager)

markstory: I hear you on it feeling inconsistent, my concern is how much logic adding a parameter/return value forces onto the caller of `pluginEvents()`. For example the bootstrap method on both `Server` and `CommmandRunner` would need to look like:

```php
        $this->app->bootstrap();
        if ($this->app instanceof EventApplicationInterface) {
            $eventManager = $this->app->events($this->getEventManager());
            $this->setEventManager($eventManager);
        }
        if ($this->app instanceof PluginApplicationInterface) {
            $this->app->pluginBootstrap();
            if ($this->app instanceof EventApplicationInterface) {
                $events = $this->app->getEventManager();
                $events = $this->app->pluginEvents($events);

                $this->app->setEventManager($events);
                $this->setEventManager($events);
            }
        }
```

To me this puts a lot of onus on the caller and breaks the encapsulation that the interfaces should be trying to provide. Perhaps the `PluginApplicationInterface`should include/consume the `EventApplicationInterface` and `EventManagerInterface`? That would allow us to have fewer checks across the framework. It would prevent 'only' using events and not plugins, but I think that is a reasonable tradeoff to make for reduced complexity.

Joining the interfaces may also allow us to make proxying eventmanager methods simpler on `Server` and `CommandRunner`. The `getEventManager()` method could have an implementation that looks something like:

```php
public function getEventManager()
{
  if ($this->app instanceof PluginApplicationInterface) {
    return $this->app->getEventManager();
  }
  return EventManager::instance();
}
```

markstory:
@robertpustulka I've updated these changes to include the merged interface and proxies we talked about today.

markstory:
@robertpustulka I've updated these changes to include the merged interface and proxies we talked about today.

src/Core/PluginApplicationInterface.php
```diff
@@ -0,0 +1,80 @@
...
+     */
+    public function pluginEvents(EventManagerInterface $eventManager);
+
+    /**
+     * Application hook for attaching events to the Application event manager instance.
+     *
+     * @param \Cake\Event\EventManagerInterface $eventManager Event manager instance.
+     * @return \Cake\Event\EventManagerInterface
+     */
+    public function events(EventManagerInterface $eventManager);
```
robertpustulka: Since now `PluginApplicationInterface` extends `EventDispatcherInterface` would it make sense to remove argument and return form `events ` and `pluginEvents` method? App that supports event hooks now has it's own event manager and listeners should be attached to that instance anyway.

```php
// App\Application.php
public function events()
{
    $this->getEventManager()->on(new MyListener);
}
```

and

```php
public function pluginEvents()
{
    $events = $this->getEventManager();
    foreach ($this->plugins->with('events') as $plugin) {
        $events = $plugin->events($events); //this stays as it is, plugin doesn't dispatch events
    }
    $this->setEventManager($events);
}
```

This way event handling could be even more encapsutaled.

markstory: I'm ok with that. I added the argument to make the routes, events, and middleware hooks consistent in that they all get a 'thing' and return that 'thing'.  Making them 'closed' would add inconsistency with other hooks but simplify framework code.

robertpustulka: Yeah, I know I’ve been advocating for consistency but now with the current API this makes little sense to have event hooks get and return something since it’s going to be always an app’s dependency.

PR #11571, opened 2017-12-23
Fix #11568

### LINKED ISSUE #11568: Truncate html text with style tag
* [x] bug
* [x] enhancement

CakePhp v 3.5.8

When i truncate html text with style tag, the content of style tag is counted.
So the returned string contain only style tag content because my desired length is shoten than style tag content.
It's normaly or it's a forgotten (so i can propose a pull request) ?

maratth:
- Added var for default excluded html tags and getter/setter
- Added check in truncate html tags loop, for seeing if tag is excluded or not

### DIFF
--- a/src/Utility/Text.php
+++ b/src/Utility/Text.php
@@ -619,7 +619,7 @@ class Text
             preg_match_all('/(<\/?([\w+]+)[^>]*>)?([^<>]*)/', $text, $tags, PREG_SET_ORDER);
             foreach ($tags as $tag) {
                 $contentLength = 0;
-                if (!in_array($tag[2], static::$_defaultHtmlNoCount)) {
+                if (!in_array($tag[2], static::$_defaultHtmlNoCount, true)) {
                     $contentLength = self::_strlen($tag[3], $options);
                 }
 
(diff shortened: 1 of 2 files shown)


src/Utility/Text.php
```diff
@@ -1120,4 +1133,25 @@ public static function slug($string, $options = [])
...
+    }
+
+    /**
+     * Set html tags not counted for truncate.
+     *
+     * @param array $htmlNoCount New html tags not counted for truncate.
+     * @return void
+     */
+    public static function setHtmlNoCount($htmlNoCount)
+    {
```
markstory: Can we have a typehint here to ensure the parameter is an array?

markstory (commented):
Do we need the new methods? Adding new methods would push this into 3.6.

src/Utility/Text.php
```diff
@@ -608,7 +618,10 @@ public static function truncate($text, $length = 100, array $options = [])
 
             preg_match_all('/(<\/?([\w+]+)[^>]*>)?([^<>]*)/', $text, $tags, PREG_SET_ORDER);
             foreach ($tags as $tag) {
-                $contentLength = self::_strlen($tag[3], $options);
+                $contentLength = 0;
+                if (!in_array($tag[2], static::$_defaultHtmlNoCount)) {
```
dereuromark: best to set third param to "true" here.

src/Utility/Text.php
LifeOrYou: I need this into 3.5.x, i must remove methods for this time ?

src/Utility/Text.php
LifeOrYou: thanks for the tip

src/Utility/Text.php
markstory: Yes. I don't think we need to make these tags configurable.

markstory:
Thanks!

PR #11574, opened 2017-12-24
Fix non interactive to only overwrite with designed force flag enabled.

dereuromark:
It seems so far the force flag wasn't used when working with file generation.
Using non-interactive mode in this case, though, can lead to unexpected data loss due to unintentional overwrite.
In my case I had git backup, but this can also go differently.

There is a force flag, we should use it:

   --force, -f            Force overwriting existing files without prompting.

Intuitively, when using non-interactive on shells, one would think it uses the default, which in this case would be "NO, do not overwrite existing file". So this fixes it to what is expected IMO.
I added tests to make sure all combinations of non-interactive and force work as expected.

Note: Also cleaned up two test class files into a standalone file.

### DIFF
--- a/src/Console/Shell.php
+++ b/src/Console/Shell.php
@@ -879,7 +879,14 @@ class Shell
 
         $this->_io->out();
 
-        if (is_file($path) && empty($this->params['force']) && $this->interactive) {
+        $fileExists = is_file($path);
+        if ($fileExists && empty($this->params['force']) && !$this->interactive) {
+            $this->_io->out('<warning>File exists, skipping</warning>.');
+
+            return false;
+        }
+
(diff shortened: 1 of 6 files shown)


dereuromark:
The failing CS seems unrelated and also invalid somehow..

bildschirmfoto vom 2017-12-25 01-41-59

markstory:
I've fixed that failing phpcs error on master in dff88e6c67ce3e34a9b3d8c0dd176120236be77b

dereuromark:
Thx, I rebased.

dereuromark:
@markstory Will we be able to release this as part of 3.5.10?

markstory:
@dereuromark Yes.

Issue #11577, opened 2017-12-27
Failed running test on PHPUnit 5.7


nojimage:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.5.9
* PHPUnit 5.7.26

### What you did

CakePHP 3.5.9 with PHPUnit 5.7, can't running test.

### What happened

I got

```
PHP Fatal error:  Declaration of Cake\TestSuite\Fixture\FixtureInjector::startTestSuite() must be compatible with PHPUnit_Framework_TestListener::startTestSuite(PHPUnit_Framework_TestSuite $suite) in /home/travis/build/cakephp/bake/vendor/cakephp/cakephp/src/TestSuite/Fixture/FixtureInjector.php on line 29
```

It seems that a problem has occurred due to this change:
https://github.com/cakephp/cakephp/commit/a17609ab16776ace2933e34a52132f0ec436857d#diff-63e46677a3b7a874996d8017691b866dR2

PHPUnit aliases is not created because `false` is specified as the second argument of `class_exists`.

I've created a patch, how about this fix?https://github.com/nojimage/cakephp/commit/28075d82ec61fa3da242cdf2e3c9e9b6984eeed3

nojimage:
refs https://travis-ci.org/cakephp/bake/jobs/322048601

ADmad:
Fixed in #11578

PR #11578, opened 2017-12-27
Load phpunit aliases in tests/bootstrap.php

nojimage:
fixes for #11577

### DIFF
--- a/tests/bootstrap.php
+++ b/tests/bootstrap.php
@@ -133,9 +133,4 @@ MutableDate::setTestNow(MutableDate::now());
 ini_set('intl.default_locale', 'en_US');
 ini_set('session.gc_divisor', '1');
 
-if (class_exists('PHPUnit_Runner_Version')) {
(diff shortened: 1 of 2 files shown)


ADmad (approved):

dereuromark:
Doesnt that break all 5.6 builds now in all plugins?
See https://travis-ci.org/cakephp/bake/builds/322319651 etc

saeideng:
@dereuromark 
related to loading  `tests/phpunit_aliases.php` by `loadPHPUnitAliases();`
because `tests/phpunit_aliases.php`  did not load before

dereuromark:
Also 7.1 tests now fail: https://travis-ci.org/dereuromark/cakephp-shim/jobs/322374951
What do you think we should do?

markstory:
@dereuromark This change was merged *after* the release of 3.5.9. I applied the wrong milestone to this change.

dereuromark:
Ah ok. You are saying that this will be fixed with the next release then?

markstory:
> Ah ok. You are saying that this will be fixed with the next release then?

I hope so.

Issue #11131, opened 2017-08-31
Cache key is too long when numerous cache groups are used


bancer:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 2.9.9.
* Platform and Target: Linux, Apache, MySQL, Memcached.

### What you did
We use cache groups. Some groups have many models therefore some values cannot be saved because the memcache key exceeds the maximum cache key length. We set the cache key in `AppModel::_readDataSource` to make it easier to debug, it gets concatenated to a very long cache key generated by the framework in `CacheEngine::key`.

An example of one cache group in core.php:
```
define('MEMCACHE_ORDERS', 'memcache_orders');
$cacheShopGroups = array('Shop', 'ShopCategory', 'ShopCategoryItem', 'ShopItem', 'ShopItemOverride');
$cacheOrderGroups = array('Order', 'ViewOrder', 'OrderCancellation', 'OrderItem', 'OrderNote', 'Payment', 'PaymentCredit');
Cache::config(MEMCACHE_ORDERS, array(
	'prefix' => $prefix,
	'servers' => $memcacheServers,
	'engine' => $memcacheType,
	'duration' => 86400,
	'groups' => array_merge($cacheOrderGroups, $cacheShopGroups),
));
```

In AppModel:
```
	protected function _readDataSource($type, $query) {
		if (isset($query['cache_config'])) {
			$cacheConfigName = $query['cache_config'];
		} else {
			$cacheConfigName = $this->_cacheConfigName;
		}
		// Cache only models that have non empty $_cacheConfigName.
		if (!empty($cacheConfigName) && $this->_isCacheGroupsEnabled) {
			$cacheName = md5(json_encode($query) . json_encode($this->hasOne) . json_encode($this->belongsTo));
			$backtrace = debug_backtrace();
			if (isset($backtrace[2]['function'])) {
				$cacheName .= '_' . $backtrace[2]['class'] . '::' . $backtrace[2]['function'];
			}
			$results = Cache::read($cacheName, $cacheConfigName);
			if ($results !== false) {
				// reset associatons to preserve the Model::_readDataSource() behavior
				$this->resetAssociations();
			} else {
				$results = parent::_readDataSource($type, $query);
				Cache::write($cacheName, $results, $cacheConfigName);
				// Temporary log the size of the cached results it is too high.
				$size = mb_strlen(serialize($results), '8bit');
[...]

PR #11581, opened 2017-12-28
Issue 11131 long cache key

bancer:
Solves https://github.com/cakephp/cakephp/issues/11131


lib/Cake/Cache/CacheEngine.php
```diff
@@ -181,7 +181,7 @@ public function key($key) {
 
 		$prefix = '';
 		if (!empty($this->_groupPrefix)) {
-			$prefix = vsprintf($this->_groupPrefix, $this->groups());
+			$prefix = md5(vsprintf($this->_groupPrefix, $this->groups()));
```
markstory: I'm not sure the `vsprintf` is necessary now. This could be `md5(implode('_', $this->groups()))` and get the same effect.

lib/Cake/Cache/CacheEngine.php
```diff
@@ -181,7 +181,7 @@ public function key($key) {
 
 		$prefix = '';
 		if (!empty($this->_groupPrefix)) {
-			$prefix = vsprintf($this->_groupPrefix, $this->groups());
+			$prefix = md5(implode('_', $this->groups()));
```
ADmad: You just dropped the usage of `_groupPrefix` property?

markstory: That property contained something like `%s_%s_%s` if the configuration had 3 groups. I don't think there is any actual difference with the current implementation other than there are fewer calls.

We should however, remove the population of this property as it is now unused.

bancer: `_groupPrefix` seems still to be used in FileEngine.

markstory: Darn. So close.

Issue #11582, opened 2017-12-28
Router not applying App.fullBaseUrl


PhantomWatson:
This is a bug.

* CakePHP Version: 3.5.9
* Platform and Target: **CLI script**

### What I did

    echo \Cake\Routing\Router::url([
        'plugin' => false,
        'prefix' => 'admin',
        'controller' => 'Deliveries',
        'action' => 'add',
        '_ssl' => true,
        '_full' => true
    ])

### What happened
`https:///admin/deliveries/add`

### What I expected to happen
`https://example.com/admin/deliveries/add`

### Notes
 - The above code works as expected in non-CLI environments, but not in CLI.
 - The above code works as expected if `_ssl` is unspecified, but not if `_ssl` is either `true` or `false`.

markstory:
I am assuming you have App.fullBaseUrl defined in your cli environment?

PhantomWatson:
Yep, it's defined, and it shows up in the URL when the `_ssl` option is left unspecified.

dereuromark:
PR now open.

PR #11583, opened 2017-12-29
Fix issue #11582

ericadestoya:
In `url()` method, looks for `isset($url['_ssl'])` first, so that `_ssl` changes are only made if `$url['_full']` is not met. Then, looks for the string `:///` and replaces it with `://`. Has been tested with all combinations of `$url['_ssl']`, `$url['_full']`, and whether or not `App.fullBaseUrl` has been set.

### DIFF
--- a/src/Routing/Router.php
+++ b/src/Routing/Router.php
@@ -617,6 +617,12 @@ class Router
             return $output;
         }
         if (is_array($url)) {
+            if (isset($url['_ssl'])) {
+                if (!isset($url['_full'])) {
+                    $url['_scheme'] = ($url['_ssl'] === true) ? 'https' : 'http';
+                }
+                unset($url['_ssl']);
+            }
             if (isset($url['_full']) && $url['_full'] === true) {
                 $full = true;
                 unset($url['_full']);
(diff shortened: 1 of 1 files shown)


dereuromark:
A small test case for this would be great. That will avoid future regressions here.

markstory:
Refs #11582

src/Routing/Router.php
```diff
@@ -617,6 +617,12 @@ public static function url($url = null, $full = false)
             return $output;
         }
         if (is_array($url)) {
+            if (isset($url['_ssl'])) {
+                if (!isset($url['_full'])) {
```
markstory: Is it possible for these conditions to be combined?  
```php
if (isset($url['_ssl']) && !isset($url['_full'])) {
  $url['_scheme'] = ($url['_ssl'] === true) ? 'https' : 'http';
}
unset($url['_ssl']);
```

dereuromark:
ping @ericadeefox

markstory:
I added tests and fixed up some behavior that didn't seem right in 82acd2beef98f2b4adcd46a8335a865d86d3a759. Previously using `_ssl=true, _full=true` you would get non-SSL URLs which didn't sound right to me.

PR #11585, opened 2017-12-30
Deprecate passing extra sort expressions by associative array

chinpei215:
Alternative to #11157.

### DIFF
--- a/src/Database/Expression/OrderByExpression.php
+++ b/src/Database/Expression/OrderByExpression.php
@@ -66,6 +66,14 @@ class OrderByExpression extends QueryExpression
      */
     protected function _addConditions(array $orders, array $types)
     {
+        foreach ($orders as $key => $val) {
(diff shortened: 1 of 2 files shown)


tests/TestCase/Database/SchemaCacheTest.php
```diff
@@ -51,6 +51,7 @@ public function setUp()
     {
         parent::setUp();
 
+        Cache::enable();
```
markstory: What is this needed for?

chinpei215: It is wired, but the test doesn't pass on my local environment as Cache is disabled if I run all tests. Probably some other test would disable Cache and not restore the settings. But I should fix the root cause. Let me try finding it.

markstory: ok

PR #11586, opened 2017-12-30
Timestamp behavior follow datetimetype set immutable

o0h:
TimestampBehavior hard construct `Cake\I18n\Time` instance. ([here](https://github.com/cakephp/cakephp/blob/f0e61d9b665249eef415e9749e8506e82e53fdc7/src/ORM/Behavior/TimestampBehavior.php#L143))
So, when new entity instance got by Table->save `created(or modified) property` is `Time` (new Time).
But with `useImmutable`,  the entity got by Table->get `created property` is `FrozenTime` (DateTimeType->toPHP()).

The same properties of same entity should be same type wherever they come from.

-------
### What you did
```php
$Table = TableRegistry::get('Users');
$entity = $Table->newEntity(['name' => 'my name']);
$saved = $Table->save($entity);
$read = $Table->get($entity->id);

echo get_class($saved->created); // Cake\I18n\Time
echo get_class($read->created); //  Cake\I18n\FrozenTime
```

```diff
-- echo get_class($saved); // Cake\I18n\Time
++ echo get_class($saved); // Cake\I18n\FrozenTime
echo get_class($read); //  Cake\I18n\FrozenTime
```

### DIFF
--- a/src/ORM/Behavior/TimestampBehavior.php
+++ b/src/ORM/Behavior/TimestampBehavior.php
@@ -14,11 +14,11 @@
  */
 namespace Cake\ORM\Behavior;
 
+use Cake\Database\Type;
 use Cake\Datasource\EntityInterface;
 use Cake\Event\Event;
-use Cake\I18n\Time;
 use Cake\ORM\Behavior;
-use DateTime;
+use DateTimeInterface;
 use UnexpectedValueException;
 
 class TimestampBehavior extends Behavior
@@ -57,7 +57,7 @@ class TimestampBehavior extends Behavior
     /**
      * Current timestamp
      *
-     * @var \DateTime
+     * @var \DateTimeInterface
      */
     protected $_ts;
 
@@ -130,19 +130,22 @@ class TimestampBehavior extends Behavior
      * If an explicit date time is passed, the config option `refreshTimestamp` is
      * automatically set to false.
      *
-     * @param \DateTime|null $ts Timestamp
+     * @param \DateTimeInterface|null $ts Timestamp
      * @param bool $refreshTimestamp If true timestamp is refreshed.
-     * @return \DateTime
+     * @return \DateTimeInterface
      */
-    public function timestamp(DateTime $ts = null, $refreshTimestamp = false)
+    public function timestamp(DateTimeInterface $ts = null, $refreshTimestamp = false)
     {
+        $class = Type::build('datetime')
+            ->getDateTimeClassName();
+
         if ($ts) {
             if ($this->_config['refreshTimestamp']) {
                 $this->_config['refreshTimestamp'] = false;
             }
-            $this->_ts = new Time($ts);
+            $this->_ts = new $class($ts);
         } elseif ($this->_ts === null || $refreshTimestamp) {
-            $this->_ts = new Time();
+            $this->_ts = new $class();
         }
 
         return $this->_ts;
--- a/src/ORM/Behavior/TimestampBehavior.php
+++ b/src/ORM/Behavior/TimestampBehavior.php
@@ -136,8 +136,9 @@ class TimestampBehavior extends Behavior
      */
     public function timestamp(DateTimeInterface $ts = null, $refreshTimestamp = false)
     {
-        $class = Type::build('datetime')
-            ->getDateTimeClassName();
+        /** @var \Cake\Database\Type\DateTimeType $type */
+        $type = Type::build('datetime');
+        $class = $type->getDateTimeClassName();
 
         if ($ts) {
             if ($this->_config['refreshTimestamp']) {
--- a/src/ORM/Behavior/TimestampBehavior.php
+++ b/src/ORM/Behavior/TimestampBehavior.php
@@ -14,6 +14,7 @@
  */
 namespace Cake\ORM\Behavior;
 
+use Cake\Database\Type;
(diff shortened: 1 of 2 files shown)


lorenzo:
Looks good to me

src/ORM/Behavior/TimestampBehavior.php
```diff
@@ -130,19 +130,23 @@ public function implementedEvents()
...
      * automatically set to false.
      *
-     * @param \DateTime|null $ts Timestamp
+     * @param \DateTimeInterface|null $ts Timestamp
      * @param bool $refreshTimestamp If true timestamp is refreshed.
-     * @return \DateTime
+     * @return \DateTimeInterface
      */
-    public function timestamp(DateTime $ts = null, $refreshTimestamp = false)
+    public function timestamp(DateTimeInterface $ts = null, $refreshTimestamp = false)
```
chinpei215: Strictly speaking, we cannot change signatures in minor releases. It may be better to change from master to 3.next.

src/ORM/Behavior/TimestampBehavior.php
```diff
@@ -130,19 +130,23 @@ public function implementedEvents()
...
     {
+        /** @var \Cake\Database\Type\DateTimeType $type */
+        $type = Type::build('datetime');
```
chinpei215: It is better than now, but if the `modified` field is declared as `timestamp`, and the `Type` settings are different between `'datetime'` and `'timestamp'`, I think it doesn't work as expected. 

See: https://github.com/cakephp/app/blob/94f120be6498f05654dfa83f6ea79302fa667661/config/bootstrap.php#L185-L188

dereuromark:
Given the typehint changes this sounds like 3.6 changes.
People could have extended the class.

src/ORM/Behavior/TimestampBehavior.php
o0h: it slipped my mind..

as default of [Type::$_types](https://github.com/cakephp/cakephp/blob/3.5.10/src/Database/Type.php#L34-L51), `datetime` and `timestamp` both are bound to `DateTimeType`.
(and DateTimeType::$_className value is shared as statically.)

i should consider there are difference  between `created` and `modified` column types ?

o0h:
thanks.
can i change this pr based-branch `master` to `3.next` ?

dereuromark:
I adjusted the PR base. But you might want to rebase on top of latest commit to resolve the conflicts.

dereuromark:
You can also create a patch from your local diff to master and re-apply that patch to a fresh 3.x branch and then force push it to this one here. That might be quickest.

o0h:
thanks. i pushed force my branch

src/ORM/Behavior/TimestampBehavior.php
chinpei215: Hmm. It seems impossible to detect correct types in this method, as it doesn't know what field uses the timestamp. I think it can be done in the `_updateField()` method instead, by using `$this->_table->getSchema()`.

markstory: We won't be able to get the class correct 100% of the time, but it is more *correct* now than it was in the past.

lorenzo: It is true that should plan for a change that actually uses the right type

markstory: How would we get the correct types based on the arguments to this method?

o0h: can i add `field` option to `timestamp()` and `_updateField()` passes field name ?

--------

```php
    /**
     * @param string  $field Field name to set timestamp
     */
public function timestamp(DateTimeInterface $ts = null, $refreshTimestamp = false, $field = 'created')
    {
        $type = Type::build('datetime');
        $class = $type->getDateTimeClassName();
```
and
```php
    protected function _updateField($entity, $field, $refreshTimestamp)
    {
        if ($entity->isDirty($field)) {
            return;
        }
        $entity->set($field, $this->timestamp(null, $refreshTimestamp, $field));
    }
}
```

o0h: if we should use the right type, maybe it would be better to to keep `$_ts` each fields...
because when `$ts` is null && !$refreshTimestamp, `timestamp()` returns cached instance.

chinpei215:
You can ignore EventManagerTest errors, as I have already fixed it in my pull request.

src/ORM/Behavior/TimestampBehavior.php
chinpei215: I think you could revert timestamp() and change `_updateField()` method like the following: 
```
protected function _updateField($entity, $field, $refreshTimestamp)
{
    if ($entity->isDirty($field)) {
        return;
    }

    $ts = $this->timestamp(null, $refreshTimestamp);

    $columnType = $this->getTable()->getScheme()->getColumnType($field);
    $type = Type::build($columnType);
    $class = $type->getDateTimeClassName();

    // Create a new proper DateTimeInterface instance here

    $entity->set($field, $ts);
}
```

o0h: thanks, i'll  do it!

o0h:
implemented in _updateField(), and added an test.

PR #11589, opened 2017-12-31
Allow `false` prefix to be used in options().

### LINKED ISSUE #11515: multiple paginator problem
This line 
https://github.com/cakephp/cakephp/commit/ed233e7167f2b693c90fa30fbba0d1ba58d3ec6d#diff-9331fd78b8c7f46ba39642f792749d81L471
should be changed to support params which we would to set it to false, like 'prefix' => false

markstory:
Allow prefixes to be disabled via options('url').

Fixes #11515


dereuromark:
Looks good.

PR #11592, opened 2017-12-31
[2.x]Fix can't mock aliased component on ControllerTestCase

tenkoma:
* CakePHP Version: 2.10.6
* Platform and Target: PHP 5.6.31

In the `ControllerTestCase::generate()` method, cannot mock components defined by aliases as follows.

```php
<?php
class PostsController extends AppController {
    public $components = array(
        'Auth' => array(
            'className' => 'CustomizedAuth'
        )
    );

    // other code
}
```

In this case, the `AuthComponent` is mocked, not the `CustomizedAuthComponent`.

To fix the problem, I imported the code to load the component from [`ComponentCollection::load()`](https://github.com/cakephp/cakephp/blob/2.x/lib/Cake/Controller/ComponentCollection.php#L97-L104)

### DIFF
--- a/lib/Cake/Test/Case/TestSuite/ControllerTestCaseTest.php
+++ b/lib/Cake/Test/Case/TestSuite/ControllerTestCaseTest.php
@@ -281,8 +281,7 @@ class ControllerTestCaseTest extends CakeTestCase {
  *
  * @return void
  */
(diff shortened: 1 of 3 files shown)


markstory:
Thanks!

Issue #11595, opened 2018-01-02
Make _name available from Route.php


EliasDeVos:
This is a (multiple allowed):

* [ ] bug
* [x] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.x

Hi,
I needed a way to get the _name option from Route.php in my controller.
I didn't really find a good way to do it so i added this to the parse() method.

I tried to create a pull request but didn't have access.
This is the code i would add (in src/Routing/Route/Route.php:482:
`// pass the name if set`
 ` if (isset($this->options['_name'])) {`
      `$route['_name'] = $this->options['_name'];`
  `}`

If there is another way to do this or you dont think this code belongs here let me know.

saeideng:
duplicate : https://github.com/cakephp/cakephp/issues/10950

Issue #10950, opened 2017-07-26
add `_name` to request object  (for using  it on Paginator , ... )


saeideng:
This is a (multiple allowed):

* [ ] bug
* [x] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.5.0

### What you did
detect action by `_name` of routes on across app specially on view 

### How
add `_name` to request object  like `_matchedRoute`

markstory:
How do you see `_name` being used?

lorenzo:
It could be displayed in debugkit

markstory:
Perhaps we should find a way to get more information about the currently matched route into debugkit?

saeideng:
> It could be displayed in debugkit

debugkit use directly `routes` and `foreach()` for showing

saeideng:
> Perhaps we should find a way to get more information about the currently matched route into debugkit?

what debugkit?
we need use this in our programs :)

markstory:
@saeideng What do you use it for?

saeideng:
> How do you see _name being used?

for example :
for detect special actions for add/show/modify view/layout

saeideng:
```
		$this->request->params['lang_params']=
		[
			'_name'=>'ads_show',
		];
```
now i use this form for sending data

markstory:
How does that connect with the Route names though?

saeideng:
`$this->request->getParam('_named');`

but I think we now already below form for 3.5??
`$this->request->route()->getName();`

markstory:
Requests don't have access to the matched route. If we're interested in having more route state available on the request, then we could use an attribute, but I'd like to better understand the use cases for that data.

saeideng:
1)i use this info for display `lang` select but just for especial action
2)now we able to detect all data (passed/controller/action) except `_name` 
3)for detect especial action we need now check both `controller` and `action` but `_name` is fast 
4)with `$this->request->route()` we can deprecate all of params (`controller` ,`action`,passed)
5).... as soon :smile:

markstory:
> with $this->request->route() we can deprecate all of params (controller ,action,passed)

I don't think we can. Route instances don't have complete information on what was in the URL.

saeideng:
@markstory 
if you define a route by `_name`
and you want use Paginator for that action you can not use it without defination `$this->Paginator->options(['url'=>[`
like 
```

markstory:
Named routes can still be matched through reverse routing though.

[...]

PR #11596, opened 2018-01-02
show _name parameter in request if set

EliasDeVos:
Hi,
I made this pull request for these issues: #11595  and #10950,

With this addition the _name parameter is also visible in the request object.

In my use case we use the _name to identify a route, and if we have seo stuff in the db we can easily retrieve it by using _name.

For me having the full Route object attached to the request isn't necessary, but if you think this would be a better solution let me know and i will try to make a pull request for it

### DIFF
--- a/src/Routing/Route/Route.php
+++ b/src/Routing/Route/Route.php
@@ -479,6 +479,11 @@ class Route
             $route['_ext'] = $ext;
         }
 
+        // pass the name if set
+        if (isset($this->options['_name'])) {
+            $route['_name'] = $this->options['_name'];
+        }
+
         // restructure 'pass' key route params
         if (isset($this->options['pass'])) {
             $j = count($this->options['pass']);
--- a/src/Routing/Router.php
+++ b/src/Routing/Router.php
@@ -746,7 +746,8 @@ class Router
             $params['requested'],
             $params['return'],
             $params['_Token'],
-            $params['_matchedRoute']
+            $params['_matchedRoute'],
+            $params['_name']
         );
         $params = array_merge($params, $pass);
         if (!empty($url)) {
--- a/tests/TestCase/Routing/RouteCollectionTest.php
+++ b/tests/TestCase/Routing/RouteCollectionTest.php
@@ -361,7 +361,7 @@ class RouteCollectionTest extends TestCase
         $routes = new RouteBuilder($this->collection, '/b', ['key' => 'value']);
         $routes->connect('/', ['controller' => 'Articles']);
         $routes->connect('/:id', ['controller' => 'Articles', 'action' => 'view']);
-        $routes->connect('/media/search/*', ['controller' => 'Media', 'action' => 'search'], ['_name' => 'media_search']);
+        $routes->connect('/media/search/*', ['controller' => 'Media', 'action' => 'search']);
 
         $request = new ServerRequest(['url' => '/b/']);
         $result = $this->collection->parseRequest($request);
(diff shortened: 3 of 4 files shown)


src/Routing/Route/Route.php
```diff
@@ -479,6 +479,11 @@ public function parse($url, $method = '')
             $route['_ext'] = $ext;
         }
 
+        // pass the name if set
+        if (isset($this->options['_name'])) {
+            $route['_name'] = $this->options['_name'];
+        }
```
markstory: Any chance tests could be updated to account for this additional key?

markstory (changes requested):

src/Routing/Route/Route.php
EliasDeVos: Hey, ofcourse
I will add some tests tomorrow

saeideng:
but I prefer to having route object in request :)
object is safe 
and we can use setter/getter for route object  property 
```php
$this->request->route()->getNamed();
$this->request->route()->setNamed(...);
```

EliasDeVos:
@saeideng and where would you set in initially ?
I was thinking in maybe here ? https://github.com/cakephp/cakephp/blob/master/src/Routing/Middleware/RoutingMiddleware.php#L88
because the parseRequest knows the route, but maybe it is not expected behavior and i don't know where else this method gets used

saeideng:
looks we can not add object until cake 4.0 IMO
so , `$route['_name']` is good now

markstory:
The route object could be added as an attribute using `withAttribute()` but that will likely require some breaking changes unfortunately? Having the name is something we can support now.

What use cases did you have in mind that require the route object @saeideng?

src/Routing/Route/Route.php
markstory: I think this key might need to be removed by `Router::reverse()` in this section https://github.com/cakephp/cakephp/blob/e3c3594e24bfc83bfbcd772b19789cdd0557f6c1/src/Routing/Router.php#L738-L754

Or the `_name` parameter will impact the results.

tests/TestCase/Routing/RouteCollectionTest.php
```diff
@@ -78,7 +78,7 @@ public function testParse()
         $routes = new RouteBuilder($this->collection, '/b', ['key' => 'value']);
         $routes->connect('/', ['controller' => 'Articles']);
         $routes->connect('/:id', ['controller' => 'Articles', 'action' => 'view']);
-        $routes->connect('/media/search/*', ['controller' => 'Media', 'action' => 'search']);
+        $routes->connect('/media/search/*', ['controller' => 'Media', 'action' => 'search'], ['_name' => 'media_search']);
```
saeideng: generally you should add new test instead of modifying existing testCases

src/Routing/Router.php
```diff
@@ -746,7 +746,8 @@ public static function reverseToArray($params)
             $params['requested'],
             $params['return'],
             $params['_Token'],
-            $params['_matchedRoute']
+            $params['_matchedRoute'],
+            $params['_name']
```
saeideng: @markstory  I think output of this method should contain  `_name` like before

saeideng:
I added a fix for test error

src/Routing/Router.php
markstory: Do you think it is useful for reverse routing?

saeideng: I never used this method , so I really don't know :)
I just say , this can be break BC, but I'm not sure
if you say :+1:  i say too :+1:

saeideng: I tested in local 
before and after these  changes `reverseToArray` returns same result

saeideng:
@EliasDeVos thanks

PR #11598, opened 2018-01-03
Added irregular word to inflector: receive

jeremyharris:

### DIFF
--- a/src/Utility/Inflector.php
+++ b/src/Utility/Inflector.php
@@ -144,7 +144,8 @@ class Inflector
         'goose' => 'geese',
         'foot' => 'feet',
         'foe' => 'foes',
-        'sieve' => 'sieves'
+        'sieve' => 'sieves',
+        'receive' => 'receives',
     ];
 
     /**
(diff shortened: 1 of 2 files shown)


src/Utility/Inflector.php
```diff
@@ -144,7 +144,8 @@ class Inflector
         'goose' => 'geese',
         'foot' => 'feet',
         'foe' => 'foes',
-        'sieve' => 'sieves'
+        'sieve' => 'sieves',
+        'receive' => 'receives',
```
ADmad: Hmm.. "recieve" isn't a noun and "receives" is not a plural of "recieve".

jeremyharris: You're right and it totally slipped my mind when adding this. I just saw "InventoryReceife" in a freelance project and opened this.

Seems it should be a custom rule for this project. Sorry about that, I'll be more thoughtful next time. Should I create a revert pr?

dereuromark: Yes please :)

PR #11599, opened 2018-01-04
Add more field value comparison validation rules

robertpustulka:
Added a number of validation rules that compare one field value to another field value.
Useful in date range checks etc.

- `equalToField` - weak type equality comparison
- `notEqualToField` - weak type non-equality comparison
- `greaterThanField`
- `greaterThanOrEqualToField`
- `lessThanField`
- `lessThanOrEqualToField`

Behind the scenes `Validation::compareFields()` now leverages `comparison` method for value checks.
I also added strict type comparisons to `Validation::comparison()`.

### DIFF
--- a/src/Validation/Validation.php
+++ b/src/Validation/Validation.php
@@ -300,48 +300,66 @@ class Validation
         $operator = str_replace([' ', "\t", "\n", "\r", "\0", "\x0B"], '', strtolower($operator));
         switch ($operator) {
             case 'isgreater':
+                /**
+                 * @deprecated 3.6.0 Use Validation::COMPARE_GREATER instead.
+                 */
                 deprecationWarning(sprintf($message, $operator, 'COMPARE_GREATER'));
-                // @deprecated 3.6.0 Use Validation::COMPARE_GREATER instead.
+                // no break
             case static::COMPARE_GREATER:
                 if ($check1 > $check2) {
                     return true;
                 }
                 break;
             case 'isless':
+                /**
+                 * @deprecated 3.6.0 Use Validation::COMPARE_LESS instead.
+                 */
                 deprecationWarning(sprintf($message, $operator, 'COMPARE_LESS'));
-                // @deprecated 3.6.0 Use Validation::COMPARE_LESS instead.
+                // no break
             case static::COMPARE_LESS:
                 if ($check1 < $check2) {
                     return true;
                 }
                 break;
             case 'greaterorequal':
+                /**
+                 * @deprecated 3.6.0 Use Validation::COMPARE_GREATER_OR_EQUAL instead.
+                 */
                 deprecationWarning(sprintf($message, $operator, 'COMPARE_GREATER_OR_EQUAL'));
-                // @deprecated 3.6.0 Use Validation::COMPARE_GREATER_OR_EQUAL instead.
+                // no break
             case static::COMPARE_GREATER_OR_EQUAL:
                 if ($check1 >= $check2) {
                     return true;
                 }
                 break;
             case 'lessorequal':
+                /**
+                 * @deprecated 3.6.0 Use Validation::COMPARE_LESS_OR_EQUAL instead.
+                 */
                 deprecationWarning(sprintf($message, $operator, 'COMPARE_LESS_OR_EQUAL'));
-                // @deprecated 3.6.0 Use Validation::COMPARE_LESS_OR_EQUAL instead.
+                // no break
             case static::COMPARE_LESS_OR_EQUAL:
                 if ($check1 <= $check2) {
                     return true;
                 }
                 break;
             case 'equalto':
+                /**
+                 * @deprecated 3.6.0 Use Validation::COMPARE_EQUAL instead.
+                 */
                 deprecationWarning(sprintf($message, $operator, 'COMPARE_EQUAL'));
-                // @deprecated 3.6.0 Use Validation::COMPARE_EQUAL instead.
+                // no break
             case static::COMPARE_EQUAL:
                 if ($check1 == $check2) {
                     return true;
                 }
                 break;
             case 'notequal':
+                /**
+                 * @deprecated 3.6.0 Use Validation::COMPARE_NOT_EQUAL instead.
+                 */
                 deprecationWarning(sprintf($message, $operator, 'COMPARE_NOT_EQUAL'));
-                // @deprecated 3.6.0 Use Validation::COMPARE_NOT_EQUAL instead.
+                // no break
             case static::COMPARE_NOT_EQUAL:
                 if ($check1 != $check2) {
                     return true;
--- a/src/Validation/Validation.php
+++ b/src/Validation/Validation.php
@@ -295,51 +295,57 @@ class Validation
             return false;
         }
 
+        $message = 'Operator `%s` is deprecated, use constant `Validation::%s` instead.';
+
         $operator = str_replace([' ', "\t", "\n", "\r", "\0", "\x0B"], '', strtolower($operator));
         switch ($operator) {
             case 'isgreater':
+                deprecationWarning(sprintf($message, $operator, 'COMPARE_GREATER'));
             case static::COMPARE_GREATER:
                 if ($check1 > $check2) {
                     return true;
                 }
                 break;
             case 'isless':
+                deprecationWarning(sprintf($message, $operator, 'COMPARE_LESS'));
             case static::COMPARE_LESS:
                 if ($check1 < $check2) {
                     return true;
                 }
                 break;
             case 'greaterorequal':
+                deprecationWarning(sprintf($message, $operator, 'COMPARE_GREATER_OR_EQUAL'));
             case static::COMPARE_GREATER_OR_EQUAL:
                 if ($check1 >= $check2) {
                     return true;
                 }
                 break;
             case 'lessorequal':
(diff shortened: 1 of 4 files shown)


src/Validation/Validation.php
```diff
@@ -314,7 +326,7 @@ public static function comparison($check1, $operator, $check2)
      */
     public static function compareWith($check, $field, $context)
     {
-        return self::compareFields($check, $field, true, $context);
+        return self::compareFields($check, $field, '===', $context);
```
dereuromark: Can we use class constant instead of magic strings? :)

src/Validation/Validator.php
```diff
@@ -1004,7 +1004,133 @@ public function notSameAs($field, $secondField, $message = null, $when = null)
...
         return $this->add($field, 'notSameAs', $extra + [
-            'rule' => ['compareFields', $secondField, false]
+            'rule' => ['compareFields', $secondField, '!==']
+        ]);
+    }
+
+    /**
+     * Add a rule to compare one field is equal to another.
+     *
+     * @param mixed $field The field you want to apply the rule to.
```
dereuromark: Please avoid mixed

src/Validation/Validation.php
```diff
@@ -324,20 +336,18 @@ public static function compareWith($check, $field, $context)
      *
      * @param mixed $check The value to find in $field.
      * @param string $field The field to check $check against. This field must be present in $context.
-     * @param bool $result The expected result of field comparison.
+     * @param string $operator Comparison operator.
```
dereuromark: I think we need to keep bool for BC instead 3.x? `string|bool ... Deprecated: bool is deprecated.` etc.
Also please add a test for the old bool to still work as expected until we can safely remove in 4.0 then.

robertpustulka: This method has been added in 3.6. Since 3.6 hasn't been released yet, we shouldn't worry about BC, am I right?

src/Validation/Validation.php
dereuromark:
```
Validation::COMPARE_EQUAL_TO
Validation::COMPARE_NOT_EQUAL_TO
Validation::COMPARE_LESS_THAN
Validation::COMPARE_LESS_THAN_OR_EQUAL
Validation::COMPARE_GREATER_THAN
Validation::COMPARE_GREATER_THAN_OR_EQUAL
```
would be nice to work with.

src/Validation/Validation.php
```diff
@@ -1216,10 +1282,10 @@ public static function uploadedFile($file, array $options = [])
         if ($options['optional'] && $error === UPLOAD_ERR_NO_FILE) {
             return true;
         }
-        if (isset($options['minSize']) && !static::fileSize($file, '>=', $options['minSize'])) {
+        if (isset($options['minSize']) && !static::fileSize($file, static::COMPARE_GREATER_THAN_OR_EQUAL_TO, $options['minSize'])) {
```
dereuromark: What do you think about keeping them only as long as possible to be descriptive enough?
E.g. `COMPARE_GREATER_OR_EQUAL` would suffice IMO, same for others. The rest is gramm. sugar which shouldnt be needed for constants, otherwise they are long and make the rest of the code less readable.

src/Validation/Validation.php
```diff
@@ -38,6 +38,62 @@ class Validation
...
+     * Equal to comparison operator.
+     *
+     * @var string
+     */
+    const COMPARE_EQUAL_TO = '==';
+
+    /**
+     * Not equal to comparison operator.
+     *
+     * @var string
```
dereuromark: Those var things are not needed for constants. They are at runtime not changeable, so the type is always clear from the code itself right away.

robertpustulka: They appear elsewhere.

dereuromark: You mean this is done elsewhere? Yeah, but it is really not needed and should be sniffer removed IMO. Duplicating these things for no reason also makes it out of sync error prone at some point and adds additional maintenance overhead.

src/Validation/Validation.php
```diff
@@ -260,41 +316,53 @@ public static function comparison($check1, $operator, $check2)
         $operator = str_replace([' ', "\t", "\n", "\r", "\0", "\x0B"], '', strtolower($operator));
         switch ($operator) {
             case 'isgreater':
```
dereuromark: Do we still need these strange magic strings then? I would remove those in favor of speaking const names now :)

src/Validation/Validation.php
robertpustulka: > You mean this is done elsewhere?
Yes.

https://github.com/robertpustulka/cakephp/blob/master/src/Validation/Validator.php#L36

src/Validation/Validation.php
robertpustulka: Those should probably stay for BC.

dereuromark: `equalto` vs `notequal` are not even following the same naming pattern.. :)
I would expect the first to be `equal` only.

src/Validation/Validation.php
```diff
@@ -38,6 +38,62 @@ class Validation
      */
     const DEFAULT_LOCALE = 'en_US';
 
+    /**
+     * Same as operator.
+     *
+     * @var string
+     */
+    const COMPARE_SAME_AS = '===';
```
dereuromark: to be consistent I would ask for

- COMPARE_SAME
- COMPARE_NOT_SAME
- COMPARE_EQUAL
- COMPARE_NOT_EQUAL
- COMPARE_GREATER
- COMPARE_GREATER_OR_EQUAL
- COMPARE_LESS
- COMPARE_LESS_OR_EQUAL

Also would be most concise while still being descriptive.

src/Validation/Validation.php
dereuromark: Yeah, should all be removed in the future, and no new ones added IMO.

src/Validation/Validation.php
robertpustulka: It's probably out of the scope of this PR for now :)

tests/TestCase/Validation/ValidatorTest.php
```diff
@@ -1425,8 +1425,9 @@ public function testNotEquals()
     public function testSameAs()
     {
         $validator = new Validator();
-        $this->assertProxyMethod($validator, 'sameAs', 'other', ['other', true], 'compareFields');
+        $this->assertProxyMethod($validator, 'sameAs', 'other', ['other', '==='], 'compareFields');
```
dereuromark: You chould use the const also for '===' etc.

dereuromark (approved):

tests/TestCase/Validation/ValidatorTest.php
robertpustulka: There is one `Validation::SAME_AS` but not used in tests.

dereuromark: Yeah, I mean you could reuse the const existing in the Validation class here to be more speaking and to avoid typos.

src/Validation/Validator.php
```diff
@@ -1004,7 +1004,133 @@ public function notSameAs($field, $secondField, $message = null, $when = null)
...
+     * Add a rule to compare one field is equal to another.
+     *
+     * @param string $field The field you want to apply the rule to.
+     * @param string $secondField The field you want to compare against.
+     * @param string|null $message The error message when the rule fails.
+     * @param string|callable|null $when Either 'create' or 'update' or a callable that returns
+     *   true when the validation rule should be applied.
+     * @see \Cake\Validation\Validation::compareFields()
+     * @return $this
+     * @since 3.6.0
```
ADmad: Thank you for adding `@since` tags.

src/Validation/Validation.php
markstory: We could emit deprecation warnings when the string words are used. That would enable us to remove them in 4.0.0.

robertpustulka: @markstory deprecation warnings added

src/Validation/Validation.php
```diff
@@ -301,41 +301,47 @@ public static function comparison($check1, $operator, $check2)
         switch ($operator) {
             case 'isgreater':
                 deprecationWarning(sprintf($message, $operator, 'COMPARE_GREATER'));
+                // @deprecated 3.6.0 Use Validation::COMPARE_GREATER instead.
```
dereuromark: I would have switched the order of comment and deprecationWarning to call to not confuse this comment and the code below.

    // deprecated
    deprecationWarning()

robertpustulka: @dereuromark I think that fixer requires that the comment is the last line in a block

I get `There must be a comment when fall-through is intentional in a non-empty case body` when I change the order.

dereuromark: Ah I see, ok.

src/Validation/Validation.php
```diff
@@ -300,48 +300,66 @@ public static function comparison($check1, $operator, $check2)
...
                 deprecationWarning(sprintf($message, $operator, 'COMPARE_LESS'));
-                // @deprecated 3.6.0 Use Validation::COMPARE_LESS instead.
+                // no break
             case static::COMPARE_LESS:
                 if ($check1 < $check2) {
                     return true;
                 }
                 break;
             case 'greaterorequal':
+                /**
```
dereuromark: Note: /** means usually a doc block

PR #11605, opened 2018-01-05
Remove deprecated function calls

antograssiot:
Just remove some function calls that have been deprecated recently from the core

### DIFF
--- a/src/Console/CommandRunner.php
+++ b/src/Console/CommandRunner.php
@@ -22,7 +22,7 @@ use Cake\Console\ConsoleIo;
 use Cake\Console\Exception\StopException;
 use Cake\Console\Shell;
 use Cake\Core\ConsoleApplicationInterface;
-use Cake\Event\EventManagerTrait;
+use Cake\Event\EventDispatcherTrait;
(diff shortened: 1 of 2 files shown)


src/TestSuite/Stub/Response.php
```diff
@@ -29,7 +29,7 @@ class Response extends Base
     public function send()
     {
         if ($this->hasHeader('Location') && $this->_status === 200) {
-            $this->statusCode(302);
+            $this->withStatus(302);
```
dereuromark: Wouldnt you need to assign the immutable response?

markstory (commented):
These changes might have already been applied to 3.next.

antograssiot:
Oh sorry @markstory, I'll rebase onto 3.next to see if  some changes are still needed and then update the target branch of this PR or just close it

src/Console/CommandRunner.php
```diff
@@ -31,7 +31,7 @@
  */
 class CommandRunner
```
robertpustulka: Could you add a `EventDiapatcherInterface` here?

PR #11607, opened 2018-01-06
Fix incorrect tag for EntityInterface::getSource()

o0h:
Fix mismatch between EntityInterface and implements in EntityTrait.

With [phan](https://github.com/phan/phan), static analyzer reports error.

```php
Inflector::singularize($entity->getSource());
```
then
```
PhanTypeMismatchArgument Argument 1 (word) is array but \Cake\Utility\Inflector::singularize() takes string defined at vendor/cakephp/cakephp/src/Utility/Inflector.php:546
```


markstory:
Thanks!

PR #11612, opened 2018-01-08
Fix cache as dependency for ORM

dereuromark:
Resolves https://github.com/sergeyklay/php-orm-benchmark/pull/6
By default cache must be enabled on production systems, otherwise it will completely kill the ORM performance.
Thus this should be a require dependency.
We should use good production defaults here - and the exception should really be avoided:

    Error: Class 'Cake\Cache\Cache' not found ...
  
  A few files more don't hurt, but the benefit of a good default is crucial, especially for those first time working with this ORM and might otherwise miss it somewhere hidden in docs.

### DIFF
--- a/src/Database/composer.json
+++ b/src/Database/composer.json
@@ -25,12 +25,12 @@
     },
     "require": {
         "php": ">=5.6.0",
+        "cakephp/cache": "^3.0.0",
         "cakephp/core": "^3.0.0",
         "cakephp/datasource": "^3.0.0"
     },
     "suggest": {
-        "cakephp/log": "Require this if you want to use the built-in query logger",
-        "cakephp/cache": "Require this if you decide to use the schema caching feature"
+        "cakephp/log": "Require this if you want to use the built-in query logger"
     },
     "autoload": {
         "psr-4": {
--- a/src/Datasource/composer.json
+++ b/src/Datasource/composer.json
@@ -25,6 +25,7 @@
     },
     "require": {
         "php": ">=5.6.0",
+        "cakephp/cache": "^3.0.0",
         "cakephp/core": "^3.0.0"
     },
     "suggest": {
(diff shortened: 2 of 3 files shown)


src/ORM/composer.json
```diff
@@ -24,6 +24,7 @@
     },
     "require": {
         "php": ">=5.6.0",
+        "cakephp/cache": "^3.0.0",
```
saeideng: I guess you need this in the 
`cakephp/datasource`
`cakephp/database`

saeideng (changes requested):

dereuromark:
@saeideng Thank you, good find!

lorenzo:
This package was left out on purpose and only set as a suggestion, what changed?

lorenzo:
I guess it is not super important, but ideally in the future we should conform to some sort of interface and let the user chose the caching library

dereuromark:
Yes, ideally we should interface contract here :)
We can still do that probably in a BC way, I think.

PS: We should also have a tool for finding the dependencies to other use statements. 
Right now it is a lot of guessing from our side, since we do not use those splits, and there is also no tooling around it for split unit testing.

lorenzo:
Can you change the readme and add some notes about configuring he cache? I think by default the cache folder will not exist, so it would make sense to show how to configure it to use a sys tmp folder

dereuromark:
What would be needed using the split setup?
If you have an idea, you can add your commit on top.

lorenzo:
Can you use the code from the php orm benchmarks repo?

dereuromark:
I wonder if we should add the note that by default even with `'cacheMetadata' => true` enabled it will use a NullLogger and not cache at all.

dereuromark:
Does it look complete now?

Issue #9188, opened 2016-07-29
About cakephp/routing


slince:
Hi guys,

I have created a route like this:
```
    $routes->connect('/thumbnail/:widthx:height/:image', ['controller' => 'Image', 'action' => 'thumb'], [
        'pass' => ['image', 'width', 'height'],
        'width' => '\d+',
        'height' => '\d+',
    ]);
```
and I want it match `/thumbnail/200x300/foo-bar.png`,however it doesn't work.
I think I need some help.

dereuromark:
This is not a help forum. The ticket tracker is reserved for possible
bugs and feature enhancements to the CakePHP framework. If you are
looking for help on how to implement a feature or to better understand
how to use the framework correctly, please visit one of the following:

[The CakePHP Manual](http://book.cakephp.org)
[The CakePHP online API](http://api.cakephp.org)
[CakePHP Official Forum](http://discourse.cakephp.org)
[Stackoverflow](http://stackoverflow.com/questions/tagged/cakephp)
[Slack channel](http://cakesf.herokuapp.com/)

or the #cakephp channel on irc.freenode.net, where we will be more than
happy to help answer your questions.

Thanks!

slince:
I have reviewed documents, unfortunately there's  no good solution because  ‘`x`’ is a valid english character; I recommend using "`{placeholder}`"  replacement "`:placeholder`"

slince:
@dereuromark

htstudios:
Optional {} to clearly mark a varianle and \ for escaping sounds good. Would you setup a PR?

slince:
I’m so sorry, I'm not familiar with the `cakephp/routing` source code @htstudios

PR #11613, opened 2018-01-09
3.next - Allow route placeholders to use {var}

markstory:
Brace placeholders allow route placeholders to be more easily identified and be placed mid-word with other alphanumeric characters.

While the original issues proposed that this new `{var}` style be the only supported style in the future, I'd like to push that decision off until 5.0 at the earliest. Removing `:var` style placeholders would be a significant impediment to upgrading to 4.0 and I'm not sure that is a good idea given that supporting both styles now is simple. If the `{var}` style proves successful we can deprecate `:var` in 4.x.

Refs #9188
Refs #10689

### DIFF
--- a/tests/TestCase/Routing/Route/RouteTest.php
+++ b/tests/TestCase/Routing/Route/RouteTest.php
@@ -191,6 +191,19 @@ class RouteTest extends TestCase
     public function testRouteCompileMixedPlaceholders()
     {
         $route = new Route(
+            '/images/{open/:id',
+            ['controller' => 'Images', 'action' => 'open']
+        );
+        $pattern = $route->compile();
+        $this->assertRegExp($pattern, '/images/{open/9', 'Need both {} to enable brace mode');
+        $result = $route->match([
+            'controller' => 'Images',
+            'action' => 'open',
+            'id' => 123,
+        ]);
+        $this->assertEquals('/images/{open/123', $result);
+
+        $route = new Route(
             '/fighters/{id}/move/{x}/:y',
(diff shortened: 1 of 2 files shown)


tests/TestCase/Routing/Route/RouteTest.php
```diff
@@ -142,6 +142,85 @@ public function testRouteBuildingSmallPlaceholders()
...
+            'controller' => 'Fighters',
+            'action' => 'move',
+            'id' => 123,
+            'x' => 8,
+            'y' => 42
+        ]);
+        $this->assertEquals('/fighters/123/move/8/42', $result);
+
+        $route = new Route(
+            '/images/{id}/{x}x{y}',
```
saeideng: hi
can you add a test for :point_down: ?
```php
{ id}
{id }
```

ADmad: Why? `{ id}` shouldn't work. :)

markstory: I can add a negative test showing that it *doesn't* work. It might also be wise to have tests for `A-Z` and `_`.

markstory: Done!

saeideng: why `i_d` and `id99` is valid ?
this can make confusing with 
```php
//id=123
'/fighters/{id99}' // returns '/fighters/12399' 
```
now 
`99` is valid character but space no

markstory:
Any other feedback on this?

Issue #11186, opened 2017-09-14
Quotes with DISTINCT inside COUNT


kicaj:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: Tested on 2.9.6 and latest 2.10.2
* Platform and Target: WAM(MSSQL)P

### What you did
Using `COUNT` with `DISTINCT` on built-in SQL Server driver.

### What happened
`SELECT COUNT(DISTINCT User.id) AS [count] FROM [users] AS [User]`

### What you expected to happen
Should add quotes/bracklets to `DISTINCT` field, like this:
`SELECT COUNT(DISTINCT [User].[id]) AS [count] FROM [auth_users] AS [User]`
but it does not.

Solutions is to rebuild lines 301-304: https://api.cakephp.org/2.10/source-class-Sqlserver.html#301-304
to:
```
	if (strpos($fields[$i], 'DISTINCT') !== false) {
		$prepend = 'DISTINCT ';
		$fields[$i] = trim(str_replace('DISTINCT', '', $fields[$i]));

		if (strpos($fields[$i], 'COUNT(') !== false) {
			$prepend = 'COUNT(' . $prepend;
			
			$fields[$i] = trim(str_replace('COUNT(', '', $this->_quoteFields($fields[$i])));
		}
	}
```
It will be works with usinq bracklets for field inside `COUNT(DISTINCT alias.field)`.

markstory:
Any chance you could make a pull request for this?

markstory:
Pull request open now.

PR #11616, opened 2018-01-09
Fix missing field identifier quoting for COUNT(DISTINCT in SQLServer.

markstory:
Refs #11186


dereuromark:
@kicaj Can you confirm maybe?

kicaj:
@dereuromark Yes, that's it.

Thanks @markstory!

Issue #11346, opened 2017-10-20
2.x support PHP 7.2?


josephzidell:
What is the official word on 2.x with PHP 7.2?

- [`.travis.yml`](https://github.com/cakephp/cakephp/blob/2.x/.travis.yml) does not list it
- [`composer.json`](https://github.com/cakephp/cakephp/blob/2.x/composer.json) seems carte blanche >= 5.3.0
- The [installation page of the cookbook](https://book.cakephp.org/2.0/en/installation.html) seems that you can, as long as you install `mcrypt` through PECL, although does not mention 7.2 explicitly.
- The [security page](https://book.cakephp.org/2.0/en/core-utility-libraries/security.html) makes no mention.

If it is supported, can we mention it in the docs? If it will *not* be supported, also mention it in the docs.

I know 2.x is old. I get that. The reason I'm going though this hassle is that 2.x will be the last version of CakePHP for us, and I want it to last as long as possible. I found no information on StackOverflow or anywhere on the Googles.

Thanks!

dereuromark:
Someone just needs to make a PR to support it. Then it would be supported.

josephzidell:
Can you confirm that it would involve:

- Installing `mcrypt` and adding 7.2 to travis
- Fix any broken tests

Anything I missed?

josephzidell:
Tried and appveyor failed. Any idea why?

ravage84:
@josephzidell if you refer to [this](https://ci.appveyor.com/project/cakephp/cakephp/build/1.0.8863), I guess it's unrelated, cause I do not see anything related changed.

ravage84:
Closing as pull request #11347 is up.

dereuromark:
7.2 as of this time is still excluded ( https://github.com/cakephp/cakephp/blob/2.x/.travis.yml#L45 ) and breaks, e.g. with 

> Cannot use 'object' as class name as it is reserved in [/var/www/html/vendor/cakephp/cakephp/lib/Cake/Core/Object.php, line 18]

stevenscg:
@dereuromark Looks like `Object.php` just does a class_alias. Seems like it could be a relatively small change to remove `Object.php` and replace usage with `CakeObject`. Good task for a first-time committer maybe.

stevenscg:
Since I stirred this up on IRC, there is also a warning being omitted along with the Fatal Error that dereuromark noted above that may or may not be something that the framework would want to address (vs php itself):
```
[...]

PR #11619, opened 2018-01-10
2.x Build and Pass with PHP7.2

tenkoma:
Refs #11346

In this Pull Request, Installed the `mcrypt` extension from pecl in the CI so that we can test it.

- added `pear config-set preferred_state snapshot`. Because pecl's mcrypt package is `snapshot`, is not `stable` https://pecl.php.net/package/mcrypt
- I changed `pecl install` to execute first. Because fails `composer require "phpunit/phpunit=$PHPUNIT"`
- remove cannot run test matrix: PHPUnit 3.7 + PHP7.2

### DIFF
--- a/lib/Cake/Console/Command/Task/ControllerTask.php
+++ b/lib/Cake/Console/Command/Task/ControllerTask.php
@@ -331,10 +331,10 @@ class ControllerTask extends BakeTask {
 	public function bake($controllerName, $actions = '', $helpers = null, $components = null) {
 		$this->out("\n" . __d('cake_console', 'Baking controller class for %s...', $controllerName), 1, Shell::QUIET);
 
-		if (is_null($helpers)) {
+		if ($helpers === null) {
 			$helpers = array();
 		}
-		if (is_null($components)) {
+		if ($components === null) {
 			$components = array();
 		}
 		$isScaffold = ($actions === 'scaffold') ? true : false;
--- a/lib/Cake/Console/Command/Task/ControllerTask.php
+++ b/lib/Cake/Console/Command/Task/ControllerTask.php
@@ -328,7 +328,7 @@ class ControllerTask extends BakeTask {
  * @param array $components Components to use in controller
  * @return string Baked controller
  */
-	public function bake($controllerName, $actions = '', $helpers = null, $components = null) {
+	public function bake($controllerName, $actions = '', $helpers = array(), $components = array()) {
 		$this->out("\n" . __d('cake_console', 'Baking controller class for %s...', $controllerName), 1, Shell::QUIET);
 
 		$isScaffold = ($actions === 'scaffold') ? true : false;
--- a/lib/Cake/Console/Command/Task/ControllerTask.php
+++ b/lib/Cake/Console/Command/Task/ControllerTask.php
@@ -328,9 +328,15 @@ class ControllerTask extends BakeTask {
  * @param array $components Components to use in controller
  * @return string Baked controller
  */
-	public function bake($controllerName, $actions = '', $helpers = array(), $components = array()) {
+	public function bake($controllerName, $actions = '', $helpers = null, $components = null) {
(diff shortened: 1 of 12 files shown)


dereuromark:
Now you are getting somewhere, errors become visible:

- count(): Parameter must be an array or an object that implements Countable
- ini_set(): A session is active. You cannot change the session module's ini settings at this time

Seem to be the issues left for PHP7.2 compatibility.

markstory:
> ini_set(): A session is active. You cannot change the session module's ini settings at this time

These are a big headache to fix.

tenkoma:
> ini_set(): A session is active. You cannot change the session module's ini settings at this time

I didn't fail when I only ran `CakeSessionTest`.
There was a problem with other tests.
But a new error was found.

> ini_set(): Cannot set 'user' save handler by ini_set() or session_module_name()

Probably related to the next ticket.

https://bugs.php.net/bug.php?id=73100

lib/Cake/Model/Datasource/CakeSession.php
```diff
@@ -552,6 +552,12 @@ protected static function _configureSession() {
 
 		if (!empty($sessionConfig['handler'])) {
 			$sessionConfig['ini']['session.save_handler'] = 'user';
+
+			// In PHP7.2.0+ session.save_handler can't be set to 'user' by the user.
+			// https://github.com/php/php-src/commit/a93a51c3bf4ea1638ce0adc4a899cb93531b9f0d
+			if (version_compare(PHP_VERSION, '7.2.0', '>=')) {
+				unset($sessionConfig['ini']['session.save_handler']);
+			}
```
tenkoma: I referred to the following code. However, I think it is PHP 7.2 or higher that this setting cannot be done

https://github.com/cakephp/cakephp/blob/8b3905ad6f1d2f17d20d495e12e9947bd69644fb/src/Network/Session.php#L111-L114

- session.c on PHP 7.1 branch: https://github.com/php/php-src/blob/PHP-7.1/ext/session/session.c#L552-L577
- session.c on PHP 7.2 branch: https://github.com/php/php-src/blob/PHP-7.2/ext/session/session.c#L559-L563

markstory: I think you have the correct code here. There is an open issue for database sessions in PHP7.2 (#11628)

lib/Cake/Console/Command/Task/ControllerTask.php
```diff
@@ -328,7 +328,7 @@ public function bakeActions($controllerName, $admin = null, $wannaUseSession = t
  * @param array $components Components to use in controller
  * @return string Baked controller
  */
-	public function bake($controllerName, $actions = '', $helpers = null, $components = null) {
+	public function bake($controllerName, $actions = '', $helpers = array(), $components = array()) {
```
chinpei215: Why did you change this signature?

tenkoma: It is a change to correct errors that occur in the following lines.

https://github.com/cakephp/cakephp/blob/2.x/lib/Cake/Console/Templates/default/classes/controller.ctp#L51

But it is not BC with existing applications.
Make `null` be converted to `array()`.

lib/Cake/Console/Command/Task/ControllerTask.php
```diff
@@ -328,9 +328,15 @@ public function bakeActions($controllerName, $admin = null, $wannaUseSession = t
  * @param array $components Components to use in controller
  * @return string Baked controller
  */
-	public function bake($controllerName, $actions = '', $helpers = array(), $components = array()) {
+	public function bake($controllerName, $actions = '', $helpers = null, $components = null) {
 		$this->out("\n" . __d('cake_console', 'Baking controller class for %s...', $controllerName), 1, Shell::QUIET);
 
+		if (is_null($helpers)) {
```
markstory: We usually favor `$helpers === null` as it uses fewer function calls.

PR #11620, opened 2018-01-10
Fixing issue with how the check is done for determining if templatePa…

darensipes:
…th already terminates with subDir

The problem the with current check:
`strrpos($templatePath, $subDir) == strlen($templatePath) - strlen($subDir)`
Lets say your have a templatePath that is equal to `Jobs` for example and you have a subDir equal to say `xlsx` both have a `strlen` of `4`, `strrpos` returns false and 4-4 = 0 therefore `false == 0` is `true` which is incorrect.

### DIFF
--- a/src/View/View.php
+++ b/src/View/View.php
@@ -1249,7 +1249,7 @@ class View implements EventDispatcherInterface
         if (strlen($this->subDir)) {
             $subDir = $this->subDir . DIRECTORY_SEPARATOR;
             // Check if templatePath already terminates with subDir
-            if (strrpos($templatePath, $subDir) == strlen($templatePath) - strlen($subDir)) {
+            if ($templatePath != $subDir && substr($templatePath, -(strlen($subDir))) == $subDir) {
                 $subDir = '';
             }
         }


markstory (commented):
If we could get a test for this case that would be great. If you need a hand let me know and I can work on a test.

src/View/View.php
```diff
@@ -1249,7 +1249,7 @@ protected function _getViewFileName($name = null)
         if (strlen($this->subDir)) {
             $subDir = $this->subDir . DIRECTORY_SEPARATOR;
             // Check if templatePath already terminates with subDir
-            if (strrpos($templatePath, $subDir) == strlen($templatePath) - strlen($subDir)) {
+            if ($templatePath != $subDir && substr($templatePath, -(strlen($subDir))) == $subDir) {
```
dereuromark: You can use `===` when you are comparing strings now.

saeideng: Shouldn't  `$templatePath` and `$subDir`   be equal sometimes ?

markstory: When would that happen? Extension directories are generally lower case, like `csv` while template paths usually end with the controller name `Csvs`.

darensipes: @dereuromark - yes, maybe `===` should be used.

@saeideng - I agree with mark that is shouldn't happen because `$templatePath` is using the controller name and the first letter is uppercase, whereas the `$subDir` is typically all lowercase.

@markstory - Wouldn't the test you added during this commit (https://github.com/cakephp/cakephp/commit/115a3547ab0c9120e3d2c6a3a2e6d2c3573ddc67) cover this case as well?

markstory: It does to some extent. But you have a new scenario that requires different code. With only the existing test, this change could be reverted and there wouldn't be a failing test to let us know there was a regression.

PR #11629, opened 2018-01-15
Simplify returns

carusogabriel:


src/Datasource/EntityTrait.php
```diff
@@ -391,14 +391,10 @@ public function has($property)
...
+        
+        return $value === null
             || (is_array($value) && empty($value)
-            || (is_string($value) && empty($value)))
-        ) {
-            return true;
-        }
-
-        return false;
+            || (is_string($value) && empty($value)));
```
dereuromark: I am not sure this is more readable.

src/Http/ServerRequest.php
```diff
@@ -799,11 +799,8 @@ protected function _is($type, $args)
         if (isset($detect['accept']) && $this->_acceptHeaderDetector($detect)) {
             return true;
         }
-        if (isset($detect['param']) && $this->_paramDetector($detect)) {
-            return true;
-        }
```
dereuromark: The early return is a better pattern in this case, easier to grasp and thus less mistakes when modifying.

dereuromark (approved):

markstory:
Thanks!

Issue #11627, opened 2018-01-15
Can not reset mailer replyTo


caladfelipe:
* [x] bug

* CakePHP Version: 3.5.10
* Platform and Target: Apache, MySQL

Tried to reset email replyTo but an error occurs:

$mailer->setReplyTo([]); // Error = "Reply-To requires only 1 email address."

### What you expected to happen
To reset the email replyTo to anybody, technically to an empty array.

markstory:
This is an issue with all of the email address setter methods. Currently it is not possible to unset/clear any email address property.

dereuromark:
I usually use reset() and then set it again.

markstory:
@dereuromark But that will reset all of the mailer state, which is not always desirable.

Pull request open now.

PR #11633, opened 2018-01-15
Allow address properties to be reset with an empty array.

markstory:
Allow an empty array to be used to reset an individual address property to its original state.

Refs #11627

### DIFF
--- a/src/Mailer/Email.php
+++ b/src/Mailer/Email.php
@@ -1012,6 +1012,12 @@ class Email implements JsonSerializable, Serializable
      */
     protected function _setEmailSingle($varName, $email, $name, $throwMessage)
     {
+        if ($email === []) {
+            $this->{$varName} = $email;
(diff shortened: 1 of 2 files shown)


ADmad:
Couldn't this change allow a property to be inadvertently reset? IMHO it would be better to provide a more explicit way to reset an address property.

dereuromark:
Both [] and null as values intuitively sound like setting it to null = resetting.
But you are right, for a patch version this can lead in some cases to unexpected behavior since this wasn't that strict yet.
Would this be sth for 3.6 then maybe?
Otherwise we would have to introduce quite a few reset methods here on top.

markstory:
@ADmad passing an array is pretty explicit. I hear you on this being risky for a patch release. I will move this to 3.6.

ADmad (approved):

Issue #11631, opened 2018-01-15
Add Table::getBehavior() method


sdustinh:
This is a (multiple allowed):

* [ ] bug
* [x] enhancement
* [ ] feature-discussion (RFC)

```bash
sdustinh@sdustinh:~/github/cakephp/app$ php --version
PHP 7.1.13-1+ubuntu16.04.1+deb.sury.org+1 (cli) (built: Jan  5 2018 13:26:45) ( NTS )
Copyright (c) 1997-2017 The PHP Group
Zend Engine v3.1.0, Copyright (c) 1998-2017 Zend Technologies
    with Zend OPcache v7.1.13-1+ubuntu16.04.1+deb.sury.org+1, Copyright (c) 1999-2017, by Zend Technologies
    with Xdebug v2.5.5, Copyright (c) 2002-2017, by Derick Rethans
sdustinh@sdustinh:~/github/cakephp/app$ mysql --version
mysql  Ver 14.14 Distrib 5.7.20, for Linux (x86_64) using  EditLine wrapper
sdustinh@sdustinh:~/github/cakephp/app$ composer show cakephp/*
cakephp/bake                1.5.4  Bake plugin for CakePHP 3
cakephp/cakephp             3.5.6  The CakePHP framework
cakephp/cakephp-codesniffer 3.0.1  CakePHP CodeSniffer Standards
cakephp/chronos             1.1.2  A simple API extension for DateTime.
cakephp/debug_kit           3.11.2 CakePHP Debug Kit
cakephp/elastic-search      0.3.5  An Elastic Search datasource and data mapper for CakePHP 3.0
cakephp/migrations          1.7.1  Database Migration plugin for CakePHP 3.0 based on Phinx
cakephp/plugin-installer    1.0.0  A composer installer for CakePHP 3.0+ plugins.
```

Currently, in my application we have a lot of methods that are provided by behaviors. One of the problems that comes up with this is it gets hard to find what methods are provided by which behavior. To this end, I have been avoiding the `Behavior::buildValidator()` event and just exposing the methods. In my table I have been doing the following:

```php
public function validationDefault(Validator $validator)
{
    $validator = $this->behaviors()->get('BehaviorName')->validationName($validator);

    // Additional validation rules...

    return $validator;
}
```

As I was typing this out, I just thought it would be nice to have a `getBehavior()` method similar to the `getTable()`, `getAssociation()`, etc. I can open a PR for this if you'd like, but wanted to put an issue in to poll support.

[...]

PR #11635, opened 2018-01-16
Table get behavior

sdustinh:
Reference: #11631

### DIFF
--- a/src/ORM/Table.php
+++ b/src/ORM/Table.php
@@ -927,9 +927,18 @@ class Table implements RepositoryInterface, EventListenerInterface, EventDispatc
      *
      * @param string $name The behavior alias to get from the registry.
      * @return \Cake\ORM\Behavior
+     * @throws \InvalidArgumentException If the behavior does not exist.
      */
     public function getBehavior($name)
     {
+        if ($this->hasBehavior($name) === false) {
+            throw new InvalidArgumentException(sprintf(
+                'The %s behavior is not defined on %s.',
+                $name,
+                get_class($this)
+            ));
+        }
+
         return $this->_behaviors->get($name);
     }
 
--- a/src/ORM/Table.php
+++ b/src/ORM/Table.php
@@ -923,6 +923,17 @@ class Table implements RepositoryInterface, EventListenerInterface, EventDispatc
     }
 
     /**
+     * Get a behavior from the registry.
+     *
+     * @param string $name The behavior alias to get from the registry.
+     * @return \Cake\ORM\Behavior|null
+     */
+    public function getBehavior($name)
+    {
+        return $this->_behaviors->get($name);
+    }
+
+    /**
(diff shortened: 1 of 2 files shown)


src/ORM/Table.php
```diff
@@ -922,6 +922,17 @@ public function behaviors()
         return $this->_behaviors;
     }
 
+    /**
+     * Get a behavior from the registry.
+     *
+     * @param string $name The behavior alias to get from the registry.
+     * @return \Cake\ORM\Behavior|null
```
dereuromark: A getter ideally doesn't return null, it throws a clear and meaningful exception instead.
Especially since a "has" method is available.

This makes code flow easier, as you can expect the object to be there to directly use the return value in chaining.
The get vs find pattern we use everywhere else, as well, where possible.

saeideng: `$this->behaviors()->get('BehaviorName')` returns `null` ...

dereuromark: Even that one shouldnt return null :) But that is now in code, and we need to wait for 4.x to fix that for BC reasons.
For now we shouldn't introduce more baggage then to clean up with the next major IMO.

sdustinh: I'd be happy to change it if you'd like @dereuromark. I added the `null` because that is what the `ObjectRegistry` had. Would you like me to update the docblock?

sdustinh: Also, should the test be updated to not check for a `null` return?

dereuromark: Let's see what the discussion around my proposal https://github.com/cakephp/cakephp/pull/11638 yields first.

markstory (approved):
Looks good to me.

dereuromark:
@markstory Shouldnt we remove the nullable part here?

markstory:
@dereuromark Yes, I posted my review before looking at the changes to make getAssociation() more strict.

src/ORM/Table.php
dereuromark: So, yeah, you can remove the nullable part and throw an exception instead. People can check with hasBehavior() if needed.

sdustinh:
Nullable return removed from the docblock and the test

src/ORM/Table.php
```diff
@@ -922,6 +922,17 @@ public function behaviors()
...
 
+    /**
+     * Get a behavior from the registry.
+     *
+     * @param string $name The behavior alias to get from the registry.
+     * @return \Cake\ORM\Behavior
+     */
+    public function getBehavior($name)
+    {
+        return $this->_behaviors->get($name);
```
ADmad: You need to check the return value and throw an exception if it's null.

sdustinh:
@ADmad Updated to throw an exception. Could you check it over for me to make sure it is what you guys want? I used the `hasBehavior()` method to do the check and bailed early instead of checking the result.

Issue #11625, opened 2018-01-12
check if route exists with $defaults and $options


EliasDeVos:
This is a (multiple allowed):

* [ ] bug
* [x] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.5

Hi,
Right now i have a use case where i need to check if a given route exists.
I would create the route like this Router::url($defaults, $options).

But right now if the route does not exist it throws an error

https://github.com/cakephp/cakephp/blob/master/src/Routing/RouteCollection.php#L347

So i have to work with try catch to solve this.

Would it be an option to add a new static function in \Cake\Core\Router routeExists($defaults, $options)

which would return false if it doesn't exist and true (or maybe the url) if it does ?

Let me know if this would be a good enhancement and i will create a pull request

markstory:
I thinks that's a reasonable addition to `Router`. It could be a simple try/catch wrapper around `Router::url()`.

saeideng:
`Router::urlOrNull()` :)

EliasDeVos:
@markstory won't there be a lot of issues if we would add try/catch in the `Router::url()` method ? because the code would keep on going instead of halting and throwing an error ?
Maybe @saeideng suggestion is safer ? let me know what you think and i will make a pull request

markstory:
@EliasDeVos I wasn't very clear. I was trying to say that `Router::routeExists()` could be a try/catch wrapper around `Router::url()`. Not that we should add try/catch inside of `Router::url()`.

markstory:
Closing as there is a pull request open now.

PR #11636, opened 2018-01-16
add urlOrNull method in router and test

EliasDeVos:
pull request for the following issue #11625,

I didn't know if i had to copy the docblocks for usage from `Router::url` or just refer to them.
I copied them now but this can be changed

### DIFF
--- a/src/Routing/Router.php
+++ b/src/Routing/Router.php
@@ -16,8 +16,8 @@ namespace Cake\Routing;
 
 use Cake\Core\Configure;
 use Cake\Http\ServerRequest;
-use Cake\Utility\Inflector;
 use Cake\Routing\Exception\MissingRouteException;
+use Cake\Utility\Inflector;
 use Psr\Http\Message\ServerRequestInterface;
 
 /**
--- a/src/Routing/Router.php
+++ b/src/Routing/Router.php
@@ -692,27 +692,7 @@ class Router
      *
      * ### Usage
      *
-     * - `Router::url('/posts/edit/1');` Returns the string with the base dir prepended.
-     *   This usage does not use reverser routing.
-     * - `Router::url(['controller' => 'posts', 'action' => 'edit']);` Returns a URL
-     *   generated through reverse routing.
-     * - `Router::url(['_name' => 'custom-name', ...]);` Returns a URL generated
-     *   through reverse routing. This form allows you to leverage named routes.
-     *
-     * There are a few 'special' parameters that can change the final URL string that is generated
-     *
-     * - `_base` - Set to false to remove the base path from the generated URL. If your application
-     *   is not in the root directory, this can be used to generate URLs that are 'cake relative'.
-     *   cake relative URLs are required when using requestAction.
-     * - `_scheme` - Set to create links on different schemes like `webcal` or `ftp`. Defaults
-     *   to the current scheme.
-     * - `_host` - Set the host to use for the link. Defaults to the current host.
-     * - `_port` - Set the port if you need to create links on non-standard ports.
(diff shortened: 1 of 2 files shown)


src/Routing/Router.php
```diff
@@ -682,6 +682,54 @@ public static function url($url = null, $full = false)
...
+     *   string.
+     * @param bool $full If true, the full base URL will be prepended to the result.
+     *   Default is false.
+     * @return string Full translated URL with base path or null if url does not exist
+     */
+    public static function urlOrNull($url = null, $full = false)
+    {
+        try {
+            return static::url($url, $full);
+        } catch (\Cake\Routing\Exception\MissingRouteException $e) {
```
ADmad: Add a "use" at top of file instead of using FQCN.

ADmad (changes requested):

tests/TestCase/Routing/RouterTest.php
```diff
@@ -129,6 +129,19 @@ public function testRouteDefaultParams()
...
+     * testRouteUrlOrNull method
+     *
+     * @return void
+     */
+    public function testRouteUrlOrNull()
+    {
+        Router::connect('/:controller/:action', ['controller' => 'posts']);
+        $this->assertEquals(Router::urlOrNull(['action' => 'view']), '/view');
+
+        $this->assertEquals(Router::urlOrNull(['action' => 'view', 'controller' => 'users', 'plugin' => 'test']), null);
```
ADmad: Use `assertNull()`

ADmad (changes requested):

src/Routing/Router.php
```diff
@@ -682,6 +683,54 @@ public static function url($url = null, $full = false)
...
+     *
+     * @param string|array|null $url An array specifying any of the following:
+     *   'controller', 'action', 'plugin' additionally, you can provide routed
+     *   elements or query string parameters. If string it can be name any valid url
+     *   string.
+     * @param bool $full If true, the full base URL will be prepended to the result.
+     *   Default is false.
+     * @return string Full translated URL with base path or null if url does not exist
+     */
+    public static function urlOrNull($url = null, $full = false)
```
dereuromark: Why would the first argument be nullable?

EliasDeVos: I just took the same default values that are passed to `Router:;url()` but i could remove the nullable for $url

EliasDeVos: @dereuromark yes that was the idea :P

src/Routing/Router.php
```diff
@@ -682,6 +683,54 @@ public static function url($url = null, $full = false)
...
+     *   cake relative URLs are required when using requestAction.
+     * - `_scheme` - Set to create links on different schemes like `webcal` or `ftp`. Defaults
+     *   to the current scheme.
+     * - `_host` - Set the host to use for the link. Defaults to the current host.
+     * - `_port` - Set the port if you need to create links on non-standard ports.
+     * - `_full` - If true output of `Router::fullBaseUrl()` will be prepended to generated URLs.
+     * - `#` - Allows you to set URL hash fragments.
+     * - `_ssl` - Set to true to convert the generated URL to https, or false to force http.
+     * - `_name` - Name of route. If you have setup named routes you can use this key
+     *   to specify it.
```
markstory: Instead of duplicating the options you should use `@see Router::url()`

markstory:
Is getting a string or null back the API we want to provide? I originally thought we were going for something like

```php
if (Router::routeExists($url)) {
  // Do something, possibly call Router::url($url);
}
```

The difference is returning a `bool` vs `string|null`

EliasDeVos:
@markstory i changed it, now it checks and returns true or false

markstory:
Thank you @EliasDeVos :clap:

PR #11638, opened 2018-01-16
Use a stricter get+has pattern for associations on tables.

dereuromark:
association() in 3.6 has been deprecated in favor of a new getAssociation().
But similar to other get/has/find pattern uses in Core, I suggest that we apply a bit stricter get syntax here too. 
I consider that cleaner code, as the flow of chaining can expect the object to be directly present - or otherwise get a meaningful exception instead of a bad fatal error.
Whoever needs, can use a "has..." check beforehand to ensure the object existence in optional cases.

This relates to https://github.com/cakephp/cakephp/pull/11635 and hasBehavior() vs getBehavior() etc.

I kept the find method protected, as get+has should usually suffice for the public API.
I will add tests once this stricter approach is approved.
I consider this reasonably BC since the method gets cleanly deprecated via migration guide and we can point out the stricter behavior there - as well as in code doc block.

### DIFF
--- a/src/ORM/Table.php
+++ b/src/ORM/Table.php
@@ -968,7 +968,7 @@ class Table implements RepositoryInterface, EventListenerInterface, EventDispatc
     {
         $association = $this->findAssociation($name);
         if (!$association) {
-            throw new InvalidArgumentException('Association does not exist: ' . $name);
+            throw new InvalidArgumentException("The {$name} association is not defined on {$this->getAlias()}.");
         }
 
         return $association;
--- a/src/ORM/Table.php
+++ b/src/ORM/Table.php
@@ -934,7 +934,7 @@ class Table implements RepositoryInterface, EventListenerInterface, EventDispatc
     }
 
     /**
-     * Returns an association object configured for the specified alias if any
+     * Returns an association object configured for the specified alias if any.
      *
      * @deprecated 3.6.0 Use getAssociation() instead.
      * @param string $name the alias used for the association.
@@ -942,9 +942,9 @@ class Table implements RepositoryInterface, EventListenerInterface, EventDispatc
      */
     public function association($name)
     {
-        deprecationWarning('Use Table::getAssociation() instead.');
+        deprecationWarning('Use Table::getAssociation() and Table::hasAssocation() instead.');
 
-        return $this->getAssociation($name);
+        return $this->findAssociation($name);
     }
 
     /**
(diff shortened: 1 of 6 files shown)


src/ORM/Table.php
```diff
@@ -934,11 +934,11 @@ public function hasBehavior($name)
     }
 
     /**
-     * Returns an association object configured for the specified alias if any
+     * Returns an association object configured for the specified alias if any.
      *
      * @deprecated 3.6.0 Use getAssociation() instead.
      * @param string $name the alias used for the association.
-     * @return \Cake\ORM\Association|null Either the association or null.
+     * @return \Cake\ORM\Association The association.
```
ADmad: I don't see the benefit of changing return type of deprecated method. Should be done just for `getAssociation()`.

src/ORM/Table.php
```diff
@@ -956,10 +956,54 @@ public function association($name)
      * $users = $this->getAssociation('Articles.Comments.Users');
      * ```
      *
-     * @param string $name the alias used for the association.
+     * Note that this method requires the association to be present or otherwise
+     * throws an exception.
+     * If you are not sure, use hasAssociation() before calling this method.
+     *
+     * @param string $name The alias used for the association.
      * @return \Cake\ORM\Association|null Either the association or null.
```
ADmad: Return type needs to be updated.

src/ORM/Table.php
dereuromark: That was an oversight, my bad.

src/ORM/Table.php
dereuromark: I changed the wrong class method. I will fix this.

dereuromark:
If this is a bit too strict of a change in a minor version, we can also hold off on this for 4.0.
It looks like even in core a lot of functionality would have to be adjusted.

src/ORM/Table.php
```diff
@@ -934,17 +934,61 @@ public function hasBehavior($name)
...
+     *
+     * @param string $name The alias used for the association.
+     * @return \Cake\ORM\Association The association.
+     * @throws \InvalidArgumentException
+     */
+    public function getAssociation($name)
+    {
+        $association = $this->findAssociation($name);
+        if (!$association) {
+            throw new InvalidArgumentException('Association does not exist: ' . $name);
```
markstory: Error message could provide more context: `The {$name} association is not defined on {$this->name()}`.

src/ORM/Table.php
```diff
@@ -934,17 +934,61 @@ public function hasBehavior($name)
...
+     * $users = $this->getAssociation('Articles.Comments.Users');
+     * ```
+     *
+     * Note that this method requires the association to be present or otherwise
+     * throws an exception.
+     * If you are not sure, use hasAssociation() before calling this method.
+     *
+     * @param string $name The alias used for the association.
+     * @return \Cake\ORM\Association The association.
+     * @throws \InvalidArgumentException
```
markstory: Adding exceptions is 🆗 in minor releases, if we document the change. I don't see this representing a significant upgrade hurdle.

markstory: Looking at the internal test failures it looks like most of the failures come from a small number of call sites.

dereuromark:
Looks good, thx for looking into this.
I added a test for the new exception being thrown.

PR #11643, opened 2018-01-19
[2.x] Make mcrypt optional

chinpei215:
Now Security::encrypt() and Security::decrypt() work with openssl if the mcrypt extension is unavailable. Note that Security::rijndael() doesn't work with openssl.

And, I have changed Security::randomBytes() to fallback to mcrypt_create_iv() if the openssl extension is unavailable.

Relates to #11346

### DIFF
--- a/lib/Cake/Test/Case/Utility/SecurityTest.php
+++ b/lib/Cake/Test/Case/Utility/SecurityTest.php
@@ -36,7 +36,7 @@ class SecurityTest extends CakeTestCase {
  */
 	public function setUp() {
 		parent::setUp();
-		Security::engine(null);
+		Configure::delete('Security.useOpenSsl');
 	}
 
 /**
@@ -46,26 +46,7 @@ class SecurityTest extends CakeTestCase {
  */
 	public function tearDown() {
 		parent::tearDown();
-		Security::engine(null);
-	}
-
-/**
- * Tests that Security::engine() works
- *
- * @return void
- */
-	public function testEngine() {
-		if (extension_loaded('mcrypt')) {
-			$this->assertEquals('mcrypt', Security::engine());
-		}
-
-		$this->assertContains(Security::engine(), array('mcrypt', 'openssl'));
-
-		Security::engine('mcrypt');
-		$this->assertEquals('mcrypt', Security::engine());
-
-		Security::engine('openssl');
-		$this->assertEquals('openssl', Security::engine());
+		Configure::delete('Security.useOpenSsl');
 	}
 
 /**
@@ -385,24 +366,24 @@ class SecurityTest extends CakeTestCase {
  */
 	public function testEncryptDecryptCompatibility($txt) {
 		$this->skipIf(!extension_loaded('mcrypt'), 'This test requires mcrypt to be installed');
-		$this->skipIf(!extension_loaded('openssl'), 'This test requires oepnssl to be installed');
-		$this->skipIf(version_compare(PHP_VERSION, '5.3.3', '<'), 'This test requires PHP 5.3.3 or grater');
+		$this->skipIf(!extension_loaded('openssl'), 'This test requires openssl to be installed');
+		$this->skipIf(version_compare(PHP_VERSION, '5.3.3', '<'), 'This test requires PHP 5.3.3 or greater');
 
 		$key = '12345678901234567890123456789012';
 
-		Security::engine('mcrypt');
+		Configure::write('Security.useOpenSsl', false);
 		$mcrypt = Security::encrypt($txt, $key);
 
-		Security::engine('openssl');
+		Configure::write('Security.useOpenSsl', true);
 		$openssl = Security::encrypt($txt, $key);
 
 		$this->assertEquals(strlen($mcrypt), strlen($openssl));
 
-		Security::engine('mcrypt');
+		Configure::write('Security.useOpenSsl', false);
 		$this->assertEquals($txt, Security::decrypt($mcrypt, $key));
 		$this->assertEquals($txt, Security::decrypt($openssl, $key));
 
-		Security::engine('openssl');
+		Configure::write('Security.useOpenSsl', true);
 		$this->assertEquals($txt, Security::decrypt($mcrypt, $key));
 		$this->assertEquals($txt, Security::decrypt($openssl, $key));
 	}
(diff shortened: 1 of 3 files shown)


chinpei215:
More readable diff: https://github.com/cakephp/cakephp/pull/11643/files?w=1#diff-03e686db7da4edec51a54aa2ec744470
You would realize that I didn't touch existing code implemented with mcrypt.

lib/Cake/Test/Case/Utility/SecurityTest.php
```diff
@@ -337,6 +376,54 @@ public function testEncryptDecrypt() {
...
+ * Tests that encrypted strings are comatible between the mcrypt and openssl engine.
+ *
+ * @dataProvider plainTextProvider
+ * @param string $txt Plain text to be encrypted.
+ * @return void
+ */
+	public function testEncryptDecryptCompatibility($txt) {
+		$this->skipIf(!extension_loaded('mcrypt'), 'This test requires mcrypt to be installed');
+		$this->skipIf(!extension_loaded('openssl'), 'This test requires oepnssl to be installed');
+		$this->skipIf(version_compare(PHP_VERSION, '5.3.3', '<'), 'This test requires PHP 5.3.3 or grater');
```
markstory: typo in 'greater'

lib/Cake/Test/Case/Utility/SecurityTest.php
```diff
@@ -337,6 +376,54 @@ public function testEncryptDecrypt() {
 		$this->assertEquals($txt, Security::decrypt($result, $key));
 	}
 
+/**
+ * Tests that encrypted strings are comatible between the mcrypt and openssl engine.
```
markstory: typo in 'compatible'

lib/Cake/Utility/Security.php
```diff
@@ -404,14 +426,42 @@ public static function decrypt($cipher, $key, $hmacSalt = null) {
...
 
+/**
+ * Set or get the encryption engine
+ *
+ * @param string $engine The encryption engine to use
+ * @return string
+ */
+	public static function engine($engine = null) {
+		if (func_num_args() > 0) {
+			static::$_engine = $engine;
```
markstory: Should we validate that the chosen engine exists? Right now someone could do `Security::engine('lol')` and not get an error.

ADmad: "lol" powered crypto :muscle:

markstory:
This seems like a fairly large change for a bugfix release. Is it unreasonable to require people to use the mcrypt pecl extension?

lib/Cake/Utility/Security.php
chinpei215: I was lazy a little bit, as I added the method for testing.

chinpei215:
Since I saw many users are struggle with installing mcrypt, I would like to make it optional.

What changes do you feel like a fairly large? If you don't like to add new `Security::engine()`, it can be removed. At first there was no such method, and I was writing `Configure::read('Security.useOpenSSL')` instead. It will default to `!extention_loaded('mcrypt') && extention_loaded('openssl') && version_compare(PHP_VERSION, '5.3.3', '>=')`.

Or if you don't like to touch `Security` class, then I can define `mcrypt_*()` functions by using `openssl_*()` functions if the mcrypt extension is not loaded. A disadvantage of this approach would be that it would be hard to test compatibility between mcrypt and openssl.

markstory:
Adding support for another encryption extension is something we wouldn't normally do in a bugfix release. I am also worried that auto selecting which extension to use is not transparent. The same application code can use different implementations with no code changes. It would be better if the developer had to opt-in to using openssl.

chinpei215:
If we could release 2.11, I would choose 2.x. But 2.10 is the final 2.x release. Since 2.10 is 'Long Term Support', I think it would be better to not require any deprecated extensions. Even if some vulnerabilities were to be found in the extensions, people would not be able to get any official patches.

I can stop auto selecting extensions. If you have any other concerns please tell me. I am planing to add some more commits on this pull request this weekend.

markstory:
I think defaulting to mcrypt is the safest option for backwards compatibility. You have good points around giving this long lived release better options for using undeprecated extensions.

lib/Cake/Test/Case/Utility/SecurityTest.php
tersmitten: And `oepnssl`

dereuromark:
Are we OK with moving forward here, has this been tested in 2.x apps to avoid any BC issues arising?

markstory:
@dereuromark Sure. From @chinpei215's last update there were a few more commits he wanted to add.

chinpei215:
Sorry, I forgot doing that. I will remove Security::engine().

nacc:
Hello,

We are moving to PHP7.2 exclusively in Ubuntu Bionic (18.04). What is the status of this MP / mcrypt support generally. I would prefer updating to a released version, so will there be a 2.10.8 containing these changes in the next week or so (our feature freeze is coming up end of the month). If not, we might just remove cakephp from Ubuntu.

-Nish

markstory:
@nacc Would you be open to having CakePHP 3.x instead? It doesn't have the same issues with mcrypt and is where we are focusing our future efforts.

markstory:
To be honest I didn't even know we were available through apt-get in debian/ubuntu, and don't know what if anything is happening in Debian.

The upgrade path from 2.x to 3.x is pretty challenging and not an easy undertaking.

nacc:
Ok, that all being said, I think we'll probably drop it from 18.04 then. I will document it in our release notes. Is there a good landing page to direct users who maybe were using 2.x on a prior release?

markstory:
The [installation guide](https://book.cakephp.org/2.0/en/installation.html) for 2.x might be a good place, as it will give them information on how they can reinstall CakePHP again.

nacc:
@markstory great, thanks!

ADmad:
@nacc You will actually be doing CakePHP users a favor by removing that outdated upstream CakePHP package from Ubuntu. Our 3.x branch released over 2 years already doesn't require mcrypt.

We have no idea who the maintainer of that debian package is. It wasn't created / maintained in consultation with the CakePHP development team.

chinpei215:
I added more commits just now. Sorry for the late update, but I didn't think 2.10.8 would be released soon. So I placed  low priority on this pull request.

PR #11646, opened 2018-01-19
Unset the save_handler for PHP7.2

### LINKED ISSUE #11628: Unable to configure the session, setting session.save_handler failed (PHP 7.2)
This is a (multiple allowed):

* [x] bug
* [ ] enhancement

* CakePHP Version: 3.5.10
* Platform and Target: Ubuntu 16.04, Apache 2.4, PHP 7.2 (7.2.1-1+ubuntu16.04.1+deb.sury.org+1)

### What you did
Updated php from 7.0.x to 7.2

### What happened
image

Found a similar issue here: https://github.com/symphonycms/symphony-2/issues/2783

The check in https://github.com/cakephp/cakephp/blob/83815ac4912656eb9e724e179bbe1dc6440e3359/src/Network/Session.php#L113 evaluates to false, so I don't quite know why this is happening, but removing line 166 from that file solves the issue

markstory:
Forward port changes from #11619 and fix #11628

### DIFF
--- a/src/Network/Session.php
+++ b/src/Network/Session.php
@@ -108,10 +108,14 @@ class Session
             $sessionConfig['ini']['session.name'] = $sessionConfig['cookie'];
         }
 
-        // In PHP7.1.0+ session.save_handler can't be set to user by the user.
-        // https://github.com/php/php-src/blob/master/ext/session/session.c#L559
-        if (!empty($sessionConfig['handler']) && version_compare(PHP_VERSION, '7.1.0', '<=')) {
+        if (!empty($sessionConfig['handler'])) {
             $sessionConfig['ini']['session.save_handler'] = 'user';
+        }
+
+        // In PHP7.2.0+ session.save_handler can't be set to user by the user.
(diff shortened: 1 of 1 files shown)


dereuromark:
Is it possible that this breaks intregration test cases?

with current master:

> Warning Error: session_set_save_handler(): Cannot change save handler when headers already sent in [/home/vagrant/Apps/.../vendor/cakephp/cakephp/src/Network/Session.php, line 224]

```php
$this->post(['controller' => 'Integration', 'action' => 'ciProject'], $data);
```
worked before.

dereuromark:
Actually, now in PHP7.2 this seems to also exist in previous versions - as it does with latest master.
We probably have to do another kind of patch for IntegrationTestCase usage.

markstory:
This shouldn't break integration tests, as we are unsetting the save handler keys so they should not be modified.

dereuromark:
Current head or dev-master is throwing this error on running integration tests and PHP7.2.

markstory:
Are you using a custom session handler?

dereuromark:
No, normal defaults and db one:
```php
    'Session' => [
        'defaults' => 'database',
        'timeout' => 60000,
```

PR #11653, opened 2018-01-23
Fix saveOrFail() to include useful messages on output, e.g. CLI/tests.

dereuromark:
Currently saveOrFail() is pretty unusable.
I so far used my "strict=>true" of Shim plugin which outputted the issue directly for CLI and test usage.
With saveOrFail() you only get:
```
1) App\Test\TestCase\Model\Table\CapabilitiesTableTest::testImport
Cake\ORM\Exception\PersistenceFailedException: Entity save failure.
```

With the provided fix we at least get
```
1) App\Test\TestCase\Model\Table\CapabilitiesTableTest::testImport
Cake\ORM\Exception\PersistenceFailedException: Entity save failure (name: "The provided value is invalid").
```
which really helps and avoids having to add debug() into the whole code everywhere and rerun it :)

For multiple messages it uses the identifier keys if possible to avoid too long messages.

### DIFF
--- a/src/ORM/Exception/PersistenceFailedException.php
+++ b/src/ORM/Exception/PersistenceFailedException.php
@@ -85,5 +85,4 @@ class PersistenceFailedException extends Exception
     {
         return $this->_entity;
     }
-
 }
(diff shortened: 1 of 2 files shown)


markstory:
:clap: for better error messages.

PR #11660, opened 2018-01-24
2.x - Change PHPUnit version as requested by test shell

davidyell:
I'm trying to run the SessionHelper tests and I can't. The closest I have got is the following.

Test shell requesting php unit 3.7

So I changed the `composer.json` to use PHP Unit 3.7 and then the test shell works.

Apologies for the misleading folder name, this is on the 2.x branch 😄

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -22,7 +22,7 @@
 		"ext-mcrypt": "*"
 	},
 	"require-dev": {
-		"phpunit/phpunit": "<6.0.0",
+		"phpunit/phpunit": "^3.7",
 		"cakephp/cakephp-codesniffer": "^1.0.0"
 	},
 	"config": {


dereuromark:
:+1: Not sure how that ever happened - but indeed 2.x only works with phpunit3.x

markstory:
@dereuromark There was some work done to improve compatibility with PHPUnit 4 and 5 at least for applications.

tenkoma:
Just like TravisCI Config, add the following to the `app/Config/bootstrap.php` , I think that you can run the core test with PHPUnit 4.* / 5.*.

```php
require_once ROOT . DS . 'vendors/autoload.php';
```

https://github.com/cakephp/cakephp/blob/49725a8f864faf11387446253cfd7b4a9582fd40/.travis.yml#L52

I think it is difficult to commit changes to the core. Because composer autoload is optional in CakePHP2.

Issue #11657, opened 2018-01-23
Automatically use custom parent field for tree behavior in find('threaded')


mrothauer:
This is a (multiple allowed):

* [ ] bug
* [x] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.5.11
* Platform and Target: Apache 2, MySQL 5.7

I created a Tree behavior in a table's initialize method, **with a custom parent_id field**

```
public function initialize(array $config)
    {
        $this->addBehavior('Tree', [
            'parent' => 'my_custom_id_field',
        ]);
    }
```
I then wanted to retrieve nested data with that piece of code

```
        $result = $this->find('threaded')
```

I expected the children to be included in the resultset, but they **were missing**. After some research, I found the (**well documented**) option to pass the my_custom_id_field as a parameter

```
        $result = $this->find('threaded', [
            'parentField' => 'my_custom_id_field'
        ]);
```
This works great, but: Is there any possiblity to automatically use the parent field that was passed on intialization of the behavior?

Thanx,
Mario

markstory:
> This works great, but: Is there any possiblity to automatically use the parent field that was passed on intialization of the behavior?

Seems like a reasonable default. Any chance you'd be willing to put together a pull request for `3.next` to implement it? I'm happy to help with tests if you need a hand.

mrothauer:
I'm happy to create a pull request. At least I will try :-)

markstory:
Closing as a pull request is open.

Issue #11240, opened 2017-09-26
RFC Explicit finderMap also for Table classes


dereuromark:
This is a (multiple allowed):

* [ ] bug
* [x] enhancement
* [x] feature-discussion (RFC)

Currently, the magic find(type) is also invoked for false positives, e.g.:
```php
    public function findByPullRequest($pr)
    {
        return $this->find()->where(['pr' => $pr]);
    }
```
as find('byPullRequest') etc, which leads to fatal errors due to the wrong typehinting.

Of course a project could try to prevent methods that start similar to findX naming scheme.
But maybe it would be cleaner to have a similar explicit finderMap for custom finders on Table level, as well.
This way there are no false positives and you can explicitly map them out the way they should work.

This also simplifies introspection and IDE helper support here (refs https://github.com/dereuromark/cakephp-ide-helper/issues/49 ).

inoas:
I don't want to declare the finder types beyond declaring the functions as I would repeat myself and then can even forget to remove finder type declarations after removing finders.

dereuromark:
also findOrCreate() is currently already wrongly accessible this way.

ADmad:
Having a map for custom finders sounds reasonable as long as it's optional. Similar to how we can optionally have a map for mixin methods and events in behaviors.

inoas:
Optional sounds okay, but is there no better solution that is DRY?

ADmad:
The "magic" detection of custom finders it's already DRY. Only those who need to make their IDEs happy need to setup the map.

dereuromark:
So an optional map makes sense then, yeah.

markstory:
Is the optional map just for IDE completion? Couldn't we do something similar with doc tags?

dereuromark:
It would also be used if you wanted a stricter mapping on project level. similar to behaviors and their mapping.

dereuromark:
Without this, I had to do https://github.com/dereuromark/cakephp-ide-helper/pull/147
It would be nice if we could revisit a more explicit mapping instead of too much magic here in the future.

PR #11662, opened 2018-01-24
findThreaded uses the tree's parent_id field as default parentField

mrothauer:
Issue #11657

This code changing works for my use case. Maybe we can remove the whole 'parentField' option? I see no use for it any more, as it gets the value from the tree options now. What do you think?

My TableTest rises a lot of errors, i guess that's due to the missing database setup. Is there some documentation how to configure the local database in phpunit.xml?

`<env name="db_dsn" value="mysql://localhost/cake_test?timezone=UTC"/>`

I created a database named "cake_test", mysql server is running and the errors persist.

Thank you,
Mario

### DIFF
--- a/src/ORM/Table.php
+++ b/src/ORM/Table.php
@@ -1478,9 +1478,10 @@ class Table implements RepositoryInterface, EventListenerInterface, EventDispatc
      */
     public function findThreaded(Query $query, array $options)
     {
+        $treeBehavior = $this->behaviors()->get('Tree');
         $options += [
             'keyField' => $this->getPrimaryKey(),
-            'parentField' => 'parent_id',
+            'parentField' => $treeBehavior ? $treeBehavior->getConfig('parent') : 'parent_id',
             'nestingKey' => 'children'
         ];
 


src/ORM/Table.php
```diff
@@ -1478,9 +1478,10 @@ public function findList(Query $query, array $options)
      */
     public function findThreaded(Query $query, array $options)
     {
+        $treeBehavior = $this->behaviors()->get('Tree');
```
saeideng: I'm not sure this method rely on `Tree behavior`

markstory: You might have to use `has()` instead. Something like

```php
$parent = 'parent_id';
if ($this->behaviors()->has('Tree')) {
  $parent = $this->behaviors->get('Tree')->getConfig('parent');
}
```

markstory (commented):
Looking good. Do you need a hand with tests?

mrothauer:
Ok I'll use the has- function. I reported my troubles with the tests in the first comment.

src/ORM/Table.php
dereuromark: Actually, table class itself has method `hasBehavior()` :)

markstory:
Sorry about missing your comment on test setup earlier. After reading the code, I forgot all about it.

If you have SQLite installed you should be able to run tests without doing any configuration as the test suite defaults to using an in-memory SQLite database.

sdustinh:
Won't these changes break tables that have a behavior aliased as `Tree` that isn't the framework provided behavior?

dereuromark:
@sdustinh That's true - but even a non-core one would most likely behave the same here after all.
Also, via 3.6 this can be clearly communicated in the migration guide for the unlikely event you describe.

sdustinh:
The only reason I bring it up is I have 2 behaviors with that name (Binary and IP subnet) tree behaviors that use this alias and don't have a `parent` configuration. I can handle the changes with minor updates but I figured I would mention it for others.

markstory:
@sdustinh What if we checked for the `Tree` alias and a `parent` config key? That should help ensure it is more specific and doesn't capture non core Tree behaviors.

sdustinh:
I don't mind the current implementation. I just figured if I had a situation that would need patches that someone else might.

If you're going the distance to check the config you might as well just check the class name.

ADmad:
I too am not thrilled about having find threaded depend on Tree behaviour. Even if we account for the issue pointed out by @sdustinh it means that method will behave differently depending on whether core Tree behaviour is used or not, that's not good IMO.

Anyone wanting the same parent field to be automatically used can easily override the findThreaded method in their own table.

mrothauer:
thank you @markstory for the infos about tesing.
I was surprised that the method find('threaded') was not part of the Tree-Behavior class, but part of the Table class. What would be a use-case for using find('threaded') without using a tree?
I will wait with the further implementation until the discussion has finished.

dereuromark:
OK, given the fact that this is a default, and that there are several ways to more easily set a different key for parent_id, I guess this introduction of magic here is not necessary after all.
I agree with @ADmad and @sdustinh then.

sdustinh:
It would be awesome for some way to configure a behavior finder to take precedence over a table finder. Then you could add a `findThreaded()` to the tree behavior and set a configuration option to tell the table to use that instead of the one provided by the table. Out of the scope of this PR, though.

ADmad:
> What would be a use-case for using find('threaded') without using a tree?

The Tree behaviour only implements MPTT. There are other tree structures like adjacency list which also use parent_id field, so findThreaded could be used with such an implementation.

dereuromark:
@sdustinh That goes in the direction of https://github.com/cakephp/cakephp/issues/11240 - a more explicit and less auto-magic way would allow for such merging and overwriting of finders probably. It will also help tools to provide proper IDE support for these finders.

PR #11665, opened 2018-01-26
Add initialize method to View::Cell

mosaxiv:
I think initialize method is  convenient to have it in the View::Cell.
initialize method is having consistency across the framework.

## use case

I think that I will use `initialize()` especially when there are multiple methods.

```php
    public function __construct(
        $request = null,
        $response = null,
        $eventManager = null,
        array $cellOptions = []) {

        parent::__construct($request, $response, $eventManager, $cellOptions);

        $this->loadModel('Banners');
    }

    public function top()
    {
        $banners = $this->Banners->find('top');
        //do something
    }

    public function side()
    {
        $banners = $this->Banners->find('side');
        //do something
    }

    public function footer()
    {
        $banners = $this->Banners->find('footer');
        //do something
    }
```

↓

```php
class BannerCell extends Cell
{
    public function initialize()
    {
        $this->loadModel('Banners');
    }

    public function top()
    {
        $banners = $this->Banners->find('top');
        //do something
    }

    public function side()
    {
        $banners = $this->Banners->find('side');
        //do something
    }

    public function footer()
    {
        $banners = $this->Banners->find('footer');
        //do something
    }
}
```


thinkingmedia:
This should be done as an event, which is consistent with how `beforeFilter` and all other callbacks are handled including table objects.

ADmad:
An event seems unnecessary when the handler is almost always going to be in the same class. I can't think of a use case where you would need an external listener for that event. We don't want to turn cells into full fledged controllers. So `initialize()` is fine.

dereuromark:
@thinkingmedia If it is close to the constructor there are no events used usually:
- https://github.com/cakephp/cakephp/blob/master/src/Controller/Component.php#L134-L158
- https://github.com/cakephp/cakephp/blob/master/src/View/Helper.php#L127

etc

markstory:
Sounds like a good idea to me. This will likely land in 3.6 though as it is a new feature.

mosaxiv:
Conflict....
Should I crate PR with 3.next branch?

markstory:
@mosaxiv I can resolve the conflicts.

markstory:
Rebased and merged into `3.next` in 408181598361983e16f9e614b7d0ae9bc19a7cf7

mosaxiv:
Thanks!

Issue #11634, opened 2018-01-15
3.6 deprecated warning 


saeideng:
This is a (multiple allowed):

* [ ] bug
* [x] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.6

1 )
```php
class PagesController extends AppController
{
  public function index(){
    $this->request->data();
  }

//returns 
ServerRequest::data() is deprecated. Use withData() and getData() or getParsedBody()
 instead. - ...vendor.../src/Controller/Controller.php, line: 594
 [...vendor.../src/Core/functions.php, line 305]
```
but this should returns 
```
.../src/Controller/PagesController.php, line: 32
```
changing  `$stackFrame` in the `function deprecationWarning($message, $stackFrame = 2)` can be solve this issue 

2 )

should update to use undeprecated response methods  
https://github.com/cakephp/cakephp/blob/308183b37d72040aff13402cf6a6da81851c6ea1/src/Error/ErrorHandler.php#L190

markstory:
You're right about the frame parameter being off 😊

saeideng:
also looks needs to update `stackFrame` in  the 'triggerWarning' 
https://github.com/cakephp/cakephp/blob/6c969fc32bc81eb49a9e76ab32313aac67a1d3d6/src/Core/functions.php#L252-L261

ADmad:
@saeideng Can you make a PR to fix that so that we can close this issue? 🙂

PR #11667, opened 2018-01-28
Use ResponseEmitter instead of deprecated Response::send().

ADmad:
Refs #11634.


markstory:
I'll get the tests fixed up. Looks like I merged poorly with master last time.

Issue #11650, opened 2018-01-22
TableRegistry not working after not being used for a while


nekkid:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.4
* Platform and Target: Apache 2.4, MariaDB (latest version), running on root-server with admin rights.

### What you did
I have the situation where I use the TableRegistry of CakePHP in a script that acts like a server. It first works, but if there was no trafic for around one day, the server still runs but once the TableRegistry is used, it produces the following error: Warning Error: Error while sending QUERY packet. PID=2536 in [..vendor\cakephp\src\Databse\Statement\MysqlStatement.php, line 39]

Googling for the error results in: MySQL have limitation for size of data

Which cannot be the case, since the query and the corresponding result is really small. I have the feeling that the CakePHP Mysql connection somehow times out internally.

Here is my coding for reporducing the error:
namespace WebsocketServer;
//load the cakePHP classes
require_once '../config/bootstrap.php';

use Cake\ORM\TableRegistry;

class Server {
    public function query() {
        //any data fetch mockup (the 
        $data = TableRegistry::get('Screens')->find()->all()->toArray();
    }
}

$server = new Server();
//querying works perfecly
$server->query();
//wait for a day
sleep(60*60*24);
//querying results in an error
$server->query();

### What happened
If the server is running for one day (doing nothing), once a query being run it produces the CakePHP-MySQL error. Warning Error: Error while sending QUERY packet. PID=2536 in [..vendor\cakephp\src\Databse\Statement\MysqlStatement.php, line 39]
I have a feeling the connection needs to be instantiated or refreshed.

### What you expected to happen
I would have expected the query that ran on the start of the server, runs the same way one day later.

dereuromark:
I fixed up your code example, please use codefencing for better readability.

dereuromark:
On a side note: Instead of a superlong running query I would instead recommend using crontab and a cronjob based (re-run every x hours/days) execution of a shell script.
It will also solve your issue and avoid problems when the script for some reason does get cut off the connection etc.

nekkid:
Hi,

thanks for that.
With the help of the comunity it turned out that the connection needs to be check before the TableRegistry queries for data. For this the ConnectionManager can be used.
If you don't see it as a bug this issue can be closed.
```
$con = ConnectionManager::get('default');
if(!$con->isConnected()){
    $con->connect();
}
```

[...]

PR #11668, opened 2018-01-28
Automatically reconnecting to the database when safe to do so

lorenzo:
This has been anf often requested feature for the database layer,
specially by people using the database in long running background tasks.

Instead of instructing them to figure out how to configure their database
to not drop connections, better just give them an extra hand by reconnecting
to the database whenever it is safe to do so.

That is, reconnection are attempted only on fresh commands to the database
outside of a transaction.

### DIFF
--- a/src/Database/Retry/ReconnectStrategy.php
+++ b/src/Database/Retry/ReconnectStrategy.php
@@ -1,0 +1,120 @@
+<?php
+/**
+ * CakePHP(tm) : Rapid Development Framework (https://cakephp.org)
+ * Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
+ *
+ * Licensed under The MIT License
+ * For full copyright and license information, please see the LICENSE.txt
+ * Redistributions of files must retain the above copyright notice.
+ *
+ * @copyright     Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
+ * @link          https://cakephp.org CakePHP(tm) Project
+ * @since         3.6.0
+ * @license       https://opensource.org/licenses/mit-license.php MIT License
+ */
+namespace Cake\Database\Retry;
+
+use Cake\Core\Retry\RetryStrategyInterface;
+use Cake\Database\Connection;
+use Exception;
+
+/**
+ * Makes sure the connection to the database is alive before authorizing
+ * the retry of an action.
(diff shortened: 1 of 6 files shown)


src/Database/Retry/ReconnectStrategy.php
```diff
@@ -0,0 +1,120 @@
...
+
+use Cake\Core\Retry\RetryStrategyInterface;
+use Cake\Database\Connection;
+use Exception;
+
+/**
+ * Makes sure the connection to the database is alive before authorizing
+ * the retry of an action.
+ */
+class ReconnectStrategy implements RetryStrategyInterface
```
markstory: Should this class be marked as internal?

tests/TestCase/Database/ConnectionTest.php
```diff
@@ -1215,4 +1219,66 @@ public function pushNestedTransactionState()
...
+            ->will($this->throwException(new Exception('server gone away')));
+
+        $newDriver->expects($this->at(1))->method('disconnect');
+        $newDriver->expects($this->at(2))->method('connect');
+        $newDriver->expects($this->at(3))
+            ->method('prepare')
+            ->will($this->returnValue($statement));
+
+        $res = $conn->query('SELECT 1');
+        $this->assertInstanceOf('Cake\Database\StatementInterface', $res);
```
markstory: You could use `StatementInterface::class` here.

markstory:
Nice work @lorenzo. I like the the decorator based approach.

src/Database/Retry/ReconnectStrategy.php
lorenzo: I’d like to eventually make this visible as a thing, for example it is useful for retrying http requests

lorenzo: Well, not this class. The CommandRetry is what I was speaking about.

it is sensible to mark this one as internal, I agree

dereuromark:
Refs 
- https://github.com/cakephp/cakephp/issues/3423
- https://github.com/cakephp/cakephp/issues/11650

lorenzo:
There is an unrelated phpstan failing check, what should we do about it?

ADmad:
@lorenzo Ignore it :).  I'll take care of it on `3.next` branch.

PR #11670, opened 2018-01-29
Allow user to specify a php.ini in ServerShell

josegonzalez:
This allows users to specify a custom php.ini for use with the built-in server shell, which is useful for local testing.

### DIFF
--- a/src/Shell/ServerShell.php
+++ b/src/Shell/ServerShell.php
@@ -43,41 +43,28 @@ class ServerShell extends Shell
      *
      * @var string
      */
-    protected $_host;
+    protected $_host = self::DEFAULT_HOST;
 
     /**
      * listen port
      *
      * @var int
      */
-    protected $_port;
+    protected $_port = self::DEFAULT_PORT;
 
     /**
      * document root
      *
      * @var string
      */
-    protected $_documentRoot;
+    protected $_documentRoot = WWW_ROOT;
 
     /**
      * ini path
      *
      * @var string
      */
-    protected $_iniPath;
-
-    /**
-     * Override initialize of the Shell
-     *
-     * @return void
-     */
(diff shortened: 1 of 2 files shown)


src/Shell/ServerShell.php
```diff
@@ -69,6 +76,7 @@ public function initialize()
         $this->_host = self::DEFAULT_HOST;
         $this->_port = self::DEFAULT_PORT;
         $this->_documentRoot = WWW_ROOT;
+        $this->_iniPath = '';
```
ADmad: This seems unnecessary, why not just set it to empty string when declaring the property?

src/Shell/ServerShell.php
```diff
@@ -100,6 +111,12 @@ public function startup()
         if (preg_match("/^([a-z]:)[\\\]+(.+)$/i", $this->_documentRoot, $m)) {
             $this->_documentRoot = $m[1] . '\\' . $m[2];
         }
+        if (substr($this->_iniPath, -1, 1) === DIRECTORY_SEPARATOR) {
```
ADmad: `$this->_iniPath = rtrim($this->_iniPath, DIRECTORY_SEPARATOR)` would be simpler.

src/Shell/ServerShell.php
josegonzalez: Followed the pattern for everything else. I can"fix" in a follow-up or.

src/Shell/ServerShell.php
josegonzalez: Ditto.

markstory:
@ADmad @josegonzalez I did some cleanup based on the feedback.

josegonzalez:
yey!

PR #11671, opened 2018-01-29
Ensure given type exists as array key.

ADmad:
This fixes the numerous `sort() expects parameter 1 to be array, null given` seen here https://travis-ci.org/cakephp/bake/jobs/334680609.


markstory:
Not sure what is up with PHPStan, but those variables are clearly defined in that context.

PR #11672, opened 2018-01-29
Fix error reported by phpstan.

ADmad:


markstory:
Nice work. I guess PHPstan assumed that anything in the `try` block might not exist in the finally?

ADmad:
I assume it might not be possible to ensure statically when an exception could be thrown and that the assignment is always executed in the `try` block. @ondrejmirtes could share more light on it 🙂.

ondrejmirtes:
Yep, you're right, current PHPStan analysis of try-catch blocks and exceptions is quite naive - it assumes that anything in `try` block can fail and thus jumping to `finally` immediately... anything assigned in `try` block might not be defined then... I aim to improve this in future versions, for example none of the internal PHP functions throw exceptions AFAIK so I could use that information for better code understanding. Also, the feature will require to properly annotate code with `@throws` annotations so if some `try-catch` won't be understood correctly, it will be possible to improve it with the right annotations :)

I'm gathering ideas around this under the [Exception tracking](https://github.com/phpstan/phpstan/milestone/11) milestone, but it's quite fresh.

ADmad:
Thanks for your input @ondrejmirtes.

PR #11675, opened 2018-01-30
Update cacert.pem

ADmad:


markstory:
Good job on remembering this :+1:

ADmad:
Relying on a package by an unknown vendor to get CA bundle can be risky.

highstrike:
its MIT license, maybe we can setup our own cake/cacert based on it, it has all the update scripts already there, all it needs is a server where it can run and push automatically to github

markstory:
Automatically updating the certfile is a better solution long term. Perhaps we could use our Jenkins server to do it for us?

CauanCabral:
Theres also the https://github.com/paragonie/certainty project for that.

highstrike:
looks solid enough, they are also the guys behind [sodium_compat](https://github.com/paragonie/sodium_compat) which has some pretty big projects [depending on it](https://packagist.org/packages/paragonie/sodium_compat/dependents) like codeigniter and joomla

PR #11678, opened 2018-01-31
Fix #11677 withStatus does not check status

### LINKED ISSUE #11677: withStatus does not check if supplied status is valid
This is a (multiple allowed):

* [x] bug
* [ ] enhancement

* CakePHP Version: 3.5.11.
* Platform and Target: nginx/1.10.3 (Ubuntu), mysql  Ver 15.1 Distrib 10.0.33-MariaDB, for debian-linux-gnu (x86_64) using readline 5.2, FriendsOfCake/Crud 5.3

### What you did
In any controller using Crud and methods mapped to the ViewAction, you can do `$this->Crud->execute()` and all is fine and well in the world. Place a `return` in front of the statement, and you get 502 Bad Gateway errors.
```
$this->Crud->mapAction('someWorkingMethod', 'Crud.View');
$this->Crud->mapAction('someBrokenMethod', 'Crud.View');

//working
public function someWorkingMethod() {
    $this->Crud->execute();
}

// crashing
public function someBrokenMethod() {
    return $this->Crud->execute();
}
```

### What happened
Through some digging, we found the statuscode in the response is null, gets transformed to 0 and nginx loses it. The statuscode being null originates from the ViewAction missing an index `api` in its defaultConfig. We encountered this problem after we updated to Crud 5.3, which now uses `$response->withStatus()` instead of `$response->statusCode()`. `withStatus()` does not check the provided status, which `statusCode` did.

We expected our methods to simply return the response, instead of completely crashing. The statuscode should never become null.

tuna-rootnet:
Add same checks as statusCode had to withStatus

Refs https://github.com/cakephp/cakephp/issues/11677

### DIFF
--- a/src/Http/Response.php
+++ b/src/Http/Response.php
@@ -925,14 +925,11 @@ class Response implements ResponseInterface
      */
     public function withStatus($code, $reasonPhrase = '')
     {
-        if ($code === null) {
-            return $this->_status;
-        }
         if (!isset($this->_statusCodes[$code])) {
             throw new InvalidArgumentException('Unknown status code');
         }
         $new = clone $this;
-        $new->_status = $code;
+        $new->_status = $code === null ? $this->_status : $code;
         if (empty($reasonPhrase) && isset($new->_statusCodes[$code])) {
             $reasonPhrase = $new->_statusCodes[$code];
         }
--- a/src/Http/Response.php
+++ b/src/Http/Response.php
@@ -925,6 +925,12 @@ class Response implements ResponseInterface
      */
     public function withStatus($code, $reasonPhrase = '')
     {
+        if ($code === null) {
+            return $this->_status;
+        }
+        if (!isset($this->_statusCodes[$code])) {
+            throw new InvalidArgumentException('Unknown status code');
+        }
         $new = clone $this;
         $new->_status = $code;
         if (empty($reasonPhrase) && isset($new->_statusCodes[$code])) {
--- a/src/Http/Response.php
+++ b/src/Http/Response.php
@@ -929,7 +929,7 @@ class Response implements ResponseInterface
             throw new InvalidArgumentException('Unknown status code');
         }
         $new = clone $this;
-        $new->_status = $code === null ? $this->_status : $code;
+        $new->_status = $code;
         if (empty($reasonPhrase) && isset($new->_statusCodes[$code])) {
             $reasonPhrase = $new->_statusCodes[$code];
         }


src/Http/Response.php
```diff
@@ -925,6 +925,12 @@ public function getStatusCode()
      */
     public function withStatus($code, $reasonPhrase = '')
     {
+        if ($code === null) {
+            return $this->_status;
```
dereuromark: This read part and $code being NULL is invalid.

src/Http/Response.php
```diff
@@ -925,6 +925,12 @@ public function getStatusCode()
...
+        }
+        if (!isset($this->_statusCodes[$code])) {
+            throw new InvalidArgumentException('Unknown status code');
```
dereuromark: Why is this necessary? It seems wrong the code that you can set any custom code, in the case not found it will just not set any reason phrase.

dereuromark:
If you want to validate your status code, you should do that on app level beforehand.
I think those changes here are not good :-1:

src/Http/Response.php
tuna-rootnet: I added it because the implementation of statusCode() explicitly checks if the provided code is known to Cake, and I image there are people relying on always having a valid response code being returned. To me, it doesn't matter if a reason phrase can be added or not, just that the responsecode is valid. But, since Cake already has this list of codes, it seemed like a good start.

src/Http/Response.php
tuna-rootnet: You're right; statusCode used it to set the default status code to the response. Changed it to set the default response code to $new.

src/Http/Response.php
```diff
@@ -925,8 +925,11 @@ public function getStatusCode()
      */
     public function withStatus($code, $reasonPhrase = '')
     {
+        if (!isset($this->_statusCodes[$code])) {
+            throw new InvalidArgumentException('Unknown status code');
+        }
         $new = clone $this;
-        $new->_status = $code;
+        $new->_status = $code === null ? $this->_status : $code;
```
ADmad: Due to the check and exception added above, the `$code === null` case will never occur, hence this check is not needed.

tuna-rootnet: @ADmad Haven't had my coffee yet... Removed the check, the check earlier in the method should suffice.

src/Http/Response.php
dereuromark: but what if an application uses custom codes? will that be a Problem then?

tuna-rootnet: It will be, just as it was before withStatus() was available. I have no problem with applications using their own statuscodes, but since CakePHP explicitly links to the RFC's for status codes and has its own internal list of valid codes and reason phrases, I'd assume it uses them as well.

I'm not against having people chose their own status code, but it should be a valid code; be that anyting between 100 and 599 or the list in Cake itself doesn't matter to me.

src/Http/Response.php
```diff
@@ -925,6 +925,9 @@ public function getStatusCode()
      */
     public function withStatus($code, $reasonPhrase = '')
     {
+        if (!isset($this->_statusCodes[$code])) {
+            throw new InvalidArgumentException('Unknown status code');
```
ADmad: You need to add `use InvalidArgumentException` at top of file.

ADmad:
@dereuromark The [ResponseInterface](https://github.com/php-fig/http-message/blob/master/src/ResponseInterface.php#L50) has `@throws \InvalidArgumentException For invalid status code arguments.`, so I don't see any problem in having the method check and throw an exception.

dereuromark:
All right.

src/Http/Response.php
tuna-rootnet: @ADmad It's already present in the file at rule 27.

ADmad: @tuna-rootnet Sorry, I missed that.

markstory: Can we the user what the invalid value was? And perhaps what the acceptable range is?

ADmad:
While I like the change I don't think throwing exception should be added in a bugfix release. Better to target this to 3.next.

PR #11679, opened 2018-02-01
Add method to generate secure random strings.

ADmad:
Generating secure random tokens is a pretty common requirement for various use cases like confirmation links, custom auth token, etc. so would be nice to have a method in core itself for it.

### DIFF
--- a/src/Utility/Security.php
+++ b/src/Utility/Security.php
@@ -142,7 +142,11 @@ class Security
      */
     public static function randomString($length)
     {
-        return bin2hex(Security::randomBytes($length / 2));
+        return substr(
+            bin2hex(Security::randomBytes(ceil($length / 2))),
+            0,
+            $length
+        );
     }
 
     /**
--- a/src/Utility/Security.php
+++ b/src/Utility/Security.php
@@ -134,6 +134,18 @@ class Security
     }
 
     /**
+     * Creates a secure random string.
+     *
+     * @param int $length String length
+     * @return string
+     * @since 3.6.0
+     */
+    public static function randomString($length)
+    {
+        return bin2hex(Security::randomBytes($length / 2));
+    }
+
+    /**
      * Like randomBytes() above, but not cryptographically secure.
      *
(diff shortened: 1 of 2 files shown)


src/Utility/Security.php
```diff
@@ -133,6 +133,18 @@ public static function randomBytes($length)
...
+    /**
+     * Creates a secure random string.
+     *
+     * @param int $length String length
+     * @return string
+     * @since 3.6.0
+     */
+    public static function randomString($length)
+    {
+        return bin2hex(Security::randomBytes($length / 2));
```
dereuromark: Doesnt the $length input here completely un-reflect the actual length returned?
If I say length 10, I would expect to get a string of length 6 (= 6 characters), so the length should be a substr() on the result of the bin2hex() before returning then IMO.

markstory: Should you use `base64` encoding instead?

markstory (commented):
I like the idea :+1:

ADmad:
@dereuromark I don't really understand what you are saying regarding the length.

@markstory I found a few examples where `bin2hex` seemed to be function of choice for converting a byte string to text string and hence chose that.

https://stackoverflow.com/a/18910943/1230763

https://github.com/thephpleague/oauth2-client/blob/aa2e3df188f0bfd87f7880cc880e906e99923580/src/Provider/AbstractProvider.php#L269

ADmad:
Should we have some reasonable default length?

markstory:
> Should we have some reasonable default length?

64 is probably a good place to start.

ADmad:
Base64 encoded string have characters like `/`, `+`, `=` which are not URL safe so an extra cleanup operation would be needed.

dereuromark:
Default lenght sounds like an Implementation detail to me.

inoas:
What about sha512in the random bytes and then cutting that off?

markstory:
@inoas Wouldn't that prevent long values from being generated as sha512 has a lower upper bound than MAX_INT.

PR #11680, opened 2018-02-01
Throw exception when invalid response status code is used.

ADmad:
Alternative to #11678.

### DIFF
--- a/src/Http/Response.php
+++ b/src/Http/Response.php
@@ -975,11 +975,19 @@ class Response implements ResponseInterface
      * @param int $code The code to set.
      * @param string $reasonPhrase The response reason phrase.
      * @return void
+     * @throws \InvalidArgumentException For invalid status code arguments.
      */
     protected function _setStatus($code, $reasonPhrase = '')
     {
+        if (!isset($this->_statusCodes[$code])) {
+            throw new InvalidArgumentException(sprintf(
+                'Invalid status code: %s. Use a valid HTTP status code in range 1xx - 5xx.',
+                $code
+            ));
+        }
+
         $this->_status = $code;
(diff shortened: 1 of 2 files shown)


src/Http/Response.php
```diff
@@ -975,11 +975,19 @@ public function withStatus($code, $reasonPhrase = '')
      * @param int $code The code to set.
      * @param string $reasonPhrase The response reason phrase.
      * @return void
+     * @throws \InvalidArgumentException For invalid status code arguments.
      */
     protected function _setStatus($code, $reasonPhrase = '')
     {
+        if (!isset($this->_statusCodes[$code])) {
+            throw new InvalidArgumentException(sprintf(
+                'Invalid status code: %s. Use a valid HTTP status code in range 1xx - 5xx.',
```
tuna-rootnet: [nagging mode]
If we tell users they must use a valid statuscode between 1xx and 5xx, the check should not be whether or not the supplied code is present in $this->_statusCodes, but if the code is in range 1xx and 5xx. That way, users can still have their own statuscodes like @dereuromark commented in #11678, but invalid codes such cannot be used. 
[/nagging]

ADmad: `$this->_statusCodes` contains list of all valid [HTTP status codes](http://www.iana.org/assignments/http-status-codes/http-status-codes.xhtml). This status code is used to generate the header of form `HTTP/<protocol> <status-code> <reason-phrase>` so any user defined would not be valid.

markstory: > That way, users can still have their own statuscodes like 

I do not want to support/allow this behavior. People should stick to the standard set of HTTP codes.

PR #11682, opened 2018-02-02
PHPUnit 7 compatibility

swiffer:
According to ChangeLog https://github.com/sebastianbergmann/phpunit/blob/7.0/ChangeLog-7.0.md

- The PHPUnit\Framework\BaseTestListener class has been removed (deprecated in PHPUnit 6.4)
- Scalar Type Declarations and Return Type Declarations are now used where possible (as a result, the API of PHPUnit\Framework\TestListener, for instance, has changed)

### DIFF
--- a/src/TestSuite/Fixture/FixtureInjector.php
+++ b/src/TestSuite/Fixture/FixtureInjector.php
@@ -17,7 +17,8 @@ namespace Cake\TestSuite\Fixture;
 loadPHPUnitAliases();
 
 use Cake\TestSuite\TestCase;
-use PHPUnit\Framework\BaseTestListener;
+use PHPUnit\Framework\TestListener;
+use PHPUnit\Framework\TestListenerDefaultImplementation;
 use PHPUnit\Framework\Test;
 use PHPUnit\Framework\TestSuite;
 
@@ -25,9 +26,10 @@ use PHPUnit\Framework\TestSuite;
  * Test listener used to inject a fixture manager in all tests that
  * are composed inside a Test Suite
  */
-class FixtureInjector extends BaseTestListener
+class FixtureInjector implements TestListener
 {
 
+    use TestListenerDefaultImplementation;
     /**
      * The instance of the fixture manager to use
      *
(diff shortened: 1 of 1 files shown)


src/TestSuite/Fixture/FixtureInjector.php
```diff
@@ -63,7 +65,7 @@ public function __construct(FixtureManager $manager)
      * @param \PHPUnit\Framework\TestSuite $suite The test suite
      * @return void
      */
-    public function startTestSuite(TestSuite $suite)
+    public function startTestSuite(TestSuite $suite): void
```
ADmad: CakePHP 3.x is PHP 5.6+, so you can't use return type typehints.

swiffer: ok, wasn't aware of that. Installed PHPUnit as suggested in Composer package and updated to 7.0.0 this morning. Fresh install leads to 7.0.0 as well.

```bash
$ composer require --dev phpunit/phpunit
Using version ^7.0 for phpunit/phpunit
./composer.json has been updated
```

src/TestSuite/Fixture/FixtureInjector.php
```diff
@@ -17,17 +17,19 @@
...
 
 /**
  * Test listener used to inject a fixture manager in all tests that
  * are composed inside a Test Suite
  */
-class FixtureInjector extends BaseTestListener
+class FixtureInjector implements TestListener
 {
 
+    use TestListenerDefaultImplementation;
```
markstory: Does this trait exist in older versions of PHPUnit?

swiffer: Exists since 6.4

robertpustulka: CakePHP 3.x needs to be compatible with PHPUnit 5.7 and PHP 5.6.

You probably need to update your composer file to require PHPUnit 6.x as the core [composer.json](https://github.com/cakephp/cakephp/blob/master/composer.json#L44).

swiffer: ok, sorry rushing here

Issue #10643, opened 2017-05-11
In 4.0 make Router fluent


lukewatts:
I think v4.0 would greatly benefit from a fluent interface for routing. So instead of the current Route with array:

```
Router::scope('/', function (RouteBuilder $routes) {
    $routes->connect('/', ['controller' => 'Pages', 'action' => 'display', 'home']);
}
```

Fluent (chained) routes:
```
use App\Controller\Pages;

Router::scope('/', function (RouteBuilder $routes) {
    $routes->connect('/', 'Pages::display')->view('home');
}
```

Benefits include:
- Better IDE support and auto-completion via IntelliSense.
- Less typing. Not so WET (We Enjoy Typing)
- More inline with conventions in every other large PHP MVC framework (Laravel/Lumen, Slim, Silex all have fluent Routes)
- It's just cooler :)

* [x] enhancement
* [x] feature-discussion (RFC)

* CakePHP Version:4.0.

markstory:
What would the `view()` method in the second example do?

As part of #10308 I'm going to be adding a `routes()` hook to the Application and deprecating all of the static route building methods on `Router`. This idea and that work together nicely, but I'd like to understand what fluent features you'd like to see.

lukewatts:
@markstory The `view()` should actually be `name()` for the named routes.

I mainly just want to see less arrays wherever possible in routing.

There could also be an optional param in connect for anything which has to be in an array (or to allow for gradual migration of options to chained methods), such as:
```
$routes->connect('/', 'Pages::display', ['something' => 'here'])->name('home');
```

markstory:
The `Pages::display` DSL needs some more thinking. A few questions that come me are:

* How do prefixes and plugins work into it? 
* What about passed arguments, or application level keys?
* How if at all does this DSL get used when making URLs?

slince:
I also think the route arguments are  too messy

jrbasso:
An option to eliminate the arrays and avoiding the DSL (which I'm not a big fan, specially when the features grow that start to becoming very hack and you need almost a compiler to parse it) is using a PHPUnit style. Something like:

```php
$route->for('/')
  ->useControllerAction('Pages', 'display')
  ->withName('home');
```

You can also use the same for specializing:
```php
$route->for('/magic')
  ->requestMethod(Request::METHOD_GET)
  ->useControllerAction('App\CustomDomain\Controller\Homepage', 'display')
  ->usingPlugin('Abc')
[...]

PR #11688, opened 2018-02-02
Implementing reverse routing syntax

burzum:
See these two links

* https://github.com/cakephp/cakephp/issues/10643#issuecomment-301652587
* https://github.com/cakephp/cakephp/issues/10689#issuecomment-304440165

### DIFF
--- a/src/Routing/RouteBuilder.php
+++ b/src/Routing/RouteBuilder.php
@@ -693,7 +693,7 @@ class RouteBuilder
      * The above route will only be matched for GET requests. POST requests will fail to match this route.
      *
      * @param string $route A string describing the template of the route
-     * @param array|string|null $defaults An array describing the default route parameters. These parameters will be used by default
+     * @param array|string $defaults An array describing the default route parameters. These parameters will be used by default
      *   and can supply routing parameters that are not dynamic. See above.
      * @param array $options An array matching the named elements in the route to regular expressions which that
      *   element should match. Also contains additional parameters such as which routed parameters should be
@@ -703,7 +703,7 @@ class RouteBuilder
      * @throws \InvalidArgumentException
      * @throws \BadMethodCallException
      */
-    public function connect($route, array $defaults = null, array $options = [])
+    public function connect($route, array $defaults = [], array $options = [])
     {
         if ($defaults === null) {
             $defaults = [];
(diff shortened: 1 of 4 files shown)


src/Routing/RouteBuilder.php
```diff
@@ -703,8 +703,12 @@ public function loadPlugin($name, $file = 'routes.php')
      * @throws \InvalidArgumentException
      * @throws \BadMethodCallException
      */
-    public function connect($route, array $defaults = [], array $options = [])
+    public function connect($route, array $defaults = null, array $options = [])
```
ADmad: Why change the default value?

src/Routing/Router.php
```diff
@@ -202,14 +202,59 @@ public static function getNamedExpressions()
...
+                        'controller' => $matches[3],
+                        'view' => $matches[4]
+                    ];
+                case 4:
+                    return [
+                        'plugin' => $matches[1],
+                        'prefix' => $matches[2],
+                        'controller' => $matches[3],
+                        'view' => $matches[4]
+                    ];
```
markstory: There should probably be an exception thrown for the default case.

tests/TestCase/Routing/RouterTest.php
```diff
@@ -3422,6 +3422,53 @@ public function testCreateRouteBuilder()
...
+            'controller' => 'Articles',
+            'view' => 'view',
+            'action' => 'index',
+            'plugin' => null,
+            '_matchedRoute' => '/admin/articles/view'
+
+        ];
+        $this->assertEquals($result, $expected);
+
+        Router::connect('/admin/blog/articles/view', 'Blog.Admin\Articles::view');
```
markstory: What about nested prefixes? Those have caused us troubles in other places.

markstory: Or all tokens separated by \ except the last one are the prefix, and the last one is the controller name.

burzum: @markstory like that?

image

markstory: Is `Foo\Bar\Admin` the prefix? Or the plugin name there?

burzum: Prefix? Just specific exactly how you would like to have it implemented? :) Can you just commit a test case for the syntax you want and I'll take care of the rest.

markstory:
@burzum I've done the changes I think needed to be done and pushed them up.

burzum:
@markstory looks good.

markstory:
I will merge this once I get the documentation updates done.

PR #11690, opened 2018-02-04
Hash matches strict compare numeric value

o0h:
## What I did

```php
$john = Hash::extract([
    ['value' => 0],
    ['value' => 'John']
], '{n}[value=John]');
return current($john);
```

## What happened
get the value as bellow.
```php
$john = ['value' => 0];
```
## What I expected to happen
like this.

```php
$john = ['value' => 'John'];
```

---
I think it's better that `Hash::_matches()`  handles `'0'` and `(string)some-value` as another value, when non numeric character token are given.
To regard some string values as '0' is counterintuitive for me ;(

### DIFF
--- a/src/Utility/Hash.php
+++ b/src/Utility/Hash.php
@@ -267,6 +267,8 @@ class Hash
                 $prop = $prop ? '1' : '0';
             } elseif ($isBool) {
                 $prop = $prop ? 'true' : 'false';
+            } elseif (is_numeric($prop)) {
+                $prop = (string)$prop;
             }
 
             // Pattern matches and other operators.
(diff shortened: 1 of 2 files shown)


dereuromark:
While this makes sense, I think it would be safer to go into 3.next (3.6), as this can alter behavior for a lot of existing code that might rely on this non-strict check. So while it would fix your issue, it might break a few others in the process.

o0h:
Thanks!
I think the patch is reasonable, but am also afraid of existing products in the world will be broken..
So, I agree this pr should not be merged as current version(3.5).

markstory:
@o0h You will need to rebase you changes onto 3.next.

src/Utility/Hash.php
```diff
@@ -267,15 +267,17 @@ protected static function _matches($data, $selector)
                 $prop = $prop ? '1' : '0';
             } elseif ($isBool) {
                 $prop = $prop ? 'true' : 'false';
+            } elseif (is_numeric($prop)) {
+                $prop = (string)$prop;
```
markstory: How will integer data be matched now?

o0h: Ex:
```php
$data = [
            ['value' => '1'],
            ['value' => 1],
            ['value' => 'string-value'],
        ];
$result = Hash::extract($data, '{n}[value=1]');
```
then
```php
$result = [
    ['value' => '3'],
    ['value' => 3],
]
```

---
When `$selector` is `'[value=3]'`, `$val` is given as string.
And `$prop` casting to string, so `$prop === $val` means `'3' === '3'`, returns true.

PR #11691, opened 2018-02-04
feat: allow modifying the Form schema class

josegonzalez:
I was working on a CMS and found it annoying to override the Schema class in use for a given form. This change allows the Form::__construct() to override `$this->_schemaClass`, making it easier to swap out the schema class in use.

The alternative is the following hack:

```
protected function _buildSchema(Schema $schema)
{
    if (count($schema->fields()) === 0) {
        $schema = new AnotherSchemaClass;
    }
}
```

### DIFF
--- a/src/Form/Form.php
+++ b/src/Form/Form.php
@@ -35,6 +35,12 @@ use Cake\Validation\Validator;
  */
 class Form
 {
+    /**
+     * Schema class.
+     *
+     * @var string
+     */
+    protected $_schemaClass = '\Cake\Form\Schema';
 
     /**
      * The schema used by this form.
@@ -70,7 +76,7 @@ class Form
     public function schema(Schema $schema = null)
     {
         if ($schema === null && empty($this->_schema)) {
-            $schema = $this->_buildSchema(new Schema());
+            $schema = $this->_buildSchema(new $this->_schemaClass);
         }
         if ($schema) {
             $this->_schema = $schema;
--- a/src/Form/Form.php
+++ b/src/Form/Form.php
@@ -14,6 +14,7 @@
  */
 namespace Cake\Form;
 
+use Cake\Form\Schema;
 use Cake\Validation\Validator;
 
 /**
@@ -40,7 +41,7 @@ class Form
      *
      * @var string
      */
-    protected $_schemaClass = '\Cake\Form\Schema';
+    protected $_schemaClass = Schema::class;
 
     /**
      * The schema used by this form.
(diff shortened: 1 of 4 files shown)


src/Form/Form.php
```diff
@@ -35,6 +35,12 @@
  */
 class Form
 {
+    /**
+     * Validator class.
```
ADmad: Schema class.

src/Form/Form.php
```diff
@@ -70,7 +76,8 @@ class Form
     public function schema(Schema $schema = null)
     {
         if ($schema === null && empty($this->_schema)) {
-            $schema = $this->_buildSchema(new Schema());
+            $schema = new $this->_schemaClass;
```
ADmad: Nitpick: Can directly do pass the new instance to `_buildSchema()` instead of assigning to a variable.

src/Form/Form.php
```diff
@@ -107,7 +118,7 @@ protected function _buildSchema(Schema $schema)
     public function validator(Validator $validator = null)
     {
         if ($validator === null && empty($this->_validator)) {
-            $validator = $this->_buildValidator(new Validator());
+            $validator = $this->_buildValidator(new $this->_validatorClass);
```
ADmad: This has been taken care of in `3.next` by making `Form` use `ValidatorAwareTrait`.

src/Form/Form.php
```diff
@@ -50,6 +54,13 @@ class Form
      */
     protected $_errors = [];
 
+    /**
+     * Validator class.
+     *
+     * @var string
+     */
+    protected $_validatorClass = '\Cake\Validation\Validator';
```
dereuromark: Should use ::class syntax both IMO.

dereuromark:
We should probably check the code for any other hardcoded `new ClassName()` syntax and allow at least some basic factory access to it.

src/Form/Form.php
josegonzalez: Can you do that in a class property?

markstory:
Looks good to me. It would ideal to have a test for the new behaviour.

src/Form/Form.php
```diff
@@ -35,6 +35,12 @@
  */
 class Form
 {
+    /**
+     * Schema class.
+     *
+     * @var string
+     */
+    protected $_schemaClass = '\Cake\Form\Schema';
```
dereuromark: ::class would be cleaner. And better for IDE discoverability.

josegonzalez:
Does someone want to volunteer to write a test? @burzum ? :D

ADmad:
Test added.

Issue #11687, opened 2018-02-02
RFC - Plugin Classes


markstory:
This is a (multiple allowed):

* [ ] bug
* [ ] enhancement
* [x] feature-discussion (RFC)

This is a formalization of #11564 that is aiming to get additional feedback and ideas before the API is finalized.

## Problem

Plugins have no central/single place to define their behavior around bootstrapping, routes and CLI commands. We rely on plugin users to copy the correct `Plugin::load()` call from plugin README files resulting in an error prone process.

With the creation of the `Application` classes the differences between applications and plugins has grown more visible. Plugins initialization has no easy way to leverage command classes, and aliased commands. Furthermore, it will be very difficult to amend plugins to use a dependency container in the future as plugins are created from snippet files and global statics.

## Goals of the Solution

The new plugin initialization should aim to solve the above problems, and have the following properties:

* Should work well with the `Application` class and be compatible with DI containers defined in the application, or by CakePHP in the future.
* Should avoid adding new global static features.
* Backwards compatibility with `Plugin` methods. This 'facade' class is a useful shortcut for accessing which plugins are loaded.
* Enable loading plugins with a single method call.
* Allow plugins to load configuration, routes, add console commands, and middleware automatically.
* Allow application developers to disable all of the above and handle those resource types manually.

## Plugin App Classes

Encapsulating the default behavior of a plugin can be supported by creating plugin 'application' classes. Plugin app classes can be added to the host application via a inside its `bootstrap()` hook.

Plugin classes can define the following hook methods:

* `bootstrap()` Load plugin configuration, constants and global event listeners.  Called after application bootstrapping is complete.
* `console(CommandCollection $commands)` Add console commands for the plugin.  Called after application commands are added to the CommandCollection.
* `middleware(MiddlewareStack $middleware)` Add middleware for the plugin.  Called after the application middleware hook is called.
* `routes(RouterBuilder $routes)` Define routes for the plugin. Called after the  application routes have been defined.

By implementing any of these hook methods, a plugin will have them automatically invoked by the application at the appropriate time.

## Loading Plugins

Plugins are added into the application during the application `bootstrap()`. You can load plugins by object, or classname:

```php
// In Application::bootstrap()

// By classname - Accepts all defaults.
$this->addPlugin(ExamplePlugin::class)

// By object, allows customization
$plugin = new OtherPlugin($this);
$this->addPlugin($plugin)
```

[...]

PR #11692, opened 2018-02-05
[RFC] Add application event hook.

robertpustulka:
Apps and plugins sometimes require configuring specific app runtime event listeners that hook into creation of models or controllers. This is usually being done in app's `bootstrap.php` file on a global `EventManager` instance. This PR adds a desiganted application hook for configuring event listeners.

Now instead of doing something like:
```php
//bootstrap.php
EventManager::instance()
    ->on(new MyEventListener())
    ->on(new OtherEventListener());
```
You can do:
```php
//Application.php
public function events(EventManagerInterface $eventManager)
{
    return $eventManager
        ->on(new MyEventListener())
        ->on(new OtherEventListener());
}
```

This PR also adds a `EventManagerInterface` which is something that in my mind has been missing for some time.

I'm looking forward for some feedback. After these changes get accepted I'll add some tests.

Refs https://github.com/cakephp/cakephp/issues/11687#issuecomment-362797267 and a discussion below.


src/Console/CommandRunner.php
```diff
@@ -70,6 +71,10 @@ public function __construct(ConsoleApplicationInterface $app, $root = 'cake')
             '--help' => 'help',
             '-h' => 'help',
         ];
+
+        if ($app instanceof EventDispatcherInterface) {
```
markstory: Shouldn't this also be the EventApplicationInterface?

src/Event/EventManager.php
```diff
@@ -22,7 +22,7 @@
  * can create multiple instances of this object to manage local events or keep a single instance
  * and pass it around to manage all events in your app.
  */
-class EventManager
+class EventManager implements EventManagerInterface
```
markstory: The other methods in this class could have their doc blocks changed to be `inheritDoc` now.

src/Event/Middleware/EventMiddleware.php
```diff
@@ -0,0 +1,74 @@
...
+     */
+    public function __construct(EventApplicationInterface $app, EventDispatcherInterface $dispatcher = null)
+    {
+        $this->app = $app;
+        if ($dispatcher === null) {
+            if ($app instanceof EventDispatcherInterface) {
+                throw new RuntimeException('Event dispatcher has not been provided.');
+            }
+            $dispatcher = $app;
+        }
```
markstory: Why is `$dispatcher` optional?

src/Event/Middleware/EventMiddleware.php
```diff
@@ -0,0 +1,74 @@
...
+     * @param \Psr\Http\Message\ServerRequestInterface $request Server request.
+     * @param \Psr\Http\Message\ResponseInterface $response Response.
+     * @param callable $next Callback to invoke the next middleware.
+     * @return \Psr\Http\Message\ResponseInterface
+     */
+    public function __invoke(ServerRequestInterface $request, ResponseInterface $response, callable $next)
+    {
+        $eventManager = $this->dispatcher->getEventManager();
+        $eventManager = $this->app->events($eventManager);
+        $this->dispatcher->setEventManager($eventManager);
```
markstory: Should events be bound after the application's bootstrap?

src/Console/CommandRunner.php
robertpustulka: This check is made in later method when the events() method is called.

src/Event/Middleware/EventMiddleware.php
robertpustulka: This falls back to the $app if it implements the interface. I wanted to avoid typehinting against the mixed type.

src/Event/Middleware/EventMiddleware.php
robertpustulka: I wasn’t sure where was the right place to call the events hook. Thought that a middleware might be a good place since routes are initialized in a similar way. Also having a middleware for that gives more control on how and when events are bound.

src/Event/EventManager.php
robertpustulka: Yes, I’m not sure if we should add more methods to the interface. Other ones looked like internal/tests related.

slince:
Great! I have added a simlar method in my Application.php

src/Event/Middleware/EventMiddleware.php
robertpustulka: Good point

src/Event/EventManagerInterface.php
```diff
@@ -0,0 +1,109 @@
...
+ * Redistributions of files must retain the above copyright notice.
+ *
+ * @copyright     Copyright (c) Cake Software Foundation, Inc. (http://cakefoundation.org)
+ * @link          http://cakephp.org CakePHP(tm) Project
+ * @since         3.6.0
+ * @license       http://www.opensource.org/licenses/mit-license.php MIT License
+ */
+namespace Cake\Event;
+
+interface EventManagerInterface
```
burzum: @robertpustulka have you seen this one? https://github.com/php-fig/fig-standards/blob/master/proposed/event-manager.md Scroll down to the interface for the manager.

robertpustulka: Yes I have. It’s in proposed for almost two years, I’m not sure if anyone’s interested in working on it.

robertpustulka:
The hooks are now executed after bootstrapping both in cli and web. The middleware has been removed.

Should we merge this before #11564?

markstory:
@robertpustulka It likely should merge before the new plugin classes. I can make sure that happens.

src/Console/CommandRunner.php
```diff
@@ -156,6 +174,22 @@ public function run(array $argv, ConsoleIo $io = null)
...
+     *
+     * Calls `bootstrap()` and `events()` if application implements `EventApplicationInterface`.
+     *
+     * @return void
+     */
+    protected function bootstrap()
+    {
+        $this->app->bootstrap();
+        if ($this->app instanceof EventApplicationInterface) {
+            $eventManager = $this->app->events($this->getEventManager());
```
markstory: Could this result in event listeners being dropped if the application's bootstrap() attaches events to the application manager? Shouldn't this be passing in the application's event manager back into events() hook?

robertpustulka: CommandRunner's and Server's event managers are set to the app's instance in `setApp` https://github.com/cakephp/cakephp/pull/11692/files#diff-5746a5b6845a932f92018ff514ae2000R110

markstory: I missed that. My bad.

robertpustulka:
It's ready to go I think.

PR #11703, opened 2018-02-08
Backported few Table class methods

bancer:
Partially backported initialize(), belongsTo(), hasOne(), hasMany() and belongsToMany(), addAssociations() methods from Table to Model.

### DIFF
--- a/lib/Cake/Model/Model.php
+++ b/lib/Cake/Model/Model.php
@@ -807,6 +807,46 @@ class Model extends CakeObject implements CakeEventListener {
 	}
 
 /**
+ * Setup multiple associations.
+ *
+ * It takes an array containing set of table names indexed by association type
+ * as argument:
+ *
+ * ```
+ * $this->Post->addAssociations([
+ *   'belongsTo' => [
+ *     'User' => ['className' => 'Users']
+ *   ],
+ *   'hasMany' => ['Comment'],
+ *   'belongsToMany' => ['Tag']
+ * ]);
+ * ```
+ *
+ * Each association type accepts multiple associations where the keys
+ * are the aliases, and the values are association config data. If numeric
+ * keys are used the values will be treated as association aliases.
+ *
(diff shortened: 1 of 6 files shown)


markstory:
I'm wary of adding features to a release that is in 'bugfix' mode. These new methods could cause strict errors with userland code.

bancer:
That would make the migration from 2.x to 3.x a bit easier.

dereuromark:
That's what the [Shim](https://github.com/dereuromark/cakephp-shim) plugin can help with IMO.

ADmad:
I too am not in favor of adding new methods/features to 2.x.

You can add these methods to your AppModel or use dereuromark's shim plugin.

bancer:
It is not easy to understand why are you so reluctant to backporting. It is quite depressing. These changes do not break any existing CakePHP interfaces or functionality and the risk that the userland code would have the same methods is minimal. But even if it would have them then it would be quite easy to rename there and the project with such code would only benefit from being closer to migrating to Cake 3. Shim plugin is great but it is more convenient to have non-breaking backports in the core code. Everybody could benefit from that but not only those who install the plugin. And maybe it would not seem so damn difficult to migrate from 2.x to 3.x.

ADmad:
> But even if it would have them then it would be quite easy to rename there...

That's easy for you to say but we can't risk breaking people's apps in bugfix releases. If you don't want use the Shim plugin just add those methods to your `AppModel`. That achieves what you want without the risk of causing problems to anyone else.

We have supported 2.x long enough and clearly communicated that it's in bugfix only mode now. We cannot keep on backporting 3.x features to 2.x and add to our maintenance overhead. As a group of volunteers contributing their spare time for free there's only so much we can do.

dereuromark:
So please PR against shim or use your own plugin base classes here.
Closing as bugfix mode for 2.x cannot extend to the proposed changes.

Issue #11697, opened 2018-02-05
WidgetRegistry is a misleading class/method name as the class is not a subclass of ObjectRegistry


raul338:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.5.11
* Platform and Target: linux / internal server - nginx

### What you did
Tried to unload DateWidget from FormHelper, I'm using a [3rd party Helpers](https://github.com/Holt59/cakephp3-bootstrap-helpers). I'm trying to use html5 native datetime pickers (or polyfills).
I expected WidgetRegistry to extend ObjectRegistry, but this was not the case.

```php
// in AppView.php
$this->loadHelper('Form', [
    'className' => 'Bootstrap.Form',
    'templates' => [
        // ....
    ],
]);
$FormHelper = $this->_helpers->get('Form');
$FormHelper->widgetRegistry()->unload('datetime');
```

### What happened
```php
Error: [Error] Call to undefined method Cake\View\Widget\WidgetRegistry::unload()
```

### What you expected to happen
Be able to unload a Helper widget

Right now the workaround is to call `$this->Form->text` and change the type via javascript, but its sloppy

markstory:
You can't unload a widget. You can remap or replace a widget with `WidgetRegistry::add()`.

raul338:
Why not? I think it should be like the others registry. If trying to use a non-loaded widget it should throw an exception.

Anyway, I'll be using an str_replace in a custom helper.

Thanks

markstory:
> Why not? I think it should be like the others registry. If trying to use a non-loaded widget it should throw an exception.

Because we've never had a use case for unloading widgets before. Why would you want a widget name to throw exceptions?

raul338:
> Why would you want a widget name to throw exceptions?

The same way like using helpers
```php
// in a view
$this->Custom->myVoid();
// Error: CustomHelper not found. Create ../src/View/Helper/CustomHelper.php
```
Sure it looks super-specific. But what confused me is that as a registry, it should extend [ObjectRegistry](https://api.cakephp.org/3.5/class-Cake.Core.ObjectRegistry.html)

markstory:
Why should it extend ObjectRegistry? We don't need most of the methods that class provides for Widgets.

raul338:
Because the name `WidgetRegistry` implies it

markstory:
[...]

PR #11704, opened 2018-02-09
Rename WidgetRegistry to WidgetLocator

raul338:
Refs #11697 

Updated the tests, I don't know if I should add tests for BC.
Should I do it in some custom helper test ?

### DIFF
--- a/src/View/Helper/FormHelper.php
+++ b/src/View/Helper/FormHelper.php
@@ -282,7 +282,10 @@ class FormHelper extends Helper
 
         parent::__construct($View, $config);
 
-        $this->widgetLocator($locator, $widgets);
+        if (!$locator) {
+            $locator = new WidgetLocator($this->templater(), $this->_View, $widgets);
+        }
+        $this->setWidgetLocator($locator);
         $this->_idPrefix = $this->getConfig('idPrefix');
     }
 
@@ -298,29 +301,37 @@ class FormHelper extends Helper
     {
         deprecationWarning('widgetRegistry is deprecated, use widgetLocator instead.');
 
-        return $this->widgetLocator($instance, $widgets);
+        if ($instance) {
+            $instance->add($widgets);
+            $this->setWidgetLocator($instance);
+        }
+
+        return $this->getWidgetLocator();
     }
 
     /**
-     * Set the widget registry the helper will use.
+     * Get the widget locator currently used by the helper.
      *
-     * @param \Cake\View\Widget\WidgetLocator|null $instance The locator instance to set.
-     * @param array $widgets An array of widgets
-     * @return \Cake\View\Widget\WidgetLocator
+     * @return \Cake\View\Widget\WidgetLocator Current locator instance
      * @since 3.6.0
      */
-    public function widgetLocator(WidgetLocator $instance = null, $widgets = [])
+    public function getWidgetLocator()
     {
-        if ($instance === null) {
-            if ($this->_locator === null) {
-                $this->_locator = new WidgetLocator($this->templater(), $this->_View, $widgets);
-            }
+        return $this->_locator;
+    }
 
-            return $this->_locator;
-        }
+    /**
+     * Set the widget locator the helper will use.
+     *
+     * @param \Cake\View\Widget\WidgetLocator $instance The locator instance to set.
+     * @return $this
+     * @since 3.6.0
+     */
+    public function setWidgetLocator(WidgetLocator $instance)
+    {
         $this->_locator = $instance;
 
-        return $this->_locator;
+        return $this;
     }
 
     /**
--- a/src/View/Helper/FormHelper.php
+++ b/src/View/Helper/FormHelper.php
@@ -24,7 +24,7 @@ use Cake\View\Form\ContextInterface;
 use Cake\View\Helper;
 use Cake\View\StringTemplateTrait;
 use Cake\View\View;
-use Cake\View\Widget\WidgetRegistry;
+use Cake\View\Widget\WidgetLocator;
 use DateTime;
 use RuntimeException;
 use Traversable;
@@ -217,11 +217,11 @@ class FormHelper extends Helper
     protected $_unlockedFields = [];
 
     /**
-     * Registry for input widgets.
+     * Locator for input widgets.
      *
-     * @var \Cake\View\Widget\WidgetRegistry
+     * @var \Cake\View\Widget\WidgetLocator
      */
-    protected $_registry;
+    protected $_locator;
 
     /**
      * Context for the current form.
@@ -260,11 +260,11 @@ class FormHelper extends Helper
      */
     public function __construct(View $View, array $config = [])
     {
-        $registry = null;
+        $locator = null;
         $widgets = $this->_defaultWidgets;
-        if (isset($config['registry'])) {
-            $registry = $config['registry'];
-            unset($config['registry']);
+        if (isset($config['locator'])) {
+            $locator = $config['locator'];
+            unset($config['locator']);
         }
         if (isset($config['widgets'])) {
             if (is_string($config['widgets'])) {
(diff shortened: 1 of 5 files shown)


src/View/Widget/WidgetLocator.php
```diff
@@ -36,7 +36,7 @@
  *
  * Widgets can ask for the current view by using the `_view` widget.
  */
-class WidgetRegistry
+class WidgetLocator
```
dereuromark: You need to keep the old class for BC.

raul338: There is an alias at the end of the file

raul338:
I don't know why codecov still says its failing. Should I do something else ?

src/View/Helper/FormHelper.php
```diff
@@ -286,19 +292,34 @@ public function __construct(View $View, array $config = [])
      * @param \Cake\View\Widget\WidgetRegistry|null $instance The registry instance to set.
      * @param array $widgets An array of widgets
      * @return \Cake\View\Widget\WidgetRegistry
+     * @deprecated 3.6.0 Use FormHelper::widgetLocator() instead.
      */
     public function widgetRegistry(WidgetRegistry $instance = null, $widgets = [])
+    {
+        trigger_error('widgetRegistry is deprecated, use widgetLocator instead.', E_USER_DEPRECATED);
```
saeideng: We have deprecationsWarnings() helper for these cases

saeideng (changes requested):

src/View/Helper/FormHelper.php
raul338: Didn't know if I had to use that, I thought it was for 4.0
I based on this deprecation warning
https://github.com/cakephp/cakephp/blob/8ca5054ad237650dec85bc69377cb64a95f59f8f/src/View/Helper/FormHelper.php#L383-L385

src/View/Widget/WidgetLocator.php
markstory: Aliases have to be in the original file so that if the old alias is used first it will work.

raul338: Done, used the same file as `Cake\Network\Http\Auth\Basic` => `Cake\Http\Client\Auth\Basic`

raul338:
Should I do anything else? Maybe squash the BC commits?

I don't know how to fix the codecov fail

markstory:
You can squash the commits if you like. Don't worry about codecov, it fails when new compatibility aliases are added.

src/View/Widget/WidgetRegistry.php
```diff
@@ -1,204 +1,5 @@
...
- * Licensed under The MIT License
- * For full copyright and license information, please see the LICENSE.txt
- * Redistributions of files must retain the above copyright notice.
- *
- * @copyright     Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
- * @link          https://cakephp.org CakePHP(tm) Project
- * @since         3.0.0
- * @license       https://opensource.org/licenses/mit-license.php MIT License
- */
-namespace Cake\View\Widget;
```
raul338: I should keep the header, right?

markstory: Yes.

raul338: Done

src/View/Widget/WidgetLocator.php
```diff
@@ -0,0 +1,204 @@
...
+ * CakePHP(tm) : Rapid Development Framework (https://cakephp.org)
+ * Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
+ *
+ * Licensed under The MIT License
+ * For full copyright and license information, please see the LICENSE.txt
+ * Redistributions of files must retain the above copyright notice.
+ *
+ * @copyright     Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
+ * @link          https://cakephp.org CakePHP(tm) Project
+ * @since         3.0.0
```
ADmad: 3.6.0

src/View/Helper/FormHelper.php
```diff
@@ -24,6 +24,7 @@
 use Cake\View\Helper;
 use Cake\View\StringTemplateTrait;
 use Cake\View\View;
+use Cake\View\Widget\WidgetLocator;
 use Cake\View\Widget\WidgetRegistry;
```
ADmad: I think this line can be removed now.

src/View/Helper/FormHelper.php
```diff
@@ -286,19 +292,34 @@ public function __construct(View $View, array $config = [])
...
+        return $this->widgetLocator($instance, $widgets);
+    }
+
+    /**
+     * Set the widget registry the helper will use.
+     *
+     * @param \Cake\View\Widget\WidgetLocator|null $instance The locator instance to set.
+     * @param array $widgets An array of widgets
+     * @return \Cake\View\Widget\WidgetLocator
+     */
```
ADmad: Please add `@since 3.6.0`

src/View/Helper/FormHelper.php
raul338: I leave it due a phpstan check, (a widgetLocator setter with typehint for BC)

ADmad: Ah sorry, it is needed for BC.

raul338:
I did the changes as requested

src/View/Helper/FormHelper.php
```diff
@@ -286,19 +292,35 @@ public function __construct(View $View, array $config = [])
...
      */
     public function widgetRegistry(WidgetRegistry $instance = null, $widgets = [])
+    {
+        deprecationWarning('widgetRegistry is deprecated, use widgetLocator instead.');
+
+        return $this->widgetLocator($instance, $widgets);
+    }
+
+    /**
+     * Set the widget registry the helper will use.
```
saeideng: also this method returns an object of  `Cake\View\Widget\WidgetLocator` 
we can not split this method to setter/getter IMO
because we have now `addWidget()`

raul338: Do you mean that both `widgetRegistry` and `widgetLocator` should return `Cake\View\Widget\WidgetLocator`?

I think it shuoldn't be a problem to have a getter/setter while keeping `addWidget`, but I'll wait for comments

saeideng: >Do you mean that both widgetRegistry and widgetLocator should return Cake\View\Widget\WidgetLocator?

no , i mean this method is not just setter 

https://book.cakephp.org/3.0/en/appendices/3-5-migration-guide.html#deprecated-combined-get-set-methods
https://book.cakephp.org/3.0/en/appendices/3-4-migration-guide.html#deprecated-combined-get-set-methods
we want to split methods into setter/getter methods where possible , like methods in above links 
but for this method  I'm not sure  @markstory

markstory: Splitting this method into `setWidgetLocator()` and `getWidgetLocator()` would match the pattern we're using elsewhere better. Having the separate methods will allow us to add return types more easily in 4.x.

raul338:
I made the split.
I didn't know what to do about the second parameter `$widgets` in the original `widgetRegistry` method. I removed it because I believe the widgets should be added to the Locator instead of using a "dual setter" (or calling `add` inside a setter).

src/View/Helper/FormHelper.php
```diff
@@ -217,11 +218,11 @@ class FormHelper extends Helper
...
 
     /**
-     * Registry for input widgets.
+     * Locator for input widgets.
      *
-     * @var \Cake\View\Widget\WidgetRegistry
+     * @var \Cake\View\Widget\WidgetLocator
      */
-    protected $_registry;
+    protected $_locator;
```
robertpustulka: Is that change BC?

dereuromark: It is highly unlikely to be relied on, but to be sure we can sure keep the former name here.
Note that we usually do not provide full BC on the internal protected methods (kind of private), especially in terms of low level infrastructure.
This would need a note in the migration guide then however.

markstory:
I will get the documentation updated and this merged in as it is looking good now.

PR #11705, opened 2018-02-09
Update BaseAuthenticate.php

allanmcarvalho:
This change suggested by me solves a need I have. I believe that many have this same need.
If the 'userIdentificationColumns' config exists with one or more values, this change allows the indentification to be performed with one or more columns of the database.
Of course the columns informed in the configuration should count distinct and unique values among them, which is not very difficult to implement. In my case I am authorizing login with username, email and the unique personal document number.

Works fine to me!

### DIFF
--- a/src/Auth/BaseAuthenticate.php
+++ b/src/Auth/BaseAuthenticate.php
@@ -147,12 +147,12 @@ abstract class BaseAuthenticate implements EventListenerInterface
         $config = $this->_config;
         $table = TableRegistry::get($config['userModel']);
 
-        if(empty($config['userIdentificationColumns'])) {
+        if (empty($config['userIdentificationColumns'])) {
             $config['userIdentificationColumns'] = $config['fields']['username'];
         }
-        if(is_array($config['userIdentificationColumns'])) {
+        if (is_array($config['userIdentificationColumns'])) {
             foreach ($config['userIdentificationColumns'] as $userIdentificationColumn) {
-                if(!empty($options['conditions']['OR'])  and is_array($options['conditions']['OR'])) {
+                if (!empty($options['conditions']['OR']) and is_array($options['conditions']['OR'])) {
                     $options['conditions']['OR'] += [$table->aliasField($userIdentificationColumn) => $username];
                 } else {
                     $options['conditions']['OR'] = [$table->aliasField($userIdentificationColumn) => $username];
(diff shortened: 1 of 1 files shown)


dereuromark:
Isn't this exactly what existing adapter extensions such as [MultiColumn](https://github.com/dereuromark/cakephp-tools/blob/master/src/Auth/MultiColumnAuthenticate.php) provide?
You seem to want to force an extension into a base class here. Whereas the linked one for example states it not only more clearly and descriptive, it also makes this a deliberate choice rather than a hidden config value.
So since the framework is this powerful to allow customization as own classes, it is better to follow SOLID here (especially open for extension and closed for modification) on this low adapter level.

allanmcarvalho:
well, I may even agree that the way I implemented my idea is not the best and the most correct. But here, I've used this (more precisely create an Authentication class that implements this my suggestion). I suggested this change because I believe that many people could take advantage of this feature, that is, it would be useful to have it natively. See big companies like Facebook, Twitter, Instagran and some others. They allow login with several types of identifiers, as long as they are unique. Like phone, email and others.

dereuromark:
> I believe that many people could take advantage of this feature

Yes, and that all already exists in plugin extensions for 10+ years now :) You just have to use it, it's free.

allanmcarvalho:
@dereuromark  Yes, plugins and tools for CakePHP have to be lost sight of on the internet. And many of them are great. I am writing here a resource that I faithfully believe to be worthy of being a CORE resource

allanmcarvalho:
But this is only my opinion!

allanmcarvalho:
For example ... You are a guy who gets good ideas and implements them in a very compact and efficient way. I've already used some of your plugins. If this your MultiColumnAuthenticate class is stable, why not implement it in Cake's CORE?

dereuromark:
Look at the cyclic complexity (tools can calculate that) of your +16 −4 code change.
There are 0 tests added, and even if so, there are tons of ways this base adapter due to the added complexity can now have bugs and other failures for this specific use case, and therefore makes all inheriting adapters inherit these issues and the overall complexity.

> why not implement it in Cake's CORE?

While some plugin code can easily be moved into core, why should it?
It would need to be used by > 80% of users to justify that, otherwise what is the harm of having it an optional composer component to add in? This way the core stays lean and the ecosystem a healthy and additive collection of tools.
Also, there might be 10 different flavors of how this multi column might be needed, and then you start the same config process again in core to allow for all use cases, rather than just making your optional component adjusted to what you need and then hook it in.

Please see the bigger picture here: Core is something essential, rest is optional (and of course often very useful) additions.

allanmcarvalho:
I'm without arguments. :)
Well, it makes sense, although it is used by many people, it may not be for the majority.

dereuromark:
Exactly. That said: Your contributions - especially in plugin form - are most welcome.
Once established, please also share them with the whole community using e.g. [awesome-cakephp](https://github.com/FriendsOfCake/awesome-cakephp). This way everyone can use, contribute these useful additions.
Closing then.

Issue #11698, opened 2018-02-06
New Link header Helper addition


mosaxiv:
This is a (multiple allowed):

* [ ] bug
* [x] enhancement
* [x] feature-discussion (RFC)

Create a [Link header](https://tools.ietf.org/html/rfc5988) 

```
Link: <'https://book.cakephp.org?page=1>; rel="first"
Link: <'https://book.cakephp.org?page=4>; rel="prev"
Link: <'https://book.cakephp.org?page=6>; rel="next"
Link: <'https://book.cakephp.org?page=10>; rel="last"
```

Easily create with withAddLink() method.

```php
$this->response->withAddLink('https://book.cakephp.org?page=1', ['rel' => 'first'])
    ->withAddLink('https://book.cakephp.org?page=4', ['rel' => 'prev'])
    ->withAddLink('https://book.cakephp.org?page=6', ['rel' => 'next'])
    ->withAddLink('https://book.cakephp.org?page=10', ['rel' => 'last'])
```

routing array
```php
$this->response->withAddLink(['controller' => 'Posts', 'action' => 'view', 5], ['rel' => 'first'])
```

For example Link header is used in [Githu API](https://developer.github.com/v3/#link-header).

markstory:
The PSR7 interface uses a `withAddedXyz()` in other methods. Might be good to follow that pattern here.

saeideng:
PR open.

PR #11706, opened 2018-02-10
Add withAddedLink()

mosaxiv:
https://github.com/cakephp/cakephp/issues/11698

### DIFF
--- a/src/Http/Response.php
+++ b/src/Http/Response.php
@@ -2003,6 +2003,7 @@ class Response implements ResponseInterface
      * @param string $url link-value
      * @param array $options link-param
      * @return static
+     * @since 3.6.0
      */
     public function withAddedLink($url, $options = [])
     {
--- a/src/Http/Response.php
+++ b/src/Http/Response.php
@@ -22,6 +22,7 @@ use Cake\Http\Cookie\CookieInterface;
 use Cake\Http\CorsBuilder;
 use Cake\Http\Exception\NotFoundException;
 use Cake\Log\Log;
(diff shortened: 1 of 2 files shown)


src/Http/Response.php
```diff
@@ -1983,6 +1984,44 @@ public function withLength($bytes)
...
+     * Link: </posts/view>; rel="next"
+     * ```
+     *
+     * @param string|array $url link-value
+     * @param array $options link-param
+     * @return static
+     */
+    public function withAddedLink($url, $options = [])
+    {
+        $url = '<' . Router::url($url) . '>';
```
markstory: Would be preferable to only accept string URLs in this method. Right now the options to `Router::url()` cannot be used.

src/Http/Response.php
```diff
@@ -1983,6 +1983,42 @@ public function withLength($bytes)
...
+     * ```
+     *
+     * Will generate:
+     *
+     * ```
+     * Link: <http://example.com?page=1>; rel="prev"
+     * Link: <http://example.com?page=3>; rel="next"
+     * ```
+     *
+     * @param string $url link-value
```
ADmad: Can we have a better descriptive statement instead of `link-value` please? :)  Same for options param.

ADmad: And `@since 3.6.0` tag.

mosaxiv: hmmm :anguished:
Sorry, I can't think of good ideas...
If revision is required, I want you to help me.

PR #11708, opened 2018-02-10
Defect 11686

T0mats0:
ObjectRegistry::normalizedArray normalizes the array the first time. When passing an array already normalized it continues trying to normalize producing a nested effect. This should resolve the issue.


markstory:
Thanks!

Issue #11533, opened 2017-12-12
New shell command forces plugin loading


dereuromark:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: latest dev-master

### What you did
Using existing 3.x apps that have a Console Shell and a plugin option, e.g.

    bin/cake upgrade all /path/ -p FooBar 

### What happened
```php
        if (!empty($this->params['plugin']) && !Plugin::loaded($this->params['plugin'])) {
            Plugin::load($this->params['plugin']);
        }
```
Forces a plugin load even though this was never indented to happen. Similar to bake and its plugin option this can very well also just be some other config or for external use.

Also the https://github.com/cakephp/upgrade is currently broken because of this: https://github.com/cakephp/upgrade/blob/master/src/Shell/Task/BaseTask.php#L37
It cannot use plugin option anymore, or dies with 

> Error: [Cake\Core\Exception\MissingPluginException] Plugin FooBar could not be found

### What you expected to happen
We shouldn't load Plugins here automatically as this is new unexpected behavior, or at least we need to find a way to disable this (or to be BC a way to enable it).

markstory:
Why is loading the plugin a problem?

markstory:
@dereuromark ping.

dereuromark:
because those do not exist in that application.
Those are the plugins of the target application.
The param name plugin itself does not warrant force loading plugins as in some scopes this is used for sth else. In this case only for path building.

markstory:
Ok. Perhaps `Shell` should catch the `MissingPluginException` and log a warning instead of dying?

dereuromark:
I think even that is too much. Lots of shells can have a plugin param and use it for doing sth else than requiring an actual plugin present.
This part looks not bc.

Instead only the shells that need it should load plugins from this.

markstory:
The line to load a plugin if the `--plugin` parameter is used was added almost 5 years ago though, in c727444880b5fc7e63493cc46fec6063e7b347f7. I'm not sure it is wise to remove it now.

dereuromark:
Interesting that shells worked so far with it. Is loading now more strict?

markstory:
Not sure. It may have become more strict as we make things stricter in general.

dereuromark:
PR open.

PR #11709, opened 2018-02-11
Remove bad plugin load magic

dereuromark:
Resolves https://github.com/cakephp/cakephp/issues/11533
Follows the plugin enhancements of https://github.com/cakephp/cakephp/pull/11564 

Makes it more explicit and less bad magic to load and use plugins.

This shouldn't even be needed anymore, as the bootstrapping process loads all wanted plugins anyway.
We should not allow hidden lazy load of plugins - especially not if that compromises the config to be used for anything else (e.g. generating sth for a plugin without loading it, as the upgrade tool does).

Any downsides of this? Does it suffice to have this outlined in the 3.6 migration guide?

### DIFF
--- a/src/Console/Shell.php
+++ b/src/Console/Shell.php
@@ -495,9 +495,6 @@ class Shell
             $this->params = array_merge($this->params, $extra);
         }
         $this->_setOutputLevel();
-        if (!empty($this->params['plugin']) && !Plugin::loaded($this->params['plugin'])) {
-            Plugin::load($this->params['plugin']);
-        }
         $this->command = $command;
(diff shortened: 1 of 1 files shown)


markstory (approved):
I think this is fine with a note in the migration guide.  This change is inline with reducing the amount of opaque automatic behavior in the framework. This code is also incompatible with the new plugin loading strategy we're working on.

saeideng:
unused namespace  
https://github.com/cakephp/cakephp/blob/e52f8c6b27b29fd3bacf73eeb7485b0b22eff273/src/Console/Shell.php#L20

PR #11711, opened 2018-02-12
Fix incorrect annotations.

ADmad:
@markstory @lorenzo What would be the proper way to avoid having to ignore the `Method PDO::lastInsertId() invoked with 2 parameters, 0-1 required`? It's caused by following line:

https://github.com/cakephp/cakephp/blob/a6c47f15e2dfbb062fcc547d6eecd0a944dc8b17/src/Database/Driver.php#L288

### DIFF
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -26,6 +26,7 @@ parameters:
         - '#Call to an undefined method object::__toString\(\)#'
         - '#Call to an undefined method object::toArray\(\)#'
         - '#Call to an undefined method object::__debugInfo\(\)#'
+        - '#Method PDO::lastInsertId\(\) invoked with 2 parameters, 0-1 required#'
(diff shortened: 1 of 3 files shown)


markstory:
> What would be the proper way to avoid having to ignore the

No idea really.

ADmad:
May in 3.6 (or 4.0) drop the second argument for `Driver::lastInsertId()`? I highly doubt anyone is going to extend it and use a non PDO connection for it.

markstory:
You're probably right that it won't be used for Non-PDO sources.

ADmad:
Let me know if making the change in 3.6 is fine and I'll make a PR.

markstory:
@ADmad We still have time to deprecate features in 3.6

ADmad:
Aye, I'll make a PR for 3.6 after this is merged.

PR #11713, opened 2018-02-13
3.next db driver

ADmad:
The `Driver` class has multiple methods using `$this->_connection` but does not declare any `$_connection` property. So theoretically the child classes can set `$_connection` to an object of any class but practically I don't think it will be used for any other type besides `PDO`.

So I have added `Driver::$_connection` declared as type `PDO` and moved all methods of `PDODriverTrait` to `Driver` itself. Also deprecated second method of `Driver::lastInsertId()`.


lorenzo:
I really don’t like this, what’s the motivation behind the change?

The Driver class is abstract in the hope that one day we can implement non-PDO drivers

ADmad:
@lorenzo It started with phpstan reporting error due with usage of `$this->_connection` in `Driver` without declaring a type for it. Based on the usages I found all methods called on `$this->_connection` has same signature as `PDO` except for `lastInsertId()`.

If we want to keep the ability to implement a non-PDO drivers in future wouldn't it be better to add a `DriverInterface` instead?

markstory:
@lorenzo Wouldn't an interface be a better way to afford non-PDO drivers than an abstract class? The abstract class and trait implementation we have right now is harder to understand than an interface and base class would be.

ADmad:
I have added `DriverInterface`.

ADmad:
I have added back `PDODriverTrait` and marked it deprecated.

ADmad:
@lorenzo Is this ok now? The addition of `DriverInterface` keeps open the possibility to implement non-PDO drivers.

src/Database/DriverInterface.php
```diff
@@ -0,0 +1,260 @@
...
+     * Disconnects from database server.
+     *
+     * @return void
+     */
+    public function disconnect();
+
+    /**
+     * Returns correct connection resource or object that is internally used.
+     *
+     * @return mixed Connection object used internally.
```
dereuromark: Can we be more specific, e.g. object or some interface?

src/Database/DriverInterface.php
```diff
@@ -0,0 +1,260 @@
...
+     * @return string
+     */
+    public function rollbackSavePointSQL($name);
+
+    /**
+     * Get the SQL for disabling foreign keys.
+     *
+     * @return string
+     */
+    public function disableForeignKeySQL();
```
dereuromark: Should those be Sql instead of acronym uppercase SQL etc as per our convention everywhere?
It is BC either way, as casing is not sensitive.

PS: disable sounds like void, should this be `get...`? Same above for save, enable etc.

src/Database/DriverInterface.php
```diff
@@ -0,0 +1,260 @@
...
+     * and the second one the compiled SQL.
+     */
+    public function compileQuery(Query $query, ValueBinder $generator);
+
+    /**
+     * Returns an instance of a QueryCompiler.
+     *
+     * @return \Cake\Database\QueryCompiler
+     */
+    public function newCompiler();
```
dereuromark: I would like getCompiler() more than newCompiler(), as for a normal internal factory call.

src/Database/DriverInterface.php
ADmad: No we can't, since the whole premise of this interface is the ability to use a non PDO connection, which at this point could be anything :)

dereuromark: not int^^ at least object I believe.

src/Database/DriverInterface.php
markstory: Yeah this method is in use already.

chinpei215:
I don't think adding DriverInterface is a nice idea for the following reasons:
- The DriverInterface type is never used in the core. And if we start using the interface, it will cause type declaration mismatch errors in users' applications.
- Our BC policy is not suitable for interface-based programming. We often add new methods to some classes in some bugfix/minor releases. But we don't add new methods to any interfaces until major releases. DriverInterface would become useless like EntitiyInterface.

I think the first implementation is better than now. And if we want to support non-PDO drivers, why not using a simpler class hierarchy like the following?

* Driver
  * PdoDriver
    * Mysql
    * Postgres
    * Sqlite
    * Sqlserver

ADmad:
> The DriverInterface type is never used in the core. And if we start using the interface, it will cause type declaration mismatch errors in users' applications.

The `Driver` class now implements `DriverInterface`. The signature of the interface methods is identical to those of existing `Driver` class so I don't see how there would be type declaration errors.

> Our BC policy is not suitable for interface-based programming. We often add new methods to some classes in some bugfix/minor releases. But we don't add new methods to any interfaces until major releases. DriverInterface would become useless like EntitiyInterface.

Adding new methods in bugfix releases has been a mistake and something to avoid in future. There have been many new interfaces added in 3.5/3.6 e.g. `CommandCollectionAwareInterface`, `EventApplicationInterface`, `EventManagerInterface` etc. We do plan to improve our versioning process and bump up versions more frequently to avoid issues like we have for `EntityInterface`.

> And if we want to support non-PDO drivers, why not using a simpler class hierarchy like the following?

I am fine with adding a `PdoDriver` class but keeping `Driver` in it's current state won't solve the issue that it uses `$this->_connection` without having a type defined for `$_connection`. Fixing that would require the methods using `$this->_connection` to be made abstract but that can't be done as it would cause backwards incompatibility.

robertpustulka:
3.6 will be the last 3.x release and with 4.0 we plan to update our interfaces with new methods and change method signatures where possible to leverage interfaces instead of concrete and abstract/base classes. So I wouldn’t worry about this not being used in the code as this is going to change soon. I also think that adding more interfaces in 3.6 is a good thing as it leaves us less work for 4.0 and lets us evaluate things earlier.

markstory:
> Our BC policy is not suitable for interface-based programming. We often add new methods to some classes in some bugfix/minor releases. 

Perhaps we should revisit this decision and allow new methods to be added to interfaces in minor (4.x.0) releases in the future. I agree that right now our policies make interfaces a bit clunky to work with.

chinpei215:
@ADmad @robertpustulka Our `Type` classes directly depend on `Driver`. If we leave those signatures as they are, people cannot create any drivers witch implement `DriverInterface` but not extend `Driver`. If we change those signatures, it causes type declaration errors. We wouldn't do that at least in 3.x. So I think adding `DriverInterface` in 3.x is meaningless.

@markstory I opened #11735.

ADmad:
@chinpei215 Yes the typehints needs to be changed from `Driver` to `DriverInterface` but can't be done in 3.x as they would cause type errors. Even though we can't change the typehints, 3.6 is the right time to add the interfaces as it leaves us less work for 4.0 as @robertpustulka said.

lorenzo:
I'm ok with this new implementation. If we need to add methods to the drivers we can do it via adding more interfaces to them, instead of more methods to the interface.

PR #7332, opened 2015-08-31
Initial support of 'enum' types in MySQL

rocher:
Hi,

Here you have a small set of changes to support  `enum`  types in MySQL.

Don't know if these are enough, but at least my problem is solved: I was unable to create a Fixture with an `$import` from a table with `enum` columns. Now it's possible and all works fine.

Please let me know how can I test it in order to be integrated in the master branch (hope nothing is broken), or what additional requirements this patch should have. This is my first contribution, so I don't know very well your development-test-integration cycle. Thanks.

Hope it helps!
-- _Francesc_

PR #11719, opened 2018-02-14
Add enum column type support to schema functions

alphp:
The command "Console/cake schema generate/create" does not work with Enum type fields by default. You will always get NOTICE's while doing so, and won't be able to restore them.

Based on: https://github.com/matheuscmpm/cakephp-schema-enum

### DIFF
--- a/lib/Cake/Model/CakeSchema.php
+++ b/lib/Cake/Model/CakeSchema.php
@@ -624,10 +624,16 @@ class CakeSchema extends CakeObject {
 			if ($Obj->primaryKey === $name && !$hasPrimaryAlready && !isset($value['key'])) {
 				$value['key'] = 'primary';
 			}
-			if (!isset($db->columns[$value['type']])) {
+			$isEnum = (strpos($value['type'], 'enum') !== false);
+			if (!isset($db->columns[$value['type']]) && !$isEnum) {
 				trigger_error(__d('cake_dev', 'Schema generation error: invalid column type %s for %s.%s does not exist in DBO', $value['type'], $Obj->name, $name), E_USER_NOTICE);
 				continue;
 			} else {
+				if ($isEnum) {
+					$length = trim(str_replace(')', '', str_replace('enum(', '', $value['type'])));
+					$value['type'] = 'enum';
+					$value['length'] = $length;
+				}
 				$defaultCol = $db->columns[$value['type']];
 				if (isset($defaultCol['limit']) && $defaultCol['limit'] == $value['length']) {
 					unset($value['length']);
--- a/lib/Cake/Model/Datasource/Database/Mysql.php
+++ b/lib/Cake/Model/Datasource/Database/Mysql.php
@@ -131,7 +131,8 @@ class Mysql extends DboSource {
 		'time' => array('name' => 'time', 'format' => 'H:i:s', 'formatter' => 'date'),
 		'date' => array('name' => 'date', 'format' => 'Y-m-d', 'formatter' => 'date'),
 		'binary' => array('name' => 'blob'),
-		'boolean' => array('name' => 'tinyint', 'limit' => '1')
+		'boolean' => array('name' => 'tinyint', 'limit' => '1'),
+		'enum' => array('name' => 'enum', 'limit' => '512')
 	);
 
 /**


dereuromark:
Refs https://github.com/cakephp/cakephp/pull/7332

alphp:
PostgreSQL enum type: https://www.postgresql.org/docs/9.1/static/datatype-enum.html
Oracle enum type: https://docs.oracle.com/cd/E17952_01/mysql-5.0-en/enum.html
MySQL/MariaDB enum type: https://dev.mysql.com/doc/refman/5.7/en/enum.html

markstory:
This seems like an incomplete implementation of `enum` support that doesn't include tests. I'm very reluctant to merge this kind of change into 2.x which is in bugfix only mode. 2.x has never supported `enum` columns so I'm not sure how this counts as a bugfix to an existing feature.

alphp:
And then these error messages are just a feature of CakePHP 2?
```
2018-02-15 00:01:44 Debug: Schema generation error: invalid column type enum('Nivel Ã“ptimo','Completado satisfactoriamente','Completado insatisfactoriamente','Incompleto') for ActionPlan.rating does not exist in DBO in [C:\usr\0xFF\angulas\app\Lib\Model\CakeSchema.php, line 628]
2018-02-15 00:01:44 Debug: Schema generation error: invalid column type enum('Completado satisfactoriamente','Completado insatisfactoriamente','Incompleto') for ActionPlanItem.rating does not exist in DBO in [C:\usr\0xFF\angulas\app\Lib\Model\CakeSchema.php, line 628]
2018-02-15 00:01:44 Debug: Schema generation error: invalid column type enum('proyectos','valores','competencias','bolos') for Item.type does not exist in DBO in [C:\usr\0xFF\angulas\app\Lib\Model\CakeSchema.php, line 628]
2018-02-15 00:01:44 Debug: Schema generation error: invalid column type enum('responsable','orientador') for Orientator.rol does not exist in DBO in [C:\usr\0xFF\angulas\app\Lib\Model\CakeSchema.php, line 628]
2018-02-15 00:01:44 Debug: Schema generation error: invalid column type enum('admin','rrhh','responsable','guiado') for User.type does not exist in DBO in [C:\usr\0xFF\angulas\app\Lib\Model\CakeSchema.php, line 628]
```
If there was no solution or if it was a complex and extensive, I would understand that even minimal support was not implemented.
However, a single line added to Cake/Model/Datasource/Database/MySQL.php and six in Cake/Model/CakeSchema.php support the scheme with ENUM in MySQL.

There are already plugins to support the ENUM in the forms, so it would not be necessary to add support for the forms and you can always add the warning that the only support is for the maintenance of the scheme, which in my opinion is the most important and complex maintain.

alphp:
By the way, the existing feature is "Schema Shell" ... but if you do not want to continue maintaining it ...

We could argue that support for the SET type, or other types of other databases, will also be requested. But the truth is that this PR solves the support for the ENUM type to the "Schema Shell" ... because let's be honest the complete support for all the data types of all the databases will never be achieved, but if someone uses a type of data in your CakePHP projects and add the necessary code for your support. Why should that support be relegated to something external to CakePHP?

josegonzalez:
A few notes:

### This is a new feature, not a bugfix

The error message is actually a generic one:

```
trigger_error(__d('cake_dev', 'Schema generation error: invalid column type %s for %s.%s does not exist in DBO', $value['type'], $Obj->name, $name), E_USER_NOTICE);
```

From: https://github.com/cakephp/cakephp/blob/f1db01395b89e5b43cb3f330ba25a51a3192ecc4/lib/Cake/Model/CakeSchema.php#L627-L629

We've traditionally not supported enums for a few reasons, but as CakePHP 2 has never had this functionality, I'd consider this a new feature.

### The implementation is incomplete

When we say incomplete, we mean that your pull request does not feature is not implemented for each datastore in the CakePHP 2.x core. Even beyond the fact that not all - sqlite - databases do not support enums, in order for this to be merged, I would say that this would _definitely_ be required to support Postgres and SQL Server.

### There are no tests

We absolutely cannot introduce new functionality/bug fixes without tests. When we add code to the core, it increases our support footprint. While I'm sure the above works well, it will inevitably have bugs in random edge cases (thats just programming). We can at least ensure that basic functionality will continue working with any refactor/fix.

This isn't specific to just this pull request, but all pull requests. We - and our users! - greatly appreciate our test coverage :) .

### CakePHP 2.0 is in maintenance mode

We have supported 2.x long enough and [clearly communicated](https://bakery.cakephp.org/2017/07/22/cakephp_2100_released.html) that it's in bugfix only mode now. We both cannot:

- backport 3.x features to 2.x
- add new functionality to 2.x

Doing so will add to our maintenance overhead. As a group of volunteers contributing their spare time for free there's only so much we can do, and at the moment we are focused on maintaining 3.x and moving towards 4.x.

### Closing thoughts

I personally don't like enums in databases for various reasons.

- They are hard/annoying to change on tables will millions of records.
- They go against data normalization.
- You can use enums in code and store integer constants instead, and use validation rules to validate the range (doesnt scale across codebases/languages, but neither does using hardcoded strings).
- You may eventually need to add metadata to each constant, so a related table is better (premature optimization though).
- Forms require you to query for all possible values before returning the possibilities.
- If you need the same enum on multiple tables, it's going to eventually get out of sync somewhere. MySQL will cast invalid values to empty string - might not anymore, haven't checked = which isn't great.
- Doing i18n translation on the values is kinda weird, and can't be done within the database.

While we appreciate this pull request - all contributions are welcome! - I don't think we're in a position to merge the pull request at this time.

If you'd like to continue using this functionality, I encourage you to vendor these changes locally or use a forked version of cakephp 2.x with this patch in your composer manifest.

alphp:
Are we really going to discuss here about the suitability or not of the ENUM types?
Does anyone really think that I, who have been programming for more than 30 years, have not thought very well if I used those guys in those fields enough to determine what was the least bad option?
NO, please, do not get into useless discussions, thank you.

ADmad:
I don't think we need to get into the suitability of enums here.

The lack of enum support is a known limitation and not a bug. Since 2.x is in bugfix only mode now, we cannot accept any new feature additions. Hence closing this PR.

alphp:
As I see it, the questions to answer are:
- Should the "Shell Schema" be left without support for a type of data that is used in projects and causes errors?
- Is it necessary to include a test to admit the PR?
- Is it necessary to have full support to admit the PR?
- Does this PR break compatibility with other databases?

But apparently the "only bugfix" does not work in this way.
Enum already existed before CakePHP 2, the lack of support of "Shell Schema" has always been, at least, a TO-DO.
In the past there was already someone who was interested in adding support, and also who was interested in having support for the ENUM type.
But the same complaints were made that I summarize in: "I hate ENUM and nobody should use it"
But that is not a correct or realistic position.
You have already made the decision, I do not share it but I have to comply with it. I will continue to use and support CakePHP, of course. I will continue trying to contribute what I can. But today I am sad and angry about the decision made.

Issue #11726, opened 2018-02-16
save() updates all records if the table does not have a primary key


MCF:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.5.12
* Platform and Target: Windows Server 2008, SQL Server 2008, PHP 7.2.2

### What you did
Create a form from an auto generated entity (i.e. no Class file exists for the table).  The database table (for historical reasons) does not have a primary key set.  Submit the form, load the original entity, patch it with the form data, call the table->save() function.

### What happened
The SQL generated for the update statement did not include a where clause.  All records in the table where changed and set to the data specified in the form.

### What you expected to happen
I was unaware the table in question did not have a primary key.  I would have expected that the _save_ function would throw an error rather than executing an update without a where clause.

Is this expected behavior?  The solution to the problem is obvious, ensure the table has a primary key.  But I would think an error would be a better than the current behavior.

saeideng:
without creating a class for this sample
```php
    $entity=$this->Tags->find()->where(['id_'=>1])->first();
    $entity->isNew($bool);
    $entity->title="new";
    $this->Tags->save($entity);

    //$bool==true (Cannot insert row in "tags" table, it has no primary key. )
    //$bool==false (will update all record)
```

saeideng:
https://github.com/cakephp/cakephp/blob/e7935c97d32ca29ccdd48f6d9a8e3679c67c9325/src/ORM/Table.php#L1992-L1995

josegonzalez:
The problem here - as @saeideng showed in his example - is that the entity is not new, and therefore it goes down the update path.

If an existing entity with _no_ primary keys is saved, then it'll go through as `Entity::has()` iterates over the set of primary keys - which is empty - and checks on each. Since there are none, `has` returns `true`, and the save continues. Unfortunately, the missing primary keys means the whole thing fails.

A fix is obviously to return false if the array is empty, but I'm not sure how that will affect join table records.

markstory:
Updating all records is not expected or desirable. Failing to save when the table lacks a primary key seems reasonable to me.

markstory:
Pull request open now.

PR #11727, opened 2018-02-17
Do not allow updates to tables with no primary key

markstory:
Raise exceptions instead of allowing updates to tables without primary keys.

Refs #11726

### DIFF
--- a/src/ORM/Table.php
+++ b/src/ORM/Table.php
@@ -1989,10 +1989,10 @@ class Table implements RepositoryInterface, EventListenerInterface, EventDispatc
             return $entity;
         }
 
-        if (empty($primaryColumns)) {
+        if (count($primaryColumns) === 0) {
             $entityClass = get_class($entity);
             $table = $this->getTable();
-            $message = "Cannot update `$entityClass` the `$table` has no primary key.";
+            $message = "Cannot update `$entityClass`. The `$table` has no primary key.";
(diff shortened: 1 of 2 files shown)


src/ORM/Table.php
```diff
@@ -1989,6 +1989,13 @@ protected function _update($entity, $data)
             return $entity;
         }
 
+        if (empty($primaryColumns)) {
```
josegonzalez: Are there values that could be falsy that would accidentally trigger this to be true?

dereuromark: I dont think so.

src/ORM/Table.php
```diff
@@ -1989,6 +1989,13 @@ protected function _update($entity, $data)
...
+            $entityClass = get_class($entity);
+            $table = $this->getTable();
+            $message = "Cannot update `$entityClass` the `$table` has no primary key.";
```
dereuromark: is there a comma missing?

src/ORM/Table.php
markstory: No it should always be a list of columns.

markstory: I'll use `count()` instead

src/ORM/Table.php
markstory: Yes. Fixed now.

Issue #11735, opened 2018-02-19
[RFC] Policy about how to change existing interfaces


chinpei215:
This is a (multiple allowed):

* [ ] bug
* [ ] enhancement
* [x] feature-discussion (RFC)

* CakePHP Version: 4.0
* Platform and Target: Non-releated

### Problems

I think we don't have policies about how to change exsiting interfaces.

Sometime we add `@method` annotations to existing interfaces (e.g. EntityInterface), and call those undeclared methods. This way seems quite dangerous to me, as it can be happend even in minor/bugfix releases, and users doesn't realize the fact until those methods are called at runtime.

At other times, we add some auxiliary interfaces (e.g. InvalidPropertyInterface) to not change existing interfaces (e.g. EntityInterface). Normally, this way allows users to extend core classes safely. But I think this way also has several problems like the following:
- We sometimes take the above `@method` ways. So now we are calling undeclared methods on InvalidPropertyInterface [like this](https://github.com/cakephp/cakephp/blob/765d9053acc86ab08790c4323deeb7cc1a0f0812/src/ORM/Marshaller.php#L189-L191).
- When people want to implement their own EntityInterface classes, they have to implement not only EntityInterface, but also InvalidPropertyInterface. 

### Suggestions

I think we have several choices to solves the above problems.

**(1) Allow to change existing interfaces in minor releases**

In this way, people cannot extend core classes safely. They may encounter some errors when they upgrade their applications to next minor releases. But they can realize those errors at compile time. I think this is far better than now (runtime errors).

**(2) Stop adding new interfaces**

In theory, interfaces should not be changed in minor releases. But we often do that. Then I think it is better to not provide any interfaces. If we didn't have EntityInterface (and had AbstractEntitty instead), we could avoid the above problems.

In this way, people can extend core classes safely as long as they don't encounter name collision.

**(3) Add derived interfaces**

For example, when we want to change EntityInterface, we should create EntityInterface2 which extends EntityInterface. We shouldn't change any signatures of existing methods. Instead, we should use `instanceof EntityInterface2` checks in those methods. In new major releases, we can remove old interfaces.

In this way, people can extend core classes safely.

Personally, I think (3) is the best way. But I am OK with any ways if we can stop adding `@method` annotations to interfaces.

dereuromark:
(2) is not an option at all - if anything we should be using more and not less interfaces for typehinting and programming against.

[...]

PR #11736, opened 2018-02-19
Make view block inst. a bit less internally coupled.

dereuromark:
Follow https://github.com/cakephp/cakephp/pull/11691 and explanations around open closed @ https://github.com/cakephp/cakephp/issues/11735#issuecomment-366820612

If we dont use proper DI here (or some factory pattern), at least make it possible to overwrite on extension level without having to copy and paste the whole constructor.

We need to be a bit more careful here inside the framework to not anti-pattern here too much.
This usually will be used against it in terms of "architecture discussions" on various platforms.
Don't need to further fuel those.

### DIFF
--- a/src/Event/EventDispatcherTrait.php
+++ b/src/Event/EventDispatcherTrait.php
@@ -33,7 +33,7 @@ trait EventDispatcherTrait
      *
      * @var string
      */
-    protected $_eventClass = '\Cake\Event\Event';
+    protected $_eventClass = Event::class;
 
     /**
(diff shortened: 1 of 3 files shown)


dereuromark:
I looked into other `= new ClassName()` examples, and they seem to either use their own (protected ) pseudo factory method or can be somewhat considered so low level architecture, that extending is not something supported/desired from the framework perspective.
Still not clean here in a few places, but probably something that doesn't have to change (or can) until 4.0.

PR #11737, opened 2018-02-20
3.6: Always use consistent class usage and inclusion.

dereuromark:
Follow https://github.com/cakephp/cakephp/pull/11736

These changes are safer in 3.next.
The main issue is that comparing the string class names requires them to be consistent to what ::class returns, and that is always without a leading slash.
So this makes it consistent and easier to work with. Proper include instead of magic strings also allows discovery and IDE usage detection and refactor.

### DIFF
--- a/src/TestSuite/TestCase.php
+++ b/src/TestSuite/TestCase.php
@@ -17,6 +17,7 @@ use Cake\Core\App;
 use Cake\Core\Configure;
 use Cake\Datasource\ConnectionManager;
 use Cake\Event\EventManager;
+use Cake\ORM\Entity;
 use Cake\ORM\Exception\MissingTableClassException;
 use Cake\ORM\Locator\LocatorAwareTrait;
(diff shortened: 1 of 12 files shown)


src/TestSuite/TestCase.php
```diff
@@ -718,6 +712,28 @@ public function getMockForModel($alias, array $methods = [], array $options = []
...
 
+    /**
+     * Gets the class name.
+     *
+     * @param string $alias The model to get a mock for.
+     * @param array $options The config data for the mock's constructor.
+     * @return string
+     * @throws \Cake\ORM\Exception\MissingTableClassException
+     */
+    protected function _getClassName($alias, array $options)
```
markstory: Should this name be more specific? Perhaps `getTableClassName()`?

dereuromark:
I rebased on top of the conflicting Table Locator changes. Should be all good now.

PR #11741, opened 2018-02-20
Fix Paginator unnecessarily modifying "sort" option.

### LINKED ISSUE #11740: Paginator unnecessarily modifies value of "sort" option.
This is a (multiple allowed):

* [x] bug
* [ ] enhancement

* CakePHP Version: 3.5.12

### What you did
Accessed sorting URL like `/posts?direction=asc&sort=title` generated by `PaginatorHelper::sort()` and then used `PaginatorHelper::numbers()` to generate paging links.

### What happened
There number links generated were like `/posts?page=2&sort=Posts.title&direction=asc`

I expected the links to be like `/posts?page=2&sort=title&direction=asc`.

Internally when generation value of `order` option `Paginator::validateSort()` adds model alias as prefix to value of `sort` thereby producing `order` as `['Posts.title' => 'ASC']`. This `order` option is used to reset `sort` and that is how links generated by PaginatorHelper::numbers()` end up with `sort=Posts.title` instead of `sort=Posts.title`. The sorting link generated by `PaginatorHelper::sort()` still uses the non prefix field `sort=title`.

**Now this might seem like just a cosmetic issue but it becomes a real problem when `sortWhitelist` option is set.** 

If I set `'sortWhitelist' => ['title']` then `Paginator::validateSort()` rejects the `sort` value from query string when it's `sort=Posts.title`. So to avoid the problem and ensure links generated by `PaginatorHelper::sort()` as well as `PaginatorHelper::number()` work I have to set the prefixed as well as non-prefixed version of a field in whitelist as `'sortWhitelist' => ['title', 'Posts.title']`. Such redundancy should not be required.

To resolve the issue the original value of `sort` from request query string should be maintained and not reset from `order`. Only in case where validation of `sort` value fails should `sort` be set to `null`.

ADmad:
Closes #11740.

### DIFF
--- a/src/Datasource/Paginator.php
+++ b/src/Datasource/Paginator.php
@@ -209,7 +209,7 @@ class Paginator implements PaginatorInterface
             'prevPage' => $page > 1,
             'nextPage' => $count > ($page * $limit),
             'pageCount' => $pageCount,
-            'sort' => key($order),
+            'sort' => $options['sort'],
             'direction' => current($order),
             'limit' => $defaults['limit'] != $limit ? $limit : null,
             'sortDefault' => $sortDefault,
@@ -349,8 +349,10 @@ class Paginator implements PaginatorInterface
                 $direction = 'asc';
             }
             $options['order'] = [$options['sort'] => $direction];
+        } else {
+            $options['sort'] = null;
         }
-        unset($options['sort'], $options['direction']);
+        unset($options['direction']);
 
         if (empty($options['order'])) {
             $options['order'] = [];
@@ -365,6 +367,7 @@ class Paginator implements PaginatorInterface
             $inWhitelist = in_array($field, $options['sortWhitelist'], true);
             if (!$inWhitelist) {
                 $options['order'] = [];
+                $options['sort'] = null;
 
                 return $options;
             }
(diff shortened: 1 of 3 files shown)


markstory (approved):
Looks pretty good to me.

dereuromark:
This introduced a small regression in the pagination sort() part.
The default sort order (in my case priority ASC) is not applied anymore as class to the tr th.

        $this->paginate = [
            'order' => ['priority' => 'ASC'],
        ];

    <a class="desc" href="/controller?sort=priority&amp;direction=asc">Priority</a>

You need to first click, and then the 2nd toggle does not bring my back to just /controller, but instead always keeps now the query strings on top (and needs them to function from here on).

Click result for the same link (`$this->Paginator->sort('priority', null, ['direction' => 'asc'])`):
```
http://app.local/controller
http://app.local/controller?direction=asc&sort=priority
http://app.local/controller?sort=priority&direction=desc
http://app.local/controller?sort=priority&direction=asc
http://app.local/controller?sort=priority&direction=desc
...
```
This used to be (automatically priority asc on empty query string)
```
http://app.local/controller
http://app.local/controller?sort=priority&direction=desc
http://app.local/controller?sort=priority&direction=asc
http://app.local/controller?sort=priority&direction=desc
...
```

So the default is broken for displaying the current order. 
And what always didnt work: the URL for the 2nd (back to the default) one.

ADmad:
@dereuromark I'll look into it.

PR #11747, opened 2018-02-22
Add Command factory support.

robertpustulka:
Makes it easier to use a custom factory, for example with DIC support.

Refs #11137


src/Console/CommandFactory.php
```diff
@@ -0,0 +1,39 @@
...
+
+use InvalidArgumentException;
+
+class CommandFactory implements CommandFactoryInterface
+{
+
+    /**
+     * {@inheritDoc}
+     */
+    public function create($className, ConsoleIo $io)
```
dereuromark: Should we also have interface contracts for ConsoleIo as typehints? This couples it tightly to the concrete implementation.

robertpustulka: `Command::run` takes `ConsoleIo` as a second argument. But the class has been added in 3.6 so it's not late to change that.

src/Console/CommandFactory.php
```diff
@@ -0,0 +1,39 @@
...
+ *
+ * @copyright     Copyright (c) Cake Software Foundation, Inc. (http://cakefoundation.org)
+ * @link          http://cakephp.org CakePHP(tm) Project
+ * @license       http://www.opensource.org/licenses/mit-license.php MIT License
+ */
+namespace Cake\Console;
+
+use InvalidArgumentException;
+
+class CommandFactory implements CommandFactoryInterface
```
markstory: Would be good to have a class doc here, as this is a 'public' class that people are likely to use/extend.

src/Console/CommandFactory.php
markstory: Another option would be to remove the typehints on concrete classes. I'm not sure ConsoleIo needs an interface as it isn't something that userland code should be replacing frequently.

robertpustulka: An interface could become handy in tests. But I think we can leave this class as it is and get back to it in 4.0 if necessary,

robertpustulka: Maybe `ConsoleIo $io` should be removed from a `create()` method? 

This is needed for `Shell` contructor which takes a `ConsoleIo` instance as an optional paramter. Since Shells are going to be deprecated and `$io` is an optional parameter, the proper io instance could be set by a command runner (there is `Shell::setIo()` setter). This way we wouldn't need to change the interface after shells get removed in the future.

robertpustulka: Ok, I'll change the interface and remove the `$io` param later this week.

robertpustulka:
`$io` parameter has been removed. The `ConsoleIO` instance is set by the command runner.

PR #11748, opened 2018-02-22
Change to class constant

davidyell:
Saves on potential typos. Introduced in 5.5.0, so allowed by the framework constraint


src/Datasource/QueryTrait.php
```diff
@@ -526,6 +526,6 @@ protected function _decorateResults($result)
      */
     protected function _decoratorClass()
     {
-        return 'Cake\Datasource\ResultSetDecorator';
+        return \Cake\Datasource\ResultSetDecorator::class;
```
dereuromark: Please use FQCN in use statement then.

dereuromark (approved):

Issue #10685, opened 2017-05-23
3.5 Add Configure::consumeOrFail()


inoas:
This is a (multiple allowed):

* [x] enhancement
* [x] feature-discussion (RFC)

* CakePHP Version: 3.4.6

### What you did
Read fallback database sources.

### What happened
`Datasource` is being consumed but `FallbackDatasource` is not.
Realized I cannot move from `readOrFail()` to `consumeOrFail()` cause latter misses.

### What you expected to happen
`consumeOrFail()` to exist.

```php
    /**
     * Used to consume information stored in Configure. It's not
     * possible to store `null` values in Configure.
     *
     * Acts as a wrapper around Configure::consume() and Configure::check().
     * The configure key/value pair consumed via this method is expected to exist.
     * In case it does not an exception will be thrown.
     *
     * @param string $var Variable to consume. Use '.' to access array elements.
     * @return mixed Value stored in configure.
     * @throws \RuntimeException if the requested configuration is not set.
     * @link http://book.cakephp.org/3.0/en/development/configuration.html#reading-configuration-data
     */
    public static function consumeOrFail($var)
    {
        if (static::check($var) === false) {
            throw new RuntimeException(sprintf('Expected configuration key "%s" not found.', $var));
        }

        return static::consume($var);
    }
```

### RFC
Replace all calling instances of consume() with consumeOrFail()?

markstory:
We could just make `consume()` raise exceptions when keys are missing too :smile:

inoas:
Well I'd love too. I think both methods should actually throw exceptions. 
However it is probably not BC?

dereuromark:
In that case a new method consumeOrFail() makes sense to have BC for the old methods.

markstory:
Adding new exceptions to existing methods doesn't break backwards compatibility according to our guidelines.

dereuromark:
Only exceptions in case of unexpected outcome.
In this case here this might Break a few live Apps  that Just didnt expect it.

markstory:
>  In this case here this might Break a few live Apps that Just didnt expect it.

But its the kind of change that would fit into a minor release like 3.5.0 or 3.6.0.

inoas:
I was actually using the consumeOrFail() above in Mysql::connect() to realize that it gets called multiple times and then the configuration parameter is not available anymore.

Now I understand why, but if that changes all of a sudden in an app that you upgrade, it is no fun.

[...]

PR #11752, opened 2018-02-23
Add consumeOrFail().

dereuromark:
Took the liberty of finishing up https://github.com/cakephp/cakephp/issues/10685

### DIFF
--- a/tests/TestCase/Core/ConfigureTest.php
+++ b/tests/TestCase/Core/ConfigureTest.php
@@ -602,7 +602,7 @@ class ConfigureTest extends TestCase
     {
         $expected = 'ok';
         Configure::write('This.Key.Exists', $expected);
-        $result = Configure::readOrFail('This.Key.Exists');
+        $result = Configure::consumeOrFail('This.Key.Exists');
         $this->assertEquals($expected, $result);
     }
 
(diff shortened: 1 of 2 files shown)


inoas:
Would it make sense to replace the few places where consume() is being used within the core framework with consumeOrFail()??

dereuromark:
This can be a follow up PR that runs the tests then separately to make sure nothing goes wrong.

dereuromark:
Yeah, I fixed that in the 2nd commit.

tests/TestCase/Core/ConfigureTest.php
```diff
@@ -592,4 +592,29 @@ public function testConsumeEmpty()
...
+     *
+     * @return void
+     */
+    public function testConsumeOrFail()
+    {
+        $expected = 'ok';
+        Configure::write('This.Key.Exists', $expected);
+        $result = Configure::consumeOrFail('This.Key.Exists');
+        $this->assertEquals($expected, $result);
+    }
```
ADmad: An assertion testing that the value has been consumed and cannot be read again would be nice.

dereuromark: I think that part is covered in consume tests itself already.

dereuromark:
Docs @ https://github.com/cakephp/docs/pull/5641

PR #11754, opened 2018-02-24
Batch casting of fields in database results

lorenzo:
This adds a new `BatchCastingInterface` to type classes. The idea
is that converting fields in a batch is sometimes faster than doing
them one by one.

I also changed the code of some of the casting functions in the type
classes to avoid waste and use faster primitives.

After all these changes are applied, you can expect the process of casting
fields to be around 10 percent faster than before. It is less than I
was expecting, but it is something, at least for wide tables.

### DIFF
--- a/src/Database/FieldTypeConverter.php
+++ b/src/Database/FieldTypeConverter.php
@@ -87,7 +87,6 @@ class FieldTypeConverter
             }
         }
 
-
         // Using batching when there is onl a couple for the type is actually slower,
         // so, let's check for that case here.
         foreach ($batchingResult as $type => $fields) {
--- a/src/Database/FieldTypeConverter.php
+++ b/src/Database/FieldTypeConverter.php
@@ -87,7 +87,7 @@ class FieldTypeConverter
             }
         }
 
-        // Using batching when there is onl a couple for the type is actually slower,
+        // Using batching when there is only a couple for the type is actually slower,
         // so, let's check for that case here.
         foreach ($batchingResult as $type => $fields) {
             if (count($fields) > 2) {
--- a/src/Database/FieldTypeConverter.php
+++ b/src/Database/FieldTypeConverter.php
@@ -14,6 +14,7 @@
  */
 namespace Cake\Database;
 
+use Cake\Database\Type;
 use Cake\Database\Type\BatchCastingInterface;
 use Cake\Database\Type\OptionalConvertInterface;
 
@@ -40,6 +41,13 @@ class FieldTypeConverter
      */
     protected $batchingTypeMap;
 
+    /**
+     * An array containing all the types registered in the Type system
+     * at the moment this object is created. Used so that the types list
+     * is not fetched on each single row of the results.
+     *
+     * @var array
+     */
     protected $types;
 
     /**
@@ -69,7 +77,15 @@ class FieldTypeConverter
                 continue;
             }
 
-            if ($type instanceof BatchCastingInterface) {
+            // Because of backwards compatibility reasons, we won't allow classes
+            // inheriting Type in userland code to be batchable, even if they implement
+            // the interface. Users can implement the TypeInterface instead to have
+            // access to this feature.
+            $batchingType = $type instanceof BatchCastingInterface &&
+                $type instanceof Type &&
+                strpos(get_class($type), 'Cake\Database\Type') === false;
+
+            if ($batchingType) {
                 $batchingMap[$k] = $type;
                 continue;
             }
--- a/src/Database/FieldTypeConverter.php
+++ b/src/Database/FieldTypeConverter.php
@@ -14,6 +14,7 @@
  */
 namespace Cake\Database;
 
+use Cake\Database\Type\BatchCastingInterface;
 use Cake\Database\Type\OptionalConvertInterface;
 
 /**
@@ -30,6 +31,16 @@ class FieldTypeConverter
      * @var array
      */
     protected $_typeMap;
+
+    /**
+     * An array containing the name of the fields and the Type objects
+     * each should use when converting them using batching.
+     *
+     * @var array
+     */
+    protected $batchingTypeMap;
+
+    protected $types;
 
     /**
      * The driver object to be used in the type conversion
@@ -49,20 +60,50 @@ class FieldTypeConverter
         $this->_driver = $driver;
         $map = $typeMap->toArray();
         $types = Type::buildAll();
-        $result = [];
+
+        $simpleMap = $batchingMap = [];
+        $simpleResult = $batchingResult = [];
 
         foreach ($types as $k => $type) {
             if ($type instanceof OptionalConvertInterface && !$type->requiresToPhpCast()) {
-                unset($types[$k]);
+                continue;
             }
+
+            if ($type instanceof BatchCastingInterface) {
+                $batchingMap[$k] = $type;
+                continue;
+            }
+
+            $simpleMap[$k] = $type;
         }
 
         foreach ($map as $field => $type) {
-            if (isset($types[$type])) {
-                $result[$field] = $types[$type];
+            if (isset($simpleMap[$type])) {
+                $simpleResult[$field] = $simpleMap[$type];
+                continue;
+            }
+            if (isset($batchingMap[$type])) {
+                $batchingResult[$type][] = $field;
             }
         }
-        $this->_typeMap = $result;
+
+
+        // Using batching when there is onl a couple for the type is actually slower,
+        // so, let's check for that case here.
+        foreach ($batchingResult as $type => $fields) {
+            if (count($fields) > 2) {
+                continue;
+            }
+
+            foreach ($fields as $f) {
+                $simpleResult[$f] = $batchingMap[$type];
+            }
+            unset($batchingResult[$type]);
+        }
+
+        $this->types = $types;
+        $this->_typeMap = $simpleResult;
+        $this->batchingTypeMap = $batchingResult;
     }
 
     /**
@@ -74,8 +115,16 @@ class FieldTypeConverter
      */
     public function __invoke($row)
     {
-        foreach ($this->_typeMap as $field => $type) {
-            $row[$field] = $type->toPHP($row[$field], $this->_driver);
+        if (!empty($this->_typeMap)) {
+            foreach ($this->_typeMap as $field => $type) {
+                $row[$field] = $type->toPHP($row[$field], $this->_driver);
+            }
+        }
+
+        if (!empty($this->batchingTypeMap)) {
+            foreach ($this->batchingTypeMap as $t => $fields) {
+                $row = $this->types[$t]->manyToPHP($row, $fields, $this->_driver);
+            }
         }
 
         return $row;
--- a/src/Database/Query.php
+++ b/src/Database/Query.php
@@ -2094,7 +2094,6 @@ class Query implements ExpressionInterface, IteratorAggregate
         $driver = $this->getConnection()->getDriver();
 
         if ($this->typeCastEnabled && $typeMap->toArray()) {
-            $driver = $this->_connection->getDriver();
             $statement = new CallbackStatement($statement, $driver, new FieldTypeConverter($typeMap, $driver));
         }
 
--- a/src/Database/Query.php
+++ b/src/Database/Query.php
@@ -2035,7 +2035,7 @@ class Query implements ExpressionInterface, IteratorAggregate
     }
 
     /**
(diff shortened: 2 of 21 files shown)


tests/TestCase/Database/Type/FloatTypeTest.php
```diff
@@ -82,17 +82,36 @@ public function testToPHP()
...
         $this->assertSame(2.0, $result);
 
-        $result = $this->type->toPHP('2 bears', $this->driver);
-        $this->assertSame(2.0, $result);
+        $result = $this->type->toPHP('15.3', $this->driver);
+        $this->assertSame(15.3, $result);
+    }
 
-        $result = $this->type->toPHP(['3', '4'], $this->driver);
-        $this->assertSame(1.0, $result);
```
lorenzo: The changes here probably require a bit of history:

When I started creating the Type classes, I though that `toPHP` would also be used for marshalling user data. Therefore, I made some checks to avoid warnings when casting bad things. Reality showed that we actually need a `marshal` function, as there was need to have different logic there.

The checks were left there, despite these functions being called almost exclusively with strings, instead of other types. I decided to get rid of those now, so I can take advantage of faster (and less) functions calls.

src/Database/Type/DecimalType.php
```diff
@@ -100,10 +101,34 @@ public function toDatabase($value, Driver $driver)
     public function toPHP($value, Driver $driver)
     {
         if ($value === null) {
-            return null;
+            return $value;
+        }
+
+        // Using coersion is faster than casting
+        // @codingStandardsIgnoreStart
+        return (float)+$value;
```
lorenzo: I benchmarked this agains `(int)$value` and just `0 + $value`. This version won by a surprisingly big margin.

src/Database/Query.php
```diff
@@ -2073,8 +2090,16 @@ public function selectTypeMap(TypeMap $typeMap = null)
      */
     protected function _decorateStatement($statement)
     {
+        $typeMap = $this->getSelectTypeMap();
+        $driver = $this->getConnection()->getDriver();
+
+        if ($this->typeCastEnabled && $typeMap->toArray()) {
+            $driver = $this->_connection->getDriver();
```
markstory: Do you need to fetch the driver again? You already got it on line 2094.

src/Database/Type/DecimalType.php
markstory: Interesting, I would have expected them to come out the same.

tests/TestCase/Database/Type/FloatTypeTest.php
markstory: Makes sense. These methods are not exposed to raw user data so they don't need to handle as many scenarios.

src/Database/FieldTypeConverter.php
```diff
@@ -49,20 +60,49 @@ public function __construct(TypeMap $typeMap, Driver $driver)
...
+        $simpleMap = $batchingMap = [];
+        $simpleResult = $batchingResult = [];
 
         foreach ($types as $k => $type) {
             if ($type instanceof OptionalConvertInterface && !$type->requiresToPhpCast()) {
-                unset($types[$k]);
+                continue;
             }
+
+            if ($type instanceof BatchCastingInterface) {
```
chinpei215: I think we have to add more methods which enables/disables the batch casting feature to the Type class:
```php
class Type implements TypeInterface
    static $batchCastingEnabled = [
        StringType::class => true,
        ...
    ];

    static function isBatchCastingEnabled($className) {
        return !empty(self::$batchCastingEnabled[$className]);
    }
    static function enableBatchCasting($className) { ... }
    static function disableBatchCasting($className) { ... }
}
```
```php
if ($type instanceof BatchCastingInterface && Type::isBatchCastingEnabled(get_class($type))) { 
    ...   
}
```

Otherwise, it will be hard to extend core Type classes. The following class no longer works in 3.next:
```php
class MyStringType extends StringType
{
    public function toPHP($value, Driver $driver) 
        // Do something
    }
}
```
Because toPHP() may not be called, in favor of manyToPHP(). People would have to copy&paste the manyToPHP() code of the parent class. Actually, we have to remove toPHP() from DateType and add $setToDateStart to the DateTimeType. 

If we allow people to enable/disable batch casting, the above MyStringType class will work in 3.next without any changes, as the class name is not in the $batchCastingEnabled array by default.

lorenzo: I added $setToDateStart as an optimization to avoid calling an extra function, not to overcome any issues, but you may be right. In that this would make some methods not be called anymore.

Maybe document the change in the migration? I don’t think disabling the feature would help people not knowing that there was this change.

src/Database/Type/DecimalType.php
```diff
@@ -100,10 +101,34 @@ public function toDatabase($value, Driver $driver)
     public function toPHP($value, Driver $driver)
     {
         if ($value === null) {
-            return null;
+            return $value;
+        }
+
+        // Using coercion is faster than casting
+        // @codingStandardsIgnoreStart
+        return (float)+$value;
```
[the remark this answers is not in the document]
lorenzo: With a script similar to yours. But you are not testing anything there, the $i should be string, not an integer.

src/Database/Query.php
lorenzo: Yeah, bad copy paste. Thanks!

src/Database/FieldTypeConverter.php
lorenzo: Looks like your suggestion would work to prevent the issue you described. I’m just not too happy we have to do that.

markstory: Is there an option where when batch casting is enabled, all casting operations are done through that method, even single columns? With that and an opt-in for batch casting will not cause problems in 3.x and in 4.x we could remove `toPHP` and always use batch casting.

src/Database/Type/DecimalType.php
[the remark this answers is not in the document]
chinpei215: I read PHP source code and now I understand what is happening in PHP.

If scalars are evaluated as `IS_LONG` (e.g. `'123'`), `(float)+$value` is a little faster than `(float)$value`:
```php
$t = microtime(true);
for ($i = 0; $i < 10000000; ++$i) {
    $value = (string)$i;
    (float)$value;
}
var_dump( microtime(true) - $t );
// float(1.2263522148132)

$t = microtime(true);
for ($i = 0; $i < 10000000; ++$i) {
    $value = (string)$i;
    (float)+$value;
}
var_dump( microtime(true) - $t );
// float(1.1857590675354)
```
Becase the latter doesn't call the `zend_strtod()` and does [a simple operation](https://github.com/php/php-src/blob/414aaa207cc59c279d3fd5ed83f6b92890e57023/Zend/zend_operators.c#L3000) to get long values in such cases
 
However, scalars are evaluated as `IS_DOUBLE` (e.g. `'123.0'`), `(float)+$value` is much slower than `(float)$value`:
```php
$t = microtime(true);
for ($i = 0; $i < 10000000; ++$i) {
    $value = $i . '.0';
    (float)$value;
}
var_dump( microtime(true) - $t );
// float(1.5314469337463)

$t = microtime(true);
for ($i = 0; $i < 10000000; ++$i) {
    $value = $i . '.0';
    (float)+$value;
}
var_dump( microtime(true) - $t );
// float(1.9330399036407)
```

Since float/decimal columns contain decimal points usually, I don't think this can be an optimization.

Could you please confirm the above two benchmarks?

src/Database/FieldTypeConverter.php
chinpei215: I agree with there is no need to disable the feature. I am OK if the above MyStringType works fine in 3.next without any changes.

lorenzo: The problem is that using the batch when there is not batching to be done is a bit slower

src/Database/Type/DecimalType.php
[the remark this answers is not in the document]
lorenzo: I can confirm your numbers, it seems I only checked this for ints, and not floats. Thanks for catching this!

src/Database/FieldTypeConverter.php
lorenzo: I'll measure again to see what's the downside of doing this

lorenzo: I tried making `toPHP` always call `manyToPHP` internally for the `IntegerType` class. It resulted in a significant performance loss. So this is not the shortcut I'd like to try.

Another shortcut is to hardcode the batching only for core classes and classes not extending from Type. I think that would avoid having that enable/disable methods and would still preserve all of the benefits.

What do you think?

chinpei215: We provide BatchCastingInterface only for users' classes, and core Types don't implement it? I think it works.

lorenzo:
I think this is ready to be merged now

src/Database/Type/DecimalType.php
```diff
@@ -100,10 +101,34 @@ public function toDatabase($value, Driver $driver)
...
+    public function manyToPHP(array $values, array $fields, Driver $driver)
+    {
+        foreach ($fields as $field) {
+            if (!isset($values[$field])) {
+                continue;
+            }
+
+            // Using coercion is faster than casting
+            // @codingStandardsIgnoreStart
+            $values[$field] = (float)+$values[$field];
```
chinpei215: Still remains.

src/Database/Type/DecimalType.php
chinpei215: Still remains.

src/Database/Type/IntegerType.php
```diff
@@ -80,10 +81,33 @@ public function toDatabase($value, Driver $driver)
...
     {
         if ($value === null) {
-            return null;
+            return $value;
         }
 
-        return (int)$value;
+        // Using coercion is faster than casting
+        // @codingStandardsIgnoreStart
+        return (int)+$value;
```
chinpei215: Still remains. Though it is `(int)+$value`, it is slower than `(int)$value` in PHP 7.x at least.

src/Database/Type/IntegerType.php
```diff
@@ -80,10 +81,33 @@ public function toDatabase($value, Driver $driver)
...
+     */
+    public function manyToPHP(array $values, array $fields, Driver $driver)
+    {
+        foreach ($fields as $field) {
+            if (!isset($values[$field])) {
+                continue;
+            }
+            // Using coercion is faster than casting
+            // @codingStandardsIgnoreStart
+            $values[$field] = (int)+$values[$field];
```
chinpei215: Still remains.

chinpei215:
Oh I couldn't leave comments in time...

src/Database/Type/DecimalType.php
lorenzo: Forgot that this was here, only removed in FloatType

src/Database/Type/IntegerType.php
dereuromark: This would need to be `(int) + $value` then for coding standards?
But is this new syntax really necessary given the fact that this is quite the nano opt. somewhat? Only will really a looot of casts this might affect in the micro second area right?

tests/TestCase/Database/QueryTest.php
```diff
@@ -4460,6 +4460,79 @@ public function testGetValueBinder()
...
+        $typeMap = new TypeMap($fields + ['a' => 'integer']);
+        $results = $query
+            ->select(array_keys($fields))
+            ->select(['a' => 'id'])
+            ->from('profiles')
+            ->setSelectTypeMap($typeMap)
+            ->limit(1)
+            ->execute()
+            ->fetchAll('assoc');
+        $this->assertSame([['id' => 1, 'user_id' => 1, 'is_active' => false, 'a' => 1]], $results);
```
saeideng: @lorenzo 
this expects
```php
       'id' => 0
        'user_id' => 1
        'is_active' => false
        'a' => 0
```
on windows 
https://ci.appveyor.com/project/cakephp/cakephp/build/1.0.9754#L297

lorenzo: add an order() clause

lorenzo: actually it is odd that the is is 0, the fixture seems to be wrong

saeideng: see next test please (`testCastResultsDisable`)
when calls  `disableResultsCasting()`
`'a' => '1'` is true

lorenzo: @saeideng What should I be looking for in that test?

PR #11757, opened 2018-02-25
Ensure paging params are properly set based on "order" in settings.

ADmad:
Fixes regression caused by #11741.
/cc @dereuromark


dereuromark (approved):
regression fixed, thx!

Issue #11753, opened 2018-02-24
IntegrationTestCase and PHP7.2


dereuromark:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: latest master
* Platform and Target: CakeBox in Linux Vagrant PHP 7.2

### What you did

So far this worked:

    php phpunit.phar tests/TestCase/Controller/AccountControllerTest.php

### What happened
But with new CakeBox and PHP7.2:

> Warning Error: session_set_save_handler(): Cannot change save handler when headers already sent in [/home/vagrant/Apps/.../vendor/cakephp/cakephp/src/Network/Session.php, line 224]

Wrapping

    session_set_save_handler($this->engine($class, $config['handler']), false);

with `if ($this->started())` in Session class seems to help.
Maybe, with some kind of static attribute, we can make sure this is only executed once?

PR #11758, opened 2018-02-25
Fix Session handling in IntegrationTestCase usage.

dereuromark:
Stab on fixing https://github.com/cakephp/cakephp/issues/11753
Not sure if this is the right way, but it seems to work.


markstory:
This will not overwrite the save handler once stdout has been written to. That will prevent session handling being changed partway through a test run, but I don't think that works now either.

Issue #11749, opened 2018-02-22
ability to read utf8 files by Cake\Filesystem\File


saeideng:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: cake 3.5.12
* Platform and Target: linux
```php
    $file='نام.txt';
    $file2='name.txt';
    $file3='ä.txt';
    $path=ROOT.DS.'webroot'.DS.'upload'.DS;
    
    $path=$path.$file;

    debug(is_file($path)); // true
    debug(file_get_contents($path)); //content for file 'نام.txt'
    $file = new File($path);
    debug($file);
//
...
	name => '.txt'
	info => []
	handle => null
	lock => null
	path => '/var/www/site/webroot/upload/.txt'
//
    debug($file->exists());// false
    debug($file->readable()); //false
```

davidyell:
Seems like a case which is missing from tests for sure, from what I can see.

https://github.com/cakephp/cakephp/blob/master/tests/TestCase/Filesystem/FileTest.php

hmic:
This is intended behavior as in:
https://api.cakephp.org/3.5/source-class-Cake.Filesystem.File.html#355

JoshuaLuckers:
What PHP version do you use?

davidyell:
I don't think it'll be the PHP version, as that bug is for `5.3.13` and the minimum for Cake 3.5.12 is `5.6.0`, but it's a good spot.

saeideng:
problem is `basename()`
```php
`debug(basename('نام.txt'));` //returns `.txt`
```

dereuromark:
So this is not a framework issue, as the language itself seems to have problems here.

saeideng:
>as the language itself seems to have problems here.

not related to just one language
this method not work with all utf8 names
other method is `getFilename`
```php
$str='ä.php';
$info = new SplFileInfo($str);
echo $info->getFilename();
```

dereuromark:
I was referring to PHP as the language. This seems to be out of scope of CakePHP.

saeideng:
```php
<?php
$str=[
'ddddd/نام.php',
'نام.php',
'نام فارسی.php',
'/نام فارسی.php',
'ddd/نام فارسی.php',
'/etc/sudoers.d',
'/etc/passwd',
'/etc/',
'.',
'/',

];
foreach($str as $item){
    $info = new SplFileInfo($item);
[...]

PR #11759, opened 2018-02-26
#11749 Attempt to fix UTF8 filenames

burzum:
See #11749

### DIFF
--- a/src/Filesystem/File.php
+++ b/src/Filesystem/File.php
@@ -15,6 +15,7 @@
 namespace Cake\Filesystem;
 
 use finfo;
+use SplFileInfo;
 
 /**
  * Convenience class for reading, writing and appending to files.
@@ -82,9 +83,10 @@ class File
      */
     public function __construct($path, $create = false, $mode = 0755)
     {
-        $this->Folder = new Folder(dirname($path), $create, $mode);
+        $splInfo = new SplFileInfo($path);
+        $this->Folder = new Folder($splInfo->getPath(), $create, $mode);
         if (!is_dir($path)) {
-            $this->name = basename($path);
+            $this->name = $splInfo->getBasename();
         }
         $this->pwd();
         $create && !$this->exists() && $this->safe($path) && $this->create();
--- a/src/Filesystem/File.php
+++ b/src/Filesystem/File.php
@@ -359,14 +359,15 @@ class File
      *
      * @return string the file basename.
      */
-    protected function _basename($name, $ext=null)
+    protected function _basename($name, $ext = null)
     {
         $splInfo = new SplFileInfo($name);
         $name = ltrim($splInfo->getFilename(), DS);
-        if($ext===null || rtrim($name,$ext)===''){
+        if ($ext === null || rtrim($name, $ext) === '') {
             return $name;
         }
-        return rtrim($name,$ext);
+
+        return rtrim($name, $ext);
     }
 
     /**
--- a/src/Filesystem/File.php
+++ b/src/Filesystem/File.php
@@ -86,7 +86,7 @@ class File
         $splInfo = new SplFileInfo($path);
         $this->Folder = new Folder($splInfo->getPath(), $create, $mode);
         if (!is_dir($path)) {
-            $this->name = ltrim($splInfo->getFilename(), DS);
+            $this->name = ltrim($splInfo->getFilename(), '/\\');
         }
         $this->pwd();
         $create && !$this->exists() && $this->safe($path) && $this->create();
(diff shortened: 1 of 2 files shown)


src/Filesystem/File.php
```diff
@@ -82,9 +83,10 @@ class File
      */
     public function __construct($path, $create = false, $mode = 0755)
     {
-        $this->Folder = new Folder(dirname($path), $create, $mode);
+        $splInfo = new SplFileInfo($path);
+        $this->Folder = new Folder($splInfo->getPath(), $create, $mode);
         if (!is_dir($path)) {
-            $this->name = basename($path);
+            $this->name = $splInfo->getBasename();
```
saeideng: this method does not work on browser ( apache ) 
but console is OK 
```php
<?php
$path='tests/permissions/نام فارسی.php';
$splInfo = new \SplFileInfo($path);

var_dump($splInfo->getFilename());
var_dump($splInfo->getBasename());
var_dump(basename($path));
```

burzum:
Looks good now!

tests/TestCase/Filesystem/FileTest.php
```diff
@@ -123,6 +123,69 @@ function_exists('mime_content_type') &&
...
+     */
+    public function testBasename($path, $suffix)
+    {
+        $class = new \ReflectionClass('Cake\Filesystem\File');
+        $method = $class->getMethod('_basename');
+        $method->setAccessible(true);
+        if ($suffix === null) {
+            $this->assertEquals(basename($path), $method->invokeArgs(null, [$path]));
+        } else {
+            $this->assertEquals(basename($path, $suffix), $method->invokeArgs(null, [$path, $suffix]));
```
markstory: If you want to call this method why is it not public? Coupling tests to 'implementation details' makes tests harder to maintain as private features are now public.

saeideng: we can add/move these tests to other public tests
i just want to know `_basename()` method is equal to `basename()` 
And i no want to have public method at this time

src/Filesystem/File.php
```diff
@@ -368,7 +388,7 @@ public function safe($name = null, $ext = null)
             $ext = $this->ext();
         }
 
-        return preg_replace("/(?:[^\w\.-]+)/", '_', basename($name, $ext));
+        return preg_replace("/(?:[^\w\.-]+)/", '_', static::_basename($name, $ext));
```
saeideng: probably needs a bit test to testing this method 
`public function safe($name = null, $ext = null)`

tests/TestCase/Filesystem/FileTest.php
markstory: Can we use a public interface of the class then?

saeideng: >Can we use a public interface of the class then?

1)having extended class withing testCase and having a method for wrapping _basename() 
like the https://github.com/cakephp/cakephp/blob/master/tests/TestCase/Mailer/EmailTest.php#L89-L92 in to testcase?

2) or moving test to public methods?

in case 2 , not fully - we can not cover all case - i'm not sure

markstory: Can we not create empty files for each case and check their names?

saeideng: I'll move these tests  to `name()` without having real file

src/Filesystem/File.php
```diff
@@ -352,6 +354,24 @@ public function name()
...
+     *
+     * @param string $path Path to file
+     * @param string|null $ext The name of the extension
+     * @return string the file basename.
+     */
+    protected static function _basename($path, $ext = null)
+    {
+        $splInfo = new SplFileInfo($path);
+        $name = ltrim($splInfo->getFilename(), DS);
+        if ($ext === null || rtrim($name, $ext) === '') {
```
saeideng: `rtrim()`  must change with proper method
now it returns invalid result

saeideng (changes requested):

saeideng:
- there is problem with `/نام.ext` on Windowns yet , Linux is OK
also , `folderName/نام.ext` is OK in both 
https://ci.appveyor.com/project/cakephp/cakephp/build/1.0.9829#L274

- I added `(mb_strlen($path) === strlen($path))` for using `basename()` if file name/path was not `multibyte`

- I will move test from `_basename()` to `name()` if I able

saeideng:
>looks there is problem with /نام.ext on Windowns yet , Linux is OK
also , folderName/نام.ext is OK in both
https://ci.appveyor.com/project/cakephp/cakephp/build/1.0.9829#L274

Tests passed on windows (AppVeyor)

tests/TestCase/Filesystem/FileTest.php
markstory: Could the behavior of `_basename` be tested based on how file creation and `name()` are working?

saeideng:
finished

PR #11764, opened 2018-02-27
Improving exception msgs for database types

burzum:
It's annoying that you can't see the type and make a guess which of the many values it might be.

### DIFF
--- a/src/Database/Type/BoolType.php
+++ b/src/Database/Type/BoolType.php
@@ -67,7 +67,10 @@ class BoolType extends Type implements TypeInterface
             return (bool)$value;
         }
 
-        throw new InvalidArgumentException('Cannot convert value to bool');
+        throw new InvalidArgumentException(sprintf(
+            'Cannot convert value of type `%s` to bool',
+            getTypeName($value)
+        ));
     }
 
     /**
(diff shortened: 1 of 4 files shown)


src/Database/Type/BoolType.php
```diff
@@ -67,7 +67,10 @@ public function toDatabase($value, Driver $driver)
             return (bool)$value;
         }
 
-        throw new InvalidArgumentException('Cannot convert value to bool');
+        throw new InvalidArgumentException(sprintf(
+            'Cannot convert value of type `%s` to bool',
+            getTypeName($value)
```
dereuromark: Wouldnt it then also make sense to include the value itself directly? Or is that a bit too much then as debug info?

burzum: I wanted to do that, but how could we archive this? It's just the base \InvalidArgumentException and doesn't really have a way to attach arbitrary data to it? Also does the error renderer has the capability of showing that value? Especially in this case of wrong type values, know the content of `$value` would make debugging the cause extremely helpful.

dereuromark: Yeah, maybe a bit out of scope - or even overkill - to use some Debugger::exportVar().
Hopefully the stracktrace contains the caller arguments.

markstory:
Nice improvement :clap:

PR #11765, opened 2018-02-27
refs #10778 add router cache

steinkel:
Route collection caching added to RoutingMiddleware.
New configuration keys added 
* `Router.cache` bool, enable cache
* `Router.cacheConfig`  string, define the cache configuration name to be used

### DIFF
--- a/src/Routing/Middleware/RoutingMiddleware.php
+++ b/src/Routing/Middleware/RoutingMiddleware.php
@@ -33,11 +33,6 @@ use Zend\Diactoros\Response\RedirectResponse;
 class RoutingMiddleware
 {
     /**
-     * Name of the default cache configuration name used to store routes collection
-     */
-    const DEFAULT_ROUTER_CACHE_CONFIG = '_cake_router_';
-
-    /**
      * Key used to store the route collection in the cache engine
      */
     const ROUTE_COLLECTION_CACHE_KEY = 'routeCollection';
@@ -50,13 +45,23 @@ class RoutingMiddleware
     protected $app;
 
     /**
+     * The cache configuration name to use for route collection caching,
+     * null to disable caching
+     *
+     * @var string
+     */
+    protected $cacheConfig;
+
+    /**
      * Constructor
      *
      * @param \Cake\Http\BaseApplication $app The application instance that routes are defined on.
+     * @param string|null $cacheConfig The cache config name to use or null to disable routes cache
      */
-    public function __construct(BaseApplication $app = null)
+    public function __construct(BaseApplication $app = null, $cacheConfig = null)
     {
         $this->app = $app;
+        $this->cacheConfig = $cacheConfig;
     }
 
     /**
@@ -85,14 +90,10 @@ class RoutingMiddleware
      */
     protected function buildRouteCollection()
     {
-        $isRouterCacheEnabled = Configure::read('Router.cache');
-        if (Cache::enabled() && $isRouterCacheEnabled) {
-            $routesCacheConfig = Configure::read('Router.cacheConfig', static::DEFAULT_ROUTER_CACHE_CONFIG);
-
+        if (Cache::enabled() && $this->cacheConfig !== null) {
             return Cache::remember(static::ROUTE_COLLECTION_CACHE_KEY, function () {
-
                 return $this->prepareRouteCollection();
-            }, $routesCacheConfig);
+            }, $this->cacheConfig);
         }
 
         return $this->prepareRouteCollection();
--- a/src/Routing/Middleware/RoutingMiddleware.php
+++ b/src/Routing/Middleware/RoutingMiddleware.php
@@ -14,7 +14,6 @@
  */
 namespace Cake\Routing\Middleware;
 
-=======
 use Cake\Cache\Cache;
 use Cake\Core\Configure;
 use Cake\Core\PluginApplicationInterface;
@@ -74,10 +73,9 @@ class RoutingMiddleware
         if (!$this->app) {
             return;
         }
+
         $routeCollection = $this->buildRouteCollection();
         Router::setRouteCollection($routeCollection);
-        // Prevent routes from being loaded again
-        Router::$initialized = true;
     }
 
     /**
--- a/src/Routing/Middleware/RoutingMiddleware.php
+++ b/src/Routing/Middleware/RoutingMiddleware.php
@@ -89,9 +89,9 @@ class RoutingMiddleware
     {
         $isRouterCacheEnabled = Configure::read('Router.cache');
         if (Cache::enabled() && $isRouterCacheEnabled) {
-            $routesCacheConfig = Configure::read('Router.cacheConfig', self::DEFAULT_ROUTER_CACHE_CONFIG);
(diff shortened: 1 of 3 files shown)


src/Routing/Middleware/RoutingMiddleware.php
```diff
@@ -56,10 +69,46 @@ public function __construct(BaseApplication $app = null)
...
+    /**
+     * Check if route cache is enabled and use the configured Cache to 'remember' the route collection
+     *
+     * @return \Cake\Routing\RouteCollection
+     */
+    protected function buildRouteCollection()
+    {
+        $isRouterCacheEnabled = Configure::read('Router.cache');
+        if (Cache::enabled() && $isRouterCacheEnabled) {
+            $routesCacheConfig = Configure::read('Router.cacheConfig', self::DEFAULT_ROUTER_CACHE_CONFIG);
```
dereuromark: Can we please use static:: everywhere?

src/Routing/Router.php
```diff
@@ -1148,6 +1148,17 @@ public static function getRouteCollection()
...
 
+    /**
+     * Set the RouteCollection inside the Router
+     *
+     * @param RouteCollection $routeCollection route collection
+     * @return void
+     */
+    public static function setRouteCollection($routeCollection)
+    {
+        static::$_collection = $routeCollection;
```
lorenzo: You can set the `$initialized` property here, and avoid to do it in the middleware

steinkel: that makes a lot of sense, thanks @lorenzo

src/Routing/Middleware/RoutingMiddleware.php
```diff
@@ -56,10 +69,44 @@ public function __construct(BaseApplication $app = null)
...
+
+    /**
+     * Generate the route collection using the builder
+     *
+     * @return \Cake\Routing\RouteCollection
+     */
+    protected function prepareRouteCollection()
+    {
+        $builder = Router::createRouteBuilder('/');
+        $this->app->routes($builder);
```
markstory: You might get conflicts when the plugin changes land.

tests/TestCase/Routing/Middleware/RoutingMiddlewareTest.php
```diff
@@ -427,4 +429,91 @@ public function scopedMiddlewareUrlProvider()
...
+        Cache::setConfig('_cake_router_', [
+            'engine' => 'File',
+            'path' => TMP,
+        ]);
+        Router::$initialized = false;
+        $request = ServerRequestFactory::fromGlobals(['REQUEST_URI' => '/articles']);
+        $response = new Response();
+        $next = function ($req, $res) {
+            $routeCollection = Cache::read('routeCollection', '_cake_router_');
+            $this->assertInstanceOf('\Cake\Routing\RouteCollection', $routeCollection);
```
markstory: You could use `RouteCollection::class` here.

tests/TestCase/Routing/Middleware/RoutingMiddlewareTest.php
```diff
@@ -427,4 +429,91 @@ public function scopedMiddlewareUrlProvider()
...
+        Cache::drop('_cake_router_');
+        Cache::enable();
+    }
+
+    /**
+     * Test cache name is used
+     *
+     * @return void
+     * @expectedException InvalidArgumentException
+     * @expectedExceptionMessage The "notfound" cache configuration does not exist
```
markstory: We've been using the `setException()` method instead, as annotations are deprecated from what I remember.

markstory (commented):
Looking great! This is much cleaner than the last approach.

tests/TestCase/Routing/Middleware/RoutingMiddlewareTest.php
robertpustulka: `expectException()` is the method.

tests/TestCase/Routing/Middleware/RoutingMiddlewareTest.php
steinkel: :white_check_mark:

tests/TestCase/Routing/Middleware/RoutingMiddlewareTest.php
steinkel: :white_check_mark:

steinkel:
rebased and fixed conflicts

burzum:
Just an idea, can't we somehow inject the cache instead of using the static inside the code?

Also will this introduce new required dependencies on other cake packages? It would be cool if other projects could use the router without half of the Cake frameworks other packages. It's just a suggestion!

markstory:
We don't already support psr-6. Currently Routing is not offered as a sub-split making the internal dependencies less problematic. We could take the cache engine as a parameter instead if using `Cache`. That may add more complex logic to Router as `remember()` is nit an engine method.

src/Routing/Middleware/RoutingMiddleware.php
```diff
@@ -60,11 +73,45 @@ protected function loadRoutes()
...
+    }
+
+    /**
+     * Check if route cache is enabled and use the configured Cache to 'remember' the route collection
+     *
+     * @return \Cake\Routing\RouteCollection
+     */
+    protected function buildRouteCollection()
+    {
+        $isRouterCacheEnabled = Configure::read('Router.cache');
```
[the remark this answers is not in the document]
markstory: Passing in the cache engine name would remove the need to have a `Router.cache` configure option as well which is nice.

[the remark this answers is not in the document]
steinkel: The intention here was to provide a way to override this behavior locally, for example in development vs production, and allow an easy environment based configuration.

I can refactor the middleware constructor to pass the cache configuration name (or null to disable) then set defaults in Application and remove the need of these 2 config keys if you think it would be better to reduce the amount of config keys.

Thanks,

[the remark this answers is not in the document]
markstory: The caller can manage configuration keys too. Doing that makes the middleware easier to adapt to di containers in the future as well.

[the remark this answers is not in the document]
steinkel: true, OK I'll do these changes and remove config from the middleware and read the config from the Application class

markstory (approved):
Looks good to me.

steinkel:
Thank you, as soon as we get this merged I'll work on a PR to app project to enable router cache by default and provide a default cache config.

burzum:
@steinkel did you do any benchmarks? How much improvement will this give us?

steinkel:
Yes, some benchmarks here

Times with/without cache were:
0.00463 vs 0.00276 for 50 connected routes
1.45s vs 0.015 for 4000 connected routes

markstory:
@steinkel I'll get this merged as soon as I can get the docs updated.

steinkel:
great! let me know the plan for the docs I could work on that too, meanwhile I'll prepare a PR to the app project to enable by default and provide a cache config.

Thanks,

PR #11770, opened 2018-02-28
Use variadic isset

bcrowe:
We use variadic `isset()` everywhere else... and noticed these two.

### DIFF
--- a/src/Validation/Validation.php
+++ b/src/Validation/Validation.php
@@ -1231,7 +1231,7 @@ class Validation
         if (isset($options['width'])) {
             $validWidth = self::comparison($width, $options['width'][0], $options['width'][1]);
         }
-        if (isset($validHeight) && isset($validWidth)) {
+        if (isset($validHeight, $validWidth)) {
             return ($validHeight && $validWidth);
         }
         if (isset($validHeight)) {
--- a/src/ORM/EagerLoader.php
+++ b/src/ORM/EagerLoader.php
@@ -398,7 +398,7 @@ class EagerLoader
 
             $pointer += [$table => []];
 
-            if (isset($options['queryBuilder']) && isset($pointer[$table]['queryBuilder'])) {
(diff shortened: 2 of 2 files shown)


dereuromark:
The main problem I have with this is clarity:
Even more experienced devs do not know right away if this is an AND or OR connection here.
The previous example is more readable IMO, and we gain nothing by using the variadic approach really.

dereuromark:
To clarify:
If those are NOT checked on, just executed, then it is irrelevant.
But here, used inside if() conditions this is rather important to know - also for future refactoring.

I am against the usage inside of conditions to prevent mistakes from happening when touching those.

src/Validation/Validation.php
```diff
@@ -1231,7 +1231,7 @@ public static function imageSize($file, $options)
         if (isset($options['width'])) {
             $validWidth = self::comparison($width, $options['width'][0], $options['width'][1]);
         }
-        if (isset($validHeight) && isset($validWidth)) {
+        if (isset($validHeight, $validWidth)) {
```
dereuromark: This should actually be `!== null` both - using isset() with existing vars is a big code smell / antipattern.

bcrowe:
I don't have a strong opinion on varidiac vs. multiple `issets()`, just wanted to touch on consistency. I'm okay with PRing the inverse of this as well. If you look around, you'll see quite a few `isset($a, $b, $c)`s.

dereuromark:
But are they ever used inside a condition? Or just void execution? That is a huge difference as outlined above.

bcrowe:
All of them are inside of conditions: `/isset\([\$a-zA-Z0-9\[\'\]]*\,/g`

dereuromark:
Examples like

    if (isset($lower, $upper)) {

should also be fixed to strict null check instead.
But other than that I guess we can move forward with the array isset check - for consistency as you mentioned.

markstory:
Merging as we use variadic isset in several places already and this is adding consistency there. The familiarity with variadic isset should be a one time learning cost for most folks.

PR #11771, opened 2018-02-28
Fixed the multiple option when it is set false on belongsToMany selects

keomaborges:
I faced this problem when I tried to set `'multiple' = false` in a select field which sets a belongs to many relationship (my field calls `stores._ids`). So I fixed how these especific option is built on FormHelper class.


markstory:
These changes [cause an existing test to fail](https://travis-ci.org/cakephp/cakephp/jobs/347405514#L762)  that is usually an indication of a regression. Could you look into the failing test?

keomaborges:
Ok, I'll chek it out.

markstory:
Thanks!

Issue #11712, opened 2018-02-12
ORM does not automatically convert timezones to the application timezone upon saving


samuelszabo:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.5.12
* Platform and Target: built-in server, PHP 7.1.5, MySQL 5.6.34

### What you did
I would like to save 2 Time objects in different timezones:
```php
$entity = $this->newEntity([
    'date1' => new Time('2018-02-12 11:05:00', '+05:00'),
    'date2' => new Time('2018-02-12 11:05:00', 'UTC')
]);
$this->save($entity);
```

### What happened
Saved dates are both same *2018-02-12 11:05:00* (in DB)

### What you expected to happen
I expect *Time object timezone should be changed to UTC before saving*, so times in DB will be in UTC
`date1 = '2018-02-12 06:05:00'`
`date2 = '2018-02-12 11:05:00'`

I have UTC both in app.php (Datasources.default.timezone) and bootstrap.php (date_default_timezone_set('UTC')), on fresh app install.

setTimezone() will do what am expecting, but I think this is not expected behavior.
```php
$entity = $this->newEntity([
    'date1' => (new Time('2018-02-12 11:05:00', '+05:00'))->setTimezone('UTC'),
    'date2' => new Time('2018-02-12 11:05:00', 'UTC')
]);
$this->save($entity);
```
saves correct *2018-02-12 06:05:00* for date1.

ADmad:
Seems like duplicate of #6571

markstory:
The ORM internals don't currently automatically translate timezones for you. You'd need to do this yourself. As an alternative solution you could replace the Datetime type mappings with an implementation that has different `toDatabase` or `marshal` logic. See the following for the current code. https://github.com/cakephp/cakephp/blob/a6c47f15e2dfbb062fcc547d6eecd0a944dc8b17/src/Database/Type/DateTimeType.php#L110-L120

This is likely related to #10877 and #11517. These issues aim to add a hidden timezone input enabling the marshaller to convert request data in the user's timezone to the application's internal default timezone.

samuelszabo:
@ADmad Thanks, I did not found it before. This issue is probably part of #6571, but it needs simplier solution (as I hope).

@markstory Thanks, I understand I can fix it my way. But I still consider this (ignoring Timezone) as a bug.

[...]

PR #11780, opened 2018-03-02
Allow changing timezone when generating datetime string for db.

ADmad:
Refs #11712.

This change won't do anything unless a timezone is explicitly set, so it won't cause any backwards compatibility issues or unexpected changes to the datetime values.

### DIFF
--- a/src/Database/Type/DateTimeType.php
+++ b/src/Database/Type/DateTimeType.php
@@ -111,7 +111,7 @@ class DateTimeType extends Type implements TypeInterface
     /**
      * Convert DateTime instance into strings.
      *
-     * @param string|int|\DateTime $value The value to convert.
+     * @param string|int|\DateTime|\DateTimeImmutable $value The value to convert.
      * @param \Cake\Database\Driver $driver The driver instance to convert with.
      * @return string|null
      */
@@ -128,8 +128,9 @@ class DateTimeType extends Type implements TypeInterface
         $format = (array)$this->_format;
 
         if ($this->dbTimezone !== null
-            && $this->dbTimezone !== $value->getTimezone()->getName()) {
-            $value->setTimezone(new DateTimeZone($this->dbTimezone));
+            && $this->dbTimezone !== $value->getTimezone()->getName()
+        ) {
+            $value = $value->setTimezone(new DateTimeZone($this->dbTimezone));
         }
 
         return $value->format(array_shift($format));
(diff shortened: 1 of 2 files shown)


markstory:
This looks like it will handle converting from PHP's default timezone into the database timezone (should they be different). I thought part of #11712 was to allow end users to pick times in *their* timezone and have CakePHP automatically convert that timezone into the application's timezone.

As a contrived illustration, intended to show problems:

1. User timezone is `America/New_York`
2. Application Timezone is `UTC`
3. Database timezone is `Asia/Kolkata`

This pull request

ADmad:
To my understanding #11712 is about solving the conversion between 2 & 3. The PR is based on the diff I showed in comment and @samuelszabo seemed to be happy with the solution.

src/Database/Type/DateTimeType.php
```diff
@@ -119,9 +127,31 @@ public function toDatabase($value, Driver $driver)
 
         $format = (array)$this->_format;
 
+        if ($this->dbTimezone !== null
+            && $this->dbTimezone !== $value->getTimezone()->getName()) {
+            $value->setTimezone(new DateTimeZone($this->dbTimezone));
```
markstory: I think this needs to be `$value = $value->setTimezone(...)` to work with immutable datetime objects.

jeremyharris: Beware of `$value` in php 7.2 :D if that's still a thing

markstory: It has only been a problem in loops.

ADmad: I'll fix this today.

src/Database/Type/DateTimeType.php
```diff
@@ -119,9 +127,32 @@ public function toDatabase($value, Driver $driver)
 
         $format = (array)$this->_format;
 
+        if ($this->dbTimezone !== null
+            && $this->dbTimezone !== $value->getTimezone()->getName()
+        ) {
+            $value = $value->setTimezone(new DateTimeZone($this->dbTimezone));
```
ADmad: I am wondering if instead of creating a `DatetimeZone` instance here on each call, it would be more efficient to set `$_dbTimezone` property itself to `DatetimeZone` instance instead of string.

markstory: We could create a single timezone instance when the timezone is first setup. That would save multiple objects from being created.

ADmad:
@markstory Done

PR #11781, opened 2018-03-02
Use rule alias as rule name by default.

### LINKED ISSUE #11778: Improvement: Unneeded redundancy in validator
This is a (multiple allowed):

* [] bug
* [x] enhancement

* CakePHP Version: 3.5.x / dev-master

### What you did
```php
        $validator
            ->scalar('status')
            ->add('status', 'validateStatus', [
                'provider' => 'table',
            ]);
```

You need to explicitly define the rule as copy of the name again:
```php
        $validator
            ->scalar('status')
            ->add('status', 'validateStatus', [
                'rule' => 'validateStatus', // HERE
                'provider' => 'table',
            ]);
```

The first example should work: The rule name should inherit from the 2nd argument rule name if not specific otherwise (null).

ADmad:
Closes #11778.

### DIFF
--- a/src/Validation/Validator.php
+++ b/src/Validation/Validator.php
@@ -397,7 +397,7 @@ class Validator implements ArrayAccess, IteratorAggregate, Countable
         }
 
         foreach ($rules as $name => $rule) {
-            $field->add($name, $rule);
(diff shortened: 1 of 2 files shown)


dereuromark:
Thanks for picking this up so fast.

Issue #11745, opened 2018-02-22
Simple Query::where() syntax


Erwane:
* [ ] bug
* [x] enhancement
* [x] feature-discussion (RFC)

* CakePHP Version: 3.x

A discussion about `where` syntax in ORM

Current syntax
```php
$articles->find()->where(['id' => 3]);
```

could be
```php
$articles->find()->where('id', 3);
```

My "problem" is in the `['field' => 'value']` syntax and "heavy" to write and read

dereuromark:
While you find it heavy to write, most find it consistent.
What about all the other methods, which also accept an array?

It will also increase the cycl. complexity of the methods' internals.
More if/else that can make the method buggy or introduce unexpected side-effects.

Last but not least, in light of PHP7.1 we try to actually go to a more strict API in general, to add more typehints and not less. So what you are suggesting means we cannot array typehint here which is a step backwords, rather than forward. Methods should always have as least possible types  (besides nullable) as possible.
It would have to be its own method name (e.g. whereSingle() or sth), but is that really worth it?

I am :-1: as the current consistent API makes it actually easier to work with as you know what to expect.

inoas:
Well `where()` can do a lot more like having expressions on the left side.
However what about `whereColumn($column, $valueOrExpression)` or `whereField($field, $valueOrExpression)` or `whereProperty($property, $valueOrEpxression)` (all the same just different name). The column/field/property must then be an aliased field and exactly that, nothing else.

```php
$query->whereField('User.username', 'mark')
```

The query builder can easily check if those fields exist without any magic and can even do early type comparison (e.g. if casting the 2nd param is lossless or already the correct type).

markstory:
We could also have `addWhere()` to support this new argument set. I would prefer to avoid  overloading the parameters of `where()` further, but agree that in simple cases the current syntax is a bit clunky.

dereuromark:
I strongly advise against overloading `andWhere($conditions, $types = [])` as well.
This introduces a code smell.
If anything, I would agree that `whereField()` could be done, but do we then also need yet another `andWhereProperty()`?
If we start this, then people also would want `orWhere()` in a similar fashion. 
Same for `group()` (`groupField()`), having, andHaving, orHaving etc.

`order()` already exists by the way (`orderAsc($field...`, `orderDesc($field...`).

inoas:
[...]

Issue #7092, opened 2015-07-22
ORM: Unexpected behavior with 'WHERE NOT IN' and empty array


chinpei215:
```php
$articles->find()->where(function($exp,$q){
    return $exp->notIn('id', []);
});

# WHERE 1 != 1
```
So this query returns no rows.
This may be a feature, but this is the opposite of what I expected.

In CakePHP 2.x:
```php
$Article->find('all', [
	'conditions' => [
		'id !=' => [],
	]
]);

# WHERE `id` IS NOT NULL
```

In http://www.doctrine-project.org/jira/browse/DDC-3063 , they are talking about a same issue.

markstory:
Are you expecting that `notIn('id', [])` would generate `id IS NOT NULL`? We could also generate `WHERE 1 = 1` I guess as is a similar expression in many cases.

chinpei215:
I think that it would be better to generate <code>WHERE 1 = 1</code>.

lorenzo:
:-1: this is by design. Imagine using 1=1 on a delete query when you fail to validate or have a typo in a variable name. A SQL error is much more desirable for me than losing data

markstory:
I am not very comfortable with the ORM generating `WHERE 1=1` clauses on what can very easily be a typo/mistake.

chinpei215:
If so, I think that you should throw an exception. <code>WHERE 1 != 1</code> isn't a SQL error:
```php
$subQuery = $articles->find()->select(['id'])->where([
	'something NOT IN' => [],
]);
$query = $comments->query()->delete()->where([
	'article_id NOT IN' => $subQuery,
]);
```

markstory:
In the sub-query case I don't think you'd ever end up with `1 != 1` though.

dereuromark:
Sounds to me we should go the 2.x way here and output `WHERE `id` IS NOT NULL`.

AD7six:
> WHERE 1 != 1 isn't a SQL error:

I don't think that's what was meant. It's a clause guaranteed to return 0 rows. It's also, should you find it in your query log, an indicator of doing something wrong (why am I getting this clause? oh, there's a typo on that variable name generating an empty array etc.).

AD7six:
> Sounds to me we should go the 2.x way here and output WHERE id IS NOT NULL.

I disagree, as that's almost always equivalent to putting no clause at all.

dereuromark:
[...]

PR #11782, opened 2018-03-02
Add whereIn() and whereNotIn() and allow to behave as intuitively expected

dereuromark:
Refs https://github.com/cakephp/cakephp/issues/11745

Two important things:

a) Adds standalone wrapper to not have to add the cumbersome IN / NOT IN (since 3.x) strings into the field key when there "could" be an array.

b) Adds support for the intuitive behavior of empty result for empty array (instead of the hard and problematic exception) - and all results for empty array on NOT.

Protecting the developer from himself here is not helpful. Often times these exceptions are not happening until somewhere someday it does happen - and then it is a big deal. Which it shouldn't have been.
Additionally, it is better to not force the developer to add additional logic all over the place, including controllers - as I have seen and experienced it myself. If (!empty) then add on top, otherwise not. Makes the whole RAD thing a bit less RAD and CRUD.

With these wrappers a dev has the choice at least to some extend.

### DIFF
--- a/src/Database/Query.php
+++ b/src/Database/Query.php
@@ -906,7 +906,7 @@ class Query implements ExpressionInterface, IteratorAggregate
      * @param array $options Options
      * @return $this
      */
-    public function whereIn($field, array $values, array $options = [])
+    public function whereInArray($field, array $values, array $options = [])
     {
         $options += [
             'types' => [],
@@ -933,7 +933,7 @@ class Query implements ExpressionInterface, IteratorAggregate
      * @param array $options Options
      * @return $this
      */
-    public function whereNotIn($field, array $values, array $options = [])
+    public function whereNotInArray($field, array $values, array $options = [])
     {
         $options += [
             'types' => [],
@@ -941,7 +941,7 @@ class Query implements ExpressionInterface, IteratorAggregate
         ];
 
         if ($options['allowEmpty'] && !$values) {
-            return $this->where('1=1');
+            return $this->where([$field . ' IS NOT' => null]);
         }
 
         return $this->where([$field . ' NOT IN' => $values], $options['types']);
--- a/src/Database/Query.php
+++ b/src/Database/Query.php
@@ -893,16 +893,18 @@ class Query implements ExpressionInterface, IteratorAggregate
      * Adds an IN condition or set of conditions to be used in the WHERE clause for this
      * query.
      *
-     * This method does allow empty inputs in contrast to where().
+     * This method does allow empty inputs in contrast to where() if you set
+     * $allowEmpty to true.
      * Be careful about using it without proper sanity checks.
      *
      * @param string $field Field
      * @param array $values Array of values
+     * @param bool $allowEmpty Allow empty array
      * @return $this
      */
-    public function whereIn($field, array $values)
+    public function whereIn($field, array $values, $allowEmpty = false)
     {
-        if (!$values) {
+        if ($allowEmpty && !$values) {
             return $this->where('1=0');
         }
 
@@ -913,16 +915,18 @@ class Query implements ExpressionInterface, IteratorAggregate
      * Adds a NOT IN condition or set of conditions to be used in the WHERE clause for this
      * query.
      *
-     * This method does allow empty inputs in contrast to where().
+     * This method does allow empty inputs in contrast to where() if you set
+     * $allowEmpty to true.
      * Be careful about using it without proper sanity checks.
      *
      * @param string $field Field
      * @param array $values Array of values
+     * @param bool $allowEmpty Allow empty array
      * @return $this
      */
-    public function whereNotIn($field, array $values)
+    public function whereNotIn($field, array $values, $allowEmpty = false)
     {
-        if (!$values) {
+        if ($allowEmpty && !$values) {
             return $this->where('1=1');
         }
 
--- a/src/Database/Query.php
+++ b/src/Database/Query.php
@@ -890,6 +890,46 @@ class Query implements ExpressionInterface, IteratorAggregate
     }
 
     /**
+     * Adds an IN condition or set of conditions to be used in the WHERE clause for this
+     * query.
+     *
+     * This method does allow empty inputs in contrast to where().
+     * Be careful about using it without proper sanity checks.
+     *
+     * @param string $field Field
+     * @param array $values Array of values
+     * @return $this
+     */
+    public function whereIn($field, array $values)
+    {
+        if (!$values) {
+            return $this->where('1=0');
+        }
+
+        return $this->where([$field . ' IN' => $values]);
+    }
+
+    /**
+     * Adds a NOT IN condition or set of conditions to be used in the WHERE clause for this
+     * query.
+     *
+     * This method does allow empty inputs in contrast to where().
+     * Be careful about using it without proper sanity checks.
+     *
+     * @param string $field Field
+     * @param array $values Array of values
+     * @return $this
+     */
+    public function whereNotIn($field, array $values)
+    {
+        if (!$values) {
+            return $this->where('1=1');
+        }
+
+        return $this->where([$field . ' NOT IN' => $values]);
+    }
+
+    /**
      * Connects any previously defined set of conditions to the provided list
      * using the AND operator. This function accepts the conditions list in the same
      * format as the method `where` does, hence you can use arrays, expression objects
--- a/src/Database/Query.php
+++ b/src/Database/Query.php
@@ -894,21 +894,30 @@ class Query implements ExpressionInterface, IteratorAggregate
      * query.
      *
      * This method does allow empty inputs in contrast to where() if you set
-     * $allowEmpty to true.
+     * 'allowEmpty' to true.
      * Be careful about using it without proper sanity checks.
+     *
+     * Options:
+     * - `types` - Associative array of type names used to bind values to query
+     * - `allowEmpty` - Allow empty array.
      *
      * @param string $field Field
      * @param array $values Array of values
-     * @param bool $allowEmpty Allow empty array
+     * @param array $options Options
      * @return $this
      */
-    public function whereIn($field, array $values, $allowEmpty = false)
+    public function whereIn($field, array $values, array $options = [])
     {
-        if ($allowEmpty && !$values) {
+        $options += [
+            'types' => [],
+            'allowEmpty' => false,
+        ];
+
+        if ($options['allowEmpty'] && !$values) {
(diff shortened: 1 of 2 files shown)


src/Database/Query.php
```diff
@@ -889,6 +889,46 @@ public function where($conditions = null, $types = [], $overwrite = false)
...
+     * This method does allow empty inputs in contrast to where().
+     * Be careful about using it without proper sanity checks.
+     *
+     * @param string $field Field
+     * @param array $values Array of values
+     * @return $this
+     */
+    public function whereIn($field, array $values)
+    {
+        if (!$values) {
```
inoas: so `''`, `null`, `0` is not valid? How come?

`WHERE foo IN (1, 2, 3, NULL)` sounds reasonable to me so would `WHERE foo IN (NULL)`?

inoas (changes requested):

src/Database/Query.php
dereuromark: They are all valid as long as they are passed as array - this uses the current internal behavior, so no special things here.

inoas:
My expectation for `whereIn($field, $values)` is to behave **exactly** like `where("$field . ' IN' => $values)` and that's the least error prone and consistent interface IMHO.

src/Database/Query.php
inoas: So why is passing in `null` invalid, why is passing in `int(0)` invalid?

src/Database/Query.php
```diff
@@ -889,6 +889,46 @@ public function where($conditions = null, $types = [], $overwrite = false)
...
+     * Be careful about using it without proper sanity checks.
+     *
+     * @param string $field Field
+     * @param array $values Array of values
+     * @return $this
+     */
+    public function whereNotIn($field, array $values)
+    {
+        if (!$values) {
+            return $this->where('1=1');
```
chinpei215: I think it should be based on the previous discussion (#7092).

In the past, this `'NOT IN' => []` returned `1 != 1` and it was an intended behavior. Since it was not my expectation, I wanted to get exceptions in such cases. I thought it was better than `1 != 1`.

But if you think `'NOT IN' => []` should return `1 = 1`, I think we should bring up #7092 again and we shouldn't add a wrapper to avoid discussion. Since `1 = 1` is my first expectation, I wouldn't be opposed to return it. `IS NOT NULL` may be better though.

xaphalanx:
dereuromark, This pull request is over 9000.

Thank you.

src/Database/Query.php
markstory: @inoas Wouldn't the type hint prevent null/false/0 from entering the method?

src/Database/Query.php
markstory: Generating a `1=1` condition on empty data is a dangerous default. Passing an empty list could be a mistake that we should protect against to be consistent with our previous decisions as @chinpei215 pointed out.

src/Database/Query.php
inoas: So I miss read the check and its meaning is?:

```php
if ($values === []) {
```

Can that be written then now that we got a strict type hint?

dereuromark: Why? Same thing. Your comments are misleading to me.

src/Database/Query.php
dereuromark: OK, I hear you @markstory 
But my experience tells me exactly the opposite:
The behavior is intended like this in pretty much every case, and the result is usually the "blowup" of the application in a later state where the condition is not met anymore then for some reason.
Sure, less problematic as applying a query to too much rows.

I adjusted the code to opt-in the intuitive behavior here for RAD, keeping the default consistent to the strict way it is in where().

I would much rather have this code:
```
$query = $this->find()
    ->where(['x' => y])
    ->whereIn('field', $values, true)
    ->orderDesc('x')
    ->all();
```
Then
```
if (!$values) {
    return [];
}
$query = $this->find()
    ->where(['x' => y])
    ->whereIn('field', $values)
    ->orderDesc('x')
    ->all();
```
Especially when you start nesting with OR etc, as the complexity of the "false" case because super error prone.

When the performance is not as relevant as readability/correctness of the code, I want to not have to add application logic on top everywhere.
Using the allowEmpty option gives a dev that deliberate choice now - nothing to protect him anymore then?

src/Database/Query.php
inoas: Well if it had been the `=== []` I'd not had miss-understood the code. Now to understand how the function runs you have to look at the signature and typehint.

Anyway, seems you like all the implicit weak castings. It works.

src/Database/Query.php
inoas: Could we have `Query::ALLOW_EMPTY` (true) and `Query::DISALLOW_EMPTY` (false)?

dereuromark:
@chinpei215 You can add your changes on top, if you want to, regarding the IS NOT NULL part you mentioned in the other ticket.

src/Database/Query.php
```diff
@@ -889,6 +889,50 @@ public function where($conditions = null, $types = [], $overwrite = false)
...
+     *
+     * @param string $field Field
+     * @param array $values Array of values
+     * @param bool $allowEmpty Allow empty array
+     * @return $this
+     */
+    public function whereIn($field, array $values, $allowEmpty = false)
+    {
+        if ($allowEmpty && !$values) {
+            return $this->where('1=0');
```
lorenzo: We had this in the past. It resulted in a bug that deleted the whole table of a user. The bug can be triggered when you pass an empty array by mistake and the condition you are using is part of a NOT. The negation of this imposible condition results in matching all rows

lorenzo (changes requested):
I would remove the magic conditions

src/Database/Query.php
lorenzo: I see, what you mean. The flag makes sense to me.

lorenzo (approved):

src/Database/Query.php
dereuromark: I have seen queries like
```
->where(['tag_id IN' => Hash::extract($result, '{n}.id')])
```
that suddenly broke the whole application "just because" - it was not justified and kind of goes against the ideas of the framework making certain things simpler and less buggy :)

I see the point more for the NOT IN part, but for 99% of all IN this should have never been done - and I have so many 2.x => 3.x apps broken silently along the way because of this. Not a single one exposed to the security concerns theoretically possible.

lorenzo: I was once exposed to that bug myself, the application sent a lot of emails to people that should not have been on that list. It was so embarrassing and time consuming to fix the data corruption that puts on perspective that 1% of the cases.

An exception is more tolerable for me than something I cannot revert after the fact.

dereuromark: Indeed - that is why it is good to have the option for both ways, defaulting to the exception case where it makes sense, and allowing the opt-in where additional checking is not necessary as nothing bad can happen here.

lorenzo: I agree with you

dereuromark:
Are we good to go here?

lorenzo:
I'm just thinking that this is missing the `types` parameter. Maybe make the last parameter `$options` and the options accept both `allowEmpty` and `types` ?

dereuromark:
That will most likely kill the usability of the flag for empty allowance, since this was one of the primary concerns besides the IS/ISNOT addition.
But sounds like we don't have any other choice then. Having 3 args is probably not much better either.

src/Database/Query.php
```diff
@@ -889,6 +889,50 @@ public function where($conditions = null, $types = [], $overwrite = false)
...
+     * This method does allow empty inputs in contrast to where() if you set
+     * $allowEmpty to true.
+     * Be careful about using it without proper sanity checks.
+     *
+     * @param string $field Field
+     * @param array $values Array of values
+     * @param bool $allowEmpty Allow empty array
+     * @return $this
+     */
+    public function whereIn($field, array $values, $allowEmpty = false)
```
ADmad: By type hinting `$values` as array you are prevent this method to be used for cases where value is a string. Is that intentional?  Because for e.g. `->where(['title IN' => 'foo'])` is valid usage which generate SQL fragment `title in ('foo')`.

lorenzo: I would say it is intentional and desired, a string is not a list of values.

dereuromark: Casting beforehand is eaay. And this way no accidental usage of wrong types.

ADmad: Works for me. Just wanted to confirm.

chinpei215:
I think it would be better to change the method name `whereIn()` to `whereInArray()`, as it doesn't support subquery though SQL does support it.
```php
//For `user_id IN (SELECT id FROM users WHERE ...)
$query->whereIn('user_id', $subQuery); // Error
```

And, as I said above, it would be better to change `1=0` to `IS NOT NULL` in `whereNotIn()`. I can open another pull request for that if you want, but I tell you the reason why `IS NOT NULL` is better. If you have a table like the follwoing:

| field |
|-------|
| 1     |
| 2     |
| 3     |
| NULL  |

`whereNotIn('field', [1])` will return 2, 3 and will not reutrn NULL. However `whereNotIn('field', [])` will return also NULL. This behavior is inconsistant, as NULL is not in ether `[1]` or `[]`.

dereuromark:
Yes you can also modify this pull request.

chinpei215:
OK. I will add more commits to this pull reqeuest as soon as possible.

inoas:
@chinpei215 what about `whereNotInList()` `whereInList()` as Array is less specific (could be multi dimensional) and also longer and inList/notInList maps pretty much to our validatior names too?

dereuromark:
I agree that we shouldn't leak the specific PHP type of the "list" type here.

chinpei215:
I am fine to change InArray to InList if people prefer the name. But note that `IN` does support multi demension array theoretically:
```
SELECT (1,2) IN ( (1,2), (3,4) );
```

inoas:
@chinpei215 well SQL does but does `whereInArray($identifier, $values)`? Would $identifier be a value or string? I'd assume it can only be a single field identifier?

dereuromark:
Let's keep the same simple :)

chinpei215:
Postgres support ARRAY. It can be referred by a single identifier.

chinpei215:
@dereuromark My intention was "What PHP data type can be pass to the method". So I name the method `whereInArray()`. But if you prefer whereInList() over whereInArray(), please rename those methods again. I am fine with either.

saeideng:
`*list()` :+1:

PR #11784, opened 2018-03-04
selectAllExcept() pull request. #11025 Select columns from a table exc…

T0mats0:
This is a pull request for Add selectAllExcept method #11025.

I attempted to incorporate the suggestions and discussions made previously. 

Lorenzo: first arg is a Table.
Story: provide tests for multiple calls.

If you call this method more than once someone might expect them to stack but the way this PR is done they don't stack.


src/ORM/Query.php
```diff
@@ -194,6 +195,30 @@ public function select($fields = [], $overwrite = false)
...
+     * @param bool $overwrite Whether to reset/remove previous selected fields
+     * @return Query
+     */
+    public function selectAllExcept($table, array $excludedFields, $overwrite = false)
+    {
+        if ($table instanceof Association) {
+            $table = $table->getTarget();
+        }
+        $aliasedFields = [];
+        if ($table instanceof Table) {
```
inoas: Maybe if the target is not a Table throw an Exception instead?

src/ORM/Query.php
```diff
@@ -194,6 +195,30 @@ public function select($fields = [], $overwrite = false)
...
+            $fields = array_diff($table->getSchema()->columns(), $excludedFields);
+            $aliasedFields = $this->aliasFields($fields);
+        }
+
+        return $this->select($aliasedFields, $overwrite);
```
inoas: I don't see tests that toggle around overwrite, do you think adding them might be helpful?

inoas (changes requested):
$table should really be an instance of Table (or Association, and then table via getTarget) anything else I'd expect an Exception.

src/ORM/Query.php
```diff
@@ -194,6 +195,30 @@ public function select($fields = [], $overwrite = false)
         return parent::select($fields, $overwrite);
     }
 
+    /**
+     * Select all the columns associated to $table except the $excludedFields.
+     * Excluded fields should not be aliased names.
+     *
+     *
+     * @param Table|Association $table The table to use to get an array of columns
```
dereuromark: Classes must be FQCN here in doc blocks.

src/ORM/Query.php
dereuromark: Does it have to be a concrete table? Don't we usually try to program against interfaces for SOLID?

ADmad: Only `Table` class has `getSchema()` method so checking for `RepositoryInterface()` isn't feasible.

This also makes me wonder if it's right to have a method on `Query` with takes an `Association` or `Table` as argument.

inoas: Maybe it should take a some Interface that both implement?

xaphalanx: Lorenzo recommended that the first argument be Table so there would be no confusion about what list of columns the user was selecting. (See original thread https://github.com/cakephp/cakephp/pull/11025).  

For now, I've left the parameters alone.  We could type hint Table and make the caller know to caller know that they have to call getTarget() on their association. The good news is the doc blocks will tell me what to pass.

It now throws an exception if something other than a table or association is passed.
Updated documentation and tests as requested.

xaphalanx:
Lorenzo recommended that the first argument be Table so there would be no confusion about what list of columns the user was selecting. (See original thread https://github.com/cakephp/cakephp/pull/11025).  

For now, I've left the parameters alone.  We could type hint Table and make the caller know to caller know that they have to call getTarget() on their association. The good news is the doc blocks will tell me what to pass.

It now throws an exception if something other than a table or association is passed.
Updated documentation and tests as requested.

inoas (approved):

tests/TestCase/ORM/QueryTest.php
```diff
@@ -3548,4 +3562,142 @@ public function testNotMatchingNested()
...
+    /**
+     * Test what happens if you call selectAllExcept() more
+     * than once.
+     *
+     * @return void
+     */
+    public function testSelectAllExceptWithMulitpleCalls()
+    {
+        $table = $this->getTableLocator()->get('Articles');
+        $this->getTableLocator()->get('Articles');
```
markstory: This line looks unnecessary.

tests/TestCase/ORM/QueryTest.php
```diff
@@ -3548,4 +3562,142 @@ public function testNotMatchingNested()
...
+    /**
+     * Test that the excluded fields are not included
+     * in the final query result.
+     *
+     * @return void
+     */
+    public function testSelectAllExceptWithContains()
+    {
+        $table = $this->getTableLocator()->get('Articles');
+        $this->getTableLocator()->get('Articles');
```
markstory: This line looks unnecessary.

tests/TestCase/ORM/QueryTest.php
```diff
@@ -3548,4 +3562,142 @@ public function testNotMatchingNested()
...
+        $result = $table
+            ->find()
+            ->selectAllExcept($table, ['body'])
+            ->selectAllExcept($table, ['published']);
+        $selectedFields = $result->clause('select');
+        $expected = [
+            'Articles__id' => 'Articles.id',
+            'Articles__author_id' => 'Articles.author_id',
+            'Articles__title' => 'Articles.title',
+            'Articles__published' => 'Articles.published',
```
markstory: Shouldn't published be excluded as it was in the `selectAllExcept` call on 3650?

tests/TestCase/ORM/QueryTest.php
```diff
@@ -3548,4 +3562,142 @@ public function testNotMatchingNested()
...
+     * @return void
+     */
+    public function testSelectAllExceptThrowsInvalidArgument()
+    {
+        $table = $this->getTableLocator()->get('Articles');
+        try {
+            $table
+                ->find()
+                ->selectAllExcept([], ['body']);
+        } catch (\InvalidArgumentException $e) {
```
markstory: We generally use `expectException(InvalidArgumentException::class)`

tests/TestCase/ORM/QueryTest.php
xaphalanx: These tests show what would happen on multiple calls. As far as I know things go into $this->_parts but don't come out unless you reset/clear it. So on the first call all the columns but the excluded list go into the query. On a second call you can't then remove something from the query parts without resetting. So these tests show that behavior. This test also shows the "reset" behavior. Does that answer your question?

tests/TestCase/ORM/QueryTest.php
xaphalanx: Thanks for the pro-tip.

tests/TestCase/ORM/QueryTest.php
markstory: Ok. Could we document how multiple calls interact in the API doc block? Having that should help people understand the method better.

xaphalanx: Will do. Thanks for the feedback.

Issue #11783, opened 2018-03-03
3.6: Fatal error loading plugins


dereuromark:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: latest master

### What you did
Using 3.6 pre-beta with a 3.5 app.
And the load command as it always has been:

Plugin::load('Tools', ['bootstrap' => true]);

### What happened

> Fatal error: Cannot redeclare endsWith() (previously declared in /home/vagrant/Apps/x.local/vendor/dereuromark/cakephp-tools/config/bootstrap.php:323) in /home/vagrant/Apps/x.local/vendor/dereuromark/cakephp-tools/config/bootstrap.php on line 323

Seems like it tries to load the plugin now twice?

### What you expected to happen
Still BC - only loading the plugins once (the old way).

markstory:
Are you running tests, or is this part of a web request?

dereuromark:
Normal webrequest after adding 3.next branch to test compatibility of my 3.5 app with 3.6.

markstory:
Ok. Looks like this is caused by the `Plugin::load()` including the bootstrap file, and also the new application hook also running the bootstrap logic. I'll get this fixed up.

PR #11785, opened 2018-03-04
Fix duplicate plugin bootstrapping.

markstory:
Use the presence of an application class to defer plugin bootstrapping. Defering this logic pushes it into the pluginBootstrap() application hook.

The hook method has one caveat that plugins will need to conditionally define constants and functions as plugins can now be bootstrapped multiple times in a single testsuite run.

Refs #11783


dereuromark (approved):
Works

PR #11786, opened 2018-03-04
Adding count() and countKeys() to Collection

lorenzo:
Not having these is honestly surprising to many. Instead of trying to be always correct
by not having the method, provide it as an "advanced" feature with strings attached.

I've added a long section of warnings on when `count()` will be maybe not be what you need.

### DIFF
--- a/src/Collection/Collection.php
+++ b/src/Collection/Collection.php
@@ -73,17 +73,29 @@ class Collection extends IteratorIterator implements CollectionInterface, Serial
     }
 
     /**
-     * Throws an exception.
+     * {@inheritDoc}
      *
-     * Issuing a count on a Collection can have many side effects, some making the
-     * Collection unusable after the count operation.
-     *
-     * @return void
-     * @throws \LogicException
+     * @return int
      */
     public function count()
     {
-        throw new LogicException('You cannot issue a count on a Collection.');
+        $traversable = $this->optimizeUnwrap();
+
+        if (is_array($traversable)) {
+            return count($traversable);
+        }
+
+        return iterator_count($traversable);
+    }
+
+    /**
+     * {@inheritDoc}
+     *
+     * @return int
+     */
+    public function countKeys()
+    {
+        return count($this->toArray());
     }
 
     /**
@@ -95,7 +107,7 @@ class Collection extends IteratorIterator implements CollectionInterface, Serial
     public function __debugInfo()
     {
         return [
-            'count' => iterator_count($this),
+            'count' => $this->count(),
         ];
     }
 }
--- a/tests/TestCase/Collection/CollectionTest.php
+++ b/tests/TestCase/Collection/CollectionTest.php
@@ -1007,16 +1007,29 @@ class CollectionTest extends TestCase
     }
 
     /**
-     * Tests that issuing a count will throw an exception
+     * Tests that Count returns the number of elements
      *
+     * @dataProvider simpleProvider
      * @return void
      */
-    public function testCollectionCount()
+    public function testCollectionCount($list)
     {
-        $this->expectException(\LogicException::class);
-        $data = [1, 2, 3, 4];
-        $collection = new Collection($data);
-        $collection->count();
+        $list = (new Collection($list))->buffered();
+        $collection = new Collection($list);
+        $this->assertEquals(8, $collection->append($list)->count());
+    }
+
+    /**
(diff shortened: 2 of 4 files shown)


src/Collection/Collection.php
```diff
@@ -73,17 +73,29 @@ public function unserialize($collection)
...
     public function count()
     {
-        throw new LogicException('You cannot issue a count on a Collection.');
+        $traversable = $this->optimizeUnwrap();
+
+        if (is_array($traversable)) {
+            return count($traversable);
+        }
+
+        return iterator_count($traversable);
```
lorenzo: I'm duplicating the implementation here for the sole reason that I want the PHP docs to be visible in the api site.

src/Collection/Collection.php
```diff
@@ -73,17 +73,29 @@ public function unserialize($collection)
...
+    }
+
+    /**
+     * {@inheritDoc}
+     *
+     * @return int
+     */
+    public function countKeys()
```
dereuromark: Should this maybe be `keyCount()`? as countKeys sounds more like exec and void command than a noun.

lorenzo: not sure, `countKeys` sounds like the right choice to me.

tests/TestCase/Collection/CollectionTest.php
```diff
@@ -1007,16 +1007,29 @@ public function testInvalidConstructorArgument()
...
-    public function testCollectionCount()
+    public function testCollectionCount($list)
     {
-        $this->expectException(\LogicException::class);
-        $data = [1, 2, 3, 4];
-        $collection = new Collection($data);
-        $collection->count();
+        $list = (new Collection($list))->buffered();
+        $collection = new Collection($list);
+        $this->assertEquals(8, $collection->append($list)->count());
```
markstory: Should there be tests for any of the sharp edges this method has?

lorenzo: any case you arrnintested in? The current test shows how a collection with only 4 unique keys can have 8 elements. That odd case is covered

markstory: Good point. On second thought the other cases are likely not worth covering with tests.

src/Collection/Collection.php
```diff
@@ -73,17 +73,29 @@ public function unserialize($collection)
...
-     * Collection unusable after the count operation.
-     *
-     * @return void
-     * @throws \LogicException
+     * @return int
      */
     public function count()
     {
-        throw new LogicException('You cannot issue a count on a Collection.');
+        $traversable = $this->optimizeUnwrap();
```
chinpei215: Why don't you do the same operation as [ResultSetDecorator](https://github.com/cakephp/cakephp/blob/424f2eb1f292a64bd53dfc29e478dd995eebe38d/src/Datasource/ResultSetDecorator.php#L36-L43)?

lorenzo: I know ResultSetDecorator In all of the cases is wrapping a countable object. Here I do not know that. But here I have the functions to dive down to an array.

I could change the decorator to be similar to this.

src/Collection/Collection.php
dereuromark: My thinking was:

count() is just a short form of getCount() and means the noun(), like getCollection() alias collection() etc.
So logically, the count for keys would be getKeyCount() or just keyCount().

But no big deal.

JoshuaLuckers:
It might be a good idea to add the Countable interface.

src/Collection/Collection.php
lorenzo: That makes sense. I mostly wanted to keep the names close when autocompleting

lorenzo:
@JoshuaLuckers I'm going to hold off on that. At least the way it is now people with an editor having autocompletion will see the warnings. With the countable interface they will not

src/Collection/Collection.php
chinpei215: But if the innter iterator is a Countable, why don't you believe the return value of count()? I think the problematic iterator is a FilterIterator here. FilterIterator would need to call iterator_count() or need to count array. But I think other Collection classes can call Countable::count().

src/Collection/Collection.php
dereuromark: Good autocomplete will also find it when you start typing "key" IMO :)

PR #11788, opened 2018-03-05
Fixing batch casting

lorenzo:
I made a last minute mistake in my previous PR.

Also reverted the coercion casting, it was not actually faster as I originally measured. Thanks @chinpei215 for checking that.

### DIFF
--- a/src/Database/FieldTypeConverter.php
+++ b/src/Database/FieldTypeConverter.php
@@ -82,8 +82,8 @@ class FieldTypeConverter
             // the interface. Users can implement the TypeInterface instead to have
             // access to this feature.
             $batchingType = $type instanceof BatchCastingInterface &&
-                $type instanceof Type &&
-                strpos(get_class($type), 'Cake\Database\Type') === false;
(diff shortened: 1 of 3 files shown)


src/Database/FieldTypeConverter.php
```diff
@@ -82,8 +82,8 @@ public function __construct(TypeMap $typeMap, Driver $driver)
             // the interface. Users can implement the TypeInterface instead to have
             // access to this feature.
             $batchingType = $type instanceof BatchCastingInterface &&
-                $type instanceof Type &&
-                strpos(get_class($type), 'Cake\Database\Type') === false;
+                !($type instanceof Type &&
+                strpos(get_class($type), 'Cake\Database\Type') === false);
```
dereuromark: Should we use the actual ::class instead of 'Cake\Database\Type' string?

lorenzo: no, I just want to check that the class is not inside the cake's namespace. Not checking for any class in particular.

Issue #11536, opened 2017-12-12
3.5.6 Client::isOk() vs Client::is2xx()?


inoas:
This is a (multiple allowed):

* [x] enhancement
* [x] feature-discussion (RFC)

* CakePHP Version: 3.5.6

### What you did
Read the release notes of 3.5.8

### What happened
* Realized isOk() does not do what it suggests it does.
* All 2xx are OK by definition. They are "success" codes. I won't digg into the RFC now but here the WP article: https://en.wikipedia.org/wiki/List_of_HTTP_status_codes#2xx_Success
* However... only a few are considered ok: https://github.com/cakephp/cakephp/blob/master/src/Http/Client/Message.php#L60
* aka: 200, 201, 202, 203 and 204
* but 205, 206, 207, 208, and 226 are OK, too

### What you expected to happen
isOk() to return true for all 2xx codes.

### What do you suggest

* isOK() should return true for all 2xx codes. If this is not deemed BC compatible:
* is2xx() could be implemented which returns true for all 2xx codes, even future once (by checking the first digit of the return status code === 2)

dereuromark:
Refs https://github.com/cakephp/cakephp/pull/11521

Side-note: Widen the range as per definition via bugfix can be considered BC, only closing it would not be BC. So it is more the issue of what codes are going to be included in isOk().
I agree that all 2xx are usually meant to be that.

inoas:
Personally I am fine if isOk returns true for the 2xx range. But I am not so sure about BC... real world apps just test isOk => return... 208 => do that, 226 => do something else.

markstory:
Do you frequently use 205-226? While they could fall under 'ok' I've personally never seen them actually used, and would prefer to keep them out of the common paths until there is a good case to include them. Adding 203 & 204 in #11521 made sense as a number of APIs use 204 for DELETE and PUT responses.

inoas:
The thing is that I am convinced that it is not up to us to decide but the IETF HTTP working group and that we should follow specs/standards as closely as possible to make life easy and with less unexpected software/library behavior for developers.

Why add exceptions and make stuff harder "by default"?

For us the only question is: Is something that is successful by standard "ok". I think it is.
I can prep the PR to add the full range of 2xx codes in there explicitly or better yet we just test for the first digit, or I can add a method is2xx() that does that and extend the book.

inoas:
Ref:
https://github.com/zendframework/zend-http/blob/master/src/Response.php#L429

[...]

PR #11796, opened 2018-03-06
Add isSuccessful function to Client\Response

Iandenh:
Added the function `isSuccessful` as discussed at #11536 
In line with the `assertResponseSuccess` of `IntegrationTestCase` https://github.com/cakephp/cakephp/blob/master/src/TestSuite/IntegrationTestCase.php#L722

### DIFF
--- a/src/Http/Client/Response.php
+++ b/src/Http/Client/Response.php
@@ -244,6 +244,16 @@ class Response extends Message implements ResponseInterface
     }
 
     /**
+     * Check if the response was in the 2xx/3xx range.
+     *
+     * @return bool
+     */
+    public function isSuccessful()
+    {
+        return $this->code >= 200 && $this->code <= 308;
+    }
+
+    /**
      * Check if the response had a redirect status code.
      *
      * @return bool
--- a/tests/TestCase/Http/Client/ResponseTest.php
+++ b/tests/TestCase/Http/Client/ResponseTest.php
@@ -233,6 +233,63 @@ XML;
     }
 
     /**
+     * Test isSuccessful()
+     *
+     * @return void
+     */
+    public function testisSuccessful()
+    {
+        $headers = [
+            'HTTP/1.1 100 Continue',
+            'Content-Type: text/html'
+        ];
+        $response = new Response($headers, '');
+        $this->assertFalse($response->isSuccessful());
+
+        $headers = [
+            'HTTP/1.1 200 OK',
+            'Content-Type: text/html'
+        ];
+        $response = new Response($headers, 'ok');
+        $this->assertTrue($response->isSuccessful());
+
+        $headers = [
+            'HTTP/1.1 201 Created',
+            'Content-Type: text/html'
+        ];
+        $response = new Response($headers, 'ok');
+        $this->assertTrue($response->isSuccessful());
+
+        $headers = [
+            'HTTP/1.1 202 Accepted',
+            'Content-Type: text/html'
+        ];
+        $response = new Response($headers, 'ok');
+        $this->assertTrue($response->isSuccessful());
+
+        $headers = [
+            'HTTP/1.1 205 Reset Content',
+            'Content-Type: text/html'
+        ];
+        $response = new Response($headers, 'ok');
+        $this->assertTrue($response->isSuccessful());
+
+        $headers = [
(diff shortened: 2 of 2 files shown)


inoas:
Thank you for taking care of this!

src/Http/Client/Response.php
```diff
@@ -243,6 +243,16 @@ public function isOk()
...
     }
 
+    /**
+     * Check if the response was in the 2xx/3xx range.
+     *
+     * @return bool
+     */
+    public function isSuccessful()
+    {
+        return $this->code >= 200 && $this->code <= 308;
```
inoas: Hm I'd go `return $this->code >= 200 && $this->code <= 399;` to be future proof and work with non standard http codes that still respect the basic ranged (1xx, 2xx, 3xx, 4xx and 5xx).

Also the naming is bad. Clearly only 2xx is "success".
From the RFC: 

```

   2.  Any 2xx (Successful) response to a CONNECT request implies that
       the connection will become a tunnel immediately after the empty
       line that concludes the header fields.  A client MUST ignore any
       Content-Length or Transfer-Encoding header fields received in
       such a message.
```
https://tools.ietf.org/html/rfc7230

So is successful should be:
`return $this->code >= 200 && $this->code <= 299;`

inoas: This method can stay, but should be renamed to something else.

inoas (changes requested):

src/Http/Client/Response.php
dereuromark: @ionas Please see the discussion in the original ticket.
The method name is fine IMO.

dereuromark: As long as https://github.com/cakephp/cakephp/blob/master/src/TestSuite/IntegrationTestCase.php#L727 is not adjusted either we can probably keep the current code here as it is.
Having them in sync makes responses and their checks consistent.

inoas: @dereuromark it is not fine, instead it is completely wrong. HTTP clearly defines what is successful and what is not. Why make things work in unexpected ways?

Suggestion: Instead `isOK` could be updated to be 200-399?

dereuromark: As discusses before BC outweighs any such changes or HTTP synchr.
So instead we keep things as all already discussed - and can re-evaluate that in a major jump.
Again: see also assertResponseSuccess()/assertResponseOk()

inoas: Every new, well versed developer, will expect `isSuccessful()` to be 2xx range and without reading the source code bugs (and security issues) will happen.

One thing is not being able to change `isOk()`, the other is making yet another step to be non-standardish and use `isSuccessful`- instead maybe one of these, or we/you can come up with a better name:
* `->is2xx3xx()`
* `->is200to399()`
* `->isSuccessfulOrRedirect()`
* `->isGood()` (I know cheesy)
* `->isOkay()` (and deprecate `isOk()` right away)

dereuromark: isSuccessfulOrRedirect() could work for the 3.x series.

markstory: Aren't redirects a type of 'success'? Having the same term used in same way in the framework is one advantage of leaving the method as is.

tests/TestCase/Http/Client/ResponseTest.php
```diff
@@ -232,6 +232,63 @@ public function testIsOk()
...
 
+    /**
+     * Test isSuccessful()
+     *
+     * @return void
+     */
+    public function testisSuccessful()
+    {
+        $headers = [
+            'HTTP/1.1 100 Continue',
```
markstory: This tests would be more succinct if it used a data provider.

src/Http/Client/Response.php
inoas: Sure they are similar. That's the reason isOk should/could return true on 200-399 (in 4.x)?

markstory: Then we don't need this method.

ADmad: I would prefer not adding any new method and just modify `isOk()` in 4.x based on consensus.

Iandenh:
Okay, so I think the conclusion is to close this PR and change the behavior of isOk() in 4.0.

Issue #11800, opened 2018-03-07
[3.6.0-beta1] Regression in TimestampBehavior


HavokInspiration:
This is a bug.

* CakePHP Version: 3.6.0-beta1
* Platform and Target: PHP 7.1 / MacOS X / No DB.

### What you did

Bind the TimestampBehavior on a table that does not have either one (or none) of the `created` or `modified` fields.
When saving the entity, an exception `Unknown type ""` is thrown by the `Cake\Database\Type::build()` method.

Extract of the stack-trace :

```
⟩ Cake\Database\Type::build
CORE/src/ORM/Behavior/TimestampBehavior.php, line 202

⟩ Cake\ORM\Behavior\TimestampBehavior->_updateField
CORE/src/ORM/Behavior/TimestampBehavior.php, line 108
```

I tested the same code on a last 3.5 release which works fine.

HavokInspiration:
Sorry, misclicked, I will complete the ticket

ADmad:
I think the right thing to do would be throw a meaningful error. Don't see how silently setting value for non-existent field like 3.5 does would be useful.

jeremyharris:
Even though I think it's better to throw the error, this will definitely break apps for people like me who add the behavior to our main table class so all tables inherit it.

HavokInspiration:
Having the error caught earlier to display a better message would indeed be better, but is this kind of behavior not introducing a breaking change in a minor ?

With that being said, if anything, it's helping to do some clean up...

markstory:
> Having the error caught earlier to display a better message would indeed be better, but is this kind of behavior not introducing a breaking change in a minor ?

It feels like a break to me.

markstory:
Pull request open now to fix the break. I'm open to making missing columns an error, but would prefer that error to not be emitted during save, and instead at configuration time.

HavokInspiration:
> I'm open to making missing columns an error, but would prefer that error to not be emitted during save, and instead at configuration time.

That's a very good idea.

PR #11801, opened 2018-03-07
Don't hard fail on missing columns.

markstory:
If timestamp behavior is attached to a table that doesn't have the required columns, it should not hard fail in this way. This failure mode can be percieved as a compatibility break as it happens deep into runtime, and not during configuration/attach time.

Refs #11800


ADmad (approved):

HavokInspiration:
👏 Thanks. I confirm this fixes the issue.

PR #11804, opened 2018-03-08
PSR-16 compatible cache interface

burzum:


src/Cache/Engine/ApcuEngine.php
```diff
@@ -68,15 +75,21 @@ public function write($key, $value)
...
      * @link https://secure.php.net/manual/en/function.apcu-fetch.php
      */
-    public function read($key)
+    public function get($key, $default = null)
     {
         $key = $this->_key($key);
 
-        return apcu_fetch($key);
+        $result = apcu_fetch($key);
+        if ($result === false) {
```
ADmad: I don't think this is correct. The success or failure of `apcu_fetch()` is determined by passing 2nd argument to the method.

http://php.net/manual/en/function.apcu-fetch.php

src/Cache/Engine/WincacheEngine.php
```diff
@@ -70,14 +73,20 @@ public function write($key, $value)
...
      *   has expired, or if there was an error fetching it
      */
-    public function read($key)
+    public function get($key, $default = null)
     {
         $key = $this->_key($key);
 
-        return wincache_ucache_get($key);
+        $result = wincache_ucache_get($key);
+        if ($result === false) {
```
ADmad: Here too cache hit/miss is checked by passing second param to `wincache_ucache_get()`.
http://php.net/manual/en/function.wincache-ucache-get.php

src/Cache/Engine/RedisEngine.php
```diff
@@ -121,47 +119,42 @@ protected function _connect()
...
 
     /**
-     * Read a key from the cache
-     *
-     * @param string $key Identifier for the data
-     * @return mixed The cached data, or false if the data doesn't exist, has expired, or if there was an error fetching it
+     * @inheritDoc
      */
-    public function read($key)
+    public function get($key, $default = null)
```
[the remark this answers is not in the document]
dereuromark: If it is not a partial one, the {} are wrong.

src/Cache/BackwardCompatibilityTrait.php
```diff
@@ -0,0 +1,119 @@
...
+ * CakePHP(tm) : Rapid Development Framework (https://cakephp.org)
+ * Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
+ *
+ * Licensed under The MIT License
+ * For full copyright and license information, please see the LICENSE.txt
+ * Redistributions of files must retain the above copyright notice.
+ *
+ * @copyright     Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
+ * @link          https://cakephp.org CakePHP(tm) Project
+ * @since         1.2.0
```
markstory: This will need to be updated ⛳️

src/Cache/Cache.php
```diff
@@ -362,6 +383,37 @@ public static function read($key, $config = 'default')
...
+     * ```
+     *
+     * @param string $key The unique key of this item in the cache.
+     * @param mixed $default Default value to return if the key does not exist.
+     * @param string $config optional name of the configuration to use. Defaults to 'default'
+     * @return mixed The value of the item from the cache, or $default in case of cache miss.
+     * @throws \Psr\SimpleCache\InvalidArgumentException
+     *   MUST be thrown if the $key string is not a legal value.
+     */
+    public static function get($key, $default = null, $config = 'default')
```
markstory: For the new cache interface, should we consider making the engines implement PSR16, and `Cache` becoming more of a registry/factory/container for cache engines? Right now we can't easily inject caching objects into other things. Would it be 'better' if we used this new interface implementation as an opportunity to make the interface something like

```php
// Get an engine
$cache = Cache::engine('default');
// Use the engine
$cache->set($key, $value, $ttl);
$value = $cache->get($key, $default);
```

I recognize this would be a big departure from what we have today. I think it would let us make engines more standalone and set us up for a DI container in the future.

markstory (commented):
How does PSR6 interact with our current concepts of cache groups?

markstory:
This is looking good @burzum. It is nice that the PSR interfaces are close to what we already have.

src/Cache/Cache.php
burzum: Shouldn't this be `getEngine()`? I thought to implement PSR6 first, then 16 in a follow up PR.

burzum:
@markstory thanks but I could need some help with the engines I have not yet installed and can't install that easy. Doing a lot of commits and waiting for Travis is pretty ugly and will take time.

src/Cache/Cache.php
markstory: `Cache::engine()` only functions as a getter, so it doesn't require a `get` prefix to me.

markstory:
@burzum Which backends do you need help with?

src/Cache/Cache.php
burzum: I think we should use getX() it's more consistent and there is no need if we ever add a getter to refactor it.

burzum:
@markstory all that don't pass yet. :) Also there is one case broken that I don't know how to fix it yet. There is one test that retuns null but now false because of the BC change:

```sh
1) Cake\Test\TestCase\Cache\CacheTest::testWriteEmptyValues
Failed asserting that false is null.

tests\TestCase\Cache\CacheTest.php:628
```

src/Cache/Cache.php
```diff
@@ -240,6 +240,7 @@ public static function engine($config)
      *
      * Permanently remove all expired and deleted data
      *
+     * @deprecated Use clearExpired() instead
```
ADmad: `@deprecated 3.6.0 Use....`

Issue #11808, opened 2018-03-09
2.x validator()->remove('field', 'rule') fails when rule does not exists (anymore)


WrDX:
This is a (multiple allowed):

* [x] bug
* [x] enhancement

* CakePHP Version: 2.10.8

### What you did
Removing a rule from the validator while the rule does not exist (anymore).
```php
$this->validator()->remove('field', 'rule');
```

### What happened
It fails with an "undifined index" when the rule does not exist, for instance it was previously removed by
```php
$this->validator()->remove('field');
```

### What you expected to happen
The same as removing a full rule, nothing. Since removing a full rule uses unset(), this will not result in an error if the rule does not exist.

dereuromark:
PR open.

PR #11809, opened 2018-03-09
Don't break on removing a non-existing rule

wrdx:
Don't fail with an "undefined index" when removing a non-existing rule.

### DIFF
--- a/lib/Cake/Model/ModelValidator.php
+++ b/lib/Cake/Model/ModelValidator.php
@@ -594,7 +594,7 @@ class ModelValidator implements ArrayAccess, IteratorAggregate, Countable {
 		$this->_parseRules();
 		if ($rule === null) {
 			unset($this->_fields[$field]);
-		} else {
+		} elseif (array_key_exists($field, $this->_fields)) {
 			$this->_fields[$field]->removeRule($rule);
 		}
 		return $this;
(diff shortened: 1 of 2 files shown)


WrDX:
See issue #11808

lib/Cake/Model/ModelValidator.php
```diff
@@ -594,7 +594,7 @@ public function remove($field, $rule = null) {
 		$this->_parseRules();
 		if ($rule === null) {
 			unset($this->_fields[$field]);
-		} else {
+		} elseif (array_key_exists($field, $this->_fields)) {
 			$this->_fields[$field]->removeRule($rule);
 		}
 		return $this;
```
saeideng: maybe
```php
	public function remove($field, $rule = null) {
		$this->_parseRules();

                if(!isset($this->_fields[$field])){ // <--here
                        return $this;
                }

		if ($rule === null) {
			unset($this->_fields[$field]);
		} else {
			$this->_fields[$field]->removeRule($rule);
		}
		return $this;
	}
```

markstory (commented):
Would it be possible to get a test case to cover this scenario and help prevent any regressions in the future?

WrDX:
I'm not sure how to test this, since the error would not occur there's nothing to assert?

Would adding this to ```testRemoveRule()``` in [ValidationTest.php](https://github.com/cakephp/cakephp/blob/5d7f3a67c5b0fadd6b1272df92b356941c470b90/lib/Cake/Test/Case/Model/ModelValidationTest.php#L1994) help?

```php
$Validator->remove('other');
$Validator->remove('other', 'between');
```

This should throw an undefined index in 2.10.8 but not when this is fixed in 2.10.9.

markstory:
@WrDX That snippet sounds like a good start for a test case to me.

dereuromark:
What is the behavior on 3.x currently?

Issue #11810, opened 2018-03-10
[2.x]mock secondary datasource by getMockForModel does not connect test_xxx database.


Khigashiguchi:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: EXACT RELEASE VERSION OR COMMIT HASH, HERE.
CakePHP Version: 2.10.8
* Platform and Target: YOUR WEB-SERVER, DATABASE AND OTHER RELEVANT INFO AND HOW THE REQUEST IS BEING MADE, HERE.
**added**
CircleCI2.0
When test runned, I create database 'test' and 'test_secondary', because I expected test process did touch only test database. 
**added**

### What you did
EXPLAIN WHAT YOU DID, PREFERABLY WITH CODE EXAMPLES, HERE.
I create an test that use getMockForModel in order to mock mock which useDbConfig is secondary,
and run test.

- app/Config/database.php
```php
		$this->default = array(
			'datasource' => 'Database/Mysql',
			'persistent' => false,
			'host' => env('127.0.0.1'),
			'database' => 'first',
			'login' => env('user'),
			'password' => env('password'),
			'prefix' => '',
			'encoding' => 'utf8',
		);

		$this->secondary = array(
			'datasource' => 'Database/Mysql',
			'persistent' => false,
			'host' => env('127.0.0.1'),
			'database' => 'secondary',
			'login' => env('user'),
			'password' => env('password'),
			'prefix' => '',
			'encoding' => 'utf8',
		);


		$this->test = array(
			'datasource' => 'Database/Mysql',
			'persistent' => false,
			'host' => env('127.0.0.1'),
			'database' => 'test',
			'login' => env('user'),
			'password' => env('password'),
			'prefix' => '',
			'encoding' => 'utf8',
		);

		$this->test_secondary = array(
			'datasource' => 'Database/Mysql',
			'persistent' => false,
			'host' => env('127.0.0.1'),
			'database' => 'test_secondary',
			'login' => env('user'),
			'password' => env('password'),
			'prefix' => '',
			'encoding' => 'utf8',
		);
```

- app/Model/Sample.php
```php
class Sample extends AppModel {
	public $useDbConfig = 'secondary';
}
```

- app/Test/Case/Model/SampleTest.php -> testSaveFailure()
```php
		$Sample = $this->getMockForModel('Sample', array('save'));
```

The result of the test is as follow.

```console
[...]

PR #11811, opened 2018-03-10
[2.x]In getMockForModel, place useDbConfig test_xxx

hgsgtk:
Mock secondary datasource by getMockForModel does not connect test_xxx database. 
That's why modify test and CakeTestCake.getMockForModel.
Ref issue #11810


markstory:
Thanks!

PR #11819, opened 2018-03-12
Split TranslateBehavior::locale() into getter/setter.

ndm2:
refs #9978


markstory:
We should keep a test for the deprecated method to ensure compatibility and fix the coverage drop.

ndm2:
Coverage seems back to normal...

PR #11821, opened 2018-03-14
Remove combined getter/setter on QueryTrait.

dereuromark:
Refs https://github.com/cakephp/cakephp/issues/9978

Makes the API already way more speakable and less error prone.
Cyc. complexity will also dramatically fall with the major upgrade then, as the super complex checks on different flag combinations go away.


markstory:
Looks good to me. I can update the migration guide before merging.

PR #11827, opened 2018-03-15
Convert Dbosource order function to use foreach() for php 7.2 compati…

afistfullofash:
…bility

### DIFF
--- a/lib/Cake/Model/Datasource/DboSource.php
+++ b/lib/Cake/Model/Datasource/DboSource.php
@@ -3040,9 +3040,7 @@ class DboSource extends DataSource {
 		$keys = array_filter($keys);
 
 		$result = array();
-		while (!empty($keys)) {
-			list($key, $dir) = each($keys);
-			array_shift($keys);
+	    foreach ($keys as $key => $dir) {
 
 			if (is_numeric($key)) {
 				$key = $dir;


lib/Cake/Model/Datasource/DboSource.php
```diff
@@ -3040,9 +3040,7 @@ public function order($keys, $direction = 'ASC', Model $Model = null) {
 		$keys = array_filter($keys);
 
 		$result = array();
-		while (!empty($keys)) {
-			list($key, $dir) = each($keys);
-			array_shift($keys);
+	    foreach ($keys as $key => $dir) {
```
chinpei215: The current operation is not the same as `foreach`. In some cases, `$keys` is changed in the loop by `array_unshift()` or operator `+`.

markstory: This change has caused tests to fail, as they change ordering in how results are loaded.

tenkoma: You can remove `each()` by changing from the code before change as follows?

```php
// before
list($key, $dir) = each($keys);

// after
$key = key($keys);
$dir = current($keys);
```

chinpei215: No, you can't. The each() function changes the cursor on the array, but neither current() or key() does. We would have to call next(). But I don't understand why each() had to be deprecated, if next() has no problems.

chinpei215: https://wiki.php.net/rfc/deprecations_php_7_2#each
PHP devs seem to have wanted to change list(), not each() itself actually. So probably we could continue to use next() without any concerns.

trenshaw:
I have already commenced these changes. Feel free to assign to me.

lib/Cake/Model/Datasource/DboSource.php
trenshaw: tenkoma's suggestion is valid as array_shift() is being called after each() with list().

The each() function will advance the pointer similar to next() but this is immediately negated by array_shift() which resets the pointer.

Refer http://php.net/manual/en/function.array-shift.php

You could create a $each() lambda outside the while() for use with the existing list() construct (if it's just too hard to say R.I.P. to each() ?)
```php
$each = function(&$keys) {
  return array(key($keys), current($keys));
};
```
Then replace with:
```php
list($key, $dir) = $each($keys);
```
Or just simply assign $key and $dir per tenkoma's suggestion.

I would like to suggest replacing the is_array() check and setting of string in an array to simply cast $keys to (array)

Replace:
```php
if (!is_array($keys)) {
  $keys = array($keys);
}
```
With:
```php
$keys = array_filter((array) $keys);
```

markstory:
Ping @trenshaw

trenshaw:
Have committed but unable to push to 2.x branch.

markstory:
@trenshaw You would have to push to your fork and open a new pull request. You won't have write access to @tpine 's fork.

markstory:
Closing in favour of #12162

PR #11829, opened 2018-03-15
Update cacert.pem

Iandenh:
Latest CA bundle: https://curl.haxx.se/ca/cacert-2018-03-07.pem


markstory:
Thanks!

highstrike:
I think #11675 is worth mentioning here because why do it manually when you can have it automated 😃

Issue #11816, opened 2018-03-12
3.6: Plugins bootstrap not included while running phpunit


cleptric:
Looks like https://github.com/cakephp/cakephp/pull/11785 now causes to not load the plugins `bootstrap.php` file while running phpunit.
The application I upgraded from 3.5 contains a `Application.php`.

3.6.0-beta2

markstory:
Do you need the bootstrap files included to define global constants or functions?

markstory:
Ok. I had another solution for #11785 that involved tracking whether or not a plugin had its bootstrapping run. It involved tracking more state so I didn't go with it. I didn't consider the testsuite startup case though.

markstory:
Pull request open now.

PR #11831, opened 2018-03-16
Include plugin bootstrap on the first time.

markstory:
If a plugin is loaded through the static method it should be immediately bootstrapped, as that is how Plugin::load() has worked for a long time. Because many 3.x plugins rely on their bootstrap file only being executed once per process, we need to disable the bootstrap hook so that `app->pluginBootstrap()` doesn't re-trigger the bootstrap steps.

Refs #11816

### DIFF
--- a/src/Core/Plugin.php
+++ b/src/Core/Plugin.php
@@ -174,10 +174,7 @@ class Plugin
             );
         }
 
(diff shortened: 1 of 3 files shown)


cleptric (approved):
The test suite of the project I noticed the problem now passes again. Thanks @markstory

PR #11833, opened 2018-03-16
Update appveyor.yml

saeideng:


ADmad:
Nice work @saeideng getting appveyor running again.

Seems few tests need to be fixed to get the builds passing.

saeideng:
>Seems few tests need to be fixed to get the builds passing.
yeah 
i want to merge this for now, to see results in next PR(s)
we can add another commit in future

PR #11837, opened 2018-03-18
[RFC] HTML5 validation error messages

jeremyharris:
This is an RFC to discuss adding support for replacing default HTML5 validation errors with the custom validation messages from the app.

The PR is mostly in a POC state as of now. It's hacky because I haven't taken time to add proper support for this feature, such as adding a method to extract the "presence" and "empty" messages from a validation set (right now it "validates" dummy data to get the messages -- hacky!). Also, I only implemented this functionality for `EntityContext`.

Discuss:

- Is this feature worth it? It only adds the custom messages for `requirePresence` or `notBlank` rules
- It adds the `onvalid` and `oninvalid` attributes to controls, so not BC
- It needs JS (inlined on the control). [It's supported well.][0]
- Should it be opt-in/out?

[0]: https://caniuse.com/#feat=constraint-validation


josegonzalez:
This seems neat, I'm a +1. Anything that makes it easier to generate forms is nice.

I'd prefer if this was opt-out. I think it meaningfully improves the downstream user experience, and as such will make application building a bit more delightful. Pushing validation rules to the client is quite nice.

src/View/Form/EntityContext.php
```diff
@@ -426,6 +426,34 @@ public function isRequired($field)
...
+    /**
+     * {@inheritDoc}
+     */
+    public function getErrorMessage($field)
+    {
+        $parts = explode('.', $field);
+
+        $validator = $this->_getValidator($parts);
+        $fieldName = array_pop($parts);
+        if (!$validator->hasField($fieldName)) {
```
markstory: Can the code ever fail to find a validator, resulting in a fatal error here?

src/View/Form/EntityContext.php
```diff
@@ -426,6 +426,34 @@ public function isRequired($field)
...
+            return null;
+        }
+
+        $ruleset = $validator->field($fieldName);
+
+        $presenceErrors = $validator->errors([]);
+        $emptyErrors = $validator->errors([$fieldName => '']);
+
+        if ($ruleset->isPresenceRequired() && isset($presenceErrors[$fieldName])) {
+            return $presenceErrors[$fieldName]['_required'] ?? null;
```
markstory: The `??` operator won't work in PHP 5.6.

src/View/Helper/FormHelper.php
```diff
@@ -1433,6 +1433,13 @@ protected function _magicOptions($fieldName, $options, $allowOverride)
             $options['required'] = $context->isRequired($fieldName);
         }
 
+        $message = $context->getErrorMessage($fieldName);
+        if ($options['required'] && $message) {
+            $message = htmlspecialchars(addslashes($message));
```
markstory: Should we use `h()` to handle application specific encodings?

ADmad:
We could have this as opt-in in 3.6 and switch to opt-out in 4.0.

src/View/Form/EntityContext.php
jeremyharris: Not that I can see. It always returns a validator. This logic was copied from `isRequired` which would suffer from the same error.

dereuromark: You could however add a sanity check to be safe. And throw meaningful exception (with clear text) then instead of running into a fatal null->usage error.

inoas: Maybe https://api.cakephp.org/3.5/source-class-Cake.View.Form.EntityContext.html#442-477 could be modified instead to throw an exception (if BC is required, by setting a 2nd parameter to true, which should in 4.x then be the default, in 3.x it would not for existing calls)?

src/View/Helper/FormHelper.php
jeremyharris: Makes sense. Doing so will require `cakephp/core` in the `cakephp/validation` package.

markstory: Why? This message is being escaped in View/Helper/FormHelper.

jeremyharris: Because I'm dumb and got confused about which file I was looking at :D

src/Validation/Validator.php
```diff
@@ -1941,6 +1929,42 @@ public function regex($field, $regex, $message = null, $when = null)
...
+            : $defaultMessage;
+    }
+
+    /**
+     * Gets the empty message for a field
+     *
+     * @param string $field Field name
+     * @return string
+     */
+    public function getEmptyMessage($field)
```
dereuromark: getRequired vs getEmpty seems an API contradiction.
You seem to want to name it `getNotEmptyMessage`.
Also, didnt we go from empty to blank for the reason of confusion in PHP core code usage and behavior of empty()?

jeremyharris:
I added support for the other contexts and cleaned up the code per the comments. The flag to enable this feature is in the form helper config. I wasn't sure if it deserved a full flag method so I put it in config for now, `$formHelper->setConfig('useValidationMessages', true)`.

src/Validation/Validator.php
ADmad: > Also, didnt we go from empty to blank for the reason of confusion in PHP core code usage and behavior of empty()?

No, we changed `Validation::notEmpty()` to `Validation::notBlank()` because of confusion between the former and `Validator::notEmpty()`.

jeremyharris: You're right about the naming, it should be `getNotEmptyMessage`. As for notBlank, it doesn't handle custom not empty messages the same way notEmpty does. We would have to iterate through all the rules and find the notBlank one to extract the message, share the `_allowEmptyMessages` var, or create a new var on validator to explicitly store notBlank messages. Pick your poison (detecting the rule seems the best to me).

jeremyharris:
The appveyor failure is unrelated. I thought it was fixed in 3.next but I guess it wasn't at the time I merged.

Any other thoughts on this?

src/Validation/Validator.php
```diff
@@ -1941,6 +1943,54 @@ public function regex($field, $regex, $message = null, $when = null)
...
+     * Gets the required message for a field
+     *
+     * @param string $field Field name
+     * @return string
+     */
+    public function getRequiredMessage($field)
+    {
+        $defaultMessage = static::DEFAULT_REQUIRED_MESSAGE;
+        if ($this->_useI18n) {
+            $defaultMessage = __d('cake', $defaultMessage);
```
markstory: This will make the po file extractor skip these messages, as it cannot resolve variables or constants.

src/Validation/Validator.php
```diff
@@ -1941,6 +1943,54 @@ public function regex($field, $regex, $message = null, $when = null)
...
+        }
+
+        $notBlankMessage = null;
+        if (isset($this->_fields[$field])) {
+            foreach ($this->_fields[$field] as $rule) {
+                if ($rule->get('rule') === 'notBlank') {
+                    $notBlankMessage = $rule->get('message');
+                    break;
```
markstory: If this also checked the message truthiness, couldn't you return here and save some logic below?

src/View/Form/ContextInterface.php
```diff
@@ -70,6 +70,14 @@ public function val($field);
      */
     public function isRequired($field);
 
+    /**
+     * Gets the default "required" error message for a field
+     *
+     * @param string $field A dot separated path to the field
+     * @return string
+     */
+    public function getRequiredMessage($field);
```
markstory: Changing interfaces usually has to wait until a major version as it can break other implementations. I know that the elastic search plugin also uses this interface.

src/Validation/Validator.php
jeremyharris: TIL :)

src/View/Form/ContextInterface.php
jeremyharris: Since we don't have a 4.0 branch yet, should I just close this? Or let it sit and re-target after there is one?

dereuromark: Can't we annotate the interface method as doc block as done in other places for 3.x?

src/Validation/Validator.php
```diff
@@ -1941,6 +1929,51 @@ public function regex($field, $regex, $message = null, $when = null)
...
+     */
+    public function getNotEmptyMessage($field)
+    {
+        $defaultMessage = 'This field cannot be left empty';
+        if ($this->_useI18n) {
+            $defaultMessage = __d('cake', 'This field cannot be left empty');
+        }
+
+        $notBlankMessage = null;
+        if (isset($this->_fields[$field])) {
```
ADmad: How about doing this check at the beginning of method and simply returning `null` if no validation set is set for this field?  Also makes if wonder if this needs to be a public problem. If it were protected check wouldn't even be needed since `$field` is got by iterating over `$this>_fields` in `errors()`.

src/View/Form/ContextInterface.php
markstory: So far we've been annotating interfaces and using `method_exists` as an interim solution.

src/Validation/Validator.php
ADmad: Seem odd to me to return a validation message even when no validations are set for field provided as argument.

jeremyharris:
I think this is ready to go, unless there are any more comments.

src/View/Helper/FormHelper.php
```diff
@@ -163,7 +163,9 @@ class FormHelper extends Helper
             'textarea' => '<textarea name="{{name}}"{{attrs}}>{{value}}</textarea>',
             // Container for submit buttons.
             'submitContainer' => '<div class="submit">{{content}}</div>',
-        ]
+        ],
+        // set HTML5 validation message to custom required/empty messages
+        'useValidationMessages' => false,
```
inoas: Maybe rename it to `'injectHtml5ValidationMessages' or `injectClientValidationMessages` or `autoSetCustomValidity` or ... *something better?* ... basically anything not as ambiguous/unspecific as `useValidationMessages`, because validation messages are already being used by default (server side), and setting this option to false won't stop those.

Then:
It needs to be documented to enable this?
Will this be enabled by default in 4.x?

jeremyharris: I'm okay with renaming it; the setting is in the FormHelper context and it's annotated with a comment, so I didn't think it was too ambiguous, but I see what you mean. Of your choices, `autoSetCustomValidity` feels best to me (inject sounds danger danger).

It'll be enabled by default in 4.x if this gets merged. I will document it as well.

src/View/Helper/FormHelper.php
```diff
@@ -1445,6 +1447,15 @@ protected function _magicOptions($fieldName, $options, $allowOverride)
             $options['required'] = $context->isRequired($fieldName);
         }
 
+        if (method_exists($context, 'getRequiredMessage')) {
+            $message = $context->getRequiredMessage($fieldName);
+            if ($options['required'] && $message && $this->getConfig('useValidationMessages')) {
+                $message = h(addslashes($message));
+                $options['oninvalid'] = "this.setCustomValidity('$message')";
```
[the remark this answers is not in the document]
jeremyharris: No, but I can make one. Another option would be to always include the message in a `data-` attribute, that way users can disregard or use it.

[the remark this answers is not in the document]
ADmad: Use `data-` attributes sounds good too?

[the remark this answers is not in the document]
inoas: What about a string based template much like the form helper uses otherwise (that default template could then use data or not, I don't personally care much, as it is configurable then)?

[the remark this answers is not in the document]
jeremyharris: @inoas that's not a bad idea. I would have to place it into the control's `templateVars` so users could add `'input' => '<input type="{{type}}" name="{{name}}"{{attrs}} data-mysuperawesomeattribute="{{requiredMessage}}"/>',`.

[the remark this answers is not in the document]
jeremyharris: @ADmad I included `requiredMessage` in the templateVars if the field is "required" and has a message. This allows you to adjust the templates according to whatever data attribute you'd want, per @ionas's suggestion.

Part of the documentation I write after this is merged will include this change, and an example on using it.

src/View/Helper/FormHelper.php
```diff
@@ -1441,10 +1443,28 @@ protected function _magicOptions($fieldName, $options, $allowOverride)
...
+
         if (!isset($options['required']) && $options['type'] !== 'hidden') {
             $options['required'] = $context->isRequired($fieldName);
         }
 
+        if (method_exists($context, 'getRequiredMessage')) {
+            $message = $context->getRequiredMessage($fieldName);
+
+            if ($options['required'] && $message) {
+                $options['templateVars']['requiredMessage'] = $message;
```
ADmad: Perhaps use name `customValidityMessage` instead for similarity with `autoSetCustomValidity` option? Which also makes me think if `getRequiredMessage()` should be renamed.

I don't care much about specific names but I would be nice to have all methods and options related to this feature similarly named.

tests/TestCase/View/Helper/FormHelperTest.php
```diff
@@ -8433,6 +8433,165 @@ public function testHtml5ControlException()
...
+            'Email',
+            '/label',
+            'input' => [
+                'id' => 'email',
+                'name' => 'email',
+                'type' => 'email',
+                'value' => '',
+                'maxlength' => 255,
+                'required' => 'required',
+                'data-message' => 'Custom error message',
```
ADmad: Can you please use a string with special chars to ensure it's properly escaped when setting `data-message`.

jeremyharris: Thanks for catching that. Added a test, caught a bug 💪

src/View/Helper/FormHelper.php
```diff
@@ -1441,10 +1443,28 @@ protected function _magicOptions($fieldName, $options, $allowOverride)
...
+            'templateVars' => []
+        ];
+
         if (!isset($options['required']) && $options['type'] !== 'hidden') {
             $options['required'] = $context->isRequired($fieldName);
         }
 
+        if (method_exists($context, 'getRequiredMessage')) {
+            $message = $context->getRequiredMessage($fieldName);
+            $message = addslashes(h($message));
```
markstory: Why do you need `addslashes()` shouldn't HTML encoding encode all the quote characters?

markstory (commented):
Sorry for the late review on this. I was going through older pull requests with the intent of getting them merged up.

src/View/Helper/FormHelper.php
jeremyharris: I think that was leftover from a previous version. Removed 👍

markstory:
Looks good. I'm still thinking about where this lives in the docs, but I like the changes.

src/View/Form/ArrayContext.php
```diff
@@ -194,16 +201,36 @@ public function val($field, $options = [])
...
 
-        return (bool)$required;
+        if ($required === false) {
+            return null;
+        }
+
+        if ($required === true) {
+            $required = 'This field is required';
+
+            if (function_exists('__d ')) {
```
ADmad: Is this check necessary? We don't have a split package for view layer so won't `__d()` alway be available?

src/View/Form/ContextInterface.php
```diff
@@ -16,6 +16,8 @@
 
 /**
  * Interface for FormHelper context implementations.
+ *
+ * @method string getRequiredMessage($field) Gets the default "required" error message for a field
```
ADmad: Should be `string|null`.

src/View/Form/ArrayContext.php
ADmad: Also you have an extra space at end of `'__d '` so the check would probably never return `true` :slightly_smiling_face:.

jeremyharris: That was pulled from the validation package which required the check since it's a standalone. But now that the logic is here I can remove the check. Not sure how the extra space got in there 😄

jeremyharris:
@markstory in addition to adding `autoSetCustomValidity` to the FormHelper "options" list, I was going to add docs about it under https://book.cakephp.org/3.0/en/views/helpers/form.html#displaying-and-checking-errors probably as a subsection (e.g. Displaying messages in HTML5 validation or something). The note about the `customValidityMessage` template var could fit there as well. If there's a better place to put it, let me know!

Sorry all about the length and noise of this messy PR.

markstory:
@jeremyharris Those documentation updates sound good to me. There is no need to apologize for the amount of discussion. It has been productive and useful I think.

markstory (approved):

src/View/Helper/FormHelper.php
```diff
@@ -165,7 +165,9 @@ class FormHelper extends Helper
             'submitContainer' => '<div class="submit">{{content}}</div>',
             //Confirm javascript template for postLink()
             'confirmJs' => '{{confirm}}',
-        ]
+        ],
+        // set HTML5 validation message to custom required/empty messages
+        'autoSetCustomValidity' => false,
```
dereuromark: I wish those would be class consts or sth to avoid magic string typos when using it :)

inoas: I wish, too, that php had atoms.

Issue #11838, opened 2018-03-18
ConsoleOptionParser addSubcommand Sorting optional


t73biz:
This is a (multiple allowed):

* [ ] bug
* [x] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 235c72450449614bce13ee145853183d70f2d7ae
* Platform and Target: Linux Mint 18, PHP 7.1, Shell

### What you did
When adding subcommands to a ShellCommand, I realized that the Tasks first letter spelt a word. I would like to see the help display for subcommands as I enter them.

```
/**
     * Manage the available sub-commands along with their arguments and help
     *
     * @see http://book.cakephp.org/3.0/en/console-and-shells.html#configuring-options-and-generating-help
     *
     * @return \Cake\Console\ConsoleOptionParser
     */
    public function getOptionParser()
    {
        $parser = parent::getOptionParser();

        $parser->addSubcommand('generate', [
            'help' => 'Generate the link tables.',
            'parser' => $this->Generate->getOptionParser(),
        ]);

        $parser->addSubcommand('import', [
            'help' => 'Import the csv data.',
            'parser' => $this->Import->getOptionParser(),
        ]);

        $parser->addSubcommand('join', [
            'help' => 'Join the habtm models.',
            'parser' => $this->Join->getOptionParser(),
        ]);

        $parser->addSubcommand('organize', [
            'help' => 'Organize the model mapping.',
            'parser' => $this->Join->getOptionParser(),
        ]);

        $parser->addSubcommand('erase', [
            'help' => 'Erase the link tables.',
            'parser' => $this->Join->getOptionParser(),
        ]);

        return $parser;
    }
```

### What happened
The subcommands are sorted
```
Usage:
cake csv_migrate [subcommand] [-h] [-q] [-v]

Subcommands:

erase     Erase the link tables.
generate  Generate the link tables.
import    Import the csv data.
join      Join the habtm models.
organize  Organize the model mapping.

To see help on a subcommand use `cake csv_migrate [subcommand] --help`

```

### What you expected to happen
The subcommands listed in the same order as I entered them in to the class.
```
Usage:
cake csv_migrate [subcommand] [-h] [-q] [-v]

Subcommands:

generate  Generate the link tables.
import    Import the csv data.
join      Join the habtm models.
organize  Organize the model mapping.
erase     Erase the link tables.

To see help on a subcommand use `cake csv_migrate [subcommand] --help`

```

saeideng:
[...]

PR #11839, opened 2018-03-18
ConsoleOptionParser addSubcommand Sorting - Issue #11838

t73biz:
Issue #11838
Implemented an optional sort parameter to \Cake\Console\ConsoleOptionParser::addSubcommand
Added unit test for the new parameter.

### DIFF
--- a/src/Console/ConsoleOptionParser.php
+++ b/src/Console/ConsoleOptionParser.php
@@ -576,7 +576,7 @@ class ConsoleOptionParser
      *
      * @param \Cake\Console\ConsoleInputSubcommand|string $name Name of the subcommand. Will also accept an instance of ConsoleInputSubcommand
      * @param array $options Array of params, see above.
-     * @param boolean $sort
+     * @param boolean $sort Optional sorting of the subcommand
      * @return $this
      */
     public function addSubcommand($name, array $options = [], $sort = true)
--- a/src/Console/ConsoleOptionParser.php
+++ b/src/Console/ConsoleOptionParser.php
@@ -121,6 +121,13 @@ class ConsoleOptionParser
     protected $_subcommands = [];
 
     /**
+     * Subcommand sorting option
+     *
+     * @var bool
+     */
+    protected $_subcommandSort = true;
+
+    /**
      * Command name.
      *
      * @var string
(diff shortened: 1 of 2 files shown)


t73biz:
Not sure why Travis #26753.6 failed
```
1) Cake\Test\TestCase\Cache\Engine\MemcachedEngineTest::testIncrementDecrementExpiring
Failed asserting that 2 is false.
/home/travis/build/cakephp/cakephp/tests/TestCase/Cache/Engine/MemcachedEngineTest.php:717
```

as this is not even testing anything to do with my PR Changes

ADmad:
This should probably target `3.next`. Changing method arguments in bugfix releases is not BC.

src/Console/ConsoleOptionParser.php
```diff
@@ -576,9 +576,10 @@ public function addOptions(array $options)
      *
      * @param \Cake\Console\ConsoleInputSubcommand|string $name Name of the subcommand. Will also accept an instance of ConsoleInputSubcommand
      * @param array $options Array of params, see above.
+     * @param bool $sort Optional sorting of the subcommand
      * @return $this
      */
-    public function addSubcommand($name, array $options = [])
+    public function addSubcommand($name, array $options = [], $sort = true)
```
dereuromark: I wonder if sort could/should be part of options array.
This way it would be also not lost when doing transformation (e.g. array format export), and the signature here stays simpler.

markstory: Or a stateful aspect of the OptionParser. Right now, forgetting to pass sort=false in one subcommand will undo all of the other `true` values. That feels like undesirable behavior to me.

t73biz: @dereuromark I had first tested it this way, and it occurred to me that in this scope, the options being defined were actually for the subcommand. Since this PR relates to the parser functionality, I would have to agree with @markstory and his approach suggestion.

I will modify this to take that approach and update the PR.

src/Console/ConsoleOptionParser.php
```diff
@@ -407,6 +414,32 @@ public function epilog($text = null)
...
+    }
+
+    /**
+     * Sets the subcommandSort property.
+     *
+     * @param bool $sort Whether or not the ConsoleOptionParser should sort subcommands
+     *
+     * @return $this
+     */
+    public function setSubcommandSort($sort)
```
markstory: In other places we use `enableSubcommandSort()` as a method name.

t73biz: Is this due to the boolean type for $subcommandSort ? Also, since it's set to true by default here, shouldn't it be disableSubcommandSort() as well?

dereuromark: you can make it set to True by default.

PR #11841, opened 2018-03-18
Phpstan level 3

ADmad:


src/ORM/Association/Loader/SelectLoader.php
```diff
@@ -294,13 +294,14 @@ protected function _addFilteringJoin($query, $key, $subquery)
...
 
+        $conditions = null;
         if (is_array($key)) {
             $conditions = $this->_createTupleCondition($query, $key, $filter, '=');
         } else {
             $filter = current($filter);
         }
 
-        $conditions = isset($conditions) ? $conditions : $query->newExpr([$key => $filter]);
+        $conditions = $conditions ?: $query->newExpr([$key => $filter]);
```
dereuromark: Could conditions be defined (not null) but falsey?

ADmad: No, it can only be `null` or `TupleComparison` (returned by `_createTupleCondition()`).

src/View/Helper/UrlHelper.php
```diff
@@ -52,8 +52,10 @@ public function build($url = null, $options = false)
         }
         $options += $defaults;
 
+        /** @var string $url */
         $url = Router::url($url, $options['fullBase']);
         if ($options['escape']) {
+            /** @var string $url */
```
dereuromark: This seems quite redundant and unnecessary  - but I guess that means that phpstan needs it.

ADmad: It's need because return type for `h()` is `string|array`.

markstory (approved):
Looks good to me.

PR #11844, opened 2018-03-19
Add often-used Entity methods to EntityInterface

muglug:
This adds `EntityInterface::setDirty` and `EntityInterface::isDirty`, methods which are called in a number of places in the codebase but not defined in the interface.

### DIFF
--- a/src/Datasource/EntityInterface.php
+++ b/src/Datasource/EntityInterface.php
@@ -148,7 +148,7 @@ interface EntityInterface extends ArrayAccess, JsonSerializable
      * @return bool whether the property was changed or not
      */
     public function dirty($property = null, $isDirty = null);
-    
+
     /**
      * Sets the dirty status of a single property.
      *
--- a/src/Datasource/EntityInterface.php
+++ b/src/Datasource/EntityInterface.php
@@ -148,6 +148,24 @@ interface EntityInterface extends ArrayAccess, JsonSerializable
      * @return bool whether the property was changed or not
      */
     public function dirty($property = null, $isDirty = null);
+    
+    /**
+     * Sets the dirty status of a single property.
+     *
+     * @param string $property the field to set or check status for
+     * @param bool $isDirty true means the property was changed, false means
+     * it was not changed
+     * @return $this
+     */
+    public function setDirty($property, $isDirty);
+
+    /**
+     * Checks if the entity is dirty or if a single property of it is dirty.
+     *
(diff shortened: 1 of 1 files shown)


dereuromark:
This would have to be done in 3.next, not master.
But even so we usually are hesitant to add methods to Interfaces, as this can be a BC for people implementing the interface without the concrete implementation of Entity itself.

src/Datasource/EntityInterface.php
```diff
@@ -149,6 +149,24 @@ public function extract(array $properties, $onlyDirty = false);
...
+     * @param bool $isDirty true means the property was changed, false means
+     * it was not changed
+     * @return $this
+     */
+    public function setDirty($property, $isDirty);
+
+    /**
+     * Checks if the entity is dirty or if a single property of it is dirty.
+     *
+     * @param string $property the field to check the status for
```
dereuromark: |null
this is also wrong in master currently :)

ADmad:
@muglug We can't modify interfaces until next major release. Which is why we have added `@method` annotations at the top for the time being. `phpstan` supports them, perhaps `pslam` should too :slightly_smiling_face:.

muglug:
@Admad it definitely should. Closing...

ADmad:
@muglug Thank you for your interest.

src/Datasource/EntityInterface.php
dereuromark: Fixed in https://github.com/cakephp/cakephp/commit/1c6b605e06b29a411857563924241de6f2a6f37f

muglug:
Although support for `@method` should only apply to classes with `__call`. It should however be possible to filter errors on a per-method basis, similar to https://github.com/vimeo/psalm/issues/603

ADmad:
I was just giving `pslam` a try last night. We may or may not end up actually using it. Making static analyzers happy with existing codebase is a tedious job :slightly_smiling_face:.

muglug:
> Making static analyzers happy with existing codebase is a tedious job

I wholeheartedly agree. I've created a config that ignores lots and lots of issues here: https://github.com/cakephp/cakephp/compare/master...muglug:master. About 40 issues that it still complains about, but most of those are due to incorrect Psalm analysis which I'll fix in the coming days.

ADmad:
@muglug I would suggest working on the `3.next` branch (used for next minor release) which should have lesser errors reported than `master`.

Issue #11845, opened 2018-03-19
MemcachedEngine: groups method throws warning after a memcached daemon restart (getMulti return false instead of array)


xavifr:
This is a:

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: master.
* Platform and Target: Ubuntu 16.04.

### What happens
Sometimes. after a restart of memcached daemon, the following read/write operations return some warnings

```
Warning Error: count(): Parameter must be an array or an object that implements Countable in [/opt/cake/app/vendor/cakephp/cakephp/src/Cache/Engine/MemcachedEngine.php, line 504]
Warning Error: array_values() expects parameter 1 to be array, boolean given in [/opt/cake/app/vendor/cakephp/cakephp/src/Cache/Engine/MemcachedEngine.php, line 515]

```

After looking at MemcachedEngine.php it looks like that getMulti can return false in some cases, but the code doesn't supports that.

It can be reproduced restarting memcached daemon while an script has the Cache object initialized and performs a read operation.

### PATCH

[fix-getMulti-return-false.txt](https://github.com/cakephp/cakephp/files/1824978/fix-getMulti-return-false.txt)

PR #11846, opened 2018-03-19
Add cast to coerce false into a countable.

markstory:
Memcached::getMulti() can return false when memcached fails.

Refs #11845


markstory:
@dereuromark You're totally right. My brain was not in PHP mode.

PR #11840, opened 2018-03-18
3.next-appveyor

saeideng:
* [x] fix patch on windowns test 
* [ ] Cake\Test\TestCase\Database\QueryTest::testCastResults (https://github.com/cakephp/cakephp/pull/11754)
* [x] Cake\Test\TestCase\Database\Driver\SqlserverTest::testConnectionConfigCustom

PR #11847, opened 2018-03-19
Fixed TimestampBehavior failing when type isn't DateTimeType

jeremyharris:
I was testing out 3.next and a bunch of my tests failed. I looked into it and it turned out I had an oddly stray `created VARCHAR(45)` column made it into one my the tables, which was apparently supported as the table has a bunch of timestamp string in it :D This caused an undefined method error as  `getDateTimeClassName` doesn't exist on StringType.

This re-adds the string type compatibility. I'm off to fix my DB in the meantime...

### DIFF
--- a/src/ORM/Behavior/TimestampBehavior.php
+++ b/src/ORM/Behavior/TimestampBehavior.php
@@ -204,6 +204,13 @@ class TimestampBehavior extends Behavior
 
         /** @var \Cake\Database\Type\DateTimeType $type */
         $type = Type::build($columnType);
+
+        if (!$type instanceof Type\DateTimeType) {
+            $entity->set($field, (string)$ts);
+
+            return;
+        }
+
         $class = $type->getDateTimeClassName();
 
         $entity->set($field, new $class($ts));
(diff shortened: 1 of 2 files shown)


ADmad:
I guess re-adding support for string columns is fine for BC but I would throw a warning for it and only support datetime fields in 4.0.

jeremyharris:
I agree throwing an error in 4.0 would be good. I can open a PR for that. Should we add a deprecation noticed here in 3.next?

ADmad:
> Should we add a deprecation noticed here in 3.next?

We should IMO.

src/ORM/Behavior/TimestampBehavior.php
```diff
@@ -204,6 +204,13 @@ protected function _updateField($entity, $field, $refreshTimestamp)
 
         /** @var \Cake\Database\Type\DateTimeType $type */
         $type = Type::build($columnType);
+
+        if (!$type instanceof Type\DateTimeType) {
+            $entity->set($field, (string)$ts);
```
markstory: This would be a good place to emit a deprecation warning.

markstory:
Looks like you might need to update a test, as deprecation warnings cause our tests to fail.

jeremyharris:
The appveyor failing is unrelated and being worked on here: https://github.com/cakephp/cakephp/pull/11840

PR #11848, opened 2018-03-19
PSR-16 compatible cache interface

burzum:


src/Cache/Cache.php
```diff
@@ -388,7 +442,37 @@ public static function readMany($keys, $config = 'default')
...
+     * Cache::getMultiple(['my_data_1', 'my_data_2], null, 'long_term');
+     * ```
+     *
+     * @param array $keys an array of keys to fetch from the cache
+     * @param mixed $default Default value
+     * @param string $config optional name of the configuration to use. Defaults to 'default'
+     * @return array An array containing, for each of the given $keys, the cached data or false if cached data could not be
+     * retrieved.
+     */
+    public function getMultiple($keys, $default = null, $config = 'default')
```
dereuromark: should we have array typehint here?

src/Cache/Cache.php
```diff
@@ -486,6 +571,35 @@ public static function deleteMany($keys, $config = 'default')
...
+     * ```
+     * Cache::deleteMultiple(['my_data_1', 'my_data_2], 'long_term');
+     * ```
+     *
+     * @param array $keys Array of cache keys to be deleted
+     * @param string $config name of the configuration to use. Defaults to 'default'
+     * @return bool
+     * couldn't be removed
+     */
+    public static function deleteMultiple($keys, $config = 'default')
```
dereuromark: should we have array typehint here?

src/Cache/Cache.php
```diff
@@ -486,6 +571,35 @@ public static function deleteMany($keys, $config = 'default')
...
+     * Deleting from a specific cache configuration.
+     *
+     * ```
+     * Cache::deleteMultiple(['my_data_1', 'my_data_2], 'long_term');
+     * ```
+     *
+     * @param array $keys Array of cache keys to be deleted
+     * @param string $config name of the configuration to use. Defaults to 'default'
+     * @return bool
+     * couldn't be removed
```
dereuromark: What does this comment mean? Should this be above the params instead with a more clear message?

src/Cache/CacheEngine.php
```diff
@@ -295,4 +211,165 @@ protected function warning($message)
...
+     * @param iterable $values A list of key => value pairs for a multiple-set operation.
+     * @param null|int|DateInterval $ttl Optional. The TTL value of this item. If no value is sent and
+     *                                      the driver supports TTL then the library may set a default value
+     *                                      for it or let the driver take care of that.
+     * @return bool True on success and false on failure.
+     * @throws \Psr\SimpleCache\InvalidArgumentException
+     *   MUST be thrown if $values is neither an array nor a Traversable,
+     *   or if any of the $values are not a legal value.
+     */
+    public function setMultiple($values, $ttl = null)
```
dereuromark: You use iterable here on doc block, but typehint is not?

src/Cache/Cache.php
ADmad: https://github.com/php-fig/simple-cache/blob/master/src/CacheInterface.php#L58
So the typehint in docblock should be changed from `array` to `iterable`. Similarly for all `*Multiple()` methods.

markstory:
I'll try to fix a few of the failing tests in the next day or so.

src/Cache/Engine/RedisEngine.php
```diff
@@ -123,47 +121,84 @@ protected function _connect()
...
+     */
+    public function set($key, $data, $ttl = null)
+    {
+        $key = $this->_key($key);
+
+        if (!is_int($data)) {
+            $data = serialize($data);
+        }
+
+        $duration = $this->_config['duration'];
```
markstory: We aren't using the ttl parameter here.

src/Cache/Engine/XcacheEngine.php
```diff
@@ -71,11 +71,63 @@ public function init(array $config = [])
...
+     * Write data for key into cache
+     *
+     * @param string $key Identifier for the data
+     * @param mixed $value Data to be cached
+     * @param null|int|DateInterval $ttl Optional. The TTL value of this item. If no value is sent and
+     *                                     the driver supports TTL then the library may set a default value
+     *                                     for it or let the driver take care of that.
+     * @return bool True if the data was successfully cached, false on failure
+     */
+    public function set($key, $value, $ttl = null)
```
markstory: This method isn't using the ttl parameter.

src/Cache/Engine/NullEngine.php
```diff
@@ -113,4 +122,20 @@ public function clearGroup($group)
     {
         return false;
     }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function get($key, $default = null)
+    {
+        return false;
```
markstory: Shouldn't this return `$default`?

src/Cache/Engine/RedisEngine.php
```diff
@@ -123,47 +121,84 @@ protected function _connect()
...
+            return (int)$value;
+        }
+
+        if ($value !== false && is_string($value)) {
+            $result = unserialize($value);
+
+            return $result === false ? null : $result;
+        }
+
+        return $value === false ? null : $value;
```
markstory: We aren't using the default value in this method, we should return that on failures.

src/Cache/Engine/NullEngine.php
burzum: Hm... whats the purpose of the Null engine? In which scenario do we want the default instead of false?

src/Cache/Engine/XcacheEngine.php
burzum: Fixed it, hopefully correctly.

src/Cache/Engine/RedisEngine.php
burzum: Fixed.

src/Cache/Cache.php
```diff
@@ -500,6 +632,19 @@ public static function clear($check = false, $config = 'default')
         return $engine->clear($check);
     }
 
+    /**
+     * Clears only expired cache entries
+     *
+     * @param string $config name of the configuration to use. Defaults to 'default'
+     * @return bool
+     */
+    public function clearExpired($config = 'default')
```
markstory: If this isn't part of the PSR interfaces why are we adding it? It won't do anything on most engines as they auto-expire data.

src/Cache/CacheEngine.php
```diff
@@ -171,31 +110,6 @@ public function readMany($keys)
      */
     abstract public function delete($key);
 
-    /**
-     * Delete all keys from the cache
-     *
-     * @param bool $check if true will check expiration, otherwise delete all
-     * @return bool True if the cache was successfully cleared, false otherwise
-     */
-    abstract public function clear($check);
```
markstory: I'm not sure we can change the `clear()` signature without breaking compatibility as it could have been implemented by userland engines.

markstory: I was thinking some mote about this last night. We could add the new methods required by the PSR interfaces now, but not implement the actual interfaces until 4.0 when we can make the breaking changes to clear().

ADmad: @markstory In that case wouldn't it be better to also not change the return value for engines in case of cache miss from current `false` to `null`? It could potentially cause issues with current applications doing strict checks.

markstory: I don't think we should change return values for the existing write/read methods as it could cause compatibility issues. However, we sould implement get(), set(), and has() in a way that will be compatible with the PSR interfaces.

markstory:
Going to close this as the wrapper approach implemented in #12520 is lower risk for 3.x.

PR #11856, opened 2018-03-20
Fixed RequestHandlerComponent overwriting renderAs response

jeremyharris:
Another little regression I found via my tests.

### DIFF
--- a/src/Controller/Component/RequestHandlerComponent.php
+++ b/src/Controller/Component/RequestHandlerComponent.php
@@ -331,6 +331,7 @@ class RequestHandlerComponent extends Component
 
         if ($this->ext && $isRecognized) {
             $this->renderAs($controller, $this->ext);
+            $response = $controller->response;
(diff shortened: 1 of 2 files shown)


src/Controller/Component/RequestHandlerComponent.php
```diff
@@ -331,6 +331,7 @@ public function beforeRender(Event $event)
 
         if ($this->ext && $isRecognized) {
             $this->renderAs($controller, $this->ext);
+            $response = $controller->response;
```
saeideng: this can be remove in future accidentally
adding a bit note is useful here IMO

ADmad: @saeideng The new test added should avoid such a regression. :slightly_smiling_face:

markstory:
Thanks for taking the time to test the beta @jeremyharris

PR #11857, opened 2018-03-20
Fixed Form regression that skipped validation

jeremyharris:
The following was previously valid:

```php
<?php
namespace TestApp\Form;
use Cake\Form\Form;
class ValidateForm extends Form
{
    public function validator(\Cake\Validation\Validator $validator = null)
    {
        return parent::validator($validator)
            ->requirePresence('title');
    }
}
```

In fact, the Form class says in its doc:

> You can also define the validation and schema by chaining method calls off of `$form->schema()` and `$form->validator()`.

The only way I could figure out how to fix this without always throwing a deprecation was to check if the validator built via the event was empty, and replace it with the previously always-called `->validator()` within `validate()`.

### DIFF
--- a/tests/TestCase/View/Form/FormContextTest.php
+++ b/tests/TestCase/View/Form/FormContextTest.php
@@ -259,13 +259,9 @@ class FormContextTest extends TestCase
         $this->assertEquals([], $context->error('Alias.name'));
         $this->assertEquals([], $context->error('nope.nope'));
 
-        $mock = $this->getMockBuilder('Cake\Validation\Validator')
-            ->setMethods(['errors'])
-            ->getMock();
-        $mock->expects($this->once())
-            ->method('errors')
-            ->willReturn(['key' => 'should be an array, not a string']);
-        $form->setValidator('default', $mock);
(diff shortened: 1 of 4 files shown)


markstory:
One test failing because of a deprecation warning. I can get that fixed up if you don't get there first.

tests/TestCase/View/Form/FormContextTest.php
```diff
@@ -259,13 +259,9 @@ public function testError()
...
-        $mock = $this->getMockBuilder('Cake\Validation\Validator')
-            ->setMethods(['errors'])
-            ->getMock();
-        $mock->expects($this->once())
-            ->method('errors')
-            ->willReturn(['key' => 'should be an array, not a string']);
-        $form->setValidator('default', $mock);
+        $validator = new Validator();
+        $validator->requirePresence('key', true, 'should be an array, not a string');
+        $form->setValidator('default', $validator);
```
jeremyharris: I couldn't suss out why this new validator was mocked, so I replaced it with a concrete validator so it implements the `->count` method relied on by the Form class.

Issue #11772, opened 2018-03-01
Saving new hasMany associations records fails due to existsIn build rule


pbfman:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.5.6
* Platform and Target: Apache2, MySQL, PHP 7

### What you did
I've got two tables ``Events`` and ``Dates`` (``Events`` hasMany ``Dates``):

```php
class EventsTable extends Table
{
    public function initialize(array $config)
    {
        parent::initialize($config);

        //further config/associations ....

        $this->hasMany('Dates', [
            'foreignKey' => 'event_id',
            'className' => 'Dsc/Core.Dates'
        ]);
    }
}
```

```php
class DatesTable extends Table
{

    /**
     * Initialize method
     *
     * @param array $config The configuration for the Table.
     * @return void
     */
    public function initialize(array $config)
    {
        parent::initialize($config);

        //further config/associations ....

        $this->belongsTo('Events', [
            'foreignKey' => 'event_id',
            'joinType' => 'INNER',
            'className' => 'Dsc/Core.Events'
        ]);
    }

    public function buildRules(RulesChecker $rules)
    {
        $rules->add($rules->existsIn(['event_id'], 'Events'));

        return $rules;
    }
}
```

I want to insert a **new** Event including a **new** Date. Therefore i created a form:

```php
    <?= $this->Form->create($event) ?>
    <fieldset>
        <legend><?= (isset($edit) && $edit ? __('Edit Event') : __('Add Event')) ?></legend>
        <?php
       //further event fields
        echo $this->Form->control('dates.0.id', ['type' => 'hidden']);
        echo $this->Form->control('dates.0.date_start', ['type' => 'date']);
        echo $this->Form->control('dates.0.time_start', ['type' => 'time']);
        echo $this->Form->control('dates.0.date_end', ['type' => 'date']);
        echo $this->Form->control('dates.0.time_end', ['type' => 'time']);
        ?>
    </fieldset>
    <?= $this->Form->button(__('Submit')) ?>
    <?= $this->Form->end() ?>
```

Patching the data according to this code:

```php
$event = $this->Events->newEntity();
if ($this->request->is('post')) {
   $event = $this->Events->patchEntity($event, $this->request->getData(), ['associated' => ['Dates']]);
   debug($event);
   if ($this->Events->save($event, ['associated' => ['Dates']])) {
[...]

PR #11860, opened 2018-03-21
#11772 Saving new hasMany associations records fails due to existsIn build rule

pbfman:
Enhancement to ExistIn build rule to validate association entities when parent entity is new.

See issue https://github.com/cakephp/cakephp/issues/11772


src/ORM/Rule/ExistsIn.php
```diff
@@ -125,6 +126,51 @@ public function __invoke(EntityInterface $entity, array $options)
...
+                        $parentEntity = $parent[strtolower($target->getName())];
+                    }
+                }
+            } else {
+                $parentEntity = $parent;
+            }
+
+            if (isset($parentEntity)) {
+                //extract fields to match
+                $parent_fields = [];
```
ADmad: Please use camelCased variable name `$parentFields`.

src/ORM/Table.php
```diff
@@ -1853,7 +1853,7 @@ protected function _onSaveSuccess($entity, $options)
             $this,
             $entity,
             $options['associated'],
-            ['_primary' => false] + $options->getArrayCopy()
+            ['_primary' => false] + ['parent' => $entity] + $options->getArrayCopy()
```
ADmad: It would be better to prefix the special key `parent` with underscore, like it's done for `_primary`.

ADmad (changes requested):

src/ORM/Rule/ExistsIn.php
```diff
@@ -125,6 +126,51 @@ public function __invoke(EntityInterface $entity, array $options)
             }
         }
 
+        if (isset($options['parent']) && is_a($options['parent'], EntityInterface::class)) {
+            /**
+             * @var EntityInterface $parent
```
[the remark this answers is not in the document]
markstory: Having the full classname helps phpstan and IDEs

src/ORM/Rule/ExistsIn.php
```diff
@@ -125,6 +126,51 @@ public function __invoke(EntityInterface $entity, array $options)
...
+                    //parent is of same  as target
+                    $parentEntity = $parent;
+                } else {
+                    if (!empty($parent[strtolower($target->getName())])) {
+                        $parentEntity = $parent[strtolower($target->getName())];
+                    }
+                }
+            } else {
+                $parentEntity = $parent;
+            }
```
markstory: These conditionals have quite a bit of nesting, would it be possible to move some of that complexity into a protected method that can return early as needed? That would let you have less indentation/complexity and still be able to find the parent entity.

src/ORM/Rule/ExistsIn.php
```diff
@@ -125,6 +126,51 @@ public function __invoke(EntityInterface $entity, array $options)
...
+
+            if (isset($parentEntity)) {
+                //extract fields to match
+                $parent_fields = [];
+                if (is_array($parentEntity)) {
+                    //reagrd HM and HABTM associations
+                    $isNew = true;
+                    foreach ($parentEntity as $record) {
+                        $isNew &= $record->isNew();
```
markstory: Do you need the reference here?

src/ORM/Rule/ExistsIn.php
```diff
@@ -125,6 +126,51 @@ public function __invoke(EntityInterface $entity, array $options)
...
+                }
+            } else {
+                $parentEntity = $parent;
+            }
+
+            if (isset($parentEntity)) {
+                //extract fields to match
+                $parent_fields = [];
+                if (is_array($parentEntity)) {
+                    //reagrd HM and HABTM associations
```
markstory: This isn't clear to me, what are HM and reagrd?

src/ORM/Rule/ExistsIn.php
```diff
@@ -125,6 +126,51 @@ public function __invoke(EntityInterface $entity, array $options)
...
+
+            //extract entity from given parent regarding HM and HABTM associations
+            $parentEntity = null;
+            if ($isAssociation) {
+                if (TableRegistry::get($parent->getSource()) === $target->getTarget()) {
+                    //parent is of same  as target
+                    $parentEntity = $parent;
+                } else {
+                    if (!empty($parent[strtolower($target->getName())])) {
+                        $parentEntity = $parent[strtolower($target->getName())];
```
markstory: You should use the association's `getProperty()` here instead. The default conventions use `lower_case_underscore` for multi-word properties which this approach won't handle well.

markstory (commented):
Do you need a hand with test cases? We require test cases for all new features. Having tests helps make sure existing scenarios continue to work when changes are made in the future.

markstory:
I've moved the target branch to `3.next` as that is where net new features land. It looks like you'll need to rebase your changes. If you need a hand with how to do that let us know.

src/ORM/Rule/ExistsIn.php
pbfman: This is not a reference it is the logic AND
```php
$isNew = $isNew & $record->isNew();
```

src/ORM/Rule/ExistsIn.php
pbfman: typo.
```php
//regard HM and HABTM associations
```

If parentEntity is an array then the association is a HasMany or HasAndBelongsToMany. Therefore there are multiple connected entities (records). If there is at least one new the old ExistsIn would fail for that and therefore for the whole entity. So the following loop extract all foreign keys to check the ids against the parent entity.

src/ORM/Rule/ExistsIn.php
pbfman: But indeed it is wrong, it should be
```php
$isNew |= $record->isNew();
```

pbfman:
Missed to use ``3.next`` sorry for that.

Thank you for your reviews. I've adopted these changes. I'm not pretty sure if I followed the right workflow for you to accept these PR? If not please tell me what to do.

@markstory it would be very nice if you could provide test cases! At the moment I've just tested it live with some data. But indeed a full comprehensive test case would be perfect. Furthermore my changes seems to invalidate some existing test case if I got that right from the automatic checks here.

markstory:
I will add some tests in a few days. The failing tests will need a second look as they usually indicate a regression.

jeremyharris:
@markstory I'd be happy to write the tests if you'd like. The failing tests are just errors due to the new save option being present and the assertions not being aware of it.

markstory:
> I'd be happy to write the tests if you'd like. 

@jeremyharris That would be much appreciated.

jeremyharris:
Hm, I can't seem to get to fail on `3.next` to begin with.

@pbfman would you mind looking at this test case and letting me know if I set it up to replicate the original issue?

```php
    public function testExistsInHasManyNewEntities()
    {
        $table = $this->getTableLocator()->get('Articles');
        $table->hasMany('Comments');
        $table->Comments->belongsTo('Articles');

        $rules = $table->Comments->rulesChecker();
        $rules->add($rules->existsIn(['article_id'], $table));

        $article = $table->newEntity([
            'title' => 'new article',
            'comments' =>  [
                $table->Comments->newEntity([
                    'user_id' => 1,
                    'comment' => 'comment 1',
                ]),
                $table->Comments->newEntity([
                    'user_id' => 1,
                    'comment' => 'comment 2',
                ]),
            ]
        ]);

        $this->assertNotFalse($table->save($article));
    }
```

pbfman:
This test case is correct!
I even tried it by myself and it worked.

I also wrote one similar for BelongsToMany:

```
    public function testExistsInBelongsToManyNewEntities()
    {
        $table = $this->getTableLocator()->get('Articles');
        $table->belongsToMany('Tags');
        $table->Tags->belongsToMany('Articles');

        $rules = $table->rulesChecker();
        $rules->add($rules->existsIn(['tag_id'], $table->Tags));

        $rules = $table->Tags->rulesChecker();
        $rules->add($rules->existsIn(['article_id'], $table));

        $article = $table->newEntity([
            'title' => 'new article',
            'tags' =>  [
                [
                    'name' => 'tag 1',
                ],
                [
                    'name' => 'tag 2',
                ],
            ]
        ]);

        $this->assertNotFalse($table->save($article));
    }
```

It works aswell for ``3.next``.

Even I tried to reconstruct the behavior in my source application and I can't. This is very strange because I got bothered by this problem several times and really digged into the problem. Nevertheless it seems to be a misconfiguration of me even if I cant imagine which....

It works now with hasMany, belongsToMany and all existsIn rules activated.

I am really sorry for wasting your time!

jeremyharris:
@pbfman don't worry, you didn't waste anyone's time! Thanks for your report :)

Closing as this is not an issue.

PR #11862, opened 2018-03-21
Add Psalm, with a lot of issues suppressed

muglug:
This PR adds the necessary fixes to get Psalm (mostly) working.

The idea here is that you'd remove issues (e.g. `InvalidReturnType`) when you want to make some headway.

***Edit***: errors are all accounted for

### DIFF
--- a/psalm.xml
+++ b/psalm.xml
@@ -136,6 +136,7 @@
                 <referencedMethod name="Cake\Datasource\InvalidPropertyInterface::setInvalidField" />
                 <referencedMethod name="Cake\Database\TypedResultInterface::getReturnType" />
                 <referencedMethod name="Cake\Datasource\QueryInterface::andWhere" />
+                <referencedMethod name="Cake\Datasource\QueryInterface::getRepository" />
                 <referencedMethod name="Cake\Datasource\RepositoryInterface::getAlias" />
                 <referencedMethod name="Cake\Core\PluginApplicationInterface::getEventManager" />
                 <referencedMethod name="Cake\Core\PluginApplicationInterface::setEventManager" />
@@ -174,6 +175,7 @@
                 <file name="src/Database/Driver/Sqlserver.php" />
                 <file name="src/Database/Statement/SqlserverStatement.php" />
                 <file name="src/Utility/Crypto/Mcrypt.php" />
+                <file name="src/Cache/Engine/XcacheEngine.php" />
             </errorLevel>
         </UndefinedConstant>
 
--- a/psalm.xml
+++ b/psalm.xml
@@ -1,0 +1,199 @@
+<?xml version="1.0"?>
+<psalm
+    totallyTyped="false"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="vendor/vimeo/psalm/src/config.xsd"
+    autoloader="tests/bootstrap.php"
+>
+    <projectFiles>
+        <directory name="src" />
+        <ignoreFiles>
+            <directory name="src/TestSuite" />
+        </ignoreFiles>
+    </projectFiles>
+
+    <issueHandlers>
+        <LessSpecificReturnType errorLevel="info" />
+
+        <!-- level 3 issues - slightly lazy code writing, but provably low false-negatives -->
+
+        <DeprecatedMethod errorLevel="info" />
+        <DeprecatedProperty errorLevel="info" />
+        <DeprecatedClass errorLevel="info" />
+
+        <MissingClosureReturnType errorLevel="info" />
+        <MissingReturnType errorLevel="info" />
+        <MissingPropertyType errorLevel="info" />
+        <InvalidDocblock errorLevel="info" />
+        <MisplacedRequiredParam errorLevel="info" />
+
+        <PropertyNotSetInConstructor errorLevel="info" />
+        <MissingConstructor errorLevel="info" />
+        <MissingClosureParamType errorLevel="info" />
+        <MissingParamType errorLevel="info" />
+
+        <DocblockTypeContradiction errorLevel="info" />
+        <RedundantConditionGivenDocblockType errorLevel="info" />
+
+        <UnresolvableInclude errorLevel="info" />
+
+        <RawObjectIteration errorLevel="info" />
+
+        <!-- level 4 issues - points to possible deficiencies in logic, higher false-positives -->
+
+        <MoreSpecificReturnType errorLevel="info" />
+        <LessSpecificReturnStatement errorLevel="info" />
+        <TypeCoercion errorLevel="info" />
+        <RedundantCondition errorLevel="info" />
+
+        <PossiblyFalseArgument errorLevel="info" />
+        <PossiblyFalseIterator errorLevel="info" />
+        <PossiblyFalseOperand errorLevel="info" />
+        <PossiblyFalsePropertyAssignmentValue errorLevel="info" />
+        <PossiblyFalseReference errorLevel="info" />
+        <PossiblyInvalidArgument errorLevel="info" />
+        <PossiblyInvalidArrayAccess errorLevel="info" />
+        <PossiblyInvalidArrayAssignment errorLevel="info" />
+        <PossiblyInvalidArrayOffset errorLevel="info" />
+        <PossiblyInvalidFunctionCall errorLevel="info" />
+        <PossiblyInvalidIterator errorLevel="info" />
+        <PossiblyInvalidMethodCall errorLevel="info" />
+        <PossiblyInvalidOperand errorLevel="info" />
+        <PossiblyInvalidPropertyAssignment errorLevel="info" />
+        <PossiblyInvalidPropertyAssignmentValue errorLevel="info" />
+        <PossiblyInvalidPropertyFetch errorLevel="info" />
+        <PossiblyNullArgument errorLevel="info" />
+        <PossiblyNullArrayAccess errorLevel="info" />
+        <PossiblyNullArrayAssignment errorLevel="info" />
+        <PossiblyNullArrayOffset errorLevel="info" />
+        <PossiblyNullFunctionCall errorLevel="info" />
+        <PossiblyNullIterator errorLevel="info" />
+        <PossiblyNullOperand errorLevel="info" />
+        <PossiblyNullPropertyAssignment errorLevel="info" />
+        <PossiblyNullPropertyAssignmentValue errorLevel="info" />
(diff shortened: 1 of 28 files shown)


src/Database/Query.php
```diff
@@ -179,7 +179,7 @@ public function getConnection()
      *
      * @deprecated 3.4.0 Use setConnection()/getConnection() instead.
      * @param \Cake\Database\Connection|null $connection Connection instance
-     * @return $this|\Cake\Database\Connection
+     * @return static|\Cake\Database\Connection
```
dereuromark: Wasn't there a reason (IDEs) that we stuck to the $this for fluent interface and non static?

src/ORM/Behavior/TimestampBehavior.php
```diff
@@ -133,7 +133,7 @@ public function implementedEvents()
      *
      * @param \DateTime|null $ts Timestamp
      * @param bool $refreshTimestamp If true timestamp is refreshed.
-     * @return \DateTime
+     * @return Time
```
dereuromark: FQCN

src/Database/Query.php
muglug: I can fix the code in Psalm that demands this change (only `@return $this` is currently permitted). No good reason, though I do prefer `static` a little.

dereuromark: Afaik static broke the fluent interface approach in IDEs. We would have to test if this change is good.
The deprecated methods should not be in use anymore though, so it most likely won't be too big of a deal either way. And for future methods we sure always try to only use true fluent typehinting (no getter mixin).

dereuromark: But I do see a few methods that are not even deprecated yet, they should be separated into their getter/setter part first IMO.

muglug:
Also should I have Psalm run on the 5.6 build too (because it can)

dereuromark:
> Also should I have Psalm run on the 5.6 build too (because it can)

I don't think that is worth it. We are now all more looking into 7.1+ direction rather than back.

src/Database/Query.php
muglug: it would surprise me if IDEs (i.e. PHPStorm) didn't support `@return static` at this point.

lorenzo: I feel like this changes back and forth every other month

src/Utility/Text.php
```diff
@@ -553,6 +553,7 @@ public static function stripLinks($text)
      * @param string $text String to truncate.
      * @param int $length Length of returned string, including ellipsis.
      * @param array $options An array of options.
+     * @psalm-param array{ellipsis?: string, exact?: bool} $options
```
lorenzo: I like this

psalm.xml
```diff
@@ -0,0 +1,185 @@
+<?xml version="1.0"?>
+<psalm
+    totallyTyped="false"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="/Users/matthewbrown/Desktop/vimeo/git/cakephp/vendor/vimeo/psalm/src/config.xsd"
```
markstory: This location won't exist on most people's computers.

muglug: works on mine. Production-ready code.

src/Controller/Component/AuthComponent.php
```diff
@@ -880,7 +880,7 @@ public function constructAuthenticate()
      *
      * @param \Cake\Auth\Storage\StorageInterface|null $storage Sets provided
      *   object as storage or if null returns configured storage object.
-     * @return \Cake\Auth\Storage\StorageInterface|\Cake\Core\InstanceConfigTrait|null
```
ADmad: Removing `\Cake\Core\InstanceConfigTrait` makes `phpstan` unhappy.
https://travis-ci.org/cakephp/cakephp/jobs/356494536#L732

composer.json
```diff
@@ -42,7 +42,8 @@
     },
     "require-dev": {
         "phpunit/phpunit": "^5.7.14|^6.0",
-        "cakephp/cakephp-codesniffer": "^3.0"
+        "cakephp/cakephp-codesniffer": "^3.0",
+        "vimeo/psalm": "^1.0"
```
ADmad: Since `psalm` is being added as dependency in `travis.yml` itself, adding is here is not necessary. Most contributors aren't going to run it locally anyway.

src/Controller/Component/AuthComponent.php
muglug: This is *definitely* an off-label use of traits, treating them like interfaces. PHP won't allow traits as return types, and I don't see why they should be allowed in a union type.

Reminds me of [this RFC](https://wiki.php.net/rfc/traits-with-interfaces), except without the interface.

I can add `@method` annotations to `StorageInterface` to make PHPStan happy.

composer.json
muglug: Absolutely. Will remove

ADmad:
> ERROR: UndefinedMethod - src/Auth/ControllerAuthorize.php:93:22 - Method Cake\Controller\Controller::isAuthorized does not exist
        return (bool)$this->_Controller->isAuthorized($user);

This is a weird one as the `Controller` doesn't have an `isAuthorized()` method but developer is expected to add on in a child class extending `Controller` when using this `ControllerAuthorize` authorization adapter.

> ERROR: UndefinedMethod - src/Collection/Iterator/FilterIterator.php:71:21 - Method Traversable::getInnerIterator does not exist
        $iterator = $filter->getInnerIterator();

Not too sure about this. Perhaps @lorenzo can help.

> ERROR: AssignmentToVoid - src/Controller/Component/AuthComponent.php:386:9 - Cannot assign $result to type void
        $result = $auth->unauthenticated($controller->request, $response);

The `unauthenticated()` method of authentication adapters can either return `Cake\Http\Response` instance or return void. So maybe updating return types for `BaseAuthenticate::unauthenticated()` can help.

ERROR: InvalidPropertyAssignmentValue - src/Controller/Component/CookieComponent.php:127:30 - $this->request with declared type 'Cake\Http\ServerRequest' cannot be assigned type 'Zend\Diactoros\ServerRequest'
            $this->request = ServerRequestFactory::fromGlobals();

`CookieComponent` is deprecated so I don't think it's worth tackling this.

> ERROR: UndefinedMethod - src/Core/InstanceConfigTrait.php:187:16 - Method Cake\View\StringTemplate::getconfig does not exist
        return $this->getConfig($key);

This one seems like a bug in `psalm`. `StringTemplate` aliases `InstanceConfigTrait::getConfig()` as `get()`.
https://github.com/cakephp/cakephp/blob/3.next/src/View/StringTemplate.php#L33

> ERROR: NullReference - src/Database/Driver.php:334:16 - Cannot call method lastInsertId on null variable $this->_connection
        return $this->_connection->lastInsertId($table, $column);

The `$this->connect();` at start of method is supposed to assign a connection instance to `$this->_connection`.

> ERROR: InvalidArrayOffset - src/I18n/Parser/MoFileParser.php:130:17 - Cannot access value on variable $messages[$singularId]['_context'] using offset value of '_context', expecting int
                $messages[$singularId]['_context'][$context] = $singular;

> ERROR: InvalidArrayOffset - src/I18n/Parser/MoFileParser.php:132:21 - Cannot access value on variable $messages[$pluralId]['_context'] using offset value of '_context', expecting int
                    $messages[$pluralId]['_context'][$context] = $plurals;

No idea about these.

psalm.xml
```diff
@@ -0,0 +1,185 @@
+<?xml version="1.0"?>
+<psalm
+    totallyTyped="false"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="vendor/vimeo/psalm/src/config.xsd"
```
ADmad: Will this work if one is using `psalm.phar` instead of having it as a composer dependency?

muglug:
> The $this->connect(); at start of method is supposed to assign a connection instance to $this->_connection.

Yeah but there's a check for `PDO` and a return. So you have some sort of non-PDO connection?

psalm.xml
muglug: the schemaLocation is just so IDEs can do autocomplete on the XML.

src/Controller/Component/AuthComponent.php
ADmad: I am happy with whatever solution which makes both `phpstan` and `psalm` happy.

Mind you travis runs phpstan at level 2 but we are slowing working towards having it pass level 3. I hope the changes you make for `psalm` don't increases the errors reported for phpstan level 3. :slightly_smiling_face:

composer.json
```diff
@@ -65,7 +65,6 @@
             "TestPluginTwo\\": "tests/test_app/Plugin/TestPluginTwo/src/",
             "Company\\TestPluginThree\\": "tests/test_app/Plugin/Company/TestPluginThree/src/",
             "Company\\TestPluginThree\\Test\\": "tests/test_app/Plugin/Company/TestPluginThree/tests/",
-            "ParentPlugin\\": "tests/test_app/Plugin/ParentPlugin/src/",
```
ADmad: Why this change?

psalm.xml
ADmad: Okay.

composer.json
muglug: checkout out `master` branch composer.json by mistake. Fixed

ADmad:
> Yeah but there's a check for PDO and a return. So you have some sort of non-PDO connection?

Nope, in case of `Driver` and child classes it will always be `PDO` instance.

src/Auth/BaseAuthenticate.php
```diff
@@ -239,7 +239,7 @@ public function getUser(ServerRequest $request)
      *
      * @param \Cake\Http\ServerRequest $request A request object.
      * @param \Cake\Http\Response $response A response object.
-     * @return void
+     * @return \Cake\Http\Response|null
```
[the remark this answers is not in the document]
muglug: This was fixed (`return null` added)

src/Controller/Component/AuthComponent.php
muglug: > I hope the changes you make for psalm don't increases the errors reported for phpstan level 3

I can't imagine how they would.

ADmad:
This looks good now expect for the `$this` to `static` change.

muglug:
@ADmad I've reverted the `static` changes. They keyword still sits in `DateFormatTrait`, but it wasn't me what did that.

ADmad (approved):

Issue #11818, opened 2018-03-12
Security: increased server load with XXE with internal references


dorxy:
This is a:
* [x] bug
* [x] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.5.12 #a6c47f15e2dfbb062fcc547d6eecd0a944dc8b17.
* Platform and Target: nginx 1.10.3, MariaDB 10.0.31, PHP 7.0.22-0ubuntu0.16.04.1

### Context
Our application was PEN-tested and their requests containing the below XML example did not pass inspection.

### What you did
Parse the following XML using `XML::build`:
```xml
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE item [
  <!ENTITY item "item">
  <!ENTITY item1 "&item;&item;&item;&item;&item;&item;">
  <!ENTITY item2 "&item1;&item1;&item1;&item1;&item1;&item1;&item1;&item1;&item1;">
  <!ENTITY item3 "&item2;&item2;&item2;&item2;&item2;&item2;&item2;&item2;&item2;">
  <!ENTITY item4 "&item3;&item3;&item3;&item3;&item3;&item3;&item3;&item3;&item3;">
  <!ENTITY item5 "&item4;&item4;&item4;&item4;&item4;&item4;&item4;&item4;&item4;">
  <!ENTITY item6 "&item5;&item5;&item5;&item5;&item5;&item5;&item5;&item5;&item5;">
  <!ENTITY item7 "&item6;&item6;&item6;&item6;&item6;&item6;&item6;&item6;&item6;">
  <!ENTITY item8 "&item7;&item7;&item7;&item7;&item7;&item7;&item7;&item7;&item7;">
]>
<item>
  <description>&item8;</description>
</item>
```

### What happened
When interchanging the item in the <description> tag the memory used for parsing peaks depending on the index used. Using `&item8;` is very noticeable. On our production servers a request using this XML body takes 8 seconds to respond.

This is due to the following:
- The RequestHandlerComponent uses `Xml::build` in its `convertXml` function with default options.
- Default build options prevent entity expansion using `libxml_disable_entity_loader` but as mentioned in the PHP docs, this only handles loading of *external* entities.

### What probably should happen
- The default return type of `XML::_loadXml` is SimpleXML. There is no parameter to prevent SimpleXML from loading internal entity references, making it increase server load.
- If I load the above example into a  `\DOMDocument` object with the libxml entity loader disabled, it gives the following warning:
[...]

PR #11863, opened 2018-03-22
Fix nested entity expansion.

markstory:
By using DOMDocument we can avoid internal entity expansion issues that SimpleXMLElement will not prevent.

I've added some tests around XML conversion as there weren't any before.

Refs #11818


lorenzo:
some tests are failing

PR #11865, opened 2018-03-22
Allow context to be multiple lines in po files

Wouter0100:
Currently, the following entry in a po file throws a warning:
```
msgctxt ""
"Description for push notifications on create message with only location"
msgid "notifications.messages.create.description.location"
msgstr "{1}: 📌 {0}"
```

This is because the `msgctxt` is currently not correctly implemented to handle multiple lines. This MR fixes it.


markstory (commented):
Any chance we could get a test/update to a sample po file that covers this scenario?

Wouter0100:
Updated the default po file used for testing. `Cake\Test\TestCase\I18n\Parser\PoFileParserTest::testParse` will fail without this fix.

markstory:
Thanks!

PR #11866, opened 2018-03-22
Fixed default adapter

davidyell:
This bug prevented the client being instantiated with empty options.

```php
$client = new Client();
// Undefined index: adapter
```

### DIFF
--- a/src/Http/Client.php
+++ b/src/Http/Client.php
@@ -159,10 +159,14 @@ class Client
      */
     public function __construct($config = [])
     {
+        if (!isset($config['adapter'])) {
+            $adapter = $this->_defaultConfig['adapter'];
+        } else {
+            $adapter = $config['adapter'];
+        }
+        $this->setConfig('adapter', $adapter);
         $this->setConfig($config);
 
-        $adapter = $this->_config['adapter'];
-        $this->setConfig('adapter', null);
         if (is_string($adapter)) {
             $adapter = new $adapter();
         }
--- a/tests/TestCase/Http/ClientTest.php
+++ b/tests/TestCase/Http/ClientTest.php
@@ -59,6 +59,12 @@ class ClientTest extends TestCase
         }
     }
 
+    public function testConstructWithEmptyOptions()
+    {
+        $http = new Client();
+        $this->assertEquals(\Cake\Http\Client\Adapter\Stream::class, $http->getConfig('adapter'));
+    }
+
     /**
      * Data provider for buildUrl() tests
      *


tests/TestCase/Http/ClientTest.php
```diff
@@ -59,6 +59,12 @@ public function testConstructConfig()
         }
     }
 
+    public function testConstructWithEmptyOptions()
+    {
+        $http = new Client();
+        $this->assertEquals('Cake\Http\Client\Adapter\Stream', $http->getConfig('adapter');
```
dereuromark: You don't want to use ::class syntax?

ADmad: A closing brace `)` is missing, causing parse error.

davidyell:
Oops, sorry edited it on Github.

I could update it to `::class` I was just copying and pasting from the class for speed.

dereuromark:
It would be better, also for IDE discovery and refactoring.

src/Http/Client.php
```diff
@@ -161,7 +161,7 @@ public function __construct($config = [])
     {
         $this->setConfig($config);
 
-        $adapter = $this->_config['adapter'];
+        $adapter = $this->_defaultConfig['adapter'];
```
markstory: How would someone change the adapter? I recognize that is not possible right now, but given the https proxy issues PHP streams have I plan to build a curl adapter in the near future.

davidyell: Yes, after I had submitted this pull request, I thought a little more. I had the same thought, that perhaps the intended functionality was to set the default to the config before merging the userland configuration.

Would it be worth updating to something like this?
```php
        $adapter = $this->_defaultConfig['adapter'];
        $this->setConfig('adapter', $adapter);
        $this->setConfig($config);

        //$adapter = $this->_defaultConfig['adapter'];
        $this->setConfig('adapter', null);
        if (is_string($adapter)) {
```

davidyell:
Thanks for the feedback, I have updated the pull request, and squashed the commits.

tests/TestCase/Http/ClientTest.php
```diff
@@ -59,6 +59,18 @@ public function testConstructConfig()
...
+    public function testConstructWithEmptyOptions()
+    {
+        $http = new Client();
+        $this->assertEquals('Cake\Http\Client\Adapter\Stream', $http->getConfig('adapter'));
+    }
+
+    public function testConstructWithEmptyOptions()
+    {
+        $http = new Client();
+        $this->assertEquals(\Cake\Http\Client\Adapter\Stream::class, $http->getConfig('adapter'));
```
dereuromark: no fqcn inside code please.

tests/TestCase/Http/ClientTest.php
```diff
@@ -59,6 +59,18 @@ public function testConstructConfig()
         }
     }
 
+    public function testConstructWithEmptyOptions()
+    {
+        $http = new Client();
+        $this->assertEquals('Cake\Http\Client\Adapter\Stream', $http->getConfig('adapter'));
```
dereuromark: Your squash just have gone wrong?

src/Http/Client.php
```diff
@@ -159,10 +159,14 @@ class Client
      */
     public function __construct($config = [])
     {
+        if (!isset($config['adapter'])) {
+            $adapter = $this->_defaultConfig['adapter'];
+        } else {
+            $adapter = $config['adapter'];
+        }
+        $this->setConfig('adapter', $adapter);
```
ADmad: This is incorrect, once the adapter is extract it's supposed to be unset from config (set to null) like it was before.

tests/TestCase/Http/ClientTest.php
```diff
@@ -59,6 +59,12 @@ public function testConstructConfig()
         }
     }
 
+    public function testConstructWithEmptyOptions()
+    {
+        $http = new Client();
+        $this->assertEquals(\Cake\Http\Client\Adapter\Stream::class, $http->getConfig('adapter'));
```
dereuromark: FQCN must be in use statement.

ADmad:
I just tested with existing code and doing `$client = new Client();` does not generate any error. So not sure what you are trying to fix.

davidyell:
I realised that I'm working on 3.next locally. However the same code was in `master` when I checked.

ADmad:
So can you can provide a way to reproduce the issue or not?

davidyell:
This is the line which caused the error for me

https://github.com/davidyell/Webservice/commit/8f52ee0271313cfe002d5c3066cfba2aa3af9095#diff-47a9d5d384ab4d71dfff6909149058c6R39

ADmad:
Well you need to fix your mock or better yet don't unnecessarily mock stuff. Closing as there is nothing to fix here.

PR #11867, opened 2018-03-22
SecurityComponent regression

jeremyharris:
This fixes a regression in the SecurityComponent which previously used `$request->here()` as the URL to validate the token with, but now uses `$request->getRequestTarget()`. 

There are a couple of issues as I see it:

1. `here()` didn't encode the query string, while `UriInterface` dictates "The value returned MUST be percent-encoded"
2. `getRequestTarget()` doesn't prepend the base, where `here()` does it by default. I'm not too keen on the inverse `$base` flag, but applying the base by default caused a lot of tests to fail. Should I make the signatures the same and fix the tests?
3. The URI created by ServerInterface didn't seem to add query args that were passed in the config array, it just added query args if they were in the URL string

I'm not sure about the new `$base` flag. I'm wondering if `getRequestTarget()` should remain the same (never include base) and just prepend the base where necessary, or if it's better to have a more compatible replacement for `here()`.

### DIFF
--- a/src/Http/ServerRequest.php
+++ b/src/Http/ServerRequest.php
@@ -337,7 +337,10 @@ class ServerRequest implements ArrayAccess, ServerRequestInterface
         if (strlen($config['url'])) {
             $uri = $uri->withPath('/' . $config['url']);
         }
-        if (strlen($querystr)) {
+        if (strlen($querystr) || isset($config['query'])) {
+            if (!strlen($querystr)) {
+                $querystr = http_build_query($config['query']);
+            }
             $uri = $uri->withQuery($querystr);
         }
 
@@ -2324,9 +2327,10 @@ class ServerRequest implements ArrayAccess, ServerRequestInterface
      * application relative path without base directory, and the query string
      * defined in the SERVER environment.
      *
+     * @param bool $base Include the base path, set to false to trim the base path off.
      * @return string
      */
-    public function getRequestTarget()
+    public function getRequestTarget($base = false)
     {
         if ($this->requestTarget !== null) {
             return $this->requestTarget;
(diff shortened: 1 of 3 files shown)


src/Http/ServerRequest.php
```diff
@@ -2324,23 +2327,28 @@ public function withRequestTarget($target)
      * application relative path without base directory, and the query string
      * defined in the SERVER environment.
      *
+     * @param bool $base Include the base path, set to false to trim the base path off.
      * @return string
      */
-    public function getRequestTarget()
+    public function getRequestTarget($base = false)
```
ADmad: `getRequestTarget()` is an interface method declared in `Psr\Http\Message\ServerRequestInterface`, you can't change it's signature.

jeremyharris: Ah no wonder there's a ton of failures, I missed that bit.

jeremyharris:
I'm finding some more confusing things so I want to get a clear idea of how this all works before digging into this too much more.

First there's this test: https://github.com/cakephp/cakephp/blob/master/tests/TestCase/Http/ServerRequestTest.php#L147-L159

This says that passing query data into the constructor should **not** add it to the internal URI object. Is that correct?

Then this test: https://github.com/cakephp/cakephp/blob/master/tests/TestCase/Http/ServerRequestTest.php#L1916-L1933

Which says that the query (well, `getQueryParams` anyway) isn't available when built from globals, however, the internal URI object will have the query still.

All of this to say I'm a little confused as to why ServerRequest's state is different than the internal object's state. This difference in state is what caused the original bug, where query arguments were originally not encoded (`here()`) and now they are (`getRequestTarget()`), causing secure actions (with query strings) to fail the SecurityComponent's check.

markstory:
Was the regression around tokens changing? They will intentionally change as per the 3.6 migration guide

> Tokens generated by SecurityComponent now include the user's session id, to prevent token reuse between users/sessions. This changes the value of security tokens and will cause forms created in earlier versions of CakePHP to fail validation in 3.6.

jeremyharris:
Closed in favor of https://github.com/cakephp/cakephp/pull/11868

I'm still a little concerned about how the states differ, and how `->here()` doesn't match `->getRequestTarget()` in many instances, but I think that's a different issue.

PR #11871, opened 2018-03-23
fixes #11870 Add DboSource::expression support for Dbo VirtualFields

### LINKED ISSUE #11870: No support for  DboSource::expression in Virtual fields
This is a (multiple allowed):

* [x] bug
* [ ] enhancement

* CakePHP Version: 2.10
* Platform and Target: PHP7 / Mysql 

### What you did
I can't use DboSource::expression in Virtual fields

### What happened
No support for this

in \DboSource::_constructVirtualFields
add condition for process expressions

giboow:
Add DboSource::expression support for Dbo VirtualFields

### DIFF
--- a/lib/Cake/Test/Case/Model/Datasource/DboSourceTest.php
+++ b/lib/Cake/Test/Case/Model/Datasource/DboSourceTest.php
@@ -1650,6 +1650,42 @@ class DboSourceTest extends CakeTestCase {
 		$this->assertEquals($expected, $result[0]);
 	}
 
+	/**
+	 * Test conditionKeysToString() with virtual field
+	 *
+	 * @return void
+	 */
+	public function testConditionKeysToStringVirtualFieldExpression() {
+		$Article = ClassRegistry::init('Article');
+		$Article->virtualFields = array(
+			'extra' => DboSource::expression('something virtual')
+		);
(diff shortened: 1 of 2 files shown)


lib/Cake/Test/Case/Model/Datasource/DboSourceTest.php
```diff
@@ -1650,6 +1650,42 @@ public function testConditionKeysToString() {
...
 
+	/**
+	 * Test conditionKeysToString() with virtual field
+	 *
+	 * @return void
+	 */
+	public function testConditionKeysToStringVirtualFieldExpression() {
+		$Article = ClassRegistry::init('Article');
+		$Article->virtualFields = array(
+			'extra' => DboSource::expression('something virtual')
```
markstory: This isn't a static method. You need to use an instance like `$Articles->getDataSource()->expression()`.

markstory (changes requested):

lib/Cake/Test/Case/Model/Datasource/DboSourceTest.php
giboow: Sorry for the mistake. I will fix this soon

PR #11872, opened 2018-03-24
Improve Csrf component and middleware by encrypting and validating tokens against and expiry date

rchavik:

### DIFF
--- a/src/Http/Middleware/CsrfProtectionMiddleware.php
+++ b/src/Http/Middleware/CsrfProtectionMiddleware.php
@@ -133,7 +133,13 @@ class CsrfProtectionMiddleware
      */
     protected function _createToken()
     {
-        return hash('sha512', Security::randomBytes(16), false);
+        $expiry = new Time($this->_config['expiry']);
+        $data = json_encode([
+            'token' => hash('sha512', Security::randomBytes(16), false),
+            'expiry' => $expiry->format('U'),
+        ]);
+
+        return bin2hex(Security::encrypt($data, Security::getSalt()));
     }
 
     /**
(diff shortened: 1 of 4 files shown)


src/Http/Middleware/CsrfProtectionMiddleware.php
```diff
@@ -193,5 +199,19 @@ protected function _validateToken(ServerRequest $request)
...
+                throw new InvalidCsrfTokenException(__d('cake', 'Invalid CSRF token'));
+            }
+            $expiry = new Time('@' . $decrypted['expiry']);
+            $now = new Time();
+            if ($now > $expiry) {
+                throw new InvalidCsrfTokenException(__d('cake', 'Expired CSRF token'));
+            }
+        } catch (\InvalidArgumentException $e) {
+            throw new InvalidCsrfTokenException(__d('cake', 'Invalid CSRF token'));
+        }
```
markstory: Should we consider extracting the token creation/validation to a helper object that can be used by the middleware and component? It could be an internal class so we don't expose more api surface.

markstory (commented):
Going to move to 3.6 as changing token creation logic will cause in-flight tokens to fail which doesn't feel like something we should be doing in a bugfix release.

markstory:
Looks like there is some missing coverage/tests on the new failure paths.

lorenzo:
Can you explain why this is necessary and how can older apps migrate to this new thing?

Users having the old cookie set would suddenly get denied requests.

rchavik:
> Can you explain why this is necessary and how can older apps migrate to this new thing?

Sorry lorenzo, can you look at these archive for now as I'm on something else.

https://cakesf.slack.com/archives/G0P0V20H0/p1521775802000119
https://cakesf.slack.com/archives/G0P0V20H0/p1521812236000531
https://cakesf.slack.com/files/U0TL27UHW/F9VH5KWS1/image.png
https://cakesf.slack.com/archives/G0P0V20H0/p1521855754000031

lorenzo:
I read the discussion, but I'm not still convinced this is improving anything.

The addition of an expiry field makes this similar to what we had in cake 2, making users' life miserable as the expiry would randomly prevent the sending of forms. That was the reason we decided for the token to be long lived and use the double submit strategy instead of storing in the session.

What is this actually preventing?

rchavik:
@lorenzo how do feel about https://github.com/cakephp/cakephp/compare/master...rchavik:session-csrf as alternative? Still need to invalidate existing sessions though.

PR #11873, opened 2018-03-25
Paginator sort links should go to page 1 when changing order

AD7six:
When looking at page n of a  paginated index - the sort links generated by the paginator component keep the page arg, wheras it'd be more logical to always go to page one. On small data sets it won't be noticable/make any difference, on large data sets this behavior is confusing. 

## Example

After following the blog tutorial create a bunch of articles with random titles.

```
truncate articles;

delimiter //
CREATE PROCEDURE createArticles(num INT)
    BEGIN
        SET @i = 0;
        REPEAT
            SET @i = @i + 1;
            INSERT INTO articles (title,body,created) VALUES ('The title', 'This is the article body.', NOW());
        UNTIL @i > num END REPEAT;
    END
    //

delimiter ;
CALL createArticles(1000);
DROP PROCEDURE createArticles;

-- Randomize titles
UPDATE articles set title = CONCAT('RANDOM ', RAND() * 1000);
```

Sort by any column and use page links to arrive at e.g. 

http://cake.local:8765/articles?page=5&sort=title&direction=asc
screenshot from 2018-03-25 16-21-32

Then use the sort header to sort by a different field - the page parameter is preserved:

http://cake.local:8765/articles?page=5&sort=id&direction=asc
screenshot from 2018-03-25 16-21-12

http://cake.local:8765/articles?sort=id&direction=asc
screenshot from 2018-03-25 16-22-39

This PR removes the step in the middle.

Why? 

Well, if you're looking at a paginated index and changing the sort order it's most likely because the changed sort allows you to find the records you weren't able to find with the sort order on a different field/direction - it doesn't make any sense to persist the page parameter when changing sort, as the records shown are then effectively random.

### DIFF
--- a/tests/TestCase/View/Helper/PaginatorHelperTest.php
+++ b/tests/TestCase/View/Helper/PaginatorHelperTest.php
@@ -840,12 +840,12 @@ class PaginatorHelperTest extends TestCase
     public function urlGenerationResetsToPage1Provider()
     {
         return [
-            'Sorting the field currently sorted asc' => [
+            'Sorting the field currently sorted asc, asc' => [
                 'name',
                 ['sort' => 'name', 'direction' => 'asc'],
                 '<a class="asc" href="/index?sort=name&amp;direction=asc">Name</a>'
             ],
-            'Sorting the field currently sorted desc' => [
+            'Sorting the field currently sorted asc, desc' => [
                 'name',
                 ['sort' => 'name', 'direction' => 'desc'],
                 '<a class="asc" href="/index?sort=name&amp;direction=desc">Name</a>'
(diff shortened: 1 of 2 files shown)


tests/TestCase/View/Helper/PaginatorHelperTest.php
```diff
@@ -842,7 +842,7 @@ public function testGenerateUrlWithPrefixes()
 
         $result = $this->Paginator->sort('name', null, ['url' => $options]);
         $expected = [
-            'a' => ['href' => '/members/posts/index?page=2&amp;sort=name&amp;direction=asc'],
+            'a' => ['href' => '/members/posts/index?sort=name&amp;direction=asc'],
```
markstory: Do you think these tests are obvious enough? I know that I would be likely to miss this given it is inside another test case method.

AD7six: I'll take a closer look at the test, these changes were the minimum to get 0 fails :)

AD7six: I have added new tests to clarify/highlight expectations.

PR #11875, opened 2018-03-26
Add Middleware for parsing body data based on content type.

markstory:
This adds a middleware based option for the automatic request body parsing that RequestComponentHelper provides. A middleware option is required to enable other middleware (like authentication) access to JSON/XML based request payloads.

Refs cakephp/authentication#185

### DIFF
--- a/src/Controller/Component/RequestHandlerComponent.php
+++ b/src/Controller/Component/RequestHandlerComponent.php
@@ -212,6 +212,9 @@ class RequestHandlerComponent extends Component
             return;
         }
 
+        if ($request->getParsedBody() !== []) {
+            return;
+        }
         foreach ($this->getConfig('inputTypeMap') as $type => $handler) {
             if (!is_callable($handler[0])) {
                 throw new RuntimeException(sprintf("Invalid callable for '%s' type.", $type));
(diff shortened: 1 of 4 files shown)


lorenzo:
What happens if both this middleware and the component are present in a json payload request?

markstory:
> What happens if both this middleware and the component are present in a json payload request?

The request body will be parsed twice. The request body payload is inside a buffered stream which will be read by both layers.

Longer term, I think RequestHandlerComponent should go away, and be replaced by a component that is focused on layout/view switching as all the other responsibilities of this component have been deprecated - `beforeRedirect` handling, or moved to other places - `request->is()`, and this change.

lorenzo:
@markstory Can we set a custom arg to the request the RequestHandlerComponent can read and avoid doing the same work twice? I don't think a lot of people will remove the component after migrating to 3.6

markstory:
@lorenzo We could skip parsing input if `getParsedBody()` returns a non-empty value. That would let us not have to add an additional config option.

lorenzo:
@markstory That sounds good to me!

chrisShick:
@markstory If we were to remove the RequestHandler, would we also not need another middleware for response as ?

markstory:
@lorenzo Updated.

PR #11878, opened 2018-03-28
using JSON_UNESCAPED_UNICODE with json_encode()

mosaxiv:
It is useful for debug

```
{
    "id": 1,
    "word": "\u30c6\u30b9\u30c81",
    "created": "2018-03-29T01:41:58",
    "modified": "2018-03-29T01:41:58"
}
```

↓

```
{
    "id": 1,
    "word": "テスト1",
    "created": "2018-03-29T01:41:58",
    "modified": "2018-03-29T01:41:58"
}
```

### DIFF
--- a/src/Database/README.md
+++ b/src/Database/README.md
@@ -55,7 +55,7 @@ directly in the options array:
 use Cake\Database\Connection;
 
 $connection = new Connection([
-	'driver' => 'Cake\Database\Driver\Sqlite'
+	'driver' => 'Cake\Database\Driver\Sqlite',
 	'database' => '/path/to/file.db'
 ]);
 ```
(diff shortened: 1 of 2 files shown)


dereuromark:
This sounds like it would be safer to apply to 3.next (3.6).

markstory:
Couldn't this cause issues with some JSON parsers?

dereuromark:
Could we move the constant integer hardcoded here up into a protected class const? 
This way it could be overwritten on project level more easily?

    const OPTIONS = JSON_PRETTY_PRINT | JSON_UNESCAPED_UNICODE;

etc?

markstory:
Closing this as I'm not comfortable with the potential risk this presents. For applications that need this option you can use the `_jsonOptions` flag to control the options passed to `json_encode()`.

PR #11884, opened 2018-03-30
Write some validation method and replaced deprecated method with new one!

sohelrana820:
Write some validation method and replaced the deprecated method with the new one!


markstory:
Thanks!

Issue #11795, opened 2018-03-06
Add new fetch methods


odan:
This is a (multiple allowed):

* [ ] bug
* [x] enhancement
* [x] feature-discussion (RFC)

* CakePHP Version: 3.x
* Platform and Target: All

### What you did

Example code:

```php
$driver = new \Cake\Database\Driver\Mysql($config);
$connection = new \Cake\Database\Connection(['driver' => $driver]);

$query = $connection->newQuery()->from('books')->select(['id', 'title']);
$rows = $query->execute()->fetch('assoc');
```

### What happened
I try to get the result as array. It would be nice to have a special and strongly typed method to fetch arrays, objects and field values. Passing the string `assoc` is not very pretty, error-prone, redundant and requires more typing. 

### What you expected to happen

It would be very practical to offer these new methods in the [StatementInterface](https://github.com/cakephp/cakephp/blob/master/src/Database/StatementInterface.php).

```php
public function fetchAssoc(): array;
public function fetchColumn(int $column = 0);
public function fetchObject(string $class = 'stdClass');
public function fetchObjects(string $class = 'stdClass'): array;
```

### Usage

```php
// fetch all rows as array
$rows = $query->execute()->fetchAssoc();

// fetch the column value of the first row
$column = $query->execute()->fetchColumn(0);

// Fetch the first row as stdClass
$object = $query->execute()->fetchObject();

// Fetch all rows as array of stdClass
$objects = $query->execute()->fetchObjects();
```

odan:
Thank you very much for adding the new `public function fetchAssoc()` method.

My only concern is that it does not always return an array, because `false` is a boolean data type. I don't think a return type should be mixed. Instead, it would be better to return an empty array than `false`. Checking the result if it's not `false` requires more casting logic after the query.

Is there a way to set the return type to array (in every case)?

`public function fetchAssoc(): array`

markstory:
>Is there a way to set the return type to array (in every case)?

There is probably a way. Just needs some more code :smile:

PR #11886, opened 2018-03-31
Fetch methods 11795

T0mats0:
This adds the methods described in issue #11795

I did not add fetchClass() because it would have required using setFetchMode() to map the class to the results and I think doing so would require some interface changes. This pull request uses existing structures and is BC.


saeideng:
you selected wrong  branch

saeideng:
target changed to 3.next

src/Database/Statement/BufferedStatement.php
```diff
@@ -106,6 +105,22 @@ public function fetch($type = 'num')
...
+     */
+    public function fetchAssoc()
+    {
+        return $this->fetch(parent::FETCH_TYPE_ASSOC);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function fetchobject()
```
dereuromark: Did you mean `fetchObject`?

src/Database/Statement/PDOStatement.php
```diff
@@ -90,13 +90,13 @@ public function bindValue($column, $value, $type = 'string')
      */
     public function fetch($type = 'num')
```
dereuromark: You could then also change this magic string.

src/Database/Statement/StatementDecorator.php
```diff
@@ -35,6 +35,25 @@ class StatementDecorator implements StatementInterface, Countable, IteratorAggre
...
+     */
+    const FETCH_TYPE_NUM = 'num';
+    /**
+     * Used to designate that associated array be returned in a result when calling fetch methods
+     *
+     * @var string
+     */
+    const FETCH_TYPE_ASSOC = 'assoc';
+    /**
+     * Used to designate that objects of \StdClass be returned in a result when calling fetch methods
```
dereuromark: stdClass no StdClass

src/Database/Statement/BufferedStatement.php
xaphalanx: Yes Sir

src/Database/Statement/PDOStatement.php
xaphalanx: Wonder if a change to the statement directory where these strings are used is called for in this case.

src/Database/Statement/BufferedStatement.php
```diff
@@ -149,7 +149,7 @@ public function rowCount()
     {
         if (!$this->_allFetched) {
             $counter = $this->_counter;
-            while ($this->fetch('assoc')) {
+            while ($this->fetch(parent::FETCH_TYPE_ASSOC)) {
```
dereuromark: These runtime ones should be static:: IMO.

xaphalanx: They should be.

xaphalanx: Thanks for you feedback today.

src/Database/Statement/StatementDecorator.php
```diff
@@ -210,11 +263,21 @@ public function fetch($type = 'num')
...
     {
         return $this->_statement->fetchAll($type);
     }
 
+    /**
+     * Returns an array of \stdClass objects or false
+     *
+     * @return \stdClass[]|false
+     */
+    public function fetchAllObjects()
```
markstory: Why does `fetchAllObjects()` require a method? The other 'all' types don't have their own methods.

src/Database/Statement/StatementDecorator.php
```diff
@@ -35,6 +35,25 @@ class StatementDecorator implements StatementInterface, Countable, IteratorAggre
...
+     *
+     * @var string
+     */
+    const FETCH_TYPE_ASSOC = 'assoc';
+    /**
+     * Used to designate that objects of \stdClass be returned in a result when calling fetch methods
+     *
+     * @var string
+     */
+    const FETCH_TYPE_OBJ = 'obj';
```
lorenzo: do you use this mode in your projects?

I think having it creates a lot of confusion, and makes certain things like using collection methods more difficult.

markstory: Should we map results into stdclass like object that implements arrayaccess as well?

xaphalanx: I made the assumption that if you wanted to call one of the methods you probably were wanting lower level pdo access. This pull request doesn't fully address it but does provide a couple options and abstracts the the strings (num, assoc) out of the way.

src/Database/Statement/StatementDecorator.php
xaphalanx: Happy to add them.

markstory: Or remove this one? We usually lean towards exposing fewer methods to start with. We can always add more later, but taking methods away takes a very long time.

src/Database/Statement/StatementDecorator.php
lorenzo: I like the other changes, but I’m unsure about this one.

xaphalanx:
As per Mark's and Lorenzo's comments I removed fetchObject() and fetchAllObjects() and got rid of the FETCH_TYPE_OBJECT constant.

src/Database/Statement/BufferedStatement.php
```diff
@@ -106,13 +106,21 @@ public function fetch($type = 'num')
         return $this->_records[] = $record;
     }
 
+    /**
+     * {@inheritdoc}
+     */
+    public function fetchAssoc()
```
dereuromark: Do we need this method if we can simply do `fetch(..::class)`?

xaphalanx: The only thing better than method('string') would be  method(static::SOME_CONST) and better still would be methodConst(). It's just easier to call. 

My guess is that the original issue (11795) was about wanting more access to PDO statement object. Dbal and other lightweight PDO abstractions generally provide these short cut methods fetchAssoc/fetchColumn/fetchObject.

xaphalanx: I can see your point. 

Doesnt fetch() = fetchNum() because num is the default.

dereuromark: Sure, there is a default on fetch(), but do use cases warrant an own method for a non default option that can easily be done using the primary one?

tests/TestCase/Database/QueryTest.php
```diff
@@ -4697,4 +4669,75 @@ public function assertQuotedQuery($pattern, $query, $optional = false)
...
+                'is_active'
+            ])
+            ->from('profiles');
+        $statement = $query->execute();
+        $results = $statement->fetchAssoc();
+        $this->assertEquals('1', $results['id']);
+        $results = $statement->fetchAssoc();
+        $this->assertEquals('2', $results['id']);
+        $results = $statement->fetchColumn(0);
+        $this->assertEquals('3', $results[0]);
```
dereuromark: These should always be assertSame(), to avoid numeric 2 etc to also false positive here.

xaphalanx: Thanks for the tip.

src/Database/Statement/BufferedStatement.php
xaphalanx: Doesn't seem to be a right or wrong answer here for me, It seems a matter of preference. I get what you are saying. I like the methods they are easy to find in in my ide and I don't have to know about the constants or strings. These new methods are about convenience to me. 

Someone let me know and I throw it one way or the other.

markstory:
There are a few failing tests.

tests/TestCase/Database/QueryTest.php
```diff
@@ -4697,4 +4669,75 @@ public function assertQuotedQuery($pattern, $query, $optional = false)
...
+                'id',
+                'user_id',
+                'is_active'
+            ])
+            ->from('profiles')
+            ->where(['id' => 2])
+            ->limit(1);
+        $statement = $query->execute();
+        $results = $statement->fetchColumn(0);
+        $this->assertSame('2', $results[0]);
```
markstory: Why is fetchColumn() returning an array? Shouldn't it return the scalar value in the column?

src/Database/Statement/BufferedStatement.php
dereuromark: I let @markstory or @lorenzo decide on it as well.

markstory: I think having `fetchAssoc()` is reasonable as the use cases of `fetchNum()` are low enough that it doesn't warrant having a method right now. `fetchObj()` was removed because it causes problems with `Collection` related methods.

tests/TestCase/Database/QueryTest.php
xaphalanx: It's not actually returning an array, it's returning a string with 1 character and passing. I updated test so it is correct now. I also corrected the code got  rid of the last 2 places where the FETCH_TYPE_OBJ  constants were located. 

I wonder if I should update the fetch/fetchall docblocks to show the use of the constants instead of the strings ('num', 'assoc') or just leave it as is? Does anyone have a preference?

src/Database/Statement/BufferedStatement.php
xaphalanx: Leaving the new methods/constants as they are in this commit.

tests/TestCase/Database/QueryTest.php
markstory: > I wonder if I should update the fetch/fetchall docblocks to show the use of the constants instead of the strings ('num', 'assoc') or just leave it as is?

Constants are better imo, as they fail hard on copy/paste typo errors.

src/Database/Statement/PDOStatement.php
```diff
@@ -88,17 +88,14 @@ public function bindValue($column, $value, $type = 'string')
...
+        if ($type === static::FETCH_TYPE_NUM) {
             return $this->_statement->fetch(PDO::FETCH_NUM);
         }
-        if ($type === 'assoc') {
+        if ($type === static::FETCH_TYPE_ASSOC) {
             return $this->_statement->fetch(PDO::FETCH_ASSOC);
         }
-        if ($type === 'obj') {
-            return $this->_statement->fetch(PDO::FETCH_OBJ);
-        }
```
markstory: Won't removing this case break userland code that is currently passing 'obj'?

markstory: I think having a constant for the `obj` type is ok and good. I previously only meant to recommend removing the obj methods.

phpunit.xml.dist
```diff
@@ -50,9 +50,9 @@
         <!-- Postgres
         <env name="db_dsn" value="postgres://localhost/cake_test?timezone=UTC"/>
         -->
-        <!-- Mysql
-        <env name="db_dsn" value="mysql://localhost/cake_test?timezone=UTC"/>
-        -->
+
+        <env name="db_dsn" value="mysql://root:root@localhost/cake_test?timezone=UTC"/>
+
```
saeideng: please don't change the  `phpunit.xml.dist` file 
if you need these changes make a copy of this file as  `phpunit.xml` in your local

src/Database/Statement/PDOStatement.php
xaphalanx: I see what you are saying. I'll put the constant back with the conditional for fetching objects.

Should I update fetch/fetchall doc blocks showing usage with constants instead of strings or leave it as is?

PR #11887, opened 2018-03-31
add cakephp version into travis

saeideng:


dereuromark:
What is the point? the version is displayed already on composer install.

saeideng:
@dereuromark not always 
please see https://travis-ci.org/cakephp/cakephp/jobs/360699288#L596

PR #11892, opened 2018-04-01
Add QueryTrait::getRepository().

ADmad:
This was on the list on #9978


markstory:
We should add tests for the new method.

ADmad:
@markstory Done.

Issue #11793, opened 2018-03-06
Proposal for a better to read and write case expression


burzum:
When I read this https://stackoverflow.com/questions/49129058/cakephp-3-query-ifnull I remembered that I disliked this, it's aweful to read and remember.

```php
$query
    ->newExpr()
    ->addCase(
        [
            $query->newExpr()->add([
                'Sales.isforeign' => $isforeign,
                'Sales.source' => $source
            ])
        ],
        [1, 0],
        ['integer', 'integer']
    );
```

So I had a discussion with @lorenzo about this and we agreed to implement something like this:

```php
$case = $query->newCase()
    ->when([
        'Sales.isforeign' => $isforeign,
        'Sales.source' => $source
    ])
    ->then(1)
    ->otherWise('0', 'integer');

$query
    ->find()
    ->addCase($case);
```

The then() and otherWise() *can* take a 2nd arg to define the type. If *not* set the type of the first arg is used by default.

Any other ideas or improvements please?

josegonzalez:
Can you redo that case api with the example you have in the first bit?

burzum:
@josegonzalez done.

inoas:
shouldn't it be `->then(1, 'integer')` then?

burzum:
@ionas 

>  The then() and otherWise() can take a 2nd arg to define the type. If not set the type of the first arg is used by default.

raul338:
I would prefer define the condition and the return in the same call, like this
```php
$query->select([
    'late_status' => $query->newCase()
        ->add(
            $query->newExpr()->lt('task.diffStart', 0),
            'Sooner',
            'literal'
        )
        ->add(
            $query->newExpr()->between('task.diffStart', 0, 15),
            'In Time',
            'literal'
        )
        ->add(
            $query->newExpr()->gt('task.diffStart', 15),
            'Late'  // literal optional
        )
        ->else(
            'Task not started / Canceled',
            'literal'
        )
]);
```

markstory:
@burzum What would multiple cases look like with that API? Should `CaseExpression` be exposed via a builder on `QueryExpression`? The `newCase()` method proposed by @raul338 would be a nice way to not require users to `new` an expression object.

lorenzo:
For multiple ases I would have a completely different expression class. The syntax is so different that we could say they are different things

markstory:
Having 2 expression classes would make deprecating the old one easier.

dereuromark:
[...]

PR #11896, opened 2018-04-02
Add QueryExpression::newCaseExpr() 

raul338:
Refs #11793

CaseExpression was ~almost~ already fluid, but It was not exposed via Query.  
Now sql cases can mimic real sql cases

Rewrote part of a test, It can be used as example in the cookbook

Happy Easter


raul338:
AppVeyor fail seems unrelated

src/Database/Expression/QueryExpression.php
```diff
@@ -376,6 +376,7 @@ public function in($field, $values, $type = null)
      * @param array $types associative array of types to be associated with the values
      * passed in $values
      * @return $this
+     * @deprecated 3.6.0 use \Cake\Database\Query::newCaseExpr()
```
markstory: Should this also emit a deprecation warning? Or are you planning for the method to last until 5.0?

src/Database/Query.php
```diff
@@ -1746,6 +1747,16 @@ public function newExpr($rawExpression = null)
         return $expression;
     }
 
+    /**
+     * Returns a new CaseExpression object to build SQL Case clauses
+     *
+     * @return \Cake\Database\Expression\CaseExpression
+     */
+    public function newCaseExpr()
```
markstory: Why is this method here and not on QueryExpression like the other expression builders?

raul338: Based on my [original comment](https://github.com/cakephp/cakephp/issues/11793#issuecomment-370888809), I try to make that code working.
`$query->newExpr()->newCaseExpr()` would be more verbose I think

src/Database/Expression/QueryExpression.php
raul338: Forgot about the warning, I'll add it in a commit

raul338:
If this is good  to go, should I rewrite other `addCase` test to this format?

markstory:
@raul338 We should keep at least one test case for the existing `addCase()` method. You could add another test case if the current builder doesn't have enough coverage.

tests/TestCase/Database/QueryTest.php
```diff
@@ -4221,11 +4276,15 @@ public function testSqlCaseStatement()
             'Not published',
             'None'
         ];
+        $finalExpression = $query->newExpr();
+        $this->withErrorReporting(E_ALL ^ E_USER_DEPRECATED, function () use ($conditions, $values, $finalExpression) {
+            $finalExpression->addCase($conditions, $values);
+        });
```
raul338: I don't know if this is the best way to test for deprecated behavior. Should I change this?

lorenzo:
The new method added to the query object feels out of place. Why is it the only expression type that gets its own method in the query?

raul338:
Right now with this PR this code should work
```php
$publishedCase = $query
    ->newCaseExpr()
    ->add(
        $query->newExpr()->add(['published' => 'Y']),
        1,
        'integer'
    );
```

Would you like this way?
```php
$publishedCase = $query
    ->newExpr()
    ->newCaseExpr()
    ->add(
        $query->newExpr()->add(['published' => 'Y']),
        1,
        'integer'
    );
```
I feel the extra newExpr() a bit odd, it add no meaning

tests/TestCase/Database/QueryTest.php
markstory: We have a `$this->deprecated (function () { ... });` helper method for toggling error reporting allowing deprecated methods to be called in tests.

markstory:
The API proposed in #11793 added `when()` and `then()` methods to make defining case expressions easier.  This change gets closer to the previous proposal but doesn't deliver what I saw as the big win which was making when/then cases easier to express.

raul338:
I don't think `when()..then()` would work, what would happend if only one of when/then is called multiple times in a row()?

The original issue is that conditions and values are explicitly separated, instead of being inlined (like resulting sql). In this PR instead of using `when/then`, I just use `add(condition, value, [type])`

Again, see [my original comment](https://github.com/cakephp/cakephp/issues/11793#issuecomment-370888809)

markstory:
> what would happend if only one of when/then is called multiple times in a row?

The resulting SQL would be invalid and the developer would know they made a mistake.

raul338:
Ok, Added when/then functions, a ability to test for simpler column cases.
Maybe needs some more testing, and a bit of rewording in some docblocks

Any failing tests would be fixed tomorrow. Its late here.

Let me know what you think. There is also the issue if move `$query->newCaseExpr()` to something else , perhaps `$query->newExpr()->newCase()` ?

src/Database/Expression/CaseExpression.php
```diff
@@ -52,15 +59,28 @@ class CaseExpression implements ExpressionInterface
     /**
      * Constructs the case expression
      *
-     * @param array|\Cake\Database\ExpressionInterface $conditions The conditions to test. Must be a ExpressionInterface
+     * @param array|mixed|\Cake\Database\ExpressionInterface $conditions The conditions to test. Must be a ExpressionInterface
```
dereuromark: Why mixed? That kind of allows everything now.

tests/TestCase/Database/Expression/CaseExpressionTest.php
```diff
@@ -53,6 +54,40 @@ public function testSqlOutput()
...
+            ->then('testResult');
+        $expected = 'CASE testColumn WHEN :param0 THEN :param1 END';
+        $this->assertSame($expected, $caseExpression->sql(new ValueBinder()));
+
+        $caseExpression = new CaseExpression();
+        $caseExpression
+            ->when('test1')
+            ->then('test2')
+            ->then('test4')
+            ->when('test3')
```
markstory: Why are there two then() calls in a row?

tests/TestCase/Database/Expression/CaseExpressionTest.php
```diff
@@ -53,6 +54,40 @@ public function testSqlOutput()
...
+            ->elseValue('testElse')
+            ->caseValue('test0');
+        $expected = 'CASE :param0 WHEN :param1 THEN :param2 WHEN :param3 THEN :param4 ELSE :param5 END';
+        $this->assertSame($expected, $caseExpression->sql(new ValueBinder()));
```
markstory: Should there also be a test for improper usage creating invalid SQL?

src/Database/Expression/CaseExpression.php
raul338: Sorry, fixed and updated

markstory:
> There is also the issue if move $query->newCaseExpr() to something else , perhaps $query->newExpr()->newCase()

I would prefer case expressions to hang off QueryExpression like all the other expression sub types. Right now it is an odd ball being on Query.

tests/TestCase/Database/Expression/CaseExpressionTest.php
raul338: I test 2 when/then, as long as they're balanced it should work

markstory: Ok.

lorenzo: Was this change merged recently?
Can we please avoid calling it multiple times? I thought we would have a different `MultiCaseExpression` for multiple cases with different syntax. It makes not sense the way it is now

raul338: This PR introduces this change. I'm not a fan of the when/then calls either (I prefer them in a single call via `add`)

What would be different in the `MultiCaseExpression` than this CaseExpression ? (never saw that in the original issue). Shouldn't both handle a single case with multiple when/then sub-clauses?

raul338:
How should I continue with this. Should I drop the then/when calls? Or adding more tests only?

lorenzo:
@raul338 let me think a bit more about what a good API would be. In the mean time, I think dropping the `then/when` calls is the way to go

raul338:
In the current state the API would be like this

```php
<?php
// Deprecated
$query
    ->newExpr()
    ->addCase(
        [
            // Conditions
            $query
                ->newExpr()
                ->add(['published' => 'Y']),
            $query
                ->newExpr()
                ->add(['published' => 'N'])
        ],
        [
            // Values
            'Published',
            'Not published',
            'None' // Else value
        ]
    );

// Simple Case
$query
    ->newExpr()
    ->addCase($query->identifier('published'))
        ->add('Y', 'Published')
        ->add('N', 'Not Published')
        ->elseValue('None');

$query
    ->newExpr()
    ->addCase()
        ->caseValue('published')
        ->add('Y', 'Published')
        ->add('N', 'Not Published')
        ->elseValue('None');

// Searched Case
$query
    ->newExpr()
    ->addCase()
        ->add(
            $query->newExpr()->eq('published', 'Y'),
            'Published'
        )
        ->add(
            $query->newExpr()->eq('published', 'N'),
            'Not Published'
        )
        ->elseValue('None');
```

If this is OK, I think just some tests would be needed to merge this PR

src/Database/Expression/CaseExpression.php
```diff
@@ -246,4 +281,26 @@ public function traverse(callable $visitor)
...
+    /**
+     * Helper build a fluent case builder
+     * If Value is not an ExpressionInterface, it will be treated as a value
+     *
+     * @param string $partName property to append the part
+     * @param string|\Cake\Database\ExpressionInterface $value Boolean Condition or Case Value
+     * @param string $type value type
+     * @return $this
+     */
+    protected function _addPart($partName, $value, $type = null)
```
dereuromark: Is this method still used?

raul338: Sorry, not anymore, I'll work in this PR maybe on weekend.

PR #11899, opened 2018-04-02
Fix toDatabase() to work with non-DateTimeImmutable properly

chinpei215:
I think it would be better to not change the given DateTime objects.

### DIFF
--- a/tests/TestCase/Database/Type/DateTimeTypeTest.php
+++ b/tests/TestCase/Database/Type/DateTimeTypeTest.php
@@ -146,9 +146,11 @@ class DateTimeTypeTest extends TestCase
         $result = $this->type->toDatabase($date, $this->driver);
         $this->assertEquals('2013-08-12 15:16:17', $result);
 
+        $tz = $date->getTimezone();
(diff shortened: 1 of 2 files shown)


tests/TestCase/Database/Type/DateTimeTypeTest.php
```diff
@@ -146,9 +146,11 @@ public function testToDatabase()
         $result = $this->type->toDatabase($date, $this->driver);
         $this->assertEquals('2013-08-12 15:16:17', $result);
 
+        $tz = $date->getTimezone();
         $this->type->setTimezone('Asia/Kolkata'); // UTC+5:30
         $result = $this->type->toDatabase($date, $this->driver);
         $this->assertEquals('2013-08-12 20:46:17', $result);
+        $this->assertEquals($tz, $date->getTimezone());
```
chinpei215: I used assertEquals() here, as DateTime::getTimezone() seems to create different TimeZone instances each time it is called.

ADmad (approved):

PR #11903, opened 2018-04-03
Adding whereNotNull() and whereNull() to the Query object

burzum:
Convenience methods for NOT NULL and IS NULL expressions.

Honestly I needed to lookup the expression way of doing this almost every time I needed it because I couldn't remember it. 😆 The convenience methods are pretty self explaining and easy to discover if you're using an IDE.

```php
$this->find()
    ->whereNotNull(['parent_id'])
    ->all();
```

```php
$this->find()
    ->whereNull(['parent_id'])
    ->all();
```

This is also pretty much in line with the recent additions of `inList()`.

### DIFF
--- a/src/Database/Query.php
+++ b/src/Database/Query.php
@@ -882,7 +882,7 @@ class Query implements ExpressionInterface, IteratorAggregate
     }
 
     /**
-     * Convenience method that adds a NOT NULL condition to the query
+     * Adds a NOT NULL or IS NULL expression to the query
      *
      * @param array $fields A list of fields that should be not null
      * @param bool $isNull Toggles between NOT NULL and NULL checks
@@ -894,9 +894,9 @@ class Query implements ExpressionInterface, IteratorAggregate
             $this->where(function ($exp) use ($condition, $isNull) {
                 if ($isNull) {
                     return $exp->isNull($condition);
-                } else {
-                    return $exp->isNotNull($condition);
                 }
+
+                return $exp->isNotNull($condition);
             });
         }
 
--- a/src/Database/Query.php
+++ b/src/Database/Query.php
@@ -882,47 +882,45 @@ class Query implements ExpressionInterface, IteratorAggregate
     }
 
     /**
-     * Adds a NOT NULL or IS NULL expression to the query
-     *
-     * @param array $fields A list of fields that should be not null
-     * @param bool $isNull Toggles between NOT NULL and NULL checks
-     * @return $this
-     */
-    protected function whereNullOrNotNull($fields, $isNull = true)
-    {
-        foreach ($fields as $condition) {
-            $this->where(function ($exp) use ($condition, $isNull) {
-                if ($isNull) {
-                    return $exp->isNull($condition);
-                }
-
-                return $exp->isNotNull($condition);
-            });
-        }
-
-        return $this;
-    }
-
-    /**
      * Convenience method that adds a NOT NULL condition to the query
      *
-     * @param array $fields A list of fields that should be not null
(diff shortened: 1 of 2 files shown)


src/Database/Query.php
```diff
@@ -881,6 +881,50 @@ public function where($conditions = null, $types = [], $overwrite = false)
...
+            });
+        }
+
+        return $this;
+    }
+
+    /**
+     * Convenience method that adds a NOT NULL condition to the query
+     *
+     * @param array $fields A list of fields that should be not null
```
markstory: Couldn't the fields also be query expressions?

tests/TestCase/Database/QueryTest.php
```diff
@@ -742,6 +748,44 @@ public function testSelectWhereTypes()
...
+     */
+    public function testSelectWhereNull()
+    {
+        $this->loadFixtures('MenuLinkTrees');
+
+        $query = new Query($this->connection);
+        $result = $query
+            ->select(['id', 'parent_id'])
+            ->from('menu_link_trees')
+            ->whereNull(['parent_id'])
```
markstory: Should we have tests for expression objects in the field list?

src/Database/Query.php
```diff
@@ -881,6 +881,50 @@ public function where($conditions = null, $types = [], $overwrite = false)
...
+     * @return $this
+     */
+    public function whereNotNull($fields)
```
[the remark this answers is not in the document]
burzum: If we can start doing that... I would love to add them everywhere.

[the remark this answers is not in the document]
dereuromark: For new methods we certainly can (easier than for existing) where it makes sense.

src/Database/Query.php
```diff
@@ -881,6 +881,50 @@ public function where($conditions = null, $types = [], $overwrite = false)
...
     }
 
+    /**
+     * Adds a NOT NULL or IS NULL expression to the query
+     *
+     * @param array $fields A list of fields that should be not null
+     * @param bool $isNull Toggles between NOT NULL and NULL checks
+     * @return $this
+     */
+    protected function whereNullOrNotNull($fields, $isNull = true)
```
lorenzo: This method is confusing, I'd rather not have it

dereuromark:
Whats the difference of this doing internal $exp->..() over e.g. calling internally `where($fieldname . ' IS' => $value)` or `where($fieldname . ' IS NOT' => $value)`?

src/Database/Query.php
lorenzo: I think I would prefer having more code, yes.

lorenzo:
> Whats the difference of this doing internal $exp->..() over e.g. calling internally where($fieldname . ' IS' => $value) or where($fieldname . ' IS NOT' => $value)?
> Performance I would guess?

You cannot concatenate expression objects to a string, that's the reason

src/Database/Query.php
markstory: Would there be much duplicate code though? Couldn't `whereNull()` become:

```php
function whereNull($fields)
{
  $exp = $this->newExpr();
  // Could skip this if we want to require an array.
  if (is_array($fields) {
    $fields = [$fields];
  }
  foreach ($fields as $field) {
     $exp->isNull($field);
  }
  return $this->where($exp);
}
```

burzum: Actually I would prefer everywhere in the framework methods that don't accept lots of different types.

burzum:
@lorenzo @markstory is there something left to do on this?

lorenzo:
Probably just a test showing that the new methods accept a string

burzum:
@lorenzo @markstory the test case has been added.

lorenzo:
Thanks @burzum !

PR #11907, opened 2018-04-06
Session close

Erwane:
The php session only allow one process at a time to avoid "session_write" conflict.

The Session::close() allow to write datas and close the session, permit others ajax to run (or multiple loading page)


src/Network/Session.php
```diff
@@ -347,6 +347,26 @@ public function start()
...
+     * @throws \RuntimeException if the session was already started
+     */
+    public function close()
+    {
+        if (!$this->_started) {
+            return true;
+        }
+
+        if (!session_write_close()) {
+            throw new RuntimeException('Could not close the session');
```
markstory: Is there any more information we can give about the failure?

markstory (changes requested):
New features require test cases to be merged. If you need a hand with how to get started with tests or what tests should be added let us know.

markstory:
Moved to 3.next as that is the branch new features are merged into.

src/Network/Session.php
Erwane: session_write_close only return true or false, so we can't "improve" the failure message.

Erwane:
i've tried to do tests but in CLI, there is no session file, he can't close it OR i don't know how to test it :/

session_write_close() always return false in CLI

src/Network/Session.php
markstory: Ok.

markstory:
Merged into 3.next in 40ff5d676702ae9af3f77b34525bacf17943d472

garas:
http://php.net/manual/en/function.session-write-close.php

>  7.2.0  The return type of this function is bool now. Formerly, it has been void. 

On PHP 7.1 and earlier `$this->request->getSession()->close()` always throws exception

Issue #11916, opened 2018-04-10
Column default value is "NULL::character varying" insted of "null" for Postgres


progLamer:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.5 
* Platform and Target: Postgres 9.4

### What you did
1. Create table 
```sql
CREATE TABLE tests
(
  name character varying(255) NOT NULL DEFAULT NULL
);
```
2. Decribe table schema
```php
        /** @var TableSchema $schema */
        $schema = ConnectionManager::get('default')
            ->getSchemaCollection()
            ->describe('tests');

       var_dump($schema->defaultValues());
```

### What happened
```php
array(1) {
  'name' =>
  string(23) "NULL::character varying"
}
```

### What you expected to happen
```php
array(1) {
  'name' => 
  NULL
}
```

dereuromark:
PR open

PR #11919, opened 2018-04-10
Fix Column default value is "NULL::character varying" insted of "null" for Postgres

progLamer:
Fix issue #11916


markstory:
Thanks!

PR #11923, opened 2018-04-13
Implemented multiple lines for all stages in parsing a po file

Wouter0100:
In #11865 I have introduced a bug, lines were incorrectly parsed when there was a context and multiple translated lines. I've first tested this bug by tests/test_app/TestApp/Locale/rule_1_po/default.po line 21 and 22, so it won't happen again.

To fix this bug, I've made the multi line support more dynamic - and added support (and tests) for multi lining all supported elements of the po file (so also `msgid_plural` and array `msgstr`).


tests/test_app/TestApp/Locale/rule_1_po/default.po
```diff
@@ -18,7 +18,8 @@ msgstr "Plural Rule 1 (translated)"
 msgctxt ""
 "This is the context"
 msgid "%d = 1"
-msgstr "First Context trasnlation"
+msgstr ""
+"First Context trasnlation"
```
markstory: Typo in 'translation'

markstory (commented):
Looks like there is one test failing due to the addition of the new messages.

tests/test_app/TestApp/Locale/rule_1_po/default.po
Wouter0100: My bad, I just copied the typo :-). Both are fixed in latest commit.

Wouter0100:
Ah, I failed to commit the new message count. Should be fixed and tests are passing.

Issue #11921, opened 2018-04-12
Removing magic file handling from responses


ndm2:
This is a (multiple allowed):

* [ ] bug
* [ ] enhancement
* [x] feature-discussion (RFC)

* CakePHP Version: 4.0.0

I've seen this over and over again, people try to send files and accidentally pass non-absolute paths (filenames only, relative paths, ...) to `Response::file()/withFile()`, thus having CakePHP do it's "magic", trying to look up the file in `APP`, which by default is the `src` folder.

Even if auto-resolving non-absolute paths would be deemed useful, who would actually want to serve files from the `src` folder? I would like to suggest to remove this behavior in 4.0, completely that is, ie. make `withFile()` require an absolute path, and fail hard for anything else.

Even though the current behavior is documented in the API docs, it's IMHO an unnecessary boobytrap, and people seem to step in to it all the time.

saeideng:
https://github.com/cakephp/cakephp/blob/40ec1f4569c979f0b5d000d42560b1e8374c9fe5/src/Http/Response.php#L2397-L2414

markstory:
We should add a deprecation warning for this scenario now.

PR #11926, opened 2018-04-14
Add deprecation warning for non-absolute file paths.

markstory:
Files being used with Response::withFile() should be absolute. Prefixing with the application path is often seen as unwanted magic.

Refs #11921

### DIFF
--- a/src/Http/Response.php
+++ b/src/Http/Response.php
@@ -2430,7 +2430,7 @@ class Response implements ResponseInterface
      * - download: If `true` sets download header and forces file to be downloaded rather than displayed in browser
      *
      * @param string $path Path to file. If the path is not an absolute path that resolves
-     *   to a file, `APP` will be prepended to the path.
+     *   to a file, `APP` will be prepended to the path (this behavior is deprecated).
      * @param array $options Options See above.
      * @return void
      * @throws \Cake\Http\Exception\NotFoundException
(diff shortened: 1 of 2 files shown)


src/Http/Response.php
```diff
@@ -2587,6 +2587,7 @@ protected function validateFile($path)
             throw new NotFoundException(__d('cake', 'The requested file contains `..` and will not be read.'));
         }
         if (!is_file($path)) {
+            deprecationWarning('Using non-absolute paths with Response::file() and withFile() is deprecated.');
```
ndm2: I'm a bit late, sorry... but wouldn't a `Folder::isAbsolute()` check be better suited? `is_file()` allows non absolute paths to slip through, given that the path would be looked up based on the current working directory. Also the warning would be triggered for all non existent files, irrespective of the type of the path.

markstory: In 4.0 I was planning on deleting this block, so relative paths would continue working. The error message can be adjusted to reflect that the automatic scoping to APP is being removed.

Was your intention to also see relative file path support also removed?

ndm2: It was, yes, I should have elaborated a bit more on it.

Accepting relative paths assumes that the current working directory is a "safe" place, which IMHO isn't necessarily the case. Normally the CWD should be `webroot` folder, but that might vary.

What I personally would like to see is an exception being thrown for non absolute paths, so that people need to be explicit, and think about what they are trying to serve. Surely people would still be able to build "dangerous" paths that point to for example the `src` or `config` folder, but a mechanism that disallows certain paths would probably be a bit too much.

markstory: Ok. I'll put up another pull request for that.

PR #11930, opened 2018-04-15
Don't set session cookie name to CAKEPHP by default.

### LINKED ISSUE #11584: RFC: Don't set session cookie name to "CAKEPHP"
Setting the session cookie name to "CAKEPHP" by default and advertising the fact that an app uses CakePHP provides no benefit. It can in fact be used to try out attacks against known security vulnerabilities if the app doesn't use an up to date release.

So I propose just using the default session name set in php.in instead.

ADmad:
Closes #11584.

### DIFF
--- a/src/Http/Session.php
+++ b/src/Http/Session.php
@@ -104,7 +104,9 @@ class Session
             $sessionConfig['ini']['session.cookie_secure'] = 1;
         }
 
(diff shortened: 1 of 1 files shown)


saeideng:
https://github.com/cakephp/app/blob/cf94d1181a4e1dac18dc8ae4a758115f999c0f10/config/app.default.php#L349

Issue #11760, opened 2018-02-26
RFC: Use .php extension by default for templates.


ADmad:
Using `.ctp` extension for templates doesn't provide much value, just extra hassle to configure editors to recognize them as PHP files. Since all templates are nested under `Template` folder (and generally use underscored file names) they are easily recognizable as templates and don't need a special extension for distinguishing them from class files.

So I propose switching to `.php` extension for templates in 4.0.

inoas:
I'd say we should keep .ctp but the load order would be .twig (in case we support that) > .php > .ctp and .ctp being dropped in 5.x.

ADmad:
Personally I would avoid implicit load order for .php and .ctp files. It would be unnecessary hassle figuring out the right file to load considering things like fallback to default templates of themes and ability to override plugin templates in app. Even when upgrading from 3.x to 4.0, renaming .ctp to .php would be simple enough or one can set `AppView::$_ext` as required.

Twig templates are a separate case altogether as they need a separate view class for rendering. Even if twig support is provided in core itself, it's use should be explicit opt-in or opt-in without any fallback feature.

inoas:
Hm... I am okay with opting in things but I would not make things harder out of the box.

It can be very feasible to have .twig > .php loading - for instance in plugin-world you supply a .php file but end users can overwrite via say `app/src/Template/Plugin/Foobar/batz/quux.twig`

Obviously the path priority would be stronger than the extension/parser priority.

And again, dropping support out of the box for .ctp will be a hassle for seamless upgrading. Why?
If you want to make a clean cut, maybe a bake task that renames all .ctp files to .php would be useful?

chinpei215:
As I said in the dev channel, this change could be a security hole accidentally. If you call `class_exists('App\Template\Users\index')`, composer will include `index.php`. If we want to do that, we would also have to move the `Template` directory to the root directory like `config` or `bin`.

ADmad:
Moving template files out of autoload path `src` does make sense.

dereuromark:
```
config/
src/ (autoload)
templates/
tests/ (autoload-dev and autoload for fixture)
```
sounds reasonable to me, too.

>  maybe a bake task that renames all .ctp files to .php would be useful?

No, @ionas , that has nothing to do with bake plugin. This is an upgrade tool task. But I agree that such a task is vital for this to be adopted and is a hard requirement for me in such a case.

saeideng:
Moving template files out of autoload path src does make sense.
[...]

PR #11931, opened 2018-04-15
4.x templates

ADmad:
Refs #11760.


dereuromark:
This is definitely sth for the migration guide then.

src/View/View.php
```diff
@@ -168,7 +168,7 @@ class View implements EventDispatcherInterface
      *
      * @var string
      */
-    protected $_ext = '.ctp';
+    protected $_ext = '.php';
```
markstory: Would be good if we could provide a tool to make this kind of mechanical change easier to do.

ADmad: Could just document this 3 liner needed (probably someone can improve it) to move the files and rename :)
```sh
git mv src/Template templates ;
find plugins -name 'Template' -exec sh -c 'git mv "$0" "$0"/../../templates' {} \; 2> /dev/null ;
find plugins templates -name '*.ctp' -exec sh -c 'git mv "$0" "${0%.ctp}.php"' {} \;

```

ravage84 (commented):
I do like the folder location change, but don't like the file extension change. :man_shrugging:

saeideng:
what you think about split this PR ?
1) for moving  templates from `src/Template` to `templates` ( i think its necessary at all ) 
2) if we want to use `.php` , changing  extension of templates files to ".php"

inoas:
What about a double extension like ctp.php - the reason would be to make it easy for editors to inject macros, special highlightning, etc?

saeideng:
also we can not move `templates` without moving `src/locale`

ADmad:
I added new `UpgradeCommand` to handle moving and renaming of template files.

ADmad:
We should over time update all tests which interact with filesystem to use `vfsStream` as I have done for `UpgradeCommandTest` :nerd_face:.  

It avoids polluting the local file system and having to restore directory/file structure to prevent problems between test cases.

src/Core/Plugin.php
```diff
@@ -80,6 +80,20 @@ public static function configPath(string $name): string
...
     }
 
+    /**
+     * Returns the filesystem path for plugin's folder containing template files.
+     *
+     * @param string $name name of the plugin in CamelCase format.
+     * @return string Path to the plugin folder container config files.
+     * @throws \Cake\Core\Exception\MissingPluginException If plugin has not been loaded.
+     */
+    public static function templatePath(string $name): string
```
saeideng: why we need static method in `Plugin` class ?
this is useless IMO

ADmad: Static methods are required in `App::path()` and `ExtractTask` to get plugin template path. Without this new method you would still have to use static method `Plugin::path()` and string concatenation.

saeideng: you can use 
```php
Plugin::getCollection()->get('PluginName')->getTemplatePath()
```
instead of `templatePath()` we can have a new method `get()` and then 
```php
Plugin::get('PluginName')->getTemplatePath()
```

composer.json
```diff
@@ -92,8 +93,8 @@
             "@cs-check",
             "@test"
         ],
-        "cs-check": "phpcs --colors -p ./src ./tests",
-        "cs-fix": "phpcbf --colors ./src ./tests",
+        "cs-check": "phpcs --colors -p --ignore=*/templates/* ./src ./tests",
```
dereuromark: On case ins. file systems this can also effect any path in src with Templates/ dir.
Why do we need it if we already pull templates/ into root?

PS: the `./src ./tests` part should better read `src/ tests/` to be consistent with all the other path declarations we usually do.

src/Command/UpgradeCommand.php
```diff
@@ -0,0 +1,195 @@
...
+     *
+     * @param string $source Source path.
+     * @param string $dest Destination path.
+     * @return void
+     */
+    protected function rename($source, $dest)
+    {
+        $this->io->out("Move $source to $dest");
+
+        if (!$this->args->getOption('dry-run')) {
```
dereuromark: We could bail early here to avoid nesting.

src/Command/UpgradeCommand.php
```diff
@@ -0,0 +1,195 @@
...
+use Cake\Console\ConsoleOptionParser;
+use Cake\Core\Configure;
+use RecursiveDirectoryIterator;
+use RecursiveIteratorIterator;
+use RecursiveRegexIterator;
+use RegexIterator;
+
+/**
+ * Upgrade 3.x application to 4.0.
+ */
```
markstory: We had talked about updating rectorphp for upgrade code modification tools. Including an upgrade tool in cakephp could make future updates hard if we try to stick to server.

composer.json
ADmad: Not quite sure why I changed this. Will revert.

src/Command/UpgradeCommand.php
ADmad: We can drop this `UpgradeCommand` when rectorphp is support is ready :)

markstory: Ok. :+1:

src/Core/BasePlugin.php
```diff
@@ -77,6 +77,13 @@ class BasePlugin implements PluginInterface
      */
     protected $configPath;
 
+    /**
+     * The templates path for this plugin.
+     *
+     * @var string
+     */
+    protected $templatePath;
```
markstory: Is this also supported as a constructor option?

ADmad: I didn't notice other paths could be passed as constructor option. Will update to allow setting template path too.

ADmad: Done.

src/Core/Plugin.php
ADmad: I added `Plugin::templatePath()` for consistency. If / when we decide to get rid of other methods like `Plugin::configPath()`, `Plugin::classPath()` when can remove `Plugin::templatePath()` too.

src/Core/BasePlugin.php
```diff
@@ -175,6 +182,19 @@ public function getClassPath(): string
...
+     * {@inheritDoc}
+     */
+    public function getTemplatePath(): string
+    {
+        if ($this->templatePath) {
+            return $this->templatePath;
+        }
+        $path = $this->getPath();
+
+        return $path . 'templates' . DIRECTORY_SEPARATOR;
```
markstory: Should this method set `$this->templatePath` so we don't repeat work the 2nd+ time this method is called?

ADmad: Done

ADmad:
Anything more required to merge this?

markstory:
Have the docs been updated?

ADmad:
Nope, I can get on it once the code changes are approved.

lorenzo (approved):

PR #11933, opened 2018-04-15
4.x remove old classes for phpunit < 6

saeideng:


jeremyharris:
I'm so happy we can get rid of this jank 😄

PR #11940, opened 2018-04-16
Fix path in FileTest

DIDoS:
CORE_PATH already ends with a DS (define('CORE_PATH', CAKE_CORE_INCLUDE_PATH . DS))


saeideng:
fixed here
https://github.com/cakephp/cakephp/commit/2d168c51a703e84aad8ac043aeb7454c07b2f90b

PR #11946, opened 2018-04-16
fix Controller 3.6 deprecated message from class attribute to method

Schlaefer:
Just makes the deprecated message a little bit more explicit.


markstory:
Thanks!

PR #11948, opened 2018-04-16
Fixed request and response deprecations

DIDoS:
Fixed some deprecations. Especially response and request getter and setter instead of direct access.

PR #11952, opened 2018-04-16
Fixed request and response deprecations

DIDoS:
Fixed some deprecations. Especially response and request getter and setter instead of direct access.

#11948


src/TestSuite/IntegrationTestCase.php
```diff
@@ -476,7 +483,7 @@ public function options($url)
      * @param string $method The HTTP method
      * @param array|null $data The request data.
      * @return void
-     * @throws \Exception
+     * @throws PhpunitException
```
dereuromark: Those should be FQCN if possible.

ADmad:
@DIDoS Ping.

src/Database/Connection.php
```diff
@@ -503,6 +511,7 @@ public function begin()
      * Commits current transaction.
      *
      * @return bool true on success, false otherwise
+     * @throws \Exception
```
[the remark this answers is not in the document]
DIDoS: It is useful so that you know or your IDE knows that you didn't handle an Exception/Exception type on the action level. Many applications just throw a lot of exceptions without knowing it... They should at least handle them and throw them again if they want this behavior.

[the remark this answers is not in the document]
dereuromark: Those here are all invalid IMO. The throws should only be added as primary level (in phpstorm level 0), that means if this method itself throws the exception, otherwise not.
I have written a sniff for this that can be used to automate this. Everything else is not sniffable and thus not enforceable and only makes things worse/confusing.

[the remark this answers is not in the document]
DIDoS: If one looks in the definition/doc of a method it would be useful to know which Exception is thrown besides IDE that depending on the IDE requires PHPDoc or computes it. Especially on interfaces I would not use the level 0 approach.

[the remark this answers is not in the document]
dereuromark: Everything that is not sniffable will just get outdated and wrong down the method invocation chain.
So no, please lets stick to my suggestion.
Also: Only catchable exceptions make sense to note usually. 500 ones that are not supposed to be caught are useless either way here.

src/Cache/Cache.php
```diff
@@ -155,6 +155,7 @@ public static function registry(ObjectRegistry $registry = null)
      * @param string $name Name of the config array that needs an engine instance built
      * @return void
      * @throws \InvalidArgumentException When a cache engine cannot be created.
+     * @throws \Exception
```
dereuromark: I only see InvalidArgumentException here, the RuntimeException is caught and handled.

src/Console/Shell.php
```diff
@@ -603,6 +603,7 @@ public function getOptionParser()
      *
      * @param string $name The task to get.
      * @return \Cake\Console\Shell Object of Task
+     * @throws \Exception
```
dereuromark: Nothing here.

src/Console/ConsoleIo.php
```diff
@@ -512,6 +512,7 @@ public function setLoggers($enable)
      * @param string $name The name of the helper to render
      * @param array $settings Configuration data for the helper.
      * @return \Cake\Console\Helper The created helper instance.
+     * @throws \Exception
```
dereuromark: Nothing here.

src/Console/Shell.php
DIDoS: load->_checkDuplicate RuntimeException
load->_throwMissingClassError Exception

dereuromark: Not primary level, thus not really useful here, though.

src/Database/Connection.php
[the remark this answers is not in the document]
DIDoS: Not sure if it will be outdated - but I guess this could be a problem. I see the phpdoc in APIs as a contract that you base your code on - especially if it's not your own code this is important.
The best solution with big impact would be to split checked and unchecked Exceptions (e.g. inheritance) to differentiate between Exceptions one should catch. Still I would document this all way long to the method the user of the api calls.

src/Controller/Component/CookieComponent.php
```diff
@@ -246,7 +246,7 @@ protected function _load($key)
         if (isset($this->_loaded[$first])) {
             return;
         }
-        $cookie = $this->request->getCookie($first);
```
ADmad: I don't think removing calls to `$this->request` need to be removed. Requiring access to request/response instance in components is pretty common and those properties should be retained in components, albeit with visibility protected.

markstory: One issue with components retaining references to the request is that it makes it much easier to make errors with immutability. For example any `withX()` method called on a request held by a component will not update other components or the controller. If we centralize the request on to the controller people will be less likely to make this kind of error.

ADmad: Okay makes sense.

Issue #11944, opened 2018-04-16
URL mismatch in POST data


chtheis:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.6.0
* Platform and Target: Debian Stretch, Apache 2.4.25

### What you did
Upgrade CakePHP to version 3.6

### What happened
My project is in a subfolder 'project', so my URLs are e.g. /project/users/login for controller 'users' and action 'login'.

In webroot/.htaccess I do the rewrite:
```
<IfModule mod_rewrite.c>
    RewriteEngine On
    RewriteBase /project
    RewriteCond %{REQUEST_FILENAME} !-d
    RewriteCond %{REQUEST_FILENAME} !-f
    RewriteRule ^ index.php [L]
</IfModule>
```

I use the SecurityComponent to protect POST data against forgery.
When I submit the form I get an error 
"URL mismatch in POST data (expected '/project/users/login' but found '/users/login')"
So far I figured out that the SecurityHelper constructs the 'url' part of the token including the base folder, i.e. '/project/users/login', but that the SecurityComponent checks against `request->getRequestTarget()`, which returns the URL without the base folder, i.e. '/users/login'.

In 3.5 the URL was checked against `request->here()`, which seems to include the folder as well by default.
`ServerRequest::here` has an optional parameter $base, by default true, but `ServerRequest::getRequestTarget` does not have this parameter. 
Though it is recommended to replace here with `getRequestTarget` this may not always be possible.

### What you expected to happen
No error, of course :)

P.S. Remember, an issue is not the place to ask questions. You can use [Stack Overflow](https://stackoverflow.com/questions/tagged/cakephp)
for that or join the #cakephp channel on irc.freenode.net, where we will be more
than happy to help answer your questions.

Before you open an issue, please check if a similar issue already exists or has been closed before.

jeremyharris:
Refs https://github.com/cakephp/cakephp/pull/11867

I think we might be able to yank some of the work in there. We'll have to probably revert the changes in IntegrationTestCase after this is fixed.

markstory:
Pull request open now.

Issue #11932, opened 2018-04-15
Redirects broken after 3.6.0 for applications in sub-directories


swiffer:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.6.0
* Platform and Target: Apache 2.4 // PHP 7.2.4
  - https://example.com/subfolder
  - /var/www/html/subfolder

### What you did
upgraded to 3.6.0

### What happened
redirects (FriendsOfCake/search [PRG], loginRedirect, FormValidation errors) stopped working.
parts of the url are missing after dispatching Controller.beforeRedirect

![grafik](https://user-images.githubusercontent.com/2990373/38782637-d3c08f6a-40f6-11e8-8c6a-34c61cf8b08e.png)

### What you expected to happen
should work as before

P.S. Remember, an issue is not the place to ask questions. You can use [Stack Overflow](https://stackoverflow.com/questions/tagged/cakephp)
for that or join the #cakephp channel on irc.freenode.net, where we will be more
than happy to help answer your questions.

Before you open an issue, please check if a similar issue already exists or has been closed before.

markstory:
I don't understand the screenshot you've included, is it trying to show the missing URL? What parts are missing that should be there? How could I reproduce the problem you're having?

raul338:
I'm having the same problem with the following stacktrace
```
2018-04-15 22:41:09 Notice: Deprecated (16384): RequestHandlerComponent::beforeRedirect() is deprecated. This functionality will be removed in 4.0.0. Set the `enableBeforeRedirect` option to `false` to disable this warning. - /home/raul/repos/test/vendor/cakephp/cakephp/src/Event/EventManager.php, line: 353 in [/home/raul/repos/test/vendor/cakephp/cakephp/src/Core/functions.php, line 305]
Request URL: /
Referer URL: http://localhost:8765/
Trace:
Cake\Error\BaseErrorHandler::handleError() - CORE/src/Error/BaseErrorHandler.php, line 153
deprecationWarning - CORE/src/Core/functions.php, line 305
Cake\Controller\Component\RequestHandlerComponent::beforeRedirect() - CORE/src/Controller/Component/RequestHandlerComponent.php, line 269
Cake\Event\EventManager::_callListener() - CORE/src/Event/EventManager.php, line 353
Cake\Event\EventManager::dispatch() - CORE/src/Event/EventManager.php, line 330
[...]

PR #11954, opened 2018-04-17
Fix mismatched URL comparisions in SecurityComponent.

markstory:
When SecurityComponent was updated to use new request methods, I forgot to take into account base URL paths. In order to get those applied, we can use Router::url(). This ensures we get the same urls that FormHelper creates. I've also fixed an issue where if the session hadn't already been started secured forms would fail as session_id() would return '' instead of the actual session id.

Refs #11932 
Refs #11944


swiffer:
tested and confirmed to be working here

lorenzo (approved):

markstory:
Thank you for testing these changes @swiffer

PR #11962, opened 2018-04-17
Add ServerRequest::getPath().

ADmad:
`ServerRequest::getRequestTarget()` is not a valid alternative for deprecated
`ServerRequest::$url` as value returned by the former includes "/" prefix and
request query string.

### DIFF
--- a/src/Http/ServerRequest.php
+++ b/src/Http/ServerRequest.php
@@ -2347,21 +2347,18 @@ class ServerRequest implements ArrayAccess, ServerRequestInterface
     /**
      * Get the path of current request.
      *
-     * The returned value does not contain leading slash and is identical to
-     * the deprecated ServerRequest::$url property.
-     *
      * @return string
      * @since 3.6.1
      */
     public function getPath()
     {
-        $path = $this->uri->getPath();
-
-        if ($this->requestTarget !== null) {
-            list($path) = explode('?', $this->requestTarget);
+        if ($this->requestTarget === null) {
+            return $this->uri->getPath();
         }
 
-        return substr($path, 1);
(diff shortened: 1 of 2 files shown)


src/Http/ServerRequest.php
```diff
@@ -2344,6 +2344,23 @@ public function getRequestTarget()
         return $target;
     }
 
+    /**
+     * Get the path of current request.
```
markstory: Should we mention that this is missing the leading and trailing slashes?

markstory (commented):
Looks good. Tests are failing in PHP 5.6

src/Http/ServerRequest.php
inoas: Shouldn't it include the leading slash at least?

ADmad: @inoas This is meant to be a replacement for `Request::$url` which does not include leading slash either.

That said I personally have almost always used that property after adding leading slash, so many in 4.0 we can modify the method to include leading slash.

ADmad:
@markstory Docblock updated. Not sure why the test is failing on 5.6. The issue is probably in the code which generates `$request->uri` instance, not in the method I added. `$request->uri()->getPath()` is supposed to return `/` if you path has been specifically set.

src/Http/ServerRequest.php
markstory: We could have the leading slash now. Adding it later is a subtle change that is likely going to cause problems. Doing it now will make the minor difference part of the upgrade process.

ADmad: @markstory I agree. Done so.

PR #11956, opened 2018-04-17
skip trace + dont show deprecation warnings in html

saeideng:
Refs: https://github.com/cakephp/cakephp/pull/11889

PR #11964, opened 2018-04-18
Don't display deprecation warnings by default.

markstory:
Change the default error handling so that deprecations are not displayed by default. Instead deprecations are only logged and can be displayed through an opt-in setting.

I thinks this addresses the opt-in concern, without adding significantly more complexity. It does add another knob, but in order to achieve an 'opt-in, off by default' behavior we need an additional setting.

This also doesn't block/disrupt the changes being done in cakephp/debug_kit#605

Refs #11956
Refs #11889

### DIFF
--- a/src/Error/BaseErrorHandler.php
+++ b/src/Error/BaseErrorHandler.php
@@ -149,7 +149,14 @@ abstract class BaseErrorHandler
                 'path' => Debugger::trimPath($file)
             ];
         }
-        $this->_displayError($data, $debug);
+
+        $shouldDisplay = (
+            $error !== 'Deprecated' ||
+            ($error === 'Deprecated' && $this->_options['displayDeprecations'])
+        );
+        if ($shouldDisplay) {
+            $this->_displayError($data, $debug);
+        }
         $this->_logError($log, $data);
 
         return true;
--- a/src/Error/ErrorHandler.php
+++ b/src/Error/ErrorHandler.php
@@ -97,6 +97,7 @@ class ErrorHandler extends BaseErrorHandler
             'log' => true,
             'trace' => false,
             'exceptionRenderer' => ExceptionRenderer::class,
+            'displayDeprecations' => false,
         ];
         $this->_options = $options + $defaults;
     }
--- a/tests/TestCase/Error/ErrorHandlerTest.php
+++ b/tests/TestCase/Error/ErrorHandlerTest.php
@@ -202,6 +202,59 @@ class ErrorHandlerTest extends TestCase
     }
 
     /**
+     * test deprecation logging only
+     *
+     * @return void
+     */
+    public function testHandleErrorLogDeprecation()
+    {
+        Configure::write('debug', true);
+
+        $errorHandler = new ErrorHandler();
+        $errorHandler->register();
+        $this->_restoreError = true;
+
+        $this->_logger->expects($this->once())
+            ->method('log')
+            ->with(
+                $this->matchesRegularExpression('(notice|debug)'),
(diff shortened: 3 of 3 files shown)


ADmad:
Having the option to only log deprecation warning is nice but I am not in favor of hiding them on screen by default. Errors/warnings should be loud and in your face by default so that developers easily notice them and take action. We can simply inform about this knob and error level manipulation in the warning messages.

If we keep them hidden by default most inexperienced users won't even notice the deprecations, the type who go onto stackoverflow/slack first instead of carefully reading errors or checking logs when there is a problem.

src/Error/BaseErrorHandler.php
```diff
@@ -149,7 +149,14 @@ public function handleError($code, $description, $file = null, $line = null, $co
                 'path' => Debugger::trimPath($file)
             ];
         }
-        $this->_displayError($data, $debug);
+
+        $shouldDisplay = (
+            $error !== 'Deprecated' ||
+            ($error === 'Deprecated' && $this->_options['displayDeprecations'])
```
saeideng: why `Deprecated` ?
 and  only `E_USER_DEPRECATED`

src/Error/ErrorHandler.php
```diff
@@ -97,6 +97,7 @@ public function __construct($options = [])
             'log' => true,
             'trace' => false,
             'exceptionRenderer' => ExceptionRenderer::class,
+            'displayDeprecations' => false,
```
saeideng: with auto switching deprecation warnings into debug_kit , we can set `True` here

src/Error/BaseErrorHandler.php
markstory: This value is the result of mapping PHP constants to string values.

saeideng: I know but this will  cover both deprecated warnings
E_USER_DEPRECATED + E_DEPRECATED

src/Error/ErrorHandler.php
markstory: I'm trying to make it easier for plugin authors by not failing all their tests by default. Setting this to true by default would not address that.

src/Error/BaseErrorHandler.php
markstory: yes.

markstory:
> They cannot just break stuff everywhere now and force plugins to either provide 3.5 or 3.6 now.
They cannot just force everyone to upgrade now (especially plugin maintainers) or face penalty not to do so.

With this change, people will no longer see the output of deprecation errors in HTML or Console environments unless the `displayDeprecations`option is enabled.

> I am still in favor of simply doing #11889 and having a very clear and simple switch to turn this tooling on where desired, keeping it off for everyone who just doesnt give a dime about it at this point. 

The constant based approach still goes against how PHP error reporting works and is also not how other libraries in PHP or other toolchains work.

> He could chose to toggle on the deprecations and start migration his code base. But he CAN also just ignore and have more work later. His decision, not the framework's.

This change moves all deprecations to logs only unless the app/plugin developer opts into having deprecations displayed. Personally I don't like this, but I'm trying to reach a compromise.

inoas:
I think we should improve communication. Maybe there is a way to add  warning/notice and a hint on how to disable deprecation warnings when doing composer update/upgrade?

I don't think we should have any warnings or notices disabled by default. That is really bad and just hides problems.

It should be as easy as going into one file and changing one variable, but by default it should hit into your face. Otherwise authors will not notice and will not upgrade and we have the same problem when 4.x hits - just later.

PR #11968, opened 2018-04-18
Add deprecationWarning for TableRegistry.

mosaxiv:
Add deprecationWarning for TableRegistry.


dereuromark:
With the current opt-out deprecation issues this is not helpful right now. Only with my suggested opt-in flag this can be OKish.

markstory:
Thanks for doing this, but like `Shell` and `AuthComponent` this was intended to be a 'soft' deprecation that won't be removed until 5.0.

mosaxiv:
I see. Thanks!

PR #11972, opened 2018-04-20
Make CSRF token comparisions time constant.

markstory:
Felix Wiedemann and Nils Rokita contacted us about a possible timing side channel in CSRF token comparisons. This issue can be mitigated by using the constant time checks in Security. Exposing this method as a public method allows us and application developers to use it as needed.

If people are happy with this I can backport to 3.5 and do another release for that branch.


Nilix007:
Looks good to me.

Kind regards,
Felix

PR #11975, opened 2018-04-20
4.x - Database type

ADmad:
- Converted `Cake\Database\Type` to pure factory class instead of current hybrid.
- Added `Cake\Database\BaseType` and made type classes extend it to keep code DRY.
- Added new getter for map `Type::getMap()`. This should probably be backported to 3.x and using `Type::map()` as getter deprecated.

### DIFF
--- a/src/Database/Type.php
+++ b/src/Database/Type.php
@@ -109,15 +109,13 @@ class Type
 
     /**
      * Registers a new type identifier and maps it to a fully namespaced classname,
-     * If called with no arguments it will return current types map array
      * If $className is omitted it will return mapped class for $type
      *
-     * @param string|string[]|null $type If string name of type to map, if array list of arrays to be mapped
+     * @param string|string[] $type If string name of type to map, if array list of arrays to be mapped.
      * @param string|\Cake\Database\TypeInterface|null $className The classname or object instance of it to register.
-     * @return array|string|null If $type is null then array with current map, if $className is null string
-     * configured class name for give $type, null otherwise
+     * @return string|null If $className is null string configured class name for give $type, null otherwise
      */
-    public static function map($type = null, $className = null)
+    public static function map($type, $className = null)
     {
         if ($type === null) {
             return static::$_types;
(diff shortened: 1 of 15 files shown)


src/Database/Type/BoolType.php
```diff
@@ -26,30 +24,8 @@
  *
  * Use to convert bool data between PHP and the database types.
  */
-class BoolType extends Type implements TypeInterface, BatchCastingInterface
+class BoolType extends BaseType implements BatchCastingInterface
```
markstory: Should we also remove the checka that prevent core type classes from using batch casting?

ADmad: The type check in `FieldTypeConverter` has been removed now. I believe that's what you are referring to.

ADmad:
Can we rename `Type` to `TypeFactory` to make it more clear what the class does?

src/Database/Type.php
```diff
@@ -132,28 +99,23 @@ public static function buildAll()
...
-     * @param string|string[]|\Cake\Database\Type[]|null $type If string name of type to map, if array list of arrays to be mapped
-     * @param string|\Cake\Database\Type|null $className The classname or object instance of it to register.
-     * @return array|string|null If $type is null then array with current map, if $className is null string
-     * configured class name for give $type, null otherwise
+     * @param string|string[] $type If string name of type to map, if array list of arrays to be mapped.
+     * @param string|\Cake\Database\TypeInterface|null $className The classname or object instance of it to register.
+     * @return string|null If $className is null string configured class name for give $type, null otherwise
      */
-    public static function map($type = null, $className = null)
+    public static function map($type, $className = null)
```
dereuromark: If it can't be null anymore, we should remove the return part `if ($type === null)`.
Also, we should maybe move the array input part out of this method. This is weird, that this then assigns directly, not using the 2nd argument. An own method mapAll() or something can be used here maybe?

dereuromark: And if this is a true setter, then

    if ($className === null) {
             return isset(static::$_types[$type]) ? static::$_types[$type] : null;
         }

is also obsolete, the $className should then not be nullable anymore, either.

Issue #11974, opened 2018-04-20
RequestHandlerComponent acts differently when tested via PhpUnit (CLI ) as opposed to actual HTTP Client


iliepandia:
This is a (multiple allowed):

* [x] bug
* [x] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.6.1.
* Platform and Target: Running PHPUnit Tests; Php 7.1; PhpUnit  6.5.8 

### What you did
I am testing a new middleware that I wrote. In one test case I throw an exception from the middleware layer. I make the request with the header `"Accept: application/json"` so that the `RequestHandlerComponent` will switch views to render a json error and not a big error page. The test wants to make sure that the errors are also reported back in json format. 

The actual test:
```php
<?php
    public function testUser(){
        $this->_request['headers'] = [ "Accept" => "application/json" ];
        $this->post('/api/data-provider/user/2000', [
           'auth_key' => '1234'
        ]);
        //check response OK
        $this->assertResponseOk();

        //check responded as JSON - this assertion FAILS!!!
        $this->assertHeader('Content-Type', 'application/json; charset=UTF-8');

        //check the JSON makes sense!
        $rawJson = (string)$this->_response->getBody();
        $user = json_decode($rawJson);
        $this->assertNotEmpty($user);
    }
```

Installing the middleware in `routes.php`:
```php
<?php
Router::prefix('api', function($routes){
    /** @var \Cake\Routing\RouteBuilder $routes */
    $routes->registerMiddleware('auth_key', new \App\Middleware\AuthKeyMiddleware() );
    $routes->applyMiddleware( 'auth_key');
    $routes->fallbacks("DashedRoute" );
});
```

### What happened
While running the test case that exception is rendered as a HTML page with the content type: `text/html; charset=UTF-8` instead of the expected value of `application/json; charset=UTF-8`

When running the request via PostMan (si not via a CLI environment) things work as expected!

### What you expected to happen
I would like that when testing in a CLI environment to get the same content type header back as when running the request via an actual HTTP Client

### My own investigation into the issue

Debugging the code:
- the exception is caught be the `ErrorHandlerMiddleware`  - **good**
- the exception rendered is built using `ErrorController` that loads the `RequestHandlerComponent` - **good**
- the new controller attempts to rebuild the request (since it is not passed to it) and debugging through the code it attempts to build it via `ServerRequestFactory::fromGlobals();` - **good**
[...]

PR #11978, opened 2018-04-22
Fix for Issue 11974

iliepandia:
This is a suggested fix for #11974 - passing the original request along to the various handlers that are interested in it.

### Notable changes

The `ExceptionRenderer::__construct()` had to be changed to accept the passed along request. I've put `null` as a default so it would still work if called with the previous signature.

### Cake\TestSuite\LegacyRequestDispatcher vs Cake\TestSuite\MiddlewareDispatcher

They both implemet `execute()` but one expects to an array and the other one a ServerRequest (at least according to PHPDoc).

To fix this is have modified the MiddlewareDispatcher to honor the PHPDoc and changed the `IntegrationTestCase::_sendRequest()` to pass along an array to LegacyRequestDispatcher or a prsRequest to the new MiddlewareDispatcher. 

Cake tests are now passing. 
My own application tests are passing too. 

I don't know how to add a test in Cake to test specifically for the problem described at #11974


src/Error/ExceptionRenderer.php
```diff
@@ -78,16 +79,26 @@ class ExceptionRenderer implements ExceptionRendererInterface
      */
     public $method = '';
 
+    /**
+     * If set, this will be request used to create the controller that will render
+     * the error.
+     *
+     * @var ServerRequestInterface
```
ADmad: Please use FQCN.

src/Error/ExceptionRenderer.php
```diff
@@ -78,16 +79,26 @@ class ExceptionRenderer implements ExceptionRendererInterface
...
+     */
+    public $request = '';
```
ADmad: Should be protected with default null / unset.

src/Error/ExceptionRenderer.php
```diff
@@ -114,9 +125,13 @@ protected function _unwrap($exception)
      */
     protected function _getController()
     {
-        if (!$request = Router::getRequest(true)) {
-            $request = ServerRequestFactory::fromGlobals();
+        $request = $this->request;
```
ADmad: Using `$request = $this->request ?: Router::getRequest(true)` would avoid the nested `if`s.

src/Error/ExceptionRenderer.php
```diff
@@ -78,16 +79,26 @@ class ExceptionRenderer implements ExceptionRendererInterface
      */
     public $method = '';
 
+    /**
+     * If set, this will be request used to create the controller that will render
+     * the error.
+     *
+     * @var \Psr\Http\Message\ServerRequestInterface
```
ADmad: `|null`

src/Error/ExceptionRenderer.php
```diff
@@ -78,16 +79,26 @@ class ExceptionRenderer implements ExceptionRendererInterface
...
+     */
+    protected $request = null;
+
     /**
      * Creates the controller to perform rendering on the error response.
      * If the error is a Cake\Core\Exception\Exception it will be converted to either a 400 or a 500
      * code error depending on the code used to construct the error.
      *
      * @param \Exception $exception Exception.
+     * @param ServerRequestInterface $request The request - if this is set it will be used instead of creating a new one
```
ADmad: Need FQCN here too.

src/Error/ExceptionRenderer.php
```diff
@@ -114,9 +125,10 @@ protected function _unwrap($exception)
      */
     protected function _getController()
     {
-        if (!$request = Router::getRequest(true)) {
+        if (!$request = $this->request ?: Router::getRequest(true)) {
```
ADmad: It's better to avoid variable assignment inside `if`. So please move it above and just do `if ($request === null)` here.

ADmad (changes requested):

ADmad:
Due to method signature changes this cannot go into a patch release. Hence changed target to `3.next`.

src/TestSuite/IntegrationTestCase.php
```diff
@@ -497,8 +506,37 @@ protected function _sendRequest($url, $method, $data = [])
             throw $e;
         } catch (Exception $e) {
             $this->_exception = $e;
-            $this->_handleError($e);
+            $this->_handleError($e, $psrRequest);
```
markstory: I'm not sure we need to change this. This path is only followed when the exception middleware is not used. Because of that getting the current request out of `Router` should be good enough, as it more accurately reflects what would happen if the global default exception handler was invoked.

src/TestSuite/MiddlewareDispatcher.php
```diff
@@ -69,7 +68,7 @@ public function __construct($test, $class = null, $constructorArgs = null)
     /**
      * Run a request and get the response.
      *
-     * @param \Cake\Http\ServerRequest $request The request to execute.
+     * @param \Psr\Http\Message\RequestInterface $request The request to execute.
```
markstory: Should be `ServerRequestInterface`

tests/test_app/TestApp/Error/TestAppsExceptionRenderer.php
```diff
@@ -17,7 +17,7 @@ class TestAppsExceptionRenderer extends ExceptionRenderer
      */
     protected function _getController()
     {
-        if (!$request = Router::getRequest(true)) {
+        if (!$request = $this->request ?: Router::getRequest(true)) {
```
markstory: We generally avoid assignment inside conditionals as it makes code more complex and harder to read as `=` and `==` are easily misread.

src/Error/ExceptionRenderer.php
iliepandia: Agreed. Wrote it like that to keep close to original code. Will rewite.

src/TestSuite/IntegrationTestCase.php
iliepandia: I favor passing the $request along because you can mock it up in tests and for example test that some method was called at some point during the processing of the requesting. Would that sill work if using the `Router`?

src/TestSuite/MiddlewareDispatcher.php
iliepandia: The `$server->run($request);` at line 95 returns `\Psr\Http\Message\RequestInterface`.

If I change to `SeverRequestInterface` I will likely get a return type mismatch warning.

tests/test_app/TestApp/Error/TestAppsExceptionRenderer.php
iliepandia: Yes, fixed. I keep it close to the way it was originally written.

src/TestSuite/MiddlewareDispatcher.php
ADmad: `Psr\Http\Message\RequestInterface` doesn't exist :slightly_smiling_face:. So it's probably a typo at some place.

ADmad: > The $server->run($request); at line 95 returns \Psr\Http\Message\RequestInterface.

`Server::run()` returns a response of type `\Psr\Http\Message\ResponseInterface`, not request.

iliepandia: Argh... I was looking at the return type but the comment was about the 1st param. Sorry. Fixing now.

src/Error/ExceptionRenderer.php
```diff
@@ -114,9 +125,13 @@ protected function _unwrap($exception)
      */
     protected function _getController()
     {
-        if (!$request = Router::getRequest(true)) {
-            $request = ServerRequestFactory::fromGlobals();
+        $request = $this->request;
+        if (!$request) {
+            if (!$request = Router::getRequest(true)) {
```
markstory: This could be `!Router::getRequest(true)`

src/TestSuite/IntegrationTestCase.php
markstory: Outside of mocking this adds a place where the test harness and real behaviour could diverge again.

iliepandia:
@ADmad are there more request changes that I did not fix?

ADmad (approved):

markstory:
Is there anyway we could add additional tests for the new behavior? For example the new constructor arguments don't have any tests, nor does the original scenario these changes were fixing.

PR #11979, opened 2018-04-22
4.x db type

ADmad:
- `Type::map()` is now only used as setter for individual types.
- `Type::setMap()` is used to set complete map.
- `Type::getMap()` is used as getter for individual type or get complete map.

### DIFF
--- a/src/Database/Schema/TableSchema.php
+++ b/src/Database/Schema/TableSchema.php
@@ -440,7 +440,7 @@ class TableSchema implements TableSchemaInterface, SqlGeneratorInterface
             return null;
         }
 
-        if (Type::map($type)) {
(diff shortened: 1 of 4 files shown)


ADmad:
I also think the ability to map type instances should be removed. The map should always be classname string.

markstory:
> I also think the ability to map type instances should be removed. The map should always be classname string.

Wouldn't that remove the option of having a type instance come from a DI container? I'm not sure we want to shut that door to save ourselves a few lines of code.

ADmad:
Wouldn't DI use Type::set() instead of Type::map() ?

markstory:
@ADmad Yes a DI container could use `Type::set()` instead.

PR #11980, opened 2018-04-23
Fix deprecation message for request->here usage.

### LINKED ISSUE #11977: Unexpected behavior for request->getRequestTarget()
This is a (multiple allowed):

* [x] bug
* [ ] enhancement

* CakePHP Version: 3.6.1.
* Platform and Target: Apache/2.4.27, MySQL 5.6.37

### What you did
I upgraded my App from 3.5 to 3.6. and my app is located into a sub directory called myapp so the full URL is http://development.local/myapp/

### What happened
I got a warning error says `Accessing 'here' as a property will be removed in 4.0.0. Use request->getRequestTarget() instead. - /www/myapp/src/Template/Layout/front.ctp, Line 45` so I used `getRequestTarget() ` instead of `here`

I expected that `request->getRequestTarget()` will return the same reult from `request->here` but this was not the case. As I told you above that my app is localed at http://development.local/myapp/
in 3.5 & 3.6, `request->here` returns `/myapp/` but on 3.6 `request->getRequestTarget()` returns `/` which is a different result and it should return `/myapp/`

Thanks

ADmad:
Closes #11977


dereuromark:
Do we also need to fix the deprecation message(s)?

Issue #11934, opened 2018-04-16
Remove deprecated code


markstory:
Tracking issue for deprecation cleanup in 4.x

* [x] Auth
* [x] Cache
* [x] Collection
* [x] Command
* [x] Console
* [x] Controller
* [x] Core
* [x] Database
* [x] Datasource
* [x] Error
* [x] Event
* [x] Filesystem
* [x] Form
* [x] Http
* [x] I18n
* [x] Log
* [x] Mailer
* [x] Network
* [x] ORM
* [x] Routing
* [x] Shell
* [x] Template
* [x] TestSuite
* [x] Utility
* [x] Validation
* [x] View

ADmad:
@josegonzalez `4.x` branch.

markstory:
> It seems only email transport classes use Socket class. Can they be changed to use Http\Client instead? Or if not move Socket under Mailer?

People could use the socket classes for other socket based transports outside of email though.

saeideng:
>@markstory Network doesn't seem to have any deprecated code anymore. 

https://github.com/cakephp/cakephp/blob/4.x/src/Network/Socket.php#L86-L104

ADmad:
@markstory Okay.

@saeideng Aye, I missed that.

jeremyharris:
I have unticked "Testsuite" as 3.7 deprecates IntegrationTestCase and ConsoleIntegrationTestCase and those will need to be removed. I can get around to doing it when 3.7 releases and is merged into 4.x, but I wanted to make a note of it.

markstory:
Going to close this as 4.x is looking pretty clean. While there are a few new deprecations being added into 3.7, we can clean those up incrementally as they merge into 4.x

PR #11981, opened 2018-04-23
[4.x] Removing deprecated methods from I18n

burzum:
Refs #11934

### DIFF
--- a/src/I18n/DateFormatTrait.php
+++ b/src/I18n/DateFormatTrait.php
@@ -29,10 +29,11 @@ trait DateFormatTrait
     /**
      * The default locale to be used for displaying formatted date strings.
      *
+     * Use static::setDefaultLocale() and static::getDefaultLocale() instead.
+     *
      * @var string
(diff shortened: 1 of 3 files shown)


burzum:
@markstory do we need to change the existing prefixed `_` methods and properties as well? I would prefer to keep them.

ADmad:
> do we need to change the existing prefixed _ methods and properties as well?

I don't think we should go about renaming all existing properties/methods with underscore prefixes.

burzum:
@ADmad OK, but we should decide to use one of both conventions and not mix them?

ADmad:
We have stopped adding underscore prefix to new protected properties/methods but the mix will stay for a while due to existing properties. Also there are features like entity getter/setter which rely on having prefixed method names.

markstory:
I'm with @ADmad on the topic of renames. We shouldn't rename tons of methods/properties for cosmetic reasons. We can gradually fix them if we want later on.

Issue #11973, opened 2018-04-20
Cake\TestSuite\EmailAssertTrait


MolbioUnige:
This is a (multiple allowed):

* [x] bug
* [x] enhancement
* [x] feature-discussion (RFC)

* CakePHP Version: 3.6.0.
* Platform and Target: not relevant

### What you did
I tried to use EmailAssertTrait in an IntegrationTestCase to test emails sent by the tested controller.

### What happened
Obviously, it didn't work as I was expecting. Looking at the Traits code, I could see that it doesn't access the controller's response email object. All trait's assertions are calling the `email()` method which access a protected Email object, which is not the controller's email object.

    /* in trait EmailAssertTrait */

    /**
     * @var \Cake\Mailer\Email
     */
    protected $_email;

    public function email($new = false)
    {
        if ($new || !$this->_email) {
            $this->_email = new Email();
            $this->_email->setProfile(['transport' => 'debug'] + $this->_email->getProfile());
        }
        return $this->_email;
    }
### What you expected to happen
Modifying the `email()` method as here below, I was able to test the email generated by the controller I am testing.

    /* in trait EmailAssertTrait */
    public function email($new = false)
    {
        return $this->_controller->email;
    }

 It is also probable that I was not using `EmailAssertTrait` the way it is intended to use, but if it is the case, an example would be very useful.

jeremyharris:
Since this topic came up, I am working on a plugin and part of it is a new trait that lets you assert against *any* emails sent through a transport. The user just injects the new transport and gets assertions against emails across the board (no extra injection of mocks, etc.):

```
// in tests/bootstrap.php
Email::dropTransport('default');
Email::configTransport('default', [
    'transport' => TestEmailTransport::class
]);
```

If this looks of interest to the core, I can work up a PR and strip it out of my plugin.

(Side note: I have never used the `EmailAssertTrait` because I don't quite understand its usage myself - create an `email()` then inject the `getMockForMailer` into the code? Seems like a lot of work, especially for controllers.)

MolbioUnige:
With the slight modification I did, you don't have to bother with `getMockForMailer`, it just gets the email out of the instanciated controller.

I can then test the email as such

    $this->post('/register', $data);
    
    $this->assertEmailSubject('My fine email subject',  'Set email subject');
    $this->assertEmailTo($data['username'], $data['username'], 'Sending to');
    $this->assertEmailTextMessageContains('/activate/' . $result['activation_code'], 'Activation link');


[...]

PR #11983, opened 2018-04-23
Email assertions

jeremyharris:
This is a proposal for #11973. Replaces the `EmailAssertTrait` with an `EmailTrait`, added on `TestCase`, that allows devs to make assertions against any emails sent through the test transport. Testing emails only requires replacing the transport class in `tests/bootstrap.php`:

```php
use Cake\Mailer\Email;
use Cake\TestSuite\TestEmailTransport;

// replace with other transport configs if required
$config = Email::getConfigTransport('default');
$config['className'] = TestEmailTransport::class;
Email::dropTransport('default');
Email::setConfigTransport('default', $config);
```

Now any emails sent through the default transport will be captured for later assertions. Here's a list of the new assertions (`$message` param removed for brevity):

These assertions assert against any email was sent during the test case.

- `assertMailSentTo($address)`: Asserts email was sent to an email address
- `assertMailSentFrom($address)`: Asserts email was sent from an email address
- `assertMailContains($expected)`: Asserts email contains contents
- `assertMailContainsHtml($expected)`: Asserts email contains contents in HTML portion of message
- `assertMailContainsText($expected)`: Asserts email contains contents in text portion of message
- `assertMailSentWith($expected, $param)`: Asserts expectation against other email fields, such as `cc` or `subject`

The above assertions have a sibling method suffixed with `At`, which can be used if multiple emails are sent during a single test. Provide the `$at` argument with the position you expected the email to be sent.

```php
// same as assertMailSentTo, but at the index $at
$this->assertMailSentToAt($at, $address);
```

Finally, some count assertions:

- `assertMailCount($count)`: Asserts `$count` emails were sent
- `assertNoMailSent()`: Asserts no emails were sent

### DIFF
--- a/src/TestSuite/Constraint/Email/MailConstraintBase.php
+++ b/src/TestSuite/Constraint/Email/MailConstraintBase.php
@@ -19,6 +19,8 @@ use PHPUnit\Framework\Constraint\Constraint;
 
 /**
  * Base class for all mail assertion constraints
+ *
+ * @internal
  */
 abstract class MailConstraintBase extends Constraint
 {
--- a/src/TestSuite/Constraint/Email/MailConstraintBase.php
+++ b/src/TestSuite/Constraint/Email/MailConstraintBase.php
@@ -12,7 +12,7 @@
  * @since         3.7.0
  * @license       https://opensource.org/licenses/mit-license.php MIT License
  */
-namespace Cake\TestSuite\Constraint;
+namespace Cake\TestSuite\Constraint\Email;
 
 use Cake\TestSuite\TestEmailTransport;
 use PHPUnit\Framework\Constraint\Constraint;
--- a/src/TestSuite/EmailTrait.php
+++ b/src/TestSuite/EmailTrait.php
@@ -26,12 +26,26 @@ use Cake\TestSuite\Constraint\Email\NoMailSent;
 /**
  * Make assertions on emails sent through the Cake\TestSuite\TestEmailTransport
  *
+ * After adding the trait to your test case, replace all mail transports with the
+ * TestEmailTransport:
+ *
  * **tests/bootstrap.php**
  * ```
  * use Cake\TestSuite\TestEmailTransport;
  *
  * // replaces existing transports with the TestEmailTransport for email assertions
  * TestEmailTransport::replaceAllTransports();
+ * ```
+ *
+ * Then, in your test case's tearDown method, clean up previously sent emails:
+ *
+ * ```
+ * public function tearDown()
+ * {
+ *     // other cleanup
+ *     parent::tearDown();
+ *     TestEmailTransport::clearEmails();
+ * }
  * ```
  */
 trait EmailTrait
--- a/src/TestSuite/EmailTrait.php
+++ b/src/TestSuite/EmailTrait.php
@@ -15,6 +15,8 @@
 namespace Cake\TestSuite;
 
 use Cake\TestSuite\Constraint\Email\MailContains;
+use Cake\TestSuite\Constraint\Email\MailContainsHtml;
+use Cake\TestSuite\Constraint\Email\MailContainsText;
 use Cake\TestSuite\Constraint\Email\MailCount;
 use Cake\TestSuite\Constraint\Email\MailSentFrom;
 use Cake\TestSuite\Constraint\Email\MailSentTo;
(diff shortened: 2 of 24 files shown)


src/TestSuite/EmailTrait.php
```diff
@@ -0,0 +1,166 @@
...
+ * **tests/bootstrap.php**
+ * ```
+ * use Cake\Mailer\Email;
+ * use Cake\TestSuite\TestEmailTransport;
+ *
+ * // replace with other transport configs if required
+ * $config = Email::getConfigTransport('default');
+ * $config['className'] = TestEmailTransport::class;
+ * Email::dropTransport('default');
+ * Email::setConfigTransport('default', $config);
```
markstory: Is there a way we could make this setup easier to use?  Something like

```php
TestEmailTransport::replaceAllTransports();
```

src/TestSuite/EmailTrait.php
```diff
@@ -0,0 +1,166 @@
...
+    }
+
+    /**
+     * Asserts an email contains expected contents
+     *
+     * @param int $contents Contents
+     * @param string $message Message
+     * @return void
+     */
+    public function assertMailContains($contents, $message = null)
```
markstory: Should there be assertions for html/text only messages? What about comparing a body to a known fixture?

src/TestSuite/TestCase.php
```diff
@@ -33,6 +33,7 @@
 abstract class TestCase extends BaseTestCase
 {
 
+    use EmailTrait;
```
markstory: Having this trait be something people opt-into using would make it less magic, and there are lots of test cases where we don't need email assertions.

src/TestSuite/EmailTrait.php
jeremyharris: I like that idea!

src/TestSuite/EmailTrait.php
jeremyharris: Are you thinking different methods for each type of message? Or, add a param here to check 'html', 'text', or 'both' (for either)?

I also like comparing to a fixture. There are honestly a lot of assertions we can add here, including checking attachments.

markstory: Different methods would be easier from a documentation perspective and is more aligned with how PHPUnit works.

src/TestSuite/TestCase.php
jeremyharris: Would you also see `TestEmailTransport::clearEmails();` as something the user should do on their own in `tearDown()`?

markstory:
Looks good!

jeremyharris:
@markstory fyi I haven't removed the trait from TestCase yet; waiting for an answer for my question above :)

src/TestSuite/TestCase.php
markstory: Yeah I think so.

src/TestSuite/Constraint/Email/MailConstraintBase.php
```diff
@@ -0,0 +1,59 @@
...
+ * @since         3.7.0
+ * @license       https://opensource.org/licenses/mit-license.php MIT License
+ */
+namespace Cake\TestSuite\Constraint\Email;
+
+use Cake\TestSuite\TestEmailTransport;
+use PHPUnit\Framework\Constraint\Constraint;
+
+/**
+ * Base class for all mail assertion constraints
```
markstory: Should we mark these new constraint classes as `@internal` so they don't become part of our public API?

src/TestSuite/EmailTrait.php
```diff
@@ -0,0 +1,228 @@
...
+ * Then, in your test case's tearDown method, clean up previously sent emails:
+ *
+ * ```
+ * public function tearDown()
+ * {
+ *     // other cleanup
+ *     parent::tearDown();
+ *     TestEmailTransport::clearEmails();
+ * }
+ * ```
```
markstory: Nice work on the docs :+1:

src/TestSuite/Constraint/Email/MailConstraintBase.php
jeremyharris: I was going to bring that up that question as I'm going through the test suite and doing some work on it -- part of which is converting the assert methods in `(Console)IntegrationTestCase` to proper constraints.

EventFired and EventFiredWith are currently public; do we have to wait until 4.0 to mark those as `@internal`? The odds of them being used directly is likely slim.

jeremyharris:
Should I deprecate `EmailAssertTrait` in favor of this one? I think deprecating it will be painless since there is no documentation on how to use it that I can find. I stumbled upon it in the API one day, that's the only way I knew about it.

I'll also write up docs when this is merged.

src/TestSuite/Constraint/Email/MailConstraintBase.php
markstory: > EventFired and EventFiredWith are currently public; do we have to wait until 4.0 to mark those as @internal

No, we can do it now. We forgot earlier.

markstory:
> Should I deprecate EmailAssertTrait in favor of this one? 

Yes.

PR #11984, opened 2018-04-23
[4.x] Validation deprecations

burzum:
Refs #11934

### DIFF
--- a/src/Form/Form.php
+++ b/src/Form/Form.php
@@ -14,13 +14,11 @@
  */
 namespace Cake\Form;
 
-use Cake\Event\Event;
 use Cake\Event\EventDispatcherInterface;
 use Cake\Event\EventDispatcherTrait;
 use Cake\Event\EventListenerInterface;
 use Cake\Event\EventManager;
 use Cake\Form\Schema;
-use Cake\Validation\Validator;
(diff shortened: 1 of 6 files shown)


src/Form/Form.php
```diff
@@ -211,7 +166,7 @@ protected function _buildValidator(Validator $validator)
      */
     public function buildValidator(Event $event, Validator $validator, $name)
     {
-        $this->_buildValidator($validator);
+        $this->setValidator($name, $validator);
```
ADmad: `setValidator()` is already called by `getValidator()`. The call to `_buildValidator()` was only for BC, so now this method can just be empty.

ADmad: Actually this method as well as `implementedEvents()` can be simply removed. For normal use case one you just use `validationDefault()` method to configure validations as done in tables.

src/Form/Form.php
```diff
@@ -224,7 +179,7 @@ public function validate(array $data)
     {
         $validator = $this->getValidator();
         if (!$validator->count()) {
```
ADmad: The whole `if` block needs to be removed.

ADmad (approved):

burzum:
@ADmad @markstory can I merge it?

PR #11988, opened 2018-04-24
[4.x] Mailer Deprecations

iliepandia:
Removing deprecated code and files related to `Mailer`

Refs #11934 

### Notes

Although the tests were passing there were many lines like

```php
<?php

$this->Email->setTemplate('default', 'default');
```

that I have replaced with 
```php
<?php

        $this->Email->setLayout('default');
        $this->Email->setTemplate('default');
```

Maybe setting the layout to `default` explicitly was not needed, but I did it to keep the code close to the original.


iliepandia:
I don't think there is something I can do about the failing AppVeyor build.

saeideng:
>I don't think there is something I can do about the failing AppVeyor build.

no problem , and unrelated to this PR

src/Mailer/Email.php
```diff
@@ -411,25 +411,6 @@ public function getFrom()
...
-     * @return array|$this
-     * @throws \InvalidArgumentException
-     */
-    public function from($email = null, $name = null)
-    {
-        if ($email === null) {
-            return $this->getFrom();
-        }
-
-        return $this->setFrom($email, $name);
```
saeideng: @markstory we did not  have deprecation warning for `for()` method in 3.6

ADmad: @saeideng If we don't want add new deprecation message in patches release of 3.6 then can just add it for 3.7.

PR #11989, opened 2018-04-24
Run phpstan at level 5.

ADmad:
This should help us ensure we have proper typehints and such from the get go.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,7 +34,10 @@ matrix:
   include:
     - php: 7.2
       env: PHPCS=1 DEFAULT=0
+    - php: 7.1
+      env: PHPSTAN=1 DEFAULT=0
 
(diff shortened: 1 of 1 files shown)


ADmad:
Hmm.. Don't know why the phpstan build doesn't show up here https://travis-ci.org/cakephp/cakephp/builds/370464759.
Pretty sure I have setup the yml properly as per docs https://docs.travis-ci.com/user/customizing-the-build#Rows-that-are-Allowed-to-Fail.

dereuromark:
You need to keep it in the matrix + add it to the allow failure :)
Otherwise it won't exist anymore.

ADmad:
Ah, fixed. Thanks :)

dereuromark:
Only 415 errors  :P

PR #11990, opened 2018-04-24
Update the range of isOk as discussed in #11796

Iandenh:
Change the HTTP code range of the `isOk` method as discussed in #11796.

We could also add a `isSuccesful` method for the 2xx range.


inoas:
**isSuccessful()** would be good because then it is easy to migrate if you are really only looking for 2xx (as **isOk()** was doing before)

PR #11991, opened 2018-04-24
[4.x] View deprecations

iliepandia:
Refs #11934 

Removed deprecations from View.

I have removed the properties, methods and files that were marked as `deprecated`. 

Also I have removed code that was conditional on the deprecated properties. Examples: 

- FormHelper::create()  `$options['action']`
- FormHelper::__construct() `$config['registry']`
- View::__set() (it was marked as magic setter for **deprecated** properties)
- parts of View::__get() (most of it needs to stay as it loads Helper classes)
- and others...

### Removed tests 

I know removing tests is a bad idea but some of them I could not refactor.

- The RssHelper tests are no longer valid since the RssHelper itself was removed
- The Bread-crumbs tests from the HtmlHelper since the BreadCrumbs have been moved into their own helper.


src/View/View.php
```diff
@@ -110,14 +108,6 @@ class View implements EventDispatcherInterface
      */
     public $name;
 
-    /**
-     * Current passed params. Passed to View from the creating Controller for convenience.
-     *
-     * @var array
-     * @deprecated 3.1.0 Use `$this->request->getParam('pass')` instead.
-     */
-    public $passedArgs = [];
```
markstory: Should also remove the code that sets this property.

tests/TestCase/View/ViewBuilderTest.php
```diff
@@ -66,9 +66,16 @@ public function arrayPropertyProvider()
...
-        $this->assertNull($builder->{$property}(), 'Default value should be null');
-        $this->assertSame($builder, $builder->{$property}($value), 'Setter returns this');
-        $this->assertSame($value, $builder->{$property}(), 'Getter gets value.');
+        $setter = 'set' . ucfirst($property);
+        $getter = 'get' . ucfirst($property);
+        if (is_bool($value)) {
+            //for boolean values the setter and getter look different
+            $setter = 'enable' . ucfirst($property);
+            $getter = 'is' . ucfirst($property) . 'Enabled';
+        }
```
markstory: Boolean properties should probably be separate test cases.

src/View/View.php
iliepandia: @markstory , can you clarify a bit? Where would that code be?

tests/TestCase/View/ViewBuilderTest.php
iliepandia: makes sense - moved `autoLayout` to it's own test.

markstory:
Related to #11992

tests/TestCase/View/ViewBuilderTest.php
```diff
@@ -57,6 +68,27 @@ public function arrayPropertyProvider()
...
+     */
+    public function testBooleanProperties($property, $value)
+    {
+        $builder = new ViewBuilder();
+
+        //for boolean values the setter and getter look different than set/get
+        $setter = 'enable' . ucfirst($property);
+        $getter = 'is' . ucfirst($property) . 'Enabled';
+
+        $this->assertNull($builder->{$getter}(), 'Default value should be null');
```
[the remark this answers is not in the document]
markstory: Either works. Defaulting to false makes more sense as we'll want to add typehints and there is no good reason for these methods to have nullable return types.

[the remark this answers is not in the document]
iliepandia: OK, before changing this, the only bool property under test right now is autoLayout. But the default for that should be `true` right? By default ViewBuilder should add the layout to the view contents. Or we keep this default on `false` and let the controller be responsible to set it to `true` when it builds the view?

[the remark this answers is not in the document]
iliepandia: OK, `ViewBuilder::$_autoLayout` now defaults to `true` updated the tests, merged the 4.x branch into this PR and fixed the conflicts which was to remove again tests for deprecated code that no longer exists.

iliepandia:
The Utility tests are failing and I don't know what is the correct way to handle this as it seems most of that code was removed as deprecated in the Utility package.

iliepandia:
Hi, it's not clear for me if you're waiting for me to do something :)

ADmad:
@iliepandia There's merge conflict now. Please just fix that and we can take care of the rest.

src/View/View.php
saeideng: @iliepandia  all usage of `passedArgs` in view 
https://github.com/cakephp/cakephp/blob/843c5a480824c5429b13dd39c7437afb0e01072d/src/View/View.php#L236

and in tests

iliepandia: I have removed the reference from the line 236. There are indeed more references to this and tests, but they related to how the `Controller` builds the view and I did not touch on that. I assumed that each package fix will only work on their respective package files so I did not change how the controller works.

iliepandia:
Decided to remove the 'passedArgs' from the `Controller` `$_validViewOptions` options after all and updated the `ViewVarsTraitTest` file to reflect that.

markstory:
Thanks @iliepandia :clap:

PR #11992, opened 2018-04-24
Add warnings that were missed before.

markstory:
Refs #11988


saeideng:
@markstory 
and maybe on `Cake\View\ViewBuilder::template()` ! ?

markstory:
@saeideng Looks like I missed a bunch of methods in ViewBuilder. I think I got them all now.

PR #11993, opened 2018-04-24
Fix errors reported by phpstan in Mail, Core, Routing packages.

ADmad:


src/Core/ObjectRegistry.php
```diff
@@ -93,7 +93,7 @@ public function load($objectName, $config = [])
             list($plugin, $objectName) = pluginSplit($objectName);
             $this->_throwMissingClassError($objectName, $plugin);
         }
-        $instance = $this->_create($className, $name, $config);
+        $instance = $this->_create((string)$className, $name, $config);
```
markstory: This isn't going to work as sometimes `$className` is an object.

src/Core/functions.php
```diff
@@ -28,7 +28,7 @@
      * @param string|array|object $text Text to wrap through htmlspecialchars. Also works with arrays, and objects.
      *    Arrays will be mapped and have all their elements escaped. Objects will be string cast if they
      *    implement a `__toString` method. Otherwise the class name will be used.
-     * @param bool $double Encode existing html entities.
+     * @param bool|string $double Encode existing html entities. If string it will be used as character set.
```
dereuromark: Should we remove those shim behaviors for a cleaner api? Or do we go with $options array here?

src/View/Helper/HtmlHelper.php
```diff
@@ -1090,13 +1090,13 @@ public function div($class = null, $text = null, array $options = [])
...
      * @param string $class CSS class name of the p element.
-     * @param string $text String content that will appear inside the p element.
+     * @param string|null $text String content that will appear inside the p element.
      * @param array $options Additional HTML attributes of the P tag
      * @return string The formatted P element
      */
     public function para($class, $text, array $options = [])
     {
-        if (isset($options['escape'])) {
+        if (!empty($options['escape'])) {
```
dereuromark: That sounds actually like a bug that should be fixed ealier in 3.x? false would be isset() true.

ADmad: Yup it's a bug which should be backported.

PR #11994, opened 2018-04-24
Allow only classname strings for type map.

ADmad:


markstory:
Looks good to me.

ADmad:
@markstory @lorenzo What do you guys think about renaming `Type` to `TypeFactory` or `TypeBuilder`?

ADmad:
I just noticed that the 4.0 roadmap says _"Remove non-static methods from Cake\Database\Type and move the common instance methods into a trait."_

I added base class `BaseType` instead of a trait in previous PR. I hope that's okay :slightly_smiling_face:.

lorenzo:
Wouldn’t renaming go against the idea of 4.0 be almost the same as 3.6?

ADmad:
@lorenzo We could do the rename in 3.7 and keep old name as alias with deprecation warning as we have done for other renames. But if you think that's unnecessary hassle we can keep the class name as it is.

dereuromark:
> I added a base class BaseType instead of a trait in previous PR. I hope that's okay

A base type is in these cases always a better choice than trait IMO.
A trait should be used to dynamically hook in unrelated things into classes (e.g. logging into controllers).
But otherwise normal OOP should be used.

lorenzo:
@ADmad That sounds like a good idea, if we actually have a 3.7

markstory:
> That sounds like a good idea, if we actually have a 3.7

I'm pretty sure we will have a 3.7 at this point. It will just be small and be primarily community driven.

ADmad:
I am going ahead with the rename then :smile:.

ADmad:
`TypeFactory` or `TypeBuilder`?

ADmad:
I can do the rename in a separate PR.

markstory:
> If we add the alias in 3.7 we can remove in 4.0. I don't think keeping it till 5.0 is necessary.

We've been telling people that if they get to 3.6 with no warnings they'll be able to upgrade to 4.0 with relative ease. Is it worth adding a hard bump on a class rename? The alias will cost us nothing in the short term.

PR #11998, opened 2018-04-25
[4.x] Add isSuccess to Http/Client/Response

Iandenh:
refs: #11990 #11536 #11796

### DIFF
--- a/src/Http/Client/Response.php
+++ b/src/Http/Client/Response.php
@@ -237,6 +237,16 @@ class Response extends Message implements ResponseInterface
     }
 
     /**
+     * Check if the response status code was in the 2xx range
+     *
+     * @return bool
+     */
+    public function isSuccessful()
+    {
+        return $this->code >= 200 && $this->code <= 299;
+    }
+
+    /**
      * Check if the response had a redirect status code.
      *
      * @return bool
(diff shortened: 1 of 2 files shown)


src/Http/Client/Response.php
```diff
@@ -236,6 +236,16 @@ public function isOk()
         return $this->code >= 200 && $this->code <= 399;
     }
 
+    /**
+     * Check if the response status code was in the 2xx range
+     *
+     * @return bool
+     */
+    public function isSuccessful()
```
ADmad: Nitpick: Can we have a shorter name `isSuccess()`? :slightly_smiling_face:

Iandenh: Sure

tests/TestCase/Http/Client/ResponseTest.php
```diff
@@ -244,6 +244,77 @@ public function testIsOk()
...
+                    'HTTP/1.1 203 Non-Authoritative Information',
+                    'Content-Type: text/html'
+                ], 'ok')
+            ],
+            [
+                true,
+                new Response([
+                    'HTTP/1.1 204 No Content',
+                    'Content-Type: text/html'
+                ], 'ok')
```
markstory: Why does 204 have content?

Iandenh: Sorry, just copy the test data from `testIsOk`. Shall I also remove the content from the response there?

Issue #11957, opened 2018-04-17
Xml::build() BOM "\xEF\xBB\xBF"


salines:
Just report,

i have unable to open and read some xml files with
Xml::build() method

UPDATE:
```
$file = new File($local_xml);
$f = $file->read()
$xml = Xml::build($f);
```
Error:

Exception: Xml cannot be read. in [/vagrant/vendor/cakephp/cakephp/src/Utility/Xml.php, line 168]
```
2018-04-17 17:37:03 Error: [Cake\Utility\Exception\XmlException] Xml cannot be read. in /vagrant/vendor/cakephp/cakephp/src/Utility/Xml.php on line 168
Stack Trace:
#0 /vagrant/vendor/cakephp/cakephp/src/Utility/Xml.php(123): Cake\Utility\Xml::_loadXml('\xEF\xBB\xBF<?xml versio...', Array)
#1 /vagrant/src/Shell/AubiplusShell.php(155): Cake\Utility\Xml::build('/vagrant/tmp/au...')
#2 /vagrant/src/Shell/AubiplusShell.php(136): App\Shell\AubiplusShell->local('/vagrant/tmp/au...', Object(SimpleXMLElement))
#3 /vagrant/src/Shell/AubiplusShell.php(78): App\Shell\AubiplusShell->remote('https://service...')
#4 /vagrant/src/Shell/AubiplusShell.php(69): App\Shell\AubiplusShell->auto()
#5 /vagrant/vendor/cakephp/cakephp/src/Console/Shell.php(507): App\Shell\AubiplusShell->main()
#6 /vagrant/vendor/cakephp/cakephp/src/Console/CommandRunner.php(141): Cake\Console\Shell->runCommand(Array, true)
#7 /vagrant/bin/cake.php(12): Cake\Console\CommandRunner->run(Array)
#8 {main}
```

but success with:

`$xml = simplexml_load_file($local_xml);`

CakePHP 3.5

markstory:
Does `new SimpleXmlElement()` work with your file?

markstory:
Closing as there is a pull request up to improve error handling making this issue easier to understand.

PR #11999, opened 2018-04-25
Xml better error reporting

iliepandia:
Refs #11957 

The diagnosis of this issue would have been easier if the underlying cause was not obscured. 

We can now use the `finally` construct since CakPHP requires PHP >5.6, so we can properly chain the exception for a more helpful error message and stack trace when something fails. 

### Tests

Added a new test that will look for this chained exception when the specified XML input should cause one.


iliepandia:
I will need help with the code coverage fails. :)

markstory:
Thanks!

PR #12002, opened 2018-04-25
Revert "Fix up behavior callbacks for clean inheritance and consisten…

ADmad:
…cy."

This reverts commit 9e93097b504583a84a827a2c4a41a24e16650470.

Changing argument hints is not backwards compatible and causes error when these
methods are overridden in extending classes.

### DIFF
--- a/src/ORM/Behavior/CounterCacheBehavior.php
+++ b/src/ORM/Behavior/CounterCacheBehavior.php
@@ -14,7 +14,6 @@
  */
 namespace Cake\ORM\Behavior;
 
-use ArrayObject;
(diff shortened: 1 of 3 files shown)


dereuromark (approved):
4.0 then.
A bummer that at the time no one saw this.. 
Thx for doing this.

ADmad:
The change still already exists in `4.x` branch. Just need to ensure it doesn't get reverted during 3.x to 4.x merge :slightly_smiling_face:.

dereuromark:
... And 4.x migration guide then gets completed this time.

PR #12004, opened 2018-04-25
Removed deprecated classes in Network package.

ADmad:


tests/TestCase/Http/ResponseTest.php
```diff
@@ -1833,7 +1833,7 @@ public function testCorsParameters($request, $origin, $domains, $methods, $heade
             $response = new Response();
 
             $result = $response->cors($request, $domains, $methods, $headers);
-            $this->assertInstanceOf('Cake\Network\CorsBuilder', $result);
+            $this->assertInstanceOf('Cake\Http\CorsBuilder', $result);
```
dereuromark: Can we use ::class to make refacting easier and allow IDE discovery?

ADmad: This whole method is gonna go away once deprecations are removed. It's a temp fix to get the builds passing.

saeideng (approved):

PR #12006, opened 2018-04-25
Backport changes done to Type class in 4.x.

ADmad:
Refs #11975, #11979 and #11994.

### DIFF
--- a/src/Database/Type.php
+++ b/src/Database/Type.php
@@ -145,8 +145,12 @@ class Type implements TypeInterface
      * If called with no arguments it will return current types map array
      * If $className is omitted it will return mapped class for $type
      *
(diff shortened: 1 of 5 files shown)


src/Database/Type.php
```diff
@@ -145,8 +145,12 @@ public static function set($name, Type $instance)
      * If called with no arguments it will return current types map array
      * If $className is omitted it will return mapped class for $type
      *
-     * Deprecated: The usage of $type as \Cake\Database\Type[] is deprecated. Please always use string[] if you pass an array
-     * as first argument.
+     * Deprecated:
```
dereuromark: Do we know the version when it was deprecated and the new methods added? This will help when having to refactor plugins and to know what min requirement to specify. We should add it here then.

markstory:
Bunch of deprecations to update.

Issue #11982, opened 2018-04-23
Form validation throws deprecation messages if no rules are set


highstrike:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.6.1

### What you did
I have a modelless form which now contains the new 3.6.x `validationDefault` function (replaced the old protected `_buildValidator` function)
```
public function validationDefault(Validator $validator) {
    return $validator;
}
```

### What happened
![image](https://user-images.githubusercontent.com/2379538/39130888-dbdca758-4716-11e8-9ce1-806cd87869fd.png)

It raised a deprecation message because of this line right here -> https://github.com/cakephp/cakephp/blob/master/src/Form/Form.php#L227

Cake actually calls a deprecated function itself because I didn't explicitly set any validation rules. The form itself is very simple, it doesn't need any validation rules.

Even if I delete the function `validationDefault` from my modelless form it still raises the same deprecation message

### What you expected to happen
not raise any deprecation message 😅 

cc @jeremyharris - https://cakesf.slack.com/archives/C053DPNGT/p1524490620000178

ADmad:
This is a tricky one. The call to deprecated method `validator()` was added in #11857 in an attempt to ensure backward compatibility. Though the usage of overriding `validator()` to add validation is kinda wrong itself. As per the docs the developer was supposed to override `_buildValidator()` to configure the validator.

markstory:
Only emitting the deprecation warning when the subclass implements the `validator()` hook might let us work around this issue.

markstory:
Pull request open now.

PR #12009, opened 2018-04-26
Fix for PR #11985

highstrike:
Refs https://github.com/cakephp/cakephp/issues/11982 & https://github.com/cakephp/cakephp/pull/11985


markstory:
Thanks!

PR #12012, opened 2018-04-26
Get phpstan build passing at level 2.

ADmad:
Aiming for level 5 was a bit over ambitious at this point. Also having the build under "allow failures" means new errors could be overlooked. Bumping the levels sequentially would be a better approach. Up to level 2 now vs. level 1 on master :muscle:.

### DIFF
--- a/src/Network/Socket.php
+++ b/src/Network/Socket.php
@@ -315,7 +315,7 @@ class Socket
     /**
      * Set the last error.
      *
-     * @param int $errNum Error code
+     * @param int|null $errNum Error code
      * @param string $errStr Error string
      * @return void
      */
--- a/src/View/Helper/NumberHelper.php
+++ b/src/View/Helper/NumberHelper.php
@@ -152,7 +152,7 @@ class NumberHelper extends Helper
         $formatted = $this->_engine->format($number, $options);
         $options += ['escape' => true];
 
-        return $options['escape'] ? h($formatted) : $formatted;
+        return $options['escape'] ? (string)h($formatted) : $formatted;
     }
 
     /**
@@ -184,7 +184,7 @@ class NumberHelper extends Helper
         $formatted = $this->_engine->currency($number, $currency, $options);
         $options += ['escape' => true];
 
-        return $options['escape'] ? h($formatted) : $formatted;
+        return $options['escape'] ? (string)h($formatted) : $formatted;
     }
 
     /**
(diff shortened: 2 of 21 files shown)


src/Network/Socket.php
```diff
@@ -315,7 +315,7 @@ public function lastError()
     /**
      * Set the last error.
      *
-     * @param int $errNum Error code
+     * @param int|null $errNum Error code
```
dereuromark: Is that true? Or should instead the calling code better cast?

src/View/View.php
```diff
@@ -648,7 +648,7 @@ public function setLayout($name)
      *
      * @deprecated 3.5.0 Use getLayout()/setLayout() instead.
      * @param string|null $name Layout file name to set. If null returns current name.
-     * @return string|null
+     * @return string|false
```
dereuromark: It can still be null (if not yet set), right?

ADmad: No, the property is set to `'default'` during declaration.

dereuromark: You are right, and setTemplate($name) strictly speaking does not accept to make it null again.

src/View/Helper/NumberHelper.php
```diff
@@ -152,7 +152,7 @@ public function format($number, array $options = [])
         $formatted = $this->_engine->format($number, $options);
         $options += ['escape' => true];
 
-        return $options['escape'] ? h($formatted) : $formatted;
+        return $options['escape'] ? (string)h($formatted) : $formatted;
```
lorenzo: Why was this cast necessary? Seems wasteful

dereuromark: Because h() being a god function.. Both string and array input output.

ADmad: `h()` can return a string or array, hence without the cast phpstan is unhappy.

Perhaps in 4.x we can make it just accept and return strings. I doubt passing array to it is commonly used.

lorenzo:
Do you think this may be a perf regression? This is the type of functions that get called very often in a loop

ADmad:
@lorenzo I removed the string casing in 27537e55c32503957e67d0f5098ec8e44823a645. The earlier `string|array` return type was a lie anyway since the method has this:
```php
        } elseif (is_bool($text) || is_null($text) || is_int($text)) {
            return $text;
        }
``

Mistakenly directly commit to 4.x instead of making a PR.

lorenzo:
Great!

PR #12016, opened 2018-04-27
4.x db typefactory

ADmad:


ADmad:
In previous PRs we discussed backporting this rename to 3.x but I am not too sure about that now. In 3.x `Type` is still a hybrid class being used as factory as well as base class. So it would be lot more hassle to do the rename there.

Also since we are going to carry this alias till 5.0 doing the rename in 4.0 should be enough IMO and not cause any issue when upgrading from 3.x to 4.0.

src/Database/Type/DateType.php
```diff
@@ -27,7 +27,7 @@ class DateType extends DateTimeType
      * class is constructed. After that use `useMutable()` or `useImmutable()` instead.
      *
      * @var string
-     * @deprecated 3.2.0 Use DateType::useMutable() or DateType::useImmutable() instead.
+     * @deprecated 3.2.0 Use DateTypeFactory::useMutable() or DateTypeFactory::useImmutable() instead.
```
markstory: Looks like find and replace casualties.

PR #12019, opened 2018-04-27
[minor] Static Code Analysis

kalessil:
Followup on https://twitter.com/cakephp/status/989052189304815617

This PR doesn't contain bug-fixes, only code style, control flow and core functions usage tweaks and intends to groom the codebase a little bit.

### DIFF
--- a/src/Console/ConsoleIo.php
+++ b/src/Console/ConsoleIo.php
@@ -455,8 +455,9 @@ class ConsoleIo
             $defaultText = "[$default] ";
         }
         $this->_out->write('<question>' . $prompt . "</question>$optionsText\n$defaultText> ", 0);
+        $result = $this->_in->read();
 
-        $result = trim($this->_in->read());
+        $result = trim($result);
         if ($default !== null && ($result === '' || $result === null)) {
             return $default;
         }
@@ -550,7 +551,8 @@ class ConsoleIo
 
         if (file_exists($path) && $forceOverwrite === false) {
             $this->warning("File `{$path}` exists");
-            $key = strtolower($this->askChoice('Do you want to overwrite?', ['y', 'n', 'a', 'q'], 'n'));
+            $key = $this->askChoice('Do you want to overwrite?', ['y', 'n', 'a', 'q'], 'n');
+            $key = strtolower($key);
 
             if ($key === 'q') {
                 $this->error('Quitting.', 2);
--- a/src/Console/ConsoleIo.php
+++ b/src/Console/ConsoleIo.php
@@ -455,9 +455,8 @@ class ConsoleIo
             $defaultText = "[$default] ";
         }
         $this->_out->write('<question>' . $prompt . "</question>$optionsText\n$defaultText> ", 0);
-        $result = $this->_in->read();
 
-        $result = trim($result);
+        $result = trim($this->_in->read());
         if ($default !== null && ($result === '' || $result === null)) {
             return $default;
         }
@@ -551,8 +550,7 @@ class ConsoleIo
 
         if (file_exists($path) && $forceOverwrite === false) {
             $this->warning("File `{$path}` exists");
-            $key = $this->askChoice('Do you want to overwrite?', ['y', 'n', 'a', 'q'], 'n');
-            $key = strtolower($key);
+            $key = strtolower($this->askChoice('Do you want to overwrite?', ['y', 'n', 'a', 'q'], 'n'));
 
             if ($key === 'q') {
                 $this->error('Quitting.', 2);
--- a/src/Core/Plugin.php
+++ b/src/Core/Plugin.php
@@ -241,8 +241,7 @@ class Plugin
             }
         }
         if (Configure::check('plugins')) {
-            $plugins = array_merge($plugins, array_keys(Configure::read('plugins')));
-            $plugins = array_unique($plugins);
+            $plugins = array_unique(array_merge($plugins, array_keys(Configure::read('plugins'))));
         }
 
         $collection = static::getCollection();
--- a/src/Core/Plugin.php
+++ b/src/Core/Plugin.php
@@ -241,7 +241,8 @@ class Plugin
             }
         }
         if (Configure::check('plugins')) {
-            $plugins = array_unique(array_merge($plugins, array_keys(Configure::read('plugins'))));
+            $plugins = array_merge($plugins, array_keys(Configure::read('plugins')));
+            $plugins = array_unique($plugins);
         }
 
         $collection = static::getCollection();
--- a/src/Core/PluginCollection.php
+++ b/src/Core/PluginCollection.php
@@ -131,7 +131,7 @@ class PluginCollection implements Iterator, Countable
      */
     public function next()
     {
-        $this->position += 1;
+        ++$this->position;
     }
 
     /**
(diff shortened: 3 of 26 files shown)


src/Console/ConsoleIo.php
```diff
@@ -551,8 +550,7 @@ public function createFile($path, $contents, $forceOverwrite = false)
 
         if (file_exists($path) && $forceOverwrite === false) {
             $this->warning("File `{$path}` exists");
-            $key = $this->askChoice('Do you want to overwrite?', ['y', 'n', 'a', 'q'], 'n');
-            $key = strtolower($key);
+            $key = strtolower($this->askChoice('Do you want to overwrite?', ['y', 'n', 'a', 'q'], 'n'));
```
dereuromark: Doing more than a single operation per line is usually considered bad practice.
Harder to debug and easier to overlook issues.

src/Core/Plugin.php
```diff
@@ -241,8 +241,7 @@ public static function loadAll(array $options = [])
             }
         }
         if (Configure::check('plugins')) {
-            $plugins = array_merge($plugins, array_keys(Configure::read('plugins')));
-            $plugins = array_unique($plugins);
+            $plugins = array_unique(array_merge($plugins, array_keys(Configure::read('plugins'))));
```
dereuromark: Same, and especially here.

src/Core/functions.php
```diff
@@ -51,7 +51,7 @@ function h($text, $double = true, $charset = null)
             } else {
                 $text = '(object)' . get_class($text);
             }
-        } elseif (is_bool($text) || is_null($text) || is_int($text)) {
+        } elseif (is_bool($text) || is_int($text) || $text === null) {
```
dereuromark: If you want to do the nano opt here, then the null check should be first, not last :)

src/Datasource/EntityTrait.php
```diff
@@ -1356,7 +1356,7 @@ public function source($alias = null)
             get_called_class() . '::source() is deprecated. ' .
             'Use setSource()/getSource() instead.'
         );
-        if (is_null($alias)) {
+        if ($alias === null) {
```
dereuromark: :+1: People should just add my Spryker sniffer rule to the CakePHP rules repo :)

src/Filesystem/Folder.php
```diff
@@ -345,7 +345,7 @@ public static function isAbsolute($path)
      */
     public static function isRegisteredStreamWrapper($path)
     {
-        return preg_match('/^[^:\/\/]+?(?=:\/\/)/i', $path, $matches) &&
+        return preg_match('/^[^:\/\/]+?(?=:\/\/)/', $path, $matches) &&
```
dereuromark: Can this be a functional regression? isAbsolute() still uses /i for example.

kalessil: Nope, the regex has no any characters. `i` is not needed.

src/View/StringTemplate.php
```diff
@@ -180,7 +180,7 @@ protected function _compileTemplates(array $templates = [])
             }
 
             $template = str_replace('%', '%%', $template);
-            preg_match_all('#\{\{([\w\d\._]+)\}\}#', $template, $matches);
+            preg_match_all('#\{\{([\w\._]+)\}\}#', $template, $matches);
```
dereuromark: Do we have a test case for this to not introduce an accidental regression?

src/Datasource/EntityTrait.php
kalessil: + PHP CS Fixer has `is_null` rule. Hopefully, one of them can be added.

src/Core/Plugin.php
ADmad: I too don't like this. Reduces code readability.

src/Core/functions.php
kalessil: Ideally in separate PR addressing similar cases around the codebase.

src/Event/EventManager.php
```diff
@@ -366,8 +366,7 @@ public function listeners($eventKey)
         $globalListeners = static::instance()->prioritisedListeners($eventKey);
         $globalListeners = empty($globalListeners) ? [] : $globalListeners;
 
-        $priorities = array_merge(array_keys($globalListeners), array_keys($localListeners));
-        $priorities = array_unique($priorities);
+        $priorities = array_unique(array_merge(array_keys($globalListeners), array_keys($localListeners)));
```
ADmad: Same comment as above. Too much going on in a single statement.

src/View/StringTemplate.php
kalessil: One of the un-testable things: a greedy regex. `\d` is sub-set of `\w`, what only can hurt performance.

src/Core/Plugin.php
kalessil: Ok, I probably don't have enough context of the project. Will revert this specific inspection fixes.

src/Database/Dialect/SqliteDialectTrait.php
```diff
@@ -120,7 +120,7 @@ protected function _transformFunctionExpression(FunctionExpression $expression)
                     ->setConjunction(' ,')
                     ->iterateParts(function ($p, $key) {
                         if ($key === 0) {
-                            $value = rtrim(strtolower($p), 's');
+                            $value = strtolower(rtrim($p, 's'));
```
ADmad: This is not functionally same. Earlier a capital `S` at end of string would have been trimmed, now it won't be.

src/Core/functions.php
kalessil: OK, gimme a sec

src/Database/Dialect/SqliteDialectTrait.php
kalessil: Hm, good catch. A false positive, reverting.

src/Database/Driver/PDODriverTrait.php
```diff
@@ -158,7 +158,7 @@ public function rollbackTransaction()
             return false;
         }
 
-        return $this->_connection->rollback();
+        return $this->_connection->rollBack();
```
ADmad: Existing method casing is already correct.

https://github.com/cakephp/cakephp/blob/62944e3bc04bd62459b3fc43c33e06d245837cb1/src/Database/Connection.php#L544

dereuromark: I wonder how that tool thinks otherwise.

src/Core/PluginCollection.php
```diff
@@ -131,7 +131,7 @@ public function get($name)
      */
     public function next()
     {
-        $this->position += 1;
+        ++$this->position;
```
markstory: The old form is easier to read in my opinion.

src/Database/Driver/PDODriverTrait.php
chinpei215: The _connection seems to be an instance of PDI. So PDO::rollBack() would be correct.

src/Core/PluginCollection.php
josegonzalez: We use the prefix version? I didn't even know that was a thing in PHP.

markstory: We do use the suffix form, but generally try to avoid the prefix form as it uncommon.

markstory: I found 4 other places we are using prefix incrementing. I'll normalize all of the prefix usage after merging this.

kalessil: Perfect, thank you, Mark.

PR #12020, opened 2018-04-28
Remove usage of prefix increment operator.

markstory:
This operator is less frequently used and can be harder to understand than the more common suffix operator.

Refs #12019

### DIFF
--- a/src/Utility/Text.php
+++ b/src/Utility/Text.php
@@ -138,7 +138,7 @@ class Text
                         }
                     }
                 }
-                $offset = ++$tmpOffset;
+                $offset = $tmpOffset++;
             } else {
                 $results[] = $buffer . mb_substr($data, $offset);
                 $offset = $length + 1;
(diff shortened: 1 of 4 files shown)


src/Utility/Text.php
```diff
@@ -138,7 +138,8 @@ public static function tokenize($data, $separator = ',', $leftBound = '(', $righ
                         }
                     }
                 }
-                $offset = ++$tmpOffset;
+                $tmpOffset += 1;
+                $offset = $tmpOffset;
```
saeideng: why not 
```php
$offset = $tmpOffset+1;
```

dereuromark: That is sth different.

saeideng: @dereuromark  I mistake :+1:

src/View/Helper/HtmlHelper.php
```diff
@@ -982,10 +982,11 @@ protected function _renderCells($line, $useCount = false)
             }
 
             if ($useCount) {
+                $i += 1;
```
dereuromark: Should we then use `$i++` for consistency?

PR #12021, opened 2018-04-28
4.x type cleanup

ADmad:
- Remove deprecated property
- Immutable date time classes are now used by default

### DIFF
--- a/tests/TestCase/ORM/Behavior/TimestampBehaviorTest.php
+++ b/tests/TestCase/ORM/Behavior/TimestampBehaviorTest.php
@@ -16,6 +16,7 @@ namespace Cake\Test\TestCase\ORM\Behavior;
 
 use Cake\Database\TypeFactory;
 use Cake\Event\Event;
+use Cake\I18n\FrozenTime;
 use Cake\I18n\Time;
 use Cake\ORM\Behavior\TimestampBehavior;
 use Cake\ORM\Entity;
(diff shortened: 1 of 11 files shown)


ADmad:
Not sure why only Postgres builds are failing.

markstory:
Maybe there are postgres specific tests?

ADmad:
@markstory I don't think so.

markstory:
Odd, I have postgres installed locally and can try to reproduce.

ADmad:
@markstory Thanks.

ADmad:
` QueryRegressionTest` passes in isolation. The failures seen on travis occur only when running full test suite.

ADmad:
@markstory Tests fixed.

I removed `BaseType::__debugInfo()`. While debugging the test suite failures I found only getting type name pretty useless. I wanted to check whether internally mutable or non mutable class was being used, heence needed to see value of `DateTimeType::$_datetimeInstance` too.

tests/TestCase/ORM/Behavior/TimestampBehaviorTest.php
```diff
@@ -241,6 +240,9 @@ public function testUseImmutable()
         $entity->clean();
         $this->Behavior->handleEvent($event, $entity);
         $this->assertInstanceOf('Cake\I18n\Time', $entity->modified);
+        // Revert back to using immutable class to avoid causing problems in
+        // other test cases when running full test suite.
+        TypeFactory::build('timestamp')->useImmutable();
```
markstory: :clap: good find.

markstory:
Refs #11934

PR #12022, opened 2018-04-29
3.6 h()

ADmad:
Currently integers are returned as is but floats are converted to string, don't see why that should be the case.

### DIFF
--- a/src/Core/functions.php
+++ b/src/Core/functions.php
@@ -64,6 +64,10 @@ if (!function_exists('h')) {
             }
         }
         if (is_string($double)) {
+            deprecationWarning(
+                'Passing charset string for 2nd argument is deprecated. ' .
+                'Use the 3rd argument instead.'
+            );
             $charset = $double;
         }
 
--- a/src/Core/functions.php
+++ b/src/Core/functions.php
@@ -25,13 +25,14 @@ if (!function_exists('h')) {
     /**
      * Convenience method for htmlspecialchars.
      *
-     * @param string|array|object $text Text to wrap through htmlspecialchars. Also works with arrays, and objects.
+     * @param mixed $text Text to wrap through htmlspecialchars. Also works with arrays, and objects.
      *    Arrays will be mapped and have all their elements escaped. Objects will be string cast if they
      *    implement a `__toString` method. Otherwise the class name will be used.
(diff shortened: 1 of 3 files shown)


ADmad:
Currently the method allows passing charset string for 2nd param `$double` and does a wasteful `is_string()` check on very call.

I propose deprecating passing string to 2nd param.

/cc @lorenzo

lorenzo:
@admad deprecating that sounds good

markstory:
Looks good I will merge this after I get the migration guide updated.

ADmad:
I doubt anyone actually used the method by passing string as 2nd param. While the function sets `$charset = $double;` it never resets `$double` to a boolean as `htmlspecialchars()` expects resulting in error: 
`PHP Warning:  htmlspecialchars() expects parameter 2 to be integer, string given`.

It's best to leave it broken :smile:

src/Core/functions.php
```diff
@@ -25,13 +25,14 @@
     /**
      * Convenience method for htmlspecialchars.
      *
-     * @param string|array|object $text Text to wrap through htmlspecialchars. Also works with arrays, and objects.
+     * @param mixed $text Text to wrap through htmlspecialchars. Also works with arrays, and objects.
```
dereuromark: We shoudn't weaken it maybe? It should only be a string or array usually, it seems it can also be an object though.

ADmad: I just fixed the docblock. You can already pass practically any data type to the method.

dereuromark: That is always true in php 5 classes. We still only typehint expected types.

ADmad: PHP version doesn't matter. Even in 4.x we won't be able to add any actual type hint for `$text` argument since it can be of multiple type. So replacing `mixed` with list of half a dozen or so types would be silly.

dereuromark: I am a bit worried about this method losing documention on its primary intent: `string|string[]` - it works on strings, nothing else. The others can be passed for simplicity only, and for performance make a no-op. But passing those is actually a bit silly and as I mentioned for simplicity of the templates.
So while yours functionally correct in the strict sense, the doc block here can still state more the intend of the method IMO and the expected primary input types (rest is just noise).

PR #12028, opened 2018-04-30
findOrCreate never fires Model.afterSaveCommit

jeremyharris:
`Table::findOrCreate` would never fire `Model.afterSaveCommit`. I don't know anything about nested transactions, so perhaps someone can illuminate why it wrapped `_processFindOrCreate` in a transaction when the save operation in that method would take care of the transaction itself.

### DIFF
--- a/src/ORM/Table.php
+++ b/src/ORM/Table.php
@@ -1678,14 +1678,13 @@ class Table implements RepositoryInterface, EventListenerInterface, EventDispatc
         $options += [
             'atomic' => true,
             'defaults' => true,
-            '_primary' => true,
         ];
 
         $entity = $this->_executeTransaction(function () use ($search, $callback, $options) {
             return $this->_processFindOrCreate($search, $callback, $options);
         }, $options['atomic']);
 
-        if ($entity && $this->_transactionCommitted($options['atomic'], $options['_primary'])) {
+        if ($entity && $this->_transactionCommitted($options['atomic'], true)) {
             $this->dispatchEvent('Model.afterSaveCommit', compact('entity', 'options'));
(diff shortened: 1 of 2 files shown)


src/ORM/Table.php
```diff
@@ -1680,9 +1680,7 @@ public function findOrCreate($search, callable $callback = null, $options = [])
             'defaults' => true
         ];
 
-        return $this->_executeTransaction(function () use ($search, $callback, $options) {
```
markstory: I think the goal of this wrapping transaction was that the find and save were executed as one transaction. In repeatable-read or [higher isolation modes](https://en.wikipedia.org/wiki/Isolation_(database_systems)#Read_phenomena) this wrapping transaction aims to help prevent dirty and phantom reads.

With that said, I don't know how we solve the read phenomena and also fire the aftercommit event.

jeremyharris:
Turns out I had to dispatch the event in `findOrCreate` as setting atomic to `false` on the save won't dispatch the event because, well, it's not atomic 😄

src/ORM/Table.php
```diff
@@ -1677,10 +1677,19 @@ public function findOrCreate($search, callable $callback = null, $options = [])
     {
         $options += [
             'atomic' => true,
-            'defaults' => true
+            'defaults' => true,
+            '_primary' => true,
```
markstory: New option should go in the doc block too.

jeremyharris: Turns out this isn't necessary for this method, since it's always primary. I was going to document it on `save()` since it's undocumented, but it seems like an internal option.

PR #12031, opened 2018-05-01
Load routing in console environments.

markstory:
Ensure that the CommandRunner framework uses the application routes() hook to load routes. This enables the routes shell and other commands/shells to generate URLs properly.

Refs #12030 
Refs cakephp/app#593


lorenzo:
Nice!

PR #12039, opened 2018-05-03
Add comment that $routes parameter is currently unused.

ravage84:


src/Http/BaseApplication.php
```diff
@@ -160,7 +160,7 @@ public function pluginBootstrap()
      *
      * By default this will load `config/routes.php` for ease of use and backwards compatibility.
      *
-     * @param \Cake\Routing\RouteBuilder $routes A route builder to add routes into.
+     * @param \Cake\Routing\RouteBuilder $routes A route builder to add routes into (currently unused).
```
ADmad: This could confuse users. While the base class doesn't use it extending classes aka app's `Application` class is expected to use the `$routes` argument if overriding the method.

markstory:
Closing as this parameter *can* be used by subclasses, and the parameter is also available in the `config/routes.php` file as a local variable.

PR #12041, opened 2018-05-03
Fixed ConsoleIntegrationTestCase failing to load user bootstrap files

jeremyharris:

### DIFF
--- a/src/TestSuite/ConsoleIntegrationTestCase.php
+++ b/src/TestSuite/ConsoleIntegrationTestCase.php
@@ -253,7 +253,7 @@ abstract class ConsoleIntegrationTestCase extends TestCase
         if ($this->_useCommandRunner) {
             $applicationClassName = Configure::read('App.namespace') . '\Application';
 
-            return new CommandRunner(new $applicationClassName([CONFIG]));
+            return new CommandRunner(new $applicationClassName(CONFIG));
         }
 
         return new LegacyCommandRunner();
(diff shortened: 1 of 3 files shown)


src/TestSuite/ConsoleIntegrationTestCase.php
```diff
@@ -253,7 +253,7 @@ protected function _makeRunner()
         if ($this->_useCommandRunner) {
             $applicationClassName = Configure::read('App.namespace') . '\Application';
 
-            return new CommandRunner(new $applicationClassName([CONFIG]));
+            return new CommandRunner(new $applicationClassName(CONFIG));
```
dereuromark: It would be wise to split this into its own line, then you can add proper phpstan doc block annotation and then it will report the string vs array issue here IMO.

jeremyharris: I had done this originally, but PHPStorm didn't pick it up, so I didn't bother. Maybe phpstan will.

src/TestSuite/ConsoleIntegrationTestCase.php
```diff
@@ -252,8 +252,10 @@ protected function _makeRunner()
     {
         if ($this->_useCommandRunner) {
             $applicationClassName = Configure::read('App.namespace') . '\Application';
+            /** @var \Cake\Http\BaseApplication $applicationClass */
+            $applicationClass = new $applicationClassName([CONFIG]);
```
dereuromark: Did you on purpose revert the array to string in `new $applicationClassName(CONFIG)`?

jeremyharris: No I am just an idiot :facepalm:

PR #12042, opened 2018-05-03
4.x controller

ADmad:
Removed all deprecated code from `Controller` and components.


src/Controller/Controller.php
```diff
@@ -155,31 +151,6 @@ class Controller implements EventListenerInterface, EventDispatcherInterface
...
-     * ```
-     * public $components = ['RequestHandler', 'Acl'];
-     * ```
-     *
-     * @var array
-     * @link https://book.cakephp.org/3.0/en/controllers/components.html
-     *
-     * @deprecated 3.0.0 You should configure components in your Controller::initialize() method.
-     */
-    public $components = [];
```
markstory: For this and `$helpers` I wonder if we should trigger an error if we see them in use? We could add a `__set` that emits a warning telling people what they should be doing instead? Right now post upgrade their app would not work and figuring out why may take some time.

src/Controller/Controller.php
```diff
@@ -250,8 +213,6 @@ public function __construct(ServerRequest $request = null, Response $response =
 
         $this->initialize();
 
-        $this->_mergeControllerVars();
-        $this->_loadComponents();
```
markstory: We could put the new warning right here :smile:

markstory:
Refs #11934

ADmad:
@markstory Warning added as suggested.

tests/TestCase/View/ViewVarsTraitTest.php
```diff
@@ -186,12 +186,10 @@ public function testUptoDateViewVars()
      */
     public function testViewOptionsGetsToBuilder()
     {
-        $this->deprecated(function () {
-            $this->subject->passedArgs = 'test';
-            $this->subject->createView();
-            $result = $this->subject->viewBuilder()->getOptions();
-            $this->assertEmpty($result);
-        });
+        $this->subject->passedArgs = 'test';
```
ADmad: Don't see the point of setting an unused property. Either it should be replaced by equivalent new method or the whole test method should be removed.

saeideng: >test that removed deprecated passedArgs is no longer passed to the view

saeideng: >should be removed.

:+1:

markstory:
Nice work @ADmad :clap:

PR #12046, opened 2018-05-04
4.x core deprecations

ADmad:
Removed deprecated code from `Core` namespace.


markstory (approved):
Looks good, only phpstan errors left to fix.

ADmad:
My other PR #12042 removes `PaginatorComponent::config()` all together so that phpstan error will go away.

markstory:
Refs #11934

Issue #12038, opened 2018-05-02
MsSql query with tables having columns name with space cannot be selected


rudy1976s:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.6.2.
* Platform and Target: Nginx, MsSql 2016  Microsoft Dynamics Navision 2017

### What you did
I have created a model to connect to a MsSql existing table. This table has columns with space in name example : "No_ 2",
Quote identifiers are set to TRUE 

### What happened
The ORM result in error because is not able to recognize fields with space and return the following error : 

Error: SQLSTATE[42S22]: [Microsoft][ODBC Driver 13 for SQL Server][SQL Server]Invalid column name 'Items.No 2'.
Error could be seen here : http://sqlserver.acmecms.it/

### What you expected to happen
The query run without problem

markstory:
What does the generated query look like?

rudy1976s:
Hi mark
the query is like the following : SELECT TOP 10 [Items].[No_] AS [Items__No_], [Items.No_ 2] AS [Items__No_ 2] FROM [GLS72_ECOMMERCE$Item] [Items] . You . can check it up a t the following test site: sqlserver.acmecms.it

markstory:
The error you are getting looks like it is coming from the ODBC driver, which means the same query executed with the basic `PDO` class would likely fail in a similar way. Are you able to query this table/field with the basic PDO objects?

rudy1976s:
Hi Mark
you can check a raw conection to PDO here and it works fine

http://sqlserver.acmecms.it

dereuromark:
Why do you need spaces filled field names? Cant you refactor your db? there are many issues and regex on the way that will likely also fail. Also mapping to entity fields and properties is a mess this way.

rudy1976s:
Hi mark , no we cannot refactor because we are connecting to Microsoft Navision and the table we are going to work on  are "core" and cannot be refactored.

markstory:
@rudy1976s What does the code that uses `PDO` directly look like? I'm mostly interested in the SQL query being executed.

rudy1976s:
Hi mark do you mean the raw code used or the one used with cake ? 

The PDO direct is like the following:

`
try
{
    $pdo_object = new PDO("sqlsrv:Server=SERVER,PORT;Database=DATABASE", "USER", "PASSWORD");
    
[...]

PR #12049, opened 2018-05-04
Add identifier quoting for fields with spaces.

markstory:
There is one caveat in that fields containing `AS` will still be misquoted.

Refs #12038


dereuromark:
Sounds still like an improvement here :)

lorenzo (commented):
Nice!

Issue #12050, opened 2018-05-04
Issue with RouteBuilder when parsing compact routes.


devcorrelator:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.6.2
* Platform and Target: Not Applicable.

### What you did
Created Routes with the new Compact Routes introduced in 3.6:
This is happening with all defined compacts routes, but an example:
config/routes.php
$routes->connect('/*', 'Pages.Pages::display');

### What happened
Each defined compact route is throwing an offset error:
Notice (8): Undefined offset: 2 [CORE/src/Routing/RouteBuilder.php, line 762]

### What you expected to happen
Proper parsed

P.S. Remember, an issue is not the place to ask questions. You can use [Stack Overflow](https://stackoverflow.com/questions/tagged/cakephp)
for that or join the #cakephp channel on irc.freenode.net, where we will be more
than happy to help answer your questions.

Before you open an issue, please check if a similar issue already exists or has been closed before.

dereuromark:
PR https://github.com/cakephp/cakephp/pull/12051 open

PR #12051, opened 2018-05-04
Fix for issue: 12050

devcorrelator:
Resolves https://github.com/cakephp/cakephp/issues/12050


src/Routing/RouteBuilder.php
```diff
@@ -747,28 +747,27 @@ protected static function parseDefaults($defaults)
         if (preg_match($regex, $defaults, $matches)) {
             unset($matches[0]);
             $matches = array_filter($matches, function ($value) {
-                return $value !== '' && $value !== '::';
+                return $value !== '::';
```
dereuromark: Why removing empty string check here? That was on purpose afaik.

devcorrelator: That's what was removing $matches[2]. With it there, it could mess with the switch as it relies on the count of the array verse specifically looking to see if like $matches[4] is set.

devcorrelator: Plus, the switch is referencing specific indexes within the $marches array, and only returning specific indexes, so if $matches[11] existed because this was taken it out, ultimately it wouldn't matter (other then maybe garbage collection if $matches is huge for some reason).

devcorrelator: Also, I guess it should return an empty string instead of null? 
See failed tests:
https://travis-ci.org/cakephp/cakephp/jobs/375121707

dereuromark: This is a filter method on the array, any empty value is being stripped, same as :: which is similar empty value.

markstory:
These changes cause several tests to fail. We'll need to find another solution that doesn't break existing behavior.

markstory:
I've pushed up a change that solves the issue you reported and solves failing tests.

PR #12052, opened 2018-05-05
remove passedArgs

saeideng:
Refs : https://github.com/cakephp/cakephp/issues/11934


ADmad:
PR #12042 already takes care of this.

PR #12056, opened 2018-05-07
removed duplicate 'className' check

### LINKED ISSUE #12055: Duplicate 'className' check in TableLocator
This is a (multiple allowed):

* [x] bug
* [ ] enhancement

* CakePHP Version: 3.6.2
* Platform and Target: not relevant

### What you did
I'd like to extend table locator (\Cake\ORM\Locator\TableLocator). I'd prefer to override only _getClassName (which I believe exists only for this reason), but checking of 'className' options is duplicate, so you can never know, if user provided 'className'.

### What happened
see above

see above

segy:
fixes #12055

### DIFF
--- a/src/ORM/Locator/TableLocator.php
+++ b/src/ORM/Locator/TableLocator.php
@@ -195,8 +195,8 @@ class TableLocator implements LocatorInterface
         if ($className) {
             $options['className'] = $className;
         } else {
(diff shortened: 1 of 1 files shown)


markstory:
Your changes make a number of tests fail. Looks like the className check is not duplicate/redundant.

PR #12059, opened 2018-05-08
remove deprecated Network classes

saeideng:
clean up  part of the Network package 

Refs : https://github.com/cakephp/cakephp/issues/11934

### DIFF
--- a/src/Routing/Router.php
+++ b/src/Routing/Router.php
@@ -825,7 +825,7 @@ class Router
      * are used for CakePHP internals and should not normally be part of an output URL.
      *
      * @param \Cake\Http\ServerRequest|array $params The params array or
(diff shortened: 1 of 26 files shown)


src/Routing/Router.php
```diff
@@ -825,7 +825,7 @@ public static function reverseToArray($params)
      * are used for CakePHP internals and should not normally be part of an output URL.
      *
      * @param \Cake\Http\ServerRequest|array $params The params array or
-     *     Cake\Network\Request object that needs to be reversed.
+     *     Cake\Http\ServerRequest object that needs to be reversed.
```
dereuromark: We should backport this already to 3.6 IMO.

saeideng: done.

PR #12063, opened 2018-05-09
Fatal errors are not properly handled/logged.

dereuromark:
Fatal errors on CLI (debug on or off) are not logged at all.
This is very dangerous on production if you do not get notified about it (inside cronjobs).
In our specific case it was using a require-dev class and that one was then gone live.

Demo: `$x = new Demo();` with `class Demo extends Foo` and Foo does not exist

I found out that none of the configured handlers are invoked in that case

        set_error_handler([$this, 'handleError'], $level);
        set_exception_handler([$this, 'wrapAndHandleException']);
        register_shutdown_function(function () {}

Only the register_shutdown_function() which is then return early here:

    if (PHP_SAPI === 'cli' || PHP_SAPI === 'phpdbg') {
        return;
    }

This is very bad.

I added a handled propery - if there are better ways I am all ears.

Now it at least logs it again

    Fatal Error (1): Class 'App\Foo' not found in ...

This can only be fixed/relevant for PHP7 of course. We are running 7.1 and 7.2 here.

### DIFF
--- a/src/Error/BaseErrorHandler.php
+++ b/src/Error/BaseErrorHandler.php
@@ -38,6 +38,11 @@ abstract class BaseErrorHandler
     protected $_options = [];
 
     /**
+     * @var bool
+     */
+    protected $_handled = false;
+
+    /**
      * Display an error message in an environment specific way.
      *
      * Subclasses should implement this method to display the error as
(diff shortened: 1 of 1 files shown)


src/Error/BaseErrorHandler.php
```diff
@@ -75,7 +80,7 @@ public function register()
         set_error_handler([$this, 'handleError'], $level);
         set_exception_handler([$this, 'wrapAndHandleException']);
         register_shutdown_function(function () {
-            if (PHP_SAPI === 'cli' || PHP_SAPI === 'phpdbg') {
+            if ((PHP_SAPI === 'cli' || PHP_SAPI === 'phpdbg') && $this->_handled) {
```
markstory: What happens if we remove the skip entirely?

dereuromark: The cli output doubles somehow as the (non?)buffered fatal error output is printed and then the display part again.
Note: Could xdebug also be playing a role here? In the duplication?

dereuromark:
Can we find a way here to continue in order to track/log these fatal errors, please? :)

lorenzo (approved):
This looks good to me

PR #12065, opened 2018-05-10
Fixed return type (array) #11795

odan:
This change ensures that `fetchAll` always returns an array.

### DIFF
--- a/src/Database/Statement/BufferedStatement.php
+++ b/src/Database/Statement/BufferedStatement.php
@@ -12,7 +12,6 @@
  * @since         3.0.0
  * @license       https://opensource.org/licenses/mit-license.php MIT License
  */
-
 namespace Cake\Database\Statement;
 
 /**
--- a/src/Database/Statement/BufferedStatement.php
+++ b/src/Database/Statement/BufferedStatement.php
@@ -113,7 +113,7 @@ class BufferedStatement extends StatementDecorator
     {
         $result = $this->fetch(static::FETCH_TYPE_ASSOC);
 
-        return $result ? $result : [];
+        return $result ?: [];
     }
 
     /**
--- a/src/Database/Statement/BufferedStatement.php
+++ b/src/Database/Statement/BufferedStatement.php
@@ -116,7 +116,7 @@ class BufferedStatement extends StatementDecorator
             return $result;
         }
 
-        return array();
+        return [];
     }
 
     /**
--- a/src/Database/Statement/BufferedStatement.php
+++ b/src/Database/Statement/BufferedStatement.php
@@ -112,11 +112,9 @@ class BufferedStatement extends StatementDecorator
      */
     public function fetchAssoc()
     {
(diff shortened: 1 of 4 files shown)


src/Database/Statement/BufferedStatement.php
```diff
@@ -111,7 +112,11 @@ public function fetch($type = parent::FETCH_TYPE_NUM)
      */
     public function fetchAssoc()
     {
-        return $this->fetch(static::FETCH_TYPE_ASSOC);
+        if ($result = $this->fetch(static::FETCH_TYPE_ASSOC)) {
```
dereuromark: you shouldnt assign inside condition.

dereuromark (changes requested):

src/Database/Statement/BufferedStatement.php
odan: Done

src/Database/Statement/BufferedStatement.php
```diff
@@ -111,7 +112,9 @@ public function fetch($type = parent::FETCH_TYPE_NUM)
      */
     public function fetchAssoc()
     {
-        return $this->fetch(static::FETCH_TYPE_ASSOC);
+        $result = $this->fetch(static::FETCH_TYPE_ASSOC);
+
```
dereuromark: You can use  ?:

src/Database/Statement/StatementDecorator.php
```diff
@@ -221,11 +222,13 @@ public function fetch($type = self::FETCH_TYPE_NUM)
      * Returns the next row in a result set as an associative array. Calling this function is the same as calling
      * $statement->fetch(StatementDecorator::FETCH_TYPE_ASSOC). If no results are found false is returned.
      *
-     * @return array|false Result array containing columns and values an an associative array or false if no results
+     * @return array Result array containing columns and values an an associative array or an empty array if no results
```
markstory: I wonder how likely it is that someone has `=== false` in their code. It might be wise to move this to 3.next for that reason.

src/Database/Statement/BufferedStatement.php
```diff
@@ -12,6 +12,7 @@
  * @since         3.0.0
  * @license       https://opensource.org/licenses/mit-license.php MIT License
  */
+
```
markstory: bonus line.

src/Database/Statement/StatementDecorator.php
```diff
@@ -12,6 +12,7 @@
  * @since         3.0.0
  * @license       https://opensource.org/licenses/mit-license.php MIT License
  */
+
```
markstory: Here too.

src/Database/Statement/BufferedStatement.php
dereuromark: No, I mean the short ternary form: `return $result ?: [];`. Cleaner since you do not repeat it.

PR #12071, opened 2018-05-11
Remove @inheritDoc to correctly work with phpstan

segy:
Replaces @inheritDoc with full parent comment since phpstan does not allow to use @inheritDoc and override @param definitions in the same time.


markstory:
Thank you 👍

PR #12072, opened 2018-05-11
Testsuite improvements

jeremyharris:
This PR does two things:

- moves existing integration test case assertions into PHPUnit constraints
- moves integration test case logic into traits

### Constraints

This is mostly a cosmetic change, but it allows us to provide better assertion errors than before. For example, `assertResponseContains` would output the entire response body upon failure (useful only for small response bodies), and `assertRedirect` which would output an error like "Failed asserting that /path is empty".

### Traits

This allows devs to create a base test case that contains logic and individual test cases can then implement the appropriate integration traits.


jeremyharris:
The failures are related to https://github.com/cakephp/cakephp/pull/12041 which has been merged into master (3.next is a bit behind).

src/TestSuite/ConsoleIntegrationTestCase.php
```diff
@@ -13,307 +13,14 @@
...
-use Cake\Console\ConsoleInput;
-use Cake\Console\ConsoleIo;
-use Cake\Core\Configure;
-use Cake\TestSuite\Stub\ConsoleOutput;
-
 /**
  * A test case class intended to make integration tests of cake console commands
  * easier.
+ *
+ * @deprecated use Cake\TestSuite\ConsoleIntegrationTestTrait instead
```
markstory: I guess this makes it easier for application developers to have their own test case hierarchy.

src/TestSuite/Constraint/Console/ContentsBase.php
```diff
@@ -0,0 +1,55 @@
...
+ * @since         3.7.0
+ * @license       http://www.opensource.org/licenses/mit-license.php MIT License
+ */
+namespace Cake\TestSuite\Constraint\Console;
+
+if (class_exists('PHPUnit_Runner_Version', false)
+    && !class_exists('PHPUnit\Framework\Constraint\Constraint', false)
+) {
+    class_alias('PHPUnit_Framework_Constraint', 'PHPUnit\Framework\Constraint\Constraint');
+}
```
markstory: Could this go in with the other phpunit aliases?

src/TestSuite/ConsoleIntegrationTestCase.php
jeremyharris: Yes. Personally I have a `TestCase` base class that extended IntegrationTestCase so I had that functionality when I needed it. However, when I want a ConsoleIntegrationTest, I wasn't able to use my base class.

src/TestSuite/Constraint/Console/ContentsBase.php
jeremyharris: I think so. tbh I copied it from the EventFired constraints I wrote a while back. I don't remember if I tried putting it with the other aliases but now seems as good as time as ever to try.

Issue #12075, opened 2018-05-12
problem with _base and Application


saeideng:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.6

add an IntegrationTestCase
```php
     public function testUrl()
     {
//$this->useHttpServer(true);
        debug( \Cake\Routing\Router::url('/articles')); //debug 1
         $url = '/articles';
         $this->get($url);
         $this->assertResponseSuccess();
         $this->assertResponseEquals('');
         debug( \Cake\Routing\Router::url('/articles')); //debug 2
     }
```
and a controller
```php
class ArticlesController extends Controller
{

  public function index(){
      debug( \Cake\Routing\Router::url('/articles')); // debug 3
  }
}
```

result 
```php
//debug1
'/articles'
//debug2
'/articles'
//debug3
'/articles'
```

now add an Application class and see result 
```php
class Application extends BaseApplication
{

    public function middleware($middlewareQueue)
    {
        $middlewareQueue
            // Handle plugin/theme assets like CakePHP normally does.
            ->add(AssetMiddleware::class)

            // Add routing middleware.
            ->add(new RoutingMiddleware($this));

        return $middlewareQueue;
    }
}
```
now result

```php
//debug1
'/articles'
//debug2
'/vendor/bin/articles'
//debug3
'/vendor/bin/articles'
```

configs
```php
Cake\Core\Configure::write('App', [
  'namespace' => 'CrudJsonApi\Test\App',
  'paths' => ['templates'=>[APP.'Template'.DS]],
  'base' => false,
  'baseUrl' => false,
  'dir' => 'src',
  'webroot' => 'webroot',
  'www_root' => APP . 'webroot',
  'fullBaseUrl' => 'http://localhost',
]);
Cake\Core\Configure::write('debug', true);
```

saeideng:
fixing `base` in Application can be solve the problem
https://github.com/FriendsOfCake/crud-json-api/commit/f2df79391fad159e0ac12e0cd64001bd3d217213

but I want to know can we solve problem in core?

markstory:
You need to define $_SERVER[PHP_SELF] in your testsuite bootstrap to fix this I think.

markstory:
We could also solve this in the integration test case by setting the environment data up automatically. That might be a better solution for end users.

markstory:
Pull request open.

PR #12076, opened 2018-05-12
Set PHP_SELF in the integration test case.

markstory:
This removes the need for users to set `$_SERVER['PHP_SELF']` in their tests/bootstrap.php. While this is something included in the app template existing apps don't have it and it leads to hard to debug failures.

Refs #12075


dereuromark:
Thanks :)

PR #12078, opened 2018-05-13
fix deprecationWarning message

saeideng:


src/ORM/Table.php
```diff
@@ -960,13 +960,13 @@ public function hasBehavior($name)
...
      *
-     * @deprecated 3.6.0 Use getAssociation() and Table::hasAssocation() instead.
+     * @deprecated 3.6.0 Use getAssociation() and Table::hasAssociation() instead.
      * @param string $name the alias used for the association.
      * @return \Cake\ORM\Association|null Either the association or null.
      */
     public function association($name)
     {
-        deprecationWarning('Use Table::getAssociation() and Table::hasAssocation() instead.');
+        deprecationWarning('Use Table::getAssociation() and Table::hasAssociation() instead.');
```
dereuromark: I also wonder if we should include the version as well.
Often times it is hard to manually track down what version you can use, when you replace them in plugins etc.

markstory: What would the updated error message look like?

dereuromark: Identical to the annotation IMO:

    deprecationWarning('3.6.0: Use Table::getAssociation() and Table::hasAssociation() instead.');

saeideng: my suggest : 
adding `$since` 
```php
deprecationWarning($message, $since =null, $stackFrame = 1){

$message = sprintf(
                $since :  '%s - %s, line: %s' . "\n" .

}

PR #12088, opened 2018-05-16
4.x - Remove deprecated features in Network and most of Routing

markstory:
I didn't get all the deprecated features in Routing as I didn't want to make this set of changes even larger. I'll clean up automatic route file loading in a separate pull request.

Refs #11934


markstory:
Not sure there is much to be done about the drop in coverage as it seems like the deprecated code was well covered :cry:

Issue #12077, opened 2018-05-13
Add SQL RAND() function


dmromanov:
This is a (multiple allowed):

* [ ] bug
* [x] enhancement
* [ ] feature-discussion (RFC)

It would be nice, if CakePHP's ORM would have SQL RAND() function built in.
This function is pretty useful for selecting a random columns `ORDER BY RAND() LIMIT 1`.

Different SQL dialects vary the function's name much. E.g. MySQL use `RAND()` while Sqlite use `RANDOM()`, so having the function built-in would ease development and testing applications that rely on the SQL function.

lorenzo:
Would you like to submit the code for this?

dmromanov:
@lorenzo I'll do.

lorenzo:
thanks!

PR #12090, opened 2018-05-16
Added SQL RAND() function.

dmromanov:
Added an SQL RAND() function to ORM.

Because Sqlite is the only DBMS with RANDOM() function returning values not between 0 <= x < 1, I've added some normalization to make the function consistent.

Seeding is not supported.

Refs #12077


saeideng:
can we have  `$query->func()->rand($seed)` !?

dmromanov:
@saeideng Well, practically, we can but, I have no idea on how to test it. Different DBMS' use different values for seeding. And Sqlite can not seed at all.

src/Database/FunctionsBuilder.php
```diff
@@ -60,6 +60,16 @@ protected function _literalArgumentFunction($name, $expression, $types = [], $re
         return $this->_build($name, $expression, $types, $return);
     }
 
+    /**
+     * Returns a FunctionExpression representing a call to SQL RAND function.
+     *
+     * @return \Cake\Database\Expression\FunctionExpression
+     */
+    public function rand()
```
markstory: Should the function name be `random()`?

dmromanov: I'd stick to the short version to match other functions like "avg" and "sum".

markstory: Fair enough.

markstory:
Moved to 3.7 as this adds new methods.

tests/bootstrap.php
```diff
@@ -139,6 +139,3 @@
 // does not allow the sessionid to be set after stdout
 // has been written to.
 session_id('cli');
-
-// Fix multiple http/server requests in a single test method.
-$_SERVER['PHP_SELF'] = '/';
```
dereuromark: Why was this removed?

saeideng: removed in other PR

dmromanov: @dereuromark I did not do it. Must be one of thouse side effects of rebasing a branch in Git.

src/Database/FunctionsBuilder.php
dmromanov: `rand()` comes from SQL `RAND()` function as in MySQL and SqlServer.

PR #12092, opened 2018-05-17
IBAN validation + test added

philsweb:


dereuromark (changes requested):

src/Validation/Validation.php
```diff
@@ -181,25 +187,27 @@ public static function blank($check)
...
+     * @param string|null  $regex A custom regex can also be passed, this will be used instead of the defined regex
+     *                            values
+     *
      * @return bool Success
      * @see \Cake\Validation\Validation::luhn()
      */
     public static function cc($check, $type = 'fast', $deep = false, $regex = null)
     {
-        if (!is_scalar($check)) {
+        if (! is_scalar($check)) {
```
dereuromark: Why are you changing this to an invalid CS? There must be no space after casts.
Also, please only change exactly the lines you want to change, no other noise changes.

src/Validation/Validation.php
```diff
@@ -82,8 +83,8 @@ class Validation
      * @var array
      */
     protected static $_pattern = [
-        'hostname' => '(?:[_\p{L}0-9][-_\p{L}0-9]*\.)*(?:[\p{L}0-9][-\p{L}0-9]{0,62})\.(?:(?:[a-z]{2}\.)?[a-z]{2,})',
-        'latitude' => '[-+]?([1-8]?\d(\.\d+)?|90(\.0+)?)',
+        'hostname'  => '(?:[_\p{L}0-9][-_\p{L}0-9]*\.)*(?:[\p{L}0-9][-\p{L}0-9]{0,62})\.(?:(?:[a-z]{2}\.)?[a-z]{2,})',
```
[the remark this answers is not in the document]
dereuromark: Inline alignment is evil. We don't do that. :)

src/Validation/Validation.php
philsweb: Sure. Sorry, my bad

src/Validation/Validation.php
```diff
@@ -1559,6 +1559,41 @@ public static function hexColor($check)
...
+     * Check that the input value is a valid IBAN number
+     *
+     * @param string $check The value to check
+     * @return bool Success
+     */
+    public static function iban($check)
+    {
+        $iban = str_replace(' ', '', $check);
+
+        if (preg_match('/^[A-Z]{2}[0-9]{2}[A-Z0-9]{1,30}$/i', $iban)) {
```
dereuromark: If you invert you can bail early here.

src/Validation/Validation.php
```diff
@@ -1552,19 +1620,58 @@ public static function isScalar($value)
...
+
+            $checksum     = intval(substr($numstr, 0, 1));
+            $numstrLength = strlen($numstr);
+            for ($pos = 1; $pos < $numstrLength; $pos++) {
+                $checksum *= 10;
+                $checksum += intval(substr($numstr, $pos, 1));
+                $checksum %= 97;
+            }
+
+            return ((98 - $checksum) == $check);
```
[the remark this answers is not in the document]
dereuromark: Can we please use strict `===`?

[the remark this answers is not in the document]
philsweb: Sure

dereuromark:
Please squash all commits into a single one when you are done. There was a lot of noise changes that otherwise make 3.next/4.x etc merges more difficult and conflicts more likely where not necessary.

src/Validation/Validation.php
philsweb: Sorry, not clear

dereuromark:
Small side note for you:
You don't have your git set up correctly (email etc), so your name here is not clickable.

src/Validation/Validation.php
dereuromark:
```php
if (!preg_match)) {
    return false;
}

... //continue without extra indentation level
```

See CakePHP coding guidelines.

philsweb: Thanks!

dereuromark: Sure you inverted properly?

philsweb:
@dereuromark Sorry, is it OK if  I open a new Pull Request right now? Cannot succeed in proper squashing.

dereuromark (approved):

dereuromark:
It is fine, we can sqash on merge afaik.

philsweb:
@dereuromark Thanks for your time, very first experience of contributing

src/Validation/Validation.php
```diff
@@ -1559,6 +1559,40 @@ public static function hexColor($check)
         return static::_check($check, '/^#[0-9a-f]{6}$/iD');
     }
 
+    /**
+     * Check that the input value is a valid IBAN number
```
inoas: one number too much:

IBAN number = "International Bank Account Number number"

src/Validation/Validation.php
```diff
@@ -1559,6 +1559,40 @@ public static function hexColor($check)
...
+     *
+     * @param string $check The value to check
+     * @return bool Success
+     */
+    public static function iban($check)
```
inoas: https://www.iban.com/structure.html

This validation method will pass IBANs that have wrong country codes or country code missmatches.
I am not saying that this should be fixed but the doc header should quote what the method does and what not.
I'd also say it should be `ibanSyntax($value)`.

The method checks for the country code to exist, the checksum to exist and then checks the body with Luhn algorithm

tests/TestCase/Validation/ValidationTest.php
```diff
@@ -3172,4 +3172,19 @@ public function testHexColor()
         $this->assertFalse(Validation::hexColor('#fff'));
         $this->assertFalse(Validation::hexColor('ffffff'));
     }
+
+    /**
+     * Test IBAN
+     */
+    public function testIban()
+    {
+        $this->assertTrue(Validation::iban('AD1200012030200359100100'));
```
inoas: Could this include all of these:
https://www.iban.com/structure.html

listed as IBAN SEPA and IBAN countries?

inoas (changes requested):

src/Validation/Validation.php
```diff
@@ -1559,6 +1559,41 @@ public static function hexColor($check)
...
 
+    /**
+     * Check that the input value is a valid iban number
+     *
+     * @param string $check The value to check
+     * @return bool Success
+     */
+    public static function iban($check)
+    {
+        $iban = strtoupper(str_replace(' ', '', $check));
```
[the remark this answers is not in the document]
inoas: AFAIK IBANs are always uppercase and expected to be so. I'd not allow lowercase, instead strtoupper should happen on the data supplied by - say - the user.

philsweb:
@inoas Will deal with that in a while, thanks

src/Validation/Validation.php
dereuromark: iban() should suffice, the syntax sounds more like spacing check (or length only) or sth.

inoas: it does nothing more than that. it does not validate an iban(). If it were it would be required to check the country code against the length, and possibly even the national bank codes against a database https://en.wikipedia.org/wiki/International_Bank_Account_Number#IBAN_formats_by_country

It is a pure stateless syntax check with. Validating an IBAN *for real* requires much more work. Thus `ibanSyntax()` hints you very clearly that you are only checking the form (uppercase, no whitespaces, max length, country code exists at right spot, checksum exists at right spot, body matches against checksum via Luhn)

dereuromark: What about creditCard, ip ones, url etc?
They all also only do Syntax check, that is already very obvious here from the scope.

tests/TestCase/Validation/ValidationTest.php
philsweb: Done

src/Validation/Validation.php
```diff
@@ -1559,6 +1559,41 @@ public static function hexColor($check)
         return static::_check($check, '/^#[0-9a-f]{6}$/iD');
     }
 
+    /**
+     * Check that the input value has a valid International Bank Account Number IBAN syntax
+     * Requirements are uppercase, no whitespaces, max length 34, country code and checksum exist at right spots,
+     * body matches against checksum via Luhn algorithm
```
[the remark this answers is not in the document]
inoas: :+1: for single source of truth/error.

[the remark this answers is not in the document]
ADmad: There's already a `Validation::luhn()` :slightly_smiling_face:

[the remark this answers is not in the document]
makallio85: I guess the algorithms are different

markstory:
Thank you @philsweb :clap:

PR #12097, opened 2018-05-17
Sort route defaults consistently in the routes cli

AD7six:
In a minor refactor of some routes definitions I noticed there was no functional change - but the `bin/cake routes` output did change the order of default parameters.

This just makes it consistent.


lorenzo (approved):
looks reasonable

markstory:
Tests need to be updated. But otherwise this looks good.

Issue #12094, opened 2018-05-17
Add actual methods to interfaces instead of `@method` tags.


ADmad:
Tracking issue for interfaces that should be updated in 4.x

* [ ] ConnectionInterface
* [x] EntityInterface
* [x] InvalidPropertyInterface
* [ ] QueryInterface
* [x] RepositoryInterface
* [x] EventDispatcherInterface
* [x] LocatorInterface
* [x] ContextInterface

dereuromark:
Are the remaining ones still doable? Or do we have to keep them out of the interface(s)?

ADmad:
I am pretty sure all possible changes are done. Remaining ones we'll just have to keep.

ADmad:
This reminds me `Entity` has couple of method (maybe more) `getOriginal()`, `getOriginalValues()` values which aren't present in `EntityInterface`.

PR #12103, opened 2018-05-19
4.x orm deprecations

ADmad:
I have left the deprecated methods in `TableRegistry`. Just wanted to confirm we want to get rid of them in 4.x.

Refs #11934


src/ORM/Query.php
```diff
@@ -455,13 +429,8 @@ public function contain($associations = null, $override = false)
...
 
-        if ($associations === null) {
-            deprecationWarning(
-                'Using Query::contain() as getter is deprecated. ' .
-                'Use getContain() instead.'
-            );
-
-            return $loader->getContain();
+        if ($associations === null && $override === false) {
+            throw new InvalidArgumentException('$associations can be null only $override is true.');
```
saeideng: but `associations === null ` deprecated 
and we have `clearContain()` method for clearing contain 
`->contain( null, true)` is pointless IMO

ADmad: I guess we can remove this check all together and one can just use `contain([], true)` to clear/reset.

src/ORM/Locator/LocatorInterface.php
```diff
@@ -24,17 +24,6 @@
...
-
-    /**
-     * Stores a list of options to be used when instantiating an object
-     * with a matching alias.
-     *
-     * @param string|null $alias Name of the alias
-     * @param array|null $options list of options for the alias
-     * @return array The config data.
-     */
-    public function config($alias = null, $options = null);
```
markstory: Is this class on your list of classes that need `@method` turned into real methods?

ADmad: Oui monsieur https://github.com/cakephp/cakephp/issues/12094

ADmad:
@markstory @lorenzo  Should I remove deprecated methods from `TableRegistry` too?

PR #12111, opened 2018-05-21
4.x Add missing methods to InvalidPropertyInterface.

ADmad:
Refs #12094.

### DIFF
--- a/src/Datasource/InvalidPropertyInterface.php
+++ b/src/Datasource/InvalidPropertyInterface.php
@@ -17,12 +17,44 @@ namespace Cake\Datasource;
 /**
  * Describes the methods that any class representing a data storage should
  * comply with.
- *
- * @method array getInvalid()
- * @method mixed getInvalidField($field)
- * @method $this setInvalid($field, $value = null, $overwrite = false)
- * @method $this setInvalidField($field, $value = null, $overwrite = false)
  */
 interface InvalidPropertyInterface
 {
+
+    /**
+     * Get a list of invalid fields and their data for errors upon validation/patching
+     *
+     * @return array
+     */
+    public function getInvalid();
+
+    /**
+     * Set fields as invalid and not patchable into the entity.
+     *
+     * This is useful for batch operations when one needs to get the original value for an error message after patching.
+     * This value could not be patched into the entity and is simply copied into the _invalid property for debugging purposes
+     * or to be able to log it away.
+     *
+     * @param array $fields The values to set.
+     * @param bool $overwrite Whether or not to overwrite pre-existing values for $field.
+     * @return $this
+     */
+    public function setInvalid(array $fields, $overwrite = false);
+
+    /**
+     * Get a single value of an invalid field. Returns null if not set.
+     *
+     * @param string $field The name of the field.
+     * @return mixed
+     */
+    public function getInvalidField($field);
+
+    /**
+     * Sets a field as invalid and not patchable into the entity.
+     *
+     * @param string $field The value to set.
+     * @param mixed $value The invalid value to be set for $field.
+     * @return $this
+     */
+    public function setInvalidField($field, $value);
 }


markstory (approved):
Looking good!

src/Datasource/ConnectionInterface.php
```diff
@@ -79,4 +72,68 @@ public function disableConstraints(callable $operation);
...
+     *
+     * @param \Cake\Database\Schema\Collection $collection The schema collection object
+     * @return $this
+     */
+    public function setSchemaCollection(Collection $collection);
+
+    /**
+     * Create a new Query instance for this connection.
+     *
+     * @return \Cake\Database\Query
```
lorenzo: I don’t think this belongs here.

lorenzo (commented):
I really don’t think anything that returns or takes something from the Darabase package belongs in this interface

src/Datasource/ConnectionInterface.php
ADmad: You mean the `@return` type? I copied it from `Connection::newQuery()`. What should the type be instead?

lorenzo: Either we don’t include these methods in the interface or we abstract them so they are in Datasource.

saeideng: `newQuery` is unnecessary in interface 
because we can do like / we have other option to create new query 
```php
$driver = new Mysql([
	'database' => '',
	'username' => '',
	'password' => ''
]);
$connection = new Connection([
	'driver' => $driver
]);
$query = new Query($connection);
```

ADmad: I have removed the method from interface, let's see whether phpstan complains.

ADmad: It does: https://travis-ci.org/cakephp/cakephp/jobs/382013768#L681

Same will happen if we remove the other new methods than have been added which return objects for Database namespace. So if we don't want them in the interface than the relevant typehints which use `ConnectionInterface` will have to be changed to concrete class `Connection` instead.

lorenzo:
@ADmad I think we are trying too hard to generalise something that is not ready to be generalised. By the looks of it, most of the new methods are only used in the `TestFixture`, can you verify that?

If that is the case, we can leave those methods out and change the implementation of TestFixture so that it can safely depend on a Database implementation or a generic implementation.

src/Datasource/ConnectionInterface.php
```diff
@@ -79,4 +72,61 @@ public function disableConstraints(callable $operation);
...
      */
     public function logQueries($enable = null);
+
+    /**
+     * Sets a logger
+     *
+     * @param \Cake\Database\Log\QueryLogger $logger Logger object
+     * @return $this
+     */
+    public function setLogger($logger);
```
dereuromark: Would be nice to have at least a basic interface here to contract against instead of using a concrete and specific class.

markstory:
> By the looks of it, most of the new methods are only used in the TestFixture, can you verify that?
>
> If that is the case, we can leave those methods out and change the implementation of TestFixture so that it can safely depend on a Database implementation or a generic implementation.

The TestFixture interface is also used by the elasticsearch plugin which also uses the `ConnectionInterface`.

lorenzo:
@markstory which has been a constant source of pain each time one of the methods is typehinted with an interface since that makes the plugin introduce a big number of new methods the plugin does not need to actually implement

markstory:
@lorenzo Perhaps we're missing a smaller interface that only defines the methods both fixture implementations actually need. Or if there is no common ground the typehints should be removed.

lorenzo:
I’ll check what can be abstracted away

ADmad:
PR only contains updates to `InvalidPropertyInterface` now. Should be good to merge.

PR #12112, opened 2018-05-21
Include missing exception

davidyell:
The controllers loadComponent has an exception docblock, but the method which causes that does not. 

This pull request addresses that inconsistency.


saeideng:
>The controllers loadComponent has an exception docblock

it can be `Cake\Controller\Exception\MissingComponentException`
https://github.com/cakephp/cakephp/blob/614f906e001f5e98174d1b273119931859bd7f39/src/Controller/ComponentRegistry.php#L96-L104

Issue #11997, opened 2018-04-25
Enhancement: Implement "strategies" for TranslateBehavior


ADmad:
Current `TranslateBehavior` uses EAV data model to store the translations. Another better/more convenient approach is using "shadow tables" as implemented in [ad7six/shadow-translate](https://github.com/AD7six/cakephp-shadow-translate). So it would be nice if the core `TranslateBehavior` had support for both data models. 

Implementations for both strategies already exist so it's a matter of modify `TranslateBehavior` to allow user to select preferred strategy. To achieve that we could introduce strategy classes for the translate behavior for handling each approach and one can configure the behavior to use either.

lorenzo:
I like this!

One thing I'd love to do is to abstract the EAV logic in the behaviour to something reusable, sounds like moving the code to a strategy may be able to help us do that.

Would it make sense to have shadow-translate as an official strategy?

ADmad:
> Would it make sense to have shadow-translate as an official strategy?

Absolutely, having it as an official strategy in core itself will give it greater visibility which I am hoping for.

rchavik:
Can we have a way to auto create the shadow tables?

PR #12113, opened 2018-05-21
4.x TranslateBehavior strategies

ADmad:
Refs #11997.


src/ORM/Behavior/TranslateBehavior.php
```diff
@@ -120,254 +92,54 @@ public function __construct(Table $table, array $config = [])
...
      *
-     * @param \Cake\Event\Event $event The beforeSave event that was fired
-     * @param \Cake\Datasource\EntityInterface $entity The entity that is going to be saved
-     * @param \ArrayObject $options the options passed to the save method
-     * @return void
+     * @param object $strategy Strategy class instance.
+     * @return $this
      */
-    public function beforeSave(Event $event, EntityInterface $entity, ArrayObject $options)
+    public function setStrategy($strategy)
```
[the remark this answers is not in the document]
ADmad: Yes still is temporary. I do plan to add an interface once both strategies are added and common methods identified.

ADmad:
This is ready for a review.

Just need helper sorting out this error https://travis-ci.org/cakephp/cakephp/jobs/383864122#L712. There are no errors if you run `TranslateBehaviorShadowTableTest` or `TranslateBehaviorTest` alone.

src/ORM/Behavior/Translate/ShadowTableStrategy.php
```diff
@@ -0,0 +1,591 @@
...
+        }
+
+        if ($noFields) {
+            return;
+        }
+
+        $where = compact('id', 'locale');
+
+        $translation = $this->translationTable->find()
+            ->select(array_merge(['id', 'locale'], $fields))
```
lorenzo: I think for safety sake, these fields should be aliased

src/ORM/Behavior/Translate/ShadowTableStrategy.php
```diff
@@ -0,0 +1,591 @@
...
+            ->where($where)
+            ->enableBufferedResults(false)
+            ->first();
+
+        if ($translation) {
+            foreach ($fields as $field) {
+                $translation->set($field, $values[$field]);
```
lorenzo: isn't there a method to set by just passing the array?

src/ORM/Behavior/Translate/ShadowTableStrategy.php
```diff
@@ -0,0 +1,591 @@
...
+            }
+
+            return;
+        }
+
+        if ($noFields) {
+            return;
+        }
+
+        $where = compact('id', 'locale');
```
lorenzo: Can you not use compact, please?

src/ORM/Behavior/Translate/ShadowTableStrategy.php
```diff
@@ -0,0 +1,591 @@
...
+     * into each entity under the `_translations` key
+     *
+     * @param \Cake\Datasource\ResultSetInterface $results Results to modify.
+     * @return \Cake\Collection\CollectionInterface
+     */
+    public function groupTranslations($results): CollectionInterface
+    {
+        return $results->map(function ($row) {
+            $translations = (array)$row['_i18n'];
+            if (count($translations) === 0 && $row->get('_translations')) {
```
lorenzo: use `empty()` instead of `count`

src/ORM/Behavior/Translate/TranslateStrategyTrait.php
```diff
@@ -0,0 +1,190 @@
...
+     *
+     * @var \Cake\ORM\Table
+     */
+    protected $table;
+
+    /**
+     * The locale name that will be used to override fields in the bound table
+     * from the translations table
+     *
+     * @var string
```
lorenzo: should this be `?string` ?

lorenzo (changes requested):

lorenzo:
Great job!

src/ORM/Behavior/Translate/ShadowTableStrategy.php
ADmad: The fields in `$fields` aren't aliased either, plus the translation table doesn't have any associations / containments so there's no possibility of duplicate field name errors.

ADmad:
> Great job!

Most of the work was done by @AD7six in his plugin. It just moved things around :slightly_smiling_face:.

src/ORM/Behavior/Translate/TranslateStrategyInterface.php
```diff
@@ -0,0 +1,116 @@
...
+     * Callback method that listens to the `beforeFind` event in the bound
+     * table. It modifies the passed query by eager loading the translated fields
+     * and adding a formatter to copy the values into the main table records.
+     *
+     * @param \Cake\Event\Event $event The beforeFind event that was fired.
+     * @param \Cake\ORM\Query $query Query
+     * @param \ArrayObject $options The options for the query
+     * @return void
+     */
+    public function beforeFind(Event $event, Query $query, ArrayObject $options): void;
```
robertpustulka: You could use interfaces in callbacks.

src/ORM/Behavior/Translate/ShadowTableStrategy.php
```diff
@@ -111,11 +111,11 @@ protected function setupAssociations()
      * and adding a formatter to copy the values into the main table records.
      *
      * @param \Cake\Event\EventInterface $event The beforeFind event that was fired.
-     * @param \Cake\Datasource\QueryInterface $query Query.
+     * @param \Cake\ORM\Query $query Query.
      * @param \ArrayObject $options The options for the query.
      * @return void
      */
-    public function beforeFind(EventInterface $event, QueryInterface $query, ArrayObject $options): void
+    public function beforeFind(EventInterface $event, Query $query, ArrayObject $options): void
```
dereuromark: Now that's sad if we must go back to concrete objects.

robertpustulka: @dereuromark we can add interfaces in another PR and update the definitions later.

src/ORM/Behavior/Translate/TranslateStrategyTrait.php
```diff
@@ -0,0 +1,190 @@
...
+            // If now, the current locale property is empty,
+            // unset it completely.
+            if (empty(array_filter($translation))) {
+                unset($entity->get('_translations')[$locale]);
+            }
+        }
+
+        // If now, the whole _translations property is empty,
+        // unset it completely and return
+        if (empty($entity->get('_translations'))) {
```
dereuromark: We shouldnt use empty() method on a method. Can we just have if (!$...)?

src/ORM/Behavior/Translate/TranslateStrategyTrait.php
```diff
@@ -0,0 +1,190 @@
...
+                if (strlen($value) === 0) {
+                    $translation->unsetProperty($field);
+                }
+            }
+
+            $translation = $translation->extract($this->_config['fields']);
+
+            // If now, the current locale property is empty,
+            // unset it completely.
+            if (empty(array_filter($translation))) {
```
dereuromark: Same here

src/ORM/Behavior/Translate/TranslateStrategyTrait.php
```diff
@@ -0,0 +1,190 @@
...
+     * @var \Cake\ORM\Table
+     */
+    protected $translationTable;
+
+    /**
+     * Return translation table instance.
+     *
+     * @return \Cake\ORM\Table
+     */
+    public function getTranslationTable(): Table
```
dereuromark: Would it be enough to typehint against the RepositoryInterface? Do we need any methods from concrete Table class?

ADmad: These strategies are not going to work for non relational databases. Methods for setting associations, getting 
primary key etc. don't exist in `RepositoryInterface`.

robertpustulka: Maybe we should introduce `TableInterface` then?

dereuromark: That would be cleaner :)

ADmad: Why not, but that's beyond the scope of this PR.

lorenzo:
@ADmad I'll have a look at this during the week, sorry that it took me so long

ADmad:
@lorenzo The "duplicate constraint name" error has been take care of. Now only the tests failures on postgres need to be looked into.

ADmad:
I have ignored the failing test on Postgres for now. It passes for all other drivers so I don't think there's any actual error, just the question of adjusting the SQL expression to work for postgres too, whenever someone gets time to look into it.

Merging this to avoid having to fix repeated merge conflicts.

PR #12116, opened 2018-05-22
4.x - EntityInterface.

saeideng:
Refs : https://github.com/cakephp/cakephp/issues/12094

### DIFF
--- a/src/Datasource/EntityInterface.php
+++ b/src/Datasource/EntityInterface.php
@@ -21,29 +21,145 @@ use JsonSerializable;
  * Describes the methods that any class representing a data storage should
  * comply with.
  *
- * In 4.x the following methods will officially be added to the interface:
- *
- * @method $this setHidden(array $properties, $merge = false)
- * @method array getHidden()
- * @method $this setVirtual(array $properties, $merge = false)
- * @method array getVirtual()
- * @method $this setDirty($property, $isDirty)
- * @method bool isDirty($property = null)
- * @method array getErrors()
- * @method array getError($field)
- * @method array setErrors(array $fields, $overwrite = false)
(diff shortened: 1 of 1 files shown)


src/Datasource/EntityInterface.php
```diff
@@ -21,30 +21,146 @@
...
+    public function extractOriginal(array $properties);
+
+    /**
+     * Returns an array with only the original properties
+     * stored in this entity, indexed by property name.
+     *
+     * @param array $properties List of properties to be returned
+     * @return array
+     */
+    public function extractOriginalChanged(array $properties);
```
saeideng: @lorenzo can we have `extractOriginal` / `extractOriginalChanged` in this  `Interface`  ?

src/Datasource/EntityInterface.php
```diff
@@ -21,30 +21,146 @@
...
+
+    /**
+     * Stores whether or not a property value can be changed or set in this entity.
+     *
+     * @param string|array $property single or list of properties to change its accessibility
+     * @param bool $set true marks the property as accessible, false will
+     * mark it as protected.
+     * @return $this
+     */
+    public function setAccess($property, $set);
```
dereuromark: Should we maybe stop using `string|array $property` mix, and instead offer a separate method for single property, e.g. `setFieldAccess()` or `setAccessForField()`?
This would make typehinting possible. But no big deal, maybe this is more convenient to use in its current form.

src/Datasource/EntityInterface.php
lorenzo: Yeah, I think it is fine to do that

ADmad (approved):

PR #12120, opened 2018-05-22
Make `\Cake\Utility\Text` accept `\Transliterator` objects.

ndm2:
Sometimes the built-in rules that ship with ICU aren't enough, and one has to provide custom ones, which requires using `Transliterator` objects, as rules cannot be embedded in identifier strings.

While it's actually already possible to pass `Transliterator` objects to `Text::setTransliteratorId()` and the `transliteratorId` option of `Text::slug()`, transliterator identifier strings are something different, and without "official support" it's kind of a hack.

This PR deprecates the current methods and options in favor of ones that accept both, `Transliterator` objects as well as transliterator identifier strings.

### DIFF
--- a/src/Utility/Text.php
+++ b/src/Utility/Text.php
@@ -23,20 +23,18 @@ class Text
 {
 
     /**
-     * Default transliterator id string.
-     *
-     * @deprecated 3.7.0 Use $_defaultTransliterator instead.
-     * @var string $_defaultTransliteratorId Transliterator identifier string.
-     */
-    protected static $_defaultTransliteratorId;
-
-    /**
      * Default transliterator.
      *
-     * @var \Transliterator|string Either a Transliterator instance, or a
-     *   transliterator identifier string.
+     * @var \Transliterator Transliterator instance.
      */
-    protected static $_defaultTransliterator = 'Any-Latin; Latin-ASCII; [\u0080-\u7fff] remove';
+    protected static $_defaultTransliterator;
+
+    /**
+     * Default transliterator id string.
+     *
+     * @var string $_defaultTransliteratorId Transliterator identifier string.
+     */
+    protected static $_defaultTransliteratorId = 'Any-Latin; Latin-ASCII; [\u0080-\u7fff] remove';
 
     /**
      * Default html tags who must not be count for truncate text.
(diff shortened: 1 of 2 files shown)


ADmad:
> ...as rules cannot be embedded in identifier strings.

I don't see a test for this case. In the tests you have added you are still passing a string to `Transliterator::createFromRules()`. Couldn't that string would have been passed to existing `setTransliteratorId()`?

ndm2:
It might look like it could, but it can't. The string in test there is a set of rules (or script... gotta admit that I'm not overly familiar with the terminology), with a trailing identifier that itself refers to a script, and neither of those are valid transliterator identifier syntax, passing that to `transliterator_transliterate()` will trigger an error along the lines of:

> transliterator_create: unable to open ICU transliterator with id "...": U_INVALID_ID

src/Utility/Text.php
```diff
@@ -1054,38 +1063,86 @@ public static function parseFileSize($size, $default = false)
...
+     * @deprecated 3.7.0 Use getTransliterator() instead.
      * @return string Transliterator identifier.
      */
     public static function getTransliteratorId()
     {
-        return static::$_defaultTransliteratorId;
+        deprecationWarning(
+            'Text::getTransliteratorId() is deprecated. ' .
+            'Use Text::getTransliterator() instead.'
+        );
```
markstory: This is a fairly new method.  Deprecating it now seems a bit indecisive and creates unnecessary API churn. Why can't we keep both the ID and transliterator instance? The `transliterator_transliterate` will work with either one, and internally we can prefer the object over the id.

ndm2: There isn't really a technical reason not to do that, I just thought that such internal weighting of properties might not be overly welcome, as there are quite a few other methods that seem to prefer union types, but if separate methods are preferred in this case, then I'll gladly change it.

ndm2: Would you prefer separate options for `Text::slug()` too?

markstory: I think separate options would be better. We could raise an error if users send both options.

ndm2: I gave it a try, but somehow having two options feels weird... they would both be able to trigger the default behavior when `null` is passed, while the possibly set transliterator object would get preferred by `Text::transliterate()`, also they would both be able to disable transliteration when `false` is passed.

Not sure what users would expect if they for example pass `transliteratorId => null`, would it be reasonable if they'd assume that the default transliterator _identifier_ is used, instead of the possibly set default transliterator _object_?

dereuromark: > Deprecating it now seems a bit indecisive and creates unnecessary API churn

If the new approach here is the better one, we shouldnt be afraid of fixing the API once more - hopefully the last time. So deprecation + having a single one forward seems to be reasonable none the less.

markstory: The current transliteratorId support was added in 3.2, which wasn't that long ago. Isn't the transliteratorId something that can be converted into an instance when the id is set? Same with the get, the id can be derived from the object. Does only accepting one or the other option for `transliterate()` make the API simpler?

I'm pushing on this because I think we're sometimes eager to deprecate and replace which can often become change without improvement for end users.

dereuromark: So you are saying this is already feature complete and this change here isn't necessary at all anymore given the current API?

ndm2: Just to make it clear, I was referring to the options for `Text::slug()` in my last comment! I'm less concerned about separate methods, even though I think a single getter/setter should be the goal at some point.

As far as the options go, I think it would be more straight forward with a single option that accepts both, identifiers and objects, just like `Text::transliterate()` can accept both. If deprecation is a concern there too, then the `transliteraterId` option could also be kept as an alias for the time being?

> Isn't the transliteratorId something that can be converted into an instance when the id is set? Same with the get, the id can be derived from the object.
	
Transliterator objects can be created from identifiers, and identifiers can be retrieved from the objects, but objects built from rules don't expose valid identifiers (they can't, as identifiers must refer to ICU transliterators), they'd always return `RulesTransPHP` as the identifier.

markstory: > Transliterator objects can be created from identifiers, and identifiers can be retrieved from the objects, but objects built from rules don't expose valid identifiers ... they'd always return RulesTransPHP as the identifier.

I think that's a reasonable trade-off to make.

> I think it would be more straight forward with a single option that accepts both, identifiers and objects

I'm on-board with that approach too.

ndm2: Ok. Just so I don't misunderstand it, because I'm a little confused, it should be separate getters/setters for identifier and object (where objects are derived from identifiers, and identifiers are derived from objects in case possible), and a single option (with or without deprecation notice?) for `Text::slug()`?

ndm2: I think the use case would be the same as the one for accepting objects in the first place, the need to apply custom rules. Wouldn't it be a little weird if only identifiers could be applied selectively? I'll push an update so we have something to look at.

Issue #12130, opened 2018-05-24
Duplicate Named Route for plugin


Erwane:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.6.4

### What you did
load plugin with routes in bootstrap
`Plugin::load('I18nUrl', ['routes' => true]);`

call `Plugin::routes()` at the end of config/routes.php in base app

create a route in plugin
```php
use Cake\Routing\Router;
use Cake\Routing\RouteBuilder;

Router::plugin('Backend', ['path' => '/gestion'], function (RouteBuilder $routes) {
    $routes->connect('/', ['controller' => 'Dashboard', 'action' => 'index'], ['_name' => 'backend.dashboard']);
});
```

### What happened
`Cake\Routing\Exception\DuplicateNamedRouteException`
`Error: A route named "backend.dashboard" has already been connected to "/gestion".`

### What you expected to happen
no duplicate routes

## where to search
It's look like the RoutingMiddleware now check if $this->app is instance of PluginApplicationInterface

the src/Application extends BaseApplication which implement PluginApplicationInterface SO the standard application build plugin routes twice, first via application and bootstrap, second time via the `RoutingMiddleware::prepareRouteCollection()`

Erwane:
i've put debug on these lines

config/routes.php, before `Plugin::routes()`
Backend/config/routes.php in line 2
Cake\Routing\Middleware line 112 before `$this->app->pluginRoutes($builder);`

output : 
```
/config/routes.php (line 108)
'/var/www/html/config/routes.php'
/plugins/Backend/config/routes.php (line 2)
'/var/www/html/plugins/Backend/config/routes.php'
/src/Routing/Middleware/RoutingMiddleware.php (line 112)
'Cake\Routing\Middleware\RoutingMiddleware::prepareRouteCollection'
/plugins/Backend/config/routes.php (line 2)
'/var/www/html/plugins/Backend/config/routes.php'
```
the Backend/config/routes.php was loaded twice

markstory:
You should remove the `Plugin::routes()` call in your routes file. It is only necessary if you are using the deprecated Dispatcher classes.

markstory:
One possible solution is to make `Plugin::routes()` do nothing when an `Application` class can be found. We could emit a deprecation warning as well to let folks know they should remove the `Plugin::routes()` call.

Erwane:
[...]

PR #12138, opened 2018-05-26
Deprecate Plugin::routes()

markstory:
I missed this earlier as I had updated my applications and removed this line. With this method undeprecated plugins will have their routes loaded twice. Once by `Application::pluginRoutes()` and again by `Plugin::routes()`. Because we're trending away from static APIs, and the `BasePlugin::route()` gives more control/flexibility I'm deprecating this method.

Refs #12130

### DIFF
--- a/src/Core/Plugin.php
+++ b/src/Core/Plugin.php
@@ -337,9 +337,16 @@ class Plugin
      * @param string|null $name name of the plugin, if null will operate on all
      *   plugins having enabled the loading of routes files.
      * @return bool
+     * @deprecated 3.6.5 This method is no longer needed when using HttpApplicationInterface based applications.
+     *   This method will be removed in 4.0.0
      */
     public static function routes($name = null)
     {
+        deprecationWarning(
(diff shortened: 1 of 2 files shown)


src/Core/Plugin.php
```diff
@@ -337,9 +337,16 @@ public static function bootstrap($name)
...
      *   plugins having enabled the loading of routes files.
      * @return bool
+     * @deprecated 3.6.5 This method is no longer needed when using HttpApplicationInterface based applications.
+     *   This method will be removed in 4.0.0
      */
     public static function routes($name = null)
     {
+        deprecationWarning(
+            'You no longer need to call `Plugin::routes()` after upgrading to use Http\Server. ' .
+            'See https://book.cakephp.org/3.0/en/development/application.html#adding-the-new-http-stack-to-an-existing-application ' .
```
saeideng: can we check using `Http\Server` before throwing deprecation warning ?
```php
if (file_exists (...Application.php)){
      deprecationWarning(...);
}
```

markstory: We could. But not using Application is also deprecated.

Issue #12131, opened 2018-05-24
normalizePath does not work as expected


AlPri78:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.6.
* Platform and Target: Windows 10

### What you did
Just using Folder::normalizePath()

### What happened
normalizePath returns only one character '\' on Windows else '/'
In fact, it's just a copy of the function Folder::correctSlashFor()

### What you expected to happen
I expect a return of the complete path but with the correct set of slashes (like description of the function, if I understand well), and like standard normalizePath function in other languages.

For example 
`pr(Folder::normalizePath('D:\www\site\webroot/img/'));`
would display on Windows

> D:\www\site\webroot\img\

It's seems that this function is not finished, or I understood nothing.

If you agree, I can pull a request.

AlPri78

markstory:
> It's seems that this function is not finished, or I understood nothing.

It looks like the function is pretty silly, and has been that way for a long time. I'm not sure its wise to 'fix' the function now as it will break compatibility. However, we could add new methods and deprecate the current method name/behavior.

AlPri78:
You are right, it's better to deprecate.

I would be happy to create a new PR with a new function named : normalizeFullPath because this function can work only with full path (based on isWindowsPath).

Do you agree ?

markstory:
@AlPri78 That works for me.

PR #12142, opened 2018-05-27
Deprecation of normalizePath to fix it in the new method normalizeFullPath

AlPri78:
#12131  

Everything in the title and in the issue.

It seems that other languages change only slashes, it's what I did. It could be possible to suppress double slashes, but because it's not very problematic and rare , so I take the option to not manage that. Tell me what do you think, perhaps other kind of work to do in this new method.

I also added some directives for a correct code analysis.

It misses documentation. I don't know if deprecation is automatically calculated or not. Please tell me and I will do the necessary.

AlPri78

### DIFF
--- a/src/Filesystem/Folder.php
+++ b/src/Filesystem/Folder.php
@@ -359,6 +359,7 @@ class Folder
      */
     public static function normalizePath($path)
     {
+        deprecationWarning('Folder::normalizePath() is deprecated. Use Folder::correctSlashFor() instead.');
         return Folder::correctSlashFor($path);
     }
 
--- a/src/Filesystem/Folder.php
+++ b/src/Filesystem/Folder.php
@@ -16,6 +16,7 @@ namespace Cake\Filesystem;
 
 use DirectoryIterator;
 use Exception;
+use FilesystemIterator;
 use InvalidArgumentException;
 use RecursiveDirectoryIterator;
 use RecursiveIteratorIterator;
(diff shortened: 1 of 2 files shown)


src/Filesystem/Folder.php
```diff
@@ -354,12 +354,28 @@ public static function isRegisteredStreamWrapper($path)
      *
      * @param string $path Path to check
      * @return string Set of slashes ("\\" or "/")
+     *
+     * @deprecated 3.6.5 This method will be removed in 4.0.0. Use correctSlashFor() instead.
```
markstory: Deprecations should emit runtime warnings with `deprecationWarning()` in addition to the documentation tag.

tests/TestCase/Filesystem/FolderTest.php
```diff
@@ -680,23 +680,50 @@ public function testSlashTerm()
...
         $result = Folder::normalizePath($path);
         $this->assertEquals('/', $result);
 
         $path = '\\path\\\to\\\file';
+        /** @noinspection PhpDeprecationInspection */
         $result = Folder::normalizePath($path);
         $this->assertEquals('/', $result);
 
         $path = 'C:\\path\\to\\file';
+        /** @noinspection PhpDeprecationInspection */
```
markstory: You can use the `$this->deprecated()` helper to swallow deprecation warnings during this test. We generally don't use these comment directives either.

src/Filesystem/Folder.php
```diff
@@ -355,7 +355,7 @@ public static function isRegisteredStreamWrapper($path)
      * @param string $path Path to check
      * @return string Set of slashes ("\\" or "/")
      *
-     * @deprecated 3.6.5 This method will be removed in 4.0.0. Use correctSlashFor() instead.
+     * @deprecated 3.7 This method will be removed in 4.0.0. Use correctSlashFor() instead.
```
ADmad: Since this patch is targeted to `master` I will go into next `3.6` release hence the version this method will be deprecated in will be `3.6.5`.

AlPri78: I changed that because of Markstory put the 3.7 milestone yesterday.
Do you ask to change that or is it not correlated ?

AlPri78:
Hello, I'd need some help to finalize.
Since markstory changes the branch I encounter different errors in Travis but not on my coding. What did I miss ?

markstory:
Thank you 👍

markstory:
> If possible, may I have just a very little explanation on why I encountered errors on Travis, and you merge my modifications with that... It's just for my potential future contributions!

Travis often has partial failures due to infrastructure issues or timeouts. There isn't much you need to do in those scenarios as we can restart builds and compare individual stages of the matrix with their sibling builds.

PR #12143, opened 2018-05-27
Add missing methods to LocatorInterface.

ADmad:
Refs  #12094

### DIFF
--- a/src/ORM/Locator/LocatorInterface.php
+++ b/src/ORM/Locator/LocatorInterface.php
@@ -18,12 +18,31 @@ use Cake\ORM\Table;
 
 /**
  * Registries for Table objects should implement this interface.
- *
- * @method array getConfig($alias)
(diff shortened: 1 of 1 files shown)


src/ORM/Locator/LocatorInterface.php
```diff
@@ -18,12 +18,31 @@
...
+     * with a matching alias.
+     *
+     * @param string|array $alias Name of the alias or array to completely
+     *   overwrite current config.
+     * @param array|null $options list of options for the alias
+     * @return $this
+     * @throws \RuntimeException When you attempt to configure an existing
+     *   table instance.
+     */
+    public function setConfig($alias, $options = null);
```
dereuromark: can options be an array now?

ADmad: Probably, but I am gonna leave that for later when we start adding proper typehints across the framework.

PR #12150, opened 2018-05-29
Generate routes as part of the integration tests

### LINKED ISSUE #12146: Integration tests and application routing
This is a (multiple allowed):

* [x] bug
* [ ] enhancement

* CakePHP Version: 3.6.x
* Platform and Target: Any

### What you did
When relying on the Application class's `route()` hook to load routes, OR using a plugin with routes integration tests fail when trying to use array based routes.

A fully working example application is available: https://github.com/dakota/plugin-test/tree/application-routing

### What happened
`Cake\Routing\Exception\MissingRouteException` is thrown as demonstrated https://travis-ci.com/dakota/plugin-test/builds/74640142

Integration tests should work.

The cause of failure is https://github.com/cakephp/cakephp/blob/master/src/TestSuite/IntegrationTestCase.php#L671

An attempt is made to match the route before the Router middleware is loaded, and so no routes have been loaded at this time. Trying to load the routes manually before the tests works unless named routes are used (In which case a duplicate route exception is thrown)

The server class will need to be refactored to load middleware/routes before execution. Alternatively an `AppBuilder` needs to be created that can build the application state (Including middleware and routing) before a server instance is created.

dakota:
Closes #12146 

This fixes the issue with integration tests not loading routes for use with array based urls.

### DIFF
--- a/src/TestSuite/MiddlewareDispatcher.php
+++ b/src/TestSuite/MiddlewareDispatcher.php
@@ -14,9 +14,12 @@
 namespace Cake\TestSuite;
 
 use Cake\Core\Configure;
+use Cake\Core\HttpApplicationInterface;
+use Cake\Core\PluginApplicationInterface;
 use Cake\Event\EventManager;
 use Cake\Http\Server;
 use Cake\Http\ServerRequestFactory;
+use Cake\Routing\Router;
 use LogicException;
 use ReflectionClass;
 use ReflectionException;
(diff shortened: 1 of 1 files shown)


dakota:
Really struggling to get the tests passing for cases where

1. Routes are set in the test itself (which should not be cleared with a `Router::reload()`)
2. New style plugins where the application is loading routes via the routes.php file.

markstory:
@dakota I'm planning on looking at this issue next.

markstory:
Closing as these changes have been combined into #12184

Issue #12158, opened 2018-05-30
I18n unit tests fail on fresh clone (macOS, PHP 7.1.16)


michaelbutler:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: cfaa31c2cb09309a4db90daca1d64c7e5c3f6a07 (master on May 30, 2018 3:00 PM America\New_York)
* Platform and Target: php 7.1.16 CLI, macOS (brew installed), running unit tests

### What you did
```
git clone git@github.com:cakephp/cakephp.git
composer install
php -dmemory_limit=512M vendor/bin/phpunit tests/TestCase/I18n
```

### What happened
Got 8 failing tests:

```
PHPUnit 6.5.8 by Sebastian Bergmann and contributors.

..FF...........................................................  63 / 317 ( 19%)
............S.................................................. 126 / 317 ( 39%)
............................................................... 189 / 317 ( 59%)
............................................................... 252 / 317 ( 79%)
.....................FFFF..FF.................................. 315 / 317 ( 99%)
..                                                              317 / 317 (100%)

Time: 1.31 seconds, Memory: 12.00MB

There were 8 failures:

1) Cake\Test\TestCase\I18n\DateTest::testI18nFormat with data set "mutable" ('Cake\I18n\Date')
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'jeudi 14 janvier 2010 00:00:00 UTC'
+'jeudi 14 janvier 2010 00:00:00 Temps universel coordonné'

/Users/bmichael/tmp/cakephp/tests/TestCase/I18n/DateTest.php:112

2) Cake\Test\TestCase\I18n\DateTest::testI18nFormat with data set "immutable" ('Cake\I18n\FrozenDate')
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'jeudi 14 janvier 2010 00:00:00 UTC'
+'jeudi 14 janvier 2010 00:00:00 Temps universel coordonné'

/Users/bmichael/tmp/cakephp/tests/TestCase/I18n/DateTest.php:112

3) Cake\Test\TestCase\I18n\TimeTest::testI18nFormat with data set "mutable" ('Cake\I18n\Time')
Failed asserting that two strings are identical.
--- Expected
+++ Actual
@@ @@
-'jueves 14 de enero de 2010 13:59:28 GMT'
[...]

PR #11995, opened 2018-04-25
Attempt to get appveyor builds running on 4.x

markstory:
Fix appveyor install script.

PR #12159, opened 2018-05-30
Fix up failing tests when run against new versions of libicu.

markstory:
Refs #12158


saeideng:
@markstory  
right now tests fixed on cake 4.0 ,
do you think we need to fix them on 3.6 again ?

saeideng:
PR for cake 4 : https://github.com/cakephp/cakephp/pull/11995

markstory:
> I find it odd that it would change so much from one version to another, though I don't know much about libicu.

Welcome to the wild world of libicu.

>  Changing a assertSame into assertContains is a bit weak since it means we don't really know what the application will do. Since we're dealing with date wrappers here, could it be solved from within the wrapper? As dirty as version checks are, maybe that's needed.

If you have the energy to invest in getting tests to pass across multiple libicu versions using `assertSame()` we'd welcome that kind of contribution. Right now I don't have the patience for it.

PR #12162, opened 2018-05-31
Fix # 11827: Replace use of each() in DboSource::order for PHP 7.2 compatibility

trenshaw:
Replaced is_array() check and set of $keys in array with cast to array of $keys in array_filter()

Replaced assignment of $key and $dir through list() with each() with simply key() and current() per tenkoma's suggestion. Resolves # 11827

each() advances the pointer similar to next() but is being negated this particular case by array_shift() which resets the pointer

### DIFF
--- a/lib/Cake/Model/Datasource/DboSource.php
+++ b/lib/Cake/Model/Datasource/DboSource.php
@@ -3043,7 +3043,7 @@ class DboSource extends DataSource {
  * @return string ORDER BY clause
  */
 	public function order($keys, $direction = 'ASC', Model $Model = null) {
-		$keys = array_filter((array) $keys);
+		$keys = array_filter((array)$keys);
 		
 		$result = array();
 		while (!empty($keys)) {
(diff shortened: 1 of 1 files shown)


dereuromark:
Looks like a dup of https://github.com/cakephp/cakephp/pull/11827 ?

lib/Cake/Model/Datasource/DboSource.php
```diff
@@ -3043,15 +3043,12 @@ public function limit($limit, $offset = null) {
  * @return string ORDER BY clause
  */
 	public function order($keys, $direction = 'ASC', Model $Model = null) {
-		if (!is_array($keys)) {
-			$keys = array($keys);
-		}
-
-		$keys = array_filter($keys);
-
+		$keys = array_filter((array) $keys);
```
dereuromark: Casts dont have whitespace afterwards (see CS).

markstory:
These changes have caused tests to fail as the order clause is in a different order.

PR #12164, opened 2018-05-31
Fix #12163

### LINKED ISSUE #12163: Model.afterSaveCommit dispatch has mixed type for third argument
This is a (multiple allowed):

* [x] bug
* [ ] enhancement

* CakePHP Version: 3.6.4 (but from release notes, I found this happens since 3.6.3)
* Platform and Target: nginx, mysql, ubuntu 16.04

### What you did

`public function afterSaveCommit(Event $event, EntityInterface $entity, ArrayObject $options)`

### What happened

Called `findOrCreate()`

Got: argument 3 passed to Camptoo\ORM\Table::afterSaveCommit() must be an instance of ArrayObject, array given

No error

https://github.com/cakephp/cakephp/blob/master/src/ORM/Table.php#L1678

This line should convert options to an ArrayObject, just like the save operation does.

JoepRoebroek:
Fixes the issue mentioned in #12163

### DIFF
--- a/src/ORM/Table.php
+++ b/src/ORM/Table.php
@@ -1675,13 +1675,13 @@ class Table implements RepositoryInterface, EventListenerInterface, EventDispatc
      */
     public function findOrCreate($search, callable $callback = null, $options = [])
     {
-        $options += [
+        $options = new ArrayObject($options + [
             'atomic' => true,
             'defaults' => true,
-        ];
+        ]);
 
         $entity = $this->_executeTransaction(function () use ($search, $callback, $options) {
-            return $this->_processFindOrCreate($search, $callback, $options);
+            return $this->_processFindOrCreate($search, $callback, $options->getArrayCopy());
         }, $options['atomic']);
 
         if ($entity && $this->_transactionCommitted($options['atomic'], true)) {
--- a/tests/TestCase/ORM/TableTest.php
+++ b/tests/TestCase/ORM/TableTest.php
@@ -6150,7 +6150,7 @@ class TableTest extends TestCase
     public function testFindOrCreateTransactions()
     {
         $articles = $this->getTableLocator()->get('Articles');
-        $articles->getEventManager()->on('Model.afterSaveCommit', function ($event, $entity) {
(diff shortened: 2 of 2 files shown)


src/ORM/Table.php
```diff
@@ -1675,13 +1675,13 @@ protected function _transactionCommitted($atomic, $primary)
...
-        $options += [
+        $options = new ArrayObject($options + [
             'atomic' => true,
             'defaults' => true,
-        ];
+        ]);
 
         $entity = $this->_executeTransaction(function () use ($search, $callback, $options) {
-            return $this->_processFindOrCreate($search, $callback, $options);
+            return $this->_processFindOrCreate($search, $callback, $options->getArrayCopy());
```
ADmad: Is using `getArrayCopy()` necessary here?

JoepRoebroek: Yes, otherwise tests fail on this line:
https://github.com/cakephp/cakephp/blob/master/src/ORM/Table.php#L1892

`save()` expects an `array` for options. But depends on what is desired, not sure if eventually the `$options` argument in the `save()` function will be typehinted? If so, what will the typehint become? `ArrayObject` or `array`? And instanceof check can also be added to `save()` which then does the `getArrayCopy()`. This solution might give more flexibility, since people can then also start using `ArrayObject` in the `save()` function

markstory: > This solution might give more flexibility, since people can then also start using ArrayObject in the save() function

Having end users need to use ArrayObject sounds like more work for them than simple arrays would be.

lorenzo: `$options` for the `save()` method will remain an array, both for usability and predictability reasons.

src/ORM/Table.php
```diff
@@ -1675,13 +1675,13 @@ protected function _transactionCommitted($atomic, $primary)
      */
     public function findOrCreate($search, callable $callback = null, $options = [])
     {
-        $options += [
+        $options = new ArrayObject($options + [
```
ADmad: Creating arrayobject here and converting it back to array when calling _processFindOrCreate() seem redundant. Wouldn't it be better to create arrayobject later just before dispatching the `afterSaveCommit` event?

PR #12170, opened 2018-06-02
Create method for import html

alphp:
loadHtml parse html string and create either SimpleXmlElement object or a DOMDocument.
use of empty($options['loadEntities']) instead of !$options['loadEntities']

### DIFF
--- a/src/Utility/Xml.php
+++ b/src/Utility/Xml.php
@@ -142,7 +142,7 @@ class Xml
     {
         $hasDisable = function_exists('libxml_disable_entity_loader');
         $internalErrors = libxml_use_internal_errors(true);
-        if ($hasDisable && empty($options['loadEntities'])) {
+        if ($hasDisable && !$options['loadEntities']) {
             libxml_disable_entity_loader(true);
         }
         $flags = 0;
@@ -162,7 +162,7 @@ class Xml
         } catch (Exception $e) {
             throw new XmlException('Xml cannot be read. ' . $e->getMessage(), null, $e);
         } finally {
-            if ($hasDisable && empty($options['loadEntities'])) {
+            if ($hasDisable && !$options['loadEntities']) {
                 libxml_disable_entity_loader(false);
             }
             libxml_use_internal_errors($internalErrors);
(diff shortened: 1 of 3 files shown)


src/Utility/Xml.php
```diff
@@ -162,7 +162,46 @@ protected static function _loadXml($input, $options)
...
+
+    /**
+     * Parse the input html string and create either a SimpleXmlElement object or a DOMDocument.
+     *
+     * @param string $input The input html string to load.
+     * @param array $options The options to use. See Xml::build()
+     * @return \SimpleXMLElement|\DOMDocument
+     * @throws \Cake\Utility\Exception\XmlException
+     */
+    public static function loadHtml($input, $options)
```
markstory: New methods require tests to be merged.

markstory:
Moved to 3.next as new features do not go into bugfix releases.

src/Utility/Xml.php
```diff
@@ -142,7 +142,7 @@ protected static function _loadXml($input, $options)
     {
         $hasDisable = function_exists('libxml_disable_entity_loader');
         $internalErrors = libxml_use_internal_errors(true);
-        if ($hasDisable && !$options['loadEntities']) {
+        if ($hasDisable && empty($options['loadEntities'])) {
```
markstory: >  use of empty($options['loadEntities']) instead of !$options['loadEntities']

Why though? What does this change?

alphp: If not exists $options['loadEntities'] Notice is raised:
```
2018-06-02 02:48:02 Notice: Notice (8): Undefined index: loadEntities in [C:\usr\0xFF\bambu\app\vendor\cakephp\cakephp\src\Utility\Xml.php, line 184]
```

markstory: How would that happen? The public methods set a default value.

alphp: Ok, I see.
I based loadHtml in loadXml thats is calling by build, so in loadHtml $options don't have default value.

tests/TestCase/Utility/XmlTest.php
```diff
@@ -247,6 +247,39 @@ public function testBuildEmptyTag()
...
+     */
+    public function testLoadHtml()
+    {
+        $html = CORE_TESTS . 'Fixture/sample.html';
+        $html = file_get_contents($html);
+        $xml = Xml::loadHtml($html);
+        $xml = html_entity_decode($xml->asXML(), ENT_NOQUOTES, 'UTF-8');
+        $this->assertEquals($html, $xml);
+
+        $xml = Xml::loadHtml($html, ['parseHuge' => true]);
```
markstory: These tests are looking good. There should be a teat case covering the domdocument return tyoe as well.

PR #12172, opened 2018-06-02
4.x: Use EventInterface instead of Event for typehints.

ADmad:
This is one of the tasks listed on 4.0 roadmap.


markstory:
:clap: Great work @ADmad

PR #12178, opened 2018-06-04
Use getter/setter instead of deprecated properties.

ADmad:


src/View/HelperRegistry.php
```diff
@@ -144,7 +144,7 @@ protected function _throwMissingClassError($class, $plugin)
     protected function _create($class, $alias, $settings)
     {
         $instance = new $class($this->_View, $settings);
-        $vars = ['request', 'theme', 'plugin'];
+        $vars = ['theme', 'plugin'];
```
markstory: Won't this result in helpers no longer having a request set? We should retain that behavior as folks are likely relying on it.

ADmad: Helper's constructor already has `$this->request = $View->getRequest();` so setting again here was redundant.

markstory: ok.

PR #12179, opened 2018-06-04
3.7 view props

ADmad:
- All the properties which I have marked as deprecated have getter/setter methods.
- `View::$uuids` seems unused hence marked for removal in 4.0.
- The following public properties still need to be dealt with:
  - `$Blocks`: We already have wrapper methods like `View::fetch()` etc. for most `ViewBlocks` methods. Could add `getBlocks()` and make `$Blocks` protected.
  - `$plugin`: Should we deprecate and remove in 4.0 and just use request instance to get plugin name?
  - `$name`:  Deprecate as public and add `getName()`?
  - `$subDir`
  - `$hasRendered`
  - `$elementCache`
  - `$helpers`

### DIFF
--- a/src/View/View.php
+++ b/src/View/View.php
@@ -108,7 +108,7 @@ class View implements EventDispatcherInterface
      *
      * @var string
      */
-    public $name;
+    protected $name;
 
     /**
      * Current passed params. Passed to View from the creating Controller for convenience.
@@ -1201,6 +1201,7 @@ class View implements EventDispatcherInterface
             'response' => 'getResponse',
             'subDir' => 'getSubdir',
             'plugin' => 'getPlugin',
+            'name' => 'getName',
         ];
         if (isset($protected[$name])) {
             $method = $protected[$name];
@@ -1278,6 +1279,7 @@ class View implements EventDispatcherInterface
             'response' => 'setResponse',
             'subDir' => 'setSubDir',
             'plugin' => 'setPlugin',
+            'name' => 'setName',
             'elementCache' => 'setElementCache',
         ];
         if (isset($protected[$name])) {
--- a/src/View/View.php
+++ b/src/View/View.php
@@ -94,7 +94,7 @@ class View implements EventDispatcherInterface
      *
      * @var \Cake\View\ViewBlock
      */
-    public $Blocks;
+    protected $Blocks;
 
     /**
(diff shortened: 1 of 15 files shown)


src/View/View.php
```diff
@@ -129,6 +129,7 @@ class View implements EventDispatcherInterface
      * The name of the subfolder containing templates for this View.
      *
      * @var string
+     * @deprecated 3.7.0 The property will become protected in 4.0.0. Use getTemplatePath()/setTemplatePath() instead.
```
markstory: Should these become protected now and use `__get/__set/__isset` to emit deprecation warnings?

ADmad: I thought we weren't going to add more explicit warnings in `3.7`, but if we are then yes could make them protected and add magic methods for dealing with them.

markstory: If we don't have hard warnings I don't think we can should make the breaking change in 4.0.0. I've been thinking more and I think 3.7 could be a 'catch up' release that contains a final sweep for deprecations.

ADmad: Okay, I'll update the PR.

ADmad:
@markstory Updated. Please let me know the course of action for the remaining public properties I have listed in initial comment.

markstory:
* hasRendered could become protected with a getter
* Blocks could become protected. The view methods can be used instead of direct access.
* plugin and name could read from the request in most cases I think.
* subDir and elementCache could both become protected with setters. I don't know if we need getters for these. I have not seen them being read before.
* helpers could be deprecated like how we have deprecated components and helpers in controllers.

ADmad:
> * hasRendered could become protected with a getter

Done

> * Blocks could become protected. The view methods can be used instead of direct access.

Done

> * plugin and name could read from the request in most cases I think.

Both can be set through view builder (which also get getter / setter from them) so I just added getter / setters for it. Also `Email` was setting `View::$plugin` so can't rely on request instance to get it's value.

> * subDir and elementCache could both become protected with setters. I don't know if we need getters for these. I have not seen them being read before.

Done. Added a getter too for `subDir` as it was needed in tests.

> * helpers could be deprecated like how we have deprecated components and helpers in controllers.

I made `helpers` protected. Being able to manipulate the helper config before helpers are loaded is useful. We do it in custom view class in crud-view plugin.

PR #12185, opened 2018-06-05
Catching StopException in ConsoleIntegrationTestCase

### LINKED ISSUE #12181: abort() in command class breaking tests
This appears to be a bug in CakePHP Version: 3.6.4

I'm having trouble testing `$this->abort()` in a Command class. The docs suggest that I can just use `$this->assertExitCode(Command::CODE_ERROR);`, but the `StopException` that `abort()` throws bubbles all the way up to my test and breaks it unless if I put `exec()` in a try/catch block.

Should `ConsoleIntegrationTestCase::exec()` be catching exceptions thrown by `abort()`?

### What I did
```
class FooCommand extends Command
{
    public function execute(Arguments $args, ConsoleIo $io)
    {
        $this->abort();
    }
}
```
```
class FooCommandTest extends ConsoleIntegrationTestCase
{
    public function setUp()
    {
        parent::setUp();
        $this->useCommandRunner();
    }
	
    public function testExecute()
    {
        $this->exec('foo');
        $this->assertExitCode(Command::CODE_ERROR);
    }
}
```

### What happened
`Cake\Console\Exception\StopException : Command aborted` is thrown and a a stack trace is displayed.

### What I expected to happen
[The docs](https://book.cakephp.org/3.0/en/console-and-shells/commands.html#testing-interactive-shells) suggest that `abort()` won't break tests and can be tested with `assertExitCode()`.

jeremyharris:
Fixes #12181 

This will need to be ported to the trait in 3.7.

### DIFF
--- a/src/TestSuite/ConsoleIntegrationTestCase.php
+++ b/src/TestSuite/ConsoleIntegrationTestCase.php
@@ -95,7 +95,7 @@ abstract class ConsoleIntegrationTestCase extends TestCase
         try {
             $this->_exitCode = $runner->run($args, $io);
         } catch (StopException $exception) {
-            $this->_exitCode = Command::CODE_ERROR;
+            $this->_exitCode = $exception->getCode();
         }
     }
 
--- a/src/TestSuite/ConsoleIntegrationTestCase.php
+++ b/src/TestSuite/ConsoleIntegrationTestCase.php
@@ -13,9 +13,11 @@
  */
 namespace Cake\TestSuite;
 
+use Cake\Console\Command;
 use Cake\Console\CommandRunner;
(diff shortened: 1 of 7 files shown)


PhantomWatson:
It looks like this won't allow non-default exit codes in a command to be tested, e.g. `$this->abort($nonstandardExitCode)`, since the test changes the exit code to `Command::CODE_ERROR` whenever `StopException` is thrown.

Maybe `$this->_exitCode = $exception->getCode();` on `src/TestSuite/ConsoleIntegrationTestCase.php:98`?

dereuromark:
Instead of 127, the

    class StopException extends Exception

should default to code 1 as default exit code IMO.

PR #12188, opened 2018-06-05
4.x Replaced usages of TestCase classes with traits

jeremyharris:
refs: https://github.com/cakephp/cakephp/issues/11934

Although this isn't released yet I thought I'd get a jump on it so we can tick the TestSuite box.


dereuromark:
Why are we using a trait here? And not an extended class like usual?
Traits are harder to customize usually.

jeremyharris:
@dereuromark here's the original PR https://github.com/cakephp/cakephp/pull/12072

Basically it makes allows users to create a base TestCase class that has logic that can be shared between multiple types of test cases, see https://github.com/cakephp/cakephp/pull/12072#discussion_r187710468

jeremyharris:
Closing as there's no harm in holding onto these classes to give users a longer time to migrate.

Issue #11935, opened 2018-04-16
Add scalar typehints and return values where possible


markstory:
Tracking issue for adding scalar typehints and return values where possible in 4.x. Depends on #11934 being completed first.

* [x] Auth
* [x] Cache
* [x] Cache/Engine
* [x] Collection
* [x] Command
* [x] Console
* [x] Controller
* [x] Core
* [x] Database 
* [x] Database/Dialect
* [x] Database/Driver
* [x] Database/Expression
* [x] Database/Schema
* [x] Database/Statement
* [x] Database/Type
* [x] Datasource
* [x] Error
* [x] Event
* [x] Filesystem
* [x] Form
* [x] Http
* [x] Http/Client
* [x] Http/Middleware
* [x] I18n
* [x] Log
* [x] Mailer
* [x] Network
* [x] ORM
* [x] ORM/Association
* [x] ORM/Behavior
* [x] ORM/Rule
* [x] Routing
* [x] Shell
* [x] Template
* [x] TestSuite
* [x] TestSuite/Constraint
* [x] TestSuite/Fixture
* [x] TestSuite/Stub
* [x] Utility
* [x] Validation
* [x] View
* [x] View/Helper

dereuromark:
I also wouldn't do this too early, it will just create a lot of merge conflicts then from 3.x into 4.x.
So better to wait here a while.

lorenzo:
will it create conflicts? The only lines that would change are function signatures, which we don’t actually change in patch versions

markstory:
I'll take Collection and Console.

dereuromark:
Talking about this:
I am strongly in favor of also starting to typehint protected methods. This really helps tons in getting errors early on, both for core and extending classes in userland.
The downsides of performances in higher PHP7 versions can be neglected, especially with the upsides in typesafety.

markstory:
I've been using 

```
perl -pi -w -e 's/\<\?php\n/<?php\ndeclare(strict_types=1);\n/' <path>/*.php
```
to add declare statements if anyone else/future me needs it.

ADmad:
One can use the following to automatically add typehints and then tweak them manually if needed.

```php
vendor/bin/phpcbf --standard=SlevomatCodingStandard --sniffs=SlevomatCodingStandard.TypeHints.TypeHintDeclaration
```

ADmad:
We need to remove the `self` return type hints. It generates an error when you override a method in extended class which returns `$this`.

[...]

PR #12191, opened 2018-06-05
Auth type hints and strict types

cleptric:
Refs: #11935

### DIFF
--- a/src/Auth/AbstractPasswordHasher.php
+++ b/src/Auth/AbstractPasswordHasher.php
@@ -50,7 +50,7 @@ abstract class AbstractPasswordHasher
      *
      * @param string $password Plain text password to hash
      *   required to generate password hash.
-     * @return mixed Either the password hash string or bool
+     * @return string|bool Either the password hash string or bool
      */
     abstract public function hash(string $password);
 
--- a/src/Auth/AbstractPasswordHasher.php
+++ b/src/Auth/AbstractPasswordHasher.php
@@ -38,7 +38,6 @@ abstract class AbstractPasswordHasher
      * Constructor
      *
      * @param array $config Array of config.
-     * @return void
      */
     public function __construct(array $config = [])
     {
--- a/src/Auth/AbstractPasswordHasher.php
+++ b/src/Auth/AbstractPasswordHasher.php
@@ -1,5 +1,5 @@
 <?php
-declare(strict_types=1);
+declare(strict_types = 1);
 /**
  * CakePHP(tm) : Rapid Development Framework (https://cakephp.org)
  * Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
--- a/src/Auth/AbstractPasswordHasher.php
+++ b/src/Auth/AbstractPasswordHasher.php
@@ -1,5 +1,5 @@
 <?php
-declare(strict_types = 1);
+declare(strict_types=1);
 /**
  * CakePHP(tm) : Rapid Development Framework (https://cakephp.org)
  * Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
--- a/src/Auth/AbstractPasswordHasher.php
+++ b/src/Auth/AbstractPasswordHasher.php
@@ -1,4 +1,5 @@
 <?php
+declare(strict_types = 1);
 /**
  * CakePHP(tm) : Rapid Development Framework (https://cakephp.org)
  * Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
@@ -37,6 +38,7 @@ abstract class AbstractPasswordHasher
      * Constructor
      *
      * @param array $config Array of config.
+     * @return void
      */
     public function __construct(array $config = [])
     {
@@ -46,21 +48,21 @@ abstract class AbstractPasswordHasher
     /**
      * Generates password hash.
      *
-     * @param string|array $password Plain text password to hash or array of data
+     * @param string $password Plain text password to hash
      *   required to generate password hash.
-     * @return string Password hash
+     * @return mixed Either the password hash string or bool
      */
-    abstract public function hash($password);
+    abstract public function hash(string $password);
 
     /**
      * Check hash. Generate hash from user provided password string or data array
      * and check against existing hash.
      *
-     * @param string|array $password Plain text password to hash or data array.
+     * @param string $password Plain text password to hash or data array.
      * @param string $hashedPassword Existing hashed password.
      * @return bool True if hashes match else false.
      */
-    abstract public function check($password, $hashedPassword);
+    abstract public function check(string $password, string $hashedPassword): bool;
 
     /**
      * Returns true if the password need to be rehashed, due to the password being
@@ -72,7 +74,7 @@ abstract class AbstractPasswordHasher
      * @param string $password The password to verify
      * @return bool
      */
-    public function needsRehash($password)
+    public function needsRehash(string $password): bool
     {
         return password_needs_rehash($password, PASSWORD_DEFAULT);
     }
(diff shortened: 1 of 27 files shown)


src/Auth/AbstractPasswordHasher.php
```diff
@@ -46,21 +48,21 @@ public function __construct(array $config = [])
     /**
      * Generates password hash.
      *
-     * @param string|array $password Plain text password to hash or array of data
+     * @param string $password Plain text password to hash
      *   required to generate password hash.
-     * @return string Password hash
+     * @return mixed Either the password hash string or bool
```
dereuromark: Dont use mixed, always specify the types with | separation.

src/Auth/BaseAuthenticate.php
```diff
@@ -139,7 +142,7 @@ protected function _findUser($username, $password = null)
      * @param string $username The username/identifier.
      * @return \Cake\ORM\Query
      */
-    protected function _query($username)
+    protected function _query(string $username): Query
```
dereuromark: If you force-return something here, it sure shouldnt be a Query object, but a true contract (interface) or alike, the concrete ORM level object on the Auth namespace doesn't sound like a good idea for such a hard return type fixation.

src/Auth/BaseAuthenticate.php
```diff
@@ -168,7 +171,7 @@ protected function _query($username)
      * @throws \RuntimeException If password hasher class not found or
      *   it does not extend AbstractPasswordHasher
      */
-    public function passwordHasher()
+    public function passwordHasher(): AbstractPasswordHasher
```
dereuromark: Same here

src/Auth/Storage/StorageInterface.php
```diff
@@ -23,24 +24,24 @@
     /**
      * Read user record.
      *
-     * @return \ArrayAccess|array|null
+     * @return mixed \ArrayAccess, array or null
```
dereuromark: The mixed is wrong here, before was correct. Why should there suddenly be more types?

tests/TestCase/Auth/DigestAuthenticateTest.php
```diff
@@ -470,9 +473,10 @@ public function testPassword()
      *
      * @param string $secret The secret to use.
      * @param int $expires Time to live
+     * @param int $time Current time in microseconds
      * @return string
      */
-    protected function generateNonce($secret = null, $expires = 300, $time = null)
+    protected function generateNonce(string $secret = null, int $expires = 300, int $time = null): string
```
dereuromark: nullable is via `?int` - as per code sniffer enforced if you would use it :)

src/Auth/BaseAuthenticate.php
cleptric: So the annotation is simply wrong here?

src/Auth/BaseAuthenticate.php
cleptric: So the annotation is simply wrong here?

src/Auth/BaseAuthenticate.php
markstory: If every typehint needs to be an interface we won't have many typehints to add. Adding interfaces for each class so it can be used in a typehint seems tedious and against the goal of helping ides and preventing accidents.

ADmad:
Tests failing on PHP 7.2.

src/Auth/BaseAuthenticate.php
ADmad: Adding interfaces for each type of class is neither practical or necessary.

dereuromark: The added `use Cake\ORM\Query;` statement shows that up until now there was no hard coupling between two very concrete objects. That now changes.
But I can see how the framework is yet still very far away of doing interface contracting instead of hard class object coupling across different layers etc.
So never mind me here :)

src/Auth/BaseAuthenticate.php
dereuromark: I didnt say wrong. I basically meant "not ideal". Never mind, though, in this case.

src/Auth/AbstractPasswordHasher.php
```diff
@@ -1,4 +1,5 @@
 <?php
+declare(strict_types = 1);
```
ADmad: The [examples](https://github.com/php-fig/fig-standards/blob/master/proposed/extended-coding-style-guide.md#example) in proposed PSR12 don't show spaces around `=`. So it's better to follow that instead of having to potentially change in future.

src/Auth/AbstractPasswordHasher.php
```diff
@@ -37,6 +38,7 @@
      * Constructor
      *
      * @param array $config Array of config.
+     * @return void
```
ADmad: Constructor shouldn't have a return type since they can never return a value.

src/Auth/AbstractPasswordHasher.php
dereuromark: Also, couldnt a new sniffer rule also assert that?
Additionally, why not in the same line, no need to further space things
```
<?php declare(strict_types = 1);

/**
```

dereuromark: The example further down also shows that this is valid.

src/Auth/BaseAuthenticate.php
```diff
@@ -1,4 +1,5 @@
 <?php
+declare(strict_types = 1);
```
dereuromark: I would move this up into the same line, and instead add a newline underneath :)

src/Auth/BaseAuthenticate.php
markstory: Skipping the typehints and continuing to rely on documentation annotations and phpstan is an option. I think that unless we have a guideline on what kinds of classes can be used in typehints we're going to spend a lot of time discussing these matters.

dereuromark: We can also fix up return types to a better/cleaner approach in 5.0.
So I guess this is OK here for now.

src/Auth/AbstractPasswordHasher.php
cleptric: The example also states out: 
```
When wishing to declare strict types in files containing markup outside PHP opening and closing tags MUST, on the first line, include an opening PHP tag, the strict types declaration and closing tag.
```

ADmad: The case where strict types declaration is on same line as PHP tag is when the file has PHP code mixed with HTML. In PHP only files it's on line below the opening PHP tag, so let's keep the same.

tests/TestCase/Auth/Storage/SessionStorageTest.php
```diff
@@ -31,7 +32,7 @@ class SessionStorageTest extends TestCase
      *
      * @return void
      */
-    public function setUp()
+    public function setUp(): void
```
markstory: Do these help at all? I don't see return hints helping much on test methods.

dereuromark: They also dont hurt.

ADmad: Wasted effort does hurt :)

dereuromark: Thats why you use automated sniffers. Always.

cleptric: They were added by an automated sniff :) I just wanted to keep it consistent

dereuromark: Exactly, consistency across the code base is key.

markstory: If they can be autofixed then ok. I know I will forget to add these all the time :blush:

cleptric:
Test are failing because of the View propertie changes. I guess we can merge this one now.

PR #12192, opened 2018-06-06
Remove more deprecated features from 3.x

markstory:
Refs #11934


ADmad:
Tests failing in View layer.

I suggest you revert the changes done in view layer classes instead of spending time updating the failing tests. My on going work to make various view layer properties protected in 3.7 will take care of those. So just wait for the changes to be done and merged in 4.x.

PR #12196, opened 2018-06-06
The $order param allows both a string and array

wrdx:
Since the `field()` function parses the `$order` param to the `find()` function, the parameter supports both a `string` and an `array`. The documentation should properly reflect this.


markstory:
Thank you 👍

PR #12197, opened 2018-06-06
3.6 - deprecationWarning - PaginatorHelper::generateUrl()

saeideng:


src/View/Helper/PaginatorHelper.php
```diff
@@ -495,13 +495,21 @@ public function sort($key, $title = null, array $options = [])
...
      * @param array|bool $urlOptions Array of options or bool `fullBase` for BC reasons.
+     * The bool version of this argument is *deprecated* and will be removed in 4.0.0
      * @return string By default, returns a full pagination URL string for use in non-standard contexts (i.e. JavaScript)
      * @link https://book.cakephp.org/3.0/en/views/helpers/paginator.html#generating-pagination-urls
      */
     public function generateUrl(array $options = [], $model = null, $urlOptions = false)
     {
         if (!is_array($urlOptions)) {
             $urlOptions = ['fullBase' => $urlOptions];
+            if (func_num_args() == 3) {
```
dereuromark: `===`

src/View/Helper/PaginatorHelper.php
```diff
@@ -495,13 +495,21 @@ public function sort($key, $title = null, array $options = [])
...
+                deprecationWarning(
+                    'Passing an bool as third argument into PaginatorHelper::generateUrl() is deprecated ' .
```
dereuromark: no "an"
or `Passing boolean value`

src/View/Helper/PaginatorHelper.php
ADmad: Don't really like this check. I would rather change the default value to array in 3.7 and throw warning when a bool is passed and make required conversion.

ADmad (approved):
This looks much cleaner than previous patch.

Issue #10982, opened 2017-08-02
Discrepancy in the Cookie's default path


jorisvaesen:
This is a (multiple allowed):

* [ ] bug
* [x] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.5.0-RC1

The default `path` parameters differs between the Response's `withCookie()` function and the new Cookie object.

`Response::withCookie()`
https://github.com/cakephp/cakephp/blob/3.next/src/Http/Response.php#L2020

Cookie constructor
https://github.com/cakephp/cakephp/blob/3.next/src/Http/Cookie/Cookie.php#L128

This means that cookies differ when used like this:
```php
$response->withCookie(new Cookie('test')); // path = ''
$response->withCookie('test'); // path = '/'
```

ADmad:
There should also be a convenient way to specify default path for cookies. The CookieComponent uses `$request->webroot` as default path. There be an easy way to achieve same with new cookies implementation.

jorisvaesen:
@ADmad I agree, but how do we get this value as default in the Response and Cookie object?

markstory:
If you're making a cookie object you are responsible for making it properly. I don't think the Response should be changing the shape of cookie objects it is given.

markstory:
@ADmad Doesn't `withCookie('foo', 'bar')` make it 'easy' enough? The `withCookie()` method should probably default cookie paths to `$this->webroot` though.

ADmad:
@markstory If `withCookie()` defaults to using `$request->webroot` then it's good enough.

markstory:
Ok I can make that change.

markstory:
I was thinking about this some more, and I don't know how we get the `request->webroot` value into the response and then into the cookies.

jorisvaesen:
Was thinking exactly the same (as my comment above). Having the request object in the response object just to get the webroot to apply to a cookie doesn't seem like an elegant solution.

markstory:
No. The current behavior is also consistent with how `Response::cookie()` has worked for a number of years.

Schlaefer:
Would it be possible to move the Cookie object to array conversion from the ResponseEmitter to a  middleware and access the request there?

markstory:
That is an option. Wouldn't that require everyone to change their applications around?

Schlaefer:
> Wouldn't that require everyone to change their applications around?

Yes. Should probably go on a consideration list for a breaking release.

[...]

PR #12198, opened 2018-06-06
Deprecate the Response being able to create new Cookies

davidyell:
References: #10982

# Aim
The aim of this pull request is to deprecate the `\Cake\Http\Response` objects responsibility for creating new `\Cake\Http\Cookie\Cookie` class instances.

# Changes
To this end I have implemented a deprecated warning to advise developers that passing array data into the methods is deprecated in favour of passing `\Cake\Http\Cookie\Cookie` instances.

I have also set a default path of `/` on Cookies where the path is empty. This is inline with my understanding of [RFC6252 Section 5.1.4 point 2](https://tools.ietf.org/html/rfc6265#section-5.1.4) .

> If the uri-path is empty or if the first character of the uri-path is not a %x2F ("/") character, output %x2F ("/") and skip the remaining steps.

# Further work
I don't know if there is any desire to implement the RFC's algorithm for creating the path in the 4.0 development branch.

# Known issues
Travis will probably not be happy. 💥 

This pull request adds deprecated warnings which are thrown for a significant number of tests. I have already updated the tests which this change touches directly. If there is any desire, I can submit another pull request to update the test suite to implement the new method signature into the other tests.

[I've also had to add a sidestep into some test cases](https://github.com/cakephp/cakephp/pull/12198/files#diff-4599d57762727ecd43517492301ce1daR1667). I think this is because the `Response:: convertCookieToArray` method is preserving backwards compatibility, and is protected. I figured this sidestep was less confusing than reflecting the class for the sake of the test case. I would happily accept any insight on this bump in the road.

I wanted to keep this pull request specific, and easy to review. If you would like me to add the test updates to this pull request please let me know.

### DIFF
--- a/tests/TestCase/Http/ResponseTest.php
+++ b/tests/TestCase/Http/ResponseTest.php
@@ -1773,7 +1773,7 @@ class ResponseTest extends TestCase
                 'name' => 'urmc',
                 'value' => '{"user_id":1,"token":"abc123"}',
                 'expire' => null,
-                'path' => '',
+                'path' => '/',
                 'domain' => '',
                 'secure' => false,
                 'httpOnly' => true
--- a/tests/TestCase/Http/ResponseTest.php
+++ b/tests/TestCase/Http/ResponseTest.php
@@ -16,12 +16,14 @@ namespace Cake\Test\TestCase\Http;
 
 include_once CORE_TEST_CASES . DS . 'Http' . DS . 'server_mocks.php';
 
+use Cake\Chronos\Chronos;
 use Cake\Http\Cookie\Cookie;
 use Cake\Http\Cookie\CookieCollection;
 use Cake\Http\CorsBuilder;
 use Cake\Http\Exception\NotFoundException;
 use Cake\Http\Response;
 use Cake\Http\ServerRequest;
+use Cake\I18n\FrozenTime;
 use Cake\TestSuite\TestCase;
 use Zend\Diactoros\Stream;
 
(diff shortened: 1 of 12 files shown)


markstory:
I think the changes look good. The change in default path could be problematic is an acceptable amount of change for a minor release.

> This pull request adds deprecated warnings which are thrown for a significant number of tests. I have already updated the tests which this change touches directly. If there is any desire, I can submit another pull request to update the test suite to implement the new method signature into the other tests.

You'll likely need to update all of the internal usage of cookies as well before this can be merged.

davidyell:
I will update the tests and other code, and add that to this pull request.

tests/TestCase/Http/ResponseTest.php
```diff
@@ -1695,18 +1710,30 @@ public function testWithExpiredCookieOptions()
             'path' => '/custompath/',
             'secure' => true,
             'httpOnly' => true,
-            'expire' => (string)strtotime('+14 days'),
+            'expire' => new \DateTimeImmutable('+14 days'),
```
davidyell: Which DateTime class should this be using? I find the DateTime classes a bit confusing. Should this be `\Cake\I18n\Time`, `Chronos`, `FrozenTime` or `DateTimeImmutable` ?

markstory: `DateTimeImmutable` is fine here.

davidyell:
Thanks @markstory 👍

PR #12200, opened 2018-06-07
Move start/end calculation into datasource.

### LINKED ISSUE #11707: CakePHP 3.x Paginator Component Datasource missing range start and end keys
I am fetching paginated data with Json API request from latest 3.5..

I added paging params so I can use them on a react native app..

    public function index()
    {
        $visits = $this->paginate($this->Visits);
        $paging = $this->request->getParam('paging');
        $results = [
            'visits' => $visits,
            'paging' => $paging,
        ];
        $this->set(compact('results'));
        $this->set('_serialize', 'results');
    }

But $this->request->getParam('paging'); doesn't set "start" and "end" metadata, like: (starting on record 21, ending on 40)

{{start}} - number of the first record being displayed.
{{end}} - number of the last record being displayed.

this code from the paginator helper could be implemented with minimal changes to paginator datasource:
https://github.com/cakephp/cakephp/blob/master/src/View/Helper/PaginatorHelper.php#L689-L696

    Datasource\Paginator.php

    public function paginate($object, array $params = [], array $settings = [])
    ......

        $start = 0;
        if ($count >= 1) {
            $start = (($page - 1) * $limit) + 1;
        }

        $end = $start + $limit - 1;
        if ($count < $end) {
            $end = $count;
        }

        $paging = [
            ........
            'start' => $start,
            'end' => $end,
            ........
        ];

Thanks,

Andras

dereuromark:
Resolves https://github.com/cakephp/cakephp/issues/11707 as it moves the calculation into Datasource for better usability outside of view/helper layer and just re-uses data there.

During the tests adjustment I noticed some very strange test expectations however.
So I fixed the values up to make sense.

### DIFF
--- a/src/Datasource/Paginator.php
+++ b/src/Datasource/Paginator.php
@@ -200,12 +200,23 @@ class Paginator implements PaginatorInterface
             $directionDefault = current($defaults['order']);
         }
 
+        $start = 0;
+        if ($count >= 1) {
+            $start = (($page - 1) * $limit) + 1;
+        }
+        $end = $start + $limit - 1;
+        if ($count < $end) {
+            $end = $count;
+        }
+
         $paging = [
             'finder' => $finder,
             'page' => $page,
             'current' => $numResults,
             'count' => $count,
             'perPage' => $limit,
+            'start' => $start,
+            'end' => $end,
             'prevPage' => $page > 1,
             'nextPage' => $count > ($page * $limit),
             'pageCount' => $pageCount,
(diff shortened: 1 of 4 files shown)


lorenzo:
Great idea!

dereuromark:
The PHPCS fails are unrelated.

markstory:
Those PHPCS errors are very strange. It appears that PHPCS is interpreting return types as constants.

dereuromark:
Could be a regression in their latest release? https://github.com/squizlabs/PHP_CodeSniffer/releases/tag/3.3.0

markstory:
The sniff emitting those errors is `CakePHP.NamingConventions.UpperCaseConstantName.ConstantNotUpperCase`. I'll see if I can make sense of whats going on. There seem to have been several tokenization changes in PHPCS 3.3.0.

PR #12201, opened 2018-06-07
3.7 cell deprecations

ADmad:
* `template`, `plugin`, `helpers` properties have been removed. Having them was redundant since they can be got/set through cell's view builder.
* `View`, `request`, `response`, `action`, `args` have been made protected. I see no reason for these to be accessible from outside the cell class. Hence no getter/setter have been provided for them.

### DIFF
--- a/src/View/Cell.php
+++ b/src/View/Cell.php
@@ -299,8 +299,7 @@ abstract class Cell
         if (in_array($name, $protected, true)) {
             deprecationWarning(sprintf(
                 'Cell::$%s is now protected and shouldn\'t be accessed from outside a child class.',
-                $name,
-                $method
+                $name
             ));
         }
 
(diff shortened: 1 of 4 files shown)


markstory:
There are some [relevant looking phpstan warnings](https://travis-ci.org/cakephp/cakephp/jobs/389225300#L689)

ADmad:
@markstory Fixed.

src/View/Cell.php
```diff
@@ -302,12 +357,12 @@ public function __toString()
     public function __debugInfo()
     {
         return [
-            'plugin' => $this->plugin,
             'action' => $this->action,
             'args' => $this->args,
-            'template' => $this->template,
+            'viewClass' => $this->viewClass,
```
markstory: Do Cells have viewClass anymore? I removed it in #12187

ADmad: `Cell` uses `ViewVarsTrait` and there's `ViewVarsTrait::$viewVars` property.

ADmad: But `ViewVarsTrait::$viewClass` is deprecated, so I'll remove it. Must have snuck in when I merged `3.next` into this branch.

ADmad:
@markstory Anything remaining here? I see that the 3.7 migration guide is docs has also been updated.

PR #12203, opened 2018-06-07
Add type hints and strict types to Controller

cleptric:
Refs: #11935 

EDIT: I slipped in one addition from the `Core` package 😇 
`InstanceConfigTrait:: getConfig `


src/Controller/Component/AuthComponent.php
```diff
@@ -403,10 +406,10 @@ protected function _isLoginAction(Controller $controller)
      * Handle unauthorized access attempt
      *
      * @param \Cake\Controller\Controller $controller A reference to the controller object
-     * @return \Cake\Http\Response
+     * @return \Cake\Http\Response|null
      * @throws \Cake\Http\Exception\ForbiddenException
      */
-    protected function _unauthorized(Controller $controller)
+    protected function _unauthorized(Controller $controller): ?Response
```
markstory: Does this need to nullable if the method sometimes throws an exception?

src/Controller/Component/SecurityComponent.php
```diff
@@ -175,7 +176,7 @@ public function blackHole(Controller $controller, $error = '', SecurityException
      * @throws \Cake\Http\Exception\BadRequestException
      * @return void
      */
-    protected function _throwException($exception = null)
+    protected function _throwException(?\Cake\Controller\Exception\SecurityException $exception = null): void
     {
```
markstory: Are we able to use the imported class name that isn't fully qualified?

dereuromark: We should always have the FQN in the use statement IMO.

markstory:
You might need to use `phpcs.xml.dist` for stickler. There isn't a phpcs.xml file in the repository.

cleptric:
@markstory ... sorry

src/Controller/Component/AuthComponent.php
dereuromark: No, it should always be either Response, or the exception to bail early.

src/Controller/Component/SecurityComponent.php
cleptric: Definitely, I overlooked this one.

src/Controller/Component/AuthComponent.php
```diff
@@ -817,7 +821,7 @@ public function constructAuthenticate()
                 $class = $alias;
             }
             $className = App::className($class, 'Auth', 'Authenticate');
-            if (!class_exists($className)) {
+            if ($className === false || !class_exists($className)) {
```
dereuromark: Dont u need the same in line 865 too?

src/Controller/Component/AuthComponent.php
```diff
@@ -817,7 +821,7 @@ public function constructAuthenticate()
...
                 throw new Exception(sprintf('Authentication adapter "%s" was not found.', $class));
             }
             if (!method_exists($className, 'authenticate')) {
```
dereuromark: Do we need to keep this check in 4.x+? Maybe we can also use some interface approach in the future.

src/Controller/Component/AuthComponent.php
```diff
@@ -403,10 +406,10 @@ protected function _isLoginAction(Controller $controller)
      * Handle unauthorized access attempt
      *
      * @param \Cake\Controller\Controller $controller A reference to the controller object
-     * @return \Cake\Http\Response
+     * @return \Cake\Http\Response|null
```
dereuromark: Must be reverted.

tests/test_app/TestApp/Controller/PostsController.php
```diff
@@ -34,7 +35,7 @@ public function initialize()
      *
      * @return void
```
dereuromark: should then be Response|null as usual

src/Controller/Controller.php
```diff
@@ -178,7 +179,7 @@ class Controller implements EventListenerInterface, EventDispatcherInterface
      * @param \Cake\Event\EventManager|null $eventManager The event manager. Defaults to a new instance.
      * @param \Cake\Controller\ComponentRegistry|null $components The component registry. Defaults to a new instance.
      */
-    public function __construct(ServerRequest $request = null, Response $response = null, $name = null, $eventManager = null, $components = null)
+    public function __construct(ServerRequest $request = null, Response $response = null, ?string $name = null, ?\Cake\Event\EventManager $eventManager = null, ?\Cake\Controller\ComponentRegistry $components = null)
```
ADmad: Need to add `use` statements instead of using FQCNs. Also `EventManager` needs to be changed to `EventManagerInterface`.

src/Controller/Controller.php
```diff
@@ -249,7 +250,7 @@ public function initialize()
      *
      * @return \Cake\Controller\ComponentRegistry
      */
-    public function components($components = null)
+    public function components(?\Cake\Controller\ComponentRegistry $components = null): ComponentRegistry
```
ADmad: Need to add `use` statements instead of using FQCN.

tests/TestCase/Controller/ControllerTest.php
```diff
@@ -632,7 +632,7 @@ public function testSetAction(): void
      */
     public function testStartupProcess(): void
     {
-        $eventManager = $this->getMockBuilder('Cake\Event\EventManager')->getMock();
+        $eventManager = $this->getMockBuilder('Cake\Event\EventManagerInterface')->getMock();
```
dereuromark: We can use ::class usage here too for better refactoring.

src/Controller/Component/RequestHandlerComponent.php
```diff
@@ -121,7 +122,7 @@ public function implementedEvents()
      * @param \Cake\Http\Response $response The response instance.
      * @return void
      */
-    protected function _setExtension($request, $response)
+    protected function _setExtension(\Cake\Http\ServerRequest $request, \Cake\Http\Response $response): void
```
dereuromark: You can use my FQCN to use statement sniff if you dont want to do that manually :)

src/Controller/Component/RequestHandlerComponent.php
```diff
@@ -238,7 +239,7 @@ public function convertXml($xml)
      *   set that Content-type in the response header.
      *
      * @param \Cake\Event\EventInterface $event The Controller.beforeRender event.
-     * @return bool false if the render process should be aborted
+     * @return bool|void false if the render process should be aborted
```
dereuromark: Response|null actually

src/Controller/ComponentRegistry.php
```diff
@@ -52,9 +54,9 @@ public function __construct(Controller $controller = null)
     /**
      * Get the controller associated with the collection.
      *
-     * @return \Cake\Controller\Controller Controller instance
+     * @return \Cake\Controller\Controller|null Controller instance
      */
-    public function getController()
+    public function getController(): ?Controller
```
dereuromark: Should this instead throw an exception if getController() cannot return sth?

src/Controller/Component/RequestHandlerComponent.php
dereuromark: Fairy certain that is a dead relict from 2.x times.
Just void should be fine for these.

src/Controller/Component/AuthComponent.php
ADmad: AuthComponent and related classes themselves will go away in a future release, so not worth the effort.

src/Controller/Component/RequestHandlerComponent.php
ADmad: It's not a relic. If a listener callback returns `false` the event manager stops event propagation.  `RequestHandler::beforeRender()` checks the last modified headers and if needed stops the event propagation to prevent view rendering and just return response with appropriate headers.

The `return false` could however be replaced with `$event->stopPropgation()` to achieve the same result.

src/Controller/Component/AuthComponent.php
cleptric: `Controller::redirect` could return `null` so this should be fine

src/Controller/Component/AuthComponent.php
cleptric: @ADmad Done, let me know if it's ok this way :)

cleptric:
`TypeError: explode() expects parameter 2 to be string, null given`

Looks like we need to add the content type to the `IntegrationTestCase`.
Or would a `list($contentType) = explode(';', $request->contentType() ?? '')`, as it was basically this before adding strict types.

tests/test_app/TestApp/Controller/AuthTestController.php
```diff
@@ -41,7 +42,7 @@ public function __construct($request = null, $response = null)
         parent::__construct($request, $response);
     }
 
-    public function initialize()
+    public function initialize(): void
```
markstory: Is updating the various hook methods in an application going to add value or frustration? I left return typehints off of the hook methods in `Shell` and `Command` as I didn't see them adding much value to a developer's life, but they certainly are a big bump when upgrading.

src/Controller/Component/AuthComponent.php
dereuromark: Why can it? It should either return a response or throw an exception, right?

tests/test_app/TestApp/Controller/AuthTestController.php
dereuromark: I also see your argument.
But: Using automated sniffers/fixers (existing), it is rather easy to fix up project level or plugins at once. 

The benefit of having a consistent code base is a requirement for this, though. If only some methods had typehinting the sniffer would fix too much and then you run into fatal errors as signature changes.

Also: The benefit for hooks is especially important (since they cannot be IDE completed/validated). People might try to return a response object or other weird things. So having a clear communicated API for them seems very useful.

src/Controller/Component/AuthComponent.php
cleptric: Because of `Controller::redirect():583` 

```
if ($event->isStopped()) {
  return null;
}
```

dereuromark: OK

tests/test_app/TestApp/Controller/AuthTestController.php
markstory: > The benefit of having a consistent code base is a requirement for this, though. If only some methods had typehinting the sniffer would fix too much and then you run into fatal errors as signature to parent is not compatible.

Sub-classes can add additional type restrictions that the parent class doesn't have in general.

> People might try to return a response object or other weird things. So having a clear communicated API for them seems very useful.

This could be a valuable point of having more types on hooks. It makes the framework's expectations of userland code clear. My concern is primarily around the initial upgrade and how bumpy that is going to be.

dereuromark: Signature changes should be regex fixable etc.

cleptric:
I fixed the conflict. Should be good to go

src/Controller/ComponentRegistry.php
dereuromark: OK, we can allow this to be nullable I guess.

PR #12206, opened 2018-06-08
3.7 helper deprecations

ADmad:
- Removed `Helper::$theme` and `Helper::$plugin` then can be accessed through their respective getter/setter methods of `View` class.
- `Helper::$fieldset` and `Helper::$tags` are unsed, hence marked them and deprecated and removed from debug info.
- Made `Helper::$helpers` protected. I don't see the need from it to be accessible outside the class so haven't added any getter/setter for it.
- Removed `Helper::$request`. `View::getRequest()/seRequest()` can be used instead.

### DIFF
--- a/src/View/Helper.php
+++ b/src/View/Helper.php
@@ -175,7 +175,7 @@ class Helper implements EventListenerInterface
     public function __set($name, $value)
     {
         $removed = [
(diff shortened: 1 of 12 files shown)


ADmad:
@markstory Ping

Issue #10841, opened 2017-06-30
Cannot issue HTTPS request via Proxy using HttpClient


Antoniossss:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

Cakephp 3.4.5

```
config: 
            "proxy" => [
                "proxy" => "pl3.nordvpn.com:80"
                "username" => "xx",
                "password" => "xx"
            ]

$client=new HttpClient($config);
$client->get('http://api.ipify.org'); // works, returns proxied ip
$client->get('https://api.ipify.org'); // does not work 
```

This is respone object for https call
```

object(Cake\Http\Client\Response) {
	[protected] code => (int) 400
	[protected] reasonPhrase => 'Bad Request'
	[protected] _xml => null
	[protected] _json => null
	[protected] _exposedProperties => [
		'cookies' => '_cookies',
		'body' => '_getBody',
		'code' => 'code',
		'json' => '_getJson',
		'xml' => '_getXml',
		'headers' => '_getHeaders'
	]
	[protected] _cookies => []
	[protected] _body => null
	[protected] headers => [
		'Connection' => [
			(int) 0 => 'close'
		],
		'Server' => [
			(int) 0 => 'Cowboy'
		],
		'Date' => [
			(int) 0 => 'Fri, 30 Jun 2017 12:21:45 GMT'
		],
		'Content-Length' => [
			(int) 0 => '0'
		]
	]
	[protected] headerNames => [
		'connection' => 'Connection',
		'server' => 'Server',
		'date' => 'Date',
		'content-length' => 'Content-Length'
	]
	[private] protocol => '1.1'
	[private] stream => object(Zend\Diactoros\Stream) {
		[protected] resource => resource
		[protected] stream => 'php://memory'
	}
}
```
Response body is empty.

markstory:
That looks like a valid response. Is that 400 coming from the proxy or the destination server?

Antoniossss:
I have no idea. This response object is from
`$response=$client->get('https://api.ipify.org');`

It should be code 200 with ip string in body (like for `http://api.ipify.org` call)

markstory:
There isn't really a way to trace http requests made by the client. We rely on php streams to make the actual HTTP requests.

yomexzo:
@Antoniossss, did you make the request outside CakePHP? just to confirm the problem is from Cake.

markstory:
[...]

PR #12209, opened 2018-06-10
3.7 - Add curl based Http\Client Adapter.

markstory:
The stream based client has a few limitations around `fopen_url` settings, and proxy issues (#10841 ) that are not currently solvable. By adding a cURL based adapter we can default to using this extension which solves these problems and allows more feature support in the future.

### TODO

* [x] Verify results on timeout expiration.

### DIFF
--- a/src/Http/Client/Adapter/Curl.php
+++ b/src/Http/Client/Adapter/Curl.php
@@ -27,6 +27,74 @@ class Curl implements AdapterInterface
      */
     public function send(Request $request, array $options)
     {
-        // Do stuff here.
+        $ch = curl_init();
+        $options = $this->buildOptions($request, $options);
+        curl_setopt_array($ch, $options);
+
+        $bodyBuffer = tmpfile();
+        curl_setopt(CURLOPT_FILE, $bodyBuffer);
+
+        $this->exec($ch);
+        curl_close($ch);
+    }
+
+    /**
+     * Convert client options into curl options.
+     *
+     * @param \Cake\Http\Client\Request $request The request.
+     * @param array $options The client options
+     * @return array
+     */
+    public function buildOptions(Request $request, array $options)
+    {
+        $headers = [];
+        foreach ($request->getHeaders() as $key => $values) {
+            $headers[] = $key . ': ' . implode(', ', $values);
+        }
+
+        $out = [
+            CURLOPT_URL => (string)$request->getUri(),
+            CURLOPT_HTTP_VERSION => $request->getProtocolVersion(),
+            CURLOPT_RETURNTRANSFER => true,
+            CURLOPT_HEADER => true,
+            CURLOPT_HTTPHEADER => $headers
+        ];
+        switch ($request->getMethod()) {
+            case Request::METHOD_GET:
+                $out[CURLOPT_HTTPGET] = true;
+                break;
+
+            case Request::METHOD_POST:
+                $out[CURLOPT_POST] = true;
+                break;
+
+            default:
+                $out[CURLOPT_POST] = true;
+                $out[CURLOPT_CUSTOMREQUEST] = $request->getMethod();
+                break;
+        }
+
+        $body = $request->getBody();
+        if ($body) {
+            $body->rewind();
+            $out[CURLOPT_POSTFIELDS] = $body->getContents();
+        }
+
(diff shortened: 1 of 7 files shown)


src/Http/Client/Adapter/Curl.php
```diff
@@ -0,0 +1,143 @@
...
+        }
+
+        if (empty($options['ssl_cafile'])) {
+            $options['ssl_cafile'] = CORE_PATH . 'config' . DIRECTORY_SEPARATOR . 'cacert.pem';
+        }
+        $optionMap = [
+            'timeout' => CURLOPT_TIMEOUT,
+            'ssl_verify_peer' => CURLOPT_SSL_VERIFYPEER,
+            'ssl_verify_host' => CURLOPT_SSL_VERIFYHOST,
+            'ssl_verify_status' => CURLOPT_SSL_VERIFYSTATUS,
```
ADmad: Seems travis has an older Curl version and `CURLOPT_SSL_VERIFYSTATUS` is unavailable.
https://travis-ci.org/cakephp/cakephp/jobs/390291411#L770

PHP Manual says:
> Added in cURL 7.41.0. Available since PHP 7.0.7.

src/Http/Client/Adapter/Curl.php
```diff
@@ -0,0 +1,143 @@
...
+        ];
+        foreach ($optionMap as $option => $curlOpt) {
+            if (isset($options[$option])) {
+                $out[$curlOpt] = $options[$option];
+            }
+        }
+        if (isset($options['proxy']['proxy'])) {
+            $out[CURLOPT_PROXY] = $options['proxy']['proxy'];
+        }
+
```
lorenzo: Setting extra curl opta is not uncommon. Should we have an extra key in he array for custom opts?

src/Http/Client/Adapter/Curl.php
```diff
@@ -0,0 +1,143 @@
...
+        return $out;
+    }
+
+    /**
+     * Convert the raw curl response into an Http\Client\Response
+     *
+     * @param resource $handle Curl handle
+     * @param string $responseData string The response data from curl_exec
+     * @return \Cake\Http\Client\Response[]
```
lorenzo: How can you get more than one response?

markstory: You can't. I'll fix this.

src/Http/Client/Adapter/Curl.php
markstory: Perhaps this should be moved into an additional option like @lorenzo suggested?

src/Http/Client/Adapter/Curl.php
markstory: Would `$options['curl']` work as a place to put the additional curl options?

lorenzo: Yeah, that sounds good to me

markstory:
Arbitrary curl option support added.

BassemN:
Thanks allot @markstory & @ADmad for implementing curl. This is essential for all modern apps.

ADmad:
@BassemN Thanks for reporting the issue (on my plugin) :slightly_smiling_face:.

src/Http/Client/AdapterInterface.php
```diff
@@ -0,0 +1,28 @@
...
+use Cake\Http\Client\Request;
+
+interface AdapterInterface
+{
+    /**
+     * Send a request and get a response back.
+     *
+     * @param \Cake\Http\Client\Request $request The request object to send.
+     * @param array $options Array of options for the stream.
+     * @return \Cake\Http\Client\Response[] Array of populated Response objects
```
lorenzo: I think this returns a single Response, not many

src/Http/Client/Adapter/Curl.php
```diff
@@ -0,0 +1,161 @@
...
+
+        return $out;
+    }
+
+    /**
+     * Convert the raw curl response into an Http\Client\Response
+     *
+     * @param resource $handle Curl handle
+     * @param string $responseData string The response data from curl_exec
+     * @return \Cake\Http\Client\Response
```
lorenzo: The type here says it is a single response, but for some reason an array is returned at the end

PR #12213, opened 2018-06-11
Casting all integration test data to string

jeremyharris:
This fixes a bug where data passed to methods in IntegrationTestCase were directly passed to the request. In an actual HTTP context everything is passed as a string. This fix prevents false tests such as this:

```php
// controller action
public function add()
{
    if ($this->request->getData('status') === 1) {
        // do stuff
        // this would never be reached in a true HTTP context
    }
}

// test case
public function testAdd()
{
    $this->post('/action', [
        'status' => 1
    ]);
    // assert did stuff
}
```

This leaves file arrays alone as they are created by PHP and have `int` values on certain keys. It also leaves classes alone so testing with things like `UploadedFile` is still possible.

I can PR against 3.next as well to help prevent extra merge work if you'd like.

### DIFF
--- a/src/TestSuite/IntegrationTestCase.php
+++ b/src/TestSuite/IntegrationTestCase.php
@@ -672,7 +672,7 @@ abstract class IntegrationTestCase extends TestCase
     {
         foreach ($data as $key => $value) {
             if (is_scalar($value)) {
-                $data[$key] = (string)$value;
+                $data[$key] = $value === false ? '0' : (string)$value;
 
                 continue;
             }
--- a/src/TestSuite/IntegrationTestCase.php
+++ b/src/TestSuite/IntegrationTestCase.php
@@ -679,9 +679,11 @@ abstract class IntegrationTestCase extends TestCase
 
             if (is_array($value)) {
                 $looksLikeFile = isset($value['error']) && isset($value['tmp_name']) && isset($value['size']);
-                if (!$looksLikeFile) {
-                    $data[$key] = $this->_castToString($value);
+                if ($looksLikeFile) {
+                    continue;
                 }
+
+                $data[$key] = $this->_castToString($value);
             }
         }
 
--- a/src/TestSuite/IntegrationTestCase.php
+++ b/src/TestSuite/IntegrationTestCase.php
@@ -605,7 +605,8 @@ abstract class IntegrationTestCase extends TestCase
             $props['input'] = $data;
         }
         if (!isset($props['input'])) {
-            $props['post'] = $this->_addTokens($tokenUrl, $data);
+            $data = $this->_addTokens($tokenUrl, $data);
+            $props['post'] = $this->_castToString($data);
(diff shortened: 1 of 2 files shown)


lorenzo (approved):

src/TestSuite/IntegrationTestCase.php
```diff
@@ -660,6 +661,33 @@ protected function _addTokens($url, $data)
...
+        foreach ($data as $key => $value) {
+            if (is_scalar($value)) {
+                $data[$key] = (string)$value;
+
+                continue;
+            }
+
+            if (is_array($value)) {
+                $looksLikeFile = isset($value['error']) && isset($value['tmp_name']) && isset($value['size']);
+                if (!$looksLikeFile) {
```
dereuromark: I am usually a fan of return early to keep nesting to a minimum. But no big deal here.

ADmad: In this case you need to continue the loop not return from the method. The loop is going to continue anyway after this check, so continuing early here won't have much benefit.

dereuromark: Yes, thats why "no big deal" :)

jeremyharris: Already done :P

markstory:
> I can PR against 3.next as well to help prevent extra merge work if you'd like.

This isn't necessary, the merges aren't too tricky to do yet.

src/TestSuite/IntegrationTestCase.php
```diff
@@ -660,6 +661,35 @@ protected function _addTokens($url, $data)
...
+     * the real world
+     *
+     * @param array $data POST data
+     * @return array
+     */
+    protected function _castToString($data)
+    {
+        foreach ($data as $key => $value) {
+            if (is_scalar($value)) {
+                $data[$key] = (string)$value;
```
markstory: Should `false` be cast to `0` instead of `''` ?

src/TestSuite/IntegrationTestCase.php
```diff
@@ -660,6 +661,35 @@ protected function _addTokens($url, $data)
...
+
+            if (is_array($value)) {
+                $looksLikeFile = isset($value['error']) && isset($value['tmp_name']) && isset($value['size']);
+                if ($looksLikeFile) {
+                    continue;
+                }
+
+                $data[$key] = $this->_castToString($value);
+            }
+        }
```
markstory: This might be a good place to use [array_walk_recursive](http://php.net/array_walk_recursive)

jeremyharris: I don't think that would work well given the check for a file array, as `array_walk_recursive` is unaware of its sibling keys afaik.

dereuromark: I think the code is readable as of now.

PR #12216, opened 2018-06-11
4.x Made FormHelper autoSetCustomValidity option true by default

jeremyharris:
Refs: https://github.com/cakephp/cakephp/pull/11837#issuecomment-374095619

Doc PR to follow.


ADmad:
Tests failing, you need to rebase your branch on top of upstream 4.x.

jeremyharris:
Hm I did that and there are still failures. It looks like the latest build for 4.x is failing

markstory:
Tests were failing on 4.x because I did a bad merge in https://github.com/cakephp/cakephp/commit/f865b9ccf00efb8f259735b8774208297b77353f. Thanks to @saeideng for pointing out the mistake.

tests/TestCase/View/Helper/FormHelperTest.php
```diff
@@ -147,7 +147,9 @@ public function setUp()
         static::setAppNamespace('Cake\Test\TestCase\View\Helper');
         $this->View = new View();
 
-        $this->Form = new FormHelper($this->View);
+        $this->Form = new FormHelper($this->View, [
+            'autoSetCustomValidity' => false,
+        ]);
```
markstory: Should we have a few more tests that do cover this feature if we're going to enable it for all users?

jeremyharris: Sure, what other tests would you like to see? I disabled it here because updating all the existing tests seemed tedious, but I can do that too if you like.

markstory: I feel like there should be a test that covers each input type and some associations.

ADmad:
@jeremyharris Can you look into the test failures and wrap this up? :slightly_smiling_face:

ADmad:
Closing in favor of #12450

jeremyharris:
Sorry I didn't see this message, I've been taking a break for the last few months. Thanks for picking up my remnants, @ADmad

PR #12224, opened 2018-06-13
4.x cleanup

ADmad:


src/View/View.php
```diff
@@ -182,14 +181,6 @@ class View implements EventDispatcherInterface
      */
     protected $hasRendered = false;
 
-    /**
-     * List of generated DOM UUIDs.
-     *
-     * @var array
-     * @deprecated 3.7.0 The property is unused and will be removed in 4.0.0.
-     */
-    public $uuids = [];
```
markstory: The `uuid()` method on `View` uses this property. I'm not confident we *need* that method either though.

dereuromark: If someone needs it they can easily wrap it using a helper method :)

dereuromark: I would much more like to see the model prefixing coming back for "id" values as the collisions on forms usually otherwise are always present somehow.

ADmad: @markstory I am pretty sure I removed the method too. Dunno how it didn't get committed :) Will update.

ADmad: `View::uuid()` removed.

dereuromark:
Also for a follow up PR: I can see 5 $this|... still in the code base, e.g.

     * @return $this|array
     */
    public function join($tables = null, $types = [], $overwrite = false)

and

     * @return $this|\Cake\Datasource\EntityInterface
     */
    public function translation($language)

etc
which we need to fix up into separate setter and getter calls etc.

src/Controller/Controller.php
```diff
@@ -211,20 +211,6 @@ public function __construct(ServerRequest $request = null, Response $response =
...
-                'Use $this->loadComponent() instead in initialize().'
-            );
-        }
-
-        if (isset($this->helpers)) {
-            triggerWarning(
-                'Support for loading helpers using $helpers property is removed. ' .
-                'Use $this->viewBuilder()->setHelpers() instead.'
-            );
-        }
```
markstory: I thought we were going to keep these to save folks time when they accidentally used the old properties.

dereuromark: => https://github.com/cakephp/cakephp/commit/d838f352401f84f501d30ae9b620b3c7314c66c2 added in 4.x

ADmad: Ah, I'll revert it.

ADmad:
Update as per feedback and fixed conflicts. Merging avoid more merge conflicts.

PR #12226, opened 2018-06-13
Deprecated View::uuid() as it's unused in core.

ADmad:
I missed out deprecating `View::uuid()` in earlier PR. I don't see any usage of it in the core, probably a relic of 2.x days.


dereuromark (approved):

ADmad:
Ah the uuid() method removal is in this PR :D

PR #12231, opened 2018-06-13
4.x: Remove more deprecations, cleaner typehinting for Paginator.

dereuromark:
Follow up on https://github.com/cakephp/cakephp/pull/12224

Resolves one of the cleanup tasks of https://github.com/cakephp/cakephp/wiki/4.0-Roadmap

### DIFF
--- a/src/View/Helper/PaginatorHelper.php
+++ b/src/View/Helper/PaginatorHelper.php
@@ -19,6 +19,7 @@ use Cake\Utility\Inflector;
 use Cake\View\Helper;
 use Cake\View\StringTemplateTrait;
 use Cake\View\View;
+use RuntimeException;
 
 /**
  * Pagination Helper class for easy generation of pagination links.
@@ -613,13 +614,18 @@ class PaginatorHelper extends Helper
     /**
      * Returns true if the given result set has the page number given by $page
      *
-     * @param string|null $model Optional model name. Uses the default if none is specified.
      * @param int $page The page number - if not set defaults to 1.
+     * @param string|null $model Optional model name. Uses the default if none is specified.
      * @return bool True if the given result set has the specified page number.
      * @link https://book.cakephp.org/3.0/en/views/helpers/paginator.html#checking-the-pagination-state
+     * @throws \RuntimeException
      */
-    public function hasPage($model = null, $page = 1)
+    public function hasPage($page = 1, $model = null)
     {
+        if (!is_numeric($page)) {
+            throw new RuntimeException('First argument "page" has to be int. Note that argument order switched from 3.x to 4.x.');
+        }
+
         $paging = $this->params($model);
         if ($paging === []) {
             return false;
--- a/src/View/Helper/PaginatorHelper.php
+++ b/src/View/Helper/PaginatorHelper.php
@@ -671,26 +671,25 @@ class PaginatorHelper extends Helper
     }
 
     /**
-     * Returns a counter string for the paged result set
+     * Returns a counter string for the paged result set.
      *
      * ### Options
      *
      * - `model` The model to use, defaults to PaginatorHelper::defaultModel();
-     * - `format` The format string you want to use, defaults to 'pages' Which generates output like '1 of 5'
+     *
+     * @param string $format The format string you want to use, defaults to 'pages' Which generates output like '1 of 5'
      *    set to 'range' to generate output like '1 - 3 of 13'. Can also be set to a custom string, containing
      *    the following placeholders `{{page}}`, `{{pages}}`, `{{current}}`, `{{count}}`, `{{model}}`, `{{start}}`, `{{end}}` and any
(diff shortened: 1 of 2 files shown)


tests/TestCase/View/Helper/PaginatorHelperTest.php
```diff
@@ -2833,7 +2833,7 @@ public function testHasPage()
         $result = $this->Paginator->hasPage('Article', 2);
         $this->assertTrue($result);
 
-        $result = $this->Paginator->hasPage(2);
+        $result = $this->Paginator->hasPage(null, 2);
```
markstory: This change makes the standard use case (single model pagination) more clunky. This feels like a mixed type we should keep.

dereuromark: But it is completely undocumented so far.

And this now in sync with all the other methods that have the model as first key:
- hasPrev($model = null)
- hasNext($model = null)
- sortKey($model = null
- sortDir($model = null

If anything, the order should be inverse: `$page, $model = null` That would make way more sense :)
Similar to `public function param($key, $model = null)`

markstory: > Can we switch the order with some migration guide?

If we were going to make a change now is the time to do so.

dereuromark:
Cool, I switched the order and added a defensive exception with a clear message until we can fully add scalar typehinting to methods.

src/View/Helper/PaginatorHelper.php
```diff
@@ -613,13 +614,18 @@ public function hasNext($model = null)
...
+     * @param string|null $model Optional model name. Uses the default if none is specified.
      * @return bool True if the given result set has the specified page number.
      * @link https://book.cakephp.org/3.0/en/views/helpers/paginator.html#checking-the-pagination-state
+     * @throws \RuntimeException
      */
-    public function hasPage($model = null, $page = 1)
+    public function hasPage($page = 1, $model = null)
     {
+        if (!is_numeric($page)) {
+            throw new RuntimeException('First argument "page" has to be int. Note that argument order switched from 3.x to 4.x.');
```
ADmad: Wouldn't InvalidArgumentException be more appropriate?

src/View/Helper/PaginatorHelper.php
```diff
@@ -613,17 +614,18 @@ public function hasNext($model = null)
...
      *
-     * @param string|null $model Optional model name. Uses the default if none is specified.
      * @param int $page The page number - if not set defaults to 1.
+     * @param string|null $model Optional model name. Uses the default if none is specified.
      * @return bool True if the given result set has the specified page number.
      * @link https://book.cakephp.org/3.0/en/views/helpers/paginator.html#checking-the-pagination-state
+     * @throws \InvalidArgumentException
      */
-    public function hasPage($model = null, $page = 1)
+    public function hasPage($page = 1, $model = null)
```
ADmad: No type hints for the arguments?

tests/TestCase/View/Helper/PaginatorHelperTest.php
```diff
@@ -2763,7 +2763,7 @@ public function testCounter()
 
         $expected = 'Page 1 of 5, showing 3 records out of 13 total, starting on record 1, ';
         $expected .= 'ending on 3';
-        $result = $this->Paginator->counter($input);
+        $result = $this->Paginator->counter(['format' => $input]);
```
ADmad: This method also seems like a candidate for splitting into 2 args `count($format, $options)`.

src/View/Helper/PaginatorHelper.php
dereuromark: I didn't know how fast we now progress into the 7.1 typehints.

ADmad: We have started adding typehints to take care of #11935. So I think it would be nice to add typehints to any argument changes we are doing.

markstory: Would the type errors give enough context for people using the old call signature?

dereuromark: Yes, most likely :)

dereuromark: @ADmad Feel free to add in these PRs. I just thought a more efficient way would be the sniffer run over the namespace.

markstory: We can add typehints to the helper namespace as a separate task. This is incremental progress :+1:

inoas: Is 7.1 still the target or more like 7.2?

dereuromark: why? 7.1 is still enough right now.

markstory: @inoas Yes 7.1 is still the target.

PR #12232, opened 2018-06-13
4.x: Cleaner typehinting for Folder.

dereuromark:
Further cleanup

### DIFF
--- a/tests/TestCase/Filesystem/FolderTest.php
+++ b/tests/TestCase/Filesystem/FolderTest.php
@@ -277,11 +277,11 @@ class FolderTest extends TestCase
         $this->assertTrue($result);
 
         $copy = TMP . 'tests' . DS . 'test_folder_copy';
-        $result = $Folder->copy($copy);
+        $result = $Folder->copy(['to' => $copy]);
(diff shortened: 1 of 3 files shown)


tests/TestCase/Filesystem/FolderTest.php
```diff
@@ -277,11 +277,11 @@ public function testOperations()
         $this->assertTrue($result);
 
         $copy = TMP . 'tests' . DS . 'test_folder_copy';
-        $result = $Folder->copy($copy);
+        $result = $Folder->copy(['to' => $copy]);
```
ADmad: This is not a convenient usage. In most cases one would just specify destination path without any options. So changing the method signature to `copy(string $dest, array $options)` would be better.

dereuromark:
As own param this is indeed much cleaner. This argument was required anyway, but inside the options array rather hidden, now it is explicit.

ADmad:
`strict_types` declaration missing at top of file.

dereuromark:
Shouldn't we do that with a fixer instead of manually in files?

ADmad:
That's an option too.

PR #12237, opened 2018-06-14
Add support for Table magic findBy* methods in phpstan

segy:
see title


markstory:
Looks good. How does PHPStan know how to use these files?

inoas:
What about removing magic findByX automagic finders in 4.x?

dereuromark:
@ionas Those can maybe be extracted into an own plugin extension or sth?
This way people can use it if they want, but it wouldn't be core anymore directly.

markstory:
Why would we remove them? They are well documented, tested and easily not used if they are not your taste.

PR #12243, opened 2018-06-14
4.x phpcs

ADmad:


ADmad:
I updated the `phpcs.xml.dist` to include `PSR12` standard (added in phpcs 3.3) and additional useful sniffs from slevomat coding standard. This is a good way to try out the sniffs we want to include in next major release for our `CakePHP` standard.

dereuromark:
Looks good.
There seems to be one missing:
Additional newlines and especially the ones at opening classes etc.

```
class Foo 
{

    // Something

}
```
is not fixed yet to


```
class Foo 
{
    // Something
}
```

as suggested in that link as well as clear from the standard.

composer.json
```diff
@@ -42,7 +42,8 @@
     },
     "require-dev": {
         "phpunit/phpunit": "^6.0",
-        "cakephp/cakephp-codesniffer": "^3.0"
+        "cakephp/cakephp-codesniffer": "^3.0",
+        "slevomat/coding-standard": "^4.6.2"
```
markstory: I can get this installed in stickler this weekend.

ADmad:
@dereuromark Added sniff for that.

Regarding failing tests:
- It's interesting that tests checking for memory leak have started to fail on PHP 7.2. A bug in PHP?
- The error reported by phpstan is invalid IMO. I have opened an issue for it https://github.com/phpstan/phpstan/issues/1067.

ADmad:
Hmm coverage tests are run when PHP == 7.2. But those memory leak tests are not supposed to run when `CODECOVERAGE` env var is set, so I wonder why they are running.

dereuromark:
Cool! :)

I also opened a phpstan issue today as void issues are not detected.
Had to write a code sniffer here for in the meantime.

tests/TestCase/ORM/QueryRegressionTest.php
```diff
@@ -1605,7 +1605,7 @@ public function testEagerLoadOrderAndSubquery()
     public function testFormatResultsMemoryLeak()
     {
         $this->loadFixtures('Articles', 'Authors', 'Tags', 'ArticlesTags');
-        $this->skipIf(env('CODECOVERAGE') === 1, 'Running coverage this causes this tests to fail sometimes.');
+        $this->skipIf(env('CODECOVERAGE'), 'Running coverage this causes this tests to fail sometimes.');
```
dereuromark: Do we need to (bool) cast this env() result?

ADmad: No need to get too pedantic in tests :slightly_smiling_face:. Just checking whether the var is defined is enough for deciding whether to skip the test.

dereuromark: I am just thinking about skipIf() getter a bool typehint.. :) Then this breaks with any other casted boolish value.

ADmad: Done

ADmad:
This is ready to merge.

PR #12245, opened 2018-06-14
3.6 remove loading 'bootstrap'/'routes' ( core plugin )

saeideng:
remove loading 'bootstrap'/'routes' through 'core/Plugin' 
this can resolve duplication routes https://github.com/cakephp/cakephp/issues/12130
also with these changes we can define  bootstrap in plugin's `src/Plugin` class 

apps must use `RoutingMiddleware` in their `Application`

### DIFF
--- a/src/Core/Plugin.php
+++ b/src/Core/Plugin.php
@@ -178,10 +178,6 @@ class Plugin
                 $config['path'] . 'tests' . DIRECTORY_SEPARATOR
             );
         }
-
-        if ($config['bootstrap'] === true) {
-            static::bootstrap($plugin);
-        }
     }
 
     /**
--- a/tests/TestCase/Core/PluginTest.php
+++ b/tests/TestCase/Core/PluginTest.php
@@ -51,7 +51,7 @@ class PluginTest extends TestCase
     /**
      * simulate running the Application
      *
-     * @return void
+     * @return Cake\Http\BaseApplication
      */
     public function runApplication()
     {
@@ -62,6 +62,8 @@ class PluginTest extends TestCase
         $app->pluginBootstrap();
         $builder = Router::createRouteBuilder('/');
         $app->pluginRoutes($builder);
+
+        return $app;
     }
 
     /**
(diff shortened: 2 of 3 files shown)


src/Core/Plugin.php
```diff
@@ -318,14 +318,8 @@ public static function bootstrap($name)
...
         }
-        // Disable bootstrapping for this plugin as it will have
-        // been bootstrapped.
-        $plugin->disable('bootstrap');
 
-        return static::_includeFile(
-            $plugin->getConfigPath() . 'bootstrap.php',
-            true
-        );
+        return is_file($plugin->getConfigPath() . 'bootstrap.php');
```
markstory: This won't run the bootstrap function file anymore. What uses the return value of this?

saeideng: just tried to keeping BC without running /loading file

saeideng: also we can remove calling this method into load() and deprecating  this method in 3.6 or 3.7
I not found any usage of `Router::bootstrap()` in out of `load()`

markstory: The point of `Plugin::bootstrap()` is to *run* the file though. If we don't run the file here, applications not using `Application` will not have their plugins load correctly.

saeideng:
@markstory 
I reverted changes for both Plugin::bootstrap()/Plugin::routes()

now , this changes cant to resolve duplication routes 
unless user would remove `Plugin::routes()` from `config/bootstrap.php`

saeideng:
when running whole tests `vendor/bin/phpunit` there is not any error 
but with 
`vendor/bin/phpunit tests/TestCase/Core/PluginTest` 
you can see 
```php
1) Cake\Test\TestCase\Core\PluginTest::testLoadAllWithDefaultsAndOverride
Failed asserting that null matches expected 'loaded plugin bootstrap'.
```
that we can fix it with re-calling `pluginBootstrap`
 ( `runApplication()` should change to return `$app` instance in this case)
```php
        $app = $this->runApplication();
        $app->pluginBootstrap();
```

should I do something to fix this ?

markstory:
I wonder what is different about that test being run in isolation. There might be another test leaving plugin state behind.

saeideng:
running whole tests in local 
same result with `vendor/bin/phpunit tests/TestCase/Core/PluginTest`
```php
1) Cake\Test\TestCase\Core\PluginTest::testLoadAllWithDefaultsAndOverride
Failed asserting that null matches expected 'loaded plugin bootstrap'.
```
just travis is OK
I'll change testCase

tests/TestCase/Core/PluginTest.php
```diff
@@ -51,7 +51,7 @@ public function tearDown()
     /**
      * simulate running the Application
      *
-     * @return void
+     * @return Cake\Http\BaseApplication
```
dereuromark: \ missing

markstory:
Do we need these changes with #12256?

saeideng:
>Do we need these changes with #12256?

yes
I removed loading bootstrap from plugin:load()

saeideng:
Shoud we change the plugin::routes() ?
It is deprecated

markstory:
Is this still needed @saeideng

saeideng:
> also with these changes we can define bootstrap in plugin's src/Plugin class

>Is this still needed @saeideng

for defining `bootstrap` in plugin's src/Plugin class , it needed 

unless we deprecate this method

saeideng:
i think `Plugin::load()`

markstory:
@saeideng Ok. I think we should deprecate that and `Plugin::loadAll()`.

PR #12247, opened 2018-06-15
4.x - Collection typing

markstory:
Add typehints and `strict_types` to Collection. I did not add `: void` to all the tests because it was a lot of work and I was lazy.

I added more methods to `CollectionInterface` as we added a few methods in 3.x that were not part of the interface. I'm happy to remove any of them from the interface if those changes are contentious.

Refs #11935


lorenzo (approved):
Great job!

Issue #12249, opened 2018-06-15
EntityTrait::setErrors() removes indexes for addNestedMany validation errors


tuna-rootnet:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.5.17
* Platform and Target: Ubuntu 16.04, PHP 7.1.17.

### What you did
I created a reusable validator which calls `$validator->addNestedMany()`. This validator is used by a modelless form and an entity who both contain a key which needs to be validated the same way. Hence, the reusable validator. The entity is simply created with `newEntity()` or `patchEntity()`, while the form is validated through `$form->validate($data);`
The data being validated contains 3 records, of which the second has some invalid field which raises a validation error.
```php
$data = [
    ['someKey' => 'a', 'content' => 'x'],
    ['someKey' => 'b'], 'content' => '' // This triggers a validation error for notEmpty()
    ['someKey' => 'c', 'content' => 'z']
];
```

### What happened
The output for the form contains the error on the correct record, `$errors[1]`. The output for the entity contains the error on the very first record, `$errors[0]`. If there are more records which have errors, the form still shows them for the correct index while the entity just starts at zero and continues from there. If there are 5 records and record 2 and 5 trigger errors, the indexes for the entity will be 0 and 1.
``` javascript
// Output for modelless form, in json
{
    "errors": {
        "records": {
            "1": { // Correct index for offending record
                "content": {
                    "_empty": "Waarde mag niet leeg zijn"
                }
            }
        }
    }
}

// Output for entity
{
    "errors": {
        "records": {
            "0": { // <-- Well, there's my problem
                "content": {
                    "_empty": "Waarde mag niet leeg zijn"
                }
            }
        }
    }
}
```
### What you expected to happen
For both outputs to contain the error at the correct index for the records.

I tracked the issue down to `setErrors` in `src/Datasource/EntityTrait:889` (in CakePHP 3.5), which uses `array_merge`, whereby numeric indexes are lost.

markstory:
Closing as there is a pull request open now.

muslimrao:
Any solution for this ? I'm using 3.2.8

garas:
Upgrade CakePHP. Issue was fixed in 3.6.6.

muslimrao:
> Upgrade CakePHP. Issue was fixed in 3.6.6.

Thanks for replying.

[...]

PR #12250, opened 2018-06-15
Set errors with array_replace instead of array_merge

tuna-rootnet:
PR for issue https://github.com/cakephp/cakephp/issues/12249.

This keeps the indexes for errors in addNestedMany() validation errors, instead of resetting them to start at 0.


dereuromark:
I think we need some test cases here to show what the issue was and that it is fixed now.

tuna-rootnet:
I'm not that experienced with tests yet, just realized my second commit still contains a @group to check only that test...

saeideng:
why target of PR is `3.5` ?

tuna-rootnet:
@saeideng Because I encountered it in 3.5.17, my project is not on 3.6.x yet.

Also, just realized array_replace will break setting multiple errors on one property. Not sure if this PR should be closed, array_replace is in hindsight not such a nice replacement (no pun intended) and I'm not sure what a proper fix would be.

markstory:
Moved to master as 3.5 is only receiving security fixes as this point in time.

markstory:
@tuna-rootnet Do you have a way to reproduce the problem your having? I could help with converting a small validator class into a test case.

tunaflish:
@markstory Sure, see below.

Basic example how I ran into it:
```php
class SomeTable {
    public function doStuff()
    {
        $data = [
            'name' => 'something',
            'records' => [
                [
                    'foo' => 'This record will be fine',
                    'bar' => 'blah'
                ],
                [
                    'foo' => 'This one as well',
                    'bar' => 'blah'
                ],
                [
                    'foo' => 'This record will trigger an error'
                ]
            ]
        ];

        $entity = $this->newEntity($data);

        var_dump($entity->getErrors());
    }

    public function validationDefault(Validator $validator)
    {
        $validator->requirePresence('name');
        $validator->addNestedMany('records', $this->buildNestedValidator());

        return $validator;
    }

    private function buildNestedValidator()
    {
        $nestedValidator = new Validator();

        $nestedValidator->requirePresence('foo');
        $nestedValidator->requirePresence('bar');

        return $nestedValidator;
    }
}
```
Or, by feeding $entity->setError() the data which the validators produce:
```php
$entity = $someTable->newEntity(['name' => 'foo']);
$entity->setError('records', [
    2 => [
        'bar' => ['_required' => 'This field is required']
    ]
]);

var_dump($entity->getErrors());
```
Output of var_dump:
```
array(1) {
  ["records"]=>
  array(1) {
    [0]=> // This index should be 2, corresponding with the record which is invalid
    array(1) {
      ["bar"]=>
      array(1) {
        ["_required"]=>
        string(22) "This field is required"
      }
    }
  }
}
```

The index is actually correct right up to the moment array_merge is called in EntityTrait->setErrors. It's as the PHP manual states:
` Values in the input array with numeric keys will be renumbered with incrementing keys starting from zero in the result array. `

markstory:
Thank you. I'll take a look and see if I can come up with something using what you've started.

markstory:
I finished off the implementation in #12268

Issue #12254, opened 2018-06-15
Inflector does not singularize irregular camel cased words correctly


segy:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.6.5

### What you did
I have a Table class `CustomCookiesTable`. 

### What happened
Calling newEntity returns base `Entity` class since `Inflector::singularize('CustomCookies')` returns `CustomCooky`.

### What you expected to happen
I would expect newEntity to create existing `CustomCookie` entity class.

Simple solution is to call `Inflector::classify(Inflector::singularize(Inflector::underscore('CustomCookies')))` in `Table::getEntityClass`, but I think it's an Inflector bug. So please suggest if you prefer this simple solution or `Inflector::singularize` method should be fixed. I can take a look at that.

markstory:
Irregular words are often incorrectly inflected. Do you have the same problem with 'Cookies' as well?

segy:
cookies work and as I mentioned above, if you underscore it first, it also works. so I guess problem is in camel case

ADmad:
`singularize()` and `pluralize()` methods don't take casing into consideration. So `'CustomCookies'` is treated same as `'customcookies'`. So you will have to explicitly specify irregular inflection for `'customcookies' as it's done for `cookies` in core.

markstory:
'customcookies' is not something that fits within the scope of the built-in rules. You can use `Inflector::rules()` to add more rules for your application.

segy:
I can live with the fact, that inflector behaves this way. But I don't see any point in behaving of the Table. Because when i create database table `custom_cookies` and class `CustomCookiesTable`, do you think it's OK that I should add custom inflector rule or set entityClass directly? I'm creating PR for this and will leave it to your consideration wheter you want to merge.

dereuromark:
Yeah, let's discuss it on a concrete PR, that will make it easier to understand the issue.

PR #12259, opened 2018-06-18
Fix entity class for irregular table names

segy:
see #12254

### DIFF
--- a/src/ORM/Table.php
+++ b/src/ORM/Table.php
@@ -773,7 +773,7 @@ class Table implements RepositoryInterface, EventListenerInterface, EventDispatc
                 return $this->_entityClass = $default;
             }
 
-            $alias = Inflector::singularize(substr(array_pop($parts), 0, -5));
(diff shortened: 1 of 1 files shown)


dereuromark:
Can you provide some example cases as well?
Either in tests or just explaining them here in text form?

segy:
Hi. I mentioned that in the issue - good example is table class `CustomCookies`. This case (word cookie) is handled in `Inflector`, but `getEntityClass` uses `singularize` method, which doesn't take this into account. So it expects entity class `CustomCooky`.

markstory (commented):
Does bake need any changes? Is there different inflection being done there?

segy:
It uses `ConventionsTrait::_entityName` to create the class name (from table alias), so probably this method should be also modified.

markstory:
ConventionsTrait is bake only thing. We can't use it in the orm.

dereuromark:
What should we do here?

markstory:
I think it would be ideal to add a test covering the expected changes in inflection. Right now it could be 'optimized' away.

PR #12261, opened 2018-06-18
Use setDirty(field) as true by default as the name suggest.

dereuromark:
When we split the methods, this small oversight has been left.
I guess from the implicit meaning of the method name it is clear that the default of setDirty() should be setting it to dirty. As such we can add this default.
It is BC.

I used it as 

    $releaseElement->setDirty('branch');

and was wondering why the 2nd argument here was required.

### DIFF
--- a/src/Datasource/EntityTrait.php
+++ b/src/Datasource/EntityTrait.php
@@ -805,10 +805,10 @@ trait EntityTrait
      *
      * @param string $property the field to set or check status for
      * @param bool $isDirty true means the property was changed, false means
-     * it was not changed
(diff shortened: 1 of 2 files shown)


ADmad (approved):

ADmad:
`EntityInterface::setDirty()` will also need to be updated in `4.x`.

markstory:
Should we have a test for the default value? It changes tje public api we offer.

dereuromark:
I adjusted a test. The same one contains the verbose one still afterwards.

PR #12268, opened 2018-06-20
Fix nested validator error messages

markstory:
Continue work started in #12250 to resolve issues with nested validator error messages.


tests/TestCase/ORM/EntityTest.php
```diff
@@ -1213,6 +1213,7 @@ public function testErrors()
     /**
      * Tests error getters and setters
      *
+     * @group fish
```
tuna-rootnet: I left this line in my commits by accident, it's not necessary at all.

Issue #12271, opened 2018-06-20
Plugin paths not set automatically for Application::makePlugin()


sdustinh:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

```bash
sdustinh@sdustinh:~/github/app$ composer show cakephp/*
cakephp/bake                1.6.4              Bake plugin for CakePHP 3
cakephp/cakephp             dev-master b02dae1 The CakePHP framework
cakephp/cakephp-codesniffer 3.0.3              CakePHP CodeSniffer Standards
cakephp/chronos             1.1.4              A simple API extension for DateTime.
cakephp/debug_kit           3.14.2             CakePHP Debug Kit
cakephp/elastic-search      0.3.5              An Elastic Search datasource and data mapper for CakePHP 3.0
cakephp/migrations          1.7.2              Database Migration plugin for CakePHP 3.0 based on Phinx
cakephp/plugin-installer    1.1.0              A composer installer for CakePHP 3.0+ plugins.
```

I've been trying to update my application to 3.6 for a while but there have been a handful of issues that prevent the jump so I've been sitting at 3.5 until they are resolved. I read #12130 and figured that solved the problems that I was having so I gave it another shot. I did the following to an application running the most recent 3.5 release:

- Set my `Error.errorLevel` to `E_ALL & ~E_USER_DEPRECATED`
- Added the `_cake_routes_` cache configuration and the `\Cake\Routing\Middleware\RoutingMiddleware`
- Removed `Plugin::routes()` from my application `config/routes.php`
- Removed all `Plugin::load()` calls from `config/bootstrap.php` and `config/bootstrap_cli.php`
- Loaded all of my plugins with `src/Application->addPlugin()` AFTER a `parent::bootstrap()` call
- Ran `vendor/bin/phpunit`

I get errors relating to plugin paths. Any element on the view that use `Plugin.name` fail. When I watch the logs stream I can see logs like `2018-06-20 11:36:18 Debug: Updating last_login for N/A user: john.doe@example.com (cecba869-6b02-4de2-945d-c9ac50a9741a)` where `N/A` is a column on an associated table and the email / uuid are from the primary table which tells me that the default `Cake\ORM\Table` is being used because the plugin class wasn't found.

I did the following to debug:

```php
// Method overload
public function makePlugin($name, array $config)
{
    $plugin = parent::makePlugin($name, $config);

    if ($name === 'Users') {
        dd([
[...]

PR #12273, opened 2018-06-20
Use Plugin::load() when creating classless plugins.

markstory:
I completely missed that Plugin::load() does a bunch of work to generate the correct paths for plugins. By using Plugin::load() and fetching the created plugin we can leverage the same logic. There is an argument to be had around deprecating Plugin::load(). Given the troubles plugin classes have given us I'm not comfortable in doing that right now.

Refs #12271


saeideng:
can we duplicate code ?
 `Plugin::load()` has problem with calling bootstrap directly

markstory:
Darn I forgot about that. Duplicating code will allow us to deprecate plugin::load() more easily in the future as well I guess.

markstory:
@saeideng I think I've resolved the issue you raised and also fixed a potential problem where `cakephp-plugins.php` would not get loaded.

src/Http/BaseApplication.php
```diff
@@ -119,22 +123,22 @@ public function getPlugins()
      * @param array $config Configuration options for the plugin
      * @return \Cake\Core\PluginInterface
      */
-    public function makePlugin($name, array $config)
+    protected function makePlugin($name, array $config)
```
markstory: There is no reason this method needs to be public, and I consider it a mistake that I would like to fix now vs. later. If folks think it is a risky change I'm ok with reverting this.

ADmad: Making is protected sounds good to me.

PR #12286, opened 2018-06-25
4.x filesystem types

markstory:
I had to tweak a bunch of return types to be nullable instead of the `string|false` pattern which we can't represent with PHP types.

Refs #11935

### DIFF
--- a/src/Filesystem/File.php
+++ b/src/Filesystem/File.php
@@ -1,4 +1,5 @@
 <?php
+declare(strict_types=1);
 /**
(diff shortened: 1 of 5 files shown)


src/Filesystem/File.php
```diff
@@ -147,7 +148,7 @@ public function open($mode = 'r', $force = false)
      * @param bool $force If true then the file will be re-opened even if its already opened, otherwise it won't
      * @return string|false string on success, false on failure
```
dereuromark: Same here

PR #12287, opened 2018-06-25
Adjust App::className() to be clean nullable return.

dereuromark:
And follow up fixes.

### DIFF
--- a/src/Auth/PasswordHasherFactory.php
+++ b/src/Auth/PasswordHasherFactory.php
@@ -46,7 +46,7 @@ class PasswordHasherFactory
         }
 
         $className = App::className($class, 'Auth', 'PasswordHasher');
-        if ($className === false) {
+        if ($className === null) {
             throw new RuntimeException(sprintf('Password hasher class "%s" was not found.', $class));
         }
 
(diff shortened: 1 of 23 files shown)


src/ORM/Association.php
```diff
@@ -1103,6 +1103,7 @@ protected function _extractFinder($finderData)
      * @param string $alias The alias name you want to get.
      * @param array $options Table options array.
      * @return string
+     * @throws \Exception If class does not exist.
```
ADmad: \RuntimeException

src/ORM/Association.php
```diff
@@ -1111,6 +1112,11 @@ protected function _getClassName($alias, array $options = [])
         }
 
         $className = App::className($options['className'], 'Model/Table', 'Table') ?: 'Cake\ORM\Table';
+        if (!$className) {
+            throw new RuntimeException(
+                sprintf('Class for %s could not be found.', $options['className'])
+            );
+        }
```
markstory: Can this error ever be raised? The `?:` should make it impossible.

inoas: Can we make automagic tables and/or entities opt-out for 4.x through Configure::?

dereuromark: Fixed

markstory: @inoas Possibly. It could be an opt-in mode that makes debugging auto-table issues easier.

Issue #12295, opened 2018-06-26
Subdirectory is added again when using $routes->redirect


Bird87ZA:
This is a (multiple allowed):

* [x] bug

* CakePHP Version: 3.6.5
* Platform and Target: AWS Hosted server (Ubuntu 14.04.5 LTS (GNU/Linux 3.13.0-129-generic x86_64)) PostgreSQL DB. Website is hosted under `www.example.com`, CakePHP runs from `www.example.com/webapp`

### What you did
I created a shortlink for a specific route in my `routes.php` file like so:

    $routes->redirect(
        '/g/:id',
        ['controller' => 'Jobs', 'action' => 'view'],
        ['pass' => ['id'], 'persist' => true]
    );

### What happened
I am being redirected to `www.example.com/webapp/webapp/jobs/view/5`

### What you expected to happen
I expect it to redirect me to `www.example.com/webapp/jobs/view/5`

### More info
I did quite a few debugging and it seems that the bug slips in around the `Router::reverse()` function. You can see my original discussion around the issue with @ndm2 [here](https://stackoverflow.com/questions/51035419/routes-redirect-adds-base-to-address).

markstory:
Pull request open now.

PR #12296, opened 2018-06-26
Fixed issue #12295 - https://github.com/cakephp/cakephp/issues/12295

Bird87ZA:
I just changed the call in `Router::parse()` from `reverse()` to `reverseToArray()` for [this issue](https://github.com/cakephp/cakephp/issues/12295).

Didn't run the tests as I can't get them working for some odd reason.

### DIFF
--- a/src/Routing/Route/RedirectRoute.php
+++ b/src/Routing/Route/RedirectRoute.php
@@ -88,7 +88,7 @@ class RedirectRoute extends Route
                     }
                 }
             }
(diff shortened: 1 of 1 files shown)


markstory:
Thank you for the patch I can add a test case. What errors/problems do you have when you attempt to run the tests?

markstory:
> Suppose I just need to set up the DB.

Yes. If you have ext/pdo_sqlite installed the default phpunit configuration will use that.

PR #12302, opened 2018-06-28
Add more typing to Log

markstory:
Many log parameters are mixed types and couldn't have types assigned to them.

Refs #11935

### DIFF
--- a/src/Log/Engine/BaseLog.php
+++ b/src/Log/Engine/BaseLog.php
@@ -1,4 +1,5 @@
 <?php
+declare(strict_types=1);
 /**
  * CakePHP(tm) :  Rapid Development Framework (https://cakephp.org)
  * Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
@@ -66,7 +67,7 @@ abstract class BaseLog extends AbstractLogger
      *
      * @return array
      */
-    public function levels()
+    public function levels(): array
     {
         return $this->_config['levels'];
     }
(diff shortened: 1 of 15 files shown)


src/Log/Engine/ConsoleLog.php
```diff
@@ -86,7 +87,7 @@ public function __construct(array $config = [])
      * @param array $context Additional information about the logged message
      * @return bool success of write.
      */
-    public function log($level, $message, array $context = [])
+    public function log($level, $message, array $context = []): bool
```
saeideng: can we have 
```php
log(string $level, string $message, .....
``` 
?

dereuromark: Level can also be int.

src/Log/Engine/BaseLog.php
```diff
@@ -66,15 +67,15 @@ public function __construct(array $config = [])
...
+    public function levels(): array
     {
         return $this->_config['levels'];
     }
 
     /**
      * Get the scopes this logger is interested in.
      *
-     * @return array
+     * @return array|bool
```
dereuromark: Could we use null as alternative value?

markstory: We could but that will break compatibility. Right now `false` is used to indicate 'no scope'. I'd like to revisit scopes/tags in the future so this might go away as part of 4.x releaess.

src/Log/Engine/ConsoleLog.php
markstory: And `message` can be an array.

PR #12305, opened 2018-06-28
MultiCheckboxWidget should use checkboxWrapper, not radioWrapper.

chinpei215:
The current code seems incorrect to me, as MultiCheckboxWidget checks radioWrapper not checkboxWrapper. But I am not sure whether my fix is correct or not, as values of 'options' are not rendered when 'label' => false and 'checkboxWrapper' doesn't contain '{{input}}'. Is this correct?

### DIFF
--- a/tests/TestCase/View/Widget/MultiCheckboxWidgetTest.php
+++ b/tests/TestCase/View/Widget/MultiCheckboxWidgetTest.php
@@ -370,6 +370,77 @@ class MultiCheckboxWidgetTest extends TestCase
     }
 
     /**
+     * Test label = false with checkboxWrapper option.
+     *
+     * @return void
+     */
+    public function testNoLabelWithCheckboxWrapperOption()
+    {
+        $data = [
+            'label' => false,
+            'name' => 'test',
+            'options' => [
+                1 => 'A',
+                2 => 'B',
+            ],
+        ];
+
+        $label = new LabelWidget($this->templates);
+        $input = new MultiCheckboxWidget($this->templates, $label);
+        $result = $input->render($data, $this->context);
+        $expected = [
+            ['div' => ['class' => 'checkbox']],
+            ['input' => [
+                'type' => 'checkbox',
+                'name' => 'test[]',
+                'value' => 1,
+                'id' => 'test-1',
+            ]],
+            ['label' => ['for' => 'test-1']],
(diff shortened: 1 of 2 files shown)


tests/TestCase/View/Widget/MultiCheckboxWidgetTest.php
```diff
@@ -369,6 +369,77 @@ public function testRenderTemplateVars()
...
 
+    /**
+     * Test label = false with checkboxWrapper option.
+     *
+     * @return void
+     */
+    public function testNoLabelWithCheckboxWrapperOption()
+    {
+        $data = [
+            'label' => false,
```
markstory: Previously were there no labels?

chinpei215: Before:
```html
<!-- First -->
<div class="checkbox">
  <input type="checkbox" name="test[]" value="1" id="test-1">
  <input type="checkbox" name="test[]" value="1" id="test-1">
</div>
<div class="checkbox">
  <input type="checkbox" name="test[]" value="2" id="test-2">
  <input type="checkbox" name="test[]" value="2" id="test-2">
</div>

<!-- Second -->
<div class="checkbox">
  <input type="checkbox" name="test[]" value="1" id="test-1">
</div>
<div class="checkbox">
  <input type="checkbox" name="test[]" value="2" id="test-2">
</div>
```

After:
```html
<!-- First -->
<div class="checkbox">
  <input type="checkbox" name="test[]" value="1" id="test-1">
  <label for="test-1">A</label>
</div>
<div class="checkbox">
  <input type="checkbox" name="test[]" value="2" id="test-2">
  <label for="test-2">B</label>
</div>

<!-- Second -->
<div class="checkbox">
  <input type="checkbox" name="test[]" value="1" id="test-1">
</div>
<div class="checkbox">
  <input type="checkbox" name="test[]" value="2" id="test-2">
</div>
```

So I think I fixed something wrong, but I am not sure the above HTML is an expected result or not. Now labels will appear even when label => false given, if checkboxWrapper contains {{input}}.

markstory: The default template doesn't contain `{{input}}` so we should be ok.

PR #12307, opened 2018-06-30
4x - Add types to Utility

markstory:
Another part of #11935

### DIFF
--- a/src/Utility/Hash.php
+++ b/src/Utility/Hash.php
@@ -1,4 +1,5 @@
 <?php
+declare(strict_types=1);
 /**
  * CakePHP(tm) : Rapid Development Framework (https://cakephp.org)
  * Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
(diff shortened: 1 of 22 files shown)


lorenzo (approved):

src/Utility/Hash.php
```diff
@@ -446,14 +447,14 @@ public static function remove(array $data, $path)
      * following the path specified in `$groupPath`.
      *
      * @param array $data Array from where to extract keys and values
-     * @param string $keyPath A dot-separated string.
-     * @param string|null $valuePath A dot-separated string.
-     * @param string|null $groupPath A dot-separated string.
+     * @param string|array $keyPath A dot-separated string.
+     * @param string|array|null $valuePath A dot-separated string.
+     * @param string|array|null $groupPath A dot-separated string.
```
saeideng: I think `groupPath` should be `string|null` 
```php
static::extract($data, $groupPath)
```
extract just accept `string`

markstory: I can try, but I think there are tests that pass an array of paths.

PR #12308, opened 2018-06-30
add `direction:ltr` - Debugger::printVar()/debug()

saeideng:


dereuromark:
Does it have to be noisy inline css?
Can't it just be in the CSS file - for class cake-debug-output?

saeideng:
do you want to add `cake-debug-output` class into product css ?

dereuromark:
That class is added by default.
As such, you should modify cakephp/app to have this in the default template.
For those few that need it. All others (9x%) don't need this inline css or any other such fix.

saeideng:
we can remove `cake-debug-output` class at cake 4 :)

dereuromark:
No why should we?

saeideng:
probably we able to add this class in footer like the adding debug_kit 
```php
<style>
  .cake-debug-output{
    direction:ltr;
   }
<style>
</body>
```

dereuromark:
Why not just using basic css stylesheets?

ADmad:
Inline css for debugging code is just fine. I wouldn't want to include a css file just for debugging.

ADmad:
> Does it have to be noisy inline css?

Debug out itself is noisy.

dereuromark:
I can see the point for dd() usage.

PR #12309, opened 2018-06-30
[3.7] - deprecate the Plugin::load/loadAll/bootstrap()

saeideng:


tests/TestCase/I18n/I18nTest.php
```diff
@@ -180,7 +181,7 @@ public function testCreateCustomTranslationPackage()
      */
     public function testPluginMesagesLoad()
     {
-        Plugin::load([
+        loadPlugin::load([
```
ADmad: Should be captial "L". Same below too.

tests/test_app/TestApp/LoadPlugin.php
```diff
@@ -0,0 +1,52 @@
...
+class LoadPlugin
+{
+    /**
+     * simulate running the Application
+     *
+     * @return \Cake\Http\BaseApplication
+     */
+    public static function load($plugins = [])
+    {
+        try {
```
markstory: Should we be using PluginCollection::add() here instead? The collection can be accessed from Plugin.  This 'helper' could potentially be added to TestCase. I suspect there are a few apps and plugin test suites that need to load plugins without an application around.

saeideng:
>Should we be using PluginCollection::add() here instead? 

in that case we have to duplicate code for making Plugin

and 
```php
// src/Core/BasePlugin.php
    public function bootstrap(PluginApplicationInterface $app)
    {
    }
```
`bootstrap()` needs an instance of Application

saeideng:
>This 'helper' could potentially be added to TestCase

yes probably

markstory:
> in that case we have to duplicate code for making Plugin

I forgot about that. What you have is better than adding more duplication.

tests/TestCase/View/HelperTest.php
```diff
@@ -122,8 +122,6 @@ public function testLazyLoadingHelpers()
      */
     public function testThatHelperHelpersAreNotAttached()
     {
-        Plugin::loadAll();
```
saeideng: I dont know why we need to this!! ? :thinking:

tests/TestCase/Http/BaseApplicationTest.php
```diff
@@ -219,8 +221,9 @@ public function testPluginBootstrapRecursivePlugins()
             [$this->path]
         );
         $app->addPlugin('ParentPlugin');
-        $app->pluginBootstrap();
-
+        $this->deprecated(function () use ($app) {
+            $app->pluginBootstrap();
+        });
```
saeideng: wrapped with `deprecated` because `ParentPlugin` has a `Plugin:load()`

src/TestSuite/TestCase.php
```diff
@@ -189,6 +190,35 @@ public function loadFixtures()
         }
     }
 
+    /**
+     * simulate running the Application
+     * @param string|array $plugins list of Plugins to load
```
ADmad: Aren't we trying to avoid mixed mode arguments going forth?

saeideng: I will change it

saeideng:
@markstory now , main problem is having static loading Plugin here 
https://github.com/cakephp/cakephp/blob/9a67c9d7158d5e299418f9956d1a8af39128cf57/src/Shell/Task/ExtractTask.php#L172

markstory:
@saeideng Instead of loading the plugin we could raise an error. We don't force load plugins in the rest of CakePHP. I don't see why this has to be an exception.

markstory:
@saeideng I pushed a few changes that should fix the tests.

saeideng:
I'll open another PR to fix phpstan

markstory:
I will get the documentation updated and then get this merged.

PR #12310, opened 2018-07-01
4.x - form typing

markstory:
Add typing to Form package.

Refs #11935

### DIFF
--- a/src/Form/Form.php
+++ b/src/Form/Form.php
@@ -1,4 +1,5 @@
 <?php
+declare(strict_types=1);
 /**
  * CakePHP(tm) : Rapid Development Framework (https://cakephp.org)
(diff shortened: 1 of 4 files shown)


src/Form/Form.php
```diff
@@ -253,7 +252,6 @@ protected function _execute(array $data)
      * @param string|null $field The field name or null to get data array with
      *   all fields.
      * @return mixed
-     * @since 3.7.0
      */
     public function getData($field = null)
```
saeideng: maybe `getData(?string $field = null)`

markstory: Good catch. Fixed in 31614341e2ce9ca22ec1fcd44ca3a08751aaf916

PR #12313, opened 2018-07-02
Stick to normal php min versioning.

dereuromark:
We should stick to normal PHP min versioning.
The minors in PHP usually are rather compatible, and restricting here only creates hazzle later on.

IDEs also dont like this much, as I had to discover.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -28,7 +28,7 @@
         "source": "https://github.com/cakephp/cakephp"
     },
     "require": {
-        "php": ">=7.1.0,<7.3.0",
+        "php": ">=7.1.0",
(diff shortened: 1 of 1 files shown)


markstory:
What happens if 7.3 does contain breaking changes or deprecations that impact us? Our requirements will indicate that packages are ok when they are not.

dereuromark:
The what if should be decided then, but in most cases you just block all minors (older ones then current) and their patches to be deployed for no reason :)

dereuromark:
PS: deprecations are E_DEPRECATED, so not even those would be a problem for production, they only notify, but the code continues to work just fine here as well.

PR #12315, opened 2018-07-02
Fix up return types for Console.

dereuromark:
Nullable type.


dereuromark:
Whats going on with appveyor all the time?
Weirdest stuff going on there:

> Command exited with code -1073741819

dereuromark:
But all real tests should pass :)

markstory:
Appveyor does that every so often. I haven't been able to figure out a way to reproduce it locally though.

Issue #12304, opened 2018-06-28
Unexpected behavior for i18nFormat with some timezones


BassemN:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version:  3.5.17 & 3.6.6
* Platform and Target: Apache/2.4.27, MySQL 5.6.37, PHP 7.1.8

### What you did
Default time zone is UTC. When changing the displayed timezones format some works and others not so please check the following example
```php
// Working timezones
pr(\Cake\I18n\Time::instance($page->modified)->i18nFormat(null, "America/Adak", null));
pr(\Cake\I18n\Time::instance($page->modified)->i18nFormat(null, "America/El_Salvador", null));
pr(\Cake\I18n\Time::instance($page->modified)->i18nFormat(null, "Africa/Nouakchott", null));

// Not working timezones
pr(\Cake\I18n\Time::instance($page->modified)->i18nFormat(null, "America/Fort_Nelson", null));
pr(\Cake\I18n\Time::instance($page->modified)->i18nFormat(null, "Asia/Yangon", null));
```
output of `var_dump($page->modified)`
```
object(Cake\I18n\FrozenTime)[56]
  public 'time' => string '2016-07-01T20:14:16+00:00' (length=25)
  public 'timezone' => string 'UTC' (length=3)
  public 'fixedNowTime' => boolean false
```

### What happened
Timezone like `America/Fort_Nelson`, `Asia/Yangon` releasing the following error message
```
2018-06-28 13:25:44 Error: [Error] Call to a member function format() on null
Request URL: /pages/privacy
Stack Trace:
#0 /vendor/cakephp/cakephp/src/I18n/DateFormatTrait.php(177): Cake\I18n\Time->_formatObject(Object(Cake\I18n\Time), Array, NULL)
#1 /src/Template/Pages/view.ctp(20): Cake\I18n\Time->i18nFormat(Array, 'America/Fort_Ne...', NULL)
#2 /vendor/cakephp/cakephp/src/View/View.php(1203): include('/usr/local/ampp...')
#3 /vendor/cakephp/cakephp/src/View/View.php(1164): Cake\View\View->_evaluate('/usr/local/ampp...', Array)
#4 /vendor/cakephp/cakephp/src/View/View.php(772): Cake\View\View->_render('/usr/local/ampp...')
[...]

PR #12319, opened 2018-07-03
Add an exception when datefmt_create fails.

markstory:
Try to provide more useful feedback over method call on null.

Refs #12304


lorenzo (approved):

BassemN:
I tested it on my app and I can catch the error.

BassemN (approved):

PR #12320, opened 2018-07-03
More secure default for redirects.

dereuromark:
Lots of people just have this in their post action code:
```php
return $this->redirect($this->referer(['action' => 'index']));
```

I also found a few of those in my code/apps.

Now that has a few security issues out of the box. You can spoof the referrer to an external untrusted source and maybe exploit the site here.
So I think the sane default should be server internal, and only explicitly set it to allow external URLs.

What do you think?

refs https://www.owasp.org/index.php/Unvalidated_Redirects_and_Forwards_Cheat_Sheet


ravage84 (approved):

lorenzo:
I like this :+1:

markstory:
I too like this change. Migration guide updated.

PR #12321, opened 2018-07-03
Use cleaner api for Plugin loaded info.

dereuromark:
We need a cleaner API here.
These "god-getters" of both scalar and multiple depending on first arg are a bad idea.
They prevent typehinting, they also correctly make PHPStan and tooling complain a lot here.
You are forced to cast everywhere in code in current 3.x.

This would fix it with a clear message in case you didnt adjust code.

What do you think?
Then I can adjust the tests.

### DIFF
--- a/src/Core/Plugin.php
+++ b/src/Core/Plugin.php
@@ -14,7 +14,6 @@
  */
 namespace Cake\Core;
 
-use ArgumentCountError;
 use DirectoryIterator;
 
 /**
--- a/src/Core/Plugin.php
+++ b/src/Core/Plugin.php
@@ -14,7 +14,7 @@
  */
 namespace Cake\Core;
 
-use Cake\Core\Exception\MissingPluginException;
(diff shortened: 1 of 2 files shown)


src/Core/Plugin.php
```diff
@@ -285,18 +285,27 @@ public static function bootstrap($name)
...
+     * Returns true if the plugin $plugin is already loaded.
      *
-     * @param string|null $plugin Plugin name.
-     * @return bool|array Boolean true if $plugin is already loaded.
-     *   If $plugin is null, returns a list of plugins that have been loaded
+     * @param string $plugin Plugin name.
+     * @return bool
      */
-    public static function loaded($plugin = null)
+    public static function isLoaded(string $plugin): bool
```
markstory: These look like methods we should also backport to 3.next so that we can add deprecation warnings in 3.7.

PR #12322, opened 2018-07-03
Add typehinting to Plugin class.

dereuromark:
Also bootstrap() does not need to return anything. The result is also not used currently.


src/Core/Plugin.php
```diff
@@ -240,7 +240,7 @@ public static function path($name)
      * @return string Path to the plugin folder container class folders.
      * @throws \Cake\Core\Exception\MissingPluginException If plugin has not been loaded.
      */
-    public static function classPath($name)
+    public static function classPath($name): string
```
markstory: Should we also add parameter types?

PR #12323, opened 2018-07-03
Add typehinting to Configure class. Cleaner API.

dereuromark:
Follow https://github.com/cakephp/cakephp/pull/12321

- Cleaner API for all configured() vs specific isConfigured() with a clear migration exception.
- void for write and clear (since they always return true and are effectively void), such writing methods should always throw an exception instead, that is way cleaner than silently returning false.
- added typehints where possible

If the API sounds good, I can adjust the tests here and provide a docs migration update.

### DIFF
--- a/src/Core/Configure.php
+++ b/src/Core/Configure.php
@@ -78,10 +78,10 @@ class Configure
      * @param string|array $config The key to write, can be a dot notation value.
      * Alternatively can be an array containing key(s) and value(s).
      * @param mixed $value Value to set for var
-     * @return bool True if write was successful
+     * @return void
      * @link https://book.cakephp.org/3.0/en/development/configuration.html#writing-configuration-data
      */
-    public static function write($config, $value = null)
+    public static function write($config, $value = null): void
     {
         if (!is_array($config)) {
             $config = [$config => $value];
@@ -99,8 +99,6 @@ class Configure
                 ini_set('display_errors', $config['debug'] ? '1' : '0');
             }
         }
-
-        return true;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


src/Core/Configure.php
```diff
@@ -252,23 +251,32 @@ public static function consume($var)
...
         static::$_engines[$name] = $engine;
     }
 
+    /**
+     * Returns true if the Engine objects is configured.
+     *
+     * @param string $name Engine name.
+     * @return bool
+     */
+    public static function isConfigured(string $name): bool
```
markstory: I think this is a reasonable change. We could add a deprecation warning/backport this method to 3.next as well so anyone using the old signature can upgrade early.

src/Core/Configure.php
```diff
@@ -78,10 +79,10 @@ class Configure
      * @param string|array $config The key to write, can be a dot notation value.
      * Alternatively can be an array containing key(s) and value(s).
      * @param mixed $value Value to set for var
-     * @return bool True if write was successful
+     * @return void
      * @link https://book.cakephp.org/3.0/en/development/configuration.html#writing-configuration-data
      */
-    public static function write($config, $value = null)
+    public static function write($config, $value = null): void
```
markstory: In the vein of #12335 this seems like a similar change that could disrupt folks more than we need to. I agree that always returning `true` is a bit goofy, but it is a consistent type at least.

src/Core/Configure.php
```diff
@@ -252,21 +251,31 @@ public static function consume($var)
...
-     * @param string|null $name Engine name.
-     * @return array|bool Array of the configured Engine objects, bool for specific name.
+     * @return array
      */
-    public static function configured($name = null)
+    public static function configured(): array
     {
-        if ($name !== null) {
-            return isset(static::$_engines[$name]);
+        if (count(func_get_args())) {
```
chinpei215: func_num_args() would be better.

Issue #11854, opened 2018-03-20
New entity should not bypass validation


dereuromark:
This is a (multiple allowed):

* [x] bug
* [x] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: latest master 3.5.x

### What you did
Somewhere in the code was a left over
```
$data = [
    ...
];
$modulesTable->saveOrFail($modulesTable->newEntity());
```

### What happened
Stored in DB as entity of content:
```php
		'id' => (int) 2,
		'name' => '',
		'version' => '',
```

It silently saved it without all the validation triggered, which is defined as
```php
        $validator
            ->requirePresence('name', 'create')
            ->add('name', 'unique', ['rule' => ['validateUnique', ['scope' => ['name', 'type']]], 'provider' => 'table'])
            ->notEmpty('name');

        $validator
            ->requirePresence('version', 'create')
            ->notEmpty('version');
```

### What you expected to happen
This should not allow to be saved with the constraints here not respected (in this case name and version being not empty).

What could be done here for the code to be less error prone (by accident) and make sure the entities are not persisted like this when accidentally forgetting to pass the $data?

Doesn't newEntity() patch just as patchEntity() does? And therefore does not accept this state?

ADmad:
Validation would be triggered only if you pass `$data` argument to `newEntity()`, which isn't in your example.

dereuromark:
Yeah, but the save() operation should then not persist still IMO. Some exception seems useful in these cases. Couldn't it then trigger the patching with empty array and see what happens?

dereuromark:
The alternative is to remove the newEntity() optional argument and make the 1st arg always required.
Currently this is a dangerous thing, and should not continue to be allowed the way it is.

lorenzo:
Making the first argument required is not a bad idea, we could explore that

dereuromark:
A basic rule of DTO is also (according to clean code) that they should always be in a state of validity after creation (usually using constructor injection of data). Or at least contain then a state of "I cant be persisted".
So since none of this is the case, the DTO itself needs to be otherwise be made non-saveable.

dereuromark:
For example a state of "notValidated" could be part of the entity, and if so, it would at least trigger an empty validation on save - which would prevent this empty one to be silently stored as OK (which it clearly is not from application rule setup).

xaphalanx:
I view newEntity()/patchEntity() validation as "request" validation.
Entity validation to me are "rules".

[...]

PR #12327, opened 2018-07-04
Adds a method hasErrors() to the EntityTrait

JoshuaLuckers:
To improve performance we set the property _hasErrors to true when Entity->setErrors() is being called.

We use the function hasErrors() in Table->save() instead of going trough all errors.

Related comments in issue: #11854


src/Datasource/EntityTrait.php
```diff
@@ -911,6 +918,16 @@ public function getErrors()
...
     }
 
+    /**
+     * Returns whether this entity contains errors.
+     *
+     * @return bool
+     */
+    public function hasErrors()
+    {
+        return $this->_hasErrors;
```
lorenzo: Wouldn’t it be better to just return `!empty($this->_errors)` instead of having a new property?

JoshuaLuckers: A new property makes the check more strict. It also opens the possibility to set different types of errors without being depended on `$this->_errors`.

JoshuaLuckers:
I'm not sure why the builds are failing with these changes..

src/Datasource/EntityTrait.php
dereuromark: But wouldn't that make it more confusing? To have hasErrors true, but then reading the error list can be empty etc?
I don't think this is wise.

ADmad: On the other hand having multiple sources of truth also increases the chance of disparity and ending up returning wrong result.

JoshuaLuckers: Maybe I should rename the property and method to `canBePersisted`. 

Not having errors doesn't mean we the Entity can be persisted. Perhaps this is determined by other variables like, we only want to be sure an Entity has been validated. This also makes more sense regarding the `newEntity()` issue.

dereuromark: That might be a semantically more clear approach, yes.

markstory (commented):
New public methods require some tests to help prevent future regressions. If you need help with writing the tests or what tests to write let us know :smile:

src/Datasource/EntityTrait.php
```diff
@@ -943,6 +961,7 @@ public function getError($field)
      */
     public function setErrors(array $fields, $overwrite = false)
     {
+        $this->_canBePersisted = false;
```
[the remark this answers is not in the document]
lorenzo: The property is just asking for trouble, better just return that the errors are empty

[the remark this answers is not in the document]
JoshuaLuckers: Points taken. I will rename it back to `hasErrors()` and return `!empty($this->_errors)`.

I will also add a test for the new public method.

ndm2:
It would be a slight change in behavior I guess, given that `getErrors()` works recursively. Without recursion, possible nested errors would only be considered when the respective nested entity is being saved, which is quite different to the current behavior, and similar to how application rule errors are evaluated.

I guess from an application design perspective, this is something that the application should be able to handle anyways, but if it's about performance, then this might actually perform worse, given that saving operations are likely to be more expensive than building the errors array?

JoshuaLuckers:
@ndm2 The recursive `getErrors()` caused a massive performance impact when I tried to save `Entities` with a huge amount of associations (nested children).

ndm2:
I'm not doubting that, surely there's a break even point somewhere, I'm just wondering whether your day to day saving cycles actually reaches/crosses that point, and whether things might then perform worse for those, as the saving cycle might first go trhough 10 actual saving operations (including database operations, possible custom before/after save/delete logic, etc), and stops at the last entity which has an error.

I could image that most time in building the errors array is spent with collection operations and in `EntityTrait::_nestedErrors()`, ie adding vanilla PHP recursion that just collects the `hasErrors()` state and bails out at the first `true`, would be much faster, even with a fair amount of nested entities, so maybe the current bail out early behavior could be kept?

ADmad:
I agree with @ndm2. In most cases I would rather have my nested entities validated first instead of start saving just to have everything rolled back when the nth child entity fails validation.

ADmad:
So this becomes akin to the `validate => true` vs `validate => 'first'` we have for `saveAll()` in Cake 2.

JoshuaLuckers:
If I'm not mistaken the current behaviour also repeats the nested validation for associated entities that are already validated. I should dig a bit deeper to validate this.

ndm2:
It's not about applying validation, it's about testing the results of validation. `getErrors()` doesn't apply validation, it just collects the results, and it does so in a recursive fashion for all nested entities, so testing its return value will prevent the saving process from continuing as early as possible, and that's what I think might be worth preserving.

JoshuaLuckers:
The problem is. When saving an Entity with Associations. `Table::save()` will call `getErrors()` on the Entity that is being saved. This is fine in my case.

However `getErrors()` also checks calls `getErrors()` for the Associated Entities (as intended). But when the Associated Entity gets saved by the Marshaller and the same behaviour happens again for the associated entity.

JoshuaLuckers:
I misunderstood. That seems like a good solution.

src/Datasource/EntityTrait.php
```diff
@@ -891,6 +891,31 @@ public function isNew($new = null)
         return $this->_new = $new;
     }
 
+    /**
+     * Returns whether this entity has errors.
+     *
+     * @param bool $includeHasErrorsFromNestedEntities
```
ADmad: This is an insanely long argument name :slightly_smiling_face:

src/Datasource/EntityTrait.php
```diff
@@ -891,6 +891,31 @@ public function isNew($new = null)
...
+     * @return bool
+     */
+    public function hasErrors($includeHasErrorsFromNestedEntities = true)
+    {
+        if (empty($this->_errors) === false) {
```
ADmad: This can be written as `if (!empty($this->_errors)) {`. Similarly all other `if` statements where you compare a return value or argument which is already boolean should be updated.

src/Datasource/EntityTrait.php
```diff
@@ -1090,6 +1115,29 @@ protected function _nestedErrors($field)
         return [];
     }
 
+    /**
+     * Reads if there are errors for one or many objects.
+     *
+     * @param array|\Cake\Datasource\EntityTrait $object The object to read errors from.
```
ADmad: The param type should be `mixed`. Since you are passing all properties to this method, `$object` can be of any type.

JoshuaLuckers:
Thanks all for your time and the provided feedback.

lorenzo:
@JoshuaLuckers Did you see any performance change for your use case after this?

src/Datasource/EntityTrait.php
```diff
@@ -894,21 +894,21 @@ public function isNew($new = null)
...
-    public function hasErrors($includeHasErrorsFromNestedEntities = true)
+    public function hasErrors($includeNested = true)
     {
-        if (empty($this->_errors) === false) {
+        if (!empty($this->_errors)) {
             return true;
         }
 
-        if ($includeHasErrorsFromNestedEntities === false) {
+        if ($includeNested === false) {
```
ADmad: `!$includeNested`

JoshuaLuckers:
@Lorenzo Yes, did a few test runs and I did notice a great increase in performance in my use case.

src/Datasource/EntityTrait.php
JoshuaLuckers: Isn't it better to use `=== false` in this case to be more strict about the type of the value?
Various other if statements in the Entity class also use `=== false`.

PR #12333, opened 2018-07-05
Use clean nullable for Route and RouteCollection. Add typehinting.

dereuromark:
Allows for stricter typehinting here.
Also adjusted affected code.


markstory:
Great work @dereuromark :clap:

PR #12334, opened 2018-07-05
3.6 missing plugin - page

saeideng:


src/Template/Error/missing_plugin.ctp
```diff
@@ -34,22 +34,16 @@ $this->start('subheading');
 <?php
 $code = <<<PHP
 <?php
-Plugin::load('{$plugin}');
+// src/Application.php
+public function bootstrap()
+{
+    parent::bootstrap();
+
+    \$this->addPlugin('{$plugin}');
```
ADmad: Extra backslash at the start.

saeideng: Required for '$'

PR #12338, opened 2018-07-08
Sanitize User Input through Environment Variables

basteyy:
I recognized that cakephp is returning the variables from the server environment direct and without sanitizing. I'm not sure if that a good idea. Perhaps I'm acting far too cautiously, but in my opinion you should filter the input at least with the filter_var function.

### DIFF
--- a/src/Core/functions.php
+++ b/src/Core/functions.php
@@ -228,7 +228,7 @@ if (!function_exists('env')) {
         }
 
         if ($val !== null) {
-            return $val;
+            return filter_var($val, FILTER_SANITIZE_STRING);
         }
 
         switch ($key) {


markstory:
What problems are you solving with this change?

basteyy:
If you use env without filter_var for example for example with `env('HTTP_USER_AGENT');` you got unsanitized user input. Data enters a Web application through an untrusted source, if someone modifies his agent in the try to exploit xss attack, this change can avoid that.

garas:
This would break `env('PHP_AUTH_PW')` (used in BasicAuthenticate) if password contains some `<>` and many other legitimate uses.

It is responsibility of developer to sanitize/escape any data from untrusted sources if necessary.

markstory:
How does checking if a value is a string prevent XSS?

inoas:
What about opting out certain fields (like PHP_AUTH_PW) and by default filtering?

markstory:
@inoas Why though? Folks should be encoding their output. Why would we apply sanitization to headers and not request bodies? Those are far more likely to contain 'bad stuff'.

basteyy:
I can accept the statement that this is up to the users. It was just a proposal. I will close the pull request.

markstory:
@basteyy Thank you for taking the time to put together a pull request. If you're looking for other ways to contribute let me know, we can find an issue for you to take on :smile:

PR #12339, opened 2018-07-09
4.x Add types to core and error packages

markstory:
Add strict_types and hints to core and error packages. I had to branch out a bit to fix errors, and I've also changed the error handling code to use `Throwable` as we no longer need the PHP5 shims.

Part of #11935


src/Error/ErrorHandler.php
```diff
@@ -161,11 +162,10 @@ protected function _clearOutput()
      *
      * The PHP5 part will be removed with 4.0.
```
saeideng: I not found PHP5 part in this method

lorenzo (approved):

PR #12349, opened 2018-07-12
Add typehint to helpers.

ADmad:
Added typhints to all public methods of `View/Helper.php` and all files under `View/Helper/*`.

I skipped the 2 helpers `NumberHelper` and `TextHelper` which are basically proxies to utility classes and the tests started throws tons of errors when adding typhints to them.

Also an interesting thing I found is `HtmlHelper::tag()` takes `null` and `false` as values for 1st argument `$name` (there are tests for it) which seems quite weird to me. I would prefer changing it to only take string.

### DIFF
--- a/src/View/Helper.php
+++ b/src/View/Helper.php
@@ -68,7 +68,7 @@ class Helper implements EventListenerInterface
     /**
      * The View instance this helper is attached to
      *
-     * @var \Cake\View\View|null
+     * @var \Cake\View\View
      */
     protected $_View;
 
@@ -123,7 +123,7 @@ class Helper implements EventListenerInterface
      *
      * @return \Cake\View\View The bound view instance.
      */
-    public function getView(): ?View
+    public function getView(): View
     {
         return $this->_View;
     }
--- a/src/View/Helper.php
+++ b/src/View/Helper.php
@@ -1,4 +1,5 @@
 <?php
+declare(strict_types=1);
 /**
  * CakePHP(tm) : Rapid Development Framework (https://cakephp.org)
  * Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
@@ -67,7 +68,7 @@ class Helper implements EventListenerInterface
     /**
      * The View instance this helper is attached to
      *
-     * @var \Cake\View\View
+     * @var \Cake\View\View|null
      */
     protected $_View;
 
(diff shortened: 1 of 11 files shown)


src/View/Helper.php
```diff
@@ -67,7 +68,7 @@ class Helper implements EventListenerInterface
     /**
      * The View instance this helper is attached to
      *
-     * @var \Cake\View\View
+     * @var \Cake\View\View|null
```
dereuromark: Cant be null, see constructor.

src/View/Helper.php
```diff
@@ -122,7 +123,7 @@ public function __get($name)
      *
      * @return \Cake\View\View The bound view instance.
      */
-    public function getView()
+    public function getView(): ?View
```
dereuromark: Also cant be nullable then

src/View/Helper/FormHelper.php
```diff
@@ -223,7 +224,7 @@ class FormHelper extends Helper
     /**
      * Locator for input widgets.
      *
-     * @var \Cake\View\Widget\WidgetLocator
+     * @var \Cake\View\Widget\WidgetLocator|null
```
dereuromark: Also cant be null after constructor (at runtime).

src/View/Helper/FormHelper.php
```diff
@@ -305,7 +306,7 @@ public function __construct(View $View, array $config = [])
      * @return \Cake\View\Widget\WidgetLocator Current locator instance
      * @since 3.6.0
      */
-    public function getWidgetLocator()
+    public function getWidgetLocator(): ?WidgetLocator
```
dereuromark: Same, see also return doc block (not null here).

src/View/Helper/HtmlHelper.php
```diff
@@ -797,7 +806,7 @@ public function tableCells($data, $oddTrOptions = null, $evenTrOptions = null, $
      *
      * @param array $line Line data to render.
      * @param bool $useCount Renders the count into the row. Default is false.
-     * @return string[]
+     * @return array
```
dereuromark: Why making this less precise?

dereuromark (commented):
I agree, string $name is also the documented one.

markstory:
Refs #11935

PR #12350, opened 2018-07-12
Add typehints to Mailer.

dereuromark:
Some doc blocks were incorrect as well.

Maybe we can also get rid of some |null things in favor of just string in the neat future.

### DIFF
--- a/src/Mailer/Email.php
+++ b/src/Mailer/Email.php
@@ -167,7 +167,7 @@ class Email implements JsonSerializable, Serializable
      *
      * @var string
      */
-    protected $_domain;
+    protected $_domain = '';
 
     /**
      * The subject of the email
@@ -340,7 +340,7 @@ class Email implements JsonSerializable, Serializable
      * If null, filter_var() will be used. Use the emailPattern() method
      * to set a custom pattern.'
      *
-     * @var string
+     * @var string|null
      */
     protected $_emailPattern = self::EMAIL_PATTERN;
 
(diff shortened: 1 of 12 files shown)


markstory:
There are a few relevant looking errors in the test results.

markstory:
Refs #11935

src/Mailer/Email.php
```diff
@@ -492,7 +492,7 @@ public function getReadReceipt()
      * @return $this
```
ADmad: Shouldn't we now change the docblock too to `@return self` to match actual typehint?

src/Mailer/Mailer.php
```diff
@@ -212,7 +212,7 @@ public function viewBuilder()
      * @param array $args Method arguments
      * @return $this|mixed
      */
-    public function __call($method, $args)
+    public function __call(string $method, array $args)
```
ADmad: Adding typehints to magic methods seems quite unnecessary as these methods are called by PHP itself not the user.

dereuromark: OK

src/Mailer/Email.php
dereuromark: I think we dont have to. one is php language, the other for IDE to understand chaining.

dereuromark:
Found the issue.

    _smtpSend(?string $data, $checkCode = '250'): ?string

$checkCode is string|bool input
We should at some point fix this up to string|null input.
Not sure if that can be done for 4.0 now, though.

PR #12359, opened 2018-07-14
Add type hints to files under Validation.

ADmad:
Refs #11935

### DIFF
--- a/src/Validation/Validation.php
+++ b/src/Validation/Validation.php
@@ -900,8 +900,8 @@ class Validation
      * legal finite on this platform.
      *
      * @param mixed $check Value to check
-     * @param int|float|null $lower Lower limit
-     * @param int|float|null $upper Upper limit
+     * @param float|null $lower Lower limit
+     * @param float|null $upper Upper limit
      * @return bool Success
      */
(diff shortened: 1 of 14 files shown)


src/Validation/Validation.php
```diff
@@ -653,12 +660,12 @@ public static function decimal($check, $places = null, $regex = null)
...
      * any PHP version on a non-windows distribution
      *
-     * @param string $check Value to check
+     * @param mixed $check Value to check
      * @param bool $deep Perform a deeper validation (if true), by also checking availability of host
      * @param string|null $regex Regex to use (if none it will use built in regex)
      * @return bool Success
      */
-    public static function email($check, $deep = false, $regex = null)
+    public static function email($check, ?bool $deep = false, ?string $regex = null): bool
```
lorenzo: I think the second argument is not nullable

ADmad: The method has check like `if ($deep === false || $deep === null) {` plus there are tests which pass `null` for second param.

saeideng: tests are invalid :)

PR #12364, opened 2018-07-16
Use strict typing for Routing package.

ADmad:
Refs #11935.

### DIFF
--- a/src/Routing/RouteCollection.php
+++ b/src/Routing/RouteCollection.php
@@ -1,4 +1,5 @@
 <?php
+declare(strict_types=1);
 /**
  * CakePHP(tm) : Rapid Development Framework (https://cakephp.org)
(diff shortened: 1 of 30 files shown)


src/Routing/RouteCollection.php
```diff
@@ -233,7 +234,7 @@ protected function _getNames($url)
         if (isset($url['prefix']) && $url['prefix'] !== false) {
             $prefix = strtolower($url['prefix']);
         }
-        $controller = strtolower($url['controller']);
+        $controller = isset($url['controller']) ? strtolower($url['controller']) : null;
```
markstory: Is this a place we could use the `??` operator?

dereuromark: Not with the extra strtolower() op.

ADmad: Right, we can't.

lorenzo (approved):

Issue #12335, opened 2018-07-06
4.x - Change return types to nullable type instead of type|bool


markstory:
This is a (multiple allowed):

* [ ] bug
* [ ] enhancement
* [x] feature-discussion (RFC)

We have a number of methods in CakePHP that currently use return types like `string|bool` and `array|bool`. These methods were designed well before nullable types were conceived, and re-used the conventions that PHP standard library uses in functions like `array_search` or `strpos`. In addition to the `type|bool` return values we also sometimes use `type|null` return types. Adopting nullable types now will allow us to make to make CakePHP more consistent internally, and simplify the return types of many methods.

We have the opportunity in 4.0.0 to make these changes. While changing return types provides increased consistency, it also represents a significant and subtle set of breaking changes. For example  code like:

```php
if ($folder->create() === false) {
 //
}
```
would break in 4.x but work in 3.x

## Why discuss this

@chinpei215 raised this issue in slack a few days ago

> This will be big braking changes which contain no benefit for users. I would be surprised if PHP changes strpos() to returning null instead of false in the next major version.

I believe he has a good point that we would not be pleased if PHP decided to change its return types in such a subtle way.

### Scope of changes

There are **108 methods** potentially effected by these changes. The [full list](https://gist.github.com/markstory/5e7e6dda3e0a8b568bfc135587a90af2) of effected methods was pulled from `3.next` as we've changed several methods in 4.x already.

## Possible Approaches

I can see a few options:

1. Continue changing all the methods to use nullable string/int/array/class types. Document that this is a non-trivial breaking change in the upgrade guide.
2. Only change types on internal methods, or where the return type already includes `null`.  All other methods would be left untouched. Exceptions could be made for caching changes as they enable PSR6/16 to be adopted.
3. Leave *all* methods with `x|bool` return types alone and do not declare a return type on these methods at this point in time.
4. ? 

These options are not exhaustive and I welcome and encourage other approaches. We have already changed a handful of methods, but I want to make a decision together on how to proceed in order to minimize recurring discussions and enable us to communicate clearly with the community.

ADmad:
While changing the return type to nullable instead of `type|bool` is the technically sound thing to do I don't think it's benefits out weight the pain it will cause during migration from 3.x.

We should keep in mind that our goal is to have 4.0 as primarily a cleanup release with minimal migration effort from 3.x as long as one also removed use of deprecated code.

Hence my vote is for option 2. Only change internal methods or where it's required for PSR adoption.

[...]

PR #12365, opened 2018-07-17
Revert breaking type changes.

markstory:
As per #12335 this reverts breaking changes made in #12286

### DIFF
--- a/src/Filesystem/File.php
+++ b/src/Filesystem/File.php
@@ -318,7 +318,7 @@ class File
     /**
      * Returns the file extension.
      *
-     * @return string|bool The file extension, false if extension cannot be extracted.
+     * @return string|false The file extension, false if extension cannot be extracted.
      */
     public function ext()
     {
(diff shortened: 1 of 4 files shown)


lorenzo (approved):

src/Filesystem/File.php
```diff
@@ -450,29 +450,29 @@ public function exists(): bool
     /**
      * Returns the "chmod" (permissions) of the file.
      *
-     * @return string|null Permissions for the file, or null in case of an error
+     * @return string|bool Permissions for the file, or false in case of an error
```
dereuromark: bool must be false here
Same for the other cases, as true can never be returned (we only abuse false here for null).

dereuromark (changes requested):

src/Filesystem/File.php
markstory: PHPStan doesn't seem to have blown up so far. Why would `bool` cause issues?

markstory:
@dereuromark Yes I think we can schedule replacing the bool|other changes for 5.0. Another option would be to replace the 'Filesystem' package with an improved implementation avoiding the need to break compatibility in a subtle way.

src/Filesystem/File.php
markstory: Ok. We use `bool` quite extensively as it is a type while `false` is not. If tooling works better with `string|false` that sounds like a separate change.

dereuromark (approved):

PR #12374, opened 2018-07-19
Add missing versions to deprecation notices.

dereuromark:
This is a painfully missing info when trying to upgrade plugins and ecosystem tooling for new majors.
If you keep 3.5 compatibility, for example, you easily risk killing the whole plugin by upgrading too high.

So having the version numbers here help to find the right things to upgrade as per minimum version support in composer.json.


ADmad:
Similarly we need to be more particular about adding `@since` tag to new methods/classes.

dereuromark:
Maybe a code sniffer some day :)

PR #12378, opened 2018-07-19
Fixes #12376 bindTranslation restore fields

### LINKED ISSUE #12376: 2.10.10 TranslateBehavior fails to restore old fields when bindTranslation is used
This is a (multiple allowed):

* [x] bug
* [ ] enhancement

* CakePHP Version: 2.10.10
* Platform and Target: Apache 2.4.23, PHP 7.1.14, MySQL 5.7.14, Windows 10 x64

### What you did
I used Translate behavior to an model like so
```
public $actsAs = array(
		'Translate' => array(
			'description',
			'joinType' => 'LEFT'
                 ),
               'Containable',
	);
```
And everything works as expected.
But given the use case:
```
             $this->Model->bindTranslation(array('description' => 'description_translations'), true);
             $this->Model->find('first',array(
			 'conditions' => [
				'Model.id' => $id,
			],
			'contain' => array(
				'description_translations'
			)
	      ));
              $this->Model->find('first',array(
			'fields' => array(
				'Model.description',
			),
			'conditions' => [
				'Model.id' => $id,
			],
		));
		$this->Model->find('first',array(
			'fields' => array(
				'Model.description',
			),
			'conditions' => [
				'Model.id' => $id,
			],
		));
```
### What happened
The last call to find() will fail with:
```
Column not found: 1054 Unknown column 'Model.description' in 'field list'
```
So why does this happen?

- Looking at bindTranaslation this line https://api.cakephp.org/2.10/source-class-TranslateBehavior.html#685 executes and an value is push to array $this->runtime[$Model->alias]['restoreFields'].
- Now the first find() runs and then in afterFind() restores the fields
 https://api.cakephp.org/2.10/source-class-TranslateBehavior.html#349
- The second find() is run and in afterFind() this gets executed https://api.cakephp.org/2.10/source-class-TranslateBehavior.html#347 and `$this->runtime[$Model->alias]['fields']` is set to empty.
- Last find() tries to run but it can't because it doesn't know that is needs to make an join, because the fields are empty so no join and there is the error.

I would expect that $reset = true to bindTranslation() will just reset the fields to original after it run not to empty.
As a side note the doc is wrong https://api.cakephp.org/2.10/source-class-TranslateBehavior.html#643
According to the doc the second find() should fail but it was the third that failed.

As a fix we could replace this line https://github.com/cakephp/cakephp/blob/2.x/lib/Cake/Model/Behavior/TranslateBehavior.php#L347
with `$this->runtime[$Model->alias]['virtualFields'] = array();` 
So that if we use bindTranslation() the old fields will be used not empty. (not 100% about this fix, clearly there are multiple ways to fix this)

Or all of this could be intended and I just lost 1 day for nothing 😢

mandricmihai:
This fixes https://github.com/cakephp/cakephp/issues/12376
Makes sure that bindTranslation() when called with $reset = true, restores the fields to their original value and not empty array.

@markstory maybe add some comments about why the 2 extra finds are there?

### DIFF
--- a/lib/Cake/Test/Case/Model/Behavior/TranslateBehaviorTest.php
+++ b/lib/Cake/Test/Case/Model/Behavior/TranslateBehaviorTest.php
@@ -1144,6 +1144,16 @@ class TranslateBehaviorTest extends CakeTestCase {
 		$TestModel->bindTranslation($translations);
 
 		$result = $TestModel->find('first');
+		$TestModel->find('first', array(
+			'fields' => array(
+				'TranslatedItem.title',
+			),
+		));
+		$TestModel->find('first', array(
+			'fields' => array(
+				'TranslatedItem.title',
+			),
+		));
 		$this->assertArrayHasKey('Title', $result);
 		$this->assertArrayHasKey('content', $result['Title'][0]);
 		$this->assertArrayNotHasKey('title', $result);
--- a/lib/Cake/Model/Behavior/TranslateBehavior.php
+++ b/lib/Cake/Model/Behavior/TranslateBehavior.php
@@ -344,7 +344,7 @@ class TranslateBehavior extends ModelBehavior {
 	public function afterFind(Model $Model, $results, $primary = false) {
 		$Model->virtualFields = $this->runtime[$Model->alias]['virtualFields'];
 
-		$this->runtime[$Model->alias]['virtualFields'] = $this->runtime[$Model->alias]['fields'] = array();
+		$this->runtime[$Model->alias]['virtualFields'] = array();
 		if (!empty($this->runtime[$Model->alias]['restoreFields'])) {
 			$this->runtime[$Model->alias]['fields'] = $this->runtime[$Model->alias]['restoreFields'];
 			unset($this->runtime[$Model->alias]['restoreFields']);


mma:
Not sure why this fails

mma:
@markstory is the pull ok?Do I need to do some changes?

lib/Cake/Test/Case/Model/Behavior/TranslateBehaviorTest.php
```diff
@@ -1144,6 +1144,16 @@ public function testFieldsRestoreAfterBind() {
 		$TestModel->bindTranslation($translations);
 
 		$result = $TestModel->find('first');
+		$TestModel->find('first', array(
+			'fields' => array(
+				'TranslatedItem.title',
+			),
+		));
+		$TestModel->find('first', array(
```
bancer: Maybe assert the results of these finds?

markstory:
@mma I didn't have a chance to look at the tests in depth. I'll try to get to this over the weekend.

mma:
@markstory Were you able to find some free time to take a look?

markstory:
@mma Not yet. But it is close to the top of my list :smile:

mma:
any news?

PR #12379, opened 2018-07-20
4x database schema type hints

markstory:
Add strict_types and typehinting to Database\Schema. I had to tweak some of the doc block as the actual code used wider types than was documented.

Refs #11935


lorenzo (approved):

ADmad (changes requested):
Builds failing

ADmad (approved):

PR #12380, opened 2018-07-20
4.x Adding types to the shell package

burzum:
Refs https://github.com/cakephp/cakephp/issues/11935

### DIFF
--- a/src/Shell/CacheShell.php
+++ b/src/Shell/CacheShell.php
@@ -1,4 +1,5 @@
 <?php
+declare(strict_types = 1);
 /**
  * CakePHP(tm) : Rapid Development Framework (https://cakephp.org)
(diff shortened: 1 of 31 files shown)


markstory (commented):
You might also want to add `declare(strict_types=1)` to find more typing related issues.

ADmad (commented):
Strict typing should be enabled for the related test case files too.

src/Shell/CacheShell.php
```diff
@@ -1,4 +1,5 @@
 <?php
+declare(strict_types = 1);
```
saeideng: I think extra spaces should remove 
`_typeses = 1)` = >`_typeses=1)`

saeideng: https://github.com/cakephp/cakephp-codesniffer/pull/205

PR #12381, opened 2018-07-21
Allowing IDE to recognize Collection/sortBy() with Query

AlPri78:
A very short enhancement (method declaration) to allow right checks by IDE on Collection/sortBy() method for a Query.

### DIFF
--- a/src/ORM/Query.php
+++ b/src/ORM/Query.php
@@ -33,6 +33,7 @@ use RuntimeException;
  *
  * @see \Cake\Collection\CollectionInterface For a full description of the collection methods supported by this class
  * @method \Cake\Collection\CollectionInterface each(callable $c) Passes each of the query results to the callable
+ * @method \Cake\Collection\CollectionInterface sortBy($callback, $dir = SORT_DESC, $type = \SORT_NUMERIC) Sorts the query with the callback
  * @method \Cake\Collection\CollectionInterface filter(callable $c = null) Keeps the results using passing the callable test
  * @method \Cake\Collection\CollectionInterface reject(callable $c) Removes the results passing the callable test
  * @method bool every(callable $c) Returns true if all the results pass the callable test
--- a/src/Filesystem/Folder.php
+++ b/src/Filesystem/Folder.php
@@ -359,6 +359,7 @@ class Folder
      */
     public static function normalizePath($path)
     {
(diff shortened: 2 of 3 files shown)


src/ORM/Query.php
```diff
@@ -33,6 +33,7 @@
  *
  * @see \Cake\Collection\CollectionInterface For a full description of the collection methods supported by this class
  * @method \Cake\Collection\CollectionInterface each(callable $c) Passes each of the query results to the callable
+ * @method \Cake\Collection\CollectionInterface sortBy($callback, $dir = SORT_DESC, $type = \SORT_NUMERIC) Sorts the query with the callback
```
dereuromark: Why the leading blacklash at SORT_NUMERIC?

AlPri78: Good question. I did not drill that point but it's exactly the signature of this method in CollectionInterface.
I see in CollectionTrait/sortBy(), there is also a backslash on the default value for the second argument.
So, it could be better even to put both, or to suppress both.
- To put : I think that in a pure programming rules, it could be better to put the slashes to be sure to get the php global namespace. Thus, if someone redefine this constant in the local namespace, the results would be uncertain. 
- To suppres : I think it's not necessary to use backslash in this context because it's almost impossible that someone would redefine that kind of constant, and it results with less readable code, and all along the cakephp program the use of backslash for SORT_ASC (for example) is very different sometimes with backslash and not, but more often without.
I you and others agree to suppress, I can fix that for sortBy method in CollectionInterface and in CollectionTrait.

dereuromark: Sounds like a snowflake accident, as we dont do that anywhere in code - and probably shouldnt as it would limit the desired possibility of overwriting in certain cases.
So best to remove this and keep the code consistent here.

AlPri78: `"PS: why so many comments? Are you able to squash them into a single one afterwards?"`

I guess it's commits (not comments). If yes, you are right I will try to do it (try : because I'm a beginner on Git and Github, and it's why there's so many commits !).

`"...as we dont do that anywhere in code"`
If you look for SORT_ASC or SORT_NUMERIC you will find few backslahes...

Ok, I will purpose the fix...

markstory: The leading slashes were added to the actual code because we ran into build issues in travis and adding a slash made the problems go away.

AlPri78: 👍

PR #12383, opened 2018-07-22
Add strict typing to behaviors.

ADmad:
Refs #11935

### DIFF
--- a/src/ORM/Behavior/TimestampBehavior.php
+++ b/src/ORM/Behavior/TimestampBehavior.php
@@ -21,7 +21,7 @@ use Cake\Datasource\EntityInterface;
 use Cake\Event\EventInterface;
 use Cake\I18n\Time;
 use Cake\ORM\Behavior;
-use DateTime;
+use DateTimeInterface;
 use RuntimeException;
 use UnexpectedValueException;
 
@@ -63,7 +63,7 @@ class TimestampBehavior extends Behavior
     /**
      * Current timestamp
      *
-     * @var \DateTime
+     * @var \DateTimeInterface
      */
     protected $_ts;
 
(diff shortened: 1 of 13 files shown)


dereuromark (approved):

src/ORM/Behavior/TimestampBehavior.php
```diff
@@ -139,7 +140,7 @@ public function implementedEvents(): array
      * @param bool $refreshTimestamp If true timestamp is refreshed.
      * @return \DateTime
      */
-    public function timestamp(?DateTime $ts = null, $refreshTimestamp = false)
+    public function timestamp(?DateTime $ts = null, bool $refreshTimestamp = false): DateTime
```
markstory: Should this be `DateTimeInterface` instead?

src/ORM/Behavior/TimestampBehavior.php
```diff
@@ -136,11 +136,11 @@ public function implementedEvents(): array
...
      * automatically set to false.
      *
-     * @param \DateTime|null $ts Timestamp
+     * @param \DateTimeInterface|null $ts Timestamp
      * @param bool $refreshTimestamp If true timestamp is refreshed.
-     * @return \DateTime
+     * @return \DateTimeInterface
      */
-    public function timestamp(?DateTime $ts = null, bool $refreshTimestamp = false): DateTime
+    public function timestamp(?DateTimeInterface $ts = null, bool $refreshTimestamp = false): DateTimeInterface
```
dereuromark: As signature change this can be BC breaking and should also be noted in the migration guide IMO.
Extending classes would be in trouble.

ADmad: Huh? All the typehint additions are already going to require changing extending classes. Why is this any different which requires special mention?

markstory: Why would subclasses be in trouble? They are getting an interface that allows immutable types and mutable ones which is a wider type definition.

PR #12384, opened 2018-07-22
Add strict typing to association related files.

ADmad:
Refs #11935.

### DIFF
--- a/src/ORM/Association/BelongsToMany.php
+++ b/src/ORM/Association/BelongsToMany.php
@@ -1,4 +1,5 @@
 <?php
+declare(strict_types=1);
 /**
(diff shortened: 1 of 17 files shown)


src/ORM/Association/BelongsToMany.php
```diff
@@ -506,19 +508,19 @@ public function type()
...
      *
      * @param array $options list of options passed to attachTo method
-     * @return bool false
+     * @return array
      */
-    protected function _joinCondition($options)
+    protected function _joinCondition(array $options): array
     {
-        return false;
+        return [];
```
ADmad: This had to be changed to match parent method.

markstory:
:clap:

PR #12385, opened 2018-07-22
Phpstan 0.10

segy:
New version of phpstan has slightly changed interfaces, so I updated classes according to this changes.

### DIFF
--- a/src/Core/PluginCollection.php
+++ b/src/Core/PluginCollection.php
@@ -104,7 +104,7 @@ class PluginCollection implements Iterator, Countable
      *
      * @param string $name The plugin name to locate a path for. Will return '' when a plugin cannot be found.
      * @return string
-     * @throws Cake\Core\Exception\MissingPluginException when a plugin path cannot be resolved.
+     * @throws MissingPluginException when a plugin path cannot be resolved.
      * @internal
      */
     public function findPath($name)
--- a/src/Core/PluginCollection.php
+++ b/src/Core/PluginCollection.php
@@ -104,7 +104,7 @@ class PluginCollection implements Iterator, Countable
      *
      * @param string $name The plugin name to locate a path for. Will return '' when a plugin cannot be found.
      * @return string
-     * @throws MissingPluginException when a plugin path cannot be resolved.
+     * @throws \Cake\Core\Exception\MissingPluginException when a plugin path cannot be resolved.
      * @internal
      */
     public function findPath($name)
--- a/src/Datasource/QueryTrait.php
+++ b/src/Datasource/QueryTrait.php
@@ -481,9 +481,12 @@ trait QueryTrait
     {
         $entity = $this->first();
         if (!$entity) {
+            /** @var \Cake\ORM\Table $repository */
+            $repository = $this->getRepository();
+
             throw new RecordNotFoundException(sprintf(
                 'Record not found in table "%s"',
-                $this->getRepository()->getAlias()
+                $repository->getTable()
             ));
         }
 
--- a/src/Datasource/QueryTrait.php
+++ b/src/Datasource/QueryTrait.php
@@ -483,7 +483,7 @@ trait QueryTrait
         if (!$entity) {
             throw new RecordNotFoundException(sprintf(
                 'Record not found in table "%s"',
-                $this->getRepository()->getTable()
+                $this->getRepository()->getAlias()
             ));
         }
 
--- a/src/Http/BaseApplication.php
+++ b/src/Http/BaseApplication.php
@@ -99,7 +99,7 @@ abstract class BaseApplication implements
             $plugin = $name;
         }
         if (!$plugin instanceof PluginInterface) {
-            throw new InvalidArgumentException("The `{$plugin}` plugin does not implement Cake\Core\PluginInterface.");
+            throw new InvalidArgumentException("The " . ($name instanceof PluginInterface ? $name->getName() : (string)$name) . " plugin does not implement Cake\Core\PluginInterface.");
         }
         $this->plugins->add($plugin);
 
--- a/src/Http/BaseApplication.php
+++ b/src/Http/BaseApplication.php
@@ -99,7 +99,7 @@ abstract class BaseApplication implements
             $plugin = $name;
         }
         if (!$plugin instanceof PluginInterface) {
-            throw new InvalidArgumentException("The `{$name}` plugin does not implement Cake\Core\PluginInterface.");
+            throw new InvalidArgumentException("The `{$plugin}` plugin does not implement Cake\Core\PluginInterface.");
         }
         $this->plugins->add($plugin);
 
--- a/src/Http/Session.php
+++ b/src/Http/Session.php
@@ -390,6 +390,10 @@ class Session
             $this->start();
         }
 
+        if (!isset($_SESSION)) {
+            return false;
+        }
+
         return Hash::get($_SESSION, $name) !== null;
     }
 
@@ -406,8 +410,12 @@ class Session
             $this->start();
         }
 
+        if (!isset($_SESSION)) {
+            return null;
+        }
+
         if ($name === null) {
-            return $_SESSION;
+            return isset($_SESSION) ? $_SESSION : [];
         }
 
         return Hash::get($_SESSION, $name);
@@ -451,7 +459,7 @@ class Session
             $write = [$name => $value];
         }
 
-        $data = $_SESSION;
+        $data = isset($_SESSION) ? $_SESSION : [];
         foreach ($write as $key => $val) {
             $data = Hash::insert($data, $key, $val);
         }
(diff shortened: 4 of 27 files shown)


dereuromark:
Dont forget composer.json ( https://github.com/cakephp/cakephp/compare/phpstan-010?expand=1 ) and adjusting the failing cases.

src/Datasource/QueryTrait.php
```diff
@@ -483,7 +483,7 @@ public function firstOrFail()
         if (!$entity) {
             throw new RecordNotFoundException(sprintf(
                 'Record not found in table "%s"',
-                $this->getRepository()->getTable()
+                $this->getRepository()->getAlias()
```
saeideng: it is not same

src/Core/PluginCollection.php
```diff
@@ -104,7 +104,7 @@ protected function loadConfig()
      *
      * @param string $name The plugin name to locate a path for. Will return '' when a plugin cannot be found.
      * @return string
-     * @throws Cake\Core\Exception\MissingPluginException when a plugin path cannot be resolved.
```
ADmad: Need to add starting blackslash here to make it proper FQCN instead of changing it to just class name.

src/Http/BaseApplication.php
```diff
@@ -99,7 +99,7 @@ public function addPlugin($name, array $config = [])
             $plugin = $name;
         }
         if (!$plugin instanceof PluginInterface) {
-            throw new InvalidArgumentException("The `{$name}` plugin does not implement Cake\Core\PluginInterface.");
+            throw new InvalidArgumentException("The `{$plugin}` plugin does not implement Cake\Core\PluginInterface.");
```
saeideng: would be better to use `$name`

src/Http/Session.php
```diff
@@ -608,7 +600,7 @@ public function renew()
      */
     protected function _timedOut()
     {
-        $time = $this->read('Config.time');
+        $time = (int)$this->read('Config.time');
```
saeideng: this is not true 
see `$time !== null` in bellow

src/ORM/ResultSet.php
```diff
@@ -447,7 +449,7 @@ protected function _getTypes($table, $fields)
     {
         $types = [];
         $schema = $table->getSchema();
-        $map = array_keys(Type::map() + ['string' => 1, 'text' => 1, 'boolean' => 1]);
+        $map = array_keys((array)Type::map() + ['string' => 1, 'text' => 1, 'boolean' => 1]);
```
saeideng: Using `Type::map()` as getter is deprecated.
Use `Type::getMap()` instead.

src/Routing/RouteBuilder.php
```diff
@@ -388,7 +388,7 @@ public function namePrefix($value = null)
      */
     public function resources($name, $options = [], $callback = null)
     {
-        if (is_callable($options) && $callback === null) {
```
saeideng: why?

src/Routing/RouteBuilder.php
```diff
@@ -987,8 +987,10 @@ public function scope($path, $params, $callback = null)
             $params = [];
         }
         if (!is_callable($callback)) {
-            $msg = 'Need a callable function/object to connect routes.';
-            throw new InvalidArgumentException($msg);
+            throw new InvalidArgumentException('Need a callable function/object to connect routes.');
+        }
+        if (!is_array($params)) {
+            throw new InvalidArgumentException('Invalid params passed.');
```
saeideng: why we need to this Exception ?

src/Utility/Text.php
```diff
@@ -1034,7 +1034,7 @@ public static function parseFileSize($size, $default = false)
             $i = array_search(substr($size, -1), ['K', 'M', 'G', 'T', 'P']);
         }
         if ($i !== false) {
-            $size = substr($size, 0, $l);
+            $size = (float)substr($size, 0, $l);
```
saeideng: there was not error about this

src/Http/Session.php
```diff
@@ -390,10 +390,6 @@ public function check($name = null)
             $this->start();
         }
 
-        if (!isset($_SESSION)) {
```
dereuromark: These session changes are somewhat dangerous and should be avoided if possible.

src/Http/Session.php
dereuromark: You can cast below in the operation itself.

dereuromark: Actually, time() should be pulled up and stored in variable instead of being called twice here.
But that can be done in a separate PR.

src/ORM/Query.php
```diff
@@ -1084,6 +1084,7 @@ public function all()
     public function triggerBeforeFind()
     {
         if (!$this->_beforeFindFired && $this->_type === 'select') {
+            /** @var Table $table */
```
dereuromark: FQCN

src/ORM/Query.php
```diff
@@ -1146,11 +1147,14 @@ protected function _transformQuery()
             return;
         }
 
+        /** @var Table $repository */
```
dereuromark: Should not be necessary due to the _repository property, right? But otherwise also FQCN.

src/ORM/Table.php
```diff
@@ -2272,7 +2272,7 @@ public function saveMany($entities, $options = [])
      */
     public function delete(EntityInterface $entity, $options = [])
     {
-        $options = new ArrayObject($options + [
+        $options = new ArrayObject((array)$options + [
```
dereuromark: Should not be needed with proper doc block annotation above and is dangerous to be added as it cloaks abuse.
Better to throw exceptions IMO in future releases when the options array passed is not of type array.

src/Routing/Route/Route.php
```diff
@@ -651,7 +651,7 @@ public function match(array $url, array $context = [])
             isset($hostOptions['_port']) ||
             isset($hostOptions['_host'])
         ) {
-            $hostOptions += $context;
+            $hostOptions = (array)$hostOptions + $context;
```
dereuromark: Why needed? Should always either be an array or throw exeption.

src/Routing/RouteBuilder.php
dereuromark: Should actually be the other way around
Only check callable options if callback is null (not provided), for nano opt.

src/Routing/RouteBuilder.php
dereuromark: In case of invalid callback this sanity check seems to make sense with the current dirty input params.

src/Utility/Text.php
dereuromark: substr() returns int, so this makes sense.

src/ORM/Table.php
saeideng: it is `arrayaccess` but phpstan can not detect

src/Utility/Text.php
dereuromark: Actually, it can also return bool false, what in that case? Would just be 0 then to go into the operation below, which seems fine as well.

src/Routing/Route/Route.php
saeideng: `$hostOptions += $context;` is true

src/ORM/Table.php
dereuromark: Then it should be also added to doc block :)

saeideng: error was 
```php
  2275   Binary operation "+" between array|ArrayAccess and array('atomic' =>  
         true, 'checkRules' => true, '_primary' => true) results in an error.  
```

dereuromark: That I requested to be fixed in recent version of phpstan
Sounds like a left over issue then that should be reported there :)

ADmad: The error is valid, you can't use `+` operator for an `ArrayObject` instance and an array. But I don't think `$options` argument should be accepting an `ArrayAccess` instance.

dereuromark: :+1: I agree, but we can't fix this in a patch. As such we should documentt ArrayObject acceptance in doc block and keep the cast for now.
We can fix this for 3.next or 4.0 then properly.

ADmad: Yup, so we keep the casting for now and fix type in 4.0.

src/Routing/Route/Route.php
segy: this one gives weird error `Binary operation "+=" between (array&hasOffset('_host'))|(array&hasOffset('_port'))|(array&hasOffset('_scheme')) and array results in an error.`

probably phpstan issue, so i'm adding this to `ignoreErrors` instead

ADmad:
Lot of test failures.

ADmad:
Closing in favor of #12399. Thanks for the PR @segy.

PR #12386, opened 2018-07-23
4x - Remove more deprecated features.

markstory:
Remove deprecated features on `Plugin` and `Connection`.


lorenzo (approved):

dereuromark:
Some fails, though: Call to undefined method Cake\Core\Plugin::load()

PR #12387, opened 2018-07-23
Use const visibility and less magic strings for project level usage.

dereuromark:
Two questions:

* What do you think about using protected const visibility where those are meant to be used internally only?

* Can we use a bit less magic string usage where config values are injected from user land code? Or at least provide the option to use the const approach? Less typing (autocomplete of IDE), better discovery, docs, and refactoring possibilities here.

### DIFF
--- a/src/Validation/Validator.php
+++ b/src/Validation/Validator.php
@@ -32,11 +32,21 @@ use IteratorAggregate;
 class Validator implements ArrayAccess, IteratorAggregate, Countable
 {
     /**
+     * By using 'create' you can make fields required when records are first created.
+     */
+    public const WHEN_CREATE = 'create';
+
+    /**
+     * By using 'update', you can make fields required when they are updated.
+     */
+    public const WHEN_UPDATE = 'update';
+
+    /**
      * Used to flag nested rules created with addNested() and addNestedMany()
      *
      * @var string
      */
-    public const NESTED = '_nested';
+    protected const NESTED = '_nested';
 
     /**
      * Holds the ValidationSet objects array
@@ -85,7 +95,6 @@ class Validator implements ArrayAccess, IteratorAggregate, Countable
 
     /**
      * Constructor
-     *
      */
     public function __construct()
     {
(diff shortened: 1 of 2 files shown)


tests/TestCase/Validation/ValidatorTest.php
```diff
@@ -26,6 +26,13 @@
  */
 class ValidatorTest extends TestCase
 {
+    /**
+     * Used to flag nested rules created with addNested() and addNestedMany()
+     *
+     * @var string
+     */
+    protected const NESTED = '_nested';
```
markstory: Given that you need this here, this probably shouldn't be a protected constant on the class.

src/Validation/Validator.php
```diff
@@ -31,12 +31,22 @@
...
 {
+    /**
+     * By using 'create' you can make fields required when records are first created.
+     */
+    public const WHEN_CREATE = 'create';
+
+    /**
+     * By using 'update', you can make fields required when they are updated.
+     */
+    public const WHEN_UPDATE = 'update';
```
markstory: :+1: for more constants and fewer strings.

tests/TestCase/Validation/ValidatorTest.php
dereuromark: Given that the outside  (and as such the test) isnt supposed to use it directly :)
Not sure...

markstory: Could the tests be changed to not need 'internal' parts of the class? If not those identifiers are probably not internal.

dereuromark: It just looked a bit like protected methods (which with a trick you can also separately invoke/test).
If you need to access, using magic string in tests or a cloned constant sounds fine.
But for simplicity we can sure keep it as part of public API and to more easily access from tests etc.
Fine with me - I can revert this.

markstory: I think we can start using protected constants more in the future.

lorenzo (approved):

PR #12391, opened 2018-07-24
Fix entity class name inflection.

### LINKED ISSUE #12390: getEntityClass Regression in 3.6.8
This is a (multiple allowed):

* [x] bug
* [ ] enhancement

* CakePHP Version: 3.6.8
`$alias = Inflector::singularize(substr(array_pop($parts), 0, -5));`
was changed to:
`$alias = Inflector::classify(Inflector::singularize(Inflector::underscore(substr(array_pop($parts), 0, -5))));`
Suddenly my Entity Class "Address" is not found anymore because it is inflected to "Addres"

When I changed those line back to the previous one (in CakePHP 3.6.7), then it works as expected.

ADmad:
Closes #12390.

### DIFF
--- a/src/ORM/Table.php
+++ b/src/ORM/Table.php
@@ -773,7 +773,7 @@ class Table implements RepositoryInterface, EventListenerInterface, EventDispatc
                 return $this->_entityClass = $default;
             }
 
-            $alias = Inflector::classify(Inflector::singularize(Inflector::underscore(substr(array_pop($parts), 0, -5))));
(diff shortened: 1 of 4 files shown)


dereuromark (approved):

markusramsak:
This solves my issue. I would suggest to release 3.6.9 really soon because everyone who is using a custom Entity method will face this issue by getting a fatal error (method not found). Thanks for this quick fix.

PR #12392, opened 2018-07-24
4.x orm strict typing

ADmad:
Refs #11935

Added type hints to all method (I hope) in ORM related files expect the ones mention below:
- Methods of `ORM\Query` which are inherited from `Datasource\Query`.
- Adding typehints to 1st and 3rd args of `Marshaller::_mergeAssociation()` as per it's docblock caused tests to fail. So either the tests or the doblocks need to be fixed.

### DIFF
--- a/src/Database/Query.php
+++ b/src/Database/Query.php
@@ -248,7 +248,7 @@ class Query implements ExpressionInterface, IteratorAggregate
      * associated values for expressions
      * @return string
      */
-    public function sql(?ValueBinder $generator = null)
+    public function sql(?ValueBinder $generator = null): string
     {
         if (!$generator) {
             $generator = $this->getValueBinder();
--- a/src/Datasource/RepositoryInterface.php
+++ b/src/Datasource/RepositoryInterface.php
@@ -52,7 +52,7 @@ interface RepositoryInterface
      * @param array|\ArrayAccess $options An array that will be passed to Query::applyOptions()
      * @return \Cake\Datasource\QueryInterface
      */
-    public function find($type = 'all', $options = []);
+    public function find(string $type = 'all', $options = []);
 
     /**
      * Returns a single record after finding it by its primary key, if no record is
@@ -74,7 +74,7 @@ interface RepositoryInterface
      * @return \Cake\Datasource\EntityInterface
      * @see \Cake\Datasource\RepositoryInterface::find()
      */
-    public function get($primaryKey, $options = []);
+    public function get($primaryKey, $options = []): EntityInterface;
 
     /**
      * Creates a new Query instance for this repository
@@ -164,7 +164,7 @@ interface RepositoryInterface
      * @param array $options A list of options for the object hydration.
      * @return \Cake\Datasource\EntityInterface
      */
-    public function newEntity(?array $data = null, array $options = []);
+    public function newEntity(?array $data = null, array $options = []): EntityInterface;
 
     /**
      * Create a list of entities + associated entities from an array.
@@ -182,7 +182,7 @@ interface RepositoryInterface
      * @param array $options A list of options for the objects hydration.
      * @return \Cake\Datasource\EntityInterface[] An array of hydrated records.
      */
-    public function newEntities(array $data, array $options = []);
+    public function newEntities(array $data, array $options = []): array;
 
     /**
      * Merges the passed `$data` into `$entity` respecting the accessible
@@ -201,7 +201,7 @@ interface RepositoryInterface
      * @param array $options A list of options for the object hydration.
      * @return \Cake\Datasource\EntityInterface
      */
-    public function patchEntity(EntityInterface $entity, array $data, array $options = []);
+    public function patchEntity(EntityInterface $entity, array $data, array $options = []): EntityInterface;
 
     /**
      * Merges each of the elements passed in `$data` into the entities
@@ -221,5 +221,5 @@ interface RepositoryInterface
      * @param array $options A list of options for the objects hydration.
      * @return \Cake\Datasource\EntityInterface[]
      */
-    public function patchEntities($entities, array $data, array $options = []);
+    public function patchEntities($entities, array $data, array $options = []): array;
 }
--- a/src/ORM/Query.php
+++ b/src/ORM/Query.php
@@ -1,4 +1,5 @@
 <?php
+declare(strict_types=1);
 /**
  * CakePHP(tm) : Rapid Development Framework (https://cakephp.org)
  * Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
--- a/src/ORM/Query.php
+++ b/src/ORM/Query.php
@@ -15,6 +15,7 @@
 namespace Cake\ORM;
 
 use ArrayObject;
+use Cake\Database\Connection;
 use Cake\Database\ExpressionInterface;
 use Cake\Database\Query as DatabaseQuery;
 use Cake\Database\TypedResultInterface;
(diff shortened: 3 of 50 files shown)


src/ORM/Table.php
```diff
@@ -581,7 +581,7 @@ public function getPrimaryKey()
     /**
      * Sets the display field.
      *
-     * @param string|array $fields Name to be used as display field.
+     * @param array $field Name to be used as display field.
```
ADmad: Shouldn't the type be kept to `string|array`?

src/ORM/Query.php
```diff
@@ -15,15 +15,18 @@
 namespace Cake\ORM;
```
saeideng: missing `declare(strict_types=1);`

tests/test_app/Plugin/TestPlugin/src/Model/Table/TestPluginCommentsTable.php
```diff
@@ -21,7 +21,7 @@
  */
```
saeideng: missing `declare(strict_types=1);`

tests/test_app/TestApp/Model/Table/ArticlesTable.php
```diff
@@ -19,7 +19,7 @@
  */
```
saeideng: missing `declare(strict_types=1);`

tests/test_app/TestApp/Model/Table/ArticlesTagsTable.php
```diff
@@ -19,7 +19,7 @@
  */
```
saeideng: missing `declare(strict_types=1);`

tests/test_app/TestApp/Model/Table/AuthorsTable.php
```diff
@@ -20,7 +20,7 @@
  */
```
saeideng: missing `declare(strict_types=1);`

tests/test_app/TestApp/Model/Table/I18nTable.php
```diff
@@ -18,12 +18,12 @@
  */
```
saeideng: missing `declare(strict_types=1);`

tests/test_app/TestApp/Model/Table/PaginatorPostsTable.php
```diff
@@ -26,7 +26,7 @@ class PaginatorPostsTable extends Table
      *
```
saeideng: missing `declare(strict_types=1);`

tests/test_app/TestApp/Model/Table/TagsTable.php
```diff
@@ -19,7 +19,7 @@
  */
```
saeideng: missing `declare(strict_types=1);`

src/Database/Query.php
```diff
@@ -248,7 +248,7 @@ public function rowCountAndClose()
      * associated values for expressions
```
saeideng: missing `declare(strict_types=1);`

src/Datasource/RepositoryInterface.php
```diff
@@ -74,7 +74,7 @@ public function find($type = 'all', $options = []);
      * @return \Cake\Datasource\EntityInterface
      * @see \Cake\Datasource\RepositoryInterface::find()
      */
-    public function get($primaryKey, $options = []);
+    public function get($primaryKey, $options = []): EntityInterface;
```
inoas: If getPrimaryKey() is array every $primaryKey type hint should be, too?

tests/test_app/TestApp/Model/Table/ArticlesTable.php
dereuromark: No need to repeat this more than once!

src/Datasource/RepositoryInterface.php
dereuromark: It can be array or string afaik. That's the dirtyness here in the return value of this method.

tests/TestCase/TestSuite/TestCaseTest.php
```diff
@@ -361,7 +361,7 @@ public function testGetMockForModel()
 
         $this->assertInstanceOf('TestApp\Model\Table\PostsTable', $Posts);
         $this->assertNull($Posts->save($entity));
-        $this->assertNull($Posts->getTable());
+        $this->assertSame('', $Posts->getTable());
```
dereuromark: Should that throw exception instead?

ADmad: This is testing model instance returned by `TestCase::getMockForModel()`. I think it's intended that is behaves a bit differently.

dereuromark: Or should we better keep nullable here?

tests/test_app/TestApp/Model/Table/ArticlesTable.php
```diff
@@ -19,7 +19,7 @@
...
 class ArticlesTable extends Table
 {
-    public function initialize(array $config)
+    public function initialize(array $config): void
```
lorenzo: I think methods that are meant to always be overridden by the user should not have these type of changes.

dereuromark: Isnt it especially important here to not expect it to return and follow right signature?

lorenzo: I’m not sure that it is specially important to be honest. It is nice to have, but this particular method does not create a lot of confusion about what the return value should be

dereuromark: So why not having it?
The IDE will more easily be able to completely typehint it if the parent method already is fully equipped here.

dereuromark: Most importantaly code becomes consistent. Doing it sometimes, sometimes not is confusing to everyone IMO. And sniffers in the future can't be used to assert and auto-fix.

markstory:
>  So either the tests or the doblocks need to be fixed.

I would lean towards changing the doc blocks for the Marshaller. It is one of the few places in the ORM we need to be lenient as we're taking in arbitrary data from the request and arbitrary userland data.

ADmad:
@markstory Could you please update the docblock for `Marshaller::_mergeAssociation()` and the methods it calls appropriately when merging?

lorenzo:
Is the goal of 4.0 to be a replacement from 3.7 with no further changes, provided that you already resolved all of the deprecation warning? If that is the case, then adding `void` to `initialize` as the return type already breaks that promise.

Otherwise, we need that documented in the migration guide.

lorenzo:
Actually that would be a problem for all methods we encourage to override in Table. We will have another 3.6 situation, were plugins will lag behind and cannot be used for both 3.6 an 4.0.

dereuromark:
3.6 and 4.0 would never work out.
With your too hard deprecations you already killed even 3.x minor compatibility .

Having a major as separation however is expected. Migration effort using auto fixer sniffers is minimal.

markstory:
@lorenzo The goals we've communicated thus far are that after updating for deprecations, changed interfaces and typehinting the application should work.  In 7.2 userland hook methods shouldn't need to change as widening in sub-classes is acceptable.

ADmad:
ticktock

markstory:
@ADmad I'm going to verify my claim on type widening before merging this.

markstory:
I took a look and users will need to update the signatures of all hook methods we're adding typehints to. I tested out with `cakephp/app` on PHP 7.2.5 and had to update a bunch of methods before I could get past all the fatal errors.  The fatal errors were emitted without strict_types being set in the application code.

If we were to not add hints on methods that *could* be extended with the goal of no breaks, then we'd have to not add typehints anywhere. I don't think that is a good plan given the previous communication and goals we've set out for 4.x which was to add typing wherever possible.

We could instead invest time updating [rector](https://github.com/rectorphp/rector/blob/master/config/level/cakephp/cakephp36.yml) for 4.0. This would give users a tool not unlike `cake upgrade`. Unlike the upgrade from 1.x to 2.x the changes for going to 4.x are generally pretty mechanical thus far.

lorenzo:
Using rector makes sense

PR #12393, opened 2018-07-25
4.x Adding strict types to the Http Client

burzum:
#11935

Before merging this please take a look at the _getBody() and bod() methods. They're inconsistent and because of this I haven't yet added a return type there. I'm not sure if the code should be corrected or the doc blocks.

### DIFF
--- a/src/Http/Client/Adapter/Stream.php
+++ b/src/Http/Client/Adapter/Stream.php
@@ -1,4 +1,5 @@
 <?php
+declare(strict_types=1);
 /**
  * CakePHP(tm) : Rapid Development Framework (https://cakephp.org)
  * Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
@@ -65,7 +66,7 @@ class Stream implements AdapterInterface
     /**
      * {@inheritDoc}
      */
-    public function send(Request $request, array $options)
+    public function send(Request $request, array $options): array
     {
         $this->_stream = null;
         $this->_context = null;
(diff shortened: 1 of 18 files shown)


src/Http/Client/Adapter/Stream.php
```diff
@@ -114,14 +115,14 @@ public function createResponses($headers, $content)
...
-    protected function _buildContext(Request $request, $options)
+    protected function _buildContext(Request $request, array $options): void
     {
         $this->_buildContent($request, $options);
         $this->_buildHeaders($request, $options);
         $this->_buildOptions($request, $options);
 
         $url = $request->getUri();
-        $scheme = parse_url($url, PHP_URL_SCHEME);
+        $scheme = parse_url((string)$url, PHP_URL_SCHEME);
```
ADmad: This makes me sad. When a class has `__toString()` method explicit type casting shouldn't be necessary.

src/Http/Client/FormData.php
```diff
@@ -239,7 +240,7 @@ public function contentType()
      *
      * @return string
      */
-    public function __toString()
+    public function __toString(): string
```
ADmad: It would be best to avoid adding type hint's to PHP's magic methods to be safe for future.

ADmad:
I pushed some updates.

src/Http/Client/FormData.php
dereuromark: What do you mean?

src/Http/Client/Adapter/Stream.php
burzum: How do you know it has __toString() What if the implementation of the object changes? We can ensure this method exists on the object.

src/Http/Client/FormData.php
markstory: I wouldn't expect PHP to change the contract on `__toString` any time soon.

markstory (approved):
Changes look good to me.

PR #12394, opened 2018-07-25
Add strict typing to Http\Middleware.

ADmad:
Refs #11935.

### DIFF
--- a/src/Http/Middleware/EncryptedCookieMiddleware.php
+++ b/src/Http/Middleware/EncryptedCookieMiddleware.php
@@ -79,9 +79,9 @@ class EncryptedCookieMiddleware
      * @param \Psr\Http\Message\ServerRequestInterface $request The request.
      * @param \Psr\Http\Message\ResponseInterface $response The response.
      * @param callable $next The next middleware to call.
-     * @return \Cake\Http\Response A response.
+     * @return \Psr\Http\Message\ResponseInterface A response.
      */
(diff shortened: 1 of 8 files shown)


tests/TestCase/Http/Middleware/BodyParserMiddlewareTest.php
```diff
@@ -361,6 +376,8 @@ public function testInvokeXmlInternalEntities()
         $response = new Response();
         $next = function ($req, $res) {
```
dereuromark: We should also start the habit of typehinting closures, this way everyone and tooling can look into the object methods used there. Especially in tests this doesnt hurt. But can also be another follow up in the future.

src/Http/Middleware/EncryptedCookieMiddleware.php
```diff
@@ -78,9 +79,9 @@ public function __construct(array $cookieNames, $key, $cipherType = 'aes')
      * @param \Psr\Http\Message\ServerRequestInterface $request The request.
      * @param \Psr\Http\Message\ResponseInterface $response The response.
      * @param callable $next The next middleware to call.
-     * @return \Psr\Http\Message\ResponseInterface A response.
+     * @return \Cake\Http\Response A response.
      */
-    public function __invoke(ServerRequestInterface $request, ResponseInterface $response, $next)
+    public function __invoke(ServerRequestInterface $request, ResponseInterface $response, callable $next): Response
```
markstory: Why change this? We do return the interface here.

PR #12396, opened 2018-07-25
Add supported for prefixed ErrorController.

ADmad:
If error occurs on a prefix URL, exception renderer first tries to use
`ErrorController` from that prefix and if one does not exist falls back to the
non prefixed `ErrorController`.

### DIFF
--- a/src/Error/ExceptionRenderer.php
+++ b/src/Error/ExceptionRenderer.php
@@ -22,6 +22,7 @@ use Cake\Core\Exception\MissingPluginException;
 use Cake\Event\Event;
 use Cake\Http\Exception\HttpException;
 use Cake\Http\Response;
+use Cake\Http\ServerRequest;
 use Cake\Http\ServerRequestFactory;
 use Cake\Routing\DispatcherFactory;
 use Cake\Routing\Router;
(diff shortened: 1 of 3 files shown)


ADmad:
While working on this I noticed `ExceptionRenderer` has many public properties. I'll make a separate PR to change them to protected.

src/Error/ExceptionRenderer.php
```diff
@@ -93,9 +93,9 @@ class ExceptionRenderer implements ExceptionRendererInterface
      * code error depending on the code used to construct the error.
      *
      * @param \Exception $exception Exception.
-     * @param \Psr\Http\Message\ServerRequestInterface $request The request - if this is set it will be used instead of creating a new one
+     * @param \Cake\Http\ServerRequest $request The request - if this is set it will be used instead of creating a new one
      */
-    public function __construct(Exception $exception, ServerRequestInterface $request = null)
+    public function __construct(Exception $exception, ServerRequest $request = null)
```
ADmad: This was changed because phpstan complained about use of `$request->getParam()` since `getParam()` isn't present on the interface.

PR #12400, opened 2018-07-27
4.x - Add typing to Database/Statement

markstory:
I had to widen the types on fetch() and fetchAll() as we not only take the CakePHP constants, but also PDO constants which are integers.

Refs #11935

### DIFF
--- a/src/Database/Statement/BufferedStatement.php
+++ b/src/Database/Statement/BufferedStatement.php
@@ -99,7 +99,7 @@ class BufferedStatement implements Iterator, StatementInterface
     /**
      * {@inheritDoc}
      */
-    public function bindValue($column, $value, string $type = 'string'): void
+    public function bindValue($column, $value, $type = 'string'): void
     {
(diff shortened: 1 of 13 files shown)


lorenzo (approved):

src/Database/Statement/BufferedStatement.php
```diff
@@ -337,7 +338,7 @@ public function next()
      *
      * @return \Cake\Database\StatementInterface
      */
-    public function getInnerStatement()
+    public function getInnerStatement(): \Cake\Database\StatementInterface
```
ADmad: Should add "use" statement instead of using FQCN. There are other such cases to in the PR.

dereuromark: There are sniffers that can automate this. We have one in spryker namespace for example.

markstory: I ran a tool, and didn't see this when I was cleaning up. I'll fix this now.

PR #12404, opened 2018-07-27
4.x http strict types

ADmad:
Refs #11935

Sorry for the large PR. I release a bit later that I should have broken it down into smaller parts.


ADmad:
@markstory The `Core` package has been checked off on #11935 but I see that many methods of `HttpApplicationInterface` and `PluginApplicationInterface` have not been type hinted. Is that intentional?

markstory:
> Is that intentional?

No. That is a mistake on my part. I spot checked the directory and saw that some classes were done.  I'll uncheck it.

ADmad:
> No. That is a mistake on my part. I spot checked the directory and saw that some classes were done. I'll uncheck it.

No problem, I'll update them.

src/Http/Response.php
```diff
@@ -655,7 +656,7 @@ public function getType()
      * @param string $contentType Either a file extension which will be mapped to a mime-type or a concrete mime-type.
      * @return static
      */
-    public function withType($contentType)
+    public function withType(string $contentType): self
```
ADmad: I couldn't use the return type as `ResponseInterface` here as it cause [this](https://travis-ci.org/cakephp/cakephp/jobs/409132848#L881).

Makes me think using return type as `ResponseInterface` for the `with*` methods is incorrect and all should be changed to `self`.

src/Event/EventDispatcherInterface.php
```diff
@@ -50,7 +50,7 @@ public function dispatchEvent(string $name, $data = null, $subject = null): Even
      * @param \Cake\Event\EventManagerInterface $eventManager the eventManager to set
      * @return \Cake\Event\EventDispatcherInterface
      */
-    public function setEventManager(EventManagerInterface $eventManager): self;
+    public function setEventManager(EventManagerInterface $eventManager): EventDispatcherInterface;
```
markstory: :+1: `self` was not a good idea here.

src/Http/Cookie/Cookie.php
```diff
@@ -370,7 +371,7 @@ public function withHttpOnly($httpOnly)
      * @return void
      * @throws \InvalidArgumentException
      */
-    protected function validateBool($value)
+    protected function validateBool(bool $value): void
```
markstory: Do we need to keep this method? The type hint will raise a `TypeError` when it fails. The method body won't run unless the parameter is a `bool` now.

src/Http/Exception/ForbiddenException.php
```diff
@@ -26,10 +29,10 @@ class ForbiddenException extends HttpException
      * Constructor
      *
      * @param string|null $message If no message is given 'Forbidden' will be the message
-     * @param int $code Status code, defaults to 403
+     * @param int|null $code Status code, defaults to 403
      * @param \Exception|null $previous The previous exception.
      */
-    public function __construct($message = null, $code = null, $previous = null)
+    public function __construct(?string $message = null, ?int $code = null, ?Exception $previous = null)
```
markstory: Should this be `Throwable`? We likely didn't use it before because of PHP5 compatibility.

src/Http/ServerRequest.php
```diff
@@ -237,11 +238,8 @@ class ServerRequest implements ServerRequestInterface
      *
      * @param array $config An array of request data to create a request with.
      */
-    public function __construct($config = [])
+    public function __construct(array $config = [])
     {
-        if (is_string($config)) {
-            trigger_error('ServerRequest no longer accepts a string as its argument.', E_USER_WARNING);
```
markstory: :clap:

src/Http/Cookie/Cookie.php
ADmad: Nope, with `bool` typehints added for the `$http` and `$secure` arguments this method is no longer necessary. I'll remove it.

src/Http/Exception/ForbiddenException.php
ADmad: Yup, it was auto added by the tool based on docblock. I'll update it to `Throwable`.

PR #12405, opened 2018-07-27
Fix/improve PSR-7 Uploaded file validation support

ndm2:
`imageSize()` wasn't working, and since `extension()` supports file uploads, it might be worth adding PSR-7 support.

### DIFF
--- a/src/Validation/Validation.php
+++ b/src/Validation/Validation.php
@@ -775,13 +775,15 @@ class Validation
     /**
      * Checks that value has a valid file extension.
      *
-     * @param string|array $check Value to check
+     * @param string|array|\Psr\Http\Message\UploadedFileInterface $check Value to check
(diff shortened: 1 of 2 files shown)


src/Validation/Validation.php
```diff
@@ -775,13 +775,15 @@ public static function equalTo($check, $comparedTo)
...
+     * @param string|array|\Psr\Http\Message\UploadedFileInterface $check Value to check
      * @param array $extensions file extensions to allow. By default extensions are 'gif', 'jpeg', 'png', 'jpg'
      * @return bool Success
      */
     public static function extension($check, $extensions = ['gif', 'jpeg', 'png', 'jpg'])
     {
-        if (is_array($check)) {
+        if ($check instanceof UploadedFileInterface) {
+            return static::extension($check->getClientFilename(), $extensions);
+        } elseif (is_array($check)) {
```
dereuromark: No else after early return necessary.

PR #12413, opened 2018-07-30
Add type hints to remaining methods of View package files.

ADmad:
Refs #11935 

Adding type hints to methods I had skipped in earlier passes.

### DIFF
--- a/src/View/Widget/MultiCheckboxWidget.php
+++ b/src/View/Widget/MultiCheckboxWidget.php
@@ -235,11 +235,11 @@ class MultiCheckboxWidget implements WidgetInterface
         }
         $isArray = is_array($selected);
         if (!$isArray) {
-            return (string)$key === (string)$selected;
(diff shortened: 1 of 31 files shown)


src/View/Widget/MultiCheckboxWidget.php
```diff
@@ -248,7 +249,7 @@ protected function _isSelected($key, $selected)
      * @param array|bool|null $disabled The disabled values.
      * @return bool
      */
-    protected function _isDisabled($key, $disabled)
+    protected function _isDisabled(string $key, $disabled): bool
```
markstory: We could remove the string casts inside this method and `_isSelected`.

ADmad: I already did.

markstory: :clap:

PR #12414, opened 2018-07-31
Add hints to Database/Driver & Dialect.

markstory:
Refs #11935


markstory:
PHPStan will be fixed on next merge with 3.next.

PR #12416, opened 2018-08-01
Remove hints on controller hook methods.

markstory:
These hints don't fully capture the possible return types. Having to explicitly return `null` seems like an annoyance for end users and doesn't allow us to make CakePHP any more typesafe as we still have to handle null|Response everywhere.

Refs #11935

### DIFF
--- a/src/Controller/Controller.php
+++ b/src/Controller/Controller.php
@@ -777,12 +777,11 @@ class Controller implements EventListenerInterface, EventDispatcherInterface
      * or perform logic that needs to happen before each controller action.
      *
      * @param \Cake\Event\EventInterface $event An Event instance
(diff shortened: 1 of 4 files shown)


markstory:
Ping @cleptric @dereuromark

dereuromark:
I can see the argument for the void vs null annoyance.
I am just wondering how we could ever have a sniffer to help here if we make exceptions all over the place.
But given the fact that I don't have any good alternative that would justify the call for consistency, I am somewhat ok with keeping these callbacks the exception to the rule.

cleptric:
I'm fine with this for now as well :)

cleptric (approved):

PR #12418, opened 2018-08-01
3.7 - remove phpstan/phpstan-phpunit 

saeideng:


phpstan.neon
```diff
@@ -32,6 +29,7 @@ parameters:
         - '#Call to an undefined method Traversable::getInnerIterator\(\)#'
         - '#PHPDoc tag @throws with type PHPUnit\\Exception is not subtype of Throwable#'
         - '#Binary operation "\+=" between \(array\&hasOffset\(.*\)\) and array results in an error#'
+        - '#Call to an undefined method PHPUnit\\Framework\\MockObject\\MockObjec#'
```
dereuromark: missing t

dereuromark:
Is this really the best way forward here?

saeideng:
>Is this really the best way forward here?
using of phpstan is hard in local , when  you want to use global phpstan ( PHAR archive) + with this dependency
it uses just in `Cake\TestSuite\TestCase::loadPlugins()`

dereuromark:
That explanation would have been nice directly in the PR comment/description :)

ADmad:
Yeah it's better to just ignore that error instead of adding an extra dependency.

ADmad:
Errors reported by phpstan need to be fixed :)

PR #12419, opened 2018-08-01
4.0 - clean up deprecations and add strict_types (http\response)

saeideng:
Refs #11935

### DIFF
--- a/tests/TestCase/Http/BaseApplicationTest.php
+++ b/tests/TestCase/Http/BaseApplicationTest.php
@@ -196,9 +196,7 @@ class BaseApplicationTest extends TestCase
             [$this->path]
         );
         $app->addPlugin('ParentPlugin');
-        $this->deprecated(function () use ($app) {
-            $app->pluginBootstrap();
-        });
+        $app->pluginBootstrap();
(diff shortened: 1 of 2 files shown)


tests/TestCase/Http/BaseApplicationTest.php
```diff
@@ -196,9 +196,7 @@ public function testPluginBootstrapRecursivePlugins()
             [$this->path]
         );
         $app->addPlugin('ParentPlugin');
-        $this->deprecated(function () use ($app) {
-            $app->pluginBootstrap();
-        });
+        $app->pluginBootstrap();
```
ADmad: This call was wrapped in `deprecated()` earlier shouldn't the method be removed all together in 4.x?

saeideng: no
deprecated class (Plugin::load ) replaced in bellow  commit 
 https://github.com/cakephp/cakephp/commit/d78f0e23620a7259534e7b95fea2ccd4ad8f6c36#diff-7dfbdbac6f63af358352d41a16afddd9
so we can remove wrapper here

PR #12421, opened 2018-08-01
Remove unneeded View::$hasRendered flag.

ADmad:
Calling View::render() multiple times should simply render again with given arguments and existing context.


markstory:
Travis thinks you have some more work to do 😛

ADmad:
Fixed. The phpcs and phpstan build failures are unrelated.

PR #12426, opened 2018-08-02
Use "iterable" type hint and is_iterable() check.

ADmad:


src/Collection/Iterator/InsertIterator.php
```diff
@@ -65,7 +65,7 @@ class InsertIterator extends Collection
      * @param array|\Traversable $values The source collection from which the values will
      * be inserted at the specified path.
      */
-    public function __construct($into, $path, $values)
+    public function __construct(iterable $into, $path, iterable $values)
```
markstory: Should we update the doc block types as well?

markstory (commented):
NIce work.

src/Collection/Iterator/InsertIterator.php
markstory: Lets find out!

ADmad: Done.

PR #12429, opened 2018-08-04
4.x - Add type hints to database package

markstory:
I had to move the part lists into Query::traverse() to maintain interface consistentcy with ExpressionInterface. Moving part lists also required a more clunky implementation for SQLServer, but that driver already has a few clunky parts.

A few other types needed to become nullable as we use them in a nullable way in test cases. I was concerned that removing those tests would break real-world usage.

Refs #11935

### DIFF
--- a/src/Database/Dialect/SqlserverDialectTrait.php
+++ b/src/Database/Dialect/SqlserverDialectTrait.php
@@ -19,6 +19,8 @@ use Cake\Database\Expression\OrderByExpression;
 use Cake\Database\Expression\UnaryExpression;
 use Cake\Database\ExpressionInterface;
 use Cake\Database\Query;
+use Cake\Database\QueryCompiler;
+use Cake\Database\Schema\BaseSchema;
 use Cake\Database\Schema\SqlserverSchema;
 use Cake\Database\SqlDialectTrait;
 use Cake\Database\SqlserverCompiler;
@@ -69,7 +71,7 @@ trait SqlserverDialectTrait
             $query->order($query->newExpr()->add('(SELECT NULL)'));
         }
 
-        if ($this->_version() < 11 && $offset !== null) {
+        if ($this->version() < 11 && $offset !== null) {
             return $this->_pagingSubquery($query, $limit, $offset);
         }
 
@@ -81,8 +83,7 @@ trait SqlserverDialectTrait
      *
      * @return int
      */
-    // @codingStandardsIgnoreLine
-    public function _version()
+    protected function version()
     {
         $this->connect();
 
@@ -324,9 +325,9 @@ trait SqlserverDialectTrait
      * Used by Cake\Schema package to reflect schema and
      * generate schema.
      *
-     * @return \Cake\Database\Schema\SqlserverSchema
+     * @return \Cake\Database\Schema\BaseSchema
      */
-    public function schemaDialect()
+    public function schemaDialect(): BaseSchema
     {
         return new SqlserverSchema($this);
     }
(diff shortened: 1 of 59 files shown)


src/Database/Log/QueryLogger.php
```diff
@@ -45,7 +45,7 @@ public function log(LoggedQuery $query)
      * @param \Cake\Database\Log\LoggedQuery $query to be written in log
      * @return void
      */
-    protected function _log($query)
+    protected function _log(\Cake\Database\Log\LoggedQuery $query): void
```
ADmad: Need to add "use" statement instead. Same for `_interpolate()` below.

src/Database/SqlserverCompiler.php
```diff
@@ -62,7 +54,7 @@ class SqlserverCompiler extends QueryCompiler
      * @param \Cake\Database\ValueBinder $generator the placeholder generator to be used in expressions
      * @return string
      */
-    protected function _buildInsertPart($parts, $query, $generator)
+    protected function _buildInsertPart(array $parts, \Cake\Database\Query $query, \Cake\Database\ValueBinder $generator): string
```
ADmad: FQCN need to be changed.

src/Database/Dialect/SqlserverDialectTrait.php
```diff
@@ -335,32 +335,32 @@ public function schemaDialect(): BaseSchema
     /**
      * Returns a SQL snippet for creating a new transaction savepoint
      *
-     * @param string $name save point name
+     * @param string|int $name save point name
```
dereuromark: Why do we need to allow int here?

src/Datasource/QueryInterface.php
```diff
@@ -159,7 +159,7 @@ public function count(): int;
      * $query->limit($query->newExpr()->add(['1 + 1'])); // LIMIT (1 + 1)
      * ```
      *
-     * @param int $num number of records to be returned
+     * @param int|mixed $num number of records to be returned
```
dereuromark: The mixed here sounds bad - as an addition it makes it pretty much worthless to expect and input type.

src/Database/Query.php
```diff
@@ -277,12 +309,12 @@ public function sql(?ValueBinder $generator = null): string
...
      * @param callable $visitor A function or callable to be executed for each part
-     * @param array $parts The query clauses to traverse
      * @return $this
      */
-    public function traverse(callable $visitor, array $parts = [])
+    public function traverse(callable $visitor)
     {
-        $parts = $parts ?: array_keys($this->_parts);
+        $partList = "_{$this->type()}Parts";
+        $parts = $this->{$partList} ?? array_keys($this->_parts);
```
lorenzo: I would really like to keep the parts in the query compiler, unless you think this is a better idea.

One way to keep that design is by adding a traverseParts() function to the query, with the signature and implementation this function had before.

src/Datasource/QueryInterface.php
markstory: Datasource doesn't contain a way to describe a QueryExpression which is what the database implementation can accept in addition to an int. Ither implementations may accept othe 'expression' style objects as well.

src/Database/Query.php
markstory: I am not convinced what I have is great. Adding a `traverseParts` method to Query sounds good to me.

src/Database/Dialect/SqlserverDialectTrait.php
markstory: Tests were failing as some of the db implementationa were paasing integers.

dereuromark: I see, so casting beforehand is not feasible here.

PR #12436, opened 2018-08-06
Remove type hints from 1st argument of validation methods.

ADmad:
Since they are meant to validate user input the argument could be
of any type and methods should return boolean without throwing any errors.

### DIFF
--- a/src/Validation/Validation.php
+++ b/src/Validation/Validation.php
@@ -164,11 +164,11 @@ class Validation
      */
     public static function cc($check, $type = 'fast', $deep = false, $regex = null): bool
     {
-        if (!is_scalar($check)) {
+        if (!is_numeric($check)) {
             return false;
         }
 
-        $check = str_replace(['-', ' '], '', $check);
+        $check = str_replace(['-', ' '], '', (string)$check);
         if (mb_strlen($check) < 13) {
             return false;
         }
@@ -614,13 +614,17 @@ class Validation
      * - true => Any number of decimal places greater than 0, or a float|double. The '.' is required.
      * - 1..N => Exactly that many number of decimal places. The '.' is required.
      *
-     * @param float $check The value the test for decimal.
+     * @param mixed $check The value the test for decimal.
      * @param int|bool|null $places Decimal places.
      * @param string|null $regex If a custom regular expression is used, this is the only validation that will occur.
      * @return bool Success
      */
     public static function decimal($check, $places = null, ?string $regex = null): bool
     {
+        if (!is_scalar($check)) {
+            return false;
+        }
+
         if ($regex === null) {
             $lnum = '[0-9]+';
             $dnum = "[0-9]*[\.]{$lnum}";
@@ -649,7 +653,7 @@ class Validation
         $decimalPoint = $formatter->getSymbol(NumberFormatter::DECIMAL_SEPARATOR_SYMBOL);
         $groupingSep = $formatter->getSymbol(NumberFormatter::GROUPING_SEPARATOR_SYMBOL);
 
-        $check = str_replace([$groupingSep, $decimalPoint], ['', '.'], $check);
+        $check = str_replace([$groupingSep, $decimalPoint], ['', '.'], (string)$check);
 
         return static::_check($check, $regex);
     }
@@ -1486,7 +1490,7 @@ class Validation
     /**
      * Check that the input value is a 6 digits hex color.
      *
-     * @param string|array $check The value to check
+     * @param mixed $check The value to check
      * @return bool Success
      */
     public static function hexColor($check): bool
--- a/src/Validation/Validation.php
+++ b/src/Validation/Validation.php
@@ -1343,6 +1343,10 @@ class Validation
      */
     public static function geoCoordinate($value, array $options = []): bool
     {
+        if (!is_scalar($value)) {
+            return false;
+        }
+
         $options += [
             'format' => 'both',
             'type' => 'latLong',
--- a/src/Validation/Validation.php
+++ b/src/Validation/Validation.php
@@ -353,7 +353,7 @@ class Validation
      */
     public static function containsNonAlphaNumeric($check, int $count = 1): bool
     {
-        if (!is_scalar($check)) {
+        if (!is_string($check)) {
             return false;
         }
 
@@ -369,7 +369,7 @@ class Validation
      * @param string|null $regex If $check is passed as a string, $regex must also be set to valid regular expression
      * @return bool Success
      */
-    public static function custom(string $check, ?string $regex = null): bool
+    public static function custom($check, ?string $regex = null): bool
     {
         if ($regex === null) {
             static::$errors[] = 'You must define a regular expression for Validation::custom()';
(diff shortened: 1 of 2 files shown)


src/Validation/Validation.php
```diff
@@ -353,7 +353,7 @@ public static function compareFields($check, string $field, string $operator, ar
      */
     public static function containsNonAlphaNumeric($check, int $count = 1): bool
     {
-        if (!is_scalar($check)) {
+        if (!is_string($check)) {
```
chinpei215: This should not be changed. (int)-1 will be now return false, for example. Also, objects which implement __toString() will be no longer pass (It would be very rare case though). Instead, cast to string before passing to preg_match().
The same can be said to other is_string() check added in this change.

ADmad: POST data can only be string or array, so it really worth it worry about other types?
Also trying to type cast an array to string generates notice `Notice: Array to string conversion` https://3v4l.org/iTHHB.

So `is_string()` check is the right thing to do IMO.

ADmad: For others method like `luhn()`, `latitude()` etc. where numeric value is acceptable there `is_scalar()` check followed by string typecasting is used.

dereuromark: Cant it also come from json decoded data and from api data?

chinpei215: No. I think you may be surprised, but cake applications allow users to POST float/integer/boolean by default.

chinpei215: Sorry, @dereuromark has been already mentioned it earlier than me 😛

ADmad: I overlooked the fact that it could be json decoded data.

That said why would anyone intentionally want to use "containsNonAlphaNumeric" for non string data?

Similarly other methods like max/minLength() are only meant for string and should reject non string data.

chinpei215: > That said why would anyone intentionally want to use containsNonAlphaNumeric() for non string data?

Now it works so. If we want to change the behavior, we would need to leave it our migration guide.

> Similarly other methods like max/minLength() are only meant for string and should reject non string data.

Those can be used to restrict number of digits for integer or float values. So I don't think it is a good idea to change the behaviors.

ADmad: > Now it works so. If we want to change the behavior, we would need to write it in our migration guide.

The method checks if string contains one or more non-alphanumeric characters. So integers are going to fail validation even after conversion to string and I don't think any sane individual would expect this method to convert float value like `1.2` to string and then let it pass. So there's no change in behavior which needs to be documented.

> Those can be used to restrict number of digits for integer or float values. So I don't think it is a good idea to change the behaviors.

I'll update those to allow strings or numeric values.

ADmad: There's actually existing test case for `lengthBetween()` which fails for integer input.

https://github.com/cakephp/cakephp/blob/abbb678b23de16d0e4d15367804d7ae741510842/tests/TestCase/Validation/ValidationTest.php#L171

Hence `min/maxLength()` should also not accept integers.

chinpei215: The is_string() checks in lengthBetween() was added in [a security patch](https://github.com/cakephp/cakephp/commit/c4c27018bccde76ce3955b36430febd1786e62c5) in the past. In the first version of the patch, it used is_string(). But since it would break existing apps, I said:
> I think that type checks in the cc() and luhn() would be better `!is_scalar` than `!is_string` like a naturalNumber(). 

And those are changed to is_string(). But I noticed that I overlooked lengthBetween(), and sent another mail:
> In `lengthBetween`, it also might be better `is_scalar` than `is_string`.
For example, "Please enter valid a number length between 3 and 5 digits".
Sorry, I overlooked the change in your previous patch. The other changes seem good to me.

But it seemed to be not applied to the final version of the patch. So the is_string() check in lengthBetween()  was not intended. 

So please do not change the current behavior of minLength() and maxLength(). I would rather want to change is_string() to is_scalar() in the lengthBetween() even now.

ADmad: In over 2 years since that patch no one has had a problem with that method only accepting strings :)

Anyway, for all length changes methods I have changed to use `is_scalar()`.

chinpei215: I think application/json input is not used much. anyway, thank you!

dereuromark: It can also be user added data via setters on the entity which then get validated.
However you generate the entities, they must not fail because of this.

ADmad: It is IMO :)

dereuromark: Dont you think this is a dangerous behavior change?

chinpei215: > Is this good to merge now?

No, I found a mistake.

src/Validation/Validation.php
```diff
@@ -164,11 +164,11 @@ public static function lengthBetween($check, int $min, int $max): bool
      */
     public static function cc($check, $type = 'fast', $deep = false, $regex = null): bool
     {
-        if (!is_scalar($check)) {
+        if (!is_numeric($check)) {
```
chinpei215: This should not be changed. `is_numeric('0000-0000-0000-0000')` doesn't return true.

dereuromark: I would be more careful to change basic and stable validation code this drastically.
Lots of false positive "invalidations" because of this suddenly for a lot of apps.

markstory: This could be `is_string() || is_int()`.

ADmad: > This should not be changed. is_numeric('0000-0000-0000-0000') doesn't return true.

This mistake wouldn't have happened if there were tests cases for this format :)

src/Validation/Validation.php
ADmad: > Dont you think this is a dangerous behavior change?

@dereuromark I don't. I already explained in earlier comment https://github.com/cakephp/cakephp/pull/12436/files#r208102520 why allowing integer or float input for `containsNonAlphaNumeric()` doesn't make sense.

ADmad:
Relaxed the check for `cc()` and added test case for input with dashes.

src/Validation/Validation.php
```diff
@@ -1308,12 +1337,16 @@ public static function imageHeight(array $file, string $operator, int $height):
...
      *
-     * @param string $value Geographic location as string
+     * @param mixed $value Geographic location as string
      * @param array $options Options for the validation logic.
      * @return bool
      */
-    public static function geoCoordinate(string $value, array $options = []): bool
+    public static function geoCoordinate($value, array $options = []): bool
     {
+        if (!is_scalar($value)) {
```
markstory: Do we have existing tests for these scenarios now that they don't create `TypeError` again?

ADmad: Test added.

src/Validation/Validation.php
dereuromark: I still think this should have been kept to allow integers to pass fine. 
Especially when those come from a different input source than just basic string POST data.

PR #12438, opened 2018-08-07
Option to show the relative paths in the .pot comments

Spriz:
This change makes is possible to put in a relative path in the `.pot` file generated by the [ExtractTask](https://github.com/cakephp/cakephp/blob/master/src/Shell/Task/ExtractTask.php) used by the [I18nShell](https://github.com/cakephp/cakephp/blob/master/src/Shell/I18nShell.php).

Without using this new argument: 
```
#: /var/lib/jenkins/jobs/apacta-translations-update/workspace/plugins/Companies/src/Template/Projects/add.ctp:425
```

With this new argumentthis would be
```
#: ./plugins/Companies/src/Template/Projects/add.ctp:425
```

Not using this argument creates a lot of diffs when this file is sometimes updated through CI server, sometimes by developers or maybe by a 3rd party tool that doesn't quite understand these paths anyway - at least the full path doesn't make a lot of sense to either us or our tools :-)

I don't see a lot of tests covering the comments in the file, and my regex is not strong - so maybe someone would like to help a bit? 👍

### DIFF
--- a/src/Shell/Task/ExtractTask.php
+++ b/src/Shell/Task/ExtractTask.php
@@ -50,6 +50,13 @@ class ExtractTask extends Shell
     protected $_merge = false;
 
     /**
+     * Use relative paths in the pot files rather than full path
+     *
+     * @var bool
+     */
+    protected $_relativePaths = false;
+
+    /**
      * Current file being processed
      *
      * @var string|null
@@ -233,6 +240,10 @@ class ExtractTask extends Shell
             $this->_merge = strtolower($response) === 'y';
         }
 
+        if (isset($this->params['relative-paths'])) {
+            $this->_relativePaths = !(strtolower($this->params['relative-paths']) === 'no');
+        }
+
         if (empty($this->_files)) {
             $this->_searchFiles();
         }
@@ -320,6 +331,9 @@ class ExtractTask extends Shell
             'help' => 'Comma separated list of paths.'
         ])->addOption('merge', [
             'help' => 'Merge all domain strings into the default.po file.',
+            'choices' => ['yes', 'no']
+        ])->addOption('relative-paths', [
+            'help' => 'Use relative paths in the .pot file',
             'choices' => ['yes', 'no']
         ])->addOption('output', [
             'help' => 'Full path to output directory.'
@@ -447,6 +461,9 @@ class ExtractTask extends Shell
                         'file' => $this->_file,
                         'line' => $line,
                     ];
+                    if ($this->_relativePaths) {
+                        $details['file'] = '.' . str_replace(ROOT, '', $details['file']);
+                    }
                     if (isset($plural)) {
                         $details['msgid_plural'] = $plural;
                     }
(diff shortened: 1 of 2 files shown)


src/Shell/Task/ExtractTask.php
```diff
@@ -447,6 +461,9 @@ protected function _parse($functionName, $map)
                         'file' => $this->_file,
                         'line' => $line,
                     ];
+                    if ($this->_relativePaths) {
+                        $details['file'] = '.' . str_replace(ROOT, '', $details['file']);
```
Spriz: If anyone has a better idea for the string replacing here - please just change it! 

Also using the `.` in front of the path is more of a personal experience. I couldn't see Cakephp using either `./` or `/` so I just picked my favorite.

Just tell me if I should change it.

Spriz:
Just squashed 3 commits to a single.

src/Shell/Task/ExtractTask.php
ADmad: I personally would just use `foo/bar/baz.ctp` without any leading dot or slash.

ADmad (commented):
LGTM, needs tests.

src/Shell/Task/ExtractTask.php
dereuromark: I usually replace `ROOT . DS` completely :)
But either way is already relative and works fine.

src/Shell/Task/ExtractTask.php
```diff
@@ -233,6 +240,10 @@ public function main()
             $this->_merge = strtolower($response) === 'y';
         }
 
+        if (isset($this->params['relative-paths'])) {
```
markstory: You could use `if (!$this->param('relative-paths')) {` here.

src/Shell/Task/ExtractTask.php
```diff
@@ -321,6 +332,9 @@ public function getOptionParser()
         ])->addOption('merge', [
             'help' => 'Merge all domain strings into the default.po file.',
             'choices' => ['yes', 'no']
+        ])->addOption('relative-paths', [
+            'help' => 'Use relative paths in the .pot file',
+            'choices' => ['yes', 'no']
```
markstory: This could be a `boolean` argument. That will default it to `false` and when `--relative-paths` is set on the CLI it will be `true`.

markstory:
I can help with tests.

garas:
This PR duplicates what `--path` param does.

By default ExtractTask strips `/fullpath/src/` (`APP` const)

If you add `--extract-core`, it also strips `/fullpath/vendor/cakephp/cakephp/src/` (`CORE` const)

You can add as many paths as you wish with `--paths=/fullpath/src/,/fullpath/`.

Or use `--plugin=Companies`

ADmad:
@garas This new option is lot more intuitive :)

garas:
Then this parameter should only control whether this `str_replace` is applied.

https://github.com/cakephp/cakephp/blob/abbb678b23de16d0e4d15367804d7ae741510842/src/Shell/Task/ExtractTask.php#L490-L492

Also maybe add `ROOT` to default list of paths, but that would slowdown if app has many composer dependencies.

PR #12441, opened 2018-08-07
3.x Fix ConsoleIo:: and Shell:: constants => self:: constants

### LINKED ISSUE #12439: 3.x Reference to Shell:: constants in ConsoleIo but no use statement
Am I blind / Is it just me?

https://github.com/cakephp/cakephp/blob/abbb678b23de16d0e4d15367804d7ae741510842/src/Console/ConsoleIo.php#L176

https://github.com/cakephp/cakephp/blob/abbb678b23de16d0e4d15367804d7ae741510842/src/Console/ConsoleIo.php#L196

inoas:
Attempts to fix #12439 

Note: The book links do not work... nor did they before...

### DIFF
--- a/src/Console/ConsoleIo.php
+++ b/src/Console/ConsoleIo.php
@@ -84,7 +84,7 @@ class ConsoleIo
      *
      * @var int
      */
-    protected $_level = ConsoleIo::NORMAL;
+    protected $_level = self::NORMAL;
 
     /**
      * The number of bytes last written to the output stream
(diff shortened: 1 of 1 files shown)


src/Console/ConsoleIo.php
```diff
@@ -163,7 +163,7 @@ public function quiet($message, $newlines = 1)
      *
      * ### Output levels
      *
-     * There are 3 built-in output level. self::QUIET, self::NORMAL, self::VERBOSE.
+     * There are 3 built-in output level. ConsoleIo::QUIET, ConsoleIo::NORMAL, ConsoleIo::VERBOSE.
      * The verbose and quiet output levels, map to the `verbose` and `quiet` output switches
      * present in most shells. Using self::QUIET for a message means it will always display.
```
dereuromark: Same here etc :)

PR #12442, opened 2018-08-08
Update cacert.pem to Jun 20 version

Iandenh:
Latest CA bundle from: https://curl.haxx.se/ca/cacert-2018-06-20.pem


markstory:
I've verified the sha256 sum of the new file with the sha256 sum posted on https://curl.haxx.se/docs/caextract.html

PR #12447, opened 2018-08-09
Add assertRedirectNotContains

gemal:


gemal:
@ADmad can you review and possibly merge?

jeremyharris:
This is a new feature so shouldn't it be added to `3.next`? Constraints have changed there; this way of asserting isn't used anymore and this new feature would need to be re-done for `3.next`.

src/TestSuite/IntegrationTestCase.php
```diff
@@ -729,6 +729,25 @@ public function assertRedirectContains($url, $message = '')
         $this->assertContains($url, $result['Location'], $message);
     }
 
+    /**
+     * Asserts that the Location header does not contain a substring
+     *
+     * @param string $url The URL you expected the client to go to.
+     * @param string $message The failure message that will be appended to the generated message.
+     * @return void
```
markstory: Should we have an `@since` tag here?

markstory:
Moved to 3.next as that is where new features go.

jeremyharris:
@gemal let me know if you have trouble migrating over to the new constraints. You should be able to copy the current header contains assertion and use a new `HeaderNotContains` constraint, which you can copy from `HeaderContains` and reverse the logic.

saeideng:
you should rebase on 3.next and push again 
travis now used 3.6 core code as source , this will fix in next push

markstory:
Moved to #12464 as I didn't want to push updates to `gemal:master`.

PR #12448, opened 2018-08-09
Rename Validation::cc() to Validation::creditCard()

garas:
I think `cc` too cryptic abbreviation for credit card.

There is already `Validator::creditCard()`, so it should be also `Validation::creditCard()`

### DIFF
--- a/src/Validation/Validation.php
+++ b/src/Validation/Validation.php
@@ -185,6 +185,29 @@ class Validation
     }
 
     /**
+     * Backwards compatibility wrapper for Validation::creditCard().
+     *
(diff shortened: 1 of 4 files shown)


markstory:
@ADmad renaming does make method names consistent between `Validator` and `Validation` though.

ADmad:
Ah, well then that's a good enough reason for the rename.

saeideng:
so ,why not rename `Validator::creditCard()` to `Validator::cc()`  ? :)

ADmad (approved):

PR #12450, opened 2018-08-10
Made FormHelper autoSetCustomValidity option true by default.

ADmad:
Refs #11837

### DIFF
--- a/src/View/Helper/FormHelper.php
+++ b/src/View/Helper/FormHelper.php
@@ -165,7 +165,7 @@ class FormHelper extends Helper
             'confirmJs' => '{{confirm}}',
         ],
         // set HTML5 validation message to custom required/empty messages
-        'autoSetCustomValidity' => false,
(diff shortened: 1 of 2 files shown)


markstory:
I feel like this is something that should be featured prominently in the formhelper docs. Any ideas on where this behavior should be mentioned?

ADmad:
There a section [Using Custom Validators](https://book.cakephp.org/3.0/en/views/helpers/form.html#using-custom-validators). Perhaps that could be renamed and the section updated to include info about this option.

I also think it would be nice to be able to have `autoSetCustomValidity` as an option for `FormHelper::create()` too.

markstory:
> I also think it would be nice to be able to have autoSetCustomValidity as an option for FormHelper::create() too.

Should we be doing that change now?

ADmad:
> Should we be doing that change now?

I guess so, on the 4.x branch for docs.

Issue #12452, opened 2018-08-10
Database column names containing utf-8 chars (ä) are quoted incorrectly


benjaminpick:
This is a (multiple allowed):

* [x] bug
* [x] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.6.10
* Platform and Target: Apache, MySQL

### What you did

I am creating a connector to a live database which has "creative" column names such as `Reservierungsbestätigung Datum` (containing spaces and Umlauts and I can't change it). 

In my configuration, `quoteIdentifiers` is enabled. 

### What happened
The Database method `quoteIdentifier()` escapes
```
 Reservierungsbestätigung Datum
``` 
to 

```
`Reservierungsbest`ätigung Datum
```
which understably raises an SQL error as soon as the query executes. This is because the regexes use \w as character class.

### What you expected to happen
I hoped the regexes would be more tolerant, as AFAIK only `.` and ` ` (space) are symbols relevant to the query parsing.

### Suggestion
If we add 'u' as modifier to the concerned regexes, it would include these Umlaut chars as well, while still excluding any special chars.

This regex-change might break compatibility (although I can't come up with good examples where this behavior would be used on purpose). Shall I write a PR for this?

Similar: #12079 (spaces in column name)

benjaminpick:
(Fortunately, I do not need the columns containing the UTF-8 chars so if this can't be solved, I can specify all columns that I do need and the query still runs correctly.)

markstory:
Could you send a pull request with the changes that make identifier quoting work for your schema?

markstory:
Closing as a pull request is open now.

PR #12454, opened 2018-08-11
Quote column identifier for column names containing utf-8 chars (ä)

benjaminpick:
Ref #12452

This simple query works now with my schema:

```php
TableRegistry::getTableLocator()->get('Tablename')->find()->first();
```
I will test some more with conditions etc.

### DIFF
--- a/src/Database/SqlDialectTrait.php
+++ b/src/Database/SqlDialectTrait.php
@@ -38,19 +38,19 @@ trait SqlDialectTrait
         }
 
         // string
-        if (preg_match('/^[\w-]+$/', $identifier)) {
+        if (preg_match('/^[\w-]+$/u', $identifier)) {
             return $this->_startQuote . $identifier . $this->_endQuote;
         }
 
         // string.string
-        if (preg_match('/^[\w-]+\.[^ \*]*$/', $identifier)) {
+        if (preg_match('/^[\w-]+\.[^ \*]*$/u', $identifier)) {
             $items = explode('.', $identifier);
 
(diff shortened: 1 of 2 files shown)


benjaminpick:
How the u modifier works:

https://maxivak.com/working-with-regular-expressions-preg_-and-utf-8-strings-in-php/

If PCRE is compiled with Unicode property support, and the PCRE_UCP option is set (by the u modifier):
\w will match any character that \p{L} (Unicode letters) or \p{N} (Unicode numbers) matches, plus underscore

markstory:
Thank you 👍

benjaminpick:
Can someone check it for performance implications, please?

-- 
Theologie zum Anfassen: https://benjaminpick.com

benjaminpick:
I agree but using the u modifier might decrease performance for all who
use quoteIdentifier-escaping. In the same way as the current
implementation first checks the standard cases, then the weird cases, we
could also add the u modifier only for "extra" cases.

markstory:
@benjaminpick We're OK with that. Identifier quoting already adds significant overhead to the ORM that a bit more is just part of that cost now.

benjaminpick:
Ok, I think I was just surprised that it got merged so quickly!
Thanks ...

dereuromark:
I don't know.. Quoting I find totally acceptable, but using non ascii chars for table and field names is a total nogo for me.

benjaminpick:
I also find it very "creative" (in a bad sense) ... It is an ancient
Access database migrated to SQL Server and so on. The primary keys are
called differently in each table etc. ...

PR #12455, opened 2018-08-11
Remove return type hint for before* behavior methods.

ADmad:
These methods can return values to be used as event result and short circuit
the operation. Adding return type would prevent someone from extending and overriding
these methods to return a non void value.

Refs #11935

### DIFF
--- a/src/Controller/Component/RequestHandlerComponent.php
+++ b/src/Controller/Component/RequestHandlerComponent.php
@@ -243,7 +243,7 @@ class RequestHandlerComponent extends Component
      * @param \Cake\Event\EventInterface $event The Controller.beforeRender event.
      * @return void
      */
-    public function beforeRender(EventInterface $event): void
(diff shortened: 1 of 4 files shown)


dereuromark:
Why? event object can handle that.
Also then doc blocks would be lying.

ADmad:
I already answered the "why" in my initial comment. 

If the return type is declared as `void` and if someone for e.g. wants to extend `TreeBehavior and return `false` from it's `beforeDelete()` then PHP will throw an error.

ADmad:
One can use `Event::setResult()` and `Event::stopPropation()` but the event system allows simply returning a value too from callback method from convenience. 

If we want to remove this convenience of returning value then we should deprecate it in 3.x. Though I am not sure it's a good idea to add this BC break and additional step for migration just so that we can add return type hint to the callback methods.

dereuromark:
Ok. Sounds like a cleanup for 5.x then probably.

PR #12461, opened 2018-08-12
Upgrading phpunit to 7.x

lorenzo:
Making the FixtureListener not inherit a class that is not there anymore. Adding a trait as a default implementation in case we need to add more listeners.

### DIFF
--- a/src/Filesystem/File.php
+++ b/src/Filesystem/File.php
@@ -455,7 +455,7 @@ class File
     public function perms()
     {
         if ($this->exists()) {
-            return substr(sprintf('%o', fileperms($this->path)), -4);
+            return substr(sprintf('%o', fileperms($this->path)), -3);
         }
 
         return false;
--- a/src/Filesystem/File.php
+++ b/src/Filesystem/File.php
@@ -455,7 +455,7 @@ class File
     public function perms()
     {
         if ($this->exists()) {
-            return substr(sprintf('%o', fileperms($this->path)), -3);
+            return decoct(fileperms($this->path) & 0777);
         }
 
         return false;
--- a/src/TestSuite/EmailTrait.php
+++ b/src/TestSuite/EmailTrait.php
@@ -55,10 +55,10 @@ trait EmailTrait
      * Asserts an expected number of emails were sent
      *
      * @param int $count Email count
-     * @param string|null $message Message
+     * @param string $message Message
      * @return void
      */
-    public function assertMailCount(int $count, ?string $message = null): void
+    public function assertMailCount(int $count, string $message = ''): void
     {
         $this->assertThat($count, new MailCount(), $message);
     }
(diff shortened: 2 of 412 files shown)


ADmad:
Builds failing due to method signature mismatch.

lorenzo:
Well, I guess I'll have to fix my env and run all the test suite. I thought it would be simpler :P

src/Filesystem/File.php
```diff
@@ -455,7 +455,7 @@ public function exists(): bool
     public function perms()
     {
         if ($this->exists()) {
-            return substr(sprintf('%o', fileperms($this->path)), -4);
+            return substr(sprintf('%o', fileperms($this->path)), -3);
```
ADmad: Earlier the method return string of form `0xxx` but that was causing tests to fail as expected value generated used `deoct()` was of form `xxx` without th preceding `0`.

garas: What about `sprintf('%03o', fileperms($this->path) & 0777)`?

markstory:
Looks good I'll merge this once I get the 4.x docs updated.

dereuromark:
By the way:
Don't we have to call all assert statically? Is that strange way still a thing in 7+?
See e.g. https://github.com/spryker/spryk/blob/master/tests/SprykTest/Spryk/Model/Spryk/ConfigurationLoader/SprykConfigurationLoaderTest.php#L46

At least with a higher phpstan/analyzing level it would complain about non static usage afaik.

dereuromark:
Hah.. https://github.com/sebastianbergmann/phpunit/issues/2532 and following the links there you find https://github.com/phpstan/phpstan-strict-rules/issues/36 :) What an annoyance that they don't fix it up to use only of the two.

markstory:
Many of our assertions are not static, switching to use static calls for phpunit assertions will make tests visually inconsistent and add mental overhead around which assertions can be used statically or dynamically.

src/Filesystem/File.php
```diff
@@ -455,7 +455,7 @@ public function exists(): bool
     public function perms()
     {
         if ($this->exists()) {
-            return substr(sprintf('%o', fileperms($this->path)), -4);
+            return decoct(fileperms($this->path) & 0777);
```
ravage84: Nice!

src/TestSuite/EmailTrait.php
```diff
@@ -55,21 +55,21 @@
      * Asserts an expected number of emails were sent
      *
      * @param int $count Email count
-     * @param string|null $message Message
+     * @param string $message Message
      * @return void
      */
-    public function assertMailCount(int $count, ?string $message = null): void
+    public function assertMailCount(int $count, string $message = ''): void
```
ravage84: Yeah, much better like this :+1:

ravage84 (approved):
:+1:

PR #12462, opened 2018-08-12
Deprecate ViewVarsTrait::viewOptions() and related property.

ADmad:
Since we have deprecated all view related public properties from classes using
this trait, this method is now redundant.

### DIFF
--- a/src/View/ViewVarsTrait.php
+++ b/src/View/ViewVarsTrait.php
@@ -80,7 +80,7 @@ trait ViewVarsTrait
             $builder->setClassName($viewClass);
         }
 
-        $validViewOptions = $this->viewOptions();
+        $validViewOptions = isset($this->_validViewOptions) ? $this->_validViewOptions : [];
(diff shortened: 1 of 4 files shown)


src/View/ViewVarsTrait.php
```diff
@@ -157,6 +157,7 @@ public function set($name, $value = null)
      * @param bool $merge Whether to merge with or override existing valid View options.
      *   Defaults to `true`.
      * @return array The updated view options as an array.
+     * @deprecated 3.7.0 Use ViewBuilder::setOptions() or any one of it's setter methods instead.
      */
     public function viewOptions($options = null, $merge = true)
```
markstory: Should this method emit a deprecation warning?

ADmad: Yup, will do

ADmad: Done

PR #12464, opened 2018-08-13
Add assertRedirectNotContains & assertHeaderNotContains

markstory:
Adapt the code from #12447 to use constraints. I also added `assertHeaderNotContains()` as it was easy and could be useful.

### DIFF
--- a/src/TestSuite/Constraint/Response/HeaderNotContains.php
+++ b/src/TestSuite/Constraint/Response/HeaderNotContains.php
@@ -1,0 +1,44 @@
+<?php
+/**
+ * CakePHP(tm) : Rapid Development Framework (http://cakephp.org)
+ * Copyright (c) Cake Software Foundation, Inc. (http://cakefoundation.org)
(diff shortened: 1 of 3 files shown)


src/TestSuite/Constraint/Response/HeaderNotContains.php
```diff
@@ -0,0 +1,44 @@
...
+ * @license       http://www.opensource.org/licenses/mit-license.php MIT License
+ */
+namespace Cake\TestSuite\Constraint\Response;
+
+/**
+ * Constraint for ensuring a header does not contain a value.
+ *
+ * @internal
+ */
+class HeaderNotContains extends HeaderEquals
```
jeremyharris: Should this extend HeaderContains instead? HeaderEquals is a full comparison.

markstory: Yes it should. Thank you.

PR #12465, opened 2018-08-13
3.7 - ViewBuilder view vars

ADmad:
With this change `ViewVarsTrait::$viewVars` can be deprecated.

A bit unsure how to deal with `ViewVarsTrait::set()`. Most people will prefer continuing to use `$this->set()` instead of `$this->viewBuilder()->setVar()` so can't deprecate that.

Changing the guts of `ViewVarsTrait::set()` to proxy the calls to `ViewBuilder::setVar()/setVars()` would create a BC problem since then one won't be able to us `ViewVarsTrait::$viewVars` to read the vars.

### DIFF
--- a/src/View/ViewBuilder.php
+++ b/src/View/ViewBuilder.php
@@ -114,6 +114,66 @@ class ViewBuilder implements JsonSerializable, Serializable
     protected $_helpers = [];
 
     /**
+     * View vars
+     *
+     * @var array
+     */
+    protected $_vars = [];
+
+    /**
+     * Saves a variable for use inside a template.
+     *
+     * @param string $name A string or an array of data.
+     * @param mixed $value Value.
+     * @return $this
+     */
+    public function setVar($name, $value = null)
+    {
+        $this->_vars[$name] = $value;
+
+        return $this;
+    }
+
+    /**
+     * Saves view vars for use inside templates.
+     *
+     * @param array $data Array of data.
+     * @param bool $merge Whether to merge with existing vars, default true.
+     * @return $this
+     */
+    public function setVars($data, $merge = true)
+    {
+        if ($merge) {
+            $this->_vars = $data + $this->_vars;
+        } else {
+            $this->_vars = $data;
+        }
+
+        return $this;
+    }
+
+    /**
+     * Get view var
+     *
+     * @param string $name Var name
+     * @return mixed The var value or null if unset.
+     */
+    public function getVar($name)
+    {
+        return isset($this->_vars[$name]) ? $this->_vars[$name] : null;
+    }
+
+    /**
+     * Get all view vars.
+     *
+     * @return array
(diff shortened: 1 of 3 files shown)


markstory:
> Most people will prefer continuing to use $this->set() instead of $this->viewBuilder()->setVar() so can't deprecate that.

I know that I will personally continue using `$this->set()`. I think this is one of the places where we should prioritize ergonomics over SOLID.

src/View/ViewBuilder.php
```diff
@@ -113,6 +113,77 @@ class ViewBuilder implements JsonSerializable, Serializable
...
+    protected $_vars = [];
+
+    /**
+     * Saves a variable for use inside a template.
+     *
+     * @param string $name A string or an array of data.
+     * @param mixed $value Value.
+     * @return $this
+     */
+    public function setVar($name, $value = null)
```
markstory: Could the trait be updated to call this method if the implementing class has a `viewBuilder()` method? Would that make removing the public property easier later?

ADmad: Yes that's what I was thinking of doing, but then the viewVars property of the trait won't be populated. I personally have used the property to check if a bar is set so it would cause BC issue.

ADmad: Yup, so to avoid any BC issues let's keep ViewVarsTrait unchanged and simply mark the public $viewVars property as deprecated.

In 4.x we'll remove the property and update set() method to proxy the calls to view builder methods instead.

ADmad: Done

src/View/ViewVarsTrait.php
```diff
@@ -38,6 +38,7 @@
      * Variables for the view
      *
      * @var array
+     * @deprecated 3.7.0 Use `$this->viewBuilder()->setVar()/getVar()` instead.
```
markstory: Couldn't folks also use `$this->set()` for a simpler interface?

ADmad: Yes they can, but currently if you use `$this->set()` you will have to still use `$this->viewVars` if you want to read the vars.

markstory: My concern would be that people migrate to using the longer form and think that 'cake is too verbose for simple things'. This message could be:

> Use `$this->set()` to set variables or `$this->viewBuilder()->getVar()` to read them instead.`

ADmad: The thing is if you use `$this->set()` you can't `$this->viewBuilder()->getVar()` to read the var as `$this->set()` still sets them to `$this->viewVars` not to the viewbuilder.

markstory: Hrm.

ADmad: Done

PR #12466, opened 2018-08-13
 Free some memory in unit tests

bancer:


bancer:
SQLite test failed to load PHPUnit. Please restart the build.

lib/Cake/TestSuite/CakeTestCase.php
```diff
@@ -164,6 +166,8 @@ public function tearDown() {
 		if (isset($_GET['debug']) && $_GET['debug']) {
 			ob_flush();
 		}
+		unset($this->_configure);
+		unset($this->_pathRestore);
```
markstory: You can do all of this in one unset call like `unset($this->_configure, $this->_pathRestore)`.

lib/Cake/TestSuite/ControllerTestCase.php
```diff
@@ -418,4 +418,18 @@ public function generate($controller, $mocks = array()) {
...
+ * @return void
+ */
+	public function tearDown() {
+		parent::tearDown();
+		unset($this->contents);
+		unset($this->controller);
+		unset($this->headers);
+		unset($this->result);
+		unset($this->view);
+		unset($this->vars);
```
markstory: You could combine all of this into one unset() as well.

bancer: Of course if that is a preferable syntax. One line or multiple?

PR #12468, opened 2018-08-13
3.next email transport

ADmad:
Added `TransportFactory` and `TransportRegistry` to off load transport instance creation responsibilities from the `Email` class and deprecated related methods.


src/Mailer/TransportFactory.php
```diff
@@ -0,0 +1,114 @@
...
+            );
+        }
+
+        if (is_array(static::$_config[$name]) && empty(static::$_config[$name]['className'])) {
+            throw new InvalidArgumentException(
+                sprintf('Transport config "%s" is invalid, the required `className` option is missing', $name)
+            );
+        }
+
+        static::getRegistry()->load($name, static::$_config[$name]);
```
markstory: This method looks like what is sometimes found in the `_create` method of a registry implementation. What does having a factory + registry class provide us?

src/Mailer/TransportRegistry.php
```diff
@@ -0,0 +1,113 @@
...
+        if (is_object($class)) {
+            return $class;
+        }
+
+        $className = App::className($class, 'Mailer/Transport', 'Transport');
+
+        if (!$className) {
+            $className = App::className($class, 'Network/Email', 'Transport');
+            if ($className) {
+                trigger_error(
```
markstory: Should this use `deprecationWarning()` now? I think this codes previous home predates that function's existence.

markstory (commented):
I like the separation of concern out of `Email`.

src/Mailer/TransportRegistry.php
ADmad: Yes it should, I'll update.

src/Mailer/TransportFactory.php
markstory: What if we only had the registry? A registry in many ways is a memoized factory.

ADmad: If we only had the registry then we would have to make the registry use the `StaticConfigTrait` to provide the `setConfg()`, `getConfig()` etc. static methods.

Currently I have made the `TranportFactory`-`TranportRegistry` pair similar to `Cache`- `CacheRegistry` and `Log`-`LogEngineRegistry` pairs we have (without the extra methods `Cache` and `Log` have).

markstory: Ok. Lets roll with that until the DI apocalypse arrives :smile:

markstory: Lets keep this.

ADmad:
This is good to go now.

PR #12471, opened 2018-08-14
Small php issue

sohelrana820:
Fixed some small PHP issues.

### DIFF
--- a/src/Console/ConsoleInput.php
+++ b/src/Console/ConsoleInput.php
@@ -77,7 +77,8 @@ class ConsoleInput
     public function dataAvailable($timeout = 0)
     {
         $readFds = [$this->_input];
-        $readyFds = stream_select($readFds, null, null, $timeout);
+        $null = null;
+        $readyFds = stream_select($readFds, $null, $null, $timeout);
 
         return ($readyFds > 0);
     }
--- a/src/Console/ConsoleInput.php
+++ b/src/Console/ConsoleInput.php
@@ -77,8 +77,7 @@ class ConsoleInput
     public function dataAvailable($timeout = 0)
     {
         $readFds = [$this->_input];
-        $null = null;
-        $readyFds = stream_select($readFds, $null, $null, $timeout);
+        $readyFds = stream_select($readFds, null, null, $timeout);
 
         return ($readyFds > 0);
     }
--- a/src/Http/Client.php
+++ b/src/Http/Client.php
@@ -179,7 +179,7 @@ class Client
     /**
      * Get the cookies stored in the Client.
      *
-     * @return \Cake\Http\Client\CookieCollection
+     * @return CookieCollection
      */
     public function cookies()
     {
--- a/src/Http/Client.php
+++ b/src/Http/Client.php
@@ -179,7 +179,7 @@ class Client
     /**
      * Get the cookies stored in the Client.
      *
-     * @return CookieCollection
+     * @return \Cake\Http\Client\CookieCollection
      */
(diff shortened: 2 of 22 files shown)


tests/TestCase/Filesystem/FileTest.php
```diff
@@ -533,7 +533,7 @@ public function testWrite()
     /**
      * testAppend method
      *
-     * @return void
+     * @return boolean|void
```
dereuromark: The test should rather be fixed as this should be void.

src/Http/Client.php
```diff
@@ -179,7 +179,7 @@ public function __construct($config = [])
     /**
      * Get the cookies stored in the Client.
      *
-     * @return \Cake\Http\Client\CookieCollection
+     * @return CookieCollection
```
dereuromark: why?

sohelrana820: Cause $this->_cookies is a object of Cake\Http\Cookie\CookieCollection. But it was return the object of \Cake\Http\Client\CookieCollection

dereuromark: But please fix it to the FQCN as it was before, just the right one.

ADmad: Then you need to use the correct FQCN not just `CookieCollection`.

sohelrana820: OK, then I am reverting this changes.

dereuromark: Just fix it to the expected `\Cake\Http\Cookie\CookieCollection` then.

sohelrana820: Just want to ensure... After fixing this I need to send a new pull request, right?

dereuromark: No, you can add a 2nd commit on top of this one.

src/Console/ConsoleInput.php
```diff
@@ -77,7 +77,8 @@ public function read()
     public function dataAvailable($timeout = 0)
     {
         $readFds = [$this->_input];
-        $readyFds = stream_select($readFds, null, null, $timeout);
+        $null = null;
```
dereuromark: What is the value of this? The repeating null is acceptable in the case it was here.

tests/test_app/TestApp/Controller/CakesController.php
```diff
@@ -39,7 +39,7 @@ public function noRender()
     /**
      * invalid method
      *
-     * @return \Cake\Http\Response
+     * @return string
```
dereuromark: The point of this kind of is to show that this is an invalid and undocumented return type.
Not sure if we should switch the documentation here.

src/Console/ConsoleInput.php
ADmad: Please revert this unnecessary change.

sohelrana820: stream_select() expect reference value. And null can't be passed by a reference.

ADmad: PHP doesn't complain when passing `null` for those argument so it's a valid value.

sohelrana820: @ADmad, so I need to revert this changes?

tests/TestCase/Filesystem/FileTest.php
```diff
@@ -503,7 +503,7 @@ public function testLastChange()
     /**
      * testWrite method
      *
-     * @return void
+     * @return bool|void
```
dereuromark: Same issue here. Must stay void and instead the test should be skipped properly.

PR #12472, opened 2018-08-15
Fix potential timing side channel in digest auth.

markstory:
This change reduces the potential ability to use digest authentication as a side channel for user enumeration. Previously passwords would be hashed for digest users that did not exist, but not hashed for users that *did*. These changes ensure that if password===null no hashing is done. This also means we can remove the string cast.

Thanks to Edgaras Janušauskas for raising this issue through our responsible disclosure mailing list.


garas:
Thanks, PR fixes the issue.

But I think `$password !== null` would be better than `strlen($password) > 0`. While core Form/Basic and few plugins, which I've reviewed, all checks if `$password != ''`, however if some plugin wouldn't check for empty password, it could be another timing attack on `_findUser('existing', '')` vs `_findUser('invalid', '')`

PR #12474, opened 2018-08-15
Update CookieCryptTrait.php

hackzilla:
When trying to disable cookie encryption, I accidentally set it to null.
The error message was unclear that it could be turned off.

Specify in the error message, that false is an option.


hackzilla:
Changed from master to 3.next, and updated the tests.

ADmad:
Updating a error message should just target `master` not `3.next`. `master` is for next bugfix release while `3.next` is for next minor.

PR #12479, opened 2018-08-15
Document "escape" option.

ADmad:


ADmad:
Merging to avoid running test suite unnecessarily and contributing to global warming.

PR #12480, opened 2018-08-15
Add an option on extractTask to avoid the display of marker errors

AlPri78:
This is a (multiple allowed):
- [ ]  bug
- [X]  enhancement
- [ ]     feature-discussion (RFC)

I implemented a new option on extractTask to avoid the display of marker errors.

This display is already off for Cake Core (Example: FormHelper:913).
Because I also have some generic code, the value for the function __() is not always a string but a variable.

It's for me very useful to have this new option. It's why I purpose to integrate it.

I'm not able to add a specific test because we are in Cake, and very logically that kind of marker error are skipped thanks to line 468 with `} elseif (strpos($this->_file, CAKE_CORE_INCLUDE_PATH) === false) {`

What do you think about this improvement ?


src/Shell/Task/ExtractTask.php
```diff
@@ -674,6 +686,8 @@ protected function _formatString($string)
      */
     protected function _markerError($file, $line, $marker, $count)
     {
+        if (!$this->markerError) return;
```
dereuromark: Should be normalized with {}

AlPri78:
Thanks @dereuromark . I did it.

Does it exist a reference for code normalization ?

dereuromark:
The basic code style guide and CS sniffer repo.

src/Shell/Task/ExtractTask.php
```diff
@@ -352,6 +360,10 @@ public function getOptionParser()
             'boolean' => true,
             'default' => false,
             'help' => 'Do not write file locations for each extracted message.',
+        ])->addOption('no-marker-error', [
```
markstory: It would be good to have a test for this option. The test could use `assertNotContains()` or `assertNotRegExp()` to ensure the error markers are not present.

AlPri78:
I agree @markstory but as I explained in the opening of this PR, the line 468 in ExtractTask.php  with `}
 elseif (strpos($this->_file, CAKE_CORE_INCLUDE_PATH) === false) {` does not allow to enter in that part of code because this test is false for Cake and we are in Cake. Then, the line 469 ( `$this->_markerError($this->_file, $line, $functionName, $count);` ) is never covered for all existing tests.
I already added a line test in extract.ctp and normally the first test (testExecute) would not be ok but there is this line 468 !
Do you have an idea for a loophole ?

markstory:
@AlPri78 Hmm that is tricky to get around. I wonder if we still need to exclude core files from marker errors with the new option?

AlPri78:
Ok. After a long thinking, I purpose this new committed solution.

**Explanation** : The fact to display marker errors is not so useful, but it could be interesting to know them.
It's why I purpose to reverse the option with, by default, no display of marker errors, then with the new option 'marker-error' it's possible to display.
To alert the user, the number of marker errors is displayed if at least one is present. This count does not include marker errors for the core.
With that solution, it's possible to add a test for this new option and it's done in this commit. As before, it's not possible to test the counter because we are in the core.

What do you think about ?

src/Shell/Task/ExtractTask.php
```diff
@@ -308,6 +314,11 @@ protected function _extract()
         $this->_paths = $this->_files = $this->_storage = [];
         $this->_translations = $this->_tokens = [];
         $this->out();
+        if ($this->_countMarkerError) {
+            $this->err("{$this->_countMarkerError} marker error(s) detected.");
+            $this->err(" => Activate marker-error option to list errors.");
```
markstory: Should `marker-error` have formatting to make it visually distinct? Or perhaps say `Use the --marker-error option to display errors`?

AlPri78: Nice. I committed with phpcs errors fixed.

PR #12486, opened 2018-08-20
Add allowEmptyByFlags() and type-specific allowEmpty*() methods

chinpei215:


src/Validation/Validator.php
```diff
@@ -2095,6 +2407,7 @@ public function __debugInfo()
         return [
             '_presenceMessages' => $this->_presenceMessages,
             '_allowEmptyMessages' => $this->_allowEmptyMessages,
+            '_allowEmptyFlags' => $this->_allowEmptyFlags,
```
markstory: Should this be simplified for human use? Instead of `3` we could output `EMPTY_STRING | EMPTY_ARRAY`.

src/Validation/Validator.php
```diff
@@ -635,8 +706,92 @@ public function requirePresence($field, $mode = true, $message = null)
...
+     *
+     * @param string|array $field the name of the field or a list of fields
+     * @param int|null $flags A bitmask of EMPTY_* flags which specify what is empty
+     * @param bool|string|callable $when Indicates when the field is allowed to be empty
+     * Valid values are true (always), 'create', 'update'. If a callable is passed then
+     * the field will allowed to be empty only when the callback returns true.
+     * @param string|null $message The message to show if the field is not
+     * @return $this
+     */
+    public function allowEmptyByFlags($field, $flags, $when = true, $message = null)
```
markstory: Does this need to be public? It is a complicated to operate method that has simpler to use public methods defined below.

src/Validation/Validator.php
```diff
@@ -635,8 +706,92 @@ public function requirePresence($field, $mode = true, $message = null)
...
+     * $validator->allowEmptyByFlags(
+     *      [
+     *          'email' => [
+     *              'when' => true
+     *          ],
+     *          'content' => [
+     *              'message' => 'Content cannot be empty'
+     *          ],
+     *          'subject'
+     *      ],
```
markstory: Do we need to support this mode? It seems awkward to use.

tests/TestCase/Validation/ValidatorTest.php
```diff
@@ -706,6 +706,341 @@ public function testAllowEmptyAsArrayFailure()
...
+            'items' => [],
+        ];
+        $result = $validator->errors($data);
+        $this->assertEmpty($result);
+
+        $data = [
+            'items' => [1, 2, 3, 4, 5],
+        ];
+        $result = $validator->errors($data);
+        $this->assertNotEmpty($result);
```
markstory: Should this assert the message?

src/Validation/Validator.php
chinpei215: We could deprecate it in 3.7.

src/Validation/Validator.php
chinpei215: As I said in the dev channel, people might need to customize what is empty. For example, if it is a RESTful API, people may prefer EMPTY_ARRAY only.

src/Validation/Validator.php
chinpei215: I am not sure whether or not we should do it. If you think we should do it, I am willing to implement it.

tests/TestCase/Validation/ValidatorTest.php
chinpei215: What message should we report to users?

markstory: Would it be useful to check that the `hasAtMost` rule failed?

src/Validation/Validator.php
markstory: Could we contain this mode to only exist in the older `allowEmpty()`? That would allow us to never add it avoiding a method that is deprecated/removed right away.

src/Validation/Validator.php
markstory: Ok.

src/Validation/Validator.php
markstory: Maybe it isn't that important. We can move forward with this and improve it later if necessary.

tests/TestCase/Validation/ValidatorTest.php
chinpei215: Ah sorry, I misunderstood. I will add the assertion to make sure hasAtMost rule failed.

src/Validation/Validator.php
chinpei215: OK.

src/Validation/Validator.php
chinpei215: We could. Should I fix doc comments of type-specific allowEmpty*()s? Or should I throw an exception in those methods?

markstory: I would fix the doc comments for now. We can add stricter types in 4.0

chinpei215: Ok.

markstory (commented):
I was thinking about how to document this, and I think this change will be simpler to communicate if it is merged into `3.next` instead of `master`. Doing so will allow us to include the 'prior to 3.7' type comments and fully update the examples to use the new methods.

src/Validation/Validator.php
```diff
@@ -653,11 +809,126 @@ public function allowEmpty($field, $when = true, $message = null)
             if ($settings[$fieldName]['message']) {
                 $this->_allowEmptyMessages[$fieldName] = $settings[$fieldName]['message'];
             }
+            if ($settings[$fieldName]['flags'] !== null) {
+                $this->_allowEmptyFlags[$fieldName] = $settings[$fieldName]['flags'];
```
markstory: If `notEmpty()` is called on a field after being marked as `allowEmptyArray()` (or one of the other types) should the flags information be cleared?

src/Validation/Validator.php
chinpei215: I am OK with reverting allowEmpty() to the original state. then we would not need to care about the backward compatibility.

src/Validation/Validator.php
chinpei215: I am not sure if we should take care such use cases. Even now if someone calls notEmpty() without the 'message' option after he called allowEmpty() with the 'message' option, it will end up with an unexpected result.

src/Validation/Validator.php
lorenzo: I agree with mark, this is better if this is just an internal method.

chinpei215:
As I said before, I think this is a vulnerability. If we add this methods into only 3.7, what can people do to prevent their 3.4/3.5/3.6 applications besides upgrading to 3.7?
If there is a workaround I am OK with that. Then I would rather want to close this PR though. I can't come up with any good workaround. That was why I opened this PR.

src/Validation/Validator.php
chinpei215: I have already explained why I made it as public. But I don't stick with my thought. Feel free to add your commits.

markstory: @chinpei215 I'll push some commits up tonight.

markstory:
@chinpei215 Ok. Adding the methods is less of a problem than the communication around them. Updating the documentation to use a method that was added in a patch release doesn't feel like a good idea to me, nor is updating bake to generate code with these new methods.

chinpei215:
I don't understand why we cannot include 'prior to 3.6.12' type comments instead.
As for bake, we could bump up the version to 1.9.0.

markstory:
@chinpei215 I guess we could. We just haven't done that historically.

src/Validation/Validator.php
markstory: I pushed up my change to remove the array field support from new methods.

ravage84:
@chinpei215 

> I don't understand why we cannot include 'prior to 3.6.12' type comments instead.
As for bake, we could bump up the version to 1.9.0.

SemVer says:

> Given a version number MAJOR.MINOR.PATCH, increment the:

> 1.   MAJOR version when you make incompatible API changes,
> 2.   MINOR version when you add functionality in a backwards-compatible manner, and
> 3.    PATCH version when you make backwards-compatible bug fixes.

So if we add non-BC-breaking functionality, we MUST increase the MINOR version, according to SemVer.

Now, if we need to introduce some tiny functionality to work around a vulnerability, I'd argue we could use a PATCH version for that. But adding five new validation methods is not tiny to me...

Now, I must admit, over the years I've come to the realisation that SemVer is not always a good fit for frameworks which are a composition of various small sets of individual or connected functionality. Nonetheless, if we communicate we do SemVer, we should adhere to it.

ravage84 (changes requested):
I'm good with the change itself, but I'd prefer to target 3.7, as @markstory recommended.

chinpei215:
This is a security fix for for 3.4, 3.5 and 3.6. Targeting 3.7 doesn't make any sense.

ravage84:
> This is a security fix for 3.4, 3.5 and 3.6.

Well, I understood that so far. Nonetheless you add functionality to workaround the security issue.

> Targeting 3.7 doesn't make any sense.

SemVer has, arguably, strict rules. Following them demands bumbing the MINOR version.
But I guess it's not very common to add functionality to work around a security bug.
So the people behind SemVer didn't fully think of this.

Anyway, it would be great if we _could_ find a way to fix or work around the bug without introducing new functionality, so we do not violate the rules of SemVer.
The idea behind SemVer is that one can know without checking each and every change log of a package what - in general - is going on with the code. If we start to divert from adhering to SemVer people will loose trust in the framework's adherence to SemVer.

chinpei215:
Would you mind if I close this pull request? We could recommend enabling SecurityComponent always instead, for 3.x applications.
I reported this vulnerability over a month ago. In the first place, I suggested fixing allowEmpty() itself, but it was rejected. Hence I suggested this way as a compromise, and it was approved. But I am not sure if people want to use new allowEmpty*() methods in 4.0 or later versions. 

The root cause of this issue is the design of 3.x validation. In 2.x, validation will run **after** *marshaling*. For example, datetime arrays will have been already converted into strings on validation.
However, in 3.x, validation will run **before** marshaling. Datetime arrays will be still arrays on validation. Hence, it is very hard to dealing with them. Let's assume you develop a reservation system. You would notice that it is hard to compare two dates in 3.x. I think it would be better to discuss how it should work in 4.0.

markstory:
@chinpei215 I think these changes are still an improvement over what we have now and make a more expressive API. The contention thus far has been around adding new methods in a bugfix release. We did that frequently in the past and it often worked out poorly.

ravage84:
I totally agree with @markstory. Let's change the target to 3.next (3.7) and let's get this money.

chinpei215:
then I would like to close this pull request as this change cannot help legacy applications.

markstory:
@chinpei215 Even if this is added to 3.6 it cannot help existing applications unless the developers opt-in to using the new methods.

chinpei215:
I thought we obtained a consensus when we discussed this issue in dev channel. Closing, as targeting 3.next was not my intention. It would be better to discuss how to fix it again.

ravage84:
@markstory shall I take @chinpei215 code and open a PR targeting 3.7 as a feature addition?

markstory:
@ravage84 Sure. Its a good set of changes.

ravage84:
@markstory I've taken @chinpei215's code and opened #12696

PR #12489, opened 2018-08-21
Update phpcs rules.

ADmad:


src/Database/Type/DateTimeType.php
```diff
@@ -352,7 +352,8 @@ protected function _setClassName(string $class, string $fallback): void
             $class = $fallback;
         }
         $this->_className = $class;
-        $this->_datetimeInstance = new $this->_className;
+        // @codingStandardsIgnoreLine
```
ADmad: These ignores can be removed once stickler is updated to use phpcs `3.3.1`.

markstory: That's done now :+1:

PR #12495, opened 2018-08-23
Remove deprecated IntegrationTestCase.

ADmad:


ADmad:
The `phpcs` and `phpstan` build failures have been fixed in #12494.

src/TestSuite/IntegrationTestCase.php
```diff
@@ -1,32 +0,0 @@
...
- *
- * This test class provides a number of helper methods and features
- * that make dispatching requests and checking their responses simpler.
- * It favours full integration tests over mock objects as you can test
- * more of your code easily and avoid some of the maintenance pitfalls
- * that mock objects create.
- *
- * @deprecated 3.7.0 Use Cake\TestSuite\IntegrationTestTrait instead
- */
-abstract class IntegrationTestCase extends TestCase
```
markstory: I was thinking we could keep this to make upgrading easier. This is a low cost to us class to keep around and removing it represents another tedious update for developers.

ADmad: Sounds reasonable. I'll add it back. Should I keep the deprecated tag?

Issue #8861, opened 2016-05-19
Prefix Error Templates


steefaan:
If I create a route prefix e.g. `Admin`, I'm able to put my templates to `Templates/Admin/*`. This works fine and is pretty useful compared to Cake 2.x. My `Admin` layout is completely different to my default layout and I hoped it's also possible to create a `Templates/Admin/Error` directory. Unfortunately it isn't. I think it would be nice to add this functionality for consistency reasons, what do you think?

markstory:
Sounds reasonable to me.

dereuromark:
@steefaan Are you able to PR this enhancement?

allanmcarvalho:
I solved this problem but was changing the beforeRender function inside the **ErrorController**. So, should we propose the solution to this within the **cakephp** or **app** github project?

markstory:
Sounds like a change to be made in cakephp as that is where the default errorcontroller is for most applications.

allanmcarvalho:
I agree @markstory , because thus updating CakePHP, everyone gets this fix. But I also had another similar problem. When using a theme, I can not create a custom layout and views within the theme, as the initializer of ErrorController overrides the initializer of the AppController which is responsible for changing the theme in my case. (This can be easily fixed in my project, but we are talking about an improvement for everyone.)
I suggest debating a solid solution to this problem, making the error layouts and views as easily customizable as all other parts of the framework, which we, we call **automagic** =D.

In summary, I suggest that by default, if the error happened on a page that uses theme and/or prefix, CakePHP first checks to see if there is a layout and view error files for that prefix or theme. If it does not exist, it loads the defaults. If it is a theme being applied over a route with a prefix, it first checks if there are files in the theme, if it does not exist, check in the prefix and if it does not exist use the default. I think that would be pretty solid. What do you think?

markstory:
I think we should use the same theme/prefix conventions that are used elsewhere in the framework. That will lead to the least amount of surprise and minimize the documentation/testing required.

Propagating the theme is not going to be easy as themes can be set in many places during an application. But preserving the prefix sounds very doable. @allanmcarvalho Are you able to start a pull request for this change?

dereuromark:
This also somewhat refs https://github.com/cakephp/cakephp/issues/8099 which needs the opposite to work, if no ext template then fallback to the main one.
So I guess if we can find some more dynamic/fallbacky way to do things here it would solve two problems at once.

allanmcarvalho:
[...]

PR #12499, opened 2018-08-25
4.x error template path

ADmad:
Refs #8861.

I didn't want to make this change for 3.x and cause any potential problems with existing apps.

### DIFF
--- a/src/Controller/ErrorController.php
+++ b/src/Controller/ErrorController.php
@@ -15,6 +15,7 @@ declare(strict_types=1);
  */
 namespace Cake\Controller;
 
+use Cake\Core\Configure;
 use Cake\Event\EventInterface;
 
 /**
@@ -42,6 +43,8 @@ class ErrorController extends Controller
      */
     public function beforeRender(EventInterface $event)
     {
-        $this->viewBuilder()->setTemplatePath('Error');
+        if (Configure::read('debug')) {
+            $this->viewBuilder()->setTemplatePath('Error');
+        }
     }
 }
--- a/src/Controller/ErrorController.php
+++ b/src/Controller/ErrorController.php
@@ -43,8 +43,10 @@ class ErrorController extends Controller
      */
     public function beforeRender(EventInterface $event)
     {
-        if (Configure::read('debug')) {
-            $this->viewBuilder()->setTemplatePath('Error');
+        $builder = $this->viewBuilder();
+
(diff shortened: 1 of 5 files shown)


src/Controller/ErrorController.php
```diff
@@ -42,6 +43,8 @@ public function initialize(): void
      */
     public function beforeRender(EventInterface $event)
     {
-        $this->viewBuilder()->setTemplatePath('Error');
+        if (Configure::read('debug')) {
```
ADmad: The templates for dev errors like `missing_controller.ctp` reside in non prefixed folder hence when debug is on we always use non prefix error folder.

ADmad:
I wonder if it would be better to move the setting to template path inside `ExceptionRenderer::render()`.
If template to render is `error400` or `error500` use prefixed template path else always use non prefixed path.

src/Controller/ErrorController.php
markstory: Could this interact poorly with custom exceptions?

ADmad: Depends on from which folder people except the template to be used for those custom exceptions.

Which is why I think it might be better to move the template folder selection inside exception renderer itself as I have mentioned below. If template is error400 or error500 use prefixed folder (if it's a prefixed route) else not.

tests/TestCase/Error/ExceptionRendererTest.php
```diff
@@ -209,6 +209,43 @@ public function testControllerInstanceForPrefixedRequest()
...
+        $this->assertEquals('missingAction', $controller->viewBuilder()->getTemplate());
+        $this->assertEquals('Error', $controller->viewBuilder()->getTemplatePath());
+
+        Configure::write('debug', false);
+        $ExceptionRenderer = new ExceptionRenderer($exception, $request);
+
+        $ExceptionRenderer->render();
+        $controller = $ExceptionRenderer->__debugInfo()['controller'];
+        $this->assertEquals('error400', $controller->viewBuilder()->getTemplate());
+        $this->assertEquals('Admin/Error', $controller->viewBuilder()->getTemplatePath());
```
markstory: This assertion is failing on AppVeyor because DIRECTORY_SEPARATOR is different there.

ADmad:
`DIRECTORY_SEPARATOR` fixed in test.

PR #12500, opened 2018-08-25
Always call custom exception renderer method.

### LINKED ISSUE #9905: Custom methods of ErrorRenderer are called only on debug mode
This is a (multiple allowed):

* [ ] bug
* [x] enhancement

* CakePHP Version: 3.3.10

### What you did

1. Declare `AppExceptionRenderer`.
```php
namespace App\Error;

use Cake\Error\ExceptionRenderer;

class AppExceptionRenderer extends ExceptionRenderer
{
    public function missingController($error) {
        die;
    }
}
```

2. Configure `app.php` to use `AppExceptionRenderer` and set debug mode to false.

```php
// In app.php
'debug' => false,

// ...

'Error' => [
    'exceptionRenderer' => 'App\Error\AppExceptionRenderer'
    // ...
],
```

3. Access some non-existent page.

```
/some/non-existent/page
```

### What happened

The `AppExceptionRenderer::missingController()` method was not called.

The `AppExceptionRenderer::missingController()` method should be called. In CakePHP 2, custom error rendering methods are called regardless of debug mode.

Is there any reason why debug mode is checked [here](https://github.com/cakephp/cakephp/blob/18e5d31a8a5bd6f875790a930e3ff6d2be106f93/src/Error/ExceptionRenderer.php#L161-L166)?

(Reported by **kino** in #japanese channel on Slack)

ADmad:
Closes #9905.

### DIFF
--- a/src/Error/ExceptionRenderer.php
+++ b/src/Error/ExceptionRenderer.php
@@ -191,10 +191,7 @@ class ExceptionRenderer implements ExceptionRendererInterface
         $template = $this->_template($exception, $method, $code);
         $unwrapped = $this->_unwrap($exception);
 
-        $isDebug = Configure::read('debug');
-        if (($isDebug || $exception instanceof HttpException) &&
-            method_exists($this, $method)
-        ) {
+        if (method_exists($this, $method)) {
             return $this->_customMethod($method, $unwrapped);
         }
 
@@ -216,6 +213,8 @@ class ExceptionRenderer implements ExceptionRendererInterface
             'code' => $code,
             '_serialize' => ['message', 'url', 'code']
         ];
+
+        $isDebug = Configure::read('debug');
         if ($isDebug) {
             $viewVars['trace'] = Debugger::formatTrace($unwrapped->getTrace(), [
                 'format' => 'array',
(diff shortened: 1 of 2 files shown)


src/Error/ExceptionRenderer.php
```diff
@@ -191,10 +191,7 @@ public function render()
         $template = $this->_template($exception, $method, $code);
         $unwrapped = $this->_unwrap($exception);
 
-        $isDebug = Configure::read('debug');
-        if (($isDebug || $exception instanceof HttpException) &&
-            method_exists($this, $method)
-        ) {
+        if (method_exists($this, $method)) {
```
markstory: This change will require us to document this in the migration guide and docs. In the past custom exception pages could be 'dev' only but now they are used in production mode too. I suspect there may be custom application error pages that are not unlike our own 'missing action' error pages, that emit far too much information for a production mode.

ADmad: True but with debug off only either `erro400` or `error500` template is used, irrespective of method called.

https://github.com/cakephp/cakephp/blob/7a7eed81bc3e32ccf3382b126a15d46369243473/src/Error/ExceptionRenderer.php#L314

Also additional exception attributes are set as view vars only when debug if off 

https://github.com/cakephp/cakephp/blob/7a7eed81bc3e32ccf3382b126a15d46369243473/src/Error/ExceptionRenderer.php#L230

So I don't think there's a risk of leaking any sensitive info. Yet to be extra safe I am fine moving this patch to 4.0 instead.

markstory: If the templates are always overridden then we should be reasonably safe.

PR #12502, opened 2018-08-25
4.x request referer

### LINKED ISSUE #12402: $this->referer() returns '/' if $ref == empty. Expected a null/empty return value
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [x ] feature-discussion (RFC)

* CakePHP Version: 3.5.1

### What you did
I needed to check the referer on a controller for some params, so i used $this->referer(). This works when there is a referer, but when the page is called directly, the call to referer() returns '/'. 

 ```
   public function referer($local = false)
    {
        $ref = $this->getEnv('HTTP_REFERER');

        $base = Configure::read('App.fullBaseUrl') . $this->webroot;
        if (!empty($ref) && !empty($base)) {
            if ($local && strpos($ref, $base) === 0) {
                $ref = substr($ref, strlen($base));
                if (!strlen($ref)) {
                    $ref = '/';
                }
                if ($ref[0] !== '/') {
                    $ref = '/' . $ref;
                }

                return $ref;
            }
            if (!$local) {
                return $ref;
            }
        }

        **return '/';**
    }

```

I was expecting an empty return value for an empty referer.

### PR DESCRIPTION
**Pull request by ADmad**
Closes #12402.

### DIFF
--- a/src/Http/ServerRequest.php
+++ b/src/Http/ServerRequest.php
@@ -565,9 +565,9 @@ class ServerRequest implements ServerRequestInterface
      *
      * @param bool $local Attempt to return a local address.
      *   Local addresses do not contain hostnames.
-     * @return string The referring address for this request.
+     * @return string|null The referring address for this request or null.
      */
-    public function referer(bool $local = false): string
+    public function referer(bool $local = false): ?string
     {
         $ref = $this->getEnv('HTTP_REFERER');
 
@@ -589,7 +589,7 @@ class ServerRequest implements ServerRequestInterface
             }
         }
 
-        return '/';
+        return null;
     }
 
     /**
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: src/Http/ServerRequest.php [hunk 1]
```diff
@@ -589,7 +589,7 @@ public function referer(bool $local = false): string
             }
         }
 
-        return '/';
+        return null;
```
markstory: What do browsers do if you set `Location: ` header to an empty value?

ADmad: I made a simple test using `<?php header('Location:');` and FF seem to redirect to same page while chrome does nothing.

I think it's upto developer to check the return value if he/she want's to use it for redirection. We have `Controller::redirection()` which returns a default value if no referrer is found for such cases.

markstory: Good point about Controller::redirect() being able to handle higher level use cases.

markstory:
I will get this merged when I get the migration guide updated.

PR #12503, opened 2018-08-25
 Improve documentation 

bancer:


lib/Cake/Model/Datasource/Database/Mysql.php
```diff
@@ -141,6 +141,13 @@ class Mysql extends DboSource {
  */
 	protected $_charsets = array();
 
+/**
+ * Map of the columns contained in a result.
+ *
+ * @var array
+ */
+	public $map;
```
ADmad: Adding new public properties is a tad more than improving documentation.

bancer: This is not a new property. See https://api.cakephp.org/2.10/source-class-Mysql.html#255.

ADmad: Okay, but why not declare it in parent class `DboSource` instead of doing it in class for each database?

bancer: Good point. I was not sure whether other classes that possibly extend DboSource need that.

lib/Cake/Model/Datasource/DboSource.php
```diff
@@ -247,6 +247,13 @@ class DboSource extends DataSource {
  */
 	protected $_methodCacheChange = false;
 
+/**
+ * Map of the columns contained in a result.
+ *
+ * @var array
+ */
+	public $map;
```
markstory: Should this be initialized to an empty array?

bancer:
Can this be merged?

PR #12505, opened 2018-08-25
Fix invalid Shell class namespace

o0h:
In `phpstan.neon`,  `Cake\Shell\Shell` is set in earlyTerminatingMethodCalls params.
It seems to be incorrect, `Cake\Consle\Shell` is right?


ADmad:
Given that the phpstan didn't report any error even with the incorrect namespace, make me wonder if that line is even required in `phpstan.neon`.

markstory:
Merging as this is 'less wrong' than it was before. We might not need that line at all, but that can be a separate change.

PR #12519, opened 2018-08-30
4.x psalm

ADmad:


src/Database/Dialect/TupleComparisonTranslatorTrait.php
```diff
@@ -81,8 +81,8 @@ protected function _transformTupleComparison(TupleComparison $expression, $query
         $conditions = ['OR' => []];
         foreach ($value as $tuple) {
             $item = [];
-            foreach (array_values($tuple) as $i => $value) {
-                $item[] = [$fields[$i] => $value];
+            foreach (array_values($tuple) as $i => $value2) {
```
ADmad: I am not sure if the reuse of `$value` here was a mistake or intentional.

src/Error/ExceptionRenderer.php
```diff
@@ -142,7 +142,7 @@ protected function _getController()
                 $class = App::className('Error', 'Controller', 'Controller');
             }
 
-            /* @var \Cake\Controller\Controller $controller */
+            /** @var string $class */
```
dereuromark: Controller was correct. just wrong doc block comment type.

src/ORM/LazyEagerLoader.php
```diff
@@ -140,6 +140,7 @@ protected function _injectResults(iterable $objects, $results, array $associatio
         $injected = [];
         $properties = $this->_getPropertyMap($source, $associations);
         $primaryKey = (array)$source->getPrimaryKey();
+        /** @param \Cake\Collection\CollectionInterface */
```
dereuromark: variable missing for param?

garas: It should be `@var`

dereuromark: Unless this is something for a closure or inline method. `function ($e) use ($primaryKey)` sure looks like it.

src/Database/Connection.php
```diff
@@ -469,6 +469,7 @@ public function commit(): bool
 
         if ($this->_transactionLevel === 0) {
             if ($this->wasNestedTransactionRolledback()) {
+                /** @var \Cake\Database\Exception\NestedTransactionRollbackException */
```
dereuromark: var missing

src/ORM/LazyEagerLoader.php
garas: `$e` looks like Entity, should be typehinted or have better varname

src/Error/ExceptionRenderer.php
ADmad: No, it's `$class` which type needed to be specified since `App::clasName()` can return null too. psalm is too smart and understands type declaration for right hand side of expressions too.

dereuromark: But how does it then know the result is a controller?
I think in this case it would be cleaner to handle null, check for it (=== null) and then bail with a meaningful exception.
That way this hack of string annotation is not necessary and the original annotation that actually is useful can be kept. What do you think?

dereuromark: In your IDE can you follow ` $controller->startupProcess();`? Can you click on it?

src/Database/Dialect/TupleComparisonTranslatorTrait.php
lorenzo: Probably a mistake

lorenzo:
@ADmad great job!

Is there a way we can get a summary of all issues being reported now? The logs are super long and tedious to read.

ADmad:
@lorenzo Thanks.

I have turned of display of code snippets in the output. Don't see any other option to make the output more compact.

dereuromark:
You want `--show-info=false`.

ADmad:
That will totally turn off any output (since all errors have been fixed). @lorenzo wants a summary.

dereuromark:
Ah ok, I thought we mainly want to keep the error output as this is the most helpful.

ADmad:
Doh, the merges have added new errors :stuck_out_tongue:.

ADmad:
We could keep just errors output too on travis. It does become a bit difficult to search for errors in between all the "INFO" notices.

dereuromark:
It is easy enough to generate the rest locally if you want to.

src/I18n/RelativeTimeFormatter.php
```diff
@@ -87,15 +86,15 @@ public function diffForHumans(ChronosInterface $date, ?ChronosInterface $other =
...
      * Format a into a relative timestring.
      *
-     * @param \DateTimeInterface $time The time instance to format.
+     * @param \Cake\I18n\Time|\Cake\I18n\FrozenTime $time The time instance to format.
      * @param array $options Array of options.
      * @return string Relative time string.
      * @see \Cake\I18n\Time::timeAgoInWords()
      */
-    public function timeAgoInWords(DateTimeInterface $time, array $options = []): string
+    public function timeAgoInWords($time, array $options = []): string
```
dereuromark: This is sad to see.
Shouldnt we then just have our own interface here that combines both? To be able to typehint sth instead of allowing anything, incl. integers coming in?

src/I18n/RelativeTimeFormatter.php
```diff
@@ -87,15 +86,15 @@ public function diffForHumans(ChronosInterface $date, ?ChronosInterface $other =
...
     {
         $options = $this->_options($options, FrozenTime::class);
-        if ($options['timezone'] && $time instanceof ChronosInterface) {
+        if ($options['timezone']) {
```
lorenzo: This looks like it will cause a regression, no?

ADmad: Not really. As per docblock this method only accepts `\Cake\I18n\Time` and `\Cake\I18n\FrozenTime` instances both of which have the `timezone()` method.

As @dereuromark mentioned above we ideally need a new interface which both `Time` and `FrozenTime` implement so that we can it as type hint and avoid any type checks inside methods.

ADmad: That said `RelativeTimeFormatter` isn't used directly by users so strong type hinting or defensive type checks aren't really necessary.

PR #12525, opened 2018-08-31
Set Controller::render() arguments to view builder.

ADmad:
This allows checking template and layout values in beforeRender callbacks.

### DIFF
--- a/src/Controller/Controller.php
+++ b/src/Controller/Controller.php
@@ -642,12 +642,12 @@ class Controller implements EventListenerInterface, EventDispatcherInterface
             return $this->response;
         }
 
-        if ($builder->getTemplate() === null && $this->request->getParam('action')) {
+        if ($builder->getTemplate() === null) {
             $builder->setTemplate($this->request->getParam('action'));
         }
 
         $view = $this->createView();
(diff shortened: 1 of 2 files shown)


saeideng:
now you can remove `$layout` arg from line 642 `$contents = $view->render($template, $layout);`

ADmad:
Yeah but passing it as argument too doesn't hurt either.

saeideng:
I want to say we can change view to use `$template` from the view obj instead of passing it again via `render()`

saeideng (approved):

src/Controller/Controller.php
```diff
@@ -634,12 +642,12 @@ public function render(?string $template = null, ?string $layout = null): Respon
             return $this->response;
         }
 
-        if ($builder->getTemplate() === null && $this->request->getParam('action')) {
+        if ($builder->getTemplate() === null) {
```
markstory: Can `getParam('action')` ever return null?

ADmad: I don't think that's possible in normal execution but even if it is, you would be just setting existing null value to again null, no harm done.

PR #12527, opened 2018-09-01
Consistent result value for malformed encrypted cookies

garas:
If encrypted cookie is empty or 9 (`Q2FrZQ==.`) characters or shorter, reading cookie throws  `InvalidArgumentException('The data to decrypt cannot be empty.')`, because after removing first 9 chars, string is empty and `Security::decrypt()` receives empty string.

However reading cookie with altered encrypted data, cookie returned as empty string.

I think it should behave same in both cases and return empty string.

With exception, end user receives 500 error on each request with no explanation and must remove cookie to be able use website.

### DIFF
--- a/src/Utility/CookieCryptTrait.php
+++ b/src/Utility/CookieCryptTrait.php
@@ -128,9 +128,15 @@ trait CookieCryptTrait
         }
         $this->_checkCipher($encrypt);
         $prefix = 'Q2FrZQ==.';
-        $value = base64_decode(substr($value, strlen($prefix)));
+        $prefixLength = strlen($prefix);
 
-        if ($value === '') {
+        if (strncmp($value, $prefix, $prefixLength) !== 0) {
+            return '';
+        }
+
+        $value = base64_decode(substr($value, $prefixLength), true);
+
+        if ($value === false ||$value === '') {
             return '';
         }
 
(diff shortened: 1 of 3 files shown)


src/Utility/CookieCryptTrait.php
```diff
@@ -129,6 +129,11 @@ protected function _decode($value, $encrypt, $key)
         $this->_checkCipher($encrypt);
         $prefix = 'Q2FrZQ==.';
         $value = base64_decode(substr($value, strlen($prefix)));
```
garas: It could be checked for valid signature and base64 decoded in strict mode. Should it be changed in 3.x or 4.x?

markstory: What would the error/result be on malformed data?

garas: In all cases of malformed data, cookie should be rejected and result should be the same (currently empty string).

Currently first 9 bytes can be anything, they are removed. Cookie should be rejected with invalid signature.

Currently base64 encoded data can contain characters from outside the base64 alphabet, invalid characters are silently discarded. In strict mode, `base64_decode` would return `false`, and cookie should be rejected.

It is just for stricter validation, it shouldn't make any problem as it was always formatted as `'Q2FrZQ==.' . base64_encode()` from Cake v1 and maybe earlier.

markstory: Ok. Those improvements sound good to me. I don't see why they couldn't be done in 3.6 as a bugfix.

PR #12530, opened 2018-09-02
Remove deprecated code.

ADmad:


ADmad:
Why can't the following method be removed in 4.0 itself?

https://github.com/cakephp/cakephp/blob/0f65f4530bd1e9e4ecfdb99c59c81c92897fa7b2/src/Error/BaseErrorHandler.php#L171-L173

src/TestSuite/IntegrationTestTrait.php
```diff
@@ -223,25 +223,9 @@ public function tearDown()
...
-    /**
-     * Toggle whether or not you want to use the HTTP Server stack.
-     *
-     * @param bool $enable Enable/disable the usage of the HTTP Stack.
-     * @return void
-     */
-    public function useHttpServer(bool $enable): void
-    {
-        if ($enable === false) {
-            deprecationWarning('Calling `useHttpServer(false)` does nothing, and will be removed.');
```
markstory: We previously had talked about leaving this in to make upgrading easier. This is a pretty simple thing to remove with rector though if we want to go down that path.

ADmad: Not really a fan of keeping a no-op method. All people have to do is remove a statement.

markstory: Ok.

ADmad: @markstory I add `IntegrationTestCase::useHttpServer()` instead and most upgrading from 3.x would be still extending that class instead of directly using this trait.

PR #12534, opened 2018-09-03
Implemented a new takeLast() method in Collection

### LINKED ISSUE #12531: Enhancement to Collection::last
This is a (multiple allowed):

* [ ] bug
* [x] enhancement

* CakePHP Version: 3.6.10
* Platform and Target: PHP 7.2.8 

### What you did
```php
$scriptParts = new \Cake\Collection\Collection($script->parts);
$count = $scriptParts->count();
$finalScriptParts = $scriptParts->take(2, $count - 2);
```

### What happened
I got the last two items of the collection.

### What you would like to happen
I think it would be much easier if `Collection::last()` implemented a single parameter for the number of items to get from the end of the collection.

### Examples
```php
$collection = new Collection(['first', 'second', 'third']);
$lastTwo = $collection->last(2);
// ['second', 'third']

$collection = new Collection(['first', 'second', 'third']);
$tooMany = $collection->last(4);
// ['first', 'second', 'third']
// Or out of range exception perhaps?

$collection = new Collection(['first', 'second', 'third']);
$tooMany = $collection->last();
// ['third']

$collection = new Collection(['first', 'second', 'third']);
$tooMany = $collection->last(-1);
// Out of range exception perhaps?
```

lorenzo:
I thought it was a good idea to have this method, and also found it a fun challenge

closes #12531

### DIFF
--- a/src/Collection/CollectionInterface.php
+++ b/src/Collection/CollectionInterface.php
@@ -620,6 +620,25 @@ interface CollectionInterface extends Iterator, JsonSerializable
     public function last();
 
     /**
+     * Returns the last N elements of a collection
+     *
+     * ### Examples:
+     *
+     * ```
+     * $items = [1, 2, 3, 4, 5];
+     *
+     * $last = (new Collection($items))->last(3);
+     *
+     * // Result will look like this when converted to array
+     * [3, 4, 5];
+     * ```
+     *
+     * @param int $howMany The number of elements at the end of the collection
+     * @return \Cake\Collection\CollectionInterface
+     */
+    public function lastN($howMany);
+
+    /**
      * Returns a new collection as the result of concatenating the list of elements
      * in this collection with the passed list of elements
      *
--- a/src/Collection/CollectionInterface.php
+++ b/src/Collection/CollectionInterface.php
@@ -560,7 +560,7 @@ interface CollectionInterface extends Iterator, JsonSerializable
     /**
      * Returns the last N elements of a collection
      *
-     * ### Examples:
+     * ### Example:
      *
      * ```
      * $items = [1, 2, 3, 4, 5];
--- a/src/Collection/CollectionInterface.php
+++ b/src/Collection/CollectionInterface.php
@@ -558,6 +558,25 @@ interface CollectionInterface extends Iterator, JsonSerializable
     public function take($size = 1, $from = 0);
 
     /**
+     * Returns the last N elements of a collection
+     *
+     * ### Examples:
+     *
+     * ```
+     * $items = [1, 2, 3, 4, 5];
+     *
+     * $last = (new Collection($items))->lastN(3);
+     *
+     * // Result will look like this when converted to array
+     * [3, 4, 5];
+     * ```
+     *
+     * @param int $howMany The number of elements at the end of the collection
+     * @return \Cake\Collection\CollectionInterface
+     */
+    public function takeLast($howMany);
+
+    /**
      * Returns a new collection that will skip the specified amount of elements
      * at the beginning of the iteration.
      *
@@ -618,25 +637,6 @@ interface CollectionInterface extends Iterator, JsonSerializable
      * @return mixed The last value in the collection will be returned.
      */
     public function last();
-
-    /**
-     * Returns the last N elements of a collection
-     *
-     * ### Examples:
-     *
-     * ```
-     * $items = [1, 2, 3, 4, 5];
-     *
-     * $last = (new Collection($items))->lastN(3);
-     *
-     * // Result will look like this when converted to array
-     * [3, 4, 5];
-     * ```
-     *
-     * @param int $howMany The number of elements at the end of the collection
-     * @return \Cake\Collection\CollectionInterface
-     */
-    public function lastN($howMany);
 
     /**
      * Returns a new collection as the result of concatenating the list of elements
--- a/src/Collection/CollectionInterface.php
+++ b/src/Collection/CollectionInterface.php
@@ -565,7 +565,7 @@ interface CollectionInterface extends Iterator, JsonSerializable
      * ```
      * $items = [1, 2, 3, 4, 5];
      *
-     * $last = (new Collection($items))->lastN(3);
+     * $last = (new Collection($items))->takeLast(3);
      *
      * // Result will look like this when converted to array
      * [3, 4, 5];
--- a/src/Collection/CollectionInterface.php
+++ b/src/Collection/CollectionInterface.php
@@ -627,7 +627,7 @@ interface CollectionInterface extends Iterator, JsonSerializable
      * ```
      * $items = [1, 2, 3, 4, 5];
      *
-     * $last = (new Collection($items))->last(3);
+     * $last = (new Collection($items))->lastN(3);
      *
      * // Result will look like this when converted to array
      * [3, 4, 5];
--- a/src/Collection/CollectionTrait.php
+++ b/src/Collection/CollectionTrait.php
@@ -405,7 +405,7 @@ trait CollectionTrait
     public function takeLast($howMany)
     {
         if ($howMany < 1) {
-            throw new \InvalidArgumentException("lastN requires a number greater than 0");
+            throw new \InvalidArgumentException("takeLast requires a number greater than 0");
         }
 
         $iterator = $this->optimizeUnwrap();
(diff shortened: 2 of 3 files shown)


tests/TestCase/Collection/CollectionTest.php
```diff
@@ -2150,6 +2150,78 @@ public function testLastWithTraversableNonIterator()
...
+     * @dataProvider simpleProvider
+     * @return void
+     */
+    public function testLasNtWithNegative($data)
+    {
+        $collection = new Collection($data);
+        $this->expectException(\InvalidArgumentException::class);
+        $result = $collection->lastN(-1)->toArray();
+        $expected = [];
+        $this->assertEquals($expected, $result);
```
lorenzo: Hmm, this 2 lines should not be here, will remove them tomorrow

src/Collection/CollectionInterface.php
```diff
@@ -619,6 +619,25 @@ public function first();
...
 
+    /**
+     * Returns the last N elements of a collection
+     *
+     * ### Examples:
+     *
+     * ```
+     * $items = [1, 2, 3, 4, 5];
+     *
+     * $last = (new Collection($items))->last(3);
```
jeremyharris: I think this example should say `->lastN(3)`, no?

lorenzo: Yes

ADmad:
@lorenzo I made updates as per comments :slightly_smiling_face:.

lorenzo:
@ADmad thanks!

davidyell:
Thanks @lorenzo 👍

src/Collection/CollectionInterface.php
```diff
@@ -619,6 +619,25 @@ public function first();
...
+     * $last = (new Collection($items))->lastN(3);
+     *
+     * // Result will look like this when converted to array
+     * [3, 4, 5];
+     * ```
+     *
+     * @param int $howMany The number of elements at the end of the collection
+     * @return \Cake\Collection\CollectionInterface
+     */
+    public function lastN($howMany);
```
markstory: Is tailFetch() a better name? :bike:

src/Collection/CollectionTrait.php
```diff
@@ -399,6 +399,67 @@ public function last()
...
+            return new Collection($iterator);
+        }
+
+        $generator = function ($iterator, $howMany) {
+            $result = [];
+            $bucket = 0;
+            $offset = 0;
+            $toggleOffset = false;
+
+            foreach ($iterator as $k => $item) {
```
markstory: This loop is pretty dense are there any comments that could be added for to explain why results are being sliced into buckets?

src/Collection/CollectionInterface.php
davidyell: Why isn't it just a param for `last()`?

lorenzo: - `last` always returns a single element, not a collection
- `tail` in order languages means everything but the first element. It would be confusing to me, at least
-  I liked `lastN` because it is really close when searching for last or autocompleting.

lorenzo:
@markstory I hope my comments clarify the algorithm that I used. Also thanks for asking me to do it, since explaining it also made me realize I did not need some extra code I had before

src/Collection/CollectionInterface.php
garas: I suggest `takeLast()` as counterpart of `take()`

ADmad: > tail in order languages means everything but the first element.

Guess it's my ignorance but only other popular `tail` I know of is in Unix and it does allow specifying N number of lines to get from the end of file.

lorenzo: @ADmad well, I'm not opposed to using `tail` it was just weird for me. I agree that there are also many languages where tail has a different meaning

lorenzo: `takeLast` is also a great suggestion @garas

src/Collection/CollectionTrait.php
```diff
@@ -399,6 +399,105 @@ public function last()
...
+             * - $tail = elements to the left =  [9]
+             * - $result = $head + $tail = [6, 7, 8, 9]
+             *
+             * The logic above applies to collections of any size.
+             */
+
+            foreach ($iterator as $k => $item) {
+                $result[$bucket] = [$k, $item];
+                $bucket = (++$bucket) % $howMany;
+                $offset++;
```
garas: `$bucket` and `$offset` holds same data, only one modulo is computed inside loop, another after loop.

lorenzo: Yeah, I realised that, but somehow I find it more readable to keep 2 variables given the explanation of what the algorithm does. If it is not a big problem, I'd like to keep both

lorenzo:
Renamed to `takeLast()`

src/Collection/CollectionInterface.php
markstory: :+1: to `takeLast`. I agree that `tail` means a different thing in functional languages and we shouldn't use it for this.

src/Collection/CollectionTrait.php
```diff
@@ -399,6 +399,105 @@ public function last()
...
+             * elements to the right for the marker are the head of the final result,
+             * whereas the elements at the left are the tail. For example consider step 5)
+             * which has an offset of 1:
+             *
+             * - $head = elements to the right = [6, 7, 8]
+             * - $tail = elements to the left =  [9]
+             * - $result = $head + $tail = [6, 7, 8, 9]
+             *
+             * The logic above applies to collections of any size.
+             */
```
markstory: :+1: Thank you for this. It is much easier to understand now.

src/Collection/CollectionTrait.php
```diff
@@ -399,6 +399,105 @@ public function last()
...
+             *
+             * 0) 1 2 3 4 -- $bucket now goes back to 0, we have filled 4 elementes
+             * 1) 5 2 3 4 -- 5th iteration
+             * 2) 5 6 3 4 -- 6th iteration
+             * 3) 5 6 7 4 -- 7th iteration
+             * 4) 5 6 7 8 -- 8th iteration
+             * 5) 9 6 7 8
+             *
+             *  We can see that at the end of the iterations, the buffer contains all
+             *  the last for elements, just in the wrong order. How do we keep the
```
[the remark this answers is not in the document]
lorenzo: yes

src/Collection/CollectionInterface.php
```diff
@@ -557,6 +557,25 @@ public function sample($size = 10);
...
 
+    /**
+     * Returns the last N elements of a collection
+     *
+     * ### Examples:
+     *
+     * ```
+     * $items = [1, 2, 3, 4, 5];
+     *
+     * $last = (new Collection($items))->lastN(3);
```
saeideng: `takeLast()`

src/Collection/CollectionTrait.php
```diff
@@ -399,6 +399,105 @@ public function last()
         return $result;
     }
 
+    /**
+     * {@inheritDoc}
+     */
+    public function takeLast($howMany)
+    {
+        if ($howMany < 1) {
+            throw new \InvalidArgumentException("lastN requires a number greater than 0");
```
saeideng: `takeLast() requires .... `

src/Collection/CollectionTrait.php
```diff
@@ -399,6 +399,105 @@ public function last()
         return $result;
     }
 
+    /**
+     * {@inheritDoc}
+     */
+    public function takeLast($howMany)
+    {
+        if ($howMany < 1) {
```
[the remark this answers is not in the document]
lorenzo: It will be strict in 4.0, we are not as strict anywhere else for scalar in most other places

src/Collection/CollectionTrait.php
```diff
@@ -399,6 +399,105 @@ public function last()
...
+            if ($count === 0) {
+                return new Collection([]);
+            }
+
+            $iterator = new LimitIterator($iterator, max(0, $count - $howMany), $howMany);
+
+            return new Collection($iterator);
+        }
+
+        $generator = function ($iterator, $howMany) {
```
ravage84: Could we move that to a function or even small class which has your great explanation as doc block?

ravage84 (approved):
:+1: 
Added some small improvments and the usual nitpicking :smirk_cat:

lorenzo:
Thanks @ravage84

src/Collection/CollectionTrait.php
lorenzo: Sure, feel free to send a PR if  you want to work on that

PR #12541, opened 2018-09-05
Added a collection lazy() method

lorenzo:
Since cake 3.6 collections are not entirely lazy anymore. This is
because in the greate majority of cases, running the computations
eagerly is faster than doing it on demand.

This assumption fails, though, when the computations are very intensive
or when you actually only need a handful of elements anyways. In order
to get this feature back, I'm introducing the `lazy()` method. It will
make sure that computations only happen on demand.

I feel that this give more power to the developer who's trying to find
the most performant way of traversing a set of results.

### DIFF
--- a/src/Collection/CollectionInterface.php
+++ b/src/Collection/CollectionInterface.php
@@ -784,6 +784,16 @@ interface CollectionInterface extends Iterator, JsonSerializable
     public function compile($preserveKeys = true);
 
     /**
+     * Returns a new collection where any operations chained after it are guaranteed
+     * to be run lazily. That is, elements will be yieleded one at a time.
+     *
+     * A lazy collection can only be iterated once. A second attempt results in an error.
+     *
+     * @return \Cake\Collection\CollectionInterface
+     */
+    public function lazy();
+
+    /**
      * Returns a new collection where the operations performed by this collection.
      * No matter how many times the new collection is iterated, those operations will
      * only be performed once.
--- a/tests/TestCase/Collection/CollectionTest.php
+++ b/tests/TestCase/Collection/CollectionTest.php
@@ -2634,7 +2634,7 @@ class CollectionTest extends TestCase
     }
 
     /**
-     * Tests that elements in a lazy collection are only fetched once
+     * Tests that elements in a lazy collection are not fetched immediately.
      *
      * @return void
      */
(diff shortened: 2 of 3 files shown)


lorenzo:
I'll try to improve the code coverage a bit and then will merge

ravage84:
> Since cake 3.6 collections are not entirely lazy anymore. This is
because in the greate majority of cases, running the computations
eagerly is faster than doing it on demand.

@lorenzo  are there deeper implications because of this? Did we update the docs about this?

src/Collection/CollectionInterface.php
```diff
@@ -783,6 +783,16 @@ public function jsonSerialize();
      */
     public function compile($preserveKeys = true);
 
+    /**
+     * Returns a new collection where any operations chained after it are guaranteed
+     * to be run lazily. That is, elements will be yieleded one at a time.
+     *
+     * A lazy collection can only be iterated once. A second attempt results in an error.
```
ravage84: :+1:  Good note! This should be included in the Cookbook, too.

ravage84: Would there be any value in a test verifiing this behavior? (if we aren't doing that already)

tests/TestCase/Collection/CollectionTest.php
```diff
@@ -2632,4 +2632,21 @@ public function testArrayIteratorExtend()
         $this->assertTrue($newIterator->checkValues());
         $this->assertCount(3, $newIterator->toArray());
     }
+
+    /**
+     * Tests that elements in a lazy collection are only fetched once
```
ravage84: I think this title is a bit misleading.  While implicitly true, we deliberately test that the the filter callable will not be called for the lazy collection without exporting to array for example.

Shouldn't we add a test that verifies the actual expected behavior mention in this title?

ravage84 (changes requested):
> I feel that this give more power to the developer who's trying to find
the most performant way of traversing a set of results.

Agree, choice is most often a good thing.

tests/TestCase/Collection/CollectionTest.php
lorenzo: The test is actually verifying said behaviour. If the collection wasn't lazy, then the filters would e called immediately

ravage84: I agree. It's just a wording thing, I guess.
How about rewording to: "Tests that elements in a lazy collection are not fetched immediately"?

lorenzo: sounds good to me

ravage84 (approved):

Issue #12509, opened 2018-08-27
IntegerType returns unexpected results with UUID


sdustinh:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

```
sdustinh@sdustinh:~/github/app$ php --version
PHP 7.1.20-1+ubuntu16.04.1+deb.sury.org+1 (cli) (built: Jul 25 2018 10:06:40) ( NTS )
Copyright (c) 1997-2018 The PHP Group
Zend Engine v3.1.0, Copyright (c) 1998-2018 Zend Technologies
    with Zend OPcache v7.1.20-1+ubuntu16.04.1+deb.sury.org+1, Copyright (c) 1999-2018, by Zend Technologies
    with Xdebug v2.6.0, Copyright (c) 2002-2018, by Derick Rethans
sdustinh@sdustinh:~/github/app$ mysql --version
mysql  Ver 14.14 Distrib 5.7.23, for Linux (x86_64) using  EditLine wrapper
sdustinh@sdustinh:~/github/app$ composer show cakephp/*
cakephp/bake                1.6.4  Bake plugin for CakePHP 3
cakephp/cakephp             3.5.17 The CakePHP framework
cakephp/cakephp-codesniffer 3.0.5  CakePHP CodeSniffer Standards
cakephp/chronos             1.2.2  A simple API extension for DateTime.
cakephp/debug_kit           3.14.2 CakePHP Debug Kit
cakephp/elastic-search      0.3.5  An Elastic Search datasource and data mapper for CakePHP 3.0
cakephp/migrations          1.7.2  Database Migration plugin for CakePHP 3.0 based on Phinx
cakephp/plugin-installer    1.1.0  A composer installer for CakePHP 3.0+ plugins.
```

I understand that this is arguably a bug, but in the course of legitimate usage it has caused some problems so I'm listing it as a bug for this issue. The [IntegerType.php](https://github.com/cakephp/cakephp/blob/master/src/Database/Type/IntegerType.php) has a small but really annoying bug that pops up occasionally in my application and it a pain to track down every time. If you pass a UUID to query conditions where the field is an integer the `IntegerType` will convert the UUID to an integer and return unexpected results. Consider the following:

```php
// UsersTable.php
public function findUser(Query $query, array $options = []): Query
{
    return $query->where([
        'OR' => [
            [$this->aliasField('id') => $options['id']],
            [$this->aliasField('uuid') => $options['id']],
        ],
[...]

PR #12543, opened 2018-09-06
Check for `is_numeric()` instead of `is_scalar()` on IntegerType

sdustinh:
Reference: https://github.com/cakephp/cakephp/issues/12509

### DIFF
--- a/src/Database/Type/IntegerType.php
+++ b/src/Database/Type/IntegerType.php
@@ -98,9 +98,17 @@ class IntegerType extends Type implements TypeInterface, BatchCastingInterface
     public function manyToPHP(array $values, array $fields, Driver $driver)
     {
         foreach ($fields as $field) {
-            if (!isset($values[$field]) || !is_numeric($values[$field])) {
+            if (!isset($values[$field])) {
                 continue;
             }
+
+            if ($values[$field] !== null && $values[$field] !== '' && !is_numeric($values[$field])) {
+                throw new InvalidArgumentException(sprintf(
+                    'Cannot convert value of type `%s` to integer',
+                    getTypeName($values[$field])
+                ));
+            }
+
             $values[$field] = (int)$values[$field];
         }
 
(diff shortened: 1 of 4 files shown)


src/Database/Type/IntegerType.php
```diff
@@ -98,7 +98,7 @@ public function toPHP($value, Driver $driver)
     public function manyToPHP(array $values, array $fields, Driver $driver)
     {
         foreach ($fields as $field) {
-            if (!isset($values[$field])) {
+            if (!isset($values[$field]) || !is_numeric($values[$field])) {
```
lorenzo: It should throw instead of drop the value

markstory: :+1: we drop values in marshalling, and raise when converting to PHP or DB values.

sdustinh:
@lorenzo @markstory If I'm going to throw the exception for non-numeric values in `manyToPHP()`, do you want me to emulate `toDatabase()` and `continue` for `null` and `''` or throw the exception for them as well?

markstory:
`null` and `''` should be cast into null and don't require exceptions.

markstory:
There are a few tests that are failing. Those tests will need to be updated before this can be merged in.

sdustinh:
:+1: Just wanted to make sure we were all okay with the changes before fixing the tests.

src/Database/Type/IntegerType.php
```diff
@@ -101,6 +101,14 @@ public function manyToPHP(array $values, array $fields, Driver $driver)
             if (!isset($values[$field])) {
                 continue;
             }
+
+            if ($values[$field] !== null && $values[$field] !== '' && !is_numeric($values[$field])) {
```
lorenzo: * no need for `$values[$field] !== null` there is an `isset` check for the value a few lines above

* no need for `$values[$field] !== ''` the `is_numeric`, we want the `is_numeric` check to catch it

sdustinh: :+1: removed them

Issue #12546, opened 2018-09-07
Missing proxy username/password in Http\Client\Adapter\Curl


ludeus:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: ec33e6f7ae0ce7f9154ce20a4a71c3ce51d52b2d
* Platform and Target: 

### What you did

```php
$http = new Cake\Http\Client(['adapter' => new Cake\Http\Client\Adapter\Curl]);
$http->get('https://www.google.com', [], ['proxy' => ['proxy' => 'proxy.xxx.com', 'username' => 'user', 'password' => 'password'])->getStatusCode();
```

### What happened

Get the error:  
Cake\Http\Exception\HttpException with message 'cURL Error (56) Received HTTP code 407 from proxy after CONNECT'

### What you expected to happen

Get the return code: 200

### Proposed solution

In `Cake\Http\Client\Adapter\Curl::buildOptions()`, add the following code:
```php
if (isset($options['proxy']['username'])) {
    $password = !empty($options['proxy']['password']) ? $options['proxy']['password'] : '';
    $out[CURLOPT_PROXYUSERPWD] = $options['proxy']['username'] . ':' . $password;
}
```

NOTE: have a warning now:  
__PHP Notice:  Undefined offset: 1 in vendor/cakephp/cakephp/src/Http/Client/Response.php on line 217__

It's from _parseHeaders() :  
`list($name, $value) = explode(':', $value, 2);`  

the value is an empty string.
adding the following solved the problem:  
```php
if (strpos($value, ':') === false) {
    continue;
}
```

markstory:
Given that you know what the required fix is, would you be willing to send a pull request with the necessary changes?

ludeus:
Doing it

PR #12547, opened 2018-09-07
Fix proxy authentication in curl adapter.

ludeus:
Issue #12546


markstory:
These commits have been rebased and merged into 3.next.

PR #12551, opened 2018-09-11
Adding withCookieCollection() to the Response object

burzum:


tests/TestCase/Http/ResponseTest.php
```diff
@@ -741,6 +741,21 @@ public function testCheckNotModifiedNoHints()
...
+     * @return void
+     */
+    public function testWithCookieCollection()
+    {
+        $response = new Response();
+
+        $collection = new CookieCollection();
+        $newResponse = $response->withCookieCollection($collection);
+
+        $this->assertNotSame($response->getCookieCollection(), $newResponse->getCookieCollection());
```
markstory: You might also want to assert that the responses are not the same instance, and that the old response didn't get mutated.

ADmad (commented):
I have rebased and pushed updates.

cleptric:
That's actually something I need? 😄 
@burzum are you finishing it or shall I?

ADmad:
@cleptric I believe if have already done required changes.

But I am concerned about the `CookieCollectionInterface` that is added and using the interface for return typehint. `CookieCollection` has extra methods `addToRequest()` and `addToResponse()` which do not exist on the interface.

So usage like `$response->getCookieCollection()->addFromResponse($response, $request)` will cause static analysers / IDEs to generate error like:
```
Call to an undefined method Cake\Http\Cookie\CookieCollectionInterface::addFromResponse()
```

ADmad:
`Response::withCookieCollection()` to `3.8`.

PR #12552, opened 2018-09-11
Improved support for BINARY(n) and VARBINARY(n) columns

dmromanov:
Improved support for BINARY(n) and VARBINARY(n) columns.

### DIFF
--- a/src/Database/Schema/MysqlSchema.php
+++ b/src/Database/Schema/MysqlSchema.php
@@ -137,7 +137,7 @@ class MysqlSchema extends BaseSchema
         if ($col === 'binary' && $length === 16) {
             return ['type' => TableSchema::TYPE_BINARY_UUID, 'length' => null];
         }
-        if (strpos($col, 'blob') !== false || $col === 'binary') {
+        if (strpos($col, 'blob') !== false || in_array($col, ['binary', 'varbinary'])) {
             $lengthName = substr($col, 0, -4);
             $length = isset(TableSchema::$columnLengths[$lengthName]) ? TableSchema::$columnLengths[$lengthName] : $length;
 
@@ -359,7 +359,17 @@ class MysqlSchema extends BaseSchema
                     break;
                 case TableSchema::TYPE_BINARY:
                     $isKnownLength = in_array($data['length'], TableSchema::$columnLengths);
-                    if (empty($data['length']) || !$isKnownLength) {
+
+                    if (!empty($data['length']) && !$isKnownLength) {
+                        if ($data['length'] > 2) {
+                            $out .= ' VARBINARY(' . $data['length'] . ')';
+                        } else {
+                            $out .= ' BINARY(' . $data['length'] . ')';
+                        }
+                        break;
+                    }
+
+                    if (empty($data['length'])) {
                         $out .= ' BLOB';
                         break;
                     }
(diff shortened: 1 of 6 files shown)


src/Database/Schema/SqliteSchema.php
```diff
@@ -340,6 +339,14 @@ public function columnSql(TableSchema $schema, $name)
             }
         }
 
+        if ($data['type'] === TableSchema::TYPE_BINARY) {
+            if (isset($data['length'])) {
+                $out .= ' VARBINARY(' . (int)$data['length'] . ')';
+            } else {
+                $out .= ' BLOB';
+            }
```
markstory: I don't know if these types are actually different in sqlite. They might be cosmetically different  but behave the same.

src/Database/Schema/SqlserverSchema.php
```diff
@@ -371,12 +371,26 @@ public function columnSql(TableSchema $schema, $name)
...
-                $out .= sprintf('(%s)', TableSchema::LENGTH_TINY);
+                $out .= ' VARBINARY';
+
+                if (empty($data['length']) || in_array(
+                    $data['length'],
+                    [TableSchema::LENGTH_MEDIUM, TableSchema::LENGTH_LONG],
+                    true
+                )) {
+                    $out .= '(MAX)';
+                } else {
```
markstory: This first of condition could be folded into the if ladder inside of it avoiding a layer of nesting.

tests/TestCase/Database/Schema/PostgresSchemaTest.php
```diff
@@ -203,6 +203,15 @@ public static function convertColumnProvider()
...
                 ['type' => 'binary', 'length' => null]
             ],
+            // Binary
+            [
+                ['type' => 'BIT VARYING(3)'],
+                ['type' => 'binary', 'length' => 3]
+            ],
+            [
+                ['type' => 'BIT(1)'],
+                ['type' => 'binary', 'length' => 1]
```
markstory: Bit fields in postgres are odd as they are represented with strings. I am not sure mapping them to binary will give folks the results they are expecting.

dmromanov: Sounds reasonable

src/Database/Schema/SqlserverSchema.php
dmromanov: Done.

src/Database/Schema/MysqlSchema.php
```diff
@@ -359,7 +359,17 @@ public function columnSql(TableSchema $schema, $name)
                     break;
                 case TableSchema::TYPE_BINARY:
                     $isKnownLength = in_array($data['length'], TableSchema::$columnLengths);
-                    if (empty($data['length']) || !$isKnownLength) {
+
+                    if (!empty($data['length']) && !$isKnownLength) {
+                        if ($data['length'] > 2) {
+                            $out .= ' VARBINARY(' . strtoupper($data['length']) . ')';
+                        } else {
+                            $out .= ' BINARY(' . strtoupper($data['length']) . ')';
```
markstory: Does `strtoupper()` on a number do anything?

dmromanov: Corrected :)

PR #12555, opened 2018-09-12
Add interface for I18n's date time classes.

ADmad:
This helps in typehinting and also allows to better "hide" Chronos. Lot of users are confused regarding the various I18n classes and Chronos classes.

### DIFF
--- a/src/I18n/TimeInterface.php
+++ b/src/I18n/TimeInterface.php
@@ -204,10 +204,17 @@ interface TimeInterface extends ChronosInterface, JsonSerializable
     public static function parseTime(string $time, $format = null);
 
     /**
-     * Get the difference formatter instance or overwrite the current one.
+     * Get the difference formatter instance.
      *
-     * @param \Cake\I18n\RelativeTimeFormatter|null $formatter The formatter instance when setting.
      * @return \Cake\I18n\RelativeTimeFormatter The formatter instance.
      */
-    public static function diffFormatter($formatter = null);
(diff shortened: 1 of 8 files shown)


ADmad:
The `Cookie` class uses `Chronos` to create datetime instances for `expiresAt`. Is using `Chronos` really necessary there? At some point I would like to have the Http client available as a standalone package and this usage would add unnecessary dependency on Chronos.

saeideng:
> Is using Chronos really necessary there? 

you can replace it with `DateTime/DateTimeImmutable`

markstory:
I like the idea. Having an interface does help reduce typehinting on concrete types.

src/I18n/TimeInterface.php
```diff
@@ -0,0 +1,213 @@
...
+     */
+    public static function parseTime(string $time, $format = null);
+
+    /**
+     * Get the difference formatter instance or overwrite the current one.
+     *
+     * @param \Cake\I18n\RelativeTimeFormatter|null $formatter The formatter instance when setting.
+     * @return \Cake\I18n\RelativeTimeFormatter The formatter instance.
+     */
+    public static function diffFormatter($formatter = null);
```
dereuromark: shouldnt this also be a split getter setter? and can we typehint the class?

ADmad:
This should be good to go. I'll deal with `Cookie`'s use of `Chronos` in a separate PR.

PR #12560, opened 2018-09-13
Throw warning when using underscored fixture names.

ADmad:
This will allow us to drop the implicit inflection to CameCased names in fixture manager.

### DIFF
--- a/src/Core/ConventionsTrait.php
+++ b/src/Core/ConventionsTrait.php
@@ -30,7 +30,7 @@ trait ConventionsTrait
      */
     protected function _fixtureName($name)
     {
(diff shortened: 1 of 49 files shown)


markstory:
This will require updates to the documentation and bake as well I think.

dereuromark:
One could probably leverage/use sth like https://github.com/dereuromark/upgrade/blob/develop/src/Shell/Task/FixtureLoadingTask.php to auto-upgrade their existing include statements.

ADmad:
@markstory I have open a PR for updating docs. Will open a PR for bake too once this is merged.

PR #12561, opened 2018-09-13
Allow instance of HttpApplicationInterface in RoutingMiddleware

Iandenh:
Allow instance of HttpApplicationInterface in RoutingMiddleware.

The RoutingMiddleware only needs `routes()` and checks before executing `pluginRoutes()` if `$this->app` is an instance of `PluginApplicationInterface`.

### DIFF
--- a/src/Routing/Middleware/RoutingMiddleware.php
+++ b/src/Routing/Middleware/RoutingMiddleware.php
@@ -15,8 +15,8 @@
 namespace Cake\Routing\Middleware;
 
 use Cake\Cache\Cache;
+use Cake\Core\HttpApplicationInterface;
 use Cake\Core\PluginApplicationInterface;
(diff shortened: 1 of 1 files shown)


ADmad:
Such a change shouldn't be done in bugfix release and hence should target `3.next`.

Iandenh:
Sorry, I changed the target branch.

src/Routing/Middleware/RoutingMiddleware.php
```diff
@@ -54,10 +54,10 @@ class RoutingMiddleware
     /**
      * Constructor
      *
-     * @param \Cake\Http\BaseApplication $app The application instance that routes are defined on.
+     * @param \Cake\Core\HttpApplicationInterface $app The application instance that routes are defined on.
```
dereuromark: |null is missing

PR #12564, opened 2018-09-14
Switch from master/slave in datasource readme

josegonzalez:
This also makes it slightly clearer that connections are arbitrarily named.


lorenzo:
thanks

PR #12565, opened 2018-09-15
3.next - Start using pool() internally in Cache

markstory:
Start adopting the new pool() method and PSR16 interface internally. There are a number of current day `Cache` methods that don't fit inside the PSR16 interface. I am considering adding our own interface that encompasses these methods as static tools will complain about
`pool()->increment()` in the future.

The methods that would be needed in this expanded interface are:

- increment
- decrement
- add
- clearGroup

### DIFF
--- a/src/Cache/Cache.php
+++ b/src/Cache/Cache.php
@@ -284,19 +284,19 @@ class Cache
      */
     public static function write($key, $value, $config = 'default')
     {
-        $engine = static::engine($config);
         if (is_resource($value)) {
             return false;
         }
 
-        $success = $engine->write($key, $value);
+        $backend = static::pool($config);
+        $success = $backend->set($key, $value);
         if ($success === false && $value !== '') {
             trigger_error(
                 sprintf(
                     "%s cache was unable to write '%s' to %s cache",
                     $config,
                     $key,
-                    get_class($engine)
+                    get_class($backend)
                 ),
                 E_USER_WARNING
             );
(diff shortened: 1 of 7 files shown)


ADmad:
Tests failing.

src/Cache/Cache.php
```diff
@@ -368,6 +368,7 @@ public static function writeMany($data, $config = 'default')
      */
     public static function read($key, $config = 'default')
     {
+        // In 4.x this needs to change to use pool()
```
ravage84: This is actually a "TODO", isn't it?

ravage84 (commented):
:+1:

src/Cache/Cache.php
markstory: It *is* except our lint rules will fail the build on TODOs being present, and I'm frankly too lazy to change them.

ravage84: @markstory I thought so ;-)

Would you want to accept TODOs and/or FIXMEs in code during a fast changing development phase like for 3.next and/or 4.x?

markstory: I'm fine with it, makes finding future tasks easier to grep for.

ravage84: I, personally, love using TODOs in my and our applications' code. But I also understand that you don't want to clutter your code with it.

markstory: Yeah there's a balance to be had and I feel we might be too far on one side right now.

ravage84: > It is except our lint rules will fail the build on TODOs being present, and I'm frankly too lazy to change them.

I changed that for you with a simple exclusion of this sniff with a comment.

ravage84 (approved):

PR #12566, opened 2018-09-15
Try removing the check for cached files to make the tests runs reliably.

ADmad:


ADmad:
Probably not optimal but at least the tests runs again on appveyor (until they magically don't again in future).

PR #12568, opened 2018-09-16
Warn if CsrfComponent is used together with CsrfProtectionMiddleware

garas:


src/Controller/Component/CsrfComponent.php
```diff
@@ -61,6 +61,19 @@ class CsrfComponent extends Component
...
 
+    /**
+     * Warn if CsrfComponent is used together with CsrfProtectionMiddleware
+     *
+     * @param array $config The config data.
+     * @return void
+     */
+    public function initialize(array $config)
+    {
+        if ($this->getController()->request->getParam('_csrfToken') !== false) {
```
Iandenh: Shouldn't this be using `getRequest()`?

garas: I've seen `@deprecated .... Use Component::getController() ... instead.` at `Component::request`, but forgot about `Controller::request` 😂

Anyway, this component is removed from `4.x` and wouldn't be a problem.

saeideng:
this class has been deprecated 
and adding new warning in master is a bit annoying  for people 
also , we have `deprecatedWarning()` if you need to throw deprecation Warning

src/Controller/Component/CsrfComponent.php
```diff
@@ -61,6 +61,19 @@ class CsrfComponent extends Component
         'field' => '_csrfToken',
     ];
 
+    /**
+     * Warn if CsrfComponent is used together with CsrfProtectionMiddleware
+     *
+     * @param array $config The config data.
+     * @return void
+     */
+    public function initialize(array $config)
```
saeideng: we have `startup` callback .  why new one ?

garas: I like to warn as possible. 
Also at this place `loadComponent()` is still in callstack so it might be easier to find location of that load. `deprecationWarning()` has `$stackFrame` parameter, but it should `5`, so I wasn't sure if such depth is reliable, so I used `triggerWarning()` but it stayed here.

ravage84 (changes requested):
@garas While this is a nice thing to have, I think it's more of a documentation problem, than anything else. Also, a deprecation warning might actually solve the problem, just the same as your warning does.

Could you change it to a deprection warning?

src/Controller/Component/CsrfComponent.php
```diff
@@ -61,6 +61,19 @@ class CsrfComponent extends Component
...
+    {
+        if ($this->getController()->getRequest()->getParam('_csrfToken') !== false) {
```
markstory: Won't this emit errors if the value is null?

garas: Which one?
`getParam('_csrfToken')` returns `false` if it is not set.
`getController()` and `getRequest()` are always set.

garas:
I still argue and it should be on `3.6`, not `3.7`.

It won't warn if only component or only middleware is active, so it won't annoy with extra warnings.

It will only warn if _both_ middleware and component are active at same time and forms are anyway broken.

It is to warn developer early that he has wrong setup.

Now I think it should be fatal error and `throw new \Exception('Cannot load CsrfComponent because CsrfProtectionMiddleware already loaded and would corrupt CSRF data')`, maybe different text or exception class.

src/Controller/Component/CsrfComponent.php
markstory: Ok. I thought getParam() could return null if the parameter was not set. I'm probably conflating 3.x and 4 though.

markstory:
@garas I disagree on it being a fatal error as there are likely scenarios we haven't thought of where this could happen. At least with a deprecation warning the end user can disable the message.

src/Controller/Component/CsrfComponent.php
ADmad: `ServerRequest::getParam()` returning `false` (since the default value for 2nd param is `false`) when key does not exist is a wacky quirk of 3.x which is fixed in 4.x.

PR #12569, opened 2018-09-16
4.x deprecate filesystem

ADmad:
Our `File` and `Folder` classes are old, quirky and have unexpected/inconsistent behavior based on various tickets we have had in the past.

This PR deprecates them and replace all their internal usage with classes provided by `symfony/filesystem` and `symfony/finder` packages and `SplFileInfo` instead.


ADmad:
Only noteworthy change is `Response::_file` is now a `SplFileInfo` instance instead of `Cake\Filesystem\File` instance when using `Response::withFile()`.

I wonder if keeping this reference is even necessary since `Response::withFile()` creates a `Zend\Diactoros\Stream` stream instance. The file reference seems like a relic of 3.x days.

ADmad:
I also removed file extension to mime type mapping from `AssetMiddleware`.  Our `Reponse` instance already provides such a mapping so just used that.

src/Routing/Middleware/AssetMiddleware.php
```diff
@@ -69,20 +48,17 @@ public function __construct(array $options = [])
...
      *
      * @param \Psr\Http\Message\ServerRequestInterface $request The request.
-     * @param \Psr\Http\Message\ResponseInterface $response The response.
+     * @param \Cake\Http\Response $response The response.
      * @param callable $next Callback to invoke the next middleware.
-     * @return \Psr\Http\Message\ResponseInterface A response
+     * @return \Cake\Http\Response A response
      */
-    public function __invoke(ServerRequestInterface $request, ResponseInterface $response, callable $next): ResponseInterface
+    public function __invoke(ServerRequestInterface $request, Response $response, callable $next): Response
```
dereuromark: Shouldnt that be our response interface then instead?

    ServerRequestInterface $request, ServerResponseInterface $response

would be a bit cleaner, wouldnt it?

ADmad: I personally think interfaces should be added only if there are going to be multiple implementations for it, not just to replace typehints.

src/Routing/Middleware/AssetMiddleware.php
```diff
@@ -98,13 +74,15 @@ public function __invoke(ServerRequestInterface $request, ResponseInterface $res
...
-            $headers = $response->getHeaders();
-            $headers['Last-Modified'] = date(DATE_RFC850, $modifiedTime);
-
-            return new Response('php://memory', 304, $headers);
+            return $response
+                ->withStatus(304)
+                ->withHeader(
+                    'Last-Modified',
+                    date(DATE_RFC850, $modifiedTime)
+                );
```
markstory: A 304 response shouldn't have a body. You should overwrite the body here.

ADmad: Added `->withStringBody('')`.

src/Routing/Middleware/AssetMiddleware.php
dereuromark: In light of psr comp. some might want to exchange with their or some other response class I would imagine.

composer.json
```diff
@@ -31,14 +31,15 @@
         "php": ">=7.1.0",
         "ext-intl": "*",
         "ext-mbstring": "*",
-        "cakephp/chronos": "^1.0.1",
         "aura/intl": "^3.0.0",
+        "cakephp/chronos": "^1.0.1",
+        "mikey179/vfsStream": "^1.6",
```
garas: `vfsStream` should be in `require-dev` as in https://github.com/cakephp/cakephp/pull/11931

ADmad: Yup, my bad, will fix it.

inoas:
Did you consider using the local adapter of flysystem? https://github.com/thephpleague/flysystem

ADmad:
> Did you consider using the local adapter of flysystem?

No, cause I only needed a convenience class for interacting with files, not a full abstraction layer for various file systems. I am re-evaluating if using `symfony/filesystem` is even necessary. `symfony/finder` is what's really useful due to it's inherent recursive nature and `flysystem` doesn't offer anything similar to that.

ADmad:
> I am re-evaluating if using symfony/filesystem is even necessary. 

It allows easily deleting non empty directories, so that alone is useful enough in `AssetsTask`.

markstory:
> It allows easily deleting non empty directories, so that alone is useful enough in AssetsTask.

If we only need that one function in one place I would rather us re-implement that small part over pulling in another dependency.

ADmad:
> If we only need that one function in one place I would rather us re-implement that small part over pulling in another dependency.

Folder delete and folder copy are two features of `Filesystem` used in `AssetsTask`. Although not too difficult  to implement I personally am not in favour of doing so. If we make a utility class for it we again end up having to maintaining it and respond to potential tickets relating to it.

markstory:
> If we make a utility class for it we again end up having to maintain it and respond to potential tickets relating to it.

What if the functions were internal/protected and only used by AssetTask?

ADmad:
> What if the functions were internal/protected and only used by AssetTask?

That's an option. But then if someone needs similar feature in app/plugin they would be unnecessarily duplicating the functionality or have to use external lib anyway. For e.g. the [TwbsAssetsTask](https://github.com/FriendsOfCake/bootstrap-ui/blob/develop/src/Shell/Task/TwbsAssetsTask.php) of `friendsofcake/bootstrap-ui` plugin.

markstory:
Won't other asset tasks have the same problem if we remove our filesystem classes and use symfony? In both the scenario where we take on the dependency, or where we implement just what we need as protected methods, plugins will need to update their code.

ADmad:
> ...plugins will need to update their code.

Yes they will but if we use something like symfony/filesystem plugins can reuse it instead of reimplementing the features.

inoas:
Can we keep File:: and Folder during the 4.x cycle and remove them with 5.0 (or offer an framework choice/implementation during 4.x)?

markstory:
> Can we keep File:: and Folder during the 4.x cycle and remove them with 5.0

That's exactly what this pull request is aiming to do. The internal uses of File/Folder have been removed to save future us work.

ADmad:
`FileEngine::clearGroup()` becomes lot more readable with the use of `Finder` class.

src/Routing/Middleware/AssetMiddleware.php
```diff
@@ -39,26 +38,6 @@ class AssetMiddleware
...
-        'css' => 'text/css',
-        'json' => 'application/json',
-        'js' => 'application/javascript',
-        'ico' => 'image/x-icon',
-        'eot' => 'application/vnd.ms-fontobject',
-        'svg' => 'image/svg+xml',
-        'html' => 'text/html',
-        'rss' => 'application/rss+xml',
-        'xml' => 'application/xml',
-    ];
```
markstory: Does symfony handle this now?

ADmad: No, this not directly related to using the symfony classes. 

Maintaining as map of extension to mime types in the middleware was redundant since we already have such a map in our `Response` class and can use `Response::getMimeType()` to get mimetype for an extension.

src/Mailer/Email.php
```diff
@@ -1910,9 +1909,7 @@ protected function _attachFiles($boundary = null)
      */
     protected function _readFile($path)
     {
-        $File = new File($path);
-
-        return chunk_split(base64_encode($File->read()));
+        return chunk_split(base64_encode((string)file_get_contents($path)));
```
lorenzo: This introduces a security risk. Now it is possible to fetch remote resources thanks to `file_get_contents`. We need to make sure that the file exists first.

ADmad: Should an exception be thrown is `file_exists()` check fails?

ADmad: Actually no change is needed. `Email::setAttachments()` already uses `realpath()` and `file_exists()` check and throws exception.

ADmad:
Static analysis build failures have been fixed on `4.x`.

ravage84 (commented):
I'm OK with using ``SplFileInfo`` instead of our old ``File`` but I feel it's a step beckward if we replace our abstraction with the vanilla PHP functions. Can't we use at least some kind of abstraction?

Personally, I would love to see us using either [Flysystem](http://flysystem.thephpleague.com/docs/) or [Gaufrette](https://github.com/KnpLabs/Gaufrette) since the filesystem is often one of the hard dependencies besides date/time one can rarely get rid of if not considered from ground up.

src/Routing/Middleware/AssetMiddleware.php
ADmad: @ravage84 The getMimeType() method used on response instance in deliverAsset() method does not exist on the interface.

src/Shell/Task/LoadTask.php
```diff
@@ -89,8 +88,7 @@ protected function makeOptions(): string
      */
     protected function modifyApplication(string $app, string $plugin, string $options): void
     {
-        $file = new File($app, false);
-        $contents = $file->read();
+        $contents = file_get_contents($app);
```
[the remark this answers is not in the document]
ADmad: Why do you need an abstraction for something that can be achieved with a single PHP function call?

src/Shell/Task/ExtractTask.php
```diff
@@ -765,8 +767,14 @@ protected function _searchFiles(): void
...
         foreach ($this->_paths as $path) {
             $path = realpath($path) . DIRECTORY_SEPARATOR;
-            $Folder = new Folder($path);
-            $files = $Folder->findRecursive('.*\.(php|ctp|thtml|inc|tpl)', true);
+
+            $finder = new Finder();
+            $finder->files()
+                ->in($path)
+                ->sortByName()
+                ->name('/\.(php|ctp)$/');
```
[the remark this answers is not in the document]
ADmad: I must have probably made this commit before my PR to change template extensions was merged. As for moving this list of a property is beyond the scope of this PR.

ADmad:
@markstory With regards to your concern about adding additional external dependencies, I looked into replacing `File` and `Folder` usage with just core PHP code but that leads to lot of code repetition and hence not practical. 

I can instead modify `File` and `Folder` to behave more sanely/intuitively, but those would be non BC changes. Would that be okay?

inoas:
@ADmad I do value BC a lot. Why make 4.x more hassle that it needs to be?
You can then copy the bits and bolts of Folder and File that are mandatory for the core, modify for sanity into new classes and possibly add deprecation warnings that target removal of the old File and Folder in 5.0?

ADmad:
> I can instead modify File and Folder to behave more sanely/intuitively, but those would be non BC changes.

I gave that a shot but it would entail pretty much a rewrite of those classes and I didn't won't to invest time and effort into that.

That two related PRs I made take care of removing their usage by just using core PHP and an internal class with utility methods instead of external dependencies. Hence closing this.

PR #12571, opened 2018-09-18
[RFC] Make RouteBuilder::redirect() return route instances

Iandenh:
Following #10738, I think it's handy to make the redirect also fluent so we can do something like this:
`$routes->redirect('/old', '/forums')->setStatus(302)`

The only thing is that `setPersist` in `Route` only accepts an array, but in `redirect` a boolean is also accepted. Don't know what the best way to solve this? One way is to change the function in `Route` to also accepts `boolean`.

### DIFF
--- a/tests/TestCase/Routing/Route/RedirectRouteTest.php
+++ b/tests/TestCase/Routing/Route/RedirectRouteTest.php
@@ -236,4 +236,17 @@ class RedirectRouteTest extends TestCase
         $route = new RedirectRoute('/:lang/my_controllers', ['controller' => 'tags', 'action' => 'add'], ['lang' => '(nl|en)', 'persist' => ['lang']]);
         $route->parse('/nl/my_controllers/');
     }
+
+    /**
+     * Test setting HTTP status
+     *
+     * @return void
+     */
+    public function testSetStatus()
+    {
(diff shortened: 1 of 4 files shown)


markstory (commented):
I like it :+1:

tests/TestCase/Routing/Route/RedirectRouteTest.php
```diff
@@ -236,4 +236,17 @@ public function testParsePersistMatchesAnotherRoute()
...
+    /**
+     * Test setting HTTP status
+     *
+     * @return void
+     */
+    public function testSetStatus()
+    {
+        $route = new RedirectRoute('/home', ['controller' => 'posts']);
+        $result = $route->setStatus(302);
+        $this->assertSame($result, $route, 'Should return this');
```
ravage84: "Should return a 302 status code" or no message.

ravage84 (changes requested):

tests/TestCase/Routing/Route/RedirectRouteTest.php
Iandenh: Thanks, I removed the messages. Because it checks if the whole Route object is the same.

PR #12573, opened 2018-09-19
Update templates path and extension in docblocks.

ADmad:


ADmad:
I am wondering if `templates/Layout` should be renamed to `templates/layout` and `templates/Element` to `templates/element`. Having these two special folder lowercased makes them easier to spot.

markstory:
Do the missing template and missing layout error pages need to be updated at all?

ADmad:
They were already updated in previous PR.

PR #12576, opened 2018-09-20
Add getter/setter for `Association::className()`.

ndm2:
The registry alias juggling in `Association::getTarget()` confused me a little bit, hope I got it right with testing against the actual class name in the setter.

refs #9978


markstory (approved):
Looks reasonable to me.

PR #12579, opened 2018-09-21
Augment docblock for `Hash::combine()`

josephzidell:
Adds array as possible types into the docblock for `Hash::combine()` for `$keyPath` and `$valuePath` params.


markstory:
Thank you 👍

Issue #12585, opened 2018-09-23
Extensions being ignored/silenced through RequestHandler


dereuromark:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: dev-master

### What you did
/event.ical

This, without whitelisting the action, used to just try to access action_ical etc and as such failed with 404.

### What happened
Now, this is just being ignored and the same /event (html) action is served as without _ext definition.
So basically, it now silently allows *any* extension it does not recognize.

The following line in RequestHandler component seems to do this silencing:
```php
        $isRecognized = (
            !in_array($this->ext, ['html', 'htm']) &&
            $response->getMimeType($this->ext)
        );

        if ($this->ext && $isRecognized) {
```

The RequestHandler asks the response class which doesnt have the ical definition and as such a silent fallback to html. That also creates duplicate URLs, by the way, which is bad for multiple other reasons.

### What you expected to happen

In my case I want to serve this action via Ical view class and let the user download this event as ICAL file.

So thus the `Router::extensions(['json', 'pdf', 'ical']);` config in routes.php and
```
//EventsController::initialize()
$this->RequestHandler->setConfig('viewClassMap', ['ical' => 'Ical']);
```

So 
a) The silent fallback should be turned into a speaking 404 exception (currently silent with both Router::extensions set or not set by the way)
b) The response class must have a setter (type() one did that but is deprecated) again for adding response type definitions.

dereuromark:
Note that switching to the known type "ics" made part B work again (silencing problem of part A is also here a problem), but we must also fix this for possible new/unknown/custom extensions, thus my two resolutions here we need to fix up.

Libertie:
I believe that #12586 breaks html extension routing. The following new code always returns a false value for $isRecognized when routing with "html" or "htm" extension:

```
$isRecognized = (
   !in_array($this->ext, ['html', 'htm']) &&
   $response->getMimeType($this->ext)
);
```

And this causes the NotFoundException to appear, which is not desirable.

markstory:
@Libertie could you open a new issue with context on how to reproduce what broke in your application?

markstory:
@dereuromark Can this be closed now?

dereuromark:
I wrote in the PR:

> This fixes most of the issues from #12585
It does however not fix the silencing of non configured extensions.
[...]

PR #12586, opened 2018-09-23
Fix silencing errors in extension handling.

dereuromark:
This fixes most of the issues from https://github.com/cakephp/cakephp/issues/12585

It does however not fix the silencing of non configured extensions.
So if you didnt whitelist `ical` and then just accessed `/controller/action.ical` it would just silently serve the content of `/controller/action` without any indication of problems. One would except a 404 here, though.
The reason for this last issue seems to be that in such a case routing just ignores the ext, and sets it to null for going into the controller. That's bad.

### DIFF
--- a/src/Controller/Component/RequestHandlerComponent.php
+++ b/src/Controller/Component/RequestHandlerComponent.php
@@ -321,6 +321,7 @@ class RequestHandlerComponent extends Component
      *
      * @param \Cake\Event\Event $event The Controller.beforeRender event.
      * @return bool false if the render process should be aborted
+     * @throws \Cake\Http\Exception\NotFoundException If invoked extension is not configured.
      */
     public function beforeRender(Event $event)
     {
@@ -334,7 +335,7 @@ class RequestHandlerComponent extends Component
             $response->getMimeType($this->ext)
         );
         if ($this->ext && !$isRecognized) {
-            throw new NotFoundException('Invoked extension not recognized: ' . $this->ext);
+            throw new NotFoundException('Invoked extension not recognized/configured: ' . $this->ext);
         }
 
         if ($this->ext) {
--- a/src/Controller/Component/RequestHandlerComponent.php
+++ b/src/Controller/Component/RequestHandlerComponent.php
@@ -598,9 +598,6 @@ class RequestHandlerComponent extends Component
         $viewClass = null;
         if ($builder->getClassName() === null) {
             $viewClass = App::className($view, 'View', 'View');
-            if ($viewClass === false) {
-                throw new RuntimeException('Configured view class can not be found: ' . $view);
-            }
         }
 
         if ($viewClass) {
--- a/src/Controller/Component/RequestHandlerComponent.php
+++ b/src/Controller/Component/RequestHandlerComponent.php
@@ -21,6 +21,7 @@ use Cake\Core\App;
 use Cake\Core\Configure;
 use Cake\Core\Exception\Exception;
 use Cake\Event\Event;
+use Cake\Http\Exception\NotFoundException;
 use Cake\Http\Response;
 use Cake\Routing\Router;
 use Cake\Utility\Exception\XmlException;
@@ -332,8 +333,11 @@ class RequestHandlerComponent extends Component
             !in_array($this->ext, ['html', 'htm']) &&
             $response->getMimeType($this->ext)
         );
+        if ($this->ext && !$isRecognized) {
+            throw new NotFoundException('Invoked extension not recognized: ' . $this->ext);
+        }
 
-        if ($this->ext && $isRecognized) {
+        if ($this->ext) {
             $this->renderAs($controller, $this->ext);
             $response = $controller->response;
         } else {
@@ -594,6 +598,9 @@ class RequestHandlerComponent extends Component
         $viewClass = null;
         if ($builder->getClassName() === null) {
             $viewClass = App::className($view, 'View', 'View');
+            if ($viewClass === false) {
+                throw new RuntimeException('Configured view class can not be found: ' . $view);
+            }
         }
 
         if ($viewClass) {
--- a/src/Http/Response.php
+++ b/src/Http/Response.php
@@ -1111,7 +1111,7 @@ class Response implements ResponseInterface
     {
         deprecationWarning(
             'Response::type() is deprecated. ' .
-            'Use getType() or withType() instead.'
+            'Use getType(), setType() or withType() instead.'
         );
 
         if ($contentType === null) {
@@ -1135,6 +1135,22 @@ class Response implements ResponseInterface
         $this->_setContentType();
 
         return $contentType;
+    }
+
+    /**
+     * Sets a content type into the map.
+     *
+     * E.g.: setType('xhtml' => ['application/xhtml+xml', 'application/xhtml', 'text/xhtml'])
+     *
+     * This is needed for RequestHandlerComponent and recognition of types.
+     *
+     * @param string $type
+     * @param string|array $definition
+     * @return void
+     */
+    public function setType($type, $definition)
+    {
+        $this->_mimeTypes[$type] = $definition;
     }
 
     /**
--- a/src/Http/Response.php
+++ b/src/Http/Response.php
@@ -1145,12 +1145,12 @@ class Response implements ResponseInterface
      * This is needed for RequestHandlerComponent and recognition of types.
      *
      * @param string $type Content type.
-     * @param string|array $definition Definition of the content type.
+     * @param string|array $mimeType Definition of the mime type.
      * @return void
      */
-    public function setType($type, $definition)
+    public function setType($type, $mimeType)
     {
-        $this->_mimeTypes[$type] = $definition;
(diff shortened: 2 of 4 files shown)


src/Http/Response.php
```diff
@@ -1137,6 +1137,22 @@ public function type($contentType = null)
...
+     *
+     * E.g.: setType('xhtml' => ['application/xhtml+xml', 'application/xhtml', 'text/xhtml'])
+     *
+     * This is needed for RequestHandlerComponent and recognition of types.
+     *
+     * @param string $type Content type.
+     * @param string|array $definition Definition of the content type.
+     * @return void
+     */
+    public function setType($type, $definition)
```
ADmad: Why do we need `setType()` when there is `withType()` ?

dereuromark: See the ticket referenced. The type() deprecation lost a way to set new type definitions.
As such RequestHandler would just not be able to serve those anymore without calling

    $this->response->setType('ical', 'my/ical');

etc

ADmad: Okay, in that case `addType()` would be a more intuitive name IMO.

src/Http/Response.php
```diff
@@ -1137,6 +1137,22 @@ public function type($contentType = null)
...
 
+    /**
+     * Sets a content type into the map.
+     *
+     * E.g.: setType('xhtml' => ['application/xhtml+xml', 'application/xhtml', 'text/xhtml'])
+     *
+     * This is needed for RequestHandlerComponent and recognition of types.
+     *
+     * @param string $type Content type.
+     * @param string|array $definition Definition of the content type.
```
ADmad: `$mimeType` would be a more intuitive variable name.

src/Http/Response.php
ADmad: I don't think the fact that `addType()` can overwrite existing mapping will cause confusion to anymore. Someone requiring to overwrite existing type mapping is unlikely anyway. `mapType()` would have been the ideal name but method with that name already exists for another purpose :slightly_frowning_face:.

ADmad: How about the name `setTypeMap()` ?

dereuromark: Modified, also added tests for it.

markstory:
Always raising a 404 on unknown 'extensions' also means that an application can never have . in the last path segment as any value after the . would be treated as an extension. I suspect this is the reason for the current behavior.

dereuromark:
The existing tests also fail, how would we adjust those here?
I think they operate on a false assumption, as reality makes those behave differently.

src/Controller/Component/RequestHandlerComponent.php
```diff
@@ -594,6 +598,9 @@ public function renderAs(Controller $controller, $type, array $options = [])
         $viewClass = null;
         if ($builder->getClassName() === null) {
             $viewClass = App::className($view, 'View', 'View');
+            if ($viewClass === false) {
+                throw new RuntimeException('Configured view class can not be found: ' . $view);
+            }
```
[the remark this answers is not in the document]
ADmad: The tests failures will go away if this particular change is reverted.

ADmad:
> If `foo` is not a declared extension then I would expect `.foo` to be part of the last parameter. If that last parameter is the action then an error should be raised.

`RequestHandler::$ext` would be set to `foo` only if extension parsing has been enabled for `foo` isn't it? So I don't think the check and exception throwing added at line 336-338 should cause any problem.

src/Controller/Component/RequestHandlerComponent.php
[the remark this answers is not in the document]
dereuromark: Should we only fail if you actually configure to a specific className? I feel like silently doing nothing here prevents this developer mistake from being visible.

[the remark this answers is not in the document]
ADmad: Not sure how will you decide if it's a dev error. As @markstory one can just use `View` class to render content of any string type by relying on templates in sub directories and just set appropriate content/mime type for response instance.

dereuromark:
Is it adjusted as per your feedback now?

dereuromark:
Added a test case and also removed the deprecated initialize() call (as it didnt do anything anyway).

tests/TestCase/Controller/Component/RequestHandlerComponentTest.php
```diff
@@ -531,6 +520,26 @@ public function testNoViewClassExtension()
...
+     * Tests that configured extensions that have no configured mimetype do not silently fallback to HTML.
+     *
+     * @return void
+     * @expectedException \Cake\Http\Exception\NotFoundException
+     * @expectedExceptionMessage Invoked extension not recognized/configured: foo
+     */
+    public function testUnrecognizedExtensionFailure()
+    {
+        Router::extensions(['json', 'foo'], false);
+        $this->Controller->request = $this->Controller->request->withParam('_ext', 'foo');
```
ADmad: Using setRequest()/getRequest() instead of the $request property will save use effort when doing 3.next/4.x merges :)

dereuromark: You want me to touch the code again? :)

dereuromark:
All done.

src/Http/Response.php
```diff
@@ -1137,6 +1137,22 @@ public function type($contentType = null)
...
+     *
+     * E.g.: setType('xhtml' => ['application/xhtml+xml', 'application/xhtml'])
+     *
+     * This is needed for RequestHandlerComponent and recognition of types.
+     *
+     * @param string $type Content type.
+     * @param string|array $mimeType Definition of the mime type.
+     * @return void
+     */
+    public function setTypeMap($type, $mimeType)
```
markstory: Is this method setting a map or just a single type? The name and implementation don't appear to match.

dereuromark: A single one. Is setTypeToMap() better?
I didnt want to addType() as it would also overwrite existing one.

markstory: Hmm. On second thought I don't have a better name that isn't long.

garas: `setTypeMap()` suggests that it sets whole map of types, not single type.

Main keyword here is _map_. After look at [dictionary](https://en.oxforddictionaries.com/definition/map), I think it could be `mapType()`, `associateType()`, `linkType()`.

src/Http/Response.php
```diff
@@ -1137,6 +1137,22 @@ public function type($contentType = null)
         return $contentType;
     }
 
+    /**
+     * Sets a content type definition into the map.
+     *
+     * E.g.: setType('xhtml' => ['application/xhtml+xml', 'application/xhtml'])
```
garas: Invalid syntax. Uses wrong function name.

dereuromark: Fixed

dereuromark:
Are you ok with this as is? Or do I have to use a different method name like updateTypeMap() etc?

ADmad (approved):

ravage84:
Out of curiosity I setup a vanilla CakePHP and and tested some cases.


CakePHP Core 3.6.12
CakePHP App 3.6.5

Implemented ``TestController`` with ``event()`` and ``anotherIcal()`` , added the templates ``/Test/event.ctp`` & ``/Test/another_ical.ctp`` and added a route for ``/test/another.ical`` to ``TestController::anotherIcal()`` but used the fallback routes for everyhting else.

URL | RequestHandler active | Extensions active | Expected | Result
---- | ----------------------- | ----------------- | --------- | ------
/test/event | Yes | - | TestController::event() | As expected
/test/event.ical | Yes | - | Not sure | **The action event.ical is not defined in TestController** *2
/test/another.ical | Yes | - | TestController::anotherIcal() | As expected
/test/event | Yes | ical | TestController::event() | As expected
/test/event.ical | Yes | ical | TestController::event() | As expected
/test/another.ical | Yes | ical | TestController::anotherIcal() | **The action another is not defined in TestController** *1
/test/event | No | - | TestController::event() | As expected
/test/event.ical | No | - | Not sure | **The action event.ical is not defined in TestController** *2
/test/another.ical | No | - | TestController::anotherIcal() | As expected
/test/event | No | ical | TestController::event() | As expected
/test/event.ical | No | ical | TestController::event() | As expected
/test/another.ical | No | ical | TestController::anotherIcal() | **The action another is not defined in TestController** *1


1. Routing should  have precedence over extension handling
2. Action names with a dot, like ``event.ical``, are invalid anyway, we could raise  a not found exception with a more helpful errro message like ``The extension ".ical" is not handled nor is there a route configured.``.

src/Http/Response.php
ravage84: I also don't think ``setTypeMap()`` for setting one tpye within the map is a good name.
The best would really be ``mapType()``, unfortunately.

How about ``setTypeMapped()`` or ``setMappedType()``?
Or ``updateType()``?

``associateType()``, isn't bad either.

dereuromark:
Can u open that up as a separate ticket. This should not get pickybacked on this one.

ravage84:
@dereuromark thought it was relevant. Anyway, done so in #12624

PR #12587, opened 2018-09-23
4.x psalm

ADmad:


src/View/Form/ContextInterface.php
```diff
@@ -55,10 +55,11 @@ public function isCreate(): bool;
      *      context's schema should be used if it's not explicitly provided.
      *
      * @param string $field A dot separated path to the field a value
+     * @param array Options.
```
dereuromark: $options

src/View/ViewBuilder.php
```diff
@@ -52,14 +52,14 @@ class ViewBuilder implements JsonSerializable, Serializable
...
-     * @var string|null|false
+     * @var string|null
      */
     protected $_plugin;
 
     /**
      * The theme name to use.
      *
-     * @var string|null|false
+     * @var string|null
```
dereuromark: sure?
also, below _layout can also be null (or false if no layout should be used), right?

dereuromark: I am fixing the wrong layout one in https://github.com/cakephp/cakephp/pull/12588 then in master.

PR #12588, opened 2018-09-23
Fix doc blocks.

dereuromark:
Inspired by https://github.com/cakephp/cakephp/pull/12587/

Not being able to properly init the properties, makes also the expected result of isAutoLayoutEnabled() rather problematic.


markstory:
Test failures seem relevant.

PR #12589, opened 2018-09-23
Remove duplicate test code.

ADmad:


tests/TestCase/Database/QueryTest.php
```diff
@@ -2227,21 +2227,6 @@ public function testSelectDistinctON()
             [3, 1],
             collection($results)->sortBy('author_id')->extract('author_id')->toList()
         );
-
-        $query = new Query($this->connection);
```
ADmad: The exact same query instance and assertions are done above. So unless I am missing something the removed code was redundant.

saeideng (changes requested):
its not same 
see `->distinct('author_id')`  and `->distinct(['author_id'])`

dereuromark:
An inline comment might have clarified this.

markstory:
Closing for the reason @saeideng mentioned.

Issue #12590, opened 2018-09-24
fatal error due to not existing components


dereuromark:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.6.11 (dev-master)

### What you did
Setting a redirect in htaccess for a URL/domain change.
Suddenly we got 5xx exceptions with very weird (and hard to debug) behavior.

### What happened
```
[Error] Call to a member function allow() on boolean
Request URL: /hybrid-auth/endpoint?redirect=%2Faccount%2Flogin&hauth.start=Github&hauth.time=1537783531
Referer URL: /account/login
Stack Trace:
#0 /app/vendor/cakephp/cakephp/src/Event/EventManager.php(353): ADmad\HybridAuth\Controller\HybridAuthController->beforeFilter(Object(Cake\Event\Event))
#1 /app/vendor/cakephp/cakephp/src/Event/EventManager.php(330): Cake\Event\EventManager->_callListener(Array, Object(Cake\Event\Event))
#2 /app/vendor/cakephp/cakephp/src/Event/EventDispatcherTrait.php(114): Cake\Event\EventManager->dispatch(Object(Cake\Event\Event))
#3 /app/vendor/cakephp/cakephp/src/Controller/Controller.php(658): Cake\Controller\Controller->dispatchEvent('Controller.init...')
#4 /app/vendor/cakephp/cakephp/src/Http/ActionDispatcher.php(115): Cake\Controller\Controller->startupProcess()
#5 /app/vendor/cakephp/cakephp/src/Http/ActionDispatcher.php(94): Cake\Http\ActionDispatcher->_invoke(Object(ADmad\HybridAuth\Controller\HybridAuthController))
#6 /app/vendor/cakephp/cakephp/src/Http/BaseApplication.php(235): Cake\Http\ActionDispatcher->dispatch(Object(Cake\Http\ServerRequest), Object(Cake\Http\Response))
#7 /app/vendor/cakephp/cakephp/src/Http/Runner.php(65): Cake\Http\BaseApplication->__invoke(Object(Cake\Http\ServerRequest), Object(Cake\Http\Response), 
```

or also

```php
if ($this->Auth->user('id')) {
}
```
inside Controller method login() suddenly throws this exception.

How come the Auth component as very internal and core one can not be available in a controllers execution cycle? This should never happen.

### What you expected to happen
[...]

PR #12593, opened 2018-09-24
Show notice for non existent controller property.

ADmad:
Refs https://github.com/cakephp/cakephp/issues/12590#issuecomment-423932380

### DIFF
--- a/src/Controller/Controller.php
+++ b/src/Controller/Controller.php
@@ -370,24 +370,24 @@ class Controller implements EventListenerInterface, EventDispatcherInterface
         }
 
         list($plugin, $class) = pluginSplit($this->modelClass, true);
-        if ($class !== $name) {
-            $trace = debug_backtrace();
-            $parts = explode('\\', get_class($this));
-            trigger_error(
-                sprintf(
-                    'Undefined property: %s::$%s in %s on line %s',
-                    array_pop($parts),
-                    $name,
-                    $trace[0]['file'],
-                    $trace[0]['line']
-                ),
-                E_USER_NOTICE
-            );
-
-            return false;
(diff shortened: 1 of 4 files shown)


src/Controller/Controller.php
```diff
@@ -371,6 +371,19 @@ public function __get($name)
 
         list($plugin, $class) = pluginSplit($this->modelClass, true);
         if ($class !== $name) {
```
garas: Maybe reverse `if` and explain:

```
if ($class === $name) {
   // Autoload table if $name matches controller name
   return $this->loadModel($plugin . $class);
}

// notice

return false; // remove or explicitly return null at 4.x
```

markstory:
Looks like we have a few undefined properties being used by our tests :cry:

ADmad:
Not really, the error doesn't make sense. It's complaining about missing `Controller::$viewClass` but that property is available through the `ViewVarsTrait`.

garas:
`Controller::$viewClass` is defined in `ViewVarsTrait`, but it is `unset()` after first use.

> https://github.com/cakephp/cakephp/pull/12035 I've added an unset() for viewClass to avoid a permanent deprecation warning when RequestHandlerComponent is used.

https://github.com/cakephp/cakephp/blob/13ab4d788ce553056938b60b3039197b8edd9993/src/View/ViewVarsTrait.php#L75-L78

If second `render()` tries access value, it is no more available. It should probably leave property set.

ADmad:
`PHPUnit\Framework\Error\Notice` is not available in phpunit 5.7 :slightly_frowning_face:

ADmad:
Meh, took care of it with a `class_exists()` check.

PR #12594, opened 2018-09-24
Check disable_functions before set_time_limit

dhemeier:
On many shared hosting providers you probably can't use of `set_time_limit()`. 

When you now try to Response with a File you probably don't see anything in case of display_errors=1 and Configure::write('debug') > 0.

File Response breaks with trying to display the warning message: `set_time_limit() has been disabled for security reasons`. The implementation is based on some other frameworks to avoid this behaviour and checks first, that you can safely use this function.

### DIFF
--- a/src/Http/Response.php
+++ b/src/Http/Response.php
@@ -2708,7 +2708,7 @@ class Response implements ResponseInterface
         }
 
         $bufferSize = 8192;
-        if (strpos(ini_get('disable_functions'),'set_time_limit') === false) {
+        if (strpos(ini_get('disable_functions'), 'set_time_limit') === false) {
             set_time_limit(0);
         }
         session_write_close();
--- a/src/Http/Response.php
+++ b/src/Http/Response.php
@@ -2708,7 +2708,7 @@ class Response implements ResponseInterface
         }
 
         $bufferSize = 8192;
-        if (array_search('set_time_limit', explode(',', ini_get('disable_functions'))) === false) {
+        if (strpos(ini_get('disable_functions'),'set_time_limit') === false) {
             set_time_limit(0);
         }
         session_write_close();
--- a/src/Http/Response.php
+++ b/src/Http/Response.php
@@ -2708,7 +2708,9 @@ class Response implements ResponseInterface
         }
 
         $bufferSize = 8192;
-        set_time_limit(0);
+        if (array_search('set_time_limit', explode(',', ini_get('disable_functions'))) === false) {
+            set_time_limit(0);
+        }
         session_write_close();
         while (!feof($file->handle)) {
             if (!$this->_isActive()) {


ADmad:
Can't we just add `@` in front of function to suppress error reporting instead of doing an expensive explode?

dmromanov:
Silencing `set_time_limit` might lead into introducing a hard-to-find bug with corrupted file downloads when the function is disabled.
It would be nice to have a message saved into logs regarding this behavior.

dhemeier:
@ADmad silencing is not a good idea as @dmromanov described. But the log message is a nice idea. What kind of message do you prefer?

I would say something like this:
```
 if (array_search('set_time_limit', explode(',', ini_get('disable_functions'))) === false) {
    set_time_limit(0);
} else {
    Log::warning("Could not use set_time_limit, function is disabled by disable_functions");
}
```

Feel free to suggest another message. Don't know how exactly the message should be written in this case.

dereuromark:
Should we trigger_error() E_USER_NOTICE or E_USER_WARNING instead?
The logging otherwise might need a **scope**, otherwise, the log files for those shared server instances would fill up like crazy and you would have to disable all warning logs completely, but you might only want to silence this specific (and then known) issue on your server.

dhemeier:
The message are already stored in error.log with the current behaviour. Entry in 2.X CakePHP:

```
2018-09-24 16:56:17 Warning: Warning (2): set_time_limit() has been disabled for security reasons in [XXXXXXX/app/Vendor/cakephp/cakephp/lib/Cake/Network/CakeResponse.php, line 1475]
Trace:
ErrorHandler::handleError() - APP/Vendor/cakephp/cakephp/lib/Cake/Error/ErrorHandler.php, line 230
set_time_limit - [internal], line ??
CakeResponse::_sendFile() - APP/Vendor/cakephp/cakephp/lib/Cake/Network/CakeResponse.php, line 1475
CakeResponse::send() - APP/Vendor/cakephp/cakephp/lib/Cake/Network/CakeResponse.php, line 440
Dispatcher::dispatch() - APP/Vendor/cakephp/cakephp/lib/Cake/Routing/Dispatcher.php, line 174
[main] - APP/webroot/index.php, line 110
```

The "special" problem is, that you don't see any downloaded file and no messages if you set debug > 0 to display them (on the same time you try to open the file, headers are already set to the corresponding downloaded file)

In my personal opinion I would say, that the set_time_limit() call in most cases don't have any impact on the file responses. For large file-downloads you need to modify more than this to make them possible.

markstory:
Do you use shared hosting for your development environment? Or is a staging environment?

dhemeier:
@markstory This is the development environment on a shared hosting. To prevent any errors we develop with debug mode 2 to display all warnings and errors. But without this change (I made it already in our files) you could not test any file downloads in the dev-system.

My previous "workaround" to test file downloads was to set temporarily debug=0.

src/Http/Response.php
```diff
@@ -2708,7 +2708,9 @@ protected function _sendFile($file, $range)
         }
 
         $bufferSize = 8192;
-        set_time_limit(0);
+        if (array_search('set_time_limit', explode(',', ini_get('disable_functions'))) === false) {
```
ADmad: Can't a string search be done for "set_time_limit" using `strpos` instead of using `explode`?  I don't see it yielding a false positive.

dhemeier: Yes, this would also be possible, I change it later. `disable_functions` is a comma separated list....

ADmad (approved):

PR #12597, opened 2018-09-25
Deprecate RequestHandler's wrapper methods.

ADmad:
Also mark RequestHandler::$ext to be made protected in 4.0.


markstory:
Seems good to me.

ravage84 (approved):
First I thought why removing these convinient functions. Then I looked them up. The ``is$Type()`` methods check what the client prefers, what it accepts, not what it sends. Thus naming it ``is$Type`` is somewhat misleading.
That's why I'm fine with removing them.

PR #12599, opened 2018-09-26
Return null instead of false when unable to load model.

ADmad:


dereuromark (approved):

src/Controller/Controller.php
```diff
@@ -304,7 +304,7 @@ public function __get(string $name)
             E_USER_NOTICE
         );
 
-        return false;
+        return null;
```
ravage84: Can't we throw a new specific exception in that case, instead of triggering an error?

dereuromark: As long as we dont have a has() method for it, I dont think we can.

ravage84: Why do you think we would need a ``has()`` method before we can throw an exception in this failure case?

dereuromark: Fair enough. Not sure about all use cases, but I guess being a bit stricter doesn't hurt after all.

ravage84: Yeah, I understand where you wanted to point at, but I tried to think of a scenario where it was better to not throw an exception but just triggering an error... Didn't find one. But when a model couldn't be loaded, it's likely that sh*t hits the fan, sooner or later ...

Also, this is for 4.x where we can easily introduce something like this.

ADmad: There's no need for an exception. Showing notice when trying to use non-existent property is default PHP behavior and same has been done. `loadModel()` already throws an exception it's unable to load a model.

garas: I would even remove `__get()` and require developer to explicitly load controller’s default model.

cleptric: I don’t like this idea. Explicity moves us further away from our “rapid development” approach without a clear benefit.

saeideng (approved):

Issue #12596, opened 2018-09-25
3.7/4.x Cake\Form\Form class misses 4.x getters/setters + deprecation


inoas:
This is a (multiple allowed):

* [x] bug
* [x] enhancement

* CakePHP Version: 3.6.x

### What you did
tried to call `\Cake\Form\Form::getErrors()`

inoas:
I am using that - however it is unexpeted/inconsistent with how Entity::getErrors() works.

ADmad:
We could add an alias `Form::getErrors()` for `Form::errors()` in 3.7 and remove `Form::errors()` in 4.0 if there's a consensus for it.

markstory:
Do we need to remove the old method?  For folks who are using the current method it is change for the sake of change.

inoas:
👍 for an alias

inoas:
PR is up.

From the same class:
```
    public function validator(Validator $validator = null)
    {
        deprecationWarning(
            'Form::validator() is deprecated. ' .
            'Use Form::getValidator()/setValidator() instead.'
        );
```

TLDR:
@markstory well I am with you, but what about consistency and purity/non-duplication?

As I said month ago, in general I don't like having things forced into getters+setters but I'd have categorized classes into builders, readers and regular classes, builders dropping their set prefix (but for fetching things having a get prefix), readers dropping their get prefix (but for writing things in case, using a set prefix) and regular classes being the once having set/get prefix everywhere like all our classes (or most) aim to.

PR #12601, opened 2018-09-26
deprecate \Cake\Form\Form\errors();

inoas:
Ref https://github.com/cakephp/cakephp/issues/12596

### DIFF
--- a/src/Form/Form.php
+++ b/src/Form/Form.php
@@ -251,8 +251,27 @@ class Form implements EventListenerInterface, EventDispatcherInterface, Validato
      * to `validate()` or `execute()`.
      *
      * @return array Last set validation errors.
+     * @deprecated 3.7.0 Use Form::getErrors() instead.
      */
     public function errors()
+    {
+        deprecationWarning(
+            'Form::errors() is deprecated. ' .
+            'Use Form::getErrors() instead.'
+        );
+
+        return $this->getErrors();
+    }
+
+    /**
+     * Get the errors in the form
+     *
+     * Will return the errors from the last call
+     * to `validate()` or `execute()`.
+     *
+     * @return array Last set validation errors.
+     */
+    public function getErrors()
     {
         return $this->_errors;
     }
@@ -353,7 +372,7 @@ class Form implements EventListenerInterface, EventDispatcherInterface, Validato
     {
         $special = [
             '_schema' => $this->schema()->__debugInfo(),
-            '_errors' => $this->errors(),
+            '_errors' => $this->getErrors(),
             '_validator' => $this->getValidator()->__debugInfo()
         ];
 
--- a/tests/TestCase/Form/FormTest.php
+++ b/tests/TestCase/Form/FormTest.php
@@ -148,12 +148,11 @@ class FormTest extends TestCase
      * Test the errors methods.
      *
      * @return void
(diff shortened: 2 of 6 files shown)


dereuromark:
Afaik the consensus was to focus on the actual functional splitting, and keeping the aliases of existing ones simple, so errors() instead of getErrors() - for the sake of changing it only this would introduce hundreds of more deprecations in the code, wouldnt it?
I am not opposed to it, especially when tools like rector could help to safely migrate.
But also not sure if we have to do that all for 4.0.

inoas:
Yes, I wondered why there was not some shared interface.

src/Form/Form.php
```diff
@@ -244,6 +244,24 @@ public function validate(array $data)
      * @return array Last set validation errors.
```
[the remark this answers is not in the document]
inoas: 3.7.0?

dereuromark (approved):

inoas:
I got problems with CI.
Which branch do I need to base my PR on and against which branch should I PR?

ADmad:
> Which branch do I need to base my PR on and against which branch should I PR?

You need to add your changes on top of `3.next` and make PR against `3.next` (which is already done).

tests/TestCase/ORM/EntityTest.php
```diff
@@ -1183,30 +1183,30 @@ public function testErrors()
     {
         $this->deprecated(function () {
             $entity = new Entity();
-            $this->assertEmpty($entity->errors());
-            $this->assertSame($entity, $entity->errors('foo', 'bar'));
-            $this->assertEquals(['bar'], $entity->errors('foo'));
+            $this->assertEmpty($entity->getErrors());
+            $this->assertSame($entity, $entity->getErrors('foo', 'bar'));
```
ADmad: This should be a setter call `$entity->setErrors('foo', 'bar')`, that's what's causing tests to fail.

tests/TestCase/Form/FormTest.php
```diff
@@ -150,6 +150,18 @@ public function testValidateDeprected()
      * @return void
      */
     public function testErrors()
+    {
+        $this->deprecated(function () {
+            $this->testGetErrors();
```
saeideng: you no need to wrap this method with `deprecated` when it does not use/have deprecated methods

saeideng (changes requested):

inoas:
Looks like the appveyor is unrelated.

ADmad:
The `3.next` branch of manual also needs to be updated.

tests/TestCase/Form/FormTest.php
```diff
@@ -140,16 +140,27 @@ public function testValidateDeprected()
...
 
     /**
      * Test the errors methods.
      *
      * @return void
+     * @deprecated 3.7.0 Use Form::testGetErrors() instead.
      */
     public function testErrors()
+    {
+        $this->testGetErrors();
```
garas: Why deprecate test method?

`testErrors()` should be removed now as there is new `testGetErrors()`

Now `testGetErrors()` runs twice

ADmad: Would you mind fixing it please :)

PR #12604, opened 2018-09-26
Fix incorrect class_exists('PHPUnit_Framework_Test')

### LINKED ISSUE #12603: CakePHP's PHPUnit class aliases don't mix well with the Codeception ones
This is a (multiple allowed):

* [x] bug
* [ ] enhancement

* CakePHP Version: 3.5.11
* PHPUnit Version: 6.5.13
* Codeception Version: 2.3.9

---

When running codeception, integrated via cakephp/codeception, the following error is being thrown:

> Cannot declare interface PHPUnit\Framework\Test, because the name is already in use

The source of it is: https://github.com/cakephp/cakephp/blob/ddbfcdb479cfaa59f6e8c3433bf85aa71bff492d/tests/phpunit_aliases.php#L8

For newer PHPUnit versions, the code in that file normally wouldn't get past the first `if` that checks for the existence of `PHPUnit_Runner_Version`, but Codeception declares a `PHPUnit_Runner_Version` alias, hence the check passes.

One thing I've noticed, is that https://github.com/cakephp/cakephp/blob/ddbfcdb479cfaa59f6e8c3433bf85aa71bff492d/tests/phpunit_aliases.php#L7 should probably use `interface_exists()` for `PHPUnit_Framework_Test`, not `class_exists()`, given that `PHPUnit_Framework_Test` is an interface, so `class_exists()` will always return `false`. This will "fix" this issue, but I'm not sure if it's actually a proper/complete fix, as I'm not sure about possible side effects with older PHPUnit versions.

garas:
Fixes #12603

`PHPUnit_Framework_Test` is interface, so `class_exists('PHPUnit_Framework_Test')` is always `false`.

### DIFF
--- a/tests/phpunit_aliases.php
+++ b/tests/phpunit_aliases.php
@@ -4,7 +4,7 @@ if (class_exists('PHPUnit_Runner_Version')) {
         trigger_error(sprintf('Your PHPUnit Version must be at least 5.7.0 to use CakePHP Testsuite, found %s', \PHPUnit_Runner_Version::id()), E_USER_ERROR);
     }
 
-    if (!interface_exists('PHPUnit_Framework_Test') && class_exists('PHPUnit_Framework_TestCase')) {
+    if (!class_exists('PHPUnit\Runner\Version')) {
         class_alias('PHPUnit_Framework_Test', 'PHPUnit\Framework\Test');
         class_alias('PHPUnit_Framework_AssertionFailedError', 'PHPUnit\Framework\AssertionFailedError');
         class_alias('PHPUnit_Framework_TestSuite', 'PHPUnit\Framework\TestSuite');
--- a/tests/phpunit_aliases.php
+++ b/tests/phpunit_aliases.php
@@ -4,7 +4,7 @@ if (class_exists('PHPUnit_Runner_Version')) {
         trigger_error(sprintf('Your PHPUnit Version must be at least 5.7.0 to use CakePHP Testsuite, found %s', \PHPUnit_Runner_Version::id()), E_USER_ERROR);
     }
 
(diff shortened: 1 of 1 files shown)


ravage84:
Looks OK to me, but some jobs in Travis vail:

````
Fatal error: Declaration of Cake\Test\TestCase\DatabaseSuite::run() must be compatible with PHPUnit_Framework_Test::run(PHPUnit_Framework_TestResult $result = NULL) in /home/travis/build/cakephp/cakephp/tests/TestCase/DatabaseSuite.php on line 25
````

````
1) Cake\Test\TestCase\ORM\QueryRegressionTest::testDeepBelongsToManySubqueryStrategy
--- Expected
+++ Actual
@@ @@
 Array (
-    0 => 'tag1'
-    1 => 'tag3'
+    0 => 'tag3'
+    1 => 'tag1'
````

File: tests/phpunit_aliases.php [hunk 1]
```diff
@@ -4,7 +4,7 @@
         trigger_error(sprintf('Your PHPUnit Version must be at least 5.7.0 to use CakePHP Testsuite, found %s', \PHPUnit_Runner_Version::id()), E_USER_ERROR);
     }
 
-    if (!interface_exists('PHPUnit_Framework_Test') && class_exists('PHPUnit_Framework_TestCase')) {
+    if (!class_exists('PHPUnit\Runner\Version')) {
```
garas: I chose this class check after resolving maze of BC/FC aliases from Codeception [2.3.9](https://github.com/Codeception/Codeception/blob/2.3.9/shim.php#L16-L52) and PHPUnit `ForwardCompatibility` dir at [5.7.14](https://github.com/sebastianbergmann/phpunit/tree/5.7.14/src/ForwardCompatibility) and [5.7.27](https://github.com/sebastianbergmann/phpunit/tree/5.7.27/src/ForwardCompatibility)

PR #12605, opened 2018-09-27
Fixed bug causing message-id to be regenerated

jeremyharris:
This fixes a bug where causing `getHeaders()` more than one time will regenerate the Message-ID (when it's set to `true` aka auto-generate). As a workaround I have been extracting the actual Message-ID from the raw headers returned by `send()`.

### DIFF
--- a/src/Mailer/Email.php
+++ b/src/Mailer/Email.php
@@ -1205,10 +1205,10 @@ class Email implements JsonSerializable, Serializable
         }
         if ($this->_messageId !== false) {
             if ($this->_messageId === true) {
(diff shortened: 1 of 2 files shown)


dereuromark:
Interesting, this broke the plugin tests of Queue ( https://travis-ci.org/dereuromark/cakephp-queue/jobs/437331114#L593 ).
I had to:
https://github.com/dereuromark/cakephp-queue/commit/464ab3dfcd6b762108f88e220af2a916e055edb9

Seems not too big of a deal, but not sure why this would be an issue.

PR #12606, opened 2018-09-27
add zend-diactoros ^2.0.0 support

samsonasik:
zend-diactoros released today https://github.com/zendframework/zend-diactoros/releases/tag/2.0.0 and I think the support for zend-diactoros ^2.0.0 can be added to composer.json.


ADmad:
zend-diactoros requires PHP 7.1+ hence CakePHP 3.x can't use it. 

You are welcome to submit a patch targeting CakePHP 4.x (which will support PHP 7.1) to update to diactoros 2.0 from 1.x.

PR #12616, opened 2018-10-01
Fixes required field div classes

josephzidell:
If you have a field that is manually required (`'required' => true`), but not included in the model validation, the `div` doesn't have the `required` class. This patches that.

### DIFF
--- a/lib/Cake/View/Helper/FormHelper.php
+++ b/lib/Cake/View/Helper/FormHelper.php
@@ -1389,7 +1389,7 @@ class FormHelper extends AppHelper {
 		} elseif (is_array($div)) {
 			$divOptions = array_merge($divOptions, $div);
 		}
-		if ($this->_extractOption('required', $options) !== false &&
+		if ($this->_extractOption('required', $options) !== false ||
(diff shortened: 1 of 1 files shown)


markstory (commented):
Seems like this would break at least 1 existing test. But once again travis is not running on pull requests from you.

lib/Cake/View/Helper/FormHelper.php
```diff
@@ -1389,7 +1389,7 @@ protected function _divOptions($options) {
 		} elseif (is_array($div)) {
 			$divOptions = array_merge($divOptions, $div);
 		}
-		if ($this->_extractOption('required', $options) !== false &&
+		if ($this->_extractOption('required', $options) !== false ||
```
ravage84: From a quick grasp, that makes sense. Also the documetnation states this

> Since 2.3 the HTML5 required attribute will also be added to the input based on validation rules. __You can explicitly set required key in options array to override it for a field.__

https://book.cakephp.org/2.0/en/core-libraries/helpers/form.html#FormHelper::input

ravage84 (commented):
Looks good from what I read in the docs. But test must be made passing.

PR #12625, opened 2018-10-04
Add SimpleCacheEngine::add()

markstory:
I forgot this method last time around. I've also added soft deprecations for Cache::engine(). Having hard deprecations would require more refactoring than I want to do right now.

### DIFF
--- a/src/Cache/CacheEngineInterface.php
+++ b/src/Cache/CacheEngineInterface.php
@@ -26,6 +26,16 @@ namespace Cake\Cache;
 interface CacheEngineInterface
 {
     /**
+     * Write data for key into a cache engine if it doesn't exist already.
+     *
+     * @param string $key Identifier for the data.
+     * @param mixed $value Data to be cached - anything except a resource.
+     * @return bool True if the data was successfully cached, false on failure.
+     *   Or if the key existed already.
+     */
+    public function add($key, $value);
+
+    /**
      * Increment a number under the key and return incremented value
      *
      * @param string $key Identifier for the data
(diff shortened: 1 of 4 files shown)


src/Cache/CacheEngineInterface.php
```diff
@@ -25,6 +25,16 @@
...
 {
+    /**
+     * Write data for key into a cache engine if it doesn't exist already.
+     *
+     * @param string $key Identifier for the data.
+     * @param mixed $value Data to be cached - anything except a resource.
+     * @return bool True if the data was successfully cached, false on failure.
+     *   Or if the key existed already.
+     */
+    public function add($key, $value);
```
ADmad: I wonder how many users actually need this method? For the few who might be actually using it what this method does could be easily simulated through app code. I would prefer removing it in 4.x in which case we shouldn't add it to the interface.

markstory: It is only implemented in a few cache engines (memcached, redis, apc). In other engines it is a non-atomic operation.

ADmad: Oh, I only checked the code of `CacheEngine::add()`, if memcache, redis, apc do provide it as atomic operation then I am fine keeping it.

markstory: Ok. Is there anything else that should change with this?

ADmad: Nope

ADmad (approved):

Issue #12577, opened 2018-09-20
Regression with extensions and routing


dereuromark:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: dev-master (3.x)

### What you did

Updating core:
```
  - Updating zendframework/zend-diactoros (1.8.1 => 1.8.6):  Checking out 20da13beba
  - Updating cakephp/cakephp (3.6.4 => 3.6.11):  Checking out ddbfcdb479
  - Updating friendsofcake/bootstrap-ui (dev-develop b46d01e => dev-master b8e17bb):  Checking out b8e17bbb40
  - Updating cakephp/debug_kit dev-master (0e29ced => d5cd818):  Checking out d5cd81818b
  - Updating dereuromark/cakephp-geo dev-master (f6dae59 => 192517e):  Checking out 192517e70d
  - Updating dereuromark/cakephp-ide-helper dev-master (2731cc6 => a481a29):  Checking out a481a29fe1
  - Updating dereuromark/cakephp-setup dev-master (6336740 => 12adf45):  Checking out 12adf45c4b
  - Updating dereuromark/cakephp-test-helper dev-master (96339fe => 82e714c):  Checking out 82e714c940
  - Updating dereuromark/cakephp-tinyauth dev-master (9d5d22e => 479adb7):  Checking out 479adb7bdb
  - Updating dereuromark/cakephp-tools dev-master (0fce0f3 => 1fac073):  Checking out 1fac073732
  - Updating friendsofcake/search dev-master (7c14e0c => cc2decf):  Checking out cc2decf52c
```

### What happened
Integration tests around `_ext` now fail. See travis: https://travis-ci.org/dereuromark/cakephp-sandbox/jobs/431208574

### What you expected to happen
There should be no regression from 3.6.4 to 3.6.11
It seems it happened in 3.6.5: https://github.com/cakephp/cakephp/compare/3.6.4...3.6.5
Most likely around the legacy router part - and might only be relevant for tests.

But there is also no indication how tests would need to be refactored to work in the patch(es) after 3.6.4.

Tests are using
```php
	/**
	 * @return void
	 */
	public function setUp() {
		parent::setUp();

		Router::extensions(['rss']);
	}
```
which worked fine so far.

markstory:
[...]

PR #12628, opened 2018-10-06
Fix regression in integration testing

dereuromark:
This resolves the regression introduced in 3.6.5 via
https://github.com/cakephp/cakephp/compare/3.6.4...3.6.5#diff-3d58de79d11037069356ea3dd5556605R126

Resolves ticket https://github.com/cakephp/cakephp/issues/12577

Any side effects, any other (better) solutions to fix the issue?

Proof:
- status quo, failing: https://github.com/dereuromark/cakephp-sandbox/pull/12
- with this branch, passing: https://github.com/dereuromark/cakephp-sandbox/pull/13

### DIFF
--- a/src/TestSuite/MiddlewareDispatcher.php
+++ b/src/TestSuite/MiddlewareDispatcher.php
@@ -126,7 +126,6 @@ class MiddlewareDispatcher
         }
 
         $out = Router::url($url);
-        Router::reload();
 
         return $out;
     }


markstory:
Without this `reload()` call routes loaded as part of resolving a testcase URL are retained, and then when `RoutingMiddleware` goes to load routes again during the request applications using named routes will likely hit a duplicate route error.

dakota:
This was part of the fix that allowed our large application to be upgraded to CakePHP 3.6.

dereuromark:
Any ideas then how we can solve the regression in integration test cases without compromising the other use cases?

saeideng:
@dereuromark
what I'm seeing here , this related to this commit https://github.com/cakephp/cakephp/commit/135957df21e54436ece2a068457c9c3ea02458eb#diff-3d58de79d11037069356ea3dd5556605R81
see order of using `Router::reload();` before/after this commit

dereuromark:
I saw the same. I bet there is no test for extension routing. That would probably show the issue.

markstory:
An alternative option could be to have the test harness call (a new not yet implemented method) that just resets the connected routes. Something like:

```php
Router::getRouteCollection()->clear()
```

This would leave extensions and middleware untouched but clear routes preventing duplicate named route errors.

dereuromark:
I am fine either way as long as we find a way quickly to resolve this regression.
Tests all being red for months is not good for CI and continues integration testing.

An alternative could also be to modify

    return new MiddlewareDispatcher($this, $this->_appClass, $this->_appArgs);

call to allow the test harness to configure/flag-switch the above PR change (new optional argument?), this way we could shim it down for the cases needed.
Even though that would more be a quick fix for 3.6 for now, as your suggestion seems to aim more into a long-term solution.

markstory:
Closing in favour of #12695

Issue #12629, opened 2018-10-08
RFC: Use lower case names for special templates folders.


ADmad:
We have few special templates folder like `Element`, `Layout`, `Plugin` and `Cell`. I propose changing them to lower case names: `element`, `layout`, `plugin` and `cell` respectively. 

This helps differentiate these special folder from corresponding controller templates folders and makes them easier to spot in a directory listing, especially when you have lot of controller folders in your project.

dereuromark:
Plugin was always plugins so far (since 3.x).
The others have been inside /src/, and as such were uppercase.
Now, that they moved into /templates/ (and outside psr-4 src) it sure can make sense to rename.

ADmad:
> Plugin was always plugins so far (since 3.x).

I was referring to the templates folder in 3.x `src/Template/Plugin` which is used for overriding plugin templates, not the `plugins` folder at app root which contains full plugins. As per my proposal it would now be `templates/plugin`.

markstory:
Do you think this is a change we could automate easily with rector? If so I'm ok with doing it.

ADmad:
I haven't looked into rector but the proposed folder renames are similar to templates folder rename we have already done, so I don't see these additional changes being an issue.

PR #12634, opened 2018-10-09
4.x - Lower cased special template folders names

ADmad:
Refs #12629


src/View/View.php
```diff
@@ -282,25 +282,25 @@ class View implements EventDispatcherInterface
     public const TYPE_ELEMENT = 'element';
 
     /**
-     * Constant for name of view file 'Element'
+     * Constant for view file type 'layout'
      *
      * @var string
      */
-    public const NAME_ELEMENT = 'Element';
+    public const TYPE_LAYOUT = 'layout';
```
markstory: These constant renames should go in the list of breaking changes.

markstory (commented):
This is going to require changes in Bake as well. It will generate files with the old paths right now.

ADmad:
I will update bake and docs too after this.

PR #12640, opened 2018-10-16
4.x - Cache updates pt 1.

markstory:
This is part one of many, and lots of things are not working yet. My goal with these changes was to get the existing cache engines compatible with PSR16 at an *interface* level. There are lots of implementation gaps that will be addressed in future pull requests. Things I'm aware of and plan on fixing next:

* The `ttl` parameter for set does nothing right now.
* The `_key` method and PSR16 validateKey method are basically the same.
* The various engines don't implement any of the `*Multiple` methods yet.
* `SimpleCacheEngine` needs to be removed.

### DIFF
--- a/src/Cache/Cache.php
+++ b/src/Cache/Cache.php
@@ -317,7 +317,7 @@ class Cache
      *
      * @param string $key Identifier for the data
      * @param string $config optional name of the configuration to use. Defaults to 'default'
-     * @return mixed The cached data, or false if the data doesn't exist, has expired, or if there was an error fetching it
+     * @return mixed The cached data, or null if the data doesn't exist, has expired, or if there was an error fetching it
      */
     public static function read(string $key, string $config = 'default')
     {
--- a/src/Cache/Cache.php
+++ b/src/Cache/Cache.php
@@ -17,7 +17,6 @@ namespace Cake\Cache;
 
 use Cake\Cache\Engine\NullEngine;
 use Cake\Core\StaticConfigTrait;
-use InvalidArgumentException;
 use RuntimeException;
 
 /**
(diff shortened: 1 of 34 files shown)


lorenzo (approved):
This is looking great, good job!

src/Cache/Cache.php
```diff
@@ -383,12 +362,11 @@ public static function readMany(array $keys, string $config = 'default'): array
      */
     public static function increment(string $key, int $offset = 1, string $config = 'default')
     {
-        $engine = static::pool($config);
         if (!is_int($offset) || $offset < 0) {
```
ADmad: The `is_int()` check can now be removed since the argument is typehinted. Also for `$offset < 0` it would be better to throw an `InvalidArgumentException` now.

src/Cache/Cache.php
```diff
@@ -402,12 +380,11 @@ public static function increment(string $key, int $offset = 1, string $config =
      */
     public static function decrement(string $key, int $offset = 1, string $config = 'default')
     {
-        $engine = static::pool($config);
         if (!is_int($offset) || $offset < 0) {
```
ADmad: Same comment as for `increment()`.

src/Cache/Cache.php
```diff
@@ -457,44 +432,40 @@ public static function delete(string $key, string $config = 'default'): bool
...
-     *   will become a no-op value in 4.0 as it is deprecated.
      * @param string $config name of the configuration to use. Defaults to 'default'
      * @return bool True if the cache was successfully cleared, false otherwise
      */
-    public static function clear(bool $check = false, string $config = 'default'): bool
+    public static function clear(string $config = 'default'): bool
     {
-        $engine = static::engine($config);
+        if (func_num_args() > 1) {
+            deprecationWarning('The $check parameter has been removed from clear(). Update your usage.');
```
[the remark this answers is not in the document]
markstory: Good point.

ADmad (approved):

PR #12644, opened 2018-10-16
Remove unused constant.

ADmad:


saeideng:
since we have it in cakephp/app repo 
https://github.com/cakephp/app/blob/42c089536a32e5362e328e76e8ef5348883d5a31/config/paths.php#L18-L20
seems its enough 
maybe it is useless in core itself
but for other plugins its necessary IMO 
like https://github.com/cakephp/bake/blob/7356656e236bb62733f4fdd4894ce1f26f9e14f6/tests/bootstrap.php#L39

dereuromark:
I actually like it and use it a lot in app and plugin code. 
Much better than the long form.

markstory:
@dereuromark it is still defined in the app template.

dereuromark:
That doesnt help plugins to rely on it IMO.

dereuromark:
Baeh, thats not nice. It builds paths on Windows not consistently IMO then.
Does it hurt to keep it?

ADmad:
> Does it hurt to keep it?

Not really. Closing then.

PR #12650, opened 2018-10-18
4.x remove file folder usage

ADmad:
Cherry picked commits from #12569 which remove `File` and `Folder` usage from few classes without the need of external dependency.


lorenzo (approved):
Looks great!

Issue #12652, opened 2018-10-18
[2.x] Error in PostgreSQL describe when the default value is a function 


begnini:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 2.10.12 (commit 5b57a1e9263d288bb47abdf2d092f702a862a084)
* Platform and Target: PostgreSQL

### What you did
Create a table where one field has a function in the default value.

```php
$db->execute('CREATE TABLE test_function_default_describe (
   id integer PRIMARY KEY, 
   year int default date_part(\'year\'::text, now()))');
```

### What happened

When call `$db->describe('test_function_default_describe');` the default value returned is truncated.

```php
Array (
    'type' => 'integer'
    'null' => true
    'default' => 'date_part('year''
    'length' => null
) 
```

So, if you use this table to clone other table (for example, a fixture), the generated SQL comes broken:

```sql
ERROR:  invalid input syntax for integer: "date_part('year'"
STATEMENT:  CREATE TABLE "public"."test_function_default_describe" (
    "id" serial NOT NULL,
    "year" int DEFAULT 'date_part(''year''' NOT NULL
);
```

### What you expected to happen

The expected is the default value describing the function without truncation.

saeideng:
Closing as a pull request is open  now

begnini:
I think the master branch has the same bug. 

https://github.com/cakephp/cakephp/blob/master/src/Database/Schema/PostgresSchema.php#L204

PR #12653, opened 2018-10-18
limiting the regex to consume only words in default value in pgsql

begnini:
Patch to #12652

### DIFF
--- a/lib/Cake/Test/Case/Model/Datasource/Database/PostgresTest.php
+++ b/lib/Cake/Test/Case/Model/Datasource/Database/PostgresTest.php
@@ -1218,4 +1218,24 @@ class PostgresTest extends CakeTestCase {
 
 		$this->assertEquals('"col1" uuid', $result);
 	}
+
+/**
+ * Test that postgres describes default columns with functions correctly.
+ *
+ * @return void
+ */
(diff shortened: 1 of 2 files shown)


markstory:
Looks good. I'll merge this once I have a chance to verify if it applies to 3.x or not.

lib/Cake/Test/Case/Model/Datasource/Database/PostgresTest.php
```diff
@@ -1218,4 +1218,24 @@ public function testBuildColumnUuid() {
...
+ */
+	public function testDescribeFunctionDefault() {
+		$db = $this->Dbo;
+		$db->execute('CREATE TABLE test_function_default_describe (id integer PRIMARY KEY, year int default date_part(\'year\'::text, now()))');
+		$data = $db->describe('test_function_default_describe');
+
+		$expected = array(
+			'type' => 'integer',
+			'null' => true,
+			'default' => 'date_part(\'year\', now())',
```
markstory: Shouldn't the default value retain the cast?

markstory:
I don't think this issue exists in 3.x. I added a test to cover this scenario in ba1ddb3f360ac4470deedc5490128c9702a8fbc3

PR #12655, opened 2018-10-19
4.x internal filesystem

ADmad:
Refs #12569.

My previous PR #12650 combined with this takes care of removing all usage of `File` and `Folder` from core. I'll remove it from tests too once this approach is approved.

I have added a new internal class `Cake\Filesystem\Filesystem` with utility methods for file system operations to avoid code repetition.

### DIFF
--- a/src/Console/CommandScanner.php
+++ b/src/Console/CommandScanner.php
@@ -143,6 +143,6 @@ class CommandScanner
 
         ksort($shells);
 
-        return $shells;
+        return array_values($shells);
     }
 }
--- a/src/Console/CommandScanner.php
+++ b/src/Console/CommandScanner.php
@@ -18,7 +18,7 @@ namespace Cake\Console;
 use Cake\Core\App;
 use Cake\Core\Configure;
 use Cake\Core\Plugin;
-use Cake\Filesystem\Folder;
+use Cake\Filesystem\Filesystem;
 use Cake\Utility\Inflector;
 
 /**
@@ -109,40 +109,39 @@ class CommandScanner
      */
     protected function scanDir(string $path, string $namespace, string $prefix, array $hide): array
     {
-        $dir = new Folder($path);
-        $contents = $dir->read(true, true);
-        if (empty($contents[1])) {
+        if (!is_dir($path)) {
             return [];
         }
 
-        $classPattern = '/(Shell|Command)$/';
-        $shells = [];
-        foreach ($contents[1] as $file) {
-            if (substr($file, -4) !== '.php') {
-                continue;
-            }
-            $shell = substr($file, 0, -4);
-            if (!preg_match($classPattern, $shell)) {
-                continue;
-            }
+        $classPattern = '/(Shell|Command)\.php$/';
+        $fs = new Filesystem();
+        $files = $fs->find($path, $classPattern);
 
-            $name = Inflector::underscore(preg_replace($classPattern, '', $shell));
+        $shells = [];
+        foreach ($files as $fileInfo) {
+            $file = $fileInfo->getFilename();
+
+            $name = Inflector::underscore(preg_replace($classPattern, '', $file));
             if (in_array($name, $hide, true)) {
                 continue;
             }
 
-            $class = $namespace . $shell;
-            if (!is_subclass_of($class, Shell::class) && !is_subclass_of($class, Command::class)) {
+            $class = $namespace . $fileInfo->getBasename('.php');
+            if (!is_subclass_of($class, Shell::class)
+                && !is_subclass_of($class, Command::class)
+            ) {
                 continue;
             }
 
-            $shells[] = [
+            $shells[$path . $file] = [
                 'file' => $path . $file,
                 'fullName' => $prefix . $name,
                 'name' => $name,
                 'class' => $class,
             ];
         }
+
+        ksort($shells);
 
         return $shells;
     }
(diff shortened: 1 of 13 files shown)


src/Console/CommandScanner.php
```diff
@@ -109,41 +109,40 @@ public function scanPlugin(string $plugin): array
...
-            $shells[] = [
+            $shells[$path . $file] = [
                 'file' => $path . $file,
                 'fullName' => $prefix . $name,
                 'name' => $name,
                 'class' => $class,
             ];
         }
 
+        ksort($shells);
```
markstory: Should you `array_values()` after this?

src/Filesystem/Filesystem.php
```diff
@@ -0,0 +1,194 @@
...
+
+        $iterator = new RecursiveIteratorIterator(
+            new RecursiveDirectoryIterator($path, FilesystemIterator::SKIP_DOTS),
+            RecursiveIteratorIterator::CHILD_FIRST
+        );
+
+        $result = true;
+        foreach ($iterator as $fileInfo) {
+            switch ($fileInfo->getType()) {
+                case 'dir':
```
markstory: Does PHP give us constants for these strings?

ADmad: Not that I know of, will check.

src/Console/CommandScanner.php
ADmad: We could but not really necessary, the keys are never used.

src/Filesystem/Filesystem.php
garas: Add `RecursiveDirectoryIterator::CURRENT_AS_SELF` flag to `RecursiveDirectoryIterator` constructor. 

`$fileInfo` is becomes instance of `RecursiveDirectoryIterator`, and has 

-    `DirectoryIterator::isDir()` - Determine if current DirectoryIterator item is a directory
-    `DirectoryIterator::isDot()` - Determine if current DirectoryIterator item is '.' or '..'
-    `DirectoryIterator::isFile()` - Determine if current DirectoryIterator item is a regular file
-    `DirectoryIterator::isLink()` - Determine if current DirectoryIterator item is a symbolic link

src/Filesystem/Filesystem.php
```diff
@@ -0,0 +1,257 @@
...
+        return new CallbackFilterIterator($flatten, $filter);
+    }
+
+    /**
+     * Dump contents to file.
+     *
+     * @param string $filename File path.
+     * @param string $content Content to dump.
+     * @return void
+     * @throws \Cake\Core\Exception When dumping fails.
```
garas: this and 2 more in this file: `\Cake\Core\Exception\Exception`

src/Filesystem/Filesystem.php
```diff
@@ -0,0 +1,257 @@
...
+    /**
+     * Find files (non-recursively) in given directory path.
+     *
+     * @param string $path Directory path.
+     * @param mixed $filter If string will be used as regex for filtering using
+     *   `RegexIterator`, if callable will be as callback for `CallbackFilterIterator`.
+     * @param int|null $flags Flags for FilesystemIterator::__construct();
+     * @return \Traversable
+     */
+    public function find(string $path, $filter = null, ?int $flags = null): Traversable
```
inoas: `findFile()`? If ti finds directories and symbolic links as well... the docblock is wrong?

src/Filesystem/Filesystem.php
```diff
@@ -0,0 +1,257 @@
...
+     * Find files recursively in given directory path.
+     *
+     * @param string $path Directory path.
+     * @param mixed $filter If string will be used as regex for filtering using
+     *   `RegexIterator`, if callable will be as callback for `CallbackFilterIterator`.
+     *   Hidden directories (starting with dot e.g. .git) are always skipped.
+     * @param int|null $flags Flags for FilesystemIterator::__construct();
+     * @return \Traversable
+     */
+    public function findRecursive(string $path, $filter = null, ?int $flags = null): Traversable
```
inoas: findFilesRecursive()?

inoas: As the signature is the same as findFIles could both be called by one proxy method having recursive = true/false as the last arg?

src/Filesystem/Filesystem.php
```diff
@@ -0,0 +1,257 @@
...
+
+    /**
+     * Create directory.
+     *
+     * @param string $dir Directory path.
+     * @param int $mode Octal mode passed to mkdir(). Defaults to 0755.
+     * @return void
+     * @throws \Cake\Core\Exception\Exception When directory creation fails.
+     */
+    public function mkdir(string $dir, int $mode = 0755): void
```
inoas: Why use mkdir but then deleteDir copyDir?
what about either of these for consistency:

mkdir, rmdir, cpdir
mkDir, rmDir, cpDir
makeDir, removeDir, copyDir

inoas:
> I am still in favor of using `symfony/filesystem` instead of the poor man's implementation I have done.

Thanks for doing this, but why was that rejected?

markstory:
I like this approach. It solves the problems we have without adding a non-trivial dependency and the new code looks good.

src/Filesystem/Filesystem.php
ADmad: This class is only meant for internal use and won't be documented in the API hence there's no need to be too pedantic about the naming.

src/Filesystem/Filesystem.php
ADmad: Boolean args are frowned upon. A more explicit API is better.

ADmad:
phpstan failure fixed on 4.x.

ADmad:
Anything more needs to be done here?

markstory:
I'll take one more look :)

PR #12656, opened 2018-10-19
Improve documentation

bancer:


lib/Cake/Utility/ClassRegistry.php
```diff
@@ -91,7 +91,7 @@ public static function getInstance() {
  *  stored in the registry and returned.
  * @param bool $strict if set to true it will return false if the class was not found instead
  *	of trying to create an AppModel
- * @return $class instance of ClassName.
+ * @return boolean|$class instance of ClassName.
```
ADmad:
```suggestion
 * @return bool|$class instance of ClassName.
```
Should be short `bool`.

PR #12658, opened 2018-10-19
Add PHP extension PDO SQLite as dev dependency

ravage84:
It is required to run the test suite.

I noticed this while I was playing around with a vanilla Ubuntu 18.10 for fun...

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -42,6 +42,7 @@
     },
     "require-dev": {
         "phpunit/phpunit": "^5.7.14|^6.0",
+        "ext-pdo_sqlite": "*",
         "cakephp/cakephp-codesniffer": "^3.0"
     },
     "autoload": {


ADmad:
> It is required to run the test suite.

No it's not. One can just set the env. var `db_dsn` to use a MySql database instead of the default Sqlite.

markstory:
@ADmad Our dev dependencies wouldn't bubble up to application developers. Only folks running `composer install` on the cakephp project.

ADmad:
@markstory I am aware of that. But for e.g. why should someone wanting to make a fix for a Cache engine and run tests for it be forced to install sqlite when that wouldn't even need a db connection?

Similarly we don't force devs to install memcached extension to run the test suite do we?

markstory:
@ADmad Tests that require optional dependencies auto-skip themselves when that dependency is not present. If a user doesn't have sqlite3 tests fail/crash which is not great.

ADmad:
> If a user doesn't have sqlite3 tests fail/crash which is not great.

Not if you run tests which don't require any db interaction.

ADmad:
https://ci.appveyor.com/project/cakephp/cakephp/builds/19648193#L376

markstory:
> Instead of this change we can add a check in tests/bootstrap that if db_dsn is not set and pdo_sqlite / sqlite is not available then show a message to either install it or modify the db connection in 

That would work for me.

Issue #12648, opened 2018-10-17
LocaleSelectorMiddleware doesn't support fallbacks, only exact matches


Zuluru:
This is a (multiple allowed):

* [ ] bug
* [x] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.6.12

### What you did
Implemented LocaleSelectorMiddleware.

### What happened
Locales are selected only by exact match to the provided list.

### What you expected to happen
Locales would be selected using fallback to two-letter code if an exact match is not found, per Locale::lookup. For example, en_US would match en, if en_US isn't an option.

My server has seen en_AU, en_CA, en_GB, en_IE, en_NZ, en_US and en_ZA, just among English options. I don't want to have to list all of those just to get to a default English translation. Similar for other languages. Having a fallback seems an obvious option. This is an easy PR to put together using Locale::lookup, if people think it's a good idea.

PR #12663, opened 2018-10-20
Use Locale::lookup to handle generic options as a fallback

Zuluru:
Resolve issue #12648.

### DIFF
--- a/tests/TestCase/I18n/Middleware/LocaleSelectorMiddlewareTest.php
+++ b/tests/TestCase/I18n/Middleware/LocaleSelectorMiddlewareTest.php
@@ -102,6 +102,20 @@ class LocaleSelectorMiddlewareTest extends TestCase
     }
 
     /**
+     * The default locale should change when the request locale has an accepted fallback option
+     *
+     * @return void
+     */
+    public function testInvokeLocaleAcceptedFallback()
+    {
+        $request = ServerRequestFactory::fromGlobals(['HTTP_ACCEPT_LANGUAGE' => 'es-ES;q=0.8,da;q=0.4']);
+        $response = new Response();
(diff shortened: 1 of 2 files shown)


tests/TestCase/I18n/Middleware/LocaleSelectorMiddlewareTest.php
```diff
@@ -101,6 +101,20 @@ public function testInvokeLocaleAccepted()
...
+    /**
+     * The default locale should change when the request locale has an accepted fallback option
+     *
+     * @return void
+     */
+    public function testInvokeLocaleAcceptedFallback()
+    {
+        $request = ServerRequestFactory::fromGlobals(['HTTP_ACCEPT_LANGUAGE' => 'es-ES;q=0.8,da;q=0.4']);
+        $response = new Response();
+        $middleware = new LocaleSelectorMiddleware(['en_CA', 'es']);
```
markstory: What about the case where `en_US` is supported but the request is for `en_GB`?

Zuluru: Per one of the comments on the Locale::lookup man page, this is not a situation that PHP supports. My interpretation is that US English would be an *alternative to* British, not a *fallback for* it. Adding support above and beyond what PHP provides here is certainly possible, but gets more complicated. See http://www.thefutureoftheweb.com/blog/use-accept-language-header for an example that would bypass any usage of PHP's built-in functions for this.

markstory: Sorry, I missed the earlier reply and didn't think to check on this again.

PR #12664, opened 2018-10-21
Replace Magic Literals in SecurityHeadersMiddleware with Class Constants

### LINKED ISSUE #12666: Replace Header Value Magic Literals in SecurityHeadersMiddleware with Constants
CakePHP Version: 3.7

Recently, I was integrating the SecurityHeadersMiddleware  into some of our applications. I noticed I had to use magic literals such as ``sameorigin`` to configure the behavior of the middleware. Since it is a bad practise to use/have magic literals, I thought it would make sense to replace those magic literals in that middleware with class constants.

I wasn't sure which way was the best, so I created two PRs alternative solutions.

1. Replacing the magic literals with class constants within that middleware #12664
1. Replacing the magic literals with class constants from separate, dedicated classes #12665

Personally, I prefer #12665 over #12664, as it is more readable and has a clear separation.

ravage84:
Replace magic literals in SecurityHeadersMiddleware with class constants.
Except for the HTTP header keys. They could be constants, too, but should be placed in another class, e.g. a general HTTP related class.

Constant doc blocks as per PSR-19 (in draft)
https://github.com/php-fig/fig-standards/blob/master/proposed/phpdoc-tags.md#522-var

Closes #12666
Alternative to #12665

Personally, I prefer #12665 over this, as it is more readable and has a clear separation.

### DIFF
--- a/src/Http/Middleware/SecurityHeadersMiddleware.php
+++ b/src/Http/Middleware/SecurityHeadersMiddleware.php
@@ -23,6 +23,71 @@ use Psr\Http\Message\ServerRequestInterface;
  */
 class SecurityHeadersMiddleware
 {
+    /** @var string X-Content-Type-Option nosniff */
+    const NOSNIFF = 'nosniff';
+
+    /** @var string X-Download-Option noopen */
+    const NOOPEN = 'noopen';
+
+    /** @var string Referrer-Policy no-referrer */
+    const NO_REFERRER = 'no-referrer';
+
+    /** @var string Referrer-Policy no-referrer-when-downgrade */
+    const NO_REFERRER_WHEN_DOWNGRADE = 'no-referrer-when-downgrade';
+
+    /** @var string Referrer-Policy origin */
+    const ORIGIN = 'origin';
+
+    /** @var string Referrer-Policy origin-when-cross-origin */
+    const ORIGIN_WHEN_CROSS_ORIGIN = 'origin-when-cross-origin';
+
+    /** @var string Referrer-Policy same-origin */
+    const SAME_ORIGIN = 'same-origin';
+
+    /** @var string Referrer-Policy strict-origin */
+    const STRICT_ORIGIN = 'strict-origin';
+
+    /** @var string Referrer-Policy strict-origin-when-cross-origin */
+    const STRICT_ORIGIN_WHEN_CROSS_ORIGIN = 'strict-origin-when-cross-origin';
+
+    /** @var string Referrer-Policy unsafe-url */
+    const UNSAFE_URL = 'unsafe-url';
+
+    /** @var string X-Frame-Option deny */
+    const DENY = 'deny';
+
+    /** @var string X-Frame-Option sameorigin */
+    const SAMEORIGIN = 'sameorigin';
+
+    /** @var string X-Frame-Option allow-from */
+    const ALLOW_FROM = 'allow-from';
+
+    /** @var string X-XSS-Protection block, sets enabled with block */
+    const XSS_BLOCK = 'block';
+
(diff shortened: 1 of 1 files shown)


dereuromark:
I think this one is just fine and doesnt introduce so many new Interfaces for this one middleware.

markstory:
I agree we don't need 6 classes/files to house constants.

src/Http/Middleware/SecurityHeadersMiddleware.php
```diff
@@ -115,15 +184,15 @@ public function setXFrameOptions($option = 'sameorigin', $url = null)
...
      */
-    public function setXssProtection($mode = 'block')
+    public function setXssProtection($mode = self::XSS_BLOCK)
     {
         $mode = (string)$mode;
 
-        if ($mode === 'block') {
-            $mode = '1; mode=block';
+        if ($mode === self::XSS_BLOCK) {
+            $mode = self::XSS_ENABLED_BLOCK;
```
garas: XSS_BLOCK and XSS_ENABLED_BLOCK could be merged to single constant, as only `'1; mode=block'` is real option, and `'block'` is just shortcut. Constant would hide complexity of real string.

ravage84: @garas while I agree about hiding complexity, ``block`` is currently used a few lines above in the method signature. https://github.com/cakephp/cakephp/pull/12664/files/ec711f5b99384897f594c03a41c6ceae21f063f4#diff-660cb26ba064296df9cc50f62b2d9455R187

ravage84:
OK, I see 6 interfaces/classes i s a bit verbose for what I wanted to achieve. Nonetheless it would the most readable and most reusable solution.

What if we move all those HTTP header key and value constants into one single big interface/calss, instead?

Or potentially two. One for the HTTP header keys and one for the header values?

This way, this one or these two interfaces/classes can be reused everywhere. Since they are HTTP focused, they shouldn't be used in too many places anyway.

Anyway, if we keep the constants here, I feel we should improve the naming first before merging, as the current names are mostly verbatim.

E.g. instead of ``NOSNIFF`` for X-Content-Type-Option "nosniff", we could use ``CONTENT_TYPE_NOSNIFF`` or even  ``CONTENT_TYPE_OPTION_NOSNIFF``to be very clear.

src/Http/Middleware/SecurityHeadersMiddleware.php
garas: Default argument value also should be changed to constant of full string.

In this case, new default will not change result of method.

For BC, literal `'block'` should be used in `if` condition

ADmad:
I don't think there's any need to add constants for these headers. They are standard documented headers, not arbitrary strings chosen by us.

If you are going to add constants of headers where do we stop? There are tons of them https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers

ravage84:
[Magic numbers and literals](https://en.wikipedia.org/wiki/Magic_number_(programming)) are a bad thing. That's why I opened the issue and provided the PRs. I'm also fine with changing the constants to variables, though I think constants are a better fit.

Searching for a magic literal in your can be really a PITA because they can overlap with others etc, but if you have constants, you can easily search for them. Having constants for magic literals _can_ make the code more readable. Certainly, we have to name them well.

A good example, IMO, for such constants is this small package:

https://github.com/php-fig/http-message-util

Providing constants for HTTP method and HTTP status codes:
- https://github.com/php-fig/http-message-util/blob/master/src/RequestMethodInterface.php
- https://github.com/php-fig/http-message-util/blob/master/src/RequestMethodInterface.php

This was kind of my inspiration for this here.

PR #12673, opened 2018-10-23
Include previous exception in log message if present

ceeram:
Allow to opt-in for including previous exceptions (if present) in logging.

### DIFF
--- a/src/Error/Middleware/ErrorHandlerMiddleware.php
+++ b/src/Error/Middleware/ErrorHandlerMiddleware.php
@@ -209,8 +209,32 @@ class ErrorHandlerMiddleware
      */
     protected function getMessage($request, $exception)
     {
+        $message = $this->getMessageForException($exception);
+
+        $message .= "\nRequest URL: " . $request->getRequestTarget();
+        $referer = $request->getHeaderLine('Referer');
+        if ($referer) {
+            $message .= "\nReferer URL: " . $referer;
+        }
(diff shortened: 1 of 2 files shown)


dereuromark:
Why not including by default? Would make the config for it unneeded.

ceeram:
That would change logging output, who knows it might break someones tooling. Personally i would indeed enable it by default in our applications.

ceeram:
how would one make this pass: https://ci.appveyor.com/project/cakephp/cakephp/builds/19730534

dereuromark:
3.7 by default could be an idea then.
i dont think we need an option to enable/disable here.

lorenzo:
I'd love to have that by default :)

ceeram:
so, no opt-in config? or no config at all, and just always include it?

ceeram:
btw, we could have app.default.php in cakephp/app have it enabled by default, if we choose to keep the config, no matter the default value we use here, so we wont break old logging output and new applications get this by default. Or maybe we dont need to support BC on format of error logs at all

markstory:
I don't think we need a config flag if it is in 3.7.0. We can document that the log format is changing in the migration guide. Folks who are munging logs will have to update their tooling.

markstory (approved):
Looks good to me.

ADmad (approved):

ADmad:
@ceeram Can you please open a PR updating the docs too.

PR #12675, opened 2018-10-24
Fix MoParser if a message exist with and without context

Iandenh:
Fix MoParser if an message exist with and without context. And it also makes MoFileParser more inline with the PoFileParser.

Here a example if you have a message where one has a context:
```po
#: Template/Cart/index.ctp:153
msgid "Shipping fee"
msgstr "Verzendkosten"

#: Template/Cart/index.ctp:96
msgctxt "indication"
msgid "Shipping fee"
msgstr "Indicatie verzendkosten"
```

### DIFF
--- a/src/I18n/Parser/MoFileParser.php
+++ b/src/I18n/Parser/MoFileParser.php
@@ -134,9 +134,9 @@ class MoFileParser
                 continue;
             }
 
-            $messages[$singularId] = $singular;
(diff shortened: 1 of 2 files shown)


Ryan-DV:
Nice

markstory (commented):
Other than a nitpick this looks good.

PR #12679, opened 2018-10-26
Improve documentation

bancer:
The parameter description is copied from `CakeLog::write` method.


lib/Cake/Utility/Debugger.php
```diff
@@ -181,7 +181,9 @@ public static function dump($var, $depth = 3) {
  * as well as export the variable using exportVar. By default the log is written to the debug log.
  *
  * @param mixed $var Variable or content to log
- * @param int $level type of log to use. Defaults to LOG_DEBUG
+ * @param int|string $level Type of log to use. Defaults to LOG_DEBUG. When value is an integer
+ *    or a string matching the recognized levels, then it will
+ *    be treated log levels. Otherwise it's treated as scope.
```
markstory:
```suggestion
 *    be treated as a log level. Otherwise it's treated as a scope.
```

PR #12681, opened 2018-10-26
Collection test additions

davidyell:
I have been having some trouble today trying to get a `min` and `max` from the same query. For some reason I get the same value back for both.

I thought that I would check the core tests for examples, and there aren't really any.

This pull request adds some tests which cover my use-case, and are hopefully of some use to the project, hence me contributing them.

I don't know what the feeling is of putting Collection tests into the ResultSet test is, but as the ResultSet class implements the CollectionInterface, this feel okay to me.

Oh and I also updated a duplicate test, which seems to be a copy and paste which was never updated.

### DIFF
--- a/tests/TestCase/ORM/ResultSetTest.php
+++ b/tests/TestCase/ORM/ResultSetTest.php
@@ -21,6 +21,7 @@ use Cake\ORM\Entity;
 use Cake\ORM\ResultSet;
 use Cake\ORM\Table;
 use Cake\TestSuite\TestCase;
+use TestApp\Model\Entity\Article;
 
 /**
  * ResultSet test case.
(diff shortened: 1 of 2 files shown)


davidyell:
So I have now managed to track down the issue which I was having.

Consider this code
```php
$calls = $this->Calls->find()
    ->select(['duration' => 'TIME_TO_SEC(TIMEDIFF(finished, began))'])
    ->find('historic', ['days' => 7])
    ->having(['duration > 0']);

$min = $calls->min('duration');
$max = $calls->max('duration'); 
```

Both the `$min` and `$max` are the same value, the first result.

Turns out that the `Call` entity, has a mutator method `protected function _getDuration(): string` which is then overriding the value of the calculated field.

Feel free to close this PR, unless there is a feeling that the tests are worth having. Or let me know if I should submit this as a new test case?

davidyell:
Thanks for the merge @markstory would you like me to add a test case for the above issue as well?

markstory:
Sure. Could help catch a regression in the future.

tests/TestCase/ORM/ResultSetTest.php
```diff
@@ -416,4 +417,40 @@ public function testIsEmptyDoesNotConsumeData()
...
+
+    /**
+     * Test that ResultSet
+     *
+     * @return void
+     */
+    public function testCollectionMinAndMaxWithAggregateField()
+    {
+        $query = $this->table->find();
+        $query->select(['count' => 'length(title)']);
```
ADmad: This causes failure on SQL Server which doesn't have `length()` function but `len()` instead.

markstory: My bad. Fixed in 0da10887762b489c34a264afd675bc7e3fca4c56.

garas: No 😊 

```
1) Cake\Test\TestCase\ORM\ResultSetTest::testCollectionMinAndMaxWithAggregateField
PDOException: SQLSTATE[42S22]: [Microsoft][ODBC Driver 13 for SQL Server][SQL Server]Invalid column name 'author_id + 2'.
```

markstory: Darn. I'll try again :smile:

PR #12686, opened 2018-10-28
4.x Remove SimpleCacheEngine

markstory:
With each engine implementing the PSR16 interfaces there is no reason for this class to exist anymore. I'll be finishing off default values and TTL in set() in future pull requests.


ADmad (approved):

src/Cache/Cache.php
```diff
@@ -223,7 +223,7 @@ public static function engine(string $config): CacheEngine
      */
     public static function pool(string $config)
     {
-        return new SimpleCacheEngine(static::engine($config));
+        return static::engine($config);
```
saeideng: `static::engine()` deprecated

markstory: Yes. That is also on my list of things to change, but I wanted to take small steps.

lorenzo (approved):
LGTM

src/Cache/Cache.php
saeideng: so , goods for this step :+1:

Issue #12682, opened 2018-10-26
No way anymore to load plugins on demand and in non-integration tests?


ndm2:
This is a (multiple allowed):

* [x] bug
* [x] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 4.0.0

--------------------------

`Plugin::load()` has been deprecated in favour of `Application::addPlugin()`, and will be removed in 4.0. 

I'm using `Plugin::load()` a lot in the test environment, and having to load everything at (application) bootstrap time, I don't know, I have a feeling that this will introduce quite some overhead to get plugin loading "on demand" working, which might be required to for example avoid polluting the test environment with unnecessary or even conflicting bootstrapping logic and routes.

Furthermore, outside of integration tests, the `Application` class will not be involved in the test environment, meaning there would be no clean way to load plugins for regular unit tests at all?

Maybe I'm missing something, maybe there are (plans for) alternatives? I know there's `Plugin::getCollection()->add()`, but `Plugin::getCollection()` is marked as internal, so it cannot really be treated as a stable API.

btw, why isn't `Plugin::unload()` deprecated too?

markstory:
> btw, why isn't Plugin::unload() deprecated too?

If there is no load, there is nothing to unload 😛 but mostly because I forgot.

Are you loading plugins dynamically during your application's test suite? Wouldn't your application's tests be more representative if all plugins were loaded?

ndm2:
Well, I'd say it depends, it might not neccessarily be an actual application, but just a test application, for example when developing standalone plugins. Such a plugin, or pretty much any extensible system, might need to be tested separately with other plugins, in order to ensure things work with only additional plugin X and Y available.

There might be ways to overcome this, but I think there's at least two situations that show the need for some way to dynamically load plugins:

1. When the tests aren't integration tests, as in that case the `Application` class is not going to be involved, ie `Application::addPlugin()` cannot be used.

2. When testing standalone plugins, as the plugin must somehow be able to load itself. [**DebugKit for example uses**](https://github.com/cakephp/debug_kit/blob/918b810e42cdf36b6228f98a504121a795b675f5/tests/bootstrap.php#L116) the for internal use only marked `Plugin::getCollection()`: 

[...]

PR #12687, opened 2018-10-29
3.next - Plugin unload

markstory:
I missed this earlier when working on new Plugins. This method should go away along with `Plugin::load()`.

Refs #12682


ADmad (approved):

tests/TestCase/Auth/FormAuthenticateTest.php
```diff
@@ -298,7 +298,6 @@ public function testPluginModel()
             'updated' => new Time('2007-03-17 01:18:31')
         ];
         $this->assertEquals($expected, $result);
-        Plugin::unload();
```
saeideng: this should be there 
`Plugin::getCollection()->clear();`

PR #12688, opened 2018-10-29
Refactored ConsoleIntegrationTestTrait private vars into protected

### LINKED ISSUE #12683: Integration tests cannot access console output anymore
This is a (multiple allowed):

* [x] bug
* [x] enhancement

* CakePHP Version: 3.7.0-RC1

---------------

I often want to test the whole output, not just whether it contains something, so I often used `$this->_out/_err->messages()` instead of the available assertion methods.

#12072 changed the properties holding the input/output objects (`$_in`, `$_out`, `$_rr`) from protected to private (btw this isn't documented as a breaking change in the migration guide), so it's no longer possible to retrieve the generated output.

One could probably work around this by importing the trait directly and changing the properties visibility, but having to not do that would IMHO be nicer :) And while additional assertion methods might work for some, people might still want test/inspect the data in other ways that require access to the actual data, so being able to directly access the output wouldn't hurt I think.

jeremyharris:
Closes #12683 

Makes vars that were previously `protected` but inadvertently became `private` during a refactor back to `protected`.

### DIFF
--- a/src/TestSuite/ConsoleIntegrationTestTrait.php
+++ b/src/TestSuite/ConsoleIntegrationTestTrait.php
@@ -37,35 +37,35 @@ trait ConsoleIntegrationTestTrait
      *
      * @var bool
      */
-    private $useCommandRunner = false;
+    protected $_useCommandRunner = false;
 
     /**
      * Last exit code
      *
      * @var int|null
      */
-    private $exitCode;
+    protected $_exitCode;
 
     /**
      * Console output stub
      *
      * @var \Cake\TestSuite\Stub\ConsoleOutput|\PHPUnit_Framework_MockObject_MockObject|null
      */
-    private $out;
+    protected $_out;
 
     /**
      * Console error output stub
      *
      * @var \Cake\TestSuite\Stub\ConsoleOutput|\PHPUnit_Framework_MockObject_MockObject|null
      */
-    private $err;
+    protected $_err;
 
     /**
      * Console input mock
      *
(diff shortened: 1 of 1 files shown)


src/TestSuite/ConsoleIntegrationTestTrait.php
```diff
@@ -37,35 +37,35 @@
      *
      * @var bool
      */
-    private $useCommandRunner = false;
+    protected $_useCommandRunner = false;
```
lorenzo: no need to underscore the variable, I think

src/TestSuite/ConsoleIntegrationTestTrait.php
```diff
@@ -78,28 +78,28 @@ public function exec($command, array $input = [])
     {
         $runner = $this->makeRunner();
 
-        $this->out = new ConsoleOutput();
-        $this->err = new ConsoleOutput();
-        $this->in = $this->getMockBuilder(ConsoleInput::class)
+        $this->_out = new ConsoleOutput();
+        $this->_err = new ConsoleOutput();
+        $this->_in = $this->getMockBuilder(ConsoleInput::class)
```
markstory: :+1: This restores the property names to what they were in 3.6.

src/TestSuite/ConsoleIntegrationTestTrait.php
markstory: The variables has `_` prefixes in 3.6 though.

PR #12692, opened 2018-10-30
Add model id to all `exists()` method calls

bancer:
Related issue - https://github.com/dereuromark/cakephp-shim/issues/25
Also see https://github.com/dereuromark/cakephp-shim/blob/2.x/Model/ShimModel.php#L558.

### DIFF
--- a/lib/Cake/Console/Templates/default/actions/controller_actions.ctp
+++ b/lib/Cake/Console/Templates/default/actions/controller_actions.ctp
@@ -132,7 +132,7 @@
  */
 	public function <?php echo $admin; ?>delete($id = null) {
 		$this-><?php echo $currentModelName; ?>->id = $id;
-		if (!$this-><?php echo $currentModelName; ?>->exists()) {
+		if (!$this-><?php echo $currentModelName; ?>->exists($id)) {
 			throw new NotFoundException(__('Invalid <?php echo strtolower($singularHumanName); ?>'));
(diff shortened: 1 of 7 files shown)


dereuromark:
Tests need to be adjusted: https://travis-ci.org/cakephp/cakephp/jobs/448237264

bancer:
@dereuromark Done :).

lib/Cake/Console/Templates/default/actions/controller_actions.ctp
```diff
@@ -132,7 +132,7 @@
  */
 	public function <?php echo $admin; ?>delete($id = null) {
 		$this-><?php echo $currentModelName; ?>->id = $id;
-		if (!$this-><?php echo $currentModelName; ?>->exists()) {
+		if (!$this-><?php echo $currentModelName; ?>->exists($id)) {
```
markstory: Do we need to set the property on line 134 still?

bancer: I believe it is still needed for `delete()` call below.

dereuromark: It could be moved down to it then maybe.

bancer: Done.

PR #12695, opened 2018-10-31
Fix router reload regression in integration testing.

dereuromark:
Alternative to https://github.com/cakephp/cakephp/pull/12628
It provides a now "BC way" to shim it back for the use cases where we need it.
Not sure that is the best solution, but it would provide a way for integration tests to fix the regression manually where needed via
```php
class FeedExamplesControllerTest extends IntegrationTestCase {

	/**
	 * Allow router reloading to be disabled.
	 *
	 * @var bool
	 */
	protected $_disableRouterReload = true;
...
```
per class.

### DIFF
--- a/tests/TestCase/TestSuite/IntegrationTestCaseTest.php
+++ b/tests/TestCase/TestSuite/IntegrationTestCaseTest.php
@@ -19,6 +19,7 @@ use Cake\Core\Plugin;
 use Cake\Event\EventManager;
 use Cake\Http\Response;
 use Cake\Routing\DispatcherFactory;
+use Cake\Routing\RouteBuilder;
 use Cake\Routing\Router;
 use Cake\Routing\Route\InflectedRoute;
 use Cake\TestSuite\IntegrationTestCase;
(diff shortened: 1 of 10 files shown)


lorenzo:
looks good to me, just needs a test

dereuromark:
I tried adding a test, but I cant reproduce the issue apps or plugins are facing.
The test works even without this flag enabled, but this is not true for the actual plugin or app cases.

I need help here to finalize this PR for the upcoming patch release.

dereuromark:
As proof, I have https://github.com/dereuromark/cakephp-sandbox/pull/14 which shows this fixes the currently broken master tests ("Tests: 65, Assertions: 286, Errors: 4").

tests/TestCase/TestSuite/IntegrationTestCaseTest.php
```diff
@@ -43,7 +44,8 @@ public function setUp()
         static::setAppNamespace();
 
         Router::reload();
-        Router::scope('/', function ($routes) {
+        Router::extensions(['json']);
```
markstory: Because routes are added in the setup method I don't think the router would be reloaded as routes would not be automatically loaded.

dereuromark: Do you want to change anything on that test case?
Feel free to add a commit on top with desired changes.

markstory: It'll be ok. I think this code won't be needed in 4.x, or at least I hope it isn't.

PR #12696, opened 2018-10-31
3.7 Allow empty by Flags Validation

ravage84:
I've taken @chinpei215 's work from his PR #12486.

@chinpei215 proposed his changes as a workaround for a vulnerability. But since we couldn't merge the changes into 3.6 or even earlier versions, he withdrew his PR. https://github.com/cakephp/cakephp/pull/12486#issuecomment-419923784

Because I think the changes as they stand are a nice addition, I wanted to re-propose them and open a discussion about them. Not as a workaround for a vulnerability, but as an API improvement for CakePHP 3.7 or 4.0.

I haven't changed anything yet.

### DIFF
--- a/src/Validation/Validator.php
+++ b/src/Validation/Validator.php
@@ -771,6 +771,7 @@ class Validator implements ArrayAccess, IteratorAggregate, Countable
      * Valid values are true (always), 'create', 'update'. If a callable is passed then
      * the field will allowed to be empty only when the callback returns true.
      * @param string|null $message The message to show if the field is not
+     * @since 3.7.0
      * @return $this
      */
     public function allowEmptyFor($field, $flags, $when = true, $message = null)
@@ -797,6 +798,7 @@ class Validator implements ArrayAccess, IteratorAggregate, Countable
      * the field will allowed to be empty only when the callback returns true.
      * @param string|null $message The message to show if the field is not
      * @return $this
+     * @since 3.7.0
      * @see \Cake\Validation\Validator::allowEmptyFor() For detail usage
      */
     public function allowEmptyString($field, $when = true, $message = null)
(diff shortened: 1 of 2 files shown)


markstory (approved):
I still think these are good additions that allow us to make validators more expressive and restrictive. @lorenzo @ADmad ?

ADmad:
LGTM, just 2 things:

1. Not too thrilled about the method name `allowEmptyByFlags()`.

2. If we are going to mark `allowEmpty()` as (soft) deprecated then bake should also be updated to use these new methods.

ravage84:
@admad I agree about the naming. Now about ``allowEmotyFor()`` or ``allowEmptyForType()``?

ADmad:
`allowEmptyFor()` works for me.

ADmad:
I took care of the method rename.

ravage84:
@admad thanks and greetings from my army duty.

src/Validation/Validator.php
```diff
@@ -2051,22 +2313,61 @@ protected function _canBeEmpty($field, $context)
...
      *
-     * @param mixed $data value to check against
+     * @param mixed $data Value to check against.
      * @return bool
+     * @deprecated Use isEmpty() instead
      */
     protected function _fieldIsEmpty($data)
     {
-        if (empty($data) && !is_bool($data) && !is_numeric($data)) {
+        return $this->isEmpty($data, self::EMPTY_ALL);
```
dereuromark: Can we please use static in these cases?
Especially ALL can at some point be needed as LTB to add another missing one into it.

dereuromark:
"Add `@SInCE` annotations" :trollface: 

Apart from the static missing part this looks good to me.

ADmad:
It was hoping putting it in back ticks would avoid the auto linking but didn't :stuck_out_tongue:.

dereuromark:
I never does for commits... A bit bad for those that have annotation usernames :P But otherwise meeh.

ADmad:
Meh, not my problem. It's the user's / github's fault :slightly_smiling_face:.

dereuromark:
I still think `self::EMPTY_ALL` should better be LateStaticBinding "static::".

PR #12698, opened 2018-11-01
Add disable methods

### LINKED ISSUE #11864: RFC: Split methods of type `enableFoo()` with boolean argument.
While updating my plugins for CakePHP 3.6 I felt odd having to do method calls like `$query->enableHydration(false)`.

It would be better to separate methods `enableFoo()` for enabling a flag and `disableFoo()` for disabling a flag.

These are the enabling methods which take boolean argument which would have to be changed:

- `\Cake\ORM\Query::enableHydration(bool $enable = true)`
- `\Cake\Database\Connection::enableSavePoints(bool $enable)` // This odd one doesn't have a default value
- `\Cake\Database\Driver::enableAutoQuoting(bool $enable = true)`
- `\Cake\Database\Query::enableBufferedResults(bool $enable = true)`
- `\Cake\View\ViewBuilder::enableAutoLayout(bool $enable = true)`

ADmad:
Closes #11864.

### DIFF
--- a/src/ORM/Query.php
+++ b/src/ORM/Query.php
@@ -1011,6 +1011,22 @@ class Query extends DatabaseQuery implements JsonSerializable, QueryInterface
     }
 
     /**
+     * Disable hydrating entities.
+     *
+     * Disabling hydration will cause array results to be returned for the query
+     * instead of entities.
+     *
+     * @return $this
+     */
+    public function disableHydration()
+    {
+        $this->_dirty();
+        $this->_hydrate = false;
+
+        return $this;
+    }
+
+    /**
      * Returns the current hydration mode.
      *
      * @return bool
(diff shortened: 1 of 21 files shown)


src/ORM/Query.php
```diff
@@ -1010,6 +1010,22 @@ public function enableHydration($enable = true)
...
+     * Disable hydrating entities.
+     *
+     * Disabling hydration will cause array results to be returned for the query
+     * instead of entities.
+     *
+     * @return $this
+     */
+    public function disableHydration()
+    {
+        $this->_dirty();
```
dereuromark: Instead of duplicating the logic this must call the actual wrapper internally to prevent unDRY and copy paste bugs.

ADmad: I would rather have a couple of line of code duplication then adding additional method call.

dereuromark: Dont nano opt a syntactic sugar wrapper. Thats a nogo, the wrapping is cheap and will prevent future bugs.

markstory (approved):
Looks good to me.

dereuromark:
:cry:

PR #12699, opened 2018-11-02
4x - Support default values in cache engine

markstory:
Add support for default values in `CacheEngine::get()`. This is a necessary piece of PSR16 support. This also normalized cache 'misses' to be `null`. Previously you would get null or false depending on which cache engine you were using :cry:

I have not tested wincache locally and will use Appveyor to muddle my way through getting that fixed :blush:


markstory:
Thank you for the assist @ADmad

PR #12701, opened 2018-11-03
Allow multiple delimiters on Inflector::humanize

ahmyi:
Allow mutiple delimiters for Inflector::humanize method
eg.
```PHP
$field = "user.is_active";
echo Inflector::humanize($field,[".","_"]); 
```
Will return User Is Active instead of User.is_active or User Is_active

### DIFF
--- a/src/Utility/Inflector.php
+++ b/src/Utility/Inflector.php
@@ -644,12 +644,12 @@ class Inflector
     public static function humanize($string, $delimiter = '_')
     {
         $delimiters = $delimiter;
-        if(is_array($delimiter)){
+        if(is_array($delimiter))
+        {
             $delimiters = implode("_", $delimiter);
         }
         
         $cacheKey = __FUNCTION__ . $delimiters;
-
         $result = static::_cache($cacheKey, $string);
 
         if ($result === false) {
--- a/src/Utility/Inflector.php
+++ b/src/Utility/Inflector.php
@@ -625,7 +625,7 @@ class Inflector
      * Also replaces underscores with dashes
      *
      * @param string $string The string to dasherize.
-     * @return mixed Dashed version of the input string
+     * @return string|array Dashed version of the input string or multiple strings
      */
     public static function dasherize($string)
     {
--- a/src/Utility/Inflector.php
+++ b/src/Utility/Inflector.php
@@ -645,7 +645,7 @@ class Inflector
     {
         $delimiters = $delimiter;
         if (is_array($delimiter)) {
-            $delimiters = implode("_", $delimiter);
+            $delimiters = implode('_', $delimiter);
         }
         
         $cacheKey = __FUNCTION__ . $delimiters;
--- a/src/Utility/Inflector.php
+++ b/src/Utility/Inflector.php
@@ -647,7 +647,7 @@ class Inflector
         if (is_array($delimiter)) {
             $delimiters = implode('_', $delimiter);
         }
-        
+
         $cacheKey = __FUNCTION__ . $delimiters;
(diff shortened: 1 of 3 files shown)


src/Utility/Inflector.php
```diff
@@ -625,7 +625,7 @@ public static function underscore($string)
      * Also replaces underscores with dashes
      *
      * @param string $string The string to dasherize.
-     * @return string Dashed version of the input string
+     * @return mixed Dashed version of the input string
```
dereuromark: This looks like an invalid change.

src/Utility/Inflector.php
```diff
@@ -625,7 +625,7 @@ public static function underscore($string)
      * Also replaces underscores with dashes
      *
      * @param string $string The string to dasherize.
-     * @return string Dashed version of the input string
+     * @return string|array Dashed version of the input string or multiple strings
```
dereuromark: Why do you think this? It is not true. This is still an array.

src/Utility/Inflector.php
```diff
@@ -643,8 +643,12 @@ public static function dasherize($string)
      */
     public static function humanize($string, $delimiter = '_')
```
dereuromark: If you accept array, why not changing the doc block?

src/Utility/Inflector.php
```diff
@@ -643,8 +643,12 @@ public static function dasherize($string)
...
     {
-        $cacheKey = __FUNCTION__ . $delimiter;
-
+        $delimiters = $delimiter;
+        if(is_array($delimiter)) {
```
dereuromark: CS invalid here

dereuromark:
I am not sure if weakening the input type to allow for array use case is a good idea for the existing method.
Especially not until 4.x.

Maybe this just warrants its own (plugin/app or core) method instead?

src/Utility/Inflector.php
ahmyi: Yes. In cases like having multiple delimiters an array should be passed. As test case given  

```PHP
$this->assertSame('User Is Active', Inflector::humanize('user.is_active',['.','_']));
```
If you pass an array directly you will have a notice error at 
```PHP
 $cacheKey = __FUNCTION__ . $delimiter;
```

src/Utility/Inflector.php
ahmyi: Sorry wrong doc block

ahmyi:
Use case for Forms for with multiple models. When label with 
```PHP
$this->Form->controller(<model>.<field>); 
```
It will display Name as label for multiple fields with name. Example
```PHP
echo $this->Form->control('Father.first_name',['type'=>'text']);
echo $this->Form->control('Mother.first_name',['type'=>'text']);
```
image

Implementing in FormHelper with humanize multiple delimiters the output will be different

image


Well user can use if you want to strict with only string

```PHP
echo Inflector::humanize(Inflector::humanize($field,"."))
```

dereuromark:
Instead of master, should this point to 3.next (3.7?)

markstory:
You have mentioned that this will make form label inflection better but I don't see where those changes are.

On it's own, I don't see how this is better than calling humanize() twice. If an application needs this kind of inflection in multiple places a function could be defined to provide the desired results.

dereuromark:
Do you find a way to have this part in the FormHelper instead?
The general direction of feedback seems to lean in that direction instead of changing this low level method here.

dereuromark:
Closing in favor of discussion https://github.com/cakephp/cakephp/issues/12702 for now, lets see what new solution we can come up with.

PR #12709, opened 2018-11-07
fix phpDocumentator tag

Graziel:


garas:
It was soft-deprecated in https://github.com/cakephp/cakephp/pull/12062. Look there for reasons.

PR #12714, opened 2018-11-09
Fix paginator defaulting for multi-default-sort.

dereuromark:
```php
        $this->paginate = [
            'order' => ['released' => 'DESC', 'modified' => 'DESC'],
        ];
```
results in /index in links like

    /controller/action?page=x&direction=DESC

Currently, if you have a multi default sort set, it would always add the direction query string into even the index root (without any query string filters) pagination links it builds.
This is due to a logical error.
The result should be

    /controller/action?page=x

sort is not set here, but direction always is based on completeSort. This fixes it.


markstory (commented):
Should this have a test?

dereuromark:
Of course, I added a test case :)

PR #12715, opened 2018-11-09
- Add the functionality to accept enum MySQL datatype storing the Integer values as String

yasnaderi:
This is a:
* [x] bug
* [ ] enhancement

* CakePHP Version: 2.9.8
* Platform and Target: MySQL as Database

### What you did
I have a table with like this:
```
CREATE TABLE `configs` (
	...
  `define_time_format` enum('12','24') NOT NULL DEFAULT '24',
  ...
  )

```

### What happened
In a **HTML select** I send the **"24"** as the value but in **"...\lib\Cake\Model\Datasource\DboSource.php"** the **"value"** function it consider as numeric value  **24** and cause error in save.

### DIFF
--- a/lib/Cake/Model/Datasource/DboSource.php
+++ b/lib/Cake/Model/Datasource/DboSource.php
@@ -367,18 +367,18 @@ class DboSource extends DataSource {
 			$column = $this->introspectType($data);
 		}
 
+		$isStringEnum = false;
 		if (strpos($column, "enum") === 0) {
-			preg_match("/(enum\()(.*)(\))/", $column, $acceptingValues);
+			$firstValue = null;
 			if(preg_match("/(enum\()(.*)(\))/",$column, $acceptingValues)){
-			    $values = explode(",", $acceptingValues[2]);
-			    $firstValue = $value[0];
+				$values = explode(",", $acceptingValues[2]);
+				$firstValue = $value[0];
 			}
 			if (is_string($firstValue)) {
 				$isStringEnum = true;
 			}
 		}
 
-		
 		switch ($column) {
 			case 'binary':
 				return $this->_connection->quote($data, PDO::PARAM_LOB);
--- a/lib/Cake/Model/Datasource/DboSource.php
+++ b/lib/Cake/Model/Datasource/DboSource.php
@@ -368,7 +368,7 @@ class DboSource extends DataSource {
 		}
 
 		if (strpos($column, "enum")===0){
-			preg_match("/(enum\()(.*)(\))/",$column, $acceptingValues);
+			preg_match("/(enum\()(.*)(\))/", $column, $acceptingValues);
 			$firstValue = explode(",", $acceptingValues[2])[0];
 			if (is_string($firstValue)){
 				$isStringEnum = true;
--- a/lib/Cake/Model/Datasource/DboSource.php
+++ b/lib/Cake/Model/Datasource/DboSource.php
@@ -367,6 +367,15 @@ class DboSource extends DataSource {
 			$column = $this->introspectType($data);
 		}
 
+		if (strpos($column, "enum")===0){
+			preg_match("/(enum\()(.*)(\))/",$column, $acceptingValues);
+			$firstValue = explode(",", $acceptingValues[2])[0];
+			if (is_string($firstValue)){
+				$isStringEnum = true;
+			}
+		}
+
+		
 		switch ($column) {
 			case 'binary':
 				return $this->_connection->quote($data, PDO::PARAM_LOB);
@@ -382,11 +391,12 @@ class DboSource extends DataSource {
 				if (is_float($data)) {
 					return str_replace(',', '.', strval($data));
 				}
-				if ((is_int($data) || $data === '0') || (
+				if (((is_int($data) || $data === '0') || (
 					is_numeric($data) &&
 					strpos($data, ',') === false &&
 					$data[0] != '0' &&
 					strpos($data, 'e') === false)
+					) && empty($isStringEnum)
 				) {
 					return $data;
 				}
--- a/lib/Cake/Model/Datasource/DboSource.php
+++ b/lib/Cake/Model/Datasource/DboSource.php
@@ -370,7 +370,7 @@ class DboSource extends DataSource {
 		$isStringEnum = false;
 		if (strpos($column, "enum") === 0) {
 			$firstValue = null;
-			if(preg_match("/(enum\()(.*)(\))/",$column, $acceptingValues)){
+			if (preg_match("/(enum\()(.*)(\))/",$column, $acceptingValues)) {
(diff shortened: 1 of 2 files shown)


dereuromark:
Enum was never really supported, it is also a usually very bad thing to use in rel. databases.
Instead most people use more code driven [enum solutions](https://www.dereuromark.de/2010/06/24/static-enums-or-semihardcoded-attributes/) usually.

Never the less, we should try to fix these issues where possible, of course.

lib/Cake/Model/Datasource/DboSource.php
```diff
@@ -367,6 +367,15 @@ public function value($data, $column = null, $null = true) {
 			$column = $this->introspectType($data);
 		}
 
+		if (strpos($column, "enum")===0){
```
markstory:
```suggestion
		if (strpos($column, "enum") === 0) {
```

lib/Cake/Model/Datasource/DboSource.php
```diff
@@ -367,6 +367,15 @@ public function value($data, $column = null, $null = true) {
...
+			preg_match("/(enum\()(.*)(\))/",$column, $acceptingValues);
```
markstory:
```suggestion
			preg_match("/(enum\()(.*)(\))/", $column, $acceptingValues);
```

lib/Cake/Model/Datasource/DboSource.php
```diff
@@ -367,6 +367,15 @@ public function value($data, $column = null, $null = true) {
...
+			$firstValue = explode(",", $acceptingValues[2])[0];
+			if (is_string($firstValue)){
```
markstory:
```suggestion
			if (is_string($firstValue)) {
```

lib/Cake/Model/Datasource/DboSource.php
```diff
@@ -382,11 +391,12 @@ public function value($data, $column = null, $null = true) {
 				if (is_float($data)) {
 					return str_replace(',', '.', strval($data));
 				}
-				if ((is_int($data) || $data === '0') || (
+				if (((is_int($data) || $data === '0') || (
 					is_numeric($data) &&
 					strpos($data, ',') === false &&
 					$data[0] != '0' &&
 					strpos($data, 'e') === false)
+					) && empty($isStringEnum)
```
markstory: You should probably default `isStringEnum` to false/null. That would let you not use empty and check its truthyness instead.

lib/Cake/Model/Datasource/DboSource.php
```diff
@@ -367,6 +367,15 @@ public function value($data, $column = null, $null = true) {
 			$column = $this->introspectType($data);
 		}
 
+		if (strpos($column, "enum")===0){
+			preg_match("/(enum\()(.*)(\))/",$column, $acceptingValues);
+			$firstValue = explode(",", $acceptingValues[2])[0];
```
markstory: This is invalid syntax in PHP5.3 [a build failed because of this](https://travis-ci.org/cakephp/cakephp/jobs/452931595).

lib/Cake/Model/Datasource/DboSource.php
yasnaderi:
```suggestion
		$isStringEnum = false;
		if (strpos($column, "enum")===0){
```

lib/Cake/Model/Datasource/DboSource.php
yasnaderi:
```suggestion
					) && !$isStringEnum
```

yasnaderi (commented):
- Avoid using empty function by initializing a variable

lib/Cake/Model/Datasource/DboSource.php
```diff
@@ -367,6 +367,18 @@ public function value($data, $column = null, $null = true) {
 			$column = $this->introspectType($data);
 		}
 
+		if (strpos($column, "enum") === 0) {
+			preg_match("/(enum\()(.*)(\))/", $column, $acceptingValues);
+			if(preg_match("/(enum\()(.*)(\))/",$column, $acceptingValues)){
```
markstory: Why do you call `preg_match()` twice here? Once should be enough if you capture the matches.

yasnaderi: > Why do you call `preg_match()` twice here? Once should be enough if you capture the matches.

I used GitHub functionality and I thought it will remove the first one, but not.

lib/Cake/Model/Datasource/DboSource.php
```diff
@@ -367,6 +367,18 @@ public function value($data, $column = null, $null = true) {
 			$column = $this->introspectType($data);
 		}
 
+		if (strpos($column, "enum") === 0) {
+			preg_match("/(enum\()(.*)(\))/", $column, $acceptingValues);
```
yasnaderi:
```suggestion
			
```

yasnaderi (commented):
- Correct a typo

markstory:
It would be good to have a test for this scenario. Here is [an existing enum test](https://github.com/cakephp/cakephp/blob/320cdf98ee6e5115d66ef90de5caaaec9c556112/lib/Cake/Test/Case/Model/Datasource/DboSourceTest.php#L1026-L1048) you could use as reference.

lib/Cake/Model/Datasource/DboSource.php
yasnaderi: > Why do you call `preg_match()` twice here? Once should be enough if you capture the matches.

Done

CTRNet (approved):

markstory:
Once the [ci tests are passing](https://travis-ci.org/cakephp/cakephp/builds/457496337) this can be merged.

yasnaderi:
Done

lib/Cake/Test/Case/Model/Datasource/DboSourceTest.php
```diff
@@ -1047,6 +1047,35 @@ public function testNotNullOnEnum() {
...
+		$this->assertTrue($result);
+
+		$EnumFayaTest = ClassRegistry::init('EnumFayaTest');
+		$enumResult = $EnumFayaTest->save(array('faya' => '10'));
+
+		$query = "DROP TABLE {$name};";
+		$result = $this->db->query($query);
+		$this->assertTrue($result);
+
+		$this->assertEquals(array(
```
garas: Move this assert immediately after where `$enumResult` is retrieved.

yasnaderi: @markstory 

[All the test](https://travis-ci.org/cakephp/cakephp/builds/458401437) passes except [This](https://travis-ci.org/cakephp/cakephp/jobs/458401456), this is in _Allowed Failures_ tests

PR #12717, opened 2018-11-10
Add undocumented properties

bancer:
See https://book.cakephp.org/2.0/en/controllers/request-response.html#accessing-request-parameters
and https://api.cakephp.org/2.10/source-class-UpgradeShell.html#345-403

### DIFF
--- a/lib/Cake/Controller/Component/RequestHandlerComponent.php
+++ b/lib/Cake/Controller/Component/RequestHandlerComponent.php
@@ -73,6 +73,13 @@ class RequestHandlerComponent extends Component {
 	public $ext = null;
 
 /**
+ * Array of parameters parsed from the URL.
+ *
+ * @var array
+ */
+	public $params = null;
+
+/**
  * The template to use when rendering the given content type.
  *
  * @var string
--- a/lib/Cake/Controller/Component/RequestHandlerComponent.php
+++ b/lib/Cake/Controller/Component/RequestHandlerComponent.php
@@ -132,7 +132,7 @@ class RequestHandlerComponent extends Component {
 		if (empty($this->ext) || $this->ext === 'html') {
 			$this->_setExtension();
 		}
-		$this->params = $controller->params;
+		$this->params = $controller->request->params;
 		if (!empty($this->settings['viewClassMap'])) {
 			$this->viewClassMap($this->settings['viewClassMap']);
 		}
--- a/lib/Cake/Controller/Component/RequestHandlerComponent.php
+++ b/lib/Cake/Controller/Component/RequestHandlerComponent.php
@@ -75,7 +75,7 @@ class RequestHandlerComponent extends Component {
 /**
  * Array of parameters parsed from the URL.
  *
- * @var array
+ * @var array|null
  */
 	public $params = null;
 
--- a/lib/Cake/Controller/Controller.php
+++ b/lib/Cake/Controller/Controller.php
@@ -932,7 +932,7 @@ class Controller extends CakeObject implements CakeEventListener {
 /**
  * Instantiates the correct view class, hands it its data, and uses it to render the view output.
  *
- * @param string $view View to use for rendering
+ * @param bool|string $view View to use for rendering
  * @param string $layout Layout to use
  * @return CakeResponse A response object containing the rendered view.
  * @triggers Controller.beforeRender $this
--- a/lib/Cake/Controller/Controller.php
+++ b/lib/Cake/Controller/Controller.php
@@ -87,7 +87,7 @@ class Controller extends CakeObject implements CakeEventListener {
  *
  * The default value is `true`.
  *
- * @var mixed
+ * @var bool|array
  * @link https://book.cakephp.org/2.0/en/controllers.html#components-helpers-and-uses
  */
 	public $uses = true;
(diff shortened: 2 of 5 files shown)


garas:
These properties are handled by magic `__get()`, `__set()`, `__isset()` methods. After you have added these properties, magic methods are no longer called, breaks behavior and fails tests.

Adding `@property` tags to class PhpDoc would help IDE to see these properties, but they don't allow mark properties as `@deprecated`.

lib/Cake/Controller/Controller.php
```diff
@@ -389,15 +396,19 @@ public function __isset($name) {
  * @return mixed The requested value for valid variables/aliases else null
  */
 	public function __get($name) {
+		$message = "Property $name is deprecated. Use CakeRequest::$name instead.";
 		switch ($name) {
 			case 'base':
 			case 'here':
 			case 'webroot':
 			case 'data':
+				trigger_error($message, E_USER_DEPRECATED);
```
ADmad: New deprecation error messages shouldn't be added in bugfix releases.

saeideng:
I'm :-1:  with adding any new deprecation/error messages into cake 2.x series 
CakePHP 2.x is in maintenance mode

lib/Cake/Controller/Controller.php
bancer: Would it be okay to write to a log file instead?

chinpei215:
Please do not write logs either. If we had a chance to release release 2.11, I would change `__get()` to `&__get()` though.

lib/Cake/Controller/Controller.php
```diff
@@ -389,15 +396,19 @@ public function __isset($name) {
  * @return mixed The requested value for valid variables/aliases else null
  */
 	public function __get($name) {
+		$message = "Property $name is deprecated. Use CakeRequest::$name instead.";
 		switch ($name) {
 			case 'base':
 			case 'here':
 			case 'webroot':
 			case 'data':
+				CakeLog::write(E_USER_DEPRECATED, $message);
```
ADmad: I don't think filling up the logs is going to do any good either. Lot of 2.x plugins / apps might never get updated to remove use of deprecated code.

lib/Cake/Controller/Component/RequestHandlerComponent.php
```diff
@@ -72,6 +72,13 @@ class RequestHandlerComponent extends Component {
  */
 	public $ext = null;
 
+/**
+ * Array of parameters parsed from the URL.
+ *
+ * @var array
```
dereuromark: |null missing

lib/Cake/basics.php
```diff
@@ -227,6 +226,7 @@ function h($text, $double = true, $charset = null) {
 		}
 		if (is_string($double)) {
 			$charset = $double;
+			$double = true;
```
saeideng: this is not similar to `master` version 
https://github.com/cakephp/cakephp/blob/21c8fce5ca8b68292b948a28606006faa935b579/src/Core/functions.php#L66-L71

lib/Cake/View/View.php
```diff
@@ -455,7 +455,7 @@ public function elementExists($name) {
  * a plugin view/layout can be used instead of the app ones. If the chosen plugin is not found
  * the view will be located along the regular view path cascade.
  *
- * @param string $view Name of view file to use
+ * @param bool|string $view Name of view file to use.
```
saeideng: use of `bool` for first args  is useless 
`bool` is not acceptable here

bancer: Well, that's how it is used: `$this->render(false, $layout);`
https://github.com/cakephp/cakephp/blob/2.x/lib/Cake/Controller/Controller.php#L1017

And passed unaltered from Controller to View here - https://github.com/cakephp/cakephp/blob/2.x/lib/Cake/Controller/Controller.php#L963.

lib/Cake/basics.php
bancer: It is not clear what you mean by that. The fourth parameter to `htmlspecialchars` below must be a boolean. So, `$double` should be converted to a boolean as it is intended according to the method signature `function h($text, $double = true, $charset = null)`.

lib/Cake/Controller/Controller.php
```diff
@@ -45,14 +45,21 @@
...
  * @property      AclComponent $Acl
+ * @property      string $action The action handling the current request. Deprecated, use CakeRequest::$action instead.
  * @property      AuthComponent $Auth
+ * @property      string $base Base URL path. Deprecated, use CakeRequest::$base instead.
  * @property      CookieComponent $Cookie
+ * @property      array $data POST data. Deprecated, use CakeRequest::$data instead.
  * @property      EmailComponent $Email
  * @property      PaginatorComponent $Paginator
+ * @property      array $params Array of parameters parsed from the URL. Deprecated, use CakeRequest::$params instead.
+ * @property      string $here The full address to the current request. Deprecated, use CakeRequest::$here instead.
```
garas: Sort properties alphabetically (`$here` and `$Flash` is out of order). And maybe list components first (also it will look better as they has no description here), then your additions.

lib/Cake/basics.php
saeideng: I meant you should fix it first in `master`

lib/Cake/View/View.php
saeideng: OK, so just `string|false` is good
and no needed to `!is_bool($view) &&`  (line 470)

lib/Cake/basics.php
bancer: https://github.com/cakephp/cakephp/pull/12719

bancer:
May this be merged already?

PR #12720, opened 2018-11-11
`getMockForModel()` method can take `null` for `$methods` parameter

mirko-pagliai:
I can not really understand why the `getMockForModel()` method should not be able to accept a `null` value as `$methods` parameter.

Yes, I realize that for many it will seem useless. But the `setMethods()` method can accept that value, I do not see why it should not be so here too.

### DIFF
--- a/src/TestSuite/TestCase.php
+++ b/src/TestSuite/TestCase.php
@@ -673,12 +673,12 @@ abstract class TestCase extends BaseTestCase
      * Mock a model, maintain fixtures and table association
      *
      * @param string $alias The model to get a mock for.
-     * @param array $methods The list of methods to mock
+     * @param array|null $methods The list of methods to mock
      * @param array $options The config data for the mock's constructor.
      * @throws \Cake\ORM\Exception\MissingTableClassException
      * @return \Cake\ORM\Table|\PHPUnit_Framework_MockObject_MockObject
      */
-    public function getMockForModel($alias, array $methods = [], array $options = [])
+    public function getMockForModel($alias, $methods = [], array $options = [])
     {
         /** @var \Cake\ORM\Table $className */
         $className = $this->_getTableClassName($alias, $options);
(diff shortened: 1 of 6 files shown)


src/TestSuite/TestCase.php
```diff
@@ -673,12 +673,12 @@ protected function skipUnless($condition, $message = '')
...
      *
      * @param string $alias The model to get a mock for.
-     * @param array $methods The list of methods to mock
+     * @param array|null $methods The list of methods to mock
      * @param array $options The config data for the mock's constructor.
      * @throws \Cake\ORM\Exception\MissingTableClassException
      * @return \Cake\ORM\Table|\PHPUnit_Framework_MockObject_MockObject
      */
-    public function getMockForModel($alias, array $methods = [], array $options = [])
+    public function getMockForModel($alias, $methods = [], array $options = [])
```
dereuromark: removing a typehint sounds wrong - why not `array $methods = null`?

dereuromark: We usually do not allow an additional type `null` where not necessary, `[]` is the same and better in these cases.
So I would suggest we do not change this here for no reason.

markstory (commented):
I also don't understand the reasoning behind this. `[]` is less typing than null and IDE autocompletion should work here.

src/TestSuite/TestCase.php
mirko-pagliai: Sorry, but they are completely different. With an empty array all methods are stubs and all return `null` by default; with `null`, all methods are mocks and they run the original code contained within the method when called.

mirko-pagliai:
https://github.com/sebastianbergmann/phpunit-mock-objects/blob/master/src/MockBuilder.php#L185

`public function setMethods(array $methods = null)`

Note that for phpunit "default is to mock none of them"

dereuromark (approved):

mirko-pagliai:
I think that for now you can choose my proposal, because it does not produce any change.

    public function getMockForModel($alias, $methods = [], array $options = [])

In fact the default value of the argument is the same. I would not be wrong, but it should not change anything.

Instead, in a next release, it would be correct that the signature of the method matches that of phpunit, and then:

    public function getMockForModel($alias, array $methods = null, array $options = [])

PR #12723, opened 2018-11-13
Fix route reset and preserve extensions

markstory:
I wasn't entirely happy with the solution in #12695 and wanted to try an alternate approach. This set of changes adds state to fewer places and doesn't require plugins to opt-in to backwards compatible behavior. Instead it introduces new internal methods to reset router state that can cause conflicts/duplicate work. Global extensions and default route classes are not reset.

I've targeted this for 3.7 as it adds new methods, but I'm open to shipping it in 3.6 if people would like that more.

Refs #12695
Refs #12577

### DIFF
--- a/src/Routing/Router.php
+++ b/src/Routing/Router.php
@@ -515,6 +515,30 @@ class Router
     }
 
     /**
+     * Reset routes and related state.
+     *
+     * Similar to reload() except that this doesn't reset all global state,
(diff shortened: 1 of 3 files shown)


dereuromark:
See, I knew you would come up with a better solution ;)

So https://github.com/dereuromark/cakephp-sandbox/pull/20 looks like this works.

dereuromark (approved):

Issue #12725, opened 2018-11-13
RFC: object variables in 4.x not ucfirst anymore


dereuromark:
This is a (multiple allowed):

* [x] feature-discussion (RFC)

* CakePHP Version: 4.0

### Status Quo

All over in code

    public function __construct(View $View, array $config)

or

    $this->View = new ViewHelper();

### What people expect

This is somewhat a relict from 1.x and 2.x days and does not reflect well on the code basis for outside people looking into it.
They expect those to be consistent with the status quo other projects use.

That's how it should look like:

    public function __construct(View $view, array $config)

or

    $this->view = new ViewHelper();

This is especially inconsistent as some new code already follows the camelBack convention.

#### Scoped method variables (1)
Easy to rename, using tools like IDEs this is fully supported without the possibility of breaking anything and fully BC.

#### Class properties (2)
Less easy, as this can break BC. So only to be done in 4.0 or 5.0.
This can also be done using IDE support and therefore also be not too big of a deal.

We can do test cases already, as they dont have the BC dependency.

The actual code base should ideally not have too many public properties leaking outside anyway, so this is mainly an issue for extending code that accesses these properties.
Here we can maybe also wait till 5.0 to get rid of the baggage/inconsistency.

dereuromark:
That said, I am also fine if we keep the magic methods (part 2) in the current format - if that is what we all want to keep :)
I was mainly talking about the variable names themselves. I opened a PR for that.

ADmad:
I would preferring keeping properties for helpers, components, tables CamelCased since they match the class name and avoid inflection.

We have already removed a lot of public properties in 4.0. Protected properties like `Helper::$_View` could be changed but it would be better to defer that to 5.0.

PR #12726, opened 2018-11-13
Resolves the first part of object variable naming cleanup.

dereuromark:
This implements the first part of the RFC https://github.com/cakephp/cakephp/issues/12725 for a cleaner core code base.

I used IDE phpstorm refactoring, so no chance for BC breaks or missing parts here.
All inside of a method scope is safe.

Rest of the linked issue still remains open for further processing.


markstory (approved):
Looks good to me.

Issue #12730, opened 2018-11-16
isMobile() method of ServerRequest returns false


davalb:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 853b691a52a6653d6b8c2c3589fe29c622cfa94b
* Platform and Target: both apache and nginx

This is the HTTP_USER_AGENT string from my mobile: `Mozilla/5.0 (Linux; Android 8.1.0; Nexus 5X) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/70.0.3538.80 Mobile Safari/537.36`

I would expect the isMobile() function to return true for that.

I will provide a test case shortly.

garas:
Mobile detection provided by [mobiledetect/mobiledetectlib](https://github.com/serbanghita/Mobile-Detect) composer package. Make sure you have latest version.

davalb:
Oh, thank you. The app was missing that library (because it was migrated from 2.x to 3.x and we forgot adding the library).
Should the isMobile() method throw an error, if this library is missing?

PR #12731, opened 2018-11-16
bug 12730 - testcase for IsMobile with specific user agent

davalb:
This is the testcase for the ticket #12730 
I don't know what the issue is here, is the string for my user agent weird? Or is there a bug. If it is a bug, I would be happy to provide a fix, but could somebody point me in the right direction please. Where is the detection code located?


davalb:
I am closing this PR, it doesnt make sense, since mobiledetect is used to detect mobile user agenets.

PR #12732, opened 2018-11-16
Add Typings to some @methods in Query

Iandenh:
I also saw that there are some missing `@method` for collection methods. Shall I add them?

### DIFF
--- a/src/ORM/Query.php
+++ b/src/ORM/Query.php
@@ -37,7 +37,7 @@ use Traversable;
  *
  * @see \Cake\Collection\CollectionInterface For a full description of the collection methods supported by this class
  * @method \Cake\Collection\CollectionInterface each(callable $c) Passes each of the query results to the callable
- * @method \Cake\Collection\CollectionInterface sortBy($callback, $dir = SORT_DESC, $type = \SORT_NUMERIC) Sorts the query with the callback
+ * @method \Cake\Collection\CollectionInterface sortBy($callback, int $dir = SORT_DESC, int $type = \SORT_NUMERIC) Sorts the query with the callback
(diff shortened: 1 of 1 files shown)


src/ORM/Query.php
```diff
@@ -37,23 +37,23 @@
...
  * @method mixed reduce(callable $c, $zero = null) Folds all the results into a single value using the callable.
  * @method \Cake\Collection\CollectionInterface extract($field) Extracts a single column from each row
- * @method mixed max($field, $type = SORT_NUMERIC) Returns the maximum value for a single column in all the results.
- * @method mixed min($field, $type = SORT_NUMERIC) Returns the minimum value for a single column in all the results.
+ * @method mixed max($field, int $type = SORT_NUMERIC) Returns the maximum value for a single column in all the results.
+ * @method mixed min($field, int $type = SORT_NUMERIC) Returns the minimum value for a single column in all the results.
  * @method \Cake\Collection\CollectionInterface groupBy(string|callable $field) In-memory group all results by the value of a column.
  * @method \Cake\Collection\CollectionInterface indexBy(string|callable $field) Returns the results indexed by the value of a column.
- * @method int countBy(string|callable $field) Returns the number of unique values for a column
+ * @method \Cake\Collection\CollectionInterface countBy(string|callable $field) Returns the number of unique values for a column
```
markstory: Doesn't countBy return an int?

dereuromark: As weird as it might sound: Definitely a collection object
https://github.com/cakephp/cakephp/blob/master/src/Collection/CollectionInterface.php#L501
https://github.com/cakephp/cakephp/blob/master/src/Collection/CollectionTrait.php#L281

PR #12733, opened 2018-11-16
Fix return type in of `countBy` @method annotation in Query

Iandenh:
A return fix I noticed when I looked at the `@method`s for #12732


dereuromark (approved):

markstory (commented):
Are you sure this is correct? I was pretty sure countby returns an integer.

dereuromark:
As weird as it might sound: Definitely a collection object
https://github.com/cakephp/cakephp/blob/master/src/Collection/CollectionInterface.php#L501
https://github.com/cakephp/cakephp/blob/master/src/Collection/CollectionTrait.php#L281

PR #12735, opened 2018-11-17
Dasherize path elements by default instead of underscoring.

ADmad:
This creates more consistent URLs by default.

### DIFF
--- a/src/Routing/RouteBuilder.php
+++ b/src/Routing/RouteBuilder.php
@@ -281,7 +281,7 @@ class RouteBuilder
      * });
      * ```
      *
-     * Plugins will create lower_case underscored resource routes. e.g
+     * Plugins will create lowercase dasherized resource routes. e.g
      * `/comments/comments`
      *
      * Connect resource routes for the Articles controller in the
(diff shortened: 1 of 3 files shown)


src/Routing/RouteBuilder.php
```diff
@@ -354,7 +354,7 @@ public function resources(string $name, $options = [], $callback = null): void
         }
         $options += [
             'connectOptions' => [],
-            'inflect' => 'underscore',
+            'inflect' => 'dasherize',
```
dereuromark: I wonder if we should use class const for these things. Easy to make typos like `dasherise` etc by accident.

ADmad: If someone makes a typo then tests will fail :slightly_smiling_face:. Also I don't think this is going to change again in the foreseeable future.

dereuromark: I am also arguing more from the type of usability from the user/dev.
Setting the routes up - there are no tests, but the result silently might be different.
Too many magic strings usually are a bad thing - at least it would give devs the option to opt-out of it if they want to ;)

markstory (approved):
Seems ok. We will have to document the changes default and how people can retain their existing URL inflections.

PR #12736, opened 2018-11-17
Improve documentation and code style

bancer:

### DIFF
--- a/lib/Cake/Network/CakeResponse.php
+++ b/lib/Cake/Network/CakeResponse.php
@@ -1318,9 +1318,6 @@ class CakeResponse {
 				continue;
 			}
 			$original = $domain;
-			if (strpos($domain, '://') === false) {
-				$preg = ($requestIsSSL ? 'https://' : 'http://') . $domain;
-			}
 			$preg = '@' . str_replace('*', '.*', $domain) . '@';
 			$result[] = compact('original', 'preg');
 		}
--- a/lib/Cake/Network/CakeResponse.php
+++ b/lib/Cake/Network/CakeResponse.php
@@ -374,7 +374,7 @@ class CakeResponse {
  * Holds all the cache directives that will be converted
  * into headers when sending the request
  *
- * @var string
+ * @var array
  */
 	protected $_cacheDirectives = array();
 
@@ -672,8 +672,9 @@ class CakeResponse {
  *
  *        For more on HTTP status codes see: http://www.w3.org/Protocols/rfc2616/rfc2616-sec6.html#sec6.1
  *
- * @return mixed associative array of the HTTP codes as keys, and the message
- *    strings as values, or null of the given $code does not exist.
+ * @return array|null|true associative array of the HTTP codes as keys, and the message
+ *    strings as values, or null of the given $code does not exist. `true` if `$code` is
+ *    an array of valid codes.
  * @throws CakeException If an attempt is made to add an invalid status code
  */
 	public function httpCodes($code = null) {
@@ -717,7 +718,7 @@ class CakeResponse {
  *
  * e.g `type(array('jpg' => 'text/plain'));`
  *
- * @param string $contentType Content type key.
+ * @param array|string $contentType Content type key.
  * @return mixed current content type or false if supplied an invalid content type
  */
 	public function type($contentType = null) {
@@ -854,7 +855,7 @@ class CakeResponse {
 		}
 
 		$this->maxAge($time);
-		if (!$time) {
+		if ($time === null) {
 			$this->_setCacheControl();
 		}
 		return (bool)$public;
@@ -1160,7 +1161,11 @@ class CakeResponse {
  * @return bool whether the response was marked as not modified or not.
  */
 	public function checkNotModified(CakeRequest $request) {
-		$etags = preg_split('/\s*,\s*/', $request->header('If-None-Match'), null, PREG_SPLIT_NO_EMPTY);
+		$ifNoneMatchHeader = $request->header('If-None-Match');
+		$etags = array();
+		if (is_string($ifNoneMatchHeader)) {
+			$etags = preg_split('/\s*,\s*/', $ifNoneMatchHeader, null, PREG_SPLIT_NO_EMPTY);
+		}
 		$modifiedSince = $request->header('If-Modified-Since');
 		$checks = array();
 		if ($responseTag = $this->etag()) {
@@ -1224,7 +1229,7 @@ class CakeResponse {
  *
  * `$this->cookie((array) $options)`
  *
- * @param array $options Either null to get all cookies, string for a specific cookie
+ * @param array|string $options Either null to get all cookies, string for a specific cookie
  *  or array to set cookie.
  * @return mixed
  */
@@ -1312,8 +1317,7 @@ class CakeResponse {
 				$result[] = array('preg' => '@.@', 'original' => '*');
 				continue;
 			}
-
-			$original = $preg = $domain;
+			$original = $domain;
 			if (strpos($domain, '://') === false) {
 				$preg = ($requestIsSSL ? 'https://' : 'http://') . $domain;
 			}
@@ -1464,7 +1468,7 @@ class CakeResponse {
 		$file->open('rb');
 
 		$end = $start = false;
-		if ($range) {
+		if (!empty($range)) {
 			list($start, $end) = $range;
 		}
 		if ($start !== false) {
--- a/lib/Cake/Network/CakeResponse.php
+++ b/lib/Cake/Network/CakeResponse.php
@@ -331,7 +331,7 @@ class CakeResponse {
  * Content type to send. This can be an 'extension' that will be transformed using the $_mimetypes array
  * or a complete mime-type
  *
- * @var int
+ * @var string
  */
 	protected $_contentType = 'text/html';
 
@@ -718,8 +718,8 @@ class CakeResponse {
  *
  * e.g `type(array('jpg' => 'text/plain'));`
  *
- * @param array|string $contentType Content type key.
- * @return mixed current content type or false if supplied an invalid content type
+ * @param array|string|null $contentType Content type key.
+ * @return string|false current content type or false if supplied an invalid content type
  */
 	public function type($contentType = null) {
 		if ($contentType === null) {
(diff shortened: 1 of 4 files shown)


lib/Cake/Network/CakeResponse.php
```diff
@@ -854,7 +855,7 @@ public function sharable($public = null, $time = null) {
 		}
 
 		$this->maxAge($time);
-		if (!$time) {
+		if ($time === null) {
```
dereuromark: What about 0 or false or other fasely values?
By changing this here you possibly introduce dangerous BC breaks for existing apps.
Better not changing functionality in a maintenance release major anymore where not absolutely necessary. Here it doesnt seem like it.

lib/Cake/Network/CakeResponse.php
```diff
@@ -1464,7 +1465,7 @@ protected function _sendFile($file, $range) {
 		$file->open('rb');
 
 		$end = $start = false;
-		if ($range) {
+		if (!empty($range)) {
```
dereuromark: the previous was 
a) more correct as it doesn't cloak
b) otherwise completely identical

please see cakephp coding guidlines for this.

lib/Cake/Network/CakeResponse.php
```diff
@@ -717,7 +718,7 @@ public function httpCodes($code = null) {
  *
  * e.g `type(array('jpg' => 'text/plain'));`
  *
- * @param string $contentType Content type key.
+ * @param array|string $contentType Content type key.
```
dereuromark: `|null` missing if you want to be correct/precise.

lib/Cake/Network/CakeResponse.php
```diff
@@ -717,7 +718,7 @@ public function httpCodes($code = null) {
...
  * @return mixed current content type or false if supplied an invalid content type
```
dereuromark: string|false then?

lib/Cake/Network/CakeResponse.php
bancer: > The expression $time of type integer|null is loosely compared to false; this is ambiguous if the integer can be 0. You might want to explicitly use === null instead.

The code intention was not very clear here. Perhaps `if (empty($time)) {` would be better.

dereuromark: No, "if empty" is also bad (and functionally equal to what was before), see my other message.
`if (!$time)` is perfect to be kept.

lib/Cake/Network/CakeResponse.php
bancer: > The expression $range of type array is implicitly converted to a boolean; are you sure this is intended? If so, consider using ! empty($expr) instead to make it clear that you intend to check for an array without elements. 

According to CakePHP [coding standards](https://book.cakephp.org/2.0/en/contributing/cakephp-coding-conventions.html#comparison) regarding comparison:
> Always try to be as strict as possible.

lib/Cake/Network/CakeResponse.php
bancer: The intention of the check becomes ambiguous when an interger is converted to a boolean. Another option would be to cast to an integer `if ((int)$time === 0) {` or to set the default value to 0 instead of `null` what I think makes more sense as `$time` is supposed to be an integer: 
`public function sharable($public = null, $time = 0) {`.

`if (!$time) {` check is also inconsistent with strict check `if ($seconds !== null) {` done in `maxAge()` method that is called on line 857.

lib/Cake/Network/CakeResponse.php
dereuromark: You didnt fix strictness here, you just did the opposite. You cloaked a var, which is bad.
Read the part about not using empty() and isset() wrongly.

lib/Cake/Network/CakeResponse.php
dereuromark: If that is the case, then casting sure would be more BC here, people could rely on falsely values to also trigger the ! part.

lib/Cake/Network/CakeResponse.php
bancer: Could you be more specific in referencing to the guidelines? I have not found any mention of `empty` on that page. I do not fully understand how making the intention of the code clearer makes it worse.

dereuromark: It is pretty highlighted on the page actually:
https://book.cakephp.org/3.0/en/contributing/cakephp-coding-conventions.html#careful-when-using-empty-isset

You may also read [this](https://www.dereuromark.de/2016/06/01/use-isset-and-empty-wisely/).

bancer: > When working with variables, it is better to rely on type-coercion to boolean instead of empty().

That statement is highly arguable. In my opinion it is quite confusing and is in fact buggy when a string or an array is converted to a boolean for comparison that way. 

In this particular case `list()` that is called below expects `$range` to be an array. If `$range` is `true` or `1` (and this may happen if `true` or `1` is passed as the parameter of `_sendFile` method) this code makes no sense:
```
list($start, $end) = true;
```
The values of `$start` and `$end` are `null` in this case. 
```
$end = $start = false;
var_export($start); // outputs false
var_export($end);   // outputs false
list($start, $end) = true;
var_export($start); // outputs NULL
var_export($end);   // outputs NULL
```
And that seems to be not what is expected here. The code inside `if ($start !== false) {` below would not be executed.

So the correct check would be `if (is_array($range) && !empty($range)) {` here.

dereuromark: Let's please stop that discussion.
You are wrong here in terms of functionality. It does not make a difference, those statements (before and after your change) are identical. The downside of cloaking is explained in both linked resources. This has nothing to do with casting.
Please revert this and focus on the actual changes that matter.

lib/Cake/Network/CakeResponse.php
```diff
@@ -1464,7 +1465,7 @@ protected function _sendFile($file, $range) {
 		$file->open('rb');
 
 		$end = $start = false;
-		if ($range) {
+		if (is_array($range) && !empty($range)) {
```
dereuromark: You are changing behavior again:
If the range is not an array, the whole functionality here differs now instead of blowing up as it should.

As discusses, please revert fully.

dereuromark: Again: 2.x is in maintenance mode, security and critical bugfixes only.
No functional changes for the sake of changing anymore, please, to avoid further accidental regressions.

lib/Cake/Network/CakeResponse.php
bancer: I've got an impression that you did not fully read my previous post :(.

dereuromark: I did - but you are then not understanding how PHP works. There is no casting happening.
The $range is still what it is, even after such a check. http://sandbox.onlinephpfunctions.com/code/580d6dbc977dd40112d36dcb8976662cc9d6e0bf

bancer: I understand that no casting happens and the value of `$range` does not change. That's why I wrote "and this may happen if true or 1 is passed as the parameter of _sendFile method". The values of `$start` and `$end` can be null in some scenarios. http://sandbox.onlinephpfunctions.com/code/6623b50ebc865963924eb4fa8730c38bd2d4d922

dereuromark: Is there a test case to proof this edge case?
Otherwise the change below looks too dangerous.

bancer: If there would be a unit test for this scenario the bug would not be in the code. Why the code in http://sandbox.onlinephpfunctions.com/code/6623b50ebc865963924eb4fa8730c38bd2d4d922 is not convincing?

bancer: Maybe [this](http://sandbox.onlinephpfunctions.com/code/989185d2a0747f4cf2dbc15714b567d1b3a5ccd1) would be more convincing?

lib/Cake/Network/CakeResponse.php
dereuromark: Use `if ($range && is_array($range)) {` then to fit the guidelines.

dereuromark: You are still wrong about what it means: in PHP it doesnt mean true, but not empty in the context of arrays. Your code is cloaking which is 10x worse. Let's please stop this discussion.
The more important question is: Why the array check now? What happens if it is not an array?
Then it would silently continue doing sth different then it did before.

bancer: > in PHP it doesnt mean true, but not empty in the context of arrays.

I think you are wrong about that. From PHP documentation:

> As described in the section about expressions, expression **is evaluated to its Boolean value**.
http://php.net/manual/en/control-structures.if.php

> When necessary, an expression is **automatically converted to boolean**.
http://php.net/manual/en/language.expressions.php

> To explicitly convert a value to boolean, use the (bool) or (boolean) casts. However, in most cases the cast is unnecessary, since a value will be **automatically converted** if an operator, function or control structure requires a boolean argument.
http://php.net/manual/en/language.types.boolean.php#language.types.boolean.casting

So event when `$var` is an array `if ($var)` is evaluated as `if var is true` according to PHP documentation. `if ($var)` is equivalent to `if ($var == true)` - non-strict comparison to a boolean.

---
> The more important question is: Why the array check now? What happens if it is not an array?

If `$range` is not an array but `1` or `true` then `$start` and `$end` are assigned `null` values although `false` values are expected in such scenario - take a look [here](http://sandbox.onlinephpfunctions.com/code/6623b50ebc865963924eb4fa8730c38bd2d4d922) again and compare it with the [correct](http://sandbox.onlinephpfunctions.com/code/f1307e1bdb883a91cb6a741342cc20a535ba8e66) implementation please.

dereuromark: Then `if ($range && is_array($range)) {` is the correct code here.

bancer: Okay, but when you claimed that `in PHP it doesnt mean true` you de-railed the discussion to PHP level yourself.

ADmad: TLDR:  Don't make any changes that can potentially change the way code behaves. 2.x is **done**.

dereuromark (approved):

PR #12741, opened 2018-11-19
3.next - Add ArrayCache engine

markstory:
This cache engine gives us a cache implementation that is ideal for application test suites. Previously folks would have to use peristant data stores that could leave artifacts behind requiring manual cleanup or use NullEngine which provides none of the benefits that caching offers. Now folks can use this in memory cache that offers expiration and group support.

### DIFF
--- a/src/Cache/Engine/ArrayEngine.php
+++ b/src/Cache/Engine/ArrayEngine.php
@@ -1,0 +1,183 @@
+<?php
+/**
+ * CakePHP(tm) : Rapid Development Framework (https://cakephp.org)
+ * Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
+ *
+ * Licensed under The MIT License
+ * For full copyright and license information, please see the LICENSE.txt
+ * Redistributions of files must retain the above copyright notice.
+ *
+ * @copyright     Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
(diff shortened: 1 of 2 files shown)


src/Cache/Engine/ArrayEngine.php
```diff
@@ -0,0 +1,183 @@
...
+     */
+    public function increment($key, $offset = 1)
+    {
+        if (!$this->read($key)) {
+            $this->write($key, 0);
+        }
+        $key = $this->_key($key);
+        $this->data[$key][1] += $offset;
+
+        return true;
```
ADmad: Return value doesn't match docblock, same for decrement().

src/Cache/Engine/ArrayEngine.php
```diff
@@ -0,0 +1,183 @@
...
+ * or console tools where you don't want the overhead of interacting
+ * with a cache servers, but want the work saving properties a cache
+ * provides.
+ */
+class ArrayEngine extends CacheEngine
+{
+    /**
+     * Cached data.
+     *
+     * Structured as [key => [expiration, value]]
```
ADmad: Would using an associative instead of numerically index array for `[expiration, value]` have any drawbacks? It would make the code a bit more readable.

markstory: It would use a bit more memory, but otherwise they would behave the same.

ADmad (approved):

PR #12744, opened 2018-11-20
3.7 client response

ADmad:
This deprecates the unnecessary use of magic `__get()` and `__isset()` for providing public properties. Instead all are available as public methods now.


ADmad:
The `Response::body($parser = null)` method should also be deprecated IMO.

src/Http/Client/Response.php
```diff
@@ -148,6 +133,20 @@ class Response extends Message implements ResponseInterface
...
 
+    /**
+     * Map of deprecated magic properties.
+     *
+     * @var array
+     * @internal
+     */
+    protected $_deprecatedMagicProperties = [
+        'cookies' => 'getCookies()',
+        'body' => 'getBody()->getContents()',
```
saeideng: `getBody()` does not exist

src/Http/Client/Response.php
```diff
@@ -60,11 +52,10 @@
  * $content = $response->getBody();
  * ```
  *
- * You can also use object access to get the string version
- * of the response body:
+ * You can get the body string using:
  *
  * ```
- * $content = $response->body;
+ * $content = $response->getBody()->getContents();
```
markstory: This feels pretty clunky. I know its part of the spec, but I don't personally relish having to type that much to read a response's contents.

ADmad: We can add a `getStringBody()` method.

src/Http/Client/Response.php
```diff
@@ -557,9 +556,14 @@ public function version()
      * @param callable|null $parser The callback to use to decode
      *   the response body.
      * @return mixed The response body.
+     * @deprecated 3.7.0 Use getStringBody()/getJson()/getXml() instead.
      */
     public function body($parser = null)
     {
+        deprecationWarning(
+            'Response::body() is deprecated. Use getStringBody()/getJson()/getXml() instead.'
+        );
```
markstory: How should people replace parsed bodies? Should we have `getParsedBody($func)`?

ADmad: Do we really need the ability to pass parsers as argument? New methods are already added to getting JSON and XML parses content which are most common types.

People can just pass the return value of `$response->getStringBody()` to their custom parser functions instead.

markstory: > People can just pass the return value of $response->getStringBody() to their custom parser functions instead.

That's fair.

src/Http/Client/Response.php
```diff
@@ -148,6 +133,20 @@ class Response extends Message implements ResponseInterface
...
     ];
 
+    /**
+     * Map of deprecated magic properties.
+     *
+     * @var array
+     * @internal
+     */
+    protected $_deprecatedMagicProperties = [
+        'cookies' => 'getCookies()',
```
dereuromark: For consistency, we could also not add () here, but add it when displaying it.
Then this could also be used more programmatically for example.

Issue #12743, opened 2018-11-19
3.6.13 Bug in paginator sort


dereuromark:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.6 latest

### What you did
Multiple of my apps contain a small bug in pagination sorting order.

### What happened
```php
	/**
	 * @var array
	 */
	public $paginate = [
		'order' => ['Logs.created' => 'DESC'],
	];
```
And normal pagination using helper

The sort is working by default (URL `/logs` with correct order and correct pagination link and icon).

But when clicking on the link, even though the URL says `/logs?sort=created&direction=asc`, the actual order always stays "created DESC"

### What you expected to happen
I expect to be able to sort even with the default order set.
Please note that I am not using any whitelist.

I think this is some calculation issue somewhere due to the added model prefix.
As soon as I remove the alias in the above paginate default order statement, it works again.
But since it worked before, and also shows no error, we should remove this regression and allow the above setup to work out of the box.

dereuromark:
I will look into it today.

dereuromark:
As I suspected, the normalization doesnt happen and causes the issue:
```php
// http://sandbox.local/data/countries
[
	'Countries' => [
		'finder' => 'all',
		'page' => (int) 1,
		'current' => (int) 20,
		'count' => (int) 241,
		'perPage' => (int) 20,
		'start' => (int) 1,
		'end' => (int) 20,
		'prevPage' => false,
		'nextPage' => true,
		'pageCount' => (int) 13,
		'sort' => 'Countries.sort',
		'direction' => 'DESC',
		'limit' => null,
		'sortDefault' => 'Countries.sort',
		'directionDefault' => 'DESC',
		'scope' => null,
		'completeSort' => [
			'Countries.sort' => 'DESC'
		]
	]
]

// http://sandbox.local/data/countries?sort=sort&direction=asc
[
	'Countries' => [
		'finder' => 'all',
		'page' => (int) 1,
		'current' => (int) 20,
		'count' => (int) 241,
		'perPage' => (int) 20,
		'start' => (int) 1,
		'end' => (int) 20,
		'prevPage' => false,
		'nextPage' => true,
		'pageCount' => (int) 13,
		'sort' => 'sort',
		'direction' => 'DESC',
		'limit' => null,
		'sortDefault' => 'Countries.sort',
		'directionDefault' => 'DESC',
		'scope' => null,
[...]

Issue #10028, opened 2017-01-13
Pagination: field names with `order` and `sortWhitelist`


mirko-pagliai:
I need some clarification for pagination, using models that have associations and a custom query.

Example: I'm paginating posts. Posts belongs to posts categories, belongs to users and have many tags. Posts and users have the `created` field.

Code:

```
$query = $this->Posts->find()->contain([
    'Categories' => function ($q) {
        return $q->select(['id', 'title']);
    },
    'Tags' => function ($q) {
        return $q->order(['tag' => 'ASC']);
    },
    'Users' => function ($q) {
        return $q->select(['id', 'first_name', 'last_name']);
    },
]);

$this->paginate['order'] = ['created' => 'DESC'];
$this->paginate['sortWhitelist'] = ['id', 'title', 'Categories.title', 'Users.first_name', 'priority', 'created'];

$posts = $this->paginate($query);

$this->set(compact('posts'));
```

What happens:
- opening the view, the records are sorted correctly (`created` `DESC`);
- on the first page, I can sort by any field set in `sortWhitelist`;
- the second page **only works** for `Categories.title` and `Users.first_name` fields. Other fields **do not work** because the query string adds the name of the model.

Example. On the first page, sorting by title, this is the url (and it works):
`http://localhost/mirkopagliai/admin?sort=title&direction=asc`  
Clicking on the second page, this is the url (and it doesn't work):
`http://localhost/mirkopagliai/admin?page=2&sort=Posts.title&direction=asc`  
And pagination breaks.

Changing the code (I added the name of the model to the `title` field):
```
$this->paginate['order'] = ['created' => 'DESC'];
$this->paginate['sortWhitelist'] = ['id', 'Posts.title', 'Categories.title', 'Users.first_name', 'priority', 'created'];
```

What happens:
- opening the view, the records are sorted correctly (`created` `DESC`);
- on the first page, **I can't sort** anything;
- the other pages **do not work** under any circumstances.

Changing the code (I added the name of the model to the `created` field, on `order` and `sortWhitelist` options):
```
$this->paginate['order'] = ['Posts.created' => 'DESC'];
$this->paginate['sortWhitelist'] = ['id', 'title', 'Categories.title', 'Users.first_name', 'priority', 'Posts.created'];
```

What happens:
- opening the view, the records are sorted correctly (`created` `DESC`);
- on the first page, I can sort by any field set in `sortWhitelist`;
[...]

PR #12745, opened 2018-11-20
Fix paginator for falling back to default sort.

dereuromark:
Resolves https://github.com/cakephp/cakephp/issues/12743

They main issue is around

	/**
	 * @var array
	 */
	public $paginate = ['order' => ['Countries.sort' => 'DESC']];

and

    <?php echo $this->Paginator->sort('sort') ?>

where the aliasing silently makes this bug appear right now.

Check out live bug @ https://sandbox.dereuromark.de/data/countries
Click on the "sort" filter, see the URL generated, but never applied.
This has been in Paginator since 3.5 or 3.6 very early it seems (probably around Paginator extract).

Paginator fixes the actual bug around the query adjustment.
The PaginationHelper fix then re-adjusts the URLs for the primary model and non-aliases default columns used.
So that with default order "sort DESC" the URLs generated and clicked are:
- `/countries`
- `/countries?sort=sort&direction=asc`
- `/countries` again

Any possible side effects here?
How can we test this, as the tests do not include the options, only params?

Additional adjustment:
directionDefault is now also consolidated to lowercase version when passed on.

### DIFF
--- a/src/Datasource/Paginator.php
+++ b/src/Datasource/Paginator.php
@@ -224,7 +224,7 @@ class Paginator implements PaginatorInterface
             'direction' => isset($options['sort']) ? current($order) : null,
             'limit' => $defaults['limit'] != $limit ? $limit : null,
             'sortDefault' => $sortDefault,
-            'directionDefault' => $directionDefault ? strtolower($directionDefault) : $directionDefault,
+            'directionDefault' => $directionDefault,
             'scope' => $options['scope'],
             'completeSort' => $order,
         ];
@@ -424,12 +424,12 @@ class Paginator implements PaginatorInterface
 
             list ($alias, $currentField) = explode('.', $field);
 
-            if ($alias !== $model) {
-                $result[$field] = $sort;
+            if ($alias === $model) {
+                $result[$currentField] = $sort;
                 continue;
             }
 
-            $result[$currentField] = $sort;
+            $result[$field] = $sort;
         }
 
         return $result;
--- a/src/Datasource/Paginator.php
+++ b/src/Datasource/Paginator.php
@@ -224,7 +224,7 @@ class Paginator implements PaginatorInterface
             'direction' => isset($options['sort']) ? current($order) : null,
             'limit' => $defaults['limit'] != $limit ? $limit : null,
             'sortDefault' => $sortDefault,
-            'directionDefault' => $directionDefault,
+            'directionDefault' => $directionDefault ? strtolower($directionDefault) : $directionDefault,
             'scope' => $options['scope'],
             'completeSort' => $order,
         ];
@@ -363,7 +363,12 @@ class Paginator implements PaginatorInterface
             if (!in_array($direction, ['asc', 'desc'])) {
                 $direction = 'asc';
             }
+
             $order = (isset($options['order']) && is_array($options['order'])) ? $options['order'] : [];
+            if ($order && $options['sort'] && strpos($options['sort'], '.') === false) {
+                $order = $this->_removeAliases($order, $object->getAlias());
+            }
+
             $options['order'] = [$options['sort'] => $direction] + $order;
         } else {
             $options['sort'] = null;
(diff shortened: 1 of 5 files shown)


src/View/Helper/PaginatorHelper.php
```diff
@@ -524,6 +524,13 @@ public function generateUrlParams(array $options = [], $model = null)
         $paging = $this->params($model);
         $paging += ['page' => null, 'sort' => null, 'direction' => null, 'limit' => null];
 
+        if (!empty($paging['sort']) && !empty($options['sort']) && strpos($options['sort'], '.') === false) {
+            $paging['sort'] = $this->_removeAlias($paging['sort'], $model = null);
```
markstory: You're assigning to `model` here which creates a side-effect inside a method call.

src/Datasource/Paginator.php
```diff
@@ -401,6 +406,35 @@ public function validateSort(RepositoryInterface $object, array $options)
...
+        $result = [];
+        foreach ($fields as $field => $sort) {
+            if (strpos($field, '.') === false) {
+                $result[$field] = $sort;
+                continue;
+            }
+
+            list ($alias, $currentField) = explode('.', $field);
+
+            if ($alias !== $model) {
```
markstory: Why is this version `!==` but the one in the helper uses `===`?

tests/TestCase/Datasource/PaginatorTest.php
```diff
@@ -321,7 +321,7 @@ public function testDefaultPaginateParamsIntoRequest()
         $this->Paginator->paginate($table, [], $settings);
         $pagingParams = $this->Paginator->getPagingParams();
         $this->assertEquals('PaginatorPosts.id', $pagingParams['PaginatorPosts']['sortDefault']);
-        $this->assertEquals('DESC', $pagingParams['PaginatorPosts']['directionDefault']);
+        $this->assertEquals('desc', $pagingParams['PaginatorPosts']['directionDefault']);
```
markstory: This could break userland code. It isn't very likely but it could happen.

markstory (commented):
Why does your paginate order value and HTML sort input use different key/column names? While we could infer which model name like this, it is simpler for us and less magical for users if all the keys used need to match.

src/Datasource/Paginator.php
dereuromark: This was just a bail early here, I can invert the logic to be in sync.

markstory:
> The main issue is the silent breaking here.

Sort conditions are supposed to silently fail when they don't match up.

This appears to be caused by the `order` key containing `['sort' => 'asc', 'Countries.sort' => 'desc']`  after the default `order` and `sort` data are merged. Then the `_prefix()` method is preferring the prefixed key. Perhaps the `_prefix` method should not overwrite keys if they already exist?

markstory:
Are you certain that the behavior you want has ever worked? I can take a look at 3.5, but I get the feeling that this ever worked based on how extensive the changes required to fix this problem are.

dereuromark:
Not fully sure. But that would mean that many apps have this problem by design for no reason IMO.
This shimming is not expensive given a normal dispatcher cycle for pagination, and since the scope is limited to only this specific use case also does less harm having than not having.

Once we can revisit pagination into a common field for sortfield+direction all these issues will then hopefully resolve themselves.

dereuromark:
They do match, the incoming query strings just are not prefixed as per default bake template.
Thing is, the order in the controllers usually have to be model prefixed as otherwise the SQL error about unclear field is thrown (if the same field exists in the contained belongs to relations, which happens a lot).
So this here is really useful for the primary model to work flawless and without painfully upgrading all templates here.

I understand your caution, but let's maybe just check that the scope of the change is indeed as limited as I suspect. This way it will work better also for the default templates and existing default orders.
Check the condition about removing: Only if also the incoming one is not prefixe (same model), and also only without whitelist afaik. Otherwise all remains the same.

> Do you mean combining sort/direction into a single query parameter like sort=-created?

Yes, please see the discussion https://github.com/cakephp/cakephp/issues/10028#issuecomment-272812357
Having a common string generated for possible orders and used by both helper and component, you can easily then build all use cases, incl complex multi-sort, without all the overhead we currently have due to the n*m matrix like options. But a discussion for a future implementation.

dereuromark:
@markstory Do you rather want to put this into 3.7 (RC) then?

PR #12746, opened 2018-11-20
Include prefer lowest and simplify checks

dereuromark:
We have way too many jobs IMO, I think we can cut down a bit on the matrix of all PHP versions vs all DB types. Usually testing min and max should be enough between php versions.
But I dont want to touch too much

For now:
- Simplify checks together
- Add prefer lowest as we have more and more dependencies in core, and as such min requirements should also be tested.

This way the current job count doesnt increase.
CS on top of phpstan is fair IMO as the stickler auto-fixes them now anyway.
So having a combined "checks" run seems like a pragmatic way moving forward.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,6 @@ language: php
 php:
   - 7.0
   - 5.6
-  - 7.1
   - 7.2
   - 7.3
 
@@ -17,7 +16,7 @@ env:
   global:
     - DEFAULT=1
     - CODECOVERAGE=0
-    - PHPCS=0
+    - CHECKS=0
 
 services:
   - memcached
@@ -37,13 +36,13 @@ matrix:
   fast_finish: true
 
   include:
+    - php: 5.6
+      env: PREFER_LOWEST=1
+
     - php: 7.0
-      env: PHPCS=1 DEFAULT=0
+      env: CHECKS=1 DEFAULT=0
 
     - php: 7.1
-      env: PHPSTAN=1 DEFAULT=0
-
-    - php: 7.2
       env: CODECOVERAGE=1 DEFAULT=0
 
 before_install:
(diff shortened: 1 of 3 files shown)


dereuromark:
> PHP Fatal error:  Call to a member function toIso8601String() on null in /home/travis/build/cakephp/cakephp/tests/TestCase/I18n/TimeTest.php on line 763

.travis.yml
```diff
@@ -36,50 +36,51 @@ matrix:
 
   include:
     - php: 7.0
-      env: PHPCS=1 DEFAULT=0
+      env: CHECKS=1 DEFAULT=0
```
markstory: Combining checks builds is :+1:

markstory:
@garas Can we bump it in `require-dev`?

garas:
Composer uses `require-dev` constraint when defined in both `require` and `require-dev`, but it looks like it is not officially supported.

I'm not sure if Composer will use `require` constraint when `cakephp/cakephp` is installed as dependency.

ADmad:
Merge conflict needs to be resolved.

dereuromark:
Everything looks good except for mcrypt
How do I get that one back?

markstory:
Those PHPCS errors appear to be new. We could disable deprecated function sniff warnings for that file.

dereuromark:
How can we proceed here, how to disable those errors the best way?

ADmad:
> How can we proceed here, how to disable those errors the best way?

Exclude the file from [phpstan](https://github.com/phpstan/phpstan#exclude-files-from-analysis) analysis.

dereuromark:
Thx!

ADmad:
Hmm... that didn't fix it. Strangely using latest phpstan and PHP 7.1 locally I don't get those errors.

dereuromark:
Should we rebase upon origin/master?

ADmad:
I doubt that will help. I'll just update to use pattern in `ignoreErrors` config instead of trying to ignore the file.

ADmad:
@dereuromark You managed to increase the errors reported by phpcs :stuck_out_tongue:. Need to run the static analytic checks on PHP 7.2 so that memcache and mcrypt are available.

dereuromark:
Oh man, finally :) Quite the Odyssey

PR #12747, opened 2018-11-20
Try to fix codecov builds

markstory:
Don't include TRAVIS_PHP_VERSION in the commands that run coverage. Due to codecov/codecov-bash#133 we should try to not have a version number in the command so that codecov can detect environment variables better.

This will add more builds, but we might get coverage for all versions of PHP & all database drivers.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,7 @@ env:
     - DB=sqlite db_dsn='sqlite:///:memory:'
   global:
     - DEFAULT=1
+    - CODECOVERAGE=0
 
 services:
   - memcached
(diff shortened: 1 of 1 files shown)


dereuromark:
See also https://github.com/cakephp/cakephp/pull/12746/
Maybe we can combine those, and also reduce some of the overhead of all PHP vs all DB by using min+max per DB type of SqLite etc?

markstory:
I think we should continue running the a full matrix of php + db vendors. I know it caught a few problems during the typehint changes in 4.x.

dereuromark:
You will find those using min and max
e.g. 5.6, 7.0, 7.3 for example.

markstory:
> You will find those using min and max

@dereuromark I'm not so sure. During 4.x typehint changes we had numerous problems between 7.2 and 7.1. Is build duration something that is a problem in your day to day work on Cake?

markstory:
Merging to see if we can get updated coverage results in codecov. These changes will drop coverage for non SQLite databases as they don't get run. I'll try to find a solution for that later.

PR #12748, opened 2018-11-20
add missing namespace

saeideng:


Iandenh:
I see that my PR #12561 is overwritten by the merge of master to 3.x and then this error happend.

markstory:
@Iandenh yes, that was my mistake. I'll get it fixed.

PR #12754, opened 2018-11-22
Remove deprecated usage.

dereuromark:
Does this make sense?

### DIFF
--- a/src/Command/HelpCommand.php
+++ b/src/Command/HelpCommand.php
@@ -131,19 +131,19 @@ class HelpCommand extends Command implements CommandCollectionAwareInterface
      */
     protected function asXml($io, $commands): void
     {
-        $commands = new SimpleXMLElement('<commands></commands>');
+        $shells = new SimpleXMLElement('<shells></shells>');
         foreach ($commands as $name => $class) {
             if (is_object($class)) {
                 $class = get_class($class);
             }
-            $command = $commands->addChild('command');
-            $command->addAttribute('name', $name);
-            $command->addAttribute('call_as', $name);
-            $command->addAttribute('provider', $class);
-            $command->addAttribute('help', $name . ' -h');
+            $shell = $shells->addChild('shell');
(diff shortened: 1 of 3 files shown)


src/Command/HelpCommand.php
```diff
@@ -131,19 +131,19 @@ protected function asText($io, $commands): void
      */
     protected function asXml($io, $commands): void
     {
-        $shells = new SimpleXMLElement('<shells></shells>');
+        $commands = new SimpleXMLElement('<commands></commands>');
```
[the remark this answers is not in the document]
dereuromark: ok

tests/TestCase/Command/HelpCommandTest.php
```diff
@@ -37,7 +37,7 @@ public function setUp()
     {
         parent::setUp();
         $this->setAppNamespace();
-        $this->useCommandRunner(true);
+        $this->useCommandRunner();
```
saeideng: I think we can use commandRunner by default with changing
 https://github.com/cakephp/cakephp/blob/8baeda92fbccc426608a2d347ead7f5dfccb5af1/src/TestSuite/ConsoleIntegrationTestTrait.php#L41

saeideng: I'm saying we can remove whole `$this->useCommandRunner();` if we would change the
`$_useCommandRunner = false;` to `$_useCommandRunner = true;`
since shell will remove in v5

dereuromark: Feel free to make a PR in top here.

saeideng: I'll open , after merging this PR

dereuromark:
Any idea why the tests still fail after revert?

 > Failed asserting that '<command name="sample" call_as="sample" provider="TestApp\Shell\SampleShell" help="sample -h"' is in output.

PR #12755, opened 2018-11-22
Fix wrong dependencies.

dereuromark:
CS for 4.x branch is currently failing due to the wrong branch.
Also, the included slevomat dependency is wrong here, it is declared on the sniffer already - where it belongs.


ADmad:
I updated the `next` of our codesniffer just couple of days ago. It probably still needs some tweaking.

PR #12757, opened 2018-11-22
Raise phpstan to level 3. WIP

dereuromark:
Looks like {} for inheritdoc should only be used for partial overwrite.
For full one we should use normal tag. That also works then for level 3+.

https://github.com/phpstan/phpstan/issues/1648

WIP
more to do - maybe others can add commits on top.

From initial `140+` errors down to `51`.


src/Auth/ControllerAuthorize.php
```diff
@@ -50,7 +50,7 @@ class ControllerAuthorize extends BaseAuthorize
     protected $_Controller;
 
     /**
-     * {@inheritDoc}
+     * @inheritDoc
```
ADmad: This change makes our codesniffer unhappy. So the related sniff will also have to be updated.

src/Database/SchemaCache.php
```diff
@@ -89,11 +90,11 @@ public function clear(?string $name = null): array
     /**
      * Helper method to get the schema collection.
      *
-     * @param \Cake\Database\Connection $connection Connection object
+     * @param \Cake\Datasource\ConnectionInterface $connection Connection object
      * @return \Cake\Database\Schema\Collection|\Cake\Database\Schema\CachedCollection
      * @throws \RuntimeException If given connection object is not compatible with schema caching
      */
-    public function getSchema(Connection $connection)
+    public function getSchema(ConnectionInterface $connection)
```
ADmad: The method uses `$connection->cacheMetadata()` and `cacheMetadata()` does not exist in `ConnectionInterface`. So the typhint should be reverted to concrete class (and changed in constructor too).

Adding `cacheMetadata()` to the interface isn't an option as for e.g. elastic search plugin's connection class does not have this method.

src/Auth/ControllerAuthorize.php
dereuromark: Done in https://github.com/cakephp/cakephp-codesniffer/pull/215

src/Database/SchemaCache.php
dereuromark: Hm, that is less than ideal.
As this means the whole class needs it, also means we get IDE errors on `ConnectionManager::get('test')` return value and usage between the cache and non cache versions using either concrete or interface contracts.
I made a PR to this one, but I sure dont like it.
Better to somehow add cacheMetadata() and throw not implemented exception or alike IMO.

src/Auth/ControllerAuthorize.php
markstory: Do you know if api-gen will also follow this style of tag?

dereuromark: I do not. We should find out. However, it should, since this is the correct tag for full overwrite.

ADmad: We will have to make updates/fixes to apigen anyway to use with 4.x, regardless of this change :slightly_smiling_face:.

src/Database/SchemaCache.php
```diff
@@ -30,16 +30,16 @@ class SchemaCache
     /**
      * Schema
      *
-     * @var \Cake\Database\Schema\CachedCollection
+     * @var \Cake\Database\Schema\CachedCollection|\Cake\Database\Schema\Collection
```
ADmad: This change is making pslam unhappy https://travis-ci.org/cakephp/cakephp/jobs/458819610#L835

ADmad:
> From initial 140+ errors down to 51.

I brought it down to 47 :smile: 

@dereuromark I think we should merge this (once the pslam error is fixed) while keeping phpstan level at 2 in travis config. Otherwise we'll have to keep dealing with merge conflicts.

dereuromark:
Yes, we can track level 3 separately over time until we can resolve all issues.
Would be good however to identify phpstan bugs and report them, so that they get fixed on their end.

ADmad:
We could add a job with phpstan level 3 under "allow_failures".

dereuromark:
For 4.x this makes sense.

ADmad:
I added new job as mentioned above and also bumped up pslam to v3. Let's see if we get :heavy_check_mark:.

dereuromark:
phpstan level 3 is now not executed anymore (in allow failure list, also green).
A small thing, otherwise all good, great work.

.travis.yml
```diff
@@ -39,6 +39,13 @@ matrix:
     - php: 7.1
       env: STATIC_ANALYSIS=1 DEFAULT=0
 
+    - php: 7.2
+      env: STATIC_ANALYSIS=1 PHPSTAN_LEVEL_3=1 DEFAULT=0
```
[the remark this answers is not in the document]
ADmad: Nope, i had forgotten `$` before the var name below :).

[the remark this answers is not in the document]
garas: `PHPSTAN_LEVEL=3` would be better var name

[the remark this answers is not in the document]
dereuromark: but arent both now running level 3? normal and allow failure?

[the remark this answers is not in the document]
garas: normal - 2, failure -3

PR #12759, opened 2018-11-23
Raise phpstan to level 3. WIP

dereuromark:
But that is less than ideal.
As this means the whole class needs it, also means we get IDE errors on ConnectionManager::get('test') return value and usage between the cache and non cache versions using either concrete or interface contracts.
I made a PR to this one, but I sure dont like it.
Better to somehow add cacheMetadata() and throw not implemented exception or alike IMO.

### DIFF
--- a/src/Database/SchemaCache.php
+++ b/src/Database/SchemaCache.php
@@ -16,7 +16,6 @@ declare(strict_types=1);
 namespace Cake\Database;
 
 use Cake\Cache\Cache;
-use Cake\Datasource\ConnectionInterface;
 
 /**
  * Schema Cache.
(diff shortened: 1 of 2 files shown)


ADmad:
> Better to somehow add cacheMetadata() and throw not implemented exception or alike IMO.

That sounds fine to me. In any case the `Collection` class of elastic search and any other userland implementations will have to be updated for 4.x anyway. Let's see what other's have to say.

ADmad:
For now I suggest we maintain status quo and just ignore any error phpstan throws instead of making any typehinting change to `SchemaCache`.

PR #12761, opened 2018-11-23
Test with SQLite first to get first results faster

garas:
Travis takes about 7.5 minutes for builds with MySQL, and takes only a bit more than 2 minutes with SQLite.

Testing with SQLite first would save 5 minutes to see non-DB-specific failures, and it would allow to start fixing sooner.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,9 +11,9 @@ dist: trusty
 
 env:
   matrix:
+    - DB=sqlite db_dsn='sqlite:///:memory:'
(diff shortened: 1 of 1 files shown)


dereuromark:
I would rather cut down on the travis matrix where total overhead ( refs https://github.com/cakephp/cakephp/pull/12746 )
as Mysql and Postgres are the more important DBs. 
But sure, we can also do this here.

garas:
Yes, it delays MySQL/PostgreSQL by 2 minutes, but assuming most PRs doesn't touch DB specific areas, it would speed up in most cases.

Reducing matrix will help with total time taken, this PR will get first result faster.

PR #12764, opened 2018-11-26
Is unique custom condition

half2me:
Thought this might be useful. A small addition to the `isUnique` rule, that allows it to be optionally extended to use custom conditions.

### DIFF
--- a/src/ORM/Rule/IsUnique.php
+++ b/src/ORM/Rule/IsUnique.php
@@ -79,7 +79,7 @@ class IsUnique
             }
         }
 
-        if (array_key_exists('conditions', $options)) {
+        if (isset($options['conditions'])) {
             $conditions += $options['conditions'];
         }
 
--- a/src/ORM/Rule/IsUnique.php
+++ b/src/ORM/Rule/IsUnique.php
@@ -57,6 +57,12 @@ class IsUnique
     /**
      * Performs the uniqueness check
      *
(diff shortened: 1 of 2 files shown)


src/ORM/Rule/IsUnique.php
```diff
@@ -79,6 +79,10 @@ public function __invoke(EntityInterface $entity, array $options)
             }
         }
 
+        if (array_key_exists('conditions', $options)) {
```
dereuromark: why not just isset()?

dereuromark:
The "### Options" part also needs to be updated for documentation.
But otherwise looks good to me.

half2me:
> ### Options
Its just a bit strange though that the `$options` is being used in two different places: one in the constructor, and another time in `__invoke()`. For example the existing option for `allowMultipleNulls` seems like it can only be set in the constructor, even though it might be expected to work on invocation as well. Where should I put my option?

dereuromark:
Do we want to continue with this for 3.x?

dereuromark:
Bump - this does look useful if you want to finish it up.

saeideng:
@dereuromark  could this be part of `$rules->add()` and not rule itself (`isUnique`,.....)?

othercorey:
Closing due to no activity. For anyone interested, please open an issue to discuss.

PR #12765, opened 2018-11-27
Add deprecation warning for routing middleware

### LINKED ISSUE #12740: 3.7.0 middleware loading issues
This is a (multiple allowed):

* [x] bug
* [ ] enhancement

* CakePHP Version: 3.6 latest (3.7)

### What you did

I upgraded a 3.5 app to 3.6 latest, and while doing so I kept the middelware loading the same:

	->add(RoutingMiddleware::class);

### What happened
$app going into the middleware is null (see https://github.com/cakephp/cakephp/pull/12739).
This silently fallbacks to some weird thing where now no plugin routes are loaded anymore in Web requests. CLI works fine, though.

So it seems that this is not so bc after all, and as such dangerous to be kept silent.

That it still loads the plugin routes.
Or maybe some deprecation warning here if `$app === null` to make it easier to know that it now requires the $app to be actually passed as object to load plugin routing.

As

	->add(new RoutingMiddleware($this, '_cake_routes_'))

fixes the issue

markstory:
While routing 'works' without an application plugin routing will not work correctly.

Fixes #12740

### DIFF
--- a/src/Routing/Middleware/RoutingMiddleware.php
+++ b/src/Routing/Middleware/RoutingMiddleware.php
@@ -59,6 +59,12 @@ class RoutingMiddleware
      */
     public function __construct(HttpApplicationInterface $app = null, $cacheConfig = null)
     {
+        if ($app === null) {
+            deprecationWarning(
+                'RoutingMiddleware should be passed an application instance. ' .
+                'Failing to do so can cause plugin routes to not behave correctly.'
+            );
+        }
         $this->app = $app;
         $this->cacheConfig = $cacheConfig;
     }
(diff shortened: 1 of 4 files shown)


src/Routing/Middleware/RoutingMiddleware.php
```diff
@@ -59,6 +59,12 @@ class RoutingMiddleware
      */
     public function __construct(HttpApplicationInterface $app = null, $cacheConfig = null)
```
ADmad: So we'll remove default `null` in 4.x?

markstory: I think so

ADmad: Sounds good.

ADmad (approved):

PR #12767, opened 2018-11-27
Fix description of Memcached engine

CauanCabral:

### DIFF
--- a/src/Cache/Engine/MemcachedEngine.php
+++ b/src/Cache/Engine/MemcachedEngine.php
@@ -23,8 +23,8 @@ use Memcached;
  * control you have over expire times far in the future. See MemcachedEngine::write() for
  * more information.
  *
- * Memcached engine support binary protocol and igbinary
- * serialization (if memcached extension compiled with --enable-igbinary).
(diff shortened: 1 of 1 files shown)


src/Cache/Engine/MemcachedEngine.php
```diff
@@ -23,10 +23,9 @@
  * control you have over expire times far in the future. See MemcachedEngine::write() for
  * more information.
  *
- * Main advantage of this Memcached engine over the memcached engine is
- * support of binary protocol, and igbinary serialization
- * (if memcached extension compiled with --enable-igbinary)
- * Compressed keys can also be incremented/decremented
+ * Memcached engine support binary protocol and igbinary
```
ADmad: Should be "supports".

src/Cache/Engine/MemcachedEngine.php
```diff
@@ -23,10 +23,9 @@
...
+ * serialization (if memcached extension compiled with --enable-igbinary).
```
ADmad: "..*is* compiled with.."

PR #12770, opened 2018-11-28
Fix Association::$finder type

o0h:
phpdoc describes `type string`, but it can retrieve an array as `[$finder:string, $options:array]` .
We can use this as bellow.

```php
$Table->belongsTo('Users')
    ->setFinder(['custom' => ['opt1' => 'some value']]);
```

This works like:

```php
// ORM\Association
$this->getTarget()->find($type, $options + $opts);
```
Because `_extractFinder()` method returns array including options.
https://github.com/cakephp/cakephp/blob/3.7.0-RC3/src/ORM/Association.php#L1391-L1399

### DIFF
--- a/src/ORM/Association.php
+++ b/src/ORM/Association.php
@@ -860,7 +860,7 @@ abstract class Association
     /**
      * Gets the default finder to use for fetching rows from the target table.
      *
-     * @return string
+     * @return string|array
      */
     public function getFinder()
     {
--- a/src/ORM/Association.php
+++ b/src/ORM/Association.php
@@ -179,8 +179,9 @@ abstract class Association
 
     /**
      * The default finder name to use for fetching rows from the target table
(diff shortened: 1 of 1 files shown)


src/ORM/Association.php
```diff
@@ -886,7 +887,7 @@ public function setFinder($finder)
      *
      * @deprecated 3.4.0 Use setFinder()/getFinder() instead.
      * @param string|null $finder the finder name to use
-     * @return string
+     * @return string|array
```
ADmad: Return type of `getFinder()` method should also be updated.

o0h: pushed https://github.com/cakephp/cakephp/pull/12770/commits/1a46ad998e12bd47314b6bf1fba1e6a4e202c2ff 😄

Issue #12771, opened 2018-11-28
Entity of BelongsTo association empty when using innerjoin with match and specifying select fields


davalb:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.6.13
* Platform and Target: apache, mysql

I used `innerJoinWith` in combination with `select`.

The belongsTo association was empty. (the _matchingData-property was set - but it is now what I expected). I will post a PR shortly with a test, that demonstrates the problem.

I expected the belongsTo association entity to be present

markstory:
The `innerJoinWith()` method wraps `matching()` which why the `_matching` property is populated and not your association property. If you want to select fields from an association use `contain()` and use `select()` to subset the fields.

PR #12772, opened 2018-11-28
#12771 - test for inner join match with select specified

davalb:
This PR contains a failing test to illustrate the bug described in #12771 . I would like to get confirmation that this is not intentional behaviour. If that is the case, I would like to try to provide a fix.

### DIFF
--- a/tests/TestCase/ORM/QueryTest.php
+++ b/tests/TestCase/ORM/QueryTest.php
@@ -3688,4 +3688,39 @@ class QueryTest extends TestCase
                 ->find()
                 ->selectAllExcept([], ['body']);
     }
+
+    /**
+     * Test that the entity of a belongsTo association is not empty if we are using an inner join with match
+     * and specify a list of select fields
+     */
+    public function testInnerJoinMatchWithQueryFields()
+    {
+        $table = $this->getTableLocator()->get('Articles');
+        $table->belongsTo('Authors');
+
+        // 1. without providing a set of select fields the test succeeds
+        $query = $table->find('all', [
+            'contain' => ['Authors']
+        ]);
+        $query->innerJoinWith('Authors', function ($q) {
+            return $q->where(['Authors.name' => 'larry']);
+        });
+        $article = $query->first();
+        $this->assertNotEmpty($article->author);
+
+        // 2. when providing an array of fields to select the author-entity is empty
+        $query = $table->find('all', [
+            'contain' => ['Authors']
+        ]);
+        $query->innerJoinWith('Authors', function ($q) {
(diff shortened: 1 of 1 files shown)


tests/TestCase/ORM/QueryTest.php
```diff
@@ -3688,4 +3688,39 @@ public function testSelectAllExceptThrowsInvalidArgument()
...
+        $article = $query->first();
+        $this->assertNotEmpty($article->author);
+
+        // 2. when providing an array of fields to select the author-entity is empty
+        $query = $table->find('all', [
+            'contain' => ['Authors']
+        ]);
+        $query->innerJoinWith('Authors', function ($q) {
+            return $q->where(['Authors.name' => 'larry']);
+        });
```
markstory: You don't need to mix `contain` + `innerJoinWith`. You can use `contain()` and `where()`

davalb: You're right, and I fixed the problem in the project I got this from by replacing the `innerJoinWith` with `contain` and `where`.
Do you think that it might nonetheless be worth it to look into why the author-entity goes missing, when using this particular combination of `contain`, `innerJoinWith` and `select`? I don't want to waste my or anybodies time with it.

markstory: Figuring out why is a complicated time-intensive process that I don't have the energy for.

markstory: Before you figure out why and a potential fix, it might be good to determine what the correct behavior should be. Right now I think matching will overrule contain. Scenarios that might make the decision less simple are how it behaves when aggregate fields are in use.

othercorey:
I ran into this today.  I found that if I set an alias for the fields in the select then it won't be put into _matchingData.

```php
->select(['alias' => 'Association.field'])
->innerJoinWith('Association')
```

othercorey:
@markstory Does CakePHP internals have the capability to disable the alias generation for Association-specific columns?  I presume that's why the entity ends up in _matchingData since the column name is decorated and so the parsing matches.  Just disabling _matchingData doesn't really do much unless the raw column name is used.

markstory:
> Does CakePHP internals have the capability to disable the alias generation for Association-specific columns?

No as we we use aliases to format results and build the resulting output entities. Have you tried aliasing your field to `Association__alias` or `Association.alias` instead of `alias`?

othercorey:
I was saying that using an alias fixes the "issue" of the entity being formed in _matchingData when using innerJoinWith().  I think that's a reasonable work-around.

Issue #12773, opened 2018-11-28
Form helper maxLenght attribute from validation


lamasgergo:
This is a (multiple allowed):

* [ ] bug
* [ ] enhancement
* [x] feature-discussion (RFC)

* CakePHP Version: 3.x or 4.x.

### What you did
I have field (named for example myfield), wich has is a varchar(45) in my database.
I set maxlenght validation in my model for this field. 

`$validator->maxlength('myfield', 10);`

### What happened
Form helper generated a field with maxlength="45" attribute.

`<input name="myfield" maxlength="45" ...>`

### What you expected to happen
It would be great the generated field should get maxlength="10" attribute.

`<input name="myfield" maxlength="10" ...>`

So the model validation could be overwrite what i got from shema.

[example.zip](https://github.com/cakephp/cakephp/files/2625508/example.zip)
![form_helper](https://user-images.githubusercontent.com/879741/49162369-04888800-f32b-11e8-814e-530e4a1968d5.jpg)
![entity_context](https://user-images.githubusercontent.com/879741/49162371-04888800-f32b-11e8-8083-88e309837f60.jpg)

markstory:
Reading the maxlength out of the validation could work be a nice improvement. Is there a reason we couldn't add the same method to the other context implementations as well?

lamasgergo:
i was not sure this is the best way, but if it is of course i will check out the other contexts and i will send a pull request.

PR #12775, opened 2018-11-29
Form Helper maxlength attribute uses validation #12773

lamasgergo:
Improvement for 
https://github.com/cakephp/cakephp/issues/12773

Form helper now supports maxlength attribute from validation directly.

### DIFF
--- a/src/View/Form/ArrayContext.php
+++ b/src/View/Form/ArrayContext.php
@@ -12,6 +12,7 @@
  * @since         3.0.0
  * @license       https://opensource.org/licenses/mit-license.php MIT License
  */
+
 namespace Cake\View\Form;
 
 use Cake\Http\ServerRequest;
@@ -212,19 +213,20 @@ class ArrayContext implements ContextInterface
      * In this context class, this is simply defined by the 'maxlength' array.
      *
      * @param string $field A dot separated path to check required-ness for.
-     * @return int|false
+     * @return int|null
      */
     public function getMaxLength($field)
     {
         if (!is_array($this->_context['schema'])) {
-            return false;
+            return null;
         }
         $fieldShema = Hash::get($this->_context['schema'], $field);
-        if(is_array($fieldShema)) {
+
+        if (is_array($fieldShema)) {
             $maxLength = Hash::get($fieldShema, 'length');
         }
 
-        return $maxLength ?? false;
+        return $maxLength ?? null;
     }
 
     /**
--- a/src/View/Form/ArrayContext.php
+++ b/src/View/Form/ArrayContext.php
@@ -226,10 +226,8 @@ class ArrayContext implements ContextInterface
         if(!$fieldSchema){
             return null;
         }
-
-        if (is_array($fieldSchema)) {
-            $maxLength = Hash::get($fieldSchema, 'length');
-        }
+        
+        $maxLength = Hash::get($fieldSchema, 'length');
 
         return $maxLength ?? null;
     }
--- a/src/View/Form/ArrayContext.php
+++ b/src/View/Form/ArrayContext.php
@@ -220,9 +220,11 @@ class ArrayContext implements ContextInterface
             return false;
         }
         $fieldShema = Hash::get($this->_context['schema'], $field);
-        $maxLength =  Hash::get($fieldShema, 'length');
+        if(is_array($fieldShema)) {
+            $maxLength = Hash::get($fieldShema, 'length');
+        }
 
-        return $maxLength??false;
+        return $maxLength ?? false;
     }
 
     /**
--- a/src/View/Form/ArrayContext.php
+++ b/src/View/Form/ArrayContext.php
@@ -220,10 +220,15 @@ class ArrayContext implements ContextInterface
         if (!is_array($this->_context['schema'])) {
             return null;
         }
-        $fieldShema = Hash::get($this->_context['schema'], $field);
 
-        if (is_array($fieldShema)) {
-            $maxLength = Hash::get($fieldShema, 'length');
+        $fieldSchema = Hash::get($this->_context['schema'], $field);
+
+        if(!$fieldSchema){
+            return null;
+        }
+
+        if (is_array($fieldSchema)) {
+            $maxLength = Hash::get($fieldSchema, 'length');
         }
 
         return $maxLength ?? null;
--- a/src/View/Form/ArrayContext.php
+++ b/src/View/Form/ArrayContext.php
@@ -207,6 +207,25 @@ class ArrayContext implements ContextInterface
     }
 
     /**
+     * Get field length from validation
+     *
+     * In this context class, this is simply defined by the 'maxlength' array.
+     *
+     * @param string $field A dot separated path to check required-ness for.
+     * @return int|false
+     */
+    public function getMaxLength($field)
+    {
+        if (!is_array($this->_context['schema'])) {
+            return false;
+        }
+        $fieldShema = Hash::get($this->_context['schema'], $field);
+        $maxLength =  Hash::get($fieldShema, 'length');
+
+        return $maxLength??false;
+    }
+
+    /**
      * {@inheritDoc}
      */
     public function fieldNames()
--- a/src/View/Form/ArrayContext.php
+++ b/src/View/Form/ArrayContext.php
@@ -12,7 +12,6 @@
  * @since         3.0.0
  * @license       https://opensource.org/licenses/mit-license.php MIT License
  */
-
 namespace Cake\View\Form;
 
 use Cake\Http\ServerRequest;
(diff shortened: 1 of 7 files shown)


src/View/Form/ArrayContext.php
```diff
@@ -206,6 +206,27 @@ public function isRequired($field)
         return (bool)$required;
     }
 
+    /**
+     * Get field length from validation
+     *
+     * In this context class, this is simply defined by the 'maxlength' array.
+     *
+     * @param string $field A dot separated path to check required-ness for.
+     * @return int|false
```
dereuromark: for new API methods i would rather want to see nullable 

    int|null

also for future php typehinting

src/View/Form/ContextInterface.php
```diff
@@ -70,6 +70,14 @@ public function val($field);
      */
     public function isRequired($field);
 
+    /**
+     * Get maximum length of a filed from model validation
+     *
+     * @param string $field A dot separated path to check required-ness for.
+     * @return bool|false
```
dereuromark: same

src/View/Form/ArrayContext.php
```diff
@@ -212,19 +213,20 @@ public function isRequired($field)
...
         }
         $fieldShema = Hash::get($this->_context['schema'], $field);
-        if(is_array($fieldShema)) {
+
+        if (is_array($fieldShema)) {
             $maxLength = Hash::get($fieldShema, 'length');
         }
 
-        return $maxLength ?? false;
+        return $maxLength ?? null;
```
dereuromark: isnt $maxLength already nullable? The cloaking ?? should be avoided here then.

src/View/Form/EntityContext.php
```diff
@@ -432,16 +432,17 @@ public function isRequired($field)
      * @param string $field The dot separated path to the field you want to check.
      * @return int|false
```
dereuromark: null?

dereuromark:
@lamasgergo This cannot go into 3.6, this should go into 3.7 (next).
But we cannot fix the existing methods to nullable I think - this might be too much a BC break.
So better to stick with "false" for now as return value, sry.
We can fix this hopefully for 4.0 then.

src/View/Form/ArrayContext.php
dereuromark: Actually, since those are existing API methods, better to keep false, see below.

lamasgergo:
> @lamasgergo This cannot go into 3.6, this should go into 3.7 (next).
> But we cannot fix the existing methods to nullable I think - this might be too much a BC break.
> So better to stick with "false" for now as return value, sry.
> We can fix this hopefully for 4.0 then.

So revert to false returns everywhere?

src/View/Form/ArrayContext.php
```diff
@@ -206,6 +207,28 @@ public function isRequired($field)
...
+     */
+    public function getMaxLength($field)
+    {
+        if (!is_array($this->_context['schema'])) {
+            return null;
+        }
+        $fieldShema = Hash::get($this->_context['schema'], $field);
+
+        if (is_array($fieldShema)) {
+            $maxLength = Hash::get($fieldShema, 'length');
```
dereuromark: $maxLength is not defined before, but it should be.

dereuromark:
Please, after setting the target to 3.next, add some test cases for this new behavior.

dereuromark:
> So revert to false returns everywhere?

No, I just didnt see the correct diff :) My mistake. New API methods, all good.

lamasgergo:
> Please, after setting the target to 3.next, add some test cases for this new behavior.

I have added 3 test case into FormHelperTest. Isnt it this enough?

src/View/Form/ContextInterface.php
```diff
@@ -70,6 +70,14 @@ public function val($field);
      */
     public function isRequired($field);
 
+    /**
+     * Get maximum length of a field from model validation
+     *
+     * @param string $field A dot separated path to check required-ness for.
+     * @return bool|null
```
dereuromark: why bool?

src/View/Form/EntityContext.php
```diff
@@ -426,6 +426,27 @@ public function isRequired($field)
...
+    public function getMaxLength($field)
+    {
+        $parts = explode('.', $field);
+        $validator = $this->_getValidator($parts);
+        $fieldName = array_pop($parts);
+        if (!$validator->hasField($fieldName)) {
+            return null;
+        }
+        foreach ($validator->field($fieldName)->rules() as $rule) {
+            if ($rule->get('rule') == 'maxLength') {
```
dereuromark: `===` always for strings

src/View/Form/EntityContext.php
```diff
@@ -426,6 +426,27 @@ public function isRequired($field)
...
+                return $rule->get('pass')[0] ?? null;
+            }
+        }
```
dereuromark: missing return null
semantically, this would be void otherwise.

src/View/Form/FormContext.php
```diff
@@ -134,6 +134,24 @@ public function isRequired($field)
...
+     * {@inheritDoc}
+     */
+    public function getMaxLength($field)
+    {
+        $validator = $this->_form->getValidator();
+        if (!$validator->hasField($field)) {
+            return null;
+        }
+        foreach ($validator->field($field)->rules() as $rule) {
+            if ($rule->get('rule') == 'maxLength') {
```
dereuromark: Same

src/View/Helper/FormHelper.php
```diff
@@ -1476,6 +1476,9 @@ protected function _magicOptions($fieldName, $options, $allowOverride)
             unset($options['step']);
         }
 
+        $validationLength = $context->getMaxLength($fieldName);
+        $fieldDef['length'] = $fieldDef['length'] ?? $validationLength;
```
dereuromark: Why the check here if you already calc based on context?
Cleaner to drop the $validationLength var then and only do the lookup if really necessary.

tests/TestCase/View/Helper/FormHelperTest.php
```diff
@@ -1742,24 +1745,24 @@ public function testFormSecurityMultipleControlDisabledFields()
         $result = $this->Form->secure($this->Form->fields);
         $hash = '43c4db25e4162c5e4edd9dea51f5f9d9d92215ec%3AAddresses.0.id%7CAddresses.1.id';
         $tokenDebug = urlencode(json_encode([
-                '/articles/add',
```
dereuromark: Please avoid side effects of indentation changes, only minimal diffs, please.
Also, these are technically now incorrect, as the normalized form would have 2 levels here.

src/View/Form/EntityContext.php
```diff
@@ -426,6 +426,29 @@ public function isRequired($field)
...
+    {
+        $parts = explode('.', $field);
+        $validator = $this->_getValidator($parts);
+        $fieldName = array_pop($parts);
+        if (!$validator->hasField($fieldName)) {
+            return null;
+        }
+        foreach ($validator->field($fieldName)->rules() as $rule) {
+            if ($rule->get('rule') === 'maxLength') {
+                return $rule->get('pass')[0] ?? null;
```
ADmad: Null coleasing operator `??` is not supported in PHP 5.6.

src/View/Form/ArrayContext.php
```diff
@@ -12,6 +12,7 @@
  * @since         3.0.0
  * @license       https://opensource.org/licenses/mit-license.php MIT License
  */
+
```
ADmad: Please avoid unnecessary whitespace changes.

src/View/Form/EntityContext.php
dereuromark: Basic isset() ternary suffices.

ADmad:
Since this makes changes to public API it needs to target 3.7 aka the `next` branch.

src/View/Form/ArrayContext.php
```diff
@@ -206,6 +207,31 @@ public function isRequired($field)
...
+            return null;
+        }
+
+        $fieldSchema = Hash::get($this->_context['schema'], $field);
+
+        if(!$fieldSchema){
+            return null;
+        }
+        
+        $maxLength = Hash::get($fieldSchema, 'length');
```
dereuromark: Directly return the result. it is either there or null.

src/View/Form/EntityContext.php
lamasgergo: > Null coleasing operator `??` is not supported in PHP 5.6.

Would you like support 5.6? I can change it if you would.

lamasgergo:
> Since this makes changes to public API it needs to target 3.7 aka the `next` branch.

I have to add a new pull request to 3.next?

dereuromark:
No, you can switch the target at the top. Whatever you prefer. A clean one to 3.next also wouldnt hurt. Or rebased/squashed, depending on your git foo :)

lamasgergo:
Resubmitted in #12777 .

PR #12777, opened 2018-11-29
Form Helper maxlength attribute uses validation #12773

lamasgergo:
Improvement for
#12773

Form helper now supports maxlength attribute from validation directly.

Rebased onto 3.next and squashed.

### DIFF
--- a/src/View/Form/ArrayContext.php
+++ b/src/View/Form/ArrayContext.php
@@ -230,6 +230,33 @@ class ArrayContext implements ContextInterface
     }
 
     /**
+     * Get field length from validation
+     *
+     * In this context class, this is simply defined by the 'length' array.
+     *
+     * @param string $field A dot separated path to check required-ness for.
+     * @return int|null
+     */
+    public function getMaxLength($field)
+    {
+        if (!is_array($this->_context['schema'])) {
+            return null;
+        }
+
+        $fieldSchema = Hash::get($this->_context['schema'], $field);
+
+        if (!$fieldSchema) {
+            return null;
+        }
+
+        if (is_array($fieldSchema)) {
+            return Hash::get($fieldSchema, 'length');
+        }
+
+        return null;
+    }
+
+    /**
      * {@inheritDoc}
      */
     public function fieldNames()
--- a/src/View/Form/ArrayContext.php
+++ b/src/View/Form/ArrayContext.php
@@ -243,17 +243,7 @@ class ArrayContext implements ContextInterface
             return null;
         }
 
-        $fieldSchema = Hash::get($this->_context['schema'], $field);
-
-        if (!$fieldSchema) {
-            return null;
-        }
-
-        if (is_array($fieldSchema)) {
-            return Hash::get($fieldSchema, 'length');
-        }
-
-        return null;
+        return Hash::get($this->_context['schema'], "$field.length");
     }
 
     /**
--- a/src/View/Form/ContextInterface.php
+++ b/src/View/Form/ContextInterface.php
@@ -18,6 +18,7 @@ namespace Cake\View\Form;
  * Interface for FormHelper context implementations.
  *
  * @method string|null getRequiredMessage($field) Gets the default "required" error message for a field
+ * @method int|null getMaxLength($field) Get maximum length of a field from model validation
  */
 interface ContextInterface
 {
--- a/src/View/Form/EntityContext.php
+++ b/src/View/Form/EntityContext.php
@@ -455,6 +455,29 @@ class EntityContext implements ContextInterface
     }
 
     /**
+     * Get field length from validation
+     *
+     * @param string $field The dot separated path to the field you want to check.
+     * @return int|null
+     */
+    public function getMaxLength($field)
+    {
+        $parts = explode('.', $field);
+        $validator = $this->_getValidator($parts);
+        $fieldName = array_pop($parts);
+        if (!$validator->hasField($fieldName)) {
+            return null;
+        }
+        foreach ($validator->field($fieldName)->rules() as $rule) {
+            if ($rule->get('rule') === 'maxLength') {
+                return $rule->get('pass')[0];
+            }
+        }
+
+        return null;
+    }
+
+    /**
      * Get the field names from the top level entity.
      *
      * If the context is for an array of entities, the 0th index will be used.
--- a/src/View/Form/FormContext.php
+++ b/src/View/Form/FormContext.php
@@ -110,7 +110,6 @@ class FormContext implements ContextInterface
      * Get default value from form schema for given field.
      *
      * @param string $field Field name.
-
      * @return mixed
      */
     protected function _schemaDefault($field)
@@ -162,6 +161,24 @@ class FormContext implements ContextInterface
         }
         if (!$ruleset->isEmptyAllowed() && $emptyMessage) {
             return $emptyMessage;
+        }
+
+        return null;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function getMaxLength($field)
+    {
+        $validator = $this->_form->getValidator();
+        if (!$validator->hasField($field)) {
+            return null;
+        }
+        foreach ($validator->field($field)->rules() as $rule) {
+            if ($rule->get('rule') === 'maxLength') {
+                return $rule->get('pass')[0];
+            }
         }
 
         return null;
--- a/src/View/Helper/FormHelper.php
+++ b/src/View/Helper/FormHelper.php
@@ -1503,23 +1503,19 @@ class FormHelper extends Helper
         }
 
         $typesWithMaxLength = ['text', 'textarea', 'email', 'tel', 'url', 'search'];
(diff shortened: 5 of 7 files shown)


src/View/Helper/FormHelper.php
```diff
@@ -1502,13 +1502,21 @@ protected function _magicOptions($fieldName, $options, $allowOverride)
             unset($options['step']);
         }
 
+        $validationLength = $context->getMaxLength($fieldName);
```
ADmad: Shouldn't you first check if the field type is one of those in `$allowedTypes` below? If it's not then this method call is unnecessary overhead for other input types (e.g. `select`) which don't need `maxlength` attribute.

So I think this additional code should be move inside the `if` after `$allowedTypes` is initialized.

tests/TestCase/View/Helper/FormHelperTest.php
```diff
@@ -9574,4 +9576,114 @@ public function testControlLabelManipulationCheckboxes()
...
+    }
+
+    /**
+     * testControlMaxLengthEntityContext method
+     *
+     * Test control() with maxlength attribute in Entity Context.
+     *
+     * @return void
+     */
+    public function testControlMaxLengthEntityContext()
```
ADmad: There should also be a test showing that the maxlength set through validator takes precedence over maxlength deduced from table column length.

src/View/Form/ContextInterface.php
```diff
@@ -72,6 +72,14 @@ public function val($field);
      */
     public function isRequired($field);
 
+    /**
+     * Get maximum length of a field from model validation
+     *
+     * @param string $field A dot separated path to check required-ness for.
+     * @return int|null
+     */
+    public function getMaxLength($field);
```
ADmad: We can't change the interface in minor release. Add a `@method` annotation instead at the top similar to what's done for `getRequiredMessage()`.

src/View/Form/EntityContext.php
```diff
@@ -454,6 +454,29 @@ public function getRequiredMessage($field)
...
+    {
+        $parts = explode('.', $field);
+        $validator = $this->_getValidator($parts);
+        $fieldName = array_pop($parts);
+        if (!$validator->hasField($fieldName)) {
+            return null;
+        }
+        foreach ($validator->field($fieldName)->rules() as $rule) {
+            if ($rule->get('rule') === 'maxLength') {
+                return isset($rule->get('pass')[0]) ? $rule->get('pass')[0] : null;
```
ADmad: Is the `isset()` check necessary?  You can't use max length validator without specifying the max length argument.

ADmad (changes requested):

src/View/Form/ContextInterface.php
lamasgergo: OK i have changed

tests/TestCase/View/Helper/FormHelperTest.php
lamasgergo: Ok i have added a test for it.

src/View/Helper/FormHelper.php
ADmad: I understand that maxlength won't be set for e.g. for a textarea. But the `$context->getMaxLength()` call will still be made, which is unnecessary. That's my issue.

lamasgergo: I think we have to rewrite this. Please check out my new file.
This will checking maxlength param too. Is this wrong? What you recommend?

tests/TestCase/View/Helper/FormHelperTest.php
```diff
@@ -9574,4 +9576,227 @@ public function testControlLabelManipulationCheckboxes()
...
+            'div' => ['class'],
+            'label' => ['for'],
+            'Title',
+            '/label',
+            'input' => [
+                'id',
+                'name' => 'title',
+                'type' => 'text',
+                'required' => 'required',
+                'maxlength' => 45,
```
ADmad: Shouldn't this be `55` since that's what's set for maxLength validator? Validation rule length should take precedence over schema length.

src/View/Helper/FormHelper.php
```diff
@@ -1502,13 +1502,20 @@ protected function _magicOptions($fieldName, $options, $allowOverride)
             unset($options['step']);
         }
 
-        $autoLength = !array_key_exists('maxlength', $options)
-            && !empty($fieldDef['length'])
-            && $options['type'] !== 'select';
-
         $allowedTypes = ['text', 'textarea', 'email', 'tel', 'url', 'search'];
```
ADmad: Nitpick: Can you please rename this variable to a more descriptive `$typesWithMaxlength`.

ADmad:
I wonder if all this extra code is really worth the overhead and extra maintenance it will need. Setting a maxLength validation with length smaller than table's column is pretty uncommon and for that edge case on can set `maxlength` in the FormHelper method options.

lamasgergo:
> I wonder if all this extra code is really worth the overhead and extra maintenance it will need. Setting a maxLength validation with length smaller than table's column is pretty uncommon and for that edge case on can set `maxlength` in the FormHelper method options.

I have started this pull request because i runned into in a project where i have a textarea and i have validated in model that, but the helper not generated maxlength attribute.
Ok i can add in the view  with `'maxlength' => 3000` but if i wrote the validation it would be great same magic as this reads out from schema i think.

You choose, close if it's not applicable.

src/View/Form/FormContext.php
```diff
@@ -167,6 +166,24 @@ public function getRequiredMessage($field)
...
+     */
+    public function getMaxLength($field)
+    {
+        $validator = $this->_form->getValidator();
+        if (!$validator->hasField($field)) {
+            return null;
+        }
+        foreach ($validator->field($field)->rules() as $rule) {
+            if ($rule->get('rule') === 'maxLength') {
+                return isset($rule->get('pass')[0])? $rule->get('pass')[0] : null;
```
ADmad: The `isset()` check needs to be removed here too.

dereuromark:
I am just wondering what happened to the 100000 protection against upper values

    min($fieldDef['length'], 100000);

We dont need that anymore?
I feel like this change here could have been introduced in a less changing manner, as pure addition with falling back to the current behavior.

src/View/Form/ArrayContext.php
```diff
@@ -229,6 +229,33 @@ public function getRequiredMessage($field)
         return $required;
     }
 
+    /**
+     * Get field length from validation
+     *
+     * In this context class, this is simply defined by the 'maxlength' array.
```
markstory:
```suggestion
     * In this context class, this is simply defined by the 'length' array.
```

src/View/Helper/FormHelper.php
```diff
@@ -1502,13 +1502,20 @@ protected function _magicOptions($fieldName, $options, $allowOverride)
...
-        $autoLength = !array_key_exists('maxlength', $options)
-            && !empty($fieldDef['length'])
-            && $options['type'] !== 'select';
-
-        $allowedTypes = ['text', 'textarea', 'email', 'tel', 'url', 'search'];
-        if ($autoLength && in_array($options['type'], $allowedTypes)) {
-            $options['maxlength'] = min($fieldDef['length'], 100000);
+        $typesWithMaxLength = ['text', 'textarea', 'email', 'tel', 'url', 'search'];
+        if (in_array($options['type'], $typesWithMaxLength)) {
+            $validationLength = $context->getMaxLength($fieldName);
```
markstory: Because this method isn't in the interface you should do a `method_exists()` check.

markstory: This name could be `$maxLength` as it won't always be based on a validator.

lamasgergo: thanx updated.

lamasgergo:
> I am just wondering what happened to the 100000 protection against upper values
> 
> ```
> min($fieldDef['length'], 100000);
> ```
> We dont need that anymore?
> I feel like this change here could have been introduced in a less changing manner, as pure addition with falling back to the current behavior.

I have added this protection. But why 100000? :) I was wondering on this line when i started this patch. But its ok users can't set maxlength more than 100000.

src/View/Form/ArrayContext.php
lamasgergo: thanx updated

src/View/Form/FormContext.php
lamasgergo: thanx,  i have removed

ADmad:
The `if` block for deducing the max length was overly complicated so I have pushed some simplification.

The way it works now is:
- If `$options['maxlength']` do nothing more.
- Check if max length is available through validation rules and if it is use that.
- Check if max length is available through schema and if it is use that.

Max length set through options or validator *can* be larger than field length in schema. This is necessary as someone might be doing some pre-processing on the submitted value which could be longer than table's column length.

ADmad (approved):

dereuromark (approved):

Issue #12784, opened 2018-12-06
Support multiple arguments in finders


robertpustulka:
This is a (multiple allowed):

* [ ] bug
* [x] enhancement
* [x] feature-discussion (RFC)

* CakePHP Version: 4.0.0

Currently finders support the following API: `find($name, array $options = [])` which translates to `findName(Query $query, array $options`).

While it provides a nice way of configuring finder options it forces you to manually check and validate each option if you accept specific classes or types of options.

I was thinking that maybe we could allow anything to be passed to the finder and make PHP validate arguments with type hinting.

For example finder declared as `findSomething(Query $q, Entity $e, int $number)` could be called like `find('something', $entity, 5)`.

I think that could be achieved painlessly with variadic arguments (`...` syntax). `Table::find($type = 'all', $options = [])` can be changed into `Table::find($type = 'all', ... $args)`. It would still accept array as a second argument and pass it correctly to the finder method. 

What do you think? Should it be as easy as it looks or maybe I'm missing something?

dereuromark:
How would that be in line with current IDE annotations and also tooling like phpstan?
We should move away from magic that cannot be properly annotated and used for introspection as those are hard to understand or lookup for people, or to programmatically verify. We should rather focus on clean and speaking APIs IMO.
But if this can be done, sure, why not :)

robertpustulka:
Currently if your finder require specific arguments to be passed your IDE and phpstan can't tell you that as well. So it's always a compromise. I think that PHP yelling on runtime that's something is wrong is a bit better than having to manually check option key existence and types...

dereuromark:
The status quo sure isnt ideal in many places. All I am saying is: Future API added should have these things at least thought about.
Also, for the array options, there are now also doc blocks and typehinting available, e.g. with psalm etc.
So it is currently not impossible to already do.

markstory:
I like that there is less syntax required for custom options. If we can figure out a backwards compatible approach I'm :+1:

robertpustulka:
Would it be OK to change `RepositoryInterface::find(string $type = 'all', $options = [])` to `RepositoryInterface::find(string $type = 'all', ...$args)`? I'm not up to date with policy about interfaces, sorry.

ADmad:
Changing interfaces is fine in 4.x.

ADmad:
Wondering if `RepositoryInterface::find(string $type = 'all', $options = [], ...$args)` would be better.

robertpustulka:
[...]

PR #12787, opened 2018-12-07
Support multiple finder arguments

robertpustulka:
First take on implementing  #12784.

Existing tests pass with very few changes.
I didin't want to update the docs and haven't added any new tests as I'm not sure if this is a good way forward.


ADmad:
With this one won't be able to stack finder calls, that's sad.

Let's say one make's a custom finder `findCustom(Query $query, Entity)`, you can't do:
```php
$table->find('all', ['conditions' => '..'])->find('custom', $someEntity)`.

**Comment by robertpustulka** [comment]
@ADmad why not? What makes this impossible with this implementation?

**Comment by robertpustulka** [comment]
> We will have to add clarification to the docs that if 2nd argument for find() is array it will always be treated as options for query. 

That's my main concern with this.

I was wondering that maybe we should remove this part: https://github.com/cakephp/cakephp/pull/12787/files#diff-d36d5650ddbf69668bdd04f31f4a7a42R2217

And set options on query explicitly in build-in finders that support options array. I imagine that this change would stay BC but would require people to update those custom finders that for some reason expect that the options are set on query (even though those still should be accessible from the argument).

**Comment by ADmad** [comment]
> ...but would require people to update those custom finders...

I am not in favor of any change which requires updating custom finders. 4.0 is supposed to a be smooth upgrade from 3.x and finders are a very important and well used feature of the framework.

File: src/Datasource/RepositoryInterface.php [hunk 1]
```diff
@@ -55,7 +55,7 @@ public function hasField(string $field): bool;
      * @param array|\ArrayAccess $options An array that will be passed to Query::applyOptions()
      * @return \Cake\Datasource\QueryInterface
      */
-    public function find(string $type = 'all', $options = []);
+    public function find(string $type = 'all', ...$args);
```
markstory: Doc string needs to be updated too.

src/ORM/Behavior/TreeBehavior.php
```diff
@@ -437,7 +437,7 @@ public function childCount(EntityInterface $node, bool $direct = false): int
      * @return \Cake\ORM\Query
      * @throws \InvalidArgumentException When the 'for' key is not passed in $options
      */
-    public function findChildren(Query $query, array $options): Query
+    public function findChildren(Query $query, array $options = []): Query
```
markstory: This implies that the changes here are going to require updates to all application custom finders as well.

src/ORM/Table.php
```diff
@@ -2212,17 +2212,24 @@ public function hasFinder(string $type): bool
...
      * @throws \BadMethodCallException
      */
-    public function callFinder(string $type, Query $query, array $options = []): Query
+    public function callFinder(string $type, Query $query, ...$args): Query
     {
-        $query->applyOptions($options);
-        $options = $query->getOptions();
+        if ($args) {
+            $options = $args[0];
+            if (is_array($options)) {
```
markstory: What about `ArrayAccess` things? Those were in the old doc string type information.

tests/TestCase/ORM/TableTest.php
```diff
@@ -5593,7 +5593,7 @@ public function testSimplifiedFind()
 
         $query = (new Query($this->connection, $table))->select();
         $table->expects($this->once())->method('callFinder')
-            ->with('all', $query, []);
+            ->with('all', $query);
```
markstory: Should have more tests for the new feature. It will quickly become cement and we'll need tests to help catch future regressions.

src/ORM/Table.php
robertpustulka: `Query::applyOptions(array $options)` accepts array only. Looks like doc type info is wrong?

markstory: Possible. We may have not ever actually had checks that were array specific before either.

robertpustulka:
Does it still make sense, can I continue with that?

dereuromark:
I dont like the change tbh.

robertpustulka:
Why not?

dereuromark:
Let me ask this:

- How do static analyzers work with this change compared to array with documented types of keys? Psalm for example can introspect even the types on those assoc arrays
- How is the usage for IDEs here (for developers)? Can this easily be understood here on what arguments are available and how many are used etc?

robertpustulka:
@dereuromark I'm not able to answer your questions but let it be clear: this is not a breaking change. You can still use options array as before. This change allows you to pass anything to the `find()` method after a finder name.

For me personally being able to leverage PHP 7 type checking instead doing it manually in every finder would be a great feature and I'd be willing to sacrifice a bit of code discoverability. But this change does not force you to write finders the new way, it just opens a path for being able to write less error-prone code.

ADmad (commented):
The fact that when 2nd argument to `Table::find()` is an array it's treated as a special case to maintain BC is my only concern with this change. I hope that doesn't trip users.

Also shouldn't `BehaviorRegistry::callFinder()` also be updated similar to `Table::classFinder()`?

robertpustulka:
> The fact that when 2nd argument to Table::find() is an array it's treated as a special case to maintain BC is my only concern with this change. I hope that doesn't trip users.

Only if it isn't passed. PHP will throw an error if some other type is expected in finder so it should be easy to catch?

> Also shouldn't BehaviorRegistry::callFinder() also be updated similar to Table::classFinder()?

Good point.

ADmad:
> Looks like it doesn't need to. BehaviorRegistry::callFinder() calls call_user_func_array internally.

A test with custom finder in a behavior which takes 2 args would be nice.

tests/test_app/TestApp/Model/Table/CustomCommentsTable.php
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    /**
+     * Finder using variadic args.
+     *
+     * @param Query $query Query
+     * @param bool $published Whether or not a comment is published.
+     * @return Query
+     */
+    public function findPublished(Query $query, bool $published): Query
```
ADmad: I probably wasn't clear in previous comment. By "2 args" i meant 2 additional args besides `$query` :slightly_smiling_face:.

dereuromark:
I guess IdeHelper could check the Signature of the finder and adjust the annotation for it if we used real methods instead of magic string..
`->findActive(true)->findPublished($entity)...` etc

Any chance we can do that in the future? find('strings', nooneknowsargs) is really not helpful.

robertpustulka:
> I guess IdeHelper could check the Signature of the finder and adjust the annotation for it if we used real methods instead of magic string..

Not sure, I'm not using the `IdeHelper`. But if an IDE is able to parse option arrays based on a finder name it should be able to parse arguments as well?

tests/test_app/TestApp/Model/Table/CustomCommentsTable.php
robertpustulka: Working on it

ADmad:
@dereuromark The first `Table::find()` class would return a `ORM\Query` instance so the chaining you have showed won't work if `findActive()`, `findPublished()` are table methods.

@robertpustulka There are few places like the paginator, associations etc, config where we can specify custom finder to use using the `finder` keys. Those should also be updated to be able to benefit from this new feature.

robertpustulka:
> There are few places like the paginator, associations etc, config where we can specify custom finder to use using the finder keys.

Can you point me to those places? I'm not sure how it is related?

ADmad:
> If those finder keys are arrays they should continue to work. All tests pass.

Yes they will continue to work since BC is maintained :). But they won't be able to benefit from this feature which allows passing multiple args to the finder.

https://github.com/cakephp/cakephp/blob/f3483c1721656cb81146039cf840c6e032e91759/src/ORM/Association.php#L738-L741

I am hoping change could be done so that `->find($finder, $opts)` could potentially be changed to `            ->find($finder, ...$opts)`.

robertpustulka:
@ADmad `$opts` are usually associative arrays and you can't make sure that the order of parameters is correct so I'm not sure it is possible for associative arrays to be passed as variadic arguments.

dereuromark:
> The first Table::find() call would return a ORM\Query instance so the chaining you have showed won't work if findActive(), findPublished() are table methods

Exactly - they would not be.
They would be magic methods invoked like find(string) is currently, and then proxying into the actual finders (protected, underscored etc). This way you can annotate them properly, so both static analyzers and IDE experience is actually working here - compared to currently that would be compromised with the current solution here.

src/ORM/Query.php
```diff
@@ -1178,12 +1178,15 @@ protected function _addDefaultSelectTypes(): void
      *
      * @see \Cake\ORM\Table::find()
      */
-    public function find(string $finder, array $options = [])
+    public function find(string $finder, ...$args)
     {
         /** @var \Cake\ORM\Table $table */
         $table = $this->getRepository();
+        if (!$args) {
```
ADmad: This is sad. Those who have overridden `find()` in their models won't realize they will have to add similar `if` block to their method.

robertpustulka: What is a chance that someone overriden ORM `Query::find()` and used it instead? `Cake\ORM\Query` is more like internal class, isn't it?

ADmad: Not much chance of overridden `Query::find()`. But I see similar `if` block in `Table::find()` as well a table class of test app which has overridden `Table::find()`.

src/ORM/Table.php
```diff
@@ -2361,21 +2365,28 @@ public function hasFinder(string $type): bool
...
+     * @param array $args List of options to pass to the finder
      * @return \Cake\ORM\Query
      * @throws \BadMethodCallException
      */
-    public function callFinder(string $type, Query $query, array $options = []): Query
+    public function callFinder(string $type, Query $query, ...$args): Query
     {
-        $query->applyOptions($options);
-        $options = $query->getOptions();
+        if ($args) {
```
ADmad: Perhaps use `if (isset($args[0]) && is_array($args[0])) {` and avoid the nested `if` below?

dereuromark:
ping @robertpustulka 
Do you still want to land this into 4.x?

robertpustulka:
Yes. Does it look OK as it is or do we need another approach? I'd like to know before I fix conflicts and add tests.

othercorey:
I like the idea of this, but the interface is awkward now.

I would prefer to see query options pulled out as a separate parameter.  Treating the first argument in a variable list as the query options parameter causes a lot of one-off and note style documentation that is hard to read.

The fact that they were merged with other general table-specific finder options is more or less a miracle that just happens to work and feels nicely hidden because everyone can just read examples or docs of a single array argument ... there's never a special case where someone reads an example without an options array as the first argument.

robertpustulka:
@othercorey 

> I would prefer to see query options pulled out as a separate parameter. Treating the first argument in a variable list as the query options parameter causes a lot of one-off and note style documentation that is hard to read.

I'm not sure I understand you, but you can pass anything to the query with the proposed interface. The first argument is treated as query options by default only for the BC.

Eg:
```php
$this->find('foo', ['foo' => 'bar']); //the BC way
$this->find('bar', $foo, $bar); // $foo and $bar ca be anything and get passed to the finder as below:

public function findBar($query, $foo, $bar) {}
```

The arrayish options can still be pulled from the query object with `$query->getOptions()`.

I agree that the interface in the 3.x was a bit hacky with mixed finder and general options but this have been done to keep the BC with 2.x, where you called finders like:

```php
$this->Articles->find('all', [
    'conditons' => [...],
]);
```

Maybe it's too late now to do this for 4.x, but maybe with 5.0 we should remove the ["special" options that are proxied to the query methods](https://github.com/cakephp/cakephp/blob/4.x/src/ORM/Query.php#L793-L818) (like `conditions`, `fields` etc)?

othercorey:
> I'm not sure I understand you, but you can pass anything to the query with the proposed interface. The first argument is treated as query options by default only for the BC.

I'm saying the backwards compatible part is what makes this extremely awkward.

robertpustulka:
@othercorey OK. Do you have an idea how to handle this more conveniently?

othercorey:
> @othercorey OK. Do you have an idea how to handle this more conveniently?

I would require calling a reserve `findXXX()` method with an options parameter to start the find chain.

```
findByArray($options)->find('custom', $param1, $param2)
```

- `findByArray()` should be a declared method, not a magic method.
- Require the $options array passed to `findByArray()` to only contain supported keys and throw a deprecation warning when it contains more.
- Require $param1 to not contain the `findByArray()` $option keys and throw a deprecation warning when it contains them.

This will enforce splitting the query options from find parameters while letting the user know they need to upgrade.

robertpustulka:
With this solution you'd need to locate and replace all `find('all', [..])` calls in existing code with `findByArray()`. I'm afraid that would be an overkill.

I have another idea - maybe we should have a new find method (both on repository and query) to support the new API? This way we could still use the old API with option array and the new one with variadic arguments. This is can be introduced in a minor release without breaking people's code.

The new method could be named `findBy` so this stays in line with old `findByFieldName` convention?

```php
$this->findBy('foo', $object1, $object2)->find('bar', $array);
```

othercorey:
How is that any different?  The finder names will conflict with existing methods.

robertpustulka:
How would they conflict?

othercorey:
> With this solution you'd need to locate and replace all `find('all', [..])` calls in existing code with `findByArray()`. I'm afraid that would be an overkill.


Only for those that pass in query options.

robertpustulka:
> Only for those that pass in query options.

Yes, but that is still a big breaking change in all community plugins and userland code. Please note that this applies not only to the pre 3.x find calls with `conditions` etc options but to the all finders including custom ones that use option arrays.

othercorey:
> Yes, but that is still a big breaking change in all community plugins and userland code. Please note that this applies not only to the pre 3.x find calls with `conditions` etc options but to the all finders including custom ones that use option arrays.

Well, you can fix the mess or not, lol.

robertpustulka:
I don't think that it's a mess. I don't even think that option arrays are inconvenience, because a lot of people like "named" arguments for options. With #12784 I just wanted to introduce more convenient way of passing objects into finders while leveraging built in PHP7.1+ type checking.

othercorey:
Just offering my opinion that mixing query options + custom finder options in an array that's treated as a special parameter in variadic magic finder methods is really awkward.  

But, I it isn't up to me.

robertpustulka:
OK I see that's awkward but it had to be done one way or the other to keep the BC. But now I realise that the variadic parameter support can be introduced as a new method without breaking the existing API and the BC.

I'll come up with a new PR soon but this will probably target 4.1 now.

@ADmad please remind me - we can still introduce new method to interfaces in minors with 4.x?

dereuromark:
Usually you would soft add new methods through docblock.

ADmad:
> please remind me - we can still introduce new method to interfaces in minors with 4.x?

We can't update the interface in minor. But we could add the new method to Table class and add annotation to the RepositoryInterface as a compromise.

ADmad:
@robertpustulka Instead of making `find()` (or a new method) support variadic args why not just add a `findX()` method to your table/behavior with the arguments you want? That would also allow you to properly typehint the args.

While Cake conventionally uses methods named like `findX()` as custom finder and expects a specific signature, it doesn't really matter if you don't try to use your custom methods as finders. Or you can simply chose to not use `find` prefix.

robertpustulka:
I can’t chain finders using custom method.

ADmad:
True that. If only the ORM required `Query::all()` to be called explicitly, then `Query::__call()` could be changed to proxy to repository methods instead of resultset decorator methods :slightly_smiling_face:.

markstory:
I didn't mean for this to get closed, I was removing the now stale 4.x branch. If you could open the pull request against master that would be appreciated as I cannot change the base branch now.

PR #12788, opened 2018-12-07
Support multiple table locations.

robertpustulka:
This change makes it easier to use the ORM in custom setups or outside CakePHP apps.
It allows to configure the locator to look for tables in non standard (`Model/Table`) namespace.

I'd like to use the ORM alone in the Infrastructure layer of my future app keeping the app logic and business logic free of any framework dependency.

### DIFF
--- a/src/ORM/Locator/TableLocator.php
+++ b/src/ORM/Locator/TableLocator.php
@@ -348,7 +348,7 @@ class TableLocator implements LocatorInterface
     /**
      * Adds a namespace where tables should be looked for.
      *
-     * @param string $namespace Namespace to add.
+     * @param string $namespace Namespace to add.
      * @return $this
      */
     public function addNamespace($namespace)
--- a/src/ORM/Locator/TableLocator.php
+++ b/src/ORM/Locator/TableLocator.php
@@ -66,7 +66,7 @@ class TableLocator implements LocatorInterface
     /**
      * Constructor.
      *
-     * @param array|null $namespaces Namespaces where tables should be located located.
+     * @param array|null $namespaces Namespaces where tables should be located.
      *   If none provided, the default `Model\Table` under your app's namespace is used.
      */
     public function __construct(array $namespaces = null)
@@ -77,7 +77,9 @@ class TableLocator implements LocatorInterface
             ];
         }
 
-        $this->_namespaces = $namespaces;
+        foreach ($namespaces as $namespace) {
+            $this->addNamespace($namespace);
+        }
     }
 
     /**
@@ -341,5 +343,19 @@ class TableLocator implements LocatorInterface
             $this->_config[$alias],
             $this->_fallbacked[$alias]
         );
+    }
+
+    /**
+     * Adds a namespace where tables should be looked for.
+     *
+     * @param string $namespace Namespace to add.
+     * @return $this
+     */
+    public function addNamespace($namespace)
+    {
+        $namespace = str_replace('\\', '/', $namespace);
+        $this->_namespaces[] = trim($namespace, '/');
+
+        return $this;
     }
 }
--- a/src/ORM/Locator/TableLocator.php
+++ b/src/ORM/Locator/TableLocator.php
@@ -32,7 +32,7 @@ class TableLocator implements LocatorInterface
      *
      * @var array
      */
-    protected $_locations = [];
+    protected $locations = [];
 
     /**
      * Configuration for aliases.
@@ -273,7 +273,7 @@ class TableLocator implements LocatorInterface
             return $options['className'];
         }
 
-        foreach ($this->_locations as $location) {
+        foreach ($this->locations as $location) {
             $class = App::className($options['className'], $location, 'Table');
             if ($class !== false) {
                 return $class;
@@ -354,7 +354,7 @@ class TableLocator implements LocatorInterface
     public function addLocation($location)
     {
         $location = str_replace('\\', '/', $location);
-        $this->_locations[] = trim($location, '/');
+        $this->locations[] = trim($location, '/');
 
         return $this;
     }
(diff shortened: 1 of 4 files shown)


robertpustulka:
Can it go to 3.7 or is it too late?

dereuromark:
This seems fully BC, so adding it in 3.7 until tomorrow could be fine.

lorenzo (commented):
We should also have a method to register a new namespace after the locator has been created

src/ORM/Locator/TableLocator.php
```diff
@@ -243,7 +269,18 @@ protected function _getClassName($alias, array $options = [])
...
         }
 
-        return App::className($options['className'], 'Model/Table', 'Table');
+        if (class_exists($options['className'])) {
+            return $options['className'];
+        }
+
+        foreach ($this->_namespaces as $namespace) {
+            $class = App::className($options['className'], $namespace, 'Table');
+            if ($class !== null) {
```
ADmad: App::className() returns false on failure.

src/ORM/Locator/TableLocator.php
```diff
@@ -243,7 +269,18 @@ protected function _getClassName($alias, array $options = [])
             $options['className'] = Inflector::camelize($alias);
         }
 
-        return App::className($options['className'], 'Model/Table', 'Table');
+        if (class_exists($options['className'])) {
```
ADmad: How about checking if the string contains backslash before doing class_exists() check?

src/ORM/Locator/TableLocator.php
```diff
@@ -307,4 +344,18 @@ public function remove($alias)
             $this->_fallbacked[$alias]
         );
     }
+
+    /**
+     * Adds a namespace where tables should be looked for.
```
ADmad: Having a usage e.g. would be nice here to clarify that one needs to specify namespace without the top level. For e.g. just `Infrastructure\Table` namespace `TestPlugin\Infrastructure\Table`.

robertpustulka: Yes. Maybe "namespace" is a wrong word as it suggests the whole namespace should be used. Maybe "location" is better?

ADmad: Yes, "location" or "path" would be better.

ADmad: Since it's in TableLocator I guess `$location` is nicer.

src/ORM/Locator/TableLocator.php
```diff
@@ -28,11 +28,11 @@ class TableLocator implements LocatorInterface
 {
 
     /**
-     * Contains a list of namespaces where table classes should be looked for.
+     * Contains a list of locations where table classes should be looked for.
      *
      * @var array
      */
-    protected $_namespaces = [];
+    protected $_locations = [];
```
ADmad: Nitpick: We don't use underscore prefix anymore for protected props/methods.

tests/test_app/Plugin/TestPlugin/src/Infrastructure/Table/AddressesTable.php
```diff
@@ -0,0 +1,23 @@
+<?php
+declare(strict_types=1);
```
ADmad: We don't use strict typing in 3.x.  Same for the other test class added.

tests/test_app/Plugin/TestPlugin/src/Infrastructure/Table/AddressesTable.php
```diff
@@ -0,0 +1,23 @@
...
+/**
+ * CakePHP(tm) : Rapid Development Framework (https://cakephp.org)
+ * Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
+ *
+ * Licensed under The MIT License
+ * Redistributions of files must retain the above copyright notice
+ *
+ * @copyright     Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
+ * @since         3.6.9
```
ADmad: 3.7.0

src/ORM/Locator/TableLocator.php
```diff
@@ -56,6 +63,25 @@ class TableLocator implements LocatorInterface
...
 
+    /**
+     * Constructor.
+     *
+     * @param array|null $locations Locations where tables should be looked for.
+     *   If none provided, the default `Model\Table` under your app's namespace is used.
+     */
+    public function __construct(array $locations = null)
+    {
+        if ($locations === null) {
```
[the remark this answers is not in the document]
robertpustulka: I wanted to default to `Model/Table` if no `$locations` is passed for BC. No value I mean null, not an empty array.

tests/test_app/Plugin/TestPlugin/src/Infrastructure/Table/AddressesTable.php
robertpustulka: Sorry, I changed target from 4.x to 3.x and forgot to remove those.

src/ORM/Locator/TableLocator.php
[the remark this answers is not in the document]
markstory: What happens when someone provides `[]` though?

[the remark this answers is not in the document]
markstory: ok.

robertpustulka:
It didn't make it for 3.7.1 release. Will it be merged before 3.7.2? I'm asking because I'm too lazy to change `@since` tags twice :)

ADmad:
@robertpustulka We no longer add new features to bugfix releases. So this would either go to `3.8` or `4.0`.

markstory:
@robertpustulka I apologize for not getting this merged earlier, but it was opened pretty late in the RCs and I missed it.

robertpustulka:
OK no problem, should it target 4.x then?

markstory:
It looks like we will end up doing a small release for 3.8, which also feeds into 4.0.

PR #12795, opened 2018-12-11
Update README.md

EOM:
change now version 3.7 igual website

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ recommend using the [app skeleton](https://github.com/cakephp/app) as
 a starting point. For existing applications you can run the following:
 
 ``` bash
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -38,7 +38,7 @@ recommend using the [app skeleton](https://github.com/cakephp/app) as
 a starting point. For existing applications you can run the following:
 
 ``` bash
-$ composer require cakephp/cakephp:"~3.6"
+$ composer require cakephp/cakephp:"~3.7"
```
ADmad: It would be better to use `3.7.*` as we do in skeleton app or simply remove the version from the command.

ADmad:
Closing in favor of #12813.

Issue #12793, opened 2018-12-10
Routes with skipped dashes


andruhapx:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [x] feature-discussion (RFC)

* CakePHP Version: 3.5.17
* Platform and Target: LAMP

### What you did
Added method UsersController::forgotPassword().
It can be reached by /users/forgot-password or /users/forgotPassword URIs

### What happened
/users/forgotpassword produces Cake\View\Exception\MissingTemplateException

### What you expected to happen
/users/forgotpassword should produce Cake\Controller\Exception\MissingActionException
Actually maybe its also correct to produce MissingActionException at /users/forgotPassword?

ADmad:
Since methods in PHP are case insensitive and the only way to avoid this issue is to not call `$route->fallbacks()` which connects the routes for patterns `/:controller/:action` and `/:controller/:action/:id` and explicitly connect routes for each controller action.

Or use your own route class which inflects dashed url fragments to underscored form instead of camelBased and then use underscored method names in your controller.

ADmad:
One way to ensure case sensitive method name check could be to get public methods of the controller using `ReflectionClass::getMethods()` and then check for exact match of action name using `in_array()`.

garas:
Getting full list of methods is unnecessary.

```php
class A {
    public function abcDef() {}
}
$method = new ReflectionMethod('A', 'AbCdeF');
echo $method->getName(); // echoes 'abcDef'
```

dereuromark:
I used to have my own component to check the URL against the expected one and fail with 404 or redirect 301 to proper target based on config: https://github.com/dereuromark/cakephp-seo/blob/master/Controller/Component/SeoComponent.php#L246
Would be nice if we had sth like this easier to hook into for SEO reasons.

ADmad:
@garas That's nice, so that could be used.

ADmad:
> Most of the time, a canonical tag however does the job, too :)

Yup, that's what I do. Avoid having to worry about URL casing :)

dereuromark:
Implemented in 4.x. For 3.x it can easily be shimmed with custom code to verify the casing in Middleware or alike.

PR #12796, opened 2018-12-11
Ensure case sensitive method name matching when invoking action.

ADmad:
Refs #12793.


markstory (approved):
Looks good to me. This change should be added to the migration guide too.

dereuromark:
If we do this, shouldnt we also care about the controller, plugin and prefix?
Or is this already fixed by design - or not relevant?

ADmad:
For controller/plugin/prefix it needs to match the FQCN which is case sensitive, hence no problem there.

dereuromark:
:+1: 

The checks fails seem unrelated.

PR #12802, opened 2018-12-11
Always use "?" to specify query string params in URL array.

ADmad:
The current behavior of treating any unknown key in URL array as query string param is unclean and unsafe. This changes forces the use of `'?'` as key to specify query string params.

Mixing query string param keys with special keys like `'controller'` etc. and custom route elements is not good. It opens the possibility of key collision and causing route mismatch or generating unexpected URLs.

I have seen people do `array_merge(['controller' => 'Foo', 'action' => 'bar'], $this->request->getQuery())` which can lead to the special keys and route elements being overwritten.

This change might cause some grief to those upgrading from 3.x but I think it's worth it.


ADmad:
I'll fix up the remaining tests once the change is approved.

markstory:
This will cause a bunch of problems for folks as they upgrade. I think it is a net positive change, but it will be a bumpy one that we'll hear about again later.

dereuromark:
Cant they shim it by providing a custom fallback that merges the old way if they want to?

ADmad:
Even if we were to provide a `LegacyRoute` class everyone would have to modify their custom route classes to extend it. Not very practical.

We could add a boolean static property like `Route::$legacyMode` which users can set to `true` if they want to continue using the old style merging.

ADmad:
This is good to go. The static analysis job failure is unrelated.

P.S. The `PaginatorHelper` does mind boggling amount of array juggling.

src/View/Helper/PaginatorHelper.php
```diff
@@ -313,9 +313,12 @@ protected function _toggledLink($text, $enabled, $options, $templates): string
...
         $paging = $this->params($options['model']);
 
-        $url = array_merge(
+        if (!isset($options['url']['?'])) {
+            $options['url']['?'] = [];
+        }
+        $url = Hash::merge(
             $options['url'],
-            ['page' => $paging['page'] + $options['step']]
+            ['?' => ['page' => $paging['page'] + $options['step']]]
```
markstory: This will have the drawback of making application routes not able to place page and sort parameters in the path.

src/View/Helper/PaginatorHelper.php
```diff
@@ -531,39 +536,54 @@ public function generateUrlParams(array $options = [], ?string $model = null): a
...
-        if (isset($paging['sortDefault'], $paging['directionDefault'], $url['sort'], $url['direction']) &&
-            $url['sort'] === $paging['sortDefault'] &&
-            strtolower($url['direction']) === strtolower($paging['directionDefault'])
+        $url = Hash::merge($url, $options);
+
+        if (!empty($url['?']['page']) && $url['?']['page'] === 1) {
+            $url['?']['page'] = null;
+        }
+        if (isset($paging['sortDefault'], $paging['directionDefault'], $url['?']['sort'], $url['?']['direction']) &&
+            $url['?']['sort'] === $paging['sortDefault'] &&
```
markstory: Given that we have a limited number of pagination parameters would it be possible to shift only those keys into `?` in `generateUrl`? Or do scopes make that complicated?

src/View/Helper/PaginatorHelper.php
ADmad: True, but now we can't simply merge as before either and it would cause route mismatch by default when you haven't setup any routes with the paging params as placeholders.

Solution would be to allow configuring which paging params should be treated as routes placeholders.

src/View/Helper/PaginatorHelper.php
ADmad: The scoping feature surely makes things difficult. It adds too much complexity for a feature needed/used by a few.

ADmad: It might help to have separate arguments in `generateUrl()` for the URL and paging params.

markstory: That sounds like a great idea. :+1:

src/View/Helper/PaginatorHelper.php
```diff
@@ -53,11 +53,14 @@ class PaginatorHelper extends Helper
...
-     * - `url['sort']` the key that the recordset is sorted.
-     * - `url['direction']` Direction of the sorting (default: 'asc').
-     * - `url['page']` Page number to use in links.
+     * - `url['?']['sort']` the key that the recordset is sorted.
+     * - `url['?']['direction']` Direction of the sorting (default: 'asc').
+     * - `url['?']['page']` Page number to use in links.
      * - `model` The name of the model.
      * - `escape` Defines if the title field for the link should be escaped (default: true).
+     * - `routePlaceholders` An array specifying which paging params should be
+     *   use as route placeholder instead of query string param. The array
```
markstory:
```suggestion
     *   passed as route placeholders instead of query string parameters. The array
```

src/View/Helper/PaginatorHelper.php
```diff
@@ -502,88 +502,95 @@ public function sort(string $key, $title = null, array $options = []): string
...
+        }
+
+        if (!isset($url['?'])) {
+            $url['?'] = [];
+        }
+
+        if (!empty($this->_config['options']['routePlaceholders'])) {
+            $placeholders = array_flip($this->_config['options']['routePlaceholders']);
+            $url += array_intersect_key($options, $placeholders);
+            $url['?'] += array_diff_key($options, $placeholders);
```
markstory: Should the placeholder parameters be added to the query string as well?

tests/TestCase/View/Helper/PaginatorHelperTest.php
```diff
@@ -1767,6 +1772,55 @@ public function testNumbers()
         $this->assertHtml($expected, $result);
     }
 
+    /**
+     * Test for URLs with paging params as route placeholders instead of query string params.
+     *
+     * @return void
+     */
+    public function testRoutePlaceholder()
```
markstory: :clap: for having test coverage on this.

src/View/Helper/PaginatorHelper.php
ADmad: I don't think so. Those using the paging params as URL fragment do so because they want prettier URLs without query string.

tests/TestCase/View/Helper/PaginatorHelperTest.php
```diff
@@ -777,13 +777,13 @@ public function testDefaultSortAndNoSort()
...
-            '/a',
-        ];
-        $this->assertHtml($expected, $result);
+        // $result = $this->Paginator->sort('controller');
+        // $expected = [
+        //     'a' => ['href' => '/index?sort=controller&amp;direction=asc'],
+        //     'Controller',
+        //     '/a',
+        // ];
+        // $this->assertHtml($expected, $result);
```
markstory: Should this be removed instead?

ADmad: Oops, that was temporary. Un-commented again.

tests/TestCase/View/Helper/PaginatorHelperTest.php
```diff
@@ -1767,6 +1772,55 @@ public function testNumbers()
...
+        Router::reload();
+        Router::connect('/:controller/:action/:sort/:direction');
+
+        $this->Paginator->options(['routePlaceholders' => ['sort', 'direction']]);
+        $result = $this->Paginator->sort('title');
+        $expected = [
+            'a' => ['href' => '/index/title/asc'],
+            'Title',
+            '/a',
+        ];
```
garas: Should be some assert here?

ADmad:
@markstory Ping

dakota:
Documentation (Including migration guide) was never updated to reflect this change.

ADmad:
@dakota https://github.com/cakephp/docs/pull/5914

PR #12803, opened 2018-12-11
Remove duplicate code from IntegerType

burzum:


markstory (approved):
Seem ok to me. This will add more method calls which in aggregate can add up, but I don't think that will matter in practical use cases.

PR #12807, opened 2018-12-12
Don't marshal random non empty strings to true.

ADmad:


dereuromark (approved):

markstory:
The PHPstan build has been failing for a bit but I'm not sure its pointing out any real problems.

ADmad:
The static analysis build seems to be failing due to psalm update. I'll fix it on `4.x`.

Issue #12804, opened 2018-12-11
schema generate shell fails on enum column


bancer:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 2.10.13.
* Platform and Target: Apache, MySQL.

### What you did
Run `./Console/cake schema generate`.

### What happened
The shell exited with errors:
```
val@val-desktop:~/git/app$ ./Console/cake schema generate

Welcome to CakePHP v2.10.13 Console
---------------------------------------------------------------
App : app
Path: /home/val/git/app/
---------------------------------------------------------------
Cake Schema Shell
---------------------------------------------------------------
Generating Schema...
Notice Error: Schema generation error: invalid column type enum('option 1','option 2','option 3') for MyModel.my_column does not exist in DBO in [/home/val/git/Cake/Model/CakeSchema.php, line 628]

Error: mymodel model not constructed properly
#0 [internal function]: AppMyModel->__construct(Array)
#1 /home/val/git/Cake/Utility/ClassRegistry.php(169): ReflectionClass->newInstance(Array)
#2 /home/val/git/Cake/Model/CakeSchema.php(248): ClassRegistry::init('AppMyMo...')
#3 /home/val/git/Cake/Console/Command/SchemaShell.php(148): CakeSchema->read(Array)
#4 /home/val/git/Cake/Console/Shell.php(459): SchemaShell->generate()
#5 /home/val/git/Cake/Console/ShellDispatcher.php(219): Shell->runCommand('generate', Array)
#6 /home/val/git/Cake/Console/ShellDispatcher.php(66): ShellDispatcher->dispatch()
#7 /home/val/git/Cake/Console/cake.php(54): ShellDispatcher::run(Array)
#8 {main}
```
(I changed some paths in the output above).

### What you expected to happen
According to the [cookbook](https://book.cakephp.org/2.0/en/console-and-shells/schema-management-and-migrations.html):
> This will generate a schema.php file in your app/Config/Schema directory

markstory:
Enums are very poorly support in 2.x. Fixing this borders on a new feature.

dereuromark:
PR open.

PR #12808, opened 2018-12-12
 Improve support of enum data type

bancer:
Resolves https://github.com/cakephp/cakephp/issues/12804. Also fixes errors in `./Console/cake schema create` when enum fields are used.

### DIFF
--- a/lib/Cake/Test/Case/Model/Datasource/Database/MysqlTest.php
+++ b/lib/Cake/Test/Case/Model/Datasource/Database/MysqlTest.php
@@ -3080,14 +3080,6 @@ SQL;
 		$expected = '5,2';
 		$this->assertSame($expected, $result);
 
-		$result = $this->Dbo->length("enum('test','me','now')");
-		$expected = 4;
-		$this->assertSame($expected, $result);
-
-		$result = $this->Dbo->length("set('a','b','cd')");
-		$expected = 2;
-		$this->assertSame($expected, $result);
-
 		$result = $this->Dbo->length(false);
 		$this->assertNull($result);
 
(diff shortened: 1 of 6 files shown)


lib/Cake/Test/Case/Model/Datasource/Database/MysqlTest.php
```diff
@@ -3100,6 +3092,26 @@ public function testLength() {
...
 	}
 
+/**
+ * Tests the length of enum column.
+ *
+ * @return void
+ */
+	public function testLengthEnum() {
+		$result = $this->Dbo->length("enum('test','me','now')");
+		$this->assertNull($result);
```
markstory: This seems like a potentially breaking change in behavior.

lib/Cake/Test/Case/Model/Datasource/DboSourceTest.php
```diff
@@ -2177,12 +2177,19 @@ public function testLength() {
...
+ * @return void
+ */
+	public function testLengthEnum() {
 		$result = $this->db->length('enum("one", "longer")');
-		$this->assertEquals(6, $result);
+		$this->assertNull($result);
 
 		$result = $this->db->length("enum('One Value','ANOTHER ... VALUE ...')");
-		$this->assertEquals(21, $result);
+		$this->assertNull($result);
```
markstory: While I understand that the old behavior may not have been entirely correct. That behavior has been there for quite some time and removing it now is pretty risky.

bancer: Fixing bugs always have some degree of risk. Enum column had to return `null` length in the first place like datetime and text columns do.

markstory: I agree that fixing bugs could break untested behavior. This seems more deliberate though. Previously the length was the length of the longest accepted element, now it is `null`.

bancer: And that was a bug. The length is used in defining the column type. So if the length is 15 then the column becomes `'type' => 'enum(15)',` instead of `'type' => 'enum(\'active\',\'deleted\')',`.  Should I add more tests of methods that use `Mysql::length()` method?

Issue #12114, opened 2018-05-22
RFC for refactoring Cake\Mailer in CakePHP 4.x


burzum:
By spending some time with DDD, trying to build a hexagonal app with Cake and looking at how others frameworks do it I came across these things I'm not happy with in CakePHPs mailer code when I was working on my infrastructure layer that is used to send emails.

## Architectural Issues

* `\Cake\Mailer\Email` is a [god object](https://en.wikipedia.org/wiki/God_object)
  * It's aware of the transport
  * It's aware of rendering
* It has ~3000 lines of code 😮 

## Proposed Change

* Add `\Cake\Mailer\Transport\TransportInterface`
* Add `\Cake\Mailer\MailerInterface`
* Add `\Cake\Mailer\` implements `MailerInterface`
* Add `\Cake\Mailer\EmailInterface`
* `\Cake\Mailer\Email` implements `EmailInterface`
* `\Cake\Mailer\Email` should be a pure data transfer object
* Move the rendering of an email to a dedicated class
  * Add an interface for that class
* Add `SendingFailedException`

## Benefits
* This will make testing of each single component a lot more easier
* Less coupling
* Might be opinionated but I think it's a lot easier to understand than having this god object

## Example

```php
$mail = new Email();
$mail->setSubject('testing')
    ->setReceiver('burzum@cakephp.org')
    ->setFrom('noreply@cakephp.org')
    ->setFormat(EmailFormat::BOTH);

// MailRenderer will call Email::setHtmlContent() and/or setPlainContent()
$mailRenderer = new MailRenderer();
// Takes only instances of EmailInterface
$mailRenderer->set(['foo' => 'bar'])->render($mail);

// Alternatives?
$email->setHtmlContent($myContent);
$email->setHtmlContent($mailRenderer->render());

$mailer = new Mailer(new Transport());
$mailer->send($mail);
```

We can use another class and using __call() as a proxy to all of the three classes above to combine them in a fluid interface as we have it right now but with proper abstraction.

```php
$mailer = new SomeMailer();
$mail
     // Not required, just if default needs to be changed
    ->setRender($renderer)
     // Not required, just if default needs to be changed
    ->setTransport($transport)
    ->setSubject('testing')
    ->setReceiver('burzum@cakephp.org')
    ->setFrom('noreply@cakephp.org')
    ->setViewVars(['foo' => 'bar'])
    ->send();
```

markstory:
[...]

PR #12810, opened 2018-12-12
4.x email renderer

ADmad:
Refs #12114.


src/Mailer/Email.php
```diff
@@ -250,6 +233,13 @@ class Email implements JsonSerializable, Serializable
      */
     protected $transferEncoding;
 
+    /**
+     * Email Renderer
+     *
+     * @var \Cake\Mailer\Renderer
```
dereuromark: |null

dereuromark (approved):
Apart from my comment above

src/Mailer/Renderer.php
```diff
@@ -0,0 +1,462 @@
...
+ */
+namespace Cake\Mailer;
+
+use Cake\Core\Configure;
+use Cake\Http\Client\FormDataPart;
+use Cake\Utility\Hash;
+use Cake\Utility\Security;
+use Cake\Utility\Text;
+
+class Renderer
```
markstory: This should have a class doc. Is this class a public one that folks can extend/replace/rely on or is it intended to be an internal one?

src/Mailer/Renderer.php
```diff
@@ -0,0 +1,462 @@
...
+     * @var \Cake\Mailer\Email
+     */
+    protected $email;
+
+    /**
+     * Constructor
+     */
+    public function __construct()
+    {
+        $this->appCharset = Configure::read('App.encoding');
```
markstory: Should this be a constructor argument?

src/Mailer/Renderer.php
ADmad: We could make it internal. I personally can't think of a use case where someone might want to extend/modify it.

src/Mailer/Renderer.php
ADmad: Probably. For now I had just copied from `Email`'s constructor.

src/Mailer/Renderer.php
```diff
@@ -0,0 +1,462 @@
...
+            $attachments = $this->attachFiles($boundary);
+            $msg = array_merge($msg, $attachments);
+        }
+        if ($hasAttachments || $hasMultipleTypes) {
+            $msg[] = '';
+            $msg[] = '--' . $boundary . '--';
+            $msg[] = '';
+        }
+
+        return [$msg, $boundary, $textMessage, $htmlMessage];
```
ADmad: I should change this to associative array.

robertpustulka (changes requested):
Could the view builder go to the renderer as well? It would make more sense to remove any rendering/templating stuff from the Email class.

If not possible, maybe we could split the Email class into Renderer and Message each with their own responsibilities and leave Email as a "facade" with BC API eventually dropping it in 5.x?

Ideally sending an email could look something like this:
```php
$renderer = new Renderer();
$renderer->setVar(...)->setTemplate(...);
$message = new Message();
$message->setFrom(...)->setTo(...)->setSubject()->setContent($renderer->getContent());

$transport->send($message);

//or hidden under BC Email "facade"
$email = new Email();
$email->setVar(...)->setTemplate(...)->setFrom(...)->setTo(...)->setSubject()->send();
```

ADmad:
> Could the view builder go to the renderer as well? It would make more sense to remove any rendering/templating stuff from the Email class.

Done which proxying relevant Email methods to the Renderer's ViewBuilder. (I'll take care of the docblocks and stuff later).

> If not possible, maybe we could split the Email class into Renderer and Message each with their own responsibilities and leave Email as a "facade" with BC API eventually dropping it in 5.x?

That would be ideal but I don't think I can bring myself to put in the extra work to achieve that :slightly_smiling_face:.

dereuromark (changes requested):
My comment has not been fixed yet. But yeah, can be done at the end.

robertpustulka:
> Done, with proxying relevant Email methods to the Renderer's ViewBuilder. (I'll take care of the docblocks and stuff later).

OK, looks good for now.

robertpustulka (approved):

ADmad:
> I can add Message class in another PR later.

Great, I'll wrap this up soon and pass on the baton to you :).

src/Mailer/Email.php
```diff
@@ -1507,7 +1507,6 @@ public static function deliver(
         if (is_array($config) && !isset($config['transport'])) {
             $config['transport'] = 'default';
         }
-        /* @var \Cake\Mailer\Email $instance */
```
dereuromark: Are you sure you should remove this?
The discovery of setTo() method can not work without it.

src/Mailer/Renderer.php
```diff
@@ -72,11 +72,13 @@ class Renderer
...
-     * Constructor
+     * Constructor.
+     *
+     * @param string|null $appCharset Application's character set.
      */
-    public function __construct()
+    public function __construct(?string $appCharset = null)
     {
-        $this->appCharset = Configure::read('App.encoding');
+        $this->appCharset = $appCharset ?? Configure::read('App.encoding');
```
dereuromark: In these cases it is probably better to use non cloaking `?:` check, but no big deal.

dereuromark (approved):

ADmad:
I think everyone's comments are addressed now.

markstory:
Looks good to me. This seems like a change that should be mentioned in the docs.

ADmad:
Usage wise this doesn't make much difference to end users. Only thing is they have to use `setViewVars()` instead of `set()` for setting view vars. I'll open a docs PR to update code examples accordingly and mention the removed/added methods for `Email` on 4.0 migration page.

PR #12813, opened 2018-12-13
Remove version number from composer require command

ADmad:
This avoids having to update README with each minor/major release.


dereuromark:
wouldnt that remove ~? no big deal though imo.

PR #12816, opened 2018-12-14
Fix docblocks and better IDE and static analyzer discoverery.

dereuromark:
I was surprised to see a public trait message not being part of the interface.
This should get fixed with 4.0 then.

### DIFF
--- a/src/Collection/CollectionTrait.php
+++ b/src/Collection/CollectionTrait.php
@@ -283,10 +283,12 @@ trait CollectionTrait
         $callback = $this->_propertyExtractor($callback);
 
         $mapper = function ($value, $key, $mr) use ($callback) {
+            /** @var \Cake\Collection\Iterator\MapReduce $mr */
             $mr->emitIntermediate($value, $callback($value));
         };
 
         $reducer = function ($values, $key, $mr) {
+            /** @var \Cake\Collection\Iterator\MapReduce $mr */
             $mr->emit(count($values), $key);
         };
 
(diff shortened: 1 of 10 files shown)


src/Collection/CollectionTrait.php
```diff
@@ -885,9 +892,10 @@ public function _unwrap()
     }
 
     /**
-     * {@inheritDoc}
```
ADmad: The `{@inheritDoc}` needs to be kept for inheriting the description.

src/Collection/Iterator/FilterIterator.php
```diff
@@ -43,7 +43,7 @@ class FilterIterator extends Collection
      * in the current iteration, the key of the element and the passed $items iterator
      * as arguments, in that order.
      *
-     * @param \Iterator $items The items to be filtered.
+     * @param \Iterator|\Traversable|array $items The items to be filtered.
```
ADmad: `Iterator` extends `Traversable`, so have both is redundant. `\Iterator` can be removed now.

ADmad (changes requested):

ADmad:
> I was surprised to see a public trait message not being part of the interface.
> This should get fixed with 4.0 then.

`CollectionInterface::cartesianProduct()` has already been added on 4.x.

src/Collection/CollectionTrait.php
dereuromark: there is nothing to inherit. i didnt find any code.

PR #12819, opened 2018-12-14
allow incorrect decimal / float formatting

### LINKED ISSUE #12800: Patching non well formed field of type decimal on 3.7.0
This is a (multiple allowed):

* [x] bug
* [ ] enhancement

* CakePHP Version: 3.7.0
* Change in commit: e960656f2a8806f705ec124558963bb1969034ad
* File causing issue: src/Database/Type/DecimalType.php:106

### What you did
Our code is trying to marshall non well formed decimal value (string using comma as decimal separator) into DecimalType database field, which worked up to change in commit mentioned above. 

### What happened
DecimalType method marshall returned null due to removal of !is_scalar($value) condition.

Method should return value if it's scalar as it did in previous version (up to specified commit) as values could be parsed to correct format by entity setters.

segy:
Fixes #12800

### DIFF
--- a/src/Database/Type/DecimalType.php
+++ b/src/Database/Type/DecimalType.php
@@ -157,11 +157,11 @@ class DecimalType extends Type implements TypeInterface, BatchCastingInterface
         if (is_numeric($value)) {
             return (float)$value;
         }
-        if (is_string($value) && preg_match('/[^0-9,. ]/', $value)) {
-            return null;
+        if (is_string($value) && preg_match('/^[0-9,. ]+$/', $value)) {
+            return $value;
         }
 
-        return $value;
+        return null;
     }
 
     /**
--- a/src/Database/Type/DecimalType.php
+++ b/src/Database/Type/DecimalType.php
@@ -157,8 +157,6 @@ class DecimalType extends Type implements TypeInterface, BatchCastingInterface
         if (is_numeric($value)) {
             return (float)$value;
         }
-
-        // allow custom decimal format (@see https://github.com/cakephp/cakephp/issues/12800)
         if (preg_match('/[^0-9,. ]/', $value)) {
             return null;
         }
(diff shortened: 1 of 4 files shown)


src/Database/Type/DecimalType.php
```diff
@@ -158,7 +158,12 @@ public function marshal($value)
             return (float)$value;
         }
 
-        return null;
+        // allow custom decimal format (@see https://github.com/cakephp/cakephp/issues/12800)
+        if (preg_match('/[^0-9,. ]/', $value)) {
+            return null;
```
ADmad: You have incorrectly swapped the return values :)

src/Database/Type/DecimalType.php
```diff
@@ -158,7 +158,12 @@ public function marshal($value)
             return (float)$value;
         }
 
-        return null;
+        // allow custom decimal format (@see https://github.com/cakephp/cakephp/issues/12800)
```
ADmad: Please remove this comment. You can put the issue link in related test case.

src/Database/Type/DecimalType.php
```diff
@@ -158,7 +158,12 @@ public function marshal($value)
...
+        if (preg_match('/[^0-9,. ]/', $value)) {
```
ADmad: Aren't there some locales which use single/double quotes too as thousands separator?

dereuromark: For degrees lat/lng maybe :) But for decimal I think this is fair enough here.

src/Database/Type/DecimalType.php
dereuromark: Sure? The ^ means not afaik.

segy: I hope it's correct

src/Database/Type/DecimalType.php
segy: done. moved to test cases

segy:
I wasn't checking the type. Fixed

ADmad:
Nope, the code now allows non scalar values through and tests still fail. We'll save time if you run the tests locally :slightly_smiling_face:.

ADmad (approved):

Issue #12817, opened 2018-12-14
[RFC] Form widget override/customize (input type date datetime-local month ...)


smarek:
This is a (multiple allowed):

* [] bug
* [x] enhancement
* [x] feature-discussion (RFC)

* CakePHP Version: 3.6.13
* Platform and Target: LNPM

This discussion was ongoing for the last years, as far as I can tell. Reference tickets might be:
https://github.com/cakephp/cakephp/issues/1888
https://github.com/cakephp/cakephp/issues/12537
https://github.com/cakephp/cakephp/issues/7254

and of course lots of stackoverflows and similar

Existing date/date-time widget is technically correct, working, however not user friendly.

Existing recommendations, such as input[type=text]+jQueryUI, [Form->text instead Form->control](https://github.com/cakephp/cakephp/issues/12537#issuecomment-418738334), or [string replacing type=text with type=date](https://discourse.cakephp.org/t/cakephp3-form-control-with-type-date/3290) are not satisfying neither systematic

[Documentation on topic of providing custom widgets](https://book.cakephp.org/3.0/en/views/helpers/form.html#adding-custom-widgets) allows users to create custom implementation, however that requires every Form control call to be annotated with correct widget type, which is again unpleasant.

Would it be possible, either by documentation or new feature, to override default widgets (either configuration override or subclassing of ie. [DateTimeWidget](https://api.cakephp.org/3.7/class-Cake.View.Widget.DateTimeWidget.html) ) to provide custom implementation to use on default field types, such as `date`, `timestamp`,  etc?

markstory:
Are HTML5 date inputs widely supported enough that we should use them instead of multiple select boxes in 4.0? Not having to support date arrays would allow us simplify internal date handling in meaningful ways and browser pickers are a better experience than multiple select boxes are.

smarek:
@markstory might be, but the RFC was more asking for replacing the basic widgets (not having to meddle with setting the text widget type for example, or similar), which is, as far as i can tell, currently impossible, due to the way, how those basic widgets are hard-wired in framework core.

HTML5 date/date-time widgets are not ultimate solution either, the design and accessibility is not always perfect, so being able to configure that on run-time would be nice.

Best in my opinion would be documentation, how to globally replace widget of type `date` or `datetime`, and similar, with custom implementation, so the way to solution of choice will be pretty and straightforward.

markstory:
[...]

PR #12820, opened 2018-12-14
WIP - 3.next - Allow easily generating browser's native date/time inputs

ADmad:
Refs #12817

### DIFF
--- a/src/View/Widget/DateTimeNativeWidget.php
+++ b/src/View/Widget/DateTimeNativeWidget.php
@@ -1,0 +1,110 @@
+<?php
+/**
+ * CakePHP(tm) : Rapid Development Framework (https://cakephp.org)
+ * Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
+ *
+ * Licensed under The MIT License
+ * For full copyright and license information, please see the LICENSE.txt
+ * Redistributions of files must retain the above copyright notice.
+ *
+ * @copyright     Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
+ * @link          https://cakephp.org CakePHP(tm) Project
+ * @since         3.0.0
+ * @license       https://opensource.org/licenses/mit-license.php MIT License
+ */
+namespace Cake\View\Widget;
+
+use Cake\View\Form\ContextInterface;
+use Cake\View\StringTemplate;
+use DateTime;
+use Exception;
+use RuntimeException;
+
+/**
+ * Input widget class for generating a date time input widget.
+ *
+ * This class is intended as an internal implementation detail
+ * of Cake\View\Helper\FormHelper and is not intended for direct use.
+ */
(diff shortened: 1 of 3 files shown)


ADmad:
Needs tests with values.

`datetime-local` needs values in form `yyyy-MM-ddThh:mm` so we probably need to add a widget for it to format the value as such.

ADmad:
This is ready for some feedback (Ignore missing/incorrect docblocks).

src/View/Widget/DateTimeNativeWidget.php
```diff
@@ -0,0 +1,130 @@
...
+use Exception;
+use RuntimeException;
+
+/**
+ * Input widget class for generating a date time input widget.
+ *
+ * This class is intended as an internal implementation detail
+ * of Cake\View\Helper\FormHelper and is not intended for direct use.
+ */
+class DateTimeNativeWidget implements WidgetInterface
```
markstory: I wonder if we should just make native date inputs the default for 4.0. HTML5 date input support has gotten much better with evergreen browsers and not having to support multiple selectboxes would let us save some complexity internally as well.

ADmad: I am fine with making native datetime inputs default in 4.0 and dropping the selects altogether. In practice most sites/apps use JS based pickers anyway for better consistency and usability.

markstory: We should retain the array marshalling for datetimes for backwards compatibility.  I bet somewhere someone is using array dates in an API.

ADmad: Keeping datetime array marshaling sounds reasonable.

So should I reopen this against `4.x` instead, with the helper and widget code for selects removed?

markstory: I think that is a good idea. We could move the old widgets into a plugin that folks can use if they really need that behavior.

ADmad: Okay, closing this then. Will open a new PR targeting 4.x soon.

markstory: What would you have liked to see?

markusramsak: I prefer using dropdown to select the year for birthday, which can allow to select "9999" for the year if someone prefers say his year of birth. Therefore for birthday I like the select dropdown version.

markstory: I think `year`, and `month` inputs could stay as select boxes. The native HTML inputs are better for `date` and `time` values though.

ADmad: If we are going to switch to native inputs then by default they should be used for all available types. There's no HTML input for `year` so keeping the select box for `year` is fine but there is one for `month` so for that native input should be used.

PR #12821, opened 2018-12-16
Update typehints.

ADmad:


src/TestSuite/IntegrationTestTrait.php
```diff
@@ -1227,9 +1227,9 @@ public function assertFileResponse(string $expected, string $message = ''): void
      * Inspect controller to extract possible causes of the failed assertion
      *
      * @param string $message Original message to use as a base
-     * @return null|string
+     * @return string
      */
-    protected function extractVerboseMessage($message = null)
+    protected function extractVerboseMessage(string $message = ''): string
```
dereuromark: In this case we dont even need the default value `''` IMO. A message has to be passed for this method to make sense (concat operation).

src/Validation/Validator.php
```diff
@@ -430,7 +430,7 @@ public function count(): int
      * @param array|\Cake\Validation\ValidationRule $rule the rule to add
      * @return $this
      */
-    public function add(string $field, $name, $rule = [])
+    public function add(string $field, $name, $rule = []): self
```
dereuromark: We must never add self etc for chainable. You shoot yourself in the foot here.
There is also no need to protect anyone, we all know what return $this in docblock means, the IDE does too.
Code sniffer takes care about the rest.

dereuromark (approved):
Looks good otherwise

src/View/View.php
```diff
@@ -489,7 +489,7 @@ public function enableAutoLayout(bool $enable = true): self
      *
      * @return $this
      */
-    public function disableAutoLayout()
+    public function disableAutoLayout(): self
```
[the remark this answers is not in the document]
dereuromark: Yeah, those should all be removed as they only kill usability, and provide no gain here.

Issue #9794, opened 2016-11-24
Separating the IntegrationTestCase out into a Trait


ascii-soup:
This is a (multiple allowed):

* [ ] bug
* [x] enhancement
* [x] feature-discussion (RFC)

* CakePHP Version: 3.3.7

We are about to embark on some integration testing in a project we're working on. For unit testing we have our own base TestCase classes which we need to do the integration testing too. However, the IntegrationTestCase is a class that contains all its behaviour which means we can't re-use it very easily.

I wondered if it would be possible to move the functionality of IntegrationTestCase into a trait so that it can be easily shared amongst custom base TestCase classes. I imagine the IntegrationTestCase class remaining, but having its contents changed to something like the following:

```php
<?php

abstract class IntegrationTestCase extends TestCase
{
    use CookieCryptTrait;
    use SecureFieldTokenTrait;
    use IntegrationTestTrait; // the functionality would simply be moved here
}
```

I'm happy to do the work involved and submit a PR but I wanted to run it past you first to see if the idea is at least sound in theory. My limited initial testing seems to work but it would need more testing if it were to happen.

Any comments, suggestions, reasons it wouldn't work?

markstory:
I think if we can extract a trait that doesn't have dependencies on other traits or inherited methods then this sounds pretty reasonable.

swiffer:
@markstory 

Moving from `...extends IntegrationTestCase` to `use IntegrationTestTrait`

BREAKS:
```php
class SampleControllerTest extends TestCase
{
    use IntegrationTestTrait;

    public function setUp(){
        parent::setUp();
    }
}
```

POSSIBLE SOLUTION:
```php
class SampleControllerTest extends TestCase
{
    use IntegrationTestTrait {
        tearDown as traitTearDown;
        setUp as traitSetUp;
    };

    public function setUp(){
        self::traitSetUp();
    }
}
```

Should this be mentioned somewhere in the docs / migration guide?

I'm using setUp for additional preperation...

e.g.
```php
        // Fake out SSL connections.
        $this->configRequest([
            'environment' => ['HTTPS' => 'on']
        ]);
```

markstory:
We could use the `before` and `after` annotations in the trait. That would remove the need for workarounds or documentation.

swiffer:
what would it look like in my TestCase?

markstory:
You would continue using setup/tear down as usual, and not need to do any method aliasing for the trait.

PR #12822, opened 2018-12-16
Use annotitations for integration trait setup/teardown

markstory:
Using setup/teardown as the method names means that implementing classes lose the setup/teardown logic defined in the trait when they redeclare those methods. By using different method names and annotations phpunit will run the implementing class's setup and ours.

Targeting this to master as we recommend people use the trait in our docs and bake now generates tests using the trait.

Refs #9794


jeremyharris:
Thanks for this, it was a complete oversight on my part.

PR #12824, opened 2018-12-17
Enable subcommands to use spaces.

markstory:
Right now explicit subcommands must use a non-space character. For example `bake:test` and `bake.test` work, however `bake test` will not work. This change enables commands up with up to 3 spaces to work. I limited to 3 as I don't think more than 3 will be frequently used and I wanted to cut down on wasted work.

### DIFF
--- a/src/Console/CommandCollection.php
+++ b/src/Console/CommandCollection.php
@@ -64,6 +64,12 @@ class CommandCollection implements IteratorAggregate, Countable
                 "Cannot use '$class' for command '$name' it is not a subclass of Cake\Console\Shell or Cake\Console\Command."
             );
         }
+        if (!preg_match('/^[\p{L}\p{N}\-_:.]+(?:(?: [\p{L}\p{N}\-_:.]+){1,2})?$/ui', $name)) {
+            throw new InvalidArgumentException(
+                "The command name `{$name}` is invalid. " .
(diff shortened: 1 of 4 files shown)


saeideng:
>I limited to 3

could we add/show a notice/error when adding a command for more than 3 subs?

or showing a message on out put of `bin/cake`?
```
#bin/cake 

Available Commands:
- benchmark
- help
- test sub sub2 sub3 ( more than 3 subs does not support)
- version
```

markstory:
@saeideng Would that error show when the command is added or when commands are running? I would prefer to prevent invalid commands from being added to the collection personally.

saeideng:
>I would prefer to prevent invalid commands from being added to the collection personally.

and I :+1:

dereuromark:
Are we transforming other punctuations into this format?
I am a bit afraid of having `.`, `:` and ` ` usage all in the CLI, for multiple plugins and multiple commands etc.
Would create quite the mess.
Better to unify maybe?

markstory:
> Are we transforming other punctuations into this format?

No punctuation is being transformed. These changes make it possible for command names to have spaces used. Right now a command/shell could be added with a space in the name and never be callable.

The added validation attempts to prevent command being registered with more than 3 spaces in them. I'll simplify the pattern as it does look more complicated than it actually is.

PR #12833, opened 2018-12-21
Make PSR7 uploaded files accessible in integration tests

robertpustulka:
This fix makes it possible to test controller actions that expect uploads are accessible from `getUploadedFiles()`. 

Unfortunately files passed as POST data body are not present in `getUploadedFiles()` and to test those properly `$_FILES` had to be injected with files directly.

### DIFF
--- a/tests/TestCase/TestSuite/IntegrationTestTraitTest.php
+++ b/tests/TestCase/TestSuite/IntegrationTestTraitTest.php
@@ -514,6 +514,59 @@ class IntegrationTestTraitTest extends IntegrationTestCase
     }
 
     /**
+     * Test that the uploaded files are passed correctly to the request
+     *
+     * @return void
+     */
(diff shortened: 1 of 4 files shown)


tests/TestCase/TestSuite/IntegrationTestTraitTest.php
```diff
@@ -513,6 +513,59 @@ public function testPostDataHttpServer()
...
+                    'error' => [
+                        ['file' => 0],
+                        ['file' => 0]
+                    ],
+                    'size' => [
+                        ['file' => 17188],
+                        ['file' => 2010]
+                    ],
+                ],
+                'upload' => new UploadedFile(__FILE__, 42, 0)
```
markstory: This seems like something that should be possible.  Is making this work complicated?

PR #12836, opened 2018-12-22
Fix .json routes not rendering error pages properly

### LINKED ISSUE #12834: Exeption in json route no render json (cakephp 3.7)
This is a (multiple allowed):

* [x] bug
* [ ] enhancement

* CakePHP Version: 3.7.1 : 1f6fd21ca27ffc88fa88506a400b94a809e29254
* Platform and Target: apache php 7.0.12

### What you did
In fresh cakephp 3.7.1 install, I create an API.

In my route.php
```php
Router::scope('/', function (RouteBuilder $routes) {
    $routes->setExtensions(['json']);
	//...
});
```

In AppController
```php
public function initialize()
    {
        parent::initialize();

        $this->loadComponent('RequestHandler', [
            'enableBeforeRedirect' => false,
        ]);
        $this->loadComponent('Flash');

        /*
         * Enable the following component for recommended CakePHP security settings.
         * see https://book.cakephp.org/3.0/en/controllers/components/security.html
         */
        //$this->loadComponent('Security');
    }
```

My TestController
```php
<?php
namespace App\Controller;

use Cake\Http\Exception\NotFoundException;

class TestController extends AppController
{
    public function success()
    {
        $this->set([
            'test' => true,
            '_serialize' => ['test']
        ]);
    }

    public function error()
    {
        throw new NotFoundException();
    }
}
```

A simple view for test
```
<script>
    fetch('/Test/error.json')
    fetch('/Test/success.json')
</script>
```

### What happened

* /Test/success.json
success

* /Test/error.json
error

I would like, as with cakephp 3.6 for error.json

error2

markstory:
Due to changes in #11978 the original request state was preserved. This is a good thing as it makes control flow more explicit however, it has the side-effect of parsed extensions being 'lost' as the request error handler has access to is detached from the one Router mutated.

By using the Router state to apply routing if available we can fix #12834. I'm using the `controller` param as a hint for empty params as controller is a required router output key.

### DIFF
--- a/src/Error/ExceptionRenderer.php
+++ b/src/Error/ExceptionRenderer.php
@@ -125,9 +125,18 @@ class ExceptionRenderer implements ExceptionRendererInterface
      */
     protected function _getController()
     {
-        $request = $this->request ?: Router::getRequest(true);
+        $request = $this->request;
+        $routerRequest = Router::getRequest(true);
+        // Fallback to the request in the router or make a new one from
+        // $_SERVER
         if ($request === null) {
-            $request = ServerRequestFactory::fromGlobals();
+            $request = $routerRequest ?: ServerRequestFactory::fromGlobals();
+        }
+
+        // If the current request doesn't have routing data, but we
+        // found a request in the router context copy the params over
+        if ($request->getParam('controller') === false && $routerRequest !== null) {
+            $request = $request->withAttribute('params', $routerRequest->getAttribute('params'));
         }
 
         $response = new Response();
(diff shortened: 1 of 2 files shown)


src/Error/ExceptionRenderer.php
```diff
@@ -125,9 +125,18 @@ protected function _unwrap($exception)
...
+        // $_SERVER
         if ($request === null) {
-            $request = ServerRequestFactory::fromGlobals();
+            $request = $routerRequest ?: ServerRequestFactory::fromGlobals();
+        }
+
+        // If the current request doesn't have routing data, but we
+        // found a request in the router context copy the params over
+        if ($request->getParam('controller') === false && $routerRequest !== null) {
+            $request = $request->withAttribute('params', $routerRequest->getAttribute('params'));
```
saeideng: why not whole request ?
```
if ($request->getParam('controller') === false && $routerRequest !== null) {
            $request =$routerRequest;
```

markstory: I was concerned with overwriting headers or transformed data.

PR #12837, opened 2018-12-22
Fixes docblock for View::set()

josephzidell:
Adds `bool` and `int` for second param of `View::set()`

### DIFF
--- a/lib/Cake/View/View.php
+++ b/lib/Cake/View/View.php
@@ -807,7 +807,7 @@ class View extends CakeObject {
  * a layout or other element. Analogous to Controller::set().
  *
  * @param string|array $one A string or an array of data.
- * @param string|array $two Value in case $one is a string (which then works as the key).
+ * @param bool|int|string|array $two Value in case $one is a string (which then works as the key).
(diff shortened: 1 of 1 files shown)


lib/Cake/View/View.php
```diff
@@ -807,7 +807,7 @@ public function uuid($object, $url) {
  * a layout or other element. Analogous to Controller::set().
  *
  * @param string|array $one A string or an array of data.
- * @param string|array $two Value in case $one is a string (which then works as the key).
+ * @param bool|int|string|array $two Value in case $one is a string (which then works as the key).
```
garas:
```suggestion
  * @param mixed $two Value in case $one is a string (which then works as the key).
```

Value can be anything. It is also `mixed` in `3.x`

josephzidell: Good idea

garas: GitHub "Suggested change" somehow added 1 space too much before `*`

josephzidell: Fixed

PR #12840, opened 2018-12-24
4.x psalm

ADmad:


ADmad:
I removed the deprecated `Cache::engine()` and moved it's code directly under `Cache::pool()`, hope that's fine.

src/Cache/Cache.php
```diff
@@ -191,17 +191,12 @@ protected static function _buildEngine(string $name): void
...
      *
-     * @param string $config The configuration name you want an engine for.
-     * @return \Cake\Cache\CacheEngine When caching is disabled a null engine will be returned.
-     * @deprecated 3.7.0 Use Cache::pool() instead. In 4.0 all cache engines will implement the
-     *   PSR16 interface and this method does not return objects implementing that interface.
+     * @param string $config The name of the configured cache backend.
+     * @return \Cake\Cache\CacheEngine
      */
-    public static function engine(string $config): CacheEngine
+    public static function pool(string $config)
```
markstory: It might be nice to keep `engine()` as a deprecated alias for `pool` just in case folks are using it in their applications. It doesn't cost us anything to keep around a while longer.

markstory (approved):

src/Cache/Cache.php
ADmad: Done

PR #12841, opened 2018-12-25
4.x - PSR-18

ADmad:
I'll add more tests and exception handling as per PSR-18 specs if the updates till now seem fine.


src/Http/Client.php
```diff
@@ -396,17 +399,29 @@ protected function _mergeOptions(array $options): array
...
+    /**
+     * Sends a PSR-7 request and returns a PSR-7 response.
+     *
+     * @param \Psr\Http\Message\RequestInterface $request Request instance.
+     * @return \Psr\Http\Message\ResponseInterface Response instance.
+     * @throws \Psr\Http\Client\ClientExceptionInterface If an error happens while processing the request.
+     */
+    public function sendRequest(RequestInterface $request): ResponseInterface
+    {
+        return $response = $this->send($request, $this->_config);
```
saeideng: redundant?? `$response =`

ADmad: Yup, I'll fix it.

markstory (approved):
Looks good to me. Nice that this was so simple to do :clap:

ADmad:
The static analysis job failure is unrelated.

src/Http/Client/Adapter/Stream.php
```diff
@@ -268,7 +269,7 @@ protected function _send(RequestInterface $request): array
         fclose($this->_stream);
 
         if ($timedOut) {
-            throw new HttpException('Connection timed out ' . $url, 504);
+            throw new NetworkException('Connection timed out ' . $url, $request);
```
markstory: This change will break the behavior where an http request to a backend timing out will result in the app also responding with a 504.

ADmad: I can make `NetworkException` extend `HttpException` and set `504` status code but personally I don't think a timeout on an internal request should bubble up as `504` for the parent request. It should still be a 500 IMO.

ADmad: I checked a couple of guzzle wrappers for PSR18 support and couple of standalone PSR18 implementations and non set a `504` code.

markstory: Ok. Then perhaps our code should change too.

ADmad:
This is ready for final review. I have also added `ClientException` which is thrown when the adapter can't even initiate a request if required PHP config/module is disabled.

markstory:
Looks good other than builds failing.

PR #12842, opened 2018-12-25
Fix error code check.

ADmad:


src/Http/Client/Adapter/Curl.php
```diff
@@ -44,7 +44,7 @@ public function send(Request $request, array $options)
             curl_close($ch);
 
             $status = 500;
-            if ($error === 28) {
+            if ($errorCode === 28) {
                 $status = 504;
```
garas: Too much magic numbers here:
- `CURLE_OPERATION_TIMEDOUT (28)`
- `500 Internal Server Error`
- `504 Gateway Timeout`

ADmad: I replaced `28` with relevant constant. `Client\Response` doesn't have any constants for `500` and `504`.

dereuromark (approved):

src/Http/Client/Adapter/Curl.php
garas: Simple comment would help.

To understand these lines, both 28 and 504 need constants or explanations.

500 is more common code, description is optional/unnecessary, especially in context of Curl/HTTP Client.

ADmad: Anyone who you care to look into this code would know where to get details about `504` HTTP status code too :slightly_smiling_face:.

PR #12845, opened 2018-12-26
Fix handling of `html` and `htm` extensions.

ndm2:
As mentioned in https://github.com/cakephp/cakephp/issues/12585#issuecomment-437549557, the changes in #12586 broke how the special `html` and `htm` extensions are being treated. Before the change they would be "ignored", so that the rendering using the default application view would take place, and now an exception is being thrown.

This PR restores the old behavior for these extensions.


dereuromark:
Thank you for looking into this!
Clearly, there was a test case for this missing.
My bad, I never used verbose htm routing here.

PR #12847, opened 2018-12-27
Cleaner API for chaining

dereuromark:
Refs https://github.com/cakephp/cakephp/pull/12846

We need a cleaner API in some cases.
Here the IDE correctly warns about a dirtyness in code:

nullpointer

This can easily be solved by forcing the type to be of that object or otherwise throw an exception (we could add a more meaningful message here of course).
To check on it one could use `has...()` methods then?

We could also keep the current ones and add `...OrFail()` ones?

In this case you would otherwise be forced to do this:
```php
$possibleProvider = $this->Auth->authenticationProvider();
if ($possibleProvider && $possibleProvider->needsPasswordRehash()) {
    ...
}
```

Just wanted to get some general API feedback here on how to improve things on core level here to use less return types and especially less nullable where possible/feasible.

### DIFF
--- a/src/Controller/Component/AuthComponent.php
+++ b/src/Controller/Component/AuthComponent.php
@@ -946,7 +946,7 @@ class AuthComponent extends Component implements EventDispatcherInterface
 
         $this->Flash->set($message, $this->_config['flash']);
     }
-    
+
     /**
      * Checks if there is an BaseAuthenticate instance.
      *
@@ -956,7 +956,7 @@ class AuthComponent extends Component implements EventDispatcherInterface
     {
         return $this->_authenticationProvider !== null;
     }
-    
+
     /**
      * If login was called during this request and the user was successfully
      * authenticated, this function will return the instance of the authentication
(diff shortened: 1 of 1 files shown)


src/Controller/Component/AuthComponent.php
```diff
@@ -947,26 +947,46 @@ public function flash($message): void
         $this->Flash->set($message, $this->_config['flash']);
     }
 
+    /**
+     * Checks if there is an BaseAuthenticate instance.
+     *
+     * @return bool
+     */
+    public function hasAuthenticationProvider(): bool
```
markstory: I think this is going to be a smoother change than throwing exceptions from `authenticationProvider()`.

dereuromark: We need a non nullable return somewhere.
So you want to rather add OrFail() methods in these cases?

ADmad: This method seems redundant to me. One could just use `if ($this->Auth->user()) {` instead of `if ($this->Auth->hasAuthenticationProvider()) {`.

ADmad:
I would rather we don't make code changes to `AuthComponent` (which is supposed to be phased out) just to please IDEs.

PR #12848, opened 2018-12-27
Try not using .bat file to run phpunit

markstory:
See if this fixes build failures we've been seeing as of late.


ndm2:
@markstory Did you maybe ment to do `php vendor\phpunit\phpunit\phpunit`? Now it's just outputing the contents of the `vendor\bin\phpunit` file ;)

I did some testing and I think I found the underlying problem... I'll push a PR shortly.

markstory:
@ndm2 I did mean to run phpunit and not output the file.

PR #12849, opened 2018-12-27
Allow assetUrl() to use CDNs easier.

dereuromark:
Refs issue in IRC/Slack
People are not easily able to set a CDN host inside templating for asset URLs.
With this you about having to switch out the `Router::fullBaseUrl($new)` and reset it again afterwards.

### DIFF
--- a/src/View/Helper/UrlHelper.php
+++ b/src/View/Helper/UrlHelper.php
@@ -143,9 +143,9 @@ class UrlHelper extends Helper
      *
      * Depending on options passed provides full URL with domain name. Also calls
      * `Helper::assetTimestamp()` to add timestamp to local files.
-     * 
+     *
      * ### Options:
-     * 
+     *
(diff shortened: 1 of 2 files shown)


src/View/Helper/UrlHelper.php
```diff
@@ -146,7 +146,7 @@ public function script($path, array $options = [])
      *
      * @param string|array $path Path string or URL array
      * @param array $options Options array. Possible keys:
-     *   `fullBase` Return full URL with domain name
+     *   `fullBase` Return full URL with domain name. Bool or string.
```
ADmad: A slightly more detailed description and/or example would be nice so people know how exactly this option can be used to generate CDN URLs.

dereuromark: Fair enough
as soon as birdy247 provides a test case I am happy to detail this in text :)

PR #12851, opened 2018-12-27
 Fix `FileCacheEngine` removing files in parent of configured path.

ndm2:
When processing dot files, `..` will expand to the parent of the configured base path, causing unrelated files to be deleted.

This seems to be what caused AppVeyor to fail, it was triggered in `SimpleCacheEngineTest` when it tried to clear the cache. With a prefix of `''`, the inner `FileCacheEngine` deleted all files in the configured path's parent path, which happens to be the OS temp folder, where probably some AppVeyor internal `.bat` file is located, resulting in an error when the batch script terminates and the file doesn't exist anymore.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -98,4 +98,4 @@ before_test:
 test_script:
   - sqlcmd -S ".\SQL2012SP1" -U sa -P Password12! -Q "create database cakephp;"
   - cd C:\projects\cakephp
-  - php vendor\bin\phpunit
+  - vendor\bin\phpunit.bat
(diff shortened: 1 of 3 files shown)


markstory (commented):
Awesome stuff. How did you figure out that this was the problem?

ndm2:
@markstory With [**lots of trial & error**](https://github.com/ndm2/cakephp/commits/appveyor) :)

After ruling out the `phpunit.bat` file, timing issues, just PHP running, and figuring that it's happening after PHPUnit exited, but before the next test script ran, I suspected that it's probably some AppVeyor internal file being affected. By excluding tests I could pin it to the cache test, which made me suspect that it's probably deleting something from a shared location.

I wish I tried it on a local Windows installation first, that would have saved me quite some time, because there it crashed/froze PHP when running the cache test, which made it relatively easy to pin it down by inspecting a few breakpoints :upside_down_face:

Issue #12853, opened 2018-12-29
Problem when testing subfolder setup


Zuluru:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.6.12
* Platform and Target: phpunit

### What you did
I have been having some issues with running my app from a subdirectory in the web server configuration, and am trying to adjust my unit tests to help nail this down. To this end, I set `$_SERVER['PHP_SELF']` to /folder/index.php in my `tests/bootstrap.php`.

### What happened
`TestSuite\MiddlewareDispatcher.php` is ignoring my setting, replacing it with just /, and generated URLs are wrong as a result.

### What you expected to happen
It should honour my setting, using it to determine the baseUrl, and hence to generate various URLs in the system.

`MiddlewareDispatcher::_createRequest` is generating `$environment` with `array_merge`, but it seems the order of arguments is wrong, so that the "default" PHP_SELF of / is overwriting my setting.

saeideng:
it was added here https://github.com/cakephp/cakephp/pull/12076 for solve the PHP_SELF issue 

I think you  able to  set headers with  
```php
->configRequest(['headers'=>['PHP_SELF'=>'addYourCode']])
```

Zuluru:
The commit note there indicates that it was added to help people who don't have it set in their tests/bootstrap.php. It shouldn't cause harm to those who do. Simply reversing the order of arguments in the `array_merge` call would allow whatever I set to propagate instead of being overwritten.

dereuromark:
Sounds fair. You can open a PR and we can then discuss that concrete result.

saeideng:
>Simply reversing the order of arguments in the array_merge

I dont think this could help to someone that dont have/set $_SERVER['PHP_SELF'] in their tests/bootstrap.php
but you can open a PR for more discuss

Zuluru:
By reversing the order, it will default to the value provided here, but allow anything the user has set to override that, instead of the other way around. The code change for the PR is easy, but I always have problems running the test suite before submitting it. Working on it...

dereuromark:
Let Travis handle the results :)

Zuluru:
Multiple failures, but it's not clear that any of them are due to my change. Is it supposed to run clean? There's stuff about missing controllers and tables already existing that I just can't see being a result of this.

markstory:
How are you setting php_self in your test suite requests?

[...]

PR #12854, opened 2018-12-29
Changes to allow user-specified PHP_SELF to take precedence

Zuluru:
Addresses #12853.

### DIFF
--- a/src/TestSuite/MiddlewareDispatcher.php
+++ b/src/TestSuite/MiddlewareDispatcher.php
@@ -164,9 +164,12 @@ class MiddlewareDispatcher
             $spec['post'] = [];
         }
         $environment = array_merge(
-            array_merge(['REQUEST_URI' => $spec['url'], 'PHP_SELF' => '/'], $_SERVER),
+            array_merge($_SERVER, ['REQUEST_URI' => $spec['url']]),
             $spec['environment']
         );
+        if (strpos($environment['PHP_SELF'], 'phpunit') !== false) {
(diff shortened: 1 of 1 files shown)


src/TestSuite/MiddlewareDispatcher.php
```diff
@@ -164,7 +164,7 @@ protected function _createRequest($spec)
             $spec['post'] = [];
         }
         $environment = array_merge(
-            array_merge($_SERVER, ['REQUEST_URI' => $spec['url'], 'PHP_SELF' => '/']),
+            array_merge(['REQUEST_URI' => $spec['url'], 'PHP_SELF' => '/'], $_SERVER),
```
markstory: Won't this result in REQUEST_URI being wrong? That could explain some of the test failures.

Zuluru: Only if REQUEST_URI is also being specified in the bootstrap or elsewhere, which seemed unlikely. (And if it is, maybe it should take precedence? I don't know.) Tried to run the tests first on my server, but got many failures, likely due to unsupported versions of things like SQLite. Will come back to this a little later, when I have more time to really dig into it.

Zuluru: Latest commit resolves this. Any problems with it, or just haven't had time to review?

markstory: I had missed the update notification on this.

PR #12866, opened 2019-01-04
4.x Add an interface for SchemaCollections

markstory:
When doing a change in bake I wasn't pleased that I had to extend the base class in order to satisfy typehinting. This switches typehints to be interface based. This should be a mostly backwards compatible change as the base class implements the new interface.


src/Database/Schema/CollectionInterface.php
```diff
@@ -0,0 +1,50 @@
...
+ *
+ * Used to access information about the tables,
+ * and other data in a database.
+ */
+interface CollectionInterface
+{
+    /**
+     * Get the list of tables available in the current connection.
+     *
+     * @return array The list of tables in the connected database/schema.
```
dereuromark: Can we use string[] here? Or is that not true?

Issue #12867, opened 2019-01-04
Incorrect phpdoc param for IntegrationTestCase::post


softius:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.6.14
* PHP 7.2.10-0, built-in web server
* MySQL 5.7.24

### What you did
A test case was created invoking `IntegrationTestCase::post` with JSON data. 

```php 
$data = (string)json_encode(['foo' => 'bar']);
$this->post('/some/endpoint', $data)
```

### What happened
Got an error from PHPStan (max level), reporting the following:

```
Parameter #2 $data of method Cake\TestSuite\IntegrationTestCase::post() expects array, string given.
``` 

### What you expected to happen
Expected no PHPStan error related to this, since the `IntegrationTestCase::post` can handle strings as well. In particular in `IntegrationTestCase::_buildRequest` is checked whether the provided `$data` is a string, despite the fact that `$data` is type-hinted to `array|null`.

PR #12870, opened 2019-01-04
Add string|array|null typehint for data

softius:
This PR addresses #12867 by improving the phpdoc typehint for `$data` param that it's being used in the following methods:

- `IntegrationTestTrait::post`
- `IntegrationTestTrait::patch`
- `IntegrationTestTrait::put`
- `IntegrationTestTrait::_sendRequest`
- `IntegrationTestTrait::_buildRequest`

With the above additions, it is defined that these methods can also accept a string (or null) as `$data` which is already supported by `IntegrationTestTrait::_buildRequest`

### DIFF
--- a/src/TestSuite/IntegrationTestTrait.php
+++ b/src/TestSuite/IntegrationTestTrait.php
@@ -404,7 +404,7 @@ trait IntegrationTestTrait
      * response.
      *
      * @param string|array $url The URL to request.
-     * @param array $data The data for the request.
+     * @param string|array|null $data The data for the request.
      * @return void
      * @throws \PHPUnit\Exception
      */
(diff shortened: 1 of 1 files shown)


dereuromark:
For some use cases it can also be other scalar values.

mamchenkov:
> For some use cases it can also be other scalar values, right?

Maybe changing to a more generic `mixed` is a better option then?

softius:
Typehint can be changed to `bool|int|float|string|array|null` and then make the necessary adjustments so that all scalar types are being converted to string (similar to `_castToString`).

markstory:
I don't think there is a practical use case for float/int. Adding string should be good enough for now.

PR #12871, opened 2019-01-04
Remove bad chaining typehints and fix up RepositoryInterface methods.

dereuromark:
As discussed, this brings extendability and chaining back for $this methods.
Also fixed up RepositoryInterface methods.

A sniffer ( https://github.com/spryker/code-sniffer/pull/149 ) then can assert this is done properly and consistently, and also helps to avoid bugs here. 

Also: Chainable methods (return $this) must never have a 2nd return type, we still violate this in a few places afaik (2x):
- TranslateTrait::translation()
- Mailer::__call()


markstory (commented):
Looks good to me.

PR #12873, opened 2019-01-05
Add missing param for parent call.

dereuromark:
right?

### DIFF
--- a/src/ORM/Query.php
+++ b/src/ORM/Query.php
@@ -1229,7 +1229,7 @@ class Query extends DatabaseQuery implements JsonSerializable, QueryInterface
         $repository = $this->getRepository();
         $this->from([$repository->getAlias() => $repository->getTable()]);
 
-        return parent::delete();
+        return parent::delete($table);
     }
 
     /**


src/ORM/Query.php
```diff
@@ -1229,7 +1229,7 @@ public function delete(?string $table = null)
         $repository = $this->getRepository();
         $this->from([$repository->getAlias() => $repository->getTable()]);
 
-        return parent::delete();
+        return parent::delete($table);
```
markstory: This will overwrite the table + alias set on line 1230 which I don't think is ideal.

saeideng:
you can remove `?string $table = null` parameter
it never used in 3.x series https://github.com/cakephp/cakephp/commit/0d6f20251fb99fd6efa6d624cb24a73e2a8faca8

src/ORM/Query.php
dereuromark: Then we should add a comment why the parent is called without the argument, as this is not usual :)

markstory:
@saeideng But someone else could be using it.

dereuromark:
Closing in favor of https://github.com/cakephp/cakephp/pull/12877 then

dereuromark:
@markstory We cannot clean this up for 4.x?

PR #12874, opened 2019-01-05
4.x client request

ADmad:
I removed the combined getter/setter `Cake\Http\Client\Messsage::body()`.

### DIFF
--- a/src/Http/Client/Message.php
+++ b/src/Http/Client/Message.php
@@ -151,13 +151,6 @@ class Message
     protected $_cookies = [];
 
     /**
-     * Body for the message.
-     *
-     * @var string|null
-     */
-    protected $_body;
-
-    /**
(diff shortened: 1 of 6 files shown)


src/Http/Client/Message.php
```diff
@@ -166,20 +159,4 @@ public function cookies(): array
...
         return $this->_cookies;
     }
-
-    /**
-     * Get/set the body for the message.
-     *
-     * @param string|null $body The body for the request. Leave null for get
-     * @return mixed Either $this or the body value.
-     */
-    public function body($body = null)
```
markstory: This wasn't deprecated before. We should document this removal in the 4.0 migration guide.

ADmad: Done

PR #12876, opened 2019-01-05
Use validator for prefer-lowest.

dereuromark:
Using the validator we get info for free about where we cannot promise/test the minimum versions declared.

Fow now, this only issues a warning output (build still passes) as require-dev raises the chronos dependency ( https://travis-ci.org/cakephp/cakephp/jobs/475716058#L766 ):
```
1 warning (impossible to test minimum version here):
 - cakephp/chronos: Defined `1.0.1.0` as minimum, but is `1.2.1.0`
```

Hopefully, we can clean this up in the future (just requiring `^1.2.1` right away).
Until then this helps as a reminder.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -85,9 +85,11 @@ before_install:
 before_script:
   - if [[ $PREFER_LOWEST != 1 ]]; then composer install --prefer-source --no-interaction; fi
   - if [[ $PREFER_LOWEST == 1 ]]; then composer update --prefer-lowest --prefer-stable --prefer-dist --no-interaction; fi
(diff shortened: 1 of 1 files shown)


ADmad:
Why don't really understand what your tool does.

dereuromark:
Read https://www.dereuromark.de/2019/01/04/test-composer-dependencies-with-prefer-lowest/
It reports if we are testing against impossible minimums, prefer-lowest only pulls what it can, so it can silently ignore the actual minimum we promise to regression-test. As such we could use newer functionality from one of the dependencies that would break hard on a project because they have a lower constraint for that dependency locked in.

ADmad:
You website's SSL certificate seems to have expired.

dereuromark:
Oh no, must have been yesterday

PR #12879, opened 2019-01-06
Remove deprecated code.

ADmad:


ADmad:
Should `Validator::notEmpty()` and `Validator::allowEmpty()` also be removed?

markstory (approved):
I think we can remove anything with a warning. If there is now warning in 3.7 we should leave the method behind.

ADmad:
Validator::notEmpty() and Validator::allowEmpty() don't throw deprecation warning. Don't remember if that's intentional or oversight.

PR #12888, opened 2019-01-10
Fix arguments type and add typehint.

ADmad:
This also fixes [this](https://travis-ci.org/cakephp/cakephp/jobs/476220922#L713) reported by phpstan at level 3.

### DIFF
--- a/src/Event/EventManager.php
+++ b/src/Event/EventManager.php
@@ -103,14 +103,14 @@ class EventManager implements EventManagerInterface
             return $this;
         }
         $argCount = func_num_args();
-        if ($argCount === 2) {
+        if ($eventKey && $argCount === 2) {
             $this->_listeners[$eventKey][static::$defaultPriority][] = [
                 'callable' => $options,
             ];
 
             return $this;
         }
-        if ($argCount === 3) {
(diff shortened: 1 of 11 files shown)


ADmad:
> Do we need to backport to 3.x?

Sure, why not.

New release of psalm is spewing new errors :unamused:.

dereuromark:
yeah looks like a bad release. better to hardcode to previous one.

ADmad:
Nope, they seem like genuine errors which weren't reported before.

src/Event/EventManager.php
```diff
@@ -103,14 +103,14 @@ public function on($eventKey = null, $options = [], $callable = null)
             return $this;
         }
         $argCount = func_num_args();
-        if ($argCount === 2) {
+        if ($eventKey && $argCount === 2) {
```
saeideng: `public function on($eventKey = null, `
if `$eventKey` can not be `null` in within this method, and it throw `InvalidArgumentException`
why it is `null`  by default ?

ADmad: Good question, I don't see why `$eventKey` is nullable in the interface. I'll remove it.

PR #12889, opened 2019-01-10
Fix invalid method calls.

ADmad:
2nd and 3rd arguments of stream_select() must be variables.


ADmad:
`Console::dataAvailable()` seems to be unused in core nor does it have code coverage. Should it be dropped in 4.0?

markstory:
This method was added in https://github.com/cakephp/cakephp/pull/1017 writing tests for this kind of wrapper is pretty hard which is why the changes were merged without tests.

Issue #12885, opened 2019-01-10
HttpSocket request config without query params


kicaj:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 2.10.14
* Platform and Target: WAMP, PHP 7.2.7

### What you did
I use one configuration to stop repeatable arrays for API connection

### What happened
Prepare my resuable configuration for requests:
```
$hc = new HttpSocket(array(
	'host' => 'api.address.pl',
	'request' => array(
		'uri' => array(
			'scheme' => 'https',
			'host' => 'api.address.pl',
			'port' => 443,
			'query' => array(
				'token' => 'abc123',
			),
		),
	),
));
```

Then I can using code like below:
```
$user = $hc->get(array(
	'path' => '/api/user',
), array(
	'id' => 100,
)); // Get one user by query user id
```

Works perfeclty and generate URL like this:
`https://api.address.pl/api/user?token=abc123&id=100`

But...
When I use something like:
```
$users = $hc->get(array(
	'path' => '/api/users',
)); // Get all users without query data
```
or
```
$users = $hc->get(array(
	'path' => '/api/users',
), array()); // Get all users with empty query data
```

Generate URL like this:
`http://api.address.pl/api/user?token=abc123`
HttpSocket connect to `api.address.pl` by `http` and `80` port.

Should still using global variables from HttpSocket and prepare URL like this:
`https://api.address.pl/api/user?token=abc123` (with `https`)

kicaj:
Solution is to remove `if` statement from line 455: https://api.cakephp.org/2.10/source-class-HttpSocket.html#455

markstory:
Would you be willing to make a pull request with the suggested changes?

kicaj:
Yes I will. You think I should also remove the same `if` statement at line 481 for `head` method?

markstory:
If the tests pass go ahead :smile:

markstory:
Closing as a pull request is open.

PR #12893, opened 2019-01-11
Remove statement from get and head method

kicaj:
Remove `if` statement: #12885


markstory (commented):
Looks good but the tests need to be updated as well.

PR #12897, opened 2019-01-12
Remove deprecated tag.

ADmad:
Only using method with argument is deprecated.


dereuromark (approved):
We could make a soft inline comment

Issue #12676, opened 2018-10-24
'errorField' should accept array of properties when adding new rules


efanzini:
This is a (multiple allowed):

* [ ] bug
* [x] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.6.12
* Platform and Target: not relevant

### What you did
I added an isUnique rule to my table, referencing two properties ("code" and "revision").
I also set the 'errorField' option to the same properties.

```php
$properties = ['code', 'revision'];
$rules->add($properties, 'code_revision_UNIQUE', [
    'errorField' => $properties,
    'message' => __('Code and revision combination already in use')
]);
```
### What happened
```php
Warning (2): Illegal offset type [CORE\src\Datasource\EntityTrait.php, line 993]
Notice (8): Array to string conversion [CORE\src\Datasource\RuleInvoker.php, line 135]
```
And no errors showing on my form controls

### What you expected to happen
I expected both of my fields to be invalidated and my form controls to show the error

ADmad:
As a user I would find seeing the same error message for 2 fields pretty weird.

efanzini:
@ADmad It doesn't sound weird to me, as long as the error message is pertinent.

ravage84:
@efanzini I used to create a CakePHP plugin for this for my CakePHP 2.x applications.
I never upgraded it, though.

https://github.com/ravage84/cakephp-multi-column-uniqueness

dereuromark:
Or use the Tools plugin 2.x branch (0.x versions).

dereuromark:
It is fair to say that this can be done on user land code and exists as multiple plugin code, so since 2.x is in maintenance mode, we an close this then now.

ravage84:
@dereuromark I agree that this can be done in user land code, but I'd keep it for 3.x, if it can be done.

dereuromark:
Let's see if someone wants to provide a PR here for 3.7 before it lands.

PR #12898, opened 2019-01-12
Allow field to be an array for validation and rules

kevinquinnyo:
Allow for multiple fields to be passed to application rules, and when creating validation rules.

Issue reference: #12676

I thought this was already possible.  When looking into it I noticed the open issue so I decided to give it a shot as it seemed easy enough.  A few concerns with this change:

- It seems like it was *too easy* which makes me think I am probably overlooking something (or a lot of things). I don't have a lot of knowledge in this area of the framework.
- This does add a performance penalty due to casting the field as an array and iterating (even if only one field is passed, iteration now occurs). Not sure if the performance penalty is negligable or not.
- Given how easy it is to achieve this in userland code, I'm honestly not sure it's a worthy addition, so take it or leave it.  For example:

```php
$fields = ['title', 'description'];
foreach ($fields as $field) {
    $rules->add(//...
}
```

### DIFF
--- a/src/Validation/Validator.php
+++ b/src/Validation/Validator.php
@@ -426,6 +426,9 @@ class Validator implements ArrayAccess, IteratorAggregate, Countable
      * then rules list for the field will be replaced with second argument and
      * third argument will be ignored.
      *
+     * The first argument can be an array of fields if you wish for the rule to
+     * apply to multiple fields.
+     *
      * ### Example:
      *
      * ```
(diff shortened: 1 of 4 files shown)


kevinquinnyo:
>  The main issue is the API here, which now cannot add a typehint here on the arguments in 4.x.

Ah good point. The string typehint in 4.x adds more value to the framework overall than this PR would, IMO. So if it's a pain to make this work, I'd trash the PR -- just my opinion.

src/Validation/Validator.php
```diff
@@ -426,6 +426,9 @@ public function count()
      * then rules list for the field will be replaced with second argument and
      * third argument will be ignored.
      *
+     * The first argument can be an array of fields if you wish for the rule to
+     * apply to multiple fields.
```
markstory: The rule would be applied to each field individually. This seems to imply that both fields would be arguments to one validation rule call.

markstory (commented):
I too would prefer an addMany or addMultiple method instead of overloading existing methods.

saeideng:
>I too would prefer an addMany or addMultiple method instead of overloading existing methods.

how we can use `isUnique` rule now
it just applies to `name` now 
```php
$rules->add($rules->isUnique(['name','name2'],'name + name2 already exist'));
```
passing multi args into `isUnique` is one of feature of this rule 

`foreach` does not work in this sample ( we can not use `addMany()` )

dereuromark:
@saeideng The current implementation basically does a addMany(), it doesn't too a combined isUnique() as you suggest. At least from what I saw in code. That would be a different, and real, use case for an array as input, I agree.

ADmad:
Closing as the consensus it to add separate method instead of changing existing ones to allow array. You are to open a new PR accordingly.

PR #12901, opened 2019-01-15
Fix docblocks.

ADmad:


dereuromark (dismissed):

dereuromark:
Just wondering: What is the use case for false here?
I think the (always) just referred to the normal case of true being invoked always. To clarify.

ADmad:
Deprecated `notEmpty('field)` is now to be replaced by `allowEmptyString('field', false)` or similar.

Issue #12903, opened 2019-01-16
Cake\I18n\Time toQuarter() returns Float instead of Int


jonmc1:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.7.2
* Platform and Target: apache, php 7.1

### What you did
use Cake\I18n\Time;

$date = new Time();
$quarter = $date->toQuarter();

### What happened
$quarter is a float 

### What you expected to happen
$quarter should be an int. 

The documentation says that toQuarter() returns an int but it returns a float. The value is rounded up to a whole number, but it's type is still float. This can cause problems if you're using strict comparison:
```
Time::parse('2019-01-01 00:00:00')->toQuarter() === 1 // false
Time::parse('2019-01-01 00:00:00')->toQuarter() === 1.0 // true
```
Or you could fail validation like this:
```
$validator
    ->integer('quarter')
    ->allowEmpty('quarter');
```

dereuromark:
We should cast it then before returning.

lorenzo:
Yeah, we should do that

PR #12904, opened 2019-01-16
Cake\I18n\Time toQuarter() return Int

jonmc1:
GitHub Issue #12903

Cast return value of toQuarter() to an int

### DIFF
--- a/src/I18n/Time.php
+++ b/src/I18n/Time.php
@@ -174,7 +174,7 @@ class Time extends MutableDateTime implements JsonSerializable
     {
         $quarter = ceil($this->format('m') / 3);
         if ($range === false) {
-            return $quarter;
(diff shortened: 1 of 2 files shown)


src/I18n/Time.php
```diff
@@ -174,7 +174,7 @@ public function toQuarter($range = false)
     {
         $quarter = ceil($this->format('m') / 3);
```
dereuromark: technically, the cast here is better

dereuromark (approved):

dereuromark:
https://github.com/cakephp/cakephp/blob/master/tests/TestCase/View/Helper/TimeHelperTest.php#L156
These kind of comparisons are the problem of all eval :)

lorenzo (approved):

Issue #12678, opened 2018-10-25
CakePHP Console: Unable to configure the session on PHP 7.2.0


trenshaw:
* [x] bug
* CakePHP Version: 2.10.12
* Platform and Target: Ubuntu 18.04, MySQL 5.7.24, PHP 7.2.10

### Description
A behavior attached to a model which calls AuthComponent::user() in attempts to obtain the logged in user fails when used in a console shell/task.

Have re-tested on and confirmed bug / issue does not manifest in shells / tasks on PHP 5.3.3. Issue appears to be related to the following (fixed) PHP bugs:-

Bug #71038 | session_start() returns TRUE on failure
https://bugs.php.net/bug.php?id=71038

Raised in PHP 7.1.0 and fixed in 7.2.0. CakeSession (in console / shells / tasks) appears to have relied on session_start() falsely returning TRUE and/or $_SESSION still being usable after headers have been sent. Refer:-

Bug #75227 | You cannot change the session module's ini settings at this time
https://bugs.php.net/bug.php?id=75227

Bug #73245 | session_start() doesn't start the session if headers have been sent
https://bugs.php.net/bug.php?id=73245

### Result
Error: Unable to configure the session, setting session.use_trans_sid failed.
#0 /var/www/html/cakephp-2.10.12/lib/Cake/Model/Datasource/CakeSession.php(619): CakeSession::_configureSession()
#1 /var/www/html/cakephp-2.10.12/lib/Cake/Model/Datasource/CakeSession.php(632): CakeSession::_cookieName()
#2 /var/www/html/cakephp-2.10.12/lib/Cake/Model/Datasource/CakeSession.php(248): CakeSession::_hasSession()
#3 /var/www/html/cakephp-2.10.12/lib/Cake/Controller/Component/AuthComponent.php(670): CakeSession::check('Auth.User')

### Expected
AuthComponent::user() should return NULL in console shells/task. CakeSession should not rely on session_start() always returning TRUE the same for any INI settings related to a session that failed to start.

CakeSession should gracefully handle the absence of session in the CLI. As a workaround, I added a !static::$_isCLI check to if(empty($_SESSION) statement on line 578 in Cake/Model/Datasource/CakeSession.php to ensure ini_set() is not called for session.* related settings in the CLI.

markstory:
> CakeSession should not rely on session_start() always returning TRUE the same for any INI settings related to a session that failed to start.

How would you expect the session to behave properly if there is no session?

[...]

PR #12906, opened 2019-01-16
resolving ini_set error in php 7.2

akeiper:
Fixing the issue where ini_set errors if headers were already sent. https://github.com/cakephp/cakephp/issues/12678

Based the solution on CakePHP v3 session controller:
https://github.com/cakephp/cakephp/blob/master/src/Http/Session.php#L316

### DIFF
--- a/lib/Cake/Model/Datasource/CakeSession.php
+++ b/lib/Cake/Model/Datasource/CakeSession.php
@@ -575,7 +575,7 @@ class CakeSession {
 			$sessionConfig['cacheLimiter'] = 'must-revalidate';
 		}
 
-		if (empty($_SESSION)) {
+		if (!headers_sent() && session_status() !== \PHP_SESSION_ACTIVE  && empty($_SESSION)) {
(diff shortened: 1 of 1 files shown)


lib/Cake/Model/Datasource/CakeSession.php
```diff
@@ -575,7 +575,7 @@ protected static function _configureSession() {
 			$sessionConfig['cacheLimiter'] = 'must-revalidate';
 		}
 
-		if (empty($_SESSION)) {
+		if (!headers_sent() && session_status() !== \PHP_SESSION_ACTIVE  && empty($_SESSION)) {
```
garas: `PHP Fatal error:  Call to undefined function session_status()` on PHP: 5.3. Add check with `function_exists()`.

CakePHP 2 is namespaceless, so `\` is inconsistent here.

`578 | ERROR | Double space found`

akeiper:
@garas updated to include `function_exists()`

PR #12913, opened 2019-01-18
Add unit test for timezone conversion

bancer:
According to https://www.timeanddate.com/worldclock/converter.html?iso=20190131T030000&p1=28&p2=69&p3=136 '2019-01-31 10:00:00' in Bangkok is '2019-01-31 03:00:00' in UTC. Does this failing test mean that there is a bug in CakeTime class?

### DIFF
--- a/lib/Cake/Utility/CakeTime.php
+++ b/lib/Cake/Utility/CakeTime.php
@@ -369,7 +369,7 @@ class CakeTime {
 			$format = static::$niceFormat;
 		}
 		$convertedFormat = static::convertSpecifiers($format, $timestamp);
-		return static::__strftime($convertedFormat, $timestamp, $date, $timezone);
+		return static::__strftimeWithTimezone($convertedFormat, $timestamp, $date, $timezone);
 	}
 
 /**
@@ -394,19 +394,19 @@ class CakeTime {
 		$timestamp = static::fromString($date, $timezone);
 
 		if (static::isToday($date, $timezone)) {
-			$formattedDate = static::__strftime("%H:%M", $timestamp, $date, $timezone);
+			$formattedDate = static::__strftimeWithTimezone("%H:%M", $timestamp, $date, $timezone);
 			return __d('cake', 'Today, %s', $formattedDate);
 		}
 		if (static::wasYesterday($date, $timezone)) {
-			$formattedDate = static::__strftime("%H:%M", $timestamp, $date, $timezone);
+			$formattedDate = static::__strftimeWithTimezone("%H:%M", $timestamp, $date, $timezone);
 			return __d('cake', 'Yesterday, %s', $formattedDate);
 		}
 		if (static::isTomorrow($date, $timezone)) {
-			$formattedDate = static::__strftime("%H:%M", $timestamp, $date, $timezone);
+			$formattedDate = static::__strftimeWithTimezone("%H:%M", $timestamp, $date, $timezone);
 			return __d('cake', 'Tomorrow, %s', $formattedDate);
 		}
 
-		$d = static::__strftime("%w", $timestamp, $date, $timezone);
+		$d = static::__strftimeWithTimezone("%w", $timestamp, $date, $timezone);
 		$day = array(
 			__d('cake', 'Sunday'),
 			__d('cake', 'Monday'),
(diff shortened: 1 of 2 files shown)


markstory:
We don't do any special handling around timezone conversions. You may have to update your timezone DB or PHP.

bancer:
@markstory I do not understand what you mean by that. According to [documentation](https://api.cakephp.org/2.10/class-CakeTime.html#_format): `CakeTime::format('2012-02-15 23:01:01', '%c', 'N/A', 'America/New_York'); // converts passed date to timezone`. And there is a unit test `CakeTimeTest::testCorrectTimezoneConversion`.  I suppose that means that CakeTime class is supposed to perform timezone conversions. What do you mean by "special handling"?

markstory:
> What do you mean by "special handling"?
>
> CakeTime class does not use a database and CakeTimeTest::testTimezoneConversionToUTC test fails in all PHP versions on Travis. How updating the timezone DB or PHP would fix that?

I mean that we don't actually do any timezone offset calculations in CakePHP. We use PHP's timezone features. Have you tried the same timezone changes with native PHP objects? This might be a problem with PHP or the timezone database it uses.

bancer:
> Have you tried the same timezone changes with native PHP objects?

I added `testTimezoneConversionToUTCPlainPHP` test and it passes.

bancer:
> I mean that we don't actually do any timezone offset calculations in CakePHP.

Actually, it seems there is at least one place where the offset is added - https://github.com/cakephp/cakephp/blob/2.x/lib/Cake/Utility/CakeTime.php#L330

bancer:
@markstory I split `testFromStringWithDateTime` test into two. Is there any chance you would recall what scenario `testFromStringWithDateTimeAsia` method is supposed to test ([initial commit](https://github.com/cakephp/cakephp/commit/5fe22cfc8defe7204d2411a12b9771851127b436#diff-8321c4ab53a01784e44932fd0083b764R830))? There are 3 time zones in that test and is quite unclear what functionality is tested.

markstory:
I missed that offset munging earlier. That could be part of the cause for this problem.

>  committed on May 13, 2012

I don't remember much from 7 years ago sorry.

bancer:
@markstory I think I figured out what's the problem. 

`$format = strftime($format, $date);` ([source](https://github.com/cakephp/cakephp/blob/e2781a536adc9491e363e2c79d2da9a417778a0d/lib/Cake/Utility/CakeTime.php#L1046)) converts `$date` timestamp to the server's datetime. It seems that `_strftime` method needs `$timezone` parameter and to adjust the timezone by  `date_default_timezone_set($timezone)` before `strftime` is called.

bancer:
The changes are ready to be reviewed.

lib/Cake/Utility/CakeTime.php
```diff
@@ -1176,4 +1182,29 @@ protected static function _strftime($format, $date) {
...
+ *
+ * Adjusts the timezone when necessary before formatting the time.
+ *
+ * @param string $format Format string.
+ * @param int $timestamp Timestamp to format.
+ * @param int|string|DateTime $date Timestamp, strtotime() valid string or DateTime object.
+ * @param string|DateTimeZone $timezone Timezone string or DateTimeZone object.
+ * @return string Formatted date string with correct encoding.
+ */
+	private static function __strftime($format, $timestamp, $date, $timezone) {
```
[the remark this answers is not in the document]
bancer: Yes, I agree. Any suggestions?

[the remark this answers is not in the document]
markstory: strftimeWithTimezone?

lib/Cake/Utility/CakeTime.php
```diff
@@ -1176,4 +1182,29 @@ protected static function _strftime($format, $date) {
...
+		$serverTimeZone = date_default_timezone_get();
+		if (
+			!empty($timezone) &&
+			$date instanceof DateTime &&
+			$date->getTimezone()->getName() != $serverTimeZone
+		) {
+			date_default_timezone_set($timezone);
+		}
+		$result = static::_strftime($format, $timestamp);
+		date_default_timezone_set($serverTimeZone);
```
markstory: Do you need to always do this or only when the default was changed?

lib/Cake/Utility/CakeTime.php
```diff
@@ -1176,4 +1182,29 @@ protected static function _strftime($format, $date) {
...
+ *
+ * Adjusts the timezone when necessary before formatting the time.
+ *
+ * @param string $format Format string.
+ * @param int $timestamp Timestamp to format.
+ * @param int|string|DateTime $date Timestamp, strtotime() valid string or DateTime object.
+ * @param string|DateTimeZone $timezone Timezone string or DateTimeZone object.
+ * @return string Formatted date string with correct encoding.
+ */
+	private static function __strftimeWithTimezone($format, $timestamp, $date, $timezone) {
```
garas: `protected`?

markstory: :+1: to protected we generally don't use many private methods.

PR #12914, opened 2019-01-18
Patching search data in findOrCreate to ensure valid entity

jeremyharris:
`Table::findOrCreate` can create and persist invalid entities as it doesn't apply any validation when creating a new entity from the `$search`.

I'm not sure if this is a bug or a feature. I personally see it as a bug, but I can see how `findOrCreate` not returning a persisted entity could be seen as a bug too. It's also a behavior change, so it might be that this needs to go in 3.next.

### DIFF
--- a/src/ORM/Table.php
+++ b/src/ORM/Table.php
@@ -1730,7 +1730,7 @@ class Table implements RepositoryInterface, EventListenerInterface, EventDispatc
         }
         $entity = $this->newEntity();
         if ($options['defaults'] && is_array($search)) {
-            $entity->set($search, ['guard' => false]);
+            $entity = $this->patchEntity($entity, $search, ['guard' => false]);
         }
         if ($callback !== null) {
             $entity = $callback($entity) ?: $entity;
--- a/src/ORM/Table.php
+++ b/src/ORM/Table.php
@@ -1730,7 +1730,8 @@ class Table implements RepositoryInterface, EventListenerInterface, EventDispatc
         }
         $entity = $this->newEntity();
         if ($options['defaults'] && is_array($search)) {
-            $entity = $this->patchEntity($entity, $search, ['guard' => false]);
+            $accessibleFields = array_combine(array_keys($search), array_fill(0, count($search), true));
+            $entity = $this->patchEntity($entity, $search, ['accessibleFields' => $accessibleFields]);
         }
         if ($callback !== null) {
             $entity = $callback($entity) ?: $entity;
--- a/src/ORM/Table.php
+++ b/src/ORM/Table.php
@@ -1737,7 +1737,13 @@ class Table implements RepositoryInterface, EventListenerInterface, EventDispatc
         }
         unset($options['defaults']);
 
-        return $this->save($entity, $options) ?: $entity;
+        $result = $this->save($entity, $options);
+
+        if ($result === false) {
+            throw new PersistenceFailedException($entity, ['findOrCreate']);
+        }
+
+        return $entity;
     }
 
     /**
--- a/tests/TestCase/ORM/TableTest.php
+++ b/tests/TestCase/ORM/TableTest.php
@@ -53,6 +53,7 @@ class ProtectedEntity extends Entity
 {
     protected $_accessible = [
         'id' => true,
+        'title' => false,
     ];
 }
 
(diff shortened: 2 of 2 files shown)


ADmad:
I too see findOrCreate not returning a persisted entity as a bug.

tests/TestCase/ORM/TableTest.php
```diff
@@ -5839,6 +5839,24 @@ public function testFindOrCreateNoTransaction()
...
+     */
+    public function testFindOrCreateValidatesNewEntity()
+    {
+        $articles = $this->getTableLocator()->get('Articles');
+        $validator = new Validator();
+        $validator->notBlank('title');
+        $articles->setValidator('default', $validator);
+
+        $article = $articles->findOrCreate(['title' => '']);
+        $this->assertTrue($article->isNew());
```
saeideng: The output of   `findOrCreate` should be a valid entity on DB , or `false`/`exception`  otherwise imo
so this `$this->assertTrue($article->isNew());` should not happened

markstory: I would prefer an exception over `false`. Returning false makes our typehints hard and isn't as clear as an exception would be.

dereuromark:
Maybe this has to be splitted into 2 parts.
patching now in master as bugfix, persisting in 3.next as it does change the behavior more dramatically.

markstory:
@dereuromark Only fixing by applying validation would result in `findOrCreate()` returning unsaved entities which isn't great.

tests/TestCase/ORM/TableTest.php
markstory: The exception could 'break' application code, but currently that code is broken in a more serious way. Returning an unsaved entity goes against the goal/purpose of this method.

jeremyharris: I would prefer an exception as well, like saveOrFail, but how could the user retrieve the invalid entity to possibly do work on? Is getting the entity useful after a failed save?

ADmad: It could help in debugging/logging. We could create a new exception which contains the entity.

jeremyharris:
Per consensus I am throwing a PersistenceFailedException (which contains a getter for the entity) when the save fails during `findOrCreate`. Also re-targeted against 3.next

dereuromark (approved):

src/ORM/Table.php
```diff
@@ -1730,14 +1730,20 @@ protected function _processFindOrCreate($search, callable $callback = null, $opt
         }
         $entity = $this->newEntity();
         if ($options['defaults'] && is_array($search)) {
-            $entity->set($search, ['guard' => false]);
+            $entity = $this->patchEntity($entity, $search, ['guard' => false]);
```
ADmad: `guard` is not a valid option for `RepositoryInterface::patchEntity()` / `Marshaller::merge()`.  You need to use `accessibleFields` instead but requires specific field names as keys.

ADmad (changes requested):

tests/TestCase/ORM/TableTest.php
```diff
@@ -48,6 +49,13 @@ class UsersTable extends Table
 
 }
 
+class ProtectedEntity extends Entity
+{
+    protected $_accessible = [
+        'id' => true,
+    ];
```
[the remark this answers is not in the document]
jeremyharris: I went ahead and added it, even though I think accessibility is well enough tested in the marshaller ;)

ADmad (approved):

src/ORM/Table.php
```diff
@@ -1730,14 +1730,21 @@ protected function _processFindOrCreate($search, callable $callback = null, $opt
         }
         $entity = $this->newEntity();
         if ($options['defaults'] && is_array($search)) {
-            $entity->set($search, ['guard' => false]);
+            $accessibleFields = array_combine(array_keys($search), array_fill(0, count($search), true));
```
dereuromark: I am not too happy about this.
This creates still a dangerous case where only the submitted fields are validated, right?
As such a half validated entity (some of the required fields as per validation are not in errors list) could be persisted?

dereuromark: OK, never mind: The test case proofs that this is only about accessible fields, not about the validation fields on required at "create": https://github.com/cakephp/cakephp/pull/13066

PR #12917, opened 2019-01-20
4.x static analysis

ADmad:


src/View/Widget/SelectBoxWidget.php
```diff
@@ -228,14 +228,14 @@ protected function _renderOptgroup(
...
-     * @param array|\ArrayAccess|null $disabled The options to disable.
+     * @param array $options The options to render.
+     * @param array|null $disabled The options to disable.
      * @param array|string|null $selected The options to select.
      * @param array $templateVars Additional template variables.
      * @param bool $escape Toggle HTML escaping.
      * @return array Option elements.
      */
-    protected function _renderOptions($options, $disabled, $selected, $templateVars, $escape): array
+    protected function _renderOptions($options, ?array $disabled, $selected, $templateVars, $escape): array
```
markstory: Could disabled be an arrayaccess if an orm result found it's way down here?

ADmad: Nope, an entity never makes it's way as argument to this function.

PR #12924, opened 2019-01-25
Fix errors reported by phpstan level 3.

ADmad:


src/Database/ExpressionInterface.php
```diff
@@ -35,7 +35,7 @@ public function sql(ValueBinder $generator): string;
      * being iterated.
      *
      * @param callable $visitor The callable to apply to all nodes.
-     * @return $this
+     * @return $this|void
```
dereuromark: This cannot be. Didnt I remove this as the void is only inside the callable etc?

src/Console/Shell.php
```diff
@@ -37,7 +37,7 @@
  * Is the equivalent of Cake\Controller\Controller on the command line.
  *
  * @deprecated 3.6.0 ShellDispatcher and Shell will be removed in 5.0
- * @method int|bool|null main(...$args)
+ * @method int|bool|void main(...$args)
```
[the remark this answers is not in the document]
ADmad: I know but shells are deprecated anyway and I don't want to got about updating `main()` of all shells to ruturn null.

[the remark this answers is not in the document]
dereuromark: better to silence this then IMO. as this is dead wrong and will cause all inherit to declare the same wrong thing.

src/Database/ExpressionInterface.php
ADmad: `traverse()` method of all expression classes which implement this interface return void. Only `Query::traverse()` returns `$this`.

dereuromark: Couldnt all return this then? for a clean api?

src/Console/Shell.php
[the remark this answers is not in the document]
ADmad: Okay, I'll revert this and ignore the error then.

src/Database/ExpressionInterface.php
ADmad: Fine by me if @lorenzo / @markstory agree.

phpstan.neon
```diff
@@ -27,6 +27,7 @@ parameters:
         -
             message: '#Binary operation "-" between (.)* results in an error#'
             path: 'src/I18n/RelativeTimeFormatter.php'
+        - '#Static property Cake\\Chronos\\(Date|Chronos|MutableDate|MutableDateTime)::\$diffFormatter \(Cake\\Chronos\\DifferenceFormatter\) does not accept Cake\\I18n\\RelativeTimeFormatter#'
```
ADmad: A proper fix for this would be to add an interface for diff formatter to Chronos so that it property can be type hinted with the interface instead of concrete class.

phpstan.neon
```diff
@@ -27,6 +27,17 @@ parameters:
...
+        - '#Return type \(void\) of method Cake\\Shell\\[A-Za-z]+Shell::main\(\) should be compatible with return type \(bool|int|null\) of method Cake\\Console\\Shell::main\(\)#'
+        - '#Property Cake\\TestSuite\\Fixture\\TestFixture::\$_schema \(Cake\\Database\\Schema\\SqlGeneratorInterface&Cake\\Database\\Schema\\TableSchemaInterface\) does not accept Cake\\Database\\Schema\\TableSchemaInterface#'
+        - '#Method Cake\\ORM\\Query::find\(\) should return \$this\(Cake\\Datasource\\QueryInterface\) but returns Cake\\ORM\\Query#'
+        -
+            message: '#Offset mixed does not exist on array\(\)|array\(2 => mixed\)|array\(string, string\)#'
+            path: 'src/Database/Dialect/SqlserverDialectTrait.php'
+        -
+            message: '#Offset 0 does not exist on array\(\)#'
+            path: 'src/Database/Dialect/SqlserverDialectTrait.php'
+        - '#Return type \(null\) of method Cake\\Collection\\Iterator\\NoChildrenIterator::getChildren\(\) should be compatible with return type \(RecursiveIterator\) of method RecursiveIterator::getChildren\(\)#'
```
ADmad: @lorenzo What would be a proper fix for this?

ADmad:
There shouldn't be any errors reported at phpstan level 3 now :muscle:.

src/Cache/Engine/FileEngine.php
```diff
@@ -319,25 +319,31 @@ protected function _clearDirectory(string $path): void
...
+     * @return false
      * @throws \LogicException
      */
-    public function decrement(string $key, int $offset = 1): void
+    public function decrement(string $key, int $offset = 1)
     {
         throw new LogicException('Files cannot be atomically decremented.');
+
+        // phpcs:ignore
+        return false;
```
markstory: Why is this necessary? Seems like a phpstan issue.

src/Database/ExpressionInterface.php
markstory: Going from void -> this is safe.

src/Cache/Engine/FileEngine.php
ADmad: Docblock of CacheEngineInterface::decrement() says it should return a boolean or int, so phpstan is unhappy if there's no retuy value.

But yes logically return value should not be necessary due to the exception. I can revert and ignore the error instead.

dereuromark: We should Report this as phpstan bug. It clearly is.

ADmad: Turned out the only problem was the `void` return type hint.

PR #12928, opened 2019-01-27
Use ::class constant wherever possible.

ADmad:


inoas:
This should happen for 3.x, too as PHP 7.4 may deprecate the other usages?

markstory:
@inoas We can update 3.x when the deprecations happen in PHP.

PR #12934, opened 2019-01-28
4.x native datetime

ADmad:
As discussed in #12820, this PR updates `FormHelper` to generate native HTML inputs for all available datetime types. I have modified `DateTimeWidget` accordingly and added new `YearWidget` to generate selectbox for years.

Still have to do lot of cleanup.

What should be done for the individual `FormHelper()` methods for generating single selects like `day()`, `hour()`, `minute()`, `seconds()`, `meridian()`?  My vote is to just drop them. They can still be generated using `select()` with appropriate options list if some needs them.


dereuromark:
We should make sure that all the "old" ways have an easy upgrade path to an existing method somewhere.
That eases the migration efforts for existing apps IMO.

src/View/Widget/YearWidget.php
```diff
@@ -0,0 +1,147 @@
...
+            $data['end'] = $data['max'];
+        }
+
+        if (!empty($data['val'])) {
+            $data['start'] = min($data['val'], $data['start']);
+            $data['end'] = max($data['val'], $data['end']);
+        }
+
+        if ($data['end'] < $data['start']) {
+            throw new RuntimeException('Max year cannot be less than min year');
```
markstory: Could be a `LogicException` or `InvalidArgumentException` 🏌️

markstory (commented):
I second moving the existing widget into a plugin for people who want to upgrade but can't update their UI.

ADmad:
Along with existing DateTimeWidget the shim plugin will also need a FormHelper with retains all the options juggling that is currently done for datetime inputs.

markstory:
> will also need a FormHelper with retains all the options juggling that is currently done for datetime inputs.

Yeah that is going to be awkward 😢

src/Database/Type/DateTimeType.php
```diff
@@ -50,6 +50,7 @@ class DateTimeType extends BaseType
      */
     protected $_format = [
         'Y-m-d H:i:s',
+        'Y-m-d\TH:i:s',
```
dereuromark: Should we backport this to 3.x?

dereuromark:
All right, we have a branch now: https://github.com/dereuromark/cakephp-shim/tree/4.x#new-shims

ADmad:
Ancillary work of PR to Shim plugin and docs PR is done. This should be good to merge now.

Issue #10172, opened 2017-02-04
Assets files on theme plugins not loaded when set base URLs


BassemN:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.3.13.
* Platform and Target: Apache/2.4.23 , 10.1.16-MariaDB, AND PHP 5.5.38.

### What you did
`app.php` file
```php
        'imageBaseUrl' => 'http://cdn.domain.com/img/',
        'cssBaseUrl' => 'http://cdn.domain.com/css/',
        'jsBaseUrl' => 'http://cdn.domain.com/js/',
```
`bootstrap.php` file
```php
Plugin::load('CustomTheme', ['ignoreMissing' => true, 'bootstrap' => true, 'routes' => true]);
```
`AppController.php` file
```php
    public function beforeRender(Event $event)
    {
        parent::beforeRender($event);
        $this->viewBuilder()->theme('CustomTheme');
    }
```
`src/Template/Layout/default.ctp` file
```php
<?= $this->Html->css('app'); ?>
```
### What happened
The CSS file path is not loaded from the plugin although the app.css file located in the plugin as you can see below
```html
<link rel="stylesheet" href="http://cdn.domain.com/css/app.css"/>
```
### What you expected to happen
The css file should be loaded from the plugin as the following
```html
<link rel="stylesheet" href="http://cdn.domain.com/custom_theme/css/app.css"/>
```
**This issue only happens when setting imageBaseUrl, cssBaseUrl & jsBaseUrl**

markstory:
Is the problem that the app version's of `app.css` was used instead of the theme/plugin's `app.css`?

BassemN:
@markstory Yes, that's it.

BassemN:
@markstory Please note that this only happens when setting imageBaseUrl, cssBaseUrl & jsBaseUrl.

markstory:
[...]

PR #12935, opened 2019-01-28
New test to show how plugin based assets interact with BaseUrl config

steinkel:
Plugin based assets were not adding the plugin prefix when baseUrl was defined, for example, for working with a pull origin CDN.

### DIFF
--- a/tests/TestCase/View/Helper/UrlHelperTest.php
+++ b/tests/TestCase/View/Helper/UrlHelperTest.php
@@ -570,9 +570,5 @@ class UrlHelperTest extends TestCase
         $result = $this->Helper->css('TestTheme.app.css');
         $expected = $cdnPrefix . 'app.css';
         $this->assertSame($expected, $result);
-
-        Configure::write('App.imageBaseUrl', $imageBaseUrl);
-        Configure::write('App.jsBaseUrl', $jsBaseUrl);
-        Configure::write('App.cssBaseUrl', $cssBaseUrl);
     }
 }
--- a/tests/TestCase/View/Helper/UrlHelperTest.php
+++ b/tests/TestCase/View/Helper/UrlHelperTest.php
@@ -544,4 +544,35 @@ class UrlHelperTest extends TestCase
 
         Configure::write('App.wwwRoot', $webRoot);
     }
+
+    /**
+     * Test plugin based assets will NOT use the plugin name
+     *
+     * @return void
+     */
+    public function testPluginAssetsPrependImageBaseUrl()
+    {
+        $cdnPrefix = 'https://cdn.example.com/';
+        $imageBaseUrl = Configure::read('App.imageBaseUrl');
+        $jsBaseUrl = Configure::read('App.jsBaseUrl');
+        $cssBaseUrl = Configure::read('App.cssBaseUrl');
+        Configure::write('App.imageBaseUrl', $cdnPrefix);
+        $result = $this->Helper->image('TestTheme.text.jpg');
+        $expected = $cdnPrefix . 'text.jpg';
+        $this->assertSame($expected, $result);
+
+        Configure::write('App.jsBaseUrl', $cdnPrefix);
+        $result = $this->Helper->script('TestTheme.app.js');
+        $expected = $cdnPrefix . 'app.js';
+        $this->assertSame($expected, $result);
+
+        Configure::write('App.cssBaseUrl', $cdnPrefix);
+        $result = $this->Helper->css('TestTheme.app.css');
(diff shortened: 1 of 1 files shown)


src/View/Helper/UrlHelper.php
```diff
@@ -176,6 +177,10 @@ public function assetUrl($path, array $options = [])
             list($plugin, $path) = $this->_View->pluginSplit($path, false);
         }
         if (!empty($options['pathPrefix']) && $path[0] !== '/') {
+            if (isset($plugin)) {
+                $path = Inflector::underscore($plugin) . '/' . $path;
```
dereuromark: Is this really underscore, not dashed?

steinkel:
code should pass all tests now, I'm thinking a refactor to reduce the complexity...

steinkel:
Moving forward, there could be an explicit "CDN" configuration key, to be used as a full base when dealing with assets, and a xxxBaseUrl to deal with the relative prefix based on the type of asset, so we can build the url's using the same structure via `CDN . xxxBaseUrl . path` for assets. Fonts are a pretty common asset too, having no specific xxxBaseUrl prefix or methods in Html/Url helpers.

markstory:
This came up once before in #10172

src/View/Helper/UrlHelper.php
```diff
@@ -185,9 +198,10 @@ public function assetUrl($path, array $options = [])
             $path .= $options['ext'];
         }
         if (preg_match('|^([a-z0-9]+:)?//|', $path)) {
+
```
markstory: Extra line here.

src/View/Helper/UrlHelper.php
```diff
@@ -176,7 +177,19 @@ public function assetUrl($path, array $options = [])
             list($plugin, $path) = $this->_View->pluginSplit($path, false);
         }
         if (!empty($options['pathPrefix']) && $path[0] !== '/') {
-            $path = $options['pathPrefix'] . $path;
+            if (preg_match('|^([a-z0-9]+:)?//|', $options['pathPrefix'])) {
```
markstory: Are you guarding against the path prefix containing a URI / full path?

markstory (commented):
This has the potential to impact people's deployments as we're changing what an input path maps to in their CDN.

src/View/Helper/UrlHelper.php
steinkel: fixed

steinkel:
Yes, reading about the context in #10172 (thanks for the link) I see others could be using a flat directory structure to copy all assets, as mentioned there.
This behavior would lead to tricky issues to debug, as cdn enabled sites usually appear later in the release cycle. As assets could overwrite other assets with the same name.

The changes implemented here pass existing tests for Html and UrlHelper, but this case was not  documented or covered in test cases, so my initial reaction was "fixing" it :)
We could:
* fix the tests and push them alone, updating the docs to clarify how the UrlHelper will process plugin based assets & absolute urls in base paths. In this case I can work on a plugin to override UrlHelper and deal with cdn's to cover these issues. This is a 100% BC scenario.
* go ahead with this change, understanding the current behavior is not correct for that specific case. In this case we could wait for a minor release, or push it to 4.x

Please let me know your thoughts, :sun_with_face:

markstory:
Pushing this change to 4.0 might be safer. We could also use 4.x to add new/more complete configuration options to handle CDN deployments easier.

ADmad:
I think it would be best to add a new option in 4.x.

steinkel:
I'll rewrite the unit tests and push them alone, so others can see how the behavior works right now, thanks!

steinkel:
unit test left only in this PR

tests/TestCase/View/Helper/UrlHelperTest.php
```diff
@@ -544,4 +544,35 @@ public function testWebrootPaths()
...
+        $this->assertSame($expected, $result);
+
+        Configure::write('App.cssBaseUrl', $cdnPrefix);
+        $result = $this->Helper->css('TestTheme.app.css');
+        $expected = $cdnPrefix . 'app.css';
+        $this->assertSame($expected, $result);
+
+        Configure::write('App.imageBaseUrl', $imageBaseUrl);
+        Configure::write('App.jsBaseUrl', $jsBaseUrl);
+        Configure::write('App.cssBaseUrl', $cssBaseUrl);
```
markstory: You shouldn't need to manually restore these values. The test case base class snapshots configure on each test.

steinkel: fixed, a little too paranoid...

PR #12936, opened 2019-01-29
Improve error messages when persistence fails

### LINKED ISSUE #12931: SaveOrFail with errors in hasMany associated data -> "Array to string conversion" error
This is a:

* [x] bug
* [ ] enhancement

CakePHP Version: 3.7.3

Gist with necessary files to reproduce here:
[https://gist.github.com/andres-ml/93aea170698a0e1607c2cf6e135a0028](https://gist.github.com/andres-ml/93aea170698a0e1607c2cf6e135a0028)

### What you did
Save an entity with nested associations that don't pass rules check using the saveOrFail() method.

### What happened
Notice Error: Array to string conversion in [(...)/vendor/cakephp/cakephp/src/ORM/Exception/PersistenceFailedException.php, line 76]
Also, the exception's error message shows "Array" instead of an error message:
Entity save failure (association: "Array")

At least, the error shouldn't trigger. As of how the error message should be, I'm not sure; maybe a comma-separated list of all errors.

### Possible fix
Making PersistenceFailedException::buildError recursive (line 73):
`$errors[] = is_int($key) ? static::buildError($message) : $key;`

markstory:
When persistence fails error messages should include all the validation errors recursively.

Fixes #12931

### DIFF
--- a/src/ORM/Exception/PersistenceFailedException.php
+++ b/src/ORM/Exception/PersistenceFailedException.php
@@ -14,6 +14,7 @@ namespace Cake\ORM\Exception;
 
 use Cake\Core\Exception\Exception;
 use Cake\Datasource\EntityInterface;
+use Cake\Utility\Hash;
 
 /**
  * Used when a strict save or delete fails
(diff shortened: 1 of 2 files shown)


src/ORM/Exception/PersistenceFailedException.php
```diff
@@ -14,6 +14,7 @@
 
 use Cake\Core\Exception\Exception;
 use Cake\Datasource\EntityInterface;
+use Cake\Utility\Hash;
```
markstory: cakephp/orm already depends on utility. https://github.com/cakephp/cakephp/blob/8b7ffba5cd82cdbaea0a3c75bcbc54aebcd2b2c6/src/ORM/composer.json#L32

src/ORM/Exception/PersistenceFailedException.php
```diff
@@ -47,35 +48,17 @@ public function __construct(EntityInterface $entity, $message, $code = null, $pr
...
         parent::__construct($message, $code, $previous);
     }
 
-    /**
-     * @param string|array $error Error message.
-     * @return string
-     */
-    protected function buildError($error)
-    {
-        if (!is_array($error)) {
```
dereuromark: We dont need to keep this part? String check?

markstory: Hash::flatten() always returns a flat key : value pairing.

dereuromark: Neat then as it simplies this a lot here :)

Issue #12933, opened 2019-01-28
Cookie::getExpiresTimestamp() returns string - Type error when calling `set_cookie`


cleptric:
`ResponseEmitter::emitCookies` calls `set_cookie`, which throws an `TypeError`, as the third parameter should be an `integer`.
The reason is the return value of `Cookie::getExpiresTimestamp()`, which is a `string`.
I'm not sure if we simply can change the return value of `getExpiresTimestamp` or typecast the value on the way to `emitCookies`

ADmad:
First issue reported which is caused by strict typing, yay! :tada:

dereuromark:
wouldnt phpstan have done the same if the code was cleaner (level4+)^^ 

jokes aside: nice work!

ADmad:
The docblock of `Cookie::getExpiresTimestamp()` states it return string to prevent integer overflow on 32-bit systems. But even core PHP methods like `DateTime::getTimestamp()` return integer so 32-bit systems are gonna have problem one way or another when dealing with datetime/timestamps values far in future.

So I think we could just change `Cookie::getExpiresTimestamp()` to return integer.

dereuromark:
what timestamp would cause this? must be many years in the future?

ADmad:
> must be many years in the future?

Yup

ADmad:
Since values than max int overflow to float PHP itself will take care of throwing error if the method has return type hint as `int`.

cleptric:
PR open.

PR #12939, opened 2019-01-29
Change Cookie::getExpiresTimestamp() return type

cleptric:
see #12933

### DIFF
--- a/src/Http/Cookie/CookieInterface.php
+++ b/src/Http/Cookie/CookieInterface.php
@@ -114,7 +114,7 @@ interface CookieInterface
     /**
      * Get the timestamp from the expiration time
      *
-     * @return int|null The expiry time as a string timestamp.
+     * @return int|null The expiry time as an integer.
      */
     public function getExpiresTimestamp(): ?int;
 
--- a/src/Http/Cookie/CookieInterface.php
+++ b/src/Http/Cookie/CookieInterface.php
@@ -114,9 +114,6 @@ interface CookieInterface
     /**
      * Get the timestamp from the expiration time
      *
-     * Timestamps are integer as large timestamps can overflow MAX_INT
-     * in 32bit systems.
-     *
      * @return int|null The expiry time as a string timestamp.
      */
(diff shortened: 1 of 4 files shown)


src/Http/Cookie/CookieInterface.php
```diff
@@ -114,12 +114,12 @@ public function getExpiry();
     /**
      * Get the timestamp from the expiration time
      *
-     * Timestamps are strings as large timestamps can overflow MAX_INT
+     * Timestamps are integer as large timestamps can overflow MAX_INT
```
ADmad: This statement needs to be removed, it makes no sense now :).

src/Http/Cookie/CookieInterface.php
```diff
@@ -114,12 +114,9 @@ public function getExpiry();
     /**
      * Get the timestamp from the expiration time
      *
-     * Timestamps are strings as large timestamps can overflow MAX_INT
-     * in 32bit systems.
-     *
-     * @return string|null The expiry time as a string timestamp.
+     * @return int|null The expiry time as a string timestamp.
```
ADmad: s/string/

cleptric: `The expiry time as an integer timestamp.` ?

markstory: Or 'expiry time as an integer'

src/Http/Cookie/CookieInterface.php
```diff
@@ -114,7 +114,7 @@ public function getExpiry();
     /**
      * Get the timestamp from the expiration time
      *
-     * @return int|null The expiry time as a string timestamp.
+     * @return int|null The expiry time as an integer.
```
dereuromark: How can this be null? I understand that 0 exists as unlimited, but what is null?

ADmad: It's probably null because setcookie() allows the $expires argument to be null.

ADmad (approved):

Issue #12925, opened 2019-01-25
Paginator::paginate should be splitted


rochamarcelo:
The Paginator::paginate should be splitted in at least two methods, one responsible to create find options and query for results, and other method to define paging params. Additionally we could move to other method the code to create the find options.

This would make easier to overwrite this class and make the code cleaner.

https://github.com/cakephp/cakephp/blob/d2bd61dbe40ec9db28057bc7e5004703452568e6/src/Datasource/Paginator.php#L187

If this is okay, I could open a pull request for it. Please let me know what you think.

markstory:
Factoring out the different concerns sounds like a reasonable idea to me as long as we don't modify the existing public interfaces

> This would make easier to overwrite this class and make the code cleaner.

Instead of subclassing you could create your own Paginator and modify it to suit your needs. Adding more hooks for subclasses to use isn't something I would plan on. We don't offer backwards compatibility promises on non-public methods and your code could break in a future minor release.

rochamarcelo:
When I create a custom paginator I don't want to copy/paste all the methods used to create the options (getDefaults, mergeOptions, validateSort and checkLimit, _pagingParams). To avoid copy/paste without extending we would need to move these methods to a trait or a custom class, what do you think?

dereuromark:
@ramiroaraujo I recommend making a basic PR that showcases the changes you want to do.
Then it will be easier to approve or decline it. You dont have to also adjust test cases just yet, only the src code directly as a discussion starter.

dereuromark:
PR open.

PR #12943, opened 2019-01-31
Paginator::paginate() splitted

rochamarcelo:
An implementation for issue #12925


dereuromark:
https://github.com/cakephp/cakephp/pull/13524 superseeds this one as the public API is more BC.

PR #12947, opened 2019-02-02
4.x Make a QueryInterface::getRepository() a real method

markstory:
This method was marked as a virtual method but there isn't a good reason it can't be a real method in 4.x

### DIFF
--- a/src/Datasource/QueryInterface.php
+++ b/src/Datasource/QueryInterface.php
@@ -21,7 +21,6 @@ namespace Cake\Datasource;
  *
  * @method $this andWhere($conditions, $types = [])
  * @method $this select($fields = [], $overwrite = false)
- * @method \Cake\Datasource\RepositoryInterface getRepository()
  */
 interface QueryInterface
 {
@@ -258,13 +257,21 @@ interface QueryInterface
     public function toArray(): array;
 
     /**
-     * Returns the default repository object that will be used by this query,
-     * that is, the repository that will appear in the from clause.
+     * Set the default Table object that will be used by this query
+     * and form the `FROM` clause.
      *
      * @param \Cake\Datasource\RepositoryInterface $repository The default repository object to use
      * @return $this
      */
     public function repository(RepositoryInterface $repository);
+
+    /**
+     * Returns the default repository object that will be used by this query,
+     * that is, the repository that will appear in the from clause.
+     *
+     * @return \Cake\Datasource\RepositoryInterface|mixed $repository The default repository object to use
+     */
+    public function getRepository();
 
     /**
      * Adds a condition or set of conditions to be used in the WHERE clause for this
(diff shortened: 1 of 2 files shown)


src/Datasource/QueryInterface.php
```diff
@@ -258,14 +257,22 @@ public function page(int $num, ?int $limit = null);
...
     /**
-     * Returns the default repository object that will be used by this query,
-     * that is, the repository that will appear in the from clause.
+     * Set the default Table object that will be used by this query
+     * and form the `FROM` clause.
      *
      * @param \Cake\Datasource\RepositoryInterface $repository The default repository object to use
      * @return $this
      */
     public function repository(RepositoryInterface $repository);
```
cleptric: Shall we call it `setRepository()`?

saeideng: in other cases we dont have setX methods for query methods , and just `x()` for setter  and `getX()`

src/Datasource/QueryInterface.php
```diff
@@ -258,14 +257,22 @@ public function page(int $num, ?int $limit = null);
...
 
+    /**
+     * Returns the default repository object that will be used by this query,
+     * that is, the repository that will appear in the from clause.
+     *
+     * @return \Cake\Datasource\RepositoryInterface|mixed $repository The default repository object to use
```
dereuromark: why mixed?

dereuromark:
> This method was marked as a virtual method but there isn't a good reason it can't be a real method in 4.x

Yeah, I wouldnt call this a virtual method, it looks to me this was mainly done to keep BC. Using an annotation on the interface allowed it to be visible without breaking code.
We can now indeed remove all those annotations in favor of real interface methods now.

src/Datasource/QueryInterface.php
```diff
@@ -258,14 +257,22 @@ public function page(int $num, ?int $limit = null);
...
+     */
+    public function getRepository();
```
dereuromark: Since the setter uses typehints, it might be better to also use typehint here: `: RepositoryInterface`

src/Datasource/QueryInterface.php
cleptric: yep, I remeber this now 👍

src/Datasource/QueryInterface.php
markstory: I'll switch it.

src/Datasource/QueryInterface.php
markstory: Won't that cause phpstan to whine when callers use methods outside of the repository interface?

dereuromark: But how is the setter working then?

markstory: I don't think we call the setter often.

PR #12948, opened 2019-02-03
4.x - Middleware queue updates

ADmad:
- `MiddlewareQueue` now implements `SeekableIterator` to avoid having to maintain iteration index in `Runner`.
- Using a separate array to maintain resolved middleware instances seemed unnecessary, so I got rid of `MiddlewareQueue::$middlewares` and `MiddlewareQueue::$queue` is directly updated now.

### DIFF
--- a/src/Http/MiddlewareQueue.php
+++ b/src/Http/MiddlewareQueue.php
@@ -97,6 +97,10 @@ class MiddlewareQueue implements Countable, SeekableIterator
             $middleware = new $className();
         }
 
+        if ($middleware instanceof MiddlewareInterface) {
+            return $middleware;
+        }
+
         if (!$middleware instanceof Closure) {
             return new DoublePassDecoratorMiddleware($middleware);
         }
--- a/src/Http/MiddlewareQueue.php
+++ b/src/Http/MiddlewareQueue.php
@@ -252,7 +252,8 @@ class MiddlewareQueue implements Countable, SeekableIterator
      * @return void
      * @see \SeekableIterator::seek()
      */
-    public function seek($position) {
+    public function seek($position)
+    {
         if (!isset($this->queue[$position])) {
             throw new OutOfBoundsException("Invalid seek position ($position)");
         }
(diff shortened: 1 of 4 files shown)


ADmad:
@markstory Any feedback?

src/Http/Runner.php
```diff
@@ -72,9 +65,10 @@ public function run(
      */
     public function handle(ServerRequestInterface $request): ResponseInterface
     {
-        $middleware = $this->queue->get($this->index++);
+        if ($this->queue->valid()) {
+            $middleware = $this->queue->current();
+            $this->queue->next();
```
markstory: Operating an iterator like this is a bit clunky would `yield` be usable here?

ADmad: Nope it isn't.

Runners in other libs I check also traverse the queue manually. `Relay` use `current()` and `next()` functions on it's queue which is an array. `zend-stratigility`'s queue is an `SplQueue` and it's runner uses `dequeue()` method to get next middleware.

ADmad:
I also got rid of `MiddlewareQueue::get()` method. It was only used in tests. Picking middlewares from arbitrarily positions isn't required in normal usage of the queue.

src/Http/Runner.php
markstory: Ok. It was worth a shot.

src/Http/MiddlewareQueue.php
```diff
@@ -55,37 +57,15 @@ public function __construct(array $middleware = [])
...
     }
 
-    /**
-     * Get the middleware at the provided index.
-     *
-     * @param int $index The index to fetch.
-     * @return \Psr\Http\Server\MiddlewareInterface|null Either the middleware or null
-     *   if the index is undefined.
-     */
-    public function get(int $index): ?MiddlewareInterface
```
markstory: Do you think someone will be using this? We haven't deprecated this method in 3.7.

ADmad: I don't see any reason why someone would need to get a middleware from queue themselves.

ADmad: We can mark it deprecated in 4.0 and keep it until 5.0 but I really don't see the need.

markstory: Ok. Let's remove it and see what happens.  :smile;

ADmad: Cool, this should be good to merge then :slightly_smiling_face:.

PR #12952, opened 2019-02-05
Use field instead of property for Entity.

dereuromark:
I would like to propose API improvement for Entities for a more consistent API here.

Main changes:

- getVisible() for visible fields (in sync with getHidden, getVirtual, getDirty, getInvalid, getOriginal method names)
- unsetField() for unsetting fields

For BC and since the interface was quite public for a long time:

- visibleProperties() kept until 5.0, but deprecated
- unsetProperty() kept until 5.0, but deprecated

still there to be used via trait, but not part of the interface contract for core.

All variable renames are BC as they are scoped inside the methods.
The protected properties have a low reach in terms of BC.

So for most people nothing changes in usability here when upgrading.

```
/**
 * An entity represents a single result row from a repository. It exposes the
 * methods for retrieving and storing fields associated in this row.
 */
```

As the entity docblock states: The main idea is to use "field" as name, in sync with DB fields.
The word property should only be used when referring to the technical part of exposing them as such on the object ("object property access" read/write). The term should not leak into the high level description of having multiple fields here in an entity. Especially since they are not even stored as single properties, but grouped as key=>value fields inside an array property here.

Follows https://github.com/cakephp/docs/pull/5961 for 3.0 now as 4.0 cleanup.

With the included deprecations this change seems to be rather harmless.
What do you think? It would make a lot of sense in terms of harmonizing the wording/naming-scheme in the public API.

Things to discuss:

- Can we split setAccess() into single and multi to add typehints? Or can we simply say setAccess() is array always, similar to some other methods?

### DIFF
--- a/src/Datasource/EntityInterface.php
+++ b/src/Datasource/EntityInterface.php
@@ -162,10 +162,10 @@ interface EntityInterface extends ArrayAccess, JsonSerializable
      * Returns an array with the requested original fields
      * stored in this entity, indexed by field name.
      *
-     * @param array $field List of fields to be returned
+     * @param array $fields List of fields to be returned
      * @return array
      */
-    public function extractOriginal(array $field);
+    public function extractOriginal(array $fields);
 
     /**
      * Returns an array with only the original fields
--- a/src/Datasource/EntityInterface.php
+++ b/src/Datasource/EntityInterface.php
@@ -27,57 +27,57 @@ use JsonSerializable;
 interface EntityInterface extends ArrayAccess, JsonSerializable
 {
     /**
-     * Sets hidden properties.
+     * Sets hidden fields.
      *
-     * @param array $properties An array of properties to hide from array exports.
-     * @param bool $merge Merge the new properties with the existing. By default false.
+     * @param array $fields An array of fields to hide from array exports.
+     * @param bool $merge Merge the new fields with the existing. By default false.
      * @return $this
      */
-    public function setHidden(array $properties, bool $merge = false);
+    public function setHidden(array $fields, bool $merge = false);
 
     /**
-     * Gets the hidden properties.
+     * Gets the hidden fields.
      *
      * @return array
      */
     public function getHidden(): array;
 
     /**
-     * Sets the virtual properties on this entity.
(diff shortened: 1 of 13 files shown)


src/Datasource/EntityInterface.php
```diff
@@ -159,91 +159,91 @@ public function setSource(string $alias);
...
 
     /**
-     * Get the list of visible properties.
+     * Get the list of visible fields.
      *
-     * @return array A list of properties that are 'visible' in all representations.
+     * @return array A list of fields that are 'visible' in all representations.
      */
-    public function visibleProperties();
+    public function getVisible(): array;
```
markstory: Creating backwards breaking changes for terminology cleanup seems harsh. Having both available lets us mark one as deprecated and document the preferred one.

dereuromark: This is in this PR.

markstory: My mistake I missed that.

lorenzo:
I like this!

src/Datasource/EntityInterface.php
josegonzalez: Is dropping it from the trait BC?

src/Datasource/EntityTrait.php
```diff
@@ -500,80 +512,91 @@ public function getVirtual(): array
...
-        return array_diff($properties, $this->_hidden);
+    /**
+     * Gets the list of visible fields.
+     *
+     * @deprecated 4.0.0 Use getVisible() instead. Will be removed in 5.0.
+     * @return array
+     */
+    public function visibleProperties(): array
+    {
+        return $this->getVisible();
```
josegonzalez: Should this trigger a warning?

ADmad: We have added any warnings for new deprecations in 4.0.

dereuromark: Not inside 4.0.0 IMO, I guess these kinds of things could be useful more towards 4.1, 4.2 when starting to prepare for 5.0 migration.

josegonzalez: I guess so :)

PR #12953, opened 2019-02-05
Remove noise docblock annotations.

dereuromark:
These annotations shouldn't be needed, and the other trait usage proofs this.
The tools understand the "use ..." already, so adding this on top is duplicate info. 
Let's remove this noise then.


dereuromark:
tests look green.

ADmad (approved):
I bet these must have added at some point by IDE users, good riddance :slightly_smiling_face:

dereuromark:
Otherwise I would have been forced to write an Annotator for this in IdeHelper. To keep the annotations in sync with use statements. No thank you :trollface:

Issue #11751, opened 2018-02-23
RFC: 4.0+ Roadmap - Remove magic method


davidyell:
This is a (multiple allowed):

* [ ] bug
* [ ] enhancement
* [x] feature-discussion (RFC)

* CakePHP Version: 4.0 onwards

### What you did
Tried to mock the `\Cake\ORM\Query` class, to create a fixed return. My IDE was suggesting that I could mock `Cake\ORM\Query::extract()` even though this method does not exist.

[It's linked via a docblock](https://github.com/cakephp/cakephp/blob/master/src/ORM/Query.php#L42), but doesn't exist in the class.

### What happened
I was unable to achieve this and had to use a Fixture, and actually execute the query with data. Rather than just mocking the query and having a fixed return.

### What you expected to happen
I expected to find the methods which I wanted to mock in the class. Then setup a mock of the `Cake\ORM\Query` object and configure a fixed return of either an array, or `\Cake\ORM\ResultSet`.

### How did I solve the issue
I had to ask both my colleagues and in the Slack channel to find out how the `\Cake\ORM\Query` class was implementing Collection class methods, when there was no obvious extension, trait, or interface implementation shown in the code.

### What was the solution
It turns out the `\Cake\Datasource\QueryTrait` has a `__call()` method. [Shown here](https://github.com/cakephp/cakephp/blob/master/src/Datasource/QueryTrait.php#L458-L477)

As I am not overly familiar with magic methods. I was reading the source code, and 'chasing functions' in my IDE, I was unable to work out how all the methods were being made available. The code did not describe it's behaviour to me, making it hard to reason about.

### RFC
I'd like to suggest an addition to the 4.x Roadmap to remove this magic method, and add concrete implementation in the ORM, so that the source code is easier to reason about, requires less docblock links and is more obvious when being read.

### Side note
At PHP Conf UK one year I was talking to attendees about  frameworks and they had a poor opinion of CakePHP because of its use of magic methods and it's "black box" approach. I feel that things like this are validating that opinion, rather than challenging it.

markstory:
Fair points.

dereuromark:
This will then also require less [ide annotations](https://github.com/dereuromark/cakephp-ide-helper/) maybe in the future for some areas.
More direct tooling and IDE - and developer - understanding of the code itself then sounds like a good idea.

But I am not sure if we should keep this ticket open here, if there are no more concrete things to discuss or create a plan of attack from this.

davidyell:
[...]

PR #12959, opened 2019-02-07
Collection trait: instantiate new collections through overridable method

dorxy:
When using the collection trait in any class, appending items or using other methods which return new collections will always return a Collection object.

If I for example wish to create my own ResultSetInterface implementation, containing its own collection the following will happen:

```
$resultSet = new ResultSet();
$a = $resultSet->append('item');
// $a is now a Collection object, but I want it to remain a ResultSet object
```

With the changes proposed, I can implement the following in my ResultSet class:
```
protected $myItems;
public function __construct($myFirstArgument, $myItems){
      $this->myItems = $myItems;
      // do something else with other arguments
}
protected function newCollection(...$args){
   return new static($this->getMyFirstArgument(), ...$args);
}
```

### DIFF
--- a/src/Collection/CollectionTrait.php
+++ b/src/Collection/CollectionTrait.php
@@ -48,7 +48,7 @@ trait CollectionTrait
      * Allows classes which use this trait to determine their own
      * type of returned collection interface
      *
-     * @param mixed ...$args
+     * @param mixed $args,.. Constructor arguments.
      * @return CollectionInterface
      */
     protected function newCollection(...$args)
--- a/src/Collection/CollectionTrait.php
+++ b/src/Collection/CollectionTrait.php
@@ -49,7 +49,7 @@ trait CollectionTrait
      * type of returned collection interface
      *
      * @param mixed $args,.. Constructor arguments.
-     * @return CollectionInterface
+     * @return \Cake\Collection\CollectionInterface
      */
     protected function newCollection(...$args)
     {
--- a/src/Collection/CollectionTrait.php
+++ b/src/Collection/CollectionTrait.php
@@ -48,7 +48,7 @@ trait CollectionTrait
      * Allows classes which use this trait to determine their own
      * type of returned collection interface
      *
-     * @param mixed $args,.. Constructor arguments.
+     * @param mixed ...$args Constructor arguments.
      * @return \Cake\Collection\CollectionInterface
      */
     protected function newCollection(...$args)
--- a/src/Collection/CollectionTrait.php
+++ b/src/Collection/CollectionTrait.php
@@ -545,7 +545,7 @@ trait CollectionTrait
      */
     public function prepend($items)
     {
-        return ($this->newCollection($items))->append($this);
+        return $this->newCollection($items)->append($this);
     }
 
     /**
@@ -645,7 +645,7 @@ trait CollectionTrait
             $parents[$key][$nestingKey] = $children;
         };
 
-        return ($this->newCollection(new MapReduce($this->unwrap(), $mapper, $reducer)))
+        return $this->newCollection(new MapReduce($this->unwrap(), $mapper, $reducer))
             ->map(function ($value) use (&$isObject) {
                 /** @var \ArrayIterator $value */
                 return $isObject ? $value : $value->getArrayCopy();
--- a/src/Collection/CollectionTrait.php
+++ b/src/Collection/CollectionTrait.php
@@ -521,7 +521,7 @@ trait CollectionTrait
     {
         $list = new AppendIterator();
         $list->append($this->unwrap());
-        $list->append(($this->newCollection($items))->unwrap());
+        $list->append($this->newCollection($items)->unwrap());
 
         return $this->newCollection($list);
     }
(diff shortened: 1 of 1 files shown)


src/Collection/CollectionTrait.php
```diff
@@ -35,13 +35,27 @@
...
     use ExtractTrait;
 
+    /**
+     * Returns a new collection.
+     *
+     * Allows classes which use this trait to determine their own
+     * type of returned collection interface
+     *
+     * @param mixed $args,.. Constructor arguments.
+     * @return CollectionInterface
```
dereuromark: fqcn please

dorxy: Fixed :)

dereuromark:
https://travis-ci.org/cakephp/cakephp/jobs/490021009#L579
You need to find a way to keep this compatible here for 3.x

src/Collection/CollectionTrait.php
```diff
@@ -35,13 +35,27 @@
...
+    /**
+     * Returns a new collection.
+     *
+     * Allows classes which use this trait to determine their own
+     * type of returned collection interface
+     *
+     * @param mixed $args,.. Constructor arguments.
+     * @return \Cake\Collection\CollectionInterface
+     */
+    protected function newCollection(...$args)
```
dereuromark: Why not just using the same typehint as the collection constructor?

    public function __construct(iterable $items)

dereuromark:
As this aims to be a fully BC compatibility fix, I think master is fine in this case.
Going from `new X()` to a protected wrapper for X seems safe to me.

ceeram:
How about having it create instance of the class that is actually using the trait by default. I ran into this same issue this week, formatResults was passing a ResultSet. I hadded some chained invokable classes, only the first one can typehint ResultSet, the others will get Collections once the first invokable did some filtering/mapping. I noticed this when changing the order of those invokeable classes.

My suggestion would be to have it by default look something like:

```
    /**
     * Helper method to ensure Collection of same type is being returned
     *
     * @param array|\Traversable $items Items.
     */
    protected function newCollection($items)
    {
        $collection = get_class($this);

        return new $collection($items);
    }
```

ADmad:
Why not just change the `new Collection()` calls in methods to `new static()`?

dereuromark:
@ADmad
Could this be a discovery issue of the class for tooling?

dereuromark:
@ADmad
Could this be a discovery issue of the class for tooling?

ADmad:
Why should it be?  `new static()` is already used in multiple other classes.

ADmad:
Following by idea you end up with errors like this when running the test suite, so scratch that :slightly_smiling_face:.

`ArgumentCountError: Too few arguments to function Cake\Collection\Iterator\ReplaceIterator::__construct(), 1 passed in /srv/sites/cake-apps/cakephp/src/Collection/CollectionTrait.php on line 694 and exactly 2 expected`

src/Collection/CollectionTrait.php
ADmad: Because child classes of `Collection` can have different constructor signatures. For e.g. `ReplaceIterator` has `__construct($items, callable $callback)`.

dereuromark:
If we can't find a PHP5.6 compatible solution, we will have to target 4.0 here.

src/Collection/CollectionTrait.php
```diff
@@ -507,9 +521,9 @@ public function append($items)
     {
         $list = new AppendIterator();
         $list->append($this->unwrap());
-        $list->append((new Collection($items))->unwrap());
+        $list->append(($this->newCollection($items))->unwrap());
```
ADmad: Removing the extra parenthesis around `$this->newCollection($items)` should make PHP 5.6 happy.

ADmad:
Shouldn't this target `3.next`?

src/Collection/CollectionTrait.php
dorxy: Somehow I did not receive updates on this pull request. I have removed the double parenthesis.

dorxy:
> Shouldn't this target `3.next`?

If I need to change the target please let me know :)

T0mats0:
Thank you.  This will be very useful.

src/Collection/CollectionTrait.php
```diff
@@ -531,7 +545,7 @@ public function appendItem($item, $key = null)
      */
     public function prepend($items)
     {
-        return (new Collection($items))->append($this);
+        return ($this->newCollection($items))->append($this);
```
ADmad: Extra braces here too which PHP 5.6 does not like.

dorxy: Found one more, both removed

src/Collection/CollectionTrait.php
```diff
@@ -35,13 +35,27 @@
...
 
     use ExtractTrait;
 
+    /**
+     * Returns a new collection.
+     *
+     * Allows classes which use this trait to determine their own
+     * type of returned collection interface
+     *
+     * @param mixed $args,.. Constructor arguments.
```
ADmad: Should be `...$args` instead of `$args,..`

ADmad:
I think this enhancement could also help resolve #11751. `ORM\Query` could then use `CollectionTrait` and override the `getCollection()` method to return decorator instance and we could get rid of the `QueryTrait::__call()` method.

ADmad:
@markstory Anything more to be done here? I just think it would better to target this to `3.next`.

PR #12961, opened 2019-02-07
Fix classes to standalone class files. Added missing mixin.

dereuromark:
Resolves the first part of https://github.com/cakephp/cakephp-codesniffer/pull/230
- Controller

also fixes some missing inline docs for typehinting, and shows the missing $mixin that is required to be added here.


markstory (approved):
Seems ok.

dereuromark:
OK, then I prepare the rest over the next days.

PR #12964, opened 2019-02-07
Revert DigestAuthenticate default secret 

cleptric:


cleptric:
I need to set a default value for `Security::salt` while running the tests.

PR #12966, opened 2019-02-09
Make optional dependency more clear.

### LINKED ISSUE #12908: Core makes use of Cache but it's composer.json doesn't require it
This is a (multiple allowed):

* [x] bug
* [ ] enhancement

* CakePHP Version: 3.7.

### What you did
I'm trying to make usage of the standalone Cakephp packages such as utility, core, etc. to be able to start migrating an application to latest version of cake. I just notice Cake\Core uses Cake\Cache but it's composer.json doesn't require it.

markstory:
Fixes #12908

### DIFF
--- a/src/Core/Configure.php
+++ b/src/Core/Configure.php
@@ -446,6 +446,9 @@ class Configure
         if ($data === null) {
             $data = static::$_values;
         }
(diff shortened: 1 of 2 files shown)


ADmad:
I fixed the missing comma in composer.json.

PR #12967, opened 2019-02-11
4.x Improve missing template errors

markstory:
Add more context to missing template/layout/element/cell view errors. I often wonder if the file is missing because the search paths are wrong. Now error messages will show which paths were searched.

I've changed the view exception inheritance so that all missing X exceptions are subclasses of `MissingTemplateException`. I think this makes sense and will enable callers to write fewer catch blocks when interacting with `View`.

I've not updated the error message HTML output but will do so in another pull request.


src/View/Exception/MissingCellViewException.php
```diff
@@ -13,15 +13,48 @@
  */
 namespace Cake\View\Exception;
 
-use Cake\Core\Exception\Exception;
-
 /**
  * Used when a view file for a cell cannot be found.
  */
-class MissingCellViewException extends Exception
+class MissingCellViewException extends MissingTemplateException
```
ADmad: Nitpick: Perhaps the class name could be changed to `MissingCellTemplateException`.

Issue #12958, opened 2019-02-07
Session is cleared after setting Flash message in tests


robertpustulka:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.7.3

### What you did
I have a controller action like this:

```php
//FooController.php
public function bar()
{
    $this->Flash->success('Info');
    $data = $this->request->getSession()->read('data');
    // do things with data
}
```

### What happened
While performing integration tests session data is missing and my logic fails. Session contains the flash message only. 

```php
public function testBar()
{
    $this->session(['data' => 'some data']);
    $this->post('/foo/bar');

    debug($this->_requestSession->read()); // 'data' key is no longer there
}
```

### What you expected to happen

I expected my session data stays there after setting flash messages.

### What do you think happened

`FlashComponent` stores the session object on component instantiation and operates on that object afterwards. Changing `$this->_session->...()` calls with `$this->getController()->getRequest()->getSession()->...()` seems to solve the issue so I suspect it's something related to request immutability.

dereuromark:
Isn't there a retainSessionData flag somewhere?

robertpustulka:
There's `enableRetainFlashMessages` but that's not it.

markstory:
Do you have a small controller action & test method that could help reproduce this?

robertpustulka:
I wasn't able to reproduce this on a clean app install. I narrowed it down to the DI plugin I use. This plugin provides a [custom controller factory](https://github.com/robotusers/cakephp-di/blob/master/src/Http/ControllerFactory.php) which sets the request and response after the controller is instantiated. Since FlashComponent stores it's request during constructor time, the controller and component get different request objects and different session objects.

This is a limitation to the plugin of course, but it also means that swapping the request after the controller is created will not change the request referenced in a component which can be considered as another limitation. Maybe we could change the FlashComponent behavior and get the currently referenced request from the controller itself each time the flash messages are accessed?

markstory:
>  Maybe we could change the FlashComponent behavior and get the currently referenced request from the controller itself each time the flash messages are accessed?

I don't see why that shouldn't be possible.

PR #12970, opened 2019-02-13
Add FlashComponent Session getter.

robertpustulka:
Refs #12958

This prevents a situation when the outdated session object is referenced by a controller.
Since PSR-7 request and response objects are immutable, request may be changed at runtime along with the session object.

### DIFF
--- a/src/Controller/Component/FlashComponent.php
+++ b/src/Controller/Component/FlashComponent.php
@@ -35,7 +35,7 @@ class FlashComponent extends Component
      * The Session object instance
      *
      * @var \Cake\Http\Session
-     * @deprecated 3.8.0 This property will be removed in 4.0.0 in favor of `getSession()` method.
+     * @deprecated 3.7.5 This property will be removed in 4.0.0 in favor of `getSession()` method.
      */
     protected $_session;
 
--- a/src/Controller/Component/FlashComponent.php
+++ b/src/Controller/Component/FlashComponent.php
@@ -35,6 +35,7 @@ class FlashComponent extends Component
      * The Session object instance
      *
      * @var \Cake\Http\Session
+     * @deprecated 3.8.0 This property will be removed in 4.0.0 in favor of `getSession()` method.
      */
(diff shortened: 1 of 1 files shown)


src/Controller/Component/FlashComponent.php
```diff
@@ -31,13 +30,6 @@
 class FlashComponent extends Component
 {
 
-    /**
-     * The Session object instance
-     *
-     * @var \Cake\Http\Session
-     */
-    protected $_session;
```
saeideng: you can not delete `protected` member in bugfix,  for BC reasons

dereuromark: We could have __...() to proxy then, right?
But yeah, people could be using the property internally right now.

src/Controller/Component/FlashComponent.php
```diff
@@ -51,18 +43,6 @@ class FlashComponent extends Component
...
-    /**
-     * Constructor
-     *
-     * @param \Cake\Controller\ComponentRegistry $registry A ComponentRegistry for this component
-     * @param array $config Array of config.
-     */
-    public function __construct(ComponentRegistry $registry, array $config = [])
-    {
-        parent::__construct($registry, $config);
-        $this->_session = $registry->getController()->getRequest()->getSession();
```
markstory: We should leave this behind and just stop using it internally. App extensions might still be using it.

src/Controller/Component/FlashComponent.php
saeideng: now we can add a note for this that it will remove in 4.x

dereuromark:
You could still add a helping deprecation to

    protected $_session;

though

robertpustulka:
This should probably be targeting 3.8.0

markstory:
> This should probably be targeting 3.8.0

Why? It looks like a pretty safe change to me.

dereuromark (approved):

PR #12975, opened 2019-02-14
Entity::hasErrors() should return true only if field's errors array is non-empty.

### LINKED ISSUE #12974: hasErrors() returns true if field errors array is empty.
This is a (multiple allowed):

* [x] bug
* [ ] enhancement

* CakePHP Version: 3.7.3

### What you did

I set an empty error array:
```php
$entity->setError('foo', []);
```

### What happened
`hasErrors()` indicates that there are errors for this entity.
```php
$entity->hasErrors(); //true
$entity->getErrors('foo'); //empty array []
```

I expect that `hasErrors()` return false if all fields have empty errors arrays.

ADmad:
Fixes #12974

### DIFF
--- a/src/Datasource/EntityTrait.php
+++ b/src/Datasource/EntityTrait.php
@@ -15,6 +15,7 @@
 namespace Cake\Datasource;
 
 use Cake\Collection\Collection;
+use Cake\Utility\Hash;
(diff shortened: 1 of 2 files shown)


robertpustulka:
Thanks!

PR #12976, opened 2019-02-14
Update return type of log adapters.

### LINKED ISSUE #12971: 4.x: Log contract as per interface
I would like to point out that the Psr\Log\LoggerInterface has the following contract specified:
```
    /**
     * Logs with an arbitrary level.
     *
     * @param mixed  $level
     * @param string $message
     * @param array  $context
     *
     * @return void
     */
    public function log($level, $message, array $context = array());
```

Our FileLog and others:

    public function log($level, $message, array $context = []): bool

which technically violates the void return type.

We should fix this and remove all returning values here to comply with the interface.

ADmad:
Retun type now matches that of LoggerInterface::log().

Closes #12971

### DIFF
--- a/src/Log/Engine/FileLog.php
+++ b/src/Log/Engine/FileLog.php
@@ -116,9 +116,9 @@ class FileLog extends BaseLog
      *    See Cake\Log\Log::$_levels for list of possible levels.
      * @param string $message The message you want to log.
      * @param array $context Additional information about the logged message
-     * @return bool success of write.
+     * @return void
      */
-    public function log($level, $message, array $context = []): bool
+    public function log($level, $message, array $context = [])
     {
         $message = $this->_format($message, $context);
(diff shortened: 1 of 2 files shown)


dereuromark:
From what I saw some test classes / stubs that implement this interface might also still return
But if the tests are green that wouldnt matter much.

src/Log/Engine/FileLog.php
```diff
@@ -116,9 +116,9 @@ public function __construct(array $config = [])
      *    See Cake\Log\Log::$_levels for list of possible levels.
      * @param string $message The message you want to log.
      * @param array $context Additional information about the logged message
-     * @return bool success of write.
+     * @return void
      */
-    public function log($level, $message, array $context = []): bool
+    public function log($level, $message, array $context = [])
```
dereuromark: Can we add `:void`?
Or will that break due to the different interface?

ADmad: The interface doesn't have return type hint so best to not add our own.

PR #12980, opened 2019-02-16
Create controller instance if one is not provided to constructor.

ADmad:
This ensures static analyzers/IDE don't complain about calling methods on possibly null values.

Fixes error this thrown by psalm
```
INFO: PossiblyNullReference - src/Controller/Component/PaginatorComponent.php:196:55 - Cannot call method getRequest on possibly null value
        $request = $this->_registry->getController()->getRequest();
```

### DIFF
--- a/src/Controller/ComponentRegistry.php
+++ b/src/Controller/ComponentRegistry.php
@@ -44,17 +44,15 @@ class ComponentRegistry extends ObjectRegistry implements EventDispatcherInterfa
      */
     public function __construct(?Controller $controller = null)
     {
-        if ($controller) {
-            $this->setController($controller);
-        }
+        $this->setController($controller ?: new Controller());
     }
 
     /**
      * Get the controller associated with the collection.
      *
-     * @return \Cake\Controller\Controller|null Controller instance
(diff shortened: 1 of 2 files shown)


ADmad:
Alternatively the constructor argument could be made non-optional as tests updated to pass an instance. In actual usage the instance is always passed anyway.

dereuromark:
I was thinking the same yesterday. This makes total sense.

dereuromark:
Others like Table

    public function getConnection(): ?Connection

have the same issue:

nullpointer

markstory:
I think we can make this a required dependency in 4.0. It will break plugin tests and should be included in the migration guide but this change has a simple fix.

ADmad:
Cool, I'll update the PR

ADmad:
`Controller`'s constructor also takes `ComponentRegistry` as an (optional) argument so we kinda have an chicken and egg situation. On second thought I am sure not making `Controller` a required constructor argument for `ComponentRegistry` is the right thing to do.

We can keep the constructor unchanged and instead make `getController()` method throw an exception if controller is not set.

markstory (approved):
Looks good to me.

PR #12982, opened 2019-02-16
Fix return type issues reported by psalm.

ADmad:


dereuromark:
The failing 7.1 tests I fixed in my PR and should now be in master.

> CurlTest::testSendLiveResponseCheck

ADmad:
Boo.. A bad merge seems to have added back the `get_class()` usages making phpcs unhappy.

ADmad:
Any feedback on this?

dereuromark:
The inline suppress are a bit annoying.
But they seem to be outweight by the benefits, the correct types in other places.
:+1:

ADmad:
The inline comments ensure error at that particular location only is ignored. Using the psalm config would ignore that error throughout the file and thus new ones could be missed.

Psalm has another feature which generates a baseline file with existing errors avoiding the need to add specific ignores. I'll look into using that one of these days.

PR #12984, opened 2019-02-16
Better getter usage for internal framework handling.

dereuromark:
Follow https://github.com/cakephp/cakephp/pull/12980

nullpointer

The whole framework code always uses getConnection() without checking for null:

    ->getConnection()->...()

as such it only seems logical to also provide more sane defaults here.

### DIFF
--- a/src/ORM/Table.php
+++ b/src/ORM/Table.php
@@ -464,7 +464,9 @@ class Table implements RepositoryInterface, EventListenerInterface, EventDispatc
     public function getConnection(): Connection
     {
         if (!$this->_connection) {
-            $this->_connection = ConnectionManager::get(static::defaultConnectionName());
+            /** @var \Cake\Database\Connection $connection */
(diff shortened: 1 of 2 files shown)


dereuromark:
As an alternative, we could also throw a runtime exception as this then is a fail in test case setup.

dereuromark:
Now that's interesting:
```
 ------ ----------------------------------------------------------------------- 
  Line   ORM/Table.php                                                          
 ------ ----------------------------------------------------------------------- 
  467    Property Cake\ORM\Table::$_connection (Cake\Database\Connection|null)  
         does not accept Cake\Datasource\ConnectionInterface.                   
  470    Method Cake\ORM\Table::getConnection() should return                   
         Cake\Database\Connection but returns                                   
         Cake\Datasource\ConnectionInterface.                                   
 ------ ----------------------------------------------------------------------- 
```

We need to start doing better interface contracting instead of concrete objects ;)

PR #12985, opened 2019-02-16
Simplify local phpstan usage.

dereuromark:
As long as we dont have those in require-dev directly, it is painful to work locally here.

This should make it simpler as you can now run locally

    composer stan-setup
    composer stan

etc
Travis uses the same wrappers to keep things DRY.


ADmad (approved):
Locally one should just use the phar files files for phpstan and psalm instead of wasting time installing them through composer.

PR #12986, opened 2019-02-17
Allow more data to be counted as empty

markstory:
date, and time fields should all allow `[]` as an empty value. This makes the behavior of the new methods compatible with `Validator::notEmpty()`'s definition of empty. This change also makes
`allowEmptyDate(..., false)` work as a complement to its other modes.

I ran across this issue while trying to build the `notEmptyDate()`, and `notEmptyDateTime()` methods. I was unable to get them to behave opposite of the `allowEmpty*()` methods and reject `[]`.

I'm targeting this for 3.7 as I think it is a mistake in the logic that creates inconsistent behavior with `allowEmpty()` and `notEmpty()`.

### DIFF
--- a/src/Validation/Validator.php
+++ b/src/Validation/Validator.php
@@ -2339,7 +2339,7 @@ class Validator implements ArrayAccess, IteratorAggregate, Countable
             return true;
         }
 
-        $arrayTypes = (self::EMPTY_ARRAY | self::EMPTY_DATE | self::EMPTY_TIME | self::EMPTY_FILE);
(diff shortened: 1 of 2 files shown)


src/Validation/Validator.php
```diff
@@ -2339,7 +2339,8 @@ protected function isEmpty($data, $flags)
             return true;
         }
 
-        if ($data === [] && ($flags & self::EMPTY_ARRAY)) {
+        $arrayTypes = (self::EMPTY_ARRAY | self::EMPTY_DATE | self::EMPTY_TIME | self::EMPTY_FILE);
```
ADmad: The braces are not necessary. 4.x ruleset will complain about it when merging :).

ADmad (approved):

PR #12987, opened 2019-02-17
Include ArrayEngine in DSN and docblock

CauanCabral:
Although ArrayEngine is included in CakePHP 3.7.x, there's no mention for it in the Cache source, neither support using DSN notation.

I'm also putting the engines list in alphabetical order.

### DIFF
--- a/src/Cache/Cache.php
+++ b/src/Cache/Cache.php
@@ -47,8 +47,8 @@ use RuntimeException;
  * There are 7 built-in caching engines:
  *
  * - `ApcuEngine` - Uses the APCu object cache, one of the fastest caching engines.
- * - `ArrayEngine` - Uses only memory to store all data, not actually a persistent engin.
- *    Can be useful in test or CLI enviroment.
+ * - `ArrayEngine` - Uses only memory to store all data, not actually a persistent engine.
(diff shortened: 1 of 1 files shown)


src/Cache/Cache.php
```diff
@@ -44,18 +44,20 @@
  * In general all Cache operations are supported by all cache engines.
  * However, Cache::increment() and Cache::decrement() are not supported by File caching.
  *
- * There are 6 built-in caching engines:
+ * There are 7 built-in caching engines:
  *
+ * - `ApcuEngine` - Uses the APCu object cache, one of the fastest caching engines.
+ * - `ArrayEngine` - Uses only memory to store all data, not actually a persistent engin.
```
garas:
```suggestion
 * - `ArrayEngine` - Uses only memory to store all data, not actually a persistent engine.
```

src/Cache/Cache.php
```diff
@@ -44,18 +44,20 @@
...
+ *    Can be useful in test or CLI enviroment.
```
garas:
```suggestion
 *    Can be useful in test or CLI environment.
```

PR #12989, opened 2019-02-17
Move test classes into their own files.

ADmad:
Refs #232


dereuromark:
only some cs issues left

PR #12990, opened 2019-02-17
3.next - Add notEmpty* validator methods

markstory:
A few folks have found the current `allowEmpty*` methods to be clunky to use when trying to port `notEmpty()` to the stricter methods. I've added the missing methods now to make migration easier.

I've kept the signatures of these methods consistent with `notEmpty()` instead of being consistent with the `allowEmpty*()` methods as I think having the message first is a better default than having the condition first as the condition is used less frequently and when it is a callback it is nice to have the callback at the end of the arguments instead of in the middle. With that said I'm open to changing the argument order if people feel consistency across all the new methods is better.

### DIFF
--- a/src/Validation/Validator.php
+++ b/src/Validation/Validator.php
@@ -809,8 +809,7 @@ class Validator implements ArrayAccess, IteratorAggregate, Countable
     protected function sortMessageAndWhen($first, $second, $method)
     {
         // Called with `$message, $when`. No order change necessary
-        if (
-            (
+        if ((
                 in_array($second, [true, false, 'create', 'update'], true) ||
                 is_callable($second)
             ) && (
@@ -847,6 +846,7 @@ class Validator implements ArrayAccess, IteratorAggregate, Countable
     public function allowEmptyString($field, $message = null, $when = true)
     {
         list($message, $when) = $this->sortMessageAndWhen($message, $when, __METHOD__);
+
         return $this->allowEmptyFor($field, self::EMPTY_STRING, $when, $message);
     }
 
--- a/src/Validation/Validator.php
+++ b/src/Validation/Validator.php
@@ -812,11 +812,11 @@ class Validator implements ArrayAccess, IteratorAggregate, Countable
      * Opposite to allowEmptyString()
      *
      * @param string $field The name of the field.
-     * @param string|null $message The message to show if the field is not
+     * @param string|null $message The message to show if the field is empty.
      * @param bool|string|callable $when Indicates when the field is not allowed
-     *   to be empty. Valid values are false (always), 'create', 'update'. If a
-     *   callable is passed then the field will allowed to be empty only when
-     *   the callback returns false.
+     *   to be empty. Valid values are false (never), 'create', 'update'. If a
+     *   callable is passed then the field will be required to be not empty when
+     *   the callback returns true.
      * @return $this
      * @see \Cake\Validation\Validator::allowEmptyString()
      * @since 3.8.0
@@ -853,10 +853,11 @@ class Validator implements ArrayAccess, IteratorAggregate, Countable
      * Opposite to allowEmptyArray()
      *
      * @param string $field The name of the field.
-     * @param bool|string|callable $when Indicates when the field is allowed to be empty
-     * @param string|null $message The message to show if the field is not
-     * Valid values are false (always), 'create', 'update'. If a callable is passed then
-     * the field will allowed to be empty only when the callback returns false.
+     * @param string|null $message The message to show if the field is empty.
+     * @param bool|string|callable $when Indicates when the field is not allowed
+     *   to be empty. Valid values are false (never), 'create', 'update'. If a
+     *   callable is passed then the field will be required to be not empty when
+     *   the callback returns true.
      * @return $this
      * @see \Cake\Validation\Validator::allowEmptyArray()
      * @since 3.8.0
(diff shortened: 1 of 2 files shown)


src/Validation/Validator.php
```diff
@@ -806,6 +806,28 @@ public function allowEmptyString($field, $when = true, $message = null)
...
+     * @param string|null $message The message to show if the field is empty.
+     * @param bool|string|callable $when Indicates when the field is not allowed
+     *   to be empty. Valid values are false (never), 'create', 'update'. If a
+     *   callable is passed then the field will be required to be not empty when
+     *   the callback returns true.
+     * @return $this
+     * @see \Cake\Validation\Validator::allowEmptyString()
+     * @since 3.8.0
+     */
+    public function notEmptyString($field, $message = null, $when = false)
```
dereuromark:
```
public function allowEmptyArray($field, $when = true, $message = null)
```

I am a bit concerned about the difference of order here on the public API: when/message vs message/when between the notEmpty and allow methods.

markstory:
@ADmad @dereuromark The following diff is an example of how we could support both method signatures and provide a deprecation warning. This would allow us to normalize on the `message, when` ordering for 4.0 and ship the new methods with the preferred argument order now.

```diff
diff --git a/src/Validation/Validator.php b/src/Validation/Validator.php
index 3e38187275..1a166450bc 100644
--- a/src/Validation/Validator.php
+++ b/src/Validation/Validator.php
@@ -713,13 +713,19 @@ class Validator implements ArrayAccess, IteratorAggregate, Countable
     }
 
     /**
-     * Indicate that a field can be empty.
+     * Low-level method to indicate that a field can be empty.
      *
-     * Using an array will let you provide the following keys:
+     * This method should generally not be used and instead you should
+     * use:
      *
-     * - `flags` individual flags for field
-     * - `when` individual when condition for field
-     * - `message` individual message for field
+     * - `allowEmptyString()`
+     * - `allowEmptyArray()`
+     * - `allowEmptyFile()`
+     * - `allowEmptyDate()`
+     * - `allowEmptyDatetime()`
+     * - `allowEmptyTime()`
+     *
+     * Should be used as their APIs are simpler to operate and read.
      *
      * You can also set flags, when and message for all passed fields, the individual
      * setting takes precedence over group settings.
@@ -787,22 +793,60 @@ class Validator implements ArrayAccess, IteratorAggregate, Countable
         return $this;
     }
 
+    /**
+     * Compatibility shim for the allowEmpty* methods that enable
+     * us to support both the `$when, $message` signature (deprecated)
+     * and the `$message, $when` format which is preferred.
+     *
+     * A deprecation warning will be emitted when a deprecated form
+     * is used.
+     *
+     * @param mixed $first The message or when to be sorted.
+     * @param mixed $second The message or when to be sorted.
+     * @param string $method The called method
+     * @return array A list of [$message, $when]
+     */
+    protected function sortMessageAndWhen($first, $second, $method)
+    {
+        // Called with `$message, $when`. No order change necessary
+        if (
+            (
+                in_array($second, [true, false, 'create', 'update'], true) ||
+                is_callable($second)
+            ) && (
+                is_string($first) || $first === null
+            )
+        ) {
+            return [$first, $second];
+        }
+        deprecationWarning(
+            "You are using a deprecated argument order for ${method}. " .
+            "You should reverse the order of your `when` and `message` arguments " .
+            "so that they are `message, when`."
+        );
+
+        // Called with `$when, $message`. Reverse the
+        // order to match the expected return value.
+        return [$second, $first];
+    }
+
     /**
      * Allows a field to be an empty string.
      *
      * This method is equivalent to calling allowEmptyFor() with EMPTY_STRING flag.
      *
      * @param string $field The name of the field.
+     * @param string|null $message The message to show if the field is not
      * @param bool|string|callable $when Indicates when the field is allowed to be empty
      * Valid values are true, false, 'create', 'update'. If a callable is passed then
      * the field will allowed to be empty only when the callback returns true.
-     * @param string|null $message The message to show if the field is not
      * @return $this
      * @since 3.7.0
      * @see \Cake\Validation\Validator::allowEmptyFor() For detail usage
      */
-    public function allowEmptyString($field, $when = true, $message = null)
+    public function allowEmptyString($field, $message = null, $when = true)
     {
+        list($message, $when) = $this->sortMessageAndWhen($message, $when, __METHOD__);
         return $this->allowEmptyFor($field, self::EMPTY_STRING, $when, $message);
     }
```

I can either put add this change to this pull request or change the parameter order on the new methods to match the existing ones. I personally prefer the deprecation approach as we could leave the deprecations in place for 4.0 as well.

ADmad:
@markstory Adding the deprecations in this PR itself for 3.8 sounds good.

markstory:
Should be good for review now.

ADmad (approved):

src/Validation/Validator.php
```diff
@@ -813,95 +879,221 @@ public function allowEmptyString($field, $when = true, $message = null)
...
+     * @param string|null $message The message to show if the field is empty.
+     * @param bool|string|callable $when Indicates when the field is not allowed
+     *   to be empty. Valid values are false (never), 'create', 'update'. If a
+     *   callable is passed then the field will be required to be not empty when
+     *   the callback returns true.
+     * @return $this
+     * @see \Cake\Validation\Validator::allowEmptyArray()
+     * @since 3.8.0
+     */
+    public function notEmptyArray($field, $message = null, $when = false)
```
saeideng: could we have `$when = true`  by default ?
now callback can be return `true` ( as `false` after calling `invertWhenClause()`) ,
and this can be make confusing 
```php
        } elseif (is_callable($when)) {
            return function ($context) use ($when) {
                return !$when($context);
            };
        }
```
my suggest : 
`create` : can not be empty `when==create` // `update` is allowed
`update` : can not be empty `when==update` // `create` is allowed
`true` : can not be empty in both `create` + `update` (never)  
`false` : can be empty in both `create` + `update` // negates the behavior like `allowEmpty*()`
with this change return of `callback` and and method argument will same ( `true` as is not allowed )

this change can be hard to refactoring from `notEmpty()` to `notEmpty*()`  methods

markstory: The condition inversion was part of how notEmpty() worked as well. That is why I made these methods default to false as well.

The inversion is required because when rules are evaluated, the validator needs to know if a field can be empty or not  as these methods and the allowEmpty* methods change the same internal state.

andrii-pukhalevych: @markstory `$when = false` by default in `notEmpty*` medothds is awful behavior confusing for years.
Is it possible to change to be `$when = true` by default and return `!$when` in the last line of 
`invertWhenClause()`?

At least do this in new version release.

dereuromark: Please do not comment on old PRs, open a new ticket or ask in the public channel if you have a question about future changes in the core.

PR #12991, opened 2019-02-17
fix FlashComponent Session getter 

saeideng:
Refs : https://github.com/cakephp/cakephp/pull/12970

### DIFF
--- a/src/Controller/Component/FlashComponent.php
+++ b/src/Controller/Component/FlashComponent.php
@@ -150,11 +150,14 @@ class FlashComponent extends Component
      *
      * @param string $name Element name to use.
      * @param array $args Parameters to pass when calling `FlashComponent::set()`.
-     * @return void
+     * @return \Cake\Http\Session|void
      * @throws \Cake\Http\Exception\InternalErrorException If missing the flash message.
      */
     public function __call($name, $args)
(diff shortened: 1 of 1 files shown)


src/Controller/Component/FlashComponent.php
```diff
@@ -150,11 +150,14 @@ public function set($message, array $options = [])
      *
      * @param string $name Element name to use.
      * @param array $args Parameters to pass when calling `FlashComponent::set()`.
-     * @return void
+     * @return \Cake\Http\Session|void
```
dereuromark: Then you need to use `|null` IMO
void cannot coexist with anything else as it disallows return value use/checking completely.
Which clearly is not the case when you start allowing this.

dereuromark:
Why not just reverting the visibility?
https://github.com/cakephp/cakephp/commit/7d36ab355b9fc6473a64aee25dd2a1768efbd220#diff-67cec2192f1efe8681e2108d254550abR39

    public ...

This way, together with the deprecation notice, all is still clear IMO - without compromising existing code or needing this method hack here.

dereuromark:
Or we just do this: https://github.com/cakephp/cakephp/pull/12992

PR #12992, opened 2019-02-17
Fix visibility for session method for outside use

dereuromark:
Alternative to https://github.com/cakephp/cakephp/pull/12991

Follow up on the support of outside usage after https://github.com/cakephp/cakephp/pull/12970

Makes this work again:

    $this->Flash->success(__('The feature has been saved.'));
    $session = $this->Flash->getSession();

Not sure though if we need this at all.


dereuromark:
I actually think people shouldnt be calling the session from the outside on that component.
Instead

    $this->getSession();

from the controller itself should be done.

PR #12995, opened 2019-02-18
4.x cacheengine

ADmad:


src/Cache/CacheEngine.php
```diff
@@ -94,20 +98,25 @@ protected function ensureValidKey($key)
...
-     * @throws \Cake\Cache\InvalidArgumentException When the keys are not valid.
+     * @throws \Cake\Cache\InvalidArgumentException
      */
-    protected function ensureValidKeys($keys)
+    protected function ensureValidType($iterable, string $check = self::CHECK_VALUE)
     {
-        if (!is_array($keys) && !($keys instanceof \Traversable)) {
-            throw new InvalidArgumentException('A cache key set must be either an array or a Traversable.');
+        if (!is_iterable($iterable)) {
+            throw new InvalidArgumentException('A cache set / keys set must be either an array or a Traversable.');
```
markstory: This error message is confusing to me. Can we use `$check` to make it easier for the user to know which mode failed?

tests/TestCase/Cache/Engine/ArrayEngineTest.php
```diff
@@ -279,4 +281,62 @@ public function testAdd()
...
+     */
+    public function testWriteMany()
+    {
+        $data = new \ArrayObject([
+            'a' => 1,
+            'b' => 'foo',
+        ]);
+
+        $result = Cache::writeMany($data, 'array');
+        $this->assertTrue($result);
```
markstory: Should you also check that the object can be fetched out of the cache?

tests/TestCase/Cache/Engine/ArrayEngineTest.php
```diff
@@ -279,4 +281,62 @@ public function testAdd()
         $result = Cache::add('test_add_key', 'test data 2', 'array');
         $this->assertFalse($result);
     }
+
+    /**
+     * Test writeMany() with Traversable
+     *
+     * @return void
+     */
+    public function testWriteMany()
```
markstory:
```suggestion
    public function testWriteManyTraversable()
```

ADmad:
@markstory updates done as per feedback.

Issue #13001, opened 2019-02-20
RedisEngine->__destruct() undefined index notice error


benjaminstout:
This is a:

* [x] bug

* CakePHP Version: 664af068516402585564a57dff8f722c05649bea
* Platform and Target: CakePHP 2.10.15 running in PHP-FPM 7.2 docker container.

### What you did
`app/cake test`

### What happened
Received the usage output, as expected, followed by:
```
Notice Error: Undefined index: persistent in [app/Vendor/cakephp/cakephp/lib/Cake/Cache/Engine/RedisEngine.php, line 231]

PHP Fatal error:  Uncaught Error: Call to a member function close() on null in app/Vendor/cakephp/cakephp/lib/Cake/Cache/Engine/RedisEngine.php:232
Stack trace:
#0 [internal function]: RedisEngine->__destruct()
#1 {main}
  thrown in app/Vendor/cakephp/cakephp/lib/Cake/Cache/Engine/RedisEngine.php on line 232
PHP Stack trace:
PHP   1. {main}() app/Console/cake.php:0
PHP   2. ShellDispatcher::run() app/Console/cake.php:35
PHP   3. ShellDispatcher->_stop() app/Vendor/cakephp/cakephp/lib/Cake/Console/ShellDispatcher.php:66

Fatal error: Uncaught Error: Call to a member function close() on null in app/Vendor/cakephp/cakephp/lib/Cake/Cache/Engine/RedisEngine.php on line 232

Call Stack:
    0.0001     387280   1. {main}() app/Console/cake.php:0
    0.0010     458280   2. ShellDispatcher::run() app/Console/cake.php:35
    0.0338    4480824   3. ShellDispatcher->_stop() app/Vendor/cakephp/cakephp/lib/Cake/Console/ShellDispatcher.php:66

Error: Call to a member function close() on null in app/Vendor/cakephp/cakephp/lib/Cake/Cache/Engine/RedisEngine.php on line 232

Call Stack:
    0.0001     387280   1. {main}() app/Console/cake.php:0
    0.0010     458280   2. ShellDispatcher::run() app/Console/cake.php:35
    0.0338    4480824   3. ShellDispatcher->_stop() app/Vendor/cakephp/cakephp/lib/Cake/Console/ShellDispatcher.php:66
    0.0339    4481232   4. RedisEngine->__destruct() app/Vendor/cakephp/cakephp/lib/Cake/Cache/Engine/RedisEngine.php:0
```

### What you expected to happen
[...]

PR #13002, opened 2019-02-20
Wrap index reference in empty() check - closes cakephp#13001

stoutput:
Avoid direct array index reference without checking isset() first – thus we can wrap in empty() check instead. Closes issue #13001.

### DIFF
--- a/lib/Cake/Cache/Engine/RedisEngine.php
+++ b/lib/Cake/Cache/Engine/RedisEngine.php
@@ -228,7 +228,7 @@ class RedisEngine extends CacheEngine {
  * Disconnects from the redis server
  */
 	public function __destruct() {
-		if (empty($this->settings['persistent']) && !is_null($this->_Redis)) {
+		if (empty($this->settings['persistent']) && $this->_Redis !== null) {
 			$this->_Redis->close();
 		}
 	}
--- a/lib/Cake/Cache/Engine/RedisEngine.php
+++ b/lib/Cake/Cache/Engine/RedisEngine.php
@@ -228,7 +228,7 @@ class RedisEngine extends CacheEngine {
  * Disconnects from the redis server
  */
 	public function __destruct() {
-		if (!$this->settings['persistent']) {
+		if (empty($this->settings['persistent'])) {
 			$this->_Redis->close();
 		}
 	}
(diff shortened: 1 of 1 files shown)


lib/Cake/Cache/Engine/RedisEngine.php
```diff
@@ -228,7 +228,7 @@ public function clearGroup($group) {
  * Disconnects from the redis server
  */
 	public function __destruct() {
-		if (!$this->settings['persistent']) {
+		if (empty($this->settings['persistent'])) {
```
benjaminstout: Perhaps we should also add:
```suggestion
		if (empty($this->settings['persistent']) && !is_null($this->_Redis)) {
```

To prevent an attempt at closing an uninitialized redis instance, as documented in the PHP Fatal Error in issue #13001.

markstory: Sounds like a good idea to me.

benjaminstout: Done!

lib/Cake/Cache/Engine/RedisEngine.php
```diff
@@ -228,7 +228,7 @@ public function clearGroup($group) {
  * Disconnects from the redis server
  */
 	public function __destruct() {
-		if (!$this->settings['persistent']) {
+		if (empty($this->settings['persistent']) && !is_null($this->_Redis)) {
```
markstory:
```suggestion
		if (empty($this->settings['persistent']) && $this->_Redis !== null) {
```

Using !== is our idiom as it doesn't use a function call.

benjaminstout: Makes sense. Done!

benjaminstout:
Here's the same line from CakePHP 3: https://github.com/cakephp/cakephp/blob/master/src/Cache/Engine/RedisEngine.php#L306

which uses `$this->_Redis instanceof Redis` instead of `$this->_Redis !== null`

dereuromark:
It makes sense to keep 2.x in sync with 3.x.
Even though technically any other object wouldnt be allowed - so it is kind of overkill and not necessary to check the type.
But thats fine here.

PR #13003, opened 2019-02-20
4.x connection interface in favor #12094

lpj145:
In favor #12094 update interface.


lpj145:
i follow https://book.cakephp.org/3.0/en/contributing/code.html and not know whats go wrong...

Issue #6571, opened 2015-05-16
Database/Type/DateTimeType not respecting DB and PHP timezones.


jhawkwind:
If in your configuration you specify one timezone for the DB settings, and one timezone for the PHP settings, the time value that gets passed from the DB server through the client, then through PHP will simply pass text over, ignoring that all three time zones could be different and need conversion of the actual time value.

Let me provide a sample code first as it is easier to illustrate this way. Currently the code looks like this and just passes the value into a DateTime object instance:

``` php
    /**
     * Convert strings into DateTime instances.
     *
     * @param string $value The value to convert.
     * @param Driver $driver The driver instance to convert with.
     * @return \Cake\I18n\Time|DateTime
     */
    public function toPHP($value, Driver $driver)
    {
        if ($value === null || strpos($value, '0000-00-00') === 0) {
            return null;
        }

        if (strpos($value, '.') !== false) {
            list($value) = explode('.', $value);
        }

        $instance = clone $this->_datetimeInstance;
        return $instance->modify($value);
    }
```

Recommended changes are like this to take into different timezone settings into account:

``` php
    /**
     * Convert strings into DateTime instances.
     *
     * @param string $value The value to convert.
     * @param Driver $driver The driver instance to convert with.
     * @return \Cake\I18n\Time|DateTime
     */
    public function toPHP($value, Driver $driver)
    {
        if ($value === null || strpos($value, '0000-00-00') === 0) {
            return null;
        }

        if (strpos($value, '.') !== false) {
            list($value) = explode('.', $value);
        }

        $instance = new static::$dateTimeClass($value,$this->_DBTimeZone);
        $instance->modify($value);
        $instance->setTimeZone($this->_PHPTimeZone);
        return $instance;
    }
```

The short version is that, this will take into account the timezone settings you may have set in the application configuration for the DB and recognize that the value passed from the DB belong to a certain timezone that is different from PHP.

So let's say the DB client is configured for `America/Los_Angeles`; it will figure out what timezone you have set as default in PHP, let's say `America/New_York`; then it will take the DB value it recieves, advance the time 3 hours, and then present it for use for the PHP code that expects all times to be in **Eastern Time**.

[...]

PR #13005, opened 2019-02-20
Take db timezone into account when creating datetime objects.

ADmad:
Refs #6571.

### DIFF
--- a/tests/TestCase/Database/Type/DateTimeTypeTest.php
+++ b/tests/TestCase/Database/Type/DateTimeTypeTest.php
@@ -125,6 +125,20 @@ class DateTimeTypeTest extends TestCase
             $expected,
             $this->type->manyToPHP($values, array_keys($values), $this->driver)
         );
+
+        $this->type->setTimezone('Asia/Kolkata'); // UTC+5:30
+        $values = [
+            'a' => null,
+            'b' => '2001-01-04 12:13:14',
+        ];
+        $expected = [
+            'a' => null,
+            'b' => new Time('2001-01-04 06:43:14'),
+        ];
+        $this->assertEquals(
+            $expected,
+            $this->type->manyToPHP($values, array_keys($values), $this->driver)
+        );
     }
 
     /**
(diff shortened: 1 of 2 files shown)


ADmad:
@markstory @lorenzo @dereuromark Any thoughts on this?

Since the timezone is taken into account when converting PHP object to datetime string for db, it should be taken into account when creating PHP object back from db datetime string too.

tests/TestCase/Database/Type/DateTimeTypeTest.php
```diff
@@ -94,6 +94,16 @@ public function testToPHPString()
         $this->assertEquals('12', $result->format('H'));
         $this->assertEquals('13', $result->format('i'));
         $this->assertEquals('14', $result->format('s'));
+
+        $this->type->setTimezone('Asia/Kolkata'); // UTC+5:30
```
markstory: Why is the resulting datetime' hour 6 if the offset is 5:30?

ADmad: Since this timezone it UTC+5:30 you need to go back 5:30 hours to get UTC time 

Going back 5:30 hours from 12:00 gives you 6:30.

dereuromark:
Great first step towards timezone handling then - there are still a few things left, like https://github.com/cakephp/cakephp/pull/11517 right?

ADmad:
Yes dealing with conversion to/from user's timezone which #11517 relates to remains. But keeping that in mind when updating `DateTimeWidget` to generate native HTML input, I added the possibility to provide a `timezone` key in data array. If provided the timezone for datetime object will be set to that timezone before generating datetime string to populate the input's value.

tests/TestCase/Database/Type/DateTimeTypeTest.php
markstory: Derp. My bad.

markstory: Should we have a similar test for the `manyToPhp()` method?

ADmad: Done

ADmad:
So is this good to merge?  The phpcs and stan failure are unrelated.

PR #13006, opened 2019-02-20
Fix xdebug removal for travis runs.

dereuromark:
xdebug should also be removed from 7.3 - only exception is the codecoverage run, whatever php version

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,7 +47,7 @@ matrix:
 
 before_install:
   - echo cakephp version && tail -1 VERSION.txt
-  - if [[ $CODECOVERAGE != 1 ]]; then phpenv config-rm xdebug.ini; fi
+  - phpenv config-rm xdebug.ini
 
   - if [ $DB = 'mysql' ]; then mysql -u root -e 'CREATE DATABASE cakephp_test;'; fi
   - if [ $DB = 'mysql' ]; then mysql -u root -e 'CREATE DATABASE cakephp_test2;'; fi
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -47,7 +47,7 @@ matrix:
 
 before_install:
   - echo cakephp version && tail -1 VERSION.txt
-  - if [[ ${TRAVIS_PHP_VERSION} != "7.3" ]]; then phpenv config-rm xdebug.ini; fi
+  - if [[ $CODECOVERAGE != 1 ]]; then phpenv config-rm xdebug.ini; fi
```
ADmad: The PHP 7.3 was done because 7.3 compatible xdebug didn't exist, does it now? Also for code codecoverage we use `phpdbg` and don't need xdebug for that job either.

So the line should be just changed to `phpenv config-rm xdebug.ini` is PHP 7.3 check isn't required.

dereuromark: My findings say it works this way there too.

ADmad: Where's "there"? :)

dereuromark: Travis green shows no check necessary. That is better than doing check and missing to remove it.

ADmad: I said to remove remove check and just use `phpenv config-rm xdebug.ini` :slightly_smiling_face:.

dereuromark: We can sure try :)

dereuromark: But wait.. coverage runs on 7.1 and that needs it. So no, I think the current way is correct and mergable.

ADmad: Repeating what I said above:

> Also for code codecoverage we use phpdbg and don't need xdebug for that job either.

dereuromark: ;)

PR #13008, opened 2019-02-21
Serialized response

mirko-pagliai:
For a plugin I'm developing, I tried to cache a `Cake\Http\Client\Response` instance (suppose that you are using `Cake\Http\Client` to make multiple requests and it is possible that the same request is made multiple times).

I realized that, by going to read the object cached, the response body was always missing.
This happens because the body is a `Stream` instance.

Consider this PR as a draft for the following reasons:
1) I'm not sure this is the best method;
2) `Cake\Http\Response` and `Cake\Http\Client\Response` classes have many equal methods,  besides what I added. Would not it be better to use a trait at this point?

Tell me what you think about. Thanks.


mirko-pagliai:
I forgot: the same applies to `\Cake\Http\Response`.
Precisely for this reason I was talking about a possible trait.

markstory:
What is the use case around serializing a server response? I can see the use case for a client response but not a server response.

src/Http/Client/Response.php
```diff
@@ -119,6 +120,20 @@ class Response extends Message implements ResponseInterface
...
+     * @var string
+     */
+    protected $_streamMode = 'wb+';
+
+    /**
+     * Stream target or resource object.
+     *
+     * @var string|resource
+     */
+    protected $_streamTarget = 'php://memory';
```
markstory: Why is the response needing to know about the stream properties?

src/Http/Client/Response.php
```diff
@@ -228,6 +253,32 @@ protected function _parseHeaders($headers)
...
+     *
+     * @param string $serialized The string representation of the object
+     * @return Cake\Http\Client\Response
+     */
+    public function unserialize($serialized)
+    {
+        list($response, $body) = unserialize($serialized);
+
+        $response->_createStream();
+        $response->stream->write((string)$body);
```
markstory: Couldn't we use a stream class that uses memory storage here? There is likely one in diactoros already.

src/Http/Client/Response.php
mirko-pagliai: > What is the use case around serializing a server response? I can see the use case for a client response but not a server response.

In fact I find it more useful for a `Cake\Http\Client\Response` instance (this is just what I would need), rather than for a `Cake\Http\Response` instance.

I said to make it possible even for a server response just because the code would be the same. I added the `$_streamTarget` property just for this reason: to make the two classes homogeneous, not because we really need them. Actually this is not necessary.

mirko-pagliai:
@markstory 
> What is the use case around serializing a server response? I can see the use case for a client response but not a server response.

In fact I find it more useful for a `Cake\Http\Client\Response` instance (this is just what I would need), rather than for a `Cake\Http\Response` instance.

I said to make it possible even for a server response just because the code would be the same. I added the `$_streamTarget` property just for this reason: to make the two classes homogeneous, not because we really need them. Actually this is not necessary.

src/Http/Client/Response.php
```diff
@@ -228,6 +253,32 @@ protected function _parseHeaders($headers)
...
+
+        return $response;
```
othercorey: I don't believe `unserialize()` supports a return type.

https://www.php.net/manual/en/class.serializable.php

src/Http/Client/Response.php
```diff
@@ -165,6 +180,16 @@ public function __construct($headers = [], $body = '')
...
     }
 
+    /**
+     * Creates the stream object.
+     *
+     * @return void
+     */
+    protected function _createStream()
+    {
+        $this->stream = new Stream($this->_streamTarget, $this->_streamMode);
```
othercorey: This needs to work the same as __construct().

src/Http/Client/Response.php
```diff
@@ -228,6 +253,32 @@ protected function _parseHeaders($headers)
...
     }
 
+    /**
+     * String representation of object
+     *
+     * @return string The string representation of the object
+     */
+    public function serialize()
+    {
+        return serialize([$this, (string)$this->getBody()]);
```
othercorey: Why is getBody() the only thing you're serializing?

PR #13009, opened 2019-02-21
Fix nested errors generation for missing layout.

ADmad:
Since MissingLayoutException now extends MissingTemplateException
explicit check is required for it when catching MissingTemplateException.

### DIFF
--- a/tests/TestCase/Error/ExceptionRendererTest.php
+++ b/tests/TestCase/Error/ExceptionRendererTest.php
@@ -764,6 +764,37 @@ class ExceptionRendererTest extends TestCase
     }
 
     /**
+     * Test that missing layout don't cause other fatal errors.
+     *
+     * @return void
+     */
+    public function testMissingLayoutRenderSafe()
+    {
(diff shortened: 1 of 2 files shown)


tests/TestCase/Error/ExceptionRendererTest.php
```diff
@@ -763,6 +763,37 @@ function (EventInterface $event) {
...
+            }
+        );
+        $controller->setRequest(new ServerRequest());
+        $ExceptionRenderer->setController($controller);
+
+        $response = $ExceptionRenderer->render();
+        $this->assertEquals('text/html', $response->getType());
+        $this->assertContains('Not Found', (string)$response->getBody());
+        $this->assertTrue($this->called, 'Listener added was not triggered.');
+        $this->assertEquals('', $controller->viewBuilder()->getLayoutPath());
```
dereuromark: We should use assertSame() here, as falsely values would also false positive.

dereuromark (approved):

PR #9803, opened 2016-11-26
Revert #9577

chinpei215:
The change in #9577 breaks bake. Because the removed variable `$boot` is used apparently. See https://github.com/cakephp/cakephp/blob/57bc0f9c457974378b708c79752a4ee1de26c389/lib/Cake/bootstrap.php#L432

### CakepPHP <= 2.9.0

```
C:\var\www\localhost\webroot\demo>Vendor\cakephp\cakephp\app\Console\cake bake project .


Welcome to CakePHP v2.9.0 Console
---------------------------------------------------------------
App : demo
Path: C:\var\www\localhost\webroot\demo\
---------------------------------------------------------------
Skel Directory: C:\var\www\localhost\webroot\demo\Vendor\cakephp\cakephp\lib\Cake\Console\Templates\skel
Will be copied to: .
---------------------------------------------------------------
Look okay? (y/n/q)
```

### CakePHP >= 2.9.1
```
C:\var\www\localhost\webroot\demo>Vendor\cakephp\cakephp\app\Console\cake bake project .

PHP Warning:  include(C:\var\www\localhost\webroot\demo\Config\core.php): failed to open stream: No such file or directory in C:\var\www\localhost\webroot\demo\Vendor\cakephp\cakephp\lib\Cake\Core\Configure.php on line 72
PHP Stack trace:
PHP   1. {main}() C:\var\www\localhost\webroot\demo\Vendor\cakephp\cakephp\app\Console\cake.php:0
PHP   2. ShellDispatcher::run() C:\var\www\localhost\webroot\demo\Vendor\cakephp\cakephp\app\Console\cake.php:47
PHP   3. ShellDispatcher->__construct() C:\var\www\localhost\webroot\demo\Vendor\cakephp\cakephp\lib\Cake\Console\ShellDispatcher.php:65
PHP   4. ShellDispatcher->_initEnvironment() C:\var\www\localhost\webroot\demo\Vendor\cakephp\cakephp\lib\Cake\Console\ShellDispatcher.php:54
PHP   5. ShellDispatcher->_bootstrap() C:\var\www\localhost\webroot\demo\Vendor\cakephp\cakephp\lib\Cake\Console\ShellDispatcher.php:100
PHP   6. require() C:\var\www\localhost\webroot\demo\Vendor\cakephp\cakephp\lib\Cake\Console\ShellDispatcher.php:138
[...]

PR #13010, opened 2019-02-21
Use CONFIG constant

bancer:


lib/Cake/Console/ShellDispatcher.php
```diff
@@ -140,8 +140,11 @@ protected function _bootstrap() {
 			define('TMP', CAKE_CORE_INCLUDE_PATH . DS . 'Cake' . DS . 'Console' . DS . 'Templates' . DS . 'skel' . DS . 'tmp' . DS);
 		}
 
+		if (!defined('CONFIG')) {
+			define('CONFIG', ROOT . DS . APP_DIR . DS . 'Config' . DS);
+		}
 		// $boot is used by Cake/bootstrap.php file
-		$boot = file_exists(ROOT . DS . APP_DIR . DS . 'Config' . DS . 'bootstrap.php');
+		$boot = file_exists(CONFIG . 'bootstrap.php');
```
bancer: I just saw https://github.com/cakephp/cakephp/pull/9803. Perhaps a better solution would be to move `$boot` variable to the file where it is actually used, i.e. to Cake/bootstrap.php. @chinpei215, @dereuromark, what would you say?

PR #13013, opened 2019-02-21
Make AssetMiddleware extend PSR 15's MiddlewareInterface.

ADmad:
I missed updating this one earlier :slightly_smiling_face:.


src/Routing/Middleware/AssetMiddleware.php
```diff
@@ -54,30 +57,29 @@ public function __construct(array $options = [])
...
-            return $next($request, $response);
+            return $handler->handle($request);
         }
 
         $file = new SplFileInfo($assetFile);
         $modifiedTime = $file->getMTime();
         if ($this->isNotModified($request, $file)) {
-            return $response
+            return (new Response())
                 ->withStringBody('')
```
garas: With fresh `Response`, you are now sure body is empty, so resetting body unnecessary.

PR #13016, opened 2019-02-23
Run phpcs tests parallel.

ADmad:
Testing locally I got no time improvement beyond value of `4` on 2 cores. So values of `16` should suffice to use someone's 8 core CPU optimally.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -95,8 +95,8 @@
             "@cs-check",
             "@test"
         ],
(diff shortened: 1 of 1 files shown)


dereuromark:
Nice one
On travis here only 49->45 sec
But locally, with xdebug on, core cs-check went down from 3 minutes to 1 :)

PR #13017, opened 2019-02-23
Add Message class for email.

ADmad:
This PR adds separate `Message` class for email as proposed by @robertpustulka in https://github.com/cakephp/cakephp/pull/12810#pullrequestreview-184533973.

The transport classes need to be updated to now use instance of `Message` instead of `Email`. This would be a BC break but it shouldn't need much change in userland code as directly using transport classes wasn't practical in 3.x anyway. Currently tests do pass since call to non-existent method on `Email` are proxied to `Message` for BC.

Also I think the `Renderer` class I added earlier should be changed. Currently apart from rendering templates it also handles stitching together various parts of message like attachments, html message and text message with appropriate boundary. I think the stitching together of various parts belongs in `Message` class instead of `Renderer`.


src/Mailer/AbstractTransport.php
```diff
@@ -34,10 +34,10 @@ abstract class AbstractTransport
     /**
      * Send mail
      *
-     * @param \Cake\Mailer\Email $email Email instance.
+     * @param \Cake\Mailer\Message $message Email mesage.
      * @return array
      */
-    abstract public function send(Email $email): array;
+    abstract public function send(Message $message): array;
```
markstory: Won't this be a breaking change that we didn't have a deprecation for in 3.x?

This could impact any plugins or application transport implementations.

markstory (commented):
I wasn't able to review the entire change as I was on a tiny screen. I can take another look in a few days.

src/Mailer/AbstractTransport.php
ADmad: Yes custom transports would be affected.

A compromise would be removing the typehint and allowing method to accept both `Email` and `Message` instances.

ADmad: Actually apart for adding the type hint no change will be required in transport implementations of plugins and apps and `Message` has the same public methods as `Email` to get headers and message.

Add typehints is something that's going to be required in general for all classes updating from 3.x to 4.0.

markstory: Good point. Other transports will need to add *a* typehint either way. This change just changes the class name that ends up in the hint.

ADmad:
Anything else needs to be done here?

PR #13026, opened 2019-02-28
Offload phpcs and static analysis checks to Scrutinizer

ADmad:
- Moved phpcs and static analysis checks from travis to Scrutinizer so that overall time to run all checks is reduced. I had to make the travis' job matrix a more verbose to ensure all jobs for code coverage with each db type are run first.
- Coverage report is also now submitted to scrutinizer instead of codecov.
- The various other checks that were done by scrutinizer by default have been disabled. In all these years I don't think anyone has actually checked them and made updates as per it's recommendations.

### DIFF
--- a/.scrutinizer.yml
+++ b/.scrutinizer.yml
@@ -12,10 +12,7 @@ build:
     phpcs:
       tests:
         override:
-          - command: vendor/bin/phpcs --report=checkstyle src/ tests/ > phpcs-checks.xml
-            analysis:
-              file: phpcs-checks.xml
-              format: "php-cs-checkstyle"
+          - phpcs-run src/ tests/
 
     static-analysis:
       environment:
@@ -35,4 +32,5 @@ build:
           - vendor/bin/psalm --show-info=false
 
 build_failure_conditions:
-- 'project.metric_change("scrutinizer.test_coverage", < -0.01)'
+  - 'issues.label("coding-style").exists'
+  - 'project.metric_change("scrutinizer.test_coverage", < -0.01)'
--- a/.scrutinizer.yml
+++ b/.scrutinizer.yml
@@ -32,5 +32,5 @@ build:
           - vendor/bin/psalm --show-info=false
 
 build_failure_conditions:
-  - 'issues.label("coding-style").exists'
+  - 'issues.count > 0'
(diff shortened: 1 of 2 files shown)


dereuromark:
What is the advantage of this mixture? Besides a bit of time saved?
For me scrut always looks a bit less easy to use and buggy somewhat.

Why do we have to do coverage for all engines now? To have the edge cases also counted in?
Will that in itself then increase the runtime maybe?

ADmad:
> What is the advantage of this mixture? Besides a bit of time saved?

Time saved is the main advantage.

> For me scrut always looks a bit less easy to use and buggy somewhat.

With all it's default checks turned off it now only shows issues for CS failures and it's pretty clean.
e.g. https://scrutinizer-ci.com/g/ADmad/cakephp/inspections/743d6193-ab15-4883-8ea2-621a8a30460d/issues/

> Why do we have to do coverage for all engines now? To have the edge cases also counted in?
> Will that in itself then increase the runtime maybe?

There are some tests which are engine specific hence coverage checks need to be run for all engines to get proper coverage report. Earlier too coverage jobs were run for all engines on travis, before the matrix was tweaked recently to circumvent the bug in codecov's report uploader.

dereuromark:
Interesting that the progress cannot be seen from this PR
https://scrutinizer-ci.com/g/cakephp/cakephp/inspections/4a872190-60a0-4e2f-a212-06ce81fb6205
is not linked anywhere until completion I assume?

> With all it's default checks turned off ...

Yeah, that is definitely necessary with this tool, lots of noise and often even false positives reported.

ADmad:
It seems the checks are not being shown on this PR as scrutinizer checks were limited to master branch through it's config earlier. You can see them on the test PR on my fork https://github.com/ADmad/cakephp/pull/14.

ADmad:
Ah, someone with admin perms needs to the scrutinizer app for the repo.

markstory:
We can try scrutinizer again. In the past it was pretty inconsistent with when and what pulls it reported on. I'll approve the app so we can try it out on this repo before rolling it out to other repositories.

ADmad:
Merging this so we can see how Scrutinizer checks show up on new PRs. If we have problems I'll revert the changes to travis and disable scrutinizer checks again.

PR #13028, opened 2019-03-02
Try to avoid installing PHP each time on appveyor.

ADmad:


ADmad:
A test for console input keeps failing. Should we just skip that test for Windows?

markstory:
@ADmad I would skip it.

PR #13030, opened 2019-03-04
Fix errors reported by phpstan level 4.

ADmad:


ADmad:
Most of the remaining errors need to be ignored and generate pretty generic message. But since phpstan only allows ignoring using regex to match error message it could lead of actual errors being ignored too nor do I don't want to spend time filling up the ignore list in phpstan's config file.

src/Database/Schema/TableSchema.php
```diff
@@ -456,7 +456,7 @@ public function defaultValues(): array
      * {@inheritDoc}
      * @throws \Cake\Database\Exception
      */
-    public function addIndex(string $name, array $attrs): TableSchemaInterface
+    public function addIndex(string $name, $attrs): TableSchemaInterface
```
dereuromark: Should we instead remove the dead string option here?

src/I18n/Translator.php
```diff
@@ -58,7 +58,7 @@ public function translate($key, array $tokensValues = [])
             unset($tokensValues['_context']);
         }
 
-        if (!$tokensValues) {
+        if (empty($tokensValues)) {
```
dereuromark: We shouldnt do this.

ADmad: Why not? `$tokenValues` is type hinted as array so no other empty values can sneak in.

src/Database/Schema/TableSchema.php
ADmad: There are no tests for using string value for `$attrs` but code does have `is_string()` check.

We could remove passing string if other's agree.

src/I18n/Translator.php
dereuromark: Exactly,  thats why this not necessary and against conventions. 
Dont use cloaking where var is defined. If someone makes a typo here it will not throw notice.

src/Database/Schema/TableSchema.php
dereuromark: We can add it then to migration guide.

src/I18n/Translator.php
ADmad: If you make a typo then pslam's `UndefinedVariable` check kicks in, so it's a non-issue.

https://psalm.dev/r/00a573221e

dereuromark: Ok. still think its bad to use cloaking to please tools bad reporting, but fair enough then.

src/Console/ConsoleInput.php
```diff
@@ -65,7 +65,12 @@ public function read(): string
             return $line;
         }
 
-        return fgets($this->_input);
+        $line = fgets($this->_input);
```
markstory: Would `return (string) fgets($this->_input);` also appease the linter?

src/Core/functions.php
```diff
@@ -61,9 +61,6 @@ function h($text, bool $double = true, ?string $charset = null)
         static $defaultCharset = false;
         if ($defaultCharset === false) {
             $defaultCharset = mb_internal_encoding();
-            if ($defaultCharset === null) {
```
markstory: If there is no encoding set, this method returns null and I think we should fallback to `utf-8` still.

src/Utility/Security.php
```diff
@@ -263,31 +263,18 @@ public static function decrypt(string $cipher, string $key, ?string $hmacSalt =
...
-        }
-        $originalLength = mb_strlen($original, '8bit');
-        $compareLength = mb_strlen($compare, '8bit');
-        if ($originalLength !== $compareLength) {
-            return false;
-        }
-        $result = 0;
-        for ($i = 0; $i < $originalLength; $i++) {
-            $result |= (ord($original[$i]) ^ ord($compare[$i]));
-        }
```
markstory: :clap: yay

ADmad:
@markstory Updates done as per feedback.

ADmad:
Merging so I continue with more fixes.

PR #13034, opened 2019-03-07
Add information to deprecation message

davidyell:
I was struggling to find the correct method to replace this property, and as I was accessing at as a getter being told to check a setter method was confusing. Hopefully this small change will help others find the correct replacement methods.

### DIFF
--- a/src/View/ViewVarsTrait.php
+++ b/src/View/ViewVarsTrait.php
@@ -38,7 +38,7 @@ trait ViewVarsTrait
      * Variables for the view
      *
      * @var array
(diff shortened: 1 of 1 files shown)


src/View/ViewVarsTrait.php
```diff
@@ -38,7 +38,7 @@ trait ViewVarsTrait
      * Variables for the view
      *
      * @var array
-     * @deprecated 3.7.0 Use `$this->set()` instead.
+     * @deprecated 3.7.0 Use `$this->set()` instead, also see `\Cake\View\View::get()`.
```
ADmad: The correct alternate getter would be `$this->viewBuilder()->getVar()`.

PR #13047, opened 2019-03-13
Speed up ExtractTask by skipping files without translation functions

garas:
Speed up ExtractTask 1.5-3x times by skipping files without translation functions. Regular expressions are faster than parsing file.

With `--extract-core=yes`, time taken decreased 10 seconds to 3.7 seconds, because from 633 core src/ files, only 16 uses translation functions.
Time saved in apps depends how much files can be skipped.


ADmad (approved):

markstory:
Great work @garas :clap:

Issue #13046, opened 2019-03-13
2.x - deleting dependent HABTM records ignores conditions defined on the model


garethellis36:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 2.10.15
* Platform and Target: PHP 7.2, IIS, Windows

I'm trying to create a single table to house all `tag` relationships. So we have a `tags` table with fields `id` and `name`, and then `tagged_entities` table, which has fields `entity_type`, `entity_id`, `tag_id`. This way we can attach tags to multiple models in the application without having to have a join table for each relationship.

The relationship is created automatically in `AppModel::__construct()`:

```php
<?php
        $this->bindModel(
            [
                "hasAndBelongsToMany" => [
                    "Tag" => [
                        "joinTable" => Inflector::tableize(TaggedEntity::class),
                        "conditions" => [
                            TaggedEntity::class . ".entity_type" => $this->useTable,
                        ],
                        "foreignKey" => "entity_id",
                        "with" => TaggedEntity::class,
                    ],
                ],
            ],
            false
        );
```

The provided `conditions` enable the `contain` usage with `Model::find()` to work as expected.

However, when I call `Model::delete()` on a 'taggable' entity, it cascades the deletes to the join table and ignores the `conditions`.

The code responsible for this is `Model::_deleteLinks()`:

```php
<?php
	protected function _deleteLinks($id) {
		foreach ($this->hasAndBelongsToMany as $data) {
			list(, $joinModel) = pluginSplit($data['with']);
			$Model = $this->{$joinModel};
			$records = $Model->find('all', array(
				'conditions' => array($Model->escapeField($data['foreignKey']) => $id),
				'fields' => $Model->primaryKey,
				'recursive' => -1,
				'callbacks' => false
			));

			if (!empty($records)) {
				foreach ($records as $record) {
					$Model->delete($record[$Model->alias][$Model->primaryKey]);
				}
			}
		}
	}
```

It is a fairly straightforward fix to address this, and I do not think this would be breaking BC. I am happy to submit a PR.

Revised version would look something like:

```php
<?php
	protected function _deleteLinks($id) {
		foreach ($this->hasAndBelongsToMany as $data) {

            $defaultConditions = isset($data["conditions"])
                ? ($data["conditions"] ?: array())
                : array();
		    
			list(, $joinModel) = pluginSplit($data['with']);
			$Model = $this->{$joinModel};
[...]

PR #13048, opened 2019-03-14
Separates derivation of conditions for cascaded deletion of HABTM records to a protected method #13046

garethellis36:
Allows overriding of the default behaviour without having to override `_deleteLinks()` entirely, as discussed in #13046


markstory:
Thanks!

PR #13050, opened 2019-03-14
Fix up extract() doc.

dereuromark:
Fixed docblocks for extract() for IDE and tooling, also found some old relics of dead code.


dereuromark:
Looks like a follow up PR is needed :)
```
Ignored error pattern #Variable \$plural in isset\(\) is never defined# was not matched in reported errors.   
  Ignored error pattern #Variable \$context in isset\(\) is never defined# was not matched in reported errors.  
```

dereuromark:
Done in https://github.com/cakephp/cakephp/commit/475481d8469f9763fa14b7beb469c043177c3df9

PR #13051, opened 2019-03-14
Fix up include of config via defaulting.

dereuromark:
I would like to see if this is a better/cleaner approach and also makes phpstan error go away.

### DIFF
--- a/src/Core/Configure/Engine/PhpConfig.php
+++ b/src/Core/Configure/Engine/PhpConfig.php
@@ -85,12 +85,14 @@ class PhpConfig implements ConfigEngineInterface
     {
         $file = $this->_getFilePath($key, true);
 
+        $config = null;
+
         $return = include $file;
         if (is_array($return)) {
             return $return;
         }
 
-        if (!isset($config)) {
+        if ($config === null) {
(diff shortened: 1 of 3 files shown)


src/Core/Configure/Engine/PhpConfig.php
```diff
@@ -85,12 +85,14 @@ public function read($key)
...
+        $config = null;
+
         $return = include $file;
         if (is_array($return)) {
             return $return;
         }
 
-        if (!isset($config)) {
+        if ($config === null) {
             throw new Exception(sprintf('Config file "%s" did not return an array', $key . '.php'));
```
dereuromark: The message seems weird here, do we need to maybe use a different once?

dereuromark:
I am not sure where this follow up issue came from:
```
------ ---------------------------- 
  Line   Http/Client/Auth/Oauth.php  
 ------ ---------------------------- 
  87     Undefined variable: $value  
 ------ ---------------------------- 
```
Fixed it, too

dereuromark:
Ah, they literally just released https://github.com/phpstan/phpstan/releases/tag/0.11.4 which found this.

PR #13052, opened 2019-03-14
 Improve generated .pot references compatibility with PO editors

garas:
Any PO editor I tried does not support references like `File.php:23;56`, each line must be listed separately to support _Go to source_ function.

```
#: ./vendor/cakephp/cakephp/src/I18n/RelativeTimeFormatter.php:47;126;316
msgid "{0} year"
```
 will become

```
#: ./vendor/cakephp/cakephp/src/I18n/RelativeTimeFormatter.php:47
#: ./vendor/cakephp/cakephp/src/I18n/RelativeTimeFormatter.php:126
#: ./vendor/cakephp/cakephp/src/I18n/RelativeTimeFormatter.php:316
msgid "{0} year"
```

Generated file size will increase a bit, but it will be possible to look at message with context in source code.

### DIFF
--- a/src/Shell/Task/ExtractTask.php
+++ b/src/Shell/Task/ExtractTask.php
@@ -519,16 +519,18 @@ class ExtractTask extends Shell
                 foreach ($contexts as $context => $details) {
                     $plural = $details['msgid_plural'];
                     $files = $details['references'];
-                    $occurrences = [];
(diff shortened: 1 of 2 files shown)


markstory (commented):
Looks good to me one you have a passing build.

garas:
`phpstan` is failing at completely differently location. Need to fix that first

PR #13055, opened 2019-03-15
4.x static analysis

ADmad:


src/Auth/DefaultPasswordHasher.php
```diff
@@ -47,6 +47,7 @@ class DefaultPasswordHasher extends AbstractPasswordHasher
      */
     public function hash(string $password)
```
dereuromark: Can we `: string` typehint here? Or at least string|null then?

src/Database/Statement/StatementDecorator.php
```diff
@@ -354,6 +355,7 @@ public function lastInsertId(?string $table = null, ?string $column = null)
      */
     public function getInnerStatement()
     {
+        /** @psalm-suppress NullableReturnStatement */
```
dereuromark: I feel like this would be cleaner if we catch this and throw meaninful exception?

ADmad: I don't know if the property can actually be null in practice. Some of the db layer internals are not easy to understand so I just chose the safe option :).

src/Auth/DefaultPasswordHasher.php
ADmad: AuthComponents and it's related classes are effectively deprecated so I prefer doing the bare min. to keeps the tools happy.

dereuromark: No more hasher? OK, didnt know that.

ADmad: The hashers are included in the new authentication plugin.

dereuromark (approved):

src/Database/Statement/StatementDecorator.php
```diff
@@ -64,7 +64,7 @@ class StatementDecorator implements StatementInterface, Countable, IteratorAggre
      *  such as PDOStatement.
      * @param \Cake\Database\Driver|null $driver Driver instance
      */
-    public function __construct($statement = null, $driver = null)
+    public function __construct($statement, $driver = null)
```
dereuromark: The doc block states |null still.

dereuromark: Ah you want it to allow null, but the argument has to be present.

ADmad: Actually i don't want to change anything. I was just trying something and forgot to revert :stuck_out_tongue:.

src/Error/Debugger.php
```diff
@@ -914,6 +914,7 @@ public static function printVar($var, array $location = [], ?bool $showHtml = nu
                 $lineInfo = sprintf('%s (line %s)', $file, $line);
             }
         }
+        /** @psalm-suppress RedundantCondition */
```
muglug: Sorry about this - it'll be fixed in the next release

ADmad: No problem. BTW can psalm report unneeded ignores?

src/View/SerializedView.php
```diff
@@ -103,10 +103,7 @@ public function render($view = null, $layout = null): ?string
 
             return $result;
         }
-        if ($view !== false && $this->_getViewFileName($view)) {
-            return parent::render($view, false);
-        }
 
-        return null;
+        return parent::render($view, false);
```
ADmad: Can someone confirm this is okay? I don't see any value in the render method returning `null`, for serialized content or otherwise.

src/Error/Debugger.php
muglug: Not at the moment - it's ticketed here: https://github.com/vimeo/psalm/issues/1444

Next release should happen in a few hours

ADmad: Thanks, I'll keep an eye on that issue.

src/View/SerializedView.php
saeideng: render method here and otherwise returns 
```php
return $this->Blocks->get('content');
```
and it should returns `''` or a string 
so it is OK imo
just you can have a condition above this line for reducing calling  the parent class 
```php
        if ($view === false ) {
            return '';
        }
```
I dont see any benefit of  `render($view = false):`

src/Error/Debugger.php
muglug: Next release (3.2.2) is out. You'll find one annoyance - `src/Core/ObjectRegistry.php:122:9 - Class or interface Cake\Core\InstanceConfigTrait does not exist`

Don't know how you want to handle that - you might suppress it for the moment. I'm interested [in a longer term solution, too](https://github.com/vimeo/psalm/issues/1478).

src/View/SerializedView.php
ADmad: Meh, let's keep the parent call. Basically if serialization is off let the parent `View` class handle rendering instead of making decisions here.

src/Error/Debugger.php
ADmad: Yup I ignored them for now.

PR #13057, opened 2019-03-17
remove useless `SET NAMES`, it because chaset is set in dsn

dmnlk:
I found too many executed queris when I have profiled my app.
Most of queries are `SET NAMES utf8`.  This query sets charset, but MySQL can set charset in dsn.

`SET NAMES ${encoding}` query is already useless, charset is et in dsn.
https://github.com/cakephp/cakephp/blob/master/src/Database/Driver/Mysql.php#L101

`chaset` option  can use grater than PHP 5.3.6, cakephp 3.x compatible `>=5.6.0` , therefore nothing problem.
http://www.php.net/manual/en/ref.pdo-mysql.connection.php

BTW, I want to port this changes to 2.x branch. (My app still 2.x 😿 
But, 2.x branch supports  `>= 5.3.0`
https://github.com/cakephp/cakephp/blob/2.x/composer.json#L21
I have 2 way.
first, apply same patch to 2.x and update support php version '>= 5.3.6'.
second, branch for each version like below
```
		if (version_compare(PHP_VERSION, '5.3.6', '<') && !empty($config['encoding'])) {
			$flags[PDO::MYSQL_ATTR_INIT_COMMAND] = 'SET NAMES ' . $config['encoding'];
		}
                ~~
		if (empty($config['unix_socket'])) {
                         // charset was silently ignored,  < 5.3.6 
			$dsn = "mysql:host={$config['host']};port={$config['port']};dbname={$config['database']};charset={$config['encoding']}";
		} else {
			$dsn = "mysql:unix_socket={$config['unix_socket']};dbname={$config['database']};charset={$config['encoding']}";
		}
```

Which do you think more better?

### DIFF
--- a/src/Database/Driver/Mysql.php
+++ b/src/Database/Driver/Mysql.php
@@ -79,9 +79,6 @@ class Mysql extends Driver
         if (!empty($config['timezone'])) {
             $config['init'][] = sprintf("SET time_zone = '%s'", $config['timezone']);
         }
-        if (!empty($config['encoding'])) {
-            $config['init'][] = sprintf('SET NAMES %s', $config['encoding']);
-        }
 
         $config['flags'] += [
             PDO::ATTR_PERSISTENT => $config['persistent'],
@@ -100,7 +97,7 @@ class Mysql extends Driver
         if (empty($config['unix_socket'])) {
             $dsn = "mysql:host={$config['host']};port={$config['port']};dbname={$config['database']};charset={$config['encoding']}";
         } else {
-            $dsn = "mysql:unix_socket={$config['unix_socket']};dbname={$config['database']}";
+            $dsn = "mysql:unix_socket={$config['unix_socket']};dbname={$config['database']};charset={$config['encoding']}";
         }
 
         $this->_connect($dsn, $config);
--- a/tests/TestCase/Database/Driver/MysqlTest.php
+++ b/tests/TestCase/Database/Driver/MysqlTest.php
@@ -57,7 +57,7 @@ class MysqlTest extends TestCase
             'flags' => [],
             'encoding' => 'utf8mb4',
             'timezone' => null,
-            'init' => ['SET NAMES utf8mb4'],
+            'init' => []
         ];
 
         $expected['flags'] += [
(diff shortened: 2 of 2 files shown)


ADmad:
Currently charset is not set in dsn when unix socket is used. So don't you need to modify that since you removed the `SET NAMES`?

ADmad:
Please squash your commits into a single once once you are done the changes so that all your `f` commit messages go away.

dmnlk:
@ADmad sorry, I' ll squash commit after passed all tests

dmnlk:
@ADmad Fixed chaset opt when unix socket used,and related tests. thank you :)

markstory:
I would prefer we not change behavior in 2.10. It is very close to a maintenance only schedule and I would rather not rock the boat right now. If your application needs to reduce the number of queries you could make a custom driver.

dmnlk:
@markstory thank you, I understood about 2.10.x policy.

lorenzo:
I think this should be accompanied by a test in which we verify that we can write unicode char and get it back correctly.

chinpei215:
This change should not be done in a bugfix release at least. Because the charset option is unavailable if the mysql server is configured with skip-character-set-client-handshake and without character-set-server.

For example, in the environment, the following script will get corrupt characters if you don't exec `SET NAMES utf8;`:
```php
$pdo = new PDO('mysql:dbname=test_cakephp;charset=utf8', 'cakephp', 'cakephp');
// $pdo->exec('SET NAMES utf8');
$stmt = $pdo->prepare('SELECT value FROM tests');
$stmt->execute();
var_dump( $stmt->fetchAll(PDO::FETCH_ASSOC) );
```
```
array(1) {
  [0]=>
  array(1) {
    ["value"]=>
    string(3) "???"
  }
}
```

In addition, if the server is configured with skip-character-set-client-handshake and with charaset-server-set=cp932, the following script will be affected by SQL injection:
```php
$pdo = new PDO('mysql:dbname=test_cakephp;charset=utf8', 'cakephp', 'cakephp');
// $pdo->exec('SET NAMES utf8');
$stmt = $pdo->prepare('SELECT ?');
$stmt->execute(["\x95\x5c', CURRENT_DATE#"]);
var_dump( $stmt->fetchAll(PDO::FETCH_ASSOC) );
```
```
array(1) {
  [0]=>
  array(2) {
    ["\\"]=>
    string(3) "\\"
    ["CURRENT_DATE"]=>
    string(10) "2019-03-18"
  }
}
```

Adding a new configuration option that suppress `SET NAMES` query will be acceptable. If your are familiar with the mysql server configuration, you could use the option.

ADmad:
Do extra `'SET NAMES ...'` queries add any noticeable overhead? If not then we don't need to change anything.

lorenzo:
@ADmad under heavy load, they may become significant.

ADmad:
Closing in favor of #13648

PR #13065, opened 2019-03-21
Add newEmptyEntity() to make newEntity() validation save.

dereuromark:
Resolves https://github.com/cakephp/cakephp/issues/11854

This needs migration notes about newEntity() requiring an actual array now.
But this has usually been the usage anyway, and all other cases will clearly respond with a meaningful message here when trying.


src/Datasource/RepositoryInterface.php
```diff
@@ -161,6 +161,17 @@ public function save(EntityInterface $entity, $options = []);
...
+    /**
+     * This creates a new entity object.
+     *
+     * Careful: This does not trigger any field validation.
+     * This entity can be persisted without validation error as empty record.
+     * Always patch in required fields before saving.
+     *
+     * @return \Cake\Datasource\EntityInterface
+     */
+    public function createEntity(): EntityInterface;
```
markstory: We use 'create' in a few other context to generate persisted records eg. `findOrCreate()` This re-uses that term but without the persistent aspect which could be confusing. An alternative name could be `makeEntity()`

markstory (commented):
There don't appear to be any tests specifically asserting the state of the created entity. That could be useful to prevent future regressions.

src/Datasource/RepositoryInterface.php
dereuromark: True - or newEmptyEntity(), blankEntity(), initializeEntity()?

saeideng: `newEmptyEntity()` is good name , but a bit long

dereuromark: Method length is not too relevant with no arguments and IDEs to autocomplete IMO.

dereuromark:
> There don't appear to be any tests specifically asserting the state of the created entity. That could be useful to prevent future regressions.

What kind of tests do you have in mind?
Currently they are invoked indirectly from the wrapping newEntity() one.

markstory:
Is there nothing different about the behavior now? If someone updates their newEntity() calls to be newEmptyEntity() would they no longer run the risk of persisting an empty record?

dereuromark:
No, the behavior is the same, just more speaking/clear on the API level.
To have the protection for persisting, we would still need a separate "persistable" flag of some sorts to toggle on when patching happened or when you are sure you want to proceed.

dereuromark:
So are we OK with this first step in making the API more clear and less accidental towards "ooops, no validation, I didnt know that"?.

markstory:
I think this is a good change. We will have to document it though as it is a breaking change.

dereuromark:
Are we OK to move forward with this?

ADmad (approved):

Issue #13069, opened 2019-03-22
loadModel with fully qualified name of class


samuelszabo:
This is a (multiple allowed):

* [ ] bug
* [x] enhancement
* [x] feature-discussion (RFC)

* CakePHP Version: 3.7.5
* Platform and Target: built-in webserver

### What you did
```php
$this->loadModel(ProductsTable::class);
$this->Products->find();
```

### What happened
`Call to a member function find() on boolean`

### What you expected to happen
It would be great if I can _optionally_ load model directly by fully qualified name. We have an application with 300+ table classes divided into tens of plugins. For me, the top reasons are:
- ease of pointing to table (thanks to class names autocomplete)
- easier refactoring (changing model name)

Also, `TableLocator::get()` accept fully qualified name.

dereuromark:
As a side note:
I dont like that loadModel silently returns false here. Much better would be to throw meaningful exception in such cases (in 4.x)?

celsowm:
Would be nice for **contain()** too :)

saeideng:
>But this is fixed in 4.0 afaik.

no
```php
//debug($this->loadModel('InvalidExample'));
object(Cake\ORM\Table) {

	'registryAlias' => 'InvalidExample',
	'table' => 'invalid_example',
	'alias' => 'InvalidExample',
	'entityClass' => 'Cake\ORM\Entity',
	'associations' => [],
	'behaviors' => [],
	'defaultConnection' => 'default',
	'connectionName' => 'default'

}
```

saeideng:
this is useful when someone need to load a Model , but he did not bake(create ) it 
but in other side this is painful for someone that uses incorrect file name (model file name) and silent ignoring the model and hard debugging

dereuromark:
fixed = not returning false (no object)

saeideng:
same return  in cake 3.7.x

dereuromark:
I don't think that is true for all cases.
See the above "Call to a member function find() on boolean" error message.
My point still stands, that should never happen ideally.

samuelszabo:
Guys, I think we have three subjects here. To make conversation a little bit clearer, let me make a quick summary.

(a) An option to loadModel with fully qualified name of a class, which is an enhancement.

[...]

PR #13072, opened 2019-03-27
Allow using FQCN with ModelAwareTrait::loadModel().

ADmad:
Refs #13069.

### DIFF
--- a/src/Datasource/ModelAwareTrait.php
+++ b/src/Datasource/ModelAwareTrait.php
@@ -77,7 +77,8 @@ trait ModelAwareTrait
      * If a repository provider does not return an object a MissingModelException will
      * be thrown.
      *
-     * @param string|null $modelClass Name of model class to load. Defaults to $this->modelClass
+     * @param string|null $modelClass Name of model class to load. Defaults to $this->modelClass.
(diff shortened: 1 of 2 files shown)


dereuromark (approved):

src/Datasource/ModelAwareTrait.php
```diff
@@ -97,7 +98,19 @@ public function loadModel($modelClass = null, $modelType = null)
...
+        $options = [];
+        if (strpos($modelClass, '\\') === false) {
+            list(, $alias) = pluginSplit($modelClass, true);
+        } else {
+            $options['className'] = $modelClass;
+            $alias = substr(
+                $modelClass,
+                strrpos($modelClass, '\\') + 1,
+                -strlen($modelType)
+            );
```
markstory: Would this be easier to read if it was

```php
$parts = explode('\\', $modelClass);
$alias = array_pop($parts);
```

ADmad: But my code is more efficient with 2 string operations as opposed to 2 array operations :).  Also your example will still need an additional operation to remove the "Table" suffix.

PR #13081, opened 2019-04-02
Ensure precision is not lost in DecimalType.

### LINKED ISSUE #10233: 3.x decimal precision lost before saving data
* [x] bug
* [ ] enhancement

* CakePHP Version: v3.3.13
* Platform and Target:  NGINX v1.11 with MySQL v5.7.17

When I try to save a data in table the precision of decimal value is lost, I checked the code and the class DecimalType, before save the data, convert the value to a float using sprintf function, so why it is converting a decimal value to a float?? It's a decimal value the precision is larger than an float.

The example code with the problem:
```php
$scoreData = [
    'distributor_id' => 1,
    'points' => '19.701592653589',
    'month_ref' => '2017-02-06',
];
...
$score = $this->Distributors->Scores->newEntity($scoreData);
...
$this->Distributors->Scores->save($score);
```
The value saved on database is rounded to: 19.701593;

In MySQL table the column is DECIMAL(18,12);

ADmad:
Closes #10233 

While the tests have changed I don't think it will make any practical difference when values are saved to db and this change should be safe for bugfix release.

### DIFF
--- a/tests/TestCase/Database/Type/DecimalTypeTest.php
+++ b/tests/TestCase/Database/Type/DecimalTypeTest.php
@@ -131,13 +131,16 @@ class DecimalTypeTest extends TestCase
         $this->assertSame('0.000000', $result);
 
         $result = $this->type->toDatabase(2, $this->driver);
-        $this->assertSame('2.000000', $result);
+        $this->assertSame('2', $result);
 
         $result = $this->type->toDatabase(2.99, $this->driver);
-        $this->assertSame('2.990000', $result);
+        $this->assertSame('2.99', $result);
 
         $result = $this->type->toDatabase('2.51', $this->driver);
         $this->assertSame('2.51', $result);
+
+        $result = $this->type->toDatabase(0.12345678, $this->driver);
+        $this->assertSame('0.12345678', $result);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


tests/TestCase/Database/Type/DecimalTypeTest.php
```diff
@@ -131,13 +131,16 @@ public function testToDatabase()
...
+        $this->assertSame('2', $result);
 
         $result = $this->type->toDatabase(2.99, $this->driver);
-        $this->assertSame('2.990000', $result);
+        $this->assertSame('2.99', $result);
 
         $result = $this->type->toDatabase('2.51', $this->driver);
         $this->assertSame('2.51', $result);
+
+        $result = $this->type->toDatabase(0.12345678, $this->driver);
```
markstory: Should we add a test for a value with 10 or 12 decimal places?

ADmad: `QueryTest::testSelectLargeNumbers()` does test that and it fails due to values being converted to scientific notation :slightly_frowning_face:

dereuromark: Sounds like this should better go into 3.next then.
And it could be a separate Type class if we want to make sure.

ADmad: Yup, the change isn't safe for a bugfix release. Closing the PR.

PR #13086, opened 2019-04-03
4.x decimal type

ADmad:
Refs #10233

### DIFF
--- a/src/Database/Type/DecimalType.php
+++ b/src/Database/Type/DecimalType.php
@@ -45,50 +45,55 @@ class DecimalType extends BaseType implements BatchCastingInterface
     /**
      * Convert integer data into the database format.
      *
-     * @param string|int|float $value The value to convert.
+     * @param mixed $value The value to convert.
      * @param \Cake\Database\Driver $driver The driver instance to convert with.
-     * @return string|null
+     * @return mixed
      * @throws \InvalidArgumentException
      */
-    public function toDatabase($value, Driver $driver): ?string
+    public function toDatabase($value, Driver $driver)
     {
         if ($value === null || $value === '') {
             return null;
         }
-        if (!is_scalar($value)) {
-            throw new InvalidArgumentException(sprintf(
-                'Cannot convert value of type `%s` to a decimal',
-                getTypeName($value)
-            ));
-        }
+
         if (is_numeric($value)) {
+            return $value;
+        }
+
+        if (is_object($value)
+            && method_exists($value, '__toString')
+            && is_numeric(strval($value))
+        ) {
             return strval($value);
         }
 
-        return sprintf('%F', $value);
+        throw new InvalidArgumentException(sprintf(
+            'Cannot convert value of type `%s` to a decimal',
+            getTypeName($value)
+        ));
     }
 
     /**
      * Convert float values to PHP floats
      *
-     * @param null|string|resource $value The value to convert.
+     * @param mixed $value The value to convert.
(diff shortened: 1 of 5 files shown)


ADmad:
Not sure why `DecimalType` supports parsing of localized string. Anyone using localized string for decimals is asking for trouble.

dereuromark:
Well, the input format for it can differ per language. We use a comma not a dot. As such this was always a usability issue to allow for this.

ADmad:
@dereuromark Yeah I figured. Though `Number::parseFloat()` is used for parsing which as the name suggests returns floats so you would end up with same problems too large / small values.

ADmad:
It seems sqlite is unable to deal with large numers https://travis-ci.org/cakephp/cakephp/jobs/515170690#L674. Same test passes for other dbs. Guess I'll just skip the test for sqlite.

inoas:
I don't think the DecimalType should support localization.
A LocalizedDecimalType could but for writing between the database and PHP a string and a dot for separation sounds consistent, expected and reliable.

src/Database/Type/DecimalType.php
```diff
@@ -45,50 +45,55 @@ class DecimalType extends BaseType implements BatchCastingInterface
...
-        if (!is_scalar($value)) {
-            throw new InvalidArgumentException(sprintf(
-                'Cannot convert value of type `%s` to a decimal',
-                getTypeName($value)
-            ));
-        }
-        if (is_string($value) && is_numeric($value)) {
+
+        if (is_numeric($value)) {
             return $value;
```
markstory: Should this string cast float/ints into string?

tests/TestCase/Database/Type/DecimalTypeTest.php
```diff
@@ -177,6 +194,12 @@ public function testMarshal()
 
         $result = $this->type->marshal(['3', '4']);
         $this->assertNull($result);
+
+        // This test is to indicate the problem that will occur if you use
+        // float/double values which get converted to scientific notation by PHP.
+        // To avoid this problem always using strings to indicate decimals values.
+        $result = $this->type->marshal(1234567890123456789.2);
```
markstory: Marshalling should generally get strings, so this situation should be rare. Should there be a test added for values with 10+ of decimal places not losing precision?

markstory:
@ionas @ADmad removing localized float support will be interpreted as a bug/defect by a few people. It was added after issues were opened about localized values being corrupted.

ADmad:
> I don't think the DecimalType should support localization.

@ionas It's an opt in feature which has to be explicitly enable. So those who don't need it are safe.

src/Database/Type/DecimalType.php
ADmad: We already specify the binding type as `PDO::PARAM_STR` through the `toStatement()`. So there's no need for explicit casting, let PDO handle the value conversion as needed.

tests/TestCase/Database/Type/DecimalTypeTest.php
ADmad: I have added more tests.

dereuromark:
I would also be fine if the core shipped with two separate Type classes: Decimal and LocalizedDecimal. The latter could extend the first. We might need sth similar for other types, too?

ADmad:
I don't think a separate class to handle localised values is necessary. Supporting localised values only needs few lines of extra code and doesn't add any significant complexity.

ADmad:
@markstory Anything else to be done here?

PR #13089, opened 2019-04-03
Improve code readability for humans and IDEs

ravage84:


src/TestSuite/TestCase.php
```diff
@@ -197,21 +197,22 @@ public function loadFixtures()
      * Useful to test how plugins being loaded/not loaded interact with other
      * elements in CakePHP or applications.
      *
-     * @param array $plugins list of Plugins to load
+     * @param array $plugins List of Plugins to load.
      * @return \Cake\Http\BaseApplication
      */
     public function loadPlugins(array $plugins = [])
     {
+        /** @var BaseApplication $app */
```
dereuromark: Technically this is `|MockObject`
But as long as the mock is not used inside of the method, this is fine.

dereuromark (approved):

markstory:
Looks good if you can convince the typecheckers that you know what you're doing.

src/TestSuite/TestCase.php
ADmad: Should use FQCN.

ravage84: @dereuromark agree, but for this I focused on readability and on the actual relevant use.
@ADmad done.

src/TestSuite/TestCase.php
```diff
@@ -197,21 +197,22 @@ public function loadFixtures()
      * Useful to test how plugins being loaded/not loaded interact with other
      * elements in CakePHP or applications.
      *
-     * @param array $plugins list of Plugins to load
+     * @param array $plugins List of Plugins to load.
      * @return \Cake\Http\BaseApplication
      */
     public function loadPlugins(array $plugins = [])
     {
+        /** @var \Cake\Http\BaseApplication $app */
```
saeideng: after adding this line you can remove 
https://github.com/cakephp/cakephp/blob/bd2cee377d9b63f01a0a6cd443e580adfb588a10/phpstan.neon#L27 line

ravage84: @saeideng done, thanks for pointing out.

Issue #13099, opened 2019-04-06
Command doesnt autoload own model


dereuromark:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.x latest

### What you did
```php
class MyCommand extends Command
{
    public $modelClass = 'Tools.Emails';

    public function execute(Arguments $args, ConsoleIo $io)
    {
        $this->Emails->doSth();
    }
}
```
### What happened

Notice: Notice (8): Undefined property

+ Exception of method not found on null.

### What you expected to happen
Like with controllers and shells before, the defined primary model should be available without having to first call
```
$this->loadModel();
```

ADmad:
I actually prefer command not having this auto-magic.

dereuromark:
why that? we should lazyload it, but otherwise the whole property would be without purpose.

markstory:
I am not sure this is a bug as I don't remember this ever being something we discussed implementing for commands.

I am not opposed to adding this behavior as it adds consistency with controllers.

saeideng:
>I actually prefer commands not having this auto-magic.

:+1:  even in controllers in future (4.x or 5.x)

dereuromark:
PR open for 3.next.

PR #13100, opened 2019-04-07
Add missing autoload to make $modelClass property useful.

dereuromark:
Resolves https://github.com/cakephp/cakephp/issues/13099 

This is to sync it with Shells (before) and other classes like controller.
The important part is: This is opt-in.
If you don't specify any $modelClass, it will just be as before.
But a manual loadModel() call without any args required is just bonkers here if there is a $modelClass property for it.

### DIFF
--- a/src/Console/Command.php
+++ b/src/Console/Command.php
@@ -63,6 +63,10 @@ class Command
         $this->modelFactory('Table', function ($alias) {
             return $this->getTableLocator()->get($alias);
         });
+
+        if (isset($this->modelClass)) {
+            $this->loadModel();
+        }
     }
 
     /**
--- a/tests/test_app/TestApp/Command/AutoLoadModelCommand.php
+++ b/tests/test_app/TestApp/Command/AutoLoadModelCommand.php
@@ -1,0 +1,9 @@
+<?php
+namespace TestApp\Command;
+
+use Cake\Console\Command;
+
+class AutoLoadModelCommand extends Command
+{
+    public $modelClass = 'Posts';
+}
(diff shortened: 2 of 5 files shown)


src/Console/Command.php
```diff
@@ -63,6 +63,10 @@ public function __construct()
         $this->modelFactory('Table', function ($alias) {
             return $this->getTableLocator()->get($alias);
         });
+
+        if (isset($this->modelClass)) {
```
ADmad: Loading the model in constructor itself isn't lazy loading. Shouldn't the __set() magic method be overridden for that like it's done in controller for that?

dereuromark: For shells and commands nano opt lazyloading sounds wrong.

markstory: I don't think we need lazy loading either. Commands don't really need the same performance profile that web controllers do.

markstory (commented):
Any way to add a test for this?

dereuromark:
I added one :)

tests/test_app/TestApp/Command/AutoLoadModelCommand.php
```diff
@@ -0,0 +1,9 @@
+<?php
+namespace TestApp\Command;
+
+use Cake\Console\Command;
+
+class AutoLoadModelCommand extends Command
+{
+    public $modelClass = 'Posts';
```
ADmad: Not in scope of this PR but this is on public property we seem to have forgotten to deal with.

dereuromark: You want to make this protected for 4.0? :) Sure, sounds like a good idea.

PR #13101, opened 2019-04-08
Save fails if associated HasMany entities array is indexed by string keys

garas:
When associated entities are indexed by string keys and save strategy is SAVE_REPLACE, saving fails because generated `'OR'` conditions. Array keys should be ignored.

Ex. in test case when saving entity, unused dependent articles are deleted by generated condition
```php
[
        'NOT' => [
                'OR' => [
                        'one' => ['id' => 5],
                        'two' => ['id' => 6],
                ]
        ],
        ['author_id' => 5]
]
```
which generates 
```sql
DELETE FROM articles
WHERE (NOT ((one = :c0 OR two = :c1)) AND author_id = :c2)
```

Or even worse if keys has spaces, generated delete condition is

```php
[
        'NOT' => [
                'OR' => [
                        'Another Random Post' => ['id' => 5],
                        'One more random post' => ['id' => 6],
                ]
        ],
        ['author_id' => 5]
]
```
which generates invalid
```sql
DELETE FROM articles
WHERE (NOT ((Another random post :c0 OR One more random post :c1)) AND author_id = :c2)
```

It should be deleted by generated condition
```php
[
        'NOT' => [
                'OR' => [
                        ['id' => 5],
                        ['id' => 6],
                ]
        ],
        ['author_id' => 5]
]
```
which generates 
```sql
DELETE FROM articles
WHERE (NOT ((id = :c0 OR id = :c1)) AND author_id = :c2)
```

### DIFF
--- a/tests/TestCase/ORM/Association/HasManyTest.php
+++ b/tests/TestCase/ORM/Association/HasManyTest.php
@@ -1141,7 +1141,7 @@ class HasManyTest extends TestCase
             ]
         ], ['associated' => ['Articles']]);
 
-        $entity = $authors->save($entity, ['associated' => ['Articles']]);
+        $entity = $authors->saveOrFail($entity, ['associated' => ['Articles']]);
         $sizeArticles = count($entity->articles);
         $this->assertSame($sizeArticles, $authors->Articles->find('all')->where(['author_id' => $entity['id']])->count());
 
--- a/tests/TestCase/ORM/Association/HasManyTest.php
+++ b/tests/TestCase/ORM/Association/HasManyTest.php
@@ -24,6 +24,7 @@ use Cake\ORM\Association;
 use Cake\ORM\Association\HasMany;
 use Cake\ORM\Entity;
 use Cake\TestSuite\TestCase;
+use Cake\Utility\Hash;
 
 /**
  * Tests HasMany class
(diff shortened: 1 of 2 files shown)


tests/TestCase/ORM/Association/HasManyTest.php
```diff
@@ -1121,6 +1122,42 @@ public function testSaveReplaceSaveStrategyDependent()
...
+
+        $articleId = $entity->articles[0]->id;
+        $entity->articles = [
+            'one' => $entity->articles[1],
+            'two' => $entity->articles[2],
+        ];
+
+        $authors->save($entity, ['associated' => ['Articles']]);
+
+        $this->assertEquals($sizeArticles - 1, $authors->Articles->find('all')->where(['author_id' => $entity['id']])->count());
```
dereuromark: Best to always use assertSame() with comparing scalar values.

tests/TestCase/ORM/Association/HasManyTest.php
```diff
@@ -1121,6 +1122,42 @@ public function testSaveReplaceSaveStrategyDependent()
...
+        $sizeArticles = count($entity->articles);
+        $this->assertEquals($sizeArticles, $authors->Articles->find('all')->where(['author_id' => $entity['id']])->count());
+
+        $articleId = $entity->articles[0]->id;
+        $entity->articles = [
+            'one' => $entity->articles[1],
+            'two' => $entity->articles[2],
+        ];
+
+        $authors->save($entity, ['associated' => ['Articles']]);
```
dereuromark: Should this be saveOrFaill()?

tests/TestCase/ORM/Association/HasManyTest.php
```diff
@@ -1121,6 +1121,42 @@ public function testSaveReplaceSaveStrategyDependent()
...
+        $entity = $authors->newEntity([
+            'name' => 'mylux',
+            'articles' => [
+                ['title' => 'One Random Post', 'body' => 'The cake is not a lie'],
+                ['title' => 'Another Random Post', 'body' => 'The cake is nice'],
+                ['title' => 'One more random post', 'body' => 'The cake is forever']
+            ]
+        ], ['associated' => ['Articles']]);
+
+        $entity = $authors->save($entity, ['associated' => ['Articles']]);
```
dereuromark: saveOrFail() ?

garas: I've modified test case based on your review. It was just copy-paste of test case above this one with few modifications, so I didn't look much at those lines.

ADmad:
Nice catch @garas.

PR #13105, opened 2019-04-09
Exclude more files from being exported

ravage84:
I was toying around with a deployment script and noticed these unnecessary (in a deployment context) files.

One could argue that even ``src/TestSuite/`` could be excluded

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -55,3 +55,8 @@ phpunit.xml.dist export-ignore
 tests/test_app export-ignore
 tests/TestCase export-ignore
 .github export-ignore
+.mailmap export-ignore
+.varci.yml export-ignore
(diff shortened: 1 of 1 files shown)


.gitattributes
```diff
@@ -55,3 +55,8 @@ phpunit.xml.dist export-ignore
 tests/test_app export-ignore
 tests/TestCase export-ignore
 .github export-ignore
+.mailmap export-ignore
+.varci.yml export-ignore
+phpcs.xml.dist export-ignore
+phpstan.neon export-ignore
```
ADmad: This needs to be kept. If someone wants to use phpstan for their app then can include this config file in their app's phpstan.neon.

PR #13113, opened 2019-04-13
Improve console help output

markstory:
Group commands/shells based on their namespaces instead of trying to do it based on registered command name. Plugin commands can be explicitly registered with no `plugin.command` variant. These commands would end up grouped under 'app' in the previous implementation. Now
they are correctly classified under the matching plugin.

## Example output now

```
$ bin/cake -h
Current Paths:

* app:  src
* root: /Users/markstory/Sites/cake4app
* core: /Users/markstory/Sites/cake4app/vendor/cakephp/cakephp

Available Commands:

App:
 - console

Bake:
 - bake
 - bake all
 - bake behavior
 - bake cell
 - bake command
 - bake component
 - bake controller
 - bake controller all
 - bake fixture
 - bake fixture all
 - bake form
 - bake helper
 - bake mailer
 - bake middleware
 - bake model
 - bake model all
 - bake plugin
 - bake shell
 - bake shell_helper
 - bake task
 - bake template
 - bake template all
 - bake test

CakePHP:
 - cache
 - completion
 - help
 - i18n
 - plugin
 - routes
 - schema_cache
 - server
 - upgrade
 - version

To run a command, type `cake command_name [args|options]`
To get help on a specific command, type `cake command_name --help`
```

### DIFF
--- a/src/Command/HelpCommand.php
+++ b/src/Command/HelpCommand.php
@@ -23,7 +23,7 @@ use Cake\Console\ConsoleIo;
 use Cake\Console\ConsoleOptionParser;
 use Cake\Console\ConsoleOutput;
 use Cake\Core\Configure;
-use Cake\Utility\Inflector;
+use Cake\Core\Plugin;
 use SimpleXMLElement;
 
 /**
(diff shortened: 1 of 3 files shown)


ADmad:
The output looks good but a single list / column could get really long if one has a handful of plugins. Would it be feasible to list commands within a namespace in 2-3 columns?

markstory:
Multiple columns could be done. It requires a more complex layout system similar to what we have in TableHelper but without borders.

dakota:
Might be a good idea to allow borderless tables? Definitely something that I can see being used in my own applications.

markstory:
@Dakota Agreed, I could see it being useful in a few places as well. I think it could be a separate shell helper though. 

If it is ok with you folks, I would like to merge this and we can explore the column layout separately.

ADmad:
@markstory :+1: for progressive enhancement.

ADmad:
The static analyzers needed to be made happy before merging :)  https://scrutinizer-ci.com/g/cakephp/cakephp/inspections/9ffa7883-75ad-46b2-96cd-5832902954c3

dereuromark:
2 of 3 checks passed (only AppVeyor build failed), why is the Scrut thingie not part of that checklist?

ADmad:
https://github.com/cakephp/cakephp/pull/13113/checks does show scrutinizer failure and it showed on the PRs primary view too before PR was merged. Not sure why the consolidated view after merge doesn't show it.

markstory:
@ADmad Sorry. Still used to scrutinizer being mostly lies.

markstory:
Fixed in 1ca0760a6678875e0711ef7f3e0c5eb88f431f4f

ADmad:
@markstory No problem, in 4.x scrutinizer only does phpcs and static analysis checks so the reports are pertinent.

PR #13117, opened 2019-04-15
Try to fix appveyor build.

markstory:
SQLServer is returning decimal values with no 0.


ADmad:
> SQLServer is returning decimal values with no 0.

Just M$ things :man_shrugging:

PR #13119, opened 2019-04-15
Allow using DateTimeImmutable instance.

### LINKED ISSUE #13087: Use DateTimeInterface for withModified transformation in Response
This is a (multiple allowed):

* [x] bug
* [ ] enhancement

### What you did
I return a response `$response->withModified($myEntity->modified)`.
My locale was set to `de_DE`, and the modified property is a FrozenTime object. 

### What happened
The `_getUTCDate` function in Response.php will check if the object given is an instance of DateTime, which FrozenTime is not. It will fallback to creating a new DateTime object, using toString of the original FrozenTimeobject, and it tries to resolve 'dd.mm.yy, HH:ii' (the default german format) which fails.

I expect `_getUTCDate` to handle DateTimeInterface instead of DateTime, allowing for support of FrozenTime and other objects which to not extend DateTime but do implement the interface which should be sufficient for conversion to the UTC timezone.

ADmad:
Closes #13087.

### DIFF
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -27,6 +27,7 @@ parameters:
         - '#Call to an undefined method DOMNode::setAttribute\(\)#'
         - '#Binary operation "\+" between array|false and array results in an error#'
         - '#Result of method Cake\\Auth\\BaseAuthenticate::unauthenticated\(\) \(void\) is used#'
+        - '#Call to an undefined method DateTimeInterface::setTimezone\(\)#'
     earlyTerminatingMethodCalls:
         Cake\Console\Shell:
             - abort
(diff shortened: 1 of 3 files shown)


phpstan.neon
```diff
@@ -27,6 +27,7 @@ parameters:
         - '#Call to an undefined method DOMNode::setAttribute\(\)#'
         - '#Binary operation "\+" between array|false and array results in an error#'
         - '#Result of method Cake\\Auth\\BaseAuthenticate::unauthenticated\(\) \(void\) is used#'
+        - '#Call to an undefined method DateTimeInterface::setTimezone\(\)#'
```
ADmad: Apparently the interface does not have the `setTimezone()` method but since both `DateTime` and `DateTimeImmutable` do and userland code can't implement the `DateTimeInterface` I don't see any practical problem in ignoring this.

PR #13120, opened 2019-04-15
Do not show AppVeyor PHP download progress

garas:
AppVeyor currently generates very long log with duplicate lines like `Progress: Downloading php 7.2.17... 92%`


garas:
Also install was blocked because 7z asked if it should overwrite files.

And backport bc0a9898b8093610eedf37e853ea71e3a9dea819

PR #13121, opened 2019-04-15
Doc block return type array specification.

dereuromark:
small phpstan/IDE improvement

### DIFF
--- a/src/Cache/Cache.php
+++ b/src/Cache/Cache.php
@@ -496,8 +496,7 @@ class Cache
      *
      * @param array $keys Array of cache keys to be deleted
      * @param string $config name of the configuration to use. Defaults to 'default'
-     * @return array of boolean values that are true if the value was successfully deleted, false if it didn't exist or
-     * couldn't be removed
+     * @return bool True if the items were successfully removed. False if there was an error.
      */
     public static function deleteMany($keys, $config = 'default')
     {
--- a/src/Console/CommandCollection.php
+++ b/src/Console/CommandCollection.php
@@ -162,7 +162,7 @@ class CommandCollection implements IteratorAggregate, Countable
      * the long name (`plugin.command`) will be returned.
      *
      * @param string $plugin The plugin to scan.
-     * @return array Discovered plugin commands.
+     * @return string[] Discovered plugin commands.
      */
     public function discoverPlugin($plugin)
     {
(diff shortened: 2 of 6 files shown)


src/Cache/Cache.php
```diff
@@ -496,8 +496,7 @@ public static function delete($key, $config = 'default')
      *
      * @param array $keys Array of cache keys to be deleted
      * @param string $config name of the configuration to use. Defaults to 'default'
-     * @return array of boolean values that are true if the value was successfully deleted, false if it didn't exist or
-     * couldn't be removed
+     * @return bool True if the items were successfully removed. False if there was an error.
```
ADmad: But `CacheEngine::deleteMany()` returns an array. So that method will also need to be updated.

dereuromark: That would be a functional change then, right? Still a bugfix?

garas: It should be `bool[]` here and `CacheEngine::deleteMany()`. And old description was correct.

garas: Oh, PSR-16 `CacheInterface::deleteMultiple()` returns single `bool`.

Probably correct way is to fill array by calling `CacheInterface::delete()` for each key?

dereuromark: If at least one fails, the return is false, otherwise true. Probably.

ADmad: This seems to be a discrepancy which has crept in when `SimpleCache` for PSR-16 support was added. Return value for `CacheInterface::deleteMultiple()` is `bool` while our old `CacheEngine::deleteMany()` method returned `bool[]`.

dereuromark: Should we merge this and have a separate 3.8 and 4.0 fix for the functional part?

ADmad: I checked the good and this docblock change is good to merge. 

`Cache::deleteMany()` uses `SimpleCacheEngine::deleteMultiple()` which iterates over the array returned by `deleteMany()` of cache engines and returns a boolean.

ADmad: Though change is return value of `Cache::deleteMany()` breaks backwards compatibility. Perhaps the method should be updated to fill array by calling CacheInterface::delete() for each key as @garas suggested.

ADmad: https://github.com/cakephp/cakephp/pull/13122

PR #13122, opened 2019-04-16
Fix backwards compatibility breakage.

ADmad:
`Cache::deleteMany()` again returns `bool[]` instead of `bool` as in v3.6 and lower.


dereuromark:
All right. We fix it to bool then for 4.0.

ADmad:
The docblock and actual return value are already boolean in 4.x.

PR #13126, opened 2019-04-17
Fix shutdown warnings in PHP7.2+

### LINKED ISSUE #13085: 2.x file cache errors on PHP 7.2 when running a single database test by itself
This is a (multiple allowed):

* [x] bug
* [ ] enhancement

* CakePHP Version: 2.10.15
* Platform and Target: Windows 10, PHP 7.2.16 (but bug present from PHP 7.2.0 onwards)

When I run a single test which hits the database, I get the following errors in my console. These are spawned by the `Mysql::__destruct()` method, which tries to update `method_cache` cache. The actual test passes and I get the PhpUnit green output in the console followed by the errors.

If I run a single test which doesn't hit the database, I do not get the error ouput. If I run the entire test suite, I do not get the error ouput.

I have tested this on PHP 7.1.latest and the error does not occur. It occurs from PHP 7.2.0. 

I am using PHPUnit 5.7.27, in case that is relevant.

Therefore I assume it is a PHP bug. However I am having real difficulty reproducing it with a plain PHP script so I'm not at the stage where I can submit a PHP bug report. If anyone else is able to reproduce this error, please do let me know!

```
» app/Console/cake test app Jobs/Queue/Database --configuration=phpunit.xml

CakePHP Test Shell
PHPUnit 5.7.27 by Sebastian Bergmann and contributors.

Runtime:       PHP 7.2.16
Configuration: C:\inetpub\my_project\phpunit.xml

.....                                                               5 / 5 (100%)

Time: 2.63 seconds, Memory: 24.00MB

OK (5 tests, 8 assertions)

Warning: flock(): supplied resource is not a valid stream resource in C:\inetpub\my_project\app\Vendor\cakephp\cakephp\lib\Cake\Cache\Engine\FileEngine.php on line 138

Warning: SplFileObject::rewind(): stream does not support seeking in C:\inetpub\my_project\app\Vendor\cakephp\cakephp\lib\Cake\Cache\Engine\FileEngine.php on line 141

Fatal error: Uncaught RuntimeException: Cannot rewind file c:\inetpub\my_project\app\tmp\cache\persistent\my_project_dev_cake_core_method_cache in C:\inetpub\my_project\app\Vendor\cakephp\cakephp\lib\Cake\Cache\Engine\FileEngine.php:141
Stack trace:
#0 C:\inetpub\my_project\app\Vendor\cakephp\cakephp\lib\Cake\Cache\Engine\FileEngine.php(141): SplFileObject->rewind()
#1 C:\inetpub\my_project\app\Vendor\cakephp\cakephp\lib\Cake\Cache\Cache.php(317): FileEngine->write('my_project_dev_cake...', 'a:3:{i:0;b:0;s:...', 10)
#2 C:\inetpub\my_project\app\Vendor\cakephp\cakephp\lib\Cake\Model\Datasource\DboSource.php(3748): Cache::write('method_cache', Array, '_cake_core_')
#3 [internal function]: DboSource->__destruct()
#4 {main}
  thrown in C:\inetpub\my_project\app\Vendor\cakephp\cakephp\lib\Cake\Cache\Engine\FileEngine.php on line 141
```

I have spent a lot of time debugging this without any joy so far. I have established that when running a single test, `FileEngine::_setKey()` does not enter the if block (line 356) when doing the write call, because the previous cache operation was the read for the same key. Whereas, when running all tests, another cache operation has happened somewhere, so it has to enter the if block and re-establish the `SplFileObject` instance.

markstory:
Check if the current file is valid before re-using it. This fixes warnings emitted during process shutdown when DboSource is persisting the method cache.

Fixes #13085

### DIFF
--- a/lib/Cake/Cache/Engine/FileEngine.php
+++ b/lib/Cake/Cache/Engine/FileEngine.php
@@ -352,7 +352,11 @@ class FileEngine extends CacheEngine {
 		if (!$createKey && !$path->isFile()) {
 			return false;
 		}
-		if (empty($this->_File) || $this->_File->getBaseName() !== $key) {
+		if (
+			empty($this->_File) ||
+			$this->_File->getBaseName() !== $key ||
+			$this->_File->valid() === false
+		) {
 			$exists = file_exists($path->getPathname());
 			try {
 				$this->_File = $path->openFile('c+');


garethellis36:
Any idea when 2.10.18 will be released please? :)

markstory:
@garethellis36 Current target is may 11.

PR #13127, opened 2019-04-17
Fix warnings from FileEngine on shutdown

markstory:
Check if the current file is valid before re-using it. This fixes warnings emitted during process shutdown when DboSource is persisting the method cache.

Forward port fix for #13085 to 3.x


garethellis36:
Is a separate PR needed to fix this for 2.x?

Issue #13129, opened 2019-04-17
Cake 4 - Why is templates/element/Flash uppercase?


davidyell:
This is a (multiple allowed):

* [ ] bug
* [ ] enhancement
* [x] feature-discussion (RFC)

* CakePHP Version: 4.0.0-alpha1
* Platform and Target: PHP 7.3.2
### What you did
I was chatting with @ADmad on Slack and they mentioned that the templates folder should be lowercase for things which are not mapped to a Controller.

I wondered why the Flash and Email folders were uppercase?

### What happened
@ADmad Told me to open a ticket 😉

x-zolezzi:
I agree and I think the same question for the folder Email

rochamarcelo:
The migration guide (https://book.cakephp.org/4.0/en/appendices/4-0-migration-guide.html) says that Email was renamed to email: "Templates have been moved from src/Template/ to templates/ folder on app and plugin root. Special templates folder like Cell, Element, Email and Plugin have be renamed to lower case cell, element, email and plugin respectively."

ADmad:
Renaming `Email` to `email` is surely something I missed in the skeleton. The core code does use `email`.

ADmad:
Renaming `Flash` to `flash` is something that needs to be done after corresponding change is done to `FlashComponent` which sets the element to use.

PR #13130, opened 2019-04-17
Rename element/Flash to element/flash.

ADmad:
This is inline with other template folder renames where lower case name
is used for special folders and those which don't correspond to a class name.

Refs #13129.


ADmad:
@markstory Thanks for taking care of the failing tests.

PR #13131, opened 2019-04-17
#13088 - Validation::datetime

### LINKED ISSUE #13088: Validation::datetime
This is a (multiple allowed):

* [x] bug
* [x] enhancement

* CakePHP Version: 3.6.14.
* Platform and Target: Apache 2.4/PHP7.2/MySQL 5.5

### What you did
I wish to benefit the Validation::datetime rule but using an ISO8601 format which implies a T as a separator.
ISO8601 extended hm, without Z `2019-09-06T15:00:00+02:00`

### What happened
At line 561 of Validation\Validation.php, the instruction tries to determine date and time parts by splitting the string but only considers space caracter as a valid separator.
```php
$parts = explode(' ', $check);
```
In regards to ISO8601 datetime format, could we improve the method and consider T as valid separator?

it-mob:
This PR is an attempt to fix #13088


src/Validation/Validation.php
```diff
@@ -14,6 +14,7 @@
  */
 namespace Cake\Validation;
 
+use Cake\Chronos\Date;
```
saeideng: why you need this namespace ?

saeideng (dismissed):

tests/TestCase/Validation/ValidationTest.php
```diff
@@ -1589,6 +1589,19 @@ public function testDateTimeWithMeriadian()
...
+     * Tests that it is possible to pass a date with a T separator
+     *
+     * @return void
+     */
+    public function testDateTimeISO()
+    {
+        $this->assertFalse(Validation::dateTime('2007/10/04T1:50'));
+        $this->assertTrue(Validation::dateTime('2007/10/04T01:50'));
+        $this->assertTrue(Validation::dateTime('2017/12/04T15:38'));
+        $this->assertFalse(Validation::dateTime('2007/10/04T58:38'));
```
markstory: Should we also have a test for ISO8859-1 style datetimes?

src/Validation/Validation.php
it-mob: It' a mistake and must be removed. I attempted to support the whole ISO8601 format using Date::createFromFormat however, it parses 2019-06-05T58:58:00Z+0200 to something like 2019-06-07T10:58:00Z+0200

it-mob:
I wonder if that modification makes sense. It's an attempt to support ISO8601 but it fails to parse fraction of seconds and time zone designator parts. A better idea might be to create a new method dedicated to ISO8601 date format? Or extend `$dateFormat` argument with a new option 'iso' that would split up the chain according to ISO8601 specification?

markstory:
Adding separate support for 8859-1 datetimes works for me as well.

it-mob:
@markstory the primary idea of this PR was to recognize T as a valid separator. But anyway, I'll work on the 8601 support.

src/Validation/Validation.php
```diff
@@ -576,6 +577,10 @@ public static function datetime($check, $dateFormat = 'ymd', $regex = null)
         if (is_object($check)) {
             return false;
         }
+        if ($dateFormat === 'iso8601' && !static::iso8601($check)) {
```
dereuromark: Can we make 'iso8601' magic string a reusable class constant here?

src/Validation/Validation.php
```diff
@@ -577,7 +582,7 @@ public static function datetime($check, $dateFormat = 'ymd', $regex = null)
         if (is_object($check)) {
             return false;
         }
-        if ($dateFormat === 'iso8601' && !static::iso8601($check)) {
+        if ($dateFormat === self::DATETIME_ISO8601 && !static::iso8601($check)) {
```
dereuromark: we use static::

src/Validation/Validation.php
```diff
@@ -585,12 +590,41 @@ public static function datetime($check, $dateFormat = 'ymd', $regex = null)
...
 
+    /**
+     * Validates an iso8601 datetime format
+     * ISO8601 recognize datetime like 2019 as a valid date. To validate and check date integrity, use @see \Cake\Validation\Validation::datetime()
+     *
+     * @param string|\DateTimeInterface $check Value to check
+     *
+     * @return bool True if the value is valid, false otherwise
+     *
+     * @copyright Regex credits: https://www.myintervals.com/blog/2009/05/20/iso-8601-date-validation-that-doesnt-suck/
```
ADmad: This is not necessary. That blog post doesn't state any copyright license for the shown.

it-mob: Ok but that doesn't hurt. He did a good job and deserve it.

dereuromark: `@see` is probably the normal link tag right?

it-mob: Ok, that's fair enough :)

PR #13136, opened 2019-04-18
Remove unused constant.

ADmad:
The constant is not used in core anymore.

It's still defined in app skeleton in `config/paths.php` so shouldn't affect app code, though I would like to remove it from app too.

### DIFF
--- a/src/Core/functions.php
+++ b/src/Core/functions.php
@@ -16,13 +16,6 @@ declare(strict_types=1);
 
 use Cake\Core\Configure;
 
-if (!defined('DS')) {
-    /**
-     * Define DS as short form of DIRECTORY_SEPARATOR.
-     */
-    define('DS', DIRECTORY_SEPARATOR);
-}
-
 if (!function_exists('h')) {
     /**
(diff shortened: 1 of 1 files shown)


saeideng (approved):
>It's still defined in app skeleton in config/paths.php

and in `core/tests/bootstrap.php`

dereuromark:
I like the constant to be there, as it can more easily be used also by plugins and other code extending Spryker.
But fair enough if one has to manually define it now.

ADmad:
> and in core/tests/bootstrap.php

Removed from there too now.

ADmad:
@dereuromark I don't mind keeping it in app skeleton.

markstory:
I liked DS for use in tests as it made for shorter cross platform path generation.

ADmad:
Okay, I have reverted the change to tests.

dereuromark:
I reverted it in https://github.com/cakephp/cakephp/pull/13926 as it is too painful also for plugins, you cannot easily add those there across the plugin classes - and usability also suffers here.
No need to overkill cleanup on those convenience things.

PR #13137, opened 2019-04-18
Fix errors reported by static analysers.

ADmad:


markstory:
Thanks :smile: I have to get into the habit of running these locally when doing merges.

PR #13145, opened 2019-04-20
Use Null Coalesce Operator when possible

carusogabriel:
Just a small refactoring to use the Null Coalesce Operator when possible 💅

### DIFF
--- a/src/Http/ServerRequest.php
+++ b/src/Http/ServerRequest.php
@@ -1745,11 +1745,8 @@ class ServerRequest implements ServerRequestInterface
         if (in_array($name, $this->emulatedAttributes, true)) {
             return $this->{$name};
         }
-        if (array_key_exists($name, $this->attributes)) {
-            return $this->attributes[$name];
-        }
 
-        return $default;
+        return $this->attributes[$name] ?? $default;
     }
 
     /**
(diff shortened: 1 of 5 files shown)


src/Http/ServerRequest.php
```diff
@@ -1745,11 +1745,8 @@ public function getAttribute($name, $default = null)
         if (in_array($name, $this->emulatedAttributes, true)) {
             return $this->{$name};
         }
-        if (array_key_exists($name, $this->attributes)) {
-            return $this->attributes[$name];
-        }
 
-        return $default;
+        return $this->attributes[$name] ?? $default;
```
carusogabriel: Should we keep the Early Returns in this case?

ADmad:
Your changes will yield different results if an array key is set to null. That's the reason `array_key_exits()` check is used.

dereuromark:
array_key_exists() int the above cases cannot be replaced as ADmad pointed out.
In cases where null can be a valid value neither isset() or ?? are valid operators as they now change behavior, returning not the desired result.

dereuromark:
That said, I find it alarming how many tests actually passed here :D

ADmad:
@dereuromark Yeah, we do need more tests for keys with null values :blush:.

PR #13146, opened 2019-04-20
Use stricter PHP internal functions.

dereuromark:
When we are now using strict typehinting and especially now the `declare(strict_types=1);` part, it is a bit sad to see that the code internally uses a lot of dirty "auto-casting" inside functions like array_search() and in_array(), potentially cloaking issues still.
I started a first batch here to harmonize the code base towards what is already used in quite a few other places. 

Maybe we also want a sniffer of some sorts in the future, as well.

### DIFF
--- a/src/Auth/BaseAuthenticate.php
+++ b/src/Auth/BaseAuthenticate.php
@@ -130,8 +130,8 @@ abstract class BaseAuthenticate implements EventListenerInterface
             $result->unset($passwordField);
         }
         $hidden = $result->getHidden();
-        if ($password === null && in_array($passwordField, $hidden)) {
-            $key = array_search($passwordField, $hidden);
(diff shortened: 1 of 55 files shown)


ADmad:
> a lot of dirty "auto-casting" inside functions like array_search() and in_array(), potentially cloaking issues still

To be fair in many of those cases the needle and values in haystack are always going to be string so auto casting isn't an issue.

dereuromark:
True :) At least given normal framework input.
If you start extending and adding your own wrappers around, that can not always be guaranteed - and might result in false positives without clear strictness in place.

dereuromark:
One mistake in userland code when calling the core and you get sth like this:
```php
$values = array("apple","orange","pear","grape");
var_dump(in_array(0, $values)); // returns true!
```

Issue #9678, opened 2016-10-31
An empty string should be NULL


steefaan:
I faced this issue multiple times already. You can work around this issue but I think it should be a part of the ORM.

Table:
```
CREATE TABLE categories
(
    id INTEGER PRIMARY KEY NOT NULL,
    name VARCHAR(128) NOT NULL,
    meta_title VARCHAR(255)
);
```

ORM usage:
```
$entity = $this->Categories->newEntity($this->request->data);
$this->Categories->save($entity);
```

Current result:

| id | name | meta_title  |
| -------- |:------------:| ---------:|
| 1 | Test | |

Should be:

| id | name | meta_title  |
| -------- |:------------:| ---------:|
| 1 | Test | _NULL_ |

As I mentioned above you can work around this behavior with `Behaviors/Callbacks` but I think this should be something which is handled directly inside the ORM. Saving an empty string inside a NULLable field isn't the proper usage of `NULL`.

inoas:
I had this issue recently and I think we need a treatment of `NULL` vs `emptyString`.

I am also in favour of the common case of a (truncated) string being empty resulting into NULL by default. If that's not BC both options could be supported and ''-string could then - for now - be the default.

dereuromark:
This is especially important with foreign keys (which are supposed to be nullable if optional), but also with string fields and alike.
If an empty string is posted one would want to have null here saved in most cases.
Otherwise you risk data inconsistency and false results on find().

You either make it not nullable and can use != '' for checking (which I did in 2.x always), or you use null and IS/ISNOT null check. 
But having to allow both means you will end up with both in your app (from different ways of input) - and as a consequence you have always conditions like `where field != '' AND field IS NOT NULL` etc. 
if you ever forget to add both you miss out some records on finding, which introduces bugs.
I have seen so much of that error prone code everywhere in apps..

:+1: For a more core internal solution in the long run, but I am fine with any behavior opt-in for 3.3/3.4.

inoas:
Maybe something along these lines: http://book.cakephp.org/3.0/en/orm/schema-system.html
```php
<?php
    /**
     * Initialize method
     *
     * @param array $config The configuration for the Table.
     * @return void
     */
    public function initialize(array $config)
    {
        parent::initialize($config);

        $this->table('users');
		$this->schema()->alterColumn('profile_id', array(
			'nullStrategy' => 'null'
		));
```

* By default, this should work for strings
[...]

PR #13147, opened 2019-04-20
Marshal empty string to null for BoolType.

ADmad:
Refs #9678


markstory:
Looks good to me, but this is something that should be included in the migration docs.

ADmad:
@markstory Docs PR up.

PR #13148, opened 2019-04-21
Set baseline file for psalm.

ADmad:
This ensures no new errors creep into the code due to ignored rules
and allows chipping away at existing errors easily.


markstory (commented):
Can we add the baseline file to the archive ignore as well? It doesn't need to be part if the zipball payload people install.

ADmad:
@markstory Done.

PR #13150, opened 2019-04-22
Remove invalid chainable typehints as per cs assert.

dereuromark:
Runs https://github.com/cakephp/cakephp-codesniffer/pull/243 over code base and finds necessary fixes.
Chaining must never use typehinting, doing so compromises its use.
As such we must implement the rule immediately, before the 4.x core including strict typehinting becomes more stable and the wrong typehints do more harm here.

The code sniffer asserts that no accidental "other" return types than $this can leak, together with PHPStan (as shown in https://github.com/FriendsOfCake/bootstrap-ui/pull/273/files#diff-f78b0cf313e75baaee6fc88f3a81a8daR71 and the resulting fail report).
Then we dont need to protect it any further using typehinting. As long as PHP doesnt support chainable properly incl those, we cannot use them here.
The alternative is to go away from chaining and use void, but this is usually less useful.

### DIFF
--- a/src/Cache/CacheRegistry.php
+++ b/src/Cache/CacheRegistry.php
@@ -103,7 +103,7 @@ class CacheRegistry extends ObjectRegistry
      * @param string $name The adapter name.
      * @return $this
      */
-    public function unload(string $name): ObjectRegistry
+    public function unload(string $name)
     {
         unset($this->_loaded[$name]);
 
--- a/src/Console/Command.php
+++ b/src/Console/Command.php
@@ -80,7 +80,7 @@ class Command
      * @return $this
      * @throws \InvalidArgumentException
      */
-    public function setName(string $name): self
+    public function setName(string $name)
     {
         if (strpos($name, ' ') < 1) {
             throw new InvalidArgumentException(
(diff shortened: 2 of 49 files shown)


dereuromark:
Failing stan seems unrelated:
```
ERROR: PossiblyNullOperand - src/I18n/FrozenTime.php:232:69 - Cannot concatenate with a possibly null null|string
                    $groupedIdentifiers[$item[0]][$tz] = $item[1] . $abbr;


ERROR: PossiblyNullOperand - src/I18n/FrozenTime.php:234:72 - Cannot concatenate with a possibly null null|string
                    $groupedIdentifiers[$item[0]] = [$tz => $item[0] . $abbr];


ERROR: PossiblyNullOperand - src/I18n/Time.php:312:69 - Cannot concatenate with a possibly null null|string
                    $groupedIdentifiers[$item[0]][$tz] = $item[1] . $abbr;


ERROR: PossiblyNullOperand - src/I18n/Time.php:314:72 - Cannot concatenate with a possibly null null|string
                    $groupedIdentifiers[$item[0]] = [$tz => $item[0] . $abbr];


------------------------------
4 errors found
------------------------------
```

ADmad:
I have fixed the unrelated failures on `4.x`, they were caused by new psalm release.

PR #13151, opened 2019-04-22
Fix inline docblocks.

dereuromark:
Running [InlineDocBlock](https://github.com/spryker/code-sniffer/blob/master/Spryker/Sniffs/Commenting/InlineDocBlockSniff.php) sniff
I recommend adding this to the core one as well to avoid this happening in the future.


ADmad:
Your changes helped fixed some errors earlier reported by pslam and found some new ones. So I tacked on some additional fixes and update psalm's baseline file.

dereuromark:
I didnt expect that one :)

markstory:
Looks good to me. There is one more phpstan error.

ADmad:
I took care of the phpstan error.

PR #13152, opened 2019-04-22
Fix FQCN for exceptions via sniffer run.

dereuromark:
Running [Spryker UseStatement sniff](https://github.com/spryker/code-sniffer/blob/master/Spryker/Sniffs/Namespaces/UseStatementSniff.php)
It seems the used (slevomatik) CS rule is a bit weak and doesnt see all inline usages.
The Spryker one seems to have some of those also found and fixed as seen here.


ADmad:
I don't think we have any slevomat sniff which checks for inline FQCN. We only have one which removed unneeded "use" statements.

dereuromark:
OK, we can check if we can enable one, or port this one used (or parts of it).

PR #13153, opened 2019-04-23
Fix side-effect in destructor

markstory:
SmtpTransport had the potential to create a harmful side-effect in its destructor should an untrusted value ever be deserialized. This solves that by removing the socket on wakeup.

### DIFF
--- a/src/Mailer/Transport/SmtpTransport.php
+++ b/src/Mailer/Transport/SmtpTransport.php
@@ -79,6 +79,18 @@ class SmtpTransport extends AbstractTransport
     }
 
     /**
+     * Unserialize handler.
+     *
+     * Ensure that the socket property isn't reinitialized in a broken state.
+     *
+     * @return void
+     */
+    public function __wakeup()
+    {
+        $this->_socket = null;
+    }
+
+    /**
      * Connect to the SMTP server.
      *
      * This method tries to connect only in case there is no open
--- a/tests/TestCase/Mailer/Transport/SmtpTransportTest.php
+++ b/tests/TestCase/Mailer/Transport/SmtpTransportTest.php
@@ -709,4 +709,25 @@ class SmtpTransportTest extends TestCase
 
         $this->SmtpTransport->send($email);
     }
+
+    /**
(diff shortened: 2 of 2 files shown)


cleptric (approved):

garas (approved):

lorenzo (approved):

stripthis:
Is it possible to get more information about this issue, such as how it could be used in practice? Where would the deserialization take place here? The release notes and the CVE report are a bit terse on the practicalities of an exploit, so it's hard to evaluate the impact for our users.

markstory:
@stripthis If your application accepts serialized data from an untrusted or user provided source, and then deserializes it, your application is vulnerable. An attacker can locally craft a serialized object that replaces the `_socket` property of `SmtpTransport` with a `Cake\Filesystem\File` instance. Then upon destruction `SmtpTransport` will write `QUIT` into the referenced file.

stripthis:
@markstory thanks for your reply. Where would the replacement take place? in SmtpTransport::_generateSocket?

stripthis:
@markstory Got it thanks

carnil:
[CVE-2019-11458](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2019-11458) was assigned for this issue.

Is there any information on when the issue was introduced? For instance Debian ships older versions (e.g. 2.8.5 based version in Debian stretch and 2.10.11 based version in Debian buster).

garas:
Vulnerability was introduced in CakePHP 3.0.0.

In CakePHP 2, SmtpTransport had slightly different implementation, CakePHP 2 also had different autoloader.

carnil:
On Thu, May 09, 2019 at 09:59:38AM +0000, Edgaras Janušauskas wrote:
> Vulnerability was introduced in CakePHP 3.0.0.
> 
> In CakePHP 2, SmtpTransport had slightly different implementation, CakePHP 2 also had different autoloader.

Thanks for the confirmation!

ADmad:
Any one relying of the distro's package manager to get CakePHP instead of using composer or keeping it updated manually is doing a real disservice to themselves anyway.

PR #13154, opened 2019-04-23
4.x psalm

ADmad:


src/View/Helper/UrlHelper.php
```diff
@@ -280,9 +280,10 @@ public function webroot(string $file): string
         $webPath = $request->getAttribute('webroot') . $asset[0];
         $file = $asset[0];
 
-        if (!empty($this->_View->getTheme())) {
+        $themeName = $this->_View->getTheme();
+        if (!empty($themeName)) {
```
dereuromark: Can we use non cloaking, e.g. ! or !== ?

src/View/HelperRegistry.php
```diff
@@ -66,8 +66,9 @@ public function __isset($helper)
         try {
             $this->load($helper);
         } catch (MissingHelperException $exception) {
-            if ($this->_View->getPlugin()) {
-                $this->load($this->_View->getPlugin() . '.' . $helper);
+            $plugin = $this->_View->getPlugin();
+            if (!empty($plugin)) {
```
dereuromark: Could also be now `!== null` right?

src/View/View.php
```diff
@@ -1344,9 +1344,10 @@ public function pluginSplit(string $name, bool $fallback = true): array
     protected function _getLayoutFileName(?string $name = null): string
     {
         if ($name === null) {
+            /** @var string $name */
```
dereuromark: Technically, it could be also the documented false when you submit it via setLayout before calling renderLayout.
I would prefer a clean exception handling in this case here, with meaningful message that this is bonkers to do as dev :)

src/View/Helper/UrlHelper.php
ADmad: Empty string have no value either hence nothing is being cloaked here. The method can only return string or null.

src/View/HelperRegistry.php
ADmad: Same reason as above.

src/View/View.php
ADmad: I added check for `$this->layout` being empty.

src/View/Helper/UrlHelper.php
dereuromark: Cloaking on variable name. If you misspell $themName etc you get a hidden never true.
I guess PHPStan covers this now, so this is not a big deal, but somewhat still a code smell usually to use cloaking checks (silent isset).

src/View/HelperRegistry.php
dereuromark: Thats my point: It can only ever be a non-empty string or null. If it can also contain empty string, then `!$plugin` would be the safer bet.

src/View/View.php
```diff
@@ -1338,17 +1338,20 @@ public function pluginSplit(string $name, bool $fallback = true): array
...
-    protected function _getLayoutFileName(?string $name = null): string
+    protected function _getLayoutFileName(?string $name = null): ?string
     {
         if ($name === null) {
-            /** @var string $name */
             $name = $this->layout;
         }
-        $subDir = '';
 
+        if (empty($name)) {
```
dereuromark: Wouldnt the exception be better? Check line 985 ( $parent = $this->_getLayoutFileName($name); ) where it later tries to then compare false to a non false value and also tries to find an array value to this as key.

dereuromark: So the return value of this should actually never be null IMO. But throw a meaningful exception why the input then was wrong here.

ADmad: No, if `View::$layout` is set to `false` means you don't want a layout to be rendered. That's not an exception case.

ADmad: Practically `renderLayout()` is never called if `View::$layout` is set to false.

src/View/Helper/UrlHelper.php
ADmad: Right, use of uninitialized variables is already reported by phpstan and psalm so it's not an issue.

src/View/View.php
dereuromark: Exactly my point: No need to then make this nullable return if the input should not happen and should throw an exception instead to show a developer error instead of hiding this to blow up later in code.

dereuromark: Could be InvalidArgumentException, LogicException or RuntimeException.

ADmad: And yet `renderLayout()` does check the return value `_getLayoutFileName()` :slightly_smiling_face: and bails out if it's empty, that's why I added the `null` return.

dereuromark: not extend() and what I just explained there happening in code.
Also not, if you extended/replaced the code and badly order/code the execution chain of these protected methods.
Thus: better safe than sry IMO. And with a clear exception message. The in_array check with false as key does not seem safe/sane to me.

ADmad: `extend()` is always called with a layout name which is passed to `_getLayoutFileName()` hence `View::$layout` is never used in that case.

I really don't spend more time on this. With current code there's no code path which generates an error in actual usage. Feel free to make changes you see fit later on if you feel strongly about it.

dereuromark: I just dont see the need or the intention of making _getLayoutFileName() nullable return.
We shouldnt weaken the API for no reason, that's all.

dereuromark: I am fine cleaning this up with a follow up commit if I may.

psalm-baseline.xml
```diff
@@ -3327,6 +3327,9 @@
     </UninitializedProperty>
   </file>
   <file src="src/View/View.php">
+    <PossiblyFalseArgument occurrences="1">
+      <code>$name</code>
+    </PossiblyFalseArgument>
```
dereuromark: :(

src/View/View.php
ADmad: I have reverted the changes to `_getLayoutFileName()`.

dereuromark: I am a bit how this discussion and the resolution went.

psalm-baseline.xml
dereuromark: I opened a PR to this one.

dereuromark (approved):

PR #13155, opened 2019-04-23
Clean up the expected input scope.

dereuromark:
I might have been misunderstood in terms of how to resolve this in the cleanest way without the need to suppress stan checks.
=> https://github.com/cakephp/cakephp/pull/13154 addon

### DIFF
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -3327,9 +3327,6 @@
     </UninitializedProperty>
   </file>
   <file src="src/View/View.php">
-    <PossiblyFalseArgument occurrences="1">
-      <code>$name</code>
-    </PossiblyFalseArgument>
     <PossiblyNullPropertyAssignmentValue occurrences="4">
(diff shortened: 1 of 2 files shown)


dereuromark:
If stan does not like this, we can also first assign and then, and then use $name to check.

ADmad:
The root cause of the problem is `View::$layout` is used to hold the layout file name and well as control whether layout should be rendered. Ideally there should be a separate property/method to control layout rendering.

ADmad:
If a minor BC break is acceptable we can just disallow setting layout to false. One can instead just use `ViewBuilder::disableAutoLayout()` / `View::disableAutoLayout()` to turn off layout rendering.

dereuromark:
I agree on the ideal case. But until then now, and cleanup can be done I think this is the best way forward without hiding issues. Especially since they can only occur if you really coded/extended it in a bad and unexpected way.

ADmad (approved):

PR #13157, opened 2019-04-24
Remove checks for params/keys no longer used in routing.

ADmad:
They are a relic of Cake 2's URL format and the defunct "request action" feature.


markstory:
Awesome stuff :clap:

PR #13158, opened 2019-04-24
Don't allow View::$layout to be set to false.

ADmad:
View::$autoLayout can be used to control layout rendering.
This allows for cleaner code.

### DIFF
--- a/src/View/View.php
+++ b/src/View/View.php
@@ -133,7 +133,7 @@ class View implements EventDispatcherInterface
      * is the filename of the layout in /templates/Layout without the .php
      * extension.
      *
-     * @var string|false
+     * @var string
      */
     protected $layout = 'default';
 
@@ -549,9 +549,9 @@ class View implements EventDispatcherInterface
      * The name specified is the filename of the layout in /templates/Layout
      * without the .php extension.
      *
-     * @return string|false
+     * @return string
      */
-    public function getLayout()
+    public function getLayout(): string
     {
         return $this->layout;
     }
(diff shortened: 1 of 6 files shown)


src/View/View.php
```diff
@@ -707,12 +711,15 @@ public function render($view = null, $layout = null): string
...
 
-        if ($this->layout && $this->autoLayout) {
+        if ($this->autoLayout) {
             $this->Blocks->set('content', $this->renderLayout('', $this->layout));
         }
         if ($layout !== null) {
             $this->layout = $defaultLayout;
         }
+        if ($defaultAutoLayout !== null) {
+            $this->autoLayout = $defaultAutoLayout;
```
markstory: If the view rendering raises an exception do we need to worry about resetting this state? Could we use local variables instead of modifying the instance state?

ADmad: One could check and/or modify this property from view file too. Of we just use a local variable one would no longer be able to turn off layout rendering from view which is possible right now by setting layout to false instead.

Also in case of an exception the exception renderer will use a new View instance isn't it? So we don't have to worry about resetting state.

src/View/View.php
```diff
@@ -707,12 +711,15 @@ public function render($view = null, $layout = null): string
             $this->dispatchEvent('View.afterRender', [$viewFileName]);
         }
 
-        if ($this->layout && $this->autoLayout) {
+        if ($this->autoLayout) {
```
markstory: If users are supposed to be able to set `$this->layout = false` in templates shouldn't this also check the layout property?

ADmad: No, one could set `$this->layout = false`  earlier but with this change the alternative to disable layout rendering is use `$this->autoLayout = false`.

Not allowing `View::$layout` to be set to false is the whole point of this PR. Now it showed always contain a string layout name and controlling of layout rendering should be done using `View::$autoLayout` only.

PR #13161, opened 2019-04-24
Remove charset on json content-type

### LINKED ISSUE #13156: [3.x] Chrome throwing CORB (Cross-Origin-Read-Blocking) warnings for default JSON header
It seems [Chrome has recently started](https://stackoverflow.com/questions/55619729/corb-options-requests-blocked-in-chrome-73) to throw CORB (Cross-Origin-Read-Blocking) warnings when the encoding is part of the HTTP content-type header, which is the [default behavior of CakePHP 3.x](https://github.com/cakephp/cakephp/search?q=%22application%2Fjson%3B+charset%3Dutf-8%22&unscoped_q=%22application%2Fjson%3B+charset%3Dutf-8%22) (possibly CakePHP 2.x, too).

Instead of ``application/json; charset=utf-8``, we should simply return ``application/json`` without the ``; charset=utf-8`` suffix.

Slim had the same issue:
https://github.com/slimphp/Slim/issues/2629

What IANA says about the JSON content-type header:

> Note:  No "charset" parameter is defined for this registration.
   Adding one really has no effect on compliant recipients.

https://www.iana.org/assignments/media-types/application/json

Some related links:
https://www.chromium.org/Home/chromium-security/corb-for-developers
https://medium.com/@ethicalevil/nosniff-and-the-rabbit-hole-of-mime-sniffing-in-browsers-9f764a454a46
https://www.searchenginejournal.com/nosniff-response-headers/261171/#close
https://www.chromestatus.com/feature/5780195579527168
https://stackoverflow.com/questions/18337630/what-is-x-content-type-options-nosniff
https://developer.mozilla.org/de/docs/Web/HTTP/Headers/X-Content-Type-Options
https://book.cakephp.org/3.0/en/controllers/middleware.html#security-header-middleware

Edit: accept -> content-type :coffee:

markstory:
Don't emit a charset parameter on the json content type. Chrome has started emitting warnings when this parameter is present as it isn't compliant with the spec.

Fixes #13156

### DIFF
--- a/src/Http/Response.php
+++ b/src/Http/Response.php
@@ -581,7 +581,7 @@ class Response implements ResponseInterface
             return;
         }
         $whitelist = [
-            'application/javascript', 'application/json', 'application/xml', 'application/rss+xml'
+            'application/javascript', 'application/xml', 'application/rss+xml'
         ];
 
         $charset = false;
--- a/tests/TestCase/Error/Middleware/ErrorHandlerMiddlewareTest.php
+++ b/tests/TestCase/Error/Middleware/ErrorHandlerMiddlewareTest.php
@@ -167,7 +167,7 @@ class ErrorHandlerMiddlewareTest extends TestCase
         $this->assertNotSame($result, $response);
         $this->assertEquals(404, $result->getStatusCode());
         $this->assertContains('"message": "whoops"', '' . $result->getBody());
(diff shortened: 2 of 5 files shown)


ravage84:
@markstory fixed the broken tests.

Issue #12549, opened 2018-09-09
Command class is missing a dispatchCommand method


suhaboncukcu:
This is a  
* [x] feature-discussion (RFC)

* CakePHP Version: 3.6
* Platform and Target: N\A

### What you did
I was able to dispatch a shell within another shell as 
`$this->dispatchShell('schema', 'create', 'Blog', '--plugin', 'Blog');`

Since the class is deprecated I tried to achieve the same thing with console commands. 

### What happened
The `dispatchCommand` method is missing from the `Command` class. 

### What you expected to happen
I usually separate commands in granular classes and code a few commands to use them in different aspects according to arguments.  
As an example;

```php 
class DeployCommand extends Command
{
    public function buildOptionParser(ConsoleOptionParser $parser)
    {
        $parser = parent::buildOptionParser($parser);
        $parser
            ->addOption('type', [
                'help' => 'type of deployment',
                'required' => true,
                'choices' => [
                    'cms', 'app'
                ]
            ]);

        return $parser;
    }

    public function execute(Arguments $args, ConsoleIo $io)
    {
         if($args->getOption('type') === 'cms') {
            // call deploy cms command here
        } else {
            // call deploy cms here
            // call deploy app also here
        }
    }
}

```

So it would be useful to have  a `dispatchCommand` method just like `dispatchShell` method.

markstory:
Would you want to call other commands using an array of arguments or a string as if you were using the cli?

robertpustulka:
Instead of dispatching commands from within commands I'd suggest refactoring your code so the reusable logic is not tied to the command anymore.

Dispatching commands from another command doesn't sound like a good idea and we should avoid that.

markstory:
@robertpustulka That is sound advice :+1:

suhaboncukcu:
I can't follow the logic behind this. I have commands. They are not sub-commands. They do work separately  as they should. There are some commands firing other ones with some logic.  

For instance;
I have a command running every night. 
It checks the database and fires a command with different arguments according to the count of visitor countries. But I should be able to use that command with whatever argument I choose during the day since I need some daily reports time to time. 

Within one of my old projects; I had a plugin including different shells for different cron jobs. A daily shell and an hourly shell. So I was able to manage cron jobs within my software layer easily. 

[...]

PR #13163, opened 2019-04-25
Add Command::executeCommand()

markstory:
This method enables commands to dispatch each other without needing to use the CommandRunner. This provides commands with the ability to be composed from other commands.

Refs #12549

### DIFF
--- a/src/Console/Command.php
+++ b/src/Console/Command.php
@@ -258,7 +258,7 @@ class Command
      * @param \Cake\Console\ConsoleIo $io The ConsoleIo instance to use for the executed command.
      * @return int|null The exit code or null for success of the command.
      */
-    public function executeCommand($command, array $args = [], ?ConsoleIo $io = null)
+    public function executeCommand($command, array $args = [], ConsoleIo $io = null)
     {
         if (is_string($command)) {
             if (!class_exists($command)) {
--- a/src/Console/Command.php
+++ b/src/Console/Command.php
@@ -249,4 +249,34 @@ class Command
     {
         throw new StopException('Command aborted', $code);
     }
+
+    /**
+     * Execute another command with the provided set of arguments.
+     *
+     * @param string|\Cake\Console\Command $command The command class name or command instance.
+     * @param \Cake\Console\ConsoleIo $io The ConsoleIo instance to use for the executed command.
+     * @param array $args The arguments to invoke the command with.
+     * @return int|null The exit code or null for success of the command.
+     */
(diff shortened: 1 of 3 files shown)


themrwilliams (approved):

src/Console/Command.php
```diff
@@ -249,4 +249,34 @@ public function abort($code = self::CODE_ERROR)
...
+
+    /**
+     * Execute another command with the provided set of arguments.
+     *
+     * @param string|\Cake\Console\Command $command The command class name or command instance.
+     * @param \Cake\Console\ConsoleIo $io The ConsoleIo instance to use for the executed command.
+     * @param array $args The arguments to invoke the command with.
+     * @return int|null The exit code or null for success of the command.
+     */
+    public function executeCommand($command, ConsoleIo $io, array $args = [])
```
ADmad: Can the method name and arguments order be harmonized with `CommandRunner::runCommand()`?
Could the `$io` argument be made optional and internally create a new instance if none provided?

markstory: I could do that. Making a new instance will do bad things in tests which is why I made it a required argument.

markstory: Done now :+1:

src/Console/Command.php
```diff
@@ -249,4 +249,35 @@ public function abort($code = self::CODE_ERROR)
...
+
+    /**
+     * Execute another command with the provided set of arguments.
+     *
+     * @param string|\Cake\Console\Command $command The command class name or command instance.
+     * @param array $args The arguments to invoke the command with.
+     * @param \Cake\Console\ConsoleIo $io The ConsoleIo instance to use for the executed command.
+     * @return int|null The exit code or null for success of the command.
+     */
+    public function executeCommand($command, array $args = [], ?ConsoleIo $io = null)
```
ADmad: `?ConsoleIo` typehint won't work in PHP < 7.1 :stuck_out_tongue:.

dereuromark: Dropping the ? should be fine here :)

markstory: Derp. I forgot that this was for 3.x

markstory:
> Fatal error: Declaration of Cake\TestSuite\TestCase::setUp() must be compatible with PHPUnit\Framework\TestCase::setUp(): void in /home/travis/build/cakephp/cakephp/src/TestSuite/TestCase.php on line 36

Going to rollback the PHPUnit version until we can make a separate effort to do that.

ADmad (approved):

PR #13164, opened 2019-04-25
4.x view cleanup

ADmad:
Cleaned up check for defunct view options and harmonized the use of "template" instead of "view".

### DIFF
--- a/src/View/View.php
+++ b/src/View/View.php
@@ -676,14 +676,14 @@ class View implements EventDispatcherInterface
      * a plugin template/layout can be used instead of the app ones. If the chosen plugin is not found
      * the template will be located along the regular view path cascade.
      *
-     * @param string|false|null $template Name of template file to use
+     * @param string|null $template Name of template file to use
      * @param string|false|null $layout Layout to use. False to disable.
      * @return string Rendered content.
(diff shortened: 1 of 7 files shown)


ADmad:
Another change I would like to do is to now allow passing `false` to `$template` argument of `View::render()` method.

I don't see a practical use case where that would be useful. `Controller::render()` and `ViewBuilder::setTemplate()` don't accept `false` as valid value for the template name either.

src/View/View.php
```diff
@@ -703,12 +695,12 @@ public function render($view = null, $layout = null): string
             $this->layout = $layout;
         }
 
-        $viewFileName = $view !== false ? $this->_getViewFileName($view) : null;
-        if ($viewFileName) {
+        $templateFileName = $template !== false ? $this->_getTemplateFileName($template) : null;
```
markstory: We should document this protected method change. It will impact a few plugins.

ADmad: Hmm.. The migration guide will get bloated if we start documenting changes to protected props/methods too. Once they get a `method not found` error it shouldn't be too difficult for anyone to figure out the new method name to use :slightly_smiling_face:.

markstory: Ok. We can always add it later if it becomes an issue.

Issue #13103, opened 2019-04-09
Raise minimum version of PHP to 7.2 for 4.0


markstory:
We are currently planning to release CakePHP 4.0 requiring at least PHP 7.1. However, PHP 7.1 will become [end-of-life at the end of this year](https://www.php.net/supported-versions.php). Given that we are about to start the release process for 4.0, we have the opportunity to raise the minimum PHP version to 7.2.

We'd like to gather feedback from the community on which version we should require. 

* Add a :+1: to vote for PHP 7.2
* Add a :-1: to vote for PHP 7.1

dmcdaniel12:
Ideally it would be 7.3 but 7.2 EOL isn't until next year so plenty of time to have this discussion again in a few months :)

Bird87ZA:
Was literally about to post what @dmcdaniel12 posted. My feeling - even though I understand that EOL isn't until next year - is that new major releases should require newest available versions where possible. Since the step up between 7.2 and 7.3 isn't huge, I feel like you might just as well require 7.3

AshleyPinner:
I wouldn't require 7.3 just yet as getting it in even 3rd party repos isn't quite there yet (webtatic as an example stops at 7.2). If the step between 7.2 and 7.3 isn't huge then there's no reason to require 7.3 either, the sword cuts both ways.

If you can argue that there's a feature of 7.3 that Cake4 could make great use of then the discussion might be a bit different, but given the current facts of 7.1's impending EoL, requiring 7.2 is a no-brainer. Moving that requirement to 7.3 is absurd at this point though with 7.2 not being EOL for a long while yet.

Bird87ZA:
This is true. As you say, it cuts both ways. And I didn't think about 3rd party repos (amateur, I know)

Let's then have this discussion again in a few months.

ADmad:
Just being the devils advocate here: The decision about what min. PHP version framework should support should be based on the benefits the core and app code can achieve by bumping to higher version, not just EOL of PHP version. We know that in the *real* world there are always systems which are forced to continue using older PHP version for X reason.

Checking the [new](https://www.php.net/manual/en/migration72.new-features.php) feature of PHP 7.2, off top of my head I don't see any being particularly helpful in core code. The new `object` type hinting and parameter widening would still be usable in app code if one deploys on PHP 7.2.

[...]

PR #13165, opened 2019-04-26
Make PHP 7.2 the minimum required version for 4.x

markstory:
After a healthy discussion the consensus seems to be that requiring 7.2 is reasonable given that 7.1 will soon be end of life and that all major distros already have PHP7.2 or will very soon.

Refs #13103

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,12 +24,6 @@ matrix:
   fast_finish: true
 
   include:
-    - php: 7.1
-      env: CODECOVERAGE=1 DB=mysql db_dsn='mysql://root@127.0.0.1/cakephp_test?init[]=SET sql_mode = "STRICT_TRANS_TABLES,NO_ZERO_IN_DATE,NO_ZERO_DATE,ERROR_FOR_DIVISION_BY_ZERO,NO_AUTO_CREATE_USER,NO_ENGINE_SUBSTITUTION"'
-    - php: 7.1
-      env: CODECOVERAGE=1 DB=pgsql db_dsn='postgres://postgres@127.0.0.1/cakephp_test'
-    - php: 7.1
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -48,12 +42,7 @@ before_install:
   - phpenv config-rm xdebug.ini
 
   - if [ $DB = 'mysql' ]; then mysql -u root -e 'CREATE DATABASE cakephp_test;'; fi
-  - if [ $DB = 'mysql' ]; then mysql -u root -e 'CREATE DATABASE cakephp_test2;'; fi
```
ADmad: The additional databases aren't needed anymore?

ADmad:
phpunit should also be bumped to v8 which is PHP 7.2+. Sticking with an older version could bite use in the future, we already had enough trouble support both phpunit 5.7 and 6 for Cake 3.

.travis.yml
markstory: Nope. I don't think we ever used them even in 3.x

ADmad: Huh, I always thought they were for cross db tests.

ADmad:
Zoinks! Updating to phpunit v8 broke apart our test suite :stuck_out_tongue:.

.travis.yml
markstory: They were, but I don't think we've had those in Cake 3.

markstory:
> Fatal error: Declaration of Cake\TestSuite\TestCase::setUp() must be compatible with PHPUnit\Framework\TestCase::setUp(): void in /home/travis/build/cakephp/cakephp/src/TestSuite/TestCase.php on line 36

Going to rollback the PHPUnit version until we can make a separate effort to do that.

markstory:
Appveyor is out to lunch again, but otherwise builds look good. Are there any other changes that need to be made right now?

lorenzo:
Looks good to me

PR #13166, opened 2019-04-26
Add Response::withCookieCollection().

ADmad:
Refs #12551.

The new method can be added in 3.8 itself instead of 4.0.

### DIFF
--- a/src/Http/Response.php
+++ b/src/Http/Response.php
@@ -2388,6 +2388,20 @@ class Response implements ResponseInterface
     }
 
     /**
+     * Get a new instance with provided cookie collection.
+     *
+     * @param \Cake\Http\Cookie\CookieCollection $cookieCollection Cookie collection to set.
+     * @return static
+     */
+    public function withCookieCollection(CookieCollection $cookieCollection)
+    {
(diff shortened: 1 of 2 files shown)


src/Http/Response.php
```diff
@@ -2387,6 +2387,20 @@ public function getCookieCollection()
         return $this->_cookies;
     }
 
+    /**
+     * Get a new instance with provided cookie collection.
+     *
+     * @param \Cake\Http\Cookie\CookieCollection $cookieCollection Cookie collection to set.
+     * @return static
+     */
+    public function withCookieCollection(CookieCollection $cookieCollection)
```
markstory: Should this hint be for the interface?

ADmad: `CookieCollection` doesn't implement any interface besides `IteratorAggregate` and `Countable`.

markstory: My bad. I thought we had interfaces.

markstory:
Docs updated in cakephp/docs@64781fe0a4d4b55a452fbc3144309bff0c87ea01

PR #13168, opened 2019-04-27
Generate column SQL for char(36) in postgres schema

### LINKED ISSUE #13167: Postgres Schema: CHAR(36) results in CHAR(1) columns
This is a:

* [x] bug
* [ ] enhancement

* CakePHP Version: 3.6.7
* Platform and Target: PHP 7.2.17, **PostgreSQL 10.7** (Ubuntu 10.7-0ubuntu0.18.10.1)

### What you did
Fixtures using `$import` on tables that contain a `char(36)` column are created incorrectly as a `char` (which results in `char(1)`).

### What happened
Inserting fixture data generates error:
```
ERROR:  value too long for type character(1)
```

I expected a char column with a size of 36 to be created.

We are using postgres UUID columns in the app and that works as expected, but we are using `char(36)` for another non-standard ID.

For now I'll just resize our columns to make it work.

It can be fixed by removing the second part of the conditional [here](https://github.com/cakephp/cakephp/blob/13ab4d788ce553056938b60b3039197b8edd9993/src/Database/Schema/PostgresSchema.php#L402).

```
-- Part of the generated SQL: 
CREATE TABLE "dev" (
    "id" BIGSERIAL,
    "dev_id" CHAR DEFAULT NULL, -- This here should be CHAR(36)
    PRIMARY KEY ("id")
)

markstory:
Don't omit length for 36 byte length CHAR columns. This is likely an artifact of how UUID columns were handled in the past.

Fixes #13167

### DIFF
--- a/src/Database/Schema/PostgresSchema.php
+++ b/src/Database/Schema/PostgresSchema.php
@@ -399,7 +399,7 @@ class PostgresSchema extends BaseSchema
                 $type = ' CHAR';
             }
             $out .= $type;
-            if (isset($data['length']) && $data['length'] != 36) {
+            if (isset($data['length'])) {
                 $out .= '(' . (int)$data['length'] . ')';
             }
         }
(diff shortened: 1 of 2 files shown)


commercial-hippie:
Thanks for the quick fix!

PR #13171, opened 2019-04-29
Improved doc block

bancer:
Session value can be not only string or array.


dereuromark:
True, technically you can also store e.g. int, float or alike.
I am just wondering if they stay this type if they come out of the session again.

bancer:
The data types stored in session are preserved per my knowledge.

Issue #13170, opened 2019-04-28
Return what is actually stored cake 3 and 4


jimgwhit:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: EXACT RELEASE VERSION OR COMMIT HASH, HERE.
* Platform and Target: YOUR WEB-SERVER, DATABASE AND OTHER RELEVANT INFO AND HOW THE REQUEST IS BEING MADE, HERE.

### What you did
To Cakephp team:

First see:  https://discourse.cakephp.org/t/cake-php-3-7-date/6001/6

I just tested the same query in laravel 5.8 and cakephp 4:

### cakephp 4

```
    public function itest()
    {
        $this->autoRender = false;
        $petid = 9;
        $pet = TableRegistry::getTableLocator()->get('Pets');
            $data = $pet
                    ->find()
                    ->where(['petid' => $petid])
                    ->first();
            $testdate = $data->odate;
            $json = json_encode($testdate);
            echo $json;
            exit();
    }

}

```

### Returns:
```ruby
"1998-09-02T00:00:00+00:00"
```

### Laravel 5.8

```
    public function itest()
    {
        $petid = 9;
        $pet = Pet::find($petid);
        $testdate = $pet->odate;
        $json = json_encode($testdate);
        echo $json;
        exit();
    }
```

### returns
```ruby
"1998-09-02"
```

Why is the orm and query builder manipulated to return what you wanted it to return instead of a value that is stored.

I am not trying to cut anyone down, but I just don't understand the core team manipulating users stored data.

In this case 1998-09-02 is what's actually stored.

And I am not new to this stuff, I am currently number 14 on laravels top 50, I have used it since ver 4.2, and have used cakephp for some apps.

Why do you determine what's returned instead of just letting mysql and pdo do their job.

### What you expected to happen
EXPLAIN WHAT IS TO BE EXPECTED, HERE.

Cakephp let pdo and mysql do their job, and not return a format you decide on.

How did you decide on that anyway?

josegonzalez:
As a convenience feature, the ORM maps column types to specific classes. That is done automatically, though if the user wishes, they may specify a specific type for a given column.

In this case, we appear to be mapping the `datetime` column to a [DateTimeType](https://github.com/cakephp/cakephp/blob/master/src/Database/Type/DateTimeType.php) instance.

[...]

PR #13172, opened 2019-04-29
Return only the date part when JSON encoding Date/FrozenDate objects.

ADmad:
Refs #13170.

### DIFF
--- a/src/I18n/Date.php
+++ b/src/I18n/Date.php
@@ -46,7 +46,7 @@ class Date extends MutableDate implements I18nDateTimeInterface
     protected static $_toStringFormat = [IntlDateFormatter::SHORT, -1];
 
     /**
-     * The format to use when when converting this object to json
+     * The format to use when converting this object to JSON.
      *
      * The format should be either the formatting constants from IntlDateFormatter as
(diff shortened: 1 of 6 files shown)


src/I18n/Date.php
```diff
@@ -45,6 +45,22 @@ class Date extends MutableDate implements I18nDateTimeInterface
      */
     protected static $_toStringFormat = [IntlDateFormatter::SHORT, -1];
 
+    /**
+     * The format to use when when converting this object to json
```
dereuromark: JSON is probably an acronym here.

ADmad: I simply cut - pasted this from the `DateTimeTrait`. There's a double "when" too in the statement :)

saeideng: `... when when ...`

markstory:
Seems reasonable but also something that we'll need to document in the migration guide as it will change the output and behavior of people's public APIs.

ADmad:
@markstory Already opened a PR for the migration guide to :)

ADmad:
@markstory Good to merge?

dereuromark:
He already merged docs one anyway :)

PR #13173, opened 2019-04-29
Document correct types for $modelClass

dereuromark:
I was working on a controller that doesnt need a model.
This clarifies the types here - also needed for [IdeHelper](https://github.com/dereuromark/cakephp-ide-helper) to detect annotations that need adding.

### DIFF
--- a/src/Datasource/ModelAwareTrait.php
+++ b/src/Datasource/ModelAwareTrait.php
@@ -35,7 +35,10 @@ trait ModelAwareTrait
      * Plugin classes should use `Plugin.Comments` style names to correctly load
      * models from the correct plugin.
      *
-     * @var string
+     * Use false to not use auto-loading on this object. Null auto-detects based on
+     * controller name.
+     *
+     * @var string|false|null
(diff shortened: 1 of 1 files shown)


src/Datasource/ModelAwareTrait.php
```diff
@@ -35,7 +35,10 @@ trait ModelAwareTrait
      * Plugin classes should use `Plugin.Comments` style names to correctly load
      * models from the correct plugin.
      *
-     * @var string
+     * Use false to not use auto-loading on this object. Null auto-detects based on
+     * controller name.
+     *
+     * @var string|false|null
```
ADmad: I have the feeling the static analyses will be unhappy with this in 4.x.

dereuromark: We can use empty string and strict === '' check here for fasely value.

dereuromark: I can fix that up after merge for 4.x as target.

PR #13174, opened 2019-04-29
4.x - Update to phpunit 8

ADmad:
I had started working on this yesterday :slightly_smiling_face:.

Only warnings like the ones below remain. I am not sure about their replacement and they don't cause builds to fail so I left them for now.

```
attributeEqualTo() is deprecated and will be removed in PHPUnit 9.
attribute() is deprecated and will be removed in PHPUnit 9.
readAttribute() is deprecated and will be removed in PHPUnit 9.
getObjectAttribute() is deprecated and will be removed in PHPUnit 9.
```


ADmad:
Psalm is unhappy with the use of various method/classes marked as internal. Just ignoring those for now.

https://scrutinizer-ci.com/g/cakephp/cakephp/inspections/3af03b26-556a-4307-8d57-820ecf3969f1

ADmad:
The static analysis job failure is unrelated and fixed on `4.x`.

dereuromark:
I fixed the phpstan inspection error in https://github.com/cakephp/cakephp/pull/13183/commits/155a22d9efda4d36516b42a78a3b454aff133bf1

tests/TestCase/Command/CacheCommandsTest.php
```diff
@@ -42,7 +42,7 @@ public function setUp()
      *
      * @return void
      */
-    public function tearDown()
+    public function tearDown(): void
```
markstory: Do we have sniffer rules that do this work or should we plan to add this to rector when we get to that point?

dereuromark: Strict one of slevomatik could add those, but if enabled it also might add ones we dont want.
Better to run this manually from time to time maybe.

markstory: I was more thinking about the upgrade path for applications and plugins, as I think all those builds/tests will now fail.

ADmad: There are [rectors](https://github.com/rectorphp/rector/blob/master/docs/AllRectorsOverview.md#phpunit) available for updating the various assert methods.

PR #13175, opened 2019-04-30
4.x Update Cache shell to commands

markstory:
Convert the cache shell commands to command classes. I wanted to get any feedback on this approach before fixing up the completion shell tests (which would break with bake anyways)

I've converted the shell tasks for `CacheShell` into a set of command classes, and manually mapped the command names as the automatic discovery isn't smart enough to figure out related commands. I wanted to see if anyone had thoughts on a convention we could use for the core and potentially application commands that would allow us to indicate command name groups using conventions.


markstory:
Outside of the clunky manually defined class map that @saeideng brought up in chat is there anything else that folks would like to see fixed?

I can put together a conventions based approach to converting class names to command names for the CakePHP provided commands as a separate pull request.

PR #13177, opened 2019-04-30
Update ErrorHandlerMiddleware.php

cleptric:
`Throwable` is the base interface for `Error` and `Exception`, so this part of the code is never reached.


dereuromark (changes requested):
:-1: 
It is in PHP5. Thus both

cleptric:
True, already fixed in `4.x` so I'll close 😄

PR #13178, opened 2019-04-30
Use stricter handling of model class check.

dereuromark:
Follow https://github.com/cakephp/cakephp/pull/13173

loadModel() without argument loads the $modelClass which, if set to empty string
would run into a cloaked error later on.
As such it is best to catch early.

Also doc block nullable part was removed from master merge, I restored it.

### DIFF
--- a/src/Console/Command.php
+++ b/src/Console/Command.php
@@ -64,7 +64,7 @@ class Command
             return $this->getTableLocator()->get($alias);
         });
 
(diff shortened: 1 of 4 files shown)


markstory (approved):
Looks good to me.

markstory:
Build failures look relevant.

dereuromark:
Closing in favor of https://github.com/cakephp/cakephp/pull/13179

PR #13179, opened 2019-05-01
Make ModelAwareTrait::$modelClass protected.

ADmad:
Allow setting the property to empty string in controller to indicate
that controller does not use a default model.

Alternate to #13178.

### DIFF
--- a/tests/TestCase/Datasource/ModelAwareTraitTest.php
+++ b/tests/TestCase/Datasource/ModelAwareTraitTest.php
@@ -32,10 +32,10 @@ class ModelAwareTraitTest extends TestCase
     public function testSetModelClass()
     {
         $stub = new Stub();
-        $this->assertNull($stub->modelClass);
+        $this->assertNull($stub->getProp('modelClass'));
 
         $stub->setProps('StubArticles');
-        $this->assertEquals('StubArticles', $stub->modelClass);
+        $this->assertEquals('StubArticles', $stub->getProp('modelClass'));
     }
 
     /**
--- a/tests/TestCase/Datasource/ModelAwareTraitTest.php
+++ b/tests/TestCase/Datasource/ModelAwareTraitTest.php
@@ -32,10 +32,10 @@ class ModelAwareTraitTest extends TestCase
     public function testSetModelClass()
     {
         $stub = new Stub();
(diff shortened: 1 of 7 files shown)


ADmad:
I can add a getter method for the property if deemed necessary.

tests/test_app/TestApp/Stub/Stub.php
```diff
@@ -15,4 +15,9 @@ public function setProps($name)
     {
         $this->_setModelClass($name);
     }
+
+    public function getProp($name)
```
dereuromark: Should we typehint string here too?

ADmad: This is a test case, not worth bothering.

dereuromark: setProps($name) => setProps(?string $name)

dereuromark: My logic is: If it is used wrong, then the expected results are also not what you want.
So probably worth also in those cases. But no biggie.

dereuromark:
I checked. After this one we still have 61 public properties left

ADmad:
The scrutinizer failure is irrelevant, I will fix them on `4.x`.

tests/TestCase/Datasource/ModelAwareTraitTest.php
```diff
@@ -32,10 +32,10 @@ class ModelAwareTraitTest extends TestCase
     public function testSetModelClass()
     {
         $stub = new Stub();
-        $this->assertNull($stub->modelClass);
+        $this->assertNull($stub->getProp('modelClass'));
 
         $stub->setProps('StubArticles');
-        $this->assertEquals('StubArticles', $stub->modelClass);
+        $this->assertEquals('StubArticles', $stub->getProp('modelClass'));
```
markstory: You could use `assertAttributeEquals` to avoid having to build your own property reader.

ADmad: Done

dereuromark: Shouldnt we using Same here? This is safer and will be used more common with scalar comparison.

dereuromark (approved):

Issue #13181, opened 2019-05-01
Translate behavior breaks loadInto for belongsTo associations


Zuluru:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.7.7

### What you did
I'm implementing the Translate behavior on some tables.

### What happened
When `loadInto` is used to load a `belongsTo` association, the property is set to `null` instead of the entity. It's not a passive failure of some sort; if the property was otherwise set before, it is *changed to* null.

### What you expected to happen
The property should be set to the associated (and translated) entity, exactly as it is if it's `contain`ed in the initial read. Note that the `loadInto` method works fine for `hasMany` associations.

robertpustulka:
I can confirm the issue. Lazy loading an entity associated as belongsTo with Translate behavior attached to the table sets the value to null. 

For some reason enabling auto fields in `LazyEagerLoader::_getQuery()` method resolved an issue for me. I'll try to apply this fix in #13182 and see if it helps.

othercorey:
> For some reason enabling auto fields in `LazyEagerLoader::_getQuery()` method resolved an issue for me.

Isn't this because that query is only selecting the primary key?

`->select((array)$primaryKey)`

robertpustulka:
@othercorey it's ok to select primary key only on a source table. The associated records that are required to load into the entity are configured using `contain()`.

PR #13182, opened 2019-05-01
Add unit test demonstrating failure of loadInto with Translate behavior

Zuluru:
This pull request adds a unit test that demonstrates issue #13181.


markstory:
Thanks for the failing test case :+1:

PR #13183, opened 2019-05-02
Fix doc blocks, add object typehint and remove static typehints.

dereuromark:
- With php7.2 we should be able to use object typehint
- return static has the same issues as return $this (chainable), making added methods in the extended classes not usable/seeable with those typehints added. Better we add a sniffer for this too. Note that already a lot of static returns are properly declared, so this is consolidating the differences to a single standard.
- I only didnt touch Response class yet, we should fix that too then afterwards
- Small fixes around doc blocks
- `public function unserialize($data): void` should always be void as per PHP specs.

Part 1/2

### DIFF
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -29,7 +29,6 @@ parameters:
         - '#Return type \(null\) of method Cake\\Collection\\Iterator\\NoChildrenIterator::getChildren\(\) should be compatible with return type \(RecursiveIterator\) of method RecursiveIterator::getChildren\(\)#'
         - '#Call to an undefined method DateTimeInterface::setTimezone\(\)#'
         - '#Call to an undefined method Cake\\Datasource\\QueryInterface::eagerLoaded\(\)#'
-        - '#Property Cake\\Mailer\\Transport\\SmtpTransport::\$_socket \(Cake\\Network\\Socket\) does not accept null#'
         -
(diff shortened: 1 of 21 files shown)


dereuromark:
Scrut:

> Comment missing for @throws tag in function comment

Since when do we need comments for throws tags? We shoudnt enforce this verbosity IMO. Any wy to disable this?

markstory:
> Since when do we need comments for throws tags? We shoudnt enforce this verbosity IMO. Any way to disable this?

I think this rule is part of our sniffer rules and has been for quite some time. Doesn't it help provide more context to IDE users?

ADmad:
> Since when do we need comments for throws tags?

It has been part of our ruleset through the `FunctionCommentSniff` since codesniffer 1.x but kept disabled. I re-enabled it for `next` branch.

ADmad:
Sadly we have lot of failures in core for missing description for `@throw` tags. So I'll disable it again, don't really want to spend time fixing them all.

ADmad:
@dereuromark Please run `pslam --update-baseline` locally after each round of docblock/typehint updates and commit changes if `psalm-baseline.xml` if any.

I have done so for this PR and we have a few ignores removed :clap:.

ADmad (approved):

PR #13184, opened 2019-05-02
4.x ".ctp" => ".php" in error templates

saeideng:
show `.php` ext instead of `.ctp` in error-suggestion section


dereuromark (approved):
Nice find!

PR #13185, opened 2019-05-02
Fix more array doc blocks to be more precise. Stricter typehints.

dereuromark:
Part 2/2 of https://github.com/cakephp/cakephp/pull/13183

I wasnt sure if we need Entity:get() to allow null as input - but I didnt remove that yet.
IMO only real string makes sense, and we can then simplify the API here.

`@mixin \Cake\Core\InstanceConfigTrait` is needed on interface contracts where the trait methods are being used in code.

Hopefully, this allows PHPStan and CO to better introspect the code and find issues even more precise and early on now.

### DIFF
--- a/src/Core/functions.php
+++ b/src/Core/functions.php
@@ -74,7 +74,7 @@ if (!function_exists('pluginSplit')) {
      * @param string $name The name you want to plugin split.
      * @param bool $dotAppend Set to true if you want the plugin to have a '.' appended to it.
      * @param string|null $plugin Optional default plugin to use if no plugin is found. Defaults to null.
-     * @return array Array with 2 indexes. 0 => plugin name, 1 => class name.
+     * @return string[] Array with 2 indexes. 0 => plugin name, 1 => class name.
      * @link https://book.cakephp.org/3.0/en/core-libraries/global-constants-and-functions.html#pluginSplit
      */
     function pluginSplit(string $name, bool $dotAppend = false, ?string $plugin = null): array
@@ -100,7 +100,7 @@ if (!function_exists('namespaceSplit')) {
      * Commonly used like `list($namespace, $className) = namespaceSplit($class);`.
      *
      * @param string $class The full class name, ie `Cake\Core\App`.
-     * @return array Array with 2 indexes. 0 => namespace, 1 => classname.
+     * @return string[] Array with 2 indexes. 0 => namespace, 1 => classname.
      */
     function namespaceSplit(string $class): array
     {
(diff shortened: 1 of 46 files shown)


src/Shell/Helper/ProgressHelper.php
```diff
@@ -115,7 +115,7 @@ public function init(array $args = [])
      * @param int $num The amount of progress to advance by.
      * @return $this
      */
-    public function increment($num = 1)
+    public function increment(int $num = 1)
```
markstory: This could be a float too. I know we don't have it documented that way, but it could happen and I think we should be ok with a float.

dereuromark: I can fix the docblock

ADmad:
Some thing got messed up. Why is the PR showing so many commits now?

dereuromark:
Oops, I accidently pointed it to master. Fixed now.

ADmad:
psalm job on scrutinizer is showing errors which I believe are already fixed. Can you please merge / rebase `4.x` on this branch to be sure.

src/Core/functions.php
```diff
@@ -74,7 +74,7 @@ function h($text, bool $double = true, ?string $charset = null)
      * @param string $name The name you want to plugin split.
      * @param bool $dotAppend Set to true if you want the plugin to have a '.' appended to it.
      * @param string|null $plugin Optional default plugin to use if no plugin is found. Defaults to null.
-     * @return array Array with 2 indexes. 0 => plugin name, 1 => class name.
+     * @return string[] Array with 2 indexes. 0 => plugin name, 1 => class name.
```
ADmad: This incorrect since plugin name (value at 0 index) can be `null`. Psalm also complains about the same.

src/View/View.php
```diff
@@ -1316,7 +1317,7 @@ protected function _checkFilePath(string $file, string $path): string
      *
      * @param string $name The name you want to plugin split.
      * @param bool $fallback If true uses the plugin set in the current Request when parsed plugin is not loaded
-     * @return array Array with 2 indexes. 0 => plugin name, 1 => filename
+     * @return string[] Array with 2 indexes. 0 => plugin name, 1 => filename
```
ADmad: Here too plugin name (value at 0 index) can be `null` hence `string[]` is incorrect.

ADmad (dismissed):
psalm failures seem relevant and need to be addressed.

src/Core/functions.php
dereuromark: Ah ok

ADmad:
I took care of the errors reported by psalm.

PR #13189, opened 2019-05-04
Use \Throwable instead of \Exception.

ADmad:


markstory:
Looks good once tests are good.

PR #13190, opened 2019-05-04
4.x typehints

ADmad:
I did another pass for typehints for namespaces `Auth` to `Core`.

### DIFF
--- a/src/Cache/Cache.php
+++ b/src/Cache/Cache.php
@@ -346,11 +346,11 @@ class Cache
      *
      * @param iterable $keys An array or Traversable of keys to fetch from the cache
      * @param string $config optional name of the configuration to use. Defaults to 'default'
-     * @return array An array containing, for each of the given $keys,
+     * @return iterable An array containing, for each of the given $keys,
      *   the cached data or false if cached data could not be retrieved.
      * @throws \Cake\Cache\InvalidArgumentException
      */
-    public static function readMany(iterable $keys, string $config = 'default'): array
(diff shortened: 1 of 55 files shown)


src/Controller/Controller.php
```diff
@@ -288,7 +288,7 @@ public function loadComponent(string $name, array $config = []): Component
      * @param string $name Property name
      * @return \Cake\Datasource\RepositoryInterface|null The model instance or null
      */
-    public function __get(string $name)
+    public function __get(string $name): ?\Cake\Datasource\RepositoryInterface
```
dereuromark: Doesnt any of the existing sniffers fix the FCQN to use statements? Otherwise I have one here to recommend.

src/Cache/Cache.php
```diff
@@ -199,7 +199,7 @@ protected static function _buildEngine(string $name): void
      * @return \Cake\Cache\CacheEngine
      * @deprecated 3.7.0 Use Cache::pool() instead. This method will be removed in 5.0.
      */
-    public static function engine(string $config)
+    public static function engine(string $config): CacheEngine
```
markstory: Should this be the psr6 interface hint instead?

ADmad: If I do that I am pretty sure static analyzers will complain about use of methods like `increment()` on the returned instances which are not PSR6 interface methods. So think it would be best to remove the return typehint and update the docblock to union of `CacheInterface` and `CacheEngineInterface`.

ADmad: Done

PR #13191, opened 2019-05-04
Add check for non empty value for View::$layout.

ADmad:
This helps provide verbose message for those upgrading from 3.x.

### DIFF
--- a/src/View/View.php
+++ b/src/View/View.php
@@ -705,7 +705,7 @@ class View implements EventDispatcherInterface
             if (empty($this->layout)) {
                 throw new RuntimeException(
                     'View::$layout must be a non-empty string.' .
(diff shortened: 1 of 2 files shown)


src/View/View.php
```diff
@@ -702,6 +702,13 @@ public function render(?string $template = null, $layout = null): string
         $this->dispatchEvent('View.afterRender', [$templateFileName]);
 
         if ($this->autoLayout) {
+            if (empty($this->layout)) {
+                throw new RuntimeException(
+                    'View::$layout must be a non-empty string.' .
+                    'To disable layout rendering set View::$autoLayout to false instead.'
```
markstory: Should we tell folks to use the disableAutoLayout() method instead of modifying the property?

ADmad: Done

PR #13192, opened 2019-05-05
4.x typehints 2

ADmad:
Typehints for `Database`, `Datasource` and `ORM`.

### DIFF
--- a/src/Database/Schema/TableSchema.php
+++ b/src/Database/Schema/TableSchema.php
@@ -668,7 +668,7 @@ class TableSchema implements TableSchemaInterface, SqlGeneratorInterface
     /**
      * @inheritDoc
      */
-    public function setOptions(array $options): TableSchemaInterface
+    public function setOptions(array $options)
     {
         $this->_options = array_merge($this->_options, $options);
 
--- a/src/Database/Schema/TableSchema.php
+++ b/src/Database/Schema/TableSchema.php
@@ -456,7 +456,7 @@ class TableSchema implements TableSchemaInterface, SqlGeneratorInterface
      * {@inheritDoc}
      * @throws \Cake\Database\Exception
      */
-    public function addIndex(string $name, $attrs): TableSchemaInterface
+    public function addIndex(string $name, $attrs)
     {
         if (is_string($attrs)) {
             $attrs = ['type' => $attrs];
@@ -536,7 +536,7 @@ class TableSchema implements TableSchemaInterface, SqlGeneratorInterface
      * {@inheritDoc}
      * @throws \Cake\Database\Exception
      */
-    public function addConstraint(string $name, $attrs): TableSchemaInterface
+    public function addConstraint(string $name, $attrs)
     {
         if (is_string($attrs)) {
             $attrs = ['type' => $attrs];
@@ -592,7 +592,7 @@ class TableSchema implements TableSchemaInterface, SqlGeneratorInterface
     /**
      * @inheritDoc
      */
-    public function dropConstraint(string $name): TableSchemaInterface
+    public function dropConstraint(string $name)
     {
         if (isset($this->_constraints[$name])) {
             unset($this->_constraints[$name]);
--- a/src/Database/Schema/TableSchema.php
+++ b/src/Database/Schema/TableSchema.php
@@ -453,8 +453,7 @@ class TableSchema implements TableSchemaInterface, SqlGeneratorInterface
     }
 
     /**
-     * {@inheritDoc}
-     * @throws \Cake\Database\Exception
+     * @inheritDoc
      */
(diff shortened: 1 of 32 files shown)


ADmad:
As I had feared trying to typehint Database/Datasource/ORM makes the static analyzers pretty unhappy :disappointed:.

src/Database/Schema/TableSchemaInterface.php
```diff
@@ -225,9 +225,9 @@ public function indexes(): array;
      * @param string $name The name of the constraint.
      * @param array|string $attrs The attributes for the constraint.
      *   If string it will be used as `type`.
-     * @return \Cake\Database\Schema\TableSchemaInterface
+     * @return self
      */
-    public function addConstraint(string $name, $attrs);
+    public function addConstraint(string $name, $attrs): self;
```
dereuromark: Arent those chaining in a way as well? Those should also not have typehints here.

src/Database/Schema/TableSchema.php
```diff
@@ -592,7 +592,7 @@ public function addConstraint(string $name, $attrs): TableSchemaInterface
     /**
      * @inheritDoc
      */
-    public function dropConstraint(string $name)
+    public function dropConstraint(string $name): TableSchemaInterface
```
dereuromark: $this = chainable
no typehint here.

src/Database/Schema/TableSchemaInterface.php
ADmad: Return typehint removed.

src/Database/Schema/TableSchema.php
ADmad: Return typehint removed.

src/Database/Schema/TableSchemaInterface.php
```diff
@@ -241,9 +241,9 @@ public function getConstraint(string $name): ?array;
      * Remove a constraint.
      *
      * @param string $name Name of the constraint to remove
-     * @return \Cake\Database\Schema\TableSchemaInterface
+     * @return self
```
dereuromark: `static` then AFAIK, since those are chainable use and as such extended classes can have added methods on top.

dereuromark:
Looks good otherwise.

ADmad:
@dereuromark phpstan becomes unhappy if `@return $this` in docblock of methods in interface. How do we make it happy? 

```
 ------ ------------------------------------------------------------------- 
  Line   Database/Schema/TableSchema.php                                    
 ------ ------------------------------------------------------------------- 
  498    Method Cake\Database\Schema\TableSchema::addIndex() should return  
         $this(Cake\Database\Schema\TableSchemaInterface) but returns       
         $this(Cake\Database\Schema\TableSchema).                           
 ------ ------------------------------------------------------------------- 

 ------ ------------------------------------------------------------------- 
  Line   ORM/Query.php                                                      
 ------ ------------------------------------------------------------------- 
  1183   Method Cake\ORM\Query::find() should return                        
         $this(Cake\Datasource\QueryInterface) but returns Cake\ORM\Query.  
 ------ ------------------------------------------------------------------- 
```

dereuromark:
Isnt that still this weird bug with phpstan not understanding partial overwrite `{@inheritDoc}` maybe?

ADmad:
Aha, that's the case. So I guess we'll just have to ignore those errors.

ADmad:
Or i'll just copy the param and return tags in the implementing classes.

dereuromark:
Jep - I would just use partial overwrite for the textual part (description), never for the actual param/return values.
That usually works fine.

ADmad:
Hmm the error for `Query::find()` is actually valid since it doesn't return `$this` but the return value of `Table::callFinder()` instead.

dereuromark:
So this is `static` then, right?

src/Datasource/QueryCacher.php
```diff
@@ -122,7 +122,7 @@ protected function _resolveKey($query): string
     /**
      * Get the cache engine.
      *
-     * @return \Cake\Cache\CacheEngine
+     * @return \Psr\SimpleCache\CacheInterface&\Cake\Cache\CacheEngineInterface
```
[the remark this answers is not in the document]
ADmad: Yes, it's union type supported by static analysers. The CacheEngine class implements both these interfaces.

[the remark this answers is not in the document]
garas: I didn't yet know about _intersection type_.

https://medium.com/@ondrejmirtes/union-types-vs-intersection-types-fd44a8eacbb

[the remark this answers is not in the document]
ADmad: Sorry, I mixed up the terms, that's intersection type not union :)

PR #13193, opened 2019-05-05
4.x typehints 3

ADmad:
Update packages `Error` to `Http`.

### DIFF
--- a/src/Http/Cookie/Cookie.php
+++ b/src/Http/Cookie/Cookie.php
@@ -236,10 +236,22 @@ class Cookie implements CookieInterface
     }
 
     /**
-     * {@inheritDoc}
-     * @psalm-suppress InvalidReturnType
+     * Gets the cookie value as a string.
+     *
+     * This will collapse any complex data in the cookie with json_encode()
+     *
+     * @return mixed
+     * @deprecated 4.0.0 Use getScalarValue() instead.
      */
     public function getStringValue()
+    {
+        return $this->getScalarValue();
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function getScalarValue()
     {
         if ($this->isExpanded) {
             return $this->_flatten($this->value);
--- a/src/Http/Cookie/CookieInterface.php
+++ b/src/Http/Cookie/CookieInterface.php
@@ -49,13 +49,13 @@ interface CookieInterface
     public function getValue();
 
     /**
-     * Gets the cookie value as a string.
+     * Gets the cookie value as scalar.
      *
      * This will collapse any complex data in the cookie with json_encode()
      *
-     * @return string
+     * @return mixed
      */
-    public function getStringValue();
+    public function getScalarValue();
 
     /**
      * Create a cookie with an updated value.
--- a/src/Http/ServerRequest.php
+++ b/src/Http/ServerRequest.php
@@ -1885,7 +1885,7 @@ class ServerRequest implements ServerRequestInterface
      * @param bool $preserveHost Whether or not the host should be retained.
      * @return static
      */
-    public function withUri($uri, $preserveHost = false)
+    public function withUri(UriInterface $uri, $preserveHost = false)
     {
         $new = clone $this;
         $new->uri = $uri;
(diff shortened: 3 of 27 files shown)


src/Http/Cookie/Cookie.php
```diff
@@ -236,10 +236,22 @@ public function getValue()
...
+     * Gets the cookie value as a string.
+     *
+     * This will collapse any complex data in the cookie with json_encode()
+     *
+     * @return mixed
+     * @deprecated 4.0.0 Use getScalarValue() instead.
      */
     public function getStringValue()
+    {
+        return $this->getScalarValue();
```
inoas: `return (string)$this->getScalarValue();`?

aka could this not be an integer? The method name suggests otherwise.

ADmad: The old method name was misleading, hence it's deprecated and new method added instead.

src/Http/Cookie/Cookie.php
```diff
@@ -236,10 +236,22 @@ public function getValue()
...
 
     /**
-     * {@inheritDoc}
-     * @psalm-suppress InvalidReturnType
+     * Gets the cookie value as a string.
+     *
+     * This will collapse any complex data in the cookie with json_encode()
+     *
+     * @return mixed
+     * @deprecated 4.0.0 Use getScalarValue() instead.
```
markstory: When is the return not a string?

ADmad: Whenever it's not an array :)

Assertions in `ResponseTest::testWithCookieScalar()` (which I haven't changed) fail if the return value for the method is type hinted as `string`.

src/Http/Cookie/CookieInterface.php
```diff
@@ -49,13 +49,13 @@ public function getName(): string;
...
-     * Gets the cookie value as a string.
+     * Gets the cookie value as scalar.
      *
      * This will collapse any complex data in the cookie with json_encode()
      *
-     * @return string
+     * @return mixed
      */
-    public function getStringValue();
+    public function getScalarValue();
```
dereuromark: Do we need to document this in the migration guide etc?

src/Http/ServerRequest.php
```diff
@@ -380,12 +380,16 @@ protected function _processGet(array $query, string $queryString = ''): array
     /**
      * Process uploaded files and move things onto the post data.
      *
-     * @param array $post Post data to merge files onto.
+     * @param mixed $post Post data to merge files onto.
      * @param array $files Uploaded files to merge in.
      * @return array merged post + file data.
      */
     protected function _processFiles($post, array $files)
```
dereuromark: When can this ever not operate on array? I checked, the code before for sure returns an array

    $config['post'] = $this->_processPost($config['post']);

As such I wonder if we should even allow for this case here and weaken the API?

src/Http/ServerRequest.php
```diff
@@ -1881,7 +1885,7 @@ public function getUri()
      * @param bool $preserveHost Whether or not the host should be retained.
      * @return static
      */
-    public function withUri(UriInterface $uri, $preserveHost = false)
+    public function withUri($uri, $preserveHost = false)
```
dereuromark: Could this be kept? Also, bool addition possible?

dereuromark:
Rest looks good to me.

src/Http/Cookie/CookieInterface.php
ADmad: The old method still exists with deprecated tag but yeah mentioning it in migration guide would also be nice.

src/Http/ServerRequest.php
ADmad: These methods are inherited from the PSR interface so can't add additional typehints for arguments.

src/Http/ServerRequest.php
ADmad: There's a test case where POST data is string:

https://github.com/cakephp/cakephp/blob/b1673c25d8955cc69e551fa97e7a159bb335f82a/tests/TestCase/Http/ServerRequestTest.php#L2437-L2443

I can change the `mixed` to `array|string` if you prefer.

src/Http/ServerRequest.php
ADmad: My bad, the interface method does have the `UriInterface` typehint.

src/Http/ServerRequest.php
dereuromark: sounds fair.

src/Http/ServerRequest.php
dereuromark: What about the 2nd arg? bool typehint?

PR #13194, opened 2019-05-05
4.x - throw RuntimeException when using subcommands in console

saeideng:


src/Console/Command.php
```diff
@@ -116,7 +117,14 @@ public function getOptionParser(): ConsoleOptionParser
         $parser = new ConsoleOptionParser($name);
         $parser->setRootName($root);
 
-        return $this->buildOptionParser($parser);
+        $parser = $this->buildOptionParser($parser);
+        if ($parser->subcommands()) {
+            throw new RuntimeException(
+                'The use of the ConsoleOptionParser::addSubcommand() is not supported by Console'
```
markstory:
```suggestion
                'You cannot add sub-commands to `Command` sub-classes. Instead make a separate command.'
```

PR #13198, opened 2019-05-06
Improve doc block

bancer:


lib/Cake/Utility/ClassRegistry.php
```diff
@@ -300,7 +300,7 @@ public static function config($type, $param = array()) {
  *
  * @param string $alias Alias to check.
  * @param string $class Class name.
- * @return bool
+ * @return bool|object Object stored in registry or `false` if the object does not exist.
```
dereuromark: Technically, this is `object|false` then though. True is never returned.

PR #13200, opened 2019-05-07
4.x typehints 4

ADmad:
Packages `I18n` to `Routing` except `ORM`.


src/I18n/Formatter/IcuFormatter.php
```diff
@@ -35,7 +35,7 @@ class IcuFormatter implements FormatterInterface
      * @throws \Aura\Intl\Exception\CannotFormat
      * @throws \Aura\Intl\Exception\CannotInstantiateFormatter
      */
-    public function format($locale, $message, array $vars): string
+    public function format($locale, $message, $vars): string
```
dereuromark: whats wrong with this one?

ADmad: Nothing, removed mistakenly. The perils of PHP 7.2 parameter type widening.

PR #13201, opened 2019-05-07
Fix assertion method.

ADmad:


ADmad:
Appveyor is green again.

PR #13202, opened 2019-05-07
4.x typehints 5

ADmad:
Updated `TestSuite` to `View`.

### DIFF
--- a/src/Utility/Hash.php
+++ b/src/Utility/Hash.php
@@ -209,11 +209,11 @@ class Hash
     /**
      * Check a key against a token.
      *
-     * @param string $key The key in the array being searched.
+     * @param mixed $key The key in the array being searched.
      * @param string $token The token being matched.
      * @return bool
      */
-    protected static function _matchToken($key, $token): bool
+    protected static function _matchToken($key, string $token): bool
     {
         switch ($token) {
             case '{n}':
@@ -296,7 +296,7 @@ class Hash
      *
      * @param array $data The data to insert into.
      * @param string $path The path to insert at.
-     * @param array|null $values The values to insert.
+     * @param mixed $values The values to insert.
      * @return array The data with $values inserted.
      * @link https://book.cakephp.org/3.0/en/core-libraries/hash.html#Cake\Utility\Hash::insert
      */
@@ -346,7 +346,7 @@ class Hash
      * @param mixed $values The values to insert when doing inserts.
      * @return array data.
      */
-    protected static function _simpleOp($op, $data, $path, $values = null)
+    protected static function _simpleOp(string $op, array $data, array $path, $values = null): array
     {
         $_list =& $data;
 
@@ -380,6 +380,8 @@ class Hash
                 $_list =& $_list[$key];
             }
         }
+
+        return $data;
     }
 
     /**
@@ -649,7 +651,7 @@ class Hash
      * @param mixed $var Array to filter.
      * @return bool
      */
-    protected static function _filter($var)
+    protected static function _filter($var): bool
     {
         return $var === 0 || $var === 0.0 || $var === '0' || !empty($var);
     }
@@ -767,7 +769,7 @@ class Hash
      * @param array $return The return value to operate on.
      * @return void
      */
-    protected static function _merge($stack, &$return)
+    protected static function _merge(array $stack, array &$return): void
     {
         while (!empty($stack)) {
             foreach ($stack as $curKey => &$curMerge) {
@@ -872,7 +874,7 @@ class Hash
      * @return array An array of the modified values.
      * @link https://book.cakephp.org/3.0/en/core-libraries/hash.html#Cake\Utility\Hash::map
      */
-    public static function map(array $data, $path, $function): array
+    public static function map(array $data, string $path, callable $function): array
     {
         $values = (array)static::extract($data, $path);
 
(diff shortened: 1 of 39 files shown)


src/Utility/Hash.php
```diff
@@ -209,11 +209,11 @@ protected static function _splitConditions(string $token): array
     /**
      * Check a key against a token.
      *
-     * @param string $key The key in the array being searched.
+     * @param mixed $key The key in the array being searched.
```
dereuromark: We should avoid mixed by using string|int etc

src/Utility/Hash.php
```diff
@@ -296,7 +296,7 @@ protected static function _matches($data, string $selector): bool
      *
      * @param array $data The data to insert into.
      * @param string $path The path to insert at.
-     * @param array|null $values The values to insert.
+     * @param mixed $values The values to insert.
```
dereuromark: Can we keep this more specific one?

src/Validation/RulesProvider.php
```diff
@@ -59,9 +59,9 @@ public function __construct($class = Validation::class)
      *
      * @param string $method the validation method to call
      * @param array $arguments the list of arguments to pass to the method
-     * @return bool whether or not the validation rule passed
+     * @return mixed whether or not the validation rule passed
```
dereuromark: You should check/revert all mixed changes.

src/Utility/Hash.php
ADmad: No since you can insert any type of value to particular key.

src/Validation/RulesProvider.php
ADmad: All mixed changes are based on actual code. Adding typehints as per old docblocks causes tests to fail.

src/Utility/Hash.php
ADmad: Well you can't since `Hash` class deals with arrays where keys don't always have to be string.

dereuromark: True, but usually scalar. As such better to be explicit. Right now you would also allow objects or other arrays etc, thats bad.

src/Utility/Hash.php
dereuromark: Isnt this iterating inside, so it must be an array of things - saying mixed means you could also work on random objects like Response class, I doubt that is possible or intended.
We usually annotate the intended target input, not all remotely possible ones you could pass without crashing it IMO.

src/Validation/RulesProvider.php
dereuromark: What does a method return that so far was boolean? See comment beside it.

ADmad:
With this PR I have done a pass for all packages. 

The updates for holy trinity of Database, Datasource and ORM in #13192 still remains. I'll get back to it after taking a breather as trying to property resolve types in those packages leads to quite a bit of hair pulling.

dereuromark:
I am strongly against adding more mixed typehints where it is not really true.
We need to be more precise here, as mixed kills basically all introspection, which in those cases above is not really true. Those are not random value storage, but concrete input to operate on.

src/Validation/RulesProvider.php
ADmad: This is for the magic `__call()` which proxies calls to other to other classes so we really can't guarantee that the return type will always be boolean.

ADmad:
@dereuromark For the 3 instances where the type is changed to `mixed` I have stated the reason. Each change is logically justified and as per actual test cases.

src/Validation/RulesProvider.php
dereuromark: No one can ever guarantee that, nor did we ever do that.
We always have been annotating the expected input and output types, that is what we should continue to do, not all possible wrong ones on top. Otherwise documentation here really doesnt help much, neither for humans nor for tooling.

dereuromark: To clarify: The follow up code will blow up if the documented expectations are not met. That is then outside of this documentation here. 
We could, if we want, also add a bool typehint to assert that on code level of course.

dereuromark:
I disagree. Each instance is different, but none of them justify going to mixed which basically means anything. That is not true in those cases. Ideally, you break it down more specifically here.

src/Utility/Hash.php
ADmad: Specifying a list of all scalar type vs mixed isn't going to provide any practical benefit. Won't make any difference to tools and humans can already know it takes valid array key types by reading description.

src/Utility/Hash.php
ADmad: Fixing incorrect type was more valuable than listing the possible one. You are welcome to check the code / test case and list the possible types instead.

src/Validation/RulesProvider.php
ADmad: Same comment as above. I fixed the incorrect type, you can update it for accuracy.

ADmad:
Changing those to `mixed` makes the docblock correct. You can update them to be more accurate.

ADmad:
Appveyour build success didn't last even for a day :laughing: 
https://ci.appveyor.com/project/cakephp/cakephp/builds/24355793#L24929

src/Utility/Hash.php
dereuromark: Why? You then know that it cannot be any random object for example. As such the benefit is 10x more than mixed already by design.

dereuromark (approved):

PR #13208, opened 2019-05-08
WIP - Generate output for static analysers in checkstyle format.

ADmad:
This allow error reports to be seen on GH pull request's "Checks" tab itself.

### DIFF
--- a/.scrutinizer.yml
+++ b/.scrutinizer.yml
@@ -14,7 +14,7 @@ build:
         override:
           - phpcs-run src/ tests/
 
-    static-analysis:
+    phpstan:
       environment:
         php:
           ini:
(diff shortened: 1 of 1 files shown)


ADmad:
Currently psalm doesn't allow listing issues only with severity "error" in checkstyle file, hence can only try to use checkstyle file for phpstan.

But even for phpstan scrutinizer keeps complaining it can't parse the file :slightly_frowning_face:
https://scrutinizer-ci.com/g/cakephp/cakephp/inspections/41ad2c26-b20c-4fa3-80f1-9ad053a87b79

ADmad:
Will revisit this sometime again in future.

PR #13209, opened 2019-05-09
Fix plural form calculation for Turkish

jiru:
CakePHP assumes that Turkish has only one plural form (in other words, it assumes that "{n} tomato" and "{n} tomatoes" translates the same in Turkish). However, Turkish has two plural forms, n=1 and n!=1, as show on these pages:

• https://unicode.org/cldr/charts/latest/supplemental/language_plural_rules.html
• https://translations.launchpad.net/+languages/tr

This PR sets the plural form calculation as it is described on these two pages. For reference, here is the [original issue](/Tatoeba/tatoeba2/issues/1866) in our project that prompted me to submit this PR.

### DIFF
--- a/src/I18n/PluralRules.php
+++ b/src/I18n/PluralRules.php
@@ -116,7 +116,7 @@ class PluralRules
         'th' => 0,
         'ti' => 2,
         'tk' => 1,
-        'tr' => 0,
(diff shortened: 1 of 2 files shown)


markstory:
Thanks :+1:

PR #13210, opened 2019-05-09
4.x psalm

ADmad:


src/View/XmlView.php
```diff
@@ -95,9 +95,9 @@ class XmlView extends SerializedView
      *   'format' as 'attributes' instead of 'tags'.
      *
      * @param array|string|true $serialize The name(s) of the view variable(s) that need(s) to be serialized
-     * @return string The serialized data
+     * @return string|false The serialized data
      */
-    protected function _serialize($serialize): string
+    protected function _serialize($serialize)
```
dereuromark: This will compromise the render() part (typehinted return string there) and other higher levels. Why not catching the false and returning empty string insdead?

src/Controller/ComponentRegistry.php
```diff
@@ -85,9 +85,9 @@ public function setController(Controller $controller)
      * Part of the template method for Cake\Core\ObjectRegistry::load()
      *
      * @param string $class Partial classname to resolve.
-     * @return string|null Either the correct class name or null.
+     * @return string|object|null Either the correct class name or null.
      */
-    protected function _resolveClassName($class): ?string
+    protected function _resolveClassName($class)
```
dereuromark: How can this return an object? The code does not show this.

src/View/XmlView.php
ADmad: `Serialized::render()` throws exception if this method returns false, so there's no comprise at the level of public API. Allowing this method to return false was just convenient as that's the return value of `json_encode()` in case of failure.

src/Controller/ComponentRegistry.php
ADmad: `ComponentRegistry` doesn't but there are other registry classes where this method can return object. Hence the return typehint is removed to match parent class.

dereuromark: I didnt see any. If so, we might want to fix that instead maybe?

src/View/XmlView.php
dereuromark: Wouldnt it be cleaner/clearer to let this method blow up?
If somewhere later the false leads to an issue it is harder to track down, the cause is cloaked then.

I am for more defensive coding in the right places, and meaningful exceptions including the correct stack trace then. This also allows us to keep the typehints that are supposed to document the code :)

src/Controller/ComponentRegistry.php
ADmad: `ConnectionRegistry` is one such registry. I don't think removing the ability to return objects would be an acceptable BC break.

Issue #12957, opened 2019-02-07
4.x saveAll()/deleteAll() for callbacks


dereuromark:
This is a (multiple allowed):

* [x] enhancement
* [x] feature-discussion (RFC)

CakePHP Version: 4+

### Status Quo
Every week people confuse deleteAll() to be a wrapper for delete() in loops.
But in fact, as atomic operations, there is no callbacks..

### Proposal
Are you open to harmonize the API here in 4.0 or 5.0 to reflect this a bit more cleanly?

- save(Entity $entity) - stays, incl callbacks
- saveMany(Entity[] $entities) - added, incl callbacks then
- delete(Entity $entity) - stays, incl callbacks
- ~~deleteAll~~ deleteMany(Entity[] $entities) - added, incl callbacks then
- updateAll($conditions) - stays as it is, atomic
- removeAll($conditions) - renamed from deleteAll(), atomic

The rename should clarify this is not as related to delete() as one might think.
The updateAll() addition clarifies that updateAll() is sth different then saveAll().

### Interface?
Do we then need the atomic operations to be part of the RepositoryInterface?
What methods are needed e.g. for abstractions like TwitterApi or Rss reading and should be part of the generic interface, which ones should go on top for more concrete cases like ORM?

### Transaction safe?
We can either keep transactional outside of the class, or we could add it by default. This example does the first:
```
	/**
	 * save() calls for multiple entities.
	 *
	 * Wrap it to be transaction safe for all save calls:
	 *
	 *  // In a controller.
	 *  $articles->connection()->transactional(function () use ($articles, $entities) {
	 *      $articles->saveAll($entities, ['atomic' => false]);
	 *  }
	 *
	 * @param \Cake\Datasource\EntityInterface[] $entities
	 * @param array $options
	 * @return bool
	 */
	public function saveAll(array $entities, array $options = []) {
		$success = true;
		foreach ($entities as $entity) {
			$success = $success & (bool)$this->save($entity, $options);
		}
		return (bool)$success;
	}
```

Looking into https://github.com/UseMuffin/Webservice/blob/master/src/Model/Endpoint.php it seems that all methods could be part of the RepositoryInterface.

markstory:
I would prefer we don't break the API on `deleteAll` and instead add new methods that accept entity lists. `deleteMany($entities)` as an example.

Having to supply a list of entities would help make the new methods distinct and differently flexible from the existing bulk operation methods.

dereuromark:
[...]

PR #13212, opened 2019-05-09
Add saveManyOrFail().

dereuromark:
Refs https://github.com/cakephp/cakephp/issues/12957

### DIFF
--- a/src/ORM/Table.php
+++ b/src/ORM/Table.php
@@ -2067,7 +2067,7 @@ class Table implements RepositoryInterface, EventListenerInterface, EventDispatc
      * @param array|\ArrayAccess $options Options used when calling Table::save() for each entity.
      * @return \Cake\Datasource\EntityInterface[]|\Cake\Datasource\ResultSetInterface Entities list.
      * @throws \Exception
-     * @throws \Cake\ORM\Exception\PersistenceFailedException
+     * @throws \Cake\ORM\Exception\PersistenceFailedException If an entity couldn't be saved.
      */
     public function saveManyOrFail(iterable $entities, $options = []): iterable
     {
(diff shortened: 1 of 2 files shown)


tests/TestCase/ORM/TableTest.php
```diff
@@ -2736,6 +2736,70 @@ public function testSaveManyFailedWithException()
...
+            ->first();
+        $this->assertSame('admad', $first->name);
+    }
+
+    /**
+     * Test saveManyOrFail() with failed save
+     *
+     * @return void
+     */
+    public function testSaveManyOrFailFailed()
```
markstory: Would be good to have a test that covers a unique constraint failure. And also a test covering the attribute roll back.

dereuromark: I am happy if someone wants to help out here, adding more tests. So far I copied the tests of saveMany()

src/ORM/Table.php
```diff
@@ -2086,6 +2086,61 @@ public function saveMany(iterable $entities, $options = [])
...
+     * any one of the records fails to save due to failed validation or database
+     * error.
+     *
+     * @param \Cake\Datasource\EntityInterface[]|\Cake\Datasource\ResultSetInterface $entities Entities to save.
+     * @param array|\ArrayAccess $options Options used when calling Table::save() for each entity.
+     * @return \Cake\Datasource\EntityInterface[]|\Cake\Datasource\ResultSetInterface Entities list.
+     * @throws \Exception
+     * @throws \Cake\ORM\Exception\PersistenceFailedException
+     */
+    public function saveManyOrFail(iterable $entities, $options = [])
```
ADmad: There's lot of code duplication between this method and `saveMany()`, only difference being this method stores reference to entity which failed to save and then throwing exception. So the common code should be consolidated into a helper method.

dereuromark: I refactored.
We shouldnt use too many exceptions for workflow :) But I hope the current solution is maximum DRY while still allowing both cases.

dereuromark:
The fails are unrelated afaik.

markstory:
Both scrutinizer and stickler have some relevant errors.

dereuromark:
Sry, I must have overlooked them. With the failing appveeyor it also always looks failing somewhere.

PR #13213, opened 2019-05-09
Add deleteMany() and deleteManyOrFail().

dereuromark:
Completes/Resolves https://github.com/cakephp/cakephp/issues/12957

Refs https://github.com/cakephp/cakephp/pull/13212

### DIFF
--- a/src/ORM/Table.php
+++ b/src/ORM/Table.php
@@ -2185,9 +2185,9 @@ class Table implements RepositoryInterface, EventListenerInterface, EventDispatc
     /**
      * @param \Cake\Datasource\EntityInterface[]|\Cake\Datasource\ResultSetInterface $entities Entities to delete.
      * @param array|\ArrayAccess $options Options used.
-     * @return EntityInterface|null
+     * @return \Cake\Datasource\EntityInterface|null
      */
-    protected function _deleteMany(iterable $entities, $options = [])
+    protected function _deleteMany(iterable $entities, $options = []): ?EntityInterface
     {
         $options = new ArrayObject((array)$options + [
                 'atomic' => true,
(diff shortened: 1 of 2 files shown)


src/ORM/Table.php
```diff
@@ -2135,6 +2135,98 @@ public function delete(EntityInterface $entity, $options = []): bool
...
+        if ($success && $this->_transactionCommitted($options['atomic'], $options['_primary'])) {
+            foreach ($entities as $entity) {
+                $this->dispatchEvent('Model.afterDeleteCommit', [
+                    'entity' => $entity,
+                    'options' => $options,
+                ]);
+            }
+        }
+
+        if ($success === false) {
```
markstory: Would you be able to call `deleteMany()` and check the return value instead of duplicating the code?

dereuromark: I didnt know how to Track the failed entity here.

ADmad: You can move the closure for transaction used in this method and other common code into a helper function which return the failed entity. Then both this method and `deleteMany()` can use that helper method with the latter simply ignoring the failed entity reference.

dereuromark:
good to go?

src/ORM/Table.php
```diff
@@ -2135,6 +2135,88 @@ public function delete(EntityInterface $entity, $options = []): bool
...
+        $options = new ArrayObject((array)$options + [
+                'atomic' => true,
+                'checkRules' => true,
+                '_primary' => true,
+            ]);
+
+        $failed = $this->_executeTransaction(function () use ($entities, $options) {
+            foreach ($entities as $entity) {
+                if (!$this->_processDelete($entity, $options)) {
+                    return $entity;
```
saeideng: this now is useless `return $entity;` even is incorrect 
you are working with array of entities not just an entity , so why this returns just one entity ? 
just return false

and you can return `return true;` (instead of `null` )

dereuromark: It returns the last one that was processed before failing, see also the other many operations in this class.

Issue #11789, opened 2018-03-05
3.6: Fix hidden query bugs a.k.a. IS NULL issue of != NULL


dereuromark:
This is a (multiple allowed):

* [x] bug
* [x] enhancement

Motivation:
- Hidden bugs in software, often times from 2.x to 3.x upgrades
- High level needs to know about things it shouldnt need to know
- Methods like findOrCreate() and using $conditions array for finding + creating is way more complex currently

## IN
This has been annoying me since day one of 3.x, and every time I upgrade a 2.x app, this annoys me even more.
The main reason: It contains hidden bugs as the resulting SQL query is invalid but silently just returns 0 results.
So no one ever knows if those queries work since no results just means "no hits" usually, no one would suspect the query to be invalid then, of course.

```
$x = $this->Logs->find()
    ->where(['comment' => null])
    ->all();
```
SQL:
```
WHERE 
  `comment` = NULL
```
=> 0 rows, silently returning nothing

**That makes no sense to me**. In what use case would one ever want to produce such a query? It does not throw a DB exception but is an invalid query that will never yield any useful result.
The desired behavior is pretty clear - make the DB check for the fields that are NULL (the developer doesnt want/need to know the specifics of the query building IMO).

```
$x = $this->Logs->find()
    ->where(['comment IS' => null])
    ->all();
```
and the resulting
```
WHERE 
  (`comment`) IS NULL
```
suddenly returns the expected 3 records.

So two things we can do for 3.6:

### Solution A
The obvious one: Throw an exception for invalid queries instead of silently doing nothing creates a false sense of correctness here.
This would be somewhat in line with IN / NOT IN behavior for empty rows.

### Solution B
Go back to the very intuitive and IMO most correct way of using 2.x behavior:
Once a null is passed, use that for the query with proper IS or IS NOT usage.

Especially for the positive case (IS) I do not see any harm here - especially compared to the current bugs we have due to the silent query fault.

That said I am fine with adding extra methods that would provide this. But I honestly do not think those should be necessary.
Cake ORM should go back to the dev friendly way of converting this to what the DB needs without having to force the developer to add extras everywhere.

I still have hundreds of these silent issues somewhere in upgraded apps and they are currently impossible to detect really, unless you use 100% test coverage
and always at least fixture data of >= 1 returned (never allow empty result collection for reading in tests). This is not feasable.
Let's please fix this now.

## NOT IN
[...]

PR #8621, opened 2016-04-09
CheckNull option for IsUnique

JayPHP:
This is a suggestion for issue https://github.com/cakephp/cakephp/issues/8620

PHPUnit tests;

> Time: 1.54 minutes, Memory: 164.00Mb
> 
> OK, but incomplete, skipped, or risky tests!
> Tests: 6883, Assertions: 21233, Skipped: 144.

PR #13214, opened 2019-05-09
4.x: IS NULL exception fix

dereuromark:
Test cases for "hidden query bugs a.k.a. IS NULL issue of != NULL"

Addresses https://github.com/cakephp/cakephp/issues/11789

Can someone please help to get this fixed for 4.0?
I dont know how to address the concrete fix in the ORM.

### DIFF
--- a/src/Database/Expression/QueryExpression.php
+++ b/src/Database/Expression/QueryExpression.php
@@ -768,6 +768,10 @@ class QueryExpression implements ExpressionInterface, Countable
             $operator = '!=';
         }
 
+        if ($value === null && $this->_conjunction !== ',') {
+            throw new \InvalidArgumentException();
+        }
+
         return new Comparison($expression, $value, $type, $operator);
     }
 
--- a/src/Database/Expression/QueryExpression.php
+++ b/src/Database/Expression/QueryExpression.php
@@ -22,6 +22,7 @@ use Cake\Database\Query;
 use Cake\Database\TypeMapTrait;
 use Cake\Database\ValueBinder;
 use Countable;
+use InvalidArgumentException;
 
 /**
  * Represents a SQL Query expression. Internally it stores a tree of
@@ -717,6 +718,7 @@ class QueryExpression implements ExpressionInterface, Countable
      * be extracted.
      * @param mixed $value The value to be bound to a placeholder for the field
      * @return string|\Cake\Database\ExpressionInterface
+     * @throws \InvalidArgumentException If operator is invalid or missing on NULL usage.
      */
     protected function _parseCondition(string $field, $value)
     {
@@ -769,7 +771,7 @@ class QueryExpression implements ExpressionInterface, Countable
         }
 
         if ($value === null && $this->_conjunction !== ',') {
-            throw new \InvalidArgumentException();
+            throw new InvalidArgumentException('Invalid or missing operator together with `null` usage.');
         }
 
         return new Comparison($expression, $value, $type, $operator);
--- a/src/ORM/Association/BelongsToMany.php
+++ b/src/ORM/Association/BelongsToMany.php
@@ -1324,23 +1324,20 @@ class BelongsToMany extends Association
         $belongsTo = $junction->getAssociation($target->getAlias());
         $hasMany = $source->getAssociation($junction->getAlias());
         $foreignKey = (array)$this->getForeignKey();
+        $foreignKey = array_map(function ($key) {
+            return $key . ' IS';
+        }, $foreignKey);
         $assocForeignKey = (array)$belongsTo->getForeignKey();
+        $assocForeignKey = array_map(function ($key) {
+            return $key . ' IS';
+        }, $assocForeignKey);
         $sourceKey = $sourceEntity->extract((array)$source->getPrimaryKey());
 
         $unions = [];
         foreach ($missing as $key) {
-            $conditions = array_merge(
-                array_combine($foreignKey, $sourceKey),
-                array_combine($assocForeignKey, $key)
-            );
-            foreach ($conditions as $k => $v) {
-                if ($v === null) {
-                    $conditions[$k . ' IS'] = $v;
-                    unset($conditions[$k]);
-                }
-            }
             $unions[] = $hasMany->find()
-                ->where($conditions);
+                ->where(array_combine($foreignKey, $sourceKey))
+                ->where(array_combine($assocForeignKey, $key));
         }
 
         $query = array_shift($unions);
--- a/src/ORM/Association/BelongsToMany.php
+++ b/src/ORM/Association/BelongsToMany.php
@@ -1329,9 +1329,18 @@ class BelongsToMany extends Association
 
         $unions = [];
         foreach ($missing as $key) {
+            $conditions = array_merge(
+                array_combine($foreignKey, $sourceKey),
+                array_combine($assocForeignKey, $key)
+            );
+            foreach ($conditions as $k => $v) {
+                if ($v === null) {
+                    $conditions[$k . ' IS'] = $v;
+                    unset($conditions[$k]);
+                }
+            }
             $unions[] = $hasMany->find()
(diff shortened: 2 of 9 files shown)


tests/TestCase/Database/QueryTest.php
```diff
@@ -4050,6 +4050,42 @@ public function testDirectIsNull()
...
+     */
+    public function testIsNullInvalid()
+    {
+        $this->expectException(\InvalidArgumentException::class);
+
+        $this->loadFixtures('Authors');
+        (new Query($this->connection))
+            ->select(['name'])
+            ->from(['authors'])
+            ->where(['name' => null])
```
saeideng: this is acceptable !! no ?

markstory:
I can take a look at this.

dereuromark:
Thank you, appreciated!

ADmad:
My guess is `QueryExpression::_addConditions()` is where the exception would be thrown :slightly_smiling_face:.

ADmad:
This uncovered generation of `foo = null` condition at various places in the ORM.

Only dealing with failures for `IsUnique` remains. I don't understand how that `allowMultipleNulls` option is supposed to work.

src/Database/Expression/QueryExpression.php
```diff
@@ -768,6 +768,10 @@ protected function _parseCondition(string $field, $value)
             $operator = '!=';
         }
 
+        if ($value === null && $this->_conjunction !== ',') {
+            throw new \InvalidArgumentException();
```
ADmad: @dereuromark I leave it to you to add proper error message here :)

src/ORM/Rule/ExistsIn.php
```diff
@@ -140,6 +140,12 @@ public function __invoke(EntityInterface $entity, array $options): bool
             $primary,
             $entity->extract($fields)
         );
+        foreach ($conditions as $k => $v) {
+            if ($v === null) {
+                $conditions[$k . ' IS'] = $v;
```
lorenzo: Using the same array seems wrong, better create a different one

src/ORM/Association/BelongsToMany.php
```diff
@@ -1329,9 +1329,18 @@ protected function _collectJointEntities(EntityInterface $sourceEntity, array $t
...
+            $conditions = array_merge(
+                array_combine($foreignKey, $sourceKey),
+                array_combine($assocForeignKey, $key)
+            );
+            foreach ($conditions as $k => $v) {
+                if ($v === null) {
+                    $conditions[$k . ' IS'] = $v;
+                    unset($conditions[$k]);
+                }
+            }
```
lorenzo: I think I would prefer building 2 new arrays for the keys where they get `IS` appended, irregardless of the value they have. Then you can `array_combine` them with the values

lorenzo (changes requested):

src/ORM/Rule/ExistsIn.php
markstory: Or don't use array_combine and instead iterate the primary fields and build a condition array with a foreach.

ADmad:
Updated as per @lorenzo's feedback.

ADmad:
Either the `allowNullableNulls` option for `IsUnique` is a joke or the tests are invalid. If I make the following change all tests pass, so the option is useless.

```diff
diff --git a/src/ORM/Rule/IsUnique.php b/src/ORM/Rule/IsUnique.php
index 6265ccbb7a..b34abcda80 100644
--- a/src/ORM/Rule/IsUnique.php
+++ b/src/ORM/Rule/IsUnique.php
@@ -98,11 +98,7 @@ class IsUnique
     {
         $aliased = [];
         foreach ($conditions as $key => $value) {
-            if ($multipleNulls) {
-                $aliased["$alias.$key"] = $value;
-            } else {
-                $aliased["$alias.$key IS"] = $value;
-            }
+            $aliased["$alias.$key IS"] = $value;
         }
```

lorenzo (approved):

src/ORM/Rule/IsUnique.php
```diff
@@ -86,23 +77,15 @@ public function __invoke(EntityInterface $entity, array $options): bool
...
-    protected function _alias(string $alias, array $conditions, bool $multipleNulls): array
+    protected function _alias(string $alias, array $conditions): array
     {
         $aliased = [];
         foreach ($conditions as $key => $value) {
-            if ($multipleNulls) {
-                $aliased["$alias.$key"] = $value;
-            } else {
-                $aliased["$alias.$key IS"] = $value;
-            }
```
markstory: I think this option was trying to enable the creation of unique rules that disallowed multiple null values in a multi-column index. While valid in SQL they are sometimes not what people wanted. With that said, if there is no test for it, or a broken test there isn't much we can do.

ADmad: Keeping an obscure feature without test coverage doesn't make sense, hence I removed it. If you prefer we can retain the `$options` argument for constructor so that we can re-add the option later if someone comes up with a proper test case.

dereuromark: We might want to look into this also for 3.x then as this seems to have similar issues then there.

dereuromark: Refs
- https://github.com/cakephp/cakephp/pull/11029/files
- https://github.com/cakephp/cakephp/pull/8621/files (!)
- https://github.com/cakephp/docs/pull/4954/files (!)

ADmad: #11029 if for `Existsin`. The `allowNullableNulls` option there does seem to have proper test coverage.

https://github.com/cakephp/cakephp/pull/8621#issuecomment-217716001 probably sheds light on the fact that `allowMultipleNulls` option was introduced to fix generation of invalid `foo = null`. It isn't actually a new feature. Now since the `= null` problem is solved by simply using `'foo IS'` for all fields, there's no need for that option.

ADmad: https://stackoverflow.com/questions/36031664/cakephp-3-multiple-isunique-allowing-null-duplicate

src/ORM/Rule/IsUnique.php
```diff
@@ -68,14 +60,16 @@ public function __invoke(EntityInterface $entity, array $options): bool
...
         $alias = $options['repository']->getAlias();
-        $conditions = $this->_alias($alias, $entity->extract($this->_fields), $allowMultipleNulls);
+        $conditions = $this->_alias($alias, $entity->extract($this->_fields));
         if ($entity->isNew() === false) {
             $keys = (array)$options['repository']->getPrimaryKey();
-            $keys = $this->_alias($alias, $entity->extract($keys), $allowMultipleNulls);
-            if (array_filter($keys, 'strlen')) {
+            $keys = $this->_alias($alias, $entity->extract($keys));
+            $filtered = array_filter($keys, function ($v) {
+                return $v !== null;
```
ADmad: Explicit `null` check makes code much easier to understand than the earlier wacky `strlen`.

ADmad: Or do we need to consider empty strings too?

ADmad: NVM, I reverted the change.

tests/TestCase/Database/QueryTest.php
```diff
@@ -4058,6 +4058,42 @@ public function testDirectIsNull()
...
+     * Tests that using the wrong NULL operator will throw meaningful exception instead of
+     * cloaking as always-empty result set.
+     *
+     * @return void
+     */
+    public function testIsNullInvalid()
+    {
+        $this->expectException(\InvalidArgumentException::class);
+
+        $this->loadFixtures('Authors');
```
saeideng: could you add/check exception message too ?
now tracking source of this exception is bit hard

dereuromark: Done, now we can probably merge.

PR #13215, opened 2019-05-10
Updated Table::setTable doc to show setting database scheme name

othercorey:
The database schema can already be set using the form 'schema.table' for the table name.  This mentions the feature and the format including the need to use automatic identifier quoting if the table name requires quoting.

You cannot manually quote the table name as the code in Collection::describe() doesn't currently handle it.

### DIFF
--- a/src/ORM/Table.php
+++ b/src/ORM/Table.php
@@ -339,6 +339,9 @@ class Table implements RepositoryInterface, EventListenerInterface, EventDispatc
     /**
      * Sets the database table name.
      *
+     * This can include the database schema name in the form 'schema.table'.
+     * If the name must be quoted, enable automatic identifier quoting.
+     *
      * @param string $table Table name.
      * @return $this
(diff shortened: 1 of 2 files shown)


src/ORM/Table.php
```diff
@@ -352,6 +355,8 @@ public function setTable($table)
     /**
      * Returns the database table name.
      *
+     * This can include the database schema name if set using setTable.
```
ADmad:
```suggestion
     * This can include the database schema name if set using `setTable()`.
```

othercorey: Is there a way to verify formatting in comments shows up correctly?

ADmad: None other than generating the api docs yourself locally using https://github.com/cakephp/cakephp-api-docs.

Else you can just compare existing docblocks and the generated api docs at api.cakephp.org.

Issue #13206, opened 2019-05-08
4.x EmailTrait::assertMailSentWith()


mirko-pagliai:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 4.x

This is the signature of the method:
```php
public function assertMailSentWith(string $expected, string $parameter, string $message = ''): void
```

Suppose I want to test attachments
Attachments are an array, as returned by the `Message::getAttachments()` method.

So, I can't do something like this:
```php
$this->assertMailSentWith(['myfile.txt' => ['file' => '...', 'mimetype' => '...'], 'attachments');
```

Because the method wants a string as first argument:
```
There was 1 error:

1) DatabaseBackup\Test\TestCase\Utility\BackupManagerTest::testSend
TypeError: Argument 1 passed to DatabaseBackup\Test\TestCase\Utility\BackupManagerTest::assertMailSentWith() must be of the type string, array given, called in /home/mirko/Libs/Plugins/cakephp-database-backup/tests/TestCase/Utility/BackupManagerTest.php on line 180

/home/mirko/Libs/Plugins/cakephp-database-backup/vendor/cakephp/cakephp/src/TestSuite/EmailTrait.php:229
/home/mirko/Libs/Plugins/cakephp-database-backup/tests/TestCase/Utility/BackupManagerTest.php:180
```

So, in my opinion, there are two possibilities:

1. allow the first argument of `assertMailSentWith()` to be an array;
2. add a new method, something like `assertMailSentWithAttachment()`
```php
public function assertMailSentWithAttachment(string $expectedFilename, array $expectedFile, string $message = ''): void
```
or
```php
public function assertMailSentWithAttachments(array $expectedFiles, string $message = ''): void
```

I want to remember that the `EmailAssertTrait::assertEmailAttachmentsContains()` method existed in CakePHP 3.x.

dereuromark:
Do you want to make a PR with your suggested changes?

mirko-pagliai:
OK, I'll do it on Monday

markstory:
Pull request merged now.

PR #13216, opened 2019-05-10
added `assertMailContainsAttachment()`

mirko-pagliai:
See https://github.com/cakephp/cakephp/issues/13206

I'm not very sure that this code is good, tell me and suggest me improvements. Also keep in mind that this is the first time I write a constraint.

### DIFF
--- a/src/TestSuite/Constraint/Email/MailContainsAttachment.php
+++ b/src/TestSuite/Constraint/Email/MailContainsAttachment.php
@@ -11,7 +11,7 @@ declare(strict_types=1);
  *
  * @copyright     Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
  * @link          https://cakephp.org CakePHP(tm) Project
- * @since         3.7.0
+ * @since         4.0.0
  * @license       https://opensource.org/licenses/mit-license.php MIT License
  */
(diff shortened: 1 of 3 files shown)


src/TestSuite/Constraint/Email/MailContainsAttachment.php
```diff
@@ -0,0 +1,77 @@
...
+ * CakePHP(tm) : Rapid Development Framework (https://cakephp.org)
+ * Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
+ *
+ * Licensed under The MIT License
+ * For full copyright and license information, please see the LICENSE.txt
+ * Redistributions of files must retain the above copyright notice.
+ *
+ * @copyright     Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
+ * @link          https://cakephp.org CakePHP(tm) Project
+ * @since         3.7.0
```
ADmad: Lies

mirko-pagliai: Yep. Which version?

cleptric: 4.0.0, as this targets 4.x 🙂

PR #13218, opened 2019-05-10
Allow request whitelisting in CsrfProtectionMiddleware.

ADmad:


dereuromark:
Your descriptions get more verbose every time :P 

Jokes aside: There seem to be test fails due to some invalid tokens (PHP7 vs PHP5?)

tests/TestCase/Http/Middleware/CsrfProtectionMiddlewareTest.php
```diff
@@ -309,4 +310,24 @@ public function testConfigurationValidate()
...
+        $request = new ServerRequest([
+            'environment' => [
+                'REQUEST_METHOD' => 'POST',
+                'ALLOW_ME' => 'let-me-pass'
+            ],
+        ]);
+        $response = new Response();
+
+        $middleware = new CsrfProtectionMiddleware();
+        $middleware->whitelistCallback(function (ServerRequestInterface $request): ?bool {
```
dereuromark: No return types for PHP5 :)

ADmad: Doh, I should have made PR for 4.x and let someone else backport it :P

ADmad:
> Your descriptions get more verbose every time :P

The code speaks for itself :loudspeaker: :slightly_smiling_face:.

ADmad:
I'll make a PR to the docs once this is approved.

src/Http/Middleware/CsrfProtectionMiddleware.php
```diff
@@ -66,6 +66,13 @@ class CsrfProtectionMiddleware
      */
     protected $_config = [];
 
+    /**
+     * Callback for allowing to skip token check for particular request.
```
markstory:
```suggestion
     * Callback for deciding whether or not to skip the token check for particular request.
```

markstory: Might also want to mention that returning true from this callback will skip the csrf check.

src/Http/Middleware/CsrfProtectionMiddleware.php
```diff
@@ -108,6 +121,22 @@ public function __invoke(ServerRequest $request, Response $response, $next)
         return $next($request, $response);
     }
 
+    /**
+     * Set callback for allowing to skip token check for particular request.
+     *
+     * The callback will receive request instance as argument and must return
+     * `true` if you want to skip token check for the particular request.
```
markstory:
```suggestion
     * `true` if you want to skip token check for the current request.
```

tests/TestCase/Http/Middleware/CsrfProtectionMiddlewareTest.php
```diff
@@ -309,4 +310,24 @@ public function testConfigurationValidate()
...
+                'REQUEST_METHOD' => 'POST',
+                'ALLOW_ME' => 'let-me-pass'
+            ],
+        ]);
+        $response = new Response();
+
+        $middleware = new CsrfProtectionMiddleware();
+        $middleware->whitelistCallback(function (ServerRequestInterface $request) {
+            return !empty($request->getServerParams()['ALLOW_ME']);
+        });
```
markstory: Couldn't you assert that the next middleware layer is called?

ADmad:
Updates done as per feedback.

markstory:
Nice work @ADmad

PR #13219, opened 2019-05-10
Update composer configs for split packages.

ADmad:


dereuromark:
We might still want to build some basic dependency checker tool etc, to verify if those constraints are sufficient and no packages that are required are missing etc.

PR #13220, opened 2019-05-11
4.x - Convert Completion shell to a command

markstory:
Convert the CompletionShell to a Command and retain the same CLI interface. Keeping the same CLI interface was important so that we don't have to have different bash-completion scripts for different versions of CakePHP.


dereuromark:
I fixed your docblock mistakes in https://github.com/cakephp/cakephp/pull/13233/commits/8915442595eb1f324046c0a0a2e3ee4ccecd1b2c

PR #13222, opened 2019-05-11
Avoid unnecessary inflection to generate file name for error message.

ADmad:
The View class already handles inflecting template name and in case of error the filename can be fetched from previous exception.

### DIFF
--- a/src/View/Cell.php
+++ b/src/View/Cell.php
@@ -181,12 +181,6 @@ abstract class Cell
 
             $builder = $this->viewBuilder()->setLayout(false);
 
-            if ($template !== null &&
-                strpos($template, '/') === false &&
-                strpos($template, '.') === false
-            ) {
-                $template = Inflector::underscore($template);
-            }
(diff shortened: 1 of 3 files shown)


src/View/Cell.php
```diff
@@ -181,12 +181,6 @@ public function render($template = null)
 
             $builder = $this->viewBuilder()->setLayout(false);
 
-            if ($template !== null &&
-                strpos($template, '/') === false &&
-                strpos($template, '.') === false
-            ) {
-                $template = Inflector::underscore($template);
```
markstory: Do we have any tests for template with absolute or relative paths.

ADmad: Not for cells specifically but I believe we do for `View` class itself.

markstory: Looks like we have `testCellRenderPluginTemplate` which was added in b1041c6b90c  for the `.` scenario.

PR #13223, opened 2019-05-11
Cleanup up static returns, fix array doc blocks.

dereuromark:
Collections that hidden iterate over sub-objects are quite the code smell if not documented properly.
I tried to fix that.

### DIFF
--- a/src/Http/Client/Response.php
+++ b/src/Http/Client/Response.php
@@ -95,7 +95,7 @@ class Response extends Message implements ResponseInterface
     /**
      * Cookie Collection instance
      *
-     * @var \Cake\Http\Cookie\CookieCollection|\Cake\Http\Cookie\Cookie[]
+     * @var \Cake\Http\Cookie\CookieCollection
      */
     protected $cookies;
 
--- a/src/Http/Client/Response.php
+++ b/src/Http/Client/Response.php
@@ -330,11 +330,13 @@ class Response extends Message implements ResponseInterface
     public function getCookie(string $name)
     {
         $this->buildCookieCollection();
-        if (!$this->cookies->has($name)) {
+
+        $cookie = $this->cookies->get($name);
+        if ($cookie === null) {
             return null;
         }
 
-        return $this->cookies->get($name)->getValue();
+        return $cookie->getValue();
     }
 
     /**
@@ -347,11 +349,10 @@ class Response extends Message implements ResponseInterface
     {
         $this->buildCookieCollection();
 
-        if (!$this->cookies->has($name)) {
+        $cookie = $this->cookies->get($name);
+        if ($cookie === null) {
             return null;
         }
-
-        $cookie = $this->cookies->get($name);
 
         return $this->convertCookieToArray($cookie);
     }
--- a/src/Http/Client/Response.php
+++ b/src/Http/Client/Response.php
@@ -95,7 +95,7 @@ class Response extends Message implements ResponseInterface
     /**
      * Cookie Collection instance
      *
-     * @var \Cake\Http\Cookie\CookieCollection
+     * @var \Cake\Http\Cookie\CookieCollection|\Cake\Http\Cookie\Cookie[]
      */
     protected $cookies;
 
@@ -256,9 +256,9 @@ class Response extends Message implements ResponseInterface
      *
      * @param int $code The status code to set.
      * @param string $reasonPhrase The status reason phrase.
-     * @return $this A copy of the current object with an updated status code.
+     * @return static A copy of the current object with an updated status code.
      */
-    public function withStatus($code, $reasonPhrase = ''): self
+    public function withStatus($code, $reasonPhrase = '')
     {
         $new = clone $this;
         $new->code = $code;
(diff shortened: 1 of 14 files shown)


src/Http/Client/Response.php
```diff
@@ -95,7 +95,7 @@ class Response extends Message implements ResponseInterface
     /**
      * Cookie Collection instance
      *
-     * @var \Cake\Http\Cookie\CookieCollection
+     * @var \Cake\Http\Cookie\CookieCollection|\Cake\Http\Cookie\Cookie[]
```
ADmad: Hmm.. how does it end up being an array?

dereuromark: It is not, that means, that it is a collection of Cookie objects, it was only an array if the first class typehint wasnt there.
Same goes for `\ArrayObject|Type[]` etc

ADmad: Okay, though that's a strange way to denote the type of elements the collection contains.

dereuromark: Afaik it is the only way for this kind of php magic of hidden iterating over an object.

ADmad:
Compiling PHP 7.2 with redis has starting failing on scrutinizer for some reason. Hence the static analysis job is not running. Could you try switching to PHP 7.3?

src/Http/Client/Response.php
ADmad: If phpstan/pslam are happy with this then meh.

dereuromark:
psalm is so strange.. it runs into parent folders.. 
```
/vendor/zendframework/zend-diactoros/src/RequestTrait.php:316:48 - Cannot concatenate with a possibly null null|int
```
which it shouldnt according to projectDirs spec.
```
<projectFiles>
        <directory name="src" />
```

ADmad:
It does go up the parent chain when extending from other libs, I actually like that.

dereuromark:
It shouldnt, this is super annoying. https://github.com/vimeo/psalm/issues/1479#issuecomment-491529577
When there is a project dir definition that must be uphold.

ADmad:
Mostly like you are facing problems because you are installing psalm through composer, I don't have any such problem running it this branch using `pslam.phar`.

src/Http/Client/Response.php
ADmad: psalm doesn't like the `|\Cake\Http\Cookie\Cookie[]` you have added. Needs to check it's doc if it supports a more sensible way of denoting the elements the collection contains.

ADmad:
Switch to PHP 7.3 didn't help, I have skiped installing redis extension for now.

ADmad:
I am communicating with scrutinizer support to get PHP compilation working again.

dereuromark:
Can we add the shim (phar) versions into stan-setup composer command?
This way local dev will be easier then.

ADmad:
Sure, the composer command can be updated in a separate PR. I just wanted to get scrutinizer to be fixed for now so the changes can be checked.

ADmad:
I have reverted the docblock changes for cookie collection property/argument which was causing pslam to throw errors https://scrutinizer-ci.com/g/cakephp/cakephp/inspections/abc6789f-1f76-47c3-a4a8-db71008917ed.

We can later come with an alternative which keeps all tools happy.

dereuromark:
But then you need to add it below in code, where it iterates over the collection:
```
/** \Cake\Http\Cookie\Cookie[] $var */
```
etc

ADmad:
psalm does not throw any errors when iterating over a collection collection instance. It knows it's iterable since `CookieCollection` implements `IteratorAggregate`.

dereuromark:
No, it sure doesnt.
But if you look into the code, the usage of it is yellow and cannot be followed/clicked through.

dereuromark:
Untitled

ADmad:
I don't use phpstorm, so I am content with making the static analyzers happy :slightly_smiling_face:. But you can add the require inline annotation if you think it's necessary.

I pushed another commit which helped removed couple of errors from pslam baseline.

I had earlier forced pushed to this branch so please use `git pull --rebase` to update your local branch if you want to add additional commits.

src/Http/Client/Response.php
```diff
@@ -330,11 +330,13 @@ public function getCookieCollection(): CookieCollection
     public function getCookie(string $name)
     {
         $this->buildCookieCollection();
-        if (!$this->cookies->has($name)) {
```
ADmad: The `has()` and `get()` methods are almost identical with the only different being `has()` return boolean if cookie exists while `get()` returns the cookie itself. So calling `has()` beforehand had no benefit.

Also since `get()` can return `null` pslam about potentially null value for the `$this->cookies->get($name)->getValue()` chain.

dereuromark: If there is a has(), a get() should not return null. the whole idea is then to use has() to check for nullable.
I would rather make this more consistent with this pattern, if you dont mind.

markstory:
Looks good to me :+1:

src/Http/Client/Response.php
ADmad: That can be addressed in a separate issue/PR if needed.

dereuromark: https://github.com/cakephp/cakephp/pull/13239

PR #13230, opened 2019-05-13
Handle host & scheme in array URLs

### LINKED ISSUE #13204: Router seems to break when host is specified in integration test
This is a **bug**

* CakePHP Version: 3.7.7

### What you did
In an integration test,
```
$this->get([
    'prefix' => 'v1',
    'controller' => 'Events',
    'action' => 'future',
    '?' => ['apikey' => $this->getApiKey()],
    '_host' => 'api.app.localhost',
    '_ssl' => true
]);
```
[The docs](https://book.cakephp.org/3.0/en/development/routing.html#matching-specific-hostnames) suggest that using `_host` this way should work for URL generation, and I'm not aware of a reason why specifying `_host` for integration tests shouldn't work.

### Why
I'm working to combine a website and its API into one codebase. It's currently set up to reject any requests for pages under the `v1` prefix if the `api.` subdomain is missing, so I'm updating the requests in API integration tests to have `api.` in their host names.

### What happened
A 404 response with this error message:
> A route matching "/https://api.app.localhost/v1/events/future" could not be found.

(Note the leading slash in the URL)

I expected the provided URL to be routed correctly and return a successful response. The array correctly resolves to `https://api.app.localhost/v1/events/future` when put through `Router::url()` and returns a normal, non-error response when loaded in a browser.

### Notes
The `MissingRouteException` exception is getting thrown in `RouteCollection::parseRequest()`, where `$request->url` is `https://api.app.localhost...`, but `$request->here` is `/https://api.app.localhost...` with a leading slash.

`$request = new ServerRequest(['url' => $url]);` in `IntegrationTestTrait::_url($url)` is taking the correct URL and returning a `$request` with the incorrect `here` property.

This code in `ServerRequest::_setConfig()`
```
if (strlen($config['url'])) {
    $uri = $uri->withPath('/' . $config['url']);
}
...
$this->here = $this->base . '/' . $this->url;
```
causes `$request->uri->path` and `$request->here` to be set with a leading slash.

Then in the `foreach` loop in `RouteCollection::parseRequest`, each possible matching path (e.g. `"/v1/events/future"` is compared against the entire URL that's pulled from `$request->uri->path`, (e.g. `"/https://api.app.localhost/v1/events/future"`) and rejected if the request's path doesn't _begin_ with the candidate path, which `$request->uri->path` never will if a protocol and host are included.

So it's looking like the router can correctly translate the provided URL array into a string (actually, the same `MissingRouteException` is encountered if a URL string is provided), but can't then get the correct route back out of that string as long as the host name is manually specified.  **Is this a bug, or am I going about this wrong?**

### Conclusion
I'm not sure what specifically to suggest for fixing it if it were a bug, but comparing `$paths` in `RouteCollection::parseRequest` to a normalized version of the current path, with `$uri->scheme . '://' . $uri->host` removed, instead of to (potentially) the full URL, might be a good direction to take.

markstory:
When handling array URLs or absolute URLs IntegrationTestCase should forward https and host data to the right place. I was able to replace our URL parsing logic with the Uri implementation from diactoros letting us save a few cycles in each test method.

Fixes #13204

### DIFF
--- a/tests/test_app/TestApp/Controller/PostsController.php
+++ b/tests/test_app/TestApp/Controller/PostsController.php
@@ -116,6 +116,18 @@ class PostsController extends AppController
         return $this->getResponse()->withHeader('X-Cake', 'custom header');
     }
 
+    public function hostData()
+    {
+        $data = [
+            'host' => $this->request->host(),
+            'isSsl' => $this->request->is('ssl'),
+        ];
+
+        debug($data);
+
+        return $this->getResponse()->withStringBody(json_encode($data));
+    }
+
     public function empty_response()
     {
         return $this->getResponse()->withStringBody('');
--- a/tests/test_app/TestApp/Controller/PostsController.php
+++ b/tests/test_app/TestApp/Controller/PostsController.php
@@ -123,8 +123,6 @@ class PostsController extends AppController
             'isSsl' => $this->request->is('ssl'),
         ];
 
-        debug($data);
-
         return $this->getResponse()->withStringBody(json_encode($data));
     }
 
--- a/src/TestSuite/IntegrationTestTrait.php
+++ b/src/TestSuite/IntegrationTestTrait.php
@@ -53,10 +53,10 @@ use Cake\Utility\Hash;
 use Cake\Utility\Security;
 use Cake\Utility\Text;
 use Cake\View\Helper\SecureFieldTokenTrait;
-use Zend\Diactoros\Uri;
 use Exception;
 use LogicException;
 use PHPUnit\Exception as PhpunitException;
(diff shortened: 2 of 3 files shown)


tests/test_app/TestApp/Controller/PostsController.php
```diff
@@ -116,6 +116,18 @@ public function header()
...
     }
 
+    public function hostData()
+    {
+        $data = [
+            'host' => $this->request->host(),
+            'isSsl' => $this->request->is('ssl'),
+        ];
+
+        debug($data);
```
ADmad: Leftover `debug()`.

PR #13234, opened 2019-05-14
Use more precise array return docs.

dereuromark:
Follow https://github.com/cakephp/cakephp/pull/13233


ADmad:
Noooo......... Both phpstan and pslam show multiple errors for this PR :slightly_frowning_face:.

ADmad:
Or maybe they were from earlier PR and fixed in another PR :man_shrugging:

markstory:
@ADmad yeah they were from an earlier pull request.

PR #13236, opened 2019-05-14
4.x 13233 followup

ADmad:
`NullEngine` returning `true` for some methods and `false` for others didn't make much sense. I changed them all to truthy values.

### DIFF
--- a/src/Cache/Engine/NullEngine.php
+++ b/src/Cache/Engine/NullEngine.php
@@ -46,7 +46,7 @@ class NullEngine extends CacheEngine
      */
     public function setMultiple($data, $ttl = null): bool
     {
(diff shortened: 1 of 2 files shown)


src/Cache/Engine/NullEngine.php
```diff
@@ -70,15 +70,15 @@ public function getMultiple($keys, $default = null): array
...
+        return 1;
     }
 
     /**
      * @inheritDoc
      */
     public function decrement(string $key, int $offset = 1)
     {
-        return false;
+        return 1;
```
dereuromark: This must be 0 then, or you keep false.

dereuromark (approved):

PR #13237, opened 2019-05-14
Return int or null from all methods writing to stdout/stderr.

ADmad:


src/TestSuite/Stub/ConsoleOutput.php
```diff
@@ -59,7 +59,7 @@ public function write($message, int $newlines = 1)
             $newlines--;
         }
 
-        return false;
+        return 0;
```
markstory: Should this be the length of the `$message`?

ADmad: Done

src/TestSuite/Stub/ConsoleOutput.php
```diff
@@ -59,7 +59,7 @@ public function write($message, int $newlines = 1)
             $newlines--;
         }
 
-        return false;
+        return strlen($message);
```
dereuromark: strlen() can operate on arrays?

ADmad: I overlooked the fact that $message can be array. Will fix it.

Returning the bytes written is necessary as it possible to overwrite a message in which case the length of previous msg is required to back track.

src/TestSuite/Stub/ConsoleOutput.php
ADmad: Actually returning a non zero value causes tests for `ProgressHelper` tests to fail. This is a stub class (though why is it under `src` beats me) and seems to intentionally returns empty value to make testing easier.

markstory: It is in src and it is part of the integration test system.

markstory (approved):

PR #13238, opened 2019-05-14
Fix return type errors reported by psalm.

ADmad:


src/View/ViewBuilder.php
```diff
@@ -526,6 +526,7 @@ public function build(
         ];
         $data += $this->_options;
 
+        /** @var \Cake\View\View */
```
dereuromark: I didnt know this is possible or necessary :)
:+1:

ADmad: Leaving out the variable name is possible https://psalm.dev/docs/supported_annotations/#off-label-usage-of-the-var-tag

PR #13239, opened 2019-05-15
Fix up cookie get() vs has(). Fixed docblocks.

dereuromark:
Follows https://github.com/cakephp/cakephp/pull/13223 
and addresses the API issues around has() vs get().
Cleaner expectation for both devs and static analyzers.

This is an important follow up, as other high level code like
CookieEquals etc, that rely on the clean nullable API here, whereas
other more internal parts should use the existing has vs get pattern as designed.

I was not sure if it is worth providing orFail() wrappers for the high levels, like Response:

```
    /**
     * Read a single cookie from the response.
     *
     * This method provides read access to pending cookies. It will
     * not read the `Set-Cookie` header if set.
     *
     * @param string $name The cookie name you want to read.
     * @return array Cookie data.
     * @throws \InvalidArgumentException If cookie not found.
     */
    public function getCookieOrFail(string $name): array
    {
        $cookie = $this->getCookie($name);
        if ($cookie === null) {
            throw new InvalidArgumentException('Cookie not found: ' . $name);
        }

        return $cookie;
    }
```

As such, I left this out for now.

### DIFF
--- a/src/Http/Client/Response.php
+++ b/src/Http/Client/Response.php
@@ -331,10 +331,11 @@ class Response extends Message implements ResponseInterface
     {
         $this->buildCookieCollection();
 
-        $cookie = $this->cookies->get($name);
-        if ($cookie === null) {
+        if (!$this->cookies->has($name)) {
             return null;
         }
+
+        $cookie = $this->cookies->get($name);
 
         return $cookie->getValue();
     }
@@ -349,7 +350,7 @@ class Response extends Message implements ResponseInterface
     {
         $this->buildCookieCollection();
 
-        $cookie = $this->cookies->get($name);
+        $cookie = $this->getCookie($name);
         if ($cookie === null) {
             return null;
         }
@@ -401,12 +402,14 @@ class Response extends Message implements ResponseInterface
     {
         $this->buildCookieCollection();
 
-        $cookies = [];
-        foreach ($this->cookies as $cookie) {
-            $cookies[$cookie->getName()] = $this->convertCookieToArray($cookie);
+        $out = [];
+        /** @var \Cake\Http\Cookie\Cookie[] $cookies */
+        $cookies = $this->cookies;
+        foreach ($cookies as $cookie) {
+            $out[$cookie->getName()] = $this->convertCookieToArray($cookie);
         }
 
-        return $cookies;
+        return $out;
     }
 
     /**
--- a/src/Http/Client/Response.php
+++ b/src/Http/Client/Response.php
@@ -335,9 +335,7 @@ class Response extends Message implements ResponseInterface
             return null;
         }
 
-        $cookie = $this->cookies->get($name);
-
-        return $cookie->getValue();
+        return $this->cookies->get($name)->getValue();
     }
 
     /**
--- a/src/Http/Client/Response.php
+++ b/src/Http/Client/Response.php
@@ -350,10 +350,11 @@ class Response extends Message implements ResponseInterface
     {
         $this->buildCookieCollection();
 
-        $cookie = $this->getCookie($name);
-        if ($cookie === null) {
+        if (!$this->cookies->has($name)) {
             return null;
         }
+
+        $cookie = $this->cookies->get($name);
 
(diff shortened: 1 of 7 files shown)


ADmad:
I am not in favor of `CookieCollection::get()` throwing exception. Cookies would often be unset and I don't want to wrap every call to `get()` with a call for `has()`, that's not user friendly.

No other methods for data stores populated from user/client data throw exceptions, for e.g. `ServerRequest::getData()`, `Session::read()` nor do methods for reading from cache or log either.

ADmad:
Even `ObjectRegistry::get()` does not throw exception in spite of have method `ObjectRegistry::has()`.

ADmad:
That said if the higher level methods `ServerRequest::getCookie()` and `Response::getCookie()` don't throw exceptions I can live with this change.

dereuromark:
I am all for consistency, but if there is has() get() should not return nullable, defeats the whole purpose of it.
We can also remove has() then so people check on the nullable return instead.

Current way would be a fair compromise on the low vs high level usage IMO.

src/Http/Client/Response.php
```diff
@@ -401,12 +403,14 @@ protected function _getCookies(): array
     {
         $this->buildCookieCollection();
 
-        $cookies = [];
-        foreach ($this->cookies as $cookie) {
-            $cookies[$cookie->getName()] = $this->convertCookieToArray($cookie);
+        $out = [];
+        /** @var \Cake\Http\Cookie\Cookie[] $cookies */
```
markstory: Isn't $this->cookies a CookieCollection?

markstory (commented):
Can't say I am a fan of throwing an exception from get(). We don't do that in other collection methods and it is a breaking change that doesn't have a related deprecation warning.

src/Http/Client/Response.php
dereuromark: Exactly, the iterating over it does provide the Cookie objects.
A bit too much hidden magic for my taste, but thats how this works here and also in other collections in Cake.

dereuromark: The correct approach would be this: https://github.com/cakephp/cakephp/pull/13223/commits/4bd026a0f716f1767409c213f1761bf864b88b6c#diff-6d806173a45816b0375aaaece3a987a7R98
But apparently, some static analyzers are not happy with this.

dereuromark:
Then you need to deprecate the has() as we apply this pattern otherwise quite consistent with this pattern.

- public function hasBehavior(string $name): bool
- public function getBehavior(string $name): Behavior - throws exception if not existent

etc

and having a has() with a nullable get() seems duplicate then for no reason.
The deprecation could easily be added in 3.8 now along with other deprecation backports.

markstory:
`getBehavior()` isn't a collection method though. It is a wrapper on `Table`. I was more thinking of how the get method works on `AssociationCollection`, `TransltorRegistry` and the various `ObjectRegistry` subclasses work. However, I also see that `CommandCollection`, `PluginCollection`, `ConnectionManager` and `WidgetRegistry` raise errors on missing items.

Given that we're split already perhaps normalizing on throwing would be better, and we could add deprecation warnings to 3.8.

dereuromark:
Exactly, if we are talking about pure data, get() or read() etc can be nullable.
But here we talk about concrete objects of those collections, and here, especially with the has() being present, we should keep things clean and consistent.
Static analyzers will also be happier when the get() result is used somewhere directly (error "possible null pointer exception" otherwise). So we should more and more go into this direction on those object collections and the has/get pattern.

ADmad (approved):

src/Http/Cookie/CookieCollection.php
```diff
@@ -127,7 +128,12 @@ public function get(string $name): ?CookieInterface
             }
         }
 
-        return null;
+        throw new InvalidArgumentException(
+            sprintf(
+                'Cookie %s not found. Use has() to check first for existence.',
```
markstory: :+1: nice error message.

PR #13241, opened 2019-05-15
Fix Router fullBaseUrl

dereuromark:
I had type errors here which should have been speaking dev errors:

> 14) Tools\Test\TestCase\View\Helper\FormHelperTest::testCreate
TypeError: Return value of Cake\Routing\Router::fullBaseUrl() must be of the type string, null returned

### DIFF
--- a/src/Routing/Router.php
+++ b/src/Routing/Router.php
@@ -52,7 +52,7 @@ class Router
      * Contains the base string that will be applied to all generated URLs
      * For example `https://example.com`
      *
-     * @var string
+     * @var string|null
      */
     protected static $_fullBaseUrl;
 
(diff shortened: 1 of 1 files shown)


markstory:
Looks like we are relying on baseUrl being undefined in our tests. I would expect plugins and applications to also not define this value.

dereuromark:
The value injected was an empty string, thus the check with strict comparison didnt work without casting.
Now it works.
The value must be at least an empty string, null is not allowed.
Null basically means you have forgotten it - which now will be more clear/verbose than the error message above.

markstory:
Why not cast on read instead of failing only when the key is missing/defined as null. Casting would let us ensure that all downstream code sees a string and cover the null/false/undefined cases.

dereuromark:
I had this specific issue in tests for a plugin, as the config key was missing.
This was just making it more obvious what was needed to be done.
We could sure cloak it - let the code further down the way fail then. But isnt this only an issue for these tests? As in normal app mode you would always have this config key defined anways?

dereuromark:
I guess we could try to make it cast to string instead. I am just a bit worried if that makes sense?

markstory:
> I guess we could try to make it cast to string instead. I am just a bit worried if that makes sense?

All the places fullBaseUrl() are used it is appended to another string. Whether the key is undefined, null, false, or '' the effective outcome would be `''`. I think this is a place where we can favour ergonomics to help provide type safety elsewhere.

dereuromark:
All good here?

PR #13247, opened 2019-05-16
4.x types

dereuromark:
Switching assertEquals() to assertSame() for scalar values finds quite some issues in both src return types but also the tests itself. Wrong expectations here can lead to bad coding.
It is generally a code smell to assert too weakly on scalar values, we should go away from assertEquals() here in general more and more.

I switched over a few trivial cases, we can do more batches in the future.
Especially around falsey values this is super important, but also now on all scalar returns that are not yet typehinted but expected as a type to go into another typehinted method. The stricter the checks here in tests, the more likely this will not blow up later in code that uses the return values (and trusts the documented type).

### DIFF
--- a/tests/TestCase/BasicsTest.php
+++ b/tests/TestCase/BasicsTest.php
@@ -40,19 +40,19 @@ class BasicsTest extends TestCase
         $two = ['one' => 'one', 'two' => 'two'];
         $result = array_diff_key($one, $two);
         $expected = ['three' => 3];
-        $this->assertEquals($expected, $result);
+        $this->assertSame($expected, $result);
 
         $one = ['one' => ['value', 'value-two'], 'two' => 2, 'three' => 3];
         $two = ['two' => 'two'];
         $result = array_diff_key($one, $two);
         $expected = ['one' => ['value', 'value-two'], 'three' => 3];
-        $this->assertEquals($expected, $result);
+        $this->assertSame($expected, $result);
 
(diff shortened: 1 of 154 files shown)


tests/TestCase/BasicsTest.php
```diff
@@ -40,19 +40,19 @@ public function testArrayDiffKey()
...
         $expected = ['three' => 3];
-        $this->assertEquals($expected, $result);
+        $this->assertSame($expected, $result);
 
         $one = ['one' => ['value', 'value-two'], 'two' => 2, 'three' => 3];
         $two = ['two' => 'two'];
         $result = array_diff_key($one, $two);
         $expected = ['one' => ['value', 'value-two'], 'three' => 3];
-        $this->assertEquals($expected, $result);
+        $this->assertSame($expected, $result);
```
markstory: Won't assertSame() fail if the array key order changes? Is that a useful constraint to put on ourselves?

ADmad: Yeah assertSame() should be used for scalars only not arrays.

dereuromark: In simple arrays like this it might also OK, but I see your point.

dereuromark: So should we keep these simple arrays in strict assert? It has the benefit of checking all the types here correctly, too.
As long as their order is to be expected exactly like this, we can keep them IMO to have the better assert. More high level comparisons sure should stay less strict.

dereuromark:
fails are unrelated.

PR #13248, opened 2019-05-16
Fix up get() for object registry.

dereuromark:
Follow https://github.com/cakephp/cakephp/pull/13239

This is another important fix for API.
If you get() the object, you must be able to trust it exists, especially since has() exists to check.

As you can see from the baseline (and the code), everyone blindly uses it in chaining, risking null pointer errors right now

    $this->get('Foo')->doSth();

Static analyzers, IDEs are not happy about this, for a good reason.
We need to become a bit cleaner here on those things.

I wasn't sure if we should also change the `__get()` call, as property access would then require real isset() wrapping to avoid the exception. Maybe sth for 5.0?


src/Core/ObjectRegistry.php
```diff
@@ -208,26 +208,27 @@ public function has(string $name): bool
...
-     * @return object|null Object instance if loaded else null.
+     * @return object Object instance.
+     * @throws \RuntimeException If not loaded or found.
      */
-    public function get(string $name): ?object
+    public function get(string $name): object
     {
-        if (isset($this->_loaded[$name])) {
-            return $this->_loaded[$name];
+        if (!isset($this->_loaded[$name])) {
```
ADmad: Instead of negating the return of `isset()` could return the value here and throw exception after the `if` block.

src/Core/ObjectRegistry.php
```diff
@@ -208,26 +208,27 @@ public function has(string $name): bool
...
      * Provide public read access to the loaded objects
      *
      * @param string $name Name of property to read
-     * @return mixed
+     * @return object|null
      */
     public function __get(string $name)
     {
-        return $this->get($name);
+        return $this->has($name) ? $this->get($name) : null;
```
ADmad: `return $this->_loaded[$name] ?? null` would avoid extra methods calls and their internal checks. It would be easier to read too.

src/ORM/Table.php
```diff
@@ -788,16 +788,17 @@ public function behaviors(): BehaviorRegistry
      */
     public function getBehavior(string $name): Behavior
     {
-        /** @var \Cake\ORM\Behavior|null $behavior */
-        $behavior = $this->_behaviors->get($name);
-        if ($behavior === null) {
+        if (!$this->_behaviors->has($name)) {
```
ADmad: Instead of this `has()` check we could just call `get()` in a try / catch block and re-throw the exception with specific message if `has()` throws exception.

In majority of the cases the `has()` call will succeed and hence the extra method call will be avoided.

ADmad (commented):
Changing `ObjectRegistry::get()` to throw exception while retaining existing behavior of `ObjectRegistry::__get()` returning `null` is a reasonable compromise :+1:.

ADmad:
Psalm is throwing errors from files not related to registry classes changes. Seems the ignores from it's baseline file have been prematurely removed.

dereuromark:
How can this happen?

src/ORM/Table.php
dereuromark: Try catch is probably slower and i find it for code flow usually less nice than normal methods.

ADmad:
I don't know :)  Trying reverting to the baseline from parent commit. Regenerating it again could mask issues introduced in these changes (if any).

src/Core/ObjectRegistry.php
dereuromark: I can catch the value but the flow in code should always prioritize the happy case towards the end.

dereuromark:
Can we restore the previous and try to re-do it?

src/ORM/Table.php
```diff
@@ -788,16 +788,17 @@ public function behaviors(): BehaviorRegistry
...
             throw new InvalidArgumentException(sprintf(
                 'The %s behavior is not defined on %s.',
                 $name,
                 static::class
             ));
         }
 
+        /** @var \Cake\ORM\Behavior|null $behavior */
+        $behavior = $this->_behaviors->get($name);
```
ADmad: Why the `|null`? `get()` doesn't return `null` anymore right? Even the assignment is not necessary, can directly return the value:

```
/** @var \Cake\ORM\Behavior */
return $this->_behaviors->get($name);
```

dereuromark: True

ADmad: Ah yes since the parent class has `object` as return type. Nevermind then.

dereuromark:
I know what happened.. I switched branches back from https://github.com/cakephp/cakephp/pull/13247
And then I reran it. Not sure why it wouldnt change that file back, but if you merge the other PR the errors here should also disappear as fixed.

ADmad:
#13247 updates only 1 file under `src` rest all all changes to test files. So those changes can't cause so many changes to the baseline.

ADmad:
I pushed a commit to revert changes to the baseline file.

ADmad (approved):

ADmad:
I pushed an additional commit to have cleaner typing for `_resolveClassName()`.

Also made a docs PR.

PR #13251, opened 2019-05-17
Set reponse instance returned by action to Controller::$response.

### LINKED ISSUE #13249: Controller::afterFilter() as no access to the response returned from an action
[x] enhancement
* CakePHP Version: master

**AController.php:**
```
...
public function some() {
    return $this->response->withStringBody('empty');
}
public afterFilter($event) {
    return $this->response->withHeader('additional', 'stuff');
}
...
```

I expected a response with the string 'empty' as the body and the header 'additional' set to 'stuff'. I got a response with the header, but not the string body.

FROM IRC:

> hmic: Is there a way to access the resposne returned from an action in a afterFilter callback? - Of course I can access $this->response from it, but that does not (necessarily) reflect whats returned from the controller action. - Or any other solution on how to change the response of any action (i just want to add additional headers) in the afterFilter callback?
It's just weird that it (afterFilter) has no access to the response object that has been returned by an action, but is able to return a response that will be send...

> admad: hmm.. yeah Controller::$response never get's updated if an action explicitly return response.  It's updated only when render() is called.

> hmic: Either i need that be accurate in the afterFilter callback, or I need access to the response object that has been returned from the action in the $event data, somehow.

> admad: Fix is easy, ActionDispatcher::_invoke() should check and set return value of invokeAction() call.

ADmad:
This ensures afterFilter() has access to the updated response.
Closes #13249.

### DIFF
--- a/src/Controller/Controller.php
+++ b/src/Controller/Controller.php
@@ -607,7 +607,12 @@ class Controller implements EventListenerInterface, EventDispatcherInterface
         /* @var callable $callable */
         $callable = [$this, $request->getParam('action')];
 
-        return $callable(...array_values($request->getParam('pass')));
+        $result = $callable(...array_values($request->getParam('pass')));
+        if ($result instanceof Response) {
+            $this->response = $result;
+        }
+
+        return $result;
     }
 
     /**
--- a/tests/TestCase/Controller/ControllerTest.php
+++ b/tests/TestCase/Controller/ControllerTest.php
@@ -131,7 +131,7 @@ class TestController extends ControllerTestAppController
 
     public function returner()
     {
-        return 'I am from the controller.';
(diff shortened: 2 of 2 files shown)


src/Controller/Controller.php
```diff
@@ -607,7 +607,12 @@ public function invokeAction()
         /* @var callable $callable */
         $callable = [$this, $request->getParam('action')];
 
-        return $callable(...array_values($request->getParam('pass')));
+        $result = $callable(...array_values($request->getParam('pass')));
+        if ($result instanceof Response) {
```
ADmad: Ideally this should check for `ResponseInterface` but I have kept it as `Response` since that's what `ActionDispatcher` also uses to check return value of various controller methods and event listeners.

We can try and change those to `ResponseInterface` checks in 4.0.

PR #13252, opened 2019-05-17
Add CSP middleware

Iandenh:
Adds a CSP middleware to 4.x based on #11115 

I'm not sure if the changes in composer.json are the right way.

### DIFF
--- a/src/Http/Middleware/CspMiddleware.php
+++ b/src/Http/Middleware/CspMiddleware.php
@@ -1,0 +1,84 @@
+<?php
+declare(strict_types=1);
+
+/**
+ * CakePHP(tm) : Rapid Development Framework (http://cakephp.org)
+ * Copyright (c) Cake Software Foundation, Inc. (http://cakefoundation.org)
+ *
+ * Licensed under The MIT License
+ * For full copyright and license information, please see the LICENSE.txt
+ * Redistributions of files must retain the above copyright notice.
+ *
+ * @copyright     Copyright (c) Cake Software Foundation, Inc. (http://cakefoundation.org)
+ * @link          http://cakephp.org CakePHP(tm) Project
+ * @since         4.0.0
+ * @license       http://www.opensource.org/licenses/mit-license.php MIT License
+ */
+
+namespace Cake\Http\Middleware;
+
+use Cake\Core\Configure;
+use InvalidArgumentException;
+use ParagonIE\CSPBuilder\CSPBuilder;
+use Psr\Http\Message\ResponseInterface;
+use Psr\Http\Message\ServerRequestInterface;
+use Psr\Http\Server\MiddlewareInterface;
+use Psr\Http\Server\RequestHandlerInterface;
+
+/**
+ * Content Security Policy Middleware
+ */
+class CspMiddleware implements MiddlewareInterface
+{
+    /**
+     * CSP Builder
+     *
+     * @var \ParagonIE\CSPBuilder\CSPBuilder|array|null $csp CSP Builder or config array
+     */
+    protected $csp;
+
+    /**
+     * Constructor
+     *
+     * @param \ParagonIE\CSPBuilder\CSPBuilder|array|null $csp CSP object or config array
+     * @throws \Exception
+     */
+    public function __construct($csp = null)
+    {
+        if ($csp === null) {
+            $cspConfig = (array)Configure::read('App.CSP');
+        } else {
+            $cspConfig = $csp;
+        }
+
+        if (!empty($cspConfig) && is_array($cspConfig)) {
+            $this->csp = CSPBuilder::fromData(json_encode($cspConfig));
+
+            return;
+        }
+
+        if (!$csp instanceof CSPBuilder) {
+            throw new InvalidArgumentException(sprintf(
+                'Expected `%s`, `%s` given.',
+                CSPBuilder::class,
+                gettype($csp)
+            ));
+        }
+
+        $this->csp = $csp;
+    }
+
+    /**
+     * Serve assets if the path matches one.
+     *
+     * @param \Psr\Http\Message\ServerRequestInterface $request The request.
+     * @param \Psr\Http\Server\RequestHandlerInterface $handler The request handler.
+     * @return \Psr\Http\Message\ResponseInterface A response.
+     */
+    public function process(ServerRequestInterface $request, RequestHandlerInterface $handler): ResponseInterface
+    {
+        $response = $handler->handle($request);
+
+        return $this->csp->injectCSPHeader($response);
+    }
+}
--- a/src/Http/Middleware/CspMiddleware.php
+++ b/src/Http/Middleware/CspMiddleware.php
@@ -52,7 +52,7 @@ class CspMiddleware implements MiddlewareInterface
         }
 
         if (!empty($cspConfig) && is_array($cspConfig)) {
-            $this->csp = CSPBuilder::fromData(json_encode($cspConfig));
+            $this->csp = new CSPBuilder($cspConfig);
 
             return;
         }
--- a/src/Http/Middleware/CspMiddleware.php
+++ b/src/Http/Middleware/CspMiddleware.php
@@ -33,7 +33,7 @@ class CspMiddleware implements MiddlewareInterface
     /**
      * CSP Builder
      *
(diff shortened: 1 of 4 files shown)


src/Http/Middleware/CspMiddleware.php
```diff
@@ -0,0 +1,87 @@
...
+    /**
+     * Constructor
+     *
+     * @param \ParagonIE\CSPBuilder\CSPBuilder|array|null $csp CSP object or config array
+     * @throws \Exception
+     */
+    public function __construct($csp = null)
+    {
+        if ($csp === null) {
+            $cspConfig = (array)Configure::read('App.CSP');
```
ADmad: It's better to avoid directly reading from `Configure`. Users can just pass array to the constructor.

src/Http/Middleware/CspMiddleware.php
```diff
@@ -0,0 +1,87 @@
...
+            $this->csp = CSPBuilder::fromData(json_encode($cspConfig));
+
+            return;
+        }
+
+        if (!$csp instanceof CSPBuilder) {
+            throw new InvalidArgumentException(sprintf(
+                'Expected `%s`, `%s` given.',
+                CSPBuilder::class,
+                gettype($csp)
```
ADmad: You can use `getTypeName()` instead from `Core\functions.php`.

src/Http/Middleware/CspMiddleware.php
```diff
@@ -0,0 +1,87 @@
...
+     * @param \Psr\Http\Message\ServerRequestInterface $request The request.
+     * @param \Psr\Http\Server\RequestHandlerInterface $handler The request handler.
+     * @return \Psr\Http\Message\ResponseInterface A response.
+     */
+    public function process(ServerRequestInterface $request, RequestHandlerInterface $handler): ResponseInterface
+    {
+        $response = $handler->handle($request);
+
+        /** @var \Psr\Http\Message\ResponseInterface $response */
+        $response = $this->csp->injectCSPHeader($response);
```
ADmad: Can directly return response by changing to:

```
/** @var \Psr\Http\Message\ResponseInterface */
return $this->csp->injectCSPHeader($response);
```

src/Http/Middleware/CspMiddleware.php
dereuromark: I like the internal handling in case it wasnt provided. We should do this more often where input is optional, and keeps it more DRY.

dereuromark: CSP would better be Csp as we dont do acronym uppercases in code and config.
But if we are dropping this config now, never mind.

src/Http/Middleware/CspMiddleware.php
dereuromark: He tried afaik, but psalm wasnt happy. See the previous commit.

ADmad: Earlier commit which directly returns instance doesn't have the `@var` annotation like I have shown. For returns you have to specify just the type without variable name.

dereuromark: Right, that new thing :)

src/Http/Middleware/CspMiddleware.php
ADmad: Passing the array to constructor is easy enough when adding the middleware to queue in your application class so no need to try reading from `Configure`. Things are less magical this way and improves testability of middleware as all data is explicitly injected.

src/Http/Middleware/CspMiddleware.php
```diff
@@ -0,0 +1,87 @@
...
+    public function __construct($csp = null)
+    {
+        if ($csp === null) {
+            $cspConfig = (array)Configure::read('App.CSP');
+        } else {
+            $cspConfig = $csp;
+        }
+
+        if (!empty($cspConfig) && is_array($cspConfig)) {
+            $this->csp = CSPBuilder::fromData(json_encode($cspConfig));
```
ndm2: Couldn't the config be passed directly to `CSPBuilder::fromArray()`, or the `CSPBuilder` constructor?

src/Http/Middleware/CspMiddleware.php
Iandenh: I tried that locally, but looks like PHPstan doesn't understead the annotation correctly.
Or should I add the PHPstan error to `ignoreErrors`

src/Http/Middleware/CspMiddleware.php
Iandenh: It now also possible to pass a `CSPBuilder` instance to `CspMiddleware` see https://github.com/cakephp/cakephp/pull/13252/files#diff-3faf87dd86f33d27dcb8b089b5644176R96

src/Http/Middleware/CspMiddleware.php
Iandenh: This way was copied from #11115 
But I can remove it of course

src/Http/Middleware/CspMiddleware.php
ADmad: @landenh But why json encode the array and pass to fromData() instead of using fromArray()?

Iandenh: ohh, sorry I misunderstood

src/Http/Middleware/CspMiddleware.php
```diff
@@ -52,7 +52,7 @@ public function __construct($csp = null)
         }
 
         if (!empty($cspConfig) && is_array($cspConfig)) {
```
dereuromark: Why do we need is_array()? If the payload is here, wouldnt the CSPBuilder class throw exception for the wrong type?

ADmad:
@Iandenh I bet you thought this would be a simple PR and didn't expect so many comments :slightly_smiling_face:. I hope you can appreciate our attention to detail and understand why they are important.

src/Http/Middleware/CspMiddleware.php
dereuromark: I see, you are later using the object.

ADmad: Would passing an empty array to the builder even be useful? If not I would throw an exception if `if (empty($cspConfig))` succeeds.

Iandenh: I can also make it as simple as:
```php
    public function __construct($csp = null)
    {
        if ($csp === null) {
            $csp = (array)Configure::read('App.CSP');
        }

        if (!$csp instanceof CSPBuilder) {
            $csp = new CSPBuilder($csp);
        }

        $this->csp = $csp;
    }
```

ADmad: No use of `Configure::read()` please. Allow passing array or `CSPBuilder()` as argument, without the default `null`.

markstory:
> I'm not sure if the changes in composer.json are the right way.

Putting csp builder as an optional dependency looks good to me.

src/Http/Middleware/CspMiddleware.php
```diff
@@ -0,0 +1,69 @@
...
+    /**
+     * Constructor
+     *
+     * @param \ParagonIE\CSPBuilder\CSPBuilder|array $csp CSP object or config array
+     * @throws \Exception
+     */
+    public function __construct($csp)
+    {
+        if (!$csp instanceof CSPBuilder) {
+            $csp = new CSPBuilder($csp);
```
markstory: Should we throw an exception if the CSPBuilder class doesn't exist and let the user know how they can install it?

ADmad (approved):

Iandenh:
@ADmad No problem, it results in better code :smile: 

@markstory I have added an exception if the class `CSPBuilder` doesn't exist

src/Http/Middleware/CspMiddleware.php
```diff
@@ -0,0 +1,72 @@
...
+     *
+     * @var \ParagonIE\CSPBuilder\CSPBuilder $csp CSP Builder or config array
+     */
+    protected $csp;
+
+    /**
+     * Constructor
+     *
+     * @param \ParagonIE\CSPBuilder\CSPBuilder|array $csp CSP object or config array
+     * @throws \Exception
```
dereuromark: RuntimeException

PR #13228, opened 2019-05-12
PHPStan fixes.

dereuromark:
In my experience PHPStan level 5 is a good target to reach, as it finds important interface contracting issues, while not being too pedantic about nullable etc.
But we are still quite a bit away from reaching this clean code state yet.

I managed to get level 5 fails down from 54 to 45, remaining is now:

```
 ------ ------------------------------------------------------------------------------------------ 
  Line   Collection/CollectionTrait.php (in context of class Cake\Collection\Collection)           
 ------ ------------------------------------------------------------------------------------------ 
  528    Parameter #1 $it of method AppendIterator::append() expects Iterator, Traversable given.  
  529    Parameter #1 $it of method AppendIterator::append() expects Iterator, Traversable given.  
 ------ ------------------------------------------------------------------------------------------ 

 ------ -------------------------------------------------------------------------------------------- 
  Line   Collection/CollectionTrait.php (in context of class Cake\Collection\Iterator\TreeIterator)  
 ------ -------------------------------------------------------------------------------------------- 
  528    Parameter #1 $it of method AppendIterator::append() expects Iterator, Traversable given.    
  529    Parameter #1 $it of method AppendIterator::append() expects Iterator, Traversable given.    
 ------ -------------------------------------------------------------------------------------------- 

 ------ ------------------------------------------------------------------------------------------- 
  Line   Collection/CollectionTrait.php (in context of class Cake\Collection\Iterator\TreePrinter)  
 ------ ------------------------------------------------------------------------------------------- 
  528    Parameter #1 $it of method AppendIterator::append() expects Iterator, Traversable given.   
  529    Parameter #1 $it of method AppendIterator::append() expects Iterator, Traversable given.   
 ------ ------------------------------------------------------------------------------------------- 

 ------ ------------------------------------------------------------------------------------------- 
  Line   Collection/CollectionTrait.php (in context of class Cake\Collection\Iterator\ZipIterator)  
 ------ ------------------------------------------------------------------------------------------- 
  528    Parameter #1 $it of method AppendIterator::append() expects Iterator, Traversable given.   
  529    Parameter #1 $it of method AppendIterator::append() expects Iterator, Traversable given.   
 ------ ------------------------------------------------------------------------------------------- 

 ------ ------------------------------------------------------------------------------------------ 
  Line   Collection/CollectionTrait.php (in context of class Cake\ORM\ResultSet)                   
 ------ ------------------------------------------------------------------------------------------ 
[...]

PR #13254, opened 2019-05-17
4.x phpstan5

dereuromark:
Follow https://github.com/cakephp/cakephp/pull/13228

We are very close to proudly say that Cake runs on phpstan level 5.

We are lying a bit - but overall it is better to exclude 2-3 interface contract violations in favor of having the level 5 protection in place now for new code.
That said, we should still try to clean up the silenced issues, and have cleaner interfaces and interface aggregation here in general moving forward.

Are we OK with silencing the baseline in psalm here for those?

### DIFF
--- a/src/Database/Schema/CachedCollection.php
+++ b/src/Database/Schema/CachedCollection.php
@@ -70,7 +70,7 @@ class CachedCollection implements CollectionInterface
     /**
      * {@inheritDoc}
      */
-    public function describe(string $name, array $options = []): TableSchema
+    public function describe(string $name, array $options = []): TableSchemaInterface
     {
         $options += ['forceRefresh' => false];
         $cacheConfig = $this->getCacheMetadata();
--- a/src/Database/Schema/CachedCollection.php
+++ b/src/Database/Schema/CachedCollection.php
@@ -70,7 +70,7 @@ class CachedCollection implements CollectionInterface
     /**
      * {@inheritDoc}
      */
-    public function describe(string $name, array $options = []): TableSchemaInterface
+    public function describe(string $name, array $options = []): TableSchema
     {
         $options += ['forceRefresh' => false];
         $cacheConfig = $this->getCacheMetadata();
--- a/src/Database/Schema/Collection.php
+++ b/src/Database/Schema/Collection.php
@@ -86,10 +86,10 @@ class Collection implements CollectionInterface
      *
      * @param string $name The name of the table to describe.
      * @param array $options The options to use, see above.
-     * @return \Cake\Database\Schema\TableSchemaInterface Object with column metadata.
+     * @return \Cake\Database\Schema\TableSchema Object with column metadata.
      * @throws \Cake\Database\Exception when table cannot be described.
      */
-    public function describe(string $name, array $options = []): TableSchemaInterface
+    public function describe(string $name, array $options = []): TableSchema
     {
         $config = $this->_connection->config();
         if (strpos($name, '.')) {
--- a/src/Database/Schema/Collection.php
+++ b/src/Database/Schema/Collection.php
@@ -89,7 +89,7 @@ class Collection implements CollectionInterface
      * @return \Cake\Database\Schema\TableSchema Object with column metadata.
      * @throws \Cake\Database\Exception when table cannot be described.
      */
-    public function describe(string $name, array $options = []): TableSchema
+    public function describe(string $name, array $options = []): TableSchemaInterface
     {
         $config = $this->_connection->config();
         if (strpos($name, '.')) {
--- a/src/Database/Schema/SqlGeneratorInterface.php
+++ b/src/Database/Schema/SqlGeneratorInterface.php
@@ -16,7 +16,7 @@ declare(strict_types=1);
  */
 namespace Cake\Database\Schema;
 
-use Cake\Database\Connection;
+use Cake\Datasource\ConnectionInterface;
 
 /**
  * An interface used by TableSchema objects.
@@ -29,11 +29,11 @@ interface SqlGeneratorInterface
      * Uses the connection to access the schema dialect
      * to generate platform specific SQL.
      *
-     * @param \Cake\Database\Connection $connection The connection to generate SQL for.
+     * @param \Cake\Datasource\ConnectionInterface $connection The connection to generate SQL for.
      * @return array List of SQL statements to create the table and the
      *    required indexes.
      */
-    public function createSql(Connection $connection): array;
+    public function createSql(ConnectionInterface $connection): array;
 
     /**
      * Generate the SQL to drop a table.
@@ -41,32 +41,32 @@ interface SqlGeneratorInterface
      * Uses the connection to access the schema dialect to generate platform
      * specific SQL.
      *
-     * @param \Cake\Database\Connection $connection The connection to generate SQL for.
+     * @param \Cake\Datasource\ConnectionInterface $connection The connection to generate SQL for.
      * @return array SQL to drop a table.
      */
-    public function dropSql(Connection $connection): array;
+    public function dropSql(ConnectionInterface $connection): array;
 
     /**
      * Generate the SQL statements to truncate a table
      *
-     * @param \Cake\Database\Connection $connection The connection to generate SQL for.
+     * @param \Cake\Datasource\ConnectionInterface $connection The connection to generate SQL for.
      * @return array SQL to truncate a table.
      */
-    public function truncateSql(Connection $connection): array;
+    public function truncateSql(ConnectionInterface $connection): array;
 
     /**
      * Generate the SQL statements to add the constraints to the table
      *
-     * @param \Cake\Database\Connection $connection The connection to generate SQL for.
+     * @param \Cake\Datasource\ConnectionInterface $connection The connection to generate SQL for.
      * @return array SQL to add the constraints.
      */
-    public function addConstraintSql(Connection $connection): array;
+    public function addConstraintSql(ConnectionInterface $connection): array;
 
     /**
      * Generate the SQL statements to drop the constraints to the table
      *
-     * @param \Cake\Database\Connection $connection The connection to generate SQL for.
+     * @param \Cake\Datasource\ConnectionInterface $connection The connection to generate SQL for.
      * @return array SQL to drop a table.
      */
-    public function dropConstraintSql(Connection $connection): array;
+    public function dropConstraintSql(ConnectionInterface $connection): array;
 }
--- a/src/Database/Schema/SqlGeneratorInterface.php
+++ b/src/Database/Schema/SqlGeneratorInterface.php
@@ -16,7 +16,7 @@ declare(strict_types=1);
  */
 namespace Cake\Database\Schema;
 
-use Cake\Datasource\ConnectionInterface;
+use Cake\Database\Connection;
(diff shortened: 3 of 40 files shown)


dereuromark:
The baseline looks now like this - which looks also much better:
```
Index: psalm-baseline.xml
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
--- psalm-baseline.xml	(revision 8f223cd862d3a775e73f5f47c7beff300f973503)
+++ psalm-baseline.xml	(date 1558125307000)
@@ -32,9 +32,6 @@
     <PropertyNotSetInConstructor occurrences="1">
       <code>DigestAuthenticate</code>
     </PropertyNotSetInConstructor>
-    <TypeDoesNotContainType occurrences="1">
-      <code>empty($req)</code>
-    </TypeDoesNotContainType>
   </file>
   <file src="src/Auth/FormAuthenticate.php">
     <PossiblyInvalidArgument occurrences="2">
@@ -521,15 +518,6 @@
       <code>static::$_dsnClassMap</code>
     </UndefinedPropertyFetch>
   </file>
-  <file src="src/Core/functions.php">
-    <InvalidScalarArgument occurrences="5">
-      <code>$name</code>
-      <code>$name</code>
-      <code>$filename</code>
-      <code>env('DOCUMENT_ROOT')</code>
-      <code>env('SCRIPT_FILENAME')</code>
-    </InvalidScalarArgument>
-  </file>
   <file src="src/Database/Connection.php">
     <MissingClosureReturnType occurrences="1">
       <code>function () use ($callback) {</code>
@@ -1396,13 +1384,6 @@
     <InvalidScalarArgument occurrences="1">
       <code>$time</code>
     </InvalidScalarArgument>
-    <PossiblyInvalidArgument occurrences="5">
-      <code>env('HTTP_ACCEPT_ENCODING')</code>
-      <code>env('HTTP_ACCEPT_ENCODING')</code>
-      <code>$agent</code>
-      <code>$agent</code>
-      <code>$httpRange</code>
-    </PossiblyInvalidArgument>
     <PossiblyInvalidOperand occurrences="3">
       <code>$end</code>
       <code>$end</code>
@@ -1527,13 +1508,6 @@
     </MissingParamType>
   </file>
   <file src="src/I18n/Number.php">
-    <InvalidArgument occurrences="5">
-      <code>$size</code>
-      <code>$size / 1024</code>
-      <code>$size / 1024 / 1024</code>
-      <code>$size / 1024 / 1024 / 1024</code>
-      <code>$size / 1024 / 1024 / 1024 / 1024</code>
-    </InvalidArgument>
     <PossiblyInvalidOperand occurrences="1">
       <code>$value</code>
     </PossiblyInvalidOperand>
@@ -1573,20 +1547,6 @@
     </PossiblyUndefinedArrayOffset>
   </file>
   <file src="src/I18n/RelativeTimeFormatter.php">
-    <InvalidArgument occurrences="12">
-      <code>$count</code>
-      <code>$count</code>
-      <code>$count</code>
-      <code>$count</code>
-      <code>$count</code>
-      <code>$count</code>
-      <code>$count</code>
-      <code>$message</code>
-      <code>$message</code>
-      <code>$message</code>
-      <code>$message</code>
-      <code>'just now'</code>
-    </InvalidArgument>
     <PossiblyNullArgument occurrences="1">
       <code>$other</code>
     </PossiblyNullArgument>
@@ -1682,9 +1642,6 @@
       <code>$item</code>
       <code>$key</code>
     </MissingClosureParamType>
-    <PossiblyInvalidArgument occurrences="1">
-      <code>env('HTTP_HOST')</code>
-    </PossiblyInvalidArgument>
     <PossiblyNullArgument occurrences="1">
       <code>$this-&gt;appCharset</code>
     </PossiblyNullArgument>
@@ -1698,9 +1655,6 @@
     </PossiblyNullArgument>
   </file>
   <file src="src/Mailer/Transport/SmtpTransport.php">
-    <PossiblyInvalidArgument occurrences="1">
-      <code>$httpHost</code>
-    </PossiblyInvalidArgument>
     <PossiblyInvalidIterator occurrences="1">
       <code>$lines</code>
     </PossiblyInvalidIterator>
@@ -1927,8 +1881,7 @@
     </PropertyNotSetInConstructor>
   </file>
   <file src="src/ORM/EagerLoadable.php">
-    <PropertyNotSetInConstructor occurrences="5">
-      <code>$_instance</code>
+    <PropertyNotSetInConstructor occurrences="4">
       <code>$_aliasPath</code>
       <code>$_propertyPath</code>
       <code>$_forMatching</code>
@@ -1951,11 +1904,8 @@
     <PossiblyNullArrayOffset occurrences="1">
       <code>$collectKeys</code>
     </PossiblyNullArrayOffset>
-    <PossiblyNullReference occurrences="4">
-      <code>getSource</code>
+    <PossiblyNullReference occurrences="1">
       <code>normalized</code>
-      <code>getTarget</code>
-      <code>requiresKeys</code>
     </PossiblyNullReference>
   </file>
   <file src="src/ORM/LazyEagerLoader.php">
@@ -2431,7 +2381,7 @@
     </PropertyNotSetInConstructor>
   </file>
   <file src="src/TestSuite/Fixture/FixtureManager.php">
-    <MissingClosureParamType occurrences="9">
+    <MissingClosureParamType occurrences="7">
       <code>$db</code>
       <code>$fixtures</code>
       <code>$db</code>
@@ -2800,9 +2750,7 @@
     <ImplementedReturnTypeMismatch occurrences="1">
       <code>string</code>
     </ImplementedReturnTypeMismatch>
-    <InvalidArgument occurrences="3">
-      <code>$modelName</code>
-      <code>$modelName</code>
+    <InvalidArgument occurrences="1">
       <code>$options['disabled']</code>
     </InvalidArgument>
     <PossiblyNullArgument occurrences="3">
```

src/Database/Schema/CachedCollection.php
```diff
@@ -70,7 +70,7 @@ public function listTables(): array
     /**
      * {@inheritDoc}
      */
-    public function describe(string $name, array $options = []): TableSchemaInterface
+    public function describe(string $name, array $options = []): TableSchema
```
markstory: This might cause problems with DebugKit as it has an implementation of Schema\Collection as well.

src/Database/Schema/Collection.php
```diff
@@ -86,10 +86,10 @@ public function listTables(): array
      *
      * @param string $name The name of the table to describe.
      * @param array $options The options to use, see above.
-     * @return \Cake\Database\Schema\TableSchemaInterface Object with column metadata.
+     * @return \Cake\Database\Schema\TableSchema Object with column metadata.
      * @throws \Cake\Database\Exception when table cannot be described.
      */
-    public function describe(string $name, array $options = []): TableSchemaInterface
+    public function describe(string $name, array $options = []): TableSchema
```
ADmad: We are going in the opposite direction by replacing the interface with concrete class. I would keep the interfacce typehint and add inline annotation if possible or simply ignore the error in phpstan's config file.

src/Http/ActionDispatcher.php
```diff
@@ -47,7 +47,7 @@ public function __construct(?ControllerFactory $factory = null)
      * Dispatches a Request & Response
      *
      * @param \Cake\Http\ServerRequest $request The request to dispatch.
-     * @param \Psr\Http\Message\ResponseInterface $response The response to dispatch.
+     * @param \Cake\Http\Response $response The response to dispatch.
```
ADmad: I don't think the docblock not matching the actual typehint is a good idea. Inline `@var` can be added inside the method where required to make phpstan happy.

src/Http/BaseApplication.php
```diff
@@ -233,7 +234,7 @@ public function pluginConsole(CommandCollection $commands): CommandCollection
      * - Create the controller that will handle this request.
      * - Invoke the controller.
      *
-     * @param \Psr\Http\Message\ServerRequestInterface $request The request
+     * @param \Cake\Http\ServerRequest $request The request
```
ADmad: Same as above, docblock should not mismatch actual typehint.

src/ORM/EagerLoadable.php
```diff
@@ -167,10 +167,15 @@ public function associations(): array
...
      *
-     * @return \Cake\ORM\Association|null
+     * @return \Cake\ORM\Association
+     * @throws \RuntimeException
      */
-    public function instance(): ?Association
+    public function instance(): Association
     {
+        if ($this->_instance === null) {
+            throw new \RuntimeException('No instance set.');
```
ADmad: A more descriptive message like `Association instance not set` would be better.

src/Database/Schema/SqlGeneratorInterface.php
```diff
@@ -29,44 +29,44 @@ interface SqlGeneratorInterface
      * Uses the connection to access the schema dialect
      * to generate platform specific SQL.
      *
-     * @param \Cake\Database\Connection $connection The connection to generate SQL for.
+     * @param \Cake\Datasource\ConnectionInterface $connection The connection to generate SQL for.
      * @return array List of SQL statements to create the table and the
      *    required indexes.
      */
-    public function createSql(Connection $connection): array;
+    public function createSql(ConnectionInterface $connection): array;
```
ADmad: Why is this necessary? `ConnectionInterface` is too generic here. The `Database\Connection` class is relational db centric and provides the methods and drivers which use SQL. Since this interface is also specific to generating SQL for relational db, typhinting with `Connection` is the right thing to do.

In general all classes under `Database` package should just use `Connection` for typehint as they all relate to relational database and  `ConnectionInterface` doesn't have all the requisite methods.

We could add a new `Database\ConnectionInterface` which extends `Datasource\ConnectionInterface` but it wouldn't serve any purpose as no one is going to write an alternate implementation for `Database\Connection`.

ADmad: I would in fact remove the `disableConstraints()` method from `ConnectionInterface` as it would only be applicable for relational databases.

ADmad: As seen in the connection class of elastic search plugin, the method had to be made into a no-op 

https://github.com/cakephp/elastic-search/blob/c1394953fd09f38d157092f86760fac653c31b41/src/Datasource/Connection.php#L167-L173

ADmad: `ConnectionInterface::setLogger()` and `ConnectionInterface::getLogger()` use `\Cake\Database\Log\QueryLogger` for typehinting which too is technically incorrect since the `database` package is not a dependency of `datasource` packages. It's the other way round.

So `Datasource\ConnectionInterface` serves no real purpose. We would be actually lying by using it for typehints. We should directly use the `Database\Connection`. I would suggest simply dropping `ConnectionInterface`.

The `ConnectionManager` and `ConnectionRegistry` should be able to take any object as connection instance, which they already do. For e.g. it these classes load instances of `Muffin\Webservice\Connection` which doesn't even implement `ConnectionInterface` and that behavior should be preserved.

ADmad:
I pushed few commits improving the typehints while keeping phpstan and psalm happy.

ADmad:
I am going to try revert the `TableSchema` vs `TableSchemaInterface` changes next.

dereuromark:
Nice work Completing it!

PR #13260, opened 2019-05-19
Use TableSchemaInterface for typehints instead of TableSchema.

ADmad:
It's only `TestFixture` which requires the methods provided by `SqlGeneratorInterface` and implemented in `TableSchema`.

Refs #13254


ADmad:
Merging as the changes (or the lack of) can still be reviewed in #13254.

Issue #13256, opened 2019-05-18
`$this->{'NotLoadedModel'}` silently reverts to `$this`


mehov:
This is a (multiple allowed):

* [x] bug
* [x] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.7.2

### What you did

```
// in Model/Table/SomeTable.php
function getDataFromAnotherTable($id) {
    $tableLocator = \Cake\ORM\TableRegistry::getTableLocator();
    $anotherTable = $tableLocator->get('AnotherTable');
    // the following was copy-pasted from elsewhere
    $this->{'AnotherTable'}->get($id); 
}
```

### What happened

> Exception: Table "App\Model\Table\SomeTable" is not associated with "AnotherTable"

Apparently, after failing to locate `$this->{'AnotherTable'}`, it reverted to just `$this`, which happened to be the initial `SomeTable`.

### What you expected to happen

> Exception: Trying to access "AnotherTable" which is not loaded

I understand this was my fault, but still if the system can detect that, making the reason obvious would help to detect the problem much faster

ADmad:
It didn't "revert" to `$this`. You are already in context of SomeTable (referenced by $this) trying to access it's non existent property.

Perhaps changing the order of the names mentioned in the error message will avoid confusion: `'AnotherTable is not associated with 'SomeTable'`.

markstory:
The exception you're seeing happens when `Table::__get()` attempts to resolve `$this->{'AnotherTable'}` which fails as the association doesn't exist.

There is no fallback to $this happening. Would the error message be clearer if it said

> You've attempted to access `$this->AnotherTable` from `App\Model\Table\SomeTable` but that association is not defined.

mehov:
I got confused because I tried to access `AnotherTable` (as a property of `SomeTable`), but instead I got the base `SomeTable` itself – this is why I used the word "revert" (although @ADmad I don't insist on it). The object I got looked and acted like a table, and it took me some non-zero time to figure out it's just a wrong one.

I'm not sure why is the framework doing that: is it due to the architecture, or it's just trying to "help", but it feels misleading. For me, this good old PHP notice would seem more precise and straightforward:

> Notice (8): Undefined property: SomeTable::$AnotherTable

However, the message wording is completely up to you. Moreover, I understand there must be reasons for such implementation, so I'm not trying to argue. I was just trying to see if there's a way to somehow avoid any future confusion.

ADmad:
[...]

PR #13263, opened 2019-05-20
Improve error message for missing associations.

markstory:
Try to make the missing association error message clearer. Indicate that the property is undefined and that it was interpreted as an association.

Refs #13256

### DIFF
--- a/src/ORM/Table.php
+++ b/src/ORM/Table.php
@@ -2532,10 +2532,10 @@ class Table implements RepositoryInterface, EventListenerInterface, EventDispatc
         if (!$association) {
             throw new RuntimeException(sprintf(
                 'Undefined property `%s`. ' .
-                'You have tried to use the `%s` property on %s, but that association does not exist.',
+                'You have not defined the `%s` association on `%s`.',
                 $property,
                 $property,
-                get_class($this),
+                static::class
             ));
         }
 
--- a/src/ORM/Table.php
+++ b/src/ORM/Table.php
@@ -2531,9 +2531,11 @@ class Table implements RepositoryInterface, EventListenerInterface, EventDispatc
         $association = $this->_associations->get($property);
         if (!$association) {
             throw new RuntimeException(sprintf(
-                'Table "%s" is not associated with "%s"',
+                'Undefined property `%s`. ' .
+                'You have tried to use the `%s` property on %s, but that association does not exist.',
+                $property,
+                $property,
                 get_class($this),
-                $property
             ));
         }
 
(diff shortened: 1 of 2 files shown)


src/ORM/Table.php
```diff
@@ -2531,9 +2531,11 @@ public function __get($property)
         $association = $this->_associations->get($property);
         if (!$association) {
             throw new RuntimeException(sprintf(
-                'Table "%s" is not associated with "%s"',
+                'Undefined property `%s`. ' .
+                'You have tried to use the `%s` property on %s, but that association does not exist.',
```
othercorey: If you don't mind an outside comment, I would keep this simple for the user.  "You did not set an association between X and Y.  I think talking about properties is less clear since this is strictly dealing with associations.

othercorey:
That feels better to me.

src/ORM/Table.php
```diff
@@ -2531,9 +2531,11 @@ public function __get($property)
         $association = $this->_associations->get($property);
         if (!$association) {
             throw new RuntimeException(sprintf(
-                'Table "%s" is not associated with "%s"',
-                get_class($this),
-                $property
+                'Undefined property `%s`. ' .
+                'You have not defined an association between `%s` and `%s` tables.',
```
saeideng: this is not clear now 

I think this should tell to user that which is base table  name and which target table

markstory: How about now?

ADmad:
@markstory Seems you mistakenly ran the 4.x CS fixer :)

markstory:
@ADmad Sounds like something I would do as I have my editor setup to auto-run phpcbf now.

src/ORM/Table.php
othercorey: I think the information is provided by the exception itself:

$this->Articles->Tags
You have not set the association between Articles and Tags.

Working with a team new to cakephp, I think the people confused at this point are always going to be confused until they spend a little more time with cakephp. Associations in general are confusing!  Most of the emails I received early on was about queries not working because I forgot to set the right binding key on associations/etc and the SQL error makes no sense to them.

saeideng: @markstory or sync with `getAssociation()`
https://github.com/cakephp/cakephp/blob/32d9d38cca72e1795b940a91e1154c8253dc7e56/src/ORM/Table.php#L1007

markstory: So the two error messages need to be the same? Could that message be updated separately?

markstory: So the two error messages need to be the same? Could that message be updated separately?

markstory: So the two error messages need to be the same? Could that message be updated separately?

saeideng: >So the two error messages need to be the same?

no,  just a suggestion 
now is good :+1:

PR #13264, opened 2019-05-20
Fix nested plugin enumeration corrupting iterator state.

### LINKED ISSUE #13246: 4.x: loading plugins defect
This is a (multiple allowed):

* [x] bug
* [ ] enhancement

* CakePHP Version: 4.x

### What you did
loading plugins, and trying bin/cake

### What happened
As soon as you load Bake plugin, all other plugins are not loaded properly anymore, the console() command for their CLI command discovery does not get invoked anymore, not showing those in the CLI overview of bin/cake.

Bake loading other plugins internally should not mess with the general loading of plugins.

markstory:
Because PluginCollection is an Iterator implementation it maintains the current offset which can be corrupted when the iterator is iterated in two nested loops. Using the inner array solves this problem.

This scenario commonly happens when a plugin uses the PluginCollection to discover other objects. While this error has only really manifested in 4.x, the issue also exists in 3.7.

Fixes #13246

### DIFF
--- a/src/Core/PluginCollection.php
+++ b/src/Core/PluginCollection.php
@@ -273,7 +273,7 @@ class PluginCollection implements Iterator, Countable
         if (!in_array($hook, PluginInterface::VALID_HOOKS)) {
             throw new InvalidArgumentException("The `{$hook}` hook is not a known plugin hook.");
         }
-        foreach ($this as $plugin) {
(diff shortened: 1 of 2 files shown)


ADmad:
Test failing.

ADmad (approved):

PR #13265, opened 2019-05-20
4.x post merge type fixes

ADmad:
Fixed typeing errors reported by psalm after `3.next` to `4.x` merge.

### DIFF
--- a/src/TestSuite/Constraint/Console/ExitCode.php
+++ b/src/TestSuite/Constraint/Console/ExitCode.php
@@ -25,7 +25,7 @@ use PHPUnit\Framework\Constraint\Constraint;
 class ExitCode extends Constraint
 {
     /**
-     * @var int
+     * @var int|null
      */
     private $exitCode;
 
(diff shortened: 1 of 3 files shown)


src/TestSuite/Constraint/Console/ExitCode.php
```diff
@@ -57,6 +57,6 @@ public function matches($other): bool
      */
     public function toString(): string
     {
-        return sprintf('matches exit code %d', $this->exitCode);
+        return sprintf('matches exit code %s', $this->exitCode ?? 'null');
```
dereuromark: Why is this needed if you  have exist code defined and either null or int?
$this->exitCode should suffice then?

ADmad: Nope, cause I want string "null" to be used if exit code is `null`.

dereuromark: I see. I didnt see the quotes around it ;) Early for me.

PR #13266, opened 2019-05-20
More phpstan/psalm fixes

dereuromark:
Most importantly, the helper wasnt in sync with the lib.
But also the |bool are a lie as only dirty false is also returned.
If you otherwise check for false return via `if ()` the tools will always complain that true wasnt handled, but it shouldnt be as true is not a valid return value.

### DIFF
--- a/src/View/Form/EntityContext.php
+++ b/src/View/Form/EntityContext.php
@@ -19,10 +19,10 @@ namespace Cake\View\Form;
 use ArrayAccess;
 use Cake\Collection\Collection;
 use Cake\Datasource\EntityInterface;
-use Cake\Datasource\RepositoryInterface;
 use Cake\Http\ServerRequest;
 use Cake\ORM\Entity;
 use Cake\ORM\Locator\LocatorAwareTrait;
+use Cake\ORM\Table;
 use Cake\Utility\Inflector;
 use Cake\Validation\Validator;
 use RuntimeException;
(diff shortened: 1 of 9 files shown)


dereuromark:
I cant resolve the

    Call to an undefined method Cake\ORM\Association::junction().

error, even though I use inline annotations
Do we need to exclude this one in ignoreErrors?

src/View/Form/EntityContext.php
```diff
@@ -534,10 +547,10 @@ protected function _getValidator(array $parts): Validator
     /**
      * Get the table instance from a property path
      *
-     * @param array|string|\Cake\Datasource\EntityInterface $parts Each one of the parts in a path for a field name
+     * @param string[]|string|\Cake\Datasource\EntityInterface $parts Each one of the parts in a path for a field name
      * @param bool $fallback Whether or not to fallback to the last found table
      *  when a non-existent field/property is being encountered.
-     * @return \Cake\ORM\Table|bool Table instance or false
+     * @return \Cake\ORM\Table|false Table instance or false
```
ADmad: How about making it return `null` instead of `false` if no matching table is found?

dereuromark: I agree with you. But can we still do that here in 4.x?

ADmad: Sure can.

ADmad (approved):

PR #13271, opened 2019-05-21
Cleanup ignore list. Add missing inline annotation for IDE.

dereuromark:
The alternative would be to throw an exception if false, but given that this would never really happen in these cases, the ternary seems to be the easier way to keep return type clean.


src/View/View.php
```diff
@@ -796,7 +796,7 @@ public function set($name, $value = null)
     {
         if (is_array($name)) {
             if (is_array($value)) {
-                $data = array_combine($name, $value);
+                $data = array_combine($name, $value) ?: [];
```
ADmad: We probably should throw an exception here since it's a public method and user could mistakenly provide array with different number of values in each array.

PR #13276, opened 2019-05-21
Make ConnectionInterface use LoggerInterface instance.

ADmad:
Update the internal class QueryLogger to implement LoggerInterface.
Query interpolation code has been moved to LoggedQuery itself.

This would allow people to use any PSR-3 complaint logger for the database package and remove hard dependency on `cakephp/log`.

I can finish up and update tests if the change seems reasonable.

### DIFF
--- a/src/Database/Connection.php
+++ b/src/Database/Connection.php
@@ -32,6 +32,7 @@ use Cake\Database\Schema\CollectionInterface as SchemaCollectionInterface;
 use Cake\Datasource\ConnectionInterface;
 use Cake\Log\Log;
 use Exception;
+use Psr\Log\LoggerInterface;
 
 /**
  * Represents a connection with a database server.
@@ -87,7 +88,7 @@ class Connection implements ConnectionInterface
     /**
      * Logger object instance.
      *
-     * @var \Cake\Database\Log\QueryLogger|null
+     * @var \Psr\Log\LoggerInterface|null
      */
     protected $_logger;
 
@@ -828,10 +829,10 @@ class Connection implements ConnectionInterface
     /**
      * Sets a logger
      *
-     * @param \Cake\Database\Log\QueryLogger|null $logger Logger object
+     * @param \Psr\Log\LoggerInterface $logger Logger object
      * @return $this
      */
-    public function setLogger(?QueryLogger $logger)
+    public function setLogger(LoggerInterface $logger)
     {
         $this->_logger = $logger;
 
@@ -841,9 +842,9 @@ class Connection implements ConnectionInterface
     /**
      * Gets the logger object
      *
-     * @return \Cake\Database\Log\QueryLogger logger instance
+     * @return \Psr\Log\LoggerInterface logger instance
      */
-    public function getLogger(): QueryLogger
+    public function getLogger(): LoggerInterface
     {
         if ($this->_logger === null) {
             $this->_logger = new QueryLogger();
@@ -862,7 +863,7 @@ class Connection implements ConnectionInterface
     {
         $query = new LoggedQuery();
         $query->query = $sql;
-        $this->getLogger()->log($query);
+        $this->getLogger()->debug((string)$query, ['query' => $query]);
     }
 
     /**
--- a/src/Database/Connection.php
+++ b/src/Database/Connection.php
@@ -831,6 +831,7 @@ class Connection implements ConnectionInterface
      *
      * @param \Psr\Log\LoggerInterface $logger Logger object
      * @return $this
+     * @psalm-suppress ImplementedReturnTypeMismatch
      */
     public function setLogger(LoggerInterface $logger)
     {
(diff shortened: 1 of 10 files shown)


lorenzo:
Fantastic!

src/Database/Connection.php
```diff
@@ -841,9 +842,9 @@ public function setLogger(?QueryLogger $logger)
...
      * Gets the logger object
      *
-     * @return \Cake\Database\Log\QueryLogger logger instance
+     * @return \Psr\Log\LoggerInterface logger instance
      */
-    public function getLogger(): QueryLogger
+    public function getLogger(): LoggerInterface
     {
         if ($this->_logger === null) {
             $this->_logger = new QueryLogger();
```
lorenzo: This is now a potentially breaking change for us eres of the database package, since QueryLogger extends something that may not be there

src/Database/Log/LoggedQuery.php
```diff
@@ -59,13 +59,58 @@ class LoggedQuery
...
+        $keys = [];
+        $limit = is_int(key($params)) ? 1 : -1;
+        foreach ($params as $key => $param) {
+            $keys[] = is_string($key) ? "/:$key\b/" : '/[?]/';
+        }
+
+        return preg_replace($keys, $params, $this->query, $limit);
+    }
+
     /**
```
lorenzo: Would you mind adding a Json serialize method?

src/Database/Connection.php
ADmad: If we remove `cakephp/log` as hard dependency of `cakephp/database` then it would be up to the user to either add `cakephp/log` as dependency for their app or use package of their choice and use `setLogger()` to explicitly set a logger instance after creating connection instance.

`getLogger()` can do a check whether a logger is already set and if not first ensure `BaseLog` class is available before trying to create `QueryLogger` instance else throw exception.

src/Database/Log/LoggedQuery.php
ADmad: Sure

src/Database/Connection.php
ADmad: Or if we don't mind having extra files pulled in even if user doesn't want to user our logger, then we keep `cakephp/log` as dependency thereby achieving backward compatibility as nothing changes for end user.

src/Database/Connection.php
```diff
@@ -828,10 +829,10 @@ public function isQueryLoggingEnabled(): bool
     /**
      * Sets a logger
      *
-     * @param \Cake\Database\Log\QueryLogger|null $logger Logger object
+     * @param \Psr\Log\LoggerInterface $logger Logger object
      * @return $this
      */
-    public function setLogger(?QueryLogger $logger)
+    public function setLogger(LoggerInterface $logger)
```
markstory: How does someone remove the logger now?

src/Database/Connection.php
markstory: We could also do a class check on `Cake\Log\Log` and fallback to `error_log()` which dumps to the SAPI error output, which is stderr in CLI and webserver error logs in apache/fpm.

I'm also ok with adding cake/log as an `suggest` dependency. We could throw an error here if `Cake\Log\Log` doesn't exist.

src/Database/Connection.php
ADmad: In what scenario would removing a logger be required?

One should either not set a logger at all if he doesn't want logging or to prevent logging temporarily just disable logging.

ADmad: `setLogger()` internally creates of `QueryLogger` if `Connection::$_log` property is `null`. So using this method to unset the logger doesn't actually achieve anything currently either.

lorenzo: could we also provide a `Null` logger?

ADmad: `psr/log` already provides a `NullLogger` :slightly_smiling_face:

ADmad: Making the argument non-nullable also allows to make the class implement `LoggerAwareInterface`.

lorenzo: The we just need to add to the docs how using a null logger is possible. Even if it is obvious and redundant, I did not remember that `psr/log` had it.

ADmad:
Any objection to making `ConnectionInterface` extend `LoggerAwareInterface` since it already has `setLogger()` method with the latter has?

`LoggingStatement` also has `setLogger()` but since it's marked as `@internal` I don't see much value in making it explicitly implement `LoggerAwareInterface`.

src/Database/Connection.php
markstory: Passing a null logger works for me.

lorenzo:
> Any objection to making `ConnectionInterface` extend `LoggerAwareInterface` since it already has `setLogger()` method with the latter has?

No objections from me

markstory:
This seems like the kind of change that should be included in the migration guide. I can take care of that tomorrow.

markstory (approved):

ADmad:
@markstory Thanks for taking care of docs update.

BTW `LoggerAwareInterface::setLogger()` is supposed to return void but `Connection::setLogger()` returns `$this` which I haven't changed. Is that okay?

PR #13277, opened 2019-05-21
Redesign Development Error Pages

cleptric:
I added a bit more space here and there and changed the colours. ~Also, both columns scroll now independently, to provider better context.~

**Before**
Screenshot 2019-05-21 at 18 49 58

**Now**
Screenshot 2019-05-21 at 18 50 04

### DIFF
--- a/templates/layout/dev_error.php
+++ b/templates/layout/dev_error.php
@@ -23,79 +23,109 @@
     </title>
     <?= $this->Html->meta('icon') ?>
     <style>
+    * {
+        box-sizing: border-box;
+    }
     body {
-        font: 14px helvetica, arial, sans-serif;
-        color: #222;
-        background-color: #f8f8f8;
-        padding:0;
+        font-family: 'Helvetica Neue', 'Helvetica', 'Arial', sans-serif;
+        color: #404041;
+        background: #F5F7FA;
+        font-size: 14px;
+        letter-spacing: .01em;
+        line-height: 1.6;
+        padding: 0;
         margin: 0;
-        max-height: 100%;
+        height: 100%;
     }
-
-    .code-dump,
-    pre {
-        background: #fefefe;
-        border: 1px solid #ddd;
-        padding: 5px;
-        white-space: pre-wrap;
-    }
-
     header {
-        background-color: #C3232D;
+        background-color: #D33C47;
         color: #ffffff;
-        padding: 16px 10px;
-        border-bottom: 3px solid #626262;
+        padding: 10px;
     }
     .header-title {
-        margin: 0;
-        font-weight: normal;
         font-size: 30px;
-        line-height: 64px;
+        margin: 0;
     }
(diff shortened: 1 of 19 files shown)


ADmad:
I prefer the horizontal separators in left panel (they could be made thinner) and the colors of the error message bar which are present currently :slightly_smiling_face:.

ADmad:
I am fine not having the horizontal separator in left panel but would prefer if the error message and suggestion bar use different primary/base colors and both not be shades of blue.

josegonzalez:
How does this compare to Rails error messages or Whoops? What can we steal from them?

cleptric:
@ADmad I'll provide some different colours to pick from :)
@josegonzalez Whoops looks like this: https://camo.githubusercontent.com/31a4e1410e740fd0ccda128cbcab8723f45e7e73/687474703a2f2f692e696d6775722e636f6d2f305651706539362e706e67

What would be neat, is to have a bit of JS to copy the suggested code to the clip board.
And maybe a refresh of the passed arguments.

cleptric:
Screenshot 2019-05-21 at 21 24 01
@ADmad

markstory:
> I prefer the horizontal separators in left panel (they could be made lighter) and the colors of the error message, suggestion bars which are present currently

Whitespace/margins are a separator :smile:

inoas:
> Screenshot 2019-05-21 at 21 24 01

Very nice aside one thing, the vertical separation between line references in the stack trace is too big...it takes a lot of screen real estate and you can have dozens of lines there.

inoas:
I'd also prefer if it said # LINE / FILE or have a tabular layout aka

ISSUE (colspan 2)
FILE | LINE

inoas:
For the stack I'd also rather see very slight and thin lines than none to be able to remove most of the vertical spacing without adding any kind of ambiguity.

markstory:
@inoas Those are all great suggestions, perhaps we can get them in as a follow up change?

templates/layout/dev_error.php
```diff
@@ -202,96 +224,92 @@
...
+            <div class="error-suggestion">
+                <?= $this->fetch('file') ?>
+            </div>
 
-        <div class="error-suggestion">
-            <?= $this->fetch('file') ?>
+            <?php if ($this->fetch('templateName')): ?>
+            <p class="customize">
+                If you want to customize this error message, create
+                <em><?= APP_DIR . DIRECTORY_SEPARATOR . 'Template' . DIRECTORY_SEPARATOR . 'Error' . DIRECTORY_SEPARATOR . $this->fetch('templateName') ?></em>
```
markstory: This path looks like the 3.x path and not the 4.x one which is

```php
<em><?= 'templates' . DIRECTORY_SEPARATOR . 'Error' . DIRECTORY_SEPARATOR . $this->fetch('templateName') ?></em>
```

PR #13278, opened 2019-05-21
Fix also ServerRequest referer to default to local.

dereuromark:
Follow https://github.com/cakephp/cakephp/pull/12320

After Controller::referer() now defaults to own/local URLs, the ServerRequest one doesnt.
This would consolidate this to a common value.


markstory:
This is something we would need to document in the migration guide.

dereuromark:
Added in https://github.com/cakephp/docs/pull/6086

PR #13279, opened 2019-05-22
Psalm fixes.

dereuromark:
Removing 10-15 lines of the basefile seems like a small win compared to its 2700 lines.
But here we go.

Given that we enabled strict mode, we need to fix more possible type issues, as they will blow up in userland code sooner than later if we release 4.0.


dereuromark:
CS errors will be gone via https://github.com/cakephp/cakephp-codesniffer/pull/251 merged.
non void is OK if we abort early via exception.

PR #13281, opened 2019-05-22
Move action's return value check to Controller::invokeAction().

ADmad:
This allows adding return typehint to `invokeAction()` and make `ActionDispatcher::_invoke()` cleaner.

### DIFF
--- a/src/Controller/Controller.php
+++ b/src/Controller/Controller.php
@@ -517,9 +517,11 @@ class Controller implements EventListenerInterface, EventDispatcherInterface
         }
 
         if (!$result instanceof ResponseInterface) {
-            throw new UnexpectedValueException(
-                'Controller actions can only return ResponseInterface instance or null.'
-            );
+            throw new UnexpectedValueException(sprintf(
+                'Controller actions can only return ResponseInterface instance or null. '
+                . 'Got %s instead.',
+                getTypeName($result)
+            ));
         }
 
         return $this->response = $result;
--- a/src/Controller/Controller.php
+++ b/src/Controller/Controller.php
@@ -29,10 +29,12 @@ use Cake\Log\LogTrait;
 use Cake\ORM\Locator\LocatorAwareTrait;
 use Cake\Routing\Router;
 use Cake\View\ViewVarsTrait;
+use Psr\Http\Message\ResponseInterface;
 use ReflectionClass;
 use ReflectionException;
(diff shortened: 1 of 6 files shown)


src/Http/ActionDispatcher.php
```diff
@@ -79,22 +77,15 @@ protected function _invoke(Controller $controller): ResponseInterface
...
-        if (!$response && $controller->isAutoRenderEnabled()) {
+        if ($response === null && $controller->isAutoRenderEnabled()) {
             $controller->render();
         }
 
         $result = $controller->shutdownProcess();
         if ($result instanceof ResponseInterface) {
             return $result;
         }
-        if (!$response) {
```
dereuromark: We dont need this part anymore? We now always fetch from $controller->getResponse()?

ADmad: We don't, since return response returned by action is also assigned to `Controller::$response`. Plus beforeRender() / afterFilter() could have modified `Controller::$response` so that what holds the reference to most up to date response instance.

src/Controller/Controller.php
```diff
@@ -504,15 +507,22 @@ public function invokeAction()
...
         $result = $callable(...array_values($request->getParam('pass')));
-        if ($result instanceof Response) {
-            $this->response = $result;
+        if ($result === null) {
+            return $result;
+        }
+
+        if (!$result instanceof ResponseInterface) {
+            throw new UnexpectedValueException(
+                'Controller actions can only return ResponseInterface instance or null.'
```
markstory: Should we tell the user the type of thing we got that was unexpected?

tests/TestCase/Controller/ControllerTest.php
```diff
@@ -756,6 +756,30 @@ public function testInvokeActionWithPassedParams(): void
...
+            'params' => [
+                'controller' => 'Test',
+                'action' => 'willCauseException',
+                'pass' => [],
+            ],
+        ]);
+        $response = $this->getMockBuilder(Response::class)->getMock();
+
+        $Controller = new TestController($url, $response);
+        $result = $Controller->invokeAction();
```
markstory:
```suggestion
        $Controller->invokeAction();
```

src/Controller/Controller.php
ADmad: Done

PR #13282, opened 2019-05-22
Always ensure FQCN returned by App::className() actually exits.

ADmad:
Earlier `class_exists()` check was done only when interpolating short name to FQCN
but not when argument passed to the method is already a FQCN.

### DIFF
--- a/src/Console/Shell.php
+++ b/src/Console/Shell.php
@@ -305,7 +305,7 @@ class Shell
     {
         foreach ($this->_taskMap as $taskName => $task) {
             $class = App::className($task['class'], 'Shell/Task', 'Task');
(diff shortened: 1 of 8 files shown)


lorenzo:
are you sure this is not a breaking change? I think this method is used by plugins for example to only write strings and not necessarily because you want to instant oats the class.

ADmad:
@lorenzo I don't think so. AFAIK this method is always used by registries/loaders to get FQCN right before trying to instantiate an object. So ensuring the class exists shouldn't be a problem.

In fact the previous inconsistency of skipping class exists check when the string looked like a FQCN (when it contained a backslash) but doing the check only when interpolating the short name to FQCN made no sense.

Issue #13272, opened 2019-05-21
Use view builder options instead of special view vars like "_serialize"


ADmad:
Currently we use special view vars like `_serialize`, `_jsonOptions`, `_xmlOptions` etc. for various options for `JsonView` and `XmlView`.

Instead of abusing data vars for configuring the serialized view classes, we can use `ViewBuilder::options()` which allows passing custom options to them.

I can whip a PR if there's an agreement.

dereuromark:
Isn't the main benefit of those options, similar to Routes, that you can store them as array, or even pass them down more easily like this from different APIs?
When making them pure class options this will not work this way anymore. But I dont think this is a strong case against it.

markstory:
What would the upgrade path be for folks using the current system?

ADmad:
We can check for the new view options and if unset check for the corresponding view vars currently used. In debug mode throw a warning to use new option instead. Drop the check for old view vars in 4.1.

markstory:
That could work well.

PR #13285, opened 2019-05-23
4.x view custom options

ADmad:
Refs #13272

### DIFF
--- a/src/View/JsonView.php
+++ b/src/View/JsonView.php
@@ -81,23 +81,34 @@ class JsonView extends SerializedView
     /**
      * List of special view vars.
      *
+     * Use ViewBuilder::setOption()/setOptions() to set these vars.
+     *
      * @var array
      */
-    protected $_specialVars = ['_serialize', '_jsonOptions', '_jsonp'];
+    protected $_specialVars = ['serialize', 'jsonOptions', 'jsonp'];
+
+    /**
+     * Options for json_encode().
+     *
+     * e.g. `JSON_HEX_TAG | JSON_HEX_APOS`.
+     *
+     * @var int|null
+     */
+    protected $jsonOptions;
+
+    /**
+     * Enables JSONP support and wraps response in callback function provided in query string.
+     *
+     * - Setting it to true enables the default query string parameter "callback".
+     * - Setting it to a string value, uses the provided query string parameter
+     *     for finding the JSONP callback name.
+     *
+     * @var true|string|null
+     */
+    protected $jsonp;
 
     /**
      * Render a JSON view.
-     *
-     * ### Special parameters
-     * `_serialize` To convert a set of view variables into a JSON response.
-     *   Its value can be a string for single variable name or array for multiple
-     *   names. If true all view variables will be serialized. If unset normal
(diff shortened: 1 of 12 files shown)


src/View/SerializedView.php
```diff
@@ -107,4 +135,30 @@ public function render(?string $template = null, $layout = null): string
...
+        if (isset($this->{$name})) {
+            return $this->{$name};
+        }
+
+        if (isset($this->viewVars["_{$name}"])) {
+            deprecationWarning(sprintf(
+                'Setting special view var "_%s" is deprecated. Use ViewBuilder::setOption(\'%s\', $value) instead.',
+                $name,
+                $name
+            ));
```
markstory: I like this :+1:

src/View/XmlView.php
```diff
@@ -80,33 +80,44 @@ class XmlView extends SerializedView
...
+
     /**
      * List of special view vars.
      *
+     * Use ViewBuilder::setOption()/setOptions() to set these vars.
+     *
      * @var array
      */
-    protected $_specialVars = ['_serialize', '_rootNode', '_xmlOptions'];
+    protected $_specialVars = ['serialize', 'rootNode', 'xmlOptions'];
```
markstory: Is specialVars the best name for this? Would `context` or `options` be better?

ADmad:
@markstory I removed the `$_specialVars` property and instead added `$options` for `View` itself which holds hash of custom view options. Having separate property for each custom option wasn't necessary. The `$options` hash will allow managing view option for user defined classes too.

But after implementing this I realized I partly re-implemented what `InstanceConfigTrait` provides :grimacing:. So should I make `View` just use `InstanceConfigTrait` instead?

src/View/JsonView.php
```diff
@@ -79,33 +79,27 @@ class JsonView extends SerializedView
...
+     *   names. If true all view variables will be serialized. If null or false
+     *   normal view template will be rendered.
+     * - `jsonOptions`: Options for json_encode(). For e.g. `JSON_HEX_TAG | JSON_HEX_APOS`.
+     * - `jsonp`: Enables JSONP support and wraps response in callback function provided in query string.
+     *   - Setting it to true enables the default query string parameter "callback".
+     *   - Setting it to a string value, uses the provided query string parameter
      *     for finding the JSONP callback name.
      *
-     * @var true|string|null
+     * @var array{serialize:string|bool|null, jsonOptions: int|null, jsonp: bool|string|null}
```
markstory: Is this something new in psalm?

ADmad: Not new, just something we haven't used before. We should try to use it wherever dealing with arrays with fixed keys. https://psalm.dev/docs/docblock_type_syntax/#object-like-arrays

ADmad:
I made an alternate branch making `View` use `InstanceConfigTrait`:
https://github.com/cakephp/cakephp/compare/4.x...ADmad:4.x-view-config

markstory:
I like the InstanceConfig approach as it reuses concepts we use everywhere else in the framework.

ADmad:
Closing in favor of #13301

PR #13287, opened 2019-05-23
4.x connection cache

ADmad:
Similar to #13276, this change allows connection to now use injectable simple cache instance.


src/Database/Connection.php
```diff
@@ -790,6 +799,33 @@ public function cacheMetadata($cache): void
...
+     */
+    public function getCacher(): CacheInterface
+    {
+        if ($this->cacher === null) {
+            $configName = $this->_config['cacheMetadata'] ?? '_cake_model_';
+            if (!is_string($configName)) {
+                $configName = '_cake_model_';
+            }
+
+            $this->cacher = Cache::pool($configName);
```
lorenzo: I think that   we should somehow not force this dependency, and maybe provide a null cacher implementation instead of defaulting in this way.

ADmad: If we remove this then we will need to come up with another way to set default cacher for cake apps.

Currently we just call `ConnectionManager` in bootstrap to setup db config arrays and the rest happens "magically".

ADmad: I can make `getCacher()` return null if no cacher is explicitly set and move the fetching to default cache using `Cache` back into `CachedCollection`.

markstory: Would it be better if we raised an error if `cakephp/cache` isn't available?

ADmad: Done. I also removed `cakehp/cache` as dependency.

ADmad:
The tests seem to be failing because same connection instance is used across tests and it's state is not reset.

I cloned the connection instance in few tests and locally I don't get any errors when running phpunit for the `TestCase/Database` folder or `SchemaCacheShellTest.php` separately. But still can't figure out what's causing failure on travis.

src/Database/composer.json
```diff
@@ -25,7 +25,6 @@
     },
     "require": {
         "php": ">=7.2.0",
-        "cakephp/cache": "^4.0",
```
markstory: Should we add the psr/simple-cache package instead?

tests/TestCase/Shell/SchemaCacheShellTest.php
```diff
@@ -65,8 +67,23 @@ public function tearDown(): void
         parent::tearDown();
         Cache::drop('orm_cache');
 
-        $ds = ConnectionManager::get('test');
-        $ds->cacheMetadata(false);
+        unset($this->connection);
+    }
+
+    protected function getShell()
```
markstory: You and I will conflict on this.

markstory (approved):
Looks good but we need to update the upgrade guide as there are a few breaking changes here. I don't think they will have significant impacts in applications but still good to document the changes.

tests/TestCase/Shell/SchemaCacheShellTest.php
ADmad: Yeah :disappointed:

ADmad:
I fixed the issue that was causing build failures earlier.

But now the testsuite crashes for MySql and Postgres with duplicate foreign key contraint error. It's the same error that has been causing Appveyor/MSSQL build to fail since a long time. So surely not caused by my changes.

PR #13290, opened 2019-05-24
4.x Add scoped middleware to the app queue

cleptric:
Applying a scoped middleware inside a route scope skips all remaining middleware from being invoked, that are applied after the `RoutingMiddleware`.

The `RoutingMiddleware::process`  method creates a new `MiddlewareQueue` and `Runner` instance, if the current `RouteCollection` contains applicable middleware.
Doing so results in the second `Runner` instance being created that drains a second `MiddlewareQueue` and calling `BaseApplication::handle`, which invokes the Cake application.

All remaining items in the first  `MiddlewareQueue` are ignored and not invoked.

I tested it in a Cake 4.x app, which seems to work fine.

### DIFF
--- a/src/Routing/Middleware/RoutingMiddleware.php
+++ b/src/Routing/Middleware/RoutingMiddleware.php
@@ -158,6 +158,6 @@ class RoutingMiddleware implements MiddlewareInterface
         $middleware = new MiddlewareQueue($matching);
         $runner = new Runner();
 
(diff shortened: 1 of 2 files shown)


markstory:
Makes sense to me. The `$handler` already has the app at its center.

ADmad (approved):

PR #13294, opened 2019-05-27
Remove cakephp/log as dependency of database package.

ADmad:
Throw execption if logger is not explicitly set and cakephp/log package is not loaded.

Refs #13276

### DIFF
--- a/src/Database/Connection.php
+++ b/src/Database/Connection.php
@@ -854,7 +854,7 @@ class Connection implements ConnectionInterface
 
         if (!class_exists(QueryLogger::class)) {
             throw new RuntimeException(
-                'For logginer you must either set logger using Connection::setLogger()' .
+                'For logging you must either set logger using Connection::setLogger()' .
                 ' or require the cakephp/log package in your composer config.'
             );
         }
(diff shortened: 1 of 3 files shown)


src/Database/Connection.php
```diff
@@ -847,11 +848,18 @@ public function setLogger(LoggerInterface $logger)
...
-        if ($this->_logger === null) {
-            $this->_logger = new QueryLogger();
+        if ($this->_logger !== null) {
+            return $this->_logger;
         }
 
-        return $this->_logger;
+        if (!class_exists(QueryLogger::class)) {
+            throw new RuntimeException(
+                'For logginer you must either set logger using Connection::setLogger()' .
```
garas:
```suggestion
                'For logging you must either set logger using Connection::setLogger()' .
```

src/Database/Connection.php
```diff
@@ -847,11 +848,18 @@ public function setLogger(LoggerInterface $logger)
...
-        if ($this->_logger === null) {
-            $this->_logger = new QueryLogger();
+        if ($this->_logger !== null) {
+            return $this->_logger;
         }
 
-        return $this->_logger;
+        if (!class_exists(QueryLogger::class)) {
+            throw new RuntimeException(
+                'For logging you must either set logger using Connection::setLogger()' .
```
markstory:
```suggestion
                'For logging you must either set a logger using Connection::setLogger()' .
```

PR #13297, opened 2019-05-27
4.x psalm

ADmad:


src/Mailer/Message.php
```diff
@@ -922,11 +922,11 @@ public function getHeaders(array $include = []): array
         $headers['MIME-Version'] = '1.0';
         if ($this->attachments) {
             $headers['Content-Type'] = 'multipart/mixed; boundary="' . (string)$this->boundary . '"';
-        } elseif ($this->emailFormat === static::MESSAGE_BOTH) {
+        } elseif ($this->emailFormat === self::MESSAGE_BOTH) {
```
dereuromark: why is this necssary?

ADmad: This isn't necessary for psalm but the class was using a mix of `static` and `self` so I harmonized all to `self`.

dereuromark: In modern php shouldnt this all be LSB? It is theoretical, but better for extension, right?

ADmad: Theoretically yes but in practice I don't see any use case for overriding these constants. Regardless I'll change them to static instead.

src/Log/Log.php
```diff
@@ -214,7 +214,9 @@ protected static function _loadConfig(): void
      */
     public static function reset(): void
     {
-        static::$_registry = null;
+        if (!empty(static::$_registry)) {
```
dereuromark: can we use strict null checks for object init check?

dereuromark (approved):

src/Log/Log.php
ADmad: If I do strict null check then phpstan complains the the check is redundant since the property's docblock doesn't say it's nullable. If I add `|null` for the property both phpstan and psalm start complaining that usage like `static::$_registry->foo` being potentially null access :slightly_smiling_face:

PR #13298, opened 2019-05-27
Remove invalid code for setting output type of ConsoleOutput.

ADmad:
Constructor of ConsoleOutput already handles setting appropriate default output type.

### DIFF
--- a/src/Log/Engine/ConsoleLog.php
+++ b/src/Log/Engine/ConsoleLog.php
@@ -32,7 +32,7 @@ class ConsoleLog extends BaseLog
         'stream' => 'php://stderr',
         'levels' => null,
         'scopes' => [],
-        'outputAs' => 'see constructor'
+        'outputAs' => null,
     ];
 
     /**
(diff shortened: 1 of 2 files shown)


src/Log/Engine/ConsoleLog.php
```diff
@@ -57,14 +57,6 @@ class ConsoleLog extends BaseLog
      */
     public function __construct(array $config = [])
     {
-        if ((DIRECTORY_SEPARATOR === '\\' && !(bool)env('ANSICON') && env('ConEmuANSI') !== 'ON') ||
-            (function_exists('posix_isatty') && !posix_isatty($this->_output))
```
ADmad: `$this->_output` property is unset at this point so this check makes no sense.

markstory: Does changing the output style to plain exist somewhere else too? Without this we will end up outputting bash escape codes into redirected output handles.

ADmad: Yes, ConsoleOutput's constructor does similar check in a proper manner and switches to plain text output if needed

Issue #13283, opened 2019-05-22
Must set field alias for setDefaultTypes() to hydrate dates


othercorey:
* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.7.5
* Platform and Target: apache 2.4, postgres 9.6

### What you did
I selected a date from a table along with other aggregate fields after grouping results.

```
        $collection = $query
            ->select([
                'service_date'
            ])
            ->setDefaultTypes(['service_date' => 'date'])
```

### What happened
The date field was not hydrated to a FrozenDate object but left as a string.  This only worked if  I set an alias for the field: select([`service_date` => `service_date`]).

### What you expected to happen
I was expecting the fields to be converted to a FrozenDate object because the field was specified in `setDefaultTypes()`.

markstory:
What is the generated query without the alias? You might need to call `setDefaultTypes()` with the `Model.service_date`. Internally the Query object sets both `Model.field` and `Model__field` keys into the type map, and userland usage would have to do the same.

othercorey:
```sql
SELECT WssEntries.service_date AS "WssEntries__service_date" FROM wss_entries WssEntries
```

So, you need to pass the Association.Field name to setDefaultTypes().

markstory:
> ... but maybe documentation could be cleared up.
I'm happy to do that, which documentation were you referencing?

othercorey:
I'm not really sure of the intended use and am not sure what to recommend.  After re-reading the documentation, I'm just resolved to avoid it.

markstory:
If you need you set types for aliases fields you can also use the $types parameter of select() which should handle aliasing from what I remember.

othercorey:
I think it could be helpful to link the setDefaultTypes documentation to the TypeMap setDefaults documentation which does the actual work.

Perhaps a sentence explaining if there is a difference between setting types here vs other $types parameters.

othercorey:
@markstory I'm not sure what $types parameter of select() is

>If you need you set types for aliases fields you can also use the $types parameter of select()

markstory:
The `$type` parameter of select/where is the last argument and lets you specify how a field should be mapped in the current query. You can see an example [here](https://book.cakephp.org/3.0/en/orm/query-builder.html#automatically-creating-in-clauses). That is the interface I would recommend using unless you would need to set types multiple times in a query. Instead of setting types multiple types you can use defaults.

othercorey:
[...]

PR #13299, opened 2019-05-28
Attempt to clarify usage of TypeMap methods

markstory:
Refs cakephp/cakephp#13283

### DIFF
--- a/src/Database/TypeMapTrait.php
+++ b/src/Database/TypeMapTrait.php
@@ -77,10 +77,14 @@ trait TypeMapTrait
     }
 
     /**
-     * Allows setting default types when chaining query.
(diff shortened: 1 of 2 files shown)


src/Database/TypeMapTrait.php
```diff
@@ -77,10 +77,14 @@ public function typeMap($typeMap = null)
     }
 
     /**
-     * Allows setting default types when chaining query.
+     * Overwrite the default type mappings for fields
+     * in the implementing object.
+     *
+     * If you need to se
```
ADmad: Statement seems incomplete

markstory: Whoops

PR #13301, opened 2019-05-29
4.x view config

ADmad:
Refs #13272


ADmad:
Psalm failures are unrelated and fixed in #13302

PR #13311, opened 2019-06-01
[2.x]Fix When PHPUnit 5.2.0+, test ends before calling FixtureManager::shutdown()

tenkoma:
In PHPUnit 5.2.0+, program termination at the end of `TestRunner::doRun()` is the default behavior.

https://github.com/sebastianbergmann/phpunit/blob/5.2.0/src/TextUI/TestRunner.php#L566-L586

`CakeTestRunner` calls `FixtureManager::shutdown()` after calling `parent::doRun()`.
Therefore, in PHPUnit 5.2.0+, the table remains after the test.

https://github.com/cakephp/cakephp/blob/2.x/lib/Cake/TestSuite/CakeTestRunner.php#L76

It causes the programmer not to notice the lack of fixtures when creating test cases.


tenkoma:
@markstory This change is for CakePHP 2.x. Please change the milestone. thanks.

PR #13312, opened 2019-06-02
Updated DateTimeFormat format documentation

othercorey:
I copied the format parameter documentation from the protected variables to the actual public functions used to set them for json and string conversion.  This is what the book mentions so that's what you'd normally look up.

https://book.cakephp.org/3.0/en/core-libraries/time.html#Cake\I18n\Time::setJsonEncodeFormat

I updated the reference to setting default format and locales to use the public setter instead of the direct variables..

### DIFF
--- a/src/I18n/DateFormatTrait.php
+++ b/src/I18n/DateFormatTrait.php
@@ -125,9 +125,7 @@ trait DateFormatTrait
      * $time->i18nFormat(Time::UNIX_TIMESTAMP_FORMAT); // outputs '1398031800'
      * ```
      *
-     * If you wish to control the default format to be used for this method, you can alter
-     * the value of the static `Time::$defaultLocale` variable and set it to one of the
-     * possible formats accepted by this function.
+     * You can control the default format used by setting through `Time::setToStringFormat()`.
      *
      * You can read about the available IntlDateFormatter constants at
      * https://secure.php.net/manual/en/class.intldateformatter.php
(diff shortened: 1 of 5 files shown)


src/I18n/DateFormatTrait.php
```diff
@@ -125,9 +125,7 @@ public function nice($timezone = null, $locale = null)
      * $time->i18nFormat(Time::UNIX_TIMESTAMP_FORMAT); // outputs '1398031800'
      * ```
      *
-     * If you wish to control the default format to be used for this method, you can alter
-     * the value of the static `Time::$defaultLocale` variable and set it to one of the
-     * possible formats accepted by this function.
+     * You can control the default format used by setting through `Time::setToStringFormat()`.
```
markstory:
```suggestion
     * You can control the default format used through `Time::setToStringFormat()`.
```

I think the extra words make this sentence harder to understand.

src/I18n/DateFormatTrait.php
```diff
@@ -147,9 +145,8 @@ public function nice($timezone = null, $locale = null)
      * $time->i18nFormat(\IntlDateFormatter::FULL, 'Europe/Berlin', 'de-DE');
      * ```
      *
-     * You can control the default locale to be used by setting the static variable
-     * `Time::$defaultLocale` to a valid locale string. If empty, the default will be
-     * taken from the `intl.default_locale` ini config.
+     * You can control the default locale used by setting through `Time::setDefaultLocale()`.
```
markstory:
```suggestion
     * You can control the default locale used through `Time::setDefaultLocale()`.
```
Same as above.

src/I18n/DateFormatTrait.php
othercorey: Funny, I actually debated the same thing.

othercorey:
@markstory Do you prefer the formatting details provided at every place they are used or is there some way to condense them down to a single place that the i18nformat/json/etc documentation can reference?  Is it more annoying for people to have to look up a "Must be a format as described in ...." than it is to maintain multiple copies of this?

markstory:
I think the maintenance cost on these is pretty low. The only good place we have for consolidated documentation is in the cookbook.

Issue #13313, opened 2019-06-02
Completion command should ignore public methods for Shell


rochamarcelo:
This is a (multiple allowed):

* [ ] bug
* [x] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 4.x.
* Platform and Target: PHP 7.2.13

### What you did
baked a shell and overwritten the methods runCommand  and getOptionParser
bin/cake completion subcommands hello_world

### What happened
The output contains runCommand and  getOptionParser

### What you expected to happen
Completion for shells should ignore all public methods defined at \Cake\Console\Shell. This is the relevant line https://github.com/cakephp/cakephp/blob/7dca886d7739779515fa3390d1e96da9ea046bea/src/Command/CompletionCommand.php#L205  at CompletionCommand.php

markstory:
Pull request open now.

PR #13314, opened 2019-06-02
4.x Completion command should ignore public methods for Shell

rochamarcelo:
#13313 Completion command should ignore methods from core Shell


markstory (commented):
Looks good, but the code style errors from scrutinizer need to be fixed up.

PR #13318, opened 2019-06-04
3.next - Improve nested validator support in EntityContext

### LINKED ISSUE #13217: Form validation for nested field
This is a (multiple allowed):

* [x] bug
* [ ] enhancement

* CakePHP Version: 3.7.7
* Platform and Target: ubuntu, php7.2-fpm, nginx, mariadb

### What you did
Saving some custom features into a JSON field `options` i want to add rules into some of theses options

    /**
     * @param Validator $validator
     *
     * @return Validator
     */
    public function validationMenuSubPages(Validator $validator)
    {
        $validator = $this->validationDefault($validator);

        $subpages = new Validator();
        $subpages
                ->requirePresence('subpages', true)
                ->allowEmptyString('subpages', false);

        $validator
                ->addNested('options', $subpages)
                ->requirePresence('options', true)
                ->allowEmptyString('options', false);

        return $validator;
    }

This is my form

    <?= $this->Form->control('options.subpages', [
        'label' => __d("content_builder", "Pages à afficher"),
        'type' => 'select',
        'multiple' => true,
        'options' => $entity->get('subpages')
    ]) ?>

### What happened

Submitting the form without data, the validation work, but is not cathed by FormHelper to be reinjected into my form view...

    +"[errors]": array:1 [▼
        "options" => array:1 [▼
            "subpages" => array:1 [▼
                "_empty" => "This field cannot be left empty"
            ]
        ]
    ]

Seems related to #11206 

The validation error should be related to the view form and not out of context, here i need to manually bind theses validation (requierements / errors) into my form, breaking the need of using FormHelper

markstory:
Nested validation errors were previously hard to interact with. Both `Entity::getError()` and `EntityContext::error()` did not support them well. These changes update these methods to better handle nested validator scenarios.

I've not fixed the gaps in `$context->isRequired()` as it is a separate beast.

Fixes #13217

### DIFF
--- a/src/View/Form/EntityContext.php
+++ b/src/View/Form/EntityContext.php
@@ -313,11 +313,12 @@ class EntityContext implements ContextInterface
     }
 
     /**
-     * Fetch the leaf entity for the given path.
+     * Fetch the entity or data value for a given path
      *
-     * This method will traverse the given path and find the leaf
-     * entity. If the path does not contain a leaf entity false
-     * will be returned.
+     * This method will traverse the given path and find the entity
+     * or array value for a given path.
+     *
+     * If you only want the terminal Entity for a path use `leafEntity` instead.
      *
      * @param array|null $path Each one of the parts in a path for a field name
      *  or null to get the entity passed in constructor context.
@@ -349,7 +350,6 @@ class EntityContext implements ContextInterface
             $prop = $path[$i];
             $next = $this->_getProp($entity, $prop);
             $isLast = ($i === $last);
-
             if (!$isLast && $next === null && $prop !== '_ids') {
                 $table = $this->_getTable($path);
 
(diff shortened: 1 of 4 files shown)


src/View/Form/EntityContext.php
```diff
@@ -372,6 +372,74 @@ public function entity($path = null)
...
+    protected function leafEntity($path = null)
+    {
+        if ($path === null) {
+            return $this->_context['entity'];
+        }
+
+        $oneElement = count($path) === 1;
+        if ($oneElement && $this->_isCollection) {
+            throw new RuntimeException(sprintf(
+                'Unable to fetch property "%s"',
```
lorenzo: should this explain why it was not possible? or this is caught by the framework in all cases?

src/View/Form/EntityContext.php
```diff
@@ -372,6 +372,74 @@ public function entity($path = null)
...
+
+        $len = count($path);
+        $last = $len - 1;
+        $leafEntity = $entity;
+        for ($i = 0; $i < $len; $i++) {
+            $prop = $path[$i];
+            $next = $this->_getProp($entity, $prop);
+
+            // Did not dig into an entity, return the current one.
+            if (is_array($entity) && !($next instanceof EntityInterface || $entity instanceof Traversable)) {
```
lorenzo: I think the `$entity instanceof Traversable` is redundant, the first conditions rules out that possibility already.

src/View/Form/EntityContext.php
markstory: Right now I catch this exception all the time. I used an exception instead of adding a 'dirty' return value of `[false, []]`. Tuple returns values don't have great support in PHP so I didn't want to add more type variance.

src/View/Form/EntityContext.php
markstory: Whoops, this is supposed to be `$next instanceof Traversable`.

PR #13320, opened 2019-06-05
Backport 4.x model aware fix for inflected assignment.

dereuromark:
Backports 4.x fix of assignment only on true null (not on disable value)
See
https://github.com/cakephp/cakephp/blob/4.x/src/Datasource/ModelAwareTrait.php

Found via https://github.com/dereuromark/cakephp-ide-helper/issues/128 and that the inflected value is overwriting the false here which it shouldnt.

This should be fully BC for proper use of the designed functionality.

### DIFF
--- a/src/Datasource/ModelAwareTrait.php
+++ b/src/Datasource/ModelAwareTrait.php
@@ -66,7 +66,7 @@ trait ModelAwareTrait
      */
     protected function _setModelClass($name)
     {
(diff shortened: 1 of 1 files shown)


dakota (approved):
This definately fixes the issues where `$this->modelClass` is overwritten when explicitly set to false (Causing issues like https://github.com/dereuromark/cakephp-ide-helper/issues/128).

PR #13324, opened 2019-06-06
Backport 4.x deprecations for easier upgrading.

dereuromark:
Maybe someone can  help with it, or add more missing ones?
So far I extracted the obvious ones from 4.x migration guide here.

### DIFF
--- a/src/Datasource/EntityTrait.php
+++ b/src/Datasource/EntityTrait.php
@@ -556,7 +556,7 @@ trait EntityTrait
     }
 
     /**
-     * Get the list of visible properties.
+     * Gets the list of visible properties.
      *
      * The list of visible properties is all standard properties
      * plus virtual properties minus hidden properties.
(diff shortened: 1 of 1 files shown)


src/View/Helper/UrlHelper.php
```diff
@@ -48,6 +48,9 @@ public function build($url = null, $options = false)
             'escape' => true,
         ];
         if (!is_array($options)) {
+            deprecationWarning(
+                'The `bool` array option is deprecated. Use an array here instead with `\'fullBase\' => true`.'
+            );
```
markstory: This one causes a ton of failures in the test suite, and I'm not sure it is a high value change to be making this late in 3.x's life. It also doesn't add a significant burden to us to maintain going forwards.

dereuromark: You mean we shouldnt throw the deprecation warning?

dereuromark: I added it to shim plugin instead: https://github.com/dereuromark/cakephp-shim/pull/53

markstory: Yes, if this deprecation causes many of our tests to fail it will likely be tedious for application developers too.

saeideng: I'm :+1:  with dont adding more deprecation warning to cake 3.8
we can add it since cake 4.0

dereuromark:
Is this ok?

PR #13325, opened 2019-06-06
Remove deprecated method of 3.x.

dereuromark:
With https://github.com/cakephp/cakephp/pull/13324 and 3.8 landing before 4.0, we can safely remove this then as it was deprecated in 3.x.
One less thing to keep along until 5.0


dereuromark:
After 3.8 part is merged, we could now move forward on this one.

PR #13326, opened 2019-06-06
Backport 4.x validation

saeideng:
Refs  #12436
this will break your application if you are using 
https://github.com/FriendsOfCake/cakephp-upload + these methods

cc @chinpei215

### DIFF
--- a/src/Validation/Validation.php
+++ b/src/Validation/Validation.php
@@ -852,6 +852,10 @@ class Validation
      */
     public static function minLength($check, $min)
     {
+        if (!is_scalar($check)) {
+            return false;
+        }
+
         return mb_strlen($check) >= $min;
     }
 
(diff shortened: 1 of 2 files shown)


dereuromark (approved):
Thank you for this

saeideng:
this is not full BC probably :)
but fixes the incorrect use of these methods

saeideng:
```php
//before this change
Validation::maxLength(['abc'], ....) //true 
//now
Validation::maxLength(['abc'], ....) //false

**Comment by dereuromark** [comment]
Running any of them on an array shouldn't have been done in the first place.
But if we wanted, we could also ship this with 3.8 then?

**Review by markstory — APPROVED** [review]
Looks good to me. All of these methods would emit warning or notice errors without these checks.

File: src/Validation/Validation.php [hunk 1]
```diff
@@ -876,6 +884,10 @@ public static function maxLength($check, $max)
      */
     public static function minLengthBytes($check, $min)
     {
+        if (!is_scalar($check)) {
```
inoas: Couldn't a resource be checked against its byte length?

markstory: How would a resource come in request data?

PR #13327, opened 2019-06-07
Make sure we are not accidently reusing other property content.

dereuromark:
We should use a return type hint to avoid accidental use of existing properties for
```php
        if (isset($this->{$alias})) {
            return $this->{$alias};
        }
```
E.g. a loaded component or service with a colliding name.

### DIFF
--- a/src/Datasource/ModelAwareTrait.php
+++ b/src/Datasource/ModelAwareTrait.php
@@ -89,7 +89,7 @@ trait ModelAwareTrait
      * @throws \UnexpectedValueException If $modelClass argument is not provided
      *   and ModelAwareTrait::$modelClass property value is empty.
      */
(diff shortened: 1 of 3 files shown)


markstory:
Seems like a good idea. Test failures look relevant.

dereuromark:
Are those remaining fails still relevant? The error messages look weirdly unrelated.

dereuromark:
I could need some review/help on this one.

ADmad:
The test failure is quite strange. The test method does have `$this->expectException(\InvalidArgumentException::class)` yet the exception bubbles up and isn't being handled by phpunit.

Also don't see how changes done in this PR could be affecting that.

dereuromark:
I will abandon this then.

PR #13328, opened 2019-06-07
4.x serverShell => ServerCommand

saeideng:


src/Command/ServerCommand.php
```diff
@@ -106,33 +118,38 @@ public function startup(): void
...
+     * @param \Cake\Console\Arguments $args The command arguments.
+     * @param \Cake\Console\ConsoleIo $io The console io
+     * @return null|int The exit code or null for success
      */
-    public function main(): void
+    public function execute(Arguments $args, ConsoleIo $io): ?int
     {
+        $this->io = $io;
+        $this->args = $args;
+        $this->startup();
```
markstory: Any reason to use properties over passing arguments to `startup()`? Won't we get warnings from typecheckers that `$this->args` and `$this->io` are potentially null?

saeideng: >Any reason to use properties over passing arguments to startup()?

no reason, this method was a way to modify header of this shell 
I can change to add arguments , and I can merge  the`_welcome()` and `startup()`

>Won't we get warnings from typecheckers that $this->args and $this->io are potentially null?

since order of use is : `run()`->`execute()` ->`startup()`->`_welcome()`
we won't receive any warnings IMO

markstory: There are warnings in the psalm output for these properties.

markstory (approved):
Looks good to me.

PR #13331, opened 2019-06-10
Add composer command for running psalm

markstory:
I often want to run psalm by itself, and having this makes that easier.


dereuromark:
we could also have one for phpstan.

PR #13332, opened 2019-06-10
Split composer stan command to phpstan and psalm

garas:
Follow up of https://github.com/cakephp/cakephp/pull/13331#issuecomment-500305166


dereuromark:
I would keep the stan one for combined check as convenience wrapper. Doesnt hurt.

PR #13339, opened 2019-06-11
4.0 support underscored name in command

saeideng:
now support in core's command
I can apply to app/plugin's commands 

```
bin/cake --help
Current Paths:

* app:  src
* root: /var/www/html/test
* core: /var/www/html/test/vendor/cakephp/cakephp

Available Commands:

App:
 - sample

CakePHP:
 - cache clear
 - cache clear_all
 - cache list
 - completion
 - help
 - i18n
 - plugin
 - routes
 - routes check
 - routes generate
 - schema_cache build
 - schema_cache clear
 - server
 - upgrade
 - version

To run a command, type `cake command_name [args|options]`
To get help on a specific command, type `cake command_name --help`
```


src/Console/Command.php
```diff
@@ -104,6 +106,26 @@ public function getName(): string
...
     }
 
+    /**
+     * Proxy for static calls.
+     *
+     * @param string $name Method name.
+     * @param array $arguments Method arguments.
+     * @return mixed
+     */
+    public static function __callStatic($name, $arguments)
```
ADmad: Why use the magic method instead of just having a `Command::defaultName()` method? You can throw an exception is the method is called directly on `Command` class instead of a child class.

markstory: Or provide a default implementation on the base class.

src/Console/CommandScanner.php
```diff
@@ -64,18 +64,11 @@ public function scanCore(): array
...
-
-            // Maintain backwards compatibility for schema_cache as it is likely
-            // hooked up to people's deploy tooling
-            if (strpos($command['name'], 'schemacache') === 0) {
-                $compat = [
-                    'name' => str_replace('schemacache', 'schema_cache', $command['name']),
-                    'fullName' => str_replace('schemacache', 'schema_cache', $command['fullName']),
-                ];
-                $commands[] = $compat + $command;
+            $name = $command['class']::getDefaultName();
```
markstory: Would we also want this same behavior for app & plugin shells?

saeideng: yes , I'll add ,and I think this is Full BC

src/Console/Command.php
saeideng: is `getDefaultName()` good name ?

ADmad: We could just use `defautName()` as there won't be any corresponding setter.

src/Console/CommandScanner.php
markstory: It is if we include a default implementation.

saeideng:
just there is an error in `completion commands` now

src/Console/CommandScanner.php
saeideng: if user uses `UpdateTableCommand` this will show `update table` 
but documentation (in cake 3) says this should be `update_table`

saeideng: I can change the default to `underscore` if it should be

src/Console/Command.php
```diff
@@ -104,6 +105,20 @@ public function getName(): string
...
 
+    /**
+     * Get the command name.
+     *
+     * @return string
+     */
+    public static function defaultName(): string
+    {
+        $class = static::class;
+        $class = preg_replace('/(Shell|Command)$/', '', $class);
```
ADmad: I don't think `Shell` is needed.

src/Console/CommandScanner.php
```diff
@@ -64,19 +65,15 @@ public function scanCore(): array
...
-
-            // Maintain backwards compatibility for schema_cache as it is likely
-            // hooked up to people's deploy tooling
-            if (strpos($command['name'], 'schemacache') === 0) {
-                $compat = [
-                    'name' => str_replace('schemacache', 'schema_cache', $command['name']),
-                    'fullName' => str_replace('schemacache', 'schema_cache', $command['fullName']),
-                ];
-                $commands[] = $compat + $command;
+            $file = new SplFileInfo($command['file']);
```
ADmad: Why do you need to use `SplFileInfo`? If `$command['file']` is full file path just use string functions to get the filename.

src/Console/CommandScanner.php
```diff
@@ -65,8 +64,7 @@ public function scanCore(): array
     protected function inflectCommandNames(array $commands): array
     {
         foreach ($commands as $i => $command) {
-            $file = new SplFileInfo($command['file']);
-            $file = $file->getFilename();
+            $file = substr($command['file'], (int)strripos($command['file'], DIRECTORY_SEPARATOR) + 1);
```
ADmad: The following should also yield same result and is much more readable:

```php
$file = basename($command['file'], '.php');
$name = Inflector::underscore(preg_replace('/(Shell|Command)$/', '', $file));
```

src/Console/Command.php
```diff
@@ -104,6 +105,23 @@ public function getName(): string
...
+     */
+    public static function defaultName(): string
+    {
+        $pos = strrpos(static::class, '\\');
+        $name = substr(static::class, $pos + 1, -7);
+
+        return str_replace(
+            '_',
+            ' ',
+            Inflector::underscore($name)
```
saeideng: @ADmad we should use underscore name for user's commands 
your commit does not show this

ADmad: Hmm.. Why should user command not follow the same convention as core commands?

saeideng: I dont know 
https://book.cakephp.org/3.0/en/console-and-shells/commands.html#testing-commands ( search the `user_table` )
in document and cake 36./ 3.7 it is underscore by default

ADmad: Boo.. Okay, I'll update.

saeideng: in cake 3.x file names point to underscore (shell + command ) and subcommand to `space` 
we are facing with 2 concept :)

ADmad: Damn BC :rage:

src/Console/Command.php
```diff
@@ -104,6 +105,29 @@ public function getName(): string
...
+     *
+     * @return string
+     */
+    public static function defaultName(): string
+    {
+        $pos = strrpos(static::class, '\\');
+        $name = substr(static::class, $pos + 1, -7);
+        $name = Inflector::underscore($name);
+
+        if (substr(static::class, 0, 4) === 'Cake') {
```
ADmad: I think it would be best to remove this core specific code and simply add `defaultName()` to all core commands which return static strings. That would be a good micro optimization too.

saeideng: yes we can do it,  just there are 22 command classes :)
I'll add them

ADmad: Actually adding static name is necessary only for multi worded core commands. I have done that :slightly_smiling_face:.

saeideng:
ready to merge

src/Command/I18nExtractCommand.php
```diff
@@ -31,6 +31,14 @@
  */
 class I18nExtractCommand extends Command
 {
+    /**
+     * @inheritDoc
```
garas: It should be `{@inheritDoc}` everywhere

saeideng: we  used both syntax in cake but now `stickler-ci ` is unhappy with `@inheritDoc'
I can change it to see what happen

ADmad: `{@inheritDoc}` for inline heriting. It's supposed to be `@inheritDoc` to inherit the whole docblock.

Our 3.x usage inaccurate which has been fixed in 4.x.

ADmad (approved):

PR #13349, opened 2019-06-15
4.x fixtures

ADmad:
The relational DB specific `createConstraints()` and `dropConstraints()` methods of `FixtureInterface` have been moved to new `Cake\Database\ConstraintsInterface` (suggestions for alternate name for interface are welcome).

This will allow removing no-op methods 
`Cake\ElasticSearch\TestSuite\TestFixture::createConstraints()/dropConstraints()`.

### DIFF
--- a/src/Database/ConstraintsInterface.php
+++ b/src/Database/ConstraintsInterface.php
@@ -1,0 +1,45 @@
+<?php
+declare(strict_types=1);
+
+/**
+ * CakePHP(tm) : Rapid Development Framework (https://cakephp.org)
+ * Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
+ *
+ * Licensed under The MIT License
+ * For full copyright and license information, please see the LICENSE.txt
+ * Redistributions of files must retain the above copyright notice.
+ *
+ * @copyright     Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
+ * @link          https://cakephp.org CakePHP(tm) Project
+ * @since         4.0.0
+ * @license       https://opensource.org/licenses/mit-license.php MIT License
+ */
+namespace Cake\Database;
+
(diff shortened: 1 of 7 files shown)


src/Database/ConstraintsInterface.php
```diff
@@ -0,0 +1,45 @@
...
+ * @license       https://opensource.org/licenses/mit-license.php MIT License
+ */
+namespace Cake\Database;
+
+use Cake\Datasource\ConnectionInterface;
+
+/**
+ * Defines the interface for managing constraints.
+ */
+interface ConstraintsInterface
```
markstory: Should this interface be in datasources alongside the FixtureInterface? Having it here would require a plugin implementation of a datastore that includes constraints to depend on the datasource + database packages.

ADmad: Are there any non RDBMS which have constraints? I believe any implementation requiring ConstraintsInterface would be using the database package already.

markstory: Depends how strictly you define RDBMS. There are systems like cockroach that are relational and have constraints but wouldn't get much from extending PDO based query objects.

ADmad: Apparently the recommended way to use cockroachdb in PHP is through PDO's pgsql driver :slightly_smiling_face: 

https://www.cockroachlabs.com/docs/stable/build-a-php-app-with-cockroachdb.html

markstory: Didn't expect to see that 😮

ADmad: So is it okay to keep the `ConstraintsInterface` here or you still prefer moving it under `Datasource`?

markstory: It can stay here, I'm not too fussy on where it lives.

Issue #13351, opened 2019-06-15
isInteger() validation returns true when provided with boolean value "true"


georgeconstantinou:
* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.7.8
* Platform and Target: PHP's build-in webserver.

### What you did
```php
<?php
use Cake\Validation\Validation;

$value = true;
$result = Validation::isInteger($value);
```
### What happened
`$result` is **true**.

### What you expected to happen
`$result` should be **false**.

saeideng:
PR open.

PR #13352, opened 2019-06-15
Fix Validation::isInteger() for boolean values

georgeconstantinou:
Issue link: https://github.com/cakephp/cakephp/issues/13351

Validation::isInteger() method will now return `false` when provided with a boolean value.

### DIFF
--- a/src/Validation/Validation.php
+++ b/src/Validation/Validation.php
@@ -1552,7 +1552,7 @@ class Validation
      */
     public static function isInteger($value)
     {
-        if (!is_scalar($value) || is_float($value) || is_bool($value)) {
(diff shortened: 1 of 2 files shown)


saeideng:
this is not fully BC IMO
and I think it is Intentional, we can not change it in master at least

markstory:
@saeideng I think fixing gaps in validators should be ok in bugfixes. This doesn't change any existing tests or API signatures and the method name implies the behavior which is being added now.

saeideng:
@markstory OK
my concern was about `validator`
```php
        $validator
            ->integer('....')
```
but looks was wrong

saeideng:
what happened if someone uses `integer('....')` with data like `active = true'?

PR #13354, opened 2019-06-16
Improve Psalm bootstrapping

muglug:
This helps Psalm recognise the import paths, so it can understand `debug` function calls etc.

It also allows `Memcached` and `Redis` to not exist wherever Psalm is run. That second change may not be ideal, happy to undo it (though neither extension is required in your composer.json).

PR #13355, opened 2019-06-17
4.x error handler

ADmad:
Made code DRY by making `ErrorHandlerMiddleware` use `ErrorHandler`. Also delegated error logging to new `ErrorLogger`.

Refs #11323.

### DIFF
--- a/src/Console/ConsoleErrorHandler.php
+++ b/src/Console/ConsoleErrorHandler.php
@@ -36,16 +36,17 @@ class ConsoleErrorHandler extends BaseErrorHandler
     /**
      * Constructor
      *
-     * @param array $options Options for the error handler.
+     * @param array $config Config options for the error handler.
      */
-    public function __construct(array $options = [])
+    public function __construct(array $config = [])
     {
-        $options += [
+        $config += [
             'stderr' => new ConsoleOutput('php://stderr'),
             'log' => false,
         ];
-        $this->_stderr = $options['stderr'];
-        $this->_options = $options + $this->_options;
+
+        $this->setConfig($config);
+        $this->_stderr = $this->_config['stderr'];
     }
 
     /**
--- a/src/Console/ConsoleErrorHandler.php
+++ b/src/Console/ConsoleErrorHandler.php
@@ -60,7 +60,7 @@ class ConsoleErrorHandler extends BaseErrorHandler
     public function handleException(Throwable $exception): void
     {
         $this->_displayException($exception);
-        $this->_logException($exception);
+        $this->logException($exception);
         $code = $exception->getCode();
         $code = $code && is_int($code) ? $code : 1;
         $this->_stop($code);
--- a/src/Console/ConsoleErrorHandler.php
+++ b/src/Console/ConsoleErrorHandler.php
@@ -34,24 +34,18 @@ class ConsoleErrorHandler extends BaseErrorHandler
     protected $_stderr;
 
     /**
-     * Options for this instance.
-     *
-     * @var array
-     */
-    protected $_options;
-
-    /**
      * Constructor
      *
      * @param array $options Options for the error handler.
      */
     public function __construct(array $options = [])
     {
-        if (empty($options['stderr'])) {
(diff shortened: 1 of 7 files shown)


ADmad:
The change in `ErrorHandlerMiddleware`'s constructor is non BC but I think it's acceptable given the benefits of this change. Only those explicitly passing a exception renderer to the middleware will have to update the middleware instantiation call.

src/Console/ConsoleErrorHandler.php
```diff
@@ -33,25 +33,19 @@ class ConsoleErrorHandler extends BaseErrorHandler
...
      *
      * @param array $options Options for the error handler.
      */
     public function __construct(array $options = [])
     {
-        if (empty($options['stderr'])) {
-            $options['stderr'] = new ConsoleOutput('php://stderr');
-        }
+        $options += [
+            'stderr' => new ConsoleOutput('php://stderr'),
```
dereuromark: Is this lazy generated in php 7?

ADmad: IDK, what's your concern?

dereuromark: Performance on lots of loops where this wouldnt be needed.
That was sure the reason this was lazy added via if statement.
Or is there some service singleton pattern on this, so this is only instant. once anyway?

ADmad: This is the console error handler for CLI, I doubt this will have any noticeable impact even over hundreds of calls.

src/Error/BaseErrorHandler.php
```diff
@@ -302,107 +312,36 @@ protected function _logError($level, array $data): bool
...
-        return $message;
-    }
-
-    /**
-     * Generates a formatted error message
-     *
-     * @param \Throwable $exception Exception instance
-     * @return string Formatted message
-     */
-    protected function _getMessage(Throwable $exception): string
```
markstory: We should note in the migration guide that the internals of Error handling have changed and that any sub-classes may need to be re-worked.

src/Error/ErrorHandler.php
```diff
@@ -96,11 +98,9 @@ class ErrorHandler extends BaseErrorHandler
     public function __construct(array $options = [])
     {
         $defaults = [
-            'log' => true,
-            'trace' => false,
             'exceptionRenderer' => ExceptionRenderer::class,
         ];
-        $this->_options = $options + $defaults;
+        $this->_options = $options + $defaults + $this->_options;
```
markstory: Should this class use `InstanceConfigTrait`?

src/Error/ErrorLogger.php
```diff
@@ -0,0 +1,145 @@
...
+
+use Cake\Core\Configure;
+use Cake\Core\Exception\Exception;
+use Cake\Core\InstanceConfigTrait;
+use Cake\Log\Log;
+use Psr\Http\Message\ServerRequestInterface;
+use Throwable;
+
+/**
+ * Log errors.
```
markstory:
```suggestion
 * Log errors and unhandled exceptions to `Cake\Log\Log`
```

src/Error/Middleware/ErrorHandlerMiddleware.php
```diff
@@ -53,42 +49,41 @@ class ErrorHandlerMiddleware implements MiddlewareInterface
      *   ```
      *
      * - `trace` Should error logs include stack traces?
+     * - `exceptionRenderer` The renderer instance or class name to use or a callable factory
+     *   which return \Cake\Error\ExceptionRendererInterface instance.
```
markstory:
```suggestion
     *   which returns a \Cake\Error\ExceptionRendererInterface instance.
```

src/Error/ErrorHandler.php
ADmad: Yeah I guess making `BaseErrorHandler` use `InstanceConfigTrait` would be a good idea now.

ADmad:
The `static-analysis` build failure is unrelated and should be fixed once #13354 is merged.

PR #13356, opened 2019-06-17
Add unit test to test Auth with custom model

bancer:
Using custom model for authentication with BasicAuthenticate results in SQL error.
It seems that the problem is [here](https://github.com/cakephp/cakephp/blob/2.x/lib/Cake/Controller/Component/Auth/BaseAuthenticate.php#L125) in BaseAuthenticate:
`$result = ClassRegistry::init($userModel)->find('first', array(`.

The model is constructed with wrong alias - AuthUser (or other name of the custom model) instead of User.

What would be the best approach to fix this? To add 'User' alias to `$userModel`?

I got this error in tests:
```
1) BasicAuthenticateTest::testAuthenticateSuccessCustomModel
PDOException: SQLSTATE[42S22]: Column not found: 1054 Unknown column 'AuthUser.user' in 'where clause'

/home/val/git/CakePHP-framework/lib/Cake/Model/Datasource/DboSource.php:502
/home/val/git/CakePHP-framework/lib/Cake/Model/Datasource/DboSource.php:468
/home/val/git/CakePHP-framework/lib/Cake/Model/Datasource/DboSource.php:715
/home/val/git/CakePHP-framework/lib/Cake/Model/Datasource/DboSource.php:1226
/home/val/git/CakePHP-framework/lib/Cake/Model/Model.php:3053
/home/val/git/CakePHP-framework/lib/Cake/Model/Model.php:3025
/home/val/git/CakePHP-framework/lib/Cake/Controller/Component/Auth/BaseAuthenticate.php:129
/home/val/git/CakePHP-framework/lib/Cake/Controller/Component/Auth/BasicAuthenticate.php:97
/home/val/git/CakePHP-framework/lib/Cake/Controller/Component/Auth/BasicAuthenticate.php:75
/home/val/git/CakePHP-framework/lib/Cake/Test/Case/Controller/Component/Auth/BasicAuthenticateTest.php:218
/home/val/git/CakePHP-framework/lib/Cake/TestSuite/CakeTestCase.php:84
/home/val/git/CakePHP-framework/lib/Cake/TestSuite/CakeTestRunner.php:75
/home/val/git/CakePHP-framework/lib/Cake/TestSuite/CakeTestSuiteCommand.php:98
/home/val/git/CakePHP-framework/lib/Cake/Console/Command/TestShell.php:283
/home/val/git/CakePHP-framework/lib/Cake/Console/Command/TestShell.php:268
/home/val/git/CakePHP-framework/lib/Cake/Console/Shell.php:462
/home/val/git/CakePHP-framework/lib/Cake/Console/ShellDispatcher.php:222
/home/val/git/CakePHP-framework/lib/Cake/Console/ShellDispatcher.php:66
```

### DIFF
--- a/lib/Cake/Controller/Component/Auth/BaseAuthenticate.php
+++ b/lib/Cake/Controller/Component/Auth/BaseAuthenticate.php
@@ -102,7 +102,8 @@ abstract class BaseAuthenticate implements CakeEventListener {
  */
 	protected function _findUser($username, $password = null) {
 		$userModel = $this->settings['userModel'];
-		$alias = 'User';
+		list(, $model) = pluginSplit($userModel);
+		$alias = stripslashes($model);
 		$fields = $this->settings['fields'];
 
 		if (is_array($username)) {
--- a/lib/Cake/Controller/Component/Auth/BaseAuthenticate.php
+++ b/lib/Cake/Controller/Component/Auth/BaseAuthenticate.php
@@ -122,10 +122,10 @@ abstract class BaseAuthenticate implements CakeEventListener {
 			$userFields[] = $alias . '.' . $fields['password'];
 		}
 
-		$User = ClassRegistry::init([
+		$User = ClassRegistry::init(array(
 			'class' => $userModel,
 			'alias' => $alias,
-		]);
+		));
 		$result = $User->find('first', array(
 			'conditions' => $conditions,
 			'recursive' => $this->settings['recursive'],
--- a/lib/Cake/Controller/Component/Auth/BaseAuthenticate.php
+++ b/lib/Cake/Controller/Component/Auth/BaseAuthenticate.php
@@ -102,14 +102,14 @@ abstract class BaseAuthenticate implements CakeEventListener {
  */
 	protected function _findUser($username, $password = null) {
 		$userModel = $this->settings['userModel'];
-		list(, $model) = pluginSplit($userModel);
+		$alias = 'User';
 		$fields = $this->settings['fields'];
 
 		if (is_array($username)) {
 			$conditions = $username;
 		} else {
 			$conditions = array(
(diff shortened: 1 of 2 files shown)


ADmad:
The error says `Unknown column 'AuthUser.user'`. How does is end up looking for column named `user`?  I doubt the alias is the problem here.

ADmad:
There are existing tests with `userModel` name besides `User`.

https://github.com/cakephp/cakephp/blob/396e342601d74175df981d8cda0a33dd64c348c6/lib/Cake/Test/Case/Controller/Component/Auth/FormAuthenticateTest.php#L276

bancer:
@ADmad You are right. I had `'fields' => array('username' => 'user', 'password' => 'password'),` in the test at first.

I tried to reproduce the problem we encountered when User model has a namespace. I thought I got the same problem just with a different model name.

I got `"SQLSTATE[42S22]: Column not found: 1054 Unknown column 'App\Model\User.id' in 'field list'"` on localhost when User model has `namespace App\Model;`.

It seems that it is fixed by using 'User' alias though with all tests passed.

ADmad:
namespace?  2.x doesn't support namespaces.

ADmad:
Your change which sets the alias to `User` always will break existing apps which use the `scope` option and non default model name like `'scope' => array('Foor.active' => 1)`.

bancer:
Would it be better to add an alias to the component's configuration array?
```
$this->Auth->authenticate = array(
    'Basic' => array(
        'userModel' => 'Member', 
        'userModelAlias' => 'User',
    ),
);

```

ADmad:
No, we are no longer adding new features to 2.x.

ADmad:
You can just use a custom authenticator class with the special features you need. We can't create inconsistency in the codebase by allowing just the authenticator class to use namespaced model names.

Issue #13342, opened 2019-06-12
FormHelper overwrites "idPrefix" when rendering RadioWidget: no way to influence ids


advitum:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [x] feature-discussion (RFC)

* CakePHP Version: 3.7.8; current master (commit e15000344b97fdcd5882db187b2ef0ca13d82991); current 4.x branch
* Platform and Target: Apache, PHP 7.3, Web-Request (Form generation in frontend)

### What you did
When rendering a Radio Widget, neither the "id" nor "idPrefix" are used. Thus, the ids cannot be modified when using the FormHelper.

```
$this->Form->control('my-name', [
	'type' => 'radio',
	'options' => ['value1', 'value2'],
	'id' => 'willNotBeUsed',
	'idPrefix' => 'willBeOverwrittenInFormHelper',
]);
```

### What happened
The inputs are generated using the "idPrefix" set to the Form (null in most cases) and using the name of the input. So the ids will be "my-name-value1", "my-name-value2" etc.

### What you expected to happen
I expected the ids to be "willNotBeUsed-value1" and "willNotBeUsed-value2", or at least "willBeOverwrittenInFormHelper-my-name-value1" or "willBeOverwrittenInFormHelper-my-name-value2" (or something similar).

### Additional information
It seems the problem lies in this line in the FormHelper: https://github.com/cakephp/cakephp/blob/e15000344b97fdcd5882db187b2ef0ca13d82991/src/View/Helper/FormHelper.php#L1693 (which is still present in the 4.x branch: https://github.com/cakephp/cakephp/blob/c05c4151e6479441742f7e9486f32b3805e28f58/src/View/Helper/FormHelper.php#L1571) - for no apparent reason, the "idPrefix" is overwritten without checking if it is already present, thus overwriting the passed value for "idPrefix".

Since the passed "id"-value is (correctly?) not being used in the RadioWidget, and the idPrefix never reaches the RadioWidget thanks to the FormHelper, that leaves us with no option to modify the ids (except extending the FormHelper).

As I see it, the idPrefix should only be set when not already present.

markstory:
Any chance you would want to make a pull request to fix this problem? I agree that `idPrefix` should be conditionally set by FormHelper.

advitum:
Sure thing, will do tomorrow! I just wanted to make sure that this is indeed a defect.

advitum:
[...]

PR #13357, opened 2019-06-18
Fix for issue 13342 - RadioWidget and MultiCheckboxWidget now use passed "id"

advitum:
Fixing this issue: https://github.com/cakephp/cakephp/issues/13342

The original issue was that in the RadioWidget and MultiCheckboxWidget, there is no way to influence the generated ids. I originally thought the Widgets should use the idPrefix that you can pass as an option.

But after working with the code of the FormHelper and other Widgets, it turns out all the other widgets use the idPrefix of the FormHelper too. So this was not the bug.

The actual problem lies in the fact that the "id" you can pass to the widget gets ignored by the RadioWidget and MultiCheckboxWidget. All other widgets use the passed id.

The cookbook clearly states:

> To reduce repetition, the common options shared by all control methods are as follows:
> 'id' - Set this key to force the value of the DOM id for the control. This will override the 'idPrefix' that may be set.
> [...]
https://book.cakephp.org/3.0/en/views/helpers/form.html#common-options-for-specific-controls

So the logical fix would be to change the behaviour of the RadioWidget and MultiCheckboxWidget so they use the passed "id" to.

### DIFF
--- a/src/View/Helper/IdGeneratorTrait.php
+++ b/src/View/Helper/IdGeneratorTrait.php
@@ -59,7 +59,21 @@ trait IdGeneratorTrait
     protected function _id($name, $val)
     {
         $name = $this->_domId($name);
+        $suffix = $this->_idSuffix($val);
 
+        return trim($name . '-' . $suffix, '-');
+    }
+
+    /**
+     * Generate an ID suffix.
+     *
+     * Ensures that id's for a given set of fields are unique.
+     *
+     * @param string $val The ID attribute value.
+     * @return string Generated id suffix.
+     */
+    protected function _idSuffix($val)
+    {
         $idSuffix = mb_strtolower(str_replace(['/', '@', '<', '>', ' ', '"', '\''], '-', $val));
         $count = 1;
         $check = $idSuffix;
@@ -68,7 +82,7 @@ trait IdGeneratorTrait
         }
         $this->_idSuffixes[] = $check;
 
-        return trim($name . '-' . $check, '-');
+        return $check;
     }
 
     /**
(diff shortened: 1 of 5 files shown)


src/View/Helper/IdGeneratorTrait.php
```diff
@@ -59,7 +59,21 @@ protected function _clearIds()
     protected function _id($name, $val)
     {
         $name = $this->_domId($name);
+        $suffix = $this->_idSuffix($val);
 
+        return trim($name . '-' . $suffix, '-');
```
dereuromark: Why do we need trim() here?

advitum: It is here in case the fieldname starts or the value of the option ends with a dash. The HTML id attribute must not start with a dash, hence the trim. Might break some backwards compatibility when we remove it if someone uses field names starting with dashes.

ADmad:
With the addition of new protected method it's best to target this to `3.next` branch. 3.8 stable release is pretty close anyway.

src/View/Widget/MultiCheckboxWidget.php
```diff
@@ -164,7 +164,14 @@ protected function _renderInputs($data, $context)
             $checkbox['checked'] = $this->_isSelected($checkbox['value'], $data['val']);
             $checkbox['disabled'] = $this->_isDisabled($checkbox['value'], $data['disabled']);
             if (empty($checkbox['id'])) {
-                $checkbox['id'] = $this->_id($checkbox['name'], $checkbox['value']);
+                if(isset($data['id'])) {
```
markstory:
```suggestion
                if (isset($data['id'])) {
```

src/View/Widget/RadioWidget.php
```diff
@@ -165,7 +165,14 @@ protected function _renderInput($val, $text, $data, $context)
         }
 
         if (empty($radio['id'])) {
-            $radio['id'] = $this->_id($radio['name'], $radio['value']);
+            if(isset($data['id'])) {
```
markstory:
```suggestion
            if (isset($data['id'])) {
```

PR #13359, opened 2019-06-18
Fix Type::buildAll() not returning type instance set using Type::set()

### LINKED ISSUE #13358: Let Cake\Database\Type::buildAll also return types added with Type::set
This is a (multiple allowed):

* [x] bug

* CakePHP Version: 3.7.5
* Platform and Target: not relevant for this issue

### What you did
I register a type with
```php
Type::set('somename', $someInstance);
```

Note that I can't use `Type::map` because I want an *instance*, because I have things configured on it that are needed when converting data to/from the database.

### What happened
My fields are not correctly converted from the database to PHP and vice versa.

I would expect that my data would get correctly converted using the type.

### Further info
After some debugging and digging, I found out that if I change `Type::buildAll()` to return `static::$_builtTypes` instead of `$result` all works well, because any types that are set using `Type::set` will then be returned.
I am not sure if this is indeed the best fix, but it seems logical to me. Either that or `Type::set` should be changed/removed.

p.s. the documentation of `Type::set` also seems confusing to me.

ADmad:
Closes #13358

### DIFF
--- a/src/Database/Type.php
+++ b/src/Database/Type.php
@@ -138,6 +138,7 @@ class Type implements TypeInterface
     public static function set($name, Type $instance)
     {
         static::$_builtTypes[$name] = $instance;
+        static::$_types[$name] = get_class($instance);
     }
 
     /**
--- a/src/ORM/ResultSet.php
+++ b/src/ORM/ResultSet.php
@@ -155,6 +155,7 @@ class ResultSet implements ResultSetInterface
      * Converters are indexed by alias and column name.
      *
      * @var array
+     * @deprecated 3.2.0 Not used anymore. Type casting is done at the statement level
      */
     protected $_types = [];
 
@@ -443,6 +444,7 @@ class ResultSet implements ResultSetInterface
      * @param \Cake\ORM\Table $table The table from which to get the schema
      * @param array $fields The fields whitelist to use for fields in the schema.
      * @return array
+     * @deprecated 3.2.0 Not used anymore. Type casting is done at the statement level
      */
     protected function _getTypes($table, $fields)
     {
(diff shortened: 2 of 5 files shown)


ADmad:
By the way one can use `Type::map('foo', $instance)` too :slightly_smiling_face:.

nickygerritsen:
> By the way one can just use `Type::map('foo', $instance)` in 3.x 🙂.

Note that it does give a deprecation warning if you do so :)

src/Database/Type.php
```diff
@@ -138,6 +138,7 @@ public static function buildAll()
     public static function set($name, Type $instance)
     {
         static::$_builtTypes[$name] = $instance;
+        static::$_types[$name] = get_class($instance);
```
nickygerritsen: Wait, shouldn't this just be 
```
static::$_types[$name] =$instance;
```
? `Type::build` will understand this as it will only construct a new entity when the value is a class. And this allows one to pass instances to `Type::set` that actually contain configuration data and/or have constructors that do not only accept `$name`.

ADmad: > And this allows one to pass instances to Type::set that actually contain configuration data and/or have constructors that do not only accept $name.

Nothing preventing you to do so now either.

You suggestion will work for 3.x but not 4.x where `$_types` only contains mapping to class name strings. My fix resolves the issue for both versions.

ADmad:
> Note that it does give a deprecation warning if you do so :)

Right, then `Type::setMap('foo', $instance)`.

src/Database/Type.php
nickygerritsen: Ah wait you are right; it will only loose this information when you call `clear` :)

PR #13362, opened 2019-06-19
4.0 move plugin shell to command

saeideng:


ADmad:
It would be nice if the current history of shell and tasks can be retained. For that you would have to first rename the shell/task files to corresponding commands using `git mv` and then make required code changes.

saeideng:
@ADmad  except `PluginAssetsCopyCommand` / `PluginAssetsRemoveCommand` / `PluginAssetsCopyCommand`
that  it was not possible , I used `git mv` for all other classes 
I dont know why `git` does not show correctly

ADmad:
@saeideng Okay, no problem.

saeideng:
@ADmad I pushed again with separate  commits 
again Github does not show correctly , but you can review commits  separately , it shows correct now

src/Command/PluginAssetsTrait.php
```diff
@@ -14,70 +14,32 @@
...
 use Cake\Filesystem\Filesystem;
 use Cake\Utility\Inflector;
 
 /**
- * Task for symlinking / copying plugin assets to app's webroot.
+ * trait for symlinking / copying plugin assets to app's webroot.
+ * @internal
  */
-class AssetsTask extends Shell
+trait PluginAssetsTrait
```
markstory: Instead of a trait you could make simple class that exposes the required methods as public methods. We could do that change in a follow-up pull request as well though.

src/Command/PluginUnloadCommand.php
```diff
@@ -0,0 +1,94 @@
...
+
+    /**
+     * Get the option parser.
+     *
+     * @param \Cake\Console\ConsoleOptionParser $parser The option parser to update
+     * @return \Cake\Console\ConsoleOptionParser
+     */
+    public function buildOptionParser(ConsoleOptionParser $parser): ConsoleOptionParser
+    {
+        $parser->addArgument('plugin', [
```
markstory: Should this have a description as well?

src/Command/PluginAssetsTrait.php
saeideng: probably yes but I no have idea for good class/methods name and place it 
I marked this as `@internal`
I think we can do it in next PR

tests/TestCase/Command/PluginUnloadCommandTest.php
```diff
@@ -0,0 +1,182 @@
...
+
+    /**
+     * Data provider for various forms.
+     *
+     * @return array
+     */
+    public function variantProvider()
+    {
+        return [
+            //  $this->addPlugin('TestPlugin', [
```
ADmad: The commented code needs to be removed.

saeideng: these are sample to show how would be value provider items 
see old code
https://github.com/cakephp/cakephp/blob/b879ff9e15c8519ecda8e8a2708f7aa08493e6af/tests/TestCase/Shell/Task/UnloadTaskTest.php#L110-L114

ADmad: Ah, never mind than.

saeideng:
ready to merge

markstory:
Looks good to me. I'll get this merged when bake is ready with a forwards/backwards compatible change so we don't break people trying out the beta.

PR #13363, opened 2019-06-20
Add return typehint for ModelAwareTrait::loadModel().

ADmad:


ADmad:
`static-analysis` failure is unrelated and fixed on `4.x`.

PR #13365, opened 2019-06-20
Fixed MailContains constraint when asserting against regexp characters

jeremyharris:
Fixes a bug when asserting for mail contents that contain regular expression characters.

This could be seen as a BC break if the user was already escaping the characters, in which case I can retarget.


markstory:
This should be fine. Basic string contains should not also support regex characters. We can have different assertions for that.

PR #13369, opened 2019-06-23
4.x security

ADmad:
Just some cleanup in `Security` class.

With `libsodium` being available by default in PHP 7.2, should we switch the default encryption engine from `openssl` to `libsodium`?

At the very least we should formalize the setting/usage of encryption engine in `Security` by providing an interface for the engines and removing `@internal` tag from `Utility\Crypto\OpenSsl`.

### DIFF
--- a/src/Utility/Crypto/OpenSsl.php
+++ b/src/Utility/Crypto/OpenSsl.php
@@ -19,9 +19,6 @@ namespace Cake\Utility\Crypto;
 /**
  * OpenSSL implementation of crypto features for Cake\Utility\Security
  *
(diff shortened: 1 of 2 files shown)


markstory:
If libsodium is compatible with existing openssl data then switching would be nice as it removes an optional dependency.

If the two implementations are not compatible we could add an adapter for libsodium.

ADmad:
The `OpenSsl` class uses `AES-256-CBC` while sodium only seems to provide functions for `AES-256-GCM`. So I don't think we can switch to sodium in a backwards compatible way.

ADmad:
This should be good to merge. Adding adapter for libsodium can be done in a follow up PR.

PR #13372, opened 2019-06-24
php-redis 'delete' alias function is deprecated since php-redis v5. 

cedricalfonsi:
The php-redis 'delete' alias function is deprecated since php-redis v5.
Use the php-redis 'del' function instead.


ADmad (approved):

markstory:
Thank you 👍

PR #13379, opened 2019-06-27
Remove duplicate tests.

ADmad:
All removed asserts are already present in `testAlphaNumeric()` method above. Seems they got duplicated at some point due to bad merge conflict resolution.


ravage84:
@saeideng  wouldn't the milestone for this be 3.8.1, insteadd of 3.8.0?

saeideng:
yes fixed

PR #13380, opened 2019-06-27
Implement a `removeClass()` option in addition to `addClass()` in src/View/StringTemplate.php

groovenectar:
Implement the option to remove default classes as well as adding them. Uses similar logic as `addClass()`, except that it will `array_diff()` instead of `array_merge()`

A example case where this comes in handy are custom Bootstrap helpers. Perhaps a button usually gets class `btn-primary` assigned by default, and it may be desirable to remove that in order to use a different or blank button class

### DIFF
--- a/src/View/StringTemplate.php
+++ b/src/View/StringTemplate.php
@@ -369,4 +369,46 @@ class StringTemplate
 
         return $input;
     }
+
+    /**
+     * Removes a class and returns a unique list either in array or space separated
+     *
+     * @param array|string $input The array or string to remove the class from
+     * @param array|string $removeClass the new class or classes to remove
+     * @param string $useIndex if you are inputting an array with an element other than default of 'class'.
+     * @return array|string
+     */
+    public function removeClass($input, $removeClass, $useIndex = 'class')
+    {
+        // NOOP
+        if (empty($removeClass)) {
+            return $input;
+        }
+
+        if (is_array($input)) {
+            $class = Hash::get($input, $useIndex, []);
+        } else {
+            $class = $input;
+            $input = [];
+        }
+
+        // Convert and sanitise the inputs
+        if (!is_array($class)) {
+            if (is_string($class) && !empty($class)) {
+                $class = explode(' ', $class);
+            } else {
+                $class = [];
+            }
+        }
+
+        if (is_string($removeClass)) {
+            $removeClass = explode(' ', $removeClass);
+        }
+
+        $class = array_unique(array_diff($class, $removeClass));
+
+        $input = Hash::insert($input, $useIndex, $class);
+
+        return $input;
+    }
 }
--- a/src/View/StringTemplate.php
+++ b/src/View/StringTemplate.php
@@ -331,10 +331,10 @@ class StringTemplate
     /**
      * Adds a class and returns a unique list either in array or space separated
      *
-     * @param array|string $input The array or string to add the class to
-     * @param array|string $newClass the new class or classes to add
+     * @param string[]|string $input The array or string to add the class to
+     * @param string[]|string $newClass the new class or classes to add
      * @param string $useIndex if you are inputting an array with an element other than default of 'class'.
-     * @return array|string
+     * @return string[]|string
      */
     public function addClass($input, $newClass, $useIndex = 'class')
     {
@@ -373,10 +373,10 @@ class StringTemplate
     /**
      * Removes a class and returns a unique list either in array or space separated
      *
-     * @param array|string $input The array or string to remove the class from
-     * @param array|string $removeClass the new class or classes to remove
+     * @param string[]|string $input The array or string to remove the class from
+     * @param string[]|string $removeClass the new class or classes to remove
      * @param string $useIndex if you are inputting an array with an element other than default of 'class'.
-     * @return array|string
+     * @return string[]|string
      */
     public function removeClass($input, $removeClass, $useIndex = 'class')
     {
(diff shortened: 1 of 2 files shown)


dereuromark:
This requires some test cases.

src/View/StringTemplate.php
```diff
@@ -369,4 +369,46 @@ public function addClass($input, $newClass, $useIndex = 'class')
...
+     *
+     * @param array|string $input The array or string to remove the class from
+     * @param array|string $removeClass the new class or classes to remove
+     * @param string $useIndex if you are inputting an array with an element other than default of 'class'.
+     * @return array|string
+     */
+    public function removeClass($input, $removeClass, $useIndex = 'class')
+    {
+        // NOOP
+        if (empty($removeClass)) {
```
markstory: What if the class is `'0'`?

ADmad: CSS identifiers can't start with a digit, so I don't think it's a problem.

https://www.w3.org/TR/CSS21/syndata.html#characters

src/View/StringTemplate.php
```diff
@@ -369,4 +369,46 @@ public function addClass($input, $newClass, $useIndex = 'class')
 
         return $input;
     }
+
+    /**
+     * Removes a class and returns a unique list either in array or space separated
+     *
+     * @param array|string $input The array or string to remove the class from
+     * @param array|string $removeClass the new class or classes to remove
```
dereuromark: array could be string[] here, right?

src/View/StringTemplate.php
```diff
@@ -369,4 +369,46 @@ public function addClass($input, $newClass, $useIndex = 'class')
...
+     * @param string $useIndex if you are inputting an array with an element other than default of 'class'.
+     * @return array|string
```
dereuromark: Same here probably.

groovenectar:
> This requires some test cases.

That makes sense! I overlooked this. Hopefully will be able to get those in there within a week

src/View/StringTemplate.php
groovenectar: Yep, I think if it's an array, it's supposed to only be string[].. The `addClass()` one seems pretty flexible and intuitive, and `removeClass()` is using the same methodology

groovenectar:
This is my first PR for Cake that is an update to something testable

src/View/StringTemplate.php
groovenectar: Just verified that that it does work like `addClass()` in that it supports both string and string[], I do need to add official test cases though

src/View/StringTemplate.php
groovenectar: These would be the same with `addClass()`

When I get back to this, should I change those to:

`* @param array string[]|string`

or

`* @param string[]|string`

Or is there a particular format?

ADmad: `@param string[]|string`

groovenectar: This is updated, and I'll work on getting tests up and running ASAP

dereuromark:
ping @groovenectar 
Are you able to provide some test cases for us here? So we can merge this?

groovenectar:
Sorry for my delay on this @markstory. I just now looked for tests on `addClass()` to use as an example, and it seems there aren't any tests for that either unless I'm missing something...

dereuromark:
It is possible, best to add some for your new method.

tests/TestCase/View/StringTemplateTest.php
```diff
@@ -440,4 +443,98 @@ public function testAddClassMethodUseIndex()
...
+        $result = $this->template->removeClass('keep_class remove_class', 'remove_class');
+        $this->assertEquals($result, ['class' => ['keep_class']]);
+
+        $result = $this->template->removeClass(null, 'remove_class');
+        $this->assertEquals($result, ['class' => []]);
+
+        $result = $this->template->removeClass(false, 'remove_class');
+        $this->assertEquals($result, ['class' => []]);
+
+        $result = $this->template->removeClass(new \StdClass(), 'remove_class');
```
dereuromark: stdClass()

groovenectar:
@dereuromark I found the tests, and have added some for the `removeClass()` method. There may be updates to the `addClass()` method upstream by now. I adapted the tests from `addClass()`. The main one I'm questioning is the last `assertEquals` test in the file... Normally I don't change that `$useIndex` parameter in practice...

Also, the original pull request may not be very DRY... I was thinking about adding another method to separate out the repeated parts of the two methods...

tests/TestCase/View/StringTemplateTest.php
groovenectar: @dereuromark, I updated this for both `addClass()` and `removeClass()`

othercorey:
@groovenectar Did you delete the fork?

groovenectar:
> @groovenectar Did you delete the fork?

Sorry 😬 I had to make another fork, and deleted the first one to do it

I guess it makes sense that it would yank it from PRs after that

I think that my tests were not super polished or comprehensive anyway for this, and the code could have been more DRY

I'm sure that I can find the method again or re-implement it. Until then, if it's gone I guess it would make sense to just close it. Whoops 😑

groovenectar:
This was to go into Cake3, and now I'm setup to contribute to Cake4 anyway, so if I can re-implement or find it I'll make the request for version 4 instead

othercorey:
You can view the files changed on this PR to see your work.

groovenectar:
> You can view the files changed on this PR to see your work.

Ah, nice! That will help, thanks. Would it make more sense if I redo it for the latest 4.x branch?

othercorey:
Yes. You can add it to 3.next or 4.x.  The 4.x branch is best and then can backport if needed.

PR #13382, opened 2019-06-28
4.x - Update tests to not use deprecated PHPUnit features

markstory:
Fix the various deprecation warnings we're getting from PHPUnit about
methods that are now deprecated.

### DIFF
--- a/tests/TestCase/Database/QueryTest.php
+++ b/tests/TestCase/Database/QueryTest.php
@@ -28,6 +28,7 @@ use Cake\Datasource\ConnectionManager;
 use Cake\TestSuite\TestCase;
 use DateTimeImmutable;
 use InvalidArgumentException;
+use ReflectionProperty;
 use TestApp\Database\Type\BarType;
 
 /**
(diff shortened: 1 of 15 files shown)


tests/TestCase/ORM/TableTest.php
```diff
@@ -2997,33 +2997,30 @@ public function testDeleteCallbacks()
...
             ->method('dispatch')
-            ->with($this->logicalAnd(
-                $this->attributeEqualTo('_name', 'Model.beforeDelete'),
-                $this->attributeEqualTo(
-                    '_data',
-                    ['entity' => $entity, 'options' => $options]
-                )
-            ));
+            ->with($this->callback(function (EventInterface $event) use ($entity, $options) {
+                return (
```
ADmad: phpcs doesn't like the use of parentheses for `return`.

https://github.com/cakephp/cakephp/pull/13382/checks?check_run_id=157743076

tests/TestCase/Database/QueryTest.php
```diff
@@ -2321,11 +2322,13 @@ public function testSelectOffset()
             ->limit(1)
             ->offset(1)
             ->execute();
-        $dirty = $this->readAttribute($query, '_dirty');
-        $this->assertFalse($dirty);
+
+        $reflect = new ReflectionProperty($query, '_dirty');
+        $reflect->setAccessible(true);
+        $this->assertFalse($reflect->getvalue($query));
```
garas: `ReflectionProperty::getValue()` here and few lines below

PR #13383, opened 2019-06-28
Fix error reported by psalm.

ADmad:
I have suppressed errors related to missing typehints for closure arguments and return types. It would be too much work to add typehints to closures with limited benefit.


psalm.xml
```diff
@@ -22,6 +22,8 @@
         <RedundantConditionGivenDocblockType errorLevel="info" />
         <TypeCoercion errorLevel="info" />
         <DocblockTypeContradiction errorLevel="info" />
+        <MissingClosureParamType errorLevel="info" />
```
ravage84: Good riddance of the above :+1:

ravage84 (approved):

PR #13384, opened 2019-06-28
4.0 make configApplication

saeideng:
will use in https://github.com/cakephp/cakephp/pull/13362

### DIFF
--- a/src/TestSuite/ConsoleIntegrationTestTrait.php
+++ b/src/TestSuite/ConsoleIntegrationTestTrait.php
@@ -43,6 +43,20 @@ trait ConsoleIntegrationTestTrait
     protected $_useCommandRunner = false;
 
     /**
+     * The customized application class name.
+     *
+     * @var string|null
+     */
+    protected $_appClass;
+
+    /**
+     * The customized application constructor arguments.
+     *
+     * @var array|null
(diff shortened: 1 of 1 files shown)


src/TestSuite/ConsoleIntegrationTestTrait.php
```diff
@@ -42,6 +42,20 @@ trait ConsoleIntegrationTestTrait
...
+     * The customized application class name.
+     *
+     * @var string|null
+     */
+    protected $_appClass;
+
+    /**
+     * The customized application constructor arguments.
+     *
+     * @var array|null
```
dereuromark: If you dont accept null strictly (empty array behaves the same), then you can remove nullable and just use empty array as default usage below.

saeideng: we accept `null` just like `IntegrationTestTrait`'s configApplication
https://github.com/cakephp/cakephp/blob/85ad45df8fa691058de7dfd59972f9e16e34f1c1/src/TestSuite/IntegrationTestTrait.php#L90-L92
both `null` and `[]` are same for this method yes , and we can accept only array for both methods in this trait and `IntegrationTestTrait`

PR #13386, opened 2019-06-29
Remove `checkboxFormGroup` key

kicaj:
Remove `checkboxFormGroup` key from form templates. It's not used anywhere.


saeideng:
it uses in
https://github.com/cakephp/cakephp/blob/2e19b7852c1fe997b26c8ecd548a141c712d667f/src/View/Helper/FormHelper.php#L1269
when `$options['options']['type']` == `checkbox`

PR #13394, opened 2019-07-03
4.x - Use Throwable on Connection.php

albertcansado:


ravage84 (approved):
May be additional tests covering the extended catches would be nice.

Issue #13067, opened 2019-03-22
RFC: Plugin/App path methods need adjusting


dereuromark:
This is a (multiple allowed):

* [] bug
* [x] enhancement
* [x] feature-discussion (RFC)

* CakePHP Version: 3.x

### What you did
With the change from 2.x to 3.x and plugins not required to be loaded anymore due to direct autoload-inclusion, we have an issue with Plugin::path() and App::path() to work in those cases.

Loading a plugin is usually only required for certain types of classes contained.
So often you dont "load" it explicitly.

My concrete issue:
The IdeHelper is supposed to run on both loaded and not loaded plugins.
But with the not loaded ones this does not work.

### What happened
```
bin/cake annotations all -p Markup
```
Exception: Plugin not found even though it is there.

It required `Plugin::load('Markup');` in bootstrap or the Application version of it.
Even though none of the classes require this as they are already in the autoload map.

### What you expected to happen
To still be able to use those methods and find the right paths.

So, are we able to provide a way in core to support those methods above also for those use cases?
My current "workaround" is custom wrappers ( https://github.com/dereuromark/cakephp-ide-helper/compare/plugin-path ), but I am not sure this is a good idea, especially since those methods are internal/deprecated and won't work beyond 3.x.

markstory:
We could try to 'autoload' plugins if they have not already been explicitly loaded.

dereuromark:
Yeah this is basically the current workaround: https://github.com/dereuromark/cakephp-ide-helper/compare/plugin-path#diff-fec1982b7a04c04113b90e7913c6f84fR24

dereuromark:
Until we "fix" this in core I released https://github.com/dereuromark/cakephp-ide-helper/releases/tag/0.13.9 with shimming in place.

ADmad:
Everything installed plugin in listed in `vendor/cakephp-plugins.php` generated by the plugin installer. So that could be leveraged if a plugin isn't explicitly loaded and added to plugin collection.

PR #13400, opened 2019-07-06
Try autoloading plugin in PluginCollection::get().

ADmad:
Refs #13067.

### DIFF
--- a/src/Core/PluginCollection.php
+++ b/src/Core/PluginCollection.php
@@ -198,7 +198,10 @@ class PluginCollection implements Iterator, Countable
     }
 
     /**
-     * Get the a plugin by name
+     * Get the a plugin by name.
+     *
+     * If a plugin isn't already loaded it will be autoloaded on first access
+     * and that plugins loaded this way may miss some hook methods.
      *
      * @param string $name The plugin to get.
      * @return \Cake\Core\PluginInterface The plugin.
--- a/src/Core/PluginCollection.php
+++ b/src/Core/PluginCollection.php
@@ -210,20 +210,38 @@ class PluginCollection implements Iterator, Countable
             return $this->plugins[$name];
         }
 
-        $config = ['name' => $name];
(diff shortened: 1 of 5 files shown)


src/Core/PluginCollection.php
```diff
@@ -206,11 +206,42 @@ public function has(string $name): bool
...
     {
-        if (!$this->has($name)) {
-            throw new MissingPluginException(['plugin' => $name]);
+        if ($this->has($name)) {
+            return $this->plugins[$name];
         }
 
-        return $this->plugins[$name];
+        $plugin = $this->create($name);
+        $this->add($plugin);
```
markstory: The docblock should mention that if the plugin hasn't already been loaded it will be loaded on first access and that plugins loaded this way may miss some hook methods.

src/Http/BaseApplication.php
```diff
@@ -96,16 +93,10 @@ public function pluginMiddleware(MiddlewareQueue $middleware): MiddlewareQueue
     public function addPlugin($name, array $config = [])
     {
         if (is_string($name)) {
-            $plugin = $this->makePlugin($name, $config);
+            $plugin = $this->plugins->create($name, $config);
```
markstory: This is a nice cleanup of the Application methods :+1:

src/Core/PluginCollection.php
markstory: Should `Plugin::loaded()` also autoload any installed plugins?

ADmad: We can add an argument to `loaded()` to autoload available plugins but it shouldn't be default.

Autoloading a plugin could have unknown and undesirable effects on the applications working.

markstory (approved):
Looks good to me, some builds are failing though.

ADmad:
The failures were unrelated. I haved rebased the branch on top of `4.x` which should fix the builds.

PR #13402, opened 2019-07-07
cakephp/utility should suggest ext-simplexml

brandonkelly:
The Xml utility relies on the SimpleXMLElement class existing, which is only available if the SimpleXML extension is installed. Therefore the package should be requiring that `ext-simplexml` is present.

### DIFF
--- a/src/Utility/composer.json
+++ b/src/Utility/composer.json
@@ -26,12 +26,12 @@
     },
     "require": {
         "php": ">=5.6.0",
-        "ext-simplexml": "*",
         "cakephp/core": "^3.6.0"
     },
     "suggest": {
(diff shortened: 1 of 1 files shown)


garas:
I would prefer it in `suggest` section like for `ext-intl`.

It is only required when using `Xml` utility.

Some users have difficulties with enabling extensions, so it is better soft dependency.

`Security` also requires `ext-openssl`. There is suggestion in root `composer.json`, but it could be copied to this one too.

dereuromark:
Only parts of it require it as such suggest seems better here.

brandonkelly:
OK, I’ve switched it to a suggestion.

PR #13385, opened 2019-06-29
4.x - Add suggestions for missing command name

markstory:
When a command is run and it doesn't exist, we can suggest potential commands based on edit distances and prefix matching.

If folks like this approach I can extend it to missing options as well. I'm not really happy with the name `NoOptionException` and welcome any suggestions for a better name.

PR #13405, opened 2019-07-08
[REDIS] Use SCAN instead of KEYS to find all matching keys

### LINKED ISSUE #13404: [Redis] Optimize Cache::clear by using SCAN instead of KEYS
* [ ] bug
* [x] enhancement

* CakePHP Version: 3

Currently the RedisEngine uses the `KEYS` command to clear all cache objects.
This is not efficient, according to the Redis docs: https://redis.io/commands/keys

They propose to use the `SCAN` command to find all matching keys.

annuh:
This PR changes the way how cache is cleared. First the inefficient KEYS command was used to find all matching keys in Redis.

According to the [Redis docs](https://redis.io/commands/keys), this doesn't perform well and they advice to use the `SCAN` command.

> Warning: consider KEYS as a command that should only be used in production environments with extreme care. It may ruin performance when it is executed against large databases. This command is intended for debugging and special operations, such as changing your keyspace layout. Don't use KEYS in your regular application code. If you're looking for a way to find keys in a subset of your keyspace, consider using SCAN or sets.

Fixes: https://github.com/cakephp/cakephp/issues/13404

Used sample code from PhpRedis https://github.com/phpredis/phpredis#scan

### DIFF
--- a/src/Cache/Engine/RedisEngine.php
+++ b/src/Cache/Engine/RedisEngine.php
@@ -236,7 +236,7 @@ class RedisEngine extends CacheEngine
         $pattern = $this->_config['prefix'] . '*';
 
         while (true) {
-            $keys = $this->_Redis->scan($iterator, $pattern));
+            $keys = $this->_Redis->scan($iterator, $pattern);
 
             if ($keys === false) {
                 break;
--- a/src/Cache/Engine/RedisEngine.php
+++ b/src/Cache/Engine/RedisEngine.php
@@ -235,7 +235,13 @@ class RedisEngine extends CacheEngine
         $iterator = null;
         $pattern = $this->_config['prefix'] . '*';
 
-        while ($keys = $this->_Redis->scan($iterator, $pattern)) {
+        while (true) {
+            $keys = $this->_Redis->scan($iterator, $pattern));
+
+            if ($keys === false) {
+                break;
+            }
+
             foreach ($keys as $key) {
                 $isDeleted = ($this->_Redis->del($key) > 0);
                 $isAllDeleted = $isAllDeleted && $isDeleted;
--- a/src/Cache/Engine/RedisEngine.php
+++ b/src/Cache/Engine/RedisEngine.php
@@ -229,7 +229,7 @@ class RedisEngine extends CacheEngine
             return true;
         }
 
-        $result = true;
+        $isAllDeleted = true;
         $iterator = null;
         $pattern = $this->_config['prefix'] . '*';
         do {
@@ -238,12 +238,13 @@ class RedisEngine extends CacheEngine
             // Redis may return empty results, so protect against that
             if ($keys !== false) {
                 foreach($keys as $key) {
-                    $result = $result && ($this->_Redis->del($key) > 0);
+                    $isDeleted = ($this->_Redis->del($key) > 0);
+                    $isAllDeleted = $isAllDeleted && $isDeleted;
                 }
             }
         } while ($iterator > 0);
 
-        return $result;
+        return $isAllDeleted;
     }
 
     /**
--- a/src/Cache/Engine/RedisEngine.php
+++ b/src/Cache/Engine/RedisEngine.php
@@ -237,7 +237,7 @@ class RedisEngine extends CacheEngine
 
             // Redis may return empty results, so protect against that
             if ($keys !== false) {
-                foreach($keys as $key) {
+                foreach ($keys as $key) {
                     $isDeleted = ($this->_Redis->del($key) > 0);
                     $isAllDeleted = $isAllDeleted && $isDeleted;
                 }
--- a/src/Cache/Engine/RedisEngine.php
+++ b/src/Cache/Engine/RedisEngine.php
@@ -229,6 +229,7 @@ class RedisEngine extends CacheEngine
             return true;
         }
 
+        $result = true;
         $iterator = null;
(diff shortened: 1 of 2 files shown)


src/Cache/Engine/RedisEngine.php
```diff
@@ -228,14 +228,22 @@ public function clear($check)
...
+        $result = true;
+        $iterator = null;
+        $pattern = $this->_config['prefix'] . '*';
+        do {
+            $keys = $this->_Redis->scan($iterator, $pattern);
+
+            // Redis may return empty results, so protect against that
+            if ($keys !== false) {
+                foreach($keys as $key) {
+                    $result = $result && ($this->_Redis->del($key) > 0);
```
garas: After `$result` becomes `false`, `del()` won't be called.

Either return `false` immediately (but this would be different behavior), or use something similar to previous code.

annuh: Ah stupid, thanks for the feedback. I'll fix this, so this behavior doesn't change.

src/Cache/Engine/RedisEngine.php
```diff
@@ -228,14 +228,23 @@ public function clear($check)
...
-            $result[] = $this->_Redis->del($key) > 0;
-        }
+        $isAllDeleted = true;
+        $iterator = null;
+        $pattern = $this->_config['prefix'] . '*';
+        do {
+            $keys = $this->_Redis->scan($iterator, $pattern);
+
+            // Redis may return empty results, so protect against that
+            if ($keys !== false) {
```
markstory: Is there a scenario where `$keys == false` and `$iterator > 0`? It looks like you have a potential infinite loop. You could do something like

```php
if ($keys === false || $iterator === 0) {
  break;
}
```

to avoid this scenario.

annuh: According to the PHPRedis [documentation](https://github.com/phpredis/phpredis#example-6) this should not be possible.
`$iterator` loops over _all_ keys, while `$keys` only contain results that matches the pattern, within the current iterator window.

I could simplify this method by using the `Redis::SCAN_RETRY` option if you like, because currently it does indeed look a bit strange.

markstory: Using SCAN_RETRY sounds like a good idea to me.

annuh:
I needed to remove the `composer.lock` file in order to fix issues with Appveyor (https://github.com/cakephp/cakephp/pull/13385)

src/Cache/Engine/RedisEngine.php
```diff
@@ -228,14 +228,21 @@ public function clear($check)
...
-        foreach ($keys as $key) {
-            $result[] = $this->_Redis->del($key) > 0;
+        $this->_Redis->setOption(Redis::OPT_SCAN, Redis::SCAN_RETRY);
+
+        $isAllDeleted = true;
+        $iterator = null;
+        $pattern = $this->_config['prefix'] . '*';
+
+        while ($keys = $this->_Redis->scan($iterator, $pattern)) {
+            foreach ($keys as $key) {
```
markstory: You could solve the warning from scrutinizer by doing

```php
while (true) {
  $keys = $this->_Redis->scan($iterator, $pattern)
  if ($keys === false) {
    break;
  }
  foreach ($keys as $key) {
  }
```

annuh: Normally I'm not a fan of `while(true)` loops, but we should do everything we can to make Scrutinizer happy, right? ;-)

src/Cache/Engine/RedisEngine.php
```diff
@@ -228,14 +228,27 @@ public function clear($check)
         if ($check) {
             return true;
         }
-        $keys = $this->_Redis->getKeys($this->_config['prefix'] . '*');
 
-        $result = [];
-        foreach ($keys as $key) {
-            $result[] = $this->_Redis->del($key) > 0;
+        $this->_Redis->setOption(Redis::OPT_SCAN, Redis::SCAN_RETRY);
```
ADmad: This is causing type errors in 4.x  https://travis-ci.org/cakephp/cakephp/jobs/567855601#L822.
I am not sure what version of redis/predis travis is using. But even it only occurs for older version it could be an issue for our users.

ADmad: So the issue has been https://github.com/phpredis/phpredis/pull/1540. Not sure in which version.

ADmad: Fix is available in 5.0.0. So essentially one won't be able to used Redis 4.x with Cake 4.0.

garas: Do `(string)Redis::SCAN_RETRY` typecast would help?

josegonzalez: > So essentially one won't be able to use Redis 4.x with Cake 4.0.

You mean the client library cannot be 4.x, not that you cannot use Redis 4.x.

ADmad: @garas Maybe, I haven't checked.

@josegonzalez Yes, I should have worded it better.

ADmad: Typecasting to string worked https://github.com/cakephp/cakephp/pull/13469/files#diff-61e014e894b3dcbac9bd2425ec47490bR237

PR #13411, opened 2019-07-11
Enable autoFields in order to resolve lazy loading translated entities.

### LINKED ISSUE #13181: Translate behavior breaks loadInto for belongsTo associations

robertpustulka:
This resolves #13181, the fix also resolves a failing test case #13182

### DIFF
--- a/tests/TestCase/ORM/Behavior/TranslateBehaviorTest.php
+++ b/tests/TestCase/ORM/Behavior/TranslateBehaviorTest.php
@@ -780,6 +780,29 @@ class TranslateBehaviorTest extends TestCase
     }
 
     /**
+     * Tests that it is possible to translate belongsTo associations using loadInto
+     *
+     * @return void
+     */
+    public function testFindSingleLocaleBelongstoLoadInto()
+    {
+        $table = $this->getTableLocator()->get('Articles');
+        $table->addBehavior('Translate', ['fields' => ['title', 'body']]);
+        $authors = $table->belongsTo('Authors')->getTarget();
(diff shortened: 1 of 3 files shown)


ADmad (approved):

tests/TestCase/ORM/Behavior/TranslateBehaviorTest.php
```diff
@@ -779,6 +779,29 @@ public function testFindSingleLocaleBelongsto()
...
+        $table->addBehavior('Translate', ['fields' => ['title', 'body']]);
+        $authors = $table->belongsTo('Authors')->getTarget();
+        $authors->addBehavior('Translate', ['fields' => ['name']]);
+
+        $table->setLocale('eng');
+        $authors->setLocale('eng');
+
+        $entity = $table->get(1);
+        $result = $table->loadInto($entity, ['Authors']);
+        $this->assertSame($entity, $result);
```
markstory: Should there be assertions that the Authors + translation were loaded?

PR #13413, opened 2019-07-12
Update typehints.

ADmad:
There were a few methods which always returned `true` (or threw exception), I changed those to return `void`, for e.g. `Configure::write()`.

### DIFF
--- a/src/Console/ConsoleOptionParser.php
+++ b/src/Console/ConsoleOptionParser.php
@@ -886,17 +886,15 @@ class ConsoleOptionParser
         } else {
             $value = $option->defaultValue();
         }
-        if ($option->validChoice($value)) {
-            if ($option->acceptsMultiple()) {
-                $params[$name][] = $value;
-            } else {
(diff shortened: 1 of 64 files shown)


src/Console/ConsoleOptionParser.php
```diff
@@ -938,13 +936,10 @@ protected function _parseArg(string $argument, array $args): array
...
 
-        if ($this->_args[$next]->validChoice($argument)) {
-            $args[] = $argument;
-
-            return $args;
-        }
+        $this->_args[$next]->validChoice($argument);
+        $args[] = $argument;
 
-        throw new ConsoleException('Invalid choice for argument: ' . $argument);
```
markstory: Don't we need this exception to signal that an invalid choice was made for the option?

ADmad: The `ConsoleInputArgument::validChoice()` already throws an exception.

PR #13421, opened 2019-07-16
Fix response type documentation and better invalid response.

dereuromark:
I tried to get phpstan level 5 to pass for plugins, but the invalid doc blocks prevent this.
Also, we should be failing with better feedback on invalid arguments instead of silently passing "false" along, which then even silently create wrong cache invalidation timestamps on the response.

### DIFF
--- a/src/Http/Response.php
+++ b/src/Http/Response.php
@@ -1342,7 +1342,7 @@ class Response implements ResponseInterface
         if (!is_int($time)) {
             $time = strtotime($time);
             if ($time === false) {
-                throw new InvalidArgumentException('Time parameter invalid.');
+                throw new InvalidArgumentException('Invalid time parameter. Ensure your time value can be parsed by strtotime');
             }
         }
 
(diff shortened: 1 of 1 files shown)


src/Http/Response.php
```diff
@@ -1352,14 +1357,18 @@ public function cache($since, $time = '+1 day')
...
+     * @param string|int $time A valid time for cache expiry
      * @return static
+     * @throws \InvalidArgumentException
      */
     public function withCache($since, $time = '+1 day')
     {
         if (!is_int($time)) {
             $time = strtotime($time);
+            if ($time === false) {
+                throw new InvalidArgumentException('Time parameter invalid.');
```
markstory:
```suggestion
                throw new InvalidArgumentException('Invalid time parameter. Ensure your time value can be parsed by strtotime');
```

src/Http/Response.php
```diff
@@ -1339,6 +1341,9 @@ public function cache($since, $time = '+1 day')
 
         if (!is_int($time)) {
             $time = strtotime($time);
+            if ($time === false) {
+                throw new InvalidArgumentException('Time parameter invalid.');
```
markstory:
```suggestion
                throw new InvalidArgumentException('Invalid time parameter. Ensure your time value can be parsed by strtotime');
```

Issue #12954, opened 2019-02-06
Internationalization adds '_locale' property in contain associations


jcv-pt:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.5.17
* Platform and Target: Ubuntu 16.04, Apache, PHP 7.0.32

### What you did
Added the translate behaviour in the model, and changed my language to a non default via I18n::setLocale
$this->Model->find('all')->matching('SomeModel');

### What happened
The property "_locale" appears on the array containing the "contain" associations when using "matching" method in a model query.

Bellow is an excerpt of the query:

```php
[product_variations] => Array(
	[0] => App\Model\Entity\ProductsVariation Object(
		[id] => 15
		[product_id] => 12
		...
		[_locale] => pt
		...
		[[invalid]] => Array
			(
			)
		[[repository]] => ProductVariations
	),
	[_locale] => pt
)
```

### What you expected to happen
The property "_locale" should be present in the entity and not within the array of associations.

Stack Overflow link with isssue: [https://stackoverflow.com/questions/54543105/cakephp-internationalization-adds-locale-property-in-contain-associations/54543493#54543493](url)

ndm2:
```php
public function testLocalePropertyShouldNotBleedIntoNtoManyAssociationPropertiesWhenUsingMatching()
{
    $table = $this->getTableLocator()->get('Articles');
    $table->hasMany('Comments');

    $table->Comments->addBehavior('Translate', [
        'translationTable' => '\TestApp\Model\Table\I18nTable'
    ]);
    $table->Comments->setLocale('xyz');

    $this->assertNotEquals($table->Comments->getLocale(), I18n::getLocale());

    $result = $table
        ->find()
        ->contain('Comments')
        ->matching('Comments')
        ->first();

    $this->assertArrayNotHasKey('_locale', $result->comments);
}
```

The problem seems to stem from the fact that matching creates a joinable containment, which is being processed in `\Cake\ORM\Association::attachTo()` using the same property path/name as the regular, non-matching containment.

I'm not tooo familiar with the inner workings, but maybe for `matching` containments the `propertyPath` built in `\Cake\ORM\EagerLoader::_normalizeContain()` should point to `_matchingData.$alias` instead of the association property name?

https://github.com/cakephp/cakephp/blob/b0b9a3713720b30cd8e90c8b1a317a8294b2513a/src/ORM/EagerLoader.php#L560

PR #13422, opened 2019-07-16
Fix wrong association and _locale properties set when using matching.

ndm2:
When using matching, the translate behavior sets wrong association and `_locale` properties, because the eager loadables hold wrong property paths that point to the properties that are used for regular containments, instead of the `_matchingData` property where matching records are set.

I think that `EagerLoader::_normalizeContain()` is the right place to fix this, as it seems to be the place where the path is initially being created, and containments for matching only ever seem to be placed in the `_matchingData` property, but I have to admit that I'm not 100% sure.

refs #12954

### DIFF
--- a/tests/TestCase/ORM/Behavior/TranslateBehaviorTest.php
+++ b/tests/TestCase/ORM/Behavior/TranslateBehaviorTest.php
@@ -45,6 +45,7 @@ class TranslateBehaviorTest extends TestCase
      */
     public $fixtures = [
         'core.Articles',
+        'core.ArticlesTags',
         'core.Authors',
         'core.Groups',
         'core.SpecialTags',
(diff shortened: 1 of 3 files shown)


lorenzo:
This looks like the right solution to me, good job!

lorenzo (approved):

tests/TestCase/ORM/Behavior/TranslateBehaviorTest.php
```diff
@@ -1856,4 +1857,178 @@ public function testCustomTableLocator()
...
+        $table->Comments->setLocale('abc');
+
+        $this->assertNotEquals($table->Comments->getLocale(), I18n::getLocale());
+
+        $result = $table
+            ->find()
+            ->matching('Comments')
+            ->first();
+
+        $this->assertArrayNotHasKey('comments', $result->toArray());
```
ADmad: This assertion is invalid. There are no `fields` specified when attaching the translate behavior so it essentially does nothing. 

If you specify `fields` the behavior will generate a contain internally and you will get `comments` property in result even though the "user" query only does a matching.

PR #13428, opened 2019-07-18
Improve error message.

ADmad:


dereuromark (approved):
Personally, I like code fencing in exception messages also with markdown backticks `this` as this is also nice to copy and paste. But looks good.

PR #13429, opened 2019-07-18
bugfix: Query orderBy control for deprecation warning

tedicela:
How to reproduce the bug:
----------------------------------

Using PostgreSQL database your should be able to make a query like this below:

```
$this->Users->find()->order(['email'=>'ASC NULLS LAST'])->all();
```

This throws a Deprecation warning:

```
Passing extra sort expressions by associative array is deprecated. Use QueryExpression or numeric array instead.
```

This PR fix this bug!


garas:
You can use `order(['email ASC NULLS LAST'])` for any complex expression.

Splitting on space without checking right side would reopen https://github.com/cakephp/cakephp/issues/11148 as it would allow any expression after space. 

4.x will throw exception in this case:

https://github.com/cakephp/cakephp/blob/15da42fdedfefc612b886293a8807acdd9db47c2/src/Database/Expression/OrderByExpression.php#L69-L86

markstory:
Thanks for taking the time to put this together, but this change would reopen a previously fixed issue. I recommend you follow the deprecation warning instead.

PR #13432, opened 2019-07-23
Set expiration atomically in RedisEngine::add()

ethanpooley:
My company has observed occasions where the `RedisEngine::add()` method sets a key but fails to set its expiration. That is very bad for simple locking implementations because the lock will never expire. This PR uses newer (Redis version >= 2.6.12) options in `Redis::set()` to set the expiration atomically when the key is set.

### DIFF
--- a/src/Cache/Engine/RedisEngine.php
+++ b/src/Cache/Engine/RedisEngine.php
@@ -258,7 +258,7 @@ class RedisEngine extends CacheEngine
      * @param string $key Identifier for the data.
      * @param mixed $value Data to be cached.
      * @return bool True if the data was successfully cached, false on failure.
-     * @link https://github.com/phpredis/phpredis#setnx
+     * @link https://github.com/phpredis/phpredis#set
      */
     public function add($key, $value)
     {
@@ -269,9 +269,8 @@ class RedisEngine extends CacheEngine
             $value = serialize($value);
         }
 
-        // setNx() doesn't have an expiry option, so follow up with an expiry
-        if ($this->_Redis->setNx($key, $value)) {
-            return $this->_Redis->expire($key, $duration);
+        if ($this->_Redis->set($key, $value, ['nx', 'ex' => $duration])) {
+            return true;
         }
 
         return false;


src/Cache/Engine/RedisEngine.php
```diff
@@ -269,9 +269,8 @@ public function add($key, $value)
             $value = serialize($value);
         }
 
-        // setNx() doesn't have an expiry option, so follow up with an expiry
-        if ($this->_Redis->setNx($key, $value)) {
-            return $this->_Redis->expire($key, $duration);
+        if ($this->_Redis->set($key, $value, ['nx', 'ex' => $duration])) {
```
markstory: What will happen for older versions of redis?

ADmad: @ethanpooley Ping

ethanpooley: Gah, sorry. Yes. I have no idea what would happen for older versions. And perhaps I need to stop proposing things if I'm not going to stick with them! Are we checking versions anywhere else, and treating different ones differently, within the Redis engine? If so we could simply version this feature and use the old method for pre-2.6.12.

Alternatively, it seems we should be aware of a minimum version number of Redis that we actually intend to support, and support elsewhere in the engine. 2.6.12 was [released six and a half years ago](https://github.com/antirez/redis/releases/tag/2.6.12), and the version of 2.x that seemed to have much of a legacy life beyond that point was 2.8. Do we have an explicit version support target for our Redis engine?

ethanpooley:
I should add that my company is now using the new approach in production (bypassing `\Cake\Cache\Cache` and using native Redis, for now) and it is behaving as expected for us.

src/Cache/Engine/RedisEngine.php
markstory: No specifically. This is the first time we have added functionality that depends on 'new' behavior in redis. It looks like ubuntu 18.04 comes with [redis 4.6](https://launchpad.net/ubuntu/bionic/+source/redis) which should make this relatively safe.

PR #13433, opened 2019-07-23
Run testsuite on PHP 7.4.

ADmad:
Who wants to track down segfaults? :smiley:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ php:
   - 5.6
   - 7.2
   - 7.3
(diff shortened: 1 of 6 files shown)


markstory:
Booooo segfaults are very tedious to track down.

ADmad:
Can we merge this? The changes are good for all currently supported PHP versions and 7.4.

We can try figuring out the segfaults on PHP 7.4 later. Or maybe it will resolve itself with never builds of 7.4 :slightly_smiling_face:.

PR #13435, opened 2019-07-23
Added Char columns to reflect as char instead of string

karmicdice:
Refers to : https://github.com/cakephp/migrations/issues/362

Added 'TYPE_CHAR' in TableSchemaInterface, MySqlSchema and corrected test to reflect the change.

### DIFF
--- a/src/Database/Schema/TableSchemaInterface.php
+++ b/src/Database/Schema/TableSchemaInterface.php
@@ -79,6 +79,13 @@ interface TableSchemaInterface extends SchemaInterface
     const TYPE_STRING = 'string';
 
     /**
+     * String column type
+     *
+     * @var string
+     */
+    const TYPE_CHAR = 'char';
+
+    /**
      * Text column type
      *
      * @var string
--- a/src/Database/Schema/MysqlSchema.php
+++ b/src/Database/Schema/MysqlSchema.php
@@ -123,7 +123,7 @@ class MysqlSchema extends BaseSchema
             return ['type' => TableSchema::TYPE_UUID, 'length' => null];
         }
         if ($col === 'char') {
-            return ['type' => TableSchema::TYPE_STRING, 'fixed' => true, 'length' => $length];
+            return ['type' => TableSchema::TYPE_CHAR, 'fixed' => true, 'length' => $length];
         }
         if (strpos($col, 'char') !== false) {
             return ['type' => TableSchema::TYPE_STRING, 'length' => $length];
--- a/src/Database/Type.php
+++ b/src/Database/Type.php
@@ -49,6 +49,7 @@ class Type implements TypeInterface
         'time' => 'Cake\Database\Type\TimeType',
         'timestamp' => 'Cake\Database\Type\DateTimeType',
         'uuid' => 'Cake\Database\Type\UuidType',
+        'char' => 'Cake\Database\Type\CharType',
     ];
 
     /**
--- a/src/Database/Type/CharType.php
+++ b/src/Database/Type/CharType.php
@@ -111,4 +111,3 @@ class CharType extends Type implements OptionalConvertInterface, TypeInterface
         return false;
     }
 }
-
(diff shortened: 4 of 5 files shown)


src/Database/Schema/TableSchemaInterface.php
```diff
@@ -78,6 +78,13 @@ interface TableSchemaInterface extends SchemaInterface
      */
     const TYPE_STRING = 'string';
 
+    /**
+     * String column type
+     *
+     * @var string
+     */
+    const TYPE_CHAR = 'char';
```
markstory: This new type doesn't have a TypeClass or a mapping to make it the same as `string`. The `char` type will also need SQL generation code in all of the database adapter layers or fixture schema will not be generated using `char` columns when the fixture has `type => char`.

markstory (changes requested):
This change is missing a few important pieces and because it is adding new behavior (and potentially breaking existing code) it cannot be merged into master, perhaps this should go into 3.next or 4.x instead?

src/Database/Schema/TableSchemaInterface.php
karmicdice: Oh damn! My bad. Sorry, yea should have checked that bit. It would result in breaking as 'Unknown type'.
Got it... On it...

karmicdice:
> This change is missing a few important pieces and because it is adding new behavior (and potentially breaking existing code) it cannot be merged into master, perhaps this should go into 3.next of 4.x instead?

Sure. Which branch do you want me to push at? Just saw the comment, have added CharType which returns as string. Under license comment, I have put since: 3.8.1, but will change to the branch you recommend.

saeideng:
you should add some tests

saeideng:
>Which branch do you want me to push at?

Ideally new features in new version : 4.x

karmicdice:
> you should add some tests

Fixtures? Cool...

ADmad:
@karmicdice You can make a new feature branch off `4.x` and make all your changes on that.

karmicdice:
> This change is missing a few important pieces and because it is adding new behavior (and potentially breaking existing code) it cannot be merged into master, perhaps this should go into 3.next of 4.x instead?

Will resend PR at 4.x instead

saeideng:
> Fixtures? Cool...

for new  type class  :  see `tests/TestCase/Database/Type/`
for other changes :  related testCases

and if needed you can add new fixture

karmicdice:
> @karmicdice You can make a new feature branch off `4.x` and make all your changes on that.

Ok, will do 4.x-feature

karmicdice:
> > Fixtures? Cool...
> 
> for new type class : see `tests/TestCase/Database/Type/`
> for other changes : related testCases
> 
> and if needed you can add new fixture

Oh! okay! gotcha! Yes, fixtures will useful because I reckon this is a breaking change for existing setup

karmicdice:
Going to push new branch for 4.x-feature
Also...
There are few more tickets related to this in [Migrations](https://github.com/cakephp/migrations/issues/). longblob, snapshot failing to calculate length, etc.

Do you think we ought to support more column types? The logic is _sortof_ similar.

Please let me know.

markstory:
Historically we've limited how many data types that are only provided by a single vendor we support. MySQL, Postgres, and SQLServer all have very deep catalogs of datatypes and I don't think we can support them all while also providing a 'cross-platform abstraction layer'. Finding a reasonable balance has been our current goal.

karmicdice:
> Historically we've limited how many data types that are only provided by a single vendor we support. MySQL, Postgres, and SQLServer all have very deep catalogs of datatypes and I don't think we can support them all while also providing a 'cross-platform abstraction layer'. Finding a reasonable balance has been our current goal.

Yea, did my research and thought why you had the limitations, totally justified and perhaps the right way to go. I will only go ahead with the char then because char is common, not really as required as longblob (from arch pov having a longblob vs fs is obnoxious design but... who cares haha)...
I will make the tests and push it to another branch across this weekend.

karmicdice:
Have pushed it to 4x-feature in my forked repo. Cannot find 4x-feature branch. Should I send PR at 4x?

ADmad:
> Should I send PR at 4x?

Yes

ADmad:
Closing as this is going to be targeted to 4.x, not 3.x.

PR #10154, opened 2017-02-02
Throw an exception if association's target table has an incorrect class name.

robertpustulka:
See a [comment](https://github.com/cakephp/cakephp/pull/10142#issuecomment-276918905)

PR #13436, opened 2019-07-24
Ensure Association::$_className is always set.

ADmad:
I don't see a reason why currently `$_className` is set to alias only when alias has plugin prefix.

### DIFF
--- a/src/ORM/Association.php
+++ b/src/ORM/Association.php
@@ -226,7 +226,7 @@ abstract class Association
             }
         }
 
-        if (empty($this->_className) && strpos($alias, '.')) {
+        if (empty($this->_className)) {
             $this->_className = $alias;
         }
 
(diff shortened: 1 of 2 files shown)


src/ORM/Association.php
```diff
@@ -226,7 +226,7 @@ public function __construct(string $alias, array $options = [])
             }
         }
 
-        if (empty($this->_className) && strpos($alias, '.')) {
+        if (empty($this->_className)) {
             $this->_className = $alias;
```
ADmad: The `_getClassName()` function I removed the did camelize the alias but I have skipped it. Conventionally we use camel cased aliases for associations and if someone doesn't follow that convention then they should just explicitly set the `className` config.

markstory: This check was added in #6054 but there isn't much context on why. If all the tests pass I think we should move forward, these changes make the implementation simpler.

src/ORM/Association.php
```diff
@@ -1125,24 +1125,6 @@ protected function _extractFinder($finderData): array
...
-     */
-    protected function _getClassName(string $alias, array $options = []): string
-    {
-        if (empty($options['className'])) {
-            $options['className'] = Inflector::camelize($alias);
-        }
-
-        $className = App::className($options['className'], 'Model/Table', 'Table') ?: Table::class;
-
-        return ltrim($className, '\\');
```
markstory: This came in #10154 but I don't see its absence causing any tests to fail.

markstory (approved):

PR #13437, opened 2019-07-24
StaticConfigTrait::getConfig($key) return type

andrii-pukhalevych:
StaticConfigTrait::getConfig($key) may return int/string/float with custom configs

### DIFF
--- a/src/Core/StaticConfigTrait.php
+++ b/src/Core/StaticConfigTrait.php
@@ -110,7 +110,7 @@ trait StaticConfigTrait
      * Reads existing configuration.
      *
      * @param string $key The name of the configuration.
-     * @return array|null Array of configuration data.
+     * @return mixed Configuration data.
      */
     public static function getConfig($key)
     {
--- a/src/Utility/Hash.php
+++ b/src/Utility/Hash.php
@@ -928,8 +928,8 @@ class Hash
      *
      * ### Sort directions
      *
-     * - `asc` Sort ascending.
-     * - `desc` Sort descending.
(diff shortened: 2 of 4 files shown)


andrii-pukhalevych:
@dereuromark 
> `mixed` covers all types so it should be the only type mentioned now.

Should I leave only `mixed` or `mixed|null`. I see that in some places @return has all types:
[`EventInterface::getData($key = null)`](https://github.com/cakephp/cakephp/blob/8005880409fce6afebd03053a6060cfd8a3c91d1/src/Event/EventInterface.php#L79) has `@return array|mixed|null`

andrii-pukhalevych:
@dereuromark Is cakephp:3.next correct branch?

dereuromark:
For fixes in docs, usually master is correct.

andrii-pukhalevych:
@dereuromark In this case I want to have this update asap, because its not breaking but affects code analyzer functionality.  At 3.next is 3.9 right? 

Also AppVeyor [failed to run test](https://ci.appveyor.com/project/cakephp/cakephp/builds/26209564#L196):  "./composer.lock" does not contain valid JSON                           
How it can be resolved?

dereuromark:
:-1: I dont think this should have been merged. Not valid as is.

markstory:
Whoops, I'll revert this merge I didn't see that it pulled in other changes from 3.next along with it.

markstory:
Whoops, I'll revert this merge I didn't see that it pulled in other changes from 3.next along with it.

PR #13439, opened 2019-07-24
Expand typehints for getConfig()

markstory:
It can return any type so indicate that.

Refs #13437


dereuromark:
What about setter and other methods? Sounds a bit weird to basically make the documentation useless (mixed) here for setter only.

PR #13442, opened 2019-07-26
Improve dev error page styling

markstory:
* Reduce contrast of 'toggle vendor frames' button. It is not the
  primary action on the page.
* Add better separation between frame buttons.
* Remove decorative > it wasn't adding anything.
* Make file paths fit on a single line. Instead of wrapping use a text
  ellipsis to trim the left side off.
* Select the first stack frame by default so we have some useful content
  to show.

Screen Shot 2019-07-25 at 22 40 28


ADmad:
Nice!

cleptric:
Awesome 👏

PR #13445, opened 2019-07-27
Update docblock.

ADmad:
If table doesn't have field named "title" or "name" and no primary key
then Table::$_displayField property would be null.


ADmad:
Static analyzers are happy, hence merging.

PR #13446, opened 2019-07-29
2.x - Moves creation of ControllerTestDispatcher to its own protected method

garethellis36:
We've extended the core `Dispatcher` class in our application and we now have some incompatibilities with the test dispatcher. Moving the creation to a protected method will allow us to extend the test dispatcher and fix those incompatibilities.


markstory:
Looks good outside of a [style error](https://travis-ci.org/cakephp/cakephp/jobs/564892184).

PR #8878, opened 2016-05-23
[WIP] 3.x - Add a rule that checks link constraints.

ndm2:
This is a first implementation of #6327, for now I've just ported it from my plugin into the core and added additional support for `belongsToMany` associations and configurable error behavior - that means, the rule still includes the highly debatable exception mechanism.

Before eventually removing exceptions, and adding/shifting (more) tests into the rules checker integration test, I wanted to put this up for further feedback as is.

PR #13447, opened 2019-07-30
4.x - Link constraint rule

ndm2:
Revival of the long forgotten #8878, with simplified link counting, and without throwing exceptions for validation failures.

Sorry @lorenzo, still no error buffering for implicit save/delete operations in associations :pensive:

### DIFF
--- a/src/ORM/Rule/LinkConstraint.php
+++ b/src/ORM/Rule/LinkConstraint.php
@@ -1,0 +1,201 @@
+<?php
+declare(strict_types=1);
+
+/**
+ * CakePHP(tm) : Rapid Development Framework (https://cakephp.org)
+ * Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
+ *
+ * Licensed under The MIT License
+ * For full copyright and license information, please see the LICENSE.txt
+ * Redistributions of files must retain the above copyright notice.
+ *
+ * @copyright     Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
+ * @link          https://cakephp.org CakePHP(tm) Project
+ * @since         4.0.0
+ * @license       https://opensource.org/licenses/mit-license.php MIT License
+ */
+namespace Cake\ORM\Rule;
+
+use Cake\Datasource\EntityInterface;
+use Cake\ORM\Association;
+use Cake\ORM\Table;
+
+/**
+ * Checks whether links to a given association exist / do not exist.
+ */
+class LinkConstraint
+{
+    /**
+     * Status that requires a link to be present.
+     *
+     * @var string
+     */
+    public const STATUS_LINKED = 'linked';
+
+    /**
+     * Status that requires a link to not be present.
+     *
+     * @var string
+     */
+    public const STATUS_NOT_LINKED = 'notLinked';
+
+    /**
+     * The association that should be checked.
+     *
+     * @var \Cake\ORM\Association|string
+     */
+    protected $_association;
+
+    /**
+     * The link status that is required to be present in order for the check to succeed.
+     *
+     * @var string
+     */
+    protected $_requiredLinkState;
+
+    /**
+     * Constructor.
+     *
+     * @param \Cake\ORM\Association|string $association The alias of the association that should be checked.
+     * @param string $requiredLinkStatus The link status that is required to be present in order for the check to
+     *  succeed.
+     */
+    public function __construct($association, string $requiredLinkStatus)
+    {
+        if (
+            !is_string($association) &&
+            !($association instanceof Association)
+        ) {
+            throw new \InvalidArgumentException(sprintf(
+                'Argument 1 is expected to be of type `\Cake\ORM\Association|string`, `%s` given.',
+                getTypeName($association)
+            ));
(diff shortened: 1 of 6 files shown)


src/ORM/Rule/LinkConstraint.php
```diff
@@ -0,0 +1,193 @@
...
+ * Checks whether links to a given association exist / do not exist.
+ */
+class LinkConstraint
+{
+    /**
+     * Link status that requires a link to be present.
+     *
+     * @var string
+     */
+    public const LINK_STATUS_LINKED = 'linked';
```
dereuromark: `LINK_` prefix might not be necessary in scope of LinkConstraint class.

ndm2: Yeah, I'm not really sure what the conventions for that are... gotta take a look at the const usages in the core.

ndm2:
Looks like Stickler and PHPCS are using different comparison algorithms for `use` statement order checking, seems that Stickler uses standard string comparison, while the PHPCS sniff uses natural string comparison.

Not sure why this isn't triggered for me locally: https://scrutinizer-ci.com/g/cakephp/cakephp/inspections/37eff0e9-a3ad-42b5-82f0-6ead409b99c8... guess I can just add it to the ignored errors regardless.

ADmad:
@ndm2 Can you please rebase and resolve the conflicts. You shouldn't need to update the ` psalm-baseline.xml`. If pslam is reporting a false positive please use inline annotation like `/** @psalm-suppress Foo` to suppress the particular error type.

ndm2:
I don't know why GitHub says that the ` tests/TestCase/ORM/Rule/LinkConstraintTest.php` and `tests/TestCase/ORM/RulesCheckerIntegrationTest.php` files are empty, they were shown correctly just a moment ago, and then something happened.

Maybe it was just a coincidence that it happened just when it failed, but could this have something to do with Stickler?

src/ORM/Rule/LinkConstraint.php
```diff
@@ -0,0 +1,189 @@
...
+     * @param array $options Options passed from the rules checker.
+     * @return bool Whether the check was successful.
+     */
+    public function __invoke(EntityInterface $entity, array $options): bool
+    {
+        if (!isset($options['repository']) ||
+            !($options['repository'] instanceof Table)
+        ) {
+            throw new \InvalidArgumentException(
+                'Argument 2 is expected to have a `repository` key that holds an instance of `\Cake\ORM\Table`.'
```
dereuromark: You can use Table::class here, it will also remove the unnecessary \ leading slash then :)

ndm2: I could, but that's not the only exception message in this PR that is using a fully qualified name. Is there a specific convention?

src/ORM/Rule/LinkConstraint.php
```diff
@@ -0,0 +1,193 @@
...
+     */
+    public function __construct($association, string $requiredLinkStatus)
+    {
+        if (
+            !is_string($association) &&
+            !($association instanceof Association)
+        ) {
+            throw new \InvalidArgumentException(sprintf(
+                'Argument 1 is expected to be of type `\Cake\ORM\Association|string`, `%s` given.',
+                is_object($association) ? get_class($association) : gettype($association)
```
markstory: You can use `getTypeName()` now.

src/ORM/Rule/LinkConstraint.php
```diff
@@ -0,0 +1,193 @@
...
+     * @param \Cake\Datasource\EntityInterface $entity The entity involved in the operation.
+     * @return int The number of links.
+     */
+    protected function _countLinks(Association $association, EntityInterface $entity): int
+    {
+        $source = $association->getSource();
+
+        $primaryKey = $source->getPrimaryKey();
+        if (!$entity->has($primaryKey)) {
+            throw new \InvalidArgumentException('All primary key values are required.');
```
markstory: This message could use some more context. Where are the primary keys required?

src/ORM/RulesChecker.php
```diff
@@ -108,6 +110,137 @@ public function existsIn($field, $table, $message = null): RuleInvoker
...
+        } elseif (is_string($association)) {
+            $associationAlias = $association;
+
+            if ($errorField === null) {
+                $errorField = Inflector::underscore($association);
+            }
+        } else {
+            throw new \InvalidArgumentException(sprintf(
+                'Argument 1 is expected to be of type `\Cake\ORM\Association|string`, `%s` given.',
+                is_object($association) ? get_class($association) : gettype($association)
```
markstory:
```suggestion
                getTypeName($association)
```

src/ORM/Rule/LinkConstraint.php
dereuromark: No, only that we try to use the ::class name to allow the namespace to be pulled by itself.

dereuromark:
Small conflict here.

ADmad (approved):

ndm2:
Sorry for pushing new stuff after review, but this "wrongly" inferred association property really bugged me.

PR #13451, opened 2019-08-01
2.x Saving Acl permission via alias, not class name

ghost:
Currently the ACL permission saves the class name of the model used, instead of its alias.

### DIFF
--- a/lib/Cake/Model/Behavior/AclBehavior.php
+++ b/lib/Cake/Model/Behavior/AclBehavior.php
@@ -85,7 +85,7 @@ class AclBehavior extends ModelBehavior {
 			}
 		}
 		if (empty($ref)) {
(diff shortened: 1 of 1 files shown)


ADmad:
What issue are you having with current code?

dereuromark:
While I agree that this could help if you use the class under several aliases, this is a heavy BC break as this has been like this for years. Many people rely on this current behavior. As such a bugfix patch can't be applied here for this.

Also please note: 2.x is in maintenance mode, so no changes that are not critical anymore.

As such I would recommend just extending or copying the behavior on plugin or app level and adjusting it there for your needs.

Additionally, you can check if 3.x and 4.x are OK here or if they need adjustment in a new minor or better major here.

ae-bogdan:
I'll close this then.

PR #13453, opened 2019-08-01
Set default value for argument.

ADmad:
EntityTrait::setDirty() already has $isDirty argument as true by default.


dereuromark (approved):

dereuromark:
That didn't trigger a notice on the signature change? Probably stricter on implementation is OK it seems for PHP.

PR #13455, opened 2019-08-02
Split combined getter/setter EntityInterface::isNew().

### LINKED ISSUE #13454: Split combined getter/setter EntityInterface::isNew()
Should we add a new method `EntityInterface::setNew(bool $new)` and change `isNew()` to `EntityInterface::isNew(): bool`?

ADmad:
isNew() now only acts as getter and new method setNew() has been added as getter.

Closes #13454.

### DIFF
--- a/src/Datasource/EntityInterface.php
+++ b/src/Datasource/EntityInterface.php
@@ -252,17 +252,20 @@ interface EntityInterface extends ArrayAccess, JsonSerializable
     public function clean(): void;
 
     /**
-     * Returns whether or not this entity has already been persisted.
-     * This method can return null in the case there is no prior information on
-     * the status of this entity.
+     * Set the status of this entity.
      *
-     * If called with a boolean, this method will set the status of this instance.
-     * Using `true` means that the instance has not been persisted in the database, `false`
-     * that it already is.
+     * Using `true` means that the entity has not been persisted in the database,
+     * `false` that it already is.
      *
-     * @param bool|null $new Indicate whether or not this instance has been persisted.
-     * @return bool If it is known whether the entity was already persisted
-     * null otherwise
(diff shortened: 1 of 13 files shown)


ADmad:
I'll rebase this once #13456 is merged to get the builds passing.

src/Datasource/EntityInterface.php
```diff
@@ -252,17 +252,20 @@ public function extract(array $fields, bool $onlyDirty = false): array;
...
+     * Using `true` means that the entity has not been persisted in the database,
+     * `false` that it already is.
      *
-     * If called with a boolean, this method will set the status of this instance.
-     * Using `true` means that the instance has not been persisted in the database, `false`
-     * that it already is.
+     * @param bool $new Indicate whether or not this entity has been persisted.
+     * @return $this
+     */
+    public function setNew(bool $new);
```
garas: Default to `true` like `setDirty()`?

ADmad: Usually entity is new by default and in most cases this method is called with `false` argument. So default `true` doesn't feel right to me. So `setAccess()` does not have a default value so `setNew()` not having one is fine IMO.

src/Datasource/EntityInterface.php
```diff
@@ -252,17 +252,20 @@ public function extract(array $fields, bool $onlyDirty = false): array;
     public function clean(): void;
 
     /**
-     * Returns whether or not this entity has already been persisted.
-     * This method can return null in the case there is no prior information on
-     * the status of this entity.
+     * Set the status of this entity.
+     *
+     * Using `true` means that the entity has not been persisted in the database,
+     * `false` that it already is.
```
markstory:
```suggestion
     * `false` indicates that the entity has been persisted.
```

PR #13457, opened 2019-08-02
Fix doc blocks

bancer:


dereuromark:
They most likely were an overwrite that was there by intention - failing phpstan will probably support this.

dereuromark:
https://travis-ci.org/cakephp/cakephp/jobs/566940236#L629-L630 confirms it.
I am sure 4.x can fix those things (or already did).

ADmad:
The typehints won't change in 4.x either since there are methods like `Connection::inTransaction()` which are specific to RDBMS and don't belong in `Datasource\ConnectionInterface`.

bancer:
@ADmad There is inconsistency here - the type hint is `ConnectionInterface` but documented as `Connection`. Why the docs should be different then the type hint?

ADmad:
Ah I missed that. In that case the typehint should be changed.

dereuromark:
> Why the docs should be different than the type hint?

To make dev/IDE/tooling aware of the more concrete object needed for some cases, while the technical typehint is still agnostic.
This is a compromise - especially when working with BC promises here.

bancer:
Sounds like the interface is incomplete. 

> ... since there are methods like Connection::inTransaction() which are specific to RDBMS and don't belong in Datasource\ConnectionInterface.

If they do not really belong to ConnectionInterface why not to place them in an abstract class that implements ConnectionInterface and implement them there like `... inTransaction() {return false;}` and use that abstract class as a typehint?

ADmad:
No one is going to actually create their own collection class which extends that abstract class.

Adding abstract classes and interfaces just for technical purity gives no real benefit and just more potentially API breakages in future.

bancer:
> No one is going to actually create their own collection class...

Probably true. Do you mean that ConnectionInterface is useless?

ADmad:
> Do you mean that ConnectionInterface is useless?

No, non RBMS datasource would have their own implementation of connection which adheres to this interface. For e.g. connection class of our elastic-search plugin.

PR #13458, opened 2019-08-02
Fix invalid type hint

bancer:


dereuromark:
That is a heavy BC break
Please see the discussion of https://github.com/cakephp/cakephp/pull/13457 for details why this is not a good change at this time.

ADmad:
Yeah technically typehints shouldn't be changed in bugfix releases (or even minor as per our BC guidelines).

This change can be done in 4.x instead.

bancer:
@dereuromark Sorry, I saw your comments there only after I committed here.

dereuromark:
Instead of removing interfaces, we should have more interface contracting in 4.x.
Maybe we should get better at interface segregation and providing more correct "contracts" per scope instead.

ADmad:
Please reopen a new PR targeting `4.x` branch.

PR #13459, opened 2019-08-02
Use pcov for code coverage.

ADmad:


markstory:
Pcov/clobber is a neat little library.

ADmad:
Yup, clobber is only required for phpunit < 8 which don't have in built support for pcov. So clobber patches it's xdebug driver files :)

PR #13464, opened 2019-08-05
Fix translate behavior tests.

ADmad:
EavStrategy of TranslateBehavior needs translated fields to be explicity specified.

Refs #13422

//cc @ndm2

### DIFF
--- a/tests/TestCase/ORM/Behavior/TranslateBehaviorTest.php
+++ b/tests/TestCase/ORM/Behavior/TranslateBehaviorTest.php
@@ -1859,29 +1859,6 @@ class TranslateBehaviorTest extends TestCase
     }
 
     /**
-     * Tests that using matching doesn't cause an association property to be created.
-     *
-     * @return void
-     */
-    public function testMatchingDoesNotCreateAssociationProperty()
-    {
-        $table = $this->getTableLocator()->get('Articles');
-        $table->hasMany('Comments');
-
-        $table->Comments->addBehavior('Translate', ['fields' => ['comment']]);
-        $table->Comments->setLocale('abc');
-
-        $this->assertNotEquals($table->Comments->getLocale(), I18n::getLocale());
-
-        $result = $table
-            ->find()
-            ->matching('Comments')
-            ->first();
-
-        $this->assertArrayNotHasKey('comments', $result->toArray());
-    }
-
-    /**
      * Tests that using deep matching doesn't cause an association property to be created.
      *
      * @return void
@@ -1908,34 +1885,6 @@ class TranslateBehaviorTest extends TestCase
             ->first();
 
         $this->assertArrayNotHasKey('author', $result->comments);
-    }
-
-    /**
-     * Tests that using contained matching doesn't cause an association property to be created.
-     *
-     * @return void
-     */
-    public function testContainedMatchingDoesNotCreateAssociationProperty()
-    {
-        $table = $this->getTableLocator()->get('Authors');
-        $table->hasMany('Comments')->setForeignKey('user_id');
-        $table->Comments->belongsTo('Articles');
-
-        $table->Comments->Articles->addBehavior('Translate', ['fields' => ['title', 'body']]);
-        $table->Comments->Articles->setLocale('xyz');
-
-        $this->assertNotEquals($table->Comments->Articles->getLocale(), I18n::getLocale());
-
-        $result = $table
-            ->find()
-            ->contain([
-                'Comments' => function ($query) {
-                    return $query->matching('Articles');
-                },
-            ])
-            ->first();
-
-        $this->assertArrayNotHasKey('article', $result->comments[0]->toArray());
     }
 
     /**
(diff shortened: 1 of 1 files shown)


ADmad:
I am not sure how to fix the failing tests.

ADmad:
"EavStrategy" is reference from 4.x as that's where I started digging the issue :)  It's the only and default implementation in 3.x.

ndm2:
> I am not sure how to fix the failing tests.

These tests were ment for the exact situation when the translate behavior "does nothing" so to speak, in that case it shouldn't cause an association property to be created, but it's basically a misconfiguration scenario, so that might not be the best way to test the problem.

I'm not sure though that the translate behavior should even cause the associated data to be loaded/injected when fields _are_ specified, without an explicit `contain()` this seems rather unexpected. @lorenzo Any thoughts on that?

ADmad:
> These tests were ment for the exact situation when the translate behavior "does nothing" so to speak, in that case it shouldn't cause an association property to be created, but it's basically a misconfiguration scenario, so that might not be the best way to test the problem.

Yeah having tests relying on mis-configuration of plugin makes no sense. (I am going to make EavStrategy show an exception in 4.x if `fields` are not specified).

I'm not sure though that the translate behavior should even cause the associated data to be loaded/injected when fields are specified, without an explicit contain() this seems rather unexpected. @lorenzo Any thoughts on that?

While there is no explicit contain there is an implicit once, in `beforeFind()` of the behavior:

https://github.com/cakephp/cakephp/blob/bb06ced5d450c00d484039a6b22a4bb026cb17f3/src/ORM/Behavior/TranslateBehavior.php#L261

ndm2:
> While there is no explicit contain in your tests there is an implicit once, in `beforeFind()` of the behavior

I know, but that's a "sub-containment" for the translation table, and since it's applied to the `Comments` association query, it shouldn't cause association data to be placed in the results when the `Comments` association is not explicitly contained. The `Comments` association is implicitly contained by `matching()`, but that shouldn't cause any association data in the results except for in `_matchingData`.

At some point the containment from the behaviors `beforeFind()` results in an additional, non-joinable, non-matching `Comments` containment to slip in for the main query, causing the association map in `ResultSet::_calculateAssociationMap()` to look like this:

```php
[
	(int) 0 => [
		'alias' => 'Comments',
		'instance' => object(Cake\ORM\Association\HasMany) {
			// ...
		},
		'canBeJoined' => true,
		'entityClass' => 'Cake\ORM\Entity',
		'nestKey' => 'Comments',
		'matching' => true,
		'targetProperty' => 'comments'
	],
	(int) 1 => [
		'alias' => 'Comments',
		'instance' => object(Cake\ORM\Association\HasMany) {
			// ...
		},
		'canBeJoined' => false,
		'entityClass' => 'Cake\ORM\Entity',
		'nestKey' => 'Comments',
		'matching' => false,
		'targetProperty' => 'comments'
	]
]
```

I can't tell whether that might be a bug, but IMHO it's at least undesirable, it doesn't affect the main queries SQL, and it wasn't explicitly requested.

ADmad:
For find call in `testMatchingDoesNotCreateAssociationProperty()` due to the `contain()` in `beforeFind()` you eventually end up with nested/deep contain

```php
        'contain' => [
                'Comments' => [
                        'Comments_comment_translation' => [
                                'queryBuilder' => object(Closure) {

                                }
                        ]
                ]
        ],
```
for the query. (debug the query before calling `first()` and you will see).

And that's why we end up with `comments` property for the article entity.

So from what I understand the issue cannot be resolved without replacing use of `contain()` with explicit joins and field selection as I suggested above.

On a side note I personally avoid using `matching()` at all. It's too magical for my liking.

ndm2:
> due to the `contain()` in `beforeFind()` you eventually end up with nested/deep contain

I know, I'm just not sure that this is really the correct behavior. From what I could figure so far, this happens in `Association::_bindNewAssociations()`, there it will grab the containment that the behavior added to the association query, and applies it on the main query.

I don't fully understand the original purpose of `Association::_bindNewAssociations()` yet, but the association for which that happens here is a matching-containment, so I'm wondering whether that could maybe be avoided without breaking things (the current tests don't break when bailing out for matching containments, but that doesn't mean too much of course, it could very well be an untested scenario).

ndm2:
OK after some more digging it seems this is generally used for chaining nested matching and containments that are being defined in callbacks with query builders, and there's actually two tests that fail when also disabling the matching inheritance.

I guess my question boils down to whether this

```php
$mainQuery->matching('Foo', function ($query) {
    return $query->contain('Bar');
});
```

and this:

```php
$mainQuery->innerJoinWith('Foo', function ($query) {
    return $query->contain('Bar');
});
```

is valid, ie whether this nested `contain()` should affect the main queries contain graph. If it should, then I guess just ditch the two tests, and if it shouldn't, then maybe it can be fixed, and then the tests should remain.

Sorry for all the noise :)

markstory:
I would think both those scenarios *should* impact the contain graph or should emit an exception indicating that the developer has asked for something we cannot complete.

ADmad:
I removed the 2 tests which are not expected to pass.

Issue #13444, opened 2019-07-26
CakePHP 4: assertSession() signature


mirko-pagliai:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 4.x

### What happened
This is the signature for `IntegrationTestTrait:assertSession()`:

    public function assertSession(string $expected, string $path, string $message = ''): void

It expects that `$expected` to be necessarily a string.

This doesn't seem to make sense to me.
I expect to be able to check, for example, an entire array.
Yes, I could do the verification for every single value, but for example a value could also be an integer.

dereuromark:
I agree, that inside a session, there could also be a numeric value or other values.
As for array: Could we just make another non-scalar one for this, assertSessionArray() or sth?

mirko-pagliai:
The same applies to the `assertCookieEncrypted()` method, whereas instead (in contradiction) the `assertCookie()` method can take any value.

Again, while the `assertCookieNotSet()` method exists, the `assertSessionNotSet()` method does not exist (in my tests, for example, I need to verify that a value in session has been deleted ...).

> I agree, that inside a session, there could also be a numeric value or other values.
> As for array: Could we just make another non-scalar one for this, assertSessionArray() or sth?

PHP allows any type for a session value. The same applies to the [Session::write()](https://github.com/cakephp/cakephp/blob/6b9efe83043650ea47d6311e3357bcfd4e61b675/src/Http/Session.php#L467) method:

    public function write($name, $value = null): void

A method should be implemented for each type of data. I would simply change the current method.

markstory:
Removing the type constraints on the current assertions is a good solution to me. Adding an assertion type for each primitive seems excessive in a high level API. If we want the additional type constraints `assertSession()` could use `assertSame()` instead of `assertEquals()`.

dereuromark:
PR https://github.com/cakephp/cakephp/pull/13466 merged

PR #13466, opened 2019-08-05
CakePHP 4: removed some useless type constraints

mirko-pagliai:
See https://github.com/cakephp/cakephp/issues/13444


ADmad:
The build failures are unrelated and caused by 3.x => 4.x merge. I'll take care of that.

PR #13468, opened 2019-08-05
[4.next] Add a convenience method to load plugins

ravage84:
This is an RFC to add three convenience methods to load plugins:

- ``addOptionalPlugin()``, do not halt when the plugin is not installed
- ``addCliPlugin()``, only load the plugin when in CLI
- ``addOptionalCliPlugin()``, only load the plugin when in CLI and do not halt when the plugin is not installed

TODO:

- ~It will need tests~.
- ~I didn't extend the ``PluginApplicationInterface`` where ``addPlugin()`` is located.~

### DIFF
--- a/src/Http/BaseApplication.php
+++ b/src/Http/BaseApplication.php
@@ -14,11 +14,13 @@ declare(strict_types=1);
  * @since         3.3.0
  * @license       https://opensource.org/licenses/mit-license.php MIT License
  */
+
 namespace Cake\Http;
 
 use Cake\Console\CommandCollection;
 use Cake\Controller\ControllerFactory;
 use Cake\Core\ConsoleApplicationInterface;
+use Cake\Core\Exception\MissingPluginException;
 use Cake\Core\HttpApplicationInterface;
 use Cake\Core\Plugin;
 use Cake\Core\PluginApplicationInterface;
@@ -117,6 +119,60 @@ abstract class BaseApplication implements
             $plugin = $name;
         }
         $this->plugins->add($plugin);
+
+        return $this;
+    }
+
+    /**
+     * Add an optional plugin
+     *
+     * If it isn't available, ignore it.
+     *
+     * @param string|\Cake\Core\PluginInterface $name The plugin name or plugin object.
+     * @param array $config The configuration data for the plugin if using a string for $name
+     * @return $this
+     */
+    public function addOptionalPlugin($name, array $config = [])
+    {
+        try {
+            $this->addPlugin($name, $config);
+        } catch (MissingPluginException $e) {
+            // Do not halt if the plugin is missing
+        }
+
+        return $this;
+    }
+
+    /**
+     * Add a plugin, when in CLI
+     *
+     * @param string|\Cake\Core\PluginInterface $name The plugin name or plugin object.
+     * @param array $config The configuration data for the plugin if using a string for $name
+     * @return $this
+     */
+    public function addCliPlugin($name, array $config = [])
+    {
+        if (PHP_SAPI === 'cli') {
+            $this->addPlugin($name, $config);
+        }
+
+        return $this;
+    }
+
+    /**
+     * Add an optional plugin, when in CLI
+     *
+     * If it isn't available, ignore it.
+     *
+     * @param string|\Cake\Core\PluginInterface $name The plugin name or plugin object.
+     * @param array $config The configuration data for the plugin if using a string for $name
+     * @return $this
+     */
+    public function addOptionalCliPlugin($name, array $config = [])
+    {
+        if (PHP_SAPI === 'cli') {
+            $this->addOptionalCliPlugin($name, $config);
+        }
 
         return $this;
     }
--- a/src/Http/BaseApplication.php
+++ b/src/Http/BaseApplication.php
@@ -144,40 +144,6 @@ abstract class BaseApplication implements
     }
 
     /**
-     * Add a plugin, when in CLI
-     *
-     * @param string|\Cake\Core\PluginInterface $name The plugin name or plugin object.
-     * @param array $config The configuration data for the plugin if using a string for $name
-     * @return $this
-     */
-    public function addCliPlugin($name, array $config = [])
-    {
-        if (PHP_SAPI === 'cli') {
-            $this->addPlugin($name, $config);
-        }
-
-        return $this;
-    }
-
-    /**
-     * Add an optional plugin, when in CLI
-     *
-     * If it isn't available, ignore it.
-     *
-     * @param string|\Cake\Core\PluginInterface $name The plugin name or plugin object.
-     * @param array $config The configuration data for the plugin if using a string for $name
-     * @return $this
-     */
-    public function addOptionalCliPlugin($name, array $config = [])
-    {
-        if (PHP_SAPI === 'cli') {
-            $this->addOptionalCliPlugin($name, $config);
-        }
-
-        return $this;
-    }
-
-    /**
      * Get the plugin collection in use.
      *
      * @return \Cake\Core\PluginCollection
--- a/tests/TestCase/Http/BaseApplicationTest.php
+++ b/tests/TestCase/Http/BaseApplicationTest.php
@@ -29,6 +29,8 @@ use TestPlugin\Plugin as TestPlugin;
 
 /**
  * Base application test.
+ *
+ * @coversDefaultClass \Cake\Http\BaseApplication
  */
 class BaseApplicationTest extends TestCase
 {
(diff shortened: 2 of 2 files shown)


ADmad:
This unnecessarily increases the API surface IMO. Adding a SAPI check or try catch block in Application itself is easy enough.

markstory:
I can see them being useful to keep application class code high-level and concise. They do add surface area where we don't necessarily need it. Could we get similar results with options in `addPlugin()`?

dereuromark:
The optinal part could be an option. Then we only have two speaking methods.

ADmad:
I am find handling this through options for `addPlugin()`.

dereuromark:
Since you modify the interface, this would be best to put into 4.x right now.
In 3.x people can inject this via trait from plugin or app if they want to use it (could be placed into Shim plugin).

ravage84:
@markstory 

> Could we get similar results with options in addPlugin()?

Sure we could, but in the past we had a lot fo such code, e.g. in ORM and we extracted a few of them into proper methods.

> In 3.x people can inject this via trait from plugin or app if they want to use it (could be placed into Shim plugin).

The framework built-in trait could be provided as a bridge to 4.x.

> Or we backport it without changing the interface and adding them as soft @method annotations into the interface in 3.next.

I wasn't sure, if we really need to change the interface. Do we?

ravage84:
I'm going to reduce the list of the new methods to only ``addOptionalPlugin()``, as discussed https://github.com/cakephp/cakephp/pull/13476#discussion_r311192751 and https://github.com/cakephp/cakephp/pull/13476#issuecomment-519025250.

But first, I would like to get a consensus on where and how to put that method in:

1. Generally in 3.x yes or no?
2. If in 3.x, as a method or as a new trait (name suggestions, please)?
3. If in 3.x, changing the interface or not (I presume not)
4. For 4.x, as a method and with changing the interface, correct?

If we add this to 3.x, I'm going to create another PR for 4.x.

dereuromark:
I would say:
- 3.x yes as method on the class and annotation for interface (soft addition as usual)
- 4.x as real interface method.

saeideng:
for cli I prefer to 
1) adding `addPlugin(...)` to `bootstrapCli()` method in `Apllication.php` file
2) of by argument `addPlugin(... , ['onlyCli'=>true])`

dereuromark:
@saeideng As discussed this is not necessary since there is a cli hook method already present and documented. I would prefer not going down this less verbose and clear direction. See linked thread for details.

ADmad:
Besides `DebugKit` can someone provide a practical example of an optional plugin?

dereuromark:
I have Plugin::load('TestHelper'); etc which also, based on the require/require-dev part you put it in could blow up maybe.

ravage84:
@ADmad sorry for not including a rationale.

Bake is such a plugin we load optionally.
Also, the Fixturize plugin.

Both are development plugins, which we do not deploy, so they cannot be loaded.
But they don't need to in production, thus optional

ADmad:
> But they don't need to in production, thus optional

I see them as "conditional" not "optional" and load dev only plugins based on environment checks. But I guess others view it differently.

robertpustulka:
> Both are development plugins, which we do not deploy, so they cannot be loaded.

If those are dev plugins maybe you should wrap them with debug option check? So maybe a config option `when` is what you need which accepts bool or callable, for example:

```php
$this->addPlugin('Bake', [
    'when' => $isDebug && $isCli
]);
```

dereuromark:
@ravage84 Do you want to finish this up? Or do we keep the status quo?

ravage84:
@dereuromark yes, if there is a real consensus on  how to implement it. But I'm not going to get to it before next week.

ADmad:
I am happy with Robert's proposal.

ravage84:
I am OK with @robertpustulka 's proposal, if we can add the ``optional`` array key, which was proposed as alternative from @dereuromark , as I disagree that it's not  *just conditional*, but __optional__.

My prposed four methods could be merged into one and would look like this, then:

```` php
// Mandatory plugin
$this->addPlugin('Bake');

// Optional plugin
$this->addPlugin('Bake', [
    'optional' => true,
]);

// CLI plugin
$this->addPlugin('Bake', [
    'when' => PHP_SAPI === 'cli'
]);

// Optoinal CLI plugin
$this->addPlugin('Bake', [
    'when' => PHP_SAPI === 'cli',
    'optional' => true,
]);
````

**Comment by inoas** [comment]
I prefer @ravage84 original suggestion aka a bigger API surface over Hash key configuration that fails with typos (php still misses atoms and pattern matching).

But both work well.

**Comment by dereuromark** [comment]
Yeah, at some point the methods seem to be more speakable than this config string key thingie :)
Especially if there is both when and optional now as keys.

**Comment by ADmad** [comment]
If people really want the extra methods then we can add them to `BaseApplication` but not to the `PluginApplicationInterface` as these are just additional convenience functions.

**Comment by ADmad** [comment]
@ravage84 We have shell / command to load plugins, would you be updating that too?

**Comment by robertpustulka** [comment]
@ravage84 can you elaborate on when loading the plugin is *optional*? I can't think of a use case for this. You gave examples of a `Bake` plugin but this is a dev plugin IMO and loading it should be conditional on whether the debug mode is enabled and/or on the environment (cli/web). DebugKit is conditional on dev mode as well.

**Comment by dereuromark** [comment]
ping @ravage84

**Comment by ravage84** [comment]
> @ravage84 We have shell / command to load plugins, would you be updating that too?

I would consider it, at least.

> ping @ravage84

Didn't have time up to now. Will be on vacation next, week unfortunately.

**Comment by ravage84** [comment]
> @ravage84 can you elaborate on when loading the plugin is optional? I can't think of a use case for this. You gave examples of a Bake plugin but this is a dev plugin IMO and loading it should be conditional on whether the debug mode is enabled and/or on the environment (cli/web). DebugKit is conditional on dev mode as well.

While loading such plugins simply depending on the debug mode is an option, I think it is too simple.

1. There are times, where I disable the debug mode when developing locally. 
2. Also there are times when I enable the debug mode in, for example, test environments (for inspecting run time issues). Since test enviornments are installed from the same built artifact as the production server (1 code, 1 build artifact, many installations mindset) which does not contain any dev dependencies, it would mean the application would try to load plugins that aren't installed.

**Comment by Ali1** [comment]
> > Both are development plugins, which we do not deploy, so they cannot be loaded.
> 
> If those are dev plugins maybe you should wrap them with debug option check? So maybe a config option `when` is what you need which could accept bool or callable, for example:
> 
> ```
> $this->addPlugin('Bake', [
>     'when' => $isDebug && $isCli
> ]);
> ```

How is this easier than just adding an if statement
```
if ($isDebug && $isCli) {
     $this->addPlugin('Bake);
}
```

I think `addDevelopmentPlugin` or `addDebugPlugin` (only loads when in Debug mode) and `addCliPlugin` would be quite neat cool additions to BaseApplication

**Comment by Ali1** [comment]
While we're on the topic, would allowing this feature be useful: `bin/cake annotations models --plugins=IdeHelper` (https://github.com/dereuromark/cakephp-ide-helper/tree/master/docs) as some cli plugins only need to be loaded with certain commands and need not clutter Application.php....

**Comment by dereuromark** [comment]
>  would allowing this feature be useful

What part of it? This already works even for not loaded ones, I dont see how this related to the ticket here.

As for the topic at hand:
I am not sure what to think. I am somewhat undecided if we need those methods, or if the current try catch solution suffices. The latter (somewhat a `loadOptionalPlugin` way) has the advantage of just silently not loading it if it is not available.

Usually you decide on this anyway through require and require-dev in your composer, creating a new separation in Application seems a bit weird then to me, too.

**Comment by othercorey** [comment]
@ravage84 Is this something you can finish or should we close and leave for a future issue?

**Comment by dereuromark** [comment]
@ravage84 
I thought about this some more and I think there can be a consensus.
Are you ok simplifying this to only one method addition?

    public function addOptionalPlugin()

This is all that is needed IMO, it is syntactic sugar that avoids a try/catch block (which we currently use).
The cli check is simple enough (`PHP_SAPI !== ''cli'`) and already usually then delegating to a protected method doing batch adding.
So having 2 more methods seems overkill.

Is this a compromise we can all live with? To move things forward.

**Comment by ravage84** [comment]
> Is this a compromise we can all live with? To move things forward.

@dereuromark yes, I can live with that. We solve the rest in our application, then. :man_shrugging:  :roll_eyes: 
Though, I still wait for other core members to voice their views, first.

**Comment by ADmad** [comment]
> I can't think of a use case for this. You gave examples of a Bake plugin but this is a dev plugin IMO and loading it should be conditional on whether the debug mode is enabled and/or on the environment (cli/web). DebugKit is conditional on dev mode as well.

I remain of the same opinion as above but I can live with one additional method.

**Comment by markstory** [comment]
I'm good with an `addOptionalPlugin()` method. Removing that boilerplate would be nice.

**Comment by dereuromark** [comment]
I updated as per consensus.
Maybe you want to add a small test case?

**Comment by dereuromark** [comment]
Should this target 4.next now?

**Comment by ADmad** [comment]
Yes it needs to target 4.next.

**Comment by ravage84** [comment]
> 
> 
> Yes it needs to target 4.next.

@ADmad what is your reasoning for that? This here could land in 3.9.0. I deliberately made it target 3.x so I/we as a company have it (in the core) for our 3.x applications.

**Comment by dereuromark** [comment]
After 4.next (4.1) merge, you can make a backport PR towards 3.next (3.9).
We just usually don't add new methods in the current minor.

**Comment by othercorey** [comment]
@ravage84 The master branch has changed since you opened this PR. With the release of 4.0, master branch is now 4.x.

**Comment by ravage84** [comment]
> @ravage84 The master branch has changed since you opened this PR. With the release of 4.0, master branch is now 4.x.

Thanks for pointint that out. My intent was 3.x but 3.next would probably have been better.

> After 4.next (4.1) merge, you can make a backport PR towards 3.next (3.9).
We just usually don't add new methods in the current minor.

What is the reason for the wait?

**Comment by dereuromark** [comment]
You can open also both in parallel, but if there are any further changes necessary on the main 4.x one, then you would need to re-adjust also the other one.
If you just let the primary one pass/merge, then you can just quickly create and touch the 3.x one (from patch).

**Comment by othercorey** [comment]
We can just change this to 4.next and easily back-port when committed.

File: src/Http/BaseApplication.php [hunk 1]
```diff
@@ -109,6 +111,26 @@ public function addPlugin($name, array $config = [])
...
+    /**
+     * Add an optional plugin
+     *
+     * If it isn't available, ignore it.
+     *
+     * @param string|\Cake\Core\PluginInterface $name The plugin name or plugin object.
+     * @param array $config The configuration data for the plugin if using a string for $name
+     * @return $this
+     */
+    public function addOptionalPlugin($name, array $config = [])
```
markstory, line 123: If we get a test for this it looks 😍

ravage84:
Tests added.

dereuromark:
Looks good. After merge, we can also backport this to 3.x

PR #13470, opened 2019-08-06
Support for char column type

karmicdice:
Refers to https://github.com/cakephp/migrations/issues/362


src/Database/TypeFactory.php
```diff
@@ -44,6 +44,7 @@ class TypeFactory
         'float' => Type\FloatType::class,
         'json' => Type\JsonType::class,
         'string' => Type\StringType::class,
+        'char' => Type\CharType::class,
```
ADmad: `CharType` is identical to `StringType` isn't it? If so then I don't see the need for code duplication. The `char` type can just be mapped to `StringType` similar to `text`.

src/Database/Schema/PostgresSchema.php
```diff
@@ -393,6 +394,10 @@ public function columnSql(TableSchema $schema, string $name): string
             $out .= ' BYTEA';
         }
 
+        if ($data['type'] === TableSchema::TYPE_CHAR) {
+            $out .= '(' . (int)$data['length'] . ')';
```
ADmad: Why cast to int when doing string concatenation?

karmicdice: A condition below this has cast it to (int). I found it weird too.. so for the sake of consistency.

src/Database/TypeFactory.php
karmicdice: Nope.
Since 7.2 The flag is "PDO::PARAM_STR_CHAR" Ref:  https://www.php.net/manual/en/pdo.constants.php 
However, I was not sure of the compatibility i.e. what is the minimum php we need for 4x so I kept it as "PDO::PARAM_STR".
Created a new Type so that once we deprecate support for < 7.2, we can use the new flag.
Turns out, Travis is set to minimum 7.2 , should I go ahead and put the new flag instead? What do you advise?

ADmad: CakPHP required PHP 7.2+ so that's not an issue. The description for `PDO::PARAM_STR` says `Represents the SQL CHAR, VARCHAR, or other string data type.` so it seems good enough for `char` fields too.

For `PDO::PARAM_STR_CHAR` it says `Flag to denote a string uses the regular character set.`. I am not sure what that means and in what case it should be used. Maybe someone more knowledge on that matter does.

So for now I suggest we simply use the `StringType` class. We can always update later if needed.

src/Database/Schema/PostgresSchema.php
ADmad: Let's try removing the casting and see if any tests fail :)

karmicdice: Alrighty!

src/Database/TypeFactory.php
karmicdice: Cool! On it!

karmicdice:
@ADmad Ok! So removing int cast did not fail any tests, so I reckon we are good.
Can you help me understand what this it? https://github.com/cakephp/cakephp/pull/13470/checks?check_run_id=186164738 ?

ADmad:
@karmicdice Those failures are due to a bug in slevomat coding standards we are using. They are unrelated to your changes.

tests/TestCase/Database/Type/CharTypeTest.php
```diff
@@ -0,0 +1,103 @@
+<?php
```
ADmad: This file needs to be removed.

karmicdice: Done.

ADmad:
It would be good to add `'char' => 'text'` mapping in `FormHelper`'s `typeMap` for the new `char` type even though generating inputs with `type="text"` is the default.

tests/TestCase/Database/Schema/MysqlSchemaTest.php
```diff
@@ -114,7 +114,7 @@ public static function convertColumnProvider()
             ],
             [
                 'CHAR(25)',
-                ['type' => 'string', 'length' => 25, 'fixed' => true],
+                ['type' => 'char', 'length' => 25, 'fixed' => true],
```
markstory: Do we need to retain the `fixed` option on `char` types? The type=char should be enough to infer that it is a fixed length type.

markstory: Should migrations be updated to understand the fixed flag? That might allow us to not have another abstract type.

karmicdice: Having fixed flag in test will respond with the same as without flag. I kept it as is.
Because char infers that default value of fixed is true.
I haven't figured the implication on migrations yet with respect to the flag.

karmicdice:
> It would be good to add `'char' => 'text'` mapping in `FormHelper`'s `typeMap` for the new `char` type even though generating inputs with `type="text"` is the default.

Yep. Noticed it and kept it as-is leaving it to default on the fall back.
If you want, I can still go ahead and explicitly define it as text.. Let me know.

karmicdice:
Anything I can do to push this further? :)

markstory:
What does FormHelper do with char fields?

karmicdice:
> What does FormHelper do with char fields?

Falls to default - 'text'

karmicdice:
Wassup guys...
Any update on this one?

ADmad (dismissed):

tests/TestCase/Database/Schema/MysqlSchemaTest.php
```diff
@@ -1075,7 +1075,7 @@ public function testCreateSql()
                 'type' => 'json',
             ])
             ->addColumn('hash', [
-                'type' => 'string',
+                'type' => 'char',
                 'fixed' => true,
```
markstory: I was writing the docs up and I'm wondering what the behavior should be when a fixture has `type => string, fixed => true` now? Does the generated column end up as a CHAR or a VARCHAR? Should this scenario emit an error, does the `fixed` flag even make sense anymore?

karmicdice: CHAR _implies_ that the column is fixed. I think having a flag should not make sense anymore but, I am not sure about the users. Is it intuitive enough...? Well, I am confused...

markstory: Could we remove the `fixed` flag? I agree it is confusing especially in the context of both a `string` and `char` type.

ADmad: Now that we have a separate type itself for fixed length columns having the `fixed` flag doesn't make sense.

karmicdice: Cool!!! Will do it this weekend...... 🤘

markstory:
You have a few conflicts now.

karmicdice:
Test for `Coding standards` is telling me that there was an extra character found at the end of interface file. I could not find the character. I am on Windows 10, VS Code / Sublime, tried both.

ADmad:
@karmicdice We have conflicts again :slightly_smiling_face:. The CS issues can be auto fixed by using `phpcbf` (utility which is included with phpcs) locally.

dereuromark:
ping @karmicdice

karmicdice:
> @karmicdice We have conflicts again . The CS issues can be auto fixed by using `phpcbf` (utility which is included with phpcs) locally.

Owh bummer!

karmicdice:
> @karmicdice We have conflicts again . The CS issues can be auto fixed by using `phpcbf` (utility which is included with phpcs) locally.

After trying various methods (including using vscode), it turns out that using Sublime Text with phpcs is more helpful... However, it requires each file to be opened, edited and saved; to trigger the checker.

karmicdice:
What is "Coverage decreased"? How do I increase it?

ADmad:
> What is "Coverage decreased"? How do I increase it?

Ignore that. No one has been able to figure out how these coverage reporting services work :stuck_out_tongue: .

karmicdice:
> > What is "Coverage decreased"? How do I increase it?
> 
> Ignore that. No one has been able to figure out how these coverage reporting services work  .

😆 Haha
So is it ready to be merged?

tests/TestCase/Database/Schema/MysqlSchemaTest.php
```diff
@@ -535,7 +535,7 @@ public static function columnSqlProvider()
             ],
             [
                 'id',
-                ['type' => 'uuid'],
```
othercorey: Did you intend to remove CHAR(36) => UUID mapping for mysql?

karmicdice: The mapping doesn't work?

karmicdice: It is already there at L120

karmicdice: Added additional test.

karmicdice:
Can someone help me understand what phpstan is trying to convey?

othercorey:
I get it too.  I think some merges collided.

ADmad:
@karmicdice @othercorey Ignore that error. That particular line uses new signature for `setcookie()` available only since PHP 7.3.  It have pushed a commit to `4.x` to run phpstan on PHP 7.3 now.

othercorey:
@karmicdice Looks good! I hope I didn't block anything with my comment.

karmicdice:
@othercorey Not the slightest...  Infact adding more test cases is never a bad thing :)

tests/TestCase/Database/Schema/PostgresSchemaTest.php
```diff
@@ -990,9 +989,11 @@ public function testConstraintSql($name, $data, $expected)
...
             'type' => 'string',
             'length' => 255,
-        ])->addColumn('author_id', [
-            'type' => 'integer',
-        ])->addConstraint($name, $data);
+            ])
+        ->addColumn('author_id', [
+                'type' => 'integer',
+                ])
+        ->addConstraint($name, $data);
```
markstory: What's going on with the formatting here?

ADmad:
@karmicdice Please revert all the unnecessary whitespace and formatting changes in the test files.

tests/TestCase/Database/Schema/PostgresSchemaTest.php
karmicdice: Sorry phpcs asked me to make these changes...

karmicdice:
> @karmicdice Please revert all the unnecessary whitespace and formatting changes in the test files.

phpcs would let me save things? Nevermind, I will make the changes.

tests/TestCase/Database/Schema/PostgresSchemaTest.php
karmicdice: Removed whitespaces. I had phpcs installed with my IDE and I made changes exactly as it recommended. Reverted. Please find the next commit.

tests/TestCase/Database/Schema/SqlserverSchemaTest.php
```diff
@@ -983,7 +983,7 @@ public function testCreateSql()
 [title] NVARCHAR(255) NOT NULL,
 [body] NVARCHAR(MAX),
 [data] NVARCHAR(MAX),
-[hash] NCHAR(40) COLLATE Latin1_General_BIN NOT NULL,
+[hash] CHAR(40) COLLATE Latin1_General_BIN NOT NULL,
```
markstory: This column type cannot store unicode. Can we continue using `NCHAR` please?

othercorey:
It feels a little painful that a non-unicode character array is considered a char type but a unicode character array is considered a string in sqlserver.

karmicdice:
> It feels a little painful that a non-unicode character array is considered a char type but a unicode character array is considered a string in sqlserver.

Exactly my thoughts except the pain. I said, "Hmm, weird. Well! That's life." and moved on :stuck_out_tongue:

othercorey:
> Exactly my thoughts except the pain. I said, "Hmm, weird. Well! That's life." and moved on stuck_out_tongue

Well it is caused by cakephp here.  It is missing a `unicode` attribute for`TYPE_CHAR`.  With that the `fixed` attribute of `TYPE_STRING` wouldn't be needed right?

karmicdice:
> > Exactly my thoughts except the pain. I said, "Hmm, weird. Well! That's life." and moved on stuck_out_tongue
> 
> Well it is caused by cakephp here. It is missing a `unicode` attribute for`TYPE_CHAR`. With that the `fixed` attribute of `TYPE_STRING` wouldn't be needed right?

Sorry, I am lost here. Can you propose the exact change?

karmicdice:
> Well it is caused by cakephp here. It is missing a `unicode` attribute for`TYPE_CHAR`. With that the `fixed` attribute of `TYPE_STRING` wouldn't be needed right?

Tried understanding... You mean, `fixed` flag be removed and instead of string we use `char`... because char implies fixed and will take the length value?

Also, `char` is a non-unicode data type. whereas nchar, nvarchar, ntext are unicode data types.

Somehow your comment is making and not making sense to me :D Haha

othercorey:
`TYPE_CHAR` would support both `char` and `nchar` while `TYPE_STRING` would support both `varchar` and `nvarchar`.  A `unicode` (or probably better `wide_char`) attribute would control which one is used.

Just a thought.  As it is, you will be supporting both fixed length strings and fixed length char arrays differently in sqlserver.

src/Database/Schema/SqlserverSchema.php
```diff
@@ -357,6 +361,7 @@ public function columnSql(TableSchema $schema, string $name): string
             TableSchema::TYPE_BIGINTEGER => ' BIGINT',
             TableSchema::TYPE_BINARY_UUID => ' UNIQUEIDENTIFIER',
             TableSchema::TYPE_BOOLEAN => ' BIT',
+            TableSchema::TYPE_CHAR => ' CHAR',
```
markstory: This should be `NCHAR` not `CHAR`. The test shouldn't have been updated to use `nchar` as that isn't an abstract type the ORM supports.

markstory:
@othercorey The abstract types we provide should be mapping to the unicode character types in SQLServer. All other databases use unicode types and SQLServer should follow suit.

src/Database/Schema/SqlserverSchema.php
karmicdice: Cool!
I have updated it.
Maybe, would be wise to put a note in docs... that orm maps to unicode d types and not non unicode like `char` as considered by sql server or is that an implied understanding with users who use sql server?

src/Database/Schema/PostgresSchema.php
```diff
@@ -112,7 +112,7 @@ protected function _convertColumn(string $column): array
             return ['type' => TableSchema::TYPE_UUID, 'length' => null];
         }
         if ($col === 'char' || $col === 'character') {
-            return ['type' => TableSchema::TYPE_STRING, 'fixed' => true, 'length' => $length];
+            return ['type' => TableSchema::TYPE_CHAR, 'fixed' => true, 'length' => $length];
```
ADmad: Why is the `fixed` key removed only from `MysqlSchema` and not from schema classes for other dbs?

karmicdice: oversight! Thanks for pointing that...

src/Database/Schema/SqlserverSchema.php
```diff
@@ -138,7 +137,7 @@ protected function _convertColumn(
         }
 
         if (strpos($col, 'char') !== false) {
-            return ['type' => TableSchema::TYPE_STRING, 'fixed' => true, 'length' => $length];
+            return ['type' => TableSchema::TYPE_CHAR, 'length' => $length];
```
karmicdice: @ADmad It wasn't implemented here. Now it is.

ADmad:
@karmicdice Your last commit messed up formatting again for `if` statements :slightly_frowning_face:. 

Make sure you have latest `phpcs` v3.5.2 locally and then use can use it's `phpcbf` to fix the formatting again.

karmicdice (commented):
> @karmicdice Your last commit messed up formatting again for `if` statements .
> 
> Make sure you have latest `phpcs` v3.5.2 locally and then use can use it's `phpcbf` to fix the formatting again.

Yes, it is at 3.2.3. Will update it. `phpcbf` acted on its own :-|

karmicdice:
> @karmicdice Your last commit messed up formatting again for `if` statements .
> 
> Make sure you have latest `phpcs` v3.5.2 locally and then use can use it's `phpcbf` to fix the formatting again.

upgrading version did not help. It did the same, is there a doc or a setting I should be doing? Please advice.

othercorey:
@markstory TYPE_CHAR > nchar resolves my concern.

karmicdice:
@markstory  We removed the fixed flag but here (https://github.com/cakephp/cakephp/pull/13470/commits/4689c7a4f52d93cbe39196ec93907a0d3f727432#diff-317ab200bb44bed550418ac58205f615L417) , it is being used to identify the column type.

Is there an alternative way you could suggest?

markstory:
> Is there an alternative way you could suggest?

Could use use the column type being equal to TYPE_CHAR?

karmicdice:
> Could use use the column type being equal to TYPE_CHAR?

The condition at https://github.com/cakephp/cakephp/blob/4689c7a4f52d93cbe39196ec93907a0d3f727432/src/Database/Schema/SqlserverSchema.php#L411-L416
 does not permit anything other than `TYPE_STRING` OR `TYPE_TEXT`
It does not go in L390 (https://github.com/cakephp/cakephp/blob/4689c7a4f52d93cbe39196ec93907a0d3f727432/src/Database/Schema/SqlserverSchema.php#L390)

othercorey:
@karmicdice Why do you want to check for fixed/char in the string logic now?  Why can't you just remove that fixed check?

karmicdice:
> @karmicdice Why do you want to check for fixed/char in the string logic now? Why can't you just remove that fixed check?

If I remove the fixed check - it will default to `string`. :(

othercorey:
@karmicdice The test cases for sqlserver need to be adjusted.  NCHAR should map to TYPE_CHAR not TYPE_STRING.

markstory:
@karmicdice I'm suggesting that fixture generation be unable to generate `CHAR` column types for SQLServer. Insetad `NCHAR` would always be used. The logic for string/text types will need to be adjusted but there is good test coverage on that logic which should help.

othercorey:
@karmicdice Would you like help with this?  If you merge latest 4.x and rebase so it's a single change, that would be easier for me to take.

karmicdice:
> @karmicdice The test cases for sqlserver need to be adjusted. NCHAR should map to TYPE_CHAR not TYPE_STRING.

Please check my comment here.
https://github.com/cakephp/cakephp/pull/13470#issuecomment-547760669

karmicdice:
> @karmicdice I'm suggesting that fixture generation be unable to generate `CHAR` column types for SQLServer. Insetad `NCHAR` would always be used. The logic for string/text types will need to be adjusted but there is good test coverage on that logic which should help.

I believe we need a unicode flag for SQL server. Mapping to nvarchar and nchar is a bit tricky. When you run migrations on sql server and write 'string' with a length, there is no way to tell that you want to generate an `nvarchar`. Introducing that flag will be better in longer run. What do you think?

ADmad:
> When you run migrations on sql server and write string with a length, there is no way to tell that you want to generate an nvarchar

@karmicdice You always generate column with nvarchar like markstory said, problem solved.

karmicdice:
> You always generate column with nvarchar like markstory said, problem solved.

So we always enforce column types as `nvarchar` and `nchar`?

ADmad:
> So we always enforce column types as nvarchar and nchar?

Yes, always use the unicode version.

karmicdice:
> > So we always enforce column types as nvarchar and nchar?
> 
> Yes, always use the unicode version.

Ok. Cool!

markstory:
Always using unicode variants makes the SQL server schema generation work consistently with other database platforms which also use unicode.

karmicdice:
> > When you run migrations on sql server and write string with a length, there is no way to tell that you want to generate an nvarchar
> 
> @karmicdice You always generate column with nvarchar like markstory said, problem solved.

@markstory @ADmad I am running tests on local. It is passing when I implementing Fixed flag. However, when I do that, Tests for Mysql, Postgres, etc are failing at `CHAR`
Having a fixed flag will make life easier. What do you think? Implementing a fixed flag is only making a double check on stuff.

ADmad:
Not sure I understand why an extra flag is required when `TYPE_CHAR` implies fixed length.

karmicdice:
> Not sure I understand why an extra flag is required when `TYPE_CHAR` implies fixed length.

Sorry. I must have gotten everyone confused.
Let me try again.

When we have type as `string`. It can either be... a text, varchar or char.
by default it is text, `if isset(length)` then it is `varchar` OR a `char`.
but `if length & fixed` then it is `char`

Pls let me know if it makes sense.

karmicdice:
Ping @markstory @ADmad 
Please check my comment and let me know.

ADmad:
> When we have type as `string`.

Where?

tests/TestCase/Database/Schema/SqlserverSchemaTest.php
```diff
@@ -208,15 +208,15 @@ public static function convertColumnProvider()
                 10,
                 null,
                 null,
-                ['type' => 'string', 'fixed' => true, 'length' => 10, 'collate' => 'Japanese_Unicode_CI_AI'],
+                ['type' => 'string', 'length' => 10, 'collate' => 'Japanese_Unicode_CI_AI'],
```
ADmad: Shouldn't the type be `char` here now?

tests/TestCase/Database/Schema/SqlserverSchemaTest.php
```diff
@@ -208,15 +208,15 @@ public static function convertColumnProvider()
...
             ],
             [
                 'NCHAR',
                 10,
                 null,
                 null,
                 // SQLServer returns double length for unicode columns.
-                ['type' => 'string', 'fixed' => true, 'length' => 5, 'collate' => 'Japanese_Unicode_CI_AI'],
+                ['type' => 'string', 'length' => 5, 'collate' => 'Japanese_Unicode_CI_AI'],
```
ADmad: And here too?

ADmad:
After making the changes I pointed out all tests pass

https://travis-ci.org/ADmad/cakephp/builds/606789518
https://github.com/cakephp/cakephp/compare/cakephp:4.x...ADmad:4.x-char-column-type

karmicdice:
> After making the changes I pointed out all tests pass
> 
> https://travis-ci.org/ADmad/cakephp/builds/606789518
> [cakephp:4.x...ADmad:4.x-char-column-type](https://github.com/cakephp/cakephp/compare/cakephp:4.x...ADmad:4.x-char-column-type)

No. The build is pointing at this commit : https://github.com/ADmad/cakephp/commit/d1256c9a1fbeb751824f0c8507bb5c56843642bb

tests/TestCase/Database/Schema/SqlserverSchemaTest.php
karmicdice: It fails?

tests/TestCase/Database/Schema/SqlserverSchemaTest.php
karmicdice: Fine! I will perform the change/.

ADmad:
> No. The build is pointing at this commit : ADmad@d1256c9

That was fixing the whitespacing mess you made :)  Check the commit before to that one on the branch.

karmicdice:
> > No. The build is pointing at this commit : [ADmad/cakephp@d1256c9](https://github.com/ADmad/cakephp/commit/d1256c9)
> 
> That was fixing the whitespacing mess you made :) Check the commit before to that one on the branch.

"mess you made" :) Ok.

karmicdice:
@ADmad Please check https://github.com/cakephp/cakephp/blob/4.x/tests/TestCase/Database/Schema/PostgresSchemaTest.php#L12-L15 (This is on current branch).
The white spacing already exists.
I made changes in the other file, which was reverted.
We all make mistakes. I could sense your frustration. Please try to be polite next time.

karmicdice:
> > No. The build is pointing at this commit : [ADmad/cakephp@d1256c9](https://github.com/ADmad/cakephp/commit/d1256c9)
> 
> That was fixing the whitespacing mess you made :) Check the commit before to that one on the branch.

@ADmad My whitespaces were reverted :)
Please check this 
https://github.com/cakephp/cakephp/blob/4.x/tests/TestCase/Database/Schema/PostgresSchemaTest.php#L12-L15
From current branch, not the mess I made. :)

Please review the changes. Is everything ok now?

ADmad:
I didn't mean to be impolite. I just pointed it out since you focussed on the commit where I fixed the white spacing instead of the one where I fixed the tests.

FYI there are still multiple white space change on your branch which didn't exist before. Check the files view of the PR https://github.com/cakephp/cakephp/pull/13470/files. 

You can just use my commit diff https://github.com/cakephp/cakephp/commit/d1256c9a1fbeb751824f0c8507bb5c56843642bb.diff  to get it sorted. Or just enable [Allow edits from maintainers](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/committing-changes-to-a-pull-request-branch-created-from-a-fork
) for your PR branch and I'll fix it myself.

karmicdice:
@ADmad Is it alright now? :)

src/Database/Schema/TableSchema.php
```diff
@@ -140,6 +140,10 @@ class TableSchema implements TableSchemaInterface, SqlGeneratorInterface
             'fixed' => null,
             'collate' => null,
         ],
+        'char' => [
+            'collate' => null,
+            'fixed' => true,
```
ADmad: This should be removed now?  And `'fixed' => null` above too?

tests/TestCase/Database/Schema/PostgresSchemaTest.php
```diff
@@ -9,10 +9,10 @@
  * For full copyright and license information, please see the LICENSE.txt
  * Redistributions of files must retain the above copyright notice.
  *
- * @copyright     Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
- * @link          https://cakephp.org CakePHP(tm) Project
- * @since         3.0.0
- * @license       https://opensource.org/licenses/mit-license.php MIT License
+ * @copyright    Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
```
ADmad: These remain :)

karmicdice: arughhhhh

karmicdice: Is there an automated way to see this??

src/Database/Schema/TableSchema.php
karmicdice: Yes for both char and string.

tests/TestCase/Database/Schema/PostgresSchemaTest.php
karmicdice: I have made changes as per the PR link. That was helpful.

karmicdice:
I will update the tests.

karmicdice:
Sent a PR #13843 instead

Issue #13471, opened 2019-08-06
Make metadata cache key configurable


bancer:
This is a (multiple allowed):

* [ ] bug
* [x] enhancement
* [x] feature-discussion (RFC)

* CakePHP Version: 3.x.
* Platform and Target: any.

### What you did
Use case: each client has its own database. Client database schemas are identical. There are thousands of clients and therefore thousands of databases.

### What happened
The framework caches metadata for each database using the connection name as part of the cache key [(1)](https://api.cakephp.org/3.8/source-class-Cake.Database.Schema.CachedCollection.html#72-81). This results in many identical value caches stored in the cache engine. This is a performance hit as queries to `information_schema`  are very slow and inefficient use of space.

### What you expected to happen
I would expect the metadata cache key name to be configurable like 'cacheMetadata' key:
```
'cacheMetadata' => true,
'cacheMetadataKey' => 'client_database',
```
So that when 'cacheMetadataKey' is not empty then it would be used as the cache key instead of the connection name. This would greatly reduce the number of queries to `information_schema` and improve performance.

bancer:
@ADmad we need this to be fixed in 3.x in order to be able to migrate from 2.x.

bancer:
In some cases queries to `information_schema` last almost 2 seconds.

![Screenshot from 2019-08-07 11:04:04](https://user-images.githubusercontent.com/3830106/62610104-44186280-b903-11e9-8d15-971089c8d81e.png)

![Screenshot from 2019-08-07 11:06:32](https://user-images.githubusercontent.com/3830106/62610185-6f02b680-b903-11e9-81ed-bd13e4e71a97.png)

ADmad:
Personally I have no interest in adding new features to 3.x unless they help a large number of users.

You can either backport the enhancement to 3.next (and wait for 3.9 release) or just use custom connection and schemacollection class and override relevant bits of code.

bancer:
Okay, could you reopen the issue please?

ADmad:
The schema collection is cached and you would see queries for `information_schema` only during development. If your schema doesn't change frequently then simply set a long duration for `_cache_model_` cache config even when debug is on.

bancer:
> Use case: each client has its own database. Client database schemas are identical. There are thousands of clients and therefore thousands of databases.

Yes, it is cached. But in this scenario such caching does not help much. We have to clear caches after each deployment. And deployments are few times per week. It takes several days to refresh all caches for thousands of databases.

[...]

PR #13472, opened 2019-08-06
Allow specifying custom cache key prefix for Schema\CachedCollection.

ADmad:
Refs #13471.

### DIFF
--- a/src/Database/Connection.php
+++ b/src/Database/Connection.php
@@ -120,7 +120,14 @@ class Connection implements ConnectionInterface
     /**
      * Constructor.
      *
-     * @param array $config configuration for connecting to database
+     * @param array $config Configuration array.
+     * ### Available options:
+     * - `driver` Sort name or FCQN for driver.
+     * - `log` Boolean indicating whether to use query logging.
+     * - `name` Connection name.
+     * - `cacheMetaData` Boolean indicating whether metadata (datasource schemas) should be cached.
+     *    If set to a string it will be used as the name of cache config to use.
+     * - `cacheKeyPrefix` Custom prefix to use when generation cache keys. Defaults to connection name.
      */
     public function __construct(array $config)
     {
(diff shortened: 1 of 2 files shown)


src/Database/Connection.php
```diff
@@ -380,7 +380,7 @@ public function getSchemaCollection(): SchemaCollectionInterface
         if (!empty($this->_config['cacheMetadata'])) {
             return $this->_schemaCollection = new CachedCollection(
                 new SchemaCollection($this),
-                $this->configName(),
+                empty($this->_config['cacheKeyPrefix']) ? $this->configName() : $this->_config['cacheKeyPrefix'],
```
bancer: I think [CachedCollection::cacheKey()](https://api.cakephp.org/3.8/source-class-Cake.Database.Schema.CachedCollection.html#72-81) that is used for metadata caching in [CachedCollection::describe()](https://api.cakephp.org/3.8/source-class-Cake.Database.Schema.CachedCollection.html#54) should be adjusted. 

Something like `return $this->_connection->getCacheKeyPrefix() . '_' . $name;` instead `return $this->_connection->configName() . '_' . $name;`.

bancer: Let me give you an example. We have cache keys like `_cake_model_client_33_areas` where `_cake_model_` is cache prefix as configured in app.php, `client_33` is the connection name, `areas` is the table name. What is needed is to set cache metadata key to `client` so that the final cache key would be `_cake_model_client_areas`.

bancer: That should affect only metadata cache keys. Cache keys to any tables other than in `information_schema` should not be affected. Queries to different client databases return different results.

ADmad: This change exactly what you want, it's just that it's targetted at 4.x not 3.x.

markstory:
Looks good to me. Shoukd this new option be in the doc block for the cached collection? The use case it is affording seems uncommon enough that this doesn't need to be in the cookbook.

ADmad:
@markstory The cache key prefix is an argument of `CacheCollection`'s constructor and documented as such. What we probably need the config options for `Connection`'s constructor to be documented.

Issue #13448, opened 2019-07-30
Unstrict loadmodel() creates followup issues


inoas:
### Code snipplet

`$this->loadmodel('tariffs');`

### Expected

To load TariffsTable or fail.

### What happens

Loads TariffsTable but sets its alias to 'tariffs'.
SQLSTATE[42S22]: Column not found: 1054 Unknown column 'Tariffs.id' in 'field list'
sql debug: `FROM tariffs tariffs`

Suggested fix:
3.x make loadModel upper case the first character
4.x make loadModel strict and not find the table if it is not {$modelname}Table

dereuromark:
I agree that there shouldnt be too much inflection going on here to avoid these kind of casing issues.

markstory:
Failing when the Table class doesn't exist will break existing applications and can't be done until 4.x.

Uppercasing the first letter will solve some scenarios but not all as multi word table names will fail in a similar way.

inoas:
then let's only fix it for 4.x and not do any juggling at all?

inoas:
(and maybe add a warning to the 3.x docs?)

dereuromark:
Can you open a PR against 4.x here?

ADmad:
> 4.x make loadModel strict and not find the table if it is not {$modelname}Table

This isn't a practical solution. Creating table classes is not mandatory.

dereuromark:
Are we OK with disallowing `'tariffs'` here? In favour of the proper alias `Tariffs`?

ADmad:
The fact that using `'tariffs'` loads `TariffsTable` is caused by this:

https://github.com/cakephp/cakephp/blob/c31eaeccb5a10017bd0607cff663417234056e47/src/ORM/Locator/TableLocator.php#L224-L226

We should see if we get a more reasonable error if that inflection is removed.

dereuromark:
I can imagine this inflection to be used for lazy-using "DB table names" directly.
But we shouldn't allow this then maybe or people need to accept the alias being used exactly as such.
I opened https://github.com/cakephp/cakephp/pull/13475 for discussing and seeing how tests react.

PR #13475, opened 2019-08-06
Use less inflection on aliasing.

dereuromark:
Let's see if this resolves https://github.com/cakephp/cakephp/issues/13448

### DIFF
--- a/src/ORM/Locator/TableLocator.php
+++ b/src/ORM/Locator/TableLocator.php
@@ -157,7 +157,7 @@ class TableLocator implements LocatorInterface
      * key in the registry. This means that if two plugins, or a plugin and app provide
      * the same alias, the registry will only store the first instance.
      *
-     * @param string $alias The alias name you want to get.
+     * @param string $alias The alias name you want to get. Must be in CamelCase format.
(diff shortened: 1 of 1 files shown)


src/ORM/Locator/TableLocator.php
```diff
@@ -157,7 +157,7 @@ public function getConfig(?string $alias = null): array
      * key in the registry. This means that if two plugins, or a plugin and app provide
      * the same alias, the registry will only store the first instance.
      *
-     * @param string $alias The alias name you want to get.
+     * @param string $alias The alias name you want to get. Must be in CamelCase format.
```
dereuromark: Or "Should be ..." maybe?

ADmad:
No test failures :slightly_smiling_face:.

ADmad:
Could you add a test showing what error/exception to expect (if at all) if some uses non conventional alias.

dereuromark:
I dont expect errors, more like using different auto created objects, one with core fallback. 
If someone has time to add tests, I'd appreciate the help.

ADmad:
You are right, there won't be any error.

ADmad (approved):

PR #13476, opened 2019-08-06
Add convenience wrappers for plugin loading.

dereuromark:
Alternative to https://github.com/cakephp/cakephp/pull/13468 for a simpler API since this is not that often used in all 4 flavors.

Could be backported either to 3.next core or in Shim plugin.

I added a more complete CLI check similar to other places in core (`$this->_isCLI = (PHP_SAPI === 'cli' || PHP_SAPI === 'phpdbg');` etc) and also made a wrapper method that allows easier customization here on project level.

### DIFF
--- a/src/Core/PluginApplicationInterface.php
+++ b/src/Core/PluginApplicationInterface.php
@@ -38,9 +38,20 @@ interface PluginApplicationInterface extends EventDispatcherInterface
      *
      * @param string|\Cake\Core\PluginInterface $name The plugin name or plugin object.
      * @param array $config The configuration data for the plugin if using a string for $name
+     * @param bool $optional If an exception for missing plugin should not be thrown.
      * @return $this
      */
-    public function addPlugin($name, array $config = []);
+    public function addPlugin($name, array $config = [], bool $optional = false);
+
+    /**
+     * Adds a plugin, but only when in CLI mode.
+     *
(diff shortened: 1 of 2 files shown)


src/Core/PluginApplicationInterface.php
```diff
@@ -38,9 +38,20 @@ interface PluginApplicationInterface extends EventDispatcherInterface
...
+
+    /**
+     * Adds a plugin, but only when in CLI mode.
+     *
+     * @param string|\Cake\Core\PluginInterface $name The plugin name or plugin object.
+     * @param array $config The configuration data for the plugin if using a string for $name
+     * @param bool $optional If an exception for missing plugin should not be thrown.
+     * @return $this
+     */
+    public function addCliPlugin($name, array $config = [], bool $optional = false);
```
ADmad: I still don't see the need to this method. We already provide a `Application::bootstrapCli()` method in app skeleton where user's can load their CLI only plugins.

ADmad: We can also add a `--cli` flag to `PluginLoadCommand` to add the `addPlugin()` statement in above method instead of `bootstrap()`.

dereuromark:
In that case we can indeed have one added convenience method
- addOptionalPlugin()

nothing else required.

ravage84:
@dereuromark agreed, I will.

PR #13481, opened 2019-08-07
Correct phpdoc for env()

o0h:
The function can return array as `getenv()`'s value.
e.g. 

```
bash-4.2# bin/cake console
You can exit with `CTRL-C` or `exit`

Psy Shell v0.9.9 (PHP 7.1.28 — cli) by Justin Hileman
env>>> env('argv')
=> [
     "/var/www/html/bin/cake.php",
     "console",
   ]
```

FYI:
But, as [php.net/getenv](https://www.php.net/manual/en/function.getenv.php), the return value is  a string unless `$varname` is omitted 🤔 

IMO, because this function can be used to get array value In the real world, so it is reliable to refer about the return value type in phpdoc.

### DIFF
--- a/src/Core/functions.php
+++ b/src/Core/functions.php
@@ -205,7 +205,9 @@ if (!function_exists('env')) {
                 return (!empty($_SERVER['HTTPS']) && $_SERVER['HTTPS'] !== 'off');
             }
 
-            return (strpos((string)env('SCRIPT_URI'), 'https://') === 0);
+            $val = env('SCRIPT_URI');
+
+            return is_string($val) ? (strpos($val, 'https://') === 0) : false;
         }
 
         if ($key === 'SCRIPT_NAME' && env('CGI_MODE') && isset($_ENV['SCRIPT_URL'])) {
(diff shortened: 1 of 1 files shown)


ADmad:
Updating the return types seems right, though it might require adding inline `@var` tags when value is expected to be string to please static analyzers :slightly_frowning_face:.

o0h:
sorry 💦 
I ran phpstan(level:7) and fix them warnings.

ADmad:
I wonder if we should go with practicality here instead of technical correctness. Using inline `@var array` when one does use `env('argv')` is better than having to adding `@var` for all other usages which are going to return `string|null`.

markstory:
@o0h Outside of argv do you know of other environment variables that contain arrays?

o0h:
Thanks for your reply 😄 

> Outside of argv do you know of other environment variables that contain arrays?

Nothing comes to my mind, $argv is an only case i know. 🤔 
Then, as @ADmad said, it's better _override_ return type of `env()` as array with inline doc for using `argv`. I don't be bothered to work around with the way.

In most other case, this function doesn't return arrays. So function's doc should describe common usage, otherwise it might be misleading.

ADmad:
Well then I vote to keep things as they are.

dereuromark:
@o0h 
You can add yourself with email @ https://github.com/cakephp/cakephp/blob/master/.mailmap
Then you will show up less cryptic on the releases like https://github.com/cakephp/cakephp/releases/tag/3.8.8 :)

o0h:
Thanks! I opend request. #14135 😃

PR #13482, opened 2019-08-07
Unset element in EntityTrait::$_original in EntityTrait::unsetProperty()

andrii-pukhalevych:
Unset element in EntityTrait::$_original along with elements in EntityTrait::$_properties and EntityTrait::$_dirty in EntityTrait::unsetProperty()

### DIFF
--- a/src/Datasource/EntityTrait.php
+++ b/src/Datasource/EntityTrait.php
@@ -440,7 +440,7 @@ trait EntityTrait
     {
         $property = (array)$property;
         foreach ($property as $p) {
-            unset($this->_properties[$p], $this->_dirty[$p]);
+            unset($this->_properties[$p], $this->_original[$p], $this->_dirty[$p]);
         }
 
         return $this;
--- a/src/Datasource/EntityTrait.php
+++ b/src/Datasource/EntityTrait.php
@@ -438,8 +438,8 @@ trait EntityTrait
      */
     public function unsetProperty($property)
     {
-        $property = (array)$property;
-        foreach ($property as $p) {
+        $field = (array)$field;
+        foreach ($field as $p) {
             unset($this->_properties[$p], $this->_original[$p], $this->_dirty[$p]);
         }
 
--- a/src/Datasource/EntityTrait.php
+++ b/src/Datasource/EntityTrait.php
@@ -440,7 +440,7 @@ trait EntityTrait
     {
         $field = (array)$field;
         foreach ($field as $p) {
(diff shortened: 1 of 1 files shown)


ADmad:
Why? If a property currently has a value and you unset it then it should remain marked as dirty.

ADmad:
Sorry, I see the dirty flag is already removed, I guess in that case the original value should be removed too.

andrii-pukhalevych:
Is there is a way to fix/restart `continuous-integration/appveyor/pr` tests?
It was not started:
`Command "cinst --no-progress --params ""/InstallDir:C:\php"" --ignore-checksums -y php --version" failed with exit code 1. `

garas:
AppVeyor was fixed in `master`, but not yet merged to `3.next`

dereuromark:
You can ignore AppVeyor
But we might want to add a test for this part.

markstory:
Won't this remove the ability to see what a property's value was before it was unset? Unset properties won't change the persisted entity (as the dirty flag is removed) but I could see wanting to know what the previous state was.

andrii-pukhalevych:
@markstory Unsetting property is not equal to set it to `null`. For this method unsetting property without cleanup of original value looks like uncollected garbage.

dereuromark (dismissed):
I find it interesting that
https://book.cakephp.org/3.0/en/orm/entities.html
doesnt even mention this method yet.

andrii-pukhalevych:
@dereuromark Its also used in `EntityTrait::__unset()` implicitly to users

dereuromark:
Well, it would be used by magic invocation of the usage of `unset($entity->prop)` then.

That said:
We have indeed 11 usage of this inside the core code - some of those might actually be intended to only set it to null instead of unsetting.
We have to check this IMO.

One is for example in BaseAuthenticate
```
$result->unsetProperty($passwordField);
```

markstory:
> For this method unsetting property without cleanup of original value looks like uncollected garbage.

Fair enough, but what problems does the current behavior cause? Changing methods in EntityTrait can have far reaching and hard to detect impact, so unless there is a problem being caused by the current behavior, I'm hesitant to change it.

ADmad:
How about we make this change in 4.x and keep 3.x as is?

dereuromark:
@andrii-pukhalevych Did you open a PR against 4.x?

andrii-pukhalevych:
Created [PR for 4.x](https://github.com/cakephp/cakephp/pull/13621) instead

Issue #13071, opened 2019-03-26
Testing a controller action that use the SecurityComponent with unlockedFields fails


alaxos:
This is a (multiple allowed):

* [ ] bug
* [x] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.7.5
* Platform and Target: Linux, PHP 7.0

### What you did
I'm trying to test a controller action that is used to add a new organization. The view contains two input fields (`field1` and `field2`) that are added to the form by using Javascript. There is a validation rule that makes at least one value required for these two fields.

In order to make the posted data pass the SecurityComponent check, I added these two fields into the `unlockedFields`:

_OrganizationsController:_
```
$this->Security->setConfig('unlockedFields', ['field1', field2']);
```

In order to test the `add` action I put some values for `field1` and `field2` as they are required to pass validation.

_OrganizationsControllerTest:_
```
$this->enableCsrfToken();
$this->enableSecurityToken();
$data = [
	'name'   => 'Test organisation',
	'field1' => ['value1'],
	'field2' => ['value2']
];
$this->post('/organizations/add', $data);
```
### What happened
The test gives the following error:
```
Possibly related to Cake\Controller\Exception\AuthSecurityException: "Invalid security debug token." 
Cake\Controller\Component\SecurityComponent->_validatePost
```
### What you expected to happen
The expected result would be a successful test instead of this error.

Please not that if I don't set values for `field1` and `field2`, the Security exception is not raised. But my validation rules make the test fail as the new organization is not created.

markstory:
Unlocked field information is normally part of the submitted form data. Right now the `enableSecurityToken()` helper doesn't have arguments to add unlocked fields though.

dereuromark:
@alaxos You want to make a PR for 3.8 to include this addition for fields?

dereuromark:
PR open.

PR #13483, opened 2019-08-07
Allow testing action with unlocked fields

DBnR1:
Enhancement #13071

This allows testing of actions with unlocked fields

### DIFF
--- a/src/TestSuite/IntegrationTestTrait.php
+++ b/src/TestSuite/IntegrationTestTrait.php
@@ -197,7 +197,7 @@ trait IntegrationTestTrait
     /**
      * List of fields that are excluded from field validation.
      *
-     * @var array
+     * @var string[]
      */
     protected $_unlockedFields = [];
 
@@ -270,7 +270,7 @@ trait IntegrationTestTrait
      * compatible token to be added to request data. This
      * lets you easily test actions protected by SecurityComponent.
      *
-     * @param array $unlockedFields List of fields that are excluded from field validation.
+     * @param string[] $unlockedFields List of fields that are excluded from field validation.
      * @return void
      */
     public function enableSecurityToken($unlockedFields = [])
@@ -283,7 +283,7 @@ trait IntegrationTestTrait
     /**
      * Set list of fields that are excluded from field validation.
      *
-     * @param array $unlockedFields List of fields that are excluded from field validation.
+     * @param string[] $unlockedFields List of fields that are excluded from field validation.
      * @return void
      */
     public function setUnlockedFields($unlockedFields = [])
--- a/src/TestSuite/IntegrationTestTrait.php
+++ b/src/TestSuite/IntegrationTestTrait.php
@@ -195,6 +195,13 @@ trait IntegrationTestTrait
     protected $_cookieEncryptionKey;
 
     /**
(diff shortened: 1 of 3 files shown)


dereuromark:
Looks like you want to rebase this on 3.next

dereuromark:
@markstory Or could we also put this into master? This seems to only affect tests part and seems quite BC.

src/TestSuite/IntegrationTestTrait.php
```diff
@@ -282,6 +282,9 @@ public function enableSecurityToken($unlockedFields = [])
 
     /**
      * Set list of fields that are excluded from field validation.
+     *
+     * @param array $unlockedFields List of fields that are excluded from field validation.
```
dereuromark: Isnt this `string[]` ?

DBnR1: I copied the comment from SecureFieldTokenTrait but you are right. Maybe fix it in both files.

dereuromark: I will take care of it in 4.x: https://github.com/cakephp/cakephp/pull/13485
but you are welcome to fix in 3.x

markstory:
@dereuromark Sure. Would you be able to rebase your changes onto `master`?

saeideng:
please rebase your branch

src/TestSuite/IntegrationTestTrait.php
```diff
@@ -263,13 +270,27 @@ public function configApplication($class, $constructorArgs)
      * compatible token to be added to request data. This
      * lets you easily test actions protected by SecurityComponent.
      *
+     * @param string[] $unlockedFields List of fields that are excluded from field validation.
      * @return void
      */
-    public function enableSecurityToken()
+    public function enableSecurityToken($unlockedFields = [])
```
saeideng: why we need this argument when we can use `$this->setUnlockedFields($unlockedFields);` ??

saeideng:
I rebased changes onto` master`

src/TestSuite/IntegrationTestTrait.php
markstory: Agreed. Having both seems redundant.

DBnR1: A good point. I can't think of a use case where `$this->setUnlockedFields($unlockedFields);` wouldn't do.

src/TestSuite/IntegrationTestTrait.php
```diff
@@ -263,13 +270,27 @@ public function configApplication($class, $constructorArgs)
...
         $this->_securityToken = true;
     }
 
+    /**
+     * Set list of fields that are excluded from field validation.
+     *
+     * @param string[] $unlockedFields List of fields that are excluded from field validation.
+     * @return void
+     */
+    public function setUnlockedFields($unlockedFields = [])
```
ADmad: Not a fan of public API changes in bugfix release even if it only affects testsuite. Let's at least add a `@since` tag to the docblock.

ADmad: The argument should be typehinted as `array`.

PR #13486, opened 2019-08-07
3.x: Backport of "Make metadata cache key configurable"

bancer:
Issue https://github.com/cakephp/cakephp/issues/13471

https://github.com/cakephp/cakephp/pull/13472 4.x backport to 3.x

[edited by dereuromark - please always add these things into description instead of title]

### DIFF
--- a/src/Database/Connection.php
+++ b/src/Database/Connection.php
@@ -107,7 +107,6 @@ class Connection implements ConnectionInterface
     /**
      * Constructor.
      *
-     * @param array $config Configuration array.
      * ### Available options:
      * - `driver` Sort name or FCQN for driver.
      * - `log` Boolean indicating whether to use query logging.
@@ -115,6 +114,8 @@ class Connection implements ConnectionInterface
      * - `cacheMetaData` Boolean indicating whether metadata (datasource schemas) should be cached.
      *    If set to a string it will be used as the name of cache config to use.
      * - `cacheKeyPrefix` Custom prefix to use when generation cache keys. Defaults to connection name.
+     *
+     * @param array $config Configuration array.
      */
     public function __construct($config)
     {
--- a/src/Database/Connection.php
+++ b/src/Database/Connection.php
@@ -107,7 +107,14 @@ class Connection implements ConnectionInterface
     /**
      * Constructor.
      *
-     * @param array $config configuration for connecting to database
+     * @param array $config Configuration array.
+     * ### Available options:
+     * - `driver` Sort name or FCQN for driver.
+     * - `log` Boolean indicating whether to use query logging.
+     * - `name` Connection name.
+     * - `cacheMetaData` Boolean indicating whether metadata (datasource schemas) should be cached.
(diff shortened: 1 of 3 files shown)


src/Database/Connection.php
```diff
@@ -107,7 +107,14 @@ class Connection implements ConnectionInterface
     /**
      * Constructor.
      *
-     * @param array $config configuration for connecting to database
+     * @param array $config Configuration array.
+     * ### Available options:
```
dereuromark: I would prefer those going on top before params start, as we usually do in other places.

src/Database/Schema/CachedCollection.php
```diff
@@ -77,7 +77,12 @@ public function describe($name, array $options = [])
      */
     public function cacheKey($name)
     {
-        return $this->_connection->configName() . '_' . $name;
+        $cachePrefix = $this->_connection->configName();
+        $config = $this->_connection->config();
+        if (isset($config['cacheKeyPrefix'])) {
+            $cachePrefix = $config['cacheKeyPrefix'];
+        }
+        return $cachePrefix . '_' . $name;
```
dereuromark: I would consider a "true" prefix to include the `_` here, no magic internal additions IMO. Or is this how it works in 4.x now as well?

bancer: I do not understand that. It is `return $this->prefix . '_' . $name;` in https://github.com/cakephp/cakephp/blob/4.x/src/Database/Schema/CachedCollection.php#L98.

dereuromark: I see.
I would like to get feedback here from @ADmad @markstory etc on if we can change 4.x here also to make prefixing behave as in other cases - and how intuitively expected (to not magically append another glue char.

ADmad: I am fine with not auto adding `_` if prefix has been explicitly specified if someone wants to update 4.x.

dereuromark: But I see also
- https://github.com/cakephp/cakephp/blob/69643f58dd110a4c5cce7ff867707de176c64bc4/src/View/Helper/IdGeneratorTrait.php#L98

which is in line with this change however. Even though that is for HTML DOM - and not as file, path or cache prefix.

dereuromark: The doc block on 4.x states

 > Typically the connection name.

As such it is probably fine to keep it as is (and auto append the underscore).

markstory: I don't see a problem with the underscore. It is a consistent input to the cache key and makes keys more readable should you need to inspect them.

ADmad:
Even though there's no API change here adding feature flags in bugfix releases makes it difficult to keep track of since when a feature is available.

dereuromark:
Putting it into 3.next then sounds like a fair compromise

dereuromark:
This also needs a documentation update for the migration guide and specific docs on the topic.

PR #13487, opened 2019-08-07
Add default value for enable* methods.

ADmad:


src/Database/Connection.php
```diff
@@ -569,7 +569,7 @@ public function rollback(?bool $toBeginning = null): bool
      * @param bool $enable Whether or not save points should be used.
      * @return $this
      */
-    public function enableSavePoints(bool $enable)
+    public function enableSavePoints(bool $enable = true)
```
dereuromark: Are there interface methods behind we also need to address?

PR #13498, opened 2019-08-09
Set "autoSetCustomValidity" to false by default.

ADmad:
That way it works it's likely to be undesirable by most users.
For e.g. if you have a required email field, by default the browser
will show message like "Please fill out this field" if you leave
the field blank.  If you enter an invalid email string it will show
"Please enter an email".

With the custom validity message feature it always shows same message
"This field cannot be left empty". It takes no account of additional
validations for the field even if set.

P.S. The default message I have mentioned is what Firefox shows me, they are different for Chrome.

### DIFF
--- a/src/View/Helper/FormHelper.php
+++ b/src/View/Helper/FormHelper.php
@@ -146,7 +146,7 @@ class FormHelper extends Helper
             'confirmJs' => '{{confirm}}',
         ],
         // set HTML5 validation message to custom required/empty messages
-        'autoSetCustomValidity' => true,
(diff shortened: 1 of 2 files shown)


ADmad:
Also `EntityContent::getRequiredMessage()` gives priority to `Validator::getRequiredMessage()` message over `Validator::getNotEmptyMessage()`.

The presence required validation would fail when the field isn't present at all in the form (and hence POSTed data), so using that message makes no sense.

ADmad:
Hmm I might have a way to salvage this, closing for now.

PR #13499, opened 2019-08-09
Show custom validity message only when field is empty.

ADmad:
This allows browser to still show its default message
for other validation failures like invalid string for email field.


ADmad:
https://jsfiddle.net/oembq2rw/

ADmad:
Also change the context classes to return "notEmpty" message instead of "requirePresence" message as I explained here https://github.com/cakephp/cakephp/pull/13498#issuecomment-519908619.

src/View/Helper/FormHelper.php
```diff
@@ -1345,7 +1345,7 @@ protected function _magicOptions(string $fieldName, array $options, bool $allowO
 
             if ($this->getConfig('autoSetCustomValidity')) {
                 $options['oninvalid'] = "this.setCustomValidity(''); "
-                    . "if (!this.validity.valid) this.setCustomValidity('$message')";
+                    . "if (!this.value) this.setCustomValidity('$message')";
```
markstory: Will this still take HTML5 pattern validations into account?

tests/TestCase/View/Form/FormContextTest.php
```diff
@@ -53,7 +53,7 @@ public function setUp(): void
     public function testGetRequiredMessage()
     {
         $validator = new Validator();
-        $validator->requirePresence('title', true, 'Don\'t forget a title!');
+        $validator->notEmpty('title', 'Don\'t forget a title!');
```
markstory: Seems like this entire test could be removed if it is for testing requirePresence() interactions with message generation.

src/View/Helper/FormHelper.php
ADmad: I updated the fiddle. As for the email field the custom validity message is shown only when field is empty. When failing to match the pattern it shows bower's default message.

tests/TestCase/View/Form/FormContextTest.php
ADmad: The test is required. Istead of requirePresence() message it does need to check for notEmpty() message. There are no other tests for the same.

PR #13502, opened 2019-08-12
Use data attribute to hold confirm message.

### LINKED ISSUE #13424: 4.x: FormHelper confirm via data-confirm?
This is a (multiple allowed):

* [ ] bug
* [x] enhancement

* CakePHP Version: 3.x/4.x

### What you did
```php
echo $this->Form->postLink(..., ..., ['confirm' => __('Are you sure you want to delete # {0}?', $country->name)]);
```

I wanted to [intercept the form post and do an ajax delete instead](https://sandbox.dereuromark.de/sandbox/ajax-examples/table), re-using the confirm string in a different way here via .ajax().

### What happened
It is inline JS here:
```html
onclick="if (confirm(&quot;Are you sure you want to delete # \u00d6sterreich?&quot;)) { document.post_5d2f83592564f777576700.submit(); } event.returnValue = false; return false;"
```

The core contains
```php
$message = $this->_cleanConfirmMessage($message);
$confirm = "if (confirm({$message})) { {$okCode} } {$cancelCode}";
```

It would be nice if the confirm part would just become a re-usable `data-confirm="Are you sure you want to delete # \u00d6sterreich?"` attribute on the link element and the JS code then would be simpler and directly just used this:

```js
if (confirm(document.{...}.dataset.confirm) ...
```

Then the following code would work:

```js
$('table.list a.postLink').removeAttr('onclick').click(function(e) {
            e.preventDefault();

            var alert = $(this).data('confirm');
            if (!confirm(alert)) {
                return false;
            }
            var form = $(this).prev();
            var url = $(form).attr("action");
            var tr = $(this).closest('tr');
            url = url + '.json';
            $.post(url).success(function(res) {
                if (res.error) {
                    alert(res.error);
                    return false;
                }
                tr.fadeOut(200);
            }).error(function() {
                alert("Error");
            });
            return false;
        });
```

Currently, I would have to duplicate this in the Form->postLink() call as separate data attribute for it.

What do people think?

ADmad:
Closes #13424.


garas:
This could be moved to `data-` attribute too:

https://github.com/cakephp/cakephp/blob/3ec66ac7370e1046b0842499cf8a3edde4094995/src/View/Helper/FormHelper.php#L1417

ADmad:
While working on this I realized `button()` also needs 2 separate options:`escapeTitle` and `escape`, to handle escaping of button text and it's attributes, like `link()` method does.

Currently it only uses `escape` which defaults to `false` and hence any extra attributes you set for button tag are not escape which is bad.

So should I included the change in PR to make it similar to `link()` with button text no escaped by default (as current) and any other attributes escaped by default?

markstory:
I think aligning the escape options for button() and link() is a good idea.

ADmad:
Add `escapeTitle` for `button()`. The title is still unescaped by default to maintain BC. Other attributes are now escaped by default.

dereuromark:
Hmm shouldnt we use this change to make the title escaped by default? It sounds like there could be a lot of accidental XSS out there, as many might not be aware of this.
It only breaks for HTML cases, and there it still works (pure HTML output then) in a way, so upgrading should be doable here without too much hazzle, right?

ADmad:
I am fine with button text also being escaped by default. Safer defaults are better.

PR #13503, opened 2019-08-12
Add SerializationFailureException.

ADmad:
Throwing a specific exception instead of `RuntimeException` is better.

### DIFF
--- a/src/View/JsonView.php
+++ b/src/View/JsonView.php
@@ -145,6 +145,10 @@ class JsonView extends SerializedView
             $jsonOptions |= JSON_PRETTY_PRINT;
         }
 
+        if (defined('JSON_THROW_ON_ERROR')) {
+            $jsonOptions |= JSON_THROW_ON_ERROR;
+        }
+
         $return = json_encode($data, $jsonOptions);
         if ($return === false) {
             throw new RuntimeException(json_last_error_msg(), json_last_error());
--- a/src/View/JsonView.php
+++ b/src/View/JsonView.php
@@ -147,7 +147,7 @@ class JsonView extends SerializedView
 
         $return = json_encode($data, $jsonOptions);
         if ($return === false) {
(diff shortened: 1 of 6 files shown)


src/View/XmlView.php
```diff
@@ -151,10 +148,6 @@ protected function _serialize($serialize)
             $options['pretty'] = true;
         }
 
-        if (isset($options['return']) && strtolower($options['return']) === 'domdocument') {
```
ADmad: Both `DOMDocument` and `SimpleXMLElement` have `saveXML()` method, so this check was redundant.

ADmad:
By default `JsonView`' uses `JSON_PARTIAL_OUTPUT_ON_ERROR` option for `json_encode()`. Personally I don't think it's a good default and devs should set it explicitly.

src/View/JsonView.php
```diff
@@ -147,7 +145,12 @@ protected function _serialize($serialize)
             $jsonOptions |= JSON_PRETTY_PRINT;
         }
 
-        return json_encode($data, $jsonOptions);
+        $return = json_encode($data, $jsonOptions);
+        if ($return === false) {
+            throw new RuntimeException(json_last_error_msg(), json_last_error());
```
markstory: Should this be the new exception type? It would save you a try/catch in the parent.

ADmad: The try/catch in parent would still be needed to catch potential XmlException from XmlView and throwing SerializationFailureException instead.

Throwing this RuntimeException is necessary only because JsonException is available since PHP 7.3 :(

ADmad: The try/catch in `SerializedView` is also necessary for non core view classes extending `SerializedView`, for .e.g. `CsvView` of `friendsofcake/cakephp-csvview`.

markstory:
Looks good, but tests are failing.

ADmad:
Tests fixed.

PR #13504, opened 2019-08-12
Update doblock / typehints.

ADmad:


src/Utility/Hash.php
```diff
@@ -1181,7 +1181,7 @@ public static function normalize(array $data, bool $assoc = true): array
      *
      * @param array $data The data to nest.
      * @param array $options Options are:
-     * @return array of results, nested
+     * @return array[] of results, nested
```
dereuromark: Is that helpful? Usually, array is already sufficient here.

PR #13506, opened 2019-08-13
Throw exception if context cannot be generated for given entity.

ADmad:
Currently if you pass `false` or `string` value as 1st argument to `FormHelper::create()` it gets silently ignored. Primarily apps upgraded from 2.x (and the core itself) are often littered with such invalid usage.

So I made this change to throw an exception if `ContextFactory` is unable create a context for provided entity.


markstory:
Looks good. This needs updates in the book around the breaking change and to update any examples.

ADmad:
I'll make a docs PR soon.

PR #13515, opened 2019-08-14
Revised ExtractTask::_extractTokens() to delegate pass FormHelper::in…

trenshaw:
…put(), ::radio() and PaginatorHelper::sort() as array callables to ExtractTask::_parse()

Revised ExtractTask::_parse() to check for and match T_OBJECT_OPERATOR tokens against passed array callables skipping any 'label' or 'legend' strings passed in $options array (in either full or short array syntax) passing labels that would be generated by these helpers to new protected ::_formatLabel() before passing to ::_addTranslation()

Fixes # 13303


trenshaw:
@todo fix Travis-CI failure for PHP 7

Issue #13303, opened 2019-05-30
I18N ExtractTask does not extract those strings generated and passed to Gettext through core helpers


trenshaw:
* [ ] bug
* [x] enhancement
* [x] feature-discussion (RFC)

* CakePHP Version: 2.10.18
* Platform and Target: Nginx 1.14.0, PHP 7.2.17, MySQL 5.7.26

### Current functionality
FormHelper::input(), ::radio() and Paginator::sort() all use Inflector to generate strings for HTML output and pass this output to Gettext.

However, ExtractTask::_extractTokens() currently only calls ExtractTask::_parse() to process user-defined strings through the __?() functions and not those generated and passed to Gettext through its core helpers.  

### Proposed enhancement
Form::input(), Form::radio() and Paginator::sort() should be passed in as an array to the $functionName argument when calling ExtractTask::_parse() (similar to call_user_func() when passing a class method) and ExtractTask::_parse() statements added to ExtractTask::_extractTokens() to process these.

ExtractTask::_parse() should have a if() check for token of type T_OBJECT_OPERATOR and if the token being parsed matches the $functionName an additional if() check be performed for an array of options and if there is no match for a 'label' or 'legend' option pass the token string as $fieldName argument to a new protected ::_formatLabel() to split and humanize the token string (e.g. a Model.Field token string) before passing to ::_addTranslation()

FormHelper::label() currently performs similar functionality to the proposed ::_formatLabel() - it passes the strings it generates to Gettext but these strings are never extracted which this proposed enhancement will address.

### Enhancement as plugin
The proposed enhancement extends the core I18nShell and ExtractTask. Attached is a pluginized version that we use across all our projects / CakePHP apps. This alleviates the need to define a 'label' or 'legend' option on every FormHelper::input() or ::radio() just to pass the same string as the default inflected field name to __() when the core helpers already do this.

[I18n.zip](https://github.com/cakephp/cakephp/files/3234866/I18n.zip)

markstory:
Thank you for sharing this. We're not adding any new features to 2.x at this point in time though. Have you considered publishing your changes as a plugin so that others could use it? Alternatively it could be adapted to 3.x and added into the existing `I18nShell` code.

trenshaw:
[...]

PR #13516, opened 2019-08-14
Issue 13303

trenshaw:
Fixes issue #13303 with ExtractTask not extracting those strings generated and passed to Gettext through core helpers.

### DIFF
--- a/lib/Cake/Console/Command/Task/ExtractTask.php
+++ b/lib/Cake/Console/Command/Task/ExtractTask.php
@@ -375,6 +375,10 @@ class ExtractTask extends AppShell {
 				}
 			}
 			unset($allTokens);
+			$this->_parse(array('Form', 'input'), array('singular'));
+			$this->_parse(array('Form', 'radio'), array('singular'));
+			$this->_parse(array('Paginator', 'sort'), array('singular'));
+			
 			$this->_parse('__', array('singular'));
 			$this->_parse('__n', array('singular', 'plural'));
 			$this->_parse('__d', array('domain', 'singular'));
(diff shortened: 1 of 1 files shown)


trenshaw:
Missed milestone for 2.10.19 please include for 2.10.20

ADmad:
As already mentioned here https://github.com/cakephp/cakephp/issues/13303#issuecomment-497174772 we are no longer accepting feature enhancements for 2.x.

You are welcome to open a pull request for `3.next` instead.

trenshaw:
Happy to put into 3.next but this defect / oversight / shortcoming has been in the I18NShell since the 2.x branch was created.

It should be back ported to 2.10.20 at least to ease in migration to 3.x. Migrating to 3.x is major enough without having to deal with missed / untranslated strings from the core helpers.

Many users likely added unnecessary 'label' and/or 'legend' options to every form element just to wrap the same string as the core helper would generate through __() or worse attempt to manually track core helper generated labels and add these one by one to their POTs.

ADmad:
We have supported 2.x since October 2011 and have [clearly communicated](https://bakery.cakephp.org/2017/07/22/cakephp_2100_released.html) that it will receive only bug and security fixes from June 2017 going forward. 

We can't keep adding new features to 2.x and increase our maintenance overhead. We appreciate your participation in the CakePHP project, but as a small group of volunteers contributing their spare time for free there's only so much we can do, and at the moment we are focused on maintaining 3.x and moving towards 4.x.

You can publish your enhanced shell as a plugin and those upgrading from 2.x could use that. 

Once your enhancements are merged into 3.next you can also see if @dereuromark is interested in incorporating those changes in his [shim](https://github.com/dereuromark/cakephp-shim) which is meant to help those upgrading between major CakePHP versions.

PR #13517, opened 2019-08-14
Revised ExtractTask::_extractTokens() to delegate pass FormHelper::in…

### LINKED ISSUE #13303: I18N ExtractTask does not extract those strings generated and passed to Gettext through core helpers

trenshaw:
…put(), ::radio() and PaginatorHelper::sort() as array callables to ExtractTask::_parse()

Revised ExtractTask::_parse() to check for and match T_OBJECT_OPERATOR tokens against passed array callables skipping any 'label' or 'legend' strings passed in $options array (in either full or short array syntax) passing labels that would be generated by these helpers to new protected ::_formatLabel() before passing to ::_addTranslation()

Fixes #13303 pull requested for 3.next and 2.10.20 milestone

### DIFF
--- a/src/Shell/Task/ExtractTask.php
+++ b/src/Shell/Task/ExtractTask.php
@@ -385,6 +385,9 @@ class ExtractTask extends Shell
                 }
             }
             unset($allTokens);
+            $this->_parse(array('Form', 'input'), array('singular'));
+            $this->_parse(array('Form', 'radio'), array('singular'));
+            $this->_parse(array('Paginator', 'sort'), array('singular'));
             $this->_parse('__', ['singular']);
             $this->_parse('__n', ['singular', 'plural']);
             $this->_parse('__d', ['domain', 'singular']);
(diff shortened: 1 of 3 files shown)


ADmad:
Merge conflicts need to be resolved. It would be easier if you add your changes on top of current `3.next` HEAD.

ADmad:
Tests also required.

trenshaw:
Sorry I can't do the tests for 3.x - we are not there yet! Migration of our apps to 3.x won't start until 2020. Hopefully someone else can contrib the tests?

dereuromark:
@trenshaw You can give us some of the code snippets that would now be parsed.
That could help for someone who is able to write a few tests on top here.

trenshaw:
@dereuromark happy to .. will do!

trenshaw:
@dereuromark just saw 3.8.6 release .. any chance we can get the I18n enhancements in for 3.8.7? I don't think I have write access to resolve the conflict?

othercorey:
@trenshaw Your branch is very old.  You need to merge from upstream for this conflict to resolve.

PR #13523, opened 2019-08-15
Document behavior of true/false values on UUIDs

Spriz:
https://github.com/UseMuffin/Footprint changed from using `false` to using `null` in case of no Footprint existing in the ORM calls - which made me check what happened before when those `false` values were moving on towards the UUID type in our db - and I found out that they actually ended up being `''` in case of `false`.

I do not know if this is desired behavior (I would have expected it to be `null` rather than `''`) - but in any case I went to the tests to check for the behavior only to find it was not covered explicitly.

Feel free to close the issue if this is obvious to everyone else, or if the behavior should rather be changed. (changing https://github.com/cakephp/cakephp/blob/master/src/Database/Type/UuidType.php#L35 to include `|| is_boolean($value)` or `|| $value`

### DIFF
--- a/tests/TestCase/Database/Type/UuidTypeTest.php
+++ b/tests/TestCase/Database/Type/UuidTypeTest.php
@@ -79,6 +79,12 @@ class UuidTypeTest extends TestCase
 
         $result = $this->type->toDatabase('', $this->driver);
         $this->assertNull($result);
+        
+        $result = $this->type->toDatabase(true, $this->driver);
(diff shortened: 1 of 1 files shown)


ADmad:
Having booleans cast to string when saving to db is a bit wacky. We can probably fix that in `3.next`.

ADmad:
Fixed in #13551.

Spriz:
Much love from here @ADmad

PR #13525, opened 2019-08-16
4.x cache cleanup

ADmad:
Increased code coverage a bit.

### DIFF
--- a/src/Cache/CacheEngine.php
+++ b/src/Cache/CacheEngine.php
@@ -306,10 +306,7 @@ abstract class CacheEngine implements CacheInterface, CacheEngineInterface
      * @param string $group name of the group to be cleared
      * @return bool
      */
-    public function clearGroup(string $group): bool
-    {
-        return false;
-    }
+    abstract public function clearGroup(string $group): bool;
 
     /**
      * Does whatever initialization for each group is required
--- a/src/Cache/CacheEngine.php
+++ b/src/Cache/CacheEngine.php
@@ -165,7 +165,6 @@ abstract class CacheEngine implements CacheInterface, CacheEngineInterface
             $this->setConfig('duration', $ttl);
         }
         try {
-            $return = [];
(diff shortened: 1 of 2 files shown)


ADmad:
I noticed that none of the engine classes use the specific methods provided by their corresponding engines to get / set multiple cache keys, why is that? All cache use the implementation in `CacheEngine::setMultiple()` and `CacheEngine::getMultiple()` which simply loops over the methods for getting/setting single key.

src/Cache/CacheEngine.php
```diff
@@ -307,10 +306,7 @@ public function add(string $key, $value): bool
      * @param string $group name of the group to be cleared
      * @return bool
      */
-    public function clearGroup(string $group): bool
-    {
-        return false;
-    }
+    abstract public function clearGroup(string $group): bool;
```
ADmad: All engines have their own implementation for this method so this implementation was never used.

tests/TestCase/Cache/Engine/FileEngineTest.php
```diff
@@ -189,11 +189,13 @@ public function testSetWithTtl()
         $this->assertTrue($engine->set('default_ttl', $data));
         $this->assertTrue($engine->set('int_ttl', $data, 1));
         $this->assertTrue($engine->set('interval_ttl', $data, new DateInterval('PT1S')));
+        $this->assertTrue($engine->setMultiple(['multi' => $data], 1));
```
ADmad: `CacheEngine::setMultiple()` with non-null `$ttl` didn't have any test earlier.

markstory:
> All classes use the implementation in CacheEngine::setMultiple() and CacheEngine::getMultiple() which simply loops over the methods for getting/setting single key.

Probably an oversight. We could fix that though and use multi-set where available.

ADmad:
Okay, merging this. We can deal with the multi-methods in a separate PR.

PR #13529, opened 2019-08-16
Update @param tag for MiddlewareQueue method.

### LINKED ISSUE #13528: Add MiddlewareInterface to MiddlewareQueue's @param hints
This is a (multiple allowed):

* [ ] bug
* [x] enhancement

* CakePHP Version: 4.x.

### What you did
`$middlewareQueue->add();`

### What happened
Psalm reports an error when adding a Middleware that implements MiddlewareInterface to the MiddlewareQueue.

Psalm should not report an error when using MiddlewareQueue's `add`, `push`, etc. methods.

That's easily fixable by add `MiddlewareInterface` to the @param hints in `MiddlewareQueue`.

ADmad:
Closes #13528.

### DIFF
--- a/src/Http/MiddlewareQueue.php
+++ b/src/Http/MiddlewareQueue.php
@@ -61,7 +61,7 @@ class MiddlewareQueue implements Countable, SeekableIterator
     /**
      * Resolve middleware name to a PSR 15 compliant middleware instance.
      *
-     * @param string|callable $middleware The middleware to resolve.
+     * @param string|callable|\Psr\Http\Server\MiddlewareInterface $middleware The middleware to resolve.
      * @return \Psr\Http\Server\MiddlewareInterface
      * @throws \RuntimeException If Middleware not found.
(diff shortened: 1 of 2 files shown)


src/Http/MiddlewareQueue.php
```diff
@@ -97,7 +97,7 @@ protected function resolve($middleware): MiddlewareInterface
     /**
      * Append a middleware to the end of the queue.
      *
-     * @param callable|string|array $middleware The middleware(s) to append.
+     * @param callable|string|array|\Psr\Http\Server\MiddlewareInterface $middleware The middleware(s) to append.
```
ADmad: This is one type too many. Should be remove the ability to pass array of middlewares? I don't see it overly useful. At least I didn't even remember that `add()` and other methods allow passing array.

ADmad:
Merging. We can look into not allowing arrays separately.

PR #13530, opened 2019-08-16
4.x mail

ADmad:
- Remove exception if `From` is not set. It's not mandatory for sending mails.
- Moved check that at least one recipient/destination is set to transport. IMO that's the concern of transport not Email class. This allows possibility to use custom transport for services where you can specify pre-made recipient lists without having to set a dummy `To` in email/message.
- Made view builder instance injectable through `Mailer\Renderer`'s constructor.

### DIFF
--- a/src/Mailer/AbstractTransport.php
+++ b/src/Mailer/AbstractTransport.php
@@ -60,9 +60,9 @@ abstract class AbstractTransport
      */
     protected function checkRecipient(Message $message): void
     {
-        if (empty($message->getTo())
-            && empty($message->getCc())
-            && empty($message->getBcc())
+        if ($message->getTo() === []
+            && $message->getCc() === []
+            && $message->getBcc() === []
         ) {
             throw new Exception('You must specify at least one recipient of to, cc or bcc.');
         }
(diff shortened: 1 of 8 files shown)


src/Mailer/AbstractTransport.php
```diff
@@ -50,6 +51,23 @@ public function __construct(array $config = [])
...
+    /**
+     * Check that at least one destination header is set.
+     *
+     * @param \Cake\Mailer\Message $message
+     * @return void
+     * @throws \Cake\Core\Exception\Exception If at least one of to, cc or bcc is not specified.
+     */
+    protected function checkRecipient(Message $message): void
+    {
+        if (empty($message->getTo())
```
dereuromark: We shouldnt use empty() on methods. Can we avoid that here please?

src/Mailer/AbstractTransport.php
```diff
@@ -50,6 +51,23 @@ public function __construct(array $config = [])
...
+     * @return void
+     * @throws \Cake\Core\Exception\Exception If at least one of to, cc or bcc is not specified.
+     */
+    protected function checkRecipient(Message $message): void
+    {
+        if ($message->getTo() === []
+            && $message->getCc() === []
+            && $message->getBcc() === []
+        ) {
+            throw new Exception('You must specify at least one recipient of to, cc or bcc.');
```
markstory:
```suggestion
            throw new Exception('You must specify at least one recipient. Use one of `setTo`, `setCc` or `setBcc` to define a recipient.');
```

PR #13533, opened 2019-08-16
4.x completion

saeideng:
before:
```php
$ bin/cake completion options i18n extract

--help -h --quiet -q --verbose -v //result
```
after:
```php
$bin/cake completion options i18n extract

--app --exclude --exclude-plugins --extract-core --files --help -h --marker-error --merge --no-location --output --overwrite --paths --plugin --quiet -q --relative-paths --verbose -v
```

### DIFF
--- a/tests/TestCase/Command/CompletionCommandTest.php
+++ b/tests/TestCase/Command/CompletionCommandTest.php
@@ -192,6 +192,24 @@ class CompletionCommandTest extends ConsoleIntegrationTestCase
     }
 
     /**
+     * test that nested command returns subcommand's options not command.
+     *
+     * @return void
+     */
+    public function testOptionsNestedCommand()
+    {
+        $this->exec('completion options i18n extract');
+        $this->assertExitCode(Shell::CODE_SUCCESS);
+
+        $expected = [
+            '--plugin',
+            '--app',
+        ];
+        foreach ($expected as $value) {
+            $this->assertOutputContains($value);
+        }
+    }
+    /**
      * test that subCommands with a existing CORE command returns the proper sub commands
      *
      * @return void
--- a/tests/TestCase/Command/CompletionCommandTest.php
+++ b/tests/TestCase/Command/CompletionCommandTest.php
@@ -157,6 +157,10 @@ class CompletionCommandTest extends ConsoleIntegrationTestCase
      */
     public function testOptionsShellTask()
     {
+        //details: https://github.com/cakephp/cakephp/pull/13533
+        $this->markTestIncomplete(
+            'This test does not work correctly with shells.'
+        );
         $this->exec('completion options sample sample');
         $this->assertExitCode(Shell::CODE_SUCCESS);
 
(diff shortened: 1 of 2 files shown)


markstory:
This caused an existing test to fail. Is the older test invalid?

saeideng:
for a shell like this
```php
<?php
namespace App\Shell;
use Cake\Console\Shell;
class SampleShell extends Shell
{
    public function main()
    {
        $this->out('This is the main method called from SampleShell');
    }
    public function example()
    {
        $this->out('This is the example method called from TestPlugin.SampleShell');
    }
}
```
cake 3.7
```php
$ bin/cake completion options sample
--help -h --quiet -q --verbose -v

$ bin/cake completion options sample example

// empty
```
cake 4.x before this change
```php
$ bin/cake completion options sample
--help -h --quiet -q --verbose -v

$ bin/cake completion options sample example
--help -h --quiet -q --verbose -v
```

after this change with cake 4.x
```php
$ bin/cake completion options sample example

// empty
```

saeideng:
also for this shell ( cake 3.x )
https://github.com/cakephp/cakephp/blob/cfc3a961251a218805834acf924d6baf270a73d3/tests/test_app/Plugin/TestPlugin/src/Shell/SampleShell.php

I dont know why this tests 
`$this->Shell->runCommand(['options', 'sample', 'sample']);`

and not
`$this->Shell->runCommand(['options', 'sample', 'example']);`
see
https://github.com/cakephp/cakephp/blob/cfc3a961251a218805834acf924d6baf270a73d3/tests/TestCase/Shell/CompletionShellTest.php#L172

markstory:
The `sample example` case should show the full option list though. The 'subcommands' in a shell inherit the shell's option parser unless they have an explicit one defined. While not a great system that is how it has worked for a while. Is there a way we can traverse into the explict parser if it exists or use the shell's parser if the subcommand has no parser?

saeideng:
> Is there a way we can traverse into the explict parser if it exists or use the shell's parser if the subcommand has no parser?

I think this occur  when subcommand is in main shell together , not in a task class

I should test about this 

should I fix it in this PR ?

markstory:
> should I fix it in this PR ?

Don't have to, but we should mark the currently failing test with an incomplete marker.

saeideng:
ready to merge 
the https://github.com/cakephp/migrations/pull/399 needs this PR to be merge

tests/TestCase/Command/CompletionCommandTest.php
```diff
@@ -191,6 +195,24 @@ public function testOptionsShellCommand()
...
 
+    /**
+     * test that nested command returns subcommand's options not command.
+     *
+     * @return void
+     */
+    public function testOptionsNestedCommand()
+    {
+        $this->exec('completion options i18n extract');
+        $this->assertExitCode(Shell::CODE_SUCCESS);
```
dereuromark: Shouldnt those be Command:: constants?

Issue #13535, opened 2019-08-17
4.x - Combine Mailer and Email classes.


ADmad:
With the creation of `Renderer` and `Message` classes `Email` is now just a glue class which uses the two to send email using a `Transport`. `Mailer` is a higher level convenience class which forwards most of the calls to `Email` class using `__call()` anyway.

So having both `Email` and `Mailer` seem unnecessary. The responsibility of both can be handled by a single class. If we move/add function for managing profiles and transport to `Mailer` then it can directly send mails using `Message`, `Renderer` and `Transport`, making `Email` redundant which can be marked as deprecated.

ADmad:
Currently calling `Mailer::viewBuilder()` calls `Email::viewBuilder()` which calls `Renderer::viewBuilder()`. That's some crazy misdirection :slightly_smiling_face:.

markstory:
What would the migration path be for user land code?

ADmad:
I haven't worked out the details yet but there would be minimal BC break if at all.

markstory:
Might be worth trying to outline the migration/upgrade path in a wiki page/issue first.

ADmad:
I already have a PR up :)

PR #13536, opened 2019-08-18
4.x mailer

ADmad:
Refs #13535

### Notable change to `Mailer`:

- It's constructor now takes config array / profile name instead of email instance as argument.
- It's no longer abstract and one can use the new `deliver()` method (non-static unlike `Email::deliver()`) to send email without using a mailer action.

### Todo:
Reverting back would have been more work and would lose the typehinting so I didn't.
- [x] `Email::setConfig()`/`getConfig()` need to be overridden to call `Mailer::setConfig()`/`getConfig()` internally.
- [x] Remove `Mailer::getName()`. I don't think it serves any purpose. `MailerAwareTrait` uses `App::className()` to extrapolate short name to FQCN.
- [x] ~~Since `Email` is deprecated I think it would be best to revert all it's internal changes and make it same as in 3.x.~~ 
This would have been an unnecessary chore and would have to redo all the typehinting work.
- [x] ~~Make `Mailer` implement `JsonSerializable`, `Serializable`. Do people use this feature?~~
Given that the `Message` class and `Renderer`'s `ViewBuilder` are serializable making `Mailer` serializable is not necessary. Earlier when `Email` was made serializable it was a "god" class.

### DIFF
--- a/src/Mailer/Mailer.php
+++ b/src/Mailer/Mailer.php
@@ -14,10 +14,13 @@ declare(strict_types=1);
  */
 namespace Cake\Mailer;
 
+use Cake\Core\Exception\Exception;
+use Cake\Core\StaticConfigTrait;
 use Cake\Datasource\ModelAwareTrait;
 use Cake\Event\EventListenerInterface;
 use Cake\Mailer\Exception\MissingActionException;
 use Cake\View\ViewBuilder;
+use InvalidArgumentException;
 
 /**
  * Mailer base class.
@@ -132,6 +135,7 @@ use Cake\View\ViewBuilder;
 abstract class Mailer implements EventListenerInterface
 {
     use ModelAwareTrait;
+    use StaticConfigTrait;
 
     /**
      * Mailer's name.
@@ -156,18 +160,53 @@ abstract class Mailer implements EventListenerInterface
     protected $_clonedEmail;
 
     /**
-     * Constructor.
+     * The transport instance to use for sending mail.
      *
-     * @param \Cake\Mailer\Email|null $email Email instance.
+     * @var \Cake\Mailer\AbstractTransport|null
      */
-    public function __construct(?Email $email = null)
+    protected $transport;
+
+    /**
+     * Message class name.
+     *
+     * @var string
+     */
+    protected $messageClass = Message::class;
+
+    /**
+     * Message instance.
+     *
+     * @var \Cake\Mailer\Message
+     */
+    protected $message;
+
+    /**
+     * Constructor
+     *
+     * @param array|string|null $config Array of configs, or string to load configs from app.php
+     */
+    public function __construct($config = null)
     {
-        if ($email === null) {
-            $email = new Email();
-        }
+        $email = new Email();
 
         $this->_email = $email;
         $this->_clonedEmail = clone $email;
+
+        $this->message = new $this->messageClass();
+
+        if ($config === null) {
+            $config = static::getConfig('default');
+        }
+
+        $this->viewBuilder()
+            ->setClassName(View::class)
+            ->setTemplate('')
+            ->setLayout('default')
+            ->setHelpers(['Html']);
+
+        if ($config) {
+            $this->setProfile($config);
+        }
     }
 
     /**
@@ -199,7 +238,7 @@ abstract class Mailer implements EventListenerInterface
     }
 
     /**
-     * Magic method to forward method class to Email instance.
+     * Magic method to forward method class to Message instance.
      *
      * @param string $method Method name.
      * @param array $args Method arguments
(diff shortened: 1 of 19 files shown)


ADmad:
- Since `Email` is deprecated I think it would be best to revert all it's internal changes and make it same as in `3.x`.
- `Email::setConfig()/getConfig()` need to be overridden to call `Mailer::setConfig()/getConfig()` internally.

ADmad:
Notable change to `Mailer`:

- It's constructor now takes config array / profile name instead of email instance as argument.
- It's no longer abstract and one can use the new (non-static unlike `Email`) `deliver()` method to send email without using a mailer action.

ADmad:
Any feedback on this?

src/Mailer/Mailer.php
```diff
@@ -244,34 +313,189 @@ public function send(string $action, array $args = [], array $headers = []): arr
...
+     *
+     * @return $this
+     */
+    protected function restore()
+    {
+        foreach (array_keys($this->clonedInstances) as $key) {
+            if ($this->clonedInstances[$key] === null) {
+                $this->{$key} = null;
+            } else {
+                $this->{$key} = clone $this->clonedInstances[$key];
```
markstory: Should the clonedInstances be cleared after all properties have been restored?

ADmad: Done

src/Mailer/Mailer.php
```diff
@@ -322,22 +340,75 @@ public function send(string $action, array $args = [], array $headers = []): arr
...
+     */
+    public function deliver(string $content)
+    {
+        $this->render($content);
+
+        return $this->getTransport()->send($this->message);
+    }
+
+    /**
+     * Set message body.
```
markstory: Should mention that setting a body will disable automatic rendering of the template.

ADmad: Done

ADmad:
I copied relevant tests from `EmailTest` to `MailerTest` and `MessageTest` so that test coverage isn't lost when `Email` and `EmailTest` are removed in future.

src/Mailer/Mailer.php
```diff
@@ -224,58 +290,351 @@ public function __call(string $method, array $args)
...
+     *
+     * Setting body content explicity will disable auto rendering.
+     *
+     * @param array $content Content array with keys "text" and/or "html" with
+     *   content string of respective type.
+     * @return $this
+     */
+    public function setBody(array $content)
```
markstory: How do you see these new methods being used? Are they a longer term replacement for Email features? To me the separation between Mailer and Email was that Mailer applied more conventions and had more batteries-included for tasks like rendering templates. Having lower-level setters makes the purpose of each class a bit more murky.

ADmad: You are right, having this lower level method on `Mailer` does muddle things.

I added the method for setting message body mainly to control the auto rendering as I wanted a user to be able to directly set text/HTML content for the message and skip using `Renderer`. But I guess that's not necessary. A user can directly use `Message` and `Transport` in that case.

I have removed these methods.

ADmad:
@markstory The stripping isn't related to user input sanitizing (if so I would be required in SMTP too). It's related to this issue https://github.com/cakephp/cakephp/issues/2209, `\r\n` added by `mb_encode_mimeheader()` need to be removed as it causes problems with qmail.

As per details provided in that issue I updated the code to strip out `\r\n` pairs not `\r` and `\n` individually.

ADmad:
This should be good to merge after a final review. The static analysis job failure is unrelated and fixed on `4.x`.

markstory (commented):
Should we add deprecation warnings to the static Email methods and Email constructor?

src/Mailer/Message.php
```diff
@@ -1179,20 +1211,12 @@ public function addAttachments(array $attachments)
...
     /**
-     * Get generated message body.
+     * Get generated message body as array.
      *
-     * @param string|null $type Use MESSAGE_* constants or null to return the full message as array
-     * @return string|array String if type is given, array if type is null
+     * @return array
      */
-    public function getBody(?string $type = null)
+    public function getBody()
```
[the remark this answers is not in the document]
ADmad: Message class doesn't exist in 3.x :)

tests/TestCase/Mailer/MailerTest.php
```diff
@@ -14,85 +14,383 @@
...
+     * test that default profile is used by constructor if available.
+     *
+     * @return void
+     */
+    public function testDefaultProfile()
+    {
+        $config = ['to' => 'foo@bar.com', 'from' => 'from@bar.com'];
+
+        Configure::write('Mailer.default', $config);
+        Mailer::setConfig(Configure::consume('Mailer'));
```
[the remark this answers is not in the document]
ADmad: Yes, but to avoid BC breakage I have added Email::setConfig()/getConfig() which proxies calls to Mailer::setConfig()/getConfig(). Same for all methods provided by StaticConfigTrait.

ADmad (commented):
> Should we add deprecation warnings to the static Email methods and Email constructor?

I have marked the class as deprecated. We haven't added explicit warnings to other classes/methods which are deprecated in 4.0 and will be retained till 5.0.

I'll submit manual update to replace Email usage examples with Mailer.

PR #13537, opened 2019-08-19
Wrap queries in parentheses when passed as function expression argument.

ndm2:
Queries passed as function arguments need to be wrapped in parentheses.

refs #11410

### DIFF
--- a/src/Database/Expression/FunctionExpression.php
+++ b/src/Database/Expression/FunctionExpression.php
@@ -15,6 +15,7 @@
 namespace Cake\Database\Expression;
 
 use Cake\Database\ExpressionInterface;
+use Cake\Database\Query;
 use Cake\Database\TypedResultInterface;
 use Cake\Database\TypedResultTrait;
 use Cake\Database\Type\ExpressionTypeCasterTrait;
@@ -171,8 +172,10 @@ class FunctionExpression extends QueryExpression implements TypedResultInterface
     {
         $parts = [];
         foreach ($this->_conditions as $condition) {
-            if ($condition instanceof ExpressionInterface) {
-                $condition = sprintf('%s', $condition->sql($generator));
+            if ($condition instanceof Query) {
+                $condition = sprintf('(%s)', $condition->sql($generator));
+            } elseif ($condition instanceof ExpressionInterface) {
+                $condition = $condition->sql($generator);
             } elseif (is_array($condition)) {
                 $p = $generator->placeholder('param');
                 $generator->bind($p, $condition['value'], $condition['type']);
--- a/tests/TestCase/Database/Expression/FunctionExpressionTest.php
+++ b/tests/TestCase/Database/Expression/FunctionExpressionTest.php
@@ -16,6 +16,7 @@ namespace Cake\Test\TestCase\Database\Expression;
 use Cake\Database\Expression\FunctionExpression;
 use Cake\Database\Expression\QueryExpression;
 use Cake\Database\ValueBinder;
+use Cake\Datasource\ConnectionManager;
 use Cake\TestSuite\TestCase;
 
 /**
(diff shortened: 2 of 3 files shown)


tests/TestCase/Database/Expression/FunctionExpressionTest.php
```diff
@@ -96,6 +97,45 @@ public function testFunctionNestingQueryExpression()
...
+            ->setSchema(['column' => 'integer'])
+            ->find()
+            ->select(['column']);
+
+        $binder = new ValueBinder;
+        $function = new FunctionExpression('MyFunction', [$query]);
+        $this->assertEquals(
+            'MyFunction((SELECT Articles.column AS Articles__column FROM articles Articles))',
+            preg_replace('/[`"\[\]]/', '', $function->sql($binder))
+        );
```
markstory: It would be good to have a test that hits the database as well to ensure that we aren't introducing incompatible syntax with any of our supported DB platforms.

src/Database/Expression/FunctionExpression.php
```diff
@@ -171,8 +172,10 @@ public function sql(ValueBinder $generator)
     {
         $parts = [];
         foreach ($this->_conditions as $condition) {
-            if ($condition instanceof ExpressionInterface) {
-                $condition = sprintf('%s', $condition->sql($generator));
+            if ($condition instanceof Query) {
```
raul338: Maybe it should be QueryInterface?

tests/TestCase/Database/Expression/FunctionExpressionTest.php
lorenzo: I agree, we should have several functional test for this change. I having something like this in the early days of QueryExpression, and it turned out to be a problem because some database servers did not like the extra parens

ndm2: Sure, I can add tests for that, but I'm pretty positive that this is not only valid, but required in all supported dialects, even in unsupported ones like Oracle or Firebase.

src/Database/Expression/FunctionExpression.php
ndm2: That would unfortunately only cover ORM queries, as the base query class from the `\Cake\Database` namespace doesn't implement `QueryInterface`.

ndm2: Also `QueryInterface` doesn't even have an `sql()` method :upside_down_face:

ndm2:
Added some tests (didn't really know where to best put them). Their usefulness is highly questionable in terms of real world scenarios that actually query useful data, I've dumbed things down to make them simple, and to use functions that are available in all dialects, and functions that are processed by the transpilers.

I don't really know what else to test for, other than checking that subqueries do work with parentheses, respectively that they do not work without parentheses. The 3 scenarios are single argument functions, multi argument functions, and transpiled functions, each with regular subqueries, and correlated subqueries (though I don't really know how that would make any difference).

tests/TestCase/ORM/QueryRegressionTest.php
```diff
@@ -1757,4 +1758,192 @@ public function testFormatDeepDistantAssociationRecords2()
...
+                                ->select(1.23456),
+                            2
+                        ],
+                        [null, 'integer']
+                    )
+                ];
+            });
+
+        $result = $query->first()->get('value');
+        $this->assertEquals('1.23', $result);
```
garas: This should be `assertSame()`.

There is regression here on `4.x` with SQL Server on AppVeyor.

https://ci.appveyor.com/project/cakephp/cakephp/builds/26914135#L296

Decimal in `3.x` is float and string in `4.x`, and `'1.23' == 1.23000`, but `'1.23' !== '1.23000'`

ndm2: I don't think there's any type conversion on CakePHP's part involved, given that `value` isn't a mapped column, both values are strings. I think it's PHPUnit doing a conversion somewhere, probably a loose check in earlier versions, and a strict one in later versions, ie something like `'1.23' == '1.230000'` vs `'1.23' === '1.230000'`.

Assigning a float return type to the function and comparing against a float should work I think.

garas: It was just my guess of failure reason, so you are probably right.

Anyway, this should be fixed in some way to pass AppVeyor 4.x build.

PR #13538, opened 2019-08-19
Add ascii validation methods

### LINKED ISSUE #13376: Validation::alphaNumeric() allows non-alphanumeric characters
This is a (multiple allowed):

* [ ] bug
* [ ] enhancement

* CakePHP Version: 4.0
* Platform and Target: Non-related

### What you did
```php
debug( Validation::alphaNumeric('日本語') );
```

### What happened
`true`

`false`

Though I think it's a well-known feature,
https://github.com/cakephp/cakephp/blob/e39e85e3fc441e60bfe8f71d7a30794e9f053481/tests/TestCase/Validation/ValidationTest.php#L121-L125
it would be nice if it could be changed in 4.0. Since `containsNonAlphaNumeric()` assumes `[a-zA-Z0-9]` as alphanumeric, I think we can apply the same rule to `alphaNumeric()`.

markstory:
* Add new validation methods for checking ascii-only contents.
* Add `notAlphaNumeric()` to be a proper inversion of `alphaNumeric()`
  unlike `containsNonAlphaNumeric()`.

Fixes #13376

### DIFF
--- a/src/Validation/Validation.php
+++ b/src/Validation/Validation.php
@@ -120,9 +120,10 @@ class Validation
     }
 
     /**
-     * Checks that a string contains only integer or letters
+     * Checks that a string contains only integer or letters.
      *
-     * Returns true if string contains only integer or letters
+     * This method's definition of letters and integers includes unicode characters.
+     * Use `asciiAlphaNumeric()` if you want to exclude unicode.
      *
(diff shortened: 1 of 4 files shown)


dereuromark:
Looks good
One test fail though.

Issue #13132, opened 2019-04-17
Separate dirty array|scalar returns


dereuromark:
This is a (multiple allowed):

* [ ] bug
* [x] enhancement
* [x] feature-discussion (RFC)

* CakePHP Version: 4.x

### What you did
Inside controller actions patching
```php
$stateMachineProcess = $this->StateMachineProcesses
    ->patchEntity($stateMachineProcess, $this->request->getData());
```

### What happened
getData() returns too much, especially array|scalar which is somewhat a code smell here.

Phpstan with level 6+ in apps always fails for
```
Parameter #2 $data of method StateMachine\Model\Table\StateMachineProcessesTable::patchEntity() expects array, array|string|null given.
```

### What you expected to happen
To have separate methods here for all vs specific input, making the optional first argument invalid.
We could in a first step introduce the single version, and later then deprecate the god wrapper.

What about
- getDataArray() / getAllData()
- getData(string $key)

?

The current workaround is to add manual (array) casts everywhere.

And I am sure there are other methods that suffer the same issue here. This is only one of the examples I found in our API.

markstory:
We could use `getParsedBody()` which is the PSR7 method to get all the request data. That would avoid needing to add another method.

dereuromark:
I checked, that method is annotated as `null|array|object` which actually makes it worse IMO.

dereuromark:
If 
```php
    ->patchEntity($stateMachineProcess, (array)$this->request->getData());
```
is the only and recommended way to go, I am fine adding the cast into my bake templates.
It works for phpstan level 7 at least now.

dereuromark:
I am fine with casting, for now. Closing then as we need to focus on more important matters now.

dereuromark:
After https://github.com/cakephp/cakephp/pull/13539 and thinking about it:
- `getDataOrFail()` would be nice, always returning an array or failing with meaningful exception.

PR #13539, opened 2019-08-19
Empty post body for json will put null value in request->data

mvdriel:
Empty post body for json will put null value in request->data

This will cause the following error when using : request->data('test')

Error: [TypeError] Argument 1 passed to Hash::get() must be of the type array, null given, called in /lib/Cake/Network/CakeRequest.php on line 976


dereuromark:
See https://github.com/cakephp/cakephp/issues/13132 :)
I raised that concern already in the past - as static analyzers detect this usually way before even such a use case arrises.

mvdriel:
This a PR is for the 2.x branch. @dereuromark Shouldn't the milestone be a 2.x too? Could we merge this bug fix into the 2.x branch and create another PR for 3.x (if it is relevant)?

markstory:
Thank you 👍

PR #13540, opened 2019-08-19
Fix errors reported by static analyzers.

ADmad:


psalm.xml
```diff
@@ -12,7 +12,9 @@
         <directory name="src" />
         <ignoreFiles>
             <directory name="src/Filesystem" />
+            <directory name="src/Auth" />
             <directory name="vendor" />
+            <file name="src/Controller/Component/AuthComponent.php" />
```
dereuromark: Is it a good idea to ignore a whole file? Don't we might miss some other valuable ones?

src/Http/ResponseEmitter.php
```diff
@@ -42,12 +42,33 @@
...
+     * Emit a response.
+     *
+     * Emits a response, including status line, headers, and the message body,
+     * according to the environment.
+     *
+     * @param \Psr\Http\Message\ResponseInterface $response The response to emit.
+     * @return bool
      */
-    public function emit(ResponseInterface $response, $maxBufferLength = 8192): bool
+    public function emit(ResponseInterface $response): bool
```
markstory: Although I doubt anyone is using it we should document this change in the migration guide.

src/Http/Session/DatabaseSession.php
```diff
@@ -164,8 +168,10 @@ public function write($id, $data): bool
      */
     public function destroy($id): bool
     {
+        /** @var string $pkField */
+        $pkField = $this->_table->getPrimaryKey();
```
markstory: What harm would a string cast do here? Seems like we'd be able to skip the annotation and we'd get better runtime errors.

psalm.xml
ADmad: Deprecated classes aren't going to get any updates. All existing major issues are already fixed, so ignoring remaining noisy issues isn't a problem IMO.

src/Http/ResponseEmitter.php
ADmad: Will do, this was necessary as the interface changed and the method no longer has the 2nd parameter.

src/Http/Session/DatabaseSession.php
ADmad: I did try casting first but then psalm complained about potential invalid array to string cast :(

ADmad:
Merging so I can continue with next round. Happy to revisit any of these changes if needed.

PR #13542, opened 2019-08-20
Move serialization/unserialization code in helper methods.

ADmad:
Besides making code cleaner this allows overriding the serialized
format to non PHP specific formats like JSON.

### DIFF
--- a/src/Cache/Engine/RedisEngine.php
+++ b/src/Cache/Engine/RedisEngine.php
@@ -315,7 +315,7 @@ class RedisEngine extends CacheEngine
      *
      * @param mixed $value Value to serialize.
      * @return string
-     * @https://github.com/phpredis/phpredis/issues/81
+     * @link https://github.com/phpredis/phpredis/issues/81
      */
     protected function serialize($value): string
     {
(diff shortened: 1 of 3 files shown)


tests/TestCase/Cache/Engine/RedisEngineTest.php
```diff
@@ -388,6 +388,27 @@ public function testIncrement()
...
+     * @return void
+     */
+    public function testIncrementAfterWrite()
+    {
+        Cache::delete('test_increment', 'redis');
+        $result = Cache::write('test_increment', 1, 'redis');
+        $this->assertTrue($result);
+
+        $result = Cache::read('test_increment', 'redis');
+        $this->assertEquals(1, $result);
```
dereuromark: Should we use assertSame here()? As true otherwise would also falsely say true.

src/Cache/Engine/RedisEngine.php
```diff
@@ -316,6 +307,40 @@ public function clearGroup(string $group): bool
...
 
+    /**
+     * Serialize value for saving to Redis.
+     *
+     * This is needed instead of using Redis' in built serialization feature
+     * as it creates problems incrementing/decrementing intially set integer value.
+     *
+     * @param mixed $value Value to serialize.
+     * @return string
+     * @https://github.com/phpredis/phpredis/issues/81
```
garas:
```suggestion
     * @link https://github.com/phpredis/phpredis/issues/81
```

PR #13544, opened 2019-08-20
Fix errors reported by psalm and update it's baseline file.

ADmad:


markstory (commented):
Looks good to me.

src/Cache/Engine/MemcachedEngine.php
```diff
@@ -111,7 +111,7 @@ public function init(array $config = []): bool
             'json' => Memcached::SERIALIZER_JSON,
             'php' => Memcached::SERIALIZER_PHP,
         ];
-        if (defined('Memcached::HAVE_MSGPACK') && Memcached::HAVE_MSGPACK) {
+        if (defined('Memcached::HAVE_MSGPACK')) {
```
dereuromark: failed tests indicate that this should be reverted.

ADmad: This check and failing test were incorrect, I have fixed it.

Only the existence of `Memcached::HAVE_MSGPACK` needed to be checked to add `msgpack` to the serializer list. The actual value of `Memcached::HAVE_*` constants is checked separately in `_setOptions()` methods and exception thrown accordingly.

PR #13546, opened 2019-08-21
Removed deprecated tag.

ADmad:
Mention the deprecation in description instead.


ADmad:
Since the deprecations are context based using `@deprecated` confuses IDE users.

PR #13552, opened 2019-08-22
Socket connect works with unix file sockets using port null

steinkel:
Socket currently adds the `:port` to the resource, preventing `unix://` file based socket connections.

### DIFF
--- a/src/Network/Socket.php
+++ b/src/Network/Socket.php
@@ -161,8 +161,12 @@ class Socket
         }
 
         set_error_handler([$this, '_connectionErrorHandler']);
-        $this->connection = stream_socket_client(
-            $scheme . $this->_config['host'] . ':' . $this->_config['port'],
+        $remoteSocketTarget = $scheme . $this->_config['host'];
+        if ($this->_config['port'] !== null) {
+            $remoteSocketTarget .= ':' . $this->_config['port'];
+        }
+        $this->connection = $this->_getStreamSocketClient(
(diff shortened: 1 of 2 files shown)


ADmad:
If you skip the new method I think this could go to `master` itself.

steinkel:
I'll send it another to master as bugfix and keep it here for the added test in 3.next OK?

ADmad (approved):

src/Network/Socket.php
```diff
@@ -161,8 +161,13 @@ public function connect()
         }
 
         set_error_handler([$this, '_connectionErrorHandler']);
-        $this->connection = stream_socket_client(
-            $scheme . $this->_config['host'] . ':' . $this->_config['port'],
+        $remoteSocketTarget = $scheme . $this->_config['host'];
+        $port = $this->_config['port'];
+        if (is_int($port)) {
+            $remoteSocketTarget .= ':' . (int)$port;
```
ADmad: Why the cast to int when already `is_int()` check earlier?  I still preferred the `!== null` check just incase someone sets the port as string.

steinkel: ok, I'll allow string but cast it to int to ensure we're dealing with an int port

markstory:
Is it ok to close this? Similar changes have been merged into master.

steinkel:
I've extracted the socket instance creation in this branch, and added specific unit tests for the file case.

Issue #13554, opened 2019-08-22
Use `UploadedFileInterface` objects in request data.


ndm2:
This is a (multiple allowed):

* [ ] bug
* [x] enhancement
* [x] feature-discussion (RFC)

---

#9407 introduced `UploadedFileInterface` objects to the server request, but for backwards compatibility reasons, it had to keep the array style in the request data.

I think it would be nice if at some point having objects instead of arrays in the request data would be the default, IMHO it's way more comfortable to work with. However, such a hard break is hard to sell of course, so how about introducing it stepwise, making it optional and disabled by default at first?

In my personal projects I've usually used a custom middleware that merges the upload objects into the request data, it's simple and works fine, but introducing a middleware for handling file uploads in the core, probably isn't the best way to solve this.

So I'd like to propose a configuration option approach that could be used by the request factory. `ServerRequestFactory::fromGlobals()` already reads the session config, so this wouldn't introduce any new dependencies.

* [ ] The feature could be introduced immediately,
* [ ] then made the default (in the application template) at some point,

and in case the old behavior should be dropped later on,

* [ ] there could possibly a phase of deprecation next,
* [ ] until it could finally be made the default behavior of the `ServerRequest` class, dropping the configuration option.

ADmad:
We can add the feature as opt-in in 3.9 and then make it default in 4.0 with the ability to opt-out for those upgrading from 3.x.

markstory:
@ADmad Do you mean to default the configuration flag to on for new applications in 4.0? We should allow folks upgrading a smooth path where they can retain the old behavior for a while.

ADmad:
@markstory Yes I meant default to `on` for new 4.0 applications. New users / applications shouldn't have to bear the burden of legacy. How to maintain old behaviour can be documented in migration guide.

PR #13555, opened 2019-08-22
Add initial support for file upload objects in the request data.

ndm2:
Uses an opt-in configuration approach in order to keep things backwards compatible.

I couldn't really decide with what names I should go. I've opted for the property "workaround" because changing the signature of `_processFiles()` wouldn't be backwards compatible, and removing the private property later on won't have any side effects on userland code.

refs #13554

### DIFF
--- a/src/Http/ServerRequestFactory.php
+++ b/src/Http/ServerRequestFactory.php
@@ -54,7 +54,7 @@ abstract class ServerRequestFactory extends BaseFactory
             'webroot' => $uri->webroot,
             'base' => $uri->base,
             'session' => $session,
-            'mergeFilesAsObjects' => Configure::read('App.uploadedFilesAsObjects'),
+            'mergeFilesAsObjects' => Configure::read('App.uploadedFilesAsObjects', false),
         ]);
 
         return $request;
--- a/src/Http/ServerRequestFactory.php
+++ b/src/Http/ServerRequestFactory.php
@@ -54,7 +54,7 @@ abstract class ServerRequestFactory extends BaseFactory
             'webroot' => $uri->webroot,
             'base' => $uri->base,
             'session' => $session,
-            'mergeFilesAsObjects' => Configure::read('ServerRequest.mergeFilesAsObjects'),
+            'mergeFilesAsObjects' => Configure::read('App.uploadedFilesAsObjects'),
         ]);
 
         return $request;
--- a/src/Http/ServerRequestFactory.php
+++ b/src/Http/ServerRequestFactory.php
@@ -54,6 +54,7 @@ abstract class ServerRequestFactory extends BaseFactory
             'webroot' => $uri->webroot,
             'base' => $uri->base,
             'session' => $session,
+            'mergeFilesAsObjects' => Configure::read('ServerRequest.mergeFilesAsObjects'),
         ]);
 
         return $request;
--- a/src/Http/ServerRequest.php
+++ b/src/Http/ServerRequest.php
@@ -446,7 +446,9 @@ class ServerRequest implements ArrayAccess, ServerRequestInterface
      */
     protected function _processFiles($post, $files)
     {
-        if (!is_array($files)) {
+        if (empty($files) ||
+            !is_array($files)
+        ) {
             return $post;
         }
         $fileData = [];
--- a/src/Http/ServerRequest.php
+++ b/src/Http/ServerRequest.php
@@ -244,6 +244,13 @@ class ServerRequest implements ArrayAccess, ServerRequestInterface
     ];
 
     /**
+     * Whether to merge file uploads as objects (`true`) or arrays (`false`).
+     *
+     * @var bool
+     */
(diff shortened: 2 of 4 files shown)


ADmad:
Since this a new feature (even if opt-in) it needs to target `3.next`.

ndm2:
Oops, forgot about that... I'll rebase.

ndm2:
If this is going to be just a temporary thing, maybe the option should be placed under a special key, something like `_upgrade` or `_features` maybe :thinking:

ADmad:
The ability to continue having file arrays will need to remain until 5.0 at least. So it won't be short lived.

ndm2:
> The ability to continue having file arrays will need to remain until 5.0 at least. So it won't be short lived.

Sure, it just feels a little bit out of the ordinary, like a workaround. But the more I think about it, the less different it seems to be from session configuration.

src/Http/ServerRequestFactory.php
```diff
@@ -54,6 +54,7 @@ public static function fromGlobals(
             'webroot' => $uri->webroot,
             'base' => $uri->base,
             'session' => $session,
+            'mergeFilesAsObjects' => Configure::read('ServerRequest.mergeFilesAsObjects'),
```
markstory: This configuration flag could go under `App` so that we don't need to add another top level key.

ADmad: Also the flag could be named `uploadedFilesAsObjects`. The merging is an internal detail and doesn't need to reflect in flag name.

ndm2:
So I just wasted a few hours of my life trying to figure out why appveyor was failing, I first thought it was directly related to my changes, but after lots of unnecessary trial and error it looks like it isn't, the same error occours in various differrent situations, which can be observed when running tests partially, funny stuff like for example a second `DatabaseSession::write()` call in the database session test causes things to fail, but only when the form tests run too...

Thanks, I hate it :confused:

PHP is crashing with a fault in `ODBC32.dll`:

> Faulting application name: php.exe, version: 7.2.21.0, time stamp: 0x5d4041ec
> Faulting module name: ODBC32.dll, version: 6.3.9600.17415, time stamp: 0x5450504d
> Exception code: 0xc0000005
> Fault offset: 0x00000000000034e1
> Faulting process id: 0xa90
> Faulting application start time: 0x01d55a91cf511303
> Faulting application path: C:\php\php.exe
> Faulting module path: C:\windows\SYSTEM32\ODBC32.dll

Upgrading `php_pdo_sqlsrv` to at least `5.6.0` (not sure if that would be acceptable here) makes the problem go away, might be somehow related to microsoft/msphpsql#434.

src/Http/ServerRequestFactory.php
dereuromark: If the flag is renamed, shouldn't also the config key be the same renamed one for consistency?

ADmad:
@ndm2 I am sorry you had to learn the hard way that the best way to deal with appveyor failures is to ignore them :P

ndm2:
Yeah, from now on I'll probably refrain from spending too much time with stuff like that... my mistake was that I've crossed a certain effort/reward boundary, so I had no choice, I had to go on and find something in order to not feel completely defeated :)

src/Http/ServerRequestFactory.php
ndm2: Maybe... maybe not. The class kinda already has uploads as objects, as all uploads are available as objects via `getUploadedFiles()`.

src/Http/ServerRequestFactory.php
```diff
@@ -54,6 +54,7 @@ public static function fromGlobals(
             'webroot' => $uri->webroot,
             'base' => $uri->base,
             'session' => $session,
+            'mergeFilesAsObjects' => Configure::read('App.uploadedFilesAsObjects'),
```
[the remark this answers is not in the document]
ndm2: It certainly wouldn't hurt, but it shouldn't create any problems, given that `null` is `false`-ish, and there's a test for it.

ADmad (approved):
This seems good to merge.

Validation functions can already accept `UploadedFileInterface` instance and `FileWidget` unsets the value so no other change should be required.

ADmad:
@markstory Ping

ndm2:
I can add some docs later on unless someone else wants to do it :)

ADmad:
Please do :)

ADmad:
We overlooked the fact that `Validator::isEmpty()` needs to be updated to account for uploaded file object when `EMPTY_FILE` flag is used.

ndm2:
I didn't even knew this check existed... I think I've never used the built-in empty validation on file fields.

ADmad:
@ndm2 No problem, I opened a PR to fixed it. Just FYI, `Validator::allowEmpty()` is deprecated and one should now use type specific methods like `allowEmptyString()`, `allowEmptyDate()`, `allowEmptyFile()` etc. :slightly_smiling_face:.

PR #13556, opened 2019-08-24
Make Controller _viewPath() method public

Ali1:
This method can be useful when building components that use view builder.

```
        $this->getController()->disableAutoRender();
        $this->getController()->viewBuilder()->setTemplatePath($this->getController()->_viewPath());
        $this->getController()->viewBuilder()->setTemplate($this->request->getParam('action'));
        $build = $this->getController()->createView();
        $html = $build->render();
```

### DIFF
--- a/src/Controller/Controller.php
+++ b/src/Controller/Controller.php
@@ -804,7 +804,7 @@ class Controller implements EventListenerInterface, EventDispatcherInterface
      *
      * @return string
      */
-    protected function _viewPath()
+    public function _viewPath()
     {
         $viewPath = $this->name;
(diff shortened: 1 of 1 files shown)


ADmad:
Why do you need to directly use controller's view builder for rendering instead of `Controller::render()` ?

Ali1:
I need to do things with the generated html within the component, not output it to the browser.

ADmad:
Component isn't the right place to manipulate html. Nor is adding public methods to controller related to view rendering is something we want to do.

Please use one of our support forums to get ideas for a better way to achieve what you need.

PR #13557, opened 2019-08-24
Use template / generic type for annotating registries object types.

ADmad:
https://psalm.dev/docs/annotating_code/templated_annotations/
https://arnaud.le-blanc.net/post/phpstan-generics.html

Support for generic types in phpstan is still WIP and partially available using it's `master` branch currently.

Edit: phan supports it too https://github.com/phan/phan/wiki/Generic-Types

Issue #13559, opened 2019-08-26
3.next/4.0: Split getter/setter TODO


dereuromark:
We have a few more that we didnt find and fix yet in terms of clean API:

ConsoleOutput:

- [x] public function styles(?string $style = null, $definition = null) {}

The API here is especially wacky in terms of true/false input/output mixed in to allow removing :)

RequestHandlerComponent

- [ ] accepts()
- [ ] requestedWith()

etc

ADmad:
`RequestHandlerComponent` as a whole is a bit wacky and it's code is quite difficult to walk through.

dereuromark:
Closing then as good enough for now.

PR #13560, opened 2019-08-26
Fix up doc blocks and issues around mixed return type.

dereuromark:
Follow https://github.com/cakephp/cakephp/pull/13557

Found a few more issues around too broad "mixed". mixed here is evil, as it basically kills all possible static analysis.
We could check if even more concrete (ideally interface) types could be set as objects in some of those cases.

Fixed also a few other issues found with missing typehints along the way.

### DIFF
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -78,9 +78,6 @@
     </PossiblyNullReference>
   </file>
   <file src="src/Console/ConsoleOutput.php">
-    <PossibleRawObjectIteration occurrences="1">
-      <code>$style</code>
-    </PossibleRawObjectIteration>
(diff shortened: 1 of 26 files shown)


dereuromark:
Issues found in static anal. are side effect issues of bad API as referenced in https://github.com/cakephp/cakephp/issues/13559
We should silence those until API is fixed.

ADmad (commented):
Psalm is unhappy with some of the changes.

dereuromark:
See my comment "We should silence those until API is fixed."

ADmad:
I pushed a commit. Once docblock type was incorrect and adding inline annotations took care of the remaining errors.

PR #13562, opened 2019-08-26
4.x psalm

ADmad:
Use psalm specific annotations until IDEs and other tools catch up with support for generics and object-like array types.

### DIFF
--- a/src/View/HelperRegistry.php
+++ b/src/View/HelperRegistry.php
@@ -88,8 +88,7 @@ class HelperRegistry extends ObjectRegistry implements EventDispatcherInterface
      * Provide public read access to the loaded objects
      *
      * @param string $name Name of property to read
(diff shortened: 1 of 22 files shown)


src/View/HelperRegistry.php
```diff
@@ -88,7 +88,8 @@ public function __isset(string $helper): bool
      * Provide public read access to the loaded objects
      *
      * @param string $name Name of property to read
-     * @return \Cake\View\Helper|null
+     * @return object|null
+     * @psalm-return \Cake\View\Helper|null
```
dereuromark: We might be able to keep the more specific one here?

ADmad: phpstan throws error since the parent class has `object|null` but I guess we could add an ignore for it.

PR #13563, opened 2019-08-26
More concrete array doc for string value arrays.

dereuromark:
Follow https://github.com/cakephp/cakephp/pull/13549

for phpstan etc


dereuromark:
Thanks for the help :)

dereuromark:
I wonder if we can/should backport some of those changes into 3.x

PR #13565, opened 2019-08-27
Clear uncleared property in TableLocator::clear method

bancer:
As `$this->_config` is reset then it makes sense to reset `$this->_options` too.


ADmad:
A test would be nice.

bancer:
I'm not sure it can be unit tested. There is no public getter for `$_options` property.

ADmad (approved):

saeideng:
you can do it  in `remove()` method too
```php
    public function remove($alias)
    {
        unset(
            $this->_instances[$alias],
            $this->_config[$alias],
            $this->_fallbacked[$alias]
        );
    }
```

PR #13566, opened 2019-08-28
Fix memory leak in unit tests - improve clean up in tearDown method

bancer:
There is a similar clean up in 2.x - https://github.com/cakephp/cakephp/blob/2.x/lib/Cake/TestSuite/CakeTestCase.php#L168

### DIFF
--- a/src/ORM/Locator/LocatorAwareTrait.php
+++ b/src/ORM/Locator/LocatorAwareTrait.php
@@ -70,7 +70,7 @@ trait LocatorAwareTrait
      */
     public function getTableLocator()
     {
-        if (!$this->_tableLocator) {
+        if (!isset($this->_tableLocator)) {
             $this->_tableLocator = TableRegistry::getTableLocator();
         }
 
--- a/src/ORM/Locator/LocatorAwareTrait.php
+++ b/src/ORM/Locator/LocatorAwareTrait.php
@@ -70,7 +70,7 @@ trait LocatorAwareTrait
      */
     public function getTableLocator()
     {
-        if (!isset($this->_tableLocator)) {
+        if ($this->_tableLocator === null) {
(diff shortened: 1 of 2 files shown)


src/TestSuite/TestCase.php
```diff
@@ -161,6 +161,7 @@ public function tearDown()
             Configure::write($this->_configure);
         }
         $this->getTableLocator()->clear();
+        unset($this->_configure, $this->_pathRestore, $this->_tableLocator);
```
ADmad: Unsetting declared properties will make static analyzers unhappy. Set them to null / initial declared value instead.

bancer: Setting to `null` or an empty array does not free all the memory though.

src/ORM/Locator/LocatorAwareTrait.php
```diff
@@ -70,7 +70,7 @@ public function setTableLocator(LocatorInterface $tableLocator)
      */
     public function getTableLocator()
     {
-        if (!$this->_tableLocator) {
+        if (!isset($this->_tableLocator)) {
```
ADmad: This should be reverted or changed to `$this->_tableLocator === null`

dereuromark:
Nice work :+1:

src/ORM/Locator/LocatorAwareTrait.php
bancer: @ADmad, @dereuromark The problem with `$this->_tableLocator === null` and `!$this->_tableLocator` is that it results in: 
```
Undefined property: MyTest::$_tableLocator

.../Vendor/cakephp/orm/Locator/LocatorAwareTrait.php:73
.../src/TestSuite/MyTestCase.php:59
```

PR #13567, opened 2019-08-28
WIP - PHP 7.4

ADmad:
We need to figure out how to deal with the `Function ReflectionType::__toString() is deprecated` error generated in phpunit's mock generator class. I tried using `syfony/phpunit-bridge` but [didn't work](https://travis-ci.org/ADmad/cakephp/jobs/577837769).

### DIFF
--- a/src/Console/ConsoleOutput.php
+++ b/src/Console/ConsoleOutput.php
@@ -239,7 +239,7 @@ class ConsoleOutput
             }
         }
 
-        return "\033[" . implode($styleInfo, ';') . 'm' . $matches['text'] . "\033[0m";
+        return "\033[" . implode(';', $styleInfo) . 'm' . $matches['text'] . "\033[0m";
     }
 
     /**
--- a/src/Shell/CompletionShell.php
+++ b/src/Shell/CompletionShell.php
@@ -168,7 +168,7 @@ class CompletionShell extends Shell
     protected function _output($options = [])
     {
         if ($options) {
-            return $this->out(implode($options, ' '));
+            return $this->out(implode(' ', $options));
         }
     }
 }
--- a/src/View/Helper/SecureFieldTokenTrait.php
+++ b/src/View/Helper/SecureFieldTokenTrait.php
@@ -50,8 +50,8 @@ trait SecureFieldTokenTrait
         ksort($locked, SORT_STRING);
         $fields += $locked;
 
-        $locked = implode(array_keys($locked), '|');
-        $unlocked = implode($unlockedFields, '|');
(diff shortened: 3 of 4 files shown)


dereuromark:
Tests reveal a warning

> PHP Warning:  Fixture creation for "products" failed "SQLSTATE[HY000]: General error: 1 table "products" already exists" in /home/travis/build/cakephp/cakephp/src/TestSuite/Fixture/TestFixture.php on line 313
Warning: Fixture creation for "products" failed "SQLSTATE[HY000]: General error: 1 table "products" already exists" in /home/travis/build/cakephp/cakephp/src/TestSuite/Fixture/TestFixture.php on line 313

Otherwise looks good :)

dereuromark (approved):

garas:
Except `ReflectionType::__toString()`, there is only `Database\ConnectionTest::testBufferedStatementCollectionWrappingStatement` [failure](https://travis-ci.org/cakephp/cakephp/jobs/577876963#L10559) on Sqlite.

Maybe new version of Sqlite bundled with PHP 7.4 has something new with buffered results.

ADmad:
@garas Probably, we could just skip the test on PHP 7.4 for now.

ndm2:
Since there won't be any new `sebastianbergmann/phpunit-mock-objects` releases anymore, maybe just modify it on the fly:

```bash
if [[ ${TRAVIS_PHP_VERSION} == "7.4snapshot" ]]; then
  sed -i 's/(string) $parameter->getType()/$parameter->getType()->getName()/g' vendor/phpunit/phpunit-mock-objects/src/Generator.php;
fi
```

dereuromark:
Good plan B. Before that, maybe try https://github.com/sminnee/phpunit-mock-objects first
That via composer replace would probably be also working locally a bit easier to get right :)

ADmad:
@dereuromark I tried using `sminnee/phpunit-mock-objects`, it doesn't work with phpunit 6. Throws errors like
```
Argument 1 passed to PHPUnit_Framework_MockObject_MockBuilder::__construct() must be an instance of PHPUnit_Framework_TestCase, instance of Cake\Test\TestCase\ORM\TableTest given
```

markstory:
Should we start thinking about moving off of phpunit mocks and using mockery instead?

ADmad:
@markstory That's worth considering depending on the effort involved.

I wish we could move off phpunit itself. It drops support for older versions (PHP and it's own releases) too fast to be practical for real-world usage. Also it's supposedly backwards compatible releases often require us having to tweak our testsuite.

dereuromark:
Merging, as the rest can be a follow up.

PR #13568, opened 2019-08-28
Throw exception from Event::getSubject() when subject is null.

ADmad:
This avoids static analyzers from complaining about possibly null value when using return value of the method.


dereuromark:
Looks like there is need a try/catch required even in core.
But I think this is an acceptable way to save a nullable getter method to be added for such rare cases

:+1:

PR #13571, opened 2019-08-29
Backport docblock fixes from 4.0

dereuromark:
Makes 3.x closer to improved 4.x on this semantic level already.

### DIFF
--- a/src/Datasource/EntityTrait.php
+++ b/src/Datasource/EntityTrait.php
@@ -1207,7 +1207,7 @@ trait EntityTrait
      * Get a single value of an invalid field. Returns null if not set.
      *
      * @param string $field The name of the field.
-     * @return mixed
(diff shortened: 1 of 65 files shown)


src/Datasource/EntityTrait.php
```diff
@@ -1207,7 +1207,7 @@ public function getInvalid()
      * Get a single value of an invalid field. Returns null if not set.
      *
      * @param string $field The name of the field.
-     * @return mixed
+     * @return mixed|null
```
ADmad: woops

dereuromark: having null as verbose type here actually makes nullable part more clear, dont u think?

ADmad: Meh, if the static analyzers stay happy I don't care.

Issue #13418, opened 2019-07-15
Paginator count query is so dumb with SELECT count(*) FROM (....)


NightCat419:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.7.8
* Platform and Target: Apache, MySQL
### What you did

        $fields=[
            'id'=>[ ... ],
            'ship_date'=>[ ... ],
                ... ... ... ...
            'total_value'=>[ ... ],
        ];
        $defaults=[];
        $default_display= 'id'=>true,'ship_date'=>true,'warehouse'=>true,'status'=>true,'age'=>true];

        // apply default sort
        if(!isset($this->request->query['sort']))
        {
           $this->request->query['sort']='id';
           $this->request->query['direction']='desc';
        }
        // apply default limit
        if(!isset($this->request->query['limit']))
           $this->request->query['limit']=20;

        $finder=$this->Filter
            ->filter_to_conditions(
                $fields,
                $filter,
                [
                    'a',
                    'b',
                    'c',
                    'd',
                    ....
                ],
                ['Shipments.id']
            );
        $fields=$this->Filter->assign_defaults($fields,$filter,$defaults,$default_display);
        $this->set('_serialize', ['shipments']);

        $this->set('fields',$fields);
        $this->set('shipments', $this->paginate($finder));

### What happened
This is executing so long query to get count.

SELECT 
  (
    COUNT(*)
  ) AS `count` 
FROM 
  (
    SELECT 
      Shipments.id AS `Shipments__id`, 
      Shipments.ship_date AS `Shipments__ship_date`, 
      ....... ......... .........
      Shipments.status AS `Shipments__status` 
    FROM 
      shipments Shipments 
      LEFT JOIN ... 
      LEFT JOIN ...
      LEFT JOIN ...
      LEFT JOIN ...
      LEFT JOIN 1
      LEFT JOIN 2
      LEFT JOIN 3 
      LEFT JOIN 4
      LEFT JOIN 5
      LEFT JOIN 6
      LEFT JOIN 7
    GROUP BY 
      Shipments.id
  ) count_source

This query takes even about 20000 milliseconds.
It's so terrible.

lorenzo:
Ok... do you have any suggestions to make it better?

NightCat419:
Why cannot use more simple count query excluded other fields and left join?
[...]

PR #13572, opened 2019-08-29
Simple paginator

ADmad:
Added `SimplePaginator` which skips total count query. This is useful for paginating repositories with very large number of records where count query is very expensive or in case total count is unknown. 

Usage through `PaginatorComponent` would be like this:
```php
// in initialize()
$this->loadComponent('Paginator', [
    'paginator' => new \Cake\Datasource\SimplePaginator(),
]);
```

Refs https://github.com/cakephp/cakephp/issues/13418#issuecomment-526176622


src/Datasource/Paginator.php
```diff
@@ -265,26 +263,31 @@ protected function buildParams(array $data)
...
             'page' => $page,
-            'current' => $data['numResults'],
+            'current' => $numResults,
             'count' => $count,
             'perPage' => $limit,
             'start' => $start,
             'end' => $end,
             'prevPage' => $page > 1,
-            'nextPage' => $count > ($page * $limit),
+            'nextPage' => $count === null ? true : ($count > ($page * $limit)),
```
markstory: Should determining the next/prev pages be moved into methods that implementations can override? Same for start/end?

src/Datasource/SimplePaginator.php
```diff
@@ -0,0 +1,35 @@
...
+ * @link          http://cakephp.org CakePHP(tm) Project
+ * @since         3.9.0
+ * @license       http://www.opensource.org/licenses/mit-license.php MIT License
+ */
+namespace Cake\Datasource;
+
+use Cake\Datasource\Exception\PageOutOfBoundsException;
+
+/**
+ * Simplified paginator which avoids query to get total count of records.
```
markstory:
```suggestion
 * Simplified paginator which avoids potentially expensives queries 
 * to get the total count of records.
 * 
 * When using a simple paginator you will not be able to generate page numbers.
 * Instead use only the prev/next pagination controls, and handle 404 errors
 * when pagination goes past the available result set.
```

tests/TestCase/Datasource/SimplePaginatorTest.php
```diff
@@ -0,0 +1,160 @@
...
+ * @since         3.5.0
+ * @license       http://www.opensource.org/licenses/mit-license.php MIT License
+ */
+namespace Cake\Test\TestCase\Datasource;
+
+use Cake\Core\Configure;
+use Cake\Datasource\SimplePaginator;
+use Cake\ORM\Entity;
+
+class SimplePaginatorTest extends PaginatorTest
```
markstory: This seems efficient now, but will likely bite us as new tests are added and cause failures in a seemingly un-related test class. Another approach could be to move the shared test code into a trait that both tests include?

src/Datasource/Paginator.php
ADmad: Okay, with do.

tests/TestCase/Datasource/SimplePaginatorTest.php
ADmad: Sounds good, I make a trait.

src/Datasource/SimplePaginator.php
```diff
@@ -0,0 +1,40 @@
...
+ * Redistributions of files must retain the above copyright notice.
+ *
+ * @copyright     Copyright (c) Cake Software Foundation, Inc. (http://cakefoundation.org)
+ * @link          http://cakephp.org CakePHP(tm) Project
+ * @since         3.9.0
+ * @license       http://www.opensource.org/licenses/mit-license.php MIT License
+ */
+namespace Cake\Datasource;
+
+use Cake\Datasource\Exception\PageOutOfBoundsException;
```
josegonzalez: This isn't actually used?

ADmad: No, my phpcs precommit hook failed me.

ADmad:
Updates done per per feedback.

ADmad:
Anything else to be done here?

ADmad:
The phpcs failures are unrelated.

garas:
`$this->buildParams()` is called twice with same data

https://github.com/cakephp/cakephp/blob/c02272f84027b33784971d90cb41ac1d0abec506/src/Datasource/Paginator.php#L177-L179

ADmad:
@garas  Fixed :slightly_smiling_face:

Issue #13577, opened 2019-08-30
3.x: command abort() vs shell abort()


dereuromark:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

3.x latest master

### What you did

migrate shell to command

CLI coding on 3.x Shell class:

    $this->abort('Invalid project path: ' . $path);

CLI coding on 3.x Command class would create error:

    $this->abort('Invalid project path: ' . $path);
    // needs to be now
    $io->error('Invalid project path: ' . $path);
    throw new StopException(self::CODE_ERROR)

Note also the requirement for the code as verbose input.
otherwise code is 244 on CLI.

The problem here is also 
- a use statement needed
- throws docblock needed/added via IDE/tooling

Quite some overhead compared to before.

### Suggestions

Re-add a convenience wrapper similar to what is in shells again.

We probably also want to default the code here for the exception:
```php
class StopException extends Exception
{
    /**
     * Default exception code
     *
     * @var int
     */
    protected $_defaultCode = 1;
}
```

markstory:
Converting a Shell early termination to Command can look like:

```php
$this->abort('Invalid project path: ' . $path);

// needs to be now
$io->error('Invalid project path: ' . $path);
$this->abort();
```

Part of the reason this change was made was so that folks could have more control over how the error messages are rendered. We could expand the types on `Command::abort()` to allow strings to set the message to make simple scenarios more ergonomic.

dereuromark:
I would like to see
```php
    /**
     * Halt the the current process with a StopException.
     *
     * @param int $code The exit code to use.
     * @throws \Cake\Console\Exception\StopException
     * @return void
     */
    public function error(string $message, $code = self::CODE_ERROR)
    {
        $this->io->error($message);
        
        throw new StopException($code);
    }
```
but there is no io object an the class anymore to make this happen.

As a general convenience method.
If people want more control they are free to use the standalone code lines - but 99% of my shell and command errors are perfectly fine to use such a convenience wrapper :) And so is probably the rest of the community.

markstory:
What if we had that kind of method on `ConsoleIo`?

```php
$io->abort('Something terrible', 99)
```

dereuromark:
PRs for both fix (master = 3.8) and enhancement (3.next = 3.9) are open.

PR #13578, opened 2019-08-30
Fix exception to be valid error default code 1 instead of 244.

dereuromark:
First part of https://github.com/cakephp/cakephp/issues/13577
```php
$io->error('Project path not found: ' . $args->getArgumentAt(0));
throw new StopException();
```
`echo $?` shows 244 as code instead of default 1 as per CLI standards.


dereuromark:
I prefer this fix however: https://github.com/cakephp/cakephp/pull/13579

PR #13579, opened 2019-08-30
Fix exception to be valid error default code 1 instead of 244 - Alternative

dereuromark:
Alternative to https://github.com/cakephp/cakephp/pull/13578

This fixes all error codes for CLI by extending the base Console exception.


garas:
In addition to this, `CommandRunner ` should check not to return anything outside 0-255, and instead return 1 or something else.

PR #13580, opened 2019-08-30
Add abort() wrapper similar to Shell::abort().

dereuromark:
Resolves the 2nd (enhancement) part of https://github.com/cakephp/cakephp/issues/13577
Convenience wrapper to avoid having to use 2+ lines for simple cases.

### DIFF
--- a/src/Console/ConsoleIo.php
+++ b/src/Console/ConsoleIo.php
@@ -251,6 +251,19 @@ class ConsoleIo
     }
 
     /**
+     * @param string $message
+     * @param int $code
+     * @return void
+     * @throws \Cake\Console\Exception\StopException
+     */
+    public function abort($message, $code = Command::CODE_ERROR)
+    {
+        $this->error($message);
+
+        throw new StopException($code);
+    }
+
+    /**
      * Wraps a message with a given message type, e.g. <warning>
      *
      * @param string $messageType The message type, e.g. "warning".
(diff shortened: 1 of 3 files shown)


dereuromark:
The failing CS is sth new coming from code sniffer directly. I fixed those in master already.

src/Console/ConsoleIo.php
```diff
@@ -250,6 +250,21 @@ public function success($message = null, $newlines = 1, $level = self::NORMAL)
...
+     * @param string $message Error message.
+     * @param int $code Error code.
+     * @return void
+     * @throws \Cake\Console\Exception\StopException
+     */
+    public function abort($message, $code = Command::CODE_ERROR)
+    {
+        $this->error($message);
+
+        throw new StopException($code);
```
ADmad: The 1st argument of `StopException`'s constructor is string `$message` not int `$code`. This creates type error doing 3.x => 4.x merge.

ADmad: This is why I prefer all future enhancements to be done on 4.x and backported to 3.next. It avoids such errors and problems during 3.x => 4.x merges.

ADmad: I changed it to `throw new StopException($message, $code);` in 4.x.

dereuromark: We have a much deeper underlying issue here IMO we should fix instead.
The first argument, message is not used at all right now which is really bad. It forces one to always add a ->error() before, which makes no sense since the message argument is there.
=> https://github.com/cakephp/cakephp/pull/13588

PR #13585, opened 2019-09-01
Fix errors reported by psalm.

ADmad:


src/View/Helper.php
```diff
@@ -95,7 +95,7 @@ public function __construct(View $view, array $config = [])
      *
      * @param string $method Method to invoke
      * @param array $params Array of params for the method.
-     * @return void
+     * @return mixed|void
```
dereuromark: Should then rather be mixed|null right?

dereuromark:
Looks good.

src/View/Helper.php
ADmad: Nah, the method doesn't return anything so psalm expects `void`.

src/View/Helper.php
```diff
@@ -144,11 +144,11 @@ protected function _confirm(string $okCode, string $cancelCode): string
...
      *
      * @param array $options Array options/attributes to add a class to
-     * @param string|null $class The class name being added.
-     * @param string $key the key to use for class.
+     * @param string $class The class name being added.
+     * @param string $key the key to use for class. Defaults to `'class'`.
      * @return array Array of options with $key set.
      */
-    public function addClass(array $options = [], ?string $class = null, string $key = 'class'): array
+    public function addClass(array $options, string $class, string $key = 'class'): array
```
ADmad: I removed the default `null` for class. Doesn't make sense for a method meant to add class accepting `null` value for class.

src/View/Helper.php
dereuromark: Why is mixed then necessary? It shouldnt. Having it both not return and return is contradicting.

src/Http/ResponseEmitter.php
```diff
@@ -267,9 +267,10 @@ protected function emitCookies(array $cookies): void
                 $key = strtolower($key);
                 $data[$key] = $value;
             }
-            if (!empty($data['expires'])) {
+            if (is_string($data['expires'])) {
```
dereuromark: This is wrong and a hard BC break!
It can be an integer, as also seen in line 252 default value - and often people use that for the argument to strtotime() as well.

src/View/Helper.php
ADmad: Because the method is overridden in `NumberHelper` and `TextHelper` which do have return values, so keeping only `void` here makes pslam complain again :)

src/Http/ResponseEmitter.php
ADmad: If the values is already integer then it doesn't need `strtotime()` conversion does it?  It's argument is supposed to be a string.

dereuromark: So it seems if we fix the default value here in line 252 to null it could be also fine.
The cases for integer here do not seem valid.

Issue #13583, opened 2019-08-30
Memory leak in unit tests - $fixtures property in TestCase class


bancer:
This is a (multiple allowed):

* [ ] bug
* [x] enhancement
* [x] feature-discussion (RFC)

* CakePHP Version: all
* Platform and Target: all

`$fixtures` property in unit tests leaks the memory and it is not possible to unset it in `tearDown` method.
I would suggest to refactor that to a getter and to remove `$fixtures` property:
```
public function getFixtures() : array {
    return ['app.Posts', 'app.Articles'];
}
```
And in the parent TestCase it could be like this for backwards compatibility:
```
public function getFixtures() : array {
    if (isset($this->fixtures) {
        return $this->fixtures;
    }
    return [];
}
```

Other properties in TestCase class also leak memory. I think it would be useful to refactor them to getters too.

markstory:
Adopting a method for getting the fixture list seems reasonable to me.

dereuromark:
Are there any numbers on how much memory we are talking about here?
If you run the current (core) suite with memory profiling output and your suggested changes, how much of a difference is there?

bancer:
In order to measure that all `public $fixture = ...;` declarations in unit tests must be replaced by `public function getFixtures() {...};` method as if `$this->fixtures` is not empty in `tearDown` method then the memory leak occurs. 10 elements array may use 300 bytes. A project with 10000 tests that use fixtures may get 30000000 bytes (28 MB) leaked. That should be multiplied by 5 for code coverage runs.

dereuromark:
This seems to be of very little win at this point, especially during 3.x.
Would you be able to make a PR for 4.x that also makes the property protected?
That would be a double-win then for this major, as we try to make more properties non-public and at the same time the getter then allows some better memory allocation, right?

bancer:
I can provide more specific numbers later after we apply this improvement in our 2.x code.

I can try to make 4.x PR. The memory usage of public and protected properties is the same though per my knowledge. 

The memory leaks in unit tests is a well known problem by the way. Most often cited [handling](https://kriswallsmith.net/post/18029585104/faster-phpunit) suggests to unset all properties that do not belong to 'PHPUnit_' classes. Any non-empty property in `tearDown` leaks memory whether it is public, protected or private.

dereuromark:
The protected is more for clean code and less side effects (clean getter access).
But we can see about the rest and possible ways to improve here also in terms of memory.

PR #13586, opened 2019-09-01
Refactor $fixtures property to a getter method

bancer:
Fixes https://github.com/cakephp/cakephp/issues/13583.


src/TestSuite/TestCase.php
```diff
@@ -804,4 +804,21 @@ public static function setAppNamespace($appNamespace = 'TestApp')
...
+     * Gets fixtures.
+     *
+     * @return array
+     */
+    public function getFixtures()
+    {
+        if (!isset($this->fixtures)) {
+            return [];
+        }
+        if (is_string($this->fixtures)) {
```
ADmad: Why does this have to be a string?  In 4.x `$fixtures` property is already declared / initialized as array.

ADmad (commented):
It would be better to make in enhancement to 4.x first and then backport to 3.x. That way the typing issue I pointed out can be avoided and also adding return type hint to new method won't be missed.

src/TestSuite/Fixture/FixtureManager.php
```diff
@@ -94,12 +94,9 @@ public function setDebug($debug)
     public function fixturize($test)
     {
         $this->_initDb();
-        if (empty($test->fixtures) || !empty($this->_processed[get_class($test)])) {
+        if (empty($test->getFixtures()) || !empty($this->_processed[get_class($test)])) {
```
dereuromark: This method call should be wrapped by empty().

src/TestSuite/TestCase.php
bancer: It is moved from FixtureManager::fixturize() - https://github.com/cakephp/cakephp/pull/13586/files#diff-00c99fa549dd6a8e04203fff9bcc684aL100.

src/TestSuite/Fixture/FixtureManager.php
bancer: What do you mean? It is already wrapped.

dereuromark: should **not**
sry
In this case just use if () or if (!)

dereuromark:
Something seems wrong with your GitHub credentials. The username is not registered on GitHub here.

src/TestSuite/Fixture/FixtureManager.php
bancer: What's the point of that? `$test->getFixtures()` does not return a boolean. It may return an empty array.

bancer: http://sandbox.onlinephpfunctions.com/code/5106681ba751da5d3e3edf2429cd754f1c913078

dereuromark: The point is that we don't do that, and we shouldn't start now.

bancer: I don't understand you. `empty($test->fixtures)` is replaced with `empty($test->getFixtures())`. So `empty()` has already been used here. Besides that `!$test->getFixtures()` does not work here as it evaluates to `false` when the fixtures are an empty array. Please take a look at the example in the sandbox I posted above.

dereuromark: empty() on a method name is the issue. Before it was not a method. And `!` usage is the non cloakling equal to that. 100% same result. Please do not argue here on this. This has always been this way.

ADmad:
Please re-open the PR against `4.x` so that we don't waste time dealing with potential issues related to typing.

bancer:
@ADmad I do not understand what you mean. I see no problems related to typing and why this could not be merged to 3.x. We do not use 4.x and the memory leak needs to be fixed in 3.x first.

ADmad:
`3.x` is regularly merged into `4.x`. That's when the issues will occur. For e.g. static analyzers will complain about the `is_string($this->fixtures)` check since `TestCase::$fixtures` is declared as array in `4.x`. Also in `4.x` we use strict typing and return types to methods, which can't be done for `3.x`. So whenever new methods get added to `3.x` we need to remember adding return types when merging to `4.x` and sometimes that's missed as I has to be manually.

In general the strict typing in 4.x ensure we write better and less error prone code and as maintainers it's easier to then backport the enhancements to 3.x instead of dealing with type issue when updates done to 3.x are merged into 4.x.

bancer:
But what about backwards compatibility? 3.x expects `$this->fixtures` to be a string in some cases.

ADmad:
The code can be updated to maintain BC when backporting for 3.x.

bancer:
You guys make it quite difficult to contribute. Now I think that I would rather had spent that time on implementing the fix in our project on in a plugin.

ADmad:
If this frustrates you just consider what we deal with on an almost daily basis, contribution our time for free to develop and maintain the project.

src/TestSuite/TestCase.php
dereuromark: That part will be taken care of follow up https://github.com/cakephp/cakephp/pull/13593

src/TestSuite/Fixture/FixtureManager.php
```diff
@@ -450,7 +444,8 @@ protected function _fixtureConnections($fixtures)
      */
     public function unload($test)
     {
-        if (empty($test->fixtures)) {
+        $fixtures = $test->getFixtures();
+        if (empty($fixtures)) {
```
dereuromark: `if (!$fixtures)` - same non-cloaking check can be used.

src/TestSuite/Fixture/FixtureManager.php
```diff
@@ -168,10 +165,11 @@ protected function _initDb()
      */
     protected function _loadFixtures($test)
     {
-        if (empty($test->fixtures)) {
+        $fixtures = $test->getFixtures();
+        if (empty($fixtures)) {
```
dereuromark: Same

src/TestSuite/TestCase.php
```diff
@@ -804,4 +804,25 @@ public static function setAppNamespace($appNamespace = 'TestApp')
...
+            deprecationWarning(
+                'Setting fixtures as string is deprecated and will be removed in 4.0.' .
+                ' Set TestCase::$fixtures as array instead.'
```
bancer: Using `TestCase::$fixtures` would still lead to the memory leak. I would recommend to set fixtures in `getFixtures()` method like in `TableTest::getFixtures()` below without initiating `$fixtures` property.

ADmad: The memory leak figures you mentioned [here](https://github.com/cakephp/cakephp/issues/13583#issuecomment-526929833 are not really noteworthy TBH. Most users who actually have such large number of tests would have GBs worthy of memory at their disposable and wouldn't care about such a small memory bump. So smaller projects the memory bump would be insignificant.

So for majority of the users the convenience of setting the property is good enough.

tests/TestCase/ORM/TableTest.php
```diff
@@ -64,20 +64,27 @@ class ProtectedEntity extends Entity
 class TableTest extends TestCase
 {
 
-    public $fixtures = [
```
ADmad: This change in unnecessary. The tests updates you have done in `FixtureManagerTest` suffice.

src/TestSuite/TestCase.php
bancer: I may agree that the memory usage improvement will not be noticeable in smaller projects. But in larger projects where many tests have 20-30 fixtures it may easily reach 10% of memory improvement. It is also worthy to mention that memory leaks have greater effect on inspections when the code coverage report is generated. 

I do not insist here but in general the usage of public properties is discouraged.

ADmad: You can a note to the 3.next docs suggesting devs with large projects to avoid using the property. There's no point directing all users to do extra work by using the method, when the issue affects only a small minority.

With the addition of the getter method `TestCase::$fixtures` will be made protected in 4.x. We have already done that for most public properties.

src/TestSuite/TestCase.php
```diff
@@ -804,4 +804,26 @@ public static function setAppNamespace($appNamespace = 'TestApp')
...
+            );
+
+            return array_map('trim', explode(',', $this->fixtures));
+        }
+
+        return $this->fixtures;
```
markstory: Did the `fixtures` property get removed? That seems like it will make linters sad. Could we initialize it to null?

src/TestSuite/TestCase.php
```diff
@@ -44,6 +44,13 @@ abstract class TestCase extends BaseTestCase
      */
     public $fixtureManager;
 
+    /**
+     * Fixtures used by this test case.
+     *
+     * @var array|string
+     */
+    public $fixtures = null;
```
dereuromark: The = null is not necessary.
Also if null is still there at the end of constructor, |null should be added to docblock.

bancer:
> PHP Fatal error:  Cake\TestSuite\TestCase and Cake\Test\TestCase\Datasource\PaginatorTestTrait define the same property ($fixtures) in the composition of Cake\Test\TestCase\Datasource\PaginatorTest. However, the definition differs and is considered incompatible. Class was composed in /home/travis/build/cakephp/cakephp/tests/TestCase/Datasource/PaginatorTest.php on line 25

tests/TestCase/Datasource/PaginatorTest.php
```diff
@@ -26,6 +26,16 @@ class PaginatorTest extends TestCase
 {
     use PaginatorTestTrait;
 
+    /**
+     * fixtures property
+     *
+     * @var array
+     */
+    public $fixtures = [
```
ADmad: Hmm.. There fixtures weren't added to `SimplePaginatorTest` after removal from the trait.  Yet there are no test failures, which makes me think setting these fixtures is probably not required.

bancer: Multiple tests fail if `$fixtures` is not added to PaginatorTest.

bancer:
Is there anything that stops this pull request from being merged?

PR #13587, opened 2019-09-01
Fix abort() method.

dereuromark:
Follow https://github.com/cakephp/cakephp/pull/13580/files#r319734608

This resolves the difference between 3.x and 4.x.
There is also a follow up PR around the topic of proper message delegation coming up.

### DIFF
--- a/tests/TestCase/Console/ConsoleIoTest.php
+++ b/tests/TestCase/Console/ConsoleIoTest.php
@@ -15,6 +15,7 @@
 namespace Cake\Test\TestCase\Console;
 
 use Cake\Console\ConsoleIo;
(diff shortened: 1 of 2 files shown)


tests/TestCase/Console/ConsoleIoTest.php
```diff
@@ -280,7 +295,7 @@ public function testErr()
      * Tests abort() wrapper.
      *
      * @expectedException \Cake\Console\Exception\StopException
-     * @expectedExceptionMessage 1
+     * @expectedExceptionMessage Some error
```
ADmad: Can you move these annotations to `expectedException*()` calls please as I have done in 4.x to avoid merge conflicts.

PR #13588, opened 2019-09-01
Fix message to be properly used

dereuromark:
Follow https://github.com/cakephp/cakephp/pull/13587 

Currently in your code you have to do:
```php
$message = 'Project path not found: ' . $args->getArgumentAt(0);
$io->error($message);
throw new StopException($message);
```

As the message of StopException is not used at all we have to un-dry duplicate it across the error() call as well as the StopException (as documented and already used). But the 2nd intuitively more important message is not used.

That is bad - we should use that in a proper way where we catch the exception and then issue the return with the error code.

Now we would only have to do:
```php
$message = 'Project path not found: ' . $args->getArgumentAt(0);
throw new StopException($message);
```

I wonder if we can do that already in 3.next or if we can only fix this in 4.x.

### DIFF
--- a/src/Console/Command.php
+++ b/src/Console/Command.php
@@ -277,6 +277,8 @@ class Command
         try {
             return $command->run($args, $io);
         } catch (StopException $e) {
+            $io->error($e->getMessage());
+            
             return $e->getCode();
         }
     }
--- a/src/Console/Command.php
+++ b/src/Console/Command.php
@@ -278,7 +278,7 @@ class Command
             return $command->run($args, $io);
         } catch (StopException $e) {
             $io->error($e->getMessage());
-            
+
             return $e->getCode();
         }
     }
--- a/src/Console/CommandRunner.php
+++ b/src/Console/CommandRunner.php
@@ -366,6 +366,8 @@ class CommandRunner implements EventDispatcherInterface
         try {
             return $command->run($argv, $io);
         } catch (StopException $e) {
+            $io->error($e->getMessage());
+            
             return $e->getCode();
         }
     }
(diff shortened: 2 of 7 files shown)


dereuromark:
The 1 remaining part in ShellDispatcher has to stay the old way it seems:
```php
        try {
            $result = $this->_dispatch($extra);
        } catch (StopException $e) {
            return $e->getCode();
        }
```

dereuromark:
I dont like the 
```
    public function abort($code = self::CODE_ERROR)
    {
        throw new StopException('Command aborted', $code);
    }
```
part that makes it not possible to properly set the message here :/

src/Console/Command.php
```diff
@@ -277,6 +277,8 @@ public function executeCommand($command, array $args = [], ConsoleIo $io = null)
         try {
             return $command->run($args, $io);
         } catch (StopException $e) {
+            $io->error($e->getMessage());
```
markstory: Couldn't this result in duplicate error messages in existing use cases?

dereuromark: Yes, that is what's happening and why I think there is an issue for this to be fixed in 3.x now.
Still sad that we can't use the exception message for what it is supposed to be used for.

markstory: I didn't expect people to directly throw `StopException` from their command classes. Using the `abort()` helper removes the need for additional imports and results in more terse code.

dereuromark:
Anything here we want or can do for 4.x then?
Or do we need to close this as an acceptable way of how the CLI part works?
Ignoring the message in the exception in all cases basically?

markstory:
I think we can make these changes in 4.0. We'd also need to update `io->abort()` so that the error message isn't emitted twice.

dereuromark:
Do we want to continue this? If so, whats the best approach?

markstory:
If we can get this into a place where messages aren't emitted twice then it could be merged into 3.next or 4.next.

PR #13590, opened 2019-09-01
Backport expires part.

dereuromark:
Backport https://github.com/cakephp/cakephp/pull/13585/files#diff-fcb1f8437cdd541f36c6f48eff4796f1R270

I wasn't sure about whether this is fine for master, so I targeted 3.next.


ADmad:
> I wasn't sure about whether this is fine for master, so I targeted 3.next.

Yup better to err on the side of caution.

PR #13591, opened 2019-09-01
Remove untested and unneeded feature.

ADmad:
`TestCase::$fixtures` is declared as array in `4.x` and there's no real need to
support declaring fixtures list as string.

### DIFF
--- a/src/TestSuite/Fixture/FixtureManager.php
+++ b/src/TestSuite/Fixture/FixtureManager.php
@@ -100,9 +100,7 @@ class FixtureManager
         if (empty($test->fixtures) || !empty($this->_processed[get_class($test)])) {
             return;
         }
-        if (!is_array($test->fixtures)) {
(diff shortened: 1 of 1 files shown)


dereuromark (approved):

dereuromark:
I recommend triggering a deprecation notice for this in 3.next
This helps to migrate away from this by knowing sth will break in 4.0.

ADmad:
> I recommend triggering a deprecation notice for this in 3.next

Sounds good. While we have agreed not to generate anymore warnings in 3.next this is for testsuite so it's okay. Also the chances of sometime using a string to declare the fixtures list is pretty slim.

dereuromark:
Some deprecations are still valid to be added to make it possible to more easily migrate.
This is sure one of those cases :)

markstory:
I think a hard break is reasonable here. We've not had the string form of fixtures documented in many many years.

PR #13595, opened 2019-09-01
Fix type errors reported by psalm.

ADmad:


src/Datasource/RepositoryInterface.php
```diff
@@ -65,10 +65,10 @@ public function hasField(string $field): bool;
      * type of search that was selected.
      *
      * @param string $type the type of query to perform
-     * @param array|\ArrayAccess $options An array that will be passed to Query::applyOptions()
+     * @param array $options An array that will be passed to Query::applyOptions()
      * @return \Cake\Datasource\QueryInterface
      */
-    public function find(string $type = 'all', $options = []);
+    public function find(string $type = 'all', array $options = []);
```
ADmad: While this is technically a BC break using `ArrayAccess` never actually worked. We don't have any tests for it and the `$options` eventually get passed to `applyOptions()` which does `ksort()` and iterates on the options, neither of which are possible for an `ArrayAccess` instance.

I updated the argument type for `RepositoryInterface:find()` also to keep things in sync.

markstory:
Looks good, but there is one psalm error in scrutinizer.

ADmad:
I fixed the error and added few more fixed.

There's one actual error which I am not sure how to deal with:

https://github.com/cakephp/cakephp/blob/77814cb12a3de3bd7c5da802bba8be9e0cc749b7/src/ORM/Association/BelongsToMany.php#L618

The `$hasMany->getConditions()` can return a closure, in which case the above would generate an error. How to handle this?

markstory:
That's a tricky one. We could skip the merge in the case the association conditions are not an array. It looks like `Database\Query::_conjucate()` and `QueryExpression::add()` know how to handle callables inside a list.


```php
$associationConditions = $hasMany->getConditions();
if (is_array($associaticationCondition)) {
  $conditions = array_merge($conditions, $associationConditions);
} else {
  $conditions[] = $associationConditions
}
```

Something like this _might_ work.

src/ORM/EagerLoadable.php
```diff
@@ -59,7 +59,7 @@ class EagerLoadable
      * A dotted separated string representing the path of associations
      * that should be followed to fetch this level.
      *
-     * @var string|null
+     * @var string
```
dereuromark: Is this really true at runtime? After constructor is done?

dereuromark:
Looks good so far.

src/ORM/EagerLoadable.php
ADmad: Yes, the constructor takes an array of config and that's used to set various properties one of which is this.

PR #13600, opened 2019-09-02
Fix errors reported by psalm for Database and Datasource packages.

ADmad:
Phew!


src/Database/Driver.php
```diff
@@ -32,7 +32,7 @@ abstract class Driver implements DriverInterface
     /**
      * Instance of PDO.
      *
-     * @var \PDO|null
+     * @var \PDO
```
ADmad: While this property is not set in constructor I chose to remove the `|null` to avoid having to replace all `$this->_connection` usage with `$this->getConnection()` calls in all drivers. With numerous usage of PDO connection instance during a request cycle the methods calls (which have no null) checks could end up adding a non trivial amount of overhead.

src/Database/Type/DateTimeType.php
```diff
@@ -105,7 +107,7 @@ public function __construct($name = null)
     /**
      * Convert DateTime instance into strings.
      *
-     * @param string|int|\DateTime|\DateTimeImmutable $value The value to convert.
+     * @param mixed $value The value to convert.
```
dereuromark: I am really not a big fan of reverting this. Some effort went into this to make sure we have introspection here on different levels to make sure the "expected" types match. If you, in tests or production, use invalid values to create noops or even exceptions, that is a different thing that we should not use as reason to compromise the good cases for having the expected values annotated here.

ADmad: The type is set as `mixed` for the interface so static analysers are unhappen if you set specific type in the implementation classes. These methods are meant to deal with user input which doesn't have specific type. Also these methods are almost never going to be used directly by end user so types shown by IDEs also isn't really a consideration.

src/Database/Statement/PDOStatement.php
```diff
@@ -131,6 +145,10 @@ public function fetchAll($type = parent::FETCH_TYPE_NUM)
             return $this->_statement->fetchAll(PDO::FETCH_OBJ);
         }
 
+        if (!is_int($type)) {
+            throw new Exception('Fetch type for PDOStatement must be an integer');
```
markstory: Should we include the provided type here?

ADmad:
@markstory Requested changes done. 

The pslam failure shown by scrutinizer is unrelated, I'll sort it out later if needed.

src/Database/Statement/BufferedStatement.php
```diff
@@ -155,7 +156,7 @@ public function execute(?array $params = null): bool
     public function fetchColumn(int $position)
     {
         $result = $this->fetch(static::FETCH_TYPE_NUM);
-        if (isset($result[$position])) {
+        if ($result && isset($result[$position])) {
```
lorenzo: can you make this `$result !== false && ...` ?

ADmad: Done

PR #13601, opened 2019-09-03
4.x psalm

ADmad:


src/Log/Engine/ConsoleLog.php
```diff
@@ -77,10 +77,11 @@ public function __construct(array $config = [])
     /**
      * Implements writing to console.
      *
-     * @param string $level The severity level of log you are making.
+     * @param mixed $level The severity level of log you are making.
```
dereuromark: why mixed?

ADmad: Because of https://github.com/php-fig/log/blob/c4421fcac1edd5a324fda73e589a5cf96e52ffd0/Psr/Log/LoggerInterface.php#L116.

dereuromark: That's just crazy. The truth is string|int right?
Can't we properly narrow it down?
Having mixed here is just a lie as you cant pass objects or float or alike. It is just not valid/expected in any way.

dereuromark: Those are valid cases to have ignores in the static analyzer list. The base code is clearly not pragmatic/usable on this. As such we should keep it specific and adjusted to the truth for us.

ADmad: Okay, I'll revert and ignore the errors. 

Do you want to try get it changed to `int|string` for the PSR interface so that we can remove the ignore in future? :stuck_out_tongue:

dereuromark: ;) I trust in your politically sensitive way of approaching this.

ADmad: Currently you can only pass levels as string to our loggers. I would like to allow using `LOG_*` constants too for the `$level` argument. So let's keep the `mixed` for now and I'll make a separate PR which makes required code changes and change docblock type to `int|string`.

ADmad:
I am done for this round. Will make a separate PR for further updates.

src/TestSuite/ConsoleIntegrationTestTrait.php
```diff
@@ -66,21 +66,21 @@ trait ConsoleIntegrationTestTrait
...
      *
-     * @var \Cake\TestSuite\Stub\ConsoleOutput|\PHPUnit\Framework\MockObject\MockObject|null
+     * @var \Cake\TestSuite\Stub\ConsoleOutput
      */
     protected $_err;
 
     /**
      * Console input mock
      *
-     * @var \Cake\Console\ConsoleInput|\PHPUnit\Framework\MockObject\MockObject|null
```
markstory: Weren't these added to fix an IDE? Are these still needed?

src/TestSuite/ConsoleIntegrationTestTrait.php
```diff
@@ -105,6 +105,7 @@ public function exec(string $command, array $input = []): void
         try {
             $this->_exitCode = $runner->run($args, $io);
         } catch (StopException $exception) {
+            /** @var int $this->_exitCode */
             $this->_exitCode = $exception->getCode();
```
markstory: Should we use a cast instead of a comment?

src/View/Helper/PaginatorHelper.php
```diff
@@ -296,7 +296,7 @@ protected function _toggledLink($text, $enabled, $options, $templates): string
         $text = $options['escape'] ? h($text) : $text;
 
         $templater = $this->templater();
-        $newTemplates = !empty($options['templates']) ? $options['templates'] : false;
+        $newTemplates = $options['templates'] ?? false;
```
markstory: Do we care if templates are `''`?

src/TestSuite/ConsoleIntegrationTestTrait.php
ADmad: The `exec()` method set's the property to concrete class `\Cake\Console\ConsoleInput` so the `|\PHPUnit\Framework\MockObject\MockObject` isn't necessary.

src/TestSuite/ConsoleIntegrationTestTrait.php
ADmad: All our exception classes do actually return int so a casting is necessary, hence the ignore. I'll  try and get the issue sorted out with psalm itself to allow us to override the return type for our `Cake\Core\Exception\Exception` class.

src/View/Helper/PaginatorHelper.php
ADmad: I made `StringTemplate::load()` an exception instead if argument is empty string.

src/TestSuite/ConsoleIntegrationTestTrait.php
ADmad: The pslam issue is fixed https://github.com/vimeo/psalm/issues/2095.

So once a new release is done we can remove all unnecessary annotations / casting  where `getCode()` is used :slightly_smiling_face:.

Issue #13570, opened 2019-08-29
RFC: Generate URLs using compact action name


garas:
This is a (multiple allowed):

* [ ] bug
* [x] enhancement
* [x] feature-discussion (RFC)

* CakePHP Version: 3.8

### Current state

Since 3.6.0, route builder supports specifying target in `[Plugin].[Prefix]/[Controller]::[action]` compact string format
```php
$routes->connect('/admin/cms/articles', 'Cms.Admin/Articles::index');
```

To generate to such URL, it still necessary to use expanded array
```php
Router::url(['prefix' => 'admin', 'plugin' => 'Cms', 'controller' => 'Articles', 'action' => 'index']);
```

Some keys can be skipped and determined from current request context.

But it can generate unexpected URLs. Ex. same email template generates different URL for `['controller' => 'Articles', 'action' => 'index']` when sent from frontend and admin backend, and all links must have `'prefix' => false` to avoid this.

There is also problem locating all links to the action when renaming/removing action. `'action' => 'index'` without prefix/plugin/controller might mean this or another controller and requires extra thinking.

Option `_name` helps to shorten, but it needs to be explicitly specified when connecting. It also often just duplicates target action.
```php
$routes->connect('/admin/cms/articles', 'Cms.Admin/Articles::index', ['_name' => 'cms:admin:articles:index']);

Router::url(['_name' => 'cms:admin:articles:index']);
```

### Suggestion
I suggest to extend `action` key functionality and allow specifying everything using same `[Plugin].[Prefix]/[Controller]::[action]` format.
```php
Router::url(['action' => 'Cms.Admin/Articles::index']);
```

Internally, if `action` contains `::`, first expand `['action' => 'Cms.Admin/Articles::index', ...]` to full `['prefix' => 'admin', 'plugin' => 'Cms', 'controller' => 'Articles', 'action' => 'index', ...]`, then continue as usually.

It should fail (throw exception) if prefix, plugin or controller specified too.

It would be always absolute, i.e. `['action' => 'Articles::index']` would mean `['prefix' => false, 'plugin' => null, 'controller' => 'Articles', 'action' => 'index']` even if used in prefix or plugin context.

It would let avoid specifying `_name` in most cases, it would avoid relative route array ambiguity, and it would be in single canonical searchable form.

### Further possibilities

[...]

PR #13605, opened 2019-09-04
4.x: Generate URLs using short string syntax

garas:
See  #13570.

Generate URLs using short string syntax:

```php
Router::pathUrl('Articles::index']);
Router::pathUrl('Articles::view', [3]);
Router::pathUrl('Articles::read', ['slug' => 'title']);
Router::pathUrl('Admin/Articles::index');
Router::pathUrl('Cms.Admin/Articles::index');
Router::pathUrl('Vendor/Cms.Management/Admin/Articles::view']);
```

Also in array within `_path` key. Also works in `HtmlHelper::link()`, `UrlHelper::build()` etc:
```php
Router::url(['_path' => 'Articles::index']);
Router::url(['_path' => 'Articles::view', 3]);
Router::url(['_path' => 'Articles::read', 'slug' => 'title']);
Router::url(['_path' => 'Admin/Articles::index']);
Router::url(['_path' => 'Cms.Admin/Articles::index']);
Router::url(['_path' => 'Vendor/Cms.Management/Admin/Articles::view']);
```


dereuromark:
One question:
What happens with colliding information, like _path + action etc?

    Router::url(['_path' => 'Articles::view', 'action' => 'foo', 3]);

Which wins? view or foo? Should this silently take e.g. view or throw meaningful InvalidArgumentException etc?

garas:
Usage of any of `'plugin', 'prefix', 'controller', 'action'` with  `_path` is forbidden by ``throw new RuntimeException("The `_path` string cannot be overriden by `action` key.");``.

src/Routing/Router.php
```diff
@@ -924,4 +928,29 @@ public static function setRouteCollection(RouteCollection $routeCollection): voi
...
+    /**
+     * Inject route defaults from `_path` key
+     *
+     * @param array $url Route array with `_path` key
+     * @return array
+     */
+    protected static function unwrapShortString(array $url)
+    {
+        foreach (['plugin', 'prefix', 'controller', 'action'] as $key) {
+            if (array_key_exists($key, $url)) {
```
dereuromark: Wouldn't isset() suffice? As it shouldn't be non-null for it to become an issue right?

src/Routing/Router.php
```diff
@@ -924,4 +928,29 @@ public static function setRouteCollection(RouteCollection $routeCollection): voi
...
+                throw new RuntimeException("The `_path` string cannot be overriden by `$key` key.");
+            }
+        }
+
+        $url += RouteBuilder::parseShortString($url['_path']);
+        $url += [
+            'plugin' => null,
+            'prefix' => false,
```
dereuromark: Why are we using plugin and prefix differently? I would assume in 4.x both to be using the same "falsely" value, probably `null` as cleaner one.

src/Routing/Router.php
dereuromark: For isset() and possible key existance we might need

    $url += RouteBuilder::parseShortString($url['_path']);

to be

    $url = RouteBuilder::parseShortString($url['_path']) + $url;

though.

garas: Should then `Router::url(['_path' => 'Cms.Articles::index', 'plugin' => null]);` silently ignore `null` and use plugin 'Cms'?

src/Routing/Router.php
garas: `plugin`: `null` is used everywhere.

`prefix`: `false` is used in docs, mix of `false`/`null` in test cases.

ADmad: `false` and `null` for `prefix` have different meanings. `null` will use current prefix if available, `false` will always unset/disable prefix.

src/Routing/Router.php
dereuromark: That depends on if null is the same meaning as false. The latter seems like a deliberate deactivation, whereas the null can be just default case from nullable setup, without meaning.
We might have to clarify what we want here as behavior of those two values.

src/Routing/Router.php
dereuromark: Given this, shouldn't prefix then also be null by default?

garas: It should stay `false` here.

One of purpose of this PR is to avoid specifying `prefix => false` in reusable elements/cells/emails/pdfs, so it removes plugin/prefix from current request.

I would like avoid workaround I sometimes use: in email, use links without `prefix => false` , but replace current request while sending email:
```php
$params = $controller->getRequest()->getAttribute('params');
unset($params['prefix']);
Router::pushRequest($controller->getRequest()->withAttribute('params', $params));

// send email

Router::popRequest();
```

dereuromark: But why then not for plugin too?

src/Routing/Router.php
garas: I will then remove override check here, so it could be overridden. If there is no override, it shouldn't use current request.

`['_path' => 'Articles::index', 'action' => 'view]` same as `'Articles::view'`
`['_path' => 'Cms.Articles::index', 'plugin' => null]` same as `'Articles::index'`
`['_path' => 'Cms.Articles::index', 'prefix' => 'admin']` same as `'Cms.Admin/Articles::index'`

dereuromark: @markstory What do u think?

src/Routing/Router.php
garas: Plugin can be removed by both null and false.
I've never seen `plugin => false` anywhere and I didn't know `false` works too.

ADmad: If you want to generate a non plugin URL from within a plugin you would need to use `plugin => false`

src/Routing/Router.php
saeideng: dont allow to override please

src/Routing/Router.php
garas: I've updated to `'plugin' => false` here

src/Routing/Router.php
garas: I left exception to not allow overriding. Only changed to `InvalidArgumentException`.

I replaced to `isset()` check. `null` overrides will be silently ignored.

src/Routing/Router.php
```diff
@@ -924,4 +929,29 @@ public static function setRouteCollection(RouteCollection $routeCollection): voi
...
+     * Inject route defaults from `_path` key
+     *
+     * @param array $url Route array with `_path` key
+     * @return array
+     */
+    protected static function unwrapShortString(array $url)
+    {
+        foreach (['plugin', 'prefix', 'controller', 'action'] as $key) {
+            if (isset($url[$key])) {
+                throw new InvalidArgumentException("`$key` cannot be used to override `_path` value.");
```
markstory:
```suggestion
                throw new InvalidArgumentException("`$key` cannot be used when defining routes targets with `_path`.");
```

src/Routing/Router.php
```diff
@@ -924,4 +929,29 @@ public static function setRouteCollection(RouteCollection $routeCollection): voi
...
         static::$_collection = $routeCollection;
     }
+
+    /**
+     * Inject route defaults from `_path` key
+     *
+     * @param array $url Route array with `_path` key
+     * @return array
+     */
+    protected static function unwrapShortString(array $url)
```
markstory: Should we cache the results of this method into an in-memory array that is cleared by `Router::reload()`? I can see this method getting hit quite a few times in a request and doing redundant work.

garas: For each `Router::url()` call with `_path` key, parsing string will be done once. Further route matching at collection takes much more work.

markstory: What about pages that have 20+ links on them? We'll be going through this block of code 20+ times and we could pretty easily cache the results.

src/Routing/Router.php
```diff
@@ -159,6 +159,13 @@ class Router
      */
     protected static $_defaultExtensions = [];
 
+    /**
+     * Cache of parsed short strings
+     *
+     * @var array
+     */
+    protected static $_shortStrings = [];
```
dereuromark: Do we need a way to clear those? For test runs etc? Or is this sufficient as is?

garas: `Router::reload()` resets all static properties.

ADmad:
The lazy dev in me thinks that having to use array with `_path` is still something that should be avoided :slightly_smiling_face:

Could we perhaps have a special prefix to the path which allows using string itself like `Router::url('__::Articles::index')`.

dereuromark:
With this you start a prefix war^^

```
Router::url('~Articles::index')
Router::url('^Articles::index')
Router::url('_Articles::index')
```

`?`, `#`, `&`, `/` and `:` etc are usually out of the game.

garas:
But what about `Router::url(['_path' => 'Articles::view', 3]);` with additional parameters?

dereuromark:
@garas, In that case, you can't/don't want to use additional params probably.

ADmad:
@garas We could implement something like this if we want to

 ```php
Router::url('Articles::read', ['slug' => 'foo'])
```

**Comment by ADmad** [comment]
I had a quick look at symfony and laravel and in both the method to general URL takes route/path name as 1st argument and 2nd argument is array for additional route/path elements.

**Comment by ADmad** [comment]
Do we even need a special prefix to denote that path string? If string starts with protocol it's a full URL, if it start with '/' its an in app URI and thing else if either route path/name isn't it?

Personally I have never used relative string URIs without starting '/` like `Router::url('foo/bar')`.

**Comment by dereuromark** [comment]
@ADmad While I agree with you, it could be cheaper to check without having to regex the string. By using the _path element or the prefix it is usually a much cheaper operation (isset is the fastest), especially relevant with tons of links on a page. Just mentioning this as this could factor in into the decision here.

**Comment by ADmad** [comment]
@dereuromark Let the user decide whether he prefers convenience of use or cheaper operation. Also the regex is required for the new path syntax. If one uses route name no regex check would be required and we could support usage like `Router::url('_myRouteName', ['slug' => 'foo'])` without any overhead.

**Comment by ADmad** [comment]
It would be great if we could auto generate names for routes as @markstory suggested here https://github.com/cakephp/cakephp/issues/13570#issuecomment-526428963.

File: src/Routing/RouteBuilder.php [hunk 6]
```diff
@@ -698,27 +698,47 @@ protected static function parseDefaults($defaults): array
...
+     * String examples:
+     * - Bookmarks::view
+     * - Admin/Bookmarks::view
+     * - Cms.Articles::edit
+     * - Vendor/Cms.Management/Admin/Articles::view
+     *
+     * @param string $routeString Short string in [Plugin.][Prefix/]Controller::action format
+     * @return string[]
+     */
+    public static function parseShortString(string $routeString): array
```
ADmad: Name like `parseRoutePath()` would be more appropriate IMO.

garas:
I've updated to use string directly in first argument.

```php
Router::url('Articles::index');
Router::url('Articles::view', [3]);
Router::url('Articles::read', ['slug' => 'title']);
```

Currently it is checked if first char is `[a-z0-9]` and contains `::`.

It would be safer however with some prefix. My suggestion would be `Router::url('@Articles::index');`

ADmad:
> It would be safer however with some prefix. My choice would be Router::url('@Articles::index');

I agree. We could after another prefix char to denote use of route name and there by avoid having to use `['_name' => 'foo']`.

src/Routing/Router.php
```diff
@@ -488,12 +497,47 @@ public static function url($url = null, bool $full = false): string
...
+            }
+
+            if (ctype_alnum($url[0]) && strpos($url, '::') > 0) {
+                $url = RouteBuilder::parseRoutePath($url) + ['plugin' => false, 'prefix' => false];
+
+                if (is_array($full)) {
+                    foreach (['plugin', 'prefix', 'controller', 'action'] as $key) {
+                        if (isset($full[$key])) {
+                            throw new InvalidArgumentException(
+                                "`$key` cannot be used when defining routes targets with route path string."
```
markstory:
```suggestion
                                "`$key` cannot be used when defining route targets with a string route path."
```

tests/TestCase/Routing/RouterTest.php
```diff
@@ -3116,6 +3117,125 @@ public function testConnectShortStringSyntax()
...
+
+        $result = Router::url('Articles::view', [3]);
+        $expected = '/articles/view/3';
+        $this->assertEquals($result, $expected);
+
+        $result = Router::url('Articles::read', ['slug' => 'title']);
+        $expected = '/article/title';
+        $this->assertEquals($result, $expected);
+
+        $result = Router::url('Admin/Articles::index');
```
markstory: I wonder if we should also support bare strings for named routes 🤔 . It should be a separate set of changes after these ones.

tests/TestCase/Routing/RouterTest.php
```diff
@@ -3116,6 +3117,125 @@ public function testConnectShortStringSyntax()
...
+     * @param array $params
+     * @return void
+     * @dataProvider invalidShortStringArrayProvider
+     */
+    public function testUrlGenerationOverridingShortString(array $params)
+    {
+        Router::connect('/articles', 'Articles::index');
+
+        $this->expectException(InvalidArgumentException::class);
+        $this->expectExceptionMessage('cannot be used when defining routes targets with route path string.');
```
markstory:
```suggestion
        $this->expectExceptionMessage('cannot be used when defining route targets with a string route path.');
```

src/Routing/Router.php
```diff
@@ -488,12 +497,47 @@ public static function url($url = null, bool $full = false): string
...
+                strpos($url, '?') === 0 ||
+                strpos($url, '//') === 0 ||
+                strpos($url, '://') !== false
+            );
+
+            if ($plainString) {
+                return $url;
+            }
+
+            if ($url[0] === '@' && strpos($url, '::') > 1) {
```
ADmad: I am wondering if instead of using `::` as separator between controller and action names could be use `@` instead, like `Articles@view`?

By doing so we would just have to check if string has `@` character to determine if it's "path" string. We could use the `@` prefix instead to denote a named route. So the check would become like:

```php
$pos = strpos('@', $url);
// It's a named or path route
if ($pos !== false) {
    // It's a named route
    if ($pos === 1) {
    }
}
```

src/Routing/RouteBuilder.php
```diff
@@ -692,33 +692,13 @@ public function connect($route, $defaults = [], array $options = []): Route
      * @param string|array $defaults Defaults array from the connect() method.
      * @return array
      */
-    protected static function parseDefaults($defaults): array
+    protected function parseDefaults($defaults): array
```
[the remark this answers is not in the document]
garas: `RouteBuilder` has only non-static methods, `Router` has only static methods.

Method is also invoked with `$this->parseDefaults($target);` since introducing, so it was likely made static erroneously.

src/Routing/Router.php
dereuromark: The @ is a bit harder to type though on most Keyboards.

garas: `::` is already used by `connect()`, so I wouldn't like one more format. 

I would prefer different prefix for named routes.

ADmad: > The @ is a bit harder to type though on most Keyboards.

You already have to type @ as prefix in current proposal, so I don't see how your argument helps :) My proposal avoids having to type two additional colons.

> :: is already used by connect(), so I wouldn't like one more format.

Ah right, maintaining similarly when setting up routes and generating URLs makes sense.

But with the @ prefix is the additional overhead of strpos() call to check for 2 colons necessary?

dereuromark: Isnt the @ optional and you can still use the _path key?
Also once is different than multiple times.

ADmad: > Isnt the @ optional and you can still use the _path key?

There's no `_path` key. The consensus was to directly pass string as 1st argument to `Router::url()` and using array with key is not convenient enough.

> Also once is different than multiple times.

I proposal was to remove `@` as 1st character and replace `::` with `@`, so it's still only once.

garas: Problem with string as 1st argument is how to use this from `HtmlHelper::link()`, `UrlHelper::build()` etc, as there is no way to pass additional params with string.

I'm thinking if I need to return `_path` as alternate so those methods could use array but that would one more way to do same...
```php
HtmlHelper::link('View', ['_path' => 'Articles::view', 3]);
HtmlHelper::link('View', Router::url('Articles::view', [3]));
```

 Any other ideas?

ADmad: @garas It seems we'll have to add back the `_path` key then. `['_path' => 'Articles::view', 3]` is still more convenient than having to use `Router::url('@Articles::view', [3])`.

markstory: Would using a format like `['Articles::view', [3]]` work? The tuple of string with `::` in the 0th slot and an array in the 1st slot could help us detect short names with parameters.

garas: I like tuple format, more compact and no magic `_path` string.

dereuromark: What if you omit 2nd argument? Only path argument, the first?

garas: Then string only `@Articles::index`, no array.

markstory: If we can skip the `@` that would be ideal as we don't use `@` when connecting routes and only having one syntax would be :+1: It should be rare for an existing URL to match both the new short syntax and be a useful URL.

Instead of making the new short syntax complex to preserve compatibility with what should be a rare situation could we provide an escape hatch for this situation? Perhaps something like

```php
Router::url(['_path' => 'MyThing:photo'])
```

Which is treated like a bare string URL would have in the past?

garas: So which of these it should support? With or without `@` prefix?

1. Bare string. Cleanest, but extra params not available for `HtmlHelper::link()`. Safer with `@` prefix.
```php
Router::url('Articles::index');
Router::url('Articles::view', [3]);
```
2. Tuple format. Cleanest for functions like `HtmlHelper::link()`.
```php
Router::url(['Articles::index', []]); // empty extra params array required
Router::url(['Articles::view', [3]]);
```
3. Using special key. Safest. Very similar to `_name`.
```php
Router::url(['_path' => 'Articles::index']);
Router::url(['_path' => 'Articles::view', 3]);
```

All variants has pros and cons.

dereuromark: 1 with the prefix and 3 could both be valid and supported, right?
I don't think 2 is good due to the decreased usability.

ndm2: Ok let, me take the "path part only" back, looks like currently the "plain string" check looks for `?` and `#` at position `0`. But testing only the path part for `::` should be easy enough :)

markstory: Combining 1 and 3 would let us handle 0 and many parameters and shouldn't overlap with existing signature forms. We'd need to add integration tests to each place where we generate URLs though to make sure with any approach we choose.

garas:
Now both formats works.
```php
Router::url('Articles::index']);
Router::url('Articles::view', [3]);

Router::url(['_path' => 'Articles::index']);
Router::url(['_path' => 'Articles::view', 3]);
HtmlHelper::link('Title', ['_path' => 'Articles::index']);
HtmlHelper::link('Title', ['_path' => 'Articles::view', 3]);
```

Slash prefixed `Router::url('/Articles::index']);` will not match.

markstory:
Do we need the `_path` key to be able to identify short URLs? The `['Articles::view', [3]]` form is a bit simpler to type.

dereuromark:
See discussion above. The tuple form is a bit bulky in usability.

ADmad:
I find `['_path' => 'Articles::view', 3]` to be more "bulky" to type than `['Articles::view', [3]]`.

dereuromark:
The 2nd array is required here, which is not really user friendly or easy to guess.

ADmad:
I do agree that for a new user reading the code the format would be not obvious.

dereuromark:
Thus short string form + _path array one seem to be a good compromise.

src/Routing/Router.php
ndm2: I don't want anyone to blame me later on, so I had a further look at using `/` for relative URLs to avoid conflicts :upside_down_face:  It looks like it shouldn't be a problem for non asset URLs, as relative URLs processed by the router get a `/` prepended anyways.

For assets, `Router::url()` seems to be only used by the helper for URLs where a scheme is present, that is URLs containing either `://`, `a-z0-9://`, or `a-z:`. The latter currently poses a limitation to using `::`, it would currently only work when the part before the `::` would contain non-`a-z0-9` chars, otherwise it would be interpreted as a URI scheme, ie `foo/bar::baz` would work and for example map to `css/foo/bar::baz.css`, while `foo::bar` would not work, as `foo:` would be interpreted as a scheme.

Long story short, so far the only string URLs that I found that might not be easily upgradable to avoid possible conflicts, would be asset URLs like `foo/bar::baz` (one could of course do `/css/foo/bar::baz`, but that would mean loosing path prefixing), and ones that have a scheme but don't use `://`, for example `cid:my::asset::id`.

garas: Assets should now use `Asset` class. 

Docs could be updated to recommend use `Router::url()` for MVC type actions first, also keep support what currently checked with `$plainString`, other bare string uses should avoided. `Router::url()` is now used basically for any type of URL, should be less usages.

`Asset` itself should not use `Router::url()` - it currently uses it with scheme URLs, but URL is just returned unmodified via `$plainString` check. And the `a-z:` part doesn't not work when base URL is used: `foo:bar` becomes `base/foo:bar` (except `mailto/javascript/tel/sms`) - I'm not sure if this behavior is intended.

markstory:
Should we support the 

```php
Router::url('Articles::index');
Router::url('Articles::view', [3]);
HtmlHelper::link('Title', 'Articles::index');
```

Forms if we are going to require a `_path` key elsewhere? Having multiple ways of using the short path is going confuse people. If we always required the `_path` key our code would be simpler and application developers would have a single way of using short paths.

We could add another method to `Router` if we want to also support the string + array form. Perhaps `Router::path('Articles::index', [3])`

dereuromark:
I think it should rather be named `urlByPath()` then to make clear this returns the URL, not some path.
But yeah, then we could just keep _path part for normal url() method. 
Sounds fair.

markstory:
The `urlByPath()` method is a reasonable option, but doesn't seem like it would help with Helper scenarios which are the more challenging part of the short URL change.

The `_path` key is a good option as it builds on the `_name` approach and allows us to support new path urls in all the helpers and various places that routes are resolved. An alternative would be to add a new interface like you've described and require the output of that method to be passed into helpers.

```php
// Instead of this
$this->Html->link('Click me', ['_path' => 'Articles::index']);

// We would do this
$this->Html->link('Click me', Router::pathUrl('Articles::index'));
```

This new method would work well for forms, but be more challenging when we have to generate URLs for pagination.

dereuromark:
The advantage of own method with clean API would also be that IdeHelper can autocomplete a list of possible controller::action strings :)
It can parse the code base and make such a list to chose from here using .meta file, at least for larger IDEs this makes now URL typehinting a piece of Cake then!

garas:
Separate method is safe and could be backported safely to 3.9. And requires modifying `Router::url()` much less.

IDE autocomplete would also nice bonus.

`$this->Html->link('Click me', Router::pathUrl('Articles::index'))` is not always safe as base path will be applied twice by `Router::url()` (`/basepath/articles` -> `/basepath/basepath/articles`).

garas:
I've simplified to array `_path` key for use in `HtmlHelper::link()`, `UrlHelper::build()` etc
```php
Router::url(['_path' => 'Articles::index']);
Router::url(['_path' => 'Articles::view', 3]);
```

And wrapper method
```php
Router::pathUrl('Articles::index']);
Router::pathUrl('Articles::view', [3]);
```

It could be also similar `Router::namedUrl()` for `_name` key.

dereuromark:
Is there then in view templates Url and Html helper methods for both as well? Or just use _path there?

garas:
What name and signature would you suggest for `HtmlHelper::pathLink` ?

```
HtmlHelper::link( string|array $title , string|array|null $url = null , array $options = [] );
HtmlHelper::pathLink( string $title , string $path, array $params = [] , array $options = [] );

$this->Html->pathLink('Click me', 'Articles::index');
$this->Html->pathLink('Click me', 'Articles::view', [3]);

//but with options and no params
$this->Html->pathLink('<img> Click me', 'Articles::index', [], ['escapeTitle' => false]);
```

markstory:
For Helpers would it be unreasonable to use a type for route targets instead of duplicating all the methods?

Something like 

```php
$this->Html->link('Click me', Url::make('Articles::view', 3));
```

The name `Url::make()` could be replaced with a global function or another class. The output of that function is a simple value object that allows helpers/Router to choose the right method to call.

dereuromark:
The static call or any other such workaround could proof a bit cumbersome in templates - with use statements needed or otherwise.
While I agree, that duplicating methods is not ideal, the wrapping could also become an issue, especially around accidental duplications of /prefix/paths for people not using root (sub)domains.

It might make more and more sense to use such a '@' prefix etc here in view layer at least, that would then go through Router::pathUrl(). 

I would have to check if IDE autocomplete here is possible on controller:action pairs for Html->url() 1st arg and Html->link() 2nd argument as they currently are.
But having a typehinted string input usually is a bit cleaner none the less for such a new string path.

markstory:
>  While I agree, that duplicating methods is not ideal, the wrapping could also become an issue, especially around accidental duplications of /prefix/paths for people not using root (sub)domains.

Not ideal is a bit of an understatement. Having double the number of methods on helpers is a significant amount of maintenance overhead. With a value object we could consolidate the condition in `Router::url()` or in `UrlHelper::build()`.

dereuromark:
True. How would a value object work?

markstory:
We would need new methods/classes to create the value object. Some ideas that come to mind are:

```php
// Factory on Router. Saves having to import additional classes.
// Other name options: make, resolve, generate
$url = Router::build('Articles::view', [1]);

// Class method on value object. Router stays compact, but additional imports needed
use Cake\Routing\Url;
$url = Url::create('Articles::view', [1]);

// Use the value object in a URL.
Router::url($url);
```

Using an actual value object is fairly high risk in my opinion as it has the opportunity to impact all helpers and other places URLs are used. Instead these factory methods could return arrays using the `_path` key, or the short path could be resolved by the factory method. These approaches would us give a terse API to end users without creating a potentially large breaking change to both plugins and userland code.

dereuromark:
@garas 
Regarding your examples
```php
Router::url(['_path' => 'Cms.Admin/Articles::index']);
Router::url(['_path' => 'Vendor/Cms.Management/Admin/Articles::view']);
```

I use a similar way for TinyAuth for quite some time already.
Here the string URLs are `Cms.admin/Articles` and `Vendor/Cms.management/admin/Articles` though.
This also makes way more sense for me, since the prefixes are underscore/camelBacked, and never uppercase on the first letter, right? And we should avoid additional inflection usually:

```php
Router::url(['_path' => 'Cms.admin/MyControllerName::myActionName']);
Router::url(['_path' => 'Vendor/Cms.management/admin/MyControllerName::myActionName']);
```

garas:
`Cms.Admin/Articles::index` case was used by `RouteBuilder::connect()` which I've reused. It also matches PHP class name case `Cms\Controller\Admin\ArticlesController::index()`.

`RouteBuilder::prefix()` inflects string, both upper/lower case should be supported in short string too.

Only inconsistent here is that `FooBar` prefix in short string now is same as `foobar`, so `strtolower()` needs to be changed to `Inflector::underscore()` to match `RouteBuilder::prefix()` behavior and inflect to `foo_bar`.

markstory:
How do people feel about moving forward with what we have here. The `Router::pathUrl()` method gives a simple API to use in non-view contexts, and the `Router::parseRoutePath()` method or the `_path` key can be used in views for now.

We can improve the ergonomics of creating path based URLs in a future release by adding new helpers or global functions.

This is the last key feature that I think we need to land for 4.0.0 to move into an RC which is why I'm asking to compromise on the ergonomics of this API for now given we have the foundational pieces in place.

dereuromark:
Sounds fair

> Only inconsistent here is that FooBar prefix in short string now is same as foobar, so strtolower() needs to be changed to Inflector::underscore() to match RouteBuilder::prefix() behavior and inflect to foo_bar.

Is this part still open? Sounds like this inconsistency might be worth fixing before the merge.

dereuromark:
Awesome, looks great.

PS: https://github.com/dereuromark/cakephp-ide-helper/pull/148 is fully working now in 4.x :)
All basic URLs that can be discovered are now typehinted and autocompletable for Router::pathUrl() calls.

src/Routing/Router.php
```diff
@@ -988,7 +988,7 @@ public static function parseRoutePath(string $url): array
             $defaults['plugin'] = $matches['plugin'];
         }
         if ($matches['prefix'] !== '') {
-            $defaults['prefix'] = strtolower($matches['prefix']);
+            $defaults['prefix'] = Inflector::underscore($matches['prefix']);
```
ADmad: It's a bit odd that values for `controller` and `plugin` are in CamelCase but `prefix` is underscored even though it maps to namespace part like the other two.

garas: I agree CamelCase would be consistent. But it would BC and may be hard to upgrade apps.

There was lowercase prefixes since [1.x](https://book.cakephp.org/1.3/en/The-Manual/Developing-with-CakePHP/Configuration.html#prefix-routing).

This change is to be consistent with `RouteBuilder::prefix()` 
https://github.com/cakephp/cakephp/blob/d68155f315b8fc1bb00ee08df3ba562c1cc51c96/src/Routing/RouteBuilder.php#L866

ADmad: Yeah :disappointed:

ionas: This has bugged me again and again and confused the heck out of me making me waste a lot of hours in routing setup. 4.x is a good moment to fix that having everything ClassNamy but actions als camelCase.

dereuromark: Hehe, same here.
I spend tons of hours also normalizing this for TinyAuth and TinyAuthBackend explaining to me and also to others in docs when it is under_scored and when CamelCased.

We should maybe even think about fixing this before RC for 4.x to be consistent CamelCase always (except for in the URL itself when inflected due to route class).
But should be a follow up to not blocked this being merged now.

markstory: I agree that if we want to switch to CamelCase for prefixes 4.0.0 is the time to do it. However, it doesn't have to be done in this pull request. It can be done separately before the RC.

dereuromark:
IDE support for the new method has been merged to master now too: https://github.com/dereuromark/cakephp-ide-helper/pull/148

PR #13607, opened 2019-09-04
Harmonize default value type across db classes to null.

ADmad:
Currently a mix of `null` and `'string'` was used.

Improve code readability of `Comparison` and `TupleComparison` classes.
Required methods are now overridden in `TupleComparison` instead of trying
to make `Comparison`'s code deal with types array too.

### DIFF
--- a/src/Database/Expression/Comparison.php
+++ b/src/Database/Expression/Comparison.php
@@ -72,15 +72,12 @@ class Comparison implements ExpressionInterface, FieldInterface
      *
      * @param string|\Cake\Database\ExpressionInterface $field the field name to compare to a value
      * @param mixed $value The value to be used in comparison
-     * @param string|array|null $type the type name used to cast the value
+     * @param string|null $type the type name used to cast the value
      * @param string $operator the operator used for comparing field and value
      */
-    public function __construct($field, $value, $type = null, string $operator = '=')
+    public function __construct($field, $value, ?string $type = null, string $operator = '=')
     {
-        if (is_string($type)) {
-            $this->_type = $type;
-        }
-
+        $this->_type = $type;
         $this->setField($field);
         $this->setValue($value);
         $this->_operator = $operator;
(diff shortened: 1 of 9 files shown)


ADmad:
The db type casting process is quite though to follow. It's split across so many files.

ADmad:
I chose to harmonize the default type to `null` instead of `'string'` as `PDOStatment::bindValue()` ultimately converts `null` to `'string'` before converting string types to its `PDO::PARAM_*` constants.

src/Database/Expression/Comparison.php
```diff
@@ -75,7 +75,7 @@ class Comparison implements ExpressionInterface, FieldInterface
      * @param string|array|null $type the type name used to cast the value
      * @param string $operator the operator used for comparing field and value
      */
-    public function __construct($field, $value, $type, string $operator)
+    public function __construct($field, $value, $type = null, string $operator = '=')
```
dereuromark: I wonder if this was more readable if we converted the magic string `=` operator char into a speaking class const?
```php
public const EQUALS = '=';
```
and 
```php
__construct($field, $value, $type = null, string $operator = self::EQUALS)
```

Especially if this char is used a bit more also as input from other sides.
But could also be a bit overkill.

ADmad: This class is too lower level to bother with constants. Also it's not practical to add constants for all possible operators so adding it just for `=` would just create inconsistency.

ADmad:
The phpstan failures are unrelated and will go away after #13612 is merged.

tests/TestCase/Database/ValueBinderTest.php
```diff
@@ -39,7 +39,7 @@ public function testBind()
         $expected = [
             ':c0' => [
                 'value' => 'value0',
-                'type' => 'string',
+                'type' => null,
```
lorenzo: shouldn't there always be a type here?

dereuromark: See above:

> I chose to harmonize the default type to null instead of 'string' as PDOStatment::bindValue() ultimately converts null to 'string' before converting string types to its PDO::PARAM_* constants.

ADmad:
With this cleanup the builds also pass on PHP 7.4 now.

lorenzo (approved):
looks great!

Issue #13598, opened 2019-09-01
4.x - Use Closure instead of callable wherever possible


ADmad:
The benefits of using `Closure` instead of `callable` are mentioned [here](https://wiki.php.net/rfc/closurefromcallable#why_would_you_use_this).

The primary consideration is whether public methods which accept callables should be changed to accept closure only.

ADmad:
This change would be especially beneficial in the db layer for methods which accept arrays as well as callable which makes type checking tricky. It would allow us to do away with methods like this:

https://github.com/cakephp/cakephp/blob/12aaeb157cc58dbb9cbc6aeba6909a9f7e7c0b6a/src/Database/Expression/QueryExpression.php#L602-L623

markstory:
Seems like a postitive change to me. I've never really liked the `isCallable` method and requiring `Closure` prevents several potential attack vectors where we miss an `isCallable` check.

lorenzo:
I personally don’t like this as creating invokable classes is more meaningful that passing plain closures. This would also make it a bit of a hassle for me to update many of my apps.

The benefits can be realized internally if we convert the callable to closures if we plan to invoke in loops.

ADmad:
Done for db layer in #13620

PR #13611, opened 2019-09-04
Replace callable with Closure in DB and ORM layer.

ADmad:
Refs #13598 

Barring couple of places where invokable classes can be used as argument, I have replaced all `callable`s with `Closure` in the `Database` and `ORM` packages :slightly_smiling_face:.

Edit: As stated below some of the changes might not be desirable. The reason I changed all possible  is so that they are highlighted in the diff and we can identify which ones should be changed to Closure and which retained as callable.

### DIFF
--- a/src/Datasource/QueryTrait.php
+++ b/src/Datasource/QueryTrait.php
@@ -19,6 +19,7 @@ namespace Cake\Datasource;
 use BadMethodCallException;
 use Cake\Collection\Iterator\MapReduce;
 use Cake\Datasource\Exception\RecordNotFoundException;
+use Closure;
 use InvalidArgumentException;
 use Traversable;
 
@@ -57,7 +58,7 @@ trait QueryTrait
      * List of formatter classes or callbacks that will post-process the
      * results when fetched
      *
-     * @var callable[]
+     * @var \Closure[]
      */
     protected $_formatters = [];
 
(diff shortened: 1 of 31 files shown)


lorenzo:
Please don’t, I use many invokable clases, that would make my apps and several I have worked on a pain to migrate.

ADmad:
@lorenzo Yeah I realized changing all callables might be too much. Can you suggest where would using only Closures be acceptable? The related ticket has the reasoning for this change.

garas:
All `callable|othertype` should become `\Closure|othertype` if it might be ambiguous.

It should stay if it is only `callable`, where `othertype` is never callable (null, int etc)

lorenzo:
I would say that methods that are only meant for configuring are ok to change (select, where, modelFactory...)

Methods that deal with manipulating dat are not Ok to change (formatResults, traverse, map...)

src/Datasource/QueryTrait.php
```diff
@@ -319,13 +320,13 @@ public function toArray(): array
...
      *
-     * @param callable|null $mapper The mapper callable.
-     * @param callable|null $reducer The reducing function.
+     * @param \Closure|null $mapper The mapper callable.
+     * @param \Closure|null $reducer The reducing function.
      * @param bool $overwrite Set to true to overwrite existing map + reduce functions.
      * @return $this
      * @see \Cake\Collection\Iterator\MapReduce for details on how to use emit data to the map reducer.
      */
-    public function mapReduce(?callable $mapper = null, ?callable $reducer = null, bool $overwrite = false)
```
lorenzo: This one is not ok, for instance.

lorenzo: If we want to reap the benefits of making it faster we could remove the internal type checks for callable and only keep them at boundary methods such as this one. 
The boundary methods can convert to closure to make the internals faster

ADmad:
Thanks for the feeback guys. I update this PR (or open new one) after #13607 is merged to avoid potential conflicts.

ADmad:
@lorenzo Can argument of `ExpressionInterface::traverse()` be changed to Closure?

lorenzo:
> @lorenzo Can argument of `ExpressionInterface::traverse()` be changed to Closure?

Yes

markstory:
This looks good to me, but will need a rebase to be mergable.

ADmad:
You probably meant to comment on #13620 :slightly_smiling_face:. This is the old PR, I had kept it open just for reference.

PR #13615, opened 2019-09-06
Add SQLite decimal length/precision

dereuromark:
I  was surprised that decimal/float didnt quite work the same way in SQLite, despite SQLite documentation saying it supports length/precision.

I set up a fixture with
```php
'amount_required' => ['type' => 'decimal','length' => 10, 'precision' => 6, 'null' => false],		'amount_nullable' => ['type' => 'decimal', 'length' => 10, 'precision' => 6, 'null' => true],
```

When running tests the schema for SQLite DB revealed this in cake using `dd($this->Table->getSchema())`:

    'length' => null,
    'precision' => null,

After this fix it now says the same as mysql/postgres schema:

    'length' => 10,
    'precision' => 6,

I am sure though there is more to be done here.

Refs https://github.com/cakephp/cakephp/pull/1281 and https://github.com/cakephp/cakephp/pull/1691

### DIFF
--- a/src/Core/StaticConfigTrait.php
+++ b/src/Core/StaticConfigTrait.php
@@ -273,7 +273,7 @@ REGEXP;
                 $parsed['className'] = $classMap[$parsed['scheme']];
             }
         }
-        
+
         return $parsed;
     }
 
--- a/src/Core/StaticConfigTrait.php
+++ b/src/Core/StaticConfigTrait.php
@@ -273,7 +273,7 @@ REGEXP;
                 $parsed['className'] = $classMap[$parsed['scheme']];
             }
         }
-
+        
         return $parsed;
     }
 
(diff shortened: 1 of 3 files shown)


dereuromark:
What part needs to be changed so the schema is also applied to the DB?
https://github.com/dereuromark/cakephp-decimal/compare/cake4-core-sqlite-decimal?expand=1 and travis result reveals that behavior for SQLite - not respecting the length/precision - didn't change with this here yet.

markstory:
Any reason to not apply this to master?

dereuromark:
No, I just didnt know the details of why this wasn't done yet in 3.x.

markstory:
> No, I just didnt know the details of why this wasn't done yet in 3.x.

Probably not many folks using decimal + sqlite together.

PR #13617, opened 2019-09-06
4.x psalm

ADmad:
All issues in pslam's baseline file have now been dealt with :muscle:.

Only errors in the baseline are now related to usage of internal or deprecated code which need to be dealt with at some time in future (mainly related to phpunit).


src/Collection/Iterator/NoChildrenIterator.php
```diff
@@ -39,10 +39,10 @@ public function hasChildren(): bool
     /**
      * Returns null as there are no children for this iteration level
      *
-     * @return null
+     * @return \RecursiveIterator
```
garas: Description above is incorrect now

ADmad:
@muglug Any idea why the following error is thrown?

```
ERROR: MissingReturnType - src/basics.php:46:14 - Method debug does not have a return type
    function debug($var, $showHtml = null, $showFrom = true)
```

It seems to randomly appear. I don't get this error locally using PHP `7.2.19-0ubuntu0.18.04.2`.

Issue #13604, opened 2019-09-04
4.x: Plugin path resolution for non-src


dereuromark:
This is a (multiple allowed):

* [x] RFC/enhancement

4.x 

### What you did

App::path() method still resolves non-src paths (no classes, but templates/locales etc).

### What you expected to happen

Mixing src path generation with other non src paths is a bit weird moving forward.
Note the mix of classPath() vs path() vs templatePath() usage inside of this wrapper now.

The cleaner strategy is to not have the App::path() being used anymore for such root folders.
We should use a different method for non class lookup IMO.

E.g.

- `App::templates(?string $plugin = null)`
- `App::locales(?string $plugin = null)`

Or we separate the App class similar to Plugin and its path() vs classPath() method:

- `App::classPath($type, ?$plugin)` for actual classes ("Model/Datasource" etc) inside src/
- `App::path($type, ?$plugin)` for the root folders (and use "templates" instead of "Template"), same for `App::path('Locale', $plugin);` using "locales" then to sync with the actual path element?

What do people think?
A clear and consistent API here would be probably quite helpful.

dereuromark:
Related: https://github.com/cakephp/app/blob/master/config/app.default.php#L66

ADmad:
We could add `App::templatePath()` and `App::resourcePath()`. The latter could have a `$type` or `$subfolder` as argument so `App::resourcePath('locales')` would return you path to locales. I don't think we need a method specific for locales folder path.

ADmad:
In reference to the `app.default.php` you have linked, it's for `3.x`. Those paths updated for 4.x:

https://github.com/cakephp/app/blob/fb5fa72862f3347acf61033ba1915958bb406f3d/config/app.default.php#L66

markstory:
> App::path() method still resolves non-src paths (no classes, but templates/locales etc).

Why is this weird? `App::path()` provides lets people use/remember fewer methods and get what they need. I could see a potential issue if folks try to create `Locale` or `Template` namespaces in their applications, but we could solve that by having the non-class keys be lowercased (`locale` and `templates`).

I don't think we need a plugin argument on any of the `App::*path` methods as we also have `Plugin::path()` and `Plugin::classPath()` to resolve paths for plugins.

Whatever changes we make to path methods on `App` we should also mirror that behavior for `Plugin` so we have consistency between both.

dereuromark:
When working with class paths in 4.x now you need to do weird stuff like

[...]

PR #13618, opened 2019-09-07
Use lowercased type names for paths to non source/class files.

ADmad:
Refs #13604

I agree with @markstory that `App::path()` doesn't need a `$plugin` argument since `Plugin` class already has methods to get plugin paths. Should I drop the argument?

### DIFF
--- a/src/I18n/MessagesFileLoader.php
+++ b/src/I18n/MessagesFileLoader.php
@@ -160,7 +160,7 @@ class MessagesFileLoader
 
         $searchPaths = [];
 
-        $localePaths = App::path('Locale');
+        $localePaths = App::path('locales');
         if (empty($localePaths) && defined('APP')) {
             $localePaths[] = ROOT . 'resources' . DIRECTORY_SEPARATOR . 'locales' . DIRECTORY_SEPARATOR;
         }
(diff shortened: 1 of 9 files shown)


ADmad:
We should maybe add methods `Plugin::resourcePath($plugin, $type)` and `PluginInterface::getResourcePath($type)` to avoid string concatenation like 
`Plugin::path('Foo') . DIRECTORY_SEPARATOR . 'resources' . DIRECTORY_SEPARATOR . 'locales'` in i18n related classes.

src/I18n/MessagesFileLoader.php
```diff
@@ -160,7 +160,7 @@ public function translationsFolders(): array
 
         $searchPaths = [];
 
-        $localePaths = App::path('Locale');
+        $localePaths = App::path('locales');
```
markstory: Is this something we should call out in the migration guide? I don't think rector can change argument values.

markstory:
> Should I drop the argument?

I think we could start emitting a deprecation warning in 4.x when the $plugin parameter is provided. This will let people update more smoothly and doesn't require a breaking change right now.

src/I18n/MessagesFileLoader.php
ADmad: Yes it should be mentioned in migration guide, I'll make a PR.

markstory: Another option would be to continue handling the old names but with deprecation warnings.

ADmad: Removing the special handing of `Template` and `Locale` types would allow people to use them as namespaces if needed, so I would rather not keep the old names.

PR #13620, opened 2019-09-07
4.x closure database

ADmad:
Refs #13611

/cc @lorenzo


src/Database/Query.php
```diff
@@ -315,7 +316,7 @@ public function sql(?ValueBinder $generator = null): string
      * @param callable $visitor A function or callable to be executed for each part
      * @return $this
      */
-    public function traverse(callable $visitor)
+    public function traverse($visitor)
```
ADmad: Had to remove type hint and rely on parameter widening here as `Query implements ExpressionInterface` and `ExpressionInterface::traverse()` only accepts `Closure` now.

ADmad:
@lorenzo Ping

src/Database/Query.php
markstory: Is the concern that userland code is using `traverse()` on a query object?

markstory (commented):
Looks good to me as most of the changes effect inward facing APIs.

src/Database/Query.php
ADmad: Yes, lorenzo commented [here](https://github.com/cakephp/cakephp/pull/13611#issuecomment-527997495) that the accepted type shouldn't be narrowed down for query methods like formatResults, traverse, map... etc.

ADmad:
phpstan and CS job failures are unrelated. I'll fix them directly on 4.x branch.

dereuromark:
Has this been covered in docs and migration guide yet?

ADmad:
This was mainly about cleaning up the internals of db layer. Most of the function changed are quite low level and don't even have any mention in the docs.

It does affect high level API like `Query::where()` but I don't think anyone would be using array callable or invokable classes are argument for it.

PR #13626, opened 2019-09-08
Refactor $fixtures property to a getter method

bancer:
https://github.com/cakephp/cakephp/issues/13583

### DIFF
--- a/src/TestSuite/TestCase.php
+++ b/src/TestSuite/TestCase.php
@@ -835,9 +835,9 @@ abstract class TestCase extends BaseTestCase
     /**
      * Gets fixtures.
      *
-     * @return array
+     * @return string[]
      */
-    public function getFixtures()
+    public function getFixtures(): array
     {
         return $this->fixtures;
     }
--- a/src/TestSuite/TestCase.php
+++ b/src/TestSuite/TestCase.php
@@ -50,7 +50,7 @@ abstract class TestCase extends BaseTestCase
      *
      * @var string[]
      */
-    public $fixtures = [];
+    protected $fixtures = [];
 
     /**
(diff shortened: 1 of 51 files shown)


ADmad (dismissed):

src/TestSuite/TestCase.php
```diff
@@ -831,4 +831,14 @@ public static function setAppNamespace(string $appNamespace = 'TestApp'): void
     {
         Configure::write('App.namespace', $appNamespace);
     }
+
+    /**
+     * Gets fixtures.
+     *
+     * @return array
+     */
+    public function getFixtures()
```
ADmad: `array` return type hint required.

dereuromark:
Shouldnt we make $fixtures protected then?

src/TestSuite/TestCase.php
markstory:
```suggestion
    public function getFixtures(): array
```

src/TestSuite/TestCase.php
```diff
@@ -831,4 +831,14 @@ public static function setAppNamespace(string $appNamespace = 'TestApp'): void
     {
         Configure::write('App.namespace', $appNamespace);
     }
+
+    /**
+     * Gets fixtures.
+     *
+     * @return array
```
dereuromark: Isn't this `string[]` ?

ADmad:
```suggestion
@return string[]
```

dereuromark:
public $fixtures = [];

I still think we should then make it protected in the process, right?

bancer:
@dereuromark Should it be renamed to $_fixtures?

ADmad:
> Should it be renamed to $_fixtures?

No.

PR #13627, opened 2019-09-08
Allow validation constants to be overridden in child classes.

ADmad:


tests/TestCase/ORM/TableTest.php
```diff
@@ -3262,7 +3262,7 @@ public function testValidationWithDefiner()
     public function testValidationWithBadDefiner()
     {
         $this->expectException(\RuntimeException::class);
-        $this->expectExceptionMessage('The Cake\ORM\Table::validationBad() validation method must return an instance of Cake\Validation\Validator.');
+        $this->expectExceptionMessageRegExp('/validationBad\(\) validation method must return an instance of Cake\\\Validation\\\Validator/');
```
garas: why RegExp if there is no expressions

dereuromark (approved):

PR #13628, opened 2019-09-09
4.x psalm

ADmad:


ADmad:
Merging to prevent unnecessary fails on other PRs.

PR #13629, opened 2019-09-09
Fix newlines around psr2

dereuromark:
This was never really done, and causes some some noise in PRs often.

new cake4 standard/code should be already fine here, so this could hopefully also minimize diff and merge in either direction.


dereuromark:
Failing tests are unrelated (master).

PR #13630, opened 2019-09-09
4.x url helper

ADmad:
`UrlHelper::assetUrl()/image()/script()/css()` now only take path as string. This allows having a cleaner API without multiple argument types.

I don't see a reason why these methods should take an array argument. Anyone dynamically generating assets through controllers can just use `UrlHelper::build()`.

### DIFF
--- a/src/View/Helper/HtmlHelper.php
+++ b/src/View/Helper/HtmlHelper.php
@@ -663,14 +663,18 @@ class HtmlHelper extends Helper
      * - `fullBase` If true the src attribute will get a full address for the image file.
      * - `plugin` False value will prevent parsing path as a plugin
      *
-     * @param string $path Path to the image file, relative to the app/webroot/img/ directory.
+     * @param string|array $path Path to the image file, relative to the webroot/img/ directory.
      * @param array $options Array of HTML attributes. See above for special options.
      * @return string completed img tag
      * @link https://book.cakephp.org/3.0/en/views/helpers/html.html#linking-to-images
      */
-    public function image(string $path, array $options = []): string
+    public function image($path, array $options = []): string
     {
-        $path = $this->Url->image($path, $options);
+        if (is_string($path)) {
+            $path = $this->Url->image($path, $options);
+        } else {
+            $path = $this->Url->build($path, $options);
+        }
         $options = array_diff_key($options, ['fullBase' => null, 'pathPrefix' => null]);
 
         if (!isset($options['alt'])) {
--- a/src/View/Helper/HtmlHelper.php
+++ b/src/View/Helper/HtmlHelper.php
@@ -219,7 +219,11 @@ class HtmlHelper extends Helper
         $out = '';
 
         if (isset($options['link'])) {
-            $options['link'] = $this->Url->assetUrl($options['link']);
+            if (is_array($options['link'])) {
+                $options['link'] = $this->Url->build($options['link']);
+            } else {
+                $options['link'] = $this->Url->assetUrl($options['link']);
+            }
             if (isset($options['rel']) && $options['rel'] === 'icon') {
                 $out = $this->formatTemplate('metalink', [
                     'url' => $options['link'],
@@ -659,12 +663,12 @@ class HtmlHelper extends Helper
      * - `fullBase` If true the src attribute will get a full address for the image file.
      * - `plugin` False value will prevent parsing path as a plugin
      *
-     * @param string|array $path Path to the image file, relative to the app/webroot/img/ directory.
+     * @param string $path Path to the image file, relative to the app/webroot/img/ directory.
      * @param array $options Array of HTML attributes. See above for special options.
      * @return string completed img tag
      * @link https://book.cakephp.org/3.0/en/views/helpers/html.html#linking-to-images
      */
-    public function image($path, array $options = []): string
+    public function image(string $path, array $options = []): string
     {
         $path = $this->Url->image($path, $options);
         $options = array_diff_key($options, ['fullBase' => null, 'pathPrefix' => null]);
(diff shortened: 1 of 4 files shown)


dereuromark:
We should probably add deprecations in 3.9 to notify about this inside 3.x and make the upgrade process smoother.

src/View/Helper/HtmlHelper.php
```diff
@@ -219,7 +219,11 @@ public function meta($type, $content = null, array $options = []): ?string
         $out = '';
 
         if (isset($options['link'])) {
-            $options['link'] = $this->Url->assetUrl($options['link']);
+            if (is_array($options['link'])) {
+                $options['link'] = $this->Url->build($options['link']);
```
dereuromark: Shouldnt the array case also run into assetUrl() to append the asset part? Someone just provided the link as non-string form yet.
That hasnt been the case so far either, from what I read. But maybe it should be this way for completeness/consistency?
```php
if (is_array($options['link'])) {
    $url = $this->Url->build($options['link']);
    $options['link'] = $this->Url->assetUrl($url);
```
etc?

ADmad: The only additional thing that `assetUrl()` does after generating a string URL is check and add timestamp to query string. If the URL is an array then it's for a controller in which case timestamp check becomes redundant.

ADmad: Also if you see the diff, earlier `assetUrl()` simply returned the output of `build()` method without doing anything extra.

dereuromark: I know ("That hasnt been the case so far either"). But it makes sense to not have to be done for controller ones I guess. Or you have to do it manually if you serve assets this way :)

src/View/Helper/HtmlHelper.php
```diff
@@ -659,12 +663,12 @@ public function style(array $data, bool $oneLine = true): string
      * - `fullBase` If true the src attribute will get a full address for the image file.
      * - `plugin` False value will prevent parsing path as a plugin
      *
-     * @param string|array $path Path to the image file, relative to the app/webroot/img/ directory.
+     * @param string $path Path to the image file, relative to the app/webroot/img/ directory.
```
garas:
```suggestion
     * @param string $path Path to the image file, relative to the webroot/img/ directory.
```

dereuromark: We often also use ROOT as string:

    ROOT/webroot/img/

but either way works.

garas: Anything non-`app/` is OK.

`app/` is from CakePHP 2.

src/View/Helper/HtmlHelper.php
```diff
@@ -659,12 +663,12 @@ public function style(array $data, bool $oneLine = true): string
...
      * @param array $options Array of HTML attributes. See above for special options.
      * @return string completed img tag
      * @link https://book.cakephp.org/3.0/en/views/helpers/html.html#linking-to-images
      */
-    public function image($path, array $options = []): string
+    public function image(string $path, array $options = []): string
```
markstory: This seems like something that is going to break for people.

src/View/Helper/UrlHelper.php
```diff
@@ -153,15 +153,12 @@ public function script($path, array $options = []): string
...
+     * @param string $path Path string or URL array
      * @param array $options Options array.
      * @return string Generated URL
      */
-    public function assetUrl($path, array $options = []): string
+    public function assetUrl(string $path, array $options = []): string
     {
-        if (is_array($path)) {
-            return $this->build($path, $options);
-        }
```
markstory: Is the amount of breakage that we're going to push onto users worth the small number of typehints? I think this will make template more verbose and complicated when the helpers should be helping make template code simple.

ADmad:
I believe the BC break is only potential and would affect only a handful of users if at all.

I would like to hear from users who actually use array for asset URLs and why?

src/View/Helper/HtmlHelper.php
dereuromark: That is why i made the PR with speaking deprecations in 3.9.

src/View/Helper/UrlHelper.php
dereuromark: With my deprecations and maybe some tooling fixes, yes.

challgren:
I've gone through 20 pages of https://github.com/search?q=%24this-%3EHtml-%3Eimage%28%5B&type=Code so far and don't find any real world examples using an array as the first parameter, Its always a string

ADmad:
@challgren Thanks :slightly_smiling_face:

dereuromark:
As mentioned/linked above the deprecations on 3.9 are vital and important for the smooth migration, where as at the same time also very unlikely to hit a lot of people. And the few people it hits are sure very thankful to get to adjust their code already in 3.x prior to 4.x migration and some harder fails.
Especially on plugin development, these things are important to know beforehand as you often have to support also both majors and therefore would otherwise have to deal with more merge conflicts than necessary. If 3.x code is already as close to 4.x due to such info given, the whole process gets a lot easier for all involved.

ADmad:
@challgren did finally find examples of HtmlHelper::image() usage with array (he  mentioned it on Slack) :).

So we should retain arrays in HtmlHelper methods for assets. Otherwise users generating those assets dynamically would first have to use Router::url() first which isn't user friendly.

So the question remains is whether only updating UrlHelper methods is worth it.

ADmad:
Jeez things are inconsistent in `HtmlHelper`. Currently for `image()` method if `$path` is a URL it's treated as URL array but for other asset methods `script()` and `css()` it's treated as array of multiple URLs which generate multiple script/css tags and using an URL array is not possible.

ADmad:
I have reverted the change to `HtmlHelper::image()`.

dereuromark:
Now it looks like it should look like :+1:

ADmad:
Yeah I think we have a good balance now. Clean API for `UrlHelper` while maintaining convenience for `HtmlHelper`.

PR #13634, opened 2019-09-10
Add class to statically get asset URLs similar to Router::url() for routes.

ADmad:
Refs https://www.reddit.com/r/cakephp/comments/alnoz8/how_to_generate_an_asseturl_with_the_urlhelper_in/

### DIFF
--- a/src/Routing/Asset.php
+++ b/src/Routing/Asset.php
@@ -153,7 +153,7 @@ class Asset
         }
 
         if (isset($plugin)) {
-            $path = Inflector::underscore($plugin) . '/' . $path;
+            $path = static::inflectString($plugin) . '/' . $path;
         }
 
         $optionTimestamp = null;
@@ -263,7 +263,7 @@ class Asset
         $themeName = $options['theme'];
         if ($themeName) {
             $file = trim($file, '/');
-            $theme = static::inflectThemeName($themeName) . '/';
+            $theme = static::inflectString($themeName) . '/';
 
             if (DIRECTORY_SEPARATOR === '\\') {
                 $file = str_replace('/', '\\', $file);
@@ -287,14 +287,14 @@ class Asset
     }
 
     /**
-     * Inflect the theme name to its underscored version.
+     * Inflect the theme/plugin name to its underscored version.
      *
-     * @param string $name Name of the theme which should be inflected.
+     * @param string $string String inflected.
      * @return string Inflected name of the theme
      */
-    protected static function inflectThemeName(string $name): string
+    protected static function inflectString(string $string): string
     {
-        return Inflector::underscore($name);
+        return Inflector::underscore($string);
     }
 
     /**
--- a/src/Routing/Asset.php
+++ b/src/Routing/Asset.php
@@ -157,7 +157,15 @@ class Asset
             [$plugin, $path] = static::pluginSplit($path);
         }
         if (!empty($options['pathPrefix']) && $path[0] !== '/') {
-            $path = $options['pathPrefix'] . $path;
+            $pathPrefix = $options['pathPrefix'];
+            $placeHolderVal = '';
+            if (!empty($options['theme'])) {
+                $placeHolderVal = static::inflectString($options['theme']) . '/';
+            } elseif (isset($plugin)) {
+                $placeHolderVal = static::inflectString($plugin) . '/';
+            }
+
+            $path = str_replace('{plugin}', $placeHolderVal, $pathPrefix) . $path;
         }
         if (!empty($options['ext']) &&
             strpos($path, '?') === false &&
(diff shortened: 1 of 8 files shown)


ADmad:
Once this is approved I can retrofit `UrlHelper` to use this class in it's methods and avoid code duplication.

One thing to note is this class only does URL encoding and doesn't do any HTML entities encoding. I believe the latter should be done only when this class is used through helpers where are meant to generate HTML.

src/View/Asset.php
```diff
@@ -0,0 +1,336 @@
...
+use Cake\Core\Configure;
+use Cake\Core\Exception\Exception;
+use Cake\Core\Plugin;
+use Cake\Routing\Router;
+use Cake\Utility\Inflector;
+
+/**
+ * Class for generating asset URLs.
+ */
+class Asset
```
markstory: Should this be in Routing package?

src/View/Asset.php
```diff
@@ -0,0 +1,336 @@
...
+
+    /**
+     * Inflect the theme name to its underscored version.
+     *
+     * @param string $name Name of the theme which should be inflected.
+     * @return string Inflected name of the theme
+     */
+    protected static function inflectThemeName(string $name): string
+    {
+        return Inflector::underscore($name);
```
markstory: Do we need this additional function call/indirection?

markstory (commented):
If we add this class would we replace the guts of UrlHelper with this?

src/View/Asset.php
ADmad: Probably, I wasn't too sure of the location.

src/View/Asset.php
ADmad: I copied this from UrlHelper, I can remove it.

ADmad:
> If we add this class would we replace the guts of UrlHelper with this?

Yes

src/View/Asset.php
markstory: Routing seems like a better location for this if we want to make a general asset URL building class.

ADmad:
- Moved `Asset` class under `Routing`
- Updated `UrlHelper` to use `Asset` class
- Added the ability to control the inflection type when inflecting theme/plugin name. I have kept default inflection to `underscore` for now but would like to change it to `dasherize` to consistent with Router.

src/Routing/Asset.php
```diff
@@ -0,0 +1,356 @@
...
+    protected static $inflectionType = 'underscore';
+
+    /**
+     * Set inflection type.
+     *
+     * @param string $inflectionType Inflection type
+     * @return void
+     * @psalm-param 'dasherize'|'underscore' $inflectionType
+     */
+    public static function setInflectionType(string $inflectionType)
```
dereuromark: you can set :void here as return type, right?

src/Routing/Asset.php
```diff
@@ -0,0 +1,356 @@
...
+        return $request->getAttribute('webroot');
+    }
+
+    /**
+     * Splits a dot syntax plugin name into its plugin and filename.
+     * If $name does not have a dot, then index 0 will be null.
+     * It checks if the plugin is loaded, else filename will stay unchanged for filenames containing dot.
+     *
+     * @param string $name The name you want to plugin split.
+     * @return array Array with 2 indexes. 0 => plugin name, 1 => filename.
```
dereuromark: Isnt this string[] then?

src/Routing/Asset.php
ADmad: yup, it's an oversight

src/Routing/Asset.php
ADmad: No, since plugin can be null

dereuromark:
Nice SOLID clean up :)

src/Routing/Asset.php
```diff
@@ -26,17 +26,17 @@
...
      * @psalm-var 'dasherize'|'underscore'
      */
     protected static $inflectionType = 'underscore';
 
     /**
-     * Set inflection type.
+     * Set inflection type to use when inflecting plugin/theme name.
      *
-     * @param string $inflectionType Inflection type
+     * @param string $inflectionType Inflection type. Valid values: 'dasherize', 'underscore'.
```
dereuromark: I recommend making those class const then for clarity and to be easier used by dev.

ADmad: Theoretically one can use any `Inflector` method. Someone can use `'humanize'` if they prefer. I don't want to add constants for all `Inflector` methods. `RouterBuilder` methods also have `inflect` option for similar purpose.

I'll change the description to mention the value should be a valid `Inflector` method name.

dereuromark: Cool, ok.

markstory:
Looks good :+1: Do you think this is a class we should document in the book? I can write something up in the routing chapter.

ADmad:
@markstory Yes please, thanks.

PR #13635, opened 2019-09-10
Fix undefined property $_tableLocator

bancer:
The problem with `$this->_tableLocator === null` and `!$this->_tableLocator` is that it results in: 
```
Undefined property: MyTest::$_tableLocator

.../Vendor/cakephp/orm/Locator/LocatorAwareTrait.php:73
.../src/TestSuite/MyTestCase.php:59
```

### DIFF
--- a/src/ORM/Locator/LocatorAwareTrait.php
+++ b/src/ORM/Locator/LocatorAwareTrait.php
@@ -70,7 +70,7 @@ trait LocatorAwareTrait
      */
     public function getTableLocator()
     {
-        if ($this->_tableLocator === null) {
+        if (!isset($this->_tableLocator)) {
             $this->_tableLocator = TableRegistry::getTableLocator();
         }
 


src/ORM/Locator/LocatorAwareTrait.php
```diff
@@ -70,7 +70,7 @@ public function setTableLocator(LocatorInterface $tableLocator)
      */
     public function getTableLocator()
     {
-        if ($this->_tableLocator === null) {
```
ADmad: How can it throw undefined property error when the trait itself has `$_tableLocator` declared?
https://github.com/cakephp/cakephp/blob/4c543ecd4a8e2d6e2db7fdf65ef3bebef3722175/src/ORM/Locator/LocatorAwareTrait.php#L30

bancer: Declared does not mean that it is defined.

bancer: http://sandbox.onlinephpfunctions.com/code/dfc84393b4ecf03d9d3bf2dd1a3cd0a3d5cca838

ADmad: Are you trying to troll me with that pastebin?  You are using local var `$_b` instead of property `$this->_b`.

ADmad: This would be a valid example which doesn't throw any error http://sandbox.onlinephpfunctions.com/code/dfc84393b4ecf03d9d3bf2dd1a3cd0a3d5cca838

```
<?php
class A {
    
    protected $_b;
    
    public function getB() {
        if ($this->_b === null) {
            $this->_b = true;
        }
        return $this->_b;
    }
    
}

$a = new A();
$a->getB();
```

bancer: My bad in the sandbox. I wonder why do I get undefined in LocatorAwareTrait.php:73.

ADmad: No problem :slightly_smiling_face:  Do leave a comment here when you figure out what's causing the issue for you. We have similar property === null checks at multiple places throughout the core.

bancer: It seems that `unset($this->_tableLocator);` was called at some point in our code.

Issue #13639, opened 2019-09-10
cURL Get requests come across as POST


KainGNX:
https://github.com/cakephp/cakephp/blob/a2baa58e0150da4109313954975b6a3681cf8e86/src/Http/Client/Adapter/Curl.php#L112-L114

I have been trying to use a get request to call to another Cake API I have running in the same environment, but when using the `Cake\Http\Client::get()` method with something like `$client->get($fullUrl);` the request is received as POST no matter what causing a route mismatch in my other Cake app.

I even tried a wrapped PSR-7 request object in conjunction with `Cake\Http\Client::sendRequest()` and got the same result.

I then commented out setting `CURLOPT_POSTFIELDS` key in a brief test and it works as GET.

I was going to put in a pull request to put back the conditional around setting `CURLOPT_POSTFIELDS` that used to be there.

But wanted to ask first, was there a reason it was removed? With this change, am I missing something else to make sure the request is GET?

EDIT: If I am not mistaken, I think I get why the conditional was removed, it would likely wind up evaluating 'true` and always execute within anyway. 

Thanks in advance!

markstory:
Thank you for the issue. Going to close this as we can continue the discussion in your pull request.

PR #13640, opened 2019-09-10
`CURLOPT_POSTFIELDS` only set when the request body content is not empty

jason-horvath:
I did open an issue on this with an explainer but figured it couldn't hurt to put in a PR. Sorry if it seems spammy.  :)
https://github.com/cakephp/cakephp/issues/13639

### DIFF
--- a/src/Http/Client/Adapter/Curl.php
+++ b/src/Http/Client/Adapter/Curl.php
@@ -108,14 +108,14 @@ class Curl implements AdapterInterface
                 $out[CURLOPT_CUSTOMREQUEST] = $request->getMethod();
                 break;
         }
-        
+
         $body = $request->getBody();
         $body->rewind();
         $out[CURLOPT_POSTFIELDS] = $body->getContents();
         if ($out[CURLOPT_POSTFIELDS] === '') {
             unset($out[CURLOPT_POSTFIELDS]);
         }
-        
+
         if (empty($options['ssl_cafile'])) {
             $options['ssl_cafile'] = CaBundle::getBundledCaBundlePath();
         }
--- a/src/Http/Client/Adapter/Curl.php
+++ b/src/Http/Client/Adapter/Curl.php
@@ -110,9 +110,10 @@ class Curl implements AdapterInterface
         }
 
         $body = $request->getBody();
-        $body->rewind();
-        $out[CURLOPT_POSTFIELDS] = $body->getContents();
-
+        if (!empty($body->getContents())) {
+            $body->rewind();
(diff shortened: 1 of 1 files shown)


src/Http/Client/Adapter/Curl.php
```diff
@@ -110,9 +110,10 @@ public function buildOptions(RequestInterface $request, array $options): array
         }
 
         $body = $request->getBody();
-        $body->rewind();
-        $out[CURLOPT_POSTFIELDS] = $body->getContents();
-
+        if (!empty($body->getContents())) {
+            $body->rewind();
+            $out[CURLOPT_POSTFIELDS] = $body->getContents();
```
garas: Retrieve (after rewind) contents once, then check `!== ''`.

Your `if` now checks non-rewinded contents and might be at end of stream. Also `empty` is useful on variables, not function calls.

src/Http/Client/Adapter/Curl.php
```diff
@@ -110,9 +110,10 @@ public function buildOptions(RequestInterface $request, array $options): array
         }
 
         $body = $request->getBody();
-        $body->rewind();
-        $out[CURLOPT_POSTFIELDS] = $body->getContents();
```
markstory: Could you unset the field key if the value has no length?

src/Http/Client/Adapter/Curl.php
KainGNX: I could make these changes, I see the other proposed change by @markstory , I take it these are two different solutions.

src/Http/Client/Adapter/Curl.php
KainGNX: Pushed up a commit that just unsets `CURLOPT_POSTFIELDS` based on this comment.

src/Http/Client/Adapter/Curl.php
```diff
@@ -108,14 +108,14 @@ public function buildOptions(RequestInterface $request, array $options): array
                 $out[CURLOPT_CUSTOMREQUEST] = $request->getMethod();
                 break;
         }
-
```
KainGNX: Thanks @stickler-ci that was fast. I was wondering where these went, I was gonna do the same! :D

markstory:
Thank you 👍

dereuromark:
Do we need to baclport to 3.x?

markstory:
@dereuromark No 3.x didn't have the changes that caused this issue applied as they were part of the psalm/phpstan changes.

PR #13644, opened 2019-09-12
Add GH action for CI

ADmad:
- GH actions seem to provide lot more parallelism and speed than other CI services.
- The VM only has PHP 7.1 - 7.3 so if we want to run tests for 5.6 / 7.4 we would have to get PHP using docker. Actions have first class docker support so someone proficient with docker can update the config https://help.github.com/en/articles/workflow-syntax-for-github-actions.
[ppa:ondrej/php](https://launchpad.net/~ondrej/+archive/ubuntu/php) does seem to be used so potentially PHP 7.4 should be readily available once released.
- Not sure why that `ConsoleInputTest` fails.
- It does provide [windows VM](https://help.github.com/en/articles/software-in-virtual-environments-for-github-actions#windows-server-2019) too but it doesn't seem to have SqlServer pre-installed, so if someone can figure out how to do that we could get off AppVeyor too.


ADmad:
Since CI support through GH actions is pretty new Codecov doesn't recognize the service. So the codecov's repo token available at https://codecov.io/gh/cakephp/cakephp/settings will have to be manually added here https://github.com/cakephp/cakephp/settings/secrets under name `CODECOV_TOKEN`.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,129 @@
...
+      run: |
+        if [[ ${{ matrix.php-version }} == '7.2' ]]; then
+          # Setting CODECOVERAGE is necessary to skip memory usage tests
+          CODECOVERAGE=1 vendor/bin/phpunit --verbose --coverage-clover=clover.xml
+        else
+          vendor/bin/phpunit
+        fi
+
+    - name: Submit code coverage data
+      run: |
```
garas:
```suggestion
      if: matrix.php-version == '7.2'
      run: |
```
I think this should remove skipped `Submit code coverage data` step from non 7.2 jobs.

A bit more readable than bash `if`.

ADmad: It still does show the `Submit code coverage data` step for PHP 7.3 jobs.

garas: At least it has white square icon to tell it skipped

ADmad:
This should be good to merge.

- Jobs with tests on PHP 7.4 with code coverage remain on travis. All other checks are now run using Github actions
- I couldn't get apcu php extesnion enabled for CLI on the Github VM, even though I have followed the generally required steps for enabling it on Ubuntu. It's not a biggie as apcu tests are run on Travis. Someone could help figure out getting it running on Github's VM too.
- Some jobs now run phpunit in verbose mode which helps see details about skipped tests.

markstory:
The CI build looks good, simple enough to understand. Would you want to use this instead of travis for all builds? Is there a need to have travis for master and github actions for pull requests?

ADmad:
The GitHub VM only has PHP 7.1 - 7.3. So we need to continue using Travis for PHP 7.4 and for master older versions of PHP.

We could have potentially all PHP versions on Github's VM itself using docker but I am not too experienced with docker and don't want to spend time tinkering with it.

markstory:
:+1: We could run with both build suites for a few months and make sure Github's is doing what we need and then pare down the travis build to not overlap with Github's

PR #13647, opened 2019-09-14
New Feature: Allow overriding of button title

aleph1:
This is a non-breaking feature that allows the button title to be passed via the options array. This can be useful when using FormHelper::inputs as it allows adding buttons into the same fieldset as other inputs. For example:

$fields = [];
$fields[] = [ 'name' => '', 'type' => 'button', 'label' => false, 'title' => 'Select All' ];
$fields[] = [ 'name' => 'data[Products][]', 'type' => 'checkbox', 'label' => 'Product 1', 'value' => 1 ];
$fields[] = [ 'name' => 'data[Products][]', 'type' => 'checkbox', 'label' => 'Product 2', 'value' => 2 ];
etc.

echo $this->Form->inputs( $fields, null, [
	'fieldset' => 'products',
	'legend' => 'Product Access'
] );

### DIFF
--- a/lib/Cake/View/Helper/FormHelper.php
+++ b/lib/Cake/View/Helper/FormHelper.php
@@ -1823,6 +1823,9 @@ class FormHelper extends AppHelper {
  */
 	public function button($title, $options = array()) {
 		$options += array('type' => 'submit', 'escape' => false, 'secure' => false);
+		if (isset($options['title'])) {
+			$title = $options['title'];
+		}
 		if ($options['escape']) {
 			$title = h($title);
 		}


aleph1:
The only issue with this is that the HTML global attribute title (https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/title) is a valid attribute on input elements and this effectively forces the button text and the title attribute to have the same text. Perhaps it would make more sense to use the array key text, for example:

$fields[] = [ 'name' => '', 'type' => 'button', 'label' => false, 'text' => 'Select All', 'title' => 'Select all products'];

This would have the button appear with the text "Select All" and the title attribute on the input itself would be "Select all products" which would ensure the browser can still display a tooltip, and that the longer text can be read by a screen reader.

saeideng:
infact button is not input , and cake crated it as last possible case 
https://github.com/cakephp/cakephp/blob/46068ac65b935919b7c13dee6cb0532c53081929/lib/Cake/View/Helper/FormHelper.php#L1162

you can add `title` as field name in array if you want
```php
$fields['SelectAll'] = [ 'type' => 'button', 'label' => false ];
```

saeideng:
also CakePHP 2.x is in maintenance mode
and can not be accept new features

We have supported 2.x since October 2011 and have [clearly communicated](https://bakery.cakephp.org/2017/07/22/cakephp_2100_released.html) that it will receive only bug and security fixes from June 2017 going forward. We both cannot accept any pull requests or issues which:

- backport 3.x features to 2.x
- add/request new functionality for 2.x

aleph1:
Sorry. I didn't realize feature requests were closed.

markstory:
@aleph1 We're happy to accept new features in the 3.next or 4.x branches :smile:

PR #13648, opened 2019-09-15
4.x mysql charset

ADmad:
Refs #13057

Given the concerns raised by @chinpei215 it's safer to make this change in 4.x. I have tweaked the patch to allow using `SET NAMES ...` to set the charset if specifying charset in connection DSN is not feasible for someone. We can document this in skeleton app's config file.


markstory:
We should also document the configuration requirements as the skip-client-encoding configuration could be off.

ADmad:
I have opened PR for both the manual and skeleton app.

PR #13649, opened 2019-09-15
2.x: Fix blackHole parameter

oppara:
2.x book says the blackhole callback should expect a parameter indicating the type of error bellow.
"auth", "csrf", "get", "post", "put", "delete" and "secure".  
https://book.cakephp.org/2.0/en/core-libraries/components/security-component.html#SecurityComponent::$blackHoleCallback

However, in the current 2.x version, only “auth” or “secure” is set.

This PR fixes the above blackhole callback's parameter bug.

### DIFF
--- a/lib/Cake/Controller/Component/SecurityComponent.php
+++ b/lib/Cake/Controller/Component/SecurityComponent.php
@@ -397,9 +397,11 @@ class SecurityComponent extends Component {
 				$require = $this->$property;
 				if (in_array($this->_action, $require) || $this->$property === array('*')) {
 					if (!$controller->request->is($method)) {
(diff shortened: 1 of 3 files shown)


ADmad:
If the documentation doesn't match actual behavior then it's the documentation that should be changed. 2.x is now is bugfix mode only and we don't accept any enhancements.

markstory:
The documentation has been updated to match the code in cakephp/docs@247b1e2b7578413d0d2cd4a10702a72d629373a9. Thank you for taking the time to put this patch together but at this stage in 2.x's life updating the documentation is the right way to correct this error.

Issue #13654, opened 2019-09-17
Memcached - wrong expiration time when duration is set as timestamp


bancer:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.x.
* Platform and Target: any.

### What you did
```
$config['duration'] = time() - 1;
Cache::setConfig($name, $config);
```

### What happened
The cache does not expire. Looks like expiration time is in the future instead of in the past.

It seems that [this line](https://github.com/cakephp/cakephp/blob/master/src/Cache/Engine/MemcachedEngine.php#L305) is the cause of the problem.

### What you expected to happen
> Some storage commands involve sending an expiration value (relative to an item or to an operation requested by the client) to the server. In all such cases, the actual value sent may either be Unix time (number of seconds since January 1, 1970, as an integer), or a number of seconds starting from current time. 

https://www.php.net/manual/en/memcached.expiration.php

According to that I would expect that expiration time would be 1 second in the past.

ADmad:
Jeez, that's only wacky interpretation of expiration value. BTW why are you setting expiration value in the past instead of using `delete()` method?

bancer:
The problem I try to solve is to disable all cache configurations except `_cake_core_` and `_cake_model_` in unit tests. 

Another use case is when it is necessary to set the expiration time greater than 30 days then a Unix timestamp can be used, f.ex. `$config['duration'] = strtottime('+40 days')`.

garas:
You could also use `NullEngine` to disable some configurations.

PR #13655, opened 2019-09-17
Remove unnecessary 30 days memcached duration check.

bancer:
https://github.com/cakephp/cakephp/issues/13654


src/Cache/Engine/MemcachedEngine.php
```diff
@@ -301,10 +301,6 @@ public function getOption($name)
     public function write($key, $value)
     {
         $duration = $this->_config['duration'];
-        if ($duration > 30 * DAY) {
```
ADmad: Same change should be done in `add()` method too for consistency.

PR #13656, opened 2019-09-17
Remove unneeded ServerRequest properties.

ADmad:
They are not needed often enough and their values can be deduced
from other attributes.

Keeping them just adds cognitive load IMO.


ADmad:
I would like to remove `here` as an emulated attribute too, I doubt people use it often enough in their apps.

markstory:
I would be wary of `here`. In the past when its value has been changed/broken we get several issues opened. While we don't use it internally I suspect many applications and plugins do.

PR #13662, opened 2019-09-20
Cleanup

ADmad:
- Remove unused property from Inflector
- Allow plugin auto loading where it makes sense. Plugin class now tries to autoload plugin if not already loaded and throws exception if it's unable to.


src/Utility/Inflector.php
```diff
@@ -162,243 +162,6 @@ class Inflector
         'pokemon', 'proceedings', 'research', 'sea[- ]bass', 'series', 'species', 'weather',
     ];
 
-    /**
-     * Default map of accented and special characters to ASCII characters
-     *
-     * @var array
-     */
-    protected static $_transliteration = [
```
garas: There is still `Inflector::rules('transliteration',...)` phpdoc

PR #13669, opened 2019-09-22
4.x request params cleanup

ADmad:
It was just for legacy reasons that various extra data like security token, paging params were stuffed into request params. Request params should be limited to routing relate keys only. The extra data is now set using separate request attributes.

### DIFF
--- a/src/Controller/Component/RequestHandlerComponent.php
+++ b/src/Controller/Component/RequestHandlerComponent.php
@@ -177,7 +177,7 @@ class RequestHandlerComponent extends Component
         }
 
         $isAjax = $request->is('ajax');
-        $controller->setRequest($request->withParam('isAjax', $isAjax));
+        $controller->setRequest($request->withAttribute('isAjax', $isAjax));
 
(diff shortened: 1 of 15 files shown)


src/Controller/Component/RequestHandlerComponent.php
```diff
@@ -177,7 +177,7 @@ public function startup(EventInterface $event): void
         }
 
         $isAjax = $request->is('ajax');
-        $controller->setRequest($request->withParam('isAjax', $isAjax));
+        $controller->setRequest($request->withAttribute('isAjax', $isAjax));
```
markstory: These parameters could be in use by applications. We can likely migrate these with rector. I don't see an existing rector that can do this but I can try adding one.

ADmad: I hope it can be handled with rector. Would be real sad to be not be able to clean up cruft even in a major release.

ADmad:
Conflict resolved.

PR #13670, opened 2019-09-23
4.x router cleanup

ADmad:
- Fix extra `_params` key being set in `$context` passed to `Route::match()` call on line 484.
-  Optimize `Router::fullBaseUrl()` to return early when called as getter with `Router::$_fullBaseUrl` already set.

### DIFF
--- a/src/Routing/Router.php
+++ b/src/Routing/Router.php
@@ -532,7 +532,7 @@ class Router
     public static function routeExists($url = null, bool $full = false): bool
     {
         try {
-            $route = static::url($url, $full);
+            static::url($url, $full);
 
             return true;
         } catch (MissingRouteException $e) {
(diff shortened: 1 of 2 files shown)


ADmad:
There's currently one inconsistent behaviour regarding full URL generation which should be addressed.

Normally when generating full URL `Router::url()` takes the host from `$_requestContext` but when `_ssl` key is set it get's the URL by parsing the URL returned by `Router::fullBaseUrl()`. Normally this is not an issue as the host is usually same for both but it would create discrepancy if `App.fullBaseUrl` is set with a host which is not the same as current host got from current request URI.

markstory:
> Also calling parse_url() everytime _ssl key is set seems quite wasteful. Shouldn't we just update Router::$_requestContext['_host'] when Router::fullBaseUrl() is called? I believe the host available from App.fullBaseUrl should always take precedence.

Do you mean parsing the URL the first time the full base url is read from configure? We could store that in the `_requestContext` and use that instead.

ADmad:
> Do you mean parsing the URL the first time the full base url is read from configure? We could store that in the _requestContext and use that instead.

Yup, that's what I meant.

PR #13672, opened 2019-09-23
Avoid potentially parsing URL multiple times.

ADmad:
In order for this optimization to work `Router::fullBaseUrl()` needs to be called initially during app bootup so that the request context in populated.

This would require replacing the `Configure::write('App.fullBaseUrl', ...)` in app's `config/boostrap.php` with call to `Router::fullBaseUrl()` instead. So this is something those upgrading from 3.x would need to care of.

### DIFF
--- a/tests/TestCase/Routing/RouterTest.php
+++ b/tests/TestCase/Routing/RouterTest.php
@@ -1884,6 +1884,7 @@ class RouterTest extends TestCase
      */
     public function testGenerationWithSslOption()
     {
+        Router::fullBaseUrl('http://app.test');
         Router::connect('/:controller/:action/*');
         $request = new ServerRequest([
             'url' => '/images/index',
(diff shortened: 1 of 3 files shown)


ADmad:
I added a commit which makes explicitly setting `App.fullBaseUrl` optional. `Router::fullBaseUrl()` can generate the full URL from request context instead and set `App.fullBaseUrl`.

dereuromark:
Ins't the main reason we need `App.fullBaseUrl` the CLI run for commands? To generate valid URLs there?
As such this config remains super important IMO.
We should probably do it the other way around and set the `Router::fullBaseUrl()` from the config?
Or how can we sync CLI vs web requests here for this?

ADmad:
> Ins't the main reason we need App.fullBaseUrl the CLI run for commands?
Yes

> We should probably do it the other way around and set the Router::fullBaseUrl() from the config?
Yes

tests/TestCase/Routing/RouterTest.php
```diff
@@ -1897,52 +1916,12 @@ public function testGenerationWithSslOption()
...
-        ]);
-        $this->assertSame('http://localhost/images/index', $result);
-    }
-
-    /**
-     * Test that the _ssl + _full options work together.
-     *
-     * @return void
-     */
-    public function testGenerationWithSslAndFullOption()
```
ADmad: It might be not clear in the diff but I dropped this method `testGenerationWithSslAndFullOption()`.  Using `_full => false` in combination with `_ssl` made no sense. If you set `_ssl` you will always get a full URL and value of `_full` should be redundant.

PR #13677, opened 2019-09-25
Fix header comparison debug output

dereuromark:
Currently, it is super annoying to work with the header comparison fails in Integration tests.
```
1) App\Test\TestCase\Controller\AccountControllerTest::testActivate
Failed asserting that 'http://app.local/login/foobar' equals content in 
header 'Location'.
```

So now you need to manually inject things like 
```php
	dd($this->_response->getHeader('Location'));
```
everywhere to find out whats going on:
```
[
        (int) 0 => 'http://app.local/login?with=foobar'
]
```

With the change, the output is speaking by design and helps to verify the result of the test:
```
1) App\Test\TestCase\Controller\AccountControllerTest::testLogin
Failed asserting that 'http://app.local/login/foobar' equals content in 
header 'Location': `http://app.local/login?with=foobar`.
```

Other comparisons could also be adjusted as follow up.


markstory:
Great work :clap:

PR #13678, opened 2019-09-25
4.x: Remove remaining pieces after requestAction removal

garas:
Remove remaining 'autoRender',  'bare', 'return', 'requested' used by `requestAction()`.

### DIFF
--- a/tests/TestCase/Routing/RouterTest.php
+++ b/tests/TestCase/Routing/RouterTest.php
@@ -2491,30 +2491,6 @@ class RouterTest extends TestCase
         Router::connect('/:controller', [], ['routeClass' => 'Object']);
     }
 
-    /**
-     * test reversing parameter arrays back into strings.
-     *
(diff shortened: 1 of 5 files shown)


tests/TestCase/Routing/RouterTest.php
```diff
@@ -2491,30 +2491,6 @@ public function testCustomRouteException()
...
     }
 
-    /**
-     * test reversing parameter arrays back into strings.
-     *
-     * Mark the router as initialized so it doesn't auto-load routes
-     *
-     * @return void
-     */
-    public function testReverseToken()
```
ADmad: Shouldn't need method retained and just the extra key in $params array removed? Or there's already another test for the same?

garas: #13669 removed `'_Token' => ['key' => 'sekret'],` and based on test name, it was object of test, so I removed completely.

PR #13679, opened 2019-09-25
Fix up Constraints.

dereuromark:
Follow https://github.com/cakephp/cakephp/pull/13677

Turns out we don't need the more stateful approach. I reverted that, especially since classes inherit this class that do not use this new property.

Also fixed some missing code fencing where recommended, e.g. around regex pattern, paths or URLs.
Copy and paste for those have a better visual experience. Also easier to detect leading or trailing whitespace, accidental or otherwise. And URLs into GitHub or SO or other sites won't auto-create followable links from those which is usually always undesired with those.

In general, we should not use `"` for code fencing, this is for human quoting. Also `'` character is not good for this. Instead, we should try to use proper ` for code fencing more in exception messages and alike.

Also:
- Added missing test for assertHeaderNotContains()

Once this is merged also into 4.x code base I would like to follow up on the code fencing issues there around paths and URLs etc. I wanted to keep these changes rather small (internal) for 3.x series.


markstory (approved):

src/TestSuite/Constraint/Response/ContentType.php
```diff
@@ -43,6 +43,6 @@ public function matches($other)
      */
     public function toString()
     {
-        return 'was set as the Content-Type';
+        return 'was set as the Content-Type (`' . $this->response->getType() . '`)';
```
garas:
```suggestion
        return 'is set as the Content-Type (`' . $this->response->getType() . '`)';
```

Other constraints uses present time.

dereuromark: I guess I now need to fix some tests :)

dereuromark: I updated some other past tense too now for consistency.

PR #13684, opened 2019-09-26
Fix generation of datetime control.

ADmad:


ADmad:
phpcs failure is unrelated. psalm failure is due to bug in pslam which will be fixed in it's next release.

Issue #11249, opened 2017-09-28
Radio : customize css selected class


Erwane:
This is a (multiple allowed):

* [ ] bug
* [x] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.5.x and 3.x generally

### What you did
Displaying a form with 2 radio elements and 1 is selected (default value)

### What happened
html code is something like
```html
<label class="label selected" for="guild-edit-0">
    <input type="radio" name="guild_edit" value="0" id="guild-edit-0" checked="checked">
    <i class="fa fa-minus"></i>
</label>
```

The "selected" class was added by RadioWidget :
```php
if (!is_bool($data['label']) && isset($radio['checked']) && $radio['checked']) {
    $data['label'] = $this->_templates->addClass($data['label'], 'selected');
}
```

### What you expected to happen
I really need to customize that class by "active" instead of "selected" for compatibility with buttons bootstrap 3 javascript : https://getbootstrap.com/docs/3.3/javascript/#buttons-checkbox-radio

Bootstrap3 fix the class too BUT it's more easy for me to PR cake code :)

inoas:
That won't be BC. Making it configurable with defaults to 'selected' would be ok, I guess.

markstory:
The selected class could also be a 'template' value. That would make it use the same extension points that other form helper customizations use.

andrej-griniuk:
@markstory could you clarify about "template value"? Or give any example where it's done this way.

markstory:
Sure. FormHelper uses 'template' strings as the main way that people can customize the output of FormHelper. You can see the [list of templates](https://api.cakephp.org/3.5/class-Cake.View.Helper.FormHelper.html#$_defaultConfig). I could see the `selectedClass` being a template option. This would allow developers to make their 'theme' changes self contained.

dereuromark:
@Erwane Do you want to make a PR here with the suggested changes?

Erwane:
@dereuromark with https://github.com/Erwane/cakephp/commit/3e6ef467efc3383e75fe97e1832220c07b792913 yes, but not with the "template" idea :(

Erwane:
Sorry bot. I will do it.

dereuromark:
PR open.

PR #13686, opened 2019-09-27
#11249: custom selected class

Erwane:
allow custom selected class for radio and checkbox label.

```php
$this->loadHelper('Form', [
    'templates' => [
        'selectedClass' => 'active',
    ],
]);
```

related to #11249

### DIFF
--- a/src/View/Widget/MultiCheckboxWidget.php
+++ b/src/View/Widget/MultiCheckboxWidget.php
@@ -210,7 +210,8 @@ class MultiCheckboxWidget implements WidgetInterface
             ];
 
             if ($checkbox['checked']) {
-                $labelAttrs = $this->_templates->addClass($labelAttrs, 'selected');
+                $selectedClass = $this->_templates->format('selectedClass', []);
+                $labelAttrs = $this->_templates->addClass($labelAttrs, $selectedClass);
             }
 
             $label = $this->_label->render($labelAttrs, $context);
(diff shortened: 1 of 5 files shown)


Erwane:
missing templates in tests. I will fix that.

src/View/Widget/MultiCheckboxWidget.php
```diff
@@ -210,7 +210,8 @@ protected function _renderInput($checkbox, $context)
             ];
 
             if ($checkbox['checked']) {
-                $labelAttrs = $this->_templates->addClass($labelAttrs, 'selected');
+                $selectedClass = $this->_templates->format('selectedClass', []);
```
markstory: What happens when a user is using a custom set of templates that doesn't define `selectedClass`?

Erwane: I'm not sure it is relevant.
If the selectedClass is not included in custom set of templates, the stringTemplate will thrown an exception, like all missing required templates.

The selectedClass is set to `selected` by default in the `FormHelper`

markstory: Wouldn't that template missing error cause applications to break when they update?

ADmad: @markstory He has added the new template in FormHelper, that's where all templates used by widgets are set.

Issue #13676, opened 2019-09-25
Number::toPercentage() does not respect I18n rules regarding placement of the percentage sign


DBnR1:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.8.4

### What you did
Using `Number::toPercentage()` I expect the returned string to be localized based on the locale specified in `$options`, specifically regarding the placement of the percentage sign. Currently only the number itself is localized.

Code example:
```
$n = new \NumberFormatter('fi_FI', \NumberFormatter::PERCENT);
// format using Intl NumberFormatter
$intlFormatted = $n->format(0.13);
// format using Cake formatter
$cakeFormatted = \Cake\I18n\Number::toPercentage(13, 0, ['locale' => 'fi_FI']);

printf('"%s" vs "%s"', $intlFormatted, $cakeFormatted);
```

### What happened
`"13 %" vs "13%"`

### What you expected to happen
`"13 %" vs "13 %"`

In the Finnish language (possibly others, too) there is supposed to be a space between the number and the percentage sign.

PR #13691, opened 2019-09-29
Fix Number::toPercentage() format

littleylv:
Refs #13676 .

Note: I change 

```
        if ($options['multiply']) {
            $value *= 100;
        }
```

to

```
        if (!$options['multiply']) {
            $value /= 100;
        }
```
 because `NumberFormatter::PERCENT` use value `0.13` instead of `13` for `13%`.

For CakePHP, Number::toPercentage(13) => `13%` because it uses `NumberFormatter::DECIMAL` before.

### DIFF
--- a/src/I18n/Number.php
+++ b/src/I18n/Number.php
@@ -112,12 +112,12 @@ class Number
      */
     public static function toPercentage($value, $precision = 2, array $options = [])
     {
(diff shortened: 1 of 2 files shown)


markstory:
Thank you 👍

PR #13692, opened 2019-09-29
Fix CS errors

ADmad:
I'll move the updates to `phpcs.xml.dist` to ruleset of `next` branch of our codesniffer.

### DIFF
--- a/config/bootstrap.php
+++ b/config/bootstrap.php
@@ -12,6 +12,7 @@
  * @since         0.2.9
  * @license       https://opensource.org/licenses/mit-license.php MIT License
  */
+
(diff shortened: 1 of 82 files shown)


src/Auth/AbstractPasswordHasher.php
```diff
@@ -1,4 +1,5 @@
 <?php
+
```
dereuromark: Thats more than silly :/ declare() should be on the same line, by common sense.

ADmad: Common sense and PSR standards don't always go hand in hand.

dereuromark (approved):

PR #13693, opened 2019-09-29
Allow using UploadedFileInterface instance as email attachment.

ADmad:


ADmad:
@markstory Can we make stickler ignore PRs to 4.x?  It's a bit annoying as it uses the old standard and reports incorrect errors.

ADmad:
Ignore `Coding Standard` job failure until issues related to PSR12 standard are sorted out.

markstory:
@ADmad Done in t574718c518c499537b40c8dc1bcc6f94a876699f

src/Mailer/Message.php
```diff
@@ -1747,12 +1758,19 @@ protected function decodeForHeader(string $text): string
...
      * @return string File contents in base64 encoding
      */
-    protected function readFile(string $path): string
+    protected function readFile($file): string
     {
-        return chunk_split(base64_encode((string)file_get_contents($path)));
+        if (is_string($file)) {
+            $content = (string)file_get_contents($file);
+        } else {
+            $content = (string)$file->getStream();
```
markstory: Shouldn't this be the type we check for instead of `string`?

ADmad: Umm.. no?  `$file` can either be a string filesystem path or an `UploadedFileInterface` instance, that's what the `is_string()` check is for.

markstory: But we could also get an `array` or other invalid value which will turn into a bad method call here. Should we give an exception on a non-string/non uploadinterface object?

ADmad: I added check and exception in `setAttachments()` where various other checks are done too.

Issue #13689, opened 2019-09-28
Make TableSchema class name configurable


bancer:
This is a (multiple allowed):

* [ ] bug
* [x] enhancement
* [x] feature-discussion (RFC)

* CakePHP Version: 3.x.
* Platform and Target: Any.

### What you did
I would like to override TableSchema in order to add more functionality.

### What happened
I found that the construction is hard coded by using `new TableSchema` that makes it very difficult to extend TableSchema.

### What you expected to happen
I would like to make it configurable - what class to use as TableSchema.
One option would be to set it like connection and database driver classes are set for `ConnectionManager::setConfig` method:
```
    'className' => Connection::class,
    'driver' => Mysql::class,
    'tableSchemaClass' => TableSchema::class,
```
Another option would be to introduce a new method to Cake\Database\Driver class:
```
public function getTableSchemaClass() 
{
    return TableSchema::class;
}
```
So that instead of hard coded `new TableSchema` the class name would be taken from configuration or Driver. That would allow to extend the functionality of TableSchema.

markstory:
The method on the connection sounds like a good approach to me. Instead of returning the classname would a factory method on the driver work for what you're doing?

bancer:
If you mean something like this in Driver class then it would be okay:
```
public function constructTableSchema($table, $columns => []) {
    return new TableSchema($table, $columns);
}
```

markstory:
Yes that's what I was thinking @bancer

PR #13696, opened 2019-09-30
Refactor TableSchema construction to a method in Driver class

bancer:
#13689

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,6 +36,8 @@ Thumbs.db
 *.tmPreferences.cache
 # Eclipse
 .settings/*
+/.project
+/.buildpath
 # JetBrains, aka PHPStorm, IntelliJ IDEA
 .idea/*
 # NetBeans
--- a/src/Database/Driver.php
+++ b/src/Database/Driver.php
@@ -15,6 +15,7 @@
 namespace Cake\Database;
 
 use Cake\Database\Query;
+use Cake\Database\Schema\TableSchema;
 use Cake\Database\Statement\PDOStatement;
 use InvalidArgumentException;
 use PDO;
@@ -424,6 +425,18 @@ abstract class Driver implements DriverInterface
     public function newCompiler()
     {
         return new QueryCompiler();
+    }
+
+    /**
+     * Constructs new TableSchema.
+     *
+     * @param string $table The table name.
+     * @param array $columns The list of columns for the schema.
+     * @return \Cake\Database\Schema\TableSchema
+     */
+    public function newTableSchema($table, $columns = [])
+    {
+        return new TableSchema($table, $columns);
     }
 
     /**
--- a/src/Database/Driver.php
+++ b/src/Database/Driver.php
@@ -436,7 +436,9 @@ abstract class Driver implements DriverInterface
      */
     public function newTableSchema($table, array $columns = [])
     {
-        return new TableSchema($table, $columns);
+        $className = $this->_config['tableSchema'] ?? TableSchema::class;
+
+        return new $className($table, $columns);
     }
 
     /**
--- a/src/Database/Driver.php
+++ b/src/Database/Driver.php
@@ -432,9 +432,9 @@ abstract class Driver implements DriverInterface
      *
      * @param string $table The table name.
      * @param array $columns The list of columns for the schema.
-     * @return \Cake\Database\Schema\TableSchema
+     * @return \Cake\Database\Schema\TableSchemaInterface
      */
-    public function newTableSchema($table, $columns = [])
+    public function newTableSchema($table, array $columns = [])
     {
         return new TableSchema($table, $columns);
     }
--- a/src/Database/Driver.php
+++ b/src/Database/Driver.php
@@ -436,7 +436,10 @@ abstract class Driver implements DriverInterface
      */
     public function newTableSchema($table, array $columns = [])
     {
-        $className = $this->_config['tableSchema'] ?? TableSchema::class;
+        $className = TableSchema::class;
+        if (isset($this->_config['tableSchema'])) {
+            $className = $this->_config['tableSchema'];
+        }
 
         return new $className($table, $columns);
     }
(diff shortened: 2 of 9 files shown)


src/Database/Driver.php
```diff
@@ -426,6 +427,18 @@ public function newCompiler()
         return new QueryCompiler();
     }
 
+    /**
+     * Constructs new TableSchema.
+     *
+     * @param string $table The table name.
+     * @param array $columns The list of columns for the schema.
+     * @return \Cake\Database\Schema\TableSchema
```
ADmad: Should be the interface `TableSchemaInterface`.

src/Database/Driver.php
```diff
@@ -426,6 +427,18 @@ public function newCompiler()
...
+     */
+    public function newTableSchema($table, $columns = [])
```
ADmad: 2nd argument can be typehinted as `array`.

src/Database/Driver.php
```diff
@@ -426,6 +427,18 @@ public function newCompiler()
...
+    {
+        return new TableSchema($table, $columns);
```
ADmad: You could check for presence of `$_config['tableSchema']` key and if present use that as class name else fallback to `TableSchema`. That would allow users to easily use their custom class without having to override this method, as you proposed in #13689.

src/ORM/Table.php
```diff
@@ -583,7 +584,11 @@ public function setSchema($schema)
                 unset($schema['_constraints']);
             }
 
-            $schema = new TableSchema($this->getTable(), $schema);
+            $connection = $this->getConnection();
+            if ($connection === null) {
```
ADmad: The null check isn't necessary. Throughout the `Table` class there are multiple calls to `$this->getConnection()` which don't check it's return type.

Ensuring that `$this->_connection` is set could be done in `getConnection()` as it's done in 4.x.
https://github.com/cakephp/cakephp/blob/11805ef531f6c7368c93ed0816ea65f6a26cd9f2/src/ORM/Table.php#L479-L488

src/Database/Driver.php
bancer: That I do not understand. For example in TestFixture.php https://github.com/cakephp/cakephp/pull/13696/files#diff-56561cb12d6a5162f76fe28b07514e1cL182 we have `$this->_schema = $connection->getDriver()->newTableSchema($this->table);` now. How would users use their custom table schema class there without overriding `newTableSchema`?

ADmad: They would set the `tableSchema` key in connection key. That doesn't change whether it's your production connection or test suite connection.

src/ORM/Table.php
bancer: I did this check here because many tests failed with null pointer exception as there are many places where tables are constructed with `new` keyword like `new Table(['schema' => $schema])` with no connection. I have no problem in moving that part to `getConnection` but I am afraid that may break someones code as it would set the connection in the constructor before `initialize` method of table class is invoked where the connection could be overridden.

ADmad: It won't create an issue. Except for tests table instances are created by `TableLocator` which always sets the `connection` key in config passed to `Table::_construct()`.

src/Database/Driver.php
bancer: Do you mean the same what I wrote in the [issue](https://github.com/cakephp/cakephp/issues/13689)?
> One option would be to set it like connection and database driver classes are set for ConnectionManager::setConfig method:

    'className' => Connection::class,
    'driver' => Mysql::class,
    'tableSchemaClass' => TableSchema::class,

@markstory suggested the second option with the getter:
> The method on the connection sounds like a good approach to me. 

Or do you suggest both?

ADmad:
```php
$className = $this->_config['tableSchema'] ?? TableSchema::class;

return new $className($table, $columns);
```

src/Datasource/ConnectionInterface.php
```diff
@@ -18,6 +18,7 @@
  * This interface defines the methods you can depend on in
  * a connection.
  *
+ * @method object getDriver() Gets the driver instance.
```
ADmad: We'll need a similar annotation in `DriverInterface` too for the `newTableSchema()` method.

tests/TestCase/Database/DriverTest.php
```diff
@@ -267,6 +268,17 @@ public function testNewCompiler()
...
 
+    /**
+     * Test newTableSchema().
+     *
+     * @return void
+     */
+    public function testNewTableSchema()
+    {
+        $actual = $this->driver->newTableSchema('articles');
+        $this->assertInstanceOf(TableSchema::class, $actual);
```
markstory: Should also include an assertion to check that the table schema has the right table property?

.gitignore
```diff
@@ -36,6 +36,8 @@ Thumbs.db
 *.tmPreferences.cache
 # Eclipse
 .settings/*
+/.project
+/.buildpath
```
garas: Directories should be `dir/*` for consistency.

But overall this list shouldn't extended to much, there are too many specific paths, Eclipse alone can have [many](https://github.com/github/gitignore/blob/master/Global/Eclipse.gitignore).

Better location would be your local `~/.gitconfig`.

ADmad: https://help.github.com/en/articles/ignoring-files#create-a-global-gitignore

bancer: `.project` and `.buildpath` are files but not folders. There is nothing wrong in having a bit longer list of entries in .gitignore especially when new entries fit well into the existing list. I never add Eclipse configuration files to global gitignore and probably will never do. Please do not make it more difficult for developers to contribute.

ndm2: I never understood why people are so keen on stuffing their project level `.gitignore`'s with all sorts of stuff unrelated to the actual project code, given that every Git repository has an `info/exclude` file that can be used for ignoring user environment specific files and folders.

Issue #13509, opened 2019-08-13
Unquoted alias and field names when using 'join' array


bancer:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.7.
* Platform and Target: Apache, Linux.

### What you did
```
$options = [
    ...,
    'join' => [
        'table' => 'orders',
        'alias' => 'Order',
        ...,
    ],
];
$logs = $this->find('all', $options)
    ->enableHydration(false)
    ->toArray();
```
Data source is configured with `'quoteIdentifiers' => true,`.

### What happened
`Order` alias was not escaped in SQL.
```
PDOException: SQLSTATE[42000]: Syntax error or access violation: 1064 You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'Order
```
SQL (partial):
```
... LEFT JOIN orders Order ON ...
```

### What you expected to happen
I would expect it to be quoted:
```
... LEFT JOIN `orders` `Order` ON ...
```

markstory:
I wasn't able to reproduce this problem. I added a test for `join()` in 1cd1aed9625c78a9853449d1e7455d0dff408e9e and I also got the expected result from

```php
        $this->loadFixtures('Articles');
        $table = $this->getTableLocator()->get('Articles');

        $table->getConnection()->getDriver()->enableAutoQuoting(true);
        $query = $table->find('all', [
            'join' => [
                'table' => 'orders',
                'alias' => 'Order',
                'conditions' => '1 = 1',
            ]
        ]);
        debug($query->sql());
```

How would I go about reproducing the problem you're seeing?

bancer:
You have
```
'alias' => 'orders',
'table' => 'Order',
```

Have you tried
```
'alias' => 'Order',
'table' => 'orders',
```

markstory:
Yes. That's in the snippet I posted in my previous comment. Both forms generate quoted names.

```sql
'SELECT "Articles"."id" AS "Articles__id", "Articles"."author_id" AS "Articles__author_id", "Articles"."title" AS "Articles__title", "Articles"."body" AS "Articles__body", "Articles"."published" AS "Articles__published"
FROM "articles" "Articles"
INNER JOIN "orders" "Order" ON 1 = 1'
```
If I invert the names used for alias/table I get:

```sql
[...]

PR #13699, opened 2019-10-01
Add unit test to illustrate the problem with orders join

bancer:
#13509

### DIFF
--- a/tests/TestCase/ORM/TableTest.php
+++ b/tests/TestCase/ORM/TableTest.php
@@ -6705,7 +6705,7 @@ class TableTest extends TestCase
         );
     }
 
-    public function testJoinOrders()
+    public function testJoinOrdersWithAlias()
     {
         $expected = [
             0 => [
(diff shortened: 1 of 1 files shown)


tests/TestCase/ORM/TableTest.php
```diff
@@ -6702,4 +6704,49 @@ public function skipIfSqlServer()
...
+            ],
+            'join' => [
+                [
+                    'table' => 'orders',
+                    'alias' => 'Order',
+                    // With escaped alias the test passes.
+                    //'alias' => '`Order`',
+                    'type' => 'INNER',
+                    'conditions' => [
+                        'Order.product_id = products.id',
```
markstory: This string condition won't be identifier quoted by the ORM.

bancer: This condition creates no problems. The test passes with that. But the alias is a problem.

markstory: This test will never fully pass right now, as it will be run with and without identifier quoting. Without identifier quoting this query is always invalid.

With identifier quoting enabled the generated query is 

```sql
'SELECT "products"."id" AS "products__id", "products"."category" AS "products__category", "products"."name" AS "products__name", "products"."price" AS "products__price", "Order"."id" AS "Order__id", "Order"."product_category" AS "Order__product_category", "Order"."product_id" AS "Order__product_id" FROM "products" "products" INNER JOIN "orders" "Order" ON Order.product_id = products.id'
```

The only unquoted Order alias is the one in your condition string.

bancer: Not really.

With `'alias' => 'Order',` this test fails:
```
/src/Database/Statement/MysqlStatement.php (line 37)
########## DEBUG ##########
object(PDOStatement) {
	queryString => 'SELECT products.id AS `products__id`, products.category AS `products__category`, products.name AS `products__name`, products.price AS `products__price`, Order.id AS `Order__id`, Order.product_category AS `Order__product_category`, Order.product_id AS `Order__product_id` FROM products products INNER JOIN orders Order ON Order.product_id = products.id'
}
```

With `'alias' => '`Order`',` this test passes:
```
/src/Database/Statement/MysqlStatement.php (line 37)
########## DEBUG ##########
object(PDOStatement) {
	queryString => 'SELECT products.id AS `products__id`, products.category AS `products__category`, products.name AS `products__name`, products.price AS `products__price`, Order.id AS `Order__id`, Order.product_category AS `Order__product_category`, Order.product_id AS `Order__product_id` FROM products products INNER JOIN orders `Order` ON Order.product_id = products.id'
}
```

You can copy-paste these queries to MySQL Workbench to see that the first one is invalid but the second one is valid.

ADmad: The failing query is generated with quoting disabled. The test suite runs all db related tests twice, once with auto quoting disabled and once with enabled. 

Add this statement `$table->getConnection()->getDriver()->enableAutoQuoting(true);` before doing the find call and you won't get any failure.

We already have tests checking that both the table name and alias are quoted with auto quoting enabled.

https://github.com/cakephp/cakephp/blob/448f30918f4a16fc288d84b54bba885d60fe3ebc/tests/TestCase/Database/QueryTest.php#L4059-L4065

ADmad:
Closing as the issue is not reproducible and the related issue #13509 is also closed.

PR #13700, opened 2019-10-02
Fix non-set precision bug in MysqlSchema::columnSql

### LINKED ISSUE #13698: Invalid precision with float data type in MySQL
This is a (multiple allowed):

* [x] bug
* [ ] enhancement

* CakePHP Version: 3.8.1.
* Platform and Target: Apache, MySQL 5.6.

### What you did
Set float data type $fields in AreasFixture:
```
        'latitude' => [
            'type' => 'float',
            'length' => 53,
            'precision' => 7,
            'unsigned' => true,
            'null' => true,
            'default' => null,
            'comment' => ''
        ],
        'longitude' => [
            'type' => 'float',
            'length' => 53,
            'precision' => 7,
            'unsigned' => true,
            'null' => true,
            'default' => null,
            'comment' => ''
        ],
```
We use `double` data type in production database for that though.

Added to $fields in AreasFixture:
```
            [
                'id' => 5,
                'name' => 'My area',
                'latitude' => 55.6760968,
                'longitude' => 12.5683371,
            ],
```

Wrote unit test in AreasTableTest.php:
```
    public function testGet()
    {
        $expected = [
            'id' => 5,
            'name' => 'My area',
            'latitude' => 55.6760968,
            'longitude' => 12.5683371,
        ];
        $actual = $this->Areas->get(5)->toArray();
        $this->assertEquals($expected, $actual);
    }
```

Then I run `vendor/bin/phpunit tests/TestCase/Model/Table/AreasTableTest.php`.

### What happened
The test failed:
```
1) App\Test\TestCase\Model\Table\AreasTableTest::testGet
--- Expected
+++ Actual
@@ @@
-    'latitude' => 55.6760968
-    'longitude' => 12.5683371
+    'latitude' => 55.6760979
+    'longitude' => 12.5683374
```

I would expect the test to pass.

bancer:
Fix #13698

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,7 @@
 /phpunit.xml
 /phpcs.xml
 /vendor
+/vendors
 /composer.phar
 *.mo
 debug.log
--- a/src/Database/Schema/MysqlSchema.php
+++ b/src/Database/Schema/MysqlSchema.php
@@ -389,13 +389,11 @@ class MysqlSchema extends BaseSchema
         }
 
         $hasPrecision = [TableSchema::TYPE_FLOAT, TableSchema::TYPE_DECIMAL];
-        if (in_array($data['type'], $hasPrecision, true)) {
-            if (isset($data['length'])) {
-                if (isset($data['precision'])) {
-                    $out .= '(' . (int)$data['length'] . ',' . (int)$data['precision'] . ')';
-                } else {
-                    $out .= '(' . (int)$data['length'] . ')';
-                }
+        if (in_array($data['type'], $hasPrecision, true) && isset($data['length'])) {
+            if (isset($data['precision'])) {
+                $out .= '(' . (int)$data['length'] . ',' . (int)$data['precision'] . ')';
+            } else {
+                $out .= '(' . (int)$data['length'] . ')';
             }
         }
 
--- a/src/Database/Schema/MysqlSchema.php
+++ b/src/Database/Schema/MysqlSchema.php
@@ -389,10 +389,14 @@ class MysqlSchema extends BaseSchema
         }
 
         $hasPrecision = [TableSchema::TYPE_FLOAT, TableSchema::TYPE_DECIMAL];
-        if (in_array($data['type'], $hasPrecision, true) &&
-            (isset($data['length']) || isset($data['precision']))
-        ) {
-            $out .= '(' . (int)$data['length'] . ',' . (int)$data['precision'] . ')';
+        if (in_array($data['type'], $hasPrecision, true)) {
+            if (isset($data['length'])) {
+                if (isset($data['precision'])) {
+                    $out .= '(' . (int)$data['length'] . ',' . (int)$data['precision'] . ')';
+                } else {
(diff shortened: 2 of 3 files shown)


src/Database/Schema/MysqlSchema.php
```diff
@@ -389,10 +389,14 @@ public function columnSql(TableSchema $schema, $name)
         }
 
         $hasPrecision = [TableSchema::TYPE_FLOAT, TableSchema::TYPE_DECIMAL];
-        if (in_array($data['type'], $hasPrecision, true) &&
-            (isset($data['length']) || isset($data['precision']))
-        ) {
-            $out .= '(' . (int)$data['length'] . ',' . (int)$data['precision'] . ')';
+        if (in_array($data['type'], $hasPrecision, true)) {
+            if (isset($data['length'])) {
```
ADmad: This check can be combined with check in outer `if` itself and avoid an nested `if`.

.gitignore
```diff
@@ -7,6 +7,7 @@
 /phpunit.xml
 /phpcs.xml
 /vendor
+/vendors
```
markstory: We don't use a /vendors directory.

markstory (commented):
Any reason you are targeting 3.next with this? It looks like it could go in master for a 3.8.x release instead of waiting for 3.9

.gitignore
bancer: It is used in 2.x. Otherwise it is impossible to see changes after checkout from 2.x to 3.x in the same repo.

bancer:
> It looks like it could go in master for a 3.8.x release instead of waiting for 3.9

Sounds good.

.gitignore
markstory: How is it impossible?

markstory:
Switching based branches didn't work well in github. I will backport the fixes post merge.

.gitignore
bancer: I mean to find my changes among those listed in vendors folder by git status.

saeideng: this change could be done in separated PR 
in PRs you can focus only in one concept in this case `Fix non-set precision bug in MysqlSchema::columnSq`

bancer: Because when you are in 2.x branch you have vendors folder. When you checkout to 3.x then vendors folder is not removed as it is in gitignore in 2.x.

saeideng: your problem is very limited to only a few developers 
I and most cake developers use clean git repo in local for cake 3

saeideng: right now I have changed gitignore in my local but it is only affect my local 
I think this case is very especial case that you can do it only for your gitignore in your local IMO

bancer: @saeideng why do you want to make contributing more difficult? I really have difficulties to understand such an attitude. I had the problem while working on `columnSql` changes. The fix was one line in gitignore. No side effects. Hurts nobody. That allowed me to fix `columnSql` faster.

markstory:
Merged into master in 37acb61bd95e1d7c781f202ca99924e28f15a794

PR #13701, opened 2019-10-02
Cleanup SecurityComponent.

ADmad:
Cleaned up docblock and leftover code and did some optimizations.

### DIFF
--- a/src/Controller/Component/SecurityComponent.php
+++ b/src/Controller/Component/SecurityComponent.php
@@ -31,12 +31,10 @@ use Cake\Utility\Security;
 
 /**
  * The Security Component creates an easy way to integrate tighter security in
- * your application. It provides methods for various tasks like:
+ * your application. It provides methods for these tasks:
  *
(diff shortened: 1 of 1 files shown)


ADmad:
Few other things that could be done:

- Deprecate remove the `requireSecure` feature. Using `$request->is('ssl')` and throwing exception is easy enough.
- Rename the component to `FormTamperingProtectionComponent` to make it clear what exactly it does.
- Rename `AuthSecurityException` also to something more meaningful. That's a pretty misleading name for an exception used when form tampering check fails.

markstory:
> Deprecate / remove the requireSecure feature. Using $request->is('ssl') and throwing exception is easy enough.

We could add a middleware to do this. Generally requiring SSL is an application wide concern.

ADmad:
> We could add a middleware to do this. Generally requiring SSL is an application wide concern.

Yup a middleware sounds good. 

I'll tackle the extra points I mentioned in a separate PR(s).

PR #13704, opened 2019-10-03
Add HttpsEnforcerMiddleware.

ADmad:
This will allow deprecating / removing the "requireSecure" feature of SecurityComponent.

### DIFF
--- a/src/Http/Middleware/HttpsEnforcerMiddleware.php
+++ b/src/Http/Middleware/HttpsEnforcerMiddleware.php
@@ -84,7 +84,7 @@ class HttpsEnforcerMiddleware implements MiddlewareInterface
         }
 
         throw new BadRequestException(
-            'Request to this URL must be made with HTTPS'
+            'Requests to this URL must be made with HTTPS.'
         );
     }
 }
--- a/src/Http/Middleware/HttpsEnforcerMiddleware.php
+++ b/src/Http/Middleware/HttpsEnforcerMiddleware.php
@@ -68,8 +68,7 @@ class HttpsEnforcerMiddleware implements MiddlewareInterface
      */
     public function process(ServerRequestInterface $request, RequestHandlerInterface $handler): ResponseInterface
     {
-        /** @psalm-suppress UndefinedInterfaceMethod */
-        if ($request->is('ssl')) {
+        if ($request->getUri()->getScheme() === 'https') {
             return $handler->handle($request);
         }
 
--- a/src/Http/Middleware/HttpsEnforcerMiddleware.php
+++ b/src/Http/Middleware/HttpsEnforcerMiddleware.php
@@ -16,6 +16,7 @@ declare(strict_types=1);
  */
 namespace Cake\Http\Middleware;
 
+use Cake\Core\Configure;
 use Cake\Http\Exception\BadRequestException;
 use Psr\Http\Message\ResponseInterface;
 use Psr\Http\Message\ServerRequestInterface;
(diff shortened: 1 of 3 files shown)


src/Http/Middleware/HttpsEnforcerMiddleware.php
```diff
@@ -0,0 +1,90 @@
...
+     * Depending on configuration either redirects to same URL with https or throws exception.
+     *
+     * @param \Psr\Http\Message\ServerRequestInterface $request The request.
+     * @param \Psr\Http\Server\RequestHandlerInterface $handler The request handler.
+     * @return \Psr\Http\Message\ResponseInterface A response.
+     */
+    public function process(ServerRequestInterface $request, RequestHandlerInterface $handler): ResponseInterface
+    {
+        /** @psalm-suppress UndefinedInterfaceMethod */
+        if ($request->is('ssl')) {
```
garas:
```suggestion
        if ($request->getUri()->getScheme() === 'https') {
```

This would check same `Uri` scheme as below, and would avoid psalm/phpstan suppression.

src/Http/Middleware/HttpsEnforcerMiddleware.php
```diff
@@ -0,0 +1,90 @@
...
+ * Redistributions of files must retain the above copyright notice.
+ *
+ * @copyright     Copyright (c) Cake Software Foundation, Inc. (http://cakefoundation.org)
+ * @link          http://cakephp.org CakePHP(tm) Project
+ * @since         4.0.0
+ * @license       http://www.opensource.org/licenses/mit-license.php MIT License
+ */
+namespace Cake\Http\Middleware;
+
+use Cake\Controller\Exception\SecurityException;
```
garas: Exception is associated with controllers, some HTTP related exception like `Cake\Http\Exception\BadRequestException` (parent of `SecurityException`) is better.

src/Http/Middleware/HttpsEnforcerMiddleware.php
ADmad: I don't know why I didn't think of using that for checking when I used the same for setting :P.

src/Http/Middleware/HttpsEnforcerMiddleware.php
```diff
@@ -0,0 +1,90 @@
...
+
+            return new RedirectResponse(
+                $uri,
+                $this->config['statusCode'],
+                $this->config['headers']
+            );
+        }
+
+        throw new BadRequestException(
+            'Request to this URL must be made with HTTPS'
```
markstory:
```suggestion
            'Requests to this URL must be made with HTTPS'
```

tests/TestCase/Http/Middleware/HttpsEnforcerMiddlewareTest.php
```diff
@@ -0,0 +1,97 @@
...
+use TestApp\Http\TestRequestHandler;
+use Zend\Diactoros\Response\RedirectResponse;
+use Zend\Diactoros\Uri;
+
+/**
+ * Test for HttpsEnforcerMiddleware
+ */
+class HttpsEnforcerMiddlewareTest extends TestCase
+{
+    public function testForRequestWithHttp()
```
markstory:
```suggestion
    public function testForRequestWithHttps()
```

tests/TestCase/Http/Middleware/HttpsEnforcerMiddlewareTest.php
```diff
@@ -0,0 +1,97 @@
...
+    {
+        $uri = new Uri('https://localhost/foo');
+        $request = new ServerRequest();
+        $request = $request->withUri($uri);
+
+        $handler = new TestRequestHandler(function ($req) {
+            return new Response();
```
markstory: Should you set something on this response and check that header/body content is the one returned?

ADmad:
Updates done as per feedback.

dereuromark:
In my 3.x HttpsMiddleware I do the same
I also have

	if (Configure::read('debug')) {
		return false;
	}

in there, for local dev to not having to put much logic here inside Application.php

ADmad:
Instead of skipping the check always when debug is on, we can make it configurable. On a staging environment for e.g. someone might prefer the check to be done even with debug on.

dereuromark:
Also, I wonder, should be handle non GET requests differently?

	$method = $request->getMethod();
	if ($method !== 'GET') {
	}

As here a redirect is probably often not useful. Would an exception be better here, if you posted or patched to this URL?

ADmad:
Good catch, redirects should be done only for GET requests.

ADmad:
Changes done.

markstory:
Looks good. I'll write some docs up for this.

PR #13705, opened 2019-10-03
Security: randomToken and shortHash

Erwane:
I often use this methods in my projects, so, i propose them if you think they are cool ;)

**randomToken** : generate a random sensitive case token. It's not hexadecimal but full alphanumeric.
**shortHash** : return a short version of `hash` (like git commit). Really useful to assign uniq id for a string or array.
maybe `shortHash` could be included in `Security::hash` with a new `$length` param.

### DIFF
--- a/src/Utility/Security.php
+++ b/src/Utility/Security.php
@@ -388,4 +388,18 @@ class Security
 
         return static::$_salt = (string)$salt;
     }
+
+    /**
+     * Creates a secure random sensitive case token.
+     *
+     * @param int $length String length. Default 12.
+     * @return string
+     */
+    public static function randomToken($length = 12)
+    {
+        $random = base64_encode(static::randomBytes($length * 4));
+        $clean = preg_replace('/[^A-Za-z0-9]/', '', $random);
+
+        return substr($clean, random_int(1, $length * 2), $length);
+    }
 }
--- a/src/Utility/Security.php
+++ b/src/Utility/Security.php
@@ -405,7 +405,8 @@ class Security
 
     /**
      * Generate a 12 chars short hash (like git signature).
-     * @param  mixed $input Data to hash. Can be string or array
+     *
+     * @param  mixed $input Data to hash. Can be string or array.
      * @return string
      */
     public static function shortHash($input)
--- a/src/Utility/Security.php
+++ b/src/Utility/Security.php
@@ -402,4 +402,18 @@ class Security
 
         return substr($clean, random_int(1, $length * 2), $length);
     }
+
+    /**
+     * Generate a 12 chars short hash (like git signature).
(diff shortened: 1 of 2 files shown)


src/Utility/Security.php
```diff
@@ -388,4 +388,33 @@ public static function salt($salt = null)
...
+        return substr($clean, random_int(1, $length * 2), $length);
+    }
+
+    /**
+     * Generate a 12 chars short hash (like git signature).
+     *
+     * @param  mixed $input Data to hash. Can be string or array.
+     * @return string
+     */
+    public static function shortHash($input)
```
ndm2: This is kind of "false sense of security" territory and IMHO shouldn't be in the core, shortening a hash has rather limited use cases, and it can easily turn into a security problem when not handled with the necessary care!

Git's short hashes are pretty much just aliases that may be used for convenience, internally actual operations will use the full hash. Also they're not just shortened, but shortened with respect to the existing commit hashes, where the specified length is a minimum value that is being increased when required in order to avoid collisions.

markstory: I agree a short hash cannot be used in a security context as it won't have enough entropy to keep it unique or safe from timing/dictionary attacks.

ADmad:
As commented by others the `shortHash()` method is a no-go. As for `randomToken()` having 2 similar methods for random strings/tokens (since there's already `randomString()` doesn't make much sense.

Erwane:
I understand about shortHash.

for the token, it's different of `randomString` because it's full alpha case sensitive, not hexadecimal and more secure for generating secret keys i think.

If you think it's redundant or useful only in 0.001% of projects, we can close this.

src/Utility/Security.php
```diff
@@ -388,4 +388,18 @@ public static function salt($salt = null)
...
+    /**
+     * Creates a secure random sensitive case token.
+     *
+     * @param int $length String length. Default 12.
+     * @return string
+     */
+    public static function randomToken($length = 12)
+    {
+        $random = base64_encode(static::randomBytes($length * 4));
+        $clean = preg_replace('/[^A-Za-z0-9]/', '', $random);
```
othercorey: I'm not sure I understand how this is secure.  secureString only encodes, does not modify the generated bytes.

src/Utility/Security.php
```diff
@@ -388,4 +388,18 @@ public static function salt($salt = null)
...
+
+        return substr($clean, random_int(1, $length * 2), $length);
```
othercorey: Also, it seems like this could fail to produce an appropriate length token depending on the generated bytes.  It would need to verify the length of $clean and re-try or throw an error.

ADmad:
The `randomString()` method also has problems pointed out by @othercorey.

Hence closing the PR.

PR #13711, opened 2019-10-04
Fix reading of form tampering protection token in FormHelper.

ADmad:
Updating the FormHelper was missed when the token was moved from
request param to request attribute. Refs #13669.

### DIFF
--- a/src/Controller/Component/SecurityComponent.php
+++ b/src/Controller/Component/SecurityComponent.php
@@ -486,9 +486,9 @@ class SecurityComponent extends Component
             'unlockedFields' => $this->_config['unlockedFields'],
         ];
 
-        $request->getSession()->write('_Token', $token);
+        $request->getSession()->write('_formToken', $token);
 
-        return $request->withAttribute('securityToken', [
+        return $request->withAttribute('_formToken', [
             'unlockedFields' => $token['unlockedFields'],
         ]);
     }
--- a/src/Controller/Component/SecurityComponent.php
+++ b/src/Controller/Component/SecurityComponent.php
@@ -489,7 +489,7 @@ class SecurityComponent extends Component
             'unlockedFields' => $this->_config['unlockedFields'],
         ];
 
-        return $request->withAttribute('_formToken', [
+        return $request->withAttribute('formToken', [
             'unlockedFields' => $token['unlockedFields'],
         ]);
     }
--- a/src/Controller/Component/SecurityComponent.php
+++ b/src/Controller/Component/SecurityComponent.php
@@ -315,7 +315,7 @@ class SecurityComponent extends Component
         if (strpos($token, ':')) {
             [$token, $locked] = explode(':', $token, 2);
         }
-        unset($check['_Token'], $check['_csrfToken']);
+        unset($check['_Token']);
 
         $locked = explode('|', $locked);
         $unlocked = explode('|', $unlocked);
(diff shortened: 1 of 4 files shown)


src/Controller/Component/SecurityComponent.php
```diff
@@ -486,9 +486,9 @@ public function generateToken(ServerRequest $request): ServerRequest
             'unlockedFields' => $this->_config['unlockedFields'],
         ];
 
-        $request->getSession()->write('_Token', $token);
+        $request->getSession()->write('_formToken', $token);
```
ADmad: I couldn't find anywhere where this value is read back from session. So is this statement required?

ADmad:
Every time I walk through the code of SecurityComponent I keep more more and more left over cruft :laughing:. Latest cleanup included in this PR is related to "disabledFields" config which was deprecated in 3.0.0 itself.

src/Controller/Component/SecurityComponent.php
markstory: Probably not anymore. This may be a relic of the `allowedControllers` and `allowedActions` features.

tests/TestCase/Controller/Component/SecurityComponentTest.php
```diff
@@ -488,7 +488,6 @@ public function testValidatePostIgnoresCsrfToken(): void
         $debug = 'not used';
 
         $this->Controller->setRequest($this->Controller->getRequest()->withParsedBody([
-            '_csrfToken' => 'abc123',
```
markstory: The `_csrfToken` input would still be in request data.

tests/TestCase/Controller/Component/SecurityComponentTest.php
```diff
@@ -789,32 +788,6 @@ public function testValidatePostHidden(): void
         $this->assertNull($result);
     }
 
-    /**
-     * testValidatePostWithDisabledFields method
-     *
-     * @return void
-     * @triggers Controller.startup $this->Controller
-     */
-    public function testValidatePostWithDisabledFields(): void
```
markstory: Do we no longer support disabled fields?

tests/TestCase/Controller/Component/SecurityComponentTest.php
ADmad: Clearing `_csrfToken` from request data is `CsrfProtectionMiddleware`'s job.

https://github.com/cakephp/cakephp/blob/6fec49c1462802c700aaf7fdd0dbf3a0ff92a36e/src/Http/Middleware/CsrfProtectionMiddleware.php#L162-L166

tests/TestCase/Controller/Component/SecurityComponentTest.php
ADmad: Disabled fields were deprecated and super seeded by unlocked fields in 3.0.0 itself. It's no longer documented even. Here's the related commit from 2011 https://github.com/cakephp/cakephp/commit/c1363499191eef0a1829295ba1afb9e5f12807c7.

src/Controller/Component/SecurityComponent.php
ADmad: Okay, I have removed it.

markstory (commented):
Do the integration test case helpers also need to be updated?

ADmad:
> Do the integration test case helpers also need to be updated?

I don't find any reference to the token in request attribute/param in `IntegrationTestTrait` or related classes. It only sets the token in POST data when the key is still `_Token`.

ADmad:
Should I rename `formToken` attribute to `formProtectionToken` to be more descriptive?

ADmad:
So good to merge?

PR #13716, opened 2019-10-05
First draft of http package

markstory:
I've had to move an exception around and add a new interface to allow separation between the controller and http interfaces. While there is one reference to `Router` in `ActionDispatcher` the `class_exists()` check should make it an optional dependency.


src/Http/composer.json
```diff
@@ -0,0 +1,40 @@
...
+        {
+            "name": "CakePHP Community",
+            "homepage": "https://github.com/cakephp/form/graphs/contributors"
+        }
+    ],
+    "support": {
+        "issues": "https://github.com/cakephp/cakephp/issues",
+        "forum": "https://stackoverflow.com/tags/cakephp",
+        "irc": "irc://irc.freenode.org/cakephp",
+        "source": "https://github.com/cakephp/form"
```
saeideng:
```suggestion
        "source": "https://github.com/cakephp/http"
```

src/Http/composer.json
```diff
@@ -0,0 +1,40 @@
...
+        "http",
+        "psr7",
+        "psr15"
+    ],
+    "homepage": "https://cakephp.org",
+    "license": "MIT",
+    "authors": [
+        {
+            "name": "CakePHP Community",
+            "homepage": "https://github.com/cakephp/form/graphs/contributors"
```
saeideng:
```suggestion
            "homepage": "https://github.com/cakephp/http/graphs/contributors"
```

src/Http/ControllerInterface.php
```diff
@@ -0,0 +1,73 @@
...
+     * @throws \UnexpectedValueException If return value of action method is not null or ResponseInterface instance.
+     */
+    public function invokeAction(): ?ResponseInterface;
+
+    /**
+     * Returns true if an action should be rendered automatically.
+     *
+     * @return bool
+     */
+    public function isAutoRenderEnabled(): bool;
```
ADmad: This method feels quite out of place on the interface.

markstory: I agree but the ActionDispatcher does call this method so it was necessary to get typing to work.

src/Http/Exception/MissingControllerException.php
```diff
@@ -0,0 +1,43 @@
...
+
+    /**
+     * Constructor.
+     *
+     * @param string|array $message Either the string of the error message, or an array of attributes
+     *   that are made available in the view, and sprintf()'d into Exception::$_messageTemplate
+     * @param int|null $code The code of the error, is also the HTTP status code for the error. Defaults to 404.
+     * @param \Throwable|null $previous the previous exception.
+     */
+    public function __construct($message, ?int $code = 404, ?Throwable $previous = null)
```
garas: I've looked why there is `404` here, but `500` in `ExceptionsTest::exceptionProvider()`.

It should be no `__construct()` here, but it should be `protected $_defaultCode = 404;` instead. With second argument as `null`, it fallbacks to `500`.

Core exceptions has mixed between `__construct()` and `protected $_defaultCode`, they should be probably normalized in next PR.

src/Http/ControllerInterface.php
ADmad: Couldn't the called checking using `isAutoRenderEnabled()` and the rendering be moved inside `Controller::invokeAction()` itself?

markstory: Yes, that could work. We could move the isAutoRenderEnabled() and render calls there.

markstory: Done :+1:

markstory:
@ADmad Anything else you'd like to see here?

src/Http/ControllerInterface.php
```diff
@@ -0,0 +1,66 @@
...
+
+    /**
+     * Dispatches the controller action. Checks that the action
+     * exists and isn't private.
+     *
+     * @return \Psr\Http\Message\ResponseInterface The resulting response.
+     * @throws \Cake\Controller\Exception\MissingActionException If controller action is not found.
+     * @throws \UnexpectedValueException If return value of action method is not null or ResponseInterface instance.
+     */
+    public function invokeAction(): ?ResponseInterface;
```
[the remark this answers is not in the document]
markstory: If we normalize on ResponseInterface, linters will complain about the methods we've added. We should probably use Response.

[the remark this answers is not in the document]
ADmad: Okay, let's keep the return type as they are for now.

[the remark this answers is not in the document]
markstory: Done.

PR #13718, opened 2019-10-05
4.x csrfmiddleware

ADmad:
- Some minor optimization
- Ensured CSRF token is always cleared from POST data (so that unsetting it isn't necessary in SecurityComponent)
- Changed the request attribute from `csrfToken` to `_csrfToken` for naming consistency and in line with our convention of prefixing special keys / attributes with underscore.
- Removed the `$_defaultConfig` property and set defaults in `$_config` itself. The middleware doesn't use `InstanceConfigTrait` so two properties were unnecessary.

### DIFF
--- a/src/Http/Middleware/CsrfProtectionMiddleware.php
+++ b/src/Http/Middleware/CsrfProtectionMiddleware.php
@@ -102,8 +102,13 @@ class CsrfProtectionMiddleware implements MiddlewareInterface
      */
     public function process(ServerRequestInterface $request, RequestHandlerInterface $handler): ResponseInterface
     {
+        $method = $request->getMethod();
+        $hasData = in_array($method, ['PUT', 'POST', 'DELETE', 'PATCH'], true)
+            || $request->getParsedBody();
+
         if (
-            $this->whitelistCallback !== null
+            $hasData
+            && $this->whitelistCallback !== null
             && call_user_func($this->whitelistCallback, $request) === true
         ) {
             $request = $this->_unsetTokenField($request);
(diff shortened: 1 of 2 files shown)


src/Http/Middleware/CsrfProtectionMiddleware.php
```diff
@@ -102,30 +95,40 @@ public function __construct(array $config = [])
...
         if ($cookieData !== null && strlen($cookieData) > 0) {
-            $request = $request->withAttribute('csrfToken', $cookieData);
+            $request = $request->withAttribute('_csrfToken', $cookieData);
         }
 
-        $method = $request->getMethod();
         if ($method === 'GET' && $cookieData === null) {
             $token = $this->_createToken();
-            $request = $request->withAttribute('csrfToken', $token);
+            $request = $request->withAttribute('_csrfToken', $token);
```
markstory: Is this something we should note in the migration guide? I know some folks use csrftokens in javascript blocks so that xhr requests can have csrf tokens applied.

Instead of predicting this should we remove the prefix from this and _formToken?

ADmad: I am fine removing the prefix from `_formToken` instead. 

Though previously the token was set as a request param, so anyone using `$request->getParam('csrfToken')` will still have to update the call. But I guess if we keep the key same the update can be easily done with rector.

ADmad: I reverted prefixing the attribute name.

markstory: Renaming methods based on the first parameter is something I've added to rector. I could add more patterns for csrfToken too.

ADmad: Okay, so this is good to merge?

ADmad: Sorry for bringing this up again but in 3.x the param name contain underscore prefix so I would prefer we keep it for attribute name too so `getParam('_csrfToken')` calls can be easily replaced with `getAttribute('_csrfToken')`.

I can add back prefix to `formToken` too in the form protector PR.

PR #13721, opened 2019-10-06
Added sqlserver support to travis

othercorey:
I made this change to allow testing against the full suite of databases for future contributions such as fractional seconds support.

This has two key changes:

- Move to xenial distribution (the current default for travis)
- Added sqlserver db support

The xenial distribution was required for sqlserver packages.  This bump makes sense to do anyway. 
 The hidden change that happens here is mysql for xenial is 5.7 by default.  I can work on adding mysql 5.6 back to xenial, but it is non-trivial in the way sqlserver is.

We can install sqlserver pdo for versions >= 7.1.  I enabled just for 7.3 as it does take a little bit of time.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -77,16 +77,10 @@ before_install:
       fi
 
   - if [ $DB = 'sqlserver' ]; then /opt/mssql-tools/bin/sqlcmd -b -S localhost -U SA -P "Strong!Passw0rd" -Q "CREATE DATABASE cakephp_test;"; fi
-  - if [ $DB = 'sqlserver' ]; then /opt/mssql-tools/bin/sqlcmd -b -S localhost -U SA -P "Strong!Passw0rd" -Q "CREATE DATABASE cakephp_test2;"; fi
-  - if [ $DB = 'sqlserver' ]; then /opt/mssql-tools/bin/sqlcmd -b -S localhost -U SA -P "Strong!Passw0rd" -Q "CREATE DATABASE cakephp_test3;"; fi
 
   - if [ $DB = 'mysql' ]; then mysql -u root -e 'CREATE DATABASE cakephp_test;'; fi
-  - if [ $DB = 'mysql' ]; then mysql -u root -e 'CREATE DATABASE cakephp_test2;'; fi
-  - if [ $DB = 'mysql' ]; then mysql -u root -e 'CREATE DATABASE cakephp_test3;'; fi
 
   - if [ $DB = 'pgsql' ]; then psql -c 'CREATE DATABASE cakephp_test;' -U postgres; fi
-  - if [ $DB = 'pgsql' ]; then psql -c 'CREATE SCHEMA test2;' -U postgres -d cakephp_test; fi
-  - if [ $DB = 'pgsql' ]; then psql -c 'CREATE SCHEMA test3;' -U postgres -d cakephp_test; fi
 
   - |
       if [[ ${TRAVIS_PHP_VERSION} != "5.6" ]]; then
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,16 +1,17 @@
 language: php
 
 php:
-  - 7.0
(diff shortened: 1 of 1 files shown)


othercorey:
Would be nice if could order the included jobs so the sqlserver 7.2 runs first as it takes longer.  It would optimize the wait time some.

.travis.yml
```diff
@@ -56,6 +64,22 @@ before_install:
...
+        sudo apt-get install -y mssql-server
+        sudo MSSQL_SA_PASSWORD="Strong!Passw0rd" MSSQL_PID=Express /opt/mssql/bin/mssql-conf -n setup accept-eula
+        sudo ACCEPT_EULA=Y apt-get install -y mssql-tools msodbcsql17 unixodbc-dev
+        pecl install sqlsrv
+        pecl install pdo_sqlsrv
+      fi
+
+  - if [ $DB = 'sqlserver' ]; then /opt/mssql-tools/bin/sqlcmd -b -S localhost -U SA -P "Strong!Passw0rd" -Q "CREATE DATABASE cakephp_test;"; fi
+  - if [ $DB = 'sqlserver' ]; then /opt/mssql-tools/bin/sqlcmd -b -S localhost -U SA -P "Strong!Passw0rd" -Q "CREATE DATABASE cakephp_test2;"; fi
+  - if [ $DB = 'sqlserver' ]; then /opt/mssql-tools/bin/sqlcmd -b -S localhost -U SA -P "Strong!Passw0rd" -Q "CREATE DATABASE cakephp_test3;"; fi
```
markstory: I'm not sure we need all these databases. We don't use any beyond the first in any database server :blush:

othercorey: Let's give it a shot, searching shows none of them are being used afaik.

ADmad:
We already run sqlserver tests on windows on appveyor. People using combination of sqlserver and linuxin production must be quite rare.

So do we really need to contribute to global warming by running sqlserver tests on Linux too? :slightly_smiling_face:

othercorey:
I didn't know it was run somewhere else.  I simply needed a way to run the sqlserver test suite.  

The linux + sqlserver test combination isn't important -- it's just a supported build of sqlserver.

markstory:
@ADmad If SQLServer + linux is more stable that could be a benefit over AppVeyor. We could continue using AppVeyor to ensure things are working on windows and use SQLite as the database to reduce the build time.

othercorey:
If testing on windows is the main concern with the AppVeyor build, I think it's possible to get php + sqlite3 on windows with travis now.  I don't know if you like having a test on a separate CI for other reasons though.

You'd basically have to do an sh environment and use chocolatey to install php + sqlite3.

.travis.yml
```diff
@@ -46,6 +52,19 @@ matrix:
...
+      env: DB=sqlserver db_dsn='sqlserver://SA:Strong!Passw0rd@localhost/cakephp_test?timezone=UTC'
+
+    - php: 7.0
+      env: DB=sqlserver db_dsn='sqlserver://SA:Strong!Passw0rd@localhost/cakephp_test?timezone=UTC'
+
+    - php: 7.2
+      env: DB=sqlserver db_dsn='sqlserver://SA:Strong!Passw0rd@localhost/cakephp_test?timezone=UTC'
+
+    - php: 7.4snapshot
+      env: DB=sqlserver db_dsn='sqlserver://SA:Strong!Passw0rd@localhost/cakephp_test?timezone=UTC'
```
markstory: Instead of using `exclude` you could use `include` to add a specific build for sqlserver + 7.3

othercorey: I tried both (both are in the pull actually if you look at the first commit).

The include puts the build at the end which ends up making the full test suite run longer as sqlserver is the longest test.  This was showing how to make it run first so the total time is the same.

othercorey:
Feel free to close this pull request if you're sure you don't want the changes.

Issue #13694, opened 2019-09-29
Support for convert expression in database


mcsknp:
This is a (multiple allowed):

* [ ] bug
* [x] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.x

### Enhancement description
Add `CONVERT` or `CAST` or both expression to use for
* Get date part from an datetime
* Convert int to varchar to use in concat
* etc...

### Usage:
Using `$query->newExpr()->convert('db_field', 'target_type')` to generate convert expression

markstory:
Wouldn't these methods be `$query->func()->convert(...)`? As all the database agnostic function wrappers are provided by `FunctionsBuilder` not the Query class.

mcsknp:
If there's method to use its, both of `function` and `expression` are ok.

dereuromark:
PR open: https://github.com/cakephp/cakephp/pull/13725

othercorey:
> * Get date part from an datetime

For postgresql, I would use the `to_char()`, `date_part()`, `date_trunct()` or `extract()` functions which are specifically designed for this.

mcsknp:
It seems have not fixed

othercorey:
`FunctionsBuilder::cast()` was added in 4.1.

PR #13723, opened 2019-10-07
Add convert method for convert expression #13694

kingofnull:
I just add the convert method as it requested in issue. It just make `CAST(expression AS datatype)` expression which should works on most DBMS' .


othercorey:
This is missing test cases.  Hard to tell if it works.

PR #13725, opened 2019-10-07
Add convert method for convert expression

### LINKED ISSUE #13694: Support for convert expression in database

kingofnull:
I just add the convert method as it requested in issue. It just make CAST(expression AS datatype) expression which should works on most DBMS' .

Resolves https://github.com/cakephp/cakephp/issues/13694

### DIFF
--- a/src/Database/Expression/QueryExpression.php
+++ b/src/Database/Expression/QueryExpression.php
@@ -531,7 +531,7 @@ class QueryExpression implements ExpressionInterface, Countable
 
         return $this->eq($wrapIdentifier($left), $wrapIdentifier($right));
     }
-    
+
     /**
      * Adds a new type convertion to the expression object in the form "CAST(field AS type)".
      *
@@ -542,9 +542,9 @@ class QueryExpression implements ExpressionInterface, Countable
     public function convert($field, $type)
     {
         $this->add("CAST($field AS $type)");
+
         return $this;
     }
-
 
     /**
      * Returns the string representation of this object so that it can be used in a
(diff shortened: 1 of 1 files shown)


othercorey:
This is missing test cases.

src/Database/Expression/QueryExpression.php
```diff
@@ -532,6 +532,20 @@ public function equalFields($left, $right)
...
     }
 
+    /**
+     * Adds a new type convertion to the expression object in the form "CAST(field AS type)".
+     *
+     * @param string $field Database field to be cast against type
+     * @param string $type target data type to cast field to it
+     * @return $this
+     */
+    public function convert($field, $type)
```
ADmad: Why not name the function `cast()` to match the SQL function name?

kingofnull: the issue request convert name. because of portability I used the  cast instead of convert.

dereuromark:
Should this go into 4.x first? And then get backported to 3.next (3.9)?

src/Database/Expression/QueryExpression.php
```diff
@@ -532,6 +532,20 @@ public function equalFields($left, $right)
...
+    {
+        $this->add("CAST($field AS $type)");
```
ndm2: I have rather conflicted feelings about this. Building the SQL immediately basically turns the cast into a non-expression, as it cannot be traversed, transformed, etc., and it's different to what all other expression builder methods do. Furthermore this neither supports automatic identifier quoting, nor does it allow to pass in expressions for the field, which is different to other expressions too.

In my own projects I've used dedicated function expression classes for casts, as they can benefit from having return types applied.

ADmad: Yup, this looks pretty unsafe.

PR #13726, opened 2019-10-07
return value for __invoke was wrong

eather009:
Corrected  __invoke return value.


dereuromark:
I wonder how this was passing tests and stuff?

eather009:
It should give Fatal error instead.

ADmad:
Maybe this class isn't actually used in tests? If so we can remove it.

markstory:
Merging as this is more correct than before. Removing the class entirely can be done as a separate change.

PR #13727, opened 2019-10-07
`Query::__call()` should be compatible with `QueryTrait::__call()`

mirko-pagliai:
`Query::__call()` should be compatible with `QueryTrait::__call()`

### DIFF
--- a/src/ORM/Query.php
+++ b/src/ORM/Query.php
@@ -1277,7 +1277,7 @@ class Query extends DatabaseQuery implements JsonSerializable, QueryInterface
      * @return mixed
      * @throws \BadMethodCallException if the method is called for a non-select query
      */
-    public function __call($method, $arguments)
+    public function __call(string $method, array $arguments)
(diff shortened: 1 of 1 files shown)


dereuromark:
Is there a test case missing for this scenario?

mirko-pagliai:
> Is there a test case missing for this scenario?

No, I don't think so.
Currently `Query::__call()` calls `QueryTrait::__call()`, therefore current test cases are already valid, because they respect the type hinting.

mirko-pagliai:
However, I have extended `Query` class. Something like that:

```
use Cake\Datasource\QueryTrait;
use Cake\ORM\Query as BaseQuery;

class Query extends BaseQuery
{
    use QueryTrait {
        cache as protected _cache;
    }

    public function cache($key, $config = '')
    {
        //some code here...

        return $this->_cache($key, $config);
    }
}
```
In this case you get a failure.

PR #13728, opened 2019-10-07
4.x form protector

ADmad:
Currently the code form tampering protection token generation and verification is split across `SecurityComponent` and `FormHelper`.  I created a new `FormProtector` class which now encapsulates all related code.

`FormHelper` has been updated to delegate all form token generated related stuff to the new class.

For token verification I don't want to try and retrofit the `SecurityComponent` to use the new class. Two possible are we either create a new `FormProtectionComponent` or a new `FormProtectionMiddleware` and simply deprecate `SecurityComponent`.

The benefit of creating a component is configuring unlocked field and turning of POST validation on per action basis becomes convenient.

TODO:

- [x] Create `FormProtectionComponent` or `FormProtectionMiddleware`.
- [x] Remove / deprecate `SecureFieldTokenTrait` trait.


src/Form/FormProtector.php
```diff
@@ -0,0 +1,563 @@
...
+ * @copyright     Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
+ * @link          https://cakephp.org CakePHP(tm) Project
+ * @since         4.0.0
+ * @license       https://opensource.org/licenses/mit-license.php MIT License
+ */
+namespace Cake\Form;
+
+use Cake\Core\Configure;
+use Cake\Utility\Hash;
+use Cake\Utility\Security;
```
markstory: Does this add new dependencies to cake/form?

src/Form/FormProtector.php
```diff
@@ -0,0 +1,563 @@
...
+
+/**
+ * Protects against form tampering. It ensures that:
+ *
+ * - Form's action (URL) is not modified.
+ * - Unknown / extra fields are not added to the form.
+ * - Existing fields have not been removed from the form.
+ * - Values of hidden inputs have not been changed.
```
markstory: Should this class be marked `@internal`?

tests/TestCase/View/Helper/FormHelperTest.php
```diff
@@ -266,15 +278,18 @@ public function testOrderForRenderingWidgetAndFetchingSecureFields()
...
         $result = $this->Form->widget('select', ['secure' => true, 'name' => '']);
         $this->assertSame('<select name=""></select>', $result);
-        $this->assertEquals([], $this->Form->fields);
+        $result = $this->getProtectedProperty($this->Form->getFormProtector(), 'fields');
+        $this->assertEquals([], $result);
 
         $result = $this->Form->widget('select', ['secure' => true, 'name' => '0']);
         $this->assertSame('<select name="0"></select>', $result);
-        $this->assertEquals(['0'], $this->Form->fields);
+        $result = $this->getProtectedProperty($this->Form->getFormProtector(), 'fields');
```
markstory: Should FormProtector provide get methods for inspecting its state? That would save us using reflection.

src/Form/FormProtector.php
dereuromark: Nope, as validation https://github.com/cakephp/validation/blob/master/composer.json#L27 which already is required.

tests/TestCase/View/Helper/FormHelperTest.php
ADmad: Except for tests I don't see a need to have getters for the properties. Perhaps just adding `__debugInfo()` would suffice?

src/Form/FormProtector.php
ADmad: As @dereuromark pointed out it doesn't. Though I doubt anyone uses the `Form` package as standalone anyway.

dereuromark:
I dont know how relevant this is, but the old FormHelper you added to Shim 4.x branch, seems to fail with changes to 4.x now:
https://travis-ci.org/dereuromark/cakephp-shim/jobs/595381701#L317

tests/TestCase/View/Helper/FormHelperTest.php
markstory: Yeah, pulling out of `__debugInfo()` is reasonable for tests to me.

ADmad:
@dereuromark I'll add tests for 3.x style multiple selects for datetime and check.

ADmad:
I checked and the tests for the Shim plugins FormHelper will have to be updated as the property it's using in tests doesn't exist anymore. There won't be any actual issues in using the datetime selects.

ADmad:
So for the form token verification should I add a `FormProtectionMiddleware` or `FormProtectionComponent`?

markstory:
@ADmad I think a component makes sense here for the reasons you outlined around unlockedFields and disabling per action.

ADmad:
- Added `FormProtectionComponent`
- Removed `SecureFieldTokenTrait`.
- Deprecated `SecurityComponent`.

ADmad:
`SecurityComponent` does some funky juggling related to which expection it throws and with what message.

I simplified `FormProtectionComponent` to always throw `BadRequestException` and just use more detailed error message if debug is turned on.

src/Controller/Component/FormProtectionComponent.php
```diff
@@ -0,0 +1,162 @@
...
+    {
+        if ($this->_config['validationFailureCallback']) {
+            return $this->executeCallback($this->_config['validationFailureCallback'], [$formProtector]);
+        }
+
+        if (!Configure::read('debug')) {
+            throw new BadRequestException(static::DEFAULT_EXCEPTION_MESSAGE);
+        }
+
+        $errorMessage = $formProtector->getError() ?: static::DEFAULT_EXCEPTION_MESSAGE;
```
markstory: Shouldn't the `getError()` method always return a reason?

src/Controller/Component/FormProtectionComponent.php
```diff
@@ -0,0 +1,162 @@
...
+    }
+
+    /**
+     * Execute callback.
+     *
+     * @param callable $callback A valid callable
+     * @param array $params Params
+     * @return \Cake\Http\Response|null
+     */
+    protected function executeCallback(callable $callback, array $params): ?Response
```
markstory: Should we require a closure? Users will have access to `Closure::fromCallable()`

src/Controller/Component/SecurityComponent.php
```diff
@@ -37,6 +37,8 @@
  * - Requiring that SSL be used.
  *
  * @link https://book.cakephp.org/3.0/en/controllers/components/security.html
+ * @deprecated 4.0.0 Use FormProtectionComponent instead, for form tampering protection
+ *   or HttpsEnforceMiddleware to enforce use of HTTPS (SSL) for requests.
```
markstory:
```suggestion
 *   or HttpsEnforcerMiddleware to enforce use of HTTPS (SSL) for requests.
```

src/Form/FormProtector.php
```diff
@@ -0,0 +1,605 @@
...
+ */
+class FormProtector
```
markstory: Should `SecurityComponent` also use this class? That would let us ensure compatibility between `FormProtectionComponent` and `SecurityComponent` and maintain only one copy of complex/sensitive code.

src/Form/FormProtector.php
```diff
@@ -0,0 +1,605 @@
...
+        } else {
+            $this->unlockField($field);
+        }
+
+        return $this;
+    }
+
+    /**
+     * Parses the field name to create a dot separated name value for use in
+     * field hash. If filename is of form Model[field] or Model.fiedl an array of
```
markstory:
```suggestion
     * field hash. If filename is of form Model[field] or Model.field an array of
```

src/Form/FormProtector.php
```diff
@@ -0,0 +1,605 @@
...
+     */
+    protected function extractToken($formData): ?string
+    {
+        if (!is_array($formData)) {
+            $this->debugMessage = 'Request data is not an array.';
+
+            return null;
+        }
+
+        $message = '\'%s\' was not found in request data.';
```
markstory:
```suggestion
        $message = '`%s`' was not found in request data.';
```
Will get formatted better by the development error pages.

src/Form/FormProtector.php
```diff
@@ -0,0 +1,605 @@
...
+
+            return null;
+        }
+        if (Configure::read('debug') && !isset($formData['_Token']['debug'])) {
+            $this->debugMessage = sprintf($message, '_Token.debug');
+
+            return null;
+        }
+        if (!Configure::read('debug') && isset($formData['_Token']['debug'])) {
+            $this->debugMessage = 'Unexpected \'_Token.debug\' found in request data';
```
markstory:
```suggestion
            $this->debugMessage = 'Unexpected `_Token.debug` found in request data';
```

src/Form/FormProtector.php
```diff
@@ -0,0 +1,605 @@
...
+        }
+
+        $expectedParts = json_decode(urldecode($formData['_Token']['debug']), true);
+        if (!is_array($expectedParts) || count($expectedParts) !== 3) {
+            return 'Invalid form protection debug token.';
+        }
+        $expectedUrl = Hash::get($expectedParts, 0);
+        $url = Hash::get($hashParts, 'url');
+        if ($expectedUrl !== $url) {
+            $messages[] = sprintf('URL mismatch in POST data (expected \'%s\' but found \'%s\')', $expectedUrl, $url);
```
markstory:
```suggestion
            $messages[] = sprintf('URL mismatch in POST data (expected `%s` but found `%s`)', $expectedUrl, $url);
```

If you are on-board with using backticks, I can update all the other error messages and fix tests.

src/View/Helper/SecureFieldTokenTrait.php
```diff
@@ -1,71 +0,0 @@
...
- */
-namespace Cake\View\Helper;
-
-use Cake\Utility\Security;
-
-/**
- * Provides methods for building token data that is
- * compatible with SecurityComponent.
- */
-trait SecureFieldTokenTrait
```
markstory: We should add this removal to the migration guide.

ADmad:
@markstory Updates done as per feedback. It would be a big chore to update the tests of `SecurityComponent` if it was updated to use the `FormProtector` class :slightly_frowning_face:. 

`SecurityComponent` is deprecated now and should only be used by those upgrading from 3.x, for a short term until they can update their code to use the new `FormProtectionComponent`.

ADmad:
"Static Analysis" job failure are unrelated and fixed on `4.x`.

src/Form/FormProtector.php
ADmad: Done

PR #13732, opened 2019-10-09
Set value for Limit if parameter found

eather009:
There was an issue with selecting a limit value if any default value was set. Updated 'value' options if has 'limit' query.

`echo $this->Paginator->limitControl([25 => 25, 50 => 50],50); `

Screen Shot 2019-10-09 at 12 12 47

### DIFF
--- a/src/View/Helper/PaginatorHelper.php
+++ b/src/View/Helper/PaginatorHelper.php
@@ -1264,10 +1264,11 @@ class PaginatorHelper extends Helper
             ];
         }
 
+        $query = $this->_View->getRequest()->getQueryParams();
         $out .= $this->Form->control('limit', $options + [
                 'type' => 'select',
                 'label' => __('View'),
-                'value' => $default,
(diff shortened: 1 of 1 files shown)


src/View/Helper/PaginatorHelper.php
```diff
@@ -1264,10 +1264,11 @@ public function limitControl(array $limits = [], $default = null, array $options
             ];
         }
 
+        $query = $this->_View->getRequest()->getQueryParams();
         $out .= $this->Form->control('limit', $options + [
                 'type' => 'select',
                 'label' => __('View'),
-                'value' => $default,
+                'value' => !empty($query['limit'])?$query['limit']:$default,
```
ADmad: Instead of this conditional we can just use the `default` key to specify default value and this line can be changed to:

```php
'value' => $this->_View->getRequest()->getQuery('limit'),
```

dereuromark:
@eather009 Also a small test case for this specific issue would be nice.

PR #13740, opened 2019-10-12
Move route hook into a separate interface

markstory:
Having the route() hook be on `HttpApplicationInterface` complicates using the Http libraries on their own or with an alternate routing implementation. This change makes that scenario easier with no impact to existing applications.


ADmad (commented):
LGTM, the failing test needs to be addressed.

ADmad:
I took care of the CS and stan job failures.

PR #13744, opened 2019-10-13
Use Closure instead of callable for various methods of Routing package.

ADmad:
I don't think anyone would be using invokable classes for connecting routes. For various router/routerbuilder methods we allow passing callback for 2nd params/options argument which is an array so using Closures only for callback are better for type checks.

### DIFF
--- a/src/Routing/RouteBuilder.php
+++ b/src/Routing/RouteBuilder.php
@@ -23,6 +23,7 @@ use Cake\Core\Plugin;
 use Cake\Routing\Route\RedirectRoute;
 use Cake\Routing\Route\Route;
 use Cake\Utility\Inflector;
+use Closure;
 use InvalidArgumentException;
 use RuntimeException;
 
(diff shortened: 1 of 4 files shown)


josegonzalez:
Its funny because I'm pretty sure I demo'd this exact thing - invokable classes for connected routes - a few years back for a microframework...

markstory:
@josegonzalez Do you think we should retain compatibility for invokable classes to be used in route building? How does it help build a microframework app?

josegonzalez:
I think its a nice way to evolve your app into a more complex one from simple closures. [Here](http://www.slimframework.com/docs/v4/objects/routing.html#container-resolution) is a list of what slimphp supports, one of which is an invokable class.

I personally enjoy working with invokable classes as they can be registered easily and don't require a convention around the framework, instead relying on generic php :)

ADmad:
While not clear from my comment my aim wasn't to prevent use of invokable but simply avoid callables as it covers a very wide range of values and callable checks can be expensive.

I can remove the `\Clousure` typehint with `object` which would allow closures and invokable class instances. There are additional type checks already for type of `$callback` argument due the argument juggling in various methods so I will update the checks to allow invokables too.

ADmad:
But using `object` typehint (or as docblock type) will most likely make static analyzers unhappy when trying to invoke the objects. Omitting typehint and specifying type as `callable` only in docblocks would be misleading as users my think all valid callables can be used :disappointed:.

I wish PHP have a specific type for invokable classes :slightly_smiling_face:.

markstory:
Perhaps we should stick with callable given the awkward alternative typehints.

ADmad:
Yup, there might be some clean up I can salvage out of this so I will keep this open for now and update the PR or close later.

PR #13745, opened 2019-10-13
Use Closure instead of callable in MiddlewareQueue

ADmad:


src/Http/MiddlewareQueue.php
```diff
@@ -92,13 +92,13 @@ protected function resolve($middleware): MiddlewareInterface
             return new DoublePassDecoratorMiddleware($middleware);
         }
 
-        return new CallableDecoratorMiddleware($middleware);
+        return new ClosureDecoratorMiddleware($middleware);
```
markstory: If we wanted to maximize compatibility we could leave the accepted types of `add()` and `prepend()` as they are today, and use `Closure::fromCallable()` here when the provided type isn't a Closure.

ADmad: This class has always handled closures only, the class naming and typehints were an oversight on my part. Any other callable is assumed to be 3.x style double pass middleware and is handled by the `DoublePassDecoratorMiddleware` class.

markstory:
Looks good to me.

ADmad:
Should we throw a warning for old style double pass middlewares and nudge people towards updating to new PSR15 middlewares?

PR #13746, opened 2019-10-13
Remove unneeded inline annotations.

ADmad:


src/ORM/Table.php
```diff
@@ -2101,7 +2101,6 @@ protected function _saveMany(iterable $entities, $options = []): iterable
         /** @var bool[] $isNew */
         $isNew = [];
         $cleanup = function ($entities) use (&$isNew): void {
-            /** @var \Cake\Datasource\EntityInterface[] $entities */
```
dereuromark: Sure those are not needed for IDE?

src/View/Form/EntityContext.php
```diff
@@ -649,7 +648,6 @@ protected function _getTable($parts, $fallback = true): ?Table
                     continue;
                 }
             } else {
-                /** @var \Cake\ORM\AssociationCollection $associationCollection */
```
dereuromark: Same maybe here.

src/ORM/Table.php
ADmad: Neither phpstan or psalm needs this annotation. In general we should only worry about type inference by IDEs for  methods/properties used by devs, not framework internals.

src/View/Form/EntityContext.php
ADmad: The `$table` var is already annotated as `ORM\Table` earlier in the method and return type of `associations()` method as `AssociationCollection` typehint, so this annotation is redundant.

src/ORM/Table.php
dereuromark: Those contribute to understanding and ide refactoring.
I strongly disagree with this sentence.

dereuromark: If you want to remove potentielly useful and needed annotations please never do so without using an IDE to back it up and to avoid regressions where not necessary.
A change to remove sth working for the sake of it is a bad change.

dereuromark: Working = followable clickable method link etc.

markstory: Do different IDEs handle type inference differently? Could testing in one still result in another breaking?

ADmad: IDE auto completions are a convenience not a necessity. So while they might be good to have they aren't essentially *needed*.

> Do different IDEs handle type inference differently?

Most have their own static analyzers so they would have varying levels of type inference, though most people who heavily reply on IDE auto completion features are phpstorm users.

> Could testing in one still result in another breaking?

Does any PHP IDE provide a way to use its static analysis feature from command line? If something isn't testable through CI then supporting it is not feasible.

src/Mailer/Email.php
```diff
@@ -501,7 +501,6 @@ public static function deliver(
         if (is_array($config) && !isset($config['transport'])) {
             $config['transport'] = 'default';
         }
-        /** @var \Cake\Mailer\Email $instance */
```
dereuromark: This also kills IDE compatibility completely here

issue1

src/ORM/Table.php
dereuromark: issue2

>  If something isn't testable through CI then supporting it is not feasible.

That is so wrong. That statement in itself explains why this PR and the intension is bad by design.
We do not code nice and shiny documented code for machines or CI cystems, but for humans. You seem to have forgotten that. Just because CI tooling does seem to not care about this, you take away something now from IDE tooling and human usability that people have put work and effort and time into to have.
We worked all hard to get IDE compatibility to a reasonable level and PRs like this just compromise this. It is a slap in the face to all those that put in that effort to get a nice core code to click and read through.
:disappointed:

dereuromark (changes requested):
This PR is reverting necessary IDE compatibility parts that people have worked hard on getting to work.
This is not acceptable.

It also reverts usability in terms of IDE "usage" of methods, "refactor" ability from the IDE and other tools that build on those being present in code.

Again: Never remove sth just because CI system says it is ok to do so. Always only remove if you can back it up by the other more important tool we work on supporting here.

src/ORM/Table.php
dereuromark: He didn't test it. He just blindly removed because CI system didn't complain.
If at least one IDE needs them it is worth keeping for that one, and in no way ever removing a valid one just for the sake of removing.

ADmad:
I'll find better for my time and avoid changes in future which incur the wrath of IDE users, who can't live without the fancy pants features that IDEs provide.

PR #13747, opened 2019-10-13
Fix invalid argument types in SqlserverSchema

garas:
Fix bad merge b55efaa.

Also add `@uses` for BaseSchema methods as IDE can't find usages for dynamically built method names.


markstory:
Thank you for fixing my mistake. :blush:

PR #13750, opened 2019-10-14
Allow dates before 1600 to be valid

### LINKED ISSUE #13441: Validation::date() invalidates dates before 1600
* [x] bug ?
* [x] enhancement ?

* CakePHP Version: 3.x
* Platform and Target: Windows Server 2012, PHP 7.0

### What you did
I tried to insert create a entity whose Table has a validation to Date:

```
<?php

$myTable->newEntity([
    'date' => '1062-07-01'
]);

// and of course:

$validator->date('date');
 
```

### What happened
Date is invalid

To be valid :)

I don't know why but there is explicit restriction on this regex: https://api.cakephp.org/3.0/source-class-Cake.Validation.Validation.html#327-402

That invalidates everything before 1600 !
I even isolated an example here: https://3v4l.org/#preview

Can we accept dates before that? No? Why?

markstory:
Expand the valid date range to start at 0001 instead of 1600. This makes validating historical dates simpler as they are now valid.

Fixes #13441


tests/TestCase/Validation/ValidationTest.php
```diff
@@ -1439,7 +1461,6 @@ public function testDateY()
         $this->assertFalse(Validation::date('20009', ['y']));
         $this->assertFalse(Validation::date(' 2012', ['y']));
         $this->assertFalse(Validation::date('3000', ['y']));
-        $this->assertFalse(Validation::date('1799', ['y']));
```
othercorey: Should we replace this with an equivalent (although weird) somewhere? 31-12-0000

Is simply '0' or '0000' tested somewhere?

tests/TestCase/Validation/ValidationTest.php
```diff
@@ -1429,6 +1450,7 @@ public function testDateYmNumeric()
      */
     public function testDateY()
     {
+        $this->assertTrue(Validation::date('0001', ['y']));
```
othercorey: Are 3-digit years now supported?

tests/TestCase/Validation/ValidationTest.php
markstory: The year 0 is currently not supported. From what I could understand from the gregorian calendar wikipedia's pages there isn't a year 0.

tests/TestCase/Validation/ValidationTest.php
markstory: They shouldn't be. I can add a test for that though.

dereuromark: That also means 2 and 1 digit years then, right?

tests/TestCase/Validation/ValidationTest.php
othercorey: I meant technically that's the first invalid date.

tests/TestCase/Validation/ValidationTest.php
markstory: Supporting one, and two digit years gets hard as the 2 digit years are ambiguous with the modern day two digit format (which is a disaster, but still in use). Three digit years should be possible, but I'm not convinced the added complexity is justified given how infrequently these dates will be used.

othercorey: I'm not convinced it's necessary either.  I think it'd be nice to have a failure test case.

PR #13752, opened 2019-10-14
Added DateTimeFractionalType to support datetimes with fractional seconds

othercorey:
This adds a database datetime type with fractional seconds support.  I know there will be a lot of feedback, but have to start somewhere.  I've added comments to unexpected changes with the reason.

To support fractional seconds, some changes to the base DateTimeType were made to allow a single base class to handle all the cases still.

Marshaling:

The marshal formats now parse the string to build the DateTimeInterface instance instead of using them to compare to the original string.  This is required because fractional seconds cannot be compared to the original string.  You can have 1-6 digits when only 3 and 6 are possible formatting options.

The main issue with this change is Chronos will throw an exception in its overload of DateTime::createFromFormat().  The original createFromFormat() function would return 'false' if parsing failed.  Would be nice if we could configure Chronos to return 'false' in this situation for easier looping.

Cleanup:

- Moved marshal formats to separate property instead of hiding them in $_format.
- Updated property and helper function names used by marshal() to be clear what they do.
- Updated many docblocks for format and helper functions.
- Updated several type hints based on what DateTime type is actually expected.

Missing:

This does not include the changes to map the database column types to DateTimeFractionalType. That change is not complex, but will add it after this type is approved.  We'll have a separate debate on what to call those types and what default to use.

This does not include a datetime type with timezone support.  That type requires a further change to toPHP and manyToPHP that will need a separate debate.


tests/TestCase/Database/Type/DateTimeTypeTest.php
```diff
@@ -212,7 +212,6 @@ public function marshalProvider()
             ['', null],
             ['derpy', null],
             ['2013-nope!', null],
-            ['13-06-26', null],
```
othercorey: 2-digit years are valid in DateTime.  It will produce a year offset from 0 with the 'Y' formatter: '0013'

src/Database/Type/DateTimeType.php
```diff
@@ -408,17 +420,46 @@ public function useMutable()
...
 
-        return $class::parseDateTime($value, $this->_localeFormat);
+        foreach ((array)$this->_marshalFormat as $format) {
+            try {
+                $dateTime = $class::createFromFormat($format, $value);
+                // Check for false in case DateTime is used directly
+                if ($dateTime !== false) {
+                    return $dateTime;
+                }
+            } catch (InvalidArgumentException $e) {
```
othercorey: Unfortunately, Chronos throws this exception instead of returning false as the base class does.

https://github.com/cakephp/chronos/blob/af24b6046827491fb4d10048e42c1fb4a671d840/src/Traits/FactoryTrait.php#L214

src/Database/Type/DateTimeFractionalType.php
```diff
@@ -0,0 +1,46 @@
...
+{
+    /**
+     * @inheritDoc
+     */
+    protected $_format = 'Y-m-d H:i:s.u';
+
+    /**
+     * @inheritDoc
+     */
+    protected $_marshalFormat = [
```
othercorey: Tried to put these in most likely order.

src/Database/Type/DateTimeType.php
```diff
@@ -228,6 +237,9 @@ public function manyToPHP(array $values, array $fields, DriverInterface $driver)
     /**
      * Convert request data into a datetime object.
      *
+     * Does not respect the timezone set by `setTimezone()` when marshalling
```
othercorey: This is the wrong wording.  I added this for something else, but will remove it.

src/Database/Type/DateTimeType.php
```diff
@@ -161,6 +168,8 @@ public function setTimezone($timezone)
     /**
      * Convert strings into DateTime instances.
      *
+     * Respects the timezone set by `setTimezone()`.
```
othercorey: This is not needed.  I added it for something else, but will remove it.

src/Database/Type/DateTimeType.php
```diff
@@ -267,7 +268,7 @@ public function marshal($value): ?DateTimeInterface
         if (is_array($value) && implode('', $value) === '') {
             return null;
         }
-        $value += ['hour' => 0, 'minute' => 0, 'second' => 0];
+        $value += ['hour' => 0, 'minute' => 0, 'second' => 0, 'microsecond' => 0];
```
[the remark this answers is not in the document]
othercorey: I can't find any use that requires changing.  FormHelper seems to just date a string and pass it to DateTime constructor.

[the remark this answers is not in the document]
ADmad: Since the multi selects for datetime inputs were removed from FormHelper in 4.x, no change in FormHelper is necessary.  

Personally I can't even think of a use case for user entered fractional seconds :slightly_smiling_face:.

src/Database/Type/DateTimeType.php
```diff
@@ -301,37 +303,23 @@ public function marshal($value): ?DateTimeInterface
...
     {
         if ($enable === false) {
-            $this->_useLocaleParser = $enable;
+            $this->_useLocaleMarshal = $enable;
 
             return $this;
         }
-        if (method_exists($this->_className, 'parseDateTime')) {
-            $this->_useLocaleParser = $enable;
+        if ($this->_datetimeInstance instanceof \Cake\I18n\I18nDateTimeInterface) {
```
ADmad: Please add a "use" statement for the interface and use `I18nDateTimeInterface::class` here.

src/Database/Type/DateTimeType.php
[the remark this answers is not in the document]
othercorey: @ADmad Should I update the Validation::_getDateString()?  I'm not sure of the use case here.

othercorey:
@markstory @ADmad I want to double check with you on the need for validating the datetime in `DateTimeType::marshal()`.  This is the only type that does this kind of "strict" validation. 

I'm not sure what we (cakephp) guarantee if anything because `DateTimeType` already accepts data in a format that does not represent what it will output.

```
    protected $_marshalFormat = [
        'Y-m-d H:i:s',
        'Y-m-d\TH:i:s',
        'Y-m-d\TH:i:sP',
    ];
```

This includes P which is the timezone offset.  However, the original `DateTimeType` does not support outputting the timezone offset in `toDatabase()`.

I had to change the validation from formatted string comparison to formatted string parsing.  Both have a cost especially as we add more supported formats.

markstory:
Accepting the timezone is useful when handling ISO8601 format dates as many other tools will include the UTC timezone offset. If the application is configured to convert all data into UTC then we don't necessarily need to retain the timezone information past the marshalling stage.

othercorey:
> Accepting the timezone is useful when handling ISO8601 format dates as many other tools will include the UTC timezone offset. If the application is configured to convert all data into UTC then we don't necessarily need to retain the timezone information past the marshalling stage.

I agree.  My point was unclear.  Why don't we just accept the parsing attempt of DateTime constructor instead of trying to parse 7 different formats to see if it matches what we expected.  We'll get the information we want out of it regardless.

ADmad:
>  Why don't we just accept the parsing attempt of DateTime constructor instead of trying to parse 7 different formats to see if it matches what we expected.

If the DateTime constructor can handle all the formats without having to explicitly specify them and return object with correct values set, then why not.

othercorey:
> If the DateTime constructor can handle all the formats without having to explicitly specify each of them and return object with correct values set, then why not.

From local testing, I haven't found any problems with the constructor.  `DateTime::modify()` has an annoying limitation, but that'll come up in the DateTimeTimeZoneType review.

Basically, I would remove the parsing restrictions but still keep the type-specific formatting.  That means DateTimeType, DateTimeFractionalType and DateTimeTimezoneType would all accept the same inputs, but each would retain their unique format when converting to database.

In fact, DateTimeType retains microseconds when converting from database to php through `toPHP()` due to using `DateTime::modify()` which simply parses the full string.  This isn't really any different than someone assigning a DateTime instance that has microseconds to a datetime field.  The microseconds are lost in the conversion to database as expected.

If you don't mind the extra formats in the initial commits, I'll propose this change after the remaining type work is done.

othercorey:
Is there anything blocking this review that I can work on?

markstory:
@othercorey I don't think so, I will try to do another review this weekend.

othercorey:
> @othercorey I don't think so, I will try to do another review this weekend.

No worries.  Take note on the change to Validation::time().  The original docblock says seconds are not permitted, but they were allowed in the regex. I updated the docblock to reflect this after adding fractional seconds.

tests/TestCase/Database/Type/DateTimeFractionalTypeTest.php
```diff
@@ -0,0 +1,396 @@
...
+            // invalid types.
+            [null, null],
+            [false, null],
+            [true, null],
+            ['', null],
+            ['derpy', null],
+            ['2013-nope!', null],
+
+            // valid string types
+            ['2014-02-14 00:00:00.123456', new FrozenTime('2014-02-14 00:00:00.123456')],
```
markstory: Could we have an invalid fractional seconds? Like 7 or 8 points?

markstory (approved):

tests/TestCase/Database/Type/DateTimeFractionalTypeTest.php
othercorey: Yes and no.  In the marshal() format, it will throw an error on extra trailing data.  On the toPHP() format, it will silently ignore them.

I'm hoping to get rid of the strict formatting in the future, but I can add the failure case for marshal and the success case to toPHP.

othercorey: I should note that the original DateTimeType would have failed in the same way essentially.

othercorey:
The static analysis test started failing on unrelated files.

othercorey:
@ADmad can you explain the appveyor error? 

>2019-10-22 21:45:27 Notice: Notice (8): This is a notice error in [/some/file, line 275]

src/Database/Type/DateTimeFractionalType.php
```diff
@@ -0,0 +1,40 @@
...
+ * CakePHP(tm) : Rapid Development Framework (https://cakephp.org)
+ * Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
+ *
+ * Licensed under The MIT License
+ * For full copyright and license information, please see the LICENSE.txt
+ * Redistributions of files must retain the above copyright notice.
+ *
+ * @copyright     Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
+ * @link          https://cakephp.org CakePHP(tm) Project
+ * @since         3.0.0
```
ADmad: 4.0.0

src/Database/Type/DateTimeFractionalType.php
```diff
@@ -0,0 +1,40 @@
...
+ * @license       https://opensource.org/licenses/mit-license.php MIT License
+ */
+namespace Cake\Database\Type;
+
+/**
+ * Extends DateTimeType with support for fractional seconds up to microseconds.
+ */
+class DateTimeFractionalType extends DateTimeType
```
ADmad: Wouldn't  DateTimeMicrosecondType be a better name? Current name sounds quite odd to me.

othercorey: I agree it is a little awkward.  I was trying to do a few things:

1. Keep the name aligned with the various standards language.  
2. Keep 'DateTime' as the prefix to read easier
3. Allow TimeFractionalType later
4. Allow the new types to line up with new type map names: 'datetime??', 'timestamp??'

You can specify fractional seconds kept in the various datetime/timestamp data types.  Both mysql and postgresql allow you to specify 0-6.

The idea is you can actually have a field that stores milliseconds or microseconds or a single digit.  Fractional seconds describes any amount of precision which means it works with any of these type declarations.

If we have DateTimeMicrosecondsType, it could imply that it only works with DATETIME(6) and TIMESTAMPTZ(6) types.

othercorey: Let me know what you want to name it, and I'll change it.  Keep in mind that DateTimeTimezoneType will come after.

ADmad: NVM, existing class name is fine.

othercorey: The alternative is DateTimeWithFractionalSecondsType and TimeWithFractionalSecondsType.

I went with slightly awkward over extremely verbose lol.

src/Database/Type/DateTimeType.php
```diff
@@ -45,36 +47,44 @@ class DateTimeType extends BaseType
...
+     * The DateTime format used when converting to string.
+     *
+     * @var string
+     */
+    protected $_format = 'Y-m-d H:i:s';
+
+    /**
+     * The DateTime formats allowed by `marshal()`.
      *
      * @var string|array
```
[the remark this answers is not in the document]
ADmad: Yes the string type was for BC only. The variable name itself has changed now so BC goes out the window :)

othercorey:
@ADmad I don't want to rush a good review, but as this is mostly clean up I'd love to get a pull request in for the timezone type.  That change requires modifying toPHP() in a way that I'd like to get feedback on.

A preview:  DateTime::modify() allows setting with an absolute time, but it does not use timezone information. It only modifies off the currently set timezone.  That means, I can't use the clone method to instance Time objects when converting from db format.

ADmad:
@othercorey Let's wrap this up first, then you can make a new PR for timezone handling. Smaller PRs are easier to review.

othercorey:
@ADmad Absolutely.  That's what I'm trying to do with these.

markstory (approved):
Looks good to me.

ADmad (approved):

othercorey:
I don't want to be annoying, but is there anything blocking this review?

ADmad:
@othercorey Nothing besides limited amount of time :slightly_smiling_face:

markstory must mostly be working on the docs update before merging thing.

othercorey:
No worries.  I understand that completely.

This change doesn't really add a feature yet which is why I don't have proposed doc changes.  I'm waiting to do that in a separate review so it's easier to review.

Although this adds a new type, it does not hook up to the schema reflection system at all nor does it all new schema types.

markstory:
Sorry, I wanted to write the docs before merging and have been short on time. I made some time tonight though :smile:

othercorey:
@markstory Thanks! Adding support for timestamptz is my main goal so I'm just pushing a bit.  I will expand the docs with the next set of changes.

src/Database/Type/DateTimeFractionalType.php
```diff
@@ -0,0 +1,40 @@
...
+{
+    /**
+     * @inheritDoc
+     */
+    protected $_format = 'Y-m-d H:i:s.u';
+
+    /**
+     * @inheritDoc
+     */
+    protected $_marshalFormats = [
```
ADmad: Going through the PHP manual page for [DatetimeInterface](https://www.php.net/manual/en/class.datetimeinterface.php) I came across the constant `DateTimeInterface::RFC3339_EXTENDED` who's format string has placeholder `v` for milliseconds.

So should we included formats with `v` too in this list?

othercorey: So `v` is milliseconds only.  The reason I switched to createFromFormat() instead of comparing strings is to allow 1-6 digits as you can't really control how many digits the database will return or how many digits a user needs to include.

I hope to remove the strict formatting restrictions in the future.

PR #13755, opened 2019-10-15
4.x routing

ADmad:
Removed multiple checks for callable type. It's now done only once in `RouterBuilder::scope()` which is eventually called by all method related to connecting routes.

### DIFF
--- a/src/Routing/RouteBuilder.php
+++ b/src/Routing/RouteBuilder.php
@@ -980,7 +980,7 @@ class RouteBuilder
      * scope or any child scopes that share the same RouteCollection.
      *
      * @param string $name The name of the middleware. Used when applying middleware to a scope.
-     * @param callable|string $middleware The middleware callable or class name to register.
+     * @param string|\Closure|\Psr\Http\Server\MiddlewareInterface $middleware The middleware to register.
      * @return $this
      * @see \Cake\Routing\RouteCollection
      */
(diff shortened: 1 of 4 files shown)


ADmad:
@markstory Is `RouterCollection::applyMiddleware()` a leftover method? Removing it doesn't case any routing test to fail.

src/Routing/RouteBuilder.php
```diff
@@ -932,13 +926,12 @@ public function plugin(string $name, $options = [], ?callable $callback = null)
...
     {
-        if (is_callable($params)) {
+        if (!is_array($params)) {
             $callback = $params;
             $params = [];
         }
         if (!is_callable($callback)) {
-            $msg = 'Need a callable function/object to connect routes.';
-            throw new InvalidArgumentException($msg);
+            throw new InvalidArgumentException('Need a callable function/object to connect routes.');
```
markstory: Could we include what type we got here to help the developer find where they went wrong?

src/Routing/RouteCollection.php
```diff
@@ -477,36 +470,6 @@ public function middlewareExists(string $name): bool
...
     }
 
-    /**
-     * Apply a registered middleware(s) for the provided path
-     *
-     * @param string $path The URL path to register middleware for.
-     * @param string[] $middleware The middleware names to add for the path.
-     * @return $this
-     */
-    public function applyMiddleware(string $path, array $middleware)
```
markstory: Nice find :+1:

PR #13758, opened 2019-10-15
Fix ConsoleOutput styles() API.

dereuromark:
Resolves https://github.com/cakephp/cakephp/issues/13559

One of the last god wrapper beasts dismantled into a clean API.

### DIFF
--- a/src/Console/ConsoleIo.php
+++ b/src/Console/ConsoleIo.php
@@ -414,6 +414,8 @@ class ConsoleIo
     }
 
     /**
+     * Gets defined styles.
+     *
      * @return array
      * @see \Cake\Console\ConsoleOutput::styles()
      */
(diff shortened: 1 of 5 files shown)


markstory:
Looks good to me. I'm pretty sure rector can handle refactoring this method, and it should be infrequently used by applications/plugins.

ADmad (approved):
LGTM, just the CS error needs to be fixed.

ADmad:
> I'm pretty sure rector can handle refactoring this method

We should create an issue to keep track of add such refactoring which we need to added else many would be missed.

markstory:
@Admad I will add the rector pull tonight. Thus far I have been adding them as we go. There are likely some we have missed but we can fix that as we and others start using the upgrade tool/rector.

markstory:
Rector rule added in rectorphp/rector#2172

PR #13762, opened 2019-10-16
Enable Console lib split

markstory:
Key changes here:

* The `ConsoleErrorHandler` has moved to the Error package. This lets us have much simpler dependencies in the Console package.
* The Command base class includes a bunch of integrations with the rest of CakePHP that make it burdensome to use. By adding a CommandInterface we can narrow the dependencies for standalone usage.

Applications that use the console standalone library would need to setup their own `run()` method that uses the optionparser if they want. I'm open to adding a leaner Command base class but don't have a good name for it right now. Any suggestions are welcome.


ADmad:
> I'm open to adding a leaner Command base class but don't have a good name for it right now.

We have `BasePlugin`,  `BaseApplication` etc. so perhaps `BaseCommand`?

saeideng:
> I'm open to adding a leaner Command base class but don't have a good name for it right now. Any suggestions are welcome.

for new class or for renaming `src/console/Command` to new name ?
if is new  what you want to put into it ?

markstory:
Renaming/changing the existing `Command` class would cause a big break for applications and I don't think it is necessary. A name like `BaseCommand` or `AbstractCommand` could work.

markstory:
I've added a `BaseCommand` class. I'm happy to rename it if folks have an alternative name.

src/Console/BaseCommand.php
```diff
@@ -0,0 +1,271 @@
...
+    }
+
+    /**
+     * Implement this method with your command's logic.
+     *
+     * @param \Cake\Console\Arguments $args The command arguments.
+     * @param \Cake\Console\ConsoleIo $io The console io
+     * @return int|null The exit code or null for success
+     */
+    public function execute(Arguments $args, ConsoleIo $io): ?int
```
saeideng: since we have `command` class that extended this class 
probably you can define  this method as abstract

markstory: Good idea :+1:

markstory:
Anything else for this?

src/Console/BaseCommand.php
```diff
@@ -0,0 +1,268 @@
...
+     * the options and arguments are validated and processed.
+     *
+     * @return void
+     */
+    public function initialize(): void
+    {
+    }
+
+    /**
+     * {@inheritDoc}
```
ADmad:
```suggestion
     * @inheritDoc
```

Issue #13763, opened 2019-10-16
Collection::sumOf() requires an int return type


othercorey:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 4.0.0-beta4.

### What you did
Ran `sumOf()` on a float field.

### What happened
PHP error on return types not matching.

>Uncaught TypeError: Return value of Cake\Collection\Collection::sumOf() must be of the type int, float returned in /var/wasterev/vendor/cakephp/cakephp/src/Collection/CollectionTrait.php:331

### What you expected to happen
Allow both float and int values.

I'm 99% sure this is just a return typo, but seemed odd to run into it this late.

CollectionInterface::sumOf() does not specify a return type, but does show 'float|int' in the type hint.

```
     * @return float|int
     */
    public function sumOf($matcher = null);
```

CollectionTrait::sumOf() does specify a return type but inherits docblock.

```
    /**
     * @inheritDoc
     */
    public function sumOf($matcher = null): int
```

ADmad:
Close as related PR is open.

PR #13764, opened 2019-10-16
Removed return type from CollectionTrait::sumOf()

othercorey:
Fixes https://github.com/cakephp/cakephp/issues/13763

`sumOf()` should allow float and integer results.

### DIFF
--- a/tests/TestCase/Collection/CollectionTest.php
+++ b/tests/TestCase/Collection/CollectionTest.php
@@ -1738,9 +1738,16 @@ class CollectionTest extends TestCase
             ['invoice' => ['total' => 200]],
         ];
 
+        $floatItems = [
+            ['invoice' => ['total' => 100.0]],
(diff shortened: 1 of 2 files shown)


tests/TestCase/Collection/CollectionTest.php
```diff
@@ -1769,6 +1769,35 @@ public function testSumOfCallable($items)
         $this->assertEquals(600, $sum);
     }
 
+    /**
+     * Provider for sumOf float tests
+     *
+     * @return array
+     */
+    public function sumOfFloatProvider()
```
ADmad: If there's an existing provider method for `sumOf()` tests it would be better to use the same for float values to.

othercorey: Still getting used to php unit test providers.

ADmad: No worries :)

PR #13766, opened 2019-10-16
Add support for cookies with "SameSite" option.

ADmad:
https://www.owasp.org/index.php/SameSite

I can add / update tests if the changes seem reasonable.

Todo:
- [x]  Add / updates tests
- [x] Update `Http\Client\Response`.


src/Http/Cookie/Cookie.php
```diff
@@ -138,13 +148,50 @@ public function __construct(
...
+     * Factory method to create Cookie instances.
+     *
+     * @param string $name Cookie name
+     * @param string|array $value Value of the cookie
+     * @param array $options Cookies options. Can contain one of following keys:
+     *  expires, path, domain, secure, httponly and samesite.
+     *  (Keys must be lowercase).
+     * @return static
+     */
+    public static function create(string $name, $value, array $options = [])
```
markstory: :+1: This is a bit simpler than the constructor for basic usage.

src/Http/Cookie/Cookie.php
```diff
@@ -535,6 +624,26 @@ public function isExpanded(): bool
         return $this->isExpanded;
     }
 
+    /**
+     * @inheritDoc
+     */
+    public function getOptions(): array
```
markstory: How do you see this being used? Is it for tests, or as part of an application?

src/Http/ResponseEmitter.php
```diff
@@ -264,16 +269,28 @@ protected function emitCookies(array $cookies): void
...
+
+            $name = $data['name'];
+            $value = $data['value'];
+            unset($data['name'], $data['value']);
+            if (!$data['samesite']) {
+                unset($data['samesite']);
+            }
+
+            if (PHP_VERSION_ID >= 70300) {
+                setcookie($name, $value, $cookie['options']);
```
markstory: Should we raise an exception if people try and use samesite cookies when they are on PHP<7.3?

src/Http/Cookie/Cookie.php
ADmad: It is used to gets args for new signature of setcookie() call in ResponseEmitter.

src/Http/ResponseEmitter.php
ADmad: I'll add an exception.

ADmad:
There are few other issues which I want to address:

1. One would expect `Cake\Http\Response::getCookie()` and `Cake\Http\Client\Response::getCookie()` to return similar data but they don't. The former returns an array representation of cookie with keys `name`, `value`, `path` etc. while the latter just return the cookie value.
2. `Cake\Http\Client\Response::getCookieData()` returns cookie data array similar `Cake\Http\Response::getCookie()` but there is discrepancy in key names: `expires` vs `expire`, `httponly` vs `httpOnly`.  
Also for `Http\Response` the `expire` key is timestamp while for `Http\Client\Response` it's a formatted string. 
Is this comment `This method is compatible with older client code that expects date strings instead of timestamps.` for `Http\Cllient\Reponse::convertCookieToArray()`  still relevant?

So my proposal is:

1. Make `getCookie()` method of both response classes return `Cookie` instance instead of array. Using the object methods is lot more convenient dealing using array keys which can easily have typos (especially considering the discrepancy in key name).
2. Add a `CookieInterface::toArray()` method which returns array of form
`['name' => '', 'value' => '', 'options' => ['path' => '', ...]]`. The `convertCookieToArray()` methods of both response classes could then be dropped.

src/Http/ResponseEmitter.php
garas: Could it generate header with `Cookie::toHeaderValue()` and send it using `header()` instead?

Exceptions might be missed if development and production uses different PHP version.

ADmad: I guess that's an option too.

ADmad:
Either I have made a silly mistake which I cannot spot or PHP is trolling me. Using the alternate `setcookie()` signature with 3rd argument as array for PHP 7.3+ throws error:

`ArgumentCountError: Too few arguments to function Cake\Http\setcookie(), 3 passed in /home/runner/work/cakephp/cakephp/src/Http/ResponseEmitter.php on line 228 and at least 5 expected`

garas:
Troll lives in https://github.com/cakephp/cakephp/blob/4.x/tests/TestCase/Http/server_mocks.php

ADmad:
Thanks, I got the hint when i saw `Cake\Http\setcookie()` as function names instead of `setcookie()` but it's too late in the day to go troll hunting, so I'll take care of it tomorrow :stuck_out_tongue:.

ADmad:
I added a new static method `Cookie::createFromHeaderString()` to avoid the duplicate string parsing code in `ResponseEmitter` and `Client/Response`.

Only dealing with the `SameSite` for PHP < 7.3 and changing the return type of `getCookie()` methods of the two response classes remains.

src/Http/ResponseEmitter.php
ADmad: Directly switching to `header()` for PHP < 7.3 would making testing a pain and we would need different set of tests for different PHP versions.

Instead we can use the namespaced `Cake\Http\setcookie()` currently only used in tests in `ResponseEmitter` too which can internally either call PHP's `setcookie()` or `header()`.

ADmad:
I am a bit sure whether `none` should also be allowed as a valid value for `SameSite`. [OWASP](https://www.owasp.org/index.php/SameSite) only mentions `lax` and `strict` as valid values but I have seen various articles which seems to indicate that chrome can have `none` too.

src/Http/ResponseEmitter.php
```diff
@@ -213,68 +214,47 @@ protected function emitHeaders(ResponseInterface $response): void
...
+        }
 
-            foreach ($parts as $part) {
-                if (strpos($part, '=') !== false) {
-                    [$key, $value] = explode('=', $part);
-                } else {
-                    $key = $part;
-                    $value = true;
-                }
+        header('Set-Cookie: ' . $cookie);
```
markstory: Have you checked that this will play nicely with sessions and setcookie()?

How will this work in integration tests?

tests/TestCase/Http/Client/ResponseTest.php
```diff
@@ -354,13 +354,13 @@ public function testGetCookies()
 
         $result = $response->getCookieData('expiring');
         $this->assertSame('soon', $result['value']);
-        $this->assertTrue($result['httponly']);
-        $this->assertTrue($result['secure']);
+        $this->assertTrue($result['options']['httponly']);
+        $this->assertTrue($result['options']['secure']);
```
markstory: This options key and renaming expire to expires is going to break applications. Could we keep this interface the same and add in `samesite`?

src/Http/ResponseEmitter.php
garas: You need to pass in `false` as the second argument to allow multiple `Set-Cookie` headers.

src/Http/ResponseEmitter.php
garas: I would use same `header()` always independently of PHP.

src/Http/Response.php
```diff
@@ -1295,34 +1293,12 @@ public function getCookies(): array
         /** @var \Cake\Http\Cookie\Cookie[] $cookies */
         $cookies = $this->_cookies;
         foreach ($cookies as $cookie) {
-            $out[$cookie->getName()] = $this->convertCookieToArray($cookie);
+            $out[$cookie->getName()] = $cookie->toArray();
```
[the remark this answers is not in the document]
ADmad: I have changed back to flat array without the nesting for `options` key.

src/Http/Cookie/Cookie.php
```diff
@@ -535,6 +701,38 @@ public function isExpanded(): bool
...
+
+    /**
+     * @inheritDoc
+     */
+    public function toArray(): array
+    {
+        return [
+            'name' => $this->name,
+            'value' => $this->getScalarValue(),
+            'options' => $this->getOptions(),
```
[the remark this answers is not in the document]
ADmad: I have changed back to flat array without the nesting for `options` key. But I really want to keep the `expire` to `expires` rename for consistency.

src/Http/ResponseEmitter.php
ADmad: > Have you checked that this will play nicely with sessions and setcookie()?

IDK, I just followed @garas's suggestion. If it doesn't work then users of PHP 7.3 just won't be able to use `SameSite`.

> How will this work in integration tests?

There are no test failures :) 

> You need to pass in false as the second argument to allow multiple Set-Cookie headers.

Fixed

ADmad: Seems the `SameSite` restriction for session cookie can be handled this way https://github.com/delight-im/PHP-Cookie/blob/master/src/Session.php#L146.

markstory: I'm not concerned about SameSite session cookies, more that using the `set-cookie` header can cause sessions to not work. This commit 7ebc2614fd9e32beec80a2c5dee9cf73f288bb1e has a removal in the `ResponseTransformer` that was a workaround for session cookies not being emitted.

src/Http/Cookie/Cookie.php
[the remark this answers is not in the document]
markstory: Would outputting both be an option?

src/Http/ResponseEmitter.php
ADmad: Okay, I really don't want to go down that rabbit hole for now. I'll revert the use of `header()` and simply throw exception is samesite is set for php 7.2.

We can look into supporting samesite attribute for PHP 7.2 at a later date.

src/Http/ResponseEmitter.php
```diff
@@ -227,34 +227,36 @@ protected function emitCookies(array $cookies): void
...
+            return setcookie($cookie->getName(), $cookie->getScalarValue(), $cookie->getOptions());
         }
 
-        header('Set-Cookie: ' . $cookie, true);
+        $path = $cookie->getPath();
+        $sameSite = $cookie->getSameSite();
+        if ($sameSite !== null) {
+            // Temporary hack for PHP 7.2 to set "SameSite" attribute
+            // https://stackoverflow.com/questions/39750906/php-setcookie-samesite-strict
+            $path .= '; samesite=' . $sameSite;
```
markstory: Neat solution.

ADmad:
This is ready for final review.

@markstory The only non BC change is change of key names in array returned by `Http\Response::getCookie()`. But I think the use cases for reading back cookies meant to be sent with response would be pretty rare. Usually those cookies are just set and emitted by server.

ADmad:
The `Static Analysis` job failure is unrelated and fixed on `4.x`.

src/Http/Cookie/Cookie.php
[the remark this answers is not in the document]
ADmad: If we output old keys too then even those staring fresh with 4.0 could inadvertently use the old keys and we wouldn't be able to do this clean up even in 5.0.

src/Http/Cookie/Cookie.php
```diff
@@ -535,6 +701,37 @@ public function isExpanded(): bool
...
+
+    /**
+     * @inheritDoc
+     */
+    public function toArray(): array
+    {
+        return [
+            'name' => $this->name,
+            'value' => $this->getScalarValue(),
+        ] + $this->getOptions();
```
markstory: :clap:

tests/TestCase/Http/Middleware/CsrfProtectionMiddlewareTest.php
```diff
@@ -278,10 +278,10 @@ public function testConfigurationCookieCreate()
         $cookie = $response->getCookie('token');
         $this->assertNotEmpty($cookie, 'Should set a token.');
         $this->assertRegExp('/^[a-f0-9]+$/', $cookie['value'], 'Should look like a hash.');
-        $this->assertWithinRange(strtotime('+1 hour'), $cookie['expire'], 1, 'session duration.');
+        $this->assertWithinRange(strtotime('+1 hour'), $cookie['expires'], 1, 'session duration.');
         $this->assertSame('/dir/', $cookie['path'], 'session path.');
         $this->assertTrue($cookie['secure'], 'cookie security flag missing');
-        $this->assertTrue($cookie['httpOnly'], 'cookie httpOnly flag missing');
+        $this->assertTrue($cookie['httponly'], 'cookie httpOnly flag missing');
```
markstory: These key renames should be added to the migration guide. I don't think rector is going to help us here as we can't easily discern valid replacements.

markstory (approved):

tests/TestCase/Http/Middleware/CsrfProtectionMiddlewareTest.php
ADmad: I'll make a PR for the migration guide.

Issue #13767, opened 2019-10-17
CsrfProtectionMiddleware::_addTokenCookie() failed on unauthenticatedRedirect


steinkel:
This is a (multiple allowed):

* [x] bug

* CakePHP Version: 
  * cakephp/cakephp                       4.x-dev 531751d 
  * cakephp/authentication                2.x-dev 287d30f
  * See al packages > https://gist.github.com/steinkel/392dd45cc523d155d72690b833e73b82
* Platform and Target: ngixn / php7.3-fpm

### What you did
Configured authentication using 
```php
            ->add(new AuthenticationMiddleware($this, [
                'unauthenticatedRedirect' => '/users/login',
                'queryParam' => 'redirect',
            ]));
```

See Application.php here https://gist.github.com/steinkel/a0829216e884ba7a7bdf811d916669a1

### What happened
`Argument 3 passed to Cake\Http\Middleware\CsrfProtectionMiddleware::_addTokenCookie() must be an instance of Cake\Http\Response, instance of Zend\Diactoros\Response\RedirectResponse given, called in /var/virtual/cakefest2019/vendor/cakephp/cakephp/src/Http/Middleware/CsrfProtectionMiddleware.php on line 125`

ADmad:
I had a feeling this would come up eventually :slightly_smiling_face:. `_addTokenCookie()` does need `Cake\Http\Response` response instance since it uses methods to add cookies which are not part of the PSR interface. 

We could change the typehint and add methods checks but similar issue can occur for other middlewares too.

A more reasonable solution would be to make `RoutingMiddleware` return a `Cake\Http\Response` instance instead of `Zend\Diactoros\Response\RedirectResponse` when it catches a `RedirectExcepton`.

steinkel:
We could do something like
```php
            if ($response instanceof RedirectResponse) {
                return $response;
            }
            return $this->_addTokenCookie($token, $request, $response);
```

ADmad:
@steinkel That works too, doesn't make sense setting cookies for a redirect. Mind opening a PR?

steinkel:
closed per https://github.com/cakephp/cakephp/issues/13767

PR #13776, opened 2019-10-18
refs #13767 don't set csrf cookie on redirect

steinkel:
Fix type issue on unauthenticated redirects trying to set the csrf cookie


ADmad:
Test case please.

steinkel:
Stan is not happy right now, do you recall any other place with a similar issue fixed?

ADmad:
phpstan only understands `@var` annotations above variable assignment statements. Just add an ignore for that error.

steinkel:
stan fixed, issue reported is not related with this PR

PR #13780, opened 2019-10-18
Stricter Instance and Static Config traits.

dereuromark:
Concrete issue:
```php
$roles = Configure::read($this->getConfig('rolesKey'));
```
But since the config value for this table was returning null in one instance, suddenly, the whole config array was returned instead of just the subset of the right key (which would be roles array).
With a bit more strict API access, these things are less easy to happen and easier to prevent by design.

### orFail
Adding orFail methods prevents null returns where you don't expect those under normal circumstances.
In the above example:
```php
$roles = Configure::read($this->getConfigOrFail('rolesTable'));
```

### get...() with non nullable key
- config() returns the whole array if needed, similar like the wrappers in other places
- set and get explicitly work on specific keys then.

This way you can typehint the key as string, and no accidental null read returns the whole config, as currently getConfig() would do with accidental null input.

Those changes should be rector auto-fixable.

### DIFF
--- a/src/Core/InstanceConfigTrait.php
+++ b/src/Core/InstanceConfigTrait.php
@@ -18,6 +18,7 @@ namespace Cake\Core;
 
 use Cake\Core\Exception\Exception;
 use Cake\Utility\Hash;
+use RuntimeException;
 
 /**
  * A trait for reading and writing instance config
@@ -39,6 +40,21 @@ trait InstanceConfigTrait
      * @var bool
      */
     protected $_configInitialized = false;
+
+    /**
+     * Gets the whole config array.
+     *
+     * @return array
+     */
+    public function config(): array
+    {
+        if (!$this->_configInitialized) {
+            $this->_config = $this->_defaultConfig;
+            $this->_configInitialized = true;
+        }
+
+        return $this->_config;
+    }
 
     /**
      * Sets the config.
@@ -86,12 +102,6 @@ trait InstanceConfigTrait
      *
      * ### Usage
      *
-     * Reading the whole config:
-     *
-     * ```
-     * $this->getConfig();
-     * ```
-     *
      * Reading a specific value:
      *
      * ```
@@ -110,11 +120,11 @@ trait InstanceConfigTrait
      * $this->getConfig('some-key', 'default-value');
      * ```
      *
-     * @param string|null $key The key to get or null for the whole config.
-     * @param mixed $default The return value when the key does not exist.
-     * @return mixed Configuration data at the named key or null if the key does not exist.
+     * @param string $key The key to get.
+     * @param mixed|null $default The return value when the key does not exist.
+     * @return mixed|null Configuration data at the named key or null if the key does not exist.
      */
-    public function getConfig(?string $key = null, $default = null)
+    public function getConfig(string $key, $default = null)
     {
         if (!$this->_configInitialized) {
             $this->_config = $this->_defaultConfig;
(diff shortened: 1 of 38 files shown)


src/Core/InstanceConfigTrait.php
```diff
@@ -40,6 +41,21 @@ trait InstanceConfigTrait
      */
     protected $_configInitialized = false;
 
+    /**
+     * Gets the whole config array.
+     *
+     * @return array
+     */
+    public function config(): array
```
ADmad: Not too sure about the method name. Given that in the past getConfig() was named config() it can be confusing.

Perhaps name this method allConfig() or getAllConfig().

othercorey: Is `getEntireConfig` against cakephp naming conventions? I think that might read easier.

ADmad: It's not against convention but we don't have any other method with "Entire" in its name. `getAllConfig()` would be more consistent naming.

markstory: config() is also the name of the deprecated method in 3.x. Someone upgrading could have a subtle failure if they didn't address all their deprecations.

src/Core/InstanceConfigTrait.php
```diff
@@ -110,11 +120,11 @@ public function setConfig($key, $value = null, $merge = true)
...
      *
-     * @param string|null $key The key to get or null for the whole config.
-     * @param mixed $default The return value when the key does not exist.
-     * @return mixed Configuration data at the named key or null if the key does not exist.
+     * @param string $key The key to get.
+     * @param mixed|null $default The return value when the key does not exist.
+     * @return mixed|null Configuration data at the named key or null if the key does not exist.
      */
-    public function getConfig(?string $key = null, $default = null)
+    public function getConfig(string $key, $default = null)
```
markstory: This is a pretty notable breaking change that we haven't had a deprecation warning for, which is rough for users.

markstory (commented):
I like the orFail method but the changes to getConfig() are high impact with an unequal amount of benefit to us or end users.

src/Core/InstanceConfigTrait.php
dereuromark: We can backport some of it into 3.x.
I can see the impact on user land code. We have to decide if it is worth it, given the benefits of the typehinting.

markstory: I generally agree that typehinting improves the code. This adds a single typehint, but at the cost of breaking a pretty core method.

dereuromark: Pretty is relative, the current way allows dangerous bugs like above, silently. :)

ADmad: I too don't like this change. Splitting a method which already takes just a single type + null as argument is going too far.

Back porting

src/Core/InstanceConfigTrait.php
```diff
@@ -110,11 +120,11 @@ public function setConfig($key, $value = null, $merge = true)
      * $this->getConfig('some-key', 'default-value');
      * ```
      *
-     * @param string|null $key The key to get or null for the whole config.
-     * @param mixed $default The return value when the key does not exist.
-     * @return mixed Configuration data at the named key or null if the key does not exist.
+     * @param string $key The key to get.
+     * @param mixed|null $default The return value when the key does not exist.
```
ADmad: Why the `|null`?  `mixed` covers all types.

dereuromark: To explicitly state the nullish part over non null mixed.

src/Core/InstanceConfigTrait.php
dereuromark: It is the only way to catch the silent bug of reading all config by accident as above.

ADmad: In the example you have provided the argument to `getConfig()` is a static string, so how does preventing `null` argument relate to that?  In my experience it's quite rare to pass a variable to `getConfig()`, which could unintentionally have null value.

Also you have added the new `getConfigOrFail()` which doesn't allow null argument so one can just use that.

src/Core/InstanceConfigTrait.php
ADmad: That seems quite unnecessary and just adds noise IMO.

src/Core/InstanceConfigTrait.php
dereuromark: Those are two different things.
I am OK with this going into a 5.0 cleanup though, if we want to postpone.
But IMO with rector this should also be solvable at this point.

dereuromark:
Can I add this to the 5.0 todo/wishlist then and we close this?

Issue #13688, opened 2019-09-28
Response status with constants


lpj145:
https://github.com/cakephp/cakephp/blob/cde7a1c39138eadd2ffb2e2d524ad09672ca6c66/src/Http/Response.php#L47

Shouldn't be a constants ? if this turn to constants i think is more reusable and clarity to work with this.

````php
$response
    ->withStatus(\Cake\Http\Response::HTTP_BAD_REQUEST)
`````

maybe on interface off response this make more sense!

Whats your think ? thanks for attention.

lpj145:
Maybe the same rule for the *application types*, all we know about http from init of protocol the things about status and application type don't change.

markstory:
Having constants for commonly used status codes would be good. Having constants would help prevent typos and enable IDE auto-completion.

I'm not sure we'd want constants for every mime type as many are very infrequently used, but for commonly used ones constants could be beneficial.

lpj145:
Okay, i'm can make a pr ?

ADmad:
https://github.com/php-fig/http-message-util
https://github.com/lmc-eu/http-constants

lpj145:
ahaha @ADmad i'm not surprise with you comment, okay, if not have a interest, close this issue!

othercorey:
The http-message-util does not include all of the status codes that cakephp enumerates such as `499` and `599`.

However, does CakePHP allow mentioning such a package for users to use when calling `withStatus()` and checking against `getStatusCode()` if they want constants in their code instead of raw integers?

dereuromark:
> does CakePHP allow mentioning such a package

What you add to your project's composer.json is completely up to you. So yes, of course, the framework doesn't prevent you from using this library :) It just doesnt ship with it by default.

othercorey:
> What you add to your project's composer.json is completely up to you. So yes, of course, the framework doesn't prevent you from using this library :) It just doesnt ship with it by default.

lol, of course :). I meant in terms of documentation like in the doc header of the class.

dereuromark:
I am sure mentioning it in the docs or docblock is a good idea here.
Feel free to make such pull requests.

PR #13785, opened 2019-10-21
Mentioned HTTP status code constant packages in Http\Response

othercorey:
https://github.com/cakephp/cakephp/issues/13688

Mention HTTP status code constant packages exist and remind user that not all status codes are allowed.

Fixed random name/punctuation issue.

### DIFF
--- a/src/Http/Response.php
+++ b/src/Http/Response.php
@@ -34,13 +34,20 @@ use Zend\Diactoros\Stream;
 
 /**
  * Responses contain the response text, status and headers of a HTTP response.
+ *
+ * There are external packages such as `http-message-util` that provide HTTP
+ * status code constants. These can be used with any method that accepts or
+ * returns a status code integer. Keep in mind that these consants might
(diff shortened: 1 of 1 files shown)


src/Http/Response.php
```diff
@@ -34,13 +34,20 @@
 
 /**
  * Responses contain the response text, status and headers of a HTTP response.
+ *
+ * There are external packages such as `http-message-util` that provide HTTP
```
dereuromark: That is not the package name. Vendor part is missing.

src/Http/Response.php
```diff
@@ -975,7 +988,7 @@ public function withStatus($code, $reasonPhrase = '')
     /**
      * Modifier for response status
      *
-     * @param int $code The code to set.
+     * @param int $code The staus code to set.
```
dereuromark: Typo

src/Http/Response.php
othercorey: Thanks for catching that!

PR #13786, opened 2019-10-21
Fixed typos in http status code constants docblocks

othercorey:
Missed typo in https://github.com/cakephp/cakephp/pull/13785.

### DIFF
--- a/src/Http/Response.php
+++ b/src/Http/Response.php
@@ -35,7 +35,7 @@ use Zend\Diactoros\Stream;
 /**
  * Responses contain the response text, status and headers of a HTTP response.
  *
- * There are external packages such as `http-message-util` that provide HTTP
+ * There are external packages such as `fig/http-message-util` that provide HTTP
  * status code constants. These can be used with any method that accepts or
  * returns a status code integer. Keep in mind that these consants might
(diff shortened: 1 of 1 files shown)


src/Http/Response.php
```diff
@@ -35,7 +35,7 @@
 /**
  * Responses contain the response text, status and headers of a HTTP response.
  *
- * There are external packages such as `http-message-util` that provide HTTP
+ * There are external packages such as `fig/http-message-util` that provide HTTP
```
ADmad: Should be `psr` not `fig`.

othercorey: Is it?  I thought I pulled the official name.  I'll double-check.

othercorey: It is fig/http-message-util for the constants not the interface.  Am I missing something?

ADmad: My mistake, seems they use `psr` only for standards packages and `fig` for others.

dereuromark: Quite the bad behavior and antipattern to not match the URL:
https://github.com/php-fig/http-message-util

The correct naming would IMO be `php-fig/http-message-util` - but meeeh

PR #13789, opened 2019-10-21
Add config orFail() wrapper.

dereuromark:
orFail() part from https://github.com/cakephp/cakephp/pull/13780

### DIFF
--- a/src/Core/StaticConfigTrait.php
+++ b/src/Core/StaticConfigTrait.php
@@ -125,7 +125,7 @@ trait StaticConfigTrait
      * The config value for this key must exist, it can never be null.
      *
      * @param string $key The name of the configuration.
-     * @return mixed|null Configuration data at the named key.
+     * @return mixed Configuration data at the named key.
      * @throws \InvalidArgumentException If value does not exist.
      */
     public static function getConfigOrFail(string $key)
--- a/src/Core/StaticConfigTrait.php
+++ b/src/Core/StaticConfigTrait.php
@@ -112,11 +112,29 @@ trait StaticConfigTrait
      * Reads existing configuration.
      *
      * @param string $key The name of the configuration.
-     * @return mixed Configuration data at the named key or null if the key does not exist.
+     * @return mixed|null Configuration data at the named key or null if the key does not exist.
      */
     public static function getConfig(string $key)
     {
         return static::$_config[$key] ?? null;
+    }
+
+    /**
+     * Reads existing configuration for a specific key.
+     *
+     * The config value for this key must exist, it can never be null.
+     *
+     * @param string $key The name of the configuration.
+     * @return mixed|null Configuration data at the named key.
+     * @throws \InvalidArgumentException If value does not exist.
+     */
+    public static function getConfigOrFail(string $key)
+    {
+        if (!isset(static::$_config[$key])) {
+            throw new InvalidArgumentException(sprintf('Expected configuration `%s` not found.', $key));
+        }
+
+        return static::$_config[$key];
     }
 
     /**
--- a/src/Core/Configure.php
+++ b/src/Core/Configure.php
@@ -428,6 +428,7 @@ class Configure
      * @param string $cacheConfig The cache configuration to save into. Defaults to 'default'
      * @param array|null $data Either an array of data to store, or leave empty to store all values.
      * @return bool Success
+     * @throws \RuntimeException
      */
     public static function store(string $name, string $cacheConfig = 'default', ?array $data = null): bool
     {
@@ -448,6 +449,7 @@ class Configure
      * @param string $name Name of the stored config file to load.
      * @param string $cacheConfig Name of the Cache configuration to read from.
      * @return bool Success.
+     * @throws \RuntimeException
      */
     public static function restore(string $name, string $cacheConfig = 'default'): bool
     {
--- a/src/Core/InstanceConfigTrait.php
+++ b/src/Core/InstanceConfigTrait.php
@@ -112,8 +112,8 @@ trait InstanceConfigTrait
      * ```
      *
      * @param string|null $key The key to get or null for the whole config.
-     * @param mixed|null $default The return value when the key does not exist.
-     * @return mixed|null Configuration data at the named key or null if the key does not exist.
+     * @param mixed $default The return value when the key does not exist.
+     * @return mixed Configuration data at the named key or null if the key does not exist.
      */
     public function getConfig(?string $key = null, $default = null)
     {
--- a/src/Core/InstanceConfigTrait.php
+++ b/src/Core/InstanceConfigTrait.php
@@ -18,6 +18,7 @@ namespace Cake\Core;
 
 use Cake\Core\Exception\Exception;
 use Cake\Utility\Hash;
+use InvalidArgumentException;
 
 /**
  * A trait for reading and writing instance config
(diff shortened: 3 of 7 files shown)


src/Core/StaticConfigTrait.php
```diff
@@ -112,13 +113,29 @@ public static function setConfig($key, $config = null): void
...
-     * @return mixed Configuration data at the named key or null if the key does not exist.
+     * @return mixed|null Configuration data at the named key or null if the key does not exist.
      */
     public static function getConfig(string $key)
     {
         return static::$_config[$key] ?? null;
     }
 
+    /**
+     * Reads existing configuration. The value must exist.
```
othercorey: Would it be more correct to say the **key** must exist?

dereuromark: No, the key must exist for this. I corrected the wrong docblock here.

othercorey: Right, that's what I said. The docblock says 'The value must exist', but since you're passing in a key, I think it's easier to say the key must exist.

src/Core/StaticConfigTrait.php
```diff
@@ -112,13 +113,29 @@ public static function setConfig($key, $config = null): void
...
+     *
+     * @param string $key The name of the configuration.
+     * @return mixed|null Configuration data at the named key.
+     * @throws \RuntimeException If value does not exist.
+     */
+    public static function getConfigOrFail(string $key)
+    {
+        if (!isset(static::$_config[$key])) {
+            throw new RuntimeException(sprintf('Expected configuration `%s` not found.', $key));
```
markstory: Should this be `InvalidArgumentException`?

src/Core/StaticConfigTrait.php
dereuromark: No they key can exist, but if it is null the method would still fail. The value for the key must exist we can say.

othercorey: You've lost me.  But, I'll let you guys figure this out!

dereuromark: I adjusted the text to clarify.

dereuromark:
If this is all good, I can do a backport to 3.9 here as well.

dereuromark:
After my fix with adding explicit nullable behavior to those dirty methods, Psalm now correctly outputs a few more warnings in terms of possible null value usage:

```
ERROR: PossiblyNullArgument - src/ORM/Behavior/TreeBehavior.php:494:43 - Argument 1 of Cake\ORM\Table::aliasField cannot be null, possibly null value provided
        $left = $this->_table->aliasField($this->getConfig('left'));


ERROR: PossiblyNullArgument - src/View/StringTemplate.php:184:48 - Argument 3 of str_replace cannot be null, possibly null value provided
            $template = str_replace('%', '%%', $template);


ERROR: PossiblyNullArgument - src/View/Widget/MultiCheckboxWidget.php:203:52 - Argument 1 of strpos cannot be null, possibly null value provided
        if ($checkbox['label'] === false && strpos($this->_templates->get('checkboxWrapper'), '{{input}}') === false) {


ERROR: PossiblyNullArgument - src/View/Widget/RadioWidget.php:220:20 - Argument 1 of strpos cannot be null, possibly null value provided
            strpos($this->_templates->get('radioWrapper'), '{{input}}') === false


------------------------------
15 errors found
------------------------------
```

Should we now adjust those to the proper orFail() methods right away here then?

src/Core/StaticConfigTrait.php
```diff
@@ -112,13 +112,31 @@ public static function setConfig($key, $config = null): void
...
+     *
+     * The config value for this key must exist, it can never be null.
+     *
+     * @param string $key The name of the configuration.
+     * @return mixed|null Configuration data at the named key.
+     * @throws \InvalidArgumentException If value does not exist.
+     */
+    public static function getConfigOrFail(string $key)
+    {
+        if (!isset(static::$_config[$key])) {
```
ADmad: Would be better to avoid the negation and instead return from the `if` block and move the exception below it.

ADmad:
Psalm reporting those errors has it's drawbacks too. For eg. it would unnecessarily complain about nullable return even for usage where 2nd argument is passed a non-null value like `getConfig('foo', 'default');`

ADmad:
Also static analysers can't know if defaults are set for a key through `$_defaultConfig` property and will unnecessary complain for practically every use to `getConfig()`.  I definitely don't want that much noise in my app.

src/Core/StaticConfigTrait.php
dereuromark: That would go abit against the flow of normal bail early principle.
The happy case goes without indentation to the last line of normal execution IMO.
But either way would be no big deal and is small semantics.

dereuromark:
Sure sounds like indeed some drawbacks. We could revert those typehints then for now.

src/Core/StaticConfigTrait.php
ADmad: The negation is an actual extra operation, the rest are just semantics.

src/Core/StaticConfigTrait.php
```diff
@@ -112,13 +112,31 @@ public static function setConfig($key, $config = null): void
...
         return static::$_config[$key] ?? null;
     }
 
+    /**
+     * Reads existing configuration for a specific key.
+     *
+     * The config value for this key must exist, it can never be null.
+     *
+     * @param string $key The name of the configuration.
+     * @return mixed|null Configuration data at the named key.
```
ADmad: The `|null` should be removed there too.

dereuromark: If psalm doesnt fail now, woudnt these help people to use the correct methods as per use case a bit more cleanly? without it they could accidently use the wrong one, with this they have the chance to either use nullable and check for empty, or use the orFail() one with clear exception path.
What do you think?

src/Core/StaticConfigTrait.php
dereuromark: The operation has to be done, either way, the bool cast is nano opt IMO, and the code flow cleaner as per normal standards IMO. Bail early is what we should consistently use, having the happy case following afterwards as clear happy path.

dereuromark:
> I definitely don't want that much noise in my app.

Actually, that noise is helping to guide one when using the wrong method.
Usually you and core should always use the proper method as per use case. If you don't check emptiness and you expect the key to be there, then it is your responsibility to use the non-null return method here IMO.

We should probably revisit adding a cleaner API and the nullable docblocks back some time in the future.

src/Core/StaticConfigTrait.php
ADmad: Pslam doesn't report any failure because there are new usages of this method in the core. If you use the return value of this method for a non nullable argument it will complain. Same for `InstanceConfigTrait::getConfigOrFail()`.

dereuromark: Exactly, and those new code pieces should be using the correct method already. To avoid someone the config not being filled this way, and silently dragging along this longer than necessary.

dereuromark: If others feel also that we "dont care" here, then we can also revert this one, but it sure wouldn't help in terms of strictness and helping to prevent bugs and alike.

ADmad: Huh?? This is the new `getConfigOrFail()` method you added. The description says "The config value for this key must exist, **it can never be null**" and on the other hand the return type claims it **can** be `null`. So the type in docblock contradicts the description, how does that make any sense?

dereuromark: Oh sry, did I rebase that wrongly and now the typehint was added to the wrong method?
You are right here, I am sry, I didnt have the diff in front of me for the discussion.

dereuromark:
I spotted my mistake, now it should be correct as per docblocks.

ADmad (approved):

PR #13790, opened 2019-10-21
Use newer ubuntu distro.

ADmad:
`trusty` aka Ubuntu 14.04 is pretty old now.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,9 +25,6 @@ services:
   - postgresql
   - mysql
 
-addons:
-  postgresql: "9.4"
-
 cache:
   directories:
     - vendor
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,13 +1,13 @@
 language: php
 
+dist: xenial
+
 php:
   - 7.0
   - 5.6
   - 7.2
   - 7.3
   - '7.4snapshot'
-
-dist: trusty
 
 env:
   matrix:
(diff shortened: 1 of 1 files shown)


othercorey:
From my previous pull request for this, FYI that xenial includes mysql 5.7.

othercorey:
@ADmad I don't know if you care, but you don't need the postgres service when you have the postgres 9.4 addon.  You can just remove that line.

ADmad:
@swijsen If your comments were meant to propose using your GH action then I suggest in future please add 1-2 introductory statements :slightly_smiling_face:. Having checked it I don't think it's something we need right now, thanks.

@othercorey I have removed the `addon` instead. It makes more sense to use the default version provided by `xenial` (9.5) as that would be what more people are likely to use.

othercorey:
> @othercorey I have removed the `addon` instead. It makes more sense to use the default version provided by `xenial` (9.5) as that would be what more people are likely to use.

Works for me.  Even though 9.6 is the latest 9.x supported version, using the default version is a nice test case.

One thing I've noticed is cakephp doesn't really have a mechanism for testing new releases of mysql and postgresql.  Do you ever run into issues with release candidates of future versions?

ADmad:
@othercorey Until now we never had any issues with their newer DB releases to warrant having a complex testing matrix for it.

The only issue that had come up was with MySql changing the default behavior related to aggregate/group by column, which is easily resolved by issues a `SET sql_mode` command when server is initialized.

PR #13792, opened 2019-10-22
Use ::class magic constant instead of string.

ADmad:


src/basics.php
```diff
@@ -105,7 +105,7 @@ function stackTrace(array $options = []): void
      */
     function breakpoint(): ?string
     {
-        if ((PHP_SAPI === 'cli' || PHP_SAPI === 'phpdbg') && class_exists('\Psy\Shell')) {
+        if ((PHP_SAPI === 'cli' || PHP_SAPI === 'phpdbg') && class_exists(\Psy\Shell::class)) {
```
dereuromark: Then this should go into use statements.

dereuromark (approved):

PR #13795, opened 2019-10-23
Backport 4.x orFail() for config.

dereuromark:
Backport https://github.com/cakephp/cakephp/pull/13789 to 3.x (3.next).

### DIFF
--- a/src/Core/InstanceConfigTrait.php
+++ b/src/Core/InstanceConfigTrait.php
@@ -192,7 +192,7 @@ trait InstanceConfigTrait
      *
      * The config value for this key must exist, it can never be null.
      *
-     * @param string $key The key to get.
+     * @param string|null $key The key to get.
      * @return mixed Configuration data at the named key
      * @throws \InvalidArgumentException
      */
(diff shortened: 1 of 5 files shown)


dereuromark:
To get phpstan to pass
```
Line   Core/InstanceConfigTrait.php (in context of class Cake\Auth\AbstractPasswordHasher)  
 ------ ------------------------------------------------------------------------------------- 
  201    Variable $key in isset() always exists and is not nullable.
```

Both param docblock lines would have to be modified to have `|null` added.
I think that would be better for 4.x compatibility than silently allowing null to be used as empty string inside, destroying the functionality of the method.
Feedback?

src/Core/InstanceConfigTrait.php
```diff
@@ -186,6 +187,29 @@ public function config($key = null, $value = null, $merge = true)
...
+     *
+     * The config value for this key must exist, it can never be null.
+     *
+     * @param string $key The key to get.
+     * @return mixed Configuration data at the named key
+     * @throws \InvalidArgumentException
+     */
+    public function getConfigOrFail($key)
+    {
+        if (!isset($key)) {
```
markstory: I think this is reasonable given that this is a 'strict mode' method.

markstory (approved):

PR #13796, opened 2019-10-23
Cleanup exceptions to at least use RuntimeException and include previous

dereuromark:
CakePHP should never throw just the most generic exception here, but instead a bit more precise.
Fixing it to at least Runtime for now, as it is already in other similar places.

Also fixed previous to be included where forgotten and fixed error code to expected int type.

### DIFF
--- a/src/Error/ErrorHandler.php
+++ b/src/Error/ErrorHandler.php
@@ -21,8 +21,8 @@ namespace Cake\Error;
 use Cake\Core\App;
 use Cake\Http\ResponseEmitter;
 use Cake\Routing\Router;
-use Exception;
 use Psr\Http\Message\ServerRequestInterface;
+use RuntimeException;
 use Throwable;
 
 /**
@@ -149,7 +149,7 @@ class ErrorHandler extends BaseErrorHandler
      * @param \Throwable $exception The exception being rendered.
      * @param \Psr\Http\Message\ServerRequestInterface|null $request The request.
      * @return \Cake\Error\ExceptionRendererInterface The exception renderer.
-     * @throws \Exception When the renderer class cannot be found.
+     * @throws \RuntimeException When the renderer class cannot be found.
      */
     public function getRenderer(
         Throwable $exception,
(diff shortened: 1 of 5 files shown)


dereuromark:
Failing phpstan must be a defect on master?
```
 ------ ----------------------------------------------------------------------- 
  Line   Http/ResponseEmitter.php                                               
 ------ ----------------------------------------------------------------------- 
  240    Parameter #3 $expires of function setcookie expects int, array given.  
 ------ ----------------------------------------------------------------------- 
```

ADmad:
> Failing phpstan must be a defect on master?

It's a defect in phpstan, it doesn't recognize the new signature of `setcookie()` added in PHP 7.3 even though it's run on PHP 7.3.

src/Core/InstanceConfigTrait.php
```diff
@@ -296,7 +297,7 @@ protected function _configDelete(string $key): void
 
         foreach ($stack as $i => $k) {
             if (!is_array($update)) {
-                throw new Exception(sprintf('Cannot unset %s value', $key));
+                throw new RuntimeException(sprintf('Cannot unset %s value', $key));
```
ADmad: This is an unnecessary change. The exception here is `\Cake\Core\Exception\Exception` which extends `\RuntimeException`, it's not PHP's `\Exception`.

src/View/SerializedView.php
```diff
@@ -127,7 +127,7 @@ function ($v) {
             } catch (Exception | TypeError $e) {
                 throw new SerializationFailureException(
                     'Serialization of View data failed.',
-                    null,
+                    0,
```
ADmad: This should stay `null` as the `Cake\Core\Exception\Exception` sets the error code to value of it's `$_defaultCode` property and if the argument is `null`.

src/Core/InstanceConfigTrait.php
dereuromark: Ah, you are right about this.

src/View/SerializedView.php
dereuromark: Ah, makes sense then.

PR #13805, opened 2019-10-25
Improve console error display

markstory:
I've often found that the console error log format includes way too much noise in the stack trace. I'm often most interested in the file/line numbers in the stack trace, so that's all that remains.

I've also added newlines and small formatting changes to how messages are display so they don't wrap quite so often.


garas:
Related tests broken on Windows

PR #13808, opened 2019-10-27
Account for BaseAuthenticate::unauthenticated() returning true.

### LINKED ISSUE #13807: Auth Component _unauthenticated should accept return value as boolean
This is a (multiple allowed):

* [x] bug
* [ ] enhancement

* CakePHP Version: 4.0.0-beta4
* Platform and Target: Windows 10 FIrefox Nightly 72.0a1 (2019-10-21) (64-bit) Wamp Apache php 7.2.14, Mysql 5.7.24

### What you did
Custom unauthenticated function in plugin to check guest permission of action.

### What happened
If guest permission is available then this function should return true (Boolean).

Auth component following functions must accept return value as Boolean
startup()
authCheck()
_unauthenticated()

ADmad:
Fixes #13807

### DIFF
--- a/src/Controller/Component/AuthComponent.php
+++ b/src/Controller/Component/AuthComponent.php
@@ -354,7 +354,7 @@ class AuthComponent extends Component implements EventDispatcherInterface
         }
         $result = $auth->unauthenticated($controller->getRequest(), $response);
         if ($result !== null) {
-            return $result;
+            return $result instanceof Response ? $result : null;
         }
 
         if (!$controller->getRequest()->is('ajax')) {


markstory (commented):
Seems like something we should have a test for so it doesn't regress again.

ADmad:
> Seems like something we should have a test for so it doesn't regress again.

Yup, though it should be added in 3.x.

src/Controller/Component/AuthComponent.php
```diff
@@ -354,7 +354,7 @@ protected function _unauthenticated(Controller $controller): ?Response
         }
         $result = $auth->unauthenticated($controller->getRequest(), $response);
         if ($result !== null) {
-            return $result;
+            return $result instanceof Response ? $result : null;
```
othercorey: Doesn't this technically treat true and false the same?  Should incorrect results be allowed?

ADmad: Yes, but I am just interested in getting things working again rather than make deprecated code perfect :slightly_smiling_face:.

othercorey: No worries, just poking at things that are of interest to me.

PR #13812, opened 2019-10-29
4.x body parser

ADmad:
- Updated `BodyParserMiddleware` to use `Closure`s instead of `callable`s.
- Removed input data parsing feature from `RequestHandlerMiddleware` as that feature is available through `BodyParserMiddleware`.

### DIFF
--- a/src/Controller/Component/RequestHandlerComponent.php
+++ b/src/Controller/Component/RequestHandlerComponent.php
@@ -172,6 +172,17 @@ class RequestHandlerComponent extends Component
         if (!$this->ext && $isAjax) {
             $this->ext = 'ajax';
         }
+
+        if (
+            !$request->is(['get', 'head', 'options'])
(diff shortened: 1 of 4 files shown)


src/Controller/Component/RequestHandlerComponent.php
```diff
@@ -92,10 +86,6 @@ public function __construct(ComponentRegistry $registry, array $config = [])
                 'xml' => 'Xml',
                 'ajax' => 'Ajax',
             ],
-            'inputTypeMap' => [
```
markstory: This wasn't emitting deprecations in 3.x removing it now could end up breaking applications without much warning.

ADmad: That's an unfortunate oversight. I have added deprecation warning in `startup()` which is shown if input data is populated but request data/parsed body isn't.

PR #13815, opened 2019-10-29
Use constants where possible.

ADmad:


dereuromark (approved):

src/Datasource/QueryTrait.php
```diff
@@ -388,7 +388,7 @@ public function getMapReducers(): array
      * @return $this
      * @throws \InvalidArgumentException
      */
-    public function formatResults(?callable $formatter = null, $mode = 0)
+    public function formatResults(?callable $formatter = null, $mode = self::APPEND)
```
saeideng: this defined in `ORM/Query` not here in trait

dereuromark: You might have to make a follow up PR here.

PR #13816, opened 2019-10-29
Ensure Table::save() always returns documented type.

ADmad:
refs https://github.com/cakephp/docs/issues/6242

### DIFF
--- a/src/ORM/Table.php
+++ b/src/ORM/Table.php
@@ -1832,7 +1832,19 @@ class Table implements RepositoryInterface, EventListenerInterface, EventDispatc
         $event = $this->dispatchEvent('Model.beforeSave', compact('entity', 'options'));
 
         if ($event->isStopped()) {
-            return $event->getResult();
+            $result = $event->getResult();
+            if ($result === null) {
+                return false;
+            }
+
+            if ($result !== false && !($result instanceof EntityInterface)) {
+                throw new RuntimeException(sprintf(
+                    'beforeSave callback must return `false` or `EntityInterface` instance. Got `%s` instead.',
+                    getTypeName($result)
+                ));
+            }
+
+            return $result;
         }
 
         $saved = $this->_associations->saveParents(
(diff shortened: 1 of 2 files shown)


ADmad:
If wonder if making `Table::save()` return `null` instead of `false` would cause too much grief to those upgrading from 3.x. That would allow us to add return typehint and avoid explicit type checks.

src/ORM/Table.php
```diff
@@ -1832,7 +1832,19 @@ protected function _processSave(EntityInterface $entity, ArrayObject $options)
...
         if ($event->isStopped()) {
-            return $event->getResult();
+            $result = $event->getResult();
+            if ($result === null) {
+                return false;
+            }
+
+            if ($result !== false && !($result instanceof EntityInterface)) {
+                throw new RuntimeException(sprintf(
+                    'beforeSave callback must return `false` or `EntityInterface` instance. Got `%s` instead.',
```
markstory:
```suggestion
                    'The `beforeSave` callback must return `false` or `EntityInterface` instance. Got `%s` instead.',
```

tests/TestCase/ORM/TableTest.php
```diff
@@ -2052,6 +2053,52 @@ public function testBeforeSaveStopEvent()
...
+    }
+
+    /**
+     * @group save
+     * @return void
+     */
+    public function testBeforeSaveException()
+    {
+        $this->expectException(RuntimeException::class);
+        $this->expectExceptionMessage('beforeSave callback must return `false` or `EntityInterface` instance. Got `integer` instead.');
```
markstory:
```suggestion
        $this->expectExceptionMessage('The `beforeSave` callback must return `false` or `EntityInterface` instance. Got `integer` instead.');
```

markstory (commented):
I am wary of switching to `null` here. This is an important and frequently use method. The new exception could be a deprecation warning if we wanted to play it safe.

ADmad:
> I am wary of switching to null here. This is an important and frequently use method.

Yup I was vary of that too, so didn't change the return type for `save()`.

> The new exception could be a deprecation warning if we wanted to play it safe.

I think we should keep the new exception. If the callback returns null/void then it's auto converted to `false`. Returning any other value type is just wrong and throwing an exception is the right thing to do.

dereuromark:
I noted down the nullable typehint part for the 5.x topics: https://github.com/cakephp/cakephp/wiki/5.x-Ideas
Until then I guess we can live with the status quo.

PR #13826, opened 2019-11-01
4.x controller response

ADmad:
- Change return type of `ControllerInterface::invokeAction()` to void as it's return value is never used by `ActionDispatcher`.
-  Remove `ControllerInterface::getResponse()`.
    
    Having a method with return type `Response` while other interface methods
   have return type `ResponseInterface` looked fugly. As a consequence
 `shutdownProcess()` now always returns a response as it's the last step in
  in controller action invocation.

### DIFF
--- a/src/Controller/Controller.php
+++ b/src/Controller/Controller.php
@@ -492,11 +492,11 @@ class Controller implements EventListenerInterface, EventDispatcherInterface, Co
      * Dispatches the controller action. Checks that the action
      * exists and isn't private.
      *
-     * @return \Psr\Http\Message\ResponseInterface The resulting response.
+     * @return void
      * @throws \Cake\Controller\Exception\MissingActionException If controller action is not found.
      * @throws \UnexpectedValueException If return value of action method is not null or ResponseInterface instance.
      */
-    public function invokeAction(): ?ResponseInterface
+    public function invokeAction(): void
     {
         $request = $this->request;
 
@@ -526,8 +526,6 @@ class Controller implements EventListenerInterface, EventDispatcherInterface, Co
         if ($result) {
             $this->response = $result;
         }
-
-        return $this->response;
     }
 
     /**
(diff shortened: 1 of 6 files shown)


ADmad:
I am still wondering if having the 3 methods on `ControllerInterface` is necessary. `startupProcess()` and `shutdownProcess()` feel like internal detail to me.

Can we have just `ControllerInterface::invokeAction(): ResponseInterface` and have `Controller::invokeAction()` itself call `Controller::startupProcess()` and `Controller::shutdownProcess()`?

src/Controller/Controller.php
```diff
@@ -492,12 +492,17 @@ public function setResponse(Response $response)
      * Dispatches the controller action. Checks that the action
      * exists and isn't private.
      *
-     * @return \Psr\Http\Message\ResponseInterface The resulting response.
+     * @return \Psr\Http\Message\ResponseInterface
      * @throws \Cake\Controller\Exception\MissingActionException If controller action is not found.
      * @throws \UnexpectedValueException If return value of action method is not null or ResponseInterface instance.
      */
-    public function invokeAction(): ?ResponseInterface
+    public function invokeAction(): ResponseInterface
```
markstory: This is a good simplification of the interface. I am left wondering if it is too opaque though for future plans. We have talked about adding a DI container in 4.x and injecting parameters into controller actions. This interface would require us to give the controller a reference to the container which isn't a pattern I like.

Perhaps instead of invokeAction() controllers should have a method that returns a Closure with the same signature as the action which could allow us to use a container to inject parameters more easily.

There are likely other/better solutions but I would like to avoid giving controllers a reference to the DI container.

ADmad:
We discussed in the dev channel that in order to address the concerns related to DIC raised by markstory, it would be better to let the controller factory handle the action invocation as seen here https://github.com/robotusers/cakephp-di/blob/80c5ed4c69b2e10d91f83a709be649b056800cfa/src/Http/ControllerFactory.php#L84.

To achieve that `ControllerInterface` will be removed and `ControllerFactoryInterface` will be added instead.

@robertpustulka Please share your thoughts in case you have any feedback on this.

PR #13828, opened 2019-11-01
4.x controller factory

ADmad:
Alternative to #13826 

### Todo
- [x]  ~~Move action invocation code under `ControllerFactory::invoke()` and remove `Controller::invokeAction()`.~~  I added a `Controller::getAction()` instead.
- [x]  Move `Http\ControllerFactory` to under `Controller`?


src/Controller/Controller.php
```diff
@@ -509,10 +508,20 @@ public function invokeAction(): ?ResponseInterface
             ]);
         }
 
-        /** @var callable $callable */
-        $callable = [$this, $request->getParam('action')];
+        return Closure::fromCallable([$this, $action]);
```
markstory: Does this allow Reflection to introspect the parameters of the action?

src/Controller/Controller.php
```diff
@@ -509,10 +508,20 @@ public function invokeAction(): ?ResponseInterface
...
+     * Dispatches the controller action.
+     *
+     * @param \Closure $action The action closure.
+     * @param array $args The arguments to be passed when invoking action.
+     * @return void
+     * @throws \UnexpectedValueException If return value of action is not `null` or `ResponseInterface` instance.
+     */
+    public function invokeAction(Closure $action, array $args): void
+    {
+        $result = $action(...$args);
```
markstory: Are you thinking that a DI container could be another closure that wraps the action's closure and performs injection?

src/Controller/ControllerFactory.php
```diff
@@ -14,27 +14,33 @@
  * @since         3.3.0
  * @license       https://opensource.org/licenses/mit-license.php MIT License
  */
-namespace Cake\Http;
+namespace Cake\Controller;
```
markstory: :+1: to the move. Can we include a `class_alias` and deprecation stub for the old name?

src/Controller/Controller.php
ADmad: `ReflectionFunction` does allow introspecting closure args, I'll add a test for it

src/Controller/Controller.php
ADmad: I was going by what's done here https://github.com/robotusers/cakephp-di/blob/master/src/Http/ControllerFactory.php#L106-L134. It manually goes through the arguments list of the closure and generates list of args to pass.

Happy to change if there's a better way to invoke the closure with required args.

src/Controller/ControllerFactory.php
ADmad: Done

src/Controller/Controller.php
markstory: This will work well. Building a new argument list with the injected parameters is simpler than what I was thinking.

src/Controller/ControllerFactory.php
```diff
@@ -0,0 +1,149 @@
...
+            $this->missingController($request);
+        }
+        /** @var string $className */
+        $reflection = new ReflectionClass($className);
+        if ($reflection->isAbstract() || $reflection->isInterface()) {
+            $this->missingController($request);
+        }
+
+        /** @var \Cake\Controller\Controller $controller */
+        $controller = $reflection->newInstance($request);
```
markstory: This is a nice simplification. The controller can create its own response as needed.

src/Http/composer.json
```diff
@@ -26,7 +26,9 @@
         "php": ">=7.2.0",
         "cakephp/core": "^4.0",
         "cakephp/event": "^4.0",
-        "cakephp/utility": "^4.0"
+        "cakephp/utility": "^4.0",
+        "psr/http-server-handler": "^1.0",
+        "zendframework/zend-diactoros": "^2.1"
```
markstory: :+1:

Issue #12282, opened 2018-06-23
[convention][Foreign Key too long]


HumanG33k:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [x] feature-discussion (RFC)

* CakePHP Version: 3.6.5
* Platform and Target: Debian current testing (buster), apache 2.4.33-3+b1, postgresql v10

### What you did
I design my database with cake convention for FK. I currently have several many to many to many to many relations. 

### What happened
 So some FK  have long long long name and it is truncate by pgsql. 

### What you expected to happen
It’s work.
Maybe there is a way to respect length and cakephp conventions ? Or i need to do it without conventions. I think maybe more experiment dev and users should talk about it maybe something about it can be done in V4.

markstory:
You are free to rename the foreign keys on your tables/associations. You will need to use the association options, but after that you shouldn't have any issues.

Out of curiosity what are your table names?

HumanG33k:
Maybe have done something wrong, i'm really tired this days.

zip_codes(__id__, [...]);
towns(__id__, [...], #zip_codes_id);
addresses(__id__, [...], #town_id);
companies(__id__, [...]);
products(__id__, [...]);
roles(__id__, [...])
addresses_companies(__#company_id__, __#address_id__,[...]);
addresses_companies_products(__#addresses_company_company_id__, __#addresses_company_address_id__, __#product_id__, [...]);
addresses_companies_products_roles(__#addresses_companies_product_addresses_company_company_id__, __#addresses_companies_product_addresses_company_address_id__, __#addresses_companies_product_product_id__, __#role_id__, [...]);

I want represent all place used by a company and for each show what is done with a product.
ie :
- Company A product B on C
- Company A sell B on D
- Company A sell E on F
- Company A product E on F

markstory:
> Is bake support other name for FK ?

It should. If you are using `FOREIGN KEY` constraints in your tables bake should read them when generating models. Bake will then use the associations defined in your models when generating views and controllers.

markstory:
Closing as while the conventions can result in long key names in some schemas, you can use the association options to chose shorter names.

PR #13830, opened 2019-11-02
account for PostgreSQL identifier max length

swiffer:
Identifiers longer than 63 characters can be used, but they will be truncated to the allowed length of 63.

### DIFF
--- a/src/Database/Driver.php
+++ b/src/Database/Driver.php
@@ -459,7 +459,7 @@ abstract class Driver implements DriverInterface
         // In PostgreSQL, identifiers — table names, column names, constraint names, etc. — are limited to a
         // maximum length of 63 bytes. Identifiers longer than 63 characters can be used, but they will be truncated
         // to the allowed length of 63.
-        if (PDO::ATTR_DRIVER_NAME === 'pgsql') {
+        if ($this->_connection->getAttribute(PDO::ATTR_DRIVER_NAME) === 'pgsql') {
             return substr($key, 0, 63);
         }
 
--- a/src/Database/Driver.php
+++ b/src/Database/Driver.php
@@ -448,21 +448,14 @@ abstract class Driver implements DriverInterface
     }
 
     /**
-     * Truncates an identifier if the database does
+     * Generates a column identifier respecting database constraints
      *
-     * @param string $key Aliased field name
+     * @param string $key Identifier
      *
-     * @return string Aliased field name as returned by Database
+     * @return string Identifier as returned by Database
      */
-    public function sanitizeIdentifier(string $key)
+    public function generateIdentifier(string $key)
     {
-        // In PostgreSQL, identifiers — table names, column names, constraint names, etc. — are limited to a
-        // maximum length of 63 bytes. Identifiers longer than 63 characters can be used, but they will be truncated
-        // to the allowed length of 63.
-        if ($this->_connection->getAttribute(PDO::ATTR_DRIVER_NAME) === 'pgsql') {
-            return substr($key, 0, 63);
-        }
-
         return $key;
     }
 }
--- a/src/Database/Driver.php
+++ b/src/Database/Driver.php
@@ -446,4 +446,23 @@ abstract class Driver implements DriverInterface
             'connected' => $this->_connection !== null
         ];
     }
+
+    /**
+     * Truncates an identifier if the database does
+     *
+     * @param string $key Aliased field name
+     *
+     * @return string Aliased field name as returned by Database
+     */
+    public function sanitizeIdentifier(string $key)
+    {
+        // In PostgreSQL, identifiers — table names, column names, constraint names, etc. — are limited to a
+        // maximum length of 63 bytes. Identifiers longer than 63 characters can be used, but they will be truncated
+        // to the allowed length of 63.
+        if (PDO::ATTR_DRIVER_NAME === 'pgsql') {
+            return substr($key, 0, 63);
+        }
+
+        return $key;
+    }
 }
--- a/src/Database/Driver/Postgres.php
+++ b/src/Database/Driver/Postgres.php
@@ -129,4 +129,15 @@ class Postgres extends Driver
     {
         return true;
     }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function generateIdentifier(string $key)
+    {
+        // In PostgreSQL, identifiers — table names, column names, constraint names, etc. — are limited to a
+        // maximum length of 63 bytes. Identifiers longer than 63 characters can be used, but they will be truncated
+        // to the allowed length of 63.
+        return substr($key, 0, 63);
+    }
 }
--- a/src/Database/Driver/Postgres.php
+++ b/src/Database/Driver/Postgres.php
@@ -129,29 +129,4 @@ class Postgres extends Driver
     {
         return true;
     }
-
-    /**
-     * {@inheritDoc}
-     */
-    public function calculateColumnMap(Query $query)
-    {
-        $map = [];
-        $defaultAlias = $query->getRepository()->getAlias();
-        foreach ($query->clause('select') as $key => $field) {
-            $key = trim($key, '"`[]');
-
-            if (strpos($key, '__') <= 0) {
-                $map[$defaultAlias][$key] = $key;
-                continue;
-            }
-
-            $parts = explode('__', $key, 2);
-            // In PostgreSQL, identifiers — table names, column names, constraint names, etc. — are limited to a
-            // maximum length of 63 bytes. Identifiers longer than 63 characters can be used, but they will be truncated
-            // to the allowed length of 63.
-            $map[$parts[0]][substr($key, 0, 63)] = $parts[1];
-        }
-
-        return $map;
-    }
 }
--- a/src/Database/Driver/Postgres.php
+++ b/src/Database/Driver/Postgres.php
@@ -129,4 +129,29 @@ class Postgres extends Driver
     {
         return true;
     }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function calculateColumnMap(Query $query)
+    {
+        $map = [];
+        $defaultAlias = $query->getRepository()->getAlias();
+        foreach ($query->clause('select') as $key => $field) {
+            $key = trim($key, '"`[]');
+
+            if (strpos($key, '__') <= 0) {
+                $map[$defaultAlias][$key] = $key;
+                continue;
+            }
+
+            $parts = explode('__', $key, 2);
+            // In PostgreSQL, identifiers — table names, column names, constraint names, etc. — are limited to a
+            // maximum length of 63 bytes. Identifiers longer than 63 characters can be used, but they will be truncated
+            // to the allowed length of 63.
+            $map[$parts[0]][substr($key, 0, 63)] = $parts[1];
+        }
+
+        return $map;
+    }
 }
--- a/src/ORM/ResultSet.php
+++ b/src/ORM/ResultSet.php
@@ -400,7 +400,18 @@ class ResultSet implements ResultSetInterface
      */
     protected function _calculateColumnMap($query)
     {
-        $map = $this->_driver->calculateColumnMap($query);
+        $map = [];
+        foreach ($query->clause('select') as $key => $aliasedField) {
+            $key = trim($key, '"`[]');
+
+            if (strpos($key, '__') <= 0) {
+                $map[$this->_defaultAlias][$key] = $key;
+                continue;
+            }
+
+            $parts = explode('__', $key, 2);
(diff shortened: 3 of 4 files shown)


src/ORM/ResultSet.php
```diff
@@ -524,8 +525,26 @@ protected function _groupResult($row)
             }
         }
 
+        // In PostgreSQL, identifiers — table names, column names, constraint names, etc. — are limited to a
+        // maximum length of 63 bytes. Identifiers longer than 63 characters can be used, but they will be truncated
+        // to the allowed length of 63.
+        $sanitize_alias_length = function ($a, $b) {
+            if ($this->_driver instanceof Postgres) {
```
ndm2: It would probably be better to use the driver check to conditionally apply `array_intersect_ukey()`, so that the custom function is only applied for Postgres.

But I'm not sure if this is really appreciated by Postgres users in the first place, as the number of calls to `$sanitize_alias_length()` can easily go into the tens of thousands with larger results (10 fields with 10 matching keys results in 32 calls per row, and every additional field (for example associations) means 7 additional calls per row), that's quite some overhead that could easily be avoided by simply sticking to the Postgres limitations.

swiffer: I've added a new approach avoiding your concerns. thanks pointing this out!

Primarly stumpled upon this when converting a cakephp application from mariadb to postgresql.

How are you handling table and column names in case they get to long vs. cakephp naming conventions?

swiffer: to further limit the performance impact we can modify in `protected function _calculateColumnMap($query)` directly ?

src/ORM/ResultSet.php
```diff
@@ -410,7 +411,14 @@ protected function _calculateColumnMap($query)
             }
 
             $parts = explode('__', $key, 2);
-            $map[$parts[0]][$key] = $parts[1];
+            // In PostgreSQL, identifiers — table names, column names, constraint names, etc. — are limited to a
+            // maximum length of 63 bytes. Identifiers longer than 63 characters can be used, but they will be truncated
+            // to the allowed length of 63.
+            if ($this->_driver instanceof Postgres) {
+                $map[$parts[0]][substr($key, 0, 63)] = $parts[1];
```
swiffer: @ADmad / @ndm2 what do you think?

Primarly stumpled upon this when converting a cakephp application from mariadb to postgresql.

How are you handling table and column names in case they get to long vs. cakephp naming conventions?

othercorey:
What is your use case for this?  Have you run into an issue?

swiffer:
> 
> 
> What is your use case for this? Have you run into an issue?

yes, as some of my columns got longer than 63 charts as CakePHP adds the alias name to the column name mapping into entities was not possible / I was missing some properties / PHP was throwing warnings because length of arrays in array_combine were not equal in ResultSet.php:528

src/ORM/ResultSet.php
ndm2: > How are you handling table and column names in case they get to long vs. cakephp naming conventions?

Personally I'd just break with the conventions in that case, and provide the required/limited aliases/identifiers manually.

> to further limit the performance impact we can modify in `protected function _calculateColumnMap($query)` directly ?

I'm not sure if the shortening will work out in the first place, you'd need to add some tests, but I could imagine that this won't help when for example foreign keys don't match.

I'm also not sure if users would really want this behavior instead of being forced to configure the application and the schema accordingly. Personally I wouldn't feel overly confident leaving an application in such a fragile state.

othercorey:
> yes, as some of my columns got longer than 63 chars as CakePHP adds the alias name to the column name. mapping into entities was not possible / I was missing some properties / PHP was throwing warnings because length of arrays in array_combine were not equal in ResultSet.php:528

Is truncating the right solution here?  We run into the problem of identifiers that are different only after character 63 being treated the same.

Should we find a way to warn users and explain that their table + column names need to allow for cakephp's aliasing?

swiffer:
@othercorey it at least mimics the behavior of PostgreSQL in this case.

othercorey:
> @othercorey it at least mimics the behavior of PostgreSQL in this case.

I feel like that's hiding the mess.  Users interacting with postgres directly will see the results and find the field name changed or missing.  With cakephp, they won't see into that transformation.

I would prefer to either warn or find a way to optimize alias name length.

swiffer:
it's not about truncating field names - it's just accounting for columns where the identifier of a field (containing the cakephp table alias) is exceeding 63 chars. with the current behavior there is neither a direct warning by cakephp and the logik just fails as array_combine is not working anymore. with the slightly modified behaviour everything is fine as long as column names are still unique. and in case they aren't (as they get truncated to the same name) PostgreSQl will return a sql error anyway: https://www.postgresql.org/message-id/flat/20130110162334.GA24281%40momjian.us#729eb1521a1bf5ff71870284321cf7f9

so while perhaps it's still best practice to ensure identifiers are within 63 chars this is definitly an improved in many cases compared to the current behaviour and has no negative impact.

src/ORM/ResultSet.php
swiffer: why is it fragile? it's default postres behaviour and perfectly fine as long as identifiers are still unique. (see my comment below).

works fine and is not affecting foreign key matches or anything else. would be willing to add tests if appreciated.

ndm2: I'd call it "fragile" because the application is configured with keys that practically do not exist, IMHO that's not an overly reliable construct, there's so many places where things could go wrong.

Foreign keys were just one example, there's lots of other places where things might break, practically everywhere where conventions are used or the full hardcoded names. Forms for example, where the field names wouldn't match the identifiers in the entities. Or patching, where the marshaller will have shortened identifiers in the entities, and non-shortened identifiers in the request data. Or application rules that run tests against non-shortened identifiers.

swiffer: The property names of entities are not affected by this change.

ndm2: I know, I probably expressed myself a little vaguely, I didn't mean to say that it's your patch that will break these things, I mean that the shortening by Postgres can break these things, and that  this patch won't prevent that. While it circumvents the error in the result set, it will leave the application in this "fragile" state where lots of things can go wrong.

ADmad:
How is truncating the identifier as you have done even going to help? Let's assume `Foo_longish_title` is longer than 63 chars and your converted it to `Foo_long` after truncation.

So in the result set the field name will become `long` instead of `longish_title'`.  How is the rest of you app code going to know that the field name has changed? You are just going to create other problems as @ndm2 pointed out.

ADmad:
Similar issue #12282.

swiffer:
@ADmad I'm not truncating field names of the result set. Trying to explain:

Postgres:

Table Name:  a_long_but_valid_table_name_with_41_chars
Field Name: a_long_but_valid_field_name_with_41_chars
Second Field Name: an_even_longer_valid_field_name_with_45_chars

When building a Query in CakePHP the query will look something like this

```sql
SELECT
  a_long_but_valid_field_name_with_41_chars as ALongButValidTableNameWith41Chars__ALongButValidFieldNameWith41Chars,
  an_even_longer_valid_field_name_with_45_chars as ALongButValidTableNameWith41Chars__AnEvenLongerValidFieldNameWith45Chars
FROM a_long_but_valid_table_name_with_41_chars ALongButValidTableNameWith41Chars
```

PostgreSQL will return a result set with truncated field identifiers

- `ALongButValidTableNameWith41Chars__ALongButValidFieldNameWith41`
- `ALongButValidTableNameWith41Chars__AnEvenLongerValidFieldNameWi`

currently CakePHP tries to map these columns in ResultSet.php:L521/L536 via

```php
$results[$table] = array_combine($keys, array_intersect_key($row, $keys));
```
as the array keys for fields with identifiers longer than 63 chars have been truncated array_intersect_key is not working anymore causing array_combine to fail (array count is not equal).

currently cakephp assumes column names are unchanged when returned from the database server. as this is not the case i modified the column names in the relevant $_map.

```php
    /**
     * Map of fields that are fetched from the statement with
     * their type and the table they belong to
     *
     * @var array
     */
    protected $_map = [];

    ...

    // In PostgreSQL, identifiers — table names, column names, constraint names, etc. — are limited to a
    // maximum length of 63 bytes. Identifiers longer than 63 characters can be used, but they will be truncated
    // to the allowed length of 63.
    if ($this->_driver instanceof Postgres) {
        $map[$parts[0]][substr($key, 0, 63)] = $parts[1];
    } else {
        $map[$parts[0]][$key] = $parts[1];
    }
```

ADmad:
Okay, even if your changes don't affect the field names in the generated result your fix seems pretty hackish and driver specific code doesn't belong in the `ResultSet`.

The proper way to handle this would be to create a  postgres specific query compiler and the identifiers truncated before hand or simply throw an excepting if identifier length exceeds limit.

othercorey:
> The proper way to handle this would be to create a postgres specific query compiler and the identifiers truncated before hand or simply throw an exception if identifier length exceeds limit.

As mentioned before, I would vote for warning the user or finding a way to optimize generated alias names other than truncating.

ADmad:
> As mentioned before, I would vote for warning the user or finding a way to optimize generated alias names other than truncating.

I agree, simply making the dev aware of the potential truncation should suffice and let the dev use custom shorter aliases.

othercorey:
> let the dev use custom shorter aliases.

Maybe there is a discussion for generating unique, short hashes to use instead for people who want long identifier support.

swiffer:
Fully understood this should not go into ResultSet as it's driver specific. However I think it's not possible to modify this anywhere else from my current understanding of how _calculateColumnMap works. Truncating the field names before would indeed result into truncated entity field names, wouldn't it?

swiffer:
@markstory moved the logic to the driver so it's encapsulated for postgres

src/Database/Driver/Postgres.php
```diff
@@ -129,4 +129,29 @@ public function supportsDynamicConstraints()
...
         return true;
     }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function calculateColumnMap(Query $query)
+    {
+        $map = [];
+        $defaultAlias = $query->getRepository()->getAlias();
```
ADmad: `Cake\Database\Query` doesn't have a `getRepository()` method and since the driver is in database package it can't depend on `Cake\ORM\Query`.

In general truncating the identifiers after they have been already set in query is the wrong approach. The length check should be done when the identifer/alias are being generated/set.

`Cake\ORM\Query::aliasField()` is where the `Foo_bar` style identifiers are generated. So there a check using the driver should be done to validate the identifier lengh.

swiffer: Hey @ADmad thanks reviewing again. as you've asked I've updated to the pull request again.
Would you check again if this is going in the right direction (truncate when set via `aliasField`) modified `_calculateColumnMap()` to not rely on query field alias name for mapping to entity property.

It's not checking for the driver yet. can you give me a hint on how to get from `Cake\Datasource\QueryTrait` ? I was not able to locate `Cake\ORM\Query::aliasField()`

swiffer: ok, truncating in when building the query heavily affects test suite as Cake relies on these array keys! i would still go the route @markstory suggested and I've tried to approach previously.

src/Database/Driver.php
```diff
@@ -446,4 +446,23 @@ public function __debugInfo()
...
     }
+
+    /**
+     * Truncates an identifier if the database does
+     *
+     * @param string $key Aliased field name
+     *
+     * @return string Aliased field name as returned by Database
+     */
+    public function sanitizeIdentifier(string $key)
```
markstory: This name is misleading. There isn't any sanitizing being done, could we call this `generateIdentifier()`?

src/Database/Driver.php
```diff
@@ -446,4 +446,23 @@ public function __debugInfo()
...
+    {
+        // In PostgreSQL, identifiers — table names, column names, constraint names, etc. — are limited to a
+        // maximum length of 63 bytes. Identifiers longer than 63 characters can be used, but they will be truncated
+        // to the allowed length of 63.
+        if ($this->_connection->getAttribute(PDO::ATTR_DRIVER_NAME) === 'pgsql') {
```
markstory: Instead of a conditional in the base class, the base implementation should only return `$key` while the Postgres driver can apply this logic with no condition.

swiffer: sure, done!

src/Database/Driver.php
swiffer: yes, done as well

swiffer:
Closing this one as per #14086. I'll keep overloading the class per my PR until temporary alias names are thing. Thanks for reviewing this, overhead is minimal now and looks good in specific project.

PR #13832, opened 2019-11-02
Simplify PHPStan installation

szepeviktor:
~~There is no dev version in phpstan-shim repo.~~


dereuromark:
Why is this simpler?

szepeviktor:
There is no back-and-forth in it.

szepeviktor:
I've just noticed that 5 days ago phpstan-shim got a dependency: https://github.com/phpstan/phpstan-shim/blob/master/composer.json#L7

PR #13833, opened 2019-11-03
Avoid unnecessary reflection

ADmad:


garas:
Abstract/interface controllers will now respond with error 500 instead of 404.

dereuromark:
I agree, the 2nd iteration (removing MissingControllerException wrapping) was a bit too much here.

src/Controller/ControllerFactory.php
```diff
@@ -38,24 +37,19 @@ class ControllerFactory implements ControllerFactoryInterface
...
     public function create(ServerRequestInterface $request): Controller
     {
         $className = $this->getControllerClass($request);
-        if (!$className) {
+        if ($className === null) {
             $this->missingController($request);
         }
-        /** @var string $className */
-        $reflection = new ReflectionClass($className);
-        if ($reflection->isAbstract() || $reflection->isInterface()) {
```
markstory: As others mentioned this will cause 500s when the framework used to emit 404, this will generally only impact applications that use fallback routing and abstract/interfaces in their controllers.

Does the usage or reflection have any runtime impact? I know that in early PHP5 Reflection could be a performance problem, but I'm not sure that holds true today.

ADmad:
I have added try/catch to ensure 404 is returned as before.

src/Controller/ControllerFactory.php
dereuromark: Controller reflection check is IMO never a performance impact, as this is always only a singular PHP invocation per request lifecycle. I think this is in the nano-area of relevance.

The main idea of this change was probably to have less complicated code. And somewhere in the middle we can probably find a good solution here.

src/Controller/ControllerFactory.php
```diff
@@ -38,23 +38,23 @@ class ControllerFactory implements ControllerFactoryInterface
...
+
+        $controller = null;
+        try {
+            /** @psalm-suppress UndefinedClass */
+            $controller = new $className($request);
+        } catch (Error $e) {
             $this->missingController($request);
```
dereuromark: Before, you had a 2nd argument to rethrow $e into the chain. Might be worth resurrecting that change from before.

ADmad: Can't you just pretend you didn't see that commit? :stuck_out_tongue:

garas: `Error` can be thrown not only because controller is not instantiable. Many things can happen in many parts of constructor.

I would stay with reflection. Only maybe simplify to `ReflectionClass::isInstantiable()`.

dereuromark: @garas One could argue that once the controller is instantiable, that anything else that goes wrong warrants an internal error 500, since then there is a real issue.
But `ReflectionClass::isInstantiable()` sounds also like a good idea.

ADmad:
I have reverted all changes and just removed the unneeded `isInterface()` check. Since `App::className()` always does `class_exists()` check it can never return a interface name.

PR #13839, opened 2019-11-05
4.x - Cookie defaults

ADmad:
Currently it was a bit tedious having to specify common non default options/attributes when creating cookies.

For e.g. if your app is under a subfolder of webserver's document root a sensible default is to have the path for all cookies set to `/subfolder` instead of `/` (that's what `CookieComponent` did). But in 4.x you would have to specify the subfolder path each time when creating a new cookie.

So I have added a new static method `Cookie::setDefaults()` to avoid having to repeatedly pass common options.

### DIFF
--- a/src/Http/Cookie/Cookie.php
+++ b/src/Http/Cookie/Cookie.php
@@ -197,8 +197,14 @@ class Cookie implements CookieInterface
      */
     public static function setDefaults(array $options): void
     {
+        if (isset($options['expires'])) {
+            $options['expires'] = static::dateTimeInstance($options['expires']);
+        }
+        if (isset($options['samesite'])) {
+            static::validateSameSiteValue($options['samesite']);
+        }
+
         static::$defaults = $options + static::$defaults;
-        static::$defaults['expires'] = static::dateTimeInstance(static::$defaults['expires']);
     }
 
     /**
@@ -242,6 +248,13 @@ class Cookie implements CookieInterface
         if ($expires instanceof DateTimeInterface) {
             /** @psalm-suppress UndefinedInterfaceMethod */
             return $expires->setTimezone(new DateTimeZone('GMT'));
+        }
+
+        if (!is_string($expires) && !is_int($expires)) {
+            throw new InvalidArgumentException(sprintf(
+                'Invalid type `%s` for expires.',
+                getTypeName($expires)
+            ));
         }
 
         if (!is_numeric($expires)) {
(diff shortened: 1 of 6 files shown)


src/Http/Cookie/Cookie.php
```diff
@@ -160,6 +176,17 @@ public function __construct(
...
 
+    /**
+     * Set default options for the cookies.
+     *
+     * @param array $options Default options.
+     * @return void
+     */
+    public static function setDefaults(array $options): void
+    {
+        static::$defaults = $options + static::$defaults;
```
othercorey: If you set defaults twice with a sparse array, do you expect it to reset the other fields or merge with previously set defaults?

src/Http/Cookie/Cookie.php
```diff
@@ -160,6 +176,17 @@ public function __construct(
         $this->expiresAt = $expiresAt;
     }
 
+    /**
+     * Set default options for the cookies.
```
othercorey: Is there a docblock explaining the allowed fields?

tests/TestCase/Http/Cookie/CookieTest.php
```diff
@@ -510,4 +511,17 @@ public function testCreateFromHeaderString()
         // https://tools.ietf.org/html/draft-west-first-party-cookies-07#section-4.1
         $this->assertNull($result->getSameSite());
     }
+
+    public function testDefaults()
+    {
+        Cookie::setDefaults(['path' => '/cakephp', 'expires' => time()]);
```
markstory: We should have a test for invalid same site and expires values.

ADmad: Convenience of array means losing benefit of type hinting :sob:. I need array shapes :smile:.

ADmad: I added checks and tests :)

src/Http/Cookie/Cookie.php
```diff
@@ -206,13 +233,50 @@ public static function create(string $name, $value, array $options = [])
...
+        }
+
+        if ($expires instanceof DateTimeInterface) {
+            /** @psalm-suppress UndefinedInterfaceMethod */
+            return $expires->setTimezone(new DateTimeZone('GMT'));
+        }
+
+        if (!is_string($expires) && !is_int($expires)) {
+            throw new InvalidArgumentException(sprintf(
+                'Invalid type `%s` for expires.',
```
markstory:
```suggestion
                'Invalid type `%s` for expires. Expected an string, integer or DateTime object.',
```

markstory (approved):

Issue #13681, opened 2019-09-26
On Hash::combine() grouping allow keyPath be empty in order to use numbered array


SpartakusMd:
In case there is no unique field to index by the data when using Hash::combine(), I still want to have a way to extract the grouped data. For this case I propose to send `$keyPath = null` so we don't use a field but create numbered arrays.

### Issue type:

* [ ] bug
* [x] enhancement
* [ ] feature-discussion (RFC)

### Component
`Hash` from https://github.com/cakephp/utility

### Example

```
$a = [
    [
        'User' => [
            'group_id' => 1,
            'Data' => [
                'user' => 'mariano.iglesias',
                'name' => 'Mariano Iglesias'
            ]
        ]
    ],
    [
        'User' => [
            'group_id' => 1,
            'Data' => [
                'user' => 'phpnut',
                'name' => 'Larry E. Masters'
            ]
        ]
    ],
];

$result = Hash::combine($a, null, '{n}.User.Data', '{n}.User.group_id');
```

Expected `$result`:

```
$result = [
        [1] => [
                [0] => [ // PHP arrays start from index 0
                        [user] => mariano.iglesias
                        [name] => Mariano Iglesias
                ],
                [1] => [
                        [user] => phpnut
                        [name] => Larry E. Masters
                ]
        ]
    ]
```

dereuromark:
PR open.

PR #13848, opened 2019-11-06
 #13681 - Allow null $keyPath on Hash::combine and use numbered array if it is

victoreassis:
This should solve issue #13681 

It's my first contribution though, so i don't know if it's a perfect one.

### DIFF
--- a/src/Utility/Hash.php
+++ b/src/Utility/Hash.php
@@ -466,12 +466,12 @@ class Hash
         if (is_array($keyPath)) {
             $format = array_shift($keyPath);
             $keys = static::format($data, $keyPath, $format);
-        } elseif (is_null($keyPath)) {
+        } elseif ($keyPath === null) {
             $keys = $keyPath;
         } else {
             $keys = static::extract($data, $keyPath);
         }
-        if (!is_null($keyPath) && empty($keys)) {
+        if ($keyPath !== null && empty($keys)) {
             return [];
         }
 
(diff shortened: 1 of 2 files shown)


src/Utility/Hash.php
```diff
@@ -466,10 +466,12 @@ public static function combine(array $data, $keyPath, $valuePath = null, $groupP
         if (is_array($keyPath)) {
             $format = array_shift($keyPath);
             $keys = static::format($data, $keyPath, $format);
+        } elseif (is_null($keyPath)) {
```
dereuromark: `$keyPath === null`
and others

dereuromark:
You might want to switch the target to 3.next here (master is 3.current = 3.8 and for bugfixes only).

src/Utility/Hash.php
victoreassis: Ok. I will do that.

victoreassis:
Got it.

dereuromark:
I managed to do it for you

victoreassis:
Thanks

src/Utility/Hash.php
victoreassis: Done.

dereuromark:
This change requires some test cases in HashTest class.

dereuromark:
Please note: Your Git credentials is not set up correctly locally. You are commiting and pushing here as anonymous user.

victoreassis:
Ok. I will check the HashTest and the credentials.

victoreassis:
I can't tell what exactly happened on the check that failed. It appears to have passed tests but failed anyway.

dereuromark:
Great work!
The appv build can be ignored for now.

victoreassis:
Thanks.

I have a question, the method description shouldn't be updated to include array as a possible value of $keyPath and $valuePath? As they can receive one to format them.

markstory:
Thank you 👍

andrii-pukhalevych:
Does it preserve keys from source array or just generates new numeric indexes?

victoreassis:
It generates new ones.

PR #13850, opened 2019-11-06
Update Form.php

andrii-pukhalevych:
public function Form::set() - Saves a variable or an associative array of variables for use inside form data

Allows to set single element inside form data, similar to Session::write() or View::set()

### DIFF
--- a/src/Form/Form.php
+++ b/src/Form/Form.php
@@ -271,6 +271,28 @@ class Form implements EventListenerInterface, EventDispatcherInterface, Validato
     }
 
     /**
+     * Saves a variable or an associative array of variables for use inside form data.
+     *
+     * @param string|array $name A string or an array of data.
+     * @param mixed $value Value in case $name is a string (which then works as the key).
+     *   Unused if $name is an associative array, otherwise serves as the values to $name's keys.
+     * @return $this
+     */
+	public function set($name, $value = null) 
+	{
+        $write = $name;
+        if (!is_array($name)) {
+            $write = [$name => $value];
+        }
+
+        foreach ($write as $key => $val) {
+            $this->_data = Hash::insert($this->_data, $key, $val);
+        }
+
+		return $this;
+	}
+
+    /**
      * Set form data.
      *
      * @param array $data Data array.


andrii-pukhalevych:
Please create tests if PR is acceptable. I didnt work with phpunit

ADmad:
What's your use case which requires setting individual values?  Can't you just aggregate all data beforehand and use `Form::setData()`?

andrii-pukhalevych:
In my form default values depends on user type, so code looks more clean when data items set separately, without aggregating data into additional variable.

ADmad:
You can still have separate variables for common and user specific data and then merge only when making th `setData()` call.

andrii-pukhalevych:
Its more clean to write shorter and without additional variable, like using Controller::set()
```
$Form->setData(['var1'=>'val1']);
if ($condition)
   $Form->set('var2', 'val2');
```

Instead of this:
```
$data = ['var1'=>'val1'];
if ($condition)
    $data['var2'] = 'val2';
$Form->setData($data);
```

src/Form/Form.php
```diff
@@ -270,6 +270,28 @@ public function getData(?string $field = null)
...
 
+    /**
+     * Saves a variable or an associative array of variables for use inside form data.
+     *
+     * @param string|array $name A string or an array of data.
+     * @param mixed $value Value in case $name is a string (which then works as the key).
+     *   Unused if $name is an associative array, otherwise serves as the values to $name's keys.
+     * @return $this
+     */
+	public function set($name, $value = null)
```
markstory: If we're going to add a new method, could it only add a single key? We already have `setData()` to set all the keys.

andrii-pukhalevych: `setData()` overrides all existing data. `set()` appends new data using `Hash::insert()`.

		$form->set('var1', 'val1'); //array ( 'var1' => 'val1', )
		$form->set(['var2'=>'val2', 'var3'=>'val3']); //array ( 'var1' => 'val1', 'var2' => 'val2', 'var3' => 'val3', )
		$form->set(['var1'=>'val1_new']); //array ( 'var1' => 'val1_new', 'var2' => 'val2', 'var3' => 'val3', )
		$form->setData(['var1'=>'val1_new2']); //array ( 'var1' => 'val1_new2', )

markstory: Ok. Are you aiming to replicate the behavior of `ViewVarsTrait::set()` on Form?

andrii-pukhalevych: Yes. Actually I want to replicate `EntityTrait::set()` first, and `EntityTrait::setError()` in next PR.
Both are used as forms context, but set() and setError() works for entity only.

markstory: >  Both are used as forms context, but set() and setError() works for entity only.

I don't see `set()` and `setError()` being used in the `FormContext`?

andrii-pukhalevych: I mean that entities and forms are used as form context. Not methods.

andrii-pukhalevych:
@markstory Did you completely decline this PR?

saeideng:
@andrii-pukhalevych branch `4.x` has been deleted so PR(s) on it closed automatically 
current branch : 
`master` => for version 4.0.x
`4.next` =>for 4.1.x

saeideng:
you can open a new Issue to discussion

markstory:
I didn't mean for this to get closed, I was removing the now stale 4.x branch. If you could open the pull request against master that would be appreciated as I cannot change the base branch now.

PR #13857, opened 2019-11-07
Fix route name generation when using braced placeholders.

ADmad:
Earlier reverse routing failed when using placeholder like {controller} or {action}
in URL template since automatic route names are generated based on such special keys
and are used by RouteCollection for matching.

This probably need to be backported to 3.x too. I made the patch for 4.x as that's what I was working on when I discovered the bug and was unsure if it affected 3.x too.


ADmad:
Also updated the fallback route templates to use new braced placeholders.

PR #13861, opened 2019-11-08
Encode Content-Disposition header fields

### LINKED ISSUE #13854: Email attachment filenames are not properly encoded
This is a (multiple allowed):

* [x] bug
* [ ] enhancement

* CakePHP Version: 3.8.5

### What you did
Send an email with an attachment using `Email::addAttachments()` and `Email::send()`

### What happened
The attachment's filename set in the `Content-Disposition` header parameter does not get properly encoded to contain ASCII-only characters.  
The filename string will just be wrapped in double quotes without an respect to the charset used within the string. Thus non-ASCII characters do not get escaped or re-encoded.  
This **breaks mail attachments** with some providers if the attachment filenames contain e.g. German umlauts.

Example header set by CakePHP:
```
Content-Disposition: attachment; filename="File_with_Ümlaut.pdf"
Content-Type: application/pdf
Content-Transfer-Encoding: base64
```

The string is percent-encoded and the charset of the header field is set according to [RFC 2231](https://tools.ietf.org/html/rfc2231) if there are non-ASCII characters in the filename.

Example from above according to RFC 2231:
```
Content-Type: application/pdf
Content-Disposition: attachment; filename*=utf-8''File_with_%C3%9Cmlaut.pdf
Content-Transfer-Encoding: base64
```

### Proposed Fix
My proposed fix would be to always set the encoding for parameter in the `Content-Disposition` header in `FormDataPart::__toString()`

Something alike:
```
if ($this->_disposition) {
    $out .= 'Content-Disposition: ' . $this->_disposition;
    if ($this->_name) {
        $out .= '; name*=' . $this->_charset . '\'\'' . rawurlencode($this->_name);
    }
    if ($this->_filename) {
        $out .= '; filename*=' . $this->_charset . '\'\'' . rawurlencode($this->_filename);
    }
    $out .= "\r\n";
}
```
This would also be in line with [RFC 8187](https://tools.ietf.org/html/rfc8187) for indicating character encodings in HTTP header fields. Though said RFC or its predecessor RFC 5987 should be supported in all current major browsers, I don't know about more exotic or antique browsers.

mrtnmtth:
Set character encoding for Content-Disposition header parameters in line with RFC 2231 for email and RFC 8187 for HTTP.

This PR fixes #13854

I have not adapted any test cases yet, because I wanted to wait for feedback on my changes first.

### DIFF
--- a/src/Http/Client/FormDataPart.php
+++ b/src/Http/Client/FormDataPart.php
@@ -13,6 +13,9 @@
  */
 namespace Cake\Http\Client;
 
+use Cake\Utility\Inflector;
+use Cake\Utility\Text;
+
 /**
  * Contains the data and behavior for a single
  * part in a Multipart FormData request body.
@@ -74,17 +77,26 @@ class FormDataPart
     protected $_contentId;
 
     /**
+     * The charset attribute for the Content-Disposition header fields
+     *
+     * @var string|null
+     */
+    protected $_charset;
+
+    /**
      * Constructor
      *
      * @param string $name The name of the data.
      * @param string $value The value of the data.
      * @param string $disposition The type of disposition to use, defaults to form-data.
+     * @param string|null $charset The charset of the data.
      */
-    public function __construct($name, $value, $disposition = 'form-data')
+    public function __construct($name, $value, $disposition = 'form-data', $charset = null)
     {
         $this->_name = $name;
         $this->_value = $value;
         $this->_disposition = $disposition;
+        $this->_charset = $charset;
     }
 
     /**
@@ -198,10 +210,10 @@ class FormDataPart
         if ($this->_disposition) {
             $out .= 'Content-Disposition: ' . $this->_disposition;
             if ($this->_name) {
-                $out .= '; name="' . $this->_name . '"';
+                $out .= '; ' . $this->_headerParameterToString('name', $this->_name);
             }
             if ($this->_filename) {
-                $out .= '; filename="' . $this->_filename . '"';
+                $out .= '; ' . $this->_headerParameterToString('filename', $this->_filename);
             }
             $out .= "\r\n";
         }
@@ -218,6 +230,27 @@ class FormDataPart
         $out .= (string)$this->_value;
 
         return $out;
+    }
+
+    /**
+     * Get the string for the header parameter.
+     *
+     * If the value contains non-ASCII letters an additional header indicating
+     * the charset encoding will be set.
+     *
+     * @param string $name The name of the header parameter
+     * @param string $value The value of the header parameter
+     * @return string
+     */
+    protected function _headerParameterToString($name, $value)
+    {
+        $transliterated = Text::transliterate(str_replace('"', '', $value));
+        $return = sprintf('%s="%s"', $name, $transliterated);
+        if ($this->_charset !== null && $value !== $transliterated) {
+            $return .= sprintf("; %s*=%s''%s", $name, strtolower($this->_charset), rawurlencode($value));
+        }
+
+        return $return;
     }
 }
 
--- a/src/Mailer/Email.php
+++ b/src/Mailer/Email.php
@@ -2521,7 +2521,7 @@ class Email implements JsonSerializable, Serializable
                 !isset($fileInfo['contentDisposition']) ||
                 $fileInfo['contentDisposition']
             );
-            $part = new FormDataPart(false, $data, false);
+            $part = new FormDataPart('', $data, '', $this->getHeaderCharset());
 
             if ($hasDisposition) {
                 $part->disposition('attachment');
@@ -2571,7 +2571,7 @@ class Email implements JsonSerializable, Serializable
             $data = isset($fileInfo['data']) ? $fileInfo['data'] : $this->_readFile($fileInfo['file']);
 
             $msg[] = '--' . $boundary;
-            $part = new FormDataPart(false, $data, 'inline');
+            $part = new FormDataPart('', $data, 'inline', $this->getHeaderCharset());
             $part->type($fileInfo['mimetype']);
             $part->transferEncoding('base64');
             $part->contentId($fileInfo['contentId']);
--- a/tests/TestCase/Mailer/EmailTest.php
+++ b/tests/TestCase/Mailer/EmailTest.php
@@ -1589,6 +1589,63 @@ class EmailTest extends TestCase
     }
 
     /**
+     * Test an attachment filename with non-ASCII characters.
+     *
+     * @return void
+     */
+    public function testSendWithNonAsciiFilenameAttachments()
+    {
+        $this->Email->setTransport('debug');
+        $this->Email->setFrom('cake@cakephp.org');
+        $this->Email->setTo('cake@cakephp.org');
+        $this->Email->setSubject('My title');
+        $this->Email->setEmailFormat('both');
+        $this->Email->setAttachments([
+            'gâteau.png' => [
+                'file' => CORE_PATH . 'VERSION.txt',
+                'contentId' => 'abc123'
+            ]
+        ]);
+        $result = $this->Email->send('Hello');
+
+        $boundary = $this->Email->getBoundary();
+        $this->assertContains('Content-Type: multipart/mixed; boundary="' . $boundary . '"', $result['headers']);
+        $expected = "--$boundary\r\n" .
+            "Content-Type: multipart/related; boundary=\"rel-$boundary\"\r\n" .
+            "\r\n" .
+            "--rel-$boundary\r\n" .
+            "Content-Type: multipart/alternative; boundary=\"alt-$boundary\"\r\n" .
+            "\r\n" .
+            "--alt-$boundary\r\n" .
+            "Content-Type: text/plain; charset=UTF-8\r\n" .
+            "Content-Transfer-Encoding: 8bit\r\n" .
+            "\r\n" .
+            'Hello' .
+            "\r\n" .
+            "\r\n" .
+            "\r\n" .
+            "--alt-$boundary\r\n" .
+            "Content-Type: text/html; charset=UTF-8\r\n" .
+            "Content-Transfer-Encoding: 8bit\r\n" .
+            "\r\n" .
+            'Hello' .
+            "\r\n" .
+            "\r\n" .
+            "\r\n" .
+            "--alt-{$boundary}--\r\n" .
+            "\r\n" .
+            "--rel-$boundary\r\n" .
+            "Content-Disposition: inline; filename=\"gateau.png\"; filename*=utf-8''g%C3%A2teau.png\r\n" .
+            "Content-Type: text/plain\r\n" .
+            "Content-Transfer-Encoding: base64\r\n" .
+            "Content-ID: <abc123>\r\n" .
+            "\r\n";
+        $this->assertContains($expected, $result['message']);
+        $this->assertContains('--rel-' . $boundary . '--', $result['message']);
+        $this->assertContains('--' . $boundary . '--', $result['message']);
+    }
+
+    /**
      * testSendWithLog method
      *
      * @return void


mrtnmtth:
This PR does always set an encoding. An alternative would be to set the charset in the constructor and only set the encoding in the `Content-Disposition` header if a charset is set.

ADmad:
> An alternative would be to set the charset in the constructor and only set the encoding in the Content-Disposition header if a charset is set.

Passing the charset through constructor would be a better solution. Then the appropriate value of  `getContentTypeCharset()` (or `getCharset()`  or `getHeaderCharset()`) of `Email` class can be passed to it.

ndm2:
Is that actually supported anywhere other than in E-Mail clients and in HTTP _responses_ for browsers? Specifically, is this supported in HTTP _requests_ in general, and especially by webservers and languages like PHP, Java, Perl, etc? If there is no wide support for requests as well as reponses, then this should definitely be optional.

And could there be clients that support encoded filenames (`filename`), but not encoded fieldnames (`name`), ie in multipart requests?

Also should there maybe be a fallback for clients that don't support the charset syntax at all?

```
Content-Disposition: attachment;
                     filename="ASCII fallback";
                     filename*=utf-8''Encoded-Variant
```

mrtnmtth:
> And could there be clients that support encoded filenames (`filename`), but not encoded fieldnames (`name`), ie in multipart requests?

The RFCs are about non-ASCII encoded strings in header fields in general. Therefore this should not be the case.


> Also should there maybe be a fallback for clients that don't support the charset syntax at all?

A quick search on the topic lead me to [RCF 6266, sec 4.3](https://tools.ietf.org/html/rfc6266#section-4.3):
>    Many user agent implementations predating this specification do not
>   understand the "filename*" parameter.  Therefore, when both
>   "filename" and "filename*" are present in a single header field
>   value, recipients SHOULD pick "filename*" and ignore "filename".

ndm2:
> > And could there be clients that support encoded filenames (`filename`), but not encoded fieldnames (`name`), ie in multipart requests?
> 
> The RFCs are about non-ASCII encoded strings in header fields in general. Therefore this should not be the case.

RFCs are one thing, actual implementations are often something very different. Stuff like this unfortunately usually needs to adapt to the dirty implementations, not the specifications.

A quick check with PHP shows that it doesn't seem to support encoded parameter values in requests at all.

> > Also should there maybe be a fallback for clients that don't support the charset syntax at all?
> 
> A quick search on the topic lead me to [RCF 6266, sec 4.3](https://tools.ietf.org/html/rfc6266#section-4.3):

So, that's a yes? :upside_down_face:  Not sure what you're trying to say?

mrtnmtth:
> So, that's a yes? upside_down_face Not sure what you're trying to say?

It's a yes. I will add the fallback to my patch.

src/Http/Client/FormDataPart.php
```diff
@@ -198,10 +217,16 @@ public function __toString()
...
             if ($this->_name) {
-                $out .= '; name="' . $this->_name . '"';
+                $out .= '; name="' . preg_replace('/[^\x20-\x7E]/', '', $this->_name) . '"';
+                if ($this->_charset !== null) {
+                    $out .= '; name*=' . $this->_charset . '\'\'' . rawurlencode($this->_name);
+                }
             }
             if ($this->_filename) {
-                $out .= '; filename="' . $this->_filename . '"';
+                $out .= '; filename="' . preg_replace('/[^\x20-\x7E]/', '', $this->_filename) . '"';
```
garas: `name` and `filename` blocks looks same, add separate method to deduplicate.

It could also check if string contains any non-ASCII, and only then add `*=` value so in most cases it would leave as it was with simple `=` value.

src/Http/Client/FormDataPart.php
```diff
@@ -73,18 +73,27 @@ class FormDataPart
...
+     */
+    protected $_charset;
+
     /**
      * Constructor
      *
      * @param string $name The name of the data.
      * @param string $value The value of the data.
      * @param string $disposition The type of disposition to use, defaults to form-data.
+     * @param string $charset The charset of the data.
```
ADmad: In all docblocks it should be `string|null`.

src/Http/Client/FormDataPart.php
```diff
@@ -185,6 +194,16 @@ public function value()
         return $this->_value;
     }
 
+    /**
+     * Get the charset.
+     *
+     * @return string
+     */
+    public function charset()
```
ADmad: What's the use case for this public method?

mrtnmtth: None. Got removed.

src/Http/Client/FormDataPart.php
```diff
@@ -219,6 +228,26 @@ public function __toString()
...
+     *
+     * @param string $name The name of the header parameter
+     * @param string $value The value of the header parameter
+     * @return string
+     */
+    protected function _headerParameterToString($name, $value)
+    {
+        $return = $name . '="' . preg_replace('/[^\x20-\x7E]/', '', $value) . '"';
+        if (!mb_check_encoding($value, 'ASCII') && $this->_charset != null) {
+            $return .= '; ' . $name . '*=' . $this->_charset . '\'\'' . rawurlencode($this->_filename);
```
garas: Use double quotes to avoid `\'\'`. Use `sprintf` as there is too much concatenation.

Replace `$this->_filename` to `$value`.

src/Http/Client/FormDataPart.php
```diff
@@ -219,6 +228,26 @@ public function __toString()
...
+     * the charset encoding will be set.
+     *
+     * @param string $name The name of the header parameter
+     * @param string $value The value of the header parameter
+     * @return string
+     */
+    protected function _headerParameterToString($name, $value)
+    {
+        $return = $name . '="' . preg_replace('/[^\x20-\x7E]/', '', $value) . '"';
+        if (!mb_check_encoding($value, 'ASCII') && $this->_charset != null) {
```
garas: Same regexp would work. `"` can be added to regexp as it probably not allowed in quotes.

src/Http/Client/FormDataPart.php
```diff
@@ -73,18 +73,27 @@ class FormDataPart
...
     /**
      * Constructor
      *
      * @param string $name The name of the data.
      * @param string $value The value of the data.
      * @param string $disposition The type of disposition to use, defaults to form-data.
+     * @param string|null $charset The charset of the data.
      */
-    public function __construct($name, $value, $disposition = 'form-data')
+    public function __construct($name, $value, $disposition = 'form-data', $charset = null)
```
ndm2: > The charset of the data

Could that possibly be a little misleading/confusing? One could think that this refers to the charset of the data part, ie the `charset` parameter for the `Content-Type` header, not to the possible global/default charset and that it affects the `Content-Disposition` header.

src/Http/Client/FormDataPart.php
```diff
@@ -219,6 +228,28 @@ public function __toString()
...
+     * @param string $name The name of the header parameter
+     * @param string $value The value of the header parameter
+     * @return string
+     */
+    protected function _headerParameterToString($name, $value)
+    {
+        // This regex matches non-ASCII characters and double quotes
+        $pattern = '/[^\x20-\x21|\x23-\x7E]/';
+        $return = sprintf('%s="%s"', $name, preg_replace($pattern, '', $value));
+        if (preg_match($pattern, $value) && $this->_charset != null) {
```
ADmad: Use `!== null` as suggested by Scrutinizer, also move the null check first as it's cheaper than preg match.

src/Mailer/Email.php
```diff
@@ -2521,7 +2521,7 @@ protected function _attachFiles($boundary = null)
                 !isset($fileInfo['contentDisposition']) ||
                 $fileInfo['contentDisposition']
             );
-            $part = new FormDataPart(false, $data, false);
+            $part = new FormDataPart(false, $data, false, $this->getHeaderCharset());
```
ADmad: The 1st argument of `FormDataPart::__construct()` is string, you can't pass it `false`.

ADmad: Same for the 3rd argument. Use empty string instead for both.

mrtnmtth:
Concerning test cases, should I write a new one or just adapt an existing one?

src/Http/Client/FormDataPart.php
garas: `headerCharset` would match naming in `Email` and reduce misleading

tests/TestCase/Mailer/EmailTest.php
```diff
@@ -1588,6 +1588,63 @@ public function testSendWithNoContentDispositionAttachments()
...
+            "Content-Transfer-Encoding: 8bit\r\n" .
+            "\r\n" .
+            'Hello' .
+            "\r\n" .
+            "\r\n" .
+            "\r\n" .
+            "--alt-{$boundary}--\r\n" .
+            "\r\n" .
+            "--rel-$boundary\r\n" .
+            "Content-Disposition: inline; filename=\"gteau.png\"; filename*=utf-8''g%C3%A2teau.png\r\n" .
```
markstory: Is having the filename defined twice correct? I couldn't find any examples in the RFC that included the same parameter multiple times.

ADmad: There's a subtle difference `filename` vs `filename*`.

garas: When both "filename" and "filename*" are present in a single header field value, "filename*" is preferred over "filename" when both are understood.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Disposition

markstory: @garas Yes in http headers. Is that also true in multipart email messages?

jimbo2150: According to NodeMailer, [yes](https://blog.nodemailer.com/2017/01/27/the-mess-that-is-attachment-filenames/). They indicate that the filename parameter (without the asterisk) cannot be encoded and should only include Latin characters (any other characters should be stripped or trans-coded to Latin equivalent). The filename* parameter (with the asterisk) is supposed to be URL parameter encoded to allow for non-Latin characters -- though not all clients may recognize or even use either of the filename fields. If you really want to support older clients you would also need to include the "name" parameter encoded (using encoded-word) as part of the Content-Type header (e.g. the pre-1997 standard).

src/Http/Client/FormDataPart.php
ndm2: That's kinda what I had in mind, and the parameter should have a description that reflects that.

tests/TestCase/Mailer/EmailTest.php
ndm2: To be fair, it doesn't say anything about fallbacks, and [it seems to generate only the encoded variant for unicode filenames](https://github.com/nodemailer/nodemailer/blob/v6.3.1/test/mime-node/mime-node-test.js#L591-L604). Looking at for example Mozilla's stuff (ie Thunderbird, etc.) [it seems to support it](https://dxr.mozilla.org/comm-central/rev/2a29ee0adb310b54a6a2df72034953fed8f2b043/comm/mailnews/mime/jsmime/test/test_header.js#45), but that doesn't mean too much of course.

Theoretically unrecognized parameters should be ignored according to https://tools.ietf.org/html/rfc2183#section-2.8, but I couldn't find anything specific about fallbacks for encoded parameter values in general. But as stated earlier, you can't necessarily rely on RFCs anyways, so this probably either needs some real world testing, or the fallback needs to go and people need to make their filenames ASCII-only in beforehand if they need maximum compatibility and the "old" parameter style.

src/Http/Client/FormDataPart.php
```diff
@@ -219,6 +228,28 @@ public function __toString()
...
+     * the charset encoding will be set.
+     *
+     * @param string $name The name of the header parameter
+     * @param string $value The value of the header parameter
+     * @return string
+     */
+    protected function _headerParameterToString($name, $value)
+    {
+        // This regex matches non-ASCII characters and double quotes
+        $pattern = '/[^\x20-\x21|\x23-\x7E]/';
```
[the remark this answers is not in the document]
mrtnmtth: `Inflector::slug()` replaces all non word characters. I think `Text::transliterate()` is more appropriate.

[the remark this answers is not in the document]
ADmad: My bad, I actually meant `Text:: transliterate()` :)

src/Http/Client/FormDataPart.php
```diff
@@ -219,6 +231,27 @@ public function __toString()
...
+     *
+     * @param string $name The name of the header parameter
+     * @param string $value The value of the header parameter
+     * @return string
+     */
+    protected function _headerParameterToString($name, $value)
+    {
+        $transliterated = Text::transliterate(str_replace('"', '', $value));
+        $return = sprintf('%s="%s"', $name, $transliterated);
+        if ($this->_charset !== null && $value != $transliterated) {
```
ADmad: Use `$value !== $transliterated` please.

markstory:
Thank you 👍

mrtnmtth:
Thanks to all of you for the constructive code review, too.

PR #13864, opened 2019-11-08
4.x - MySQL + PostgreSQL - Mapped DATETIME(N) and TIMESTAMP(N)

othercorey:
This adds proper type mapping for the datetime types with precision now that we have a type that supports them.

**mysql**
``DATETIME`` => ``TYPE_DATETIME``
``DATETIME(0)`` => ``TYPE_DATETIME``
``DATETIME(1-6)`` => ``TYPE_DATETIME_FRACTIONAL``
``TIMESTAMP`` => ``TYPE_TIMESTAMP``
``TIMESTAMP(0)`` => ``TYPE_TIMESTAMP``
``TIMESTAMP(1-6)`` => ``TYPE_TIMESTAMP_FRACTIONAL``

**postgresql**
``TIMESTAMP`` => ``TYPE_TIMESTAMP_FRACTIONAL``
``TIMESTAMP(0)`` => ``TYPE_TIMESTAMP``
``TIMESTAMP(1-6)`` => ``TYPE_TIMESTAMP_FRACTIONAL``

In postgres, ``TIMESTAMP`` without specifying the precision is the same as ``TIMESTAMP(6)``.  This changes the default type to support that as users don't spell out ``TIMESTAMP(6)`` with posgresql.

### DIFF
--- a/src/Database/Schema/MysqlSchema.php
+++ b/src/Database/Schema/MysqlSchema.php
@@ -100,9 +100,18 @@ class MysqlSchema extends BaseSchema
             $precision = (int)$precision;
         }
 
-        if (in_array($col, ['date', 'time', 'datetime', 'timestamp'])) {
+        if (in_array($col, ['date', 'time'])) {
             return ['type' => $col, 'length' => null];
         }
+        if (in_array($col, ['datetime', 'timestamp'])) {
+            $typeName = $col;
+            if ($length > 0) {
+                $typeName = $col . 'fractional';
+            }
+
+            return ['type' => $typeName, 'length' => null, 'precision' => $length];
+        }
+
         if (($col === 'tinyint' && $length === 1) || $col === 'boolean') {
             return ['type' => TableSchema::TYPE_BOOLEAN, 'length' => null];
         }
@@ -326,7 +335,9 @@ class MysqlSchema extends BaseSchema
             TableSchema::TYPE_DATE => ' DATE',
             TableSchema::TYPE_TIME => ' TIME',
             TableSchema::TYPE_DATETIME => ' DATETIME',
+            TableSchema::TYPE_DATETIME_FRACTIONAL => ' DATETIME',
             TableSchema::TYPE_TIMESTAMP => ' TIMESTAMP',
+            TableSchema::TYPE_TIMESTAMP_FRACTIONAL => ' TIMESTAMP',
             TableSchema::TYPE_CHAR => ' CHAR',
             TableSchema::TYPE_UUID => ' CHAR(36)',
             TableSchema::TYPE_JSON => $nativeJson ? ' JSON' : ' LONGTEXT',
@@ -393,13 +404,18 @@ class MysqlSchema extends BaseSchema
             $out .= '(' . $data['length'] . ')';
         }
 
-        $hasPrecision = [TableSchema::TYPE_FLOAT, TableSchema::TYPE_DECIMAL];
-        if (in_array($data['type'], $hasPrecision, true) && isset($data['length'])) {
+        $hasLengthAndPrecision = [TableSchema::TYPE_FLOAT, TableSchema::TYPE_DECIMAL];
+        if (in_array($data['type'], $hasLengthAndPrecision, true) && isset($data['length'])) {
             if (isset($data['precision'])) {
                 $out .= '(' . (int)$data['length'] . ',' . (int)$data['precision'] . ')';
             } else {
                 $out .= '(' . (int)$data['length'] . ')';
             }
+        }
+
+        $hasPrecision = [TableSchema::TYPE_DATETIME_FRACTIONAL, TableSchema::TYPE_TIMESTAMP_FRACTIONAL];
+        if (in_array($data['type'], $hasPrecision, true) && isset($data['precision'])) {
+            $out .= '(' . (int)$data['precision'] . ')';
         }
 
         $hasUnsigned = [
@@ -444,16 +460,28 @@ class MysqlSchema extends BaseSchema
         ) {
             $out .= ' AUTO_INCREMENT';
         }
-        if (isset($data['null']) && $data['null'] === true && $data['type'] === TableSchema::TYPE_TIMESTAMP) {
+
+        $timestampTypes = [TableSchema::TYPE_TIMESTAMP, TableSchema::TYPE_TIMESTAMP_FRACTIONAL];
+        if (isset($data['null']) && $data['null'] === true && in_array($data['type'], $timestampTypes)) {
             $out .= ' NULL';
             unset($data['default']);
         }
+
+        $dateTimeTypes = [
+            TableSchema::TYPE_DATETIME,
+            TableSchema::TYPE_DATETIME_FRACTIONAL,
+            TableSchema::TYPE_TIMESTAMP,
+            TableSchema::TYPE_TIMESTAMP_FRACTIONAL,
+        ];
         if (
             isset($data['default']) &&
-            in_array($data['type'], [TableSchema::TYPE_TIMESTAMP, TableSchema::TYPE_DATETIME]) &&
-            in_array(strtolower($data['default']), ['current_timestamp', 'current_timestamp()'])
+            in_array($data['type'], $dateTimeTypes) &&
+            strpos(strtolower($data['default']), 'current_timestamp') !== false
         ) {
             $out .= ' DEFAULT CURRENT_TIMESTAMP';
+            if (isset($data['precision'])) {
+                $out .= '(' . $data['precision'] . ')';
+            }
             unset($data['default']);
         }
         if (isset($data['default'])) {
--- a/src/Database/Schema/MysqlSchema.php
+++ b/src/Database/Schema/MysqlSchema.php
@@ -404,8 +404,8 @@ class MysqlSchema extends BaseSchema
             $out .= '(' . $data['length'] . ')';
         }
 
-        $hasLengthAndPrecision = [TableSchema::TYPE_FLOAT, TableSchema::TYPE_DECIMAL];
-        if (in_array($data['type'], $hasLengthAndPrecision, true) && isset($data['length'])) {
+        $lengthAndPrecisionTypes = [TableSchema::TYPE_FLOAT, TableSchema::TYPE_DECIMAL];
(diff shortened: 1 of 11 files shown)


src/Database/Schema/TableSchemaInterface.php
```diff
@@ -51,6 +51,13 @@ interface TableSchemaInterface extends SchemaInterface
      */
     public const TYPE_DATETIME = 'datetime';
 
+    /**
+     * Datetime with fractional seconds column type
+     *
+     * @var string
+     */
+    public const TYPE_DATETIME_FRACTIONAL = 'datetimef';
```
ADmad: A single extra character is not easy to spot. If `datetimefractional` seems to long then we can shorten it to `datetimefract`.

othercorey: I went back and forth on this.  I decided to submit the PR with the smallest version and get feedback.

ADmad: Normally it's best to maintain consistency with other constants and also keep the constant's value similar to it's name, so I would suggest keeping the value `datetime_fractional`.

othercorey: Sure, that is my preference as well.

othercorey:
I will submit a docs update to go with this in a bit.

tests/TestCase/Database/Schema/PostgresSchemaTest.php
```diff
@@ -493,11 +515,11 @@ public function testDescribeTableWithDefaults()
                 'autoIncrement' => null,
             ],
             'created' => [
-                'type' => 'timestamp',
+                'type' => 'timestampfractional',
```
markstory: Changing the default reflected type could be problematic for applications. Is there a way we could make this feature opt-in?

markstory (commented):
FormHelper will also need some changes to generate the right input type for fractional datetime/timestamp columns.

tests/TestCase/Database/Schema/PostgresSchemaTest.php
ADmad: Wouldn't those using fractional datetime/timestamp field actually want they fractional part?

othercorey: I tried to work on this, but I'm not sure the best way. You can certainly `opt-out` by remapping `timestampfractional` to `DateTimeType`.

When converting SQL to Schema, postgres doesn't show the difference between TIMESTAMP and TIMESTAMP(6).  They both have the same type name and precision value.

othercorey: > Wouldn't those using fractional datetime/timestamp field actually want they fractional part?

I believe he is talking about legacy projects that upgrade to 4.0 where using postgresl + TIMESTAMP they have broken code that accidentally sets the microseconds.

With this, they would get the microseconds back out which could cause ... I don't know a parsing error?  Maybe a display error.  It would probably be minor unless it causes an overflow of some kind in a validation suite.

othercorey: I suppose another bug it could cause is SQL logic that looks for rows with the same timestamp. If their logic is accidentally setting microseconds in the I18n\Time instance that we are accidentally scrubbing, that could be a subtle error in future inserts.

markstory: My concern was about people having conditional logic on abstract schema types. Previously they would have gotten `datetime` or `timestamp`, now they'll get something different. It could impact plugins like Bake too.

othercorey:
FormHelper now handles the fractional types.  datetime-local always formats times with milliseconds.

src/View/Widget/DateTimeWidget.php
```diff
@@ -86,16 +99,35 @@ public function render(array $data, ContextInterface $context): string
...
-        if ($data['type'] === 'datetime-local' || $data['type'] === 'time') {
-            $data += ['step' => '1'];
+        if (!isset($data['step'])) {
+            $step = null;
+            if (isset($this->defaultStep[$data['type']])) {
+                $step = $this->defaultStep[$data['type']];
+            }
+
+            if (isset($data['fieldName'])) {
+                $schemaType = $context->type($data['fieldName']);
```
ADmad: Having the widget use db schema type feels wrong to me. 

Perhaps all this introspection should be done in `FormHelper` itself since it holds the db type to input type mapping. `FormHelper::_magicOptions()` sets `step` for `number` inputs depending on whether the db type is `decimal` or `float`. ` So it can do the same for datetime inputs too.

othercorey: Makes you wonder what the point of the $context parameter is!

tests/TestCase/Database/Schema/PostgresSchemaTest.php
othercorey: @markstory I ran bake with datetime(6) and timestamp(6) with this change without any issue.  I'm not sure that anything changed really.  Is there something specific you'd like tested?

src/View/Widget/DateTimeWidget.php
ADmad: Yup, there's definitely some work needed with regards to context introspection and setting of input/widget attributes.

But that's beyond the scope of this PR and needs to be addressed separately.

With regards to your changes for consistency please just add required changes to FormHelper::_magicOptions().

src/Database/Schema/MysqlSchema.php
```diff
@@ -393,15 +404,20 @@ public function columnSql(TableSchema $schema, string $name): string
...
+        $hasLengthAndPrecision = [TableSchema::TYPE_FLOAT, TableSchema::TYPE_DECIMAL];
+        if (in_array($data['type'], $hasLengthAndPrecision, true) && isset($data['length'])) {
             if (isset($data['precision'])) {
                 $out .= '(' . (int)$data['length'] . ',' . (int)$data['precision'] . ')';
             } else {
                 $out .= '(' . (int)$data['length'] . ')';
             }
         }
 
+        $hasPrecision = [TableSchema::TYPE_DATETIME_FRACTIONAL, TableSchema::TYPE_TIMESTAMP_FRACTIONAL];
```
dereuromark: Rather `$precisionTypes` right?

src/Database/Schema/MysqlSchema.php
```diff
@@ -444,16 +460,28 @@ public function columnSql(TableSchema $schema, string $name): string
         ) {
             $out .= ' AUTO_INCREMENT';
         }
-        if (isset($data['null']) && $data['null'] === true && $data['type'] === TableSchema::TYPE_TIMESTAMP) {
+
+        $timestampTypes = [TableSchema::TYPE_TIMESTAMP, TableSchema::TYPE_TIMESTAMP_FRACTIONAL];
+        if (isset($data['null']) && $data['null'] === true && in_array($data['type'], $timestampTypes)) {
```
dereuromark: `, true` for in_array() is possible here.

src/Database/Schema/MysqlSchema.php
othercorey: I was just following the existing naming.  I can update it.

src/View/Widget/DateTimeWidget.php
markstory: The `context` objects aim to abstract the different potential sources that FormHelper has to work with. For example ORM object, elasticsearch entities, or arrays.

I don't think using the context inside the widget is incorrect. We've used `_magicOptions` to consolidate logic so we don't have duplicate logic between widgets or to generate data based on view context.

tests/TestCase/Database/Schema/PostgresSchemaTest.php
markstory: That is what I was hoping to have tested. The FormHelper test you added is great :clap:

othercorey: I think FormHelperTest is missing a test where it builds a ->control() based on the field name. That should look up the default mapping for types.  I know I've added that mapping, but I want to make sure I test it.

src/View/Widget/DateTimeWidget.php
ADmad: @markstory Should we always include the fieldname in data array for all widgets then? Because without the fieldname the context is unusable. Or maybe add an additional argument to render() for fieldname.

src/View/Widget/DateTimeWidget.php
```diff
@@ -104,22 +107,25 @@ public function render(array $data, ContextInterface $context): string
             'templateVars' => [],
         ];
 
+        if (!isset($this->formatMap[$data['type']])) {
+            throw new InvalidArgumentException(sprintf(
+                'Invalid type "%s" for input tag',
```
markstory: Could we include the list of valid types here? So its easier to figure out what a valid type is.

tests/TestCase/View/Helper/FormHelperTest.php
```diff
@@ -7413,6 +7413,41 @@ public function testDateTimeWithFractional()
...
+     */
+    public function testControlWithFractional()
+    {
+        $this->Form->create([
+            'schema' => [
+                'created' => ['type' => 'datetimefractional'],
+            ],
+        ]);
+        $result = $this->Form->control('created', [
+            'val' => new FrozenTime('2019-09-27 02:52:43'),
```
markstory: Should this value have microseconds to ensure they work their way down into the value?

src/View/Widget/DateTimeWidget.php
othercorey: Sure. This was just moved up from the function below though.

tests/TestCase/View/Helper/FormHelperTest.php
```diff
@@ -7399,13 +7399,13 @@ public function testDateTimeWithFractional()
             ],
         ]);
         $result = $this->Form->datetime('created', [
-            'val' => new FrozenTime('2019-09-27 02:52:43'),
+            'val' => new FrozenTime('2019-09-27 02:52:43.123'),
         ]);
         $expected = [
-            'input' => [
+            'input' => [\
```
ADmad: Stray backslash

othercorey: I fixed that. Must have hit the keyboard right before committing or something.

PR #13868, opened 2019-11-10
Convert prefix names to be CamelCased

markstory:
Update the routing scope helpers to create CamelCased prefix names. This is a breaking change but helps normalize routing the prefix parameter with other routing parameters.

I've added a deprecation for prefix inflection as I think we can remove it down the road if prefixes are created in PascalCase.

### DIFF
--- a/tests/TestCase/Routing/RouteBuilderTest.php
+++ b/tests/TestCase/Routing/RouteBuilderTest.php
@@ -462,7 +462,7 @@ class RouteBuilderTest extends TestCase
             $this->assertInstanceOf(RouteBuilder::class, $r);
             $this->assertCount(0, $this->collection->routes());
             $this->assertSame('/path/admin', $r->path());
-            $this->assertEquals(['prefix' => 'admin', 'key' => 'value', 'param' => 'value'], $r->params());
+            $this->assertEquals(['prefix' => 'Admin', 'key' => 'value', 'param' => 'value'], $r->params());
         });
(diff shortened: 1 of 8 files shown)


tests/TestCase/Routing/RouteBuilderTest.php
```diff
@@ -494,7 +494,7 @@ public function testNestedPrefix()
         $routes = new RouteBuilder($this->collection, '/admin', ['prefix' => 'admin']);
         $res = $routes->prefix('api', ['_namePrefix' => 'api:'], function ($r) {
             $this->assertSame('/admin/api', $r->path());
-            $this->assertEquals(['prefix' => 'admin/api'], $r->params());
+            $this->assertEquals(['prefix' => 'admin/Api'], $r->params());
```
ADmad: Shouldn't it be `Admin` too instead of `admin`?

markstory: No because the 'admin' value was passed into the RouteBuilder as a lower case value. This likely won't happen often in applications because people use the scope() helpers.

dereuromark: So `'_namePrefix' => 'contacts:']` also won't change then?

garas: `_namePrefix` is used for building `_name`, it is not related to `prefix`.

ADmad (approved):

PR #13870, opened 2019-11-10
Fix partially rendered content with errors

markstory:
When we did some refactoring of the error page rendering, the clearOutput() method ceased to be called. This resulted in buffers not being cleared when an error is encountered part way through the view rendering.

### DIFF
--- a/src/Error/ErrorHandler.php
+++ b/src/Error/ErrorHandler.php
@@ -136,7 +136,6 @@ class ErrorHandler extends BaseErrorHandler
                 Router::getRequest()
             );
             $response = $renderer->render();
-            $this->_clearOutput();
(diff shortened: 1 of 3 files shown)


src/Error/ErrorHandler.php
```diff
@@ -136,7 +136,6 @@ protected function _displayException(Throwable $exception): void
                 Router::getRequest()
             );
             $response = $renderer->render();
-            $this->_clearOutput();
```
othercorey: Curious why this was never called.

markstory: The ErrorHandlerMiddleware now gets an instance of the ExceptionRenderer and calls render() on it directly and no longer calls handleException().

PR #13872, opened 2019-11-10
Fix missing stack frame in error pages

### LINKED ISSUE #13869: Exception call stack shows frame with the parent frame's source
This is a (multiple allowed):

* [x] bug

* CakePHP Version: 4.x-dev

### What you did
Throw class not found exception

### What happened
The stack frames are linking the function being called with the current source file instead of the function currently executing.

stack

markstory:
Exceptions in PHP don't include the frame that the error originated from in the result of Exception::getTrace(). To workaround that we need to use the exception's file & line attributes as an additional frame.

Fixes #13869

### DIFF
--- a/src/Error/ExceptionRenderer.php
+++ b/src/Error/ExceptionRenderer.php
@@ -206,12 +206,18 @@ class ExceptionRenderer implements ExceptionRendererInterface
 
         $isDebug = Configure::read('debug');
         if ($isDebug) {
-            $viewVars['trace'] = Debugger::formatTrace($exception->getTrace(), [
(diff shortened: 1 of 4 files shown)


dereuromark:
Is this specific to 4.x or do we also need this in master?

markstory:
@dereuromark It should be backported to 3.x as well.

dereuromark:
Ah ok
I thought we made bugfixes to master first, and new features to 4.x (to backport).
But if that also works, sure.

othercorey:
Looks ok to me.  Unit tests are saying it shouldn't run though!

PR #13873, opened 2019-11-12
4.x - Added timezone parameter to Date and FrozenDate

othercorey:
This passes the $tz parameter down to Chronos which was added in https://github.com/cakephp/chronos/pull/218

`Date::parseDateTime()` now respects local time when grabbing the date.  This does not add support for time zones to `parseDateTime()`.

### DIFF
--- a/src/I18n/Date.php
+++ b/src/I18n/Date.php
@@ -17,7 +17,6 @@ declare(strict_types=1);
 namespace Cake\I18n;
 
 use Cake\Chronos\MutableDate;
-use DateTimeInterface;
 use IntlDateFormatter;
 
 /**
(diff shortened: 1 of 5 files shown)


src/I18n/Date.php
```diff
@@ -115,20 +115,23 @@ class Date extends MutableDate implements I18nDateTimeInterface
...
+     * The $timezone parameter is ignored if $time is a DateTimeInterface
+     * instance.
+     *
      * Date instances lack time components, however due to limitations in PHP's
      * internal Datetime object the time will always be set to 00:00:00, and the
      * timezone will always be UTC. Normalizing the timezone allows for
      * subtraction/addition to have deterministic results.
      *
      * @param string|int|\DateTimeInterface|null $time Fixed or relative time
+     * @param \DateTimeZone|string|null $tz The timezone in which the date is taken
```
ADmad: Repeating the fact that timezone is used only when `$time` is a string would be nice.

src/I18n/DateFormatTrait.php
```diff
@@ -323,20 +329,20 @@ public static function parseDateTime(string $time, $format = null)
                 is_subclass_of(static::class, MutableDate::class);
         }
 
-        $defaultTimezone = static::$_isDateInstance ? 'UTC' : date_default_timezone_get();
```
saeideng: looks you killed the `UTC` here

othercorey: That is on purpose to match how Chronos Date works.

PR #13875, opened 2019-11-12
4.x - Form Context

ADmad:
I moved the generation of default html attributes using context from `FormHelper` to widgets wherever possible.

I will update method docblocks if the changes seem reasonable.


ADmad:
Currently if not explicitly specified values for attributes like `maxlength` are only set if using `FormHelper::contro()` and not when using individual input methods like `FormHelper::text()/textarea()`.

Doesn't is make sense to always set them whenever context is available?

dereuromark:
> Doesn't it make sense to always set them whenever context is available?

I would totally say yes. Usually the main reason to use specific ones instead of control() is to remove the diff (since this is not possible that easy anymore in 3.x now). The whole form input part should be the same IMO.

src/View/Form/EntityContext.php
```diff
@@ -538,15 +538,20 @@ public function getMaxLength(string $field): ?int
...
+        if ($validator->hasField($fieldName)) {
+            foreach ($validator->field($fieldName)->rules() as $rule) {
+                if ($rule->get('rule') === 'maxLength') {
+                    return $rule->get('pass')[0];
+                }
             }
         }
 
+        $attributes = $this->attributes($field);
+        if (!empty($attributes['length'])) {
```
markstory: If the length is 0 should we return it?

src/View/Widget/BasicWidget.php
```diff
@@ -72,6 +74,24 @@ public function render(array $data, ContextInterface $context): string
         $data['value'] = $data['val'];
         unset($data['val']);
 
+        $fieldName = $data['fieldName'] ?? null;
+        if (
+            $data['type'] === 'number'
+            && !isset($data['step'])
+            && $fieldName
+        ) {
+            $data = $this->setStep($data, $context, $data['fieldName']);
```
markstory: Should this be `$fieldName` instead of `$data['fieldName']`?

src/View/Widget/HtmlAttributesTrait.php
```diff
@@ -0,0 +1,32 @@
...
+ * @copyright     Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
+ * @link          https://cakephp.org CakePHP(tm) Project
+ * @since         4.0.0
+ * @license       https://opensource.org/licenses/mit-license.php MIT License
+ */
+namespace Cake\View\Widget;
+
+use Cake\View\Form\ContextInterface;
+
+trait HtmlAttributesTrait
```
markstory: Could we mark this trait as an internal one with a doc string?

ADmad:
- Default HTML attributes based on context introspection are now set when directly using input specific methods too.
- `FormHelper::select()` now sets `empty` option to `true` by default if the field has been explicitly allowed to be empty in validator. Earlier it always defaulted to `'empty' => false` regardless of whether the field was allowed to be empty or not. 
This required changed required allowing `ContextInterface::isRequired()` return nullable, so that `EntityContext::isRequired()` returns `null` if no validations are set for the field. This way the `empty` option still defaults to `false` is no `notEmpty()` or `allowEmpty()` is not explicity set for the field.

src/View/Widget/HtmlAttributesTrait.php
ADmad: Done

src/View/Widget/BasicWidget.php
ADmad: Yup, done.

src/View/Form/EntityContext.php
ADmad: I don't think db columns with length `0` or `maxlength` attribute with `'0'` has any practical use case, so such a default doesn't make much sense.

ADmad:
The failure on travis for PHP 7.4 and Mysql seems related to the changes https://travis-ci.org/cakephp/cakephp/jobs/611047166#L577

But there are no failures for other PHP versions and mysql on GH so not sure what's causing the failure on travis.

src/View/Widget/DateTimeWidget.php
```diff
@@ -110,25 +109,15 @@ public function render(array $data, ContextInterface $context): string
         }
 
         if (!isset($data['step'])) {
-            $step = $this->defaultStep[$data['type']];
+            $data['step'] = $this->defaultStep[$data['type']];
```
othercorey: Shouldn't this default step go in the magic setStep() function too?

othercorey (approved):

src/View/Widget/DateTimeWidget.php
ADmad: The `setStep()` has `$fieldName` as an argument so I called it only if `$data['fieldName']` is set.

Also I got rid of the trait since put it methods directly in `BasicWidget` since all other widgets extend it anyway. Also `setStep()` is now overridden in `DateTimeWidget()` so introspection related to datetime fields is contained within same class.

othercorey: It looks good the way it is now.  I approved.

markstory (approved):

src/View/Form/ArrayContext.php
```diff
@@ -199,11 +199,25 @@ public function val(string $field, array $options = [])
...
      * @param string $field A dot separated path to check required-ness for.
-     * @return bool
+     * @return bool|null
      */
-    public function isRequired(string $field): bool
+    public function isRequired(string $field): ?bool
     {
-        return (bool)$this->getRequiredMessage($field);
+        if (!is_array($this->_context['required'])) {
+            return null;
```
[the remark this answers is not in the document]
ADmad: I have explained in 2nd point of my earlier comment https://github.com/cakephp/cakephp/pull/13875#issuecomment-553090729.

PR #13878, opened 2019-11-13
Updated phpunit.xml.dist default mysql DB_DSN

othercorey:
Updated phpunit.xml.dist default mysql DB_DSN to disable ONLY_FUL_GROUP_BY for local tests

Updated the readme test steps to be more specific about what setting up local db for tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,9 +48,10 @@ Assuming you have PHPUnit installed system wide using one of the methods stated
 tests for CakePHP by doing the following:
 
 1. Copy `phpunit.xml.dist` to `phpunit.xml`.
-2. Add the relevant database credentials to your `phpunit.xml` if you want to run tests against
-   a non-SQLite datasource.
-3. Run `phpunit`.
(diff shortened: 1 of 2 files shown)


dereuromark:
Why updating the default?
Shouldnt we instead make an example in docs somewhere to explain how to do that?

othercorey:
> Why updating the default?
> Shouldnt we instead make an example in docs somewhere to explain how to do that?

Because it's required? Following the readme steps didn't work.

ADmad:
Ideally the relevant tests should be be updated so that disabling `ONLY_FULL_GROUP_BY` is not required.

othercorey:
> Ideally the relevant tests should be be updated so that disabling `ONLY_FULL_GROUP_BY` is not required.

This is just doing what you do in the travis and github unit tests.  I'll cancel it though.

ADmad:
>  This is just doing what you do in the travis and github unit tests

I know, but disabling it on travis and GH was just me being lazy :P

PR #13880, opened 2019-11-13
Remove unneeded variable assignments.

ADmad:


src/Utility/Xml.php
```diff
@@ -141,9 +141,8 @@ public static function build($input, array $options = [])
      */
     protected static function _loadXml(string $input, array $options)
     {
-        $hasDisable = function_exists('libxml_disable_entity_loader');
```
markstory: Is this function always present now?

ADmad: `libxml_disable_entity_loader()` and well as the other method used below `libxml_use_internal_errors()` are both provided by `libxml` and available in PHP 7.  So I don't see why the exits check is done for this method when it's not done for the other one. `libxml_disable_entity_loader()` was added in PHP 5.2 so maybe this check was required earlier when PHP versions older than PHP 5.2 were supported.

dereuromark:
The actions need to be reverted into travis, otherwise we are blind on the additional tooling checks.

markstory:
@dereuromark We're trying to get the cakephp org in a position where we can continue to use actions.

ADmad:
Also the tests are being run on PHP 7.4 on travis. If stan or cs build do start failing I can easily sort them out later.

PR #13882, opened 2019-11-14
Check if the scheme and port is set in `Router.php:fullBaseUrl()`

groovenectar:
When I run `bin/cake --help`, I get the following output without this update:

```$ bin/cake --help
Notice Error: Undefined index: scheme
In [/var/www/vhosts/dev/vendor/cakephp/cakephp/src/Routing/Router.php, line 623]

Notice Error: Undefined index: host
In [/var/www/vhosts/dev/vendor/cakephp/cakephp/src/Routing/Router.php, line 624]
```
Does anyone else have that notice appear?

### DIFF
--- a/src/Routing/Router.php
+++ b/src/Routing/Router.php
@@ -620,8 +620,8 @@ class Router
 
         $parts = parse_url(static::$_fullBaseUrl);
         static::$_requestContext = [
-            '_scheme' => $parts['scheme'],
-            '_host' => $parts['host'],
+            '_scheme' => $parts['scheme'] ?? null,
+            '_host' => $parts['host'] ?? null,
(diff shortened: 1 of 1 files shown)


groovenectar:
It fails one of the tests, but the error is strange:

    Workflows can't be executed on this repository. Please check your payment method or billing status.

ADmad:
> It fails one of the tests, but the error is strange:

Github had made changes to what services are available based on the GH plan one it using. We are working to get it things sorted out so the Github actions for CI run again.

ADmad:
The changes seems fine. In CLI full base URL would not be available unless you have explicitly set a URL in the `Router::fullBaseUrl()` call in boostrap, so the error can occur.

PR #13887, opened 2019-11-15
4.x timehelper

ADmad:


src/View/Helper/TimeHelper.php
```diff
@@ -269,12 +269,9 @@ public function timeAgoInWords($dateTime, array $options = []): string
...
         $options['timezone'] = $this->_getTimezone($options['timezone']);
-        if (
-            $options['timezone'] &&
-            ($dateTime instanceof Time ||
-            $dateTime instanceof FrozenTime)
-        ) {
-            $dateTime = $dateTime->timezone($options['timezone']);
+        /** @psalm-suppress UndefinedInterfaceMethod */
+        if ($options['timezone'] && $dateTime instanceof DateTimeInterface) {
+            $dateTime = $dateTime->setTimezone($options['timezone']);
```
markstory: Should we be suppressing this warning? Would it be better to use method_exists here as there is no good interface type to use.

src/View/Helper/TimeHelper.php
```diff
@@ -373,7 +370,7 @@ public function format($date, $format = null, $invalid = false, $timezone = null
      * Returns a formatted date string, given either a Datetime instance,
      * UNIX timestamp or a valid strtotime() date string.
      *
-     * @param int|string|\DateTime $date UNIX timestamp, strtotime() valid string or DateTime object
+     * @param int|string|\DateTimeInterface|null $date UNIX timestamp, strtotime() valid string or DateTime object
```
othercorey: If this officially supports null, shouldn't that the be the default for $date?  Feels weird as it returns $invalid.

src/View/Helper/TimeHelper.php
ADmad: I want with practicality over technical correctness here.

Both DateTime and DateTimeImmutable has setTimezone() but sadly it's not part of DatetimeInterface. I really don't see someone coming up with a new implementation of DatetimeInterface which doesn't have setTimezone().

src/View/Helper/TimeHelper.php
ADmad: The null check is only a safeguard. I had to add it to the docblock too to prevent static analyzers from complaining.

PR #13889, opened 2019-11-16
Add assertRedirectEquals, to allow assertions against non-full URLs

Zuluru:
Addition per discussion at https://github.com/cakephp/authentication/issues/305.

### DIFF
--- a/src/TestSuite/IntegrationTestTrait.php
+++ b/src/TestSuite/IntegrationTestTrait.php
@@ -858,7 +858,7 @@ trait IntegrationTestTrait
     }
 
     /**
-     * Asserts that the Location header is correct.
+     * Asserts that the Location header is correct. Comparison is made against a full URL.
      *
      * @param string|array|null $url The URL you expected the client to go to. This
      *   can either be a string URL or an array compatible with Router::url(). Use null to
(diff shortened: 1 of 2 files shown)


src/TestSuite/IntegrationTestTrait.php
```diff
@@ -858,7 +858,7 @@ public function assertResponseCode($code, $message = null)
     }
 
     /**
-     * Asserts that the Location header is correct.
+     * Asserts that the Location header is correct. Comparison is made against a full URL.
```
markstory:
```suggestion
     * Asserts that the Location header is correct. Comparison is made against full URL containing the host and scheme.
```

src/TestSuite/IntegrationTestTrait.php
```diff
@@ -876,6 +876,25 @@ public function assertRedirect($url = null, $message = '')
...
+    /**
+     * Asserts that the Location header is correct. Comparison is made against exactly the URL provided.
+     *
+     * @param string|array|null $url The URL you expected the client to go to. This
+     *   can either be a string URL or an array compatible with Router::url(). Use null to
+     *   simply check for the existence of this header.
+     * @param string $message The failure message that will be appended to the generated message.
+     * @return void
+     */
+    public function assertRedirectEquals($url = null, $message = '')
```
markstory: Should this be `assertRedirectSame()`? PHPUnit uses Same to indicate additional precision/type checks.

Zuluru: I took my cues from functions in here like `assertResponseEquals` and helper classes like `BodyEquals`, `HeaderEquals`, etc., but can certainly change it if `Same` is preferred.

PR #13890, opened 2019-11-17
4.x - Http Server

ADmad:
- Allow `Runner` instance to be injected into `Server` constructor and remove `Server::setRunner()`. I can't think of a use case where you would need to change the runner after server instance is created.
- `Server::setEventManager/getEventManager()` now check if application implements `EventDispatcherInterface` instead of `PluginApplicationInterface`. An application should be able to use events without using plugins.
- Allow passing `MiddlewareQueue` instance to `Server::run()` which was previously hard coded in the method.


markstory:
Nice cleanup :clap:

PR #13893, opened 2019-11-18
Add support for accounting currency formatting provided by the ICU library

Zuluru:
The general standard in accounting is not to display negative currency amounts with a minus sign, but with parentheses around them. This is supported by the ICU library which the PHP INTL module is based on, but not directly by INTL itself. This patch adds support for specifying FORMAT_CURRENCY_ACCOUNTING instead of FORMAT_CURRENCY as the preferred display mode when using the currency formatter.

This doesn't change anything about existing functionality, just adds the new option.


src/I18n/Number.php
```diff
@@ -39,6 +39,20 @@ class Number
      */
     const FORMAT_CURRENCY = 'currency';
 
+    /**
+`     * Format type to format as currency, accounting style (negative numbers in parentheses)
```
dereuromark: Invalid character ` here at the beginning

src/I18n/Number.php
```diff
@@ -39,6 +39,20 @@ class Number
...
+     *
+     * @var string
```
dereuromark: these are not necessary for strings, they cant change at runtime anymore

src/I18n/Number.php
```diff
@@ -39,6 +39,20 @@ class Number
...
+     */
+    const FORMAT_CURRENCY_ACCOUNTING = 'currency_accounting';
+
+    /**
+     * ICU Constant for accounting format; not yet supported by INTL library
+     *
+     * @var int
```
dereuromark: same

src/I18n/Number.php
```diff
@@ -263,6 +284,31 @@ public static function defaultCurrency($currency = null)
...
+     * Getter/setter for default currency format
+     *
+     * @param null $currencyFormat Default currency format to be used by currency()
+     * if $currencyFormat argument is not provided. If boolean false is passed, it will clear the
+     * currently stored value
+     * @return string CurrencyFormat
+     */
+    public static function defaultCurrencyFormat($currencyFormat = null)
+    {
+        if (!empty($currencyFormat)) {
```
dereuromark: !empty() is bad here if format can also be false (next line). Use better strict `!== null` here

src/I18n/Number.php
```diff
@@ -263,6 +284,31 @@ public static function defaultCurrency($currency = null)
...
+            return self::$_defaultCurrencyFormat = $currencyFormat;
```
dereuromark: `static::` in framework context.

src/I18n/Number.php
```diff
@@ -263,6 +284,31 @@ public static function defaultCurrency($currency = null)
         return self::$_defaultCurrency;
     }
 
+    /**
+     * Getter/setter for default currency format
+     *
+     * @param null $currencyFormat Default currency format to be used by currency()
```
dereuromark: wrong types (missing ones)

src/I18n/Number.php
```diff
@@ -263,6 +284,31 @@ public static function defaultCurrency($currency = null)
...
+    {
+        if (!empty($currencyFormat)) {
+            return self::$_defaultCurrencyFormat = $currencyFormat;
+        }
+
+        if ($currencyFormat === false) {
+            return self::$_defaultCurrencyFormat = null;
+        }
+
+        if (empty(self::$_defaultCurrencyFormat)) {
```
dereuromark: Also `!== null then`

src/I18n/Number.php
```diff
@@ -296,6 +342,8 @@ public static function formatter($options = [])
             $type = $options['type'];
             if ($options['type'] === static::FORMAT_CURRENCY) {
                 $type = NumberFormatter::CURRENCY;
+            } elseif ($options['type'] === static::FORMAT_CURRENCY_ACCOUNTING) {
+                $type = static::CURRENCY_ACCOUNTING;
```
dereuromark: Isnt that FORMAT_CURRENCY_ACCOUNTING?
You need to add test cases for this specific one, also please consider new features going to 4.x first, and 3.next (3.9) as following backport please.

markstory: This is currently targeting master but also adding new features. Could you change this pull request to target 3.next instead?

othercorey:
Is there a php bug opened on this?

src/I18n/Number.php
```diff
@@ -39,6 +39,20 @@ class Number
...
+     * @var string
+     */
+    const FORMAT_CURRENCY_ACCOUNTING = 'currency_accounting';
+
+    /**
+     * ICU Constant for accounting format; not yet supported by INTL library
+     *
+     * @var int
+     */
+    const CURRENCY_ACCOUNTING = 12;
```
othercorey: This needs documentation on where the value comes from and why it would work without explicit support.

src/I18n/Number.php
Zuluru: But if it's `false`, it *does* need to continue on.

src/I18n/Number.php
Zuluru: Stickler broke the type, because I had the variable name wrong.

src/I18n/Number.php
Zuluru: Again, just copied how `defaultCurrency` works. Should I clean all that up while I'm here?

src/I18n/Number.php
dereuromark: Exactly, and it would with strict check.

Zuluru: `false` is `!== null`, so this check would be true, and it would return right away. I can check for both `!== null` and `!== false`.

src/I18n/Number.php
```diff
@@ -246,21 +260,46 @@ public static function currency($value, $currency = null, array $options = [])
...
+     * currently stored value
+     * @return string CurrencyFormat
+     */
+    public static function defaultCurrencyFormat($currencyFormat = null)
+    {
+        if ($currencyFormat !== null && $currencyFormat !== false) {
+            return static::$_defaultCurrencyFormat = $currencyFormat;
+        }
+
+        if ($currencyFormat === false) {
```
garas: Swap with first `if` to avoid checking twice for `false`

src/I18n/Number.php
```diff
@@ -246,21 +260,46 @@ public static function currency($value, $currency = null, array $options = [])
...
+            static::$_defaultCurrency = $formatter->getTextAttribute(NumberFormatter::CURRENCY_CODE);
+        }
+
+        return static::$_defaultCurrency;
+    }
+
+    /**
+     * Getter/setter for default currency format
+     *
+     * @param string|bool|null $currencyFormat Default currency format to be used by currency()
```
garas: `false` instead of `bool`

othercorey:
> Is there a php bug opened on this?

Can we link a php bug to maybe track a fix for php 8?

src/I18n/Number.php
```diff
@@ -239,28 +253,53 @@ public static function currency($value, $currency = null, array $options = [])
...
+            static::$_defaultCurrency = $formatter->getTextAttribute(NumberFormatter::CURRENCY_CODE);
+        }
+
+        return static::$_defaultCurrency;
+    }
+
+    /**
+     * Getter/setter for default currency format
+     *
+     * @param string|false|null $currencyFormat Default currency format to be used by currency()
```
markstory: This is a complex type that we can't express with a typehint. Would we be able to restrict this to `string|null`? Where null clears the current value and string values are stored? By providing `FORMAT_CURRENCY` it is straightforward to set a useful format option and `defaultCurrencyFormat(Number::FORMAT_CURRENCY)` is simpler to understand than `defaultCurrencyFormat(null)`

Zuluru: Actually, why has the `defaultCurrency` function (which I modelled this on) not been deprecated and replaced with a separate getter/setter pair? That's probably the better solution here?

markstory: Likely an oversight as we had a ton of methods to change. We could add the get/set pair for the new feature and add a deprecation in a separate pull request.

Zuluru:
> Can we link a php bug to maybe track a fix for php 8?

https://bugs.php.net/bug.php?id=78912

src/I18n/Number.php
dereuromark: You can probably now rebase on top of 3.next.

markstory:
Looks good other than the [PHPStan warning](https://travis-ci.org/cakephp/cakephp/jobs/622394135#L319)

othercorey:
Looks like we'll get the feature in php 7.4.1

ADmad:
That constant usage is already wrapped in a `defined()` check, so the phpstan error needs to be ignored.

Zuluru:
> That constant usage is already wrapped in a `defined()` check, so the phpstan error needs to be ignored.

Doesn't look like there's a comment I can add in the code that will make phpstan ignore it? Wasn't sure if the "needs to be ignored" was for me to resolve or not.

ADmad:
I took care of ignoring the phpstan error.

PR #13894, opened 2019-11-18
4.x commands

ADmad:


src/Console/ConsoleInputOption.php
```diff
@@ -93,18 +93,23 @@ public function __construct(
...
     ) {
         $this->_name = $name;
         $this->_short = $short;
         $this->_help = $help;
         $this->_boolean = $boolean;
-        $this->_default = is_bool($default) ? $default : (string)$default;
         $this->_choices = $choices;
         $this->_multiple = $multiple;
 
+        if ($boolean) {
```
dereuromark: Should we name this $isBoolean to be more clear? Currently it reads as bool true/false input.

markstory (commented):
Looks good to me.

PR #13897, opened 2019-11-19
Forward port 3.next into 4.x.

dereuromark:
merges 3.next => 4.x.

### DIFF
--- a/src/Datasource/EntityTrait.php
+++ b/src/Datasource/EntityTrait.php
@@ -302,7 +302,7 @@ trait EntityTrait
      */
     public function getOriginal(string $field)
     {
-        if (!strlen((string)$field)) {
+        if (!strlen($field)) {
             throw new InvalidArgumentException('Cannot get an empty field');
         }
         if (array_key_exists($field, $this->_original)) {
@@ -354,7 +354,7 @@ trait EntityTrait
      * When checking multiple fields. All fields must not be null
      * in order for true to be returned.
      *
-     * @param string|array $field The field or fields to check.
+     * @param string|string[] $field The field or fields to check.
      * @return bool
      */
     public function has($field): bool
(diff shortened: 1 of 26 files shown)


src/Validation/Validation.php
```diff
@@ -330,12 +330,12 @@ public static function comparison($check1, string $operator, $check2): bool
                 }
                 break;
             case static::COMPARE_EQUAL:
-                if ($check1 === $check2) {
+                if ($check1 == $check2) {
```
dereuromark: How did that happen?

src/Datasource/EntityTrait.php
```diff
@@ -434,7 +434,7 @@ public function hasValue(string $field): bool
      * $entity->unset(['name', 'last_name']);
      * ```
      *
-     * @param string|array $field The field to unset.
+     * @param string|string[] $field The field to unset.
```
ADmad: This change needs to be done for `EntityInterface::unset()` too.

src/Utility/CookieCryptTrait.php
```diff
@@ -92,10 +92,10 @@ protected function _checkCipher(string $encrypt): void
     /**
      * Decrypts $value using public $type method in Security class
      *
-     * @param array|string $values Values to decrypt
+     * @param string[]|string $values Values to decrypt
      * @param string|false $mode Encryption mode
      * @param string|null $key Used as the security salt if specified.
-     * @return string|array Decrypted values
+     * @return string|string[] Decrypted values
```
ADmad: Same need to be done for return type of  `_decode()` method called by `_decrypt()` else psalm is unhappy.

src/Validation/Validation.php
markstory: I think it likely happened a long time ago when we changed all the `==` to `===` and didn't notice this.

PR #13898, opened 2019-11-19
Fix App path for usability.

dereuromark:
Resolves problematic usability described in https://github.com/cakephp/cakephp/issues/13604

- `App::path()` remains for all non class path elements.
- `App::classPath()` has been added and can proxy Plugin::classPath()
- App::path() is allowed to also use plugin, same as others for convenience.

Does that make sense?

The changes mean that e.g. my IdeHelper can keep the simple App::className($type, $plugin) syntax here for simplicity.

I wasn't sure if we should extract also templatePath(), since locales and other such keys don't have this yet either.
Or do we plan to add `public function getLocalePath()` to Plugin class as well?

### DIFF
--- a/src/Core/App.php
+++ b/src/Core/App.php
@@ -179,31 +179,45 @@ class App
      * Will return the path for tables `src/Model/Table`.
      *
      * @param string $type type of path
-     * @param string|null $plugin This argument is deprecated.
-     *   Use \Cake\Core\Plugin::classPath()/templatePath() instead for plugin paths.
-     * @return array
+     * @param string|null $plugin This argument is deprecated for class paths.
+     *   Use \Cake\Core\Plugin::classPath() instead or directly the method on \Cake\Core\Plugin.
+     * @return string[]
+     * @deprecated 4.0 Use \Cake\Core\App::classPath()/templatePath() instead.
      * @link https://book.cakephp.org/4/en/core-libraries/app.html#finding-paths-to-namespaces
      */
     public static function path(string $type, ?string $plugin = null): array
     {
-        if (empty($plugin)) {
-            if ($type[0] === strtolower($type[0])) {
-                return (array)Configure::read('App.paths.' . $type);
-            }
-
-            return [APP . $type . DIRECTORY_SEPARATOR];
+        if ($plugin === null && $type[0] === strtolower($type[0])) {
+            return (array)Configure::read('App.paths.' . $type);
         }
-
-        deprecationWarning(
-            'Using App::path() with 2nd argument $plugin is deprecated.'
-            . ' Use \Cake\Core\Plugin::classPath()/templatePath() instead.'
-        );
 
         if ($type === 'templates') {
             return [Plugin::templatePath($plugin)];
         }
 
-        return [Plugin::classPath($plugin) . $type . DIRECTORY_SEPARATOR];
+        deprecationWarning(
+            'App::path() is deprecated for class path.'
+            . ' Use \Cake\Core\App::classPath() instead or directly the method on \Cake\Core\Plugin.'
+        );
+
+        return static::classPath($type, $plugin);
+    }
+
+    /**
+     * @param string $type
+     * @param string|null $plugin
+     *
+     * @return string[]
+     */
+    public static function classPath(string $type, ?string $plugin = null): array
+    {
+        if ($plugin !== null) {
+            return [
+                Plugin::classPath($plugin) . $type . DIRECTORY_SEPARATOR,
+            ];
+        }
+
+        return [APP . $type . DIRECTORY_SEPARATOR];
     }
 
     /**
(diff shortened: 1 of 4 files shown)


src/Core/App.php
```diff
@@ -179,31 +179,45 @@ protected static function _classExistsInBase(string $name, string $namespace): b
...
-            return [APP . $type . DIRECTORY_SEPARATOR];
+        if ($type === 'templates') {
+            return [Plugin::templatePath($plugin)];
         }
 
         deprecationWarning(
-            'Using App::path() with 2nd argument $plugin is deprecated.'
-            . ' Use \Cake\Core\Plugin::classPath()/templatePath() instead.'
+            'App::path() is deprecated for class path.'
+            . ' Use \Cake\Core\App::classPath() instead or directly the method on \Cake\Core\Plugin.'
```
markstory:
```suggestion
            . ' Use \Cake\Core\App::classPath() or \Cake\Core\Plugin::classPath() instead.'
```

src/Core/App.php
```diff
@@ -179,31 +179,45 @@ protected static function _classExistsInBase(string $name, string $namespace): b
...
         );
 
-        if ($type === 'templates') {
-            return [Plugin::templatePath($plugin)];
+        return static::classPath($type, $plugin);
+    }
+
+    /**
```
markstory:
```suggestion
    /**
     * Get the path to a class type in the application or a plugin.
     *
```

src/Core/App.php
```diff
@@ -179,31 +179,45 @@ protected static function _classExistsInBase(string $name, string $namespace): b
      * Will return the path for tables `src/Model/Table`.
      *
      * @param string $type type of path
-     * @param string|null $plugin This argument is deprecated.
-     *   Use \Cake\Core\Plugin::classPath()/templatePath() instead for plugin paths.
-     * @return array
+     * @param string|null $plugin This argument is deprecated for class paths.
+     *   Use \Cake\Core\Plugin::classPath() instead or directly the method on \Cake\Core\Plugin.
+     * @return string[]
+     * @deprecated 4.0 Use \Cake\Core\App::classPath()/templatePath() instead.
```
ADmad: But there's not `App::templatePath()` currently.

dereuromark: Do we want to create it? Or rather keep all lowercase "templates"/"locales" here in App::path()?

src/Core/App.php
```diff
@@ -178,32 +178,63 @@ protected static function _classExistsInBase(string $name, string $namespace): b
...
+     * Example:
+     *
+     * ```
+     * App::path('Model/Table');
+     * ```
+     *
+     * Will return the path for tables `src/Model/Table`.
+     *
+     * ```
+     * App::path('My/Plugin.Model/Table');
```
ADmad: Method name as well the argument doesn't seem right.

dereuromark: Copy and paste ;) thx!

dereuromark:
I also added a note about default types for path(), since this is otherwise hard to retrieve from code.

src/Core/App.php
```diff
@@ -172,38 +177,63 @@ protected static function _classExistsInBase(string $name, string $namespace): b
...
+     * Example:
+     *
+     * ```
+     * App::classPath('Model/Table');
+     * ```
+     *
+     * Will return the path for tables `src/Model/Table`.
+     *
+     * ```
+     * App::classPath('My/Plugin.Model/Table');
```
ADmad: The argument still doesn't seem right to me. Shouldn't it be `App::classPath('Model/Table', 'My/Plugin')`?

dereuromark: I guess my brain was too much into Router::path()^^

tests/TestCase/Core/AppTest.php
```diff
@@ -216,6 +216,22 @@ public function shortNameProvider()
...
+     * test classPath() with a plugin.
+     *
+     * @return void
+     */
+    public function testClassPathWithPlugins()
+    {
+        $basepath = TEST_APP . 'Plugin' . DS;
+        $this->loadPlugins(['TestPlugin', 'Company/TestPluginThree']);
+        $result = App::classPath('Controller', 'TestPlugin');
+        $this->assertPathEquals($basepath . 'TestPlugin' . DS . 'src' . DS . 'Controller' . DS, $result[0]);
```
ADmad: Nitpick: A newline after this assert would have been nice for easier reading :slightly_smiling_face:.

dereuromark: True. I just copied them from the other test case.

ADmad (approved):
Changes LGTM. The plugins would also have to be updated to use the new `classPath()` method instead of `path()` wherever applicable.

dereuromark:
Docs: https://github.com/cakephp/docs/pull/6279

PR #13903, opened 2019-11-19
Fix CS for easier 3.x/4.x comparison/merge.

dereuromark:
When doing 4.x to 3.x backports (and the other way around), I noticed that the noise in diff is huge on those non functional CS changes.

Ideally we would enable those non-php7 ones on master (3.x) as well as new codesniffer minor release or alike.
But for now just making them in sync here enables easier diffs in the future, especially for the trailing commas, as it doesnt create diff-noise on added lines then.


dereuromark:
There seem to be unrelated issues in cs we can make in separate pr.

dereuromark:
https://github.com/cakephp/cakephp/pull/13908
I will then merge this to have final review on new sniff changes and green master.

Issue #13901, opened 2019-11-19
Missing validation rule allowNull()


bancer:
This is a (multiple allowed):

* [ ] bug
* [x] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.8.5.
* Platform and Target: Linux, Apache, MySQL.

### What you did
I tried to create validation rules for a column that may contain `string` min length 36 or `null` but not an empty string.

### What happened
 `allowEmpty()` worked fine but is deprecated and allowed to save an empty string. `allowEmptyString()` cannot be used because empty strings are not allowed to be saved according to the business rules. 

### What you expected to happen
I would expect to have something like `allowNull()` validation rule. I think it is pretty common to have nullable varchar columns in a database.

bancer:
@ADmad It is good that it is fixed in 4.x but the problem was 3.x.

dereuromark:
@bancer You can backport it to 3.x

PR #13904, opened 2019-11-19
Make Validator::allowEmptyFor() only allow null as empty when no flags are specified.

ADmad:
Refs #13901.

### DIFF
--- a/src/Validation/Validator.php
+++ b/src/Validation/Validator.php
@@ -202,7 +202,7 @@ class Validator implements ArrayAccess, IteratorAggregate, Countable
 
             $canBeEmpty = $this->_canBeEmpty($field, $context);
 
-            $flags = static::EMPTY_ALL;
+            $flags = 0;
             if (isset($this->_allowEmptyFlags[$name])) {
                 $flags = $this->_allowEmptyFlags[$name];
             }
@@ -699,7 +699,12 @@ class Validator implements ArrayAccess, IteratorAggregate, Countable
         foreach ($field as $fieldName => $setting) {
             $settings = $this->_convertValidatorToArray($fieldName, $defaults, $setting);
             $fieldName = array_keys($settings)[0];
-            $this->allowEmptyFor($fieldName, null, $settings[$fieldName]['when'], $settings[$fieldName]['message']);
+            $this->allowEmptyFor(
+                $fieldName,
+                static::EMPTY_ALL,
+                $settings[$fieldName]['when'],
+                $settings[$fieldName]['message']
+            );
         }
 
         return $this;
@@ -1149,6 +1154,7 @@ class Validator implements ArrayAccess, IteratorAggregate, Countable
             $whenSetting = $this->invertWhenClause($settings[$fieldName]['when']);
 
             $this->field($fieldName)->allowEmpty($whenSetting);
+            $this->_allowEmptyFlags[$fieldName] = static::EMPTY_ALL;
             if ($settings[$fieldName]['message']) {
                 $this->_allowEmptyMessages[$fieldName] = $settings[$fieldName]['message'];
             }
--- a/src/Validation/Validator.php
+++ b/src/Validation/Validator.php
@@ -53,6 +53,15 @@ class Validator implements ArrayAccess, IteratorAggregate, Countable
     /**
      * A flag for allowEmptyFor()
      *
+     * When `null` is given, it will be recognized as empty.
+     *
+     * @var int
+     */
+    public const EMPTY_NULL = 0;
+
+    /**
+     * A flag for allowEmptyFor()
+     *
      * When an empty string is given, it will be recognized as empty.
      *
      * @var int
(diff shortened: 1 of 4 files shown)


tests/TestCase/Validation/ValidatorTest.php
```diff
@@ -600,6 +600,34 @@ public function testCustomErrorsWithPresenceRequiredAsArray()
...
+        $results = $validator->errors(['title' => '']);
+        $this->assertSame(['title' => ['minLength' => 'Min. length 5 chars']], $results);
+
+        $validator
+            ->allowEmptyFor('name', Validator::EMPTY_STRING);
+
+        $results = $validator->errors(['name' => null]);
+        $this->assertSame([], $results);
+
+        $results = $validator->errors(['name' => '']);
```
markstory: Should there be a test for 0 and empty array as well?

src/Validation/Validator.php
```diff
@@ -202,7 +202,7 @@ public function errors(array $data, bool $newRecord = true): array
 
             $canBeEmpty = $this->_canBeEmpty($field, $context);
 
-            $flags = static::EMPTY_ALL;
+            $flags = 0;
```
markstory: What is 0?

tests/TestCase/Validation/ValidatorTest.php
ADmad: Done

src/Validation/Validator.php
ADmad: It has no special meaning except doing an logic `&` with it using any number will results in `0` itself.

ADmad:
This should be merged only after ` 4.x-validator-deprecations` is merged to ensure there are no other unwanted regressions.

othercorey:
Would it be possible to have an EMPTY_NULL_ONLY flag that would help indicate more than just `null`?

I see that the absence of flags indicates that all other empty checks would fail, but it's hard to read that sometimes.

ADmad:
@othercorey I wanted to first ensure that the changes don't cause any unwanted side affects. Once  #13906 is merged I'll rebase and everything seems fine I'll look into adding additional flag and/or method.

ADmad:
@othercorey I have updated the docblock for `allowEmptyFor` stating the behavior in case no flag/bitmask is provided.

src/Validation/Validator.php
othercorey: This is normal for a bitwise field.

markstory: Should we have a constant for that to make its purpose more explicit? eg. `EMPTY_NULL`?

ADmad: Will do.

ADmad: Done

src/Validation/Validator.php
```diff
@@ -771,7 +785,8 @@ public function allowEmpty($field, $when = true, $message = null)
      * ```
      *
      * @param string $field The name of the field.
-     * @param int|null $flags A bitmask of EMPTY_* flags which specify what is empty
+     * @param int|null $flags A bitmask of EMPTY_* flags which specify what is empty.
```
othercorey: This could technically default to EMPTY_NULL now.

ADmad: Yes, but if for e.g. someone wants to pass the 3rd or 4th argument and wants to keeps flags to default that using `null` is more convenient than `Validator::EMPTY_NULL`.

src/Validation/Validator.php
```diff
@@ -202,7 +211,7 @@ public function errors(array $data, bool $newRecord = true): array
 
             $canBeEmpty = $this->_canBeEmpty($field, $context);
 
-            $flags = static::EMPTY_ALL;
+            $flags = static::EMPTY_NULL;
```
markstory: I can't shake the feeling that this behavior change is going to come back and bite us. Should we instead be requiring people to use `allowEmptyFor('field', Validator::EMPTY_NULL)` to get the new behavior?

othercorey: My understanding from this change was it moved the default of EMPTY_ALL from `errors()` to the `allowEmptyFor()` call.  I'm not sure implicit null == EMPTY_NULL is the cleanest, but I don't know how many other options follow that.

ADmad: > My understanding from this change was it moved the default of EMPTY_ALL from errors() to the allowEmptyFor() call.

Yup.

@markstory All test cases still pass. The only difference the change causes is that error message shown. For e.g. earlier if you didn't use `notEmptyString()` for a field as just use `minLength()` then submitting a field with empty string you got the generic `This field cannot be empty` message. But not since empty string is not considered empty by default the validator is run and you get the error message specific to `minLength()` validator. I don't think this will cause any real world problems.

ADmad: If users do have issues which can swap back the constants in a bugfix release.

burzum:
@markstory this has not yet been backported to 3.x right?

markstory:
@burzum No, it is a new behavior so there wasn't a reason to backport it.

PR #13906, opened 2019-11-19
Trigger deprecation warning for Validator::allowEmpty()/notEmpty().

ADmad:


markstory (approved):
Looks good. I don't think we can automate any of these changes with rector sadly.

PR #13908, opened 2019-11-20
Master cs psr12

dereuromark:
Green master CS again with https://github.com/cakephp/cakephp-codesniffer/pull/271


dereuromark:
Darn, stickler is reverting the changes :(

dereuromark:
@markstory Can you please check how stickler can maybe not destroy future PRs here? Do we need to make a patch release on the sniffer here?
Master is now green, but future PRs would have auto-stickler-reverts.

PR #13911, opened 2019-11-20
Add missing assert.

ADmad:


tests/TestCase/Validation/ValidatorTest.php
```diff
@@ -630,14 +630,14 @@ public function testAllowEmptyFor()
         $results = $validator->errors(['name' => null]);
         $this->assertSame([], $results);
 
-        $this->assertSame([], $results);
         $results = $validator->errors(['name' => '']);
-
         $this->assertSame([], $results);
-        $results = $validator->errors(['title' => 0]);
 
+        $results = $validator->errors(['title' => 0]);
```
othercorey: Did you mean 'name' here?

tests/TestCase/Validation/ValidatorTest.php
```diff
@@ -630,14 +630,14 @@ public function testAllowEmptyFor()
...
         $this->assertSame(['title' => ['minLength' => 'Min. length 5 chars']], $results);
+
         $results = $validator->errors(['title' => []]);
+        $this->assertSame([], $results);
```
ADmad: @markstory I think this test shows that the changes I have done in #13904 are actually necessary. Even though `->allowEmptyFor('name', Validator::EMPTY_STRING)` is used for the field it treats empty array as empty and skips validation.

After my changes it runs the validations as expected and generates error https://github.com/cakephp/cakephp/pull/13904/files#diff-2f1913506d5e9a181af4bb924b1a3018R640

tests/TestCase/Validation/ValidatorTest.php
ADmad: Doh, I have messed the field names in both PR. Thanks for point it out.

othercorey: I missed it too and merged it :(

PR #13914, opened 2019-11-20
4.x - Added sqlserver support for fractional datetime schema types

othercorey:
This adds fractional schema type support as well as some sqlserver maintenance.

- Maps types to the new datatimefractional type
- Changes the data type to schema type mapping from `TYPE_TIMESTAMP` to `TYPE_DATETIME`
- Changes the SQL generation for 'datetime' fixture type to `datetime2` sqlserver data type
- Allows for default values to be functions like `sysdatetime()` which return `datetime2` types
- Make sure boolean types are cast to int only after a null check

### DIFF
--- a/src/Database/Schema/SqlserverSchema.php
+++ b/src/Database/Schema/SqlserverSchema.php
@@ -91,7 +91,12 @@ class SqlserverSchema extends BaseSchema
             return ['type' => $col, 'length' => null];
         }
         if (strpos($col, 'datetime') !== false) {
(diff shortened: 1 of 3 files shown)


othercorey:
@ADmad @markstory This is finally ready for review.

markstory (approved):
Looks good to me.

markstory:
What does the timestamp abstract type generate in sqlserver now?

othercorey:
> What does the timestamp abstract type generate in sqlserver now?

It generates a `datetime2` column data type the same as `datetime`.  It's just an alias ... similar to how postgres works.

PR #13918, opened 2019-11-21
Drop "safe" option for HtmlHelper::script()/scriptBlock().

ADmad:
Using CDATA tag was only required for XHTML which is defunct now.

### DIFF
--- a/src/View/Helper/HtmlHelper.php
+++ b/src/View/Helper/HtmlHelper.php
@@ -533,8 +533,6 @@ class HtmlHelper extends Helper
      *
      * ### Options
      *
-     * - `safe` (boolean) Whether or not the $script should be wrapped in `<![CDATA[ ]]>`.
-     *   Defaults to `false`.
      * - `block` Set to true to append output to view block "script" or provide
      *   custom block name.
      *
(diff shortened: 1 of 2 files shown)


ADmad:
There's also `HtmlHelper::doctype()` tag which can generate `<!DOCTYPE ...>` tags for various versions of HTTML4, XHTML and HTML5. Given that all other doctypes besides HTML5 are defunct now and directly using `<!DOCTYPE html>` is simply enough I proper dropping this method altogether.

dereuromark (approved):
makes sense to simlify

dereuromark:
Regarding docType() method.
What would be useful if that method was to also automatically add the lang attribute as per current language (I18n).
That way it would be useful again as some kind of convenience wrapper.

https://developer.paciellogroup.com/blog/2016/06/using-the-html-lang-attribute/
Quite useful for search engines.

ADmad:
@dereuromark Yes that would be using, but that's the `<html>` tag not `<!DOCTYPE>` tag, both are different.

dereuromark:
You are right, that would be a different topic :) So I agree with your removal proposal.

markstory:
This will need to be noted in the migration guide butnotherwise seems ok to me. The likely hood of someone using an xhtml strict doctype and wanting to upgrade is low.

inoas:
AFAIK this is not true. HTML5 can be served as XML complaint or not. If you serve it xml complaint u want self closing tags, attributes that always require a value and CDATA. That in turn enables you to do XSL(T) processing, MathML, complaint inline SVG etc.

XHTML 1.0/1.1 may be defunct but HTML as XML is not.

ADmad:
The framework doesn't need to support every edge case usage. The chances of someone rendering HTML5 in XML mode is slim to none. They can always use that own helper methods.

othercorey:
If we merged the doc changes is this ok?

PR #13920, opened 2019-11-21
Remove unneeded method HtmlHelper::docType().

ADmad:
HTML4 and XHTML are now defunct and doctype for HTML5 is pretty short
and easy to type out directly.

### DIFF
--- a/src/View/Helper/HtmlHelper.php
+++ b/src/View/Helper/HtmlHelper.php
@@ -93,51 +93,6 @@ class HtmlHelper extends Helper
     protected $_scriptBlockOptions = [];
 
     /**
-     * Document type definitions
-     *
-     * @var string[]
-     */
-    protected $_docTypes = [
-        // phpcs:disable Generic.Files.LineLength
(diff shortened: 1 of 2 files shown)


dereuromark:
We should probably soft deprecate those in 3.x (3.9)
Adding deprecated tags should suffice, no need to throw real notices IMO.

markstory:
If we don't include notices won't people using those methods get fatal errors when they attempt to move from 3.9 to 4.0 which isn't a great experience and is not inline with what we've been telling people to expect of the upgrade process.

inoas:
I'd not remove that in 4.x but in 5.x and just remove it from cakephp/app for now. Otherwise it is much hassle for no gain.

dereuromark:
@ionas Who is actually using those? And most of them will stay on 3.9 long enough that they can switch easily in time to upgrade.

inoas:
Probably almost everyone as they are part of cakephp/app aint they?

dereuromark:
Rector could easily replace afaik.

ADmad:
> Probably almost everyone as they are part of cakephp/app aint they?

App skeleton doesn't use this method https://github.com/cakephp/app/blob/master/src/Template/Layout/default.ctp#L19.

PR #13921, opened 2019-11-21
Mark HtmlHelper::docType() as deprecated.

ADmad:


src/View/Helper/HtmlHelper.php
```diff
@@ -185,6 +185,7 @@ public function addCrumb($name, $link = null, array $options = [])
      * @param string $type Doctype to use.
      * @return string|null Doctype string
      * @link https://book.cakephp.org/3/en/views/helpers/html.html#creating-doctype-tags
+     * @deprecated 3.9.0 This method will be removed in 4.0.0.
```
markstory: Should we also add a deprecation warning?

dereuromark: I don't think that is necessary in those trivial cases.

ADmad: @dereuromark suggested just adding deprecation tag so that's what I did. Also didn't we decided to avoid throwing new warning in 3.9?

I don't care either way, happy to do whatever is preferred.

dereuromark: This is usually more useful if there is a replacement to be done.
For just dropped stuff not sure how useful those are.

markstory: The warnings let people avoid fatal errors before they upgrade.

dereuromark: Fair enough

ADmad: I have added explicit warning.

Issue #13923, opened 2019-11-21
4.x - Cache list engines command not found?


davidyell:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 4.0.0-RC1
* Platform and Target: Nginx 1.17.3 & PHP 7.3.10

### What you did
* `bin/cake cache`
* `bin/cake cache clear`
* `bin/cake cache list_engines`

### What happened
I got an error for unknown command, even though the help text told me about the command.

![Terminal screenshot](https://i.imgur.com/NZH1kns.png)

### What you expected to happen
I wanted to clear all the caches.

### What's the issue?
The clear_all command works `bin/cake cache clear_all` still, but exploring the commands from `bin/cake` and diving into commands that way seems to have some issues in the help text, and also says about `list_engines` which then doesn't seem to work.

markstory:
It should be suggesting `cache list`.

PR #13924, opened 2019-11-21
Fix incorrect help.

markstory:
Refs #13923


src/Command/CacheClearCommand.php
```diff
@@ -54,7 +54,7 @@ public function buildOptionParser(ConsoleOptionParser $parser): ConsoleOptionPar
             ->addArgument('engine', [
                 'help' => 'The cache engine to clear.' .
                     'For example, `cake cache clear _cake_model_` will clear the model cache' .
```
davidyell: I'm not sure these are concatenating with a space. Leading to conjoined sentences.

othercorey: Needs periods and spaces.

Issue #13930, opened 2019-11-22
Separate Console package has missing dependencies


othercorey:
* [x] bug

* CakePHP Version: cakephp/console 4.0.0-RC1

### What you did
Created a project with cakephp/console package.

### What happened

* Fails to load `VersionCommand`. These files are not part of the repo.
* Fails to load `LocatorAwareTrait` in Command.php.  Datasource and ORM are not dependencies.

I can get a command to run through `CommandRunner` in the stand-alone package if I remove the addition of `VersionCommand` and `HelpCommand` and replace references to `Command` with `CommandInterface`.

PR #13931, opened 2019-11-22
Use the CommandInterface for constants instead of Command class.

ADmad:
Refs #13930.

The use of `VersionCommand` and `HelpCommand` needs to be addressed. We can either move then under `Console` or add check to use them only if the class exists.

### DIFF
--- a/config/config.php
+++ b/config/config.php
@@ -14,7 +14,7 @@
  * @since         1.1.11
  * @license       https://opensource.org/licenses/mit-license.php MIT License
  */
-$versionFile = file(CORE_PATH . 'VERSION.txt');
+$versionFile = file(dirname(__DIR__) . '/VERSION.txt');
 
 return [
     'Cake.version' => trim(array_pop($versionFile)),
--- a/src/Console/Command.php
+++ b/src/Console/Command.php
@@ -2,62 +2,10 @@
 declare(strict_types=1);
 
 /**
- * CakePHP(tm) : Rapid Development Framework (https://cakephp.org)
- * Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
- *
- * Licensed under The MIT License
- * For full copyright and license information, please see the LICENSE.txt
- * Redistributions of files must retain the above copyright notice.
- *
- * @copyright     Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
- * @link          https://cakephp.org CakePHP(tm) Project
- * @since         3.6.0
- * @license       https://opensource.org/licenses/mit-license.php MIT License
+ * @deprecated 4.0.0 Use Cake\Command\Command instead.
  */
-namespace Cake\Console;
 
-use Cake\Datasource\ModelAwareTrait;
-use Cake\Log\LogTrait;
-use Cake\ORM\Locator\LocatorAwareTrait;
-
-/**
- * Base class for commands using the full stack
- * CakePHP Framework.
- *
- * Includes traits that integrate logging
- * and ORM models to console commands.
- */
-class Command extends BaseCommand
-{
-    use LocatorAwareTrait;
-    use LogTrait;
-    use ModelAwareTrait;
-
-    /**
-     * Constructor
-     *
-     * By default CakePHP will construct command objects when
-     * building the CommandCollection for your application.
-     */
-    public function __construct()
-    {
-        $this->modelFactory('Table', function ($alias) {
-            return $this->getTableLocator()->get($alias);
-        });
-
-        if (isset($this->modelClass)) {
-            $this->loadModel();
-        }
-    }
-
-    /**
-     * Implement this method with your command's logic.
-     *
-     * @param \Cake\Console\Arguments $args The command arguments.
-     * @param \Cake\Console\ConsoleIo $io The console io
-     * @return int|null|void The exit code or null for success
-     */
-    public function execute(Arguments $args, ConsoleIo $io)
-    {
-    }
-}
+class_alias(
+    'Cake\Command\Command',
+    'Cake\Console\Command'
+);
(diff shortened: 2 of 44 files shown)


othercorey:
Maybe not for this PR, but would it make sense to move Command.php to Commands/?  It stands out in the package and would be unused.

ADmad:
I renamed couple of protected methods. Since shells will be removed in future it's better to use "command" in method names unless the method deals with shells only.

ADmad:
@markstory Should I move `Command` class under `Command` namespace as @othercorey suggested and add class alias for it's existing location?

othercorey:
Are you fixing the dependency on Core\Configuration as well with this?

ADmad:
> Are you fixing the dependency on Core\Configuration as well with this?

`composer.json` of the console package does have `cake/core` as a requirement which contains the `Configure` class. 

The problem though is that `Configure::version()` uses `CORE_PATH` constant to generate path to core's `config/config.php` which too uses `CORE_PATH` constant. We could replace the use of constants with relative paths.

othercorey:
I am not sure the usefulness of cake's version number in an independent console though.

ADmad:
I don't have a strong opinion on it. So I have fine we doing what I suggested or moving back the `VersionCommand` under `Command` and loading in conditionally.

othercorey:
I think a common version command would help.  I'm sure we can come up wit something in a PR after this.  I'll let mark decide though.

ADmad:
So I just went ahead and moved the `Command` class and updated code to fetch CakePHP version not rely on the `CORE_PATH` constant.

ADmad:
> updated code to fetch CakePHP version not rely on the CORE_PATH constant.

Still doesn't fix the problem as neither `config/config.php` nor `src/VERSION.txt` are going to be available in the standalone console package.

src/Console/Command.php
```diff
@@ -1,63 +1,10 @@
...
-    public function execute(Arguments $args, ConsoleIo $io)
-    {
-    }
-}
+class_alias(
+    'Cake\Command\Command',
+    'Cake\Console\Command'
+);
+deprecationWarning(
+    'Use Cake\Command\Command instead of Cake\Console\Command.'
```
markstory: This is going to be a noisy deprecation. Can we leave this as a silent alias for the time being? This adds toil to upgrading with little value, as we learned with the ConsoleErrorHandler move uses get tripped up by these.

tests/TestCase/Console/CommandCollectionTest.php
```diff
@@ -272,7 +270,6 @@ public function testAutoDiscoverCore()
         // These have to be strings as ::class uses the local namespace.
         $this->assertSame(RoutesCommand::class, $collection->get('routes'));
         $this->assertSame(SampleShell::class, $collection->get('sample'));
-        $this->assertSame(VersionCommand::class, $collection->get('version'));
```
markstory: Shouldn't the version command still be present?

markstory (commented):
I am not concerned about the VersionCommand being available in standalone usage. It could stay in it's current namespace and not be added to the command collection conditionally.

It would be good to have the HelpCommand though, as it is tedious to reimplement in each application.

tests/TestCase/Console/CommandCollectionTest.php
ADmad: The test is for auto discovered commands. Auto discovery only looks under src/Command so it's not present. 

Anyway I'll move it back as getting version is not possible in standalone packages.

src/Console/Command.php
ADmad: Yup I was thinking that too. Will remove the warning.

markstory: > Should I move Command class under Command namespace as @othercorey suggested and add class alias for it's existing location?

The `Command` class being in the package is only a problem if people try to use it. Could we solve that problem with documentation instead of adding deprecation warnings that impact applications and plugins for a standalone package that may not have widespread adoption.

othercorey: So, as a new user, I followed the documentation on the README as that was the first and only thing I saw.  There is no mention of the Command class there.  However, if you click on the links a the bottom for other things, you can find the Command Objects documentation which starts with explaining using the Command class.

It's awkward it exists but isn't usable, but this is just FYI.  I'm not making an argument either way.

ADmad:
- Moved back `VersionCommand` under `Command`.
- Removed the warning triggered in `src/Console/Command.php`.

ADmad:
The checks job failure is due to pslam crashing. I'll open a ticket for it.

othercorey:
Looks ok to me.

config/config.php
```diff
@@ -14,7 +14,7 @@
  * @since         1.1.11
  * @license       https://opensource.org/licenses/mit-license.php MIT License
  */
-$versionFile = file(CORE_PATH . 'VERSION.txt');
+$versionFile = file(dirname(__DIR__) . '/VERSION.txt');
```
markstory: I was thinking about this problem some more, and will put up a pull request to make `Configure::version()` better.

src/Console/CommandCollection.php
```diff
@@ -203,6 +203,9 @@ protected function resolveNames(array $input): array
             }
         }
 
+        // This entry is created for the `Command` class, which we don't need.
+        unset($out['']);
```
markstory: Could we exclude the base class in CommandScanner so it doesn't show up in other places as well?

config/config.php
othercorey: It would be kind of nice to have a build in version command that you could just configure ... but it doesn't make or break Console imo.

src/Console/CommandCollection.php
ADmad: Done

ADmad:
Anything else to be done here?

PR #13934, opened 2019-11-23
Improve error messages when replies run out.

markstory:
When an console integration test runs out of replies we currently tell you what the replies were but not what question went un-answered. This should make that easier.

Targeting master with this as while it does add a new class it is primarily trying to make testing easier.


src/TestSuite/ConsoleIntegrationTestTrait.php
```diff
@@ -88,6 +89,10 @@ public function exec($command, array $input = [])
 
         try {
             $this->_exitCode = $runner->run($args, $io);
+        } catch (MissingConsoleInputException $e) {
+            $messages = $this->_out->messages();
+            $e->setQuestion($messages[count($messages) - 1]);
```
othercorey: I guess this could fail with 0 messages, but if it isn't an issue in test cases then no worries.

markstory: I can guard against that scenario. It could happen if someone tried hard.

PR #13936, opened 2019-11-23
Improve support for HTTP/2 in Http\Client

### LINKED ISSUE #13922: Curl version 1.1 instead of 2
This is a (multiple allowed):
* [x] bug

* CakePHP Version: 3.7.7
* Platform and Target: PHP 7.2

### What you did
Http client request to a https website - happenend overnight on working code/multiple servers.

### What happened
got an error: cURL Error (56) OpenSSL SSL_read: Success
tracked it down to Cake\Http\Client\Adapter\Curl.php where the $request->getProtocolVersion() function on line 75 returns '1.1' instead of the '2.0' my server provides. 

Also when executing env('SERVER_PROTOCOL') it returns HTTP/2.0 as expected in the getProtocolVersion() of ServerRequest. 

Request.php - getProtocolVersion()  to return protocol 2.0 instead of 1.1

markstory:
Add `protocolVersion` as a constructor option. Previously making HTTP/2 requests was tedious and not exposing this option earlier was an oversight.

Add support for HTTP2 requiring TLS. This mode is preferred over HTTP2 without TLS as most clients and server implementations require TLS for HTTP2.

Fixes #13922

### DIFF
--- a/src/Http/Client/Adapter/Curl.php
+++ b/src/Http/Client/Adapter/Curl.php
@@ -148,11 +148,15 @@ class Curl implements AdapterInterface
                 return CURL_HTTP_VERSION_1_0;
             case '1.1':
                 return CURL_HTTP_VERSION_1_1;
+            case '2':
             case '2.0':
+                if (defined('CURL_HTTP_VERSION_2TLS')) {
+                    return CURL_HTTP_VERSION_2TLS;
+                }
                 if (defined('CURL_HTTP_VERSION_2_0')) {
                     return CURL_HTTP_VERSION_2_0;
                 }
-                throw new HttpException('libcurl 7.33 needed for HTTP 2.0 support');
+                throw new HttpException('libcurl 7.33 or greater required for HTTP/2 support');
         }
 
         return CURL_HTTP_VERSION_NONE;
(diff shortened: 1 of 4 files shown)


src/Http/Client/Adapter/Curl.php
```diff
@@ -148,11 +148,15 @@ protected function getProtocolVersion(Request $request)
                 return CURL_HTTP_VERSION_1_0;
             case '1.1':
                 return CURL_HTTP_VERSION_1_1;
+            case '2':
             case '2.0':
+                if (defined('CURL_HTTP_VERSION_2TLS')) {
```
othercorey: I wish we could warn that it is falling back to non-tls.  It might make a connection and they think tls is configured right.

othercorey: Maybe can update the minimum for 4.x

markstory: HTTP/2 without TLS is very rare, and if this constant is used the request should fail based on the documentation.

markstory: HTTP/2 without TLS is very rare, and if this constant is used the request should fail based on the documentation.

othercorey: I was talking about the constant not being defined.

markstory: We would need to see what version of curl is in the LTS of major distros before we could bump our requirement up. Curl is also an optional dep of the client so we could only enforce the dependency at runtime. We could document the suggested versions though.

PR #13939, opened 2019-11-25
4.x - Removed DateTimeType::_parseValue()

othercorey:
Removed the extensive list of formats we try to parse in `DateTimeType::_parseValue()`.  We can just let `DateTime` constructor parse everything it supports and pull out the values we need on saving.

This is to reduce things before adding more formats like time zone formats.  

We discussed this in https://github.com/cakephp/cakephp/pull/13752, but kept format parsing in place to help test the first set of changes.

### DIFF
--- a/src/Database/Type/DateTimeType.php
+++ b/src/Database/Type/DateTimeType.php
@@ -54,17 +54,6 @@ class DateTimeType extends BaseType
     protected $_format = 'Y-m-d H:i:s';
 
     /**
-     * The DateTime formats allowed by `marshal()`.
-     *
-     * @var array
-     */
-    protected $_marshalFormats = [
-        'Y-m-d H:i:s',
-        'Y-m-d\TH:i:s',
-        'Y-m-d\TH:i:sP',
-    ];
-
-    /**
      * Whether `marshal()` should use locale-aware parser with `_localeMarshalFormat`.
      *
      * @var bool
@@ -257,7 +246,7 @@ class DateTimeType extends BaseType
             } elseif ($isString && $this->_useLocaleMarshal) {
                 return $this->_parseLocaleValue($value);
             } elseif ($isString) {
-                return $this->_parseValue($value);
+                return new $class($value);
             }
         } catch (Exception $e) {
             return null;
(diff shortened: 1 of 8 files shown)


othercorey:
The rationale for this change is you would typically validate user input earlier than this.  Some validation is more lenient already (time zones, etc) and some is more strict (fractional digits, time).  You want to fail there vs here.

src/Database/Type/DateTimeType.php
```diff
@@ -257,7 +245,8 @@ public function marshal($value): ?DateTimeInterface
             } elseif ($isString && $this->_useLocaleMarshal) {
                 return $this->_parseLocaleValue($value);
             } elseif ($isString) {
-                return $this->_parseValue($value);
+                /** @var \DateTimeInterface */
+                return new $class($value);
```
markstory: Won't this also enable users to send things like `+2 minutes -1 days` in API requests? Previously that would fail marshalling but would now pass.

othercorey: Yes.  It would pass at this level.

ADmad: It would, I am unsure if allowing relative time strings when marshalling is a good idea.

othercorey: Does it make a difference?

othercorey:
This change isn't required, but I'm trying to avoid having types with 8 different string formats.

```
    protected $_marshalFormats = [
        'Y-m-d H:i:s.u',
        'Y-m-d H:i:s',
        'Y-m-d H:i:s.uP',
        'Y-m-d H:i:sP',
        'Y-m-d\TH:i:s.u',
        'Y-m-d\TH:i:s',
        'Y-m-d\TH:i:s.uP',
        'Y-m-d\TH:i:sP',
    ];
```

src/Database/Type/DateTimeType.php
ADmad: I am unsure :)   But not having to maintain a long list of valid formats is certainly nice.

markstory: I think it isn't desirable as it changes the behavior of application API endpoints to accept data that is not expected. Is the big list of formats hard to maintain?

markstory: If the big list is hard to maintain could we do a quick pattern check on the input data to ensure it quacks like a datetime string?

othercorey: That is fine.  I just put this up to follow up on the idea.

Looks like I can use a single time zone format to catch: +0200, +02:00 and Europe/Paris.

That'll make the longest marshal list 8 formats (see below).  Shouldn't be too bad for now since it's just processing 1-3 inputs at most likely per request.

Issue #13940, opened 2019-11-25
URL mismatch in POST data in Form Protection Component


chetanvarshney:
This is a (multiple allowed):

* [x] bug

* CakePHP Version: 4.0.0-RC1
* Platform and Target: Windows 10 Mozilla, Localhost

### What you did
I used Form Protection component, on posting form I am getting URL mismatch error.

### What happened
URL mismatch in POST data (expected/cakephp4/users/add but found /users/add)
I am testing on localhost using wamp server.
There is no issue with Security Component.

### What you expected to happen
I have fixed issue by editing FormProtectionComponent.php (in forked project)
I will send pull request now.

ADmad:
Closing as PR is open.

PR #13941, opened 2019-11-25
Fixed URL mismatch in POST data in FormProtectionComponent.php

chetanvarshney:
I have fixed URL mismatch in POST data issue in FormProtectionComponent.php, which was not working on wamp localhost.
If you look in SecurityComponent.php at line number 298 a valid URL used
but in FormProtectionComponent.php it is incorrect.

Fixed issue [#13940](https://github.com/cakephp/cakephp/issues/13940)


ADmad:
Thanks, for the fix.

chetanvarshney:
Happy to contribute.
Thanks

PR #13942, opened 2019-11-26
Make Configure::version() more resiliant

markstory:
In addition to checking for the config/config.php file we should also honour the return type even when we don't know which version of CakePHP is being run.

### DIFF
--- a/tests/TestCase/Core/ConfigureTest.php
+++ b/tests/TestCase/Core/ConfigureTest.php
@@ -464,8 +464,14 @@ class ConfigureTest extends TestCase
      */
     public function testVersion()
     {
-        $result = Configure::version();
-        $this->assertTrue(version_compare($result, '4.0', '>='));
+        $original = Configure::version();
+        $this->assertTrue(version_compare($original, '4.0', '>='));
+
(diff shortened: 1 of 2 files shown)


tests/TestCase/Core/ConfigureTest.php
```diff
@@ -464,8 +464,14 @@ public function testStoreAndRestoreWithData()
...
     {
-        $result = Configure::version();
-        $this->assertTrue(version_compare($result, '4.0', '>='));
+        $original = Configure::version();
+        $this->assertTrue(version_compare($original, '4.0', '>='));
+
+        Configure::write('Cake.version', 'banana');
+        $this->assertSame('banana', Configure::version('Cake.version'));
+
+        Configure::delete('Cake.version');
```
othercorey: So, if you delete Cake.version or don't have it in your config, it will force overwrite your config with config/config.php values?

markstory: Kind of, you could define `Cake.version` to skip the need to load the file. However if the value is undefined/null the config/config.php file will be loaded.

PR #13943, opened 2019-11-26
4.x - Added DateTimeTimezoneType

othercorey:
This type is meant to reflect the `timestamptz` and `timestamp with time zone` data types in PostgreSQL.  It could potentially reflect the `datetimeoffset` type in SqlServer as well.

The key differences between DateTimeTimezoneType and DateTimeFractionalType are:

* Timezone type allows time zone in the marshal string format
* Timezone type formats database string with time zone offset

To support use cases where users want to keep the db-specific time zone, I added `DateTimeType::setKeepDatabaseTimeZone()` which simply sets a flag that wraps existing behavior of converting to default time zone.

I also created `DateTimeType::setDatabaseTimezone()` which should supersede `setTimezone()` eventually as the clearer name.

The unit tests are probably had to follow, but they are essentially the same as the fractional type test with some different time zones in the strings and a couple conversion changes.


src/Database/Type/DateTimeType.php
```diff
@@ -179,18 +194,20 @@ public function toPHP($value, DriverInterface $driver)
...
                 return null;
             }
-            $instance = $instance->modify($value);
+            $instance = new $class($value, $this->dbTimezone);
         }
 
-        if ($instance->getTimezone()->getName() !== date_default_timezone_get()) {
-            $instance = $instance->setTimezone(new DateTimeZone(date_default_timezone_get()));
+        if (!$this->keepDatabaseTimezone) {
+            if ($instance->getTimezone()->getName() !== $this->defaultTimezone->getName()) {
```
ADmad: This nesting can be avoided by hoisting the check in outer `if` block.

othercorey: I will change this, but I'm wondering if nested checks can look cleaner than multi-line if checks sometimes.

```
        if (
            !$this->keepDatabaseTimezone &&
            $instance->getTimezone()->getName() !== $this->defaultTimezone->getName()
        ) {
```

ADmad: Asthetics are not really a concern here as the nested `if` creates an unnecessary code branch.

othercorey: That's pretty pedantic lol.

src/Database/Type/DateTimeType.php
```diff
@@ -179,18 +194,21 @@ public function toPHP($value, DriverInterface $driver)
...
+        $class = $this->_className;
         if (is_int($value)) {
-            $instance = $instance->setTimestamp($value);
+            $instance = new $class('@' . $value);
         } else {
             if (strpos($value, '0000-00-00') === 0) {
                 return null;
             }
-            $instance = $instance->modify($value);
+            $instance = new $class($value, $this->dbTimezone);
```
othercorey: Caching off the DateTimeZone instance removes the overhead of creating new DateTime instances here.

src/Database/Schema/PostgresSchema.php
```diff
@@ -382,8 +389,10 @@ public function columnSql(TableSchema $schema, string $name): string
             TableSchema::TYPE_TIME => ' TIME',
             TableSchema::TYPE_DATETIME => ' TIMESTAMP',
             TableSchema::TYPE_DATETIME_FRACTIONAL => ' TIMESTAMP',
+            TableSchema::TYPE_DATETIME_TIMEZONE => ' TIMESTAMPTZ',
```
markstory: Should there be two abstract type if they share a concrete type in the only database that supports timezones right now? I think having a single type is ok.

src/Database/Schema/TableSchemaInterface.php
```diff
@@ -79,6 +86,13 @@ interface TableSchemaInterface extends SchemaInterface
      */
     public const TYPE_TIMESTAMP_FRACTIONAL = 'timestampfractional';
 
+    /**
+     * Timestamp with time zone column type
+     *
+     * @var string
+     */
+    public const TYPE_TIMESTAMP_TIMEZONE = 'timestamptimezone';
```
markstory: Given that MySQL and SQLite don't have a way of supporting this type should we fallback to a timezone-less storage column or raise an exception indicating that a fixture schema object cannot be converted into this dialect of SQL?

src/Database/Type/DateTimeType.php
```diff
@@ -144,33 +151,41 @@ public function toDatabase($value, DriverInterface $driver): ?string
...
+     * This is the time zone used when converting database strings to DateTime
+     * instances and converting DateTime instances to database strings.
+     *
+     * @see DateTimeType::setKeepDatabaseTimezone
      *
      * @param string|\DateTimeZone|null $timezone Database timezone.
      * @return $this
      */
-    public function setTimezone($timezone)
+    public function setDatabaseTimezone($timezone)
```
markstory: Why are we adding an alias now?

src/Database/Type/DateTimeType.php
```diff
@@ -200,6 +218,26 @@ public function toPHP($value, DriverInterface $driver)
...
     }
 
+    /**
+     * Set whether DateTime object created from database string is converted
+     * to default time zone.
+     *
+     * If your database date times are in a specific time zone that you want
+     * to keep in the DateTime instance then set this to true.
+     *
+     * Defaults to false. Time zone is converted to default time zone.
```
markstory:
```suggestion
     * When false, datetime timezones are converted to default time zone. This is default behavior.
```

src/Database/Type/DateTimeType.php
```diff
@@ -212,22 +250,7 @@ public function manyToPHP(array $values, array $fields, DriverInterface $driver)
...
-            if ($instance->getTimezone()->getName() !== date_default_timezone_get()) {
-                $instance = $instance->setTimezone(new DateTimeZone(date_default_timezone_get()));
-            }
-
-            if ($this->setToDateStart) {
-                $instance = $instance->setTime(0, 0, 0);
-            }
-
-            $values[$field] = $instance;
+            $values[$field] = $this->toPHP($values[$field], $driver);
```
markstory: I think the goal with the duplication here was to avoid the overhead of additional method calls. While small these add up in aggregate and the `manyToPHP()` methods are intended to be an optimization.

src/Database/Type/DateTimeType.php
othercorey: I want to clarify what this setting does.  It is the time zone used to interpret db values only.  In toPHP() there is logic that converts the time zone.  This does not affect that. It also does not affect marshaling.

src/Database/Type/DateTimeType.php
othercorey: I can re-build it.  I changed it to see what the team thought.

src/Database/Schema/PostgresSchema.php
othercorey: In postgres, `TIMESTAMPTZ` is an alias for `TIMESTAMP WITH TIME ZONE`.  It lets you do `TIMESTAMPTZ(3)` instead of `TIMESTAMP(3) WITH TIME ZONE`.  They keep the long name for compatibility with the SQL Standard names.

othercorey: Oh, nevermind, you're asking about `TYPE_DATETIME_TIMEZONE`.  There is one type that would use it: `datetimeoffset` in SqlServer.

https://docs.microsoft.com/en-us/sql/t-sql/data-types/datetimeoffset-transact-sql?view=sql-server-ver15

I added it for completion, but you're correct no concrete types map to it right now.

src/Database/Schema/TableSchemaInterface.php
othercorey: Let me see if there's a consistent way to do that.

othercorey: This would require throwing an exception from BaseSchema::columnSql().  Nothing else checks or throws an error here.  The TestFixture wrapper will catch an exception (which it shouldn't) when the create sql fails to run and log it.

othercorey: Added default behavior for mysql, sqlserver and sqlite so fixtures can use these types without unexpected results.

src/Database/Type/DateTimeType.php
```diff
@@ -200,6 +218,27 @@ public function toPHP($value, DriverInterface $driver)
...
+     * When false, datetime timezones are converted to default time zone.
+     * This is default behavior.
+     *
+     * @param bool $keep If true, database time zone is kept when converting
+     *      to DateTime instances.
+     * @return $this
+     */
+    public function setKeepDatabaseTimezone(bool $keep)
+    {
+        $this->keepDatabaseTimezone = $keep;
```
markstory: What do non-timezone aware type mappers do when this is set to true? Should the timezone type be the class that implements this instead?

othercorey: They do the same thing.  Here's the most complex example:

1. MySQL is configured for 'Europe/London'
2. PHP is configured for 'America/Toronto'
3. `DateTimeType::setDatabaseTimezone()` is set to 'Europe/London'
4. `DateTimeType::toPHP()` converts a value from the database

With the default of `keepDatabaseTimezone=false` `toPHP` will return the `Time` instance converted to 'America/London'.  With `true`, it will leave the `Time` instance as-is in 'Europe/London'.

othercorey: This also allows time zone conversion to happen on the DB-side for databases that support time zone conversion that return a time zone in the string.  You'll get the instance in the time zone specified by the string.

Issue #13945, opened 2019-11-26
Cannot change save handler when session is active (PHP 7.2 / Cake 2.10.x)


pascalwoerde:
* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 2.10.19
* Platform and Target: PHP 7.2

### What you did
While upgrading to PHP 7.2 with Cake 2.10.x the well known warning 'Cannot change save handler when session is active' comes from the CakeSession.

### What happened
Warning (2)
: session_set_save_handler(): Cannot change save handler when session is active [
APP/Vendor/cakephp/cakephp/lib/Cake/Model/Datasource/CakeSession.php
, line
598
]

### What you expected to happen
A working application without the above warning.

pascalwoerde:
Further investigation shows that the internal method CakeSession::_getHandler creates a new SessionHandler which by default for example the DatabaseSession calls the registry which creates on his own some dependencies which result in recursive call and a duplicate session_set_save_handler call. I will open a PR with the fix soon.

PR #13947, opened 2019-11-26
Fix duplicate session_set_save_handler call from handler instantiation.

pascalwoerde:
Will fix issue #13945.


markstory:
If the session is active does the configured session handler still work, or will the previous session handler be used?

pascalwoerde:
If the session is active it is not possible to set (or override) the current save handler. When we want to do that we have to destroy the current session and then we are able to override the save handler, but that is not what we want to achieve in this case. Only the first handler will be set and used for sessions. Before php 7.2 this works exactly the same, except the warning which will be thrown since php 7.2.

Issue #13952, opened 2019-11-28
Validator::errors() is ambiguous


othercorey:
We have `Validator::errors()` which validates the passed in data and returns the validation result + error messages.  This name suggests that some validation already occurred as you normally see with other `getErrors()` methods.

The docblock tries to explain it which suggests that this is ambiguously named.

>     * Returns an array of fields that have failed validation. On the current model. This method will
>     * actually run validation rules over data, not just return the messages.

As a user, I'd expect something like `Validator::validateData()` or just `Validator::validate()`.

dereuromark:
Also `Validator::result()` or alike could work.

markstory:
Closing as there is a pull request open.

PR #13953, opened 2019-11-28
3.9 - Renamed Validator::errors() to Validator::validate()

othercorey:
A suggestion for fixing https://github.com/cakephp/cakephp/issues/13952.

I think this tells the user what is happening and you still expect results back.  This would be nice to have available for the 4.0 transition.

### DIFF
--- a/src/Validation/Validator.php
+++ b/src/Validation/Validator.php
@@ -160,14 +160,26 @@ class Validator implements ArrayAccess, IteratorAggregate, Countable
     }
 
     /**
-     * Returns an array of fields that have failed validation. On the current model. This method will
-     * actually run validation rules over data, not just return the messages.
+     * Validates and returns an array of failed fields and their error messages.
      *
      * @param array $data The data to be checked for errors
      * @param bool $newRecord whether the data to be validated is new or to be updated.
-     * @return array Array of invalid fields
+     * @return array Array of failed fields
+     * @deprecated 3.9.0 Renamed to validate()
      */
     public function errors(array $data, $newRecord = true)
+    {
+        return $this->validate($data, $newRecord);
+    }
+
+    /**
+     * Validates and returns an array of failed fields and their error messages.
+     *
+     * @param array $data The data to be checked for errors
+     * @param bool $newRecord whether the data to be validated is new or to be updated.
+     * @return array Array of failed fields
+     */
+    public function validate(array $data, $newRecord = true)
     {
         $errors = [];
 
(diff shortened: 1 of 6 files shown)


src/Validation/Validator.php
```diff
@@ -160,14 +160,26 @@ public function __construct()
...
+    }
+
+    /**
+     * Validates and returns an array of failed fields and their error messages.
+     *
+     * @param array $data The data to be checked for errors
+     * @param bool $newRecord whether the data to be validated is new or to be updated.
+     * @return array Array of failed fields
+     */
+    public function validate(array $data, $newRecord = true)
```
ADmad: Do remember to add typehints after this is merged into 4.x :)

othercorey: I can do that.

ADmad (approved):

src/Validation/Validator.php
```diff
@@ -160,14 +160,26 @@ public function __construct()
...
      *
      * @param array $data The data to be checked for errors
      * @param bool $newRecord whether the data to be validated is new or to be updated.
-     * @return array Array of invalid fields
+     * @return array Array of failed fields
+     * @deprecated 3.9.0 Renamed to validate()
      */
     public function errors(array $data, $newRecord = true)
+    {
+        return $this->validate($data, $newRecord);
```
dereuromark: This probably should trigger deprecation notes.
If your code extends this method in plugins or project, it will otherwise silently not run anymore right?

saeideng: deprecation warning can be add to 4.x 
and then remove the old methods in cake 5

dereuromark: Do we also backport to 3.9? Or keep as is?

saeideng: we should not add more deprecation warning to cake 3.x ( when no needed ) 
because we are in RC(cake 4) level and we will see final version soon 
so people should not force to upgrade their apps to 3.9 IMO ( that there is not yet) 
I think 3.8 is enough 
but 3.9 can be receive backport features from 4.x ( until 3.9 get out )

having this method ( `errors` ) does not heart code and can be kept as an alias to new method

saeideng: if we merge this now , this method must throw deprecation warning in cake 4

src/Validation/Validator.php
```diff
@@ -160,14 +160,26 @@ public function __construct()
...
     /**
-     * Returns an array of fields that have failed validation. On the current model. This method will
-     * actually run validation rules over data, not just return the messages.
+     * Validates and returns an array of failed fields and their error messages.
      *
      * @param array $data The data to be checked for errors
      * @param bool $newRecord whether the data to be validated is new or to be updated.
-     * @return array Array of invalid fields
+     * @return array Array of failed fields
+     * @deprecated 3.9.0 Renamed to validate()
```
saeideng: `, and will remove in future`

markstory: This is a method rename we could add a rector for to the 40 ruleset.

markstory (approved):
Looks good to me.

PR #13954, opened 2019-11-28
Add connection id to the queries log

### LINKED ISSUE #13768: Log connection id to queries log or use different log files
This is a (multiple allowed):

* [ ] bug
* [x] enhancement

* CakePHP Version: 3.8.5.
* Platform and Target: Linux, Apache, MySQL.

### What you did
Enabled queries log.

### What happened
All queries are logged to the same file. But it is quite difficult to understand when multiple database connections are used and it is impossible to debug database transactions as it is not clear on what connection the query is executed.

I would expect each database connection to use it's own query log or if one log file is used then to record also the connection id.

bancer:
Fixes #13768

### DIFF
--- a/src/Database/Connection.php
+++ b/src/Database/Connection.php
@@ -970,7 +970,7 @@ class Connection implements ConnectionInterface
     }
 
     /**
-     * Gets the logger object
+     * Gets the database connection id.
      *
      * @return int|null
      */
--- a/src/Database/Connection.php
+++ b/src/Database/Connection.php
@@ -45,6 +45,13 @@ class Connection implements ConnectionInterface
     protected $_config;
 
     /**
+     * The connection id.
+     *
+     * @var int|null
+     */
+    protected $_connectionId;
+
+    /**
      * Driver object, responsible for creating the real connection
      * and provide specific SQL dialect.
      *
@@ -956,9 +963,24 @@ class Connection implements ConnectionInterface
     {
         if ($this->_logger === null) {
             $this->_logger = new QueryLogger();
+            $this->_logger->setConnectionId($this->getConnectionId());
         }
 
         return $this->_logger;
+    }
+
+    /**
+     * Gets the logger object
+     *
+     * @return \Cake\Database\Log\QueryLogger logger instance
+     */
+    public function getConnectionId()
+    {
+        if ($this->_connectionId === null) {
+            $this->_connectionId = $this->_driver->getConnectionId();
+        }
+
+        return $this->_connectionId;
     }
 
     /**
--- a/src/Database/Connection.php
+++ b/src/Database/Connection.php
@@ -963,7 +963,7 @@ class Connection implements ConnectionInterface
     {
         if ($this->_logger === null) {
             $this->_logger = new QueryLogger();
-            $this->_logger->setConnectionId($this->getConnectionId());
+            $this->_logger->addMetaData(['connection_id' => $this->getConnectionId()]);
         }
 
         return $this->_logger;
--- a/src/Database/Connection.php
+++ b/src/Database/Connection.php
@@ -972,7 +972,7 @@ class Connection implements ConnectionInterface
     /**
      * Gets the logger object
      *
-     * @return \Cake\Database\Log\QueryLogger logger instance
+     * @return int|null
      */
     public function getConnectionId()
     {
--- a/src/Database/Driver/Mysql.php
+++ b/src/Database/Driver/Mysql.php
@@ -176,4 +176,17 @@ class Mysql extends Driver
 
         return $this->_supportsNativeJson = version_compare($this->_version, '5.7.0', '>=');
     }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function getConnectionId()
+    {
+        $PDOStatement = $this->getConnection()->query('SELECT CONNECTION_ID()');
+        $row = $PDOStatement->fetch();
+        if ($row === false) {
+            return null;
+        }
+        return (int)$row[0];
+    }
 }
--- a/src/Database/Driver/Mysql.php
+++ b/src/Database/Driver/Mysql.php
@@ -187,6 +187,7 @@ class Mysql extends Driver
         if ($row === false) {
             return null;
         }
+
         return (int)$row[0];
     }
 }
--- a/src/Database/Log/QueryLogger.php
+++ b/src/Database/Log/QueryLogger.php
@@ -25,21 +25,21 @@ use Cake\Log\Log;
 class QueryLogger
 {
     /**
-     * The database connection id.
+     * Meta data.
      *
-     * @var int|null
+     * @var array
      */
-    protected $_connectionId;
+    protected $_metaData = [];
 
     /**
-     * Sets the database connection id.
+     * Adds meta data such as a connection id.
      *
-     * @param int|null $connectionId Database connection id.
+     * @param array $data Meta data.
      * @return void
      */
-    public function setConnectionId($connectionId)
+    public function addMetaData($data)
     {
-        $this->_connectionId = $connectionId;
+        $this->_metaData += $data;
     }
 
     /**
@@ -66,8 +66,8 @@ class QueryLogger
     protected function _log($query)
     {
         $message = $query->__toString();
-        if ($this->_connectionId !== null) {
-            $message = "connection_id={$this->_connectionId} $message";
+        if (isset($this->_metaData['connection_id'])) {
+            $message = "connection_id={$this->_metaData['connection_id']} $message";
         }
         Log::write('debug', $message, ['queriesLog']);
     }
--- a/src/Database/Log/QueryLogger.php
+++ b/src/Database/Log/QueryLogger.php
@@ -25,6 +25,24 @@ use Cake\Log\Log;
 class QueryLogger
 {
     /**
+     * The connection id.
+     *
+     * @var int|null
+     */
+    protected $_connectionId;
+
+    /**
+     * Gets the logger object
+     *
+     * @param int|null $connectionId Database connection id.
+     * @return void
+     */
+    public function setConnectionId($connectionId)
+    {
+        $this->_connectionId = $connectionId;
+    }
+
+    /**
      * Writes a LoggedQuery into a log
      *
      * @param \Cake\Database\Log\LoggedQuery $query to be written in log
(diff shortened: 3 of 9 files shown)


bancer:
Any idea why sqlite and pgsql tests fail? Is it something wrong with my code or should `testAutomaticReconnect` be adjusted?

src/Database/Connection.php
```diff
@@ -956,11 +963,26 @@ public function getLogger()
...
 
+    /**
+     * Gets the database connection id.
+     *
+     * @return int|null
+     */
+    public function getConnectionId()
+    {
+        if ($this->_connectionId === null) {
+            $this->_connectionId = $this->_driver->getConnectionId();
```
othercorey: I don't think this should query on get.

src/Database/Driver/Mysql.php
```diff
@@ -176,4 +176,18 @@ public function supportsNativeJson()
 
         return $this->_supportsNativeJson = version_compare($this->_version, '5.7.0', '>=');
     }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function getConnectionId()
```
othercorey: Sorry, but I don't think querying connection id just for mysql is helpful here.

I'd rather see the query logger use the connection name. It is safe and easy to configure.

bancer: Different connections may have the same connection name.

src/Database/Connection.php
bancer: There is no other way to get the connection id per my knowledge.

src/Database/Driver/Mysql.php
bancer: F.ex. if you have a database transaction that executes several SQL statements and the same URL is requested at the same time from two or more different clients it is impossible to determine to what transaction (connection) belongs each SQL statement without logging the transaction id.

othercorey: I see.  I would suggest a client side id then.  This only works for mysql.

src/Database/Connection.php
othercorey: This should not run the sql query in a getter.  The id should be set once.

src/Database/Driver/Mysql.php
bancer: I am not sure how the client id would help if the requests are send by two different tabs opened in the same browser for example.

othercorey: client-side for the db so cakephp.

bancer: `getConnectionId()` can be overridden in driver classes. I believe that Postgres and SQLServer provide ways to get the connection id or something similar.

src/Database/Connection.php
bancer: It is not clear what you mean. `SELECT CONNECTION_ID()` query is executed only once per connection.

markstory:
Do you generally want to track these connection ids back to the active database connection, or are you aiming to understand the query logs when multiple servers are concurrently querying the database, and need to separate which server/thread/request is doing what operations?

othercorey:
> Do you generally want to track these connection ids back to the active database connection, or are you aiming to understand the query logs when multiple servers are concurrently querying the database, and need to separate which server/thread/request is doing what operations?

My understanding is he just wants to track multiple php processes on the same server making the same query to the same db.

I'm suggesting a client-side id is generated and set for the logger then. That would give you a per-instance query log.

bancer:
> Do you generally want to track these connection ids back to the active database connection, or are you aiming to understand the query logs when multiple servers are concurrently querying the database, and need to separate which server/thread/request is doing what operations?

In our configuration we have multiple databases - one per client (thousands in total) and one common database. What is needed is to be able to clearly see on what database the queries are executed and to what database transactions the queries belong. Writing each connection statements to its own log file (desirable but not essential) would help to separate queries to different databases but it still would not allow to see to what database transaction (in the same database) the queries belong (essential). Writing the connection id solves the major problem - makes it easier to debug the concurrent database transactions. Also that's what the native MySQL logger does - writes the connection id together with the statement and the time. One HTTP request often uses 2-3 database connections to different databases.

> I'm suggesting a client-side id is generated and set for the logger then. That would give you a per-instance query log.

If I understand correctly you suggest the same what MySQL already does - generates unique connection id. I see no point in re-implementing the same in PHP what is already implemented on the database level. Perhaps it would make sense to do that for SQLite or some other database drivers if it does not provide any way to get the connection id.

src/Database/Log/QueryLogger.php
```diff
@@ -47,7 +65,8 @@ public function log(LoggedQuery $query)
      */
     protected function _log($query)
     {
-        Log::write('debug', $query, ['queriesLog']);
+        $message = "connection_id={$this->_connectionId} " . $query->__toString();
```
markstory: If the connection id is null could this be skipped?

src/Database/Connection.php
```diff
@@ -956,11 +963,26 @@ public function getLogger()
     {
         if ($this->_logger === null) {
             $this->_logger = new QueryLogger();
+            $this->_logger->setConnectionId($this->getConnectionId());
```
markstory: This is a very specific solution to a more general problem of enriching query logs with additional metadata.

Could we generalize this with a method like `addMetadata()` that takes a key:value array of data that is added to emitted logs? This might also let us refactoring how query timing is collected and included in logs.

Connection configuration could define a closure that prepares the logging metadata the first time it is used. That would let us avoid adding methods to Driver that have only one implementation.

bancer: Do you mean `$this->_logger->addMetadata(['connection_id' => $this->getConnectionId()]);`?

src/Database/Log/QueryLogger.php
bancer: Of course.

src/Database/Connection.php
bancer: That would be possible but in my opinion that is a wrong thing to do. Usually when an array of items is passed as a method parameter (or a method has many parameters that is practically the same) that is considered as code smell in OOP and refactoring guru M.Fowler suggests to use [Introduce Parameter Object](https://refactoring.guru/introduce-parameter-object) pattern instead. I think it is better for an object to have a list of basic type properties than one array type property (which would contain a list of mixed type elements - int, string, etc.).

othercorey:
> If I understand correctly you suggest the same what MySQL already does - generates unique connection id. I see no point in re-implementing the same in PHP what is already implemented on the database level. Perhaps it would make sense to do that for SQLite or some other database drivers if it does not provide any way to get the connection id.

You seem to be missing the point that this solution only works for mysql.  It also adds the cost of a query for every connection.

A simple id can be generated on the client to make grouping queries in the log easier to solve your problem.  This would work for every db that cake supports without much cost.

bancer:
@othercorey I am afraid that may be confusing if the connection id in the logs would be not the same connection id as returned by `SELECT CONNECTION_ID()`. I think anybody who works with MySQL would expect a real connection id to be in the log but not the fake one. For example if a transaction is terminated without `COMMIT;` or `ROLLBACK;` for some reason (what I unfortunately experienced in the past) it would still be active (hanging) in MySQL and possibly blocking some tables and will be present in `information_schema.innodb_trx` table. I think it would be quite useful to be able to find the logs for such transaction. That would be impossible if a fake transaction id would be logged.

It is true that it is an extra query to the database but only if the query log is enabled and that query itself is very fast.

bancer:
It looks like Postgres [has](https://stackoverflow.com/a/5070953/316041) something similar - `pg_backend_pid()`.

othercorey:
I understand.  I feel this needs to work for everything or be something that the user can optionally add.

src/Database/Connection.php
markstory: I don't see how a key value list of logging attributes creates a smell. It would be intended as a way for applications to add application relevant attributes to query logs. The types could be constrained to `array{string: string}`.

ADmad: `$context` argument of `LoggerInterface::debug()` is also an array :wink:. (In 4.x `QueryLogger` is an `LoggerInterface` instance.)

bancer: Whenever I see an associative array passed as a parameter that smells bad. Sadly but CakePHP has a lot of that. I believe that can be called [data clump](https://en.wikipedia.org/wiki/Data_clump). It is much better to use objects instead. The benefits are briefly explained in [this](https://stackoverflow.com/a/3862899/316041) post.

othercorey: This isn't the PR to discuss overhauling cake's options architecture.

However, some of what you're talking about is due to the language itself some of it is historical.   One of the things that php lacks is nested classes.

The options aren't really data clumps per say.  They are structured options which aren't passed around.  It's similar to creating a function in javascript.

However, insisting that cake be re-designed to avoid aligning your changes with current convention is just going to result in a stale PR.

bancer: It's not a big deal to change that as asked above and I will do that. But I really have difficulties to understand why would you consider `$this->_logger->addMetadata(['connection_id' => $this->getConnectionId()]);` to be better than `$this->_logger->setConnectionId($this->getConnectionId());`. 
1. It allows typing errors like `$this->_logger->addMetadata(['conection_id' => $this->getConnectionId()]);` that are quite difficult to debug,
2. It allows to set keys that are not handled by the logger class like `$this->_logger->addMetadata(['useless' => true]);`,
3. It does not allow IDEs to autocomplete the name of the array key. There is no such problem with `setConnectionId()` implementation.
4. It will not allow to enforce a different data type for each of array elements, f.ex. 'connection_id' to be int and 'duration' to be a float. There would be no such problem with `function setConnectionId(int $connectionId)` in 4.x.

markstory: For the `addMedata` method I proposed, mistyped keys would still be included in the log message as log messages would format all metadata keys into emitted messages.

I'm not sure the type safety is required as all the values will be coerced into strings when added to log messages.

bancer:
@othercorey This is a pull request to 3.x, not to 4.x.

saeideng:
cake 4.x now released , so you have to put new features to cake 4.x first 
and then you can back port them to 3.next

bancer:
@saeideng I have little interest in doing that. We are not going to use 4.x any time soon.

saeideng:
@bancer yes , but how we could add new features to 3.x without having them in 4.x ?
when a feature added to core you are not only person that want to use it , it will use by other person that expected to see it in 4.x too 

also adding typehints to new features from cake 3 => 4 could  be forgot
add fix it is another trouble for BC reasons
or even could be unexpected error on cake 4 for any reason 

so we can not accept new features on cake3 , unless it already exist on cake 4

see wiki 
https://github.com/cakephp/cakephp/wiki

>We will do a 3.9 release to add additional backports

othercorey:
Closing in favor of alternative PR.

PR #13955, opened 2019-11-29
Fix query string set to uri.

dereuromark:
Bug:
```php
		$server = [
			'REQUEST_URI' => '/posts/home',
		];
		$query = [
			'foo' => 'bar',
			'key' => 'val',
		];

		$request = ServerRequestFactory::fromGlobals(
			$server,
			$query
		);
		dd($request->getRequestTarget()); // Just '/posts/home'
```

Same for
```php
        $request = new ServerRequest([
            'url' => '/posts/home',
            'query' => $query,
        ]);
```

Seems like it never properly sets the uri->query part here.
So when you remove deprecations in Cake3 from 
`request->here()` to `request->getRequestTarget()` as outlined in the deprecation messages thrown, all breaks.

This here could be a fix, it seems to work.
Might need some follow up - I am happy to pass this on at this point.

### DIFF
--- a/src/Http/ServerRequest.php
+++ b/src/Http/ServerRequest.php
@@ -354,6 +354,9 @@ class ServerRequest implements ArrayAccess, ServerRequestInterface
         $this->url = substr($uri->getPath(), 1);
         $this->here = $this->base . '/' . $this->url;
 
+        $this->query = $this->_processGet($config['query'], $querystr);
+        $this->uri = $this->uri->withQuery(http_build_query($this->query, null, '&', PHP_QUERY_RFC3986));
+
         if (isset($config['input'])) {
             $stream = new Stream('php://memory', 'rw');
             $stream->write($config['input']);
@@ -365,7 +368,6 @@ class ServerRequest implements ArrayAccess, ServerRequestInterface
 
         $config['post'] = $this->_processPost($config['post']);
         $this->data = $this->_processFiles($config['post'], $config['files']);
-        $this->query = $this->_processGet($config['query'], $querystr);
         $this->params = $config['params'];
         $this->session = $config['session'];
     }
@@ -423,6 +425,9 @@ class ServerRequest implements ArrayAccess, ServerRequestInterface
             parse_str($queryString, $queryArgs);
             $query += $queryArgs;
         }
+
+        parse_str($this->uri->getQuery(), $queryArgs);
+        $query += $queryArgs;
 
         return $query;
     }
--- a/src/TestSuite/IntegrationTestTrait.php
+++ b/src/TestSuite/IntegrationTestTrait.php
@@ -752,7 +752,8 @@ trait IntegrationTestTrait
     {
         $uri = new Uri($url);
         $path = $uri->getPath();
-        $query = $uri->getQuery();
+        parse_str($uri->getQuery(), $query);
+        $query = http_build_query($query, null, '&', PHP_QUERY_RFC3986);
 
         $hostData = [];
         if ($uri->getHost()) {
--- a/tests/TestCase/Controller/Component/RequestHandlerComponentTest.php
+++ b/tests/TestCase/Controller/Component/RequestHandlerComponentTest.php
@@ -1250,7 +1250,7 @@ XML;
                 $this->Controller->response
             );
             $data = json_decode($response, true);
-            $this->assertEquals('/request_action/params_pass', $data['here']);
+            $this->assertEquals('/request_action/params_pass?a=b&x=y%3Fish', $data['here']);
 
             $response = $this->RequestHandler->beforeRedirect(
                 $event,
--- a/tests/TestCase/Http/ServerRequestTest.php
+++ b/tests/TestCase/Http/ServerRequestTest.php
@@ -155,7 +155,7 @@ class ServerRequestTest extends TestCase
         ];
         $request = new ServerRequest($data);
         $this->assertEquals($request->getQueryParams(), $data['query']);
-        $this->assertEquals('/some/path', $request->getRequestTarget());
+        $this->assertEquals('/some/path?one=param&two=banana', $request->getRequestTarget());
     }
 
     /**


ADmad:
I have pushed a new commit to fix the issue. There was reverse issue too where `ServerRequest::$query` was not populated based on value of `Uri::getQuery()`.

The failing test case is due to issue with generation of Uri instance based on environment params.

For this test case 

https://github.com/cakephp/cakephp/blob/e16d98815b14e3e482f4b0e9a0c5c119a6388e9c/tests/TestCase/Http/ServerRequestTest.php#L2389-L2414

`uri->getQuery()` returns `'/posts/add&'`.  If the test does actually mimic how the environment values are populated then `ServerRequestFactory::marshalUriFromServer()` need to be updated to account for this wackiness and not populate the uri instance with value of `QUERY_STRING` as query string.

ADmad:
Aargh.. The change also cause various integration tests to fail.

@dereuromark Seems you have opened a can of worms :stuck_out_tongue:.

dereuromark:
Thanks for checking it out.
Better we fix it now before we release now 3.9 or 4.0 tags :)

ADmad:
I have updated the patch and now all tests pass expect the ones for nginx.

But I am vary of this change going into a patch release. The flip floping of query params between array and string representation could lead to unexpected results in real world scenarios.

As you can see I had make change in `IntegrationTestTrait`'s url handling code. That is because if you create a `Uri` instance with argument `path?foo=/` then query string return by `Uri::getQuery()` will be `foo=/` while if you generate query string using `http_build_query(['foo' => '/'])` the `/` will be encoded and you get `foo=%2F`. The docblock of `QueryInterface::getQuery()` does that the value returned MUST be percent-encoded as per RFC 3986, and `Uri` also seems to have some regex based code for encoding chars in query string, so I am unsure why it returns the un-encoded `/`. Someone more proficient with reading ABNF forms can check the relevant sections of RFC 3986 :)

Also when one populates environment params as well the `query` option like the example given in initial post, you are essentially adding two sources of truth. Supporting the mix of options is probably necessary for BC but in 4.x I would rather get rid of this ambiguity.

src/Http/ServerRequest.php
```diff
@@ -354,6 +354,9 @@ protected function _setConfig($config)
         $this->url = substr($uri->getPath(), 1);
         $this->here = $this->base . '/' . $this->url;
 
+        $this->query = $this->_processGet($config['query'], $querystr);
```
markstory: Do we still need to pass `querystr` here? If the querystr has been set into the uri object, we'll do some duplicate work now.

src/Http/ServerRequest.php
```diff
@@ -424,6 +426,9 @@ protected function _processGet($query, $queryString = '')
             $query += $queryArgs;
         }
 
+        parse_str($this->uri->getQuery(), $queryArgs);
```
markstory: I think we should be able to remove the block that operates on `$queryString` here as the already has the query set, so we'd be parsing it twice now.

src/TestSuite/IntegrationTestTrait.php
```diff
@@ -752,7 +752,8 @@ protected function _url($url)
     {
         $uri = new Uri($url);
         $path = $uri->getPath();
-        $query = $uri->getQuery();
+        parse_str($uri->getQuery(), $query);
+        $query = http_build_query($query, null, '&', PHP_QUERY_RFC3986);
```
markstory: Why are we parsing and then re-encoding the query string now?

src/Http/ServerRequest.php
ADmad: At this point `$this->uri` object won't have the required query string set if the URL was specified using `url` config in constructor.

src/TestSuite/IntegrationTestTrait.php
ADmad: It's due to difference in how query string params are encoded in values generated by `http_build_query()` and value returned by `Uri::getQuery()`. I have explained it my comment https://github.com/cakephp/cakephp/pull/13955#issuecomment-559854498.

ADmad:
I am of the opinion that this should be simply closed. If one is setting the request uri through environment params then full url info should be set through that and not reply on other config params. Trying to merge info from various options creates a mess in the code.

dereuromark:
We dont fix/shim this inside 3.x?
The current way is not adequate - and people are forced to upgrade due to the deprecations thrown, only to have all failing with hard to debug code.
If you close this, you need to make the deprecation soft here, so we can keep here() until 4.x for sure.

dereuromark:
Maybe you need to also throw deprecation warnings at least for this kind of partial "setting" scenario then, to clarify the issue instead of silently doing noops.

ADmad:
Sorry, having picked this up again after few days I forget that the main issue here is difference in values returned by `request->here()` to `request->getRequestTarget()`.  That's is something that should be fixed. I'll see if I can come up with a simpler fix for that.

ADmad:
So I checked the output of `request->here()` and `request->getRequestTarget()` in actual 3.x app and URL returned by both have query string, so I don't think anything needs to be done regarding that.

As for your examples if your includes the query string in the URL you set for either `REQUEST_URI` or `'url'` then both methods should return the url with query string.

dereuromark:
It is true, normal requests don't necessarily face this.
But there are other use cases where you manually use this to introspect a request, or to test.
And in those cases it would silently behave differently than PSR documented this to work.
Or does the docs here forbid this usage despite the params being available on the constructor?

Then let's just add the deprecation notices on partial setup here, so you can't silently run into this and know how to resolve this. I think that could be a good compromise here.

ADmad:
I checked Diactoros' `ServerRequest` class and for that too if you pass the query params using it's `$queryParams` but don't include the query string with same values in the value for `$uri` argument then value returned by `getRequestTarget()` won't include the query string.

So if one is constructing the request object by hand then it's the devs responsibility to include the query string in uri to match the query params.

Issue #13944, opened 2019-11-26
Collision uuid in PHP 5.6


nojimage:
This is a (multiple allowed):

* [ ] bug
* [x] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.x
* Platform and Target: PHP 5.6

### What you did

In PHP 5.6 `Cake\Utility\Text::uuid()` uses `mt_rand`.
If the seed is inadvertently initialized with `mt_srand`, the generated UUID collides.

Sample code: https://gist.github.com/nojimage/df00550e9cbc412e479ec9d54939cb67

I think Text::uuid should always use `random_int` instead of `mt_rand` and install [paragonie/random_compat](https://github.com/paragonie/random_compat).

nojimage:
I think to solve this problem, just add `"paragonie/random_compat": "^2.0|9.99.99"` into composer.json.

rchavik:
Is there anyway to NOT add `random_compat` as dependency when already using php > 7.1?

markstory:
@rchavik I don't think there is.

othercorey:
> Is there anyway to NOT add `random_compat` as dependency when already using php > 7.1?

Is there a downside to having the dependency?

rchavik:
I'd rather it throws exceptions when used in 5.6.  It's already past EOL and next year reaches end of support.

othercorey:
You want `Text::uuid()` to throw an exception in 5.6?

rchavik:
Or something like this:

```diff
diff --git a/src/Utility/Text.php b/src/Utility/Text.php
index ddcef7eb3a..e1cc10aac2 100644
--- a/src/Utility/Text.php
+++ b/src/Utility/Text.php
@@ -61,6 +61,9 @@ class Text
     public static function uuid()
     {
         $random = function_exists('random_int') ? 'random_int' : 'mt_rand';
+        if ($random === 'mt_rand') {
+            trigger_error('Please install random_compat');
+        }
 
         return sprintf(
             '%04x%04x-%04x-%04x-%04x-%04x%04x%04x',
```

Affected installations can then install random_compat or upgrade to supported php versions.

nojimage:
"paragonie/random_compat" v9.99.99 installs with PHP 7, but it does nothing provide.

https://github.com/paragonie/random_compat/tree/v9.99.99

nojimage:
@rchavik How about using `deprecationWarning` instead of trigger_error?

[...]

PR #13958, opened 2019-11-29
Deprecation of mt_rand in uuid generation

nojimage:
fixes for #13944

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -35,7 +35,8 @@
         "aura/intl": "^3.0.0",
         "psr/log": "^1.0.0",
         "psr/simple-cache": "^1.0.0",
-        "zendframework/zend-diactoros": "^1.4.0"
+        "zendframework/zend-diactoros": "^1.4.0",
+        "paragonie/random_compat": "^2.0|9.99.99"
     },
     "suggest": {
         "ext-openssl": "To use Security::encrypt() or have secure CSRF token generation.",
--- a/src/Utility/Text.php
+++ b/src/Utility/Text.php
@@ -60,25 +60,23 @@ class Text
      */
     public static function uuid()
     {
-        $random = function_exists('random_int') ? 'random_int' : 'mt_rand';
-
         return sprintf(
             '%04x%04x-%04x-%04x-%04x-%04x%04x%04x',
             // 32 bits for "time_low"
-            $random(0, 65535),
-            $random(0, 65535),
+            random_int(0, 65535),
+            random_int(0, 65535),
             // 16 bits for "time_mid"
-            $random(0, 65535),
+            random_int(0, 65535),
             // 12 bits before the 0100 of (version) 4 for "time_hi_and_version"
-            $random(0, 4095) | 0x4000,
+            random_int(0, 4095) | 0x4000,
             // 16 bits, 8 bits for "clk_seq_hi_res",
             // 8 bits for "clk_seq_low",
             // two most significant bits holds zero and one for variant DCE1.1
-            $random(0, 0x3fff) | 0x8000,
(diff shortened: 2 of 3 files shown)


rchavik:
This adds a dependency and also throw a depreciationWarning.

It should be one OR the other.

src/Utility/Text.php
```diff
@@ -61,6 +61,9 @@ class Text
     public static function uuid()
     {
         $random = function_exists('random_int') ? 'random_int' : 'mt_rand';
```
rchavik: if random_compat is installed, we can immediately do: 

```php
$random = 'random_int';
```

am i missing something?

nojimage: When upgrading an existing application, there seems to has a short time random_int to not exist, between updating the package with composer.
If we can ignore this, we can unify to random_int.

dereuromark:
I would recommend letting people opt in rather than requiring this.

composer.json
```diff
@@ -35,7 +35,8 @@
         "aura/intl": "^3.0.0",
         "psr/log": "^1.0.0",
         "psr/simple-cache": "^1.0.0",
-        "zendframework/zend-diactoros": "^1.4.0"
+        "zendframework/zend-diactoros": "^1.4.0",
+        "paragonie/random_compat": "^2.0|9.99.99"
```
othercorey: What does requiring 9.99.99 do?

garas: PHP5 will install v2.0 with `lib/random.php` autoloaded.

PHP7 will install v9.99.99 with no file autoloaded.

othercorey: I see now. I was looking in master branch.  Very nice, so this solves our problem.

src/Utility/Text.php
othercorey: > When upgrading an existing application, there seems to has a short time random_int to not exist, between updating the package with composer.

The composer.json change makes that impossible though, right?

othercorey:
I think 3.9 is probably better for this.  We can add it to migration notes.

othercorey:
> I would recommend letting people opt in rather than requiring this.

symfony seems to use this in their polyfill for php7

src/Utility/Text.php
othercorey: I removed the deprecation notice and defaulted to `random_int()`. If PHP suggests this implementation in their official doc, it's good enough for me.

https://www.php.net/manual/en/function.random-int.php

othercorey:
I'm not sure why it is failing to run.  I'll have to clone locally.

nojimage:
> I'm not sure why it is failing to run. I'll have to clone locally.

It seems to polyfilled `Error` class in ramdom_compat. but `ParseError` is not.
 https://github.com/paragonie/random_compat/blob/9e071856f328b3d249df75b3256394d398809ed7/lib/error_polyfill.php#L29

othercorey:
> > I'm not sure why it is failing to run. I'll have to clone locally.
> 
> It seems to polyfilled `Error` class in ramdom_compat. but `ParseError` is not.
> https://github.com/paragonie/random_compat/blob/9e071856f328b3d249df75b3256394d398809ed7/lib/error_polyfill.php#L29

I was just looking at why Error was defined, thanks!  I fixed it with a proper version check like everywhere else.

othercorey:
I made the PR list all messy by messing with 3.next trying to find the build break.  Doesn't matter for the merge though.

ADmad (approved):
Changes look good but there's lot of commit noise. Please squash them into a single commit.

othercorey:
> Changes look good but there's lot of commit noise. Please squash them into a single commit.

Done.

PR #13959, opened 2019-11-29
Use a single argument for ErrorHandlerMiddleware::__construct().

ADmad:
This avoids the fugliness of passing null as 1st argument when you want to use config array.

### DIFF
--- a/src/Error/Middleware/ErrorHandlerMiddleware.php
+++ b/src/Error/Middleware/ErrorHandlerMiddleware.php
@@ -80,6 +80,15 @@ class ErrorHandlerMiddleware implements MiddlewareInterface
      */
     public function __construct($errorHandler = [])
     {
+        if (func_num_args() > 1) {
+            deprecationWarning(
+                'The signature of ErrorHandlerMiddleware::__construct() has changed. '
+                . 'Pass the config array as 1st argument instead.'
+            );
+
+            $errorHandler = func_get_arg(1);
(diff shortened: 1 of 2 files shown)


ADmad:
This change will required updating the app skeleton:

https://github.com/cakephp/app/blob/6f5bf5b1ab2807d1e1fe5cb44be87681ab9d451c/src/Application.php#L69

src/Error/Middleware/ErrorHandlerMiddleware.php
```diff
@@ -72,18 +75,25 @@ class ErrorHandlerMiddleware implements MiddlewareInterface
...
-     *   These options will be ignored if an ErrorHandler instance is passed as
-     *   first argument.
+     * @param \Cake\Error\ErrorHandler|array $errorHandler The error handler instance
+     *  or config array.
      */
-    public function __construct($errorHandler = null, array $config = [])
+    public function __construct($errorHandler = [])
     {
-        $this->setConfig($config);
+        if (is_array($errorHandler)) {
```
markstory: If you wanted to allow a more graceful upgrade process we could use `func_get_args()` to allow the second parameter to continue being used and emit a deprecation warning? Perhaps its not a big deal though as we've not finalized everything yet.

ADmad: Done

PR #13965, opened 2019-12-01
Fix GET requests with bodies being sent as POST

### LINKED ISSUE #13406: GET request with body is being sent as POST request (HTTP Curl Client)
This is a (multiple allowed):

* [x] bug
* [x] enhancement

* CakePHP Version: from 3.7

### What you did
```php
$client = new Client();
$request = new Request(
    'http://localhost:8888/requestbin.php', 
    'GET',
    [],
    'param1=foo'
);

$response = $client->send($request);
```

### What happened
POST Request

GET Request

I know it's a very unusual use case (but not prohibit by the HTTP spec). But i don't think a Request should be converted to something that's unexpected. 

### Issue
cURL will automatically change method to PUT or POST if CURLOPT_POSTFIELDS is set. 
https://github.com/cakephp/cakephp/blob/master/src/Http/Client/Adapter/Curl.php#L98

markstory:
While infrequently used, GET requests with bodies are allowed and used by applications like elasticsearch. Fix GET requests being sent as post.

Fixes #13406

### DIFF
--- a/src/Http/Client/Adapter/Curl.php
+++ b/src/Http/Client/Adapter/Curl.php
@@ -96,6 +96,10 @@ class Curl implements AdapterInterface
         if ($body) {
             $body->rewind();
             $out[CURLOPT_POSTFIELDS] = $body->getContents();
+            // GET requests with bodies require custom request to be used.
+            if (isset($out[CURLOPT_HTTPGET])) {
+                $out[CURLOPT_CUSTOMREQUEST] = 'get';
+            }
         }
 
         if (empty($options['ssl_cafile'])) {
--- a/tests/TestCase/Http/Client/Adapter/CurlTest.php
+++ b/tests/TestCase/Http/Client/Adapter/CurlTest.php
@@ -119,6 +119,42 @@ class CurlTest extends TestCase
      *
      * @return void
      */
+    public function testBuildOptionsGetWithBody()
+    {
+        $options = [
+            'timeout' => 5,
+        ];
+        $request = new Request(
+            'http://localhost/things',
+            'GET',
+            ['Cookie' => 'testing=value'],
+            '{"some":"body"}'
+        );
+        $result = $this->curl->buildOptions($request, $options);
+        $expected = [
+            CURLOPT_URL => 'http://localhost/things',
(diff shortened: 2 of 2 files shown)


src/Http/Client/Adapter/Curl.php
```diff
@@ -96,6 +96,10 @@ public function buildOptions(Request $request, array $options)
         if ($body) {
             $body->rewind();
             $out[CURLOPT_POSTFIELDS] = $body->getContents();
+            // GET requests with bodies require custom request to be used.
+            if (isset($out[CURLOPT_HTTPGET])) {
+                $out[CURLOPT_CUSTOMREQUEST] = 'get';
```
othercorey: I was wondering if just moving the CURLOPT_HTTPGET option to the end of the array would work.

However, this should work nicely because it doesn't matter if CURLOPT_HTTPGET is reset by curl.

othercorey (approved):

src/Http/Client/Adapter/Curl.php
fwille: I know this discussion is almost 6 years old. 
Has anybody tested this with CakePHP 5? I recently wanted to send a GET Request with body (Elasticsearch) and always got a 405 Response when using the CakePHP Client. As soon as I omit the body it works.
Something appears to be wrong with the library when using GET requests with body.

It is such a rare use case that I don't know if it just broke recently or it has been like this since this change was merged.

In my case I can just use a POST instead and it works fine - still, GET requests with body are unusual but allowed in theory and should probably be working with this library.

fwille: Is there a reason why this is a lower case 'get' instead of uppercase 'GET'? My request appears to work fine once I change it to uppercase and I no longer receive the status code for Method Not Allowed (405) from the ES endpoint.

markstory: Generally the casing of HTTP methods doesn't matter. If having it be uppercase solves an issue for you, feel free to open a pull request or new issue. We can't do much with comments on a 6 year old pull request :)

Issue #13492, opened 2019-08-08
The return code of the (2.x) testrunner is > 0 although there are no failing tests


tersmitten:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 2.10.19

### What you did
Run our unit test (using `exec ${phpBinary} ${defineIni} -q "${PWD}/Console/cake.php" -working "${PWD}" test app All --stderr "$@";` like we always did)

### What happened
The return code is not `0` anymore, although there are no failing tests.

### What you expected to happen
The return code should be `0`, because the only thing I changed is upgrade CakePHP from `2.10.18` to `2.10.19`

The issue is probably related to https://github.com/cakephp/cakephp/commit/396e342601d74175df981d8cda0a33dd64c348c6.

Something that might be specific for our testsuite is that we have `2 warnings` and `22 skipped` tests.

dereuromark:
Are you able to provide a PR here with your suggested changes to fix this?

tersmitten:
I do not have a suggested change (yet). I'm not sure what the change was anyway.

tersmitten:
I'm using `phpunit` version `5.7.27`

tersmitten:
Do you have any idea @tenkoma?

tenkoma:
I'm sorry to be late. I'll try to find out the cause.

othercorey:
This needs to check `wasSuccessfulIgnoringWarnings()` instead of `wasSuccessful()` here https://github.com/cakephp/cakephp/blob/5eed56b6e45cc2e01d0ce0de6808e67e624e6502/lib/Cake/TestSuite/CakeTestSuiteCommand.php#L104

othercorey:
You're going to have to manually check error and failure count on older phpunits unfortunately.  There is no wasSuccessfulIgnoreWarnings() in those versions.

PR #13966, opened 2019-12-01
2.x - Ignore warnings when generating exit code for phpunit doRun() result

othercorey:
Fixes https://github.com/cakephp/cakephp/issues/13492.

This syncs up with newer phpunit versions.  The 5.2.0 version this was originally based on incorrectly failed on warnings by default.


othercorey:
Seems that's not available on the supposed phpunit versions.

PR #13969, opened 2019-12-02
2.x - PHP 7.4

dereuromark:
Adding php74 check

Also: running all possible php versions in a matrix is a bit silly
each php version also already has a specific DB type job as well

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,6 +13,9 @@ env:
 
 services:
   - memcached
+  - redis
+  - postgresql
+  - mysql
 
 matrix:
   fast_finish: true
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,13 +5,10 @@ dist: precise
 
 php:
   - 5.3
-  - 5.4
-  - 5.5
   - 5.6
   - 7.0
-  - 7.1
-  - 7.2
   - 7.3
+  - 7.4snapshot
 
 env:
   matrix:
@@ -31,10 +28,10 @@ matrix:
     - php: 5.4
       env: DB=pgsql
 
-    - php: 5.4
+    - php: 5.5
       env: DB=sqlite
 
-    - php: 7.0
+    - php: 7.3
       env: PHPCS=1
 
     - php: 7.1
@@ -50,7 +47,6 @@ matrix:
       env: DB=mysql
     - php: 7.3
       env: DB=mysql
-
 
 before_script:
   - if [[ ${TRAVIS_PHP_VERSION:0:3} =~ ^7\.[23]$ ]] ; then pear config-set preferred_state snapshot && yes "" | pecl install mcrypt ; fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,6 @@
 language: php
 
+os: linux
 dist: xenial
 
 php:
@@ -8,7 +9,7 @@ php:
   - 7.4
 
 env:
-  matrix:
+  jobs:
     - DB=mysql
   global:
     - PHPUNIT=5.7.19
@@ -19,7 +20,7 @@ services:
   - postgresql
   - mysql
 
-matrix:
+jobs:
   fast_finish: true
   include:
     - php: 5.3
@@ -41,11 +42,8 @@ matrix:
     - php: 7.3
       env: DB=sqlite
 
-    - php: 7.2
+    - php: 7.0
       env: PHPCS=1
-
-  allow_failures:
-    - php: 7.4
 
 before_script:
   - if [[ ${TRAVIS_PHP_VERSION:0:3} =~ ^7\.[234]$ ]] ; then pear config-set preferred_state snapshot && yes "" | pecl install mcrypt ; fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,7 @@
 language: php
 
-# We require precise here, as 5.3 doesn't exist on trusty.
-dist: precise
-
 php:
-  - 5.3
   - 5.6
-  - 7.0
   - 7.3
   - 7.4snapshot
 
@@ -22,13 +17,10 @@ services:
 matrix:
   fast_finish: true
   include:
-    - php: 5.3
-      env: DB=mysql PHPUNIT=4.8.35
-
-    - php: 5.4
+    - php: 5.6
       env: DB=pgsql
 
-    - php: 5.5
+    - php: 7.0
       env: DB=sqlite
 
     - php: 7.3
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,15 +1,17 @@
 language: php
 
+dist: xenial
+
 php:
   - 5.6
   - 7.3
-  - 7.4snapshot
+  - 7.4
 
 env:
   matrix:
     - DB=mysql
   global:
-    - PHPUNIT=3.7.33
+    - PHPUNIT=5.7.19
 
 services:
   - memcached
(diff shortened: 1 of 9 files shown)


dereuromark:
Looks like dist: precise is not shipping 7.4+ anymore...
I think here we now should drop PHP5.5 and below. Everyone running an old cake2 app by now is long working with php5.6+ and that can easily be expected by now.

garas:
PHP 5.3 could stay, just add `dist: precise` in it's `include:`. CakePHP 2.x supports 5.3+, so continue support by tests.

5.6, 7.3, 7.4 versions can run on latest distribution.

dereuromark:
I disagree.
But feel free to make an alternative PR here.

.travis.yml
```diff
@@ -1,17 +1,9 @@
 language: php
 
-# We require precise here, as 5.3 doesn't exist on trusty.
-dist: precise
```
othercorey: This will jump all the way up to xenial, not trust.  Should we be explicit?

The default mysql version will be 5.7.

dereuromark: You can add more commits to do that, or open a 2nd PR with your suggestions.
We can merge whatever becomes green first :)

othercorey: I don't see why I would add commits or a PR when I don't know what I'm trying to change.

dereuromark:
> Function create_function() has been deprecated

Do we want to silence deprecations here? Probably not worth fixing
At this point we should outline/document that using E_ALL with silenced E_USER_DEPRECATED and E_DPRECATED should be used instead for such an old code.

ADmad:
For 2.x ensuring the code runs on the whatever the minimum supported PHP version is more important than testing against PHP 7.4. So I am not in favour of removing older PHP versions from the test matrix.

garas:
There is only one `create_function()` in `RssHelperTest` and easily converted to closure.

Any other hard to fix deprecation notice could be silenced.

dereuromark:
@ADmad 
In these cases servers soon also update to more modern versions (and they must never run e.g. 5.3 these days anyways).

At this point in time, if you are running a 2.x app and code base, you need to make sure that you at least keep it inside the PHP releases of that major version, so by now 5.6.
We need to maintain critical security fixes and minimal compatibility, as such it is reasonable to drop support for everything <= 5.5 at this point and state that 5.6...7.4 is fine to run it with for the next months.
If you have older PHP servers, you wont upgrade anything, neither the code base nor anything else for that matter.

So I disagree with your assessment in that way, that we should just set the composer and documented limits correctly now as per reasonable MIN version support possible at this time.

garas:
I've added back PHP 5.3 on `precise`, all others will run on `xenial`. Non-`snapshot` 7.4 is also now available on Travis.

`cakephp-codesniffer:1.*` is not compatible with PHP 7.3, run it with 7.2.

dereuromark:
I still think that bumping PHP to a reasonable minimum inside 5.x series is the right thing to do.
In the meantime we can better educate here maybe: https://github.com/cakephp/docs/pull/6294

ADmad:
@dereuromark Let's not get into what users "should" or "must" do. Ideally everyone should have upgrade to 3.x by now but that's not how things work in the real world. 

Currently the min. supported version for 2.x is PHP 5.3 and the test suite should be run against it. Since 2.x is essentially EOL bumping up the PHP version doesn't make sense. Also it would require bumping minor version for 2.x which isn't something we should get into.

Since @garas has managed to get the builds running on PHP 5.3 as well as PHP 7.4 there's no need to drop PHP 5.3 support.

dereuromark:
If thats the case OK. But it still doesnt invalidate my points.

dereuromark:
Can we silence the deprecations for travis? As there is no point (and often no possibility) in fixing them for PHP7+ etc here.

markstory:
Changing the minimum required php version for 2.x could cut people off of future security fixes should they be necessary. We could choose to not support php7.4 in 2.x instead.  The support window for 7.3 will be longer than we plan on supporting 2.x.

ADmad:
> We could choose to not support php7.4 in 2.x instead. The support window for 7.3 will be longer than we plan on supporting 2.x.

:+1:

dereuromark:
> We could choose to not support php7.4 in 2.x instead. The support window for 7.3 will be longer than we plan on supporting 2.x.

:-1: : 

That is completely the wrong approach
From experience I can say that people are using 2.x still for lack of being able to upgrade that monster to 3.x functionality, as business functionality would break hard with months of refactor needs.
Not because of infrastructure (which even requires you to usually to to PHP7 and beyond since old servers are also force upgraded to more modern versions - due to security concerns).

As such you rather need to drop off old 5year+ old PHP instead of not supporting new PHP minors.

Completely the wrong thoughts here.

dereuromark:
But why are we still discussing this then, if the current matrix has those enabled again and no such change has been made to composer require?
Afaik, if 7.4 now has deprecations silenced it proofs that it is working fine and no further cutting off is necessary at this point now.

The prio must still be to further enable newer PHP versions. That is all I am saying.
If old 5.3 etc still work, so be it. But this is not someone should ever count on, especially with responsible server infrastructure running in production.

saeideng:
>Changing the minimum required php version for 2.x could cut people off of future security fixes should they be necessary. We could choose to not support php7.4 in 2.x instead.

:+1:

LuizMarin:
Who am I to give an opinion to such experienced people as you ..!

But for over 30 years I have always taught not to live the past and stay up to date. What is badly done needs to be redone. Patches on old clothes always end badly. (* I don't know if google translates the expressions well .... probably not. But anyway ...)

I installed cake on my machine always updated daily with php 7.3, but I have 7.4 for testing in the same project.

What did i see? Php improvements. If I had a 'monster project' done in some old version, would I change it? Certainly! The customer deserves to have competitive systems with the current market reality. It means better business for me (and ease of maintenance as well).

I chose cakephp a few months ago because it has a constant update commitment and the developer team is accessible.

In diplomacy, we try to keep old clients and gain new ones. Cake has evolved from v3 to v4. The php too. Operating systems as well. The rogues too, and that drove security improvements.

Finally, the name and logo "cake" suggests a tasty cake with easy to use and make recipe. Let the cake get old in the bakery and see how many will want a piece ....
Old 'cake' is bad, bad for your health and going to waste.

Please keep the 'Cake' new, beautiful and attractive and many will be attracted and will want it, without complaining, regardless of the 'cost' of refactoring or extra work: because it's hot!

Best regards,

dereuromark:
@LuizMarin 
What I don't understand in the above discussion is the "argument of security bugfixes" while allowing/using unsecure operation systems and PHP versions by design, even requiring them with the "php53 first, php7.x+ last" approach.
I can only re-emphasize the importance of staying up to date, even with a 2.x code base.
That means at least PHP5.6 these days, as I am sure most responsible devs/companies use anyways, if for platform requirements or other reasons. 
Therefore even with bumps of PHP version here to reasonable minimums there is no blocking at all.
Whoever uses such outdated versions as 5.3 still has other issues than pulling in bugfixes in Cake core IMO.
And dropping it from travis also doesnt mean dropping the versions as composer requirement (even if this would also make sense at this point in time, 2019).

But that said, we can now finalize the current PR as inclusive to all versions, even if I am really sad about the general opinion here in not understanding the irony of keeping 5.3 possible while argumenting security or operability.
5.3 was EOL in 2014, that is 5+ years ago.

dereuromark:
LTS (long term support) does not mean you have to support these old PHP versions, it means you make sure the major version is compatible with the reasonable PHP versions out there on platforms/servers.
But that is now a discussion for a future PR probably.

markstory:
Doesn't dropping 5.3 from our test matrix make it hard for us to catch future breaks when 5.4+ syntax or features are used?

I agree that we don't need to test 5.4 and 5.5, or the intermediate php7 versions. Without a 5.3 build we will be unable to ensure we stay compatible with our composer constraints.

dereuromark:
Yeah, thats why it is cleaner to move away from it alltogehter
For now we can still support all versions, so all good.

dereuromark:
@garas Nice work on the shimming in tests.

garas:
I've fixed all failures on PHP 7.4.

Most of failures were `Trying to access array offset on value of type bool/int/null`. Few test warnings revealed wrong tests.

lib/Cake/Test/Case/Model/Datasource/Database/MysqlTest.php
```diff
@@ -356,41 +356,6 @@ public function testIndexDetection() {
...
 	}
 
-/**
- * testBuildColumn method
- *
- * @return void
- */
-	public function testBuildColumn() {
-		$restore = $this->Dbo->columns;
-		$this->Dbo->columns = array('varchar(255)' => 1);
```
garas: `$this->Dbo->columns` should be array of `'string' => array('name' => 'varchar', 'limit' => '255')` like arrays.

Behaviour with array like `array('varchar(255)' => 1)` is undefined, generates invalid ``'`testName`  DEFAULT NULL COMMENT \'test\''`` (note double space).

Test caused many failures on further tests, as it didn't restored original columns after failure.

othercorey:
If you are planning to move the minimum php version up for 2.x, I would suggest making this a new minor release not a patch release.

Just a thought as I don't use 2.x.

dereuromark:
@othercorey 
At this point in time though even a patch should be acceptable for the reasons outlined above.
But a minor would then indeed be cleaner.

I don't think it is planned anymore (and keep full 5.3 compat), though. Seems like travis can still be used in a way to support the full range somehow.

ADmad:
Since @garas has got the testsuite running and passing on PHP 5.3 as well as 7.4 there's no need to make any changes to the PHP version constraints.

PR #13970, opened 2019-12-03
4.x - Updated supported phpunit versions to ^8.4

othercorey:
Just set to ^8.4 instead of ^8.3.3|^8.4


othercorey:
This is only to ensure the versions don't break. I am guessing point releases of phpunit can contain breaking changes.

PR #13972, opened 2019-12-03
Nested Joins acting incorrectly

ruanpepe:
If table A is linked to table B by a left join, and the table B is linked to table C by a inner join, when you try to access the table C from table A, the query acts like table A was linked to table B by a inner join.
I modified  the "_buildJoinPart" method to group nested joins by parentheses, and that is working fine now.

### DIFF
--- a/src/Database/QueryCompiler.php
+++ b/src/Database/QueryCompiler.php
@@ -13,7 +13,9 @@
  * @license       https://opensource.org/licenses/mit-license.php MIT License
  */
 namespace Cake\Database;
+
 use Cake\Database\Expression\QueryExpression;
+
 /**
  * Responsible for compiling a Query object into its SQL representation
  *
@@ -38,6 +40,7 @@ class QueryCompiler
         'offset' => ' OFFSET %s',
         'epilog' => ' %s',
     ];
+
     /**
      * The list of query clauses to traverse for generating a SELECT statement
      *
@@ -47,24 +50,28 @@ class QueryCompiler
         'select', 'from', 'join', 'where', 'group', 'having', 'order', 'limit',
         'offset', 'union', 'epilog',
     ];
+
     /**
      * The list of query clauses to traverse for generating an UPDATE statement
      *
      * @var array
      */
     protected $_updateParts = ['update', 'set', 'where', 'epilog'];
+
     /**
      * The list of query clauses to traverse for generating a DELETE statement
      *
      * @var array
      */
     protected $_deleteParts = ['delete', 'modifier', 'from', 'where', 'epilog'];
+
     /**
      * The list of query clauses to traverse for generating an INSERT statement
      *
      * @var array
      */
     protected $_insertParts = ['insert', 'values', 'epilog'];
+
     /**
      * Indicate whether or not this query dialect supports ordered unions.
      *
@@ -73,6 +80,7 @@ class QueryCompiler
      * @var bool
      */
     protected $_orderedUnion = true;
+
     /**
      * Returns the SQL representation of the provided query after generating
      * the placeholders for the bound values using the provided generator
@@ -89,6 +97,7 @@ class QueryCompiler
             $this->_sqlCompiler($sql, $query, $generator),
             $this->{'_' . $type . 'Parts'}
         );
+
         // Propagate bound parameters from sub-queries if the
         // placeholders can be found in the SQL statement.
         if ($query->getValueBinder() !== $generator) {
@@ -99,8 +108,10 @@ class QueryCompiler
                 }
             }
         }
+
         return $sql;
     }
+
     /**
      * Returns a callable object that can be used to compile a SQL string representation
      * of this query.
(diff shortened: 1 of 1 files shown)


markstory:
Any chance you could include a way to reproduce the problem you're having? Your fix has caused many tests to fail.

ruanpepe:
> Any chance you could include a way to reproduce the problem you're having? Your fix has caused many tests to fail.

My model:

JuridicalPeople model
`
class JuridicalPeopleTable extends Table
{
    public function initialize(array $config)
    {
        parent::initialize($config);

        $this->setTable('juridical_people');
        $this->setDisplayField('id');
        $this->setPrimaryKey('id');

        $this->hasOne('Teams', [
            'foreignKey' => 'juridical_person_id',
            'dependent' => true,
            'cascadeCallbacks' => true
        ]);
    }

    public function validationDefault(Validator $validator)
    {
        $validator
            ->integer('id')
            ->allowEmptyString('id', null, 'create');

        $validator
            ->scalar('logo')
            ->maxLength('logo', 255)
            ->allowEmptyString('logo');

        return $validator;
    }
}

class JuridicalPerson extends Entity
{
    protected $_accessible = [
        'logo' => true,
        'team' => true
    ];
}
`

My Teams model:
`
class TeamsTable extends Table
{
    public function initialize(array $config)
    {
        parent::initialize($config);

        $this->setTable('teams');
        $this->setDisplayField('id');
        $this->setPrimaryKey('id');

        $this->belongsTo('JuridicalPeople', [
            'foreignKey' => 'juridical_person_id',
            'joinType' => 'INNER'
        ]);
        
        $this->hasMany('Athletes', [
            'foreignKey' => 'team_id'
        ]);
    }

    public function validationDefault(Validator $validator)
    {
        $validator
            ->integer('id')
            ->allowEmptyString('id', null, 'create');

        return $validator;
    }

    public function buildRules(RulesChecker $rules)
    {
        $rules->add($rules->existsIn(['juridical_person_id'], 'JuridicalPeople'));

        return $rules;
    }
}

class Team extends Entity
{
    protected $_accessible = [
        'juridical_person_id' => true,
        'juridical_person' => true,
        'athletes' => true,
    ];
}
`

My Athletes model:
`
class AthletesTable extends Table
{
    public function initialize(array $config)
    {
        parent::initialize($config);

        $this->setTable('athletes');
        $this->setDisplayField('id');
        $this->setPrimaryKey('id');

        $this->belongsTo('Teams')
            ->setForeignKey('team_id');
    }

    public function validationDefault(Validator $validator)
    {
        $validator
            ->integer('id')
            ->allowEmptyString('id', null, 'create');

        return $validator;
    }

    public function buildRules(RulesChecker $rules)
    {
        $rules->add($rules->existsIn(['team_id'], 'Teams'));

        return $rules;
    }
}
`

My Athletes Controller:
`class AthletesController extends AppController
{
    public function index()
    {
        $this->paginate = [
            'contain' => [
                'NaturalPeople',
                'Teams.JuridicalPeople']
        ];
        
        $athletes = $this->paginate($this->Athletes->find());

        $this->set(compact('athletes'));
    }
}`

At my athletes/index, if the athlete doesn't match with a team, he should be showed becouse the Athletes table is joined to Teams table by a left join. But this is not happening becouse the teams table is joined to JuridicalPeople table by inner join, and the current way to build query insert all joins at same level. So, at query, the Athletes table is joined to the JuridicalPeople table by a inner join.

othercorey:
@ruanpepe Your editor is changing every coding standard.  Please adjust it to stop.

src/Database/QueryCompiler.php
```diff
@@ -234,12 +256,17 @@ protected function _buildJoinPart($parts, $query, $generator)
                 $join['table'] = '(' . $join['table'] . ')';
             }
 
-            $joins .= sprintf(' %s JOIN %s %s', $join['type'], $join['table'], $join['alias']);
+            $joins .= sprintf(' %s JOIN (%s %s', $join['type'], $join['table'], $join['alias']);
```
markstory: Could we only use parenthesis on join expressions that contain child associations? I think this change is why so many tests are failing.

ruanpepe: > Could we only use parenthesis on join expressions that contain child associations? I think this change is why so many tests are failing.

Yes. Te query join structure was changed from
```
LEFT JOIN teams Teams 
ON Teams.id = (Athletes.team_id) 

INNER JOIN juridical_people JuridicalPeople 
ON JuridicalPeople.id = (Teams.juridical_person_id)
```
to
```
LEFT JOIN (teams Teams 
	INNER JOIN (juridical_people JuridicalPeople) 
	ON JuridicalPeople.id = (Teams.juridical_person_id)
) 
ON Teams.id = (Athletes.team_id) 
```
and it's working fine to me.
Another way to do this, would be inserting a subquery, 
```
LEFT JOIN 
(
	SELECT 
		...
		...
	FROM teams Teams
	INNER JOIN juridical_people JuridicalPeople
	ON JuridicalPeople.id = Teams.juridical_person_id
) Teams  
ON Teams.id = (Athletes.team_id) 
```
but i think it would cause many performance troubles.

othercorey:
We are going to approach this a different way in 4.3.

PR #13973, opened 2019-12-04
4.x stan

ADmad:
Bumping phpstan to 0.12 threw a lot of new errors so I have just updated it's baseline file for now.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -110,7 +110,7 @@
             "@phpstan",
             "@psalm"
         ],
-        "stan-setup": "cp composer.json composer.backup && composer require --dev phpstan/phpstan-shim:^0.11.18 psalm/phar:^3.5 && mv composer.backup composer.json"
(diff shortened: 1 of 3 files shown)


composer.json
```diff
@@ -110,7 +110,7 @@
             "@phpstan",
             "@psalm"
         ],
-        "stan-setup": "cp composer.json composer.backup && composer require --dev phpstan/phpstan-shim:^0.11.18 psalm/phar:^3.5 && mv composer.backup composer.json"
+        "stan-setup": "cp composer.json composer.backup && composer require --dev phpstan/phpstan:^0.12.0 psalm/phar:^3.7 && mv composer.backup composer.json"
```
othercorey: Can we just require-dev it now?

ADmad: I don't expect the average contributor to run pslam/phpstan so I would rather not add extra dependencies.

People like me who use them for multiple libs/repos have it installed globally and having them installed per project is just wasteful.

othercorey: bizarre but ok.

Issue #13581, opened 2019-08-30
Session handlers silently ignore errors that occur during session write


sukihub:
This is a (multiple allowed):

* [x] bug
* [x] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.6 (but master still seems to have this).
* Platform and Target: Apache & mod_php, but all are probably affected (I replicated it on Nginx & php cgi).

### What you did
We use Cake's `DatabaseSession`. A bug in the application code caused an exception during session write. DatabaseSession uses ORM's tables and entities, so this can easily happen.

### What happened
Obviously, the session did not get written. But no error was emitted which made the problem incredibly difficult to find. Also, all the requests were finishing correctly, with 2xx responses, etc. The only visible symptom was that session sometimes did not get written.

I think this might be because of `session_write_close()` being called/registered as a shutdown function (`session_register_shutdown()` does that). There's not much that PHP can do if shutdown handler throws an exception. Yet it would be very helpful if the error was somehow reported.

### What you expected to happen
Since both `DatabaseSession` and `CacheSession` perform some non-trivial tasks during session write, I would expect my app to crash if session write fails.

Even though PHP itself cannot do much about errors in shutdown functions, CakePHP (as a provider of some session handler implementations) can.

One solution I propose is wrapping session handler's write method in a try catch:
```php
public function write($id, $data)
{
    try {
        return parent::write($id, $data); // or original implementation of write
    } catch (Throwable $exception) {
        Log::error($exception);
        return false;
    }
}
```
This solution does not make the app crash if session write fails, but at least it reports the error correctly into a log.

Actually crashing an application might not be possible as long as `session_write_close()` is being called as a shutdown function. Which seems to be the preferred way of doing it in newer PHPs. But I also tested calling `session_write_close()` manually, earlier in the request cycle and error was correctly reported. So that's option no. 2.

markstory:
Throwing an exception on write failure is something we can add, but for the reasons you've outlined it might not work in all scenarios. Logging directly from the session adds additional cross package dependencies and might not be what all users want, whereas an exception can be caught and converted into logs, or sent to an upstream for collection.

sukihub:
The exception already gets thrown on write failure (probably, most of the time), doesn't it?

I see what you're saying with cross package dependencies. But exceptions in shutdown functions are silently ignored, so this will solve nothing.

markstory:
> But exceptions in shutdown functions are silently ignored, so this will solve nothing.

[...]

PR #13976, opened 2019-12-04
4.x - Explicit write & close of the session

sukihub:
- so that any exceptions in session write go through regular error handling process
- see #13581 for details

I would like to implement some more tests:
- for the assumption, that the exception from `SessionHandlerInterface::close` propagates up correctly
- and that a call to `session_write_close` on already closed session is a noop

I'm also not sure that calling `session_register_shutdown` is a good idea, discussion in #13581 

Otherwise this should be good enough.

### DIFF
--- a/src/TestSuite/Constraint/Session/SessionEquals.php
+++ b/src/TestSuite/Constraint/Session/SessionEquals.php
@@ -16,6 +16,7 @@ declare(strict_types=1);
 namespace Cake\TestSuite\Constraint\Session;
 
 use Cake\Http\Session;
+use Cake\Utility\Hash;
 use PHPUnit\Framework\AssertionFailedError;
 use PHPUnit\Framework\Constraint\Constraint;
 
@@ -65,7 +66,10 @@ class SessionEquals extends Constraint
      */
     public function matches($other): bool
     {
-        return $this->session->read($this->path) === $other;
+        // Server::run calls Session::close at the end of the request.
+        // Which means, that we cannot use Session object here to access the session data.
+        // Call to Session::read will start new session (and will erase the data).
+        return Hash::get($_SESSION, $this->path) === $other;
     }
 
     /**
--- a/src/Http/ResponseEmitter.php
+++ b/src/Http/ResponseEmitter.php
@@ -34,9 +34,6 @@ use Zend\HttpHandlerRunner\Emitter\EmitterInterface;
  *
  * - It logs headers sent using CakePHP's logging tools.
  * - Cookies are emitted using setcookie() to not conflict with ext/session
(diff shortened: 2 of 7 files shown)


markstory:
The test failures in travis seem relevant.

sukihub:
Well. `Session::close` did not have any implementation for running in CLI, but `Session::start` did.

(An alternative solution would be putting an alternative implementation of `session_write_close` here https://github.com/cakephp/cakephp/blob/master/tests/TestCase/Http/server_mocks.php)

markstory:
> An alternative solution would be putting an alternative implementation of session_write_close here

I'm not sure that will work unless the test harness also loads the namespaced mocks (which it currently doesn't).

sukihub:
> I'm not sure that will work unless ...

OK, I like the solution I implemented better :)

But. PHPStan makes a good point. Sometimes, `Server::run` might get PSRs `ServerRequestInterface` and that one does not have session. Which means something else "created" the session and should probably be responsible for closing it.

Would doing something like `if ($request instanceof ServerRequest) $sesssion->close()` be a good idea?

markstory:
> Would doing something like if ($request instanceof ServerRequest) $sesssion->close() be a good idea?

It would.

src/TestSuite/Constraint/Session/SessionEquals.php
```diff
@@ -65,7 +66,10 @@ public function __construct(?Session $session, string $path)
      */
     public function matches($other): bool
     {
-        return $this->session->read($this->path) === $other;
+        // Server::run calls Session::close at the end of the request.
+        // Which means, that we cannot use Session object here to access the session data.
+        // Call to Session::read will start new session (and will erase the data).
+        return Hash::get($_SESSION, $this->path) === $other;
```
ADmad: Directly using `$_SESSION` here means one will never be able to mock the session class. Though I don't have an alternative suggestion.

markstory: If you're using a mock, you can't use these methods anyways.

sukihub: I'm not sure what did you mean by "mock the session class", but I can try to offer a solution.

`Session` currently has few `isCLI` if-s, that are true while running tests. It means that it basically uses different implementation for CLI/tests. I can imagine having completely different classes for request/cli/test sessions. And the test one could allow us to read its contents here.

markstory: I was referring to the scenario where a test has mocked the session. In that case you couldn't really have used these methods before either as the read() or write() method would likely have been mocked.

PR #13989, opened 2019-12-07
Add getter/setter methods for currency

Zuluru:
This deprecates the `defaultCurrency` method and adds separate getter and setter methods, as discussed in #13893.


src/I18n/Number.php
```diff
@@ -237,30 +237,62 @@ public static function currency($value, $currency = null, array $options = [])
...
+        deprecationWarning(
+            'Number::defaultCurrency() is deprecated. ' .
+            'Use Number::setDefaultCurrency()/getDefaultCurrency() instead.'
+        );
 
         if ($currency === false) {
-            return self::$_defaultCurrency = null;
+            static::setDefaultCurrency(null);
+
+            // This doesn't sem like a useful result to return, but it's what the old version did.
```
markstory:
```suggestion
            // This doesn't seem like a useful result to return, but it's what the old version did.
```

Zuluru: D'oh! Typ'oh!

othercorey:
Can you fetch and merge 3.next to resolve the conflict?

PR #13992, opened 2019-12-08
Update limit

dereuromark:
I found this old branch, from 2016
Is it still useful? Or shall we delete it?


othercorey:
I need some sort of context to evaluate this.

othercorey:
I think we could re-create this if needed. I don't know the use-case so I can't say if it's complete.

PR #13993, opened 2019-12-09
4.x - Added missing docblocks for traits, classes and interfaces

othercorey:
Found generating api docs.

### DIFF
--- a/src/View/Widget/YearWidget.php
+++ b/src/View/Widget/YearWidget.php
@@ -21,10 +21,10 @@ use Cake\View\StringTemplate;
 use InvalidArgumentException;
 
 /**
- * Input widget class for generating a year select box.
+ * Input widget class for generating a calendar year select box.
  *
- * This class is intended as an internal implementation detail
- * of Cake\View\Helper\FormHelper and is not intended for direct use.
+ * This class is usually used internally by `Cake\View\Helper\FormHelper`,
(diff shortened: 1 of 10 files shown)


src/View/Widget/YearWidget.php
```diff
@@ -20,6 +20,12 @@
 use Cake\View\StringTemplate;
 use InvalidArgumentException;
 
+/**
+ * Input widget class for generating a year select box.
+ *
+ * This class is intended as an internal implementation detail
```
ADmad: This statement can be misleading. Someone might read it the class being an internal class which should not be extended, but that's not the case. Similar statement in other widgets should also be changed / removed.

othercorey: Let me know what you want it to say and I'll clean them up.

ADmad: I don't have a suggestion for rewording it, so I would just remove it.

Let's see what other's have to say.

markstory: The widget classes aren't entirely implementation details. If someone wanted to re-use this to make standalone year widgets they could. Perhaps something like

> This class can be used to generate standalone select boxes for choosing calendar years.

othercorey: I added that with a note about being used in FormHelper.  See if that works.

PR #14001, opened 2019-12-10
4.x - Add unit test for contain() with queryBuilder using limit() 

othercorey:
https://github.com/cakephp/docs/issues/4552

Is this not testing that example correctly?  It confirms `limit(1)` does not restrict the count of rows returned by the outer query.  It also confirms that the ordered select is restricted.

Is this not valid in general?

### DIFF
--- a/tests/TestCase/ORM/QueryTest.php
+++ b/tests/TestCase/ORM/QueryTest.php
@@ -2151,6 +2151,33 @@ class QueryTest extends TestCase
     }
 
     /**
+     * Ensure using limit() in the queryBuilder option works.
+     *
+     * @return void
+     */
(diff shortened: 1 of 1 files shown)


ndm2:
It is valid in the sense of that the query is valid and that you'll get a result, but it won't do what people might expect, that is applying a limit per author, ie retrieving 1 article per author, instead it will limit the overall number of retrieved articles to 1. So in this test the first author will have 1 article, all other authors won't have any articles at all.

othercorey:
The description from saeideng was that the limit(1) was hoisted to the outer query.  I'm not sure that's happening, but we should update the example to something simple that people can follow.

othercorey:
Closing as this was just for example.

PR #14010, opened 2019-12-11
Fixed preserving spaces in Text::slug

jeremyharris:
Previously the regex used a space as a temporary placeholder so it would be replaced regardless of its presence in the "preserve" option.

### DIFF
--- a/src/Utility/Text.php
+++ b/src/Utility/Text.php
@@ -1179,7 +1179,7 @@ class Text
             '/[' . $regex . ']/mu' => $options['replacement'],
             sprintf('/^[%s]+|[%s]+$/', $quotedReplacement, $quotedReplacement) => '',
         ];
-        if ($options['replacement']) {
+        if (isset($options['replacement']) && strlen($options['replacement'])) {
             $map[sprintf('/[%s]+/mu', $quotedReplacement)] = $options['replacement'];
         }
         $string = preg_replace(array_keys($map), $map, $string);
--- a/src/Utility/Text.php
+++ b/src/Utility/Text.php
@@ -1170,16 +1170,18 @@ class Text
             $string = static::transliterate($string, $options['transliteratorId']);
         }
 
-        $regex = '^\s\p{Ll}\p{Lm}\p{Lo}\p{Lt}\p{Lu}\p{Nd}';
+        $regex = '^\p{Ll}\p{Lm}\p{Lo}\p{Lt}\p{Lu}\p{Nd}';
         if ($options['preserve']) {
             $regex .= preg_quote($options['preserve'], '/');
         }
         $quotedReplacement = preg_quote($options['replacement'], '/');
         $map = [
-            '/[' . $regex . ']/mu' => ' ',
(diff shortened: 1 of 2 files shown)


src/Utility/Text.php
```diff
@@ -1170,16 +1170,18 @@ public static function slug($string, $options = [])
...
         }
         $quotedReplacement = preg_quote($options['replacement'], '/');
         $map = [
-            '/[' . $regex . ']/mu' => ' ',
-            '/[\s]+/mu' => $options['replacement'],
+            '/[' . $regex . ']/mu' => $options['replacement'],
             sprintf('/^[%s]+|[%s]+$/', $quotedReplacement, $quotedReplacement) => '',
         ];
+        if ($options['replacement']) {
+            $map[sprintf('/[%s]+/mu', $quotedReplacement)] = $options['replacement'];
```
othercorey: Why not just change the above pattern?  Also if this is checking for empty string can you use ===?

jeremyharris: I'm not sure what you mean by change the above pattern. This pattern replaces duplicates "replacements" with a single one after the first regexp.

I was under the impression we generally used loose checks for "empty" but I'm happy to add the strict equality check instead.

markstory: You could use `isset()`

jeremyharris: I don't think that will work for an empty string though.

markstory: It wouldn't you would have to do `isset() && strlen()`.

othercorey: The key is assumed to exist elsewhere with a default.  I think just `=== ''` works.

jeremyharris: Before these changes, `null` and `false` were valid replacement options (or at least the code didn't complain, just replaced with nothing) and this change is BC to that. I added test cases (which pass in master too) for those situations.

I'm fine with `isset() && strlen()`.

ADmad: The doc states `- replacement: Replacement string. Default '-'.`  So see no need to to account for `replacement` being set to `null` and hence `isset()` check is not necessary.

tests/TestCase/Utility/TextTest.php
```diff
@@ -1879,6 +1879,22 @@ public function slugInputProvider()
...
+            [
+                'Foo Bar: Not just for (breakfast) any-more', ['preserve' => ' ()'],
+                'Foo Bar- Not just for (breakfast) any-more',
+            ],
+            [
+                'Foo Bar: Not just for breakfast any-more', ['replacement' => null],
+                'FooBarNotjustforbreakfastanymore',
+            ],
+            [
+                'Foo Bar: Not just for breakfast any-more', ['replacement' => false],
```
ADmad: Since you are using `strlen()` setting `replacement` to `false` will generate type error when this is merged into 4.x.

jeremyharris: Nice catch. I changed the check once more to `is_string` instead of `isset`.

PR #14012, opened 2019-12-12
4.x - Fixed markdown lists in docblocks

othercorey:

### DIFF
--- a/src/Http/Response.php
+++ b/src/Http/Response.php
@@ -430,10 +430,12 @@ class Response implements ResponseInterface
      * Constructor
      *
      * @param array $options list of parameters to setup the response. Possible values are:
+     *
      *  - body: the response text that should be sent to the client
      *  - status: the HTTP status code to respond with
      *  - type: a complete mime-type string or an extension mapped in this class
      *  - charset: the charset for the response body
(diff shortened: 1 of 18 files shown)


src/Http/Response.php
```diff
@@ -430,10 +430,12 @@ class Response implements ResponseInterface
      * Constructor
      *
      * @param array $options list of parameters to setup the response. Possible values are:
+     *
```
dereuromark: In those cases they should be extracted to the top as own part IMO.

src/I18n/FrozenDate.php
```diff
@@ -145,10 +145,12 @@ public function __construct($time = 'now', $tz = null)
      * - `from` => another Date object representing the "now" date
      * - `format` => a fall back format if the relative time is longer than the duration specified by end
      * - `accuracy` => Specifies how accurate the date should be described (array)
+     *
```
dereuromark: do we really need a leading newline?
afaik markdown needs the trailing one.

othercorey: We'll see.  I'll re-adjust if it doesn't work somewhere.

Issue #14015, opened 2019-12-13
Lost timezone part of FrozenTime JSON serialization on libicu < 51


nojimage:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [x] feature-discussion (RFC)

* CakePHP Version: >=3.3
* Platform and Target: libicu < 51 (RHEL/CentOS 7, AmazonLinux2, etc)

### What you did

```php
echo FrozenTime::now()->jsonSerialize();
```

### What happened

```
2019-12-13T02:13:39
```

### What you expected to happen

```
2019-12-13T02:14:04+00:00
```

Lost timezone part.

I created a repository that reproduces this problem: [nojimage/cakephp-time-i18nformat-test](https://github.com/nojimage/cakephp-time-i18nformat-test)

#### 1. Use `ZZZZZ` instead of `xxx`

The difference between `ZZZZZ` and `xxx` is that `'Z'` is returned instead of `'+00:00'` when in UTC timezone.
'Z' is also correctly interpreted by the JavaScript `Date` class.

Quick fix: Add the below code in app's `config/bootstrap.php`,

```php
\Cake\I18n\FrozenTime::setJsonEncodeFormat("yyyy-MM-dd'T'HH':'mm':'ssZZZZZ");
\Cake\I18n\Time::setJsonEncodeFormat("yyyy-MM-dd'T'HH':'mm':'ssZZZZZ");
```

#### 2. Add `$useI18nFormat` to `setJsonEncodeFormat`'s 2nd argument

Add `$useI18nFormat` to `setJsonEncodeFormat`'s 2nd argument.
The param be used to control the formatting method.

```php
// `$useI18nFormat` to false, then jsonSerialize will use `format` method.
FrozenTime::setJsonEncodeFormat(DATE_ATOM, false);
// For backward compatibility, Default `$useI18nFormat` is `true`.
FrozenTime::setJsonEncodeFormat("yyyy-MM-dd'T'HH':'mm':'ssZZZZZ");
```

`format` is faster than `i18nFormat`:

```bash
docker-compose run --rm remi php /data/performance.php
Number of loops: 1,000,000
use i18nFormat: 8.725sec
use format: 0.854sec
```

I want to fix the problem with 2.

ADmad:
Screw these distros which still provide ICU version older than 51.

51.1 was released in *2013*.

> Version: 51.1 (Initial Public Release of ICU4C 51)
> Release Date: 2013-03-22

[...]

PR #14016, opened 2019-12-13
Time::setJsonEncodeFormat() can set a closure for json serialization

nojimage:
This param are determining format method when converting time object to json.

refs #14015

### DIFF
--- a/src/I18n/DateFormatTrait.php
+++ b/src/I18n/DateFormatTrait.php
@@ -52,7 +52,7 @@ trait DateFormatTrait
      * will be used for formatting the date part of the object and the second position
      * will be used to format the time part.
      *
-     * @var string|array|int
+     * @var string|array|int|callable
      * @see \Cake\I18n\Time::i18nFormat()
      */
     protected static $_jsonEncodeFormat = "yyyy-MM-dd'T'HH':'mm':'ssxxx";
@@ -284,8 +284,11 @@ trait DateFormatTrait
      * will be used for formatting the date part of the object and the second position
      * will be used to format the time part.
      *
+     * Alternatively, the format can provide a callback. In this case, the callback
+     * can receive this datetime object and return a formatted string.
+     *
      * @see \Cake\I18n\Time::i18nFormat()
-     * @param string|array|int $format Format.
+     * @param string|array|int|callable $format Format.
      * @return void
      */
     public static function setJsonEncodeFormat($format)
@@ -422,6 +425,10 @@ trait DateFormatTrait
      */
     public function jsonSerialize()
     {
+        if (is_callable(static::$_jsonEncodeFormat)) {
+            return call_user_func(static::$_jsonEncodeFormat, $this);
+        }
+
         return $this->i18nFormat(static::$_jsonEncodeFormat);
     }
 
--- a/tests/TestCase/I18n/DateTest.php
+++ b/tests/TestCase/I18n/DateTest.php
@@ -164,6 +164,23 @@ class DateTest extends TestCase
     }
 
     /**
+     * Tests change json encoding format
+     *
+     * @dataProvider classNameProvider
+     * @return void
+     */
+    public function testSetJsonEncodeFormat($class)
+    {
+        $date = new $class('2015-11-06 11:32:45');
+
+        $class::setJsonEncodeFormat(static function ($d) { return $d->format(DATE_ATOM); });
+        $this->assertEquals('"2015-11-06T00:00:00+00:00"', json_encode($date));
+
+        $class::setJsonEncodeFormat("yyyy-MM-dd'T'HH':'mm':'ssZZZZZ");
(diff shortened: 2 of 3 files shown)


tests/TestCase/I18n/DateTest.php
```diff
@@ -163,6 +163,23 @@ public function testJsonSerialize($class)
...
+     * @return void
+     */
+    public function testSetJsonEncodeFormat($class)
+    {
+        $date = new $class('2015-11-06 11:32:45');
+
+        $class::setJsonEncodeFormat(DATE_ATOM, false);
+        $this->assertEquals('"2015-11-06T00:00:00+00:00"', json_encode($date));
+
+        $class::setJsonEncodeFormat("yyyy-MM-dd'T'HH':'mm':'ssZZZZZ");
```
markstory: Is it unwise to change the default formatting to use the timezone specifier that works with a wider set of icu versions?

ADmad:
Based on the ICU version check in the test case it seems existing default format works for ICU 50.0+

https://github.com/cakephp/cakephp/blob/9ff52cf29f268ba80a35ccc35456a17f5bd8f33e/tests/TestCase/I18n/DateTest.php#L156-L164

ICU 50.1 was release on 2012-12-17

> Version: 50.1.2
> Release Date: 2012-12-17

http://site.icu-project.org/download/50

I see no reason to add unnecessary complexity to our code to support ICU versions older than 7 years. I am **strongly** against making any change to the core to support such edges cases. We can instead just updated the requirements check in app skeleton to check for ICU version too https://github.com/cakephp/app/blob/master/config/requirements.php#L30

tests/TestCase/I18n/DateTest.php
ADmad: @markstory Yes it would be unwise. Changing the default format could break API consumers which expect the datetime string in specific format.

src/I18n/DateFormatTrait.php
```diff
@@ -286,11 +293,13 @@ public static function setToStringFormat($format)
      *
      * @see \Cake\I18n\Time::i18nFormat()
      * @param string|array|int $format Format.
+     * @param bool $useI18nFormat When set to false, using `format` method when converting this object to json.
      * @return void
      */
-    public static function setJsonEncodeFormat($format)
+    public static function setJsonEncodeFormat($format, $useI18nFormat = true)
```
othercorey: I think adding a parameter here and defaulting to true will be hard to explain in the docblock.  Unfortunately, this function does not include i18n in the name like i18nFormat() does.

However, if you want a non-i18n serialization for speed, maybe a separate function to enable/disable might work.

nojimage:
In fact, the ICU version 50 is included in RHEL7 and it will be supported by 2024. I think should support this.
I think at least this issue and how to fix needs to be add a note to the documentation.

src/I18n/DateFormatTrait.php
ADmad: Switching between 2 types of placeholders is also a bad internal implementation.

ADmad:
https://github.com/cakephp/app/pull/752

markstory:
> In fact, the ICU version 50 is included in RHEL7 and it will be supported by 2024. I think should support this.

Is rhel not going to pull in additional libicu updates?

src/I18n/DateFormatTrait.php
nojimage: > I think adding a parameter here and defaulting to true will be hard to explain in the docblock.

I think so too.
How about adding a method `setJsonEncodeFormatAsNoneI18n($format)` ?

othercorey:
> > In fact, the ICU version 50 is included in RHEL7 and it will be supported by 2024. I think should support this.
> 
> Is rhel not going to pull in additional libicu updates?

FYI, I put in an official request for libicu update with AWS Linux2 through my org.  I hope other people have asked as it is a slow process and they have many deprecated php builds.

othercorey:
I was thinking about this situation.  I can't think of a way to switch between ICU and non-ICU format cleanly with the existing function names.

However, maybe we could allow setting a closure for json serialization.  That would be called in `jsonSerialize()` if set.  Then, you can format it however you want.

nojimage:
@othercorey sounds good.

Set like this?

```php
FrozenTime::setJsonEncodeFormat(function ($time) { return $time->format(DATE_ATOM); });
```

nojimage:
Alternatively, I think if the `_setClassName` method of DateTimeType is public, it will rewrite to a class that overridden `jsonSerialize`.

dereuromark:
ping @nojimage

nojimage:
I completely forgot about it.  😝 
I'll modify setJsonEncodeFormat to accept the callback, 
 like as https://github.com/cakephp/cakephp/pull/14016#issuecomment-568971163
Even tonight if possible.

nojimage:
I modified to accept callback.
thanks.

PR #14017, opened 2019-12-13
4.x - Added TableSchemaInterface::getPrimaryKey()

othercorey:
Shouldn't we align this name with other getters()?

### DIFF
--- a/src/Database/Schema/TableSchema.php
+++ b/src/Database/Schema/TableSchema.php
@@ -522,6 +522,7 @@ class TableSchema implements TableSchemaInterface, SqlGeneratorInterface
 
     /**
      * @inheritDoc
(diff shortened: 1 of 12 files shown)


src/Database/Schema/TableSchema.php
```diff
@@ -521,9 +521,21 @@ public function getIndex(string $name): ?array
     }
 
     /**
-     * @inheritDoc
+     * Get the column(s) used for the primary key.
+     *
+     * @return array Column name(s) for the primary key. An
+     *   empty list will be returned when the table has no primary key.
+     * @deprecated 4.0.0 Renamed to getPrimaryKey()
```
markstory: :+1: This works for me. Having a soft deprecation is safe at this point, we can add a yelling deprecation in 4.1

PR #14018, opened 2019-12-13
4.x - Added ContextInterface::getPrimaryKey() to use instead of primaryKey()

othercorey:
Shouldn't we align this name with other getters()?

### DIFF
--- a/src/View/Form/ArrayContext.php
+++ b/src/View/Form/ArrayContext.php
@@ -107,19 +107,7 @@ class ArrayContext implements ContextInterface
      */
     public function primaryKey(): array
     {
-        if (
-            empty($this->_context['schema']['_constraints']) ||
-            !is_array($this->_context['schema']['_constraints'])
-        ) {
-            return [];
-        }
-        foreach ($this->_context['schema']['_constraints'] as $data) {
-            if (isset($data['type']) && $data['type'] === 'primary') {
-                return (array)($data['columns'] ?? []);
-            }
-        }
-
-        return [];
+        return $this->getPrimaryKey();
     }
 
     /**
--- a/src/View/Form/ArrayContext.php
+++ b/src/View/Form/ArrayContext.php
@@ -103,8 +103,31 @@ class ArrayContext implements ContextInterface
      * Get the fields used in the context as a primary key.
      *
      * @return string[]
+     * @deprecated 4.0.0 Renamed to getPrimaryKey()
      */
     public function primaryKey(): array
+    {
+        if (
+            empty($this->_context['schema']['_constraints']) ||
+            !is_array($this->_context['schema']['_constraints'])
+        ) {
+            return [];
+        }
+        foreach ($this->_context['schema']['_constraints'] as $data) {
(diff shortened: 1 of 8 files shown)


ADmad:
I am not thrilled about unnecessary public API changes after 2 RCs. 

If we do want to want to add `getPrimaryKey()` then `primaryKey()` should be removed from the interface and kept only in the implementing classes with deprecated tag.

Same for #14017.

ADmad:
I am not thrilled about unnecessary public API changes after 2 RCs. 

If we do want to want to add `getPrimaryKey()` then `primaryKey()` should be removed from the interface and kept only in the implementing classes with deprecated tag.

Same for #14017.

othercorey:
> If we do want to want to add `getPrimaryKey()` then `primaryKey()` should be removed from the interface and kept only in the implementing classes with deprecated tag.

Agreed.  I'm only putting these up for awareness.  I defer to the team about what's correct.  I'll submit the requested changes.

dereuromark:
My main problem with primaryKey() is its unusefulness by default for most cases, since you get string|array.
I would much rather have a getFirstPrimaryKey() or getPrimaryKeyScalarOrFail() kind of wrappers added instead that you can use directly in code (or fail early) instead of having to array_shift() or other operations on the result until you can trust the "primary" use case (99%) of dealing with a single primary key. Most of my apps and plugins always need a lot of overhead logic to deal with this.

Just adding get prefix here for the sake of naming doesn't solve too much here for most use cases.

othercorey:
> Just adding get prefix here for the sake of naming doesn't solve too much here for most use cases.

Not dismissing your other points, but this is the reason I brought this up.  Reading this code, I thought what is different about `getPrimaryKey()` and `primaryKey()`?

This adds consistency which makes code more reliable.

```
class Table
{
    public function getPrimaryKey()
    {
        if ($this->_primaryKey === null) {
            $key = (array)$this->getSchema()->primaryKey();
            if (count($key) === 1) {
                $key = $key[0];
            }
            $this->_primaryKey = $key;
        }
```

dereuromark:
Sry, so it wouldnt be a complete alias, but the said scalar return?
I didnt check the code exactly, so I didnt see this.

PS: https://github.com/cakephp/cakephp/blob/4.x/src/Http/Session/DatabaseSession.php#L155-L156
We are still lying everywhere here - quiet badly IMO.
It would be cleaner to have the correct way and otherwise fail with meaningful exception early.

src/View/Form/ArrayContext.php
```diff
@@ -121,12 +122,34 @@ public function primaryKey(): array
         return [];
     }
 
+    /**
+     * Get the fields used in the context as a primary key.
+     *
+     * @return string[]
+     */
+    public function getPrimaryKey(): array
```
markstory: This looks like the same implementation as `primaryKey()` has. Why duplicate the code?

othercorey: Just an accident.  I left it in fixing a warning.

PR #14019, opened 2019-12-14
Remove code and tests for non string log message.

ADmad:
$message argument of LoggerInterface::log() must be a string.


src/Log/Engine/BaseLog.php
```diff
@@ -81,34 +79,17 @@ public function scopes()
...
+     * The context can optionally be used by log engines to interpolate variables
      * or add additional info to the logged message.
      *
-     * @param mixed $data The data to be converted to string and logged.
+     * @param string $message The message to be formatted.
      * @param array $context Additional logging information for the message.
      * @return string
      */
-    protected function _format($data, array $context = []): string
+    protected function _format(string $message, array $context = []): string
```
ADmad: Even though this methods simply returns the message augment now, I kept this method as it would be helpful in implementing this feature on 4.1 todo list.

> Improve log message creation so that context variables are formatted into the message string. This allows usage like $logger->log('info', 'Thing {color}', ['color' => $value])

src/Log/Engine/FileLog.php
```diff
@@ -116,7 +116,6 @@ public function __construct(array $config = [])
      */
     public function log($level, $message, array $context = []): void
```
othercorey:
```suggestion
    public function log($level, string $message, array $context = []): void
```

Can we do this?

ADmad: No we can't because `LoggerInterface::log()` doesn't have the `string` typehint.

othercorey: Ah yes, the PSR versioning issue.

src/Log/Engine/FileLog.php
```diff
@@ -116,7 +116,6 @@ public function __construct(array $config = [])
...
     {
-        $message = $this->_format($message, $context);
```
markstory: Shouldn't we still call format() to enable subclasses to do their own formatting?

ADmad: Yup this statement was mistakenly removed, reverted.

PR #14030, opened 2019-12-16
Use Arguments::getOption() instead of hasOption().

### LINKED ISSUE #14027: i18n extract command fails
This is a (multiple allowed):

* [x] bug
* [ ] enhancement

* CakePHP Version: 4.x-dev e9818adacf758b3057f4d7e3868c8b2791c6b3df
* Platform and Target: XAMPP, Windows 10, cmd.exe
* PHP 7.2.2 (cli) (built: Jan 31 2018 19:51:55) ( ZTS MSVC15 (Visual C++ 2017) x86 )

### What you did
(1) Installed latest CakePHP by "composer create-project --prefer-dist --stability=dev cakephp/app:4.x caketest"
(2) Moved to "caketest" folder
(3) Tried to extract i18n strings to .pot file from console by running "bin\cake i18n extract"
(4) This is the console output:

> C:\htdocs\caketest>bin\cake i18n extract
> Current paths: None
> What is the path you would like to extract?
> [Q]uit [D]one
> [C:\htdocs\caketest\src\] >
> 
> Current paths: C:\htdocs\caketest\src\
> What is the path you would like to extract?
> [Q]uit [D]one
> [C:\htdocs\caketest\templates\] >
> 
> Current paths: C:\htdocs\caketest\src\, C:\htdocs\caketest\templates\
> What is the path you would like to extract?
> [Q]uit [D]one
> [D] >
> 
> The output directory \ was not found or writable.
> 

### What happened
After prompting for paths following error is issued: "The output directory \ was not found or writable."

CakePHP to extract my localization strings into .pot file. This might be windows specific.

### Why it happened
Because following lines return true when they should not. There are no paths or plugin arguments. If I negate the condition in mentioned line, everything works fine.
https://github.com/cakephp/cakephp/blob/767b164747df789db8c0fbcf3bcf9eaf3be55b38/src/Command/I18nExtractCommand.php#L191
https://github.com/cakephp/cakephp/blob/767b164747df789db8c0fbcf3bcf9eaf3be55b38/src/Command/I18nExtractCommand.php#L193

ADmad:
That latter always returns true for any option declared in ConsoleOptionParser
regardless of whether value for the option is passed during command execution.

Closes #14027

### DIFF
--- a/src/Command/I18nExtractCommand.php
+++ b/src/Command/I18nExtractCommand.php
@@ -197,7 +197,7 @@ class I18nExtractCommand extends Command
             $this->_getPaths($io);
         }
 
-        if ($args->hasOption('extract-core')) {
+        if ($args->getOption('extract-core')) {
             $this->_extractCore = !(strtolower((string)$args->getOption('extract-core')) === 'no');
         } else {
             $response = $io->askChoice(
@@ -208,7 +208,7 @@ class I18nExtractCommand extends Command
             $this->_extractCore = strtolower((string)$response) === 'y';
         }
 
-        if ($args->hasOption('exclude-plugins') && $this->_isExtractingApp()) {
+        if ($args->getOption('exclude-plugins') && $this->_isExtractingApp()) {
             $this->_exclude = array_merge($this->_exclude, App::path('plugins'));
         }
 
@@ -216,9 +216,9 @@ class I18nExtractCommand extends Command
             $this->_paths[] = CAKE;
         }
 
-        if ($args->hasOption('output')) {
+        if ($args->getOption('output')) {
             $this->_output = (string)$args->getOption('output');
-        } elseif ($args->hasOption('plugin')) {
+        } elseif ($args->getOption('plugin')) {
             $this->_output = Plugin::path($plugin)
                 . 'resources' . DIRECTORY_SEPARATOR
                 . 'locales' . DIRECTORY_SEPARATOR;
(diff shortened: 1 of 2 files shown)


dereuromark:
Where there no test cases for this?

ADmad:
Seems all existing test cases explicitly pass values for the various options and thus default behavior isn't checked.

othercorey:
> That latter always returns true for any option declared in ConsoleOptionParser
> regardless of whether value for the option is passed during command execution.

Ewwww.

ADmad:
Maybe we should change `ConsoleOptionParser::parse()` instead to make `Arguments::hasOption()` actually useful?

dereuromark:
what is hasOption() supposed to do?

 > Check if an option is defined and not null.

For boolean values this currently could never work :) Sure.

markstory:
hasOption() isn't great for options with default values. Updating parse() so that hasOption() behaves correctly for options with no default sounds like a good idea. I can take care of that.

ADmad:
> hasOption() isn't great for options with default values

It's not just about options with default values. Even if options doesn't have a default value `hasOption()` always return's true as the array returned by `parser()` return the option set to empty string.

saeideng:
looks code changed in cake 4 
https://github.com/cakephp/cakephp/commit/e2b8505cd
now `hasOption()` checks by `isset()` and can be change to handle `''`

ADmad:
Closing in favor of #14034

PR #14031, opened 2019-12-16
Adding validation rules for ISBN numbers

burzum:


ADmad:
Requiring to validate ISBN numbers is a pretty niche use case. I don't enough people would require it to warrant having it in core. The related validators could easily go into a plugin.

Also in general new features now need to go into `4.next`. Only those deemed necessary to make upgrading from 3.x to 4.x easier will be backported to `3.next`.

dereuromark:
I agree, a simple plugin addition similar as https://github.com/cakephp/localized/tree/master/src/Validation sounds reasonable for this.

burzum:
🙄

PR #14034, opened 2019-12-17
Fix default value for console options

markstory:
Don't set a non-null default value for options. Doing so prevents `Arguments::hasOption()` from working correctly.


ADmad:
Refs #14027

ADmad:
I checked in an app and this change fixed the reported issue and pot file(s) are generated properly.

PR #14037, opened 2019-12-17
Fix CSRF tokens not being set on diactoros responses

markstory:
Update typehints and implementation to be compatible with diactoros response. Thankfully Cookie now has a `toHeaderValue()` method which enables CSRF tokens to be set on diactoros response objects.


src/Http/Middleware/CsrfProtectionMiddleware.php
```diff
@@ -199,8 +198,11 @@ protected function _addTokenCookie(string $token, ServerRequestInterface $reques
                 'httponly' => $this->_config['httpOnly'],
             ]
         );
+        if ($response instanceof Response) {
+            return $response->withCookie($cookie);
+        }
 
-        return $response->withCookie($cookie);
+        return $response->withAddedHeader('Set-Cookie', $cookie->toHeaderValue());
```
ADmad: :clap:

othercorey:
Restarted failed test. Looks good!

Issue #14038, opened 2019-12-17
Validation rule can not be added without a name


markusramsak:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 4.0.0

### What you did

this doesn't work:
```
public function validationDefault(Validator $validator): Validator
    {
        return $validator
            ->add('date_field', [
                ['rule' => 'date', 'provider' => 'custom'],
            ])
     }
```

this works:
```
public function validationDefault(Validator $validator): Validator
    {
        return $validator
            ->add('date_field', [
                'validationRuleName' => ['rule' => 'date', 'provider' => 'custom'],
            ])
     }
```

### What happened
Argument 1 passed to Cake\Validation\ValidationSet::add() must be of the type string, int given, called in vendor/cakephp/cakephp/src/Validation/Validator.php on line 475

### What you expected to happen
Validation rule added successfully without having to add a name

### Possible solution
revert add-method in ValidationSet class
from
`public function add(string $name, $rule)`
back to
`public function add($name, $rule)`

ADmad:
Why can't you just provide a name as expected?

ADmad:
Validations rules were always meant to be added with a name. It's just due to lack of strict typing in earlier versions of PHP and implicit type coercion that not providing a name (associative array key) didn't throw an error.

markusramsak:
I use CakePHP for 9 years now und I have really really large projects where I don't WANT to think about an extra validation rule name every time I add a validation rule to the validation set because the validation rule in most cases has a self explaining name. There is no benefit in forbidding to skip the name, so why don't just allow numeric indexes? I really don't see the benefit when I name the validation rule 'date' to 'date' when adding to validation set - it is just redundant in my opinion and it wasn't documented that I HAVE TO use a name, so please remove this requirement. I want to upgrade over 10 CakePHP 3.8 applications to CakePHP 4.0 und I would add names like '0' or '1' or '2' or 'date' (with no benefit at all). I think that the name of the application rules are self explaining and don't need an extra name in the validation set.

dereuromark:
We could add this to the shim plugin to auto add a name here if not given?

markusramsak:
But why force the name it in the first place? What is the benefit? I really don't see it ...

markusramsak:
[...]

PR #14041, opened 2019-12-18
Shim add for usability with numeric array

dereuromark:
This could mitigate the issues people face when upgrading:
https://github.com/cakephp/cakephp/issues/14038

It does indeed make not much sense to un-dry rules if we allow such an array config for setup of validation rules. Most of the time people have to use the same key twice, redundant.
So why not allowing them to auto use the rule name here as long as there are no collisions, and only then require more creative setup?

What do you think?
We can add then some tests for this.

### DIFF
--- a/src/Validation/Validator.php
+++ b/src/Validation/Validator.php
@@ -472,6 +472,13 @@ class Validator implements ArrayAccess, IteratorAggregate, Countable
             if (is_array($rule)) {
                 $rule += ['rule' => $name];
             }
+            if (!is_string($name)) {
+                $name = $rule['rule'];
+                if (isset($this->_rules[$name])) {
+                    throw new InvalidArgumentException('You cannot add a rule without a unique name, already existing rule found: ' . $name);
+                }
+            }
+
             $validationSet->add($name, $rule);
         }
 
--- a/src/Validation/Validator.php
+++ b/src/Validation/Validator.php
@@ -485,7 +485,7 @@ class Validator implements ArrayAccess, IteratorAggregate, Countable
                 }
 
                 deprecationWarning(
-                    'Using numeric indexes for adding validation rules is deprecated. Use string indexes instead.'
+                    'Adding validation rules without a name key is deprecated. Update rules array to have string keys.'
                 );
             }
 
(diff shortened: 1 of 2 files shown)


markusramsak:
my backwards compatible simple solution (if the name exists) was:
`if (is_int($name)) { $name = (string)$name; }`
then the names would be "0" oder "1" which I would expect.

dereuromark:
My concern with your simple solution: Then remove() and other methods that require a "known" key would not operate cleanly.
You cannot easily trust numeric indexes. As such I would recommend we use the rule name here to set the key, that is a compromise I think everyone can live with.

markusramsak:
I am fine with it

markusramsak:
And if the rule is an anonymous function you should always throw an error or generate a random name or cast the name to string.

```
['rule' => function ($value, $context) {
                    return $this->find()->where(['id' => $context['data']['id'], 'passwd' => (new WeakPasswordHasher())->hash($value)])->count() > 0;
                }, 'message' => 'password wrong']
```

markusramsak:
I think I will manually add a name for every validation rule added to the validation set to avoid issues ...

markusramsak:
Maybe it is the best if it throws an error which really explains what happend and what the developer has to change. For example: "You haven't provided a name to the validation rule. Please add one - we don't allow numeric indexes anymore."
Or maybe the Shim gives a deprecationWarning and you throw an error in 4.1 oder 5.0?

dereuromark:
I added a deprecation message - and some test cases for it.

markusramsak:
It still breaks anonymous function validation rules but the deprecation warning or error is already a lot less frustrating to find what I did wrong.

src/Validation/Validator.php
```diff
@@ -472,6 +473,22 @@ public function add(string $field, $name, $rule = [])
...
+                    $name = array_shift($name);
+                }
+
+                if ($validationSet->offsetExists($name)) {
+                    $message = 'You cannot add a rule without a unique name, already existing rule found: ' . $name;
+                    throw new InvalidArgumentException($message);
+                }
+
+                deprecationWarning(
+                    'Using numeric indexes for adding validation rules is deprecated. Use string indexes instead.'
```
othercorey:
```suggestion
                    'Adding validation rules without a name key is deprecated. Update rules array to have string keys.'
```

Maybe that's easier to read outside of the code?

dereuromark: I am fine either way.

PR #14046, opened 2019-12-19
prevent TypeError on preg_match when param is an integer

maxchene:
When trying to access : 
`http://localhost:8080/cars`

With this route : 
`
    $builder
        ->connect('/cars/{id}', ['controller' => 'Cars', 'action' => 'view'])
        ->setPatterns(['id' => '\d+'])
        ->setPass(['id']);
`

their's a TypeError from preg_match that require a string on second parameter

### DIFF
--- a/src/Routing/Route/Route.php
+++ b/src/Routing/Route/Route.php
@@ -707,7 +707,7 @@ class Route
         // check patterns for routed params
         if (!empty($this->options)) {
             foreach ($this->options as $key => $pattern) {
-                if (isset($url[$key]) && !preg_match('#^' . $pattern . '$#u', $url[$key])) {
+                if (isset($url[$key]) && !preg_match('#^' . $pattern . '$#u', (string)$url[$key])) {
                     return null;
                 }
             }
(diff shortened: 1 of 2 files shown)


ADmad:
I don't see how does that route end up generating integer values. We various similar routes in tests and there are no errors.

kaliel86:
Actually, on a fresh install : i have these 2 routes : 
`    $builder->connect('/cars', ['controller' => 'Cars', 'action' => 'index']);
    $builder
        ->connect('/cars/{id}', ['controller' => 'Cars', 'action' => 'view'])
        ->setPatterns(['id' => '\d+'])
        ->setPass(['id']);`

ADmad:
I tests using a fresh 4.0 app by creating a `CarsController` and setting the routes you have shown. I could access both `/cars` and `/cars/1` without any errors and the `index()` and `view()` methods were called respectively as expected.

kaliel86:
i created a new project and apprently it works as expected like you did.
The project where i got preg_match error only adds Authentication and Authorization middlewares.
I'll dig into it.

garas:
It fails when there is route defined with pattern. If there is no pattern, everything is OK.
```php
$route = new Route('/:controller/:action/:id', [], ['id' => '[0-9]+']);
// ok
$result = $route->match(['controller' => 'posts', 'action' => 'view', 'id' => '9']);
// fails
$result = $route->match(['controller' => 'posts', 'action' => 'view', 'id' => 9]);
```

@kaliel86 Could you add test with integer `'id' => 9` here:

https://github.com/cakephp/cakephp/blob/c6b9a2eb1be5760c1800cc48ec478c059b1992ac/tests/TestCase/Routing/Route/RouteTest.php#L1005-L1013

PR #14047, opened 2019-12-19
Fix silent association failures

### LINKED ISSUE #13467: belongsTo association with strategy=select
This is a (multiple allowed):

* [ ] bug
* [x] enhancement

* CakePHP Version: 3.8.

### What you did
In table: defined belongsTo association with strategy=select
In controller: feetched results without assoc foreign key in select expression:
`$this->Table->find()->select(['id', 'name'])->contain(['Assoc'])->first();`

### What happened
I got results from primary table, but property with assoc entity is null.

`RuntimeException` is thrown with message `'Could not find primary key value for source entity'`.

It works only when I specify assoc foreign key in select expression or skip select:
`$this->Table->find()->select(['id', 'name', 'assoc_id'])->contain(['Assoc'])->first();`
`$this->Table->find()->contain(['Assoc'])->first();`

markstory:
When selecting specific fields if foreign key columns are omitted, associations would silently fail to load. When we know that the source query returned results but that no foreign keys could be collected the ORM raises an error. This should help prevent missing associations. We do not raise an error for nested association loads as the nested association could be dependent on an optional association.

I can backport this fix to 3.x if folks are happy with it.

Fixes #13467

### DIFF
--- a/src/ORM/EagerLoader.php
+++ b/src/ORM/EagerLoader.php
@@ -653,7 +653,8 @@ class EagerLoader
                 // If the path or alias has no key the required assoication load will fail.
                 // Nested paths are not subject to this condition because they could
                 // be attached to joined associations.
-                if (strpos($path, '.') === false &&
+                if (
+                    strpos($path, '.') === false &&
                     (!array_key_exists($path, $collected) || !array_key_exists($alias, $collected[$path]))
                 ) {
                     $message = "Unable to load `{$path}` association. Ensure foreign key on `{$alias}` is selected.";
--- a/src/ORM/EagerLoader.php
+++ b/src/ORM/EagerLoader.php
@@ -650,14 +650,14 @@ class EagerLoader
 
             $requiresKeys = $instance->requiresKeys($config);
             if ($requiresKeys) {
-                // If the path or alias has no key the required assoication load will fail.
+                // If the path or alias has no key the required association load will fail.
                 // Nested paths are not subject to this condition because they could
                 // be attached to joined associations.
                 if (
                     strpos($path, '.') === false &&
                     (!array_key_exists($path, $collected) || !array_key_exists($alias, $collected[$path]))
                 ) {
-                    $message = "Unable to load `{$path}` association. Ensure foreign key on `{$alias}` is selected.";
+                    $message = "Unable to load `{$path}` association. Ensure foreign key in `{$alias}` is selected.";
                     throw new InvalidArgumentException($message);
                 }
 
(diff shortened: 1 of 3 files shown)


src/ORM/EagerLoader.php
```diff
@@ -644,8 +649,23 @@ public function loadExternal(Query $query, StatementInterface $statement): State
             $path = $meta->aliasPath();
 
             $requiresKeys = $instance->requiresKeys($config);
-            if ($requiresKeys && empty($collected[$path][$alias])) {
-                continue;
+            if ($requiresKeys) {
+                // If the path or alias has no key the required assoication load will fail.
```
othercorey:
```suggestion
                // If the path or alias has no key the required association load will fail.
```

src/ORM/EagerLoader.php
```diff
@@ -644,8 +649,23 @@ public function loadExternal(Query $query, StatementInterface $statement): State
...
+                // Nested paths are not subject to this condition because they could
+                // be attached to joined associations.
+                if (
+                    strpos($path, '.') === false &&
+                    (!array_key_exists($path, $collected) || !array_key_exists($alias, $collected[$path]))
+                ) {
+                    $message = "Unable to load `{$path}` association. Ensure foreign key on `{$alias}` is selected.";
```
othercorey:
```suggestion
                    $message = "Unable to load `{$path}` association. Ensure foreign key in `{$alias}` is selected.";
```

tests/TestCase/ORM/Association/BelongsToTest.php
```diff
@@ -427,4 +428,32 @@ public function testAttachToNoFieldsSelected()
...
+        $this->assertNotEmpty($result->author);
+        $this->assertSame(1, $result->author->id);
+
+        $query = $articles->find()
+            ->select(['Articles.title'])
+            ->where(['Articles.id' => 1])
+            ->contain('Authors');
+
+        $this->expectException(InvalidArgumentException::class);
+        $this->expectExceptionMessage('Unable to load `Authors` association. Ensure foreign key on `Articles`');
```
othercorey:
```suggestion
        $this->expectExceptionMessage('Unable to load `Authors` association. Ensure foreign key in `Articles`');
```

tests/TestCase/ORM/Association/HasManyTest.php
```diff
@@ -452,6 +454,25 @@ public function testEagerLoaderMultipleKeys()
...
+     *
+     * @return void
+     */
+    public function testEagerloaderNoForeignKeys()
+    {
+        $authors = $this->getTableLocator()->get('Authors');
+        $authors->hasMany('Articles');
+
+        $this->expectException(InvalidArgumentException::class);
+        $this->expectExceptionMessage('Unable to load `Articles` association. Ensure foreign key on `Authors`');
```
othercorey:
```suggestion
        $this->expectExceptionMessage('Unable to load `Articles` association. Ensure foreign key in `Authors`');
```

othercorey:
This looks good to me!  I think it covers a significant percentage of use cases.

PR #14048, opened 2019-12-19
Improve performance in eagerloader

markstory:
When loading external associations we need to enumerate all the records in the statement. Using fetchAll() results in fewer lines of code running and fewer allocations as the buffered results are not resized multiple times.

### DIFF
--- a/src/ORM/EagerLoader.php
+++ b/src/ORM/EagerLoader.php
@@ -809,7 +809,7 @@ class EagerLoader
     protected function _groupKeys(BufferedStatement $statement, array $collectKeys): array
     {
         $keys = [];
-        while ($result = $statement->fetch('assoc')) {
+        foreach ($statement->fetchAll('assoc') as $result) {
             foreach ($collectKeys as $nestKey => $parts) {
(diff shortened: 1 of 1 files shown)


src/ORM/EagerLoader.php
```diff
@@ -809,7 +809,7 @@ protected function _collectKeys(array $external, Query $query, $statement): arra
     protected function _groupKeys(BufferedStatement $statement, array $collectKeys): array
     {
         $keys = [];
-        while ($result = $statement->fetch('assoc')) {
+        foreach ($statement->fetchAll('assoc') as $result) {
```
othercorey:
```suggestion
        foreach (($statement->fetchAll('assoc') ?: []) as $result) {
```

There is `fetchAssoc()` but not all `fetchAssocAll()` :(

markstory: Thanks. I didn't consider the scenario where fetchAll() would return false.

othercorey:
That `$this->buffer[] = $record` is a real hidden gotcha.

PR #14049, opened 2019-12-19
Update Github CI workflow and travis config.

ADmad:


ADmad:
Do not merge. Codeverage reported aren't being uploaded to codecov. I'll look into it tomorrow.

ADmad:
So the CODECOV_TOKEN is not available to the runner due to a limitation of the service.

https://help.github.com/en/actions/automating-your-workflow-with-github-actions/creating-and-using-encrypted-secrets

> With the exception of GITHUB_TOKEN, secrets are not passed to the runner when a workflow is triggered from a forked repository.

Here's the related discussion https://github.community/t5/GitHub-Actions/Make-secrets-available-to-builds-of-forks/td-p/30678

The issue still exists even though that thread is marked as "Solved".

Guess we'll have to continue using Travis to submit code coverage report. I'll update accordingly.

ADmad:
Travis and GA config are sorted now. Should we get rid of Scrunitizer, I don't see much value in having it since we already run checks using 2 static analysers.

othercorey:
> Should we get rid of Scrunitizer?

I say remove it.  I don't think we use the metrics from it.

ADmad:
We don't have any scrutinizer config file in the repo. I turned off pull request tracking from it's site.

ADmad:
To summarize testsuite on PHP 7.2 and 7.4, static analysis and CS check jobs now run on GA workflow and testsuite on PHP 7.3 with code coverage reporting now runs on Travis.

PR #14056, opened 2019-12-20
Add support for placeholder in log messages.

ADmad:


src/Log/Engine/BaseLog.php
```diff
@@ -90,6 +91,46 @@ public function scopes()
...
+                    continue;
+                }
+
+                if (method_exists($value, 'toArray')) {
+                    $replacements['{' . $key . '}'] = print_r($value->toArray(), true);
+                    continue;
+                }
+            }
+
+            $replacements['{' . $key . '}'] = print_r($value, true);
```
ADmad: This `print_r()` causes PHP to run out of memory in tests :slightly_smiling_face:.

Similar statement exists in 3.x too but I guess it doesn't have any code coverage.
https://github.com/cakephp/cakephp/blob/2f68c080ad27ecb9c9ee19e58961497a75703f7e/src/Log/Engine/BaseLog.php#L110

So I guess the best course of action is to simply ignore objects which cannot be serialized to string or json encoded.

tests/test_app/TestApp/Log/Engine/TestBaseLog.php
```diff
@@ -12,17 +12,27 @@
...
-     * @return string
+     * @return void
      */
     public function log($level, $message, array $context = [])
     {
-        return $this->_format($message, $context);
+        $this->message = $this->_format($message, $context);
+    }
+
+    public function getMessage()
```
dereuromark: We should add string typehint here also for test methods. This avoids accidental usage as other type etc.

src/Log/Engine/BaseLog.php
```diff
@@ -90,6 +91,46 @@ public function scopes()
      */
     protected function _format(string $message, array $context = []): string
     {
-        return $message;
+        if (strpos($message, '{') === false && strpos($message, '}') === false) {
+            return $message;
+        }
+
+        preg_match_all('/\{([a-z][a-z0-9-_]*)\}/i', $message, $matches);
```
markstory: Do you want to use `Text::insert()`? It would require adding cakephp/utility as a dependency, and pre-casting data into strings.

src/Log/Engine/BaseLog.php
markstory: I'm :+1: for removing this code. If we find a non-scalar value that can't be formatted can we include `[unhandled value of $typeOfThing]` placeholder so the user can figure out why data is missing?

tests/TestCase/Log/Engine/BaseLogTest.php
```diff
@@ -52,8 +56,36 @@ private function assertUnescapedUnicode(array $needles, $haystack)
...
+            'string' => 'a-string',
+            'bool' => true,
+            'json' => new Entity(['foo' => 'bar']),
+            'array' => ['arr'],
+            'obj' => function () {
+            },
+        ];
+        $this->logger->log(
+            LogLevel::INFO,
+            '1: {string}, 2: {bool}, 3: {json}, 4: {not a placeholder}, 5: {array}, 6: {obj}, 8: {valid-ph-not-in-context}',
```
markstory: Should we support `.` separated values to create paths? We could use `Hash::get()` to pull values out. I'm also ok with only supporting one dimensional arrays to stay simple.

src/Log/Engine/BaseLog.php
ADmad: > It would require adding cakephp/utility as a dependency

That's the reason I didn't use it. The code of Text::insert() looks lot more expensive than this regex check.

tests/TestCase/Log/Engine/BaseLogTest.php
dereuromark: I think simple Should be fine.

ADmad: I too don't think we need to get all fancy here and add dependency on utility package for being able to use `Hash::get()`.

othercorey:
What happens if a user tries to log something like their own placeholder system text? Can you escape/disable this?

ADmad:
@stickler-ci You are a dumb bot, I am gonna slap you silly. There is no CS error, the GA CS job is passing.

ADmad:
@othercorey I am not sure how trying to escape placeholders would work. We could add an option to display placeholder parsing though.

othercorey:
Can we at least skip placeholders if they don't exist in the context for now? That would let someone pass-through existing text with {} in it.

ADmad:
> Can we at least skip placeholders if they don't exist in the context for now?

That's already done. Check the test cases. The string `{valid-ph-not-in-context}` is left unchanged because there's no matching key in context array, even though it is matched as a valid placeholder.

src/Log/Engine/BaseLog.php
```diff
@@ -90,6 +93,66 @@ public function scopes()
...
+        foreach ($placeholders as $key) {
+            $value = $context[$key];
+
+            if (is_scalar($value)) {
+                $replacements['{' . $key . '}'] = (string)$value;
+                continue;
+            }
+
+            if (is_array($value)) {
+                $replacements['{' . $key . '}'] = print_r($value, true);
```
markstory: Could this have the same memory exhaustion error that we had before? Should we use json_encode here instead?

ADmad: It potentially could. There are few other `print_r()` uses too, should I replace them all with `json_encode()`?

othercorey:
> That's already done. Check the test cases. The string `{valid-ph-not-in-context}` is left unchanged because there's no matching key in context array, even though it is matched as a valid placeholder.

My mistake. I missed the array_intersection call.

src/Log/Engine/BaseLog.php
markstory: I think it depends on how print_r() is being used. I personally find its multiline output annoying in logs especially when they are forwarded via syslog. JSON in that scenario is easier to manipulate downstream.

ADmad: I don't like its multi line output either. I used it just cause that's what was done in the 3.x code. I'll replace it with json_encode().

ADmad: Done

ADmad:
@othercorey I added support for allowing the placeholder to be escaped.

I also relaxed the regex to allow placholder string to start with underscore or digit since such strings are valid array keys.

src/Log/Engine/BaseLog.php
```diff
@@ -98,7 +98,7 @@ protected function _format(string $message, array $context = []): string
         }
 
         preg_match_all(
-            '/(?<!' . preg_quote('\\', '/') . ')\{([a-z][a-z0-9-_]*)\}/i',
+            '/(?<!' . preg_quote('\\', '/') . ')\{([a-z0-9-_]+)\}/i',
```
othercorey: Does this allow `-` in the place holder name?

Issue #13293, opened 2019-05-27
MSSQL query having condition on aggregated aliased field


inspectorman:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.6.11

### What you did
I have been trying to add a Having condition for an aliased max field in a Query, I'm using MSSQL Server for the database.
The select looks like:
`'diam_1'=>$q->func()->max('(tmls.nom_diam)'),`
and the having function looks like:
`$q->having(['diam_1 >='=>5])`

### What happened
Sql Srv throughs an error telling that diam_1 is not a recognized field.

### What you expected to happen
I would expect that query builder would detect that the diam_1 is actually an aggregated field and must convert the having statement to: Having(max(tmls.nom_diam)>=5) as it does with order clauses on aggregated named fields.

markstory:
I don't think this was ever implemented. If you want to build it I can help with tests.

othercorey:
> convert the having statement to: Having(max(tmls.nom_diam)>=5) as it does with order clauses on aggregated named fields

There is code that replaces a column alias with the expression sql in order by clauses?

markstory:
Closing as there is a pull request open for this now.

PR #14057, opened 2019-12-20
Test having on aggregated field

Zwartpet:
Fix an issue where using Having on a aggregated field was not working in all database drivers

https://github.com/cakephp/cakephp/issues/13293


src/Database/Dialect/SqlserverDialectTrait.php
```diff
@@ -368,11 +368,11 @@ public function rollbackSavePointSQL($name)
     /**
      * {@inheritDoc}
      *
-     * @return \Cake\Database\SqlserverCompiler
+     * @return \Cake\Database\PostgresCompiler
      */
     public function newCompiler()
     {
-        return new SqlserverCompiler();
+        return new PostgresCompiler();
```
ADmad: Why is the `SqlserverDialectTrait` returning a `PostgresCompiler`?

src/Database/PostgresCompiler.php
```diff
@@ -0,0 +1,83 @@
...
+                    $matches
+                );
+
+                if (empty($matches[0])) {
+                    continue;
+                }
+
+                $parts[$k] = preg_replace(
+                    ['/"/', '/\b' . trim($selectKey, '"') . '\b/i'],
+                    ['', $selectPart->sql($generator)],
```
markstory: Will this still work if the function contains placeholders? For example `SUM(COALESCE(value, 1)` where the expression equivalent would be 
```php
$col = $query->func()->sum(
  $query->func()->coalesce(['author_id' => 'identifier', 1])
);
```

othercorey:
@Zwartpet Are you going to follow up on this?

Zwartpet:
I'll look into it in the upcoming days

src/Database/PostgresCompiler.php
Zwartpet:
```
$query
            ->select([
                'posts.author_id',
                'post_count' => $query->func()->count('posts.id'),
            ])
            ->group(['posts.author_id'])
            ->having([$query->newExpr()->gte('post_count', 2, 'integer')]);
```

This is in the test cases, don't know if this covers your question? Can't think of a case you would use function in the having condition

Zwartpet:
@othercorey any idea why i get `Error : Undefined class constant` on `PDO::SQLSRV_ATTR_ENCODING` when i rebase on 4.next (see failing tests)

tests/TestCase/Database/Driver/PostgresTest.php
```diff
@@ -173,4 +173,38 @@ public function testInsertReturning()
...
     }
+
+    /**
+     * Test that having queries replace the aggregated alias field.
+     *
+     * @return void
+     */
+    public function testHavingReplacesAlias()
+    {
+        $driver = $this->getMockBuilder('Cake\Database\Driver\Sqlserver')
```
markstory: This test should be in the Driver/SqlserverTest.php file instead. That is likely why you're getting undefined constant errors.

Zwartpet: Thanks for the hint, the mocked driver was wrong indeed. But the constant errors were there because the mocked functions were not prepended with underscore anymore.

Zwartpet:
Done, if there are no further comments this can be merged

markstory:
Thanks :+1:

PR #14061, opened 2019-12-21
Supress PHPUnit ReflectionType deprecation warnings on PHP 7.4

garas:
Fix 2 `Trying to access array offset on value of type null` warnings which revealed after suppressing reflection errors.

Supress `ReflectionType` deprecation errors when generating mocks in `TestCase::getMockBuilder() -> MockBuilder::getMock()`, `TestCase::getMockClass()` and `TestCase::getMockForAbstractClass()`.

Plugins, which uses `Cake\TestSuite\TestCase` in their tests, will get suppressed warnings too.

### DIFF
--- a/src/Database/Expression/TupleComparison.php
+++ b/src/Database/Expression/TupleComparison.php
@@ -96,7 +96,7 @@ class TupleComparison extends Comparison
             if ($isMulti) {
                 $bound = [];
                 foreach ($value as $k => $val) {
-                    $valType = $multiType ? $type[$k] : $type;
+                    $valType = $multiType && isset($type[$k]) ? $type[$k] : $type;
                     $bound[] = $this->_bindValue($generator, $val, $valType);
                 }
 
--- a/src/TestSuite/MockBuilder.php
+++ b/src/TestSuite/MockBuilder.php
@@ -1,0 +1,39 @@
+<?php
+/**
+ * CakePHP(tm) : Rapid Development Framework (https://cakephp.org)
+ * Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
+ *
+ * Licensed under The MIT License
+ * For full copyright and license information, please see the LICENSE.txt
+ * Redistributions of files must retain the above copyright notice.
+ *
+ * @copyright     Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
+ * @link          https://cakephp.org CakePHP(tm) Project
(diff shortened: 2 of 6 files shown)


src/Database/Expression/TupleComparison.php
```diff
@@ -96,7 +96,7 @@ protected function _stringifyValues($generator)
             if ($isMulti) {
                 $bound = [];
                 foreach ($value as $k => $val) {
-                    $valType = $multiType ? $type[$k] : $type;
+                    $valType = $multiType && isset($type[$i]) ? $type[$k] : $type;
```
markstory: Why do you check for `$i` when the read is using `$k`?

src/TestSuite/MockBuilder.php
```diff
@@ -0,0 +1,37 @@
...
+ * @link          https://cakephp.org CakePHP(tm) Project
+ * @since         3.8.8
+ * @license       https://opensource.org/licenses/mit-license.php MIT License
+ */
+namespace Cake\TestSuite;
+
+use PHPUnit\Framework\MockObject\MockBuilder as BaseMockBuilder;
+
+/**
+ * PHPUnit MockBuilder with muted Reflection errors
```
markstory: Should we flag this as `@internal`?

src/Database/Expression/TupleComparison.php
garas: Just stupid copy/paste...

src/TestSuite/MockBuilder.php
garas: Done

markstory:
Great work. This is a simple solution compared to the alternatives.

dereuromark:
You might need some shimming for PHP5.6 packages it seems
I have now these fails in 5.6 builds: https://travis-ci.org/dereuromark/cakephp-dto/jobs/630604416

> Fatal error:  Class 'PHPUnit\Framework\MockObject\MockBuilder' not found in /home/travis/build/dereuromark/cakephp-dto/vendor/cakephp/cakephp/src/TestSuite/MockBuilder.php on line 25

garas:
Try to add `loadPHPUnitAliases()`. 

Aliases autoloaded in `FixtureInjector`, `FixtureManager` and `TestSuite`, and `tests/bootstrap.php`. It seems your plugin doesn't use any of these. Adding this to shim `MockBuilder` can be good idea.

Issue #13666, opened 2019-09-21
Translate behavior sets `dirty` on contained property if the parent doesn't have the behavior


Zuluru:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.8.4
* Platform and Target: PHP7.2.21

### What you did
To use an example from core unit tests, Articles belongs to Authors. Add Translate behavior to Authors, but not Articles. (Assume that in the actual use case, there's nothing worth translating in Articles.) Load an Article (or Articles; it doesn't seem to matter if you use `find` or `get`), containing Authors.

### What happened
The Article entity has the author property flagged as dirty.

### What you expected to happen
The author property should not be flagged as dirty.

If Articles has the Translate behavior on it, or if neither Articles nor Authors does, or if you use `loadInto` instead of `contain`, the author property is correctly not flagged.

This can be easily replicated in the core tests by commenting out `$table->addBehavior('Translate', ['fields' => ['title', 'body']]);` (and therefore also `$table->setLocale('eng');`) in `Cake\Test\TestCase\ORM\Behavior\TranslateBehaviorTest::testFindSingleLocaleBelongstoLoadInto`.

Zuluru:
Note, I have only tested this so far on the `belongsTo` association. Not sure if it affects others.

Zuluru:
I've done a little more digging into this, and found the following. Hopefully, it helps someone to track down and resolve the issue; I don't know nearly enough about ORM internals to fix it.

I looked at where `setDirty` is being called from. In the case where the error occurs, the call stack looks like this:

```
Table.php:1624, App\Model\Table\GamesTable->get()
QueryTrait.php:483, Cake\ORM\Query->firstOrFail()
QueryTrait.php:472, Cake\ORM\Query->first()
CollectionTrait.php:386, Cake\Datasource\ResultSetDecorator->first()
CollectionTrait.php:386, LimitIterator->rewind()
CollectionTrait.php:386, Cake\Datasource\ResultSetDecorator->rewind()
BufferedIterator.php:112, Cake\Collection\Iterator\BufferedIterator->rewind()
BufferedIterator.php:112, Cake\Collection\Iterator\BufferedIterator->rewind()
BufferedIterator.php:112, Cake\Datasource\ResultSetDecorator->rewind()
InsertIterator.php:118, Cake\Collection\Iterator\InsertIterator->current()
[...]

PR #14062, opened 2019-12-21
Add test cases to demonstrate issue 13666

Zuluru:
Test cases to demonstrate issue #13666. Tests are included for both `loadInto` and `contain` methods, for all four types of containment. Some pass, some do not.

### DIFF
--- a/tests/TestCase/ORM/Behavior/TranslateBehaviorTest.php
+++ b/tests/TestCase/ORM/Behavior/TranslateBehaviorTest.php
@@ -828,6 +828,113 @@ class TranslateBehaviorTest extends TestCase
     }
 
     /**
+     * Tests that parent entity isn't dirty when containing a translated association
+     *
+     * @return void
+     */
+    public function testGetAssociationNotDirtyBelongsTo()
+    {
+        $table = $this->getTableLocator()->get('Articles');
+        $authors = $table->belongsTo('Authors')->getTarget();
+        $authors->addBehavior('Translate', ['fields' => ['name']]);
+
+        $authors->setLocale('eng');
+
+        $entity = $table->get(1);
+        $this->assertNotEmpty($entity);
+        $entity = $table->loadInto($entity, ['Authors']);
+        $this->assertFalse($entity->isDirty());
+        $this->assertNotEmpty($entity->author);
+        $this->assertFalse($entity->author->isDirty());
+
+        $entity = $table->get(1, ['contain' => ['Authors']]);
+        $this->assertNotEmpty($entity);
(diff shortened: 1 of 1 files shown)


tests/TestCase/ORM/Behavior/TranslateBehaviorTest.php
```diff
@@ -827,6 +827,113 @@ public function testFindSingleLocaleBelongsToMany()
...
+    /**
+     * Tests that parent entity isn't dirty when containing a translated association
+     *
+     * @return void
+     */
+    public function testGetAssociationNotDirtyBelongsToMany()
+    {
+        $table = $this->getTableLocator()->get('Articles');
+        $specialTags = $this->getTableLocator()->get('SpecialTags');
+        $specialTags->addBehavior('Translate', ['fields' => ['extra_info']]);
```
markstory: Should the translate behavior be on Tags and not on the junction table?

markstory:
I was able to track down the source of the problem for this issue, but haven't had the time to think of a non-terrible fix. The `TranslateBehavior` appends additional associations and result formatters in its `beforeFind` here https://github.com/cakephp/cakephp/blob/5bfb78f1ecf4e37872ecfc49619e27ca48790860/src/ORM/Behavior/TranslateBehavior.php#L261-L264

Because association `beforeFind` handlers are fed 'dummy' queries by the association `attachTo` method, the ORM also rebinds any result formatters here https://github.com/cakephp/cakephp/blob/5bfb78f1ecf4e37872ecfc49619e27ca48790860/src/ORM/Association.php#L1257-L1286

The `results->insert()` call is the culprit here as it sets the modified author entities onto the articles records once their translation fields have been shuffled around. The Article entity set() method sees this as a property change flagging the property as dirty.

tests/TestCase/ORM/Behavior/TranslateBehaviorTest.php
Zuluru: I copied the basic setup of this test from `testFindSingleLocaleBelongsToMany`.

Zuluru:
I found in the exploration of the original bug report that `set` seems to be used a lot, which sets the entity as dirty, but then various things clean it after that. For example, the ReplaceIterator is used (and results in this cleaning) when both the parent and child have the TranslateBehavior on them, but not in this case. So, not sure that the bug is really that it's being set as dirty so much as it's not being reset as it's supposed to be after this.

Or maybe all this dirtying and cleaning is an indication that there should be an extra flag that gets passed around in such cases to skip setting things as dirty in the first place when initially reading them? Not sure if that would be an improvement or a violation of separation of concerns.

Issue #13978, opened 2019-12-04
4.x - Email with links with route raises MissingRouteException on tests


mentisy:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 4.x-dev https://github.com/cakephp/cakephp/commit/7751403267ba4c8196cc444d70ac3877eac0911b
* Platform and Target: Attempted on Windows and Linux

### What you did
1. Downloaded the latest version of CakePHP 4.x-dev using composer create-project cakephp/app:4.x-dev
2. Baked a Mailer
3. Added an HTML link it the mailer template trying to route to `['controller' => 'pages', 'action' => 'display', 'home' `
4. Inside the baked MailerTest I tried to assert that an email was sent and contained `pages/home`

### What happened
A `Cake\Routing\Exception\MissingRouteException` is raised saying `A route matching ['controller' => 'pages', 'action' => 'display', 0 => 'home'] could not be found`

### What you expected to happen
I expected the test to pass as an email was sent and contained the link routed.

### Reproduce raised MissingRouteException ###
* I cloned this repository, ran all the tests inside `tests/TestCase/Mailer/` using `bin/vendor/phpunit tests/TestCase/Mailer` and they all passed.
* I added
`<?php echo $this->Html->link('Test', ['controller' => 'pages', 'action' => 'display', 'home']);`
 to the bottom of `templates/email/html.php` and ran the tests again. This raised the MissingRouteException.

### Full stack trace ###
```
4) Cake\Test\TestCase\Mailer\MailerTest::testSendRenderWithHTML
Cake\Routing\Exception\MissingRouteException: A route matching "array (
  'controller' => 'pages',
  'action' => 'display',
  0 => 'home',
  'plugin' => NULL,
  '_ext' => NULL,
)" could not be found.

X:\xampp\htdocs\cakephp-mail\src\Routing\RouteCollection.php:341
X:\xampp\htdocs\cakephp-mail\src\Routing\Router.php:494
X:\xampp\htdocs\cakephp-mail\src\View\Helper\UrlHelper.php:51
X:\xampp\htdocs\cakephp-mail\src\View\Helper\HtmlHelper.php:257
X:\xampp\htdocs\cakephp-mail\tests\test_app\templates\email\html\html.php:16
X:\xampp\htdocs\cakephp-mail\src\View\View.php:1160
X:\xampp\htdocs\cakephp-mail\src\View\View.php:1121
[...]

PR #14067, opened 2019-12-23
Add helper method to load routes in tests

markstory:
Add a helper method to load routes similar to how we have helper methods to load plugins. This will make testing mailers and helpers simpler and require less duplicate code as the real application routes can be used.

Refs #13978

### DIFF
--- a/src/TestSuite/TestCase.php
+++ b/src/TestSuite/TestCase.php
@@ -29,6 +29,7 @@ use Cake\TestSuite\Constraint\EventFired;
 use Cake\TestSuite\Constraint\EventFiredWith;
 use Cake\Utility\Inflector;
 use PHPUnit\Framework\TestCase as BaseTestCase;
+use ReflectionClass;
 use RuntimeException;
 
 /**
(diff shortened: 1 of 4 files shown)


tests/TestCase/TestSuite/TestCaseTest.php
```diff
@@ -468,4 +465,24 @@ public function testGetMockForModelSetTable()
...
+     * @return void
+     */
+    public function testLoadRoutes()
+    {
+        $url = ['controller' => 'Articles', 'action' => 'index'];
+        try {
+            Router::url($url);
+        } catch (MissingRouteException $e) {
+            $this->assertTrue(true, 'Route should fail.');
+        }
```
saeideng: `assertTrue` now  is useless 
could changes to this
```php
        try {
            Router::url($url);
            $this->assertTrue(false, 'Route should fail.');
        } catch (MissingRouteException $e) {
            
        }
```

src/TestSuite/TestCase.php
```diff
@@ -195,6 +197,33 @@ public function loadFixtures(): void
...
+     * `Cake\TestSuite\IntegrationTestCaseTrait` to better simulate all routes
+     * and plugins being loaded.
+     *
+     * @param array|null $appArgs Constuctor parameters for the application class.
+     * @return void
+     * @since 4.0.1
+     */
+    public function loadRoutes(?array $appArgs = null): void
+    {
+        $appArgs = $appArgs === null ? [CONFIG] : $appArgs;
```
ADmad: The trailing slash from `CONFIG` needs to be removed as all methods of `BaseApplication` that use `$this->configDir` again add a slash.

https://github.com/cakephp/cakephp/blob/9886fa7dcef259a1ff497a5b8fdf739401499c89/src/Http/BaseApplication.php#L139

src/TestSuite/TestCase.php
```diff
@@ -195,6 +198,33 @@ public function loadFixtures(): void
...
+    public function loadRoutes(?array $appArgs = null): void
+    {
+        $appArgs = $appArgs ?? [rtrim(CONFIG, DIRECTORY_SEPARATOR)];
+        $className = Configure::read('App.namespace') . '\\Application';
+        try {
+            $reflect = new ReflectionClass($className);
+            /** @var \Cake\Routing\RoutingApplicationInterface $app */
+            $app = $reflect->newInstanceArgs($appArgs);
+        } catch (ReflectionException $e) {
+            throw new LogicException(sprintf('Cannot load "%s" to load routes from.', $className), null, $e);
```
othercorey:
```suggestion
            throw new LogicException(sprintf('Cannot load "%s" to load routes from.', $className), 0, $e);
```

PR #14079, opened 2019-12-27
4.1 - Added AggregateExpression for aggregate functions

othercorey:

This is the first part of window function support. This adds the necessary classes and API in Query to start building.

The AggregateExpression class will support the extended API for window functions.  We can also add full support for extended aggregate expression clauses in the future.

~The existing aggregate functions in FunctionsBuilder will still work, but they will not be extendable. Eventually, we can deprecate them in favor of the extensible versions.~

As there is no new behavior in this copy, the unit tests are just copying the FunctionExpression tests.

### DIFF
--- a/src/Database/Expression/AggregateExpression.php
+++ b/src/Database/Expression/AggregateExpression.php
@@ -1,0 +1,27 @@
+<?php
+declare(strict_types=1);
+
+/**
+ * CakePHP(tm) : Rapid Development Framework (https://cakephp.org)
+ * Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
+ *
+ * Licensed under The MIT License
+ * For full copyright and license information, please see the LICENSE.txt
+ * Redistributions of files must retain the above copyright notice.
+ *
+ * @copyright     Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
+ * @link          https://cakephp.org CakePHP(tm) Project
+ * @since         4.1.0
+ * @license       https://opensource.org/licenses/mit-license.php MIT License
+ */
+namespace Cake\Database\Expression;
+
+/**
+ * This represents a SQL aggregate function expression in a SQL statement.
+ * Calls can be constructed by passing the name of the function and a list of params.
+ * For security reasons, all params passed are quoted by default unless
+ * explicitly told otherwise.
+ */
+class AggregateExpression extends FunctionExpression
+{
+}
--- a/src/Database/FunctionsBuilder.php
+++ b/src/Database/FunctionsBuilder.php
@@ -16,6 +16,7 @@ declare(strict_types=1);
  */
 namespace Cake\Database;
 
+use Cake\Database\Expression\AggregateExpression;
 use Cake\Database\Expression\FunctionExpression;
 use InvalidArgumentException;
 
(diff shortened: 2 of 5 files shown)


src/Database/AggregatesBuilder.php
```diff
@@ -0,0 +1,149 @@
...
+class AggregatesBuilder
+{
+    /**
+     * Returns a AggregateExpression representing a call to SQL SUM function.
+     *
+     * @param mixed $expression the function argument
+     * @param array $types list of types to bind to the arguments
+     * @return \Cake\Database\Expression\AggregateExpression
+     */
+    public function sum($expression, $types = []): AggregateExpression
```
markstory: Would the planned window related methods be available on the sum() expression?

markstory (commented):
Do we need to separate the builders for generic functions and aggregates? It sounded like we would be able to expose the required windowing features on a the window expression class, without requiring more builders.

src/Database/AggregatesBuilder.php
othercorey: Yes, any supported aggregate function can be extend to a window function.

Even without that, we can add FILTER support to this interface for non-window functions.

othercorey:
> Do we need to separate the builders for generic functions and aggregates? It sounded like we would be able to expose the required windowing features on a the window expression class, without requiring more builders.

Sure, that's up for debate.  The separate builder does two things:

1. Ensure an AggregateExpression is returned and typed
2. Avoid user mistakes when mixing non-aggregate functions with aggregate clauses and window clauses.

markstory:
> Ensure an AggregateExpression is returned and typed

We could do this with the current function builder too if each window-able function was typed to the window expression type.

How do you see the API working for this approach on window functions? How would someone use them in a query?

othercorey:
> We could do this with the current function builder too if each window-able function was typed to the window expression type.

How do we support the arbitrary function building with that?  That's the main reason I switched to this.  I'm not going to wrap every aggregate and window function.

> 
> How do you see the API working for this approach on window functions? How would someone use them in a query?

There are 2 common approaches I had in mind.  There are some caveats with the vendor implementations and nuances on customization to the window when in query vs column and with duplicate names/etc.  I'm hoping to hit the common stuff.

```
// set existing window
$sum = $query->aggregate()->sum('field')->window('name' or $window);
```

```
// build window
$sum = $query->aggregate()->sum('field');
$sum
    ->window()
    ->partition()
    ->order()
    ->frame();
```

You could add a separate window to the query:

```
$query->window('name', $window);
```

markstory:
> How do we support the arbitrary function building with that? That's the main reason I switched to this. I'm not going to wrap every aggregate and window function.

Ok. Using the function builder would require using the aggregate expression class in specific functions. We could also have a `window()` (bad name) as a way to support other function names.

With having another builder I would be concerned with the two sum() methods being different.

othercorey:
> With having another builder I would be concerned with the two sum() methods being different.

Maybe there is another approach.  I don't think having `aggregate()` just for the call builder will be clean.  Users won't understand why `func()` for pre-defined is different from `aggregate()`.

To avoid diverging implementations/double implementation, why don't we give `FunctionsBuilder` an `AggregatesBuilder` instance that it uses to build the existing aggregate functions wrapped by `FunctionBuilder`.  That way they both use the same implementation and require the same interface and types.

This would also fix the issue where current `func()->sum()` expressions would not be extendable to window functions.

markstory:
> Maybe there is another approach. I don't think having aggregate() just for the call builder will be clean. Users won't understand why func() for pre-defined is different from aggregate().

Are you referring to the `__call` method? Could we have a `buildAggregate()` that returns an `AggregateExpression` instead of the vanilla `FunctionExpression`? I don't think we need to have `__call` support for aggregates if we have a build method. We can add more concrete methods as requests come in.

othercorey:
> I don't think we need to have `__call` support for aggregates if we have a build method. We can add more concrete methods as requests come in.

That feels like I'm going to break someone.  

What about the option of just having `FunctionExpression` support everything and leave it up to the user if they added the wrong option for a non-aggregate function?

markstory:
> What about the option of just having FunctionExpression support everything and leave it up to the user if they added the wrong option for a non-aggregate function?

Is the scenario you're concerned about

```php
$query->func()->customaggregate()->window()
```

Not working? Wouldn't the `__call` return type let the developer know that this would fail as the return of `__call` is a `FunctionExpression`? We have concrete functions on the `FunctionsBuilder` for aggregate functions (`sum`, `count` etc). Those functions could return an `AggregateFunctionExpression` which supports the new windowing features. If users need more functions with windowing we can either add more concrete methods to `FunctionsBuilder` or those people can use 

```php
$query->func()->buildAggregate('customaggregate', ...)->window()
```

othercorey:
- Removed `AggregatesBuilder`
- Updated `FunctionsBuilder` to return `AggregateExpression` for existing aggregates
- Added `FunctionsBuilder::aggregate()` to support building arbitrary aggregates.

markstory (approved):
Looks good to me. We can expand this in future pull requests :+1:

PR #14081, opened 2019-12-27
4.1 - Renamed database schema dialect classes

othercorey:
Renamed to `Dialect` to avoid confusion with `TableSchema`.  These classes are already referred to as `dialect` in the code.

Marked the dialect classes as internal.

This helps clarify the dependencies when trying to walk through the code.


src/Database/Dialect/MysqlDialectTrait.php
```diff
@@ -47,7 +47,7 @@ trait MysqlDialectTrait
     /**
      * The schema dialect class for this driver
      *
-     * @var \Cake\Database\Schema\MysqlSchema
+     * @var \Cake\Database\Schema\Dialect\MysqlDialect
```
markstory: Should this be `Cake\Database\Dialect\MysqlSchemaDialect` ? Four namespaces feels like a lot.

src/Database/Schema/Dialect/MysqlDialect.php
```diff
@@ -14,14 +14,17 @@
  * @since         3.0.0
  * @license       https://opensource.org/licenses/mit-license.php MIT License
  */
-namespace Cake\Database\Schema;
+namespace Cake\Database\Schema\Dialect;
```
markstory: Historically we have added class_alias() stubs to the old file location. Having aliases prevents breaking any external uses of these classes and give us a good spot to put deprecation warnings.

src/Database/Dialect/MysqlDialectTrait.php
othercorey: I was torn on that.  The database namespaces are a bit of a mess IMO.

We have `Database\Driver` but then `Database\Dialect` which is a driver trait.  If we move `MysqlSchemaDialect` into `Database\Dialect` would that be clear?

It probably could be as long as we use the full name `MysqlSchemaDialect` and not `MysqlDialect`.

othercorey: The `MysqlDialectTrait` should probably be `MysqlDialectDriverTrait` to match.

othercorey:
Moved files to `Database\Dialect` and added class aliases for old locations.

src/Database/Dialect/MysqlDialectTrait.php
markstory: > Maybe it shouldn't even be a trait.

Probably not. The drivers wereobne of the places we thought that traits could  help fix deep inheritance trees. It has created another set of problems though :smile:

othercorey: I think the methods/properties in the trait should just be in the driver.  They aren't shared by anything.

src/Database/Dialect/MysqlDialectTrait.php
```diff
@@ -16,8 +16,7 @@
  */
 namespace Cake\Database\Dialect;
 
-use Cake\Database\Schema\BaseSchema;
-use Cake\Database\Schema\MysqlSchema;
+use Cake\Database\Schema\SchemaDialect;
```
ADmad: Why not have `SchemaDialect` class under `Database\Dialect` too?

othercorey: I thought the cake pattern was to put the base class in the parent directory.

ADmad: Well it's not in parent directory is it? :slightly_smiling_face: `Dialect` and `Schema` are sibling directories.

othercorey: It was before moving to Dialect is why I left it there. Easy fix.

src/Database/Dialect/MysqlSchemaDialect.php
```diff
@@ -621,3 +623,8 @@ protected function _keySql(string $prefix, array $data): string
         return $prefix . ' (' . implode(', ', $columns) . ')';
     }
 }
+
+// phpcs:disable
+// @deprecated 4.1.0 Add backwards compatible alias.
+class_alias('Cake\Database\Dialect\MysqlSchemaDialect', 'Cake\Database\Schema\MysqlSchema');
```
markstory: This stub needs to be in ` src/Database/Schema/MysqlSchema.php` so when the old class is autoloaded composer loads the shim that sets an alias and autoloads the new class name.

src/Database/Dialect/PostgresSchemaDialect.php
```diff
@@ -654,3 +656,8 @@ public function dropTableSql(TableSchema $schema): array
         return [$sql];
     }
 }
+
+// phpcs:disable
+// @deprecated 4.1.0 Add backwards compatible alias.
+class_alias('Cake\Database\Dialect\PostgresSchemaDialect', 'Cake\Database\Schema\PostgresSchema');
```
markstory: Same for this alias shim.

src/Database/Dialect/SchemaDialect.php
```diff
@@ -283,3 +284,8 @@ public function convertOptionsDescription(TableSchema $schema, array $row): void
      */
     abstract public function truncateTableSql(TableSchema $schema): array;
 }
+
+// phpcs:disable
+// @deprecated 4.1.0 Add backwards compatible alias.
+class_alias('Cake\Database\Dialect\SchemaDialect', 'Cake\Database\Schema\BaseSchema');
```
markstory: Same for this alias shim.

src/Database/Dialect/SqliteSchemaDialect.php
```diff
@@ -583,3 +585,8 @@ public function hasSequences(): bool
         return $this->_hasSequences;
     }
 }
+
+// phpcs:disable
+// @deprecated 4.1.0 Add backwards compatible alias.
+class_alias('Cake\Database\Dialect\SqliteSchemaDialect', 'Cake\Database\Schema\SqliteSchema');
```
markstory: Same for this alias shim.

src/Database/Dialect/SqlserverSchemaDialect.php
```diff
@@ -653,3 +656,8 @@ public function truncateTableSql(TableSchema $schema): array
         return $queries;
     }
 }
+
+// phpcs:disable
+// @deprecated 4.1.0 Add backwards compatible alias.
+class_alias('Cake\Database\Dialect\SqlserverSchemaDialect', 'Cake\Database\Schema\SqlserverSchema');
```
markstory: Same for this alias shim.

src/Database/Dialect/MysqlSchemaDialect.php
garas: `class_alias()` in new file, and `class_exists()` in old file.

For same reasons as in 946478e7a129d4a403580f31310bbbede4817680.

othercorey:
Added `class_exist` to old file bases on @garas comment.  This matches the behavior of the Cake\Network to Cake\Http move. I missed the class_exist part when following the alias example.

othercorey:
Moved the schema dialect classes back to `Cake\Database\Schema` namespace. I didn't make sense to have them in a non-schema namespace.  The difference is not putting them into a Dialect namespace.

If we remove the weird driver traits from `Cake\Database\Dialect` then we can remove that weird namespace completely.

This change now just renames the dialect classes appropriately.

src/Database/DriverInterface.php
```diff
@@ -180,9 +180,9 @@ public function queryTranslator(string $type): Closure;
      * If all the tables that use this Driver specify their
      * own schemas, then this may return null.
      *
-     * @return \Cake\Database\Schema\BaseSchema
+     * @return \Cake\Database\Schema\SchemaDialect
      */
-    public function schemaDialect(): BaseSchema;
+    public function schemaDialect(): SchemaDialect;
```
ADmad: Change return type would be a non BC change even of a new minor.

othercorey:
This will have to wait.

src/Database/DriverInterface.php
garas: There is `class_alias('Cake\Database\Schema\SchemaDialect', 'Cake\Database\Schema\BaseSchema');` so it shouldn't be so BC.

3.x had same with `Response` etc

ADmad: Oops, sorry I overlooked the fact that it's just a rename.

othercorey:
Moving files that you have to deprecate just makes a mess.

othercorey:
This has the db specific dialects marked internal after renaming.

PR #14086, opened 2019-12-28
Validate table schema against max alias length

othercorey:
https://github.com/cakephp/cakephp/pull/13830

Until we can generate temporary alias names and use them when matching results, this prevents aliases from being truncated quietly causing unknown errors.

This is really only valid for PostgreSQL and SqlServer where the max alias length is the same as column identifier length.

### DIFF
--- a/src/Database/Driver/Mysql.php
+++ b/src/Database/Driver/Mysql.php
@@ -31,6 +31,11 @@ class Mysql extends Driver
     use MysqlDialectTrait;
 
     /**
+     * @var int|null Maximum alias length or null if no limit
+     */
+    protected const MAX_ALIAS_LENGTH = 256;
+
+    /**
      * Base configuration settings for MySQL driver
      *
      * @var array
--- a/src/ORM/Table.php
+++ b/src/ORM/Table.php
@@ -19,6 +19,7 @@ namespace Cake\ORM;
 use ArrayObject;
 use BadMethodCallException;
 use Cake\Core\App;
+use Cake\Core\Configure;
 use Cake\Database\Connection;
 use Cake\Database\Schema\TableSchemaInterface;
 use Cake\Database\TypeFactory;
@@ -498,6 +499,9 @@ class Table implements RepositoryInterface, EventListenerInterface, EventDispatc
                     ->getSchemaCollection()
                     ->describe($this->getTable())
             );
+            if (Configure::read('debug')) {
+                $this->checkAliasLengths();
+            }
         }
 
         return $this->_schema;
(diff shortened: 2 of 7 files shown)


src/Database/Driver/Sqlite.php
```diff
@@ -32,6 +32,11 @@ class Sqlite extends Driver
 {
     use SqliteDialectTrait;
 
+    /**
+     * @var int|null Maximum alias length or null if no limit
+     */
+    protected const MAX_ALIAS_LENGTH = null;
```
ADmad: Instead of this we can just add a default `getMaxAliasLength()` implementation in `Driver` class which returns `null`.

src/ORM/Table.php
```diff
@@ -530,10 +531,40 @@ public function setSchema($schema)
...
         return $this;
     }
 
+    /**
+     * Validates the identifiers are valid for generating queries.
+     *
+     * @return void
+     * @throws \RuntimeException When identifiers are invalid
+     */
+    protected function validateSchema(): void
```
ADmad: The method name is pretty non-intuitive.

src/ORM/Table.php
```diff
@@ -530,10 +531,40 @@ public function setSchema($schema)
...
+        if (method_exists($this->getConnection()->getDriver(), "getMaxAliasLength")) {
+            $maxLength = $this->getConnection()->getDriver()->getMaxAliasLength();
+        }
+
+        if ($maxLength) {
+            $table = $this->getAlias();
+            foreach ($this->_schema->columns() as $name) {
+                if (strlen($table . '__' . $name) > $maxLength) {
+                    throw new RuntimeException(
+                        "Unable to use table `{$table}` and column `{$name}` without truncating names."
```
ADmad: Error message should also suggest how the problem can be fixed, that is by using a shorter alias for the table.

src/ORM/Table.php
```diff
@@ -530,10 +531,40 @@ public function setSchema($schema)
...
+        if ($this->_schema === null) {
+            throw new RuntimeException("Unable to validate `{$this->getAlias()}` without schema.");
+        }
+
+        $maxLength = null;
+        if (method_exists($this->getConnection()->getDriver(), "getMaxAliasLength")) {
+            $maxLength = $this->getConnection()->getDriver()->getMaxAliasLength();
+        }
+
+        if ($maxLength) {
```
ADmad: You can check for `null` instead and return early.

src/ORM/Table.php
```diff
@@ -530,10 +531,40 @@ public function setSchema($schema)
...
+            $table = $this->getAlias();
+            foreach ($this->_schema->columns() as $name) {
+                if (strlen($table . '__' . $name) > $maxLength) {
```
ADmad: This feels like a lot of unnecessary overheard. I would suggest simply skipping calling this method when debug is turned off.

othercorey: Agreed. This is the minimal place I found to check, but I didn't want to try caching the result in the schema.

src/ORM/Table.php
```diff
@@ -530,10 +534,46 @@ public function setSchema($schema)
...
+            return;
+        }
+
+        $table = $this->getAlias();
+        foreach ($this->_schema->columns() as $name) {
+            if (strlen($table . '__' . $name) > $maxLength) {
+                $nameLength = $maxLength - 2;
+                throw new RuntimeException(
+                    "The combination of `{$table}` and `{$name}` must be less than {$nameLength}. " .
+                    "You must shorten the one of those names before `Table` can generate queries."
```
markstory: This will be a hard failure for some folks, and they won't be able to use the ORM without schema changes. While I think the early failure is good we might want to suggest that they alter the column name, or rename the table to be more explicit on what the next step should be.

src/Database/Driver/Mysql.php
```diff
@@ -185,4 +190,15 @@ public function supportsNativeJson(): bool
...
         return $this->_supportsNativeJson = version_compare($this->_version, '5.7.0', '>=');
     }
+
+    /**
+     * Returns the maximum alias length allowed.
+     * This can be different than the maximum identifier length for columns.
+     *
+     * @return int|null Maximum alias length or null if no limit
+     */
+    public function getMaxAliasLength(): ?int
```
ADmad: This exact same code is still being repeated in 3 classes. I guess it would be better to define the constant also in `Driver` and just override it's value in child classes and make `Driver::getMaxAliasLength()` return the constant value.

othercorey: Is that ok for user-defined drivers? They wouldn't know to override the constant so we'd have to avoid using `MAX_ALIAS_LENGTH` outside of the `getMaxAliasLength()` wrapper, right?

I don't mind it.  Just want to make sure I don't introduce anything too painful.

src/ORM/Table.php
othercorey: By hard failure you mean their ORM is broken and they don't realize why? I put it in Table because any query you make at the ORM level needs this to work, right?

src/Database/Driver/Mysql.php
ADmad: Even with currently implementation any custom drivers extending `Driver` will have their `getMaxAliasLength()`. So overriding the constant is no worse than overriding the method.

othercorey:
Updated exception message to explain more.

Created MAX_ALIAS_LENGTH constant in base Driver class.

othercorey:
Should we move this to 4.1 or is the error critical enough to add to 4.0 as early as possible?

markstory:
I think this is fine for a big fix. The previous behavior was incorrect and now we are letting folks know what is wrong.

PR #14089, opened 2019-12-28
Fix optional routing keys with braced placeholders

### LINKED ISSUE #13997: possible Router BC break
This is a (multiple allowed):

* [x] bug
* [ ] enhancement

* CakePHP Version: 3.8 latest

### What you did
Travis tests on cakephp-tags

### What happened
Now red after the release of
https://github.com/cakephp/cakephp/compare/3.8.6...3.8.7#diff-37dbf1f85d9888de3ac3c50006f2704fR1023-R1024

```
1) Tags\Test\TestCase\View\Helper\TagCloudHelperTest::testDisplayUlLi
--- Expected
+++ Actual
@@ @@
-'<ul class="tag-cloud"><li style="font-size: 80%"><a href="/?by=Foo" id="tag-1">Foo</a></li><li style="font-size: 160%"><a href="/?by=Bar" id="tag-2">Bar</a></li><li style="font-size: 120%"><a href="/?by=X-Y-Z" id="tag-3">X Y Z</a></li></ul>'
+'<ul class="tag-cloud"><li style="font-size: 80%"><a href="/{controller}?by=Foo" id="tag-1">Foo</a></li><li style="font-size: 160%"><a href="/{controller}?by=Bar" id="tag-2">Bar</a></li><li style="font-size: 120%"><a href="/{controller}?by=X-Y-Z" id="tag-3">X Y Z</a></li></ul>'
/home/travis/build/dereuromark/cakephp-tags/vendor/cakephp/cakephp/src/TestSuite/TestCase.php:317
/home/travis/build/dereuromark/cakephp-tags/tests/TestCase/View/Helper/TagCloudHelperTest.php:109
```
https://travis-ci.org/dereuromark/cakephp-tags/jobs/622451557?utm_medium=notification&utm_source=github_status

Is this a regression, if the URLs change in a way that they dont fit expected one anymore?

markstory:
Optional parameters were not working correctly with braced parameters. While non-core parameters can be optional the controller + action parameters should never be null as that will result in invalid URLs generated or incorrect results that only contain the action parameter.

I've removed a few invalid test cases that were providing keys as both in the template and as a default. This has not worked for a long time and these tests were relying on the now invalid behavior.

Fixes #13997

### DIFF
--- a/src/Routing/Route/Route.php
+++ b/src/Routing/Route/Route.php
@@ -753,6 +753,14 @@ class Route
         }
         $url += $hostOptions;
 
+        // Ensure controller/action keys are not null.
+        if (
+            (isset($keyNames['controller']) && !isset($url['controller'])) ||
+            (isset($keyNames['action']) && !isset($url['action']))
+        ) {
+            return false;
+        }
+
         return $this->_writeUrl($url, $pass, $query);
     }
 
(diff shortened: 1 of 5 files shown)


dereuromark:
With this I had to fix my tests including routes.
But the prefer lowest now seems to have issues: https://travis-ci.org/dereuromark/cakephp-calendar/builds/630675605

PR #14090, opened 2019-12-29
Fix trace/context for errors wrapped by debug kit

markstory:
Add a magic `_trace_frame_offset` context value that allows error handling wrappers to offset the stack trace that is displayed. This is not a great solution, but given how PHP error handling works and our backwards compatibility requirements this is what I could come up with.

I'm open to suggestions or alternate approaches we could take.

### DIFF
--- a/src/Error/BaseErrorHandler.php
+++ b/src/Error/BaseErrorHandler.php
@@ -148,9 +148,19 @@ abstract class BaseErrorHandler
 
         $debug = Configure::read('debug');
         if ($debug) {
+            // By default trim 3 frames off for the public and protected methods
+            // used by ErrorHandler instances.
+            $start = 3;
+
+            // Can be used by error handlers that wrap other error handlers
+            // to coerce the generated stack trace to the correct point.
+            if (isset($context['_trace_offset'])) {
+                $start += $context['_trace_offset'];
+                unset($context['_trace_offset']);
+            }
             $data += [
                 'context' => $context,
-                'start' => 3,
+                'start' => $start,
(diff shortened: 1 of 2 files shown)


src/Error/BaseErrorHandler.php
```diff
@@ -148,9 +148,19 @@ public function handleError($code, $description, $file = null, $line = null, $co
 
         $debug = Configure::read('debug');
         if ($debug) {
+            // By default trim 3 frames off for the public and protected methods
+            // used by ErrorHandler instances.
+            $start = 3;
+
+            // Can be used by error handlers that wrap other error handlers
+            // to coerce the generated stack trace to the correct point.
+            if (isset($context['_trace_offset'])) {
```
othercorey: I would probably make this name more direct.  `_trace_frame_offset` and change `start` to `start_frame`.

It could then be described as an offset from the starting frame which makes sense with the math here.

markstory: Changing `start` would be a breaking change as it modifies the API of `Debugger::outputError()`.

othercorey: I meant to say would be nice to change `start`. I still think making this name mention `frame` is helpful for when random cake devs debug in the future.

othercorey:
@markstory I didn't merge this because I saw the debug_kit PR was using `_trace_frame_offset`. Were you going to change this one?

markstory:
Derp. I thought I had changed this one too. I will update this code when I get to a real computer.

PR #14092, opened 2019-12-29
allow IdentifierExpression as argument in Function

swiffer:
The following code was valid in CakePHP 3.x

```php
$query->func()->dateAdd(new IdentifierExpression($this->aliasField('valid_until')), 1, 'day'), (new FrozenTime('now')), 'datetime'),
```

but is invalid as $expression has enforcement of type string now. However this affects date functions only. It's still fine for other functions so this proposed to revert this change

### DIFF
--- a/src/Database/FunctionsBuilder.php
+++ b/src/Database/FunctionsBuilder.php
@@ -186,11 +186,11 @@ class FunctionsBuilder
      * Returns the specified date part from the SQL expression.
      *
      * @param string $part Part of the date to return.
-     * @param string $expression Expression to obtain the date part from.
+     * @param mixed $expression Expression to obtain the date part from.
      * @param array $types list of types to bind to the arguments
      * @return \Cake\Database\Expression\FunctionExpression
      */
-    public function datePart(string $part, string $expression, array $types = []): FunctionExpression
+    public function datePart(string $part, $expression, array $types = []): FunctionExpression
     {
         return $this->extract($part, $expression, $types);
     }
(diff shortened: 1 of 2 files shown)


ADmad:
Change seems fine. Can you please add the example you have shown as test to avoid regression in future?

swiffer:
extended the test for dateAdd - adding this to all functions tested seems to much imo

src/Database/FunctionsBuilder.php
```diff
@@ -186,11 +186,11 @@ public function dateDiff(array $args, array $types = []): FunctionExpression
      * Returns the specified date part from the SQL expression.
      *
      * @param string $part Part of the date to return.
-     * @param string $expression Expression to obtain the date part from.
+     * @param mixed $expression Expression to obtain the date part from.
```
ADmad: Using `string|\Cake\Database\ExpressionInterface` would be better than `mixed`.

ADmad:
> extended the test for dateAdd - adding this to all functions tested seems to much imo

That's fine.

> is a release date for 4.0.1 planned already?

Very soon :)

src/Database/FunctionsBuilder.php
ADmad: I'll take care of it post merge.

markstory:
> is a release date for 4.0.1 planned already?

I am hoping to do it tonight. :smile:

swiffer:
Yeah :) can you Take a Look at #14095 before, might be a Regression but Not 100% Sure about that

PR #10821, opened 2017-06-24
Attempt no. 2 to fix PHPUnit aliases

jeremyharris:
Refs: https://github.com/cakephp/cakephp/issues/10814

For reasons I cannot figure out (probably gremlins), autoloading in the plugin from the original issue happens in a different order than that of the cake core. 

To help alleviate this discrepancy, I created a new helper function `loadPHPUnitAliases()` which is guaranteed to exist on all apps and plugins, which in turn uses `require_once` to load a file that defines aliases. If any future breaks happen in PHPUnit 🤞  we should be able to add aliases to the file and use the global function in all files that require said aliases.

I went with `require_once` instead of checking that classes exist because I was still getting redeclare errors even when wrapping them in `class_exists`.

PR #14098, opened 2019-12-29
Autoload PHPUnit aliases in MockBuilder

garas:
Fix https://github.com/cakephp/cakephp/pull/14061#issuecomment-569508411

Tests might fail if PHPUnit aliases not loaded. Extend https://github.com/cakephp/cakephp/pull/10821.

### DIFF
--- a/src/TestSuite/MockBuilder.php
+++ b/src/TestSuite/MockBuilder.php
@@ -14,6 +14,8 @@
  */
 namespace Cake\TestSuite;
 
+loadPHPUnitAliases();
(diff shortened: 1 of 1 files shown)


dereuromark:
I tested it on the plugin, same issue: https://travis-ci.org/dereuromark/cakephp-dto/jobs/630652107
Might I be missing something else?

dereuromark:
I tested it and it works :)
Now we need another patch release soonish to make plugin tests green here.

Issue #14085, opened 2019-12-28
Test with FormProtectionComponent enabled


swiffer:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 4.0.1
* Platform and Target: PHP 7.4

### What you did
Enabled FormProtectionComponent in AppController.php
```
$this->loadComponent('FormProtection');
```
Enable security Token in TestSuite
```
$this->enableSecurityToken();
```

### What happened
Tests fail

### What you expected to happen
Tests should pass?

P.S. Remember, an issue is not the place to ask questions. You can use [Stack Overflow](https://stackoverflow.com/questions/tagged/cakephp)
for that or join the #cakephp channel on irc.freenode.net, where we will be more
than happy to help answer your questions.

Before you open an issue, please check if a similar issue already exists or has been closed before.

ADmad:
Sounds like duplicate to #13925.

Does your `tests/bootstrap.php` have the following statement to fixate session id https://github.com/cakephp/app/blob/master/tests/bootstrap.php#L52?

swiffer:
The statement in `tests/bootstrap.php` is present:
```php
// Fixate sessionid early on, as php7.2+
// does not allow the sessionid to be set after stdout
// has been written to.
session_id('cli');
```

Form protection is enabled in AppController::initialize
```php
$this->loadComponent('FormProtection');
```

TestCase method has:
```php
$this->enableCsrfToken();
$this->enableSecurityToken();
```

markstory:
I'll see if I can reproduce this problem in our tests.

swiffer:
I'm not unlocking a field in the particular request. Updated to "dev-master as 4.0.2" but still the issue persists. Read the docs again:

https://book.cakephp.org/4/en/development/testing.html#testing-actions-protected-by-csrfcomponent-or-securitycomponent

> It is also important to enable debug in tests that use tokens to prevent the SecurityComponent from thinking the debug token is being used in a non-debug environment.

swiffer:
Having debug = true when running the testsuite causes excessive memory usage and process timeout. This seems to be related to DebugKit, when the plugin is disabled this does not happen.

I've now disabled DebugKit in cli via

src/Application.php

```php
        /*
         * Only try to load DebugKit in development mode
         * Debug Kit should not be installed on a production system
         */
        if (Configure::read('debug') && PHP_SAPI !== 'cli') {
[...]

PR #14100, opened 2019-12-30
Fix FormProtection not comparing unlocked fields properly

markstory:
While looking into #14085 further I started off by updating our integration tests to use the FormProtectionComponent and an existing test was failing. The failing test ensured that tokens generated with no unlocked fields are not accepted by controller actions that unlock fields. This scenario was failing as token generation under FormProtectionComponent was not forwarding the unlockedFields configuration.


ADmad:
Opps, my bad :stuck_out_tongue:

PR #14102, opened 2019-12-30
Pass URL and session ID as args to token data building method.

ADmad:
This gets token building inline with token verification and makes API more intuitive.
Earlier it wasn't clear that passing URL and ID to constructor was only necessary for token
data building.

`FormProtector` is marked as internal as we are still quite early in 4.0 release cycle so I don't think modifying it's API should be a problem.


markstory:
> FormProtector is marked as internal as we are still quite early in 4.0 release cycle so I don't think modifying it's API should be a problem.

We have no promises around internal classes. They are effectively private as far as I'm concerned.

Issue #13493, opened 2019-08-08
3.next: Backport 4.x functionality for easier migration


dereuromark:
This is a (multiple allowed):

* [ ] bug
* [x] enhancement

in 4.x we added API that should ideally be backported into 3.next (3.9).
The migration guide is quite empty so far ( https://github.com/cakephp/docs/blob/3.next/en/appendices/3-9-migration-guide.rst ). We should fix that - and make the diff between 3.x and 4.x easier for future migration then.

According to docs (catch all ticket here to tick off):

ORM
---

* [x] isNew() => setNew()

* [x] ``Table::saveManyOrFail()`` method has been added that will throw ``PersistenceFailedException``
  with the specific entity that failed in case of an error. The entities are saved transaction safe.
* [x] ``Table::deleteMany()`` and ``Table::deleteManyOrFail()`` methods have been added for removing many
  entities at once including callbacks. The entities are removed transaction safe.

( https://github.com/cakephp/cakephp/pull/13212 )

Http
----

* [x] ``Cake\Http\Client\Response::isSuccess()`` was added. This method returns true
  if the response status code is 2xx.
* [x] ``CspMiddleware`` was added to make defining Content Security Policy headers simpler.
* [ ] Support for `SameSite` attribute in cookies. #13766 #13806

TestSuite
---------

* [x] ``TestSuite\EmailTrait::assertMailContainsAttachment()`` was added. https://github.com/cakephp/cakephp/pull/13216

Console
-------

* [x] Command classes can implement the ``defaultName()`` method to overwrite the
  conventions based CLI name.

More to be added? You can create list items via `* [ ] ...`.

---

Already included in [Shim](https://github.com/dereuromark/cakephp-shim) plugin and thus not necessarily needed (can easily be opt-in if desired):
- Controller referer default security improvement
- newEmptyEntity() method for Table (unless you want to move this also into 3.next)
- FormHelper BC behavior is included here as helper
- [this maybe](https://github.com/dereuromark/cakephp-shim/issues/56)

dereuromark:
Small question:
- Do we need to also backport newEmptyEntity() to core 3.9? (Currently, this is in Shim plugin as opt-in FC shim + deprecation and suffices as such IMO).

You can use :+1:  / :-1: on the comment for your feedback.

If not, I will tag a release here for Shim shortly documenting it.

othercorey:
[...]

PR #14105, opened 2019-12-31
Ticket 13493 backport http

Tiborhajos:
I tried my best to add the HTTP part of this ticket:

Note that I had to downgrade paragonie/random_compat to ^1.4, because paragonie/csp-builder ^1.4 requires it. I could not use a later version of the csp-builder as it requires php ^7.0.


markstory:
This doesn't look like a backport. `master` is currently for `4.x`.

Tiborhajos:
Sorry, my bad. I forgot to select the branch.

PR #14106, opened 2019-12-31
Ticket 13493 backport http

Tiborhajos:
I tried my best to add the HTTP part of this ticket: 

Note that I had to downgrade paragonie/random_compat to ^1.4, because paragonie/csp-builder ^1.4 requires it. I could not use a later version of the csp-builder as later versions require php ^7.0.


src/Http/Cookie/Cookie.php
```diff
@@ -118,6 +126,7 @@ class Cookie implements CookieInterface
      * @param string $domain Domain
      * @param bool $secure Is secure
      * @param bool $httpOnly HTTP Only
+     * @param null $sameSite Samesite
```
dereuromark: ...|

Tiborhajos: Sorry, I fail to see what you mean. Do you mean | string?

composer.json
```diff
@@ -36,12 +36,14 @@
         "psr/log": "^1.0.0",
         "psr/simple-cache": "^1.0.0",
         "zendframework/zend-diactoros": "^1.4.0",
-        "paragonie/random_compat": "^2.0|9.99.99"
+        "paragonie/random_compat": "^1.4|9.99.99",
+        "paragonie/csp-builder": "^1.4"
```
garas: It could be `"paragonie/random_compat": "^1.4|^2.0|9.99.99"`

- 1.4 when on PHP 5.6 and `paragonie/csp-builder` (unfortunately there is no tagged release with relaxed dependency https://github.com/paragonie/csp-builder/compare/v1.4.0...v1.x)
- 2.0 when  on PHP 5.6 without `paragonie/csp-builder`
- 9.99 when on PHP 7

And `paragonie/csp-builder` should be moved to `require-dev` and updated to `^1.4|^2.3` as it is optional.

src/Http/Cookie/Cookie.php
Tiborhajos: Cool, thank you

composer.json
Tiborhajos: Thanks for the feedback! I've struggled with how I should do this. Great suggestions!

src/Http/Cookie/Cookie.php
```diff
@@ -466,6 +485,45 @@ public function withExpired()
...
+    }
+
+    /**
+     * Check that value passed for SameSite is valid.
+     *
+     * @param string $sameSite SameSite value
+     * @return void
+     * @throws \InvalidArgumentException
+     */
+    protected static function validateSameSiteValue(string $sameSite)
```
garas:
```suggestion
    protected static function validateSameSiteValue($sameSite)
```

tests/TestCase/Http/Middleware/CspMiddlewareTest.php
```diff
@@ -0,0 +1,100 @@
...
+ * CakePHP(tm) : Rapid Development Framework (http://cakephp.org)
+ * Copyright (c) Cake Software Foundation, Inc. (http://cakefoundation.org)
+ *
+ * Licensed under The MIT License
+ * For full copyright and license information, please see the LICENSE.txt
+ * Redistributions of files must retain the above copyright notice.
+ *
+ * @copyright     Copyright (c) Cake Software Foundation, Inc. (http://cakefoundation.org)
+ * @link          http://cakephp.org CakePHP(tm) Project
+ * @since         4.0.0
```
garas:
```suggestion
 * @since         3.9.0
```

ADmad:
Combining multiple features in a single PR makes reviewing difficult. It's easy to miss out some detail this way. It would be better if you do a separate PR for each feature.

Tiborhajos:
You're right. 
At first I deemed this small enough for a single pull request. Yet in retrospect it would have been better to do one request per feature. I will definitely keep this in mind for future contributions. 

I've no idea how I should proceed from here. If you could help me out, that would be great.
Should I just close this request and start three separate ones? And is there a way to do this and preserve the feedback others have given?

ADmad:
You could start new branches for each feature and cherry pick the relevant commits from current branch. Preserving feedback is not necessary.

Tiborhajos:
That suggestion would not work as - and this makes me feel really stupid, like it should - I've done everything in one commit.

ADmad:
`ResponseEmitter.php` changes are related to SameSite support for cookies, not CSP middleware :).  New/updated tests for it in ResponseEmitterTest are missing.

Tiborhajos:
Thanks for pointing out the mistake in files and the missing tests :)
I'm on it.

Tiborhajos:
@ADmad I'll go with your suggestion to make 3 different PRs. 
I want to give you a heads-up and don't just close this one without a comment.

I'm stuck on the samesite cookie for now and I don't want it to hinder the other two functionalities. Besides, it's more concise and cleaner to do it this way, as you pointed out :)

PR #14107, opened 2020-01-01
Fix dirty associations that use formatResults

markstory:
This is a continuation of #14062 with a solution. When a joined association uses formatResults the parent entity should not be marked dirty if the result formatter modifieds the association entity.

While this may not be the optimal place to do this performance wise it does solve the problem and I'd like to revise the dirty/clean flagging that happens multiple times during association construction anyways.

~~Targeting 3.next as that is what the original pull request was aimed at. I can backport these changes to `master` and `3.x` or rebase just the fix and BelongsTo test if folks are ok with that.~~

Will port to 4.x once this has merged.

### DIFF
--- a/tests/TestCase/ORM/Behavior/TranslateBehaviorTest.php
+++ b/tests/TestCase/ORM/Behavior/TranslateBehaviorTest.php
@@ -859,36 +859,6 @@ class TranslateBehaviorTest extends TestCase
      *
      * @return void
      */
-    public function testGetAssociationNotDirtyBelongsToMany()
-    {
-        $table = $this->getTableLocator()->get('Articles');
-        $specialTags = $this->getTableLocator()->get('SpecialTags');
-        $specialTags->addBehavior('Translate', ['fields' => ['extra_info']]);
-
-        $table->belongsToMany('Tags', [
-            'through' => $specialTags,
-        ]);
-        $specialTags->setLocale('eng');
-
-        $entity = $table->get(2);
-        $this->assertNotEmpty($entity);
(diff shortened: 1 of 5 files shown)


tests/TestCase/ORM/Behavior/TranslateBehaviorTest.php
```diff
@@ -827,6 +827,113 @@ public function testFindSingleLocaleBelongsToMany()
         $this->assertEquals('Translated Info', $result->tags[0]->special_tags[0]->extra_info);
     }
 
+    /**
+     * Tests that parent entity isn't dirty when containing a translated association
+     *
+     * @return void
+     */
+    public function testGetAssociationNotDirtyBelongsTo()
```
othercorey: Can't we just add the dirty flag check to the above existing test cases?

markstory: We could. I'll reduce the amount of duplication in the tests. I think they were initially provided to help show the cases that were and were not working.

othercorey:
Looks ok to me. Would be nice to not have to iterate over all results, but better than being broken.

markstory:
@othercorey I agree the extra map isn't ideal, and I want to remove it in the future. But I think it will require a broader refactor to do so.

Zuluru:
Looks like when hydration is disabled on a query, this fails by trying to call `clean` on an array.

markstory:
> Looks like when hydration is disabled on a query, this fails by trying to call clean on an array.

Would be great if this could be a new issue. I'm 100% going to lose track of a comment on a closed pull request.

PR #14111, opened 2020-01-02
Moved schemaDialect() from driver dialect traits to drivers

othercorey:
This is part of merging the driver traits with the drivers. This moves the `schemaDialect()` function from the driver dialect traits. 

This copies the functions instead of implementing a shared function in `Driver` for a couple reason:

- Need to support existing user drivers that implement `Driver::schemaDialect()`.
- `SchemaDialect::__construct()` calls `Driver::connect()`.

   If we create the `SchemaDialect` instance in `Driver::__construct()`, it will try to connect immediately. That could happen before the db-specific constructor has finished initializing.

### DIFF
--- a/src/Database/Dialect/SqlserverDialectTrait.php
+++ b/src/Database/Dialect/SqlserverDialectTrait.php
@@ -22,8 +22,6 @@ use Cake\Database\Expression\UnaryExpression;
 use Cake\Database\ExpressionInterface;
 use Cake\Database\Query;
 use Cake\Database\QueryCompiler;
-use Cake\Database\Schema\SchemaDialect;
-use Cake\Database\Schema\SqlserverSchemaDialect;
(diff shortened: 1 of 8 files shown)


src/Database/Dialect/SqlserverDialectTrait.php
```diff
@@ -324,19 +322,6 @@ protected function _transformFunctionExpression(FunctionExpression $expression):
...
-     * Get the schema dialect.
-     *
-     * Used by Cake\Schema package to reflect schema and
-     * generate schema.
-     *
-     * @return \Cake\Database\Schema\SchemaDialect
-     */
-    public function schemaDialect(): SchemaDialect
-    {
-        return new SqlserverSchemaDialect($this);
```
othercorey: This was never cached.

PR #14113, opened 2020-01-02
4.1 - Merged driver dialect traits with drivers

othercorey:
This finishes the merge and removal of `Cake\Database\Dialect`. This doesn't change any implementation. 

- `SqlDialectTrait` moved to `Cake\Database\Driver` where it is used.
- Created alias for `Cake\Database\SqlDialectTrait` as it wasn't marked internal.
- Made drivers use `SqlDialectTrait` directly
- Moved `Driver` overrides directly into drivers
- Marked db-specific drivers as internal

We could make `SqlDialectTrait` methods part of the `Driver` base class, but I left them alone for this merge.  Don't know if we want to support protected methods in the `Driver` class or not.

Not sure if `SqlDialectTrait` needs an alias, but following standard procedure.


src/Database/Driver/Mysql.php
```diff
@@ -27,10 +26,12 @@
 
 /**
  * Class Mysql
+ *
+ * @internal
```
ADmad: Do we really need to mark the driver classes as internal?

ADmad:
Bye bye traits, you won't be missed.

src/Database/Driver/Mysql.php
othercorey: I leave it up to you.  Are they supportable classes?

We want people to extend `Driver` and implement `DriverInterface`. We want people to extend `SchemaDialect`, but do we want anything else?

ADmad: I have a feeling someone might want to extend say the Mysql driver and override something to suite their special needs.

othercorey: I can revert the @internal flag. Just means we can't change any of the protected implementation functions.

ADmad: As per our BC guidelines protected methods can be changed in new minor release.

src/Database/Driver/SqlDialectTrait.php
```diff
@@ -0,0 +1,312 @@
...
+     */
+    public function rollbackSavePointSQL($name): string
+    {
+        return 'ROLLBACK TO SAVEPOINT LEVEL' . $name;
+    }
+}
+
+// phpcs:disable
+// Add backwards compatible alias.
+class_alias('Cake\Database\Driver\SqlDialectTrait', 'Cake\Database\SqlDialectTrait');
```
markstory: Should this be in the old file location?

othercorey: This follows the pattern used in the other file moves. The alias is in the new file and a `class_exists` or `trait_exists` is in the old location which forces the new file to load.

garas: Traits can't be typehinted as method arguments, so they should safe to have only `class_alias` in old location without cluttering new file.

Issue #14109, opened 2020-01-01
Zend Framework Abandoned


jimbo2150:
This is a (multiple allowed):

* [ ] bug
* [x] enhancement
* [ ] feature-discussion (RFC)

The Zend framework has been abandoned and forked as the Laminas framework as of January 1, 2020 ([announced in April](https://framework.zend.com/blog/2019-04-17-announcing-laminas.html)) and is now operated by the Linux Foundation. A notification appears in composer when install or update is run on a CakePHP application.

dereuromark:
I assume it is about
https://github.com/cakephp/cakephp/blob/master/composer.json#L42-L43

        "zendframework/zend-diactoros": "^2.0",
        "zendframework/zend-httphandlerrunner": "^1.0"

Quite the hard cut to complete move the repo and rename namespace and files.
I wonder if we should move away from this library/framework.

ADmad:
There's no need to panic. While Zend framework might be abandoned it don't think it's github repo or packagist packages would go away anytime soon.

For 4.0 and 3.9 we could simply replace the relevant packages with the one's from Laminas (with class aliasing if required).

ADmad:
It seems laminas packages are still not available on packagist but i guess they are still in the processing of moving/setting up things.

jimbo2150:
@ADmad I agree, over the short-term the packages will remain so it isn't an emergency - just wanted to let everyone know and get the conversation started. I'm sure it will end up in packagist soon since they have documentation on adding it via composer.

@dereuromark I think it is premature to consider changing yet. I agree with @ADmad in that it should be switched over to Laminas over the medium-term so that any security updates will be pulled in (it is unlikely any updates will be made to the Zend repos from now on). In the long-term we should keep an eye on the Laminas repos to make sure it is being developed and utilized. If so, there may be no reason to switch. However, if it ends up being required, I looked at alternatives and there are a number of them and some are being used almost as much as Zend/Luminas is.

dereuromark:
PR open.

PR #14114, opened 2020-01-02
Replace Zend framework packages with Laminas.

ADmad:
The new laminas packages also include aliases for the old names within `Zend` namespace so app/plugin code using them shouldn't be affected.

I also opened a PR to fix few issues I found with the aliased functions https://github.com/laminas/laminas-diactoros/pull/30

Refs #14109

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -34,13 +34,13 @@
         "aura/intl": "^3.0.0",
         "cakephp/chronos": "^2.0",
         "composer/ca-bundle": "^1.2",
+        "laminas/laminas-diactoros": "^2.2.2",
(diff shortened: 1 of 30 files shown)


othercorey:
I wonder what prevented them from reusing the Zend name.

dereuromark:
We should wait then until they released a patch here. Quite the alias fail there.

ADmad:
> I wonder what prevented them from reusing the Zend name.

U.S. copyrights law I presume :slightly_smiling_face:

ADmad:
My patch to `laminas-diactoros` has been merged and new release done so this is good to go.

Issue #14115, opened 2020-01-02
LogTrait behaviour does not match documentation


BGehrels:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 4.0.1
* Platform and Target: irrelevant

### What you did
Within my AppController, i use the following code:
```
public function invokeAction(Closure $action, array $args): void
    {
         ConnectionManager::get('default')->transactional(
                    function () use ($action, $args) {
                        parent::invokeAction($action, $args);
                    }
                );
            }
        } catch (ExpectedException $e) {
            $this->log($e, LogLevel::INFO);
            [...]
       }
   }
```

### What happened

     Error: Argument 1 passed to Cake\Log\Engine\BaseLog::_format() must be of the type string, object given, called in /vagrant/oyaso/vendor/cakephp/cakephp/src/Log/Engine/FileLog.php on line 119

### What you expected to happen

The [documentation](https://book.cakephp.org/4/en/core-libraries/logging.html#logging-trait) says that
>  If $msg isn’t a string it will be converted with print_r before being logged.

so i would have expected that a stringified version of my exception is logged (as it has been in CakePHP 3.x)

markstory:
This behavior was changed in 4.0 in fa745835a79 as the old behavior was not aligned with the PSR4 specs. You'll need to do `$e->getMessage()` print_r() yourself before logging. I can update the documentation.

PR #14116, opened 2020-01-02
Fix docblocks and add typehints

ADmad:
Refs #14115

### DIFF
--- a/src/Log/Log.php
+++ b/src/Log/Log.php
@@ -341,7 +341,7 @@ class Log
      *
      * @param int|string $level The severity level of the message being written.
      *    The value must be an integer or string matching a known level.
-     * @param mixed $message Message content to log
+     * @param string $message Message content to log
      * @param string|array $context Additional data to be used for logging the message.
      *  The special `scope` key can be passed to be used for further filtering of the
      *  log engines to be used. If a string or a numerically index array is passed, it
@@ -350,7 +350,7 @@ class Log
      * @return bool Success
      * @throws \InvalidArgumentException If invalid level is passed.
      */
-    public static function write($level, $message, $context = []): bool
+    public static function write($level, string $message, $context = []): bool
     {
         static::_init();
         if (is_int($level) && in_array($level, static::$_levelMap, true)) {
--- a/src/Log/LogTrait.php
+++ b/src/Log/LogTrait.php
@@ -27,13 +27,13 @@ trait LogTrait
      * Convenience method to write a message to Log. See Log::write()
      * for more information on writing to logs.
      *
-     * @param string $msg Log message.
+     * @param string $message Log message.
      * @param int|string $level Error level.
      * @param string|array $context Additional log data relevant to this message.
      * @return bool Success of log write.
      */
-    public function log(string $msg, $level = LogLevel::ERROR, $context = []): bool
+    public function log(string $message, $level = LogLevel::ERROR, $context = []): bool
     {
-        return Log::write($level, $msg, $context);
+        return Log::write($level, $message, $context);
     }
 }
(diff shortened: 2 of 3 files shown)


src/Log/Log.php
```diff
@@ -350,7 +350,7 @@ public static function engine(string $name): ?LoggerInterface
      * @return bool Success
      * @throws \InvalidArgumentException If invalid level is passed.
      */
-    public static function write($level, $message, $context = []): bool
+    public static function write($level, string $message, $context = []): bool
```
dereuromark: Adding typehints on stable releases technically is a BC break, as any class extending Log would break with a fatal error.
We could either make this a minor (4.next) or say that this is OK for now since we are early in 4.0.

othercorey: I think we have to commit this as it's a bug fix. Unfortunately, the bug is the missing type hint.

dereuromark: That is not completely true, though. The fix is the docblock itself. The added typehint is an additional typehint that would break BC.
We can decide to do that, but it would not be necessary for the fix itself to be complete or working (and having 0 BC break).

othercorey: Up to you. With the existing code, we're giving the user broken api because of https://github.com/cakephp/cakephp/commit/fa745835a795ea97995b026bd7494802c24fcc37.

dereuromark: You misread. The code itself is already fixed without the typehint (as it would already allow string to work fine). There is no need for the typehint here other than making it "cleaner". Functionally it would work both ways, only that the typehint breaks the public API in a BC breaking way.

dereuromark: Concrete: You break the whole ecosystem that uses Log::write() as shown e.g. here: https://github.com/dereuromark/CakePHP-DatabaseLog/blob/master/src/Log/Engine/DatabaseLog.php#L49

othercorey: It is a big change come in post release.

src/Log/LogTrait.php
```diff
@@ -27,13 +27,13 @@ trait LogTrait
...
      * for more information on writing to logs.
      *
-     * @param mixed $msg Log message.
+     * @param string $message Log message.
      * @param int|string $level Error level.
      * @param string|array $context Additional log data relevant to this message.
      * @return bool Success of log write.
      */
-    public function log($msg, $level = LogLevel::ERROR, $context = []): bool
+    public function log(string $message, $level = LogLevel::ERROR, $context = []): bool
```
dereuromark: Will this break public API of e.g. https://github.com/dereuromark/CakePHP-DatabaseLog/blob/cake4/src/Log/Engine/DatabaseLog.php#L49 in a BC breaking way?
Then we dont need to add the typehint here either, the docblock fix should suffice.

src/Log/Log.php
markstory: I think changing this typehint will also cause conflicts with the PSR interfaces.

ADmad: > Adding typehints on stable releases technically is a BC break, as any class extending Log::write() would break with a fatal error.

There is practically zero chance of someone using extended version of `Cake\Log\Log`. Lack of typehint here will simply cause an error at engine level.

> I think changing this typehint will also cause conflicts with the PSR interfaces.

No it doesn't. As you see the test suite is passing. `Cake\Log\Log.php` and `Cake\Log\LogTrait` are our extra classes. Only the engine under `Cake\Log\Engine` use the PSR interface and their `log()` methods still have exact same signature as that of `LogInterface::log()`.

src/Log/LogTrait.php
ADmad: No it won't. Log engines use neither `Cake\Log\Log.php` or `Cake\Log\LogTrait`. You engine extends `Cake\Log\Engine\BaseLog`.

ADmad:
@markstory @dereuromark You are confusing `Cake\Log\Log` with `Cake\Log\Engine\BaseLog`. The log engines which extends `BaseLog` still retain the same signature as required by the PSR interface.

PR #14117, opened 2020-01-02
4.0: Fix config path to be normalized

dereuromark:
Note the double slash:

> PHP Warning:  Uncaught require(/.../sandbox.local/vendor/dereuromark/cakephp-feedback/config//routes.php): failed to open stream: No such file or directory

As our paths are all properly trailing slashed with one exception (ROOT), the extra slash should be removed.

### DIFF
--- a/src/Http/BaseApplication.php
+++ b/src/Http/BaseApplication.php
@@ -136,7 +136,7 @@ abstract class BaseApplication implements
      */
     public function bootstrap(): void
     {
-        require_once $this->configDir . '/bootstrap.php';
+        require_once $this->configDir . 'bootstrap.php';
     }
 
     /**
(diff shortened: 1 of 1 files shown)


ADmad:
The app passes the path without trailing slash. https://github.com/cakephp/app/blob/master/webroot/index.php#L37. So this change will break every app. 

The double slashes occur only in testsuite and the trailing slash should be removed when using `CONFIG` constant https://github.com/cakephp/cakephp/blob/b081f396f6436a94425c516e13b2220c09d13852/src/TestSuite/MiddlewareDispatcher.php#L85 and here https://github.com/cakephp/cakephp/blob/b081f396f6436a94425c516e13b2220c09d13852/src/TestSuite/ConsoleIntegrationTestTrait.php#L298

dereuromark:
That is literally the first path I see that is not normalized... besides ROOT
Maybe we can consolidate this for 4.1 then.

OK, fair enough.

dereuromark:
On 2nd thought:
What if we just normalized the incoming path directly?
Then it would be consistent again inside, right?

ADmad (approved):

dereuromark:
Thank you :)

PR #14121, opened 2020-01-03
Added 4.x backport for Client/Response->isSuccess() and tests.

Tiborhajos:
I tried my best to add the Client/Response->isSuccess() part of this ticket:


dereuromark:
We will also need some migration note docs probably in https://github.com/cakephp/docs/tree/3.next

Tiborhajos:
Thanks for reviewing! I'll go check out the docs.

PR #14122, opened 2020-01-03
Added 4.x backport for CspMiddleware and tests.

Tiborhajos:
I tried my best to add the CspMiddleware, as described in this ticket:

Note that I had to downgrade paragonie/random_compat to ^1.4, because paragonie/csp-builder ^1.4 requires it. I could not use a later version of the csp-builder as later versions require php ^7.0.


dereuromark:
Looks good

Note: Your github user/email is not set on your local machine, thus the commit here is anonymous (not linked).

markstory:
Thank you 👍

PR #14126, opened 2020-01-04
[4.x - cakephp/http] - add composer/ca-bundle

saeideng:
required  in https://github.com/cakephp/cakephp/blob/597c1e92cf1c66c17b113d8e8fa3e766cd8ebec1/src/Http/Client/Adapter/Curl.php#L25

### DIFF
--- a/src/Http/composer.json
+++ b/src/Http/composer.json
@@ -27,6 +27,7 @@
         "cakephp/core": "^4.0",
         "cakephp/event": "^4.0",
         "cakephp/utility": "^4.0",
+        "composer/ca-bundle": "^1.2",
         "psr/http-client": "^1.0",
         "psr/http-server-handler": "^1.0",
         "psr/http-server-middleware": "^1.0",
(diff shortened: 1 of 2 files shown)


src/Http/composer.json
```diff
@@ -27,6 +27,7 @@
         "cakephp/core": "^4.0",
         "cakephp/event": "^4.0",
         "cakephp/utility": "^4.0",
+        "composer/ca-bundle": "^1.2",
```
garas: Sync version constraint with main `composer.json`

saeideng: I can update main file to `^1.2`

garas: Both variants are OK.

dereuromark: Where is the prefer lowest run now on master? With travis this used to be a) visible and b) daily run using cronjobs. Now I cant find it anymore.

saeideng: removed when merging 3.x to 4.x

dereuromark: Thats bad :/

PR #14128, opened 2020-01-06
Improve how debug() handles cyclic references

markstory:
Right now if you dump an object that contains cyclic references somewhere in the requested depth the same objects are output multiple times. While this is annoying with small objects it makes reading debug() output for Table objects, or other complex objects really hard as the same content is output multiple times.

Now instead of duplicate content each object is assigned a context 'id'. When an object is encountered after the first time it is output only the short representation is emitted with the reference id. In a separate block of work I want to enhance the HTML output to include links to the original value from the cyclic reference. However, I felt this was a small/useful change on its own.

### DIFF
--- a/src/Error/Debugger.php
+++ b/src/Error/Debugger.php
@@ -228,14 +228,14 @@ class Debugger
      * Recursively formats and outputs the contents of the supplied variable.
      *
      * @param mixed $var The variable to dump.
-     * @param int $depth The depth to output to. Defaults to 3.
+     * @param int $maxDepth The depth to output to. Defaults to 3.
      * @return void
      * @see \Cake\Error\Debugger::exportVar()
      * @link https://book.cakephp.org/4/en/development/debugging.html#outputting-values
      */
-    public static function dump($var, int $depth = 3): void
+    public static function dump($var, int $maxDepth = 3): void
     {
-        pr(static::exportVar($var, $depth));
+        pr(static::exportVar($var, $maxDepth));
     }
 
     /**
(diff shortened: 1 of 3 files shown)


src/Error/Debugger.php
```diff
@@ -495,18 +495,19 @@ protected static function _highlight(string $str): string
      */
     public static function exportVar($var, int $depth = 3): string
```
othercorey: Can you rename this to `$maxDepth` to represent the value better?

src/Error/Debugger.php
```diff
@@ -522,15 +523,15 @@ protected static function _export($var, int $depth, int $indent): string
 
                 return "'" . $var . "'";
             case 'array':
-                return static::_array($var, $depth - 1, $indent + 1);
+                return static::_array($var, $context->withAddedDepthAndIndent());
```
othercorey: This function name `withAddedDepthAndIndent()` feels like you're increasing max depth, not jumping the context to the next depth level. Not sure there's a much better name for it though.

tests/TestCase/Error/DebuggerTest.php
```diff
@@ -310,14 +311,14 @@ public function testExportVar()
 
         $result = Debugger::exportVar($View);
         $expected = <<<TEXT
-object(Cake\View\View) {
-	Html => object(Cake\View\Helper\HtmlHelper) {}
-	Form => object(Cake\View\Helper\FormHelper) {}
+object(Cake\View\View) #0 {
```
othercorey: The `#` makes it feel like a count/depth marker not an id marker. Not sure if an `ID` prefix would look better or not.

src/Error/Debugger.php
othercorey: Maybe `atNextDepthAndIndent()` or `withNextDepthAndIndent()`. Just a thought.

markstory: I couldn't come up with a better name. I chose to use the `with` style naming as the DumpContext is a mostly immutable object. I first implemented it as mutable one and tracking depth/indent got really messy.

tests/TestCase/Error/DebuggerTest.php
markstory: I can do `id:0` if you think it would be clearer.

othercorey: Yea, that is what I was thinking.

PR #14131, opened 2020-01-06
Update config for the "setup-php" action to avoid warning.

ADmad:


dereuromark:
Can we also please have prefer-lowest back in travis?
This seems to be lost now with 3.x => 4.x migration. And this will soon become relevant in terms of minimum versions compatibility.

ADmad:
@dereuromark Done

PR #14145, opened 2020-01-09
4.0: Fix element cache key generation

### LINKED ISSUE #14144: Default FileEngine Cache keys of elements in subfolders of templates/element
This is a (multiple allowed):

* [x] bug
* [ ] enhancement

* CakePHP Version: 4.0.1

### What you did
* Set className of default Cache config to FileEngine::class (default from install)
* Create element file in app/templates/element/en_US/modal_signout.php
* Call the element in a view using <?= $this->element('en_US/modal_signout',[],['cache' => true]); ?>

### What happened
Cake\Cache\InvalidArgumentException thrown:
Cache keycake_element__en_US/modal_signout contains invalid characters. You cannot use /, \, <, >, ?, :, |, *, or " in cache keys.

Characters disallowed for file created would be replaced with a valid character, e.g. '-'

dereuromark:
Resolves https://github.com/cakephp/cakephp/issues/14144

Also fixed another issue that would occur with plugin namespace:

```php
->element('MyNamespace/MyPlugin.test_element', [], ['cache' => true])
```

This might be useful go get backported to 3.x.

### DIFF
--- a/src/View/View.php
+++ b/src/View/View.php
@@ -1583,15 +1583,16 @@ class View implements EventDispatcherInterface
 
         [$plugin, $name] = $this->pluginSplit($name);
 
-        $underscored = null;
+        $pluginKey = null;
         if ($plugin) {
-            $underscored = Inflector::underscore($plugin);
+            $pluginKey = str_replace('/', '_', Inflector::underscore($plugin));
         }
+        $elementKey = str_replace(['\\', '/'], '_', $name);
 
         $cache = $options['cache'];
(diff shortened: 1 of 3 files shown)


ADmad:
I wonder why the jobs for mysql and pgsql have started failing. They are failing on my PR #14146 too. They are passing on travis and appveyor so I presume it must be related to Github's VM.

dereuromark:
Jep, they seem unrelated and now globablly failing for some reason.

PR #14156, opened 2020-01-11
Split Form::schema() into getter and setter

chinpei215:
Although `\Cake\Form\Form::schema()` was not listed in #9978 , it seems like a combined getter/setter. Since `\Cake\ORM\Table::schema()` has been deprecated in 3.x and removed in 4.0, this method also should be deprecated. It will be removed in 5.0.

### DIFF
--- a/src/Form/Form.php
+++ b/src/Form/Form.php
@@ -136,25 +136,56 @@ class Form implements EventListenerInterface, EventDispatcherInterface, Validato
     }
 
     /**
+     * Set the schema for this form.
+     *
+     * @since 4.1.0
+     * @param \Cake\Form\Schema $schema The schema to set
+     * @return $this
+     */
+    public function setSchema(Schema $schema)
+    {
+        $this->_schema = $schema;
+
+        return $this;
+    }
+
+    /**
+     * Get the schema for this form.
+     *
+     * This method will call `_buildSchema()` when the schema
+     * is first built. This hook method lets you configure the
+     * schema or load a pre-defined one.
+     *
+     * @since 4.1.0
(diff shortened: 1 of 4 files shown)


src/Form/Form.php
```diff
@@ -135,26 +135,57 @@ public function implementedEvents(): array
...
      * @return \Cake\Form\Schema the schema instance.
      */
     public function schema(?Schema $schema = null): Schema
     {
-        if ($schema === null && empty($this->_schema)) {
-            $schema = $this->_buildSchema(new $this->_schemaClass());
-        }
-        if ($schema) {
-            $this->_schema = $schema;
+        if ($schema !== null) {
```
markstory: Can we add a deprecation warning for this method?

ADmad: Throwing explicit deprecation warning in 4.1 would be quite annoying for users IMO.

dereuromark: We should add those laters, like 4.3 or 4.4 maybe.

markstory: Ok. Lets try not to get too many deprecations stacked up before we add warnings. It took me a good month to add all the warnings in preparation for 4.x.

ADmad: @markstory We should also consider simply not adding explicit deprecation warnings in future. A lot of devs use IDEs these days which already warn them when using deprecated methods plus static analysers like psalm can also do the same.

markstory: I'm open to trying that. I've also started a rector ruleset for 4.1 in my fork. Once it has some more substance I'll open a pull request to upstream rector.

PR #14160, opened 2020-01-11
[WIP] [Experiment] Add Rector CI

TomasVotruba:
Rector CI has been very helpful to few smaller projects or application.
It can be big value to framworks, since there is prefixed rector.phar now.

**Don't mind this PR yet, it's just experiment of what CI can do and how useful it can be here.**

I'll ping you when all the edge-cases will be covered and configuration will be tailored to cakephp repository.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -14,7 +14,6 @@ trim_trailing_whitespace = true
 end_of_line = crlf
 
 [*.yml]
-indent_style = space
 indent_size = 2
 
 [Makefile]
--- a/.github/workflows/rector_ci.yml
+++ b/.github/workflows/rector_ci.yml
@@ -1,0 +1,19 @@
+name: Rector CI
+
+on:
+  pull_request:
+
+jobs:
+  rector_ci:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v1
+      - uses: shivammathur/setup-php@v1
+        with:
+          php-version: 7.2
+          extensions: mbstring, intl, apcu
+          coverage: none
+
+      - run: composer install
+
+      - run: vendor/bin/rector process --dry-run --config rector-ci.yaml --ansi --no-progress-bar
--- a/composer.json
+++ b/composer.json
@@ -53,7 +53,7 @@
         "mikey179/vfsstream": "^1.6",
         "paragonie/csp-builder": "^2.3",
         "phpunit/phpunit": "~8.5.0",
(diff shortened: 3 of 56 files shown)


dereuromark:
So, looking at the result, most changes are not good/useful:
```
-    abstract public function authenticate(ServerRequest $request, Response $response);
+    abstract public function authenticate(ServerRequest $request);
```
and
```
-    public function __construct(ServerRequest $request, Response $response, array $config = [])
+    public function __construct(ServerRequest $request, array $config = [])
```
etc

Just because some method of an interface implementation or some constructor doesnt make direct use of params, those method arguments are still valid and need to stay for BC reasons anyway.
It would be much more interesting to see actual dead code that can also safely be removed IMO.

TomasVotruba:
Please do wait, I need to check every case first :)

ADmad:
Can you please first example how is the new rector config you are adding going to help? 

We already use rector for our upgrade tool https://github.com/cakephp/upgrade/tree/4.x.

dereuromark:
@ADmad I think it is showcasing code mess detection and alike for active PRs.

Another interesting one:
```
-        $out .= $options['after'];

-        return $out;
+        return $out . $options['after'];
```

This kind of change is actually a code smell for me.
The variable exists already, as such completing it before returning is cleaner than returning a half-done concat one. Kind of reduces cognitive load as well to keep as it was instead of mixing the forms to save one line.
It is also easier to debug (and step through) if you have a var completed first, before returning it as a whole.

ADmad:
@dereuromark We had Scrutinizer CI earlier which suggested such noisy changes and we got rid of it as it was a PITA. If Rector CI is going to do similar then I already don't like it :slightly_smiling_face:.

dereuromark:
@ADmad It sounds like it can be configured, and maybe there are a few useful ones in there :)
Let's wait for his ping.

TomasVotruba:
> P.S. You can do experiments on your own fork too before you have something ready to show / for review slightly_smiling_face.

I would, but I don't want to replicate all the CI settings there already is.

dereuromark:
The clenaup looks neat and promising so far.
Still a few false positives probably.
```
-     * @var (\Psr\SimpleCache\CacheInterface&\Cake\Cache\CacheEngineInterface)|null
+     * @var (\Psr\SimpleCache\CacheInterface&\Cake\Cache\CacheEngineInterface|null)
```
This kind of changes the meaning :)

dereuromark:
Closing as stale, requires a new PR.

PR #14162, opened 2020-01-12
Fix schema type tests

othercorey:


markstory:
Nice work in tracking this down :clap::clap:

PR #14164, opened 2020-01-12
4.next - Begin decoupling variable dumping and formatting

markstory:
I would like to dramatically improve the output format of `debug()` for 4.1. While the plain text format we have today works it could be much better. For example in webserver contexts we could be outputting formatted HTML with collapsible sections and styled output. In console environments we could be using ANSI escape codes to give syntax highlighting.

In order to achieve these results first the 'dumping' and output stages need to be separated. The dumping stage will be shared across all output formats while each formatter will implement its own formatting logic. The dumping and formatting stages will communicate via a new AST.

This pull request adds the AST classes and modifies `Debugger` to generate the AST and use a backwards compatible `TextFormatter`. In future pull requests I plan to add:

* Additonal formatters for CLI and HTML outputs.
* Automatic environment detection.
* Methods to coerce a formatter if you really hate the detected ones, or want to use your own formatter.
* A formatter interface that userland code can implement.

Having a rich HTML dump formatter in core will allow us to remove the 'neat array' code from DebugKit and only have one dumping implementation.

I'm looking for feedback on the general approach and if the AST classes make sense. I'll finish off the doc blocks/psalm errors once the rough approach is agreed on.


othercorey:
I think I'd prefer to see `DumpContext` in a `Debug` namespace along with the nodes. Something to move it away from the public top-level `Error` namespace.

markstory:
> I think I'd prefer to see DumpContext in a Debug namespace along with the nodes. Something to move it away from the public top-level Error namespace.

How would that work in relation to Debugger? Would it stay in `Error` or also move to `Debug`?

othercorey:
> How would that work in relation to Debugger? Would it stay in `Error` or also move to `Debug`?

Sorry, I meant to say move it into `Cake\Error\Debug` so `Debug` inside of `Error` containing the bits used to handle debugging. I think `Debugger` would stay in `Cake\Error`.

PR #14165, opened 2020-01-13
Do not try to compare null passwords

cleptric:
If the hashed password from the DB is null, `DefaultPasswordHasher::check` throws a type error because it expects a string.

### DIFF
--- a/src/Auth/BaseAuthenticate.php
+++ b/src/Auth/BaseAuthenticate.php
@@ -123,7 +123,6 @@ abstract class BaseAuthenticate implements EventListenerInterface
 
         if ($password !== null) {
             $hasher = $this->passwordHasher();
-            $hashedPassword = $result->get($passwordField);
             if (!$hasher->check($password, $hashedPassword)) {
                 return false;
             }
--- a/src/Auth/BaseAuthenticate.php
+++ b/src/Auth/BaseAuthenticate.php
@@ -104,10 +104,25 @@ abstract class BaseAuthenticate implements EventListenerInterface
     protected function _findUser(string $username, ?string $password = null)
     {
         $result = $this->_query($username)->first();
+
+        if ($result === null) {
+            // Waste time hashing the password, to prevent
+            // timing side-channels. However, don't hash
+            // null passwords as authentication systems
+            // like digest auth don't use passwords
+            // and hashing *could* create a timing side-channel.
+            if ($password !== null) {
+                $hasher = $this->passwordHasher();
+                $hasher->hash($password);
+            }
+
+            return false;
+        }
+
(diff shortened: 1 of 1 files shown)


src/Auth/BaseAuthenticate.php
```diff
@@ -119,7 +121,6 @@ protected function _findUser(string $username, ?string $password = null)
             return false;
         }
 
-        $passwordField = $this->_config['fields']['password'];
         if ($password !== null) {
             $hasher = $this->passwordHasher();
             $hashedPassword = $result->get($passwordField);
```
othercorey: Getting hashed password twice?

src/Auth/BaseAuthenticate.php
```diff
@@ -120,9 +120,24 @@ protected function _findUser(string $username, ?string $password = null)
         }
 
         $passwordField = $this->_config['fields']['password'];
+        $hashedPassword = $result->get($passwordField);
```
cleptric: Not ideal, shall I create a new method to hash and return false?

src/Auth/BaseAuthenticate.php
```diff
@@ -120,9 +120,24 @@ protected function _findUser(string $username, ?string $password = null)
...
+            // null passwords as authentication systems
+            // like digest auth don't use passwords
+            // and hashing *could* create a timing side-channel.
+            if ($password !== null) {
+                $hasher = $this->passwordHasher();
+                $hasher->hash($password);
+            }
+
+            return false;
+        }
```
cleptric: Not ideal, shall I create a new method to hash and return false?

garas: First `if` could be modified to `if ($result === null || $result->get($passwordField) === null)`

cleptric: Is this in line with our guidelines? wasn't sure about making a call inside an if condition.

garas: But on password-less authentication systems, there is no password, so `$hashedPassword` is always `null`.

Currently, PR is completely wrong.

cleptric: Then I am worried why all our tests pass.

garas: Add before `if (!$hasher->check(...))`:

```php
if ($hashedPassword === null) {
    $hasher->hash($password);
    return false;
}
```

markstory (commented):
Any way we could have a test on this so it doesn't regress in the future?

cleptric:
@markstory yes, I'll refine this PR a bit further and add tests.

ADmad:
Replaced by #14168

PR #14168, opened 2020-01-14
Fix type error in authenticator with null passwords

garas:
Replaces #14165.

Fix type error in  `DefaultPasswordHasher::check()` when the saved hashed password is `null`.

Because `password_verify($password, $hash)` with empty `$hash` takes less time than with normal hash,  call `hash()` instead of `check()` to prevent timing side-channels.

Add tests to check if all cases (user not exists, user has no password, user has password) _always_ takes same time by calling either `hash()` or `check()`.

Also add tests to check if password-less authenticators takes same time  and _never_ uses password hasher to prevent issue from #12472. `DigestAuthenticateTest` used hard coded username, so it always found user and not triggered error. Also use dummy password hasher classname in case it tries to init password hasher.

### DIFF
--- a/src/Auth/BaseAuthenticate.php
+++ b/src/Auth/BaseAuthenticate.php
@@ -123,6 +123,16 @@ abstract class BaseAuthenticate implements EventListenerInterface
         if ($password !== null) {
             $hasher = $this->passwordHasher();
             $hashedPassword = $result->get($passwordField);
+
+            if ($hashedPassword === null || $hashedPassword === '') {
+                // Waste time hashing the password, to prevent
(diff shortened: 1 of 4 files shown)


dereuromark:
Sounds like we should also backport this to 3.x?

garas:
Yes. In backport to 3.x, `CallCounterPasswordHasher::check()` will need extended check `if ($hashedPassword === null || $hashedPassword === '')`.

ADmad (approved):

markstory:
Backport to 3.x done.

PR #14170, opened 2020-01-14
Correct & clarify length parameter of File::read()

ravage84:


dereuromark (approved):

ravage84:
Shall we fix this in 4.x, too? (even if the package is deprecated)

https://github.com/cakephp/cakephp/blob/master/src/Filesystem/File.php#L146

PR #14173, opened 2020-01-14
[3.x] Skip unchanged files when extracting translation strings

ravage84:
We heavily use translations in our applications. In some cases we have multiple translation domains (strored in several translation templates; POTs).

Every time we extract the translation strings, even though there were no changed made to any translation string, it leads to slightly changed POTs as they have an updated "POT-Creation-Date" line.
These POTs need to be reverted afterwards every time.

This improvement prevents "POT-Creation-Date" being the only line changing in the file every time one extracts the translation strings with unchanged translation strings.

### DIFF
--- a/src/Shell/Task/ExtractTask.php
+++ b/src/Shell/Task/ExtractTask.php
@@ -622,7 +622,7 @@ class ExtractTask extends Shell
             $filename = str_replace('/', '_', $domain) . '.pot';
             $File = new File($this->_output . $filename);
 
-            if ($this->_checkUnchanged($File, $headerLength, $output) === true) {
+            if ($File->exists() && $this->_checkUnchanged($File, $headerLength, $output) === true) {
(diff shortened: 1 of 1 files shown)


src/Shell/Task/ExtractTask.php
```diff
@@ -669,6 +678,26 @@ protected function _writeHeader()
...
+     * @param int $headerLength The length of the file header in bytes.
+     * @param string $newFileContent The content of the new file.
+     * @return bool Whether or not the old and new file are unchanged.
+     */
+    protected function _checkUnchanged(File $oldFile, $headerLength, $newFileContent)
+    {
+        $oldFileContent = $oldFile->read();
+
+        $oldChecksum = sha1(substr($oldFileContent, $headerLength));
+        $newChecksum = sha1(substr($newFileContent, $headerLength));
```
garas: It is really no advantage of hashing here. Content can be compared directly.

PR #14174, opened 2020-01-14
Fix static call of _stop on Console Error Handler

seanrabey:
Uncaught Error: Using $this when not in object context in /cakephp/lib/Cake/Console/ConsoleErrorHandler.php:63

Throwing an uncaught Exception in the Cake Shell would cause the error above. this patch fixes the error.

### DIFF
--- a/lib/Cake/Console/ConsoleErrorHandler.php
+++ b/lib/Cake/Console/ConsoleErrorHandler.php
@@ -60,7 +60,7 @@ class ConsoleErrorHandler {
 		));
 		$code = $exception->getCode();
 		$code = ($code && is_int($code)) ? $code : 1;
-		return $this->_stop($code);
+		return static::_stop($code);
 	}
 
 /**


saeideng:
you are using `handleException()` that is non-static 
`_stop()` is not static method

garas:
Current code is correct, stop should be called as `$this->_stop($code)`.

`ConsoleErrorHandler::handleException()` is _non-static_ method, and it is used as `[$instanceOfConsoleErrorHandler, 'handleException']` when registering as exception handler.

Similar `ErrorHandler::handleException()` is _static_ method, and it used as `'ErrorHandler::handleException'` when registering as exception handler.

Check how you have configured exception handler in your app.

seanrabey:
Hey all,
Thanks for taking a look at this. I guess I should have just opened an issue as I don't understand how to configure `consoleHandler` as it's not detailed in the documentation https://book.cakephp.org/2/en/development/exceptions.html#exception-configuration

Following the way the web handler is configured my exception handler configuration looks like this
```
Configure::write('Exception', array(
	'handler' => 'ErrorHandler::handleException',
	'renderer' => 'ExceptionRenderer',
	'consoleHandler' => 'ConsoleErrorHandler::handleException',
	'log' => true
));
```

Using a test Shell:
```
<?php

class TestShell extends AppShell
{
    public function main()
    {
        throw new Exception('Test');
    }
}
```
I get a second exception thrown
```
Welcome to CakePHP v2.10.19 Console
App : app
Path: /Downloads/cakephp-2.10.19/app/
Error: Test
#0 /Downloads/cakephp-2.10.19/lib/Cake/Console/Shell.php(462): TestShell->main()
#1 /Downloads/cakephp-2.10.19/lib/Cake/Console/ShellDispatcher.php(222): Shell->runCommand(NULL, Array)
#2 /Downloads/cakephp-2.10.19/lib/Cake/Console/ShellDispatcher.php(66): ShellDispatcher->dispatch()
#3 /Downloads/cakephp-2.10.19/app/Console/cake.php(47): ShellDispatcher::run(Array)
#4 {main}
PHP Fatal error:  Uncaught Error: Using $this when not in object context in /Downloads/cakephp-2.10.19/lib/Cake/Console/ConsoleErrorHandler.php:63
Stack trace:
#0 [internal function]: ConsoleErrorHandler::handleException(Object(Exception))
#1 {main}
  thrown in /Downloads/cakephp-2.10.19/lib/Cake/Console/ConsoleErrorHandler.php on line 63

Fatal error: Uncaught Error: Using $this when not in object context in /Downloads/cakephp-2.10.19/lib/Cake/Console/ConsoleErrorHandler.php:63
Stack trace:
#0 [internal function]: ConsoleErrorHandler::handleException(Object(Exception))
#1 {main}
  thrown in /Downloads/cakephp-2.10.19/lib/Cake/Console/ConsoleErrorHandler.php on line 63
Fatal Error Error: Uncaught Error: Using $this when not in object context in /Downloads/cakephp-2.10.19/lib/Cake/Console/ConsoleErrorHandler.php:63
Stack trace:
#0 [internal function]: ConsoleErrorHandler::handleException(Object(Exception))
#1 {main}
  thrown in [/Downloads/cakephp-2.10.19/lib/Cake/Console/ConsoleErrorHandler.php, line 63]
```

markstory:
> Following the way the web handler is configured my exception handler configuration looks like this

If you're not building your own console handler you don't need to configure anything as the `ShellDispatcher` will set a default handler in 2.x

seanrabey:
Thanks for the help and taking the time to look at this,
I am actually building my own handler for Rollbar, which then calls the built in handler. But to figure out the baseline, I tried to configure the built in consoleHandler but can't seem to make it 
 or my custom handler work. I've configured my custom handler like so
```php
Configure::write('Exception', array(
	'handler' => '\pump\spadmin\errors\CakeRollbarErrorHandler::handleException',
	'renderer' => 'ExceptionRenderer',
	'consoleHandler' => '\pump\spadmin\errors\CakeRollbarConsoleErrorHandler::handleException',
	'log' => true
));
```
How should consoleHandler be configured?
Here is my handler (it's actually a trait which both CakeRollbarErrorHandler and CakeRollbarConsoleErrorHandler inherit)
```php
/**
 * Collect the error data, send it to Rollbar, check if the Rollbar send worked, then call the baked in Cake Error handler
 * @param int $code Code of error
 * @param string $error_description Error description
 * @param string $file File on which error occurred
 * @param int $line Line that triggered the error
 * @param array $context Context
 * @return bool true if error was handled
 */
public static function handleError($code, $error_description, $file = null, $line = null, $context = null) // can't typehint as parent class is not typehitned
{
    $extra_data = [
        'code' => $code,
        'file' => $file,
        'line' => $line,
        'context' => $context
    ];

    $rollbar_log_level_name = self::getRollbarLogLevelNameFromPhpErrorCode($code);
    self::sendAndCheckRollbar($rollbar_log_level_name, $error_description, $extra_data);

    // Remove namespace from Class name
    $short_class_name = (new \ReflectionClass(static::class))->getShortName();
    if (strpos($short_class_name, 'Console') !== false) {
        // Cake ConsoleErrorHandler methods are NOT static ??
        return (new \ConsoleErrorHandler())->handleError($code, $error_description, $file, $line, $context);
    }

    return parent::handleError($code, $error_description, $file, $line, $context);
}
```

garas:
You can configure it as `[new YourErrorHandler(), 'handleException']` if it is non static method and `YourErrorHandler::handleException` if it is static.

You should check method if it is static or not.

Both `handler` and `consoleHandler` are configured same, and expects [callable](https://www.php.net/manual/en/language.types.callable.php). Difference is how classes are implemented.

seanrabey:
Great, thanks for all the help :)

PR #14181, opened 2020-01-16
Fix Notice Error: Trying to access array offset on value of type null

ravage84:
On PHP 7.4 when providing ``null`` for the ``$path``

### DIFF
--- a/src/View/Helper/UrlHelper.php
+++ b/src/View/Helper/UrlHelper.php
@@ -174,7 +174,7 @@ class UrlHelper extends Helper
         if (!array_key_exists('plugin', $options) || $options['plugin'] !== false) {
             list($plugin, $path) = $this->_View->pluginSplit($path, false);
         }
-        if (!empty($options['pathPrefix']) && $path[0] !== '/') {
+        if (!empty($options['pathPrefix']) && ($path === null || $path[0] !== '/')) {
             $path = $options['pathPrefix'] . $path;
         }
         if (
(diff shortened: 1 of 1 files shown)


ADmad:
A null path doesn't make sense, I would rather throw an exception.

dereuromark:
Or we cast :) Thats safer

ADmad:
It's not safer, it's masking an invalid argument.

dereuromark:
It might not be documented, but so far it worked. Breaking it for no reason is also not reasable IMO.

ravage84:
> A null path doesn't make sense, I would rather throw an exception.

@ADmad  while I agree that null is actually an invalid argument, throwing an exception would blow up the whole website. A bit extreme...

dereuromark:
@ravage84 Do you have an example where this would happen and make sense to have null and/or empty string supported?

ravage84:
> 
> 
> @ravage84 Do you have an example where this would happen and make sense to have null and/or empty string supported?

@dereuromark to be honest, I found this case in a data provider set in one of our test cases. Not sure, it was deliberately or one of our junior devs wanted to be thorough, as I thaught them to be. :laughing:

ADmad:
Ah this is for 3.x, well then ignore my comments, do whatever hack is necessary to avoid the error :)

ADmad (approved):

src/View/Helper/UrlHelper.php
```diff
@@ -174,7 +174,7 @@ public function assetUrl($path, array $options = [])
         if (!array_key_exists('plugin', $options) || $options['plugin'] !== false) {
             list($plugin, $path) = $this->_View->pluginSplit($path, false);
         }
-        if (!empty($options['pathPrefix']) && $path[0] !== '/') {
+        if (!empty($options['pathPrefix']) && ($path === null || $path[0] !== '/')) {
```
dereuromark: With an empty string you have the same issue...
Why not using `substr((string)...` here, that is safer.

PR #14182, opened 2020-01-16
4.1 - Moved class_alias for SqlDialectTrait to old location

othercorey:
Cleaning up alias from file moves. We don't need the special trait_exists > class_alias setup for traits as they cannot be function parameters and will always load.

@garas

### DIFF
--- a/src/Database/Driver/SqlDialectTrait.php
+++ b/src/Database/Driver/SqlDialectTrait.php
@@ -305,8 +305,3 @@ trait SqlDialectTrait
         return 'ROLLBACK TO SAVEPOINT LEVEL' . $name;
     }
 }
-
-// phpcs:disable
-// Add backwards compatible alias.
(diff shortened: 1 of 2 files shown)


dereuromark:
Bear in mind that these kind of class_alias() usages break deployment websites if composer is used with recommended optimizations -o and -a.
https://github.com/cakephp/docs/pull/6412/files

We should move away from those to a hack that works:
https://dev.to/greg0ire/how-to-deprecate-a-type-in-php-48cf

dereuromark:
A demo implementation: https://github.com/spryker/search/releases/tag/8.11.1

dereuromark:
@markstory It does. But that breaks composer in optimized mode. 
See the if(false) thingie: https://github.com/spryker/search/compare/8.11.0...8.11.1#diff-7b3ac20f465a3ac5ed6078283c6a0bd8R20

markstory:
Is that a frequently used mode/option? It seems like a lot of code to maintain.

ADmad:
I do use `-o` aka `--optimize` option.

PR #14185, opened 2020-01-17
Quote SELECT identifiers only if autoquoting is enabled.

ADmad:


ADmad:
Postgres is playing spoilt sport :cry:. I don't have it locally to check if there's some easy fix that can be done to avoid the failures.

ADmad:
So apparently Posgres auto converts all unquoted identifier to lower cased which is what's causing the failures.

We could make a new `PostgresCompiler` (similar to existing `SqliteCompiler`) and retain the quoting only for Postgres.

ADmad:
If creating a `PostgresCompiler` is acceptable I'll make required changes and reopen this against `4.next`. Even though query compilers are marked internal I don't want to much around with them in `4.0`

markstory:
@ADmad I think having a postgres specific compiler is fair. We have compilers for Sqlite and Sqlserver to work around their specific requirements.

ADmad:
@markstory I eventually decided to add `PostgresCompiler` in this PR itself for `4.0`. I think it's best to make this quoting related change early in 4.x life cycle. Waiting till 4.1 increases the chance of causing problems to users and quite a few would have already adopted Cake 4.

src/Database/QueryCompiler.php
```diff
@@ -85,6 +85,13 @@ class QueryCompiler
      */
     protected $_orderedUnion = true;
 
+    /**
+     * Indicate whether identifiers of SELECT clause need to be always quoted.
+     *
+     * @var bool
+     */
+    protected $_quotedSelectIdentifiers = false;
```
othercorey: This is specifically for quoting aliases right, not column identifiers.

ADmad: Yes, should I rename the property to better reflect that?

othercorey: I think so.

src/Database/QueryCompiler.php
```diff
@@ -85,6 +85,13 @@ class QueryCompiler
      */
     protected $_orderedUnion = true;
 
+    /**
+     * Indicate whether aliases in SELECT clause need to be always quoted.
+     *
+     * @var bool
+     */
+    protected $_quotedSelectAliases = false;
```
markstory: I'm half tempted to not have protected property overrides and instead duplicate `_buildSelectPart` into the Postgres compiler so that both the postgres and non-postgres versions have no additional branches. Our test suite is thorough enough that we should be able to keep changes aligned between the two implementations.

ADmad: I too had contemplated the same thing :)

I will update accordingly.

ADmad: @markstory Duplicating `_buildSelectPart()` will only avoid a `if` check in `PostgresCompiler`. The `if` check will still be required in `QueryCompiler` if auto quoting is enabled.

markstory: Well that doesn't seem worth it.

Issue #14080, opened 2019-12-27
Add sql window function support


othercorey:
Tracking issue for implementing window function support.

Goals:

- Any aggregate function can be extended to a window function
  - List of supported function is driver specific, up to user to pick right one
- Windows can be declared separate from functions and shared
- Support for all standard/common window definitions
  - Window Name
  - OVER()
  - PARTITION
  - ORDER BY
  - FRAME_START + FRAME_END
- Pre-defined wrappers for common window functions like `lag()` and `lead()`

inoas:
Ref https://github.com/cakephp/cakephp/issues/10670

ndm2:
I'm a little late to the party... but, @othercorey do you have any plans for adding support for [**named windows**](https://dev.mysql.com/doc/refman/8.0/en/window-functions-named-windows.html)? Looking at the current state, it would probably not possible to implement without breaking changes.

Also would you share my concern that the fact that all window functions that the builder can create are implemented as `AggregateExpression`, could be a little misleading, and perhaps awkward to use, given that not all window functions actually are aggregate functions, like `RANK()`, `ROW_NUMBER()`, `LAG()`, `LEAD()`, etc...?

othercorey:
Can you open a new issue so we aren't discussing in one that isn't listed?

ndm2:
Sure thing, I've create a new issue.

PR #14186, opened 2020-01-17
4.1 - Added WindowExpression with support for empty OVER () clause.

othercorey:

I didn't implement `WindowInterface` in this PR for easier review. It will be implemented by both 
`AggregateExpression` and `WindowExpression`.

### DIFF
--- a/src/Database/Expression/AggregateExpression.php
+++ b/src/Database/Expression/AggregateExpression.php
@@ -16,6 +16,9 @@ declare(strict_types=1);
  */
 namespace Cake\Database\Expression;
 
+use Cake\Database\ValueBinder;
+use Closure;
+
 /**
  * This represents a SQL aggregate function expression in a SQL statement.
  * Calls can be constructed by passing the name of the function and a list of params.
@@ -24,4 +27,64 @@ namespace Cake\Database\Expression;
  */
 class AggregateExpression extends FunctionExpression
 {
+    /**
+     * @var \Cake\Database\Expression\WindowExpression
+     */
+    protected $window;
+
+    /**
+     * Adds an empty `OVER()` window expression.
+     *
+     * If the window expression for this aggregate is already
+     * initialized, this does nothing.
+     *
+     * @return $this
+     */
+    public function over()
+    {
+        if ($this->window === null) {
+            $this->window = new WindowExpression();
+        }
+
+        return $this;
+    }
+
+    /**
+     * @inheritDoc
+     */
+    public function sql(ValueBinder $generator): string
+    {
+        $sql = parent::sql($generator);
+        if ($this->window !== null) {
+            $sql .= ' ' . $this->window->sql($generator);
+        }
+
+        return $sql;
+    }
+
+    /**
+     * @inheritDoc
+     */
+    public function traverse(Closure $visitor)
+    {
+        parent::traverse($visitor);
+        if ($this->window !== null) {
+            $this->window->traverse($visitor);
+        }
+
+        return $this;
+    }
+
+    /**
+     * @inheritDoc
+     */
+    public function count(): int
+    {
+        $count = parent::count();
+        if ($this->window !== null) {
+            $count = $count + 1;
+        }
+
+        return $count;
+    }
 }
--- a/src/Database/Expression/WindowInterface.php
+++ b/src/Database/Expression/WindowInterface.php
@@ -1,0 +1,110 @@
+<?php
+declare(strict_types=1);
+
+/**
+ * CakePHP(tm) : Rapid Development Framework (https://cakephp.org)
+ * Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
+ *
+ * Licensed under The MIT License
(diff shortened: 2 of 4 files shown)


src/Database/Expression/AggregateExpression.php
```diff
@@ -24,4 +27,64 @@
...
+
+    /**
+     * Adds an empty `OVER()` window expression.
+     *
+     * If the window expression for this aggregate is already
+     * initialized, this does nothing.
+     *
+     * @return $this
+     */
+    public function over()
```
markstory: Once someone has created a window how do they get access to it?

src/Database/Expression/WindowInterface.php
```diff
@@ -0,0 +1,110 @@
...
+     * overriden by passing an array with the order as the key. The SQL standard
+     * for ordering must be followed.
+     *
+     * ```
+     * // this is the same as '1 FOLLOWING`
+     * $window->range(['following' => 1]);
+     * ```
+     *
+     * The SQL keywords `UNBOUNDED` and `CURRENT ROW` can be used directly or
+     * easier to read substitutes `null` and `0` instead.
```
markstory:
```suggestion
     * easier to read substitutes for `null` and `0` instead.
```

src/Database/Expression/WindowInterface.php
```diff
@@ -0,0 +1,110 @@
...
+     * ```
+     * // this is the same as 'CURRENT ROW'
+     * $window->range(0);
+     *
+     * // this is the same as 'UNBOUNDED PRECEDING'
+     * $window->range(null)
+     * ```
+     *
+     * @param array|int|string|null $start Frame start
+     * @param array|int|string|null $end Frame end
```
markstory: Do we need an array in the second parameter? We could eliminate the need for string types with a constant. A constant would give better autocompletion than a string would as well.

src/Database/Expression/WindowInterface.php
```diff
@@ -0,0 +1,110 @@
...
+     * specified per window.
+     *
+     * See `range()` for details on `$start` and `$end` format.
+     *
+     * @param array|int|string|null $start Frame start
+     * @param array|int|string|null $end Frame end
+     *  If not passed in, only frame start SQL will be generated.
+     * @return $this
+     */
+    public function rows($start, $end = 0);
```
markstory: Could use the constants here too.

src/Database/Expression/WindowInterface.php
```diff
@@ -0,0 +1,110 @@
...
+    public function range($start, $end = 0);
+
+    /**
+     * Adds a rows frame clause to the window. Only one frame clause can be
+     * specified per window.
+     *
+     * See `range()` for details on `$start` and `$end` format.
+     *
+     * @param array|int|string|null $start Frame start
+     * @param array|int|string|null $end Frame end
```
markstory: Does end need array?

src/Database/Expression/WindowInterface.php
```diff
@@ -0,0 +1,110 @@
...
+    public function rows($start, $end = 0);
+
+    /**
+     * Adds a groups frame clause to the window. Only one frame clause can be
+     * specified per window.
+     *
+     * See `range()` for details on `$start` and `$end` format.
+     *
+     * @param array|int|string|null $start Frame start
+     * @param array|int|string|null $end Frame end
```
markstory: Does end need array?

src/Database/Expression/WindowInterface.php
```diff
@@ -0,0 +1,110 @@
...
+    public function groups($start, $end = 0);
+
+    /**
+     * Adds a frame exclusion to the window.
+     *
+     * Known exclusion keywords are:
+     *  - CURRENT ROW
+     *  - GROUP
+     *  - TIES
+     *  - NO OTHERS
```
markstory: Should we have constants for these?

src/Database/Expression/AggregateExpression.php
othercorey: `AggregateExpression` will implement `WindowInterface` which passes the calls to `$window` and returns `$this` for chaining.

src/Database/Expression/WindowInterface.php
othercorey: Both end can be both 'preceding' and 'following'. We assume they want start = preceding and end = follow, but need to allow both.

src/Database/Expression/WindowInterface.php
```diff
@@ -0,0 +1,148 @@
...
+{
+    /**
+     * 'CURRENT ROW' frame start, end or exclusion
+     *
+     * @var int
+     */
+    public const CURRENT_ROW = 0;
+
+    /**
+     * 'UNBOUNDED PRECEDING' and '(offset) PRECEDING' frame start or end
```
othercorey: These can only work for the array format otherwise how can we tell the difference between offset integers and constants.  Users will have to specify `null` to mean unbounded.

markstory (approved):
Looking good to me. You'll learn more about the interface as you implement it :smile:

PR #14189, opened 2020-01-18
Fix CS errors and issues reported by static analyzers

ADmad:
The phpcs ignores can be removed once https://github.com/slevomat/coding-standard/issues/851 is addressed.

### DIFF
--- a/src/Cache/CacheRegistry.php
+++ b/src/Cache/CacheRegistry.php
@@ -36,8 +36,7 @@ class CacheRegistry extends ObjectRegistry
      * Part of the template method for Cake\Core\ObjectRegistry::load()
      *
      * @param string $class Partial classname to resolve.
-     * @return string|null Either the correct classname or null.
-     * @psalm-return class-string
+     * @return class-string|null Either the correct classname or null.
      */
     protected function _resolveClassName(string $class): ?string
     {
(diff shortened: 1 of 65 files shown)


src/Console/HelperRegistry.php
```diff
@@ -55,8 +55,7 @@ public function setIo(ConsoleIo $io): void
      * Part of the template method for Cake\Core\ObjectRegistry::load()
      *
      * @param string $class Partial classname to resolve.
-     * @return string|null Either the correct class name or null.
-     * @psalm-return class-string
+     * @return class-string|null Either the correct class name or null.
```
othercorey: TypeResolver doesn't seem to handle `class-string` yet. Maybe I can add it in a PR.

src/Cache/CacheRegistry.php
```diff
@@ -36,8 +36,7 @@ class CacheRegistry extends ObjectRegistry
      * Part of the template method for Cake\Core\ObjectRegistry::load()
      *
      * @param string $class Partial classname to resolve.
-     * @return string|null Either the correct classname or null.
-     * @psalm-return class-string
+     * @return class-string|null Either the correct classname or null.
```
dereuromark: We shouldnt do this yet.

othercorey:
Opened PR for class-string support in TypeResolver so we don't have to rely on psalm-specific annotations.

https://github.com/phpDocumentor/TypeResolver/pull/90

ADmad:
Both phpstan and psalm (and even phan) support `class-string`, so others should catch up.

dereuromark:
Maybe not IDEs and other tools? For them it is still string.

PR #14192, opened 2020-01-19
4.1 - Implemented partition() and order() for WindowExpression

othercorey:

Will add `WindowInterface` to `AggregateExpression` with wrappers when `WindowExpression` is done.

### DIFF
--- a/src/Database/Expression/WindowExpression.php
+++ b/src/Database/Expression/WindowExpression.php
@@ -141,11 +141,12 @@ class WindowExpression implements ExpressionInterface, WindowInterface
     {
         $partitionSql = '';
         if ($this->partitions) {
-            $partitions = array_map(function ($partition) use ($generator) {
-                return $partition->sql($generator);
-            }, $this->partitions);
+            $expressions = [];
+            foreach ($this->partitions as $partition) {
+                $expressions[] = $partition->sql($generator);
+            }
 
-            $partitionSql = 'PARTITION BY ' . implode(', ', $partitions);
+            $partitionSql = 'PARTITION BY ' . implode(', ', $expressions);
         }
 
         return sprintf(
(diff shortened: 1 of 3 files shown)


src/Database/Expression/WindowExpression.php
```diff
@@ -19,25 +19,161 @@
...
+        foreach ($partitions as &$partition) {
+            if (is_string($partition)) {
+                $partition = new IdentifierExpression($partition);
+            }
+
+            if (
+                $partition instanceof WindowExpression ||
+                ($partition instanceof AggregateExpression && $partition->getWindow())
+            ) {
+                throw new InvalidArgumentException('Window partition cannot contain other window functions.');
```
othercorey: This should happen in sql() since the user could add the window after. Not sure if it is important to validate this or make the user responsible for valid sql.

src/Database/Expression/WindowExpression.php
```diff
@@ -23,21 +23,154 @@
...
      * @inheritDoc
      */
     public function sql(ValueBinder $generator): string
     {
-        return 'OVER ()';
+        $partitionSql = '';
+        if ($this->partitions) {
+            $partitions = array_map(function ($partition) use ($generator) {
+                return $partition->sql($generator);
+            }, $this->partitions);
```
markstory: Using a foreach would yield better performance, but it likely doesn't matter here.

src/Database/Expression/WindowInterface.php
```diff
@@ -80,69 +48,77 @@ public function partition($partitions);
...
      *
-     * @param array|int|null $start Frame start
-     * @param array|int|null $end Frame end
+     * @param int|null $start Frame start
+     * @param int|null $end Frame end
      *  If not passed in, only frame start SQL will be generated.
      * @return $this
      */
-    public function rows($start, $end = 0);
+    public function groups(?int $start, ?int $end = 0);
```
markstory: These changes are great :clap: The simpler types are much better, and the additional methods to handle the special cases will be simper to read down the road.

src/Database/Expression/WindowExpression.php
othercorey: Sure, can do.

Issue #14143, opened 2020-01-08
IntegrationTestTrait - get() & post() do not need to throw an exception


tfont:
`@throws \PHPUnit\Exception`

Since:
```
$this->get('path/dir');
$this->assertResponseCode(200);
```

will be catched with a negative test as:
```
$this->get('path/dir');
$this->assertResponseCode(404);
```

Note; this would be considered an 'enhancement' but should be discussed.

dereuromark:
Yes, those can be removed IMO. You can make a PR here.

tfont:
PR has been made and can be found:
https://github.com/cakephp/cakephp/pull/14193

markstory:
Thanks @tfont

PR #14193, opened 2020-01-20
removal of Throwables through _sendRequest from get(), post(), patch(), put(), delete, head(), & options()

travisfont:
IntegrationTestTrait removed all throwables from send request methods:

- get()
- post()
- patch()
- put()
- delet()
- head()
- options()

See discussion/issue at:
https://github.com/cakephp/cakephp/issues/14143

### DIFF
--- a/src/TestSuite/IntegrationTestTrait.php
+++ b/src/TestSuite/IntegrationTestTrait.php
@@ -17,7 +17,6 @@ namespace Cake\TestSuite;
 
 use Cake\Controller\Controller;
 use Cake\Core\Configure;
-use Cake\Database\Exception as DatabaseException;
 use Cake\Error\ExceptionRenderer;
 use Cake\Event\EventInterface;
 use Cake\Form\FormProtector;
@@ -60,7 +59,6 @@ use Cake\Utility\Text;
 use Exception;
 use Laminas\Diactoros\Uri;
 use LogicException;
-use PHPUnit\Framework\Error\Error as PhpUnitError;
 use Throwable;
 
 /**
@@ -501,9 +499,9 @@ trait IntegrationTestTrait
             }
             $this->_response = $response;
             $this->_exception = null;
-        } catch (PhpUnitError $e) {
+        } catch (\PHPUnit\Framework\Error\Error $e) {
             $this->_exception = $e;
-        } catch (DatabaseException $e) {
+        } catch (\Cake\Database\Exception $e) {
             $this->_exception = $e;
         } catch (LogicException $e) {
             $this->_exception = $e;
@@ -559,7 +557,7 @@ trait IntegrationTestTrait
      *
      * This method will attempt to use the configured exception renderer.
      * If that class does not exist, the built-in renderer will be used.
-     * @param \Throwable|\LogicException|DatabaseException|PhpUnitError|null $exception Exception to handle.
+     * @param \Throwable|\LogicException|\Cake\Database\Exception|\PHPUnit\Framework\Error\Error|null $exception Exception to handle.
      * @return void
      */
     protected function _handleError($exception): void
--- a/src/TestSuite/IntegrationTestTrait.php
+++ b/src/TestSuite/IntegrationTestTrait.php
@@ -107,7 +107,7 @@ trait IntegrationTestTrait
     /**
      * The exception being thrown if the case.
      *
-     * @var \Throwable|null
+     * @var Throwable|null
      */
     protected $_exception;
 
@@ -567,7 +567,7 @@ trait IntegrationTestTrait
      * This method will attempt to use the configured exception renderer.
      * If that class does not exist, the built-in renderer will be used.
      *
-     * @param \Throwable|\LogicException|\LogicException|DatabaseException|PhpUnitError $exception Exception to handle.
+     * @param Throwable|LogicException|DatabaseException|PhpUnitError $exception Exception to handle.
      * @return void
      */
     protected function _handleError($exception): void
--- a/src/TestSuite/IntegrationTestTrait.php
+++ b/src/TestSuite/IntegrationTestTrait.php
@@ -377,7 +377,6 @@ trait IntegrationTestTrait
      * response.
      *
      * @param string|array $url The URL to request.
-     *
      * @return void
      */
     public function get($url): void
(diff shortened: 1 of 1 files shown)


src/TestSuite/IntegrationTestTrait.php
```diff
@@ -509,16 +509,17 @@ protected function _sendRequest($url, $method, $data = []): void
...
         } catch (PhpUnitError $e) {
-            throw $e;
+            $this->_exception = $e;
         } catch (DatabaseException $e) {
-            throw $e;
+            $this->_exception = $e;
         } catch (LogicException $e) {
-            throw $e;
+            $this->_exception = $e;
         } catch (Throwable $e) {
```
garas: `Throwable` covers all exceptions

src/TestSuite/IntegrationTestTrait.php
```diff
@@ -566,10 +567,10 @@ public function controllerSpy(EventInterface $event, ?Controller $controller = n
      * This method will attempt to use the configured exception renderer.
      * If that class does not exist, the built-in renderer will be used.
      *
-     * @param \Throwable $exception Exception to handle.
+     * @param \Throwable|\LogicException|\LogicException|DatabaseException|PhpUnitError $exception Exception to handle.
```
markstory: You have LogicException in this annotation twice.

othercorey:
@tfont are you going to follow up on this?

tfont:
> @tfont are you going to follow up on this?

@othercorey Yes, I can update this.

src/TestSuite/IntegrationTestTrait.php
```diff
@@ -377,8 +377,8 @@ public function cookieEncrypted(string $name, $value, $encrypt = 'aes', $key = n
      * response.
      *
      * @param string|array $url The URL to request.
+     *
      * @return void
-     * @throws \PHPUnit\Framework\Error\Error|\Throwable
```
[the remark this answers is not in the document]
dereuromark: Like if any business logic or in the controller throws exceptions, and we disable the error catching then this would still be true, right?

tfont:
@othercorey @markstory 
feel free to review and leave feedback!

src/TestSuite/IntegrationTestTrait.php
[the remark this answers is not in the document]
dereuromark: These are still not answered here. I still think those should be kept as those can through exceptions if error handler middleware is disabled.

src/TestSuite/IntegrationTestTrait.php
```diff
@@ -566,10 +559,10 @@ public function controllerSpy(EventInterface $event, ?Controller $controller = n
      * This method will attempt to use the configured exception renderer.
      * If that class does not exist, the built-in renderer will be used.
      *
-     * @param \Throwable $exception Exception to handle.
+     * @param \Throwable|\LogicException|\Cake\Database\Exception|\PHPUnit\Framework\Error\Error $exception Exception to handle.
```
dereuromark, line 562: Aren't `\LogicException|\Cake\Database\Exception|\PHPUnit\Framework\Error\Error` all extending Throwable?

tfont: Yes. Is it more ideal to remove them all and use the inheritable Throwable instead?

dereuromark: I don't see the point in outlining them all here, if they are subclasses of it.

src/TestSuite/IntegrationTestTrait.php
```diff
@@ -508,16 +498,19 @@ protected function _sendRequest($url, $method, $data = []): void
...
-            throw $e;
+            $this->_exception = $e;
         } catch (Throwable $e) {
             $this->_exception = $e;
-            // Simulate the global exception handler being invoked.
-            $this->_handleError($e);
+        }
+
+        // Simulate the global exception handler being invoked.
+        if ($this->_exception instanceof Throwable) {
```
dereuromark, line 512: Isnt this a behavior change?
They all are instance of this class afaik. Before only in the last catch block this was invoked, now in every case, right?

tfont: Yes, previously it was:

```php
try {
   // ...
   $this->_response = $response;
   } catch (PhpUnitError $e) {
      throw $e;
   } catch (DatabaseException $e) {
      throw $e;
   } catch (LogicException $e) {
      throw $e;
   } catch (Throwable $e) {
      $this->_exception = $e;
      // Simulate the global exception handler being invoked.
      $this->_handleError($e);
}
```

I've updated all throwables to be added to the exception property and then passed to the handleError() to handle the exception. Thoughts?

src/TestSuite/IntegrationTestTrait.php
```diff
@@ -508,16 +498,19 @@ protected function _sendRequest($url, $method, $data = []): void
                 $this->_requestSession->write('Flash', $this->_flashMessages);
             }
             $this->_response = $response;
-        } catch (PhpUnitError $e) {
-            throw $e;
-        } catch (DatabaseException $e) {
-            throw $e;
+        } catch (\PHPUnit\Framework\Error\Error $e) {
+            $this->_exception = $e;
```
garas, line 502: At least PHPUnit exceptions should be rethrown. Normal applications shouldn't use PHPUnit classes.

However it could be widened to `\PHPUnit\Framework\Exception` class or `\PHPUnit\Exception` interface.

All PHPUnit asserts should immediately abort test, but not rendered with `ExceptionRenderer`. Otherwise it makes testing hard.

src/TestSuite/IntegrationTestTrait.php
dereuromark: I think the reason here is to have different behaviors.

garas: Previously it was different behaviors, but now all `catch` blocks have same code, so all blocks could handled in single `catch (Throwable $e)`.

dereuromark:
Can you explain again please what the reason for this is?
The original issue was about removing some docblock lines, which given that they shouldn't occur in normal testing, made sense

    @throws \PHPUnit\Exception

etc
But the scope of the changes now here go way beyond, and in a way that is not acceptable as is.

It would be good to have the main goal again in front of us, to be able to find a solution.

markstory:
Closing in favour of #14549 which achieves a similar result with a simpler approach.

PR #14194, opened 2020-01-20
Update phpstan's config

ADmad:


src/View/Helper.php
```diff
@@ -106,7 +106,7 @@ public function __call(string $method, array $params)
      * Lazy loads helpers.
      *
      * @param string $name Name of the property being accessed.
-     * @return \Cake\View\Helper|null Helper instance if helper with provided name exists
+     * @return \Cake\View\Helper|void Helper instance if helper with provided name exists
```
dereuromark: null|void
Both then?

PR #14198, opened 2020-01-21
4.1 - Setup mysql8 for github actions

othercorey:
This switches github actions to mysql 8 via docker container.

- Ignored integer display widths except for `TINYINT(1)`
- Properly sorted foreign keys by declaration order when building schema
- Rename groups fixture as groups is now reserved keyword

### DIFF
--- a/tests/TestCase/Database/Schema/MysqlSchemaTest.php
+++ b/tests/TestCase/Database/Schema/MysqlSchemaTest.php
@@ -85,36 +85,36 @@ class MysqlSchemaTest extends TestCase
                 ['type' => 'boolean', 'length' => null],
             ],
             [
-                'TINYINT(2)',
-                ['type' => 'tinyinteger', 'length' => 2, 'unsigned' => false],
+                'TINYINT(1) UNSIGNED',
+                ['type' => 'boolean', 'length' => null],
             ],
             [
                 'TINYINT(3)',
-                ['type' => 'tinyinteger', 'length' => 3, 'unsigned' => false],
+                ['type' => 'tinyinteger', 'length' => null, 'unsigned' => false],
             ],
             [
                 'TINYINT(3) UNSIGNED',
-                ['type' => 'tinyinteger', 'length' => 3, 'unsigned' => true],
+                ['type' => 'tinyinteger', 'length' => null, 'unsigned' => true],
             ],
             [
                 'SMALLINT(4)',
-                ['type' => 'smallinteger', 'length' => 4, 'unsigned' => false],
+                ['type' => 'smallinteger', 'length' => null, 'unsigned' => false],
             ],
             [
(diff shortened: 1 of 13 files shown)


src/Database/Schema/MysqlSchema.php
```diff
@@ -112,22 +112,22 @@ protected function _convertColumn(string $column): array
             return ['type' => $typeName, 'length' => null, 'precision' => $length];
         }
 
-        if (($col === 'tinyint' && $length === 1) || $col === 'boolean') {
+        if ($col === 'tinyint' || $col === 'boolean') {
```
ndm2: Are you just testing stuff? I'd not be overly happy when all my `TINYINT` columns were suddenly treated as flags after a bugfix update 🤔

othercorey: yes, trying to find out what's different.

tests/TestCase/Database/Schema/MysqlSchemaTest.php
```diff
@@ -85,36 +85,36 @@ public static function convertColumnProvider()
...
             [
-                'TINYINT(2)',
-                ['type' => 'tinyinteger', 'length' => 2, 'unsigned' => false],
+                'TINYINT(1) UNSIGNED',
+                ['type' => 'boolean', 'length' => null],
             ],
             [
                 'TINYINT(3)',
-                ['type' => 'tinyinteger', 'length' => 3, 'unsigned' => false],
+                ['type' => 'tinyinteger', 'length' => null, 'unsigned' => false],
```
markstory: Will dropping display lengths impact older versions of MySQL where it wasn't just decorative?

othercorey: It should not. They were onlu used with the `zerofill` attribute which fixtures/schemas don't support.

othercorey: https://dev.mysql.com/doc/refman/8.0/en/numeric-type-syntax.html the top paragraph has the notes on this.

markstory: Ok. It looks like integer widths were only for display purposes as [far back as 5.6 ](https://dev.mysql.com/doc/refman/5.6/en/numeric-type-attributes.html). With that in mind, ignoring display widths seems reasonable as it doesn't impact storage types.

markstory: To avoid surprising people in a bug fix release we might want to move this to 4.1.0.

othercorey: Sounds good. I started it in 4.0 thinking it would be simpler, but with these changes 4.1 makes sense.

PR #14199, opened 2020-01-22
Add ConsoleFormatter for Debugger::exportVar()

markstory:
This formatter adds ANSI escape code formatted output for use in console environments. I've not gone deep into 256 color land as I wanted to maximize compatibility. This also adds methods to Debugger to autoselect the formatter or users will be able to use `Debugger::configInstance()` to choose their preferred output format.

I chose to mostly duplicate the code in `TextFormatter` as I didn't want to clutter that class up with no-op `style()` calls. I'm also not confident in what a good abstraction would be or if its even useful. I hope to know more once I've implemented the HTML formatter.

### DIFF
--- a/src/Error/Debug/ConsoleFormatter.php
+++ b/src/Error/Debug/ConsoleFormatter.php
@@ -185,7 +185,6 @@ class ConsoleFormatter implements FormatterInterface
                 $this->style('punct', ' {}');
         }
 
-        /** @var \Cake\Error\Debug\ClassNode $var */
         $out = $this->style('punct', 'object(') .
             $this->style('class', $var->getValue()) .
             $this->style('punct', ') id:') .
--- a/src/Error/Debug/ConsoleFormatter.php
+++ b/src/Error/Debug/ConsoleFormatter.php
@@ -83,7 +83,7 @@ class ConsoleFormatter implements FormatterInterface
     {
         $code = $this->styles[$style];
 
-        return "\033[{$code};m{$text}\033[0m";
+        return "\033[{$code}m{$text}\033[0m";
     }
 
     /**
--- a/src/Error/Debug/ConsoleFormatter.php
+++ b/src/Error/Debug/ConsoleFormatter.php
@@ -1,0 +1,219 @@
+<?php
+declare(strict_types=1);
+
+/**
+ * CakePHP(tm) : Rapid Development Framework (https://cakephp.org)
+ * Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
+ *
+ * Licensed under The MIT License
+ * For full copyright and license information, please see the LICENSE.txt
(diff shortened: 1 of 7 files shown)


src/Error/Debug/ConsoleFormatter.php
```diff
@@ -0,0 +1,219 @@
...
+     *
+     * @param string $style The style name to use.
+     * @param string $text The text to style.
+     * @return string The styled output.
+     */
+    protected function style(string $style, string $text): string
+    {
+        $code = $this->styles[$style];
+
+        return "\033[{$code};m{$text}\033[0m";
```
ndm2: I'm not sure, but is that closing semicolon there actually valid?

```suggestion
        return "\033[{$code}m{$text}\033[0m";
```

src/Error/Debug/ConsoleFormatter.php
```diff
@@ -0,0 +1,219 @@
...
+     *
+     * @param \Cake\Error\Debug\ArrayNode $var The array to export.
+     * @param int $indent The current indentation level.
+     * @return string Exported array.
+     */
+    protected function _array(ArrayNode $var, int $indent): string
+    {
+        $out = $this->style('punct', '[');
+        $break = "\n" . str_repeat("  ", $indent);
+        $end = "\n" . str_repeat("  ", $indent - 1);
```
ndm2: Could we get 4 spaces of indentation?

src/Error/Debug/ConsoleFormatter.php
```diff
@@ -0,0 +1,219 @@
...
+        }
+
+        $out = $this->style('punct', 'object(') .
+            $this->style('class', $var->getValue()) .
+            $this->style('punct', ') id:') .
+            $this->style('number', (string)$var->getId()) .
+            $this->style('punct', ' {');
+
+        $break = "\n" . str_repeat("  ", $indent);
+        $end = "\n" . str_repeat("  ", $indent - 1) . $this->style('punct', '}');
```
ndm2: Could we get 4 spaces of indentation?

src/Error/Debug/ConsoleFormatter.php
markstory: I went to 2 spaces to try and keep long content in narrow terminals better. I can go back to 4 though.

src/Error/Debug/ConsoleFormatter.php
markstory: It seems to work for me in iTerm with and without the `;` so we can do without.

src/Error/Debug/ConsoleFormatter.php
ndm2: That's a good point... given that everything's consistently formatted, 2 spaces aren't too bad :)

PR #14200, opened 2020-01-22
4.1 - Implemented window frames and exclusions

othercorey:

This completes the sql spec for window expressions. Only integer offsets are implemented in this version. Intervals will be implemented later when query tests are working.

### DIFF
--- a/src/Database/Expression/WindowExpression.php
+++ b/src/Database/Expression/WindowExpression.php
@@ -19,6 +19,7 @@ namespace Cake\Database\Expression;
 use Cake\Database\ExpressionInterface;
 use Cake\Database\ValueBinder;
 use Closure;
+use InvalidArgumentException;
 
 /**
  * This represents a SQL window expression used by aggregate and window functions.
@@ -34,6 +35,16 @@ class WindowExpression implements ExpressionInterface, WindowInterface
      * @var \Cake\Database\Expression\OrderByExpression|null
      */
     protected $order;
+
+    /**
+     * @var array
+     */
+    protected $frame;
+
+    /**
+     * @var string
+     */
+    protected $exclusion;
 
     /**
      * @inheritDoc
@@ -83,7 +94,11 @@ class WindowExpression implements ExpressionInterface, WindowInterface
      */
     public function range(?int $start, ?int $end = 0)
     {
-        return $this;
+        if (func_num_args() === 1) {
+            return $this->frame(self::RANGE, $start, self::PRECEDING);
+        } else {
+            return $this->frame(self::RANGE, $start, self::PRECEDING, $end, self::FOLLOWING);
+        }
     }
 
     /**
@@ -91,7 +106,11 @@ class WindowExpression implements ExpressionInterface, WindowInterface
      */
     public function rows(?int $start, ?int $end = 0)
     {
-        return $this;
+        if (func_num_args() === 1) {
+            return $this->frame(self::ROWS, $start, self::PRECEDING);
+        } else {
+            return $this->frame(self::ROWS, $start, self::PRECEDING, $end, self::FOLLOWING);
+        }
     }
 
     /**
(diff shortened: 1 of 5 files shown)


src/Database/Expression/WindowExpression.php
```diff
@@ -83,22 +94,102 @@ public function order($fields)
...
+
+        $this->frame = [
+            'type' => $type,
+            'start' => [
+                'offset' => $startOffset,
+                'direction' => $startDirection,
+            ],
+        ];
+
+        if (func_num_args() > 3) {
```
lorenzo: What if the user actaully passed all the arguments, but with the null defaults? Would this break?

src/Database/Expression/WindowExpression.php
```diff
@@ -83,22 +94,102 @@ public function order($fields)
...
+
+        if (
+            $type === self::RANGE &&
+            (
+                ($startOffset !== null && !is_int($startOffset) && !is_string($startOffset)) ||
+                ($endOffset !== null && !is_int($endOffset) && !is_string($endOffset))
+            )
+        ) {
+            throw new InvalidArgumentException('Frame offsets for RANGE must be null, integers or interval strings.');
+        }
```
ndm2: What about expression objects? Using functions and other expressions for the frame start/end offset is supported in one way or another, depending on the DBMS.

src/Database/Expression/WindowExpression.php
```diff
@@ -149,12 +248,45 @@ public function sql(ValueBinder $generator): string
...
 
-        return sprintf(
+        $sql = sprintf(
             'OVER (%s%s%s)',
             $partitionSql,
             $partitionSql && $this->order ? ' ' : '',
             $this->order ? $this->order->sql($generator) : ''
         );
+
+        if ($this->frame) {
```
ndm2: The frame bound needs to be placed inside of the `OVER` clause.

src/Database/Expression/WindowExpression.php
othercorey: `func_num_args()` returns the actual number of arguments passed. Even though there are defaults set, this won't include them. This way we can tell if the user wants 'BETWEEN ... FOLLOWING' generated or if they want the default implementation without writing that out.

src/Database/Expression/WindowExpression.php
othercorey: Can you give an specific documentation example that you'd like to see implemented?

src/Database/Expression/WindowExpression.php
othercorey: That is true lol. Looks like I kept following one mistyped test case.

Fortunately, there will be several annoying db query tests implemented for this later.

src/Database/Expression/WindowExpression.php
ndm2: In MySQL for example there's intervals that can be expressed like `INTERVAL 1 DAY PRECEDING` (**https://dev.mysql.com/doc/refman/8.0/en/window-functions-frames.html**).

The docs don't state it in an overly obvious fashion, but Postgres seems to generally allow expressions that evaluate to a compatible type (with a few limitations), being it function calls `SomeFunction() PRECEDING`, mathematical expressions `(1 + 1) PRECEDING`, or even subqueries `(SELECT foo FROM bar) PRECEDING` (**https://www.postgresql.org/docs/12/sql-expressions.html#SYNTAX-WINDOW-FUNCTIONS**).

othercorey: Getting intervals to work is a goal. Beyond that I'm sure we'll run into issues with non-standard support. I want to avoid supporting too many non-standard implementations.

I don't think we I wrote works for intervals, but it's a starting point to have more than just the integer offset. If adding generic expressions is the solution then that will happen.

src/Database/Expression/WindowExpression.php
```diff
@@ -83,22 +94,102 @@ public function order($fields)
      */
     public function range(?int $start, ?int $end = 0)
     {
-        return $this;
+        if (func_num_args() === 1) {
+            return $this->frame(self::RANGE, $start, self::PRECEDING);
+        } else {
```
ADmad: The `else` is not necessary since the `if` part returns from the method. Same for other methods below.

src/Database/Expression/WindowExpression.php
lorenzo: My question is what if the user actually pass more then 3 arguments like this `->frame($a, $b, $c, null, 0)`
Won't this cause a problem aftewards as the offset will be `null`?

src/Database/Expression/WindowExpression.php
```diff
@@ -83,22 +94,102 @@ public function order($fields)
...
+    /**
+     * @inheritDoc
+     */
+    public function frame(
+        int $type,
+        $startOffset,
+        int $startDirection,
+        $endOffset = null,
+        int $endDirection = self::FOLLOWING
+    ) {
```
lorenzo: I have the impression that this should be an object builder instead of a function. It has many arguments of similar types and it will be difficult to remember the order when reading the code. Would it be possible to make this function return a builder or maybe accept a closure with a builder?

src/Database/Expression/WindowExpression.php
lorenzo: I also think that accepting `ExpressionInterface` would make a lot of sense here

src/Database/Expression/WindowExpression.php
```diff
@@ -35,6 +36,16 @@ class WindowExpression implements ExpressionInterface, WindowInterface
      */
     protected $order;
 
+    /**
+     * @var array
+     */
+    protected $frame;
```
ADmad: Wouldn't it be better to intialize the property to empty array?

src/Database/Expression/WindowExpression.php
```diff
@@ -35,6 +36,16 @@ class WindowExpression implements ExpressionInterface, WindowInterface
...
+
+    /**
+     * @var string
+     */
+    protected $exclusion;
```
ADmad: There's a `$this->$exclusion !== null` check in a method, should shouldn't the type be `string|null`?

src/Database/Expression/WindowExpression.php
ADmad: Also even though there are constants for the directions the `int` typehint is disconcerting to read. Personally I would prefer the constants were string.

src/Database/Expression/WindowExpression.php
othercorey: null is a valid offset argument.

src/Database/Expression/WindowExpression.php
othercorey: Well mysql only accepts placeholder, integer literals or the interval value. I was hoping to avoid allowing a lot of invalid expressions here. Keep it simple.

Postgres has similar restrictions even if it might allow some expressions.

src/Database/Expression/WindowExpression.php
othercorey: How would you chain arguments if this returns a builder?

If we have another builder here, that would simply have 2 functions. One for start and one for end. Is that really necessary? Wouldn't it be confusing to have the simle range/rows/groups helper and then a builder?

othercorey: I think strings would be cleaner too. I went with constants based on feedback, but maybe constants as strings is a good way of doing both.

src/Database/Expression/WindowExpression.php
othercorey: I can if that's cleaner, but it doesn't make a difference in terms of implementation since the array is assigned when configured.

src/Database/Expression/WindowExpression.php
othercorey: yup!

src/Database/Expression/WindowExpression.php
othercorey: For this iteration, I reduced the allows offsets to integers. Will tackle how to implement intervals when the query tests are in-place so can validate them.

src/Database/Expression/WindowExpression.php
othercorey: Changed constants to strings and removed the verbose validation in frame().

I'm against a builder as it makes this very complex for two calls: start() and end(). 90% of users will just use the above helpers.

src/Database/Expression/WindowExpression.php
```diff
@@ -174,4 +271,27 @@ public function traverse(Closure $visitor)
...
+    protected function buildOffsetSql(ValueBinder $generator, ?int $offset, string $direction): string
+    {
+        if ($offset === 0) {
+            return 'CURRENT ROW';
+        }
+
+        $sql = sprintf(
+            '%s %s',
+            $offset ?? 'UNBOUNDED',
+            $direction
```
othercorey: FYI, the one issue with using direct string constants/types is allowing user-specified SQL to end up here. Both direction and type strings are used to build the sql string.

Users can force this to happen elsewhere so I don't know if that's an issue.

PR #14201, opened 2020-01-22
Small doc improvement EntityTrait

Iandenh:
In the text above we talk about disabling the guard. I think it makes sense that the example is disabling the guard to prevent confusion.


lorenzo (approved):
Thanks!

PR #14204, opened 2020-01-24
4.1 - Improve unit test times

othercorey:


othercorey:
@ADmad This gets it down to 7m on mysql 8 and 3m on mysql 5.7. 

Due to the way `loadSingle()` works, it's hard to move the `dropConstraints()` call out of `unload()` as well.

However, it's technically correct either way. The table will either be dropped or loaded again which truncates (in both $fixtures and loadSingle format).

ADmad:
Good find :+1:

src/TestSuite/Fixture/FixtureManager.php
```diff
@@ -435,12 +435,6 @@ public function unload(TestCase $test): void
                     $fixture->dropConstraints($db);
                 }
             }
-
-            foreach ($fixtures as $fixture) {
-                if ($this->isFixtureSetup($configName, $fixture)) {
-                    $fixture->truncate($db);
-                }
-            }
```
markstory: This code has been around for a looooong time. However, it doesn't appear to be impacting the testsuite of today, and wouldn't be part of the fixtures recovery behavior for when a test crashes, so lets try it out.

PR #14205, opened 2020-01-25
Handle viewbuilder methods in EmailTrait

### LINKED ISSUE #14078: Mailer test assert on viewBuilder properties
This is a (multiple allowed):

* [x] bug
* [ ] enhancement

* CakePHP Version: 3.8.7

### What you did
```php
$this->assertMailSentWith('default', 'template');
```

### What happened
Deprecation warning about the use of getTemplate() instead of viewBuilder()->getTemplate().

No deprecation warning, maybe provide an assertion on the mailer viewBuilder properties.

markstory:
As a high-level API the `assertMailSentWith` method is more useful if the user doesn't have to worry too much with where properties are coming from.

Fixes #14078

### DIFF
--- a/src/TestSuite/Constraint/Email/MailSentWith.php
+++ b/src/TestSuite/Constraint/Email/MailSentWith.php
@@ -73,7 +73,7 @@ class MailSentWith extends MailConstraintBase
     {
         $viewBuilderMethods = [
             'template', 'layout', 'helpers',
-            'theme', 
+            'theme',
         ];
         if (in_array($this->method, $viewBuilderMethods)) {
             return $email->viewBuilder()->{'get' . ucfirst($this->method)}();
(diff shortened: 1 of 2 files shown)


src/TestSuite/Constraint/Email/MailSentWith.php
```diff
@@ -63,6 +63,25 @@ public function matches($other)
...
+    /**
+     * Read a value from the email
+     *
+     * @param \Cake\Mailer\Email $email The email to read properties from.
+     * @return mixed
+     */
+    protected function getValue($email)
+    {
+        $viewBuilderMethods = [
+            'template', 'layout', 'helpers',
```
othercorey: the split lines are a little odd.

src/TestSuite/Constraint/Email/MailSentWith.php
```diff
@@ -63,6 +63,25 @@ public function matches($other)
...
+            'theme',
+        ];
+        if (in_array($this->method, $viewBuilderMethods)) {
```
othercorey:
```suggestion
        if (in_array($this->method, $viewBuilderMethods, true)) {
```

PR #14209, opened 2020-01-25
Start Html Debug output

markstory:
These changes add an unstyled HTML dump formatter and does some cleanup in the console/text formatters. I will add CSS and Javascript to the HTML dump in an upcoming pull request as to not make this one really big.

I have marked all the output formatters and their interface as internal/unstable. I would like to introduce the concept of unstable APIs into CakePHP as a way to help us be able to ship features without being required to fully commit to not changing them. I envision unstable to be something that lasts on an class/interface for one minor release.


src/Error/Debug/HtmlFormatter.php
```diff
@@ -0,0 +1,218 @@
...
+        $html = $this->export($node);
+
+        return '<div class="cake-dbg">' . $html . '</div>';
+    }
+
+    /**
+     * Convert a tree of NodeInterface objects into HTML
+     *
+     * @param \Cake\Error\Debug\NodeInterface $var The node tree to dump.
+     * @param int $indent The current indentation level.
```
othercorey:
```suggestion
```

src/Error/Debug/HtmlFormatter.php
```diff
@@ -0,0 +1,218 @@
...
+        }
+
+        return $out . $close;
+    }
+
+    /**
+     * Handles object to string conversion.
+     *
+     * @param \Cake\Error\Debug\ClassNode|\Cake\Error\Debug\ReferenceNode $var Object to convert.
+     * @param int $indent Current indentation level.
```
othercorey:
```suggestion
```

src/Error/Debug/HtmlFormatter.php
```diff
@@ -0,0 +1,218 @@
...
+     */
+    protected function exportObject($var): string
+    {
+        $objectId = "cake-db-object-{$this->id}-{$var->getId()}";
+        $out = sprintf(
+            '<div class="cake-dbg-object" id="%s">',
+            $objectId,
+        );
+
+        $end = $this->style('punct', '}') . '</div>';
```
othercorey: Think can move these below the early return to where they are used.

othercorey (approved):

src/Error/Debug/HtmlFormatter.php
```diff
@@ -0,0 +1,215 @@
...
+     * @param \Cake\Error\Debug\ClassNode|\Cake\Error\Debug\ReferenceNode $var Object to convert.
+     * @return string
+     * @see \Cake\Error\Debugger::exportVar()
+     */
+    protected function exportObject($var): string
+    {
+        $objectId = "cake-db-object-{$this->id}-{$var->getId()}";
+        $out = sprintf(
+            '<div class="cake-dbg-object" id="%s">',
+            $objectId,
```
othercorey:
```suggestion
            $objectId
```

PR #14212, opened 2020-01-26
Fix withStatus() overwriting content-type

### LINKED ISSUE #14211: Can't set Content-Type via Response::withHeader()
This is a (multiple allowed):

* [x] bug
* [ ] enhancement

* CakePHP Version: 3.8.x

### What you did

```php
	    if ($controller->getRequest()->is('json')) {
	    	return (new Response())
				->withHeader('Content-Type', 'application/json')
				->withStatus(403)
				->withStringBody(json_encode([
					'redirectUrl' => '/',
				]));
	    }
```

Doing this fixed it.

```php
	    if ($controller->getRequest()->is('json')) {
	    	return (new Response())
				->withType('application/json') // <-- See this (added)
				->withStatus(403)
				->withStringBody(json_encode([
					'redirectUrl' => '/',
				]));
	    }
```

### What happened

Got a text/html response...

Getting an application/json type response.

If the above is intended behavior then this sucks big time. If I'm forced to use withType() for that instead of being able to use withHeader() as well, then withHeader() should throw an exception telling me that I'm not allowed to do so instead of silently overriding my set value and causing me a nice debug orgy.

markstory:
The withStatus() method would use the `_contentType` property which can contain stale information if `withHeader()` is used to set the content-type. Instead of storing potentially divergent information we can make the content-type header authoritative and do type munging in getType().

While this could be a breaking change we make precious few promises around protected properties. This change should be backported to 3.x as well to resolve the same problem there.

Fixes #14211

### DIFF
--- a/src/Http/Response.php
+++ b/src/Http/Response.php
@@ -478,7 +478,7 @@ class Response implements ResponseInterface
      * Formats the Content-Type header based on the configured contentType and charset
      * the charset will only be set in the header if the response is of type text/*
      *
-     * @param string|null $type The type to set.
+     * @param string $type The type to set.
      * @return void
      */
     protected function _setContentType(string $type): void
--- a/src/Http/Response.php
+++ b/src/Http/Response.php
@@ -632,6 +632,7 @@ class Response implements ResponseInterface
         }
         $this->_reasonPhrase = $reasonPhrase;
 
+        // These status codes don't have bodies and can't have content-types.
         if (in_array($code, [304, 204], true)) {
             $this->_clearHeader('Content-Type');
         }
--- a/src/Http/Response.php
+++ b/src/Http/Response.php
@@ -362,14 +362,6 @@ class Response implements ResponseInterface
     protected $_status = 200;
 
     /**
-     * Content type to send. This can be an 'extension' that will be transformed using the $_mimetypes array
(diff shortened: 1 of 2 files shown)


src/Http/Response.php
```diff
@@ -637,7 +631,10 @@ protected function _setStatus(int $code, string $reasonPhrase = ''): void
             $reasonPhrase = $this->_statusCodes[$code];
         }
         $this->_reasonPhrase = $reasonPhrase;
-        $this->_setContentType();
+
+        if (in_array($code, [304, 204], true)) {
```
othercorey: I think this one might deserve a comment.

src/Http/Response.php
```diff
@@ -464,10 +456,11 @@ public function __construct(array $options = [])
             $options['charset'] = Configure::read('App.encoding');
         }
         $this->_charset = $options['charset'];
+        $type = 'text/html';
```
ADmad: I am wondering if this default type should be set by `View` instead of here. Might not be worth doing it right now and potentially break some app but worth considering in future.

markstory: Could be something we do in a future minor release. It shouldn't be a *breaking* change but it would be a small change in behavior.

PR #14216, opened 2020-01-27
Sync arguments of NumberHelper::precision() with Number::precision().

### LINKED ISSUE #14215: NumberHelper parameters issue
* CakePHP Version: 3.8.8

In official documentation, we found 3 parameters for both `\Cake\I18n\Number::precision()` and `\Cake\View\Helper\NumberHelper::precision()`.  But in `\Cake\View\Helper\NumberHelper::precision()` definition, we found only 2 parameters.

ADmad:
Closes #14215

### DIFF
--- a/tests/TestCase/View/Helper/NumberHelperTest.php
+++ b/tests/TestCase/View/Helper/NumberHelperTest.php
@@ -19,8 +19,11 @@ declare(strict_types=1);
 namespace Cake\Test\TestCase\View\Helper;
 
 use Cake\Core\Configure;
+use Cake\I18n\Number;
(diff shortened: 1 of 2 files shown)


tests/TestCase/View/Helper/NumberHelperTest.php
```diff
@@ -76,7 +79,6 @@ public function methodProvider()
             ['toPercentage'],
             ['currency'],
             ['format'],
-            ['addFormat'],
```
othercorey: Is this test working if this wasn't a valid method?

Issue #14213, opened 2020-01-26
`ServerRequest` shouldn't try to build itself.


ADmad:
- Methods like `_processPost()`, `_processFiles()`, `_processGet()` shouldn't exist in `ServerRequest`. All that data pre processing should be consolidated within `ServerRequestFactory` and then passed to `ServerRequest`'s constructor.

- `ServerRequest::_processPost()` has special handling for content type `'application/x-www-form-urlencoded'` to convert the string input data into array. This should be moved into `BodyParserMiddleware` where other content types like `application/json` and `application/xml` are parsed.

- `ServerRequest::input()` should be deprecated. One can use `ServerRequest::getBody()` to get the stream containing the input string. Also the `$callback` argument of the method which allows converting input string to other representations is unnecessary as `BodyParserMiddleware` can handle the same.

markstory:
> Methods like _processPost(), _processFiles(), _processGet() shouldn't exist in ServerRequest

Would you still want to retain backwards compatibility with the current `ServerRequest` constructor? I worry about thrashing a non-trivial amount of tests for a change like this.

ADmad:
> Would you still want to retain backwards compatibility with the current ServerRequest constructor?

The constructor could remain same.

dereuromark:
Can this be closed?

ADmad:
Not yet.

PR #14218, opened 2020-01-27
Move form urlencoded request body parsing from ServerRequest to BodyParserMiddleware

ADmad:
Refs #14213.

### DIFF
--- a/src/Http/Middleware/BodyParserMiddleware.php
+++ b/src/Http/Middleware/BodyParserMiddleware.php
@@ -54,6 +54,7 @@ class BodyParserMiddleware implements MiddlewareInterface
      *
      * ### Options
      *
+     * - `form-urlencoded` Set to false to disable parsing 'application/x-www-form-urlencoded' encoded data.
      * - `json` Set to false to disable json body parsing.
      * - `xml` Set to true to enable XML parsing. Defaults to false, as XML
      *   handling requires more care than JSON does.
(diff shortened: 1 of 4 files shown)


ADmad:
Following this change I think the `BodyParserMiddlware` should be added to the skeleton app's `Application` class by default.

src/Http/Middleware/BodyParserMiddleware.php
```diff
@@ -63,7 +64,13 @@ class BodyParserMiddleware implements MiddlewareInterface
      */
     public function __construct(array $options = [])
     {
-        $options += ['json' => true, 'xml' => false, 'methods' => null];
+        $options += ['form-urlencoded' => true, 'json' => true, 'xml' => false, 'methods' => null];
+        if ($options['form-urlencoded']) {
```
markstory: This change is going to trip people up as they upgrade. I think we can mitigate that a bit by updating the application skeleton now like you suggested. We could also add a deprecation warning in 4.0.x or 4.1.0 that ServerRequest parsing PUT/DELETE/PATCH data is going to move and that they should adopt the middleware.

ADmad: I have created a PR for the app skeleton. Adding deprecation warning is tricky. We tried doing that for `RequestHandler` where JSON parsing was removed but it caused too many false positives and we had to eventually get rid of it. It's difficult to distinguish whether a request body hasn't been parsed at all or an empty request body has been parsed.

markstory: Yeah. I'm concerned we'll break applications in a way that is hard for application developers to detect.

ADmad: In that case I can move this form-urlencoded parsing too into `ServerRequestFactory` along with other things.

markstory: That would be great. Having two copies isn't great but it gives us a low-risk path forward.

ADmad: > Having two copies isn't great

It isn't, so if we are moving it to `ServerRequestFactory` then I wouldn't update `BodyParserMiddleware`, otherwise the input would be unnecessarily parsed twice.

markstory: Ok. Let's keep form-encoded in the request factory as it is a shim around a php limitation.

ADmad:
I'll open a new PR

PR #14219, opened 2020-01-27
Move files array processing into ServerRequestFactory.

ADmad:
Refs #14213


src/Http/ServerRequest.php
```diff
@@ -238,7 +230,6 @@ public function __construct(array $config = [])
             'base' => '',
             'webroot' => '',
             'input' => null,
-            'mergeFilesAsObjects' => true,
```
ADmad: I can retain just the merging of files data into `ServerRequest::$data` as either `UploadedFileInterface` instances or array in `ServerRequest` if that's preferable.

src/Http/ServerRequestFactory.php
```diff
@@ -55,35 +56,75 @@ abstract class ServerRequestFactory implements ServerRequestFactoryInterface
...
+     * Process uploaded files and move things onto the post data.
+     *
+     * @param array $files Files array for normalization and merging in post body.
+     * @param array $post Post body to merge files onto.
+     * @return array
+     * @psalm-return array{files: array, post: array}
+     */
+    protected static function processFiles($files, $post)
+    {
+        $files = normalizeUploadedFiles($files ?: $_FILES);
```
markstory: :clap: Nice that we don't have to maintain this anymore. I like that you kept a few integration tests to ensure laminas keeps their end of the contract up.

ADmad:
Anything else needed to be done here?

PR #14223, opened 2020-01-28
4.1 - Added basic window function tests

othercorey:

### DIFF
--- a/tests/TestCase/Database/WindowQueryTests.php
+++ b/tests/TestCase/Database/WindowQueryTests.php
@@ -1,0 +1,87 @@
+<?php
+declare(strict_types=1);
+
+/**
+ * CakePHP(tm) : Rapid Development Framework (https://cakephp.org)
+ * Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
+ *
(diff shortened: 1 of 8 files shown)


othercorey:
I don't understand why sqlite fails to parse the `OVER` keyword in the PREFER_LOWEST config only.

https://travis-ci.org/cakephp/cakephp/jobs/642780923?utm_medium=notification&utm_source=github_status

tests/TestCase/Database/WindowQueryTests.php
```diff
@@ -0,0 +1,84 @@
...
+     * @var \Cake\Database\Connection
+     */
+    protected $connection = null;
+
+    public function setUp(): void
+    {
+        parent::setUp();
+        $this->connection = ConnectionManager::get('test');
+
+        $skip = env('SKIP_WINDOW_TESTS');
```
markstory: Can we put the skip condition logic into the test here? Without that logic here tests will fail when people run test locally.

othercorey: Updated the skip test logic.

Issue #14226, opened 2020-01-28
Datetime field cleared after saving


jwvanschie:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 4.0.3

### What you did
- Started a new project using the command `composer create-project --prefer-dist cakephp/app:4.*  .`
- From an existing database I baked all the code with the command `bin\cake bake all Games`, including the migration below:
`this->table('games', ['id' => false, 'primary_key' => ['id']])
            ->addColumn('id', 'uuid', [
                'default' => null,
                'limit' => null,
                'null' => false,
            ])
            ->addColumn('event_id', 'uuid', [
                'default' => null,
                'limit' => null,
                'null' => false,
            ])
            ->addColumn('created_at', 'datetime', [
                'default' => null,
                'limit' => null,
                'null' => true,
            ])
            ->addIndex(
                [
                    'event_id',
                ]
            )
            ->create();
`

### What happened
When I go to the view `games/add`, I could save a record for the entity `Game`.
When I go to the view `games/edit/ID` I can change the date and it is saved correctly.
When I go to the view `games/edit/ID` and I change only the event, the date is saved as an empty value.

### What you expected to happen
As the date has not changed, I expect the value remains.

After some research I figured  out it is caused by this commit:
https://github.com/cakephp/cakephp/commit/c63778cbcc6baf5905e2b9454e77babd115057d0

markstory:
Closing as a pull request is open now.

PR #14227, opened 2020-01-28
#14226 - DateTimeWidget now accepts milliseconds. 

jwvanschie:
See ticket #14226.
The commit https://github.com/cakephp/cakephp/commit/c63778cbcc6baf5905e2b9454e77babd115057d0
introduced the milliseconds for fractional types. But it wasn't taking care of saving this format to the database.

### DIFF
--- a/src/Database/Type/DateTimeType.php
+++ b/src/Database/Type/DateTimeType.php
@@ -62,6 +62,7 @@ class DateTimeType extends BaseType
         'Y-m-d H:i:s',
         'Y-m-d\TH:i:s',
         'Y-m-d\TH:i:sP',
+        'Y-m-d\TH:i:s.v',
     ];
 
     /**


markstory (commented):
The fractional part of a datetime column is only saved by datetimefractional type which is a type mapper you have to enable in your application.

ndm2:
The problem is that the date time widget adds fractions to the output by default, irrespective of whether the field actually is of a type that supports fractions.

Since the the regular, non-fraction datetime database type doesn't accept fractions when marshalling, the value is not being accepted, and one ends up with `null` for the field after patching. On top the default `datetime` validation doesn't prevent saving, as it accepts fractions by default.

I think the date time widget should only generate fractions for fields with a fraction type.

othercorey:
> I think the date time widget should only generate fractions for fields with a fraction type.

Agreed. I did test the html widgets in a browser, but I missed something critical here. Not sure if I changed the code after or just never tested both column types like I thought I did.

othercorey:
Closing in favor of https://github.com/cakephp/cakephp/pull/14236

PR #14231, opened 2020-01-29
4.1 - Add CSS to HTML debug output.

markstory:
Add CSS to the HTML formatter. I've used a basic colour scheme from solarized as it is my favorite. I'm happy to refine/change/replace that colour scheme if people want to use a different set of colours.

I've moved the wrapper generation to each formatter as I think it belongs there. While I've maintained backwards compatibility with the `showHtml` parameter I would like to remove this parameter in the future as it doesn't make much sense with the new formatter system.


src/Error/Debug/FormatterInterface.php
```diff
@@ -30,4 +30,13 @@ interface FormatterInterface
...
     public function dump(NodeInterface $node): string;
+
+    /**
+     * Output a dump wrapper with location context.
+     *
+     * @param string $contents The contents to wrap and return
+     * @param array $location The file and line the contents came from.
+     * @return string
+     */
+    public function formatWrapper(string $contents, array $location);
```
othercorey:
```suggestion
    public function formatWrapper(string $contents, array $location): string;
```

src/Error/Debug/HtmlFormatter.php
```diff
@@ -55,19 +60,41 @@ public static function environmentMatches(): bool
     }
 
     /**
-     * Style text with HTML class names
+     * {@inheritDoc}
+     */
+    public function formatWrapper(string $contents, array $location)
```
othercorey:
```suggestion
    public function formatWrapper(string $contents, array $location): string
```

src/Error/Debugger.php
```diff
@@ -920,53 +920,23 @@ public static function getType($var): string
...
+            $location['file'] = static::trimPath((string)$location['file']);
         }
-        if ($showHtml === null && $template !== $text) {
-            $showHtml = true;
+
+        $debugger = static::getInstance();
+        $restore = null;
+        if ($showHtml !== null) {
+            $restore = $debugger->getConfig('exportFormatter');
+            $debugger->setConfig('exportFormatter', $showHtml ? HtmlFormatter::class : TextFormatter::class);
```
othercorey: The default here is a little confusing. Can you better explain what $showHtml = null does in the docblock?

src/Error/Debug/ConsoleFormatter.php
```diff
@@ -75,17 +75,23 @@ public static function environmentMatches(): bool
     }
 
     /**
-     * Style text with ANSI escape codes.
-     *
-     * @param string $style The style name to use.
-     * @param string $text The text to style.
-     * @return string The styled output.
+     * {@inheritDoc}
```
ADmad:
```suggestion
     * @inheritDoc
```

src/Error/Debug/HtmlFormatter.php
```diff
@@ -55,19 +60,41 @@ public static function environmentMatches(): bool
     }
 
     /**
-     * Style text with HTML class names
+     * {@inheritDoc}
```
ADmad:
```suggestion
     * @inheritDoc
```

src/Error/Debug/TextFormatter.php
```diff
@@ -28,6 +28,26 @@
  */
 class TextFormatter implements FormatterInterface
 {
+    /**
+     * {@inheritDoc}
```
ADmad:
```suggestion
     * @inheritDoc
```

src/Error/Debugger.php
markstory: Sure thing.

PR #14236, opened 2020-01-31
Only render DateTimeWidget value with milliseconds for fractional types

othercorey:
https://github.com/cakephp/cakephp/issues/14226

### DIFF
--- a/src/View/Widget/DateTimeWidget.php
+++ b/src/View/Widget/DateTimeWidget.php
@@ -59,7 +59,7 @@ class DateTimeWidget extends BasicWidget
      * @var string[]
      */
     protected $formatMap = [
-        'datetime-local' => 'Y-m-d\TH:i:s.v',
+        'datetime-local' => 'Y-m-d\TH:i:s',
         'date' => 'Y-m-d',
         'time' => 'H:i:s',
         'month' => 'Y-m',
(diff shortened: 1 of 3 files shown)


src/View/Widget/DateTimeWidget.php
```diff
@@ -192,7 +194,21 @@ protected function formatDateTime($value, array $options): string
             $dateTime = $dateTime->setTimezone($timezone);
         }
 
-        return $dateTime->format($this->formatMap[$options['type']]);
+        $format = $this->formatMap[$options['type']];
+        if ($fieldName !== null) {
+            $fractionalTypes = [
```
ADmad: This array is now duplicated at 2 places in the class, so better to set it to a property. If the same set it used at other places in the framework too?

othercorey: I thought this too, but wasn't sure if it was good style to make a list of fractional datetime types in the schema.

I do want something where we don't miss new types again though.

markstory: Could TableSchema have a constant that contains the fractional types?

ADmad:
I was wondering if we could take a simpler approach to fix the issue. Simply truncate the milliseconds from the formatted values if values of `step` is an integer / not decimal. That would be just a 2-3 lines patch.

othercorey:
> Simply truncate the milliseconds from the formatted datetime string if value of `step` attribute is an integer / not decimal. That would be just a 2-3 lines patch.

If users set the step to 0.5 but had a datetime type, the insert/update would fail.

ADmad:
> If users set the step to 0.5 but had a datetime type, the insert/update would fail.

That's users fault for setting invalid attribute.

PR #14239, opened 2020-01-31
Fix and add additional console formatter environment checks.

ndm2:
Guess they wouldn't be pretty, but should there maybe be some tests for the environment check?

### DIFF
--- a/src/Error/Debug/ConsoleFormatter.php
+++ b/src/Error/Debug/ConsoleFormatter.php
@@ -66,7 +66,10 @@ class ConsoleFormatter implements FormatterInterface
         // Windows environment checks
         if (
             DIRECTORY_SEPARATOR === '\\' &&
-            (!(bool)env('ANSICON') || env('ConEmuANSI') !== 'ON')
+            strpos(strtolower(php_uname('v')), 'windows 10') === false &&
+            strpos(strtolower((string)env('SHELL')), 'bash.exe') === false &&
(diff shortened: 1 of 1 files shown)


src/Error/Debug/ConsoleFormatter.php
```diff
@@ -66,7 +66,10 @@ public static function environmentMatches(): bool
         // Windows environment checks
         if (
             DIRECTORY_SEPARATOR === '\\' &&
-            (!(bool)env('ANSICON') || env('ConEmuANSI') !== 'ON')
+            strpos(strtolower(php_uname('v')), 'windows 10') === false &&
```
markstory: Do we need this? If we are running on windows and in bash.exe can't we safely assume we are in a bash emulator?

markstory (commented):
I thought about tests but getting them to pass consistently across windows and linux felt like they would be duplicating the implementation in skip logic

src/Error/Debug/ConsoleFormatter.php
ndm2: That's actually for two different scenarios, as of Windows 10 the console host supports colors out of the box, the bash check would be for earlier Windows versions.

ndm2:
> I thought about tests but getting them to pass consistently across windows and linux felt like they would be duplicating the implementation in skip logic

Yeah, that's what I thought too.

PR #14241, opened 2020-01-31
Fix exceptions not raised when disableErrorHandlerMiddleware() is used.

### LINKED ISSUE #14237: 4.x - Exceptions not testable using disableErrorHandlerMiddleware()
This is a (multiple allowed):

* [x] bug
* [ ] enhancement

* CakePHP Version: 4.0.3
* Platform and Target: PHP 7.2, apache 2.4, MySQL 5.7, 

### What you did

1. Create new project with `composer create-project --prefer-dist cakephp/app`
2. Create a DB with table "mytable"
3. Bake controller `bin/cake bake controller Mytable`
4. Throw an exception in controller

```
class MytableController extends AppController
{
     public function index()
    {
        throw new \Exception();
```

5. Check for the exception in the test created by `bake`

```
class MytableControllerTest extends TestCase
{
     public function testIndex(): void
    {
        $this->disableErrorHandlerMiddleware();
        $this->expectException(\Exception::class);
        $this->get('/mytable/index');
```
6. Run test

```
vendor/bin/phpunit tests/TestCase/Controller/MytableControllerTest.php --filter testIndex
```

### What happened

```
There was 1 failure:

1) App\Test\TestCase\Controller\MytableControllerTest::testIndex
Failed asserting that exception of type "Exception" is thrown.
```

The test should not fail but detect the thrown exception.

Using `disableErrorHandlerMiddleware()` a exception was testable in Cake 3, but this isn't the case in CakePHP 4 (at the moment)?

markstory:
Fix the test to actually hit the code paths we are interested in as they
weren't before.

Fixes #14237

### DIFF
--- a/src/Error/Middleware/ErrorHandlerMiddleware.php
+++ b/src/Error/Middleware/ErrorHandlerMiddleware.php
@@ -131,9 +131,10 @@ class ErrorHandlerMiddleware implements MiddlewareInterface
     public function handleException(Throwable $exception, ServerRequestInterface $request): ResponseInterface
     {
         $errorHandler = $this->getErrorHandler();
+        $renderer = $errorHandler->getRenderer($exception, $request);
 
         try {
-            $response = $errorHandler->getRenderer($exception, $request)->render();
+            $response = $renderer->render();
             $errorHandler->logException($exception, $request);
         } catch (Throwable $internalException) {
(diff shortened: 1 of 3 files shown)


ADmad:
CS errors need to be fixed

PR #14243, opened 2020-02-01
Generate milliseconds for datetime-local inputs only if "step" is decimal.

### LINKED ISSUE #14226: Datetime field cleared after saving

ADmad:
Fixes #14226

Alternative to #14236

### DIFF
--- a/src/View/Widget/DateTimeWidget.php
+++ b/src/View/Widget/DateTimeWidget.php
@@ -59,7 +59,7 @@ class DateTimeWidget extends BasicWidget
      * @var string[]
      */
     protected $formatMap = [
-        'datetime-local' => 'Y-m-d\TH:i:s.v',
+        'datetime-local' => 'Y-m-d\TH:i:s',
         'date' => 'Y-m-d',
         'time' => 'H:i:s',
         'month' => 'Y-m',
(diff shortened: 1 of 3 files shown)


markstory:
This seems like the simpler solution. What do you think @othercorey ?

othercorey:
I'm fine with it. I made the change already.

markstory:
So which pull request are we merging? :smile:

othercorey:
> So which pull request are we merging? smile

Your choice.

PR #14246, opened 2020-02-03
4.1 - Remove request from context classes

diegosardina:
This PR removes the hard-coded ServerRequest in the various Context classes and will use values sources of FormHelper for a more fine-grained and flexible way of reading and writing form's data.

This means that FormHelper will move from ['context'] to ['data', 'context'] to keep the actual behavior. However if one doesn't want that context data get overridden by request data, it's now possible to change priority to ['context', 'data'].

Short summary of changes:

- EntityContext now uses the (already state-holder) entity to store its values, however for usability purpose it must show also invalid values that previously were restored by request data. For this reason val() method will give priority to invalid values.

- FormContext now uses the (unused) data field in Form class to hold and restore its values.

- ArrayContext now has a data section that stands for the current supplied values.

- NullContext is now a true null object :-)

Tests against request data are removed or changed to test against the internal state, added a test for invalid values in the case of EntityContext.

Feel free to modify or even reject it.


ADmad:
This needs to target `4.next` instead of master.

src/Form/Form.php
```diff
@@ -265,6 +265,8 @@ public function setErrors(array $errors)
      */
     public function execute(array $data): bool
     {
+        $this->_data = $data;
```
markstory: Won't this leave potentially invalid data in the object? Could we only set the data once it has been validated?

src/View/Form/ContextFactory.php
```diff
@@ -59,44 +58,44 @@ public static function createWithDefaults(array $providers = [])
         $providers = [
             [
                 'type' => 'orm',
-                'callable' => function ($request, $data) {
+                'callable' => function ($data) {
```
markstory: This is a potentially breaking change for contexts defined in userland code. We should continue passing the request into factory functions just in case someone is using it.

src/View/Form/ContextFactory.php
```diff
@@ -133,18 +132,17 @@ public function addProvider(string $type, callable $check)
      *
      * If no type can be matched a NullContext will be returned.
      *
-     * @param \Cake\Http\ServerRequest $request Request instance.
      * @param array $data The data to get a context provider for.
      * @return \Cake\View\Form\ContextInterface Context provider.
      * @throws \RuntimeException When a context instace cannot be generated for given entity.
      */
-    public function get(ServerRequest $request, array $data = []): ContextInterface
+    public function get(array $data = []): ContextInterface
```
markstory: This should be reverted as well, as it is a breaking change.

src/View/Form/EntityContext.php
```diff
@@ -279,6 +265,12 @@ public function val(string $field, array $options = [])
 
         if ($entity instanceof EntityInterface) {
             $part = end($parts);
+
+            $val = $entity->getInvalidField($part);
```
markstory: This method isn't part of the `EntityInterface` you'd need to do a check for `InvalidPropertyInterface` here as well.

markstory (commented):
Looking good, there are a few breaking changes that we should revert to avoid breaking userland code.

src/Form/Form.php
diegosardina: In the case the form is invalid, `FormContext` wouldn't be able to show previous supplied data (they were restored by request data before).

`data` is (to be considered as) the supplied data, being valid or not.

Otherwise we may split in `$_requestData` and `$_data`.

However in the current behavior we can set any kind of data with $form->setData($invalidData) and we don't know if these are good (except if $form->validate($invalidData) is executed). For this reason I considered reasonable to directly set the request data.

markstory: Good point about `setData()` also allowing invalid data.

othercorey:
Is the postrgres fail related? Is this something we can merge?

The README change isn't needed.

diegosardina:
> Is the postrgres fail related?
> 
If I remember correct that fail happened after the pull request was moved to the correct branch 4.next.

There were some changes related to postgres in 4.next.

othercorey:
github actions is failing to pull the commit for some reason. Can you try rebasing into a single commit and force commit again?

diegosardina:
After merging upstream/4.next to this branch the only failing test was:

_CakePHP CI / Windows - PHP 7.4 & SQL Server (pull_request)_

Later, tests restarted again and reported this git related problem to all CakePHP CI tests.

> Can you try rebasing into a single commit and force commit again?

Unfortunately I'm not familiar with rebasing (I use mercurial) I'm very sorry :-( 

If it doesn't take long, could you do it? Alternatively I can create a new pull request (targeting the correct branch)

othercorey:
@markstory I rebased this on 4.next to allow merging.

PR #14247, opened 2020-02-04
Bring docblock of dateAdd() in line with 4.x

johanmeiring:
This PR updates the documentation of `FunctionsBuilder::dateAdd()` to indicate that both `string` and `\Cake\Database\ExpressionInterface` are acceptable types for the first argument, and both `string` and `int` for the second argument.  This brings the documentation for 3.x in line with what is currently in 4.x for the same method.

### DIFF
--- a/src/Database/FunctionsBuilder.php
+++ b/src/Database/FunctionsBuilder.php
@@ -203,8 +203,8 @@ class FunctionsBuilder
     /**
      * Add the time unit to the date expression
      *
-     * @param string $expression Expression to obtain the date part from.
-     * @param string $value Value to be added. Use negative to subtract.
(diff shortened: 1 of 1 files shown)


dereuromark:
Wouldn't dayOfWeek() and the other "mixed" documented types in this class also be in need of update?
https://github.com/cakephp/cakephp/blob/master/src/Database/FunctionsBuilder.php#L239

johanmeiring:
Yes indeed.  I'll make the changes and push shortly 👍

johanmeiring:
@dereuromark Done

dereuromark:
Nice work :) Thank you!

markstory:
Thank you 👍

PR #14251, opened 2020-02-06
Add getAccessible()

### LINKED ISSUE #14249: Get accessible properties in an Entity
This is a (multiple allowed):

* [ ] bug
* [ ] enhancement

* CakePHP Version: CakePHP 4.0.3
* Platform and Target: Irrelevant

I'm creating an import entity plugin where I want to fetch which fields are accessible in an entity, but it seems like there is no such method available.

The property `$_accessible` is protected, so can't use it directly. Is it ok if I create a PR creating a `getAccessible()` method in the `EntityTrait`? Should I rather get the database schema and loop through the fields using `isAccessible` to check for accessibility? Just seems like unnecessary effort when there's already a property readily available to be returned.

dereuromark:
Resolves https://github.com/cakephp/cakephp/issues/14249

Is this missing getter OK to be added in 4.0? Or do we have to move it to 4.1?
It is fully BC as such.

We can also backport this to 3.x then afterwards.

### DIFF
--- a/src/Datasource/EntityInterface.php
+++ b/src/Datasource/EntityInterface.php
@@ -24,6 +24,7 @@ use JsonSerializable;
  * comply with.
  *
  * @property mixed $id Alias for commonly used primary key.
+ * @method bool[] getAccessible()
  */
 interface EntityInterface extends ArrayAccess, JsonSerializable
 {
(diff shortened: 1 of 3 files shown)


ADmad:
> Is this missing getter OK to be added in 4.0?

We had decided to avoid adding new methods in patch releases. It makes it difficult to keep track of and properly document when a feature was added.

dereuromark:
I changed base to 4.1 then.

Issue #14250, opened 2020-02-06
Description of Table::buildRules is incorrect.


mtak3:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 4.0.3
* Platform and Target: 

### What you did

`Table::buildRules()` 
https://github.com/cakephp/cakephp/blob/4.0.3/src/ORM/Table.php#L2914
description is incorrect
https://github.com/cakephp/cakephp/blob/4.0.3/src/ORM/Table.php#L99-L100

`Behavior::buildRules()` is correct
https://github.com/cakephp/cakephp/blob/4.0.3/src/ORM/Behavior.php#L64-L66
https://github.com/cakephp/cakephp/blob/4.0.3/src/ORM/Behavior.php#L277
https://github.com/cakephp/cakephp/blob/4.0.3/src/Datasource/RulesAwareTrait.php#L99-L100

### What happened

https://github.com/cakephp/docs/blob/4.x/en/appendices/4-0-upgrade-guide.rst
```
 bin/cake upgrade rector --rules cakephp40 <path/to/app/src>
```
execute...

https://github.com/rectorphp/rector/blob/v0.6.14/config/set/cakephp/cakephp40.yaml#L98-L99
"Type declarations" is incorrect

markstory:
Thank you for the great bug report. I've updated our documentation and opened a pull request to fix rector.

PR #14255, opened 2020-02-07
Update table docs.

markstory:
The buildRules() method is slightly different than other hooks as
it doesn't use an event on the Table class.

Refs #14250

### DIFF
--- a/src/ORM/Table.php
+++ b/src/ORM/Table.php
@@ -96,8 +96,10 @@ use RuntimeException;
  * - `buildValidator(EventInterface $event, Validator $validator, string $name)`
  *   Allows listeners to modify validation rules for the provided named validator.
  *
- * - `buildRules(EventInterface $event, RulesChecker $rules)`
- *   Allows listeners to modify the rules checker by adding more rules.
+ * - `buildRules(RulesChecker $rules)`
+ *   Allows table classes to modify the rules checker by adding more rules. The
+ *   `Model.buildRules` event will also include an `$event` parameter before the 
+ *   `$rules` parameter.
  *
  * - `beforeRules(EventInterface $event, EntityInterface $entity, ArrayObject $options, string $operation)`
  *   Fired before an entity is validated using the rules checker. By stopping this event,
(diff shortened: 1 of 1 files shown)


src/ORM/Table.php
```diff
@@ -96,8 +96,10 @@
  * - `buildValidator(EventInterface $event, Validator $validator, string $name)`
  *   Allows listeners to modify validation rules for the provided named validator.
  *
- * - `buildRules(EventInterface $event, RulesChecker $rules)`
- *   Allows listeners to modify the rules checker by adding more rules.
+ * - `buildRules(RulesChecker $rules)`
+ *   Allows table classes to modify the rules checker by adding more rules. The
+ *   `Model.buildRules` event will also include an `$event` parameter before the
```
othercorey: This whole docblock section is confusing. Why are we mixing callbacks and events together?

The comment about `Model.buildRules` is confusing as it doesn't explain the relationship between Table::buildRules and Model.buildRules.

src/ORM/Table.php
```diff
@@ -96,8 +96,10 @@
  * - `buildValidator(EventInterface $event, Validator $validator, string $name)`
```
othercorey: Why do we document these without the `Model.` prefix?

markstory: These are the table methods. Your comment about conflating the methods and events is spot on. I will split up the two sections.

src/ORM/Table.php
```diff
@@ -81,51 +81,59 @@
  * You can use Table::updateAll() and Table::deleteAll() to do bulk updates/deletes.
  * You should be aware that events will *not* be fired for bulk updates/deletes.
  *
- * ### Callbacks/events
+ * ### Events
  *
- * Table objects provide a few callbacks/events you can hook into to augment/replace
- * find operations. Each event uses the standard event subsystem in CakePHP
+ * Table objects emit several events during as life-cycle hooks during find, delete and save
```
garas: ... during as life-cycle hooks during ...

Issue #14254, opened 2020-02-06
Validation of decimal numbers vs non-breaking space as a grouping separator


micdobro:
This is a:

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.8.2
* Platform and Target: LEMP

### What you did
Adding some serious multilang in the past days. In locale `pl_PL` the grouping separator is a "space", but in reality it's a non-breaking space aka `\xc2\xa0`. Validation code does a regular `str_replace()` - and that works as long as there is a value in the form actually containing the non-breaking space. Entering an amount like `1 200,99` is perfectly valid (in the human-input realm) - but obviously that cannot be validated, because the code tries to replace a non-breaking space. 

It works for editing - as my `Decimal` widget outputs via `Number::format()`. 

### What happened
Validation fails for amounts with space as a separator (entered via a form). 

### What you expected to happen
I would expect the validation to take into account the fact that non-breaking space can be also a space. 

In file `Validation/Validation.php` function `decimal()` we have today: 
```php
// account for localized floats.
        $locale = ini_get('intl.default_locale') ?: static::DEFAULT_LOCALE;
        
        $formatter = new NumberFormatter($locale, NumberFormatter::DECIMAL);
        $decimalPoint = $formatter->getSymbol(NumberFormatter::DECIMAL_SEPARATOR_SYMBOL);
        $groupingSep = $formatter->getSymbol(NumberFormatter::GROUPING_SEPARATOR_SYMBOL);

        $check = str_replace([$groupingSep, $decimalPoint], ['', '.'], $check);
       
        return static::_check($check, $regex);
```

I propose: 
```php
// account for localized floats.
        $locale = ini_get('intl.default_locale') ?: static::DEFAULT_LOCALE;
        
        $formatter = new NumberFormatter($locale, NumberFormatter::DECIMAL);
        $decimalPoint = $formatter->getSymbol(NumberFormatter::DECIMAL_SEPARATOR_SYMBOL);
        $groupingSep = $formatter->getSymbol(NumberFormatter::GROUPING_SEPARATOR_SYMBOL);
        
        // Some locales use non-br as separator - add a human space to the list
        if($groupingSep == "\xc2\xa0") {
            $check = str_replace([' ', $groupingSep, $decimalPoint], ['', '', '.'], $check);
        }
        else {
            $check = str_replace([$groupingSep, $decimalPoint], ['', '.'], $check);
        }
        
        
[...]

PR #14256, opened 2020-02-07
Issue #14254 - accounting for human input in decimal validation and w…

micdobro:
Accounting for human input when validating decimals - some locales use non-breaking spaces (whitespace) as separators - and user inputs a normal space. That has to be artificially injected into the `str_replace()` call - otherwise it will "never" validate as it is trying to replace "the wrong whitespace".

### DIFF
--- a/src/Validation/Validation.php
+++ b/src/Validation/Validation.php
@@ -783,6 +783,14 @@ class Validation
         $decimalPoint = $formatter->getSymbol(NumberFormatter::DECIMAL_SEPARATOR_SYMBOL);
         $groupingSep = $formatter->getSymbol(NumberFormatter::GROUPING_SEPARATOR_SYMBOL);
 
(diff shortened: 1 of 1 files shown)


src/Validation/Validation.php
```diff
@@ -783,6 +783,13 @@ public static function decimal($check, $places = null, $regex = null)
         $decimalPoint = $formatter->getSymbol(NumberFormatter::DECIMAL_SEPARATOR_SYMBOL);
         $groupingSep = $formatter->getSymbol(NumberFormatter::GROUPING_SEPARATOR_SYMBOL);
 
+        // There are two types of non-breaking spaces - we inject a space to account for human input
+        if ($groupingSep == "\xc2\xa0" || $groupingSep == "\xe2\x80\xaf") {
```
markstory: It would be good to expand the tests for `decimal()` to cover this new logic so we don't regress in the future.

PR #14258, opened 2020-02-07
Add annotations for GH actions failing test runs.

ADmad:
Check the image at https://github.com/mheap/phpunit-github-actions-printer

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -54,6 +54,10 @@ jobs:
     - name: composer install
       run: composer install
 
+    - name: Configure PHPUnit matcher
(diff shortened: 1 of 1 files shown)


ADmad:
Hmm.. We end up with one annotation from each job for phpunit. 6 annotations for each failure would be annoying. I guess we'll have to use it for only one PHP-database pair.

markstory:
The output from PHPUnit is really hard to read now :cry:

ADmad:
> The output from PHPUnit is really hard to read now

The idea is you now can just see the annotation comments instead of having to go to check the CLI output.

Anyway the annotations output is now used for only 1 job to avoid multiple annotations for same failure.

PR #14261, opened 2020-02-08
Add methods to support editor links

markstory:
This adds methods to Debugger that will help generate editor links in error pages. I'll be updating the error pages in a separate pull request.

### Questions

* Are there other editors that we should include?
* Suggestions on where the editor links should be added in the error page?

### DIFF
--- a/src/Error/Debugger.php
+++ b/src/Error/Debugger.php
@@ -115,10 +115,12 @@ class Debugger
      * @var array
      */
     protected $editors = [
-        'sublime' => 'subl://open?url=file://{file}&line={line}',
-        'phpstorm' => 'phpstorm://open?file={file}&line={line}',
-        'macvim' => 'mvim://open/?url=file://{file}&line={line}',
+        'atom' => 'atom://core/open/file?filename={file}&line={line}',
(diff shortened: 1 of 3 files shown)


src/Error/Debugger.php
```diff
@@ -61,6 +62,7 @@ class Debugger
     protected $_defaultConfig = [
         'outputMask' => [],
         'exportFormatter' => null,
+        'editor' => 'phpstorm',
```
ADmad: This default should be configurable. The constructor could use `Configure::read('Debugger.editor')`.

markstory: :+1: on making it configurable from the application config. I was considering adding an `editor` option to the `Error` key, but if have a few other options having Debugger be separate could be nice.

ADmad: A key under `Error` works too.

josegonzalez:
Maybe a few good ones to add

- textmate
- atom
- intellij

markstory:
I've added atom and textmate. Intellij [seems more complicated though](https://youtrack.jetbrains.com/issue/IDEA-65879).

PR #14265, opened 2020-02-10
Move request method marshalling into ServerRequestFactory.

ADmad:
Parsing of "application/x-www-form-urlencoded" content type for PUT/PATCH/DELETE
methods is also moved to the factory class.

Refs #14213

### DIFF
--- a/src/Auth/DigestAuthenticate.php
+++ b/src/Auth/DigestAuthenticate.php
@@ -123,7 +123,12 @@ class DigestAuthenticate extends BasicAuthenticate
         $password = $user[$field];
         unset($user[$field]);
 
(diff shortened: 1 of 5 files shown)


ADmad:
Testing the parsing of raw php input is an issue right now.

markstory:
> Testing the parsing of raw php input is an issue right now.

Mocking the stream is really hard. Could we conditionally read from a `CAKE_` prefixed environment variable and fallback to reading the stream, or use vfs for this?

ADmad:
> Could we conditionally read from a CAKE_ prefixed environment variable and fallback to reading the stream

Good idea, done so.

PR #14266, opened 2020-02-10
Added support for mediumtext and longtext

wouter-corpoflow:
Added support for mediumtext and longtext. Before this change all mediumtext and longtext columns were changed to text.


wouter-corpoflow:
@dereuromark shouldn't this be added to milestone for 2.10.21 since 2.10.20 is the current release?

othercorey:
I think this will need tests added before it is considered.

othercorey:
@wouter-corpoflow do you plan to follow up? If not, we will close this as 2.x isn't receiving any new features.

wouter-corpoflow:
I would love to add some tests and fix some more issues, for example char isn't recognised either, but I currently don't have time for this :(

othercorey:
Feel free to start a discussion again when ready, but remember 2.x isn't going to release a new minor version so only bug fixes are allowed.

PR #14272, opened 2020-02-11
4.next Allows to set single element inside form data

andrii-pukhalevych:
`public function Form::set()` - Saves a variable or an associative array of variables for use inside form data
Allows to set single element inside form data, similar to `Session::write()`, `Entity::set()` or `View::set()`

Replacement for https://github.com/cakephp/cakephp/pull/13850


dereuromark:
This seems to be missing test cases 
Do we want to continue here?

andrii-pukhalevych:
Added tests similar to `ViewVarsTrait::set()` tests

PR #14275, opened 2020-02-12
Move "url" config processing into separate method.

ADmad:
Refs #14213

Generating the `UriInterface` instance and populating to `$query` property from the `url` option is something that needs to be retained for convenience. There are lot of tests which using the `url` option.

### DIFF
--- a/src/Http/ServerRequest.php
+++ b/src/Http/ServerRequest.php
@@ -258,7 +258,7 @@ class ServerRequest implements ServerRequestInterface
 
         if (isset($config['uri'])) {
             if (!$config['uri'] instanceof UriInterface) {
-                throw new Exception('The `uri` key must be an instanceof Psr\Http\Message\UriInterface.');
+                throw new Exception('The `uri` key must be an instance of ' . UriInterface::class);
             }
             $uri = $config['uri'];
         } else {
--- a/src/Http/ServerRequest.php
+++ b/src/Http/ServerRequest.php
@@ -285,7 +285,7 @@ class ServerRequest implements ServerRequestInterface
 
         $this->data = $config['post'];
         $this->uploadedFiles = $config['files'];
-        $this->query = $this->_processGet($config['query']);
+        $this->query = $config['query'];
         $this->params = $config['params'];
         $this->session = $config['session'];
     }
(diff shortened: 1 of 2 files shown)


ADmad:
All the `ServerRequest::_processGet()` method does now is handle removal of request paths from query.

@markstory Is that still needed? While there are test cases for it as shown below, I am not sure if such cases are possible in practice with our current rewrite URL suggested in docs https://book.cakephp.org/4/en/installation.html. Even when not using URL rewriting the URLs would be like `index.php/controller/action?foo=bar` which would not included the path `/controller/action` in `$_GET`.

https://github.com/cakephp/cakephp/blob/081003ae86022cc0f30d5841801f25b0cf2e07f4/tests/TestCase/Http/ServerRequestTest.php#L1957-L1982

https://github.com/cakephp/cakephp/blob/081003ae86022cc0f30d5841801f25b0cf2e07f4/tests/TestCase/Http/ServerRequestTest.php#L1703-L1724

src/Http/ServerRequest.php
```diff
@@ -257,28 +254,21 @@ protected function _setConfig(array $config): void
             $config['environment']['REQUEST_METHOD'] = 'GET';
         }
 
-        $this->_environment = $config['environment'];
         $this->cookies = $config['cookies'];
 
-        if (isset($config['uri']) && $config['uri'] instanceof UriInterface) {
+        if (isset($config['uri'])) {
+            if (!$config['uri'] instanceof UriInterface) {
+                throw new Exception('The `uri` key must be an instanceof Psr\Http\Message\UriInterface.');
```
othercorey:
```suggestion
                throw new Exception('The `uri` key must be an instance of ' . UriInterface::class);
```

Should we use the generated name?

markstory:
> Is that still needed? While there are test cases for it as shown below, I am not sure if such cases are possible in practice with our current rewrite URL suggested in docs 

They might not be valid anymore. Those tests have been around for a long time and were for non-standard server configurations even then. I think reducing the rewriting setups we need to support is worth trying.

src/Http/ServerRequest.php
```diff
@@ -295,26 +285,47 @@ protected function _setConfig(array $config): void
...
+        if (strpos($config['url'], '?') !== false) {
+            [$config['url'], $config['environment']['QUERY_STRING']] = explode('?', $config['url']);
+
+            parse_str($config['environment']['QUERY_STRING'], $queryArgs);
+            $config['query'] += $queryArgs;
+        }
+
+        $config['environment']['REQUEST_URI'] = $config['url'];
+
+        return $config;
```
markstory: Nice to get this out of the hot path for request handling :clap:

ADmad:
> They might not be valid anymore. Those tests have been around for a long time and were for non-standard server configurations even then. I think reducing the rewriting setups we need to support is worth trying.

Great, I got rid of `_processGet()` and related tests.

Also cleaned up tests to avoid manipulating PHP globals.

markstory:
Looks good to me.  There is a static analysis build failing though.

ADmad:
@markstory That failure is unrelated and I have already fixed it on master.

PR #14279, opened 2020-02-13
Don't try to update counter cache when foreign key value is null.

### LINKED ISSUE #14276: CakePHP 4.0.3 CounterCache on null association
This is a (multiple allowed):

* [x] bug
* [ ] enhancement

* CakePHP Version: 4.0.3
* Platform and Target: N/A

### What you did
UsersTable has Behavior CounterCache with ['Tenants' => ['user_count']]. Save a user with a tenant_id = null

### What happened
Exception is thrown `Expression is missing operator (IS, IS NOT) with null value.`

No exception to be thrown.

ADmad:
Closes #14276

### DIFF
--- a/src/ORM/Behavior/CounterCacheBehavior.php
+++ b/src/ORM/Behavior/CounterCacheBehavior.php
@@ -219,11 +219,24 @@ class CounterCacheBehavior extends Behavior
         array $settings
     ): void {
         $foreignKeys = (array)$assoc->getForeignKey();
-        $primaryKeys = (array)$assoc->getBindingKey();
         $countConditions = $entity->extract($foreignKeys);
-        $updateConditions = array_combine($primaryKeys, $countConditions);
-        $countOriginalConditions = $entity->extractOriginalChanged($foreignKeys);
+        $allNulls = true;
+        foreach ($countConditions as $field => $value) {
+            if ($value === null) {
+                $countConditions[$field . ' IS'] = $value;
+                unset($countConditions[$field]);
+            } else {
+                $allNulls = false;
+            }
+        }
+        if ($allNulls) {
+            return;
+        }
 
+        $primaryKeys = (array)$assoc->getBindingKey();
+        $updateConditions = array_combine($primaryKeys, $countConditions);
+
+        $countOriginalConditions = $entity->extractOriginalChanged($foreignKeys);
         if ($countOriginalConditions !== []) {
             $updateOriginalConditions = array_combine($primaryKeys, $countOriginalConditions);
         }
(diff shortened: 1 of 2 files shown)


src/ORM/Behavior/CounterCacheBehavior.php
```diff
@@ -219,11 +219,24 @@ protected function _processAssociation(
         array $settings
     ): void {
         $foreignKeys = (array)$assoc->getForeignKey();
-        $primaryKeys = (array)$assoc->getBindingKey();
         $countConditions = $entity->extract($foreignKeys);
+        $allNulls = true;
+        foreach ($countConditions as $field => $value) {
```
ADmad: This loop is not pretty but it's to account for multi field foreign key with a mix of null and non null values.

tests/TestCase/ORM/Behavior/CounterCacheBehaviorTest.php
```diff
@@ -175,6 +175,27 @@ public function testAddScope()
...
+            'Users' => [
+                'comment_count',
+            ],
+        ]);
+
+        $entity = new Entity([
+            'title' => 'Orphan comment',
+            'user_id' => null,
+        ]);
+        $this->comment->save($entity);
```
dereuromark: Should this be saveOrFail? Or the assert maybe checks if saved.

ADmad: Without this fix an exception was thrown so no exception is a test enough.

dereuromark: Sure,  but the fake assert could still be a useful one maybe.

dereuromark: Imagine an (accidental) regression where no exception is being thrown still, but now also not saved anymore. Thus my concern.

chopstik:
Hi @dereuromark @ADmad had the same issue and for the most part the commit/version change resolved the issue on the CounterCache behaviour. I still see the same Exception (IS/IS NOT NULL) when using a finder method on the count field. See below:
```
$this->addBehavior('CounterCache', [
	'ContractorCompanies' => [
		'threshold_jobs_count' => [
			'finder' => 'ThresholdJobsCount',
		],
	],
	'Instructions'        => ['jobs_count'],
]);
```
```
public function findThresholdJobsCount(Query $query, array $options)
{
	$query->andWhere(['Jobs.assigned' => true]);
	$currentWeek = Time::parse('monday this week');
	$query->andWhere(
		$query->newExpr()->gte('Jobs.created', $currentWeek->format('Y-m-d H:i:s'))
	);

	return $query;
}
```
Can you see a way to adjust [CounterCacheBehaviour.php](https://github.com/cakephp/cakephp/pull/14279/files/4fe9a9610db72aa487e7a2a00afda6c9e2331dbe#diff-0795798ecd11b7b1d2151551def1a027) to take a finder method into account?

othercorey:
@chopstik Please create a new issue that references this one.

chopstik:
Will do.

chopstik:
Opened issue #14362

PR #14283, opened 2020-02-13
Fix EagerLoader loading optional association with leftJoinWith and contain

### LINKED ISSUE #14281: EagerLoader with leftJoin loading optional association exception when there's no matching keys
This is a (multiple allowed):

* [x] bug
* [ ] enhancement

* CakePHP Version: 4.0.3
* Platform and Target: Any

### What you did
Let's say I have TableA related to TableB, as such:
TableA:
```php
        $this->belongsTo('TableB', [
            'foreignKey' => 'table_b_id',
        ]);
```
TableB:
```php
        $this->hasMany('TableA', [
            'foreignKey' => 'table_b_id',
        ]);
```
Normally I would query then with innerJoin, but I then need to perform a leftJoin because that relation is optional and I want to list both for some reason. Then I build a query like so:
```php
        $query = $this->TableA->find()
            ->contain([
                'TableB',
            ])
            ->leftJoinWith('TableB');
```

### What happened
It works fine if there's at least one relation between the two, like, if I have 2 records and one of then have the relation, it works as intended, but if both doesn't have one, it will blow following exception:

> "Unable to load `TableB` association. Ensure foreign key in `TableA` is selected."

I expected to work fine even if there's no relation on the result set.

I looked into the EagerLoader class and it appear the issue comes from the _groupKeys that causes a if statement on loadExternal method to not match in this case.
https://github.com/cakephp/cakephp/blob/master/src/ORM/EagerLoader.php#L666
```php
               if (empty($collected[$path][$alias])) {
                    continue;
                }
```
Will not be true because the first condition will throw the exception because the $collected array is empty, on _groupKeys the $keys returned is an empty array in this because of this condition:
https://github.com/cakephp/cakephp/blob/master/src/ORM/EagerLoader.php#L834
```php
                if ($parts[2] === true && !isset($result[$parts[1][0]])) {
                    continue;
                }
```
Removing this condition and inserting one inside the second one in _groupKeys appear to fix the issue and return the expected result to the loadExternal method.
```php
                if ($parts[2] === true) {
                    if (empty($keys[$nestKey][$parts[0]]) && !isset($result[$parts[1][0]])) {
                        $keys[$nestKey][$parts[0]] = [];
                    } else {
                        $value = $result[$parts[1][0]];
                        $keys[$nestKey][$parts[0]][$value] = $value;
                    }
                    continue;
                }
```
I would be happy to provide a PR for this if my reasoning is correct.

victoreassis:
Solves #14281 

It seems to work.

I added a test on ORM/QueryTest.php to check:
- If it works when there's some results with the associated filled and some not (this already worked as intended).
- If it works when all the results have a null associate (this was the case that needed a fix).
- If the table doesn't have an association (the exception that was thrown before).

I really don't know if it is on the right place and if it is a good enough test.

### DIFF
--- a/src/ORM/EagerLoader.php
+++ b/src/ORM/EagerLoader.php
@@ -832,6 +832,9 @@ class EagerLoader
             foreach ($collectKeys as $nestKey => $parts) {
                 if ($parts[2] === true) {
                     // Missed joins will have null in the results.
+                    if (!array_key_exists($parts[1][0], $result)) {
+                        continue;
+                    }
                     // Assign empty array to avoid not found association when optional.
                     if (empty($keys[$nestKey][$parts[0]]) && !isset($result[$parts[1][0]])) {
                         $keys[$nestKey][$parts[0]] = [];
@@ -850,7 +853,6 @@ class EagerLoader
                 $keys[$nestKey][$parts[0]][implode(';', $collected)] = $collected;
             }
         }
-
         $statement->rewind();
 
         return $keys;
--- a/src/ORM/EagerLoader.php
+++ b/src/ORM/EagerLoader.php
@@ -830,13 +830,15 @@ class EagerLoader
         $keys = [];
         foreach (($statement->fetchAll('assoc') ?: []) as $result) {
             foreach ($collectKeys as $nestKey => $parts) {
-                // Missed joins will have null in the results.
-                if ($parts[2] === true && !isset($result[$parts[1][0]])) {
-                    continue;
-                }
                 if ($parts[2] === true) {
-                    $value = $result[$parts[1][0]];
-                    $keys[$nestKey][$parts[0]][$value] = $value;
+                    // Missed joins will have null in the results.
+                    // Assign empty array to avoid not found association when optional.
+                    if (empty($keys[$nestKey][$parts[0]]) && !isset($result[$parts[1][0]])) {
+                        $keys[$nestKey][$parts[0]] = [];
+                    } else {
+                        $value = $result[$parts[1][0]];
+                        $keys[$nestKey][$parts[0]][$value] = $value;
+                    }
                     continue;
                 }
 
--- a/src/ORM/EagerLoader.php
+++ b/src/ORM/EagerLoader.php
@@ -836,8 +836,10 @@ class EagerLoader
                         continue;
                     }
                     // Assign empty array to avoid not found association when optional.
-                    if (empty($keys[$nestKey][$parts[0]]) && !isset($result[$parts[1][0]])) {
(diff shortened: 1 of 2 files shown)


src/ORM/EagerLoader.php
```diff
@@ -830,13 +830,18 @@ protected function _groupKeys(BufferedStatement $statement, array $collectKeys):
...
-                }
                 if ($parts[2] === true) {
-                    $value = $result[$parts[1][0]];
-                    $keys[$nestKey][$parts[0]][$value] = $value;
+                    // Missed joins will have null in the results.
+                    if (!array_key_exists($parts[1][0], $result)) {
+                        continue;
+                    }
+                    // Assign empty array to avoid not found association when optional.
+                    if (empty($keys[$nestKey][$parts[0]]) && !isset($result[$parts[1][0]])) {
```
markstory: Should this be `isset()` instead of empty()? empty() can swallow '' and 0.

tests/TestCase/ORM/QueryTest.php
```diff
@@ -3412,6 +3412,105 @@ public function testLeftJoinWithAutoFields()
...
+        $expected = [
+            [
+                'id' => 5,
+                'author_id' => null,
+                'title' => 'Fourth Article',
+                'body' => 'Fourth Article Body',
+                'published' => 'N',
+                'author' => null,
+            ],
+        ];
```
markstory: What is this test covering that the first insert/read isn't? They look very similar to me.

markstory (commented):
Thanks for taking the initiative to put this together :clap:

tests/TestCase/ORM/QueryTest.php
victoreassis: The first test covers when there's result with and without the association, some associations are null, and the second covers when there's no results with association, all associations are null.

src/ORM/EagerLoader.php
victoreassis: I guess it could be. I will change it right away, although I guess it will never have those values as the way the array is built, but is better to be safe.

tests/TestCase/ORM/QueryTest.php
victoreassis: I made some changes to tests to remove the second insert doing a query only on null associations.

src/ORM/EagerLoader.php
victoreassis: I did the change and some other one to avoid the presence of ```['' => null]``` on the result keys set when there's some null associations.

markstory:
Thank you 👍

PR #14284, opened 2020-02-13
Deprecate ServerRequest::input().

ADmad:
Refs #14213

### DIFF
--- a/src/Http/ServerRequest.php
+++ b/src/Http/ServerRequest.php
@@ -1250,6 +1250,9 @@ class ServerRequest implements ServerRequestInterface
      *
      * Any additional parameters are applied to the callback in the order they are given.
      *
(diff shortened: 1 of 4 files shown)


markstory (approved):
Looks good to me.

src/Http/ServerRequest.php
```diff
@@ -1258,6 +1261,11 @@ public function getData(?string $name = null, $default = null)
      */
     public function input(?callable $callback = null, ...$args)
     {
+        deprecationWarning(
```
ADmad: I am having second thoughts about this explicit warning, it can be annoying to people. The `@deprecated` tag will already warn IDE users.

markstory: This won't show up for more people unless they are calling it though.

PR #14285, opened 2020-02-13
Change autoParagraph method of TextHelper to accept null as parameter

### LINKED ISSUE #14271: CakePHP 4.0.3 - Relax $this->Text->autoParagraph(h($model->field)); to accept NULL.
This is a:

* [x] bug
* [x] enhancement

* CakePHP Version: 4.0.3.

In my view.php I have $this->Text->autoParagraph(h($model->field)); as per the baked view. It should show the field value.

If the field is NULL, an error is raised:

> Argument 1 passed to Cake\View\Helper\TextHelper::autoParagraph() must be of the type string, null given

It would be handy if this function handled NULL values gracefully. I'm currently working around this by using:

$this->Text->autoParagraph(h($model->field ?? ''))

victoreassis:
Closes #14271 

Simple change on the method docblock, his signature and a test case to cover the change.

### DIFF
--- a/src/View/Helper/TextHelper.php
+++ b/src/View/Helper/TextHelper.php
@@ -295,12 +295,13 @@ class TextHelper extends Helper
      *  <br /> added for single line return
      *  <p> added for double line return
      *
-     * @param string $text Text
+     * @param string|null $text Text
      * @return string The text with proper <p> and <br /> tags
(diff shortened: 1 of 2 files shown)


markstory:
Thank you 👍

PR #14287, opened 2020-02-13
AJAX Flash Header support (if AJAX HTTP request contains "X-Get-Flash: yes")

Ali1:
At times, it would be useful if flash messages could also be sent with AJAX responses. These would need to be handled client side with javascript notifying the user of the message.

### DIFF
--- a/src/Controller/Component/FlashComponent.php
+++ b/src/Controller/Component/FlashComponent.php
@@ -73,7 +73,7 @@ class FlashComponent extends Component
             return null;
         }
 
-        $flash = $session->read("Flash.$key");
+        $flash = $session->read("Flash");
         if (!is_array($flash)) {
             throw new UnexpectedValueException('Value for Flash setting must be an array');
         }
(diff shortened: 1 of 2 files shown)


Ali1:
Example response header:
`X-Flash: {"flash":[{"message":"250.00 recorded with method: Cash","type":"success","params":[]}]}`

src/Controller/Component/FlashComponent.php
```diff
@@ -40,11 +42,72 @@ class FlashComponent extends Component
...
+        $controller = $event->getSubject();
+
+        if (!$controller->getRequest()->is('ajax')) {
+            return null;
+        }
+
+        if (
+            !in_array('yes', array_map( // case insensitive
+                'strtolower',
+                $controller->getRequest()->getHeader('X-Get-Flash')
```
markstory: Could these extensions be done in a sub-class? Without other tools that support this header and its response header we end up with a bespoke implementation that nothing else supports.

Ali1: Yep - just wanted to canvas if there was interest in this in core.

ADmad:
This additional complexity doesn't need to be in the core. It's a good candidate for a community plugin.

dereuromark:
https://github.com/dereuromark/cakephp-flash/pull/8#issuecomment-586034562

PR #14289, opened 2020-02-15
4.1 - Clean up ServerRequest tests

ADmad:
Removed duplicate tests and moved tests using `ServerRequestFactory` to it's test file.


markstory:
Nice work 👏

PR #14298, opened 2020-02-18
Use FILEINFO_MIME_TYPE constant for mime type validation

andrii-pukhalevych:
Use FILEINFO_MIME_TYPE constant for mime type validation
https://www.php.net/manual/en/fileinfo.constants.php#constant.fileinfo-mime-type


markstory:
Thank you 👍

PR #14306, opened 2020-02-20
Fix HasMany::_saveTarget method assign property to wrong entity

### LINKED ISSUE #14303: HasMany::_saveTarget attempts to set original to failed target entity
The flow in `Cake\ORM\Association\HasMany::_saveTarget` seems to indicate that when one of the target entities fails to save because of any errors, the parent entity's property is reverted to the original set of entities, with the appropriate errors.

The line at which this happens seems to be incorrect, as it uses `$entity->set($this->getProperty(), $original)` instead of `$parentEntity->set($this->getProperty(), $original)`, see https://github.com/cakephp/cakephp/blob/master/src/ORM/Association/HasMany.php#L235

E.g. Users hasMany Articles, save a user A with a new set of articles X, one of the articles Y has an error because of rules, this line will attempt to set the set of articles X (with the errors of Y set to the original Y found X) on the article Y instead of the user A.

victoreassis:
Fixes #14303 

- Simple removal of the unnecessary set of property on entity.
- Add test case to cover it.
- Set _invalid content of original entity according to the value of the cloned one.

There's a reason for the getInvalid()/setInvalid() is not part of the EntityInterface?

### DIFF
--- a/src/ORM/Association/HasMany.php
+++ b/src/ORM/Association/HasMany.php
@@ -21,6 +21,7 @@ use Cake\Collection\Collection;
 use Cake\Database\Expression\FieldInterface;
 use Cake\Database\Expression\QueryExpression;
 use Cake\Datasource\EntityInterface;
+use Cake\Datasource\InvalidPropertyInterface;
 use Cake\ORM\Association;
 use Cake\ORM\Association\Loader\SelectLoader;
 use Cake\ORM\Query;
@@ -197,9 +198,10 @@ class HasMany extends Association
      * target entity, and the parent entity.
      * @param \Cake\Datasource\EntityInterface $parentEntity The source entity containing the target
      * entities to be saved.
-     * @param array $entities list of entities to persist in target table and to
-     * link to the parent entity
+     * @param \Cake\Datasource\EntityInterface[]|\Cake\Datasource\InvalidPropertyInterface[] $entities list of entities
+     * to persist in target table and to link to the parent entity
      * @param array $options list of options accepted by `Table::save()`.
+     *
      * @return bool `true` on success, `false` otherwise.
      */
     protected function _saveTarget(
@@ -231,8 +233,12 @@ class HasMany extends Association
             }
 
             if (!empty($options['atomic'])) {
-                $original[$k]->setErrors($entity->getErrors());
-                $original[$k]->setInvalid($entity->getInvalid());
+                if ($original[$k] instanceof EntityInterface && $entity instanceof EntityInterface) {
+                    $original[$k]->setErrors($entity->getErrors());
+                }
+                if ($original[$k] instanceof InvalidPropertyInterface && $entity instanceof InvalidPropertyInterface) {
+                    $original[$k]->setInvalid($entity->getInvalid());
+                }
 
                 return false;
             }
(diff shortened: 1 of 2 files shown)


dereuromark:
Probably forgotten. We should add them as annotational methods though in https://github.com/cakephp/cakephp/blob/master/src/Datasource/EntityInterface.php#L26

ADmad:
https://github.com/cakephp/cakephp/pull/14176#issuecomment-575861804

src/Datasource/EntityInterface.php
```diff
@@ -24,6 +24,11 @@
  * comply with.
  *
  * @property mixed $id Alias for commonly used primary key.
+ *
+ * @method array getInvalid()
+ * @method mixed|null getInvalidField(string $field)
+ * @method $this setInvalid(array $fields, bool $overwrite = false)
+ * @method $this setInvalidField(string $field, $value)
```
dereuromark: I actually see them now in InvalidPropertyInterface.

dereuromark: `class Entity implements EntityInterface, InvalidPropertyInterface`
So we dont need them after all. Sry.

victoreassis:
Right. I will undo the last commit then, I had no idea.

The static check could not find the methods declaration either because it's looking into the ```EntityInterface``` and not on the ```Entity``` that implements both ```EntityInterface``` and ```InvalidPropertyInterface```.

What should be done? Just ignore this specific test or change the docblock of the method to be ```Entity[] $entities```

dereuromark:
I would probably ignore the test issue here, or we annotate using `Interface1|Interface2` etc.

victoreassis:
```\Cake\Datasource\EntityInterface[]&\Cake\Datasource\InvalidPropertyInterface[]``` seems to work fine.

**Comment by dereuromark** [comment]
In a future version we might want to fix up

    if (!($entity instanceof EntityInterface)) {

to the concrete Entity object, as that is more true based on the other interface also needed in some cases.

But for now this is probably OK.

**Comment by victoreassis** [comment]
Well, it works on the IDE type hint (PHP Storm here), but not on the static analysis...

**Comment by dereuromark** [comment]
The clean way would be to
```php
if (!empty($options['atomic']) && $entity instanceof InvalidPropertyInterface) {
    ...
}
```

@ADmad Is that acceptable? Silent no op for the interface missing?

src/ORM/Association/HasMany.php
```diff
@@ -197,9 +197,10 @@ public function saveAssociated(EntityInterface $entity, array $options = [])
      * target entity, and the parent entity.
      * @param \Cake\Datasource\EntityInterface $parentEntity The source entity containing the target
      * entities to be saved.
-     * @param array $entities list of entities to persist in target table and to
-     * link to the parent entity
+     * @param \Cake\Datasource\EntityInterface[]|\Cake\Datasource\InvalidPropertyInterface[] $entities list of entities
```
ADmad: It should be `(\Cake\Datasource\EntityInterface&\Cake\Datasource\InvalidPropertyInterface)[]`

markstory:
Thank you :+1:

Issue #14307, opened 2020-02-20
CakePHP 4.03 installed with improper permissions on tmp SQLSTATE[HY000] [14] unable to open database file


cnizzardini:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 4.0.3

### What you did

1. Installed CakePHP using `composer create-project --prefer-dist cakephp/app:4.* myapp` and said Yes to Set Folder Permissions ? (Default to Y).

2. After configuring a virtual host in Apache/2.4.29 on Ubuntu 18.04 LTS I browsed to the localhost. 

### What happened

Encountered error: `SQLSTATE[HY000] [14] unable to open database file`

```
PDO->__construct
CORE/src/Database/Driver.php:102
Cake\Database\Driver->_connect
CORE/src/Database/Driver/Sqlite.php:79
Cake\Database\Driver\Sqlite->connect
CORE/src/Database/Schema/BaseSchema.php:47
Cake\Database\Schema\BaseSchema->__construct
CORE/src/Database/Dialect/SqliteDialectTrait.php:173
Cake\Database\Driver\Sqlite->schemaDialect
CORE/src/Database/Schema/Collection.php:53
Cake\Database\Schema\Collection->__construct
CORE/src/Database/Connection.php:391
Cake\Database\Connection->getSchemaCollection
ROOT/vendor/cakephp/debug_kit/src/Model/Table/LazyTableTrait.php:44
DebugKit\Model\Table\RequestsTable->ensureTables
ROOT/vendor/cakephp/debug_kit/src/Model/Table/RequestsTable.php:53
DebugKit\Model\Table\RequestsTable->initialize
CORE/src/ORM/Table.php:310
Cake\ORM\Table->__construct
CORE/src/ORM/Locator/TableLocator.php:263
Cake\ORM\Locator\TableLocator->_create
CORE/src/ORM/Locator/TableLocator.php:217
Cake\ORM\Locator\TableLocator->get
ROOT/vendor/cakephp/debug_kit/src/ToolbarService.php:232
DebugKit\ToolbarService->saveData
ROOT/vendor/cakephp/debug_kit/src/Middleware/DebugKitMiddleware.php:66
DebugKit\Middleware\DebugKitMiddleware->process
CORE/src/Http/Runner.php:73
Cake\Http\Runner->handle
CORE/src/Http/Runner.php:58
Cake\Http\Runner->run
[...]

PR #14308, opened 2020-02-20
Improve error message when driver cannot connect to database.

ADmad:
There are various places where Driver::connect() is called directly and not through Connection::connect().
So making Driver::connect() itself deal with PDOException allows showing a better error message.

Refs #14307

### DIFF
--- a/templates/Error/missing_connection.php
+++ b/templates/Error/missing_connection.php
@@ -17,15 +17,7 @@ $this->layout = 'dev_error';
 $this->assign('templateName', 'missing_connection.php');
 $this->assign('title', 'Missing Database Connection');
 
-
-$this->start('subheading'); ?>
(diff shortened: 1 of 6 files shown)


templates/Error/missing_connection.php
```diff
@@ -17,15 +17,7 @@
...
-
-$this->start('subheading'); ?>
-A Database connection using was missing or unable to connect.
-<br>
-<?php
-if (isset($reason)):
-    echo sprintf('The database server returned this error: %s', h($reason));
-endif;
-$this->end();
+$this->assign('subheading', h($message));
```
ADmad: We can directly use the error message here as it's decent enough now. For e.g.
```
Connection to Sqlite could not be established: SQLSTATE[HY000] [14] unable to open database file 
```

PR #14310, opened 2020-02-21
Remove redundant instanceof check.

ADmad:
Fix type in docblock.

Refs #14306

### DIFF
--- a/src/ORM/Association/HasMany.php
+++ b/src/ORM/Association/HasMany.php
@@ -198,7 +198,7 @@ class HasMany extends Association
      * target entity, and the parent entity.
      * @param \Cake\Datasource\EntityInterface $parentEntity The source entity containing the target
      * entities to be saved.
-     * @param \Cake\Datasource\EntityInterface[]|\Cake\Datasource\InvalidPropertyInterface[] $entities list of entities
+     * @param array $entities list of entities
      * to persist in target table and to link to the parent entity
      * @param array $options list of options accepted by `Table::save()`.
      *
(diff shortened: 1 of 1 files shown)


src/ORM/Association/HasMany.php
```diff
@@ -198,7 +198,7 @@ public function saveAssociated(EntityInterface $entity, array $options = [])
      * target entity, and the parent entity.
      * @param \Cake\Datasource\EntityInterface $parentEntity The source entity containing the target
      * entities to be saved.
-     * @param \Cake\Datasource\EntityInterface[]|\Cake\Datasource\InvalidPropertyInterface[] $entities list of entities
+     * @param array $entities list of entities
```
ADmad: This needs to remain as `array`. The `foreach` loop for entities checks whether the array item is an `EntityInterface` instance, which means it could be something else too. We even have test for it https://github.com/cakephp/cakephp/blob/0cbace0fa29fda713f611f29135bdf7db443366d/tests/TestCase/ORM/Association/HasManyTest.php#L583.

src/ORM/Association/HasMany.php
```diff
@@ -233,10 +233,8 @@ protected function _saveTarget(
             }
 
             if (!empty($options['atomic'])) {
-                if ($original[$k] instanceof EntityInterface && $entity instanceof EntityInterface) {
```
ADmad: This check was redundant as `$entity instanceof EntityInterface` is the 1st thing that's checked as start of this loop.

PR #14317, opened 2020-02-22
Possible Missiong Inflector Class methods

Shabasha:
It is not considered an issue neither a bug , it is a kind of improvement that can affect cakephp 2.x , 3.x and 4.x branchs , this request is about the Inflector that is converting plural to singular and singular to plural , i think that before making any further conversion , it would be better to check whether the word is already plural or singular , doing so may let developer or maintainer to check before to convert , i don't know if it is usefull for all of you but let me explain first :

in current Inflector Class implementation in Cake Utility Namespace , the inflector plays singularizations and pluralizations without any check , in this request that i have opened i have added in the Inflector Class these two methods that will do this check as follows 👍 

/**
     * Return true if $word is in plural form.
     *
     * @param string $word Word in singular
     * @return string Word in plural
     * @link https://book.nativephp.org/4/en/core-libraries/inflector.html#creating-plural-singular-forms
     */
	 public static function is_plural(string $word): int 
	 {
		      $singular = $word;
              $plural   = static::pluralize($word);
              if($singular==$plural) { return 1; }
              else                   { return 0;  }			  
	 }
	 
	 /**
     * Return true if $word is in singular form.
     *
     * @param string $word Word in singular
     * @return string Word in plural
     * @link https://book.nativephp.org/4/en/core-libraries/inflector.html#creating-plural-singular-forms
     */
	 public static function is_singular(string $word): int 
	 {
		      $singular = $word;
              $plural   = static::singularize($word);
              if($singular==$plural) { return 1; }
              else                   { return 0;  }			  
	 }

the first method is used when to find out if the given word is already plural or not , we will assume that given word is not plural and it is a singular , by trying to pluralize it , we must get in the final a match between this word and the plural version of it , if it is so , then the function return TRUE or 1 , else FALSE or 0. 

in the second one , the given word is checked as plural not as a singular , by making the same , if the given matches the singular form of it then it is a singular otherwise it is a plural. 

I think by doing like this we can make any conversion from singular to plural and from plural to singular by checking the word and applying that conversion in our context. 

Hope it helps. 

N.B : Consider doing the same for the others methods , camelize , humanize ...  and so on.


othercorey:
I don't know what this is supposed to be but you're merging 4.next into master.

othercorey:
Please open a ticket with the explanation or a new pull request with the correct branch.

PR #14319, opened 2020-02-23
Add job to run testsuite with MSSQL on Windows

ADmad:
Does anybody have ideas how to fix the failing tests for CURL and currency formatting (related to ICU) on Windows?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -78,6 +78,38 @@ jobs:
           vendor/bin/phpunit --verbose
         fi
 
+  testsuite-windows:
+    runs-on: windows-2019
+    name: Windows - PHP 7.4 & mssql
+
+    steps:
+    - uses: actions/checkout@v1
+      with:
+        fetch-depth: 1
+
+    - name: Setup PHP
+      uses: shivammathur/setup-php@v1
+      with:
+        php-version: '7.4'
+        extensions: mbstring, intl, apcu, pdo_sqlsrv, openssl, curl
+        ini-values: apc.enable_cli = 1, extension = php_fileinfo.dll
+        coverage: none
+
+    - name: Setup SQLServer
+      run: |
+        # MSSQLLocalDB is the default SQL LocalDB instance
+        SqlLocalDB start MSSQLLocalDB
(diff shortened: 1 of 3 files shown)


garas:
CLDR 36 has new `en_IN` currency format in [CLDR-13097](https://unicode-org.atlassian.net/browse/CLDR-13097). Tests on Linux will start to fail too once ICU/CLDR will be updated.

Curl tests expects not empty content on `http://localhost`, but server responds with empty content. It is unclear what server listens on port 80 on Github Actions Windows VM. Need some more reliable target server, or to skip test.

ADmad:
> CLDR 36 has new en_IN currency format in CLDR-13097. Tests on Linux will start to fail too once ICU/CLDR will be updated.

I updated the tests to not use `en_IN` currency format.

> Curl tests expects not empty content on http://localhost, but server responds with empty content. It is unclear what server listens on port 80 on Github Actions Windows VM. Need some more reliable target server, or to skip test.

I removed the assertion for non empty content. Lack of exception means the request was successful and there's also an assertion for non empty headers which does pass.

Also updated the test for `http://dummy/?sleep` URL to account for both possible error messages on linux and windows.

ADmad:
> Tests: 10129, Assertions: 28788, Skipped: 211, Incomplete: 5

On AppVeyor

> Tests: 10129, Assertions: 29998, Skipped: 179, Incomplete: 5.

On Github, so 32 more tests run :slightly_smiling_face:

.github/workflows/ci.yml
```diff
@@ -78,6 +78,38 @@ jobs:
           vendor/bin/phpunit --verbose
         fi
 
+  testsuite-windows:
+    runs-on: windows-2019
+    name: Windows - PHP 7.4 & mssql
```
othercorey:
```suggestion
    name: Windows - PHP 7.4 & SQL Server
```

othercorey:
I can't seem to find what version of SQL Server is provided in the `windows-2019` image. I see the various add-on packages, but not the actual server version.

ADmad:
@othercorey https://github.com/actions/virtual-environments/issues/413#issuecomment-589949674

othercorey (approved):

PR #14320, opened 2020-02-24
Add slug() to TextHelper

dereuromark:
In the view templates once is forced to use the static access

    echo \Cake\Utility\Text::slug($name)

where needed - formerly known as Inflector::slug().

If we had the helper delegating to this (new) method as well, we could  do

    echo $this->Text->slug($name);

Any objections?

### DIFF
--- a/src/View/Helper/TextHelper.php
+++ b/src/View/Helper/TextHelper.php
@@ -396,6 +396,32 @@ class TextHelper extends Helper
     }
 
     /**
+     * Returns a string with all spaces converted to dashes (by default),
+     * characters transliterated to ASCII characters, and non word characters removed.
+     *
+     * ### Options:
+     *
+     * - `replacement`: Replacement string. Default '-'.
+     * - `transliteratorId`: A valid transliterator id string.
+     *   If `null` (default) the transliterator (identifier) set via
+     *   `Text::setTransliteratorId()` or `Text::setTransliterator()` will be used.
+     *   If `false` no transliteration will be done, only non words will be removed.
+     * - `preserve`: Specific non-word character to preserve. Default `null`.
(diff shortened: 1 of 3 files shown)


src/View/Helper/TextHelper.php
```diff
@@ -395,6 +395,32 @@ public function toList(array $list, ?string $and = null, string $separator = ',
...
+     *   If `false` no transliteration will be done, only non words will be removed.
+     * - `preserve`: Specific non-word character to preserve. Default `null`.
+     *   For e.g. this option can be set to '.' to generate clean file names.
+     *
+     * @param string $string the string you want to slug
+     * @param array|string $options If string it will be use as replacement character
+     *   or an array of options.
+     * @return string
+     * @see \Cake\Utility\Text::setTransliterator()
+     * @see \Cake\Utility\setTransliteratorId()
```
ADmad: `Text::` missing

othercorey:
Why not add this to 4.0 ?

markstory:
> Why not add this to 4.0 ?

We've been trying not to add methods in bug fix releases as it makes documentation and plugins harder to use because of patch release constraints.

tests/TestCase/View/Helper/TextHelperTest.php
```diff
@@ -634,4 +634,36 @@ public function testAutoParagraph()
...
+    }
+
+    /**
+     * @return array
+     */
+    public function slugInputProvider(): array
+    {
+        return [
+            [
+                'Foo Bar: Not just for breakfast any-more', [],
```
othercorey: The split array lines here looks weird, but don't know that it matters for tests.

dereuromark: Copy and paste from utility tests

PR #14322, opened 2020-02-24
Use caching to speed up CI on GH workflow

ADmad:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -47,12 +47,16 @@ jobs:
     - name: Get composer cache directory
       id: composer-cache
       run: echo "::set-output name=dir::$(composer config cache-files-dir)"
+
+    - name: Get date part for cache key
+      id: key-date
+      run: echo "::set-output name=date::$(date +'%Y-%m')"
+
     - name: Cache composer dependencies
       uses: actions/cache@v1
       with:
         path: ${{ steps.composer-cache.outputs.dir }}
-        key: ${{ runner.os }}-composer-${{ hashFiles('**/composer.json') }}
-        restore-keys: ${{ runner.os }}-composer-
+        key: ${{ runner.os }}-composer-${{ steps.key-date.outputs.date }}-${{ hashFiles('composer.json') }}
 
     - name: Install packages
       run: |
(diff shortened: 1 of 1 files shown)


othercorey:
I don't know how to review this or know if it works, but I see this output for example in `PHP 7.2 & sqlite`

> Cache not found for input keys: Linux-composer-9acc5a587e3793a4a11b986d4f0f8b53e6550c393541eefd60ecdd48c2d26ed3, Linux-composer-.

> [warning]Cache already exists. Scope: refs/pull/14322/merge, Key: Linux-composer-9acc5a587e3793a4a11b986d4f0f8b53e6550c393541eefd60ecdd48c2d26ed3, Version: (null)

garas:
`Cache not found for input keys` - I think it is normal on first run without no cache existing. Next run should be OK.

`[warning]Cache already exists` - _Coding Standard_ job finished first and has no warning and successfully saved cache file, other jobs finished later and tried to upload with same cache key.

othercorey:
I tried to re-run the tests to see what happened now that the cache was initialized, but github actions doesn't let you do this unless there is an error?

.github/workflows/ci.yml
```diff
@@ -102,6 +112,16 @@ jobs:
...
         sqlcmd -S "(localdb)\MSSQLLocalDB" -Q "create database cakephp;"
 
+    - name: Get composer cache directory
+      id: composer-cache
+      run: echo "::set-output name=dir::$(composer config cache-files-dir)"
+    - name: Cache composer dependencies
+      uses: actions/cache@v1
+      with:
+        path: ${{ steps.composer-cache.outputs.dir }}
+        key: ${{ runner.os }}-composer-${{ hashFiles('**/composer.json') }}
```
markstory: Our composer.json rarely changes. Wouldn't this prevent us from having tests run against new patch releases of our dependencies as they are released?

ADmad: Yeah ideally the `composer.lock` file should be used, but we don't have that committed in the repo.

ADmad: > https://help.github.com/en/actions/configuring-and-managing-workflows/caching-dependencies-to-speed-up-workflows#usage-limits-and-eviction-policy

Unused cache keys are removed after 7 days but I don't see any info about forcibly clearing cache. So we can just take a break of 7 days every month or so to clear the cache :smile:.

markstory: Won't our key always be 'used'? Could we include a date based component to the cache key so it automatically rotates once a week?

.github/workflows/ci.yml
```diff
@@ -44,6 +44,18 @@ jobs:
...
         coverage: pcov
 
+    - name: Get composer cache directory
+      id: composer-cache
+      run: echo "::set-output name=dir::$(composer config cache-files-dir)"
+
+    - name: Cache composer dependencies
+      uses: actions/cache@v1
+      env:
+        KEY_DATE=$(date '+%y-%m')
```
markstory: Why is this one different than the others?

ADmad: It's still WIP. I still haven't figured how to get the required date string is a variable which I can use in the next step :)

ADmad:
@markstory The cache key now contains a date component so that that key gets changed every month.

PR #14324, opened 2020-02-25
Fix null values in Hash::get

jeroenbos22:
Did not work as described. The default value should be returned when the path does not exist according to the docs. Because of the use of isset it will also return the default value when the provided field holds null as value. Since already checked if its an array its better to use array_key_exists.

### DIFF
--- a/lib/Cake/Utility/Hash.php
+++ b/lib/Cake/Utility/Hash.php
@@ -61,7 +61,7 @@ class Hash {
 		}
 
 		foreach ($parts as $key) {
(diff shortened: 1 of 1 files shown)


dereuromark:
Null is usually not considered a value. Thus the expected behavior .
We also cannot change this due to the maintenance mode, and for BC as many apps rely on this current behavior. 
You can make your own class extension and own method for this, though.

ADmad:
Sorry but it's too late in the 2.x life cycle to make such behaviour changes.

PR #14327, opened 2020-02-26
Relay SMTP buffer in exception message.

### LINKED ISSUE #14228: Smtp Transport says timeout when its an error response
This is a (multiple allowed):

* [ ] bug
* [x] enhancement

* CakePHP Version: 3.8.2
* Platform and Target: apache,mysql

### What you did
Send an e-mail with SMTP

### What happened
SMTP Transport class said it timeouted, but it got a response with an error message.
It never stopped when it got the error back and waited till the timeout triggered.
There is no "\r\n" in the error response.

Catch or log the correct error/response and just dont show "SMTP Timeout" which is incorrect.

Additional:
Spent some time to debug this just to find out its not a timeout, its a "Message size too large" Error.
in $response was
```
554 5.2.0 STOREDRV.Submission.Exception:MessageSubmissionExceededException.MapiExceptionMaxSubmissionExceeded; Failed to process message due to a permanent exception with message Cannot save changes made to an item to store.
```

https://github.com/cakephp/cakephp/blob/8b07fc218597156d4f2dacdcc9b8f6bfdee4924f/src/Mailer/Transport/SmtpTransport.php#L468

https://github.com/cakephp/cakephp/blob/8b07fc218597156d4f2dacdcc9b8f6bfdee4924f/src/Mailer/Transport/SmtpTransport.php#L447-L486

jadb:
On Windows, the "\r\n" was not included in the error message as required by the spec which was causing the default "SMTP timeout" message to be passed to SocketException instead of the actual error.

This PR will use the default message only if the response is empty, otherwise passing the raw message.

Closes #14228

### DIFF
--- a/src/Mailer/Transport/SmtpTransport.php
+++ b/src/Mailer/Transport/SmtpTransport.php
@@ -465,7 +465,7 @@ class SmtpTransport extends AbstractTransport
                 $response .= $bytes;
             }
             if (substr($response, -2) !== "\r\n") {
-                throw new SocketException('SMTP timeout.');
+                throw new SocketException($response ?: 'SMTP timeout.');
             }
             $responseLines = explode("\r\n", rtrim($response, "\r\n"));
             $response = end($responseLines);
--- a/src/Mailer/Transport/SmtpTransport.php
+++ b/src/Mailer/Transport/SmtpTransport.php
@@ -464,7 +464,9 @@ class SmtpTransport extends AbstractTransport
                 }
                 $response .= $bytes;
             }
+            // Catch empty or malformed responses.
             if (substr($response, -2) !== "\r\n") {
+                // Use response message or assume operation timed out.
                 throw new SocketException($response ?: 'SMTP timeout.');
             }
             $responseLines = explode("\r\n", rtrim($response, "\r\n"));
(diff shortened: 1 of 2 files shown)


src/Mailer/Transport/SmtpTransport.php
```diff
@@ -465,7 +465,7 @@ protected function _smtpSend(?string $data, $checkCode = '250'): ?string
                 $response .= $bytes;
             }
             if (substr($response, -2) !== "\r\n") {
-                throw new SocketException('SMTP timeout.');
+                throw new SocketException($response ?: 'SMTP timeout.');
```
markstory: Won't `response` always be not null as it is initialized to `''` on line 458.

othercorey: This checks for a falsy or empty string this way.  I think maybe it should be more explict though.

jadb: With an `IF` statement vs. the short way? 

I didn't want to add another nesting level already had `while` and `if` but if you think its better for readability, I can do so.

othercorey: No, I was just thinking how to explain what's happening. A comment would do well to explain why timeout is the default and why response is sent if not empty. Looks fine to me technically.

markstory: My bad I conflated ?? And ?:.

PR #14332, opened 2020-02-27
Adding two examples of how to use the accessibleFields option

nook24:
I added two examples to show how to set `accessibleFields` option of associated entities using `Table::patchEntity()`

Took me quite a while to figure this out! Many thanks to @ADmad

### DIFF
--- a/src/ORM/Marshaller.php
+++ b/src/ORM/Marshaller.php
@@ -160,6 +160,14 @@ class Marshaller
      * ]);
      * ```
      *
+     * ```
+     * $result = $marshaller->one($data, [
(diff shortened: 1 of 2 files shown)


dereuromark:
I wonder if this should go into docs instead.

jadb:
@dereuromark i'd say in docs too (vs. instead). I personally find it comes in handy to have short samples when reading docbloock in IDE.

nook24:
I guess it was there for the whole time and I just didn't saw it: https://book.cakephp.org/4/en/orm/saving-data.html#changing-accessible-fields

PR #14339, opened 2020-03-02
Table::_saveMany() won't rollback if some entity fail to save

victoreassis:
Table::_saveMany() is not working properly, it won't rollback when there's a failed entity and if the fail happens after any save, i.e any entity beside the first, that entity will be saved on the database although the cleanup will remove the given ids and set them as new again.

I changed the testSaveManyFailed to check the count on the table to make sure that the rollback worked. I passed the $failed variable as reference for the Connection::transaction() callback, then it's possible to return false and rollback on the transaction and still have the failed entity for the thrown PersistenceFailedException.

### DIFF
--- a/src/ORM/Table.php
+++ b/src/ORM/Table.php
@@ -2178,13 +2178,16 @@ class Table implements RepositoryInterface, EventListenerInterface, EventDispatc
             }
         };
 
+        $failed = null;
         try {
-            $failed = $this->getConnection()
-                ->transactional(function () use ($entities, $options, &$isNew) {
+            $this->getConnection()
+                ->transactional(function () use ($entities, $options, &$isNew, &$failed) {
                     foreach ($entities as $key => $entity) {
                         $isNew[$key] = $entity->isNew();
                         if ($this->save($entity, $options) === false) {
-                            return $entity;
+                            $failed = $entity;
+
+                            return false;
                         }
                     }
                 });
--- a/src/ORM/Table.php
+++ b/src/ORM/Table.php
@@ -2178,6 +2178,7 @@ class Table implements RepositoryInterface, EventListenerInterface, EventDispatc
             }
         };
 
+        /** @var \Cake\Datasource\EntityInterface|null $failed */
         $failed = null;
         try {
             $this->getConnection()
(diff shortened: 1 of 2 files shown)


src/ORM/Table.php
```diff
@@ -2194,7 +2197,7 @@ protected function _saveMany(iterable $entities, $options = []): iterable
             throw $e;
         }
 
-        if ($failed !== null) {
+        if (isset($failed)) {
```
dereuromark: You can keep this if you move $failed = null before the try block.

tests/TestCase/ORM/TableTest.php
```diff
@@ -2835,6 +2836,7 @@ public function testSaveManyFailed()
         $result = $table->saveMany($entities);
 
         $this->assertFalse($result);
+        $this->assertEquals($expectedCount, $table->find()->count());
```
dereuromark: Same

src/ORM/Table.php
victoreassis: Will do that.

victoreassis:
Still getting a failing test with static analysis. Any suggestion regarding the ```if ($failed !== null)``` placement? Inside the try block should work, right?

dereuromark:
This looks like a bug in phpstan, a false positive as it doesnt understand the reference passing and thus thinks it will never by anything but null.
You could inline annotate the var as `/** @var null|false $failed` maybe?

src/ORM/Table.php
```diff
@@ -2160,7 +2160,10 @@ public function saveManyOrFail(iterable $entities, $options = []): iterable
 
     /**
      * @param \Cake\Datasource\EntityInterface[]|\Cake\Datasource\ResultSetInterface $entities Entities to save.
-     * @param array|\ArrayAccess|\Cake\ORM\SaveOptionsBuilder $options Options used when calling Table::save() for each entity.
+     * @param array|\ArrayAccess|\Cake\ORM\SaveOptionsBuilder                        $options  Options used when
```
ADmad: This needs to be reverted.

victoreassis: Done.

dereuromark:
Do we have to backport this to 3.x as well?

victoreassis:
3.x doesn't have this issue, it has only a saveMany() method that returns false in case of any fail attempting at saving an entity, not an exception like this one.
3.next has it, though.

PR #14344, opened 2020-03-08
Forward port changes from #13848 from 3.9 to 4.1

markstory:
Allow $keyPath to be null which results in a numerically indexed output.


othercorey:
Why are the DSN strings failing to parse on travis?

PR #14345, opened 2020-03-09
Fix command runner mocking

dereuromark:
Or is there a better way to resolve the issue of

> Cake\TestSuite\Stub\MissingConsoleInputException: There are no more input replies available. This is the 1st read operation, only 0 replies were set.
The provided replies are: 
The question asked was: <question>Does this table already contain records? If so, we need to create a nullable field.</question> (no/yes) 

for more complex scenarios than just a simple input (which would be the 2nd arg)?

With the fix this one could set up a custom mock here if needed:
```php
	$this->io = $this->getMockBuilder(ConsoleIo::class)->getMock();
	...
	$this->io->expects($this->once())->method('askChoice')->willReturn('yes');

	$this->exec('add_exposed_field Users');
	$this->assertExitCode(Command::CODE_SUCCESS);
```

### DIFF
--- a/src/TestSuite/ConsoleIntegrationTestTrait.php
+++ b/src/TestSuite/ConsoleIntegrationTestTrait.php
@@ -108,7 +108,7 @@ trait ConsoleIntegrationTestTrait
         if ($this->_in === null) {
             $this->_in = new ConsoleInput($input);
         } elseif ($input) {
-            throw new RuntimeException('You can use $input only if $_in property is null and will be reset.');
+            throw new RuntimeException('You can use `$input` only if `$_in` property is null and will be reset.');
         }
 
         $args = $this->commandStringToArgs("cake $command");
--- a/src/TestSuite/ConsoleIntegrationTestTrait.php
+++ b/src/TestSuite/ConsoleIntegrationTestTrait.php
@@ -29,6 +29,7 @@ use Cake\TestSuite\Constraint\Console\ExitCode;
 use Cake\TestSuite\Stub\ConsoleInput;
 use Cake\TestSuite\Stub\ConsoleOutput;
 use Cake\TestSuite\Stub\MissingConsoleInputException;
+use RuntimeException;
 
 /**
  * A test case class intended to make integration tests of cake console commands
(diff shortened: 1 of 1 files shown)


markstory:
Any reason you can't use the response list parameter of `exec()`? For example your test could be

```php
$this->exec('add_exposed_field Users', ['y'])
```

src/TestSuite/ConsoleIntegrationTestTrait.php
```diff
@@ -90,15 +91,25 @@ trait ConsoleIntegrationTestTrait
...
+            $this->_out = new ConsoleOutput();
+        }
+        if ($this->_err === null) {
+            $this->_err = new ConsoleOutput();
+        }
+        if ($this->_in === null) {
+            $this->_in = new ConsoleInput($input);
+        } elseif ($input) {
+            throw new RuntimeException('You can use $input only if $_in property is null and will be reset.');
+        }
```
markstory: Not sure why all of this is needed if you can use the `$input` array to stub out input requests. Are there other situations you need to use mocks?

dereuromark: You can use the ->once() ->any() and ->..() calls depending on your needs.
Right now it is always the once exact order match.

But I can also understand if that is out of scope for this.

I just don't see any disadvantage of this conditional setup, as this is fully BC and side-by-side working to the existing way of doing it.

src/TestSuite/ConsoleIntegrationTestTrait.php
```diff
@@ -90,15 +91,25 @@ trait ConsoleIntegrationTestTrait
...
+        if ($this->_out === null) {
+            $this->_out = new ConsoleOutput();
+        }
+        if ($this->_err === null) {
+            $this->_err = new ConsoleOutput();
+        }
+        if ($this->_in === null) {
+            $this->_in = new ConsoleInput($input);
+        } elseif ($input) {
+            throw new RuntimeException('You can use $input only if $_in property is null and will be reset.');
```
othercorey:
```suggestion
            throw new RuntimeException('You can use `$input` only if `$_in` property is null and will be reset.');
```
Do we quote variable names?

src/TestSuite/ConsoleIntegrationTestTrait.php
markstory: Ok, I was just checking if there was a gap in how input stubbing worked, but it sounds like you prefer to use mocks for certain kinds of tests.

PR #14351, opened 2020-03-11
Re-enable retrieval of model alias per controller.

dereuromark:
When upgrading a plugin one faces the challenge of accessing the now protected class property.
In this case $modelClass and the resulting alias per controller

```php
// Inside component
$modelName = $this->Controller->modelClass;
		
list(, $modelName) = pluginSplit($modelName);
$this->setConfig('modelName', $modelName);
if (!$this->Controller->{$modelName}->behaviors()->has('...')) {
	$this->Controller->{$modelName}->behaviors()->load('...', $this->_config);
}
```

With this suggested change this would be able again
```php
$modelName = $this->Controller->getModelAlias();

$this->setConfig('modelName', $modelName);
if (!$this->Controller->{$modelName}->behaviors()->has('...')) {
	$this->Controller->{$modelName}->behaviors()->load('...', $this->_config);
}
```
Or other generic operations done by a generic component assisting:
```php
$modelName = $this->Controller->getModelAlias();
if ($this->Controller->{$modelName}->do()) {
	...
}
```

Or are there better alternatives?

Force reading the field with reflection is an option but sounds like a less clean approach.

### DIFF
--- a/src/Datasource/ModelAwareTrait.php
+++ b/src/Datasource/ModelAwareTrait.php
@@ -129,10 +129,13 @@ trait ModelAwareTrait
         if ($modelType === null) {
             $modelType = $this->getModelType();
         }
-
         if ($modelClass === null) {
             $modelClass = $this->modelClass;
         }
+        if (empty($modelClass)) {
+            throw new UnexpectedValueException('Default modelClass is empty');
+        }
+
         $alias = $this->getModelAlias($modelClass, $modelType);
 
         if (isset($this->{$alias})) {
--- a/src/Datasource/ModelAwareTrait.php
+++ b/src/Datasource/ModelAwareTrait.php
@@ -73,8 +73,8 @@ trait ModelAwareTrait
     }
 
     /**
-     * @param string|null $modelClass
-     * @param string|null $modelType
+     * @param string|null $modelClass Name of model class to load. Defaults to $this->modelClass.
+     * @param string|null $modelType The type of repository to load. Defaults to the modelType() value.
      * @throws \UnexpectedValueException If $modelClass argument is not provided
      *   and ModelAwareTrait::$modelClass property value is empty.
      * @return string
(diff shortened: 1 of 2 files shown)


jadb (approved):

ADmad:
This those target `4.next`. We no longer add new feature or change public API (even if it's BC) in patch releases.

src/Datasource/ModelAwareTrait.php
```diff
@@ -73,52 +73,81 @@ protected function _setModelClass(string $name): void
...
-     *  The name can be an alias like `'Post'` or FQCN like `App\Model\Table\PostsTable::class`.
      * @param string|null $modelType The type of repository to load. Defaults to the modelType() value.
-     * @return \Cake\Datasource\RepositoryInterface The model instance created.
-     * @throws \Cake\Datasource\Exception\MissingModelException If the model class cannot be found.
      * @throws \UnexpectedValueException If $modelClass argument is not provided
      *   and ModelAwareTrait::$modelClass property value is empty.
+     * @return string
      */
-    public function loadModel(?string $modelClass = null, ?string $modelType = null): RepositoryInterface
+    public function getModelAlias(?string $modelClass = null, ?string $modelType = null): string
```
dereuromark: Or should this be named `getModelName()` ?

src/Datasource/ModelAwareTrait.php
```diff
@@ -73,52 +73,81 @@ protected function _setModelClass(string $name): void
...
 
     /**
-     * Loads and constructs repository objects required by this object
-     *
-     * Typically used to load ORM Table objects as required. Can
-     * also be used to load other types of repository objects your application uses.
-     *
-     * If a repository provider does not return an object a MissingModelException will
-     * be thrown.
-     *
```
markstory: Shouldn't this method have a description?

markstory:
Closing in favour of #14356

PR #14353, opened 2020-03-11
Fix argument type declaration

### LINKED ISSUE #14352: Declaration of Table incompatible with RepositoryInterface
This is a (multiple allowed):

* [x] bug
* [ ] enhancement

* CakePHP Version: 4.0.4
* Platform and Target: Docker: php:7.2-apache-stretch. I'm using `cakephp/orm` as standalone (see: https://github.com/lorenzo/cake-orm-standalone-example).

### What you did
I tried to get a Table from the Table locator like this:
```php
$test = \Cake\ORM\TableRegistry::getTableLocator()->get('TableName');
```

### What happened
A fatal error is thrown. Full error:
```
Fatal error: Declaration of Cake\ORM\Table::get($primaryKey, $options = Array): Cake\Datasource\EntityInterface must be compatible with Cake\Datasource\RepositoryInterface::get($primaryKey, array $options = Array): Cake\Datasource\EntityInterface
```

I expected that no fatal exception regarding the compatibility of the original `Cake\Orm\Table` implementation is thrown.

>P.S. Remember, an issue is not the place to ask questions. You can use [Stack Overflow](https://stackoverflow.com/questions/tagged/cakephp)
for that or join the #cakephp channel on irc.freenode.net, where we will be more
than happy to help answer your questions.
Before you open an issue, please check if a similar issue already exists or has been closed before.

I tried to find this specific issue in your tracker and on google but had no success. I checked the Code and it actually seems to be wrong. So i'm opting to open an issue instead of asking on Stack Overflow.

The only thing missing in the `get`-Declaration in `Table` is the Type `array` in front of the parameter `options`.

I can provide a pull-request if desired.

jadb:
Closes #14352

### DIFF
--- a/src/ORM/Table.php
+++ b/src/ORM/Table.php
@@ -1463,7 +1463,7 @@ class Table implements RepositoryInterface, EventListenerInterface, EventDispatc
      *      incorrect number of elements.
      * @psalm-suppress InvalidReturnType
      */
-    public function get($primaryKey, $options = []): EntityInterface
+    public function get($primaryKey, array $options = []): EntityInterface
     {
         $key = (array)$this->getPrimaryKey();
         $alias = $this->getAlias();


dereuromark (approved):

jadb:
In light of these fixes, I started quickly looking at the other methods and it looks like there are other possible conflicts?

```
    public function sql(?ValueBinder $generator = null): string
```

vs.

```
    public function sql(ValueBinder $generator): string;
```

Do we want to update the interface for that? Wouldn't that be a BC?

dereuromark:
For BC we should go the other way where possible. Use interface as truth and only in next minor or major we can discuss BC breaking adjustments.

jadb:
Ok, so upon further investigation, I noticed @ADmad changed the `Closure` to a `callable` in the `Query::traverse()` to widen possibilities but the interface wasn't updated accordingly.

Which makes me wonder:

- should the interface be updated in that case?
- how come those error took so long to popup?

src/Database/Query.php
```diff
@@ -327,7 +327,7 @@ public function sql(?ValueBinder $generator = null): string
      * @param callable $visitor A function or callable to be executed for each part
      * @return $this
      */
-    public function traverse($visitor)
+    public function traverse(Closure $visitor)
```
ADmad: This is intentional parameter type widening.

dereuromark: Did this throw a warning too in PHP74? Otherwise we want to keep this probably.

jadb: I know, mentioned in my last comment. But the interface wasn't updated accordingly and doing it now wouldn't be BC?

ADmad:
> should the interface be updated in that case?

No it shouldn't, you can read the comment on related PR https://github.com/cakephp/cakephp/pull/13620#discussion_r321966175 and go down the rabbit hole.

> how come those error took so long to popup?

There was no error because method parameter type is allowed in PHP 7.2.

Issue #14326, opened 2020-02-26
Cannot supress PHPUnit ReflectionType deprecation warnings on PHP 7.4


andres-ml:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.8.10
* Platform and Target: php 7.4, both with phpunit 5.7 and 6.5

### What you did
Use mocks within my Test cases with php 7.4

### What happened
The following deprecation error appears:
`Deprecated Error: Function ReflectionType::__toString() is deprecated in [.../vendor/phpunit/phpunit-mock-objects/src/Generator.php, line 1090]`

### What you expected to happen
No deprecation error to appear.

I thought #14061 solved this problem, but it's still showing the deprecation error for me. From what I can gather, even though the call to `getMock()` is wrapped to disable E_DEPRECATED errors, the error is still being caught by the handler in `src/Error/BaseErrorHandler::handleError` and ends up being displayed regardless of the current `error_reporting()` level. Note that this does not happen if your app's errorLevel is configured to ignore E_DEPRECATED errors as a whole, since the handler is not registered for that level.

A simple (but kind of breaking) fix would be replacing the following in BaseErrorHandler:
```php
if (error_reporting() === 0) {
    return false;
}
```
for

```php
if (!(error_reporting() & $code)) {
    return false;
}
```

Note: this issue is not limited to the ReflectionType deprecation but also to other deprecations; for example:
```php
public function testErrorReporting()
{
    error_reporting(E_ALL ^ E_DEPRECATED);
    is_real(3);
}
```
Yields a similar error:
```
PHPUnit 6.5.14 by Sebastian Bergmann and contributors.

Deprecated Error: Function is_real() is deprecated in [...]
```

markstory:
Have you tried disabling deprecation warnings? We run the CakePHP suite on 7.4 and use mocks extensively,  so I am not sure what causes the errors to show up for you and not us.

andres-ml:
Yes; as I commented, configuring the app's `errorLevel` to ignore all deprecation warnings makes it so it doesn't show up.
The problem is I want to keep those warnings (so I can fix them), just not the ReflectionType one which I have no control over and keeps popping all over my tests. Besides, if disabling them as a whole is the only solution, what was the point of introducing [this class](https://github.com/cakephp/cakephp/blob/3.x/src/TestSuite/MockBuilder.php)?

dereuromark:
[...]

PR #14355, opened 2020-03-11
Use error handler to supress `ReflectionType::__toString()` deprecation warnings on PHP 7.4

garas:
Fixes https://github.com/cakephp/cakephp/issues/14326. As reported in the issue, `error_reporting()` level should be handled by error handler.

PHPUnit error handler [checks](https://github.com/sebastianbergmann/phpunit/blob/6.5.14/src/Util/ErrorHandler.php#L47-L49) current error level, so for simple tests previous solution was OK.

But in integration tests, CakePHP error handler is [registered](https://github.com/cakephp/cakephp/blob/3.8.10/src/Error/BaseErrorHandler.php#L78-L79), but it [doesn't check](https://github.com/cakephp/cakephp/blob/3.8.10/src/Error/BaseErrorHandler.php#L78-L79) active error level. Full fix would be to check active error level like PHPUnit error handler does, but it is not completely BC safe.

So instead of temporary changing error level, temporary set custom error handler to catch and skip only specific `ReflectionType::__toString()` deprecation warning. I've tested it with fresh app skeleton and there is now no warning.

### DIFF
--- a/src/TestSuite/MockBuilder.php
+++ b/src/TestSuite/MockBuilder.php
@@ -30,12 +30,28 @@ class MockBuilder extends BaseMockBuilder
      */
     public function getMock()
     {
-        $errorLevel = error_reporting();
-        error_reporting(E_ALL ^ E_DEPRECATED);
+        static::setSupressedErrorHandler();
+
         try {
             return parent::getMock();
         } finally {
-            error_reporting($errorLevel);
+            restore_error_handler();
         }
+    }
+
+    /**
+     * Set error handler to supress `ReflectionType::__toString()` deprecation warning
+     *
+     * @return void
+     */
+    public static function setSupressedErrorHandler()
+    {
+        $previousHandler = set_error_handler(function ($code, $description, $file = null, $line = null, $context = null) use (&$previousHandler) {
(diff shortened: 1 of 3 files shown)


othercorey:
Is it possible to make internal cakephp and integration tests work the same? It's surprising that the error handling is different.

markstory:
> Full fix would be to check active error level like PHPUnit error handler does, but it is not completely BC safe.

What are the compatibility risks with adding an error level check against the current value in Configure?

garas:
I think it would be safest to use `Configure` value only for initial error level of `error_reporting()`, but  `set_error_handler()` should catch all levels.

https://github.com/cakephp/cakephp/blob/a5b75c3a9d0c3ed99a2220d9114571d383acf214/src/Error/BaseErrorHandler.php#L78-L79

And error handler then should filter by `error_reporting() & $code`.

https://github.com/cakephp/cakephp/blob/a5b75c3a9d0c3ed99a2220d9114571d383acf214/src/Error/BaseErrorHandler.php#L132-L134

If app/plugin doesn't use any run time error level changes, it should be completely compatible. It is most likely error level changes only exist in unit tests only.

I think fix could be done in next minor at 3.9 and 4.1.

PR #14356, opened 2020-03-11
4.next component model alias

dereuromark:
Alternative to https://github.com/cakephp/cakephp/pull/14351 for 4.next

### DIFF
--- a/src/Datasource/ModelAwareTrait.php
+++ b/src/Datasource/ModelAwareTrait.php
@@ -73,6 +73,41 @@ trait ModelAwareTrait
     }
 
     /**
+     * @param string|null $modelClass Name of model class. Defaults to $this->modelClass.
+     * @param string|null $modelType The type of repository. Defaults to the getModelType() value.
+     * @throws \UnexpectedValueException If $modelClass argument is not provided
+     *   and ModelAwareTrait::$modelClass property value is empty.
+     * @return string
+     */
+    public function getModelAlias(?string $modelClass = null, ?string $modelType = null): string
+    {
+        if ($modelClass === null) {
+            $modelClass = $this->modelClass;
+        }
+
+        if (empty($modelClass)) {
+            throw new UnexpectedValueException('Default modelClass is empty');
+        }
+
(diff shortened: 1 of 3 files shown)


src/Datasource/ModelAwareTrait.php
```diff
@@ -72,6 +72,41 @@ protected function _setModelClass(string $name): void
         }
     }
 
+    /**
+     * @param string|null $modelClass Name of model class. Defaults to $this->modelClass.
```
markstory:
```suggestion
    /**
     * Get the current model alias based on modelClass property.
     *
     * @param string|null $modelClass Name of model class. Defaults to $this->modelClass.
```

ADmad:
Is the new method really necessary? Currently one can use `$controller->loadModel()->getAlias()` to get the default model's alias.

dereuromark:
That could also be enough.

ADmad:
Then it would be better to avoid increasing the API surface.

dereuromark:
This also reduces side efects inside methods as it reduces complexity and delegates those into smaller chunks.
But I can see that the benefit here is really small - as such I would be fine with the status quo.

dereuromark:
There is one problem now though

The core states:
```php
    /**
     * ...
     *
     * Use empty string to not use auto-loading on this object. Null auto-detects based on
     * controller name.
     *
     * @var string|null
     */
    protected $modelClass;
```

And then tools like bake are forced to use `loadModel()->...` as they cannot know beforehand that here no model is supposed to be used.

Causing errors like

> [UnexpectedValueException] Default modelClass is empty in /.../vendor/cakephp/cakephp/src/Datasource/ModelAwareTrait.php on line 98

Sure, one could use try catch here, but a dedicated method might be cleaner before any loadModel class is done.

ADmad:
> And then tools like bake are forced to use loadModel()->... as they cannot know beforehand that here no model is supposed to be used.

Can you give a specific example/use case? Baking works just fine now

dereuromark:
Actually, it doesnt. I can make a fix

PR #14357, opened 2020-03-11
Test to show a bug when using a custom event implementation

burzum:
This test shows a problem with the type hint of _callListener() in the EventManager when using a custom event implementation. The current implementation is this:

> protected function _callListener(callable $listener, **Event** $event)

and by using Event and not EventInterface, it will fail when using a custom event class. It needs to be changed to this:

> protected function _callListener(callable $listener, **EventInterface** $event)

Also there is yet another bug:

```php
EventManager::instance()->on($event->getName(), [$this, $handlerMethod]);
```

Will cause a type error on the callable type hint. The line above is at least according to the documentation valid code. https://book.cakephp.org/3/en/core-libraries/events.html#interacting-with-existing-listeners

So to fix the we need to remove the callable and check for array or callable and throw an invalid argument exception if none of both matches and change Event to EventInterface.

### DIFF
--- a/tests/TestCase/Event/EventManagerTest.php
+++ b/tests/TestCase/Event/EventManagerTest.php
@@ -15,10 +15,94 @@
 namespace Cake\Test\TestCase\Event;
 
 use Cake\Event\Event;
+use Cake\Event\EventInterface;
 use Cake\Event\EventList;
 use Cake\Event\EventListenerInterface;
 use Cake\Event\EventManager;
 use Cake\TestSuite\TestCase;
+
+/**
+ * TestEvent
+ */
+class TestEvent implements EventInterface
+{
+    /**
+     * @var string
+     */
+    protected $name;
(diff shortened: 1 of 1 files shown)


dereuromark:
What about

    ConditionDecorator::_evaluateCondition($condition, Event $event)

and

    SubjectFilterDecorator::canTrigger(Event $event)

and alike?
Do they also need changing?

burzum:
@dereuromark I think all places that still use Event as type need to be changed to the interface.

dereuromark:
Which happened in 4.x
We need to make sure we prioritize in 3.x what is required as bug, and what can optionally be done with a certain BC promise in mind
Probably has to go then in 3.next either way.

dereuromark:
Indeed, the tests are failing.
Can you add your recommended changes to the src code here as well now?

PR #14361, opened 2020-03-13
Allow changing log date format

mtak3:
# Allow changing log date format

Allow changing log date format for LogEngine  provided by CakePHP.
no change default date format. (compatible changes)

I understand that it is possible to change the date format using monolog.
https://book.cakephp.org/4/en/core-libraries/logging.html#using-monolog

But I'm satisfied with the log engine provided by CakePHP, except that I can't change the date format.

Because the Application TimeZone and the OS TimeZone may be different, it will be confusing if the date has no TimeZone information.

### DIFF
--- a/src/Log/Engine/BaseLog.php
+++ b/src/Log/Engine/BaseLog.php
@@ -37,6 +37,7 @@ abstract class BaseLog extends AbstractLogger
     protected $_defaultConfig = [
         'levels' => [],
         'scopes' => [],
+        'dateFormat' => 'Y-m-d H:i:s',
     ];
 
     /**
--- a/src/Log/Engine/BaseLog.php
+++ b/src/Log/Engine/BaseLog.php
@@ -161,11 +161,13 @@ abstract class BaseLog extends AbstractLogger
     }
 
     /**
-     * Get the date.
+     * Returns date formatted according to given `dateFormat` option format.
+     *
+     * This function affects `FileLog` or` ConsoleLog` datetime information format.
      *
      * @return string
      */
-    protected function _date()
+    protected function _getFormattedDate()
     {
         return date($this->_config['dateFormat']);
     }
--- a/src/Log/Engine/BaseLog.php
+++ b/src/Log/Engine/BaseLog.php
@@ -159,4 +159,14 @@ abstract class BaseLog extends AbstractLogger
 
         return str_replace(array_keys($replacements), $replacements, $message);
     }
+
+    /**
+     * Get the date.
+     *
+     * @return string
+     */
+    protected function _date()
+    {
+        return date($this->_config['date_format']);
+    }
 }
--- a/src/Log/Engine/BaseLog.php
+++ b/src/Log/Engine/BaseLog.php
@@ -167,6 +167,6 @@ abstract class BaseLog extends AbstractLogger
      */
     protected function _date()
     {
-        return date($this->_config['date_format']);
+        return date($this->_config['dateFormat']);
     }
 }
--- a/src/Log/Engine/ConsoleLog.php
+++ b/src/Log/Engine/ConsoleLog.php
@@ -34,6 +34,7 @@ class ConsoleLog extends BaseLog
         'levels' => null,
         'scopes' => [],
         'outputAs' => null,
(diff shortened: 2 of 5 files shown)


saeideng:
by default it uses current local time , `date_default_timezone_get()`
it is `UTC` by default right now
https://github.com/cakephp/app/blob/a1ecb83e73068880874852dbac6d3f18cb556ec7/config/app.php#L54

>Because the Application TimeZone and the OS TimeZone may be different, it will be confusing if

how you want to handle it when probably your local timezone and server was different ?

I think best way is keeping as default timezone

or even maybe
```php
date('Y-m-d H:i:s e'); // e = Timezone identifier // echos `2020-03-13 09:20:59 UTC`
```

mtak3:
> how you want to handle it when probably your local timezone and server was different ?
>
> I think best way is keeping as default timezone

using Docker (with Cloud)
- application
- Docker container
- Host OS
- etc ...

I need to adjust the TimeZone of many layers and it is very hard.
Various things can be solved if the date has TimeZone information.

> or even maybe printing time zone

This fix is to do that : )

ADmad:
I changed the PR base to `4.next` and new features need to target next minor release.

src/Log/Engine/ConsoleLog.php
```diff
@@ -34,6 +34,7 @@ class ConsoleLog extends BaseLog
         'levels' => null,
         'scopes' => [],
         'outputAs' => null,
+        'date_format' => 'Y-m-d H:i:s',
```
markstory: Most of our options are camelBacked so this would need to be `dateFormat`.

src/Log/Engine/BaseLog.php
```diff
@@ -159,4 +159,14 @@ protected function _format(string $message, array $context = []): string
 
         return str_replace(array_keys($replacements), $replacements, $message);
     }
+
+    /**
+     * Get the date.
+     *
+     * @return string
+     */
+    protected function _date()
```
othercorey: I would prefer a better name and a more complete description.

ndm2: It should especially note that one has to implement the required format option, or the option should be moved into the base log.

tests/TestCase/Log/Engine/ConsoleLogTest.php
```diff
@@ -76,4 +77,22 @@ public function testDefaultOutputAs()
...
+    {
+        $filename = tempnam(sys_get_temp_dir(), 'cake_log_test');
+        $log = new ConsoleLog([
+            'stream' => $filename,
+            'dateFormat' => 'c',
+        ]);
+        $log->log('error', 'oh noes');
+        $fh = fopen($filename, 'r');
+        $line = fgets($fh);
+        $this->assertRegExp('/^2[0-9]{3}-[0-9]+-[0-9]+T[0-9]+:[0-9]+:[0-9]+\+\d{2}:\d{2} Error: oh noes/', $line);
```
markstory: This test is failing in appveyor. Perhaps the `^` anchor should be removed?

src/Log/Engine/BaseLog.php
```diff
@@ -159,4 +159,16 @@ protected function _format(string $message, array $context = []): string
...
+    /**
+     * Returns date formatted according to given `dateFormat` option format.
+     *
+     * This function affects `FileLog` or` ConsoleLog` datetime information format.
+     *
+     * @return string
+     */
+    protected function _getFormattedDate()
+    {
+        return date($this->_config['dateFormat']);
```
ADmad, line 172: `dateFormat` needs to be added to `BaseLog::$_defaultConfig` too.

src/Log/Engine/ConsoleLog.php
```diff
@@ -52,6 +53,7 @@ class ConsoleLog extends BaseLog
      * - `scopes` string or array, scopes the engine is interested in
      * - `stream` the path to save logs on.
      * - `outputAs` integer or ConsoleOutput::[RAW|PLAIN|COLOR]
+     * - `dateFormat` php date() format.
```
ADmad, line 56:
```suggestion
     * - `dateFormat` PHP date() format.
```

PR #14364, opened 2020-03-17
4.1 - Return previous app namespace

othercorey:
Easy access to previous app namespace without having to directly query the configure that's being wrapped here.

### DIFF
--- a/src/TestSuite/TestCase.php
+++ b/src/TestSuite/TestCase.php
@@ -869,11 +869,14 @@ abstract class TestCase extends BaseTestCase
      * Set the app namespace
      *
      * @param string $appNamespace The app namespace, defaults to "TestApp".
-     * @return void
+     * @return string|null The previous app namespace or null if not set.
      */
-    public static function setAppNamespace(string $appNamespace = 'TestApp'): void
+    public static function setAppNamespace(string $appNamespace = 'TestApp'): ?string
     {
+        $previous = Configure::read('App.namespace');
         Configure::write('App.namespace', $appNamespace);
+
+        return $previous;
     }
 
     /**
--- a/tests/TestCase/Http/MiddlewareQueueTest.php
+++ b/tests/TestCase/Http/MiddlewareQueueTest.php
@@ -16,7 +16,6 @@ declare(strict_types=1);
  */
 namespace Cake\Test\TestCase\Http;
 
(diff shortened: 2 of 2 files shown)


jadb:
I like the addition it brings but I feel it's against common logic around _setters_? I might be wrong here but I think they either don't return anything or return the updated object / value. 

Maybe a new method (i.e. `changeAppNamespace()`) that wraps both the `Configure::read()` and `setAppNamespace()` instead?

ADmad:
@jadb This method is only used in tests so I think it's fine to make it return previous name. PHP does have similar returns, for e.g. `set_exception_handler()` also returns previous handler name.

jadb:
@ADmad I have to agree on context reasoning 😄

src/TestSuite/TestCase.php
```diff
@@ -869,11 +869,14 @@ protected function _getTableClassName(string $alias, array $options): string
...
-     * @return void
+     * @return string|null The previous app namespace or null if not set.
      */
-    public static function setAppNamespace(string $appNamespace = 'TestApp'): void
+    public static function setAppNamespace(string $appNamespace = 'TestApp'): ?string
     {
+        $previous = Configure::read('App.namespace');
         Configure::write('App.namespace', $appNamespace);
+
+        return $previous;
```
markstory: How do you see this being used in our test suites or in an app's?

othercorey: I added an example conversion. There are a few cases for this. The reason I added it as return is this wraps the setter but you have to manually read the value if you want to save it which makes this wrapper pointless.

We have tests where the app namespace is set and never restored -- probably because nothing implies you can/should save it.

We have cases where the configure is manually queried to save when this wrapper is meant to hide that.

We have unit tests where the app namespace cannot be 'TestApp' which can happen accidentally depending on the order of execution because we don't have a clean save/restore pattern.

othercorey:
> I like the addition it brings but I feel it's against common logic around _setters_? I might be wrong here but I think they either don't return anything or return the updated object / value.
> 
> Maybe a new method (i.e. `changeAppNamespace()`) that wraps both the `Configure::read()` and `setAppNamespace()` instead?

So, the reason I made this setter special is the use case is always a temporary override. You never need to query the value through a wrapper.

PR #14366, opened 2020-03-17
Fixed type for formatResults callback

othercorey:
Fixes phpstan error.


othercorey:
Now we have a situation where the return type for `CollectionInterface::listNested()` is `CollectionInterface` but it actually returns `TreeIterator` where `printer()` is defined.

PR #14369, opened 2020-03-18
Fix docblock and revert missing iterable annotation.

dereuromark:
Can we please have this useful annotation back?

I had now several times the requirement of unnessarily needing to inline annotate the basic looping over the collection:
```php
/** @var \Cake\ORM\Association[] $associations */
$associations = $modelInstance->associations();
foreach ($associations as $association) {
	$relationType = $association->type();
```

With the proper docblock typehinting restored for the iterable behavior, this is then clear by itself.
And it is valid as the primary type still dictates this as object.

### DIFF
--- a/src/ORM/Table.php
+++ b/src/ORM/Table.php
@@ -950,7 +950,7 @@ class Table implements RepositoryInterface, EventListenerInterface, EventDispatc
     /**
      * Get the associations collection for this table.
      *
-     * @return \Cake\ORM\AssociationCollection The collection of association objects.
+     * @return \Cake\ORM\AssociationCollection|\Cake\ORM\Association[] The collection of association objects.
      */
     public function associations(): AssociationCollection
     {


dereuromark:
psalm doesnt understand this or at least warns here, which needs to be ignored.

ADmad:
> psalm doesnt understand this or at least warns here, which needs to be ignored.

No, you addition is simply wrong. It's a hack to please your IDE.

dereuromark:
The hack here is the code, quite the codesmell to hide a specific class iterable behind a non speaking collection class.
This here makes it not only visible and clear to the IDE, but also all other stakeholders.

ADmad:
There are proper notations to specify the type of individual item of an iterable but they are not used because again your IDE phpstorm doesn't not understand them (and will probably complain).

dereuromark:
Then they are not proper either :)
Might be worth adding both then for now, complementing either other.

ADmad:
https://blog.jetbrains.com/phpstorm/2016/06/phpstorm-2016-2-eap-162-844/  shows the proper way of specifying types for `IteratorAggregate` implementations (which `AssociationCollection` is).

Closing in favor of #14371,

PR #14370, opened 2020-03-18
$when = true by default for Validator::notEmpty*() methods

andrii-pukhalevych:
Its much logical to use $when = true by default for Validator::notEmpty*() methods, and allow to pass false to disable rule

### DIFF
--- a/src/Validation/Validator.php
+++ b/src/Validation/Validator.php
@@ -864,7 +864,7 @@ class Validator implements ArrayAccess, IteratorAggregate, Countable
      * @param string $field The name of the field.
      * @param string|null $message The message to show if the field is empty.
      * @param bool|string|callable $when Indicates when the field is not allowed
-     *   to be empty. Valid values are false (never), 'create', 'update'. If a
+     *   to be empty. Valid values are true, false, 'create', 'update'. If a
      *   callable is passed then the field will be required to be not empty when
      *   the callback returns true.
      * @return $this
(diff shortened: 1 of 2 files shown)


ADmad:
We can't flip the default value in a minor release and break backwards compatibility. This change (if approved in priciple) would have to wait till 5.0.

saeideng:
this could not be change even in 5.0 unless we remove it and use the string/const
`always` / 'create' / `update`

markstory:
What would breaking compatibility on these methods improve? What usage gets better with this change?

andrii-pukhalevych:
It will be the same logic for $when as for allowEmpty* methods:
true: rule is always enabled, false: rule is always disabled

ADmad:
`allowEmpty*` and `notEmpty*` are opposites, so it's logical that their `$when` argument have opposite default value.

andrii-pukhalevych:
Why if `$when` callable used for `notEmpty*`, it must return `true` to enable rule?
true sounds like rule is always enabled. 
And now its false by default, sounds like rule is disabled, and setting true even is not supported.

markstory:
I'm not a fan of this change. It is not something we can easily add a deprecation warning for, and is a minor improvement with potentially large consequences (invalid data getting by validation rules).

dereuromark:
We will have to revisit this then for 5.0. I put it into the roadmap https://github.com/cakephp/cakephp/wiki/5.x-Ideas#cleanup

PR #14371, opened 2020-03-18
Fix typehint for iterables.

ADmad:
Improve docblock types to help static analyzers and IDEs detect the type of item returned by iterables.

https://blog.jetbrains.com/phpstorm/2016/06/phpstorm-2016-2-eap-162-844/

### DIFF
--- a/src/Console/CommandCollection.php
+++ b/src/Console/CommandCollection.php
@@ -20,6 +20,7 @@ use ArrayIterator;
 use Countable;
 use InvalidArgumentException;
 use IteratorAggregate;
+use Traversable;
 
 /**
  * Collection for Commands.
@@ -34,6 +35,7 @@ class CommandCollection implements IteratorAggregate, Countable
      * Command list
      *
      * @var array
+     * @psalm-var (\Cake\Console\Shell|\Cake\Console\CommandInterface|class-string)[]
      */
     protected $commands = [];
 
@@ -60,7 +62,6 @@ class CommandCollection implements IteratorAggregate, Countable
      */
     public function add(string $name, $command)
     {
-        /** @psalm-suppress DeprecatedClass */
         if (!is_subclass_of($command, Shell::class) && !is_subclass_of($command, CommandInterface::class)) {
             $class = is_string($command) ? $command : get_class($command);
             throw new InvalidArgumentException(sprintf(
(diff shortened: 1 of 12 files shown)


dereuromark:
How does this solve the inline annotation needed?
```php
/** @var \Cake\ORM\Association[] $associations */
$associations = $modelInstance->associations();
foreach ($associations as $association) {
	$relationType = $association->type();
```

othercorey:
Wonder when we can officially use the class-string type I added to TypeResolver.

src/Console/CommandCollection.php
```diff
@@ -34,6 +35,7 @@ class CommandCollection implements IteratorAggregate, Countable
      * Command list
      *
      * @var array
```
othercorey:
```suggestion
     * @var (\Cake\Console\Shell|\Cake\Console\CommandInterface|string)[]
```
Can we at least do this?

src/Core/ObjectRegistry.php
```diff
@@ -375,10 +376,10 @@ public function unload(string $objectName)
     /**
      * Returns an array iterator.
      *
-     * @return \ArrayIterator
-     * @psalm-return \ArrayIterator<string, TObject>
+     * @return \Traversable
```
othercorey:
```suggestion
     * @return \Traversable<string, TObject>
```
TypeResolver should able to handle this. It generates a Collection type.

ADmad:
> How does this solve the inline annotation needed?

Did you read type hinting section on https://blog.jetbrains.com/phpstorm/2016/06/phpstorm-2016-2-eap-162-844/?  It's shows an example for exactly the usage you show.

ADmad:
> Wonder when we can officially use the class-string type I added to TypeResolver.

Probably not until phpstorm starts supporting it, same for the other changes you have suggested. Everyone's beloved IDE is what's preventing us from using better annotation types.

othercorey:
> Probably not until phpstorm starts supporting it, same for the other changes you have suggested. Everyone's beloved IDE is what's preventing us from using better annotation types.

phpstorm supports compound array types at least, right?

ADmad:
> phpstorm supports compound array types at least, right?

IDK, if someone confirms I can make the changes.

dereuromark:
I can confirm that it works in PHPStorm 2019.3.4 etc
@ADmad thank you for looking into this.

dereuromark:
@othercorey If you want me to test something else, please make a branch and give that to me.

PR #14373, opened 2020-03-19
Allow using "Unassigned" HTTP status codes within 1xx - 5xx range.

### LINKED ISSUE #14372: Setting response code out of 1xx-5xx range (rest/restful api)
This is a (multiple allowed):

* [ ] bug
* [x] enhancement

* CakePHP Version: 3.8.10/4.x
* Platform and Target: nginx php7.3+

Currently, to be able to send custom HTTP Status Code for REST API use-case, i've had to patch the Http/Response.php (diff against 3.8.10 version) like this
```diff
> git diff
diff --git a/src/Http/Response.php b/src/Http/Response.php
index 54e68c6..488c0dc 100644
--- a/src/Http/Response.php
+++ b/src/Http/Response.php
@@ -992,16 +992,9 @@ class Response implements ResponseInterface
      */
     protected function _setStatus($code, $reasonPhrase = '')
     {
-        if (!isset($this->_statusCodes[$code])) {
-            throw new InvalidArgumentException(sprintf(
-                'Invalid status code: %s. Use a valid HTTP status code in range 1xx - 5xx.',
-                $code
-            ));
-        }
-
         $this->_status = $code;
         if (empty($reasonPhrase)) {
-            $reasonPhrase = $this->_statusCodes[$code];
+            $reasonPhrase = isset($this->_statusCodes[$code]) ? $this->_statusCodes[$code] : 'API Response';
         }
         $this->_reasonPhrase = $reasonPhrase;
```

Incorrect and awful solution (second part could be solved by changing default reasonPhrase or always passing some string), yes, but there is currently no supported framework way

I came up with checklist, that I'd like to see implemented, and I can contribute directly via PRs to do that, but I'd like to know your opinion on this subject first.

For CakePHP 3.8.10
 - [ ] There should be non-deprecated way to set statusCodes (method `Cake\Http\Response::httpCodes($code = null)` is deprecated and to be removed in 4.0.0)
 - [ ] Method `Response::httpCodes($code = null)` accepts any http status-code definition, if status code number is in range **100<=statusCode<=999** this filter should be optional and it does not match next issue
 - [ ] `InvalidArgumentException` in `Response::_setStatus($code, $reasonPhrase = '')` message is misleading, since the validation (`httpCodes`) allows codes out of range 1xx-5xx
 - [ ] Should the status codes number range be validated, validation should be optionally turned off, not hard-coded

For CakePHP 4.x
 - [ ] There should be way to set statusCodes (method `httpCodes` is already removed, and no other documented way is available)
 - [ ] Should the status codes number range be validated, validation should be optionally turned off, not hard-coded

ADmad:
Refs #14372

### DIFF
--- a/src/Http/Response.php
+++ b/src/Http/Response.php
@@ -44,6 +44,10 @@ class Response implements ResponseInterface
 {
     use MessageTrait;
 
+    public const MIN_STATUS_CODE_VALUE = 100;
+
+    public const MAX_STATUS_CODE_VALUE = 599;
+
     /**
      * Allowed HTTP status codes and their default description.
      *
@@ -619,7 +623,7 @@ class Response implements ResponseInterface
      */
     protected function _setStatus(int $code, string $reasonPhrase = ''): void
     {
-        if (!isset($this->_statusCodes[$code])) {
+        if ($code < static::MIN_STATUS_CODE_VALUE || $code > static::MAX_STATUS_CODE_VALUE) {
             throw new InvalidArgumentException(sprintf(
                 'Invalid status code: %s. Use a valid HTTP status code in range 1xx - 5xx.',
                 $code
@@ -627,7 +631,7 @@ class Response implements ResponseInterface
         }
 
         $this->_status = $code;
-        if (empty($reasonPhrase)) {
+        if ($reasonPhrase === '' && isset($this->_statusCodes[$code])) {
             $reasonPhrase = $this->_statusCodes[$code];
         }
         $this->_reasonPhrase = $reasonPhrase;
--- a/tests/TestCase/Http/ResponseTest.php
+++ b/tests/TestCase/Http/ResponseTest.php
@@ -1464,7 +1464,7 @@ class ResponseTest extends TestCase
     }
 
     /**
-     * Test with protocol.
+     * Test with status code.
      *
      * @return void
      */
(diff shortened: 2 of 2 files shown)


dereuromark (approved):

markstory (commented):
Looks good. I also checked that our Client\Response is more permissive and allows any status code. This is a good thing in my opinion as the we need to be permissive in what we accept but stricter in what we output.

smarek:
One issue, that might be with merged implementation, in my environment, if you provide just HTTP status code, but no message, Apache webserver will replace the set status code with 500. I was not able to reproduce with Nginx, so this might be Apache issue

I propose to patch https://github.com/cakephp/cakephp/blob/master/src/Http/ResponseEmitter.php#L169 to ensure, that the status code is not sent without Reason-Phrase

RFC says (https://tools.ietf.org/html/rfc7230#section-3.1.2) :
     status-line = HTTP-version SP status-code SP reason-phrase CRLF

So it should not be really correctly possible to send status-code and omit reason-phrase, if I read it correctly

Tested in environment:
```
Server version: Apache/2.4.38 (Debian) or 
Server built:   2019-10-15T19:53:42
#and
Server version: Apache/2.4.41 (Debian)
Server built:   2019-08-16T19:44:16

PHP 7.3.14-1~deb10u1 (cli) (built: Feb 16 2020 15:07:23) ( NTS )
Copyright (c) 1997-2018 The PHP Group
Zend Engine v3.3.14, Copyright (c) 1998-2018 Zend Technologies
    with Zend OPcache v7.3.14-1~deb10u1, Copyright (c) 1999-2018, by Zend Technologies

Debian 10 buster
```

PR #14374, opened 2020-03-19
Backport changes from #14373 to 3.x

markstory:
Refs #14372

### DIFF
--- a/src/Http/Response.php
+++ b/src/Http/Response.php
@@ -46,6 +46,10 @@ class Response implements ResponseInterface
 {
     use MessageTrait;
 
+    public const MIN_STATUS_CODE_VALUE = 100;
+
+    public const MAX_STATUS_CODE_VALUE = 599;
+
     /**
      * Allowed HTTP status codes and their default description.
      *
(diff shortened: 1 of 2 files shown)


src/Http/Response.php
```diff
@@ -46,6 +46,10 @@ class Response implements ResponseInterface
 {
     use MessageTrait;
 
+    public const MIN_STATUS_CODE_VALUE = 100;
```
dereuromark: public is not good for php5

src/Http/Response.php
```diff
@@ -46,6 +46,10 @@ class Response implements ResponseInterface
...
+
+    public const MAX_STATUS_CODE_VALUE = 599;
```
othercorey:
```suggestion
    const MIN_STATUS_CODE_VALUE = 100;

    const MAX_STATUS_CODE_VALUE = 599;
```

src/Http/Response.php
```diff
@@ -46,6 +46,10 @@ class Response implements ResponseInterface
 {
     use MessageTrait;
 
+    const MIN_STATUS_CODE_VALUE = 100;
+
+    const MAX_STATUS_CODE_VALUE = 599;
```
dereuromark: Usually you would group constants:

- STATUS_CODE_MAX
- STATUS_CODE_MIN

othercorey: Agreed. I can fix it up.

dereuromark: You would also have to do that in master now.

Issue #12238, opened 2018-06-14
BelongsToMany Ignores bindingKey


shayllis:
* [x] bug
* [x] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.5.13
* Platform and Target: MacOS, localhost (AJAX request), MySQL Database

### What you did
Set up BelongsToMany relation without a joining class

### What happened
Relation was created without my given field for bindingKey

### What you expected to happen
Association between my tables to use my given bindingKey

### My Solution
As you can set bindingKey, CakePHP does not it at all. It turned that I could (and should) fix it myself.
He is what I have done:

I updated every junctionGeneration method under "Core/ORM/Association/BelongsToMany".
See:

```
// Core/ORM/Association/BelongsToMany
// Implemented method
public function getBindingKey()
{
     if ($this->_bindingKey === null) {
         $this->_bindingKey = $this->_modelKey($this->getSource()->getTable());
     }

     return $this->_bindingKey;
}
// Changed method
protected function _generateJunctionAssociations($junction, $source, $target)
    {
        $tAlias = $target->getAlias();
        $sAlias = $source->getAlias();

        if (!$junction->association($tAlias)) {
            $junction->belongsTo($tAlias, [
                'foreignKey' => $this->getTargetForeignKey(),
	        'bindingKey' => $this->getBindingKey(), // New line
                'targetTable' => $target
            ]);
        }
        if (!$junction->association($sAlias)) {
            $junction->belongsTo($sAlias, [
                'foreignKey' => $this->getForeignKey(),
                'bindingKey' => $this->getBindingKey(), // New line
                'targetTable' => $source
            ]);
        }
    }

```

lorenzo:
I think we need a different name for this as `bindingKey` is already used for something else, and probably we would need one for the `sourceTable` and another for the `targetTable`.

Do you think you can provide a pull request for this new feature?

shayllis:
Sure: [https://github.com/cakephp/cakephp/pull/12239](https://github.com/cakephp/cakephp/pull/12239).

BindingKey is also a name given in belongsTo Associations, as you said, to set origin table key. I think bindingKey could be used in a belongsToMany because its referring to the current table. To avoid confusing, it could be given another name as well. I suggest "currentTableKey".

markstory:
Refs #12061

markstory:
This will be fixed in 4.1

Issue #12061, opened 2018-05-08
belongsToMany uses primaryKey instead of bindingKey


stefanniedermaier:
This is a (multiple allowed):

* [x] bug ?
* [ ] enhancement
* [x] feature-discussion (RFC)

* CakePHP Version: 3.6.2

### What you did
I created a `UsersTable` and a `LabelsTable` associated with a `belongsToMany`-Association. Both Tables use a `id` column used for foreignKeys and a `uid` column as primaryKey.
```php
class UsersTable extends BaseTable
{
    public function initialize(array $config)
    {
        $this->setPrimaryKey('uid');

        $this->belongsToMany('Labels')
            ->setForeignKey('user_id')
            ->setBindingKey('id')
            ->setTargetForeignKey('label_id')
            ->setThrough('LabelsUsers');
    }
}
```
```php
class LabelsUsersTable extends Table
{
    public function initialize(array $config)
    {
        $this->belongsTo('Labels')
            ->setForeignKey('label_id')
            ->setJoinType('INNER')
            ->setBindingKey('id');

        $this->belongsTo('Users')
            ->setForeignKey('user_id')
            ->setJoinType('INNER')
            ->setBindingKey('id');
    }
}
```
```php
class LabelsTable extends Table
{
    public function initialize(array $config)
    {
        $this->setPrimaryKey('uid');

        $this->belongsToMany('Users')
            ->setForeignKey('label_id')
            ->setBindingKey('id')
            ->setTargetForeignKey('user_id')
            ->setThrough('LabelsUsers');
    }
}
```

### What happened
When saving a User Entity including associated Labels the `BelongsToMany` class is using the primaryKey of the LabelsTable instead of the configured bindingKey of the association table.    

`public/vendor/cakephp/cakephp/src/ORM/Association/BelongsToMany.php`
the issue occures cause of line 836
```php
$targetKeys = array_combine($assocForeignKey, $e->extract($targetPrimaryKey));
```
what i would expect is something like
```php
$targetKeys = array_combine($assocForeignKey, $e->extract((array)$belongsTo->getBindingKey()));
```
This allows the data to be saved as expected, but isn't a full fix, cause reading the User with a `contain => ['Labels']` doesn't return the data as expected.
Is this a Bug in the ORM or did i misconfigure the association?

stefanniedermaier:
The database uses different keys than the models.
Database: 
`uid string(50) unique`
`id int(11) primary autoIncrement`
[...]

PR #14377, opened 2020-03-19
4.1 - Use junction association bindingKey instead of target primaryKey

othercorey:
Fixes https://github.com/cakephp/cakephp/issues/12061
Fixes https://github.com/cakephp/cakephp/issues/12238

This allows a BelongsToMany association to use a different bindingKey than the primaryKey for the table.  The default value for the bindingKey is the primary key of the target table.

### DIFF
--- a/src/ORM/Association/BelongsToMany.php
+++ b/src/ORM/Association/BelongsToMany.php
@@ -312,10 +312,17 @@ class BelongsToMany extends Association
     {
         $junctionAlias = $junction->getAlias();
         $sAlias = $source->getAlias();
+        $tAlias = $target->getAlias();
+
+        $targetBindingKey = null;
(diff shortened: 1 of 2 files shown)


dereuromark:
Do we need to backport this to 3.next? Or any way people can use it on their own as project overwrite?

othercorey:
Because the default behavior doesn't change, we could consider back-porting this. Although very unlikely, if someone did manually add a `setBindingKey()` call to the junction table association, it could throw an error if they typed the column name wrong.

markstory (commented):
Would be good to have an integration test for this. Right now its hard to ensure we actually have the code working correctly.

othercorey:
> Would be good to have an integration test for this. Right now its hard to ensure we actually have the code working correctly.

Added unit test using contain and link with custom binding key.

PR #14380, opened 2020-03-21
4.1 - Renamed associations() to getAssociations()

othercorey:


ADmad:
This is just going to be an annoyance, having to change the method when upgrading from 4.0 to 4.1. I don't see much benefit of it.

othercorey:
I thought it would be very low use.

PR #14383, opened 2020-03-22
add afterMarshal event

saeideng:
how to use ?
```php
//model
use Cake\Event\EventInterface;

public function afterMarshal(EventInterface $event, EntityInterface $entity)
{
    //
}
```

why we need to this ?
for validation (set error) and manipulate  the result after `patchEntity()` there not any hook for this

other options ??
- buildValidator

works well but it suitable  for manipulate fields separately not several fields together and impossible to conditionally mark them as invalid/error 

- beforeRules
- afterRules

too late for manipulate entity because if there was an error on entity 
even `save()` and these callbacks/events  will not call/trigger

### DIFF
--- a/tests/TestCase/ORM/MarshallerTest.php
+++ b/tests/TestCase/ORM/MarshallerTest.php
@@ -3242,6 +3242,105 @@ class MarshallerTest extends TestCase
     }
 
     /**
+     * Test Model.afterMarshal event.
+     *
+     * @return void
+     */
+    public function testAfterMarshalEvent()
+    {
+        $data = [
+            'title' => 'original title',
+            'body' => 'original content',
+            'user' => [
+                'name' => 'Robert',
+                'username' => 'rob',
+            ],
+        ];
+
+        $marshall = new Marshaller($this->articles);
+
+        $this->articles->getEventManager()->on(
+            'Model.afterMarshal',
+            function ($e, $entity, $data, $options) {
+                $this->assertInstanceOf('Cake\ORM\Entity', $entity);
+                $this->assertArrayHasKey('validate', $options);
+                $this->assertFalse($options['isMerge']);
+
+                $data['title'] = 'Modified title';
+                $data['user']['username'] = 'robert';
+                $options['associated'] = ['Users'];
+
+                $entity->body = 'Modified body';
+            }
+        );
+
+        $entity = $marshall->one($data);
+
+        $this->assertSame('original title', $entity->title, '$data is immutable');
+        $this->assertSame('Modified body', $entity->body);
+        // both $options and $data are unchangeable
+        $this->assertIsArray($entity->user, '$options[\'associated\'] is ignored');
+        $this->assertSame('Robert', $entity->user['name']);
+        $this->assertSame('rob', $entity->user['username']);
+    }
+
+    /**
+     * Test Model.afterMarshal event on patchEntity.
+     * when $options['fields'] is set and is empty
+     *
+     * @return void
+     */
+    public function testAfterMarshalEventOnPatchEntity()
+    {
+        $data = [
+            'title' => 'original title',
+            'body' => 'original content',
+            'user' => [
+                'name' => 'Robert',
+                'username' => 'rob',
+            ],
+        ];
+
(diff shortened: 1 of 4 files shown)


dereuromark:
I think this makes perfect sense.
I have some cake2 apps that are really hard to update without this - on validating based on the marshalling results of several fields together.
buildValidator() can only do so much here in terms of multiple fields and their dependencies.

	/**
	 * @param \Cake\Event\EventInterface $event
	 * @param \Cake\Datasource\EntityInterface $entity
	 *
	 * @return void
	 */
	public function afterMarshal(EventInterface $event, EntityInterface $entity) {
		if ($entity->one && $entity->two && ...) {
			$entity->setError('one', 'Can only be used without the others or alike.');
		}
	}

Setting errors in the marshalling is not possible yet due to it being an array yet, and no entity.

And domain rules are the wrong part of the system, being invoked separately and creating weird usability as it a) fails, then b) passes those fields only to c) get other fails.
As such I am never using domain rules for anything but low-level integrity parts, and especially not for data validation.

saeideng:
>	if ($entity->one && $entity->two && ...) {
		$entity->setError('one', 'Can only be used without the others or alike.');
	}

what I had in mind

> And domain rules are the wrong part of the system

even I sometimes dont use `save()` I just need to know if there is any error on entity in controller for example to check email and send it and `$entity->getErrors()` is enough to me
so for this purpose we should have a hook to check multiple fields in same time (out of application rule ) 
and this event could helps

markstory:
Would this new event be fired after each entity like beforeMarshall is, or at the end of the entire process? Would you also want access to the original request data?

dereuromark:
I think for simplicity and keeping things in sync also after each entity.
I don't think that is a must have, as the main goal is to work on the entity now to properly add errors where needed now based on outcome.

saeideng:
> Would you also want access to the original request data?

no idea , but do we need to `$options` like `beforeMarshal`
any way these ( `$options` , '$data` will be immutable not like the `beforeMarshal`)

dereuromark:
I ran a real-life test using updated https://github.com/dereuromark/executionorder/pull/8 app
And this would be the scenario/flow of events for the patch only use case:
```
Controller::initialize
FooComponent::beforeFilter
Controller::beforeFilter
FooComponent::startup
Controller.action
TokensTable:initialize
AlphaBehavior:initialize
AlphaBehavior:beforeMarshal
TokensTable:beforeMarshal
TokensTable:validationDefault
AlphaBehavior:buildValidator
TokensTable:buildValidator
AlphaBehavior:afterMarshal
TokensTable:afterMarshal
FooComponent::beforeRender
Controller::beforeRender
FooComponent::shutdown
Controller::afterFilter
```

markstory:
>  any way these ( $options , $data ) will be immutable and not like the beforeMarshal)

Yes I think having access to the options + data will be useful for the kinds of scenarios you're hoping to address with this new event.

dereuromark:
@saeideng Will you add the options in?
I think then we can proceed.

saeideng:
```php
//model
use ArrayObject;
use Cake\Event\EventInterface;

public function (EventInterface $event, EntityInterface $entity, ArrayObject $data, ArrayObject $options)
{
    //
}
```

tests/TestCase/ORM/MarshallerTest.php
```diff
@@ -3241,6 +3241,105 @@ function ($e, $data) {
...
+            }
+        );
+
+        $entity = $marshall->one($data);
+
+        $this->assertSame('original title', $entity->title, '$data is immutable');
+        $this->assertSame('Modified body', $entity->body);
+        // both $options and $data are unchangeable
+        $this->assertIsArray($entity->user, '$options[\'associated\'] is ignored');
+        $this->assertSame('Robert', $entity->user['name']);
```
dereuromark, line 3283: $entity->user->name right?

saeideng: no 
since `$data` and `$options` are unchangeable
```$options['associated'] = ['Users'];``` will not affect 
so `user` is not a entity

dereuromark:
Looks good!

PR #14384, opened 2020-03-22
Add RedirectMiddleware

dereuromark:
In 3.x we could use a hack with $reponse->send() and exit(); to make redirects from within the components and alike.
In 4.x this is not possible anymore, a cleaner approach is using the RedirectException available already.

But there is a core Middleware missing for handling this.

I added one on project level as demo here:
https://github.com/dereuromark/cakephp-sandbox/commit/b4fb8635fd8ea6bef3ac057d2351b776dc083ea4

You can see how it can be used
https://sandbox.dereuromark.de/sandbox/search-examples/table?page=12 (200 OK)
But when you hit the next page usually, you get a hard exception
This is not very user-friendly, as often, just an item got deleted at the end and thus the page count is one less. So if userland code wants to easily implement a paginator that does this, such middleware is vital here.
https://sandbox.dereuromark.de/sandbox/search-examples/table?page=13 (302 redirect)
This one now redirects to the last page on its own.

With this change, it is easier inside apps to leverage the given exception now.

### DIFF
--- a/src/Routing/Middleware/RedirectMiddleware.php
+++ b/src/Routing/Middleware/RedirectMiddleware.php
@@ -1,0 +1,52 @@
+<?php
+declare(strict_types=1);
+
+/**
+ * CakePHP(tm) : Rapid Development Framework (https://cakephp.org)
+ * Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
+ *
+ * Licensed under The MIT License
+ * For full copyright and license information, please see the LICENSE.txt
+ * Redistributions of files must retain the above copyright notice.
+ *
+ * @copyright     Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
+ * @link          https://cakephp.org CakePHP(tm) Project
+ * @since         4.1.0
+ * @license       https://opensource.org/licenses/mit-license.php MIT License
+ */
+namespace Cake\Routing\Middleware;
+
+use Cake\Routing\Exception\RedirectException;
+use Laminas\Diactoros\Response\RedirectResponse;
+use Psr\Http\Message\ResponseInterface;
+use Psr\Http\Message\ServerRequestInterface;
+use Psr\Http\Server\MiddlewareInterface;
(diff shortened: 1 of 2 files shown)


src/Routing/Middleware/RedirectMiddleware.php
```diff
@@ -0,0 +1,51 @@
...
+class RedirectMiddleware implements MiddlewareInterface
+{
+    /**
+     * Ensures any RedirectException thrown is redirected.
+     *
+     * @param \Psr\Http\Message\ServerRequestInterface $request The request.
+     * @param \Psr\Http\Server\RequestHandlerInterface $handler The request handler.
+     * @return \Psr\Http\Message\ResponseInterface A response.
+     */
+    public function process(ServerRequestInterface $request, RequestHandlerInterface $handler): ResponseInterface
```
markstory: Do we need a Middleware for this? Handling `RedirectException` seems like behavior we could build into `ControllerFactory::invoke()` or into Http/Server. If we handled it in Http/Server we would need to update the ErrorHandlerMiddleware to not swallow those errors which is a bit clunky and not idea in my opinion.

dereuromark: I am also fine with that :)
This was just my first idea.

markstory: Throwing the exception from a component or controller class is another opt-in mechanism :smile:

One drawback to handing `RedirectException` in the `ControllerFactory` is that RedirectExceptions raised in middleware won't work the same.  Perhaps the `ErrorHandlerMiddleware` is the best place for these exceptions to be handled. It already handles lots of other kinds of exceptions, and this one has a pretty straightforward action to take.

Should we also expand the `RedirectException` to allow headers to be added?

dereuromark: I am fine if someone draws up a better alternative PR.

dereuromark:
With https://github.com/cakephp/cakephp/pull/14397 continuing this I will close this one.

PR #14389, opened 2020-03-25
Encode binary data in query logs.

### LINKED ISSUE #14381: BinaryUUIDs shouldnt be logged in binary
This is a (multiple allowed):

* [x] bug
* [x] enhancement

* CakePHP Version: master

### What you did

Working with binary-16 UUID (type `binaryuuid`) columns in a Cake4 app.

### What happened

image

DebugKit cannot display/format the value.
The value is missing (or hidden in output) and it is hard to retrieve in some cases the exact value here, especially when dynamically fetched and added to another query.

If you debug the array logged, you will find the value to be there, but as a binary string
'÷¬\x01#©8N€„\x0Eïè’\x05\x132' inside the queries log array:
```
^ array:4 [▼
  0 => array:3 [▼
    "query" => b"duration=1 rows=1 SELECT ... FROM `exposed_users` `ExposedUsers` WHERE `ExposedUsers`.`uuid` = '÷¬\x01#©8N€„\x0Eïè’\x05\x132' ORDER BY `ExposedUsers`.`name` ASC LIMIT 1 ◀"
    "took" => 1
    "rows" => 1
  ]
```

When logging away queries with binaryuuid I think it is best to use the non-binary form here.
As it seems the info "uuid" is not passed along with it, making it harder later in DebugKit or other tools to make this byte string visible again with the proper scope/conversion applied.
And mixing binary and normal text in log files is also not a good idea IMO.

markstory:
If a string value has non-printable data we can encode it in hexadecimal. This ensures the value will be visible in logs and not break HTML output or log forwarders that don't handle non-ascii bytes.

Fixes #14381

### DIFF
--- a/tests/TestCase/Database/Log/LoggedQueryTest.php
+++ b/tests/TestCase/Database/Log/LoggedQueryTest.php
@@ -17,8 +17,8 @@ declare(strict_types=1);
 namespace Cake\Test\TestCase\Database\Log;
 
 use Cake\Database\Log\LoggedQuery;
-use Cake\Utility\Text;
 use Cake\TestSuite\TestCase;
+use Cake\Utility\Text;
 
 /**
  * Tests LoggedQuery class
--- a/tests/TestCase/Database/Log/LoggedQueryTest.php
+++ b/tests/TestCase/Database/Log/LoggedQueryTest.php
@@ -17,6 +17,7 @@ declare(strict_types=1);
 namespace Cake\Test\TestCase\Database\Log;
 
 use Cake\Database\Log\LoggedQuery;
+use Cake\Utility\Text;
 use Cake\TestSuite\TestCase;
 
 /**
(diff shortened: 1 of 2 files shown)


tests/TestCase/Database/Log/LoggedQueryTest.php
```diff
@@ -111,6 +112,22 @@ public function testStringInterpolationSpecialChars()
...
+    /**
+     * Tests that query placeholders are replaced when logged
+     *
+     * @return void
+     */
+    public function testBinaryInterpolation()
+    {
+        $query = new LoggedQuery();
+        $query->query = 'SELECT a FROM b where a = :p1';
+        $uuid = str_replace('-', '', Text::uuid());
```
othercorey: Why would a normal text uuid be considered binary data? Why isn't the hex string printable? Wouldn't want want a normal uuid to be printed?

dereuromark: Binary16 here, see issue.

othercorey: My mistake, I misread the query code below is all. I didn't have a problem with the actual change :)

dereuromark:
```
WHERE 
  `ExposedUsers`.`uuid` = 'c94be2399546431a97f4aa351fcd697f' 
```

works like a charm, thank you!
One question: Having these kind of (primary) keys being represented in default UUID form with - as separator (`c94be239-9546-431a-97f4-aa351fcd697f`) in logs and alike is probably not possible, right?
It would be easier to match them when searching/filtering.
No big deal, just wanted to clarify this as follow up.

dereuromark:
Also, I was kind of hoping we could map this based on the schema detection in general.
"asdf\n\r\t" would also be true for the ctype_print() check, and thus doing this primarily based on Type map class being BinaryUuid or alike would have less false positives.
If we want we could also check on (byte) length of 16 (`strlen()`) here to avoid possible false hex conversations.

dereuromark:
See https://github.com/cakephp/cakephp/compare/binary
So if we cannot make this based on the type/map data and being a true binary input, I would suggest merging my follow up that we don't scramble other data, containing tabs, or alike.

PR #14391, opened 2020-03-25
Change collection function return types

### LINKED ISSUE #14367: Return specific types for CollectionInterface methods
`CollectionInterface` methods all return type `CollectionInterface`.

```
public function listNested($dir = 'desc', $nestingKey = 'children'): CollectionInterface;
```

However, these methods are sometimes implemented with specific types that provide extra methods that we reference in documentation. However, there is no type hinting for these methods.

https://book.cakephp.org/4/en/core-libraries/collections.html#Cake\Collection\Collection::listNested

This causes issues with static analyzers and other type resolvers that require special work-arounds.

```
/** @var \Cake\Collection\Iterator\TreeIterator $nested */
$nested = $results->listNested();

return $nested->printer($options['valuePath'], $options['keyPath'], $options['spacer']);
```

Zwartpet:
Fixes #14367

Collection function all returned a CollectionInterface, but this gives issues with typehinting.

### DIFF
--- a/src/Collection/CollectionTrait.php
+++ b/src/Collection/CollectionTrait.php
@@ -75,7 +75,7 @@ trait CollectionTrait
     /**
      * {@inheritDoc}
      *
-     * @return \Cake\Collection\CollectionInterface
+     * @return \Cake\Collection\Iterator\FilterIterator
      */
     public function filter(?callable $c = null): CollectionInterface
     {
@@ -91,7 +91,7 @@ trait CollectionTrait
     /**
      * {@inheritDoc}
      *
-     * @return \Cake\Collection\CollectionInterface
+     * @return \Cake\Collection\Iterator\FilterIterator
      */
     public function reject(callable $c): CollectionInterface
     {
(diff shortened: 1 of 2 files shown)


Zwartpet:
@othercorey proposed this change for [the issue you made](https://github.com/cakephp/cakephp/issues/14367), I've changed the actual return types as well but maybe that should stay a CollectionInterface?

othercorey:
I don't know if we can do this for 4.1, however we rely on these types returned by these methods. For example, a generic `CollectionInterface` returned by `listNested()` doesn't help.

All of the internal implementations use `CollectionTrait`, but it's possible someone out there implemented an entirely different `CollectionInterface`. It wouldn't be compatible with documented use of these methods though.

ADmad:
Also having methods of Collectioninterface be type hinted with methods of its child interfaces feels odd.

othercorey:
> Also having methods of Collectioninterface be type hinted with methods of its child interfaces feels odd.

That is awkward. I would like to find something better than non-hinted method calls.

dereuromark:
This breaks the public API and usually cannot be done before the next major.
Could we add those as docblock return annotations?
That would be BC.
Just not sure if it should replace or be a combination of both (using `&` or `|` ?)

Zwartpet:
Something like this? Or only the docblock and removing the actual return type in the trait?

dereuromark:
Does this work for you regarding typehinting? And for everyone else?

othercorey:
@Zwartpet I will revisit this when I have time to see if the types fix the psalm errors at least.

PR #14394, opened 2020-03-25
Fix binary log conversion.

dereuromark:
Follow https://github.com/cakephp/cakephp/pull/14389
as the current check could false positive.

### DIFF
--- a/src/Database/Log/LoggedQuery.php
+++ b/src/Database/Log/LoggedQuery.php
@@ -78,8 +78,8 @@ class LoggedQuery implements JsonSerializable
             }
 
             if (is_string($p)) {
-                // Likely binary data.
-                if (!ctype_print($p)) {
+                // Likely binary UUID.
+                if (strlen($p) === 16 && !ctype_print($p)) {
                     $p = bin2hex($p);
                 }
 
--- a/src/Database/Log/LoggedQuery.php
+++ b/src/Database/Log/LoggedQuery.php
@@ -78,8 +78,9 @@ class LoggedQuery implements JsonSerializable
             }
 
             if (is_string($p)) {
-                // Likely binary UUID.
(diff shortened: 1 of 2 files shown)


othercorey:
Do we need to worry about binary blob data?

src/Database/Log/LoggedQuery.php
```diff
@@ -78,8 +78,8 @@ protected function interpolate(): string
             }
 
             if (is_string($p)) {
-                // Likely binary data.
-                if (!ctype_print($p)) {
+                // Likely binary UUID.
+                if (strlen($p) === 16 && !ctype_print($p)) {
```
othercorey: I think the only complete solution is regex. This will limit all other non-ascii characters as well which will convert a LOT of non-basic english strings even if we replaced all the whitespace characters before checking.

> Only ascii 32 thru 126 (inclusive) are considered printable

markstory: Another approach could be to use a pattern like `'/[\x00-\x1F\x7F]/u'` and see if the results are different than the original value. If they are we can infer the string was bytes.

othercorey: Do we have a test that logs a couple utf-8 unicode characters?

dereuromark: > Do we have a test that logs a couple utf-8 unicode characters?

Can you add what you have in mind? Or post such a string here that I can test it?
The current solution now here already looks less problematic than the check before.

othercorey: It's fine. I leave it to you two :)

dereuromark: Strings like https://www.soscisurvey.de/tools/view-chars.php work fine.

PR #14397, opened 2020-03-28
Add Redirect exception support

markstory:
An alternative to #14384. I've added a new RedirectException class that isn't coupled to Routing package and is integrated with the ErrorHandlerMiddleware which enables any middleware/controller/component to raise a RedirectException and get the intended effect.

### DIFF
--- a/src/Http/Exception/RedirectException.php
+++ b/src/Http/Exception/RedirectException.php
@@ -33,7 +33,6 @@ use Cake\Core\Exception\Exception;
  */
 class RedirectException extends Exception
 {
-
     /**
      * Headers to include in the response.
      *
--- a/src/Http/Exception/RedirectException.php
+++ b/src/Http/Exception/RedirectException.php
@@ -1,0 +1,95 @@
+<?php
+declare(strict_types=1);
+
+/**
+ * CakePHP(tm) : Rapid Development Framework (https://cakephp.org)
+ * Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
+ *
+ * Licensed under The MIT License
+ * For full copyright and license information, please see the LICENSE.txt
+ * Redistributions of files must retain the above copyright notice.
+ *
+ * @copyright     Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
+ * @link          https://cakephp.org CakePHP(tm) Project
+ * @since         4.1.0
+ * @license       https://opensource.org/licenses/mit-license.php MIT License
+ */
+namespace Cake\Http\Exception;
+
+use Cake\Core\Exception\Exception;
+
+/**
+ * An exception subclass used by routing and application code to
+ * trigger a redirect.
+ *
+ * The URL and status code are provided as constructor arguments.
+ *
+ * ```
+ * throw new RedirectException('http://example.com/some/path', 301);
+ * ```
+ *
+ * Additional headers can also be provided in the constructor, or
+ * using the addHeaders() method.
+ */
+class RedirectException extends Exception
+{
+
+    /**
+     * Headers to include in the response.
+     *
+     * @var array
+     */
+    protected $headers = [];
+
+    /**
+     * Constructor
+     *
+     * @param string $target The URL to redirect to.
+     * @param int $code The exception code that will be used as a HTTP status code
+     * @param array $headers The headers that should be sent in the unauthorized challenge response.
+     */
+    public function __construct(string $target, int $code = 302, array $headers = [])
+    {
+        parent::__construct($target, $code);
+        $this->addHeaders($headers);
+    }
+
+    /**
+     * Add headers to be included in the response generated from this exception
+     *
+     * @param array $headers An array of `header => value` to append to the exception.
+     *  If a header already exists, the new values will be appended to the existing ones.
+     * @return $this
+     */
+    public function addHeaders(array $headers)
+    {
+        foreach ($headers as $key => $value) {
+            $this->headers[$key][] = $value;
+        }
+
+        return $this;
+    }
+
+    /**
+     * Remove a header from the exception.
+     *
+     * @param string $key The header to remove.
+     * @return $this
+     */
+    public function removeHeader(string $key)
+    {
+        unset($this->headers[$key]);
+
+        return $this;
+    }
+
+    /**
+     * Get the headers from the exception.
+     *
+     * @return array
+     */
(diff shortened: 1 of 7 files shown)


src/Http/Exception/RedirectException.php
```diff
@@ -0,0 +1,95 @@
...
+ * ```
+ * throw new RedirectException('http://example.com/some/path', 301);
+ * ```
+ *
+ * Additional headers can also be provided in the constructor, or
+ * using the addHeaders() method.
+ */
+class RedirectException extends Exception
+{
+
```
ADmad:
```suggestion
```

src/Routing/Exception/RedirectException.php
```diff
@@ -27,6 +27,9 @@
  * ```
  * throw new RedirectException('http://example.com/some/path', 301);
  * ```
+ *
+ * If you need a more general purpose redirect exception use
+ * Cake\Http\Exception\RedirectException instead of this class.
```
othercorey: Isn't it confusing to have two identically named `RedirectException` classes?

dereuromark: I dont think we need this 2nd exception.
The Error package already heavily relies on the Http one:
- ErrorHandler
- ErrorHandlerMiddleware
- ExceptionRenderer

tests/TestCase/Error/Middleware/ErrorHandlerMiddlewareTest.php
```diff
@@ -143,6 +145,55 @@ public function testHandleException()
...
+    {
+        $request = ServerRequestFactory::fromGlobals();
+        $middleware = new ErrorHandlerMiddleware();
+        $handler = new TestRequestHandler(function () {
+            throw new RedirectException('http://example.org/login');
+        });
+        $result = $middleware->process($request, $handler);
+        $this->assertInstanceOf(ResponseInterface::class, $result);
+        $this->assertEquals(302, $result->getStatusCode());
+        $this->assertEmpty('' . $result->getBody());
```
dereuromark: You could also (string) cast here.

tests/TestCase/Error/Middleware/ErrorHandlerMiddlewareTest.php
```diff
@@ -143,6 +145,55 @@ public function testHandleException()
...
+     * Test creating a redirect response
+     *
+     * @return void
+     */
+    public function testHandleRedirectException()
+    {
+        $request = ServerRequestFactory::fromGlobals();
+        $middleware = new ErrorHandlerMiddleware();
+        $handler = new TestRequestHandler(function () {
+            throw new RedirectException('http://example.org/login');
```
dereuromark: For security purposes, shouldn't we only use internal (relative) URLs?
https://github.com/cakephp/cakephp/pull/14384/files#diff-fd25dba57f71dd4309af4df2494d43e2R55
Or is that up to the user at this point?

src/Routing/Exception/RedirectException.php
ADmad: How about we deprecate this exception and make RedirectRoute also throw the new Http/Exception/RedirectException?

markstory: @dereuromark Using the routing class would make standalone http package usage more cumbersome. I would rather we deprecate the Routing class and have routing depend on Http as that dependency makes more sense.

tests/TestCase/Error/Middleware/ErrorHandlerMiddlewareTest.php
markstory: I think that is up to the user. The exceptions and redirect response don't modify the provided url at all.

src/Routing/Exception/RedirectException.php
dereuromark: Thats fine :)

markstory: I've marked the Routing\RedirectException as deprecated, but not changed any of the code as I didn't want to potentially break compatibility over a cosmetic change.

ADmad:
`RedirectRoute` should also be updated to throw the new exception and accordingly `RoutingMiddleware::proces()` too.

Perhaps the check for redirect exception in `RoutingMiddleware` could be just done away with since `ErrorHandlerMiddleware` now checks and handles it.

dereuromark:
I fixed up the remaining part.

> Perhaps the check for redirect exception in RoutingMiddleware could be just done away with since ErrorHandlerMiddleware now checks and handles it.

That part should probably be kept for the old Exception class to work.
As the new one only checks for the new `Cake\Http\Exception\RedirectException`.

So the PR as it stands now seems fine to me.
Routing handles Route one, otherwise normal ErrorHandler handles the new one.

dereuromark:
How do we want to adjust the tests?
$this->expectException()?
Or should we also catch the new exception in the RoutingMiddleware?

src/Routing/Middleware/RoutingMiddleware.php
```diff
@@ -150,6 +151,11 @@ public function process(ServerRequestInterface $request, RequestHandlerInterface
                 $e->getMessage(),
```
ADmad: Why do we need to catch the new exception too in routing middleware? It will be handled by errorhandler middleware isn't it?

dereuromark: Then we would need to adjust the tests, as I wrote above.
We could also do that.

markstory: By raising the new exception we are risking breaking any application/plugin tests that directly use the current exception class. Is such a break worth the trouble?

dereuromark: This is a minor release, and affects then only tests, easy to fix with the namespace change.
So I think if it makes it cleaner and more straightforward then yes.

ADmad: Another option is make `Routing\Exception\RedirectionException` an alias of `Http\Exception\RedirectException`.

dereuromark: Actually, nothing breaks now the way it is now
The Routing part still catches those exceptions - and otherwise ErrorHandler jumps in.
So the current way is maybe the most pragmatic one forward.
And it can be cleaned up in 5.x.

dereuromark:
Are we OK with the current status, including backward compatibility for tests?

Issue #14398, opened 2020-03-28
IntegrationTestCase regression on session flash


dereuromark:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: dev-master

### What you did
3.x plugin test code
```php
	public function testGenerate() {
		$this->disableErrorHandlerMiddleware();

		$data = [
			'plugin' => 'Tools',
			'name' => 'MyCoolRecords',
		];

		$this->post(['plugin' => 'TestHelper', 'controller' => 'TestFixtures', 'action' => 'generate'], $data);

		$this->assertResponseCode(302);

		$flash = $this->_requestSession->read('Flash.flash');
		$flash = Hash::combine($flash, '{n}.element', '{n}.message');
		$this->assertTextContains('bake fixture MyCoolRecords -q -p Tools', $flash['Flash/info']);
		$this->assertSame('MyCoolRecordsFixture generated.', $flash['Flash/success']);
	}
```

Upgraded to 4.x
```php
	public function testGenerate() {
		$this->disableErrorHandlerMiddleware();
		//$this->_retainFlashMessages = true; // Only necessary for rendering involved.

		$data = [
			'plugin' => 'Tools',
			'name' => 'MyCoolRecords',
		];

		$this->post(['plugin' => 'TestHelper', 'controller' => 'TestFixtures', 'action' => 'generate'], $data);

		$this->assertResponseCode(302);

		$flash = $this->_requestSession->read(); dd($flash);	
		//FIXME - why null?
		/*
		$flash = $this->_requestSession->read('Flash.flash');
		$flash = Hash::combine($flash, '{n}.element', '{n}.message');
		$this->assertTextContains('bake fixture MyCoolRecords -q -p Tools', $flash['flash/info']);
		$this->assertSame('MyCoolRecordsFixture generated.', $flash['flash/success']);
		*/
	}
```

### What happened
The session is empty, despite the 302 and thus no rendering happening.

### What you expected to happen
This shouldn't happen.
The session data, especially the flash messages (I double-checked, they are written to session in controller code), should still be there for test assertion.

--- 

Reproducible directly when cloning master: https://github.com/dereuromark/cakephp-test-helper

markstory:
This limitation is documented in the migration guide from what I remember. The cause is that the session is closed at the end of each simulated request and because  all the code runs in the CLI, the session restarts fresh. This problem was also reported in  #14270

[...]

PR #14399, opened 2020-03-29
Add session access for IntegrationTest testing.

### LINKED ISSUE #14270: CakePHP 4.0.3: Asserting Session Variable after Request in Test Case clears $_SESSION
Let me know if you need any more info from me. Possible I am missing something here, but this worked before.

___

This is a (multiple allowed):

* [x] bug
* [ ] enhancement

* CakePHP Version: 4.0.3
* Platform and Target: Apache 2.4, Ubuntu 18.04, Microsoft SQL Server 2017

### What you did
We're trying to upgrade a Cake 3.8 application to Cake 4.0.x. Some of our test cases do the following:
```
    public function testSomePostRequest()
    {
        $this->enableCsrfToken();
        $this->enableSecurityToken();
       
        // form data
        $data = [
            'FormItem1' => 'something'
        ];
       
        // this action will proceed normally, and set a session key in 'sessionKey'
        $this->post('some-plugin/some-controller/some-action', $data);

        /* We want to assert now that certain session variables were set */
         
        // this code works (but is not desired)
        $this->assertArrayHasKey($_SESSION['sessionKey']);
        $this->assertNotNull($_SESSION['sessionKey']);

        // this code worked in 3.8, but stopped working in 4.0.x
        $value = $this->_requestSession->read('sessionKey'); // alternatively, was also possible to use $this->_controller->getRequest()->getSession(), but that returns null also
        $this->assertNotNull($value); // causes error

        $this->assertRedirect('/redirect/somehwere/else');
    }
```

### What happened
$_SESSION is getting cleared when $this->_requestSession->read() or $this->_controller->getRequest()->getSession()->read() is being called.

When I dug a little deeper, for some reason, $this->started() in Session.php is returning false, which is causing $this->start() to be executed under the first IF condition in ->read(). 

$_SESSION should contain the session variables that were set by the controller POST action, as in Cake 3.8.

dereuromark:
Resolves https://github.com/cakephp/cakephp/issues/14270 and https://github.com/cakephp/cakephp/issues/14398

Fixes the session testing regression introduced in cake 4.

Now, you can do this (again):
```php
	$this->post(['plugin' => 'TestHelper', 'controller' => 'TestFixtures', 'action' => 'generate'], $data);

	$this->assertResponseCode(302);

	$flash = $this->getTestSession()->read('Flash.flash');
	$flash = Hash::combine($flash, '{n}.element', '{n}.message');
	$this->assertTextContains('bake fixture MyCoolRecords -q -p Tools', $flash['flash/info']);
	$this->assertSame('MyCoolRecordsFixture generated.', $flash['flash/success']);
```

The wrapper is read only, thus it only exposes read() and check().
It is also protected as there is no need afaik to have public access from a third party location.

It seems like the Session itself doesn't have a readOrFail(), so I didn't add one here either.
This could be a follow up addition for both classes.
See https://github.com/dereuromark/cakephp-test-helper/commit/0a8eaf56b02d39c25293ac51d9b1591050a437ed#diff-74bfea6f6fb4add7746dde591fe332efR76-R83 as possible implementation idea.

### DIFF
--- a/src/TestSuite/IntegrationTestTrait.php
+++ b/src/TestSuite/IntegrationTestTrait.php
@@ -1300,7 +1300,7 @@ trait IntegrationTestTrait
     /**
      * @return \Cake\TestSuite\TestSession
      */
-    protected function getTestSession(): TestSession
+    protected function getSession(): TestSession
     {
         return new TestSession($_SESSION);
     }
--- a/src/TestSuite/IntegrationTestTrait.php
+++ b/src/TestSuite/IntegrationTestTrait.php
@@ -1296,4 +1296,12 @@ trait IntegrationTestTrait
             PHP_EOL .
             $exception->getTraceAsString();
     }
+
+    /**
+     * @return \Cake\TestSuite\TestSession
+     */
+    protected function getTestSession(): TestSession
+    {
+        return new TestSession($_SESSION);
+    }
 }
--- a/src/Http/Session.php
+++ b/src/Http/Session.php
@@ -419,6 +419,10 @@ class Session
             return false;
         }
 
+        if ($name === null) {
+            return (bool)$_SESSION;
+        }
+
         return Hash::get($_SESSION, $name) !== null;
     }
 
--- a/src/TestSuite/TestSession.php
+++ b/src/TestSuite/TestSession.php
@@ -50,6 +50,10 @@ class TestSession
             return false;
         }
 
+        if ($name === null) {
+            return (bool)$this->session;
+        }
+
         return Hash::get($this->session, $name) !== null;
     }
 
--- a/src/TestSuite/TestSession.php
+++ b/src/TestSuite/TestSession.php
@@ -1,0 +1,75 @@
+<?php
+declare(strict_types=1);
+
+/**
+ * A class to contain and retain the session during integration testing.
+ *
+ * CakePHP(tm) : Rapid Development Framework (https://cakephp.org)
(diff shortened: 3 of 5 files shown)


dereuromark:
I pointed this to master as this is needed as a regression fix coming from 3.x.
Fully BC and only effecting tests as opt-in.

I can open a 2nd PR against 4.next for the readOrFail() enhancement both for Session and this TestSession class later.

dereuromark:
While writing tests I found another bug in core Session check()
As this method is nullable, check() itself without arguments should check on the session being not empty.
This was broken and is now also fixed.

src/TestSuite/IntegrationTestTrait.php
```diff
@@ -1296,4 +1296,12 @@ protected function extractExceptionMessage(Exception $exception): string
             PHP_EOL .
             $exception->getTraceAsString();
     }
+
+    /**
+     * @return \Cake\TestSuite\TestSession
+     */
+    protected function getTestSession(): TestSession
```
markstory: Should we call this `getSession()`? I don't know if the `Test` prefix helps much as the code is in a test case already.

markstory (commented):
The wrapper approach is nice. It gives the same interface as the real session.

src/TestSuite/IntegrationTestTrait.php
dereuromark: I wanted to be sure that people don't confuse this with the existing _requestSession.
But I am fine if we call the method a bit simpler.

dereuromark:
Is this good to go?

PR #14403, opened 2020-03-29
Fix errors reported by phpstan

ADmad:

### DIFF
--- a/src/Cache/CacheEngine.php
+++ b/src/Cache/CacheEngine.php
@@ -178,8 +178,6 @@ abstract class CacheEngine implements CacheInterface, CacheEngineInterface
                 $this->setConfig('duration', $restore);
             }
         }
-
-        return false;
     }
 
     /**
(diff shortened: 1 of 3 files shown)


src/Cache/CacheEngine.php
```diff
@@ -178,8 +178,6 @@ public function setMultiple($values, $ttl = null): bool
                 $this->setConfig('duration', $restore);
             }
         }
-
-        return false;
```
dereuromark: How come?

ADmad: Because the `try` block has return statement, so if there's no exception method will return and if there is exception then too this statement will never be reached.

I too was confused earlier and opened an issue with phpstan but later understood.

ADmad: https://3v4l.org/rCD4o
https://3v4l.org/MJ5VY

PR #14407, opened 2020-03-29
Cleanup querystring params with null values to prevent reverse routing failures.

### LINKED ISSUE #14404: Pagination Helper creating wrong links when using root URL
This is a (multiple allowed):

* [x] bug
* [ ] enhancement

* CakePHP Version: 4.0.5
* Platform and Target: PHP 7.3, MySQL, Ubuntu

### What you did
I added `$builder->connect('/', ['controller' => 'Pictures', 'action' => 'index']);` to my routes.php to list my pictures at the root of my url in paginated form.

### What happened
The URL generated by `$this->Paginator->first()` and ` $this->Paginator->prev()` (for the first page) is now **mywebsite.com/pictures**. When I remove `$builder->fallbacks();` from routes.php I get Missing-Route-Exception-Error.

I'd expect the first- and prev-functions to return the right URL for the first page, which after my change in the routes.php should be **mywebsite.com** instead of **mywebsite.com/pictures**. Also it shouldn't give me the Missing-Route-Exception-Error after removing the fallback as I already declared the route in my routes.php.

ADmad:
Fixes #14404

### DIFF
--- a/tests/TestCase/View/Helper/PaginatorHelperTest.php
+++ b/tests/TestCase/View/Helper/PaginatorHelperTest.php
@@ -58,7 +58,7 @@ class PaginatorHelperTest extends TestCase
             'url' => '/',
             'params' => [
                 'plugin' => null,
-                'controller' => '',
+                'controller' => 'Articles',
                 'action' => 'index',
             ],
         ]);
@@ -79,6 +79,7 @@ class PaginatorHelperTest extends TestCase
         $this->Paginator = new PaginatorHelper($this->View);
 
         Router::reload();
+        Router::connect('/', ['controller' => 'Articles', 'action' => 'index']);
         Router::connect('/:controller/:action/*');
         Router::connect('/:plugin/:controller/:action/*');
         Router::setRequest($request);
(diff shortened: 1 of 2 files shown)


tests/TestCase/View/Helper/PaginatorHelperTest.php
```diff
@@ -1151,7 +1152,7 @@ public function testGenerateUrlMultiplePagination()
     public function testGenerateUrlMultiplePaginationQueryStringData()
     {
         $request = new ServerRequest([
-            'url' => '/posts/index/',
+            'url' => '/posts//',
```
dereuromark: should that be one trailing slash?

ndru123 (approved):

othercorey:
Do we need to back-port this?

ADmad:
Yup the fix probably should be backported. But cherry picking the commit might cause lot of conflicts in the test file so better to directly make that one line change and update tests as required.

ADmad:
Actually this issue doesn't exist in 3.x because there query string params are not nested under `?` key array.

PR #14408, opened 2020-03-29
Add json ext as dependency as the core needs this.

dereuromark:
also fixed some IDE warnings about exceptions.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -29,6 +29,7 @@
     },
     "require": {
         "php": ">=7.2.0",
+        "ext-json": "*",
         "ext-intl": "*",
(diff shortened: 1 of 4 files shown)


ADmad (approved):

composer.json
```diff
@@ -29,6 +29,7 @@
     },
     "require": {
         "php": ">=7.2.0",
+        "ext-json": "*",
```
othercorey: Would be nice if these stayed alphabetical.

dereuromark: I dont think thats a big deal, but you can sure reorder them.

PR #14410, opened 2020-03-30
Fix error reported by phpstan

ADmad:


ADmad:
The CS build failure is due to new release of `slevomat/coding-standard`, I have opened an issue for it.

saeideng (approved):

PR #14412, opened 2020-03-31
Updated inheritDoc comments to follow formatting standard.

othercorey:
This updates a couple formatting issues:

- Docblocks that want to just inherit must use `@inheritDoc`
- Docblocks that want to extend the comment must start with `{@inheritDoc}` on a single line.
- `inheritdoc` is updated to `inheritDoc`
- All `@param` and `@return` tags must be valid when extending docblock with `{@inheritDoc}`

### DIFF
--- a/src/Collection/CollectionTrait.php
+++ b/src/Collection/CollectionTrait.php
@@ -73,9 +73,7 @@ trait CollectionTrait
     }
 
     /**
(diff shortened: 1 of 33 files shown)


markstory (commented):
Seems reasonable to me.

PR #14416, opened 2020-04-01
Fix unload command removing too much content.

### LINKED ISSUE #14414: Plugin Load/Unload Command generating extra lines
This is a (multiple allowed):

* [x] enhancement

* CakePHP Version: dabb56c latest master
* Platform and Target: n/a.

### What you did
use `bin/cake plugin load PluginName` and `bin/cake plugin unload PluginName`

### What happened
Extra empty line was created in file `Application.php` with each load/unload use

No extra lines remains in file

### LINKED ISSUE #14415: PluginUnloadCommand removing extra non empty line
This is a (multiple allowed):

* [x] bug

* CakePHP Version: dabb56c latest master
* Platform and Target: n/a.

### What you did
use `bin/cake plugin load PluginName` (refers to PluginLoadCommand from core)
Manually add `$this->addPlugin('OtherPlugin');` in very next line after added by `plugin load` `$this->addPlugin('PluginName');`
use `bin/cake plugin unload PluginName` (refers to PluginLoadCommand from core)

### What happened
both lines were removed

removed `$this->addPlugin('PluginName');` only

markstory:
The previous regex would include all plugins that followed the named plugin. The new pattern is a bit more deliberate about the structure of the parameters to unload.

Fixes #14414
Fixes #14415

### DIFF
--- a/src/Command/PluginUnloadCommand.php
+++ b/src/Command/PluginUnloadCommand.php
@@ -70,7 +70,8 @@ class PluginUnloadCommand extends Command
      */
     protected function modifyApplication(string $app, string $plugin): bool
     {
-        $finder = "@\\\$this\-\>addPlugin\(\s*'$plugin'(.|.\n|)+\);+@";
+        $finder = "/\s*\\\$this\-\>addPlugin\(\s*['\"]{$plugin}['\"](\s*,[\s\\n]*\[(\\n.*|.*){0,5}\][\\n\s]*)?\);/m";
+        // $finder = "@\\\$this\-\>addPlugin\(\s*'$plugin'(.|.\n|)+\);+@";
 
         $content = file_get_contents($app);
         $newContent = preg_replace($finder, '', $content);
--- a/src/Command/PluginUnloadCommand.php
+++ b/src/Command/PluginUnloadCommand.php
@@ -71,7 +71,6 @@ class PluginUnloadCommand extends Command
     protected function modifyApplication(string $app, string $plugin): bool
     {
         $finder = "/\s*\\\$this\-\>addPlugin\(\s*['\"]{$plugin}['\"](\s*,[\s\\n]*\[(\\n.*|.*){0,5}\][\\n\s]*)?\);/m";
-        // $finder = "@\\\$this\-\>addPlugin\(\s*'$plugin'(.|.\n|)+\);+@";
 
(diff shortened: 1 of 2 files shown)


src/Command/PluginUnloadCommand.php
```diff
@@ -70,7 +70,8 @@ public function execute(Arguments $args, ConsoleIo $io): ?int
      */
     protected function modifyApplication(string $app, string $plugin): bool
     {
-        $finder = "@\\\$this\-\>addPlugin\(\s*'$plugin'(.|.\n|)+\);+@";
+        $finder = "/\s*\\\$this\-\>addPlugin\(\s*['\"]{$plugin}['\"](\s*,[\s\\n]*\[(\\n.*|.*){0,5}\][\\n\s]*)?\);/m";
+        // $finder = "@\\\$this\-\>addPlugin\(\s*'$plugin'(.|.\n|)+\);+@";
```
ADmad, line 74: Commented code needs to be removed :)

markstory: whoops. Was having a hard time with this regex.

src/Command/PluginUnloadCommand.php
```diff
@@ -70,7 +70,7 @@ public function execute(Arguments $args, ConsoleIo $io): ?int
      */
     protected function modifyApplication(string $app, string $plugin): bool
     {
-        $finder = "@\\\$this\-\>addPlugin\(\s*'$plugin'(.|.\n|)+\);+@";
+        $finder = "/\s*\\\$this\-\>addPlugin\(\s*['\"]{$plugin}['\"](\s*,[\s\\n]*\[(\\n.*|.*){0,5}\][\\n\s]*)?\);/m";
```
othercorey, line 73: We really should start adding a short comment explaining complex regex.

markstory: I could use the x modifier and include comments if that would help.

othercorey: That probably isn't needed for this one. I would be happy with a 1-2 line comment stating what it is checking with a quick example. 

Something other than "Checking for correct format with regex" which just re-states the code.

markstory:
@yarkm13 Vendored plugin removal should be fixed now.

PR #14431, opened 2020-04-08
Further harden CSRF token comparisons

markstory:
Nirmal Kirubakaran contacted us via the security mailing list and disclosed a potential vulnerability in our CSRF token generation. Currently our CSRF token validation will accept any matching token pairs. If an attacker were to use an XSS vulnerabiity or physical access to fixate a CSRF token they could then exploit additional CSRF attacks.

This change improves token generation to use a random and hmac component to ensure that the token was originally generated by the receiving application.

I've put this for 4.0.6 as it is fix for a security related issue. However, I'm also ok moving it to 4.1.0 as it could potentially break multi-application setups where CSRF tokens are shared across applications or CakePHP versions < 4.0.6


src/Http/Middleware/CsrfProtectionMiddleware.php
```diff
@@ -223,16 +240,57 @@ protected function _validateToken(ServerRequestInterface $request): void
...
+        }
+        $postKey = (string)substr($post, 0, static::TOKEN_VALUE_LENGTH);
+        $postHmac = (string)substr($post, static::TOKEN_VALUE_LENGTH);
+        $cookieKey = (string)substr($cookie, 0, static::TOKEN_VALUE_LENGTH);
+        $cookieHmac = (string)substr($cookie, static::TOKEN_VALUE_LENGTH);
+
+        // Put all checks in a list
+        // so they all burn time reducing timing attack window.
+        $checks = [
+            hash_equals($postKey, $cookieKey),
```
othercorey, line 280: Wouldn't this always be the same if both $post and $cookie are ''?

ADmad:
> However, I'm also ok moving it to 4.1.0 as it could potentially break multi-application setups where CSRF tokens are shared across applications or CakePHP versions < 4.0.6

I think chances of this are very small. Just in case someone does have an issue, switching to their own middleware which follows old implementation would be pretty easy.

src/Http/Middleware/CsrfProtectionMiddleware.php
markstory: Yes it would. But we need to run all the same comparisons to reduce timing attack windows.

othercorey: I mean won't it be a false positive?

garas: `_validateToken()` has `if (!$cookie)` check for empty cookie.

But check for cookie/post minimal length could be done too.

othercorey: If they aren't long enough, substr can generate false which is cast to string which could be the same also.

othercorey: You would somehow need a key whose hmac matches `(string)false` which doesn't seem feasible.  Looks good to me.

PR #14435, opened 2020-04-10
Add setHelper() as better IDE method.

dereuromark:
You cannot always set all helpers via AppView and loadHelper() calls.
Sometimes you do it inside controller using view builder.
Currently, the array syntax has 0 autocomplete on the magic strings inside.

I propose to add a syntactic sugar that makes the IDE experience here way better, faster and cleaner coding. The merge overhead on a simple small list here should be irrelevant on that request level.

```php
// Before, no autocomplete :(
$this->viewBuilder()->setHelpers(['Tools.Format', 'Captcha.Captcha']);

// After, nice chaining with suggest/autocomplete
$this->viewBuilder()->setHelper('Tools.Format')
	->setHelper('Captcha.Captcha');
```

Makes it possible to have https://github.com/dereuromark/cakephp-ide-helper auto complete now. PR: https://github.com/dereuromark/cakephp-ide-helper/pull/180

See

ide_autocomplete

If we rather put this into 4.next, that's also fine.


src/View/ViewBuilder.php
```diff
@@ -301,6 +301,18 @@ public function getPlugin(): ?string
...
+    /**
+     * Set a helper to use.
+     *
+     * @param string $helper Helper to use.
+     * @param bool $merge Whether or not to merge existing data with the new data.
+     * @return $this
+     */
+    public function setHelper(string $helper, bool $merge = true)
+    {
+        return $this->setHelpers([$helper], $merge);
```
othercorey, line 313: I wish these weren't named `set` because that implies overwriting when it is adding by default.

dereuromark: Yeah, but then you would need two methods. add* and set* :)

ADmad: Why would we need add* and set*?  We could just have `addHelper(string $helpers)`. Requiring to overrite existing helpers list is pretty rare in which case one case use `setHelper()` which has `$overwrite` argument.

dereuromark: Fair enough.

dereuromark: Wouldnt then deprecate the merge flag be cleaner?
And
- addHelper()
- addHelpers()
- setHelpers()

which each only 1 arg?

ADmad:
> If we rather put this into 4.next, that's also fine.

Targeted 4.next. Even for BC changes we have decided to not add new methods in bugfix releases.

dereuromark:
I guess someone would need to merge master first for this to be mergeable.

dereuromark:
Closing then in favor of https://github.com/cakephp/cakephp/pull/14436

PR #14449, opened 2020-04-15
Add sql cast function #13694

### LINKED ISSUE #13694: Support for convert expression in database

mcsknp:
Add a method to generate sql `CAST`.
Resolves #13694

### DIFF
--- a/tests/TestCase/Database/QueryTest.php
+++ b/tests/TestCase/Database/QueryTest.php
@@ -4240,12 +4240,8 @@ class QueryTest extends TestCase
 
         // Postgres requires the case statement to be cast to a integer
         if ($this->connection->getDriver() instanceof \Cake\Database\Driver\Postgres) {
-            $publishedCase = $query->func()
-                ->cast([$publishedCase, 'integer' => 'literal'])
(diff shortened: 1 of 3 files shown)


tests/TestCase/Database/QueryTest.php
```diff
@@ -4240,12 +4240,8 @@ public function testSqlCaseStatement()
 
         // Postgres requires the case statement to be cast to a integer
         if ($this->connection->getDriver() instanceof \Cake\Database\Driver\Postgres) {
-            $publishedCase = $query->func()
-                ->cast([$publishedCase, 'integer' => 'literal'])
```
markstory, line 4244: Does this form now fail? In order to maintain backwards compatibility we should also continue to accept this call signature.

mcsknp: I added backward compatibility with deprecation warning message.

markstory:
I'll fix the CS problems up post merge.

PR #14450, opened 2020-04-15
[3.x] Add sql cast function #13694

### LINKED ISSUE #13694: Support for convert expression in database

mcsknp:
Add a method to generate sql CAST for 3.x.
Resolves #13694

### DIFF
--- a/src/Database/FunctionsBuilder.php
+++ b/src/Database/FunctionsBuilder.php
@@ -165,7 +165,7 @@ class FunctionsBuilder
      * @param string $type The target data type
      * @return \Cake\Database\Expression\FunctionExpression
      */
-    public function cast($field, string $type): FunctionExpression
+    public function cast($field, string $type)
     {
         $expression = $this->_literalArgumentFunction('CAST', $field);
         $expression->setConjunction(' AS')->add([$type => 'literal']);
--- a/src/Database/FunctionsBuilder.php
+++ b/src/Database/FunctionsBuilder.php
@@ -165,8 +165,17 @@ class FunctionsBuilder
      * @param string $type The target data type
      * @return \Cake\Database\Expression\FunctionExpression
      */
-    public function cast($field, $type)
+    public function cast($field, $type = '')
     {
+        if (is_array($field)) {
+            deprecationWarning(
+                'Build cast function by FunctionsBuilder::cast(array $args) is deprecated. ' .
+                'Use FunctionsBuilder::cast($field, string $type) instead.'
+            );
+
+            return $this->_build('CAST', $field);
+        }
+
         $expression = $this->_literalArgumentFunction('CAST', $field);
         $expression->setConjunction(' AS')->add([$type => 'literal']);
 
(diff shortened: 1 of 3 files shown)


tests/TestCase/Database/QueryTest.php
```diff
@@ -4376,12 +4376,8 @@ public function testSqlCaseStatement()
 
         //Postgres requires the case statement to be cast to a integer
         if ($this->connection->getDriver() instanceof \Cake\Database\Driver\Postgres) {
-            $publishedCase = $query->func()
-                ->cast([$publishedCase, 'integer' => 'literal'])
```
markstory, line 4380: Same feedback as the 4.next pull about backwards compatibility issue.

mcsknp: I added backward compatibility with deprecation warning message.

src/Database/FunctionsBuilder.php
```diff
@@ -158,6 +158,30 @@ public function coalesce($args, $types = [])
...
     }
 
+    /**
+     * Returns a FunctionExpression representing a call to SQL CAST function.
+     *
+     * @param string|\Cake\Database\ExpressionInterface $field Field or expression to cast.
+     * @param string $type The target data type
+     * @return \Cake\Database\Expression\FunctionExpression
+     */
+    public function cast($field, $type = '')
```
othercorey, line 168: Why is `$type` defaulted to ''?

src/Database/FunctionsBuilder.php
```diff
@@ -158,6 +158,30 @@ public function coalesce($args, $types = [])
...
+        if (is_array($field)) {
+            deprecationWarning(
+                'Build cast function by FunctionsBuilder::cast(array $args) is deprecated. ' .
+                'Use FunctionsBuilder::cast($field, string $type) instead.'
+            );
+
+            return $this->_build('CAST', $field);
+        }
+
+        $expression = $this->_literalArgumentFunction('CAST', $field);
```
othercorey, line 179: Shouldn't this set the return type to $type ?

src/Database/FunctionsBuilder.php
mcsknp: Because of backwards compatibility. Function is called with only an array in first argument.

othercorey: I think we should check that $type was passed as an argument when $field is not an array.

src/Database/FunctionsBuilder.php
mcsknp: No, because $type is SQL type.

othercorey: Well the return type defaults to string. We should probably provide a way to set it.

mcsknp: I didn't found where the return type used. If it's required, I will add an arguments to set it.

src/Database/FunctionsBuilder.php
mcsknp: Firstly I think so too, but I didn't add this check because of 2 reason:
1. Both of check or no check will return an exception.
2. The default value for its will be removed when remove deprecation.

PR #14454, opened 2020-04-15
Deprecate false/string for FormHelper::create() to ease upgrade pain.

dereuromark:
Takes care of https://github.com/cakephp/cakephp/issues/13493#issuecomment-570640461

### DIFF
--- a/src/View/Helper/FormHelper.php
+++ b/src/View/Helper/FormHelper.php
@@ -397,7 +397,7 @@ class FormHelper extends Helper
      *
      * @param mixed $context The context for which the form is being defined.
      *   Can be a ContextInterface instance, ORM entity, ORM resultset, or an
-     *   array of meta data. You can use false or null to make a context-less form.
+     *   array of meta data. You can use null to make a context-less form.
      * @param array $options An array of html attributes and options.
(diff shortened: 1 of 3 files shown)


src/View/Helper/FormHelper.php
```diff
@@ -406,6 +406,10 @@ public function create($context = null, array $options = [])
     {
         $append = '';
 
+        if (is_bool($context) || is_string($context)) {
+            deprecationWarning('Using `string` or `bool` for $context is deprecated, use `null` to make a context-less form.');
```
ADmad, line 410: A string could be a valid context if someone has a custom form context class for it.

dereuromark: I didnt see tests that would match this.

dereuromark: But since this is still just a deprecation help for the "fatal error" you get on 4.0 otherwise, even if that use case exists, this should be noted as deprecated, right? As this helps to upgrade to a 4.0 compatible way.

markstory: I agree that the deprecation is helpful and could help improve someone's upgrade experience.

PR #14457, opened 2020-04-15
Debug a var when you user REST api methodology

lpj145:
This improve a lot when you simple die debug on api.

In most case you need to die and debug a var when you user REST api methodology, actually the entire response stream is ignored, now, this simple make die with some headers to test vars.

### DIFF
--- a/src/basics.php
+++ b/src/basics.php
@@ -142,7 +142,13 @@ if (!function_exists('dd')) {
             'line' => $trace[0]['line'],
             'file' => $trace[0]['file'],
         ];
-
+        
+        $origin = substr($_SERVER['HTTP_REFERER'], 0, (strlen($_SERVER['HTTP_REFERER']) - 1));
+        $headers = ['authorization', 'content-type', 'api-token'];
+        header('Access-Control-Allow-Credentials: true');
+        header('Access-Control-Allow-Headers: '.implode($headers, ' '));
+        header('Access-Control-Allow-Origin:'.str_replace('\/', '', $origin));
+        
         Debugger::printVar($var, $location, $showHtml);
         die(1);
     }
--- a/src/basics.php
+++ b/src/basics.php
@@ -144,7 +144,8 @@ if (!function_exists('dd')) {
         ];
 
         if (PHP_SAPI !== 'cli') {
-            $origin = substr($_SERVER['HTTP_REFERER'], 0, strlen($_SERVER['HTTP_REFERER'] - 1));
+            $referer = env('HTTP_SERVER', '');
+            $origin = substr($referer, 0, (strlen($referer) - 1));
             $headers = ['authorization', 'content-type', 'api-token'];
             header('Access-Control-Allow-Credentials: true');
             header(sprintf('Access-Control-Allow-Headers: %s', implode(' ', $headers)));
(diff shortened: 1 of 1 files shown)


src/basics.php
```diff
@@ -142,7 +142,13 @@ function dd($var, $showHtml = null): void
             'line' => $trace[0]['line'],
             'file' => $trace[0]['file'],
         ];
-
+        
+        $origin = substr($_SERVER['HTTP_REFERER'], 0, (strlen($_SERVER['HTTP_REFERER']) - 1));
```
dereuromark, line 146: I think this would be `env('HTTP_REFERER')`

src/basics.php
```diff
@@ -142,7 +142,13 @@ function dd($var, $showHtml = null): void
...
+        $headers = ['authorization', 'content-type', 'api-token'];
+        header('Access-Control-Allow-Credentials: true');
+        header('Access-Control-Allow-Headers: '.implode($headers, ' '));
+        header('Access-Control-Allow-Origin:'.str_replace('\/', '', $origin));
```
dereuromark, line 150: I dont think header output are a good idea inside such a generic helper function, especially since this is also used in CLI context.

src/basics.php
lpj145: maybe all moments you need to debug, you are on dev env, and not need verify referer.

src/basics.php
lpj145: make sense, need to verify SAPI env.

src/basics.php
dereuromark: `$_SERVER['HTTP_REFERER']` could not be set and would trigger a notice, right?

src/basics.php
```diff
@@ -143,6 +143,15 @@ function dd($var, $showHtml = null): void
             'file' => $trace[0]['file'],
         ];
 
+        if (PHP_SAPI !== 'cli') {
+            $referer = env('HTTP_REFERER') ?? '';
```
dereuromark, line 147: $referer = (string)env('HTTP_REFERER');

suffices

src/basics.php
```diff
@@ -143,6 +143,15 @@ function dd($var, $showHtml = null): void
             'file' => $trace[0]['file'],
         ];
 
+        if (PHP_SAPI !== 'cli') {
```
dereuromark, line 146: You could add here

    && !empty($_SERVER['HTTP_REFERER'])

this way you don't need env() usage either.

src/basics.php
lpj145: next pr, i make from ide, worrying about this.

src/basics.php
```diff
@@ -143,6 +143,18 @@ function dd($var, $showHtml = null): void
             'file' => $trace[0]['file'],
         ];
 
+       if (
+            PHP_SAPI !== 'cli'
+            && $_SERVER['HTTP_REFERER']
```
dereuromark, line 148: There was no need to make this multiline.
Best to use an IDE, the commits are getting out of hand a bit here.

dereuromark: Also, this does not meet the comment I gave, about using !empty() to avoid the undefined index notice.

PR #14458, opened 2020-04-15
Debug a var when you user REST api methodology

lpj145:
This make more sense, now, tests are already and ok.

### DIFF
--- a/src/basics.php
+++ b/src/basics.php
@@ -143,6 +143,15 @@ if (!function_exists('dd')) {
             'file' => $trace[0]['file'],
         ];
 
+        if (PHP_SAPI !== 'cli' && !empty($_SERVER['HTTP_REFERER'])) {
+            $referer = $_SERVER['HTTP_REFERER'];
+            $origin = substr($referer, 0, strlen($referer) - 1);
+            $headers = ['authorization', 'content-type', 'api-token'];
+            header('Access-Control-Allow-Credentials: true');
+            header(sprintf('Access-Control-Allow-Headers: %s', implode(' ', $headers)));
+            header(sprintf('Access-Control-Allow-Origin: %s', $origin));
+        }
+
         Debugger::printVar($var, $location, $showHtml);
         die(1);
     }


dereuromark (approved):
Nice, ok
If there are no drawbacks I don't see why we shouldn't add this snippet.

markstory:
I'm not sure I understand why debug output needs to set CORS headers.

src/basics.php
```diff
@@ -143,6 +143,15 @@ function dd($var, $showHtml = null): void
             'file' => $trace[0]['file'],
         ];
 
+        if (PHP_SAPI !== 'cli' && !empty($_SERVER['HTTP_REFERER'])) {
+            $referer = $_SERVER['HTTP_REFERER'];
+            $origin = substr($referer, 0, strlen($referer) - 1);
+            $headers = ['authorization', 'content-type', 'api-token'];
+            header('Access-Control-Allow-Credentials: true');
```
markstory, line 150: You should also check to make sure that headers haven't already been sent. Not doing so will emit additional warnings.

ADmad:
> I'm not sure I understand why debug output needs to set CORS headers.

Me too. Why not use a middleware to always set those headers when debug is enabled?

lpj145:
> I'm not sure I understand why debug output needs to set CORS headers.



> > I'm not sure I understand why debug output needs to set CORS headers.
> 
> Me too. Why not use a middleware to always set those headers when debug is enabled?

Of course, if headers is already emitted, but is not this case, this case all **ENTIRE** stream is put off, and a simple die with some vardumps are flush on stream, but maybe you cannot understand this case, close pr, i'm happy to make wrapper for this function on my projects.

dereuromark:
He is right that the dd() before the already processed middleware wouldn't have any effect.

lpj145:
> He is right that the dd() before the already processed middleware wouldn't have any effect.
> Or if just added to the response but not using it as here.

Can use a main stream, make a exception, make middleware, get the exception by the way and return normal stream, because i already added CORS headers, is works right, but, i'm only need dd var 😢 😸

ADmad:
This seems like an edge case usecase and I don't think these changes would be needed by enough users to warrant adding it in core. Adding unneeded headers to every debug out seems unnecessary. 

Also one can easily add their own `dd()` function is app's bootstrap and that function would be used instead of the one provided by CakePHP core. Hence closing this issue.

PR #14463, opened 2020-04-16
Add locale path into BasePlugin config.

dereuromark:
```
public static function path(string $type, ?string $plugin = null): array
```
Docblock states:
```
     * Used to read information stored path.
     *
     * The 1st character of $type argument should be lower cased and will return the
     * value of `App.paths.$type` config.
     *
     * Default types:
     * - plugins
     * - templates
     * - locales
```

locales here, however, are different from the others, the only one that doesn't properly work for plugins as 2nd arg:
```php
    $paths = App::path('locales'); // OK
    $paths = App::path('locales', $plugin); // broken
```

This fixes it for 4.1+.

We don't necessarily have to go full config here, we could also let the BasePlugin read by convention from the resources/locales/ folder without the check of the property.
But I figured it ain't that much overhead after all.

### DIFF
--- a/src/Core/App.php
+++ b/src/Core/App.php
@@ -198,6 +198,10 @@ class App
             /** @psalm-suppress PossiblyNullArgument */
             return [Plugin::templatePath($plugin)];
         }
+        if ($type === 'locales') {
+            /** @psalm-suppress PossiblyNullArgument */
+            return [Plugin::localePath($plugin)];
+        }
 
         deprecationWarning(
             'App::path() is deprecated for class path.'
--- a/src/Core/BasePlugin.php
+++ b/src/Core/BasePlugin.php
@@ -86,6 +86,13 @@ class BasePlugin implements PluginInterface
     protected $templatePath;
 
     /**
+     * The templates path for this plugin.
+     *
+     * @var string
+     */
+    protected $localePath;
+
+    /**
      * The name of this plugin
      *
      * @var string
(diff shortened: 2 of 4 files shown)


ADmad:
The docblock is wrong here. Instead of `- locales` it needs to be `- resources`. One can then just 

I see no need for all the additional methods to get locales folder path. The locales folder is required for only 1 specific implementation of translation messages loader.

If we really want to add extra methods for plugins then it should be a more general `resourcesPath()`.

dereuromark:
I see, so we fix the docblocks instead in master then.

dereuromark:
Still, `App::path('locales');` is used all over the core, and as such one would expect it to not break with 2nd argument plugin.

ADmad:
Hmm actually app config does set `App.paths.locales` https://github.com/cakephp/app/blob/c703461a3859c396ef29189e915234ba85d80211/config/app.php#L67.

You are right that using `App::path('locales')` with 2nd argument will not work currently. That could be handled by doing the same that's done in `MessageFileLoader`

https://github.com/cakephp/cakephp/blob/8718b2e9bbd092b991a1da6472d1a28e89b63dac/src/I18n/MessagesFileLoader.php#L176

dereuromark:
I am fine either way. I just need a reliable API to work with in tooling ;)

ADmad:
Okay, I'll open PR for `master` itself. I don't think using `App::path('Locale')` with 2nd argument will work in 3.x either.

markstory:
Closing as #14464 was merged.

PR #14465, opened 2020-04-16
Fix errors reported by psalm

ADmad:

### DIFF
--- a/src/Collection/CollectionInterface.php
+++ b/src/Collection/CollectionInterface.php
@@ -69,7 +69,7 @@ interface CollectionInterface extends Iterator, JsonSerializable
      * @param callable|null $c the method that will receive each of the elements and
      *   returns true whether or not they should be in the resulting collection.
      *   If left null, a callback that filters out falsey values will be used.
-     * @return \Cake\Collection\CollectionInterface
+     * @return self
      */
     public function filter(?callable $c = null): CollectionInterface;
 
@@ -94,7 +94,7 @@ interface CollectionInterface extends Iterator, JsonSerializable
      *
      * @param callable $c the method that will receive each of the elements and
      * returns true whether or not they should be out of the resulting collection.
-     * @return \Cake\Collection\CollectionInterface
+     * @return self
      */
     public function reject(callable $c): CollectionInterface;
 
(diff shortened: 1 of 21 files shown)


src/Collection/CollectionInterface.php
```diff
@@ -44,7 +44,7 @@ interface CollectionInterface extends Iterator, JsonSerializable
      * in this collection
      * @return $this
      */
-    public function each(callable $c): CollectionInterface;
+    public function each(callable $c);
```
ADmad, line 47: This is technically a BC break but I highly doubt there's any custom implementation of `CollectionInterface` out there.

I can instead just ignore the error if that's preferred by everyone.

othercorey: What error requires this change? Is @return $this not valid anymore?

dereuromark: Adding return types on chainable is just bad. It breaks.

dereuromark: I dont think removing return types is a BC break, only adding them would be.
If extended classes would have that defined, it would still be just fine.

ADmad: @othercorey This is the error shown with current code.

```
ERROR: LessSpecificImplementedReturnType - src/Collection/CollectionTrait.php:66:40 - The inherited return type 'Cake\Collection\CollectionInterface&static' for Cake\Collection\CollectionInterface::each is more specific than the implemented return type for Cake\Collection\CollectionTrait::each 'Cake\Collection\CollectionInterface' (see https://psalm.dev/166)
    public function each(callable $c): CollectionInterface
```

othercorey: I think that's because the doc block for CollectionInterface::each() is `@return $this` instead of `@return \Cake\Collection\CollectionInterface` like all the other functions.

ADmad: Well the implementation in `CollectionTrait::each()` does return `$this` hence `@return $this`. Unlike most other collection methods `each()` return the same collection instead of a new one.

othercorey: I saw that too. I think there's some nonsense with psalm going on. Still, we could probably get around it by adjusting the @return.

dereuromark: Yes. Chainable must not be return typehinted.
I wonder why didnt add a cs rule yet. I could offer one.

PR #14466, opened 2020-04-16
Merge 3.x into 3.next

dereuromark:


.appveyor.yml
```diff
@@ -8,9 +8,9 @@ cache:
 
 branches:
   only:
-    - master
+    - 3.x
     - 3.next
-    - 4.x
```
saeideng, line 13: should not we keep the `master` ?

dereuromark: i dont think so for 3.x ones.

markstory: Appveyor might have only one configuration for the project.

PR #14470, opened 2020-04-19
4.next - Port 14016 to 4.1

markstory:
Port the changes from #14016 to 4.x.

### DIFF
--- a/src/I18n/DateFormatTrait.php
+++ b/src/I18n/DateFormatTrait.php
@@ -19,6 +19,7 @@ namespace Cake\I18n;
 use Cake\Chronos\Date as ChronosDate;
 use Cake\Chronos\DifferenceFormatterInterface;
 use Cake\Chronos\MutableDate;
+use Closure;
 use DateTime;
 use DateTimeZone;
 use IntlDateFormatter;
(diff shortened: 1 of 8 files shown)


src/I18n/DateFormatTrait.php
```diff
@@ -48,6 +48,22 @@ trait DateFormatTrait
...
+     * as specified in (http://www.icu-project.org/apiref/icu4c/classSimpleDateFormat.html#details)
+     *
+     * It is possible to provide an array of 2 constants. In this case, the first position
+     * will be used for formatting the date part of the object and the second position
+     * will be used to format the time part.
+     *
+     * @var string|array|int|callable
+     * @see \Cake\I18n\Time::i18nFormat()
+     */
+    protected static $_jsonEncodeFormat = "yyyy-MM-dd'T'HH':'mm':'ssxxx";
```
othercorey, line 65: I'm not sure this should exist in DateTimeFormat because there is no common format. The @var type needs to be updated for all the Date and Time classes.

src/I18n/DateFormatTrait.php
```diff
@@ -276,8 +292,11 @@ public static function setToStringFormat($format): void
      * will be used for formatting the date part of the object and the second position
      * will be used to format the time part.
      *
+     * Alternatively, the format can provide a callback. In this case, the callback
```
othercorey, line 295: This should be moved to I18nDateTimeInterface then @inheritDoc instead.

src/I18n/DateFormatTrait.php
othercorey: I would like to have this in a base trait/class so we can just @inheritDoc it everywhere. I suppose this should be fine as every date/time type should override it.

src/I18n/DateFormatTrait.php
```diff
@@ -48,6 +48,22 @@ trait DateFormatTrait
...
+     *
+     * The format should be either the formatting constants from IntlDateFormatter as
+     * described in (https://secure.php.net/manual/en/class.intldateformatter.php) or a pattern
+     * as specified in (http://www.icu-project.org/apiref/icu4c/classSimpleDateFormat.html#details)
+     *
+     * It is possible to provide an array of 2 constants. In this case, the first position
+     * will be used for formatting the date part of the object and the second position
+     * will be used to format the time part.
+     *
+     * @var string|array|int|callable
```
ADmad, line 62: Can we limit it to just `Closure` instead of `callable`?

markstory: Sure.

src/I18n/DateFormatTrait.php
markstory: Going to remove this as the property is defined on each class that implements this trait.

dereuromark:
Do we need to readjust the 3.next merge?

PR #14473, opened 2020-04-19
Fix class name.

ADmad:


dereuromark (approved):

phpstan.neon
```diff
@@ -28,6 +28,7 @@ parameters:
         - '#Binary operation "\+" between array|false and array results in an error#'
         - '#Result of method Cake\\Auth\\BaseAuthenticate::unauthenticated\(\) \(void\) is used#'
         - '#Call to an undefined method DateTimeInterface::setTimezone\(\)#'
+        - '#Class APCuIterator referenced with incorrect case: APCUIterator#'
```
dereuromark, line 31: Is the tooling confused here?

ADmad: Yeah, it's a bug in `v0.11` of phpstan. [APCUIterator](https://www.php.net/manual/en/class.apcuiterator.php) is the correct casing.

PR #14474, opened 2020-04-19
4.next path url

dereuromark:
Follow https://github.com/cakephp/cakephp/pull/13605

Uses https://github.com/dereuromark/cakephp-ide-helper/pull/191#issuecomment-616225774 to have template support for new route path strings.

Currently, Router::pathUrl() works fine already, but it has like 1% usage inside most apps.
The other 99% are inside the templates, and there having the chance to leverage those now as autocompletable strings is really a game changer.

See
ide_url
and
ide_html

The overhead here is minimal, but the benefit of having this more strict syntactic sugar wrapper and the full IDE support behind it makes coding/developing 10x as fast, especially if you need to re-use similar actions a lot, and if those happen to be plugins or prefixed even more so.

### DIFF
--- a/src/View/Helper/HtmlHelper.php
+++ b/src/View/Helper/HtmlHelper.php
@@ -309,14 +309,14 @@ class HtmlHelper extends Helper
      *   over value of `escape`)
      * - `confirm` JavaScript confirmation message.
      *
-     * @param string|array $title The content to be wrapped by `<a>` tags.
-     *   Can be an array if $url is null. If $url is null, $title will be used as both the URL and title.
+     * @param string $title The content to be wrapped by `<a>` tags.
      * @param string $path Cake-relative route path.
      * @param array $options Array of options and HTML attributes.
      * @return string An `<a />` element.
+     * @see \Cake\Routing\Router::pathUrl()
      * @link https://book.cakephp.org/3/en/views/helpers/html.html#creating-links
      */
-    public function linkFromPath($title, string $path, array $options = []): string
+    public function linkFromPath(string $title, string $path, array $options = []): string
     {
         return $this->link($title, ['_path' => $path], $options);
     }
--- a/src/View/Helper/HtmlHelper.php
+++ b/src/View/Helper/HtmlHelper.php
@@ -311,14 +311,16 @@ class HtmlHelper extends Helper
      *
      * @param string $title The content to be wrapped by `<a>` tags.
      * @param string $path Cake-relative route path.
+     * @param array $params An array specifying any additional parameters.
+     *   Can be also any special parameters supported by `Router::url()`.
      * @param array $options Array of options and HTML attributes.
      * @return string An `<a />` element.
      * @see \Cake\Routing\Router::pathUrl()
      * @link https://book.cakephp.org/3/en/views/helpers/html.html#creating-links
      */
-    public function linkFromPath(string $title, string $path, array $options = []): string
+    public function linkFromPath(string $title, string $path, array $params = [], array $options = []): string
     {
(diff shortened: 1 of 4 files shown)


src/View/Helper/HtmlHelper.php
```diff
@@ -299,6 +299,28 @@ public function link($title, $url = null, array $options = []): string
...
+     *
+     * ### Options
+     *
+     * - `escape` Set to false to disable escaping of title and attributes.
+     * - `escapeTitle` Set to false to disable escaping of title. Takes precedence
+     *   over value of `escape`)
+     * - `confirm` JavaScript confirmation message.
+     *
+     * @param string|array $title The content to be wrapped by `<a>` tags.
+     *   Can be an array if $url is null. If $url is null, $title will be used as both the URL and title.
```
ADmad, line 313: This line probably needs to be removed. There's no `$url` argument.

src/View/Helper/HtmlHelper.php
```diff
@@ -299,6 +299,28 @@ public function link($title, $url = null, array $options = []): string
...
+     * @param string $path Cake-relative route path.
+     * @param array $options Array of options and HTML attributes.
+     * @return string An `<a />` element.
+     * @link https://book.cakephp.org/3/en/views/helpers/html.html#creating-links
+     */
+    public function linkFromPath($title, string $path, array $options = []): string
```
ADmad, line 319: Should `$title` be type hinted as `string`?

src/View/Helper/UrlHelper.php
```diff
@@ -57,6 +57,24 @@ public function build($url = null, array $options = []): string
...
+     * ### Options:
+     *
+     * - `escape`: If false, the URL will be returned unescaped, do only use if it is manually
+     *    escaped afterwards before being displayed.
+     * - `fullBase`: If true, the full base URL will be prepended to the result
+     *
+     * @param string $path Cake-relative route path.
+     * @param array $options Array of options.
+     * @return string Full translated URL with base path.
+     */
```
ADmad, line 72: A `@see` tag pointing to `Router::pathUrl()` would be nice so users can easily check what are route paths. Same for the new method in HtmlHelper too.

src/View/Helper/HtmlHelper.php
```diff
@@ -299,6 +299,28 @@ public function link($title, $url = null, array $options = []): string
...
+     * @param string $title The content to be wrapped by `<a>` tags.
+     * @param string $path Cake-relative route path.
+     * @param array $options Array of options and HTML attributes.
+     * @return string An `<a />` element.
+     * @see \Cake\Routing\Router::pathUrl()
+     * @link https://book.cakephp.org/3/en/views/helpers/html.html#creating-links
+     */
+    public function linkFromPath(string $title, string $path, array $options = []): string
+    {
+        return $this->link($title, ['_path' => $path], $options);
```
markstory, line 321: How would links that need parameters work?

dereuromark: They would not use this method. They would use normal one.
This is mainly the syntactic sugar entry point for the 70% use case where this applies and where you can use the above speed up tooling as well.

Or do we have an idea how to append params or query strings in such a string, or through options?

dereuromark: We could also just add the $params arg into them. One arg more wouldn't be a big issue.

dereuromark:
I adjusted it as per your feedback, and in sync with what Router does as well:

```php
// in Router
return static::url(['_path' => $path] + $params, $full);
```

dereuromark:
The stan failures seem to be unrelated and already fixed on 4.next branch.

PR #14476, opened 2020-04-20
Test reply to mailsentwith

### LINKED ISSUE #14475: EmailTrait test for replyTo and sender headers
This is a (multiple allowed):

* [x] bug
* [x] enhancement

* CakePHP Version: 4.0

### What you did
Trying to test the reply-to and sender headers

### What happened
`Cake\TestSuite\Constraint\Email`MailSentWith::matches` return an array for this headers but only allow string in `expected`

Erwane:
Resolves #14475


dereuromark:
If this is a bugfix, and it looks like it, shouldn't this go into master?

Erwane:
If you prefer, i can create the PR for master.

PR #14477, opened 2020-04-20
Fix mail headers sender replyto tests

### LINKED ISSUE #14475: EmailTrait test for replyTo and sender headers

Erwane:
Resolves #14475


markstory:
Thank you 👍

PR #14479, opened 2020-04-20
Added supportsWindowFunctions() to replace env variable

othercorey:

### DIFF
--- a/src/Database/Driver/Mysql.php
+++ b/src/Database/Driver/Mysql.php
@@ -76,6 +76,13 @@ class Mysql extends Driver
     protected $_supportsNativeJson;
 
     /**
+     * Whether or not the connected server supports window functions.
+     *
+     * @var bool
+     */
+    protected $_supportsWindowFunctions;
+
+    /**
      * String used to start a database identifier quoting to make it safe
      *
      * @var string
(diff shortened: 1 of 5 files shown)


src/Database/Driver/Sqlite.php
```diff
@@ -62,6 +62,20 @@ class Sqlite extends Driver
...
+     * The connected server version.
+     *
+     * @var string
+     */
+    protected $_version;
+
+    /**
+     * Whether or not the connected server supports windows.
+     *
+     * @var bool
```
dereuromark, line 75: Those seem to be |null here

src/Database/Driver/Mysql.php
```diff
@@ -75,6 +75,13 @@ class Mysql extends Driver
      */
     protected $_supportsNativeJson;
 
+    /**
+     * Whether or not the connected server supports windows.
+     *
+     * @var bool
+     */
+    protected $_supportsWindows;
```
ADmad, line 83: Can we name this `$_supportsWindowFunctions` instead?

dereuromark: I was also like "wuuut?" :P

othercorey: Sure. I was just using the technical name for the expression, but can be confusing.

markstory:
Nice work :clap:

PR #14481, opened 2020-04-21
Wrong type declaration in DocBlock

tobse:
IDE-Typehinting proposes value of type "string". It should be of type "array".


dereuromark:
Check again: it says array of strings.

Tobse:
Oh, sorry. I havent seen that before... I tought it was a typo.

Issue #14482, opened 2020-04-21
Entity data passed through an event returns notice when accessed but data is present


burzum:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.8.x
* Platform and Target: Win 10, php 7.4

### What you did

Created a test (put the code below in EventTest.php) to to show the problem we came accross in our app:

The issue is that when passing an entity to the event it gets converted to an array, so far so good. But when you try to access it as an array, you get a undefined index notice.

![image](https://user-images.githubusercontent.com/162789/79839971-8dd3ab00-83b5-11ea-84b7-86e64eeb1029.png)

Remove the dd() to get to this point. If you run the code as shown below, you'll see it checks if the author is emtpy and if it is, it will do the dd() - OK now check the output...

![image](https://user-images.githubusercontent.com/162789/79839599-07b76480-83b5-11ea-9dc1-218ac630dfc1.png)

I think this is actually not an issue of the event but the convenient but prone to failure "magic" of the internals of the entity object. Maybe it is also a php bug because $data is a real array and author still an object implementing array access.

```php
    public function testPassingEntityThroughEventBehavior()
    {
        $manager = EventManager::instance();
        $articles = TableRegistry::getTableLocator()->get('Articles');
        $articles->belongsTo('Authors');
        $article = $articles->find()
            ->contain([
                'Authors'
            ])
            ->first();

        $manager->on('TestEntity', function ($event) {
            $data = $event->getData();

            if (empty($data['author'])) {
                dd($data);
            }

            echo $data['author']['name'];
        });

        $manager->dispatch(new Event('TestEntity', null, $article));
    }
```

### What happened
Getting a notice for an undefined array index but the data is clearly there.

### What you expected to happen
No notice.

othercorey:
If the event data implements `ArrayAccess`, we could skip the conversion to array in the constructor and `getData()` method.

dereuromark:
This looks like a PHP issue of not being able to traverse through a mix of array and ArrayAccess.
So @othercorey s solution sounds like a good idea to me if that solves it.

ADmad:
I can't reproduce this issue with plain PHP https://3v4l.org/obeOH.

dereuromark:
@ADmad That is because you are not using mixed array and ArrayAccess. The bug only exists for this specific case, and because the event is casting to array, but only the root element is effected, not recursive.

[...]

PR #14486, opened 2020-04-21
Don't convert Entity data if instance of ArrayAccess

othercorey:
Fixes https://github.com/cakephp/cakephp/issues/14482

Issues:

- Event::setData() will operate on the ArrayAccess instance

  Should we create a copy?

- IterateAggregate might provide a valid array conversion. 

  Not sure if we care about ArrayObject.

### DIFF
--- a/src/Event/Event.php
+++ b/src/Event/Event.php
@@ -16,7 +16,9 @@ declare(strict_types=1);
  */
 namespace Cake\Event;
 
+use ArrayAccess;
 use Cake\Core\Exception\Exception;
+use InvalidArgumentException;
 
 /**
  * Class Event
@@ -43,7 +45,7 @@ class Event implements EventInterface
     /**
      * Custom data for the method that receives the event
      *
-     * @var array
+     * @var array|\ArrayAccess
      */
     protected $_data;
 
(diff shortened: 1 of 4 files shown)


dereuromark:
If we want to make this change without a clone, we should better put this into 4.next as this changes behavior now due to the mutable character of the input as you pointed out.

ADmad:
> and a recursive cast here

PHP won't do a recursive cast.

ADmad:
This problem has arisen because we assumed/forced event data to be an array in the past and have `getData()`, `setData()` which allow getting/setting individual keys.

We missed the opportunity in 4.0 to do a clean break and just store and return whatever is passed to the constructor without forcing/assuming it's type.

markstory:
Event data has historically been expected to be an associative map of keys and values. For the linked issue the author could do `dispatch('event.name', $this, ['entity' => $thing])` and not have an issue.

othercorey:
> Event data has historically been expected to be an associative map of keys and values. For the linked issue the author could do `dispatch('event.name', $this, ['entity' => $thing])` and not have an issue.

True. Perhaps instead of this work-around, we should have changed Event to enforce the array requirement so they don't end up with mangled data.

othercorey:
I pushed an alternative fix to discuss.

- Changed default event data to `[]`
- Require event data to be an array or ArrayAccess. No casting.
- Throw exceptions when invalid event data is set.

src/Event/Event.php
```diff
@@ -172,22 +179,26 @@ public function getData(?string $key = null)
...
      * @param mixed $value The value to set.
      * @return $this
+     * @throws \InvalidArgumentException When $key is not a string, array or does not implement \ArrayAccess.
      */
     public function setData($key, $value = null)
     {
-        if (is_array($key)) {
-            $this->_data = $key;
-        } else {
+        if (is_string($key)) {
```
ADmad, line 195: While it would be unwise one could use integer/numeric keys too.

src/Event/EventDispatcherInterface.php
```diff
@@ -33,13 +33,13 @@ interface EventDispatcherInterface
...
      * @param string $name Name of the event.
-     * @param array|null $data Any value you wish to be transported with this event to
+     * @param array $data Any value you wish to be transported with this event to
      * it can be read by listeners.
      * @param object|null $subject The object that this event applies to
      * ($this by default).
      * @return \Cake\Event\EventInterface
      */
-    public function dispatchEvent(string $name, ?array $data = null, ?object $subject = null): EventInterface;
+    public function dispatchEvent(string $name, array $data = [], ?object $subject = null): EventInterface;
```
ADmad, line 42: Changing interface method signature would be technically a BC break.

src/Event/Event.php
othercorey: True, but the $key typehint was already `string` so I used that.

src/Event/EventDispatcherInterface.php
othercorey: Yea, but this shows what a better fix is. The default of null was a very odd choice which requires the array cast.

PR #14490, opened 2020-04-22
[4.next] Add table() convenience function for CakePHP REPL

ravage84:
This adds the ``table()`` convenience function for the [CakePHP REPL](https://book.cakephp.org/4/en/console-commands/repl.html).

Instead of needing to do ``Cake\ORM\TableRegistry::getTableLocator()->get('TableName')`` to fetch a table in the CakePHP REPL

````
bin/cake console

>>> $articles = Cake\ORM\TableRegistry::getTableLocator()->get('Articles');
// object(Cake\ORM\Table)(
//
// )
>>> $articles->find()->all();
````

One can do ``table('TableName')``

````
bin/cake console

>>> $articles = table('Articles');
// object(Cake\ORM\Table)(
//
// )
>>> $articles->find()->all();
````

I tried to conditionally define the new function when the Psy shell class is available, but this makes it impossible test it, as this class does not seem to be available during tests.

```` php
if (!function_exists('table') && (PHP_SAPI === 'cli' || PHP_SAPI === 'phpdbg') && class_exists('\Psy\Shell')) {
````

This was inspired by @steinkel's CakePHP training session.

### DIFF
--- a/src/basics.php
+++ b/src/basics.php
@@ -17,9 +17,6 @@ declare(strict_types=1);
 
 use Cake\Core\Configure;
 use Cake\Error\Debugger;
-use Cake\ORM\Table;
-use Cake\ORM\TableRegistry;
-use Psy\Shell as PsyShell;
 
 define('SECOND', 1);
 define('MINUTE', 60);
@@ -63,7 +60,6 @@ if (!function_exists('debug')) {
 
         return $var;
     }
-
 }
 
 if (!function_exists('stackTrace')) {
@@ -92,50 +88,6 @@ if (!function_exists('stackTrace')) {
         /** @var string $trace */
         $trace = Debugger::trace($options);
         echo $trace;
-    }
-
-}
-
-if (!function_exists('breakpoint')) {
-    /**
-     * Command to return the eval-able code to startup PsySH in interactive debugger
-     * Works the same way as eval(\Psy\sh());
-     * psy/psysh must be loaded in your project
-     *
-     * @link http://psysh.org/
-     * ```
-     * eval(breakpoint());
-     * ```
-     * @return string|null
-     */
-    function breakpoint(): ?string
-    {
-        if ((PHP_SAPI === 'cli' || PHP_SAPI === 'phpdbg') && class_exists(PsyShell::class)) {
-            return 'extract(\Psy\Shell::debug(get_defined_vars(), isset($this) ? $this : null));';
-        }
-        trigger_error(
-            'psy/psysh must be installed and you must be in a CLI environment to use the breakpoint function',
-            E_USER_WARNING
-        );
-
-        return null;
-    }
-}
-
-if (!function_exists('table') && (PHP_SAPI === 'cli' || PHP_SAPI === 'phpdbg')) {
-    /**
-     * REPL convenience function to load a table
-     *
-     * Do **NOT** use this function in your app or plugin code,
-     * it is meant only for use in the CakePHP interactive console (REPL).
-     *
-     * @param string $alias The alias name you want to get.
-     * @param array $options The options you want to build the table with.
-     * @return \Cake\ORM\Table
-     */
-    function table(string $alias, array $options = []): Table
-    {
-        return TableRegistry::getTableLocator()->get($alias, $options);
     }
 }
 
--- a/src/basics.php
+++ b/src/basics.php
@@ -126,6 +126,9 @@ if (!function_exists('table') && (PHP_SAPI === 'cli' || PHP_SAPI === 'phpdbg')) 
     /**
      * REPL convenience function to load a table
      *
+     * Do **NOT** use this function in your app or plugin code,
+     * it is meant only for use in the CakePHP interactive console (REPL).
+     *
      * @param string $alias The alias name you want to get.
      * @param array $options The options you want to build the table with.
      * @return \Cake\ORM\Table
--- a/src/basics.php
+++ b/src/basics.php
@@ -17,6 +17,8 @@ declare(strict_types=1);
 
 use Cake\Core\Configure;
 use Cake\Error\Debugger;
+use Cake\ORM\Table;
+use Cake\ORM\TableRegistry;
 use Psy\Shell as PsyShell;
 
 define('SECOND', 1);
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: src/basics.php [hunk 1]
```diff
@@ -120,6 +120,20 @@ function breakpoint(): ?string
...
 
+if (!function_exists('table') && (PHP_SAPI === 'cli' || PHP_SAPI === 'phpdbg') ) {
+    /**
+     * REPL convenience function to load a table
+     *
+     * @param string $alias The alias name you want to get.
+     * @param array $options The options you want to build the table with.
+     * @return \Cake\ORM\Table
+     */
+    function table($alias, array $options = [])
```
dereuromark, line 131: This might better go into 4.next
=> why not stict typehing here for string $alias?

ravage84: > => why not stict typehing here for string $alias?

Simply an oversight because the code started in a 3.x branch. Fixed.

ravage84: > This might better go into 4.next

Changed and rebased.

src/basics.php
```diff
@@ -120,6 +121,20 @@ function breakpoint(): ?string
...
 
+if (!function_exists('table') && (PHP_SAPI === 'cli' || PHP_SAPI === 'phpdbg')) {
+    /**
+     * REPL convenience function to load a table
+     *
+     * @param string $alias The alias name you want to get.
+     * @param array $options The options you want to build the table with.
+     * @return \Cake\ORM\Table
+     */
+    function table(string $alias, array $options = []): Table
```
saeideng, line 132: maybe `getTable()` !!??

src/basics.php
```diff
@@ -120,6 +121,20 @@ function breakpoint(): ?string
...
+    {
+        return Cake\ORM\TableRegistry::getTableLocator()->get($alias, $options);
```
ADmad, line 134: You can add `use` statement and just use class name here.

src/basics.php
ravage84: The idea was to be as brief as possible without going down to an unreadable ``t()``.

dereuromark: I dont think in that context a get prefix makes sense.

I rather have some concerns having this now globally available, anywhere in the system, including templates.
Can be easily misused.

It would be nice if these things could be scoped into the repl only.

ravage84:
> I rather have some concerns having this now globally available, anywhere in the system, including templates.
> Can be easily misused.
> It would be nice if these things could be scoped into the repl only.

@dereuromark I agree.

src/basics.php
ADmad: `table()` is fine. For e.g. to get a collection instance the function is `collection()` not `getCollection()`.

@dereuromark I too am concerned about the misuse, but perhaps SAPI name checks can avoid general use.

dereuromark:
Actually, you can:

     "argv" => [
       "/.../bin/cake.php",
       "console",
     ],

is inside $_SERVER in that case

Maybe this should be a separate file that can be included in that case.
So you can add also other syntactic sugar for REPL only.

ADmad:
> Maybe this should be a separate file that can be included in that case.
> So you can add also other syntactic sugar for REPL only.

:+1:

ravage84:
> Actually, you can:
> 
> ```
>  "argv" => [
>    "/.../bin/cake.php",
>    "console",
>  ],
> ```
> 
> is inside $_SERVER in that case

@dereuromark  sounds a bit fragile/specific but would be an option.

> Maybe this should be a separate file that can be included in that case.
> So you can add also other syntactic sugar for REPL only.

:+1:  I thought about that, too. But didn't want to overdo it right from the start.

If you want to go to extremes, the REPL could even be a plugin and this syntactic sugar could be moved there.

ravage84:
> SAPI name checks

@ADmad do you mean additional checks inside the function? The function is only defined in CLI, anyway.

https://github.com/cakephp/cakephp/pull/14490/files#diff-929f6c44aa6aa503f170c56f0245c9c1R124

dereuromark:
If you make this a plugin, you can do it even right now for 4.x :)

But yeah, for now the CLI check could be enough to include a repl_functions.php file for example.

ADmad:
> The function is only defined in CLI, anyway.

@ravage84 Yes, that's what I was pointing out to.

src/basics.php
ravage84: Fixed

src/basics.php
```diff
@@ -120,6 +122,20 @@ function breakpoint(): ?string
     }
 }
 
+if (!function_exists('table') && (PHP_SAPI === 'cli' || PHP_SAPI === 'phpdbg')) {
+    /**
+     * REPL convenience function to load a table
```
ADmad, line 127: Would be nice to add explicit disclaimer too to not use this function in app/plugin code.

ravage84: How about:

``Do **NOT** use this function in your app or plugin code?``

dereuromark: I would still remove this here into its own separate file, which can have this as file doc block or warning on the top.

ADmad: Moving it in separate file would allow us to load the file using `require-dev` and address our concerns of it's misuse.

ADmad: `breakpoint()` could also be moved to the new file then.

ravage84: Yep, that was my thought, too.

I also had in mind to contribute the pretty print function @lorenzo used in one of his workshops once.
But I couldn't find his code, unfortunately.
Such a function could be placed there, too.

ravage84: @ADmad added the disclaimer.

@dereuromark  & @ADmad moved both functions into ``repl_functions.php``.

> Moving it in separate file would allow us to load the file using require-dev and address our concerns of it's misuse.

This can be done in cakephp/app, once this PR is ready/merged.

dereuromark:
Looks good, but how are you loading it now?
We are missing the autoload-dev part of this I guess.

ravage84:
> Looks good, but how are you loading it now?

We could load in the REPL console command.

dereuromark:
In that case the CLI checks might not even be necessary, as the command itself is only CLI only.

ravage84:
https://github.com/cakephp/app/pull/786 in combination with https://github.com/cakephp/cakephp/pull/14491 is an alternative to this PR.

ADmad:
Closing as the final consensus as discussed on cakephp/app#786 is to have a separate plugin for REPL.

Issue #14471, opened 2020-04-19
Cake 4.0.5 to 4.0.6 upgrade: Missing CSRF token body Cake\Http\Exception\InvalidCsrfTokenException


cnizzardini:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 4.0.6
* Platform and Target: Ubuntu 18 LTS, Apache, PHP 7.2 FPM

### What you did
1. Upgrade to 4.0.5 from 4.0.6
2. Went to do login in on my website
3. Encountered `Missing CSRF token body Cake\Http\Exception\InvalidCsrfTokenException`

### What happened
This happens on any form submission. A workaround is clearing cookie, but iis it reasonable to force clearing this data on each release to avoid any potential CSRF errors? I have locked my CakePHP version to 4.0.5 for now.

**Stack Trace:** 
```
Missing CSRF token body
Cake\Http\Exception\InvalidCsrfTokenException
Toggle Vendor Stack Frames
CORE/src/Http/Middleware/CsrfProtectionMiddleware.php:254
Cake\Http\Middleware\CsrfProtectionMiddleware->_validateToken
CORE/src/Http/Middleware/CsrfProtectionMiddleware.php:133
Cake\Http\Middleware\CsrfProtectionMiddleware->process
CORE/src/Http/Runner.php:73
Cake\Http\Runner->handle
CORE/src/Http/Runner.php:58
Cake\Http\Runner->run
CORE/src/Routing/Middleware/RoutingMiddleware.php:162
Cake\Routing\Middleware\RoutingMiddleware->process
CORE/src/Http/Runner.php:73
Cake\Http\Runner->handle
CORE/src/Routing/Middleware/AssetMiddleware.php:68
Cake\Routing\Middleware\AssetMiddleware->process
CORE/src/Http/Runner.php:73
Cake\Http\Runner->handle
CORE/src/Http/Middleware/DoublePassDecoratorMiddleware.php:74
Cake\Http\Middleware\DoublePassDecoratorMiddleware->Cake\Http\Middleware\{closure}
ROOT/vendor/dereuromark/cakephp-setup/src/Middleware/MaintenanceMiddleware.php:48
Setup\Middleware\MaintenanceMiddleware->__invoke
CORE/src/Http/Middleware/DoublePassDecoratorMiddleware.php:75
Cake\Http\Middleware\DoublePassDecoratorMiddleware->process
[...]

PR #14492, opened 2020-04-22
#14471 support pre 4.0.6 CSRF tokens

cnizzardini:
See #14471 for details. This commit enables support for reading pre 4.0.6 CSRF tokens.

### DIFF
--- a/src/Http/Middleware/CsrfProtectionMiddleware.php
+++ b/src/Http/Middleware/CsrfProtectionMiddleware.php
@@ -241,13 +241,17 @@ class CsrfProtectionMiddleware implements MiddlewareInterface
         $body = $request->getParsedBody();
         if (is_array($body) || $body instanceof ArrayAccess) {
             $post = Hash::get($body, $this->_config['field']);
-            if ($this->_compareToken($post, $cookie)) {
+            if (Security::constantEquals($post, $cookie)) {
+                return;
+            } else if ($this->_compareToken($post, $cookie)) {
                 return;
             }
         }
 
         $header = $request->getHeaderLine('X-CSRF-Token');
-        if ($this->_compareToken($header, $cookie)) {
+        if (Security::constantEquals($header, $cookie)) {
+            return;
+        } else if ($this->_compareToken($header, $cookie)) {
             return;
         }
 
--- a/src/Http/Middleware/CsrfProtectionMiddleware.php
+++ b/src/Http/Middleware/CsrfProtectionMiddleware.php
@@ -241,9 +241,13 @@ class CsrfProtectionMiddleware implements MiddlewareInterface
         $body = $request->getParsedBody();
         if (is_array($body) || $body instanceof ArrayAccess) {
             $post = Hash::get($body, $this->_config['field']);
+            // Allow non-hardened tokens until 4.1.0
             if (Security::constantEquals($post, $cookie)) {
                 return;
-            } else if ($this->_compareToken($post, $cookie)) {
+            }
+            // Validates hardened tokens generated since 4.0.6
+            // The above check will circumvent this until 4.1.0
+            if ($this->_compareToken($post, $cookie)) {
                 return;
             }
         }
--- a/src/Http/Middleware/CsrfProtectionMiddleware.php
+++ b/src/Http/Middleware/CsrfProtectionMiddleware.php
@@ -241,22 +241,25 @@ class CsrfProtectionMiddleware implements MiddlewareInterface
         $body = $request->getParsedBody();
         if (is_array($body) || $body instanceof ArrayAccess) {
             $post = Hash::get($body, $this->_config['field']);
-            // Allow non-hardened tokens until 4.1.0
-            if (Security::constantEquals($post, $cookie)) {
-                return;
-            }
-            // Validates hardened tokens generated since 4.0.6
-            // The above check will circumvent this until 4.1.0
-            if ($this->_compareToken($post, $cookie)) {
+            if (
+                // Allow non-hardened tokens until 4.1.0
+                Security::constantEquals($post, $cookie) ||
+                $this->_compareToken($post, $cookie)
+            ) {
                 return;
             }
         }
 
-        $header = $request->getHeaderLine('X-CSRF-Token');
-        if (Security::constantEquals($header, $cookie)) {
-            return;
-        } else if ($this->_compareToken($header, $cookie)) {
-            return;
+        if ($request->hasHeader('X-CSRF-Token')) {
+            $header = $request->getHeaderLine('X-CSRF-Token');
+            if (
+                // Allow arbitrary tokens until 4.1.0
+                Security::constantEquals($header, $cookie) || 
+                // Validate hmac tokens added in 4.0.6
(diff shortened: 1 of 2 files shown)


src/Http/Middleware/CsrfProtectionMiddleware.php
```diff
@@ -241,13 +241,17 @@ protected function _validateToken(ServerRequestInterface $request): void
         $body = $request->getParsedBody();
         if (is_array($body) || $body instanceof ArrayAccess) {
             $post = Hash::get($body, $this->_config['field']);
-            if ($this->_compareToken($post, $cookie)) {
+            if (Security::constantEquals($post, $cookie)) {
+                return;
+            } else if ($this->_compareToken($post, $cookie)) {
```
dereuromark, line 246: no else required after a return;

cnizzardini: They could also just be combined into a single if statement.

othercorey:
The problem with this change is it disables the new token validation.

dereuromark:
@othercorey You had this self-healing idea? Maybe we try that one?

othercorey:
> @othercorey You had this self-healing idea? Maybe we try that one?

I think @garas had an idea, not me.

As long as we don't simply accept old tokens and generate new ones, then that's ok. There needs to be some place we reject old tokens without authenticating.

markstory:
@othercorey Accepting old tokens is necessary in the short term though. In addition to accepting the old tokens, we could add a step to 'upgrade' the cookie for subsequent requests. Upgrading the token could cause previously opened tabs/pages to fail CSRF but the process would be a bit smoother than before.

othercorey:
> @othercorey Accepting old tokens is necessary in the short term though. In addition to accepting the old tokens

Agreed. I might be misunderstanding the logic here, but this will always validate the token. We don't have a cut-off that requires the new token validation.

If we regenerated the token, this will still trigger old token validation.

markstory:
> If we regenerated the token, this will still trigger old token validation.

Yes, but doing the old check after the new process fails would have the same effect. This change could be short lived and removed in 4.1.0. That would give some time for people to get new tokens in place and upgrade with fewer bumps.

othercorey:
> This change could be short lived and removed in 4.1.0. That would give some time for people to get new tokens in place and upgrade with fewer bumps.

Just as long as we're acknowledging the token hardening is not enabled until 4.1.0. We'll have to adjust the failing unit test to compare different strings instead of non-hardened tokens.

cnizzardini:
My understanding of this change is:

**Old** 4.0.5 and lower tokens are read-only.
**New** 4.0.6 and up tokens are read and write.

So if I have **old** tokens in my application, they will continue to work, but as they expire only the **new** tokens will be created.

The goal being, allow "nature" to take its course as the old tokens slowly get expired in the wild. By the time 4.1 comes around, there should be far fewer instances of old tokens in existence. Those that are will have to force clear cookies in their application. A "breaking" change is more reasonable for a minor version bump than a revision bump.

If this code does not accomplish that, then the PR should be scrapped.

src/Http/Middleware/CsrfProtectionMiddleware.php
```diff
@@ -241,13 +241,17 @@ protected function _validateToken(ServerRequestInterface $request): void
...
                 return;
             }
```
othercorey, line 248:
```suggestion
            // Allow non-hardened tokens until 4.1.0
            if (Security::constantEquals($post, $cookie)) {
                return;
            }
            // Validates hardened tokens generated since 4.0.6
            // The above check will circumvent this until 4.1.0
            if ($this->_compareToken($post, $cookie)) {
                return;
            }
```

othercorey:
> The goal being, allow "nature" to take its course as the old tokens slowly get expired in the wild.If this code does not accomplish that, then the PR should be scrapped or corrected to accomplish that goal.

It does. We have to make a change to 4.1.0 to explicitly remove this and re-document the change since the 4.0.6 release claims the token validation was hardened.

garas:
The main problem with invalid CSRF cookie is that it is left until browser session end and reloading page won't help. Cookie might become invalid not only when upgrading CakePHP to 4.0.6+, but also if `Security.salt` is changed.

To be able use app again, either user must to close browser, or app has to have some logic to detect and clear old cookie.

So CSRF middleware itself has either remove invalid cookie, or accept any cookie, but authenticate it by other ways:

- Regenerate cookie if it is invalid, but it needs to check HMAC on each request.

- Delete CSRF cookie when CSRF validation fails, but I don't know how to modify response (send expired cookie) + throw exception at once in middleware.

- Change CSRF format once again: cookie is any data (no HMAC - compatible with both pre-4.0.6 and 4.0.6 cookie), but change request attribute format to `nonce:cookie:hmac` or even just `nonce:hmac` without repeating cookie data in post/header.

1 & 3 would need generate/check HMAC on each request, but  HMAC is fast to use on each request. 2 would be the most directly targeted solution, but I don't know how to implement it in middleware.

markstory:
> Cookie might become invalid not only when upgrading CakePHP to 4.0.6+, but also if Security.salt is changed.

If an application has Security.salt changed much more than just CSRF tokens is going to have troubles though. 

> Change CSRF format once again: 

I'm reluctant to do that given how problematic this change has proven to be.

>  but change request attribute format to nonce:cookie:hmac or even just nonce:hmac without repeating cookie data in post/header.

Not including the cookie data in the post/header data would circumvent CSRF protection. We need both to compare some data between the cookie and the request data to ensure that the request is both from the current user and from the originating domain. Using only a `nonce:hmac` would enable token reuse between an attacker and the victim.

The change suggested by @othercorey will restore short term compatibility and give us a path to having better tokens in the future.

garas:
> If an application has Security.salt changed much more than just CSRF tokens is going to have troubles though.

The main problem here cookie is persistently stored for browser session and is not replaced with valid token with minimal user action - page reload.

In other cases like FormProtector/DigestAuth, fresh token is generated on each page load, encrypted data in database can be removed massively, encrypted cookie becomes invalid and handled same as not existing. But in CSRF case, cookie is persistent and handled as valid, and fresh valid token never generated.

Single page reload to get new valid token is easy and minor problem which most users will try first, but restarting whole browser is big annoyance and most users won't try this and will just leave the app. Even bigger problem is if CSRF is set to some non-session expire option.

> > Change CSRF format once again:
> I'm reluctant to do that given how problematic this change has proven to be.

As I described, cookie is any data is always valid so would be compatible with any previous format.

> > but change request attribute format to `nonce:cookie:hmac` or even just `nonce:hmac` without repeating cookie data in post/header.
> 
> Not including the cookie data in the post/header data would circumvent CSRF protection. 

Post/header would contain HMAC of cookie data, which is the proof that it comes from the current user and from the originating app. `nonce` is included so attacker couldn't just get HMAC for any data by providing cookie with chosen data, so additional random junk is necessary.

---

I'll go with solution 1 and provide alternate PR: Regenerate cookie if it is invalid by checking cookie HMAC on each request.

markstory:
Closing in favor of the changes in #14504

Issue #14494, opened 2020-04-23
4.1: urlFromPath in general?


dereuromark:
This is a (multiple allowed):

* [] bug
* [x] enhancement
* [x] feature-discussion (RFC)

* CakePHP Version: 4.1+

### What you did

Looking more into what we are currently doing in 4.1 and fromPath usability improvements:

I use a lot of URLs everywhere (preferably always array to be agnostic, never the final routed one!):

    return $this->redirect(
        ['plugin' => 'Messaging', 'controller' => 'Conversations', 'action' => 'index']
    );

in config (auth):

    'loginAction' => [
        'controller' => 'Account', 'action' => 'login', 'prefix' => false, 'plugin' => false
    ],

or in tests:

    $this->get(
        ['prefix' => 'Admin', 'plugin' => 'MyPlugin', 'controller' => 'MyController', 'action' => 'myAction']
    );

### What happened
Right now you need to manually write those, or use copy and paste.

### What you expected to happen
Given, that we can typehint/autocomplete the string paths from a generated list of available actions, it would be nice if we could generate the array from the string path.

What if we created similar methods we did before for collection(), __() and more?
Syntactic sugar that can do that?

    function url(string $path): array {
        return ...
    }

And then we could just inject that in any (even custom) method we need it:
```php
// controller and alike
return $this->redirect(
    url('Messaging.Conversations:index')
);

// config or custom usage
'loginAction' => url('Account::login'],

// testing
$this->get(
    url('MyPlugin.Admin/MyController::myAction')
);
```
The non-declared parts automatically become false (prefix/plugin).

It stays fully optional, would make the rest of us though 300% faster and more productive.
It only needs one meta directive to make this possible for all possible use cases, so we don't need tons of more ways to get this implemented throughout the whole system.

Naming of this global function?
- url()
- urlFromPath()
- ?

Thoughts? Alternatives?

ADmad:
I am onboard for having a `url()` method, I too have thought of having the same in the past but never got around to proposing/implementing it.

ADmad:
Though I would prefer being able to pass other URL parts too to the function, not the path string.

dereuromark:
You are absolutely right, the 2nd $params array must not be forgotten here to make this able to give us back a proper complete array form.

dereuromark:
PR open.

PR #14497, opened 2020-04-23
Add urlArray() convenience function.

dereuromark:
An idea of what was discussed in https://github.com/cakephp/cakephp/issues/14494

### Functionality
This works nicely if we expect the inflection of https://github.com/cakephp/cakephp/pull/14496 to be gone.

I tested it with e.g.

```php
    public function index()
    {
        $this->Flash->success('Test redirect');

        $url = urlArray('DatabaseLog.Admin/DatabaseLog::index');

        return $this->redirect($url);
    }
```

Questions:
- Should this be the _path only still? Or do we want this conversion to happen here right away?
- Is the method name clear and speaking enough, while still being not too long and verbose? 

---

### Usability

With the directive of 
```
    expectedArguments(
        \urlArray(),
        0,
        argumentsSet('routePaths')
    );
```
I got the desired autocomplete now for all URL relevant configs and methods this way:

url_array
=> https://github.com/dereuromark/cakephp-ide-helper/pull/195

### DIFF
--- a/src/Routing/functions.php
+++ b/src/Routing/functions.php
@@ -29,8 +29,12 @@ if (!function_exists('urlArray')) {
      */
     function urlArray(string $path, array $params = []): array
     {
-        $array = Router::parseRoutePath($path);
+        $url = Router::parseRoutePath($path);
+        $url += [
+            'plugin' => false,
+            'prefix' => false,
+        ];
 
-        return $array + $params;
+        return $url + $params;
     }
 }
--- a/src/Routing/functions.php
+++ b/src/Routing/functions.php
@@ -1,0 +1,36 @@
+<?php
+declare(strict_types=1);
+
+/**
+ * CakePHP(tm) : Rapid Development Framework (https://cakephp.org)
+ * Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
+ *
+ * Licensed under The MIT License
+ * For full copyright and license information, please see the LICENSE.txt
+ * Redistributions of files must retain the above copyright notice.
+ *
+ * @copyright     Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
+ * @link          https://cakephp.org CakePHP(tm) Project
+ * @since         4.1.0
+ * @license       https://opensource.org/licenses/mit-license.php MIT License
+ */
+
+use Cake\Routing\Router;
+
+if (!function_exists('urlArray')) {
+    /**
+     * Returns an array URL from a route path string.
+     *
+     * @param string $path Route path.
+     * @param array $params An array specifying any additional parameters.
+     *   Can be also any special parameters supported by `Router::url()`.
+     * @return array URL
+     * @see \Cake\Routing\Router::pathUrl()
+     */
+    function urlArray(string $path, array $params = []): array
+    {
+        $array = Router::parseRoutePath($path);
+
+        return $array + $params;
+    }
+}
--- a/tests/TestCase/Routing/RouterTest.php
+++ b/tests/TestCase/Routing/RouterTest.php
@@ -2270,6 +2270,39 @@ class RouterTest extends TestCase
     }
 
     /**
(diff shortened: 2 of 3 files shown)


src/Routing/functions.php
```diff
@@ -0,0 +1,36 @@
...
+    /**
+     * Returns an array URL from a route path string.
+     *
+     * @param string $path Route path.
+     * @param array $params An array specifying any additional parameters.
+     *   Can be also any special parameters supported by `Router::url()`.
+     * @return array URL
+     * @see \Cake\Routing\Router::pathUrl()
+     */
+    function urlArray(string $path, array $params = []): array
```
markstory, line 30: Why not a name like `url()`, `route()` or `urlRoute()`? This method doesn't return an array, and can be used without any array parameters.

dereuromark: My idea was to return an array here to be used further inside the application.
The array is the ideal data transport to be easily merged and modified, whereas the string seems to be the ideal input.
And they shouldn't be localized or finalized until the very end of the request, in order for the introspections and modifications to still be possible.

dereuromark: This is purely opt-in right now, for those that want to use it instead of normal/classic strings.
I don't think we need to move to use those on core level or if you don't have the IDE support to actually be faster this way.

ADmad: Why not have another function too which returns string? Users can use whatever they prefer/need.

dereuromark: So just url()? How would that function work?

garas: If `$url = urlArray('Users::index');` returns string, and app in subdir, it would return `'/subdir/users'`.

But if it then passed to `Router::url()` or `$this->redirect($url)`, subdir is applied again: `'/subdir/subdir/users'`.

So I would go with array.

dereuromark: Yes, the final string is the worst format to use already, and continues to be so.
Unflexible and error prone when using it down the line, also not easily comparable where needed (as it is already final route string).

ADmad: @garas Yes, that can be issue, though if we allow additional params one can control whether then want the URL string with or without base.

ADmad: But rather than dealing with users reports of base getting prefixed twice, I can add my own `url()` function which wraps `Router::pathUrl()` :slightly_smiling_face:

dereuromark: Feel free to add commits on top :)

ADmad: I am good with current state of PR. I'll put some more thought into the possibility of adding a function which returns string later.

dereuromark:
Cool, all right. Anything else missing?
I started docs in https://github.com/cakephp/docs/pull/6605

src/Routing/functions.php
garas: `Router::pathUrl()` and `urlArray()` now has different behavior.

`Router::pathUrl()` will reset prefix/plugin to `false`, but `urlArray()` will inherit prefix/plugin from current route.

`urlArray()` should better be just shortcut of `Router::pathUrl()`.

dereuromark: I thought it behaved the same. Indeed, it should do so.

dereuromark: I updated the code and tests.

tests/TestCase/Routing/RouterTest.php
```diff
@@ -2269,6 +2269,42 @@ public function testParseInvalidRoutePath(string $value)
...
+        $expected = [
+            'controller' => 'Bookmarks',
+            'action' => 'view',
+            'plugin' => false,
+            'prefix' => false,
+        ];
+        $this->assertSame($expected, urlArray('Bookmarks::view'));
+
+        $expected = [
+            'prefix' => 'admin',
```
markstory, line 2289: Shouldn't this be Admin? Prefixes should be CamelBacked now.

dereuromark:
I rebased on top of 4.next.

Issue #12014, opened 2018-04-27
Creating a custom Locator


davidyell:
This is a (multiple allowed):

* [ ] bug
* [x] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.6.1
* Platform and Target: Nginx, PHP 7.2.4

### What you did
```php
class EndpointLocator implements LocatorInterface
{
    public function set($alias, Endpoint $object)
    {

    }

    // Rest of class
}
```

### What happened
An error is generated because the interface stipulates that the `set()` method must take a Table class instance.

https://github.com/cakephp/cakephp/blob/master/src/ORM/Locator/LocatorInterface.php#L62

### What you expected to happen
I was expecting the interface to be more generic to allow for more varying implementations. So I could locate other types of repository class.

dereuromark:
Refs also https://github.com/cakephp/cakephp/issues/11861 etc.

davidyell:
Also seems that the `config()` method is not deprecated in the interface, but is in the `TableLocator` implementation.

davidyell:
It also seems that the interface specifies the `get()` method as being not static. Yet the FactoryLocator class is trying to call it statically.

```php
\Cake\Datasource\FactoryLocator::add('Endpoint', [Muffin\Webservice\Model\EndpointLocator::class, 'get']);

// Deprecated: Non-static method Muffin\Webservice\Model\EndpointLocator::get() should not be called statically in /Users/david/Sites/MuffinWebservice/vendor/cakephp/cakephp/src/Datasource/FactoryLocator.php on line 36
```

ADmad:
Changing the typehint to `RepositoryInterface` would probably cause PHP to complain but I think it's worth doing in 3.6 itself.

ADmad:
Actually given that `LocatorInterface` is under `Cake\ORM` namespace using `Table` typehint makes sense.

If needed we can add a new `Cake\Datasource\LocatorInterface` which non-table data sources can implement.

robertpustulka:
Yes, this interface is intended to work with ORM Tables only.

> It also seems that the interface specifies the get() method as being not static. Yet the FactoryLocator class is trying to call it statically.

@davidyell So why are you trying to call non-static method statically? :)
Locators are intended to be objects. so:

```php
$locator = new \Muffin\Webservice\Model\EndpointLocator;

\Cake\Datasource\FactoryLocator::add(
    'Endpoint', 
    [$locator, 'get']
);
```

[...]

PR #14499, opened 2020-04-24
Update TableLocator Interface to add more flexibility

### LINKED ISSUE #14495: 3.x | LocatorInterface only allows getting Tables?
This is a (multiple allowed):

* [ ] bug
* [x] enhancement

* CakePHP Version: 3.8.11
* Platform and Target: PHP 7.4.1

## What you found
Why does the `\Cake\ORM\Locator\LocatorInterface` stipulate that the `get()` method must return `\Cake\ORM\Table` and not `\Cake\Datasource\RepositoryInterface`?

If I was to implement this interface into my own locator to find my custom repositories it would 
 fail the interface? Unless my custom repository classes also extended `\Cake\ORM\Table`?

The class `\Cake\ORM\Locator\TableLocator` does return `\Cake\ORM\Table` as its name implies. So it seems to me the interface has been missed.

## Suggested change
The interface be updated to return an interface, allowing more flexibility.

I can open a pull request to update this if needed.

## References
Follows on from #12014
https://github.com/cakephp/cakephp/blob/3.8.11/src/ORM/Locator/LocatorInterface.php#L37-L44

davidyell:
Resolves #14495

This change creates more flexibility in the interface by allowing custom Locator implementations such as the use-case shown in #12014

It should actually make no difference to the core framework code as `\Cake\ORM\Table` already implements `\Cake\Datasource\RepositoryInterface`.

The only downside I can think of, might be [the stand-alone ORM package](https://github.com/cakephp/orm), as I can't see the interface in there.

### DIFF
--- a/src/ORM/Locator/LocatorInterface.php
+++ b/src/ORM/Locator/LocatorInterface.php
@@ -39,7 +39,7 @@ interface LocatorInterface
      *
      * @param string $alias The alias name you want to get.
      * @param array $options The options you want to build the table with.
-     * @return \Cake\ORM\Table
+     * @return \Cake\Datasource\RepositoryInterface
      */
     public function get($alias, array $options = []);
 


ndm2:
Is this something that is ment to go into 4.x too? Because when mapped to real type declarations this would be invalid, as return type covariance is only available as of PHP 7.4

markstory:
@ndm That's unfortunate. Changing the typing on the current interface would break compatibility :cry:

othercorey:
We could add it to a wiki page tracking changes we can make when PHP 7.4 is the minimum version.

othercorey:
Added to 5.0 roadmap.

PR #14500, opened 2020-04-24
Generate new CSRF cookie token if invalid

garas:
Alternate to https://github.com/cakephp/cakephp/pull/14492.

Validate CSRF cookie on each request and generate new CSRF cookie token if necessary.


tests/TestCase/Http/Middleware/CsrfProtectionMiddlewareTest.php
```diff
@@ -100,6 +100,48 @@ public function testSettingCookie()
...
+    /**
+     * Test invalid cookie is replaced with valid cookie
+     *
+     * @return void
+     */
+    public function testInvalidCookieTokenRegenerated()
+    {
+        $request = new ServerRequest([
+            'environment' => ['REQUEST_METHOD' => 'GET'],
+            'cookies' => ['csrfToken' => 'invalidToken'],
```
markstory, line 112: Should we have a test that covers the arbitrary cookie data working when it is submitted with an hmac'ed value?

garas: There are few tests which uses hmac'ed value (testValidTokenInHeader, testValidTokenRequestData) so I think no extra test necessarily.

ADmad:
> Validate CSRF cookie on each request and generate new CSRF cookie token if necessary.

Wouldn't this add a non trivial amount of overhead to every request?

garas:
> 
> 
> > Validate CSRF cookie on each request and generate new CSRF cookie token if necessary.
> 
> Wouldn't this add a non trivial amount of overhead to every request?

HMAC is fast (2 × sha1 + bit more data mangling).

---

However I just learned how to send headers with exceptions with `\Cake\Core\Exception\Exception::responseHeader()`.

I can now implement and delete CSRF cookie when CSRF validation fails, as this is the most directly targeted solution.

garas:
Replaced by more direct solution without validating CSRF cookie on each `GET` request

https://github.com/cakephp/cakephp/pull/14504

PR #14502, opened 2020-04-24
Added View::$prefixPath to support self-contained features

othercorey:
This adds the configuration necessary for Bake to run without hacks. Features can be self-contained throughout ecosystem in a sub-directory under templates.

This is similar to prefix route controller rendering with 2 differences:

- The prefix path is applied to all files
- The prefix path is not inflected or generated

### DIFF
--- a/src/View/View.php
+++ b/src/View/View.php
@@ -116,6 +116,13 @@ class View implements EventDispatcherInterface
     protected $helpers = [];
 
     /**
+     * The name of the subfolders cotaining all files for this View.
+     *
+     * @var string
+     */
+    protected $prefixPath;
+
+    /**
      * The name of the subfolder containing templates for this View.
      *
      * @var string
@@ -216,7 +223,7 @@ class View implements EventDispatcherInterface
      */
     protected $_passedVars = [
         'viewVars', 'autoLayout', 'helpers', 'template', 'layout', 'name', 'theme',
-        'layoutPath', 'templatePath', 'plugin',
+        'layoutPath', 'templatePath', 'prefixPath', 'plugin',
     ];
 
     /**
@@ -412,6 +419,29 @@ class View implements EventDispatcherInterface
     public function setResponse(Response $response)
     {
         $this->response = $response;
+
+        return $this;
+    }
+
+    /**
+     * Gets prefix path for files.
+     *
+     * @return string
+     */
+    public function getPrefixPath(): string
+    {
+        return $this->prefixPath;
+    }
+
+    /**
+     * Sets prefix path for files.
+     *
+     * @param string $path Prefix path for files
+     * @return $this
+     */
+    public function setPrefixPath(string $path)
+    {
+        $this->prefixPath = $path;
 
         return $this;
     }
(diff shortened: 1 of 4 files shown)


src/View/View.php
```diff
@@ -115,6 +115,13 @@ class View implements EventDispatcherInterface
      */
     protected $helpers = [];
 
+    /**
+     * The nname of the subfolders cotaining all files for this View.
+     *
+     * @var string
```
dereuromark, line 121: |null
Typo in "nname".

dereuromark:
I could have needed that probably yesterday ;)
https://github.com/dereuromark/cakephp-captcha/pull/16

I just copied the template to all possible subdirs, and it worked then for me.
So I am glad if I can fix this up some time in the future.

src/View/View.php
othercorey: I didn't add |null because all the other config variables don't use it. I think that's something to change separately.

ADmad:
> This adds the configuration necessary for Bake to run without hacks.

Exactly what hacks are you referring to?

Currently templates paths are already prefixed based on value of `View::$plugin` and `View::$subDir`. Adding yet another way to prefix paths make thing murkier and View code more difficult to follow.

othercorey:
Bake was doing two things:

- Modifying the App.paths.templates config
- Modifying the generated paths

The first can be changed to simply use proper template naming with `Bake.` plugin convention.

This moves the path modification to a config.

The View configuration is very confusing and not well documented. Some of the configs don't seem to provide anything specific. This provides a simple "here is the templates" path that isn't split between template, layout and controller.

Maybe we don't want it, but I wish that system was more clear.

othercorey:
I will say the name `$prefixPath` might be too subtle if we use this. Maybe it should be more explicit so we don't run into the issue that exists now with "path" and "subdir" configs.

Maybe there is a forward looking way to configure View paths instead of setting these individual and only slightly different sub paths.

src/View/View.php
```diff
@@ -1531,7 +1561,12 @@ protected function _paths(?string $plugin = null, bool $cached = true): array
                 return $this->_pathsForPlugin[$plugin];
             }
         }
-        $templatePaths = App::path(static::NAME_TEMPLATE);
+
+        $prefixPath = $this->prefixPath ? $this->prefixPath . DIRECTORY_SEPARATOR : '';
+        $templatePaths = array_map(function (string $path) use ($prefixPath) {
+            return $path . $prefixPath;
+        }, App::path(static::NAME_TEMPLATE));
```
markstory, line 1568: If prefixPath is empty this loop just burns time. Could we skip it?

src/View/View.php
```diff
@@ -1531,7 +1561,12 @@ protected function _paths(?string $plugin = null, bool $cached = true): array
                 return $this->_pathsForPlugin[$plugin];
             }
         }
-        $templatePaths = App::path(static::NAME_TEMPLATE);
+
+        $prefixPath = $this->prefixPath ? $this->prefixPath . DIRECTORY_SEPARATOR : '';
```
markstory, line 1565: We have a `subDir` already which is inserted in a similar position in the template path. Would that potentially work for bake as well?

othercorey: So, `subdir` only affects template paths and is inserted after the `templatePath`. It does not affect element and layout paths.

ADmad:
I would rather leave couple of special path modifications in Bake than add more complexity to the View class. We can look into cleaning up the path manipulation in View in next major.

othercorey:
> I would rather leave couple of special path modifications in BakeView than add more complexity to the View class. We can look into cleaning up the path manipulation in View in next major.

I created this PR for the discussion. I'd also like to think of a better way to configure View.

ADmad:
As mentioned in this comment https://github.com/cakephp/bake/issues/683#issuecomment-619301449 probably need additional changes in bake anyway to allow use to include multiple themes. Currently we use plugins as themes so a plugin can only have one theme.

PR #14507, opened 2020-04-27
2.x: Fix argument order for implode method

JakubOnderka:
Different order is deprecated and will cause fatal error in PHP8

> implode() can, for historical reasons, accept its parameters in either order. For consistency with explode(), however, it is deprecated not to use the documented order of arguments.

– https://www.php.net/manual/en/function.implode.php

### DIFF
--- a/lib/Cake/Console/ConsoleOutput.php
+++ b/lib/Cake/Console/ConsoleOutput.php
@@ -265,7 +265,7 @@ class ConsoleOutput {
 				$styleInfo[] = static::$_options[$option];
 			}
 		}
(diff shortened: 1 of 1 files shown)


ADmad:
There will probably be more fixes which will be needed to be done, so it's better to wait till the PHP ecosystem updates to be able to run our test suite for PHP 8.

ADmad:
I see that this issue is preventing fixing an issue in your project, so I'll merge it :)

dereuromark:
This should also already produce deprecations in PHP7 afaik.
Either way, good to get rid of it asap.

PR #14515, opened 2020-04-29
Improve ErrorLogger for PHP errors

markstory:
I was integrating an error tracking service with a CakePHP application and had hoped to be able to replace the `errorLogger` with a different implementation that forwarded errors to the remote service and then also logged to syslog.

This was challenging because exceptions are logged by the ErrorLogger but PHP errors are logged via the ErrorHandler. This change unifies logging to always be done in the ErrorLogger and adds an interface so we don't have to be lucky when implementing custom error loggers.

### DIFF
--- a/src/Error/BaseErrorHandler.php
+++ b/src/Error/BaseErrorHandler.php
@@ -18,9 +18,9 @@ namespace Cake\Error;
 
 use Cake\Core\Configure;
 use Cake\Core\InstanceConfigTrait;
-use Cake\Log\Log;
 use Cake\Routing\Router;
 use Psr\Http\Message\ServerRequestInterface;
+use RuntimeException;
 use Throwable;
 
 /**
@@ -322,9 +322,8 @@ abstract class BaseErrorHandler
             }
             $message .= "\nTrace:\n" . $trace . "\n";
         }
-        $message .= "\n\n";
 
-        return Log::write($level, $message);
+        return $this->getLogger()->logMessage($level, $message);
     }
 
     /**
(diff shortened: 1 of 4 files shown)


src/Error/ErrorLoggerInterface.php
```diff
@@ -0,0 +1,50 @@
...
+use Psr\Http\Message\ServerRequestInterface;
+use \Throwable;
+
+/**
+ * Interface for error logging handlers.
+ *
+ * Used by the ErrorHandlerMiddleware and global
+ * error handlers to log exceptions and errors.
+ */
+interface ErrorLoggerInterface
```
ADmad, line 28: It would have been nice if we could just use or extend `LoggerInterface` (as we did for `Connection`). But since existing `ErrorLogger::log()` has a different signature it would be a BC break.

src/Error/BaseErrorHandler.php
```diff
@@ -351,11 +350,21 @@ public function logException(Throwable $exception, ?ServerRequestInterface $requ
...
+            /** @var \Cake\Error\ErrorLoggerInterface $logger */
             $logger = new $this->_config['errorLogger']($this->_config);
+
+            if (!$logger instanceof ErrorLoggerInterface) {
+                // Set the logger so that the next error can be logged.
+                $this->logger = new ErrorLogger($this->_config);
+
+                $interface = ErrorLoggerInterface::class;
+                $type = getTypeName($logger);
+                throw new RuntimeException("Cannot create logger. {$type} does not implement {$interface}");
```
othercorey, line 362:
```suggestion
                throw new RuntimeException("Cannot create logger. `{$type}` does not implement `{$interface}`.");
```

src/Error/ErrorLoggerInterface.php
dereuromark: A cleanup for 5.x maybe then :)

markstory: Yeah, we can't do that today unfortunately.

dereuromark (approved):

othercorey:
How are we going to fix the getRequestContext() error?

dereuromark:
Adding a 2nd logger interface maybe that you need to implement if you are interested in the request 
 context:

    RequestContextAwareLoggerInterface

or sth like that
and then if (instance of interface && $request)

ADmad:
These changes would make static analysers happy and make `BaseErrorHandler` cleaner IMO:

https://gist.github.com/ADmad/d906fd769dcc45fc6e4b92da20a1908f

othercorey:
I too would prefer to avoid another interface and just decorate the result.

markstory:
I like passing the request in as a context variable and letting loggers do what they need with it. Getting request context in a custom logger could be done with inheritance or copy/paste.

PR #14516, opened 2020-04-29
Fix ISO8601 validation from Validator

johanmeiring:
When using the `dateTime()` proxy method in `Validator` with a date format of `[iso8601]`, the validation does not pass, even with a valid value.  A small addition to the test case has been provided to prove the problem, and a non-invasive solution has been implemented.


markstory:
Thank you 👍

PR #14520, opened 2020-04-29
Remove unused property.

ADmad:
DateFormatTrait::$_isDateInstance was made defunct in #13873.


othercorey:
Looks like I was unsure what to touch back then :)

PR #14522, opened 2020-04-30
Allow disabling lenient IntlDateFormatter parsing

### LINKED ISSUE #14519: Use setLenient(false) when parsing localized date(time)
This is a (multiple allowed):

* [x] bug
* [x] enhancement

* CakePHP Version: 4.0.
* Platform and Target: All.

### What you did

At [I18n\DateFormatTrait](https://github.com/cakephp/cakephp/blob/master/src/I18n/DateFormatTrait.php) we have methods to parse strings of date and datetime into objects.

### What happened

When passing invalid dates for some locales, the parser silent fail and return a non-predictive date object.

Parsing should fail.

I think it should expect string localized representing valid dates/datetimes.

### Propose

Disable, or allow us to disable, the lenient mode from IntlDateFormatter, [as pointed in PHP documentation](https://www.php.net/manual/en/intldateformatter.setlenient.php).

Code sample:

```php
// If you don't want to setup pt_BR locale, use a string with '25-12-2000' (25 of december)
// It will be parsed as 2002-01-12
Cake\I18n\I18n::setLocale('pt_BR');

/**
 * Day: 04
 * Month: 21   <-- This is invalid
 * Year: 1980
 */
$invalid = '04/21/1980';
$result = Cake\I18n\Date::parseDate($invalid);

// Expected: null|false
// Result: Cake\I18n\Date Object
// (
//     [time] => 1981-09-04 00:00:00.000000+00:00
//     [timezone] => UTC
//     [fixedNowTime] => 2020-04-29T13:50:04.505974+00:00
// )
debug($result);
```

othercorey:
Closes https://github.com/cakephp/cakephp/issues/14519

### DIFF
--- a/src/I18n/DateFormatTrait.php
+++ b/src/I18n/DateFormatTrait.php
@@ -43,6 +43,15 @@ trait DateFormatTrait
     protected static $defaultLocale;
 
     /**
+     * Whether lenient parsing is enabled for IntlDateFormatter.
+     *
+     * Defaults to true which is the default for IntlDateFormatter.
+     *
+     * @var bool
+     */
+    protected static $lentientParsing = true;
+
+    /**
      * In-memory cache of date formatters
      *
      * @var \IntlDateFormatter[]
@@ -69,12 +78,44 @@ trait DateFormatTrait
     /**
      * Sets the default locale.
      *
-     * @param string|null $locale The default locale string to be used or null.
+     * Set to null to use IntlDateFormatter default.
+     *
+     * @param string|null $locale The default locale string to be used.
      * @return void
      */
     public static function setDefaultLocale(?string $locale = null): void
     {
         static::$defaultLocale = $locale;
+    }
+
+    /**
+     * Gets whether locale format parsing is set to lenient.
+     *
+     * @return bool
+     */
+    public static function lentientParsingEnabled(): bool
+    {
+        return static::$lentientParsing;
+    }
+
+    /**
+     * Enables lentient parsing for locale formats.
+     *
+     * @return void
+     */
+    public static function enableLentientParsing(): void
+    {
+        static::$lentientParsing = true;
+    }
+
+    /**
+     * Enables lentient parsing for locale formats.
+     *
+     * @return void
+     */
+    public static function disableLentientParsing(): void
+    {
+        static::$lentientParsing = false;
     }
 
     /**
@@ -335,6 +376,8 @@ trait DateFormatTrait
             null,
             $pattern ?? ''
         );
+        $formatter->setLenient(static::$lentientParsing);
+
         $time = $formatter->parse($time);
         if ($time !== false) {
             $dateTime = new DateTime('@' . $time);
--- a/src/I18n/DateFormatTrait.php
+++ b/src/I18n/DateFormatTrait.php
@@ -49,7 +49,7 @@ trait DateFormatTrait
      *
      * @var bool
      */
-    protected static $lentientParsing = true;
+    protected static $lenientParsing = true;
 
     /**
      * In-memory cache of date formatters
@@ -93,29 +93,29 @@ trait DateFormatTrait
      *
      * @return bool
      */
-    public static function lentientParsingEnabled(): bool
+    public static function lenientParsingEnabled(): bool
     {
-        return static::$lentientParsing;
+        return static::$lenientParsing;
     }
 
     /**
-     * Enables lentient parsing for locale formats.
+     * Enables lenient parsing for locale formats.
      *
      * @return void
      */
-    public static function enableLentientParsing(): void
+    public static function enableLenientParsing(): void
     {
-        static::$lentientParsing = true;
+        static::$lenientParsing = true;
     }
 
     /**
-     * Enables lentient parsing for locale formats.
+     * Enables lenient parsing for locale formats.
      *
      * @return void
      */
-    public static function disableLentientParsing(): void
+    public static function disableLenientParsing(): void
     {
-        static::$lentientParsing = false;
+        static::$lenientParsing = false;
     }
 
     /**
@@ -376,7 +376,7 @@ trait DateFormatTrait
             null,
             $pattern ?? ''
         );
-        $formatter->setLenient(static::$lentientParsing);
+        $formatter->setLenient(static::$lenientParsing);
 
         $time = $formatter->parse($time);
         if ($time !== false) {
--- a/tests/TestCase/I18n/DateTest.php
+++ b/tests/TestCase/I18n/DateTest.php
@@ -217,6 +217,25 @@ class DateTest extends TestCase
     }
 
     /**
+     * Tests disabling leniency when parsing locale format.
+     *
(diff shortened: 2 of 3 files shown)


othercorey:
Maybe this should be `enableLenientParsing()` and `disableLenientParsing()` and not worry about clearing. No one is going to need that scenario and this reads easier.

src/I18n/DateFormatTrait.php
```diff
@@ -42,6 +42,15 @@ trait DateFormatTrait
...
     protected static $defaultLocale;
 
+    /**
+     * Whether the IntlDateFormatter is set to lenient for parsing.
+     *
+     * Defaults to null which uses IntlDateFormatter default (true).
+     *
+     * @var bool|null
+     */
+    protected static $lentientParsing;
```
ADmad, line 52: Why don't you just use true/false instead of null/false?

ADmad: Using two falsey values to indicate different states is not intuitive.

othercorey: I was trying not to hide the intldateformatter default but that's too complicated for this.

src/I18n/DateFormatTrait.php
```diff
@@ -42,6 +42,15 @@ trait DateFormatTrait
...
     protected static $defaultLocale;
 
+    /**
+     * Whether lenient parsing is enabled for IntlDateFormatter.
+     *
+     * Defaults to true which is the default for IntlDateFormatter.
+     *
+     * @var bool
+     */
+    protected static $lentientParsing = true;
```
markstory, line 52:
```suggestion
    protected static $lenientParsing = true;
```

src/I18n/DateFormatTrait.php
```diff
@@ -69,14 +78,46 @@ public static function getDefaultLocale(): ?string
...
     {
         static::$defaultLocale = $locale;
     }
 
+    /**
+     * Gets whether locale format parsing is set to lenient.
+     *
+     * @return bool
+     */
+    public static function lentientParsingEnabled(): bool
```
markstory, line 96:
```suggestion
    public static function lenientParsingEnabled(): bool
```

src/I18n/DateFormatTrait.php
```diff
@@ -69,14 +78,46 @@ public static function getDefaultLocale(): ?string
...
+    {
+        return static::$lentientParsing;
```
markstory, line 98:
```suggestion
        return static::$lenientParsing;
```

src/I18n/DateFormatTrait.php
```diff
@@ -69,14 +78,46 @@ public static function getDefaultLocale(): ?string
...
+    }
+
+    /**
+     * Enables lentient parsing for locale formats.
```
markstory, line 102:
```suggestion
     * Enables lenient parsing for locale formats.
```

src/I18n/DateFormatTrait.php
```diff
@@ -69,14 +78,46 @@ public static function getDefaultLocale(): ?string
...
+     *
+     * @return void
+     */
+    public static function enableLentientParsing(): void
```
markstory, line 106:
```suggestion
    public static function enableLenientParsing(): void
```

src/I18n/DateFormatTrait.php
```diff
@@ -69,14 +78,46 @@ public static function getDefaultLocale(): ?string
...
+    {
+        static::$lentientParsing = true;
```
markstory, line 108:
```suggestion
        static::$lenientParsing = true;
```

src/I18n/DateFormatTrait.php
```diff
@@ -69,14 +78,46 @@ public static function getDefaultLocale(): ?string
...
+    }
+
+    /**
+     * Enables lentient parsing for locale formats.
```
markstory, line 112:
```suggestion
     * Enables lenient parsing for locale formats.
```

src/I18n/DateFormatTrait.php
```diff
@@ -69,14 +78,46 @@ public static function getDefaultLocale(): ?string
...
+     *
+     * @return void
+     */
+    public static function disableLentientParsing(): void
```
markstory, line 116:
```suggestion
    public static function disableLenientParsing(): void
```

src/I18n/DateFormatTrait.php
```diff
@@ -69,14 +78,46 @@ public static function getDefaultLocale(): ?string
...
+    {
+        static::$lentientParsing = false;
```
markstory, line 118:
```suggestion
        static::$lenientParsing = false;
```

src/I18n/DateFormatTrait.php
```diff
@@ -335,6 +376,8 @@ public static function parseDateTime(string $time, $format = null, $tz = null)
             null,
             $pattern ?? ''
         );
+        $formatter->setLenient(static::$lentientParsing);
```
markstory, line 379:
```suggestion
        $formatter->setLenient(static::$lenientParsing);
```

tests/TestCase/I18n/DateTest.php
```diff
@@ -216,6 +216,25 @@ public function testParseDateTime($class)
...
+     * Tests disabling leniency when parsing locale format.
+     *
+     * @dataProvider classNameProvider
+     * @return void
+     */
+    public function testLenientParseDate($class)
+    {
+        $class::setDefaultLocale('pt_BR');
+
+        $class::disableLentientParsing();
```
markstory, line 229:
```suggestion
        $class::disableLenientParsing();
```

tests/TestCase/I18n/DateTest.php
```diff
@@ -216,6 +216,25 @@ public function testParseDateTime($class)
...
+        $date = $class::parseDate('04/21/2013');
+        $this->assertSame(null, $date);
+
+        $class::enableLentientParsing();
```
markstory, line 233:
```suggestion
        $class::enableLenientParsing();
```

tests/TestCase/I18n/TimeTest.php
```diff
@@ -916,6 +916,25 @@ public function testParseTime($class)
...
+     * Tests disabling leniency when parsing locale format.
+     *
+     * @dataProvider classNameProvider
+     * @return void
+     */
+    public function testLenientParseDate($class)
+    {
+        $class::setDefaultLocale('pt_BR');
+
+        $class::disableLentientParsing();
```
markstory, line 929:
```suggestion
        $class::disableLenientParsing();
```

tests/TestCase/I18n/TimeTest.php
```diff
@@ -916,6 +916,25 @@ public function testParseTime($class)
...
+        $time = $class::parseDate('04/21/2013');
+        $this->assertSame(null, $time);
+
+        $class::enableLentientParsing();
```
markstory, line 933:
```suggestion
        $class::enableLenientParsing();
```

PR #14523, opened 2020-04-30
Improve error message for EmailTrait::MailContains

### LINKED ISSUE #14518: Improve error message for EmailTrait::MailContains
This is a (multiple allowed):

* [ ] bug
* [x] enhancement

* CakePHP Version: 4.0.6

If the assert fails, I get a message similar to this
`Failed asserting that 'my text' is in an email.
`

It would be great, If I could see the actual message that was available, like:
`Failed asserting that 'my text' is in an email. Was: 'your text'
`

mrothauer:
This PR fixes #14518 

This changing would have helped my with my use case.

Questions

* The changings need to be implemented for all classes that extend MailContains (for now only implemented for MailContainsText). Maybe there is a smarter way of adding the error message to the result of the __toString()-method than appending it in every class?
* $this->at is not yet considered - I don't know what it's for.
* Maybe each message should be truncated (show 100 characters?) and not the concatenated result of all messages (as implemented)?

### DIFF
--- a/src/TestSuite/Constraint/Email/MailContains.php
+++ b/src/TestSuite/Constraint/Email/MailContains.php
@@ -75,7 +75,7 @@ class MailContains extends MailConstraintBase
         }
         $result = join(PHP_EOL, $messageMembers);
 
-        return PHP_EOL . 'was: ' . substr($result, 0, 1000);
+        return PHP_EOL . 'was: ' . mb_substr($result, 0, 1000);
     }
 
     /**
--- a/src/TestSuite/Constraint/Email/MailContains.php
+++ b/src/TestSuite/Constraint/Email/MailContains.php
@@ -55,7 +55,7 @@ class MailContains extends MailConstraintBase
     /**
      * @return string
      */
-    private function getTypeMethod(): string
+    protected function getTypeMethod(): string
     {
         return 'getBody' . ($this->type ? ucfirst($this->type) : 'String');
     }
--- a/src/TestSuite/Constraint/Email/MailContains.php
+++ b/src/TestSuite/Constraint/Email/MailContains.php
@@ -73,9 +73,9 @@ class MailContains extends MailConstraintBase
             $method = $this->getTypeMethod();
             $messageMembers[] = $message->$method();
         }
-        $result = join(LF, $messageMembers);
+        $result = join(PHP_EOL, $messageMembers);
 
-        return LF . 'was: ' . substr($result, 0, 1000);
+        return PHP_EOL . 'was: ' . substr($result, 0, 1000);
     }
 
     /**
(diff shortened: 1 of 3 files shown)


src/TestSuite/Constraint/Email/MailContains.php
```diff
@@ -52,6 +52,32 @@ public function matches($other): bool
...
+    {
+        $messageMembers = [];
+        $messages = $this->getMessages();
+        foreach ($messages as $message) {
+            $method = $this->getTypeMethod();
+            $messageMembers[] = $message->$method();
+        }
+        $result = join(LF, $messageMembers);
+
+        return LF . 'was: ' . substr($result, 0, 1000);
```
dereuromark, line 78: mb_ function is important here
And usually we use PHP_EOL.

src/TestSuite/Constraint/Email/MailContains.php
```diff
@@ -52,6 +52,32 @@ public function matches($other): bool
         return false;
     }
 
+    /**
+     * @return string
+     */
+    private function getTypeMethod(): string
```
dereuromark, line 58: Why not protected?

src/TestSuite/Constraint/Email/MailContains.php
mrothauer: ok I just changed it

src/TestSuite/Constraint/Email/MailContains.php
mrothauer: because the method is only called from the class itself

dereuromark: We don't do that in a framework context where not absolutely required, as this also prohibits further extension where needed.

dereuromark:
Looks good
Can we have a unit test for this to show that it is working?

src/TestSuite/Constraint/Email/MailContains.php
mrothauer: ok I changed it

mrothauer:
I fixed the one unit test that failed due to my changings. Unfortunately the message in the unit test is empty, so it's not actually a proof that it works as suspected.

I'll be back on Monday and it's totally ok for me if you finish that PR without me "help"? But I'm also happy to continue with your support after the weekend.

markstory:
Thank you 👍

mrothauer:
From my point of view the PR was not yet finished. The other existing toString() methods of classes that inherit from MailContains need to be adapted (including the `if ($this->at) {`-cases) and a working unit test are missing.

markstory:
@mrothauer Ok. Could you send another pull request with the remaining changes you think are needed?

PR #14524, opened 2020-04-30
Adjust interface contracts as per implementation.

dereuromark:


src/Datasource/QueryInterface.php
```diff
@@ -34,7 +34,7 @@ interface QueryInterface
      * If `true` is passed in the second argument, any previous selections will
      * be overwritten with the list passed in the first argument.
      *
-     * @param mixed $fields Fields to be added to the list.
+     * @param array|\Cake\Database\ExpressionInterface|string|callable|\Cake\ORM\Table|\Cake\ORM\Association $fields Fields.
```
othercorey, line 37:
```suggestion
     * @param array|string|callable|\Cake\Database\ExpressionInterface|\Cake\ORM\Table|\Cake\ORM\Association $fields Fields.
```

PR #14525, opened 2020-05-01
4.next - Adds basic support for common table expressions.

### LINKED ISSUE #10666: Add support for Common Table Expressions (CTE)
This is a (multiple allowed):

* [x] enhancement

* CakePHP Version: 3.4
* Platform and Target: MariaDB 10.2

### What you did

```sql
# This is a Common Table Expression (this issue):
WITH cms_elements_translations AS (
  SELECT
    cms_elements_translations.id AS id,
    cms_elements_translations.locale_code AS locale_code,
    cms_elements_translations.body AS body,
    # This is a Window Function (for demonstration, only):
    ROW_NUMBER() OVER (PARTITION BY id ORDER BY (CASE
      # These conditions need to be build based on HTTP_ACCEPT_LANGUAGE:
      WHEN cms_elements_translations.locale_code LIKE 'en-US' THEN 0
      WHEN cms_elements_translations.locale_code LIKE 'de-__' THEN 1
      WHEN cms_elements_translations.locale_code LIKE 'en-__' THEN 2
      WHEN cms_elements_translations.locale_code LIKE 'fr-__' THEN 3
      WHEN cms_elements_translations.locale_code LIKE 'de-DE' THEN 4
    END) ASC) AS rank
  FROM cms_elements_translations
  # These are some cms_elements.id from contain():
  WHERE cms_elements_translations.id IN (
    '1bc23a23-d9ec-4a48-981d-27a74a8b2446', 'ef6368e3-92f4-47cd-946e-f2633ac36016')
)
SELECT * FROM cms_elements_translations
WHERE cms_elements_translations.rank = 1;
```

### What happened

... cannot easily create this with the 3.x ORM builder. I cannot use an SQL view for the CTE as the CASE WHEN conditions within the window function are programmatically build by reading HTTP_ACCEPT_LANGUAGE

A way to add/prepend CTEs to queries.

I have found `Query::epiloque()` but there is neither `Query::prepend()` (raw) nor `Query::addCommonTableExpression()` or similar.

### Workaround/Hack

Use derived tables, aka `->from(['CmsElementsTranslations' => $rankingQuery])`, then I just had to convert the window function to CakePHP3 ORM, which brought its own issues: https://github.com/cakephp/cakephp/issues/10670

### Vendor Docs for Common Table Expressions
- MariaDB 10.2 https://mariadb.com/kb/en/mariadb/non-recursive-common-table-expressions-overview/
- Postgres (8.4+) https://www.postgresql.org/docs/9.2/static/queries-with.html
- MS SQL Server https://msdn.microsoft.com/de-de/library/ms175972.aspx as per @stmeyer 
- Oracle http://docs.oracle.com/cd/E11882_01/server.112/e41084/statements_10002.htm#SQLRF30048

ndm2:
I had to implement support for CTE's in a recent project, and I didn't stumble over any real problems, so I thought I could put something together that might fit into the core, and get some feedback on it.

This is pretty much directly copied from the project, with minor adjustments, and some additional tests and cross schema compatibility added on top, so it's probably a little rough around the edges, and there might be a few parts that don't really adhere to the conventions.

I don't really like how I had to cram in these `WithExpression::*Keywords()` methods to be able to transform the output for SQL Server, they feel rather forced.

Not sure how keen people are on MariaDB support, but it's listed as supported in the docs, hence these awkward version checks.

---

Basically this introduces `WITH` clause support for the following formats:

```
WITH ... SELECT
WITH ... UPDATE
WITH ... INSERT
WITH ... DELETE
```

with the `SELECT` variant supporting usage in subqueries too.

Using recursive CTEs is supported, ie the `WITH RECURSIVE ...` format, and modifiers (currently only supported by Postgres if I'm not mistaken) are as well, ie `WITH cte AS MATERIALIZED ...`.

Furthermore parameterized CTEs are supported, ie `WITH cte(col1, col2) ...`.

Things are split into two new expressions:

* `\Cake\Database\Expression\WithExpression`
* `\Cake\Database\Expression\CommonTableExpression`

The former hosting the latter on the query objects. `\Cake\Database\Query::with()` is added to accept common table expressions, which can be created via the helper method `\Cake\Database\Query::commonTableExpression()` (similar to the `identifier()` helper method).

---

Basic example:

```php
$cteQuery = $connection
    ->newQuery()
    ->select('*')
    ->from('employees')
    ->where([
        'dept' => 'Engineering'
    ]);

$query = $connection
    ->newQuery()
    ->with(function (QueryExpression $exp, Query $query) use ($cteQuery) {
        return $query
            ->commonTableExpression('engineers', $cteQuery);
    })
    ->select('*')
    ->from('engineers');
```

```sql
WITH engineers AS (
    SELECT *
    FROM employees
    WHERE dept = 'Engineering'
)
SELECT *
FROM engineers
```

Complex example:

```php
$anchorQuery = $connection
    ->newQuery()
    ->select([
        'sub_part', 'part'
    ])
    ->where([
        'part' => 'our_product'
    ]);

$recursiveQuery = $connection
    ->newQuery()
    ->select([
        'p.sub_part', 'p.part'
    ])
    ->from([
        'pr' => 'included_parts',
        'p' => 'parts'
    ])
    ->where(function (QueryExpression $exp) {
        return $exp->equalFields('p.part', 'pr.sub_part');
    });

$cteQuery = $anchorQuery->unionAll($recursiveQuery);

$query = $connection
    ->newQuery()
    ->with(function (QueryExpression $exp, Query $query) use ($cteQuery) {
        return $query
            ->commonTableExpression('included_parts', $cteQuery)
            ->setFields([
                'sub_part', 'part'
            ])
            ->setRecursive(true);
    })
    ->delete()
    ->from('parts')
    ->where([
        'part IN' => $connection
            ->newQuery()
            ->select('part')
            ->from('included_parts')
    ]);
```

```sql
WITH RECURSIVE included_parts(sub_part, part) AS (
    SELECT sub_part, part
    FROM parts
    WHERE part = 'our_product'
  UNION ALL
    SELECT p.sub_part, p.part
    FROM included_parts pr, parts p
    WHERE p.part = pr.sub_part
)
DELETE FROM parts
WHERE part IN (SELECT part FROM included_parts)
```


othercorey:
Can you give a very quick summary of what this introduces?

src/Database/Driver.php
```diff
@@ -256,6 +263,16 @@ public function supportsSavePoints(): bool
         return true;
     }
 
+    /**
+     * Returns true if the server supports common table expressions.
+     *
+     * @return bool
+     */
+    public function supportsCommonTableExpressions(): bool
```
ADmad, line 271: Nitpick: The method and related property name is a bit too long. Can be shorten it to `supportsCTE()` or `supportsCte()`?

ndm2:
> Can you give a very quick summary of what this introduces?

Sure, I've updated the description.

src/Database/Driver/Mysql.php
```diff
@@ -230,6 +231,32 @@ public function supportsDynamicConstraints(): bool
...
+     * Receives a WithExpression and changes it so that it conforms to this
+     * SQL dialect.
+     *
+     * @param \Cake\Database\Expression\WithExpression $expression The expression to transform.
+     * @return void
+     */
+    protected function _transformWithExpression(WithExpression $expression): void
+    {
+        foreach ($expression->getExpressions() as $expression) {
+            $expression->setModifiers([]);
```
othercorey, line 256: Should we throw an exception here? I don't think it's normal cake convention to hide unsupported sql, is it?

src/Database/Driver.php
ndm2: If you insist :)

src/Database/Driver/Mysql.php
```diff
@@ -245,6 +272,28 @@ public function getVersion(): string
...
+    /**
+     * Returns true if the server supports common table expressions.
+     *
+     * @return bool
+     */
+    public function supportsCommonTableExpressions(): bool
+    {
+        if ($this->supportsCommonTableExpressions === null) {
+            $version = $this->getVersion();
+            if (strpos($version, 'MariaDB') !== false) {
```
othercorey, line 284: We should make this a `isMariaDb()` helper if we're going to support it.

src/Database/Driver/Mysql.php
ndm2: I'm not sure, I guess most transformers do usually transform into an equivalent. Gotta have a look at it.

src/Database/Query.php
```diff
@@ -369,6 +372,58 @@ public function traverseParts(callable $visitor, array $parts)
...
     }
 
+    /**
+     * Creates a common table expression (CTE).
+     *
+     * @param string $name The CTE name.
+     * @param \Cake\Database\ExpressionInterface|\Closure|string $query The CTE query definition.
+     * @return \Cake\Database\Expression\CommonTableExpression
+     */
+    public function commonTableExpression(string $name, $query): CommonTableExpression
```
othercorey, line 382: I'm thinking of the complex SQL statements that would use this. I think `cte()` might be much cleaner even if it's less verbose.

src/Database/Expression/CommonTableExpression.php
```diff
@@ -0,0 +1,297 @@
...
+     * The modifiers to use for the CTE.
+     *
+     * @var \Cake\Database\ExpressionInterface[]|string[]
+     */
+    protected $modifiers = [];
+
+    /**
+     * The CTE query definition.
+     *
+     * @var \Cake\Database\ExpressionInterface|string
```
othercorey, line 41: What is the use case for supporting raw query strings?

ndm2: I didn't really had one in mind, I've just added it because of all the other places in the query builder and the expressions that accept raw SQL additionally to expressions.

src/Database/Query.php
ndm2: That's two votes for CTE then...

othercorey:
This looks good in general! I'm going to finish reviewing when I have time -- expect us to nitpick so it feels consistent.

Thanks!

src/Database/Expression/CommonTableExpression.php
ndm2: If we want to go a more stricter route and require expressions, then I'm all in :)

src/Database/Driver/Mysql.php
othercorey: We could use it for window functions if we keep this.

src/Database/Driver/Mysql.php
othercorey: Unless there is a different sql syntax, I prefer that we simply generate the sql regardless. Users will receive a sql error if they use unsupported features.

src/Database/Driver.php
ADmad: Pretty please :slightly_smiling_face:

src/Database/Expression/CommonTableExpression.php
```diff
@@ -0,0 +1,297 @@
...
+        foreach ($fields as $index => $field) {
+            if (is_string($field)) {
+                $fields[$index] = $field = new IdentifierExpression($field);
+            }
+
+            if (!($field instanceof ExpressionInterface)) {
+                throw new InvalidArgumentException(sprintf(
+                    'The `$fields` argument must contain only instances of `%s`, or strings, `%s` given at index `%d`.',
+                    ExpressionInterface::class,
+                    Debugger::getType($field),
```
ADmad, line 102:
```suggestion
                    getTypeName($field),
```

src/Database/Expression/CommonTableExpression.php
```diff
@@ -0,0 +1,297 @@
...
+        return $this;
+    }
+
+    /**
+     * Constructor.
+     *
+     * @param string $name The CTE name.
+     * @param \Cake\Database\ExpressionInterface|string $query The CTE query definition.
+     */
+    public function __construct(string $name, $query)
```
ADmad, line 214: Can you move the constructor at the top please.

src/Database/Expression/CommonTableExpression.php
```diff
@@ -0,0 +1,297 @@
...
+    {
+        $this->setName($name);
```
othercorey, line 216: Since the name is required, I would make it non-optional in the constructor and remove the `setName()` helper.

src/Database/Query.php
```diff
@@ -369,6 +372,58 @@ public function traverseParts(callable $visitor, array $parts)
...
+                throw new \InvalidArgumentException(
+                    'Resetting the WITH clause only works when overwriting is enabled.'
+                );
+            }
+
+            return $this;
+        }
+
+        if (!$this->_parts['with']) {
+            $this->_parts['with'] = new WithExpression();
```
othercorey, line 415: I don't think I understand why a separate `WithExpression` is needed. Why can't we just store an array of CommonTableExpressions?

src/Database/Query.php
inoas: I don't mind the verbosity as it is not much interface. Only if the interface is huge long names become an issue imho as they clump up app source everywhere.

src/Database/Expression/CommonTableExpression.php
inoas: Sounds like a 5.x feature to drop string support over objects for all expressions?

src/Database/Query.php
ndm2: In the project where this stems from I had to test for recursive expressions quite a lot, and I didn't had to support multiple dialects. On project level I usually use concrete expressions for everything, as they're easy to test/reuse/extend/transform, extending the compiler on project level is always a little annoying.

Since the core needs compiler adjustments anyways, and given the fact that in the end it's just a stack of expressions, moving responsibilities into the compilers shouldn't be a problem.

src/Database/Driver/Mysql.php
ndm2: Would you prefer if the version string was prepared accordingly then, so that it returns the MariaDB version number only? Currently for MariaDB it can return something like this:

```
5.5.5-10.3.22-MariaDB-0ubuntu0.16.04.1
```

where `5.5.5` is the MySQL version that is being replicated, and `0ubuntu0.16.04.1` either the build environment or the server environment, not really sure.

src/Database/Driver/Mysql.php
ndm2: Fair enough. It can be revisited in case other dialects implement equivalent modifiers

src/Database/Expression/CommonTableExpression.php
ndm2: done

src/Database/Expression/CommonTableExpression.php
ndm2: done

src/Database/Query.php
ndm2: done

src/Database/Driver/Mysql.php
ndm2: done

src/Database/Driver.php
ndm2: done

src/Database/Expression/CommonTableExpression.php
ndm2: done

ndm2:
I've moved this out of the draft stage as the CI wouldn't run properly.

src/Database/Driver/Mysql.php
ndm2: Perhaps an additional method for specifically retrieving the MariaDB version might make sense, something like `getMariaDbVersion()`, which, given the above version string, would return `10.3.22`, while `getVersion()` would return `5.5.5`?

src/Database/Query.php
othercorey: As discussed in the other PR, we could do something similar here:

`with('name', $cte)` with closure support.

That would move the CTE name to the query. I don't know if that would cause other issues.

src/Database/Expression/CommonTableExpression.php
markstory: SQL snippets are useful when getting the querybuilder to generate the SQL you want is to convoluted or too complicated. With that said we could/should use a wrapper object so that one can't accidentally enable raw-sql mode.

src/Database/Driver/Mysql.php
othercorey: I don't know mariadb, but I think returning the maria-db specific version is best. Relying on (probably) outdated emulation version would be trouble.

`isMariaDb()` and `version()`

src/Database/Query.php
ndm2: If there was only that signature, then pre-constructing CTEs would be rather awkward, either the name would need to be completely removed from the CTE, which feels weird given how essential the name is for a CTE, or the name would need to be made an optional property.

But pre-constructing CTEs without names has a weird vibe too, and from a syntactic point of view, the CTE query definition is required too, also it can refer to the name of the CTE, and refering to that name before it's even defined for the CTE itself, it's not a technical problem, but again feels rather weird.


```php
$cte2 = new CommonTableExpression($cteQuery1);
$cte2 = new CommonTableExpression(
    $connection->newQuery()->select('*')->from('cte1')->...
);
$cte3 = new CommonTableExpression(
    $connection->newQuery()
        ->select('*')
        ->from('cte2')
        ->unionAll(
            $connection->newQuery()
                ->select('*')
                ->from('cte3')
                ->...
        )
);

// ...

$query
    ->with('cte1', $cte1)
    ->with('cte2', $cte2)
    ->with('cte3', $cte3);
```

```php
$query->with('name', function (CommonTableExpression $cte, Query, $query) {
    return $cte
        ->setQuery($query->getConnection->newQuery()->...);
});
```

```php
$query->with('name', function (CommonTableExpression $cte, Query, $query) {
    return $query
        ->cte('name', $query->getConnection->newQuery()->...)
});
```

```php
$query->with('name', function (CommonTableExpression $cte, Query, $query) {
    return new CommonTableExpression(
        $query->getConnection->newQuery()->...
    );
});
```

`with()` could of course also be made to accept multipe signatures, like `with($cte, bool)` _and_ `with('name', $cte, bool)`, but I feel like this would make things just more confusing.

I have to think about this for a litte bit longer. I'll start with dropping `WithExpression`.

src/Database/Driver/Mysql.php
ndm2: I'm not overly familiar with MariaDB either, but that sounds like a fair point.

src/Database/Expression/CommonTableExpression.php
ndm2: Makes sense, and creating `QueryExpression` objects is easy enough.

ndm2: done

src/Database/Query.php
othercorey: If the name needs to be part of the cte then that's fine. But, once we decide on a signature for the closure, it's not going to change.

ndm2: I've tried various variants back and forth, and I'm afraid it just doesn't seem to fit like it does for the window expression, which has the advantage that it's syntactically valid even if it's empty, ie `OVER ()` and `win AS ()` are both valid.

I think if we wanted to go with passing an instance to the callable, then both the name and the query would need to be made optional and settable. Hover that deviates from other expressions, which do require the essential parts that are neccesary to build valid SQL, so it might not be overly appreciated. On the other hand, constructing nameless CTEs is weird, and having the name "optional" (by moving it out of the expression) but the query being required, is even weirder, and if the query would be made optional too, well then we're back to square one :)

othercorey: Let's not make it overly complicated/difficult to use. We can just allow setting the name outside the constructor and if you don't, it's an invalid expression.

ndm2: I tend to overthink things... Not sure if I understood you right, you mean like this?

```php
with(function (CommonTableExpression $cte, Query $query) {
    return $cte
        ->setName('cte1')
        ->setQuery(...);
})
```

othercorey: Yep. I was wrong earlier when I said `setName()` isn't needed because the name is required. This let's us use consistent conventions in places like the closure, I think.

ndm2: All right, let's go with that and see if there's any objections.

src/Database/Expression/CommonTableExpression.php
```diff
@@ -0,0 +1,265 @@
...
+        return $this->query;
+    }
+
+    /**
+     * Sets the CTE query definition.
+     *
+     * @param \Cake\Database\ExpressionInterface $query The CTE query definition.
+     * @return $this
+     */
+    public function setQuery(ExpressionInterface $query)
```
othercorey, line 165: Should we require a Query instead of an expression? Is there a valid non-query expression for CTE? Should we support `VALUES()` expressions?

ndm2: That might be a little too restrictive, for raw SQL one would usually use `QueryExpression`, and forcing custom expressions to extend `Query` might not be overly appreciated.

`VALUES()` is supported in Postgres and Sqlite, so I'd say it should be supported. There's currently no clean way to to create it with queries or directly via `ValuesExpression` though.

othercorey: We could support `QueryExpression|ValuesExpression` in the docblock, but this is good for now.

tests/TestCase/Database/CommonTableExpressionIntegrationTest.php
```diff
@@ -0,0 +1,942 @@
...
+            ->where(['col !=' => 3], ['col' => 'integer']);
+        $cteQuery = $query1->unionAll($query2);
+
+        $query = $this->connection
+            ->newQuery()
+            ->with(function (QueryExpression $exp, Query $query) use ($cteQuery) {
+                return $query
+                    ->cte('cte', $cteQuery)
+                    ->setFields(['col'])
+                    ->setRecursive(true);
```
markstory, line 203: Should this method be an enableRecursive/disableRecurisve pair?

tests/TestCase/Database/CommonTableExpressionIntegrationTest.php
```diff
@@ -0,0 +1,942 @@
...
+    }
+
+    public function testWithOverwrite(): void
+    {
+        $query = $this->connection->newQuery();
+
+        $this->assertEmpty($query->clause('with'));
+
+        $query
+            ->with(function (QueryExpression $exp, Query $query) {
```
markstory, line 387: If the CommonTableExpression had setName() and setQuery() methods would we need Query::cte()?

src/Database/Query.php
```diff
@@ -369,6 +371,70 @@ public function traverseParts(callable $visitor, array $parts)
...
     }
 
+    /**
+     * Creates a common table expression (CTE).
+     *
+     * @param string $name The CTE name.
+     * @param \Cake\Database\ExpressionInterface|\Closure $query The CTE query definition.
+     * @return \Cake\Database\Expression\CommonTableExpression
+     */
+    public function cte(string $name, $query): CommonTableExpression
```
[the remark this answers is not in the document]
othercorey: I'm not sure I see the need for a helper. Users can just `new CommonTableExpression()`.

src/Database/SqlserverCompiler.php
```diff
@@ -49,10 +49,30 @@ class SqlserverCompiler extends QueryCompiler
...
+     * @return string
+     */
+    protected function _buildWithPart(array $parts, Query $query, ValueBinder $generator): string
+    {
+        $expressions = [];
+        foreach ($parts as $expression) {
+            $expressions[] = $expression->sql($generator);
+        }
+
+        return sprintf('WITH %s ', implode(', ', $expressions));
```
othercorey, line 73: We don't need this override. Users shouldn't create a recursive cte with sqlserver. It won't execute properly anyway.

tests/TestCase/Database/CommonTableExpressionIntegrationTest.php
othercorey: I wonder if we should follow the sql builder convention of `query()` and `modifier()`.

src/Database/SqlserverCompiler.php
ndm2: SQL Server does support recursive CTEs, it's just that unlike other DBMS (and contrary to the SQL standard) it just doesn't use the `RECURSIVE` keyword.

If we'd drop this, then we're forcing people to write rather DBMS specific queries when it's not really necessary.

tests/TestCase/Database/CommonTableExpressionIntegrationTest.php
ndm2: I was unsure about that, as it doesn't really enable/disable recursion for the CTE itself, but rather flags it at using a recursive query.

src/Database/Query.php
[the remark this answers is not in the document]
ndm2: I'd agree, I've already removed it in the process of changing the `with()` signature, as it now seems to be less relevant.

I'm still skeptical about the `with($name, $cte)` syntax, my feeling was that `with($cte)` would still be needed in order to make preconstructing CTEs not overly awkward, but the general trend seemed to be to move away from these overly overloaded signatures with varying argument numbers.

`withNamed()` gives me a weird vibe, it kinda sounds as if there is something like unnamed common table expressions 🤔 

Let me push the current state, and let's see how it works for you, and if you still feel like there's additional syntax needed.

src/Database/SqlserverCompiler.php
othercorey: I was not familiar with sql server recursive notation. As long as we aren't hiding bad sql then this looks good to me.

I think another approach is modifying the expression on compilation ... but this looks clean.

tests/TestCase/Database/CommonTableExpressionIntegrationTest.php
othercorey: Yea, enable/disable doesn't work because you don't want to "enable" a recursive cte without it being a recursive cte. Enabling implies you allow CTE's to be recursive instead of flagging them.

tests/TestCase/Database/CommonTableExpressionIntegrationTest.php
ndm2: Someone should probably write down the required conventions, AFAIR it was to use non `get/set` in the query builder only? Guess it's debatable whether expressions are close enough to the builder to be affected by that convention, but currently it seems that expressions generally use the `get/set` notation.

src/Database/Driver/Mysql.php
ndm2: I could squeeze this in here, but maybe it should rather be in a separate PR, and accompanied by adding at least _one_ test run with an actual MariaDB backend?

tests/TestCase/Database/CommonTableExpressionIntegrationTest.php
markstory: We generally don't use get/set in the query builder, as we didn't want to add 'noise words' when you're dealing with SQL concepts. The builder is generally all 'set' anyways.

tests/TestCase/Database/CommonTableExpressionIntegrationTest.php
markstory: Ok. setRecursive() works for me.

tests/TestCase/Database/CommonTableExpressionIntegrationTest.php
ndm2: I take that as a "*use get/set for expressions*"...

src/Database/Expression/CommonTableExpression.php
```diff
@@ -0,0 +1,313 @@
...
+    {
+        $this->recursive = $recursive;
+
+        return $this;
+    }
+
+    /**
+     * Converts the Node into a SQL string fragment.
+     *
+     * @param \Cake\Database\ValueBinder $generator Placeholder generator object
```
othercorey, line 227: You can use `{@inheritDoc}` to just add throws tags.

ndm2: I've tried that but PHPCS still complained, trying it again it works. Might have been a different error and I just didn't pay attention.

othercorey: You will get a mix of errors depending on what else is in the docblock since cakephp-codesniffer can't tell what it *should* be.

ndm2: Now PHPCS is happy, but Stickler complains again. @markstory Does Stickler use the different rules than the default CS check?

src/Database/Expression/CommonTableExpression.php
```diff
@@ -0,0 +1,311 @@
...
+    public function setModifiers(array $modifiers)
+    {
+        foreach ($modifiers as $index => $modifier) {
+            if (
+                !($modifier instanceof ExpressionInterface) &&
+                !is_string($modifier)
+            ) {
+                throw new InvalidArgumentException(sprintf(
+                    'The `$modifiers` argument must contain only instances of `%s`, or strings, ' .
+                        '`%s` given at index `%d`.',
```
markstory, line 165: Nice error message :clap:

markstory (approved):
Looks great to me. Is there any other feedback from others that needs to be addressed?

src/Database/Driver/Mysql.php
othercorey: I think we'd leave out MariaDB-specific code and expect that users would extend the Mysql driver to provide specific checks/support.

othercorey:
> Looks great to me. Is there any other feedback from others that needs to be addressed?

The only thing I'm not sure about the interface is `setModifiers()`. I think we should just have specific `set()` functions for materialized/not materialized. It's also weird that it accepts an expression.

Anything else we can clean up later.

othercorey:
> The only thing I'm not sure about the interface is `setModifiers()`. I think we should just have specific `set()` functions for materialized/not materialized. It's also weird that it accepts an expression.

I'll fix this in a follow-up PR.

PR #14527, opened 2020-05-01
4.next - Added IntervalExpression to handle datetime interval addition/subtraction in the orm

jimbo2150:
This PR adds support for datetime intervals without having to use a workaround and will translate the syntax across different servers.

The MySQL syntax is `dateOrField + INTERVAL 2 YEAR + INTERVAL 3 DAY`
The PostgrSQL syntax is `dateOrField + INTERVAL '2 YEAR 3 DAY'`
The SQLite syntax is `strftime('%Y-%m-%d %H:%M:%f', dateOrField, '2 year', '3 day')`
- The strfttime syntax is for the use of fractional time. Other functions such as `date()` do not support fractions of a second. The result can then be passed through date/time conversion to get the desired output.

The MSSQL syntax is `date_add('DAYS', 3, date_add('YEARS', 2, dateOrField))`

Example usage:
```php
$interval = \DateInterval::createFromDateString('2 year + 3 day');
$iExp = new IntervalExpression(new FrozenTime('2020-04-17 02:03:02.21'), $interval);
(new Query($connection))->select([ $iExp ]);
```
Would produce output like
```sql
SELECT '2020-04-17 02:03:02.21' + INTERVAL 2 YEAR + INTERVAL 3 DAY;
```
in MySQL and should produce a result `2022-04-20 02:03:02.21`.

The requiring of a date or expression was necessary for supporting other databases since SQLite and MSSQL do not support INVERVAL expressions, only date modifier functions which require an expression or date to be passed as a parameter.

Special and relative intervals such as "Second Thursday in October" are not supported because there is no meaningful way to acquire that interval from the PHP interval object and most databases do not support relative/special intervals.


src/Database/Expression/IntervalExpression.php
```diff
@@ -0,0 +1,309 @@
...
+     * Key mappings for DateInterval array conversion.
+     *
+     * @var array
+     */
+    protected const KEYS_REQUIRED = ['y' => 1, 'm' => 1, 'd' => 1, 'h' => 1, 'i' => 1, 's' => 1, 'f' => 1];
+
+    /**
+     * SQL unit values for DateInterval array conversion.
+     *
+     * @var array
```
dereuromark, line 42: string[]

src/Database/Expression/IntervalExpression.php
```diff
@@ -0,0 +1,309 @@
...
+     * intervals.
+     *
+     * @param DateInterval $interval An interval object.
+     * @throws Exception When interval value is not valid.
+     * @return self
+     */
+    public function setInterval(DateInterval $interval): self
+    {
+        $isValid = ($interval->y + $interval->m + $interval->d +
+            $interval->h + $interval->i + $interval->s + $interval->f) != 0;
```
dereuromark, line 218: !==

src/Database/Expression/IntervalExpression.php
```diff
@@ -0,0 +1,309 @@
...
+        if ($isValid == false) {
```
dereuromark, line 219: if (!$isValid)

src/Database/Expression/IntervalExpression.php
jimbo2150: Corrected

src/Database/Expression/IntervalExpression.php
jimbo2150: Corrected

othercorey:
Is the goal to use this with WindowExpressions or just adding for general use?

jimbo2150:
> Is the goal to use this with WindowExpressions or just adding for general use?

Both. Though WindowExpressions will be a second iteration since use with a date requires a value to be added due to not all servers supporting the interval expression.

src/Database/Driver/Postgres.php
```diff
@@ -297,4 +299,22 @@ public function newCompiler(): QueryCompiler
...
+     *   to postgres SQL.
+     * @return void
+     */
+    protected function transformIntervalExpression(IntervalExpression $intervalExp): void
+    {
+        $intervalExp->combineIntervalSqlOptions([
+            'glue' => ' ',
+            'sql-prefix' => 'INTERVAL ',
+            'prefix' => '\'',
+            'suffix' => '\'',
```
markstory, line 317: Could use `"'"` to avoid needing to escape quotes.

src/Database/Driver/Sqlite.php
```diff
@@ -336,4 +338,30 @@ public function supportsWindowFunctions(): bool
...
+     * Receives a IntervalExpression and changes it so that it conforms to this
+     * SQL dialect.
+     *
+     * @param IntervalExpression $intervalExp The interval expression to convert
+     *   to Sqlite SQL.
+     * @return void
+     */
+    protected function transformIntervalExpression(IntervalExpression $intervalExp): void
+    {
+        $fOrV = $intervalExp->getFieldOrValue();
```
markstory, line 352: Could we use a more readable variable name here? Perhaps `$fieldValue`?

src/Database/Driver/Sqlite.php
```diff
@@ -336,4 +338,30 @@ public function supportsWindowFunctions(): bool
...
+        $typeAry = [];
```
markstory, line 353: Could be `$types` instead.

src/Database/Driver/Sqlite.php
```diff
@@ -336,4 +338,30 @@ public function supportsWindowFunctions(): bool
...
+        if ($fOrV instanceof \DateTimeInterface) {
+            $typeAry[] = 'datetimefractional';
+        }
+        $sign = $intervalExp->getIntervalSign();
+        $fncExp = (new FunctionExpression('strftime', ['\'%Y-%m-%d %H:%M:%f\'' => 'literal']))
```
markstory, line 358: Could be `$function` instead of `fncExp`

src/Database/Expression/IntervalExpression.php
```diff
@@ -0,0 +1,316 @@
...
+     * @return array
+     */
+    public function getIntervalSqlOptions(): array
+    {
+        return $this->intervalSqlOptions;
+    }
+
+    /**
+     * Method that sets an overriding expression for servers that require
+     * the use of en expression (usually a function) instead of an interval
```
markstory, line 157:
```suggestion
     * the use of an expression (usually a function) instead of an interval
```

src/Database/Expression/IntervalExpression.php
```diff
@@ -0,0 +1,316 @@
...
+    }
+
+    /**
+     * Method that accepts a date or expression object that serves as the
+     * value the interval will be applied to.
+     *
+     * @param \DateTimeInterface|\Cake\Database\ExpressionInterface $fov A DateTimeInterface or ExpressionInterface object.
+     * @return self
+     */
+    public function setFieldOrValue($fov): self
```
markstory, line 187: Could this be 'target' or 'subject'? FieldOrValue is a lot of words to repeat.

markstory (commented):
Should interval expressions be exposed on the function builder?

I also question the need for all the public methods on `IntervalExpression` it creates a wide public API that doesn't seem to be needed based on what is in the tests. Making those methods protected/private would help us adapt to change more easily in the future.

src/Database/Expression/IntervalExpression.php
```diff
@@ -0,0 +1,329 @@
...
+        }
+        $this->reset();
+
+        return $sql;
+    }
+
+    /**
+     * @inheritDoc
+     */
+    public function traverse(Closure $callable): self
```
dereuromark, line 111: No typehint please on return, as this is chainable.

src/Database/Expression/IntervalExpression.php
```diff
@@ -0,0 +1,329 @@
...
+        return $this->intervalSqlOptions;
+    }
+
+    /**
+     * Method that sets an overriding expression for servers that require
+     * the use of an expression (usually a function) instead of an interval
+     * statement.
+     *
+     * @param \Cake\Database\ExpressionInterface|null $exp An ExpressionInterface object or null value.
+     * @return self
```
dereuromark, line 153: $this

src/Database/Expression/IntervalExpression.php
```diff
@@ -0,0 +1,329 @@
...
+     */
+    public function setOverrideExpression(?ExpressionInterface $exp): self
```
dereuromark, line 155: same here, no return typehint allowed for chainable

jimbo2150:
> Should interval expressions be exposed on the function builder?
> 
> I also question the need for all the public methods on `IntervalExpression` it creates a wide public API that doesn't seem to be needed based on what is in the tests. Making those methods protected/private would help us adapt to change more easily in the future.

@markstory, Do you mean within the driver-specific methods? I see your point about having so many exposed functions. I am going to split the expression and move the date code into it's own class since having a date involved is very different from one server to another. It should also make it easier to work with intervals for window expressions.

src/Database/Expression/IntervalExpression.php
jimbo2150: @dereuromark These are fixed locally. The editor wasn't warning so I looked and found the setting was still on PS12. Thanks for your help with these!

othercorey:
I think this will be easier if use a standard SQL format for intervals.

`DAYS HH:MM:SS.MICROSECOND`

jimbo2150:
> I think this will be easier if use a standard SQL format for intervals.
> 
> `DAYS HH:MM:SS.MICROSECOND`

@othercorey Thanks. I just tested it in pg and it does work even though I don't see that format in the documentation. I could use that to shorten the length of those that include days/hours/minutes/seconds/micro.

tests/TestCase/Database/Expression/IntervalExpressionTest.php
```diff
@@ -0,0 +1,120 @@
...
+     * @return void
+     */
+    public function testInterval()
+    {
+        // Query using direct date value
+        $iExp = new IntervalExpression(
+            new FrozenTime(self::$data['date'], self::$data['tz']),
+            self::$data['interval']
+        );
+        $query = new Query(self::$connection);
```
othercorey, line 85: This unit test is difficult to read. Can we follow the format used by other query tests without splitting up every line?

tests/TestCase/Database/Expression/IntervalExpressionTest.php
```diff
@@ -0,0 +1,120 @@
...
+     */
+    public static function tearDownAfterClass(): void
+    {
+        foreach (self::$data['tmpTable']->dropSql(self::$connection) as $query) {
+            self::$connection->query($query);
+        }
+        self::$connection = self::$data = null;
+        parent::tearDownAfterClass();
+    }
+
```
othercorey, line 72: This needs tests that validate the generated sql.

tests/TestCase/Database/Expression/IntervalExpressionTest.php
```diff
@@ -0,0 +1,120 @@
...
+    /**
+     * Tests interval values using an expression.
+     *
+     * @return void
+     */
+    public function testIntervalWithExpression()
+    {
+        // Create temporary table and populate with date
+        $query = new Query(self::$connection);
+        self::$connection->execute('INSERT INTO interval_test VALUES (\'' . self::$data['date'] . '\')');
```
othercorey, line 105: We don't need to do this. Just query an existing table like Comments.

tests/TestCase/Database/Expression/IntervalExpressionTest.php
```diff
@@ -0,0 +1,120 @@
...
+        // Query using subquery
+        $iExp = new IntervalExpression(
+            (new Query(self::$connection))->select(['interval_date'])->from('interval_test')->limit(1),
```
othercorey, line 108: We don't need to test subqueries.

jimbo2150:
Any assistance with SQL Server would be helpful as I am not as familiar with it. I ran a bunch of tests locally and the only issue I could find with it complaining about date conversion is if you include a timezone offset. I tried with formatting it as a string without the timezone and got the same error.

tests/TestCase/Database/Expression/IntervalExpressionTest.php
jimbo2150: I can collapse it into fewer lines. It is split up as I was doing debugging when I needed to see what sql was being produced. Most of the other tests don't actually run queries, however the output of an interval involving a date will be different depending on the server. I found that actually running a query then checking to make sure the result was valid was the best way to test it.

jimbo2150:
> Any assistance with SQL Server would be helpful as I am not as familiar with it. I ran a bunch of tests locally and the only issue I could find with it complaining about date conversion is if you include a timezone offset. I tried with formatting it as a string without the timezone and got the same error.

So it turns out SQL Server will throw that type error for datetimes including more than 3 fractional second digits unless you explicitly cast it as a `datetime2`.
https://superuser.com/questions/1026511/convert-datetime-with-decimal-time-longer-than-3-digits

tests/TestCase/Database/Expression/IntervalExpressionTest.php
```diff
@@ -0,0 +1,159 @@
...
+use Cake\Database\Expression\DateTimeIntervalExpression;
+use Cake\Database\Expression\IntervalExpression;
+use Cake\Database\ExpressionInterface;
+use Cake\Database\Query;
+use Cake\Database\Type;
+use Cake\Database\ValueBinder;
+use Cake\Datasource\ConnectionManager;
+use Cake\I18n\FrozenTime;
+use Cake\TestSuite\TestCase;
+use DateInterval as CakeDateInterval;
```
markstory, line 27: Why is this alias being used?

tests/TestCase/Database/Expression/IntervalExpressionTest.php
```diff
@@ -0,0 +1,159 @@
...
+    /**
+     * @var array
+     */
+    protected static $data = [];
+
+    public function setUp(): void
+    {
+        parent::setUp();
+        $this->connection = ConnectionManager::get('test');
+        self::$data['tz'] = $utc = new DateTimeZone('UTC');
```
markstory, line 57: Is there a reason you chose to use class properties instead of instance properties?

src/Database/DateInterval.php
```diff
@@ -0,0 +1,158 @@
...
+ * @license       https://opensource.org/licenses/mit-license.php MIT License
+ */
+namespace Cake\Database;
+
+use UnexpectedValueException;
+
+/**
+ * TODO: Add documentation
+ */
+class DateInterval extends \DateInterval
```
markstory, line 24: I didn't see this used in the tests anywhere, do we still need it?

src/Database/DateInterval.php
```diff
@@ -0,0 +1,158 @@
...
+
+    /**
+     * Constructor that takes an interval spec and an optional clone object. If the clone object is provided,
+     * the properties from the clone object are applied to the new object.
+     *
+     * @param string $interval_spec A string containing an interval specification.
+     * @param \DateInterval $clone Copy values from another DateInterval object.
+     * @throws \Exception
+     */
+    public function __construct($interval_spec, ?\DateInterval $clone = null)
```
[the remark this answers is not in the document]
jimbo2150: Corrected

tests/TestCase/Database/Expression/IntervalExpressionTest.php
jimbo2150: Corrected. Was left over from a previous test though not sure how it got on the base class. It does remind me that I was going to ask and make sure (question posted on the file review you did below).

src/Database/DateInterval.php
jimbo2150: Added tests and missing documentation.

jimbo2150: Also wanted to make sure: Do you think this class will be confusing for people? If so, I can rename it. I don't see any classes with the same name as their base and wanted to make sure there wasn't a requirement against that.

tests/TestCase/Database/Expression/IntervalExpressionTest.php
jimbo2150: Left over from when I was testing the static method. Changed to instance property.

src/Database/DateInterval.php
markstory: I wasn't sure why we needed this interval class. I couldn't figure that out from the tests/code.

markstory: It seems to duplicate some of the behavior in ChronosInterval. If folks need the additional precision beyond what PHP native objects provide they could use that package.

jimbo2150: I moved the `\DateInterval`-related code out of the `IntervalExpression` to keep it separate, so it's more of a utility class. It takes the `\DateInterval` properties (y, m, d, h, etc) and maps them into the database-specific units (YEAR, MONTH, DAY, HOUR, etc).

For comparison, `ChronosInterval` is used to reverse a `\DateInterval` object back into a period specification string `P1Y10D5H`.

This class instead creates an array `[ 'YEAR' => 1, 'DAY' => 10, 'HOUR' => 5 ]` that is used by `IntervalExpression` to generate `INTERVAL '1-0' YEAR_MONTH + INTERVAL '10 05:00:00.00' DAY_MICROSECOND` or the more specific `INTERVAL 1 YEAR + INTERVAL 10 DAY + INTERVAL 5 HOUR`. It also does some checks such as ensuring there is at least one valid value specified (creating an object with `P0Y` would not produce any change in a date it was applied to) as well as preventing relative or special formats (`second tuesday of march`) as those do not produce usable properties and are not supported by most databases. Since the methods are specific to database I left it under the `Cake\Database` namespace.

markstory:
The changes in #14738 should cover making interval expression with less complexity than a new expression object would.

jimbo2150:
> The changes in #14738 should cover making interval expression with less complexity than a new expression object would.

It does cover window expression date ranges but does not appear to cover regular date arithmetic (`SELECT dateField + INTERVAL ...` or date function equivalents in other databases). Since it leaves the user up to adding text for the actual range (`1 day`, `2 month`, etc), MySQL common syntax is expected; It may not work on SQL Server since it expects slightly different syntax.

Doesn't appear to be interest in the arithmetic portion; closing.

PR #14529, opened 2020-05-01
Use FQCN imports.

dereuromark:
Complements https://github.com/cakephp/cakephp/pull/14527

Thanks to my trailing comma sniff fixes, this should not create any merge conflicts.

### DIFF
--- a/src/Database/Driver/SqlDialectTrait.php
+++ b/src/Database/Driver/SqlDialectTrait.php
@@ -127,7 +127,7 @@ trait SqlDialectTrait
      * objects to conform with the specific SQL dialect. Keys are class names
      * and values a method in this class.
      *
-     * @return array
+     * @return string[]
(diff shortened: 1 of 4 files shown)


src/Database/Driver/SqlDialectTrait.php
```diff
@@ -127,7 +127,7 @@ public function queryTranslator(string $type): Closure
      * objects to conform with the specific SQL dialect. Keys are class names
      * and values a method in this class.
      *
-     * @return array
+     * @return string[]
      */
```
othercorey, line 131:
```suggestion
      @psalm-return array<class-string, string>
     */
```
I think we could get away with this here.

dereuromark: I added it into it, just in case.

PR #14531, opened 2020-05-02
4.next - Add required option support

### LINKED ISSUE #12498: Allow marking shell options as required
This is a (multiple allowed):

* [ ] bug
* [x] enhancement

* CakePHP Version: 3.6
* Platform and Target: N/A

### What you did
It is useful sometimes to specify cli options as required, but at the moment we cant. 

This is most relevant If a cli has multiple required inputs - in this case it is currently only possible to enforce via the command parser by using (positional) arguments. 

Example, consider a shell with this options parser:

```
    public function getOptionParser()
    {
        $parser = parent::getOptionParser();

        foreach(range(1,3) as $i) {
            $parser->addArgument("thing$i", [
                'help' => "Required thing $i",
                'required' => true
            ]);
        }

        return $parser;
    }
```

```
-> bin/cake my-cli --help
Usage:
cake my_cli [-h] [-q] [-v] <thing1> <thing2> <thing3>

Options:

--help, -h     Display this help.
--quiet, -q    Enable quiet output.
--verbose, -v  Enable verbose output.

Arguments:

thing1  Required thing 1
thing2  Required thing 2
thing3  Required thing 3

-> bin/cake my-cli
Error: Missing required arguments. thing1 is required.
```
Although helpful, having to specify args in order is sometimes pretty awkward.

It's be much easier with this example to create a parser like so:

```
    public function getOptionParser()
    {
        $parser = parent::getOptionParser();

        foreach(range(1,10) as $i) {
            $parser->addOption("thing$i", [ <---- changed
                'help' => "Required thing $i",
                'required' => true <---- ignored 
            ]);
        }

        return $parser;
    }
```

But that only achieves the syntax I'm looking for:

```
-> cake/bin/cake my-cli --help
Usage:
cake my_cli [-h] [-q] [-v] [--thing1] [--thing2] [--thing3]

Options:

--help, -h     Display this help.
--quiet, -q    Enable quiet output.
--verbose, -v  Enable verbose output.
--thing1       Required thing 1
--thing2       Required thing 2
--thing3       Required thing 3
```
It does not enforce `required` as options are, according to the Shell class, always optional:

```
-> cake/bin/cake my-cli
Hello now going to try and do something in the main method and fail because required things are missing...
```

So, the ask here is simply to allow marking options required.

markstory:
Allow CLI flags to be marked as required flags/options.

Fixes #12498

### DIFF
--- a/src/Console/ConsoleInputOption.php
+++ b/src/Console/ConsoleInputOption.php
@@ -217,7 +217,7 @@ class ConsoleInputOption
      */
     public function isRequired(): bool
     {
-        return (bool)$this->required;
+        return $this->required;
     }
 
     /**
--- a/src/Console/ConsoleInputOption.php
+++ b/src/Console/ConsoleInputOption.php
@@ -77,6 +77,13 @@ class ConsoleInputOption
     protected $_choices;
 
     /**
+     * Is the option required.
+     *
+     * @var bool
+     */
+    protected $required;
+
+    /**
      * Make a new Input Option
      *
      * @param string $name The long name of the option, or an array with all the properties.
@@ -86,6 +93,7 @@ class ConsoleInputOption
      * @param string|bool|null $default The default value for this option.
      * @param string[] $choices Valid choices for this option.
      * @param bool $multiple Whether this option can accept multiple value definition.
+     * @param bool $required Whether this option is required or not.
      * @throws \Cake\Console\Exception\ConsoleException
      */
     public function __construct(
@@ -95,7 +103,8 @@ class ConsoleInputOption
         bool $isBoolean = false,
         $default = null,
         array $choices = [],
-        bool $multiple = false
+        bool $multiple = false,
+        bool $required = false
     ) {
         $this->_name = $name;
         $this->_short = $short;
@@ -103,6 +112,7 @@ class ConsoleInputOption
         $this->_boolean = $isBoolean;
         $this->_choices = $choices;
         $this->_multiple = $multiple;
+        $this->required = $required;
 
         if ($isBoolean) {
             $this->_default = (bool)$default;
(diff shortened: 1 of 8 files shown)


src/Console/ConsoleInputOption.php
```diff
@@ -192,6 +210,16 @@ public function defaultValue()
...
     }
 
+    /**
+     * Check if this option is required
+     *
+     * @return bool
+     */
+    public function isRequired(): bool
+    {
+        return (bool)$this->required;
```
ADmad, line 220: The casting shouldn't be necessary.

src/Console/ConsoleOptionParser.php
```diff
@@ -434,7 +435,8 @@ public function addOption($name, array $options = [])
                 $options['boolean'],
                 $options['default'],
                 $options['choices'],
-                $options['multiple']
+                $options['multiple'],
+                $options['required'],
```
othercorey, line 439:
```suggestion
                $options['required']
```

tests/TestCase/Console/ConsoleOptionParserTest.php
```diff
@@ -330,6 +332,38 @@ public function testAddMultipleOptionsWithMultiple()
...
+            $result[0]
+        );
+
+        $result = $parser->parse(['--no-default', 'value']);
+        $this->assertSame(
+            ['no-default' => 'value', 'help' => false, 'test' => 'default value'],
+            $result[0]
+        );
+
+        $this->expectException(ConsoleException::class);
```
othercorey, line 363: If the first parse throws an exception, does phpunit mark the remaining asserts as fails?

markstory: Anything that happens before expectException should still get counted as passed. Nothing will happen after the exception is raised though.

othercorey: I prefer to have exception tests in a separate function to avoid capturing it for other tests so nothing is skipped.

markstory: Ok. I can split them up.

PR #14533, opened 2020-05-02
Simplify Debugger::getType()

ADmad:


dereuromark:
If you read all the issues this method has,
E.g. Comments in https://www.php.net/manual/de/function.gettype.php
Not sure if you want to switch to using it instead of thr separate checks.

ADmad:
All the comments on PHP manual are years old, who knows if they are even relevant now. Plus all our tests pass after the change. In fact after this change `resource` type is now reported instead of being shown as `unknown` earlier.

dereuromark:
https://wiki.php.net/rfc/get_debug_type php8

ADmad:
`is_object($bar) ? get_class($bar) : gettype($bar)`  that's what our `getTypeName()` function already does, which is what I have used.

dereuromark (approved):

ADmad:
Pushed another commit to make Debugger:export() DRYer too.

PR #14535, opened 2020-05-02
Fix test failure on MySQL 8

garas:
Do not assume database records are returned in order they were inserted.

Condition matches all except first comment.
MySQL 8 returned last one (`'Second Comment for Second Article'`), other databases returned `'Second Comment for First Article'` record. 
https://github.com/cakephp/cakephp/blob/06027e7fc10b68e31a0383b98f73470eb2a44d20/tests/Fixture/CommentsFixture.php#L45-L52


othercorey:
Perfect, thanks!

Issue #14469, opened 2020-04-17
Named windows and non-aggregate window functions


ndm2:
This is a (multiple allowed):

* [ ] bug
* [x] enhancement
* [x] feature-discussion (RFC)

* CakePHP Version: 4.x

---------------------------------

Continuation from https://github.com/cakephp/cakephp/issues/14080

Seeing the first implementation of window function support, I'm wondering whether there are any plans for adding support for [**named windows**](https://dev.mysql.com/doc/refman/8.0/en/window-functions-named-windows.html)? Looking at the current state, it would probably not possible to implement without breaking changes.

To be able to implement named window usage, it would need to be possible to generate `OVER` clauses like `OVER window_name` and `OVER (window_name further_window_spec_clauses...)`, and it it would need to be possible to generate a window specification without an over clause, so that it can fit in the `WINDOW` clause, like `WINDOW window_name AS (window_spec_clauses...)`.

Also would you share my concern that the fact that all window functions that the builder can create are implemented as `AggregateExpression`, could be a little misleading, and perhaps awkward to use, given that not all window functions actually are aggregate functions, like `RANK()`, `ROW_NUMBER()`, `LAG()`, `LEAD()`, etc...?

othercorey:
For named windows, I am not sure if we want to support extending the named window clause. That might be confusing interface for users who don't know there are special limitations. I think it would be minimal overhead for users to simply create the full window they want and name it.

Implementing named windows is something we would like to do.

othercorey:
For the concern about AggregateExpression being used for all window functions, we found this was a user-experience compromise.

We discussed creating separate FunctionsBuilder versions for aggregate functions/window functions, but found that to be confusing for the user. We would have to duplicate previous functions and enforce that the user call the right builder.

For now, we chose to use a single FunctionsBuilder with an `aggregate()` helper. If we added a `window()` helper, that would feel more confusing than helpful as it suggests you can't use aggregate functions.

So, generally, while AggregateExpression might have options in the future that don't strictly belong to a handful of window functions, it seems like the least confusing compromise.

othercorey:
Named windows ended up being easier to implement with full support. Users can chain windows freely as long as their database supports it.

ndm2:
That's great news 👍

othercorey:
Added in https://github.com/cakephp/cakephp/pull/14536

PR #14536, opened 2020-05-02
4.1 - Added named window expression support

othercorey:
Closes https://github.com/cakephp/cakephp/issues/14469.

The interface change is small, but the sql generation for window expressions had to change a lot.

- WindowExpression no longer generates the `OVER` clause
- AggregateExpression generates the `OVER` clause
- WindowExpression only generates sql when there is more than a name.

Because `OVER` and `WINDOW` clauses render window expressions differently, `WindowExpression` provides an `isEmpty()` helper.

# OVER clause

`SELECT COUNT(*) OVER name`  is different than `SELECT COUNT(*) OVER (name)`

We always generate an empty window expression that has a name without `()`. Users will not expect the edge case of `OVER (name)`.

# WINDOW clause

`WINDOW name AS (other_name)` is the only allowed syntax.

We always generate window clauses inside `()`.

# Examples

```
$query
    ->select('sum' => $query->func()->sum('field')->over('second_window')->order('field2')
    ->window('first_window', new WindowExpression())
    ->window('second_window', function ($expr) {
        return $expr->setName('first_window');
    });
```

### DIFF
--- a/src/Database/Query.php
+++ b/src/Database/Query.php
@@ -21,6 +21,7 @@ use Cake\Database\Expression\OrderByExpression;
 use Cake\Database\Expression\OrderClauseExpression;
 use Cake\Database\Expression\QueryExpression;
 use Cake\Database\Expression\ValuesExpression;
+use Cake\Database\Expression\WindowExpression;
 use Cake\Database\Statement\CallbackStatement;
 use Closure;
 use InvalidArgumentException;
@@ -85,6 +86,7 @@ class Query implements ExpressionInterface, IteratorAggregate
         'where' => null,
         'group' => [],
         'having' => null,
+        'window' => [],
         'order' => null,
         'limit' => null,
         'offset' => null,
@@ -1324,6 +1326,35 @@ class Query implements ExpressionInterface, IteratorAggregate
     public function andHaving($conditions, $types = [])
     {
         $this->_conjugate('having', $conditions, 'AND', $types);
+
+        return $this;
+    }
+
+    /**
+     * Adds a named window expression.
+     *
+     * You are responsible for adding windows in the order your database requires.
+     *
+     * @param string $name Window name
+     * @param \Cake\Database\Expression\WindowExpression|\Closure $window Window expression
+     * @param bool $overwrite Clear all previous query window expressions
+     * @return $this
+     */
+    public function window(string $name, $window, bool $overwrite = false)
+    {
+        if ($overwrite) {
+            $this->_parts['window'] = [];
+        }
+
+        if ($window instanceof Closure) {
+            $window = $window(new WindowExpression(), $this);
+            if (!($window instanceof WindowExpression)) {
+                throw new RuntimeException('You must return a `WindowExpression` from closure passed to `window()`.');
+            }
+        }
+
+        $this->_parts['window'][] = ['name' => $name, 'window' => $window];
+        $this->_dirty();
 
         return $this;
     }
--- a/src/Database/QueryCompiler.php
+++ b/src/Database/QueryCompiler.php
@@ -51,7 +51,7 @@ class QueryCompiler
      * @var array
      */
     protected $_selectParts = [
-        'select', 'from', 'join', 'where', 'group', 'having', 'order', 'limit',
+        'select', 'from', 'join', 'where', 'group', 'having', 'window', 'order', 'limit',
         'offset', 'union', 'epilog',
     ];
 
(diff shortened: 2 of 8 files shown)


src/Database/Query.php
```diff
@@ -1328,6 +1330,29 @@ public function andHaving($conditions, $types = [])
...
+     *
+     * If the window exprssion is copying from an existing window name, the
+     * target window expression must be added to the query first.
+     *
+     * @param string $name Window name
+     * @param \Cake\Database\Expression\WindowExpression $window Window expression
+     * @param bool $overwrite Clear all previous query window expressions
+     * @return $this
+     */
+    public function window(string $name, WindowExpression $window, bool $overwrite = false)
```
ndm2, line 1344: Any chance for closure support for `$window`, similar to the other query builder methods, so we could do stuff like:

```php
->window('win', function (Query $query) {
    return (new WindowExpression())
        ->partition($query->func()->foo([$query->identifier('bar')])); 
})
```

It would also give me one more reason to ask for expression support for frame bound offsets 😉

othercorey: I can add closure support, but as your example shows we'll have to break with the convention of passing a `QueryExpression`.

othercorey: Maybe we should add `WindowExpression::setName()` and pass the function a `new WindowExpression()` instance similar to how other builders pass an expression and expect it in return.

I was avoiding `->setName('other_window')->partition()` because I thought it looked odd.

ndm2: Oh I just omitted that by accident :) But I also wouldn't have a problem with it not being passed, one can always easily do `$query->newExpr()`.

othercorey: But `QueryExpression` has no use here. We'd need to pass a `WindowExpression` instance which they could return.

ndm2: You mean like this?

```php
->window(function (WindowExpression $window) {
    return $window->setName('win') ...
})
```

I guess that could work, it's similar enough to how you get expression objects in other methods.

othercorey: That's what I'm implementing. You're free to construct your own expression of course, but want a consistent experience.

ndm2: Ok, now I have to think about whether that might make sense for CTEs too :)

othercorey: That's true -- but CTE is awkward because the name is required, but I suppose you can create an invalid expression in other builders that accept a closure.

othercorey: This is the wrong place -- but while I'm thinking about it, we can just have: `->with('name', $cte)` which accepts either closure or expression. That would solve the awkwardness.

src/Database/Query.php
```diff
@@ -1328,6 +1330,32 @@ public function andHaving($conditions, $types = [])
...
+     * @return $this
+     */
+    public function window(string $name, $window, bool $overwrite = false)
+    {
+        if ($overwrite) {
+            $this->_parts['window'] = [];
+        }
+
+        if ($window instanceof Closure) {
+            $window = $window(new WindowExpression(), $this);
```
markstory, line 1350: We should raise an error when the closure doesn't return a windowexpression.

src/Database/QueryCompiler.php
```diff
@@ -270,6 +270,29 @@ protected function _buildJoinPart(array $parts, Query $query, ValueBinder $gener
...
+     * @param array $parts List of windows to be transformed to string
+     * @param \Cake\Database\Query $query The query that is being compiled
+     * @param \Cake\Database\ValueBinder $generator the placeholder generator to be used in expressions
+     * @return string
+     */
+    protected function _buildWindowPart(array $parts, Query $query, ValueBinder $generator): string
+    {
+        $windows = [];
+        foreach ($parts as $window) {
+            $expr = $window['window'];
```
markstory, line 285: Should this also have a runtime check to ensure we don't have a fatal error/type error on the method calls?

src/Database/Query.php
othercorey: I didn't see other builders do this. Is that something we should change or does it happen elsewhere?

src/Database/QueryCompiler.php
othercorey: If we add a type check in `Query::window()` we shouldn't need one here.

src/Database/Query.php
markstory: Association `queryBuilder` callbacks raise a `RuntimeException` if the callback does not return a Query instance.

PR #14538, opened 2020-05-03
Fix colours in dark terminals

markstory:
Don't use the black ansi code as it is hard to see in dark terminals. Instead use the foreground color which *should* be visible.

Screen Shot 2020-05-02 at 22 17 03


dereuromark:
Your colors are feeling awfully British today :)

But it looks like it is working, on my CLI anyway.

PR #14539, opened 2020-05-03
4.1 - Added exception for FunctionsBuilder::cast() when parameter is missing

othercorey:
Added exception when the backwards compatible parameter is missing.

Added description explaining the return type is not automatically set like other `FunctionsBuilder` wrappers.

### DIFF
--- a/src/Database/FunctionsBuilder.php
+++ b/src/Database/FunctionsBuilder.php
@@ -127,10 +127,13 @@ class FunctionsBuilder
     }
 
     /**
-     * Returns a FunctionExpression representing a call to SQL CAST function.
+     * Returns a FunctionExpression representing a SQL CAST.
+     *
+     * The `$type` parameter is a SQL type. The return type for the returned expression
+     * is the default type name. Use `setReturnType()` to update it.
      *
      * @param string|\Cake\Database\ExpressionInterface $field Field or expression to cast.
-     * @param string $type The target data type
+     * @param string $type The SQL data type
      * @return \Cake\Database\Expression\FunctionExpression
      */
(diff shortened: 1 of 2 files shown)


othercorey:
I'm not a big fan of this being in FunctionsBuilder, but since we don't have any way to cleanly map sql to db types moving it won't make it much cleaner.

src/Database/FunctionsBuilder.php
```diff
@@ -127,10 +127,13 @@ public function coalesce(array $args, array $types = []): FunctionExpression
     }
 
     /**
-     * Returns a FunctionExpression representing a call to SQL CAST function.
+     * Returns a FunctionExpression representing a SQL CAST.
+     *
+     * The `$type` parameter is a SQL type. The return type for the returned expression
+     * is the default type name. Use `setReturnType()` to update it.
```
markstory, line 133: Should we fix this caveat when using the new call signature? Seems like a good time to do it.

othercorey: I can't imply the return type from $type. I don't know how to fix the new signature.

markstory: Are we not able to use a mapping of common SQL types and their PHP equivalents?

othercorey: All we have is the sql dialect column conversion. We don't have any sort of default mapping.

othercorey: Do we want to create a generic sql-type-to-database-type mapper?

markstory: Probably not necessary given how frequently cast() is used it wouldn't be worth the investment.

PR #14541, opened 2020-05-03
Cleanup MiddlewareDispatcher.

ADmad:
Remove duplicate code and update type annoations.

### DIFF
--- a/src/TestSuite/MiddlewareDispatcher.php
+++ b/src/TestSuite/MiddlewareDispatcher.php
@@ -26,8 +26,6 @@ use Cake\Routing\RoutingApplicationInterface;
 use Laminas\Diactoros\Stream;
 use LogicException;
 use Psr\Http\Message\ResponseInterface;
-use ReflectionClass;
-use ReflectionException;
 
 /**
  * Dispatches a request capturing the response for integration
@@ -89,14 +87,11 @@ class MiddlewareDispatcher
         $this->_class = $class;
         $this->_constructorArgs = $constructorArgs ?: [CONFIG];
 
-        try {
-            $reflect = new ReflectionClass($this->_class);
-            /** @var \Cake\Core\HttpApplicationInterface $app */
-            $app = $reflect->newInstanceArgs($this->_constructorArgs);
-            $this->app = $app;
-        } catch (ReflectionException $e) {
-            throw new LogicException("Cannot load `{$this->_class}` for use in integration testing.", 0, $e);
+        if (!class_exists($this->_class)) {
+            throw new LogicException("Cannot load `{$this->_class}` for use in integration testing.", 0);
         }
+
+        $this->app = new $this->_class(...$this->_constructorArgs);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


ADmad:
Btw why does the constructor use reflection to create application instance instance instead of just using `new $this->_class(...$this->_constructorArgs)`?

src/TestSuite/MiddlewareDispatcher.php
```diff
@@ -48,7 +48,7 @@ class MiddlewareDispatcher
      * The application class name
      *
      * @var string
-     * @psalm-var class-string
+     * @psalm-var \Cake\Core\HttpApplicationInterface::class
```
othercorey, line 51:
```suggestion
     * @psalm-var class-string<\Cake\Core\HttpApplicationInterface>
```
This is the syntax that other parsers have adopted.

src/TestSuite/MiddlewareDispatcher.php
```diff
@@ -74,15 +74,19 @@ class MiddlewareDispatcher
      * @param array|null $constructorArgs The constructor arguments for your application class.
      *   Defaults to `['./config']`
      * @throws \LogicException If it cannot load class for use in integration testing.
-     * @psalm-param \Cake\Core\HttpApplicationInterface::class|\Cake\Core\ConsoleApplicationInterface::class|null $class
+     * @psalm-param \Cake\Core\HttpApplicationInterface::class|null $class
```
othercorey, line 77:
```suggestion
     * @psalm-param class-string<\Cake\Core\HttpApplicationInterface>|null $class
```
Same here.

src/TestSuite/MiddlewareDispatcher.php
```diff
@@ -74,15 +74,19 @@ class MiddlewareDispatcher
...
      */
     public function __construct(
         TestCase $test,
         ?string $class = null,
         ?array $constructorArgs = null
     ) {
         $this->_test = $test;
-        $this->_class = $class ?: Configure::read('App.namespace') . '\Application';
+        if ($class === null) {
+            /** @psalm-var \Cake\Core\HttpApplicationInterface::class */
```
othercorey, line 86:
```suggestion
            /** @psalm-var class-string<\Cake\Core\HttpApplicationInterface> */
```

src/TestSuite/MiddlewareDispatcher.php
```diff
@@ -74,25 +72,26 @@ class MiddlewareDispatcher
...
 
-        try {
-            $reflect = new ReflectionClass($this->_class);
-            /** @var \Cake\Core\HttpApplicationInterface $app */
-            $app = $reflect->newInstanceArgs($this->_constructorArgs);
-            $this->app = $app;
-        } catch (ReflectionException $e) {
-            throw new LogicException("Cannot load `{$this->_class}` for use in integration testing.", 0, $e);
+        if (!class_exists($this->_class)) {
+            throw new LogicException("Cannot load `{$this->_class}` for use in integration testing.", 0);
```
ADmad, line 91: This feels like a `RuntimeException` rather than `LogicException` to me but I didn't change it to keep existing test case passing.

PR #14543, opened 2020-05-04
Update Filesystem internal description

### LINKED ISSUE #14532: Should Filesystem be @internal?
Should we leave this marked internal if we're using it in external plugins?

https://github.com/cakephp/cakephp/blob/efd878ce4158eebcb3d124372d498ad947738aeb/src/Filesystem/Filesystem.php#L31

othercorey:
Closes https://github.com/cakephp/cakephp/issues/14532

This seems to be a commonly used helper in external code. 

Do we want to keep it internal or move it to a more general utility location so the filesystem namespace can be removed when File and Folder are removed?

### DIFF
--- a/src/Filesystem/Filesystem.php
+++ b/src/Filesystem/Filesystem.php
@@ -27,7 +27,11 @@ use RegexIterator;
 use SplFileInfo;
 
 /**
- * @since 4.0.0
+ * This provides convenience wrappers around common filesystem queries.
+ *
+ * This is an internal helper class that should not be used in application code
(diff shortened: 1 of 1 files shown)


ADmad:
The reason for adding `@internal` tag was do deter people from using it in their app code and us having to provide BC promises. Us using it in other core plugins is an informed decision and if the class gets changed we would change it's usage in the plugins too if required.

othercorey:
I agree that's reasonable. I've changed it to simply document the expected use better.

PR #14549, opened 2020-05-06
Removed @throws tags from IntegrationTestTrait http methods

othercorey:
Alternative to https://github.com/cakephp/cakephp/pull/14193

### DIFF
--- a/src/TestSuite/IntegrationTestTrait.php
+++ b/src/TestSuite/IntegrationTestTrait.php
@@ -379,7 +379,6 @@ trait IntegrationTestTrait
      *
      * @param string|array $url The URL to request.
      * @return void
-     * @throws \PHPUnit\Framework\Error\Error|\Throwable
      */
     public function get($url): void
     {
@@ -396,7 +395,6 @@ trait IntegrationTestTrait
      * @param string|array $url The URL to request.
      * @param string|array $data The data for the request.
      * @return void
-     * @throws \PHPUnit\Framework\Error\Error|\Throwable
      */
     public function post($url, $data = []): void
     {
@@ -413,7 +411,6 @@ trait IntegrationTestTrait
      * @param string|array $url The URL to request.
      * @param string|array $data The data for the request.
      * @return void
-     * @throws \PHPUnit\Framework\Error\Error|\Throwable
      */
     public function patch($url, $data = []): void
     {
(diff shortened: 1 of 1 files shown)


src/TestSuite/IntegrationTestTrait.php
```diff
@@ -478,7 +472,6 @@ public function head($url): void
      *
      * @param string|array $url The URL to request.
      * @return void
-     * @throws \PHPUnit\Framework\Error\Error|\Throwable
```
dereuromark, line 481: But this is not really correct, see my intial Comments on the original pr.

src/TestSuite/IntegrationTestTrait.php
```diff
@@ -509,11 +502,7 @@ protected function _sendRequest($url, $method, $data = []): void
                 $this->_requestSession->write('Flash', $this->_flashMessages);
             }
             $this->_response = $response;
-        } catch (PhpUnitError $e) {
-            throw $e;
-        } catch (DatabaseException $e) {
-            throw $e;
-        } catch (LogicException $e) {
+        } catch (PhpUnitError | DatabaseException | LogicException $e) {
```
garas, line 505: `PHPUnit\Framework\Error\Error as PhpUnitError` only catches PHP errors/warning/notices.

`\PHPUnit\Framework\Exception` class or `\PHPUnit\Exception` interface would be better as it catch all PHPUnit asserts too.

src/TestSuite/IntegrationTestTrait.php
dereuromark: As soon as you disableErrorHandlerMiddleware() you can actually expect the \Throwable at least.

src/TestSuite/IntegrationTestTrait.php
tfont: This is a great approach! Well done.

markstory: I don't think we want to catch phpunit assertion errors here though.

garas: I wanted to say "it should re-throw all PHPUnit asserts".

Currently they are catched in `\Throwable` block and rendered by `ExceptionRenderer`.

othercorey:
So, do we need to revert this?

PR #14551, opened 2020-05-06
4.1 - Added connection name to QueryLogger log entries

othercorey:
Alternative to https://github.com/cakephp/cakephp/pull/13954

If users need to add more complex metadata to their query logs, they'll have to extend and set a custom logger.

### DIFF
--- a/src/Database/Log/QueryLogger.php
+++ b/src/Database/Log/QueryLogger.php
@@ -46,7 +46,7 @@ class QueryLogger extends BaseLog
     {
         Log::write(
             'debug',
-            (string)$context['query'],
+            'connection=' . $this->getConfig('connection', '') . ' ' . (string)$context['query'],
             ['scope' => $this->scopes() ?: ['queriesLog']]
         );
     }
--- a/src/Database/Connection.php
+++ b/src/Database/Connection.php
@@ -905,7 +905,7 @@ class Connection implements ConnectionInterface
             );
         }
 
-        return $this->_logger = new QueryLogger();
+        return $this->_logger = new QueryLogger(['connection' => $this->configName()]);
     }
 
     /**
--- a/tests/TestCase/Database/ConnectionTest.php
+++ b/tests/TestCase/Database/ConnectionTest.php
@@ -70,12 +70,18 @@ class ConnectionTest extends TestCase
     protected $connection;
 
     /**
+     * @var use Cake\Database\Log\QueryLogger
+     */
+    protected $defaultLogger;
+
+    /**
      * @return void
      */
     public function setUp(): void
     {
         parent::setUp();
         $this->connection = ConnectionManager::get('test');
+        $this->defaultLogger = $this->connection->getLogger();
 
         $this->logState = $this->connection->isQueryLoggingEnabled();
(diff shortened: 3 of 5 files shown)


ADmad:
This can go into `4.next`.

othercorey:
I think users might expect we default the name to the connection manager config name.

ADmad (approved):
LGTM

othercorey:
Fixed `ConnectionTest` to restore original QueryLogger each test. It should log the connection `test` by default.

src/Database/Log/QueryLogger.php
```diff
@@ -36,6 +43,9 @@ public function __construct(array $config = [])
     {
         $this->_defaultConfig['scopes'] = ['queriesLog'];
 
+        $this->connection = $config['connection'] ?? '';
```
ADmad, line 46: Why use a separate property and unset `'connection'` key instead of just using `$this->getConfig('connection')` later on?

othercorey: I was trying to keep the overhead low, but it's probably minimal.

othercorey: I simplified it down to getting the config value.

src/Database/Log/QueryLogger.php
```diff
@@ -46,7 +46,7 @@ public function log($level, $message, array $context = [])
     {
         Log::write(
             'debug',
-            (string)$context['query'],
+            'connection=' . $this->getConfig('connection', '') . ' ' . (string)$context['query'],
             ['scope' => $this->scopes() ?: ['queriesLog']]
```
markstory, line 50: Should we be adding the connection name here, and using the new log formatting features that replace `{foo}` expressions to format the connection name into the message?

Having the connection name as a context value would allow structured logging adapters to use the connection name as data instead of text.

othercorey: I looked at adding it as a context value. The `LoggingStatement` would need the connection name to add it. Can we do it another way?

ADmad: You can  `$this->_driver->getConnection()->getName()` to get the connection name directly in `LoggingStatement`.

othercorey: Should we just provide the full connection object in the context then? Users can add any custom info they want to the connection like getting/setting connection ids.

ADmad: My mistake, `$this->_driver->getConnection()` return the driver's `PDO` instance. So without modifying the `LoggingStatement` there's not way to pass the connection name to the logger.

markstory: I'm ok with the constructor argument to QueryLogger, my suggestion was that instead of appending the string here, we pass the connection as a context value. I can do that in a follow up pull request though, as it would be good to move duration and numRows into context to for consistency.

ADmad: > ..it would be good to move duration and numRows into context to for consistency.

Duration and numRows are already available through context since the last round of cleanup I did for query logging. Just they are not set directly set as context keys but available through the `LoggedQuery` instance which is passed to context.

PR #12693, opened 2018-10-30
Add test case for issue #12440

dereuromark:
@ADmad is this still sth to PR? or can we remove the branch?

Refs:  #12440

PR #14555, opened 2020-05-06
Throw exception when contain() in query builder for matching

### LINKED ISSUE #12440: Query error when using contain() inside innerJoinWith()
This is a (multiple allowed):

* [x] bug
* [ ] enhancement

* CakePHP Version: 3.5.17

I've found a problem using the ShadowTranslate behavior ([AD7six/cakephp-shadow-translate](url)).

Please, see also [AD7six/cakephp-shadow-translate/issues/52](url) for more information.

I have the associations: Destinations hasMany Countries hasMany CountriesI18n

This is an example that shows what the behavior is trying to do:

```php
$DestinationsTable
  ->find()
  ->innerJoinWith('Countries', function ($q) {
    return $q->contain('CountriesI18n');
   })
  ->where(['CountriesI18n.slug' => 'aaa']);
```
The result:
> PDOException with message 'SQLSTATE[42S22]: Column not found: 1054 Unknown column 'CountriesI18n.slug' in 'where clause''

This is the resulting query:

```
=> Cake\ORM\Query {#334                     
     +"(help)": "This is a Query object, to get the results execute or iterate it.",
     +"sql": "SELECT Destinations.id AS `Destinations__id`, Destinations.created AS
`Destinations__created`, Destinations.type AS `Destinations__type`,
Destinations.province_id AS `Destinations__province_id`,
Destinations.country_id AS `Destinations__country_id`,
Destinations.parent_id AS `Destinations__parent_id`, Destinations.enabled AS
`Destinations__enabled`, Destinations.name AS `Destinations__name`
FROM destinations Destinations
INNER JOIN countries Countries ON Countries.id = (Destinations.country_id)
WHERE CountriesI18n.slug = :c0",
     +"params": [                          
       ":c0" => [             
         "value" => "aaa",                  
         "type" => null,                   
         "placeholder" => "c0",
       ],                                   
     ],                                    
     +"defaultTypes": [
....
   ],
     +"decorators": 0,
     +"executed": false,
     +"hydrate": true,
     +"buffered": true,
     +"formatters": 0,
     +"mapReducers": 0,
     +"contain": [
       "Countries" => [
         "CountriesI18n" => [],
       ],
     ],
     +"matching": [
       "Countries" => [
         "queryBuilder" => Closure {#348
           parameters: {
             $q: {},
           },
         },
         "matching" => true,
         "joinType" => "INNER",
         "fields" => false,
       ],
     ],
     +"extraOptions": [],
     +"repository": App\Model\Table\DestinationsTable {#301
       +"registryAlias": "Destinations",
       +"table": "destinations",
       +"alias": "Destinations",
       +"entityClass": "App\Model\Entity\Destination",
       +"associations": [
         "countries",
       ],
       +"behaviors": [],
       +"defaultConnection": "default",
       +"connectionName": "default",
     },
   }
```

I'm trying to understand the issue and fix it but I'm not sure where and how to fix it.

Thanks.

othercorey:
Fixes https://github.com/cakephp/cakephp/issues/12440

Original test case
https://github.com/cakephp/cakephp/pull/12693

The original issue doesn't describe the failure properly because `hasMany` will load through select.

This tests against the `matching` option set by `matching()` or `innerJoinWith()` that gets passed to the association from the eager loader although it's not one of the options consumed by `attachTo()`.

### DIFF
--- a/src/ORM/Association.php
+++ b/src/ORM/Association.php
@@ -751,6 +751,16 @@ abstract class Association
             }
         }
 
+        if (
+            !empty($options['matching']) &&
+            $this->_strategy === static::STRATEGY_JOIN &&
+            $dummy->getContain()
+        ) {
+            throw new RuntimeException(
+                "`{$this->getName()}` association cannot contain() associations when using JOIN strategy."
+            );
+        }
+
         $dummy->where($options['conditions']);
         $this->_dispatchBeforeFind($dummy);
 
--- a/tests/TestCase/ORM/QueryTest.php
+++ b/tests/TestCase/ORM/QueryTest.php
@@ -32,6 +32,7 @@ use Cake\ORM\Query;
 use Cake\ORM\ResultSet;
 use Cake\TestSuite\TestCase;
 use ReflectionProperty;
+use RuntimeException;
 
 /**
  * Tests Query class
@@ -46,6 +47,7 @@ class QueryTest extends TestCase
     protected $fixtures = [
         'core.Articles',
         'core.ArticlesTags',
+        'core.ArticlesTranslations',
         'core.Authors',
         'core.Comments',
         'core.Datatypes',
@@ -3574,6 +3576,28 @@ class QueryTest extends TestCase
             ->matching('articles')
             ->toArray();
         $this->assertEquals($expected, $results);
+    }
+
+    /**
+     * Tests contain() in query returned by innerJoinWith throws exception.
+     *
+     * @return void
+     */
+    public function testInnerJoinWithContain()
+    {
+        $comments = $this->getTableLocator()->get('Comments');
+        $articles = $comments->belongsTo('Articles');
+        $articles->hasOne('ArticlesTranslations');
+
+        $this->expectException(RuntimeException::class);
+        $this->expectExceptionMessage('`Articles` association cannot contain() associations when using JOIN strategy');
+        $comments->find()
+            ->innerJoinWith('Articles', function (Query $q) {
+                return $q
+                    ->contain('ArticlesTranslations')
+                    ->where(['ArticlesTranslations.title' => 'Titel #1']);
+            })
+            ->sql();
     }
 
     /**


src/ORM/Association.php
```diff
@@ -749,6 +749,15 @@ public function attachTo(Query $query, array $options = []): void
                     $this->getName()
                 ));
             }
+            if (
+                !empty($options['matching']) &&
+                $this->_strategy === static::STRATEGY_JOIN &&
+                !empty($dummy->getContain())
```
dereuromark, line 755: there should be no method usage here in empty() calls, it can be removed without issues here.

tests/TestCase/ORM/QueryTest.php
```diff
@@ -3576,6 +3578,29 @@ public function testInnerJoinWithSelect()
...
+    {
+        $comments = $this->getTableLocator()->get('Comments');
+        $articles = $comments->belongsTo('Articles');
+        $articles->hasOne('ArticlesTranslations');
+
+        $this->expectException(RuntimeException::class);
+        $this->expectExceptionMessage('Cannot contain associations in query returned by query builder');
+        $comments
+            ->find()
+            ->innerJoinWith('Articles', function ($q) {
```
dereuromark, line 3596: Can you please typehint $q here, as the ->contain() and ->where() are not detected right now (not followable).

src/ORM/Association.php
othercorey: Can you rephrase that?

dereuromark: `&& $dummy->getContain()`
suffices

The cloaking method check is not needed and an anti pattern in these cases of existing property, key or method.

othercorey: That's fine, just couldn't understand the sentence so wanted to double check.

src/ORM/Association.php
```diff
@@ -749,6 +749,15 @@ public function attachTo(Query $query, array $options = []): void
                     $this->getName()
                 ));
             }
+            if (
+                !empty($options['matching']) &&
+                $this->_strategy === static::STRATEGY_JOIN &&
+                $dummy->getContain()
+            ) {
+                throw new RuntimeException(
+                    "Cannot contain associations in query returned by query builder for `{$this->getName()}`."
```
markstory, line 758:
```suggestion
                    "Cannot contain associations in queries returned by query builder Closures for `{$this->getName()}`."
```

tests/TestCase/ORM/QueryTest.php
```diff
@@ -3576,6 +3578,28 @@ public function testInnerJoinWithSelect()
...
+     * @return void
+     */
+    public function testInnerJoinWithContain()
+    {
+        $comments = $this->getTableLocator()->get('Comments');
+        $articles = $comments->belongsTo('Articles');
+        $articles->hasOne('ArticlesTranslations');
+
+        $this->expectException(RuntimeException::class);
+        $this->expectExceptionMessage('Cannot contain associations in query returned by query builder');
```
markstory, line 3593:
```suggestion
        $this->expectExceptionMessage('Cannot contain associations in queries returned by query builder');
```

src/ORM/Association.php
othercorey: Should we fix the other exception messages to use `Closure`?

`'Query builder for association "%s" did not return a query'`

othercorey: Never mind -- closure isn't really relevant. I adjusted the test to happen no matter how the contain is set.

PR #9058, opened 2016-06-30
adds case insensitive lookup for AssociaationCollection::type()

thinkingmedia:
Closes #9056

PR #14556, opened 2020-05-06
Remove lower casing of association names in AssociationCollection.

### LINKED ISSUE #14554: Case insensitive use of association names in AssociationCollection causes problems.
I have seen this issue come up many times in our support channels. 

### The problem
This diff below shows how the result gets changed if one mistakenly uses incorrect inflection when containing an association. If one uses `authors` instead of `Authors` in `contain()` the related record is now under property `Authors` instead of `author` and also it's an array instead of entity.

```php
diff --git a/tests/TestCase/ORM/Association/BelongsToTest.php b/tests/TestCase/ORM/Association/BelongsToTest.php
index 3fe85b2229..6cc56e870f 100644
--- a/tests/TestCase/ORM/Association/BelongsToTest.php
+++ b/tests/TestCase/ORM/Association/BelongsToTest.php
@@ -415,12 +415,12 @@ class BelongsToTest extends TestCase
         $query = $articles->find()
             ->select(['Authors.name'])
             ->where(['Articles.id' => 1])
-            ->contain('Authors');
+            ->contain('authors');
         $result = $query->firstOrFail();
 
-        $this->assertNotEmpty($result->author);
-        $this->assertSame('mariano', $result->author->name);
-        $this->assertSame(['author'], array_keys($result->toArray()), 'No other properties included.');
+        $this->assertNotEmpty($result->Authors);
+        $this->assertSame('mariano', $result->Authors['name']);
+        $this->assertSame(['Authors'], array_keys($result->toArray()), 'No other properties included.');
     }
 
     /**
```

### The cause
The source of the problem is that all methods of `AssociationCollection` lower case the association names.

### The solution
Remove the lower casing of assocation names/aliases. That would cause the ORM to throw exception with message like `"The authors association is not defined on Articles."` so users are made aware of the problem and they can fix it.

ADmad:
It causes undeterministic behavior in the ORM. For e.g. using different casing
in contain() calls than the one used when declaring association caused related
records to be under different entity property in the resultset entities. Also related
record remained as an array instead of being hydrated as an entity.

Fixes #14554.

### DIFF
--- a/src/ORM/Table.php
+++ b/src/ORM/Table.php
@@ -894,7 +894,13 @@ class Table implements RepositoryInterface, EventListenerInterface, EventDispatc
     {
         $association = $this->findAssociation($name);
         if (!$association) {
-            throw new InvalidArgumentException("The {$name} association is not defined on {$this->getAlias()}.");
+            $assocations = $this->associations()->keys();
+
+            $message = "The `{$name}` association is not defined on `{$this->getAlias()}`.";
+            if ($assocations) {
+                $message .= "\nValid associations are: " . implode(', ', $assocations);
+            }
+            throw new InvalidArgumentException($message);
         }
 
         return $association;
--- a/src/ORM/AssociationCollection.php
+++ b/src/ORM/AssociationCollection.php
@@ -71,7 +71,7 @@ class AssociationCollection implements IteratorAggregate
     {
         [, $alias] = pluginSplit($alias);
 
-        return $this->_items[strtolower($alias)] = $association;
+        return $this->_items[$alias] = $association;
     }
 
     /**
@@ -110,7 +110,6 @@ class AssociationCollection implements IteratorAggregate
      */
     public function get(string $alias): ?Association
     {
-        $alias = strtolower($alias);
         if (isset($this->_items[$alias])) {
             return $this->_items[$alias];
         }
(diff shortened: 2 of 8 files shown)


dereuromark:
I like it :) Why didnt we do that before?

Given that this wouldn't impact users that have been using the ORM correctly, this could fit into this minor release.

ADmad:
> I like it :) Why didnt we do that before?

I never knew `AssociationCollection` did this wackiness else I would have fixed it in 4.0 itself :)  It came to my attention recently after I saw multiple users on Slack having problems after using incorrect casing in `contain()` calls.

dereuromark:
Refs https://github.com/cakephp/cakephp/pull/9058

ADmad:
> Refs #9058

Nope, that issue is only related to making the association types case insensitive. The case insensitivity of association name must be from 2.x days.

src/ORM/Table.php
```diff
@@ -894,7 +894,13 @@ public function getAssociation(string $name): Association
     {
         $association = $this->findAssociation($name);
         if (!$association) {
-            throw new InvalidArgumentException("The {$name} association is not defined on {$this->getAlias()}.");
+            $assocations = $this->associations()->keys();
+
+            $message = "The `{$name}` association is not defined on `{$this->getAlias()}`.";
+            if ($assocations) {
+                $message .= "\nValid associations are: " . implode(', ', $assocations);
```
markstory, line 901: This is a nice touch 💅

markstory (commented):
Looks good to me. This is something we'll need to call out in the 4.1 upgrade guide.

PR #14559, opened 2020-05-07
Fix false being set in form inputs

### LINKED ISSUE #14548: Cannot store "false" values as a hidden field in a reliable way
This is a (multiple allowed):

* [x] bug
* [ ] enhancement

* CakePHP Version: 4.0
* Platform and Target: Debian Buster

### What you did
See #5639 
I have a column declared as "TINYINT(1) DEFAULT 0 NOT NULL". When I store the value "false" as a hidden field in a form it is read back and converted to NULL, which cannot be written back into the table.

### What happened
I get a database error indicating that NULL instead of false was written back

That the value is converted back to false and written back as such instead of NULL

"Before you open an issue, please check if a similar issue already exists or has been closed before."
Yes, I did :)

In my form I write
`echo $this->Form->control("<table>.<column>", "hidden")`
which produces the output
`<input id="..." type="hidden" name="...">`
Note, that there is no `value="..."` in the input field.
On submit, the POST parameter is empty, which translates to an empty string, which is marshaled as NULL in patchEntity.

In my opinion, we have 4 different values of bool: true, false, and "not a value" aka null, and "not specified". A hidden field should somehow be able to distinguish between them, though I have no idea how to do that with the last 2 (null and undefined).

markstory:
Convert false values into 0 so that they can be marshalled back into false for boolean fields.

Fixes #14548

### DIFF
--- a/src/View/Widget/BasicWidget.php
+++ b/src/View/Widget/BasicWidget.php
@@ -79,13 +79,13 @@ class BasicWidget implements WidgetInterface
 
         $data['value'] = $data['val'];
         unset($data['val']);
+        if ($data['value'] === false) {
+            $data['value'] = '0';
+        }
 
         $fieldName = $data['fieldName'] ?? null;
         if ($fieldName) {
-            if (
-                $data['type'] === 'number'
-                && !isset($data['step'])
-            ) {
+            if ($data['type'] === 'number' && !isset($data['step'])) {
                 $data = $this->setStep($data, $context, $fieldName);
             }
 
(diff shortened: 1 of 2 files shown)


src/View/Widget/BasicWidget.php
```diff
@@ -79,13 +79,13 @@ public function render(array $data, ContextInterface $context): string
 
         $data['value'] = $data['val'];
         unset($data['val']);
+        if ($data['value'] === false) {
+            $data['value'] = '0';
```
othercorey, line 83:
```suggestion
            // explicitly convert to 0 to avoid empty string which is marshaled as null
            $data['value'] = '0';
```

ADmad (approved):

ADmad:
Should this be backported to 3.8 or 3.9?

othercorey:
Users probably have a lot of accidentally nullable columns so this would probably help although it would definitely change behavior.

But, since we merged this into 4.0.x, we might as well merge it into 3.8.x.

PR #14561, opened 2020-05-07
Set ajax response type in initialize()

othercorey:
This allows the response type to be set on the default Response. Using initialize() also avoids overriding the constructor.

### DIFF
--- a/src/View/AjaxView.php
+++ b/src/View/AjaxView.php
@@ -16,10 +16,6 @@ declare(strict_types=1);
  */
 namespace Cake\View;
 
-use Cake\Event\EventManager;
-use Cake\Http\Response;
-use Cake\Http\ServerRequest;
-
 /**
  * A view class that is used for AJAX responses.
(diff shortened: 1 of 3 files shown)


src/View/AjaxView.php
```diff
@@ -33,23 +29,11 @@ class AjaxView extends View
...
-    ) {
-        if ($response) {
-            $response = $response->withType('ajax');
-        }
-
-        parent::__construct($request, $response, $eventManager, $viewOptions);
+    public function initialize(): void
+    {
+        parent::initialize();
+        $this->setResponse($this->getResponse()->withType('ajax'));
```
ADmad, line 37: There's a tiny chance this can cause problem for someone who has overridden `initialize()` and forgot to call `parent::initialize()` (as you did in previous commit :slightly_smiling_face: ). So it would be safer to target this to `4.next`.

othercorey: Sounds good.

PR #14562, opened 2020-05-07
Improve error message for EmailTrait::MailContains

mrothauer:
additional changes for #14518 , first PR was #14523

I added the necessary code changings to cover these missing functionalities

* Add new error message for MailContains and MailContainsHtml
* changings also work with usage of $this->at
* adapted unit tests

**For my project's tests those changings work well, but a unit test for the changings is still missing because I have no idea how to test the assert error message of a unit test.**


markstory:
Thank you 👍 We don't generally have a ton of tests for error messages from test assertions beyond the ones you changed. More tests would likely end up being tautological and fragile.

PR #14563, opened 2020-05-07
Add iterable type into generator doc.

dereuromark:
This makes it possible for IDE and stan to know the iterable type returned
I think it is actually wrong to set Generator as return type, should be iterable as base type here.
But that's probably for a minor or major to replace.

This just aims at better discoverability and to avoid having to inline annotate the with() usage everywhere.

I also saw other references as

    \Generator<\FQCN>

and

    iterable<\FQCN>

etc.
But those are not yet fully understood by IDEs usually.
See https://phpstan.org/r/51d234b0-e6fd-4408-a5b7-4151ea6a0646

### DIFF
--- a/src/Core/PluginCollection.php
+++ b/src/Core/PluginCollection.php
@@ -325,7 +325,7 @@ class PluginCollection implements Iterator, Countable
      * Filter the plugins to those with the named hook enabled.
      *
      * @param string $hook The hook to filter plugins by
-     * @return \Generator A generator containing matching plugins.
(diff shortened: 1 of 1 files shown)


dereuromark:
Looks like psalm is not happy about this.
But all other forms still require inline annotation then.

dereuromark:
I found a way that works, as per [this suggestion](https://github.com/vimeo/psalm/issues/3316#issuecomment-625443939)
If that is acceptable.

PR #14566, opened 2020-05-07
4.1 - Clean up cloning of Query and EagerLoader

othercorey:


markstory:
What does this fix? Does it help fix state being held between executions?

othercorey:
The only real issue this fixes is the double-cloning of $_eagerLoader in Query.

I thought there was a held state, but it looks like EagerLoader will clear before re-calculating arrays -- although that code is a little disconnected. I reverted the cloning of those instances/arrays.

PR #14570, opened 2020-05-09
4.next - Explicit list of supported valueSources

diegosardina:
No functional change, just removed hardcoded part. This will reduce code of my next PR.

Tests already cover for valid values sources

### DIFF
--- a/src/View/Helper/FormHelper.php
+++ b/src/View/Helper/FormHelper.php
@@ -219,7 +219,14 @@ class FormHelper extends Helper
     protected $_lastAction = '';
 
     /**
-     * The sources to be used when retrieving prefilled input values.
+     * The supported sources that can be used to populate input values.
+     *
+     * @var string[]
+     */
+    protected $_supportedValueSources = ['context', 'data', 'query'];
+
+    /**
+     * The default sources.
      *
      * @var string[]
      */
@@ -2481,9 +2488,20 @@ class FormHelper extends Helper
     }
 
     /**
+     * Extracts valid value sources.
+     *
+     * @param string|string[] $sources A string or a list of strings identifying a source.
+     * @return array
+     */
+    protected function _extractValidValueSources($sources): array
+    {
+        return array_values(array_intersect((array)$sources, $this->_supportedValueSources));
+    }
+
+    /**
      * Sets the value sources.
      *
-     * Valid values are `'context'`, `'data'` and `'query'`.
+     * Valid values are in $this->_supportedValueSources.
      * You need to supply one valid context or multiple, as a list of strings. Order sets priority.
      *
      * @param string|string[] $sources A string or a list of strings identifying a source.
@@ -2491,7 +2509,7 @@ class FormHelper extends Helper
      */
     public function setValueSources($sources)
     {
-        $this->_valueSources = array_values(array_intersect((array)$sources, ['context', 'data', 'query']));
+        $this->_valueSources = $this->_extractValidValueSources($sources);
 
         return $this;
     }
--- a/src/View/Helper/FormHelper.php
+++ b/src/View/Helper/FormHelper.php
@@ -223,7 +223,7 @@ class FormHelper extends Helper
      *
      * @var string[]
      */
-    protected $_supportedValueSources = ['context', 'data', 'query'];
+    protected $supportedValueSources = ['context', 'data', 'query'];
 
     /**
      * The default sources.
--- a/src/View/Helper/FormHelper.php
+++ b/src/View/Helper/FormHelper.php
@@ -2490,7 +2490,7 @@ class FormHelper extends Helper
     /**
      * Extracts valid value sources.
      *
-     * @param string[] $sources A a list of strings identifying a source.
+     * @param string[] $sources A list of strings identifying a source.
      * @return string[]
      */
     protected function _extractValidValueSources(array $sources): array
--- a/src/View/Helper/FormHelper.php
+++ b/src/View/Helper/FormHelper.php
@@ -2501,7 +2501,7 @@ class FormHelper extends Helper
     /**
      * Sets the value sources.
      *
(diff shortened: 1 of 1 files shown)


src/View/Helper/FormHelper.php
```diff
@@ -2480,18 +2487,29 @@ public function getValueSources(): array
         return $this->_valueSources;
     }
 
+    /**
+     * Extracts valid value sources.
+     *
+     * @param string|string[] $sources A string or a list of strings identifying a source.
+     * @return array
```
dereuromark, line 2494: string[] then?

diegosardina: It receives the same type of setValueSources()

dereuromark: I don't think the return result does, if you look into what the code does.

src/View/Helper/FormHelper.php
```diff
@@ -2480,18 +2487,29 @@ public function getValueSources(): array
...
+     * Valid values are in $this->_supportedValueSources.
      * You need to supply one valid context or multiple, as a list of strings. Order sets priority.
      *
      * @param string|string[] $sources A string or a list of strings identifying a source.
      * @return $this
      */
     public function setValueSources($sources)
     {
-        $this->_valueSources = array_values(array_intersect((array)$sources, ['context', 'data', 'query']));
+        $this->_valueSources = $this->_extractValidValueSources($sources);
```
garas, line 2512: Use `(array)` here, and require `array` as argument innew method.

src/View/Helper/FormHelper.php
```diff
@@ -2480,18 +2487,29 @@ public function getValueSources(): array
         return $this->_valueSources;
     }
 
+    /**
+     * Extracts valid value sources.
+     *
+     * @param string[] $sources A a list of strings identifying a source.
```
garas, line 2493:
```suggestion
     * @param string[] $sources A list of strings identifying a source.
```

src/View/Helper/FormHelper.php
```diff
@@ -2480,18 +2487,29 @@ public function getValueSources(): array
         return $this->_valueSources;
     }
 
+    /**
+     * Extracts valid value sources.
+     *
+     * @param string[] $sources A list of strings identifying a source.
+     * @return string[]
+     */
+    protected function _extractValidValueSources(array $sources): array
```
othercorey, line 2496:
```suggestion
    protected function extractValidValueSources(array $sources): array
```

src/View/Helper/FormHelper.php
```diff
@@ -2480,18 +2487,29 @@ public function getValueSources(): array
...
+    {
+        return array_values(array_intersect($sources, $this->_supportedValueSources));
+    }
+
     /**
      * Sets the value sources.
      *
-     * Valid values are `'context'`, `'data'` and `'query'`.
+     * Valid values are in $this->_supportedValueSources.
```
othercorey, line 2504:
```suggestion
     * Valid values are returned by `supportedValueSources()`.
```

src/View/Helper/FormHelper.php
```diff
@@ -2480,18 +2487,29 @@ public function getValueSources(): array
...
      * You need to supply one valid context or multiple, as a list of strings. Order sets priority.
      *
      * @param string|string[] $sources A string or a list of strings identifying a source.
      * @return $this
      */
     public function setValueSources($sources)
     {
-        $this->_valueSources = array_values(array_intersect((array)$sources, ['context', 'data', 'query']));
+        $this->_valueSources = $this->_extractValidValueSources((array)$sources);
```
othercorey, line 2512:
```suggestion
        $this->_valueSources = $this->extractValidValueSources((array)$sources);
```

src/View/Helper/FormHelper.php
```diff
@@ -219,7 +219,14 @@ class FormHelper extends Helper
     protected $_lastAction = '';
 
     /**
-     * The sources to be used when retrieving prefilled input values.
+     * The supported sources that can be used to populate input values.
+     *
+     * @var string[]
+     */
+    protected $_supportedValueSources = ['context', 'data', 'query'];
```
othercorey, line 226:
```suggestion
    protected $supportedValueSources = ['context', 'data', 'query'];
```

src/View/Helper/FormHelper.php
```diff
@@ -2480,18 +2487,29 @@ public function getValueSources(): array
...
 
+    /**
+     * Extracts valid value sources.
+     *
+     * @param string[] $sources A list of strings identifying a source.
+     * @return string[]
+     */
+    protected function _extractValidValueSources(array $sources): array
+    {
+        return array_values(array_intersect($sources, $this->_supportedValueSources));
```
othercorey, line 2498:
```suggestion
        return array_values(array_intersect($sources, $this->supportedValueSources));
```

markstory:
Seems fine, I don't understand why we need the new method yet though.

dereuromark:
Extendability probably, if you want to add in another source.

diegosardina:
> Seems fine, I don't understand why we need the new method yet though.

It is for my next PR. I found a case where it's convenient to override the valueSources just for a single field. The code is very small and doesn't add complexity to the FormHelper and imho it is worth to have.

However I didn't like to have this in _two places_:

```
... = array_values(array_intersect((array)$sources, ['context', 'data', 'query']));
```

Another reason to isolate that part is to raise an exception in the case the list will be empty (for example if I mistakenly wrote ['adta'] instead of ['data'], the field will never get a value). But I didn't because it's a behavioural change (it also requires tests adaptation) and I wanted this to be less invasive as possible in a minor release.

diegosardina:
> Extendability probably, if you want to add in another source.

Yup, maybe one day 'session' could be added, but I don't need it now :-)

Issue #14571, opened 2020-05-09
4.0: Shell to Command loses interactive disable flag.


dereuromark:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 4.0

### What happened
Shells upgraded to Commands.

Some Upgraded shells now do not work anymore as before in non-inactivate mode (via -q flag).

Shells used to have this
```
    public function in(string $prompt, $options = null, ?string $default = null): ?string
    {
        if (!$this->interactive) {
            return $default;
        }
```

The commands now always prompt even if not necessary and -q desires defaults to be auto-used.

### What you expected to happen
Commands should also have a non-interactive mode as the Shells did, that try to auto-set the fields as per defaults, to avoid prompting in e.g. CI setup or deploy etc.

dereuromark:
https://github.com/cakephp/cakephp/pull/14576

PR #14572, opened 2020-05-09
Interactive false for Commands.

dereuromark:
Addresses https://github.com/cakephp/cakephp/issues/14571

Restores the ability to  use non-interactive mode

The command then has to set

    $io->setInteractive(false);

and e.g.

    bin/cake i18n extract -q --overwrite --extract-core=yes --relative-paths

would run without issues now using the defaults.
This can then further be used programmatically.

Right now, even quit and overwrite enabled, it still forces input confirmation:
```
Current paths: None
What is the path you would like to extract?
[Q]uit [D]one
[src/] > 
```
The default here is `src/`.

Additionally:
It also auto detects this based on -q (quiet) flag, as you would never use this in normal interactive mode, but for crons, for CI checks, for deploy scripts etc, and here you cannot interfere really anyway.
We could move this part to 4.1 if we wanted, to put this into a minor release.

### DIFF
--- a/src/Console/ConsoleIo.php
+++ b/src/Console/ConsoleIo.php
@@ -129,8 +129,7 @@ class ConsoleIo
     }
 
     /**
-     * @param bool $value
-     *
+     * @param bool $value Value
      * @return void
      */
     public function setInteractive(bool $value): void
--- a/src/Console/ConsoleIo.php
+++ b/src/Console/ConsoleIo.php
@@ -32,6 +32,27 @@ use SplFileObject;
 class ConsoleIo
 {
     /**
+     * Output constant making verbose shells.
+     *
+     * @var int
+     */
+    public const VERBOSE = 2;
+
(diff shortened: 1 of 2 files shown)


src/Console/ConsoleIo.php
```diff
@@ -123,6 +128,15 @@ public function __construct(
         $this->_helpers->setIo($this);
     }
 
+    /**
+     * @param bool $value Value
+     * @return void
+     */
+    public function setInteractive(bool $value): void
```
markstory, line 135: We've been trying to avoid adding public methods in bugfix releases. Should this move to 4.next?

dereuromark: We can do that.

dereuromark:
Since 4.1 is pretty much there, we can move this regression fix into a cleaner fix+enhancement for 4.1 then.

PR #14573, opened 2020-05-09
Improve CSRF validation errors

markstory:
Improve the CSRF token validation errors with the aims of helping reduce confusion for application developers and end users.


src/Http/Middleware/CsrfProtectionMiddleware.php
```diff
@@ -272,7 +272,10 @@ protected function _validateToken(ServerRequestInterface $request): void
             return;
         }
 
-        throw new InvalidCsrfTokenException(__d('cake', 'Missing CSRF token body'));
+        throw new InvalidCsrfTokenException(__d(
+            'cake',
+            'Missing or invalid CSRF token. A CSRF token could not be matched in either the request body or request headers.'
```
othercorey, line 277:
```suggestion
            'CSRF token from either the request body or request headers did not match or is missing.'
```

PR #14574, opened 2020-05-09
Consolidate logging formatting for query logging

markstory:
Move context formatting out of LoggedQuery::__toString() and leverage the new log formatting that has been added to 4.1. This is a small behavior change but not an API breaking one in my opinion.

### DIFF
--- a/src/Database/Log/QueryLogger.php
+++ b/src/Database/Log/QueryLogger.php
@@ -35,6 +35,7 @@ class QueryLogger extends BaseLog
     public function __construct(array $config = [])
     {
         $this->_defaultConfig['scopes'] = ['queriesLog'];
+        $this->_defaultConfig['connection'] = '';
 
         parent::__construct($config);
     }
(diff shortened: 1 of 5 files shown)


src/Database/Log/QueryLogger.php
```diff
@@ -44,10 +44,13 @@ public function __construct(array $config = [])
...
-            'debug',
-            'connection=' . $this->getConfig('connection', '') . ' ' . (string)$context['query'],
-            ['scope' => $this->scopes() ?: ['queriesLog']]
-        );
+        $context['scope'] = $this->scopes() ?: ['queriesLog'];
+        $context['connection'] = $this->getConfig('connection');
+
+        if ($context['query'] instanceof LoggedQuery) {
+            $context = $context['query']->getContext() + $context;
+            $message = 'connection={connection} duration={took} rows={numRows} ' . $message;
```
othercorey, line 52: Does this print `connection={connection}` if the connection name context isn't set?

markstory: Yes.

dereuromark: Maybe it should be empty string.

markstory: :+1: Added a default value for connection.

PR #14575, opened 2020-05-10
Fix form-urlencoded PUT bodies not working in tests

### LINKED ISSUE #14565: PUT Request with x-www-form-urlencoded header has no data on IntegrationTest
This is a (multiple allowed):

* [x] bug

* CakePHP Version: 3.5.18
* Platform and Target: Only Integration Test class which extends IntegrationTestCase. this issue not happen on product.

### What you did
I tried to test PUT Request which change user data by Integration Test.
```php
class ChangeUserDataApiTest extends IntegrationTestCase
{
    public function testPut()
    {
        $this->configRequest([
            'headers' => [
                'Accept' => 'application/json',
                'Content-Type' => 'application/x-www-form-urlencoded',
            ],
        ]);

        // user cookie
        $this->cookie('hoge', 'hoge');
        $this->cookieEncrypted('fuga', 'fuga');

        $this->put('/change-user-data-api', ['name' => 'test']);

        $this->assertResponseCode(204);
    }
}
```

### What happened
this test failed and return 500 status code.

I injected following code in controller which I try to test.
```php
var_export($this->request->getData());
```

I expected that `array ('name' => 'test' )` is printed, but `array()`.

this not reproduced on product code. only Integration Test.

I want to test PUT request with `x-www-form-urlencoded` header.
I expected that `ServerRequest::getData()` return second argument of `$this->put` on Integration Test.

markstory:
When using application/www-form-urlencoded content type request bodies are expected to be present in the input stream. Propagate the input data through the request factory so that the stream is in the correct state during ServerRequest's constructor.

Fixes #14565

### DIFF
--- a/src/Http/ServerRequestFactory.php
+++ b/src/Http/ServerRequestFactory.php
@@ -49,6 +49,7 @@ abstract class ServerRequestFactory implements ServerRequestFactoryInterface
      * @param array $body $_POST superglobal
      * @param array $cookies $_COOKIE superglobal
      * @param array $files $_FILES superglobal
+     * @param string $input php://input Used to stub request streams in testing.
      * @return \Cake\Http\ServerRequest
      * @throws \InvalidArgumentException for invalid file values
      */
@@ -57,7 +58,8 @@ abstract class ServerRequestFactory implements ServerRequestFactoryInterface
         ?array $query = null,
         ?array $body = null,
         ?array $cookies = null,
-        ?array $files = null
+        ?array $files = null,
+        ?string $input = null
     ): ServerRequest {
         $server = normalizeServer($server ?: $_SERVER);
         $uri = static::createUri($server);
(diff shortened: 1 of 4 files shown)


src/Http/ServerRequestFactory.php
```diff
@@ -79,6 +81,7 @@ public static function fromGlobals(
             'base' => $uri->base,
             'session' => $session,
             'mergeFilesAsObjects' => Configure::read('App.uploadedFilesAsObjects', true),
+            'input' => $input,
```
ADmad, line 84: This create merge conflict with `4.next` since this is how passing input stream data is handled there
https://github.com/cakephp/cakephp/blob/b1436bf69b40cda9ec7bfcdae2a68406e7d7d998/src/Http/ServerRequestFactory.php#L82

src/Http/ServerRequestFactory.php
```diff
@@ -57,7 +58,8 @@ public static function fromGlobals(
         ?array $query = null,
         ?array $body = null,
         ?array $cookies = null,
-        ?array $files = null
+        ?array $files = null,
+        ?string $input = null
```
ADmad, line 62: Should we be changing constructor signature in patch releases?

PR #14577, opened 2020-05-10
Updated CommonTableExpression

othercorey:
This updates CommonTableExpression to try and simplify the API and fix a couple issues. The interface felt a little over-engineered.

The unit tests were updated to read a little easier and with docblock comments. I removed a couple Sql Server tests that only tested internal functionality and nothing to do with cake sql syntax generation.

- Removed the need for excessive setters and getters in CommonTableExpression

  Everything is just an implicit setter now. I tried to align it with other expressions.
  ````
  $cte
      ->name('test')
      ->query(function() { 
          return new Query(); 
      })
      ->field(['col1, col2]);
  ```

- Added support for closures when setting query in CommonTableExpression

  ```
  new CommonTableExpression('test', function() {
      return new Query();
  });
  ```

- ~Added `Query::withRecursive()` and removed `CommonTableExpression::setRecursive()`.~

- Cleaned up the way unit tests ran and validated results


src/Database/Expression/CommonTableExpression.php
```diff
@@ -98,190 +106,106 @@ public function setName(string $name)
...
-            }
-
-            $modifiers = ' ' . implode(' ', $modifiers);
-        }
+        $suffix = $this->materialized ? $this->materialized . ' ' : '';
 
-        return sprintf('%s%s AS%s (%s)', $this->name, $fields, $modifiers, $this->query->sql($generator));
+        return sprintf(
+            '%s%s AS %s(%s)',
+            $this->name,
```
dereuromark, line 189: A string cast maybe?

src/Database/Expression/CommonTableExpression.php
```diff
@@ -30,258 +29,164 @@ class CommonTableExpression implements ExpressionInterface
     /**
      * The CTE name.
      *
-     * @var string|null
+     * @var string
```
dereuromark, line 32: For this to be true you need to init it as empty string ''

othercorey: I'm not sure what the correct "trick" is here. If you never allow it to be set to null, phpstan and psalm are ok with saying it's required to be set.

I don't mind initializing it to `''` if that's more proper.

othercorey: I can just say it's nullable and add the checks. Any external use of it is hidden now so what's the difference right.

dereuromark: Yeah, this is more for internal correctness.
I would probably prefer nullable and string cast where necessary
Or a protected getter that throws exception if needed as string (orFail pendant).

othercorey: Agreed. I see now harm in generating empty values. Users will get a sql error instead of a crash which makes more sense.

Took me a while to simplify it down.

src/Database/Query.php
```diff
@@ -420,68 +420,44 @@ public function traverseParts(callable $visitor, array $parts)
...
     }
 
+    /**
+     * Adds a new recursive common table expression (CTE) to the query.
+     *
+     * @param \Closure|\Cake\Database\Expression\CommonTableExpression $cte The CTE to add.
+     * @param bool $overwrite Whether to reset the list of CTEs.
+     * @return $this
+     */
+    public function withRecursive($cte, $overwrite = false)
```
dereuromark, line 456: Do we want to have typehints here for bool $overwrite?

othercorey: I assumed there was some historical reason they aren't here and simply copied.

dereuromark: Bc for existing maybe. For new we can always

src/Database/Expression/CommonTableExpression.php
```diff
@@ -56,230 +48,143 @@ class CommonTableExpression implements ExpressionInterface
...
-                    'The `$fields` argument must contain only instances of `%s`, or strings, `%s` given at index `%d`.',
-                    ExpressionInterface::class,
-                    getTypeName($field),
-                    $index
-                ));
+        if ($query instanceof Closure) {
+            $query = $query();
+            if (!($query instanceof ExpressionInterface)) {
+                throw new RuntimeException(
+                    'You must return an `ExpressionInterface` from closure passed to `query()`.'
```
markstory, line 99:
```suggestion
                    'You must return an `ExpressionInterface` from a Closure passed to `query()`.'
```

src/Database/Query.php
```diff
@@ -408,80 +408,49 @@ public function traverseParts(callable $visitor, array $parts)
...
 
-        if ($expression === null) {
-            if (!$overwrite) {
-                throw new \InvalidArgumentException(
-                    'Resetting the WITH clause only works when overwriting is enabled.'
+        if ($cte instanceof Closure) {
+            $cte = $cte(new CommonTableExpression(), $this);
+            if (!($cte instanceof CommonTableExpression)) {
+                throw new RuntimeException(
+                    'You must return a `CommonTableExpression` from closure passed to `with()`.'
```
markstory, line 431:
```suggestion
                    'You must return a `CommonTableExpression` from a Closure passed to `with()`.'
```

src/Database/Query.php
```diff
@@ -408,80 +408,49 @@ public function traverseParts(callable $visitor, array $parts)
...
      * @return $this
-     * @throws \InvalidArgumentException When passing `null` for the `$expression` argument but not enabling
-     *  `$overwrite`.
-     * @throws \InvalidArgumentException When an invalid type is passed or returned for the `$expression` argument.
-     * @throws \InvalidArgumentException When the given CTE object has no name set.
-     * @throws \InvalidArgumentException When the given CTE object has no query set.
-     * @throws \InvalidArgumentException When a CTE object with the same name is already attached to this query.
      */
-    public function with($expression, $overwrite = false)
+    public function with($cte, bool $recursive = false, bool $overwrite = false)
```
markstory, line 421: Having recursive be a mode on the expression will be more readable. With this form a long with will have a trailing `true` and figuring out what that means will require IDE jump/hover. A named method doesn't have that drawback.

src/Database/Query.php
```diff
@@ -408,80 +408,49 @@ public function traverseParts(callable $visitor, array $parts)
...
+    /**
+     * Adds a new recursive common table expression (CTE) to the query.
+     *
+     * @param \Closure|\Cake\Database\Expression\CommonTableExpression $cte The CTE to add.
+     * @param bool $overwrite Whether to reset the list of CTEs.
+     * @return $this
+     */
+    public function withRecursive($cte, bool $overwrite = false)
+    {
+        return $this->with($cte, true, $overwrite);
```
markstory, line 451: This method could be skipped if the expression had a recursive mode on it.

src/Database/Query.php
```diff
@@ -408,80 +408,49 @@ public function traverseParts(callable $visitor, array $parts)
...
-            if ($existing->getName() === $name) {
-                throw new InvalidArgumentException(sprintf(
-                    'A common table expression with the name `%s` is already attached to this query.',
-                    $name
-                ));
-            }
         }
 
-        $this->_parts['with'][] = $expression;
+        $this->_parts['with'][] = ['cte' => $cte, 'recursive' => $recursive];
```
markstory, line 436: How does a visitor callback access the `recursive` aspect of a cte?

tests/TestCase/Database/CommonTableExpressionIntegrationTest.php
```diff
@@ -1,1027 +0,0 @@
...
-        $expected = [
-            [
-                'col' => '1',
-            ],
-        ];
-
-        $stmt = $query->execute();
-        $result = $stmt->fetchAll('assoc');
-        $stmt->closeCursor();
-        $this->assertEquals($expected, $result);
```
markstory, line 110: We should keep a bunch of these integration tests. They will be helpful for ensuring the code works in the future and give us a reference point for the generated SQL and results that are created.

othercorey: As far as I know, I didn't discard any actual integration tests. I renamed the file to CommonTableExpressionQueryTests to match WindowQueryTests.

The only integration tests I removed were 2 that tested how sql server lets you use the CTE name in some place ... but they aren't relevant to how we generate any sql.

src/Database/Query.php
othercorey: There isn't one to traverse. It's just a query with a union.

src/Database/Query.php
othercorey: It sounds like we prefer `$cte->recursive()`. I had that change and then switched to this to see how it feels. I'll switch back to that. I think that's reasonable to want to know which one is recursive at the expression level.

tests/TestCase/Database/CommonTableExpressionIntegrationTest.php
markstory: :+1: It was hard to see that change in the diff output.

PR #14578, opened 2020-05-10
Clean up tests using PHPStan level 1.

dereuromark:
At some point we might want to think about using at least level 1 here for `tests/`.

All the issues found here was using this level
Now down from 1600 to 1000 errors for that directory

### DIFF
--- a/tests/TestCase/Command/PluginAssetsCommandsTest.php
+++ b/tests/TestCase/Command/PluginAssetsCommandsTest.php
@@ -32,10 +32,13 @@ class PluginAssetsCommandsTest extends TestCase
 {
     use ConsoleIntegrationTestTrait;
 
+    /**
+     * @var string
+     */
     protected $wwwRoot;
 
     /**
-     * @var Cake\Filessytem\Filesystem;
+     * @var \Cake\Filesystem\Filesystem
      */
     protected $fs;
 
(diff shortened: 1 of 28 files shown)


dereuromark:
The phpstan fail is unrelated and due to phpstan release .24 , so both master and 4.next are red now.

tests/TestCase/Command/PluginAssetsCommandsTest.php
```diff
@@ -122,7 +121,7 @@ public function testSymlinkWhenVendorDirectoryExits()
         $this->exec('plugin assets symlink');
         $path = $this->wwwRoot . 'company' . DS . 'test_plugin_three';
         if (DS === '\\') {
-            $this->assertDirectoryExits($path);
+            $this->assertDirectoryExists($path);
```
othercorey, line 124: I am guessing this test never runs?

dereuromark: Jep, sure looks like it then. Only relevant for windows from what I see.

othercorey: Should we remove it or mark it as window-specific to force it to run? Feels bad to leave a test that doesn't run.

dereuromark: There are several more of these in this file.
Admad added this:

        $this->skipIf(
            DS === '\\',
            'Skip AssetsTask tests on windows to prevent side effects for UrlHelper tests on AppVeyor.'
        );

Thats why those are dead. Not sure what to do here.

dereuromark: We could probably look into the windows issue as separate PR.

PR #14580, opened 2020-05-10
4.1 - Cleaned up WindowExpression sql generation

othercorey:
This simplifies some sql generation to help clean up the window expression for 4.1.

- The expression is always generated by `sql()`.
- The parent expression/query builder does not need to extract the name to special case.
- The get/setName() is simplified to just an implicit setter.
- The meaning of "empty" is clarified in helper name.


markstory:
Nice work. :+1:

PR #14581, opened 2020-05-11
4.1 - Added closure support to WindowExpression partition() and order()

### LINKED ISSUE #14537: 4.1 - Add closure support to window expression partition and order clauses
We should allow closures for `WindowInterface::partition()` and `WindowInterface::order()` like other query builders.

othercorey:
Closes https://github.com/cakephp/cakephp/issues/14537

This allows passing closures to `partition()` and `order()` as they return expressions.

I made `order()` require an `OrderByExpression` because having a closure return individual fields felt strange.

### DIFF
--- a/src/Database/Expression/WindowExpression.php
+++ b/src/Database/Expression/WindowExpression.php
@@ -20,6 +20,7 @@ use Cake\Database\ExpressionInterface;
 use Cake\Database\ValueBinder;
 use Closure;
 use InvalidArgumentException;
+use RuntimeException;
 
 /**
  * This represents a SQL window expression used by aggregate and window functions.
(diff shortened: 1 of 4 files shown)


src/Database/Expression/WindowExpression.php
```diff
@@ -119,6 +129,18 @@ public function order($fields)
...
         }
 
+        if ($fields instanceof Closure) {
+            $fields = $fields();
+            if (!($fields instanceof OrderByExpression)) {
+                throw new RuntimeException(
+                    'You must return an `OrderByExpression` from a Closure passed to `order()`.'
+                );
+            }
+            $this->order = $fields;
```
othercorey, line 139: This should act the same as when we add closure support to Query::order(). If we want it to only return fields there this needs to change.

PR #14583, opened 2020-05-11
Upate psalm's annotations/config.

ADmad:

### DIFF
--- a/src/Core/ObjectRegistry.php
+++ b/src/Core/ObjectRegistry.php
@@ -331,8 +331,9 @@ abstract class ObjectRegistry implements Countable, IteratorAggregate
      *
      * @param string $objectName The name of the object to set in the registry.
      * @param object $object instance to store in the registry
-     * @psalm-param TObject $object
      * @return $this
+     * @psalm-param TObject $object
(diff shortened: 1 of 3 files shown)


src/Core/ObjectRegistry.php
```diff
@@ -331,8 +331,9 @@ public function reset()
      *
      * @param string $objectName The name of the object to set in the registry.
      * @param object $object instance to store in the registry
-     * @psalm-param TObject $object
      * @return $this
+     * @psalm-param TObject $object
+     * @psalm-suppress MoreSpecificReturnType
```
othercorey, line 336: Do we really need both this and the one above the return statement?

dereuromark:
If this gets master green again, looks fair enough for now.

src/Core/ObjectRegistry.php
ADmad: We do if we want to keep pslam happy :)  My understanding of generics annotation is still WIP. So I am not sure whether our annotation are incorrect or it's a bug in pslam, so suppressing the error is the easiest option right now.

PR #14584, opened 2020-05-11
Clean up tests using PHPStan level 2/3.

dereuromark:
Follow https://github.com/cakephp/cakephp/pull/14578

using phpstan level 2/3 a few more issues pop up

as a side effect also some of the src/ issues appear, e.g. wrong annotations.

---

level 1: 800 phpstan issues left for tests/
level 2: ~1000
level 3: ~3000

### DIFF
--- a/tests/TestCase/Controller/Component/RequestHandlerComponentTest.php
+++ b/tests/TestCase/Controller/Component/RequestHandlerComponentTest.php
@@ -797,7 +797,7 @@ class RequestHandlerComponentTest extends TestCase
 
         $event = new Event('Controller.beforeRender', $this->Controller);
         $requestHandler = new RequestHandlerComponent($this->Controller->components());
-        $this->assertNull($requestHandler->beforeRender($event));
+        $requestHandler->beforeRender($event);
         $this->assertTrue($event->isStopped());
         $this->assertEquals(304, $this->Controller->getResponse()->getStatusCode());
         $this->assertSame('', (string)$this->Controller->getResponse()->getBody());
@@ -822,7 +822,7 @@ class RequestHandlerComponentTest extends TestCase
         $event = new Event('Controller.beforeRender', $this->Controller);
 
         $requestHandler = new RequestHandlerComponent($this->Controller->components());
-        $this->assertNull($requestHandler->beforeRender($event));
+        $requestHandler->beforeRender($event);
         $this->assertTrue($event->isStopped());
         $this->assertEquals(304, $this->Controller->getResponse()->getStatusCode());
         $this->assertSame('', (string)$this->Controller->getResponse()->getBody());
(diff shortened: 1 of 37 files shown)


dereuromark:
Does this not say anything about the file locations?
```
##[error]The declared return type 'Cake\Core\ObjectRegistry&static' for Cake\Core\ObjectRegistry::set is more specific than the inferred return type 'Cake\Core\ObjectRegistry<TObject:Cake\Core\ObjectRegistry as mixed>&Cake\Event\EventDispatcherInterface'
##[error]The type 'Cake\Core\ObjectRegistry<TObject:Cake\Core\ObjectRegistry as mixed>&Cake\Event\EventDispatcherInterface' is more general than the declared return type 'Cake\Core\ObjectRegistry&static' for Cake\Core\ObjectRegistry::set
##[error]The declared return type 'Cake\Core\ObjectRegistry&static' for Cake\Core\ObjectRegistry::unload is more specific than the inferred return type 'Cake\Core\ObjectRegistry<TObject:Cake\Core\ObjectRegistry as mixed>&Cake\Event\EventDispatcherInterface'
##[error]The type 'Cake\Core\ObjectRegistry<TObject:Cake\Core\ObjectRegistry as mixed>&Cake\Event\EventDispatcherInterface' is more general than the declared return type 'Cake\Core\ObjectRegistry&static' for Cake\Core\ObjectRegistry::unload
##[error]Argument 1 of array_map expects callable|null, array{Cake\ORM\Association, string(aliasField)} provided
##[error]Process completed with exit code 1.
```

ADmad:
Those psalm errors are already fixed on master.

tests/TestCase/Controller/Component/RequestHandlerComponentTest.php
```diff
@@ -797,7 +797,7 @@ public function testCheckNotModifiedByEtagStar(): void
 
         $event = new Event('Controller.beforeRender', $this->Controller);
         $requestHandler = new RequestHandlerComponent($this->Controller->components());
-        $this->assertNull($requestHandler->beforeRender($event));
```
ADmad, line 800: Removing assertions might not be a good idea.

dereuromark: They are not removed :)

ADmad: > They are nonsense from the start for void.

But `beforeRender()` callbacks can return a response instance.

ADmad: I see we have added return typehint `void` for `RequestHander::beforeRender()`. That was probably not the best idea.

dereuromark: ;) yeah, if we fix that, we can also undo this part here.

PR #14585, opened 2020-05-11
Revert constructor signature change introduced in #14575.

ADmad:
It's best not to change constructor signature in a bugfix release.
Also this change will avoid merge conflict with 4.next.


src/TestSuite/MiddlewareDispatcher.php
```diff
@@ -148,6 +148,7 @@ protected function _createRequest(array $spec): ServerRequest
     {
         if (isset($spec['input'])) {
             $spec['post'] = [];
+            $spec['environment']['CAKEPHP_INPUT'] = $spec['input'];
```
markstory, line 151: Didn't know about this being in 4.1. This is much better.

PR #14587, opened 2020-05-13
4.next - Allow fields to override valueSources

diegosardina:
In my previous PR, I removed Request from Context classes to allow for ['context', 'data'], so FormHelper is now able to show changes that comes directly from the entity instead of restoring request data.

However, this is not enough if we _add_ or _remove_ fields from entity. In that case only ['context'] works, otherwise FormHelper restores values from request data in wrong places causing a lot of mess.

But with ['context'] we can't have fields in form that are not part of entity. For this reason my proposal is to allow for fields override of valueSources:

```
<?= $this->Form->create($entity, ['valueSources' => ['context']]) ?>

[...]

<?= $this->Form->control('non_entity_field', ['valueSources' => ['data']]) ?>
```

### DIFF
--- a/src/View/Helper/FormHelper.php
+++ b/src/View/Helper/FormHelper.php
@@ -2287,7 +2287,9 @@ class FormHelper extends Helper
             $valOptions = [
                 'default' => $options['default'] ?? null,
                 'schemaDefault' => $options['schemaDefault'] ?? true,
+                'valueSources' => $options['valueSources'] ?? null,
             ];
+            unset($options['valueSources']);
             $options['val'] = $this->getSourceValue($field, $valOptions);
         }
         if (!isset($options['val']) && isset($options['default'])) {
@@ -2527,7 +2529,15 @@ class FormHelper extends Helper
             'data' => 'getData',
             'query' => 'getQuery',
         ];
-        foreach ($this->getValueSources() as $valuesSource) {
+
+        $valueSources = $this->getValueSources();
+
+        if (isset($options['valueSources'])) {
+            $valueSources = $this->extractValidValueSources((array)$options['valueSources']);
+            unset($options['valueSources']);
+        }
+
+        foreach ($valueSources as $valuesSource) {
             if ($valuesSource === 'context') {
                 $val = $this->_getContext()->val($fieldname, $options);
                 if ($val !== null) {
--- a/tests/TestCase/View/Helper/FormHelperTest.php
+++ b/tests/TestCase/View/Helper/FormHelperTest.php
@@ -8312,7 +8312,7 @@ class FormHelperTest extends TestCase
     }
 
     /**
-     * Test the basic setters and getters for value sources
+     * Test the basic setters and getters and field override for value sources
      *
      * @return void
      */
@@ -8344,15 +8344,31 @@ class FormHelperTest extends TestCase
         $result = $this->Form->getSourceValue('id');
         $this->assertEquals($expected, $result);
 
+        $expected = '1';
+        $result = $this->Form->getSourceValue('id', ['valueSources' => ['data']]);
+        $this->assertEquals($expected, $result);
+
         $this->Form->setValueSources(['data']);
         $expected = '1';
         $result = $this->Form->getSourceValue('id');
+        $this->assertEquals($expected, $result);
+
+        $expected = '2';
+        $result = $this->Form->getSourceValue('id', ['valueSources' => ['query']]);
         $this->assertEquals($expected, $result);
 
         $this->Form->setValueSources(['query', 'data']);
         $expected = '2';
         $result = $this->Form->getSourceValue('id');
         $this->assertEquals($expected, $result);
+
+        $expected = '1';
+        $result = $this->Form->getSourceValue('id', ['valueSources' => ['data', 'query']]);
+        $this->assertEquals($expected, $result);
+
+        $result = $this->Form->getSourceValue('id', ['valueSources' => ['context']]);
+        $this->assertNull($result);
+
     }
 
     /**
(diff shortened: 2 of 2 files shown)


markstory:
What kind of use cases do these requirements come up for you? I've not ever run into this kind of situation so I'd like to know more about how it came up. This feels like a very specific scenario that adds complexity with marginal benefits to few.

diegosardina:
> What kind of use cases do these requirements come up for you? I've not ever run into this kind of situation so I'd like to know more about how it came up. This feels like a very specific scenario that adds complexity with marginal benefits to few.

If one is using a ['context'] only form, sooner or later surely needs to override valueSources for fields that are not part of entity and live in request data.

The use case for a ['context'] only form is a strong one, adding and removing associated entities, or fields in general, dynamically:
* **without JavaScript**,
* keeping **FormProtection active** (to protected associated ids),
* dynamic forms that are **phpunit testable**.

Adding and removing fields won't work well both with ['data', 'context'] and ['context', 'data'].

For the record: Symfony and Zend have the so called _form collections_ that allow easily to add and remove associated entities dynamically (both with or without JavaScript) and since I started with CakePHP, I wasn't able to find a non-JS working solution that doesn't require these changes to the FormHelper.
In the mentioned frameworks the entity plays the big role as in the case of ['context'], if you want a field that only lives in request data, you have to set that is not mapped to the entity (that is equivalent to se valueSource to ['data']).

The accepted solution among users on the net was [this](https://waltherlalk.com/blog/dynamic-form-input-fields) (the site isn't avaiable anymore). It uses jQuery, disabled the old SecurityComponent and it's not testable without headless browser solutions (or other tools I don't know, I'm not a JavaScript lover).

The code below is sub-optimal, it has been oversimplified for the sake of shortness. However the complete code is not a complex use case.

```
//src/Model/Table/MentorsTable.php
    public function beforeMarshal(EventInterface $event, ArrayObject $data, ArrayObject $options)
    {
        if ($data->offsetExists('student_add')) {
            $data['students'][] = [];
            $options['validate'] = false;
            $options['associated']['students']['validate'] = false;
        }
        if ($data->offsetExists('student_delete')) {
            $index = $data['student_delete'];
            unset($data['students'][$index]);
            $options['validate'] = false;
            $options['associated']['students']['validate'] = false;
        }
    }
```
```
// templates/Mentors/edit.php
            <?= $this->Form->create($mentor, ['valueSources' => ['context'], 'novalidate' => true]) ?>
            <fieldset>
                <legend><?= __('Edit Mentor') ?></legend>
                <?= $this->Form->control('name') ?>
                <fieldset>
                    <legend>Students</legend>
                    <?php foreach((array) $mentor->students as $key => $student): ?>
                        <?= $this->Form->controls(["students.$key.id", "students.$key.name"], ['legend' => false]) ?>
                        <?= $this->Form->button(__('Remove student'), ['name' => 'student_delete', 'value' => "$key"]) ?>
                    <?php endforeach; ?>
                </fieldset>
            </fieldset>
            <?= $this->Form->button(__('Add student'), ['name' => 'student_add']) ?>
            <?= $this->Form->button(__('Edit mentor'), ['name' => 'submit']) ?>
            <?= $this->Form->end() ?>
```

```
// src/Controller/MentorsController.php
    public function edit($id = null)
    {
        $mentor = $this->Mentors->get($id, ['contain' => ['Students']]);
        if ($this->request->is(['patch', 'post', 'put'])) {
            $mentor = $this->Mentors->patchEntity($mentor, $this->request->getData());
            if ($this->request->getData('submit') !== null) {
                if ($this->Mentors->save($mentor)) {
                    TableRegistry::getTableLocator()->get('Students')->deleteAll([
                        'mentor_id' => $mentor->id,
                        'id NOT IN' => (new Collection($mentor->students))
                            ->extract('id')->filter()->toList(),
                    ]);
                    $this->Flash->success(__('The mentor has been saved.'));
                }
            }
        }
        $this->set(compact('mentor'));
    }
```

image

ADmad:
I don't see how SecurityComponent/FormProtectionComponent are related to values sources. Those components are meant to prevent form field manipulation and if you are dynamically adding/removing from fields using javascript you obviously have to disable the component or unlock relevant fields.

Do you get new student entities created for the empty array you add using `$data['students'][] = [];` in `beforeMarshal()`?  If not that's where your problem lies. Trying using something like `$data['students'][] = ['id' => null]`;

diegosardina:
> I don't see how SecurityComponent/FormProtectionComponent are related to values sources. Those components are meant to prevent form field manipulation and if you are dynamically adding/removing from fields using javascript you obviously have to disable the component or unlock relevant fields.
> 

I didn't say that FormProtection caused problems or is related to value sources.

The point is I don't want to use JavaScript, just the old multiple submission, state-update and show, and there is a huge advantage in having ['context'] only form: There is no need to disable FormProtection :-) 


> Do you get new student entities created for the empty array you add using `$data['students'][] = [];` in `beforeMarshal()`? If not that's where your problem lies. Trying using something like `$data['students'][] = ['id' => null]`;

No, that code doesn't have any problem.

I just showed a _use-case_ where we can have dynamic forms with FormProtector that are phpunit testable.

The problem is the ['context'] only form, this is a limit in the case we have fields that live only in request data.

ADmad:
> No, that code doesn't have any problem.

What problem do you have with your example code without this patch? Please provide some debug info highlighting the problem.

diegosardina:
> > No, that code doesn't have any problem.
> 
> What problem do you have with your example code without this patch? Please provide some debug info highlighting the problem.


```
<?= $this->Form->create($entity, ['valueSources' => ['context']]) ?>

[...]

<?= $this->Form->control('non_entity_field') ?>
```

the value will get lost after the first form submission that does just patchEntity() to update the state of the entity.

image


Solution, being able to override valueSources for this field:
```
<?= $this->Form->control('non_entity_field', ['valueSources' => ['data']]) ?>
```

"Just use ['data', 'context'] or ['context', 'data']" is not the answer. FormHelper will restore deleted values by shifting or keep delete values. In this case only ['context'] makes sense.


image

ADmad:
> FormHelper will restore deleted values by shifting or keep delete values.

That's because you are only unsetting them from request data but the related entities are still still actually present without the parent entity that is patched.

Instead of unsetting records you want to delete, you should for e.g. set a "deleted" flag from them and use that to hide/skip them in form and then use same flag to filter out the records to be deleted upon final submission.

diegosardina:
> Instead of unsetting records you want to delete, you should for e.g. set a "deleted" flag from them and use that to hide/skip them in form and then use same flag to filter out the records to be deleted upon final submission.

What if you add a new one between two existing entities and not at the end?

ADmad:
Another way to handle your use case is to use the `Cake\Form\Form` class. Populate the Form instance's data from entity initially. Then on each add/delete submission replace the Form instance's data with request data. Eventually create back entity from the request data and save. If you set your hasMany association config to use replace strategy instead of merge you won't even have to keep track of deleted student records.

ADmad:
> What if you add a new one between two existing entities and not at the end?

That's when you stop trying to patch the entity at each step, use custom Form class and use entity just for initial population and final saving after all data array manipulations are completed.

diegosardina:
These aren't clean/straightforward ways :-) by the way now we can have ['context'] and it works very well for these use cases.
My proposal was to overcome the limitation in this case of not being able to have fields that live in request data, just that.

> This feels like a very specific scenario that adds complexity with marginal benefits to few.

I'm honest to say that I may be the only one in 2020 that uses dynamic forms without JavaScript :-) 

However in the patch I proposed I don't see any complexity there, just **6 lines** of effective code in FormHelper

markstory:
The complexity comes from having to keep the this behavior working, and maintaining documentation explaining how and when to use it.

ADmad:
Also specifying values source per input just seems conceptually wrong and hackish.

diegosardina:
> Also specifying value sources per input just seems conceptually wrong and hackish.

My opinion is different.

Having _more sources_ (with the others as fallback) seems conceptually wrong, since everwhere the entity/model plays the big role in keeping the state of the form and having a second source as fallback it's clearly a [POLA](https://en.wikipedia.org/wiki/Principle_of_least_astonishment) in the case you add or remove a field from the form.

When I add or remove a field or an associated from entity I expect that the form adapts well (as in Symfony, Zend, .NET and even Django). CakePHP was the first that caused a total mess with the request (it is clear FormHelper wasn't initially designed to deal with dynamic forms).

Even more wrong when the request was hardcoded in Context classes as the first choice where to get values, without that change the above dynamic form wouldn't work in any way.

Note: this is not a bash against CakePHP, otherwise I wouldn't even bother to improve it.

ADmad:
> Having _more sources_ (with the others as fallback) seems conceptually wrong, since everwhere the entity/model plays the big role in keeping the state of the form and having a second source as fallback it's clearly a [POLA](https://en.wikipedia.org/wiki/Principle_of_least_astonishment) in the case you add or remove a field from the form.

But you do have the option to control the value sources per form now. It's manipulating the source per input where we are drawing the line.

> When I add or remove a field or an associated from entity I expect that the form adapts well (as in Symfony, Zend, .NET and even Django). CakePHP was the first that caused a total mess with the request (it is clear FormHelper wasn't initially designed to deal with dynamic forms).

You are having problems because you are adding/removing associated records in request data, patching original entity with that and wanting FormHelper to figure out all the manipulation. That's expecting a bit too much.

I am not too familiar with the other frameworks you have listed but AFAIK they all require explicitly specifying the form fields info in one way or the other isn't? CakePHP is quite unique in it's ability to directly generate forms from entities. For use cases like yours (which is quite an edge case since you are trying to avoid using javascript) CakePHP does provide `Form` class to build your custom forms.

> Even more wrong when the request was hardcoded in Context classes as the first choice where to get values, without that change the above dynamic form wouldn't work in any way.

I don't remember exactly why that was done. There must be some historical reason for it and that behavior have been retained over the years since no one before you complained about it :slightly_smiling_face: 

> Note: this is not a bash against CakePHP, otherwise I wouldn't even bother to improve it.

We really appreciate your interest and efforts in helps us improve the framework. There might be just some things were we have to agree to disagree :slightly_smiling_face:.

othercorey:
@diegosardina If this PR isn't merged, is there a separate change that is needed to help create your custom `Form` subclass to implement this? Maybe we can help improve the ability to customize before 4.1 is released.

diegosardina:
It seems that there is already a way to override valueSources for a single field:

```
<?= $this->Form->create($entity, ['valueSources' => ['context']]) ?>
[...]
<?php $this->Form->setValueSources(['data']) ?>
<?= $this->Form->control('non_entity_field') ?>
<?php $this->Form->setValueSources(['context']) ?>
[...]
```
A bit overkilling for a single field but good if they are more than one. It is a more general solution.

Well, there is a solution and I'm satisfied with it, this Pull Request can be closed. I don't think someone is interested to it since I was the only one in favour to it :-)

However nobody pointed me it out this possibility, so it may not pretty straightforward.
Would be a nice idea to mention it in the doc?

PR #14589, opened 2020-05-14
4.1 formhelper value sources

ADmad:
- Throw exception for invalid source values instead of silently ignoring them.
- Improved docblocks

// @diegosardina

### DIFF
--- a/src/View/Helper/FormHelper.php
+++ b/src/View/Helper/FormHelper.php
@@ -2522,6 +2522,7 @@ class FormHelper extends Helper
      * @see FormHelper::$supportedValueSources for valid values.
      * @param string|string[] $sources A string or a list of strings identifying a source.
      * @return $this
+     * @throws \InvalidArgumentException If sources list contains invalid value.
      */
     public function setValueSources($sources)
     {
--- a/src/View/Helper/FormHelper.php
+++ b/src/View/Helper/FormHelper.php
@@ -2516,7 +2516,7 @@ class FormHelper extends Helper
     /**
      * Sets the value sources.
      *
-     * You need to supply one or more valid source, as a list of strings.
+     * You need to supply one or more valid sources, as a list of strings.
      * Order sets priority.
      *
      * @see FormHelper::$supportedValueSources for valid values.
--- a/src/View/Helper/FormHelper.php
+++ b/src/View/Helper/FormHelper.php
@@ -28,6 +28,7 @@ use Cake\View\Helper;
 use Cake\View\StringTemplateTrait;
 use Cake\View\View;
 use Cake\View\Widget\WidgetLocator;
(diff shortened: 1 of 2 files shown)


src/View/Helper/FormHelper.php
```diff
@@ -2488,28 +2494,41 @@ public function getValueSources(): array
...
      *
-     * Valid values are set in `$supportedValueSources`.
-     * You need to supply one valid context or multiple, as a list of strings. Order sets priority.
+     * You need to supply one or more valid source, as a list of strings.
+     * Order sets priority.
      *
+     * @see FormHelper::$supportedValueSources for valid values.
      * @param string|string[] $sources A string or a list of strings identifying a source.
      * @return $this
      */
```
othercorey, line 2525:
```suggestion
     * @throws \InvalidArgumentException If sources list contains invalid value.
     */
```

Isn't this the public api for users?

src/View/Helper/FormHelper.php
```diff
@@ -2488,28 +2494,42 @@ public function getValueSources(): array
...
-     * Extracts valid value sources.
+     * Validate value sources.
      *
      * @param string[] $sources A list of strings identifying a source.
-     * @return string[]
+     * @return void
+     * @throws \InvalidArgumentException If sources list contains invalid value.
      */
-    protected function extractValidValueSources(array $sources): array
+    protected function validateValueSources(array $sources): void
```
saeideng, line 2503: removing old `extractValidValueSources` methods is not BC !! is ?

diegosardina: No, it was added by me some days ago because sourceValues array was hardcoded in FormHelper

dereuromark: If it is a protected method, then in a minor this wouldn't be too big of a deal. At least if it is very unlikely to be overwritten - which it seems in this case.

ADmad: > removing old extractValidValueSources methods is not BC !! is ?

No it's not. This method is not present in any release. It was added just few days ago to 4.next branch.

Even if it was present in 4.0, our BC policy allows changing protected methods in new minors.

src/View/Helper/FormHelper.php
```diff
@@ -2488,28 +2494,42 @@ public function getValueSources(): array
...
+            ));
+        }
     }
 
     /**
      * Sets the value sources.
      *
-     * Valid values are set in `$supportedValueSources`.
-     * You need to supply one valid context or multiple, as a list of strings. Order sets priority.
+     * You need to supply one or more valid source, as a list of strings.
```
markstory, line 2519:
```suggestion
     * You need to supply one or more valid sources, as a list of strings.
```

Issue #14588, opened 2020-05-14
CsrfProtectionMiddleware / Cookie / ResponseEmitter does not allow to set SameSite security


smarek:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.8.12
* Platform and Target: nginx, php-7.4.5-fpm, linux

See scan report for cakephp.org https://securityheaders.com/?q=cakephp.org&followRedirects=on

Configuring session, it's possible to configure session cookie to be pretty much secure

```php
<?php
// snip from app_local.php
   return [ 'Session' => [
        'cookie' => '__Host-CAKEPHP',
        'defaults' => 'cake',
        'ini' => [
            'session.cookie_samesite' => 'Strict',
            'session.cookie_path' => '/',
            'session.cookie_httponly' => 1,
            'session.cookie_lifetime' => 60 * 60 * 24 * 7
        ],
        'timeout' => 60 * 60 * 24 * 7 // week
    ]];
```

However, as can be seen in scan report, it's not possible to configure samesite security for csrfToken cookie, because `Cake\Http\Cookie\Cookie` does not allow setting additional cookie params, and ResponseEmitter uses the [`setcookie`](https://www.php.net/manual/en/function.setcookie.php) call variant, which does not allow the samesite 'Strict'/'Lax' setting

Used setcookie function is available since PHP 5.2.0
```php
function setcookie ($name, $value = "", $expire = 0, $path = "", $domain = "", $secure = false, $httponly = false)
```
Desired function is available since PHP 7.3.0
```php
function setcookie($name, $value = '', array $options = []) :bool
```

Or one could avoid using PHP `setcookie()` function, and use `header()` directly, see https://www.php.net/manual/en/function.setcookie.php#124644

Tasks, as I see them
- [ ] Backport 4.x ResponseEmitter `protected function setCookie($cookie): bool` which already supports the newer call and workaround for samesite, to **3.x branch**
- [ ] Backport 4.x Cookie class to 3.x branch, because it supports samesite setting ($defaults)
- [ ] Update CsrfProtectionMiddleware in both 3.x and 4.x to suport SameSite option

Also, there is some new security measurement, called "Cookie Prefixing", see https://scotthelme.co.uk/tough-cookies/#cookieprefixes
[...]

PR #14590, opened 2020-05-14
Allow specifying "SameSite" attribute for CSRF protection cookie.

ADmad:
Refs #14588

### DIFF
--- a/src/Http/Middleware/CsrfProtectionMiddleware.php
+++ b/src/Http/Middleware/CsrfProtectionMiddleware.php
@@ -54,6 +54,9 @@ class CsrfProtectionMiddleware implements MiddlewareInterface
      *    Defaults to browser session.
      *  - `secure` Whether or not the cookie will be set with the Secure flag. Defaults to false.
      *  - `httpOnly` Whether or not the cookie will be set with the HttpOnly flag. Defaults to false.
+     *  - 'samesite' "SameSite" attribute for cookies. Defaults to `null`.
+     *    Valid values: `CookieInterface::SAMESITE_LAX`, `CookieInterface::SAMESITE_STRICT`,
+     *    `CookieInterface::SAMESITE_NONE` or `null`.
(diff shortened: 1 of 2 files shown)


src/Http/Middleware/CsrfProtectionMiddleware.php
```diff
@@ -64,6 +67,7 @@ class CsrfProtectionMiddleware implements MiddlewareInterface
         'expiry' => 0,
         'secure' => false,
         'httpOnly' => false,
+        'samesite' => null,
```
dereuromark, line 70: I wonder if the config should be `sameSite`, camelBacked same as `httpOnly`.

ADmad: It's `httpOnly` that should have been lowercased to match the casing using in `Cookie` methods.

dereuromark: The internal config string doesnt necessarily have to match the actual cookie value.
If httpOnly is like this and to avoid inconsistency, maybe we could keep camelBacked.

ADmad: I am not in favor of adding more inconsistency just to match a previous one. We can instead rename `httpOnly` to `httponly` and the constructor could check and copy old key to new one to maintain BC.

ADmad: The alternative signature of PHP's `setcookie()` method using `$options` array also uses all lowercased keys hence I had chosen the same for all options arrays in `Cookie` class.

markstory: Deprecating `httpOnly` and using `httponly` to increase consistency with the Cookie class seems like the right move here.

ADmad: Done

PR #14593, opened 2020-05-16
current evaluates to false on empty array

swiffer:
current evaluates to false on empty array

### DIFF
--- a/src/Database/FunctionsBuilder.php
+++ b/src/Database/FunctionsBuilder.php
@@ -75,7 +75,7 @@ class FunctionsBuilder
      */
     public function max($expression, $types = []): AggregateExpression
     {
-        return $this->aggregate('MAX', $this->toLiteralParam($expression), $types, current($types) ?? 'float');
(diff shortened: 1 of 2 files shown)


ADmad:
Change looks good but a test case would be nice.

swiffer:
done ;)

dereuromark:
As a bugfix shouldnt this go into master?

ADmad:
> As a bugfix shouldnt this go into master?

`master` already uses `?:`. Guess it was changed in 4.next only during some refactoring.

https://github.com/cakephp/cakephp/blob/e8d2d973339f130010b8d35b50ad04bfb6c1a3ff/src/Database/FunctionsBuilder.php#L121

swiffer:
This started to happen when testing 4.1 Preview (interested in cte support)

othercorey:
It must be part of the AggregateExpression change, but I can't tell you why I changed it. Probably re-typed it thinking it was the same.

PR #14595, opened 2020-05-17
There's a notice half way through the tests

Phally:
See: https://travis-ci.com/github/cakephp/cakephp/jobs/335780299#L472

Not sure if there's a reason this isn't suppressed and the test right underneath is:
https://github.com/cakephp/cakephp/blob/master/tests/TestCase/Error/ConsoleErrorHandlerTest.php#L81

Interestingly, if you run just that single test case the output doesn't appear.


Phally:
Oh never mind ha!

Phally:
No one saw that first commit right? Anyway, I fixed it properly now

markstory:
Thank you 👍

PR #14601, opened 2020-05-19
Fixed invalid tags

othercorey:


dereuromark (approved):

dereuromark:
By the way, if you want clickable links in the IDE the following syntax works:

    @deprecated Use {@link \FQCN\SomeInterface::someMethod()} instead.

PR #14605, opened 2020-05-20
Updated parameter names for inherited methods to match.

othercorey:
`DateTimeType` was also missing `implements BatchCastingInterface`.


src/Database/Type/DateTimeType.php
```diff
@@ -34,7 +34,7 @@
  *
  * Use to convert datetime instances to strings & back.
  */
-class DateTimeType extends BaseType
+class DateTimeType extends BaseType implements BatchCastingInterface
```
markstory, line 37: Nice find :+1:

PR #14606, opened 2020-05-20
Added sameSite-Attribute on CsrfProtectionMiddleware

tobse:
Since CakePHP 4.0 the Cookie-object supports the samesite-attribute. In the CsrfProtectionMiddleware this feature was still missing.


ADmad:
Ability to set "Samesite" attribute has been added to 4.1 #14590.

PR #14607, opened 2020-05-20
Fix CS: Use single quotes where applicable.

dereuromark:
For https://github.com/cakephp/cakephp-codesniffer/pull/307

also fixed some other small issues.

### DIFF
--- a/src/Controller/ControllerFactory.php
+++ b/src/Controller/ControllerFactory.php
@@ -108,8 +108,8 @@ class ControllerFactory implements ControllerFactoryInterface
             if ($firstChar !== strtoupper($firstChar)) {
                 deprecationWarning(
                     "The `{$prefix}` prefix did not start with an upper case character. " .
-                    "Routing prefixes should be defined as CamelCase values. " .
-                    "Prefix inflection will be removed in 5.0"
+                    'Routing prefixes should be defined as CamelCase values. ' .
+                    'Prefix inflection will be removed in 5.0'
                 );
 
                 if (strpos($prefix, '/') === false) {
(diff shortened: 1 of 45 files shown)


ADmad (approved):
Nice

templates/element/exception_stack_trace.php
```diff
@@ -13,6 +13,7 @@
  * @link          https://cakephp.org CakePHP(tm) Project
  * @since         1.3.0
  * @license       https://opensource.org/licenses/mit-license.php MIT License
+ * @var array $trace
```
othercorey, line 16: This doesn't look right.

src/Controller/ControllerFactory.php
```diff
@@ -108,8 +108,8 @@ public function getControllerClass(ServerRequest $request): ?string
             if ($firstChar !== strtoupper($firstChar)) {
                 deprecationWarning(
                     "The `{$prefix}` prefix did not start with an upper case character. " .
-                    "Routing prefixes should be defined as CamelCase values. " .
-                    "Prefix inflection will be removed in 5.0"
+                    'Routing prefixes should be defined as CamelCase values. ' .
```
othercorey, line 111: Not sure how I feel about having to perfectly switch when editing/writing multi-line strings.

templates/element/exception_stack_trace.php
dereuromark: We dont use inline allignment

src/Controller/ControllerFactory.php
ADmad: For such cases sprintf() could be used to avoid using both type of quotes.

othercorey: Sounds reasonable to me.

PR #14609, opened 2020-05-21
4.1 - Grouped query tests under QueryTests/

othercorey:
Moved the feature query tests into a QueryTests/ folder where we would break up the other feature tests to avoid clutter.


othercorey:
@markstory what did you fix?

markstory:
@othercorey I rebased as there was a conflict on the WindowExpression tests.

Issue #14611, opened 2020-05-21
Remove seconds from time/date FormHelper inputs


ryanolton:
This is a (multiple allowed):

* [ ] bug
* [x] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 4.1.0-beta1 (all of 4.x)
* Platform and Target: Debian 10.4, nginx 1.14.2, PostgreSQL 11.7, PHP 7.3.14

### What you did
I used FormHelper to create a control and set the type to "time".

`$this->Form->control( 'bedtime', [ 'type' => 'time' ] )`

### What happened
The resultant input was of type "time" and had seconds (and a "step" attribute).

### What you expected to happen
I knew what to expect based on the documentation, I just need the ability to remove seconds.

othercorey:
Does setting the step not work for you? Can you show the html output that is generated and what you expect?

ryanolton:
> Does setting the step not work for you? Can you show the html output that is generated and what you expect?

Setting the step does work to remove the step attribute, however, removing the step alone does not remove the seconds (at least not in Firefox), you must also remove the seconds from the input value:

**No step and no value (and no seconds):**

```
<html>
  <body>
    <form>
      <input type="time">
    </form>
  </body>
</html>
```

**No step but with a value (and with seconds):**
```
<html>
  <body>
    <form>
      <input type="time" value="11:00:00">
    </form>
  </body>
</html>
```

Even if a time without seconds is supplied to FormHelper, "11:00" for example, there is a function within the DateTimeWidget that converts it into predefined formats:

```
protected $formatMap = [
  'datetime-local' => 'Y-m-d\TH:i:s',
  'date' => 'Y-m-d',
  'time' => 'H:i:s',
  'month' => 'Y-m',
  'week' => 'Y-\WW',
];

...

$data['value'] = $this->formatDateTime($data['val'], $data);
```

Per the [Mozilla docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/time#Time_value_format), if seconds are supplied in the value, then seconds show in the input. Due to this reason, the FormHelper format defaults need overwritten so that seconds are omitted from the value used in the final HTML output.

It's the combination of having no step **and** having no seconds in the input's value that allows the format to omit seconds.

See PR #14612

othercorey:
Fixed in https://github.com/cakephp/cakephp/pull/14614.

PR #14612, opened 2020-05-21
Remove second from time input

ryanolton:
See [ Issue #14611 ](https://github.com/cakephp/cakephp/issues/14611)

The HTML5 "datetime-local" and "time" input types allow for the omission of seconds, but the FormHelper provides no way to omit seconds.

I modified the DateTimeWidget to allow for the omission of seconds by changing the default time/date format as well as removing the "step" attribute.

I also added the ability to supply a custom time/date format, if the defaults are inadequate.

### DIFF
--- a/src/View/Widget/DateTimeWidget.php
+++ b/src/View/Widget/DateTimeWidget.php
@@ -91,6 +91,9 @@ class DateTimeWidget extends BasicWidget
      * - `escape` Set to false to disable escaping on all attributes.
      * - `type` A valid HTML date/time input type. Defaults to "datetime-local".
      * - `timezone` The timezone the input value should be converted to.
+     * - `step` The stepping interval to use for this input (seconds). Default is based upon given type.
+     * - `format` The format of the time/date input. Default is based upon given type.
+     * - `second` Set to false to disable seconds and remove default step when using "datetime-local" or "time" type.
      *
      * All other keys will be converted into HTML attributes.
      *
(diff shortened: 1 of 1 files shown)


src/View/Widget/DateTimeWidget.php
```diff
@@ -91,6 +91,9 @@ class DateTimeWidget extends BasicWidget
      * - `escape` Set to false to disable escaping on all attributes.
      * - `type` A valid HTML date/time input type. Defaults to "datetime-local".
      * - `timezone` The timezone the input value should be converted to.
+     * - `step` The stepping interval to use for this input (seconds). Default is based upon given type.
+     * - `format` The format of the time/date input. Default is based upon given type.
+     * - `second` Set to false to disable seconds and remove default step when using "datetime-local" or "time" type.
```
ADmad, line 96: I don't see the need for a `second` option. Hiding seconds value can be achieved using setting required `format` only.

ryanolton: We can certainly achieve the same result simply by setting `step` to `null` and `format` to something like `'H:i'`, and I'm not against this simplified solution, but there was some precedence in [version 3.x](https://book.cakephp.org/3/en/views/helpers/form.html#options-for-time-related-controls) that I thought might be beneficial for those migrating to 4.x.

ADmad: We wanted to make the code simpler which is why we got rid of the multiple selects for datetime and related options of 3.x :slightly_smiling_face:.

I have opened an alternate PR #14614.

ryanolton: I understand -- PR looks good to me, thanks!

dereuromark:
If you are happy with the alternative PR you can also close this one.

PR #14614, opened 2020-05-21
4.1 datetime widget

### LINKED ISSUE #14611: Remove seconds from time/date FormHelper inputs

ADmad:
Alternative to #14612

- You can now set `step` attribute to `null` or `false` to prevent it form being added to HTML.
- You can use `format` option to explicitly specify the format to generate value strings for the input.

### DIFF
--- a/src/View/Widget/DateTimeWidget.php
+++ b/src/View/Widget/DateTimeWidget.php
@@ -94,9 +94,9 @@ class DateTimeWidget extends BasicWidget
      * - `step` The "step" attribute. Defaults to `1` for "time" and "datetime-local" type inputs.
      *   You can set it to `null` or `false` to prevent explicit step attribute being added in HTML.
      * - `format` A `date()` function compatible datetime format string.
-     *   By default the widget will use suitable format based on input type and
-     *   database type for the context. If explicit format is provided then no
-     *   default value will be set for `step` attribute and it too has to be
+     *   By default the widget will use a suitable format based on the input type and
+     *   database type for the context. If an explicit format is provided, then no
+     *   default value will be set for the `step` attribute, and it needs to be
      *   explicitly set if required.
      *
      * All other keys will be converted into HTML attributes.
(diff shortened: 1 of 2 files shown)


othercorey (approved):

src/View/Widget/DateTimeWidget.php
```diff
@@ -91,6 +91,13 @@ class DateTimeWidget extends BasicWidget
      * - `escape` Set to false to disable escaping on all attributes.
      * - `type` A valid HTML date/time input type. Defaults to "datetime-local".
      * - `timezone` The timezone the input value should be converted to.
+     * - `step` The "step" attribute. Defaults to `1` for "time" and "datetime-local" type inputs.
+     *   You can set it to `null` or `false` to prevent explicit step attribute being added in HTML.
+     * - `format` A `date()` function compatible datetime format string.
+     *   By default the widget will use suitable format based on input type and
```
markstory, line 97:
```suggestion
     *   By default the widget will use a suitable format based on the input type and
```

src/View/Widget/DateTimeWidget.php
```diff
@@ -91,6 +91,13 @@ class DateTimeWidget extends BasicWidget
...
+     *   database type for the context. If explicit format is provided then no
```
markstory, line 98:
```suggestion
     *   database type for the context. If an explicit format is provided, then no
```

src/View/Widget/DateTimeWidget.php
```diff
@@ -91,6 +91,13 @@ class DateTimeWidget extends BasicWidget
...
+     *   default value will be set for `step` attribute and it too has to be
```
markstory, line 99:
```suggestion
     *   default value will be set for the `step` attribute, and it needs to be
```

PR #14615, opened 2020-05-21
Ignore invalid CS error.

ADmad:
Prep for codesniffer v4.2.

### DIFF
--- a/src/Core/ObjectRegistry.php
+++ b/src/Core/ObjectRegistry.php
@@ -103,6 +103,7 @@ abstract class ObjectRegistry implements Countable, IteratorAggregate
             }
         }
 
+        // phpcs:ignore SlevomatCodingStandard.Namespaces.FullyQualifiedClassNameInAnnotation.NonFullyQualifiedClassName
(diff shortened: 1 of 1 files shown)


src/Core/ObjectRegistry.php
```diff
@@ -103,6 +103,7 @@ public function load(string $objectName, array $config = [])
             }
         }
 
+        // phpcs:ignore SlevomatCodingStandard.Namespaces.FullyQualifiedClassNameInAnnotation.NonFullyQualifiedClassName
```
othercorey, line 106: I wish slevomat would stop trying to validate psalm annotations.

othercorey (approved):

src/Core/ObjectRegistry.php
dereuromark: We cannot write `\TObject` here?

othercorey: It might do a blind string comparison, but I would prefer it not look like a qualified type since it's meant to be a `T` generics type.

PR #14617, opened 2020-05-21
Allow chainable fixture adding in TestCase

dereuromark:
RFC
What do you think about allowing chainable adding of fixtures in TestCases using `->addFixture()` instead of the array property directly?
This would also allow autocompleting on it then in IDEs.

           /**
             * @return void
             */
            public function initialize(): void
            {
                $this->addFixture('core.Articles')
                    ->addFixture('app.SomeModel')
                    ->addFixture('plugin.Some/ComplicatedName.SomePluginModel');
            }

`initialize()` is called from the constructor as it has to be quite early for us allow setting them as such, otherwise we get a "table not found" error it seems.

We cannot use `setUp()` as it is too late.

### DIFF
--- a/src/TestSuite/TestCase.php
+++ b/src/TestSuite/TestCase.php
@@ -909,7 +909,7 @@ abstract class TestCase extends BaseTestCase
      * @param string $fixture Fixture
      * @return $this
      */
-    public function addFixture(string $fixture)
+    protected function addFixture(string $fixture)
     {
         $this->fixtures[] = $fixture;
 
--- a/src/TestSuite/TestCase.php
+++ b/src/TestSuite/TestCase.php
@@ -81,25 +81,6 @@ abstract class TestCase extends BaseTestCase
     protected $_configure = [];
 
     /**
-     * @param string $name Name
-     * @param array $data Data
-     * @param string $dataName Data name
-     */
-    public function __construct($name = null, array $data = [], $dataName = '')
-    {
-        $this->initialize();
-
-        parent::__construct($name, $data, $dataName);
-    }
-
-    /**
-     * @return void
-     */
-    public function initialize(): void
-    {
-    }
-
-    /**
      * Overrides SimpleTestCase::skipIf to provide a boolean return value
      *
      * @param bool $shouldSkip Whether or not the test should be skipped.
--- a/src/TestSuite/TestCase.php
+++ b/src/TestSuite/TestCase.php
@@ -906,6 +906,9 @@ abstract class TestCase extends BaseTestCase
      * - app.MyRecords
      * - plugin.MyPluginName.MyModelName
      *
+     * Use this method inside your test cases' {@link getFixtures()} method
+     * to build up the fixture list.
+     *
      * @param string $fixture Fixture
      * @return $this
      */
--- a/src/TestSuite/TestCase.php
+++ b/src/TestSuite/TestCase.php
@@ -81,6 +81,25 @@ abstract class TestCase extends BaseTestCase
     protected $_configure = [];
 
     /**
+     * @param string $name Name
+     * @param array $data Data
+     * @param string $dataName Data name
+     */
+    public function __construct($name = null, array $data = [], $dataName = '')
+    {
+        $this->initialize();
+
+        parent::__construct($name, $data, $dataName);
+    }
+
+    /**
+     * @return void
+     */
+    public function initialize(): void
+    {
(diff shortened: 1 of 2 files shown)


src/TestSuite/TestCase.php
```diff
@@ -80,6 +80,25 @@ abstract class TestCase extends BaseTestCase
...
     protected $_configure = [];
 
+    /**
+     * @param string $name Name
+     * @param array $data Data
+     * @param string $dataName Data name
+     */
+    public function __construct($name = null, array $data = [], $dataName = '')
+    {
+        $this->initialize();
```
markstory, line 90: Could we call this method from `getFixtures()` instead?  That method was added in 4.0 to enable defining fixtures outside of the property.

dereuromark: My main concern with the getFixtures() method was the side effect of having to call add..() inside.
But I guess that's a fair compromise for now.

dereuromark:
I changed it so you can now extend `getFixtures()` directly:

       /**
         * @return string[]
         */
        public function getFixtures(): array
        {
            $this->addFixture('core.Articles')
                ->addFixture('app.SomeModel')
                ->addFixture('plugin.Some/ComplicatedName.SomePluginModel');

            return parent::getFixtures();
        }

etc

src/TestSuite/TestCase.php
```diff
@@ -898,6 +898,24 @@ public static function setAppNamespace(string $appNamespace = 'TestApp'): ?strin
...
+     *
+     * Examples:
+     * - core.Tags
+     * - app.MyRecords
+     * - plugin.MyPluginName.MyModelName
+     *
+     * @param string $fixture Fixture
+     * @return $this
+     */
+    public function addFixture(string $fixture)
```
dereuromark, line 912: Should this be protected? Any reason to have this public?

ADmad: Can't think of any reason for it to be public.

dereuromark: Fixed.

dereuromark:
We can either backport this to 3.next - or use shimming ( https://github.com/dereuromark/cakephp-shim/pull/76 ) as opt-in.

src/TestSuite/TestCase.php
```diff
@@ -898,6 +898,24 @@ public static function setAppNamespace(string $appNamespace = 'TestApp'): ?strin
         return $previous;
     }
 
+    /**
+     * Adds a fixture to this test case.
+     *
+     * Examples:
+     * - core.Tags
+     * - app.MyRecords
+     * - plugin.MyPluginName.MyModelName
```
markstory, line 907: Should we give some guidance on how this method is to be used? For example 'Use this method inside your test cases' `getFixtures()` method to build up the fixture list'

dereuromark: Sounds good. Here on in docs?

ADmad:
If one wants IDE autocompletion can't they just use FQCNs?

dereuromark:
Nah, thats the opposite of autocomplete, as you just get thousands of classes instead of the magic strings.
Plus a ton of use statement mess on top of it.

src/TestSuite/TestCase.php
dereuromark: I added the note in the docblock.

chinpei215:
I agree with @ADmad . If people don't want thousands of classes or a ton of use statements,  they can do like the following:

```php
use Cake\Test\Fixture as f;
```
```php
public $fixtures = [
    f\ArticlesFixture::class,
    f\CommentsFixture::class,
];
```

dereuromark:
That is the worst idea. It obfuscates everything.
Also, see the complex use cases as outlined in https://github.com/dereuromark/cakephp-ide-helper/pull/203 (nested ones)
To find them manually is way more difficult than using the documented and de facto standard string usage.

markstory:
>  The one method doesn't cost anything, and is fully opt-in to be used (or not).

The one potential cost is people trying to use it inside a setup() or test method and it not working. In addition to not working it doesn't fail either adding a silent 'failure'.

dereuromark:
Thus the documention added as per your suggestion.

markstory:
I agree that using imports would be simpler for us, but having that many imports in a file could get unwieldy. This seems like a low-maintenance method for us to provide and help those who use IDE autocompletion.

Issue #6108, opened 2015-03-19
patchEntity with null values not possible


markusramsak:
Is there any chance to set a field to "null" using patchEntity? I overwrote patchEntity to allow this...

markstory:
How does null get sent in POST data?

mdlouhy:
I have a routine that generates a PDF.  In the database I have a field "pdf_generated_date" which needs to be set to null before starting the generation process.
I wanna do.
```
$order = $this->Orders->get($orderId, ['fields' => ['id']]);
$order = $this->Orders->patchEntity($order, [ 'pdf_generated_date' => null], ['validate' => false]);
```

before newEntity() method was changed I could even do
```
$order = $this->Orders->newEntity([ 'id' => $orderId, 'pdf_generated_date' => null], ['validate' => false]);
```

but now I have to do
```
$order = $this->Orders->get($orderId, ['fields' => ['id']]);
$order->pdf_generated_date = null;
```

Database operations do not always depend on data sent by forms.
Think of having routines (for example triggered by cronjobs) which just does some updates, inserts, deleted based on data NOT sent by forms.

markstory:
You don't need patchEntity to set a single field though. You can just update the field. If you need to set multiple fields, Entity::set() can do that too.

mdlouhy:
I understand. So basically patchEntity is designed to work only with form data?! I didn't know that. In the docs there is no mention of that. Thanks. I think this solves my problem.

lorenzo:
@mdlouhy Would you add that to the docs?

mdlouhy:
I am not so good in English - I don't think that this is what you really want :P

markstory:
@mdlouhy The headings for those methods are under [Converting request data into entities](http://book.cakephp.org/3.0/en/orm/saving-data.html#converting-request-data-into-entities) what would make that more clear.

dereuromark:
Wasn't there recently a change regarding null and patching?

markstory:
Yes, but it was for a regression that was introduced. #9389

mehov:
As stupid as this sounds, if someone googles this like I just did, make sure your database column isn't set to `NOT NULL`

PR #14618, opened 2020-05-22
Auto lock closed issues after period of inactivity

ADmad:
This will prevent people commenting on old closed issues instead of opening new ones.

### DIFF
--- a/.github/workflows/lock-closed-issues.yml
+++ b/.github/workflows/lock-closed-issues.yml
@@ -1,0 +1,30 @@
+name: 'Lock inactive closed issues'
+
+on:
+  schedule:
+    - cron: '0 * * * *'
+
+jobs:
+  lock:
+    name: "Lock"
+
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: "Lock"
+        uses: dessant/lock-threads@v2
+        with:
+          github-token: ${{ github.token }}
+          issue-lock-inactive-days: '31'
+          issue-exclude-created-before: '2020-05-01T00:00:00Z'
+          issue-exclude-labels: ''
+          issue-lock-labels: ''
+          issue-lock-comment: ''
+          issue-lock-reason: 'resolved'
+          pr-lock-inactive-days: '31'
+          pr-exclude-created-before: '2020-05-01T00:00:00Z'
+          pr-exclude-labels: ''
+          pr-lock-labels: ''
+          pr-lock-comment: ''
+          pr-lock-reason: 'resolved'
+          process-only: ''


.github/workflows/lock-closed-issues.yml
```diff
@@ -0,0 +1,27 @@
...
+    steps:
+      - name: "Lock"
+        uses: dessant/lock-threads@v2
+        with:
+          github-token: ${{ github.token }}
+          issue-lock-inactive-days: '31'
+          issue-exclude-created-before: ''
+          issue-exclude-labels: ''
+          issue-lock-labels: ''
+          issue-lock-comment: >
```
othercorey, line 22: Will I get emails for every old issue?

dereuromark: Overall this could be more harmful than useful.

ADmad: @othercorey You can turn off notification emails and just use web notifications :)

But I agree even that would be annoying. I'll check if it's possible to make it not generate notifications but I doubt that's possible. Will just close the PR in that case.

garas: Does locking without comment generate notification?

othercorey: If it can lock without comment notification that seems ok.

gmponos: maybe if you could set a starting day that would be cool...

gmponos: Ah!.. isn't this the option `issue-exclude-created-before`? So maybe old issues could stay as they are and have a starting day.. that would ease the pain of notifications...

garas: There is `issue-exclude-created-before` to exclude very old issues

ADmad: Event without the comment GH will send notification to at least the author about issue/PR closure. That's not good either.

othercorey: If it's just the author, that might not be so bad.

gmponos: Yes.. the problem I believe is with old issues.. if you set a starting date i believe it will be fine...

dereuromark: Still, we are not solving any issues, only create them.
The 1 per year issue would be on the other side of constant noise of this closing.
I don't think that is worth it. Here it suffices to copy and paste a quick message to reopen new.

In fact, in some cases it is even warranted to just add to the existing.
So by forcing this we might even more shoot ourselves int the foot than necessary.

othercorey: I say it was a nice idea, but maybe we try again when there are new options.

markstory: Have we confirmed that locking will send notifications?

ADmad: * Locking without comment doesn't seem to generate notifications hence I have removed the comment.
* Updated config to lock both issues and PRs.
* Issues/PRs prior to 1st May 2020 won't be processed.

We can give it a try and if it does create annoyance we can always get rid of it.

othercorey: mark and I tested. Did not receive notifications just from locking.

dereuromark: I am still against this. If a PR got closed and could be opened, this would be blocked unnecessarily
and same for tickets - it creates more fuss and noise where nothing needed to be done IMO.

I see that in the phpstan repo right now, getting tons of noise for nothing.

dereuromark:
Cases like https://github.com/cakephp/cakephp/issues/6108 actually show how it can be useful to add data to existing tickets etc
And it doesn't hurt too much here either.
I stand by it: We don't need this IMO.

ADmad:
For me that comment is actually an example of why locking would be better. It's a useless comment.

Anyway I don't mind if this isn't merged.

Issue #14455, opened 2020-04-15
RFC ORM: Add support to string aggregation


celsowm:
Hi folks !
Well, since Cake 4.1 will adopt [window functions](https://github.com/cakephp/docs/pull/6527), I think the timing would be very good to add a very useful feature adopted by all major SGBDs in the last years:

- Mysql: http://sqlfiddle.com/#!9/146217/3
- Postgres:  http://sqlfiddle.com/#!17/fb46f/1/0
- SQLite: http://sqlfiddle.com/#!7/fb46f/1
- SQLServer 2017 >=: http://sqlfiddle.com/#!18/2cb0e/5 

As you can see, there are two different syntaxes: **STRING_AGG** and **GROUP_CONCAT**, this [article ](https://database.guide/mysql-group_concat-vs-t-sql-string_agg/)has a very good comparison.

An usage on Cake ORM would be something like:

```php
$q = $this->Classes->find();
$q
    ->select(['description', $q->func()->strAgg('name')])
    ->innerJoinWith('Students')
    ->groupBy('description');
```

This could be very useful to produce a fast and simple query for reports and similars

* [x] enhancement
* [x] feature-discussion (RFC)

dereuromark:
@celsowm Are you interested to make a PR against 4.next branch?

celsowm:
Sorry my ignorance, but what is PR?

dereuromark:
Pull Request (some call it merge request). A change as seen in the tab above.

celsowm:
Ah okay, hehe. I can try if you guys give me some tipos on which classes I need to add new methods

jimbo2150:
Working on this, should have a PR in the next day.

dereuromark:
@jimbo2150 Are you interested in finishing the PR https://github.com/cakephp/cakephp/pull/14620 ?
Lets continue the discussion there.

PR #14620, opened 2020-05-22
Added GROUP_CONCAT/STRING_AGG support

jimbo2150:
Adds support for the `GROUP_CONCAT()` (known as `STRING_AGG()` in PostgreSQL and MS SQL/tSQL) aggregate functions.

This is a PR for issue #14455


jimbo2150:
Have an issue where SQLite is returning different results - appears to due to fixtures duplicating data when run more than once (and it does). All the other databases seem to assume DISTINCT on the grouped associations except SQLite. Looking at it deeper I would agree that SQLite is likely in the right whereas the others are assuming deduplication. Since only MySQL supports the DISTINCT modifier within the function expression, would be better to do a subquery across all (with DISTINCT) but am having an issue getting the subquery to be generated rather than a join.

src/Database/Expression/AggregateExpression.php
```diff
@@ -179,9 +184,9 @@ public function sql(ValueBinder $generator): string
         $sql = parent::sql($generator);
         if ($this->window !== null) {
             if ($this->window->isNamedOnly()) {
-                $sql .= ' OVER ' . $this->window->sql($generator);
+                $sql .= " {$this->clause} " . $this->window->sql($generator);
```
othercorey, line 187: What is the point of this change?

jimbo2150: MS SQL Server uses a syntax similar to SQLite for the `ORDER BY` clause for result sorting within certain aggregate functions but instead of using the `OVER` clause, it uses `WITHIN GROUP`. I modified `AggregateExpression` to make the `OVER` clause dynamic so that it could be changed for TSQL.

SQL Server uses `STRING_AGG(field1, ',') WITHIN GROUP (ORDER BY field2 DESC)` rather than
`GROUP_CONCAT(field1, ';') OVER (ORDER BY field2 DESC)` for SQLite.

https://docs.microsoft.com/en-us/sql/t-sql/functions/string-agg-transact-sql?view=sql-server-ver15

othercorey: This clause is specifically for window expressions.

tests/TestCase/Database/FunctionsBuilderTest.php
```diff
@@ -40,7 +40,6 @@ class FunctionsBuilderTest extends TestCase
      */
     public function setUp(): void
     {
-        parent::setUp();
```
markstory, line 43: Why was this necessary?

jimbo2150: Thanks, looks like I accidentally deleted that line I moved the tests out of that file.

jimbo2150:
The errors don't appear to have to do with this PR.

Note that the tests will not cover the MS SQL server code since the test server is using SQL Server 2016 (version 13). The STRING_AGG and other aggregate related functions/expressions were added in SQL Server 2017 (version 14).

src/Database/Expression/GroupedExpression.php
```diff
@@ -0,0 +1,113 @@
...
+     * Overwrites or initializes the orderContainer property with a
+     * FunctionExpression object.
+     *
+     * @param bool $initialize If true, will not use an existing value.
+     * @return $this
+     */
+    protected function resetOrderContainer($initialize = false)
+    {
+        $this->orderContainer = (new FunctionExpression($initialize ? 'OVER ' : $this->getOrderContainer()->getName()))
+            ->setConjunction(' ', false, false);
```
markstory, line 57: Why does this replace the previous container with a new copy?

src/Database/Expression/GroupedExpression.php
```diff
@@ -0,0 +1,113 @@
...
+namespace Cake\Database\Expression;
+
+use Cake\Database\ValueBinder;
+use Closure;
+
+/**
+ * Handles the few special aggregate functions that require special syntax
+ * for grouped ordering.
+ */
+class GroupedExpression extends FunctionExpression
```
markstory, line 26: GroupConcat might be a better name for this based on how it is used.

src/Database/Expression/GroupedExpression.php
```diff
@@ -0,0 +1,113 @@
...
+
+    /**
+     * @var \Cake\Database\Expression\OrderByExpression|null
+     */
+    protected $order;
+
+    /**
+     * @inheritDoc
+     */
+    public function __construct(string $name, array $params = [], array $types = [], string $returnType = 'string')
```
markstory, line 41: If this expression is only used for GroupConcat does the name need to be a parameter?

src/Database/Expression/SelectExpression.php
```diff
@@ -0,0 +1,148 @@
...
+
+    /**
+     * Method that returns the modifiers as either an associative array where
+     * the modifiers are keys or as an array where the modifiers are array
+     * values.
+     *
+     * @param bool $asValue Boolean determining return format.
+     * @return string[]
+     */
+    public function getModifiers(bool $asValue = true): array
```
markstory, line 144: Do we need this parameter? It makes the return value inconsistent and doesn't appear to have any practical usage that the default value doesn't cover.

PR #14629, opened 2020-05-24
4.1 datasource locatorinterface

ADmad:
## The problem

Let's say someone want to use a custom locator for `Table` class instances throughout their app. The  global way to achieve that is customize the table locator using `FactoryLocator::set()` and set a callable for type `Table`.

Doing so would then make `ModelAwareTrait::loadModel()` (used in `Controller`) to use the callable one has set to get table instances. 

Another way to get table instances is using the `LocatorAwareTrait` (`Controller` uses it) and use `LocatorAwareTrait::getTableLocator()->get()`. Now the problem is `LocatorAwareTrait::getTableLocator()` by default will always return a `TableLocator` instance. So one would have to use `LocatorAwareTrait::setTableLocator()` to set a custom table locator which achieves the same as what the callable used for `FactoryLocator`.

This isn't practical. There should be a way to globally set a custom table locator which is used throughout the ORM as single source for table instances.

### Allow the default table locator class in `TableRegistry` to be customizable :-1: 
`LocatorAwareTrait::getTableLocator()` use `TableRegistry::getTableLocator()` to get the table locator. `TableRegistry` currently doesn't provide anyway to customize the default table locator class. So we could add public method to be able to do so.

But again this solution isn't practical. One would then have to configure `FactoryLocator` as well as `TableRegistry`, plus the former uses callable while the latter `ORM\LocatorInterface` instances. So this isn't much of a solution. Moreover `TableRegistry` is essentially a relic for BC compatibility which should be removed in future.

### Add a `Datasource\LocatorInterface` and allow `FactoryLocator` to use both callable and the interface instances. :+1: 

This is the solution I have implemented here.

- Added `Datasource\LocatorInterface` (there's also another issue for this)
- Updated `FactoryLocator` to use both callable and `Datasource\LocatorInterface` instances.
- `ORM\LocatorInterface` now extends `Datasource\LocatorInterface`.
- Updated the codebase to always use `FactoryLocator` as the single consistent source for table locator instances.

## Implications
- To allow `ORM\LocatorInterface` to `Datasource\LocatorInterface` the `Table` typehint for `$object` argument of `ORM\LocatorInterface::set()` had to be removed. This is technically a BC break and will throw an error for anyone who has extended `TableLocator` and customized the method. But IMO that's acceptable and anyone using a custom locator would actually appreciate this change, to be able to customize table locator from a single location.

If the changes seem reasonable I'll go ahead with making static analyzers happy and any other related potential clean ups that can be done.

### DIFF
--- a/src/Datasource/FactoryLocator.php
+++ b/src/Datasource/FactoryLocator.php
@@ -16,7 +16,7 @@ declare(strict_types=1);
  */
 namespace Cake\Datasource;
 
-use Cake\ORM\TableRegistry;
+use Cake\ORM\Locator\TableLocator;
 use InvalidArgumentException;
 
 /**
@@ -64,7 +64,7 @@ class FactoryLocator
     public static function get(string $type)
     {
         if (!isset(static::$_modelFactories['Table'])) {
-            static::$_modelFactories['Table'] = [TableRegistry::getTableLocator(), 'get'];
+            static::$_modelFactories['Table'] = new TableLocator();
         }
 
         if (!isset(static::$_modelFactories[$type])) {
--- a/src/Datasource/FactoryLocator.php
+++ b/src/Datasource/FactoryLocator.php
@@ -27,7 +27,7 @@ class FactoryLocator
     /**
      * A list of model factory functions.
      *
-     * @var callable[]
+     * @var (callable|\Cake\Datasource\LocatorInterface)[]
      */
     protected static $_modelFactories = [];
 
@@ -35,10 +35,10 @@ class FactoryLocator
      * Register a callable to generate repositories of a given type.
      *
      * @param string $type The name of the repository type the factory function is for.
-     * @param callable $factory The factory function used to create instances.
+     * @param callable|\Cake\Datasource\LocatorInterface $factory The factory function used to create instances.
      * @return void
      */
-    public static function add(string $type, callable $factory): void
+    public static function add(string $type, $factory): void
     {
         static::$_modelFactories[$type] = $factory;
     }
@@ -59,9 +59,9 @@ class FactoryLocator
      *
      * @param string $type The repository type to get the factory for.
      * @throws \InvalidArgumentException If the specified repository type has no factory.
-     * @return callable The factory for the repository type.
+     * @return callable|\Cake\Datasource\LocatorInterface The factory for the repository type.
      */
-    public static function get(string $type): callable
+    public static function get(string $type)
     {
         if (!isset(static::$_modelFactories['Table'])) {
             static::$_modelFactories['Table'] = [TableRegistry::getTableLocator(), 'get'];
--- a/src/Datasource/FactoryLocator.php
+++ b/src/Datasource/FactoryLocator.php
@@ -40,6 +40,14 @@ class FactoryLocator
      */
     public static function add(string $type, $factory): void
     {
+        if (!$factory instanceof LocatorInterface && !is_callable($factory)) {
+            throw new InvalidArgumentException(sprintf(
+                '`$factory` must be an instance of Cake\Datasource\LocatorInterface or a callable.'
+                . ' Got type `%s` instead.',
+                getTypeName($factory)
+            ));
+        }
+
         static::$_modelFactories[$type] = $factory;
     }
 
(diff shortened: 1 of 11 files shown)


src/ORM/Locator/TableLocator.php
```diff
@@ -274,7 +274,7 @@ public function exists(string $alias): bool
     /**
      * @inheritDoc
      */
-    public function set(string $alias, Table $object): Table
+    public function set(string $alias, $object): Table
```
dereuromark, line 277: Could this be `object` ?

ADmad: Yes, `object` typehint can be used that argument here (and in the interfaces).

CauanCabral:
Could that changes also remove `Table` as the return typehint?

That way, others datasources, like `CakePHP\ElasticSearch` can utilize the same interfaces and traits to load its repositories.

src/Datasource/FactoryLocator.php
```diff
@@ -27,18 +27,18 @@ class FactoryLocator
...
     /**
      * Register a callable to generate repositories of a given type.
      *
      * @param string $type The name of the repository type the factory function is for.
-     * @param callable $factory The factory function used to create instances.
+     * @param callable|\Cake\Datasource\LocatorInterface $factory The factory function used to create instances.
      * @return void
      */
-    public static function add(string $type, callable $factory): void
+    public static function add(string $type, $factory): void
```
markstory, line 41: Should we do checks here to ensure that $factory is of an acceptable type?

ADmad:
> Could that changes also remove Table as the return typehint?

> That way, others datasources, like CakePHP\ElasticSearch can utilize the same interfaces and traits to load its repositories.

The new `Cake\Datasource\LocatorInterface` does not use `Table` typehint. Only `Cake\ORM\Locator\LocatorInterface` and `Cake\ORM\Locator\LocatorAwareTrait` which are under the `ORM` package use `Table` typehint for return value.

The elastic search's `Cake\ElasticSearch\IndexRegistry` would implement `Cake\Datasource\LocatorInterface`. It already all the required methods except `getConfig()` and `setConfig()`. Besides that just the `Index` typehint for the argument of `IndexRegistry::set($alias, Index $object)` would have to be removed.

src/Datasource/FactoryLocator.php
ADmad: Done

ADmad:
The implementations of methods like `setConfig()`, `getConfig()`, `exits()` of `LocatorInterface` will generally be identical across all classes. So I am wondering if providing a base/abstract class which implements these methods would be nice to avoid code duplication.

src/ORM/Locator/TableLocator.php
```diff
@@ -27,7 +28,7 @@
 /**
  * Provides a default registry/factory for Table objects.
  */
-class TableLocator implements LocatorInterface
+class TableLocator extends AbstractLocator implements LocatorInterface
```
dereuromark, line 31: The interface is already implemented and should be removed here.

ADmad: This is `ORM\LocatorInterace` not `Datsource\Locator\LocatorInterface`.

dereuromark: Hah ok :) My bad.

ADmad:
I have added a `Cake\Datasource\Locator\AbstractLocator` with common methods, but I now question the usefulness of the `setConfig()` and `getConfig()` method. 

They are meant to set(and get) per alias default config which are merged with `$options` passed to `get()` calls. Do people actually use this feature in their apps? I personally have always passed all required options to `get()` calls. (Also the naming of args `$options` vs `$config` and properties `$_config` vs `$_options` is quite confusing.)

Elasticsearch's `IndexRegistry` doesn't have these `setConfig()` and `getConfig()` methods.

So I think having these for the new `LocatorInterface` under `Datasource` is not necessary. We can retain they only for `ORM\LocatorInterface` for backwards compatibility.

markstory (approved):
Looks good to me.

ADmad:
I removed `getConfig()`, `setConfig()` from `Cake\Datasource\Locator\LocatorInterface` and updated `AbstractLocator`.

PR #14630, opened 2020-05-24
Abort deletion of primary entity if deletion of dependent entity fails.

### LINKED ISSUE #14434: Status of delete() of associated model is ignored when  cascadeCallback is true.
This is a (multiple allowed):

* [x] bug
* [x] enhancement

* CakePHP Version: 3.8.5
* Platform and Target: Linux, Apache, Mysql.

When associated model is set to be dependent and cascadeCallbacks is set to true, cakephp triggers separate delete fro each associated entity. If this delete is prohibited by application rules, this information is silently ignored (since return status of delete is not checked). This results in DB corruption when parent entity is deleted, and associated records are still present in DB without any errors being detected.

ADmad:
Fixes #14434

Would be great if someone can help with tests for each association type :slightly_smiling_face:.

### DIFF
--- a/tests/TestCase/ORM/TableTest.php
+++ b/tests/TestCase/ORM/TableTest.php
@@ -3025,9 +3025,38 @@ class TableTest extends TestCase
             'cascadeCallbacks' => true,
         ]);
 
+        $articles = $table->getAssociation('articles')->getTarget();
+        $articles->getEventManager()->on('Model.buildRules', function ($event, $rules) {
+            $rules->addDelete(function ($entity) {
+                if ($entity->author_id === 3) {
+                    return false;
+                } else {
+                    return true;
+                }
+            });
+        });
+
         $entity = $table->get(1);
         $result = $table->delete($entity);
         $this->assertTrue($result);
+
+        $query = $articles->find('all', [
+            'conditions' => [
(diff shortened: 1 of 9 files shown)


markstory:
I can help with tests.

markstory:
@ADmad I added tests for the the association methods. Are there other tests you wanted to see?

ADmad:
Thanks @markstory. I added a additional integration for `Table::delete()` too showing that deletion of primary record is also aborted if deleting dependent records fails.

tests/TestCase/ORM/TableTest.php
```diff
@@ -3025,9 +3025,38 @@ public function testDeleteDependentHasMany()
...
+        $entity = $table->get(3);
+        $result = $table->delete($entity);
+        $this->assertFalse($result);
+
+        $query = $articles->find('all', [
+            'conditions' => [
+                'author_id' => $entity->id,
+            ],
+        ]);
+        $this->assertFalse($query->all()->isEmpty(), 'Should find any rows.');
```
markstory, line 3059:
```suggestion
        $this->assertFalse($query->all()->isEmpty(), 'Should find some rows.');
```

PR #14636, opened 2020-05-27
Fix typing errors

ADmad:

### DIFF
--- a/src/Mailer/Message.php
+++ b/src/Mailer/Message.php
@@ -881,14 +881,14 @@ class Message implements JsonSerializable, Serializable
         ];
         foreach ($relation as $var => $header) {
             if ($include[$var]) {
-                $headers[$header] = current($this->formatAddress($this->{$var}));
+                $headers[$header] = (string)current($this->formatAddress($this->{$var}));
             }
         }
         if ($include['sender']) {
             if (key($this->sender) === key($this->from)) {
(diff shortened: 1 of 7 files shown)


src/Mailer/Message.php
```diff
@@ -868,7 +868,7 @@ public function getHeaders(array $include = []): array
                 'from', 'sender', 'replyTo', 'readReceipt', 'returnPath',
                 'to', 'cc', 'bcc', 'subject',
             ],
-            false
+            ''
```
othercorey, line 871: Why is this wrong? It uses `true` above and seems to check for true/false in the code below.

ADmad: Thanks for spotting this. Need to ensure that the array returned by `getHeaders()` contains strings only but this was the wrong fix.

tests/TestCase/Mailer/MessageTest.php
```diff
@@ -410,7 +410,7 @@ public function testSender()
         $this->assertSame($expected, $this->message->getSender());
 
         $headers = $this->message->getHeaders(['from' => true, 'sender' => true]);
-        $this->assertFalse($headers['From']);
+        $this->assertSame('', $headers['From']);
```
ADmad, line 413: The earlier asserting was incorrect IMO. `getHeaders()` is supposed to return an array of strings.

Issue #14628, opened 2020-05-24
Optional routing parameter


chtheis:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [x] feature-discussion (RFC)

* CakePHP Version: 4.0.7

I don't know if this is a bug, a feature, or something impossible ...

I want to switch my datasource based on an optional prefix in my url, E.g. /test/controller/action would use datasource "test" and /controller/action would use datasource "default". 

I adopted the routing class of admad/cakephp-i18n (@ADmad: Thank you very much for this project, I wouldn't have succeeded without that!), changed {lang} to {ds} and added it to defaults, too. So I have nice patterns where {ds} in /{ds}/{controller} is an optional subpattern and both, /test/controller and /controller match the route.

But: preg_match called for /controller returns an empty string for {ds}. Therefore, in the request object the parameter 'ds' is set to an empty string. When I call Route::url(['controller' => 'other']) the (supposed to be optional) key 'ds' in the url array is set to en empty string in Route::match (precisley: _persistParams). At the end of the method match all options are checked against their corresponding keys in the $url array with preg_match using an on-the-fly created pattern, using my definition of {ds} (i.e. 'test'), without making it optional. And that, of course, would fail.

As a workaround I overloaded the method parse in my route class and replace it with null when empty. The preg_match above would succeed checking against null.

My question: when compiling the route all parameters from the templates are removed from defaults so the information that a parameter was optional is lost. If it were somehow stored, or if the generated pattern for the parameters were stored it could detect in parse that a parameter was not part of the url, or it could use it in the match method instead of the on-the-fly pattern,

ADmad:
You should be connecting 2 routes, 1 for pattern `/{ds}/{controller}` and another for `/{controller}`, have you done that?

chtheis:
I thought about that but then I had to duplicate every route, including those in plugins and also the fallback-routes. It would not be DRY. 

But I can append a '|' to my pattern so it should match an empty string, too.
It would still be a workaround but a nicer one.

ADmad:
Currently the router doesn't have any support for optional parameters/elements. I am not sure how feasible that is. @markstory would have a better idea about it.

chtheis:
It is almost there, but it might be the famous 80%.
If I put ['ds' => null] into defaults the resulting pattern for this parameter will have a '?' appended and /controller will match. Unfortunately it will then continue to use {ds} as if it were found and had been an empty string, and the information, it had a default, is removed.

[...]

PR #14643, opened 2020-05-29
Route will keep optional keys in default

chtheis:
#14628 Allow keys in a route to be optional
Beside removing the code which removes keys from defaults, I had to do a few more changes:
1) At the end of match(...) the code checks that defaults are present and unchanged. To stay compatible I had to create a subset of defaults without keys.

2) When compiling the route preg_match will return an empty string for each optional key in the route, and omit optional keys at the end. I check the result for empty strings and replace them with the default, as it would be done with optional keys at the end. 
Because there is a difference where the optional key is I added 2 tests.

3) Again in match url parameters are tested against options. A null value may pass the test, but to be clearer I explicitly allow the default value.

I hope it is OK if I target 4.next instead of master

### DIFF
--- a/src/Routing/Route/Route.php
+++ b/src/Routing/Route/Route.php
@@ -439,11 +439,11 @@ class Route
         // Assign defaults, set passed args to pass
         foreach ($this->defaults as $key => $value) {
             if (isset($route[$key])) {
-				// preg_match returns an empty string if an optional key was not fournd
-				// We treat this as if the key was not found
-				if ($route[$key] !== '') {				
-	                continue;
-				}
+                // preg_match returns an empty string if an optional key was not fournd
+                // We treat this as if the key was not found
+                if ($route[$key] !== '') {
+                    continue;
+                }
             }
             if (is_int($key)) {
                 $route['pass'][] = $value;
@@ -652,12 +652,12 @@ class Route
             return null;
         }
         unset($url['_method'], $url['[method]'], $defaults['_method']);
-		
-		// The key names
-		$keyNames = array_flip($this->keys);
-		
-		// All defaults which are not a key
-		$defaultsNotKey = array_diff_key($defaults, $keyNames);
+
+        // The key names
+        $keyNames = array_flip($this->keys);
+
+        // All defaults which are not a key
+        $defaultsNotKey = array_diff_key($defaults, $keyNames);
 
         // Missing defaults, which are not a key, is a fail
         if (array_diff_key($defaultsNotKey, $url) !== []) {
@@ -712,11 +712,11 @@ class Route
         // check patterns for routed params, if they are not the default value
         if (!empty($this->options)) {
             foreach ($this->options as $key => $pattern) {
-				if (!array_key_exists($key, $defaults) || $defaults[$key] !== $url[$key]) {
-					if (isset($url[$key]) && !preg_match('#^' . $pattern . '$#u', (string)$url[$key])) {
-						return null;
-					}
-				}
+                if (!array_key_exists($key, $defaults) || $defaults[$key] !== $url[$key]) {
+                    if (isset($url[$key]) && !preg_match('#^' . $pattern . '$#u', (string)$url[$key])) {
+                        return null;
+                    }
+                }
             }
         }
         $url += $hostOptions;
--- a/src/Routing/Route/Route.php
+++ b/src/Routing/Route/Route.php
@@ -339,11 +339,6 @@ class Route
         $this->_compiledRoute = '#^' . $parsed . '[/]*$#' . $mode;
         $this->keys = $names;
 
-        // Remove defaults that are also keys. They can cause match failures
-        foreach ($this->keys as $key) {
-            unset($this->defaults[$key]);
-        }
-
         $keys = $this->keys;
         sort($keys);
         $this->keys = array_reverse($keys);
@@ -444,7 +439,11 @@ class Route
         // Assign defaults, set passed args to pass
         foreach ($this->defaults as $key => $value) {
             if (isset($route[$key])) {
-                continue;
+				// preg_match returns an empty string if an optional key was not fournd
+				// We treat this as if the key was not found
+				if ($route[$key] !== '') {				
+	                continue;
+				}
             }
             if (is_int($key)) {
                 $route['pass'][] = $value;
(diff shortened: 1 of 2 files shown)


src/Routing/Route/Route.php
```diff
@@ -444,7 +439,11 @@ public function parse(string $url, string $method): ?array
         // Assign defaults, set passed args to pass
         foreach ($this->defaults as $key => $value) {
             if (isset($route[$key])) {
-                continue;
+                // preg_match returns an empty string if an optional key was not fournd
```
markstory, line 442:
```suggestion
                // preg_match returns an empty string if an optional key was not found
```

src/Routing/Route/Route.php
```diff
@@ -444,7 +439,11 @@ public function parse(string $url, string $method): ?array
...
+                // We treat this as if the key was not found
+                if ($route[$key] !== '') {
```
markstory, line 444: Couldn't this condition be combined with the one above?

src/Routing/Route/Route.php
```diff
@@ -705,11 +709,13 @@ public function match(array $url, array $context = []): ?string
...
         }
 
-        // check patterns for routed params
+        // check patterns for routed params, if they are not the default value
         if (!empty($this->options)) {
             foreach ($this->options as $key => $pattern) {
-                if (isset($url[$key]) && !preg_match('#^' . $pattern . '$#u', (string)$url[$key])) {
-                    return null;
+                if (!array_key_exists($key, $defaults) || $defaults[$key] !== $url[$key]) {
+                    if (isset($url[$key]) && !preg_match('#^' . $pattern . '$#u', (string)$url[$key])) {
```
markstory, line 716: Could these conditions be combined with those on the line above?

tests/TestCase/Routing/Route/RouteTest.php
```diff
@@ -1247,6 +1248,99 @@ public function testParseWithPassDefaults()
...
+        ];
+        $this->assertSame($expected, $result);
+
+        // Test route against path without {pref}
+        $result = $route->parse('/posts', 'GET');
+        $expected = [
+            'start' => null,
+            'controller' => 'posts',
+            'pass' => [],
+            '_matchedRoute' => '/{start}/{controller}',
```
markstory, line 1281: I don't think this should be a match. You have `start` defined as needing to match a pattern and both posts and `null` do not match that pattern.

tests/TestCase/Routing/Route/RouteTest.php
```diff
@@ -1247,6 +1248,99 @@ public function testParseWithPassDefaults()
...
+        $this->assertSame($expected, $result);
+
+        // Test route against path without {pref}
+        $result = $route->parse('/posts/', 'GET');
+        $expected = [
+            'controller' => 'posts',
+            'pass' => [],
+            'end' => null,
+            '_matchedRoute' => '/{controller}/{end}',
+        ];
```
markstory, line 1334: This looks the same as the test on 1316-1323.

src/Routing/Route/Route.php
chtheis: Done

src/Routing/Route/Route.php
chtheis: Done.

src/Routing/Route/Route.php
chtheis: Done.
I had them separate because I thought it will be easier to follow the change and it would be easier to read.

tests/TestCase/Routing/Route/RouteTest.php
chtheis: Good question. I'd say it is valid because start is the same as the default.

We have a pattern which restricts what can be the value of start, and we have a default to be used if the pattern is not found, i.e. the key is not present. Does that mean the pattern must include default? I don't think so, because then having a default doesn't make sense: it would be part of the pattern and having a default is superfluous, thus the key is no longer optional but required.

My interpretation is that a value must either match the pattern or be the same as the default, so I don't have to make the default value match the pattern. Which could be difficult if the default is null.

tests/TestCase/Routing/Route/RouteTest.php
chtheis: The difference is the trailing '/' in parse in this case. The result should, of course, be the same.

markstory: :+1: I missed the trailing slash.

tests/TestCase/Routing/Route/RouteTest.php
markstory: I'm a bit concerned that this will create more opportunity for ambiguity to be created in routes. Optional leading and trailing parameters can create overlaps with other routes. I guess that can be mitigated by including 'be careful' blocks in the documentation.

chtheis: It's not really a new feature, it was possible before, just with some rough edges. 

I think it was already possible to have optional keys, which may not require to match a pattern. There is a test for it in RouteTest, testMatchNullValueOptionalKey. That's why I added only tests for the parse part and not the match part. 

The new part is that it is now possible to have optional keys which, when present, must match a pattern like a set of allowed values, and that the default (usually null) doesn't have to match the pattern. I think this way it is possible to write more restrictive routes than it was possible before, with less hassle and, what I believe, more intuitive.

But you are right: users should be be careful with optional keys especially at the start of a route and they should be advised to make use of pattern matching to minimize the risk of ambiguities.

othercorey: Should we have some sort of validation in debug mode to warn users?

chtheis: One additional remark on my last reply:
With the old code I had to write my route as Route('/{ds}/{controller}, ['ds' => ''], ['ds' => 'test|'])
With the new code I can write my route as Route('/{ds}/{controller}, ['ds' => null], ['ds' => 'test'])
So it was possible to have optional keys. I consider the new way more intuitive because I'm less restricted what to choose as a default and thus I can choose a more reasonable one. And that is what the changes are about: not introducing new ways but better ways.

src/Routing/Route/Route.php
```diff
@@ -705,11 +707,14 @@ public function match(array $url, array $context = []): ?string
             return null;
         }
 
-        // check patterns for routed params
+        // check patterns for routed params, if they are not the default value
         if (!empty($this->options)) {
             foreach ($this->options as $key => $pattern) {
-                if (isset($url[$key]) && !preg_match('#^' . $pattern . '$#u', (string)$url[$key])) {
-                    return null;
+                if (
```
ADmad, line 713: The conditions for this `if` are practically unreadable. It would be better to split the checks and assign them to well named variables and then use them inside the `if` so that it's easier to understand what's going on here.

ADmad:
The route matching code is already not easy to follow and these changes seem to complicate it further. I also feature that allowing optional route params by default could cause false matches.

Perhaps explicitly configuring which params are optional would with both these issues?

othercorey:
I would prefer an alternative solution to avoid optional parameters at arbitrary positions which will have issues we don't want to support/fix in the future.

- Only support optional parameter at the root position `/{optional}/rest/of/url`
- Special case a default for `I18nRoute` that is used if the route matches without the lang prefix.

This should reduce the overhead and support your specific situation.

chtheis:
Optional keys are an **existing** feature, not something I invent with this PR. There is even a test for it with an optional key in the middle. Removing them or restricting the use of them may break backwards compatibility.

All I do in this PR is to not delete the key from $defaults, so the fact, a key was optional, and its default values will be known later. Which in turn allows for null as the default instead of forcing one to make an empty string the default. You could also use that info to mark optional keys when printing out all routes if no match was found, something which afaik is not done today.

If you want to address your concerns you may think about a warning in debug mode, when optional keys without a pattern are used in a route.

Issue #14642, opened 2020-05-29
Wrong magic method signature


othercorey:
https://wiki.php.net/rfc/magic-methods-signature

PHP 8 is considering better signature checks and found an error in cakephp.

wrong native return type of __call(): void
https://github.com/cakephp/cakephp/blob/8718b2e9bbd092b991a1da6472d1a28e89b63dac/src/Controller/Component/FlashComponent.php#L137

dereuromark:
I agree, those should not have void as return type.
Semantically speaking such a method can both return something and null, but not really void (as it is not fully assertable to only be used as setter and not getter).

So let's just remove it in 4.1.

ADmad:
It's strange that the RFC only mentions `__toString()` with regarding to future runtime time. Why not just add return type `string` for the method?

othercorey:
Looks like they have by making it bound to type Stringable.

https://3v4l.org/jIg7b/rfc#git-php-master

ADmad:
Ah so we can add `string` return typehint to all `__toString()` usages.

othercorey:
Btw, only have one other type hinted `__call()`, and it wasn't flagged. That should indicate it's compatible with their intended typehint.

https://github.com/cakephp/cakephp/blob/a98ed4ef6a55cb524d67da33e668fbde76edb78f/src/Database/FunctionsBuilder.php#L293

ADmad:
So since `void` is a special type they don't consider it as covariant with `mixed`. Using any other type except `void` won't generate an error.

dereuromark:
Yes.

PR #14645, opened 2020-05-29
Removed void return typehint from __call()

othercorey:
https://github.com/cakephp/cakephp/issues/14642

Let's see what complains. Leaving the @return hint for now.


dereuromark:
We might want to add a sniffer adjustment to also not add such return typehint for magic methods.

Issue #14623, opened 2020-05-23
Add $session->readOrFail() and $session()->readOrCreate() 


inoas:
This is a (multiple allowed):

* [x] enhancement
* [x] feature-discussion (RFC)

* CakePHP Version: 3.7 / 3.8

### What you did
Used sessions to read/write data.

### What happened
Got silent null values.

### What you expected to happen
Stuff to crash.

dereuromark:
orCreate() is usually an antipattern - or not applicable in this context IMO. 

    read(?string $name = null)

could also become

    read(?string $name = null, $default = null)

or alike, having a default value where needed.

But otherwise the orFail() sounds reasonable.

markstory:
I prefer the `read()` with a default value over `readOrCreate()`

inoas:
@dereuromark however null is a legit value so ?? $default only works for keys that are not considered valid when set to null - right?

dereuromark:
Well, null in that context it is not. Check also all other default cases.
The same for *orFail() as that one for all our methods so far also excludes null as value.
If null would would be a valid value, you would need custom logic here either way - if we change that behavior we create a snowflake.

The orFail() just makes more sense here as defaulting is already super easy anyway.

Rule of thumb: Don't store null as meaningful value, cast it to the type you need if you want to use such methods.

othercorey:
Isn't this signature more consistent with cake? I'm not sure if we like the nullable name convention in cake still.

`read(?string $name = null, $default = null)`

dereuromark:
Yes it would be :) just saying that it is not null value aware.

dereuromark:
If you want a defaulting or orFail() added, now would be the time, as RC will now be finalized for both 4.1 and 3.9 backport.

dereuromark:
I opened https://github.com/cakephp/cakephp/pull/14647 for further fine-tuning.

PR #14647, opened 2020-05-30
Add defaulting and *orFail() to Session.

dereuromark:
This resolves https://github.com/cakephp/cakephp/issues/14623

The advantage would be to have a string typehint for readOrFail(), same as for Configure.

### DIFF
--- a/src/Http/Session.php
+++ b/src/Http/Session.php
@@ -460,7 +460,7 @@ class Session
      */
     public function readOrFail(string $name)
     {
-        if ($this->check($name) === false) {
+        if (!$this->check($name)) {
             throw new RuntimeException(sprintf('Expected session key "%s" not found.', $name));
         }
 
--- a/src/Http/Session.php
+++ b/src/Http/Session.php
@@ -430,10 +430,11 @@ class Session
      * Returns given session variable, or all of them, if no parameters given.
      *
      * @param string|null $name The name of the session variable (or a path as sent to Hash.extract)
+     * @param mixed $default The return value when the path does not exist
      * @return string|array|null The value of the session variable, null if session not available,
      *   session not started, or provided name not found in the session.
      */
-    public function read(?string $name = null)
+    public function read(?string $name = null, $default = null)
     {
         if ($this->_hasSession() && !$this->started()) {
             $this->start();
@@ -447,7 +448,23 @@ class Session
             return $_SESSION ?: [];
         }
 
-        return Hash::get($_SESSION, $name);
+        return Hash::get($_SESSION, $name, $default);
+    }
+
+    /**
+     * Returns given session variable, or throws Exception if not found.
+     *
+     * @param string $name The name of the session variable (or a path as sent to Hash.extract)
+     * @throws \RuntimeException
+     * @return array|string|null
+     */
+    public function readOrFail(string $name)
+    {
+        if ($this->check($name) === false) {
+            throw new RuntimeException(sprintf('Expected session key "%s" not found.', $name));
+        }
+
+        return $this->read($name);
     }
 
     /**
--- a/src/Http/Session.php
+++ b/src/Http/Session.php
@@ -431,7 +431,7 @@ class Session
      *
      * @param string|null $name The name of the session variable (or a path as sent to Hash.extract)
      * @param mixed $default The return value when the path does not exist
-     * @return string|array|null The value of the session variable, null if session not available,
+     * @return mixed|null The value of the session variable, null if session not available,
      *   session not started, or provided name not found in the session.
      */
     public function read(?string $name = null, $default = null)
@@ -456,7 +456,7 @@ class Session
      *
      * @param string $name The name of the session variable (or a path as sent to Hash.extract)
      * @throws \RuntimeException
-     * @return array|string|null
+     * @return mixed|null
      */
     public function readOrFail(string $name)
     {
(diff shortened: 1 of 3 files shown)


dereuromark:
Are those fails a psalm issue?
```
##[error]Argument 2 of str_replace expects array<array-key, string>|string, array{string, int} provided
##[error]Argument 2 of str_replace expects array<array-key, string>|string, array<non-empty-string, false|null|string> provided
```

src/Http/Session.php
```diff
@@ -447,7 +448,23 @@ public function read(?string $name = null)
...
-        return Hash::get($_SESSION, $name);
+        return Hash::get($_SESSION, $name, $default);
+    }
+
+    /**
+     * Returns given session variable, or throws Exception if not found.
+     *
+     * @param string $name The name of the session variable (or a path as sent to Hash.extract)
+     * @throws \RuntimeException
+     * @return array|string|null
```
ADmad, line 459: One could write and read integer/float too isn't?

dereuromark: Yes, I copy and pasted this from the read() method. I guess both should be mixed|null then.

ADmad: Just `mixed`. `mixed` includes `null`. https://wiki.php.net/rfc/mixed_type_v2#nullability

src/Http/Session.php
```diff
@@ -430,10 +430,11 @@ public function check(?string $name = null): bool
      * Returns given session variable, or all of them, if no parameters given.
      *
      * @param string|null $name The name of the session variable (or a path as sent to Hash.extract)
-     * @return string|array|null The value of the session variable, null if session not available,
+     * @param mixed $default The return value when the path does not exist
+     * @return mixed|null The value of the session variable, null if session not available,
```
ADmad, line 434:
```suggestion
     * @return mixed The value of the session variable, null if session not available,
```

src/Http/Session.php
```diff
@@ -447,14 +448,30 @@ public function read(?string $name = null)
...
-        return Hash::get($_SESSION, $name);
+        return Hash::get($_SESSION, $name, $default);
+    }
+
+    /**
+     * Returns given session variable, or throws Exception if not found.
+     *
+     * @param string $name The name of the session variable (or a path as sent to Hash.extract)
+     * @throws \RuntimeException
+     * @return mixed|null
```
ADmad, line 459:
```suggestion
     * @return mixed
```

src/Http/Session.php
```diff
@@ -447,14 +448,30 @@ public function read(?string $name = null)
...
+     */
+    public function readOrFail(string $name)
+    {
+        if ($this->check($name) === false) {
```
ADmad, line 463:
```suggestion
        if (!$this->check($name)) {
```

src/Http/Session.php
```diff
@@ -447,14 +448,30 @@ public function read(?string $name = null)
...
+
+        return $this->read($name);
     }
 
     /**
      * Reads and deletes a variable from session.
      *
      * @param string $name The key to read and remove (or a path as sent to Hash.extract).
-     * @return mixed The value of the session variable, null if session not available,
+     * @return mixed|null The value of the session variable, null if session not available,
```
ADmad, line 474:
```suggestion
     * @return mixed The value of the session variable, null if session not available,
```

src/Http/Session.php
dereuromark: No, that is on purpose. It must be |null - as you can see also on RFC passed for mixed in PHP8.
mixed does really not include null - as I have been telling you guys for years. The future PHP version will also enforce that.

src/Http/Session.php
dereuromark: See above.

src/Http/Session.php
dereuromark: See above.

src/Http/Session.php
dereuromark: Sure, thats fair. I just copied this from the other check() method, and as such it should be the same in both cases, though.

src/Http/Session.php
dereuromark: Oh darn, they merged the other one? https://wiki.php.net/rfc/mixed_type_v2
Where mixed is basically `?mixed`? That's sad...

dereuromark: They really screwed this up :(
https://wiki.php.net/rfc/union_types_v2 also is in line what should have been `?mixed`...

dereuromark: As long as there is still a chance for them to fix this, I would suggest we keep the proper nullable declarations in place - as this can easily be cleaned up by tooling then either way as long as we have the additional info in place. Losing it now makes it harder to revive.
Also, this is still way clearer in terms of nullable communication if we adhere to this verbosity.

othercorey: I don't know about php union types, but I do prefer to have some indication of nullable in the docblock types.

ADmad: Okay, you guys can have it your way until static analyzers or PHP itself starts complaining about it :slightly_smiling_face:.

ADmad (approved):

src/Http/Session.php
othercorey: I agree mixed a difficult notation ... but just for consistency until we find out what blows up in the php 8 analyzers.

PR #14652, opened 2020-05-30
Cake 4: Microoptimization for ServerRequest::is() and ServerRequest::isAll()

andrii-pukhalevych:
Microoptimization for `ServerRequest::is()` and `ServerRequest::isAll()`to stop and return on first matched/unmatched type.

Instead of checking all passed types (calling `array_map()` + `array_filter()`), it breaks on first type via `return`.

### DIFF
--- a/src/Http/ServerRequest.php
+++ b/src/Http/ServerRequest.php
@@ -628,9 +628,13 @@ class ServerRequest implements ServerRequestInterface
      */
     public function isAll(array $types): bool
     {
-        $result = array_filter(array_map([$this, 'is'], $types));
+        foreach ($types as $type) {
+            if (!$this->is($type)) {
+                return false;
+            }
+        }
 
-        return count($result) === count($types);
+        return true;
     }
 
     /**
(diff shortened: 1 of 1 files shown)


dereuromark:
The basic loops also look somewhat more readable :+1:

dereuromark:
the docblock for is() could also probably be

    @param string|string[] $type

andrii-pukhalevych:
What is correct way to pull this into 3.9/4.0?  Maybe even to cake 2, if next release will be happening.

ADmad:
Ideally you could have opened the PR against `master` then it would have been merged to `4.next` will regular merges we do.

You can now open new PR targetting `3.0`, `master` and `2.x`.

andrii-pukhalevych:
> You can now open new PRs targetting `3.next`, `master` and `2.x`.
Added PRs to these branches.
Initially I didnt expect that this update will affect all versions.
But I believe that it will be useful because `ServerRequest::is()` is used very often.

PR #14656, opened 2020-06-01
Removed MariaDB version check

othercorey:
This was left from the CommonTableExpression work. We don't support MariaDB in the mysql driver directly anywhere else. I assume users will have a separate driver for this.

If we do need to support it, we should decide and fix it now so we can have proper version checks in other places.

### DIFF
--- a/src/Database/Driver/Mysql.php
+++ b/src/Database/Driver/Mysql.php
@@ -253,15 +253,7 @@ class Mysql extends Driver
     public function supportsCTEs(): bool
     {
         if ($this->supportsCTEs === null) {
-            $version = $this->getVersion();
-            if (strpos($version, 'MariaDB') !== false) {
-                preg_match('/(\d+\.\d+.\d+)-MariaDB/i', $version, $matches);
-                $version = $matches[1];
-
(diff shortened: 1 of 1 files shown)


ADmad:
While we may not officially document it I am sure there are people using the `Mysql` driver for MariaDB so we should keep support for it. We can move the check to a more appropriate location if necessary.

othercorey:
> While we may not officially document it I am sure there are people using the `Mysql` driver for MariaDB, so we should keep support for it. We can move the check to a more appropriate location if necessary.

I'm ok with that. To summarize my thoughts/discussion in the CTE PR:

- Makes sense to have a separate MariaDB version
- Doing versions checks off the mysql version that mariadb version "supports" sounds dangerous

I think in most places we could just do something like `if mysql version > X or maria version > Y` and not check if it's actually one or the other. So the empty maria version should fail any version check.

ADmad:
@markstory Wasn't this merged prematurely? The alternate code to removal of MariaDB version check in current PR hasn't been implemented.

markstory:
My bad. Sorry about that.

ADmad:
No problem. I'll create a new separate issue.

Issue #14004, opened 2019-12-10
Schema Metadata Cache is always activated


ravage84:
* CakePHP Version: 3.8.7

It seems ``SchemaCache::_getSchema()`` always uses an activated cache, even when disabled through setting ``$config['cacheMetadata']`` to  ``false`` or ``0``.

https://github.com/cakephp/cakephp/blob/ed758075d1a12abebcc958baa450c49961c81dea/src/Database/SchemaCache.php#L103-L106

Proof:
https://3v4l.org/3p4ph

This line was likey copied from ``OrmCacheShell::_getSchema()`` in PR https://github.com/cakephp/cakephp/pull/10977

https://github.com/cakephp/cakephp/pull/10977/files#diff-2e213f8f3145bfa87b3f292c3bb2f2dcL109-L113.

But I'm wondering what these lines actually want to achieve.

Probably, is should be:

````php
$config = $connection->config();
if (!isset($config['cacheMetadata'])) {
    $connection->cacheMetadata(true);
}
````

This way, the cache would be activated when not set deliberately.

May be @burzum  can shed some light?

Found by @odan

ADmad:
`SchemaCache` is only used by `SchemaCacheShell` which is meant to build/clear schema cache. It probably force enables metadata caching for the connection as other it won't make much to use the shell would it?

othercorey:
@ravage84 Can we fix this for 3.9?

PR #14658, opened 2020-06-01
Added info message about enabling metadata cache when using SchemaCache.

othercorey:
Closes https://github.com/cakephp/cakephp/issues/14004

This restores the notice about always enabling metadata cache when using SchemaCache.

### DIFF
--- a/src/Database/SchemaCache.php
+++ b/src/Database/SchemaCache.php
@@ -16,6 +16,7 @@ namespace Cake\Database;
 
 use Cake\Cache\Cache;
 use Cake\Database\Connection;
+use Cake\Log\Log;
 use RuntimeException;
 
 /**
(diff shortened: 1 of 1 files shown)


othercorey:
The previous code that did this was https://github.com/cakephp/cakephp/blob/3a6bd7583204e3619b8bb98b385cc1b27d2724eb/src/Shell/OrmCacheShell.php#L111

src/Database/SchemaCache.php
```diff
@@ -104,6 +105,7 @@ public function getSchema(Connection $connection)
 
         $config = $connection->config();
         if (empty($config['cacheMetadata'])) {
+            Log::info("Metadata cache was disable in config for `{$connection->configName()}`. Enabling to clear cache.");
```
ADmad, line 108:
```suggestion
            Log::info("Metadata cache was disabled in config for `{$connection->configName()}`. Enabling to clear cache.");
```

Nitpick: I personally prefer using `sprintf()` instead of doing method calls inside strings.

othercorey: done.

ravage84:
Thanks, @othercorey

PR #14659, opened 2020-06-01
SaveMany should call `afterSaveCommit`

dakota:
This fills a current gap where `saveMany()` does not call/trigger the `afterSaveCommit` callback (While other `many()` uses, like `deleteMany()` correctly trigger their `commit` callback)

Unsure if this should be in 4.next, or 4.0?

Will create tests shortly

### DIFF
--- a/src/ORM/Table.php
+++ b/src/ORM/Table.php
@@ -2225,13 +2225,13 @@ class Table implements RepositoryInterface, EventListenerInterface, EventDispatc
 
             throw new PersistenceFailedException($failed, ['saveMany']);
         }
-        
+
         if ($this->_transactionCommitted($options['atomic'], $options['_primary'])) {
             foreach ($entities as $entity) {
                 $this->dispatchEvent('Model.afterSaveCommit', compact('entity', 'options'));
             }
         }
-        
+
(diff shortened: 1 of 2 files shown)


dereuromark:
Almost sounds like a bugfix for master, but with 4.1 now already in RC, there is probably no need to force that now.

ADmad:
With 4.1 almost ready we can target this to 4.next.

src/ORM/Table.php
```diff
@@ -2180,6 +2180,14 @@ public function saveManyOrFail(iterable $entities, $options = []): iterable
      */
     protected function _saveMany(iterable $entities, $options = []): iterable
     {
+        $options = new ArrayObject(
```
othercorey, line 2183: Does it need to be an ArrayObject?

markstory: Yes, the options in ORM events are ArrayObject instances.

dakota:
Bump :)

markstory:
What needs bumping here? This was merged.

dakota:
😳

Well, this is embarrassing. I didn't even see that. Sorry!

PR #14666, opened 2020-06-03
Deprecate TableRegistry.

ADmad:

### DIFF
--- a/src/ORM/TableRegistry.php
+++ b/src/ORM/TableRegistry.php
@@ -55,6 +55,9 @@ use Cake\ORM\Locator\LocatorInterface;
  * // Prior to 3.6.0
  * $table = TableRegistry::get('Users', $config);
  * ```
+ *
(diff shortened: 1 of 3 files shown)


src/ORM/TableRegistry.php
```diff
@@ -55,6 +55,9 @@
  * // Prior to 3.6.0
  * $table = TableRegistry::get('Users', $config);
  * ```
+ *
+ * @deprecated 4.1.0 Use \Cake\ORM\Locator\LocatorAwareTrait::getTableLocator()
```
dereuromark, line 59: Can we add a {@link or see tag here too, to be able to follow the link?

ADmad: The second example has an argument too, will IDEs be able to navigate `@see`/`@link` tag for that?

dereuromark: We might not need to, the first is usually useful enough.

dereuromark (approved):

PR #14668, opened 2020-06-04
Add ORM dependency to Datasource

othercorey:


src/Datasource/composer.json
```diff
@@ -26,6 +26,7 @@
     "require": {
         "php": ">=7.2.0",
         "cakephp/core": "^4.0",
+        "cakephp/orm": "^4.0",
```
ADmad, line 29: Why?

othercorey: https://github.com/cakephp/cakephp/blob/8e81c28c33f941b65ae6d73c609abd58c6af4785/src/Datasource/FactoryLocator.php#L76

othercorey: Of course adding cakephp/orm is incorrect, but we can't ignore this.

ADmad: We sure can't, have been doing it for years :slightly_smiling_face:. 

Adding a circular dependency isn't a solution here.

othercorey: It isn't meant to be a solution, just tracking the issue for discussion.

Sure, we can ignore it. It just makes the Datasource package throw confusing errors as it's not mentioned as a requirement in the readme.

ADmad: > It isn't meant to be a solution, just tracking the issue for discussion.

Then opening an issue would have been better instead of this PR, or at least keeping the PR in draft state. In current state I though this is the change you want to make.

PR #14679, opened 2020-06-06
3.next backports

dereuromark:
After diffing 4.next and 3.next, I found the following docblock and code changes that could be backported.

Some methods seem to be completely missing in API (validation), and here we would also need to backport tests then, of course.
https://github.com/cakephp/cakephp/pull/13538
=> I pulled those now into a separate PR https://github.com/cakephp/cakephp/pull/14682 for easier review.
Once that is done, I can rebase this one.

Note:
Validation has been fixed towards how 4.x behaves.
Now:

    $this->assertTrue(Validation::lengthBetween(1, 1, 3));

We should then also document this in the changelogs if we continue with it.

### DIFF
--- a/tests/TestCase/Auth/DigestAuthenticateTest.php
+++ b/tests/TestCase/Auth/DigestAuthenticateTest.php
@@ -61,14 +61,12 @@ class DigestAuthenticateTest extends TestCase
     {
         parent::setUp();
 
-        $salt = 'foo.bar';
-        Security::setSalt($salt);
         $this->Collection = $this->getMockBuilder(ComponentRegistry::class)->getMock();
         $this->auth = new DigestAuthenticate($this->Collection, [
             'realm' => 'localhost',
             'nonce' => 123,
             'opaque' => '123abc',
-            'secret' => $salt,
+            'secret' => Security::getSalt(),
             'passwordHasher' => 'ShouldNeverTryToUsePasswordHasher',
         ]);
 
--- a/tests/TestCase/Auth/DigestAuthenticateTest.php
+++ b/tests/TestCase/Auth/DigestAuthenticateTest.php
@@ -61,12 +61,14 @@ class DigestAuthenticateTest extends TestCase
     {
         parent::setUp();
 
+        $salt = 'foo.bar';
+        Security::setSalt($salt);
         $this->Collection = $this->getMockBuilder(ComponentRegistry::class)->getMock();
         $this->auth = new DigestAuthenticate($this->Collection, [
             'realm' => 'localhost',
             'nonce' => 123,
             'opaque' => '123abc',
-            'secret' => Security::getSalt(),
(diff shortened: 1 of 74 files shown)


ADmad:
This will again cause multiple conflicts when doing `3.next` to `master` merge :slightly_frowning_face:

dereuromark:
I thought we didnt do that anymore? And thus so many changes here both ways anyway?

ADmad:
Oh right, we only backport required changes. Sorry, I forgot.

dereuromark:
I am still no git expert, but if the lines are exactly the same, on both sides, it should also go through without a conflict to resolve from my experience. But I could be wrong.

markstory:
We haven't been doing merges from 3.next -> master in months. Doing backports by hand is a bit easier as we make fewer mistakes backporting incompatible typehints and don't have to deal with conflicts.

othercorey:
The downside is 3.8 to 3.9 is going to be a huge diff with the formatting/doc changes we've made.

dereuromark:
We can merge those first and i can reapply then these changes on top.

dereuromark:
I rebased. btw, the newly added see links are not valid as they are ( https://github.com/cakephp/cakephp/pull/14679/commits/b018d3ab735e25a4b227d79d5aa9c4dc4c3baa09 ), missing the leading slash to work.

I am, however, a bit confused about the failing tests. I checked and cannot find the reason why any of the changes would affect the tests this way.

dereuromark:
@othercorey Is master merged? And any idea what the failing test might be about?

othercorey:
The failures are caused by the salt value. If you revert the change and return null when the Salt isn't set then it succeeds.

dereuromark:
Do you think we should do that? Or would it be better to try to resolve the issue?
I am fine either way, but feel like that backporting 4.x could be the safer one here.

dereuromark:
After your tip I found it looking into https://github.com/cakephp/cakephp/pull/12964
The tests were not fully correct without all changes, and I had to backport the whole PR here.

tests/TestCase/Auth/DigestAuthenticateTest.php
```diff
@@ -61,12 +61,14 @@ public function setUp()
     {
         parent::setUp();
 
+        $salt = 'foo.bar';
+        Security::setSalt($salt);
```
othercorey, line 65: This would need to be reset although I'm not sure why it needs to be changed for this test.

dereuromark: That is a valid question. But we should check that on 4.x first, and if we find the reason to do so, I can update this PR.

dereuromark: You were right, in 4.next latest this is also not there anymore, so I cleaned this up now too.

Issue #14675, opened 2020-06-06
Issues of SecurityComponent and CsrfProtectionMiddleware


smarek:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.8.12
* Platform and Target: nginx, php 7.4

When external tester tried malforming requests to app running on mentioned environment/framework version, few issues came up in logs, and I'm posting stacktraces, for your consideration.

**first**
```
Jun 04 17:20:27 first [293607]: warning: Warning (2): urldecode() expects parameter 1 to be string, array given in [/var/www/redacted/vendor/cakephp/cakephp/src/Controller/Component/SecurityComponent.php, line 371]
Jun 04 17:20:27 first [293607]: warning: Request URL: /redacted
Jun 04 17:20:27 first [293607]: warning: Client IP: 1.2.3.4
Jun 04 17:20:27 first [293607]: warning: Trace:
Jun 04 17:20:27 first [293607]: warning: Cake\Error\BaseErrorHandler::handleError() - CORE/src/Error/BaseErrorHandler.php, line 168
Jun 04 17:20:27 first [293607]: warning: ErrorEmail\Error\ErrorHandler::_callParentHandleError() - ROOT/vendor/ebrigham1/cakephp-error-email/src/Error/ErrorHandler.php, line 113
Jun 04 17:20:27 first [293607]: warning: ErrorEmail\Error\ErrorHandler::handleError() - ROOT/vendor/ebrigham1/cakephp-error-email/src/Error/ErrorHandler.php, line 63
Jun 04 17:20:27 first [293607]: warning: urldecode - [internal], line ??
Jun 04 17:20:27 first [293607]: warning: Cake\Controller\Component\SecurityComponent::_validToken() - CORE/src/Controller/Component/SecurityComponent.php, line 371
Jun 04 17:20:27 first [293607]: warning: Cake\Controller\Component\SecurityComponent::_validatePost() - CORE/src/Controller/Component/SecurityComponent.php, line 327
Jun 04 17:20:27 first [293607]: warning: Cake\Controller\Component\SecurityComponent::startup() - CORE/src/Controller/Component/SecurityComponent.php, line 124
Jun 04 17:20:27 first [293607]: warning: Cake\Event\EventManager::_callListener() - CORE/src/Event/EventManager.php, line 352
[...]

PR #14680, opened 2020-06-06
Fix missing typechecks in SecurityComponent and CsrfProtectionMiddleware

markstory:
Add additional type checks to prevent warnings reported in #14675


ADmad:
This will need to be forward ported to 4.x and similar check added to `FormProtector` class.

PR #14689, opened 2020-06-08
Pin phpstan for now.

dereuromark:
latest PHPStan release has a few internal bugs
we should pin current to keep master green

pending resolution of
- [ ] https://github.com/phpstan/phpstan/issues/3409
- [ ] https://github.com/phpstan/phpstan/issues/3404

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -111,7 +111,7 @@
             "@phpstan",
             "@psalm"
         ],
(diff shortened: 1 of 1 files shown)


ondrejmirtes:
Both issues resolved in PHPStan 0.12.27.

PR #14692, opened 2020-06-09
Update phpstan config.

ADmad:

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -111,7 +111,7 @@
             "@phpstan",
             "@psalm"
         ],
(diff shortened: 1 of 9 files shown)


ADmad:
Awaiting following issues to be fixed:

https://github.com/phpstan/phpstan/issues/3422
https://github.com/phpstan/phpstan/issues/3423

othercorey:
Looks ok to me.

dereuromark:
Looking at the shrinking https://github.com/cakephp/cakephp/blob/master/phpstan-baseline.neon file it seems we have now by far the cleanest codebase (in terms of static analysis) - regarding all larger repos out there in terms of popularity.

PR #14696, opened 2020-06-10
add branch alias to ease testing of 4.next in projects

swiffer:
https://getcomposer.org/doc/articles/aliases.md#branch-alias

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -116,5 +116,10 @@
     "config": {
         "sort-packages": true,
         "process-timeout": 900
+    },
+    "extra": {
+        "branch-alias": {
+            "dev-4.next": "4.1.x-dev"
+        }
     }
 }


dereuromark:
The same is true for normal inline aliasing. It works and is as valid to be used as branch aliasing.
Saying anything different is wrong.
The warning is meant for the production use cases, not for trying out a branch.
Branch aliasing is as good or bad here to be honest.

dereuromark:
That said: 
I think its fair to consider including branch aliasing.
The only downside I could think of is merge conflicts down the line. But those wouldnt be too frequent anyway.

ADmad:
Closing as the related bug in composer v2 has been fixed.

GrahamCampbell:
As I say, expecting your users to use aliasing in their projects is not the right way to go. It is only intended as a temporary hack, as per the composer docs. The correct way to go is branch aliasing in this repo itself, or just name the branch properly.

swiffer:
I'm just needing it as a temporary hack (until RC2 is released). All fine :)

PR #14697, opened 2020-06-11
Deprecate whitelist in Paginator settings

markstory:
Replace the `whitelist` and `sortWhitelist` options with more appropriate and descriptive names. The added methods ensure that the new and old names behave as aliases for both configuration set via setConfig() and settings passed to paginate().

### DIFF
--- a/src/Datasource/Paginator.php
+++ b/src/Datasource/Paginator.php
@@ -36,7 +36,7 @@ class Paginator implements PaginatorInterface
      * - `maxLimit` - The maximum limit users can choose to view. Defaults to 100
      * - `limit` - The initial number of items per page. Defaults to 20.
      * - `page` - The starting page, defaults to 1.
-     * - `whitelist` - A list of parameters users are allowed to set using request
+     * - `allowedParameters` - A list of parameters users are allowed to set using request
      *   parameters. Modifying this list will allow users to have more influence
      *   over pagination, be careful with what you permit.
      *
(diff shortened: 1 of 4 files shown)


src/Datasource/Paginator.php
```diff
@@ -99,19 +99,19 @@ class Paginator implements PaginatorInterface
...
-     * sorting on. You can define the whitelist in the `$settings` parameter:
+     * you will need to define an allowed list of all the columns you wish to allow
+     * sorting on. You can define the allowed sort fields in the `$settings` parameter:
      *
      * ```
      * $settings = [
      *   'Articles' => [
      *     'finder' => 'custom',
-     *     'sortWhitelist' => ['title', 'author_id', 'comment_count'],
+     *     'allowedSort' => ['title', 'author_id', 'comment_count'],
```
othercorey, line 109:
```suggestion
     *     'allowedSortFields' => ['title', 'author_id', 'comment_count'],
```
If we're going to make it clearer, should we specify that they are fields or are they allowed to be expressions?

markstory: This list has to be strings as it is for checking the URL parameters for safeness.

garas:
```suggestion
     *     'sortableFields' => ['title', 'author_id', 'comment_count'],
```

`allowedSort` doesn't sound naturally

ADmad: I too like `sortableFields`.

markstory: I'll get this updated to sortableFields :+1:

markstory: Done now.

ADmad (approved):

PR #14719, opened 2020-06-18
Remove code duplication.

ADmad:

### DIFF
--- a/src/Database/Driver/Mysql.php
+++ b/src/Database/Driver/Mysql.php
@@ -62,13 +62,6 @@ class Mysql extends Driver
     protected $_schemaDialect;
 
     /**
-     * The server version
-     *
-     * @var string
-     */
-    protected $_version;
-
-    /**
      * Whether or not the server supports native JSON
      *
      * @var bool|null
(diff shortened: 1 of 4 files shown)


othercorey (approved):

src/Database/Driver/Mysql.php
```diff
@@ -230,21 +223,6 @@ public function supportsDynamicConstraints(): bool
         return true;
     }
 
-    /**
-     * Returns connected server version.
-     *
-     * @return string
-     */
-    public function getVersion(): string
```
markstory, line 238: Is this a new method? Or are we breaking compatibility here?

othercorey: I split some internal version checks into a `getVersion()` helper in this commit https://github.com/cakephp/cakephp/commit/57848d731c87c03e04fa11f53810f961da3e8731 for 4.1.

ADmad: @markstory It's a new method, so there is no BC break. I went with `version()` instead of `getVersion()` as `Sqlserver` class already has it and there isn't going to be a setter for version so the "get" prefix isn't really necessary.

PR #14721, opened 2020-06-18
Rollback transaction on any Throwable

### LINKED ISSUE #14717: 3.x Connection::transactional doesn't catch Throwable
This is a (multiple allowed):

* [x] bug
* [ ] enhancement

* CakePHP Version: 3.8.12

If an Error occurs in a transaction, rollback isn't called. This is because Exception is caught instead of Throwable.

**3.x** implementation:
```php
    public function transactional(callable $transaction)
    {
        $this->begin();

        try {
            $result = $transaction($this);
        } catch (Exception $e) {
            $this->rollback(false);
            throw $e;
        }
```

**4.x** implementation:
```php
    public function transactional(callable $transaction)
    {
        $this->begin();

        try {
            $result = $transaction($this);
        } catch (Throwable $e) {
            $this->rollback(false);
            throw $e;
        }
```

notpron:
Fixes #14717 

Whenever an Error is thrown in a transaction, the expectation is that this will be caught and the transaction is rolled back. For PHP 5.6 compatibility reasons, only Exception was caught in `Connection::transactional`. With PHP 7+ all Throwables should be handled also.

This change catches Throwable in addition to Exception, similar to (for instance) ErrorHandlerMiddleware handles the same. This is to ensure PHP backwards compatibility.

### DIFF
--- a/src/Database/Connection.php
+++ b/src/Database/Connection.php
@@ -29,6 +29,7 @@ use Cake\Database\Schema\Collection as SchemaCollection;
 use Cake\Datasource\ConnectionInterface;
 use Cake\Log\Log;
 use Exception;
+use Throwable;
 
 /**
  * Represents a connection with a database server.
@@ -735,6 +736,9 @@ class Connection implements ConnectionInterface
 
         try {
             $result = $transaction($this);
+        } catch (Throwable $e) {
+            $this->rollback(false);
+            throw $e;
         } catch (Exception $e) {
             $this->rollback(false);
             throw $e;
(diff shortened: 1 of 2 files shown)


src/Database/Connection.php
```diff
@@ -735,6 +736,9 @@ public function transactional(callable $transaction)
 
         try {
             $result = $transaction($this);
+        } catch (Throwable $e) {
```
othercorey, line 739: If we catch `Throwable` then don't need to catch `Exception` below, right?

ADmad: We do, since `Throwable` doesn't exist in PHP 5.6 (this PR is for 3.x).

dereuromark: We do. Since this version needs PHP5 support.
See the link/example I posted in the ticket.
There are like 5-10 other examples btw that might also need a look in terms of this issue, and if it could be relevant to also apply this exception handling there.

othercorey: Got it. I missed it was for 3.x.

dereuromark:
We now need to put this into 3.x as bugfix
cc @markstory

PR #14726, opened 2020-06-19
Update feature checks for MariaDB.

ADmad:

### DIFF
--- a/src/Database/Driver/Mysql.php
+++ b/src/Database/Driver/Mysql.php
@@ -90,6 +90,34 @@ class Mysql extends Driver
     protected $_endQuote = '`';
 
     /**
+     * Server type.
+     *
+     * If the underlying server is MariaDB, it's value will get set to `'mariadb'`
+     * after `version()` method is called.
+     *
+     * @var string
+     */
+    protected $serverType = 'mysql';
+
+    /**
+     * Mappping of feature to db server version for feature availability checks.
+     *
+     * @var array
+     */
+    protected $featuresToVersionMap = [
+        'mysql' => [
+            'json' => '5.7.0',
+            'cte' => '8.0.0',
+            'window' => '8.0.0',
+        ],
+        'mariadb' => [
+            'json' => '10.2.7',
+            'cte' => '10.2.1',
+            'window' => '10.2.0',
+        ],
+    ];
+
+    /**
      * Establishes a connection to the database server
      *
      * @return bool true on success
@@ -224,6 +252,27 @@ class Mysql extends Driver
     }
 
     /**
+     * Returns connected server version.
+     *
+     * @return string
+     */
+    public function version(): string
+    {
+        if ($this->_version === null) {
+            $this->connect();
+            $this->_version = (string)$this->_connection->getAttribute(PDO::ATTR_SERVER_VERSION);
+
+            if (strpos($this->_version, 'MariaDB') !== false) {
+                $this->serverType = 'mariadb';
+                $parts = explode(':', $this->_version);
+                $this->_version = array_pop($parts);
+            }
+        }
+
+        return $this->_version;
+    }
+
+    /**
      * Returns true if the server supports common table expressions.
      *
      * @return bool
@@ -231,7 +280,11 @@ class Mysql extends Driver
     public function supportsCTEs(): bool
     {
         if ($this->supportsCTEs === null) {
-            $this->supportsCTEs = version_compare($this->version(), '8.0.0', '>=');
+            $this->supportsCTEs = version_compare(
+                $this->version(),
+                $this->featuresToVersionMap[$this->serverType]['cte'],
+                '>='
+            );
         }
 
         return $this->supportsCTEs;
@@ -245,7 +298,11 @@ class Mysql extends Driver
     public function supportsNativeJson(): bool
     {
         if ($this->_supportsNativeJson === null) {
-            $this->_supportsNativeJson = version_compare($this->version(), '5.7.0', '>=');
+            $this->_supportsNativeJson = version_compare(
+                $this->version(),
+                $this->featuresToVersionMap[$this->serverType]['json'],
+                '>='
+            );
         }
 
         return $this->_supportsNativeJson;
@@ -259,7 +316,11 @@ class Mysql extends Driver
     public function supportsWindowFunctions(): bool
     {
         if ($this->_supportsWindowFunctions === null) {
-            $this->_supportsWindowFunctions = version_compare($this->version(), '8.0.0', '>=');
+            $this->_supportsWindowFunctions = version_compare(
+                $this->version(),
+                $this->featuresToVersionMap[$this->serverType]['window'],
+                '>='
+            );
         }
 
         return $this->_supportsWindowFunctions;
--- a/src/Database/Driver/Mysql.php
+++ b/src/Database/Driver/Mysql.php
@@ -290,8 +290,8 @@ class Mysql extends Driver
 
             if (strpos($this->_version, 'MariaDB') !== false) {
                 $this->serverType = static::SERVER_TYPE_MARIADB;
-                $parts = explode(':', $this->_version);
-                $this->_version = array_pop($parts);
+                preg_match('/^(?:5\.5\.5-)?(\d+\.\d+\.\d+.*-MariaDB[^:]*)/', $this->_version, $matches);
+                $this->_version = $matches[1];
             }
         }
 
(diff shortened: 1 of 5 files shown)


ADmad:
Working on this made me wonder if having separate function for each feature check is necessary. The code is practically the same for each. Instead we could have a `Driver::supports(string $feature): bool` function and add constants for various features.

othercorey:
> Instead we could have a `Driver::supports(string $feature): bool` function and add constants for various features.

One problem with that method is users won't necessarily know they're missing a feature check like they would if implementing an interface.

othercorey:
I don't think mariadb supports CTE's with UPDATE statements. I can't find specific documentation on what they support -- the documentation is very limited. 

The errors don't match what MySQL used to throw in this situation, but mariadb might early out in the parsing phase.

ADmad:
> One problem with that method is users won't necessarily know they're missing a feature check like they would if implementing an interface.

Realistically I don't see any else implementing a new driver besides the core team.

> I don't think mariadb supports CTE's with UPDATE statements. I can't find specific documentation on what they support -- the documentation is very limited.

I guess we'll just have to leave to the users to not not features that are not supported. Or perhaps update the feature checks in future when we can get more info.

othercorey:
> Realistically I don't see any else implementing a new driver besides the core team.

Agreed, but we also don't allow changing DriverInterface for years :disappointed:

src/Database/Driver/Mysql.php
```diff
@@ -223,6 +251,39 @@ public function supportsDynamicConstraints(): bool
...
+    public function version(): string
+    {
+        if ($this->_version === null) {
+            $this->connect();
+            $this->_version = (string)$this->_connection->getAttribute(PDO::ATTR_SERVER_VERSION);
+
+            if (strpos($this->_version, 'MariaDB') !== false) {
+                $this->serverType = 'mariadb';
+                $parts = explode(':', $this->_version);
+                $this->_version = array_pop($parts);
```
ndm2, line 280: For what kind of version string format is this intended?

From my experience the MariaDB version string usually looks something like `5.5.5-10.3.23-MariaDB` sometimes with additional trailing information like `-0ubuntu0.16.04.1`.

tests/TestCase/Database/QueryTests/CommonTableExpressionQueryTests.php
```diff
@@ -323,6 +323,11 @@ public function testWithInInsertWithValuesQuery()
      */
     public function testWithInUpdateQuery()
     {
+        $this->skipIf(
+            $this->connection->getDriver() instanceof Mysql && $this->connection->getDriver()->isMariadb(),
+            'MariaDB does not support window functions in UPDATE query.'
```
othercorey, line 328:
```suggestion
            'MariaDB does not support CTEs in UPDATE query.'
```

tests/TestCase/Database/QueryTests/CommonTableExpressionQueryTests.php
```diff
@@ -384,6 +389,11 @@ public function testWithInUpdateQuery()
      */
     public function testWithInDeleteQuery()
     {
+        $this->skipIf(
+            $this->connection->getDriver() instanceof Mysql && $this->connection->getDriver()->isMariadb(),
+            'MariaDB does not support window functions in DELETE query.'
```
othercorey, line 394:
```suggestion
            'MariaDB does not support CTEs in DELETE query.'
```

tests/TestCase/Database/Schema/MysqlSchemaTest.php
```diff
@@ -406,6 +406,12 @@ public function testDescribeTable()
                 'comment' => null,
             ],
         ];
+
+        if (ConnectionManager::get('test')->getDriver()->isMariadb()) {
+            $expected['created_with_precision']['default'] = 'current_timestamp(3)';
+            $expected['created_with_precision']['default'] = '';
```
othercorey, line 412:
```suggestion
            $expected['created_with_precision']['comment'] = '';
```

src/Database/Driver/Mysql.php
ADmad: Testing on Debian 9 I got string like `5.5.5-10.3.23-MariaDB-1:10.3.23+maria~stretch`.

I guess we can update the check to account for `5.5.5-10.3.23-MariaDB` too.

ADmad: @ndm2 I added a `debug()` to show the version and on the Ubuntu 18.04 VM also it shows similar format `5.5.5-10.4.13-MariaDB-1:10.4.13+maria~focal`.

ADmad: Sorry the server used isn't one provided through `apt` on Ubuntu 18.04 but a dockerized on provided by the `getong/mariadb-action` action, though the point remains the same.

ADmad: On Debian 9 

|`$pdo->getAttribute(PDO::ATTR_SERVER_VERSION)` |`SELECT VERSION()`                       |
|-----------------------------------------------|-----------------------------------------|
|`5.5.5-10.3.23-MariaDB-1:10.3.23+maria~stretch`|`10.3.23-MariaDB-1:10.3.23+maria~stretch`|

ndm2: I can check it again later, but from my testing when I implemented the CTE stuff, my local Ubuntu VM produced the `...-MariaDB-0ubuntu0.16.04.1` variant, and on Windows it was just `...-MariaDB`.

ADmad: I have updated the code to use regex match to account for both possible formats.

garas: `5.5.5-` is hardcoded and always returned by MariaDB

https://github.com/MariaDB/server/commit/c50ee6c23dbeb090963580754bec2f0a96ac0557

https://github.com/php/php-src/commit/f9f4a68368406a2ba2bd9fab0494261090183e8e

ADmad: @garas Yup. The regex used should account for all known possible strings format mentioned in the comments above.

ADmad:
@othercorey There are lot of `supportsFoo()` methods and I don't feel like putting in the effort to replaced all to use the `supports()` I proposed, so I'll just let them be for now.

othercorey:
> @othercorey There are a lot of `supportsFoo()` methods and I don't feel like putting in the effort to replace all to use the `supports()` I proposed insted, so I'll just let them be for now.

I'm not against string-based features. It's actually a cleaner api and doesn't require us to modify the interface. I just wanted to point out that features can be hidden.

src/Database/Driver/Mysql.php
```diff
@@ -290,8 +290,8 @@ public function version(): string
 
             if (strpos($this->_version, 'MariaDB') !== false) {
                 $this->serverType = static::SERVER_TYPE_MARIADB;
-                $parts = explode(':', $this->_version);
-                $this->_version = array_pop($parts);
+                preg_match('/^(?:5\.5\.5-)?(\d+\.\d+\.\d+.*-MariaDB[^:]*)/', $this->_version, $matches);
```
othercorey, line 293: This will be difficult to test. Users will just have to report issues with specific builds.

ADmad: We can update it as and when needed if users report a string which isn't covered.

othercorey (approved):

markstory:
> I guess we'll just have to leave it to the users to not use features that are not supported. Or perhaps update the feature checks in future when we can get more info.

Won't invalid SQL errors from the database help users figure out they are asking questions that MariaDB can't answer?

src/Database/Driver/Mysql.php
```diff
@@ -89,6 +103,34 @@ class Mysql extends Driver
      */
     protected $_endQuote = '`';
 
+    /**
+     * Server type.
+     *
+     * If the underlying server is MariaDB, it's value will get set to `'mariadb'`
```
markstory, line 109:
```suggestion
     * If the underlying server is MariaDB, its value will get set to `'mariadb'`
```

src/Database/Driver/Mysql.php
```diff
@@ -89,6 +103,34 @@ class Mysql extends Driver
...
+     * after `version()` method is called.
+     *
+     * @var string
+     */
+    protected $serverType = self::SERVER_TYPE_MYSQL;
+
+    /**
+     * Mappping of feature to db server version for feature availability checks.
```
markstory, line 117:
```suggestion
     * Mapping of feature to db server version for feature availability checks.
```

src/Database/Driver/Mysql.php
markstory: Should we have a testable way to access this so that we can add build names to the test suite as we collect them? Without that we'll have to hope that adding new names won't break the old ones.

ADmad: I have added tests for `version()` method now.

PR #14731, opened 2020-06-22
Avoid use of call_user_func() and call_user_func_array().

ADmad:
These methods are not very efficient.
See https://github.com/fab2s/call_user_func

### DIFF
--- a/tests/TestCase/ORM/Behavior/TranslateBehaviorTest.php
+++ b/tests/TestCase/ORM/Behavior/TranslateBehaviorTest.php
@@ -894,7 +894,7 @@ class TranslateBehaviorTest extends TestCase
         $this->assertSame('New translated article', $article->get('title'));
         $this->assertSame('Content #1', $article->get('body'));
 
-        $table->setLocale(false);
+        $table->setLocale(null);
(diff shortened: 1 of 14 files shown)


tests/TestCase/ORM/Behavior/TranslateBehaviorTest.php
```diff
@@ -894,7 +894,7 @@ public function testUpdateSingleLocale()
         $this->assertSame('New translated article', $article->get('title'));
         $this->assertSame('Content #1', $article->get('body'));
 
-        $table->setLocale(false);
+        $table->setLocale(null);
```
othercorey, line 897: Did changing the calls trigger argument validation somewhere?

othercorey (approved):

tests/TestCase/ORM/Behavior/TranslateBehaviorTest.php
ADmad: Yup, the argument for `TranslateStrategyInterface::setLocale()` is typehinted as `?string`. I don't know why TypeError was not thrown earlier.

PR #14737, opened 2020-06-22
Fix path docs

dereuromark:
https://github.com/cakephp/app/tree/master/templates

- layout
- element

and consolidate paths

### DIFF
--- a/src/Controller/Component/FlashComponent.php
+++ b/src/Controller/Component/FlashComponent.php
@@ -116,7 +116,7 @@ class FlashComponent extends Component
      * Magic method for verbose flash methods based on element names.
      *
      * For example: $this->Flash->success('My message') would use the
-     * `success.php` element under `templates/Element/Flash` for rendering the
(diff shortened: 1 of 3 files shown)


othercorey:
We use \*\*bold\*\* in the book docs for paths. Should we do the same here? Users won't always see it rendered as bold.

dereuromark:
I dont think we should use ** inside docblocks, the code fencing is especially for this case (and was also already in a lot of other places).
We could instead render those code blocks in the API docs if we want, as <code></code> etc.

markstory:
> We use **bold** in the book docs for paths. Should we do the same here? Users won't always see it rendered as bold.

We've been using inline code for paths in the API, which is inconsistent with the docs. We could change the API docs, to also use bold formatting, but I'm not sure the API docs and book need to be consistent.

ADmad:
I too don't see the need to use bold for paths.

PR #14739, opened 2020-06-23
Remove the GITHUB_USER variable as it is unused

markstory:
I long ago updated the scripts to only use GITHUB_TOKEN so having a USER set is no longer necessary.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,6 @@
 # The following env variables need to be set:
 # - VERSION
-# - GITHUB_USER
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -17,8 +16,6 @@ REMOTE=origin
 
 ifdef GITHUB_TOKEN
```
othercorey, line 17: Should this be an ifdef or should we make it required?

markstory: The various tasks that need a token use `guard-GITHUB_TOKEN` as a dependency which will cause a fail if the variable is undefined.

PR #14740, opened 2020-06-23
Abort atomic save if deletion of junction record fails.

ADmad:

### DIFF
--- a/src/ORM/Association/BelongsToMany.php
+++ b/src/ORM/Association/BelongsToMany.php
@@ -1195,6 +1195,10 @@ class BelongsToMany extends Association
                 $jointEntities = $this->_collectJointEntities($sourceEntity, $targetEntities);
                 $inserts = $this->_diffLinks($existing, $jointEntities, $targetEntities, $options);
 
+                if ($inserts === false) {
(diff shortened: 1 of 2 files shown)


othercorey (approved):

src/ORM/Association/BelongsToMany.php
```diff
@@ -1228,14 +1232,14 @@ function () use ($sourceEntity, $targetEntities, $primaryValue, $options) {
...
-     * @return array
+     * @return array|false Array of entities not deleted or false in case of deletion failure for atomic saves.
      */
     protected function _diffLinks(
         Query $existing,
         array $jointEntities,
         array $targetEntities,
         array $options = []
-    ): array {
+    ) {
```
dereuromark, line 1242: Why not ?array
then the null would indicate that and we can keep the typehint in place.

ADmad: In pretty much all ORM methods save failure is indicated by returning false. So I prefer being consistent here. Since it's an internal method the lack of type hint won't really matter, plus static analyzers would catch any type errors based on docblock types.

PR #14743, opened 2020-06-25
Improve doc block parameters

bancer:
Fixes phpstan error:
```
 ------ -------------------------------------------------------------------- 
  1337   Parameter #1 $table of method Cake\Database\Query::innerJoin()      
         expects array<string>|string, array<string, Cake\ORM\Query> given.  
 ------ --------------------------------------------------------------------
```

### DIFF
--- a/src/Database/Query.php
+++ b/src/Database/Query.php
@@ -715,7 +715,7 @@ class Query implements ExpressionInterface, IteratorAggregate
      * The arguments of this method are identical to the `leftJoin()` shorthand, please refer
      * to that methods description for further details.
      *
-     * @param string|string[] $table The table to join with
+     * @param string|(string|\Cake\ORM\Query)[] $table The table to join with
      * @param string|array|\Cake\Database\ExpressionInterface $conditions The conditions
      * to use for joining.
      * @param array $types a list of types associated to the conditions used for converting
@@ -730,7 +730,7 @@ class Query implements ExpressionInterface, IteratorAggregate
     /**
      * Returns an array that can be passed to the join method describing a single join clause
      *
-     * @param string|string[] $table The table to join with
+     * @param string|(string|\Cake\ORM\Query)[] $table The table to join with
      * @param string|array|\Cake\Database\ExpressionInterface $conditions The conditions
      * to use for joining.
      * @param string $type the join type to use
--- a/src/Database/Query.php
+++ b/src/Database/Query.php
@@ -675,7 +675,7 @@ class Query implements ExpressionInterface, IteratorAggregate
      *
      * See `join()` for further details on conditions and types.
      *
-     * @param string|string[] $table The table to join with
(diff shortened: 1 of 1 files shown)


src/Database/Query.php
```diff
@@ -715,7 +715,7 @@ public function rightJoin($table, $conditions = [], $types = [])
      * The arguments of this method are identical to the `leftJoin()` shorthand, please refer
      * to that methods description for further details.
      *
-     * @param string|string[] $table The table to join with
+     * @param string|(string|\Cake\ORM\Query)[] $table The table to join with
```
ndm2, line 718: That syntax isn't supported in most IDEs, even the (not in PSR-5 anymore) `Type<...>` generic syntax has barely been adopted.

```suggestion
     * @param string|string[]|\Cake\Database\ExpressionInterface[] $table The table to join with
```

ps. don't forget `leftJoin()` and `_makeJoin()`.

dereuromark: The proposed syntax can go into psalm-param tag on top if it provides additional info.

bancer: Logically `string[]|\Cake\Database\ExpressionInterface[]` is wrong because the array may contain both strings and objects. It is not only array of strings or array of objects.

bancer: @dereuromark, could you elaborate or give an example?

bancer: @ndm2 https://github.com/php-fig/fig-standards/blob/master/proposed/phpdoc.md#arrays

dereuromark: @bancer Quoting back other resources that not make it any more valid.
We explained why you need to use `psalm-param` syntax on top for now for this syntax. You can find other pieces in the code here as reference already.

bancer: @dereuromark, well @ndm2 stated that `(string|\Cake\ORM\Query)[]` syntax is not in PSR-5 anymore that is not true.

ndm2: I stated that the `Type<..>` syntax (`array<string|ExpressionInterface>`) is not in PSR-5 anymore, even though it had at least _some_ adoption.

dereuromark: It doesn't matter if it is in or out. The facts remain: We cannot use it in tags yet, and thus we use the special tags on top.
The minor difference in meaning is therefore acceptable if that means we do not lose major IDE support here.
Please understand that this is why those special tags exist in the first place.

ndm2: The meaning is actually perfectly valid in this case, as `_makeJoin()` expects / makes use of exactly one array entry.

othercorey:
Changed to 3.x branch. The 3.next branch was merged into 3.x for 3.9 release.

othercorey:
@bancer If you update this PR, we should be able to merge it.

Issue #14722, opened 2020-06-18
3.x Http Client no ssl_local_pk option


ns-group-dev:
This is a (multiple allowed):

* [x] enhancement

* CakePHP Version: 3.4.14.

### What you did
Working with a third party API that requires a certificate they signed and gave me, I found that the only solution to succeed the requests was to use a stream context (for fopen) with local_cert (pointing to that .cer certificate) and local_pk (pointing to my private key). For some reason using the cafile didn't work (don't know why). 
I tried to reproduce this with the Http Client. I saw the ssl_local_cert option in documentation, great, so I naively put also a ssl_local_pk parameter to the Client...

### What happened
I was returned an error about the missing private key (the ssl_local_cert was ok). Searching into the Cake Http Client code, I found no ssl option for "local_pk", which is a common parameter in stream context?

For the moment I put the option manually to the ssl options (/src/Http/Client/Adapter/Stream.php, _buildSslContext() method) : 
```
$sslOptions = [
            'ssl_verify_peer',
            'ssl_verify_peer_name',
            'ssl_verify_depth',
            'ssl_allow_self_signed',
            'ssl_cafile',
            'ssl_local_cert',
            'ssl_local_pk', // here
            'ssl_passphrase',
        ];
```
Why is this parameter missing? Was it an intention?
Wouldn't it be a good thing to add it as a PR?

markstory:
> Why is this parameter missing? Was it an intention?

I was not aware of that option and have never used it or seen it used.

> Wouldn't it be a good thing to add it as a PR?

A pr to add that option would be awesome.

PR #14756, opened 2020-06-29
Added ssl_local_pk option for Http/Client

ns-group-dev:
Ticket #14722

Just added an "ssl_local_pk" to the ssl context options for the Http Client, to use in conjunction with the "ssl_local_cert" option when parameting it.

Like :
```
new \Cake\Http\Client([
    'ssl_local_cert' => 'path/to/local.cer',
    'ssl_local_pk' => 'path/to/local.key'
]);
```

The ticket was tagged with the 3.next version, but @ADmad suggested me to make a PR on 4.next version.

### DIFF
--- a/src/Http/Client/Adapter/Stream.php
+++ b/src/Http/Client/Adapter/Stream.php
@@ -217,6 +217,7 @@ class Stream implements AdapterInterface
             'ssl_allow_self_signed',
             'ssl_cafile',
             'ssl_local_cert',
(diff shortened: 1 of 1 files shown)


markstory:
> The ticket was tagged with the 3.next version, but @ADmad suggested me to make a PR on 4.next version.

Yes, this change can be backported to 3.x for use in 3.9.

PR #14763, opened 2020-07-03
Changed CollectionTrait::shuffle() to return all elements in collection.

### LINKED ISSUE #14757: Collection : Append and shuffle problem
This is a (multiple allowed):

* [x] bug

* CakePHP Version: 3.7.9
* Platform and Target: apache2, mysql on ubuntu 18

### What you did
I was just making two requests and append the two results Set :
`$Trouvables = $this->Lesobjets->find('Region',['region_id' => $options['position']['region']['id']])->find('action', ['action' => $options['action']])->distinct('Lesobjets.id');`
(gives 12 results)
`$TrouvablesAussi = $this->Lesobjets->find('NoRegion')->find('action', ['action' => $options['action']])->distinct('Lesobjets.id');`
(gives 1 result)
Then I try to pick a random element : 
`$laListe = $Trouvables->append($TrouvablesAussi)->shuffle();`
### What happened
In fact,  after applying shuffle(), the resulting set contain only 12 objects.

There was 13 objects before applying shuffle() so I expected to obtain 13 objects after :)

After a close examination of code, It seems that this is because shuffle() use a toArray() transformation on resultSet(), instead of toList(). As the first collection has object numbered from 0 to 11, appending another collection with 1 object at key 0 leads to overwrite of the first one by toArray().

othercorey:
Fixes https://github.com/cakephp/cakephp/issues/14757

### DIFF
--- a/src/Collection/CollectionTrait.php
+++ b/src/Collection/CollectionTrait.php
@@ -330,7 +330,7 @@ trait CollectionTrait
      */
     public function shuffle(): CollectionInterface
     {
-        $elements = $this->toArray();
+        $elements = $this->toList();
         shuffle($elements);
 
         return $this->newCollection($elements);
--- a/tests/TestCase/Collection/CollectionTest.php
+++ b/tests/TestCase/Collection/CollectionTest.php
@@ -925,6 +925,20 @@ class CollectionTest extends TestCase
     }
 
     /**
+     * Tests shuffle with duplicate keys.
+     *
+     * @return void
+     */
+    public function testShuffleDuplicateKeys()
+    {
+        $collection = (new Collection(['a' => 1]))->append(['a' => 2])->shuffle();
+        $result = $collection->toArray();
+        $this->assertCount(2, $result);
+        $this->assertContainsEquals(1, $result);
+        $this->assertContainsEquals(2, $result);
+    }
+
+    /**
      * Tests sample
      *
      * @dataProvider simpleProvider
--- a/tests/TestCase/Collection/CollectionTest.php
+++ b/tests/TestCase/Collection/CollectionTest.php
@@ -934,6 +934,7 @@ class CollectionTest extends TestCase
         $collection = (new Collection(['a' => 1]))->append(['a' => 2])->shuffle();
         $result = $collection->toArray();
         $this->assertCount(2, $result);
+        $this->assertEquals([0, 1], array_keys($result));
         $this->assertContainsEquals(1, $result);
         $this->assertContainsEquals(2, $result);
     }
--- a/src/Collection/CollectionInterface.php
+++ b/src/Collection/CollectionInterface.php
@@ -789,7 +789,8 @@ interface CollectionInterface extends Iterator, JsonSerializable
     public function toList(): array;
 
     /**
-     * Convert a result set into JSON.
+     * Returns the data that can be converted to JSON. This returns the same data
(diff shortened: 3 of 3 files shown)


othercorey:
`gethostbynamel()` is returning 2 addresses for 'localhost' now:

```
1) Cake\Test\TestCase\Network\SocketTest::testAddresses
Failed asserting that two arrays are identical.
--- Expected
+++ Actual
@@ @@
 Array &0 (
     0 => '127.0.0.1'
+    1 => '127.0.0.1'
 )
```

src/Collection/CollectionTrait.php
```diff
@@ -330,7 +330,7 @@ public function sumOf($matcher = null)
      */
     public function shuffle(): CollectionInterface
     {
-        $elements = $this->toArray();
+        $elements = $this->toList();
```
ADmad, line 333: While this is for a bug fix, it's a change in behavior that could surprise people, so safer to target 4.1 and document in migration guide.

ADmad:
Here's an alternate fix:

```diff
diff --git a/src/Collection/CollectionTrait.php b/src/Collection/CollectionTrait.php
index 1a7690c652..12e113da83 100644
--- a/src/Collection/CollectionTrait.php
+++ b/src/Collection/CollectionTrait.php
@@ -330,7 +330,12 @@ trait CollectionTrait
      */
     public function shuffle(): CollectionInterface
     {
-        $elements = $this->toArray();
+        $preserveKeys = true;
+        $iterator = $this->unwrap();
+        if (get_class($iterator) === AppendIterator::class) {
+            $preserveKeys = false;
+        }
+        $elements = $this->toArray($preserveKeys);
         shuffle($elements);
 
         return $this->newCollection($elements);
```

othercorey:
The shuffle will reset keys regardless.

src/Collection/CollectionTrait.php
othercorey: I added it to 4.0.x in case we wanted to backport to 3.9.x. Up to you though.

markstory: Dropping the keys is both good and bad. It will fix one but and potentially create another for existing users. One benefit of this change is it makes the behavior consistent with PHP's `shuffle()` method which also drops keys.

tests/TestCase/Collection/CollectionTest.php
```diff
@@ -924,6 +924,20 @@ public function testShuffle($data)
...
+     *
+     * @return void
+     */
+    public function testShuffleDuplicateKeys()
+    {
+        $collection = (new Collection(['a' => 1]))->append(['a' => 2])->shuffle();
+        $result = $collection->toArray();
+        $this->assertCount(2, $result);
+        $this->assertContainsEquals(1, $result);
+        $this->assertContainsEquals(2, $result);
```
markstory, line 938: It might also be good to assert that there is no `a` key anymore.

src/Collection/CollectionTrait.php
othercorey: All I can say is this is the current documentation for shuffle(). To me, that suggests that you want all the elements. If users need to have unique keys only, then they can re-create the collection first.

```
     * Returns a new collection with the elements placed in a random order,
     * this function does not preserve the original keys in the collection.
```

ADmad:
> The shuffle will reset keys regardless.

In that case I rescind my alternative fix :)

Issue #14203, opened 2020-01-24
Disable select alias generation for ORM query used as join subquery


othercorey:
* [x] bug
* [x] enhancement
* [x] feature-discussion (RFC)

When an ORM query is used as a subquery for a join, the field aliases are still generated which requires the user to map new aliases for every field/column they want to use from the join.

```
LEFT JOIN (
    SELECT
      ClientSites.id AS "ClientSites__id"
    FROM client_sites ClientSites
    WHERE
      client_id = :c0
  ) Access ON Access.id = ClientSites.id
```

Would be nice to have a way to disable the alias generation for the query.

ADmad:
Don't we need to populate entities from the subquery when using subqueries for join? Without the aliases the ORM wouldn't be able to do so?

othercorey:
Does it work like that? I couldn't find any subquery tests for ORM queries.

markstory:
If we didn't generate aliases in subqueries we could break existing queries that people have built on the aliases existing.

othercorey:
It would have to be an option set if anything.

othercorey:
This would also be useful for CTE/with clauses.

othercorey:
Now that we have CTE/with, I'm going to try to get a PR for 4.1.

othercorey:
Implemented in https://github.com/cakephp/cakephp/pull/14764

PR #14764, opened 2020-07-03
Added Query::subquery() which disables field aliasing

othercorey:
Closes https://github.com/cakephp/cakephp/issues/14203.

This implementation creates a `ORM\Query::subquery()` wrapper that disables aliasing on a new Query instance. This style implies that the user needs to create subqueries for CTE or where/etc subqueries from an ORM\Query instance.

If users aren't building a subquery off an ORM query, then they probably want to use a regular `Database\Query` instance.


othercorey:
I wasn't sure where to put this.

`Database\Query::with()` uses `Connection::newQuery()` to create an instance for closures. That always creates a new `Database\Query` even when called from `ORM\Query`.

Adding `subquery()` to `Database\Query` doesn't add much because it's not really modifying any behavior in `Database\Query`. Users who need ORM Table queries with aliasing disable will have to create their own Query instance and not rely on closure arguments.

src/ORM/Query.php
```diff
@@ -1285,6 +1300,20 @@ public function insert(array $columns, array $types = [])
...
 
+    /**
+     * Returns a new Query that has automatic field aliasing disabled.
+     *
+     * @param \Cake\ORM\Table $table The table this query is starting on
+     * @return self
+     */
+    public function subquery(Table $table): self
+    {
+        $query = new self($this->getConnection(), $table);
```
ADmad, line 1311: `static` instead of `self` perhaps?

othercorey: `static` should be ok.

othercorey: On second thought, other code explicitly creates `Database\Query` so maybe this should explicitly create `ORM\Query` instances.

dereuromark: Still. If you ever needed to override on plugin or project level, it would not be easy to overcome this typehint then, would it? The static is just a bit more lenient here while it will provide the same functionality out of the box.

src/ORM/Query.php
```diff
@@ -1285,6 +1300,20 @@ public function insert(array $columns, array $types = [])
         return parent::insert($columns, $types);
     }
 
+    /**
+     * Returns a new Query that has automatic field aliasing disabled.
+     *
+     * @param \Cake\ORM\Table $table The table this query is starting on
+     * @return static
+     */
+    public function subquery(Table $table)
```
markstory, line 1309: How would you see this being used in window and CTE expressions?

othercorey: The goal is being able to use `ORM\Query` without the field aliasing. So, that lets me use the same style of query building as the outer query with where/innerJoinWith/etc that references associations.

markstory:
> Database\Query::with() uses Connection::newQuery() to create an instance for closures. That always creates a new Database\Query even when called from ORM\Query.

Should we change this? If `Orm\Query` re-implemented `with()` to pass instances of `Orm\Query` that have been converted to subqueries? This might make building with() queries more ergonomic as the ORM\Query interface is the one people are more familiar with.

Given that `Query::subquery()` is returning new instances it could be a static method, or perhaps the method should be `newSubquery()` and we can drop the Table parameter, and use the current query's repository?

Another approach would be to leave it as an instance method that traverses the various query parts and removes aliases. That would enable more natural usage like:

```php
$query->where(['id IN' => $otherQuery->where(...)->subquery()]);
```

othercorey:
> Another approach would be to leave it as an instance method that traverses the various query parts and removes aliases. That would enable more natural usage like:
> 
> ```
> $query->where(['id IN' => $otherQuery->where(...)->subquery()]);
> ```

How would we know which aliases to remove?

markstory:
All aliases in the query should be safe to remove if we also update conditions. Removing aliases after the subquery is built will be more expensive than turning off aliasing before hand. Perhaps we shouldn't pursue that path given the additional cost, complexity and potential for bugs during alias removal.

I think that the factory method might be a better approach then. Creating an instance only to create another query instance seems wasteful when we could go directly to the subquery instance.

othercorey:
> All aliases in the query should be safe to remove if we also update conditions. Removing aliases after the subquery is built will be more expensive than turning off aliasing before hand. Perhaps we shouldn't pursue that path given the additional cost, complexity and potential for bugs during alias removal.

But, that would remove aliases for user-specified fields as well, right?

markstory:
Yes stripping the aliases would remove all aliases. After thinking through it more, this would be fraught with peril as there could be aliases embedded in expression objects that we cannot modify.

othercorey:
> Given that Query::subquery() is returning new instances it could be a static method, or perhaps the method should be newSubquery() and we can drop the Table parameter, and use the current query's repository?

The sample implementation uses the connection from the current `Query` instance. If we make it a static helper, which connection should it use?

markstory:
> The sample implementation uses the connection from the current Query instance. If we make it a static helper, which connection should it use?

It could take a connection from the passed table? Something like `Query::subquery($table)`

othercorey:
Changed `ORM\Query::subquery()` to a static function and use the connection from `Table`.

src/ORM/Query.php
```diff
@@ -1285,6 +1300,20 @@ public function insert(array $columns, array $types = [])
...
 
+    /**
+     * Returns a new Query that has automatic field aliasing disabled.
+     *
+     * @param \Cake\ORM\Table $table The table this query is starting on
+     * @return static
+     */
+    public static function subquery(Table $table)
+    {
+        $query = new static($table->getConnection(), $table);
```
othercorey, line 1311: I think always using the `default` connection in `Table` is a design limitation that prevents read/write pool configurations. We should consider allowing read/write connection defaults in the future.

PR #14778, opened 2020-07-08
Deprecate method that throws deprecation along the way

dereuromark:
Follow https://github.com/cakephp/cakephp/pull/14777

### DIFF
--- a/src/View/Helper/NumberHelper.php
+++ b/src/View/Helper/NumberHelper.php
@@ -222,6 +222,7 @@ class NumberHelper extends Helper
      * if $currency argument is not provided. If boolean false is passed, it will clear the
      * currently stored value. Null reads the current default.
      * @return string|null Currency
+     * @deprecated 4.1.0 Use setDefaultCurrency()/getDefaultCurrency() instead.
      */
     public function defaultCurrency($currency): ?string
     {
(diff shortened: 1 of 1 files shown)


src/View/Helper/NumberHelper.php
```diff
@@ -222,6 +222,7 @@ public function formatDelta($value, array $options = []): string
      * if $currency argument is not provided. If boolean false is passed, it will clear the
      * currently stored value. Null reads the current default.
      * @return string|null Currency
+     * @deprecated 4.1.0 Use setDefaultCurrency()/getDefaultCurrency() instead.
```
saeideng, line 225:
```suggestion
     * @deprecated 3.9.0 Use setDefaultCurrency()/getDefaultCurrency() instead.
```

saeideng: if you check these methods exist in `4.0.0` too

dereuromark: Yes but we just started to declare, no?

saeideng: no , since 4.0.0
see https://github.com/cakephp/cakephp/blob/4.0.0/src/I18n/Number.php#L275-L278

PR #14787, opened 2020-07-11
Fix missing asset timestamps on prefixed plugins

### LINKED ISSUE #14784: Timestamps missing for Org/Plugin assets
* [x] bug
* [ ] enhancement

* CakePHP Version: 4.1.0, and 4.0.x
* Platform and Target: Ubuntu, PHP 7.2, Apache, MySQL

### What you did
When configuration for asset timestamps is enabled
```
'Asset' => [
    'timestamp' => true,
],
```
and including assets
```
$this->Html->script(
    [
        'MyPlugin.assetname',
        'Org/MyOtherPlugin.assetname'
    ],
    ['block' => 'extra-scripts']
);
$this->Html->css(
    [
        'MyPlugin.assetname',
        'Org/MyOtherPlugin.assetname'
    ],
    ['block' => 'css']
);
```

### What happened
HTML output only assigns timestamps to plugins that aren't org name-spaced
```
<link rel="stylesheet" href="/my_plugin/css/assetname.css?1583177993">
<link rel="stylesheet" href="/org/my_other_plugin/css/assetname.css">
...
<script src="/my_plugin/js/assetname.js?1583177993"></script>
<script src="/org/my_other_plugin/js/assetname.js"></script>
```

I expect both HTML outputs to contain the asset timestamp

### Initial findings

The code [here in Asset.php](https://github.com/cakephp/cakephp/blob/166e8f7b9d036880aebb8783db0a5f2b06f0f812/src/Routing/Asset.php#L256) only checks the first segment split by "/", so `Plugin::isLoaded()` is only checking against the org instead of the full plugin name.

Adding the following (right before the if statement) will correctly identify that the plugin has been loaded, but then the assets are expected to live in `<plugin-folder>/webroot/plugin_name/{css,js}/asset.ext` instead of `<plugin-folder>/webroot/{css,js}/asset.ext`, breaking a consistent file structure.
```
if (!Plugin::isLoaded($plugin) && count($segments) > 1) {
    $plugin = implode('/', [$plugin, Inflector::camelize($segments[1])]);
}
```

markstory:
Fixes #14784

### DIFF
--- a/src/Routing/Asset.php
+++ b/src/Routing/Asset.php
@@ -254,6 +254,10 @@ class Asset
             }
             $segments = explode('/', ltrim($filepath, '/'));
             $plugin = Inflector::camelize($segments[0]);
+            if (!Plugin::isLoaded($plugin) && count($segments) > 1) {
+                $plugin = implode('/', [$plugin, Inflector::camelize($segments[1])]);
+                unset($segments[1]);
+            }
             if (Plugin::isLoaded($plugin)) {
                 unset($segments[0]);
                 $pluginPath = Plugin::path($plugin)
--- a/src/Routing/Asset.php
+++ b/src/Routing/Asset.php
@@ -252,6 +252,7 @@ class Asset
             if (file_exists($webrootPath)) {
                 return $path . '?' . filemtime($webrootPath);
             }
+            // Check for plugins and org prefixed plugins.
             $segments = explode('/', ltrim($filepath, '/'));
             $plugin = Inflector::camelize($segments[0]);
             if (!Plugin::isLoaded($plugin) && count($segments) > 1) {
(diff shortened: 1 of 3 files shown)


othercorey:
php nightly is core dumping:

> /home/travis/.travis/functions: line 113:  5293 Segmentation fault      (core dumped) vendor/bin/phpunit

garas:
Segfault fixed in https://github.com/php/php-src/commit/1c967df5a037f50cc894f38765371b8ccb2249ff

But PHP 8 has now `match` keyword reserved, and PHPUnit 8 doesn't work, PHPUnit 9 fixed in https://github.com/sebastianbergmann/phpunit/commit/b57cd45d8d8626d5b9b38547b85094a8a7160712

src/Routing/Asset.php
```diff
@@ -254,6 +254,10 @@ public static function assetTimestamp(string $path, $timestamp = null): string
             }
             $segments = explode('/', ltrim($filepath, '/'));
             $plugin = Inflector::camelize($segments[0]);
+            if (!Plugin::isLoaded($plugin) && count($segments) > 1) {
```
othercorey, line 257: Can we comment was is happening in this section below the timestampEnabled check.

othercorey (approved):

Issue #14785, opened 2020-07-11
Route issues, template Debugger::exportVar formatted Html instead of Text


smarek:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 4.1 stable

In templates https://github.com/cakephp/cakephp/blob/master/templates/Error/missing_route.php#L52-L53 and https://github.com/cakephp/cakephp/blob/master/templates/Error/duplicate_named_route.php#L54-L55 exported route defaults and options get formatted by Debugger HtmlFormatter instead of TextFormatter, and `h()` will turn the resulting Html into escaped Html, making the output non-readable

I've not found quick way to change the current formatter in the call-chain, so i'm submitting bug report instead

garas:
It is probably better leave dump in HTML, but remove `h()`,

smarek:
@garas if you can keep the html expanded, why not, but in my case all the 2nd and 3rd level of information is collapsed, and i'ts impractical to have to click them all

PR #14789, opened 2020-07-12
Fix double encoding in route errors

markstory:
Fix the double encoding errors reported in #14785. I've also improved the route error layouts a bit and improved the styling of the dumps so that they are formatted the same as normal debug() dumps, and made the table flow better and be full width.

### DIFF
--- a/templates/Error/duplicate_named_route.php
+++ b/templates/Error/duplicate_named_route.php
@@ -38,17 +38,12 @@ Remove duplicate route names in your route configuration.</p>
     <h3>Duplicate Route</h3>
     <table cellspacing="0" cellpadding="0">
     <tr><th>Template</th><th>Defaults</th><th>Options</th></tr>
-    <?php
-    $other = $attributes['duplicate'];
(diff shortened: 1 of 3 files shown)


templates/Error/duplicate_named_route.php
```diff
@@ -34,28 +34,16 @@
...
-        h($other->template),
-        h(Debugger::exportVar($other->defaults)),
-        h(Debugger::exportVar($other->options))
-    );
-    echo '</tr>';
-    ?>
+    <?php $other = $attributes['duplicate']; ?>
+    <tr>
+        <td><?= h($other->template) ?></td>
+        <td><div class="cake-debug" data-open-all="true"><?= Debugger::exportVar($other->defaults) ?></div></td>
```
othercorey, line 44: Should we make this expanded an option in the exporter?

markstory: We could. I can add that to 4.next.

othercorey (approved):

PR #14790, opened 2020-07-13
Ensure Comparison alias is created

### LINKED ISSUE #14788: 4.1: ORM regression with callbacks
This is a (multiple allowed):

* [x] bug
* [ ] enhancement

* CakePHP Version: 4.1

### What you did
Travis Test Suite for Tool started to become red with 4.1 release (7 days ago)
https://travis-ci.com/github/dereuromark/cakephp-tools/builds

Mysql
> 1) Tools\Test\TestCase\Model\Behavior\BitmaskedBehaviorTest::testSaveBasic
PDOException: SQLSTATE[42S22]: Column not found: 1054 Unknown column 'statuses' in 'where clause'

Postgres
> 1) Tools\Test\TestCase\Model\Behavior\BitmaskedBehaviorTest::testSaveBasic
PDOException: SQLSTATE[42703]: Undefined column: 7 ERROR:  column "statuses" does not exist
LINE 1: ...OM "bitmasked_comments" "BitmaskedComments" WHERE "statuses"...
HINT:  Perhaps you meant to reference the column "BitmaskedComments.status".

And also Sqlite (but here it lets the unknown field slide and only fails on assert afterwards):
> 1) Tools\Test\TestCase\Model\Behavior\BitmaskedBehaviorTest::testSaveBasic
Failed asserting that false is true.

This seems to be not specific to DB, so the callback (logic) in CakePHP itself must have changed in a BC breaking way.

### What happened

The beforeFind() callback seems compromised here and doesn't exchange the array values with the int bitmask as it used to do in 3.x/4.0.

So the search query stays

        [protected] _conditions => [
                (int) 0 => object(Cake\Database\Expression\Comparison) {
                        [protected] _value => [
                                (int) 0 => (int) 2,
                                (int) 1 => (int) 4
                        ]
                        [protected] _type => 'string[]'
                        [protected] _operator => 'in'
                        [protected] _isMultiple => true
                        [protected] _valueExpressions => []
                        [protected] _field => 'BitmaskedComments.statuses'
                }
        ]

instead of expected `status` of bitmask value `6`.

I didn't find any relevant part in migration guide: https://book.cakephp.org/4/en/appendices/4-1-migration-guide.html

The weird part about this is that locally it passes...
I use PHP 7.3 locally, but on Travis it fails for all versions, so that cannot be it either.

othercorey:
Closes https://github.com/cakephp/cakephp/issues/14788

### DIFF
--- a/src/Database/Expression/Comparison.php
+++ b/src/Database/Expression/Comparison.php
@@ -2,3 +2,5 @@
 declare(strict_types=1);
 
 class_exists('Cake\Database\Expression\ComparisonExpression');
+
+deprecationWarning('`Comparison` deprecated in 4.1.0, use `ComparisonExpression` instead.');
--- a/src/Database/Expression/Comparison.php
+++ b/src/Database/Expression/Comparison.php
@@ -1,5 +1,4 @@
 <?php
 declare(strict_types=1);
 
-// @deprecated 4.1.0 Add backwards compatible alias.
-class_alias('Cake\Database\Expression\ComparisonExpression', 'Cake\Database\Expression\Comparison');
+class_exists('Cake\Database\Expression\ComparisonExpression');
--- a/src/Database/Expression/ComparisonExpression.php
+++ b/src/Database/Expression/ComparisonExpression.php
@@ -315,3 +315,9 @@ class ComparisonExpression implements ExpressionInterface, FieldInterface
         return [$result, $expressions];
     }
 }
+
+// phpcs:disable
+// Comparison will not load during instanceof checks so ensure it's loaded here
+// @deprecated 4.1.0 Add backwards compatible alias.
+class_alias('Cake\Database\Expression\ComparisonExpression', 'Cake\Database\Expression\Comparison');
+// phpcs:enable


src/Database/Expression/ComparisonExpression.php
```diff
@@ -315,3 +315,8 @@ protected function _collectExpressions($values): array
         return [$result, $expressions];
     }
 }
+
+// phpcs:disable
+// Comparison will not load during instanceof checks so ensure it's loaded here
```
garas, line 320: Move `@deprecated` here

src/Database/Expression/Comparison.php
```diff
@@ -1,5 +1,6 @@
 <?php
 declare(strict_types=1);
 
-// @deprecated 4.1.0 Add backwards compatible alias.
-class_alias('Cake\Database\Expression\ComparisonExpression', 'Cake\Database\Expression\Comparison');
+class_exists('Cake\Database\Expression\ComparisonExpression');
+
+deprecationWarning('`Comparison` deprecated in 4.1, use `ComparisonExpression` instead.');
```
dereuromark, line 6: Wont this be a bit too harsh at this point?

othercorey: Removed, but it's only for users who manually instantiate Comparison objects.

src/Database/Expression/ComparisonExpression.php
```diff
@@ -315,3 +315,9 @@ protected function _collectExpressions($values): array
...
+// @deprecated 4.1.0 Add backwards compatible alias.
```
ADmad, line 321: I am pretty sure IDEs and static analyzers won't warn about deprecated class usage unless the `@deprecated` tag is present in a docblock in `Comparion.php`.

othercorey: We never had them in class docblocks for the aliases. Does it warn if it's just in the file? What should I do?

garas: This could work [[source]](https://translate.google.com/translate?hl=en&sl=de&tl=en&u=https%3A%2F%2Fwww.schmengler-se.de%2F2016%2F09%2Fphp-class_alias-verwenden-um-klassen-rueckwaertskompatibel-zu-verschiebenumzubenennen%2F&prev=search):

```php
if (!class_exists(OldClass::class)) {
    /** @deprecated this is an alias for NewClass */
    class OldClass {}
}
```

othercorey: I feel like we could just add the deprecationWarning() to Comparison.php and the few users who manually create Comparison objects will get a warning.

dereuromark: OK, fair enough I guess.

othercorey: That doesn't really work either apparently. If we create the alias in ComparisonExpression then Comparison.php is never loaded when users instantiate Comparison objects. They will only see the deprecation warning if they create them before running any cakephp code that creates ComparisonExpression objects.

othercorey: I tried to create a unit test, but it's unreliable, so I'm just adding the warning for that edge case.

Issue #14747, opened 2020-06-26
TestCase::getMockForModel generate phpunit deprecations notices


CauanCabral:
This is a (multiple allowed):

* [x] bug
* [x] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 4.1-RC2.
* Platform and Target: phpunit 8.5.8, php 7.2

### What you did
Have test cases with model linked with behaviors and tried mock some of behavior methods.

### What happened
Get that message when tests are run:

`Trying to set mock method "myBehaviorMethod" with onlyMethods, but it does not exist in class "App\Model\Table\ArticlesTable". Use addMethods() for methods that don't exist in the class.`

### What you expected to happen
Mocked methods and silent run

I found a PR on PHPUNIT about that deprecation: https://github.com/sebastianbergmann/phpunit/pull/3687

So, inspecting `Cake\TestSuite\TestCase::getMockForModel` I can fix that message changing: 

```php
        $options += $locator->getConfig($alias);

        /** @var \Cake\ORM\Table $mock */
        $mock = $this->getMockBuilder($className)
            ->onlyMethods($methods)
            ->setConstructorArgs([$options])
            ->getMock();
```

to

```php
        $options += $locator->getConfig($alias);
        $existingMethods = array_intersect($methods, get_class_methods($className));
        $nonExistingMethods = array_diff($methods, $existingMethods);

        /** @var \Cake\ORM\Table $mock */
        $mock = $this->getMockBuilder($className)
            ->onlyMethods($existingMethods)
            ->addMethods($nonExistingMethods)
            ->setConstructorArgs([$options])
            ->getMock();
```

dereuromark:
Do you want to make a PR here?

CauanCabral:
Sure, I just don't know how to include a test for this (a test for the TestCase class?)

othercorey:
Fixed in https://github.com/cakephp/cakephp/pull/14748

hakito:
This solution shifts the deprecation warning to the case when trying to mock protected methods. get_class_methods returns public methods only.

Maybe ReflectorClass should be used instead?

CauanCabral:
You are right @hakito 

That way we have all methods reflected, in my codebase, worked to:

```php
        [, $baseClass] = pluginSplit($alias);
        $options += ['alias' => $baseClass, 'connection' => $connection];
        $options += $locator->getConfig($alias);

        $reflection = new ReflectionClass($className);
[...]

PR #14792, opened 2020-07-14
A correct fix to getMockForModel issue #14747

CauanCabral:
This solves mock for non-public models methods and a bug when only mocking non-existing methods.


CauanCabral:
Tests broken on nightly PHP are not related, are?

othercorey:
> Tests broken on nightly PHP are not related, are?

phpunit is broken on php nightly. You can ignore.

PR #14797, opened 2020-07-16
Backports from 4.next.

andrii-pukhalevych:
Backport phpdoc from 4.next to fix 'Expected parameter of type 'array|array[]', 'string' provided' hint


dereuromark:
Arent those already fixed in 3.x? https://github.com/cakephp/cakephp/blob/3.x/src/I18n/functions.php#L23
there is no 3.next at this point.

dereuromark:
I took care of that in https://github.com/cakephp/cakephp/commit/7073ab77cd561d98be0159c4091e9d0693851052 already.

PR #14801, opened 2020-07-16
Updated psalm to 3.12.2

othercorey:


src/TestSuite/IntegrationTestTrait.php
```diff
@@ -692,6 +692,7 @@ protected function _castToString(array $data): array
     {
         foreach ($data as $key => $value) {
             if (is_scalar($value)) {
+                /** @psalm-suppress RedundantCondition */
                 $data[$key] = $value === false ? '0' : (string)$value;
```
othercorey, line 696: https://github.com/vimeo/psalm/issues/3829

src/Collection/CollectionTrait.php
```diff
@@ -626,6 +626,7 @@ public function nest($idPath, $parentPath, string $nestingKey = 'children'): Col
             }
             if (empty($key) || !isset($parents[$key])) {
                 foreach ($values as $id) {
+                    /** @psalm-suppress PossiblyInvalidArgument */
                     $parents[$id] = $isObject ? $parents[$id] : new ArrayIterator($parents[$id], 1);
```
othercorey, line 630: This could be a real issue, but this code is too dense to evaluate easily.

src/ORM/Behavior/Translate/ShadowTableStrategy.php
```diff
@@ -483,6 +483,7 @@ protected function rowMapper($results, $locale)
             unset($row['translation']);
 
             if ($hydrated) {
+                /** @psalm-suppress PossiblyInvalidMethodCall */
                 $row->clean();
```
othercorey, line 487: Resetting $row type with @var didn't work here (it is set above). This code does a lot of type juggling that might be written simpler.

othercorey:
https://github.com/cakephp/cakephp-codesniffer/pull/315 to fix the CS error

src/Database/Type/ExpressionTypeCasterTrait.php
```diff
@@ -49,7 +49,7 @@ protected function _castToExpression($value, ?string $type = null)
         $multi = $type !== $baseType;
 
         if ($multi) {
-            /** @psalm-suppress InvalidArgument */
+            /** @var \Cake\Database\Type\ExpressionTypeInterface $converter */
```
ADmad, line 52: IDEs might not like this too without the variable assignment, so could use `@psalm-var` instead.

othercorey: I mean, they wouldn't throw an error though, right?

othercorey: Changed to psalm-var for now since it's a special case.

PR #14809, opened 2020-07-18
Update travis distro to bionic

othercorey:


.travis.yml
```diff
@@ -1,6 +1,6 @@
 language: php
 
-dist: xenial
+dist: bionic
```
ADmad, line 3: Any particular reason for changing this? Bionic is already used for GitHub Actions, so better to keep testing on Xenial on Travis.

othercorey: I was trying to see what issues there were running on travis bionic.

PR #14811, opened 2020-07-19
Use postgresql 9.6 for travis builds

othercorey:
With github actions running latest postgresql, we should try testing on the minimal version supported since 9.6 is still more popular.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,9 @@ services:
   - postgresql
   - mysql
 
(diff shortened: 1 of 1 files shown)


dereuromark:
or even both maye?

othercorey:
I don't know how to run both on travis. 9.6 has the same features as 11 for cakephp. It should be the same coverage.

PR #14813, opened 2020-07-20
Made Http\Client accept url without // and scheme

CauanCabral:
Before 4.0 release, I can use url on `Http\Client` only with DNS ou IP address, but if I do the same now, I'll get a domain with 3 slashes (`///example.com`).

This make Client BC with 3.x releases.

### DIFF
--- a/tests/TestCase/Http/ClientTest.php
+++ b/tests/TestCase/Http/ClientTest.php
@@ -175,6 +175,15 @@ class ClientTest extends TestCase
                 ],
                 'protocol relative url',
             ],
+            [
+                'http://example.com/test.html',
+                'example.com/test.html',
+                [],
+                [
+                    'scheme' => 'http',
+                ],
+                'url without any prefix',
+            ],
         ];
     }
 
--- a/src/Http/Client.php
+++ b/src/Http/Client.php
@@ -512,7 +512,10 @@ class Client implements ClientInterface
         if ($options['port'] && (int)$options['port'] !== $defaultPorts[$options['scheme']]) {
             $out .= ':' . $options['port'];
         }
-        $out .= '/' . ltrim($url, '/');
+        if ($options['host'] !== null) {
+            $out .= '/';
+        }
+        $out .= ltrim($url, '/');
 
         return $out;
     }


CauanCabral:
Build fail from phpstans is not related.

tests/TestCase/Http/ClientTest.php
```diff
@@ -175,6 +175,15 @@ public static function urlProvider()
...
                 'protocol relative url',
             ],
+            [
+                'http://example.com/test.html',
+                'example.com/test.html',
+                [],
+                [
+                    'scheme' => 'http',
+                ],
+                'url without any prefix',
```
ADmad, line 185: Why should URLs without any protocol be allowed when require even explicitly specifying URL with relative protocol using the `protocolRelative` option?

If we accept this patch then we should get rid of `protocolRelative` option too and implicity check if URL starts with `//`.

dereuromark: In which case this would then rather target 4.2+

ADmad: I would rather treat arguments without any protocol as relative paths only as it's currently done.

ADmad: Also the `buildUrl()` method seems to be public only for testing convenience. I don't see why a user would need to use that method directly.

CauanCabral: @ADmad thats not a case of using `buildUrl()` directly, I just traced where the extra slash have been put on a code.

I thought its a BC-break, but giving a better look at both versions and my differents environments - they works (or doesn't works) equal - so, I will close that PR.

Just to clarify what I want:

> Define into `Client` constructor a common base point to next calls (like Guzzle support), example:

```php
$c = new Client(['host' => 'https://example.com/subproject/api/']);

$response = $c->get('/status');
// should call https://example.com/subproject/api/status
// actually call http://https://example.com/subproject/api//status
```

Implicit inferring the scheme of host is just a part of it (in the example, use 'example.com/subproject/api/').

I know this not correct place to put it, but supporting that is wanted at framework?

ADmad: You can open a separate issue/PR regarding the ability to set a base path for the requests so you don't have to misuse the `host` option.

PR #14814, opened 2020-07-21
Allow PDOStatement driver options to be set by SqlServer::prepare

### LINKED ISSUE #14806: SQL Stored Procedure running through sp_cursorprepexec
This is a (multiple allowed):

* [x] bug
* [ ] enhancement

* CakePHP Version: 4.1.0
* PHP Version: 7.4.5
* Platform and Target: Azure WebApp Microsoft Windows NT 10.0.14393.0, Microsoft SQL Azure (RTM) - 12.0.2000.8
* SQLSRV Version: 5.7.1

### What you did
Execute an stored procedure using CakePHP:
```php
$this->connection = ConnectionManager::get('default');
$query = 'EXEC PROCEDURE_NAME PARAM';
$stmt = $this->connection->execute($query);
$result = $stmt->fetchAll('assoc');
$stmt->closeCursor();
```

### What happened
The database server receives the query, but is executed as a cursor such as:
```sql
declare @p1 int
set @pq=1
declare @p2 int
set @p2=0
declare @p5 int
set @p5=8
declare @p6 int
set @p6=8193
declare @p7 int
set @p7=0
exec sp_cursorprepexec @p1 output, @p2 output,NULL,N'EXEC PROCEDURE_NAME PARAMS',@p5 output,@p6 output,@p7 output
select @p1,@p2,@p5,@p6,@p7
```

The DB receives the query as sent
```sql
EXEC PROCEDURE_NAME PARAMS
```

This was reported to me from the DBA which said this is affecting the database performance.
Since we use stored procedures to process large data and retrieve summarized output to present it into dashboards/reports.
And this execution is in loop like every 3-5 seconds.

I noticed that at **src/Database/Driver/Sqlserver.php#195**
https://github.com/cakephp/cakephp/blob/74b320a4fca22280ff311dac57b611fa0a0affd6/src/Database/Driver/Sqlserver.php#L195
Is enforcing this PDO::ATTR_CURSOR attribute as PDO::CURSOR_SCROLL.

To be honest I have no idea if this is what is causing this...
Or if this is an expected behaviour using Microsoft's SQLSRV Driver.
Hopefully there's some way to circumvent it that I missed out and you guys are better suitable to enlighten me.

iomariani:
Fixes #14806

As stated at the issue, the `ATTR_CURSOR` attribute for `PDOStatement` is pre-defined at the driver not allowing us to set specific cursor types types, which can vary depending on our usage.

This PR adds the driver options argument `$options` to be able to include other options when instantiating an PDOStatement for the Sqlserver driver.

So we can simply enforce the cursor type by doing:

```php
$this->connection = ConnectionManager::get('default');
$query = 'EXEC PROCEDURE_NAME PARAM';
$driver = $this->connection->getDriver();
$stmt = $driver->prepare($query, [
    \PDO::SQLSRV_ATTR_CURSOR_SCROLL_TYPE => \PDO::SQLSRV_CURSOR_BUFFERED
]);
$stmt->execute();
$result = $stmt->fetchAll('assoc');
$stmt->closeCursor();
```

### DIFF
--- a/src/Database/Driver/Sqlserver.php
+++ b/src/Database/Driver/Sqlserver.php
@@ -193,7 +193,7 @@ class Sqlserver extends Driver
     public function prepare($query, array $options = []): StatementInterface
     {
         $this->connect();
-        if (!isset($options[PDO::ATTR_CURSOR]) {
+        if (!isset($options[PDO::ATTR_CURSOR])) {
             $options = [PDO::ATTR_CURSOR => PDO::CURSOR_SCROLL] + $options;
         }
         $isObject = $query instanceof Query;
--- a/src/Database/Driver/Sqlserver.php
+++ b/src/Database/Driver/Sqlserver.php
@@ -193,7 +193,9 @@ class Sqlserver extends Driver
     public function prepare($query, array $options = []): StatementInterface
     {
         $this->connect();
-        $options = [PDO::ATTR_CURSOR => PDO::CURSOR_SCROLL] + $options;
+        if (!isset($options[PDO::ATTR_CURSOR]) {
+            $options = [PDO::ATTR_CURSOR => PDO::CURSOR_SCROLL] + $options;
+        }
         $isObject = $query instanceof Query;
         /** @psalm-suppress PossiblyInvalidMethodCall */
         if ($isObject && $query->isBufferedResultsEnabled() === false) {
--- a/src/Database/Driver/Sqlserver.php
+++ b/src/Database/Driver/Sqlserver.php
@@ -187,12 +187,13 @@ class Sqlserver extends Driver
      * Prepares a sql statement to be executed
      *
      * @param string|\Cake\Database\Query $query The query to prepare.
+     * @param array $options The driver options to set on the statement.
      * @return \Cake\Database\StatementInterface
      */
-    public function prepare($query): StatementInterface
+    public function prepare($query, array $options = []): StatementInterface
     {
         $this->connect();
-        $options = [PDO::ATTR_CURSOR => PDO::CURSOR_SCROLL];
+        $options += [PDO::ATTR_CURSOR => PDO::CURSOR_SCROLL];
         $isObject = $query instanceof Query;
         /** @psalm-suppress PossiblyInvalidMethodCall */
         if ($isObject && $query->isBufferedResultsEnabled() === false) {
--- a/src/Database/Driver/Sqlserver.php
+++ b/src/Database/Driver/Sqlserver.php
@@ -193,7 +193,7 @@ class Sqlserver extends Driver
     public function prepare($query, array $options = []): StatementInterface
     {
         $this->connect();
-        $options += [PDO::ATTR_CURSOR => PDO::CURSOR_SCROLL];
+        $options = [PDO::ATTR_CURSOR => PDO::CURSOR_SCROLL] + $options;
         $isObject = $query instanceof Query;
         /** @psalm-suppress PossiblyInvalidMethodCall */
         if ($isObject && $query->isBufferedResultsEnabled() === false) {


src/Database/Driver/Sqlserver.php
```diff
@@ -187,12 +187,13 @@ public function enabled(): bool
...
      * @param string|\Cake\Database\Query $query The query to prepare.
+     * @param array $options The driver options to set on the statement.
      * @return \Cake\Database\StatementInterface
      */
-    public function prepare($query): StatementInterface
+    public function prepare($query, array $options = []): StatementInterface
     {
         $this->connect();
-        $options = [PDO::ATTR_CURSOR => PDO::CURSOR_SCROLL];
+        $options = [PDO::ATTR_CURSOR => PDO::CURSOR_SCROLL] + $options;
```
othercorey, line 196: Don't you want the defaults added afterwards? This won't let `PDO::ATTR_CURSOR` be overridden by $options.

iomariani: Actually, the goal was to _add_ custom driver options, alongside `PDO::ATR_CURSOR`.

But what you said is true, in case someone wants to override that attribute, that won't work with the `+` operator

iomariani: I've submitted a check for _in case_ `PDO::ATTR_CURSOR` is set, just go through with the execution.

iomariani (commented):
Check for in case PDO::ATTR_CURSOR is set, just go through with the execution & syntax fix

src/Database/Driver/Sqlserver.php
```diff
@@ -187,12 +187,15 @@ public function enabled(): bool
...
      * @return \Cake\Database\StatementInterface
      */
-    public function prepare($query): StatementInterface
+    public function prepare($query, array $options = []): StatementInterface
     {
         $this->connect();
-        $options = [PDO::ATTR_CURSOR => PDO::CURSOR_SCROLL];
+        if (!isset($options[PDO::ATTR_CURSOR]) {
+            $options = [PDO::ATTR_CURSOR => PDO::CURSOR_SCROLL] + $options;
+        }
```
othercorey:
```suggestion
    $options = $options + [PDO::ATTR_CURSOR => PDO::CURSOR_SCROLL];
```

iomariani: @othercorey if we do that the `PDO::ATTR_CURSOR` will be overridden, but it will be appended, being the last index of the array. Unfortunately, this would throw the exception:

>SQLSTATE[IMSSP]: The PDO::SQLSRV_ATTR_CURSOR_SCROLL_TYPE attribute may only be set when PDO::ATTR_CURSOR is set to PDO::CURSOR_SCROLL in the $driver_options array of PDO::prepare.

I had it happening to me, seems the `PDO::ATTR_CURSOR` must be set before the other options... hence why I used that check...

othercorey: My mistake, I forgot it was an integer constant.

othercorey: Do you always want ATTR_CURSOR set to something? If don't need to add the defaults if $options is not empty.

iomariani: Actually not... there can be many cases it won't even be required, for instance when you want to fetch all results without iterating.

However I left it there as it was set by default by the framework, so perhaps it's required for the usage of cursor methods later on some other Cake Sqlserver classes.

iomariani: Or you mean something like..

```php
if (empty($options)) {
    $options = [PDO::ATTR_CURSOR => PDO::CURSOR_SCROLL]
}
```

or just set it on the parameter as default?

```php
public function prepare($query, array $options = [PDO::ATTR_CURSOR => PDO::CURSOR_SCROLL]): StatementInterface
```

othercorey: Yes, like the one with only default set. I don't know the history of CURSOR_SCROLL in CakePHP though.

markstory (commented):
Could we get a test covering the usage of this option? There are tests for the SQLServer driver and you could create a stored procedure, call and then drop that procedure to ensure the new mode works.

src/Database/Driver/Sqlserver.php
iomariani: Yeah, to be honest I have no idea either... that's why I left it that way. 😅

I traced it back to 2014 at this commit https://github.com/cakephp/cakephp/commit/3bf580dbe72a278ba2c0276504ac69e5877087b1#diff-6d58c9b7093fe3eba95388ad543a0da1R104 seems it's a default since then.

othercorey: Can you run a test with that attribute removed? I don't know where CakePHP need to scroll backwards in the cursor.

iomariani: @othercorey I tried without any parameters for cursor set and it throws the following error:
> array size cannot be less than zero
> InvalidArgumentException 

**Trace:**
https://github.com/cakephp/cakephp/blob/47912247fa64dc35066c8fe74d24bee38480d4b6/src/ORM/ResultSet.php#L177
https://github.com/cakephp/cakephp/blob/47912247fa64dc35066c8fe74d24bee38480d4b6/src/ORM/Query.php#L1114
https://github.com/cakephp/cakephp/blob/47912247fa64dc35066c8fe74d24bee38480d4b6/src/Datasource/QueryTrait.php#L292
https://github.com/cakephp/cakephp/blob/47912247fa64dc35066c8fe74d24bee38480d4b6/src/ORM/Query.php#L1059
https://github.com/cakephp/cakephp/blob/47912247fa64dc35066c8fe74d24bee38480d4b6/src/Datasource/QueryTrait.php#L141

Although, following the trace seems I'd need to readapt some code to avoid this.

Wherever I use ORM to find Database results I'd need to call ORM's `disableBufferedResults` method:

```php
$users = $this->getTableLocator()->get('Users')->find('all')->disableBufferedResults();
```

Once I did that it worked flawlessly. I will check how the database receives the EXEC queries, however.

iomariani: @othercorey  indeed, I've just tested that and it was much more performatic as well, and the query was received by the server as it was sent.

Screen Shot 2020-07-21 at 00 33 36

Ran three procedures with the settings:
**red** is using `CURSOR_SCROLL`
**green** is using `CURSOR_FWDONLY`

The `CURSOR_SCROLL` enforces the usage of `sp_cursorprepexec` which runs a lot of cursor operations to create a handle, fetch result, and return.


According to [Microsoft documentation](https://docs.microsoft.com/en/sql/connect/php/pdo-prepare?view=sql-server-ver15#remarks) the default is `CURSOR_FWDONLY`.
Even since when I use the `CURSOR_SCROLL` and `SQLSRV_CURSOR_BUFFERED` seems equally performatic as `CURSOR_FWDONLY`, I will keep only `CURSOR_FWDONLY` on your app.

iomariani:
> Could we get a test covering the usage of this option? There are tests for the SQLServer driver and you could create a stored procedure, call and then drop that procedure to ensure the new mode works.

Sorry, I'm not exactly sure what you mean, do you have any example you can point me into how to do that?

src/Database/Driver/Sqlserver.php
iomariani: @othercorey I think it's safe to say rather set Cake's default also to `CURSOR_FWDONLY`? 😅

othercorey: Probably. Like I said earlier - I don't know where cakephp needs to scroll backwards so FWDONLY should work.

othercorey:
> Could we get a test covering the usage of this option? There are tests for the SQLServer driver and you could create a stored procedure, call and then drop that procedure to ensure the new mode works.

We always need a cursor to buffer, but should switch to FWDONLY to fix the performance issue.

src/Database/Driver/Sqlserver.php
```diff
@@ -187,12 +187,15 @@ public function enabled(): bool
...
+     * @param array $options The driver options to set on the statement.
      * @return \Cake\Database\StatementInterface
      */
-    public function prepare($query): StatementInterface
+    public function prepare($query, array $options = []): StatementInterface
     {
         $this->connect();
-        $options = [PDO::ATTR_CURSOR => PDO::CURSOR_SCROLL];
+        if (!isset($options[PDO::ATTR_CURSOR])) {
+            $options = [PDO::ATTR_CURSOR => PDO::CURSOR_SCROLL] + $options;
```
ADmad, line 197:
```suggestion
            $options[PDO::ATTR_CURSOR] = PDO::CURSOR_SCROLL;
```

othercorey:
Looks like there is some timing error when using FWDONLY. 

https://github.com/cakephp/cakephp/pull/14821

src/Database/Driver/Sqlserver.php
iomariani: Unfortunately this would throw the same exception as: https://github.com/cakephp/cakephp/pull/14814#discussion_r457783780

ADmad: That's pretty silly that the options need to be set in a specific order.

iomariani: I know right, seems they do a validation per option set, instead of the whole array.

PR #14817, opened 2020-07-21
check datatype for localizedTime cakephp#14800

### LINKED ISSUE #14800: Validation cannot validate invalid datetime
This is a (multiple allowed):

* [x] bug
* [ ] enhancement

* CakePHP Version: 4.1
* Platform and Target: Windows, NGINX, PHP 7.3, SQL Server

### What you did
Define validation method in table class:
```php
public function validationDefault(Validator $validator): Validator
{
    // ...
    $validator->notEmptyDateTime('field');
    // ...
    return $validator;
}
```

Patch entity in controller:
```
$entity = $table->patchEntity($entity, $requestData);
$table->save($entity);
```

Request data:
```
field=test
```

### What happened
SQL error occurred: Column 'field' cannot be null

Local time format (`H:i, Y/m/d` in `vi_VN` locale) cannot be validated if I add below code to validator
```
$validator->dateTime('field', ['ymd', 'dmy']);
```

And if I don't use `H:i, Y/m/d` format, time string cannot be parsed by local parser. See #14071 

Error on `field` of entity

mcsknp:
Fix #14800

### DIFF
--- a/src/Validation/Validation.php
+++ b/src/Validation/Validation.php
@@ -658,7 +658,7 @@ class Validation
         if ($check instanceof DateTimeInterface) {
             return true;
         }
-        if (is_object($check)) {
+        if (!is_string($check)) {
             return false;
         }
         static $methods = [
--- a/tests/TestCase/Validation/ValidationTest.php
+++ b/tests/TestCase/Validation/ValidationTest.php
@@ -1759,6 +1759,8 @@ class ValidationTest extends TestCase
 
         $this->assertFalse(Validation::localizedTime('', 'date'));
         $this->assertFalse(Validation::localizedTime('invalid', 'date'));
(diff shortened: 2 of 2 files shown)


src/Validation/Validation.php
```diff
@@ -658,7 +658,7 @@ public static function localizedTime($check, string $type = 'datetime', $format
         if ($check instanceof DateTimeInterface) {
             return true;
         }
-        if (is_object($check)) {
+        if (!is_string($check)) {
```
markstory, line 661: It would be good to expand the test cases for this method to cover this scenario. Do you need a hand with that?

mcsknp: I added some test cases.

PR #14818, opened 2020-07-21
fix Validation::custom function arguments cakephp#14816

### LINKED ISSUE #14816: Error at Validator::regex()
This is a (multiple allowed):

* [x] bug
* [ ] enhancement

* CakePHP Version: 4.x
* Platform and Target: Any

### What you did
Use `Validator::regex()` function and test with below request data
```
field[]=sample&field[]=value
```

### What happened
System error occurred at https://github.com/cakephp/cakephp/blob/47912247fa64dc35066c8fe74d24bee38480d4b6/src/Validation/Validation.php#L443

Remove datatype for argument and add `is_string` check

mcsknp:
Fix #14816

### DIFF
--- a/src/Validation/Validation.php
+++ b/src/Validation/Validation.php
@@ -436,12 +436,16 @@ class Validation
     /**
      * Used when a custom regular expression is needed.
      *
-     * @param string $check The value to check.
+     * @param mixed $check The value to check.
      * @param string|null $regex If $check is passed as a string, $regex must also be set to valid regular expression
      * @return bool Success
      */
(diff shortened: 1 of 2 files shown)


dereuromark (approved):
This is a small BC break on the method signature
But it fixes this method to be used properly, and as such I think it is warranted in this case.

ADmad (approved):

markstory:
@dereuromark Widening the parameter type shouldn't break backwards compatibility. Any value that was accepted before would continue to be accepted now.

PR #14819, opened 2020-07-21
cast array key to string cakephp#14810

### LINKED ISSUE #14810: Error when create/patch entity with array data has numeric key
This is a (multiple allowed):

* [x] bug
* [ ] enhancement

* CakePHP Version: 4.1
* Platform and Target: Windows, NGINX, PHP7.3, SQL server

### What you did
Patch entity with below request data
```
0=sample&1=value&....
```

### What happened
Below error occurred
```
Argument 1 passed to Cake\Database\Schema\TableSchema::getColumnType() must be of the type string, int given
```
at
https://github.com/cakephp/cakephp/blob/4f635e709ad0a9610a0ca4a1e2b6129c08ec6cd0/src/ORM/Marshaller.php#L75

`$prop` must be a scalar and cast to string before pass to `TableSchema::getColumnType()`

mcsknp:
Fix #14810

### DIFF
--- a/tests/TestCase/ORM/EntityTest.php
+++ b/tests/TestCase/ORM/EntityTest.php
@@ -71,11 +71,11 @@ class EntityTest extends TestCase
         $this->assertEquals(1, $entity->getOriginal('id'));
 
         $entity->set(['foo', 'bar']);
-        $this->assertEquals('foo', $entity->{0});
-        $this->assertEquals('bar', $entity->{1});
+        $this->assertEquals('foo', $entity->get('0'));
+        $this->assertEquals('bar', $entity->get('1'));
 
         $entity->set(['sample']);
-        $this->assertEquals('sample', $entity->{0});
+        $this->assertEquals('sample', $entity->get('0'));
     }
 
     /**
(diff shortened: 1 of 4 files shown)


tests/TestCase/ORM/EntityTest.php
```diff
@@ -69,6 +69,13 @@ public function testSetMultiplePropertiesNoSetters()
         $this->assertSame(3, $entity->thing);
         $this->assertSame('bar', $entity->getOriginal('foo'));
         $this->assertEquals(1, $entity->getOriginal('id'));
+
+        $entity->set(['foo', 'bar']);
+        $this->assertEquals('foo', $entity->{0});
```
ADmad, line 74: Better to use `$entity[0]` or `$entity->get('0')` instead of `$entity->{0}`.

mcsknp: I fixed it.

tests/TestCase/ORM/MarshallerTest.php
```diff
@@ -132,6 +132,25 @@ public function testOneSimple()
...
+    public function testOneWithNumericField()
+    {
+        $data = [
+            'sample',
+            'username' => 'test',
+            'password' => 'secret',
+            1,
+        ];
+        $marshall = new Marshaller($this->articles);
+        $result = $marshall->one($data, []);
```
dereuromark, line 149: assertSame() are usually the better fit for scalar values.

PR #14822, opened 2020-07-21
Fix TypeError on PHP 8

ADmad:

### DIFF
--- a/src/Utility/Text.php
+++ b/src/Utility/Text.php
@@ -226,13 +226,13 @@ class Text
 
         $dataKeys = array_keys($data);
         $hashKeys = array_map('crc32', $dataKeys);
-        /** @var array<string, string> $tempData */
+        /** @var array<string, string|int> $tempData */
         $tempData = array_combine($dataKeys, $hashKeys);
         krsort($tempData);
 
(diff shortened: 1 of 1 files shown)


src/Utility/Text.php
```diff
@@ -232,7 +232,7 @@ public static function insert(string $str, array $data, array $options = []): st
 
         foreach ($tempData as $key => $hashVal) {
             $key = sprintf($format, preg_quote($key, '/'));
-            $str = preg_replace($key, $hashVal, $str);
+            $str = preg_replace($key, (string)$hashVal, $str);
```
othercorey, line 235: If $tmpData is type array<string, string> what type is getting passed in?

ADmad: Apparently `$tempData` isn't exactly `array<string, string>` :slightly_smiling_face: 

https://travis-ci.org/github/ADmad/cakephp/jobs/710436188#L498

ADmad: I updated the inline docblock for `$tempData`.

othercorey: I see, it's from the call to `crc32` above. Since `crc32` will enforce string type anyway I suppose this is fine to cast here instead.

PR #14825, opened 2020-07-22
Add session based CSRF token middleware

markstory:
The current double submit cookie CSRF middleware has a few potential weaknesses:

1. The tokens it uses are portable between users. If a user has their    browser compromised, CSRF tokens can be re-used between other    compromised users.
2. Tokens don't have an expiration time. If a CSRF token is leaked and   a user is compromised an old token can be used again.

These changes introduce an alternative CSRF middleware that stores the compare token in the session. This removes the need to do HMAC signing as the compare value is in the session where an attacker cannot change it. Furthermore, it also couples CSRF tokens to the user, and to the specific session. This mitigates both weaknesses with our existing CSRF middleware.

I didn't want to make the existing CSRF middleware more complicated and risk breaking backwards compatibility which is why this is a new middleware. A separate middleware makes opt-ing into session based tokens explicit.

### DIFF
--- a/src/Http/Middleware/SessionCsrfProtectionMiddleware.php
+++ b/src/Http/Middleware/SessionCsrfProtectionMiddleware.php
@@ -11,7 +11,7 @@ declare(strict_types=1);
  *
  * @copyright     Copyright (c) Cake Software Foundation, Inc. (http://cakefoundation.org)
  * @link          http://cakephp.org CakePHP(tm) Project
- * @since         3.5.0
+ * @since         4.2.0
  * @license       http://www.opensource.org/licenses/mit-license.php MIT License
  */
 namespace Cake\Http\Middleware;
--- a/src/Http/Middleware/SessionCsrfProtectionMiddleware.php
+++ b/src/Http/Middleware/SessionCsrfProtectionMiddleware.php
@@ -1,0 +1,217 @@
+<?php
+declare(strict_types=1);
+
+/**
+ * CakePHP(tm) : Rapid Development Framework (http://cakephp.org)
+ * Copyright (c) Cake Software Foundation, Inc. (http://cakefoundation.org)
+ *
+ * Licensed under The MIT License
+ * For full copyright and license information, please see the LICENSE.txt
+ * Redistributions of files must retain the above copyright notice.
+ *
+ * @copyright     Copyright (c) Cake Software Foundation, Inc. (http://cakefoundation.org)
+ * @link          http://cakephp.org CakePHP(tm) Project
+ * @since         3.5.0
+ * @license       http://www.opensource.org/licenses/mit-license.php MIT License
+ */
+namespace Cake\Http\Middleware;
+
+use ArrayAccess;
+use Cake\Http\Exception\InvalidCsrfTokenException;
+use Cake\Http\Session;
+use Cake\Utility\Hash;
+use Cake\Utility\Security;
+use Psr\Http\Message\ResponseInterface;
+use Psr\Http\Message\ServerRequestInterface;
+use Psr\Http\Server\MiddlewareInterface;
+use Psr\Http\Server\RequestHandlerInterface;
+use RuntimeException;
+
+/**
+ * Provides CSRF protection via session based tokens.
+ *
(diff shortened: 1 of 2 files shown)


src/Http/Middleware/SessionCsrfProtectionMiddleware.php
```diff
@@ -0,0 +1,217 @@
...
+ * CakePHP(tm) : Rapid Development Framework (http://cakephp.org)
+ * Copyright (c) Cake Software Foundation, Inc. (http://cakefoundation.org)
+ *
+ * Licensed under The MIT License
+ * For full copyright and license information, please see the LICENSE.txt
+ * Redistributions of files must retain the above copyright notice.
+ *
+ * @copyright     Copyright (c) Cake Software Foundation, Inc. (http://cakefoundation.org)
+ * @link          http://cakephp.org CakePHP(tm) Project
+ * @since         3.5.0
```
othercorey, line 14:
```suggestion
 * @since         4.2.0
```

src/Http/Middleware/SessionCsrfProtectionMiddleware.php
```diff
@@ -0,0 +1,217 @@
...
+    /**
+     * Config for the CSRF handling.
+     *
+     *  - `key` The session key to use. Defaults to `csrfToken`
+     *  - `field` The form field to check. Changing this will also require configuring
+     *    FormHelper.
+     *
+     * @var array
+     */
+    protected $_config = [
```
othercorey, line 59:
```suggestion
    protected $config = [
```
What is the cakephp policy for underscore protected? Are we transitioning away? It seems we randomly avoid them and randomly copy them.

othercorey:
> 1. The tokens it uses are portable between users. If a user has their    browser compromised, CSRF tokens can be re-used between other    compromised users.
> 2. Tokens don't have an expiration time. If a CSRF token is leaked and   a user is compromised an old token can be used again.

This essentially only protects other users, right? If the browser is compromised then the user's session is essentially compromised. If they do fix the browser exploit then they are only vulnerable until the session is changed.

src/Http/Middleware/SessionCsrfProtectionMiddleware.php
ADmad: We generally avoid the underscore for new protected properties but for `$_config` I would prefer keeping it for consistency with other classes.

We'll have to live with this till v5.

src/Http/Middleware/SessionCsrfProtectionMiddleware.php
```diff
@@ -0,0 +1,217 @@
...
+                return;
+            }
+        }
+
+        $header = $request->getHeaderLine('X-CSRF-Token');
+        if (hash_equals($header, $token)) {
+            return;
+        }
+
+        throw new InvalidCsrfTokenException(__d(
```
dereuromark, line 212: I thought we dont translate internal messages anymore? Or are those visible to the end user on the page?

src/Http/Middleware/SessionCsrfProtectionMiddleware.php
markstory: We could start making things better now so there are fewer things to change in the future.

src/Http/Middleware/SessionCsrfProtectionMiddleware.php
markstory: These can show up in error pages. The other CSRF middleware also has these translated.

ADmad (approved):

smarek:
May I ask why this did not land with 4.1.3 or 4.1.4 release?

othercorey:
New features generally go into the next point release. This makes it easier to document changes.

garas:
CakePHP follows [semver](https://semver.org/).

New features will be in next _minor_ release (4.2.0).
_Patch_ releases (like 4.1.4) fix bugs in existing code.

markstory:
If you need it *right now* you could copy/paste the code as an application middleware until 4.2 ships.

smarek:
Thank you both for explanation :)

PR #14826, opened 2020-07-22
Pin phpstan to 0.12.32

othercorey:
Check that this still passes and we can pin until next update.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -112,7 +112,7 @@
             "@phpstan",
             "@psalm"
         ],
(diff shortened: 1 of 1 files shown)


ADmad:
Alternative would be regenerate the baseline file instead of pinning to new version. That way we wouldn't miss any other improvements in the new release.

othercorey:
I'm starting to prefer manually updating every so often. Unfortunately, phpstan doesn't use major versioning so it's not as easy as psalm to pin to bug releases only.

ADmad:
> I'm starting to prefer manually updating every so often

That works too.

PR #14827, opened 2020-07-22
Use specific test methods to assert PHP errors generation.

ADmad:
This will avoid deprecation errors in PHPUnit 9.


othercorey:
phpstan 9 doesn't let you test for a specific exception only exception message? I suppose the message should be unique, but still.

ADmad:
> phpstan 9 doesn't let you test for a specific exception only exception message?

It does, but for `PHPUnit\Framework\Error\Error|Warning|Notice` you need to use the corresponding methods instead of the general `expectException()`.

othercorey:
oh, got it. I was reading messages and assumed it was a cake exception.

PR #14829, opened 2020-07-22
Use md5 as temporary placeholder in Text::insert()

garas:
`Text::insert()` currently uses `crc32` as temporary placeholder.

It can accidentally return unexpected output with common keys or values, because `crc32` value can be very short (`crc32('ob') = 26083`, `crc32('jskng') = 400`).

`md5` is long enough to not accidentally collide with another value, also it is always same length string, so I undid 
https://github.com/cakephp/cakephp/pull/14822.

### DIFF
--- a/src/Utility/Text.php
+++ b/src/Utility/Text.php
@@ -198,19 +198,8 @@ class Text
             'before' => ':', 'after' => '', 'escape' => '\\', 'format' => null, 'clean' => false,
         ];
         $options += $defaults;
-        $format = $options['format'];
-        $data = $data;
         if (empty($data)) {
             return $options['clean'] ? static::cleanInsert($str, $options) : $str;
-        }
-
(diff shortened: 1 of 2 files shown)


garas:
This function also supports `?` placeholders `Text::insert('this is a ? string', ['test']) = 'this is a test string'`, but it is undocumented anywhere.

Should this be documented or deprecated?

dereuromark:
Is there any reason to keep it?
As sprintf() can easily take care of basic replacing.
So I see Text::insert() only useful for named ones.

markstory:
> This function also supports ? placeholders Text::insert('this is a ? string', ['test']) = 'this is a test string', but it is undocumented anywhere.

We should deprecate this in 4.2. I agree that we aren't providing any value that sprintf() can't do better with that style of placeholder.

PR #14830, opened 2020-07-22
Fix errors in MysqlTest

ADmad:


othercorey:
How was this not failing with the missing use statement.

PR #14841, opened 2020-07-25
Fix open cursors

othercorey:
Fixes a couple more open cursors found where only execute() is called.

### DIFF
--- a/tests/TestCase/Database/QueryTest.php
+++ b/tests/TestCase/Database/QueryTest.php
@@ -1709,18 +1709,17 @@ class QueryTest extends TestCase
         $query = new Query($this->connection);
         $query->select(['id'])
             ->from('articles')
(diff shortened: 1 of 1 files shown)


tests/TestCase/Database/QueryTest.php
```diff
@@ -1709,18 +1709,17 @@ public function testWhereInArrayEmpty()
...
         $this->assertQuotedQuery(
             'SELECT <id> FROM <articles> WHERE 1=0',
-            $sql,
+            $query->sql(),
             !$this->autoQuote
         );
+
+        $statement = $query->execute();
+        $this->assertFalse($statement->fetch('assoc'));
+        $statement->closeCursor();
```
othercorey, line 1722: This closeCursor() is probably not necessary because fetch() returns false. The issue was the double execute() above.

PR #14842, opened 2020-07-25
Use client-side buffered cursor type

othercorey:
Closes https://github.com/cakephp/cakephp/issues/14806

Based on PDO_SQLSRV documentation, CURSOR_SCROLL types can be configured as SQLSRV_CURSOR_BUFFERED which moves the cursor client-side like you'd expect with buffered queries for other dbs.

https://docs.microsoft.com/en-us/sql/connect/php/cursor-types-pdo-sqlsrv-driver?view=sql-server-ver15
https://docs.microsoft.com/en-us/sql/connect/php/pdo-prepare?view=sql-server-ver15

The CURSOR_FWDONLY type (default) does not allow type attributes.

As noted in the docs, this is for small to medium size result sets. You can configure the maximum buffered size through attributes or php.ini settings.

### DIFF
--- a/src/Database/Driver/Sqlserver.php
+++ b/src/Database/Driver/Sqlserver.php
@@ -192,7 +192,10 @@ class Sqlserver extends Driver
     public function prepare($query): StatementInterface
     {
         $this->connect();
-        $options = [PDO::ATTR_CURSOR => PDO::CURSOR_SCROLL];
+        $options = [
+            PDO::ATTR_CURSOR => PDO::CURSOR_SCROLL,
+            PDO::SQLSRV_ATTR_CURSOR_SCROLL_TYPE => PDO::SQLSRV_CURSOR_BUFFERED,
+        ];
         $isObject = $query instanceof Query;
         /** @psalm-suppress PossiblyInvalidMethodCall */
         if ($isObject && $query->isBufferedResultsEnabled() === false) {
--- a/psalm.xml
+++ b/psalm.xml
@@ -38,6 +38,7 @@
         <UndefinedConstant>
             <errorLevel type="suppress">
                 <file name="src/Cache/Engine/ApcuEngine.php" />
+                <file name="src/Database/Driver/Sqlserver.php" />
             </errorLevel>
         </UndefinedConstant>
         <PropertyNotSetInConstructor errorLevel="suppress"/>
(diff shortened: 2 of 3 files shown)


othercorey:
One of the benefits of a client side cursor is the num_rows call returns valid data which is why the unit test changed.

src/Database/Driver/Sqlserver.php
```diff
@@ -192,7 +192,10 @@ public function enabled(): bool
     public function prepare($query): StatementInterface
     {
         $this->connect();
-        $options = [PDO::ATTR_CURSOR => PDO::CURSOR_SCROLL];
+        $options = [
+            PDO::ATTR_CURSOR => PDO::CURSOR_SCROLL,
+            PDO::SQLSRV_ATTR_CURSOR_SCROLL_TYPE => PDO::SQLSRV_CURSOR_BUFFERED,
+        ];
```
markstory, line 198: Should we allow connection/driver level configuration to choose the cursor type? If users are loading large results they might not want client side cursors.

othercorey: `Query::disableBufferedResults()` will clear the cursor setting (just below this code). 

Do we have a global way to set that for other dbs? I don't think postgres has any buffering settings. Mysql uses the same `Query::disableBufferedResults()` flag.

othercorey: I wonder if any users need options like PDO::SQLSRV_CURSOR_STATIC for server-side cursors.

iomariani: I wouldn't personally use server-side cursors due to performance, but since they are an option available in the drivers, I believe any chance for users to _customize_ it are well welcomed.

markstory (approved):
These changes look good to me. @iomariani do they help solve your problem?

src/Database/Driver/Sqlserver.php
othercorey: I think we can solve this through a Configure setting that applies defaults globally instead of per-call.

iomariani:
> These changes look good to me. @iomariani do they help solve your problem?

Definitely, like I mentioned, the changes are what I've left as defaults on our apps.
For me looks great to merge, thank you all for your help! 😉

PR #14844, opened 2020-07-28
Fix integration test helpers for session based CSRF

markstory:
* Fix the CSRF test helpers to work with session based tokens.
* Add integration tests for both session and cookie CSRF tokens  as we didn't actually have integration tests for these.

### DIFF
--- a/src/TestSuite/IntegrationTestTrait.php
+++ b/src/TestSuite/IntegrationTestTrait.php
@@ -674,7 +674,7 @@ trait IntegrationTestTrait
                 $token = $middleware->createToken();
             } elseif (isset($this->_cookie['csrfToken'])) {
                 $token = $this->_cookie['csrfToken'];
-            } elseif (isset($this->_session['csrfToken'])) {
+            } else {
                 $token = $this->_session['csrfToken'];
             }
 
--- a/src/TestSuite/IntegrationTestTrait.php
+++ b/src/TestSuite/IntegrationTestTrait.php
@@ -583,7 +583,6 @@ trait IntegrationTestTrait
             'defaults' => 'php',
         ];
         $session = Session::create($sessionConfig);
-        $session->write($this->_session);
         [$url, $query, $hostInfo] = $this->_url($url);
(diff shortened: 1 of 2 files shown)


src/TestSuite/IntegrationTestTrait.php
```diff
@@ -669,11 +669,25 @@ protected function _addTokens(string $url, array $data): array
...
+                $token = $this->_cookie['csrfToken'];
+            } else {
+                $token = $this->_session['csrfToken'];
+            }
+
+            // Add the token to both the session and cookie to cover
+            // both types of CSRF tokens. We generate the token with the cookie
+            // middleware as cookie tokens will be accepted by session csrf, but not
+            // the inverse.
+            if ($token) {
```
othercorey, line 685: In what scenario is this not set here?

tests/TestCase/TestSuite/IntegrationTestTraitTest.php
```diff
@@ -69,6 +71,16 @@ public function setUp(): void
             $routes->options('/options/:controller/:action', []);
             $routes->connect('/:controller/:action/*', []);
         });
+        Router::scope('/cookie-csrf/', ['csrf' => 'cookie'], function (RouteBuilder $routes) {
+            $routes->registerMiddleware('cookieCsrf', new CsrfProtectionMiddleware());
```
othercorey, line 75: I wonder if we should rename this to CookieCsrfProtectionMiddleware.

src/TestSuite/IntegrationTestTrait.php
markstory: You're right it will always be set, I was being overly cautious.

tests/TestCase/TestSuite/IntegrationTestTraitTest.php
markstory: We could do that, we'd need an alias for backwards compatibility.

othercorey (approved):
Looks ok to me.

PR #14846, opened 2020-07-28
Backport 4.x change to 3.next

andrii-pukhalevych:
Removing original property on unset

### DIFF
--- a/src/Datasource/EntityTrait.php
+++ b/src/Datasource/EntityTrait.php
@@ -440,7 +440,7 @@ trait EntityTrait
     {
         $property = (array)$property;
         foreach ($property as $p) {
-            unset($this->_properties[$p], $this->_dirty[$p]);
+            unset($this->_properties[$p], $this->_original[$p], $this->_dirty[$p]);
         }
 
         return $this;


dereuromark:
Can you please rebase? I fixed the tests.

andrii-pukhalevych:
Yes, but I don't follow how to do it. I don't see your changes on 3.next

dereuromark:
Well, i merged 3.x into it.

othercorey:
Just pull upstream then merge upstream/3.next into your branch.

dereuromark:
Or git rebase origin/3.next

andrii-pukhalevych:
Thanks. Will do at morning, writing from phone now

dereuromark:
Do we want to put this into 3.x or 3.next?
It looks reasonably BC, no?

andrii-pukhalevych:
In https://github.com/cakephp/cakephp/pull/13482 was agreement to not push it into 3.x and add in 3.9 only

markstory:
Is this destined for a 3.10 release? It might be a while before that comes out. If this is a bugfix it should go into 3.x instead.

othercorey:
FYI, this was changed in 4.0 in https://github.com/cakephp/cakephp/pull/13621

othercorey:
This was also agreed to only be in 4.x here - https://github.com/cakephp/cakephp/pull/13482#issuecomment-520526473

ADmad:
Reading through the comments of the original PR, this is the type of change that could potentially cause unexpected behaviour for existing 3.x apps and would be difficult to track down. So let's keep it for 4.x only as agreed.

dereuromark:
3.next is fair here, no?

andrii-pukhalevych:
Next for 3.8 was 3.9 ;)
This change does not look unsafe for me, specially for 3.9.

dereuromark:
Now it would be 3.10.

andrii-pukhalevych:
So should I change base to 3.next?

othercorey:
I think if this was going to be in a 3.x release then it should be 3.9. However, it seems that was declined. 3.10 will probably never ship - if we are being honest.

dereuromark:
I think 3.10 could still land, given our lpng support.

dereuromark:
So are we going to add a 3.10 milestone and branch?
There is now also a 2nd topic (Mailer/Email) that would need to land there.

markstory:
> So are we going to add a 3.10 milestone and branch?

Yeah we can. I won't have much time to dedicate towards that branch but it can be useful for community contributions.

othercorey:
We should resolve this. I am not taking sides on 3.10, but we should do *something* to make it official if we are decide to take changes.

Should we set a rough release date?

othercorey:
Closing as we've decided this too risky of a change at the end of 3.x -- as discussed in the original PR.

PR #14853, opened 2020-07-30
Use CURLOPT_NOBODY when sending HEAD requests. Fixes #14851

### LINKED ISSUE #14851: Fix HTTP Client Curl adapter with HEAD requests with non-zero content-length in response
This is a:

* [x] bug
* [ ] enhancement

* CakePHP Version: 3.9.1 (but is also broken in older and newer versions)
* Platform and Target: PHP 7.4.3, Ubuntu 20.04, libcurl 7.68.0

### What you did
Do a `HEAD` request to a webservice that returns a `Content-Length > 0` but no body, in accordance with the [HTTP spec](https://tools.ietf.org/html/rfc2616#section-9.4).
For example Elasticsearch, see https://github.com/elastic/elasticsearch/pull/21123 .

### What happened
I get an error `cURL Error (18) transfer closed with 431 bytes remaining to read`.

The HTTP request succeeds

### What is the fix?
Add a case [here](https://github.com/cakephp/cakephp/blob/master/src/Http/Client/Adapter/Curl.php#L105):

```php
            case Request::METHOD_HEAD:
                $out[CURLOPT_NOBODY] = true;
                break;
```

If you have a webserver which sends a non-zero content length for HEAD requests, you can reproduce it using the curl CLI:

```bash
curl -H 'Connection: close' -X HEAD [url here]
```

Note that using `-I` instead of `X HEAD` does work, since that will set `CURLOPT_NOBODY` correctly.

nickygerritsen:
This is the same fix as #14852 but then for CakePHP 4.x

### DIFF
--- a/src/Http/Client/Adapter/Curl.php
+++ b/src/Http/Client/Adapter/Curl.php
@@ -103,6 +103,10 @@ class Curl implements AdapterInterface
                 $out[CURLOPT_POST] = true;
                 break;
 
+            case Request::METHOD_HEAD:
+                $out[CURLOPT_NOBODY] = true;
+                break;
+
             default:
                 $out[CURLOPT_POST] = true;
                 $out[CURLOPT_CUSTOMREQUEST] = $request->getMethod();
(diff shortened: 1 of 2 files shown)


markstory:
Thank you 👍

PR #14855, opened 2020-07-30
Test psalm 3.13.0

othercorey:
Confirming that psalm 3.13.0 never finishes.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -112,7 +112,7 @@
             "@phpstan",
             "@psalm"
         ],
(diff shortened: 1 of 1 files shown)


ADmad:
v3.13.0 is crashing. But I reported it and it has already been fixed in v3.13.1. Just need to wait few mins until the new phar is available.

othercorey:
ah, I couldn't tell what was happening. It self-destructed my vm.

PR #14857, opened 2020-07-30
Fix inconsistencies with datefmt_create constants

othercorey:
This fixes some issue with `IntlDateFormatter` constants used by datefmt_create. Some of these were existing and some were caused by other fixes in 4.0 release.

Previously, both phpstan and psalm would throw invalid errors on the datefmt_create parameters. Both of these errors are now fixed which lets us clean things up.

- Use `int[]` instead of `array` for the formatter constants
- Replace -1 with `IntlDateFormatter::NONE`
- Don't default to 0 when one of the date/time formatters is null. 0 is `IntlDateFormatter::FULL`.
- When a single constant is passed as the argument for $format, this should be the format use for both date and time.
  This was the documented behavior and is now implemented by both formatter and parser interfaces.

This also removes the invalid exception when a single constant was passed. This was added in 4.0 due to a misunderstanding from the inconsistent implementation and static analyzer bugs.

### DIFF
--- a/src/I18n/DateFormatTrait.php
+++ b/src/I18n/DateFormatTrait.php
@@ -347,8 +347,9 @@ trait DateFormatTrait
             $pattern = $format;
         }
 
+        $locale = static::$defaultLocale ?? I18n::getLocale();
         $formatter = datefmt_create(
-            static::$defaultLocale,
+            $locale,
             $dateFormat,
             $timeFormat,
             $tz,
--- a/src/I18n/DateFormatTrait.php
+++ b/src/I18n/DateFormatTrait.php
@@ -21,7 +21,6 @@ use Closure;
 use DateTime;
 use DateTimeZone;
 use IntlDateFormatter;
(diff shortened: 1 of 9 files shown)


src/I18n/DateFormatTrait.php
```diff
@@ -325,41 +324,37 @@ public static function setJsonEncodeFormat($format): void
...
-            if (is_int($pattern)) {
-                throw new InvalidArgumentException(
-                    'If $format is an IntlDateFormatter constant, must be an array.'
-                );
-            }
+            $dateFormat = $timeFormat = IntlDateFormatter::FULL;
+            $pattern = $format;
         }
 
+        $locale = static::$defaultLocale ?? I18n::getLocale();
```
othercorey, line 350: I'm not sure `$defaultLocale` should be used here. It is documented as the default locale for formatting objects not parsing them. Should we just use `I18n::getLocale()` ?

PR #14860, opened 2020-07-31
Remove unneeded method call.

ADmad:
loadConfig() is already called from constructor.

### DIFF
--- a/src/Core/PluginCollection.php
+++ b/src/Core/PluginCollection.php
@@ -122,8 +122,6 @@ class PluginCollection implements Iterator, Countable
      */
     public function findPath(string $name): string
     {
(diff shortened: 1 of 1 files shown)


ADmad:
Seems like we'll have to wait till tomorrow for the "timeAgoInWords" failures to go away.

garas:
2 years, 5 months, 2 days ago = ~2018-02-29~ = 2018-03-01 = 2 years, 4 months, 4 weeks, 2 days ago

ADmad:
@garas Yeah it's valid just not in format the test case expects.

garas:
Maybe `-3 days` could be used so day 28 always exist.

Are there more dates to fail?

PR #14863, opened 2020-08-01
4.next - Add a getter for obtaining a query's results casting mode.

ndm2:
This aligns the results casting API with hydration and results buffering, which both have getters for the current mode (`isHydrationEnabled()`, `isBufferedResultsEnabled()`).

### DIFF
--- a/src/Database/Query.php
+++ b/src/Database/Query.php
@@ -2188,7 +2188,11 @@ class Query implements ExpressionInterface, IteratorAggregate
     }
 
     /**
-     * Disables the automatic casting of fields to their corresponding PHP data type
+     * Disables result casting.
+     *
+     * When disabled, the fields will be returned as received from the database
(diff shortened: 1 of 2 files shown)


ADmad (approved):

othercorey (approved):
Looks good to me.

othercorey:
We could do a better job with the docblock for this setting probably.

ndm2:
> We could do a better job with the docblock for this setting probably.

Copy pasta was very convenient :)

I can add a line or two so that it tells a little something about what happens in enabled/disabled mode.

ndm2:
Hope you don't mind me testing the re-request review button here for the first time, it's interesting that approvals don't automatically get at least "soft"-invalidated when pushing changes afterwards.

othercorey (approved):

PR #5696, opened 2015-01-19
Remove query as an argument to result formatter methods.

markstory:
The Query argument is not always passed consistently, and is a bit unnecessary as one can always close over the query argument if it is needed.

Refs #5689

Issue #5689, opened 2015-01-18
Query::formatResults() receives only a single argument of type \Cake\Collection\Collection for hasOne/belongsTo


ndm2:
I can't tell for sure whether this is a bug, but it's certainly unexpected jugding from the docs and the existing tests.

```php
public function testFormatResultsForHasOneAssociation()
{
	$table = TableRegistry::get('authors');
	$table->hasOne('articles');

	$articles = TableRegistry::get('articles');

	$articles->eventManager()->attach(function ($event, $query) {
		$query->formatResults(function ($results, $query) {
			$this->assertInstanceOf('\Cake\Datasource\ResultSetInterface', $results);
			$this->assertInstanceOf('\Cake\Database\Query', $query);

			return $results->map(function ($entity) {
				$entity->foo = 'bar';
				return $entity;
			});
		});
	}, 'Model.beforeFind');

	$result = $table->get(1, ['contain' => ['articles']]);

	$this->assertEquals('bar', $result->article->foo);
}

public function testFormatResultsForBelongsToAssociation()
{
	$table = TableRegistry::get('articles');
	$table->belongsTo('authors');

	$authors = TableRegistry::get('authors');

	$authors->eventManager()->attach(function ($event, \Cake\ORM\Query $query) {
		$query->formatResults(function ($results, $query) {
			$this->assertInstanceOf('\Cake\Datasource\ResultSetInterface', $results);
			$this->assertInstanceOf('\Cake\Database\Query', $query);

			return $results->map(function ($entity) {
				$entity->foo = 'bar';
				return $entity;
			});
		});
	}, 'Model.beforeFind');

	$result = $table->get(1, ['contain' => ['authors']]);

	$this->assertEquals('bar', $result->author->foo);
}

public function testFormatResultsForHasManyAssociation()
{
	$table = TableRegistry::get('authors');
	$table->hasMany('articles');

	$articles = TableRegistry::get('articles');

	$articles->eventManager()->attach(function ($event, $query) {
		$query->formatResults(function ($results, $query) {
			$this->assertInstanceOf('\Cake\Datasource\ResultSetInterface', $results);
			$this->assertInstanceOf('\Cake\Database\Query', $query);

			return $results->map(function ($entity) {
				$entity->foo = 'bar';
				return $entity;
			});
		});
	}, 'Model.beforeFind');

	$result = $table->get(1, ['contain' => ['articles']]);

	$this->assertEquals('bar', $result->articles[0]->foo);
}

[...]

PR #14864, opened 2020-08-01
4.next - Add support for receiving the query object in result formatter callables

ndm2:
Just like with other query builder methods, I'd find it handy to receive the query object as an argument in the results formatter callable. 

This would be especially helpful in situations where the result formatter stems from the query builder for a joined association, which can make it hard to impossible to get access to the actual/source query object, say for example in a `belongsTo` scenario where the result formatter is being defined in a finder or a `beforeFind` callback in the associated table class.

The specific situation where I stumbled over this, was when I needed to inject additional results in a formatter of a `belongsTo` association finder, and needed to know whether the source query has hydration and results casting enabled/disabled, so that I could match that and query and inject the additional results in the expected format.

### DIFF
--- a/src/Datasource/QueryTrait.php
+++ b/src/Datasource/QueryTrait.php
@@ -400,7 +400,7 @@ trait QueryTrait
      *
      * ```
      * $query->formatResults(function ($results, $query) {
-     *     return $results->map(function ($row) {
+     *     return $results->map(function ($row) use ($query) {
      *         $data = [
      *             'bar' => 'baz',
      *         ];
@@ -416,13 +416,14 @@ trait QueryTrait
      * });
      * ```
      *
-     * Retaining access to the original query instance by inheriting the query variable:
+     * Retaining access to the association target query instance of joined associations,
+     * by inheriting the contain callback's query argument:
      *
      * ```
      * // Assuming a `Articles belongsTo Authors` association that uses the join strategy
      *
      * $articlesQuery->contain('Authors', function ($authorsQuery) {
-     *     return $query->formatResults(function ($results, $query) use ($authorsQuery) {
+     *     return $authorsQuery->formatResults(function ($results, $query) use ($authorsQuery) {
      *         // Here `$authorsQuery` will always be the instance
      *         // where the callback was attached to.
      *
--- a/src/Datasource/QueryTrait.php
+++ b/src/Datasource/QueryTrait.php
@@ -356,30 +356,85 @@ trait QueryTrait
      * Registers a new formatter callback function that is to be executed when trying
      * to fetch the results from the database.
      *
-     * Formatting callbacks will get a first parameter, an object implementing
-     * `\Cake\Collection\CollectionInterface`, that can be traversed and modified at will.
+     * If the second argument is set to true, it will erase previous formatters
+     * and replace them with the passed first argument.
      *
      * Callbacks are required to return an iterator object, which will be used as
      * the return value for this query's result. Formatter functions are applied
      * after all the `MapReduce` routines for this query have been executed.
      *
-     * If the second argument is set to true, it will erase previous formatters
-     * and replace them with the passed first argument.
+     * Formatting callbacks will receive two arguments, the first one being an object
+     * implementing `\Cake\Collection\CollectionInterface`, that can be traversed and
+     * modified at will. The second one being the query instance on which the formatter
+     * callback is being applied.
      *
-     * ### Example:
+     * Usually the query instance received by the formatter callback is the same query
+     * instance on which the callback was attached to, except for in a joined
+     * association, in that case the callback will be invoked on the association source
+     * side query, and it will receive that query instance instead of the one on which
+     * the callback was originally attached to - see the examples below!
+     *
+     * ### Examples:
+     *
+     * Return all results from the table indexed by id:
      *
      * ```
-     * // Return all results from the table indexed by id
      * $query->select(['id', 'name'])->formatResults(function ($results) {
-     *   return $results->indexBy('id');
+     *     return $results->indexBy('id');
      * });
+     * ```
      *
-     * // Add a new column to the ResultSet
+     * Add a new column to the ResultSet:
+     *
+     * ```
      * $query->select(['name', 'birth_date'])->formatResults(function ($results) {
-     *   return $results->map(function ($row) {
-     *     $row['age'] = $row['birth_date']->diff(new DateTime)->y;
-     *     return $row;
-     *   });
+     *     return $results->map(function ($row) {
+     *         $row['age'] = $row['birth_date']->diff(new DateTime)->y;
+     *
+     *         return $row;
+     *     });
+     * });
+     * ```
+     *
+     * Add a new column to the results with respect to the query's hydration configuration:
+     *
+     * ```
+     * $query->formatResults(function ($results, $query) {
+     *     return $results->map(function ($row) {
+     *         $data = [
+     *             'bar' => 'baz',
+     *         ];
+     *
+     *         if ($query->isHydrationEnabled()) {
+     *             $row['foo'] = new Foo($data)
+     *         } else {
+     *             $row['foo'] = $data;
+     *         }
+     *
+     *         return $row;
+     *     });
+     * });
+     * ```
+     *
+     * Retaining access to the original query instance by inheriting the query variable:
+     *
+     * ```
+     * // Assuming a `Articles belongsTo Authors` association that uses the join strategy
+     *
+     * $articlesQuery->contain('Authors', function ($authorsQuery) {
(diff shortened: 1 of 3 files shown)


ndm2:
I'm not sure if it maybe could make sense for joined associations, to also pass the target query to the result formatter callable, that would for example allow to retrieve possible options that have been applied to the target query, and cannot be retrieved from the source query?

AFAICT `\Cake\ORM\Association::attachTo()` does not merge the options of the target query into the source query, possibly purposefully.

tests/TestCase/ORM/QueryTest.php
```diff
@@ -2221,6 +2221,64 @@ public function testFormatResults()
...
+    /**
+     * Tests that results formatters do receive the query object.
+     *
+     * @return void
+     */
+    public function testResultFormatterReceivesTheQueryObject()
+    {
+        $resultFormatterQuery = null;
+
+        $query = $this->getTableLocator()->get('authors')
```
othercorey, line 2233:
```suggestion
        $query = $this->getTableLocator()->get('Authors')
```

src/ORM/Association.php
```diff
@@ -1014,7 +1014,7 @@ protected function _formatAssociationResults(Query $query, Query $surrogate, arr
             }
             $extracted = new Collection($extracted);
             foreach ($formatters as $callable) {
-                $extracted = new ResultSetDecorator($callable($extracted));
+                $extracted = new ResultSetDecorator($callable($extracted, $query));
```
othercorey, line 1017: This `formatResults()` should now receive `$query` as an argument. We should update to use the new parameter.

tests/TestCase/ORM/QueryTest.php
```diff
@@ -2221,6 +2221,64 @@ public function testFormatResults()
...
+
+        $resultFormatterTargetQuery = null;
+        $resultFormatterSourceQuery = null;
+
+        $authors->getEventManager()->on(
+            'Model.beforeFind',
+            function ($event, Query $targetQuery) use (&$resultFormatterTargetQuery, &$resultFormatterSourceQuery) {
+                $resultFormatterTargetQuery = $targetQuery;
+
+                $targetQuery->formatResults(function ($results, $query) use (&$resultFormatterSourceQuery) {
```
othercorey, line 2264: Won't this behave differently for different association types?

ndm2: I'm not sure what exactly you're pointing to, but it would behave differently in so far as that for non-joined associations, the formatter would receive the association target query instead instead of source query.

I can add a test to show the different behavior.

othercorey: I clicked on the wrong line, but yes that's what I mean.

I think that might be confusing, but something we could possibly fix with clear documentation.

ndm2: Yeah, if accepted, I'll add some explanation in the `formatResults()` docblock once the code is finished.

ndm2: Wait a minute, I remember this section in the docs very differently: https://book.cakephp.org/4/en/orm/query-builder.html#adding-calculated-fields

I thought it would be a good opportunity to expand on it, but it actually already says that the second argument would be a query instance 🙃 I must be mixing things up, I remember this section as a short only recently ago.

ndm2:
Well look what I found, of course _I_ was the one who shot me in the foot:

* [Query::formatResults() receives only a single argument of type \Cake\Collection\Collection for hasOne/belongsTo](https://github.com/cakephp/cakephp/issues/5689)
* [Remove query as an argument to result formatter methods.](https://github.com/cakephp/cakephp/pull/5696)

ndm2:
> I'm not sure if it maybe could make sense for joined associations, to also pass the target query to the result formatter callable, that would for example allow to retrieve possible options that have been applied to the target query, and cannot be retrieved from the source query?

On a second thought after reading the arguments for why this feature was initially removed, one could of course always pull in the association target query into the closure via `use()`, then one would have both the source query and the target query.

@markstory @lorenzo I wonder what you think of it now that a query would always be passed?

As @othercorey mentioned it might be confusing when one would receive different query objects depending on the context where the formatters are being applied, even though from a purely technical point of view it seems to make sense that it would retrieve the query on which it was actually being used.

I guess another option would be to always was the query where the formatter was _initially_ applied, and in the context of a joined-association, pass the source query where formatter is finally being applied, as the third argument to the callable?

othercorey:
> ```
> callable(CollectionInterface $results, Query $query, Query $sourceQuery = null)
> ```

I think if we add the query to the callback, it needs to be something like this. It shouldn't default to null though. If they are the same, then both should be set to the source.

markstory:
Do we need the additional parameter if the resultFormatter can close over the original query? That seems like it would solve the use case you originally presented.

ndm2:
> Do we need the additional parameter if the resultFormatter can close over the original query? That seems like it would solve the use case you originally presented.

It would solve my use case, yes, with the current state of the PR one can easily obtain both the original query as well as the association source query.

The question is whether we would want to avoid the possible confusion of the passed query not-being the original query depending on the context in which the formatter is being used, ie non-join (original query) vs join (association source query)?

Let's look at two examples, first the context dependent query one:

```php
$originalQuery
    ->applyOptions(['foo' => 'bar'])
    ->formatResults(function ($results, $assocSourceOrOriginalQuery) use ($originalQuery) {
        if (!$assocSourceOrOriginalQuery->isHydrationEnabled()) {
            // ...
        }
        
        $originalOptions = $originalQuery->getOptions();
        if ($options['foo'] ?? null) {
            // ...
        }
        
        // ...
    });
```

And here the two queries variant, where the first one would always be the original query where the formatter was initially attached:

```php
$originalQuery
    ->applyOptions(['foo' => 'bar'])
    ->formatResults(function ($results, $originalQuery, $assocSourceOrOriginalQuery) {
        if (!$assocSourceOrOriginalQuery->isHydrationEnabled()) {
            // ...
        }
        
        $originalOptions = $originalQuery->getOptions();
        if ($options['foo'] ?? null) {
            // ...
        }
        
        // ...
    });
```

I could see how both variants have the potential to confuse people, depending on how aware they are about the behavior.

othercorey:
Are we interested in providing this? We'll have to provide both queries for consistency with any combination I think.

There would need to be clear documentation and better names.

markstory:
I'm not sure we have use cases that require it. The problems this would solve seem to be solvable by retaining references to queries in closures.

ndm2:
@markstory Do you mean in general, or are you referring to the "two query arguments" solution?

markstory:
I don't really understand when/why you would need both queries in the formatter. That scenario seems somewhat theoretical to me so far.

ndm2:
Well, I still can only think of the options scenario, ie if one wanted to access options that have been set on the original query. Other than that, I'm not sure...

othercorey:
Closing in favor of one-off closure capture.

ndm2:
But it's not possible to capture the actual query in a joined association scenario, that was the whole point to begin with 😉

othercorey:
It seemed like there was no support, but if you want to re-open to discuss again.

ndm2:
The doubt seemed to be about the two query arguments variant. @markstory ?

Personally I would be fine with the single argument variant, where the query object for that argument would not be the original query in the joined association scenario, ie the current state of the PR.

It think it should be fine if it's properly documented. I mean, people have lived the whole time without the query argument being present, despite it being documented as being present.

markstory:
Yes, I'm not a fan of the two query variant. The single query version is ok with me as we can document why the query is sometimes the joined query.

othercorey:
If we're going to include this in 4.2, we need to merge soon. This needs migration notes in the 4.2 docs.

othercorey:
@markstory If we don't want this for 4.2, going to close it.

src/Datasource/QueryTrait.php
```diff
@@ -356,30 +356,85 @@ public function getMapReducers(): array
...
+     * });
+     * ```
+     *
+     * Retaining access to the original query instance by inheriting the query variable:
+     *
+     * ```
+     * // Assuming a `Articles belongsTo Authors` association that uses the join strategy
+     *
+     * $articlesQuery->contain('Authors', function ($authorsQuery) {
+     *     return $query->formatResults(function ($results, $query) use ($authorsQuery) {
```
markstory, line 425:
```suggestion
     *     return $articlesQuery->formatResults(function ($results, $query) use ($authorsQuery) {
```

ndm2: Actually `$authorsQuery`, I hope that was more or less just a typo, and not a testament to this feature's potential to even confuse experienced users 🙃

markstory: Just muddy brain from a long day :)

PR #14867, opened 2020-08-02
Adds hal+json, hal+xml, and ld+json #14796

### LINKED ISSUE #14796: RFC to add additional MIME types (HAL and JSON-LD) to Cake\HTTP\Response MIME types
This is a (multiple allowed):

* [ ] bug
* [ ] enhancement
* [X] feature-discussion (RFC)

Add HAL and JSON-LD mime types to [Response](https://github.com/cakephp/cakephp/blob/master/src/Http/Response.php).

Proposal:

```php
        'hal+json' => ['application/hal+json', 'application/vnd.hal+json'],
        'hal+xml' => ['application/hal+xml', 'application/vnd.hal+xml'],
        'ld+json' => 'application/ld+json',
```

I am not sure whether indexes for `hal+json` and `hal+xml` should instead by expressed as `vnd.hal+json` and `vnd.hal+xml`. There might be some debate there.

Not having these in necessitates some userland hacks to modify the Response at runtime such as including this code during the bootstrap phase of a plugin: https://gist.github.com/cnizzardini/1cf144d4d66b9f94f7d55b1c99d7ce62

Particularly, my use-case is building a HalJsonView and HalXmlView (and eventually a JsonLdView). It seems silly to have to add these MIME types in at runtime, though maybe there is a better way that I don't know about. Regardless, these are common enough message formats to be included IMO.

Reference:
- https://www.iana.org/assignments/media-types/media-types.xhtml
- http://stateless.co/hal_specification.html 
- https://json-ld.org/

cnizzardini:
Add hal+json, hal+xml, and ld+json #14796

I decided against modifying ServerRequest detectors, that can be handled by the user easily during bootstrap.


src/Http/Response.php
```diff
@@ -161,6 +161,8 @@ class Response implements ResponseInterface
         'gz' => 'application/x-gzip',
         'bz2' => 'application/x-bzip',
         '7z' => 'application/x-7z-compressed',
+        'hal+json' => ['application/hal+json', 'application/vnd.hal+json'],
```
othercorey, line 164: Wouldn't this just be `hal` ?

src/Http/Response.php
```diff
@@ -169,6 +171,7 @@ class Response implements ResponseInterface
         'js' => 'application/javascript',
         'jsonapi' => 'application/vnd.api+json',
         'latex' => 'application/x-latex',
+        'ld+json' => 'application/ld+json',
```
othercorey, line 174: Wouldn't this just be `jsonld` ?

src/Http/Response.php
cnizzardini: There is a hal+xml specification too. How else should these be distinguished?

src/Http/Response.php
cnizzardini: Yes, I think that makes more sense.

src/Http/Response.php
othercorey: I never see `+` in a type name, but there doesn't seem to be any other name for hal objects. *shrug*

cnizzardini: Agreed, it's goofy. I am fine dropping hal+xml al together and just going with `'hal' => ['application/hal+json', 'application/vnd.hal+json']`.

cnizzardini:
Were we done here?

othercorey:
Not much we can do if we're supporting both. The name is what it is.

markstory:
@cnizzardini If there is more to be done, more names can be added in additional pull requests.

Issue #14845, opened 2020-07-28
Asserting email subject in EmailTrait


mrothauer:
* [x] enhancement
* CakePHP Version: 4.1.1

Hello,
I'm looking for the possibility of asserting the email **subject** in EmailTrait.

In the docs https://book.cakephp.org/4/en/core-libraries/email.html#email-testing, there is stated: "We want to check that the **subject** and body contain the user’s name:" - but the example does not assert the subject, it only uses "assertMailContainsText".

I digged deeper and found another misleading line of code:
https://github.com/cakephp/cakephp/blob/ffa1fa5715e3569d30a4e07d0ff8e6c035ef3781/src/TestSuite/EmailTrait.php#L153

It assumes, that "subject" can be asserted with the "$parameter" argument - but as the assertMethod uses "MailSentWith" it cannot work.

So my question is:
Is there any way to assert the subject of an email? And if not, is such a feature planned? I would suggest a new class "MailSubject", extending "MailConstraintBase".

I could make a PR if you want.

markstory:
Seems like we're missing subject in this list https://github.com/cakephp/cakephp/blob/ffa1fa5715e3569d30a4e07d0ff8e6c035ef3781/src/TestSuite/Constraint/Email/MailSentWith.php#L59 Would you be able to make a pull request to add that?

mrothauer:
Yes, I can make a PR for that but I'm afraid it's not that easy.
MailSentWith is not the appropriate class to use for an assertion of the email subject - of course it would work but "subject" has nothing to do with "SentWith".
I would suggest creating a new method `assertMailSubject` and new class `MailSubject` which handle the new feature - and don't add "subject" to the quoted array.

markstory:
@mrothauer Ok, a new constraint class works too.

PR #14873, opened 2020-08-04
4.next issue 14845 assert email subject

mrothauer:
ref #14845 
docs https://github.com/cakephp/docs/commit/46716c1caa1738c8bc4e996cfa2bc4cb8b08d512

### DIFF
--- a/src/TestSuite/Constraint/Email/MailSubjectContains.php
+++ b/src/TestSuite/Constraint/Email/MailSubjectContains.php
@@ -16,6 +16,8 @@ declare(strict_types=1);
  */
 namespace Cake\TestSuite\Constraint\Email;
 
+use InvalidArgumentException;
+
 /**
  * MailSubjectContains
  *
@@ -31,6 +33,11 @@ class MailSubjectContains extends MailConstraintBase
      */
     public function matches($other): bool
     {
+        if (!is_string($other)) {
+            throw new InvalidArgumentException(
+                'Invalid data type, must be a string.'
+            );
+        }
         $messages = $this->getMessages();
         foreach ($messages as $message) {
             $subject = $message->getOriginalSubject();
--- a/src/TestSuite/Constraint/Email/MailSubjectContains.php
+++ b/src/TestSuite/Constraint/Email/MailSubjectContains.php
@@ -60,7 +60,7 @@ class MailSubjectContains extends MailConstraintBase
         if ($this->at && isset($messageMembers[$this->at - 1])) {
             $messageMembers = [$messageMembers[$this->at - 1]];
         }
-        $result = join(PHP_EOL, $messageMembers);
+        $result = implode(PHP_EOL, $messageMembers);
 
         return PHP_EOL . 'was: ' . mb_substr($result, 0, 1000);
     }
--- a/src/TestSuite/Constraint/Email/MailSubjectContains.php
+++ b/src/TestSuite/Constraint/Email/MailSubjectContains.php
@@ -1,0 +1,82 @@
+<?php
+declare(strict_types=1);
+
+/**
+ * CakePHP(tm) : Rapid Development Framework (https://cakephp.org)
+ * Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
+ *
+ * Licensed under The MIT License
+ * For full copyright and license information, please see the LICENSE.txt
+ * Redistributions of files must retain the above copyright notice.
+ *
+ * @copyright     Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
+ * @link          https://cakephp.org CakePHP(tm) Project
+ * @since         4.1.2
+ * @license       https://opensource.org/licenses/mit-license.php MIT License
+ */
+namespace Cake\TestSuite\Constraint\Email;
+
+/**
+ * MailSubjectContains
+ *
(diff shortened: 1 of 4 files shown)


src/TestSuite/Constraint/Email/MailSubjectContains.php
```diff
@@ -0,0 +1,82 @@
...
+     * @param mixed $other Constraint check
+     * @return bool
+     */
+    public function matches($other): bool
+    {
+        $messages = $this->getMessages();
+        foreach ($messages as $message) {
+            $subject = $message->getOriginalSubject();
+
+            $other = preg_quote($other, '/');
```
dereuromark, line 39: You let "mixed" enter, but use it blindly as string. This can blow up
Maybe you want to either only accept string or assert it.

src/TestSuite/Constraint/Email/MailSubjectContains.php
```diff
@@ -0,0 +1,82 @@
...
+    {
+        $messageMembers = [];
+        $messages = $this->getMessages();
+        foreach ($messages as $message) {
+            $messageMembers[] = $message->getSubject();
+        }
+        if ($this->at && isset($messageMembers[$this->at - 1])) {
+            $messageMembers = [$messageMembers[$this->at - 1]];
+        }
+        $result = join(PHP_EOL, $messageMembers);
```
dereuromark, line 64: implode()

src/TestSuite/Constraint/Email/MailSubjectContains.php
mrothauer: I tried to change the method declaration to `public function matches(string $other): bool` it and got:
"bool should be compatible with PHPUnit\Framework\Constraint\Constraint::matches($other)"
what do you mean by "or assert it"

src/TestSuite/Constraint/Email/MailSubjectContains.php
mrothauer: done

src/TestSuite/Constraint/Email/MailSubjectContains.php
dereuromark: Then dont change signature, but add assert inside:

If !is_string throw exception

src/TestSuite/Constraint/Email/MailSubjectContains.php
```diff
@@ -0,0 +1,88 @@
...
+            throw new InvalidArgumentException(
+                'Invalid data type, must be a string.'
+            );
+        }
+        $messages = $this->getMessages();
+        foreach ($messages as $message) {
+            $subject = $message->getOriginalSubject();
+
+            $other = preg_quote($other, '/');
+            if (preg_match("/$other/", $subject) > 0) {
```
dereuromark, line 46: We usually use simple `'` where possible.

dereuromark: Since this method can also return false, the > comparison is not good.
Instead, just use if(preg_match()) here

dereuromark: I usually even use

    preg_match(..., $matches);
    if ($matches) {}

src/TestSuite/Constraint/Email/MailSubjectContains.php
```diff
@@ -0,0 +1,89 @@
...
+        if (!is_string($other)) {
+            throw new InvalidArgumentException(
+                'Invalid data type, must be a string.'
+            );
+        }
+        $messages = $this->getMessages();
+        foreach ($messages as $message) {
+            $subject = $message->getOriginalSubject();
+
+            $other = preg_quote($other, '/');
```
markstory, line 45: The current assertion method documentation doesn't imply that regexp patterns are supported. Is using regex an intentional feature or should this constraint be using strpos() instead?

mrothauer: ´assertMailSubjectContains´ would work fine with `strpos()`, I can change that tomorrow if everyone is fine with that.

mrothauer:
While I was adding test cases for the new method assertMailSubjectContains in https://github.com/cakephp/cakephp/pull/14873/commits/0fcd67a800dcf7e69950151282a61407d8a4e3f8 I found out, that in EmailTestTrait there already was a test case for testing a subject, so my PR is pretty much useless.
https://github.com/cakephp/cakephp/pull/14873/commits/0fcd67a800dcf7e69950151282a61407d8a4e3f8

But anyway I think my implemention is better because

1) `$this->assertMailSentWith('Hello world', 'subject');` is dirty (mixing "SentWith" and "Subject"), SentWith IMHO should be used for to, cc, bcc and so on
2) As `subject` is not in the array https://github.com/mrothauer/cakephp/blob/0fcd67a800dcf7e69950151282a61407d8a4e3f8/src/TestSuite/Constraint/Email/MailSentWith.php#L59
this line is responsible for the assert: https://github.com/mrothauer/cakephp/blob/0fcd67a800dcf7e69950151282a61407d8a4e3f8/src/TestSuite/Constraint/Email/MailSentWith.php#L64
Therefore the assert needs to equal, my implementation can also compare parts of the string (strpos)

If you consider merging this PR, we should deprecate the "old" way of asserting email subjects.

Issue #14676, opened 2020-06-06
fetch('num') / fetchAll('num') not working when Query Log is enabled


swiffer:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: dev-4.next
* Platform and Target: YOUR WEB-SERVER, DATABASE AND OTHER RELEVANT INFO AND HOW THE REQUEST IS BEING MADE, HERE.

### What you did
```php
        $connection = ConnectionManager::get('default');
        $connection->execute("SELECT * FROM articles")->fetchAll('num');
```

### What happened
When Query Logging is enabled the fetch style seems to get reset to assoc somehow.

### What you expected to happen
num should be preserved here.

P.S. Remember, an issue is not the place to ask questions. You can use [Stack Overflow](https://stackoverflow.com/questions/tagged/cakephp)
for that or join the #cakephp channel on irc.freenode.net, where we will be more
than happy to help answer your questions.

Before you open an issue, please check if a similar issue already exists or has been closed before.

ADmad:
I am unable to reproduce this in my app using `4.next`.

swiffer:
I've added a demo_project with a Test that demonstrates the problem. Running `composer test` shows an error but shouldn't.

https://github.com/swiffer/demo_project

ADmad:
Thanks for the test case @swiffer.

The issue is caused by `BufferedStatement::rowCount()` which fetch 

https://github.com/cakephp/cakephp/blob/818eddcd6b96060b64447fab667de26b126fc3ee/src/Database/Statement/BufferedStatement.php#L258-L265

It occurs only with logging enabled because `LoggingStatement::execute()` call `rowCount()`
https://github.com/cakephp/cakephp/blob/818eddcd6b96060b64447fab667de26b126fc3ee/src/Database/Log/LoggingStatement.php#L67

So essentially when used buffered statement calling `rowCount()` before doing a fetch has the potential to cause unexpected results as it fetches the rows before a `fetchAll()` call is explicitly made.

The fix would be to change `LoggingStatement` to not call `rowCount()`.

This issue mostly likely exists in 4.0 and probably even 3.x too.

ADmad:
Another solution is if the buffer is already populated `BufferedStatement::fetchAll()` should check the fetch type and either convert the records in buffer to required type or re-fetch records.

swiffer:
Yes, possible I've seen this before

[...]

PR #14877, opened 2020-08-05
Fix SELECT queries with zero matching rows not being logged.

ADmad:
Refs #14685, #14676

// @ndm2

### DIFF
--- a/tests/TestCase/Database/ConnectionTest.php
+++ b/tests/TestCase/Database/ConnectionTest.php
@@ -1021,6 +1021,12 @@ class ConnectionTest extends TestCase
 
         $messages = Log::engine('queries')->read();
         $this->assertCount(1, $messages);
+
+        $st = $this->connection->execute('SELECT * FROM things WHERE id = 0');
+        $this->assertSame(0, $st->rowCount());
+
+        $messages = Log::engine('queries')->read();
+        $this->assertCount(2, $messages, 'Select queries without any matching rows should also be logged.');
     }
 
     /**
--- a/src/Database/Log/LoggingStatement.php
+++ b/src/Database/Log/LoggingStatement.php
@@ -81,8 +81,7 @@ class LoggingStatement extends StatementDecorator
         }
 
         if (preg_match('/^(?!SELECT)/i', $this->queryString)) {
-            $this->loggedQuery->numRows = $this->rowCount();
-            $this->_log();
+            $this->rowCount();
         }
 
         return $result;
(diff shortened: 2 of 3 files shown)


othercorey:
I think this is still going to trigger the scenario in the #14676 where there is a prefix in front of the SELECT clause:

```
preg_match('/^(?!SELECT)/i', $this->queryString)
```

ndm2:
That's pretty much what I was thinking about, and from quick local test Debug Kit shows buffered 0 result queries just fine now.

Switching `_log()` for `rowCount()` instead of moving counting into `_log()` avoids requiring to keep a state, that's nice 👍

tests/TestCase/Database/ConnectionTest.php
```diff
@@ -1021,6 +1021,12 @@ public function testLoggerDecoratorDoesNotPrematurelyFetchRecords()
 
         $messages = Log::engine('queries')->read();
         $this->assertCount(1, $messages);
+
+        $st = $this->connection->execute('SELECT * FROM things WHERE id = 0');
+        $this->assertSame(0, $st->rowCount());
+
+        $messages = Log::engine('queries')->read();
+        $this->assertCount(2, $messages, 'Select queries without any matching rows should also be logged.');
```
ndm2, line 1029: The problem with 0 results was in particular for buffered queries that do create result sets, that is queries made with the query builder. I think it would be good to have an additional test case for that?

ADmad: I added additional test case. 

For buffered queries without any matching rows `ResultSet` creates an `SplFixedFixed` of size `0`. Hence when you try to iterate it a `StatementInterface::fetch()` is never executed, which is why the query was never logged.

https://github.com/cakephp/cakephp/blob/edf991c10bd365981217d436b08d5612f538884d/src/ORM/ResultSet.php#L175-L178

Thus moving the logging to `LoggingStatment::rowCount()` (which is called by `ResultSet::count()`) fixes the issue.

othercorey:
This looks better to me although it reads weird. We still have an issue with SELECT statements with a prefix like the example in #14676:

```
'WITH final_articles AS (SELECT * FROM articles) SELECT * FROM final_articles WHERE title = ?'
```

othercorey (approved):

ADmad:
I have never used such type of queries. Whoever uses them can make a PR to update the regex used in `LoggingStatment::execute()` :slightly_smiling_face:.

PR #14880, opened 2020-08-06
Update issue template to more useful links

othercorey:


dereuromark:
Those details need to go into a comment section
Otherwise some genius users will just add their text on top and we will have tons of these including links opened.

othercorey:
We should just use the new issue templates.

PR #14884, opened 2020-08-08
Added connection retry logic to Driver

### LINKED ISSUE #14875: SQL Azure transient connection error when server unavailable
This is a (multiple allowed):

* [x] bug
* [ ] enhancement

* CakePHP Version: 4.1.1
* Platform and Target: Windows, IIS (Azure App Service), SQL Server Azure, HTTPS

### What you did
I'm using SQL Server on Azure (not by my choice) and it works fine for our application, but we recently tried using an on-demand instance for the database to reduce costs. There's no issue with the application, except when the database is waking up from pausing. I currently set it up to pause after 1 hour of inactivity. When connecting after the database was 'paused', SQL Server throws an error saying to basically check again in a few moments.

I tried to get around this error by setting the `loginTimeout` option on the database connection to 600 (10 minutes), hoping CakePHP could wait until the server is available or retry the connection at some point.

### What happened
The following error is thrown by the SQL Server driver when the database is still waking up: 
`[PDOException] SQLSTATE[08001]: [Microsoft][ODBC Driver 17 for SQL Server]TCP Provider: Timeout error [258]`

And the following can also be thrown: 
```Error: [Cake\Database\Exception\MissingConnectionException] Connection to Sqlserver could not be established: SQLSTATE[HY000]: [Microsoft][ODBC Driver 17 for SQL Server][SQL Server]Database 'xxxxx' on server 'xxxxx.database.windows.net' is not currently available.  Please retry the connection later.  If the problem persists, contact customer support, and provide them the session tracing ID of '{xxxxx}'. in D:\home\site\wwwroot\vendor\cakephp\cakephp\src\Database\Driver.php on line 121

Caused by: [PDOException] SQLSTATE[HY000]: [Microsoft][ODBC Driver 17 for SQL Server][SQL Server]Database 'xxxxx' on server 'xxxxx.database.windows.net' is not currently available.  Please retry the connection later.  If the problem persists, contact customer support, and provide them the session tracing ID of '{xxxxx}'. in D:\home\site\wwwroot\vendor\cakephp\cakephp\src\Database\Driver.php on line 118
```

I noticed there is some connection retrying code in CakePHP, and I understand this is a pretty specific platform issue. I opened this issue to determine if this is something that either the framework can allow handling, if this is truly a bug (or unsupported), and if there is a workaround I can do. 

Thanks for all the great work on the framework!

othercorey:
Closes https://github.com/cakephp/cakephp/issues/14875

### DIFF
--- a/src/Database/Driver.php
+++ b/src/Database/Driver.php
@@ -17,7 +17,9 @@ declare(strict_types=1);
 namespace Cake\Database;
 
 use Cake\Core\App;
+use Cake\Core\Retry\CommandRetry;
 use Cake\Database\Exception\MissingConnectionException;
+use Cake\Database\Retry\ErrorCodeWaitStrategy;
 use Cake\Database\Schema\SchemaDialect;
 use Cake\Database\Schema\TableSchema;
 use Cake\Database\Statement\PDOStatement;
@@ -36,6 +38,11 @@ abstract class Driver implements DriverInterface
      * @var int|null Maximum alias length or null if no limit
      */
     protected const MAX_ALIAS_LENGTH = null;
+
+    /**
+     * @var int[] DB-specific error codes that allow connect retry
+     */
+    protected const RETRY_ERROR_CODES = [];
 
     /**
      * Instance of PDO.
@@ -82,6 +89,13 @@ abstract class Driver implements DriverInterface
     protected $_version;
 
     /**
+     * The last number of connection retry attempts.
+     *
+     * @var int
+     */
+    protected $connectRetries = 0;
+
+    /**
      * Constructor
      *
      * @param array $config The configuration for the driver.
@@ -110,13 +124,18 @@ abstract class Driver implements DriverInterface
      */
     protected function _connect(string $dsn, array $config): bool
     {
-        try {
-            $connection = new PDO(
+        $action = function () use ($dsn, $config) {
+            $this->setConnection(new PDO(
                 $dsn,
                 $config['username'] ?: null,
                 $config['password'] ?: null,
                 $config['flags']
-            );
+            ));
+        };
+
+        $retry = new CommandRetry(new ErrorCodeWaitStrategy(static::RETRY_ERROR_CODES, 5), 4);
+        try {
+            $retry->run($action);
         } catch (PDOException $e) {
             throw new MissingConnectionException(
                 [
@@ -126,8 +145,9 @@ abstract class Driver implements DriverInterface
                 null,
                 $e
             );
+        } finally {
+            $this->connectRetries = $retry->getRetries();
         }
-        $this->setConnection($connection);
 
         return true;
     }
@@ -484,6 +504,16 @@ abstract class Driver implements DriverInterface
     public function getMaxAliasLength(): ?int
     {
         return static::MAX_ALIAS_LENGTH;
+    }
+
+    /**
+     * Returns the number of connection retry attempts made.
+     *
+     * @return int
+     */
+    public function getConnectRetries(): int
+    {
+        return $this->connectRetries;
     }
 
     /**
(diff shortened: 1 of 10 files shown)


src/Database/Driver.php
```diff
@@ -486,6 +519,16 @@ public function getMaxAliasLength(): ?int
         return static::MAX_ALIAS_LENGTH;
     }
 
+    /**
+     * Returns the number of connection retry attempts made.
+     *
+     * @return int
+     */
+    public function getLastConnectRetries(): int
```
ADmad, line 527: This could be just `getConnectRetries()`.

othercorey: Done. I also didn't update the retry count const name in SqlServer.

src/Database/Driver.php
```diff
@@ -486,6 +519,16 @@ public function getMaxAliasLength(): ?int
         return static::MAX_ALIAS_LENGTH;
     }
 
+    /**
+     * Returns the number of connection retry attempts made.
+     *
+     * @return int
+     */
+    public function getConnectRetries(): int
```
ADmad, line 527: Would also need to add `@method` annotation to `DriverInterface` for this new method.

lorenzo:
We used to have a Retry class for this, which is used in Connectiin if I remember correctly. Would it be worth reusing it?

othercorey:
> We used to have a Retry class for this, which is used in Connectiin if I remember correctly. Would it be worth reusing it?

CommandRetry wraps `Exception` while Driver wraps `PDOException`. Is that ok? We could maybe create a strategy in `connect` and pass it to `_connect`. Just depends how much we want to do here.

othercorey:
I think `Connection::connect` double wraps `Driver::connect` to catch `Exception`.

lorenzo:
I think a custom strategy may be a good idea

ADmad:
Given the changes this should target next minor.

othercorey:
This implements connect retry as a CommandRetry strategy.

Updated CommandRetry to pull the max retries config from the strategy.

src/Database/Retry/ConnectWaitStrategy.php
```diff
@@ -0,0 +1,87 @@
...
+     */
+    public function shouldRetry(Exception $exception): bool
+    {
+        if (
+            $exception instanceof PDOException &&
+            (
+                in_array($exception->getCode(), $this->errorCodes) ||
+                (
+                    $exception->errorInfo &&
+                    in_array($exception->errorInfo[1], $this->errorCodes)
```
othercorey, line 76: This situation is caused by https://bugs.php.net/bug.php?id=64705. errorInfo is null when the exception is thrown from PDO construction.

src/Core/Retry/RetryStrategyInterface.php
```diff
@@ -24,12 +24,18 @@
  */
 interface RetryStrategyInterface
 {
+    /**
+     * Returns the maximum number of retries allowed.
+     *
+     * @return int
+     */
+    public function getMaxRetries(): int;
```
markstory, line 32: Normally adding to interfaces is not backwards compatible. I doubt that this interface has been implemented by userland code though.

othercorey: `Cake\Database\Retry\ReconnectStrategy` is marked internal so I thought the whole mechanism was internal.

othercorey: Looking at the Orcale driver - the only other code I can think of to customize this - doesn't seem to have any strategy code.

othercorey:
Re-implemented without breaking BC of CommandRetry and RetryStrategyInterface.

markstory (approved):
Looks good to me.

othercorey:
Should this be for 4.1 patch for 4.2?

markstory:
@othercorey I think this can fit in 4.1. It will be a no op for most people and no interfaces changed.

PR #14885, opened 2020-08-08
Add phpunit 9 support (to allow testing on PHP 8)

ADmad:

### DIFF
--- a/src/Validation/Validation.php
+++ b/src/Validation/Validation.php
@@ -96,6 +96,13 @@ class Validation
      */
     public const COMPARE_LESS_OR_EQUAL = '<=';
 
+    protected const COMPARE_STRING = [
+        self::COMPARE_EQUAL,
+        self::COMPARE_NOT_EQUAL,
+        self::COMPARE_SAME,
+        self::COMPARE_NOT_SAME,
+    ];
+
     /**
      * Datetime ISO8601 format
      *
@@ -315,19 +322,21 @@ class Validation
      * Used to compare 2 numeric values.
      *
      * @param string|int $check1 The left value to compare.
-     * @param string $operator Can be either a word or operand
-     *    is greater >, is less <, greater or equal >=
-     *    less or equal <=, is less <, equal to ==, not equal !=
+     * @param string $operator Can be one of following operator strings:
+     *   '>', '<', '>=', '<=', '==', '!=', '===' and '!=='. You can use one of
+     *   the Validation::COMPARE_* constants.
      * @param string|int $check2 The right value to compare.
      * @return bool Success
      */
     public static function comparison($check1, string $operator, $check2): bool
     {
-        if ((float)$check1 != $check1) {
+        if (
+            (!is_numeric($check1) || !is_numeric($check2)) &&
+            !in_array($operator, static::COMPARE_STRING)
+        ) {
             return false;
         }
 
-        $operator = str_replace([' ', "\t", "\n", "\r", "\0", "\x0B"], '', strtolower($operator));
         switch ($operator) {
             case static::COMPARE_GREATER:
                 if ($check1 > $check2) {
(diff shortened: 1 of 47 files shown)


othercorey:
```
3) Cake\Test\TestCase\Database\Type\DateTypeTest::testMarshal with data set #9 ('2014-02-14', Cake\Chronos\Date Object (...))
Failed asserting that two DateTime objects are equal.
--- Expected
+++ Actual
@@ @@
-2014-02-14T00:00:00.000000+0000
+2014-02-14T07:42:50.000000+0000
```

This looks like a fun failure.

ADmad:
Errors on PHP 8:

- > Function openssl_free_key() is deprecated

    I couldn't find any info regarding this. I tried replacing it with `openssl_pkey_free()` which is apparently it's alias but that function too gives deprecation error. So nothing sure if there's some alternative or we just need to drop it's usage for PHP 8.

- > Function libxml_disable_entity_loader() is deprecated

    Based on https://github.com/php/php-src/pull/5867 and https://github.com/symfony/symfony/pull/37763 loading external entities is disabled by default since v2.9. So the solution is to check libxml version and use the method only for < 2.9. For higher versions use `LIBXML_NOENT` flag instead of enable.

- PHP version constraint for https://github.com/auraphp/Aura.Intl/blob/3.x/composer.json#L23 would have to be bumped. Aura project seem to be dead. I'll make a PR and see how it goes.

ADmad:
Last commit to Aura.Intl was on Jan 20, 2017. Even running it's test suite with a recent phpunit version would be non trivial. I think it would be better if we just copy the classes we need and move on.

othercorey:
>   I couldn't find any info regarding this. I tried replacing it with `openssl_pkey_free()` which is apparently it's alias but that function too gives deprecation error. So nothing sure if there's some alternative or we just need to drop it's usage for PHP 8.

>> The openssl_pkey_free() function is deprecated and no longer has an effect,
 instead the OpenSSLAsymmetricKey instance is automatically destroyed if it is no
 longer referenced.

https://github.com/php/php-src/blob/master/UPGRADING#L397

ADmad:
> https://github.com/php/php-src/blob/master/UPGRADING#L397

Good, so just need to wrap the call in a PHP version check.

othercorey:
We can commit this with failures allowed so people can test fixes or will have to wait work on them here.

ADmad:
On #14885 is merged and forwarded to `4.next` the build on PHP 8 will start passing so there's no need to modify the travis matrix.

src/Validation/Validation.php
```diff
@@ -315,19 +322,21 @@ public static function numElements($check, string $operator, int $expectedCount)
      * Used to compare 2 numeric values.
      *
      * @param string|int $check1 The left value to compare.
-     * @param string $operator Can be either a word or operand
-     *    is greater >, is less <, greater or equal >=
-     *    less or equal <=, is less <, equal to ==, not equal !=
+     * @param string $operator Can be one of following operator strings:
+     *   '>', '<', '>=', '<=', '==', '!=', '===' and '!=='. You can use one of
+     *   the Validation::COMPARISON_* constants.
```
garas, line 327:
```suggestion
     *   the Validation::COMPARE_* constants.
```

.travis.yml
```diff
@@ -64,7 +61,7 @@ install:
       if [[ $PREFER_LOWEST == 1 ]]; then
         composer update --prefer-lowest --prefer-stable --no-interaction
       elif [[ $TRAVIS_PHP_VERSION == 'nightly' ]]; then
-        composer install --no-interaction --ignore-platform-reqs
```
othercorey, line 67: What does no interaction cause with php 8?

othercorey (approved):
Looks ok to me.

ADmad:
This is ready to merge now. 

The only remaining failure for DateTimeType marshalling needs a fix in Chronos (which isn't an easy fix) and dealing with the warnings for usage of `at()` for mocks will take significant amount of effort given the number of usages. Perhaps this would be a good time to start switching to Mockery instead.

PR #14886, opened 2020-08-08
Xml entity loader

ADmad:
With libxml >= 2.9.0, loading of external entities is disabled,
even if libxml_disable_entity_loader(false) is called.
Instead the flag LIBXML_NOENT needs to be used to enable it.

See https://github.com/php/php-src/pull/5867


othercorey:
I doubt there is a way to test this, right?

ADmad:
There's no way I am going to even try to add tests for this :).

othercorey (approved):
Seems like entity loading has been discouraged for a while but we have the option.

PR #14888, opened 2020-08-08
Replace classes/interfaces of Aura.Intl with our own.

ADmad:
AuraPHP project is practically dead and Aura.Intl package hasn't had any updates since 2017.


ADmad:
- Do we want to consider providing class alias for common used classes like `Package` to make upgrade easier? For e.g. [DbMessagesLoader](https://github.com/ADmad/cakephp-i18n/blob/master/src/I18n/DbMessagesLoader.php) of my I18n plugin uses `Aura\Intl\Pacakage`.

- Should I add argument/return typehints that are missing?

src/I18n/Exception/CannotFormat.php
```diff
@@ -0,0 +1,25 @@
...
+/**
+ * CakePHP(tm) : Rapid Development Framework (https://cakephp.org)
+ * Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
+ *
+ * Licensed under The MIT License
+ * For full copyright and license information, please see the LICENSE.txt
+ * Redistributions of files must retain the above copyright notice.
+ *
+ * @copyright     Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
+ * @copyright     Copyright (c) 2017-2020, Aura for PHP
```
ADmad, line 13: Is this copyright tag acceptable? Ref https://github.com/auraphp/Aura.Intl/blob/3.x/LICENSE#L3

src/I18n/Exception/CannotFormat.php
```diff
@@ -0,0 +1,25 @@
...
+ * @link          https://cakephp.org CakePHP(tm) Project
+ * @since         4.2.0
+ * @license       https://opensource.org/licenses/mit-license.php MIT License
+ */
+namespace Cake\I18n\Exception;
+
+/**
+ * Could not format a message.
+ */
+class CannotFormat extends Exception
```
garas, line 23: Exception -> I18nException
CannotFormat ->CannotFormatException
FormatterNotMapped -> FormatterNotMappedException

src/I18n/Exception/CannotFormat.php
garas:
```suggestion
 * @copyright     Copyright (c) 2017, Aura for PHP
```
No changes after 2017, so 2020 is inaccurate, especially if range 2017-2020 specified

othercorey: A range isn't necessary for copyright. We can just declare the date it was changed.

othercorey:
>     * Do we want to consider providing class alias for common used classes like `Package` to make upgrade easier? For e.g. [DbMessagesLoader](https://github.com/ADmad/cakephp-i18n/blob/master/src/I18n/DbMessagesLoader.php) of my I18n plugin uses `Aura\Intl\Pacakage`.

That gets dangerously close to just forking the repository. Do we want to do that?

src/I18n/Exception/Exception.php
```diff
@@ -0,0 +1,27 @@
...
+ * @link          https://cakephp.org CakePHP(tm) Project
+ * @since         4.2.0
+ * @license       https://opensource.org/licenses/mit-license.php MIT License
+ */
+namespace Cake\I18n\Exception;
+
+use Cake\Core\Exception\Exception as CoreException;
+
+/**
+ * Could not format a message.
```
markstory, line 23: Is this comment still accurate?

src/I18n/FormatterLocator.php
```diff
@@ -0,0 +1,88 @@
...
+ * @copyright     Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
+ * @copyright     Copyright (c) 2017-2020, Aura for PHP
+ * @link          https://cakephp.org CakePHP(tm) Project
+ * @since         4.2.0
+ * @license       https://opensource.org/licenses/mit-license.php MIT License
+ */
+namespace Cake\I18n;
+
+/**
+ * A ServiceLocator implementation for loading and retaining formatter objects.
```
markstory, line 21: Should this class be marked as internal?

src/I18n/PackageLocator.php
```diff
@@ -0,0 +1,99 @@
...
+ * @link          https://cakephp.org CakePHP(tm) Project
+ * @since         4.2.0
+ * @license       https://opensource.org/licenses/mit-license.php MIT License
+ */
+namespace Cake\I18n;
+
+use Cake\I18n\Exception\Exception;
+
+/**
+ * A ServiceLocator implementation for loading and retaining translator objects.
```
markstory, line 23: Should this class be marked as internal?

src/I18n/Translator.php
```diff
@@ -14,27 +14,91 @@
  */
 namespace Cake\I18n;
 
-use Aura\Intl\Translator as BaseTranslator;
-
 /**
- * Provides missing message behavior for CakePHP internal message formats.
- *
- * @internal
+ * Translator to translate the message.
```
markstory, line 18: Should this class stay internal? It was internal before.

ADmad:
@markstory To begin with I simply made 1:1 copy of Aura's classes and ensure our test suite was passing. But as you pointed out `Translator` and other locator classes could be made internal. They would allow us to remove the locator interfaces too and it doesn't make much sense to have interfaces for internal classes. I could probably also simply `TranslatorRegistry` a bit then and drop classes like `TranslatorFactory` (our other registries themselves have method to create object instances).

ADmad:
> That gets dangerously close to just forking the repository. Do we want to do that?

Not really, my only aim was to make 4.2 upgrade easy and try to maintain our BC promise for minors. But I guess we could make a small exception in this case.

othercorey:
If we accept an Aura.Intl interface somewhere, then I'd say we should provide support. Otherwise, users will probably need to figure out a php 8 solution long-term on their own.

ADmad:
> If we accept an Aura.Intl interface somewhere, then I'd say we should provide support

Currently our users doen't really need to use those interfaces anywhere directly. Everything is tucked away under `Translator` which is marked as `@internal`. For e.g. the only Aura.Intl class my plugin requires to use directly is the `Package` class.

ADmad:
`TranslatorRegistry` does provide access to the locators but I doubt anyone accesses them to manipulate stuff.

othercorey:
Well using a getter is different that accepting an instance. We shouldn't need to alias for that.

markstory:
Should `composer.json` also be updated?

ADmad:
> Should composer.json also be updated?

Done.

I have simplified the internals of `TranslatorRegistry` and reduced Closure nesting.

To avoid having to explicitly packages for each domain-locale pair our `TranslatorRegistry` used loader classes which are used as fallbacks to get package instances. We also fallback domain feature so that if a message id is not found in particular domain name it's fetched from `default` domain. These features let to lost of closure nesting since Aura's `PackageLocator` class didn't allow directly setting `Package` instance, it only accepted callables which returned package instances. 

So I added a new method `setupPackage()` which allows directly setting `Package` instances to our updated package locator. This considerably simplified things.

ADmad:
This is ready for final review.

ADmad:
I'll merge this tomorrow if there is no further feedback.

src/I18n/FormatterLocator.php
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    /**
+     * Sets a formatter into the registry by name.
+     *
+     * @param string $name The formatter name.
+     * @param callable $spec A callable that returns a formatter object.
+     * @return void
+     */
+    public function set(string $name, callable $spec): void
```
othercorey, line 64: Are you still pushing to change callables closures?

src/I18n/Package.php
```diff
@@ -0,0 +1,164 @@
...
+
+/**
+ * Message Catalog
+ */
+class Package
+{
+    /**
+     * Message keys and translations in this package.
+     *
+     * @var array
```
othercorey, line 28:
```suggestion
     * @var string[]
```

src/I18n/Package.php
```diff
@@ -0,0 +1,164 @@
...
+    ) {
+        $this->formatter = $formatter;
+        $this->fallback = $fallback;
+        $this->messages = $messages;
+    }
+
+    /**
+     * Sets the messages for this package.
+     *
+     * @param array $messages The messages for this package.
```
othercorey, line 67:
```suggestion
     * @param string[] $messages The messages for this package.
```

src/I18n/Package.php
```diff
@@ -0,0 +1,164 @@
...
+     */
+    public function addMessage(string $key, string $message): void
+    {
+        $this->messages[$key] = $message;
+    }
+
+    /**
+     * Adds new messages for this package.
+     *
+     * @param array $messages The messages to add in this package.
```
othercorey, line 90:
```suggestion
     * @param string[] $messages The messages to add in this package.
```

src/I18n/Package.php
```diff
@@ -0,0 +1,164 @@
...
+     */
+    public function addMessages(array $messages): void
+    {
+        $this->messages = array_merge($this->messages, $messages);
+    }
+
+    /**
+     * Gets the messages for this package.
+     *
+     * @return array
```
othercorey, line 101:
```suggestion
     * @return string[]
```

src/I18n/Package.php
```diff
@@ -0,0 +1,164 @@
...
+    public function getMessages(): array
+    {
+        return $this->messages;
+    }
+
+    /**
+     * Gets the message of the given key for this package.
+     *
+     * @param string $key the key of the message to return
+     * @return mixed The message translation string, or false if not found.
```
othercorey, line 112:
```suggestion
     * @return string|false The message translation string, or false if not found.
```

src/I18n/Translator.php
```diff
@@ -118,4 +191,14 @@ protected function resolveContext(string $key, array $message, array $vars)
 
         return $message['_context'][$context];
     }
+
+    /**
+     * An object of type Package
```
othercorey, line 196:
```suggestion
     * Returns the translator package
```

src/I18n/TranslatorRegistry.php
```diff
@@ -72,40 +92,74 @@ class TranslatorRegistry extends TranslatorLocator
...
+     *
+     * @return string
+     */
+    public function getLocale(): string
+    {
+        return $this->locale;
+    }
+
+    /**
+     * An object of type PackageLocator
```
othercorey, line 144:
```suggestion
     * Returns the translator packages
```

src/I18n/PackageLocator.php
```diff
@@ -0,0 +1,124 @@
...
+
+    /**
+     * Sets a Package loader.
+     *
+     * @param string $name The package name.
+     * @param string $locale The locale for the package.
+     * @param callable|\Cake\I18n\Package $spec A callable that returns a package or Package instance.
+     * @return void
+     */
+    public function set(string $name, string $locale, $spec): void
```
othercorey, line 71: I guess this is required for compatibility for `set` and `setPackage` are confusingly similar.

src/I18n/TranslatorRegistry.php
```diff
@@ -227,54 +299,26 @@ public function useFallback(bool $enable = true): void
...
-    {
-        $loader = $this->_loaders[$name]($name, $locale);
-        $package = $loader;
-
-        if (!is_callable($loader)) {
-            $loader = function () use ($package) {
-                return $package;
-            };
+        if ($package->getFallback()) {
+            return $package;
```
othercorey, line 311: Why would the setter return the package being set?

src/I18n/PackageLocator.php
ADmad: Initially I had thought of adding a separate method for adding `Package` instance so that the argument could be typehinted for both methods but later just updated `set()` to accept both types to keep code DRY and forgot to remove the `setPackage()` method.

What does everyone prefer? Single method on two separate ones? I can update accordingly.

src/I18n/TranslatorRegistry.php
ADmad: The fallback is meant to be set on the package object itself. So if it already has a fallback set nothing has to be done and same object is returned back without any modification.

src/I18n/FormatterLocator.php
ADmad: I am but method will practically never be used. The 2 formatters we have are already setup through the locator's constructor and I don't really see anyone needed to a new formatter type.

That said even using callables which return formatter instance is not necessary. We can just have a formatter name to class name making and the locator and create and return an instance.

src/I18n/TranslatorRegistry.php
othercorey: Sure. I just mean it's strange that the setter is returning anything.

othercorey:
Looks ok to me although I am not very familiar with the use of packages. Anything else we can just re-adjust in cakephp 5.

PR #14891, opened 2020-08-11
Ensure plugin path config is loaded for each PluginCollection::findPath() call.

### LINKED ISSUE #14889: AuditStash not found in Testsuite after upgrading to 4.1.2
This is a (multiple allowed):

* [x] bug
* [ ] enhancement

* CakePHP Version: 4.1.2
* Platform and Target: PHPUnit 8.5.8, PHP 7.4.9

### What you did
Upgrade to 4.1.2 (from 4.1.1) - PHPUnit throws a bunch of errors because AuditStash plugin can no longer be loaded.

- https://github.com/lorenzo/audit-stash
- https://github.com/cakephp/cakephp/commit/7fc5c93f454148bae6b1e8c2b731fd3f6b01aa03#comments

### What happened
Possibly related to Cake\Core\Exception\MissingPluginException: "Plugin AuditStash could not be found."
```
#0 ..\vendor\cakephp\cakephp\src\Core\PluginCollection.php(243): Cake\Core\PluginCollection->findPath('AuditStash')
#1 ..\vendor\cakephp\cakephp\src\Http\BaseApplication.php(117): Cake\Core\PluginCollection->create('AuditStash', Array)
#2 ..\src\Application.php(78): Cake\Http\BaseApplication->addPlugin('AuditStash')
```

ADmad:
This is necessary primarily for testing because the Configure::clear() call
in TestCase::tearDown() wipes out all configuration including plugin paths config.

Reverts #14860. Closes #14889.

### DIFF
--- a/src/Core/PluginCollection.php
+++ b/src/Core/PluginCollection.php
@@ -122,6 +122,11 @@ class PluginCollection implements Iterator, Countable
      */
     public function findPath(string $name): string
     {
+        // Ensure plugin config is loaded each time. This is necessary primarily
+        // for testing because the Configure::clear() call in TestCase::tearDown()
+        // wipes out all configuration including plugin paths config.
+        $this->loadConfig();
+
         $path = Configure::read('plugins.' . $name);
         if ($path) {
             return $path;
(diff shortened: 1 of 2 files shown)


othercorey (approved):

ADmad:
Searching for path is necessary only for plugins which don't have a `Plugin` class since `BasePlugin::getPath()` can return is own path. We should consider throwing warnings for plugins which don't have a `Plugin` class in next minor. This would allow us to get rid of `PluginCollection::findPath()` in future.

Issue #14775, opened 2020-07-08
SQL Server maximum of parameters


mcsknp:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 4.x
* Platform and Target: Nginx, PHP 7.4, SQL Server

### What you did
Export CSV by run select query with fetching hasMany data.
Result has over 2100 rows.

### What happened
Below error occurred on associated select query.
```
Error: [PDOException] SQLSTATE[IMSSP]: Tried to bind parameter number 2101.  SQL Server supports a maximum of 2100 parameters.
```

### What you expected to happen
No error occurred.

ADmad:
You are exceeding the limits set by your database server. What do you expect the framework to do here?

mcsknp:
I want to pass primary key values to associated query without adding any parameters.
Or adding an option to separate fetching associated data in many times.

othercorey:
Please give an example of the query you're using that has so many parameters.

mcsknp:
`Articles` has many `Comments`

- `articles` table
```
id
...
```

- `comments` table
```
id
article_id
...
```

- Sample code
```php
$query = $articlesTable->find()->contain(['Comments']);
foreach ($query as $row) {
    // ...
}
```

garas:
Your `articles` table has likely over 2000 entries. [HasMany](https://book.cakephp.org/4/en/orm/associations.html#hasmany-associations) default strategy is `select`, and CakePHP will generate 
```sql
SELECT * FROM articles;
SELECT * FROM comments WHERE article_id IN (1, 2, 3, 4, 5); // Inserted using parameters
```

You can [change strategy](https://book.cakephp.org/4/en/orm/retrieving-data-and-resultsets.html#fetching-with-the-subquery-strategy) to `subquery`:
```php
$articles->find()->contain([
    'Comments' => ['strategy' => 'subquery']
]);
```
Instead second query will be
```sql
SELECT * FROM comments WHERE article_id IN (SELECT id FROM articles);
```

mcsknp:
It's worked but it looks like query is becoming more slowly.

garas:
Does SQL query executes slowly or PHP/CakePHP?

To speed up SQL query, check if `comments.article_id` has index on it.

From CakePHP side, hydrating thousands of rows to entities might be slow. Add `disableHydration()` to get results as PHP arrays.

mcsknp:
[...]

PR #14893, opened 2020-08-11
Throw exception when exceeding maximum number of Sql Server parameters

othercorey:
Refs https://github.com/cakephp/cakephp/issues/14775

### DIFF
--- a/src/Database/Driver/Sqlserver.php
+++ b/src/Database/Driver/Sqlserver.php
@@ -192,17 +192,27 @@ class Sqlserver extends Driver
     public function prepare($query): StatementInterface
     {
         $this->connect();
+
+        $sql = $query;
         $options = [PDO::ATTR_CURSOR => PDO::CURSOR_SCROLL];
-        $isObject = $query instanceof Query;
-        /** @psalm-suppress PossiblyInvalidMethodCall */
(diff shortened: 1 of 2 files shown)


src/Database/Driver/Sqlserver.php
```diff
@@ -192,17 +193,22 @@ public function enabled(): bool
...
         $this->connect();
+
         $options = [PDO::ATTR_CURSOR => PDO::CURSOR_SCROLL];
-        $isObject = $query instanceof Query;
-        /** @psalm-suppress PossiblyInvalidMethodCall */
-        if ($isObject && $query->isBufferedResultsEnabled() === false) {
-            $options = [];
+        if ($query instanceof Query) {
+            if (count($query->getValueBinder()->bindings()) > 2100) {
+                throw new Exception(
```
ADmad, line 200: Using core exception feels a bit weird here but we don't have a generic database exception :(

othercorey: I don't know when to use PHP exceptions and when to use Cake exceptions.

markstory: We often use basic php exceptions when the class fits the situation and we don't have an need to provide a special error page.

othercorey:
Changed exception to InvalidArgumentException and updated explanation of the possible causes.

PR #14898, opened 2020-08-12
Improve email validation

markstory:
Accept addresses containing [] these are in use by GitHub bots. This might be a bad idea as `[]` isn't entirely legal in an email address unless it is in a quoted string.

### DIFF
--- a/src/Validation/Validation.php
+++ b/src/Validation/Validation.php
@@ -809,7 +809,7 @@ class Validation
 
         if ($regex === null) {
             // phpcs:ignore Generic.Files.LineLength
-            $regex = '/^[\p{L}0-9!#$%&\'*+\/=?^_`{|}~-]+(?:\.[\p{L}0-9!#$%&\'*+\/=?^_`{|}~-]+)*@' . self::$_pattern['hostname'] . '$/ui';
+            $regex = '/^[\p{L}0-9!#$%&\'*+\/=?^_`{|}\\[\\]~-]+(?:\.[\p{L}0-9!#$%&\'*+\/=?^_`{|}\\[\\]~-]+)*@' . self::$_pattern['hostname'] . '$/ui';
         }
         $return = static::_check($check, $regex);
         if ($deep === false || $deep === null) {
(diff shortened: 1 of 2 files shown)


jimbo2150:
Definitely support this as the RFC does state that bracket characters are allowed in the local/username of an email address. The validation library should only be concerned with checking characters and formatting of the address alone. The higher level email class should then be concerned with it being formatted properly (e.g. quoted) when it is an attribute of a message.

othercorey:
Would users need to validate github bot email addresses?

markstory:
> Would users need to validate github bot email addresses?

Perhaps as an application that consumes source code from GitHub. 

>  Is it possible to create an email like that anywhere else?

Yes, self hosted email servers let you configure anything you'd like.

othercorey:
I guess I'd like to see an example of this. I can't send to email addresses like this through mail. The google bots I see have [] in the name but not the email address.

I am asking because if I accepted an email address like this and then couldn't send emails to it through an external mailer, then I can't even notify the user of a problem.

jimbo2150:
I tested Gmail and Office 365 / Outlook and neither support it. Gmail shows the address in red (I assume meaning invalid) and shows an error if you try to send an email containing a bracket character. Outlook seems to have a JS validation that freaks out and tries to get you to write a name onto the address and duplicates some of the characters. Odd that it is in the spec but not supported.

As far as Github bot, the brackets are showing up in the address name (as far as I know you can include any visible characters in the name).

Not sure if other systems will accept. I will try to create an email address with a bracket in a web host later today and see if it will be accepted.

jimbo2150:
The web host does not allow brackets in the name. The only allowed characters are alphanumeric, underscores, hyphens, ampersands, and periods.

dereuromark:
Sounds like a ticket for GitHub asking them to adhere to standards in their email usage would be the better alternative then :)

jimbo2150:
Don't think it has anything to do with GitHub. As stated earlier, GitHub uses brackets in the email name, not the address.
Example: "github-actions[bot]" <notifications@github.com>

markstory:
Closing this then, if mail servers are generally not supporting this style of address it will negatively impact our end users who generally want to send emails, as their users will be able to input invalid email addresses.

PR #14901, opened 2020-08-13
4.x - Fix output buffers not being closed when the view triggers an exception.

ndm2:
When evaluating a template or caching a block, the view itself opens a buffer, and the templates/blocks that it evaluates might open buffers too. In case an exception happens, the opened buffers will not be closed, causing risky tests with PHPUnit, and possible further output to be swallowed unless the test itself closes/flushes the remaining buffers.

My thinking was that unclosed buffers aren't really a user error when caused by an exception, unlike when someone for example forgets to close an opened view block. However I'm still a little hesitant, because maybe there actually are situations where one would want such a problem to cause their tests to be marked as risky!?

### DIFF
--- a/src/View/View.php
+++ b/src/View/View.php
@@ -34,6 +34,7 @@ use Cake\View\Exception\MissingTemplateException;
 use InvalidArgumentException;
 use LogicException;
 use RuntimeException;
+use Throwable;
 
 /**
  * View, the V in the MVC triad. View interacts with Helpers and view variables passed
@@ -687,8 +688,20 @@ class View implements EventDispatcherInterface
         if ($result) {
             return $result;
         }
+
+        $bufferLevel = ob_get_level();
         ob_start();
-        $block();
+
+        try {
+            $block();
+        } catch (Throwable $exception) {
+            while (ob_get_level() > $bufferLevel) {
+                ob_end_clean();
+            }
+
+            throw $exception;
+        }
+
         $result = ob_get_clean();
 
         Cache::write($options['key'], $result, $options['config']);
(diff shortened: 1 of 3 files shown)


src/View/View.php
```diff
@@ -687,8 +687,20 @@ public function cache(callable $block, array $options = []): string
...
+
+        $bufferLevel = \ob_get_level();
         ob_start();
-        $block();
+
+        try {
+            $block();
+        } catch (\Throwable $exception) {
+            while (ob_get_level() > $bufferLevel) {
+                ob_end_clean();
```
markstory, line 698: What does this do to error pages that are triggered during view rendering? It looks like you'd end up with a half rendered view and then an error page after that.

ndm2: Good question. From a quick manual test it looks fine.

Is your concern that `ob_get_clean()` could cause output to be generated (it won't), or that discarding the buffers could cause the partially rendered data to be lost?

ndm2: I menr `ob_end_clean()` of course, sorry.

markstory: Ah I was conflating ob_get_clean() and ob_get_flush() which does flush the output buffer. Discarding the buffer on an exception will be safe.

othercorey: Not that we need to change this, but wouldn't this work as a `finally {}` block?

jimbo2150: The finally block is run regardless of if an exception is thrown or not. I believe having that run when no exception is thrown would destroy the content output from the view.

ndm2: It would work as in that it would close all opened buffers that way, but it would then always do that, ie also when no exception is being triggered. I didn't go that route because I thought that would mean that it's possibly fixing up and somewhat "hiding" user errors.

If it could be safely assumed that open buffers without exceptions are always user errors, then maybe always closing them and triggering an error, warning, or notice when no exception happened might be a good fit?

othercorey: > If it could be safely assumed that open buffers without exceptions are always user errors, then maybe always closing them and triggering an error, warning, or notice when no exception happened might be a good fit?

I think phpunit does that for us. This seems like it works best for now.

ndm2: It does, though we would have the opportunity to supply more information as to where the problem is located, and what it's possibly caused by. PHPUnit gives you no clue, and it only closes the buffers after the test method ran, so if you for example add debug output in your test method, it won't show up, which can be pretty annoying if you don't know what exactly is going on.

However I just noticed that if you for example leave a view block open (and therefore a buffer), `View::_render()` will trigger an exception telling you about that, and I certainly wouldn't want to interfere with that, so I guess I agree that it's probably best to leave it at that for now 🙃

Issue #14892, opened 2020-08-11
Setting filename for inline attachment in response is missing


mrothauer:
This is a (multiple allowed):

* [] bug
* [x] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 4.1.2

### What you did
In my response I want to send a file, but instead of forcing the download I want to set the Content-Disposition to "inline" so that the file (in this case a pdf) is opened in the browser and when the user decides to save the file, the specified filename should be used. This is not possible so far (or at least I did not find out how to do this).

A new method in Cake/Http/Response would help

```
public function withInlineAttachment(string $filename)
{
    return $this->withHeader('Content-Disposition', 'inline; filename="' . $filename . '"');
}
```

It would be nice, if this could be integrated in the existing method `withFile()`, 

Thank you!

markstory:
> It would be nice, if this could be integrated in the existing method withFile(),

What would the interface extension to `withFile` look like? Would something like 

```php
$this->response->withFile('file.pdf', ['as' => 'inline', 'name' => 'best-thing.pdf']);
```

Work? This would also make the `download => true` option an alias for `as => 'download'`. I don't think another boolean option is a good idea as then we have to handle the situation where both are set.

mrothauer:
Yes, your interface suggestion is working for my use case.

dereuromark:
I have been using https://github.com/dereuromark/cakephp-tools/blob/5753afad12101c14974fa98e1fbde8685f4bf8a7/src/Mailer/Message.php#L189 for 10 years now for embedded attachments etc. But that doesnt mean download true though. 
maybe inline means sth else than embedded here?

ADmad:
Isn't in browser display the default behavior if you don't force download? @mrothauer Have you tried just setting `'download' => false` in options of `withFile()`?

othercorey:
If you view a file inline, can you then save the file from the browser? Does the browser accept the content disposition when saving? Maybe that's what the user wants. 

I wonder if the behavior is different for mobile browsers.

mrothauer:
> Have you tried just setting 'download' => false in options of withFile()?
Yes, I tried that. Instead of the passed filename, the name of the action is used in the save-as dialog.

mrothauer:
> If you view a file inline, can you then save the file from the browser? Does the browser accept the content disposition filename when saving?
In Windows 10 / Chrome it works (with a PDF)
[...]

PR #14903, opened 2020-08-14
Added Response::withInline() to force content disposition to inline

othercorey:
Refs https://github.com/cakephp/cakephp/issues/14892

### DIFF
--- a/src/Http/Response.php
+++ b/src/Http/Response.php
@@ -1142,6 +1142,22 @@ class Response implements ResponseInterface
     }
 
     /**
+     * Create a new instnace with Content-Disposition header set.
+     *
+     * @param string|null $filename The name of the file the browser can download as after viewing file inline
+     * @return static
+     */
+    public function withInline(?string $filename = null)
+    {
+        $disposition = 'inline';
+        if ($filename) {
+            $disposition .= '; filename="' . $filename . '"';
+        }
+
+        return $this->withHeader('Content-Disposition', $disposition);
+    }
+
+    /**
      * Create a new response with the Content-Length header set.
      *
      * @param int|string $bytes Number of bytes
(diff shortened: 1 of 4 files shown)


src/Http/Response.php
```diff
@@ -1141,6 +1141,22 @@ public function withDownload(string $filename)
         return $this->withHeader('Content-Disposition', 'attachment; filename="' . $filename . '"');
     }
 
+    /**
+     * Create a new instnace with Content-Disposition header set.
```
dereuromark, line 1145:
```suggestion
     * Create a new instance with Content-Disposition header set.
```

src/Http/Response.php
```diff
@@ -1141,6 +1141,22 @@ public function withDownload(string $filename)
         return $this->withHeader('Content-Disposition', 'attachment; filename="' . $filename . '"');
     }
 
+    /**
+     * Create a new instance with Content-Disposition header set.
+     *
+     * @param string|null $filename The name of the file the browser can download as after viewing file inline
+     * @return static
+     */
+    public function withInline(?string $filename = null)
```
ADmad, line 1150: Can't we just add an `as` option to `Response::withFile()` as markstory had suggested? Not too thrilled about increasing the API surface for an uncommon use case. Mostly people just craft the URL to get the required filename.

ADmad: Also new features, especially API changes need to target next minor. Otherwise it becomes difficult for users to keep of when a feature was added.

othercorey: We can do that. Wasn't sure if there was value to just setting the disposition instead of requiring a stream using `withFile()`.

othercorey: The inline disposition is not about the filename - it is about better hinting to the browser to preview the file.

PR #14904, opened 2020-08-14
Ensure DateType marshalling works for both mutable and immutable types

othercorey:
This cleans up a few unrelated areas:

- Make sure setTime(0, 0, 0) is called for both DateTime and DateTimeImmutable
- Add missing return type hints to parse functions
- Ensure DateType marshal() is tested for both mutable and immutable types

### DIFF
--- a/src/Database/Type/DateType.php
+++ b/src/Database/Type/DateType.php
@@ -18,6 +18,7 @@ namespace Cake\Database\Type;
 
 use Cake\I18n\Date;
 use Cake\I18n\FrozenDate;
+use Cake\I18n\I18nDateTimeInterface;
 use DateTime;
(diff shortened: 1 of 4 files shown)


src/Database/Type/DateType.php
```diff
@@ -80,7 +81,9 @@ public function useMutable()
     public function marshal($value): ?DateTimeInterface
     {
         $date = parent::marshal($value);
-        if ($date instanceof DateTime) {
+        if ($date && !$date instanceof I18nDateTimeInterface) {
```
othercorey, line 84: This is the issue I noticed. When I18n isn't available (something we don't test), only mutable dates would be cleared properly.

PR #14905, opened 2020-08-15
Make phpunit methods protected as per docs

dereuromark:
Follow https://github.com/cakephp/phinx/pull/1862
Might also allow then plugins, apps to start using the documented visibility as the extending class can only be "wider", so no BC break if extending code is still public.
But it could not be protected if the core itself still has it public.


othercorey (approved):
This seems like a non-issue, but might as well fix it.

PR #14157, opened 2020-01-11
Fix url option in first() and last()

markstory:
Correct URL generation when the title is a string and the `url` option
is used.

Fixes #14155

PR #14907, opened 2020-08-15
Fix PaginatorHelper last & first with URL parameters

### LINKED ISSUE #14894: PaginatorHelper not merging URL parameters correctly
This is an apparent **bug** in CakePHP 3.8.12.

### What I did
I'm using the `PaginatorHelper` in CakePHP 3.8 and passing `'url' => ['?' => ['foo' => 'bar']]` into the options of its `first()`, `prev()`, `next()`, and `last()` buttons.

### What happened
`prev()` and `next()` are forming URLs as expected, but `first()` and `last()` are ignoring the query string parameter and not adding `'foo'` to the URL.

markstory:
Incorporate URL options when generating a single first/last link.

Fixes #14894

### DIFF
--- a/src/View/Helper/PaginatorHelper.php
+++ b/src/View/Helper/PaginatorHelper.php
@@ -1075,8 +1075,9 @@ class PaginatorHelper extends Helper
             }
         } elseif ($params['page'] > 1 && is_string($first)) {
             $first = $options['escape'] ? h($first) : $first;
+            $url = array_merge($options['url'], ['page' => 1]);
             $out .= $this->templater()->format('first', [
-                'url' => $this->generateUrl(['page' => 1], $options['model']),
+                'url' => $this->generateUrl($url, $options['model']),
                 'text' => $first,
             ]);
         }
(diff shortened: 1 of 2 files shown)


src/View/Helper/PaginatorHelper.php
```diff
@@ -1075,8 +1075,9 @@ public function first($first = '<< first', array $options = [])
             }
         } elseif ($params['page'] > 1 && is_string($first)) {
             $first = $options['escape'] ? h($first) : $first;
+            $url = array_merge($options['url'], ['page' => 1]);
```
othercorey, line 1078: Interestingly, this means you can control the page options, but I don't think that's going to cause any problems.

dereuromark:
I wonder when we fixed 4.x and forgot to backport it.

othercorey:
> I wonder when we fixed 4.x and forgot to backport it.

https://github.com/cakephp/cakephp/pull/14157

src/View/Helper/PaginatorHelper.php
markstory: Wouldn't the merge overwrite anything in options[url]?

othercorey: It won't overwrite 'page' or 'pageCount'. Should we pass $options['url'] as the 3rd parameter to generateUrl() like your other PR?

markstory: The signature of generateUrl() is different in 3.x and 4.x. In 4.x there is an addtional url parameter that is being used in the PR you referenced. I can't make a similar change here as it would require changing the public interface.

PR #14908, opened 2020-08-15
Make Route normalize HTTP method conditions

### LINKED ISSUE #14902: .json ext giving Missing Method
This is a (multiple allowed):

* [x] bug
* [ ] enhancement

* CakePHP Version:4.0.9
* Platform and Target: MAMP, PHP 7, MYSQL

### What you did
I have a relatively clean install of CakePHP 4. I have read all the docs related to JSON routing with an extension and using accept header in the AJAX call `"Accept", "application/json"`. Both methods give me a 404 missing method. I have posted the issue on both StackOverflow and CakePHP support forum but the help unfortunately didn't solve the issue. I am beginning to think it is an issue with CakePHP because I have done all in the docs.

https://discourse.cakephp.org/t/json-ext-giving-missing-method/8211

https://stackoverflow.com/questions/63282271/cakephp-json-ext-giving-missing-method

    Missing Method in StrategiesConditionsController
    Cake\Controller\Exception\MissingActionException
    
    Error The action conditions.json is not defined in StrategiesConditionsController
    
    Create StrategiesConditionsController::conditions.json() in file: src/Controller/StrategiesConditionsController.php.

From reading the doc a couple of times I am sure I have the routes correct.

routes.php

    <?php
    /**
     * Routes configuration.
     *
     * In this file, you set up routes to your controllers and their actions.
     * Routes are very important mechanism that allows you to freely connect
     * different URLs to chosen controllers and their actions (functions).
     *
     * It's loaded within the context of `Application::routes()` method which
     * receives a `RouteBuilder` instance `$routes` as method argument.
     *
     * CakePHP(tm) : Rapid Development Framework (https://cakephp.org)
     * Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
     *
     * Licensed under The MIT License
     *
     */
    
    use Cake\Http\Middleware\CsrfProtectionMiddleware;
    use Cake\Routing\Route\DashedRoute;
    use Cake\Routing\RouteBuilder;
    use Cake\Routing\Router;
    
    /*
     * The default class to use for all routes
     *
     * The following route classes are supplied with CakePHP and are appropriate
     * to set as the default:
     *
     * - Route
     * - InflectedRoute
     * - DashedRoute
     *
     * If no call is made to `Router::defaultRouteClass()`, the class used is
     * `Route` (`Cake\Routing\Route\Route`)
     *
     * Note that `Route` does not do any inflections on URLs which will result in
     * inconsistently cased URLs when used with `:plugin`, `:controller` and
     * `:action` markers.
     */
    /** @var \Cake\Routing\RouteBuilder $routes */
    $routes->setRouteClass(DashedRoute::class);
    
        $routes->setExtensions(['json']);
        $routes->resources('StrategiesConditions', [
            'map' => [
                'conditions' => [
                    'action' => 'conditions',
                    'method' => 'post'
                ]
            ],
            'only' => ['conditions']
         ]);
    });
    
    $routes->scope('/', function (RouteBuilder $builder) {
        // Register scoped middleware for in scopes.
        $builder->registerMiddleware('csrf', new CsrfProtectionMiddleware([
            'httpOnly' => true,
        ]));
    
        /*
         * Apply a middleware to the current route scope.
         * Requires middleware to be registered through `Application::routes()` with `registerMiddleware()`
         */
        $builder->applyMiddleware('csrf');
    
        /*
         * Here, we are connecting '/' (base path) to a controller called 'Pages',
         * its action called 'display', and we pass a param to select the view file
         * to use (in this case, templates/Pages/home.php)...
         */
        $builder->connect('/', ['controller' => 'Pages', 'action' => 'display', 'home']);
    
        /*
         * ...and connect the rest of 'Pages' controller's URLs.
         */
        $builder->connect('/pages/*', ['controller' => 'Pages', 'action' => 'display']);
    
        /*
         * Connect catchall routes for all controllers.
         *
         * The `fallbacks` method is a shortcut for
         *
         * ```
         * $builder->connect('/:controller', ['action' => 'index']);
         * $builder->connect('/:controller/:action/*', []);
         * ```
         *
         * You can remove these routes once you've connected the
         * routes you want in your application.
         */
        $builder->fallbacks();
    });
    
    Router::prefix('admin', function (RouteBuilder $routes) {
        $routes->connect('/', ['controller' => 'Users', 'action' => 'index']);
        $routes->fallbacks(DashedRoute::class);
    });
    
    /*
     * If you need a different set of middleware or none at all,
     * open new scope and define routes there.
     *
     * ```
     * $routes->scope('/api', function (RouteBuilder $builder) {
     *     // No $builder->applyMiddleware() here.
     *     // Connect API actions here.
     * });
     * ```
     */

strategiesConditionsController

    public function conditions()
    {
        $this->request->allowMethod(['post']);

        if ($this->request->is('ajax')) {
            $this->response = $this->response->withDisabledCache();
        }

        $strategy_id = $this->request->getData('strategy_id');
        
        $strategiesConditions = $this->StrategiesConditions->find('all', [
            'where' => ['strategy_id' => $strategy_id],
        ]);
        $this->viewBuilder()->setOption('serialize', ['strategiesConditions']);
    }

Cakephp console has 8 routes showing

    Route name	URI template	Defaults
    pages:display	/	
    {
        "controller": "Pages",
        "action": "display",
        "0": "home",
        "plugin": null
    }
    pages:display	/pages/*	
    {
        "controller": "Pages",
        "action": "display",
        "plugin": null
    }
    _controller:index	/{controller}	
    {
        "action": "index",
        "plugin": null
    }
    _controller:_action	/{controller}/{action}/*	
    {
        "plugin": null,
        "action": "index"
    }
    admin:users:index	/admin	
    {
        "controller": "Users",
        "action": "index",
        "prefix": "Admin",
        "plugin": null
    }
    admin:_controller:index	/admin/{controller}	
    {
        "action": "index",
        "prefix": "Admin",
        "plugin": null
    }
    admin:_controller:_action	/admin/{controller}/{action}/*	
    {
        "prefix": "Admin",
        "plugin": null,
        "action": "index"
    }
    strategiesconditions:conditions	/strategies-conditions/conditions	
    {
        "controller": "StrategiesConditions",
        "action": "conditions",
        "_method": "post",
        "plugin": null
    }

Update from the history tab in debug kit

    Request
    Routing Params
    controller StrategiesConditions
    action conditions.json
    pass (empty)
    plugin (null)
    _matchedRoute /{controller}/{action}/*
    _ext (null)
    Post data
    strategy_id 11
    Query string
    No querystring data.
    
    Cookie
    csrfToken dd12f852560a384d39206e511f1857f77f71da2eadb023a6c67ae346
    PHPSESSID 3d2jt7dgpo09af8b1908quljpn
    Matched Route
    template /{controller}/{action}/*

and the routes tab:

    Routes
    Toggle debugkit internal routes
    Route name	URI template	Defaults
    pages:display	/	
    {
        "controller": "Pages",
        "action": "display",
        "0": "home",
        "plugin": null
    }
    pages:display	/pages/*	
    {
        "controller": "Pages",
        "action": "display",
        "plugin": null
    }
    _controller:index	/{controller}	
    {
        "action": "index",
        "plugin": null
    }
    _controller:_action	/{controller}/{action}/*	
    {
        "plugin": null,
        "action": "index"
    }
    admin:users:index	/admin	
    {
        "controller": "Users",
        "action": "index",
        "prefix": "Admin",
        "plugin": null
    }
    admin:_controller:index	/admin/{controller}	
    {
        "action": "index",
        "prefix": "Admin",
        "plugin": null
    }
    admin:_controller:_action	/admin/{controller}/{action}/*	
    {
        "prefix": "Admin",
        "plugin": null,
        "action": "index"
    }
    strategiesconditions:conditions	/strategies-conditions/conditions	
    {
        "controller": "StrategiesConditions",
        "action": "conditions",
        "_method": "post",
        "plugin": null
    }

Request URL

http://localhost:8888/trading-journal/strategies-conditions/conditions.json

Application.php

            // Add routing middleware.
            // If you have a large number of routes connected, turning on routes
            // caching in production could improve performance. For that when
            // creating the middleware instance specify the cache config name by
            // using it's second constructor argument:
            // `new RoutingMiddleware($this, '_cake_routes_')`
            ->add(new RoutingMiddleware($this))

Thanks

markstory:
By normalizing the http method name to the correct case resource routes are a bit more ergonomic as case mistakes don't cause routes to fail matching.

Fixes #14902

### DIFF
--- a/src/Routing/Route/Route.php
+++ b/src/Routing/Route/Route.php
@@ -187,9 +187,9 @@ class Route
     {
         $methods = is_array($methods)
             ? array_map('strtoupper', $methods)
-            : [strtoupper($methods)];
+            : strtoupper($methods);
 
-        $diff = array_diff($methods, static::VALID_METHODS);
+        $diff = array_diff((array)$methods, static::VALID_METHODS);
         if ($diff !== []) {
             throw new InvalidArgumentException(
                 sprintf('Invalid HTTP method received. `%s` is invalid.', implode(', ', $diff))
--- a/src/Routing/Route/Route.php
+++ b/src/Routing/Route/Route.php
@@ -131,6 +131,12 @@ class Route
         $this->setExtensions((array)$this->options['_ext']);
         $this->setMiddleware((array)$this->options['_middleware']);
         unset($this->options['_middleware']);
+
+        if (isset($this->defaults['_method'])){
+            $this->defaults['_method'] = is_array($this->defaults['_method'])
+                ? array_map('strtoupper', $this->defaults['_method'])
+                : strtoupper($this->defaults['_method']);
+        }
     }
 
     /**
@@ -740,9 +746,10 @@ class Route
         if (empty($url['_method'])) {
             $url['_method'] = 'GET';
         }
+        $defaults = (array)$this->defaults['_method'];
         $methods = array_map('strtoupper', (array)$url['_method']);
         foreach ($methods as $value) {
-            if (in_array($value, (array)$this->defaults['_method'], true)) {
+            if (in_array($value, $defaults, true)) {
                 return true;
             }
         }
--- a/src/Routing/Route/Route.php
+++ b/src/Routing/Route/Route.php
@@ -132,7 +132,7 @@ class Route
         $this->setMiddleware((array)$this->options['_middleware']);
         unset($this->options['_middleware']);
 
-        if (isset($this->defaults['_method'])){
+        if (isset($this->defaults['_method'])) {
             $this->defaults['_method'] = is_array($this->defaults['_method'])
                 ? array_map('strtoupper', $this->defaults['_method'])
                 : strtoupper($this->defaults['_method']);
--- a/src/Routing/Route/Route.php
+++ b/src/Routing/Route/Route.php
@@ -133,9 +133,7 @@ class Route
         unset($this->options['_middleware']);
 
         if (isset($this->defaults['_method'])) {
-            $this->defaults['_method'] = is_array($this->defaults['_method'])
-                ? array_map('strtoupper', $this->defaults['_method'])
-                : strtoupper($this->defaults['_method']);
+            $this->defaults['_method'] = $this->normalizeAndValidateMethods($this->defaults['_method']);
         }
     }
 
@@ -174,16 +172,31 @@ class Route
      */
     public function setMethods(array $methods)
     {
-        $methods = array_map('strtoupper', $methods);
-        $diff = array_diff($methods, static::VALID_METHODS);
-        if ($diff !== []) {
-            throw new InvalidArgumentException(
-                sprintf('Invalid HTTP method received. %s is invalid.', implode(', ', $diff))
-            );
-        }
-        $this->defaults['_method'] = $methods;
+        $this->defaults['_method'] = $this->normalizeAndValidateMethods($methods);
 
         return $this;
+    }
+
+    /**
+     * Normalize method names to upper case and validate that they are valid HTTP methods.
+     *
+     * @param string|array $methods Methods.
+     * @return string|array
+     */
+    protected function normalizeAndValidateMethods($methods)
+    {
+        $methods = is_array($methods)
+            ? array_map('strtoupper', $methods)
+            : strtoupper($methods);
+
+        $diff = array_diff((array)$methods, static::VALID_METHODS);
+        if ($diff !== []) {
+            throw new InvalidArgumentException(
+                sprintf('Invalid HTTP method received. `%s` is invalid.', implode(', ', $diff))
+            );
+        }
+
+        return $methods;
     }
 
     /**
@@ -426,6 +439,9 @@ class Route
      */
     public function parse(string $url, string $method): ?array
     {
+        if ($method !== '') {
+            $method = $this->normalizeAndValidateMethods($method);
+        }
         $compiledRoute = $this->compile();
         [$url, $ext] = $this->_parseExtension($url);
 
@@ -747,7 +763,7 @@ class Route
             $url['_method'] = 'GET';
         }
         $defaults = (array)$this->defaults['_method'];
-        $methods = array_map('strtoupper', (array)$url['_method']);
+        $methods = (array)$this->normalizeAndValidateMethods($url['_method']);
         foreach ($methods as $value) {
             if (in_array($value, $defaults, true)) {
                 return true;
--- a/src/Routing/Route/Route.php
+++ b/src/Routing/Route/Route.php
@@ -118,10 +118,13 @@ class Route
      * - `_host` - Define the host name pattern if you want this route to only match
      *   specific host names. You can use `.*` and to create wildcard subdomains/hosts
      *   e.g. `*.example.com` matches all subdomains on `example.com`.
+     * - `_method` - Defines the HTTP method(s) the route applies to. It can be
+     *   a string or array of valid HTTP method name.
      *
      * @param string $template Template string with parameter placeholders
      * @param array $defaults Defaults for the route.
      * @param array $options Array of additional options for the Route
+     * @throws \InvalidArgumentException When `$options['_method']` are not in `VALID_METHODS` list.
      */
     public function __construct(string $template, array $defaults = [], array $options = [])
     {
@@ -168,7 +171,7 @@ class Route
      *
      * @param string[] $methods The HTTP methods to accept.
      * @return $this
-     * @throws \InvalidArgumentException
+     * @throws \InvalidArgumentException When methods are not in `VALID_METHODS` list.
      */
     public function setMethods(array $methods)
     {
@@ -180,8 +183,9 @@ class Route
     /**
      * Normalize method names to upper case and validate that they are valid HTTP methods.
      *
-     * @param string|array $methods Methods.
-     * @return string|array
+     * @param string|string[] $methods Methods.
+     * @return string|string[]
+     * @throws \InvalidArgumentException When methods are not in `VALID_METHODS` list.
      */
     protected function normalizeAndValidateMethods($methods)
     {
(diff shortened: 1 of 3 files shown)


src/Routing/Route/Route.php
```diff
@@ -131,6 +131,12 @@ public function __construct(string $template, array $defaults = [], array $optio
         $this->setExtensions((array)$this->options['_ext']);
         $this->setMiddleware((array)$this->options['_middleware']);
         unset($this->options['_middleware']);
+
+        if (isset($this->defaults['_method'])){
+            $this->defaults['_method'] = is_array($this->defaults['_method'])
```
othercorey, line 136: Should we just force this to be an array?

markstory: We _could_ but it is a change in behavior and could break application code as `defaults` is a public property. I briefly had changed it to an array when working on this but quickly reverted as many tests failed.

tests/TestCase/Routing/RouteBuilderTest.php
```diff
@@ -720,6 +720,31 @@ public function testResourcesMappings()
...
+     * @return void
+     */
+    public function testResourcesWithMapOnly()
+    {
+        $routes = new RouteBuilder($this->collection, '/api', ['prefix' => 'api']);
+        $routes->resources('Articles', [
+            'map' => [
+                'conditions' => ['action' => 'conditions', 'method' => 'DeLeTe'],
+            ],
+            'only' => ['conditions']
```
othercorey, line 735:
```suggestion
            'only' => ['conditions'],
```

othercorey (approved):
Looks ok

ADmad:
Don't we have constants for request method types?

othercorey:
Are you suggesting we validate the names instead of making them work?

ADmad:
Normally I would say "yes", but in this particular case I guess we don't need to be too pedantic about the casing user uses. Though I was surprised to use strings for method types in our codebase too instead of constants.

othercorey:
Well. it is bad that users can pass invalid methods. Forget casing, which is east to fix. What if they misspell "post"?

ADmad:
That's why we should provide constants and let those use strings suffer their typos :stuck_out_tongue:.

markstory:
We could add constants, and validate that the methods are within the accepted range, but that is a different change that would need to go into 4.next. I would likely put the constants on RouteBuilder so that users wouldn't need to do additional imports if they already are using typehinting.

dereuromark:
Those constants already exist for user land code to be used by the way in Cake\Http\Client\Message:

    /**
     * HTTP GET method
     *
     * @var string
     */
    public const METHOD_GET = 'GET';

    /**
     * HTTP POST method
     *
     * @var string
     */
    public const METHOD_POST = 'POST';

etc

markstory:
> Those constants already exist for user land code to be used by the way in Cake\Http\Client\Message:

Its a bit odd to need to import the HTTP client when defining routes though.

dereuromark:
Sure, just saying: if u wanted you could already use constants.

ADmad:
I added a new `normalizeAndValidateMethods()` method to get rid of the multiple `strtoupper` calls scattered across methods and also validate methods names in all arguments.

dereuromark (approved):

ADmad:
The test failures are because signature of `RedirectRoute::parse(string $url, string $method = '')` doesn't match `RedirectRoute::parse(string $url, string $method)`.

Is there any reason why `RedirectRoute` has `$method` as empty string by default? If not we can just remove the default value.

markstory:
> Is there any reason why RedirectRoute has $method as empty string by default? If not we can just remove the default value and update tests to pass a value.

Probably an accident/omission.

ADmad:
Apparent all child methods of `Route::parse()` which override `parse()` have a default value for `$method` and so does `RouteCollection::parse()`. So removing the default argument would cause lot of route failures.

So I am going to keep the signatures as they are.

src/Routing/Route/Route.php
```diff
@@ -187,9 +187,9 @@ protected function normalizeAndValidateMethods($methods)
     {
         $methods = is_array($methods)
             ? array_map('strtoupper', $methods)
-            : strtoupper($methods);
+            : [strtoupper($methods)];
 
-        $diff = array_diff((array)$methods, static::VALID_METHODS);
+        $diff = array_diff($methods, static::VALID_METHODS);
```
othercorey: Just cast it to an array in array_map.

markstory: I was trying to avoid always casting. The current code will only create an array if needed, otherwise there is no cast. From what I remember array_map is not as fast as creating a new array with the strtoupper result.

src/Routing/Route/Route.php
```diff
@@ -168,16 +172,31 @@ public function getExtensions(): array
...
+
+        return $this;
+    }
+
+    /**
+     * Normalize method names to upper case and validate that they are valid HTTP methods.
+     *
+     * @param string|array $methods Methods.
+     * @return string|array
+     */
```
othercorey, line 185:
```suggestion
     * @throws \InvalidArgumentException When methods are not one of `VALID_METHODS`
     */
```

src/Routing/Route/Route.php
```diff
@@ -420,6 +439,9 @@ public function parseRequest(ServerRequestInterface $request): ?array
      */
     public function parse(string $url, string $method): ?array
     {
+        if ($method !== '') {
+            $method = $this->normalizeAndValidateMethods($method);
```
othercorey, line 443: Should add a @throws tag for this.

src/Routing/Route/Route.php
```diff
@@ -182,6 +185,7 @@ public function setMethods(array $methods)
      *
      * @param string|array $methods Methods.
      * @return string|array
```
dereuromark, line 187: string[] here then?

PR #14910, opened 2020-08-17
Indent the HTML tags

markstory:
Make the HTML output include whitespace. This makes copying and pasting into other formats easier as the text has whitespace.

### DIFF
--- a/src/Error/Debug/HtmlFormatter.php
+++ b/src/Error/Debug/HtmlFormatter.php
@@ -105,7 +105,7 @@ class HtmlFormatter implements FormatterInterface
      */
     public function dump(NodeInterface $node): string
     {
-        $html = $this->export($node);
+        $html = $this->export($node, 0);
         $head = '';
         if (!static::$outputHeader) {
             static::$outputHeader = true;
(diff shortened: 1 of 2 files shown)


othercorey (approved):
Looks ok

ndm2:
The text output looks fine, and it is copyable via JS with white-spaces included, but it cannot be copied manually. Not sure if you were already aiming for the latter?

markstory:
@ndm2  I was going to add the copy button that we talked about separately.

ndm2:
I see 👍

src/Error/Debug/HtmlFormatter.php
```diff
@@ -119,9 +119,10 @@ public function dump(NodeInterface $node): string
      * Convert a tree of NodeInterface objects into HTML
      *
      * @param \Cake\Error\Debug\NodeInterface $var The node tree to dump.
+     * @param int $indent The current indentation level.
      * @return string
      */
-    protected function export(NodeInterface $var): string
+    protected function export(NodeInterface $var, int $indent): string
```
othercorey, line 125: Wonder if we should default this to 0.

markstory: I found the errors on undefined parameters were helpful as they made it really easy to track down where my mistakes are.

othercorey: I was just thinking for compatibility, but it's a protected function and no one is using it.

PR #14911, opened 2020-08-17
Fix getTable() collision. Make getName() usage consistent.

dereuromark:
This makes it consistent with Association::getName() and removes the IDE warning/error of string vs object since the naming collides here.
Table::getTable() vs Behavior::getTable()

This resolves this.
It also takes away the confusion, as in many cases getTable() would assume to actually get the table (just as the behavior does), when it is just the name.

The change must be documented in migration guide, of course.
The internal property is very internal, so I don't see an issue here with 4.2 as minor release.

### DIFF
--- a/src/ORM/Table.php
+++ b/src/ORM/Table.php
@@ -364,7 +364,7 @@ class Table implements RepositoryInterface, EventListenerInterface, EventDispatc
      *
      * @param string $name Table name.
      * @return $this
-     * @deprecated 4.2.0 Use setName() instead.
+     * @deprecated 4.2.0 Use setTableName() instead.
      */
     public function setTable(string $name)
     {
@@ -393,10 +393,10 @@ class Table implements RepositoryInterface, EventListenerInterface, EventDispatc
     /**
      * Returns the database table name.
      *
-     * This can include the database schema name if set using `setTable()`.
+     * This can include the database schema name if set using `setTableName()`.
      *
      * @return string
-     * @deprecated 4.2.0 Use getName() instead.
+     * @deprecated 4.2.0 Use getTableName() instead.
      */
     public function getTable(): string
     {
@@ -406,7 +406,7 @@ class Table implements RepositoryInterface, EventListenerInterface, EventDispatc
     /**
      * Returns the database table name.
      *
-     * This can include the database schema name if set using `setTable()`.
+     * This can include the database schema name if set using `setTableName()`.
      *
      * @return string
      */
--- a/src/ORM/Table.php
+++ b/src/ORM/Table.php
@@ -362,12 +362,30 @@ class Table implements RepositoryInterface, EventListenerInterface, EventDispatc
      * This can include the database schema name in the form 'schema.table'.
      * If the name must be quoted, enable automatic identifier quoting.
      *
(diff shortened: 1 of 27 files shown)


ADmad:
> This makes it consistent with Association::getName()

Not really, it actually adds an inconsistency. `Association::getName()` returns camel cased alias of the model which usually matches the classname. Similarly `Controller::getName()` returns the camel cased controller name which also by default matches class name. 

While the `Table::getName()` you added returns the database table name which is by convention underscored.

So using `Table::getTableName()/setTableName()` would be more intuitive and avoid confusing with `getName()` of other classes. 

Also I would avoid renaming the `Table::$_table` property. While we can rename protected properties in minor releases, there's no real need to rename it and cause grief to users who are directly using the property.  Personally I find  `Table::$_table` more clear about what it holds than  `Table::$_name`.

dereuromark:
Table::getTableName()/setTableName() sounds good too.

othercorey:
If we change that then the constructor config is a little inconsistent.
```
     * - table: Name of the database table to represent
     * - alias: Alias to be assigned to this table (default to table name)
```

This does not use "table name", just "table".

src/ORM/Table.php
```diff
@@ -378,8 +396,21 @@ public function setTable(string $table)
      * This can include the database schema name if set using `setTable()`.
      *
      * @return string
+     * @deprecated 4.2.0 Use getName() instead.
```
ADmad, line 399:
```suggestion
     * @deprecated 4.2.0 Use getTableName() instead.
```

src/ORM/Table.php
```diff
@@ -362,12 +362,30 @@ public function initialize(array $config): void
      * This can include the database schema name in the form 'schema.table'.
      * If the name must be quoted, enable automatic identifier quoting.
      *
-     * @param string $table Table name.
+     * @param string $name Table name.
      * @return $this
+     * @deprecated 4.2.0 Use setName() instead.
```
ADmad, line 367:
```suggestion
     * @deprecated 4.2.0 Use setTableName() instead.
```

src/ORM/Table.php
```diff
@@ -378,8 +396,21 @@ public function setTable(string $table)
...
      */
     public function getTable(): string
+    {
+        return $this->getTableName();
+    }
+
+    /**
+     * Returns the database table name.
+     *
+     * This can include the database schema name if set using `setTable()`.
```
ADmad, line 409:
```suggestion
     * This can include the database schema name if set using `setTableName()`.
```

dereuromark:
> If we change that then the constructor config is a little inconsistent.

I agree that this is not ideal.
The alternative would be to rename the Behavior method to `table()` since the behavior has no setter, and we agreed that the get/set prefix part is mainly about that.
The rest of the API has just basic collection() or object() references usually.

What do you think?

ADmad:
> The alternative would be to rename the Behavior method to table() since the behavior has no setter

Works for me.

Hold off until more people have a change to provide their feedback so that you don't have to do repeated changes :slightly_smiling_face:.

dereuromark:
Don't wry, it was an easy one. And this way we can see it more visually.

PR #14912, opened 2020-08-17
Rename getTable() to table() to remove naming collision.

dereuromark:
Alternative to https://github.com/cakephp/cakephp/pull/14911

This has a smaller footprint, but will still have the issue of clarification of table vs name (object vs string).

This makes sense if we want to further improve our API in the future (towards 5.x) using
- set/get prefix only for such setter and getters
- don't use the prefix for pure collection() and object() handles exposed by the API (where there is no such setter)

What do people think?

For example it should probably also be
- request->session()->...
- but object::setConfig() and object::getConfig()

### DIFF
--- a/src/ORM/Behavior.php
+++ b/src/ORM/Behavior.php
@@ -184,8 +184,19 @@ class Behavior implements EventListenerInterface
      * Get the table instance this behavior is bound to.
      *
      * @return \Cake\ORM\Table The bound table instance.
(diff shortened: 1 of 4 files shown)


markstory:
I prefer this set of changes over those in #14911 personally. This changes/deprecates fewer APIs and we have a few other places where we use unprefixed names as well (ServerRequest has a few too).

ADmad (approved):

PR #14913, opened 2020-08-18
Make test assert stricter. Less auto cast errors in expectations (and…

dereuromark:
I ran my new rector rule over the core code and found quite a few places 
with unexpected $expected values/results.

Applied rules:

 * Rector\PHPUnit\Rector\MethodCall\AssertEqualsToSameRector

                                                                                                                        
[OK] Rector is done! 144 files have been changed.  

The rule switches out only scalar values and thus finds the dirtyness of empty string vs false vs null etc.

### DIFF
--- a/tests/TestCase/Cache/Engine/ApcuEngineTest.php
+++ b/tests/TestCase/Cache/Engine/ApcuEngineTest.php
@@ -112,8 +112,7 @@ class ApcuEngineTest extends TestCase
         $this->_configCache(['duration' => 1]);
 
         $result = Cache::read('test', 'apcu');
-        $expecting = '';
-        $this->assertEquals($expecting, $result);
+        $this->assertNull($result);
 
         $data = 'this is a test of the emergency broadcasting system';
         $result = Cache::write('test', $data, 'apcu');
@@ -121,7 +120,7 @@ class ApcuEngineTest extends TestCase
 
         $result = Cache::read('test', 'apcu');
         $expecting = $data;
-        $this->assertEquals($expecting, $result);
+        $this->assertSame($expecting, $result);
 
         Cache::delete('test', 'apcu');
     }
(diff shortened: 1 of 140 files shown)


tests/TestCase/Database/QueryTest.php
```diff
@@ -1431,10 +1431,10 @@ public function testWhereWithBetween()
 
         $this->assertCount(2, $result);
         $first = $result->fetch('assoc');
-        $this->assertEquals(5, $first['id']);
+        $this->assertSame('5', $first['id']);
```
dereuromark, line 1434: Is the string here really what we want? 
Doesnt the ORM cast this to int as per field type?

ADmad: It does but it needs the field type map for that. Since `Database/Query` is directly used here and the type map hasn't been explicitly set, it can't do any conversion, hence string is expected.

tests/TestCase/Cache/Engine/ApcuEngineTest.php
```diff
@@ -113,15 +113,15 @@ public function testReadAndWriteCache()
 
         $result = Cache::read('test', 'apcu');
         $expecting = '';
-        $this->assertEquals($expecting, $result);
+        $this->assertSame($expecting, $result);
```
ADmad, line 116:
```
$expecting = '';
$this->assertSame($expecting, $result);
```
needs to be replaced with
```
$this->assertNull($result);
```
since cache miss returns `null` by default.

dereuromark:
Looks like some DBs return assoc as true int, others as string. We could revert for those, or cast the value before assertSame() it.

dereuromark:
I reverted the failing (different per DB type) ones for now.

garas:
Commit message lost during rebase.

dereuromark:
Fixed now :)

PR #14914, opened 2020-08-18
Fix (some) response deprecations

tersmitten:
Fixes for:

```
 notice: Deprecated (16384): Response::body() is deprecated. Mutable response methods are deprecated. Use `withBody()` and `getBody()` instead. - /var/www/example.com/app-3x/releases/20200817145138Z/vendor/cakephp/cakephp/src/Http/Response.php, line: 513#012 You can disable deprecation warnings by setting `Error.errorLevel` to `E_ALL & ~E_USER_DEPRECATED` in your config/app.php. in [/var/www/example.com/app-3x/releases/20200817145138Z/vendor/cakephp/cakephp/src/Core/functions.php, line 311]
```

### DIFF
--- a/src/Controller/Component/RequestHandlerComponent.php
+++ b/src/Controller/Component/RequestHandlerComponent.php
@@ -287,7 +287,7 @@ class RequestHandlerComponent extends Component
         }
         /** @var \Cake\Controller\Controller $controller */
         $controller = $event->getSubject();
-        $response = $response->withStringBody($controller->requestAction($url, [
+        $response->body($controller->requestAction($url, [
             'return',
             'bare' => false,
             'environment' => [
--- a/src/Controller/Component/RequestHandlerComponent.php
+++ b/src/Controller/Component/RequestHandlerComponent.php
@@ -287,7 +287,7 @@ class RequestHandlerComponent extends Component
         }
         /** @var \Cake\Controller\Controller $controller */
         $controller = $event->getSubject();
-        $response->withStringBody($controller->requestAction($url, [
+        $response = $response->withStringBody($controller->requestAction($url, [
             'return',
             'bare' => false,
             'environment' => [
--- a/src/Controller/Component/RequestHandlerComponent.php
+++ b/src/Controller/Component/RequestHandlerComponent.php
@@ -287,7 +287,7 @@ class RequestHandlerComponent extends Component
         }
         /** @var \Cake\Controller\Controller $controller */
         $controller = $event->getSubject();
-        $response = $response->withBody($controller->requestAction($url, [
+        $response = $response->withStringBody($controller->requestAction($url, [
             'return',
             'bare' => false,
             'environment' => [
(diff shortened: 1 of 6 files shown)


src/Controller/Component/RequestHandlerComponent.php
```diff
@@ -287,7 +287,7 @@ public function beforeRedirect(Event $event, $url, Response $response)
         }
         /** @var \Cake\Controller\Controller $controller */
         $controller = $event->getSubject();
-        $response->body($controller->requestAction($url, [
+        $response->withStringBody($controller->requestAction($url, [
```
dereuromark, line 290: I dont think this works.
You are not assigning the immutable response.

src/Http/Response.php
```diff
@@ -624,7 +624,7 @@ protected function _setContent()
         );
 
         if (in_array($this->_status, [304, 204])) {
-            $this->body('');
+            $this->withStringBody('');
```
dereuromark, line 627: Same here

src/Http/Client/Response.php
```diff
@@ -560,7 +560,7 @@ public function version()
      * For example to get the json data as an object:
      *
      * ```
-     * $body = $response->body('json_decode');
+     * $body = $response->getStringBody('json_decode');
```
garas, line 563:
```suggestion
     * $body = $response->getJson();
```

src/Controller/Component/RequestHandlerComponent.php
tersmitten: You're sure. Fixed

src/Http/Response.php
tersmitten: You're correct. Fixed

src/Http/Client/Response.php
tersmitten: Fixed

src/Http/Response.php
```diff
@@ -624,7 +624,7 @@ protected function _setContent()
         );
 
         if (in_array($this->_status, [304, 204])) {
-            $this->body('');
+            $this = $this->withStringBody('');
```
tersmitten, line 627: How would one fix this one?

dereuromark: With this, the body() call might actually be fine, no?

src/Http/Response.php
```diff
@@ -624,7 +624,8 @@ protected function _setContent()
         );
 
         if (in_array($this->_status, [304, 204])) {
-            $this->body('');
+            $this->_createStream();
+            $this->stream->write('');
```
tersmitten, line 628: It's not possible to do:

```
$this = $this->withStringBody('');
```

You can't assign `$this`;

dereuromark:
We should Squash merge this :)

tersmitten:
I think I stop here for now. There're some calls left, but I'm not confident on how to fix them.

dereuromark:
Looks good so far.

othercorey:
Do we want to merge this?

PR #14915, opened 2020-08-19
Formalize current deprecations

markstory:
Add formal warnings for deprecations that have been added so far in 4.x. I'll follow this up with a change to rector as well.

### DIFF
--- a/src/Database/Type/DateTimeType.php
+++ b/src/Database/Type/DateTimeType.php
@@ -163,7 +163,7 @@ class DateTimeType extends BaseType implements BatchCastingInterface
      */
     public function setTimezone($timezone)
     {
-        deprecationWarning('DateTimeType::setTimezone() is deprecated. use setDatabaseTimezone() instead.');
+        deprecationWarning('DateTimeType::setTimezone() is deprecated. Use setDatabaseTimezone() instead.');
 
         return $this->setDatabaseTimezone($timezone);
     }
--- a/src/Database/Type/DateTimeType.php
+++ b/src/Database/Type/DateTimeType.php
@@ -163,6 +163,8 @@ class DateTimeType extends BaseType implements BatchCastingInterface
      */
     public function setTimezone($timezone)
     {
+        deprecationWarning('DateTimeType::setTimezone() is deprecated. use setDatabaseTimezone() instead.');
+
         return $this->setDatabaseTimezone($timezone);
     }
 
--- a/src/Http/Cookie/Cookie.php
+++ b/src/Http/Cookie/Cookie.php
@@ -431,7 +431,7 @@ class Cookie implements CookieInterface
      */
     public function getStringValue()
     {
-        deprecationWarning('Cookie::getStringValue() is deprecated. Use getStringValue() instead.');
+        deprecationWarning('Cookie::getStringValue() is deprecated. Use getScalarValue() instead.');
 
         return $this->getScalarValue();
     }
--- a/src/Http/Cookie/Cookie.php
+++ b/src/Http/Cookie/Cookie.php
@@ -431,6 +431,8 @@ class Cookie implements CookieInterface
      */
     public function getStringValue()
     {
+        deprecationWarning('Cookie::getStringValue() is deprecated. Use getStringValue() instead.');
+
         return $this->getScalarValue();
     }
 
--- a/src/ORM/Behavior.php
+++ b/src/ORM/Behavior.php
@@ -188,6 +188,8 @@ class Behavior implements EventListenerInterface
      */
     public function getTable(): Table
     {
+        deprecationWarning('Behavior::getTable() is deprecated. Use table() instead.');
+
         return $this->table();
     }
 
--- a/src/View/Helper/NumberHelper.php
+++ b/src/View/Helper/NumberHelper.php
@@ -226,6 +226,8 @@ class NumberHelper extends Helper
      */
     public function defaultCurrency($currency): ?string
     {
+        deprecationWarning('NumberHelper::defaultCurreny() is deprecated. Use setDefaultCurreny() and getDefaultCurrent() instead.');
+
         return $this->_engine->defaultCurrency($currency);
     }
 
--- a/src/View/Helper/NumberHelper.php
+++ b/src/View/Helper/NumberHelper.php
@@ -226,7 +226,7 @@ class NumberHelper extends Helper
      */
     public function defaultCurrency($currency): ?string
     {
-        deprecationWarning('NumberHelper::defaultCurreny() is deprecated. Use setDefaultCurreny() and getDefaultCurrent() instead.');
+        deprecationWarning('NumberHelper::defaultCurrency() is deprecated. Use setDefaultCurrency() and getDefaultCurrency() instead.');
 
         return $this->_engine->defaultCurrency($currency);
     }
--- a/src/View/Helper/NumberHelper.php
+++ b/src/View/Helper/NumberHelper.php
@@ -226,7 +226,9 @@ class NumberHelper extends Helper
      */
     public function defaultCurrency($currency): ?string
     {
-        deprecationWarning('NumberHelper::defaultCurrency() is deprecated. Use setDefaultCurrency() and getDefaultCurrency() instead.');
+        deprecationWarning(
+            'NumberHelper::defaultCurrency() is deprecated. Use setDefaultCurrency() and getDefaultCurrency() instead.'
+        );
 
         return $this->_engine->defaultCurrency($currency);
     }
--- a/src/Cache/Cache.php
+++ b/src/Cache/Cache.php
@@ -206,6 +206,8 @@ class Cache
      */
     public static function engine(string $config)
     {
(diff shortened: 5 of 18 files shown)


src/ORM/Behavior.php
```diff
@@ -188,6 +188,8 @@ public function initialize(array $config): void
      */
     public function getTable(): Table
     {
+        deprecationWarning('Behavior::getTable() is deprecated. Use table() instead.');
```
othercorey, line 191: Are we going to throw a warning the same point release it was deprecated? I am not sure what the policy is or if there is one.

src/View/Helper/NumberHelper.php
```diff
@@ -226,6 +226,8 @@ public function formatDelta($value, array $options = []): string
      */
     public function defaultCurrency($currency): ?string
     {
+        deprecationWarning('NumberHelper::defaultCurreny() is deprecated. Use setDefaultCurreny() and getDefaultCurrent() instead.');
```
garas, line 229:
```suggestion
        deprecationWarning('NumberHelper::defaultCurrency() is deprecated. Use setDefaultCurrency() and getDefaultCurrency() instead.');
```

src/ORM/Behavior.php
dereuromark: Throwing them a bit too soon can be annoying for plugins and stuff.
As they always need to then pin to the highest version of minors, leaving a lot of people very annoyed by this.
The same issue I mentioned back when we started to do hard instead of soft deprecations (or rather opt out vs opt in).

ADmad: What benefit would delaying warnings have? For anyone keeping up to date with the minors will eventually see the warnings anyway.

src/Database/Type/DateTimeType.php
```diff
@@ -163,6 +163,8 @@ public function toDatabase($value, DriverInterface $driver): ?string
      */
     public function setTimezone($timezone)
     {
+        deprecationWarning('DateTimeType::setTimezone() is deprecated. use setDatabaseTimezone() instead.');
```
ADmad, line 166:
```suggestion
        deprecationWarning('DateTimeType::setTimezone() is deprecated. Use setDatabaseTimezone() instead.');
```

src/Http/Cookie/Cookie.php
```diff
@@ -431,6 +431,8 @@ public function getValue()
      */
     public function getStringValue()
     {
+        deprecationWarning('Cookie::getStringValue() is deprecated. Use getStringValue() instead.');
```
ADmad, line 434:
```suggestion
        deprecationWarning('Cookie::getStringValue() is deprecated. Use getScalarValue() instead.');
```

src/ORM/Behavior.php
dereuromark: Plugins wouldnt be forced to directly upgrade, and minimize constraint range.
So a wider constraint range makes it easier to support a wider range of users and their different plugin usage.
All they need would be to silence. But if plugins are force-constraint to last minor all the time, this puts a lot of pressure on all involved and makes it impossible to keep a lot of different plugins in the app - since not all plugins are well maintained.

ADmad: I don't see how these deprecation warnings force constraint to last minor when one can just silence the deprecations by modifying the error level.

I wouldn't mind waiting till the last minor release to add deprecation warnings but we don't follow fixed release cycles and don't exactly know which minor will be the last one. We started adding them in 3.6 thinking it might be the last but we are up to 3.9 now. Also waiting till next minor to add the warning doesn't work in practice as evident by this PR, we always forget to add the warnings later.

dereuromark: Thats because you only have a very small amount of plugins I do :)
I can tell you, this is super annoying.

The tests are failing, so you remove the deprecated part in favor of the new code.
But now you need to constraint, as the code would not work anymore with the pervious minor
So this is the force on last minor for all plugins that have anything related to this kind of deprecation errors thrown.

dereuromark: This all wouldnt be an issue if we went to what I advocating very loudly for years:
Opt-in instead of Opt-out notifications :)

rchavik: I agree. With PHP, I'd much prefer warnings in doc blocks.

dereuromark:
Above, I explained how too strict opt-out can lead to plugin constraint topic and thus fallout for projects.

Additionally, I would like to point out that we need to think about what we want to achieve here.
Do we want to help people upgrading? Or punish the ones that didn't (yet)?
The latter is sure in line with these deprecations and the result usually is that all projects just silence it completely, getting rid of any and all (even external or more low level) deprecations here.
So in the end (and that includes my apps as well), people really don't see anything anymore. And also can never activate it again, as at least a few plugins directly are responsible for some of them.

If we want to really help people while not annoy the others that are not ready for it yet (there is no business value in these things):
- We shouldn't throw them for anything that is not critical. Security-related, sure. But if it is just an API rename, that can be done any time, using rector this is a pure opt-in decision and does not need you to silence anything until then.
- An opt-in way would suffice, e.g. a configure driven flag that would enable this. Ideally, this is also scoped to the project in reporting. The projects usually have no control over plugin code, and having a lot of noise coming from those will just cloak the actual issues they would like to find and fix.

Maybe other people see it differently, this is just my experience being involved in all sides, and thus being able to speak from experience from project and plugin side as well.

src/ORM/Behavior.php
markstory: @dereuromark Building another mechanism on top of PHP errors seems silly when we already have access to an opt-in/out system for error warnings. Should we be setting the default error level in plugins and applications to exclude deprecations?

If people think the runtime warnings are annoying and not helpful, we can skip this process entirely and do only documentation warnings. I don't think documentation only warnings will help users though. You can't use documentation tags to verify that your application has no remaining deprecations. Runtime warnings lets you at least be confident that the code you have tests for, or code you manually test no longer has deprecations.

garas:
I would like granular opt-out by file path, similary how currently `psalm-baseline.xml` ignores warnings, also developer would always see the list of plugins or app files to upgrade.

Developer could add `ROOT` relative paths to `Error.ignoredDeprecationsPaths` config array, ex. `src/`, `src/SomeFile.php`, `vendor/org/plugin-name/`.

`$frame['file']` from here would be used.
https://github.com/cakephp/cakephp/blob/2c0273a37023044b96a4d6471092f48f51225f8d/src/Core/functions.php#L291-L293

markstory:
> I would like granular opt-out by file path, similary how currently psalm-baseline.xml ignores warnings, also developer would always see the list of plugins or app files to upgrade.

@garas Why an opt out by path? Would an opt-in by path work? Having a set of paths you want to receive deprecation warnings from could help with the problems that @dereuromark has been having as well.

othercorey:
One place where deprecation documentation is difficult is options. It's not something we have a standard way of deprecating.

garas:
> Why an opt out by path? Would an opt-in by path work? Having a set of paths you want to receive deprecation warnings from could help with the problems that @dereuromark has been having as well.

@markstory I think opt-out by path will work better as I want to have most code/plugins clean so I will run with warnings enabled everywhere first, then fill list one-by-one with paths if I don't control them or I can't fix now. List should be kept as short as possibly and perfectly empty.

I don't see how opt-in by path would help. If path is not in list, I will miss if previously-clean code becomes dirty code. That's not what I want.

List should document problems I know, but I can't fix now, and I should revise and clean up later by upgrading plugins or code.

ADmad:
@garas Something like this?

```diff
diff --git a/src/Core/functions.php b/src/Core/functions.php
index 330fc39f3c..b48b195ffd 100644
--- a/src/Core/functions.php
+++ b/src/Core/functions.php
@@ -292,6 +292,14 @@ if (!function_exists('deprecationWarning')) {
             $frame = $trace[$stackFrame];
             $frame += ['file' => '[internal]', 'line' => '??'];
 
+            // deprecation_optout.php must return array of paths relative to app ROOT.
+            // For e.g. ['src/Controller/FooFontroller.php', 'vendor/cakephp/bake/src/Plugin.php']
+            $optOutList = @include CONFIG . 'deprecation_optout.php';
+            $path = substr($frame['file'], strlen(ROOT) + 1);
+            if ($optOutList && in_array($path, $optOutList, true)) {
+                return;
+            }
+
             $message = sprintf(
                 '%s - %s, line: %s' . "\n" .
                 ' You can disable deprecation warnings by setting `Error.errorLevel` to' .
```

garas:
Something like that 👍, but list should be stored in `config/app.php`, `DS` normalized, and should be able exclude whole directory/plugin like `vendor/cakephp/bake/` if plugin has many bad files.

markstory:
To replay back what I'm understanding:

1. Enabling deprecations on/off would still be done via Error.level and the PHP constants.
2. If you have a noisy plugin or a 'bad' part of your code base you can silence or disable deprecations using the proposed configuration. The new configuration option could be `Error.disableDeprecations` and contain an array of glob compatible patterns. If a path matches one of the glob expression any deprecations that are triggered will not be rendered.

Is that what you folks understand too?

PR #14916, opened 2020-08-19
Add copy to debug output

markstory:
Add copy behavior to the debug output HTML. I've also replaced the unicode arrow with a tiny inline svg which renders better.

Screenshot from 2020-08-18 23-31-01

### DIFF
--- a/src/Error/Debug/dumpHeader.html
+++ b/src/Error/Debug/dumpHeader.html
@@ -2,6 +2,7 @@
 .cake-debug {
   --color-bg: #ECECE9;
   --color-highlight-bg: #fcf8e3;
+  --color-control-bg: hsla(0, 0%, 50%, 0.2);
 
   --color-orange: #c44f24;
   --color-green: #0b6125;
@@ -21,6 +22,7 @@
   line-height: 16px;
   font-size: 14px;
   margin-bottom: 10px;
+  position: relative;
 }
 .cake-debug:last-child {
   margin-bottom: 0;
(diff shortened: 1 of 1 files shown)


src/Error/Debug/dumpHeader.html
```diff
@@ -211,6 +225,27 @@
...
+    copy.setAttribute('title', 'Copy contents of debug output');
+    copy.appendChild(doc.createTextNode('Copy'));
+
+    // Add copy behavior
+    copy.addEventListener('click', function (event) {
+      event.preventDefault();
+      event.stopPropagation();
+
+      // Chop off last 4 to exclude copy button text.
+      navigator.clipboard.writeText(container.innerText.substring(0, container.innerText.length - 4));
```
ndm2, line 242:
```suggestion
      navigator.clipboard.writeText(container.textContent.substring(0, container.textContent.length - 4));
```

`innerText` won't include whitespaces and visually hidden elements.

ndm2, line 242: Personally I'm good with using the Clipboard API, it's way faster than the textarea hack, however in Firefox for example the API is only available in secure context's, ie it requires using HTTPS or `127.0.0.1`/`localhost` for serving the content. 

Would the people that use non localhost domains without SSL be of any concern?

ndm2:
I'm wondering, what do you think about including the line info too?

src/Error/Debug/dumpHeader.html
markstory: They aren't a concern I have. If the clipboard API doesn't work I don't think we need to add elaborate workarounds for this kind of feature.

markstory:
> I'm wondering, what do you think about including the line info too?

Currently the line information is not 'inside' the debug dump. Would having the line number be useful?

src/Error/Debug/dumpHeader.html
ndm2: OK. But maybe an alert in case the Clipboard API isn't available (or writing fails) would be helpful, so that people aren't just left with nothing happening.

ndm2:
> > I'm wondering, what do you think about including the line info too?
> 
> Currently the line information is not 'inside' the debug dump. Would having the line number be useful?

In combination with the filename, it might be. If I ask someone for debug output, or they provide it by themselves, seeing if they actually debugged it in the right place could be helpful.

PR #14922, opened 2020-08-21
Typecast numeric cell values to string in HtmlHelper::_renderCells()

### LINKED ISSUE #14921: HtmlHelper>-_renderCells should convert mixing values of $cell to string bevor passing it to tableCell()
This is a (multiple allowed):

* [ ] bug
* [x] enhancement

* CakePHP Version: 8f673ad
* Platform and Target: --

### What you did
`echo $this->Html->tableCells(
    [
        ['Foo', 'Bar'],
        ['Yve', 'Ice'],
        [42, 1.2345],
    ]
);`

### What happened
TypeError Exception in HtmlHelper::tableCell(), because this function aspects the argument strictly to be of type string

HtmlHelper::_renderCells() should convert arguments of type int, float, bool to string before passing it to HtmlHelper::tableCell() as $cell. 
So we can use mixed arrays.

YveIce:
This patch typecast numeric cell values to string in HtmlHelper::_renderCells() before passing it to HtmlHelper::tableCell(), while strict typehint "String" for HtmlHelper::tableCell() stays untouched.
I personally would avoid to check if method "__toString()" exists if an class/object is passed, at the time of PHP-7.x, on a later PHP-8.x only release Uniontype typehint like "stringable|string|int|float" on HtmlHelper::tableCell() will do the trick for us in a much better way. 
On type bool, everyone has his personal opinion. (And i hate flamewar). My thoughts are, that common functions like substr() return (bool) false if they fail. If we typecast them silent, users may not take notice of misbehave of there code as no exception is thrown. 

Resolves https://github.com/cakephp/cakephp/issues/14921

### DIFF
--- a/src/View/Helper/HtmlHelper.php
+++ b/src/View/Helper/HtmlHelper.php
@@ -790,6 +790,10 @@ class HtmlHelper extends Helper
                 }
             }
 
+            if(!is_string($cell) && is_numeric($cell)) {
+                $cell = (string) $cell;
+            }
+
             $cellsOut[] = $this->tableCell($cell, $cellOptions);
         }
 
--- a/src/View/Helper/HtmlHelper.php
+++ b/src/View/Helper/HtmlHelper.php
@@ -729,7 +729,7 @@ class HtmlHelper extends Helper
         bool $continueOddEven = true
     ): string {
         if (is_string($data) || empty($data[0]) || !is_array($data[0])) {
-            $data = [$data];
+            $data = [[$data]];
         }
 
         if ($oddTrOptions === true) {


YveIce:
And what is with:
```php
function tostring($value):string
{
   return (String) $value."<br>\n";
}
echo "bool(false) => ".tostring(false); // => returns empty string
echo "bool(true) => ".tostring(true); // => returns '1'
```
??? 
IMHO that's not a good behave.

dereuromark:
If you pass bool, that is what you get here for all stringish operations. So this is to be expected and OK IMO.
If you need to handle bool differently, do that before passing it on to any of those methods.

src/View/Helper/HtmlHelper.php
```diff
@@ -790,6 +790,10 @@ protected function _renderCells(array $line, bool $useCount = false): array
                 }
             }
 
+            if(!is_string($cell) && is_numeric($cell)) {
+                $cell = (string) $cell;
```
dereuromark, line 794: $cell = (string)$cell;

ADmad:
Fixed in #14940

PR #14925, opened 2020-08-23
Add support for ignoring deprecations by path

markstory:
Being able to have deprecations on globally but silence problematic
parts of your application or a noisy plugin that can't be upgraded just
yet allows developers more control when upgrading.

Instead of having to hope they don't add new deprecations they can fix
a part of their application and then remove it from the deprecation
ignore patterns.

I've used glob here instead of regex as I don't think the power of regex
is needed for this.

Refs cakephp/cakephp#14915
Refs cakephp/app#803

### DIFF
--- a/src/Core/functions.php
+++ b/src/Core/functions.php
@@ -303,9 +303,9 @@ if (!function_exists('deprecationWarning')) {
 
             $message = sprintf(
                 '%s - %s, line: %s' . "\n" .
-                ' You can disable deprecation warnings by setting `Error.errorLevel` to' .
+                ' You can disable all deprecation warnings by setting `Error.errorLevel` to' .
                 ' `E_ALL & ~E_USER_DEPRECATED`, or add `%s` to ' .
-                ' `Error.ignoredDeprecationPaths` in your `config/app.php` to mute deprecations.',
+                ' `Error.ignoredDeprecationPaths` in your `config/app.php` to mute deprecations from only this file.',
                 $message,
                 $frame['file'],
                 $frame['line'],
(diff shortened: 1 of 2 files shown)


src/Core/functions.php
```diff
@@ -292,10 +292,20 @@ function deprecationWarning(string $message, int $stackFrame = 1): void
...
+                if (fnmatch($pattern, $relative)) {
+                    return;
+                }
+            }
+
             $message = sprintf(
                 '%s - %s, line: %s' . "\n" .
                 ' You can disable deprecation warnings by setting `Error.errorLevel` to' .
-                ' `E_ALL & ~E_USER_DEPRECATED` in your config/app.php.',
+                " `E_ALL & ~E_USER_DEPRECATED`, or add `{$relative}` to " .
```
garas, line 307: Add `str_replace(DIRECTORY_SEPARATOR, '/', $relative)` through `sprintf` placeholder.

src/Core/functions.php
```diff
@@ -292,13 +292,24 @@ function deprecationWarning(string $message, int $stackFrame = 1): void
...
+                    return;
+                }
+            }
+
             $message = sprintf(
                 '%s - %s, line: %s' . "\n" .
                 ' You can disable deprecation warnings by setting `Error.errorLevel` to' .
-                ' `E_ALL & ~E_USER_DEPRECATED` in your config/app.php.',
+                ' `E_ALL & ~E_USER_DEPRECATED`, or add `%s` to ' .
+                ' `Error.ignoredDeprecationPaths` in your `config/app.php` to mute deprecations.',
```
garas, line 308: "You can disable **all** deprecation warnings by  ...., or add .... to mute deprecations **in this file only**."

garas (approved):

PR #14929, opened 2020-08-23
reject HH time form according to the doc comment in Validation::time()

tanden:
# Related pull request
https://github.com/cakephp/cakephp/pull/14927

# Description
Validation::time() returns true when two digit (HH) numbers like 12, 19 are given.
According to the comment, only `HH` should be rejected.
https://github.com/cakephp/cakephp/blob/4.next/src/Validation/Validation.php#L628-L655

```
Validates time as 24hr (HH:MM[:SS][.FFFFFF]) or am/pm ([H]H:MM[a|p]m)
```

 I committed the code change as the following.

- modify regular expression to reject when only `HH`, like 12, 23 are given and make regex to pass only `HH:MM` in standard clock expression.
- add test assertions to test above code change

# Discussion
- `HH` are basically the short version of standard clock expression, like 12(12:00), 18(18:00)
- So, I am also considering another option (modify doc comment like `HH[:MM][:SS][.FFFFFF]`)

### DIFF
--- a/src/Validation/Validation.php
+++ b/src/Validation/Validation.php
@@ -649,7 +649,7 @@ class Validation
         }
 
         $meridianClockRegex = '^((0?[1-9]|1[012])(:[0-5]\d){0,2} ?([AP]M|[ap]m))$';
-        $standardClockRegex = '^([01]\d|2[0-3])((:[0-5]\d){0,2}|(:[0-5]\d){2}\.\d{0,6})$';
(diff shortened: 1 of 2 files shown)


markstory:
With #14929 merged this has conflicts.

othercorey:
Fixed in https://github.com/cakephp/cakephp/pull/14969

Issue #14865, opened 2020-08-02
RFC - Adding a DI container


markstory:
CakePHP has long been without a dependency injection container, as we haven't needed one in CakePHP core. Instead of using a DI container we rely on service locators for the ORM, global configuration and constructor/method parameters instead. While CakePHP hasn't needed a dependency injection container directly it would be useful for application developers.

## Problems a Container Solves 

As applications increase in complexity they need ways to centralize object creation, reduce coupling to dependencies, and have a central place to build reusable objects and the relationships between them.

A container solves these problems by acting as a configurable factory for a variety of objects. By  fetching objects from the container, the container can resolve dependency trees based on the configured state. Having a central factory makes swapping implementations for testing or refactoring simpler as all the creation logic is located in one place.

## Proposed Implementation

This proposal does not include the creation of a container implementation. Instead, we should re-use an existing container based on PSR-11. Using a PSR-11 container and coupling to those interfaces allow us to change container libraries more easily in the future if necessary.

### High Level Goals

A dependency injection container in CakePHP should enable applications and plugins to define services and then have those services injected into controllers and commands as requested. Views, Helpers, Tables, Mailers and Behaviors will not have container services injected into them. These objects are generally created far away from the `Application` and thus hard to inject into. Furthermore, none of these class types are 'entrypoints' into an application and could have services passed into their methods as parameters by the top level 'entrypoint' classes (controllers and commands).

The container would be initialized during `Application` construction. Services could be registered during the new `register()` hook which would be called *after* `bootstrap()`. Calling `register()` after `bootstrap()` lets us ensure that the application and all plugins have initialized their configuration and constants before any services are defined.

### Registering Services

Services would be registered during the `register()` hook method implemented on both the
`BaseApplication` and `BasePlugin` classes. Application classes would implement a new interface,
(`ContainerApplicationInterface`). This interface would define the following methods:

- `register(ContainerInterface $container): ContainerInterface` - Implemented by applications to register their services.
- `getContainer(): ContainerInterface` - Creates the new container if one has not been created, registers all application and plugin services. Once the container is prepared references to the container will be returned.

Having a new interface ensures that container features are opt-in for standalone HTTP and console applications.

#### Plugin Services

Plugins would be able to register services in an application's container. Plugins would have a `register` hook method called *after* the application services are registered.

```
register(ContainerInterface $container): ContainerInterface
```

[...]

PR #14932, opened 2020-08-24
Add Core/ContainerInterface

markstory:
Add a DI container interface to give us an interface that applications and plugins can use when registering services. The PSR interfaces do not provide methods to register services. However, for us to have a useful container interface we need those methods as plugins and applications will want a stable interface. Should we need to change off of league/container this interface is small enough that we could adapt it to a new container implementation. This interface will also allow ambitious applications to use an alternative container implementation and not break their plugins.

Refs #14865

### DIFF
--- a/src/Core/ContainerInterface.php
+++ b/src/Core/ContainerInterface.php
@@ -40,10 +40,10 @@ interface ContainerInterface extends PsrInterface
      * @param mixed $concrete Either the classname an interface or name resolves to.
      *   Can also be a constructed object, Closure, or null. When null, the `$id` parameter will
      *   be used as the concrete class name.
-     * @param bool|null $shared Set to true to make a service shared.
+     * @param ?bool $shared Set to true to make a service shared.
      * @return \League\Container\Definition\DefinitionInterface
      */
-    public function add(string $id, $concrete = null, bool $shared = null): DefinitionInterface;
+    public function add(string $id, $concrete = null, ?bool $shared = null): DefinitionInterface;
 
     /**
      * Add a service provider to the container
--- a/src/Core/ContainerInterface.php
+++ b/src/Core/ContainerInterface.php
@@ -16,7 +16,7 @@ declare(strict_types=1);
  */
 namespace Cake\Core;
 
-use League\Container\DefinitionInterface;
+use League\Container\Definition\DefinitionInterface;
 use Psr\Container\ContainerInterface as PsrInterface;
 
 /**
@@ -28,7 +28,7 @@ use Psr\Container\ContainerInterface as PsrInterface;
  * The methods defined in this interface use the conventions provided
  * by league/container as that is the library that CakePHP uses.
  *
- * @experimental This interface is not final and can have additional 
+ * @experimental This interface is not final and can have additional
  *   methods and parameters added in future minor releases.
  */
 interface ContainerInterface extends PsrInterface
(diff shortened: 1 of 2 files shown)


src/Core/ContainerInterface.php
```diff
@@ -0,0 +1,55 @@
...
+use Psr\Container\ContainerInterface as PsrInterface;
+
+/**
+ * Interface for the Dependency Injection Container in CakePHP applications
+ *
+ * This interface extends the PSR-11 container interface and adds
+ * methods to add services and service providers to the container.
+ *
+ * The methods defined in this interface use the conventions provided
+ * by league/container as that is the library that CakePHP uses.
```
othercorey, line 29: Is this meant to be swappable with other container libraries at any point? The interface below exposes explicit league types.

markstory: >Is this meant to be swappable with other container libraries at any point?  

Ideally yes. I don't think we'd swap libraries on users in a bugfix release but it could happen in a feature release.

> The interface below exposes explicit league types.

Yes, to define constructor arguments for a service you need to chain off of `add()` or use `extend()` and then chain method calls of DefinitionInterface. I don't have a good solution for these problems as adding additional interfaces for Definitions and ServiceProviders means we need to do more wrapping and sub-classing of league/container.

An alternative approach would be to skip this interface all together and only use League\Container. However, there are two problems that I was hoping to solve with having our own container interface:

1. Insulating our users from the potential future scenario where league/container becomes unmaintained. In this scenario we could either try to help with maintaining the library or forking and adding backwards compatibility aliases like we did with Chronos. This problem might not ever happen to and we could be over-engineering in the present to solve a problem that never happens.
2. Having our own interface allows other DI containers to be used if they implement the interfaces via shims. Plugins could potentially provide 'cake compatible' implementations of php-di or symfony/container if they wanted. This problem is the harder one to solve right now, and the current code doesn't deliver on it.

We could solve problem 2 by changing how `add()` is defined and make it look like:

```php
function add($id, $concrete = null, ?bool shared = null, array $definition = [])
```

Providing all the definition options as an array would give us better encapsulation but we'd sacrifice typechecking on all the definition steps which is not great.

Another solution would be to wrap the league/container definition with a proxy that implements our interface. We can use a subclass as easily because of how [league/container works]https://github.com/thephpleague/container/blob/master/src/Container.php#L85). Problem 2 could also be solved if those compatibility shims were written against the interfaces in league/container as well.

othercorey: I don't think it's necessarily bad that the interface exposes library types. 

If the container-specific types are going to be passed into external libraries, then there isn't much point in hiding them because if we do switch, then the entire ecosystem has to switch with us.

The alternative is providing 2 sets of the same interface and selecting which one to use at runtime.

src/Core/ContainerInterface.php
```diff
@@ -0,0 +1,55 @@
...
+     */
+    public function add(string $id, $concrete = null, ?bool $shared = null): DefinitionInterface;
+
+    /**
+     * Add a service provider to the container
+     *
+     * @param \League\Container\ServiceProvider\ServiceProviderInterface $provider The service provider to add.
+     * @return $this
+     */
+    public function addServiceProvider($provider);
```
othercorey, line 54:
```suggestion
    public function addServiceProvider(ServiceProviderInterface $provider);
```
If this is the only type, we should add the typehint, right? The Psr interface doesn't because of PHP 5.6.

src/Core/ContainerInterface.php
markstory: There are potential solutions that wouldn't result in us passing external library interfaces into plugins/applications but they require more work on our part. I'm not sure that work is justifiable given that we can either contribute upstream to help with any maintenance of league/container or fork and fix should the upstream maintainers disappear entirely (which I don't think will happen).

Exposing league/container types means that any adapter for another container library will need to partly implement interfaces from CakePHP and partly from league. I think alternate implementations will be very rare and that most folks will just stick with the defaults, making this entire thought exercise moot.

src/Core/ContainerInterface.php
markstory: I didn't add this typehint as the implementation in League/container is missing this typehint. If we add this hint to our interface we'll need to implement it in the subclass and also deal with psalm possibly being grouchy.

othercorey: I see. The class that implements this interface will extend League\container. In php 7.2, you should be able to expand the parameter type, but maybe this is cleaner then.

PR #14933, opened 2020-08-24
Convenience instance factory for TableLocator instance.

dereuromark:
Psalm reports
```
ERROR: DeprecatedClass - src/TableFinderTrait.php:175:18 - Cake\ORM\TableRegistry is marked deprecated (see https://psalm.dev/098)
        $table = TableRegistry::getTableLocator()->get($className);
```

When trying to resolve the deprecations in existing 4.x app or plugin code, you can into really weird issues.
```diff
Index: src/TableFinderTrait.php
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
--- src/TableFinderTrait.php	(revision 64e49fb5f0cffb5772fac3a18f755b555232542c)
+++ src/TableFinderTrait.php	(date 1598257246628)
@@ -18,7 +18,6 @@
 use Cake\Database\Schema\CollectionInterface;
 use Cake\Datasource\ConnectionManager;
 use Cake\Filesystem\Filesystem;
-use Cake\ORM\TableRegistry;
 use FilesystemIterator;
 use ReflectionClass;
 
@@ -172,7 +171,7 @@
             return $tables;
         }
 
-        $table = TableRegistry::getTableLocator()->get($className);
+        $table = $this->getTableLocator()->get($className);
         foreach ($table->associations()->keys() as $key) {
             /** @psalm-suppress PossiblyNullReference */
             if ($table->associations()->get($key)->type() === 'belongsToMany') {
```
wouldn't work, as traits cannot easily reuse other traits.
One could use `(new ...)->` etc, but that is also not too convenient now.

I really liked the TableRegistry static factory call, but I understand that it should go away.
Would it make sense so simplify it back to a static factory method though inside the new class?

e.g:
```
TableRegistry::get('Queue.QueuedJobs')->do();
// becomes
TableLocator::instance()->get('Queue.QueuedJobs')->do();
```

### DIFF
--- a/src/ORM/Locator/TableLocator.php
+++ b/src/ORM/Locator/TableLocator.php
@@ -79,6 +79,16 @@ class TableLocator extends AbstractLocator implements LocatorInterface
     }
 
     /**
+     * @param array|null $locations
+     *
+     * @return static
+     */
+    public static function instance(array $locations = null)
+    {
+        return new static($locations);
+    }
+
+    /**
      * @inheritDoc
      */
     public function setConfig($alias, $options = null)
--- a/tests/TestCase/ORM/Locator/TableLocatorTest.php
+++ b/tests/TestCase/ORM/Locator/TableLocatorTest.php
@@ -62,6 +62,17 @@ class TableLocatorTest extends TestCase
     }
 
     /**
+     * Test TableLocator::instance()->...() factory.
+     *
+     * @return void
+     */
+    public function testInstance(): void
+    {
+        $instance = TableLocator::instance();
+        $this->assertInstanceOf(TableLocator::class, $instance);
+    }
+
+    /**
      * Test getConfig() method.
      *
      * @return void


ADmad:
`FactoryLocator::get('Table')` already gives you a `TableLocator` instance, so there's no need to add another static method to get the instance.

dereuromark:
Having such a string inside the method feels very hacky and dirty though.

dereuromark:
Also, PHPStan/Psalm seem to have trouble with this magic string resolver:
```
$table = TableRegistry::getTableLocator()->get($className);
to
$table = FactoryLocator::get('Table')->get($className);

 ------ -------------------------------------------------------------------------------------------- 
  Line   TableFinderTrait.php (in context of class Migrations\Command\BakeMigrationSnapshotCommand)  
 ------ -------------------------------------------------------------------------------------------- 
  177    Call to an undefined method Cake\Datasource\RepositoryInterface::associations().            
  179    Call to an undefined method Cake\Datasource\RepositoryInterface::associations().            
  181    Call to an undefined method Cake\Datasource\RepositoryInterface::associations().            
  185    Call to an undefined method Cake\Datasource\RepositoryInterface::getTable().                
 ------ -------------------------------------------------------------------------------------------- 

 ------ ---------------------------------------------------------------------------------- 
  Line   TableFinderTrait.php (in context of class Migrations\Command\Phinx\Dump)          
 ------ ---------------------------------------------------------------------------------- 
  177    Call to an undefined method Cake\Datasource\RepositoryInterface::associations().  
  179    Call to an undefined method Cake\Datasource\RepositoryInterface::associations().  
  181    Call to an undefined method Cake\Datasource\RepositoryInterface::associations().  
  185    Call to an undefined method Cake\Datasource\RepositoryInterface::getTable().  
```

As the result is now still LocatorInterface and not Table as before.

ADmad:
We have a `TableRegistry::getTableLocator()` earlier which is now deprecated. So now adding another, practically same, static method `TableLocator::instance()` makes no sense.

There's nothing hackish about using `FactoryLocator::get('Table')`. It's a container for locators of different datasource types hence the return type can only be limited to a general interface. It's no different than using `$container->get('foo')` when using a DIC. The returned value could be of any type. Your use case is uncommon and most people won't need to use `FactoryLocator::get()` directly anyway.

Regarding PHPStan/Psalm just add an inline docblock to indicate the return type to make static analyzers/IDEs aware of the type.

dereuromark:
> We have a TableRegistry::getTableLocator() method which is now deprecated. So now adding another, practically same, static method TableLocator::instance() makes no sense.

Do you plan to undeprecate this one then?

ADmad:
I am happy with the current API for accessing table locator instances and not in favour of adding a new static method for it. Let's see if any other team member is.

dereuromark:
I think I understand now where the issue comes from
Someone just deprecated the whole class (which contained 2 not deprecated ones)

https://github.com/cakephp/cakephp/pull/14934 resolved it.
This will also make PHPStan and Psalm happy for the deprecation usage part, e.g. in the case mentioned above.

PR #14934, opened 2020-08-24
Undeprecate class

dereuromark:
Resolves https://github.com/cakephp/cakephp/pull/14933

now the remaining API is as expected (and documented in the docblock above):

```php
// OK
$table = TableRegistry::getTableLocator()->get('Users', $config);
// same for the setter

// Deprecated
$table = TableRegistry::get('Users', $config);
```

### DIFF
--- a/src/ORM/TableRegistry.php
+++ b/src/ORM/TableRegistry.php
@@ -55,9 +55,6 @@ use Cake\ORM\Locator\LocatorInterface;
  * // Prior to 3.6.0
  * $table = TableRegistry::get('Users', $config);
  * ```
- *
- * @deprecated 4.1.0 Use {@see \Cake\ORM\Locator\LocatorAwareTrait::getTableLocator()}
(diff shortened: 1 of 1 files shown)


dereuromark:
We could add a sentence as such:

> Consider using the LocatorAwareTrait directly if not in trait context.

ADmad:
Undeprecating the class means all those upgrading from 3.x and 4.0 will continue using the static interface instead of using the `getTableLocator()` instance method even for controller/command classes and we are back to square one.

dereuromark:
That's why we have the inline deprecations in place where needed, and we can add that sentence and clarification of trait usage where needed.
But deprecation of an actually not deprecated method is not helpful either.

dereuromark:
I prefer this simple one over the addition of a non-deprecated new syntactic sugar method - especially if the solution is so simple.

dereuromark:
Is this acceptable as is?

ADmad:
The circle of futility is complete.

PR #14935, opened 2020-08-25
Fix: Type Missmatch in HtmlHelper::tableCells()

YveIce:
TableCells() allow $data to be of type string. If so, it will be warped into an array an later unpacked in a foreach loop, and passed as string to _renderCells(). But _renderCells() aspects the argument $line to be an array.

Double warp it as array should fix this.

### DIFF
--- a/src/View/Helper/HtmlHelper.php
+++ b/src/View/Helper/HtmlHelper.php
@@ -729,7 +729,7 @@ class HtmlHelper extends Helper
         bool $continueOddEven = true
     ): string {
         if (is_string($data) || empty($data[0]) || !is_array($data[0])) {
-            $data = [$data];
+            $data = [[$data]];
         }
 
         if ($oddTrOptions === true) {


dereuromark:
Do you have a test case for this?
What kind of code triggers this?

YveIce:
```php
echo $this->Html->tableCells("Foo");
```
Throws Exception:
`Argument 1 passed to Cake\View\Helper\HtmlHelper::_renderCells() must be of the type array, string given, called in /var/www/vhosts/devil.c-plan.work/PUBLIC/vendor/cakephp/cakephp/src/View/Helper/HtmlHelper.php on line 754`

Testcase:
```php
public function test(): bool
 return ($this->Html->tableCells("Foo") == $this->Html->tableCells(["Foo"]));
```

dereuromark:
Solved in https://github.com/cakephp/cakephp/pull/14940

PR #14939, opened 2020-08-26
Psalm

ADmad:

### DIFF
--- a/psalm.xml
+++ b/psalm.xml
@@ -42,5 +42,6 @@
         </UndefinedConstant>
         <PropertyNotSetInConstructor errorLevel="suppress"/>
         <UnresolvableInclude errorLevel="suppress"/>
+        <UnsafeInstantiation errorLevel="suppress"/>
     </issueHandlers>
 </psalm>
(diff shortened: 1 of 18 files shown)


psalm.xml
```diff
@@ -42,5 +42,6 @@
         </UndefinedConstant>
         <PropertyNotSetInConstructor errorLevel="suppress"/>
         <UnresolvableInclude errorLevel="suppress"/>
+        <UnsafeInstantiation errorLevel="suppress"/>
```
ADmad, line 45: We have multiple errors reported for this rule https://psalm.dev/docs/running_psalm/issues/UnsafeInstantiation/.

The recommended "fix" for this is to mark the constructor `final`, but I don't think that's something we can do until next major release.

dereuromark: I dont even think we should do this in the next major release - at least not without a different way of instantiation added (static factory method).

ADmad: Yes using factory methods is the other alternative. We can evaluate on per class basis which option to go for.

PR #14940, opened 2020-08-27
Bugfix table cells

ADmad:
#14921, #14935


src/View/Helper/HtmlHelper.php
```diff
@@ -728,7 +728,9 @@ public function tableCells(
         bool $useCount = false,
         bool $continueOddEven = true
     ): string {
-        if (is_string($data) || empty($data[0]) || !is_array($data[0])) {
+        if (!is_array($data)) {
```
ADmad, line 731: We should deprecate support for non array values in `4.2`.

Issue #14815, opened 2020-07-21
Skip next validation rule if data is not valid


mcsknp:
This is a (multiple allowed):

* [ ] bug
* [x] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 4.1
* Platform and Target: Windows, NGINX, PHP 7.3, SQL server

### What you did
Validate invalid request data.

### What happened
Some validation method is written with type for arguments or using function that defined with datatype for arguments but request data can be an array or string, so system error will be occurred with invalid request data.

### What you expected to happen
An option to skip next validation rule if data is invalid.

markstory:
What would the interface you'd like to use to do this look like?

mcsknp:
Add `$skipOnError` property to `ValidationRule`.
Add `$sequence` property to `Validator`.

```php
$validator->beginSequence();
// use $validator->...() to add any rule for field
$validator->endSequence();
```

When `$sequence` is on rule will be added to `ValidationSet` with `$skipOnError` flag.
And in `Validator::_processRules()`, if an error found, break `for` if `$skipOnError` flag is on.

dereuromark:
We should close this in favor of https://github.com/cakephp/cakephp/issues/14816 and the linked PR there.

markstory:
I don't fully understand what a 'sequence' is and have a few questions:

* Would a 'sequence' be a chain of validation methods that stop when the first rule fails? 
* Would you be able to define sequences within other sequences?
* What happens to sequences that are started but not ended?
* What happens with empty sequences?

mcsknp:
Sorry for bad explanation.
It isn't sequence, it's a property of validation rule.
And the `sequence` I wrote is a property to enable `skipOnError` flag of validation rule when add a rule to set.
I think adding a property to `Validator` class and use its when add a validation rule is better than add an argument to all method of it.

dereuromark:
Isnt this somewhat what last=> true does?

mcsknp:
How to set `last` to true?

dereuromark:
The old way: https://github.com/dereuromark/cakephp-sandbox/blob/master/src/Model/Table/UsersTable.php#L41
The new way is a method on it afaik.
EIther way: The default should here be true (as I think was once discussed years ago), as the next rule only makes sense usually if that one worked.
Email is a classic example: First check if min length, then regexp pattern, then domain existence (API) check and alike.
Most rules are usually always in such a chain - and thus last should be considered true here always as default case.

[...]

PR #14943, opened 2020-08-29
Validation last=true regressions

dereuromark:
While upgrading an app from 3.x to 4.x I ran into a few hard blockers.
One was the hidden behavior change of validation rules being applied differently now.
See https://github.com/cakephp/cakephp/issues/14815#issuecomment-678568430

An additional problem seems to be that a rule setup as baked like this does not allow setting last in any way:

		$validator
			->requirePresence('email', 'create')
			->notBlank('email')
			->email('email');

### Part 1 of 2

This resolves this at least in a way that you can now extend the validator on project level and globally set it to true, or invoke the flag method enableLast() for it in the specific tables and retain the old 3.x behavior again.

I targeted master here since this is quite a blocker for apps trying to upgrade, and this seems to be fully BC since you need do enable this new flag.

PS: We could also make this a static call, so it could be done from bootstrapping process, rather than inline.
Or allow `(bool)Configure::read('Validation.last')` to read it in the constructor, this way it would be easier to configure without having to overwrite/extend so much.

--- 

 Part 2 of 2 will then be an API cleanup in 4.next to allow the convenience wrappers, e.g. notBlank() email() etc to accept an additional extra argument to also set the last flag per rule (which currently seems also impossible now.

### DIFF
--- a/src/Validation/Validator.php
+++ b/src/Validation/Validator.php
@@ -201,7 +201,6 @@ class Validator implements ArrayAccess, IteratorAggregate, Countable
      * The first failing one will abort then per rule.
      *
      * @param bool $stopOnFailure If to apply last flag.
-     *
      * @return $this
      */
     public function stopOnFailure(bool $stopOnFailure = true)
--- a/src/Validation/Validator.php
+++ b/src/Validation/Validator.php
@@ -197,8 +197,10 @@ class Validator implements ArrayAccess, IteratorAggregate, Countable
     }
 
     /**
-     * Whether to apply last flag to generated rule(s).
-     * The first failing one will abort then per rule.
+     * Whether to stop validation rule evaluation on the first failed rule.
+     *
+     * When enabled the first failing rule per field will cause validation to stop.
+     * When disabled all rules will be run even if there are failures.
      *
      * @param bool $stopOnFailure If to apply last flag.
      * @return $this
--- a/src/Validation/Validator.php
+++ b/src/Validation/Validator.php
@@ -181,12 +181,34 @@ class Validator implements ArrayAccess, IteratorAggregate, Countable
     protected $_allowEmptyFlags = [];
 
     /**
+     * Whether to apply last flag to generated rule(s).
+     *
+     * @var bool
+     */
+    protected $_stopOnFailure = false;
+
+    /**
      * Constructor
      */
     public function __construct()
     {
         $this->_useI18n = function_exists('__d');
         $this->_providers = self::$_defaultProviders;
+    }
+
+    /**
+     * Whether to apply last flag to generated rule(s).
+     * The first failing one will abort then per rule.
+     *
+     * @param bool $stopOnFailure If to apply last flag.
+     *
+     * @return $this
+     */
+    public function stopOnFailure(bool $stopOnFailure = true)
+    {
+        $this->_stopOnFailure = $stopOnFailure;
+
+        return $this;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


src/Validation/Validator.php
```diff
@@ -189,6 +196,18 @@ public function __construct()
...
+    /**
+     * Whether to apply last flag to rule(s).
+     * The first failing one will abort then per rule.
+     *
+     * @param bool $last If to apply last flag.
+     * @return void
+     */
+    public function enableLast(bool $last = true): void
+    {
+        $this->_last = $last;
```
markstory, line 208: Should this method be chainable and return $this?

src/Validation/Validator.php
```diff
@@ -189,6 +196,18 @@ public function __construct()
...
     }
 
+    /**
+     * Whether to apply last flag to rule(s).
+     * The first failing one will abort then per rule.
+     *
+     * @param bool $last If to apply last flag.
+     * @return void
+     */
+    public function enableLast(bool $last = true): void
```
markstory, line 206: Would `stopOnFailure` or `stopOnError` be a better name? I know we used last in 3.x but I've never found that name to be overly clear.

dereuromark:
OK, this is the result as per feedback.
I am still not sure if we shouldn't allow a more global way here, e.g. using Configure value or a static method to be called from bootstrapping.
Right now, one would have to overwrite/set it in each model, and it is probably difficult to reach the plugin ones or other non app code directly.

markstory:
If there were global configuration, then we'd also need a way to disable that behavior and run all validation rules as we do today. Having to call methods is a bit of overhead but makes the validator's behavior explicit and easy to understand when reading it. A global configuration option would hide some of the behavior away and require you to cross reference more.

src/Validation/Validator.php
```diff
@@ -189,6 +196,20 @@ public function __construct()
         $this->_providers = self::$_defaultProviders;
     }
 
+    /**
+     * Whether to apply last flag to generated rule(s).
+     * The first failing one will abort then per rule.
```
markstory, line 201:
```suggestion
     * Whether to stop validation rule evaluation on the first failed rule.
     *
     * When enabled the first failing rule per field will cause validation to stop.
     * When disabled all rules will be run even if there are failures.
```

dereuromark:
Is this acceptable as is?

markstory:
I don't like adding new methods in bugfix releases as it makes documentation more complicated. I think we should make an exception here as upgrading applications from 3.x to 4.x is more complicated without an API like this that makes defining 'last' rules simple.

dereuromark:
Started some rough docs on https://github.com/cakephp/docs/pull/6777

dereuromark:
Are there any blockers on this one? Or viable alternatives to explore?

src/Validation/Validator.php
```diff
@@ -189,6 +196,22 @@ public function __construct()
...
+    /**
+     * Whether to stop validation rule evaluation on the first failed rule.
+     *
+     * When enabled the first failing rule per field will cause validation to stop.
+     * When disabled all rules will be run even if there are failures.
+     *
+     * @param bool $stopOnFailure If to apply last flag.
+     * @return $this
+     */
+    public function stopOnFailure(bool $stopOnFailure = true)
```
othercorey, line 208: My only comment is this interface. We have stopped making function that are positive `stop...` but have a parameter that can make them positive or negative. Can we split this into enable/disable?

dereuromark: I find a separated api counter productive. It makes also setting it from a simple input more complex.

othercorey: Then we should call it `set...()` to follow other changes, yes?

dereuromark: I changed the method name.

dereuromark:
Is there anything else needed?

PR #14944, opened 2020-08-29
Add a way to restore 3.x behavior for last=true

dereuromark:
Follows https://github.com/cakephp/cakephp/pull/14943

It basically, independent from the global setting for last, allows overwriting it on a per rule base from the convenience wrappers.
In 5.x we should clean up the API and remove some of the many args that can just be added into $extra array.

These are the baked rules, and it would be quite annoying to have to completely rewrite them only to have the flag set (and possibly other options).

WIP
so far only changed
- notBlank()
- email()

wrappers as showcase, the rest can be done once this is OK with everyone.

### DIFF
--- a/src/Validation/Validator.php
+++ b/src/Validation/Validator.php
@@ -1255,16 +1255,17 @@ class Validator implements ArrayAccess, IteratorAggregate, Countable
      * @param string|null $message The error message when the rule fails.
      * @param string|callable|null $when Either 'create' or 'update' or a callable that returns
      *   true when the validation rule should be applied.
+     * @param array $extra Extra arguments, e.g. last.
      * @see \Cake\Validation\Validation::notBlank()
      * @return $this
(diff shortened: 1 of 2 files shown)


markstory:
> In 5.x we should clean up the API and remove some of the many args that can just be added into $extra array.

Couldn't we leverage named parameters? Doing that would allow us to leverage typechecking and IDE completion better than an 'extra' array would.

dereuromark:
I dont think we can use php8 Features anytime the next 1-2 years.

markstory:
I agree we won't be able to use PHP8 features, but our user base might be able to upgrade before we can. Adding additional positional parameters instead of an extra array will make Validator better when php8 comes around. We've spent a good amount of time improving typechecking across the framework and adding more array parameters which make typechecking harder seems like a mistake to me.

PR #14945, opened 2020-08-29
Add a dependency injection container

markstory:
This branch will collect the various incremental changes as support for a dependency injection container are added. The plan for this work can be found in #14865


src/Core/ContainerInterface.php
```diff
@@ -0,0 +1,55 @@
...
+     * Add an item to the container.
+     *
+     * @param string $id The class name or name of the service being registered.
+     * @param mixed $concrete Either the classname an interface or name resolves to.
+     *   Can also be a constructed object, Closure, or null. When null, the `$id` parameter will
+     *   be used as the concrete class name.
+     * @param ?bool $shared Set to true to make a service shared.
+     * @return \League\Container\Definition\DefinitionInterface
+     */
+    public function add(string $id, $concrete = null, ?bool $shared = null): DefinitionInterface;
```
robertpustulka, line 46: Maybe this should be split into three methods each for individual case: `addClassName`, `addObject` and `addClosure`?

src/Core/ContainerInterface.php
```diff
@@ -0,0 +1,55 @@
...
+
+    /**
+     * Add a service provider to the container
+     *
+     * @param \League\Container\ServiceProvider\ServiceProviderInterface $provider The service provider to add.
+     * @return $this
+     */
+    public function addServiceProvider($provider);
```
robertpustulka, line 54: Should the parameter be type hinted in method definition?

robertpustulka (commented):
I'm not sure I like sticking to the League implementation of DIC, but this is fine by me as long as the interface is there just for the out-of-the-box plugin compatibility and that I can swap the container in my applications.

src/Core/ContainerInterface.php
ADmad: We should probably make our own interface which extends `league/container`'s interface and use that for typehints to be safe. I am vary of directly using external interfaces/classes in case the lib becomes unmaintained in future, like `Aura.Intl` for example.

src/Core/ContainerInterface.php
markstory: That would diverge from league/container. We might be swapping that library out though as we've not had much luck so far in getting php8 support added.

src/Core/ContainerInterface.php
markstory: I didn't add a typehint as I was trying to keep the interface aligned with what is in league/container. I am happy to revisit the container interface once everything is hooked up and working.

src/Core/PluginInterface.php
```diff
@@ -20,6 +20,9 @@
 
 /**
  * Plugin Interface
+ *
+ * @method \Cake\Core\ContainerInterface register(\Cake\Core\ContainerInterface $container) Register plugin services to
```
robertpustulka, line 24: Maybe this method should be declared as a part of separate `ContainerPluginInterface`?

markstory: Would we then merge the two interfaces in 5.0? I thought an interface with a single method in it was a bit of a waste.

robertpustulka: I'd like to avoid interface method confusion we had in 3.x with entity methods being added to interface in docblocks.

markstory:
I think this has all of what I outlined in the linked issue. One potential addition I could see is an event for the container being built, but I don't have a good use case for it just yet.

markstory:
Any other feedback on this?

othercorey:
Other than the `register()` method as an annotation on `PluginInterface` looks good to merge for now. I don't know that we really have a better solution than annotations until cakephp 5 is released. We can consider containers an experimental feature until then probably.

markstory:
> Other than the register() method as an annotation on PluginInterface looks good to merge for now. I don't know that we really have a better solution than annotations until cakephp 5 is released. 

Would you prefer it be a concrete method on PluginInterface?

othercorey:
> Would you prefer it be a concrete method on PluginInterface?

It should be, but since we're adding this mid-release to an interface it's what it has to be. We definitely don't want to hide the container api for cakephp 5.

I'm not personally a fan of just calling it `register`, but I leave that up to you.

robertpustulka:
> Would you prefer it be a concrete method on PluginInterface?

It can be a new interface (something like `ContainerPluginInterface` similar to `ContainerApplicationInterface`) that the base Plugin can implement.

markstory:
> I'm not personally a fan of just calling it register, but I leave that up to you.

I'm open to changing the names of the methods if we have alternatives in mind.

ADmad:
`register()` is inline with naming used in `league/container`. There's `ServiceProviderInterface::register()` https://container.thephpleague.com/3.x/service-providers/

othercorey:
I was just thinking in the context of a Plugin - register is a little generic. It doesn't really imply container-specific stuff.

PR #14953, opened 2020-09-02
check phpmetric tool

dereuromark:
You can check out this branch and run the tool

We can discuss if we want to look into specific points this tool outlines and create action points for them (as things to be done in the future)

Screenshot from 2020-09-02 16-00-16

The biggest issue it seems to have is with FormHelper, and Hash class.
Otherwise a lot of things like stability and minor other topics.

If nothing should be done, we can just close this.


markstory:
Does using this tool require checking in its output? I'm not sure we should be bundling 169K lines of HTML in our repo.

dereuromark:
This pr is not to merge. Only to check :)

PR #14959, opened 2020-09-06
4.next - Add container to ControllerFactory

markstory:
Implement the ContainerApplicationInterface on BaseApplication and integrate it into ControllerFactory so that dependencies can be injected into controller constructors and action parameters. Parameters for constructors will need to be explicitly registered into the container. Whilst action parameters are inferred by typehint. Not needing to redefine constructor parameters is something that can be removed if we use the auto-wiring container delegate. In a subsequent pull request I'll update CommandFactory to work in a similar way.

Refs #14865


src/Controller/ControllerFactory.php
```diff
@@ -73,9 +96,23 @@ public function invoke($controller): ResponseInterface
...
         }
-
         $action = $controller->getAction();
-        $args = array_values($controller->getRequest()->getParam('pass'));
+
+        $args = [];
+        $reflection = new ReflectionFunction($action);
+        $passed = array_values((array)$controller->getRequest()->getParam('pass'));
+        foreach ($reflection->getParameters() as $i => $parameter) {
+            if (isset($passed[$i])) {
```
ADmad, line 105: Checking for passed args first means the action methods would required declaring params for passed args before and additional params at the end. Wouldn't most people prefer the having params for passed args at the end instead? One reason being passed args could be optional. If for e.g. your method needs a service class it's unlikely to be optional.

src/Controller/ControllerFactory.php
```diff
@@ -73,9 +96,23 @@ public function invoke($controller): ResponseInterface
...
+                $args[$i] = $passed[$i];
+                continue;
+            }
+            $class = $parameter->getClass();
+            $value = $parameter->getDefaultValue();
+            if ($class) {
+                $value = $this->container->get($class->getName());
```
ADmad, line 112: I think you should first check if the required name is available in container, using `has()`, before trying to `get()` it, else you will get an exception. (A test should be added for this case).

src/Controller/ControllerFactory.php
markstory: Yes that is a possibility too. I could make this argument handling code more robust and check if each argument exists in the container and if not, then draw from a passed argument until we run out of passed arguments.

src/Controller/ControllerFactory.php
markstory: Ok. I'll see what a more robust approach to handling parameters would be. I think you have a good point that injected services should be first in the parameter list as they are unlikely to be optional whilst passed routing parameters are more likely to be optional.

src/Controller/ControllerFactory.php
```diff
@@ -73,9 +98,49 @@ public function invoke($controller): ResponseInterface
...
+            }
+            $typeName = $type->getName();
+            if (substr($typeName, 0, 1) === '?') {
+                $typeName = substr($typeName, 1);
+            }
+
+            // Primitive types are passed args as they can't be looked up in the container.
+            if (in_array($typeName, ['float', 'int', 'string', 'bool'], true)) {
+                if (count($passed) || !$type->allowsNull()) {
+                    $args[$position] = array_shift($passed);
```
ADmad, line 124: Passed args from URL would be always string, so wouldn't we need to typecast the arg if the param is typehinted with any primitive type besides `string`?

markstory: Yeah we would. Historically we've not done that typecasting. We could add that now, but we could run into complexity around handling invalid data. I'll change this to only look for strings instead. Adding other primitive types to passed arguments can be a separate discussion.

ADmad: Yeah the code checks for all primitive types hence my comment. If we aren't going to typecast then we need to assume that the method param can only have string type or none specified.

ADmad:
LGTM. @robertpustulka Do you have any feedback on this?

PR #14962, opened 2020-09-07
Don't allow PluginLoadCommand to add invalid plugins

### LINKED ISSUE #14870: PluginLoadCommand load should check if Plugin exists and CamelCase is correct
This is a (multiple allowed):

* [ ] bug
* [x] enhancement

* CakePHP Version: 8f673ad
* Platform and Target: 

### What you did
`bin/cake plugin load NOTEXISTINGplugin
`
### What happened

> $this->addPlugin('NOTEXISTINGplugin');

gets added to Application.php, even if the plugin does not exists, breaking everything as the plugin could not be loaded. 

Throw an exception/print an error if the user try's to load a plugin that does not exists or is spelled wrong.

markstory:
Catch mistakes earlier on as users can easily make typos when adding plugins to their applications.

Fixes #14870

### DIFF
--- a/src/Command/PluginLoadCommand.php
+++ b/src/Command/PluginLoadCommand.php
@@ -74,7 +74,7 @@ class PluginLoadCommand extends Command
         try {
             Plugin::getCollection()->findPath($plugin);
         } catch (MissingPluginException $e) {
-            $this->io->err('Could not find plugin <info>NopeNotThere</info>');
+            $this->io->err($e->getMessage());
             $this->io->err('Ensure you have the correct spelling and casing.');
 
             return static::CODE_ERROR;
--- a/src/Command/PluginLoadCommand.php
+++ b/src/Command/PluginLoadCommand.php
@@ -19,6 +19,8 @@ namespace Cake\Command;
 use Cake\Console\Arguments;
 use Cake\Console\ConsoleIo;
 use Cake\Console\ConsoleOptionParser;
+use Cake\Core\Plugin;
+use Cake\Core\Exception\MissingPluginException;
 
 /**
  * Command for loading plugins.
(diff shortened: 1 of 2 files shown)


src/Command/PluginLoadCommand.php
```diff
@@ -69,6 +71,15 @@ public function execute(Arguments $args, ConsoleIo $io): ?int
             return static::CODE_ERROR;
         }
 
+        try {
+            Plugin::getCollection()->findPath($plugin);
+        } catch (MissingPluginException $e) {
+            $this->io->err('Could not find plugin <info>NopeNotThere</info>');
```
ADmad, line 77: Need to use `$plugin`  instead of `NopeNotThere` in error message :slightly_smiling_face:.

YveIce:
```suggestion
            $this->io->err($e->getMessage());
```

tests/TestCase/Command/PluginLoadCommandTest.php
```diff
@@ -77,16 +77,31 @@ public function testHelp()
...
+    /**
+     * Test loading an unknown plugin
+     *
+     * @return void
+     */
+    public function testLoadUnknownPlugin()
+    {
+        $this->exec('plugin load NopeNotThere');
+        $this->assertExitCode(Command::CODE_ERROR);
+        $this->assertErrorContains('Could not find plugin <info>NopeNotThere</info>');
```
YveIce, line 102:
```suggestion
        $this->assertErrorContains('Plugin NopeNotThere could not be found.');
```

YveIce (changes requested):
I suggest to life with loosing the <info> tag, but using the default messageTemplate declared in /src/Command/MissingPluginException.php instead.

src/Command/PluginLoadCommand.php
markstory: I was pretty tired when I tried to do this and it shows. Sorry about that.

dereuromark:
Based on the list of plugins inside the vendors plugin file, we could even output a list of possible plugins to chose from.

PR #14963, opened 2020-09-08
Allow TimeType to marshal H:i format

### LINKED ISSUE #14957: conversion of time-input to time-instance failing
This is a:

* [x] bug
* [ ] enhancement

* CakePHP Version: 4.1.3
* Platform and Target: NGINX, mySQL, php7.4.7 in a docker-container

### What you did
Using the form-helper, I created a time-field ([https://book.cakephp.org/4/en/views/helpers/form.html#creating-time-controls](https://book.cakephp.org/4/en/views/helpers/form.html#creating-time-controls)).
Validation setting: 
```
$validator
    ->time('departuretime')
    ->requirePresence('departuretime', 'create')
    ->notEmptyTime('departuretime');
```

### What happened
When entering a time-value without seconds (i.e. 18:45), no time-value is patched into the entity. No validation-error is beeing shown. When I leave the time-field empty, the validation-error shows up. In the request-data-Array the time-value is shown.

```
[
    'name' => 'Test',
'departuretime' => '18:00',
]
```

In the entity the departuretime is missing.

I expected the time-value beeing converted into a time-instance and saved to the database. When I manually alter the time-value in a beforeMarshal()-function in my model (i.e. 18:45:00) or create a time instance everything works out fine.
```
public function beforeMarshal(EventInterface $event, ArrayObject $data, ArrayObject $options)
    {
        if ($data['departuretime']) {
             // creating time-instance
              $data['departuretime'] = new FrozenTime($data['departuretime']);
             // setting time-format manually
              $data['departuretime'] = '18:45:00';
        }
    }
```

markstory:
This aligns the behavior of TimeType with the values accepted by Validation::time(). Doing so will prevent tedious to diagnose problems where validation passes but no value is set in the entity.

The x:y format could also be marshaled as I:S but we already accept Y-m-d H:i in the DateTimeType and having consistency is important.

Fixes #14957

### DIFF
--- a/src/Database/Type/TimeType.php
+++ b/src/Database/Type/TimeType.php
@@ -35,6 +35,7 @@ class TimeType extends DateTimeType
      */
     protected $_marshalFormats = [
         'H:i:s',
+        'H:i',
     ];
 
     /**
--- a/tests/TestCase/Database/Type/TimeTypeTest.php
+++ b/tests/TestCase/Database/Type/TimeTypeTest.php
@@ -146,13 +146,13 @@ class TimeTypeTest extends TestCase
             ['', null],
             ['derpy', null],
             ['16-nope!', null],
-            ['14:15', null],
             ['2014-02-14 13:14:15', null],
 
             // valid string types
             ['1392387900', $date],
             [1392387900, $date],
             ['13:10:10', new Time('13:10:10')],
+            ['14:15', new Time('14:15:00')],
 
             // valid array types
             [


dereuromark (approved):

ADmad (approved):
LGTM. It could probably go into master itself.

othercorey:
Do we know why the notEmpty() validation isn't triggered when this marshal fails?

markstory:
@ADmad I can move this to master if you think it will be ok. I was concerned that previously invalid data becoming valid could be unexpected in a bugfix release.

ADmad:
@markstory I guess it's better to err on the side of caution, so fine by me to merge into `4.next`.

PR #14965, opened 2020-09-08
Don't run test suite on nightly/PHP 8.

ADmad:
PHP 8 support requires updating dependencies and hence a new minor release.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,11 +28,6 @@ matrix:
   include:
     - php: 7.2
       env: PREFER_LOWEST=1
-    - php: nightly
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -48,23 +43,17 @@ before_install:
         pecl install pcov
       fi
 
-  - |
-      if [[ $TRAVIS_PHP_VERSION != 'nightly' ]]; then
```
othercorey, line 52: Can we just remove the nightly include and leave this so future merges are clean?

ADmad: Done

PR #14967, opened 2020-09-08
Fixtures not truncated when both $import and $fields aren't set

### LINKED ISSUE #14958: Unit Testing: Tables not truncated when fixture is setup
This is a (multiple allowed):

* [x] bug
* [ ] enhancement

* CakePHP Version: 4.1.3
* Platform and Target: Apache, MariaDB

### What you did
Unit testing using fixtures that don't have $fields or $import set, so it uses existing database table structure. And data is inserted through the fixtures

### What happened
The first run of the tests works fine but if you run it again, you get 'Unable to insert fixture "" in "" test case:' errors because of duplicate entries. The tables are not being truncated when the fixture is loaded.

Multiple test runs in a row without having to manually truncate the database tables. It was working correctly in version 4.0

jonmc1:
Issue #14958

When both `$fixture->import` and `$fixture->fields` are not set, the fixture uses the structure of an existing table. When setting up the table, the fixture manager was using `$fixture->drop()` and `$fixture->create()` when it should be using `$fixture->truncate()` in this case. The drop and create methods just return true right away with this setup.

### DIFF
--- a/src/TestSuite/Fixture/TestFixture.php
+++ b/src/TestSuite/Fixture/TestFixture.php
@@ -310,7 +310,7 @@ class TestFixture implements ConstraintsInterface, FixtureInterface, TableSchema
         }
 
         if (empty($this->import) && empty($this->fields)) {
-            return true;
+            return $this->truncate($db);
         }
 
         try {
--- a/tests/TestCase/TestSuite/TestFixtureTest.php
+++ b/tests/TestCase/TestSuite/TestFixtureTest.php
@@ -178,7 +178,6 @@ class TestFixtureTest extends TestCase
         $this->assertSame(['id', 'letter'], $fixture->getTableSchema()->columns());
 
         $db = $this->getMockBuilder('Cake\Database\Connection')
-            ->onlyMethods(['prepare', 'execute'])
             ->disableOriginalConstructor()
             ->getMock();
         $db->expects($this->never())


src/TestSuite/Fixture/FixtureManager.php
```diff
@@ -256,7 +256,9 @@ protected function _setupTable(
 
         $hasSchema = $fixture instanceof TableSchemaAwareInterface && $fixture->getTableSchema() instanceof TableSchema;
 
-        if (($drop && $exists) || ($exists && $hasSchema)) {
+        if ($exists && !$fixture->import && !$fixture->fields) {
```
markstory, line 259: The interface for fixtures doesn't have these properties. You should be able to use the $hasSchema variable from like 257 instead as it represents whether or not the fixture defined any kind of schema.

markstory (commented):
The sqlite build failures are concerning and indicate that these changes may create different problems as well. On second look those sqlite failures in travis are related to PHP8 which has been pretty flaky lately.

src/TestSuite/Fixture/FixtureManager.php
jonmc1: Oh ok, I see. But using $hasSchema isn't working because the fixture's _schema property is always being set, even when both the imports and fields properties are empty, it's set by this method: `Cake\TestSuite\Fixture\TestFixture::_schemaFromReflection()`

othercorey: Maybe we can just change `TestFixture::drop()` and truncate if there are no fields set instead of only returning true. That seems to be the expected behavior.

This does tie the FixtureManager and TestFixture implementation together because we expect `drop()` before `create()` if the table exists. I don't think we want to truncate in `create()` as well.

markstory: > Is there any way to determine from the TableSchema object if the schema was built directly from a datasource?

Right now there isn't we would need to expand the TableSchemaAwareInterface to have that if if we need it.

PR #14969, opened 2020-09-08
Restrict Validation::time() so it requires minutes not just hours

othercorey:
I am proposing this as the alternative to https://github.com/cakephp/cakephp/pull/14930. I don't think most users expect hours only to pass validation in a 24-hour format.

PHP DateTime will not parse hours-only when passed to the constructor so it would require a custom format for users that want to accept this input.

https://3v4l.org/iYb3Q

Did not adjust the am/pm (meridian) format which will validate something like "1 pm". These formats are never marshaled so I don't know when users expect to use them.

### DIFF
--- a/src/Validation/Validation.php
+++ b/src/Validation/Validation.php
@@ -627,7 +627,7 @@ class Validation
 
     /**
      * Time validation, determines if the string passed is a valid time.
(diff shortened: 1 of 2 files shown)


markstory:
Did you mean as an alternative to #14929?

othercorey:
Either way. I'm suggesting we treat it as an implementation bug not a documentation bug.

ADmad (approved):

PR #14970, opened 2020-09-10
Add container support to CommandFactory

markstory:
This enables commands to have dependencies injected as constructor arguments.

Refs #14865
Refs #14945

### DIFF
--- a/src/Console/CommandFactory.php
+++ b/src/Console/CommandFactory.php
@@ -14,6 +14,7 @@ declare(strict_types=1);
  */
 namespace Cake\Console;
 
+use Cake\Core\ContainerInterface;
 use InvalidArgumentException;
 
 /**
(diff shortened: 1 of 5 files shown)


src/Console/CommandFactory.php
```diff
@@ -24,12 +25,32 @@
...
+    }
+
     /**
      * @inheritDoc
      */
     public function create(string $className)
     {
-        $command = new $className();
+        if ($this->container && $this->container->has($className)) {
+            $command = $this->container->get($className);
```
othercorey, line 49: Is this going to be a common pattern? Should we make a helper for it?

src/Console/CommandRunner.php
```diff
@@ -368,6 +369,14 @@ protected function runShell(Shell $shell, array $argv)
      */
     protected function createCommand(string $className, ConsoleIo $io)
     {
+        if (!$this->factory) {
```
ADmad, line 372: Static analysers will be happy if this is moved into the constructor.

markstory: Hrm. Doing that in the constructor would result in the container being built before the application's bootstrap was called which will not work great.

ADmad (approved):

PR #14972, opened 2020-09-10
Drop `+` from `hal+json` and `hal+xml` for routing file extension.

cnizzardini:
Drop `+` from `hal+json` and `hal+xml` for routing file extension. If a user wants to enable routing by extension, then I think having the plus in there is quite odd. In the absence of a standard on this I think this is fair.

See original PR https://github.com/cakephp/cakephp/pull/14867


ADmad:
Are using URL extensions normally used? I would expect the URL to be `.json`/`.xml` only.

ADmad:
Well adding things which aren't normally used increases unnecessary bloat.

othercorey:
I am ok with this since it looks like other types. Is there any code or framework that actually uses hal that we could ask about this for consistency? Someone else must be using this as a standard over us. Can we ask the owner of the type in the spec?

cnizzardini:
Not to my knowledge, symfony api platform does allow routes by `.jsonld` in their demo here https://demo.api-platform.com/, but when I try with `.haljson` its a no-go.

cnizzardini:
We're decreasing bloat by removing two characters ;-)

markstory:
Having the `+` makes URL generation tricky as + also overlaps with the URL encoded space.

PR #14973, opened 2020-09-10
Logs RedisException message to error log on connection failure.

cnizzardini:
On a connection failure to the Redis server cake currently displays the following warning:

> Cache engine %s is not properly configured.

This is not particularly useful in determining the problem. This PR adds the RedisException message to the error log and informs the user to check there. Probably some better ways of going about this, but this is a step in the right direction IMO.

### DIFF
--- a/src/Cache/CacheRegistry.php
+++ b/src/Cache/CacheRegistry.php
@@ -88,7 +88,7 @@ class CacheRegistry extends ObjectRegistry
         if (!$instance->init($config)) {
             throw new RuntimeException(
                 sprintf(
-                    'Cache engine %s is not properly configured. Check log for additional information.',
+                    'Cache engine %s is not properly configured. Check error log for additional information.',
                     get_class($instance)
                 )
             );
--- a/src/Cache/CacheRegistry.php
+++ b/src/Cache/CacheRegistry.php
@@ -87,7 +87,10 @@ class CacheRegistry extends ObjectRegistry
 
         if (!$instance->init($config)) {
             throw new RuntimeException(
-                sprintf('Cache engine %s is not properly configured.', get_class($instance))
+                sprintf(
+                    'Cache engine %s is not properly configured. Check log for additional information.',
+                    get_class($instance)
+                )
             );
         }
 
--- a/src/Cache/Engine/RedisEngine.php
+++ b/src/Cache/Engine/RedisEngine.php
@@ -118,7 +118,7 @@ class RedisEngine extends CacheEngine
             }
         } catch (RedisException $e) {
             if (class_exists(Log::class)) {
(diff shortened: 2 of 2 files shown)


src/Cache/Engine/RedisEngine.php
```diff
@@ -116,6 +117,8 @@ protected function _connect(): bool
                 );
             }
         } catch (RedisException $e) {
+            Log::error('RedisEngine RedisException: ' . $e->getMessage());
```
ADmad, line 120: This needs to be wrapped in a `class_exists(Log::class)` check other wise we will have to add `cakephp/log` as a dependency of the `cakephp/cache` package.

src/Cache/CacheRegistry.php
```diff
@@ -87,7 +87,10 @@ protected function _create($class, string $alias, array $config): CacheEngine
 
         if (!$instance->init($config)) {
             throw new RuntimeException(
-                sprintf('Cache engine %s is not properly configured.', get_class($instance))
+                sprintf(
+                    'Cache engine %s is not properly configured. Check log for additional information.',
```
ADmad, line 91:
```suggestion
                    'Cache engine %s is not properly configured. Check error log for additional information.',
```

othercorey:
I think this is ok for master since it is just a log change.

src/Cache/Engine/RedisEngine.php
cnizzardini: Thanks. Should this use the FQN?

```php
            if (class_exists(\Cake\Log\Log::class)) {
                \Cake\Log\Log::error('RedisEngine RedisException: ' . $e->getMessage());
            }
```

Or is importing the class fine?

othercorey: You can `use` the class normally as long as it's wrapped in class_exists since it might fail to load.

cnizzardini:
Changing base to master creates conflict in travis.yaml. I don't feel comfortable changing that.

othercorey:
> Changing base to master creates conflict in travis.yaml. I don't feel comfortable changing that.

You have to rebase the branch onto master then you can force-push and the merge will clean up.

src/Cache/Engine/RedisEngine.php
```diff
@@ -116,6 +117,10 @@ protected function _connect(): bool
                 );
             }
         } catch (RedisException $e) {
+            if (class_exists(Log::class)) {
+                Log::error('RedisEngine RedisException: ' . $e->getMessage());
```
markstory, line 121:
```suggestion
                Log::error('RedisEngine could not connect. Got error: ' . $e->getMessage());
```

cnizzardini: Agreed, this is strictly related to establishing a connection.

PR #14977, opened 2020-09-14
Add Configuration wrapper.

markstory:
This wrapper would provide a typehintable interface for configuration data and the DI container. Typehinting against array configuration is not possible with most container implementations. If folks are happy with this approach a Configuration instance can be added to the default container as a shared service.

Refs #14865
Refs #14945

### DIFF
--- a/src/Core/Configuration.php
+++ b/src/Core/Configuration.php
@@ -25,7 +25,6 @@ namespace Cake\Core;
  */
 class Configuration
 {
-
     /**
      * Read a configuration key
      *
@@ -42,10 +41,10 @@ class Configuration
      * Check if $path exists and has a non-null value.
      *
      * @param string $path The path to check.
-     * @return boolean True if the configuration data exists.
+     * @return bool True if the configuration data exists.
      */
-    public function has(string $key): bool
+    public function has(string $path): bool
     {
-        return Configure::check($key);
+        return Configure::check($path);
     }
 }
--- a/src/Core/Configuration.php
+++ b/src/Core/Configuration.php
@@ -1,0 +1,51 @@
+<?php
+declare(strict_types=1);
+
+/**
+ * CakePHP(tm) : Rapid Development Framework (https://cakephp.org)
+ * Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
+ *
+ * Licensed under The MIT License
+ * For full copyright and license information, please see the LICENSE.txt
+ * Redistributions of files must retain the above copyright notice.
+ *
+ * @copyright     Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
+ * @link          https://cakephp.org CakePHP(tm) Project
+ * @since         4.2.0
+ * @license       https://opensource.org/licenses/mit-license.php MIT License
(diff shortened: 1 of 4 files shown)


src/Core/Configuration.php
```diff
@@ -0,0 +1,50 @@
...
+namespace Cake\Core;
+
+/**
+ * Read-only wrapper for configuration data
+ *
+ * Intended for use with Cake\Core\Container as
+ * a typehintable way for services to have application
+ * confiugration injected as arrays cannot be typehinted.
+ */
+class Configuration
```
othercorey, line 26: This seems like a potential source of confusion for users.

robertpustulka: `ConfigurationContainer` maybe?

dereuromark: Wouldnt it be ContainerConfiguration then maybe?

othercorey: Instead of just trying to rename Configure (which is the confusing part), we could create a `Container\ConfigureSource` that wraps Configure in in a similar way.

markstory: I agree the name is awkward right now. I would like to keep it somewhat short though as I suspect people will be referencing this quite a bit in their service signatures.

markstory:
An example of how I see this being used is in application services that need some configuration. For example, a service that encapsulates an application's API requests to GitHub or Stripe. These services would need API keys and URLs based on configuration.

```php
$config = Configure::read('Stripe');
$stripe = new StripeService($config);
$plans = $stripe->getPlans();

// constructor would look like
public function __construct(array $config)
{
  $this->apiKey = $config['apiKey'];
  $this->host = $config['host'];
}
```

Is an example of a building one such service and using it (without a DI container). Because we can't typehint arrays, we would need to pre-read all the configuration when defining the container like:

```php
$container->add(StripeService::class)
  ->addArgument(Configure::read('Stripe'));
```

I worry that this doesn't play well with autowiring containers, and adds overhead to an application that has many configuration dependent services.

With the proposed class the `StripeService` could be refactored to accept a `Configuration` object in its constructor, and extract the necessary configuration out of that object. The container definition would now look like:

```php
// With autowiring
$container->add(StripeService::class);

// Without autowiring.
$container->add(StripeService::class)
  ->addArgument(Configuration::class);
```
 
The constructor of the `StripeService` would need to be refactored to

```php
public function __construct(Configuration $config)
{
  $this->apiKey = $config->read('Stripe.apiKey');
  $this->host = $config->read('Stripe.host');
}
```

This does move some of the 'pull' operations into each service which isn't ideal. We could potentially solve that with 'scoped' Configuration proxies but I've personally not needed to restrict configuration access in services before. 

I could be taking a sub-optimal approach to this problem, and any alternatives that help solve the problem are very welcome :smile:

markstory:
Some other names that come to mind (including those suggested earlier)

* ConfigurationContainer
* ContainerConfiguration
* InjectedConfig
* ServiceConfig
* ConfigureData

There are likely others but this is a list I could come up with.

ADmad:
`ServiceConfig` seems short yet descriptive enough.

PR #14980, opened 2020-09-18
Fix sameSite for PHP 7.3+

nickygerritsen:
We have some webapps that need to set the `sameSite` option for a cookie, which we do by setting the path to `/; SameSite=None`. This works on PHP < 7.3 but breaks on 7.3+.
This PR should add a fix for that.

I saw that on `master` the `Cookie` object has a sameSite property but the 4.x code also changes the path to have `; sameSite=x` in it. I can also submit a PR for `master` if this one seems good.

### DIFF
--- a/src/Http/ResponseEmitter.php
+++ b/src/Http/ResponseEmitter.php
@@ -202,15 +202,39 @@ class ResponseEmitter implements EmitterInterface
     {
         foreach ($cookies as $cookie) {
             if (is_array($cookie)) {
-                setcookie(
-                    $cookie['name'],
-                    $cookie['value'],
-                    $cookie['expire'],
-                    $cookie['path'],
-                    $cookie['domain'],
-                    $cookie['secure'],
-                    $cookie['httpOnly']
-                );
+                if (PHP_VERSION_ID >= 70300) {
+                    // PHP 7.3 and up don't support adding SameSite to the path,
+                    // but require the alternative setcookie syntax.
+                    // So we extract sameSite from the path
+                    $sameSite = null;
+                    $path = $cookie['path'];
+                    if (preg_match('/^(.*); SameSite=(.*)$/', $path, $matches) === 1) {
+                        $path = $matches[1];
+                        $sameSite = $matches[2];
+                    }
+                    setcookie(
+                        $cookie['name'],
+                        $cookie['value'],
+                        [
+                            'expires' => $cookie['expire'],
+                            'path' => $path,
+                            'domain' => $cookie['domain'],
+                            'secure' => $cookie['secure'],
+                            'httponly' => $cookie['httpOnly'],
+                            'samesite' => $sameSite,
+                        ]
+                    );
+                } else {
+                    setcookie(
+                        $cookie['name'],
+                        $cookie['value'],
+                        $cookie['expire'],
+                        $cookie['path'],
+                        $cookie['domain'],
+                        $cookie['secure'],
+                        $cookie['httpOnly']
+                    );
+                }
                 continue;
             }
 
@@ -230,6 +254,7 @@ class ResponseEmitter implements EmitterInterface
                 'domain' => '',
                 'secure' => false,
                 'httponly' => false,
+                'samesite' => null,
             ];
 
             foreach ($parts as $part) {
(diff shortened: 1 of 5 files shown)


src/Http/ResponseEmitter.php
```diff
@@ -202,15 +202,39 @@ protected function emitCookies(array $cookies)
...
+                        $cookie['name'],
+                        $cookie['value'],
+                        [
+                            'expires' => $cookie['expire'],
+                            'path' => $path,
+                            'domain' => $cookie['domain'],
+                            'secure' => $cookie['secure'],
+                            'httponly' => $cookie['httpOnly'],
+                            'samesite' => $sameSite,
+                        ],
```
ADmad, line 225:
```suggestion
                        ]
```

This causes syntax error on PHP < 7.4

nickygerritsen: Woops my bad

src/Http/ResponseEmitter.php
```diff
@@ -202,15 +202,39 @@ protected function emitCookies(array $cookies)
...
-                setcookie(
-                    $cookie['name'],
-                    $cookie['value'],
-                    $cookie['expire'],
-                    $cookie['path'],
-                    $cookie['domain'],
-                    $cookie['secure'],
-                    $cookie['httpOnly']
-                );
+                if (version_compare(PHP_VERSION, '7.3.0', '>=')) {
```
ADmad, line 205:
```suggestion
                if (PHP_VERSION_ID >= 70300) {
```

Using `PHP_VERSION_ID` would be more efficient.

nickygerritsen:
Updated the existing tests and added an extra cookie to `testEmitResponseCookies` to also test this.

tests/test_app/TestApp/Application.php
```diff
@@ -26,6 +26,18 @@ class Application extends BaseApplication
      */
     public function bootstrap()
     {
+        $this->addPlugin('TestPlugin',
```
ADmad, line 29: What's all this?

nickygerritsen: I have no clue where this comes from...

src/Http/ResponseEmitter.php
nickygerritsen: I will update it. Note that I copied it from `Session.php` line `116`, so we might want to also update that at some point.

tests/test_app/TestApp/Application.php
nickygerritsen: Ah this happens when I run all the test locally and they fail (since I have no DB set up)

nickygerritsen:
Took a bit but the tests pass :)

src/Http/Session.php
```diff
@@ -604,15 +604,39 @@ public function renew()
...
+            setcookie(
+                session_name(),
+                '',
+                [
+                    'expires' => time() - 42000,
+                    'path' => $path,
+                    'domain' => $params['domain'],
+                    'secure' => $params['secure'],
+                    'httponly' => $params['httponly'],
+                    'samesite' => $sameSite,
```
ndm2, line 626: Is this actually necessary for "deleting" the cookie?

I'm not sure if I remember this correctly, but isn't only the path and the domain required to match for a given cookie name in order to be able to update the cookie?

nickygerritsen: Ah that I don't know. I could remove this here indeed if we don't need the samesite. But do we then need the other params?

ndm2: I'm not sure about that, hence why I'm asking ;)

The PHP docs say:

> Cookies must be deleted with the same parameters as they were set with. If the value argument is an empty string, or FALSE, and all other arguments match a previous call to setcookie, then the cookie with the specified name will be deleted from the remote client. This is internally achieved by setting value to 'deleted' and expiration time to one year in past. 

But from looking at the source this is already not quite correct, the expiration date is different, and it only depends on the value, not on any other arguments: https://github.com/php/php-src/blob/php-7.4.10/ext/standard/head.c#L110-L122

From a quick test it seems that it's not required to specify all the same attributes, [**which is also what the RFC suggests**](https://tools.ietf.org/html/draft-ietf-httpbis-rfc6265bis-05#section-3.1), I can freely change `HttpOnly`, `Secure` and `SameSite`, PHP will generate the same `deleted` cookie value and `Thu, 01-Jan-1970 00:00:01 GMT` expiration value, and the browser will delete the cookie just fine (except of course for combinations like `SameSite=None` without `Secure` which would be rejected by the browser).

But who knows, maybe there are actually cases that require the same exact attributes.

markstory (commented):
Given that we have ways of testing cookies being sent via the server_mocks it would be preferrable to have a test covering the behavior of defining SameSite in the path attribute. Right now that feels a bit fragile.

nickygerritsen:
> Given that we have ways of testing cookies being sent via the server_mocks it would be preferrable to have a test covering the behavior of defining SameSite in the path attribute. Right now that feels a bit fragile.

I can try to add such a test, but do you think the current setup of splitting the path in the mock makes sense?

src/Http/Session.php
nickygerritsen: I'd say it never hurts to have all attributes? Then it might be better to leave it in.

othercorey:
According to codecov we aren’t testing several of these code paths.

nickygerritsen:
> According to codecov we aren’t testing several of these code paths. 

Some can be fixed if I add a test for parsing the path. The others aren't covered because we only do coverage on 7.4 I think

markstory:
I don't love splitting the path. Adding the `withSameSite` and `getSameSite` methods to `Http\Cookie` and using `@method` annotations in the CookieInterface would be another approach. That would be more forwards compatible than path splitting would be.

nickygerritsen:
Closed in favor of #14987

PR #14982, opened 2020-09-19
Add docs for runCommand + container

markstory:
Add notes on the limitations of runCommand() and the container. Because commands don't have access to the container they are unable resolve dependencies. Instead commands should be defined in the container and then resolved as dependencies to the calling command.

Refs #14865


othercorey:
I think we need to put any container-specific notes under a common "When Using Containers" header or something to give context.

markstory:
@othercorey I was thinking of doing that in the book for both the container and runCommand docs.

PR #14987, opened 2020-09-21
3.x cookie samesite

ADmad:
Refs https://github.com/cakephp/cakephp/pull/14980#issuecomment-695880618.

Normally these changes would go into a new minor but since we don't have any plans for doing a `3.10` currently and this has started to become an issue for users, we can probably make an exception and make these changes in a backwards compatible bugfix release of `3.9`.

### DIFF
--- a/src/Http/Cookie/Cookie.php
+++ b/src/Http/Cookie/Cookie.php
@@ -526,7 +526,7 @@ class Cookie implements CookieInterface
     {
         if (!in_array($sameSite, CookieInterface::SAMESITE_VALUES, true)) {
             throw new InvalidArgumentException(
-                'Samesite value must be either of: ' . implode(', ', CookieInterface::SAMESITE_VALUES)
+                'SameSite value must be either of: ' . implode(', ', CookieInterface::SAMESITE_VALUES)
             );
         }
     }
--- a/src/Http/Cookie/Cookie.php
+++ b/src/Http/Cookie/Cookie.php
@@ -104,6 +104,13 @@ class Cookie implements CookieInterface
     protected $httpOnly = false;
 
     /**
+     * Samesite
+     *
+     * @var string|null
+     */
+    protected $sameSite = null;
+
+    /**
      * Constructor
      *
      * The constructors args are similar to the native PHP `setcookie()` method.
@@ -118,6 +125,7 @@ class Cookie implements CookieInterface
      * @param string $domain Domain
      * @param bool $secure Is secure
      * @param bool $httpOnly HTTP Only
+     * @param string|null $sameSite Samesite
      */
     public function __construct(
         $name,
@@ -126,7 +134,8 @@ class Cookie implements CookieInterface
         $path = '/',
         $domain = '',
         $secure = false,
-        $httpOnly = false
+        $httpOnly = false,
+        $sameSite = null
     ) {
         $this->validateName($name);
         $this->name = $name;
@@ -144,6 +153,12 @@ class Cookie implements CookieInterface
 
         $this->validateBool($secure);
         $this->secure = $secure;
+
+        if ($sameSite) {
+            $this->validateSameSiteValue($sameSite);
+            $this->sameSite = $sameSite;
+        }
+
         if ($expiresAt) {
             $expiresAt = $expiresAt->setTimezone(new DateTimeZone('GMT'));
         }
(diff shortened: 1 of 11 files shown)


dereuromark:
I am still strongly for a 3.10 release, given also other required backports and open PRs waiting to go into it.
Not no mention what still is to come here most likely to further keep also 3.x maintained along the 4.1+

ADmad:
For this particular set of changes I don't really care in which release of Cake 3 they go. But given that browsers have started to become quite finicky about the `SameSite` attribute, I think a lot of users would prefer these changes are released sooner rather wait for other 4.x changes to be backported.

src/Http/Cookie/Cookie.php
```diff
@@ -466,6 +484,51 @@ public function withExpired()
...
+     *
+     * @param string $sameSite SameSite value
+     * @return void
+     * @throws \InvalidArgumentException
+     */
+    protected static function validateSameSiteValue($sameSite)
+    {
+        if (!in_array($sameSite, CookieInterface::SAMESITE_VALUES, true)) {
+            throw new InvalidArgumentException(
+                'Samesite value must be either of: ' . implode(', ', CookieInterface::SAMESITE_VALUES)
```
othercorey, line 527:
```suggestion
                'SameSite value must be either of: ' . implode(', ', CookieInterface::SAMESITE_VALUES)
```
Should this be the same spelling as constructor parameter, attribute name or normal english?

src/Http/Cookie/Cookie.php
```diff
@@ -466,6 +484,51 @@ public function withExpired()
...
+        return $this->sameSite;
+    }
+
+    /**
+     * Create a cookie with an updated SameSite option.
+     *
+     * @param string|null $sameSite Value for to set for Samesite option.
+     *   One of CookieInterface::SAMESITE_* constants.
+     * @return static
+     */
```
othercorey, line 503: Add throws tag.

src/Http/Cookie/Cookie.php
```diff
@@ -118,6 +125,7 @@ class Cookie implements CookieInterface
      * @param string $domain Domain
      * @param bool $secure Is secure
      * @param bool $httpOnly HTTP Only
+     * @param string|null $sameSite Samesite
      */
```
othercorey, line 129: Add throws tag.

src/Http/Cookie/Cookie.php
dereuromark: if it is the key, I would code fence it: `samesite` etc.

markstory:
I'm not against a 3.10 release, but I think deferring this until 3.10 isn't as useful as having it in 3.9. While normally I don't like adding features in minor releases, needing to specify the samesite attribute is an external requirement that is mostly out of our hands.

markstory (approved):

nickygerritsen:
Ha I was just about to update #14980 and I saw you already created a new PR.
What about the changes to the session? We also use it there, currently by setting `cookiePath` in the config to `/; SameSite=none`, but that also gives issues.

ADmad:
> What about the changes to the session?

Done

src/Http/Session.php
```diff
@@ -604,15 +604,25 @@ public function renew()
...
-        setcookie(
-            session_name(),
-            '',
-            time() - 42000,
-            $params['path'],
-            $params['domain'],
-            $params['secure'],
-            $params['httponly']
-        );
+        if (PHP_VERSION_ID >= 70300) {
```
nickygerritsen, line 607: I wonder if we should use https://www.php.net/manual/en/function.session-set-cookie-params.php somewhere to actually set the samesite option. Since we now only 'copy' it and never set it.

But perhaps we leave that to a new PR?

ADmad: This particular `setcookie()` call is deleting previous cookie when renewing session, I doubt `session_set_cookie_params()` works for that case, plus as the doc mentions it needs to called before `session_start()` is called which isn't the case either here.

nickygerritsen: Sorry I meant in the `start` method. But it seems we can use `defaults.session.cookie_samesite` for that in the config.

nickygerritsen: Just tested, this:

```php
    'Session'        => [
        'defaults'   => 'php',
        'ini' => [
            'session.cookie_samesite' => 'None',
        ],
    ],
```

indeed works for the session cookie :)

PR #14989, opened 2020-09-23
Throw an exception on invalid cache configuration

### LINKED ISSUE #14985: Failure to setup second Memcached instance on another port
This is a (multiple allowed):

* [x] bug
* [ ] enhancement

* CakePHP Version: 3.x.
* Platform and Target: Ubuntu, MySQL, Apache.

### What you did
1. Started another memcached instance on port 11213 by running `memcached -p 11213 -d` in the console.
2. Configured `'_cake_core_'` cache to use that instance by using `'servers' => ['localhost:11213'],`

### What happened
`'_cake_core_'` used the default memcached on port 11211 instead 11213.

The line [157](https://github.com/cakephp/cakephp/blob/master/src/Cache/Engine/MemcachedEngine.php#L157) `if (!$this->_Memcached->addServers($servers)) {` seems to be never executed because `$this->_Memcached->getServerList()` on the line [148](https://github.com/cakephp/cakephp/blob/master/src/Cache/Engine/MemcachedEngine.php#L148) in MemcachedEngine always  returns `true`.

I would expect `'_cake_core_'`  to use 'localhost:11213' memcached.

markstory:
* Add test case for the memcached config on the second port, #14985
* Port changes from #14986 to 4.x

Fixes #14985

### DIFF
--- a/src/Cache/Engine/MemcachedEngine.php
+++ b/src/Cache/Engine/MemcachedEngine.php
@@ -145,7 +145,25 @@ class MemcachedEngine extends CacheEngine
         }
         $this->_setOptions();
 
-        if (count($this->_Memcached->getServerList())) {
+        $serverList = $this->_Memcached->getServerList();
+        if (count($serverList)) {
+            if ($this->_config['persistent'] !== false) {
+                $actualServers = [];
+                foreach ($serverList as $server) {
+                    $actualServers[] = $server['host'] . ':' . $server['port'];
+                }
+                unset($server);
+                sort($actualServers);
+                $configuredServers = $this->_config['servers'];
+                sort($configuredServers);
+                if ($actualServers !== $configuredServers) {
+                    $message = "Invalid cache configuration. Multiple persistent cache configurations are detected" .
+                        " with different 'servers' values. 'servers' values for persistent cache configurations" .
+                        " must be the same when using the same persistence id.";
+                    throw new InvalidArgumentException($message);
+                }
+            }
+
             return true;
         }
 
--- a/src/Cache/Engine/MemcachedEngine.php
+++ b/src/Cache/Engine/MemcachedEngine.php
@@ -147,7 +147,7 @@ class MemcachedEngine extends CacheEngine
 
         $serverList = $this->_Memcached->getServerList();
         if ($serverList) {
-            if ($this->_config['persistent'] !== false) {
+            if ($this->_Memcached->isPersistent()) {
                 foreach ($serverList as $server) {
                     if (!in_array($server['host'] . ':' . $server['port'], $this->_config['server'], true)) {
                         throw new InvalidArgumentException(
--- a/src/Cache/Engine/MemcachedEngine.php
+++ b/src/Cache/Engine/MemcachedEngine.php
@@ -157,9 +157,9 @@ class MemcachedEngine extends CacheEngine
                 $configuredServers = $this->_config['servers'];
                 sort($configuredServers);
                 if ($actualServers !== $configuredServers) {
-                    $message = "Invalid cache configuration. Multiple persistent cache configurations are detected" .
-                        " with different 'servers' values. 'servers' values for persistent cache configurations" .
-                        " must be the same when using the same persistence id.";
+                    $message = 'Invalid cache configuration. Multiple persistent cache configurations are detected' .
+                        ' with different `servers` values. `servers` values for persistent cache configurations' .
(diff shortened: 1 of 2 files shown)


src/Cache/Engine/MemcachedEngine.php
```diff
@@ -145,7 +145,25 @@ public function init(array $config = []): bool
         }
         $this->_setOptions();
 
-        if (count($this->_Memcached->getServerList())) {
+        $serverList = $this->_Memcached->getServerList();
+        if (count($serverList)) {
```
othercorey, line 149:
```suggestion
        if ($serverList) {
```

src/Cache/Engine/MemcachedEngine.php
```diff
@@ -145,7 +145,25 @@ public function init(array $config = []): bool
...
+            if ($this->_config['persistent'] !== false) {
```
othercorey, line 150:
```suggestion
        if ($serverList && $this->_config['persistent'] !== false) {
```

src/Cache/Engine/MemcachedEngine.php
othercorey, line 149:
```suggestion
        if ($serverList) {
```

src/Cache/Engine/MemcachedEngine.php
```diff
@@ -145,7 +145,25 @@ public function init(array $config = []): bool
...
+                $actualServers = [];
+                foreach ($serverList as $server) {
+                    $actualServers[] = $server['host'] . ':' . $server['port'];
+                }
+                unset($server);
```
othercorey, line 155: why is this unset?

src/Cache/Engine/MemcachedEngine.php
othercorey, line 150:
```suggestion
            if ($this->_config['persistent']) {
```
null shouldn't be persistent, right?

othercorey: Can we use `$this->_Memcached->isPersistent()` here?

src/Cache/Engine/MemcachedEngine.php
```diff
@@ -145,7 +145,25 @@ public function init(array $config = []): bool
...
+                sort($actualServers);
+                $configuredServers = $this->_config['servers'];
+                sort($configuredServers);
+                if ($actualServers !== $configuredServers) {
+                    $message = 'Invalid cache configuration. Multiple persistent cache configurations are detected' .
+                        ' with different `servers` values. `servers` values for persistent cache configurations' .
+                        ' must be the same when using the same persistence id.';
+                    throw new InvalidArgumentException($message);
+                }
```
othercorey, line 164:
```suggestion
                foreach ($serverList as $server) {
                    if (!in_array($server['host'] . ':' . $server['port'], $this->_config['server'], true)) {
                        throw new InvalidArgumentException(
                            'Invalid cache configuration. Multiple persistent cache configurations are detected' .
                            ' with different `servers` values. `servers` values for persistent cache configurations' .
                            ' must be the same when using the same persistence id.'
                        );
                    }
                }
                $actualServers = [];
                foreach ($serverList as $server) {
                    $actualServers[] = $server['host'] . ':' . $server['port'];
                }
                unset($server);
                sort($actualServers);
                $configuredServers = $this->_config['servers'];
                sort($configuredServers);
                if ($actualServers !== $configuredServers) {
                    $message = 'Invalid cache configuration. Multiple persistent cache configurations are detected' .
                        ' with different `servers` values. `servers` values for persistent cache configurations' .
                        ' must be the same when using the same persistence id.';
                    throw new InvalidArgumentException($message);
                }
```

othercorey: I don't know if that's the best exception message, but I'm not familiar with it.

tests/TestCase/Cache/Engine/MemcachedEngineTest.php
```diff
@@ -374,6 +374,32 @@ public function testSaslAuthException()
         $MemcachedEngine->init($config);
     }
 
+    /**
+     * testConfigDifferentPorts method
+     *
+     * @expectedException InvalidArgumentException
```
othercorey, line 380: I believe this tags are deprecated. Use $this->expectException().

src/Cache/Engine/MemcachedEngine.php
markstory: We might be able to.

Issue #14897, opened 2020-08-12
Should ExceptionRenderer only set HTTP code if a cake Exception?


othercorey:
Should the exception code only be interpreted as an HTTP code if the exception extends `Cake\Core\Exception\Exception` which documents the code explicitly as an HTTP code?

**Exception.php**
```
     * @param string|array $message Either the string of the error message, or an array of attributes
     *   that are made available in the view, and sprintf()'d into Exception::$_messageTemplate
     * @param int|null $code The code of the error, is also the HTTP status code for the error.
     * @param \Throwable|null $previous the previous exception.
     */
    public function __construct($message = '', ?int $code = null, ?Throwable $previous = null)
```

**ExceptionRenderer**
```
    protected function _code(Throwable $exception): int
    {
        $code = 500;
        $errorCode = (int)$exception->getCode();
        if ($errorCode >= 400 && $errorCode < 600) {
            $code = $errorCode;
        }

        return $code;
    }
```

dereuromark:
What's the reason? If you didn't set one it will be 5xx as expected already, won't it?
I don't see an issue with current way so far.

othercorey:
There are many exception codes outside of cake's internal exceptions.

ADmad:
> There are many exception codes outside of cake's internal exceptions.

Are there any non HTTP codes within the 400-599 range?

othercorey:
> Are there any non HTTP codes within the 400-599 range?

I have no idea what other codes like PDOException throw. That's why I'm asking this.

garas:
More logical would be only use `\Cake\Http\Exception\HttpException` error code for HTTP code, any other exception would be always 500.

That would require change parents of few exceptions:

Change parent of `\Cake\Controller\Exception\MissingActionException` to `\Cake\Http\Exception\NotFoundException`.

Change parent to `\Cake\Http\Exception\NotFoundException`, but `cakephp/datasource` would require new dependency `cakephp/http`:
- `\Cake\Datasource\Exception\InvalidPrimaryKeyException`
- `\Cake\Datasource\Exception\PageOutOfBoundsException`
- `\Cake\Datasource\Exception\RecordNotFoundException`

Change parent of `\Cake\Http\Exception\InvalidCsrfTokenException` to `\Cake\Http\Exception\ForbiddenException`.

Change parent of `\Cake\Routing\Exception\RedirectException` (deprecated) to `\Cake\Http\Exception\RedirectException`.

[...]

PR #14996, opened 2020-09-26
Added interface to identify http code exceptions

othercorey:
Refs https://github.com/cakephp/cakephp/issues/14897

This adds `Cake\Core\Exception\HttpExceptionCodeInterface` which identifies exceptions that want their exception code used for the http response code. `Exception::responseHeader()` has been split into `getResponseHeaders()` and `addResponseHeaders()`.

`Cake\Core\Exception\HttpExceptionCodeTrait` implements the http response header functions for exceptions that cannot inherit from HttpException directly.

Several exceptions have changed to `HttpException` or `HttpExceptionCodeInterface` so we know which exceptions are setting a valid http response code.

## Issues

`ExceptionRenderer` tries to render custom templates for non-http exceptions, but now some of these exceptions are considered http exceptions. These now render the generic error400 template.

`ExceptionRenderer::_template()` and `ExceptionRenderer::_method()` sort of duplicate the logic of determining which template to render. Neither one allows for http-specific exception templates. I'm not sure how to fix this up -- doesn't seem like we need both methods.


othercorey:
I haven't switched existing code over to the new `addResponseHeaders()` function yet. I can't just forward the function calls in `Exception::responseHeader()` because we need to deprecate anyone using a non-http exception first. We can't guarantee user exceptions will extend `HttpException` until cakephp 5.

tests/TestCase/Error/ExceptionRendererTest.php
```diff
@@ -130,8 +130,8 @@ public function testTemplatePath()
 
         $ExceptionRenderer->render();
         $controller = $ExceptionRenderer->__debugInfo()['controller'];
-        $this->assertSame('missingAction', $controller->viewBuilder()->getTemplate());
-        $this->assertSame('Error', $controller->viewBuilder()->getTemplatePath());
+        $this->assertSame('error400', $controller->viewBuilder()->getTemplate());
+        $this->assertSame('Admin/Error', $controller->viewBuilder()->getTemplatePath());
```
othercorey, line 134: This is now a http exception.

tests/TestCase/Error/ExceptionRendererTest.php
```diff
@@ -231,7 +231,7 @@ public function testConstruction()
     public function testExceptionMessageCoercion()
     {
         Configure::write('debug', false);
-        $exception = new MissingActionException('Secret info not to be leaked');
+        $exception = new CakeException('Secret info not to be leaked');
```
othercorey, line 234: http exceptions render the message.

src/Datasource/Exception/InvalidPrimaryKeyException.php
```diff
@@ -17,12 +17,16 @@
...
 
 use Cake\Core\Exception\Exception;
+use Cake\Core\Exception\HttpExceptionCodeInterface;
+use Cake\Core\Exception\HttpExceptionCodeTrait;
 
 /**
  * Exception raised when the provided primary key does not match the table primary key
  */
-class InvalidPrimaryKeyException extends Exception
+class InvalidPrimaryKeyException extends Exception implements HttpExceptionCodeInterface
```
ADmad, line 26: The dbal throwing exceptions with HTTP status code feels pretty weird.

IMO a proper way to handle this would be to remove the 4xx/5xx exception codes from all exceptions exception the ones under http package. The exception renderer should then wrap all non http exception with a http exception with appropriate status code.

othercorey: That does seem cleaner. I don't think any non-http exceptions would need a special http code (just 404 or 500).

I'll wait for other feedback before implementing that since this PR covers a few of the recommendations in the issue.

PR #14997, opened 2020-09-26
Return all possible form context attributes.

ADmad:
Additional attributes could be for e.g. useful in custom widgets.

For e.g. `comment` attribute of a table column could be used to auto generate help text below a form input.

### DIFF
--- a/src/View/Form/ContextInterface.php
+++ b/src/View/Form/ContextInterface.php
@@ -21,6 +21,8 @@ namespace Cake\View\Form;
  */
 interface ContextInterface
 {
+    public const VALID_ATTRIBUTES = ['length', 'precision', 'comment', 'null', 'default'];
+
     /**
      * Get the fields used in the context as a primary key.
      *
--- a/src/View/Form/EntityContext.php
+++ b/src/View/Form/EntityContext.php
@@ -711,7 +711,10 @@ class EntityContext implements ContextInterface
             return [];
         }
 
-        return (array)$table->getSchema()->getColumn(array_pop($parts));
+        $column = (array)$table->getSchema()->getColumn(array_pop($parts));
+        $allowed = ['length' => null, 'precision' => null, 'comment' => null, 'null' => null, 'default' => null];
+
+        return array_intersect_key($column, $allowed);
     }
 
     /**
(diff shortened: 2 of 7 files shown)


src/View/Form/EntityContext.php
```diff
@@ -711,10 +711,7 @@ public function attributes(string $field): array
             return [];
         }
 
-        $column = (array)$table->getSchema()->getColumn(array_pop($parts));
-        $allowed = ['length' => null, 'precision' => null];
```
markstory, line 715: I agree that having the comment could be useful. Why not add only that. One drawback to adding everything is that the interface for attributes gets muddy and hard to keep consistent.

tests/TestCase/View/Form/EntityContextTest.php
```diff
@@ -395,7 +395,10 @@ public function testSchemaOnCollections($collection)
         $this->assertSame('string', $context->type('99.title'));
         $this->assertNull($context->type('0.nope'));
 
-        $expected = ['length' => 255, 'precision' => null];
+        $expected = [
+            'length' => 255, 'precision' => null, 'type' => 'string',
+            'null' => null, 'default' => null, 'comment' => null, 'collate' => null,
```
markstory, line 400: I don't see how the collate, type, or null keys would ever be useful in a form building context.

markstory: I can see how null could be useful for indicating fields are optional. Would you be ok with only exposing null and comment? We have existing APIs to get an abstract type for a field.

ADmad: I have updated `EnityContext` to addtionally expose only `null`, `comment` and `default`.

src/View/Form/ContextInterface.php
```diff
@@ -21,6 +21,8 @@
  */
 interface ContextInterface
 {
+    public const VALID_ATTRIBUTES = ['length', 'precision', 'comment', 'null', 'default'];
```
ADmad, line 24: Wondering if this should be `protected`.

markstory: I would start at protected and open up if we need to. Going from public to protected is harder.

ADmad: Err interface constant can be public only, so can't make it protected :)

markstory: boooo

PR #14998, opened 2020-09-29
Use exception code as http code for HttpException only

othercorey:
Refs https://github.com/cakephp/cakephp/issues/14897

ExceptionRenderer now maps certain non-http exceptions to their http status codes instead of requiring exceptions to extend HttpException or require core Exception to implement http status codes.

~A few exceptions that were returning 404 now return 500.~

### DIFF
--- a/src/Error/ExceptionRenderer.php
+++ b/src/Error/ExceptionRenderer.php
@@ -305,7 +305,8 @@ class ExceptionRenderer implements ExceptionRendererInterface
             $baseClass = substr($baseClass, 0, -9);
         }
 
-        $method = Inflector::variable($baseClass) ?: 'error500';
+        // $baseClass would be an empty string if the exception class is \Exception.
+        $method = $baseClass === '' ? 'error500' : Inflector::variable($baseClass);
 
         return $this->method = $method;
     }
@@ -339,24 +340,15 @@ class ExceptionRenderer implements ExceptionRendererInterface
      * Get template for rendering exception info.
      *
      * @param \Throwable $exception Exception instance.
-     * @param string $method Method name.
+     * @param string $template Template name.
      * @param int $code Error code.
      * @return string Template name
      */
-    protected function _template(Throwable $exception, string $method, int $code): string
+    protected function _template(Throwable $exception, string $template, int $code): string
     {
-        $isHttpException = $exception instanceof HttpException;
-
-        if (!Configure::read('debug') && !$isHttpException || $isHttpException) {
-            $template = 'error500';
-            if ($code < 500) {
-                $template = 'error400';
-            }
-
-            return $this->template = $template;
+        if ($exception instanceof HttpException || !Configure::read('debug')) {
+            return $this->template = $code < 500 ? 'error400' : 'error500';
         }
-
-        $template = $method ?: 'error500';
 
         if ($exception instanceof PDOException) {
             $template = 'pdo_error';
--- a/src/Error/ExceptionRenderer.php
+++ b/src/Error/ExceptionRenderer.php
@@ -340,21 +340,21 @@ class ExceptionRenderer implements ExceptionRendererInterface
      * Get template for rendering exception info.
      *
      * @param \Throwable $exception Exception instance.
-     * @param string $template Template name.
+     * @param string $method Method name.
      * @param int $code Error code.
      * @return string Template name
      */
-    protected function _template(Throwable $exception, string $template, int $code): string
+    protected function _template(Throwable $exception, string $method, int $code): string
     {
         if ($exception instanceof HttpException || !Configure::read('debug')) {
             return $this->template = $code < 500 ? 'error400' : 'error500';
         }
 
         if ($exception instanceof PDOException) {
-            $template = 'pdo_error';
+            return $this->template = 'pdo_error';
         }
 
-        return $this->template = $template;
+        return $this->template = $method;
     }
 
     /**
--- a/src/Error/ExceptionRenderer.php
+++ b/src/Error/ExceptionRenderer.php
@@ -23,7 +23,6 @@ use Cake\Core\App;
 use Cake\Core\Configure;
 use Cake\Core\Exception\Exception as CakeException;
 use Cake\Core\Exception\MissingPluginException;
-use Cake\Datasource\Exception\InvalidPrimaryKeyException;
 use Cake\Datasource\Exception\PageOutOfBoundsException;
 use Cake\Datasource\Exception\RecordNotFoundException;
 use Cake\Event\Event;
@@ -108,7 +107,6 @@ class ExceptionRenderer implements ExceptionRendererInterface
         // Controller exceptions
         MissingActionException::class => 404,
         // Datasource exceptions
-        InvalidPrimaryKeyException::class => 404,
         PageOutOfBoundsException::class => 404,
         RecordNotFoundException::class => 404,
         // Http exceptions
--- a/src/Error/ExceptionRenderer.php
+++ b/src/Error/ExceptionRenderer.php
@@ -18,12 +18,17 @@ namespace Cake\Error;
 
 use Cake\Controller\Controller;
 use Cake\Controller\ControllerFactory;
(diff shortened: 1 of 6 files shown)


tests/TestCase/ExceptionsTest.php
```diff
@@ -166,13 +166,13 @@ public function exceptionProvider()
...
             ['Cake\Database\Exception\NestedTransactionRollbackException', 500],
-            ['Cake\Datasource\Exception\InvalidPrimaryKeyException', 404],
+            ['Cake\Datasource\Exception\InvalidPrimaryKeyException', 500],
             ['Cake\Datasource\Exception\MissingDatasourceConfigException', 500],
             ['Cake\Datasource\Exception\MissingDatasourceException', 500],
             ['Cake\Datasource\Exception\MissingModelException', 500],
             ['Cake\Datasource\Exception\PageOutOfBoundsException', 404],
-            ['Cake\Datasource\Exception\RecordNotFoundException', 404],
-            ['Cake\Mailer\Exception\MissingActionException', 404],
+            ['Cake\Datasource\Exception\RecordNotFoundException', 500],
```
garas, line 174: `RecordNotFoundException` behaviour is mentioned in [Getting a Single Entity by Primary Key](https://book.cakephp.org/4/en/orm/retrieving-data-and-resultsets.html#getting-a-single-entity-by-primary-key) docs and baked by [view.twig](https://github.com/cakephp/bake/blob/master/templates/bake/element/Controller/view.twig) and [edit.twig](https://github.com/cakephp/bake/blob/master/templates/bake/element/Controller/edit.twig), so response code should stay same and handled by `ExceptionRenderer` too.

src/Error/ExceptionRenderer.php
```diff
@@ -341,20 +344,33 @@ protected function _template(Throwable $exception, string $method, int $code): s
...
-        $errorCode = (int)$exception->getCode();
-        if ($errorCode >= 400 && $errorCode < 600) {
-            $code = $errorCode;
+        if ($exception instanceof HttpException) {
+            return (int)$exception->getCode();
         }
 
-        return $code;
+        // Controller exceptions
+        if ($exception instanceof MissingActionException) {
```
ADmad, line 359: It would be better to add a property which maintains a list of exceptions that should return a `404`. That would allow users to easily additional additional ones too as per their requirements.

As garas mentioned below `RecordNotFoundException` should also be in that list.

dereuromark: Refs https://github.com/dereuromark/cakephp-tools/blob/master/src/Error/ErrorHandlerTrait.php#L36

tests/TestCase/ExceptionsTest.php
dereuromark: Jep, this is always a 404.

src/Error/ExceptionRenderer.php
ADmad: If we forsee people needing to map other codes too then we can use array format like
```php
[
    MissingActionException::class => 404,
    SomeOtherException::class => 400,
] 
```

tests/TestCase/ExceptionsTest.php
othercorey: I disagree, but I will implement the existing functionality.

src/Error/ExceptionRenderer.php
othercorey: Done.

tests/TestCase/ExceptionsTest.php
```diff
@@ -172,7 +172,7 @@ public function exceptionProvider()
             ['Cake\Datasource\Exception\MissingModelException', 500],
             ['Cake\Datasource\Exception\PageOutOfBoundsException', 404],
             ['Cake\Datasource\Exception\RecordNotFoundException', 404],
-            ['Cake\Mailer\Exception\MissingActionException', 404],
+            ['Cake\Mailer\Exception\MissingActionException', 500],
```
othercorey, line 175: Is anyone familiar with this exception. It seemed like a bug that calling the wrong mailer method would throw a 404 error. If this is intentional for some reason, let me know.

dereuromark: I agree, that this might have been done based on "missing action" being controller related on normal URLs.
But for mailer it sure sounds like it should be a 500.

src/Error/ExceptionRenderer.php
```diff
@@ -95,6 +95,15 @@ class ExceptionRenderer implements ExceptionRendererInterface
...
     protected $request;
 
+    /**
+     * Map of exceptions to http status codes.
+     *
+     * This can be customized for users that don't want specific exceptions to throw 404 errors
+     * or want their application exceptions to be automatically converted.
+     *
+     * @var array
+     * @psalm-var array<string, int>
```
garas, line 105:
```suggestion
     * @psalm-var array<class-string, int>
```
or even
```suggestion
     * @psalm-var array<class-string<\Throwable>, int>
```

src/Error/ExceptionRenderer.php
```diff
@@ -90,6 +95,26 @@ class ExceptionRenderer implements ExceptionRendererInterface
...
+     */
+    protected $exceptionHttpCodes = [
+        // Controller exceptions
+        MissingActionException::class => 404,
+        // Datasource exceptions
+        InvalidPrimaryKeyException::class => 404,
```
garas, line 111: I'm thinking if `InvalidPrimaryKeyException` should be treated as developer error and return 500.

It is about correct number of passed keys in `$table->get(null)`, `$table->get(1)`, `$table->get([1, 2])`, not value itself, it is responsibility of developer to ensure it is valid size of array, so it is not client side error.

Parameter is always null/string in default baked controller, array can't be passed from URL.

dereuromark: @garas That is not always true in all cases I think
If it comes from e.g. query string, a user or malicious person could most likely also input an array into what goes into such a method.
Depends on how the code has been written.

src/Error/ExceptionRenderer.php
```diff
@@ -90,6 +95,26 @@ class ExceptionRenderer implements ExceptionRendererInterface
...
+        PageOutOfBoundsException::class => 404,
+        RecordNotFoundException::class => 404,
+        // Http exceptions
+        MissingControllerException::class => 404,
```
ADmad, line 115: Instead of needing this I think we should just make `MissingControllerException` extend `HttpException`. It's a bit weird that it's under `Http/Exception/` and extends `Core`Exception`Exception` instead of `Http\ExceptionHttpException`.

src/Error/ExceptionRenderer.php
ADmad: I too am leaning towards `InvalidPrimaryKeyException` being a `500` instead of `404`. It's the devs job to satanize the value coming from the request.

src/Error/ExceptionRenderer.php
ADmad: Changing the parent might have been missed when it moved from under `Routing/Exception/`.

othercorey: The custom error template wont render. Maybe we can figure that out.

ADmad: Yeah we could a add a check in `_template()` method to use the special template when debug is on.

othercorey: We'll have to figure that out separate because it explicitly doesn't render http exceptions in debug.

ADmad:
Apart of `InvalidPrimaryKeyException` probably being a `500`, this seems good to go.

src/Error/ExceptionRenderer.php
markstory: I can get behind changing InvalidPrimaryKeyException to a 500 as well. It is likely to be an uncommon occurrence, both in the number of effected applications and ways to trigger this exception.

othercorey: Changed InvalidPrimaryKeyException to render with 500 code since it seems to be a strong opinion here.

PR #15001, opened 2020-09-30
Code cleanup/simplification.

ADmad:

### DIFF
--- a/src/Error/ExceptionRenderer.php
+++ b/src/Error/ExceptionRenderer.php
@@ -305,7 +305,8 @@ class ExceptionRenderer implements ExceptionRendererInterface
             $baseClass = substr($baseClass, 0, -9);
         }
 
-        $method = Inflector::variable($baseClass) ?: 'error500';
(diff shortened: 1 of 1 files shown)


src/Error/ExceptionRenderer.php
```diff
@@ -339,25 +340,16 @@ protected function _message(Throwable $exception, int $code): string
      * Get template for rendering exception info.
      *
      * @param \Throwable $exception Exception instance.
-     * @param string $method Method name.
+     * @param string $template Template name.
```
othercorey, line 343: This is confusing since the function is documented as getting the template name.

othercorey: I am hoping there is a way to combine the `_method()` and `_template()` functions because they're hard to use together. The names don't really make sense either.

ADmad: I don't mind reverting the argument name to what it was before.

PR #15005, opened 2020-10-01
Update Collection parameter names

othercorey:
This follows the parameter name changes in php 8. Some of the documentation needs cleaning up, but it takes too long to do that at the same time.

~The most significant change is `$items` to `$elements` even for functions with item in the name. Maybe some of these should be `$values` following php names in places like `array_push`.~

We are using the naming convention of `$items` to refer to either values or collections passed around.

### DIFF
--- a/src/Collection/CollectionInterface.php
+++ b/src/Collection/CollectionInterface.php
@@ -28,9 +28,7 @@ use Traversable;
 interface CollectionInterface extends Iterator, JsonSerializable
 {
     /**
-     * Executes the passed callable for each of the elements in this collection
-     * and passes both the value and key for them on each step.
-     * Returns the same collection for chaining.
+     * Applies a callback to the elements in this collection.
      *
      * ### Example:
      *
@@ -40,11 +38,10 @@ interface CollectionInterface extends Iterator, JsonSerializable
      * });
      * ```
      *
-     * @param callable $c callable function that will receive each of the elements
-     * in this collection
+     * @param callable $callback Callback to run for each element in collection.
      * @return $this
      */
-    public function each(callable $c);
+    public function each(callable $callback);
 
     /**
      * Looks through each value in the collection, and returns another collection with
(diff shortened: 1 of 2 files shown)


othercorey:
I don't know why changing the parameter name to `$elements` causes this psalm failure.

```
        $list->append($this->unwrap());
        $list->append($this->newCollection($elements)->unwrap());
```

`unwrap()` returns `Traversable` while `AppendIterator::append()` wants an `Iterator`.

The failure happens on the 2nd unwrap.

src/Collection/CollectionInterface.php
```diff
@@ -645,36 +642,36 @@ public function last();
      * Returns a new collection as the result of concatenating the list of elements
      * in this collection with the passed list of elements
      *
-     * @param iterable $items Items list.
+     * @param iterable $elements Items list.
```
ADmad, line 645: Isn't the term "a collection of items" more commonly used instead of "a collection of elements"?

othercorey: Documentation refers to the entries in the collection as "elements", but I agree that it's the wrong name for the parameter.

Since they can be collections themselves, either items or values makes sense to me. We can pick one and be consistent everywhere.

ADmad: Also we have methods `appendItem()` and `prependItem()`.

othercorey: I will switch the names to "items" and we can decide how to update documentation later.

src/Collection/CollectionInterface.php
```diff
@@ -542,23 +539,23 @@ public function shuffle(): CollectionInterface;
...
      *
-     * @param int $size the maximum number of elements to take from
+     * @param int $length the maximum number of elements to take from
      * this collection
-     * @param int $from A positional offset from where to take the elements
+     * @param int $offset A positional offset from where to take the elements
      * @return self
      */
-    public function take(int $size = 1, int $from = 0): CollectionInterface;
+    public function take(int $length = 1, int $offset = 0): CollectionInterface;
```
ADmad, line 558: Wouldn't `$count` be more appropriate?

othercorey: I was following the naming style from array_splice - https://www.php.net/manual/en/function.array-splice.php.

I used $length everywhere to see if it worked.

markstory:
Looks good to me.

src/Collection/CollectionInterface.php
ADmad: I am not convinced it's the best name but NVM, being consistent with PHP's naming is good enough.

PR #15010, opened 2020-10-02
Made a more accurate comment on Email sender methods

Richard-Strittmatter:

### DIFF
--- a/src/Mailer/Message.php
+++ b/src/Mailer/Message.php
@@ -347,13 +347,14 @@ class Message implements JsonSerializable, Serializable
     }
 
     /**
-     * Sets "sender" address.
+     * Sets the "sender" address (acting "on behalf of")
      *
      * @param string|array $email String with email,
      *   Array with email as key, name as value or email as value (without name)
      * @param string|null $name Name
      * @return $this
      * @throws \InvalidArgumentException
+     * @link https://tools.ietf.org/html/rfc2822.html#section-3.6.2
      */
     public function setSender($email, ?string $name = null)
     {
@@ -361,9 +362,10 @@ class Message implements JsonSerializable, Serializable
     }
 
     /**
-     * Gets "sender" address.
+     * Gets the "sender" address (acting "on behalf of")
      *
      * @return array
+     * @link https://tools.ietf.org/html/rfc2822.html#section-3.6.2
      */
     public function getSender(): array
     {
--- a/src/Mailer/Message.php
+++ b/src/Mailer/Message.php
@@ -347,7 +347,7 @@ class Message implements JsonSerializable, Serializable
     }
 
     /**
-     * Sets the "sender" address (acting "on behalf of")
+     * Sets the "sender" address. See rfc link below for full explanation.
      *
      * @param string|array $email String with email,
      *   Array with email as key, name as value or email as value (without name)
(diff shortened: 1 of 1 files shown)


saeideng:
according to other methods , simple and short text is enough
thank you

ravage84:
@saeideng I do not agree with your reasoning, thus I reaopen. While I could agree that the wording isn't optimal, I still think this PR has its value.

Actually, @Richard-Strittmatter opened this PR on my behalf. I recently stumpled upon this method, wasn't sure if it made the correct thing I wanted to achieve and still wasn't sure if it's the correct one after reading the abysmal description. I had to look up the code, which is a testament to the poor quality of the current state of this method's description.

ADmad:
@ravage84 I would suggest keeping the description short and just link to the relevant RFC page which explains the purpose of "Sender" header for emails.

Richard-Strittmatter:
@ADmad what do you think of my amendments?

src/Mailer/Message.php
```diff
@@ -347,23 +347,25 @@ public function getFrom(): array
...
     /**
-     * Sets "sender" address.
+     * Sets the "sender" address (acting "on behalf of")
      *
      * @param string|array $email String with email,
      *   Array with email as key, name as value or email as value (without name)
      * @param string|null $name Name
      * @return $this
      * @throws \InvalidArgumentException
+     * @link https://en.wikipedia.org/wiki/Email#Header_fields
```
ADmad, line 357: I would suggest a more *authentic* source https://tools.ietf.org/html/rfc2822.html#section-3.6.2

src/Mailer/Message.php
```diff
@@ -347,23 +347,25 @@ public function getFrom(): array
     }
 
     /**
-     * Sets "sender" address.
+     * Sets the "sender" address (acting "on behalf of")
```
othercorey, line 350:
```suggestion
     * Sets the "sender" address (sender acting "on behalf of" from)
```

src/Mailer/Message.php
```diff
@@ -347,23 +347,25 @@ public function getFrom(): array
...
+     * @link https://tools.ietf.org/html/rfc2822.html#section-3.6.2
      */
     public function setSender($email, ?string $name = null)
     {
         return $this->setEmailSingle('sender', $email, $name, 'Sender requires only 1 email address.');
     }
 
     /**
-     * Gets "sender" address.
+     * Gets the "sender" address (acting "on behalf of")
```
othercorey, line 365:
```suggestion
     * Gets the "sender" address (sender acting "on behalf of" from)
```

src/Mailer/Message.php
othercorey: I don't know the cleanest way to write that.

ADmad: Which is why it's best to leave it short `Sets the "sender" address` and let users read the RFC to figure out what it does :slightly_smiling_face:

othercorey: agreed.

othercorey:
@Richard-Strittmatter Sorry for the invalid tag. The change was well-made, but without a description it was caught up in our spam checks that day. We don't mark every PR we don't merge as invalid.

src/Mailer/Message.php
ravage84: > That's why it's best to leave it short `Sets the "sender" address` and let users read the RFC to figure out what it does slightly_smiling_face

I still don't agree. Now I might not need to check the code, but get redirected into reading a hightly technical RFC.

I don't expect the documentation of a framework code to educate me about a techical term but to tell me if a given function will do what I think it does (judging by its name and documentation).

By removing the ``"on behalf of"`` you successfully removed the last additional info (aparr from the mentioned technical RFC) this PR could have provided... :man_shrugging:

PR #15011, opened 2020-10-02
Set default error code to 500 for HttpException only

othercorey:
Refs https://github.com/cakephp/cakephp/pull/14998 https://github.com/cakephp/cakephp/issues/14897

Moves the default 500 error code to HttpException only. Now that only HttpException uses the error code as http code, this makes sure we don't imply it anywhere else.

The default for Exception is now the original default of 0.

### DIFF
--- a/src/Core/Exception/Exception.php
+++ b/src/Core/Exception/Exception.php
@@ -51,7 +51,7 @@ class Exception extends RuntimeException
      *
      * @var int
      */
-    protected $_defaultCode = 500;
+    protected $_defaultCode = 0;
 
     /**
      * Constructor.
--- a/src/Error/ConsoleErrorHandler.php
+++ b/src/Error/ConsoleErrorHandler.php
@@ -61,9 +61,7 @@ class ConsoleErrorHandler extends BaseErrorHandler
     {
         $this->_displayException($exception);
         $this->logException($exception);
-        $code = $exception->getCode();
-        $code = $code && is_int($code) ? $code : 1;
-        $this->_stop($code);
+        $this->_stop(1);
     }
 
     /**
--- a/src/Error/ConsoleErrorHandler.php
+++ b/src/Error/ConsoleErrorHandler.php
@@ -16,7 +16,9 @@ declare(strict_types=1);
  */
 namespace Cake\Error;
 
+use Cake\Command\Command;
 use Cake\Console\ConsoleOutput;
+use Cake\Console\Exception\ConsoleException;
 use Throwable;
 
 /**
@@ -61,7 +63,12 @@ class ConsoleErrorHandler extends BaseErrorHandler
     {
         $this->_displayException($exception);
         $this->logException($exception);
-        $this->_stop(1);
+
+        $exitCode = Command::CODE_ERROR;
+        if ($exception instanceof ConsoleException) {
+            $exitCode = (int)$exception->getCode();
+        }
+        $this->_stop($exitCode);
     }
 
     /**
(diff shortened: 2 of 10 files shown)


tests/TestCase/ExceptionsTest.php
```diff
@@ -69,13 +69,6 @@ public function testFatalErrorException()
         $this->assertSame(__FILE__, $exception->getFile());
         $this->assertSame(1, $exception->getLine());
         $this->assertSame($previous, $exception->getPrevious());
-
-        $exception = new FatalErrorException('message', null, __FILE__, 1, $previous);
```
othercorey, line 73: This is the same test as above but checking the default error code. There is no need to do this for a random exception.

tests/TestCase/ExceptionsTest.php
```diff
@@ -93,11 +86,6 @@ public function testPersistenceFailedException()
         $this->assertSame(100, $exception->getCode());
         $this->assertSame($previous, $exception->getPrevious());
         $this->assertSame($entity, $exception->getEntity());
-
-        $exception = new PersistenceFailedException(new Entity(), 'message', null, $previous);
```
othercorey, line 97: This is the same test as above but checking the default error code. There is no need to do this for a random exception.

src/Core/Exception/Exception.php
```diff
@@ -51,7 +51,7 @@ class Exception extends RuntimeException
      *
      * @var int
      */
-    protected $_defaultCode = 500;
+    protected $_defaultCode = 0;
```
markstory, line 54: Should we use a non-zero value? `Cake\Error\ConsoleErrorHandler::handleException()` uses the exception code as an exit code which should be non-zero.

othercorey: I was considering making it `1` but started with 0 since that's the PHP default.

Does `handleException()` expect every exception error code to be > 0? There were a couple exceptions that already defaulted to 0.

othercorey: I wonder if that should really treat exceptions as exit codes. Maybe just returning 1 by default would be better even if we changed the exception default.

Different exceptions don't really mean the same exit code.

ADmad: Treating exception codes as exit codes is as wrong as treating all exception codes as http status codes :) 

So IMO we should keep exception codes as zero by default and make ConsoleErrorHandler return exit code 1.

markstory: Ok. That is a behavior change we'll need to document though. Users could be using the exception code propagation to indicate to other tools the error state of a command/shell should it fail.

othercorey: We could create a base command center exception like http exception where the error code is treated as the exit code.

markstory: What about using the existing Cake\Console\Exception\ConsoleException? We have a few subclasses of that including `StopException` which would ideally retain custom exit codes.

othercorey: Updated ConsoleErrorHandler to use error code from ConsoleException and updated unit tests to cover exit codes.

ADmad (approved):

src/Error/ConsoleErrorHandler.php
```diff
@@ -61,9 +62,12 @@ public function handleException(Throwable $exception): void
     {
         $this->_displayException($exception);
         $this->logException($exception);
-        $code = $exception->getCode();
-        $code = $code && is_int($code) ? $code : 1;
-        $this->_stop($code);
+
+        $exitCode = 1;
```
garas, line 66:
```suggestion
        $exitCode = Command::CODE_ERROR;
```

markstory (approved):
Looks good to me.

PR #15012, opened 2020-10-02
Rename Validator's setProvider() to addProvider()

johanmeiring:
A while back, when I needed to add two custom validation providers to a certain table, I found the terminology `setProvider()` confusing, as "setting" something _usually_ infers that that something has been changed once and cannot be added to, unless changed completely (e.g. `$this->Email->setViewRenderer(...)` and `$view->setTemplatePath(...)`).  To clarify, the name of the method does not lend itself to describing something actually being added to an indexed list behind the scenes, in spite of the first parameter actually being the key itself.

This PR serves as an implemented proposal to rename `setProvider()` to `addProvider()`, and marking `setProvider()` as soft-deprecated.

### DIFF
--- a/src/Validation/Validator.php
+++ b/src/Validation/Validator.php
@@ -295,10 +295,12 @@ class Validator implements ArrayAccess, IteratorAggregate, Countable
      * @param string $name The name under which the provider should be set.
      * @param object|string $object Provider object or class name.
      * @return $this
-     * @deprecated In favour of addProvider()
+     * @deprecated 4.2.0 Use `addProvider()` instead
      */
     public function setProvider(string $name, $object)
     {
+        deprecationWarning('Validator::setProvider() is deprecated; use addProvider() instead.');
+
         return $this->addProvider($name, $object);
     }
 
--- a/src/Validation/Validator.php
+++ b/src/Validation/Validator.php
@@ -295,8 +295,24 @@ class Validator implements ArrayAccess, IteratorAggregate, Countable
      * @param string $name The name under which the provider should be set.
      * @param object|string $object Provider object or class name.
      * @return $this
+     * @deprecated In favour of addProvider()
      */
     public function setProvider(string $name, $object)
+    {
+        return $this->addProvider($name, $object);
+    }
+
+    /**
+     * Associates an object to a name so it can be used as a provider. Providers are
+     * objects or class names that can contain methods used during validation of for
+     * deciding whether a validation rule can be applied. All validation methods,
+     * when called will receive the full list of providers stored in this validator.
+     *
+     * @param string $name The name under which the provider should be set.
+     * @param object|string $object Provider object or class name.
+     * @return $this
+     */
+    public function addProvider(string $name, $object)
     {
         $this->_providers[$name] = $object;
 
@@ -535,7 +551,7 @@ class Validator implements ArrayAccess, IteratorAggregate, Countable
             }
             foreach ($this->providers() as $provider) {
                 /** @psalm-suppress PossiblyNullArgument */
-                $validator->setProvider($provider, $this->getProvider($provider));
+                $validator->addProvider($provider, $this->getProvider($provider));
             }
             $errors = $validator->validate($value, $context['newRecord']);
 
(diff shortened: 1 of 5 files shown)


src/Validation/Validator.php
```diff
@@ -295,8 +295,24 @@ public function hasField(string $name): bool
      * @param string $name The name under which the provider should be set.
      * @param object|string $object Provider object or class name.
      * @return $this
+     * @deprecated In favour of addProvider()
```
othercorey, line 298:
```suggestion
     * @deprecated 4.2.0 Use `addProvider()` instead
```

othercorey:
This needs to be made in the 4.next branch.

src/Validation/Validator.php
```diff
@@ -295,8 +295,24 @@ public function hasField(string $name): bool
...
      */
     public function setProvider(string $name, $object)
+    {
+        return $this->addProvider($name, $object);
```
othercorey, line 302: Add a test to cover deprecated function.

johanmeiring:
> This needs to be made in the 4.next branch.

Would the correct approach be to rebase the work that was done off of `4.next` (instead of `master` as originally done) and then alter the PR to merge into `4.next` instead?

othercorey:
> > This needs to be made in the 4.next branch.
> 
> Would the correct approach be to rebase the work that was done off of `4.next` (instead of `master` as originally done) and then alter the PR to merge into `4.next` instead?

yes.

dereuromark:
I changed the target. Looks good still.

src/Validation/Validator.php
markstory: Any reason we wouldn't add a deprecation warning here? We could add a rector rule to make upgrading easier.

johanmeiring: No reason from my side.  I thought this might ease people into the change in a less "holy cow my code is throwing weird notices at me after upgrading" type of way.  But if you're happy with having a deprecation warning (i.e. if it's standard practice in a situation like this), I'll add it.

johanmeiring:
I've added a test case that should cover `setProvider()`, but now, awkwardly, because I added the `deprecationWarning()` statement, the test is (correctly) throwing an error.  What's the correct/preferred approach here?

ADmad:
I don't see any problem with the current method name. Whenever a `setFoo()` method has an argument like `$name` / `$key` it's apparent that one is setting a value only for that particular name/key. I don't see why one would assume that doing so would affect other names (providers in this case) too.

We also have `ObjectRegistry::set(string $objectName, object $object)`. I doubt anyone assumes that setting a particular object name will affect other objects.

Naming it `addProvider()` would actually be more confusing as someone might think they are adding another provider for that particular name, when it's fact it's *setting* a provider for that name. For e.g. when using `Mailer::addTo()` you are adding another value for "to", not setting/reseting it.

ADmad:
We have `CacheInterface::set($key, $value, $ttl)` also, surely no one thinks setting a `$key` is going to affect any other keys.

johanmeiring:
Opinions aside (for now), for future reference, what is the accepted approach for testing a method that calls the `deprecationWarning()` function?  Or should the method itself rather get a `@codeCoverageIgnore` annotation?

ADmad:
> Opinions aside (for now), for future reference, what is the accepted approach for testing a method that calls the deprecationWarning() function? 


```php
public function testFoo()
{
        // assertions here for methods that throw deprecation warning
    });
}
```

dakota:
`addProvider` makes a lot more sense to me from an English language point of view. "I'm adding a provider called `$name`". Yes, saying "I'm setting..." is still correct, it just feels clunky and unnatural.

I would even suggest the same change for `ObjectRegistry`, `CacheInterface` is a little bit different IMO because you're `setting` a value, rather than adding something to a store.

ADmad:
> I'm adding a provider called $name

But you are not always adding. If a provider for that name was already set you would be *overwriting* / *resetting* it.

johanmeiring:
Would a decent compromise here be to have the terms be synonyms without deprecating `setProvider()`?  Or maybe to have `addProvider()` fail if a provider for the given key already exists?

ADmad:
> Would a decent compromise here be to have the terms be synonyms without deprecating setProvider()? Or maybe to have addProvider() fail if a provider for the given key already exists?

That would be unnecessarily increasing the API surface for not a good enough reason. The current name is consistent and inline with how other methods are named and behave across the framework.

If you want you can provide clarification in the docblock of the method and the manual.

othercorey:
> > I'm adding a provider called $name
> 
> But you are not always adding. If a provider for that name was already set you would be _overwriting_ / _resetting_ it.

I agree that the `$name` parameter suggests we are "setting" the provider for that name.

PR #15013, opened 2020-10-02
Update Connection parameter names

othercorey:
I left any parameter that is a sql string only as `$sql` and anything that is a string or Query as `$query`. Should we make the string|Query parameters all `$sql` ?


markstory:
> Should we make the string|Query parameters all $sql ?

I think using `$query` works as `$sql` feels like it should be a string to me.

PR #15014, opened 2020-10-03
Add backwards compatibility alias for Aura\Intl\Package.

ADmad:
Refs #14888


othercorey:
Are these still compatible?

markstory:
I don't remember us making any changes that would make the aliased classes incompatible. We can't ensure compatibility long term, but this will help smooth over upgrades in the short term.

PR #15015, opened 2020-10-03
Allow Cake\ORM\Query type in the arguments of epilog()

itosho:
`epilog()` can be used as follows.

```php
$query = $this->FooTabls
    ->query()
    ->epilog($this->BarTables->find()->limit(1));
```

But, PHPStan gives the error, `Parameter #1 $expression of method Cake\Database\Query::epilog()  expects Cake\Database\Expression\QueryExpression|string|null, Cake\ORM\Query given. `.

ref: I found the problem on [this](https://github.com/itosho/easy-query/blob/master/src/Model/Behavior/InsertBehavior.php#L98)

### DIFF
--- a/src/Database/Query.php
+++ b/src/Database/Query.php
@@ -1820,7 +1820,7 @@ class Query implements ExpressionInterface, IteratorAggregate
      *
      * Epliog content is raw SQL and not suitable for use with user supplied data.
      *
-     * @param string|\Cake\Database\Expression\QueryExpression|\Cake\ORM\Query|null $expression The expression to be appended
+     * @param string|\Cake\Database\ExpressionInterface|null $expression The expression to be appended
      * @return $this
      */
     public function epilog($expression = null)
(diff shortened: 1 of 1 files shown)


src/Database/Query.php
```diff
@@ -1820,7 +1820,7 @@ public function delete(?string $table = null)
      *
      * Epliog content is raw SQL and not suitable for use with user supplied data.
      *
-     * @param string|\Cake\Database\Expression\QueryExpression|null $expression The expression to be appended
+     * @param string|\Cake\Database\Expression\QueryExpression|\Cake\ORM\Query|null $expression The expression to be appended
```
ADmad, line 1823:
```suggestion
     * @param string|\Cake\Database\ExpressionInterface|null $expression The expression to be appended
```

ndm2, line 1823:
```suggestion
     * @param string|\Cake\Database\ExpressionInterface|null $expression The expression to be appended
```

IMHO there's no need to restrict this to concrete implementations, given that `epilog()` is ment for appending arbitrary SQL.

ndm2: I see you beat me to it.

itosho: I agree.

PR #15016, opened 2020-10-03
Update Driver parameter names

othercorey:
This updates parameter names for DriverInterface and removes unnecessary abstract functions from Driver.

### DIFF
--- a/src/Database/Connection.php
+++ b/src/Database/Connection.php
@@ -615,7 +615,7 @@ class Connection implements ConnectionInterface
     /**
      * Creates a new save point for nested transactions.
      *
-     * @param string|int $name The save point name.
+     * @param string|int $name Save point name or id
      * @return void
      */
     public function createSavePoint($name): void
@@ -626,7 +626,7 @@ class Connection implements ConnectionInterface
     /**
      * Releases a save point by its name.
      *
-     * @param string|int $name The save point name.
+     * @param string|int $name Save point name or id
      * @return void
      */
     public function releaseSavePoint($name): void
@@ -637,7 +637,7 @@ class Connection implements ConnectionInterface
     /**
      * Rollback a save point by its name.
      *
-     * @param string|int $name The save point name.
+     * @param string|int $name Save point name or id
      * @return void
      */
     public function rollbackSavepoint($name): void
--- a/src/Database/Driver.php
+++ b/src/Database/Driver.php
@@ -271,36 +271,6 @@ abstract class Driver implements DriverInterface
     /**
      * @inheritDoc
      */
-    abstract public function releaseSavePointSQL($name): string;
-
-    /**
-     * @inheritDoc
-     */
-    abstract public function savePointSQL($name): string;
-
-    /**
-     * @inheritDoc
-     */
-    abstract public function rollbackSavePointSQL($name): string;
-
-    /**
-     * @inheritDoc
-     */
-    abstract public function disableForeignKeySQL(): string;
-
-    /**
-     * @inheritDoc
-     */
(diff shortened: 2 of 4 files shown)


src/Database/Connection.php
```diff
@@ -615,7 +615,7 @@ public function isSavePointsEnabled(): bool
     /**
      * Creates a new save point for nested transactions.
      *
-     * @param string|int $name The save point name.
+     * @param string|int $name Save point name or id
      * @return void
      */
     public function createSavePoint($name): void
```
othercorey, line 621: I don't know why but these take `string|int` instead of just string. It didn't seem right changing this to `$savePoint` because we pass in the name instead of generating the save point to use elsewhere.

src/Database/DriverInterface.php
```diff
@@ -97,23 +97,23 @@ public function rollbackTransaction(): bool;
     /**
      * Get the SQL for releasing a save point.
      *
-     * @param string|int $name The table name.
+     * @param string|int $name Save point name or id
      * @return string
      */
     public function releaseSavePointSQL($name): string;
```
othercorey, line 103: I don't know why but these take string|int instead of just string. It didn't seem right changing this to $savePoint because we pass in the name instead of generating the save point to use elsewhere.

src/Database/DriverInterface.php
```diff
@@ -151,10 +151,10 @@ public function supportsSavePoints(): bool;
      * Returns a value in a safe representation to be used in a query string
      *
      * @param mixed $value The value to quote.
-     * @param int $type Type to be used for determining kind of quoting to perform.
+     * @param int $type Must be one of the \PDO::PARAM_* constants
      * @return string
      */
-    public function quote($value, $type): string;
+    public function quote($value, int $type): string;
```
othercorey, line 157: This type hint was missing, but don't know if it's too late.

src/Database/Driver.php
```diff
@@ -317,17 +287,13 @@ public function supportsCTEs(): bool
...
-     * @return string
+     * @inheritDoc
      */
-    public function quote($value, $type = PDO::PARAM_STR): string
+    public function quote($value, int $type = PDO::PARAM_STR): string
     {
         $this->connect();
 
-        return $this->_connection->quote($value, $type);
+        return $this->_connection->quote((string)$value, $type);
```
othercorey, line 296: `PDO::quote()` expects a string but we accept `mixed` for $value.

src/Database/Driver.php
```diff
@@ -268,36 +268,6 @@ public function rollbackTransaction(): bool
...
-
-    /**
-     * @inheritDoc
-     */
-    abstract public function enableForeignKeySQL(): string;
-
-    /**
-     * @inheritDoc
-     */
-    abstract public function supportsDynamicConstraints(): bool;
```
markstory, line 299: Were these removed because they are also defined in the interface?

src/Database/DriverInterface.php
markstory: This is a medium risk change as there could be implementations of DriverInterface in the wild. I don't think there will be many though.

src/Database/Driver.php
othercorey: Yes, I mentioned it a while ago in chat and remembered when going through the interface.

src/Database/DriverInterface.php
othercorey: I can revert it. Honestly, the $value parameter should be `string` the same as PDO::quote() so maybe we just leave both alone until 5.

PR #15018, opened 2020-10-04
Allow the Reply-to header of e-mail contain more than one e-mail address

rturella3:
The [RFC 2822](https://www.rfc-editor.org/rfc/rfc2822.txt) says:

> an optional reply-to field MAY also be
> included, which contains the field name "Reply-To" and a
> comma-separated list of one or more addresses.
> ...
> reply-to        =       "Reply-To:" address-list CRLF

In CakePHP, if you try to set more than one reply-to, an InvalidArgumentException is thrown.

This pull request is to allow more than 1 e-mail address as reply-to.

### DIFF
--- a/src/Mailer/Message.php
+++ b/src/Mailer/Message.php
@@ -101,7 +101,7 @@ class Message implements JsonSerializable, Serializable
     protected $sender = [];
 
     /**
-     * List of email's that the recipient will reply to
+     * List of email(s) that the recipient will reply to
      *
      * @var array
      */
--- a/src/Mailer/Message.php
+++ b/src/Mailer/Message.php
@@ -395,19 +395,6 @@ class Message implements JsonSerializable, Serializable
     }
 
     /**
-     * Add "Reply-To" address.
-     *
-     * @param string|array $email Null to get, String with email,
-     *   Array with email as key, name as value or email as value (without name)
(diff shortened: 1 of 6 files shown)


src/Mailer/Message.php
```diff
@@ -394,6 +394,19 @@ public function getReplyTo(): array
...
 
+    /**
+     * Add "Reply-To" address.
+     *
+     * @param string|array $email Null to get, String with email,
+     *   Array with email as key, name as value or email as value (without name)
+     * @param string|null $name Name
+     * @return $this
+     */
+    public function addReplyTo($email, ?string $name = null)
```
ADmad, line 405: We don't add new methods in patch releases. Other changes seem fine for `master`. So you should make a separate PR targeting `4.next` to add this method after the other changes are merged.

rturella3:
@ADmad, thanks for review.

I'm new to git, can you help me in how do I proceed? Do I remove the function (and the tests that use it) and make a new commit in my fork? If I do this, will the pull request "self-update"?

ADmad:
@rturella3 Yes you can remove the function and it's test and make a new commit. Then just push the branch and the pull request will auto update.

ADmad (approved):

PR #15019, opened 2020-10-04
Update ExpressionInterface parameter names

othercorey:
This is straight-forward.

- `$generator` > `$binder` to match other changes
- `$visitor` > `$callback` to match other changes
- Updated inheritDoc comments


src/Database/Expression/QueryExpression.php
```diff
@@ -617,19 +603,19 @@ public function iterateParts(callable $visitor)
      * as they often contain user input and arrays of strings
      * are easy to sneak in.
      *
-     * @param callable|string|array|\Cake\Database\ExpressionInterface $c The callable to check.
+     * @param callable|string|array|\Cake\Database\ExpressionInterface $callable The callable to check.
      * @return bool Valid callable.
      */
-    public function isCallable($c): bool
+    public function isCallable($callable): bool
```
othercorey, line 609: This function doesn't make sense to me. It doesn't seem to be used anywhere in cakephp either. How is ExpressionInterface a valid callable?

ADmad: IIRC this method was made redundant after I made the changes for db expressions to only accept closures instead of callable. We can mark it deprecated and unused.

markstory: We were using this to filter out strings, as they are accepted in the `callable` type but aren't something we wanted to let people us inside the query builder as it opens the potential for remote code execution in a variety of places.

src/Database/Expression/QueryExpression.php
```diff
@@ -617,19 +603,19 @@ public function iterateParts(callable $visitor)
      * as they often contain user input and arrays of strings
      * are easy to sneak in.
      *
-     * @param callable|string|array|\Cake\Database\ExpressionInterface $c The callable to check.
+     * @param callable|string|array|\Cake\Database\ExpressionInterface $callable The callable to check.
      * @return bool Valid callable.
      */
```
ADmad, line 608:
```suggestion
     * @deprecated 4.2.0 This method is unused.
     * @codeCoverageIgnore
     */
```

Issue #15020, opened 2020-10-04
Type Error in php8 str_replace(): Argument #2 ($replace) must be of type array|string, int given


Badman12:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 4.1.4.
* Platform and Target: php:8.0-rc-apache image

### What you did
I try to run cakephp application on new php8-apache docker image

### What happened
I got error str_replace(): Argument #2 ($replace) must be of type array|string, int given
In  ROOT/vendor/cakephp/cakephp/src/Utility/Text.php, line 240
![image](https://user-images.githubusercontent.com/20132986/95015517-ff4aa180-0655-11eb-8f50-01602262bbdb.png)

### What you expected to happen
It was working on image php7.4 perfectly

dereuromark:
PR open.

dereuromark:
I clarified the version map in https://github.com/cakephp/cakephp/wiki#version-map

sselvanathan:
Glad to see that im not the only one, who run into this bug.

PR #15021, opened 2020-10-04
Fix TypeError  in str_replace() call

antonvolokha:
Fix issue #15020 

According to code $tmpValue must be only string in the other case it must be a empty string
so i simply check if $tmpValue is string then return $tmpValue else return empty string

In my case it fixes problem  issue #15020

### DIFF
--- a/src/Utility/Text.php
+++ b/src/Utility/Text.php
@@ -236,7 +236,7 @@ class Text
         /** @var array<string, mixed> $dataReplacements */
         $dataReplacements = array_combine($hashKeys, array_values($data));
         foreach ($dataReplacements as $tmpHash => $tmpValue) {
-            $tmpValue = is_string($tmpValue) ? $tmpValue  : '';
+            $tmpValue = is_array($tmpValue) ? '' : (string) $tmpValue;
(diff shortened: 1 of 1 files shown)


src/Utility/Text.php
```diff
@@ -236,7 +236,7 @@ public static function insert(string $str, array $data, array $options = []): st
         /** @var array<string, mixed> $dataReplacements */
         $dataReplacements = array_combine($hashKeys, array_values($data));
         foreach ($dataReplacements as $tmpHash => $tmpValue) {
-            $tmpValue = is_array($tmpValue) ? '' : $tmpValue;
+            $tmpValue = is_string($tmpValue) ? $tmpValue  : '';
```
dereuromark, line 239: I dont think that fix is correct
If by any means an integer comes out of the loop, you don't want this to be an empty string.

You should instead use casting to solve the issue.

Please also provide a test case to back up the issue.

othercorey:
This is already fixed in 4.next branch. PHP 8 support will be in version 4.2 only.

https://github.com/cakephp/cakephp/pull/14966

dereuromark:
Closing then as 4.2+ will be PHP8 ready.

PR #15026, opened 2020-10-06
Update Expression parameter names

othercorey:
These are parameters that weren't part of ExpressionInterface changes.

### DIFF
--- a/src/Database/Expression/BetweenExpression.php
+++ b/src/Database/Expression/BetweenExpression.php
@@ -34,14 +34,14 @@ class BetweenExpression implements ExpressionInterface, FieldInterface
      *
      * @var mixed
      */
-    protected $_from;
+    protected $start;
 
     /**
      * The second value in the expression
      *
      * @var mixed
      */
-    protected $_to;
+    protected $end;
 
     /**
      * The data type for the from and to arguments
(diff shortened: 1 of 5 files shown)


src/Database/Expression/BetweenExpression.php
```diff
@@ -54,20 +54,20 @@ class BetweenExpression implements ExpressionInterface, FieldInterface
...
      *
      * @param string|\Cake\Database\ExpressionInterface $field The field name to compare for values inbetween the range.
-     * @param mixed $from The initial value of the range.
-     * @param mixed $to The ending value in the comparison range.
+     * @param mixed $start The initial value of the range.
+     * @param mixed $end The ending value in the comparison range.
      * @param string|null $type The data type name to bind the values with.
      */
-    public function __construct($field, $from, $to, $type = null)
+    public function __construct($field, $start, $end, $type = null)
```
ADmad, line 61: I have never seen `start`, `end` used in SQL parlance for `BETWEEN`.

MySQL docs for e.g. uses "min", "max" https://dev.mysql.com/doc/refman/8.0/en/comparison-operators.html#operator_between

othercorey: postgres just seems to call it a "range" with "endpoints". min and max makes sense to me, however, note that `BETWEEN 1 AND 3` is different than `BETWEEN 3 and 1` and I don't believe we sort anything.

ADmad: > is different than BETWEEN 3 and 1 and I don't believe we sort anything.

Yup, so we either keep the param names unchanged or if we want to be accurate name them `$value1`, `$value2` or `$expression1`, `$expression2`.

src/Database/Expression/OrderClauseExpression.php
```diff
@@ -32,18 +32,18 @@ class OrderClauseExpression implements ExpressionInterface, FieldInterface
...
+    protected $order;
 
     /**
      * Constructor
      *
      * @param \Cake\Database\ExpressionInterface|string $field The field to order on.
-     * @param string $direction The direction to sort on.
+     * @param string $order The direction to sort on.
      */
-    public function __construct($field, $direction)
```
ADmad, line 43: I don't mind this change but I personally prefer `$direction`.

PR #15032, opened 2020-10-07
Fix operator extraction from expression keys

### LINKED ISSUE #15008: Wrong case of alias in having/where clauses
This is a (multiple allowed):

* [x] bug
* [ ] enhancement

* CakePHP Version: 4.1.4.
* Platform and Target: Mysql 5.7.14, PHP 7.2.28.

### What you did
```
CREATE TABLE test (
	id INT NOT NULL PRIMARY KEY AUTO_INCREMENT, 
	test VARCHAR(250)
) ENGINE=INNODB CHARSET=utf8mb4;
```

```
// in a controller
$this->loadModel('Test');

$test = $this->Test->find()->group(['Test.test'])->having([
	'COUNT(DISTINCT Test.id) >' => 1,
]);

debug($test);
```

### What happened

The query generated was:

```
SELECT Test.id AS Test__id, Test.test AS Test__test FROM test Test GROUP BY Test.test  HAVING COUNT(DISTINCT test.id) > :c0 
```

And the MySQL error:

```
Error: [PDOException] SQLSTATE[42S22]: Column not found: 1054 Unknown column 'test.id' in 'having clause'
```

I was expecting the query to be:

```
SELECT Test.id AS Test__id, Test.test AS Test__test FROM test Test GROUP BY Test.test  HAVING COUNT(DISTINCT Test.id) > :c0 
```

The case of the column after the `DISTINCT` has been lowered. That was fine with a MySQL database on Windows, but when I moved the database to Linux, which is case-sensitive, an error of unknown column occurred.

Adding backticks didn't worked also:

```
$test = $this->Test->find()->group(['Test.test'])->having([
	'COUNT(DISTINCT `Test.id`) >' => 1,
]);
```

Generates:

```
SELECT Test.id AS Test__id, Test.test AS Test__test FROM test Test GROUP BY Test.test  HAVING COUNT(DISTINCT `test.id`) > :c0 
```

The same happens on `where` clause function also.

I think it it related to spaces in the string of the clause, because the following code will also lowercase the alias:

```
$test = $this->Test->find()->group(['Test.test'])->where([
	'AAA(BBB( Test.id)) >' => 1,
]);
```

Output:

```
SELECT Test.id AS Test__id, Test.test AS Test__test FROM test Test WHERE AAA(BBB( test.id)) > :c0 GROUP BY Test.test 
```

Adding backticks to the where clause will also keep the alias lowercase.
But if I remove the space before the "Test.id" in the where clause, I get the correct case of the alias:

```
$test = $this->Test->find()->group(['Test.test'])->where([
	'AAA(BBB(Test.id)) >' => 1,
]);
```

Output:

```
SELECT Test.id AS Test__id, Test.test AS Test__test FROM test Test WHERE AAA(BBB(Test.id)) > :c0 GROUP BY Test.test 
```

I was able to solve using `distinct` function of the query builder and removing unnecessary spaces.

I did not expected this behavior and was caught of surprise. Lost some time debugging and changing some of my code. Sorry if it's not a bug.

markstory:
Fix the operator extraction on expression key values. Previously we were
not handling functions with spaces well which resulted in conditions
being lowercased.

Fixes #15008

### DIFF
--- a/src/Database/Expression/QueryExpression.php
+++ b/src/Database/Expression/QueryExpression.php
@@ -737,7 +737,7 @@ class QueryExpression implements ExpressionInterface, Countable
         $spaces = substr_count($field, ' ');
         if ($spaces > 1) {
             $parts = explode(' ', $field);
-            if (preg_match('/(is not|not in)$/i', $field)) {
+            if (preg_match('/(is not|not \w+)$/i', $field)) {
                 $last = array_pop($parts);
                 $second = array_pop($parts);
                 array_push($parts, strtolower("{$second} {$last}"));
--- a/src/Database/Expression/QueryExpression.php
+++ b/src/Database/Expression/QueryExpression.php
@@ -735,6 +735,7 @@ class QueryExpression implements ExpressionInterface, Countable
         $expression = $field;
 
         $spaces = substr_count($field, ' ');
+        // Handle operators with a space in them like `is not` and `not like`
         if ($spaces > 1) {
             $parts = explode(' ', $field);
             if (preg_match('/(is not|not \w+)$/i', $field)) {
--- a/src/Database/Expression/QueryExpression.php
+++ b/src/Database/Expression/QueryExpression.php
@@ -722,7 +722,7 @@ class QueryExpression implements ExpressionInterface, Countable
      * generating the placeholders and replacing the values by them, while storing
      * the value elsewhere for future binding.
      *
-     * @param string $field The value from with the actual field and operator will
+     * @param string $field The value from which the actual field and operator will
      * be extracted.
      * @param mixed $value The value to be bound to a placeholder for the field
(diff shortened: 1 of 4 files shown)


src/Database/Expression/QueryExpression.php
```diff
@@ -722,24 +722,34 @@ protected function _addConditions(array $conditions, array $types): void
...
     protected function _parseCondition(string $field, $value)
     {
+        $field = trim($field);
         $operator = '=';
         $expression = $field;
-        $parts = explode(' ', trim($field), 2);
 
-        if (count($parts) > 1) {
+        $spaces = substr_count($field, ' ');
+        if ($spaces > 1) {
```
othercorey, line 738: Can you add a comment stating what scenario this one is catching?

PR #15033, opened 2020-10-07
Fix sorting of path segments by length.

ADmad:
`rsort()` doesn't achieve what the comment `Sort path segments matching longest paths first.` says is being done. Plus we only need length based sorting here, not alphabetical.

https://3v4l.org/gBbCs


ADmad:
garas pointed out to me that fully sorting all paths based on length is not necessary. Just ensuring that longer paths like `'/debug-kit/toolbar/clear-cache'` are checked before shorter paths like `'/debug-kit/toolbar'` with a common prefix is enough.

PR #15034, opened 2020-10-08
Deprecate Exception::responseHeader() in favor of HttpException

othercorey:
~I don't know what the cakephp naming convention is for functions that take either an array of pairs or a single pair as 2 arguments:
`setHeader()` or `setHeaders()`~

We can't entirely move the functionality to HttpException until v5 because some user exceptions might not be HttpExceptions. So, we leave the original implementation until users finish converting any custom exception that set headers.

### DIFF
--- a/src/Core/Exception/Exception.php
+++ b/src/Core/Exception/Exception.php
@@ -92,12 +92,20 @@ class Exception extends RuntimeException
      *   array of "header name" => "header value"
      * @param string|null $value The header value.
      * @return array|null
+     * @deprecated 4.2.0 Use `HttpException::setHeaders()` instead. Response headers
+     *   should be set for HttpException only.
      */
     public function responseHeader($header = null, $value = null): ?array
     {
         if ($header === null) {
             return $this->_responseHeaders;
         }
+
+        deprecationWarning(
+            'Setting HTTP response headers from Exception directly is deprecated. ' .
+            'If your exceptions extend Exception, they must now extend HttpException. ' .
+            'You should only set HTTP headers on HttpException instances via the `setHeaders()` method.'
+        );
         if (is_array($header)) {
             return $this->_responseHeaders = $header;
         }
--- a/src/Http/Exception/HttpException.php
+++ b/src/Http/Exception/HttpException.php
@@ -30,4 +30,42 @@ class HttpException extends Exception
      * @inheritDoc
      */
     protected $_defaultCode = 500;
+
+    /**
+     * @var array
+     */
+    protected $headers = [];
+
+    /**
+     * Set a single HTTP response header.
+     *
+     * @param string $header Header name
+     * @param string|string[]|null $value Header value
+     * @return void
+     */
+    public function setHeader(string $header, $value = null): void
+    {
+        $this->headers[$header] = $value;
+    }
+
+    /**
+     * Sets HTTP response headers.
+     *
+     * @param array $headers Array of header name and value pairs.
+     * @return void
+     */
+    public function setHeaders(array $headers): void
+    {
+        $this->headers = $headers;
+    }
+
+    /**
+     * Returns array of response headers.
+     *
+     * @return array
+     */
+    public function getHeaders(): array
+    {
+        return $this->headers;
+    }
 }
--- a/src/Auth/BasicAuthenticate.php
+++ b/src/Auth/BasicAuthenticate.php
@@ -95,7 +95,7 @@ class BasicAuthenticate extends BaseAuthenticate
     public function unauthenticated(ServerRequest $request, Response $response)
     {
         $unauthorizedException = new UnauthorizedException();
-        $unauthorizedException->responseHeader($this->loginHeaders($request));
+        $unauthorizedException->setHeaders($this->loginHeaders($request));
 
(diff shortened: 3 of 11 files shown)


ADmad (approved):

othercorey:
Split `setHeaders()` into `setHeader()` and `setHeaders()` to avoid complicated, combined parameters.

src/Http/Exception/HttpException.php
```diff
@@ -30,4 +30,44 @@ class HttpException extends Exception
...
+    /**
+     * @var array
+     */
+    protected $headers = [];
+
+    /**
+     * Sets http response header.
+     *
+     * @param string $header Header name
+     * @param string|null $value Header value
```
ADmad, line 43: Header values can be an array.

othercorey: I forgot about that this time. Updated type hints and changed a test to pass an array of values.

src/Http/Exception/HttpException.php
```diff
@@ -30,4 +30,44 @@ class HttpException extends Exception
...
      */
     protected $_defaultCode = 500;
+
+    /**
+     * @var array
+     */
+    protected $headers = [];
+
+    /**
+     * Sets http response header.
```
markstory, line 40:
```suggestion
     * Sets HTTP response header.
```

src/Http/Exception/HttpException.php
```diff
@@ -30,4 +30,44 @@ class HttpException extends Exception
...
+     * @param string|array|null $value Header value
+     * @return void
+     */
+    public function setHeader(string $header, $value = null): void
+    {
+        $this->setHeaders([$header => $value]);
+    }
+
+    /**
+     * Sets an array of http response headers.
```
markstory, line 52:
```suggestion
     * Sets an array of HTTP response headers.
```

src/Core/Exception/Exception.php
```diff
@@ -92,6 +92,8 @@ public function getAttributes(): array
      *   array of "header name" => "header value"
      * @param string|null $value The header value.
      * @return array|null
+     * @deprecated 4.2.0 Use `HttpException::setHeaders()` instead. Response headers
+     *   should be set for HttpException only.
```
markstory, line 96: Should this emit a warning? We could automate replacing this with rector. I'd also like to avoid spending weeks adding warnings in the future.

othercorey: I wasn't sure. I can add a warning that would require users to either use setHeaders(), but if they are not using HttpException for some reason, they need to switch to it. Should we detect this and show a custom warning or just a generic warning explaining both.

I could have this code see if the exception is an HttpException and convert the calls, but then the user might have non-HttpExceptions and it would be confusing.

I need to find out if rector can tell me what the parameters are so we can replace it with the right function.

src/Http/Exception/HttpException.php
```diff
@@ -30,4 +30,44 @@ class HttpException extends Exception
...
+     *
+     * Will replace any headers with the same name.
+     *
+     * @param array $headers Array of header name and value pairs.
+     * @return void
+     */
+    public function setHeaders(array $headers): void
+    {
+        $this->headers = $headers + $this->headers;
```
othercorey, line 61: From a coding conventions POV, should a "set" function always overwrite the entire array?

src/Core/Exception/Exception.php
othercorey: Added deprecation notice when setting only.

src/Http/Exception/HttpException.php
markstory: In the case where you're taking an array of things `set` operations should overwrite existing data.

src/Core/Exception/Exception.php
```diff
@@ -92,12 +92,20 @@ public function getAttributes(): array
...
      */
     public function responseHeader($header = null, $value = null): ?array
     {
         if ($header === null) {
             return $this->_responseHeaders;
         }
+
+        deprecationWarning(
+            'Setting HTTP response headers from Exception directly is deprecated. ' .
+            'You can only set HTTP headers through HttpException::setHeaders(). ' .
```
markstory, line 106:
```suggestion
            'You should only set HTTP headers on HttpException instances via the `setHeaders()` method. ' .
```

src/Http/Exception/HttpException.php
othercorey: Updated. I updated the tests to test both methods properly. I also split up the array values so we can test that they are properly joined by ','.

src/Http/Exception/HttpException.php
```diff
@@ -30,4 +30,44 @@ class HttpException extends Exception
...
+     */
+    public function setHeader(string $header, $value = null): void
+    {
+        $this->headers[$header] = $value;
+    }
+
+    /**
+     * Sets HTTP response headers.
+     *
+     * Existing headers with the same names are overwritten.
```
othercorey, line 54:
```suggestion
     * Sets HTTP response headers.
```

othercorey, line 54:
```suggestion
     * Sets HTTP response headers.
```

PR #15036, opened 2020-10-08
Speed up file existence checks where feasible.

ADmad:
file_exists() is a lot slower than is_file() / is_dir().
Refs: https://bugs.php.net/bug.php?id=78285

`is_file()` doesn't check if the file is actually accessible but if there are file permission issues one if going to have problems one way or another anyway.

### DIFF
--- a/src/Core/Configure.php
+++ b/src/Core/Configure.php
@@ -421,7 +421,7 @@ class Configure
         }
 
         $path = dirname(dirname(__DIR__)) . DIRECTORY_SEPARATOR . 'config/config.php';
-        if (file_exists($path)) {
+        if (is_file($path)) {
             $config = require $path;
             static::write($config);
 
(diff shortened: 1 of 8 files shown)


src/Core/Configure.php
```diff
@@ -421,14 +421,10 @@ public static function version(): string
         }
 
         $path = dirname(dirname(__DIR__)) . DIRECTORY_SEPARATOR . 'config/config.php';
-        if (file_exists($path)) {
```
ADmad, line 424: I don't see why this existence check was necessary. The `config/config.php` file is always going to exist in the cakephp package.

othercorey:
Looks ok. The check in CacheEngine doesn't seem to need file_exists, but I'm not sure what that does does exactly.

src/Core/Configure.php
dereuromark: Wasnt that because of phinx or sth?

dereuromark: Came from https://github.com/cakephp/cakephp/pull/13942

ADmad: Hmm.. Right, it could be because of split packages.

ADmad: I reinstated the check and just changed the function.

PR #15037, opened 2020-10-09
4.next - Fix missing route code

markstory:
I don't think missing routes or incorrect methods should be a 500. Historically they were a 404.


src/Error/ExceptionRenderer.php
```diff
@@ -112,6 +113,7 @@ class ExceptionRenderer implements ExceptionRendererInterface
         RecordNotFoundException::class => 404,
         // Http exceptions
         MissingControllerException::class => 404,
+        MissingRouteException::class => 404,
```
othercorey, line 116:
```suggestion
        // Routing
        MissingRouteException::class => 404,
```

PR #15038, opened 2020-10-09
Rename register() for container services

othercorey:
Suggest not using a core function name for optional feature.

If php had proper function overloading, then we could just do it that way.

### DIFF
--- a/src/Http/BaseApplication.php
+++ b/src/Http/BaseApplication.php
@@ -249,9 +249,9 @@ abstract class BaseApplication implements
         if ($this->container) {
             return $this->container;
         }
-        $container = $this->register(new Container());
-        foreach ($this->plugins->with('register') as $plugin) {
-            $container = $plugin->register($container);
+        $container = $this->services(new Container());
+        foreach ($this->plugins->with('services') as $plugin) {
+            $container = $plugin->services($container);
         }
         $this->container = $container;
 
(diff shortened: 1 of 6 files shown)


ndm2:
This feels a little bit like an outlier, given the other method names in the class/interface. What about just `services`, or are there maybe plans to introduce `registerRoutes`, `registerCommands`, etc?

ADmad:
I too was thinking if we could just go with `services()`.

othercorey:
> I too was thinking if we could just go with `services()`.

That would fit.

src/Http/BaseApplication.php
```diff
@@ -249,22 +249,22 @@ public function getContainer(): ContainerInterface
         if ($this->container) {
             return $this->container;
         }
-        $container = $this->register(new Container());
+        $container = $this->services(new Container());
         foreach ($this->plugins->with('register') as $plugin) {
```
ndm2, line 253: I guess this should then be `services` too, with the `BasePlugin::$registerEnabled` property being renamed to `servicesEnabled` accordingly.

othercorey: thats true

markstory:
I'm ok with the method being `services`. I used the name `register` as that is the name league uses in ServiceProviders. However, `services` is more aligned with the other method names in Application.

ADmad (approved):

PR #15039, opened 2020-10-10
4.2 - Fix variadic routing parameters not being set during invoke

markstory:
These would not be captured during the reflection merge as there are no parameters to read during reflection.

### DIFF
--- a/tests/test_app/TestApp/Controller/DependenciesController.php
+++ b/tests/test_app/TestApp/Controller/DependenciesController.php
@@ -41,4 +41,9 @@ class DependenciesController extends Controller
     {
         return $this->response->withStringBody(json_encode(['args' => func_get_args()]));
     }
+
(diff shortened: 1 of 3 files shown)


tests/test_app/TestApp/Controller/DependenciesController.php
```diff
@@ -36,4 +36,9 @@ public function requiredDep(stdClass $dep, $any = null, ?string $str = null)
     {
         return $this->response->withStringBody(json_encode(compact('dep', 'any', 'str')));
     }
+
+    public function variadic()
```
othercorey, line 40: I assume it used to work if you had `variadic(...$params)`?

markstory: I'll add a test for that as well.

Issue #14546, opened 2020-05-05
RFC ORM: Add collate param to QueryExpression->like() & notLike()


celsowm:
This is a (multiple allowed):

* [] bug
* [x] enhancement
* [x] feature-discussion (RFC)

* CakePHP Version: 3 or major
* Platform and Target: all

### What you did
Insensitve Accent like on all "main dbs":

**mysql**: http://sqlfiddle.com/#!9/cba20d/1
**sqlserver**: http://sqlfiddle.com/#!18/cba20/3
**postgres**: http://sqlfiddle.com/#!17/cba20/8
**sqlite**: http://sqlfiddle.com/#!5/cba20d/1

### What happened

For us, non ASCII speakers, It is very tediuos to search thing like "José" or "Jose" when your database, by default is not accent insensitive

Recently, in our [portuguese reform,](https://en.wikipedia.org/wiki/Portuguese_Language_Orthographic_Agreement_of_1990) the word idéia is now ideia, without the accent.
So, since years, in our main CakePHP app our solutions are things like:
(in mssql case):

```
 $table->find()
           ->where(['field COLLATE Latin1_general_CI_AI Like' => '%idéia%']);
```

Very very ugly....

### What you expected to happen

Something like (a pun!, hehe):

```
$query = $table->find()
    ->where(function (QueryExpression $exp, Query $q) {
        return $exp->like('field', '%idéia%', 'Latin1_general_CI_AI');
    });
```

or even maybe:

```
$query = $table->find()
    ->where(function (QueryExpression $exp, Query $q) {
        return $exp->likeAi('field', '%idéia%');
    });
```

P.S. Remember, an issue is not the place to ask questions. You can use [Stack Overflow](https://stackoverflow.com/questions/tagged/cakephp)
for that or join the #cakephp channel on irc.freenode.net, where we will be more
than happy to help answer your questions.

Before you open an issue, please check if a similar issue already exists or has been closed before.

ADmad:
I am curious why don't you configure your database to use defaults that better suit your needs? Or just change the collation of the columns on which you need to do accent insensitive matches?

celsowm:
The are severe problems when you do that after all table and data be
created on Sqlserver (shame on you Bill Gates)

Em ter, 5 de mai de 2020 01:08, ADmad <notifications@github.com> escreveu:

markstory:
>         return $exp->like('field', '%idéia%', 'Latin1_general_CI_AI');

[...]

PR #15040, opened 2020-10-10
Add StringExpression with collation support

othercorey:
Refs https://github.com/cakephp/cakephp/issues/14546

`Query::newString()` helps simplify code instead of having to called `new StringExpression` everywhere.

If we want to support changing the collation of columns in an expression, that should be done through `IdentifierExpression`.

### DIFF
--- a/src/Database/Driver/Postgres.php
+++ b/src/Database/Driver/Postgres.php
@@ -18,6 +18,7 @@ namespace Cake\Database\Driver;
 
 use Cake\Database\Driver;
 use Cake\Database\Expression\FunctionExpression;
+use Cake\Database\Expression\StringExpression;
 use Cake\Database\PostgresCompiler;
 use Cake\Database\Query;
 use Cake\Database\QueryCompiler;
@@ -223,6 +224,7 @@ class Postgres extends Driver
     {
         return [
             FunctionExpression::class => '_transformFunctionExpression',
+            StringExpression::class => '_transformStringExpression',
         ];
     }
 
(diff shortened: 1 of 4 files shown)


src/Database/Driver/Postgres.php
```diff
@@ -291,6 +293,20 @@ protected function _transformFunctionExpression(FunctionExpression $expression):
...
+     * Changes string expression into postgresql format.
+     *
+     * @param \Cake\Database\Expression\StringExpression $expression The string expression to tranform.
+     * @return void
+     */
+    protected function _transformStringExpression(StringExpression $expression): void
+    {
+        $collation = $expression->getCollation();
+        if ($collation) {
+            $expression->setCollation('"' . trim($collation, '"') . '"');
```
othercorey, line 306: The `trim()` works around a design flaw where expressions are transformed every time `sql()` is called. I don't know how to fix that in 4.2.

src/Database/Query.php
```diff
@@ -1869,6 +1870,18 @@ public function newExpr($rawExpression = null): QueryExpression
...
     }
 
+    /**
+     * Returns a new instance of `StringExpression`.
+     *
+     * @param string $string String value
+     * @param string|null $collation String collation
+     * @return \Cake\Database\Expression\StringExpression
+     */
+    public function newString(string $string, ?string $collation = null)
```
ADmad, line 1880: Is this method necessary? Requiring to specify collation doesn't seem like a common enough use case to warrant a wrapper method, that too for a single statement.

othercorey: I was thinking the people who need this would use it a lot. The orm is already very verbose.

othercorey: I think StringExpression should probably make collation required otherwise you wouldn’t need the expression

PR #15043, opened 2020-10-10
Add the method addReplyTo

rturella3:
Add the method addReplyTo to Message and Mailer, to maintain the same consistency with methods like addTo, addCc and addBcc.

### DIFF
--- a/src/Mailer/Message.php
+++ b/src/Mailer/Message.php
@@ -397,6 +397,19 @@ class Message implements JsonSerializable, Serializable
     }
 
     /**
+     * Add "Reply-To" address.
+     *
+     * @param string|array $email Null to get, String with email,
+     *   Array with email as key, name as value or email as value (without name)
+     * @param string|null $name Name
+     * @return $this
+     */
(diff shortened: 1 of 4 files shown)


src/Mailer/Message.php
```diff
@@ -396,6 +396,19 @@ public function getReplyTo(): array
         return $this->replyTo;
     }
 
+    /**
+     * Add "Reply-To" address.
+     *
+     * @param string|array $email Null to get, String with email,
```
markstory, line 402:
```suggestion
     * @param string|array $email String with email,
```
I don't think we do 'null to get' in add methods anymore.

markstory (commented):
Looks good to me once the doc block is updated. Thanks!

src/Mailer/Message.php
rturella3: I copied the same comment of the 'addTo' method. You are right, this method will always return `$this` from the `addEmail` method (the same for addTo, AddBcc and addCc).
Should I change all the other 3 comments too?

PR #15044, opened 2020-10-11
Add ServiceProvider base class

markstory:
To provide consistent method name conventions for users we will need a ServiceProvider subclass. This should also help reduce the number of times that users need to import directly from league.

Refs cakephp/docs#6847

### DIFF
--- a/src/Core/ServiceProvider.php
+++ b/src/Core/ServiceProvider.php
@@ -30,8 +30,7 @@ use League\Container\ServiceProvider\BootableServiceProviderInterface;
  * @experimental This class' interface is not stable and may change
  *   in future minor releases.
  */
-abstract class ServiceProvider extends AbstractServiceProvider
-    implements BootableServiceProviderInterface
+abstract class ServiceProvider extends AbstractServiceProvider implements BootableServiceProviderInterface
 {
     /**
      * Delegate to the bootstrap() method
--- a/src/Core/ServiceProvider.php
+++ b/src/Core/ServiceProvider.php
@@ -47,7 +47,7 @@ abstract class ServiceProvider extends AbstractServiceProvider implements Bootab
         $container = parent::getContainer();
         if (!($container instanceof ContainerInterface)) {
             $message = sprintf(
-                "Unexpected container type. Expected `%s` got `%s` instead.",
+                'Unexpected container type. Expected `%s` got `%s` instead.',
                 ContainerInterface::class,
                 getTypeName($container)
             );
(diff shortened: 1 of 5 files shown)


src/Core/ServiceProvider.php
```diff
@@ -0,0 +1,87 @@
...
+ *   in future minor releases.
+ */
+abstract class ServiceProvider extends AbstractServiceProvider
+    implements BootableServiceProviderInterface
+{
+    /**
+     * Delegate to the bootstrap() method
+     *
+     * This method primarily exists as a shim between the interface
+     * that league/container has and the one we want to offer in CakePHP.
```
ADmad, line 40: Based on this should we mark this method `final` then?

markstory: Nah, if people want to footgun themselves they can.

ADmad: 😝

othercorey, line 40:
```suggestion
     * This method wraps the league/container function so users
     * only need to use the CakePHP bootstrap() interface.
```

othercorey:
Looks ok except the code checks. Can we guarantee the container type is always valid? Then we can suppress that error.

tests/test_app/TestApp/ServiceProvider/PersonServiceProvider.php
```diff
@@ -0,0 +1,22 @@
...
+declare(strict_types=1);
+
+namespace TestApp\ServiceProvider;
+
+use Cake\Core\ContainerInterface;
+use Cake\Core\ServiceProvider;
+
+class PersonServiceProvider extends ServiceProvider
+{
+    protected $provides = ['boot', 'sally'];
```
robertpustulka, line 11: I don't like this being a property in League DIC, not very intuitive. Can we have a method for this?

ADmad: Based on the docs https://container.thephpleague.com/3.x/service-providers/  it doesn't seem there's any alternative to populating this array. So we would have to add a method ourselves if needed.

robertpustulka: That's what I had in mind.

markstory: Would the method return the list of services provides? That will end up running a bit slower than a property would as the array would be re-allocated many times during a request.

PR #15047, opened 2020-10-11
ensure its a string (php8)

swiffer:


dereuromark:
This is fixed in 4.next ( https://github.com/cakephp/cakephp/blob/4.next/src/Utility/Text.php#L239 ) which will then fully support PHP 8.
That said: I don't see any issue with backporting this specific fix already to master here.

PR #15050, opened 2020-10-12
Add collation to IdentifierExpression

othercorey:
Refs https://github.com/cakephp/cakephp/issues/14546

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,7 +34,7 @@ matrix:
 before_install:
   - echo cakephp version && tail -1 VERSION.txt
 
-  - if [[ $DB == 'mysql' ]]; then mysql -u root -e 'CREATE DATABASE cakephp_test;'; fi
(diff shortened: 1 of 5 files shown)


.travis.yml
```diff
@@ -34,7 +34,7 @@ matrix:
 before_install:
   - echo cakephp version && tail -1 VERSION.txt
 
-  - if [[ $DB == 'mysql' ]]; then mysql -u root -e 'CREATE DATABASE cakephp_test;'; fi
+  - if [[ $DB == 'mysql' ]]; then mysql -u root -e 'CREATE DATABASE cakephp_test CHARACTER SET utf8mb4;'; fi
```
ADmad, line 37: This should probably be done for `master` itself as `utf8mb4` is used as default for the Mysql driver https://github.com/cakephp/cakephp/blob/1e0d18199b38f6063ca10d6f7d5aad52ec01ca83/src/Database/Driver/Mysql.php#L66.

PR #15051, opened 2020-10-12
Made a more accurate comment on Email sender methods

TerryKern:
Analogous #15010

### DIFF
--- a/src/Mailer/Email.php
+++ b/src/Mailer/Email.php
@@ -421,13 +421,14 @@ class Email implements JsonSerializable, Serializable
     }
 
     /**
-     * Sets "sender" address.
+     * Sets the "sender" address. See rfc link below for full explanation.
      *
      * @param string|array $email String with email,
      *   Array with email as key, name as value or email as value (without name)
      * @param string|null $name Name
      * @return $this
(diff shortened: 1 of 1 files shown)


ravage84 (approved):

src/Mailer/Email.php
```diff
@@ -421,23 +421,25 @@ public function from($email = null, $name = null)
...
+     * @link https://tools.ietf.org/html/rfc2822.html#section-3.6.2
      */
     public function setSender($email, $name = null)
     {
         return $this->_setEmailSingle('_sender', $email, $name, 'Sender requires only 1 email address.');
     }
 
     /**
-     * Gets "sender" address.
+     * Gets the "sender" address. See rfc link below for full explanation.
```
dereuromark, line 439: RFC, right? it is an acronym.

TerryKern: You're right. I'll make a new PR fixing where "rfc" is written and replace it with "RFC" in the entire project.

dereuromark: No need for a new PR. just directly add a commit on top of this PR.

Issue #15049, opened 2020-10-11
"Undefined index: samesite" if an app is upgraded to 3.9.3


motooka:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.9.3
* Platform and Target: 
  * Amazon Linux 1
  * Apache 2.4.46
  * PHP 7.3.21
  * PostgreSQL 9.6.x
  * behind AWS ALB

### What you did
* deploy a CakePHP 3.9.2 application
  * `debug` : false
* issue a request to the application
* upgraded the application from CakePHP 3.9.2 to CakePHP 3.9.3
* issue a request with no cookie

### What happened
The following log appears in a log file

```
2020-10-12 03:13:31 Notice: Notice (8): Undefined index: samesite in [/var/www/html/vendor/cakephp/cakephp/src/Http/Middleware/CsrfProtectionMiddleware.php, line 213]
Trace:
Cake\Error\BaseErrorHandler::handleError() - CORE/src/Error/BaseErrorHandler.php, line 168
Sentry\ErrorHandler::handleError() - ROOT/vendor/sentry/sentry/src/ErrorHandler.php, line 379
Cake\Http\Middleware\CsrfProtectionMiddleware::_addTokenCookie() - CORE/src/Http/Middleware/CsrfProtectionMiddleware.php, line 213
Cake\Http\Middleware\CsrfProtectionMiddleware::__invoke() - CORE/src/Http/Middleware/CsrfProtectionMiddleware.php, line 123
Cake\Http\Runner::__invoke() - CORE/src/Http/Runner.php, line 65
Cake\Http\Runner::run() - CORE/src/Http/Runner.php, line 51
Cake\Routing\Middleware\RoutingMiddleware::__invoke() - CORE/src/Routing/Middleware/RoutingMiddleware.php, line 170
Cake\Http\Runner::__invoke() - CORE/src/Http/Runner.php, line 65
Cake\Routing\Middleware\AssetMiddleware::__invoke() - CORE/src/Routing/Middleware/AssetMiddleware.php, line 88
Cake\Http\Runner::__invoke() - CORE/src/Http/Runner.php, line 65
App\Middleware\SentryReportMiddleware::__invoke() - APP/Middleware/SentryReportMiddleware.php, line 12
Cake\Http\Runner::__invoke() - CORE/src/Http/Runner.php, line 65
[...]

PR #15052, opened 2020-10-12
Samesite default value

motooka:
This fixes a bug, which is described in #15049

### DIFF
--- a/src/Http/Middleware/CsrfProtectionMiddleware.php
+++ b/src/Http/Middleware/CsrfProtectionMiddleware.php
@@ -207,7 +207,7 @@ class CsrfProtectionMiddleware
             '',
             (bool)$this->_config['secure'],
             (bool)$this->_config['httpOnly'],
-            $this->_config['samesite']
+            isset($this->_config['samesite']) ? $this->_config['samesite'] : $this->_defaultConfig['samesite']
         );
 
         return $response->withCookie($cookie);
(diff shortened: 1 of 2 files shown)


src/Http/Middleware/CsrfProtectionMiddleware.php
```diff
@@ -207,7 +207,7 @@ protected function _addTokenCookie($token, ServerRequest $request, Response $res
             '',
             (bool)$this->_config['secure'],
             (bool)$this->_config['httpOnly'],
-            $this->_config['samesite']
+            isset($this->_config['samesite']) ? $this->_config['samesite'] : $this->_defaultConfig['samesite']
```
dereuromark, line 210: Shoudnt the constructor take care of this?

    $this->_config = $config + $this->_defaultConfig;

What is the issue of merging that here would require an additional ternary?

motooka: Usually, the constructor works.

The problem occurs if...

- the app is upgraded from 3.9.2 or earlier, to 3.9.3 (or later)
- and route cache is not deleted after the upgrade

because the instance of the middleware is serialized in cache, and constructor is not called.

dereuromark: That sounds like a deployment issue to me, as you need to bust the cache after each deployment.
But workarounds here in the code for this seem counterproductive.

motooka: Yes, it's a deployment issue and the new code looks ugly, but I believe this change helps some application developers.

othercorey: This will be a maintenance issue in the future without context.

dereuromark: I don't believe it helps any developers, as there are still 100 other issues that can occur when not following basic principle of cache clearing when doing a deploy.
So while this might by accident then solve one, the others remain unsolved.

markstory: At least this will be one fewer problem :smile:

PR #15053, opened 2020-10-12
Added missing comment function to ConsoleIo

GrandFelix:
Added missing function to ConsoleIo which is presented in documentation https://book.cakephp.org/4/en/console-commands/input-output.html#creating-output

### DIFF
--- a/src/Console/ConsoleIo.php
+++ b/src/Console/ConsoleIo.php
@@ -225,6 +225,24 @@ class ConsoleIo
     }
 
     /**
+     * Convenience method for out() that wraps message between <comment /> tag
+     *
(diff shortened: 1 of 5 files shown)


othercorey:
Looks ok for 4.2 if we want to add it. Need to re-target merge.

ADmad:
This seems more like a case of docs being outdated. The methods were more likely updated to match the standard error levels in 4.x.

markstory:
I think having the `comment()` method is a nice addition. It is low cost to us and makes building pleasant terminal UIs a bit easier.

markstory:
I've cherry picked the relevant changes onto 4.next. Thank you @GrandFelix

PR #15055, opened 2020-10-12
Test nightly with all dbs

othercorey:
Check if php 8 fails with any db configs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,7 @@ cache:
     - $HOME/.composer/cache
 
 php:
-  - 7.4
+  - nightly
 
 env:
   matrix:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -16,7 +16,7 @@ cache:
     - $HOME/.composer/cache
 
 php:
-  - 7.4
```
ADmad, line 19: Why drop `7.4`? It's needed for the code coverage builds.

dereuromark: You probably want to rather add 3 matrix includes below.

othercorey: This is just a draft to test.

PR #15056, opened 2020-10-13
Move container construction into a protected method.

markstory:
This should make it a bit easier for applications to customize the creation of the container should they need to do that.


othercorey:
Looks good. Now that users overwrite `buildContainer()`, we can decide how `getContainer()` should work later.

PR #15057, opened 2020-10-13
Disable transactions during DDL queries

othercorey:
This resolves a dangling transaction exception in mysql pdo with php 8.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,6 +17,7 @@ cache:
 
 php:
   - 7.4
+  - nightly
 
 env:
   matrix:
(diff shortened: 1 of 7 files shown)


tests/TestCase/Datasource/PaginatorTest.php
```diff
@@ -29,8 +29,12 @@ class PaginatorTest extends TestCase
...
      */
     protected $fixtures = [
-        'core.Posts', 'core.Articles', 'core.ArticlesTags',
-        'core.Authors', 'core.AuthorsTags', 'core.Tags',
+        'core.Posts',
+        'core.Authors',
+        'core.Articles',
+        'core.Tags',
+        'core.ArticlesTags',
+        'core.AuthorsTags',
```
markstory, line 37: Did these need to be re-ordered to satisfy foreign key constraints?

othercorey: I was hacking through trying to get it to work with the existing code. I've thrown that away and just separated the queries.

PR #15061, opened 2020-10-14
Throw exception if loadFixtures() is called when auto fixtures enabled

othercorey:
In preparation for FixtureManager clean-up. This is just an extra check that shouldn't occur in existing tests.


markstory:
Seems like a good idea to me.

othercorey:
> Seems like a good idea to me.

Actually found a test that does this. That would have thrown me off later.

PR #15065, opened 2020-10-15
Fix call_user_func_array() named arguments on PHP 8

gersonfs:
On PHP 8, call_user_func_array support named arguments.

https://wiki.php.net/rfc/named_params

### DIFF
--- a/lib/Cake/Utility/ObjectCollection.php
+++ b/lib/Cake/Utility/ObjectCollection.php
@@ -126,7 +126,7 @@ abstract class ObjectCollection {
 		}
 		$result = null;
 		foreach ($list as $name) {
-			$result = call_user_func_array(array($this->_loaded[$name], $callback), array_filter(compact('subject')) + $params);
+			$result = call_user_func_array(array($this->_loaded[$name], $callback), array_values(array_filter(compact('subject')) + $params));
 			if ($options['collectReturn'] === true) {
 				$collected[] = $result;
 			}


othercorey:
Be aware that 2.x is switching to security fixes only at the end of the year. PHP 8 support isn't planned.

https://github.com/cakephp/cakephp/wiki#2x-security-mode

gersonfs:
Hello @othercorey! Yes I am aware of the deadlines.  However I have several great apps still running on cake2.  Many of them have already started migrating in parallel to cake4, however it will still take me a long time because there are no automated tests.  Being a monorepo with cake2 and cake4 I can't use different versions of php each one.  That's why I would like to add PHP 8 support to cake2 until I finish migrating to cake4.

othercorey:
Why can't you stay on php 7.4 until you finish migrating?

gersonfs:
Nothing prevents me from staying on php 7.4, but on the other hand I also thought: what prevents me from migrating to 8.0? If you think it is not feasible to support php 8, I can do it in my private repository only.

markstory:
For us to 'support' php8 we'd need to have passing CI. The core team currently doesn't have the capacity to take that work on as 2.x is only getting bugfix (and soon to only get security fixes). While this change seems reasonable, I suspect this will not be the end of changes required for PHP8, and that you'll soon be blocked on PHPunit or another dependency.

gersonfs:
Thanks for the explanations @markstory! I don't know if Travis already supports php 8, I will try to run the automated tests locally according to the CONTRIBUTING.md manual.

othercorey:
Closing since php 8 support for 2.x is not planned.

PR #15069, opened 2020-10-16
Name acronyms correctly

ravage84:
To improve the readability/clarity about common acronyms, either name them correctly, e.g. all in caps or use quotes.

### DIFF
--- a/src/Controller/Component/RequestHandlerComponent.php
+++ b/src/Controller/Component/RequestHandlerComponent.php
@@ -380,7 +380,7 @@ class RequestHandlerComponent extends Component
      *
      * ### Usage:
      *
-     * Render the response as an 'AJAX' response.
+     * Render the response as an 'ajax' response.
      *
      * ```
      * $this->RequestHandler->renderAs($this, 'ajax');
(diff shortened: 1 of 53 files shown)


dereuromark:
Those are not abbreviations, but acronyms :)
But your point is still valid.

src/Controller/Component/RequestHandlerComponent.php
```diff
@@ -380,13 +380,13 @@ public function prefers($type = null)
      *
      * ### Usage:
      *
-     * Render the response as an 'ajax' response.
+     * Render the response as an 'AJAX' response.
```
garas, line 383: Unquote or leave as it was before.

src/View/XmlView.php
```diff
@@ -61,7 +61,7 @@
 class XmlView extends SerializedView
 {
     /**
-     * XML layouts are located in the xml sub directory of `Layouts/`
+     * XML layouts are located in the 'xml' sub directory of `Layouts/`
```
garas, line 64: Use backticks `` `xml` `` as in `Layouts` or add slash `xml/`

garas: And `Layouts/` are `layout/` in 4.x.

othercorey: Just use `layouts/xml/`

PR #15071, opened 2020-10-17
Only include URL encoded data in Oauth signatures

### LINKED ISSUE #15066: Oauth 1.0 authentication / Wrong signature
* [x] bug

* CakePHP Version: 4.1.5
* Platform and Target: Any.

### What you did
I'm trying to make a post request with Oauth 1.0 authentication with an xml body. like this :

```php
$http = new Client();
$response = $http->post('http://example.com/test', '<?xml version="1.0" encoding="utf-8"?><test></test>', [
  'headers' => ['Content-Type' => 'text/xml'],
  'auth' => [
    'type' => 'oauth',
    'consumerKey' => 'xxx',
    'consumerSecret' => 'xxx',
    'token' => 'xxx',
    'tokenSecret' => 'xxx'
  ]
]);
```

### What happened
Everything works fine except that the signature generated is wrong. Cakephp uses the request body to generate it when it shouldn't.

**Section 9.1.1 of Oauth spec (https://oauth.net/core/1.0a/#anchor13) :**

Capture

I'm expecting for cakephp dont use the body of the request to generate the Oauth Authorization when it shouldn't do it :)

markstory:
The OAuth 1.0a spec only allows urlencoded data to be included in the OAuth signature to be included. All other content types should be skipped.

Fixes #15066

### DIFF
--- a/src/Http/Client/Auth/Oauth.php
+++ b/src/Http/Client/Auth/Oauth.php
@@ -285,9 +285,11 @@ class Oauth
         parse_str((string)$query, $queryArgs);
 
         $post = [];
-        $body = (string)$request->getBody();
-        parse_str($body, $post);
-
+        $contentType = $request->getHeaderLine('Content-Type');
+        if ($contentType === '' || $contentType === 'application/x-www-form-urlencoded') {
+            $body = (string)$request->getBody();
+            parse_str($body, $post);
+        }
         $args = array_merge($queryArgs, $oauthValues, $post);
         $pairs = $this->_normalizeData($args);
         $data = [];
--- a/src/Http/Client/Auth/Oauth.php
+++ b/src/Http/Client/Auth/Oauth.php
@@ -287,8 +287,7 @@ class Oauth
         $post = [];
         $contentType = $request->getHeaderLine('Content-Type');
         if ($contentType === '' || $contentType === 'application/x-www-form-urlencoded') {
-            $body = (string)$request->getBody();
-            parse_str($body, $post);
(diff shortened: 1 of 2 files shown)


tests/TestCase/Http/Client/Auth/OauthTest.php
```diff
@@ -273,6 +273,56 @@ public function testBaseStringWithPostData()
...
+     */
+    public function testBaseStringWithXmlPostData()
+    {
+        $request = new Request(
+            'http://example.com/search?q=pogo',
+            Request::METHOD_POST,
+            [
+                'Content-Type' => 'application/xml',
+            ],
+            '<xml>stuff</xml>',
```
othercorey, line 295:
```suggestion
            '<xml>stuff</xml>'
```

src/Http/Client/Auth/Oauth.php
```diff
@@ -285,9 +285,11 @@ protected function _normalizedParams(Request $request, array $oauthValues): stri
         parse_str((string)$query, $queryArgs);
 
         $post = [];
-        $body = (string)$request->getBody();
-        parse_str($body, $post);
-
+        $contentType = $request->getHeaderLine('Content-Type');
+        if ($contentType === '' || $contentType === 'application/x-www-form-urlencoded') {
+            $body = (string)$request->getBody();
+            parse_str($body, $post);
```
othercorey, line 291:
```suggestion
            parse_str((string)$request->getBody(), $post);
```

PR #15074, opened 2020-10-18
Make sure debug as string doesn't throw error.

dereuromark:
The default app template (especially of existing apps) uses

    'debug' => filter_var(env('DEBUG', false), FILTER_VALIDATE_BOOLEAN),

Some might also still use

     'debug' => env('DEBUG', false),

or alike.

So it can also come in from env using strings, e.g. empty string or `'0'` for false or `'1'` for true.

A stringish debug value here creates a fatal error though down the line:

> [TypeError] Argument 2 passed to Cake\Error\ErrorHandler::_displayError() must be of the type boolean, string given

in  `protected function _displayError(array $error, bool $debug): void` part.

I propose casting here where the code depends on the bool value for logic/comparison.

### DIFF
--- a/src/Error/BaseErrorHandler.php
+++ b/src/Error/BaseErrorHandler.php
@@ -166,7 +166,7 @@ abstract class BaseErrorHandler
             'line' => $line,
         ];
 
-        $debug = Configure::read('debug');
+        $debug = (bool)Configure::read('debug');
         if ($debug) {
             // By default trim 3 frames off for the public and protected methods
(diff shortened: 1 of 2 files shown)


ADmad:
I would prefer the devs cast the values properly when setting.

dereuromark:
Me too, but since there could always be one app_local not doing so, the risk of a complete fatal error seems a bit high.
The cast itself seems like a good compromise here.

Issue #15075, opened 2020-10-18
SqliteSchemaDilect fails on derived columns from SQLite Views


cnizzardini:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: `4.1.5`
* Platform and Target: Ubuntu 20.04 LTS, PHP 7.2.33, SQLite 3.32.3, running `bin/cake server`

### What you did
Created a SQLite View that has a derived column using `count()`

```sql
CREATE VIEW view_film_categories AS
	SELECT
	    categories.id,
	    categories.name,
	    CAST(count(film_categories.uuid) as INT) as total
	FROM
	    categories
	LEFT JOIN
	    film_categories on film_categories.category_id = categories.id
	GROUP BY
		categories.name
```

Created a class extending `Cake\ORM\Table`

```php
<?php
declare(strict_types=1);

namespace App\Model\Table;

use Cake\ORM\Table;

class ViewFilmCategoriesTable extends Table
{
    /**
     * Initialize method
     *
     * @param array $config The configuration for the Table.
     * @return void
     */
    public function initialize(array $config): void
    {
        parent::initialize($config);
        $this->setTable('view_film_categories');
        $this->setPrimaryKey('id');
    }
}
```

Then called `Cake\Database\Schema\CachedCollection::describe` which ultimately calls `Cake\Database\Schema\SqliteSchemaDialect::_convertColumn`. See stack trace below. 

### What happened

> Unable to parse column type from ""
> Cake\Database\Exception
> CORE/src/Database/Schema/SqliteSchemaDialect.php:57
> Cake\Database\Schema\SqliteSchemaDialect->_convertColumn
> CORE/src/Database/Schema/SqliteSchemaDialect.php:172
> Cake\Database\Schema\SqliteSchemaDialect->convertColumnDescription
> CORE/src/Database/Schema/Collection.php:146
> Cake\Database\Schema\Collection->_reflect
> CORE/src/Database/Schema/Collection.php:100
> Cake\Database\Schema\Collection->describe
> CORE/src/Database/Schema/CachedCollection.php:84

### What you expected to happen

To be able to use views. It seems the SQLite is unable to determine the column affinity (https://www.sqlite.org/datatype3.html) for derived columns. In this case, it would be nice if CakePHP at least defaulted to text or some other string value to avoid the exception?

[...]

PR #15076, opened 2020-10-18
Default to string if no SQLite type is defined

cnizzardini:
See #15075. This PR defaults column type to `string` if no type is defined by SQLite.

### DIFF
--- a/src/Database/Schema/SqliteSchemaDialect.php
+++ b/src/Database/Schema/SqliteSchemaDialect.php
@@ -52,6 +52,10 @@ class SqliteSchemaDialect extends SchemaDialect
      */
     protected function _convertColumn(string $column): array
     {
+        if (empty($column)) {
+            return ['type' => TableSchema::TYPE_STRING, 'length' => null];
+        }
+
         preg_match('/(unsigned)?\s*([a-z]+)(?:\(([0-9,]+)\))?/i', $column, $matches);
         if (empty($matches)) {
             throw new Exception(sprintf('Unable to parse column type from "%s"', $column));
--- a/src/Database/Schema/SqliteSchemaDialect.php
+++ b/src/Database/Schema/SqliteSchemaDialect.php
@@ -52,8 +52,8 @@ class SqliteSchemaDialect extends SchemaDialect
      */
     protected function _convertColumn(string $column): array
     {
-        if (empty($column)) {
-            return ['type' => TableSchema::TYPE_STRING, 'length' => null];
+        if ($column === '') {
+            return ['type' => TableSchema::TYPE_TEXT, 'length' => null];
         }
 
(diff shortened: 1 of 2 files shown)


src/Database/Schema/SqliteSchemaDialect.php
```diff
@@ -52,6 +52,10 @@ class SqliteSchemaDialect extends SchemaDialect
      */
     protected function _convertColumn(string $column): array
     {
+        if (empty($column)) {
+            return ['type' => TableSchema::TYPE_STRING, 'length' => null];
```
ADmad, line 56: The fallback type at the end of this function is `return ['type' => TableSchema::TYPE_TEXT, 'length' => null];` so maybe we should use the same for consistency?

src/Database/Schema/SqliteSchemaDialect.php
```diff
@@ -52,6 +52,10 @@ class SqliteSchemaDialect extends SchemaDialect
      */
     protected function _convertColumn(string $column): array
     {
+        if (empty($column)) {
```
ADmad, line 55:
```suggestion
        if ($column === '') {
```

othercorey:
Do we have a view test for the other dbs?

cnizzardini:
> Do we have a view test for the other dbs?

@othercorey not sure, i went with the view test since that is how I discovered the problem. I could do a more narrow unit test instead (if this test is even needed).

tests/TestCase/Database/Schema/SqliteSchemaTest.php
```diff
@@ -384,6 +392,35 @@ public function testDescribeTable()
...
+                'null' => true,
+                'default' => null,
+                'precision' => null,
+                'comment' => null,
+                'collate' => null,
+            ],
+        ];
+        $this->assertInstanceOf('Cake\Database\Schema\TableSchema', $result);
+        foreach ($expected as $field => $definition) {
+            $this->assertEquals($definition, $result->getColumn($field));
```
dereuromark, line 420: This should be assertSame()

tests/TestCase/Database/Schema/SqliteSchemaTest.php
```diff
@@ -384,6 +392,35 @@ public function testDescribeTable()
         }
     }
 
+    /**
+     * Tests SQLite views
+     *
+     * @return void
+     */
+    public function testDescribeView()
```
markstory, line 400:
```suggestion
    public function testDescribeViewWithComputedValues()
```

I think this captures a bit more of the scenario this test is covering.

cnizzardini: My idea was to leave that test open for other assertions that may be needed with views. I am okay altering the name as you suggested if you don't think that's pertinent.

markstory: I'm not too fussy about the names. :smile:

PR #15081, opened 2020-10-20
Change _runOperation to runPerFixture

othercorey:
Port some of the refactor cleanup from debugging php 8 transaction issue. This is to improve code clarity for future work.

- The `_runOperation()` wrapper is renamed to `runPerFixture()` and runs the callback per-fixture in the same transaction.
- The `_fixtureConnections()` helper is renamed to `groupFixturesByConnection()`.

There only runtime change is in `FixtureManager::load()` where the `dropContraints()`, `create()/drop()/truncate()` and `createConstraints()` calls are run in separate transactions.


markstory (approved):
Looks good to me. I like how you've moved the loop into the `runPerFixture` method. I didn't see any significant logic changes.

PR #15088, opened 2020-10-21
Fix parameter types

martinstuecklschwaiger:
These functions expect arrays and only get passed arrays.

### DIFF
--- a/lib/Cake/TestSuite/CakeTestLoader.php
+++ b/lib/Cake/TestSuite/CakeTestLoader.php
@@ -31,7 +31,7 @@ class CakeTestLoader extends PHPUnit_Runner_StandardTestSuiteLoader {
  * Load a file and find the first test case / suite in that file.
  *
  * @param string $filePath The file path to load
- * @param string $params Additional parameters
+ * @param array $params Additional parameters
  * @return ReflectionClass
  */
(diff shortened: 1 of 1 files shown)


lib/Cake/TestSuite/CakeTestLoader.php
```diff
@@ -31,7 +31,7 @@ class CakeTestLoader extends PHPUnit_Runner_StandardTestSuiteLoader {
  * Load a file and find the first test case / suite in that file.
  *
  * @param string $filePath The file path to load
- * @param string $params Additional parameters
+ * @param array $params Additional parameters
  * @return ReflectionClass
  */
 	public function load($filePath, $params = '') {
```
dereuromark, line 37: But then the default value wouldnt make sense. Maybe instead of changing the docs you want to add a cast somewhere.

othercorey: Looks like these type hints and `_basePath` hints are wrong. Looks like both should be `array|string`.

othercorey:
@martinstuecklschwaiger Can you fix the hints for `_basePath`?

markstory:
Closing as 2.10 has reached end of support and will only receive security fixes going forward.

Issue #14127, opened 2020-01-04
4.x - http/client


saeideng:
without having base uri you have to use like 
 ```php
new Client([
                'host' => "example.com",
                'scheme' => "https",
            ]);
```
after adding 
```php
new Client([
                'base_uri' => "https://example.com",
            ]);
```

`base_uri` could be change

saeideng:
even we can add `uri` to accept `Uri` objects
maybe something like this 
```php
new Client([
                'uri' => new Uri("https://example.com"),
            ]);
```

markstory:
Would you want to support path and query strings in the baseUrl as well? We could use `parse_url()` to parse the provided string into the current array data.

ADmad:
I am not in favour of providing multiple ways to set uri/host/protocol. It makes the code and usage unnecessarily difficult to follow.

saeideng:
>Would you want to support path and query strings in the baseUrl as well?

probably no

>I am not in favour of providing multiple ways to set uri/host/protocol

ok, then we can have a method to do this 
```php
new Client()->setUrl(....);
new Client()->setBaseUrl(....);
```

saeideng:
A PR has been merged :  https://github.com/cakephp/cakephp/pull/15090
closing....

PR #15090, opened 2020-10-24
Create scoped client

T0mats0:
This is a pull request for issue #14127 . I read the discussion. I agree with Admad about not muddling up the constructor with mixed params. I also didn't like the idea of scoping the client after construction.  So I came up with the idea to make a static factory method that does the parsing for the string and returns a scoped client. I can see how this would be useful for folks storing string endpoints in their configuration or database. I thought about creating a factory class, but went with a method instead.

Let me know what you think.


src/Http/Client.php
```diff
@@ -227,6 +227,30 @@ public function addCookie(CookieInterface $cookie)
...
     }
 
+    /**
+     * Pass a string url to set the base domain and scheme for the request.
+     * Url must parse into a scheme and host.
+     *
+     * @param  string  $url A string url e.g. https://example.com
+     * @return self
+     */
+    public static function createScopedClientFromUrl(string $url): self
```
dereuromark, line 237: The documented type is `static`, and there should be no return typehint on these methods.

tests/TestCase/Http/ClientTest.php
```diff
@@ -898,4 +898,61 @@ public function testRedirectDifferentSubDomains()
...
+        $this->assertEquals('example.co', $client->getConfig('host'));
+    }
+
+    /**
+     * Port is set when passed to client in string
+     */
+    public function testCreateScopedClientSetsPort()
+    {
+        $client = Client::createScopedClientFromUrl('https://example.co:8765/');
+        $this->assertEquals('8765', $client->getConfig('port'));
```
dereuromark, line 926: We prefer assertSame() here to make sure the type is also as expected.

tests/TestCase/Http/ClientTest.php
```diff
@@ -898,4 +898,61 @@ public function testRedirectDifferentSubDomains()
...
         $this->assertSame($response, $result);
     }
+
+    /**
+     * Scheme is set when passed to client in string
+     */
+    public function testCreateScopedClientSetsProtocol()
+    {
+        $client = Client::createScopedClientFromUrl('https://example.co/');
+        $this->assertEquals('https', $client->getConfig('scheme'));
```
dereuromark, line 908: There are a few more of those.

T0mats0: :+1:

src/Http/Client.php
```diff
@@ -227,6 +227,31 @@ public function addCookie(CookieInterface $cookie)
         return $this;
     }
 
+    /**
+     * Pass a string url to set the base domain and scheme for the request.
+     * Url must parse into a scheme and host.
+     *
+     * @param  string  $url A string url e.g. https://example.com
```
dereuromark, line 234: There are some extra white space in there that shouldn't be.
Also, in text URL is usually used as acronym.

ADmad:
There should also be a test showing that only scheme, port and host are used. Any additional URL parts are ignored. The docblock could clarify the same too.

ADmad:
I wonder if we could add support for a `basePath` config too. So using URL `https://foo.com/api/` would result in `basePath` being `api/` and calling `get('bar')` would make a request to `https://foo.com/api/bar`.

T0mats0:
> I wonder if we could add support for a `basePath` config too. So using URL `https://foo.com/api/` would result in `basePath` being `api/` and calling `->get('bar')` would make a request to `https://foo.com/api/bar`.

I like this idea it would make the the proposed function more useful if I'm using the class to interact with an api.

src/Http/Client.php
```diff
@@ -227,6 +227,31 @@ public function addCookie(CookieInterface $cookie)
...
+     * @param  string $url A string url e.g. https://example.com
+     * @return static
+     * @throws \InvalidArgumentException
+     */
+    public static function createScopedClientFromUrl(string $url)
+    {
+        $parts = parse_url($url);
+
+        if ($parts === false) {
+            $parts = [];
```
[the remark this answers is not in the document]
T0mats0: Made this an exception, and I found a string that produces a boolean false response (see test). It's difficult to make the function return false ('', 'somestring, '123',  '_' all return an array).

src/Http/Client.php
```diff
@@ -227,6 +227,31 @@ public function addCookie(CookieInterface $cookie)
...
+        }
+
+        $config = array_intersect_key($parts, ['scheme' => '', 'port' => '', 'host' => '']);
+
+        if (!isset($config['scheme']) || !isset($config['host'])) {
+            throw new InvalidArgumentException('String ' . $url . ' did not parse');
```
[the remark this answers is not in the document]
T0mats0: Updated the error message.

tests/TestCase/Http/ClientTest.php
```diff
@@ -898,4 +898,96 @@ public function testRedirectDifferentSubDomains()
...
+
+    /**
+     * Test exception is thrown if passed URL has no domain
+     */
+    public function testCreateScopedClientThrowsInvalidArgumentExceptionWhenNoDomainProvided()
+    {
+        try {
+            Client::createScopedClientFromUrl('https://');
+        } catch (InvalidArgumentException $e) {
+            $this->assertInstanceOf(InvalidArgumentException::class, $e);
```
markstory, line 968: You can use `expectException()` instead of try/catch.

tests/TestCase/Http/ClientTest.php
```diff
@@ -898,4 +898,96 @@ public function testRedirectDifferentSubDomains()
...
+    }
+
+    /**
+     * Test exception is throw when no scheme is provided.
+     */
+    public function testCreateScopedClientThrowsInvalidArgumentExceptionWhenNoSchemeProvided()
+    {
+        try {
+            Client::createScopedClientFromUrl('example.co');
+        } catch (InvalidArgumentException $e) {
```
markstory, line 952: You can use `expectException()` instead of try/catch.

tests/TestCase/Http/ClientTest.php
```diff
@@ -898,4 +898,96 @@ public function testRedirectDifferentSubDomains()
...
+    }
+
+    /**
+     * Test exception is thrown when URL cannot be parsed
+     */
+    public function testCreateScopedClientThrowsInvalidExceptionWhenUrlCannotBeParsed()
+    {
+        try {
+            Client::createScopedClientFromUrl('htps://');
+        } catch (InvalidArgumentException $e) {
```
markstory, line 927: You can use `expectException()` and `expectExceptionMessage()` instead of this try/catch.

ADmad:
> I like this idea it would make the the proposed function more useful if I'm using the class to interact with an api.

Yup, accessing APIs is probably the most common use case for the HTTP client and often APIs have a common path prefix. So adding this feature would avoid have to repeat the common path fragment for each request call.

Let's see what others have to say about it.

src/Http/Client.php
```diff
@@ -227,6 +227,32 @@ public function addCookie(CookieInterface $cookie)
...
+    /**
+     * Client instance returned is scoped to the domain, port, and scheme parsed from the passed URL string. The passed
+     * string must have a scheme and a domain. Optionally, if a port is included in the string, the port will be scoped
+     * too. Other parts of the url string are ignored.
+     *
+     * @param  string $url A string URL e.g. https://example.com
+     * @return static
+     * @throws \InvalidArgumentException
+     */
+    public static function createScopedClientFromUrl(string $url)
```
ADmad, line 239: I don't think using "scoped" in the name is necessary and the "client" is also redundant. The name could be shortened to `createFromUrl()`.

tests/TestCase/Http/ClientTest.php
T0mats0: Corrected.

tests/TestCase/Http/ClientTest.php
T0mats0: Corrected

tests/TestCase/Http/ClientTest.php
T0mats0: Corrected

T0mats0:
I've updated the name of the method to createFromUrl. I've adjusted the tests to use expectsException().

Do we want to add 'basePath' (uri segment), to the method as proposed by Admad? If so I would propose adding it to the defaults and options that get passed around the rest of the class. 

I like the idea, it would make the class more useful for calling api's where there is a base path or uri segment after the domain.

markstory:
I think you'll need to merge 4.next back into this branch or rebase to get CI passing.

ADmad:
@T0mats0 I think we can go ahead with the addition of "basePath" option. No one has objected to it's addition:).

T0mats0:
Okay. I'll merge 4.x. My last push has basePath added.

tests/TestCase/Http/ClientTest.php
```diff
@@ -902,82 +902,69 @@ public function testRedirectDifferentSubDomains()
...
-            Client::createScopedClientFromUrl('https://');
-        } catch (InvalidArgumentException $e) {
-            $this->assertInstanceOf(InvalidArgumentException::class, $e);
-
-            return;
-        }
-        $this->fail('Client should have thrown exception');
+        $this->expectException(InvalidArgumentException::class);
+        Client::createFromUrl('https://');
+        $message = $this->getExpectedExceptionMessage();
```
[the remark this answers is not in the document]
T0mats0: It was not, but I've updated the code. It should be good now.

T0mats0:
I fetched and merged 4.next. Some database errors are still occurring. I'm not sure why.

T0mats0:
... and we're green again except for the static analyzer does not like "new static".

basePath option was added to the client class.

othercorey:
> ... and we're green again except for the static analyzer does not like "new static".

You need to add it to the baseline file so it is ignored. Here is an example https://github.com/cakephp/cakephp/blob/de98d012fbd258c77b774b891ddd3c6e37020e49/phpstan-baseline.neon#L119

I would also like to see this function move to just below the constructor.

tests/TestCase/Http/ClientTest.php
```diff
@@ -898,4 +920,98 @@ public function testRedirectDifferentSubDomains()
...
+    public function testCreateFromUrlSetsHost()
+    {
+        $client = Client::createFromUrl('https://example.co/');
+        $this->assertSame('example.co', $client->getConfig('host'));
+    }
+
+    /**
+     * basePath is set when passed to client in string
+     */
+    public function testCreateFromUrlSetsBasePath()
```
ADmad, line 945: Instead of individual test methods like these you can add a provider method for URLs and then a single test methods which asserts the scheme, host, port and basePath generated.

PR #15092, opened 2020-10-25
Run composer install before composer stan-setup

othercorey:
Composer V2 requires a lock file to perform a partial update.


dereuromark:
They sad the bug is fixed. So this might not be necessary anymore

othercorey:
Fixed in composer 2.0.2

PR #15097, opened 2020-10-28
Set max php version for split packages

othercorey:
Copy the changes from https://github.com/cakephp/cakephp/commit/a7bedf10bb41ed68d9e0281ee26945ab3960fd3f to split packages.

### DIFF
--- a/src/Cache/composer.json
+++ b/src/Cache/composer.json
@@ -22,7 +22,7 @@
         "source": "https://github.com/cakephp/cache"
     },
     "require": {
-        "php": ">=5.6.0",
+        "php": ">=5.6.0,<8.0.0",
         "psr/simple-cache": "^1.0.0",
         "cakephp/core": "^3.6.0"
     },
--- a/src/Collection/composer.json
+++ b/src/Collection/composer.json
@@ -23,7 +23,7 @@
         "source": "https://github.com/cakephp/collection"
     },
     "require": {
-        "php": ">=5.6.0"
+        "php": ">=5.6.0,<8.0.0"
     },
     "autoload": {
         "psr-4": {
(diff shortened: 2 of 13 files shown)


dereuromark:
Interesting
We don't try to allow PHP8 for Cake3?

Isn't the compatibility doable in a 3.next? And preferrable to forcing the devs running an outdated PHP 7 at some point?

othercorey:
It requires an unknown amount of work and people are trying to use it thinking it's just a single variable patch. If someone decides to backport everything to 3.10 and we release that, then we can update these.

markstory:
> Isn't the compatibility doable in a 3.next? And preferrable to forcing the devs running an outdated PHP 7 at some point?

It is potentially doable, but getting 3.next to PHP8 requires bumping PHPUnit and that requires breaking changes in CakePHP or runtime hacks.

dereuromark:
Well, the source could might still be very well compatible with it. So restricting that then could be cumbersome for many.
The testing harness is sure a different problem, though.

othercorey:
What is the concern here?

3.10 is not in development and we don't know what breaks with php 8. We don't have any way to test patches.

Are you saying it should not be restricted in case someone works on a patch in a year? We can't tell users it will be done at the moment.

dereuromark:
I would suggest breaking API now in a branch just to see how PHP8 is running with Cake3.
And then make an informed decision based on that - rather than just closing it down due to the test harness issue
As 99% the core works fine with PHP8 :)

othercorey:
Feel free.

othercorey:
Also, if you decide to test and enable support, need to revert the commit mentioned in the summary. This PR just fixes it.

ADmad:
Just stumped upon this: https://github.com/Yoast/PHPUnit-Polyfills

othercorey:
We should merge this or revert the original commit - to be consistent.

PR #15100, opened 2020-10-29
Fix regression in object value comparison when marshalling entity.

### LINKED ISSUE #15099: 4.x regression on patchEntity update query for datetime columns
This is a (multiple allowed):

* [x] bug
* [ ] enhancement

* CakePHP Version: 4.1.5
* Platform and Target: apache, php 7.3 ; .

### What you did
When I update data with patchEntity, datetime data are always updated even if the datetime is not modified.

### What happened
When using patchEntity, CakePHP 4 generate an sql query to update all datetime columns, even if the columns is not modified.    

cakePHP convert datetime columns to FrozenTime objects.

This is a regression come with commit  : https://github.com/cakephp/cakephp/commit/9bd82d7ac13642f03cde23c43e73c5a5271694b9

 src/ORM/Marshaller.php (line 562 563)
```php
 -  (is_object($value) && !($value instanceof EntityInterface) && $original == $value)
+  (is_object($value) && !($value instanceof EntityInterface) && $original === $value)
```

in cakePHP 3, $original == $value, check that FrozenTime objects are set dirty only if object data is different
in cakePHP 4, $original === $value,  FrozenTime objects are always set dirty even if FrozenTime objects data have the same date.

restore the cakePHP 3 behavior so update query are correct.

ADmad:
Closes #15099

### DIFF
--- a/tests/TestCase/ORM/MarshallerTest.php
+++ b/tests/TestCase/ORM/MarshallerTest.php
@@ -1375,6 +1375,32 @@ class MarshallerTest extends TestCase
     }
 
     /**
+     * Test merge() doesn't dirty objects which are equal.
+     *
+     * @return void
+     */
+    public function testMergeWithSameObjectValue()
+    {
+        $created = new FrozenTime('2020-10-29');
+        $entity = new Entity([
+            'comment' => 'foo',
+            'created' => $created,
+        ]);
+        $entity->setAccess('*', true);
+        $entity->setNew(false);
+        $entity->clean();
+
+        $data = [
+            'comment' => 'bar',
+            'created' => clone $created,
+        ];
+        $marshall = new Marshaller($this->comments);
+        $marshall->merge($entity, $data);
+
+        $this->assertFalse($entity->isDirty('created'));
+    }
+
+    /**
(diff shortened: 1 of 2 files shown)


othercorey:
This allows scenarios like setting the entity with Date when we marshal using FrozenDate?

tests/TestCase/ORM/MarshallerTest.php
```diff
@@ -1374,6 +1374,32 @@ public function testMergeUnchangedNullValue()
...
+            'comment' => 'foo',
+            'created' => $created,
+        ]);
+        $entity->setAccess('*', true);
+        $entity->setNew(false);
+        $entity->clean();
+
+        $data = [
+            'comment' => 'bar',
+            'created' => clone $created,
```
othercorey, line 1395: Should we create a different type here?

markstory: Why? I think this captures the regression well enough.

othercorey: I missed that cloning alone would fail the `===` test. This should work.

PR #15102, opened 2020-10-30
Update PHP 8 build job on travis

ADmad:
`8.0snapshot` has all the additional extensions we need except `memcached` but it's using an older PHP 8 snapshot than `nightly`. I have left a comment on Travis' forum so waiting for it to get sorted out.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,7 @@ cache:
 
 php:
   - 7.4
-  - nightly
+  - '8.0snapshot'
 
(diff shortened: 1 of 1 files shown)


othercorey:
The fix we need is in RC3. They should be using that at least.

othercorey:
I think we're going to drop this, right?

ADmad:
We need to shift off Travis now so is pointless.

PR #15104, opened 2020-10-30
Add FixtureManager test for unmanaged fixtures

othercorey:
This confirms the current setup will only delete then insert records.

### DIFF
--- a/tests/TestCase/TestSuite/FixtureManagerTest.php
+++ b/tests/TestCase/TestSuite/FixtureManagerTest.php
@@ -281,6 +281,63 @@ class FixtureManagerTest extends TestCase
     }
 
     /**
+     * Tests handling unmnaged fixtures which don't
+     * import a schema or defined fields. These tables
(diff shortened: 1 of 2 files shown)


tests/TestCase/TestSuite/FixtureManagerTest.php
```diff
@@ -280,6 +280,63 @@ public function testFixturizeInvalidType()
...
+        $conn->newQuery()
+            ->insert(['title', 'body'])
+            ->into('unmanaged')
+            ->values([
+                'title' => 'Existing title',
+                'body' => 'Existing body',
+            ])
+            ->execute();
+
+        $buffer = new ConsoleOutput();
```
markstory, line 310: We have an ArrayLog that could be better here if you need to collect logs.

othercorey: I copied that mess from the test. Both should be fixed.

PR #15105, opened 2020-10-30
Update RedirectException to extend HttpException.

ADmad:

### DIFF
--- a/src/Http/Exception/RedirectException.php
+++ b/src/Http/Exception/RedirectException.php
@@ -57,6 +57,8 @@ class RedirectException extends HttpException
      */
     public function addHeaders(array $headers)
     {
+        deprecationWarning('RedirectException::addHeaders() is deprecated, use setHeaders() instead.');
+
         foreach ($headers as $key => $value) {
             $this->headers[$key][] = $value;
         }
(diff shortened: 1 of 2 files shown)


src/Http/Exception/RedirectException.php
```diff
@@ -31,15 +29,8 @@
  * Additional headers can also be provided in the constructor, or
  * using the addHeaders() method.
```
othercorey, line 30: Not sure if we should update or just remove this and let people use the HttpException api naturally.

othercorey:
We need to deprecate the add/remove headers functions.

src/Http/Exception/RedirectException.php
ADmad: We could deprecated `addHeaders()`. I am not quite sure how common it is to add new headers after the exception instance has been constructed.

othercorey:
Seems ok. I don't know if setting the http code is necessarily since redirects are handled separately.

src/Http/Exception/RedirectException.php
```diff
@@ -74,21 +69,12 @@ public function addHeaders(array $headers)
      *
      * @param string $key The header to remove.
      * @return $this
+     * @deprecated 4.2.0 Use `setHeaders()` instead.
```
markstory, line 72: These methods should emit runtime warnings so we don't have to go back and add them later.

othercorey (approved):

PR #15109, opened 2020-10-31
Added package name to each base exception class

othercorey:
Add package name to the base exception for all packages. This aligns all packages and reduces the need to alias the exception names in code to avoid collisions with the name `Exception.

We can leave the alias for Core exception through 5 if needed, but it helps our internal code and new code.

Moved `DatabaseException` into Exception namespace to match other packages.

### DIFF
--- a/src/Core/Exception/CakeException.php
+++ b/src/Core/Exception/CakeException.php
@@ -115,5 +115,5 @@ class CakeException extends RuntimeException
 }
 
 // phpcs:disable
-class_exists('Cake\Core\Exception\Exception');
+class_exists('Cake\Core\Exception\CakeException');
 // phpcs:enable
--- a/src/Core/Exception/CakeException.php
+++ b/src/Core/Exception/CakeException.php
@@ -115,5 +115,5 @@ class CakeException extends RuntimeException
 }
 
 // phpcs:disable
-class_alias('Cake\Core\Exception\CakeException', 'Cake\Core\Exception\Exception');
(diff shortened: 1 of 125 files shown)


src/Core/Exception/Exception.php
```diff
@@ -9,107 +9,8 @@
  * Redistributions of files must retain the above copyright notice.
  *
  * @copyright     Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
- * @since         3.0.0
+ * @since         4.2.0
```
ADmad, line 12: Shouldn't this remain 3.0.0 and instead the new class have 4.2.0?

othercorey: Up to you. I thought maybe we would say the alias was since 4.2 but the actual exception code since the original version?

I'll switch if you're sure.

othercorey:
psalm crash:
https://github.com/vimeo/psalm/issues/4462

src/Core/Exception/CakeException.php
```diff
@@ -0,0 +1,119 @@
...
+
+use RuntimeException;
+use Throwable;
+
+/**
+ * Base class that all CakePHP Exceptions extend.
+ *
+ * @method int getCode() Gets the Exception code.
+ */
+class CakeException extends RuntimeException
```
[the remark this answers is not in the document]
othercorey: I was letting the tests pass first with the aliases. If this is approved I'll update all internal uses. I've tested the non-alias version in DatabaseException.

PR #15112, opened 2020-10-31
Allowing scalar and empty body for JSON requests

jrbasso:
When the request is set as `application/json` and contains a valid body, such as empty body or non-array/object, the `BodyParserMiddleware` was throwing a `BadRequestException`.

This change will put the content in an indexed array to conform with the PSR interface. It will keep giving `BadRequestException` for invalid JSON content.

### DIFF
--- a/src/Http/Middleware/BodyParserMiddleware.php
+++ b/src/Http/Middleware/BodyParserMiddleware.php
@@ -178,11 +178,19 @@ class BodyParserMiddleware implements MiddlewareInterface
      * Decode JSON into an array.
      *
      * @param string $body The request body to decode
-     * @return mixed
+     * @return array|null
      */
     protected function decodeJson(string $body)
     {
-        return json_decode($body, true);
+        if ($body === '') {
+            return [];
+        }
+        $decoded = json_decode($body, true);
+        if (json_last_error() === JSON_ERROR_NONE) {
+            return (array)$decoded;
+        }
+
+        return null;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


src/Http/Middleware/BodyParserMiddleware.php
```diff
@@ -182,7 +182,15 @@ public function process(ServerRequestInterface $request, RequestHandlerInterface
...
-        return json_decode($body, true);
+        if ($body === '') {
+            return [];
+        }
+        $decoded = json_decode($body, true);
+        if (json_last_error() === JSON_ERROR_NONE) {
+            return (array)$decoded;
+        }
+
+        return false;
```
othercorey, line 193: Shouldn't this return null?

dereuromark: Ideally yes, but I guess for BC reasons it should return similar values as pure json_decode.

othercorey: > NULL is returned if the json cannot be decoded

https://www.php.net/manual/en/function.json-decode.php

I thought it does.

othercorey:
I disabled the dockerhub login. If you merge it should run.

src/Http/Middleware/BodyParserMiddleware.php
jrbasso: With the current code any non-array response will be considered a bad response. To keep more compatible with `json_encode` I updated it to `null`.

othercorey: Should we update the return type hint to `array|null` ?

jrbasso: Done too.

PR #15115, opened 2020-11-01
Add TestFixture::isManaged()

othercorey:
This allows porting more of FixtureManager cleanup. 

This adds a way to flag a fixture as unmanaged where the fixture instance should not be create/destroyed, only truncated. Previously, we had to hack the drop()/create() methods to work around being called when they shouldn't be.

### DIFF
--- a/src/TestSuite/Fixture/TestFixture.php
+++ b/src/TestSuite/Fixture/TestFixture.php
@@ -274,7 +274,7 @@ class TestFixture implements ConstraintsInterface, FixtureInterface, TableSchema
             return false;
         }
 
-        if (empty($this->import) && empty($this->fields)) {
+        if (!$this->isManaged()) {
             return true;
         }
 
@@ -309,8 +309,8 @@ class TestFixture implements ConstraintsInterface, FixtureInterface, TableSchema
             return false;
         }
 
-        if (empty($this->import) && empty($this->fields)) {
-            return $this->truncate($connection);
+        if (!$this->isManaged()) {
+            return true;
         }
 
         try {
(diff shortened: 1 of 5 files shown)


othercorey:
I don't know if there's any point to adding this to `FixtureInterface`. Not sure what the use case is outside of TestFixture.

markstory:
> Not sure what the use case is outside of TestFixture (does anyone create their own fixtures?)

We have another fixture implementation in the elasticsearch plugin, and I would assume that other plugins with datasource like features may implement fixtures.

othercorey:
I added a method annotation to `FixtureInterface` to allow other fixture implementations to wrap existing fixtures/objects without needing to handle bad logic.

I don't know what the best name for this method is. I've tried:

- `isManaged()`
- `isInstanceManaged()`
- `allowCreate()`
- `shouldCreate()`

src/TestSuite/Fixture/TestFixture.php
```diff
@@ -439,6 +439,16 @@ public function truncate(ConnectionInterface $connection): bool
         return true;
     }
 
+    /**
+     * Returns whether the fixture should be created/destroyed.
+     *
+     * @return bool
+     */
+    public function allowCreate(): bool
```
markstory, line 447: I don't know about this method name. Would something like `hasManagedSchema()` work?

markstory: Derp, just saw your comment. Of the options you had `isManaged()` fits better. I'm not a fan of `allowCreate()` as fixtures can create records & schema and the method name isn't clear to me.

othercorey: I didn't like it either, but wanted feedback before switching back.

Issue #15117, opened 2020-11-02
Add a tag to right justify text in TableHelper


imo-tikuwa:
This is a (multiple allowed):

* [ ] bug
* [x] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 4.1.5
* Platform and Target: Any

### What i did
I have added a text-right tag to [TableHelper](https://github.com/cakephp/cakephp/blob/master/src/Shell/Helper/TableHelper.php) that allows right-justified output of columns.

I checked the operation with the test code in the following branch.
https://github.com/imo-tikuwa/cakephp/commit/95e783868c68d625ba2928d427dff9262a2d4458

### What you expected to happen
If there is no problem with this fix, I would like to send a pull request to merge.

othercorey:
It is easier to discuss if you open the PR.

imo-tikuwa:
I tried to create a PR.

PR #15118, opened 2020-11-02
add text-right tag to TableHelper.

imo-tikuwa:
#15117

### DIFF
--- a/src/Shell/Helper/TableHelper.php
+++ b/src/Shell/Helper/TableHelper.php
@@ -114,7 +114,7 @@ class TableHelper extends Helper
             if (!empty($options['style'])) {
                 $column = $this->_addStyle($column, $options['style']);
             }
-            if (!is_null($column) && preg_match('#<text-right>.*</text-right>#', $column)) {
+            if ($column !== null && preg_match('#<text-right>.+</text-right>#', $column)) {
                 $out .= '| ' . str_repeat(' ', $pad) . $column . ' ';
             } else {
                 $out .= '| ' . $column . str_repeat(' ', $pad) . ' ';
--- a/src/Shell/Helper/TableHelper.php
+++ b/src/Shell/Helper/TableHelper.php
@@ -16,6 +16,7 @@ declare(strict_types=1);
 namespace Cake\Shell\Helper;
 
 use Cake\Console\Helper;
+use UnexpectedValueException;
 
 /**
  * Create a visually pleasing ASCII art table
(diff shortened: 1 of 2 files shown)


dereuromark:
I think you should point a new feature such as this to 4.next branch (as master is feature frozen, only bugfixes).

src/Shell/Helper/TableHelper.php
```diff
@@ -114,7 +114,11 @@ protected function _render(array $row, array $widths, array $options = []): void
             if (!empty($options['style'])) {
                 $column = $this->_addStyle($column, $options['style']);
             }
-            $out .= '| ' . $column . str_repeat(' ', $pad) . ' ';
+            if (!is_null($column) && preg_match('#<text-right>.*</text-right>#', $column)) {
```
dereuromark, line 117:
```suggestion
            if ($column !== null && preg_match('#<text-right>.+</text-right>#', $column)) {
```

markstory: What will happen when someone does `<text-right>some</text-right>text`?

dereuromark:
> What will happen when someone does <text-right>some</text-right>text?

Yeah, I think we should have a test case for this.
It could through an exception maybe.

src/Shell/Helper/TableHelper.php
imo-tikuwa: Thank you for pointing out the code.
I took in the points that you pointed out.

When I tried `<text-right>some</ text-right>text`, it was right-justified, including the text outside the tag.
```
$outputs = [
    ['Item', 'Price per piece (yen)'],
    ['Apple', '<text-right>some</text-right>text'],
    ['Orange', '100'],
];
$io->helper('Table')->output($outputs);
```
image

markstory: I assumed that would be the output. I don't think it is expected based on the markup text though. The CLI markup language borrows conventions from HTML which would generate results like `text      some` instead. We could raise an exception if there is text outside the alignment tags as it is likely an error.

It would also be good to ensure formatting like `<text-right><comment>blue</comment>normal</text-right>` behaves correctly as well.

imo-tikuwa:
@markstory 
@dereuromark 
I modified it to throw an exception when there is a string outside the text-right tag.

> It would also be good to ensure formatting like `<text-right><comment>blue</comment>normal</text-right>` behaves correctly as well.

I tried to correspond to tags such as info included in the text-right tag, the info tag was output as a character string.
`<text-right>` and `</ text-right>` are used to determine if `str_repeat(' ', $pad)` in the TableHeler class should be done before or after the string.
Therefore, I thought that it would be okay to delete it from the character string at the timing before passing it to the output processing of the console, so I fixed it.

PR #15119, opened 2020-11-02
Test php 8 on appveyor

othercorey:


othercorey:
we are removing appveyor instead

PR #15120, opened 2020-11-03
Allowing apps to change asset URL building

### LINKED ISSUE #15108: RFC: Replace asset timestamp by asset hash on URLs
This is a (multiple allowed):

* [ ] bug
* [x] enhancement

* CakePHP Version: 5.0+
* Platform and Target: N/A

### The problem
Currently we can set the `Asset.timestamp` option or pass an option on the `HtmlHelper`/`UrlHelper` methods, so the URL will contain the file timestamp on it as query parameter. It helps to generate a new cached version/request for the file. It is great when the assets are static on the repository, however with the increase of asset bundle builder and compilers for assets (webpack, sass, rollup, grunt tasks, etc) it's making the timestamp to vary on every build.

It is not a problem in terms of functionality since the assets are going to work the same way, but when it's deployed, the browser and CDNs will have to download the assets again, increasing network throughput, slowing down the page load, etc.

### Proposal solution
Replace the asset timestamp functionality by an asset hash. It will have the same end goal (vary the cache when file is modified), but will not vary the hash when the content of the file is update. PHP offers the methods `md5_file` and `sha1_file` that can be used for that and they are pretty quick.

### Migration plan
For CakePHP 5.0:
* Replace the asset timestamp functionality by asset hash (likely `sha1_file`).
* Create the configuration called `Asset.hash` with the same values from `Asset.timestamp`
* The configuration `Asset.timestamp` will be marked as deprecated and will be an "alias" for `Asset.hash`
* Deprecate option `timestamp` from UrlHelper, making an alias to `assetHash`
* Add the option `assetHash` on the methods currently supporting `timestamp` on `UrlHelper`

### Considerations
An alternative would be supporting both options (timestamp and hash), but both will accomplish the same goal, so I don't think is necessary to create more global configurations and helper method options. It will make things more confuse when one option solves both problems.

jrbasso:
It will allow the apps to implement/extend their own version of asset URL builder.

The app will be able to manipulate the asset generated URL. Some applications:
* Customize the versioning, changing from the core supported timestamp to something else (ie, md5 hash)
* Allow to change the asset filename based on file generated by asset builder, such as webpack
* Define different CDN hostname based on some option

Closes #15108

### DIFF
--- a/src/View/Helper/UrlHelper.php
+++ b/src/View/Helper/UrlHelper.php
@@ -16,6 +16,8 @@ declare(strict_types=1);
  */
 namespace Cake\View\Helper;
 
+use Cake\Core\App;
+use Cake\Core\Exception\CakeException;
 use Cake\Routing\Asset;
 use Cake\Routing\Router;
 use Cake\View\Helper;
@@ -25,6 +27,43 @@ use Cake\View\Helper;
  */
 class UrlHelper extends Helper
 {
+    /**
+     * Default config for this class
+     *
+     * @var array
+     */
+    protected $_defaultConfig = [
+        'assetUrlClassName' => Asset::class,
+    ];
+
+    /**
+     * Asset URL engine class name
+     *
+     * @var string
+     * @psalm-var class-string<\Cake\Routing\Asset>
+     */
+    protected $_assetUrlClassName;
+
+    /**
+     * Check proper configuration
+     *
+     * @param array $config The configuration settings provided to this helper.
+     * @return void
+     */
+    public function initialize(array $config): void
+    {
+        parent::initialize($config);
+        $engineClassConfig = $this->getConfig('assetUrlClassName');
+
+        /** @psalm-var class-string<\Cake\Routing\Asset>|null $engineClass */
+        $engineClass = App::className($engineClassConfig, 'Routing');
+        if ($engineClass === null) {
+            throw new CakeException(sprintf('Class for %s could not be found', $engineClassConfig));
+        }
+
+        $this->_assetUrlClassName = $engineClass;
+    }
+
     /**
      * Returns a URL based on provided parameters.
      *
@@ -99,7 +138,7 @@ class UrlHelper extends Helper
     {
         $options += ['theme' => $this->_View->getTheme()];
 
-        return h(Asset::imageUrl($path, $options));
+        return h($this->_assetUrlClassName::imageUrl($path, $options));
     }
 
     /**
@@ -124,7 +163,7 @@ class UrlHelper extends Helper
     {
         $options += ['theme' => $this->_View->getTheme()];
 
-        return h(Asset::cssUrl($path, $options));
+        return h($this->_assetUrlClassName::cssUrl($path, $options));
     }
 
     /**
@@ -149,7 +188,7 @@ class UrlHelper extends Helper
     {
         $options += ['theme' => $this->_View->getTheme()];
 
-        return h(Asset::scriptUrl($path, $options));
+        return h($this->_assetUrlClassName::scriptUrl($path, $options));
     }
 
     /**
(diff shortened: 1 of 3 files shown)


src/View/Helper/UrlHelper.php
```diff
@@ -98,8 +98,10 @@ public function buildFromPath(string $path, array $params = [], array $options =
     public function image(string $path, array $options = []): string
     {
         $options += ['theme' => $this->_View->getTheme()];
+        /** @var class-string $className */
+        $className = App::className('Asset', 'Routing');
```
ADmad, line 102: The asset classname should be a helper config similar to https://github.com/cakephp/cakephp/blob/41903eb1751b7c7c4dc45bbe9ceed7514cd52135/src/View/Helper/TextHelper.php#L49-L52

jrbasso: Updated.

src/View/Helper/UrlHelper.php
```diff
@@ -16,15 +16,62 @@
...
  */
 class UrlHelper extends Helper
 {
+    /**
+     * Default config for this class
+     *
+     * @var array
+     */
+    protected $_defaultConfig = [
+        'engine' => Asset::class,
```
ADmad, line 37: `engine` key sounds weird in this case. Something like `assetUrlClassName` would be more appropriate.

src/View/Helper/UrlHelper.php
```diff
@@ -16,15 +16,62 @@
...
+     *
+     * - `engine` Class name to use to replace asset url building functionality.
+     *            The class needs to be placed in the `Routing` directory.
+     *
+     * @param \Cake\View\View $view the view object the helper is attached to.
+     * @param array $config Settings array Settings array
+     * @throws \Cake\Core\Exception\CakeException when the engine class could not be found.
+     */
+    public function __construct(View $view, array $config = [])
+    {
```
ADmad, line 61: Instead of overriding the constructor use the initialize() hook. The overriding in TextHelper predates the addition of initialize() hook for helpers.

src/View/Helper/UrlHelper.php
jrbasso: I thought the same. I had it as `assetEngine`, but since most of the helper is based on `Asset` class I left it as `engine` to keep more consistent with the other helpers.
I can still change if prefer to keep as `assetUrlClassName`

jrbasso: Addressed it.

src/View/Helper/UrlHelper.php
jrbasso: Updated.

ADmad (approved):

othercorey:
Do we want to merge this?

src/View/Helper/UrlHelper.php
```diff
@@ -25,6 +27,43 @@
...
+        parent::initialize($config);
+        $engineClassConfig = $this->getConfig('assetUrlClassName');
+
+        /** @psalm-var class-string<\Cake\Routing\Asset>|null $engineClass */
+        $engineClass = App::className($engineClassConfig, 'Routing');
+        if ($engineClass === null) {
+            throw new CakeException(sprintf('Class for %s could not be found', $engineClassConfig));
+        }
+
+        $this->_assetUrlClassName = $engineClass;
```
markstory, line 64: One problem I can see with having this being a config option and a separate property, is that after the helper is built someone could do `setConfig('assetUrlClassName', 'Something')` and it would take no effect. I don't think that will come up often, but it will be a 'bug' when it does.

ADmad: Yes, that can be an issue. The same issues exits for `TimeHelper` and `NumberHelper` too. We would have to override `setConfig()` to check the config keys and update the property too if require if we really want to address it.

markstory: Well those haven't been a problem this far, so let's go with this.

PR #15121, opened 2020-11-03
Add PHP 8 and coverage builds to Github Actions

ADmad:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -110,9 +110,15 @@ jobs:
 
         if [[ ${{ matrix.php-version }} == '7.2' ]]; then
           vendor/bin/phpunit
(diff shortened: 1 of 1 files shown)


ADmad:
Don't know why pgsql got salty on PHP 7.4 :disappointed:.

othercorey:
Looks ok.

I've been dropping a few of the inconsistent but repeating failures in issue to look at later.

othercorey:
My only thought is about the php to db version match. Originally, I was testing the extremes of both (7.2 + mysql 8 and 7.4 + mysql 5.6). I can see that we want to test latest db with the most builds though.

ADmad:
Feel free to switch back / change the combinations as you see fit.

PR #15122, opened 2020-11-04
Use PHP env variable in ServerCommand

### LINKED ISSUE #15116: bin/cake server ignores PHP environment variable
This is a (multiple allowed):

* [x] bug
* [x] enhancement

* CakePHP Version: 4.1.5
* Platform and Target: linux x64

### What you did
I wanted to try out a different PHP version than the one installed system-wide, compiled it and ran

    $ PHP=/path/to/my/own/cli/php bin/cake server

### What happened
The bin/cake script was running with the PHP version I compiled locally, but the server it starts with `php -S` runs with the system-wide installed PHP version.

I expected the ServerCommand.php script to also check for the PHP environment variable like the bin/cake script and use the PHP path provided in it if available, else fall back to the system-wide PHP installation

The relevant code is in [ServerCommand.php:130](https://github.com/cakephp/cakephp/blob/master/src/Command/ServerCommand.php#L130)

markstory:
Use the PHP environment variable when starting up the PHP server.

Fixes #15116

### DIFF
--- a/src/Command/ServerCommand.php
+++ b/src/Command/ServerCommand.php
@@ -126,8 +126,10 @@ class ServerCommand extends Command
     public function execute(Arguments $args, ConsoleIo $io): ?int
     {
         $this->startup($args, $io);
+        $phpBinary = env('PHP', 'php');
         $command = sprintf(
-            'php -S %s:%d -t %s',
+            '%s -S %s:%d -t %s',
+            $phpBinary,
             $this->_host,
             $this->_port,
             escapeshellarg($this->_documentRoot)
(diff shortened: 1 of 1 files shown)


src/Command/ServerCommand.php
```diff
@@ -126,8 +126,10 @@ protected function startup(Arguments $args, ConsoleIo $io): void
     public function execute(Arguments $args, ConsoleIo $io): ?int
     {
         $this->startup($args, $io);
+        $phpBinary = env('PHP', 'php');
```
othercorey, line 129:
```suggestion
        $phpBinary = (string)env('PHP', 'php');
```

PR #15123, opened 2020-11-04
Add a method to get nodes from Debugger

markstory:
I don't see this method being used frequently, but I'm planning on using it DebugKit to resolve cakephp/debug_kit#777. This should also help simplify the Variable panel code as we can remove the various code paths that try to handle the unserializable classes. Instead of storing the actual variable data, the panel could store the debug nodes.

### DIFF
--- a/src/Error/Debugger.php
+++ b/src/Error/Debugger.php
@@ -627,6 +627,23 @@ class Debugger
     }
 
     /**
+     * Convert the variable to the internal node tree.
+     *
+     * The node tree can be manipulated and serialized more easily
+     * than many object graphs can.
+     *
+     * @param mixed $var Variable to convert.
+     * @param int $maxDepth The depth to generate nodes to. Defaults to 3.
+     * @return NodeInterface The root node of the tree.
+     */
+    public static function exportNodes($var, int $maxDepth = 3): NodeInterface
+    {
+        $context = new DebugContext($maxDepth);
+
+        return static::export($var, $context);
+    }
+
+    /**
      * Protected export function used to keep track of indentation and recursion.
      *
      * @param mixed $var The variable to dump.
--- a/src/Error/Debugger.php
+++ b/src/Error/Debugger.php
@@ -634,7 +634,7 @@ class Debugger
      *
      * @param mixed $var Variable to convert.
      * @param int $maxDepth The depth to generate nodes to. Defaults to 3.
-     * @return NodeInterface The root node of the tree.
+     * @return \Cake\Error\Debug\NodeInterface The root node of the tree.
      */
     public static function exportNodes($var, int $maxDepth = 3): NodeInterface
     {
--- a/src/Error/Debugger.php
+++ b/src/Error/Debugger.php
@@ -636,7 +636,7 @@ class Debugger
      * @param int $maxDepth The depth to generate nodes to. Defaults to 3.
      * @return \Cake\Error\Debug\NodeInterface The root node of the tree.
      */
-    public static function exportNodes($var, int $maxDepth = 3): NodeInterface
+    public static function exportVarAsNodes($var, int $maxDepth = 3): NodeInterface
     {
         return static::export($var, new DebugContext($maxDepth));
     }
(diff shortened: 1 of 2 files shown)


src/Error/Debugger.php
```diff
@@ -626,6 +626,23 @@ public static function exportVar($var, int $maxDepth = 3): string
...
+     *
+     * @param mixed $var Variable to convert.
+     * @param int $maxDepth The depth to generate nodes to. Defaults to 3.
+     * @return NodeInterface The root node of the tree.
+     */
+    public static function exportNodes($var, int $maxDepth = 3): NodeInterface
+    {
+        $context = new DebugContext($maxDepth);
+
+        return static::export($var, $context);
```
othercorey, line 643:
```suggestion
        return static::export($var, new DebugContext($maxDepth));
```

src/Error/Debugger.php
```diff
@@ -626,6 +626,23 @@ public static function exportVar($var, int $maxDepth = 3): string
...
 
+    /**
+     * Convert the variable to the internal node tree.
+     *
+     * The node tree can be manipulated and serialized more easily
+     * than many object graphs can.
+     *
+     * @param mixed $var Variable to convert.
+     * @param int $maxDepth The depth to generate nodes to. Defaults to 3.
+     * @return NodeInterface The root node of the tree.
```
othercorey, line 637:
```suggestion
     * @return \Cake\Error\Debug\NodeInterface The root node of the tree.
```

tests/TestCase/Error/DebuggerTest.php
```diff
@@ -501,6 +504,41 @@ public function testExportVarSplFixedArray()
...
+    {
+        $data = [
+            1 => 'Index one',
+            5 => 'Index five',
+        ];
+        $result = Debugger::exportNodes($data);
+        $this->assertInstanceOf(NodeInterface::class, $result);
+        $this->assertCount(2, $result->getChildren());
+
+        /** @var \Cake\Error\Debug\ArrayItemNode */
```
othercorey, line 522:
```suggestion
        /** @var \Cake\Error\Debug\ArrayItemNode $item */
```

src/Error/Debugger.php
```diff
@@ -626,6 +626,23 @@ public static function exportVar($var, int $maxDepth = 3): string
...
+     */
+    public static function exportNodes($var, int $maxDepth = 3): NodeInterface
```
othercorey, line 639: This shouldn't be a big debate, but this name feels a little odd for what it does. Why don't we just expose `export` and have users pass in a context?

markstory: I agree the name is a bit clunky. Would `exportToNodes` be better? I was concerned with requiring callers to have to construct the context and be exposed to more of the conversion process guts.

othercorey: To keep it clear between `exportVar()` and this, I would say `exportVarAsNodes()`. exportNodes sounds like it takes nodes as input and exports debug string like exportVar does.

src/Error/Debugger.php
```diff
@@ -626,6 +626,21 @@ public static function exportVar($var, int $maxDepth = 3): string
...
+     * Convert the variable to the internal node tree.
+     *
+     * The node tree can be manipulated and serialized more easily
+     * than many object graphs can.
+     *
+     * @param mixed $var Variable to convert.
+     * @param int $maxDepth The depth to generate nodes to. Defaults to 3.
+     * @return \Cake\Error\Debug\NodeInterface The root node of the tree.
+     */
+    public static function exportNodes($var, int $maxDepth = 3): NodeInterface
```
othercorey, line 639:
```suggestion
    public static function exportVarAsNodes($var, int $maxDepth = 3): NodeInterface
```

tests/TestCase/Error/DebuggerTest.php
```diff
@@ -501,6 +504,41 @@ public function testExportVarSplFixedArray()
...
+     *
+     * @return void
+     */
+    public function testExportNodes()
+    {
+        $data = [
+            1 => 'Index one',
+            5 => 'Index five',
+        ];
+        $result = Debugger::exportNodes($data);
```
othercorey, line 518:
```suggestion
        $result = Debugger::exportVarAsNodes($data);
```

PR #15127, opened 2020-11-05
3.x - Include missing option in the docblock

davidyell:
I noticed that this is listed in the documentation but is missing from the docblock.

I found the options here https://book.cakephp.org/3/en/orm/associations.html#belongsto-associations

Thought it might be helpful to add it here in-case someone is viewing the source. I have copied in the text from the book link.

### DIFF
--- a/src/ORM/Table.php
+++ b/src/ORM/Table.php
@@ -1135,6 +1135,9 @@ class Table implements RepositoryInterface, EventListenerInterface, EventDispatc
      * - finder: The finder method to use when loading records from this association.
      *   Defaults to 'all'. When the strategy is 'join', only the fields, containments,
      *   and where conditions will be used from the finder.
+     * - propertyName: The property name that should be filled with data from the 
+     *   associated table into the source table results. By default this is the underscored 
+     *   & singular name of the association so user in our example.
      *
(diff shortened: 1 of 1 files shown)


ndm2:
It's missing for all association methods. `bindingKey` is missing too.

src/ORM/Table.php
```diff
@@ -1135,6 +1135,9 @@ public function addAssociations(array $params)
      * - finder: The finder method to use when loading records from this association.
      *   Defaults to 'all'. When the strategy is 'join', only the fields, containments,
      *   and where conditions will be used from the finder.
+     * - propertyName: The property name that should be filled with data from the 
+     *   associated table into the source table results. By default this is the underscored 
+     *   & singular name of the association so user in our example.
```
ndm2, line 1140: Since there is no association config example here in the doc block, this should probably either just explain the inflected format, or provide a short example like:

> for an association alias of `ProductCategories`, this would be `product_category`

davidyell:
Yes, good point. I will try and find some time to submit the extra changes, thanks.

davidyell:
I have added the missing options to the docblocks for the other associations. Let me know if there are any more updates needed. Thanks

PR #15130, opened 2020-11-05
Update GA workflow to run jobs that were run on TravisCI

ADmad:


ADmad:
The .travis.yml can also be removed now.

PR #15132, opened 2020-11-06
3.x - Add github actions and reduce travis to prefer-lowest

othercorey:


othercorey:
This is mostly working except for some locale-specific tests.

othercorey:
I cannot get the memcache clear() test to pass on 5.6 even with a 2s delay.

othercorey:
> I cannot get the memcache clear() test to pass on 5.6 even with a 2s delay.

The issue appears to be `getAllKeys()` which is returning CLIENT_ERROR.

markstory:
We could skip the problematic test if getAllKeys() fails.

othercorey:
> We could skip the problematic test if getAllKeys() fails.

I cleaned up my fixes. I'm not familiar with memcached api someone else will have to finish this.

markstory:
I can take a look.

Issue #15111, opened 2020-10-31
Bulk insert using \Cake\ORM\Table->saveMany()


Shylajhaa:
This is a (multiple allowed):

* [ ] bug
* [x] enhancement
* [x] feature-discussion (RFC)

* CakePHP Version: 3.9.3.

### What you did
There are certain use-cases, where there are too many records to insert, and the ```saveMany()``` function doesn't seem to have a solution to insert rows in bulk (though this can be achieved using ```\Cake\ORM\Query->insert()```).

### What happened
The ```save``` and ```saveMany()``` functions have similar implementations. ```saveMany()``` iterates through the entities and insert queries are run one by one.

### What you expected to happen
The 2nd parameter of ```saveMany() - $options``` can support a boolean key to decide if the records should be inserted at once.

Moving from ```saveMany()``` to ```query()->insert()``` resulted in significant performance gains. So it would be great if this could be accommodated in the ORM.

markstory:
> Moving from saveMany() to query()->insert() resulted in significant performance gains. So it would be great if this could be accommodated in the ORM.

I'm not sure a bulk insert fits in the ORM paradigm, unless it also handles all the ORM callbacks. The reason saveMany() loops over the entities is so that the beforeSave, afterSave and application rule callbacks can be fired. A bulk insert would need to also trigger all of these events. Perhaps the events could be triggered in bulk around the insert query?

Shylajhaa:
If I'm not wrong, in the current implementation, the bulk updates/deletes doesn't invoke the lifecycle events (```beforeSave(), afterSave()```, etc.). The bulk inserts can skip the lifecycle events as well.

The single insert can approach can be used in cases where the callbacks are necessary.

dakota:
Perhaps an `insertAll` method to  compliment the existing `updateAll` and `deleteAll`?

dereuromark:
I had the same thought :)

Shylajhaa:
Agree with @dakota and @dereuromark 
This sounds perfect.

othercorey:
Those operate on fields while saveMany operates on entities.  What's the difference between insertAll and just using insert?

Shylajhaa:
@othercorey Idea is to minimize the number of queries, insertAll() will insert all rows at once.

othercorey:
And why wouldn't insert?

othercorey:
I would support letting `saveMany()` perform a single insert somehow.

ADmad:
[...]

PR #15136, opened 2020-11-07
Bulk Inserts for ORM/Query

T0mats0:
The following pull request address issue #15111 

As discussed in the issue I implemented the recommended interface insertAll($columns, $values). I added as a third optional parameter $types. I also added the ability to pass as values an array of column => values or just an array of values.  If only an array of values is passed, if a column indice isn't found, execution ends.

Tests include regular expression examination of the query to ensure a single bulk insert is being made.  As suggested in the issue I didn't move updateAll or deleteAll to ORM/Query. 

Now that I'm looking at it, I'm wondering if running a raw query without the usual ORM behaviors (validation, callbacks, ect) isn't a miss direction?

### DIFF
--- a/src/ORM/Query.php
+++ b/src/ORM/Query.php
@@ -1300,6 +1300,38 @@ class Query extends DatabaseQuery implements JsonSerializable, QueryInterface
     }
 
     /**
+     * Insert multiple rows of data in one call make use of bulk inserts.
+     *
+     * @param  array  $columns The columns to insert as ['columnName', 'columnName']
+     * @param  array  $values The values to insert as [['columnName' => 'value'], ... ] or ['value', ...]
+     * @param  array  $types A map between columns & their datatypes.
+     * @return $this
+     */
+    public function insertAll(array $columns, array $values, array $types = [])
+    {
+        $query = $this->insert($columns, $types);
+        foreach ($values as $value) {
+            $preparedValue = [];
+            foreach ($value as $columnName => $val) {
+                if (is_numeric($columnName)) {
+                    if (isset($columns[$columnName]) && !empty($columns[$columnName])) {
+                        $columnName = $columns[$columnName];
+                    } else {
+                        throw new InvalidArgumentException(
+                            'No column found at index ' . $columnName . ' in passed $columns'
+                        );
+                    }
+                }
+                $preparedValue[$columnName] = $val;
+            }
+
+            $query->values($preparedValue);
+        }
+
+        return $this;
+    }
+
+    /**
      * Returns a new Query that has automatic field aliasing disabled.
      *
      * @param \Cake\ORM\Table $table The table this query is starting on
--- a/tests/TestCase/ORM/QueryTest.php
+++ b/tests/TestCase/ORM/QueryTest.php
@@ -1842,6 +1842,94 @@ class QueryTest extends TestCase
     }
 
     /**
+     * Test insertAll method inserts more than one record
+     *
+     * @return void
+     */
+    public function testInsertAll()
+    {
+        $table = $this->getTableLocator()->get('articles');
+
+        $result = $table->query()
+            ->insertAll(
+                ['title'],
+                [
+                    ['title' => 'same ol same'],
+                    ['title' => 'other same ol'],
+                    ['title' => 'more of the same'],
+                ]
+            )
+            ->execute();
+
+        $result->closeCursor();
+
+        $this->assertInstanceOf('Cake\Database\StatementInterface', $result);
+        $this->assertEquals(3, $result->rowCount());
+    }
+
+    /**
+     * Test insert method with no column name strings in the values array
+     *
+     * @return void
+     */
+    public function testInsertAllWithNoColumnNames()
+    {
+        $table = $this->getTableLocator()->get('articles');
+        $result = $table->query()
+            ->insertAll(
+                ['title'],
+                [
+                    ['same ol same'],
+                    ['other same ol'],
+                    ['more of the same'],
+                ]
+            )
+            ->execute();
+
+        $result->closeCursor();
+        $this->assertInstanceOf('Cake\Database\StatementInterface', $result);
+        $this->assertEquals(3, $result->rowCount());
+    }
+
+    /**
+     * Inspect sql to insure bulk insert is being performed
+     */
+    public function testInsertAllWithSingleBulkInsertQuery()
+    {
+        $table = $this->getTableLocator()->get('articles');
+        $query = $table->query()
+            ->insertAll(
+                ['title'],
+                [
+                    ['same ol same'],
+                    ['other same ol'],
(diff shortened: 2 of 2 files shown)


src/ORM/Query.php
```diff
@@ -1299,6 +1299,38 @@ public function insert(array $columns, array $types = [])
         return parent::insert($columns, $types);
     }
 
+    /**
+     * Insert multiple rows of data in one call make use of bulk inserts.
+     *
+     * @param  array  $columns The columns to insert as ['columnName', 'columnName']
+     * @param  array  $values The values to insert as [['columnName' => 'value'], ... ] or ['value', ...]
```
dereuromark, line 1306: I am not sure if the numeric use case for values is common enough, as this would mean a single column, which is not really often happening, is it?

src/ORM/Query.php
```diff
@@ -1299,6 +1299,38 @@ public function insert(array $columns, array $types = [])
         return parent::insert($columns, $types);
     }
 
+    /**
+     * Insert multiple rows of data in one call make use of bulk inserts.
+     *
+     * @param  array  $columns The columns to insert as ['columnName', 'columnName']
```
dereuromark, line 1305: Are those `string[]` ?

src/ORM/Query.php
T0mats0: Yes you can pass an array with just the values, those values line up with the columns.

src/ORM/Query.php
T0mats0: It's an array of string, yes. It's like the other column array for insert.

tests/TestCase/ORM/QueryTest.php
```diff
@@ -1841,6 +1841,94 @@ public function testInsert()
...
+    {
+        $table = $this->getTableLocator()->get('articles');
+        $result = $table->query()
+            ->insertAll(
+                ['title'],
+                [
+                    ['same ol same'],
+                    ['other same ol'],
+                    ['more of the same'],
+                ]
```
markstory, line 1885: I think we should choose one API or the other. Supporting both numerically indexed and keyed arrays seems un-necessary. Supporting only keyed arrays would let us drop the `$columns` parameter which is nice.

src/ORM/Query.php
```diff
@@ -1299,6 +1299,38 @@ public function insert(array $columns, array $types = [])
...
+     * @param  array  $values The values to insert as [['columnName' => 'value'], ... ] or ['value', ...]
+     * @param  array  $types A map between columns & their datatypes.
+     * @return $this
+     */
+    public function insertAll(array $columns, array $values, array $types = [])
+    {
+        $query = $this->insert($columns, $types);
```
ADmad, line 1312: `insert()` returns `$this` so there's no need to assign the retun value. You can just use `$this` instead of `$query` later in the method.

tests/TestCase/ORM/QueryTest.php
```diff
@@ -1841,6 +1841,94 @@ public function testInsert()
...
+     */
+    public function testInsertAll()
+    {
+        $table = $this->getTableLocator()->get('articles');
+
+        $result = $table->query()
+            ->insertAll(
+                ['title'],
+                [
+                    ['title' => 'same ol same'],
```
ADmad, line 1857: There should be a test with non string values too showing the that type map set from table is used.

tests/TestCase/ORM/QueryTest.php
T0mats0: I missed that note. Okay. I'll re-work it so no numerical index support. With the signature ``` public function insertAll(array $inserts, array $types = [])``` I will derive the columns from the first element/insert of the array.

Let me know if types are to be passed at all.

src/ORM/Query.php
```diff
@@ -1299,6 +1299,38 @@ public function insert(array $columns, array $types = [])
...
 
+    /**
+     * Insert multiple rows of data in one call make use of bulk inserts.
+     *
+     * @param  array  $columns The columns to insert as ['columnName', 'columnName']
+     * @param  array  $values The values to insert as [['columnName' => 'value'], ... ] or ['value', ...]
+     * @param  array  $types A map between columns & their datatypes.
+     * @return $this
+     */
+    public function insertAll(array $columns, array $values, array $types = [])
```
ADmad, line 1310: I am leaning towards the name `insertMany()`, what do others think?

dereuromark: Theoretically, I would agree.
But given that we have updateAll(), deleteAll() as similar operations, and saveMany() etc as callback ops, this seems at least more consistent in the current naming scheme if we use insertAll() here.

ADmad: Well this method is in a different class and Query doesn't have any callbacks.

lorenzo:
I think this method is misplaced. We have almost the same function in the Query class, which is a more natural place. Tables only deal with entities in all methods, and I think this could be mi leading to many, as expressed in this PR description.

src/ORM/Query.php
T0mats0: I named it insertAll at first to match other functions like @dereuromark  suggested. That was my first thought too.  insertMany describes it better than insertAll to me. The language difference is subtle. The difference might be explained this way, "I hope every insert() call inserts all the data". This insertMany() function inserts many records. 

I could live with it either way. But the word many describes it better to me.

T0mats0:
@lorenzo Yet, having these kind of shortcuts make nice features for developers. Might it be better placed in a class like TableQueries or TableUtilities? Then it would be clearer that we are not dealing with the ORM and the developer gets the ease of use of not needing to pass all the things like the table name?

othercorey:
I don't understand why this is in ORM\Query either. If we are operating on tables, then why do we pass in columns?

ADmad:
> Tables only deal with entities in all methods...

That's precisely the reason I didn't want this method on ` `Table` and proposed adding it to `ORM\Query` instead as a compromise. 

I don't see the need either this method as one can just call `Query::values()` multiple times to insert multiple rows in a single query (as [documented](https://book.cakephp.org/4/en/orm/query-builder.html#inserting-data)).

T0mats0:
How about this:

1. Create a trait or behavior that can be applied to a table that adds function like insertMany(), updateAll(), deleteAll(). That way we won't be violating SRP and anyone applying the trait/behavior to table class will know what they are getting into.
2. Add insertMany to Database\Query as a short cut to calling values() multiple times in the calling code -- having an interface on an object is the best form of documentation.

ADmad:
You can choose option 1 can create your own plugin with required trait/behavior :)

Others have created similar plugins with convenience functions, e.g. [this](https://github.com/itosho/easy-query).

T0mats0:
:skull: Looks like this PR and issue are going no where, closing.

PR #15138, opened 2020-11-08
GA CI workflow

ADmad:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -118,7 +118,7 @@ jobs:
         fi
 
     - name: Submit code coverage
-      if: success() && matrix.php-version == '7.4'
+      if: matrix.php-version == '7.4'
       uses: codecov/codecov-action@v1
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -118,7 +118,7 @@ jobs:
         fi
 
     - name: Submit code coverage
-      if: success() && matrix.php-version == '7.4'
```
othercorey, line 121: I don't have a problem with these changes, but wondering what this fixes.

othercorey: Is it so the patch codecov % isn't way off for a single failure?

othercorey: Now I'm thinking the success() is just redundant.

ADmad: > Now I'm thinking the success() is just redundant.

Yes, as stated in the commit message.

PR #15140, opened 2020-11-09
Dont use a dist for a non changing file

dereuromark:
The .dist should only be added if a local version of it can and should be made in case of custom modifications
e.g. local passwords, ports

In this case, with CS, this is a non changing file IMO.
It creates side effects if we execute this from inside vendor, so better to use the right default.


othercorey:
This should be `phpcs.xml`

dereuromark:
My bad
Fixed.

ADmad:
.gitattributes file also probably needs to be updated

Issue #15141, opened 2020-11-10
Wrong second paginator query in beforeFind


bancer:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.8.12.
* Platform and Target: Apache, MySQL.

### What you did
In the controller class: `$Paginator->paginate($Query)->toArray();`.
In the table class:
```
    public function beforeFind(Event $event, Query $query, ArrayObject $options, $primary)
    {
        \Cake\Log\Log::debug($query->sql());
    }
```

### What happened
In the queries log I got two paginator queries - one that retrieves the data and another that counts. But in the debug log I got two identical queries that retrieves the data. So the second query is supposed to be like `SELECT (COUNT(*)) AS count FROM ...` in beforeFind method but it is not.

Partial stacktraces of both paginator queries.
The query that retrieves records:
```
App\Model\Table\MyTable::beforeFind() - APP/src/Model/Table/MyTable.php, line 153
Cake\Event\EventManager::_callListener() - APP/Vendor/cakephp/event/EventManager.php, line 352
Cake\Event\EventManager::dispatch() - APP/Vendor/cakephp/event/EventManager.php, line 329
Cake\ORM\Table::dispatchEvent() - APP/Vendor/cakephp/event/EventDispatcherTrait.php, line 113
Cake\ORM\Query::triggerBeforeFind() - APP/Vendor/cakephp/orm/Query.php, line 1108
Cake\ORM\Query::_execute() - APP/Vendor/cakephp/orm/Query.php, line 1137
Cake\ORM\Query::_all() - APP/Vendor/cakephp/datasource/QueryTrait.php, line 310
Cake\ORM\Query::all() - APP/Vendor/cakephp/orm/Query.php, line 1088
Cake\Datasource\Paginator::paginate() - APP/Vendor/cakephp/datasource/Paginator.php, line 186
```
The query that counts records:
```
App\Model\Table\MyTable::beforeFind() - APP/src/Model/Table/MyTable.php, line 153
Cake\Event\EventManager::_callListener() - APP/Vendor/cakephp/event/EventManager.php, line 352
Cake\Event\EventManager::dispatch() - APP/Vendor/cakephp/event/EventManager.php, line 329
Cake\ORM\Table::dispatchEvent() - APP/Vendor/cakephp/event/EventDispatcherTrait.php, line 113
[...]

PR #5151, opened 2014-11-13
Fixes #5150

dakota:
As reported in #5150, some databases don't like including `order by` along with `count(*)`, this is usually not a problem as the `Query::count()` method creates a clean copy of the query which excludes any counts or limits. However, if the query is modified in the beforeFind callback, the count() method will not clean up any added parts.

This PR corrects that issue by triggering beforeFind as part of the clean-up process.

PR #15143, opened 2020-11-10
Remove the call to triggerBeforeFind from cleanCopy method

bancer:
#15141

### DIFF
--- a/src/ORM/Query.php
+++ b/src/ORM/Query.php
@@ -863,7 +863,6 @@ class Query extends DatabaseQuery implements JsonSerializable, QueryInterface
     {
         $clone = clone $this;
         $clone->setEagerLoader(clone $this->getEagerLoader());
-        $clone->triggerBeforeFind();
         $clone->enableAutoFields(false);
         $clone->limit(null);
         $clone->order([], true);
--- a/tests/TestCase/ORM/QueryTest.php
+++ b/tests/TestCase/ORM/QueryTest.php
@@ -2935,6 +2935,7 @@ class QueryTest extends TestCase
     public function testCountPaginatorBeforeFind()
     {
         $table = $this->getTableLocator()->get('Articles');
+        $table->getConnection()->getDriver()->enableAutoQuoting(false);
         $table->hasMany('Comments');
         $table->getEventManager()
             ->on('Model.beforeFind', function (Event $event, $query) {
--- a/tests/TestCase/ORM/QueryTest.php
+++ b/tests/TestCase/ORM/QueryTest.php
@@ -2935,7 +2935,6 @@ class QueryTest extends TestCase
     public function testCountPaginatorBeforeFind()
     {
         $table = $this->getTableLocator()->get('Articles');
-        $table->getConnection()->getDriver()->enableAutoQuoting(false);
         $table->hasMany('Comments');
         $table->getEventManager()
             ->on('Model.beforeFind', function (Event $event, $query) {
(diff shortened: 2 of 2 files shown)


ndm2:
Won't this possibly break "complex" query assesment? [**`Query::_performCount()`**](https://github.com/cakephp/cakephp/blob/master/src/ORM/Query.php#L902) will asses the complexity, and use the query as a subquery if determined neccesary. Moving triggering `beforeFind()` so that it will apply to the count query, will cause `Query::_performCount()` to possibly not see the final count query when it's being modified in `beforeFind()`, thus it will not be able to asses its final complexity.

I have a feeling that this change would generally break many applications out there that never expected count queries to end up in their `beforeFind()` callbacks, as that seems to have been the behavior ever since CakePHP 3 was released.

bancer:
@ndm2 Could you give an example so that it would be possible to write a unit test for that?

ADmad:
@ndm2 Yes, given how long this behaviour has existed it would definitely need a good amount of thought. Perhaps we should even consider only changing things in 4.x (if at all) and leave 3.x as is.

ADmad:
@bancer What use case do you have that's causing problem for you due to this behaviour?

bancer:
> @bancer What use case do you have that's causing problem for you due to this behaviour?

We need to migrate some code from 2.x to 3.x. In some 2.x models in beforeFind callback we change the count query generated by the paginator in order to achieve acceptable performance. It is not possible to do that in 3.x table class at the moment because the query object is wrong in beforeFind for that paginator's count query.

ADmad:
Since the pagination code has been extracted into a separate `Paginator` class in 3.9, what you want can be achieved in a better way by extending it and overriding its `getCount()` method. The PaginatorComponent can be configured to use your own paginator class.

ndm2:
> @ndm2 Could you give an example so that it would be possible to write a unit test for that?

I'd have to do quite some trial and error first, I'm not quite sure what underlying problems exactly `Query::_performCount()` is trying to prevent by doing this subquery for "complex" cases. Given that there's no failing test yet, the problem might not have been tested properly, or the feature might even not be required, but that would need to be evaluated properly.

ndm2:
> Since the pagination code has been extracted into a separate `Paginator` class in 3.9, what you want can be achieved in a better way by extending it and overriding its `getCount()` method. The PaginatorComponent can be configured to use your own paginator class.

Defining a custom count query in `beforeFind()` via `Query::counter()` might also work.

ADmad:
Right, using Query::counter() might be a simpler option.

bancer:
> Since the pagination code has been extracted into a separate Paginator class in 3.9, what you want can be achieved in a better way by extending it and overriding its getCount() method.

There is no `getCount()` method. In my understanding the count query is generated in `paginate()` method - https://api.cakephp.org/3.8/source-class-Cake.Datasource.Paginator.html#185:
```
$cleanQuery = clone $query;
$results = $query->all();
$numResults = count($results);
$count = $cleanQuery->count();
```
Refactoring this code to `getCount()` method would actually make it possible to override it.

> Defining a custom count query in beforeFind()...

That's what we basically do in 2.x but cannot do in 3.x because it is not possible to detect the paginator's count query in beforeFind.

> Given that there seems to be no failing existing test related to that, the problem might not have been tested properly, or the feature might even not be required, but that would need to be evaluated properly.

According to [pull request](https://github.com/cakephp/cakephp/pull/5151/files) where `$query->triggerBeforeFind();` was added to `cleanCopy()` method the intention was to remove `ORDER BY` clause from the count query - https://github.com/cakephp/cakephp/issues/5150. That unit test still passes.

bancer:
Ah... `getCount()` exists in 3.9.

othercorey:
Closing old draft.

PR #15149, opened 2020-11-11
Retain flash also for redirect

dereuromark:
This worked in 3.x
```php
$this->post(['controller' => 'ReleaseGroups', 'action' => 'initAutoRelease', $id]);

$this->assertResponseCode(302);

$flashMessages = $this->_requestSession->read('Flash.flash');
$expected = [
    'flash/success',
];
$this->assertSame($expected, Hash::extract($flashMessages, '{n}.element'));
```

Then we upgraded the app this week to 4.1+.

In 4.x I then saw this test failing and added
```php
$this->enableRetainFlashMessages();
```
at the top

Still not finding the retained messages.

I then found out why, the redirect is not covered going from 3.x to 4.x
So here the bugfix for this regression.

I still feel like the session should still contain those after the redirect (inside the test method). So not sure why the session is dropped this quickly.

### DIFF
--- a/src/TestSuite/IntegrationTestTrait.php
+++ b/src/TestSuite/IntegrationTestTrait.php
@@ -536,7 +536,8 @@ trait IntegrationTestTrait
         }
         $this->_controller = $controller;
         $events = $controller->getEventManager();
-        $events->on('Controller.beforeRedirect', function () use ($controller): void {
+        $events->on('Controller.beforeRedirect', function ($event): void {
+            $controller = $event->getSubject();
             if ($this->_retainFlashMessages) {
                 $this->_flashMessages = $controller->getRequest()->getSession()->read('Flash');
             }
--- a/src/TestSuite/IntegrationTestTrait.php
+++ b/src/TestSuite/IntegrationTestTrait.php
@@ -536,6 +536,11 @@ trait IntegrationTestTrait
         }
         $this->_controller = $controller;
         $events = $controller->getEventManager();
+        $events->on('Controller.beforeRedirect', function () use ($controller): void {
+            if ($this->_retainFlashMessages) {
+                $this->_flashMessages = $controller->getRequest()->getSession()->read('Flash');
+            }
+        });
         $events->on('View.beforeRender', function ($event, $viewFile) use ($controller): void {
             if (!$this->_viewName) {
                 $this->_viewName = $viewFile;


dereuromark:
https://github.com/cakephp/cakephp/pull/15150 shows that it should be possible actually given current code.
So there must be something else going on here...

dereuromark:
Any app I try to use flash message assert and redirect, they always are empty without this fix.
Any ideas?

dereuromark:
If I want to also check other session content after the redirect the only solution I could find to not having a wiped out session content is this:

```diff
===================================================================
--- src/TestSuite/IntegrationTestTrait.php	(revision b77db1a76285f6b4b4e752ea504828dfe47836f1)
+++ src/TestSuite/IntegrationTestTrait.php	(date 1605658105962)
@@ -505,7 +505,6 @@
         try {
             $request = $this->_buildRequest($url, $method, $data);
             $response = $dispatcher->execute($request);
-            $this->_requestSession = $request['session'];
             if ($this->_retainFlashMessages && $this->_flashMessages) {
                 $this->_requestSession->write('Flash', $this->_flashMessages);
             }
@@ -549,6 +548,7 @@
             if ($this->_retainFlashMessages) {
                 $this->_flashMessages = $controller->getRequest()->getSession()->read('Flash');
             }
+            $this->_requestSession = clone($controller->getRequest()->getSession());
         });
         $events->on('View.beforeRender', function ($event, $viewFile) use ($controller): void {
             if (!$this->_viewName) {

This way the session content is still available after the "fake" redirect, and the asserts that used to work in 3.x still work in 4.x.

**Comment by markstory** [comment]
I think we still need retainFlashMessages as it also allows you test flash messages *after* they've been consumed by the helper.

**Comment by dereuromark** [comment]
I ended up with a hacked trait that uses

    $this->_requestSessionCopy = clone($controller->getRequest()->getSession());

As the issue is only related to redirects, and would otherwise mess up the other cases.
With this, and reading the copy on redirect asserts works now...

I still don't get why this is necessary, as the core tests show it shouldn't be.
So far I checked also app and plugin callbacks but couldn't find any cause for this loss of data.

**Comment by othercorey** [comment]
Do we need to do this for 4.2?

**Comment by dereuromark** [comment]
I really don't know how to proceed with this one.
I have my hack on my apps and those work for now - so if you also don't see how we could approach this, we can also close.

**Comment by markstory** [comment]
The session being empty after a redirect could be caused by the session being closed and then reopened. I think this change is reasonably safe, but it would be good to have a test that captures what doesn't work now and will work with these changes.

File: src/TestSuite/IntegrationTestTrait.php [hunk 1]
```diff
@@ -536,6 +536,11 @@ public function controllerSpy(EventInterface $event, ?Controller $controller = n
         }
         $this->_controller = $controller;
         $events = $controller->getEventManager();
+        $events->on('Controller.beforeRedirect', function () use ($controller): void {
+            if ($this->_retainFlashMessages) {
```
markstory, line 540:
```suggestion
        $events->on('Controller.beforeRedirect', function ($event): void {
            $controller = $event->getSubject();
            if ($this->_retainFlashMessages) {
```

Saves having another closure reference.

PR #15152, opened 2020-11-11
4.2 fallback table class

ADmad:
The Cake ORM's "auto table" feature which allows accessing db tables without having to create concrete classes for them is a boon and a curse :slightly_smiling_face:.

It can often lead to a lot of hair pulling, especially for new users, if one makes a typo in a table name/alias or forgets to use prefix for plugin tables and keep wondering why their table class is not loaded. We added a list of auto generated models to debug kit's SQL Log tab but it hasn't helped much as a lot of new users are not aware of it.

So this PR adds the ability to turn off the fallback to `Cake\ORM\Table` class in the `TableLocator`, when a concrete class does not exist for a table. On can now add the following in `Application::bootstrap()` for example to turn off the fallback/"auto model" feature.

```php
FactoryLocator::add('Table', (new TableLocator())->useFallbackClass(false));
```

To maintain some balance and not require having to create concrete classes for every database table, there are some cases where use of fallback class is still always enabled:
1. Join/junction table for belongsToMany association. Unless you specify `through` option you still won't need to create concrete class for join table.
2. Tables used by `TranslateBehavior` to store translated values (for both `EavStrategy` and `ShadowTableStrategy`).

### DIFF
--- a/src/ORM/Association/BelongsToMany.php
+++ b/src/ORM/Association/BelongsToMany.php
@@ -270,7 +270,7 @@ class BelongsToMany extends Association
 
             $config = [];
             if (!$tableLocator->exists($tableAlias)) {
-                $config = ['table' => $tableName];
+                $config = ['table' => $tableName, 'useFallbackClass' => true];
 
                 // Propagate the connection if we'll get an auto-model
                 if (!App::className($tableAlias, 'Model/Table', 'Table')) {
--- a/src/ORM/Association/BelongsToMany.php
+++ b/src/ORM/Association/BelongsToMany.php
@@ -270,7 +270,7 @@ class BelongsToMany extends Association
 
             $config = [];
             if (!$tableLocator->exists($tableAlias)) {
-                $config = ['table' => $tableName, 'useFallbackClass' => true];
+                $config = ['table' => $tableName, 'allowFallbackClass' => true];
 
                 // Propagate the connection if we'll get an auto-model
                 if (!App::className($tableAlias, 'Model/Table', 'Table')) {
--- a/src/ORM/Locator/TableLocator.php
+++ b/src/ORM/Locator/TableLocator.php
@@ -73,7 +73,7 @@ class TableLocator extends AbstractLocator implements LocatorInterface
      *
      * @var bool
      */
-    protected $useFallbackClass = true;
+    protected $allowFallbackClass = true;
 
     /**
      * Constructor.
@@ -100,12 +100,12 @@ class TableLocator extends AbstractLocator implements LocatorInterface
      * Controls whether a fallback class should be used to create a table
      * instance if a concrete class for alias used in `get()` could not be found.
      *
-     * @param bool $enable Flag to enable or disable fallback
+     * @param bool $allow Flag to enable or disable fallback
      * @return $this
      */
-    public function useFallbackClass(bool $enable)
+    public function allowFallbackClass(bool $allow)
     {
-        $this->useFallbackClass = $enable;
+        $this->allowFallbackClass = $allow;
 
         return $this;
     }
@@ -223,11 +223,11 @@ class TableLocator extends AbstractLocator implements LocatorInterface
             $options += $this->_config[$alias];
         }
 
-        $useFallbackClass = $options['allowFallbackClass'] ?? $this->useFallbackClass;
+        $allowFallbackClass = $options['allowFallbackClass'] ?? $this->allowFallbackClass;
         $className = $this->_getClassName($alias, $options);
         if ($className) {
             $options['className'] = $className;
-        } elseif ($useFallbackClass) {
+        } elseif ($allowFallbackClass) {
             if (empty($options['className'])) {
                 $options['className'] = $alias;
             }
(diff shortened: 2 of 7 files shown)


src/ORM/Locator/TableLocator.php
```diff
@@ -78,6 +94,33 @@ public function __construct(?array $locations = null)
         }
     }
 
+    /**
+     * Set if fallback class should be used.
+     *
+     * @param bool $enable Flag to enable or disable fallback
+     * @return $this
+     */
+    public function useFallbackClass(bool $enable)
```
markstory, line 103: Are you thinking that this should become the default mode? How would enabling this globally work with Bake which also relies on auto-models extensively.

ADmad: Eventually yes but I would like to hear opinions on whether we should do it 4.2. We could keep the default behavior same in core and instead add the code I showed in app skeleton for 4.2 also.

In either case we can check for CLI and turn on fallback mode in `Application::bootstrap()` or in bake's `Plugin::bootstrap()` or `BakeCommand`'s constructor.

markstory: > Eventually yes but I would like to hear opinions on whether we should do it for 4.2. We could keep the default behavior same in core and instead add the code I showed in app skeleton for 4.2 also.

Ok. I agree with your earlier statement that auto-tables are a boon and a curse. They are great for getting started and prototyping but quickly become a liability.

I like your idea of enabling auto-tables in bake and the console command to smooth over the experience for new applications.

ndm2: I'd like to see this rather sooner than later, and if it basically only affects newly create applications, then it shouldn't really be a problem. The only problem might be people upgrading their bootstrap according to `cakephp/app` when upgrading to 4.2, but I would assume that the people that do that are usually the more savvy ones.

ndm2:
Should this maybe affect entities too? I haven't seen nearly as many people yet having problems with "auto-entites" as opposed to "auto-tables", but it happened.

src/ORM/Locator/TableLocator.php
ADmad: We hope to have 4.2 out "soon" to make this feature available :slightly_smiling_face:

ADmad:
> Should this maybe affect entities too?

We could probably provide a way to turn of "auto entities" too but I don't turning of "auto tables" to automatically turn off "auto entities" too. It's quite common to have entities which don't require anything special and if you aren't patching entities from user provided data then customizing the `$_accessible` property isn't required either.

dereuromark:
With entities it is more the problem of fields not being accessible and silent data loss you just don't realize. :)

othercorey:
My only thought is whether `useFallbackClass` is clear. Should it be more specific like `allowMissingTableClass` or do we call it "fallback class" in the documentation?

ADmad:
> With entities it is more the problem of fields not being accessible and silent data loss you just don't realize. :)

Yup that's another issue we should try to tackle.

> My only thought is whether useFallbackClass is clear. Should it be more specific like allowMissingTableClass or do we call it "fallback class" in the documentation?

"fallback class" is a proper term for the behavior/feature IMO. I improved the docblocks a bit for the new methods I added.  Regardless of what name we need to improve the documenting for it. I don't remember this feature being explicitly stated in the docs.

src/ORM/Association/BelongsToMany.php
```diff
@@ -270,7 +270,7 @@ public function junction($table = null): Table
 
             $config = [];
             if (!$tableLocator->exists($tableAlias)) {
-                $config = ['table' => $tableName];
+                $config = ['table' => $tableName, 'useFallbackClass' => true];
```
markstory, line 273: Should the option be `allowFallbackClass`? Could `use` be misinterpreted, as 'always use'?

dereuromark: Could then also be shorter with `allowFallback` maybe.

ADmad: Okay, I'll rename the option to `allowFallbackClass`, I prefer keeping the `Class` suffix as it clearly indicates what we are allowing :).

The method name `useFallbackClass()` is fine right?

dereuromark: Hm I wonder if the method name should also match the config key
But I am fine either way.

src/ORM/Locator/TableLocator.php
```diff
@@ -78,6 +94,40 @@ public function __construct(?array $locations = null)
...
+    /**
+     * Set if fallback class should be used.
+     *
+     * Controls whether a fallback class should be used to create a table
+     * instance if a concrete class for alias used in `get()` could not be found.
+     *
+     * @param bool $allow Flag to enable or disable fallback
+     * @return $this
+     */
+    public function allowFallbackClass(bool $allow)
```
ADmad, line 106: We already have `I18n::useFallback()`, that's why I had chose "use" prefix for consistency.

othercorey: allowFallbackClass feels correct, but if useFallback is strongly preferred then I won't object.

PR #15154, opened 2020-11-12
[2.x] php-8.0-compat: fix "Error: Unknown named parameter $subject"

mfn:
## Summary
PHP 8 added support for named parameters however will throw an error when
a named parameter, which does not exist, will be called/appears.

In this case the code uses `array_filter(compact('subject))` to conveniently
prepend `$subject` to the callback, if it's non-null.

It's then combined with `$params` but mixing a hash wih a sequential array
and then end result is `['subject' => …, ]` which then triggers this
TypeError in PHP 8.

The solution applies `array_values()` on above array as to remove the keys
and then the error disappears.

With this change, our internal test suite (2k tests) is green on both PHP8
and PHP7.4, aka this change is expecte to be backwards compatible.

### DIFF
--- a/lib/Cake/Utility/ObjectCollection.php
+++ b/lib/Cake/Utility/ObjectCollection.php
@@ -126,7 +126,7 @@ abstract class ObjectCollection {
 		}
 		$result = null;
 		foreach ($list as $name) {
-			$result = call_user_func_array(array($this->_loaded[$name], $callback), array_filter(compact('subject')) + $params);
+			$result = call_user_func_array(array($this->_loaded[$name], $callback), array_values(array_filter(compact('subject')) + $params));
 			if ($options['collectReturn'] === true) {
 				$collected[] = $result;
 			}


dereuromark:
Can you provide a test case for this, please?

mfn:
First, I think the failed builds are unrelated, as they contain errors referencing `MemcachedEngineTest::testClear` 🤷‍♀️ and they only apply to PHP 5.6; others are green.

> Can you provide a test case for this, please?

So, that's a bit my crux here right now 😅

_If_ the travis CI would run on PHP 8, it would throw the error _without_ this PR.

But currently travis does not include running on PHP 8 at all.

@dereuromark I'm unsure how to proceed, do you have a suggestion?

markstory:
While we can merge this as it doesn't cause any breakage in existing tests. As per #15065 we aren't planning to 'support' PHP8 in 2.x. There is only a month of bugfix support left in 2.x's lifetime, after which we'll only be doing security vulnerability fixes.

mfn:
I understand and I, like others, would highly appreciate if this (or better: https://github.com/cakephp/cakephp/pull/15065 which is the same, to give proper credit to the first author) can be merged nevertheless.

As I mentioned, I also made sure this change is compatible with older PHP versions and in fact we're still a bit away from upgrading to PHP8, this is just one of the issues we encountered so far.

dereuromark:
For this to be merged you need to also revert https://github.com/cakephp/cakephp/blob/2.x/composer.json#L21 then here.

ADmad:
> For this to be merged you need to also revert https://github.com/cakephp/cakephp/blob/2.x/composer.json#L21 then here.

The change is fine for lower PHP versions too regardless of whether the constraints. Also one call still use `--ignore-platform-deps` to install this on PHP 8 even with that constraint.

dereuromark:
That sounds weird
Why doing this when it would work normally, the risk of someone then being forced to use this and deploying with other platform depts unchecked is high - and not reasonable IMO.

garas:
>  Why doing this when it would work normally, the risk of someone then being forced to use this and deploying with other platform depts unchecked is high - and not reasonable IMO.

Composer 2 now allows to ignore only PHP version, but still check extensions: `--ignore-platform-req=php`.

dereuromark:
Still, whats the gain in that if it aint even true?
The opposite would be more helpful IMO. Allowing but with caution docs.

lib/Cake/Utility/ObjectCollection.php
```diff
@@ -126,7 +126,7 @@ public function trigger($callback, $params = array(), $options = array()) {
 		}
 		$result = null;
 		foreach ($list as $name) {
-			$result = call_user_func_array(array($this->_loaded[$name], $callback), array_filter(compact('subject')) + $params);
+			$result = call_user_func_array(array($this->_loaded[$name], $callback), array_values(array_filter(compact('subject')) + $params));
```
othercorey, line 129:
```suggestion
			$result = call_user_func_array(array($this->_loaded[$name], $callback), array_filter(array($subject)) + $params);
```
Let's make this simpler.

mfn: @othercorey this breaks the test suite for me (right now tested this on 7.4):
```
ArgumentCountError : Too few arguments to function ContainableBehavior::beforeFind(), 1 passed in /app/Vendor/cakephp/cakephp/lib/Cake/Utility/ObjectCollection.php on line 129 and exactly 2 expected
 /app/Vendor/cakephp/cakephp/lib/Cake/Model/Behavior/ContainableBehavior.php:93
 /app/Vendor/cakephp/cakephp/lib/Cake/Utility/ObjectCollection.php:129
 /app/Vendor/cakephp/cakephp/lib/Cake/Event/CakeEventManager.php:244
 /app/Vendor/cakephp/cakephp/lib/Cake/Model/Model.php:3110
 /app/Vendor/cakephp/cakephp/lib/Cake/Model/Model.php:3020
 /app/Model/AppModel.php:24
 /app/Vendor/cakephp/cakephp/lib/Cake/Model/Datasource/DboSource.php:636
 /app/Vendor/cakephp/cakephp/lib/Cake/Model/Model.php:834
 /app/Test/…

```

othercorey: It should be the same except without the $subject key. What is missing?

mfn: PS: was now able to test in PHP8, same issue/stacktrace.

markstory:
I'm going to merge this as it doesn't cause any breaks. We won't be removing the PHP constraints in composer.json though. As we're ~17 days from the end of active support for 2.x, we won't be adding or implying that 2.x will work in PHP8, but until December 15 I'm happy to merge community pull requests before doing one more release.

PR #15157, opened 2020-11-14
Add FlashMessage utility class.

ADmad:
I found the need to be able to set flash message from middlewares before redirection. While that could be done by directly setting the `Flash.flash` session key with required array format, having to rely on an internal implementation detail isn't pretty. Also I noticed that the `FlashComponent`s code and public API could be improved.

So here's `FlashMessage` class to solve the above issues. I can finish up the work, add tests and retro fit the `FlashComponent` to use `FlashMessage` internally if the idea is approved.

### DIFF
--- a/src/Http/FlashMessage.php
+++ b/src/Http/FlashMessage.php
@@ -60,7 +60,8 @@ class FlashMessage
     }
 
     /**
-     * Used to set a session variable that can be used to output messages in the view.
+     * Store flash messages that can be output in the view.
+     *
      * If you make consecutive calls to this method, the messages will stack
      * (if they are set with the same flash key)
      *
--- a/src/Http/FlashMessage.php
+++ b/src/Http/FlashMessage.php
@@ -36,6 +36,7 @@ class FlashMessage
     protected $_defaultConfig = [
         'key' => 'flash',
         'element' => 'default',
+        'plugin' => null,
         'params' => [],
         'clear' => false,
         'duplicate' => true,
@@ -67,7 +68,9 @@ class FlashMessage
      * ### Options:
      *
      * - `key` The key to set under the session's Flash key.
-     * - `element` The element used to render the flash message.
+     * - `element` The element used to render the flash message. You can use
+     *     `'SomePlugin.name'` style value for flash elements from a plugin.
+     * - `plugin` Plugin name to use element from.
      * - `params` An array of variables to be made available to the element.
      * - `clear` A bool stating if the current stack should be cleared to start a new one.
      * - `escape` Set to false to allow templates to print out HTML content.
@@ -86,6 +89,9 @@ class FlashMessage
         }
 
         [$plugin, $element] = pluginSplit($options['element']);
+        if ($options['plugin']) {
+            $plugin = $options['plugin'];
+        }
 
         if ($plugin) {
             $options['element'] = $plugin . '.flash/' . $element;
--- a/src/Http/FlashMessage.php
+++ b/src/Http/FlashMessage.php
@@ -157,12 +157,7 @@ class FlashMessage
      */
     public function consume(string $key): ?array
     {
-        $messages = $this->session->read("Flash.{$key}");
-        if ($messages !== null) {
-            $this->session->delete("Flash.{$key}");
-        }
-
-        return $messages;
+        return $this->session->consume("Flash.{$key}");
     }
 
     /**
--- a/src/Http/FlashMessage.php
+++ b/src/Http/FlashMessage.php
@@ -172,7 +172,6 @@ class FlashMessage
     public function success(string $message, array $options = []): void
     {
         $options['element'] = 'success';
-
         $this->set($message, $options);
     }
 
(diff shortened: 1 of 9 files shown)


dereuromark:
I like the idea of the controller code (and thus component) becoming a bit more slim again, and offloading the business logic further down into own classes that are more easily reusable across the system.
Especially with middleware now hijacking the general concept of components and how they used to be able to access request/session object, I think this is the right way to go.
We have to move those components also along into the direction of where middleware went :+1:

src/Http/FlashMessage.php
```diff
@@ -0,0 +1,195 @@
...
+
+use Cake\Core\InstanceConfigTrait;
+use Cake\Utility\Hash;
+use Throwable;
+
+/**
+ * The FlashMessage class provides a way for you to write a flash variable
+ * to the session, to be rendered in a view with the FlashHelper.
+ */
+class FlashMessage
```
markstory, line 27: Should the request have a method or an attribute that exposes flash messages? Or perhaps a method on the `Session` object?

```php
// My preference, but doesn't follow the `getXYZ` pattern.
$this->request->flash()->set(...);

// Simple & mirrors how sessions are exposed.
$this->request->getFlash()->set(...);

// Fits within the PSR spec, and we could still have getFlash() or flash()
$this->request->getAttribute('flash')->set(...);

// Is a bit more verbose, but workable.
$this->request->getSession()->getFlash()->set(...);
```

ADmad: Since there's no corresponding setter method I would prefer `request->flash()`, (even though for session we have `getSession()`). We deprecated `Behavior::setTable()` in favor of `Behavior::table()` for the same reason.

Internally it would be set as a request attribute so `request->getAttribute('flash')` would also work.

dereuromark: I think behavior table() was because of collision with Table::setTable() (meaning table string of database table)
But I agree, internally as attribute might suffice for the 4.x lifecycle
And with more experience on the subject we can still think about changes for 5.x when the time comes.

ADmad: > I think behavior table() was because of collision with Table::setTable() (meaning table string of database table)

Ah, I forgot that. Well I can live with `request->getFlash()` too, `request->getAttribute('flash')` too cumbersome and can't be auto completed by IDEs.

dereuromark: It can with IdeHelper help for PHPStorm at least ;)

But all right, getFlash() sounds reasonable then.

ADmad:
I have updated `ServerRequest` as per discussion above and updated `FlashHelper` to use the `FlashMessage` instance instead of directly using session.

Earlier I had thought of retro fitting the `FlashComponent` to use `FlashMessage` internally but with `FlashMessage` instance now easily accessing through the request instance I would rather we just deprecate `FlashComponent`.

Few notable differences currently between `FlashComponent` and `FlashMessage`:
1. Instead of magic `__call()` method `FlashMessage` has 2 concrete convenience methods `success()` and `error()`. Magic call methods don't make for good public APIs. Also I don't other element types besides these would be used commonly enough. The app skeleton also ships with additional flash elements for only these 2 types.
2. `FlashComponent::__call()` supports a `plugin` options which `set()` doesn't, which is manipulated to generate `element` option which uses a plugin's element. I haven't implemented that option for `FlashMessage`. If people think it's useful enough than it should be implemented for `set()` too IMO (which isn't the case with `FlashComponent`) instead of just the convenience methods. Personally I wasn't too keen on this option when it was initially added either.

src/View/Helper/FlashHelper.php
```diff
@@ -67,27 +66,16 @@ class FlashHelper extends Helper
      *    Supports the 'params', and 'element' keys that are used in the helper.
      * @return string|null Rendered flash message or null if flash key does not exist
      *   in session.
-     * @throws \UnexpectedValueException If value for flash settings key is not an array.
      */
     public function render(string $key = 'flash', array $options = []): ?string
     {
-        $session = $this->_View->getRequest()->getSession();
```
markstory, line 74: Should the component also be updated to use the new interface?

ADmad: See https://github.com/cakephp/cakephp/pull/15157#issuecomment-727640037 :)

dereuromark:
> has 2 concrete convenience methods success() and error(). Magic call methods don't make for good public APIs. Also I don't thing other types besides these would be used commonly enough. The app skeleton also ships with additional flash elements for only these 2 types

Well, I usually use 4 types in apps:
- error (red)
- success (green)
- warning (orange)
- info (blue)
https://github.com/dereuromark/cakephp-flash/blob/master/src/Controller/Component/FlashComponent.php#L21-L32

I don't want to blow this topic up, but we could then look into further enhancing flashs in core as per my plugin. If there is some interest.
Specifically: 
- Order (instead of mixed types as per entry in stack)
- Transient (only for this request, e.g. useful when working with AJAX)
- AJAX header support

But otherwise we can also keep them plugin specific

As for the method above:

    $this->request->getSession()->getFlash()

The downside of making the flash bag a sub part of session now is going somewhat against the easy to use component level, and what the plugin offers on top.
Session was already quite nice back in the days

    $this->Session->read()/write()

now it is

     $this->getController()->getRequest()->getSession()->...

quite cumbersome

I fear a bit similar things then for flash handling in the controller level.

So I would probably like to keep the easy access of component and helper, and only use FlashMessage library code common code, proxying into that one.

ADmad:
> I usually use 4 types in apps:

We could add the other 2 types,  though the app skeleton would also have to be updated to add templates for the additional types.

> $this->request->getSession()->getFlash()
> The downside of making the flash bag a sub part of session now is going somewhat against the easy to use component level

The method is `$this->request->getFlash()`, it's not a sub part of session, the session instance is injected into `FlashMesage` through its constructor. Also the `FlashMesage` instance is set as a request attribute so one can theoretically even replace it with their own implementation which doesn't use session.

dereuromark:
Sry i was looking only at above examples 
Then it looks good

markstory:
> Earlier I had thought of retro fitting the FlashComponent to use FlashMessage internally but with FlashMessage instance now easily accessing through the request instance I would rather we just deprecate FlashComponent

What do you think about adding the new class now and updating bake to generate code that uses `FlashMessage` for 4.2. If we don't have any issues come up around the new code we can deprecate `FlashComponent` in 4.3?

dereuromark:
Does the component access even need to change here? To keep things simple?
It also a bit more convenient and shorter.

The main aspect was to allow also other points in the framework to add to the flash stack. Not to go away from easy component and helper usage IMO.

ADmad:
I added additional convenience methods and updated `FlashComponent` to use `ServerRequest::getFlash()` internally. We can take the decision to deprecate `FlashComponent` at a later date.

ADmad:
There's a tiny BC break for `FlashComponent` that when using `set()` with exception instance the element is by default set to `error` now. But I think this is a reasonable change and we can just document it in 4.2 migration guide.

markstory (approved):

PR #15161, opened 2020-11-15
Deprecate Text::insert() support for ? placeholders

garas:
Text::insert() `?` placeholders aren't providing any value that sprintf() can't do better.

https://github.com/cakephp/cakephp/pull/14829#issuecomment-662782598

### DIFF
--- a/src/Utility/Text.php
+++ b/src/Utility/Text.php
@@ -203,6 +203,11 @@ class Text
         }
 
         if (strpos($str, '?') !== false && is_numeric(key($data))) {
+            deprecationWarning(
+                'Using Text::insert() with `?` placeholders is deprecated. ' .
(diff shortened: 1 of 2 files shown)


markstory (approved):

src/Utility/Text.php
```diff
@@ -203,6 +203,12 @@ public static function insert(string $str, array $data, array $options = []): st
         }
 
         if (strpos($str, '?') !== false && is_numeric(key($data))) {
+            deprecationWarning(
+                'Using Text::insert() with `?` placeholders is deprecated. ' .
+                'Use sprintf() with `%s` placeholders instead. ' .
+                'Support for `?` placeholders will be removed in 5.0'
```
othercorey, line 209: I don't think we need this line. Just mark it deprecated.

PR #15166, opened 2020-11-16
4.x - Fix binding conflicts with SQL Server when ordering on fields with expressions

ndm2:
So after [**fixing the parentheses problem**](https://github.com/cakephp/cakephp/pull/15164), additional problems became apparent, that is when the expression has bindings they can clash with bindings on the original query (or other subqueries), as the value binder's counter is being reset to `0` when the queries are being compiled, so the manually compiled subquery for the order clause might cause `:c0` to be created, and then a binding on the original query could create another `:c0`. This could cause the wrong values being assigned in the order clause, or possibly an SQL error.

I'm not 100% sure whether/how to fix this. Manually using the original query's value binder will make no difference, as it is being reset when the query is compiled, so it would result in the very same problem. The only thing that came to my mind so far is using order clause expressions (and fixing the missing parentheses in there), so that the subquery expression becomes part of the main query's expression tree. That way the subquery will use the main query's value binder after it has been reset, resulting in the values being bound multiple times with different placeholders instead.

The tests are, well, let's say there's room for improvment, but for now it should just illustrate the problem. I've added an additional integration test (that kinda replicates what the SQL Server driver would do) to see what generally happens in such a situation, and while it seems to work, I'm unsure about possible hidden side effects.

It seems that custom bindings would never work in these expressions, it would cause multiple identically named placedholders to end up in the SQL, and (at least current) SQL Server seems to either require the placeholder names to be unique, or to exactly match the count of bound values.

### DIFF
--- a/tests/TestCase/Database/Driver/SqlserverTest.php
+++ b/tests/TestCase/Database/Driver/SqlserverTest.php
@@ -389,6 +389,50 @@ class SqlserverTest extends TestCase
             ') _cake_paging_ ' .
             'WHERE _cake_paging_._cake_page_rownum_ > 10';
         $this->assertSame($expected, $query->sql());
+
+        $subqueryA = new Query($connection);
+        $subqueryA
+            ->select('count(*)')
+            ->from(['a' => 'articles'])
+            ->where([
+                'a.id = articles.id',
+                'a.published' => 'Y',
+            ]);
+
+        $subqueryB = new Query($connection);
+        $subqueryB
+            ->select('count(*)')
+            ->from(['b' => 'articles'])
+            ->where([
+                'b.id = articles.id',
+                'b.published' => 'N',
+                '1 = :customBinding',
+            ])
+            ->bind(':customBinding', 1, 'integer');
+
+        $query = new Query($connection);
+        $query
+            ->select([
+                'id',
+                'computedA' => $subqueryA,
+                'computedB' => $subqueryB,
+            ])
+            ->from('articles')
+            ->order([
+                'computedA' => 'ASC',
+            ])
+            ->offset(10);
+
+        $this->assertEquals(
+            'SELECT * FROM (' .
+                'SELECT id, ' .
+                '(SELECT count(*) FROM articles a WHERE (a.id = articles.id AND a.published = :c0)) AS computedA, ' .
+                '(SELECT count(*) FROM articles b WHERE (b.id = articles.id AND b.published = :c1 AND 1 = :customBinding)) AS computedB, ' .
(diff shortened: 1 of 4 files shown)


markstory (approved):
These changes seem good to me. I don't think explicitly named parameters are something we should worry about when expressions are being reused.

tests/TestCase/Database/Driver/SqlserverTest.php
```diff
@@ -389,6 +389,48 @@ public function testSelectLimitOldServer()
...
+                'computedA' => $subqueryA,
+                'computedB' => $subqueryB,
+            ])
+            ->from('articles')
+            ->order([
+                'computedA' => 'ASC',
+            ])
+            ->offset(10);
+
+        $this->assertEquals(
```
othercorey, line 424:
```suggestion
        $this->assertSame(
```

tests/TestCase/Database/QueryTest.php
```diff
@@ -4978,4 +4978,170 @@ public function testFetchColumnReturnsFalse()
...
+            'SELECT <id> FROM <articles> ORDER BY \(' .
+                'SELECT \(CASE WHEN <a>\.<published> = \:c0 THEN \:param1 ELSE \:param2 END\) ' .
+                'FROM <articles> <a> ' .
+                'WHERE a\.id = articles\.id' .
+            '\) DESC, <id> ASC',
+            $query->sql(),
+            !$this->autoQuote
+        );
+
+        $this->assertEquals(
```
othercorey, line 5030:
```suggestion
        $this->assertSame(
```

tests/TestCase/Database/QueryTest.php
```diff
@@ -4978,4 +4978,170 @@ public function testFetchColumnReturnsFalse()
...
+                '\(SELECT count\(\*\) FROM <articles> <b> WHERE \(b\.id = articles\.id AND <b>\.<published> = :c1\)\) AS <computedB> ' .
+            'FROM <articles> ' .
+            'ORDER BY \(' .
+                'SELECT count\(\*\) FROM <articles> <b> WHERE \(b\.id = articles\.id AND <b>\.<published> = :c2\)' .
+            '\) DESC, <id> ASC',
+            $query->sql(),
+            !$this->autoQuote
+        );
+
+        $this->assertEquals(
```
othercorey, line 5105:
```suggestion
        $this->assertSame(
```

tests/TestCase/Database/QueryTest.php
```diff
@@ -4978,4 +4978,170 @@ public function testFetchColumnReturnsFalse()
...
+                [
+                    'id' => '2',
+                    'computedA' => '1',
+                    'computedB' => '0',
+                ],
+            ],
+            $query->execute()->fetchAll('assoc')
+        );
+
+        $this->assertEquals(
```
othercorey, line 5126:
```suggestion
        $this->assertSame(
```

PR #15181, opened 2020-11-20
Fix incorrect URL generation

### LINKED ISSUE #15163: Possible Webserver Restriction Bypass
This is a (multiple allowed):

* [x] bug
* [ ] enhancement

* CakePHP Version: 2.x
* Platform and Target: Apache on Linux.

### What you did
While researching Citrix SD-WAN we came across a method to bypass webserver access restriction using how CakePHP2 handles URLs.
The problem is described here: https://medium.com/realmodelabs/sd-pwn-part-2-citrix-sd-wan-center-another-network-takeover-a9c950a1a27c
The relevant line is here: https://github.com/cakephp/cakephp/blob/2.x/lib/Cake/Network/CakeRequest.php#L259

Haven't found similar logic in other versions of CakePHP

### What happened
An endpoint requiring client certificate was accessible without it using CakePHP v2 URI manipulation.

Not being able to alter how the URI is analyzed by CakePHP. I don't really know what's the purpose of this code so it's hard for me to suggest an alternative

P.S. Remember, an issue is not the place to ask questions. You can use [Stack Overflow](https://stackoverflow.com/questions/tagged/cakephp)
for that or join the #cakephp channel on irc.freenode.net, where we will be more
than happy to help answer your questions.

Before you open an issue, please check if a similar issue already exists or has been closed before.

markstory:
When a fullBaseURL is being used we should check for the exact base string being a prefix and not just the string length as it could be full of garbage.

Fixes #15163

### DIFF
--- a/lib/Cake/Test/Case/Network/CakeRequestTest.php
+++ b/lib/Cake/Test/Case/Network/CakeRequestTest.php
@@ -220,7 +220,7 @@ class CakeRequestTest extends CakeTestCase {
 		$request = new CakeRequest();
 		$this->assertEquals('other/path', $request->url);
 
-		$_SERVER['REQUEST_URI'] =  str_repeat('x', strlen($base) - 4) . '://?/other/path';
+		$_SERVER['REQUEST_URI'] = str_repeat('x', strlen($base) - 4) . '://?/other/path';
 		$request = new CakeRequest();
 		$this->assertEquals('', $request->url);
 	}
--- a/lib/Cake/Test/Case/Network/CakeRequestTest.php
+++ b/lib/Cake/Test/Case/Network/CakeRequestTest.php
@@ -215,9 +215,14 @@ class CakeRequestTest extends CakeTestCase {
 		$request = new CakeRequest();
 		$this->assertEquals('some/path', $request->url);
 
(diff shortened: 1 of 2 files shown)


lib/Cake/Test/Case/Network/CakeRequestTest.php
```diff
@@ -215,9 +215,14 @@ public function testQueryStringAndNamedParams() {
...
 
-		$_SERVER['REQUEST_URI'] = Configure::read('App.fullBaseUrl') . '/other/path?url=https://cakephp.org';
+		$base = Configure::read('App.fullBaseUrl');
+		$_SERVER['REQUEST_URI'] = $base . '/other/path?url=https://cakephp.org';
 		$request = new CakeRequest();
 		$this->assertEquals('other/path', $request->url);
+
+		$_SERVER['REQUEST_URI'] =  str_repeat('x', strlen($base) - 4) . '://?/other/path';
+		$request = new CakeRequest();
+		$this->assertEquals('', $request->url);
```
othercorey, line 225: Just reading the code, why does this not equal `/`?
```
if (empty($uri) || $uri === '/' || $uri === '//' || $uri === '/index.php') {
	$uri = '/';
}
```

markstory: Not entirely sure, but I was aiming to make the minimum amount of changes to resolve the problem.

PR #15186, opened 2020-11-23
Improve doc block

bancer:


dereuromark:
This should go into 3.x though, not 3.next.

Issue #15184, opened 2020-11-22
4.2 - Enable mocking/faking of services in test cases


markstory:
This is a (multiple allowed):

* [ ] bug
* [ ] enhancement
* [x] feature-discussion (RFC)

* CakePHP Version: 4.2.0

Testsuites can't easily replace services in the container. Being able to do so is important to realize many of the benefits of dependency injection. In a test case application developers need a simple way to replace or augment services with mocks or fakes.

```php
// Inside a test case.
$this->mockService(StripeService::class, function () {
     return new StripeFake();
});
```

The `mockService` method would register factory methods that would be attached to the container during application startup.

## Container Delegation

In order to append the mocks in a unobtrusive way, we need to make the application container a delegate of the testing container. This will require an addition event to be fired by the application class. The  `Application.buildContainer` event would be fired after building the application's container. Event listeners are able to modify the result of this event with a new container.

By listening to the above event, the test suite can do something like.

```php
function onBuildContainer($event)
{
    if ($this->mockServices) {
        $container = new Container();
        // Simplified
        $container->addMany($this->mockServices);
        $container->delegate($event->getSubject());
        $event->setResult($container);
    }
}
```

I think this will make working with services and the container in tests much easier than it is today.

Related to #14865

ndm2:
I suppose the test could then overwrite the callback in order to gain full control over the new container?

And with delegates, we're going one step closer to tight `league/container` coupling? I know, I'm annoying 🙃 I don't really have a problem with tight coupling as long as it's clearly communicated that this is how it is.

markstory:
> I suppose the test could then overwrite the callback in order to gain full control over the new container?

The callback in Application would be an event, and the test suite could listen to that event via the global listener. The test suite is already using the beforeRender event to gain access to template variables.

> And with delegates, we're going one step closer to tight league/container coupling? I know, I'm annoying upside_down_face I don't really have a problem with tight coupling as long as it's clearly communicated that this is how it is.

I don't know if we're necessarily becoming more coupled. The proposed interface that developers will use doesn't expose the container. How mocked services get into the application should be less important. The container the application registers services into wouldn't have a delegate either, as the delegation would be setup in the event listener which is triggered after the container is built.

ndm2:
> The container the application registers services into wouldn't have a delegate either

Ah, of course, I somehow read that wrong, thinking the inner container would have to provide the delegation API 😴

PR #15191, opened 2020-11-27
4.next - Add test case methods for mocking services

markstory:
Implements the changes described in #15184

I wasn't able to use container delegates because if the top-level dependency isn't in the mocked container, it is forwarded to the delegate, where all dependencies are resolved. This resulted in the mocked dependencies never being used. Instead overwriting/adding mocked services is simpler and works.

I've also refactored the configApplication() and application creation code as we had duplicates and I felt those methods fit well with the new trait. I'm not sold on the new trait name and welcome suggestions for a better name.

### DIFF
--- a/src/Core/ContainerInterface.php
+++ b/src/Core/ContainerInterface.php
@@ -52,4 +52,12 @@ interface ContainerInterface extends PsrInterface
      * @return $this
      */
     public function addServiceProvider($provider);
+
+    /**
+     * Modify an existing definition
+     *
+     * @param string $id The class name or name of the service being modified.
+     * @return \League\Container\Definition\DefinitionInterface
+     */
+    public function extend(string $id): DefinitionInterface;
 }
--- a/src/Http/BaseApplication.php
+++ b/src/Http/BaseApplication.php
@@ -269,6 +269,11 @@ abstract class BaseApplication implements
             $plugin->services($container);
         }
 
+        $event = $this->dispatchEvent('Application.buildContainer', ['container' => $container]);
+        if ($event->getResult() instanceof ContainerInterface) {
+            return $event->getResult();
+        }
+
         return $container;
     }
 
--- a/src/TestSuite/ContainerStubTrait.php
+++ b/src/TestSuite/ContainerStubTrait.php
@@ -1,0 +1,168 @@
+<?php
+declare(strict_types=1);
+
+/**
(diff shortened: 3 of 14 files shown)


src/Core/ContainerInterface.php
```diff
@@ -52,4 +52,12 @@ public function add(string $id, $concrete = null, bool $shared = false): Definit
...
      */
     public function addServiceProvider($provider);
+
+    /**
+     * Modify an existing definition
+     *
+     * @param string $id The class name or name of the service being modified.
+     * @return \League\Container\Definition\DefinitionInterface
+     */
+    public function extend(string $id): DefinitionInterface;
```
othercorey, line 62: Should we rename `$id` to `$name` ?

othercorey: Nevermind. i see the psr interface uses id.

src/TestSuite/ContainerStubTrait.php
```diff
@@ -0,0 +1,168 @@
...
+     * @return null|\Cake\Core\ContainerInterface
+     */
+    public function modifyContainer(EventInterface $event, ContainerInterface $container): ?ContainerInterface
+    {
+        if (empty($this->containerServices)) {
+            return null;
+        }
+        foreach ($this->containerServices as $key => $factory) {
+            if ($container->has($key)) {
+                $container->extend($key)->setConcrete($factory);
```
othercorey, line 146: Does anything implement `extend`? I don't see it anywhere.

src/Http/BaseApplication.php
```diff
@@ -269,6 +269,11 @@ protected function buildContainer(): ContainerInterface
             $plugin->services($container);
         }
 
+        $event = $this->dispatchEvent('Application.buildContainer', ['container' => $container]);
```
ADmad, line 272: We should start adding constants for the event names. Today itself I had made a typo in an event name and has to waste a few precious seconds to realize my mistake and fix it :slightly_smiling_face:.

src/TestSuite/ContainerStubTrait.php
markstory: Yeah it is in League/Container

othercorey: oh, ok, so you're just exposing a League/Container function to the Cake interface that extends the PSR interface.

PR #15196, opened 2020-11-28
Clean up use of custom error handler in tests

othercorey:

### DIFF
--- a/tests/TestCase/Error/ErrorHandlerTest.php
+++ b/tests/TestCase/Error/ErrorHandlerTest.php
@@ -151,8 +151,6 @@ class ErrorHandlerTest extends TestCase
      */
     public function testHandleErrorTraceOffset()
     {
-        $this->_restoreError = true;
-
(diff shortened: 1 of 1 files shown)


tests/TestCase/View/CellTest.php
```diff
@@ -109,13 +109,8 @@ public function testDebugInfo()
...
-            restore_error_handler();
-            $this->assertSame(E_USER_WARNING, $errno);
-            $this->assertStringContainsString('Could not render cell - Cell template file', $msg);
-        };
-        set_error_handler($capture);
-
+        $this->expectWarning();
+        $this->expectWarningMessage('Could not render cell - Cell template file');
         $cell = $this->View->cell('Articles::teaserList');
         $cell->viewBuilder()->setTemplate('nope');
```
dereuromark, line 115: Should we move those below this line? To make sure that only the last line fails?

othercorey: Looks like it doesn't matter. I thought we could throw exceptions now, but guess not.

PR #15199, opened 2020-12-02
Fix optional parameters with default values

markstory:
The container based argument builder was setting these parameters to null before these changes.


othercorey:
Looks ok.

PR #15200, opened 2020-12-02
Rename $callback to $path for methods that take a column/path

othercorey:
I think my first pass of cleaning up collection parameter names was a little too generic. These should primarily be "paths" to the user even though a callback can represent a path. Other function are using $path and it makes sense in the documentation.

Updated any parameter that takes both `string` and `callable`.

### DIFF
--- a/src/Collection/ExtractTrait.php
+++ b/src/Collection/ExtractTrait.php
@@ -29,27 +29,27 @@ trait ExtractTrait
      * Returns a callable that can be used to extract a property or column from
      * an array or object based on a dot separated path.
      *
-     * @param string|callable $callback A dot separated path of column to follow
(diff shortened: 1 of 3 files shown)


src/Collection/ExtractTrait.php
```diff
@@ -29,27 +29,27 @@ trait ExtractTrait
...
+    protected function _propertyExtractor($path): callable
     {
-        if (!is_string($callback)) {
-            return $callback;
+        if (!is_string($path)) {
+            return $path;
         }
 
-        $path = explode('.', $callback);
+        $parts = explode('.', $path);
```
othercorey, line 43: This is renamed to $parts to separate it from the original $path. The parts are an array of path sections.

PR #15203, opened 2020-12-03
Remove Duplicate Virtual Fields in DboSource

pm-michael:
I noticed the `fields` method in DboSource.php duplicates the virtualFields when the `$quote` argument is set to *false*. This simply fixes that issue by removing duplicate keys from the `$virtual` array before the virtual fields are constructed below.


markstory:
Would it be possible to represent what this is changing with a test? With 2.x being so close to end of life, I'm reluctant to merge changes that don't fix testable issues.

markstory:
Closing as 2.10 has reached end of support and will only receive security fixes going forward.

PR #15205, opened 2020-12-04
Cleanup

ADmad:

### DIFF
--- a/src/Database/Driver/Mysql.php
+++ b/src/Database/Driver/Mysql.php
@@ -71,7 +71,7 @@ class Mysql extends Driver
     /**
      * The schema dialect for this driver
      *
-     * @var \Cake\Database\Schema\MysqlSchemaDialect
+     * @var \Cake\Database\Schema\MysqlSchemaDialect|null
      */
     protected $_schemaDialect;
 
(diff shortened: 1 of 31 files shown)


src/Database/Driver/Mysql.php
```diff
@@ -71,7 +71,7 @@ class Mysql extends Driver
     /**
      * The schema dialect for this driver
      *
-     * @var \Cake\Database\Schema\MysqlSchemaDialect
+     * @var \Cake\Database\Schema\MysqlSchemaDialect|null
```
othercorey, line 74: I assume this is required because of the way it's initialized, but doesn't feel like it should be null.

dereuromark: In the end it is important that for the consumer

    public function schemaDialect(): SchemaDialect

is true, which it is.
The change here seems "correct" from the way it is used internally, yeah.

dereuromark (approved):

src/Database/Driver/Mysql.php
ADmad: @othercorey The dialect instance instance isn't created from the constructor so if some method used the property directly without calling `schemaDialect()` it would be still null.

PR #15220, opened 2020-12-13
Allow exception renderer to fallback to error400

othercorey:
When `debug` is true, `ExceptionRenderer` will try to render a custom error template based on the exception name. However, when rendering using TwigView, these custom templates don't exist which throws a `MissingTemplateException`. This causes the generic `error500` to render.

Since custom template engines won't re-implement all of the internal error templates, this allows falling back to the appropriate generic template based on the http exception.

When debug is enabled, users will see the appropriate `error400` page for missing routes.

### DIFF
--- a/src/Error/ExceptionRenderer.php
+++ b/src/Error/ExceptionRenderer.php
@@ -222,6 +222,7 @@ class ExceptionRenderer implements ExceptionRendererInterface
         $code = $this->getHttpCode($exception);
         $method = $this->_method($exception);
         $template = $this->_template($exception, $method, $code);
+        $fallbackTemplate = $this->_fallbackTemplate($exception, $method, $code);
         $this->clearOutput();
 
         if (method_exists($this, $method)) {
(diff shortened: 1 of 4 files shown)


src/Error/ExceptionRenderer.php
```diff
@@ -365,6 +366,19 @@ protected function _template(Throwable $exception, string $method, int $code): s
...
 
+    /**
+     * Gets fallback template for rendering exception info.
+     *
+     * @param \Throwable $exception Exception instance.
+     * @param string $method Method name.
+     * @param int $code Error code.
+     * @return string Template name
+     */
+    protected function _fallbackTemplate(Throwable $exception, string $method, int $code): string
```
othercorey, line 377: This could also be `_missingTemplate` to be more specific about when it is used.

src/Error/ExceptionRenderer.php
```diff
@@ -384,9 +398,10 @@ protected function getHttpCode(Throwable $exception): int
      * Generate the response using the controller object.
      *
      * @param string $template The template to render.
+     * @param string $fallbackTemplate The fallback template to render if $template is missing.
      * @return \Cake\Http\Response A response object that can be sent.
      */
-    protected function _outputMessage(string $template): Response
+    protected function _outputMessage(string $template, string $fallbackTemplate = 'error500'): Response
```
othercorey, line 404: Provide an explicit default for easier migrations since custom exception renderers might not know what to do for this.

PR #15221, opened 2020-12-15
Throw exception on invalid path for groupBy and indexBy

othercorey:
When an invalid path is passed to `groupBy` and `indexBy`, php silently converts the `null` value returned by `_propertyExtractor` to an empty string.

This causes all mistakes to be ignored and data-loss in the case of `indexBy`.

Users can provide defaults by passing a callback, but the default implementation should throw an error since missing paths are typically not intentional. Unfortunately, there is no distinction between a null value that exists and an invalid path. Users expecting to group or index by `null` probably want to define valid default behavior anyway.

### DIFF
--- a/src/Collection/CollectionTrait.php
+++ b/src/Collection/CollectionTrait.php
@@ -267,7 +267,14 @@ trait CollectionTrait
         $callback = $this->_propertyExtractor($path);
         $group = [];
         foreach ($this->optimizeUnwrap() as $value) {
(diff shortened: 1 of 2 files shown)


markstory (commented):
I like this change. We'll need to note it in the 4.2 migration guide as it *could* break applications. Those applications would have been doing the 'wrong' thing before but it was working and now won't be.

Issue #15223, opened 2020-12-16
Unique field allow to be NULL


davidspeijer:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 4.1.0
* Platform and Target: CentOS + MariaDB

### What you did
I have a Users table;

```
CREATE TABLE `users` (
  `id` int(11) UNSIGNED AUTO_INCREMENT PRIMARY KEY,
  `firstname` varchar(50) DEFAULT NULL,
  `middlename` varchar(50) DEFAULT NULL,
  `lastname` varchar(50) NOT NULL,
  `email` varchar(255) NOT NULL,
  `password` varchar(255) NOT NULL,
  `lastlogin` datetime DEFAULT NULL,
  `published` BOOLEAN NOT NULL DEFAULT FALSE,
  `token` varchar(40) DEFAULT NULL,
  `locale` varchar(5) NOT NULL DEFAULT 'nl_NL',
  `superadmin` BOOLEAN NOT NULL DEFAULT FALSE,
  `created` datetime DEFAULT NULL,
  `modified` datetime DEFAULT NULL,
  UNIQUE KEY (email),
  UNIQUE KEY (token)
);
```

with validator
```
        $validator
            ->scalar('token')
            ->maxLength('token', 40)
            ->allowEmptyString('token')
            ->add('token', 'unique', ['rule' => 'validateUnique', 'provider' => 'table']);
```

and build rules
```
    public function buildRules(RulesChecker $rules): RulesChecker
    {
        $rules->add($rules->isUnique(['email']), ['errorField' => 'email']);
        $rules->add($rules->isUnique(['token']), ['errorField' => 'token']);
        $rules->add($rules->existsIn(['company_id'], 'Companies'), ['errorField' => 'company_id']);

        return $rules;
    }
```

### What happened
when I try to reset the token to NULL and save the user to the database 

```
            $user = $this->Users->patchEntity($user, $this->request->getData());
            $user = $this->Users->patchEntity($user, ['token' => null]);
            //print_r($user);
            
            

            if ($this->Users->save($user)) {
                $this->getMailer('User')->send('resetPassword', [$user]);
                $this->Flash->success(__('Your password new password is saved.'));
                //return $this->redirect(['controller'=>'Users','action' => 'login']);
            } else {
                print_r($user);
            }
```

I get the error;

```
[[errors]] => Array
        (
            [token] => Array
                (
[...]

Issue #15198, opened 2020-12-01
Cake4X isUnique() allowMultipleNull expected


Agacor:
* [ ] enhancement

* CakePHP Version:4x

### What you did
Saving an entity with RulesChecker isUnique(['key_a', 'key_b']) with multiple keys and some of them with null value.

### What happened
There is no option like Cake3X for allowMultipleNull.

### What you expected to happen
I was expected to have $options for the isUnique() Rule as Cake3. I think this could be forgotten in the migration to Cake4.

markstory:
The allowMultipleNull option was removed in #13214

markstory:
The option was deemed redundant because of improvements in how null conditions are handled. See https://github.com/cakephp/cakephp/pull/13214#discussion_r311835830 for the related discussion. It is also possible that our test suite did not cover the required scenarios well enough.

PR #15226, opened 2020-12-18
Add allowMultipleNulls to IsUnique rule check

othercorey:
Closes https://github.com/cakephp/cakephp/issues/15223
Closes https://github.com/cakephp/cakephp/issues/15198

This allows the `IsUnique` rule check to behave the same as UNIQUE index for MySQL, PostgreSQL and SQLite. The default behavior is BC and users will have to enable the option.

This was originally in 3.x and enabled by default but implemented differently.

The previous tests seem to be left over from previous code that was removed. Updated to test both scenarios.

### DIFF
--- a/src/ORM/Rule/IsUnique.php
+++ b/src/ORM/Rule/IsUnique.php
@@ -31,13 +31,28 @@ class IsUnique
     protected $_fields;
 
     /**
+     * The unique check options
+     *
+     * @var array
+     */
+    protected $_options = [
+        'allowMultipleNulls' => false,
+    ];
+
+    /**
      * Constructor.
      *
+     * ### Options
+     *
+     * - `allowMultipleNulls` Allows any field to have multiple null values. Defaults to false.
+     *
      * @param string[] $fields The list of fields to check uniqueness for
+     * @param array $options The options for unique checks.
      */
-    public function __construct(array $fields)
+    public function __construct(array $fields, array $options = [])
     {
         $this->_fields = $fields;
+        $this->_options = $options + $this->_options;
     }
 
     /**
@@ -54,8 +69,13 @@ class IsUnique
             return true;
         }
 
+        $fields = $entity->extract($this->_fields);
+        if ($this->_options['allowMultipleNulls'] && array_filter($fields, 'is_null')) {
+            return true;
+        }
+
         $alias = $options['repository']->getAlias();
-        $conditions = $this->_alias($alias, $entity->extract($this->_fields));
+        $conditions = $this->_alias($alias, $fields);
         if ($entity->isNew() === false) {
             $keys = (array)$options['repository']->getPrimaryKey();
             $keys = $this->_alias($alias, $entity->extract($keys));
--- a/tests/Fixture/UniqueAuthorsFixture.php
+++ b/tests/Fixture/UniqueAuthorsFixture.php
@@ -1,0 +1,47 @@
+<?php
+/**
+ * CakePHP(tm) : Rapid Development Framework (https://cakephp.org)
+ * Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
+ *
+ * Licensed under The MIT License
+ * For full copyright and license information, please see the LICENSE.txt
+ * Redistributions of files must retain the above copyright notice
+ *
+ * @copyright     Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
+ * @link          https://cakephp.org CakePHP(tm) Project
+ * @since         4.2.0
+ * @license       https://opensource.org/licenses/mit-license.php MIT License
+ */
+namespace Cake\Test\Fixture;
+
+use Cake\TestSuite\Fixture\TestFixture;
+
+/**
+ * Tables of unique author ids
+ */
+class UniqueAuthorsFixture extends TestFixture
+{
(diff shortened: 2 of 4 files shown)


src/ORM/Rule/IsUnique.php
```diff
@@ -30,14 +30,23 @@ class IsUnique
...
      * Constructor.
      *
      * @param string[] $fields The list of fields to check uniqueness for
+     * @param array $options The options for unique checks.
      */
-    public function __construct(array $fields)
+    public function __construct(array $fields, array $options = [])
     {
         $this->_fields = $fields;
+        $this->_options = $options + ['allowUniqueNulls' => false];
```
ADmad, line 49:
```suggestion
        $this->_options = $options + $this->_options;
```

src/ORM/Rule/IsUnique.php
```diff
@@ -30,14 +30,23 @@ class IsUnique
      */
     protected $_fields;
 
+    /**
+     * The unique check options
+     *
+     * @var array
+     */
+    protected $_options;
```
ADmad, line 38:
```suggestion
    protected $_options = ['allowUniqueNulls' => false];
```

ADmad:
> The default behavior is BC and users will have to enable the option.

We can update `bake` instead to generate the rule with `'allowUniqueNulls' => true`.

src/ORM/Rule/IsUnique.php
markstory: It would be good to have a blurb here on what options are supported and what they do.

markstory (commented):
Should we use the option name that was used in 3.x?

tests/Fixture/UniqueAuthorsFixture.php
```diff
@@ -0,0 +1,47 @@
...
+    ];
+
+    /**
+     * records property
+     *
+     * @var array
+     */
+    public $records = [
+        ['first_author_id' => null, 'second_author_id' => 1],
+    ];
```
saeideng, line 46: could we have 3 fields instead of  2 for testing ?

othercorey: Do you want the 3rd field to be NOT NULL?

othercorey:
> Should we use the option name that was used in 3.x?

So, we could. Looking at the 3.x code, it relied on invalid SQL behavior to perform the check. It would generate invalid sql on purpose so the check would fail with `null` values:
```sql
select * from table where first_column = null and second_column = 'actual value';
```

The new code skips the check if there is a `null` value in the column set as any `null` value generates a unique record.

However, the old `allowMultipleNulls` was enabled by default while 4.0 and 4.1 has this behavior removed so it was essentially disabled by default.

tests/Fixture/UniqueAuthorsFixture.php
saeideng: I want to test all these cases 

1)

| field_1| field_2| field_3 |  -
| ---| ---| ---| ---
| null| 1 | 1|  
 | null| 1 | 1| new record

2)

| field_1| field_2| field_3 |  -
| ---| ---| ---| ---
| null| 1 | 1|  
| null| 1 | 2| new record

3)

| field_1| field_2| field_3 |  -
| ---| ---| ---| ---
| null| 1 | 1|  
| null| null | 1| new record

4)

| field_1| field_2| field_3 |  -
| ---| ---| ---| ---
| null| 1 | 1|  
| null| null | 2| new record
| null| null | 2| new record

so when there are 2 `NULL` value , and when 2 `NOT NULL` value 
especially I want to see result of case 3 and 4 
you can add ` null| 1 | 1` in `Fixture`

saeideng: >The new code skips the check if there is a null value in the column set 

this is important for case 3

othercorey: I don't see the difference really. Why is 3 columns different than 2?

tests/TestCase/ORM/RulesCheckerIntegrationTest.php
```diff
@@ -394,86 +396,58 @@ public function testIsUniqueMultipleFields()
...
-
-        $entity = $table->get(1);
-        $entity->setDirty('author_id', true);
-        $this->assertSame($entity, $table->save($entity));
+        $this->assertEquals(['first_author_id' => ['_isUnique' => 'This value is already in use']], $entity->getErrors());
     }
 
     /**
-     * Tests isUnique with multiple fields and allowMultipleNulls
+     * Tests isUnique with allowMultipleNulles
```
markstory, line 420:
```suggestion
     * Tests isUnique with allowMultipleNulls
```

VARGA-Peter:
Hi guys, first thank you very much that this option has been re-introduced. Is there a way to set it for the whole project to `TRUE`?

_If not, could you consider to offer something like this:_
In `app_local.php` in the `Datasources` section it can be pre-set for the whole project. If it isn't set there so it defaults to `FALSE`  - or something similar.

This wouldn't break any compatibility. Thx anyway!

othercorey:
Please open a new issue.

VARGA-Peter:
> Please open a new issue.

https://github.com/cakephp/cakephp/issues/15734 - Thank you!

PR #15227, opened 2020-12-18
Fix replacing belongstomany junction records with complex types

### LINKED ISSUE #15212: BelongsToMany not "replace" associated when foreignKeys are binaryuuid
This is a (multiple allowed):

* [x] bug
* [ ] enhancement

* CakePHP Version: 4.1.6
* Platform and Target: MAMP with MySQL.

### What you did
I'm trying to update belongstomany relations with saveStrategy "replace".
When i create new one everything goes well but when i try to update (ie: removing one related and leaving others) nothing happen.
In my PostsTable.php:
```
$this->belongsToMany('Therapies', [
      'foreignKey' => 'post_id',
      'targetForeignKey' => 'therapy_id',
      'joinTable' => 'posts_therapies',
    ]);
```
MySQL tables:
```
CREATE TABLE `posts` (
  `id` binary(16) NOT NULL DEFAULT '\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0',
......,
  PRIMARY KEY (`id`)
)

CREATE TABLE `posts_therapies` (
  `id` int(10) unsigned NOT NULL AUTO_INCREMENT,
  `post_id` binary(16) NOT NULL DEFAULT '\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0',
  `therapy_id` tinyint(3) unsigned NOT NULL,
  PRIMARY KEY (`id`)
)
```
PostsController.php:
```
public function edit($id = null)
  {
    $post = $this->Posts->get($id, ['contain' => ['Therapies']]);
    if ($this->request->is(['patch', 'post', 'put'])) {
      $post = $this->Posts->patchEntity($post, $this->request->getData(), ['associated' => [ 'Therapies']]);
      if ($this->Posts->save($post)) {
        $this->Flash->success(__('The post has been saved.'));
      } else {
        $this->Flash->error(__('The post could not be saved. Please, try again.'));
      }
    }
    $therapies = $this->Posts->Therapies->find('list');
    $this->set(compact('post', 'therapies'));
  }
```

### What happened
Nothing happen when i try to update.
I've made some test without uuid as primary key and everything goes well, so i think is a matter of uuid.

When i update a record with belongstomany related must unlink existing relations and create new ones.

markstory:
When the ORM was attempting to clear the junction records for a BelongsToMany junction table there were a few related problems causing it to fail:

1. The query used to locate the junction records did not have the   correct types set for where conditions.
2. If the junction table has a surrogate primary key that has    a different datatype than the association target table, the ORM    would use the target table type. In the committed fixtures this means    that the `id` column on the junction table would be typed as    `binaryuuid` instead of `integer`.

These factors combined together would result in either junction rows not being deleted. Thankfully the query to find junction records would always fail preventing more serious data corruption issues.

These problems can be solved by applying conditions *after* adding the junction join information and type mappings, and also setting junction table types for our aliased entity query.

I optimized object traversal inside the `_appendJunctionJoin` method as there were some redundant calls being ade.

Fixes #15212

### DIFF
--- a/tests/Fixture/BinaryUuiditemsBinaryUuidtagsFixture.php
+++ b/tests/Fixture/BinaryUuiditemsBinaryUuidtagsFixture.php
@@ -1,0 +1,48 @@
+<?php
+/**
+ * CakePHP(tm) : Rapid Development Framework (https://cakephp.org)
+ * Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
+ *
+ * Licensed under The MIT License
+ * For full copyright and license information, please see the LICENSE.txt
+ * Redistributions of files must retain the above copyright notice
+ *
+ * @copyright     Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
+ * @link          https://cakephp.org CakePHP(tm) Project
+ * @since         4.1.8
+ * @license       https://opensource.org/licenses/mit-license.php MIT License
+ */
+namespace Cake\Test\Fixture;
+
+use Cake\TestSuite\Fixture\TestFixture;
+
+/**
+ * BinaryUuiditemsBinaryUuidtagsFixture
+ */
+class BinaryUuiditemsBinaryUuidtagsFixture extends TestFixture
+{
+    /**
+     * fields property
+     *
+     * @var array
+     */
+    public $fields = [
+        'id' => ['type' => 'integer'],
+        'binary_uuiditem_id' => ['type' => 'binaryuuid', 'null' => false],
+        'binary_uuidtag_id' => ['type' => 'binaryuuid', 'null' => false],
+        '_constraints' => [
+            'primary' => ['type' => 'primary', 'columns' => ['id']],
+            'unique_item_tag' => [
+                'type' => 'unique',
+                'columns' => ['binary_uuiditem_id', 'binary_uuidtag_id']
+            ]
+        ],
+    ];
+
+    /**
+     * records property
+     *
+     * @var array
+     */
+    public $records = [];
+}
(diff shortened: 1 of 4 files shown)


tests/Fixture/BinaryUuiditemsBinaryUuidtagsFixture.php
```diff
@@ -0,0 +1,48 @@
...
+    public $fields = [
+        'id' => ['type' => 'integer'],
+        'binary_uuiditem_id' => ['type' => 'binaryuuid', 'null' => false],
+        'binary_uuidtag_id' => ['type' => 'binaryuuid', 'null' => false],
+        '_constraints' => [
+            'primary' => ['type' => 'primary', 'columns' => ['id']],
+            'unique_item_tag' => [
+                'type' => 'unique',
+                'columns' => ['binary_uuiditem_id', 'binary_uuidtag_id']
+            ]
```
othercorey, line 38:
```suggestion
                'columns' => ['binary_uuiditem_id', 'binary_uuidtag_id'],
            ],
```

ADmad:
The failure for SQL Server seems related to these changes.

tests/TestCase/ORM/Association/BelongsToManyTest.php
```diff
@@ -995,6 +1003,47 @@ public function testReplaceLinkFailingDomainRules()
...
+     * Tests that replaceLinks will delete entities not present in the passed,
+     * array, maintain those are already persisted and were passed and also
+     * insert the rest.
+     *
+     * @return void
+     */
+    public function testReplaceLinkBinaryUuid()
+    {
+        $this->skipIf(
+            $this->connection->getDriver() instanceof \Cake\Database\Driver\Sqlserver,
```
othercorey, line 1016:
```suggestion
            ConnectionManager::get('test')->getDriver() instanceof \Cake\Database\Driver\Sqlserver,
```

PR #15229, opened 2020-12-21
Fix case

remicollet:
To match https://github.com/cakephp/cakephp/blob/master/tests/Fixture/BinaryUuidItemsFixture.php

### DIFF
--- a/tests/TestCase/ORM/Association/BelongsToManyTest.php
+++ b/tests/TestCase/ORM/Association/BelongsToManyTest.php
@@ -48,7 +48,7 @@ class BelongsToManyTest extends TestCase
         'core.SpecialTags',
         'core.ArticlesTags',
         'core.Tags',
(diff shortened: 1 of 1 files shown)


othercorey:
I'm not sure why just this fixture was changed previously, but many more need to be changed to align it if we want to have `binary_uuid_items` table name.

remicollet:
@othercorey btw, you see the problem, classe name case are inconsistent in code, raising issue with Psr4 autoloader (class not found)

dereuromark:
Yes, all fixture names should have been cleaned up in 4.2.
We should probably fix up the rest as well now.

dereuromark:
Ah, from the looks only this one file has a psr issue
https://github.com/cakephp/cakephp/tree/master/tests/Fixture

The others can remain in the former naming for the time being for BC reasons now.

So all good

Issue #15234, opened 2020-12-22
4.2.0 - Foreign key constraint error at fixtures creation when not loading fixtures for all referenced tables


liqueurdetoile:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 4.2.0
* Platform and Target: Mysql 4.7 with innodb tables

### What you did
For speed sake, I'm used to load only relevant fixtures in my unit tests. Therefore, some associated tables are not created if no data is needed from them. It can make a huge boost difference with central models like `users` that can be referenced in several other tables.

### What happened
When upgrading to 4.2.0, tests start failing with `SQLSTATE[HY000]: General error: 1215 Cannot add foreign key constraint error.` When adding the "missing" fixtures, it works. Reverting to 4.1.7 also solved the issue.

### What you expected to happen
No foreign key constraint error for unused fixtures.

othercorey:
We need more information on the fixtures you're loading that cause this error.

liqueurdetoile:
The schema is simple : 
- Users has many Deployments (and reverse belongsTo)
- Creleases has many Deployments (and reverse belongsTo)
- Sreleases has many Deployments (and reverse belongsTo)

Fixtures are all baked from db and Deployments references user_id, crelease_id and srelease_id as FK :
```
public $fields = [
        'id' => ['type' => 'uuid', 'length' => null, 'null' => false, 'default' => null, 'comment' => '', 'precision' => null],
        'user_id' => ['type' => 'uuid', 'length' => null, 'null' => false, 'default' => null, 'comment' => '', 'precision' => null],
        'crelease_id' => ['type' => 'uuid', 'length' => null, 'null' => true, 'default' => null, 'comment' => '', 'precision' => null],
        'srelease_id' => ['type' => 'uuid', 'length' => null, 'null' => true, 'default' => null, 'comment' => '', 'precision' => null],
        'cconstraint' => ['type' => 'string', 'length' => 45, 'null' => false, 'default' => null, 'comment' => '', 'precision' => null],
        'sconstraint' => ['type' => 'string', 'length' => 45, 'null' => false, 'default' => null, 'comment' => '', 'precision' => null],
        'domain' => ['type' => 'string', 'length' => 100, 'null' => false, 'default' => null, 'collate' => 'utf8mb4_general_ci', 'comment' => '', 'precision' => null],
[...]

PR #15239, opened 2020-12-23
Fix(FixtureManager): Add back FK checks disabling for fixtures loading

liqueurdetoile:
The last rework of FixtureManager introduces a breaking change as described here : #15234 

As stated in #4461, FK checks have been disabled for fixtures operations to ease tests writing. It has been maintained for data insertion but not for FK constraints creation. Therefore, one must load all referenced tables even if their data is not used in the test.

This PR proposes to restore previous behavior.

### DIFF
--- a/src/Database/Driver/Sqlite.php
+++ b/src/Database/Driver/Sqlite.php
@@ -59,9 +59,9 @@ class Sqlite extends Driver
     /**
      * Stores the initial state of foreign key support in sqlite db
      *
-     * @var bool
+     * @var bool|null
      */
-    protected $_isFKEnabled = false;
+    protected $_isFKEnabled = null;
 
     /**
      * The schema dialect class for this driver
@@ -146,11 +146,6 @@ class Sqlite extends Driver
             }
         }
 
-        // Fetch the current state of FK support
-        $statement = $this->getConnection()->query('PRAGMA foreign_keys');
-        $statement->execute();
-        $this->_isFKEnabled = (bool)$statement->fetch()['foreign_keys'];
-
         return true;
     }
 
@@ -193,6 +188,13 @@ class Sqlite extends Driver
      */
     public function disableForeignKeySQL(): string
     {
+        // Fetch the current state of FK support
+        if ($this->_isFKEnabled === null) {
+            $statement = $this->getConnection()->query('PRAGMA foreign_keys');
+            $statement->execute();
+            $this->_isFKEnabled = (bool)$statement->fetch()['foreign_keys'];
+        }
+
         return 'PRAGMA foreign_keys = OFF';
     }
 
--- a/src/Database/Driver/Sqlite.php
+++ b/src/Database/Driver/Sqlite.php
@@ -149,7 +149,7 @@ class Sqlite extends Driver
         // Fetch the current state of FK support
         $statement = $this->getConnection()->query('PRAGMA foreign_keys');
         $statement->execute();
-        $this->_isFKEnabled = (bool) $statement->fetch()['foreign_keys'];
+        $this->_isFKEnabled = (bool)$statement->fetch()['foreign_keys'];
 
         return true;
     }
--- a/src/Database/Driver/Sqlite.php
+++ b/src/Database/Driver/Sqlite.php
@@ -206,6 +206,7 @@ class Sqlite extends Driver
         if ($this->_isFKEnabled === null) {
             $this->_isFKEnabled = $this->isForeignKeyEnabled();
         }
+        
         return 'PRAGMA foreign_keys = OFF';
     }
 
--- a/src/Database/Driver/Sqlite.php
+++ b/src/Database/Driver/Sqlite.php
@@ -206,7 +206,7 @@ class Sqlite extends Driver
         if ($this->_isFKEnabled === null) {
             $this->_isFKEnabled = $this->isForeignKeyEnabled();
         }
-        
+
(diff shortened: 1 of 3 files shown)


src/Database/Driver/Sqlite.php
```diff
@@ -176,11 +183,30 @@ public function prepare($query): StatementInterface
         return $result;
     }
 
+    /**
+     * Query SQLite DB for current status of foreign keys support
+     *
+     * @return  bool      True if foreign key support is currently enabled
+     */
+    public function isForeignKeyEnabled(): bool
```
markstory, line 191: We don't generally add new public methods in bug fix releases. This would need to be a protected method or moved to 4.3.

markstory (commented):
While this would restore the behavior that was present in the past I am not sure that behavior was intentional or even worked in all the database drivers. Mysql is a bit of an oddity in that you can bypass foreign keys and insert otherwise invalid data. If you frequently rely on bypassing foreign keys have you considered not including them in your fixture schema?

src/Database/Driver/Sqlite.php
liqueurdetoile: I understand. It can easily be turned into a protected method as it was mainly made public for test purpose. The return value of `SqliteDriver::enableForeignKeySQL` is enough to check for validity.

ADmad:
> If you frequently rely on bypassing foreign keys have you considered not including them in your fixture schema?

Indeed. It really doesn't make much sense adding foreign key constraints to your fixtures if you are going to disable them.

liqueurdetoile:
@markstory The goal is not to permit to load invalid data but to avoid loading fixtures that are not relevant in the test.

A use case :
- Users has many Deployments
- Deployments are dependent from Users.
- Deployments belongs to Users and 2 other tables.
- Each Deployment has a string field `domain`.
- Deployments fixture data is set up carefully so FK values are really matching data from the parents table fixtures (that's the most boring thing to do I must say).
- I have 10 tests for UsersTable : 7 on user own data, 1 is for User removal, 2 for a custom finder on Users that reduces a User Deployments collection to a collection of strings from domains.

Before 4.2.0, I only have to load ['App.users', 'App.deployments'] for all tests to run for UsersTable. That sounds perfectly logic to me as only Users and Deployments schema and data are needed in my testcase. If I've planned to create or update into Deployments from the UsersTable testcase (not logical but feasible), the whole 4 fixtures should have been loaded to avoid an error.

From 4.2.0, the loading for `App.deployments` fixture fails at **foreign key constraint creation** because the two other referenced tables schema are missing in db. To solve it, I must load 2 perfectly useless additional fixtures only to have the tables created in database for keys that will not be used. It means rewriting fixtures list for many tests and that's why I'm considering it quite like a breaking change. Moreover, there's some overhead in creating tables only to avoid error at **constraints creation** and **truncate** steps.

I can't simply remove FK from Deployments schema as I'm also testing DeploymentsTable in another testcase and needs them from there (this test case obviously use the four fixtures). Moreover, it means that I have to tweak fixtures schema after baking it from database.

Finally, in 4.2.0, constraints are still disabled at **data insertion step** and therefore, you can still insert silly foreign keys values that leads to nowhere.

I think this PR is not to revive bad habits but let optimize fixtures to focus only on required tables for a given test and give more accuracy in handling Sqlite specificity for foreign keys.

I know my poor dev Mysql Server makes me more sensitive to fixtures optimization but I hope my explanation makes sense.

Merry Christmas to all 🎅

liqueurdetoile:
I've spent some time to dive deeper into the different databases handling with fixtures.

I've set up a [quick repository](https://github.com/liqueurdetoile/cakephp-fixtures-bench) with nearly the same workflow as Cakephp to test and benchmark fixtures optimization. See readme for scenarios explanation.

The results are available here : https://lite.framacalc.org/9kv2-cakephp-fixtures-benchmark. Globally, optimizing fixtures is definitely a good idea to speed up a test suite. But...

To do this, I quickly came into the limitations of each dbs to handle fixtures setup :
- SQLite have foreign keys disabled as default and constraints are set up at table creation. Any attempt to wrap an operation in `disableConstraints` add back foreign keys and SQLite fails at data insertion.
- Postgre `disableConstraints` is only based on a FK DEFERRED statement and not with a ALTER TABLE xxx DISABLE TRIGGER
- Mysql, MariaDB and Postgre complains when trying to truncate a table with FK are enabled
- ....

At this point, I've started to think about a way to make optimized fixtures available for all core databases. It's feasible by avoiding applying constraints for unloaded tables (but also in tests ?) and/or adding fixture setup specific operations for each one.

I've made a really quick and dirty implementation of the idea in the benchmark repository. I've stopped when it cames to work out a fixtures setup through dedicated database mappers with this question : Is it really useful to spend much time on this ?

I then looked at alternate systems like `cakephp-fixture-factories`. They turned over the aforementioned limitations by handling test database and fixtures in a totally different way. There is many interesting ideas, especially migrations use and generators for fake data, but I like to spot a globally coherent fixture dataset available in all tests (and I'm not using migrations in all my projects 🙄).

So, what to do next ? Is this PR finally acceptable (after casting `isForeignKeyEnabled` as protected method) ? Should I also propose an upgraded fixtures manager that go further in optimizing fixtures loading ? A plugin ?

Many questions that I was willing to share here as I certainly miss some points.

othercorey:
Closed in https://github.com/cakephp/cakephp/pull/15253. Reverting fixtures changes in 4.2.

PR #15241, opened 2020-12-24
Fix extra null parameters in controller actions

### LINKED ISSUE #15236: 4.2.0 Non empty action params
This is a (multiple allowed):

* [x] bug
* [ ] enhancement

* CakePHP Version: 4.2.0 | 4.2.1 | master
* Platform and Target: Ubuntu 20.04LTS

### What you did
```
<?php

namespace App\Controller;

class FooController extends AppController
{
    public function bar(...$baz)
    {
        var_dump($baz);
    }
}
```

### What happened

null value in key 0

```
src/Controller/FooController.php:9:
array(1) {
  [0] =>
  NULL
}
```

Expected same as previous versions of cake.  
ex: CakePHP 4.1.7:
```
src/Controller/FooController.php:9:
array(0) {
}
```

markstory:
When a controller has a required parameter or a spread parameter we should not be appending a null.

Fixes #15236

### DIFF
--- a/src/Controller/ControllerFactory.php
+++ b/src/Controller/ControllerFactory.php
@@ -104,12 +104,17 @@ class ControllerFactory implements ControllerFactoryInterface
         $passed = array_values((array)$controller->getRequest()->getParam('pass'));
         foreach ($reflection->getParameters() as $i => $parameter) {
             $position = $parameter->getPosition();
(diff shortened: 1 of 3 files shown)


othercorey:
Looks ok.

PR #15242, opened 2020-12-24
Add salted tokens to SessionCsrfProtectionMiddleware

markstory:
Our current tokens are consistent across the entire session. This can allow CSRF tokens to be exfiltrated via a BREACH vulnerability. Salting the token will ensure that it is always different can thus cannot be reliabily extracted via BREACH.

I have this targeting master the session CSRF is new in 4.2 and I think the adoption rate will be very low right now. Furthermore, there are tests ensuring that unsalted tokens are still accepted. If this pull request is good, I'll also add salting to the stateless CSRF tokens.

### DIFF
--- a/src/Http/Middleware/SessionCsrfProtectionMiddleware.php
+++ b/src/Http/Middleware/SessionCsrfProtectionMiddleware.php
@@ -167,13 +167,14 @@ class SessionCsrfProtectionMiddleware implements MiddlewareInterface
             // XOR the token and salt together so that we can reverse it later.
             $salted .= chr(ord($token[$i]) ^ ord($salt[$i]));
         }
+
         return $salted . $salt;
     }
 
     /**
      * Remove the salt from a CSRF token.
      *
-     * If the token is not TOKEN_VALUE_LENGTH * 2 it is an old 
+     * If the token is not TOKEN_VALUE_LENGTH * 2 it is an old
      * unsalted value that is supported for backwards compatibility.
      *
      * @param string $token The token that could be salty.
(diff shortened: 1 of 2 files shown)


src/Http/Middleware/SessionCsrfProtectionMiddleware.php
```diff
@@ -148,6 +148,55 @@ public function skipCheckCallback(callable $callback)
...
+     * @return string The salted token with the salt appended.
+     */
+    public function saltToken(string $token): string
+    {
+        $length = strlen($token);
+        $salt = Security::randomString($length);
+        $salted = '';
+        for ($i = 0; $i < $length; $i++) {
+            // XOR the token and salt together so that we can reverse it later.
+            $salted .= chr(ord($token[$i]) ^ ord($salt[$i]));
```
othercorey, line 168: Don't want to encrypt it with the salt?

garas: Result of XOR will be likely codepoint of binary character, so `chr()` will make binary string. Binary strings are unwanted in HTTP headers or forms.

`$token` and `$salt` are hex strings, so they should be hex unencoded first, then hex re-encoded after XOR operation.

So `unsaltToken()` should also check if token is valid hexadecimal string before hex decoding.

markstory: Good point about the binary data being bad in HTTP contexts. I'll make the suggested changes.

markstory: I switched to using base64 encoding as it was simpler to use and more compact than hex encoding is.

PR #15245, opened 2020-12-25
Unskip test for SQLServer and BelongsToMany

markstory:


tests/TestCase/ORM/Association/BelongsToManyTest.php
```diff
@@ -1027,10 +1023,11 @@ public function testReplaceLinkBinaryUuid()
         ]);
         $itemName = 'Item 1';
         $item = $items->find()->where(['BinaryUuidItems.name' => $itemName])->firstOrFail();
+        $existingTag = $tags->find()->where(['BinaryUuidTags.name' => 'Defect'])->firstOrFail();
 
         // 1=existing, 2=new tag
         $item->binary_uuid_tags = [
-            new Entity(['id' => '481fc6d0-b920-43e0-a40d-111111111111'], ['markNew' => false]),
+            $existingTag,
```
othercorey, line 1030: This looks ok, but why was the previous code passing for other dbs?

markstory: SQLServer stores UUID values as upper case bytes...

othercorey: So why does the fixture insert work but the entity lookup doesn't? Is it a problem with BinaryUuidType?

markstory: I haven't tracked down why the casing was different. I didn't see any case changing in the binary uuid type or in the sqlserver driver.

PR #15248, opened 2020-12-26
Make Router::routes() return a sorted routes list.

ADmad:
Users wouldn't be aware that the connected routes are sorted based on paths first when parsing request URL. Having the `RoutesCommand` show the routes list in the order that they will be matched would better help debugging routing issues.

I can update the tests once there is a consensus for the change. A few tests pluck the routes using specific index from the list returned by `Router::routes()` so those indexes will need to be updated.

This change was suggested by @ndm2 on Slack :slightly_smiling_face:.

> It might be helpful if the routes command would output the routes in the same order as they are being matched, seeing routes working even though the shown order suggested that it wouldn't work, was quite confusing.

### DIFF
--- a/src/Routing/RouteCollection.php
+++ b/src/Routing/RouteCollection.php
@@ -40,13 +40,6 @@ class RouteCollection
     protected $_routeTable = [];
 
     /**
-     * The routes connected to this collection.
-     *
-     * @var \Cake\Routing\Route\Route[]
-     */
-    protected $_routes = [];
-
-    /**
      * The hash map of named routes that are in this collection.
      *
      * @var \Cake\Routing\Route\Route[]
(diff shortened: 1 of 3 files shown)


src/Routing/RouteCollection.php
```diff
@@ -39,13 +39,6 @@ class RouteCollection
      */
     protected $_routeTable = [];
 
-    /**
-     * The routes connected to this collection.
-     *
-     * @var \Cake\Routing\Route\Route[]
-     */
-    protected $_routes = [];
```
markstory, line 47: Not having two lists is a nice win.

markstory (commented):
Could the sorting be done in the route command if that is the problem we are trying to solve? Changing the order in Router seems higher risk than doing the sorting in the command.

ADmad:
> Changing the order in Router seems higher risk than doing the sorting in the command.

I don't see the risk. The routes are sorted anyway before parsing request URL in `parseRequest()`. The `match()` method uses routes list maintained in `$_routeTable` property which is unchanged. `routes()` method is only used to check if routes are loaded and to get list for display routes. So IMO this change is safe enough.

markstory (approved):

PR #15257, opened 2020-12-30
Fix incorrect type checks in ErrorHandlerMiddleware

### LINKED ISSUE #15254: Custom ErrorHandler throws InvalidArgumentException
This is a (multiple allowed):

* [x] bug
* [ ] enhancement

* CakePHP Version: 4.1.7
* Platform and Target: nginx 1.19.5, PHP 7.4.13

### What you did
Implemented custom error handler based on docs at https://book.cakephp.org/4/en/development/errors.html#creating-your-own-error-handler

### What happened
The issue is the documentation says that the custom error handler should be implemented in following way:
`class AppError extends BaseErrorHandler` 
The check in ErrorHandlerMiddleware:101 checks for the following condition:
`if (!$errorHandler instanceof ErrorHandler) {` .
As the ErrorHandler is actually a child of BaseErrorHandler, this condition can't go through when custom error handler extends the BaseErrorHandler class.

The condition in ErrorHandlerMiddleware:101 should be evaluated as follows:
`if (!$errorHandler instanceof BaseErrorHandler) {`
or the documentation should be altered so that the custom error handler extends ErrorHandler rather than BaseErrorHandler.
Also the class in documentation is designated as ErrorHandlingMiddleware instead of ErrorHandlerMiddleware

markstory:
Check for the base class instead of the implementation.

Fixes #15254

### DIFF
--- a/src/Error/Middleware/ErrorHandlerMiddleware.php
+++ b/src/Error/Middleware/ErrorHandlerMiddleware.php
@@ -71,7 +71,7 @@ class ErrorHandlerMiddleware implements MiddlewareInterface
     /**
      * Error handler instance.
      *
-     * @var \Cake\Error\ErrorHandler|null
+     * @var \Cake\Error\BaseErrorHandler|null
      */
     protected $errorHandler;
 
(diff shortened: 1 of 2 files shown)


markstory:
Closing this as it is the docs that need to be updated.

PR #15258, opened 2020-12-31
Salt stateless CSRF tokens as well

markstory:
Add salting to CSRF tokens to prevent BREACH weakness. This ports the
changes made to session based CSRF tokens to stateless ones as well.
This will result in longer tokens as both the cookie and the POST data
are base64 encoded, and the POST data is now 112 bytes instead of 56
bytes.


src/Http/Middleware/CsrfProtectionMiddleware.php
```diff
@@ -240,12 +304,18 @@ public function createToken(): string
      */
     protected function _verifyToken(string $token): bool
     {
-        if (strlen($token) <= self::TOKEN_VALUE_LENGTH) {
+        $decoded = base64_decode($token);
+        // If base64 fails we're in a compatibility mode from before
+        // tokens were salted on each request.
+        if ($decoded === false) {
+            $decoded = $token;
```
othercorey, line 311: Should we add a warning here? Users shouldn't see this occur more than once per request/refresh, yes?

othercorey:
> Strict comparison using === between string and false will always evaluate to false.

This seems like a bug in phpstan. We want it to return false for string comparisons, and it should since the return type is `string|false`.

ADmad:
> This seems like a bug in phpstan.

Similar code does not throw any error on it's playground https://psalm.dev/r/510895c6c3

othercorey:
The error is in phpstan: https://phpstan.org/r/d51986a7-1515-4135-8c1e-f872ac58f2e3

src/Http/Middleware/CsrfProtectionMiddleware.php
```diff
@@ -227,9 +240,60 @@ protected function _createToken(): string
...
+     * If the token is not TOKEN_VALUE_LENGTH * 2 it is an old
+     * unsalted value that is supported for backwards compatibility.
+     *
+     * @param string $token The token that could be salty.
+     * @return string An unsalted token.
+     */
+    public function unsaltToken(string $token): string
+    {
+        $decoded = base64_decode($token);
+        if ($decoded === false || strlen($decoded) !== static::TOKEN_WITH_CHECKSUM_LENGTH * 2) {
```
ADmad, line 284: `base64_decode()` returns false only if 2nd argument is `true` https://www.php.net/manual/en/function.base64-decode.php

src/Http/Middleware/CsrfProtectionMiddleware.php
ADmad: I don't think we should show explicit warnings for internal changes which have been mitigated and don't require the dev to make any changes.

othercorey: What I am asking is whether this introduces a security risk with the fallback.

src/Http/Middleware/CsrfProtectionMiddleware.php
othercorey: Good catch. Looks like they have a special case somewhere other than the functionMap I checked. No error when 2nd argument is true.

src/Http/Middleware/CsrfProtectionMiddleware.php
markstory: I don't think there is anything new being introduced. Without this flow we wouldn't be able to accept old tokens which would break applications as they upgrade.

src/Http/Middleware/CsrfProtectionMiddleware.php
markstory: Thanks for assist on this :bow:. I added the strict parameter to all the places we decode tokens from external sources.

src/Http/Middleware/CsrfProtectionMiddleware.php
othercorey: Ok. As long as they ensure the token is refreshed. Up to them. We should probably test this though since codecov says it's untested.

Issue #15244, opened 2020-12-24
API Documentation errors and warnings for 4.2


othercorey:
~2020-12-24 19:22:22 Error: Missing @param for `expression` in `\Cake\Database\Driver\Sqlite::_transformFunctionExpression()`.~
~2020-12-24 19:22:22 Error: Missing @param for `original` in `\Cake\Database\Driver\Sqlserver::_transformDistinct()`.~
~2020-12-24 19:22:23 Error: Missing @param for `visitor` in `\Cake\Database\Expression\ComparisonExpression::traverse()`.~
~2020-12-24 19:22:36 Warning: Found invalid @see for `\Cake\I18n\PackageLocator::__construct()`.~
~2020-12-24 19:22:37 Error: Missing @var for `\Cake\I18n\TranslatorRegistry::FALLBACK_LOADER`. Using `mixed.`~
~2020-12-24 19:22:46 Error: Missing @var for `\Cake\Validation\Validation::COMPARE_STRING`. Using `mixed.`~
~2020-12-24 19:22:47 Error: Missing @var for `\Cake\View\Form\ContextInterface::VALID_ATTRIBUTES`. Using `mixed.`~

~2020-12-24 19:22:40 Warning: Missing description for @method `getAccessible` on `\Cake\Datasource\EntityInterface`.~
2020-12-24 19:22:46 Warning: Missing description for @method `extension` on `\Cake\Validation\RulesProvider`.

othercorey:
@dereuromark can you add a description for getAccessible()? git says it was added in https://github.com/cakephp/cakephp/pull/14251

othercorey:
Does anyone know how the `extension` method in `RulesProvider` should be documented?

PR #15259, opened 2020-12-31
Fix up accessible docblock

dereuromark:
https://github.com/cakephp/cakephp/issues/15244#issuecomment-752682827


othercorey:
Thanks!

PR #15260, opened 2020-12-31
Use new style place holders in docblocks

ADmad:


ADmad:
At some point we should consider showing warnings for old style placeholders.

PR #15262, opened 2021-01-02
Improve flash message capturing

### LINKED ISSUE #15268: Regression in 4.2.2 ? Flash Messages
This is a (multiple allowed):

* [x] bug
* [ ] enhancement

* CakePHP Version: 4.2.2
* Platform and Target: PHP 7.4.11

### What you did
```php
    /**
     * Test resetPassword method
     *
     * @return void
     */
    public function testResetPassword(): void
    {
        $this->enableCsrfToken();
        $this->enableSecurityToken();

        $this->enableRetainFlashMessages();
        $this->post('/users/reset-password', ['reset_password_token' => 'validresettoken']);
        $this->assertResponseSuccess();
        $this->assertSession('The Password has been reset.', 'Flash.flash.0.message');
        $this->assertRedirect(['controller' => 'Users', 'action' => 'login']);

        $this->post('/users/reset-password', ['reset_password_token' => 'unknownresetpasswordtoken']);
        $this->assertResponseSuccess();
        $this->assertSession('The Password Reset Token is unknown.', 'Flash.flash.0.message');
        $this->assertRedirect(['controller' => 'Users', 'action' => 'login']);
    }
```
### What happened
This test started to fail after updating to 4.2.2. Worked fine before.

Failed asserting that 'The Password Reset Token is unknown.' is in session path 'Flash.flash.0.message'.

using $this->assertFlashMessage didn't work either.

Test should still pass?

It still works when splitting the test into two test cases

```php

    /**
     * Test resetPassword method with valid token
     *
     * @return void
     */
    public function testResetPasswordValid(): void
    {
        $this->enableCsrfToken();
        $this->enableSecurityToken();

        $this->enableRetainFlashMessages();
        $this->post('/users/reset-password', ['reset_password_token' => 'validresettoken']);
        $this->assertResponseSuccess();
        $this->assertFlashMessageAt(0, 'The Password has been reset.');
        $this->assertRedirect(['controller' => 'Users', 'action' => 'login']);
    }

    /**
     * Test resetPassword method with invalid token
     *
     * @return void
     */
    public function testResetPasswordInvalid(): void
    {
        $this->enableCsrfToken();
        $this->enableSecurityToken();

        $this->enableRetainFlashMessages();
        $this->post('/users/reset-password', ['reset_password_token' => 'unknownresetpasswordtoken']);
        $this->assertResponseSuccess();
        $this->assertFlashMessageAt(0, 'The Password Reset Token is unknown.');
        $this->assertRedirect(['controller' => 'Users', 'action' => 'login']);
    }
```

markstory:
Some view implementations (like inertiajs) manipulate flash messages and make them into view variables during the `View::render()` method. This breaks flash assertions as the messages are deleted.

By capturing the messages with a early priority listener in the `Controller.beforeRender` event we can capture flash messages more reliably. The trade off being that we will not capture any changes made to flash messages during the controller `beforeRender` or during view initialization.

### DIFF
--- a/src/TestSuite/IntegrationTestTrait.php
+++ b/src/TestSuite/IntegrationTestTrait.php
@@ -511,18 +511,17 @@ trait IntegrationTestTrait
         }
         $this->_controller = $controller;
         $events = $controller->getEventManager();
-        $events->on('Controller.beforeRedirect', function ($event): void {
-            $controller = $event->getSubject();
-            if ($this->_retainFlashMessages) {
+        $flashCapture = function (EventInterface $event): void {
+            if ($this->_retainFlashMessages && empty($this->_flashMessages)) {
+                $controller = $event->getSubject();
                 $this->_flashMessages = $controller->getRequest()->getSession()->read('Flash');
             }
-        });
-        $events->on('View.beforeRender', function ($event, $viewFile) use ($controller): void {
+        };
+        $events->on('Controller.beforeRedirect', ['priority' => -100], $flashCapture);
+        $events->on('Controller.beforeRender', ['priority' => -100], $flashCapture);
+        $events->on('View.beforeRender', function ($event, $viewFile): void {
             if (!$this->_viewName) {
                 $this->_viewName = $viewFile;
-            }
-            if ($this->_retainFlashMessages) {
-                $this->_flashMessages = $controller->getRequest()->getSession()->read('Flash');
             }
         });
         $events->on('View.beforeLayout', function ($event, $viewFile): void {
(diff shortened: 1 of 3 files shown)


othercorey:
> By capturing the messages with a early priority listener in the `Controller.beforeRender` event we can capture flash messages more reliably. The trade off being that we will not capture any changes made to flash messages during the controller `beforeRender` or during view initialization.

What's the danger of capturing it in a normal priority listener?

markstory:
A normal priority listener would trigger after the controller's beforeRender function as that method is registered first, and thus triggered first. A lower priority ensures that the integration test case listener is fired first.

PR #15263, opened 2021-01-02
4.next - Make association queries inherit the results casting mode.

ndm2:
This brings the behavior in line with hydration, which is already being inherited accordingly.

Unfortunately it's the first time now that I ran into this, would have been good if it would've been included in #14863 for 4.2. You could call it a feature, you could call it a bug fix, it certainly is a change in behavior that could break things, even if it would probably be rather rare.

People who disabled results casting in combination with associations would have probably noticed it, and worked around it by manually disabling casting for the association query builders, just like I did now, leaving only few people who have the need to disable casting only partially, but still.

### DIFF
--- a/tests/TestCase/ORM/QueryTest.php
+++ b/tests/TestCase/ORM/QueryTest.php
@@ -4285,50 +4285,36 @@ class QueryTest extends TestCase
             ->belongsTo('Articles')
             ->setStrategy(BelongsTo::STRATEGY_SELECT);
 
-        $disabled = true;
-        $enabled = true;
-
         $articles
             ->find()
-            ->contain('Comments', function (Query $query) use (&$disabled) {
-                $disabled =
-                    $disabled &&
-                    !$query->isHydrationEnabled() &&
-                    !$query->isResultsCastingEnabled();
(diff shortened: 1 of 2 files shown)


tests/TestCase/ORM/QueryTest.php
```diff
@@ -4265,4 +4266,69 @@ public function testSubqueryJoinClause()
...
+        $disabled = true;
+        $enabled = true;
+
+        $articles
+            ->find()
+            ->contain('Comments', function (Query $query) use (&$disabled) {
+                $disabled =
+                    $disabled &&
+                    !$query->isHydrationEnabled() &&
+                    !$query->isResultsCastingEnabled();
```
markstory, line 4297: You can use assertions inside these closures to avoid needing to close over disabled and enabled.

ndm2: It's a habit... I'm always thinking, "_what if a closure isn't being invoked for some reason_", but in this case it wouldn't prevent things from not being covered anyways, as the initial values equal the expected values 🙃

PR #15266, opened 2021-01-04
Fix CSRF token backwards compatibility

markstory:
I did more thorough testing with salted CSRF tokens and this change is required to provide backwards compatibility. Because the old tokens are hexadecimal, they also decode as base64 data. When an old token is found in the cookie we shouldn't base64 decode or apply salting, as the data becomes hard to manage in subsequent requests.

### DIFF
--- a/src/Http/Middleware/CsrfProtectionMiddleware.php
+++ b/src/Http/Middleware/CsrfProtectionMiddleware.php
@@ -236,12 +236,15 @@ class CsrfProtectionMiddleware implements MiddlewareInterface
     /**
      * Test if the token predates salted tokens.
      *
-     * These tokens are vulnerable to BREACH but will rotate over time.
+     * These tokens are hexadecimal values and equal
+     * to the token with checksum length. While they are vulnerable
+     * to BREACH they should rotate over time and support will be dropped
+     * in 5.x.
      *
      * @param string $token The token to test.
      * @return bool
      */
-    protected function isOldToken(string $token): bool
+    protected function isUnsaltedToken(string $token): bool
     {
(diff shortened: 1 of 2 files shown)


src/Http/Middleware/CsrfProtectionMiddleware.php
```diff
@@ -233,6 +233,19 @@ protected function _createToken(): string
...
 
+    /**
+     * Test if the token predates salted tokens.
+     *
+     * These tokens are vulnerable to BREACH but will rotate over time.
+     *
+     * @param string $token The token to test.
+     * @return bool
+     */
+    protected function isOldToken(string $token): bool
```
othercorey, line 244: Lets be explicit as old can mean anything. isPreSaltedToken.

src/Http/Middleware/CsrfProtectionMiddleware.php
```diff
@@ -257,7 +273,10 @@ public function createToken(): string
      */
     public function saltToken(string $token): string
     {
-        $decoded = base64_decode($token);
+        if ($this->isUnsaltedToken($token)) {
```
othercorey, line 276: Why do we need to check here before salting? It doesn't read correctly.

markstory: If the token is hexadecimal only then it shouldn't be salted as it doesn't contain base64 data. I could change the method name to be `isBase64Encoded()` instead.

othercorey: I see. yea, changing the name seems necessary then so it doesn't read like a bug. I guess I'd expect this to be outside of the saltToken() function, but as long as it's consistent.

markstory: Changed the name to `isHexadecimalToken()` to be less confusing.

PR #15270, opened 2021-01-07
Fix database session entity class error

RuGa:
Fix allow use correct custom class and his mutators in table events instead base Entity class.

### DIFF
--- a/src/Http/Session/DatabaseSession.php
+++ b/src/Http/Session/DatabaseSession.php
@@ -18,7 +18,6 @@ declare(strict_types=1);
  */
 namespace Cake\Http\Session;
 
-use Cake\ORM\Entity;
 use Cake\ORM\Locator\LocatorAwareTrait;
 use SessionHandlerInterface;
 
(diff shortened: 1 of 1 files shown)


src/Http/Session/DatabaseSession.php
```diff
@@ -155,7 +154,9 @@ public function write($id, $data): bool
         /** @var string $pkField */
         $pkField = $this->_table->getPrimaryKey();
         $record[$pkField] = $id;
-        $result = $this->_table->save(new Entity($record));
+        $entityClass = $this->_table->getEntityClass();
+        $entity = new $entityClass($record);
```
ADmad, line 158: We could just `$this->_table->newEntity($record)`

RuGa: Then need add "id" to $_accessible in Entity Class for mass assigned using newEntity() or patchEntity()

markstory: You can do that using the options to newEntity().

ADmad:
@RuGa ping

othercorey:
@RuGa Can you make the changes to use `newEntity()`?

othercorey:
Implemented in https://github.com/cakephp/cakephp/pull/15345

Issue #15224, opened 2020-12-16
Supporting NOT IN with nullable columns


othercorey:
When you compare a column to a list of values using `NOT IN`, the result can be unexpected if the comparison is against `NULL`. In order to support a nullable column in the comparison, you have to add a separate condition when the column is null.

```
->where([
    'OR' => [
        ['nullable_column NOT IN' => ['value1', 'value2']],
        ['nullable_column IS' => null]
    ],
]);
```

Would it be possible to add an options to `notIn()` to automatically generate this conditions to support nullable columns.

dereuromark:
Technically, a similar problem is also with != and normal comparison, if you have both empty string and NULL in our columns.
But that is probably a user problem.

othercorey:
Is it possible to find the schema for a field in QueryExpression? All we have access to is the typemap which doesn't tell us whether the field is nullable.

markstory:
Do we need the schema?

othercorey:
You think it would be ok to generate that condition in all cases? It would make the query more complicated by default, but it's necessary when the column is nullable.

markstory:
I think its a reasonable to generate the IS NOT fragment whenever there is a null in the value options.

othercorey:
> I think its a reasonable to generate the IS NOT fragment whenever there is a null in the value options.

The issue is when the column is nullable. If the column contains the value `null` then it will fail the NOT IN check no matter what's in the list because SQL will perform a `null != value` check.

markstory:
Ah ok, then auto-detecting when to add the null clause gets harder. Should this be a separate method on the builder? Something like `$exp->notInWithNull()`?

PR #15278, opened 2021-01-11
Add Query::whereNotInListOrNull() and QueryExpression::notInOrNull()

othercorey:
Closes https://github.com/cakephp/cakephp/issues/15224

### DIFF
--- a/src/Database/Expression/QueryExpression.php
+++ b/src/Database/Expression/QueryExpression.php
@@ -375,7 +375,7 @@ class QueryExpression implements ExpressionInterface, Countable
      * @param string|null $type the type name for $value as configured using the Type map.
      * @return $this
      */
-    public function notInOrNull($field, $values, $type = null)
+    public function notInOrNull($field, $values, ?string $type = null)
     {
         $or = new static([], [], 'OR');
         $or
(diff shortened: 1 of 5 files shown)


src/Database/Expression/QueryExpression.php
```diff
@@ -366,6 +366,25 @@ public function notIn($field, $values, $type = null)
...
+    /**
+     * Adds a new condition to the expression object in the form
+     * "(field NOT IN (value1, value2) OR field IS NULL".
+     *
+     * @param string|\Cake\Database\ExpressionInterface $field Database field to be compared against value
+     * @param string|array|\Cake\Database\ExpressionInterface $values the value to be bound to $field for comparison
+     * @param string|null $type the type name for $value as configured using the Type map.
+     * @return $this
+     */
+    public function notInOrNull($field, $values, $type = null)
```
dereuromark, line 378:
```suggestion
    public function notInOrNull($field, $values, ?string $type = null)
```

othercorey: This is copied from `notIn()` above so it's the same.

dereuromark: Sure, but here you do not have to follow BC issue of the signature.

PR #15283, opened 2021-01-13
Remove incorrect argument type.

### LINKED ISSUE #15280: Cake\Validation\Validation::imageWidth() should expect an UploadedFile object instead of an array
This is a (multiple allowed):

* [x] bug
* [ ] enhancement

* CakePHP Version: 4.1.7
* Platform and Target: OpenSuse 15.2, MySQL

### What you did
I tried to patch an entity with form data that contained an UploadedImage object.
I also have the following validation rule:

```php
$validator->add('image', 'largeurImageInvalide', [
    'rule' => ['imageWidth', '==', 640],
    'message' => __('The image width must be 640 pixels.'),
]);
```

### What happened
`'Argument 1 passed to Cake\Validation\Validation::imageWidth() must be of the type array, object given'.`

Validation should not throw an error.

ADmad:
Closes #15280

### DIFF
--- a/src/Validation/Validation.php
+++ b/src/Validation/Validation.php
@@ -1265,7 +1265,7 @@ class Validation
      * Helper for reading the file out of the various file implementations
      * we accept.
      *
-     * @param string|array|\Psr\Http\Message\UploadedFileInterface $check The data to read a filename out of.
+     * @param mixed $check The data to read a filename out of.
      * @return string|false Either the filename or false on failure.
      */
     protected static function getFilename($check)
--- a/src/Validation/Validation.php
+++ b/src/Validation/Validation.php
@@ -1463,12 +1463,12 @@ class Validation
     /**
      * Validates the image width.
      *
-     * @param array $file The uploaded file data from PHP.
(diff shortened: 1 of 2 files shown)


src/Validation/Validation.php
```diff
@@ -1463,12 +1463,12 @@ public static function imageSize($file, array $options): bool
     /**
      * Validates the image width.
      *
-     * @param array $file The uploaded file data from PHP.
```
dereuromark, line 1466: Could we just be concrete here with array|UploadedFile ?
That would help also understanding of IDE/PHPStan etc.

ADmad: All other validation methods also have `mixed` since the argument comes from user input which can be of any type.

dereuromark: But then you need to guard it with a type check and early return false
Otherwise, semantically, this can blow up from user land site as internal error without it supposed to happen.

ADmad: > But then you need to guard it with a type check and early return false

Which is done by the `Valdiation::getFilename()` method which is called further down the call chain.

PR #15284, opened 2021-01-14
# Regression ? regex can be 'int'

ghost:
Hi, i start to migrate into 4.x it's my first contribute to cakephp, so excuse me if it's not the right way.
I'am not agree with the new 'custom' validator is_string check, because i use regex validation and allow some 'int' string in my field.
So the 'is_string' break the validator.
For example, user can set 'value', or simply '1' and I use regex for check that, the final fields it's a string but a string of 'int' maybe.
I don't know if explanation is great, but the code below will explain better.
Regards

### DIFF
--- a/src/Validation/Validation.php
+++ b/src/Validation/Validation.php
@@ -442,10 +442,6 @@ class Validation
      */
     public static function custom($check, ?string $regex = null): bool
     {
-        if (!is_string($check)) {
-            return false;
-        }
-
         if ($regex === null) {
             static::$errors[] = 'You must define a regular expression for Validation::custom()';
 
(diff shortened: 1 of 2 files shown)


dereuromark:
I agree that a "is string castable" would maybe be a better fit to guard against.

Thx3r:
Thank you ! :+1:

src/Validation/Validation.php
```diff
@@ -442,10 +442,6 @@ public static function containsNonAlphaNumeric($check, int $count = 1): bool
      */
     public static function custom($check, ?string $regex = null): bool
     {
-        if (!is_string($check)) {
```
markstory, line 445: This check is intended to keep non-scalar data out as arrays could come in via request data. Could we use `is_scalar()` instead of removing this check?

Thx3r: Yes, it's work great. I have patched my forked pull request version.

markstory:
Thank you 👍

PR #15287, opened 2021-01-17
4.next - Add joins when `onlyTranslated` or `filterByCurrentLocale` is enabled.

ndm2:
These options indicate the possible intent of applying filtering solely
based on translations existing, hence the join should always apply when
they are enabled.

### DIFF
--- a/src/ORM/Behavior/Translate/ShadowTableStrategy.php
+++ b/src/ORM/Behavior/Translate/ShadowTableStrategy.php
@@ -135,7 +135,10 @@ class ShadowTableStrategy implements TranslateStrategyInterface
 
         $fieldsAdded = $this->addFieldsToQuery($query, $config);
         $orderByTranslatedField = $this->iterateClause($query, 'order', $config);
-        $filteredByTranslatedField = $this->traverseClause($query, 'where', $config);
+        $filteredByTranslatedField =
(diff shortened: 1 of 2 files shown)


src/ORM/Behavior/Translate/ShadowTableStrategy.php
```diff
@@ -135,7 +135,10 @@ public function beforeFind(EventInterface $event, Query $query, ArrayObject $opt
 
         $fieldsAdded = $this->addFieldsToQuery($query, $config);
         $orderByTranslatedField = $this->iterateClause($query, 'order', $config);
-        $filteredByTranslatedField = $this->traverseClause($query, 'where', $config);
+        $filteredByTranslatedField =
+            $this->traverseClause($query, 'where', $config) ||
+            $config['onlyTranslated'] ||
+            ($options['filterByCurrentLocale'] ?? null);
```
ADmad, line 141: Is this a new option added?

ndm2: `filterByCurrentLocale`? No, it exists ever since 3.0, but it seems that it was never properly documented.

ADmad: Ah, I see now that it's an existing option, albeit undocumented.

Issue #15288, opened 2021-01-18
Different behavior for default & sprintf formatter for __n()


bartikowskiw:
* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: master (3f59335fd9), also tested with 4.2.2 and 3.9.5

### What you did
Using the plural translation functions as `__n()` with the SprintfFormatter (instead of the default IcuFormatter.)

### What happened
The way the placeholders are replaced differs between the SprintfFormatter and the IcuFormatter:

```php
echo __n('Singular: {0}', 'Plural: {0}', 1, 'singular value') . PHP_EOL;
echo __n('Singular: {0}', 'Plural: {0}', 9, 'plural value') . PHP_EOL;
```

results in

```
Singular: singular value
Plural: plural value
```

for the default formatter, but

```php
echo __n('Singular: %s', 'Plural: %s', 1, 'singular value') . PHP_EOL;
echo __n('Singular: %s', 'Plural: %s', 9, 'plural value') . PHP_EOL;
```

uses results in

```
Singular: 1
Plural: 9
```

### What you expected to happen
The translations should be the same for both formatters.

PR #15289, opened 2021-01-18
Issue #15288. Fix SprintfFormatter.

bartikowskiw:
Move cleanup of tokens to the Translator. Update tests.

See https://github.com/cakephp/cakephp/issues/15288

### DIFF
--- a/src/I18n/Formatter/IcuFormatter.php
+++ b/src/I18n/Formatter/IcuFormatter.php
@@ -37,23 +37,6 @@ class IcuFormatter implements FormatterInterface
      */
     public function format(string $locale, string $message, array $tokenValues): string
     {
-        unset($tokenValues['_singular'], $tokenValues['_count']);
-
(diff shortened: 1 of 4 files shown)


ADmad (approved):
LGTM

markstory:
I'm in favour of making the formatters consistent, but this seems like it could break applications that are using the default formatter, as their output will change in meaningful ways. Perhaps this should be moved to 4.3 where we can better communicate the behavior change and not surprise people as they upgrade through point releases.

ADmad:
@markstory I don't think the output for the `IcuFormatter` has changed changed in any way. He just removed the additional `_formatMessage()` message method and move it's code inside `format()` itself. The unsetting of keys from `$tokenValues` was removed and it's redundant now since the calling method itself cleans them up. 

We can preserve the redundancy in patch release and removed it in next minor if we want to be extra cautious.

PR #15292, opened 2021-01-19
Regression marshaling date string to Datetime

ghost:
Hello,
I am continuing my migrations to V4.x.
I want to be able to marshal / cast a date string like '2021-01-01' to a database DateTime.
In v4.x is not possible, because cake want the full string datetime format.
I understand this, but if a developper want to store a string date into a datetime  bdd field, it's beacause for this context it doesn't matter about 'H: m :s'.
I think it's important to preserve casting about 'Y-m-d' -> DateTime like v3.

In V3 :

src/Database/Type/DateTimeType.php l.246

```php
if (ctype_digit($value)) {
                $date = new $class('@' . $value);
            } elseif ($isString && $this->_useLocaleParser) {
                return $this->_parseValue($value);
            } elseif ($isString) {
                $date = new $class($value);
                $compare = true;
            }
```
In V4 : 

src/Database/Type/DateTimeType.php l.309

```php
if (ctype_digit($value)) {
                return new $class('@' . $value);
            } elseif ($isString && $this->_useLocaleMarshal) {
                return $this->_parseLocaleValue($value);
            } elseif ($isString) {
                return $this->_parseValue($value);
            }
```

I add the support of 'Y-m-d' casting in commit here.
If not possible to merge it, we have to raise an Error, because today it's return null and it's dangerous.

### DIFF
--- a/src/Database/Type/DateTimeType.php
+++ b/src/Database/Type/DateTimeType.php
@@ -64,6 +64,7 @@ class DateTimeType extends BaseType implements BatchCastingInterface
         'Y-m-d\TH:i',
         'Y-m-d\TH:i:s',
         'Y-m-d\TH:i:sP',
+        'Y-m-d',
     ];
 
     /**
--- a/src/Database/Type/DateTimeType.php
+++ b/src/Database/Type/DateTimeType.php
@@ -473,6 +473,11 @@ class DateTimeType extends BaseType implements BatchCastingInterface
 
         foreach ($this->_marshalFormats as $format) {
             try {
+                // if date format, force casting into correct DateTime with empty H:i:s
+                if($format === 'Y-m-d' && preg_match('/^[\d]{4}[-][\d]{2}[-][\d]{2}$/', $value)) {
+                    $format = 'Y-m-d H:i:s';
+                    $value .= ' 00:00:00';
+                }
                 $dateTime = $class::createFromFormat($format, $value);
                 // Check for false in case DateTime is used directly
                 if ($dateTime !== false) {
--- a/src/Database/Type/DateTimeType.php
+++ b/src/Database/Type/DateTimeType.php
@@ -474,7 +474,7 @@ class DateTimeType extends BaseType implements BatchCastingInterface
         foreach ($this->_marshalFormats as $format) {
             try {
                 // if date format, force casting into correct DateTime with empty H:i:s
-                if($format === 'Y-m-d' && preg_match('/^[\d]{4}[-][\d]{2}[-][\d]{2}$/', $value)) {
+                if ($format === 'Y-m-d' && preg_match('/^[\d]{4}[-][\d]{2}[-][\d]{2}$/', $value)) {
                     $format = 'Y-m-d H:i:s';
                     $value .= ' 00:00:00';
                 }
--- a/tests/TestCase/ORM/MarshallerTest.php
+++ b/tests/TestCase/ORM/MarshallerTest.php
@@ -221,6 +221,10 @@ class MarshallerTest extends TestCase
         $data['created'] = 1392387900;
         $result = $marshall->one($data, []);
         $this->assertSame($data['created'], $result->created->getTimestamp());
+
+        $data['created'] = '2014-02-14';
+        $result = $marshall->one($data, []);
+        $this->assertEquals('2014-02-14', $result->created->format('Y-m-d'));
     }
 
     /**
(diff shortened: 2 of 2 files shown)


tests/TestCase/ORM/MarshallerTest.php
```diff
@@ -221,6 +221,10 @@ public function testOneWithDatetimeField()
         $data['created'] = 1392387900;
         $result = $marshall->one($data, []);
         $this->assertSame($data['created'], $result->created->getTimestamp());
+
+        $data['created'] = '2014-02-14';
+        $result = $marshall->one($data, []);
+        $this->assertEquals('2014-02-14', $result->created->format('Y-m-d'));
```
othercorey, line 227:
```suggestion
        $this->assertEquals(new Time('2014-02-14 00:00:00'), $result->created);
```

src/Database/Type/DateTimeType.php
```diff
@@ -64,6 +64,7 @@ class DateTimeType extends BaseType implements BatchCastingInterface
         'Y-m-d\TH:i',
         'Y-m-d\TH:i:s',
         'Y-m-d\TH:i:sP',
+        'Y-m-d',
```
othercorey, line 67: Move this to the front of the array.

tests/TestCase/ORM/MarshallerTest.php
Thx3r: this test will fail

othercorey: Then the format needs to be changed to re-create 00:00:00 time. `new DateTime('2014-02-14')` would create `2014-02-14 00:00:00`.

src/Database/Type/DateTimeType.php
ADmad: Since date only string would be a uncommon format for datetime, wouldn't it be better to keep it last?

tests/TestCase/ORM/MarshallerTest.php
Thx3r: I am sorry i don't understand. For me we have to edit masrhall for detect if it's 'Y-m-d' format and cast as 'Y-m-d 00:00:00' . rigth ?

src/Database/Type/DateTimeType.php
Thx3r: Ok may be it's not a good idea

othercorey: I gave up trying to optimize the list and just made it readable/grouped. We can leave it at the bottom though.

tests/TestCase/ORM/MarshallerTest.php
ADmad: Yes, if only date part is passed then the time part should be always set to `00:00:00`, irrespective of current time.

othercorey:
Are we sure this worked properly before?

```
     if ($compare && $date && !$this->_compare($date, $value)) {
        return $value;
     }
```

Wouldn't this simply return the original string value? We're now adding support for the format and assuming it should be the same as `new $class(...)`. It looks like it's not marshalled into a DateTime instance.

src/Database/Type/DateTimeType.php
markstory: Shouldn't this be in the DateType? DateTimeType is for date and time.

tests/TestCase/ORM/MarshallerTest.php
markstory: Shouldn't you be using the DateType if you only want to use the date portion or have a date column?

Thx3r:
I commit a new robust version of casting.

Thx3r:
I don't know why test fail with pgsql

Thx3r:
> Are we sure this worked properly before?
> 
> ```
>      if ($compare && $date && !$this->_compare($date, $value)) {
>         return $value;
>      }
> ```
> 
> Wouldn't this simply return the original string value? We're now adding support for the format and assuming it should be the same as `new $class(...)`. It looks like it's not marshalled into a DateTime instance.

Yes it's seems to marshalled as String into validator and not DateTime instance

markstory:
Why do you need to marshal only the date component of a datetime? Could you not use the DateType instead?

ADmad:
I agreed, one should use DateTime type of they want to marshal/store only strings. If for some reason one needs to marshal date only strings for datetime columns then they can tack on the time part in `beforeMarshal()`.

Thx3r:
Ok i will do that

PR #15293, opened 2021-01-19
Include message body in serialized Message.

### LINKED ISSUE #15290: Message::jsonSerialize() omits message body
This is a (multiple allowed):

* [x] bug
* [ ] enhancement

* CakePHP Version: 4.1.3

### What you did
Tried to serialize and then unserialize fully composed email message.

### What happened
After unserialization the message misses the body.

The unserialized message should keep the message body.

https://github.com/cakephp/cakephp/blob/master/src/Mailer/Message.php#L1852-L1856

Note that `$message`, `$textMessage` and `$htmlMessage` properties are missing here.

Is that intentional?

ADmad:
Closes #15290.


robertpustulka:
Thanks!

PR #15294, opened 2021-01-20
Fix to work under PHP 8.0.1

kiang:
In PHP 8.0.1 (Ubuntu), it's not possible to execute any command as it would return this error:
Error: ReflectionMethod::__construct(): Argument #2 ($method) cannot be null when argument #1 ($objectOrMethod) is an object

Just add the fix to prevent empty value in hasMethod().

### DIFF
--- a/lib/Cake/Console/Shell.php
+++ b/lib/Cake/Console/Shell.php
@@ -355,6 +355,9 @@ class Shell extends CakeObject {
  */
 	public function hasMethod($name) {
 		try {
+			if(empty($name)) {
+				return false;
+			}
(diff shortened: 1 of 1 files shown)


lib/Cake/Console/Shell.php
```diff
@@ -355,6 +355,9 @@ public function hasTask($task) {
  */
 	public function hasMethod($name) {
 		try {
+			if(empty($name)) {
```
dereuromark, line 358:
```suggestion
			if (empty($name)) {
```

ADmad:
CakePHP 2.x has reached EOL and won't have any new releases except for security fixes.

PR #15296, opened 2021-01-20
Fix CI.

markstory:
2015 isn't in the default year options anymore.


othercorey:
How did this change?

markstory:
@othercorey DateTime widget includes the last x years. 2015 is no longer in that range.

PR #15297, opened 2021-01-20
Port changes from #15295 to 4.x

markstory:
Include frame arguments by default in stack traces generated for debug errors.


phpbench.json
```diff
@@ -0,0 +1,4 @@
+{
```
ADmad, line 1: Is the addition of this file intentional?

markstory: Nope.

PR #15310, opened 2021-01-28
4.next - Fix logging via debugger applies Console/HTML formatting.

ndm2:
With the introduction of Console/HTML formatting, logging started applying the formatting too, which I would assume to be pretty much unintended, and usually not what users would expect.

I would like to suggest introducing a new method, as the `Debugger::exportVar()` behavior changed too. Certainly the plain text formatting could be done directly in `Debugger::log()` instead, however I'm kinda missing a public shorthand method that would export as plain text, as at least for me the use case 99% of the times using `Debugger::exportVar()`, was custom logging. `Debugger::exportVarAsNodes()` is nice, giving us the option to use any formatter on the fly, but a shorthand for this specific case might be considered reasonable.

### DIFF
--- a/src/Error/Debugger.php
+++ b/src/Error/Debugger.php
@@ -348,7 +348,10 @@ class Debugger
         $source = static::trace(['start' => 1]);
         $source .= "\n";
 
-        Log::write($level, "\n" . $source . static::exportVar($var, $maxDepth));
+        Log::write(
+            $level,
+            "\n" . $source . static::exportVarAsPlainText($var, $maxDepth)
+        );
     }
 
     /**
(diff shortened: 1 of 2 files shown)


othercorey:
Maybe it's cleaner to pass a formatter to `exportVar()` for custom formatting.

markstory:
@othercorey Do you have scenarios where you're forcing HTML or console output?

othercorey:
> @othercorey Do you have scenarios where you're forcing HTML or console output?

No. I was only thinking that these wrappers and names might be confusing. `exportVar()` vs `exportVarAsPlainText()` doesn't exactly make it clear the difference.

ndm2:
I first tried modifying `exportVar()`, but it's not very verbose, and without "breaking" the signature one would always have to supply the depth, which felt a little annoying.

Maybe if restoring the old behavior would be an option, ie making `exportVar()` by default not apply any formatting, and require requesting formatting instead, then something like `exportVar($var, $depth, class-string|bool)` would be less awkward, as it would probably pretty much exclusively be used internally.

Instead of expanding the `exportVar()` signature, introducing a new method like `exportVarFormatted()` that optionally takes a formatter class might be an option too?

markstory:
> Instead of expanding the exportVar() signature, introducing a new method like exportVarFormatted() that optionally takes a formatter class might be an option too?

This would require changes to other parts of the framework and DebugKit which are using the formatted result. I'm ok with the proposed method name to get `Debugger::log()` back on track.

othercorey:
Anything left to fix before merging?

PR #15317, opened 2021-01-30
Automatically set CSP nonces.

### LINKED ISSUE #15301: Integrate CSP nonces into HTMLHelper making strict CSP policies simpler to use
This is a (multiple allowed):

* [] bug
* [ ] enhancement

* CakePHP Version: 4.3.0

We currently provide a `CspMiddleware` that uses the [paragonie/csp-builder](https://github.com/paragonie/csp-builder/blob/master/src/CSPBuilder.php) to define content-security-policies.

When defining strict CSP policies it is necessary to use either nonces or hashed script blocks. Right now making script & style tags that use those attributes is challenging as additional middleware is needed to pass either the CSPBuilder into the view.

## Proposal

When the `CspMiddleware` is active, the framework can make using stricter CSP policies much easier. The `CspMiddleware` can add request attributes for a CSP nonce (`cspNonce`). The `HtmlHelper` would also be updated to read this attribute and automatically apply it to all script, style and stylesheet link tags it creates.

markstory:
When using the CSP middleware we can automatically set nonces for script and style sources. HtmlHelper can use then use these request attributes to generate HTML attributes. Having nonces automated eases the adoption of more strict content-security-policies

Fixes #15301

### DIFF
--- a/src/View/Helper/HtmlHelper.php
+++ b/src/View/Helper/HtmlHelper.php
@@ -483,7 +483,7 @@ class HtmlHelper extends Helper
         $defaults = [
             'block' => null,
             'once' => true,
-            'nonce' => $this->_View->getRequest()->getAttribute('cspScriptNonce')
+            'nonce' => $this->_View->getRequest()->getAttribute('cspScriptNonce'),
         ];
         $options += $defaults;
 
--- a/src/View/Helper/HtmlHelper.php
+++ b/src/View/Helper/HtmlHelper.php
@@ -376,7 +376,12 @@ class HtmlHelper extends Helper
      */
     public function css($path, array $options = []): ?string
     {
-        $options += ['once' => true, 'block' => null, 'rel' => 'stylesheet'];
+        $options += [
+            'once' => true,
+            'block' => null,
+            'rel' => 'stylesheet',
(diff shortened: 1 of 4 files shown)


src/View/Helper/HtmlHelper.php
```diff
@@ -524,6 +541,10 @@ public function script($url, array $options = []): ?string
     public function scriptBlock(string $script, array $options = []): ?string
     {
         $options += ['block' => null];
+        $nonce = $this->_View->getRequest()->getAttribute('cspScriptNonce');
+        if (!isset($options['nonce']) && $nonce) {
+            $options['nonce'] = $nonce;
+        }
```
ADmad, line 547:
```suggestion
        $options += ['block' => null, 'nonce' => $this->_View->getRequest()->getAttribute('cspScriptNonce')];
```

Wouldn't this be simpler? `formatAttributes()` skips keys with `null` values.

markstory: Yes. this would be better. :bow:

ADmad:
The manual probably needs to be updated with the new info regarding nonce.

othercorey:
https://github.com/cakephp/docs/pull/6935

PR #15318, opened 2021-01-31
Remove invalid sqlserver connection flag

### LINKED ISSUE #15315: Invalid PDO attribute on SQL server database connection
This is a (multiple allowed):

* [x] bug
* [ ] enhancement

* CakePHP Version: 4.2.2
* Platform and Target: Ubuntu 18 LTS, PHP 8.0.1, sqlsrv 5.9, pdo_sqlsrv 5.9, unixOdbc 2.3.7

### What you did
Tried to connect to a Azure SQL database instance.

### What happened
PHP Error.
Connection to Sqlserver could not be established: SQLSTATE[IMSSP]: The given attribute is only supported on the PDOStatement object. 

After the connection, the query returns some result.

Based on the [documentation](https://docs.microsoft.com/en-us/sql/connect/php/pdo-prepare?view=sql-server-ver15#remarks) what I found on microsoft.com the _PDO::ATTR_EMULATE_PREPARES_  is an attribute for PDO::prepare not for the connection.

In PHP 7.4 this is not a problem, but for PHP 8+ this throws an error.

When I moved `PDO::ATTR_EMULATE_PREPARES => false,` from connect() to prepare() in the [Sqlserver.php](https://github.com/cakephp/cakephp/blob/045122bd895847d4ce2a8173d31101bd5144dcc0/src/Database/Driver/Sqlserver.php#L132) solved the problem.

UPDATE: Actually you don't even need to move _PDO::ATTR_EMULATE_PREPARES_ to prepare() bcz according to the documentation, the default value is false - just remove it.

The problem looks like the PHP version. After 8.0 the [default error mode was changed to exception](https://www.php.net/ChangeLog-8.php).

This also throws an error
`$this->MyUserTables->getConnection()->getDriver()->getConnection()->setAttribute(\PDO::ATTR_EMULATE_PREPARES, true)`
even in PHP 7.4.

othercorey:
Fixes https://github.com/cakephp/cakephp/issues/15315

This causes a failure with pdo_sqlsrv 5.9 with php8 support. The flag is disabled by default so there's no need to set it for prepares anyway.

### DIFF
--- a/src/Database/Driver/Sqlserver.php
+++ b/src/Database/Driver/Sqlserver.php
@@ -129,7 +129,6 @@ class Sqlserver extends Driver
         }
 
         $config['flags'] += [
-            PDO::ATTR_EMULATE_PREPARES => false,
             PDO::ATTR_ERRMODE => PDO::ERRMODE_EXCEPTION,
         ];
 
--- a/tests/TestCase/Database/Driver/SqlserverTest.php
+++ b/tests/TestCase/Database/Driver/SqlserverTest.php
@@ -147,7 +147,6 @@ class SqlserverTest extends TestCase
 
         $expected = $config;
         $expected['flags'] += [
-            PDO::ATTR_EMULATE_PREPARES => false,
             PDO::ATTR_ERRMODE => PDO::ERRMODE_EXCEPTION,
             PDO::SQLSRV_ATTR_ENCODING => 'a-language',
         ];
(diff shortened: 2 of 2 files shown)


markstory:
Looks like there are a few tests that need to be updated.

othercorey:
> Looks like there are a few tests that need to be updated.

Reverted the switch to php8 and updated the sqlserver specific tests.

PR #15325, opened 2021-02-11
Test with php8 on windows

othercorey:


tests/TestCase/I18n/TimeTest.php
```diff
@@ -433,55 +436,33 @@ public function testNice($class)
...
-        $result = $time->i18nFormat([\IntlDateFormatter::SHORT, \IntlDateFormatter::FULL], null, 'fa-IR@calendar=persian');
-        $expected = '۱۳۸۸/۱۰/۲۴،‏ ۱۳:۵۹:۲۸ GMT';
-        $this->assertTimeFormat($expected, $result);
-
-        $result = $time->i18nFormat(\IntlDateFormatter::FULL, null, 'en-KW@calendar=islamic');
-        $expected = 'Thursday, Muharram 29, 1431 at 1:59:28 PM GMT';
-        $this->assertTimeFormat($expected, $result);
-
-        $result = $time->i18nFormat(\IntlDateFormatter::FULL, 'Asia/Tokyo', 'ja-JP@calendar=japanese');
-        $expected = '平成22年1月14日木曜日 22時59分28秒 日本標準時';
```
markstory, line 479: Should we have some tests left behind for non-gregorian calendars?

othercorey: Yea, we should since we parse it to set the traditional calendar flag.

PR #15328, opened 2021-02-11
Allow integers to be passed to command table helper

### LINKED ISSUE #15327: Command Table Helper - Passing integers to cells
This is a (multiple allowed):

* [x] bug
* [ ] enhancement

* CakePHP Version: EXACT RELEASE VERSION OR COMMIT HASH, HERE.

### What you did
I have a console command for viewing members in my project. In the command I use the built-in table helper to output the entities with their properties and values. The entity contains integers, as well as strings.
```php
{
    $members = [
        ['Id', 'Name', 'Age'],
        [1, 'Jane', 27],
        [2, 'John', 25],
    ];

    $this->io->helper('Table')->output($members);
}
```

### What happened
I get an exception saying the argument passed to `TableHelper::_cellWidth()` must be of type string or null.
```
2021-02-11 16:14:39 Error: [TypeError] Argument 1 passed to Cake\Shell\Helper\TableHelper::_cellWidth() must be of the type string or null, int given, 
called in /vendor/cakephp/cakephp/src/Shell/Helper/TableHelper.php on line 49 in /vendor/cakephp/cakephp/src/Shell/Helper/TableHelper.php on line 65
Stack Trace:
- /vendor/cakephp/cakephp/src/Shell/Helper/TableHelper.php:49
- /vendor/cakephp/cakephp/src/Shell/Helper/TableHelper.php:150
- /src/Command/EditMemberCommand.php:68
- /vendor/cakephp/cakephp/src/Console/BaseCommand.php:179
- /vendor/cakephp/cakephp/src/Console/CommandRunner.php:336
- /vendor/cakephp/cakephp/src/Console/CommandRunner.php:172
- /bin/cake.php:12
```

I expected the helper to either accept integers or cast them into strings.

I'm unsure whether you guys qualify this as unexpected behavior. Sure, I could cast them manually myself before passing them to the helper, but that shouldn't be necessary *in my opinion*.

I believe passing integers to the table worked fine until this commit change: https://github.com/cakephp/cakephp/commit/9fc0363079d08f8ecefd5394c919ae5edd6951b2

mentisy:
## Command Table Helper - Integers
Passing integers to the Command/Shell table helper throws exceptions. With this PR they will now cast integers to string, allowing them to be output properly.

Closes #15327

### DIFF
--- a/src/Shell/Helper/TableHelper.php
+++ b/src/Shell/Helper/TableHelper.php
@@ -59,12 +59,12 @@ class TableHelper extends Helper
     /**
      * Get the width of a cell exclusive of style tags.
      *
-     * @param string|null $text The text to calculate a width for.
+     * @param string $text The text to calculate a width for.
      * @return int The width of the textual content in visible characters.
      */
-    protected function _cellWidth(?string $text): int
+    protected function _cellWidth(string $text): int
     {
-        if ($text === null) {
+        if (strlen($text) === 0) {
             return 0;
         }
 
@@ -116,7 +116,7 @@ class TableHelper extends Helper
             if (!empty($options['style'])) {
                 $column = $this->_addStyle($column, $options['style']);
             }
-            if ($column !== null && preg_match('#(.*)<text-right>.+</text-right>(.*)#', $column, $matches)) {
+            if (strlen($column) > 0 && preg_match('#(.*)<text-right>.+</text-right>(.*)#', $column, $matches)) {
                 if ($matches[1] !== '' || $matches[2] !== '') {
                     throw new UnexpectedValueException('You cannot include text before or after the text-right tag.');
                 }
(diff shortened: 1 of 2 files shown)


src/Shell/Helper/TableHelper.php
```diff
@@ -46,6 +46,7 @@ protected function _calculateWidths(array $rows): array
         $widths = [];
         foreach ($rows as $line) {
             foreach (array_values($line) as $k => $v) {
+                $v = is_integer($v) ? (string)$v : $v;
```
dereuromark, line 49: What about numeric values that would be castable?
I wonder if it should just always cast.
If it doesnt work it would thow an exception anyway.

mentisy: I was unsure whether that was the right thing, but I agree. Have committed a change.

mentisy: Always casting threw a new wrench in the works. Should we resolve the psalm issue by removing the nullable string parameter?

markstory: > Always casting threw a new wrench in the works. Should we resolve the psalm issue by removing the nullable parameter text in the _cellWidth method? That also means replacing the null checks with strlen checks instead.

That sounds like a good idea to me.

markstory:
Thank you 👍

PR #15331, opened 2021-02-17
`ServerRequest::is()`: `$args` is mixed

mirko-pagliai:
For the `ServerRequest::is()` method, now `$args` is described as a list of strings, but there is no reason why this should be so and why other types of value (array, integer, etc.) cannot be passed.


garas:
Please rebase your branch, there is too many merge commits.

othercorey:
This is still a draft.

othercorey:
@markstory can you check this?

PR #15338, opened 2021-02-22
Error handling is logging silenced (@) errors in PHP8. Issue #15336

### LINKED ISSUE #15336: Error handling is logging silenced (@) errors in PHP8
This is a (multiple allowed):

* [x] bug

* CakePHP Version: 4.2.3
* Platform and Target: PHP8

### What you did
Triggered an error that is silenced by @ when using PHP8

### What happened
It logged the silenced error to error.log that should not be logged. Works properly in PHP7.4.

It should not have logged the error.

The fix:
Error handlers that expect error_reporting to be 0 when @ is used, should be adjusted to use a mask check instead:
https://www.php.net/manual/en/migration80.incompatible.php

Can be fixed by replacing
"if (error_reporting() === 0) {"
with
"if (!(error_reporting() & $code)) {"

at line 151 in BaseErrorHandler.php.

AdmDevelopment:
As requested by Mark Story, I have created a pull request for the issue:
Error handling is logging silenced (@) errors in PHP8. Issue #15336

### DIFF
--- a/src/Error/BaseErrorHandler.php
+++ b/src/Error/BaseErrorHandler.php
@@ -148,7 +148,7 @@ abstract class BaseErrorHandler
         ?int $line = null,
         ?array $context = null
     ): bool {
-        if (error_reporting() === 0) {
+        if (!(error_reporting() & $code)) {
             return false;
         }
         $this->_handled = true;


othercorey:
Can we add a unit test for this?

src/Error/BaseErrorHandler.php
```diff
@@ -148,7 +148,7 @@ public function handleError(
         ?int $line = null,
         ?array $context = null
     ): bool {
-        if (error_reporting() === 0) {
+        if (!(error_reporting() & $code)) {
```
garas, line 151:
```suggestion
        if (error_reporting() & $code === 0) {
```

Negation makes expression harder to understand.

ADmad: Note that it's a bitwise operator `&`, not a logical operator `&&`.

AdmDevelopment: I used the bitwise operator as suggested by PHP at:
https://www.php.net/manual/en/migration80.incompatible.php

garas: Bitwise operation result is integer, so it is better to compare with integer.  `!$integer` is same as `$integer === 0`, but it makes more clear that value 0 is (un)wanted.

othercorey: I would put () around the & operation then.

AdmDevelopment: I just ran and debugged the code. It appears $code ($errno) is never 0 in PHP8:
https://www.php.net/manual/en/function.set-error-handler.php

I think we're stuck with the bitwise since this is backwards compatible with older PHP versions. We could however wrap the entire code block in "if (error_reporting() & $code) {" if we absolutely need to get rid of the negation.

FYI: The vendors Composer & PHPUnit also use equal bitwise operation with negation to handle this.

ADmad: @garas I tried locally and there's a test case failure with your suggested change.

```
1) Cake\Test\TestCase\Error\ErrorHandlerTest::testErrorSuppressed
Failed asserting that a string is empty.

../cakephp/tests/TestCase/Error/ErrorHandlerTest.php:224
```

markstory: Sounds like we have existing coverage for this change then?

ADmad: Guess you could say that :).

othercorey: Did you put () around the bitwise operation?

ADmad: No, I just tried exactly was garas has suggested.

AdmDevelopment: I have tried paras' code with and without () around the bitwise operation. It still prints / logs the silenced errors.

I've been running the original code from my pull request in local development/staging since last Friday without any issues.

Issue #15339, opened 2021-02-22
Entity::isEmpty() behaves differently than method's comments


mtak3:
* [x] bug

* CakePHP Version: 3.x, 4.x
* Platform and Target: 

### What you did
`Entity :: isEmpty()` behaves differently than method's comments
```
     * This is not working like the PHP `empty()` function. The method will
     * return true for:
     *
     * - `''` (empty string)
     * - `null`
     * - `[]`
     *
     * and false in all other cases.
```

### What happened
`"0"` is true

### What you expected to happen
`"0"` is false
https://github.com/cakephp/cakephp/pull/10709#discussion_r137542451

dereuromark:
Are you able to make a PR with the suggested changes?

markstory:
Closing as this has been fixed on 3.x and master

PR #15340, opened 2021-02-22
Entity::isEmpty() behaves differently than method's comments

mtak3:
https://github.com/cakephp/cakephp/issues/15339

### DIFF
--- a/src/Datasource/EntityTrait.php
+++ b/src/Datasource/EntityTrait.php
@@ -394,7 +394,7 @@ trait EntityTrait
                 empty($value) ||
                 (
                     is_string($value) &&
-                    empty($value)
+                    strlen($value) === 0
                 )
             )
         ) {
--- a/tests/TestCase/ORM/EntityTest.php
+++ b/tests/TestCase/ORM/EntityTest.php
@@ -1761,6 +1761,7 @@ class EntityTest extends TestCase
             'emptyArray' => [],
             'object' => new \stdClass(),
             'string' => 'string',
+            'stringZero' => '0',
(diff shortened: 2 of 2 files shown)


src/Datasource/EntityTrait.php
```diff
@@ -394,7 +394,7 @@ public function isEmpty(string $field): bool
                 empty($value) ||
                 (
                     is_string($value) &&
-                    empty($value)
+                    strlen($value) === 0
```
dereuromark, line 397: I wonder if with longer strings and lots of ops it would be a quicker nano-opt. to use `!== '0'` instead on top.
Lets see what others think.

ADmad: Given that there's already a `is_string()` check first, `!== ''` check would be better.

markstory, line 397:
```suggestion
                    $value === ''
```

othercorey:
Is this change ok for master or should it go into 4.next?

markstory:
It looks like this method has never worked the way it was documented. The method was introduced in 31ba1a33c82   and the '0' case was missed. Given the intent of the method I think this fits the scope of a bugfix, and is something we should backport to 3.x as well.

src/Datasource/EntityTrait.php
othercorey: @mtak3 You need to make this change before we can merge. I would make it, but you didn't allow maintainers to make changes.

markstory: I'll fix this up post merge.

PR #15343, opened 2021-02-25
Simplify wrapping join tables expressions in parentheses

othercorey:
Committing this cleanup before fixing nested joins so tests are run separately.


othercorey:
@markstory fyi. merged this since I was going to work in 4.2.

PR #15344, opened 2021-02-25
Remove invalid options from attachTo

othercorey:
This avoids putting invalid keys into `$options` and cleans up the docblock. Looks like a couple old commits partly removed some lines.

I added checks for `type` option when running query and association tests. Nothing was passing this invalid option in that I could find.

### DIFF
--- a/src/ORM/Association.php
+++ b/src/ORM/Association.php
@@ -696,8 +696,6 @@ abstract class Association
      * - conditions: array with a list of conditions to filter the join with, this
      *   will be merged with any conditions originally configured for this association
      * - fields: a list of fields in the target table to include in the result
-     * - type: The type of join to be used (e.g. INNER)
-     *   the records found on this association
      * - aliasPath: A dot separated string representing the path of association names
      *   followed from the passed query main table to this association.
      * - propertyPath: A dot separated string representing the path of association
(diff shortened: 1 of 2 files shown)


src/ORM/Association.php
```diff
@@ -710,21 +708,19 @@ protected function _options(array $options): void
...
         $options += [
             'includeFields' => true,
             'foreignKey' => $this->getForeignKey(),
             'conditions' => [],
+            'joinType' => $this->getJoinType(),
             'fields' => [],
-            'type' => $joinType,
             'table' => $table,
             'finder' => $this->getFinder(),
         ];
```
othercorey, line 726: This doesn't set defaults for all allowed options. I believe other helper functions check those, but will pull the defaults up to here in future changes if possible.

src/ORM/Association.php
```diff
@@ -764,9 +760,11 @@ public function attachTo(Query $query, array $options = []): void
         $dummy->where($options['conditions']);
         $this->_dispatchBeforeFind($dummy);
 
-        $joinOptions = ['table' => 1, 'conditions' => 1, 'type' => 1];
-        $options['conditions'] = $dummy->clause('where');
-        $query->join([$this->_name => array_intersect_key($options, $joinOptions)]);
+        $query->join([$this->_name => [
+            'table' => $options['table'],
+            'conditions' => $dummy->clause('where'),
+            'type' => $options['joinType'],
```
markstory, line 766: Why not use join everywhere?

othercorey: How do you mean?

markstory: My mistake, I got mixed up. This can't be `joinType` because it is a join on the query, and in the above hunk you're normalizing on `joinType` as that is the key used in `Association`.

PR #15345, opened 2021-02-27
Use newEntity() to create DatabaseSession entities

othercorey:

### DIFF
--- a/src/Http/Session/DatabaseSession.php
+++ b/src/Http/Session/DatabaseSession.php
@@ -18,7 +18,6 @@ declare(strict_types=1);
  */
 namespace Cake\Http\Session;
 
-use Cake\ORM\Entity;
 use Cake\ORM\Locator\LocatorAwareTrait;
 use SessionHandlerInterface;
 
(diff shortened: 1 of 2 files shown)


markstory:
Something is up with one of the delete tests.

othercorey:
Yea, I don't know what it is. I had this error locally before committing.

othercorey:
You can't reference an existing entity using `Table::newEntity()` because it doesn't pass the options to `Entity` constructor.

src/Http/Session/DatabaseSession.php
```diff
@@ -150,14 +149,16 @@ public function write($id, $data): bool
...
+
         /** @var string $pkField */
         $pkField = $this->_table->getPrimaryKey();
-        $record[$pkField] = $id;
-        $result = $this->_table->save(new Entity($record));
+        $session = $this->_table->newEntity([
+            $pkField => $id,
+            'data' => $data,
+            'expires' => time() + $this->_timeout,
+        ], ['guard' => false]);
```
othercorey, line 159: You can't use `guard` option with `newEntity()` either since it's an Entity constructor option.

PR #15347, opened 2021-02-27
Update psalm and phpstan's version.

ADmad:


ADmad:
The phar generation for latest psalm release is borked so need to wait for that to be fixed.

PR #15348, opened 2021-02-27
Use newEntity() to create DatabaseSession entities

othercorey:


src/Http/Session/DatabaseSession.php
```diff
@@ -150,14 +149,16 @@ public function write($id, $data): bool
...
+
         /** @var string $pkField */
         $pkField = $this->_table->getPrimaryKey();
-        $record[$pkField] = $id;
-        $result = $this->_table->save(new Entity($record));
+        $session = $this->_table->newEntity([
+            $pkField => $id,
+            'data' => $data,
+            'expires' => time() + $this->_timeout,
+        ], ['accessibleFields' => ['id' => true]]);
```
markstory, line 159: Should this also be pkField to match the key in the entity data?

othercorey: good catch, thanks.

PR #15349, opened 2021-02-27
4.x - Fix file cache engine trying to clear root on realpath failure.

ndm2:
When `getRealPath()` fails and returns `false`, the path will finally
contain only the appended directory separator, thus pointing to the root
directory.

I've stumbled over this via a **[question on StackOverflow](https://stackoverflow.com/questions/66352205/how-to-debug-cakephp-3-5-cache-basedir-issue)**, where clearing the cache "randomly" tries to clear the root directory (`/`). This might not actually be the source of the problem that the OP is experiencing, but I would still like to see this possible problem patched.

I can try to add a test for this if you like, it would require a little refactoring for obtaining the iterator and some creative mocking I guess.

### DIFF
--- a/src/Cache/Engine/FileEngine.php
+++ b/src/Cache/Engine/FileEngine.php
@@ -251,12 +251,18 @@ class FileEngine extends CacheEngine
             RecursiveIteratorIterator::SELF_FIRST
         );
         $cleared = [];
-        foreach ($contents as $path) {
-            if ($path->isFile()) {
+        /** @var \SplFileInfo $fileInfo */
+        foreach ($contents as $fileInfo) {
+            if ($fileInfo->isFile()) {
                 continue;
             }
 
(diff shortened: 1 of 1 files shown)


src/Cache/Engine/FileEngine.php
```diff
@@ -256,7 +256,12 @@ public function clear(): bool
                 continue;
             }
 
-            $path = $path->getRealPath() . DIRECTORY_SEPARATOR;
+            $realPath = $path->getRealPath();
```
garas, line 259: Here and above this line `$path` is `SplFileInfo` and `string` below.

Rename it in upper part to something like `$fileinfo`.

ndm2: done

markstory:
I don't think the mock maintenance cost is worth it here.

othercorey:
Backported to 3.x in 9c8cd8c

PR #15350, opened 2021-02-28
Simplify disabling of association target fields

othercorey:
`Query::*joinWith()` functions call `setMatching()` with `fields => false` to disable including all the target table fields.

This moves the configuration up to `attachTo()` and simplifies the appending of fields.

### DIFF
--- a/src/ORM/Association.php
+++ b/src/ORM/Association.php
@@ -725,6 +725,12 @@ abstract class Association
             'finder' => $this->getFinder(),
         ];
 
+        // This is set by joinWith to disable matching results
+        if ($options['fields'] === false) {
+            $options['fields'] = [];
+            $options['includeFields'] = false;
+        }
+
         if (!empty($options['foreignKey'])) {
             $joinCondition = $this->_joinCondition($options);
             if ($joinCondition) {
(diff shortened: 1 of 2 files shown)


tests/TestCase/ORM/Association/HasOneTest.php
```diff
@@ -173,8 +173,7 @@ public function testAttachToMultiPrimaryKey()
                 'table' => 'profiles',
             ],
         ]);
-        $query->expects($this->never())->method('select');
-        $association->attachTo($query, ['includeFields' => false]);
```
othercorey, line 177: This test is unrelated to `includeFields`. The test passes with our without this setting and avoiding the `select()` call is irrelevant. Don't know if this was copied from somewhere else.

src/ORM/Association.php
```diff
@@ -954,25 +960,17 @@ protected function _appendFields(Query $query, Query $surrogate, array $options)
...
+            $surrogate->isAutoFieldsEnabled()
+        ) {
+            $fields = array_merge($fields, $this->_targetTable->getSchema()->columns());
         }
 
-        if ($fields) {
-            $query->select($query->aliasFields($fields, $this->_name));
-        }
-        $query->addDefaultTypes($target);
+        $query->select($query->aliasFields($fields, $this->_name));
```
othercorey, line 972: I don't know if we *need* to select an alias the original fields here or not.

PR #15352, opened 2021-02-28
Use null coalescing operator where possible.

ADmad:


markstory:
Great line savings in this one :clap:

PR #15353, opened 2021-03-01
DATE_RFC7231

ADmad:
`D, j M Y G:i:s` generates invalid format. `j` and `G` don't add leading zeros as required by the [RFC7231](https://tools.ietf.org/html/rfc7231#section-7.1.1.1) datetime format.

This fix needs to be backported to 3.x too, but constant `DATE_RFC7231` doesn't exist in older PHP versions so there `D, d M Y H:i:s` will have to be used.

### DIFF
--- a/tests/TestCase/Http/ResponseTest.php
+++ b/tests/TestCase/Http/ResponseTest.php
@@ -301,7 +301,7 @@ class ResponseTest extends TestCase
         $response = new Response();
         $expected = [
             'Expires' => ['Mon, 26 Jul 1997 05:00:00 GMT'],
-            'Last-Modified' => [gmdate('D, d M Y H:i:s') . ' GMT'],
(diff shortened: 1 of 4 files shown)


tests/TestCase/Http/ResponseTest.php
```diff
@@ -325,9 +325,9 @@ public function testWithCache()
         $this->assertFalse($response->hasHeader('Date'));
         $this->assertFalse($response->hasHeader('Last-Modified'));
 
-        $this->assertSame(gmdate('D, j M Y G:i:s ', $since) . 'GMT', $new->getHeaderLine('Date'));
```
othercorey, line 328: I think we should just rely on the constant in test cases as well.

If we wanted to test a specifically formatted date, we should start with the raw date string, convert to time and then compare that.

ADmad: Done

PR #15354, opened 2021-03-02
Simplify generating nested contains in setMatching()

othercorey:
This gets rid of confusing variable names like `$pointer` and documents the nested contain traversal.

Even though `Query` calls `setMatching()` with 3 options, `Association` will actually call `setMatching()` with other associations from itself. We cannot simplify the `$options` passed in just yet.

### DIFF
--- a/src/ORM/EagerLoader.php
+++ b/src/ORM/EagerLoader.php
@@ -230,41 +230,37 @@ class EagerLoader
      *
      *  ### Options
      *
-     *  - 'joinType': INNER, OUTER, ...
-     *  - 'fields': Fields to contain
+     *  - `joinType`: INNER, OUTER, ...
+     *  - `fields`: Fields to contain
+     *  - `negateMatch`: Whether to add conditions negate match on target association
      *
-     * @param string $assoc A single association or a dot separated path of associations.
+     * @param string $path Dot separated association path
      * @param callable|null $builder the callback function to be used for setting extra
      * options to the filtering query
      * @param array $options Extra options for the association matching.
      * @return $this
      */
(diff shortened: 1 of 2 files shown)


src/ORM/EagerLoader.php
```diff
@@ -230,41 +230,37 @@ public function isAutoFieldsEnabled(): bool
...
      *
-     * @param string $assoc A single association or a dot separated path of associations.
+     * @param string $path Dot separated association path
      * @param callable|null $builder the callback function to be used for setting extra
      * options to the filtering query
      * @param array $options Extra options for the association matching.
      * @return $this
      */
-    public function setMatching(string $assoc, ?callable $builder = null, array $options = [])
+    public function setMatching(string $path, ?callable $builder = null, array $options = [])
```
ADmad, line 243: `$assocPath` to be more clearer?

othercorey: Sounds good.

src/ORM/EagerLoader.php
```diff
@@ -230,41 +230,37 @@ public function isAutoFieldsEnabled(): bool
...
+            $nested[$association] = $sharedOptions;
+            // Set to next nested level
+            $nested = &$nested[$association];
         }
 
-        $pointer[$last] = ['queryBuilder' => $builder, 'matching' => true] + $options;
-
-        $this->_matching->contain($containments);
+        // Add all options to target association contain which is the last in nested chain
+        $nested = ['matching' => true, 'queryBuilder' => $builder] + $options;
```
othercorey, line 262: Instead of trying to merge with shared options, just set the valid options here. Only so much I can do to clarify this before moving on.

PR #15357, opened 2021-03-03
Use TRUE and FALSE instead of 1 and 0 for postgres and sqlite

othercorey:
Fixes https://github.com/cakephp/debug_kit/issues/508

This makes the SQL valid when copying to analyze.

### DIFF
--- a/src/Database/Log/LoggedQuery.php
+++ b/src/Database/Log/LoggedQuery.php
@@ -16,6 +16,7 @@ declare(strict_types=1);
  */
 namespace Cake\Database\Log;
 
+use Cake\Database\Driver\Sqlserver;
 use JsonSerializable;
 
 /**
@@ -26,6 +27,13 @@ use JsonSerializable;
  */
 class LoggedQuery implements JsonSerializable
 {
+    /**
+     * Driver executing the query
+     *
+     * @var \Cake\Database\DriverInterface|null
+     */
+    public $driver = null;
+
     /**
      * Query string that was executed
      *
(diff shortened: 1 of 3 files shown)


src/Database/Log/LoggedQuery.php
```diff
@@ -73,7 +82,12 @@ protected function interpolate(): string
             if ($p === null) {
                 return 'NULL';
             }
+
             if (is_bool($p)) {
+                if ($this->driver instanceof Postgres || $this->driver instanceof Sqlite) {
```
ADmad, line 87: Wouldn't it be better to check for `Mysql` and return 0/1 and default to `TRUE`/`FALSE`?

markstory: I think MySQL will also accept TRUE/FALSE.

othercorey: SqlServer uses `bit` which accepts only 1 or the string 'TRUE' and not the literal TRUE. It was an even split.

othercorey: Just tested and you are right that MySQL accepts `TRUE` literal.

src/Database/Log/LoggedQuery.php
```diff
@@ -73,8 +81,13 @@ protected function interpolate(): string
             if ($p === null) {
                 return 'NULL';
             }
+
             if (is_bool($p)) {
-                return $p ? '1' : '0';
+                if ($this->driver instanceof Sqlserver) {
+                    return $p ? '1' : '0';
```
ADmad, line 87: Seems we need to update the GA workflow to push coverage report for windows/sqlserver build too.

PR #15364, opened 2021-03-04
Detect infinite recursion when initializing Table

### LINKED ISSUE #15351: Self-associated table may cause an infinite loop
This is a (multiple allowed):

* [x] bug

* CakePHP Version: 4.2.3.
* Platform and Target: php 7.4.15, docker image: webdevops/php-nginx:7.4

### What you did
I have created a Table inside a plugin that referces itself in a parent child relationship

```php
<?php
declare(strict_types=1);

namespace Domain\Model\Table;

// ... some lines removed to keep it simple (uses, etc)

class PersonasTable extends Table
{
    public function initialize(array $config): void
    {
        parent::initialize($config);
        // ... 

        $this->belongsTo('Titulares', [
            'foreignKey' => 'titular_id',
            'className' => 'Domain.Personas',
        ]);
        //...
    }
    //...
}
```

### What happened
Any function call chained to that belongsTo that internally fires a getTarget() call, like calling functions proxied to the target Table result in an infinite loop. For example:
 
```php
        $this->belongsTo('Titulares', [
            'foreignKey' => 'titular_id',
            'className' => 'Domain.Personas',
        ])->setEntityClass('Domain.Persona');
```

othercorey:
Closes https://github.com/cakephp/cakephp/issues/15351

Users can make the same calls in  `initialize()` since it's the same Table instance.

An exception throw in Table constructor will cause the initializing flag to remain set, but it seems the code is unable to use that Table class anyway.

### DIFF
--- a/src/ORM/Locator/TableLocator.php
+++ b/src/ORM/Locator/TableLocator.php
@@ -53,6 +53,13 @@ class TableLocator extends AbstractLocator implements LocatorInterface
     protected $instances = [];
 
     /**
+     * List of Table classes being initialized.
+     *
+     * @var array<string, bool>
+     */
+    protected $initializing = [];
+
+    /**
      * Contains a list of Table objects that were created out of the
      * built-in Table class. The list is indexed by table alias
      *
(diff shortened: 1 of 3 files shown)


src/ORM/Locator/TableLocator.php
```diff
@@ -226,6 +238,10 @@ protected function createInstance(string $alias, array $options)
         $allowFallbackClass = $options['allowFallbackClass'] ?? $this->allowFallbackClass;
         $className = $this->_getClassName($alias, $options);
         if ($className) {
+            if (!empty($this->initializing[$className])) {
```
ADmad, line 241:
```suggestion
            if (isset($this->initializing[$className])) {
```

markstory (approved):

PR #15368, opened 2021-03-06
Use passed-in class name when clearing initializing list

othercorey:
This fixes an edge case where the user looks up a table using a fully qualified class name. `get_class()` only returns a qualified name without the leading `\`.

This means a scenarios exists where infinite recursion is not detected due to a FQCN get() and then a QCN get while initializing.

Seems rare enough to keep the code simple.

### DIFF
--- a/src/ORM/Locator/TableLocator.php
+++ b/src/ORM/Locator/TableLocator.php
@@ -208,13 +208,8 @@ class TableLocator extends AbstractLocator implements LocatorInterface
      */
     public function get(string $alias, array $options = []): Table
     {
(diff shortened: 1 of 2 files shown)


ndm2:
Just to make sure there's no misunderstanding, names in strings are always treated as fully qualified, no leading slash required.

othercorey:
@ndm2 This isn't a change in functionality. There was one place in 3.x tests where the `className` option had a leading slash. Because get_class() didn't return a leading slash, the flag wasn't cleared which meant the next call to `get()` throws an exception.

PR #15371, opened 2021-03-06
4.next - Add support for schema column SQL/conversion via database types.

ndm2:
Better late than never I guess. I thought it would be best to split this off from #14993, as this seems to be the less "controversial" feature.

### DIFF
--- a/src/Database/Schema/PostgresSchemaDialect.php
+++ b/src/Database/Schema/PostgresSchemaDialect.php
@@ -90,7 +90,7 @@ class PostgresSchemaDialect extends SchemaDialect
             $length = (int)$matches[2];
         }
 
-        $type = $this->_convertColumnViaType($col, compact('length'));
+        $type = $this->_applyTypeSpecificColumnConversion($col, compact('length'));
         if ($type !== null) {
             return $type;
         }
@@ -385,7 +385,7 @@ class PostgresSchemaDialect extends SchemaDialect
         /** @var array $data */
         $data = $schema->getColumn($name);
 
-        $sql = $this->_getColumnSqlViaType($data['type'], $schema, $name);
+        $sql = $this->_getTypeSpecificColumnSql($data['type'], $schema, $name);
         if ($sql !== null) {
             return $sql;
         }
--- a/src/Database/Schema/PostgresSchemaDialect.php
+++ b/src/Database/Schema/PostgresSchemaDialect.php
@@ -85,12 +85,15 @@ class PostgresSchemaDialect extends SchemaDialect
         }
 
         $col = strtolower($matches[1]);
-        $length = null;
+        $length = $precision = $scale = null;
         if (isset($matches[2])) {
             $length = (int)$matches[2];
         }
 
-        $type = $this->_applyTypeSpecificColumnConversion($col, compact('length'));
+        $type = $this->_applyTypeSpecificColumnConversion(
+            $col,
+            compact('length', 'precision', 'scale')
+        );
         if ($type !== null) {
             return $type;
         }
--- a/src/Database/Schema/PostgresSchemaDialect.php
+++ b/src/Database/Schema/PostgresSchemaDialect.php
@@ -90,6 +90,11 @@ class PostgresSchemaDialect extends SchemaDialect
             $length = (int)$matches[2];
         }
 
+        $type = $this->_convertColumnViaType($col, compact('length'));
+        if ($type !== null) {
+            return $type;
+        }
+
         if (in_array($col, ['date', 'time', 'boolean'], true)) {
             return ['type' => $col, 'length' => null];
         }
@@ -379,6 +384,12 @@ class PostgresSchemaDialect extends SchemaDialect
     {
         /** @var array $data */
         $data = $schema->getColumn($name);
+
+        $sql = $this->_getColumnSqlViaType($data['type'], $schema, $name);
+        if ($sql !== null) {
+            return $sql;
+        }
+
         $out = $this->_driver->quoteIdentifier($name);
         $typeMap = [
             TableSchema::TYPE_TINYINTEGER => ' SMALLINT',
--- a/src/Database/Schema/SchemaDialect.php
+++ b/src/Database/Schema/SchemaDialect.php
@@ -17,6 +17,8 @@ declare(strict_types=1);
 namespace Cake\Database\Schema;
 
 use Cake\Database\DriverInterface;
+use Cake\Database\Type\SchemaAwareTypeInterface;
+use Cake\Database\TypeFactory;
 use InvalidArgumentException;
 
 /**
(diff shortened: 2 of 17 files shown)


ndm2:
Guess the tests need a master > 4.next merge.

othercorey:
> Guess the tests need a master > 4.next merge.

done.

tests/TestCase/Database/SchemaAwareTypeTest.php
```diff
@@ -0,0 +1,205 @@
...
+                $this->convertColumnDescriptionPostgresTest();
+                break;
+
+            case Sqlite::class:
+                $this->convertColumnDescriptionSqliteTest();
+                break;
+
+            case Sqlserver::class:
+                $this->convertColumnDescriptionSqlserverTest();
+                break;
```
markstory, line 86: Should this have an exception for the default case?

tests/TestCase/Database/SchemaAwareTypeTest.php
```diff
@@ -0,0 +1,205 @@
...
+            'null' => '1',
+            'default' => 'NULL',
+            'char_length' => null,
+            'precision' => null,
+            'scale' => null,
+            'collation_name' => 'Latin1_General_CI_AI',
+        ]);
+
+        $column = $schema->getColumn('field');
+        $this->assertEquals(
```
markstory, line 193: Do you need these tests if you also have the integration tests? Couldn't the comment, length and column type assertions be done for all db vendors in the integration tests as well?

markstory (commented):
Nice work! This will be helpful to map value objects through the ORM :clap:

tests/TestCase/Database/SchemaAwareTypeTest.php
ndm2: Kind of, maybe... problem is that in order to get a simple integration test running a real column type needs to be used. I've opted to use `TEXT` here, as it works for all dialects, that way I could avoid creating all sorts of custom types for the different dialects, but the side effect was that `convertSchemaColumn()` wouldn't be invoked.

Maybe if I remap the `text` type instead of mapping to `schemaawaretype`. I'll give that a try. There would still need to be some vendor specific code though, as only MySQL supports actual comment SQL.

src/Database/Type/SchemaAwareTypeInterface.php
```diff
@@ -0,0 +1,29 @@
+<?php
+declare(strict_types=1);
+
+namespace Cake\Database\Type;
+
+use Cake\Database\DriverInterface;
+use Cake\Database\Schema\TableSchemaInterface;
+
+interface SchemaAwareTypeInterface
```
othercorey, line 9: Since we have `TableSchemaAwareInterface` should this be `ColumnSchemaAwareInterface`?

I think we should put the type first not at the end so we don't name interfaces like `SomethingForSomethingInterface`.

src/Database/Schema/SchemaDialect.php
```diff
@@ -112,6 +114,53 @@ protected function _convertConstraintColumns($references): string
...
+     * Tries to use a matching database type to generate the SQL
+     * fragment for a single column in a table.
+     *
+     * @param string $columnType The column type.
+     * @param \Cake\Database\Schema\TableSchemaInterface $schema The table schema instance the column is in.
+     * @param string $column The name of the column.
+     * @return string|null An SQL fragment, or `null` in case no corresponding type was found or the type didn't provide
+     *  custom column SQL.
+     */
+    protected function _getColumnSqlViaType(string $columnType, TableSchemaInterface $schema, string $column): ?string
```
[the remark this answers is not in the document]
othercorey: I'm not a big fan of using names like `Via`.

src/Database/Type/SchemaAwareTypeInterface.php
ndm2: But is there such a thing as a "column schema"? Aren't columns more like just a part of the "table schema"?

As for the `TypeInterface` trail, I was just following `ExpressionTypeInterface`, but on a second thought this was probably only chosen to avoid confusion with the existing `ExpressionInterface`.

othercorey: I think column types that have a specific expression format have a schema. Consider the GIS types.

tests/TestCase/Database/SchemaAwareTypeTest.php
markstory: > There would still need to be some vendor specific code though, as only MySQL supports actual comment SQL.

I'm not super concerned with preserving schema comments in custom types. That feels like something we could cover in the docs.

src/Database/Type/SchemaAwareTypeInterface.php
ndm2: I couldn't find too much about it, but there's references to column definitions as schemas in for example SQL Server and .NET, so I guess it's kind of a thing.

othercorey:
> Not sure what this memory leak problem with SQL Server is about, I'd like to imagine it's not related to my changes.

It randomly fails on any configuration.

markstory:
I'll get this merged this weekend after I write some docs.

othercorey:
Looks ok, but I prefer to not have function names with `Via` in them. I commented on one, but was hoping the other would be updated.

ndm2:
> Looks ok, but I prefer to not have function names with `Via` in them. I commented on one, but was hoping the other would be updated.

It's better now... I hope :)

markstory:
Docs pull request is up now https://github.com/cakephp/docs/pull/6964

src/Database/Type/ColumnSchemaAwareInterface.php
```diff
@@ -0,0 +1,29 @@
...
+    public function getColumnSql(TableSchemaInterface $schema, string $column, DriverInterface $driver): ?string;
+
+    /**
+     * Convert a SQL column definition to an abstract type definition.
+     *
+     * @param array $definition The column definition.
+     * @param \Cake\Database\DriverInterface $driver The driver instance being used.
+     * @return array|null Array of column information, or `null` in case the column isn't processed by this type.
+     */
+    public function convertColumnDefinition(array $definition, DriverInterface $driver): ?array;
```
othercorey, line 28: I'm concerned this is not well-defined. The user does not know what `$definition` contains even if it is driver-specific.

We also don't define what the expected result should be. Should the array contain all the keys that SchemaInterface::addColumn() accepts?

ndm2: > Should the array contain all the keys that SchemaInterface::addColumn() accepts?

Basically yes, according to the column type and driver.

othercorey: I think we need to document this as well as document what to expect from each driver. At the moment, driver-specific keys are an implementation detail that we are exposing.

src/Database/Schema/SchemaDialect.php
```diff
@@ -112,6 +114,56 @@ protected function _convertConstraintColumns($references): string
...
+    protected function _getTypeSpecificColumnSql(
+        string $columnType,
+        TableSchemaInterface $schema,
+        string $column
+    ): ?string {
+        if (!TypeFactory::getMap($columnType)) {
+            return null;
+        }
+
+        $type = TypeFactory::build($columnType);
```
markstory, line 136: I was working on update the docs to be more clear on what will be inside `$definition`, and this line made me think. How is this supposed to work with custom types? Is the goal to only handle custom types that overwrite an existing 'basic' type, or handles a column type that the ORM cannot currently handle?

src/Database/Schema/PostgresSchemaDialect.php
```diff
@@ -379,6 +384,12 @@ public function columnSql(TableSchema $schema, string $name): string
     {
         /** @var array $data */
         $data = $schema->getColumn($name);
+
+        $sql = $this->_getTypeSpecificColumnSql($data['type'], $schema, $name);
```
markstory, line 388: Couldn't a custom type also need other attributes in the abstract schema? Keys like length, precision and null could be useful in a custom type.

src/Database/Schema/SchemaDialect.php
ndm2: Both. The latter was kinda shown before my change to simplify the tests, problem was that generating `TEXT` in the type would cause the made up `schemaawaretype` type to be lost between `getColumnSql()` and `convertColumnDefinition()`, so I've chosen to remap the existing type.

I could add an additional driver specific type, like for example `POINT`, and use that for MySQL/Postgres-only integration tests that show handling types unknown to CakePHP? Maybe I could even get it to work with SQL Server, but that requires a different column type IIRC, `GEOGRAPHY` I think.

markstory: I'm not sure the test needs to cover the a DB specific type as it complicates maintenance.

Issue #15372, opened 2021-03-07
Error: array_merge() does not accept unknown named parameters


antonvolokha:
* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 2.10.24 commit: cf14e6546ec44e33
* Platform and Target: docker php:8-apache

### What you did
Run console/cake bake fixture
Then choose my model, after that answer on all questions

### What happened
`
Error: array_merge() does not accept unknown named parameters
#0 /var/www/html/lib/Cake/Utility/Folder.php(230): array_merge(/var/www/html/lib/Cake/Console/Templates/default/actions: Array, /var/www/html/lib/Cake/Console/Templates/default/classes: Array, /var/www/html/lib/Cake/Console/Templates/default/views: Array)
#1 /var/www/html/lib/Cake/Console/Command/Task/TemplateTask.php(75): Folder->read()
#2 /var/www/html/lib/Cake/Console/Command/Task/TemplateTask.php(50): TemplateTask->_findThemes()
#3 /var/www/html/lib/Cake/Console/Shell.php(511): TemplateTask->initialize()
#4 /var/www/html/lib/Cake/Console/Command/Task/FixtureTask.php(288): Shell->__get('Template')
#5 /var/www/html/lib/Cake/Console/Command/Task/FixtureTask.php(270): FixtureTask->generateFixtureFile('Admin', Array)
#6 /var/www/html/lib/Cake/Console/Command/Task/FixtureTask.php(170): FixtureTask->bake('Admin', 'admins', Array)
#7 /var/www/html/lib/Cake/Console/Command/Task/FixtureTask.php(116): FixtureTask->_interactive()
#8 /var/www/html/lib/Cake/Console/Shell.php(459): FixtureTask->execute()
#9 /var/www/html/lib/Cake/Console/Shell.php(456): Shell->runCommand('execute', Array)
#10 /var/www/html/lib/Cake/Console/ShellDispatcher.php(222): Shell->runCommand('Fixture', Array)
#11 /var/www/html/lib/Cake/Console/ShellDispatcher.php(66): ShellDispatcher->dispatch()
#12 /var/www/html/app/Console/cake.php(37): ShellDispatcher::run(Array)
#13 {main}
`

othercorey:
We are not supporting PHP 8 for CakePHP 2.x.

See https://github.com/cakephp/cakephp/issues/15197

antonvolokha:
[...]

Issue #15197, opened 2020-11-28
CakePHP 2.x support PHP 8.0?


EOM:
This is a (multiple allowed):

* [] bug
* [ ] enhancement
* [x] feature-discussion (RFC)

* CakePHP Version: 2.X

markstory:
There are no plans to update 2.x to support PHP8. CakePHP 2.x is scheduled to only receive security updates as of December 15 2020.

EOM:
OK ;(

bancer:
@EOM maybe we should make it compatible with PHP 8 in a fork if it would not require too many changes. I believe it would be much simpler to make 2.x compatible with PHP 8 than to migrate the entire project to 3.x. This would be quite a bummer if 2.x would be the only reason why we could use PHP 8 in our project.

RyanTheDevGuy:
> @EOM maybe we should make it compatible with PHP 8 in a fork if it would not require too many changes. I believe it would be much simpler to make 2.x compatible with PHP 8 than to migrate the entire project to 3.x. This would be quite a bummer if 2.x would be the only reason why we could not use PHP 8 in our project.

@bancer do you know if this fork was ever created?

bancer:
@fezzy2020 Not per my knowledge.

kamilwylegala:
@bancer @fezzy2020 I actively maintain project that uses mix of CakePHP 2 and Symfony. Like you mentioned, moving to CakePHP 3.* is too expensive for me.

So I created a fork at the beginning of this year: https://github.com/kamilwylegala/cakephp2-php8 to support PHP 8.0.

I plan to migrate to PHP 8.1 in my project soon, so you can expect an update in my fork.

osd84:
You are an hero, if you do that little month before i didn't jump on CakePHP 4.
Now i will stay here, for ever and ever...

RyanTheDevGuy:
@kamilwylegala Amazing!

vincent-fletcher:
@kamilwylegala - Thank you very much for providing this :)

JoelFB:
@kamilwylegala can your fork only be used if we use Symfony as you do or this will work well for being able to support php8 on cake 2?

kamilwylegala:
@JoelFB You can use fork alone just like you use original framework. Forks contains only necessary changes to support PHP8/8.1.

JoelFB:
Please confirm that we do not need to use Symfony to make use of your fork @kamilwylegala 

Thank you

PR #15373, opened 2021-03-07
Fix issue #15372

antonvolokha:
Fix issue #15372

array_merge() expects to be passed one or more arrays to be merged. As the $arrays parameter is variadic, PHP 8.0 named parameters is not supported.
So I just replace named parameters to numeric using array_values


othercorey:
We are not supporting PHP 8 for CakePHP 2.x.

See https://github.com/cakephp/cakephp/issues/15197

Issue #15213, opened 2020-12-10
RFC to add additional MIME types (KML / KMZ) to Cake\HTTP\Response MIME types


impronta48:
This is a (multiple allowed):

* [ ] bug
* [ ] enhancement
* [X] feature-discussion (RFC)

### Proposal
Support KML / KMZ file types in order to avoid _Invoked extension not recognized/configured: kml_

### What happened
I try to define   $builder->setExtensions(['xls', 'json', 'kml', 'xml']);
but the KML extension is not recognized because it does not appear in protected $_mimeTypes

### What you expected to happen
I would like all valid extensions to be recognized, in particular common ones like KML.

### Version .2
Would be great to use an official updated source of mime types, such as
https://cdn.jsdelivr.net/gh/jshttp/mime-db@master/db.json
(more details here https://www.npmjs.com/package/mime-db)

impronta48:
In the meanwhile I discovered that I could add extra mime types like this, in my controller
```php
  public function beforeFilter($event)
  {
    $this->response->setTypeMap('kml', ['application/vnd.google-earth.kml+xml']);
    parent::beforeFilter($event);

    if ($this->request->getParam('_ext') == 'kml') {
      $this->response = $this->response->withType('kml');
    }
  }
```
Is this the proper way to handle them?

othercorey:
One issue we have with newer types is the lack of agreed extension to represent them. Those lists don't provide that.

ADmad:
> Is this the proper way to handle them?

Yes it is.

othercorey:
@impronta48 You can create a PR that adds kml and kmz types.

PR #15376, opened 2021-03-08
Add kml/kmz mime types and update name of hal+xml type

othercorey:
Closes https://github.com/cakephp/cakephp/issues/15213

Updated based on http://svn.apache.org/repos/asf/httpd/httpd/trunk/docs/conf/mime.types

The registered extension for hal+xml according to the list is just `hal`.


src/Http/Response.php
```diff
@@ -167,8 +167,8 @@ class Response implements ResponseInterface
         'gz' => 'application/x-gzip',
         'bz2' => 'application/x-bzip',
         '7z' => 'application/x-7z-compressed',
+        'hal' => ['application/hal+xml', 'application/vnd.hal+xml'],
         'haljson' => ['application/hal+json', 'application/vnd.hal+json'],
-        'halxml' => ['application/hal+xml', 'application/vnd.hal+xml'],
```
markstory, line 171: Wasn't this added in 4.2? Renaming it will break any applications using it. We could add another alias to align with what apache has though.

othercorey: Yes. Wasn't sure if fixing the name now and documenting would be ok.

othercorey: But I'll just add the alias

PR #15378, opened 2021-03-09
Revert "Backport: Detect infinite recursion when initializing Table"

othercorey:
Reverts cakephp/cakephp#15367


ADmad (approved):
Considering that this change breaks existing applications it is best to revert it for 3.x.

PR #15380, opened 2021-03-10
[3.x] Allow to compare "mail sent with" against an array

ravage84:
E.g. ``EmailTrait::assertMailSentFrom([$emailAddress => $displayName])``

https://book.cakephp.org/3/en/core-libraries/email.html#assertion-methods


markstory (commented):
Would be good to add a test to expand coverage for this new parameter style.

Issue #15383, opened 2021-03-10
Cake\I18n\PluralRules::calculate changes


hacheraw:
This is a (multiple allowed):

* [x] bug
* [x] enhancement
* [x] feature-discussion (RFC)

* CakePHP Version: 4.2.4
* Platform and Target: Windows 10 PHP 7.4

**I was having trouble with Plurals. Cake was always returning singulars.**
I found that `Cake\I18n\PluralRules::calculate` does:

```php
if (!isset(static::$_rulesMap[$locale])) {
    $locale = explode('_', $locale)[0];
}

if (!isset(static::$_rulesMap[$locale])) {
    return 0;
}
```

So it always returns 0 (singular) because I'm not using **underscode** but an **hyphen**:
using es-ES instead of es_ES

I searched on the Internet looking for the correct form and some people says the correct is the former and some others the latter.

So I have two questions ~~and a bug~~:

### Question 1
Could we accept both forms? It could be done with:
- `str_replace('-', '_', $locale)`
- or `preg_match('/^[A-Za-z]+/', $locale, $out)`

### Question 2
I don't know if this is possible but what are *.po headers for? Could we extract the `Plural-Forms` rule from it?
Example:
```gettext
"Project-Id-Version: PROJECT VERSION\n"
"POT-Creation-Date: 2020-12-15 21:32+0000\n"
"PO-Revision-Date: 2021-02-23 23:12+0000\n"
"Last-Translator: Hache_raw <EMAIL@ADDRESS>\n"
"Language-Team: LANGUAGE <EMAIL@ADDRESS>\n"
"Language: es-ES\n"
"MIME-Version: 1.0\n"
"Content-Type: text/plain; charset=UTF-8\n"
"Content-Transfer-Encoding: 8bit\n"
"Plural-Forms: nplurals=2; plural=(n != 1);\n"  <------- THIS
"X-Generator: Poedit 2.4.1\n"
```

### ~~Bug~~
~~`static::$_rulesMap[]` has a value which is: 'pt_pt'.~~
~~That will never be used because of the current `explode('_', $locale)[0]`;~~

hacheraw:
I just removed the "bug" part. I was wrong.

markstory:
> Could we accept both forms? It could be done with:

Converting `-` to `_` to increase interoperability with other tools/po files seems reasonable to me.

> I don't know if this is possible but what are *.po headers for? Could we extract the Plural-Forms from it?

[...]

PR #15384, opened 2021-03-10
Accept hyphens in locale

hacheraw:
– IETF Language Tag: uses a hyphen
– ISO 15897: uses an underscore

More info in #15383

### DIFF
--- a/src/I18n/PluralRules.php
+++ b/src/I18n/PluralRules.php
@@ -143,7 +143,7 @@ class PluralRules
         $locale = strtolower($locale);
 
         if (!isset(static::$_rulesMap[$locale])) {
-            $locale = explode('_', $locale)[0];
+            $locale = explode('_', str_replace('-', '_', $locale))[0];
         }
 
         if (!isset(static::$_rulesMap[$locale])) {


ADmad:
The `calculate()` methods links to https://localization-guide.readthedocs.io/en/latest/l10n/pluralforms.html which shows locales in `xx_XX` form. PHP's [Local::canonicalize()](https://www.php.net/manual/en/locale.canonicalize.php) also normalized the locale in same form.

So instead of adhoc string manipulations I would rather we use `Local::canonicalize()`. Only `pt_pt` in the rules map would need to be updated to `pt_PT`.

Though instead of `pt`, `pt_pt` the list should probably have `pt` and `pt_BR` IMO.

ADmad:
For 3 letter locales like `fil` for .e.g. `Local::canonicalize()` return the same string `fil`. So using this function shouldn't be a problem.

ADmad:
I haven't checked where else we can use Locale::canonicalize().

I have already opened an alternative PR https://github.com/cakephp/cakephp/pull/15386.

markstory:
Thanks for putting this together. We'll go with #15386 as using Locale will be easier to maintain.

PR #15386, opened 2021-03-11
Canonicalize locale string.

ADmad:
This avoids problems when locale string is provided in varying forms, for e.g. "en-us", "en_US" etc. Refs #15383.

Plurals rules for Portuguese and Brazilian Portuguese have been updated as per info provided by following resources:
http://localization-guide.readthedocs.org/en/latest/l10n/pluralforms.html
https://developer.mozilla.org/en-US/docs/Mozilla/Localization/Localization_and_Plurals#List_of_Plural_Rules

The class PluralRules has been marked as internal.


ADmad:
I have targeted 4.next as I believe existing map for Portuguese languages are incorrect but it's best to not change them in a minor release.

othercorey:
Looks good me.

On the PHP-side, it's confusing that they document Locale as accepting RFC 4646 tags (es-ES), but canonicalize() converts to the ICU format.

 I was going to suggest we decide if RFC 4646 style tags should be preferred in cakephp config, but this makes it very confusing.

PR #15388, opened 2021-03-11
Use proposed methods instead of deprecated ones

martinstuecklschwaiger:
This removes deprecation warnings but otherwise doesn't change anything.


othercorey:
CakePHP 2 is only receiving security updates until June 15, 2021.

https://github.com/cakephp/cakephp/wiki#2x-security-mode

PR #15389, opened 2021-03-11
Tag internal association code

othercorey:
This is code that is subject to change with association fixes/optimizations.

Users should not be referencing it directly.

### DIFF
--- a/src/ORM/Association.php
+++ b/src/ORM/Association.php
@@ -707,6 +707,7 @@ abstract class Association
      * - negateMatch: Will append a condition to the passed query for excluding matches.
      *   with this association.
      *
+     * @internal
      * @param \Cake\ORM\Query $query the query to be altered to include the target table data
      * @param array $options Any extra options or overrides to be taken in account
      * @return void
--- a/src/ORM/EagerLoader.php
+++ b/src/ORM/EagerLoader.php
@@ -27,6 +27,8 @@ use InvalidArgumentException;
  * for a table once a query is provided and delegates the job of creating the
  * required joins and decorating the results so that those associations can be
  * part of the result set.
+ *
+ * @internal
  */
 class EagerLoader
 {


src/ORM/Association.php
```diff
@@ -707,6 +707,7 @@ protected function _options(array $options): void
      * - negateMatch: Will append a condition to the passed query for excluding matches.
      *   with this association.
      *
+     * @internal
```
ndm2, line 710: This doesn't feel overly nice, I already dislike how `SelectLoader` and `SelectWithPivotLoader` are internal. Personally I have quite a few custom/extended association and loader classes in my projects.

othercorey: I am slowly simplifying/clarifying the association code. I'm trying to narrow down the places that we have to support because the association join generation has some larger issues.

If this is actually supported historically, then we'll just have to leave it alone - but I'm not sure the join generation is something public.

ndm2: Oh, I got distracted by the other `@internal` tags that are on class level, I overlooked that this one is on method level, but I'm actually using that one specifically, though only in _one_ custom association and _one_ custom loader 🙃 

Do you already have any plans for providing a different way to do the same, or maybe to drop the necessity for it?

othercorey: I don't know that it will "go away" entirely, but the implementation will change. But yes, you'll get similar functionality -- I just can't guarantee the same code.

Do you use the AssociationCollection and Trait? I included those because they seemed to go with the rest of the internal code, but it's not a requirement for me at this time.

ndm2: I see. I mean, personally I could live with it as long as I could relatively easily change up my implementation to get it continue working, but I'm not sure how much of a BC problem this is considered to be with respect to CakePHP's BC policy, it feels like something between deprecation and an actual BC break.

Never did anything with collection or normalizer trait, not, also never touched eager loader.

othercorey: I reverted those 2 changes.

othercorey: The actual changes to EagerLoader have been cleanup to make the code readable so we can review changes and determine differences in implementations.

The functionality changes are planning to be only minor-release level bug fixes. If it looks like we actually have to break BC, then it will be pushed.

An example of issues we're looking to fix long-term are associations that have joins that need to be nested with the association join. At the moment, we can't determine how to generate that.

Another lower priority fix is making it possible to have joins + contains be sorted properly.

src/ORM/EagerLoader.php
```diff
@@ -27,6 +27,8 @@
  * for a table once a query is provided and delegates the job of creating the
  * required joins and decorating the results so that those associations can be
  * part of the result set.
+ *
+ * @internal
```
othercorey, line 31: If we can't mark this internal, the function I'm looking to modify is `setMatching()` and related code. I don't know that `contain()` needs to change.

src/ORM/Association.php
ndm2: Order of joins is a long outstanding issue I remember... nested joins sounds nice, with the potential for quite a BC break :)

PR #15392, opened 2021-03-11
change Method Visibility to public

zoghal:
when using a CakePHP Console Library into a simple another project, for Automatically discover shell commands, we need public access for this methods: CommandScanner::scanDir() & CommandCollection::resolveNames().

```
<?php
namespace Tools;

use Cake\Console\CommandScanner;
use Cake\Console\CommandCollection;
use Cake\Core\ConsoleApplicationInterface;

class App implements ConsoleApplicationInterface
{
    public function bootstrap(): void
    {
    }

    public function console(CommandCollection $commands): CommandCollection
    {

        $scanner = new CommandScanner();
        $lists = $scanner->scanDir(__DIR__ . DS . 'Command' . DS, 'Tools\Command\\', '', []);
        $lists = $commands->resolveNames($lists);
        $commands->addMany($lists);

        return $commands;
    }
}

```

### DIFF
--- a/src/Console/CommandCollection.php
+++ b/src/Console/CommandCollection.php
@@ -188,7 +188,7 @@ class CommandCollection implements IteratorAggregate, Countable
      * @param array $input The results of a CommandScanner operation.
      * @return string[] A flat map of command names => class names.
      */
-    protected function resolveNames(array $input): array
+    public function resolveNames(array $input): array
     {
         $out = [];
         foreach ($input as $info) {
--- a/src/Console/CommandScanner.php
+++ b/src/Console/CommandScanner.php
@@ -108,7 +108,7 @@ class CommandScanner
      * @param string[] $hide A list of command names to hide as they are internal commands.
      * @return array The list of shell info arrays based on scanning the filesystem and inflection.
      */
-    protected function scanDir(string $path, string $namespace, string $prefix, array $hide): array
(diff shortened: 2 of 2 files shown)


markstory:
I am not sure we should expose these methods. Perhaps we could add a new public method that makes adding commands from a directory as a single operation instead?

zoghal:
> 
> 
> Perhaps we could add a new public method that makes adding commands from a directory as a single operation instead?

You mean like this:

```
class CommandScanner
{
    public function scanPath(string $path, string $namespace): array
    {
        return $this->scanDir($path, $namespace, '', [])
    }
}

class CommandCollection implements IteratorAggregate, Countable
{
    public function discoverPath(string $path, string $namespace): array
    {
        $scanner = new CommandScanner();
        $shells = $scanner->scanDir($path, $namespace);

        return $this->resolveNames($shells);
    }
}

class App implements ConsoleApplicationInterface
{
    public function console(CommandCollection $commands): CommandCollection
    {
        $shells= $commands->discoverPath(__DIR__ . DS . 'Command' . DS, 'Tools\Command\\');
        $commands->addMany($lists);

        return $commands;
    }
}
```

othercorey:
Yes, something like that.

othercorey:
Please open new PR when implemented.

PR #15395, opened 2021-03-15
Fix RouteBuilder::plugin() not forwarding namePrefix option

### LINKED ISSUE #15394: _namePrefix not working in \Cake\Routing\RouteBuilder::plugin
This is a (multiple allowed):
* [x] bug

* CakePHP Version: 4.2.4.
* Platform and Target: Ubuntu Linux 18.04, PHP 7.2.

### What you did
I've set up a plugin with a namePrefix for its routes.

```php
// my Plugin class
class Plugin extends BasePlugin
{
    public function routes(RouteBuilder $routes): void
    {
        $config = [
            'path' => '/sadmin',
            '_namePrefix' => 'sadmin.',
        ];

        $routes->plugin('SuperAdmin', $config, function (RouteBuilder $builder) {
            $builder->connect('/', 'Clientes::index', ['_name' => 'home']);
            $builder->fallbacks();
        });

        parent::routes($routes);
    }
}
```

### What happened
The route is registered with the _name "home" without the prefix.

The route should be named as "sadmin.home"

I've been looking into the code and I think the issue is on the `\Cake\Routing\RouteBuilder::plugin()` method, as it's ignoring any options key other than `"path"`:

```php
// \Cake\Routing\RouteBuilder class
    public function plugin(string $name, $options = [], $callback = null)
    {
        if (!is_array($options)) {
            $callback = $options;
            $options = [];
        }

        $params = ['plugin' => $name] + $this->_params;
        $path = $options['path'] ?? '/' . Inflector::dasherize($name);
        $this->scope($path, $params, $callback);

        return $this;
    }
```
As you can see, `$options` is not passed nor merged into `scope()` method, instead only the `"path"` key is taken.
I found this bug some time ago, in cakephp 3.x, I was almost sure I'd reported it then, but it turns out that I didn't, I'd just used the scope method directly, avoiding the use of `plugin()`.

markstory:
I also improved the API docs for plugin() and scope() to be clearer on the supported options.

Fixes #15394

### DIFF
--- a/src/Routing/RouteBuilder.php
+++ b/src/Routing/RouteBuilder.php
@@ -868,8 +868,14 @@ class RouteBuilder
      * Routes connected in the scoped collection will have the correct path segment
      * prepended, and have a matching plugin routing key set.
      *
+     * ### Options
+     *
+     * - `path` The path prefix to use. Defaults to `Inflector::dasherize($name)`.
+     * - `_namePrefix` Set a prefix used for named routes. The prefix is prepended to the
+     *   name of any route created in a scope callback.
+     *
      * @param string $name The plugin name to build routes for
-     * @param array|callable $options Either the options to use, or a callback
+     * @param array|callable $options Either the options to use, or a callback to build routes.
      * @param callable|null $callback The callback to invoke that builds the plugin routes
      *   Only required when $options is defined.
      * @return $this
@@ -881,8 +887,9 @@ class RouteBuilder
             $options = [];
         }
 
-        $params = ['plugin' => $name] + $this->_params;
         $path = $options['path'] ?? '/' . Inflector::dasherize($name);
+        unset($options['path']);
+        $params = ['plugin' => $name] + $options + $this->_params;
         $this->scope($path, $params, $callback);
 
         return $this;
@@ -894,6 +901,11 @@ class RouteBuilder
      * Scopes created with this method will inherit the properties of the scope they are
      * added to. This means that both the current path and parameters will be appended
      * to the supplied parameters.
+     *
+     * ### Special Keys in $params
+     *
+     * - `_namePrefix` Set a prefix used for named routes. The prefix is prepended to the
+     *   name of any route created in a scope callback.
      *
      * @param string $path The path to create a scope for.
      * @param array|callable $params Either the parameters to add to routes, or a callback.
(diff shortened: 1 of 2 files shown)


othercorey:
Looks good. Do we need to backport to 3.x?

src/Routing/RouteBuilder.php
```diff
@@ -881,8 +887,9 @@ public function plugin(string $name, $options = [], $callback = null)
             $options = [];
         }
 
-        $params = ['plugin' => $name] + $this->_params;
         $path = $options['path'] ?? '/' . Inflector::dasherize($name);
+        unset($options['path']);
+        $params = ['plugin' => $name] + $options + $this->_params;
```
ADmad, line 892: This mixing of options and params is confusing and doesn't make for very readable code. The constructor has two separate arguments for params and options while for thos method now both seem to be clubbed together.

We should better document which keys options and params can contain and if possible avoid such clubbing. 

Also there's inconsistent in key naming. For e.g. for the constructor's options `namePrefix` doesn't have a underscore prefix while at other places it does.

markstory: > We should better document which keys options and params can contain and if possible avoid such clubbing.

I considered copying over only the relevant keys but I got lazy. I can get unlazy if being explicit is important here.

> Also there's inconsistent in key naming. For e.g. for the constructor's options namePrefix doesn't have a underscore prefix while at other places it does.

`RouteBuilder::scope()` also uses `_namePrefix`. I'm pretty sure that option has an `_` because it was added in a minor release and we didn't want to shadow userland route keys.

ADmad: I pushed a commit to remove unneeded additional local var and array merging :slightly_smiling_face:.

Regarding the `$options` vs `$params` inconsistency/confusion; `plugin()`, `prefix()` are quite similar convenience wrappers for `scope()` which just preset certain values. Yet for `plugin()` the 2nd arg is called `$options` while for `prefix()` it's `$params` (which matches `scope()` when both args can have pretty much the same keys.

ADmad (approved):

src/Routing/RouteBuilder.php
ADmad: That said I guess we can live we the current argument names.

PR #15396, opened 2021-03-15
Use new braced route element placeholders in resource paths.

ADmad:

### DIFF
--- a/src/Routing/RouteBuilder.php
+++ b/src/Routing/RouteBuilder.php
@@ -56,9 +56,9 @@ class RouteBuilder
     protected static $_resourceMap = [
         'index' => ['action' => 'index', 'method' => 'GET', 'path' => ''],
         'create' => ['action' => 'add', 'method' => 'POST', 'path' => ''],
-        'view' => ['action' => 'view', 'method' => 'GET', 'path' => ':id'],
-        'update' => ['action' => 'edit', 'method' => ['PUT', 'PATCH'], 'path' => ':id'],
-        'delete' => ['action' => 'delete', 'method' => 'DELETE', 'path' => ':id'],
+        'view' => ['action' => 'view', 'method' => 'GET', 'path' => '{id}'],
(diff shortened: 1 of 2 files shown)


src/Routing/RouteBuilder.php
```diff
@@ -215,6 +215,11 @@ public function addExtensions($extensions)
      */
     public function path(): string
     {
+        $routeKey = strpos($this->_path, '{');
+        if ($routeKey !== false && strpos($this->_path, '}') !== false) {
```
othercorey, line 219: Shouldn't this look for `}` after `{`?

ADmad: That's what's it's doing.

othercorey: I mean after the position in the string.

ADmad: Ah, ideally yes. I copied the check from another place in the routing code. Guess I can update all, though if one has a closing brace before an opening brace then their route template is most likely messed up.

othercorey: Up to you if you want to throw a validation error if } is before {.

ADmad: I don't think such a check is necessary. A broken route would be indication enough. Being over zealous in handling dev error didn't go too well last time we tried :wink:.

src/Routing/RouteBuilder.php
```diff
@@ -421,7 +426,7 @@ public function resources(string $name, $options = [], $callback = null)
 
         if ($callback !== null) {
             $idName = Inflector::singularize(Inflector::underscore($name)) . '_id';
-            $path = '/' . $options['path'] . '/:' . $idName;
+            $path = '/' . $options['path'] . '/{' . $idName . '}';
```
garas, line 429: 4 concatenations is bit too much, better `sprintf`

ADmad: The total line length is quite reasonable and it's still easy enough to grok what's being done :).

PR #15403, opened 2021-03-19
Don't use strlen() when we just need to check if string is empty or not.

ADmad:

### DIFF
--- a/src/Mailer/Message.php
+++ b/src/Mailer/Message.php
@@ -1610,7 +1610,7 @@ class Message implements JsonSerializable, Serializable
      */
     protected function wrap(?string $message = null, int $wrapLength = self::LINE_LENGTH_MUST): array
     {
-        if ($message === null || strlen($message) === 0) {
+        if ($message === null || $message === '') {
             return [''];
         }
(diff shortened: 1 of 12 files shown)


src/Mailer/Message.php
```diff
@@ -1610,7 +1610,7 @@ protected function encodeString(string $text, string $charset): string
      */
     protected function wrap(?string $message = null, int $wrapLength = self::LINE_LENGTH_MUST): array
     {
-        if ($message === null || strlen($message) === 0) {
+        if ($message === null || $message === '') {
```
othercorey, line 1613:
```suggestion
        if (empty($message)) {
```

markstory: I would be careful with empty() as it also eats '0'.

othercorey: True, but $message is `string|null`. But was just a suggestion.

ADmad: So it can be string `0`, empty() gobble that up.

othercorey: Sorry I missed it was a string lol.

Issue #15308, opened 2021-01-28
4.next - Separate schema and data for test fixtures


markstory:
This is a (multiple allowed):

* [ ] bug
* [ ] enhancement
* [x] feature-discussion (RFC)

Maintaining and improving the current fixture system is challenging as it is complex has many hidden features and relies on deprecated PHPUnit APIs. 

## Problems

* The various ways we currently manage schema results in even more complexity as handling constraints requires workarounds and hacks.
* We rely on deprecated PHPUnit APIs which will be removed. Potentially in PHPUnit 10.
* The replacement APIs in PHPUnit are not suitable for our use case.

As there are a few orthogonal problems we could address them separately, but addressing them together could unlock a better developer experience.

### Integration with PHPUnit

Currently the test suite uses `PHPUnit\Framework\TestListener` to listen for tests being started. The test suite and test case start events are used to create schema and populate records before a test is run. This interface is deprecated, and [extensions are recommended instead](https://phpunit.readthedocs.io/en/9.5/extending-phpunit.html#extending-the-testrunner). Unfortunately, extensions do not have access to the testsuite or the test case objects. While they do have access to the class and method names, this data is not ideal and limits any future fixture APIs to being static.

We have had reasonable success using traits that leverage the `@before` annotation to do setup work. Another potential approach is to use the `TestCase::setUp()` to collaborate with shared global that wraps the database 'state'.

### Managing Schema

Currently the test suite drops and manages schema at the table level. As fixtures are loaded/used tables are created. At the end of the test run all tables are dropped. This incurs some overhead as we need to track which tables have been created, and whether the table needs to be dropped before it is created. Compounding this complexity is the variety of ways we allow developers to define schema:

1. As abstract schema in fixture classes via the `$schema` property. This form is not the actual schema, but an abstract representation, that can be leaky or incorrect. Having schema in fixtures, creates complexity for developers, as there is another place to manage schema as it changes.
2. As an 'import' from a table or a model. Using the reflection API's schema is read and then reverse engineered back into SQL. This conversion can be lossy or inaccurate.
3. As an unmanaged table that does not import or generate a schema but does truncate table rows.

These various schema generation approaches require us to have a comprehensive schema generation library in CakePHP adding complexity to the database abstractions.

### Managing Schema

There are several potential options when it comes to schema management.

[...]

PR #15404, opened 2021-03-21
4.next Start data only fixtures

markstory:
These changes rough in the PHPUnit extension, and transaction based state manager. The state manager will be swappable via the PHPUnit configuration file. More complex implementations will require complicated PHPUnit configuration.

This is part of #15308

### DIFF
--- a/src/TestSuite/Fixture/TransactionStrategy.php
+++ b/src/TestSuite/Fixture/TransactionStrategy.php
@@ -16,7 +16,9 @@ declare(strict_types=1);
  */
 namespace Cake\TestSuite\Fixture;
 
+use Cake\Database\Connection;
 use Cake\Datasource\ConnectionManager;
+use RuntimeException;
 
 /**
  * Fixture state strategy that wraps tests in transactions
@@ -69,11 +71,17 @@ class TransactionStrategy
         foreach ($map as $testConnection => $normal) {
             ConnectionManager::alias($testConnection, $normal);
             $connection = ConnectionManager::get($normal);
-            if (method_exists($connection, 'enableSavePoints') && !$connection->enableSavePoints(true)) {
-                throw new RuntimeException("Could not enable save points for the `{$normal}` connection. Your database needs to support savepoints in order to use the TransactionStrategy for fixtures.");
-            }
-            if ($enableLogging) {
-                $connection->enableQueryLogging(true);
+            if ($connection instanceof Connection) {
+                if (!$connection->enableSavePoints(true)) {
+                    throw new RuntimeException(
+                        "Could not enable save points for the `{$normal}` connection. " .
+                            'Your database needs to support savepoints in order to use the ' .
+                            'TransactionStrategy for fixtures.'
+                    );
+                }
+                if ($enableLogging) {
+                    $connection->enableQueryLogging(true);
+                }
             }
         }
     }
@@ -91,7 +99,7 @@ class TransactionStrategy
                 continue;
             }
             $db = ConnectionManager::get($connection);
-            if (method_exists($db, 'begin')) {
+            if ($db instanceof Connection) {
                 $db->begin();
                 $db->createSavePoint('__fixtures__');
             }
@@ -115,7 +123,7 @@ class TransactionStrategy
                 continue;
             }
             $db = ConnectionManager::get($connection);
-            if (method_exists($db, 'rollback')) {
+            if ($db instanceof Connection) {
                 $db->rollback(true);
             }
         }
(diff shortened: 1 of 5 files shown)


src/TestSuite/Fixture/TransactionStrategy.php
```diff
@@ -0,0 +1,131 @@
...
+    /**
+     * Before each test start a transaction.
+     *
+     * @return void
+     */
+    public function beforeTest(): void
+    {
+        $connections = ConnectionManager::configured();
+        foreach ($connections as $connection) {
+            if (strpos($connection, 'test') === false) {
```
othercorey, line 98:
```suggestion
            if (strpos($connection, 'test') !== 0) {
```

src/TestSuite/Fixture/TransactionStrategy.php
```diff
@@ -0,0 +1,131 @@
...
+                continue;
+            }
+            $db = ConnectionManager::get($connection);
+            if ($db instanceof Connection) {
```
othercorey, line 102: What is the official support/expectation for non-Connection connections when using a transaction strategy? Are we just creating a save point on possible connection and all others are on their own? Or should we fail?

markstory: :thinking: that's a good question. Failing makes it really hard to have an application that uses both DB fixtures and elasticsearch as an example. This state manager is unfit for elasticsearch, and I haven't really thought through how it could be adapted yet.

With little thought applied each 'kind' of data store could have its own state manager that focuses on resetting their state (transactions for SQL db, truncation for ES). The extension could call each statemanager in a loop. Or perhaps plugins like elasticseach provide their own phpunit extensions to enable fixtures?

src/TestSuite/Fixture/TransactionStrategy.php
```diff
@@ -0,0 +1,131 @@
...
+                $map[$connection] = substr($connection, 5);
+            } else {
+                $map['test_' . $connection] = $connection;
+            }
+        }
+        foreach ($map as $testConnection => $normal) {
+            ConnectionManager::alias($testConnection, $normal);
+            $connection = ConnectionManager::get($normal);
+            if ($connection instanceof Connection) {
+                if (!$connection->enableSavePoints(true)) {
```
othercorey, line 75:
```suggestion
                if (!$connection->enableSavePoints()) {
```
I prefer to leave out unnecessary parameters for clearer code.

src/TestSuite/Fixture/TransactionStrategy.php
```diff
@@ -0,0 +1,131 @@
...
+                    throw new RuntimeException(
+                        "Could not enable save points for the `{$normal}` connection. " .
+                            'Your database needs to support savepoints in order to use the ' .
+                            'TransactionStrategy for fixtures.'
+                    );
+                }
+                if ($enableLogging) {
+                    $connection->enableQueryLogging(true);
```
othercorey, line 83:
```suggestion
                    $connection->enableQueryLogging();
```
Same here.

src/TestSuite/Fixture/TransactionStrategy.php
```diff
@@ -0,0 +1,131 @@
...
+     * operations we should end up at a transaction depth of 0
+     * and we will rollback the root transaction started in beforeTest()
+     *
+     * @return void
+     */
+    public function afterTest(): void
+    {
+        $connections = ConnectionManager::configured();
+        foreach ($connections as $connection) {
+            if (strpos($connection, 'test') === false) {
```
othercorey, line 122:
```suggestion
            if (strpos($connection, 'test') !== 0) {
```
Sorry, I missed this one earlier. For context, this checks if it starts with test which is how we define test-specific connections.

src/TestSuite/Fixture/TransactionStrategy.php
```diff
@@ -0,0 +1,131 @@
...
+                continue;
+            }
+            $db = ConnectionManager::get($connection);
+            if ($db instanceof Connection) {
+                $db->rollback(true);
```
othercorey, line 127: I'm not very familiar with save points, but why do we create them if we are rolling back to beginning here?

markstory: Application code could be using transactions as well. Using savepoints lets you have 'nested' transactions that can be rolled back in case of an error or constraint violation.

We want to roll everything back at the end of the test so that the database is 'clean' once again. With a transaction based 'state management' fixture records and rows modified during the test would all occur within a transaction. By rolling back at the end we can 'reset' the database without having to figure out which tables need to be truncated.

othercorey: I see. I think I'm just confused by we need the `__fixtures__` save point. Perhaps this code is too early to understand that workflow.

markstory: We might not, I'm still figuring things out as well.

PR #15405, opened 2021-03-21
Fix BufferedIterator not yielding full results

### LINKED ISSUE #15341: BufferedIterator::firstMatch not able to find elements
This is a (multiple allowed):

* [x] bug
* [ ] enhancement

* CakePHP Version: 3.x/4.x (dcbf2095a61b87270b6a5baa11aef26a39f2bba2)
* Platform and Target: LAMP

### What you did
```
    public function testBufferedIterator() {
        $data = [['myField' => '1'], ['myField' => '2'], ['myField' => '3']];
        $buffered = (new \Cake\Collection\Collection($data))->buffered();
        // $buffered->count();
        debug($buffered->firstMatch(['myField' => '1']));
        debug($buffered->firstMatch(['myField' => '2']));
        debug($buffered->firstMatch(['myField' => '3']));
    }
```

### What happened
```
BufferedIteratorTest.php (line 42)
########## DEBUG ##########
[
  'myField' => '1'
]
###########################
BufferedIteratorTest.php (line 43)
########## DEBUG ##########
[
  'myField' => '2'
]
###########################
BufferedIteratorTest.php (line 44)
########## DEBUG ##########
null
###########################
```

Third item should be found.

After adding a `$buffered->count()` the issue goes away, so something is likely off with the index.

markstory:
When a Collection\BufferedIterator is partially iterated, rewound and then iterated again it would not yield complete data as the inner iterator would be advanced while buffered data was being yielded.

Fixes #15341

### DIFF
--- a/src/Collection/Iterator/BufferedIterator.php
+++ b/src/Collection/Iterator/BufferedIterator.php
@@ -159,6 +159,10 @@ class BufferedIterator extends Collection implements Countable, Serializable
     {
         $this->_index++;
 
+        // Don't move inner iterator if we have more buffer
+        if ($this->_buffer->offsetExists($this->_index)) {
+            return;
+        }
         if (!$this->_finished) {
             parent::next();
         }
--- a/tests/TestCase/Collection/CollectionTest.php
+++ b/tests/TestCase/Collection/CollectionTest.php
@@ -553,7 +553,7 @@ class CollectionTest extends TestCase
     {
         $collection = new Collection($items);
         $callback = function ($e) {
-            return $e['a']['b']['c'] * - 1;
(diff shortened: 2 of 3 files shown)


src/Collection/Iterator/BufferedIterator.php
```diff
@@ -117,6 +117,9 @@ public function rewind(): void
         }
 
         $this->_index = 0;
+        // Reset the wrapped iterator so if it was partially
+        // iterated before it can be resumed.
+        // parent::rewind();
```
othercorey, line 122: Is this supposed to be here?

markstory: Nope, I'll clean that up.

PR #15406, opened 2021-03-21
Correct documentation for isSavePointsEnabled()

markstory:
The method always returns $this now.


othercorey:
I saw this when reviewing the other PR. I'm not sure silently ignoring was the best design choice here.

PR #15409, opened 2021-03-24
4.3 - Add data fixture manager

markstory:
Add the fixture manager that TestCase will use when data-only fixtures are being used. I might extract an interface for the old/new fixture managers to implement so that typechecking is easier.

Part of #15308

### DIFF
--- a/src/TestSuite/Fixture/DataFixtureManager.php
+++ b/src/TestSuite/Fixture/DataFixtureManager.php
@@ -1,0 +1,252 @@
+<?php
+declare(strict_types=1);
+
+/**
+ * CakePHP(tm) : Rapid Development Framework (https://cakephp.org)
+ * Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
+ *
+ * Licensed under The MIT License
+ * For full copyright and license information, please see the LICENSE.txt
+ * Redistributions of files must retain the above copyright notice.
+ *
+ * @copyright     Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
+ * @link          https://cakephp.org CakePHP(tm) Project
+ * @since         4.3.0
+ * @license       https://opensource.org/licenses/mit-license.php MIT License
+ */
+namespace Cake\TestSuite\Fixture;
+
+use Cake\Core\Configure;
+use Cake\Core\Exception\CakeException;
+use Cake\Datasource\ConnectionInterface;
+use Cake\Datasource\ConnectionManager;
+use Cake\TestSuite\TestCase;
+use PDOException;
+use RuntimeException;
+use UnexpectedValueException;
+
+/**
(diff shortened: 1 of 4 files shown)


src/TestSuite/Fixture/DataFixtureManager.php
```diff
@@ -0,0 +1,252 @@
...
+ * Data only fixture manager
+ *
+ * Part of the data-only fixture system. Data-only fixtures assume that
+ * you have created the schema for the testsuite *before* running tests.
+ * They also assume that any mutations to your schema are reverted by you.
+ *
+ * This class implements a common interface with the Cake\TestSuite\FixtureManager so that
+ * test cases only need to interact with a single interface.
+ */
+class DataFixtureManager
```
othercorey, line 38: This sounds more like `FixtureDataManager` instead of `DataFixtureManager` from the description. 

However, I don't know if some other fixture system sets up the schema or if it's entirely outside of the code.

src/TestSuite/Fixture/DataFixtureManager.php
```diff
@@ -0,0 +1,252 @@
...
+{
+    /**
+     * A mapping between the fixture name (including the prefix) and the object.
+     *
+     * @var \Cake\Datasource\FixtureInterface[]
+     */
+    protected $loaded = [];
```
othercorey, line 45: I disliked the original fixture manager tracking code. Are we going to "load" all FixtureInterface instance at once now?

Let's just call this `$fixtures`.

The difference between `$loaded` and `$fixtureMap` was ambiguous. Can we have a single array somehow? If not, we need to name them better than loaded and fixtureMap.

markstory: I admit the names could be better and now is the right time to fix them.

src/TestSuite/Fixture/DataFixtureManager.php
markstory: For the new fixtures the schema would be managed externally to this class. I'm fine with FixtureDataManager if that resonates better with others.

tests/TestCase/TestSuite/Fixture/FixtureDataManagerTest.php
```diff
@@ -0,0 +1,196 @@
...
+     * @return void
+     */
+    public function testLoadSingleValid()
+    {
+        $manager = new FixtureDataManager();
+        $manager->fixturize($this);
+
+        $manager->loadSingle('Articles');
+        $db = ConnectionManager::get('test');
+        $stmt = $db->newQuery()->select(['count()'])->from('articles')->execute();
```
ADmad, line 152:
```suggestion
        $stmt = $db->newQuery()->select(['count(*)'])->from('articles')->execute();
```

othercorey:
Merging. Will continue cleanup in future PRs.

PR #15421, opened 2021-03-31
Align FrozenDate API docs with chronos

### LINKED ISSUE #15418: FrozenDate construct with timezone return wrong date
This is a (multiple allowed):

* [x] bug
* [ ] enhancement

* CakePHP Version: 4.2.3
* Platform and Target: MacOS, Nginx, PHP 7.4

### What you did
Create `FrozenDate` in `UTC` when default timezone is `Asia/Tokyo`
```
$date = new FrozenDate('2021/03/30', 'UTC');
```

### What happened
`$date` is `2021/03/30` in `Asia/Tokyo`

`$date` is `2021/03/30` in `UTC`

markstory:
Align the documentation with what is in Chronos. Chronos was changed in https://github.com/cakephp/chronos/pull/227 (a820c1b1a523df341468d536ca7a2900ccaabc05) to fix problems with timezones that are far away from UTC.

Fixes #15418

### DIFF
--- a/src/I18n/FrozenDate.php
+++ b/src/I18n/FrozenDate.php
@@ -124,7 +124,7 @@ class FrozenDate extends ChronosDate implements I18nDateTimeInterface
      *
      * Date instances lack time components, however due to limitations in PHP's
      * internal Datetime object the time will always be set to 00:00:00, and the
-     * timezone will always be UTC. Normalizing the timezone allows for
+     * timezone will always be the server local time. Normalizing the timezone allows for
      * subtraction/addition to have deterministic results.
      *
      * @param string|int|\DateTime|\DateTimeImmutable|null $time Fixed or relative time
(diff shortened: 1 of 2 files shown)


othercorey:
Should we just inherit the chronos docblock?

src/I18n/FrozenDate.php
```diff
@@ -124,7 +124,7 @@ class FrozenDate extends ChronosDate implements I18nDateTimeInterface
      *
      * Date instances lack time components, however due to limitations in PHP's
      * internal Datetime object the time will always be set to 00:00:00, and the
-     * timezone will always be UTC. Normalizing the timezone allows for
+     * timezone will always be server local time. Normalizing the timezone allows for
```
ADmad, line 127: The statement "Normalizing the timezone.." doesn't seem necessary now.

markstory:
> Should we just inherit the chronos docblock?

Do docblocks inherit across projects?

othercorey:
Have to use the same parameter name as Chronos.

othercorey:
I reverted the fix back to wording change to avoid issues with Chronos.

PR #15424, opened 2021-04-02
4.3 - Truncation strategy for new fixtures

markstory:
This adds another strategy for resetting state in the new fixtures system. This strategy will be 'slow' but will be useful when applications need to run operations that cannot be run within transactions like DDL statements in MySQL.

Part of #15308 

cc @pabloelcolombiano

### DIFF
--- a/src/TestSuite/Fixture/TruncationStrategy.php
+++ b/src/TestSuite/Fixture/TruncationStrategy.php
@@ -1,0 +1,126 @@
+<?php
+declare(strict_types=1);
+
+/**
+ * CakePHP(tm) : Rapid Development Framework (https://cakephp.org)
+ * Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
+ *
+ * Licensed under The MIT License
+ * For full copyright and license information, please see the LICENSE.txt
+ * Redistributions of files must retain the above copyright notice.
+ *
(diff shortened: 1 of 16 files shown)


src/TestSuite/Fixture/TruncationStrategy.php
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+    /**
+     * Alias non test connections to the test ones
+     * so that models reach the test database connections instead.
+     *
+     * @param bool $enableLogging Whether or not to enable query logging.
+     * @return void
+     */
+    protected function aliasConnections(bool $enableLogging): void
```
markstory, line 55: This code is somewhat duplicate but not entirely duplicate from the TransactionStrategy. I would like to clean up the duplication once I have a better handle on what the abstractions should be.

pabloelcolombiano: This piece of code is very useful, and will be needed when the Schema is being created, for example before the migrations get run.

Therefore I suggest to move it in a dedicated class, so it can be called by other instances.

My proposal would be `Cake\TestSuite\TestConnectionManager.php`. I will include that in my coming PR for the Schema management.

markstory: Separating it out makes sense to me.

Issue #15426, opened 2021-04-05
Mention correct file name to update security salt


karmicdice:
This is a (multiple allowed):

* [ ] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 4.2.5
* Platform and Target: Apache2, Ubuntu, php 8.0

### What you did
Loaded cakephp using `composer create-project cakephp/app <project>`

### What happened
It threw a notice asking to change Security.sale in app.php file.

### What you expected to happen
It should say app_local.php (since it gets overwritten).

I kept changing the Security['salt'] in app.php but it never picked up, turns out the real to update where change happens is app_local.php

PR #15427, opened 2021-04-05
Re-worded to correct filename #15426

karmicdice:
Wrote the correct file to edit so that the notice goes away.


karmicdice:
Ref: https://github.com/cakephp/cakephp/issues/15426

markstory:
You'll need to rebase your changes on top of master. The 4.0.x branch is only used to do security releases for the 4.0 branch.

PR #15428, opened 2021-04-05
Update Debugger to reflect correct file name

karmicdice:
Ref: https://github.com/cakephp/cakephp/issues/15426


markstory:
Thanks :+1:

PR #15429, opened 2021-04-06
#15308 Introduces a SchemaManager to create a sql file based test sch…

pabloelcolombiano:
This commit introduces a test schema management strategy based on SQL files.

The developer is entitled, e.g. in `tests/bootstrap.php`, to provide one file or an array of files, containing the SQl commands to be run prior to the test suite to start.

The whole schema gets dropped, recreated and the table are truncated prior to the test suite to run.

This strategy will find a corresponding, more efficient, approach in the Migrations plugin.

### DIFF
--- a/src/TestSuite/Schema/SchemaCleaner.php
+++ b/src/TestSuite/Schema/SchemaCleaner.php
@@ -33,20 +33,13 @@ class SchemaCleaner
     protected $io;
 
     /**
-     * @var bool
-     */
-    protected $isDroppingEnabled = true;
-
-    /**
      * SchemaCleaner constructor.
      *
      * @param \Cake\Console\ConsoleIo|null $io Outputs if provided.
-     * @param bool $enableDropping Enable dropping (useful for testing purposes)
      */
-    public function __construct(?ConsoleIo $io = null, ?bool $enableDropping = true)
+    public function __construct(?ConsoleIo $io = null)
     {
         $this->io = $io;
-        $this->isDroppingEnabled = $enableDropping;
     }
 
     /**
@@ -59,9 +52,7 @@ class SchemaCleaner
      */
     public function dropTables(string $connectionName, $tables = null): void
     {
-        if ($this->isDroppingEnabled === true) {
-            $this->handle($connectionName, 'dropTableSql', 'dropping', $tables);
-        }
+        $this->handle($connectionName, 'dropTableSql', 'dropping', $tables);
     }
 
     /**
--- a/src/TestSuite/Schema/SchemaCleaner.php
+++ b/src/TestSuite/Schema/SchemaCleaner.php
@@ -54,8 +54,8 @@ class SchemaCleaner
     /**
      * Truncate all tables of the provided connection.
      *
-     * @param \Cake\Console\ConsoleIo|null $io Console IO to output the processes.
      * @param string $connectionName Name of the connection.
+     * @param \Cake\Console\ConsoleIo|null $io Console IO to output the processes.
      * @return void
      * @throws \Exception if the truncation failed.
      */
@@ -117,9 +117,7 @@ class SchemaCleaner
     }
 
     /**
-     *
      * @param string $msg Message to display.
-     *
      * @param \Cake\Console\ConsoleIo|null $io Console IO.
      * @return void
      */
--- a/src/TestSuite/Schema/SchemaCleaner.php
+++ b/src/TestSuite/Schema/SchemaCleaner.php
@@ -33,34 +33,80 @@ class SchemaCleaner
     protected $io;
 
     /**
+     * @var bool
+     */
+    protected $isDroppingEnabled;
+
+    /**
      * SchemaCleaner constructor.
      *
      * @param \Cake\Console\ConsoleIo|null $io Outputs if provided.
+     * @param bool|null $enableDropping Enable dropping (useful for testing purposes)
      */
-    public function __construct(?ConsoleIo $io = null)
+    public function __construct(?ConsoleIo $io = null, ?bool $enableDropping = true)
     {
         $this->io = $io;
+        $this->isDroppingEnabled = $enableDropping;
     }
 
     /**
      * Drop all tables of the provided connection.
      *
      * @param string $connectionName Name of the connection.
+     * @param array|null $tables Tables to truncate (all if not set).
      * @return void
      * @throws \Exception if the dropping failed.
      */
-    public function dropTables(string $connectionName)
+    public function dropTables(string $connectionName, $tables = null): void
     {
-        $this->info('Dropping all tables for connection ' . $connectionName);
+        if ($this->isDroppingEnabled === true) {
+            $this->handle($connectionName, 'dropTableSql', 'dropping', $tables);
+        }
+    }
 
+    /**
+     * Truncate all tables of the provided connection.
+     *
+     * @param string $connectionName Name of the connection.
+     * @param array|null $tables Tables to truncate (all if not set).
+     * @return void
+     * @throws \Exception if the truncation failed.
+     */
+    public function truncateTables(string $connectionName, $tables = null): void
+    {
+        $this->handle($connectionName, 'truncateTableSql', 'truncating', $tables);
+    }
+
+    /**
+     * @param string $connectionName Name of the connection.
+     * @param string $dialectMethod Method applied to the SQL dialect.
+     * @param string $action Action displayed in the info message
+     * @param string[]|null $tables Tables to truncate (all if null)
+     * @return void
+     * @throws \Exception
+     */
+    protected function handle(string $connectionName, string $dialectMethod, string $action, $tables): void
+    {
         $schema = $this->getSchema($connectionName);
         $dialect = $this->getDialect($connectionName);
+        $allTables = $schema->listTables();
+        if (is_null($tables)) {
+            $tables = $allTables;
+        } else {
+            $tables = array_intersect($allTables, $tables);
+        }
+
+        $this->displayInfoMessage($connectionName, $tables, $action);
+
+        if (empty($tables)) {
+            return;
+        }
 
         $stmts = [];
-        foreach ($schema->listTables() as $table) {
+        foreach ($tables as $table) {
             $table = $schema->describe($table);
             if ($table instanceof TableSchema) {
-                $stmts = array_merge($stmts, $dialect->dropTableSql($table));
+                $stmts = array_merge($stmts, $dialect->{$dialectMethod}($table));
             }
         }
 
@@ -68,28 +114,19 @@ class SchemaCleaner
     }
 
     /**
-     * Truncate all tables of the provided connection.
+     * Display message in info.
      *
      * @param string $connectionName Name of the connection.
+     * @param array $tables Table handled.
+     * @param string $action Action performed.
      * @return void
-     * @throws \Exception if the truncation failed.
      */
-    public function truncate(string $connectionName)
+    protected function displayInfoMessage(string $connectionName, array $tables, string $action): void
     {
-        $this->info('Truncating all tables for connection ' . $connectionName);
-
-        $stmts = [];
-        $schema = static::getSchema($connectionName);
-        $dialect = static::getDialect($connectionName);
-        $tables = $schema->listTables();
-        foreach ($tables as $table) {
-            $table = $schema->describe($table);
-            if ($table instanceof TableSchema) {
-                $stmts = array_merge($stmts, $dialect->truncateTableSql($table));
-            }
-        }
-
-        $this->executeStatements(ConnectionManager::get($connectionName), $stmts);
+        $msg = [ucwords($action)];
+        $msg[] = count($tables) . ' tables';
+        $msg[] = 'for connection ' . $connectionName;
+        $this->info(implode(' ', $msg));
     }
 
     /**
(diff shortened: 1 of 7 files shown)


src/Database/Schema/MysqlSchemaDialect.php
```diff
@@ -67,7 +67,7 @@ public function describeOptionsSql(string $tableName, array $config): array
     /**
      * @inheritDoc
      */
-    public function convertOptionsDescription(TableSchema $schema, array $row): void
+    public function convertOptionsDescription(TableSchemaInterface $schema, array $row): void
```
ADmad, line 70: You need to revert these signature changes. We cannot change method signatures in minor releases as it would be a BC break.

pabloelcolombiano: Alright, will do.

src/TestSuite/Schema/SchemaCleaner.php
```diff
@@ -0,0 +1,126 @@
...
+            });
+        });
+    }
+
+    /**
+     * @param \Cake\Console\ConsoleIo|null $io Console IO.
+     * @param string         $msg Message to display.
+     * @return void
+     */
+    private static function info($io, string $msg): void
```
markstory, line 102:
```suggestion
    private static function info(?ConsoleLog $io, string $msg): void
```

Could have a typehint here.

src/TestSuite/Schema/SchemaCleaner.php
```diff
@@ -0,0 +1,126 @@
...
+{
+    /**
+     * Drop all tables of the provided connection.
+     *
+     * @param string $connectionName Name of the connection.
+     * @param \Cake\Console\ConsoleIo|null $io Console IO to output the processes.
+     * @return void
+     * @throws \Exception if the dropping failed.
+     */
+    public static function drop(string $connectionName, ?ConsoleIo $io = null)
```
markstory, line 39: Any reason these methods need to be static? We could make an instance and pass in the ConsoleIo as a constructor parameter. It makes the other methods a bit simpler.

src/TestSuite/TestConnectionManager.php
```diff
@@ -0,0 +1,94 @@
...
+        }
+    }
+
+    /**
+     * Unset the phinx migration tables from an array of tables.
+     *
+     * @param  string[] $tables Array of strings with table names.
+     * @return array
+     */
+    public static function unsetMigrationTables(array $tables): array
```
markstory, line 47: Couldn't this go on `SchemaCleaner` as well? The other methods in this class are related to connection management. This method feels out of place.

src/TestSuite/Schema/SchemaManager.php
```diff
@@ -0,0 +1,81 @@
...
+    {
+        $files = (array)$file;
+
+        $migrator = new static($verbose);
+
+        SchemaCleaner::drop($connectionName, $migrator->io);
+
+        foreach ($files as $file) {
+            if (!file_exists($file)) {
+                throw new \RuntimeException(__('cake', 'The file {0} was not found.', $file));
```
garas, line 62: Strings for the developer shouldn't be translatable.

src/TestSuite/Schema/SchemaCleaner.php
pabloelcolombiano: Yes, updated.

src/TestSuite/TestConnectionManager.php
pabloelcolombiano: Agreed, done.

src/TestSuite/Schema/SchemaManager.php
pabloelcolombiano: Alright, done

markstory (commented):
Looking good, there are a few style/typechecker errors. If you need help with those let me know.

src/TestSuite/Schema/SchemaCleaner.php
pabloelcolombiano: These methods could be called within setUp or tearDown of tests easily if they are static.
If a given test modifies intentionally the schema, the developer can decide to rebuild the schema by dropping it and re-running the Schema construction (with SchemaManager or Mogrator).

This is the motivation for these being static. The $io is there more for the SchemaManager to pass on its own $io.

Does this make sense?

src/TestSuite/Schema/SchemaCleaner.php
```diff
@@ -0,0 +1,148 @@
...
+    {
+        self::info('Dropping all tables for connection ' . $connectionName, $io);
+
+        $schema = static::getSchema($connectionName);
+        $dialect = static::getDialect($connectionName);
+
+        $stmts = [];
+        foreach ($schema->listTables() as $table) {
+            $table = $schema->describe($table);
+            $stmts = array_merge($stmts, $dialect->dropTableSql($table));
```
pabloelcolombiano, line 48: @markstory I am not sure on how to solve this error. To my understanding, there is an inconsistency in the dropTableSql(), and generally in most of the SchemaDialect's method with a TableSchema and not a TableSchemaInterface as parameter.

How would you solve that?

src/TestSuite/Schema/SchemaCleaner.php
```diff
@@ -0,0 +1,148 @@
...
+        static::info('Truncating all tables for connection ' . $connectionName, $io);
+
+        $stmts = [];
+        $schema = static::getSchema($connectionName);
+        $dialect = static::getDialect($connectionName);
+        $tables = $schema->listTables();
+        $tables = self::unsetMigrationTables($tables);
+        foreach ($tables as $table) {
+            $table = $schema->describe($table);
+            $stmts = array_merge($stmts, $dialect->truncateTableSql($table));
```
pabloelcolombiano, line 73: Same as above.

src/TestSuite/Schema/SchemaCleaner.php
pabloelcolombiano: ps: I am also O.K. to have the schema cleaner methods non-static, especially if you think that it would more fit the Cake spirit.

othercorey:
This is failing all unit tests currently. Can you check that error?

src/TestSuite/Schema/SchemaCleaner.php
markstory: You could use an `instanceof` check to only call dropTableSql on objects that implement the correct interface.

src/TestSuite/Schema/SchemaCleaner.php
markstory: `instanceof` or `method_exists()` is how we solve this in other places.

src/TestSuite/Schema/SchemaCleaner.php
pabloelcolombiano: Agreed, I'll update accordingly

src/TestSuite/Schema/SchemaCleaner.php
pabloelcolombiano: Agreed, I'll update accordingly

src/TestSuite/Schema/SchemaCleaner.php
markstory: My personal bias is to avoid static methods unless they are the only way of doing something. These methods could be exposed via the `FixtureManager` if we require a more convenient way to access them. I suspect that dropping schema is a rare task in applications and it doesn't need to be 'easy'.

src/TestSuite/TestConnectionManager.php
pabloelcolombiano: I've cleaned up a bit that part of the code, which also convinced me that there was way too much staticity there.
Let me know if this works for you!

markstory:
@pabloelcolombiano Would you like a hand with getting the tests fixed on this?

pabloelcolombiano:
@markstory @othercorey I'll have some time planned for that this end of week. I'll ping you if I am blocked somewhere.
The tests run for me locally, so I am surprised. I suspect that the truncation of the test schema has an impact on other tests. 
Any input is welcome, but yes I'll put my hands on that this weekend!

pabloelcolombiano:
@markstory @othercorey I have moved the schema dropping tests at the end of the suite, which seams to fix the tests (at least for me locally).
I am not very happy with that solution, and I am wondering why this is happening.
Aren't the fixtures supposed to create the schema from scratch before every test case?
If this has changed already (as this was the scope of 4.3), then it would explain why dropping the schema in the middle of the suite has the other tests failing.

tests/TestCase/TestSuite/Schema/SchemaManagerTest.php
```diff
@@ -0,0 +1,99 @@
...
+
+    private function listTables(): array
+    {
+        $connection = ConnectionManager::get('test');
+        /** @var SchemaDialect $dialect */
+        $dialect = $connection->getDriver()->schemaDialect();
+
+        $stmt = $dialect->listTablesSql($connection->config());
+
+        return $connection->execute($stmt[0], $stmt[1])->fetchAll(StatementInterface::FETCH_TYPE_NUM);
```
garas, line 97:
```suggestion
        [$sql, $params] = $dialect->listTablesSql($connection->config());

        return $connection->execute($sql, $params)->fetchAll(StatementInterface::FETCH_TYPE_NUM);
```

tests/TestCase/TestSuite/Schema/SchemaCleanerTest.php
```diff
@@ -0,0 +1,132 @@
...
+    public function testDropSchema()
+    {
+        $connection = ConnectionManager::get('test');
+        /** @var SchemaDialect $dialect */
+        $dialect = $connection->getDriver()->schemaDialect();
+
+        $this->createSchemas();
+
+        $stmt = $dialect->listTablesSql($connection->config());
+        $tables = $connection->execute($stmt[0], $stmt[1])->fetch();
```
garas, line 52:
```suggestion
        [$sql, $params] = $dialect->listTablesSql($connection->config());
        $tables = $connection->execute($sql, $params)->fetch();
```

pabloelcolombiano:
@markstory I am stuck right now with the few failing tests left.
* Microsoft server: dropping fails due to foreign key constrain
* MySQL 8 fails at some other place than the tests concerning the present task.
* Postgres generates a silent error.

I will keep on debugging that, but any help there is welcome.

markstory:
> Aren't the fixtures supposed to create the schema from scratch before every test case?

Kind of. The fixture manager will create the tables if they are missing. If the table was created by the fixture manager, it assumes it will also be dropping that table at the end of the test run. Tables aren't expected to be dropped in the middle of a test run.

> If this has changed already (as this was the scope of 4.3), then it would explain why dropping the schema in the middle of the suite has the other tests failing.

We haven't switched the core tests over to data-only fixtures as the way to create schema hasn't been added yet.

src/TestSuite/Schema/SchemaCleaner.php
```diff
@@ -0,0 +1,140 @@
...
+    }
+
+    /**
+     * Drop all tables of the provided connection.
+     *
+     * @param string $connectionName Name of the connection.
+     * @return void
+     * @throws \Exception if the dropping failed.
+     */
+    public function drop(string $connectionName)
```
markstory, line 52:
```suggestion
    public function dropTables(string $connectionName)
```

Should these names indicate that they will operate on all tables in the connection? I can make that change if you're ok with it.

markstory:
I'm able to reproduce the postgres crash locally, and will look into it further.

tests/TestCase/TestSuite/Schema/SchemaCleanerTest.php
```diff
@@ -0,0 +1,132 @@
...
+        $this->assertSame(
+            1,
+            $connection->newQuery()->select('id')->from('test_table2')->execute()->count()
+        );
+
+        $this->expectException(\PDOException::class);
+        $connection->delete('test_table');
+    }
+
+    public function testDropSchema()
```
markstory, line 43: The tests for `SchemaCleaner` are the ones causing tests to crash in postgres. Dropping all the tables in the connection results in an uncaught exception when the `FixtureManager` attempts to clean up fixtures. 

Some potential solutions I can think of are:

1. Test this method with a mocked cleaner that doesn't actually execute SQL.
2. Add parameters to only drop a subset of tables. If that parameter is omitted `$schema->listTables()` can be used.

I'm generally not a fan of mocks but this might be a good fit for them given the destructive nature of dropping tables.

pabloelcolombiano: Thanks for the hint. I'll give option 2 a try.

pabloelcolombiano: @markstory Did some refactoring to pass a set of tables when truncating/refactoring, in the line of the second solution you suggested. Let's see how the CI reacts to that...

markstory:
Looking much better as far as CI is concerned. Looks like there is only a few typechecker errors. Don't worry about the MySQL 8 build right now it is being strange.

pabloelcolombiano:
> Looking much better as far as CI is concerned. Looks like there is only a few typechecker errors. Don't worry about the MySQL 8 build right now it is being strange.

Did a fix to have Psalm clear. Hope that we are close to shipping. Let me know if you have any concerns on the implementation of the code or tests.

othercorey:
If this is ready for review, please mark it. It is currently Draft.

markstory:
> Did a fix to have Psalm clear. Hope that we are close to shipping. Let me know if you have any concerns on the implementation of the code or tests.

Looks good to me. If you make it ready for review we can get it merged in so that the next block of work can be done.

markstory:
Thank you for putting this together! :clap: I guess the next step is to get migrations updated so migrations can be run during tests/bootstrap.php?

PR #15434, opened 2021-04-09
Update exception message for groupBy() and indexBy() missing path

othercorey:
This also adds a missing test for callbacks that return null values.

### DIFF
--- a/src/Collection/CollectionTrait.php
+++ b/src/Collection/CollectionTrait.php
@@ -270,8 +270,8 @@ trait CollectionTrait
             $pathValue = $callback($value);
             if ($pathValue === null) {
                 throw new InvalidArgumentException(
-                    'Cannot use a nonexistent path or null value. ' .
-                    'Use a callback to provide default values if necessary.'
+                    'Cannot group by path that does not exist or contains a null value. ' .
+                    'Use a callback to return a default value for that path.'
                 );
             }
             $group[$pathValue][] = $value;
@@ -291,8 +291,8 @@ trait CollectionTrait
             $pathValue = $callback($value);
             if ($pathValue === null) {
                 throw new InvalidArgumentException(
-                    'Cannot use a nonexistent path or null value. ' .
-                    'Use a callback to provide default values if necessary.'
+                    'Cannot index by path that does not exist or contains a null value. ' .
+                    'Use a callback to return a default value for that path.'
                 );
             }
             $group[$pathValue] = $value;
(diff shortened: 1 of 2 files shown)


src/Collection/CollectionTrait.php
```diff
@@ -269,10 +269,11 @@ public function groupBy($path): CollectionInterface
...
             $pathValue = $callback($value);
             if ($pathValue === null) {
-                throw new InvalidArgumentException(
-                    'Cannot use a nonexistent path or null value. ' .
-                    'Use a callback to provide default values if necessary.'
-                );
+                throw new InvalidArgumentException(sprintf(
+                    'Cannot group by path that does not exist or contains a null value. ' .
+                    'Use a callback to return a default value.',
+                    $path
```
markstory, line 275: Psalm isn't happy about this using sprintf() with no placeholders. Perhaps these should be basic strings?

othercorey: I was printing the path name before I remembered it can be a closure. Will remove it.

src/Collection/CollectionTrait.php
```diff
@@ -270,8 +270,8 @@ public function groupBy($path): CollectionInterface
             $pathValue = $callback($value);
             if ($pathValue === null) {
                 throw new InvalidArgumentException(
-                    'Cannot use a nonexistent path or null value. ' .
-                    'Use a callback to provide default values if necessary.'
+                    'Cannot group by path that does not exist or contains a null value. ' .
+                    'Use a callback to return a default value.',
```
markstory, line 274:
```suggestion
                    'Use a callback to return a default value.'
```

src/Collection/CollectionTrait.php
```diff
@@ -291,8 +291,8 @@ public function indexBy($path): CollectionInterface
             $pathValue = $callback($value);
             if ($pathValue === null) {
                 throw new InvalidArgumentException(
-                    'Cannot use a nonexistent path or null value. ' .
-                    'Use a callback to provide default values if necessary.'
+                    'Cannot index by path that does not exist or contains a null value. ' .
+                    'Use a callback to return a default value.',
```
markstory, line 295:
```suggestion
                    'Use a callback to return a default value.'
```
The trailing commas are an error in older php

othercorey: Apparently removing a parameter is very difficult work.

PR #15437, opened 2021-04-09
Remove the "relative-paths" option.

### LINKED ISSUE #15432: RFC: use full path by default as reference in .pot files when using I18n extract
This is a (multiple allowed):

* CakePHP Version: 4.2.5

### What you did

```bash
bin/cake i18n extract (multiple directories)
```

### What happened
When extracting multiple folders like vendor, templates, src, etc, the references are not clear:

```gettext
#: Error/error400.php:39
msgid "The requested address {0} was not found on this server."
msgstr ""

#: Error/ExceptionRenderer.php:340
msgid "An Internal Error Has Occurred."
msgstr ""
```

With the current behavior, the references start with Error/ but one is in `ROOT/template/Error` and the other is inside `ROOT/vendor/.../src/Error`

To show paths from ROOT:

Applying my changes would show them this way:

```gettext
#: templates/Error/error400.php:39
msgid "The requested address {0} was not found on this server."
msgstr ""

#: vendor/cakephp/cakephp/src/Error/ExceptionRenderer.php:340
msgid "An Internal Error Has Occurred."
msgstr ""
```

I know changing this would change a LOT of lines _.pot_ and _.po_ files first time the user uses _extract command_ but I think it's way cleaner this way.

## Note:
***I found this while writing this RFC** but I still think there may be something we could do.

The default behaviour of the command is supposed to show full paths instead of relative paths. It is actually showing neither the full nor the relative ones. **So what is the expected behaviour?**

I say this because there's an option called `relative-paths` that can be passed to the to the _extract command_.
The DocBlock says:

```php
    /**
     * Use relative paths in the pot files rather than full path
     *
     * @var bool
     */
    protected $_relativePaths = false;
```

When this option is true, the paths will be like this:

```gettext
#: ./templates/Error/error400.php:39
msgid "The requested address {0} was not found on this server."
msgstr ""

#: ./vendor/cakephp/cakephp/src/Error/ExceptionRenderer.php:340
msgid "An Internal Error Has Occurred."
msgstr ""
```

Which is almost the same result I want to get.

I don't know if I have expressed myself correctly...
**At this point I do not know if the actual behavior is the expected one**.

I'm ready to send a PR which, by default, shows use the "full" paths like in the second example but since it does quite the same as `relative-paths` option does, I don't know if we should just change the description of the command and some DocBlocks to state that the default behavior does not use the full path. **What do you think?**

If we do not make any changes, there will still be, in my opinion, confusion in the routes when no `relative-paths` option is passed

ADmad:
Paths in pot file comments are now always relative to the app root. Refs #15432.

### DIFF
--- a/tests/TestCase/Command/I18nExtractCommandTest.php
+++ b/tests/TestCase/Command/I18nExtractCommandTest.php
@@ -84,38 +84,39 @@ class I18nExtractCommandTest extends TestCase
 
         $this->assertFileDoesNotExist($this->path . DS . 'cake.pot');
 
-        // extract.ctp
-        $pattern = '/\#: [\/\\\\]extract\.php:\d+\n';
-        $pattern .= '\#: [\/\\\\]extract\.php:\d+\n';
-        $pattern .= 'msgid "You have %d new message."\nmsgid_plural "You have %d new messages."/';
+        // The additional "./tests/test_app" is just due to the wonky folder structure of the test app.
+        // In a regular app the path would start with "./templates".
+        $pattern = '@\#: \./tests/test_app/templates/Pages/extract\.php:\d+\n';
+        $pattern .= '\#: \./tests/test_app/templates/Pages/extract\.php:\d+\n';
+        $pattern .= 'msgid "You have %d new message."\nmsgid_plural "You have %d new messages."@';
         $this->assertMatchesRegularExpression($pattern, $result);
 
         $pattern = '/msgid "You have %d new message."\nmsgstr ""/';
         $this->assertDoesNotMatchRegularExpression($pattern, $result, 'No duplicate msgid');
 
-        $pattern = '/\#: [\/\\\\]extract\.php:\d+\n';
-        $pattern .= 'msgid "You deleted %d message."\nmsgid_plural "You deleted %d messages."/';
+        $pattern = '@\#: \./tests/test_app/templates/Pages/extract\.php:\d+\n';
+        $pattern .= 'msgid "You deleted %d message."\nmsgid_plural "You deleted %d messages."@';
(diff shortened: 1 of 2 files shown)


tests/TestCase/Command/I18nExtractCommandTest.php
```diff
@@ -84,20 +84,21 @@ public function testExecute()
 
         $this->assertFileDoesNotExist($this->path . DS . 'cake.pot');
 
-        // extract.ctp
-        $pattern = '/\#: [\/\\\\]extract\.php:\d+\n';
-        $pattern .= '\#: [\/\\\\]extract\.php:\d+\n';
+        // The additional "./tests/test_app" is just due to the wonky folder structure of the test app.
+        // In a regular app the path would start with "./templates".
+        $pattern = '/\#: \.[\/\\\\]tests[\/\\\\]test_app[\/\\\\]templates[\/\\\\]Pages[\/\\\\]extract\.php:\d+\n';
```
garas, line 89: Paths are always generated with `/`, pattern can be simplified.

markstory (approved):
Seems reasonable to me. I can merge this when the 4.3 guide is updated.

garas:
Add notice in `buildOptionParser` that `relative-paths` is ignored/deprecated.

ADmad:
@garas You mean update the `help` text for the option to state the option is unused?

othercorey:
> @garas You mean update the `help` text for the option to state the option is unused?

We should probably remove the unused option so it throws an error.

ADmad:
> We should probably remove the unused option so it throws an error.

Right, silly me missed that, done now.

PR #15439, opened 2021-04-09
Fix asset symlink and Filesystem::deleteDir() on windows

othercorey:
Let's see if these run correctly with github actions.


markstory:
Looks good.

othercorey:
> Looks good.

Need to tweak a couple tests to work properly instead of throwing warnings (existing directory, etc).

PR #15440, opened 2021-04-10
Change rand() to mt_rand()

yama:
thanks.


ADmad:
2.x has reached EOL and won't have any new releases except for security updates.

PR #15442, opened 2021-04-10
Invert db version + php version

othercorey:
Test latest db with oldest php vesrion and oldest db with latest php version.

This might also speed up the very slow mysql 8 + php 7.4 + code coverage build.


othercorey:
Moved to 4.3 in https://github.com/cakephp/cakephp/commit/f89d2483d16746fed6fef8f6aa2529b5d871fb9a

PR #15444, opened 2021-04-10
Simplify TreeBehavior::recover()

othercorey:
This cleans up the tree recovery code so the lft/rght counter starts at 1 and level starts at 0.

The nodes are updated directly instead of updating the parent after processing.

Reduced unnecessary variables and removed aliasing which isn't used anywhere else (and isn't needed as the table isn't joined with anything).

### DIFF
--- a/src/ORM/Behavior/TreeBehavior.php
+++ b/src/ORM/Behavior/TreeBehavior.php
@@ -828,47 +828,41 @@ class TreeBehavior extends Behavior
     /**
      * Recursive method used to recover a single level of the tree
      *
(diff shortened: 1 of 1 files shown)


ADmad:
While it's highly unlikely someone would have overriden this method, it's better to make such changes which change param names or default values, in a minor release instead of patch release.

othercorey:
Of course. I didn't start with the default value change and thought of it as internal.

othercorey:
Moved to 4.3 in https://github.com/cakephp/cakephp/commit/4eedbbf5bcaeb9461a7804aac2a0380d9104678e

PR #15445, opened 2021-04-11
Fix typed variadic string parameter resolution

### LINKED ISSUE #15441: Controller action invoked with null for variadic string parameter
This is a (multiple allowed):

* [x] bug
* [ ] enhancement

* CakePHP Version: 4.2

### What you did
https://github.com/cakephp/app/issues/833

Access `PagesController::display()` from default app skeleton using URL `/pages/`.

 ```
public function display(string ...$path): ?Response
```

### What happened
In CakePHP 4.2, action is called with one argument `null`, which triggers `[TypeError] Argument 1 passed to App\Controller\PagesController::display() must be of the type string, null given`.

Action should receive no arguments like in CakePHP 4.1.

https://github.com/cakephp/cakephp/pull/15241 fixed similar problem, but this issue still exists.

### PR DESCRIPTION
**Pull request by markstory**
Fixes #15441

### DIFF
--- a/tests/test_app/TestApp/Controller/DependenciesController.php
+++ b/tests/test_app/TestApp/Controller/DependenciesController.php
@@ -47,7 +47,7 @@ class DependenciesController extends Controller
         return $this->response->withStringBody(json_encode(['args' => func_get_args()]));
     }
 
-    public function spread(...$args)
+    public function spread(string ...$args)
     {
         return $this->response->withStringBody(json_encode(['args' => $args]));
     }
--- a/tests/test_app/TestApp/Controller/DependenciesController.php
+++ b/tests/test_app/TestApp/Controller/DependenciesController.php
@@ -27,6 +27,11 @@ class DependenciesController extends Controller
         $this->inject = $inject;
     }
 
+    public function requiredString(string $str)
+    {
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: tests/test_app/TestApp/Controller/DependenciesController.php [hunk 1]
```diff
@@ -47,7 +47,7 @@ public function variadic()
         return $this->response->withStringBody(json_encode(['args' => func_get_args()]));
     }
 
-    public function spread(...$args)
+    public function spread(string ...$args)
```
othercorey, line 50: Was this receiving null and converting to an empty string somewhere passing the test?

othercorey:
Looks like there isn't a test that passes [] to the variadic action.

tests/test_app/TestApp/Controller/DependenciesController.php
markstory: It was getting a null, which was silently cast to an empty array by the spread operator.

markstory:
There is an existing test that sends no parameters. It failed when I added the string type and no longer fails.

othercorey:
> There is an existing test that sends no parameters. It failed when I added the string type and no longer fails.

I mean the code coverage error.

othercorey:
Looks ok now.

PR #15446, opened 2021-04-11
Organizes output of `bin/cake routes` into columns.

cnizzardini:
The motivation behind this PR is cleaning up the output of `bin/cake routes` so data is easier to read for developers maintaining applications with a large number of routes. This change:

- Sorts alphabetically by route name (edit: optionally via `--sort`)
- Previous JSON display is maintained (edit: with the existing) `--verbose` switch
- Updates unit test

Sample output:

```
> bin/cake routes
+-------------------------------------+-------------------------------------------+-------------+-------------------+-------------+-----------+---------------+
| Route name                          | URI template                              | Controller  | Action            | Plugin      | Prefix    | Method(s)     |
+-------------------------------------+-------------------------------------------+-------------+-------------------+-------------+-----------+---------------+
| _controller:_action                 | /{controller}/{action}/*                  |             | index             |             |           |               |
| actors:edit                         | /actors/:id                               | Actors      | edit              |             |           | PUT, PATCH    |
| admin:setup._controller:_action     | /admin/setup/{controller}/{action}/*      |             | index             | Setup       | Admin     |               |
```

Adding the `--verbose` or `-v` switch simply appends a column called Defaults to the end of the rows.

I had considered adding some additional switches such as `--plugin` for limiting to a specific plugin and `--no-plugins` for hiding all plugins, but I think these use-cases can be accomplished with `grep`. This keeps the change relatively small and eliminates the need for additional unit tests.

### DIFF
--- a/src/Command/RoutesCommand.php
+++ b/src/Command/RoutesCommand.php
@@ -35,14 +35,43 @@ class RoutesCommand extends Command
      */
     public function execute(Arguments $args, ConsoleIo $io): ?int
     {
-        $output = [
-            ['Route name', 'URI template', 'Defaults'],
-        ];
-        foreach (Router::routes() as $route) {
-            $name = $route->options['_name'] ?? $route->getName();
-            ksort($route->defaults);
-            $output[] = [$name, $route->template, json_encode($route->defaults)];
+        $header = ['Route name', 'URI template', 'Controller', 'Action', 'Plugin', 'Prefix', 'Method(s)'];
+        if ($args->getOption('verbose') !== null) {
+            $header[] = 'Defaults';
         }
+
+        $output = [];
+
+        foreach (Router::routes() as $route) {
+            $methods = '';
+            if (isset($route->defaults['_method']) && is_array($route->defaults['_method'])) {
+                $methods = implode(', ', $route->defaults['_method']);
+            }
+
+            $item = [
+                $route->options['_name'] ?? $route->getName(),
+                $route->template,
+                $route->defaults['controller'] ?? '',
+                $route->defaults['action'] ?? '',
+                $route->defaults['plugin'] ?? '',
+                $route->defaults['prefix'] ?? '',
+                $methods,
+            ];
+
+            if ($args->getOption('verbose') !== null) {
+                ksort($route->defaults);
+                $item[] = json_encode($route->defaults);
+            }
+
+            $output[] = $item;
+        }
+
+        usort($output, function ($a, $b) {
+            return strcasecmp($a[0], $b[0]);
+        });
+
+        array_unshift($output, $header);
+
         $io->helper('table')->output($output);
         $io->out();
 
@@ -57,7 +86,12 @@ class RoutesCommand extends Command
      */
     public function buildOptionParser(ConsoleOptionParser $parser): ConsoleOptionParser
     {
-        $parser->setDescription('Get the list of routes connected in this application.');
+        $parser
+            ->setDescription('Get the list of routes connected in this application.')
+            ->addOption('verbose', [
+                'help' => 'Display verbose output',
+                'short' => 'v',
+            ]);
 
         return $parser;
     }
--- a/src/Command/RoutesCommand.php
+++ b/src/Command/RoutesCommand.php
@@ -58,7 +58,7 @@ class RoutesCommand extends Command
                 $methods,
             ];
 
-            if ($args->getOption('verbose') !== null) {
+            if ($args->getOption('verbose') === true) {
                 ksort($route->defaults);
                 $item[] = json_encode($route->defaults);
             }
@@ -66,7 +66,7 @@ class RoutesCommand extends Command
             $output[] = $item;
         }
 
-        if ($args->getOption('sort') !== null) {
+        if ($args->getOption('sort') === true) {
             usort($output, function ($a, $b) {
                 return strcasecmp($a[0], $b[0]);
             });
@@ -93,6 +93,7 @@ class RoutesCommand extends Command
             ->addOption('sort', [
                 'help' => 'Sorts alphabetically by route name A-Z',
                 'short' => 's',
+                'boolean' => true,
             ]);
 
         return $parser;
--- a/src/Command/RoutesCommand.php
+++ b/src/Command/RoutesCommand.php
@@ -36,7 +36,7 @@ class RoutesCommand extends Command
     public function execute(Arguments $args, ConsoleIo $io): ?int
     {
         $header = ['Route name', 'URI template', 'Controller', 'Action', 'Plugin', 'Prefix', 'Method(s)'];
-        if ($args->getOption('verbose') !== null) {
+        if ($args->getOption('verbose')) {
             $header[] = 'Defaults';
         }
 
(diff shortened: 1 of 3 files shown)


cnizzardini:
If someone wants to point me in the right direction I can add coverage for this line: https://github.com/cakephp/cakephp/pull/15446/checks?check_run_id=2317637645

ADmad:
> Sorts alphabetically by route name

I don't like this. Currently the routes are listed in the order in which they are checked in `RoutesCollection::parseRequest()`. That's more useful when debugging route matching issue.

ndm2:
The sorting of [`Router::routes()` was just recently updated](https://github.com/cakephp/cakephp/pull/15248) so that it matches the order in which routes are actually being matched, which is not necessarily identical with the order in which they are connected, hence that specific order can help a lot to figure out problems with non-matching routes, hence personally I'd like to not see this change.

How about introducing an option for different sort orders?

cnizzardini:
I am fine with that. I added a `--sort` switch and there is no sorting by default now. The unit test uses `-s` for coverage purposes.

src/Command/RoutesCommand.php
```diff
@@ -57,7 +88,16 @@ public function execute(Arguments $args, ConsoleIo $io): ?int
...
-        $parser->setDescription('Get the list of routes connected in this application.');
+        $parser
+            ->setDescription('Get the list of routes connected in this application.')
+            ->addOption('verbose', [
+                'help' => 'Display verbose output',
+                'short' => 'v',
+            ])
+            ->addOption('sort', [
+                'help' => 'Sorts alphabetically by route name A-Z',
+                'short' => 's',
```
garas, line 99:
```suggestion
                'short' => 's',
                'boolean' => true,
```

Then `$args->getOption('sort')` will be `true` or `false`

cnizzardini: What's the downside of a simple null check?

src/Command/RoutesCommand.php
```diff
@@ -57,7 +88,16 @@ public function execute(Arguments $args, ConsoleIo $io): ?int
      */
     public function buildOptionParser(ConsoleOptionParser $parser): ConsoleOptionParser
     {
-        $parser->setDescription('Get the list of routes connected in this application.');
+        $parser
+            ->setDescription('Get the list of routes connected in this application.')
+            ->addOption('verbose', [
```
garas, line 93: `verbose`/`quite` options already exist by default

src/Command/RoutesCommand.php
garas: Booleans doesn't have value.

Option now consumes next argument as its value

`bin/cake routes --sort --verbose` will set `sort=--verbose`...

src/Command/RoutesCommand.php
```diff
@@ -35,14 +35,45 @@ class RoutesCommand extends Command
...
+                $route->options['_name'] ?? $route->getName(),
+                $route->template,
+                $route->defaults['controller'] ?? '',
+                $route->defaults['action'] ?? '',
+                $route->defaults['plugin'] ?? '',
+                $route->defaults['prefix'] ?? '',
+                $methods,
+            ];
+
+            if ($args->getOption('verbose') === true) {
```
garas, line 61:
```suggestion
            if ($args->getOption('verbose')) {
```

src/Command/RoutesCommand.php
```diff
@@ -35,14 +35,45 @@ class RoutesCommand extends Command
      */
     public function execute(Arguments $args, ConsoleIo $io): ?int
     {
-        $output = [
-            ['Route name', 'URI template', 'Defaults'],
-        ];
+        $header = ['Route name', 'URI template', 'Controller', 'Action', 'Plugin', 'Prefix', 'Method(s)'];
```
ADmad, line 38: Having `Prefix` and `Plugin` before `Controller` would be more inline with how URL parts are generally mapped.

tests/TestCase/Command/RoutesCommandTest.php
```diff
@@ -71,26 +71,46 @@ public function testRouteListHelp()
      */
     public function testRouteList()
     {
-        $this->exec('routes');
+        $this->exec('routes -v -s');
```
ADmad, line 74: There should be test for `routes` too without any additional options.

src/Command/RoutesCommand.php
garas: More exact order is `Plugin` before `Prefix`.

src/Command/RoutesCommand.php
```diff
@@ -35,14 +35,45 @@ class RoutesCommand extends Command
...
+        }
+
+        $output = [];
+
         foreach (Router::routes() as $route) {
-            $name = $route->options['_name'] ?? $route->getName();
-            ksort($route->defaults);
-            $output[] = [$name, $route->template, json_encode($route->defaults)];
+            $methods = '';
+            if (isset($route->defaults['_method']) && is_array($route->defaults['_method'])) {
```
ADmad, line 47: You aren't showing the method when it's a string?

tests/TestCase/Command/RoutesCommandTest.php
cnizzardini: I think this meets near full coverage as-is. I could add a smaller test for just `-v -s` to test order and the additional column if you think that's better.

cnizzardini: @ADmad Is there an easy way to test the order of the output when using `--sort`?

cnizzardini: Cuz you ain't gonna like what I'm cooking up buddy.

cnizzardini: I just pulled something out of the oven for you buddy. Bon Appétit!

src/TestSuite/ConsoleIntegrationTestTrait.php
```diff
@@ -144,6 +144,16 @@ public function useCommandRunner(): void
         $this->_useCommandRunner = true;
     }
 
+    /**
+     * Get an array of messages from \Cake\TestSuite\Stub\ConsoleOutput::messages()
+     *
+     * @return array
+     */
+    public function getMessages(): array
```
ADmad, line 152: If we are going to add this method then better to name is `getOutMessage()`.

cnizzardini: Would it be better just to provide a getter for `\Cake\TestSuite\Stub\ConsoleOutput`?

ADmad: Well we could just use the protected property `$this->_out` directly too.

Whether we need a public interface for the ConsoleOutput instance can be discussed in a separate issue/PR.

cnizzardini: No, getMessages() is pointless. Not sure why I thought `_out` was private. I removed it and just accessed `_out` directly.

ADmad (approved):

markstory:
Great job @cnizzardini :clap:

PR #15449, opened 2021-04-11
Use source Table connection when creating association target Table

### LINKED ISSUE #742: Connection failure checking CounterCache using non-default connection
`Association::getTarget()` constructs the Table object using default connection instead of through `ModelCommand::getTableObject()` which uses the custom connection from `-c` option.

This table has a composite key so some of the association code is skipped.

```
One moment while associations are detected.
Bake cannot generate associations for composite primary keys at this time.
Exception: Connection to Postgres could not be established: SQLSTATE[08006] [7] could not translate host name "port=" to address: Unknown host
In [vendor\cakephp\cakephp\src\Database\Driver.php, line 140]

2021-04-10 22:57:20 Error: [Cake\Database\Exception\MissingConnectionException] Connection to Postgres could not be established: SQLSTATE[08006] [7] could not translate host name "port=" to address: Unknown host in Stack Trace:
- vendor\cakephp\cakephp\src\Database\Driver\Postgres.php:109
- vendor\cakephp\cakephp\src\Database\Schema\SchemaDialect.php:47
- vendor\cakephp\cakephp\src\Database\Driver\Postgres.php:146
- vendor\cakephp\cakephp\src\Database\Schema\Collection.php:53
- vendor\cakephp\cakephp\src\Database\Connection.php:399
- vendor\cakephp\cakephp\src\ORM\Table.php:512
- vendor\cakephp\bake\src\Command\ModelCommand.php:927
- vendor\cakephp\bake\src\Command\ModelCommand.php:904
- vendor\cakephp\bake\src\Command\ModelCommand.php:132
- vendor\cakephp\bake\src\Command\ModelCommand.php:98
- vendor\cakephp\bake\src\Command\ModelCommand.php:81
- vendor\cakephp\cakephp\src\Console\BaseCommand.php:179
- vendor\cakephp\cakephp\src\Console\CommandRunner.php:336
- vendor\cakephp\cakephp\src\Console\CommandRunner.php:172
- bin\cake.php:11
```

### LINKED ISSUE #15448: Association::getTarget() always uses default connection
If the table doesn't exist in the registry, Association calls TableLocator::get() with the default connection

```
$config = [];
$exists = $tableLocator->exists($registryAlias);
if (!$exists) {
    $config = ['className' => $this->_className];
}
$this->_targetTable = $tableLocator->get($registryAlias, $config);
```

This causes errors if the source table uses a different connection. This situation is possible in Bake.

othercorey:
Fixes https://github.com/cakephp/cakephp/issues/15448
Fixes https://github.com/cakephp/bake/issues/742

There doesn't seem to be a test specific to creating a target Table instance from `getTarget()`.

### DIFF
--- a/src/ORM/Association.php
+++ b/src/ORM/Association.php
@@ -391,7 +391,7 @@ abstract class Association
             $config = [];
             $exists = $tableLocator->exists($registryAlias);
             if (!$exists) {
-                $config = ['className' => $this->_className];
+                $config = ['className' => $this->_className, 'connection' => $this->getSource()->getConnection()];
             }
             $this->_targetTable = $tableLocator->get($registryAlias, $config);
 
--- a/tests/TestCase/ORM/Behavior/TranslateBehaviorTest.php
+++ b/tests/TestCase/ORM/Behavior/TranslateBehaviorTest.php
@@ -93,7 +93,6 @@ class TranslateBehaviorTest extends TestCase
 
         $this->assertSame('CustomI18n', $i18n->getName());
         $this->assertInstanceOf(CustomI18nTable::class, $i18n->getTarget());
-        $this->assertSame('test_custom_i18n_datasource', $i18n->getTarget()->getConnection()->configName());
         $this->assertSame('custom_i18n_table', $i18n->getTarget()->getTable());
     }
 


tests/TestCase/ORM/Behavior/TranslateBehaviorTest.php
```diff
@@ -93,7 +93,6 @@ public function testCustomTranslationTable()
 
         $this->assertSame('CustomI18n', $i18n->getName());
         $this->assertInstanceOf(CustomI18nTable::class, $i18n->getTarget());
-        $this->assertSame('test_custom_i18n_datasource', $i18n->getTarget()->getConnection()->configName());
```
othercorey, line 96: Associations can't use different connections. This isn't the right place to test custom default connection name for Table class.

markstory: Associations can use different connections, however they need to use the `select` strategy and not join/subquery.

src/ORM/Association.php
```diff
@@ -391,7 +391,7 @@ public function getTarget(): Table
             $config = [];
             $exists = $tableLocator->exists($registryAlias);
             if (!$exists) {
-                $config = ['className' => $this->_className];
+                $config = ['className' => $this->_className, 'connection' => $this->getSource()->getConnection()];
```
markstory, line 394: Shouldn't the concrete class for the association target define its connection?

ADmad: Explicitly setting the connection here would mean it would override any non default connection specified in the table class isn't it? That isn't good.

othercorey: What do you propose here? Any associated generated from a key would use the same connection.

ndm2: Maybe check after `get()` if `$this->_targetTable` is `TableLocator::$fallbackClassName` (needs a public getter), and if so, set the connection on the constructed table object?

othercorey: The issue I'm trying to solve is the bake issue noted in the comment. We override the connection which is used everywhere for Table creation except here where the table doesn't exist before calling `getTarget()`.

In those cases, the default connection is overridden by the `-c` connection option.

ndm2: Ah, sorry, I should've read the initial post. If my concrete table class should use the default connection, then I certainly wouldn't want that to change in my app.

Maybe this is something that `ModelCommand::getTableObject()` should enforce? Currently it's only doing that for not yet constructed table objects, meaning if they get constructed somewhere else in the code before that, they will have the "wrong" connection, assuming the user actually expects `-c` to modify the connections of all involved tables.

othercorey: Are you suggesting that getTableObject() walk through all associations and set the connection flag? I'm not sure that's the order of operations in ModelCommand which is generating the associations.

ndm2: No, I mean just the table that `getTableObject()` obtains. Not sure if I'm missing something, but aren't the tables obtained via that method the only ones that the command actually makes use of with regards to establishing connections?

PR #15451, opened 2021-04-12
Cleanup callable for marshalling '_translations' property.

ADmad:
Removed unnecessary loop.

Also fixed setting of errors for translated records in the primary entity.
For .e.g. the error should be set for '_translations.{lang}.{field}' not '{lang}.{field}', for the error to be properly displayed in the form.

### DIFF
--- a/src/ORM/Behavior/Translate/TranslateStrategyTrait.php
+++ b/src/ORM/Behavior/Translate/TranslateStrategyTrait.php
@@ -152,27 +152,32 @@ trait TranslateStrategyTrait
 
         return [
             '_translations' => function ($value, $entity) use ($marshaller, $options) {
-                /** @var \Cake\Datasource\EntityInterface $entity */
+                if (!is_array($value)) {
+                    return null;
+                }
+
+                /** @var array<string, \Cake\Datasource\EntityInterface>|null $translations */
                 $translations = $entity->get('_translations');
-                foreach ($this->_config['fields'] as $field) {
-                    $options['validate'] = $this->_config['validator'];
-                    $errors = [];
-                    if (!is_array($value)) {
-                        return null;
+                if ($translations === null) {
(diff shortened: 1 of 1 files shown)


tests/TestCase/ORM/Behavior/TranslateBehaviorTest.php
```diff
@@ -1847,7 +1854,7 @@ public function testBuildMarshalMapBuildEntitiesValidationErrors()
                 '_empty' => 'This field cannot be left empty',
             ],
         ];
-        $this->assertEquals($expected, $entity->getError('es'));
+        $this->assertEquals($expected, $entity->getError('_translations.es'));
```
markstory, line 1857: This seems like the kind of change that could cause problems for userland code as the key errors are stored in is changing. This does match up with the documentation better now, but I suspect we might still get an issue about the change.

ADmad: With the current array structure the errors wouldn't be displayed under the related fields using `FormHelper::control()`.

There's a possibility the someone could be using `FormHelper::error()` to fetch and display the errors at appropriate location, but that seems quite unlikely. 

That said I can move the error key change to 4.next if we want to be extra safe.

markstory: I'm fine either way. We should think about what our response will be if we get issues that the behavior 'broke'.

ADmad: Meh, I will just move the array keying fix to 4.next :)

ADmad: Removed the error key changes and just kept the cleanup.

PR #15453, opened 2021-04-15
Allow new fixtures to use the data only loader

markstory:
Convert the active fixture loader to a singleton. This allows the
PHPUnit extension/listener to inject its own implementation. TestCases
now pull from this singleton as the extension cannot set properties, and
I'd like both fixture systems to have properties available with the same
sequencing.

### DIFF
--- a/src/TestSuite/Fixture/FixtureLoader.php
+++ b/src/TestSuite/Fixture/FixtureLoader.php
@@ -1,5 +1,4 @@
 <?php
-
 declare(strict_types=1);
 
 /**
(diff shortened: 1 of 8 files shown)


src/TestSuite/Fixture/FixtureLoader.php
```diff
@@ -0,0 +1,97 @@
...
+     */
+    private static $instance;
+
+    /**
+     * Set the shared instance
+     *
+     * @param \Cake\TestSuite\Fixture\FixtureLoader $instance The instance to set.
+     * @return void
+     */
+    public static function setInstance(FixtureLoader $instance): void
```
othercorey, line 41: Can we create a factory method and a protected constructor or does it have to be 'set`?

markstory: If the constructor was protected how would each fixture system set the manager it needs? I don't love the approach I have here but it was simple and would let me move the overall implementation forward.

PR #15456, opened 2021-04-16
Fixed setting of errors for translated records in the primary entity.

ADmad:
For .e.g. the error should be set for key '_translations.{lang}.{field}' not
'{lang}.{field}', for the error to be properly displayed in the form.


othercorey:
Looks ok to me, but I don't know translations very well.

Issue #15457, opened 2021-04-17
RFC: Preloading


cnizzardini:
* [ ] bug
* [ ] enhancement
* [X] feature-discussion (RFC)

https://www.php.net/manual/en/opcache.preloading.php

There was an issue (#13950) opened on this a while back and core developers wanted to wait on composer support (composer/composer#7777) which now seems dead. I think this is something CakePHP should be offering its users as other major frameworks have already implemented support.

I haven't done much with preloading, but it seems CakePHP could offer a Command for generating a preload file that might include CakePHP `vendor/cakephp` and some (or all) of userland `src/` and `plugins/`.

cnizzardini:
I might take a stab today at building some alpha code for this, but I likely need some assistance from core devs.

markstory:
I have never worked with preloading either, but I can try to help.

cnizzardini:
If I am understanding what I've read, you only gain preload performance if all of the dependencies of a preloaded file are preloaded as well. It's semi-trivial to grab a list of all `CAKE` and `APP` files. But the moment those files have dependencies on other vendor files you lose the benefit (again if I am understanding correctly). I think tools can be provided to assist users with managing their own `APP` preloads, the framework shouldn't get into that mess. For Cake specifics, not all of these dependencies are well known to me in terms of usage in the various area of Cake core:

```json
        "composer/ca-bundle": "^1.2",
        "laminas/laminas-diactoros": "^2.2.2",
        "laminas/laminas-httphandlerrunner": "^1.1",
        "league/container": "^3.2",
        "psr/http-client": "^1.0",
        "psr/http-server-handler": "^1.0",
        "psr/http-server-middleware": "^1.0",
        "psr/log": "^1.0.0",
        "psr/simple-cache": "^1.0.0"
```

I've also come across some packages which assist in managing building preloads, but I'd imagine if Cake were to leverage a library we'd want to ensure it's from a quality maintainer that won't "disappear". I'm not necessarily recommending that at this time, but it's an option.

For the additional cake core dependencies, it should be possible to grab all the files in their vendor directory. FWIW I do have some rudimentary code that can be run by command or integrated into userland code that I'll try and open a PR with this weekend or shortly after. 

Benchmarks can be tricky, but with what I wrote I am seeing about a 15% "grain of salt" boost on a dockerized application running on my local Ubuntu.

**WITHOUT: opcache.preload:**

```console
> ab -n 10000 -c 10 http://localhost:8080/public/actors.json

[...]

PR #15458, opened 2021-04-18
RFC: Generate OPcache preload file

cnizzardini:
CakePHP users would gain performance benefits from a means of easily generating preload files for use in `opcache.preload` (see: https://www.php.net/manual/en/opcache.preloading.php).

See: RFC: Preloading #15457

### Initial Implementation

By default, Preload will generate a list of Cake Core files to add to the preload file. This generates a relatively small amount of files that can be preloaded into a reasonable amount of memory depending on the user's OPcache settings (see: https://www.php.net/manual/en/opcache.configuration.php).

### Preload Command

```console
/srv/app $ bin/cake preload --help
Generate a preload file

Usage:
cake preload [options]

Options:

--app           Add your applications src directory into the preloader
--help, -h      Display this help.
--name          The preload file path. (default:
                /srv/app/preload.php)
--packages      A comma separated list of packages (e.g.
                vendor-name/package-name) to add to the preloader
--plugins       A comma separated list of your plugins to load or `*` to
                load all plugins/*
--quiet, -q     Enable quiet output.
--verbose, -v   Enable verbose output.

```

Examples:

Default only loads in CakePHP core files, excluding TestSuite and Console/Command/Shell namespaces:
```console
bin/cake preload 
```

Include composer packages:
```console
bin/cake preload --packages=cakephp/authentication,cakephp/chronos
```

Include userland application
```console
bin/cake preload --app
```

Include userland plugins
```console
bin/cake preload --plugins=*
```
or
```console
bin/cake preload --plugins=MyPlugin,MyOtherPlugin
```

### Preload Class

Writes cake core files to `ROOT . DS . 'preload.php'`

```php
(new Preload())->write();
```

Users can build something more elaborate using the Preload class methods `loadPath()` and writing their own callback method. It just uses FileSystem under the hood.

### Performance

Obviously, these types of benchmarks should be taken with a bit of a gain of salt. I benchmarked this using apache bench with this project here: https://github.com/mixerapi/app which is a dockerized REST API (LEMP stack on alpine + php-fpm 7.4). `DEBUG` was set to false.

```ini
[php]
session.auto_start = Off
short_open_tag = Off
opcache.preload_user=root
opcache.preload=/srv/app/preload.php
opcache.interned_strings_buffer = 16
opcache.max_accelerated_files = 20000
opcache.memory_consumption = 256
opcache.enable_cli = 0
opcache.enable = 1
opcache.revalidate_freq = 360
opcache.fast_shutdown = 1
realpath_cache_size = 4096K
realpath_cache_ttl = 600
```

Note: `opcache.preload_user=root` and `opcache.preload=/srv/app/preload.php` were disabled for the no preload run.

Command:

```console
ab -n 10000 -c 10 http://localhost:8080/public/actors.json
```

I ran each 3 times:

| Type      | Run 1 | Run 2 | Run 3 |
| ----------- | ----------- | ----------- | ----------- |
| No Preload | 301.30 [#/sec] (mean) |  335.12 [#/sec] (mean) | 322.41 [#/sec] (mean) |
| `CAKE` only | 447.92 [#/sec] (mean) |  448.48 [#/sec] (mean) | 446.53 [#/sec] (mean) |
| `CAKE` + `APP` | 457.62 [#/sec] (mean) |  455.40 [#/sec] (mean) | 394.89 [#/sec] (mean) |


ADmad:
Has anyone checked/used https://github.com/DarkGhostHunter/Preloader ?

cnizzardini:
> Has anyone checked/used https://github.com/DarkGhostHunter/Preloader ?

Checked yes, used no. It seems to base what is added to the preload based off of opcache script stats. Which can keep preload smaller but also might miss things. 

I should compare it with what I wrote.

cnizzardini:
I took a look at Symfony's Preloader (https://github.com/symfony/dependency-injection/blob/5.x/Dumper/Preloader.php) and how it generates for an application I can work on. 

You get output like this from it:

```php
use Symfony\Component\DependencyInjection\Dumper\Preloader;

if (in_array(PHP_SAPI, ['cli', 'phpdbg'], true)) {
    return;
}

require dirname(__DIR__, 3).'/vendor/autoload.php';
require __DIR__.'/Container1UVgufq/App_KernelDevDebugContainer.php';
// ...many more require calls...

$classes = [];
$classes[] = 'Symfony\Bundle\FrameworkBundle\FrameworkBundle';
// ... many more classes

Preloader::preload($classes);
```

That class list is likely generated from what you define in `services.yaml` for their global auto wiring approach. That method seems to use reflection and recursion to grab class properties and require those files. It will be interesting to see how that other library works compared to symfony and my prototype.

cnizzardini:
I played around with a few different ways of doing this, but ultimately settled a simpler approach which mostly leverages FileSystem but still leaves things open for users to do their own thing.

I played with DarkGhostHunter/Preloader, but I think most CakePHP users are looking for something simple that can just add to their build process and `bin/cake preload` is pretty simple.

Let me know what yall think.

src/Command/PreloadCommand.php
```diff
@@ -0,0 +1,173 @@
...
+        $packages = explode(',', $args->getOption('packages'));
+        $packages = array_filter($packages, function ($package) {
+            $path = ROOT . DS . 'vendor' . DS . $package;
+            if (file_exists($path)) {
+                return true;
+            }
+            triggerWarning("Package $package could not be located at $path");
+        });
+
+        $packages = array_map(function($package){
```
cnizzardini, line 109: Odd order of operations here, array_map() should be moved above array_filter().

src/Preload/Preloader.php
```diff
@@ -0,0 +1,168 @@
...
+     * @throws \LogicException
+     * @throws \RuntimeException
+     */
+    public function write(string $path = ROOT . DS . 'preload.php'): bool
+    {
+        if (!function_exists('opcache_get_status')) {
+            throw new RuntimeException('OPcache must be enabled');
+        }
+
+        if ((file_exists($path) && !is_writable($path)) || !is_writable(ROOT)) {
```
cnizzardini, line 119: Checking if ROOT is_writable makes no sense. Need look at this more.

othercorey:
We're thinking this should probably go into a plugin to receive community feedback. We expect a lot of churn on the "optimal" implementation and don't think that's good for patch releases.

othercorey:
> We're thinking this should probably go into a plugin to receive community feedback. We expect a lot of churn on the "optimal" implementation and don't think that's good for patch releases.

@cnizzardini Can you move this to a personal plugin repo and try posting it to the slack channel or discourse.cakephp.org for feedback?

cnizzardini:
> > We're thinking this should probably go into a plugin to receive community feedback. We expect a lot of churn on the "optimal" implementation and don't think that's good for patch releases.
> 
> @cnizzardini Can you move this to a personal plugin repo and try posting it to the slack channel or discourse.cakephp.org for feedback?

Yeah, I should have some time to that.

PR #15459, opened 2021-04-19
Add a formal interface for fixture state managers

markstory:
Add a more formal interface for data-only fixture state managers. This will be useful for userland implementations or plugins like ElasticSearch which will need their own fixture state management.

### DIFF
--- a/src/TestSuite/FixtureSchemaExtension.php
+++ b/src/TestSuite/FixtureSchemaExtension.php
@@ -62,7 +62,7 @@ class FixtureSchemaExtension implements
     /**
      * BeforeTestHook implementation
      *
-     * @param string $test The test being run.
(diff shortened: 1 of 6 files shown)


src/TestSuite/FixtureSchemaExtension.php
```diff
@@ -67,6 +79,6 @@ public function executeBeforeTest(string $test): void
      */
     public function executeAfterTest(string $test, float $time): void
     {
-        $this->state->afterTest();
+        $this->state->afterTest($test);
```
othercorey, line 82: fyi, this hook provides the full name of the class::method as $test.

markstory: Good point. I've updated the docstrings.

PR #15462, opened 2021-04-21
Show deprecation warning when provider is not an object or class name…

### LINKED ISSUE #14703: Passing data as rule providers via `Validator::setProvider()`?
This is a (multiple allowed):

* [x] bug
* [x] enhancement

* CakePHP Version: 3.x / 4.x

---

In the [**Using Custom Validation Rules**](https://book.cakephp.org/3/en/core-libraries/validation.html#using-custom-validation-rules) ([**4.x**](https://book.cakephp.org/4/en/core-libraries/validation.html#using-custom-validation-rules)) section of the docs, there's an example that shows using an array as a validation rule provider to pass additional data into the validator.

According to the API docs, that's not supported, as [**`Validator::setProvider()`**](https://github.com/cakephp/cakephp/blob/3.8.12/src/Validation/Validator.php#L251) expects `object|string`, and there's also no tests for that use case.

Either the API docs need to be updated, or the Cookbook... or both, detailing that providers don't necessarily have to be _validation rule_ providers. Or maybe there should be an additional feature specifically ment for exposing arbitrary data to the validation rules.

ADmad:
… string.

Closes #14703

### DIFF
--- a/src/Validation/Validator.php
+++ b/src/Validation/Validator.php
@@ -317,10 +317,18 @@ class Validator implements ArrayAccess, IteratorAggregate, Countable
      *
      * @param string $name The name under which the provider should be set.
      * @param object|string $object Provider object or class name.
+     * @psalm-param object|class-string $object
      * @return $this
      */
     public function setProvider(string $name, $object)
     {
+        if (!is_string($object) && !is_object($object)) {
+            deprecationWarning(sprintf(
+                'The provider must be an object or class name string. Got `%s` instead.',
+                getTypeName($object)
+            ));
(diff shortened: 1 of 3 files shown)


othercorey:
The problem matcher doesn't seem to be working.

markstory:
I like the changes. Test failures seem relevant.

ADmad:
This helped uncover an issue where the `EntityContext` would set `null` or related records `array` as provider. I didn't even know `EntityContext` sets the entity as a provider :slightly_smiling_face:.

PR #15467, opened 2021-04-27
Fix at matcher pt1

markstory:
I figured we should start the long process to update our usage of `at()` with undeprecated APIs. This is the first of many pull requests that will be needed to update all 300+ usages of `at()`.

### DIFF
--- a/tests/TestCase/Collection/Iterator/FilterIteratorTest.php
+++ b/tests/TestCase/Collection/Iterator/FilterIteratorTest.php
@@ -18,7 +18,6 @@ namespace Cake\Test\TestCase\Collection\Iterator;
 
 use Cake\Collection\Iterator\FilterIterator;
 use Cake\TestSuite\TestCase;
-use RuntimeException;
 
 /**
  * FilterIterator test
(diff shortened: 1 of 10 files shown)


othercorey:
Replacing at() in CollectionTest was all I had in me lol.

tests/TestCase/Collection/Iterator/FilterIteratorTest.php
```diff
@@ -32,21 +33,16 @@ class FilterIteratorTest extends TestCase
...
-            ->will($this->returnValue(false));
-        $callable->expects($this->at(1))
-            ->method('__invoke')
-            ->with(2, 1, $items)
-            ->will($this->returnValue(true));
-        $callable->expects($this->at(2))
-            ->method('__invoke')
-            ->with(3, 2, $items)
-            ->will($this->returnValue(false));
+        $callable = function ($key, $value, $itemArg) use ($items) {
```
othercorey, line 36: We could probably just simplify this to if $key === 2 return true otherwise return false. I don't know if we need to test internal values like this doing with the exceptions.

markstory: I can shorten the methods up. I had the arguments reversed for key/value, and have added more assertions as the keys/values are part of the interface we provide.

PR #15472, opened 2021-04-28
Update at() matchers in Console

markstory:
Update more tests to not use at(). Without at() we lose the ability to ensure method orders with mocks, as this is discouraged within PHPUnit mocks.

### DIFF
--- a/tests/TestCase/Console/CommandRunnerTest.php
+++ b/tests/TestCase/Console/CommandRunnerTest.php
@@ -513,18 +513,18 @@ class CommandRunnerTest extends TestCase
             ->setConstructorArgs([$this->config])
             ->getMock();
 
-        $app->expects($this->at(0))->method('bootstrap');
-        $app->expects($this->at(1))->method('pluginBootstrap');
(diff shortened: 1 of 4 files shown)


dereuromark:
> Without at() we lose the ability to ensure method orders with mocks

`withConsecutive` is the replacement when order matters, right?

markstory:
> withConsecutive is the replacement when order matters, right?

Yes but you can't use that to ensure order across different methods.

dereuromark:
Then I dont find their way of removing useful methods a good way forward.
If people want to use it, they can use at(), otherwise once() or any() etc.
Dictating the removal seems weird and counterproductive. Is there any good argument for it?

Sure, overusing it can create maintenance overhead or failing tests when they don't need to be.
But that's user error, and not the fault of the existence of those in general.

dereuromark:
I see some comments and details here: https://github.com/sebastianbergmann/phpunit/issues/4297

othercorey:
looks ok.

PR #15477, opened 2021-04-30
Allow specifying timezone to use when marshaling DateTimeType.

ADmad:
In apps using localized datetime, generally the values are converted to user's
timezone when used in datetime inputs of forms. When marshaling the submitted
values `DateTimeType` is unaware of the user's timezone and always marshals datetime
strings into `Time`/`FrozenTime` instances assuming they are in the system's timezone.

The new `DateTimeType::setLocalTimezone()` methods allows you to specify the
user's timezone to be used when marshaling the datetime strings.

### DIFF
--- a/src/Database/Type/DateTimeType.php
+++ b/src/Database/Type/DateTimeType.php
@@ -320,7 +320,8 @@ class DateTimeType extends BaseType implements BatchCastingInterface
     public function marshal($value): ?DateTimeInterface
     {
         if ($value instanceof DateTimeInterface) {
-            return $value;
+            /** @var \Datetime|\DateTimeImmutable $value */
+            return $value->setTimezone($this->defaultTimezone);
         }
 
         /** @var class-string<\DatetimeInterface> $class */
@@ -331,7 +332,10 @@ class DateTimeType extends BaseType implements BatchCastingInterface
             }
 
             if (ctype_digit($value)) {
-                return new $class('@' . $value);
+                /** @var \Datetime|\DateTimeImmutable $dateTime */
+                $dateTime = new $class('@' . $value);
+
+                return $dateTime->setTimezone($this->defaultTimezone);
             }
 
             if (is_string($value)) {
@@ -342,11 +346,7 @@ class DateTimeType extends BaseType implements BatchCastingInterface
                 }
 
                 /** @var \Datetime|\DateTimeImmutable $dateTime */
-                if (
-                    $dateTime !== null &&
-                    $this->userTimezone &&
-                    $this->userTimezone->getName() !== $this->defaultTimezone->getName()
-                ) {
+                if ($dateTime !== null) {
                     $dateTime = $dateTime->setTimezone($this->defaultTimezone);
                 }
 
@@ -388,7 +388,10 @@ class DateTimeType extends BaseType implements BatchCastingInterface
             $value['microsecond']
         );
 
-        return new $class($format, $value['timezone'] ?? $this->userTimezone);
+        /** @var \Datetime|\DateTimeImmutable $dateTime */
+        $dateTime = new $class($format, $value['timezone'] ?? $this->userTimezone);
+
+        return $dateTime->setTimezone($this->defaultTimezone);
     }
 
     /**
--- a/src/Database/Type/DateTimeType.php
+++ b/src/Database/Type/DateTimeType.php
@@ -320,6 +320,10 @@ class DateTimeType extends BaseType implements BatchCastingInterface
     public function marshal($value): ?DateTimeInterface
     {
         if ($value instanceof DateTimeInterface) {
+            if ($value instanceof DateTime) {
+                $value = clone $value;
+            }
+
             /** @var \Datetime|\DateTimeImmutable $value */
             return $value->setTimezone($this->defaultTimezone);
         }
--- a/src/Database/Type/DateTimeType.php
+++ b/src/Database/Type/DateTimeType.php
@@ -98,6 +98,13 @@ class DateTimeType extends BaseType implements BatchCastingInterface
     protected $dbTimezone;
 
     /**
+     * User time zone.
+     *
+     * @var \DateTimeZone|null
+     */
+    protected $userTimezone;
+
+    /**
      * Default time zone.
      *
      * @var \DateTimeZone
@@ -184,6 +191,24 @@ class DateTimeType extends BaseType implements BatchCastingInterface
             $timezone = new DateTimeZone($timezone);
         }
         $this->dbTimezone = $timezone;
+
+        return $this;
+    }
+
+    /**
+     * Set user timezone.
+     *
+     * This is the time zone used when marshalling strings to DateTime instances.
+     *
+     * @param string|\DateTimeZone|null $timezone User timezone.
+     * @return $this
+     */
+    public function setUserTimezone($timezone)
+    {
+        if (is_string($timezone)) {
+            $timezone = new DateTimeZone($timezone);
+        }
+        $this->userTimezone = $timezone;
 
         return $this;
     }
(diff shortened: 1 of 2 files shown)


ADmad:
This will avoid having to manually juggle the timezone for the datetime values as shown the "Modify add / edit actions to convert the submitted data to the current timezone of the submitting user" in this blog post https://toggen.com.au/it-tips/cakephp-4-time-zones/.

src/Database/Type/DateTimeType.php
```diff
@@ -304,13 +329,28 @@ public function marshal($value): ?DateTimeInterface
...
+                    $dateTime = $this->_parseValue($value);
+                }
+
+                /** @var \Datetime|\DateTimeImmutable $dateTime */
+                if (
+                    $dateTime !== null &&
+                    $this->localTimezone &&
+                    $this->localTimezone->getName() !== $this->defaultTimezone->getName()
+                ) {
+                    $dateTime = $dateTime->setTimezone($this->defaultTimezone);
```
ADmad, line 350: This is not strictly necessary but I think it would be nice to shift the timezone back to the system/default timezone for consistency.

ADmad: Actually it is necessary. Without this one will have to explicitly set `$dbTimezone` too to ensure the timezone if change before converting the value back to string when saving to database, even with PHP and db time are the same.

src/Database/Type/DateTimeType.php
```diff
@@ -188,6 +195,24 @@ public function setDatabaseTimezone($timezone)
...
 
+    /**
+     * Set local timezone.
+     *
+     * This is the time zone used when marshalling strings to DateTime instances.
+     *
+     * @param string|\DateTimeZone|null $timezone Database timezone.
+     * @return $this
+     */
+    public function setLocalTimezone($timezone)
```
othercorey, line 206: This might be a confusing name since it is only used for marshalling and not converting from db to php type.

I'm not sure it's the wrong name though.

src/Database/Type/DateTimeType.php
othercorey: I think we can just set this unconditionally without any cost. The cost is creating the DateTimeZone instance not setting it. I probably wrote code like this elsewhere, but not sure the comparison check is needed.

src/Database/Type/DateTimeType.php
markstory: Could this have 'user' in the name? That might help better indicate where/how it is meant to be used. I also wonder if we should have an example in the docs of how to build a middleware that uses this method to ensure requests always have their dates/times localized.

tests/TestCase/Database/Type/DateTimeTypeTest.php
```diff
@@ -308,6 +308,26 @@ public function testMarshal($value, $expected)
...
+    {
+        $this->type->setLocalTimezone('+0200');
+
+        $value = '2020-05-01 23:28:00';
+        $expected = new Time($value);
+
+        $result = $this->type->marshal($value);
+        $this->assertEquals($expected, $result->addHours(2));
+
+        $expected = new Time('2020-05-01 23:28:00', '+0200');
```
markstory, line 321: Shouldn't the output time of marshalling be in application's timezone?

src/Database/Type/DateTimeType.php
othercorey: I think user isn't distinct enough from server + local for many people including international users. We would have to be explicit like `setUserInterfaceTimezone()` or `setWebBrowserTimezone()` maybe.

ADmad: @markstory Yes, perhaps `setUserTimezone()`?
@othercorey I don't like the alternatives you have suggested :slightly_smiling_face:.

src/Database/Type/DateTimeType.php
ADmad: Okay, removing the check would make the code cleaner too.

tests/TestCase/Database/Type/DateTimeTypeTest.php
ADmad: Yup, I had forgotten about datetime array initially since it's no longer used in Cake 4. I'll update the related code and test the case.

src/Database/Type/DateTimeType.php
othercorey: `setUserTimezone()` is ok with me. It's documentable and probably covers more situations.

src/Database/Type/DateTimeType.php
```diff
@@ -188,6 +195,24 @@ public function setDatabaseTimezone($timezone)
         return $this;
     }
 
+    /**
+     * Set local timezone.
+     *
+     * This is the time zone used when marshalling strings to DateTime instances.
+     *
+     * @param string|\DateTimeZone|null $timezone Database timezone.
```
othercorey, line 203:
```suggestion
     * @param string|\DateTimeZone|null $timezone User time zone.
```

src/Database/Type/DateTimeType.php
```diff
@@ -97,6 +97,13 @@ class DateTimeType extends BaseType implements BatchCastingInterface
      */
     protected $dbTimezone;
 
+    /**
+     * Local time zone.
```
othercorey, line 101:
```suggestion
     * User time zone.
```

src/Database/Type/DateTimeType.php
```diff
@@ -188,6 +195,24 @@ public function setDatabaseTimezone($timezone)
...
 
+    /**
+     * Set local timezone.
+     *
+     * This is the time zone used when marshalling strings to DateTime instances.
+     *
+     * @param string|\DateTimeZone|null $timezone Database timezone.
+     * @return $this
+     */
+    public function setuserTimezone($timezone)
```
othercorey, line 206:
```suggestion
    public function setUserTimezone($timezone)
```

ADmad:
@markstory An example in the docs to use the feature in a middleware sounds great. Could you please take care of that?

src/Database/Type/DateTimeType.php
markstory: `setUserTimezone()` works for me.

src/Database/Type/DateTimeType.php
```diff
@@ -295,7 +320,8 @@ public function manyToPHP(array $values, array $fields, DriverInterface $driver)
     public function marshal($value): ?DateTimeInterface
     {
         if ($value instanceof DateTimeInterface) {
-            return $value;
+            /** @var \Datetime|\DateTimeImmutable $value */
+            return $value->setTimezone($this->defaultTimezone);
```
othercorey, line 324: Should we clone here for mutable instances? This is actually modifying the expected test result.

tests/TestCase/Database/Type/DateTimeTypeTest.php
```diff
@@ -308,6 +308,42 @@ public function testMarshal($value, $expected)
...
+     *
+     * @return void
+     */
+    public function testMarshalDateTimeInstance()
+    {
+        $expected = new Time('2020-05-01 23:28:00', 'Europe/Paris');
+
+        $result = $this->type->marshal($expected);
+        $this->assertEquals('UTC', $result->getTimezone()->getName());
+        $this->assertEquals($expected, $result->addHours(2));
```
othercorey, line 322: This is the same instance in `$expected` and `$result`, right? So, this is just comparison the same instance to itself.

src/Database/Type/DateTimeType.php
ADmad: Done

othercorey:
Looks ok to me.

PR #15480, opened 2021-05-02
4.x - Fix `Text file busy` errors when deleting directories.

ndm2:
I ran into this problem a few times over the years when working with VMs, where trying to delete directories would fail with an `rmdir(...): Text file busy` error, it came and go, and now I ran into it again, and it won't disappear.

[**Apparently**](https://github.com/thephpleague/flysystem/commit/d03f7e1e0f2fa47f52d445a60ec8ed93d433ddc1) unsetting the possible filesystem iterators helps releasing the locks, it certainly does so for me locally, no errors anymore using cache, folder and filesystem. I'm saying "_possible iterators_" here because these changes also unset variables that currently point to `SplFileInfo` objects, this won't do anything, it's just a preventive measure in case the code might change to use the "current as iterator" behavior.

Finally, tying into the preventive changes, I don't think I'd be able to write any tests for this, as I have no idea how to possibly force the problem.

### DIFF
--- a/src/Cache/Engine/FileEngine.php
+++ b/src/Cache/Engine/FileEngine.php
@@ -254,11 +254,13 @@ class FileEngine extends CacheEngine
         /** @var \SplFileInfo $fileInfo */
         foreach ($contents as $fileInfo) {
             if ($fileInfo->isFile()) {
+                unset($fileInfo);
                 continue;
             }
 
             $realPath = $fileInfo->getRealPath();
             if (!$realPath) {
+                unset($fileInfo);
                 continue;
             }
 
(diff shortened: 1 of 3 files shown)


markstory:
I agree there isn't a reasonable way to include tests for this.

othercorey:
> I agree there isn't a reasonable way to include tests for this.

Can we include a comment on one of them explaining why it's there? It's the kind of code that could be easily removed when cleaning up.

ndm2:
> Can we include a comment on one of them explaining why it's there?

Done in #15481

PR #15490, opened 2021-05-04
Explicitly check against null, to enable testing attributes of email #0

Zuluru:
The classes for testing email constraints do not work when specifying email #0 (the first email sent). We often send multiple emails to different recipients in one step, and need to ensure the specific contents of each, not just that *some* email contained a certain string. The code as written extracts the specified email if it is truthy, which 0 is not. Changing to specifically check for null resolves this.

I'm not sure what to do about unit testing of this, as there are currently no unit tests at all under TestSuite/Constraint/Email (that folder doesn't exist). If there were other unit tests there, I'd update them to test for this, but I don't have the time right now to build a whole set of test cases for this area of the code.

### DIFF
--- a/src/TestSuite/Constraint/Email/MailConstraintBase.php
+++ b/src/TestSuite/Constraint/Email/MailConstraintBase.php
@@ -47,7 +47,7 @@ abstract class MailConstraintBase extends Constraint
     {
         $emails = TestEmailTransport::getEmails();
 
-        if ($this->at) {
+        if ($this->at !== null) {
             if (!isset($emails[$this->at])) {
                 return [];
             }
--- a/tests/TestCase/TestSuite/EmailTraitTest.php
+++ b/tests/TestCase/TestSuite/EmailTraitTest.php
@@ -110,9 +110,6 @@ class EmailTraitTest extends TestCase
         $this->assertMailSentFromAt(0, 'default@example.com');
         $this->assertMailSentFromAt(1, 'alternate@example.com');
 
-        // Confirm that "at 0" is really testing email 0, not all the emails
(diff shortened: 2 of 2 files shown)


markstory:
> I'm not sure what to do about unit testing of this, as there are currently no unit tests at all under TestSuite/Constraint/Email (that folder doesn't exist). 

The tests for email constraints are [here](https://github.com/cakephp/cakephp/blob/master/tests/TestCase/TestSuite/EmailTraitTest.php).

Zuluru:
Okay! I'll get something added there shortly to test this.

Zuluru:
Added a test, but I don't love it. It's referencing things outside of what this file is supposed to be testing. But there's no "assertMailNotSentToAt" set of negative assertion functions, nor do I think there should be. Open to suggestions about how to improve this.

Zuluru:
...and the test fails on 5.6 because LogicalNot isn't available there?

markstory:
> Added a test, but I don't love it. It's referencing things outside of what this file is supposed to be testing. But there's no "assertMailNotSentToAt" set of negative assertion functions, nor do I think there should be. Open to suggestions about how to improve this.

Could you have a separate test that only sends two emails with different addresses and check specifics from the first email with `assertMailSentWithAt(0, ...)`

Zuluru:
The problem is that the original code would succeed with that: when "0" is specified for the email to check, it will assert true if *any* of the emails that were sent match. So such a test would not actually be testing anything useful. The only way to test the fix is to specifically have a negative test.

markstory:
Ok. You could have the current test you've added in a separate method and skip it if the `LogicalNot` class doesn't exist.

Issue #15497, opened 2021-05-05
Query encoding switch is missing in the http Client


RobRikken:
* [ ] bug
* [x] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: CakePHP version 4.1
* Platform and Target: Windows, request to Odata API

### What you did
Did a request with the http Client of CakePHP to a Odata API server.

### What happened
The request produced an error from the API, because the API doesn't support a + in the query part of the url.

### What you expected to happen
That the CakePHP http Client encoded the space character into %20 so the request succeeds.

### Notes
The http_build_query() function that is used has a option to change the encoding to a %20 instead of a + while encoding, but this is not an option in the CakePHP http client.

The php function:
```
* @param int $encoding_type By default, PHP_QUERY_RFC1738.
 *  <p>If enc_type is PHP_QUERY_RFC1738, then encoding is performed per » RFC 1738 and the application/x-www-form-urlencoded media type,
 *  which implies that spaces are encoded as plus (+) signs.
 *  <p>If enc_type is PHP_QUERY_RFC3986, then encoding is performed according to » RFC 3986, and spaces will be percent encoded (%20).
 * @return string a URL-encoded string.
 */
#[Pure]
function http_build_query ($data, string $numeric_prefix = "", string $arg_separator = "&", int $encoding_type = PHP_QUERY_RFC1738){}
```

CakePHP not using the encoding option:
```
public function buildUrl(string $url, $query = [], array $options = []): string
    {
        if (empty($options) && empty($query)) {
            return $url;
        }
        if ($query) {
            $q = strpos($url, '?') === false ? '?' : '&';
            $url .= $q;
            $url .= is_string($query) ? $query : http_build_query($query);
        }
```

ADmad:
Fixed in  #15498

PR #15498, opened 2021-05-05
Switch http client query encoding

RobRikken:
Fix for the issue #15497.

This consists of the addition of the encoding query option and the tests to make sure it works.

### DIFF
--- a/src/Http/Client.php
+++ b/src/Http/Client.php
@@ -122,6 +122,7 @@ class Client implements ClientInterface
         'ssl_verify_host' => true,
         'redirect' => false,
         'protocolVersion' => '1.1',
+        'query_encoding' => PHP_QUERY_RFC1738,
     ];
 
     /**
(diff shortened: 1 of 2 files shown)


src/Http/Client.php
```diff
@@ -519,20 +520,26 @@ public function buildUrl(string $url, $query = [], array $options = []): string
...
-            $url .= $q;
-            $url .= is_string($query) ? $query : http_build_query($query);
-        }
         $defaults = [
             'host' => null,
             'port' => null,
             'scheme' => 'http',
             'basePath' => '',
             'protocolRelative' => false,
+            'query_encoding' => PHP_QUERY_RFC1738,
```
markstory, line 529: Could this be `queryEncoding`? Do we need an option at all. Seems like we could/should default to 3986 given that 1738 is updated by 3986.

markstory (commented):
Thanks for sending this pull request. I'm not sure we need to have strict compatibility here. Servers should generally accept percent encoding.

src/Http/Client.php
RobRikken: No problem, I've changed the name to  queryEncoding.

RobRikken:
The default is now changed to PHP_QUERY_RFC3986, this does however give me the following error in the test suite:

```
3) Cake\Test\TestCase\Http\ClientTest::testHeadQuerystring
Expectation failed for method name is "send" when invoked 1 time(s)
Failed asserting that two strings are identical.
--- Expected
+++ Actual
@@ @@
-'http://cakephp.org/search?q=hi+there'
+'http://cakephp.org/search?q=hi%20there'
```

I however do agree that %20 should be generally accepted.

ADmad:
I too don't think the `queryEncoding` option is necessary, we should simply switch to PHP_QUERY_RFC3986 and update existing test cases.

RobRikken:
I have set PHP_QUERY_RFC3986 as the only option and updated the testcases to now use the %20 instead of +.

othercorey:
Because this is in 4.3, we should have an entry in the migration notes.

PR #15502, opened 2021-05-06
Improve error message.

ADmad:

### DIFF
--- a/src/View/Helper/FormHelper.php
+++ b/src/View/Helper/FormHelper.php
@@ -679,7 +679,8 @@ class FormHelper extends Helper
     {
         if ($this->formProtector === null) {
             throw new CakeException(
-                'FormHelper::create() must be called first for FormProtector instance to be created.'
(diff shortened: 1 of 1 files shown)


src/View/Helper/FormHelper.php
```diff
@@ -679,7 +679,8 @@ public function getFormProtector(): FormProtector
     {
         if ($this->formProtector === null) {
             throw new CakeException(
-                'FormHelper::create() must be called first for FormProtector instance to be created.'
+                '`FormProtector` instance has not been created. Ensure you have loaded the `FormProtectorComponent`'
```
davidyell, line 682: FormProtectionComponent

dereuromark: Right - https://github.com/cakephp/cakephp/blob/master/src/Controller/Component/FormProtectionComponent.php

ADmad: https://github.com/cakephp/cakephp/commit/fe6fd198e33f69c1d36d5df4ac2c92d26ffa4807

PR #15504, opened 2021-05-08
Don't set direction to false

### LINKED ISSUE #15499: Paginator TypeError on indirect controller pagination
This is a (multiple allowed):

* [x] bug
* [ ] enhancement

* CakePHP Version: latest master

### What you did
Upgrade cake3 app to cake4

URL `/clubs/users` without any args.
```php
	public $paginate = ['order' => ['Clubs.name' => 'ASC']];

	// ClubController users action
	$query = $this->Clubs->ClubUsers->find();
	$clubUsers = $this->paginate($query)->toArray();
```

### What happened

```
[TypeError] strtolower() expects parameter 1 to be string, bool given in
 /var/www/site/vendor/cakephp/cakephp/src/View/Helper/PaginatorHelper.php on line 566
```

I solved it on project level for now using
```php
		// HACK FOR NOW
		if ($this->request->getAttribute('paging') && $this->request->getAttribute('paging')['ClubUsers']['direction'] === false) {
			$paging = $this->request->getAttribute('paging');
			$paging['ClubUsers']['direction'] = 'asc';
			$this->request = $this->request->withAttribute('paging', $paging);
		}
```

ServerRequest content
```
protected attributes => [
'isAjax' => false,
'paging' => [
'ClubUsers' => [
'count' => (int) 614,
'current' => (int) 20,
'perPage' => (int) 20,
'page' => (int) 1,
'requestedPage' => (int) 1,
'pageCount' => (int) 31,
'start' => (int) 1,
'end' => (int) 20,
'prevPage' => false,
'nextPage' => true,
'sort' => 'Clubs.name',
'direction' => false, // !
'sortDefault' => 'Clubs.name',
'directionDefault' => 'ASC',
'completeSort' => [ ],
'limit' => null,
'scope' => null,
'finder' => 'all',
],
],
]
```

No TypeError, either meaningful exception to dev, or defaulting to a sane default for direction (asc/desc etc).

I assume there is some shenanigans going on inside the default whitelist validation for aliases models.
Which then, combined with a non-direct component paginate seems to not work with the model being used for sorting.

markstory:
When the order data is empty because it is invalid, direction should not be set to `false`. Instead `null` is better as it doesn't cause downstream problems with PaginatorHelper

Fixes #15499

### DIFF
--- a/tests/TestCase/Datasource/PaginatorTest.php
+++ b/tests/TestCase/Datasource/PaginatorTest.php
@@ -165,4 +165,27 @@ class PaginatorTest extends TestCase
         $this->assertSame(2, $pagingParams['PaginatorPosts']['end']);
         $this->assertFalse($pagingParams['PaginatorPosts']['nextPage']);
     }
+
+    /**
+     * test direction setting.
+     *
+     * @return void
+     */
+    public function testPaginateDirection()
+    {
+        $settings = [
+            'PaginatorPosts' => [
+                'order' => ['Other.title' => 'ASC'],
+            ],
+        ];
+
+        $this->loadFixtures('Posts');
+        $table = $this->getTableLocator()->get('PaginatorPosts');
+
+        $this->Paginator->paginate($table, [], $settings);
+        $pagingParams = $this->Paginator->getPagingParams();
+
+        $this->assertSame('Other.title', $pagingParams['PaginatorPosts']['sort']);
+        $this->assertNull($pagingParams['PaginatorPosts']['direction']);
+    }
 }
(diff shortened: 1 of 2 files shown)


tests/TestCase/Datasource/PaginatorTest.php
```diff
@@ -165,4 +165,27 @@ public function testPaginateCustomFinder()
         $this->assertSame(2, $pagingParams['PaginatorPosts']['end']);
         $this->assertFalse($pagingParams['PaginatorPosts']['nextPage']);
     }
+
+    /**
+     * test direction setting.
+     *
+     * @return void
+     */
+    public function testPaginateDirection()
```
othercorey, line 174:
```suggestion
    public function testPaginateDefaultDirection()
```

PR #15505, opened 2021-05-08
Fixed modifyApplication() not able to find boostrap()-method in Appli…

fabian-mcfly:
Fixed `PluginLoadCommand::modifyApplication()` not able to find `Application::boostrap()`
When loading a plugin with `bin/cake plugin load <pluginname>`, modifyApplication() wasn't able to find the bootstrap()-method when not following the CakePHP code style convention


markstory:
Thanks!

PR #15516, opened 2021-05-15
Missing clossing braces in DocBlock examples

hacheraw:
I have tried to respect the spacing by matching it to the rest of the examples within each DocBlock, but there are style differences between one block and another.


othercorey:
Thanks!

PR #15517, opened 2021-05-17
4.next - Use new fixtures for tests

markstory:
Extract schema from the existing core fixtures into a new schema file that can be used with data-only fixtures. I had to make a few API changes to the new fixtures as I hadn't wired up row loading yet.

I don't think we can safely remove schema from any of the core fixtures as they are likely referenced in plugin builds and those plugins will lag behind in adopting new fixtures.

Part of #15308

### DIFF
--- a/src/TestSuite/Fixture/FixtureLoader.php
+++ b/src/TestSuite/Fixture/FixtureLoader.php
@@ -79,12 +79,14 @@ abstract class FixtureLoader
     abstract public function load(TestCase $test): void;
 
     /**
-     * Inspects the test to to load fixture classes.
+     * Setup fixtures for the provided test.
      *
-     * @param \Cake\TestSuite\TestCase $test The test case to inspect.
+     * Called by TestCase during its setUp() method.
+     *
+     * @param \Cake\TestSuite\TestCase $test The test to inspect for fixture loading.
      * @return void
      */
-    abstract public function fixturize(TestCase $test): void;
+    abstract public function setupTest(TestCase $test): void;
 
     /**
      * Get the fixtures fixtures.
--- a/src/TestSuite/Fixture/FixtureLoader.php
+++ b/src/TestSuite/Fixture/FixtureLoader.php
@@ -89,9 +89,16 @@ abstract class FixtureLoader
     abstract public function setupTest(TestCase $test): void;
 
     /**
-     * Get the fixtures fixtures.
+     * Get the list of all fixtures that have been loaded.
      *
      * @return \Cake\Datasource\FixtureInterface[]
      */
     abstract public function loaded(): array;
+
+    /**
+     * Get the list of fixture tables that were loaded since the last call to setupTest.
+     *
+     * @return string[]
+     */
+    abstract public function lastInserted(): array;
 }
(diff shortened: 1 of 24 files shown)


othercorey:
Is the schema.php array in any specific order?

markstory:
> Is the schema.php array in any specific order?

Yes and no. The tables are generally not ordered except where foreign keys are involved.

> Is it possible to load from multiple files or do teams have to deal with maintaining a single large array?

Ideally no users ever have to interact with this format. This format is only intended to be used for core tests, as we can't require phinx or migrations due to cyclic dependencies on cake core.

Instead developers would run migrations for their application in their `tests/bootstrap.php` The start of that work is in cakephp/migrations#488. Alternative forms for loading schema include loading one or more files with the required SQL to create the schema, or users can roll their own schema generation if they have externally managed SQL via another migration tool.

src/TestSuite/Fixture/TruncationStrategy.php
```diff
@@ -96,14 +98,13 @@ public function beforeTest(string $test): void
             if (!($db instanceof Connection)) {
                 continue;
             }
+            $schema = $db->getSchemaCollection();
+            $tables = $schema->listTables();
+            $tables = array_intersect($schema->listTables(), $fixtures->lastInserted());
```
markstory, line 103: I'm not in love with the `lastInserted()` approach and would appreciate any ideas on a cleaner solution.

The problem I was aiming to solve was that truncating the 54 fixture tables we have was really slow. The current fixture system skips some of this work by only truncating tables that were used in the current test.

othercorey:
> Ideally no users ever have to interact with this format. This format is only intended to be used for core tests, as we can't require phinx or migrations due to cyclic dependencies on cake core.

Should we enforce that only core uses it?

markstory:
> Should we enforce that only core uses it?

It is marked as `@internal` and won't be included in the docs.

src/TestSuite/Fixture/TruncationStrategy.php
othercorey: > The current fixture system skips some of this work by only truncating tables that were used in the current test.

Our internal fixtures don't have foreign keys setup most of the time. Do we want to support fixtures with foreign keys and truncating subsets of the tables?

src/TestSuite/Fixture/FixtureLoader.php
```diff
@@ -79,17 +79,26 @@ abstract public function loadSingle(
...
      * @return \Cake\Datasource\FixtureInterface[]
      */
     abstract public function loaded(): array;
+
+    /**
+     * Get the list of fixture tables that were loaded since the last call to setupTest.
+     *
+     * @return string[]
+     */
+    abstract public function lastInserted(): array;
```
othercorey, line 103: This name doesn't make much sense. Fixtures aren't "inserted" and "last" doesn't really mean anything either.

I know this was a hack/temp solution but maybe clarifying this will help.

markstory: I agree the name is clunky. I am planning to make this method protected when I move state management. I didn't want to increase the scope of this change too much as I am still battling with the various database differences.

markstory:
Merging this so I can continue work on the next block of changes which is to make state reset part of the TestCase.

PR #15518, opened 2021-05-18
Always clone NumberFormatter to avoid php 8 issue when cloning

### LINKED ISSUE #15501: Number helper does not initialize NumberFormatter on PHP 8.0 correctly
This is a (multiple allowed):

* [x] bug
* [ ] enhancement

* CakePHP Version: 4.2.5
* Platform and Target: PHP 8.0.5

### What you did

While upgrading our application to PHP 8.0, we noticed a strange bug when interacting with `NumberFormatter` through `Cake\I18n\Number` that did not happen in PHP 7.4:

```php
<?php

require './vendor/autoload.php';

use Cake\I18n\Number;
var_dump(Number::parseFloat('invalid'));
var_dump(Number::precision(123.45, 2));
```

### What happened

On PHP 8.0 above code fails with message: `PHP Fatal error:  Uncaught Exception: Failed to clone NumberFormatter object in /www-root/vendor/cakephp/cakephp/src/I18n/Number.php:413`

On PHP 7.4 this works without throwing an exception.

There seems to be either a bug or behavior change in PHP 8 which occurs when cloning a NumberFormatter that is in an error-state. See vanilla example here: https://3v4l.org/CLcph

othercorey:
Closes https://github.com/cakephp/cakephp/issues/15501

### DIFF
--- a/src/I18n/Number.php
+++ b/src/I18n/Number.php
@@ -400,14 +400,17 @@ class Number
         /** @var \NumberFormatter $formatter */
         $formatter = static::$_formatters[$locale][$type];
 
-        $options = array_intersect_key($options, [
-            'places' => null,
-            'precision' => null,
-            'pattern' => null,
-            'useIntlCode' => null,
-        ]);
-        if (empty($options)) {
-            return $formatter;
+        // PHP 8.0.0 - 8.0.6 throws an exception when cloning NumberFormatter after a failed parse
+        if (version_compare(PHP_VERSION, '8.0.6', '>') || version_compare(PHP_VERSION, '8.0.0', '<')) {
+            $options = array_intersect_key($options, [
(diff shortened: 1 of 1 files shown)


ADmad:
Given that the issue is fixed in PHP can't we keep the optimization of avoiding the cloning of the formatter instance based on version check?

othercorey:
> Given that the issue is fixed in PHP can't we keep the optimization of avoiding the cloning of the formatter instance based on version check?

Up to you. It won't be fixed until 8.0.7. Anyone using this before then will get a pretty confusing issue.

othercorey:
@ADmad I misread your comment. Added version checks around optimization. I'm curious how much the clone costs compared to the tests, but worse case is they negate each other for now.

PR #15520, opened 2021-05-19
Update psalm baseline with laminas psalm-var changes

othercorey:


othercorey:
There is a psalm issues that makes these types difficult to satisfy. https://github.com/vimeo/psalm/issues/5795

Issue #15525, opened 2021-05-20
Static analysis fails on QueryExpression::not called with a closure as first parameter


liqueurdetoile:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 4.2.6
* Platform and Target: Not relevant

### What you did
I'm using `QueryExpression::not` with a closure as first parameter.

### What happened
Results as are valid but I ran into an error when performing static analysis :

```
Parameter #1 $conditions of method Cake\Database\Expression\QueryExpression::not() expects array|Cake\Database\ExpressionInterface|string, Closure(mixed): mixed given.
```

A closure is not expected as a valid type for this parameter, unlike `QueryExpression::and` and `QueryExpression::or`.

### What you expected to happen
Using a closure with `QueryExpression::not` is leading to expected results in my case.  Peeping at the code, the final call to `QueryExpression::_addConditions` seems to process closure the same way that `and` and `or` are doing. Therefore, I don't think I'm going really wrong with this.

Should doc block be updated to include `\Closure` or did I miss something and should I forget to use a closure in that case ?

Thanks

Issue #15524, opened 2021-05-20
Error on description for `saveMany()`?


mirko-pagliai:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

Maybe it's a bug, see https://github.com/phpstan/phpstan/issues/5042.

In fact, if I give an array of `Cake\Datasource\EntityInterface`, I get this error from `phpstan`:
```
  75     Parameter #1 $entities of method Cake\ORM\Table::saveMany() expects Cake\Datasource\ResultSetInterface&iterable<Cake\Datasource\EntityInterface>,  
         array<Cake\Datasource\EntityInterface> given.   
```

The description of the `$entities` parameter should be this?

```php
     * @param array<\Cake\Datasource\EntityInterface>|\Cake\Datasource\ResultSetInterface $entities Entities to save.
```

othercorey:
What cakephp version are you using?

mirko-pagliai:
> What cakephp version are you using?

I'm sorry, I think it was obvious: the last one, 4.2.6.
But it doesn't seem to me that the source has been changed lately.

othercorey:
I don't understand the phpstan logic. It seems like a bug in their phpdoc parsing.

dereuromark:
Could it be that the | should be & here and all would work fine?

mirko-pagliai:
The problem is everywhere though.

Currently it does not seem visible because `phpstan` does not parse `test/TestCase` files - in my opinion wrong, but understandable, I guess so to avoid having to handle too many things, including many false positives.

Another example (just doing the parsing of `tests/TestCase/ORM/TableTest.php`):
```
  5867   Parameter #1 $entities of method Cake\ORM\Table::patchEntities() expects iterable<Cake\Datasource\EntityInterface>&Traversable, array<int,           
         Cake\ORM\Entity> given. 
```



> I don't understand the phpstan logic. It seems like a bug in their phpdoc parsing.

@othercorey I also suspect it's a bug of the phpdoc parsing. But they say no (see link I posted).

ADmad:
So have to add stub files to fix phpstan's incorrect type inference :unamused:.

mirko-pagliai:
@AD7six I did a test (on 4.2.6 and now on master) and I understand that both of these syntaxes are valid:

```
     * @param \Cake\Datasource\EntityInterface[]&\Cake\Datasource\ResultSetInterface $entities Entities to delete.
```
```
[...]

PR #15526, opened 2021-05-20
Update types in docblocks.

ADmad:
Refs #15524, #15525.

### DIFF
--- a/src/Filesystem/Filesystem.php
+++ b/src/Filesystem/Filesystem.php
@@ -204,7 +204,7 @@ class Filesystem
 
         $result = true;
         foreach ($iterator as $fileInfo) {
-            $isWindowsLink = DS === '\\' && $fileInfo->getType() === 'link';
+            $isWindowsLink = DIRECTORY_SEPARATOR === '\\' && $fileInfo->getType() === 'link';
             if ($fileInfo->getType() === self::TYPE_DIR || $isWindowsLink) {
                 // phpcs:ignore
                 $result = $result && @rmdir($fileInfo->getPathname());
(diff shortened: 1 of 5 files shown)


dereuromark (approved):

src/ORM/Table.php
```diff
@@ -2174,9 +2174,9 @@ protected function _update(EntityInterface $entity, array $data)
      * any one of the records fails to save due to failed validation or database
      * error.
      *
-     * @param \Cake\Datasource\EntityInterface[]|\Cake\Datasource\ResultSetInterface $entities Entities to save.
+     * @param array<\Cake\Datasource\EntityInterface>|\Cake\Datasource\ResultSetInterface $entities Entities to save.
```
othercorey, line 2177: This feels like a slippery slope.

ADmad: This is a better way, we couldn't use it earlier because IDEs hadn't caught up with this syntax.

src/Filesystem/Filesystem.php
```diff
@@ -204,7 +204,7 @@ public function deleteDir(string $path): bool
 
         $result = true;
         foreach ($iterator as $fileInfo) {
-            $isWindowsLink = DS === '\\' && $fileInfo->getType() === 'link';
+            $isWindowsLink = DIRECTORY_SEPARATOR === '\\' && $fileInfo->getType() === 'link';
```
othercorey, line 207: Why does this need to use the full name?

src/ORM/Table.php
othercorey: It's fine with me but a lot of hints to change if phpstan keeps at it.

I could write a phpcs transform if we need to update more.

othercorey:
Just trying to get a green GA build here. I restarted the workflow.

src/Filesystem/Filesystem.php
ADmad: phpstan showed an error which I couldn't understand for this line. So tried the 1st thing that came to mind and it seemed to work :). Using PHP's constant is preferable either way.

PR #15533, opened 2021-05-21
Fix OAuth 1 when consumer key is base64 string

Rytiss:
Fix consumer keys not being correctly encoded when the key is base64 and ends with an equals sign

Consider a consumer key like `...O0jGs=`. When creating an OAuth 1 base string, CakePHP encodes this part as follows:
`oauth_consumer_key%3D...0jGs%3D`

However, the correct encoding is as follows:
`oauth_consumer_key%3D...0jGs%253D`

Note how CakePHP does not encode the % sign, thus creating wrong base string and thus the signature is also incorrect.

Tested on 3.7.9, but seemingly affects other versions too.

### DIFF
--- a/src/Http/Client/Auth/Oauth.php
+++ b/src/Http/Client/Auth/Oauth.php
@@ -145,6 +145,10 @@ class Oauth
         ];
         $baseString = $this->baseString($request, $values);
 
+        // Consumer key should only be encoded for base string calculation as
+        // auth header generation already encodes independently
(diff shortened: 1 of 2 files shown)


ADmad:
> Tested on 3.7.9

The `master` branch is for Cake 4.x, but no problem we can backport the fix.

ADmad:
Please add a test for the fix.

Rytiss:
Added a test case and fixed a problem with double encoding

PR #15535, opened 2021-05-21
it makes the `phpstan` rules compatible with PHP 8

mirko-pagliai:
PHP 8.0 reports some errors with slightly different syntax, as well as new errors not reported by PHP 7.x.

We need to disable the `reportUnmatchedIgnoredErrors` parameter just for backward compatibility with PHP 7.x.

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -121,7 +121,7 @@ parameters:
 			path: src/Collection/Iterator/ZipIterator.php
 
 		-
-			message: "#^Parameter \\#1 \\$var_array of function extract is passed by reference, so it expects variables only\\.$#"
+			message: "#^Parameter \\#1 \\$(var_)?array of function extract is passed by reference, so it expects variables only\\.$#"
 			count: 1
 			path: src/Command/I18nExtractCommand.php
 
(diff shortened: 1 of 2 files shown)


othercorey:
What happens when the baseline is updated which removes unmatched errors?

ADmad:
Currently I used phpstan's `--generate-baseline` option to update the `phpstan-baseline.neon` file whenever needed. These changes mean the file can no longer be auto generated. Manual updating the file is not something I am looking forward to.

mirko-pagliai:
> Currently I used phpstan's `--generate-baseline` option to update the `phpstan-baseline.neon` file whenever needed. These changes mean the file can no longer be auto generated. Manually updating the file is not something I am looking forward to.

Okay, I guess it is enough that I remove the regexes.

> What happens when the baseline is updated which removes unmatched errors?

You're right, I need to check.

othercorey:
Not running composer stan using php 8 (my default) is annoying, but not sure how to make it dual compatible.

dereuromark:
Running it only once and for one version should suffice and is most likely recommended in terms of baseline and consistent reporting.

mirko-pagliai:
> Running it only once and for one version should suffice and is most likely recommended in terms of baseline and consistent reporting.

So can we say it's better to leave it as it is now?

ADmad:
> So can we say it's better to leave it as it is now?

Yes, we will consider switching to PHP 8 in the future for running phpstan.

PR #15538, opened 2021-05-21
Add __serialize()/__unserialize() magic methods.

### LINKED ISSUE #15530: PHP 8.1: Serializable is deprecated
With PHP 8.1, Serializable is deprecated in favor of `__serialize()` and `__unserialize()`.

> PHP Deprecated:  The Serializable interface is deprecated. Implement __serialize() and __unserialize() instead (or in addition, if support for old PHP versions is necessary) in /home/runner/work/cakephp/cakephp/src/Collection/Collection.php on line 27

The deprecation is suppressed if both mechanisms are implemented for libraries that support older PHP versions. We should be able to do this.

ADmad:
The Serializable interface is deprecated in PHP 8.1 and adding these methods
avoids deprecation errors.

### DIFF
--- a/src/Collection/Iterator/BufferedIterator.php
+++ b/src/Collection/Iterator/BufferedIterator.php
@@ -202,6 +202,20 @@ class BufferedIterator extends Collection implements Countable, Serializable
     }
 
     /**
+     * Magic method used for serializing the iterator instance.
+     *
+     * @return array
+     */
+    public function __serialize(): array
+    {
+        if (!$this->_finished) {
+            $this->count();
+        }
+
+        return iterator_to_array($this->_buffer);
+    }
+
+    /**
      * Unserializes the passed string and rebuilds the BufferedIterator instance
      *
      * @param string $buffer The serialized buffer iterator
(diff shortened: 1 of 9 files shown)


src/Collection/Iterator/BufferedIterator.php
```diff
@@ -214,4 +228,20 @@ public function unserialize($buffer): void
...
+     * Magic method used to rebuild the iterator instance.
+     *
+     * @param array $data Data array.
+     * @return void
+     */
+    public function __unserialize(array $data): void
+    {
+        $this->__construct([]);
+        foreach ($data as $value) {
+            $this->_buffer->push($value);
```
ADmad, line 242: I am not 100% sure about this. There are no tests for serialization/unserialization in `BuffereredIteratorTest.php`.

othercorey: Does anything serialize it as a side-effect?

ADmad: Codecov said there was no coverage for these lines, so I doubt it. Anyway, I added a test.

src/Log/Engine/BaseLog.php
```diff
@@ -140,13 +140,18 @@ protected function _format(string $message, array $context = []): string
...
-                if (method_exists($value, '__toString')) {
-                    $replacements['{' . $key . '}'] = (string)$value;
+                if (method_exists($value, 'toArray')) {
+                    $replacements['{' . $key . '}'] = json_encode($value->toArray(), JSON_UNESCAPED_UNICODE);
                     continue;
                 }
 
-                if (method_exists($value, 'toArray')) {
-                    $replacements['{' . $key . '}'] = json_encode($value->toArray(), JSON_UNESCAPED_UNICODE);
+                if (method_exists($value, '__serialize')) {
```
othercorey, line 148: This should be ok since we're implementing both, but __serialize() won't be called for php 7.4.

ADmad: Right, but those classes already implement `Serializable` which is checked earlier. This check for `__serializable` is for future proofing when the check for Serializable will be removed and for app code which only use the magic methods and not implement Serializable.

othercorey: Ah, I see it. The order change confused me.

othercorey:
Looks good to me.

PR #15540, opened 2021-05-21
Fix TypeError on PHP 8.1 when strlen() is called with null.

### LINKED ISSUE #15539: PHP 8.1: Passing null to strlen is deprecated
Previously, passing null to strlen would just return 0. However, since the parameter type is actually `string`, this is now deprecated.

```
strlen(): Passing null to parameter #1 ($string) of type string is deprecated
```

One common scenario that triggers this deprecation is using the primary key from an entity. `EntityTrait::get()` currently returns null for a missing field. Since the primary key is an array type, this creates an array of null values.

Example from `BelongsToMany` association.

```
        $bindingKey = (array)$this->getBindingKey();
        $primaryValue = $sourceEntity->extract($bindingKey);

        if (count(array_filter($primaryValue, 'strlen')) !== count($bindingKey)) {
            $message = 'Could not find primary key value for source entity';
            throw new InvalidArgumentException($message);
        }
```

Another example from `IsUnique` vaildation rule.

```
        if ($entity->isNew() === false) {
            $keys = (array)$options['repository']->getPrimaryKey();
            $keys = $this->_alias($alias, $entity->extract($keys));
            if (array_filter($keys, 'strlen')) {
                $conditions['NOT'] = $keys;
            }
        }
```

ADmad:
Fixes #15539.

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -392,11 +392,6 @@ parameters:
 			path: src/Mailer/Email.php
 
 		-
-			message: "#^Parameter \\#2 \\$callback of function array_filter expects callable\\(mixed, mixed\\)\\: bool, 'strlen' given\\.$#"
-			count: 1
-			path: src/ORM/Association/BelongsToMany.php
-
-		-
 			message: "#^Unsafe usage of new static\\(\\)\\.$#"
 			count: 2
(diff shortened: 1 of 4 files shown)


othercorey:
Looks good to me if we just want non-null. It does let `true` through, but that's never a primary key.

PR #15541, opened 2021-05-22
Access PDOStatement::$queryString only after execution

### LINKED ISSUE #15537: Cannot access PDOStatement::$queryString before initialized
In PHP 8.1, accessing PDOStatement::$queryString before it's initialized with a value throws a type error:

```
Error: Typed property PDOStatement::$queryString must not be accessed before initialization
```

This is caused by code like this in `StatementDecorator`:
```
    public function __get(string $property)
    {
            return $this->_statement->queryString;
        }
    }
```

And in `SqliteStatement` leading to that call:
```
    public function rowCount(): int
    {
        if (
            $this->_statement->queryString &&
            preg_match('/^(?:DELETE|UPDATE|INSERT)/i', $this->_statement->queryString)
        ) {
```

This code seems to expect a `null` value returned when uninitialized. We'll need to only check this when the statement has executed.

othercorey:
Fixes https://github.com/cakephp/cakephp/issues/15537

### DIFF
--- a/src/Database/Statement/MysqlStatement.php
+++ b/src/Database/Statement/MysqlStatement.php
@@ -32,6 +32,7 @@ class MysqlStatement extends PDOStatement
      */
     public function execute(?array $params = null): bool
     {
+        $this->_hasExecuted = true;
         $connection = $this->_driver->getConnection();
 
         try {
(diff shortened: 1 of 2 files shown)


src/Database/Statement/MysqlStatement.php
```diff
@@ -32,6 +32,7 @@ class MysqlStatement extends PDOStatement
      */
     public function execute(?array $params = null): bool
     {
+        $this->_hasExecuted = true;
```
othercorey, line 35: I tried extending the unit tests for this statement, but these recursively nested statements makes it very difficult.

PR #15542, opened 2021-05-22
Backport 15533

markstory:
Backport the changes from #15533 to 3.x

### DIFF
--- a/tests/TestCase/Http/Client/Auth/OauthTest.php
+++ b/tests/TestCase/Http/Client/Auth/OauthTest.php
@@ -332,7 +332,7 @@ shqoyFXJvizZzje7HaTQv/eJTuA6rUOzu/sAv/eBx2YAPkA8oa3qUw==
 
         $result = $request->getHeaderLine('Authorization');
         $expected = '2hr/eoFyTSuWc6SfZIvkhpeRHdM=';
-        $this->assertStringContainsString(
+        $this->assertContains(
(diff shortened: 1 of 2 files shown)


Rytiss:
Lacks the critical part of the original change that runs `_encode()` on the consumer key: https://github.com/cakephp/cakephp/pull/15533/files#diff-14e47e708381dc7e32406d133527a97f94d5cad603de3dcca07e7236a2335affR144

markstory:
Interesting I wonder why that chunk didn't get applied when the commit was picked across.

ADmad:
And the test case is added twice :)

markstory:
Should be better now :blush:

tests/TestCase/Http/Client/Auth/OauthTest.php
```diff
@@ -307,6 +307,37 @@ public function testHmacSigning()
...
+            'token' => 'nnch734d00sl2jdk',
+            'nonce' => 'kllo9940pd9333jh',
+            'timestamp' => '1191242096',
+        ];
+        $auth = new Oauth();
+        $request = $auth->authentication($request, $options);
+
+        $result = $request->getHeaderLine('Authorization');
+        $expected = '2hr/eoFyTSuWc6SfZIvkhpeRHdM=';
+        $this->assertStringContainsString(
```
markstory, line 335:
```suggestion
        $this->assertContains(
```

Issue #15393, opened 2021-03-12
Add JSON format output to the ConsoleLog


bor-attila:
This is a:

* [ ] bug
* [x] enhancement
* [x] feature-discussion (RFC)

* CakePHP Version: All
* Platform and Target: Kubernetes (Google Cloud Engine), but I think all container environment is affected

### What you did
Based on the [12factor app](https://12factor.net/logs) recommendations (and ofc bcz it's logic) in Kubernetes I changed the logging output from File, Database etc ... to Console. GCE automatically offers you a very nice, feature rich logging system, where you can store and search logs very easily. 

### What happened
Multiline outputs are recorded as separate log events bcz is separated with new lines (like stack trace). So the first line of message <error> and the closing </error> 'tag' is logged separately (new insertId, new timestamp, etc ...).
But the bigger problem is I can't search in the full textPayload.

### Solution
One solution can be monolog, but only for formatting reason I didn't wanted to add a new package to my project
Another solution is the custom logger

I wrote an ugly ConsoleJsonLog class (bcz I needed a solution fast) what extends the ConsoleLog.
```php
public function log($level, $message, array $context = [])
{
    $this->_output->write(json_encode([
        'level' => $level,
        'date' => $this->_getFormattedDate(),
        'content' => $this->_format($message, $context),
        'context' => $context
    ]));
}
```

After this, all log entries (until now) are logged as 1 line (even multiline output) and I can make some ["crazy" query](https://cloud.google.com/logging/docs/view/advanced-queries#introduction).

My solution is not a really good solution, bcz:
* json_encode paramaters are not customizable
* I think BaseLog::_format not fully really neccessary
* what if $context is a resource
* maybe fallback if json_encode fails
* the 'content' key should be named 'message' : D

### Suggestion
In my opinion, the ConsoleLog should have a default 'json' (=false) option which accepts a boolean or an array (with json_encode params).
When it's enabled, the 'outputAs' must be forced to use RAW (???) and the output instead of `<level>%message</level>` must be a json string.

Before:
<img src="https://user-images.githubusercontent.com/6832955/110946520-3f1d5f80-833f-11eb-90d6-3db812162e6c.jpg" width=150 />

After:
<img src="https://user-images.githubusercontent.com/6832955/110946612-5ceac480-833f-11eb-8aa8-ee670baf57b4.jpg" width=150 />
[...]

PR #15543, opened 2021-05-23
Split message formatting from log engine into formatter.

othercorey:
Refs https://github.com/cakephp/cakephp/issues/15393

### DIFF
--- a/src/Log/Engine/BaseLog.php
+++ b/src/Log/Engine/BaseLog.php
@@ -18,7 +18,9 @@ namespace Cake\Log\Engine;
 
 use ArrayObject;
 use Cake\Core\InstanceConfigTrait;
-use DateTimeImmutable;
+use Cake\Log\Formatter\AbstractFormatter;
+use Cake\Log\Formatter\DefaultFormatter;
+use InvalidArgumentException;
 use JsonSerializable;
 use Psr\Log\AbstractLogger;
 use Serializable;
@@ -38,8 +40,13 @@ abstract class BaseLog extends AbstractLogger
     protected $_defaultConfig = [
         'levels' => [],
         'scopes' => [],
-        'dateFormat' => 'Y-m-d H:i:s',
+        'formatter' => DefaultFormatter::class,
     ];
+
+    /**
+     * @var \Cake\Log\Formatter\AbstractFormatter
+     */
+    protected $formatter;
 
     /**
      * __construct method
@@ -60,6 +67,23 @@ abstract class BaseLog extends AbstractLogger
 
         if (!empty($this->_config['types']) && empty($this->_config['levels'])) {
             $this->_config['levels'] = (array)$this->_config['types'];
+        }
+
+        // Default to DefaultFormatter for BC with custom loggers that don't set `formatter` in $_defaultConfig
+        $formatterConfig = $this->_config['formatter'] ?? DefaultFormatter::class;
+        if (!is_array($formatterConfig)) {
+            $formatterConfig = ['className' => $formatterConfig];
+        }
+
+        /** @var class-string<\Cake\Log\Formatter\AbstractFormatter> $className */
+        $className = $formatterConfig['className'];
+        $this->formatter = new $className($formatterConfig);
+        if (!$this->formatter instanceof AbstractFormatter) {
+            throw new InvalidArgumentException(sprintf(
+                'Formatter must extend `%s`, got `%s` instead',
+                AbstractFormatter::class,
+                $className
+            ));
         }
     }
 
@@ -92,8 +116,21 @@ abstract class BaseLog extends AbstractLogger
      * @param string $message The message to be formatted.
      * @param array $context Additional logging information for the message.
      * @return string
+     * @deprecated 4.3.0 Call `resolve()` directly from your log engine and format the message in a formatter.
      */
     protected function _format(string $message, array $context = []): string
+    {
+        return $this->resolve($message, $context);
+    }
+
+    /**
+     * Resolves interpolation expressions in message string.
+     *
+     * @param string $message Interpolated message
+     * @param array $context Interpolation expression values
+     * @return string
+     */
+    protected function resolve(string $message, array $context = []): string
     {
         if (strpos($message, '{') === false && strpos($message, '}') === false) {
             return $message;
@@ -166,17 +203,5 @@ abstract class BaseLog extends AbstractLogger
 
         /** @psalm-suppress InvalidArgument */
         return str_replace(array_keys($replacements), $replacements, $message);
-    }
-
-    /**
-     * Returns date formatted according to given `dateFormat` option format.
-     *
-     * This function affects `FileLog` or` ConsoleLog` datetime information format.
-     *
-     * @return string
-     */
-    protected function _getFormattedDate(): string
-    {
-        return (new DateTimeImmutable())->format($this->_config['dateFormat']);
     }
 }
--- a/src/Log/Engine/BaseLog.php
+++ b/src/Log/Engine/BaseLog.php
@@ -69,22 +69,27 @@ abstract class BaseLog extends AbstractLogger
             $this->_config['levels'] = (array)$this->_config['types'];
         }
 
-        // Default to DefaultFormatter for BC with custom loggers that don't set `formatter` in $_defaultConfig
-        $formatterConfig = $this->_config['formatter'] ?? DefaultFormatter::class;
-        if (!is_array($formatterConfig)) {
-            $formatterConfig = ['className' => $formatterConfig];
+        $formatter = $this->_config['formatter'] ?? DefaultFormatter::class;
+        if (!is_object($formatter)) {
+            if (is_array($formatter)) {
+                $class = $formatter['className'];
+                $options = $formatter;
+            } else {
+                $class = $formatter;
+                $options = [];
+            }
+            /** @var class-string<\Cake\Log\Formatter\AbstractFormatter> $class */
+            $formatter = new $class($options);
         }
 
-        /** @var class-string<\Cake\Log\Formatter\AbstractFormatter> $className */
-        $className = $formatterConfig['className'];
-        $this->formatter = new $className($formatterConfig);
-        if (!$this->formatter instanceof AbstractFormatter) {
+        if (!$formatter instanceof AbstractFormatter) {
             throw new InvalidArgumentException(sprintf(
                 'Formatter must extend `%s`, got `%s` instead',
                 AbstractFormatter::class,
-                $className
+                get_class($formatter)
             ));
         }
+        $this->formatter = $formatter;
     }
 
     /**
--- a/src/Log/Engine/ConsoleLog.php
+++ b/src/Log/Engine/ConsoleLog.php
@@ -17,6 +17,7 @@ declare(strict_types=1);
 namespace Cake\Log\Engine;
 
 use Cake\Console\ConsoleOutput;
+use Cake\Log\Formatter\DefaultFormatter;
 use InvalidArgumentException;
 
 /**
@@ -34,7 +35,10 @@ class ConsoleLog extends BaseLog
         'levels' => null,
         'scopes' => [],
         'outputAs' => null,
-        'dateFormat' => 'Y-m-d H:i:s',
+        'formatter' => [
+            'className' => DefaultFormatter::class,
+            'includeTags' => true,
+        ],
     ];
 
     /**
@@ -74,6 +78,11 @@ class ConsoleLog extends BaseLog
         if (isset($config['outputAs'])) {
             $this->_output->setOutputAs($config['outputAs']);
         }
+
+        if (isset($this->_config['dateFormat'])) {
+            deprecationWarning('`dateFormat` option should now be set in the formatter options.');
+            $this->formatter->setConfig('dateFormat', $this->_config['dateFormat']);
+        }
     }
 
     /**
@@ -88,8 +97,6 @@ class ConsoleLog extends BaseLog
     public function log($level, $message, array $context = [])
     {
         $message = $this->_format($message, $context);
-        $output = $this->_getFormattedDate() . ' ' . ucfirst($level) . ': ' . $message;
-
-        $this->_output->write(sprintf('<%s>%s</%s>', $level, $output, $level));
+        $this->_output->write($this->formatter->format($level, $message, $context));
     }
 }
--- a/src/Log/Engine/FileLog.php
+++ b/src/Log/Engine/FileLog.php
@@ -17,6 +17,7 @@ declare(strict_types=1);
 namespace Cake\Log\Engine;
 
 use Cake\Core\Configure;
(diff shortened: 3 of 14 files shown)


src/Log/Engine/ConsoleLog.php
```diff
@@ -74,6 +79,10 @@ public function __construct(array $config = [])
         if (isset($config['outputAs'])) {
             $this->_output->setOutputAs($config['outputAs']);
         }
+
+        if ($this->formatter && !isset($this->_config['formatter']['dateFormat'])) {
```
othercorey, line 83: This provides backwards compatibility where loggers don't have the dateFormat moved to the formatter config.

src/Log/Engine/BaseLog.php
```diff
@@ -61,6 +68,18 @@ public function __construct(array $config = [])
         if (!empty($this->_config['types']) && empty($this->_config['levels'])) {
             $this->_config['levels'] = (array)$this->_config['types'];
         }
+
+        $formatterConfig = $this->_config['formatter'] ?? null;
+        if (isset($formatterConfig)) {
```
othercorey, line 73: Formatters are optional for custom loggers especially for backwards compatibility.

src/Log/Engine/BaseLog.php
```diff
@@ -94,6 +113,18 @@ public function scopes()
...
+    }
+
+    /**
+     * Resolves interpolation expressions in message string.
+     *
+     * @param string $message Interpolated message
+     * @param array $context Interpolation expression values
+     * @return string
+     */
+    protected function resolve(string $message, array $context = []): string
```
othercorey, line 127: Split resolving interpolate strings from _format() so we can deprecated _format() in the future and provide engines with this common functionality.

Existing log engines require the message to be resolved before formatted.

src/Log/Engine/FileLog.php
```diff
@@ -128,13 +137,17 @@ public function log($level, $message, array $context = []): void
         $pathname = $this->_path . $filename;
         $mask = $this->_config['mask'];
         if (!$mask) {
-            file_put_contents($pathname, $output, FILE_APPEND);
+            foreach ($formatted as $output) {
+                file_put_contents($pathname, $output . "\n", FILE_APPEND);
```
othercorey, line 141: Not sure if it's cheaper to append a newline which creates a new string or call file_put_contents again with just `"\n"`.

src/Log/Formatter/AbstractFormatter.php
```diff
@@ -0,0 +1,42 @@
...
+
+    /**
+     * Formats message.
+     *
+     * @param mixed $level Logging level
+     * @param string $message Message string
+     * @param array $context Mesage context
+     * @return array<string> Array of formatted messages
+     */
+    abstract public function format($level, string $message, array $context = []): array;
```
othercorey, line 41: I would like to make level typed as string but PSR has `mixed`.

src/Log/Engine/BaseLog.php
```diff
@@ -61,6 +68,18 @@ public function __construct(array $config = [])
         if (!empty($this->_config['types']) && empty($this->_config['levels'])) {
             $this->_config['levels'] = (array)$this->_config['types'];
         }
+
+        $formatterConfig = $this->_config['formatter'] ?? null;
+        if (isset($formatterConfig)) {
+            $this->formatter = new $formatterConfig['className']($formatterConfig);
```
ADmad, line 74: We generally allow setting just class name strings too if no extra options are needed, so `'formatter' => Foo::class`.

src/Log/Engine/SyslogLog.php
```diff
@@ -81,6 +86,18 @@ class SyslogLog extends BaseLog
...
 
+    /**
+     * @inheritDoc
+     */
+    public function __construct(array $config = [])
+    {
+        parent::__construct($config);
+
+        if ($this->formatter && !isset($this->_config['formatter']['format'])) {
+            $this->formatter->setConfig('format', $this->_config['format']);
```
ADmad, line 97: Can't setting required config for the formatter be done at instance construction itself?

othercorey: There aren't shared/common configs between formatters so BaseLog doesn't know what to set.

src/Log/Engine/BaseLog.php
othercorey: Good catch. Remembered too late.

src/Log/Engine/SyslogLog.php
ADmad: Ah, I missed the fact that the log class' option is being set as the formatter's option. Still how about modifying `$config['formatter']` before calling `parent::__construct()`

src/Log/Formatter/SyslogFormatter.php
```diff
@@ -0,0 +1,47 @@
...
+
+class SyslogFormatter extends AbstractFormatter
+{
+    /**
+     * Default config for this class
+     *
+     * @var array
+     */
+    protected $_defaultConfig = [
+        'format' => '%s: %s',
```
markstory, line 27: Should we use named place holders? Positional arguments can be hard to expand in the future.

src/Log/Engine/BaseLog.php
ADmad: Could do that but that would make this one of the only class which has separate keys for class name and it's config.

src/Log/Engine/FileLog.php
dereuromark: This looks good to me like this.

src/Log/Engine/SyslogLog.php
othercorey: I was having trouble making it 100% BC. Custom loggers could override the `_defaultConfig` value, so I was using the result of `setConfig()` in `BaseLog::__construct()` to resolve that, then I looked to see it the value was set.

However, I'd like to remove the formatter-specific options from the log config in `_defaultConfig` and deprecate when it is set.

src/Log/Formatter/SyslogFormatter.php
othercorey: I have to allow copying the original log-level `format` option.

src/Log/Formatter/SyslogFormatter.php
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    /**
+     * @inheritDoc
+     */
+    public function format($level, string $message, array $context = []): array
+    {
+        return array_map(function ($line) use ($level) {
+            return sprintf($this->getConfig('format'), $level, $line);
+        }, explode("\n", $message));
```
othercorey, line 45: Should I make `SyslogEngine` split the message into lines after resolving the placeholders, and call `format()` for each one instead? This is the main reason why `format()` returns `array`. I don't know that other formatters would need to split up the message.

markstory: I think splitting the message up into multiple lines is something the log engine can do. Having that in the formatter makes formatters less reusable.

othercorey:
Deprecated formatter specific options in the log engine config. If they are set, the options are copied to the formatter. Custom log engines that set these should only see a deprecation warning.

DefaultFormatter is always created if a custom engine does not specify. This makes it easier to handle options and static analysis.

`AbstractFormatter::format()` now returns a string. SyslogEngine splits the message and formats each line.

`BaseLog::_getFormattedDate()` removed so the `dateFormat` default option in BaseLog can be removed. This function is no longer called.

src/Log/Engine/BaseLog.php
```diff
@@ -94,6 +116,18 @@ public function scopes()
      * @return string
      */
     protected function _format(string $message, array $context = []): string
```
othercorey, line 118: I would remove this as well, but it seems like a bigger BC break than removing `_getFormattedDate()`. Ideally, our log engines should only be calling `resolve()` before passing the message to the formatter.

othercorey: I can add the deprecated annotation and tell people to use `resolve()` directly and just suppress the psalm warning.

src/Log/Engine/BaseLog.php
```diff
@@ -61,6 +68,23 @@ public function __construct(array $config = [])
...
+        // Default to DefaultFormatter for BC with custom loggers that don't set `formatter` in $_defaultConfig
+        $formatterConfig = $this->_config['formatter'] ?? DefaultFormatter::class;
+        if (!is_array($formatterConfig)) {
+            $formatterConfig = ['className' => $formatterConfig];
+        }
+
+        /** @var class-string<\Cake\Log\Formatter\AbstractFormatter> $formatterClass */
+        $formatterClass = $formatterConfig['className'];
+        $this->formatter = new $formatterClass($formatterConfig);
+        if (!$this->formatter instanceof AbstractFormatter) {
```
markstory, line 81: You could use `ReflectionClass` to check this before constructing it. That would also allow you to skip the type annotation as you'd be able to use `$reflection->newInstance()`

src/Log/Engine/BaseLog.php
markstory: Instead of deprecating this, should this method call resolve and then use the formatter to format? Do engines need to do logic between these two steps?

src/Log/Engine/SyslogLog.php
```diff
@@ -106,10 +123,9 @@ public function log($level, $message, array $context = []): void
             $priority = $this->_levelMap[$level];
         }
 
-        $messages = explode("\n", $this->_format($message, $context));
-        foreach ($messages as $message) {
-            $message = sprintf($this->_config['format'], $level, $message);
-            $this->_write($priority, $message);
+        $lines = explode("\n", $this->_format($message, $context));
+        foreach ($lines as $line) {
+            $this->_write($priority, $this->formatter->format($level, $line, $context));
```
markstory, line 128: Should each line be formatted? We're mostly force-wrapping the long line to work around protocol deficiencies.

src/Log/Formatter/AbstractFormatter.php
```diff
@@ -0,0 +1,42 @@
...
+ * @copyright     Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
+ * @link          https://cakefoundation.org CakePHP(tm) Project
+ * @since         4.3.0
+ * @license       https://opensource.org/licenses/mit-license.php MIT License
+ */
+namespace Cake\Log\Formatter;
+
+use Cake\Core\InstanceConfigTrait;
+
+abstract class AbstractFormatter
```
markstory, line 21: You could also have an interface and have each class use the `InstanceConfigTrait` directly (if they want to).

src/Log/Formatter/SyslogFormatter.php
markstory: Fair point.

src/Log/Formatter/AbstractFormatter.php
othercorey: I was using an Interface, but I ended up with warnings because the ConsoleLog, FileLog and SyslogLog constructors were calling `setConfig()` on an interface.

src/Log/Engine/SyslogLog.php
othercorey: I don't use Syslog so I can't say. I was only trying to re-create the special formatting that existed. It was adding the level as a prefix to each line.

src/Log/Engine/BaseLog.php
othercorey: That's what it is doing.

src/Log/Engine/BaseLog.php
othercorey: Seems it made no difference to psalm

src/Log/Engine/BaseLog.php
```diff
@@ -61,6 +69,22 @@ public function __construct(array $config = [])
...
+
+        $class = new ReflectionClass($formatterConfig['className']);
+        if (!$class->isSubclassOf(AbstractFormatter::class)) {
+            throw new InvalidArgumentException(sprintf(
+                'Formatter must extend `%s`, got `%s` instead',
+                AbstractFormatter::class,
+                $class->getName()
+            ));
+        }
+        $this->formatter = $class->newInstance($formatterConfig);
```
ADmad, line 87: Uses reflection instead of `instanceof` check makes it more difficult for static analyzers to deduce the class type :slightly_smiling_face:.

othercorey: psalm throws the same error on assignment with `new class-string()`. Have to tell it what's being instantiated.

ADmad: You would need the inline docblock to indicate what's the class for that `class-string`, no way of avoiding that.

othercorey: I think going back to `new` + instanceof makes that cleaner. If using ReflectionClass doesn't help psalm then it's just harder to annotate.

src/Log/Engine/BaseLog.php
markstory: I meant that the call to `$this->formatter->format()` would also be in the `_format` method. But looking at it again that isn't great as the level would need to be passed requiring `_format` to change.

src/Log/Engine/SyslogLog.php
markstory: I guess this will work well with JSON formatting. For example logging a long string into a JSON format will result in multiple messages (one per line) and each line is JSON formatted with additional context.

src/Log/Formatter/AbstractFormatter.php
markstory: Oh. you'd have to have `setConfig()` in the interface as well if the logging engines are calling it. You could avoid having `setConfig()` in the interface if you shim the options differently but it would require slightly more complex shim code as you'd need to handle `formatter` being a string and an array.

You could do something like

```php
// in the logging engine
 if (isset($this->_config['format'])) {
    deprecationWarning('`format` option should now be set in the formatter options.');
    // new protected method to modify the formatter config.
    $config = $this->setFormatterOption($config, 'format');
}
parent::__construct($config);
```

othercorey: Typically, you use an Interface when passing an instance in. In this case, we're always instantiating the formatter so I'm not sure how well that works. We need it to have a constructor that accepts a $config array as well.

src/Log/Engine/BaseLog.php
```diff
@@ -61,6 +68,23 @@ public function __construct(array $config = [])
         if (!empty($this->_config['types']) && empty($this->_config['levels'])) {
             $this->_config['levels'] = (array)$this->_config['types'];
         }
+
+        // Default to DefaultFormatter for BC with custom loggers that don't set `formatter` in $_defaultConfig
+        $formatterConfig = $this->_config['formatter'] ?? DefaultFormatter::class;
```
markstory, line 73: Should we also allow the formatter to be passed as an instance? We do that for log/cache engines and it helps when someone wants to handle construction via a DI container or some more elaborate solution.

markstory (approved):
Looks good to me.

PR #15545, opened 2021-05-23
A custom message added in firstOrFail() method,

RiteshParyali:
Instead of  'Record not found in table "%s"' in response any one can add custom message.

### DIFF
--- a/src/Datasource/QueryTrait.php
+++ b/src/Datasource/QueryTrait.php
@@ -503,18 +503,22 @@ trait QueryTrait
     /**
      * Get the first result from the executing query or raise an exception.
      *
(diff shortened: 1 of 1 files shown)


othercorey:
Why does it matter what the exception message is?

ADmad:
`Record not found in table "%s"` is shown only when debug on. With debug off the `ExceptionRender` will show a general message `Not Found` for the error page. So allowing to modify the error message for `RecordNotFoundException` message doesn't make much sense.

If you want to customize the error message shown to end users with debug off then you should use a [custom exception renderer](https://book.cakephp.org/4/en/development/errors.html#custom-exceptionrenderer).

PR #15548, opened 2021-05-24
Fix error when accessing PDOStatement::$queryString.

### LINKED ISSUE #15537: Cannot access PDOStatement::$queryString before initialized

ADmad:
On PHP 8.1 trying to access PDOStatement::$queryString before it's initialized results in
"Error: Typed property PDOStatement::$queryString must not be accessed before initialization".

Closes #15537

### DIFF
--- a/src/Database/Statement/PDOStatement.php
+++ b/src/Database/Statement/PDOStatement.php
@@ -47,6 +47,20 @@ class PDOStatement extends StatementDecorator
     }
 
     /**
+     * Magic getter to return PDOStatement::$queryString as read-only.
+     *
+     * @param string $property internal property to get
+     * @return mixed
+     */
+    public function __get(string $property)
+    {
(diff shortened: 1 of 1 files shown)


othercorey:
This should work.

PR #15551, opened 2021-05-25
Move fixture reset management to test case

markstory:
Having the reset strategy be global across the application complicates a few scenarios:

1. If an application has some end-to-end tests they would be forced to    use `TruncationStrategy` even for their integration tests.
2. Having multiple state managers that are used in different tests gets    harder. With this approach if a test only uses elasticsearch it can    use the elasticsearch state reset (once it exists).

In addition to these specfic problems having the state manager on the test case makes more of the framework's logic extensible and replaceable should an application need to.

For the core tests having this will unblock us to use the `TransactionStrategy` on a large number of our tests which should yield even faster test durations.

Part of #15308

### DIFF
--- a/src/TestSuite/Fixture/TransactionStrategy.php
+++ b/src/TestSuite/Fixture/TransactionStrategy.php
@@ -33,53 +33,32 @@ class TransactionStrategy implements StateResetStrategyInterface
     /**
      * Constructor.
      *
-     * @param bool $enableLogging Whether or not to enable query logging.
+     * @param \Cake\TestSuite\Fixture\FixtureLoader $fixtures The fixture loader being used.
      * @return void
      */
-    public function __construct(bool $enableLogging = false)
+    public function __construct(FixtureLoader $fixtures)
     {
-        $this->aliasConnections($enableLogging);
+        $this->checkConnections();
     }
 
     /**
-     * Alias non test connections to the test ones
-     * so that models reach the test database connections instead.
+     * Ensure that all `Connection` connections support savepoints.
      *
-     * @param bool $enableLogging Whether or not to enable query logging.
      * @return void
      */
-    protected function aliasConnections(bool $enableLogging): void
+    protected function checkConnections(): void
     {
         $connections = ConnectionManager::configured();
-        ConnectionManager::alias('test', 'default');
-        $map = [];
-        foreach ($connections as $connection) {
-            if ($connection === 'test' || $connection === 'default') {
-                continue;
-            }
-            if (isset($map[$connection])) {
-                continue;
-            }
-            if (strpos($connection, 'test_') === 0) {
-                $map[$connection] = substr($connection, 5);
-            } else {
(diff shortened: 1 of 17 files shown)


src/TestSuite/Fixture/TransactionStrategy.php
```diff
@@ -33,65 +33,43 @@ class TransactionStrategy implements StateResetStrategyInterface
     /**
      * Constructor.
      *
-     * @param bool $enableLogging Whether or not to enable query logging.
+     * @param \Cake\TestSuite\Fixture\FixtureLoader $fixtures The fixture loader being used.
      * @return void
      */
-    public function __construct(bool $enableLogging = false)
+    public function __construct(FixtureLoader $fixtures)
```
othercorey, line 39:
```suggestion
    public function __construct(FixtureLoader $loader)
```

src/TestSuite/Fixture/TruncationStrategy.php
```diff
@@ -47,63 +46,28 @@ class TruncationStrategy implements StateResetStrategyInterface
...
-     * Alias non test connections to the test ones
-     * so that models reach the test database connections instead.
+     * Constructor.
      *
-     * @param bool $enableLogging Whether or not to enable query logging.
+     * @param \Cake\TestSuite\Fixture\FixtureLoader $fixtures The fixture loader being used.
      * @return void
      */
-    protected function aliasConnections(bool $enableLogging): void
+    public function __construct(FixtureLoader $fixtures)
```
othercorey, line 59:
```suggestion
    public function __construct(FixtureLoader $loader)
```

src/TestSuite/Fixture/TruncationStrategy.php
```diff
@@ -47,63 +46,28 @@ class TruncationStrategy implements StateResetStrategyInterface
...
-            }
-        }
-        foreach ($map as $testConnection => $normal) {
-            ConnectionManager::alias($testConnection, $normal);
-            $connection = ConnectionManager::get($normal);
-            if ($connection instanceof Connection && $enableLogging) {
-                $connection->enableQueryLogging();
-            }
-        }
+        $this->fixtures = $fixtures;
```
othercorey, line 61:
```suggestion
        $this->fixtures = $loader;
```

I don't know if `$this->fixtures` should be renamed or if used for other legacy code.

othercorey:
Looks ok from a high level.

src/TestSuite/Fixture/TruncationStrategy.php
markstory: There is no legacy here as this class is new. I can rename it.

othercorey:
> codecov/patch — 59.32%

not great

markstory:
> not great

Yes the phpunit extension runs during phpunit startup before coverage is enabled. I can add some rudimentary tests for it.

PR #15553, opened 2021-05-25
Remove php 7.4 from build matrix

othercorey:
Trying to further reduce our build matrix after removing redundant php 7.2 + mysql build.

Since we don't use php 7.4 or 8.0 features, just testing against 8.0 should cover any deprecations or changes to PDO. If we bump mariadb to 7.4, it covers most of it.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,14 +17,14 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        php-version: ['7.2', '7.4', '8.0']
+        php-version: ['7.2', '8.0']
         db-type: [sqlite, mysql, pgsql]
         prefer-lowest: ['']
         exclude:
           - php-version: '7.2'
             db-type: 'mysql'
         include:
-          - php-version: '7.2'
+          - php-version: '7.4'
             db-type: 'mariadb'
           - php-version: '7.2'
             db-type: 'mysql'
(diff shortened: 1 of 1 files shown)


inoas:
This however assumes that the php team does 100% perfect work and there are no fixes in 8.0 that are not in 7.4 (backports?).

7.4 may be prevalent server eco systems for a very long time so as a testing target that may be very valuable to catch possible differences between  7.2, 7.4 and 8.0?

dereuromark:
@ionas It is highly unlikely that a min/max setup of testing does not catch any of the issues you mentioned.
I also do that for 25+ plugins for years, as well as thousands of company packages and so far without a single incident.

dereuromark:
At some point, for the next 4.x minor it could just make more sense to raise the minimum to 7.3 or 7.4 etc, and drop long EOL versions. But that will also be an easy adjustment on the matrix here.

othercorey:
> 7.4 may be prevalent server eco systems for a very long time so as a testing target that may be very valuable to catch possible differences between 7.2, 7.4 and 8.0?

I don't know of any PDO-specific changes that were introduced in 7.4 and *removed* in 8.0. We are still running 7.4 for mariadb.

dereuromark:
> I don't know of any PDO-specific changes that were introduced in 7.4 

This is also void, as it would still need to pass 7.2 anyway (which is the min), therefore there is no such possibility for any such functionality.

othercorey:
Well, it could be behavior changes, but, again, I don't know of anything that was introduced and then reverted. PHP is typically more consistent than that. Here is an example of a PDO behavior change (in 8.1):

> PDO SQLite:  Integers and floats in results sets will now be returned using native PHP types. You can restore the previous behavior by enabling the PDO::ATTR_STRINGIFY_FETCHES option.

ADmad:
> We are still running 7.4 for mariadb.

That's good enough.

PR #15555, opened 2021-05-25
Extract action args generation into a separate method.

ADmad:
This makes the action invocation process easier to grok and also makes overriding
ControllerFactory::invoke() easier if required.


othercorey:
Looks ok.

PR #15558, opened 2021-05-26
Allow specifying middlewares at the controller level.

ADmad:
While we can setup route based middlewares at times setting them can be
a bit cumbersome. If for e.g. you want a middleware to run for just one
action you would have to setup an extra route for it.

Middlewares for controllers provides a good convenience and also allow replacing
some components with middlewares.

P. S. Laravel allows setting up middleware for controller in a similar way.

### DIFF
--- a/src/Controller/Controller.php
+++ b/src/Controller/Controller.php
@@ -162,6 +162,14 @@ class Controller implements EventListenerInterface, EventDispatcherInterface
     protected $plugin;
 
     /**
+     * Middlewares list.
+     *
+     * @var array
+     * @psalm-var array<int, array{middleware:string|\Closure|\Psr\Http\Server\MiddlewareInterface, options:array}>
+     */
+    protected $middlewares = [];
+
+    /**
      * Constructor.
      *
      * Sets a number of properties based on conventions if they are empty. To override the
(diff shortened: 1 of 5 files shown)


src/Controller/Controller.php
```diff
@@ -544,6 +552,57 @@ public function invokeAction(Closure $action, array $args): void
...
+            'options' => $options,
+        ];
+    }
+
+    /**
+     * Get middleware to be applied for this controller.
+     *
+     * @return array
+     */
+    public function getMiddleware(): array
```
othercorey, line 578:
```suggestion
    public function getMiddlewares(): array
```

This should be plural, right?

ADmad: Yes, but I was going for consistency since `RouteCollection`, `RouteBuilder` and `Route` also have methods of the same name which return arrays :slightly_frowning_face:.

markstory: Should the property also be `middleware` then?

src/Controller/Controller.php
```diff
@@ -544,6 +552,57 @@ public function invokeAction(Closure $action, array $args): void
...
+     * Register middleware for the controller.
+     *
+     * @param string|\Closure|\Psr\Http\Server\MiddlewareInterface $middleware Middleware.
+     * @param array $options Valid options:
+     *  - `only`: (array|string) Only run the middleware for specified actions.
+     *  - `except`: (array|string) Run the middleware for all actions except the specified ones.
+     * @return void
+     * @psalm-var array{?only: string|array, ?except: string|array}
+     */
+    public function middleware($middleware, array $options = [])
```
markstory, line 565: Should this method raise an exception if it is called after middleware have already been applied? It would be frustrating to use this method in a `beforeRender` and then think that Cake is 'broken' when you're using it wrong.

src/Controller/ControllerFactory.php
```diff
@@ -94,6 +102,32 @@ public function create(ServerRequestInterface $request): Controller
      */
     public function invoke($controller): ResponseInterface
     {
+        $this->controller = $controller;
+
+        $middlewares = $controller->getMiddleware();
+
+        if ($middlewares) {
+            $middlewareQueue = new MiddlewareQueue($controller->getMiddleware());
```
markstory, line 110:
```suggestion
            $middlewareQueue = new MiddlewareQueue($middlewares);
```

tests/test_app/TestApp/Controller/PostsController.php
```diff
@@ -32,6 +32,16 @@ public function initialize(): void
         $this->loadComponent('Flash');
         $this->loadComponent('RequestHandler');
         $this->loadComponent('FormProtection');
+
+        $this->middleware(function ($request, $handler) {
```
markstory, line 36: This can place nice with the DI container as middleware could be constructor parameters too :+1:

src/Controller/Controller.php
ADmad: Could be start with just documenting the fact that middlewares must be set from `initialize()` only?

I don't think we would be able to check if the middleware queue was already created without adding a new property.

tests/test_app/TestApp/Controller/PostsController.php
ADmad: Yeah the controller factory has the reference to the container instance so it opens up possibilities.

src/Controller/Controller.php
markstory: Fair enough.

src/Controller/Controller.php
markstory: Yeah we would need another property to track whether or not middleware could be added.

ADmad:
Can this be merged?

PR #15560, opened 2021-05-28
Make SQL logging in tests much simpler to operate

markstory:
Instead of having to edit the phpunit.xml and/or application configuration to enable logging, we can use the `--debug` flag that PHPUnit has to also enable debug mode for our fixtures and database package. This feature will only work with the new fixture extension.

Part of #15308

### DIFF
--- a/src/TestSuite/FixtureSchemaExtension.php
+++ b/src/TestSuite/FixtureSchemaExtension.php
@@ -36,26 +36,14 @@ class FixtureSchemaExtension implements BeforeFirstTestHook
         FixtureLoader::setInstance(new FixtureDataManager());
 
         $enableLogging = in_array('--debug', $_SERVER['argv'] ?? [], true);
-        $this->enableLogging($enableLogging);
         $this->aliasConnections($enableLogging);
-    }
-
-    /**
-     * Enables a logger for SQL queries.
-     *
-     * @param bool $enableLogging Whether or not logging should be enabled.
-     * @return void
-     */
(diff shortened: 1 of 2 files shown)


src/TestSuite/FixtureSchemaExtension.php
```diff
@@ -34,12 +37,29 @@ public function __construct()
...
         $this->aliasConnections($enableLogging);
     }
 
+    /**
+     * Configures a logger for SQL queries.
+     *
+     * @param bool $enableLogging Whether or not logging should be enabled.
+     * @return void
+     */
+    protected function configureLogging(bool $enableLogging): void
```
othercorey, line 51: Think this should be enableLogging. Calling `configure` with false that just early returns seems odd.

src/TestSuite/FixtureSchemaExtension.php
```diff
@@ -34,12 +35,29 @@ public function __construct()
...
 
+    /**
+     * Enables a logger for SQL queries.
+     *
+     * @param bool $enableLogging Whether or not logging should be enabled.
+     * @return void
+     */
+    protected function enableLogging(bool $enableLogging): void
+    {
+        if (!$enableLogging) {
```
ADmad, line 51: Is the argument necessary if it's a no-op when `false`?

markstory: I could move the conditional into constructor instead.

PR #15561, opened 2021-05-28
Export __debugInfo() variables only

othercorey:
This early return was lost when the debugger node exports were re-written.


othercorey:
@robertpustulka

PR #15563, opened 2021-05-29
Deprecate syslog formatting with custom format string

othercorey:
This renames SyslogFormatter to LegacySyslogFormatter so the intention is clear.

Added `includeDate` option to `DefaultFormatter` so it works with Syslog. We had to pick one format for the level, so switched console + file log tests. If the capitalization of the level is really important, we can change Syslog instead, but using the raw string seemed cleaner.

### DIFF
--- a/tests/TestCase/Log/Engine/SyslogLogTest.php
+++ b/tests/TestCase/Log/Engine/SyslogLogTest.php
@@ -111,6 +111,18 @@ class SyslogLogTest extends TestCase
     }
 
     /**
+     * Test deprecated format option
+     *
+     * @return void
+     */
+    public function testDeprecatedFormatMessage()
+    {
+        $this->expectDeprecation();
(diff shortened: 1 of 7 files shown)


tests/TestCase/Log/Engine/SyslogLogTest.php
```diff
@@ -110,6 +110,18 @@ public function testDeprecatedFormat()
...
 
+    /**
+     * Test deprecated format option
+     *
+     * @return void
+     */
+    public function testDeprecatedFormatMessage()
+    {
+        $this->expectDeprecation();
+        $this->expectDeprecationMessage('`format` option is now deprecated in favour of custom formatters');
```
dereuromark, line 121:
```suggestion
        $this->expectDeprecationMessage('`format` option is now deprecated in favor of custom formatters');
```
We use AE usually in coding (colour vs color etc, behavior, ...)

PR #15570, opened 2021-05-31
Fix up more controller casing issues

dereuromark:
Follow https://github.com/cakephp/cakephp/pull/15568
This removes almost all cases, with the exception of 1-2, that are hard to resolve.

The tests kind of show that the inflection not always works out of the box.
But I think we should adjust the tests to match the reality (CamelCase controllers always, same for plugin and prefix) and not the other way around.
Otherwise people can be confused if they see this syntax and try it, and it just doesn't work.

### DIFF
--- a/tests/TestCase/Routing/Route/RedirectRouteTest.php
+++ b/tests/TestCase/Routing/Route/RedirectRouteTest.php
@@ -48,8 +48,8 @@ class RedirectRouteTest extends TestCase
      */
     public function testMatch()
     {
-        $route = new RedirectRoute('/home', ['controller' => 'posts']);
-        $this->assertNull($route->match(['controller' => 'posts', 'action' => 'index']));
+        $route = new RedirectRoute('/home', ['controller' => 'Posts']);
+        $this->assertNull($route->match(['controller' => 'Posts', 'action' => 'index']));
     }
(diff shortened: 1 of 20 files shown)


tests/TestCase/Routing/Route/RedirectRouteTest.php
```diff
@@ -72,9 +72,9 @@ public function testParseMiss()
     public function testParseSimple()
     {
         $this->expectException(RedirectException::class);
-        $this->expectExceptionMessage('http://localhost/posts');
+        $this->expectExceptionMessage('http://localhost/Posts');
```
othercorey, line 75: Why would the URL change too?

othercorey: I was thinking a different route was default.

inmun: [DashedRoute.php](https://github.com/cakephp/cakephp/blob/master/src/Routing/Route/DashedRoute.php) isn't loaded.

dereuromark: Or any inflecting route for that matter :)

PR #15571, opened 2021-06-01
Remove additional wrong DS

dereuromark:
The paths were wrongly double slashed:

> vendor/dereuromark/cakephp-queue/templates/bake//Task/task.twig

Every path was already always with trailing DS, and every name including subnames always without DS.
Except for this one case it seems.

### DIFF
--- a/src/View/View.php
+++ b/src/View/View.php
@@ -1339,7 +1339,7 @@ class View implements EventDispatcherInterface
             } elseif (!$plugin || $this->templatePath !== $this->name) {
                 $name = $templatePath . $subDir . $name;
             } else {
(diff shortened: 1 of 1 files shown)


src/View/View.php
```diff
@@ -1339,7 +1339,7 @@ protected function _getTemplateFileName(?string $name = null): string
             } elseif (!$plugin || $this->templatePath !== $this->name) {
                 $name = $templatePath . $subDir . $name;
             } else {
-                $name = DIRECTORY_SEPARATOR . $subDir . $name;
+                $name = $subDir . $name;
```
ADmad, line 1342: The lack of test coverage is probably why the issue went unnoticed :)

PR #15572, opened 2021-06-04
Set allowFallbackClass to TableLocator config

cleptric:
This adds support to use the database session handler without having to enable `TableLocator::allowFallbackClass` or creating a specific `Table::class`.

### DIFF
--- a/src/Http/Session/DatabaseSession.php
+++ b/src/Http/Session/DatabaseSession.php
@@ -57,7 +57,7 @@ class DatabaseSession implements SessionHandlerInterface
         $tableLocator = $this->getTableLocator();
 
         if (empty($config['model'])) {
-            $config = $tableLocator->exists('Sessions') ? [] : ['table' => 'sessions'];
+            $config = $tableLocator->exists('Sessions') ? [] : ['table' => 'sessions', 'allowFallbackClass' => true];
             $this->_table = $tableLocator->get('Sessions', $config);
         } else {
             $this->_table = $tableLocator->get($config['model']);


ADmad (approved):
Change looks good but it's not a defect :). Fallback table class is enabled by default globally. You want to override the global config.

cleptric:
The skeleton disables the fallback currently: https://github.com/cakephp/app/blob/master/src/Application.php#L55
As this happens so early in the request, I thought this might not be the intended behavior when using database sessions.

src/Http/Session/DatabaseSession.php
```diff
@@ -57,7 +57,7 @@ public function __construct(array $config = [])
         $tableLocator = $this->getTableLocator();
 
         if (empty($config['model'])) {
-            $config = $tableLocator->exists('Sessions') ? [] : ['table' => 'sessions'];
+            $config = $tableLocator->exists('Sessions') ? [] : ['table' => 'sessions', 'allowFallbackClass' => true];
             $this->_table = $tableLocator->get('Sessions', $config);
```
othercorey, line 61: The config is ignored if the table exists, right? Do we need this extra logic?

ADmad: No, the config is not ignored. If an alias is already present is the registry and you call get() again for it with a different config then it will throw an exception. So the extra logic is necessary.

othercorey: that's right, my mistake.

PR #15574, opened 2021-06-09
Add 8.1.0alpha1 to build matrix

othercorey:
I don't see an option for 8.1.0alpha1 so this is just nightly still.


ADmad:
Testuite failure on PHP 8.1 doesn't fail CI any more hence merging. The unrelated test failure will be fixed in #15577.

PR #15578, opened 2021-06-09
Update return types for session handler methods.

### LINKED ISSUE #15575: PHP 8.1: JsonSerializable::jsonSerialize() must return mixed
> Deprecated: Declaration of Cake\I18n\Time::jsonSerialize() should be compatible with JsonSerializable::jsonSerialize(): mixed in /home/runner/work/cakephp/cakephp/src/I18n/Time.php on line 446

This can be suppressed by adding `ReturnTypeWillChange` attribute.

### LINKED ISSUE #15576: PHP 8.1: SessionHandlerInterface::gc must return int|false
> PHP Fatal error:  During inheritance of SessionHandlerInterface: Uncaught Declaration of Cake\Http\Session\CacheSession::gc($maxlifetime): bool should be compatible with SessionHandlerInterface::gc(int $max_lifetime): int|false

This is throwing a fatal error instead of a deprecation notice so the `ReturnTypeWillChange` attribute won't suppress. Not sure if that should be suppressible or not.

The implementation probably needs to be changed to return the proper type.

ADmad:
The return types now match those of `SessionHandlerInterface` methods.

https://www.php.net/manual/en/class.sessionhandlerinterface.php

### DIFF
--- a/src/Http/Session/CacheSession.php
+++ b/src/Http/Session/CacheSession.php
@@ -20,6 +20,7 @@ namespace Cake\Http\Session;
 
 use Cake\Cache\Cache;
 use InvalidArgumentException;
+use ReturnTypeWillChange;
 use SessionHandlerInterface;
 
 /**
@@ -55,11 +56,11 @@ class CacheSession implements SessionHandlerInterface
     /**
      * Method called on open of a database session.
      *
-     * @param string $savePath The path where to store/retrieve the session.
+     * @param string $path The path where to store/retrieve the session.
      * @param string $name The session name.
      * @return bool Success
      */
-    public function open($savePath, $name): bool
+    public function open($path, $name): bool
     {
         return true;
     }
@@ -78,14 +79,15 @@ class CacheSession implements SessionHandlerInterface
      * Method used to read from a cache session.
      *
      * @param string $id ID that uniquely identifies session in cache.
-     * @return string Session data or empty string if it does not exist.
+     * @return string|false Session data or false if it does not exist.
      */
-    public function read($id): string
+    #[ReturnTypeWillChange]
+    public function read($id)
     {
         $value = Cache::read($id, $this->_options['config']);
 
-        if (empty($value)) {
-            return '';
+        if ($value === null) {
+            return false;
         }
 
         return $value;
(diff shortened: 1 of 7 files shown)


ADmad:
```
PHP Fatal error:  During inheritance of SessionHandlerInterface: Uncaught Declaration of Cake\Http\Session\CacheSession::read($id) should be compatible with SessionHandlerInterface::read(string $id): string|false
```
We can't concrete return type `string|false` as we need to support lower PHP versions too :angry:. This looks like a bug to me, it should not be a fatal error.

othercorey:
> We can't add concrete return type `string|false` as we need to support lower PHP versions too 😠.
> This looks like a bug to me, it should not be a fatal error but rather a deprecated error.

This might be a side-effect of implementing an internal interface instead of extending an internal class. It probably should be supressible with `ReturnTypeWillChange` attribute. If not, probably need to submit a php ticket.

ADmad:
> This might be a side-effect of implementing an internal interface instead of extending an internal class. 

The manual actually has a warning to implement the interface instead of extension [SessionHandler]((https://www.php.net/manual/en/class.sessionhandler.php) class.

It's definitely a bug IMO, missing return type when implementing an interface should also be a deprecated error not fatal, like it's when extending classes.

othercorey:
When testing manually, it prints a deprecation message as expected. Adding the `ReturnTypeWillChange` attribute does work:

https://3v4l.org/RWpMT/rfc#git.master

ADmad:
This is good to go. The phpcs/stan failures are invalid/irrelevant which I will take care of in a separate PR.

src/Http/Session/CacheSession.php
```diff
@@ -20,6 +20,7 @@
 
 use Cake\Cache\Cache;
 use InvalidArgumentException;
+use ReturnTypeWillChange;
```
markstory, line 23: I can't find docs for this, but I'm going to assume PHP8.1 reads this attribute and turns of return type comparisons?

markstory (approved):

src/Http/Session/CacheSession.php
othercorey: It isn't well documented, but this attribute will suppress the deprecation of mismatched tentative return types.

https://wiki.php.net/rfc/internal_method_return_types

Issue #8099, opened 2016-01-24
3.x: Fallback error templates to non ext ones


dereuromark:
In 3.1 latest, when trying to work with extensions, one runs into:

```
Fatal error: [Cake\View\Exception\MissingTemplateException] Template file "Error\pdf\error500.ctp" is missing
```

When opening some `/controller/action.pdf` URL.

Sure, I don't want to have custom error templates for all extensions.
My assumption would be that it will fallback to non-ext ones (Error\error500.ctp) from Cake itself, when this specific layout is not found.
This would make it more DRY, because I only have to add error templates for the cases where I want to change error behavior (or stay inside the current type) based on the extension.

After some debugging I found out that it wanted to throw an error `'Error\pdf\missing_helper'` because of a missing helper, but ended up throwing fatals because of missing error layout it should not necessarily expect in the first place.

ionas:
:+1:

markstory:
Turning off extensions in `_outputMessageSafe` sounds reasonable to me. That method is only invoked when rendering the original error page has failed.

markstory:
Is PDF rendering being handled by a view class? I tried to replicate this with the standard View class and didn't have any success. I did the following:

1. Enable RequestHandlerComponent
2. Add route extension `$routes->extensions(['json', 'csv']);`
3. Make `src/Template/Bookmarks/csv/index.ctp` that loads an invalid helper.

The MissingHelper page was rendered correctly as expected.

dereuromark:
Yes PDFview class from foc.

hmic:
:+1: 
This buggs me long time...

Whenever you use a custom viewclass with an extension, you cannot render the default html error templates and all exceptions thrown cascade to a 500 because of missing layout/template/...

markstory:
> Yes PDFview class from foc.

Sounds like that view class needs fixing then.

markstory:
There is not much that error handling can do if custom view classes are manipulating the view template paths.

hmic:
Any custom view class I've seen does it this way.

So whats the correct option to do so?

Dakota/Excel does the same and suffers the same problems e.g.
This viewClass does binary output, like the pdf view too, whats the correct approach other than saying they need fixing? - Seems nobody has come up with a better approach or fix for now...

Shed light on us please!

dereuromark:
And none of those custom view classes actually manipulate error handling.
It should just fallback on the default ones for exception handling in the case spefied above.

markstory:
[...]

PR #15582, opened 2021-06-12
Try to find error template in parent directories

othercorey:
Closes https://github.com/cakephp/cakephp/issues/8099

This attempts to walk the parent directories looking for application error template before bailing out to safe rendering.

Want to get feedback on the implementation before writing the complicated tests for each of the scenarios.

1. Remove extension/subDir
2. Remove prefix
3. Remove plugin
4. Remove custom template name

All non-template exceptions are now considered internal exceptions. This means rendering errors don't stomp on the original exception and are handled by `ErrorHandlerMiddleware::handleInternalError()`. Most internal errors just need to be logged anyway as something is misconfigured.

### DIFF
--- a/src/Error/ExceptionRenderer.php
+++ b/src/Error/ExceptionRenderer.php
@@ -23,7 +23,6 @@ use Cake\Core\App;
 use Cake\Core\Configure;
 use Cake\Core\Container;
 use Cake\Core\Exception\CakeException;
-use Cake\Core\Exception\MissingPluginException;
 use Cake\Datasource\Exception\PageOutOfBoundsException;
 use Cake\Datasource\Exception\RecordNotFoundException;
 use Cake\Event\Event;
@@ -276,7 +275,18 @@ class ExceptionRenderer implements ExceptionRendererInterface
         }
         $this->controller->setResponse($response);
 
-        return $this->_outputMessage($template);
+        try {
+            return $this->renderTemplate($template);
+        } catch (MissingLayoutException $e) {
+            throw $e;
+        } catch (MissingTemplateException $e) {
+            $this->controller->viewBuilder()
+                ->setHelpers([], false)
+                ->setLayoutPath('')
+                ->setTemplatePath('Error');
+
+            return $this->renderTemplate('error500');
+        }
     }
 
     /**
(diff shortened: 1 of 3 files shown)


src/Error/ExceptionRenderer.php
```diff
@@ -429,11 +437,20 @@ protected function _outputMessageSafe(string $template): Response
             ->setHelpers([], false)
             ->setLayoutPath('')
             ->setTemplatePath('Error');
-        $view = $this->controller->createView('View');
+
+        try {
+            // Try with app View first with default templates
+            $view = $this->controller->createView();
```
othercorey, line 443: This helps applications with custom views like TwigView. They should be allowed to render the default error500 template first.

src/Error/ExceptionRenderer.php
```diff
@@ -390,29 +390,37 @@ protected function _outputMessage(string $template): Response
...
+                $this->controller->viewBuilder()->setPlugin(null);
+
+                return $this->_outputMessage($template);
             }
 
-            return $this->_outputMessageSafe('error500');
-        } catch (Throwable $e) {
-            return $this->_outputMessageSafe('error500');
+            if ($template !== 'error500') {
+                return $this->_outputMessage('error500');
```
markstory, line 419: :crossed_fingers: that these 10 ish additional stack frames don't blow the limit :smile:

tests/TestCase/Error/ExceptionRendererTest.php
```diff
@@ -702,61 +702,6 @@ public function testExceptionNameMangling()
...
-            ->with('missingHelper')
-            ->will($this->throwException($exception));
-
-        $ExceptionRenderer->setController($controller);
-
-        $response = $ExceptionRenderer->render();
-        $helpers = $controller->viewBuilder()->getHelpers();
-        sort($helpers);
-        $this->assertEquals([], $helpers);
-        $this->assertStringContainsString('Helper class Fail', (string)$response->getBody());
```
markstory, line 732: Does this error no longer get rendered?

src/Error/ExceptionRenderer.php
```diff
@@ -390,29 +390,37 @@ protected function _outputMessage(string $template): Response
...
-        } catch (MissingPluginException $e) {
-            $attributes = $e->getAttributes();
-            if (isset($attributes['plugin']) && $attributes['plugin'] === $this->controller->getPlugin()) {
+            if ($request->getParam('prefix')) {
+                $this->controller->setRequest($request->withParam('prefix', null));
+
+                return $this->_outputMessage($template);
+            }
+
+            if ($this->controller->getPlugin()) {
```
ADmad, line 411: We could skip this check. Having the controller's plugin property not match the one in the request should be discouraged IMO. I would even like to deprecate this method.

src/Error/ExceptionRenderer.php
othercorey: I thought of that too. This re-uses the original functions as-is, but will see if I have an idea how to iterate without recursion.

tests/TestCase/Error/ExceptionRendererTest.php
othercorey: All non-template errors are treated as internal errors now and trigger the fallback case in ErrorHandlerMiddleware.

src/Error/ExceptionRenderer.php
othercorey: This is an awkward one. I wasn't sure what to when rendering something like a pdf from a plugin which doesn't have a pdf-specific error template. Do we check for plugin-specific error template or jump back to the application templates immediately?

ADmad: I doubt anyone is going to generate an actual pdf for an error page.

othercorey: Right. The purpose of this PDF allowing the non-pdf error template to render.

othercorey:
This now loop until a non-template exceptions thrown or rendering succeeds.

All non-template exceptions are treated as internal errors. If the template is not found walking up the request hierarchy, then 'error500' is rendered with minimal controller configuration. If that fails, it's treated as an internal error as before.

Replaced `_outputMessage()` and `_outputMessageSafe()` with `renderTemplate()`.

src/Error/ExceptionRenderer.php
```diff
@@ -276,7 +276,20 @@ public function render(): ResponseInterface
         }
         $this->controller->setResponse($response);
 
-        return $this->_outputMessage($template);
+        try {
+            return $this->renderTemplate($template);
+        } catch (MissingTemplateException $e) {
+            if (!$e instanceof MissingLayoutException) {
```
othercorey, line 282: This is due to MissingLayoutException extending MissingTemplateException. I could either catch and re-throw it like in renderTemplate() or do this. This seemed cleaner for this one scenario.

ADmad:
> Replaced _outputMessage() and _outputMessageSafe() with renderTemplate().

Boo, I will have to  update the exception renderer of Crud plugin :(

othercorey:
> Boo, I will have to update the exception renderer of the Crud plugin :(

Can you link that? I'd like to see an example where that's overriden.

ADmad:
> Can you link that? I'd like to see an example where that's overridden.

https://github.com/FriendsOfCake/crud/blob/4b13c38d4f442cabcdc720f6ffbb7d00a163677d/src/Error/ExceptionRenderer.php#L67

othercorey:
That code should work the same with `renderTemplate()`. I changed it because it doesn't really output anything and there isn't a difference between outputMessaage and outputMessageSafe.

I can just remove outputMessageSafe() and keep `_outputMessage()` if it's going to be annoying.

othercorey:
This is impossible.

PR #15583, opened 2021-06-13
Use DefaultFormatter with ArrayLog

othercorey:


othercorey:
@markstory when you have time, can you double-check this looks ok?

markstory (commented):
Looks good to me.

PR #15589, opened 2021-06-16
Update test for var exporting recursive array.

ADmad:
~~$GLOBALS is gone in PHP 8.1.~~
Behavior of $GLOBALS has changed in PHP 8.1.

### DIFF
--- a/tests/TestCase/Error/DebuggerTest.php
+++ b/tests/TestCase/Error/DebuggerTest.php
@@ -748,14 +748,17 @@ TEXT;
     }
 
     /**
-     * Test that exportVar() will stop traversing recursive arrays like GLOBALS.
+     * Test that exportVar() will stop traversing recursive arrays.
      *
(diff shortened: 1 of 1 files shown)


othercorey:
It's not technically gone, but you can't modify the array.

https://github.com/php/php-src/blob/master/UPGRADING#L23

othercorey:
@markstory do we really need to modify the array passed in?

othercorey (approved):

ADmad:
I haven't looked into the innards of `Debugger::exportVar()`, so don't know if it actually modifies `$GLOBALS`. It's also possible how it maintains references internally might also have changed in 8.1.

I just updated the test to a simple case to for recursion.

With this pretty much all PHP 8.1 related issues are sorted out. I am not sure if the remaining mocks related issues needs to be fixed in PHPUnit or our tests need to be modified.

PR #15590, opened 2021-06-16
fix #15588: eliminate the sprintf error on asserting command output

### LINKED ISSUE #15588: Command test fails with error "sprintf(): Too few arguments" when asserting contents that have a format specifier (e.g. %s)
This is a **bug**.

* CakePHP Version: 4.2.6.
* Platform and Target: PHP 7.4 (`sprintf` still returns `false` when fails)

### What you did
Executed a phpunit test like: `vendor/bin/phpunit tests/TestCase/Command/SomethingCommandTest.php`.

The command outputs json formatted text containing format specifiers such as "%s".

### What happened
The test failed  with error: 
```
1) App\Test\TestCase\Command\SomethingCommandTest::testExecute
TypeError: Return value of Cake\TestSuite\Constraint\Console\ContentsContain::toString() must be of the type string, boolean returned
```

The essential problem is that the test throws `sprintf(): Too few arguments` because my `$this->contents` has format specifiers such as "%s".
```php
// Cake\TestSuite\Constraint\Console\ContentsContain::toString()

    public function toString(): string
    {
        return sprintf('is in %s,' . PHP_EOL . 'actual result:' . PHP_EOL . $this->contents, $this->output);
    }
```

Eliminate the error by moving the sprintf's closing bracket like this:
```php
        return sprintf('is in %s,' . PHP_EOL . 'actual result:' . PHP_EOL, $this->output) . $this->contents;
```

fugaco:

### DIFF
--- a/src/TestSuite/Constraint/Console/ContentsContain.php
+++ b/src/TestSuite/Constraint/Console/ContentsContain.php
@@ -40,6 +40,6 @@ class ContentsContain extends ContentsBase
      */
     public function toString(): string
     {
-        return sprintf('is in %s,' . PHP_EOL . 'actual result:' . PHP_EOL . $this->contents, $this->output);
+        return sprintf('is in %s,' . PHP_EOL . 'actual result:' . PHP_EOL, $this->output) . $this->contents;
     }
 }
--- a/tests/TestCase/TestSuite/ConsoleIntegrationTestTraitTest.php
+++ b/tests/TestCase/TestSuite/ConsoleIntegrationTestTraitTest.php
@@ -93,6 +93,19 @@ class ConsoleIntegrationTestTraitTest extends TestCase
     }
 
     /**
+     * tests that exec with a format specifier
+     *
+     * @return void
+     */
+    public function testExecCommandWithFormatSpecifier()
+    {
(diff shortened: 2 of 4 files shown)


src/TestSuite/Constraint/Console/ContentsContain.php
```diff
@@ -40,6 +40,6 @@ public function matches($other): bool
      */
     public function toString(): string
     {
-        return sprintf('is in %s,' . PHP_EOL . 'actual result:' . PHP_EOL . $this->contents, $this->output);
+        return sprintf('is in %s,' . PHP_EOL . 'actual result:' . PHP_EOL, $this->output) . $this->contents;
```
othercorey, line 43: Can we encode the newline with double quotes? `"is in %s,\n..."` ?

dereuromark: That aint better, given that the constant is Platform awake imo.

othercorey: If it needs the windows-specific end lines, then no worries.

markstory:
Thank you 👍

PR #15591, opened 2021-06-16
cake 5: Require php 8.0

othercorey:
This updates the basic configuration to require PHP 8.0.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -22,11 +22,11 @@
         }
     ],
     "require": {
-        "php": ">=7.2.0",
+        "php": ">=8.0",
         "ext-intl": "*",
         "ext-json": "*",
         "ext-mbstring": "*",
-        "cakephp/chronos": "^2.0",
+        "cakephp/chronos": "^2.2",
         "composer/ca-bundle": "^1.2",
         "laminas/laminas-diactoros": "^2.2.2",
         "laminas/laminas-httphandlerrunner": "^1.1",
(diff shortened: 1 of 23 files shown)


othercorey:
@ADmad @markstory @dereuromark @garas not sure if anyone else would like to know.

composer.json
```diff
@@ -22,11 +22,11 @@
         }
     ],
     "require": {
-        "php": ">=7.2.0",
+        "php": ">=8.0",
         "ext-intl": "*",
         "ext-json": "*",
         "ext-mbstring": "*",
-        "cakephp/chronos": "^2.0",
+        "cakephp/chronos": "^2.1.2",
```
ADmad, line 29: Should use 2.next for now to avoid errors on the PHP 8.1 build.

othercorey: Are you releasing 2.2. soon?

ADmad: Me personally no. I am no idea what else has gone into Chronos' next minor or what more needs to go. I just fixed the errors on PHP 8.1.

markstory:
Based on our current usage data going to 8.0 would drop a significant number of users. Perhaps that will change by the time we get around to releasing though. Historically ambitious release targets have worked out well.

othercorey:
With 8.0, we get the `mixed` and `static` types as well as compound types `string|int`. IMO, those are the key benefits for moving to 8.0 as they're more compatible with changes made in 8.1+.

There are other future use features like attributes and named arguments. There are some syntax improvements like nullsafe `?->` and `Stringable`.

othercorey:
With no objections from the active members, I'm going to merge this.

PR #15594, opened 2021-06-17
RFC: Raise PHP min to 7.3

dereuromark:
I would like to propose the min PHP for 4.3+ to be 7.3
This simplifies a lot, including the PHPUnit part (9, and soon 10).

Also, I think with 8.1 soon out, there is no need to stick to a long EOL PHP 7.2 (End of 2020 even security fixes)
https://www.php.net/supported-versions.php

Given that CakePHP 5.0 ~~could~~ will be 8.0+, a 7.3+ for upcoming new minor 4.3 sounds reasonably conservative to me.

In other words: There is currently no reason to not run 7.4 even on all systems for many months now, and therefore even 7.3 is quite super-conservative in a way :)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,16 +17,16 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        php-version: ['7.2', '8.0']
+        php-version: ['7.3', '8.0']
         db-type: [sqlite, mysql, pgsql]
         prefer-lowest: ['']
         exclude:
-          - php-version: '7.2'
+          - php-version: '7.3'
             db-type: 'mysql'
         include:
           - php-version: '7.4'
             db-type: 'mariadb'
-          - php-version: '7.2'
+          - php-version: '7.3'
             db-type: 'mysql'
             prefer-lowest: 'prefer-lowest'
           - php-version: '8.1'
@@ -44,19 +44,19 @@ jobs:
 
     steps:
     - name: Setup MySQL latest
-      if: matrix.db-type == 'mysql' && matrix.php-version == '7.2'
+      if: matrix.db-type == 'mysql' && matrix.php-version == '7.3'
       run: docker run --rm --name=mysqld -e MYSQL_ROOT_PASSWORD=root -e MYSQL_DATABASE=cakephp -p 3306:3306 -d mysql --default-authentication-plugin=mysql_native_password --disable-log-bin
 
     - name: Setup MySQL 5.6
-      if: matrix.db-type == 'mysql' && matrix.php-version != '7.2'
+      if: matrix.db-type == 'mysql' && matrix.php-version != '7.3'
       run: docker run --rm --name=mysqld -e MYSQL_ROOT_PASSWORD=root -e MYSQL_DATABASE=cakephp -p 3306:3306 -d mysql:5.6 --character-set-server=utf8
 
     - name: Setup PostgreSQL latest
-      if: matrix.db-type == 'pgsql' && matrix.php-version == '7.2'
+      if: matrix.db-type == 'pgsql' && matrix.php-version == '7.3'
       run: docker run --rm --name=postgres -e POSTGRES_PASSWORD=postgres -e POSTGRES_DB=cakephp -p 5432:5432 -d postgres
 
     - name: Setup PostgreSQL 9.4
-      if: matrix.db-type == 'pgsql' && matrix.php-version != '7.2'
+      if: matrix.db-type == 'pgsql' && matrix.php-version != '7.3'
       run: docker run --rm --name=postgres -e POSTGRES_PASSWORD=postgres -e POSTGRES_DB=cakephp -p 5432:5432 -d postgres:9.4
 
     - uses: getong/mariadb-action@v1.1
@@ -105,7 +105,7 @@ jobs:
         fi
 
     - name: Setup problem matchers for PHPUnit
-      if: matrix.php-version == '7.2' && matrix.db-type == 'mysql'
+      if: matrix.php-version == '7.3' && matrix.db-type == 'mysql'
       run: echo "::add-matcher::${{ runner.tool_cache }}/phpunit.json"
 
     - name: Run PHPUnit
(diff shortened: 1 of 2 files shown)


markstory:
Currently ~20% of CakePHP 4 usage is from 7.2 based on the install numbers [packagist publishes](https://packagist.org/packages/cakephp/cakephp/php-stats#4). We'd be forcing those folks to stay on 4.2 or potentially upgrade operating systems which can be a big effort.

Outside of simpler PHPUnit versions is there any other benefits to dropping 7.2 at this time?

dereuromark:
Those ~20% could very well be 99% CI 7.2 matrix checks :) For our hundreds of plugins and alike.
Just saying.

I don't see a good reason here to stay on an EOL for a new minor when the upgrade path is so simple and trivial actually.

As for other benefits:
- Avoiding potential problems with dependencies going 7.3+ as well (this creates a mess of testing and ensuring BC as well)
- A few hacks can be removed

ADmad:
> I don't see a good reason here to stay on an EOL for a new minor when the upgrade path is so simple and trivial actually.

Unfortunately the decision regarding upgrades is in the the hands of the "business people" and not the devs and they don't want to upgrade unless it brings additional direct value to the business (even if upgrades are simple).

> Avoiding potential problems with dependencies going 7.3+ as well (this creates a mess of testing and ensuring BC as well)

Even if a dependencies release a new 7.3+ version the current version we use will continue to work. For e.g. `laminas-diactoros` already bumped PHP to 7.3+ (when I submitted a PR for PHP 8 support), yet it hasn't caused us any issue. Those using PHP 7.2 get a older minor of diactoros while those with PHP 7.3+ get the latest.

> A few hacks can be removed

Which ones? A few version checks we might have in the code aren't really a problem.

dereuromark:
https://www.reddit.com/r/PHP/comments/obnqh9/php_security_release_808_7421_and_7329/
I do not think it is feasible nor responsible to try to justify 7.2 usage or support when the alternatives are so simple.
Again: The upgrade efforts are minimal and super easy and at this point every script kiddie can even use php7.4+ already on any server, let alone all actual companies and businesses using CakePHP more professionally.

At the same time, we soon force them to use 8+ anyway, here we could also otherwise argue it needs to still be on 7.2+.
There we don't, even across a major. So here, with a simple minor we also do not need to IMO.

The upping can be communicated clearly and in advance, and no one is required to use 4.3+ from the start.

markstory:
> Again: The upgrade efforts are minimal and super easy and at this point every script kiddie can even use php7.4+ already on any server, let alone all actual companies and businesses using CakePHP more professionally.

I agree that the upgrade is simple for new applications. For existing applications upgrading PHP will often entail an operating system upgrade which isn't a low-effort change.

> At the same time, we soon force them to use 8+ anyway, here we could also otherwise argue it needs to still be on 7.2+.

The timeline on 5.0 is a bit farther out than 4.3 is though.

dereuromark:
Also refs https://github.com/cakephp/docs/pull/7004/

Erwane:
Today, 7.2 + 7.3 is 20%
Even wordpress raise warning for php version less than 7.3, maybe 7.4.

I understand concept to allow max peoples to use CakePHP but allowing "not supported" versions of PHP will not "force" developers and hosters to update their packages.

7.2 is no more supported and 7.3 will have security risk soon (https://www.php.net/supported-versions.php) and i think everyone should upgrade to 7.4, plus, 7.2 to 7.4 is really not risky. Need check, but not risky.

dereuromark:
> 7.2 to 7.4 is really not risky. Need check, but not risky.

Even " 7.3 will have security risk soon", jep

We do need to make a decision here.
But given that 4.x is supposed to be LTS (with PHP7), the only sane approach IMO is
- Cake 4.3 is PHP 7.3+ (7.3 already EOL and a security risk in **less than 2 months** from now)
- Cake 4.4 is PHP 7.4+ (maybe, as 7.4 is EOL in 1 year, 2 months)
- Cake 5 is already PHP 8+

7.3 is already way more conservative given all the info above.

markstory:
In light of PHP support coming to an end soon on 7.3, going to 7.3 for 4.3 sounds good to me.

Erwane:
>     * Cake 4.3 is PHP 7.3+ (7.3 already EOL and a security risk in **less than 2 months** from now)
>     * Cake 4.4 is PHP 7.4+ (maybe, as 7.4 is EOL in 1 year, 2 months)
>     * Cake 5 is already PHP 8+

Sounds really good

othercorey:
I would caution against being too aggressive about dropping support for little gain.

ADmad:
> I would caution against being too aggressive about dropping support for little gain.

I agree, as framework maintainers we don't have much to gain. I doubt we are going to add new features to 4.x which can benefit from PHP 7.3/7.4 features.

Even after a PHP version reaches EOL major distros continue to provide security fixes for the PHP versions supported by their LTS releases.
For e.g. RHEL https://access.redhat.com/security/updates/backporting/

dereuromark:
> I would caution against being too aggressive about dropping support for little gain.

Yes of course
Thats why 7.3 is proposed while 7.4 and 8.0 are already well established for a long time.

markstory:
Are there distros that are below 7.3? I know that Debian and downstreams like Ubuntu are on 7.4 in their latest LTS. Fedora also seems to be at 7.4 as of Fedora 32.

dereuromark:
With 7.3 also being EOL (no more security fixes even) in less than two months, is >= 7.3 now OK to be merged as requirement?
The sane and safe one would be >= 7.4, but I guess 7.3 is a good compromise then.

othercorey:
I still don't see the benefit, but I leave it up to you.

dereuromark:
Not doing it will be a recipe for disaster or failure in the long run. Especially with 4.3 having to be LTS to some extend.
There will be PHP 8.2+ coming with changes that also could bind to PHPUnit 10+ and other dependencies in a way, that having that large span of PHP versions to maintain will make newer dependencies impossible to meet - and all that for no good reason as we discussed here. Some libraries are already 7.4+, or even 8.0+, leaving also older versions already behind for any fixes.

markstory:
I'm ok with bumping up to 7.3 given it is rapidly approaching end of life.

markstory:
I've been thinking about this more, and I don't feel :100: great about bumping the minimum PHP version so close to the release of 4.3.0. It feels like something that will catch users and plugin maintainers by surprise as we've not communicated the change at all. I'd like to avoid delaying the release of 4.3 longer, and suggest that in `4.4` we raise the minimum PHP version to `7.4.0`. This will give us more time to communicate the change and let community plugins get prepared.

Erwane:
upgrading to 7.4.0 could prepare old projects to properties types and transition to 8.x if it's decided to type all variables and properties.

If this case, waiting for 4.4 and communication is a good idea.

othercorey:
> upgrading to 7.4.0 could prepare old projects to properties types and transition to 8.x if it's decided to type all variables and properties.

We added type declarations for properties, parameters and returns in 5.x, but we won't change the API in 4.4.

othercorey:
Closing this as we're moving the requirement to 4.4.

PR #15595, opened 2021-06-18
cake 5: Require Event payload to be array or ArrayAccess

othercorey:
Refs https://github.com/cakephp/cakephp/issues/14482

Minimal change to fix issue. Would like to update interface to use payload instead of data.

### DIFF
--- a/src/Event/Event.php
+++ b/src/Event/Event.php
@@ -16,6 +16,7 @@ declare(strict_types=1);
  */
 namespace Cake\Event;
 
+use ArrayAccess;
 use Cake\Core\Exception\CakeException;
 
 /**
@@ -43,9 +44,9 @@ class Event implements EventInterface
     /**
      * Custom data for the method that receives the event
      *
-     * @var array
+     * @var \ArrayAccess|array
      */
-    protected $_data;
+    protected $payload;
 
     /**
      * Property used to retain the result value of the event listeners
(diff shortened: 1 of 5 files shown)


othercorey:
These template errors are annoying. Maybe can just remove them.

ADmad:
I never liked the fact that the event payload can only be an array/ArrayAccess. For e.g. if my payload is a request instance I have to unnecessarily wrap it in an array.

othercorey:
> I never liked the fact that the event payload can only be an array/ArrayAccess. For e.g. if my payload is a request instance I have to unnecessarily wrap it in an array.

I was thinking that as well. We should be able to attach anything. It does require cleaning up the payload getter/setter.

ADmad:
While I don't see much practical value, we should maybe make our event package [PSR-14](https://www.php-fig.org/psr/psr-14/) compliant too.

markstory:
How would `getData()` work if the payload is an arbitrary type? Would you raise an exception if the type was not array/arrayaccess?

othercorey:
We would not provide key access.

ADmad:
We could deprecate `getData()/setData()` in favour of `getPayload()/setPayload()` and the `*Data()` methods would throw and exception if the payload isn't an array/ArrayAccess.

othercorey:
> We could deprecate `getData()/setData()` in favour of `getPayload()/setPayload()` and the `*Data()` methods would throw an exception if the payload isn't an array/ArrayAccess.

If we deprecate in 4.3, then I would just deprecate `getData()` and `setData()`. When they switch to `getPayload()` and setPayload()`, there is only one solution so we don't need to validate the types.

ADmad:
> However, we could throw an exception if more than 1 parameter is passed to setPayload().

I don't think that's necessary. That's IDEs / static analysers' job :).

othercorey:
https://book.cakephp.org/4/en/core-libraries/events.html#getting-event-data-as-function-parameters

The event manager actually splits the array values into listener arguments. This seems rather awkward since the array can be associated, but it just does:
```php
...array_values((array)$data));
````

Issue #15528, opened 2021-05-20
RFC: Ordering union types


othercorey:
I thought I'd make an RFC issue to track this idea for moving to php 8 union types.

If we decide to convert most types to union types in cakephp 5, one of the solutions is copy the types from the existing docblocks. There are limitations to this approach since php does not support typed arrays - only the `array` type.

Union types in cakephp should follow a specific pattern to make it easier to read and update the type. It makes sense that we will copy from docblock types since we know they are correct.

I'm suggesting the following format:

```
{classes\interfaces, unsorted}|\Closure|\Traversable|\ArrayAccess|callable|resource|object|iterable|array|string|float|int|bool|true|false|null|void
```

An example would be:

`\Cake\I18n\FrozenDate|\DateTime|\Callable|string[]|string|null|false`

markstory:
I'm good with choosing an order if we can enforce it and autocorrect with linters. Without that it adds toil to review and maintenance.

dereuromark:
The order looks good to me, expect for the null/false part at the end, null should always be last IMO, as it already is for docblock types (true|false|null):

    ...|float|false|null

othercorey:
Yea, it would only make sense with a sniff. However, if we feel that is burdensome, we could enable it to fix the docblocks once.

`false|null` is ok with me if that's desired (it's probably very rare). I was trying to find php manual examples of multiple scalar return types to see if there is any conflicting ordering.

PR #15596, opened 2021-06-18
Normalize order of union types in docblocks

othercorey:
Refs https://github.com/cakephp/cakephp/issues/15528

This handles many non-object types.


ADmad:
Without a phpcs sniff we won't be ensure the ordering is followed for new/updated methods.

othercorey:
> Without a phpcs sniff we won't be able to ensure the ordering is followed for new/updated methods.

That sniff will take some work. This just cleans up a large chunk so the diffs aren't terrible in 5.9.

othercorey:
There are a few docblocks where the object type is last still. I will try to fix them.

ADmad:
On a related noted I am also wondering if we should relax our docblock requirements for 5.0. Repeating the types in docblock when we have language types is redundant. The description like `Response instance` are also pretty useless when the argument is typed as `ServerResponseInterface`.

markstory:
> On a related noted I am also wondering if we should relax our docblock requirements for 5.0. Repeating the types in docblock when we have support for types in the language itself now is redundant (unless we want to expand them in docblocks). The description like Request instance are also pretty useless when the argument is typed as ServerResponseInterface.

💯 agree. The type annotations in docs are generally redundant and it would be good to not have to type everything out twice. The description text is also used in the generated API docs so it isn't always noise.

dereuromark:
I usually only add a text/description if it provides value beyond what is already written :) Repeating it was never useful.

othercorey:
> The description text is also used in the generated API docs so it isn't always noise.

Our parameter descriptions are often terrible, but what would we show if they're missing?

dereuromark:
> Our parameter descriptions are often terrible, but what would we show if they're missing?

Could it then just auto-inflect the name from the param etc? 

    $paramName => Param Name

othercorey:
The API docs do default to the reflected types. However, it will generate a slightly awkward empty parameter description block. Maybe we optimize how those are displayed.

PR #15600, opened 2021-06-19
Update return types.

ADmad:
Remove no longer needed `ReturnTypeWillChange` attribute.


ADmad:
PHP 8.1 remains happy :tada:

PR #15601, opened 2021-06-19
Normalize order of types in return tags

othercorey:
`\Cake\Datasource\EntityInterface|$this` is a little strange.

### DIFF
--- a/src/Database/Expression/FieldTrait.php
+++ b/src/Database/Expression/FieldTrait.php
@@ -42,7 +42,7 @@ trait FieldTrait
     /**
      * Returns the field name
      *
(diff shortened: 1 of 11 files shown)


src/Database/Expression/FieldTrait.php
```diff
@@ -42,7 +42,7 @@ public function setField($field): void
     /**
      * Returns the field name
      *
-     * @return array|string|\Cake\Database\ExpressionInterface
+     * @return \Cake\Database\ExpressionInterfacearray|string
```
othercorey, line 45:
```suggestion
     * @return \Cake\Database\ExpressionInterface|array|string
```

PR #15603, opened 2021-06-19
cake 5: Remove validation deprecations

ADmad:

### DIFF
--- a/src/Validation/Validator.php
+++ b/src/Validation/Validator.php
@@ -218,21 +218,6 @@ class Validator implements ArrayAccess, IteratorAggregate, Countable
      * @param array $data The data to be checked for errors
      * @param bool $newRecord whether the data to be validated is new or to be updated.
      * @return array[] Array of failed fields
-     * @deprecated 3.9.0 Renamed to {@link validate()}.
-     */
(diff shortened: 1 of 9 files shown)


src/Validation/Validator.php
```diff
@@ -212,21 +212,6 @@ public function setStopOnFailure(bool $stopOnFailure = true)
...
 
-    /**
-     * Validates and returns an array of failed fields and their error messages.
-     *
-     * @param array $data The data to be checked for errors
-     * @param bool $newRecord whether the data to be validated is new or to be updated.
-     * @return array[] Array of failed fields
-     * @deprecated 3.9.0 Renamed to {@link validate()}.
-     */
-    public function errors(array $data, bool $newRecord = true): array
```
markstory, line 223: Do we want to document each of these removals separately or have a more generalized statement about applications needing to have 0 deprecations before they are ready to upgrade?

othercorey: When I setup the 5.0 docs, that's what 4.0 did. I assumed it would be the same.

ADmad: We already have explicit deprecation warnings for all(most) deprecations, so a general statement should suffice.

PR #15604, opened 2021-06-19
cake 5: Remove http-related deprecated code

othercorey:


tests/TestCase/TestSuite/TestEmailTransportTest.php
```diff
@@ -95,9 +95,9 @@ public function testSendGetAndClear()
     {
         TestEmailTransport::replaceAllTransports();
 
-        (new Email())
+        (new Mailer())
             ->setTo('test@example.com')
-            ->send('test');
+            ->deliver('test');
```
othercorey, line 100: @markstory Can you double check my changes here?

markstory: This looks right to me. Mailer::deliver() does the same thing that Email::send() did. The other changes in this pull request look good too. Lots of deprecation cleanup.

PR #15605, opened 2021-06-20
cake 5: Remove various deprecations

othercorey:
This removes all remaining deprecations except for Shell, File, Folder and App::path().


othercorey:
Fixed the phpstan Email.php warning in 5.0 branch.

src/Controller/Controller.php
```diff
@@ -694,29 +694,6 @@ public function redirect($url, int $status = 302): ?Response
...
-     * @param mixed ...$args Arguments passed to the action
-     * @return mixed Returns the return value of the called action
-     * @deprecated 4.2.0 Refactor your code use `redirect()` instead of forwarding actions.
-     */
-    public function setAction(string $action, ...$args)
-    {
-        $this->setRequest($this->request->withParam('action', $action));
-
-        return $this->$action(...$args);
-    }
```
saeideng, line 718: better to have `deprecationWarning()` in cake 4.x before removing this method, IMO

src/Controller/Controller.php
```diff
@@ -694,29 +694,6 @@ public function redirect($url, int $status = 302): ?Response
...
-     * ```
-     *
-     * @param string $action The new action to be 'redirected' to.
-     *   Any other parameters passed to this method will be passed as parameters to the new action.
-     * @param mixed ...$args Arguments passed to the action
-     * @return mixed Returns the return value of the called action
-     * @deprecated 4.2.0 Refactor your code use `redirect()` instead of forwarding actions.
-     */
-    public function setAction(string $action, ...$args)
-    {
```
saeideng, line 714: better to have deprecationWarning() in cake 4.x before removing this method, IMO

othercorey: You can add to 4.3 before release if you want.

PR #15610, opened 2021-06-21
cake 5: Use `null` instead of `false` to disable log scopes.

### LINKED ISSUE #15544: Change return type of `BaseLog::scopes()`
The current return type of `BaseLog::scopes()` is `array|false` in docblock which should be changed to either `array` or `array|null`. All emptish values for scopes are treated the same anyway by the loggers.

Currently our app skeleton uses `'scopes' => false` in config arrays so best to do the suggested change in next major.

ADmad:
Closes #15544.

### DIFF
--- a/src/Log/Engine/BaseLog.php
+++ b/src/Log/Engine/BaseLog.php
@@ -57,13 +57,16 @@ abstract class BaseLog extends AbstractLogger
     {
         $this->setConfig($config);
 
-        if (!is_array($this->_config['scopes']) && $this->_config['scopes'] !== false) {
+        assert(
+            $this->_config['scopes'] !== false,
+            new InvalidArgumentException('Use `null` instead of `false` to disable scopes.')
+        );
+
+        if ($this->_config['scopes'] !== null) {
             $this->_config['scopes'] = (array)$this->_config['scopes'];
         }
 
-        if (!is_array($this->_config['levels'])) {
-            $this->_config['levels'] = (array)$this->_config['levels'];
-        }
+        $this->_config['levels'] = (array)$this->_config['levels'];
 
         if (!empty($this->_config['types']) && empty($this->_config['levels'])) {
             $this->_config['levels'] = (array)$this->_config['types'];
@@ -105,9 +108,9 @@ abstract class BaseLog extends AbstractLogger
     /**
      * Get the scopes this logger is interested in.
      *
-     * @return array|false
+     * @return array|null
      */
-    public function scopes()
+    public function scopes(): ?array
     {
         return $this->_config['scopes'];
     }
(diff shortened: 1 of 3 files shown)


tests/TestCase/Log/LogTest.php
```diff
@@ -471,7 +471,7 @@ public function testScopedLoggingStrict()
             'path' => LOGS,
             'levels' => ['notice', 'info', 'debug'],
             'file' => 'debug',
-            'scopes' => false,
+            'scopes' => null,
```
othercorey, line 474: Why wouldn't this just be an empty array?

ADmad: Because empty array means capture all scopes :stuck_out_tongue_closed_eyes: .

src/Log/Engine/BaseLog.php
```diff
@@ -57,13 +57,16 @@ public function __construct(array $config = [])
     {
         $this->setConfig($config);
 
-        if (!is_array($this->_config['scopes']) && $this->_config['scopes'] !== false) {
+        assert(
```
markstory, line 60: Are you thinking that we should use `assert()` more for developer error checking? Would we also start recommending that people use `assert_options()` to optimize away assertion checks in production?

othercorey: I'm not sure we should use assertions where cake can't continue. If the data is invalid, then we should probably just throw the exception which is guaranteed.

ADmad: > Are you thinking that we should use assert() more for developer error checking? 

Yup, so that the checks/safeguards have zero impact in production. 

> Would we also start recommending that people use assert_options() to optimize away assertion checks in production?

The manual actually recommends using the ini config `zend.assertions` instead of `assert_options()` in PHP 7+ code.

> I'm not sure we should use assertions where cake can't continue. If the data is invalid, then we should probably just throw the exception which is guaranteed.

IMO `assert()` is the right thing to use in this case. The check is only there to avoid missing updating the config when upgrading from 4.x.

tests/TestCase/Log/LogTest.php
othercorey: Of course.

src/Log/Engine/BaseLog.php
othercorey: Sounds good. We could probably check !is_bool() to be comprehensive.

othercorey:
@ADmad This will need a migration note in the docs. I setup 5.0 migration notes here: https://github.com/cakephp/docs/blob/5.0/en/appendices/5-0-migration-guide.rst

ADmad:
And updating the app skeleton, will do both soon.

PR #15611, opened 2021-06-21
cake 5: Require array event payload

othercorey:
This fixes the original issue until a new Event design is completed.

Casting ArrayAccess to array (such as entities) does not work well.


markstory (approved):
This looks good to me. Having an array is a simpleish design that solves some of the problems without creating more which is nice.

PR #15613, opened 2021-06-21
cake 5: Remove deprecated syslog formatter

othercorey:


othercorey:
@markstory fyi

markstory:
Looks good. This seems like something we should mention in the upgrade guide as it is a behavior change but not a breaking API change.

othercorey:
Separate from the deprecation notice?

markstory:
Derp. I forgot about the deprecation notice.

PR #15614, opened 2021-06-21
4.3 - Consolidate fixture namespaces

markstory:
Having Schema and Fixture be separate felt off as both namespaces are for fixtures. This will require an update in cakephp/app as well.


markstory:
I've managed to make CI sad. I'm able to reproduce locally and hope to have a solution soon.

PR #15617, opened 2021-06-22
Align return type of fetchAll()

othercorey:
Fix the return type mismatches thrown by php 8.1.


othercorey:
This isn't the actual issue. There is a bug in php 8.1 that I reported.

Still, don't see why this shouldn't behave correctly.

PR #15622, opened 2021-06-24
cake 5: Add string type hint to Driver::quote() value parameter

othercorey:
This matches the type hints for PDO::quote(). https://www.php.net/manual/en/pdo.quote.php

### DIFF
--- a/src/Database/Driver.php
+++ b/src/Database/Driver.php
@@ -289,11 +289,11 @@ abstract class Driver implements DriverInterface
     /**
      * @inheritDoc
      */
-    public function quote($value, $type = PDO::PARAM_STR): string
+    public function quote(string $value, int $type = PDO::PARAM_STR): string
     {
         $this->connect();
 
-        return $this->_connection->quote((string)$value, $type);
+        return $this->_connection->quote($value, $type);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


othercorey:
https://github.com/phpstan/phpstan-src/pull/564

src/Database/Connection.php
```diff
@@ -756,6 +757,9 @@ public function inTransaction(): bool
     public function quote($value, $type = 'string'): string
     {
         [$value, $type] = $this->cast($value, $type);
+        if (!is_string($value)) {
+            throw new DatabaseException('Unable to quote non-string values.');
```
othercorey, line 761: This one might be unnecessary since it should throw a php error.

src/Database/Driver.php
```diff
@@ -289,11 +289,11 @@ public function supportsCTEs(): bool
     /**
      * @inheritDoc
      */
-    public function quote($value, $type = PDO::PARAM_STR): string
+    public function quote(string $value, int $type = PDO::PARAM_STR): string
     {
         $this->connect();
 
-        return $this->_connection->quote((string)$value, $type);
+        return $this->_connection->quote($value, $type);
```
othercorey, line 296: This no longer casts to string. I'm not sure we should be quoting non-string values here.

src/Database/Connection.php
markstory: We could skip it for now as PHP should be taking care of this.

PR #15627, opened 2021-06-24
Return array instead of false for empty fetchAll() mock return

othercorey:


othercorey:
Looks like this fixes the warnings on php 8.1:
https://github.com/cakephp/cakephp/pull/15627/checks?check_run_id=2907620434#step:17:207

ADmad:
Don't we need any changed in code which uses this method to check for `[]` instead of `false`?

othercorey:
> Don't we need any change in code which uses this method to check for `[]` instead of `false`?

It wasn't checking for a DB/driver error, it was just looking for empty results. The `false` return here was wrong in general.

PR #15628, opened 2021-06-24
Return PDOStatement instead of true for query mock return

othercorey:
This was never valid. The return type for `query()` does not include `true` for any php version.


othercorey:
https://github.com/cakephp/cakephp/blob/master/src/Database/Driver.php#L393

We don't need to mock any results since it just checks if the query failed or not.

PR #15630, opened 2021-06-25
Upgrade psalm and fix reported issues.

ADmad:


markstory:
Looks like a few errors coming from laminas still.

ADmad:
Merging as the CI - Stan build is green now.

PR #15633, opened 2021-06-26
Deprecate calling ResultSetInterface methods on query instances directly

othercorey:
This fixes every use of Collection methods on query instances in the tests.

### DIFF
--- a/src/Datasource/QueryTrait.php
+++ b/src/Datasource/QueryTrait.php
@@ -552,6 +552,11 @@ trait QueryTrait
     {
         $resultSetClass = $this->_decoratorClass();
         if (in_array($method, get_class_methods($resultSetClass), true)) {
+            deprecationWarning(sprintf(
+                'Calling result set method `%s()` directly on query instance is deprecated. ' .
+                'You must call `all()` to retrieve the results first.',
+                $method
+            ));
             $results = $this->all();
 
             return $results->$method(...$arguments);
--- a/tests/TestCase/ORM/QueryTest.php
+++ b/tests/TestCase/ORM/QueryTest.php
@@ -870,7 +870,7 @@ class QueryTest extends TestCase
         $query = new Query($this->connection, $article);
 
         $firstCount = $query->count();
-        $firstResults = $query->toList();
+        $firstResults = $query->toArray();
 
         $this->assertEquals(3, $firstCount);
         $this->assertCount(3, $firstResults);
@@ -881,7 +881,7 @@ class QueryTest extends TestCase
         $this->assertSame($return, $query);
 
         $secondCount = $query->count();
-        $secondResults = $query->toList();
+        $secondResults = $query->toArray();
 
         $this->assertEquals(2, $secondCount);
         $this->assertCount(2, $secondResults);
@@ -1690,6 +1690,7 @@ class QueryTest extends TestCase
 
         $query->select(['s' => $query->func()->rand()]);
         $result = $query
+            ->all()
             ->extract('s')
             ->first();
 
@@ -1799,6 +1800,7 @@ class QueryTest extends TestCase
             ['take', 1, $collection],
             ['append', new \ArrayIterator(), $collection],
             ['compile', true, $collection],
+            ['isEmpty', true, true],
         ];
     }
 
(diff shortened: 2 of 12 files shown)


tests/TestCase/ORM/QueryTest.php
```diff
@@ -1949,7 +1951,9 @@ public function testCollectionProxy($method, $arg, $return)
             ->with($arg, 99)
             ->will($this->returnValue($return));
 
-        $this->assertSame($return, $query->{$method}($arg, 99));
+        $this->deprecated(function () use ($return, $query, $method, $arg) {
+            $this->assertSame($return, $query->{$method}($arg, 99));
+        });
```
othercorey, line 1956: This tests all of the proxied methods.

src/Datasource/QueryTrait.php
```diff
@@ -552,6 +552,11 @@ public function __call(string $method, array $arguments)
     {
         $resultSetClass = $this->_decoratorClass();
         if (in_array($method, get_class_methods($resultSetClass), true)) {
+            deprecationWarning(sprintf(
+                'Unable to call `%s()` directly on query instance. ' .
```
ADmad, line 556: "Unable" sounds wrong since one is still able :)
An explicit message here might be quite annoying to users. Perhaps we should skip the message as we had done when deprecating `TableRegistry`'s method in 3.x.

othercorey: I was trying to say unable to on the query instance.

How would we deprecate it? Deprecate the __call() function?

ADmad: > How would we deprecate it? Deprecate the __call() function?

I just meant that the message needs to be tweaked.

othercorey: Suggestions on how to make it feel like TableRegistry?

ADmad: Hmm.. `TableRegistry` had concrete methods so `@deprecated` tag was enough. We could add the tag to the `__call()` method as you suggested. Not sure how IDE would deal with it but static analyzers should warn (if related rule/level is enabled).

We will also have to removed all the `@method` for methods of `CollectionInterface`.

othercorey: I think removing the `@method` annotations would make their static analysis fail though. Might be too harsh.

othercorey: I tried a couple things. Marking __call() as deprecated does not throw an error from psalm or my visual studio code extension. I'm not sure about phpstorm, but that's probably not good enough even if they do.

ADmad: > I think removing the @method annotations would make their static analysis fail though. Might be too harsh.

Having IDEs/tools report errors when using this magic method and not have any errors when actually running the code is our aim isn't?

markstory: Is the concern with a deprecation warning that it will be too noisy? Thes method calls are going to be hard to track down without warnings or help from static analyzers.

ADmad: > Is the concern with a deprecation warning that it will be too noisy?

Yup

> Thes method calls are going to be hard to track down without warnings or help from static analyzers.

Yup, I guess keeping the explicit warning is the best option then.

othercorey: Would "Do not call..." sound better?

garas:
```suggestion
                'Calling result set method `%s()` directly on query instance is deprecated. ' .
```

It should also explain why this method differs from others, i.e. it is not query, but result set/collection from here.

othercorey: The deprecation part is unnecessary in a deprecation warning I think.

ADmad: > The deprecation part is unnecessary in a deprecation warning I think.

We have had it in the various messages in the past and it read well. The message shown to users only has "Warning" prefix, so mentioning it's for deprecation is good.

othercorey: No worries. Any message you want is fine with me.

markstory:
I can take care of updating the docs for this.

PR #15639, opened 2021-07-02
cake 5: Remove deprecated cast() parameters

othercorey:
I didn't find a test for the deprecated parameters.

### DIFF
--- a/src/Database/FunctionsBuilder.php
+++ b/src/Database/FunctionsBuilder.php
@@ -133,26 +133,13 @@ class FunctionsBuilder
      * is the default type name. Use `setReturnType()` to update it.
      *
      * @param \Cake\Database\ExpressionInterface|string $field Field or expression to cast.
-     * @param string $type The SQL data type
+     * @param string $dataType The SQL data type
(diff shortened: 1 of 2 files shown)


src/Database/FunctionsBuilder.php
```diff
@@ -133,26 +133,13 @@ public function coalesce(array $args, array $types = []): FunctionExpression
      * is the default type name. Use `setReturnType()` to update it.
      *
      * @param \Cake\Database\ExpressionInterface|string $field Field or expression to cast.
-     * @param string $type The SQL data type
+     * @param string $sqlType The SQL data type
      * @return \Cake\Database\Expression\FunctionExpression
      */
-    public function cast($field, string $type = ''): FunctionExpression
+    public function cast($field, string $sqlType): FunctionExpression
```
dereuromark, line 139: Or
```suggestion
    public function cast($field, string $dataType): FunctionExpression
```

othercorey: We definitely need better terms for SQL vs cake types. I want to make it clear it's not the same as `$types` from other methods, at least.

PR #15640, opened 2021-07-02
cake 5: Remove only E_USER_DEPRECATED flag when hiding deprecations

othercorey:
I don't think we should reset to E_ALL inside a `deprecated()` wrapper.

### DIFF
--- a/src/TestSuite/TestCase.php
+++ b/src/TestSuite/TestCase.php
@@ -210,7 +210,7 @@ abstract class TestCase extends BaseTestCase
     public function deprecated(callable $callable): void
     {
         $errorLevel = error_reporting();
-        error_reporting(E_ALL ^ E_USER_DEPRECATED);
+        error_reporting($errorLevel & ~E_USER_DEPRECATED);
         try {
             $callable();
         } finally {


src/TestSuite/TestCase.php
```diff
@@ -210,7 +210,7 @@ public function withErrorReporting(int $errorLevel, callable $callable): void
     public function deprecated(callable $callable): void
     {
         $errorLevel = error_reporting();
-        error_reporting(E_ALL ^ E_USER_DEPRECATED);
+        error_reporting($errorLevel ^ E_USER_DEPRECATED);
```
dereuromark, line 213: This sounds more like a 4.x fix to me even, actually.

othercorey: I can backport if approved.

dereuromark: I would just fix it for 4.x as the upmerge will automatically solve it for 5.0 then afterwards.

garas:
```suggestion
        error_reporting($errorLevel & ~E_USER_DEPRECATED);
```

Current level might not include E_USER_DEPRECATED

othercorey: I wonder if we should throw a warning if E_USER_DEPRECATED is not set. Seems like a bug in the test.

garas: It could also check with custom error handler if E_USER_DEPRECATED was thrown at least once in callback so it would reveal when callback use any deprecated code.

othercorey: I thought about that, but it can cause an error in the test if an exception is thrown.

PR #15642, opened 2021-07-02
Deprecate colon prefixed route template placeholders.

ADmad:
Braced placeholders should be used instead.

### DIFF
--- a/src/Routing/Route/Route.php
+++ b/src/Routing/Route/Route.php
@@ -108,6 +108,11 @@ class Route
     public const VALID_METHODS = ['GET', 'PUT', 'POST', 'PATCH', 'DELETE', 'OPTIONS', 'HEAD'];
 
     /**
+     * Regex for matching braced placholders in route template.
+     */
+    protected const PLACEHOLDER_REGEX = '#\{([a-z][a-z0-9-_]*)\}#i';
+
+    /**
      * Constructor for a Route
      *
      * ### Options
@@ -317,10 +322,16 @@ class Route
         $parsed = preg_quote($this->template, '#');
 
         if (strpos($route, '{') !== false && strpos($route, '}') !== false) {
-            preg_match_all('/\{([a-z][a-z0-9-_]*)\}/i', $route, $namedElements);
+            preg_match_all(static::PLACEHOLDER_REGEX, $route, $namedElements);
         } else {
-            preg_match_all('/:([a-z0-9-_]+(?<![-_]))/i', $route, $namedElements);
+            $hasMatches = preg_match_all('/:([a-z0-9-_]+(?<![-_]))/i', $route, $namedElements);
             $this->braceKeys = false;
+            if ($hasMatches) {
+                deprecationWarning(
+                    'Colon prefixed route placeholders like `:foo` are deprecated.'
+                    . ' Use braced placeholders like `{foo}` instead.'
+                );
+            }
         }
         foreach ($namedElements[1] as $i => $name) {
             $search = preg_quote($namedElements[0][$i]);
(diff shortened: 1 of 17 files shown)


src/Routing/Route/Route.php
```diff
@@ -852,12 +858,26 @@ public function staticPath(): string
     {
         $routeKey = strpos($this->template, '{');
         if ($routeKey !== false && strpos($this->template, '}') !== false) {
-            return substr($this->template, 0, $routeKey);
+            $parts = explode('/', ltrim($this->template, '/'));
```
ADmad, line 861: This additional work was required to support templates like `'/api/{/{action}/*'`.

Someone actually using such a template seems quite unlikely but our test suite has a test for `'/api/{/:action/*'`, it needed to be supported with the braced placeholder too.

src/Routing/Route/Route.php
```diff
@@ -319,8 +319,14 @@ protected function _writeRoute(): void
         if (strpos($route, '{') !== false && strpos($route, '}') !== false) {
             preg_match_all('/\{([a-z][a-z0-9-_]*)\}/i', $route, $namedElements);
         } else {
-            preg_match_all('/:([a-z0-9-_]+(?<![-_]))/i', $route, $namedElements);
+            $hasMatches = preg_match_all('/:([a-z0-9-_]+(?<![-_]))/i', $route, $namedElements);
             $this->braceKeys = false;
+            if ($hasMatches) {
+                deprecationWarning(
+                    'Colon prefixed route placeholders are deprecated.'
```
othercorey, line 326: Let's add a matching example of colon prefixed for foo.

tests/TestCase/Routing/Route/RouteTest.php
```diff
@@ -39,16 +39,24 @@ public function setUp(): void
         Configure::write('Routing', ['prefixes' => []]);
     }
 
+    public function testDeprecationWarning()
```
othercorey, line 42:
```suggestion
    public function testDeprecatedPlaceholders()
```

tests/TestCase/Routing/Route/RouteTest.php
```diff
@@ -39,16 +39,24 @@ public function setUp(): void
...
+    {
+        $this->deprecated(function () {
+            $route = new Route('/:controller');
```
othercorey, line 45:
```suggestion
            $route = new Route('/:controller/:action/:id');
```

tests/TestCase/Routing/Route/RouteTest.php
ADmad: Don't be so pedantic :stuck_out_tongue:.

src/Routing/Route/Route.php
markstory: This looks like a sizable amount of code for something that could lead to ambiguous results. We could require that any `{` in URL templates require an escape character like `\{` to help clarify where the route parameters are.

ADmad: Yeah that sounds reasonable.

othercorey: I remember voicing a concern about escaping unrelated `{` characters in the text/log placeholders a long time ago. Did we do the same thing there?

src/Routing/Route/Route.php
```diff
@@ -850,14 +861,22 @@ protected function _writeUrl(array $params, array $pass = [], array $query = [])
      */
     public function staticPath(): string
     {
-        $routeKey = strpos($this->template, '{');
-        if ($routeKey !== false && strpos($this->template, '}') !== false) {
-            return substr($this->template, 0, $routeKey);
+        $matched = preg_match(
```
ADmad, line 864: @markstory I replaced the array juggling with a single preg_match. This shouldn't add too much overhead.

markstory: Much simpler :clap:

PR #15643, opened 2021-07-02
cake 5: Remove deprecated File and Folder

othercorey:
This leaves only the internal Filesystem class in the Filesystem package.


dereuromark:
It might be worth mentioning that from 5.0+ the Shim plugin ( https://github.com/dereuromark/cakephp-shim/ ) will support those classes for the 5.x series to allow easier migration path coming from 4.x
I for example see them used in my apps hundreds of times, migration away from those is just too painful.
Better to just swap out namespace and done for existing code here.
Less likely the chance to introduce more regressions when doing bigger upgrades on large apps.

othercorey:
I was going to suggest users simply copy the files for migration -- would be the same with shim.

dereuromark:
Copy from where? Old 4.x code you mean? Sure, that would also work.
But I would like to keep any necessary fixes inside a global and DRY vendor place :) Not across 10+ apps or sth to each get their own fixes on those classes if needed.

ADmad:
Shouldn't we remove `composer.json`, readme etc. too? The split package would be archived once these classes are gone.

othercorey:
> Shouldn't we remove `composer.json`, readme etc. too? The split package would be archived once these classes are gone.

This doesn't remove Filesystem.

dereuromark:
If that is the only remaining file (which is `@internal`) I wonder if such a package even makes sense.
Maybe it should be moved to Utility then?

ADmad:
> This doesn't remove Filesystem.

A published package with just an internal class doesn't make sense does it?

dereuromark:
Given that Plugins like Migrations rely on and use this file, it isn't so internal after all
We could also just remove that (misleading?) tag then ;)

othercorey:
I assumed we would remove the package when Filesystem was moved. If we don't want to move it then we can just "unpackage" it.

dereuromark:
Sure, this PR itself is already fine then. The rest can be a follow up topic/PR.

markstory:
Should we move `Filesystem` class to the Utility package? That is an alias we could add in 4.3 and also remove the `internal` annotation as we do have other packages that are using this class.

PR #15647, opened 2021-07-03
Fix parameter typehints in TestApp

othercorey:
This does not fix all typehints. Want to make sure everything else still passes.

These fail the slevomat paramter typehint sniff which we want to use in 5.0.


othercorey:
Open php-docs PR to fix the stream_open() typehint. https://github.com/php/doc-en/pull/753

PR #15653, opened 2021-07-03
Deprecate DoublePassDecoratorMiddleware.

ADmad:
Support for "double pass" middleware was added just to make upgrading from 3.x easier.

### DIFF
--- a/src/Http/Middleware/DoublePassDecoratorMiddleware.php
+++ b/src/Http/Middleware/DoublePassDecoratorMiddleware.php
@@ -38,6 +38,9 @@ use Psr\Http\Server\RequestHandlerInterface;
  * or a class with `__invoke()` method with same signature as above.
  *
  * Neither the arguments nor the return value need be typehinted.
+ *
+ * @deprecated 4.3.0 "Double pass" middleware are deprecated.
+ *   Use a `Closure` or a class which implements `Psr\Http\Server\MiddlewareInterface` instead.
  */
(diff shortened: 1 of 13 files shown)


othercorey:
Is there a way to add test coverage back?

src/Http/Middleware/DoublePassDecoratorMiddleware.php
```diff
@@ -55,6 +58,10 @@ class DoublePassDecoratorMiddleware implements MiddlewareInterface
      */
     public function __construct(callable $callable)
     {
+        deprecationWarning(
+            '"Double pass" middleware are deprecated.'
+            . ' Use a `Closure` or a class which implements `Psr\Http\Server\MiddlewareInterface` instead.'
```
markstory, line 63:
```suggestion
            . ' Use a `Closure` with the signature of `($request, $handler)` or a class which implements `Psr\Http\Server\MiddlewareInterface` instead.'
```

Should we clarify the expected arity of the closure as 3 parameter middlewares are deprecated.

ADmad: I think stating the required signature for Closures should suffice. The error message would get too long otherwise. The deprecated signature is mentioned in the migration guide.

ADmad:
> Is there a way to add test coverage back?

It's back. Apparently phpunit doesn't consider the lines as "covered" if there are no assertions in the test method, even though the code is actually run.

PR #15660, opened 2021-07-05
Move fixture cleanup to teardown

markstory:
Doing cleanup in setup causes problems for subsequent test runs as data from the previous test run is left behind in the database. Doing the cleanup in teardown attempts to ensure that the database is cleared at the end of the test.

This will work less well when tests are killed before a test run is complete as state will not be cleaned up. The old fixture system solved this by dropping tables at the beginning of a test run which can happenin the new fixture system depending on how schema is loaded.

### DIFF
--- a/src/TestSuite/Fixture/TruncationStrategy.php
+++ b/src/TestSuite/Fixture/TruncationStrategy.php
@@ -66,7 +66,7 @@ class TruncationStrategy implements ResetStrategyInterface
      *
      * @return void
      */
-    public function setupTest(): void
+    public function teardownTest(): void
     {
         $connections = ConnectionManager::configured();
         foreach ($connections as $connection) {
@@ -104,7 +104,7 @@ class TruncationStrategy implements ResetStrategyInterface
      *
      * @return void
      */
-    public function teardownTest(): void
+    public function setupTest(): void
     {
     }
 
--- a/tests/TestCase/TestSuite/Fixture/TruncationStrategyTest.php
+++ b/tests/TestCase/TestSuite/Fixture/TruncationStrategyTest.php
@@ -30,7 +30,7 @@ class TruncationStrategyTest extends TestCase
     /**
      * Test that beforeTest truncates tables from the previous test
      */
-    public function testSetupSimple(): void
+    public function testTeardownSimple(): void
     {
         $articles = TableRegistry::get('Articles');
         $articlesTags = TableRegistry::get('ArticlesTags');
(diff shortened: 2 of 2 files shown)


othercorey:
Can we add a failure event handler which triggers a cleanup in the next test setup?

markstory:
> Can we add a failure event handler which triggers a cleanup in the next test setup?

We could potentially use a file or a global variable/static property to do an initial cleanup. Each test run is a separate process, so I don't think event handlers will work well. Is using global state along the lines of what you were thinking? I'm not sure that a global will work with the `runInSeparateProcess` mode that PHPUnit has, but we could try it.

othercorey:
> Is using global state along the lines of what you were thinking?

I misread your first comment. You're talking about manual interruption of the test. In that case, all we can require is a clean db state before starting the test run. Is there any way we can enforce that? I don't think we can when using the raw sql dump file.

markstory:
> You're talking about manual interruption of the test. In that case, all we can require is a clean db state before starting the test run. Is there any way we can enforce that? I don't think we can when using the raw sql dump file.

Yes, manually killing the tests with `ctrl-c` or using `die()` in a test would cause the database to be 'broken'.

I don't think we have a reasonable solution for migration based fixtures. SQL dumps are simple as the entire database is dropped at the beginning of each test run and we re-create all the tables.

othercorey:
> I don't think we have a reasonable solution for migration based fixtures. SQL dumps are simple as the entire database is dropped at the beginning of each test run and we re-create all the tables.

We could verify that the db has no tables at the start of the test. However, I know there are existing apps out there that use special support of importing fixtures.

markstory:
> We could verify that the db has no tables at the start of the test. 

We could, that would require the migrations adapter to drop schema at the beginning of each test run. Currently it re-uses schema if the migration state matches the migrations in code. This helps make tests faster as schema only is rebuilt when migrations change.

othercorey:
> We could, that would require the migrations adapter to drop schema at the beginning of each test run. Currently it re-uses schema if the migration state matches the migrations in code. This helps make tests faster as schema only is rebuilt when migrations change.

I could argue this either way. Your best judgement on the least trouble for developers iterating locally is probably the right choice.

markstory:
While having database state left behind after a test run makes debugging after a test simpler. Having the first test that uses fixtures fail on test runs after the first is a really clunky local experience that will be treated as a bug by developers.

PR #15661, opened 2021-07-07
Improve performance of truncation strategy

markstory:
Store schema reflection results statically. This reduces the number of reflection queries we run to one per table. Previously a reflection query would run per test / per fixture.


othercorey:
Looks ok but doesn't seem to be the cause of sql server slowdown.

markstory:
Darn, well at least it removed ~1min for postgres and MySQL vs. https://github.com/cakephp/cakephp/runs/3013624704

PR #15667, opened 2021-07-09
Fix var, param and return tag type hint format

othercorey:
Refs https://github.com/cakephp/cakephp/issues/15528

This orders the type hints in tags according to the above RFC.

### DIFF
--- a/src/Console/ConsoleIo.php
+++ b/src/Console/ConsoleIo.php
@@ -155,7 +155,7 @@ class ConsoleIo
     /**
      * Output at the verbose level.
      *
-     * @param string|string[] $message A string or an array of strings to output
+     * @param array<string>|string $message A string or an array of strings to output
      * @param int $newlines Number of newlines to append
      * @return int|null The number of bytes returned from writing to stdout
      *   or null if current level is less than ConsoleIo::VERBOSE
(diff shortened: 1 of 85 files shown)


othercorey:
This is an example of `\Closure` sorting. Not sure what we prefer:
```
-   * @param \Closure|\Psr\Http\Server\MiddlewareInterface|string $middleware Middleware.
+   * @param \Psr\Http\Server\MiddlewareInterface|\Closure|string $middleware Middleware.
```

ADmad:
I thought we would be doing this sorting only for 5.0.

othercorey:
> I thought we would be doing this sorting only for 5.0.

I manually applied it to 4.next to avoid annoying merge conflicts later. I did for the other type hint cleanup.

othercorey:
Another thing to note, the array version of each type is grouped with that type:

`string[]|string|int[]|int`

We discussed converting all array versions to generics `array<string>` which would group all arrays together:

`array<string>|array<int>|string|int`

othercorey:
Updated to include var, psalm- and phpstan- tags.

src/I18n/TranslatorRegistry.php
```diff
@@ -87,7 +87,7 @@ class TranslatorRegistry
      * A CacheEngine object that is used to remember translator across
      * requests.
      *
-     * @var (\Psr\SimpleCache\CacheInterface&\Cake\Cache\CacheEngineInterface)|null
+     * @var (\Psr\SimpleCache\CacheInterface & \Cake\Cache\CacheEngineInterface)|null
```
othercorey, line 90: Looks like have to remove the spacing from intersection types. However, the space does make it a lot easier to read.

othercorey:
Updated to include `type[]` to `array<type>` array type hint conversion. The array hints are left in the original order they were encountered but grouped together with `array`.

PR #15668, opened 2021-07-09
Update league/container to v4.

ADmad:

### DIFF
--- a/src/Core/ContainerInterface.php
+++ b/src/Core/ContainerInterface.php
@@ -16,8 +16,7 @@ declare(strict_types=1);
  */
 namespace Cake\Core;
 
-use League\Container\Definition\DefinitionInterface;
-use Psr\Container\ContainerInterface as PsrInterface;
+use League\Container\DefinitionContainerInterface;
 
 /**
  * Interface for the Dependency Injection Container in CakePHP applications
@@ -31,33 +30,6 @@ use Psr\Container\ContainerInterface as PsrInterface;
  * @experimental This interface is not final and can have additional
  *   methods and parameters added in future minor releases.
  */
-interface ContainerInterface extends PsrInterface
+interface ContainerInterface extends DefinitionContainerInterface
 {
-    /**
-     * Add an item to the container.
-     *
-     * @param string $id The class name or name of the service being registered.
-     * @param mixed $concrete Either the classname an interface or name resolves to.
-     *   Can also be a constructed object, Closure, or null. When null, the `$id` parameter will
-     *   be used as the concrete class name.
-     * @param bool $shared Set to true to make a service shared.
-     * @return \League\Container\Definition\DefinitionInterface
-     */
-    public function add(string $id, $concrete = null, bool $shared = false): DefinitionInterface;
-
-    /**
-     * Add a service provider to the container
-     *
-     * @param \League\Container\ServiceProvider\ServiceProviderInterface $provider The service provider to add.
-     * @return $this
-     */
-    public function addServiceProvider($provider);
-
-    /**
-     * Modify an existing definition
-     *
-     * @param string $id The class name or name of the service being modified.
-     * @return \League\Container\Definition\DefinitionInterface
-     */
-    public function extend(string $id): DefinitionInterface;
 }
--- a/src/Core/ServiceProvider.php
+++ b/src/Core/ServiceProvider.php
@@ -18,7 +18,6 @@ namespace Cake\Core;
 
 use League\Container\ServiceProvider\AbstractServiceProvider;
 use League\Container\ServiceProvider\BootableServiceProviderInterface;
-use Psr\Container\ContainerInterface as PsrContainerInterface;
 use RuntimeException;
 
 /**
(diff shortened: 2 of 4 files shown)


ADmad:
`league/container` v4 is PHP 7.2+ so this could be backported to 4.3 too.

ADmad:
Well we have a problem, `League\Container\ServiceProvider\AbstractServiceProvider::getContainer()` has return type `\League\Container\DefinitionContainerInterface` while our `Cake\Core\getContainer()` has `League\Container\DefinitionContainerInterface` (in docblock).

Now that since `container/league` v4 has concrete return type our shenanigans won't work.

src/Core/ContainerInterface.php
```diff
@@ -40,18 +41,17 @@ interface ContainerInterface extends PsrInterface
      * @param mixed $concrete Either the classname an interface or name resolves to.
      *   Can also be a constructed object, Closure, or null. When null, the `$id` parameter will
      *   be used as the concrete class name.
-     * @param bool $shared Set to true to make a service shared.
      * @return \League\Container\Definition\DefinitionInterface
      */
-    public function add(string $id, $concrete = null, bool $shared = false): DefinitionInterface;
+    public function add(string $id, $concrete = null): DefinitionInterface;
```
ADmad, line 46: League's container now has an explicit `addShared()` method, should we add the same to our interface?

src/Core/ServiceProvider.php
```diff
@@ -40,15 +40,15 @@ abstract class ServiceProvider extends AbstractServiceProvider implements Bootab
      * This method's actual return type and documented return type differ
      * because PHP 7.2 doesn't support return type narrowing.
      *
-     * @return \Cake\Core\ContainerInterface
+     * @return \League\Container\DefinitionContainerInterface
      */
-    public function getContainer(): PsrContainerInterface
+    public function getContainer(): DefinitionContainerInterface
```
ADmad, line 45: I think we will need to create our own `DefinitionContainerInterface` which extends `DefinitionContainerInterface`.

dereuromark: Safer in the end, allows certain shimming where needed.

src/Core/ContainerInterface.php
markstory: We should.

src/Core/ServiceProvider.php
markstory: Agreed. We could start off with a simple extension and add if we need to. Having our own interface helps keep league container out of user land code more.

markstory:
These changes could be made in 4.3 if we also allow `league/container:^3.2|^4.0`. The container is an experimental feature and can have small breaking changes like these made.

ADmad:
> These changes could be made in 4.3 if we also allow league/container:^3.2|^4.0.

With the type changes I am no longer sure how feasible it would be to support 3.2+ and 4.0+ together. We can see about backporting once the changes are finished for Cake 5.

src/Core/ContainerInterface.php
```diff
@@ -31,33 +30,6 @@
  * @experimental This interface is not final and can have additional
  *   methods and parameters added in future minor releases.
  */
-interface ContainerInterface extends PsrInterface
+interface ContainerInterface extends DefinitionContainerInterface
```
ADmad, line 33: Making `Cake\Core\ContainerInterface` extend `League\Container\DefinitionContainerInterface` was the only sane way I could think of to satisfy the typing requirements.

othercorey:
> With the type changes I am no longer sure how feasible it would be to support 3.2+ and 4.0+ together.

Could just make cakephp 4.3 require container 4+.

ADmad:
> Could just make cakephp 4.3 require container 4+.

Yup that's an option.

tests/test_app/TestApp/ServiceProvider/PersonServiceProvider.php
```diff
@@ -19,4 +19,14 @@ public function services(ContainerInterface $container): void
     {
         $container->add('sally', json_decode('{"name":"sally"}'));
     }
+
+    public function provides(string $id): bool
```
markstory, line 23: Is this a new method that service providers will need to implement? Should we provide a default implementation that reads the property so that application developers don't need to update their service providers immediately?

ADmad: Yup it's a new method on the interface, I will into a default implementation.

ADmad: @markstory Done.

src/Core/ServiceProvider.php
```diff
@@ -33,6 +33,14 @@
  */
 abstract class ServiceProvider extends AbstractServiceProvider implements BootableServiceProviderInterface
 {
+    /**
+     * List of ids of services this provider provides.
+     *
+     * @var array<string>
+     * @see ServiceProvider::provides()
+     */
+    protected $services = [];
```
markstory, line 42: Isn't the old property name [provides](https://container.thephpleague.com/3.x/service-providers/)? Could we use that so the upgrade is even easier?

ADmad: Reverted to the old property name. I missed that fact that the properly already existed, just the new method was added.

markstory: Nice. This will help make upgrading much simpler. :clap:

ADmad:
So is this good to go?

PR #15671, opened 2021-07-11
cake 5: Add native type hints from docblock type hints

othercorey:

### DIFF
--- a/src/Cache/Cache.php
+++ b/src/Cache/Cache.php
@@ -239,7 +239,7 @@ class Cache
      * @param string $config Optional string configuration name to write to. Defaults to 'default'
      * @return bool True if the data was successfully cached, false on failure
      */
-    public static function write(string $key, $value, string $config = 'default'): bool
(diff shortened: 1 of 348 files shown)


src/Cache/Cache.php
```diff
@@ -239,7 +239,7 @@ public static function pool(string $config)
      * @param string $config Optional string configuration name to write to. Defaults to 'default'
      * @return bool True if the data was successfully cached, false on failure
      */
-    public static function write(string $key, $value, string $config = 'default'): bool
+    public static function write(string $key, mixed $value, string $config = 'default'): bool
```
markstory, line 242: Is mixed a native type now?

othercorey: With php 8, we get a few new types:

- mixed
- false
- static

static can only be used a return type and false cannot be used by itself, must be part of a union.

othercorey: We also get an interface called `Stringable`. https://www.php.net/manual/en/class.stringable

PR #15674, opened 2021-07-12
cake 5: Add native type hints for Core

othercorey:
This pulls in just the src/Core files and dependent changes.

The key change is removing `callable` support from a few classes that extend ObjectRegistry and adding object support to a few.

### DIFF
--- a/src/Console/HelperRegistry.php
+++ b/src/Console/HelperRegistry.php
@@ -92,14 +92,18 @@ class HelperRegistry extends ObjectRegistry
      *
      * Part of the template method for Cake\Core\ObjectRegistry::load()
      *
-     * @param string $class The classname to create.
+     * @param \Cake\Console\Helper|string $class The classname to create.
      * @param string $alias The alias of the helper.
      * @param array $config An array of settings to use for the helper.
      * @return \Cake\Console\Helper The constructed helper class.
      * @psalm-suppress MoreSpecificImplementedParamType
      */
-    protected function _create($class, string $alias, array $config): Helper
+    protected function _create(object|string $class, string $alias, array $config): Helper
     {
+        if (is_object($class)) {
+            return $class;
+        }
+
         /** @var \Cake\Console\Helper */
         return new $class($this->_io, $config);
     }
--- a/src/Core/Configure.php
+++ b/src/Core/Configure.php
@@ -82,7 +82,7 @@ class Configure
      * @return void
      * @link https://book.cakephp.org/4/en/development/configuration.html#writing-configuration-data
      */
-    public static function write($config, $value = null): void
+    public static function write(array|string $config, mixed $value = null): void
     {
         if (!is_array($config)) {
             $config = [$config => $value];
@@ -117,7 +117,7 @@ class Configure
      * @return mixed Value stored in configure, or null.
      * @link https://book.cakephp.org/4/en/development/configuration.html#reading-configuration-data
      */
-    public static function read(?string $var = null, $default = null)
+    public static function read(?string $var = null, mixed $default = null): mixed
     {
         if ($var === null) {
             return static::$_values;
@@ -160,7 +160,7 @@ class Configure
      * @throws \RuntimeException if the requested configuration is not set.
      * @link https://book.cakephp.org/4/en/development/configuration.html#reading-configuration-data
      */
-    public static function readOrFail(string $var)
+    public static function readOrFail(string $var): mixed
     {
         if (!static::check($var)) {
             throw new RuntimeException(sprintf('Expected configuration key "%s" not found.', $var));
@@ -200,7 +200,7 @@ class Configure
      * @throws \RuntimeException if the requested configuration is not set.
      * @since 3.6.0
      */
-    public static function consumeOrFail(string $var)
+    public static function consumeOrFail(string $var): mixed
     {
         if (!static::check($var)) {
             throw new RuntimeException(sprintf('Expected configuration key "%s" not found.', $var));
@@ -216,9 +216,9 @@ class Configure
      * out of configure into the various other classes in CakePHP.
      *
      * @param string $var The key to read and remove.
-     * @return array|string|null
+     * @return mixed
      */
-    public static function consume(string $var)
+    public static function consume(string $var): mixed
     {
         if (strpos($var, '.') === false) {
             if (!isset(static::$_values[$var])) {
--- a/src/Core/ObjectRegistry.php
+++ b/src/Core/ObjectRegistry.php
@@ -73,11 +73,11 @@ abstract class ObjectRegistry implements Countable, IteratorAggregate
      *
      * @param string $name The name/class of the object to load.
      * @param array $config Additional settings to use when loading the object.
-     * @return mixed
+     * @return object
      * @psalm-return TObject
      * @throws \Exception If the class cannot be found.
      */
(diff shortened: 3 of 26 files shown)


src/Core/Retry/CommandRetry.php
```diff
@@ -60,15 +60,15 @@ public function __construct(RetryStrategyInterface $strategy, int $maxRetries =
      *
      * @param callable $action The callable action to execute with a retry strategy
      * @return mixed The return value of the passed action callable
-     * @throws \Exception
+     * @throws \Throwable
```
othercorey, line 63: Updated to use Throwable and psalm still thinks there is another return path

src/Core/ContainerInterface.php
```diff
@@ -43,15 +44,15 @@ interface ContainerInterface extends PsrInterface
      * @param bool $shared Set to true to make a service shared.
      * @return \League\Container\Definition\DefinitionInterface
      */
-    public function add(string $id, $concrete = null, bool $shared = false): DefinitionInterface;
+    public function add(string $id, mixed $concrete = null, bool $shared = false): DefinitionInterface;
```
ADmad, line 47: Better to avoid making any changes to container related classes/interfaces as it will cause conflict with #15668.

src/Core/Configure.php
```diff
@@ -218,7 +218,7 @@ public static function consumeOrFail(string $var)
      * @param string $var The key to read and remove.
      * @return array|string|null
      */
-    public static function consume(string $var)
+    public static function consume(string $var): array|string|null
```
ADmad, line 221: This should be `mixed`.

src/Core/Retry/CommandRetry.php
ADmad: It probably can't figure out all the code paths due to the `while(true)`. We can just ignore the error.

src/Core/functions.php
```diff
@@ -38,7 +38,7 @@
      * @return mixed Wrapped text.
      * @link https://book.cakephp.org/4/en/core-libraries/global-constants-and-functions.html#h
      */
-    function h($text, bool $double = true, ?string $charset = null)
+    function h(mixed $text, bool $double = true, ?string $charset = null): mixed
```
ADmad, line 41: This should be `string`.

src/Core/functions.php
```diff
@@ -192,7 +192,7 @@ function pj($var)
      * @return string|bool|null Environment variable setting.
      * @link https://book.cakephp.org/4/en/core-libraries/global-constants-and-functions.html#env
      */
-    function env(string $key, $default = null)
+    function env(string $key, string|bool|null $default = null): string|bool|null
```
ADmad, line 195: I hope someone isn't setting integers.

src/Datasource/ConnectionRegistry.php
```diff
@@ -65,20 +65,13 @@ protected function _throwMissingClassError(string $class, ?string $plugin): void
...
-     * function will be the result of the callable.
-     *
-     * @param \Cake\Datasource\ConnectionInterface|callable|string $class The classname or object to make.
+     * @param \Cake\Datasource\ConnectionInterface|string $class The classname or object to make.
      * @param string $alias The alias of the object.
      * @param array $config An array of settings to use for the datasource.
      * @return \Cake\Datasource\ConnectionInterface A connection with the correct settings.
      */
-    protected function _create($class, string $alias, array $config)
+    protected function _create(object|string $class, string $alias, array $config): ConnectionInterface
```
ADmad, line 73: Not all connection classes implement `ConnectionInterface`, for e.g. https://github.com/UseMuffin/Webservice/blob/9bde1646d6ed938a5e2b2a0dd6e78b3a89b6ac2a/src/Connection.php#L16.

It's because `ConnectionInterface` has methods like `transactional()` which are irrelevant for non RDBMS datasource.

othercorey: This is the big issue. I made changes so everything passes since it expects ConnectionInterface. You can see how FakeConnection was failing as well.

I'm not sure ConnectionRegistry can or should return random objects.

src/Core/functions.php
othercorey: We could set it to mixed since the $default doesn't comply with getenv() types.

src/Core/functions.php
othercorey: We can make this change in 4.3 I think.

src/Console/HelperRegistry.php
```diff
@@ -92,14 +92,18 @@ protected function _throwMissingClassError(string $class, ?string $plugin): void
...
      *
-     * @param string $class The classname to create.
+     * @param \Cake\Console\Helper|string $class The classname to create.
      * @param string $alias The alias of the helper.
      * @param array $config An array of settings to use for the helper.
      * @return \Cake\Console\Helper The constructed helper class.
      * @psalm-suppress MoreSpecificImplementedParamType
      */
-    protected function _create($class, string $alias, array $config): Helper
+    protected function _create(object|string $class, string $alias, array $config): Helper
```
saeideng, line 101: why not `Helper|string` ??

othercorey: contravariance
https://www.php.net/manual/en/language.oop5.variance.php

Cannot restrict the parameter type.

ADmad: Because that wouldn't be contravariant, i.e. you can only widen the argument type in a child class, not narrow it down.

src/Core/functions.php
ADmad: IMO this method should only act as an abstraction to avoid modifying the actual environment vars and be complaint with `getenv()`. But that's a separate discussion.

src/Datasource/ConnectionRegistry.php
ADmad: Well currently it can return any object. Let's limit the scope of this PR to only add types as per current behavior (unless it's a trivial change).

We can discuss changes to `ConnectionRegistry`/`ConnectionInterface` separately.

othercorey: Yup, it has to be just `object` return type. 

What about `callable` support? Is that necessary? I think we can expand `object|string` to `callable|object|string` now that I think about it. We cannot restrict `object|string` to just `string` though.

ADmad: I think only connection registry has callable support to theoretically allow lazy creation of connection instance. That way connecting to a (external) datasource would be avoided unless needed.

To maintain the status quo we could just use `callable|object|string`.

othercorey:
Updated the registry changes to include everything we possibly can.

src/Core/ObjectRegistry.php
```diff
@@ -77,7 +77,7 @@ abstract class ObjectRegistry implements Countable, IteratorAggregate
      * @psalm-return TObject
      * @throws \Exception If the class cannot be found.
      */
-    public function load(string $name, array $config = [])
+    public function load(string $name, array $config = []): mixed
```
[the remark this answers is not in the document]
othercorey: agreed.

src/Core/functions.php
othercorey: I tried this in 4.3, but not comfortable casting the scalars to string. Would rather error if a bool is passed in.

PR #15675, opened 2021-07-13
Resolves out-of-date/inaccurate doc blocks for Configure::load

cnizzardini:
A boolean false is returned only if the engine is not found. When a file is not found a CakeException is raised.

\Cake\Core\Exception\CakeException:

```php
Configure::load('i_dont_exist_config', 'default');
```

Returns `false`:

```php
Configure::load('i_do_exist_config', 'but_this_engine_does_not');
```

While this resolves the doc blocks, I am not exactly sure if this is how this should work :-/ I think it should be the other way around or better yet have both throw exceptions...

### DIFF
--- a/src/Core/Configure.php
+++ b/src/Core/Configure.php
@@ -321,7 +321,8 @@ class Configure
      * @param string $key name of configuration resource to load.
      * @param string $config Name of the configured engine to use to read the resource identified by $key.
      * @param bool $merge if config files should be merged instead of simply overridden
-     * @return bool False if file not found, true if load successful.
+     * @return bool False if engine not found, true if load successful.
+     * @throws \Cake\Core\Exception\CakeException if file not found
      * @link https://book.cakephp.org/4/en/development/configuration.html#reading-and-writing-configuration-files
      */
     public static function load(string $key, string $config = 'default', bool $merge = true): bool


src/Core/Configure.php
```diff
@@ -321,7 +321,8 @@ public static function drop(string $name): bool
      * @param string $key name of configuration resource to load.
      * @param string $config Name of the configured engine to use to read the resource identified by $key.
      * @param bool $merge if config files should be merged instead of simply overridden
-     * @return bool False if file not found, true if load successful.
+     * @return bool False if engine not found, true if load successful.
+     * @throws \Cake\Core\Exception\CakeException if file not found
```
ADmad, line 325: Well the exception is thrown by engine class (which is `PhpConfig`) by default not this method (or it's direct helper method), so mentioning the exception here seems wrong.

But yes simply returning `false` when an engine cannot be loaded seems wrong. It should be an exception. Perhaps we can change that in the next major.

cnizzardini: Hmmm, what about try/catch and throwing another exception of the same kind? I don't think that would be a breaking change and would allow users to know what to expect. This can cause issues in applications since per doc blocks you don't expect an exception.

markstory: > But yes simply returning false when an engine cannot be loaded seems wrong. It should be an exception. I can't think of a use case where failure to load a config wouldn't matter. Perhaps we can change that in the next major.

We can make this change in a minor release too. New exceptions are not breaking changes as per our compatibility guidelines. The user land code wasn't working before so we are unlikely to be breaking any useful usage.

cnizzardini: Alright, I can close this and create an issue instead.

PR #15678, opened 2021-07-14
Remove deprecated behavior from App::path().

ADmad:

### DIFF
--- a/src/Core/App.php
+++ b/src/Core/App.php
@@ -16,6 +16,8 @@ declare(strict_types=1);
  */
 namespace Cake\Core;
 
+use Cake\Core\Exception\CakeException;
+
 /**
  * App is responsible for resource location, and path management.
  *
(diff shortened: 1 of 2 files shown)


src/Core/App.php
```diff
@@ -187,7 +187,7 @@ protected static function _classExistsInBase(string $name, string $namespace): b
      */
     public static function path(string $type, ?string $plugin = null): array
     {
-        if ($plugin === null && $type[0] === strtolower($type[0])) {
+        if ($plugin === null) {
             return (array)Configure::read('App.paths.' . $type);
```
garas, line 191: Both `/** @psalm-suppress PossiblyNullArgument */` can now be removed below because `$plugin` never `null`.

markstory: Couldn't a null still be provided as an argument?

garas: When `$plugin === null`, function exits here, so few lines below (GitHub didn't let comment on them directly) `Plugin::templatePath($plugin)` and `Plugin::path($plugin)` always passes `string`.

ADmad: Yup I'll remove those annotations.

PR #15679, opened 2021-07-15
cake 5: Add native type hints for Cache

othercorey:


src/Cache/CacheEngine.php
```diff
@@ -265,15 +265,15 @@ abstract public function increment(string $key, int $offset = 1);
...
+    abstract public function decrement(string $key, int $offset = 1): int|false;
 
     /**
      * Delete a key from the cache
      *
      * @param string $key Identifier for the data
      * @return bool True if the value was successfully deleted, false if it didn't exist or couldn't be removed
      */
-    abstract public function delete($key): bool;
+    abstract public function delete($key);
```
ADmad, line 276: v2 of PSR's `CacheInterface` is going to have `bool` return type https://github.com/php-fig/simple-cache/pull/24/files#diff-8f44301e3ecaf1100cc9dd79bd90d859a2e3100ff576e5d5adb54415e0828501R46

othercorey: Did not mean to remove the return type. I was only trying to remove the added type hints to parameters for psr interface.

ADmad: v1 doesn't have return type but the (yet to be released) will have it :)

othercorey: So, the code already add the return types set. It is covariant so it works. 

Do you want to remove the existing return types?

ADmad: > Do you want to remove the existing return types?

No

PR #15682, opened 2021-07-15
Configure::load() now throws an exception when config engine does not exist 

cnizzardini:
Alters `Configure::load()` to throw an exception if the `$config` engine does not exist. A test is added so there is coverage for this now.

See discussion in #15675


markstory:
Thank you 👍

PR #15683, opened 2021-07-16
PHP 8.1: Iterator returns

othercorey:
MultipleIterator tentative return type mismatch.

@markstory


othercorey:
Removed TestIterator because ArrayIterator::append() conflicts with CollectionTrait::append(). The only test this is used in doesn't seem necessary.

tests/TestCase/Collection/CollectionTest.php
```diff
@@ -2553,30 +2552,6 @@ protected function datePeriod($start, $end): DatePeriod
...
-        $this->assertTrue($iterator->checkValues());
-
-        //We need to perform at least two collection operation to trigger the issue.
-        $newIterator = $iterator
-            ->filter(function ($item) {
-                return $item < 5;
-            })
-            ->reject(function ($item) {
-                return $item > 2;
-            });
```
markstory, line 2573: This was a bug/regression from a few years back. Removing the test means we *could* regress here again, but I suspect this is unlikely to happen unless we were to make dramatic changes to `CollectionTrait`.

othercorey: Well, CollectionTrait::append() and ArrayIterator::append() are conflicting. I'm not sure what users were trying to do with that.

PR #15684, opened 2021-07-16
cake 5: Remove use of deprecated Serializable

othercorey:

### DIFF
--- a/src/Collection/Collection.php
+++ b/src/Collection/Collection.php
@@ -18,13 +18,12 @@ namespace Cake\Collection;
 
 use ArrayIterator;
 use IteratorIterator;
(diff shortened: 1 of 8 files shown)


othercorey:
I should mention that @ADmad added __serialize() and __unserialize() to all places we implemented Serializable in 4.3. This removes the dual code.

ADmad:
> I should mention that @ADmad added __serialize() and __unserialize() to all places we implemented Serializable in 4.3. This removes the dual code.

Yup going through the diff I realized after a while that we had added `__serialize()`, `__unserialize()` earlier :).

markstory:
Looks good to me.

PR #15685, opened 2021-07-16
cake 5: Add native type hints for Utility

othercorey:

### DIFF
--- a/src/Utility/Hash.php
+++ b/src/Utility/Hash.php
@@ -18,6 +18,13 @@ namespace Cake\Utility;
 use ArrayAccess;
 use InvalidArgumentException;
 use RuntimeException;
+use const SORT_ASC;
+use const SORT_DESC;
+use const SORT_LOCALE_STRING;
+use const SORT_NATURAL;
+use const SORT_NUMERIC;
+use const SORT_REGULAR;
+use const SORT_STRING;
 
 /**
  * Library of array functions for manipulating and extracting data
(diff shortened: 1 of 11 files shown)


src/Utility/Xml.php
```diff
@@ -307,7 +303,7 @@ public static function fromArray($input, array $options = [])
      * @return void
      * @throws \Cake\Utility\Exception\XmlException
      */
-    protected static function _fromArray(DOMDocument $dom, $node, &$data, $format): void
+    protected static function _fromArray(DOMDocument $dom, DOMDocument|DOMElement $node, $data, string $format): void
```
othercorey, line 306: I gave up trying to fix the $data type hint. We pass in many types even thought it only accept array.

src/Utility/Hash.php
```diff
@@ -649,7 +643,7 @@ public static function check(array $data, string $path): bool
      * @return array Filtered array
      * @link https://book.cakephp.org/4/en/core-libraries/hash.html#Cake\Utility\Hash::filter
      */
-    public static function filter(array $data, $callback = ['self', '_filter']): array
+    public static function filter(array $data, callable|array $callback = ['self', '_filter']): array
```
ADmad, line 646: Does the `$callback` arg need to have `|array` because PHP can't always figure out if the argument is a valid callable?

src/Utility/Hash.php
```diff
@@ -1070,7 +1064,7 @@ public static function sort(array $data, string $path, $dir = 'asc', $type = 're
      * @param mixed $key The key for the data.
      * @return array
      */
-    protected static function _squash(array $data, $key = null): array
+    protected static function _squash(array $data, mixed $key = null): array
```
ADmad, line 1067: Perhaps using `int|string` instead of `mixed` for `$key` might make static analyzers happy.

src/Utility/Security.php
```diff
@@ -60,7 +60,7 @@ class Security
      * @throws \RuntimeException
      * @link https://book.cakephp.org/4/en/core-libraries/security.html#hashing-data
      */
-    public static function hash(string $string, ?string $algorithm = null, $salt = false): string
+    public static function hash(string $string, ?string $algorithm = null, mixed $salt = false): string
```
ADmad, line 63:
```suggestion
    public static function hash(string $string, ?string $algorithm = null, bool|string $salt = false): string
```

src/Utility/Xml.php
ADmad: This whole Xml class needs to die :stuck_out_tongue:.

markstory: It comes in handy when you need to use XML though. Being able to create an array out of XML and XML from an array can be useful at times.

src/Utility/Hash.php
othercorey: The problem is the default. The default is just an array that gets re-interpreted at runtime when used in the context of the function. Seems invalid to me.

othercorey: I changed this to `?callable` and created the default inside the method.

PR #15686, opened 2021-07-16
Replace getTypeName() with PHP's get_debug_type().

ADmad:

### DIFF
--- a/src/Core/functions.php
+++ b/src/Core/functions.php
@@ -312,16 +312,3 @@ if (!function_exists('deprecationWarning')) {
         trigger_error($message, E_USER_DEPRECATED);
     }
 }
-
-if (!function_exists('getTypeName')) {
-    /**
-     * Returns the objects class or var type of it's not an object
-     *
-     * @param mixed $var Variable to check
-     * @return string Returns the class name or variable type
-     */
-    function getTypeName(mixed $var): string
-    {
-        return is_object($var) ? get_class($var) : gettype($var);
-    }
-}
--- a/src/Error/Debugger.php
+++ b/src/Error/Debugger.php
@@ -274,7 +274,7 @@ class Debugger
     {
         $instance = static::getInstance();
         if (!is_string($template) && !($template instanceof Closure)) {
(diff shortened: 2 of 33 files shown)


ADmad:
I dropped `getTypeName()` altogether since replacing it with `get_debug_type()` would be trivial using a search/replace.

src/Core/functions.php
```diff
@@ -312,16 +312,3 @@ function deprecationWarning(string $message, int $stackFrame = 1): void
...
 }
-
-if (!function_exists('getTypeName')) {
-    /**
-     * Returns the objects class or var type of it's not an object
-     *
-     * @param mixed $var Variable to check
-     * @return string Returns the class name or variable type
-     */
-    function getTypeName(mixed $var): string
```
markstory, line 323: Nice to see that PHP devs were thinking along the same lines as we were.

markstory (approved):

src/Error/Debugger.php
```diff
@@ -668,16 +668,19 @@ public static function exportVarAsNodes($var, int $maxDepth = 3): NodeInterface
     protected static function export($var, DebugContext $context): NodeInterface
     {
         $type = static::getType($var);
+
+        if (preg_match('/^resource /', $type)) {
```
garas, line 672:
```suggestion
        if (str_starts_with($type, 'resource ')) {
```
PHP 8

othercorey:
Can we add this to a list of changes to consider for `upgrade`?

src/Error/Debugger.php
ADmad: Thanks, will take a while for it to get the new into the daily vocabulary :)

othercorey:
> Can we add this to a list of changes to consider for `upgrade`?

Added to 5.0 wiki

ADmad:
> Added to 5.0 wiki

Thanks

markstory:
If we keep the changes we want to automate later on. I'm happy to build up the appropriate changes for `rector`.

markstory (approved):

PR #15687, opened 2021-07-17
cake 5: ZipIterator::current() should return array

othercorey:
I was trying to fix the static analysis errors from merging, but I found there is an actual error.

`MultipleIterator::current()` has a return type of `array`. `ZipIterator::current()` does not return array when a callback is passed in.

This doesn't look great because zipWith() suggests reducing to a single value. However, the return type needs to be consistent.

### DIFF
--- a/tests/TestCase/Collection/CollectionTest.php
+++ b/tests/TestCase/Collection/CollectionTest.php
@@ -1930,12 +1930,12 @@ class CollectionTest extends TestCase
         $zipped = $collection->zipWith([3, 4], function ($a, $b) {
             return $a * $b;
         });
-        $this->assertEquals([3, 8], $zipped->toList());
+        $this->assertEquals([[3], [8]], $zipped->toList());
 
         $zipped = $collection->zipWith([3, 4], [5, 6, 7], function (...$args) {
             return array_sum($args);
         });
-        $this->assertEquals([9, 12], $zipped->toList());
+        $this->assertEquals([[9], [12]], $zipped->toList());
     }
 
     /**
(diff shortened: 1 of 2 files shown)


othercorey:
To re-create the original behavior, we could wrap the ZipIterator in an iterator that unwraps the first level.

tests/TestCase/Collection/CollectionTest.php
```diff
@@ -1930,12 +1930,12 @@ public function testZipWith(): void
         $zipped = $collection->zipWith([3, 4], function ($a, $b) {
             return $a * $b;
         });
-        $this->assertEquals([3, 8], $zipped->toList());
+        $this->assertEquals([[3], [8]], $zipped->toList());
```
markstory, line 1933: This change will break user land code and how a 'zipWith' operation should work. I'm pretty sure the intent of this method is to be an analog to [haskell's zipWith](http://zvon.org/other/haskell/Outputprelude/zipWith_f.html)

othercorey: right, but how to we create that with MultipleIterator::current?

markstory: Maybe we shouldn't be extending `MultipleIterator` if it requires us to return an array?

othercorey: I guess we can just leave the ReturnTypeWillChange attribute for 5.0 and hope nothing else affects it.

markstory: I think we should be setting the php compatibility to `>8,<8.2` based on how the recent PHP releases have been going 😛

PR #15688, opened 2021-07-18
cake 5: Add native type hints for Validation

othercorey:

### DIFF
--- a/src/Validation/Validation.php
+++ b/src/Validation/Validation.php
@@ -140,7 +140,7 @@ class Validation
      * @param mixed $check Value to check
      * @return bool Success
      */
-    public static function notBlank($check): bool
+    public static function notBlank(mixed $check): bool
     {
         if (empty($check) && !is_bool($check) && !is_numeric($check)) {
             return false;
@@ -158,7 +158,7 @@ class Validation
      * @param mixed $check Value to check
      * @return bool Success
      */
-    public static function alphaNumeric($check): bool
+    public static function alphaNumeric(mixed $check): bool
     {
         if ((empty($check) && $check !== '0') || !is_scalar($check)) {
             return false;
@@ -176,7 +176,7 @@ class Validation
      * @param mixed $check Value to check
      * @return bool Success
      */
-    public static function notAlphaNumeric($check): bool
+    public static function notAlphaNumeric(mixed $check): bool
     {
         return !static::alphaNumeric($check);
     }
@@ -187,7 +187,7 @@ class Validation
      * @param mixed $check Value to check
      * @return bool Success
      */
-    public static function asciiAlphaNumeric($check): bool
+    public static function asciiAlphaNumeric(mixed $check): bool
     {
         if ((empty($check) && $check !== '0') || !is_scalar($check)) {
             return false;
@@ -202,7 +202,7 @@ class Validation
      * @param mixed $check Value to check
      * @return bool Success
      */
-    public static function notAsciiAlphaNumeric($check): bool
+    public static function notAsciiAlphaNumeric(mixed $check): bool
     {
         return !static::asciiAlphaNumeric($check);
     }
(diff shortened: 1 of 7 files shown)


src/Validation/Validation.php
```diff
@@ -324,14 +324,14 @@ public static function numElements($check, string $operator, int $expectedCount)
...
+     * @param mixed $check1 The left value to compare.
      * @param string $operator Can be one of following operator strings:
      *   '>', '<', '>=', '<=', '==', '!=', '===' and '!=='. You can use one of
      *   the Validation::COMPARE_* constants.
-     * @param string|int $check2 The right value to compare.
+     * @param mixed $check2 The right value to compare.
      * @return bool Success
      */
-    public static function comparison($check1, string $operator, $check2): bool
+    public static function comparison(mixed $check1, string $operator, mixed $check2): bool
```
othercorey, line 334: This is due to compareFields().

src/Validation/Validation.php
```diff
@@ -677,7 +677,7 @@ public static function localizedTime($check, string $type = 'datetime', $format
      * @param array $booleanValues List of valid boolean values, defaults to `[true, false, 0, 1, '0', '1']`.
      * @return bool Success.
      */
-    public static function boolean($check, array $booleanValues = []): bool
+    public static function boolean(string|int|bool $check, array $booleanValues = []): bool
```
markstory, line 680: This should probably be mixed as well. We could get anything from end user request data.

src/Validation/Validation.php
```diff
@@ -695,7 +695,7 @@ public static function boolean($check, array $booleanValues = []): bool
      * @param array $truthyValues List of valid truthy values, defaults to `[true, 1, '1']`.
      * @return bool Success.
      */
-    public static function truthy($check, array $truthyValues = []): bool
+    public static function truthy(string|int|bool $check, array $truthyValues = []): bool
```
markstory, line 698: Same here with using `mixed`.

src/Validation/Validation.php
```diff
@@ -713,7 +713,7 @@ public static function truthy($check, array $truthyValues = []): bool
      * @param array $falseyValues List of valid falsey values, defaults to `[false, 0, '0']`.
      * @return bool Success.
      */
-    public static function falsey($check, array $falseyValues = []): bool
+    public static function falsey(string|int|bool $check, array $falseyValues = []): bool
```
markstory, line 716: Same here with using `mixed`

src/Validation/Validation.php
```diff
@@ -1205,7 +1205,7 @@ public static function luhn($check): bool
      * @throws \RuntimeException when mime type can not be determined.
      * @throws \LogicException when ext/fileinfo is missing
      */
-    public static function mimeType($check, $mimeTypes = []): bool
+    public static function mimeType(UploadedFileInterface|array|string $check, array|string $mimeTypes = []): bool
```
markstory, line 1208: This should probably be mixed as well. Well formed requests should only have `UploadedFileInterface|array|string` but bad ones could have arbitrary garbage in them.

src/Validation/Validation.php
```diff
@@ -839,7 +827,7 @@ public static function equalTo($check, $comparedTo): bool
      * @param array<string> $extensions file extensions to allow. By default extensions are 'gif', 'jpeg', 'png', 'jpg'
      * @return bool Success
      */
-    public static function extension($check, array $extensions = ['gif', 'jpeg', 'png', 'jpg']): bool
+    public static function extension(UploadedFileInterface|array|string $check, array $extensions = ['gif', 'jpeg', 'png', 'jpg']): bool
```
othercorey, line 830: Should functions like extension accept mixed also?

ADmad: Validation function take user input so it could be of any type.

othercorey: This is true. However, I've had to adjust only a few tests due to restricting types. I wonder if we're properly testing "supported" input in validation outside of a few specific exceptions.

othercorey: For example, this code certainly expects the type to be at least string and does not early return otherwise.

markstory: I think so. Validation could be passed anything as it gets end user data.

othercorey: Updated extension() to mixed and  few others. Added a note in the docblock about supported values.

othercorey:
@markstory @ADmad last call. There are a lot of methods so I left the PR open for longer.

PR #15691, opened 2021-07-20
cake 5: Move Filesystem to Utility

othercorey:
This follows the removal of File and Folder classes.


dereuromark:
The alternative would be to just not split this file into its own repo anymore.

ADmad:
> The alternative would be to just not split this file into its own repo anymore.

That's what's done here. The `cakephp/filesystem` repo is killed.

dereuromark:
Sure, I meant that the need to move the file is then not given anymore
`Cake\Filesystem\Filesystem` could remain if we wanted to.

othercorey:
The `console` package does depend on it at least.

PR #15696, opened 2021-07-21
Adds an extra check before trying to set the session_id

bmilleare:
In our tests we're seeing the following error pop up frequently:
```
Warning Error: session_id(): Cannot change session id when session is active
In [vendor/cakephp/cakephp/src/Http/Session.php, line 533]
```
This adds a simple check to ensure the session_id() is empty before trying to set it.

### DIFF
--- a/src/Http/Session.php
+++ b/src/Http/Session.php
@@ -529,7 +529,7 @@ class Session
      */
     public function id(?string $id = null): string
     {
-        if ($id !== null && !headers_sent()) {
+        if ($id !== null && !headers_sent() && session_id() == '') {
             session_id($id);
         }
 


dereuromark:
Shouldn't this rather through an exception if used (wrongly) in such a case?
To not cloak and hide the fact that something went wrong and the session id was not set as the method call suggests?

bmilleare:
Interestingly, it seems your own tests pass without a warning in `tests/TestCase/Http/SessionTest.php` despite `session_id()` returning `cli` prior to the call.

The warning seems to have appeared since PHP 7.2 so was perhaps failing silently prior, but I'll investigate further and add a new PR if necessary.

PR #15697, opened 2021-07-21
Default value for View::$template

andrii-pukhalevych:
Default value for 'View::$template = ''' is needed to because `View::getTemplate(): string` does not expect to return null
After this PR template will behave the same way as templatePath

### DIFF
--- a/src/Core/Exception/CakeException.php
+++ b/src/Core/Exception/CakeException.php
@@ -115,5 +115,5 @@ class CakeException extends RuntimeException
 }
 
 // phpcs:disable
-class_exists('Cake\Core\Exception\Exception');
+class_exists('Cake\Core\Exception\CakeException');
 // phpcs:enable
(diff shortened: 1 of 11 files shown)


dereuromark:
The test case needs to also change

The alternative could be to allow null (change docblock) and cast this on getting.
Not sure what is better, though.

andrii-pukhalevych:
I'm also not sure, so just did similar to `View::templatePath`.

Regarding test - please help me here, because tests for `View::template` and `View::templatePath`also should be similar.

dereuromark:
If this breaks anything, shouldnt this go to master instead?
Or is this fine to be then for 4.3?

markstory:
I think this change is fine, but why is the template property not being set?

othercorey:
@andrii-pukhalevych Can you follow up on this?

andrii-pukhalevych:
Not sure what more is needed from me

PR #15698, opened 2021-07-22
cake 5: Add native type hints for Console

othercorey:
Did not add return type for `BaseCommand::execute()` to allow implicit null return. Otherwise, we have to enforce returning something.

@markstory

### DIFF
--- a/src/Console/Arguments.php
+++ b/src/Console/Arguments.php
@@ -140,9 +140,9 @@ class Arguments
      * Get an option's value or null
      *
      * @param string $name The name of the option to check.
-     * @return string|int|bool|null The option value or null.
+     * @return string|bool|null The option value or null.
      */
-    public function getOption(string $name)
+    public function getOption(string $name): string|bool|null
     {
         return $this->options[$name] ?? null;
     }
--- a/src/Console/ConsoleInputOption.php
+++ b/src/Console/ConsoleInputOption.php
@@ -90,7 +90,7 @@ class ConsoleInputOption
      * @param string $short The short alias for this option
      * @param string $help The help text for this option
      * @param bool $isBoolean Whether this option is a boolean option. Boolean options don't consume extra tokens
-     * @param string|bool|null $default The default value for this option.
+     * @param mixed $default The default value for this option.
      * @param array<string> $choices Valid choices for this option.
      * @param bool $multiple Whether this option can accept multiple value definition.
      * @param bool $required Whether this option is required or not.
@@ -205,7 +205,7 @@ class ConsoleInputOption
      *
      * @return string|bool|null
      */
-    public function defaultValue()
+    public function defaultValue(): string|bool|null
     {
         return $this->_default;
     }
@@ -247,7 +247,7 @@ class ConsoleInputOption
      * @return true
      * @throws \Cake\Console\Exception\ConsoleException
      */
-    public function validChoice($value): bool
+    public function validChoice(string|bool $value): bool
     {
         if (empty($this->_choices)) {
             return true;
--- a/src/Console/ConsoleIo.php
+++ b/src/Console/ConsoleIo.php
@@ -160,7 +160,7 @@ class ConsoleIo
      * @return int|null The number of bytes returned from writing to stdout
      *   or null if current level is less than ConsoleIo::VERBOSE
      */
-    public function verbose($message, int $newlines = 1): ?int
+    public function verbose(array|string $message, int $newlines = 1): ?int
     {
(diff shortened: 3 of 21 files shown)


tests/TestCase/Console/HelpFormatterTest.php
```diff
@@ -183,7 +183,7 @@ public function testHelpWithOptions(): void
         $parser->addOption('test', ['help' => 'A test option.'])
                ->addOption('number', [
                    'help' => 'The number',
-                   'default' => 2,
+                   'default' => '2',
```
othercorey, line 186: Option defaults must be boolean or a string.

src/Console/Arguments.php
```diff
@@ -142,7 +142,7 @@ public function getOptions(): array
      * @param string $name The name of the option to check.
      * @return string|int|bool|null The option value or null.
      */
-    public function getOption(string $name)
+    public function getOption(string $name): string|int|bool|null
```
markstory, line 145: If default values can only be strings should this also include `int` or should we remove `int` and have userland code convert/cast?

tests/TestCase/Console/HelpFormatterTest.php
markstory: Will non-bool|string values cause errors? We might want to add this to the list of things we fix with rector, as I'm certain this will impact applications.

othercorey: It will throw a TypeError when creating ConsoleInputOption instance.

https://github.com/cakephp/cakephp/blob/5.0/src/Console/ConsoleOptionParser.php#L431

That class casts everything to string as well:

https://github.com/cakephp/cakephp/blob/5.0/src/Console/ConsoleInputOption.php#L120

src/Console/Arguments.php
othercorey: Seems like we should remove it to align with default. The parsing code also re-casts the default to a string.

https://github.com/cakephp/cakephp/blob/5.0/src/Console/ConsoleOptionParser.php#L895

I'm not sure there is actually an "int" anywhere in the parsed option values. We get strings from php, and I don't see it converted anywhere.

tests/TestCase/Console/HelpFormatterTest.php
othercorey: We could also uptype it and let it cast everything. However the getOption should probably change.

src/Console/ConsoleInputOption.php
```diff
@@ -90,7 +90,7 @@ class ConsoleInputOption
      * @param string $short The short alias for this option
      * @param string $help The help text for this option
      * @param bool $isBoolean Whether this option is a boolean option. Boolean options don't consume extra tokens
-     * @param string|bool|null $default The default value for this option.
+     * @param mixed $default The default value for this option.
```
othercorey, line 93: @markstory Changed to mixed in annotation but removed native type hint. Everything is cast to string.

src/Console/Arguments.php
```diff
@@ -140,9 +140,9 @@ public function getOptions(): array
      * Get an option's value or null
      *
      * @param string $name The name of the option to check.
-     * @return string|int|bool|null The option value or null.
+     * @return string|bool|null The option value or null.
      */
-    public function getOption(string $name)
+    public function getOption(string $name): string|bool|null
```
othercorey, line 145: Removed `int` since all values are strings from the command line.

tests/TestCase/Console/ArgumentsTest.php
```diff
@@ -141,14 +141,14 @@ public function testGetOption(): void
         $options = [
             'verbose' => true,
             'off' => false,
-            'zero' => 0,
+            'zero' => '0',
```
othercorey, line 144: The `Arguments` class accepts an array of options, but these are expected to be parsed from command line so they must be string or bool at this point.

src/Console/ConsoleIo.php
```diff
@@ -300,7 +300,7 @@ public function success($message, int $newlines = 1, int $level = self::NORMAL):
      * @return void
      * @throws \Cake\Console\Exception\StopException
      */
-    public function abort($message, $code = CommandInterface::CODE_ERROR): void
+    public function abort(string $message, int $code = CommandInterface::CODE_ERROR): void
```
saeideng, line 303: `error()` accepts `array|string` and this is only `string` , !!?

othercorey: Looks like StopException expects a single string.

markstory (approved):
Looks good to me.

PR #15705, opened 2021-07-23
Fix radio inputs trimming leading - off causing duplicate ids

### LINKED ISSUE #15700: Form radio id when value is negative
This is a (multiple allowed):

* [x] bug
* [ ] enhancement

* CakePHP Version: 4.x and 3.9
* Browser: Firefox & Brave (chromium)

### What you did
```php
echo $this->Form->control('bounces', [
    'type' => 'radio',
    'options' => [
        0 => 'default',
        -1 => 'reject',
        1 => 'accept',
    ],
])
```

### What happened
Code above produce this html
```html
<label for="bounces-0">
    <input type="radio" name="bounces" value="0" id="bounces-0">default
</label>
<label for="bounces-1">
    <input type="radio" name="bounces" value="-1" id="bounces-1">reject
</label>
<label for="bounces-1">
    <input type="radio" name="bounces" value="1" id="bounces-1">accept
</label>
```

The `for` in label for value `1` and `-1` are exactly same.  Like `input` ids
When users click on `accept label`, the `reject` radio is checked

markstory:
Fixes #15700

### DIFF
--- a/src/View/Widget/RadioWidget.php
+++ b/src/View/Widget/RadioWidget.php
@@ -169,7 +169,7 @@ class RadioWidget extends BasicWidget
 
         if (empty($radio['id'])) {
             if (isset($data['id'])) {
-                $radio['id'] = $data['id'] . '-' . trim(
+                $radio['id'] = $data['id'] . '-' . rtrim(
                     $this->_idSuffix((string)$radio['value']),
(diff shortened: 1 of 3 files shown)


othercorey:
Looks ok.

PR #15709, opened 2021-07-24
fix class name

saeideng:
looks class name should be `Cake\Core\Exception\Exception`

see previous commit : https://github.com/cakephp/cakephp/commit/d43092a93821


othercorey:
Looks ok. Good catch.

PR #15713, opened 2021-07-25
cake 5: Add native type hints for Http

othercorey:
@markstory Please take a close look at the changes to Cookie. We expect `array|string` values but have been accepting other values due to implicit string conversion when writing the headers.

### DIFF
--- a/src/Http/Cookie/Cookie.php
+++ b/src/Http/Cookie/Cookie.php
@@ -73,7 +73,7 @@ class Cookie implements CookieInterface
     /**
      * Expiration time
      *
-     * @var \DateTime|\DateTimeImmutable|null
+     * @var \DateTimeInterface|null
      */
     protected $expiresAt;
 
@@ -136,8 +136,8 @@ class Cookie implements CookieInterface
      *
      * @link http://php.net/manual/en/function.setcookie.php
      * @param string $name Cookie name
-     * @param array|string $value Value of the cookie
-     * @param \DateTime|\DateTimeImmutable|null $expiresAt Expiration time and date
+     * @param array|string|float|int|bool $value Value of the cookie
+     * @param \DateTimeInterface|null $expiresAt Expiration time and date
      * @param string|null $path Path
      * @param string|null $domain Domain
      * @param bool|null $secure Is secure
@@ -146,7 +146,7 @@ class Cookie implements CookieInterface
      */
     public function __construct(
         string $name,
-        $value = '',
+        array|string|float|int|bool $value = '',
         ?DateTimeInterface $expiresAt = null,
         ?string $path = null,
         ?string $domain = null,
@@ -210,12 +210,12 @@ class Cookie implements CookieInterface
      * Factory method to create Cookie instances.
      *
      * @param string $name Cookie name
-     * @param array|string $value Value of the cookie
+     * @param array|string|float|int|bool $value Value of the cookie
      * @param array $options Cookies options.
      * @return static
      * @see \Cake\Cookie\Cookie::setDefaults()
      */
-    public static function create(string $name, $value, array $options = [])
+    public static function create(string $name, array|string|float|int|bool $value, array $options = []): static
     {
         $options += static::$defaults;
         $options['expires'] = static::dateTimeInstance($options['expires']);
@@ -236,9 +236,9 @@ class Cookie implements CookieInterface
      * Converts non null expiry value into DateTimeInterface instance.
      *
      * @param mixed $expires Expiry value.
-     * @return \DateTime|\DatetimeImmutable|null
+     * @return \DateTimeInterface|null
      */
-    protected static function dateTimeInstance($expires): ?DateTimeInterface
+    protected static function dateTimeInstance(mixed $expires): ?DateTimeInterface
     {
         if ($expires === null) {
             return $expires;
@@ -275,7 +275,7 @@ class Cookie implements CookieInterface
      * @return static
      * @see \Cake\Cookie\Cookie::setDefaults()
      */
-    public static function createFromHeaderString(string $cookie, array $defaults = [])
+    public static function createFromHeaderString(string $cookie, array $defaults = []): static
     {
         if (strpos($cookie, '";"') !== false) {
             $cookie = str_replace('";"', '{__cookie_replace__}', $cookie);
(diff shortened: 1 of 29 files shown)


src/Http/Cookie/Cookie.php
```diff
@@ -416,15 +417,15 @@ protected function validateName(string $name): void
...
+    public function getValue(): array|string
     {
         return $this->value;
     }
 
     /**
      * @inheritDoc
      */
-    public function getScalarValue()
+    public function getScalarValue(): array|string
```
othercorey, line 428: This one doesn't make much sense. The value is still `array|string` not `mixed` and certainly isn't converting anything.

tests/TestCase/Http/ResponseTest.php
```diff
@@ -749,26 +749,6 @@ public function testWithCookieEmpty(): void
         $this->assertEquals($expected, $result);
     }
 
-    /**
-     * Test setting cookies with scalar values
-     */
-    public function testWithCookieScalar(): void
```
othercorey, line 755: These tests don't make much sense. They are testing the value stored in the Cookie object which is not the value sent and retrieved via headers. Those are expected to be strings.

src/Http/Cookie/Cookie.php
```diff
@@ -146,7 +147,7 @@ class Cookie implements CookieInterface
      */
     public function __construct(
         string $name,
-        $value = '',
+        array|string $value = '',
```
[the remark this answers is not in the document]
othercorey: I agree that numbers are probably used very often. Do we want just `int` or `float|int` ?

src/Http/Cookie/Cookie.php
othercorey: I was returning mixed. I'm not sure if static analyzer just ignore undefined return types though.

src/Http/Cookie/Cookie.php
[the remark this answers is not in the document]
ADmad: `float|int`

src/Http/Cookie/Cookie.php
ADmad: The method calls `_flatten()` if the value is expanded (aka an array) which has return type `string`, so the return type here should should be a `string`.

src/Http/Cookie/Cookie.php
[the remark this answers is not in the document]
markstory: We're unlikely to get floats from HTTP responses as that is all strings, but users creating cookies will use all sorts of scalar values. I agree that including `float|int` is a good idea, I might also add `bool` to that union, so we can cover all the useful primitives.

src/Http/Cookie/Cookie.php
```diff
@@ -215,7 +216,7 @@ public static function setDefaults(array $options): void
      * @return static
      * @see \Cake\Cookie\Cookie::setDefaults()
      */
-    public static function create(string $name, $value, array $options = [])
+    public static function create(string $name, array|string $value, array $options = []): static
```
markstory, line 219: You might want `int|float|bool` here too.

src/Http/Cookie/Cookie.php
```diff
@@ -451,7 +452,7 @@ public function withValue($value)
      * @param array|string $value The value to store.
      * @return void
      */
-    protected function _setValue($value): void
+    protected function _setValue(array|string $value): void
```
markstory, line 455: Would this need to take float|int|bool too?

src/Http/Cookie/CookieInterface.php
```diff
@@ -79,24 +82,24 @@ public function getName(): string;
...
     /**
      * Gets the cookie value as scalar.
      *
      * This will collapse any complex data in the cookie with json_encode()
      *
-     * @return mixed
+     * @return array|string
      */
-    public function getScalarValue();
+    public function getScalarValue(): array|string;
```
markstory, line 94: Should this be string?

src/Http/Cookie/CookieInterface.php
```diff
@@ -79,24 +82,24 @@ public function getName(): string;
...
 
     /**
      * Create a cookie with an updated value.
      *
      * @param array|string $value Value of the cookie to set
      * @return static
      */
-    public function withValue($value);
+    public function withValue(array|string $value): static;
```
markstory, line 102: Could add the `int|float|bool` here too. We could also narrow the type of value to only accept string|array, but I think accepting wider types and casting values to strings makes a more ergonomic API that is aligned with how CakePHP should work.

src/Http/Cookie/CookieInterface.php
```diff
@@ -135,14 +138,14 @@ public function getDomain(): string;
...
-    public function withDomain(string $domain);
+    public function withDomain(string $domain): static;
 
     /**
      * Get the current expiry time
      *
      * @return \DateTime|\DateTimeImmutable|null Timestamp of expiry or null
      */
-    public function getExpiry();
+    public function getExpiry(): DateTime|DateTimeImmutable|null;
```
markstory, line 148: Should this be DateTimeInterface?

src/Http/ServerRequest.php
```diff
@@ -834,7 +834,7 @@ public function getHeader($name): array
      * @return string Header values collapsed into a comma separated string.
      * @link http://www.php-fig.org/psr/psr-7/ This method is part of the PSR-7 server request interface.
      */
-    public function getHeaderLine($name): string
```
markstory, line 837: Is this for compatibility with PSR interfaces?

src/Http/Cookie/Cookie.php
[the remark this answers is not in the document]
othercorey: Converting `false` to `string` creates an empty string. Will that break the header or cookie?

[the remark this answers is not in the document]
ADmad: AFAIK there's nothing that prevents setting empty string as value for a cookie either, so `false` being converted to empty string shouldn't be an issue.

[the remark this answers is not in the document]
ADmad: `setcookie()` also has default value for `$value` arg as empty string.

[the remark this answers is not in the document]
markstory: Empty strings in cookies are totally fine.

src/Http/ServerRequest.php
othercorey: I don't know why this changed. I reset it to `string`.

PR #15714, opened 2021-07-26
cake 5: Add native type hints for Core properties

othercorey:
For uninitialized properties, I use `isset()` instead of using a random initialization.

However, psalm will complain that it's not initialized in the constructor. Should we make all optional properties nullable? Should we just add that psalm warning to the global suppression list?

### DIFF
--- a/src/Console/Exception/ConsoleException.php
+++ b/src/Console/Exception/ConsoleException.php
@@ -29,5 +29,5 @@ class ConsoleException extends CakeException
      *
      * @var int
      */
-    protected $_defaultCode = CommandInterface::CODE_ERROR;
+    protected int $_defaultCode = CommandInterface::CODE_ERROR;
 }
--- a/src/Core/BasePlugin.php
+++ b/src/Core/BasePlugin.php
@@ -34,70 +34,70 @@ class BasePlugin implements PluginInterface
      *
      * @var bool
      */
-    protected $bootstrapEnabled = true;
+    protected bool $bootstrapEnabled = true;
 
     /**
      * Console middleware
      *
      * @var bool
      */
-    protected $consoleEnabled = true;
+    protected bool $consoleEnabled = true;
 
     /**
      * Enable middleware
      *
      * @var bool
      */
-    protected $middlewareEnabled = true;
+    protected bool $middlewareEnabled = true;
 
     /**
      * Register container services
      *
      * @var bool
      */
-    protected $servicesEnabled = true;
+    protected bool $servicesEnabled = true;
 
     /**
      * Load routes or not
      *
      * @var bool
      */
-    protected $routesEnabled = true;
+    protected bool $routesEnabled = true;
 
     /**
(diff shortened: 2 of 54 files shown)


ADmad:
I would not add those `isset()` checks and let PHP throw an error if it's not properly initialized before accessing. Suppress any psalm errors as required.

src/Console/Exception/ConsoleException.php
```diff
@@ -29,5 +29,5 @@ class ConsoleException extends CakeException
      *
      * @var int
```
ADmad, line 30: These annotations are quite redundant now. We could remove them after updating our CS and API docs builder.

othercorey:
> I would not add those `isset()` checks and let PHP throw an error if it's not properly initialized

ok, but I had to add them because it is accessing them.

ADmad:
> ok, but I had to add them because it is accessing them.

In case of `BasePlugin` they are needed since appropriate values are generated if not provided through constructor.

src/Console/Exception/ConsoleException.php
othercorey: I disabled the CS sniff for "useless annotation" to avoid annoying merge errors until we're done making api/docs changes to 4.3.

src/Core/BasePlugin.php
```diff
@@ -34,70 +34,70 @@ class BasePlugin implements PluginInterface
...
-    protected $templatePath;
+    protected string $templatePath;
 
     /**
      * The name of this plugin
      *
      * @var string
      */
-    protected $name;
+    protected string $name;
```
markstory, line 100: These should probably be nullable if we want to have more correct types.

othercorey: Changes them to nullable. Left the isset but it could also be !== null.

src/Core/Plugin.php
```diff
@@ -30,7 +30,7 @@ class Plugin
      *
      * @var \Cake\Core\PluginCollection|null
      */
-    protected static $plugins;
+    protected static ?\Cake\Core\PluginCollection $plugins = null;
```
ADmad, line 33: Should just use classname here. It's in the same namespace so even a `use` statement isn't required. Too bad such changes will have to be done manually.

src/Core/Retry/CommandRetry.php
```diff
@@ -31,17 +31,17 @@ class CommandRetry
      *
      * @var \Cake\Core\Retry\RetryStrategyInterface
      */
-    protected $strategy;
+    protected \Cake\Core\Retry\RetryStrategyInterface $strategy;
```
ADmad, line 34: Just classname here too.

src/Core/Plugin.php
othercorey: There is a sniff I used to fix these up when generating the parameter/return type hints, but have to manually enable it.

PR #15715, opened 2021-07-26
Remove PHP version checks no longer needed.

ADmad:

### DIFF
--- a/src/Http/ResponseEmitter.php
+++ b/src/Http/ResponseEmitter.php
@@ -231,27 +231,7 @@ class ResponseEmitter implements EmitterInterface
             $cookie = Cookie::createFromHeaderString($cookie, ['path' => '']);
         }
 
-        if (PHP_VERSION_ID >= 70300) {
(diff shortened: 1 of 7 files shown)


src/Http/ResponseEmitter.php
```diff
@@ -231,27 +231,7 @@ protected function setCookie($cookie): bool
...
-        return setcookie(
-            $cookie->getName(),
-            $cookie->getScalarValue(),
-            $cookie->getExpiresTimestamp() ?: 0,
-            $path,
-            $cookie->getDomain(),
-            $cookie->isSecure(),
-            $cookie->isHttpOnly()
-        );
+        return setcookie($cookie->getName(), $cookie->getScalarValue(), $cookie->getOptions());
```
othercorey, line 234: How is this not covered?

saeideng (approved):

src/Http/ResponseEmitter.php
ADmad: Codecov/phpunit is drunk.

PR #15722, opened 2021-07-28
Add support for RFC3339_EXTENDED in DateTimeType

Wouter0100:
This adds support and tests for the RFC3339_EXTENDED format defined [here](https://www.php.net/manual/en/class.datetimeinterface.php#datetime.constants.rfc3339_extended) to the DateTimeType.


Wouter0100:
Hmm, I tested by running: `phpunit --filter DateTimeTypeTest` but apparently that misses things.

dereuromark:
Is this already in master (4.x)?
Usually, we add new functionality first into 4.x and then backport to 3.x

othercorey:
The 4.x implementation changed how these formats are accepted, but it does accept this format. It accepts up to microseconds not just milliseconds.

PR #15723, opened 2021-07-29
cake 5: Add native type hints for Event

othercorey:

### DIFF
--- a/src/Event/Decorator/SubjectFilterDecorator.php
+++ b/src/Event/Decorator/SubjectFilterDecorator.php
@@ -34,11 +34,11 @@ class SubjectFilterDecorator extends AbstractDecorator
     /**
      * @inheritDoc
      */
-    public function __invoke()
+    public function __invoke(): mixed
     {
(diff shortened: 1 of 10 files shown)


src/Event/Decorator/SubjectFilterDecorator.php
```diff
@@ -34,11 +34,11 @@ class SubjectFilterDecorator extends AbstractDecorator
     /**
      * @inheritDoc
      */
-    public function __invoke()
+    public function __invoke(): mixed
     {
         $args = func_get_args();
         if (!$this->canTrigger($args[0])) {
-            return false;
+            return null;
```
othercorey, line 41: Is this supposed to be `false`? if so, should `ConditionalDecorator` return null or false? `false` means to stop propagation of the event while null means don't set the result, right?

ADmad: Yes, the `false` does seem odd. Returning `null` makes more sense.

PR #15727, opened 2021-08-01
Enable auto fixtures for CTE tests

othercorey:


othercorey:
@markstory I don't like hiding the issue, but this should just be auto fixtures anyway.

markstory (approved):

PR #15729, opened 2021-08-04
cake 5: Add native type hints for Datasource

othercorey:


src/Datasource/ConnectionManager.php
```diff
@@ -73,12 +73,12 @@ class ConnectionManager
...
      *
      * @param array|string $key The name of the connection config, or an array of multiple configs.
-     * @param array|null $config An array of name => config data for adapter.
+     * @param object|array|null $config An array of name => config data for adapter.
      * @return void
      * @throws \Cake\Core\Exception\CakeException When trying to modify an existing config.
      * @see \Cake\Core\StaticConfigTrait::config()
      */
-    public static function setConfig($key, $config = null): void
+    public static function setConfig(array|string $key, object|array|null $config = null): void
```
othercorey, line 81: Notice this adds `object` for $config to accept instances.

othercorey:
Notice that did not add  `ConnectionInterface` return type hint for `ConnectionManager::get()` because it does not return that type.

othercorey:
@ADmad @markstory please take a look.

src/Datasource/QueryInterface.php
```diff
@@ -40,7 +45,10 @@ interface QueryInterface
      * @param bool $overwrite whether to reset fields with passed list or not
      * @return $this
      */
-    public function select($fields, bool $overwrite = false);
+    public function select(
+        ExpressionInterface|Association|Table|callable|array|string $fields,
```
ADmad, line 49: "I heard ya like types, here ya go, have half a dozen."

src/Datasource/RepositoryInterface.php
```diff
@@ -128,7 +132,7 @@ public function updateAll($fields, $conditions): int;
      * @return int Returns the number of affected rows.
      * @see \Cake\Datasource\RepositoryInterface::delete()
      */
-    public function deleteAll($conditions): int;
+    public function deleteAll(mixed $conditions): int;
```
ADmad, line 135: Could this be narrowed down?

othercorey: I don't know. There is a big abstraction between repository and tables. Not sure if other datasource specify conditions differently.

ADmad: Okay, thx.

dereuromark: With `exists(array $conditions)`, wouldnt it be fair to also assume we could use similar limitations on this one?

src/Datasource/RepositoryInterface.php
```diff
@@ -148,7 +152,7 @@ public function exists($conditions): bool;
      * @param \ArrayAccess|array $options The options to use when saving.
      * @return \Cake\Datasource\EntityInterface|false
      */
-    public function save(EntityInterface $entity, $options = []);
+    public function save(EntityInterface $entity, ArrayAccess|array $options = []): EntityInterface|false;
```
dereuromark, line 155: Shouldnt we go for nullable stuff in 5.0? For a cleaner API?

ADmad: Allowing null here would require adding an extra check for null value or cast to array. So that would be just overhead without any benefit.

dereuromark: As return value?
I dont quite follow.

ADmad: Ah, for the return value, yes I guess that is an option.

Is returning the entity even necessary? Since it's an object it's the same entity reference that's passed as an argument.  I am not aware of a use case where the returned value is a reference to a new entity object. In which case the return value could be just boolean.

src/Datasource/EntityTrait.php
```diff
@@ -601,7 +599,7 @@ public function &offsetGet($offset)
      * @param mixed $value The value to set.
      * @return void
      */
-    public function offsetSet($offset, $value): void
+    public function offsetSet(mixed $offset, mixed $value): void
```
markstory, line 602: Shouldn't the offset be a scalar value?

src/Datasource/EntityTrait.php
```diff
@@ -612,7 +610,7 @@ public function offsetSet($offset, $value): void
      * @param string $offset The offset to remove.
      * @return void
      */
-    public function offsetUnset($offset): void
+    public function offsetUnset(mixed $offset): void
```
markstory, line 613: Shouldn't the offset be a scalar value?

src/Datasource/RepositoryInterface.php
markstory: I think this type could be narrowed to `QueryExpression|Closure|array|string` like updateAll() is. The same for exists().

src/Datasource/EntityTrait.php
othercorey: Have to be compatible with ArrayAccess. All of these implementations use mixed.

https://www.php.net/manual/en/class.arrayaccess.php

src/Datasource/RepositoryInterface.php
dereuromark: Are entities one day immutable? Probably not

src/Datasource/RepositoryInterface.php
othercorey: Those are for fields. $conditions is still `mixed`. It says it accepts whatever `where()` does which is:

`where(Closure|array|string|null $conditions = null`

We could narrow down to that for those parameters.

src/Datasource/FixtureInterface.php
```diff
@@ -47,7 +49,7 @@ public function drop(ConnectionInterface $connection): bool;
      * @return \Cake\Database\StatementInterface|bool on success or if there are no records to insert,
      *  or false on failure.
      */
-    public function insert(ConnectionInterface $connection);
+    public function insert(ConnectionInterface $connection): StatementInterface|bool;
```
ADmad, line 52: Hmm.. `Datasource` packages shouldn't be using classes from `Database`.

src/Datasource/QueryInterface.php
ADmad: `Datasource` package is using classes from `Database` and `ORM` package :(

othercorey: This datasource package is looking less useful.

src/Datasource/QueryCacher.php
```diff
@@ -72,7 +65,7 @@ public function __construct($key, $config)
      * @param object $query The query the cache read is for.
      * @return mixed|null Either the cached results or null.
      */
-    public function fetch(object $query)
+    public function fetch(object $query): mixed
```
ADmad, line 68:
```suggestion
    public function fetch(QueryInterface $query): ?ResultSetInterface
```

Similarly the signature for `store()` can also be changed to 
```
store(QueryInterface $query, ResultSetInterface $results): bool
```

and `resolveKey(object $query): string` to
```
resolveKey(QueryInterface $query): string
```

othercorey: I would prefer that to be a separate PR.

src/Datasource/FixtureInterface.php
othercorey: Removed the type hint but left the docblock annotation because this interface design is broken.

src/Datasource/QueryInterface.php
othercorey: Changed QueryInterface to type hint only non-database types and then added them in the subclasses.

src/Datasource/QueryInterface.php
```diff
@@ -148,9 +150,9 @@ public function find(string $finder, array $options = []);
      * $singleUser = $query->select(['id', 'username'])->first();
      * ```
      *
-     * @return \Cake\Datasource\EntityInterface|array|null the first result from the ResultSet
+     * @return mixed the first result from the ResultSet
```
ADmad, line 153: Why is this widened?  EntityInterface is within the same packaged and these 3 are the only possible return types AFAIK.

othercorey: Because of mapReduce()/etc.

src/Datasource/QueryInterface.php
othercorey: Added `int` because we test selecting an integer.

src/Datasource/QueryCacher.php
ADmad: Okay :+1: .

Issue #13049, opened 2019-03-14
Query field type overwritten by associated table field with the same name


berarma:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 3.7.5
* Platform and Target: PHP5.6, MySQL 5.5

### What you did

I have a model Resources that belongs to many Services that has many Deals (Resources -> Services <- Deals).

The Services model has an integer field named ```min_price```. The Deals model has a decimal field named ```price```.

I write a query like this:
```
$query = $this->Resources->find();
$query->innerJoinWith('Services.Deals');
$query->group('Resources.id'');
$query->select(['min_price' => $query->func()->min('Deals.price')]);
$query->all();
```

Notice that I'm defining an alias ```min_price``` over ```Deals.price``` and I have a field with the same name in the model Services.

### What happened

As expected, the results have the ```min_price``` value that I defined as an alias but it's of type integer instead of float. It seems to take the type from the ```Services.min_price``` field.

I've tried using this line to the same effect:

```
$query->select(['min_price' => $query->func()->min('Deals.price', ['decimal'])]);
```

### What you expected to happen

I expected the ```min_price``` defined as an alias to be of type float in the results.

I've found this line in the Query object that sets the type:

https://github.com/cakephp/cakephp/blob/0e63c784a61baf65d97177b07bd5f50574218745/src/ORM/Query.php#L272

I've tried replacing it with the following line and it worked like expected but I don't know the implications:

```
            $fields[$alias . '.' . $f] = $fields[$alias . '__' . $f] = $type;
```

I'm going to work around it by changing the alias name. I know this is a very weird behavior/bug that won't be triggered many times but may happen and it's hard to track.

Regards.

markstory:
> I've tried replacing it with the following line and it worked like expected but I don't know the implications:

You could try adding a test case for this scenario and applying your fix. That would let us see if there is an existing test scenario that fails with the new changes.

ADmad:
The type specified in function expression will now take precedence over default column types #15738.

PR #15738, opened 2021-08-08
Fix type specified for FunctionExpression not being respected.

ADmad:
Refs #13049.

While I think this is a bug I am being extra cautious here and targeting `4.next`. I can re-target it to `master` if it's generally considered a safe change.


othercorey:
This is only going to affect those who are aliasing an existing field in the schema and expecting that type. I don't think that's an empty set of users though. I've accidentally relied on it before.

ADmad:
The CTE related test failures were occurring earlier too and don't seem to be related to my changes.

othercorey:
@markstory Looks ok for 4.3 to me, can you double check?

PR #15739, opened 2021-08-09
Fix broken markdown in Table::findList() docs

larsgw:
Add a missing markdown fence after a code block for the `valueField` feature option in the documentation of `\Cake\ORM\Table::findList()`. See https://api.cakephp.org/4.2/class-Cake.ORM.Table.html#findList().

Since this is a very small patch (in a comment even) I did not make an issue, I hope that is okay.


othercorey:
Thanks!

PR #15742, opened 2021-08-12
cake 5: Add native type hints for Database

othercorey:


src/Database/Type/BinaryUuidType.php
```diff
@@ -67,10 +67,10 @@ public function newId(): string
      *
      * @param mixed $value The value to convert.
      * @param \Cake\Database\DriverInterface $driver The driver instance to convert with.
-     * @return resource|string|null
+     * @return string|null
      * @throws \Cake\Core\Exception\CakeException
      */
-    public function toPHP($value, DriverInterface $driver)
+    public function toPHP(mixed $value, DriverInterface $driver): string|null
```
othercorey, line 73: @markstory do you know why we support resources in toPHP()?

src/Database/Statement/BufferedStatement.php
```diff
@@ -186,18 +186,15 @@ public function bind(array $params, array $types): void
...
 
     /**
-     * {@inheritDoc}
-     *
-     * @param string|int $type The type to fetch.
-     * @return array|false
+     * @inheritDoc
      */
-    public function fetch($type = self::FETCH_TYPE_NUM)
+    public function fetch(string|int $type = self::FETCH_TYPE_NUM): stdClass|array|false
```
othercorey, line 197: This can return `stdClass` which was undocumented. That is causing all of the psalm errors.

src/Database/Expression/CaseExpression.php
```diff
@@ -56,14 +56,17 @@ class CaseExpression implements ExpressionInterface
...
      * conditions passed in $conditions. If there are more $values than $conditions,
      * the last $value is used as the `ELSE` value.
      * @param array $types associative array of types to be associated with the values
      * passed in $values
      */
-    public function __construct($conditions = [], $values = [], $types = [])
-    {
+    public function __construct(
+        ExpressionInterface|array $conditions = [],
+        ExpressionInterface|array|string|float|int|bool|null $values = [],
```
markstory, line 67: Type alias pain :grimacing:

src/Database/Query.php
```diff
@@ -1742,14 +1754,8 @@ public function values($data)
...
      * @return $this
      */
-    public function update($table)
+    public function update(ExpressionInterface|string $table)
     {
-        if (!is_string($table) && !($table instanceof ExpressionInterface)) {
-            $text = 'Table must be of type string or "%s", got "%s"';
-            $message = sprintf($text, ExpressionInterface::class, gettype($table));
-            throw new InvalidArgumentException($message);
-        }
```
markstory, line 1751: Great to see this kind of condition supported by the interpreter.

src/Database/Statement/BufferedStatement.php
```diff
@@ -92,12 +93,14 @@ public function __construct(StatementInterface $statement, DriverInterface $driv
...
      */
-    public function __get(string $property)
+    public function __get(string $property): mixed
     {
         if ($property === 'queryString') {
             /** @psalm-suppress NoInterfaceProperties */
             return $this->statement->queryString;
         }
+
+        throw new RuntimeException("Cannot access undefined property `$property`.");
```
markstory, line 103: This could be `InvalidArgumentException` if you wanted a more accurate exception type.

src/Database/Statement/BufferedStatement.php
markstory: Solving all of those type errors is going to be tedious :grimacing:

ADmad: Shouldn't the return type be just `mixed` to match `StatementInterface::fetch()`?

othercorey: I changed it to return `mixed`. We don't support the PDOStatement fetch type that returns true, but specifying `stdClass|array|false` causes too many type errors.

PR #15744, opened 2021-08-13
Remove mutable versions of datetime classes.

ADmad:
Maintaining both mutable and immutable versions is unnecessary.


src/Database/Type/DateTimeType.php
```diff
@@ -465,13 +465,13 @@ public function getDateTimeClassName(): string
     }
 
     /**
-     * Change the preferred class name to the mutable Time implementation.
+     * Change the preferred class name to the mutable implementation.
      *
      * @return $this
      */
     public function useMutable()
```
markstory, line 472: Shouldn't this method be removed as well?

src/I18n/RelativeTimeFormatter.php
```diff
@@ -99,7 +99,7 @@ public function diffForHumans(
      */
     public function timeAgoInWords(I18nDateTimeInterface $time, array $options = []): string
     {
-        $options = $this->_options($options, FrozenTime::class);
```
markstory, line 102: Should we retain an alias for `FrozenTime` so that users don't have to churn a bunch of code for the class rename? Or is that something that we should add to rector rules for 5.x?

markstory (commented):
Before we can remove the mutable date/datetime we'd need to deprecate it in 4.x.

src/Database/Type/DateTimeType.php
ADmad: Okay, will remove the the option to use mutable classes altogether. I have kept it as the datettime type classes fallback to `DateTime` and `DateTimeImmuble` is using `cakephp/database` standalone and `cakephp/i18n` isn't loaded.

src/I18n/RelativeTimeFormatter.php
ADmad: It would be a simple rename so we could just add rector rules for it.

ADmad:
> Before we can remove the mutable date/datetime we'd need to deprecate it in 4.x.

Okay, I'll make a PR for 4.3 deprecating them.

ADmad:
Don't know why that one test with Danish locale is throwing a hissy fit.

src/Database/Type/DateTimeType.php
markstory: Falling back to `DateTimeImmutable` is always how it has worked though. We'd only be losing the date clamping features when `i18n` isn't present which is how it has historically worked as well.

src/I18n/RelativeTimeFormatter.php
markstory: I've added it to the transform list in the 5.x wiki.

src/Database/Type/DateTimeType.php
ADmad: Yip, done.

src/Database/Type/DateType.php
```diff
@@ -49,27 +47,13 @@ class DateType extends DateTimeType
     protected $setToDateStart = true;
 
     /**
-     * Change the preferred class name to the FrozenDate implementation.
-     *
-     * @return $this
-     */
-    public function useImmutable()
```
markstory, line 56: If we aren't deprecating this in 4.3 (as we shouldn't) then we should also consider leaving this method in as a no-op to avoid breaking compatibility. We'll have guided people to call this method in every application they have and it will be an annoying bump to have to fix immediately.

src/I18n/Time.php
```diff
@@ -135,90 +135,8 @@ public function __construct($time = null, $tz = null)
...
-     *
-     * The format to be used is stored in the static property `Time::niceFormat`.
-     *
-     * @param \DateTimeZone|string|null $timezone Timezone string or DateTimeZone object
-     * in which the date will be displayed. The timezone stored for this object will not
-     * be changed.
-     * @param string|null $locale The locale name in which the date should be displayed (e.g. pt-BR)
-     * @return string Formatted date string
-     */
-    public function nice($timezone = null, $locale = null): string
```
markstory, line 152: Are these methods on the `DateFormatTrait` already?

src/Database/Type/DateType.php
ADmad: Hmm... I think it's best to deprecated this method too in 4.3. It's called from the constructor itself so generally users don't need to call it explicitly.

src/I18n/Time.php
ADmad: Yes, or inherited from `Chronos`.

othercorey:
Looks ok to merge after rebase.

ADmad:
> Looks ok to merge after rebase.

Done but can't figure out why that test with Danish locale is failing.

othercorey:
Testing this locally with danish locale, the test passes when run individually but fails when the full test suite is run. 

My guess is the removals cause a different order in the unit tests. There must be a setting that isn't being cleared or we have a flaw in the config somewhere.

ADmad:
Renamed `Cake\I18n\Time` to `Cake\I18n\DateTime`.

ADmad:
Can we skip the failing test for now and merge this?

othercorey:
> Can we skip the failing test for now and merge this?

I think so. Just create an issue to track.

ADmad:
> I think so. Just create an issue to track.

Done.

PR #15746, opened 2021-08-15
Raise error when no mock responses are found

markstory:
When the HttpClient cannot find a mocked response it should raise an error. This makes it simpler to detect mistakes in mock setup code, and make tests more explicit.

### DIFF
--- a/src/Http/Client/Exception/MissingResponseException.php
+++ b/src/Http/Client/Exception/MissingResponseException.php
@@ -15,22 +15,15 @@ declare(strict_types=1);
  */
 namespace Cake\Http\Client\Exception;
 
-use RuntimeException;
+use Cake\Core\Exception\CakeException;
 
 /**
  * Used to indicate that a request did not have a matching mock response.
  */
-class MissingResponseException extends RuntimeException
+class MissingResponseException extends CakeException
 {
     /**
-     * Constructor
-     *
-     * @param string $method The HTTP method used.
-     * @param string $url The request URL.
+     * @var string
      */
-    public function __construct(string $method, string $url)
-    {
-        $message = "Unable to find a mocked response for {$method} to {$url}.";
-        parent::__construct($message);
-    }
+    protected $_messageTemplate = 'Unable to find a mocked response for `%s` to `%s`.';
 }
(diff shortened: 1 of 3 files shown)


src/Http/Client/Exception/MissingResponseException.php
```diff
@@ -0,0 +1,36 @@
...
+/**
+ * CakePHP(tm) : Rapid Development Framework (https://cakephp.org)
+ * Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
+ *
+ * Licensed under The MIT License
+ * Redistributions of files must retain the above copyright notice.
+ *
+ * @copyright     Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
+ * @link          https://cakephp.org CakePHP(tm) Project
+ * @since         3.0.0
```
ADmad, line 13:
```suggestion
 * @since         4.3.0
```

src/Http/Client/Exception/MissingResponseException.php
```diff
@@ -0,0 +1,36 @@
...
+ * @license       https://opensource.org/licenses/mit-license.php MIT License
+ */
+namespace Cake\Http\Client\Exception;
+
+use RuntimeException;
+
+/**
+ * Used to indicate that a request did not have a matching mock response.
+ */
+class MissingResponseException extends RuntimeException
```
ADmad, line 23: Why not extend `CakeException`?  `cakephp/http` has `cakephp/core` as depedency.

src/Http/Client/Exception/MissingResponseException.php
```diff
@@ -0,0 +1,36 @@
...
+{
+    /**
+     * Constructor
+     *
+     * @param string $method The HTTP method used.
+     * @param string $url The request URL.
+     */
+    public function __construct(string $method, string $url)
+    {
+        $message = "Unable to find a mocked response for {$method} to {$url}.";
```
saeideng, line 33: if this exception class is only for mock , could we add `@internal` to the class, like mock driver ?
if it is general, this message is not general because of `mocked` :thinking:

markstory: It is only for missing mocks but application code could use the class name in a catch block.

PR #15747, opened 2021-08-15
Add a trait to make defining mock responses simpler

markstory:
The trait automatically clears up mocks and makes possible to define mocks with no static class name references.

### DIFF
--- a/src/TestSuite/HttpClientTrait.php
+++ b/src/TestSuite/HttpClientTrait.php
@@ -1,0 +1,117 @@
+<?php
+declare(strict_types=1);
+
+/**
+ * CakePHP(tm) : Rapid Development Framework (https://cakephp.org)
+ * Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
+ *
+ * Licensed under The MIT License
+ * For full copyright and license information, please see the LICENSE.txt
+ * Redistributions of files must retain the above copyright notice.
+ *
+ * @copyright     Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
+ * @link          https://cakephp.org CakePHP(tm) Project
+ * @since         4.3.0
+ * @license       https://opensource.org/licenses/mit-license.php MIT License
+ */
+namespace Cake\TestSuite;
+
+use Cake\Http\Client;
+use Cake\Http\Client\Response;
+
+/**
+ * Define mock responses and have mocks automatically cleared.
+ */
+trait HttpClientTrait
+{
+    /**
+     * Resets mocked responses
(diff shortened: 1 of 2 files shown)


tests/TestCase/TestSuite/HttpClientTraitTest.php
```diff
@@ -0,0 +1,59 @@
...
+use Cake\TestSuite\TestCase;
+
+class HttpClientTraitTest extends TestCase
+{
+    use HttpClientTrait;
+
+    /**
+     * Provider for http methods.
+     *
+     * @return void
```
othercorey, line 30: returns array<array<string>> ?

src/TestSuite/HttpClientTrait.php
```diff
@@ -0,0 +1,117 @@
...
+
+    /**
+     * Create a new response.
+     *
+     * @param int $code The response code to use. Defaults to 200
+     * @param array $headers A list of headers for the response. Example `Content-Type: application/json`
+     * @param string $body The body for the response.
+     * @return \Cake\Http\Client\Response
+     */
+    public function newResponse(int $code = 200, array $headers = [], string $body = ''): Response
```
othercorey, line 46: Should this have `Client` in the method name to match the others?

tests/TestCase/TestSuite/HttpClientTraitTest.php
```diff
@@ -0,0 +1,59 @@
...
+            ['Patch'],
+            ['Delete'],
+        ];
+    }
+
+    /**
+     * @dataProvider methodProvider
+     * @param string $httpMethod Http Method name.
+     */
+    public function testRequestMethods($httpMethod)
```
othercorey, line 47:
```suggestion
     */
    public function testRequestMethods(string $httpMethod)
```

just type hint it

tests/TestCase/TestSuite/HttpClientTraitTest.php
```diff
@@ -0,0 +1,59 @@
...
+    {
+        $traitMethod = "mockClient{$httpMethod}";
+        $clientMethod = strtolower($httpMethod);
```
othercorey, line 50: what does strtolower do?

markstory: Gets the casing to match the method names on the http client.

othercorey: Method names are case insensitive.

markstory: True, I can remove the case change.

PR #15748, opened 2021-08-17
cake 5: Add native type hints for ORM

othercorey:
@markstory Please check

### DIFF
--- a/src/ORM/Association.php
+++ b/src/ORM/Association.php
@@ -20,6 +20,8 @@ use Cake\Collection\Collection;
 use Cake\Core\App;
 use Cake\Core\ConventionsTrait;
 use Cake\Database\Expression\IdentifierExpression;
+use Cake\Database\Expression\QueryExpression;
+use Cake\Database\ExpressionInterface;
 use Cake\Datasource\EntityInterface;
 use Cake\Datasource\ResultSetDecorator;
 use Cake\ORM\Locator\LocatorAwareTrait;
@@ -113,7 +115,7 @@ abstract class Association
     /**
      * The name of the field representing the foreign key to the table to load
      *
-     * @var array<string>|string
+     * @var array<string>|string|false
      */
     protected $_foreignKey;
 
@@ -424,9 +426,9 @@ abstract class Association
      *
      * @param \Closure|array $conditions list of conditions to be used
      * @see \Cake\Database\Query::where() for examples on the format of the array
-     * @return \Cake\ORM\Association
+     * @return $this
      */
-    public function setConditions($conditions)
+    public function setConditions(Closure|array $conditions)
     {
         $this->_conditions = $conditions;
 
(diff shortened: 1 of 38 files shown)


othercorey:
Several suppressions due to foreign key being false.

src/ORM/ResultSet.php
```diff
@@ -185,8 +184,7 @@ public function __construct(Query $query, StatementInterface $statement)
      *
      * @return object|array
      */
-    #[ReturnTypeWillChange]
-    public function current()
+    public function current(): object|array
```
ADmad, line 187: Can

othercorey: Can non-entity result sets be used in plugins? I don't think so, right?

ADmad: I haven't seen resultsets having anything besides entity or array.

src/ORM/Association.php
```diff
@@ -893,7 +894,7 @@ public function exists($conditions): bool
      * @see \Cake\ORM\Table::updateAll()
      * @return int Count Returns the affected rows.
      */
-    public function updateAll(array $fields, $conditions): int
+    public function updateAll(array $fields, mixed $conditions): int
```
ADmad, line 897: Are the types for types for both params the same as for Table::updateAll()?

src/ORM/Association/BelongsToMany.php
```diff
@@ -218,7 +216,7 @@ public function getForeignKey()
      * @param mixed $sort A find() compatible order clause
      * @return $this
      */
-    public function setSort($sort)
+    public function setSort(mixed $sort)
```
ADmad, line 219: This should probably match Query::order()?

othercorey: Updated setSort() and getSort(). getSort() allows returning null.

src/ORM/Association.php
othercorey: Updated updateAll() and deleteAll().

src/ORM/Association.php
```diff
@@ -424,9 +426,9 @@ public function getTarget(): Table
      *
      * @param \Closure|array $conditions list of conditions to be used
      * @see \Cake\Database\Query::where() for examples on the format of the array
-     * @return \Cake\ORM\Association
+     * @return $this
      */
-    public function setConditions($conditions)
+    public function setConditions(Closure|array $conditions)
```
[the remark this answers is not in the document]
othercorey: We really don't test all of these pass-through signatures.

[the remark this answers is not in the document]
ADmad: I am concerned that since these values get simply passed through to the corresponding query methods, people might be using values that are valid for the query but it's type not explicitly listed in the association methods.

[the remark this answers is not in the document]
othercorey: I meant that I agree. We don't seem to test all the paths to the query methods.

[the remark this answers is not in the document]
othercorey: Updated Association::setConditions() and BelongsToMany::setConditions()

PR #15750, opened 2021-08-18
Fix issues reported by static analyzers.

ADmad:

### DIFF
--- a/src/Http/CallbackStream.php
+++ b/src/Http/CallbackStream.php
@@ -40,7 +40,8 @@ class CallbackStream extends BaseCallbackStream
     {
         $callback = $this->detach();
         $result = '';
-        if (is_callable($callback)) {
+        /** @psalm-suppress TypeDoesNotContainType */
+        if ($callback !== null) {
             $result = $callback();
         }
         if (!is_string($result)) {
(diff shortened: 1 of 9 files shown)


src/Http/CallbackStream.php
```diff
@@ -40,7 +40,8 @@ public function getContents(): string
     {
         $callback = $this->detach();
         $result = '';
-        if (is_callable($callback)) {
+        /** @psalm-suppress TypeDoesNotContainType */
+        if ($callback !== null) {
```
othercorey, line 44: Why is this needed?

ADmad: The return type of `detach()` is `?callable` so checking `!== null` is cheaper hence I made the change. Pslam gave some silly error like `Type null for $callback is always null` which I ignored.

src/I18n/DateFormatTrait.php
```diff
@@ -251,7 +251,7 @@ protected function _formatObject($date, $format, ?string $locale): string
                 $calendar,
                 $pattern
             );
-            if ($formatter === false) {
+            if (empty($formatter)) {
```
ADmad, line 254: psalm says return value of `datefmt_create()` can be false while phpstan says it can be `null`. It seems the PHP doc was updated at some point (or changed for PHP 8). Regardless, using `empty()` caters for both types.

src/Http/CallbackStream.php
othercorey: It probably doesn't consider callable a native type and discards it. Probably a bug.

PR #15752, opened 2021-08-18
Update phpunit's config file.

ADmad:
Replace deprecated `<filter>` tag with `<coverage>`.


phpunit.xml.dist
```diff
@@ -18,24 +17,24 @@
...
-    </filter>
+        </include>
+        <exclude>
+            <!--
+            This file contains a few functions that cannot be tested
+            as they contain die; or breakpoint functionality.
+            -->
+            <file>src/basics.php</file>
+            <!-- This file is deprecated and unused -->
+            <file>src/Core/ClassLoader.php</file>
```
markstory, line 35: Wasn't this file removed in 5.x?

ADmad: Yup, removed.

PR #15754, opened 2021-08-20
Call parent constructor of Constraint

frankvanhest:
Good morning,

For example when you use `assertContentType` and it fails, you'll get the error `Call to a member function export() on null` in the file `vendor/phpunit/phpunit/src/Framework/Constraint/Constraint.php` in the function `failureDescription`. This is because the constructor is not called and the exporter property is not set. By calling the Constraint constructor in the ResponseBase, this issue is resolved.

If further changes are needed, please let me know. 

Cheers,

Frank


frankvanhest:
I'm sorry for this MR. The project I'm working with, still uses phpunit 6.5. In the new version this issue doen't apply anymore.

PR #15756, opened 2021-08-20
Use hash_equals in BlowfishPasswordHasher to mitigate timing attacks

JakubOnderka:
hash_equals is available since PHP 5.6, but CakePHP 2.x still supports PHP 5.3, so we have to first check if this function exists and fallback to `===` otherwise.


lib/Cake/Controller/Component/Auth/BlowfishPasswordHasher.php
```diff
@@ -42,7 +42,12 @@ public function hash($password) {
  * @return bool True if hashes match else false.
  */
 	public function check($password, $hashedPassword) {
-		return $hashedPassword === Security::hash($password, 'blowfish', $hashedPassword);
+		if (function_exists('hash_equals')) {
+			// Use hash_equals to mitigate timing attacks
+			return hash_equals($hashedPassword, Security::hash($password, 'blowfish', $hashedPassword));
+		} else {
```
dereuromark, line 48: early return doesnt need else case.

markstory:
Thanks for taking the time to do this. Unfortunately, 2.x is end of life and no longer receiving any fixes.

PR #15758, opened 2021-08-21
cake 5: Add native type hints for I18n

othercorey:


othercorey:
The only type changes I made were adding `int` to the Number formatting functions.

src/I18n/Date.php
```diff
@@ -131,8 +134,10 @@ class Date extends MutableDate implements I18nDateTimeInterface
      * @param \DateTimeZone|string|null $tz The timezone in which the date is taken.
      *                                  Ignored if `$time` is a DateTimeInterface instance.
      */
-    public function __construct($time = 'now', $tz = null)
-    {
+    public function __construct(
+        DateTime|DateTimeImmutable|string|int|null $time = 'now',
+        DateTimeZone|string|null $tz = null
+    ) {
         deprecationWarning(
```
ADmad, line 141: I thought I cleaned up these deprecation. Probably I have them locally and forgot to make a PR.

ADmad: Actually it's because https://github.com/cakephp/cakephp/pull/15744 hasn't been merged yet and now again has conflicts 😞.

othercorey: I would wait for the View PR to merge as well then we can resolve this.

PR #15759, opened 2021-08-21
cake 5: Add native type hints for View

othercorey:


src/View/StringTemplate.php
```diff
@@ -339,13 +340,16 @@ protected function _formatAttribute(string $key, $value, $escape = true): string
...
-     * @return array<string>|string
+     * @return array<string>|string|null
      */
-    public function addClass($input, $newClass, string $useIndex = 'class')
-    {
+    public function addClass(
+        stdClass|array|string|bool|null $input,
+        array|string|bool|null $newClass,
+        string $useIndex = 'class'
+    ): array|string|null {
```
othercorey, line 352: I just added all the types that were being passed in. It requires the suppression below because $newClass might not be array ... due to bool, I think. Not sure if if `empty($newClass)` is ever false.

src/View/Widget/DateTimeWidget.php
```diff
@@ -118,7 +119,7 @@ public function render(array $data, ContextInterface $context): string
 
         $data = $this->setStep($data, $context, $data['fieldName'] ?? '');
 
-        $data['value'] = $this->formatDateTime($data['val'], $data);
+        $data['value'] = $this->formatDateTime($data['val'] === true ? new DateTimeImmutable() : $data['val'], $data);
```
othercorey, line 122: Set the default value instead of passing `true` to `formatDateTime()`

src/View/Widget/RadioWidget.php
```diff
@@ -235,14 +235,19 @@ protected function _renderInput($val, $text, $data, $context): string
...
      */
-    protected function _renderLabel(array $radio, $label, $input, $context, $escape)
-    {
+    protected function _renderLabel(
+        array $radio,
+        array|string|bool|null $label,
+        string $input,
+        ContextInterface $context,
+        bool $escape
+    ): string|false {
```
othercorey, line 250: Seems that `true` and `null` are passed in.

src/View/Widget/SelectBoxWidget.php
```diff
@@ -201,18 +199,18 @@ protected function _emptyValue($value): array
...
-        $optgroup,
+        ArrayAccess|array $optgroup,
         ?array $disabled,
-        $selected,
-        $templateVars,
-        $escape
+        array|string|bool|null $selected,
+        array $templateVars,
+        bool $escape
     ): string {
```
othercorey, line 214: Seems that `bool` is passed to `$selected`

src/View/Widget/SelectBoxWidget.php
```diff
@@ -238,13 +236,18 @@ protected function _renderOptgroup(
...
-    protected function _renderOptions(iterable $options, ?array $disabled, $selected, $templateVars, $escape): array
-    {
+    protected function _renderOptions(
+        iterable $options,
+        ?array $disabled,
+        array|string|int|bool|null $selected,
+        array $templateVars,
+        bool $escape
+    ): array {
         $out = [];
```
othercorey, line 251: Same here.

src/View/StringTemplate.php
```diff
@@ -309,11 +310,11 @@ public function formatAttributes(?array $options, ?array $exclude = null): strin
      * Works with minimized attributes that have the same value as their name such as 'disabled' and 'checked'
      *
      * @param string $key The name of the attribute to create
-     * @param array<string>|string $value The value of the attribute to create.
+     * @param array<string>|string|float|int|bool $value The value of the attribute to create.
      * @param bool $escape Define if the value must be escaped
      * @return string The composed attribute.
      */
-    protected function _formatAttribute(string $key, $value, $escape = true): string
+    protected function _formatAttribute(string $key, array|string|float|int|bool $value, bool $escape = true): string
```
othercorey, line 317: Added all the types that were passed in.

src/View/StringTemplate.php
```diff
@@ -339,13 +340,16 @@ protected function _formatAttribute(string $key, $value, $escape = true): string
...
+     * @param \stdClass|array|string|bool|null $input The array or string to add the class to
+     * @param array|string|bool|null $newClass the new class or classes to add
      * @param string $useIndex if you are inputting an array with an element other than default of 'class'.
-     * @return array<string>|string
+     * @return array<string>|string|null
      */
-    public function addClass($input, $newClass, string $useIndex = 'class')
-    {
+    public function addClass(
+        stdClass|array|string|bool|null $input,
```
ADmad, line 349: LOL.. How does this get `stdClass`?

tests/TestCase/View/Helper/FormHelperTest.php
```diff
@@ -2211,7 +2200,7 @@ public function testFormSecuredAndDisabledNotAssoc(): void
         $this->Form->text('Model.text', ['disabled']);
         $this->Form->textarea('Model.textarea', ['disabled']);
         $this->Form->password('Model.password', ['disabled']);
-        $this->Form->radio('Model.radio', [1, 2], ['disabled']);
```
ADmad, line 2214: This can very well be an array of ids, hence ints.

src/View/StringTemplate.php
othercorey: An explicit test: https://github.com/cakephp/cakephp/blob/5.0/tests/TestCase/View/StringTemplateTest.php#L332-L333

ADmad: That's to test passing an object, not stdClass specifically. So I think we should just use `mixed` here.

tests/TestCase/View/Helper/FormHelperTest.php
othercorey: I don't see how integers can be translated into attributes. I assumed this was just a lazy test to "disable" an attribute which is an array key.

othercorey: Updated RadioWidget::_renderInput() to accept `int` for `$text`

src/View/Widget/WidgetLocator.php
```diff
@@ -185,7 +185,7 @@ public function clear(): void
      * @return \Cake\View\Widget\WidgetInterface Widget instance.
      * @throws \ReflectionException
      */
-    protected function _resolveWidget($config): WidgetInterface
+    protected function _resolveWidget(mixed $config): WidgetInterface
```
ADmad, line 188:
```suggestion
    protected function _resolveWidget(string|array $config): WidgetInterface
```

Then the exception with msg `'Widget config must be a string or array.'` can be removed.

src/View/StringTemplate.php
markstory: Couldn't it be false if an empty string or `null` is passed?

src/View/StringTemplate.php
ADmad: There's a failing test since it receives a `FrozenTime` instance too as value. I think the type should be just set to `mixed` since it can accept any scalar or stringable value.

src/View/StringTemplate.php
othercorey: I'm just saying, the original cod doesn't detect or handle a false where I added the suppression. Not sure if it's ever encountered.

PR #15761, opened 2021-08-21
ARIA attributes

a-d-w-s:
ARIA attributes FormHelper.php

### DIFF
--- a/src/View/Helper/FormHelper.php
+++ b/src/View/Helper/FormHelper.php
@@ -768,7 +768,7 @@ class FormHelper extends Helper
 
         return $this->formatTemplate('error', [
             'content' => $error,
-            'id' => $field . '-aria',
+            'id' => $field . '-error',
         ]);
     }
 
@@ -1056,7 +1056,7 @@ class FormHelper extends Helper
                 'aria-label' => !empty($options['label']) ? $options['label'] : $options['placeholder'],
                 'aria-required' => $options['required'] == true ? 'true' : null,
                 'aria-invalid' => $isFieldError ? 'true' : null,
-                'aria-describedby' => $isFieldError ? $options['id'] . '-aria' : null,
+                'aria-describedby' => $isFieldError ? $options['id'] . '-error' : null,
             ];
         }
 
--- a/src/View/Helper/FormHelper.php
+++ b/src/View/Helper/FormHelper.php
@@ -1047,10 +1047,18 @@ class FormHelper extends Helper
         $options = $this->_parseOptions($fieldName, $options);
         $options += ['id' => $this->_domId($fieldName)];
 
-        if ($options['type'] != 'hidden') {
+        // Hidden inputs don't need aria.
+        // Multiple checkboxes can't have aria generated for them at this layer.
+        if ($options['type'] !== 'hidden' && ($options['type'] !== 'select' && !isset($options['multiple']))) {
             $isFieldError = $this->isFieldError($fieldName);
+            $labelText = null;
+            if (isset($options['label']) && is_string($options['label'])) {
+                $labelText = $options['label'];
+            } elseif (isset($options['placeholder'])) {
+                $labelText = $options['placeholder'];
+            }
             $options += [
-                'aria-label' => isset($options['label']) && is_string($options['label']) ? $options['label'] : $options['placeholder'],
+                'aria-label' => $labelText,
                 'aria-required' => $options['required'] == true ? 'true' : null,
                 'aria-invalid' => $isFieldError ? 'true' : null,
                 'aria-describedby' => $isFieldError ? $this->_domId($fieldName) . '-error' : null,
--- a/src/View/Helper/FormHelper.php
+++ b/src/View/Helper/FormHelper.php
@@ -768,7 +768,7 @@ class FormHelper extends Helper
 
         return $this->formatTemplate('error', [
             'content' => $error,
-            'id' => $field . '-error',
+            'id' => $this->_domId($field) . '-error',
         ]);
     }
 
@@ -1056,7 +1056,7 @@ class FormHelper extends Helper
                 'aria-label' => !empty($options['label']) ? $options['label'] : $options['placeholder'],
                 'aria-required' => $options['required'] == true ? 'true' : null,
                 'aria-invalid' => $isFieldError ? 'true' : null,
-                'aria-describedby' => $isFieldError ? $options['id'] . '-error' : null,
+                'aria-describedby' => $isFieldError ? $this->_domId($fieldName) . '-error' : null,
             ];
         }
 
--- a/src/View/Helper/FormHelper.php
+++ b/src/View/Helper/FormHelper.php
@@ -1047,24 +1047,6 @@ class FormHelper extends Helper
         $options = $this->_parseOptions($fieldName, $options);
         $options += ['id' => $this->_domId($fieldName)];
 
-        // Hidden inputs don't need aria.
-        // Multiple checkboxes can't have aria generated for them at this layer.
-        if ($options['type'] !== 'hidden' && ($options['type'] !== 'select' && !isset($options['multiple']))) {
-            $isFieldError = $this->isFieldError($fieldName);
-            $labelText = null;
-            if (isset($options['label']) && is_string($options['label'])) {
-                $labelText = $options['label'];
-            } elseif (isset($options['placeholder'])) {
-                $labelText = $options['placeholder'];
-            }
-            $options += [
-                'aria-label' => $labelText,
-                'aria-required' => $options['required'] == true ? 'true' : null,
-                'aria-invalid' => $isFieldError ? 'true' : null,
-                'aria-describedby' => $isFieldError ? $this->_domId($fieldName) . '-error' : null,
-            ];
-        }
-
         $templater = $this->templater();
         $newTemplates = $options['templates'];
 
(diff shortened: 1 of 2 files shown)


src/View/Helper/FormHelper.php
```diff
@@ -766,7 +766,10 @@ public function error(string $field, $text = null, array $options = []): string
             }
         }
 
-        return $this->formatTemplate('error', ['content' => $error]);
+        return $this->formatTemplate('error', [
+            'content' => $error,
+            'id' => $field . '-aria',
```
ndm2, line 771: This value needs to be passed through `FormHelper::_domId()`, same for the `aria-describedby` option value.

And maybe this should have a more generic suffix, so that it doesn't indicate such a strictly tied purpose.

src/View/Helper/FormHelper.php
```diff
@@ -1046,6 +1049,10 @@ public function control(string $fieldName, array $options = []): string
             'templates' => [],
             'templateVars' => [],
             'labelOptions' => true,
+            'aria-label' => empty($options['placeholder']) ? $options['label'] : $options['placeholder'],
+            'aria-required' => $options['required'] == true ? 'true' : null,
+            'aria-invalid' => $this->isFieldError($fieldName) ? 'true' : null,
+            'aria-describedby' => $this->isFieldError($fieldName) ? $fieldName . '-aria' : null,
```
ndm2, line 1055: Generally, do these ARIA attributes maybe need to _not_ apply to hidden inputs?

markstory: I don't think they're required or useful for hidden inputs. Aria agents will skip hidden inputs by default.

shadowx-jb: You're right, I didn't think of hidden fields. It probably shouldn't matter, but I'd rather do it so that ARIA isn't in the hidden fields.

ADmad:
These changes need to target `4.next` instead of `master`.

src/View/Helper/FormHelper.php
shadowx-jb: I modified the condition, now ARIA will be inserted into everything outside the hidden field and if it has a label or placeholder attribute.

src/View/Helper/FormHelper.php
shadowx-jb: I don't understand how I should do it? For ARIA it is necessary that the element has an ID, and it must have some prefix - because the text field itself has an ID. So if they had the same ID, it would not be valid according to W3C.

shadowx-jb:
> 
> 
> These changes need to target `4.next` instead of `master`.

shadowx-jb:
I rebase to `4.next` and push

src/View/Helper/FormHelper.php
ndm2: Since IDs aren't something specifically related to ARIA, I'd suggest to not use `-aria` as a suffix, but something more generic that fits what the element represents, like `-error`.

ndm2:
> I rebase to `4.next` and push

You also need to change the target branch of the PR so that it points to `4.next` (click on "edit" in the top right, then you can change the branch).

src/View/Helper/FormHelper.php
```diff
@@ -1044,6 +1047,19 @@ public function control(string $fieldName, array $options = []): string
...
+        if (
+            !empty($options['label']) && $options['type'] != 'hidden' ||
+            !empty($options['placeholder']) && $options['type'] != 'hidden'
+        ) {
+            $isFieldError = $this->isFieldError($fieldName);
+            $options += [
+                'aria-label' => !empty($options['label']) ? $options['label'] : $options['placeholder'],
+                'aria-required' => $options['required'] == true ? 'true' : null,
+                'aria-invalid' => $isFieldError ? 'true' : null,
+                'aria-describedby' => $isFieldError ? $options['id'] . '-error' : null,
```
ndm2, line 1059: Using `$options['id']` won't work, `id` has already been passed through `_domId()`, which transforms the value to include only valid characters, and also might prepend a custom prefix, so it could be different from what is being set in the `error()` method.

You'll need to explicitly pass the field name + suffix through `_domId()`, and do the same in the `error()` method so that both values match.

shadowx-jb: OK, thank You

src/View/Helper/FormHelper.php
```diff
@@ -1044,6 +1047,19 @@ public function control(string $fieldName, array $options = []): string
         $options = $this->_parseOptions($fieldName, $options);
         $options += ['id' => $this->_domId($fieldName)];
 
+        if (
+            !empty($options['label']) && $options['type'] != 'hidden' ||
+            !empty($options['placeholder']) && $options['type'] != 'hidden'
+        ) {
```
markstory, line 1053: These empty() calls could eat `'0'` which *could* be a label or placeholder. Doing a comparison on the input type should be enough as an empty aria-label won't hurt.

```suggestion
        if ($options['type'] != 'hidden') {
```

src/View/Helper/FormHelper.php
```diff
@@ -1044,6 +1047,19 @@ public function control(string $fieldName, array $options = []): string
...
+            $isFieldError = $this->isFieldError($fieldName);
+            $options += [
+                'aria-label' => !empty($options['label']) ? $options['label'] : $options['placeholder'],
+                'aria-required' => $options['required'] == true ? 'true' : null,
+                'aria-invalid' => $isFieldError ? 'true' : null,
```
markstory, line 1058:
```suggestion
                'aria-invalid' => $isFieldError ? true : null,
```

You should be able to use a boolean here. If not we can fix up the attribute generation code.

src/View/Helper/FormHelper.php
```diff
@@ -1044,6 +1047,19 @@ public function control(string $fieldName, array $options = []): string
         $options = $this->_parseOptions($fieldName, $options);
         $options += ['id' => $this->_domId($fieldName)];
 
+        if (
+            !empty($options['label']) && $options['type'] != 'hidden' ||
+            !empty($options['placeholder']) && $options['type'] != 'hidden'
+        ) {
+            $isFieldError = $this->isFieldError($fieldName);
+            $options += [
+                'aria-label' => !empty($options['label']) ? $options['label'] : $options['placeholder'],
```
markstory, line 1056: The `label` option can also be an array. We'll need to work around that.
```suggestion
                'aria-label' => isset($options['label']) && is_string($options['label']) ? $options['label'] : $options['placeholder'],
```

markstory (commented):
Looking good. I'll add some tests in a day or so.

src/View/Helper/FormHelper.php
shadowx-jb: It should be true or false:
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-invalid_attribute

shadowx-jb:
Question: if we have one error, it writes the text to the div element. If we have more errors, then we write to the div element UL LI. For a well-styled voice - for example, adding an icon in front of the text in case of one error - I would suggest wrapping one error in the p element, which is inserted into the div element. Example: 

```
            } else {
                $error = array_pop($error);
            }
```

New 

```
            } else {
                $error = $this->Html->para(null, array_pop($error));
            }
```

markstory:
> I would suggest wrapping one error in the p element, which is inserted into the div element. Example:

While I agree with you, this is more complicated for us to change at the framework level as switching tags out has a higher chance of breaking userland applications as they could be relying on specific tag `names` in their CSS or JavaScript code.

tests/TestCase/View/Helper/FormHelperTest.php
```diff
@@ -2433,8 +2437,13 @@ public function testErrorMessageDisplay(): void
...
             'input' => [
-                'type' => 'text', 'name' => 'Article[title]',
-                'id' => 'article-title', 'class' => 'form-error',
+                'type' => 'text',
+                'name' => 'Article[title]',
+                'id' => 'article-title',
+                'class' => 'form-error',
+                'aria-invalid' => 'true',
+                // TODO this seems wrong as the id doesn't exist.
+                'aria-describedby' => 'article-title-error',
```
markstory, line 2446: There are a few tests I marked with TODO as the results look off to me.

ndm2: The `control()` call should probably disable the attribute, ie pass `'aria-describedby' => null`.

tests/TestCase/View/Helper/FormHelperTest.php
```diff
@@ -2911,8 +2964,13 @@ public function testControlCustomization(): void
...
             'input' => [
-                'type' => 'text', 'name' => 'field',
-                'id' => 'field', 'class' => 'form-error',
+                'type' => 'text',
+                'name' => 'field',
+                'id' => 'field',
+                'class' => 'form-error',
+                'aria-invalid' => 'true',
+                // TODO this seems bad as the id doesn't exist.
+                'aria-describedby' => 'field-error',
```
ndm2, line 2973: Maybe hardcode the `id` attribute in the custom `error` template? The `aria-describedby` option could receive a custom value too, after all that's something that should be tested anyways.

markstory:
I've fixed the problems with `aria-describedby` and removed the `aria-label` attribute. The `aria-label` attribute was only set when an explicit label string was set which might not be very often. Furthermore, the aria-label attribute is redundant when we also have id attributes on inputs and labels with correct `for` attributes set.

shadowx-jb:
`aria-label` should be, if I have a `placeholder`, for example ... blind people then don't know what kind of field it is. I use a field, without a label and I have the field title in the placeholder.

markstory:
>  I use a field, without a label and I have the field title in the placeholder.

Hasn't the developer gone out of their way to make accessibility bad by not providing a label though? I don't know if the framework should be trying to compensate for that.

shadowx-jb:
Accessibility for the blind is that the label is in the `aria-label`. For others, he is a `placeholder `.

markstory:
So would you recommend adding an `aria-label` attribute whenever the `placeholder` is set or only if the `placeholder` is defined and the label has been disabled?

shadowx-jb:
Yes, if label disabled and placeholder are defined, I would set the aria-label field.

PR #15765, opened 2021-08-23
Allow json_encode options in JsonType

Erwane:
Add possibility to pass `json_encode()` options to database JsonType.
This can be useful in total UTF-8 context to prevent extra characters in database.

Changing the `options` in existing project don't break old data.

JsonType documentation : 
```
// Don't escape unicode and slashes
TypeFactory::build('json')->setEncodingOptions(JSON_UNESCAPED_UNICODE | JSON_UNESCAPED_SLASHES);
```

### DIFF
--- a/src/Database/Type/JsonType.php
+++ b/src/Database/Type/JsonType.php
@@ -112,10 +112,10 @@ class JsonType extends BaseType implements BatchCastingInterface
      * Set json_encode options.
      *
      * @param  int $jsonOptions Options. Use JSON_* flags. Set `0` to reset.
-     * @return self
+     * @return $this
      * @see https://www.php.net/manual/en/function.json-encode.php
      */
-    public function setJsonOptions(int $jsonOptions): self
+    public function setJsonOptions(int $jsonOptions)
     {
         $this->_jsonOptions = $jsonOptions;
 
--- a/src/Database/Type/JsonType.php
+++ b/src/Database/Type/JsonType.php
@@ -28,6 +28,11 @@ use PDO;
 class JsonType extends BaseType implements BatchCastingInterface
 {
     /**
+     * @var int
+     */
+    protected $_jsonOptions = 0;
+
+    /**
      * Convert a value data into a JSON string
      *
      * @param mixed $value The value to convert.
@@ -45,7 +50,7 @@ class JsonType extends BaseType implements BatchCastingInterface
             return null;
         }
 
-        return json_encode($value);
+        return json_encode($value, $this->_jsonOptions);
     }
 
     /**
@@ -101,5 +106,19 @@ class JsonType extends BaseType implements BatchCastingInterface
     public function marshal($value)
     {
         return $value;
+    }
+
+    /**
+     * Set json_encode options.
+     *
+     * @param  int|null $jsonOptions Options. Use JSON_* flags
+     * @return self
+     * @see https://www.php.net/manual/en/function.json-encode.php
+     */
+    public function setJsonOptions(int $jsonOptions): self
+    {
+        $this->_jsonOptions = $jsonOptions;
+
+        return $this;
     }
 }
--- a/src/Database/Type/JsonType.php
+++ b/src/Database/Type/JsonType.php
@@ -30,7 +30,7 @@ class JsonType extends BaseType implements BatchCastingInterface
     /**
      * @var int
      */
-    protected $_jsonOptions = 0;
+    protected $_encodingOptions = 0;
 
     /**
      * Convert a value data into a JSON string
@@ -50,7 +50,7 @@ class JsonType extends BaseType implements BatchCastingInterface
             return null;
         }
 
-        return json_encode($value, $this->_jsonOptions);
+        return json_encode($value, $this->_encodingOptions);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


ADmad:
New features need to target `4.next` instead of `master`.

src/Database/Type/JsonType.php
```diff
@@ -102,4 +107,18 @@ public function marshal($value)
     {
         return $value;
     }
+
+    /**
+     * Set json_encode options.
+     *
+     * @param  int $jsonOptions Options. Use JSON_* flags. Set `0` to reset.
+     * @return self
```
ADmad, line 115:
```suggestion
     * @return $this
```

src/Database/Type/JsonType.php
```diff
@@ -102,4 +107,18 @@ public function marshal($value)
...
+     * @see https://www.php.net/manual/en/function.json-encode.php
+     */
+    public function setJsonOptions(int $jsonOptions): self
```
ADmad, line 118:
```suggestion
    public function setJsonOptions(int $jsonOptions)
```

src/Database/Type/JsonType.php
```diff
@@ -102,4 +107,18 @@ public function marshal($value)
     {
         return $value;
     }
+
+    /**
+     * Set json_encode options.
+     *
+     * @param  int $jsonOptions Options. Use JSON_* flags. Set `0` to reset.
```
ADmad, line 114:
```suggestion
     * @param int $jsonOptions Options. Use JSON_* flags. Set `0` to reset.
```

ADmad (approved):

tests/TestCase/Database/Type/JsonTypeTest.php
```diff
@@ -133,4 +134,19 @@ public function testToStatement()
...
+    /**
+     * Test instance options
+     *
+     * @return void
+     */
+    public function testInstanceOptions()
+    {
+        // New instance to prevent others tests breaking
+        $instance = new JsonType();
+        $instance->setJsonOptions(JSON_UNESCAPED_UNICODE + JSON_UNESCAPED_SLASHES);
```
garas, line 147:
```suggestion
        $instance->setJsonOptions(JSON_UNESCAPED_UNICODE | JSON_UNESCAPED_SLASHES);
```

Binary flags are usually combined with bitwise OR.

Erwane: Semantically, `+` is more readable for me (adding options to others), BUT, i will change ;)

src/Database/Type/JsonType.php
```diff
@@ -102,4 +107,18 @@ public function marshal($value)
...
     }
+
+    /**
+     * Set json_encode options.
+     *
+     * @param int $jsonOptions Options. Use JSON_* flags. Set `0` to reset.
+     * @return $this
+     * @see https://www.php.net/manual/en/function.json-encode.php
+     */
+    public function setJsonOptions(int $jsonOptions)
```
othercorey, line 118: Probably not a big issue, but `setJsonEncodeOptions()` might be clearer since it's only for json_encode.

tests/TestCase/Database/Type/JsonTypeTest.php
```diff
@@ -133,4 +134,19 @@ public function testToStatement()
...
+    /**
+     * Test instance options
+     *
+     * @return void
+     */
+    public function testInstanceOptions()
+    {
+        // New instance to prevent others tests breaking
+        $instance = new JsonType();
+        $instance->setJsonOptions(JSON_UNESCAPED_UNICODE | JSON_UNESCAPED_SLASHES);
```
othercorey, line 147: I wonder if JSON_UNESCAPED_UNICODE should be the default.

src/Database/Type/JsonType.php
dereuromark: Or setEncodingOptions()

tests/TestCase/Database/Type/JsonTypeTest.php
dereuromark: Maybe in 5.x

src/Database/Type/JsonType.php
Erwane: setEncodingFlags() ?

src/Database/Type/JsonType.php
Erwane: About $this. PhpStorm recommend `@return self` and `: self` as return type.

```php
     * @return self
     * @see https://www.php.net/manual/en/function.json-encode.php
     */
    public function setEncodingOptions(int $jsonOptions): self
```
With this, there is no warning on method.

src/Database/Type/JsonType.php
ADmad: We don't have any methods with "Flags" in the name, we just using "Options" so `setEncodingOptions()` sounds good.

dereuromark: After all a flag is just a technical detail of an option :) And we try to to hide certain implementation of language details behind a more common speaking term.

Erwane:
@ADmad , could you check https://github.com/cakephp/cakephp/pull/15765#discussion_r694562153 please ?

src/Database/Type/JsonType.php
dereuromark: This is for PHP 8+, for lower PHP versions we still don't do this for several reasons, including chaining in the IDE
The CS sniffer rule here prevents any mistakes and issues, so the type would be completely necessary here anyways.

dereuromark (approved):

src/Database/Type/JsonType.php
ndm2: For a long time PhpStorm totally sperged out on `self`, it's even only working properly in PHP since PHP 7.4.

Depending on the state of other editors, maybe this can be changed with CakePHP 5, but for 4.x I'd recommend staying with `$this` in order to not break peoples experience now with slightly older editors.

dereuromark: True, and Cake4 is still PHP7.2+ (I strongly voted for PHP7.3+ for 4.3 but that still wouldn't change anything^^).

ndm2: Oh, for some reason I actually thought 4.x it was 7.4+, but yeah, with 7.2+ it's out of question as it's not compatible on language level.

PR #15766, opened 2021-08-24
cake 5: Add native type hints for Log

othercorey:
@ADmad @markstory Also forgot to push Log.

### DIFF
--- a/src/Log/Log.php
+++ b/src/Log/Log.php
@@ -17,6 +17,7 @@ namespace Cake\Log;
 
 use Cake\Core\StaticConfigTrait;
 use Cake\Log\Engine\BaseLog;
+use Closure;
 use InvalidArgumentException;
 use Psr\Log\LoggerInterface;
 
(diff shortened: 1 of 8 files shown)


src/Log/Log.php
```diff
@@ -271,11 +272,11 @@ public static function levels(): array
      * ```
      *
      * @param array|string $key The name of the logger config, or an array of multiple configs.
-     * @param array|null $config An array of name => config data for adapter.
+     * @param \Psr\Log\LoggerInterface|\Closure|array|null $config An array of name => config data for adapter.
      * @return void
      * @throws \BadMethodCallException When trying to modify an existing config.
      */
-    public static function setConfig($key, $config = null): void
+    public static function setConfig(array|string $key, LoggerInterface|Closure|array|null $config = null): void
```
othercorey, line 279: I think `LoggerInterface` is the correct base class here. Wasn't sure if it had to be a `BaseLog` instance, but I don't think so.

ADmad:
`psr/log` has new v3 release with types. Might be a good idea to update it too to avoid potentially having to change the types again.

markstory (approved):

othercorey:
> `psr/log` has a new v3 release with types. Might be a good idea to update it too to avoid potentially having to change the types again.

I can only find a tag and no release notes. What's the difference between v2 and v3?

ADmad:
> What's the difference between v2 and v3?

The addition of argument and return types (what you are doing currently :slightly_smiling_face:).

PR #15767, opened 2021-08-24
Update psr/log to v3.

ADmad:


src/Log/Engine/SyslogLog.php
```diff
@@ -92,12 +93,12 @@ class SyslogLog extends BaseLog
      * log messages, pass all messages through the format defined in the configuration
      *
      * @param mixed $level The severity level of log you are making.
-     * @param string $message The message you want to log.
+     * @param \Stringable|string  $message The message you want to log.
```
dereuromark, line 96:
```suggestion
     * @param \Stringable|string $message The message you want to log.
```

othercorey:
Thanks for cleaning up _format().

PR #15769, opened 2021-08-24
cake 5: Add native type hints for Controller properties

othercorey:
Seems like a reasonable place to add them.

### DIFF
--- a/src/Controller/Controller.php
+++ b/src/Controller/Controller.php
@@ -106,7 +106,7 @@ class Controller implements EventListenerInterface, EventDispatcherInterface
      *
      * @var string
      */
-    protected $name;
+    protected string $name = '';
 
     /**
      * An instance of a \Cake\Http\ServerRequest object that contains information about the current request.
@@ -116,7 +116,7 @@ class Controller implements EventListenerInterface, EventDispatcherInterface
      * @var \Cake\Http\ServerRequest
      * @link https://book.cakephp.org/4/en/controllers/request-response.html#request
      */
-    protected $request;
+    protected ServerRequest $request;
 
     /**
      * An instance of a Response object that contains information about the impending response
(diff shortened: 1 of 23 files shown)


src/Controller/Controller.php
```diff
@@ -106,7 +106,7 @@ class Controller implements EventListenerInterface, EventDispatcherInterface
      *
      * @var string
      */
-    protected $name;
+    protected string $name = '';
```
othercorey, line 109: This is to avoid a `?string` type because psalm complains when you check if initialized with `isset()`.

ADmad:
There are a lot of places where FQCN has been used for the type which need to be replaced with just class/interface name along with `use` statements.

othercorey:
> There are a lot of places where FQCN has been used for the type which need to be replaced with just class/interface name along with `use` statements.

I forgot to enable this sniff. Should we enable it by default?

othercorey:
Fixed type hint namespaces.

src/Controller/Controller.php
```diff
@@ -196,11 +196,14 @@ public function __construct(
     ) {
         if ($name !== null) {
             $this->name = $name;
-        } elseif ($this->name === null && $request) {
-            $this->name = $request->getParam('controller');
+        } elseif (!$this->name && $request) {
+            $controller = $request->getParam('controller');
+            if ($controller) {
+                $this->name = $request->getParam('controller');
```
markstory, line 202:
```suggestion
                $this->name = $controller;
```

othercorey: whoops, too many undo/redo.

tests/test_app/TestApp/Controller/ViewPostsController.php
```diff
@@ -24,7 +24,7 @@ class ViewPostsController extends Controller
      *
      * @var string
      */
-    public $name = 'Posts';
+    public string $name = 'Posts';
```
ADmad, line 27:
```suggestion
    protected string $name = 'Posts';
```

PR #15772, opened 2021-08-25
Remove methods from Datasource/ConnectionInterface

ADmad:
These methods only make sense for RDBMS.

### DIFF
--- a/src/TestSuite/Fixture/FixtureDataManager.php
+++ b/src/TestSuite/Fixture/FixtureDataManager.php
@@ -206,17 +206,23 @@ class FixtureDataManager extends FixtureLoader
         $dbs = $this->fixtureConnections($fixtures);
         foreach ($dbs as $connection => $fixtures) {
             $db = ConnectionManager::get($connection);
-            if ($db->getDriver() instanceof Postgres) {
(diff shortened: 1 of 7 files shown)


ADmad:
If this seems reasonable than I would also like to update `ConnectionManager` and `ConnectionRegistry` to only accept/return `ConnectionInterface` instances, unless we want to keep open the possibility for more varied connection types in the future.

src/TestSuite/Fixture/FixtureDataManager.php
```diff
@@ -205,7 +205,15 @@ protected function runOperation(array $fixtures, callable $operation): void
         $dbs = $this->fixtureConnections($fixtures);
         foreach ($dbs as $connection => $fixtures) {
             $db = ConnectionManager::get($connection);
+            if (
+                !method_exists($db, 'transactional')
+                || !method_exists($db, 'disableConstraints')
+            ) {
+                continue;
```
markstory, line 212: Should this be calling `$operation($db, $fixtures)` so that the operation still happens?

ADmad: Yup, fixed.

ADmad:
Is this good to merge?

PR #15779, opened 2021-08-26
Use instanceof Stringable check instead of method_exists().

ADmad:


markstory (commented):
This is a nice improvement :clap:

PR #15780, opened 2021-08-27
Mark stateResetStrategy and TransactionResetStrategy as experimental

othercorey:
@markstory Can we mark this as experimental for 5.0 changes?


othercorey:
I think we're doing something different.

PR #15781, opened 2021-08-27
cake 5: Re-organize fixtures into strategies

othercorey:
Re-organizes fixture management into strategies and helpers.

Each `TestCase` defines the fixture strategy it uses with `getFixtureStrategy()` which defaults to `TruncateFixtureStrategy`. The truncate strategy re-creates the original fixture behavior of truncating fixture tables after each test.

Each strategy implements `FixtureStrategyInterface` which defines the methods used by `TestCase` to setup and teardown each test. `FixtureHelper` has helpers to run callbacks per-connection as well as general solutions for inserting and truncating fixtures.

`TransactionFixtureStrategy` re-implements the experimental `TransactionResetStrategy` logic of wrapping the fixture insert in a transaction and rolling back after the test.


src/TestSuite/Fixture/FixtureLoadingTrait.php
```diff
@@ -0,0 +1,145 @@
...
+use Cake\Core\Configure;
+use Cake\Database\Connection;
+use Cake\Datasource\ConnectionManager;
+use Closure;
+use UnexpectedValueException;
+
+/**
+ * FixtureLoadingTrait
+ */
+trait FixtureLoadingTrait
```
markstory, line 28: Still not sold on this being a trait, but trait/helper object can be sorted out later.

src/TestSuite/Fixture/FixtureLoadingTrait.php
```diff
@@ -0,0 +1,145 @@
...
+     * and the second parameter is the array of fixtures using
+     * that connection.
+     *
+     * @param \Closure $callback Callback run per connection
+     * @param array<\Cake\Datasource\FixtureInterface> $fixtures Test fixtures
+     * @return void
+     */
+    protected function runPerConnection(Closure $callback, array $fixtures): void
+    {
+        $callback(ConnectionManager::get('test'), $fixtures);
```
markstory, line 107: Should this be operating on more than just the `test` connection?

src/TestSuite/Fixture/SchemaStrategyInterface.php
```diff
@@ -0,0 +1,30 @@
...
+ * SchemaStrategyInterface
+ */
+interface SchemaStrategyInterface
+{
+    /**
+     * Loads schema required by fixtures.
+     *
+     * @return void
+     */
+    public function load(): void;
```
markstory, line 29: For migrations would all the configuration options go into the constructor? Currently Migrations is using a `migrate()` function to do the bulk of its work, but we could change that to fit this interface instead.

I also wonder if we *need* to formalize this. My thinking with the schema loaders in 4.next was that at the end of `tests/bootstrap.php` your application's schema should be loaded into the test databases. How that happens is up to you. We provide a path for loading SQL dumps, running migrations, or creating indexes in elasticsearch but there is no formal interface.

A more formal interface could make using a variety of datasources/implementations easier to swap out but I'm not convinced that scenario would come up in practice often enough that we should be designing around it.

src/TestSuite/TestCase.php
```diff
@@ -50,20 +51,23 @@ abstract class TestCase extends BaseTestCase
...
 
+    /**
+     * @var class-string<\Cake\TestSuite\Fixture\FixtureStrategyInterface>
+     */
+    protected string $fixtureStrategyClass = TruncateFixtureStrategy::class;
+
+    /**
+     * @var class-string<\Cake\TestSuite\Fixture\SchemaStrategyInterface>
+     */
+    protected string $schemaStrategyClass = CakeSchemaStrategy::class;
```
markstory, line 69: I don't think this is a good default for applications. It is going to make a lot of work for application developers who never will want to use a big array to contain their schema definitions.

src/TestSuite/TestCase.php
```diff
@@ -249,9 +250,56 @@ protected function tearDown(): void
...
+        if (!isset(static::$schemaStrategies[$this->schemaStrategyClass])) {
+            $schemaStrategy = new $this->schemaStrategyClass();
+            if (!$schemaStrategy instanceof SchemaStrategyInterface) {
+                throw new RuntimeException(
+                    "Schema strategy `{$this->schemaStrategyClass}` does not implement `SchemaStrategyInterface`."
+                );
+            }
+            $schemaStrategy->load();
+            static::$schemaStrategies[$this->schemaStrategyClass] = $schemaStrategy;
+        }
```
markstory, line 277: Can't a similar result be achieved by loading schema in `tests/bootstrap.php`?

src/TestSuite/TestCase.php
```diff
@@ -249,9 +250,56 @@ protected function tearDown(): void
...
+
+        static::$fixtureStrategy = new $this->fixtureStrategyClass();
+        if (!static::$fixtureStrategy instanceof FixtureStrategyInterface) {
+            throw new RuntimeException(
+                "Fixture strategy `{$this->fixtureStrategyClass}` does not implement `FixtureStrategyInterface`."
+            );
+        }
+        static::$fixtureStrategy->setupTest($fixtureNames);
```
markstory, line 285: I like this but I think we might need to also have an 'aggregate' fixture strategy. An application that needs to straddle multiple datasources would need to load/reset state on all of those storage systems which is harder to do now.

src/TestSuite/TestCase.php
othercorey: That's up to you. This lets people just "set and run", but they will require setting up schema data so it isn't anywhere close to a free implementation.

I'm not sold on the schema strategy if it's not needed.

othercorey:
I don't know why the cake_sessions and sessions tables don't exist on teardown for SessionTest with sqlite.

markstory:
> I don't know why the cake_sessions and sessions tables don't exist on teardown for SessionTest with sqlite.

Could be related to `runInSeparateProcess`. I remember those tests acting funny until I added the globals check to not recreate the database when tests were run in separate process mode.

src/TestSuite/TestCase.php
```diff
@@ -249,9 +238,45 @@ protected function tearDown(): void
...
+     * @return void
+     */
+    protected function setupFixtures(): void
+    {
+        $fixtureNames = $this->getFixtures();
+        if (empty($fixtureNames)) {
+            return;
+        }
+
+        static::$fixtureStrategy = new $this->fixtureStrategyClass();
```
markstory, line 256: Should this continue to come from a protected method? Having a method to create the strategy will help users override it more easily. We wouldn't have to check the return type that way either as we could use a return type hint to enforce the interface requirement.

othercorey: Avoiding the class property would be nice.

othercorey:
@markstory ready for review

src/TestSuite/TestCase.php
```diff
@@ -241,52 +214,56 @@ protected function setUp(): void
...
+        $fixtureNames = $this->getFixtures();
+        if (empty($fixtureNames)) {
+            return;
         }
 
-        $args = func_get_args();
-        foreach ($args as $class) {
-            $this->fixtureManager->loadSingle($class, null);
-        }
+        $this->fixtureStrategy = $this->getFixtureStrategy();
```
othercorey, line 240: This uses a getter instead of a class config, but we need to cache off the strategy for the duration of the test.

src/TestSuite/Fixture/TransactionFixtureStrategy.php
```diff
@@ -0,0 +1,86 @@
...
+    /**
+     * @inheritDoc
+     */
+    public function setupTest(array $fixtureNames): void
+    {
+        $this->fixtures = $this->helper->loadFixtures($fixtureNames);
+
+        $this->helper->runPerConnection(function ($connection) {
+            if ($connection instanceof Connection) {
+                assert($connection->inTransaction() === false);
```
markstory, line 57: Should this assert() have a message?

othercorey: Added message explaining most likely a bug this is starting inside a transaction.

PR #15783, opened 2021-08-28
cake 5: Normalize passing  array to initialize() and documentation

othercorey:
This just adds a `$config` parameter to `initialize()` functions that are called by constructors with config arrays.

Updated the docblock to read the exact same in all places and explain the config array is unmerged.

### DIFF
--- a/src/View/View.php
+++ b/src/View/View.php
@@ -348,7 +348,7 @@ class View implements EventDispatcherInterface
         $this->request = $request;
         $this->response = $response ?: new Response();
         $this->Blocks = new $this->_viewBlockClass();
-        $this->initialize();
+        $this->initialize($viewOptions);
         $this->loadHelpers();
     }
 
@@ -360,9 +360,11 @@ class View implements EventDispatcherInterface
      * So this method allows you to manipulate them as required after view instance
      * is constructed.
      *
+     * @param array $config The original config array passed to constructor.
+     *  It is not merged with defaults.
      * @return void
      */
-    public function initialize(): void
+    public function initialize(array $config): void
     {
     }
 
--- a/src/Controller/Component.php
+++ b/src/Controller/Component.php
@@ -128,7 +128,8 @@ class Component implements EventListenerInterface
      * Implement this method to avoid having to overwrite
      * the constructor and call parent.
      *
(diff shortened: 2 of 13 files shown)


src/View/View.php
```diff
@@ -348,7 +348,7 @@ public function __construct(
         $this->request = $request;
         $this->response = $response ?: new Response();
         $this->Blocks = new $this->_viewBlockClass();
-        $this->initialize();
+        $this->initialize($viewOptions);
```
ADmad, line 351: I think this should be the diff of $viewOptions and keys set as view props similar to https://github.com/cakephp/cakephp/blob/d339441b1ce2a0164c41d3227f71f594558931c7/src/View/View.php#L337-L340.

othercorey: I think so too - if we want to make this change at all.

othercorey:
Originally, I added this to the wiki because I thought there was a mix of merged an unmerged $config passed to initialize(). Turns out, the $config is simply missing from some initialize().

ADmad:
Yeah I had a look into it too and for the ones that it's missing, it's really not needed.

othercorey:
I was originally going to remove the $config options and make users rely on the merged config, but it was suggested that to big of a break.

src/View/View.php
saeideng: config is available via `$this->getConfig()` so this Argument is useless in `View`

othercorey: The reason I made this is `$config` != $this->getConfig(). It is the original array from the constructor, not merged with defaults.

saeideng: if someone would need to `_passedVars`'s values , he can simply use like `$this->getLayout()` and other configs are available from `$this->getConfig()`
anyway I can not see good/reasonable usage for it

also there is not defaults in `View` , is there ?
infact , ```        'viewVars', 'autoLayout', 'helpers', 'template', 'layout', 'name', 'theme',
        'layoutPath', 'templatePath', 'plugin',```
are not at the output of `$this->getConfig().`

othercorey: The only reason is for consistency. I'm not a fan of $config, but we have it in other places.

othercorey:
Closing due to no interest.

PR #15784, opened 2021-08-28
Deprecate methods of Router for connecting routes.

ADmad:
The RouteBuilder's instance methods should be used instead.

### DIFF
--- a/src/Routing/Router.php
+++ b/src/Routing/Router.php
@@ -206,9 +206,14 @@ class Router
      * @throws \Cake\Core\Exception\CakeException
      * @see \Cake\Routing\RouteBuilder::connect()
      * @see \Cake\Routing\Router::scope()
+     * @deprecated 4.3.0 Use the non-static method `RouterBuilder::connect()` instead.
      */
     public static function connect($route, $defaults = [], $options = []): void
     {
+        deprecationWarning(
+            '`Router::connect()` is deprecated, use the non-static method `RouterBuilder::connect()` instead.'
+        );
+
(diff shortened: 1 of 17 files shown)


ADmad:
I tried deprecating `Router::extensions()` too but `RequestHandlerComponent`'s wacky extension guessing only works with these globally set extensions and `RequestHandlerComponent::testInitializeNoContentTypeWithMultipleAcceptedTypes()` doesn't pass without it.

src/Routing/Router.php
```diff
@@ -206,9 +206,12 @@ public static function getNamedExpressions(): array
      * @throws \Cake\Core\Exception\CakeException
      * @see \Cake\Routing\RouteBuilder::connect()
      * @see \Cake\Routing\Router::scope()
+     * @deprecated 4.3.0 Use the `RouterBuilder`'s `connect()` method instead.
```
othercorey, line 209:
```suggestion
     * @deprecated 4.3.0 Use `RouterBuilder::connect()` instead.
```
Hopefully simpler english.

ADmad, line 209: I consciously avoided using `RouterBuilder::connect()` in the messages so as to prevent the not so bright users trying to use it statically and then complain.

othercorey: I thought so. It's up you, it's just a little awkwardly phrased.

ADmad: Yeah I not thrilled with the wording either.

markstory (commented):
Looks good to me. The builder was added in 3.0.0, so I don't feel too bad about deprecating the static methods 6 years later.

PR #15786, opened 2021-08-28
cake 5: Add native property type hints for Datasource

othercorey:
This will, unfortunately, affect all Entity classes. But, if we're going to do this, then might as well do it now.

### DIFF
--- a/src/Datasource/EntityTrait.php
+++ b/src/Datasource/EntityTrait.php
@@ -34,14 +34,14 @@ trait EntityTrait
      *
      * @var array
      */
-    protected $_fields = [];
+    protected array $_fields = [];
 
     /**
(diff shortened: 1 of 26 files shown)


src/Datasource/EntityTrait.php
```diff
@@ -34,22 +34,22 @@ trait EntityTrait
...
+    protected array $_original = [];
 
     /**
      * List of field names that should **not** be included in JSON or Array
      * representations of this Entity.
      *
      * @var array<string>
      */
-    protected $_hidden = [];
+    protected array $_hidden = [];
```
markstory, line 52: These are going to be painful for users and we'll need to have rector support for this. I've added these changes to the wiki list.

PR #15787, opened 2021-08-28
Remove use of deprecated autoFixtures setting

othercorey:
This back-ports a fix from 5.0 for loading fixtures with foreign key dependencies. Only postgres needs to wrap disableConstraints() in a transaction.

### DIFF
--- a/src/TestSuite/Fixture/FixtureDataManager.php
+++ b/src/TestSuite/Fixture/FixtureDataManager.php
@@ -18,6 +18,7 @@ namespace Cake\TestSuite\Fixture;
 
 use Cake\Core\Configure;
 use Cake\Core\Exception\CakeException;
+use Cake\Database\Driver\Postgres;
 use Cake\Datasource\ConnectionInterface;
 use Cake\Datasource\ConnectionManager;
 use Cake\TestSuite\TestCase;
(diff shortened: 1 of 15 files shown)


src/TestSuite/Fixture/FixtureDataManager.php
```diff
@@ -205,11 +206,18 @@ protected function runOperation(array $fixtures, callable $operation): void
         $dbs = $this->fixtureConnections($fixtures);
         foreach ($dbs as $connection => $fixtures) {
             $db = ConnectionManager::get($connection);
-            $db->transactional(function (ConnectionInterface $db) use ($fixtures, $operation): void {
+            if ($db->getDriver() instanceof Postgres) {
+                // disabling foreign key constraints is only valid in a transaction
+                $db->transactional(function (ConnectionInterface $db) use ($fixtures, $operation): void {
```
markstory, line 211: Does the transaction make other databases behave incorrectly?

markstory (commented):
Looks good. Should we also add a deprecation warning for `autoFixtures` being set to false? I can do that in a follow-up pull request if you'd like.

src/TestSuite/Fixture/FixtureDataManager.php
othercorey: It breaks sqlite because disabling foreign keys within a transaction is ignored. The transaction isn't really necessary. A failure breaks the test either way.

othercorey:
> Looks good. Should we also add a deprecation warning for `autoFixtures` being set to false? I can do that in a follow-up pull request if you'd like.

I think we can add it to `TestCase::loadFixtures()`.

PR #15789, opened 2021-08-29
Move PHPUnit bootstrap condition to schema loader

markstory:
Having this check in the schema loader lets the user-facing API be simpler to operate and have one fewer gotcha.

### DIFF
--- a/src/TestSuite/Fixture/SchemaGenerator.php
+++ b/src/TestSuite/Fixture/SchemaGenerator.php
@@ -23,6 +23,10 @@ use RuntimeException;
 /**
  * Create database schema from the provided metadata file.
  *
+ * This class is only intended for use by CakePHP's testsuite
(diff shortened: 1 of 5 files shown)


othercorey:
So `__PHPUNIT_BOOTSTRAP` is set after the first run only?

markstory:
> So __PHPUNIT_BOOTSTRAP is set after the first run only?

It gets set after phpunit finishes its first bootstrap run. It is used [when generating code](https://github.com/sebastianbergmann/phpunit/blob/24d9338124561fb4c7d79fb3e8b0d649d5bcad99/src/Util/PHP/Template/TestCaseMethod.tpl) for running a single test in isolation. Using this key might be a terrible idea, but I've not found a better way to do it other than to create our own global to track whether or not schema has already been created.

PR #15791, opened 2021-08-29
Add ping health check to mysql startup

othercorey:
Try to wait for mysql to startup before running commands to see if it prevents connection drop when bootstrapping.


othercorey:
Nevermind, this requires another check to run.

PR #15793, opened 2021-08-30
Switch ConnectionManager::alias() parameter names

othercorey:
I find this confusing every time. The alias should be the new name and the source should be the existing connection. This matches the name used by `dropAlias()`.

~Renamed $source to $target.~

### DIFF
--- a/src/Datasource/ConnectionManager.php
+++ b/src/Datasource/ConnectionManager.php
@@ -150,20 +150,20 @@ class ConnectionManager
      * ConnectionManager::alias('test_things', 'things');
      * ```
      *
-     * @param string $alias The alias to add. Fetching $source will return $alias when loaded with get.
-     * @param string $source The connection to add an alias to.
+     * @param string $source The existing connection to alias.
+     * @param string $alias The alias name that resolves to `$source`.
      * @return void
      * @throws \Cake\Datasource\Exception\MissingDatasourceConfigException When aliasing a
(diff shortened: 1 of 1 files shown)


ADmad:
Personally I have always thought of something already existing as the source, so naming the argument of the existing connection as `$target` feels alien to me.

othercorey:
$source to $target isn't important. I just don't want $alias to be the existing connection.

ADmad:
> I just don't want $alias to be the existing connection.

:+1:

src/Datasource/ConnectionManager.php
```diff
@@ -150,20 +150,20 @@ public static function parseDsn(string $config): array
...
-     * @param string $alias The alias to add. Fetching $source will return $alias when loaded with get.
-     * @param string $source The connection to add an alias to.
+     * @param string $source The existing connection to alias.
+     * @param string $alias The alias name that resolves to `$source`.
      * @return void
      * @throws \Cake\Datasource\Exception\MissingDatasourceConfigException When aliasing a
      * connection that does not exist.
      */
-    public static function alias(string $alias, string $source): void
+    public static function alias(string $source, string $alias): void
```
markstory, line 159: Doesn't this invert the behavior as well?

othercorey: No, I changed this line `static::$_aliasMap[$alias] = $source;`

PR #15796, opened 2021-08-30
Use User-Agent set in php.ini if available.

ADmad:


dereuromark:
This could go into 4.next, IMO.

ADmad:
> This could go into 4.next, IMO.

I don't mind that, if that's the consensus.

ADmad:
Closing in favour of#15799.

PR #15797, opened 2021-08-30
Change visibility of public properties of `Socket` class.

ADmad:
Added accessor methods `isConnected()` and `isEncrypted()`.


markstory (commented):
Looks good but tests are failing.

othercorey:
We should test a failure.

ADmad:
> Looks good but tests are failing.

Fixed.

> We should test a __get call at least.

Done.

PR #15802, opened 2021-08-31
Return a `Closure` from `routes.php` files.

ADmad:
This is cleaner than including the file within a method's context and making `$routes`
variable magically available.

### DIFF
--- a/tests/test_app/Plugin/TestPlugin/config/routes.php
+++ b/tests/test_app/Plugin/TestPlugin/config/routes.php
@@ -1,13 +1,14 @@
 <?php
 
 use Cake\Core\Configure;
+use Cake\Routing\RouteBuilder;
 
 Configure::write('PluginTest.test_plugin.routes', 'loaded plugin routes');
 
-if (isset($routes)) {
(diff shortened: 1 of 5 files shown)


tests/test_app/Plugin/TestPlugin/config/routes.php
```diff
@@ -1,13 +1,14 @@
 <?php
 
 use Cake\Core\Configure;
+use Cake\Routing\RouteBuilder;
 
 Configure::write('PluginTest.test_plugin.routes', 'loaded plugin routes');
 
-if (isset($routes)) {
+return function (RouteBuilder $routes) {
```
markstory, line 8: I guess this makes it clearer where the `$routes` variable is coming from.

ADmad: Yes, more easier to understand the context.

ADmad: Do we want to throw a deprecation error if a closure is not returned. It might be quite annoying for users hence I am wary of doing do.

PR #15804, opened 2021-08-31
Re-organize SchemaManager

othercorey:
This renames `SchemaManager` to `SchemaLoader` to reflect what it does and further reduce use of "manager".

Moved the static function to a normal method and moved the config to an instance config. Cleaned up parameters that were both nullable and had defaults. Removed used of final and private.

Added parameter to control truncating the tables post-load.


src/TestSuite/Fixture/SchemaLoader.php
```diff
@@ -29,77 +31,79 @@
...
 {
+    use InstanceConfigTrait;
+
     /**
      * @var \Cake\Console\ConsoleIo
      */
     protected $io;
 
+    /**
+     * @var array
```
dereuromark, line 44: we can now annotate the assoc arrays better with the next syntax IMO
```suggestion
     * @var array<string, mixed>
```

othercorey: you're right, we did switch to `array<>` for 4.3

src/TestSuite/Fixture/SchemaLoader.php
```diff
@@ -29,77 +31,79 @@
...
+     * @param bool $dropTables Drop all tables prior to loading schema files
+     * @param bool $truncateTables Truncate all tables after loading schema files
      * @return void
-     * @throws \Exception if the truncation failed
-     * @throws \RuntimeException if the file could not be processed
      */
-    public static function create(
+    public function loadFiles(
+        $files,
         string $connectionName,
```
othercorey, line 75: Not sure if this should take a connection name. This and SchemaLoader is littered with calls to get the connection.

src/TestSuite/Fixture/SchemaLoader.php
```diff
@@ -0,0 +1,105 @@
...
+    /**
+     * @var array
+     */
+    protected $_defaultConfig = [
+        'dropTables' => true,
+        'outputLevel' => ConsoleIo::NORMAL,
+    ];
+
+    /**
+     * SchemaManager constructor.
```
markstory, line 52:
```suggestion
     * Constructor.
```

src/TestSuite/Fixture/SchemaLoader.php
```diff
@@ -0,0 +1,105 @@
...
+    public function __construct(array $config = [])
+    {
+        $this->setConfig($config);
+
+        $this->io = new ConsoleIo();
+        $this->io->level($this->getConfig('outputLevel'));
+    }
+
+    /**
+     * Import the schema from a file, or an array of files.
```
markstory, line 65:
```suggestion
     * Import schema from a file, or an array of files.
```

src/TestSuite/Fixture/SchemaLoader.php
markstory: Connection names is less work for the developer and all of this code is coupled to `Connection` already.

tests/TestCase/TestSuite/Fixture/SchemaLoaderTest.php
```diff
@@ -0,0 +1,135 @@
...
+
+        $schemaCollection = $this->createMock(Collection::class);
+        $schemaCollection
+            ->expects($this->atLeastOnce())->method('listTables')->willReturn(['schema_test']);
+        $schemaCollection
+            ->expects($this->atLeastOnce())->method('describe')->willReturn($tableSchema);
+        $connection
+            ->expects($this->atLeastOnce())->method('getSchemaCollection')->willReturn($schemaCollection);
+
+        ConnectionManager::setConfig('schema_test', $connection);
```
markstory, line 116: Could we create a table with a random name, and drop/truncate that instead? This many mocks makes the test very brittle and tedious to maintain as the test is coupled to the implementation tightly.

PR #15805, opened 2021-08-31
Start documenting the assoc arrays as per their string key.

dereuromark:
Follow https://github.com/cakephp/cakephp/pull/15804#pullrequestreview-742943993

After all, assoc arrays are defined by their assoc key.

Quite the manual work
There are for example 360 options arrays left

I wonder if there is a bit more automated (tooling based way) for it.. :)

### DIFF
--- a/psalm.xml
+++ b/psalm.xml
@@ -20,6 +20,7 @@
 
     <issueHandlers>
         <RedundantConditionGivenDocblockType errorLevel="suppress"/>
+        <RedundantCastGivenDocblockType errorLevel="suppress"/>
         <DocblockTypeContradiction errorLevel="suppress"/>
         <MissingClosureParamType errorLevel="suppress"/>
         <MissingClosureReturnType errorLevel="suppress"/>
--- a/src/Auth/BaseAuthenticate.php
+++ b/src/Auth/BaseAuthenticate.php
@@ -256,7 +256,7 @@ abstract class BaseAuthenticate implements EventListenerInterface
      *   function should have signature like `logout(EventInterface $event, array $user)`
      *   where `$user` is the user about to be logged out.
      *
-     * @return array List of events this class listens to. Defaults to `[]`.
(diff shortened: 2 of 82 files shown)


dereuromark:
Looking at the fails, the docblocks where lying all along it seems
Even though the internal classes expect assoc arrays, the reality is sometimes different :/

        $types = [];
        if ($type !== null) {
            $types = [$type, 'integer', $type]; // not assoc at all
        }

        return (new AggregateExpression('LEAD', $params, $types, $type ?? 'float'))->over();

also

    return new FunctionExpression('DATE', $p['value'], [$p['type']]);

markstory:
> Looking at the fails, the docblocks where lying all along it seems
Even though the internal classes expect assoc arrays, the reality is sometimes different :/

On the upside with the improved tools we have today we can make the docs reflect reality more :smile:

dereuromark:
> On the upside with the improved tools we have today we can make the docs reflect reality more

@markstory Done :)

psalm.xml
```diff
@@ -20,6 +20,7 @@
 
     <issueHandlers>
         <RedundantConditionGivenDocblockType errorLevel="suppress"/>
+        <RedundantCastGivenDocblockType errorLevel="suppress"/>
```
ADmad, line 23: Is this error triggered at too many places? If it's only for a few then it's better to update the baseline file instead.

dereuromark: No, not too many.
I just dont think that it is worth being reported at all, those are usually just nice in 4.3 where we cannot "actually promise" the value, so the cast just makes sure the documented type coming in is the one we expect later on.
In 5.x this can be removed then as we will have types in place to assert it for us. Until then an extra cast might not even be extra, and doesn't hurt in the other case IMO.

ADmad: Agreed, keep for 4.3 and remove for 5.0.

PR #15806, opened 2021-08-31
Document the assoc options arrays as per their string key.

dereuromark:
Follow https://github.com/cakephp/cakephp/pull/15805
but focus only on options arrays as they are always assoc for us

Fun fact: It even found some invalid (dead) code already:

    $this->_table->get($entity->get($this->_getPrimaryKey()), [here]); /// TreeBehavior:965


dereuromark:
The remaining psalm error is a false positive.
A bug of it not understanding `@inheritdoc` it seems.

Should we deprecate the mixed options usage of
```php
        if (is_int($disabledIndex)) {
            unset($options[$disabledIndex]);
            $options['disabled'] = true;
        }
```
Afaik this is not documented either, quite the hidden feature it seems.
Or do we do this for other bool attributes as well like this?

markstory:
> Afaik this is not documented either, quite the hidden feature it seems.
Or do we do this for other bool attributes as well like this?

Some of these mixed use modes go all the way back to 1.x and we've been bringing them along all these years as the maintenance cost of them has been low. I'm up for deprecating and removing the ones that are not documented.

dereuromark:
Ok I can make a separate follow up PR, then we can merge this now

othercorey:
Looks ok to me, but I haven't checks all these types. My type investigations were mostly non-array.

Issue #15803, opened 2021-08-31
Broken Mail Headers with MailTransport and PHP 8


calcosta:
This is a (multiple allowed):

* [x] bug

* CakePHP Version: 4.2.8.
* Platform and Target: Apache 2.4, PHP 8.0.8

### What you did
Upgraded existing Cake 4.2.8 application from PHP 7.4 to PHP 8. Tested applications mail functionality.
Mail Transport is configured like this:
```
'EmailTransport' => [
        'default' => [
            'className' => 'Mail',
            'host' => 'localhost',
            'port' => 25,
            'timeout' => 30,
            'client' => null,
            'tls' => null,
            'url' => env('EMAIL_TRANSPORT_DEFAULT_URL', null),
        ],
[...]
```

### What happened
While everything works as expected on PHP 7.4, on PHP 8.0.8 the emails sent have broken headers and cannot be displayed correctly or are rejected by the server.
Sending with Smtp transport works fine in PHP 8.
Below is an example of a broken Mail (see the whitespaces as first char at some header lines):
```
To: xxx
Subject: xxx
Date: Tue, 31 Aug 2021 14:17:44 +0200
 Message-ID: <xxx>
 MIME-Version: 1.0
 Content-Type: multipart/alternative; boundary="6897a799401f2eec49d3e5d63236e43b"
 Content-Transfer-Encoding: 8bit
Message-Id: <xxx>
From: xxx
```

### What you expected to happen
The mail should contain correct headers.
This seems to be related to a problem that exists in several Mail Libraries e.g. here: https://github.com/PHPMailer/PHPMailer/pull/2188

in `vendor/cakephp/cakephp/src/Mailer/Transport/MailTransport.php` line 44 is:
```
$eol = $this->getConfig('eol', PHP_EOL);

$headers = $message->getHeadersString(
    [
        'from',
        'sender',
        'replyTo',
        'readReceipt',
        'returnPath',
        'cc',
        'bcc',
    ],
    $eol,
    function ($val) {
        return str_replace("\r\n", '', $val);
    }
);
```
On linux environments PHP_EOL should be \n which is then used to implode the headers. This may cause the problem.
If you add ` 'eol' => "\r\n"` to the `EmailTransport` config everything works as expected on PHP 8.
In `vendor/cakephp/cakephp/src/Mailer/Transport/SmtpTransport.php` line 408 is:
```
return implode("\r\n", $messages);
```
[...]

PR #15807, opened 2021-09-01
Use crlf in mail headers

markstory:
Separate mail headers with \r\n to be compliant with RFC specs. This also fixes delivery issues in PHP 8. The content is also wrapped with CRLF for consistency. Testing this with a variety mail transfer agents is not something I'm able to do at this time.

Refs #15803

### DIFF
--- a/src/Mailer/Transport/MailTransport.php
+++ b/src/Mailer/Transport/MailTransport.php
@@ -41,7 +41,7 @@ class MailTransport extends AbstractTransport
         $to = $message->getHeaders(['to'])['To'];
         $to = str_replace("\r\n", '', $to);
 
-        $eol = $this->getConfig('eol', PHP_EOL);
+        $eol = $this->getConfig('eol', "\r\n");
         $headers = $message->getHeadersString(
             [
                 'from',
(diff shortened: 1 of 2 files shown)


src/Mailer/Transport/MailTransport.php
```diff
@@ -41,7 +41,7 @@ public function send(Message $message): array
         $to = $message->getHeaders(['to'])['To'];
         $to = str_replace("\r\n", '', $to);
 
-        $eol = $this->getConfig('eol', PHP_EOL);
+        $eol = $this->getConfig('eol', "\r\n");
```
othercorey, line 44: The php bug listed above was fixed in php8. https://bugs.php.net/bug.php?id=47983

markstory: I'm hoping that MTAs are ok with CRLF in headers. The other option would be to change the default based on the PHP version.

othercorey: It's the standard and if php 8 is doing it then we're as compatible as php is.

othercorey: I wonder if php 7.4 will throw errors because PHP uses `\n` while cakephp uses `\r\n`. We might need version-specific default?

Would be nice to test that at least.

markstory: Yeah, using `\n` for PHP7 would make this a safer change.

PR #15808, opened 2021-09-01
4.3 - Improve deprecations

markstory:
* Improve the deprecation in QueryTrait::_call() so it will point to application code. Having the message point to a line in ORM\Query is not helpful.
* Don't emit duplicate deprecations. When running tests is it common for the same code to be run multiple times. Instead of flooding output with duplicate content we can use md5 sums to prevent noise. This will consume some memory, but deprecations should only really be on in dev/ci so we should be relatively safe.

### DIFF
--- a/src/Core/functions.php
+++ b/src/Core/functions.php
@@ -309,6 +309,13 @@ if (!function_exists('deprecationWarning')) {
             );
         }
 
+        static $errors = [];
+        $checksum = md5($message);
+        if (isset($errors[$checksum])) {
+            return;
+        }
+        $errors[$checksum] = true;
+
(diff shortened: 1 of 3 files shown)


src/Core/functions.php
```diff
@@ -309,6 +309,13 @@ function deprecationWarning(string $message, int $stackFrame = 1): void
             );
         }
 
+        static $errors = [];
+        $checksum = md5($message);
+        if (isset($errors[$checksum])) {
+            // return;
```
othercorey, line 315: The return is commented out

markstory: derp.

PR #15811, opened 2021-09-02
Deprecate non assoc options usage.

dereuromark:
Follow https://github.com/cakephp/cakephp/pull/15806

Will simplify the documentation for this method to just `array<string, mixed>` in 5.0

### DIFF
--- a/src/View/Helper/FormHelper.php
+++ b/src/View/Helper/FormHelper.php
@@ -2262,6 +2262,7 @@ class FormHelper extends Helper
 
         $disabledIndex = array_search('disabled', $options, true);
         if (is_int($disabledIndex)) {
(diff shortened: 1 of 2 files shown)


othercorey:
Seems like this would throw a deprecation warning too many times, but @markstory did add some cache logic.

dereuromark:
What do you mean?
Doesnt seem like a feature widely used, pretty much no one even knows it exists or is expected to work :)

othercorey:
Sorry, I was looking at the _initInputHelper() function only. You're right.

ADmad (approved):

othercorey:
@dereuromark Can you add this to the 4.3 migration notes?

PR #15813, opened 2021-09-02
Make FixtureHelper::loadFixtures() public

othercorey:
Users need some way to load fixtures from the name like `app.Articles`.


othercorey:
Maybe this should be `StrategyHelper`.

markstory:
> I think it's going to be imperfect either way.

Don't stress too much about it either way. This is going to be a very infrequently implemented feature so it won't be a frequent paper cut :smile:

PR #15815, opened 2021-09-02
Form\Form::validate() new 'validate' option to pass custom validator

Erwane:
Add new option to Form\Form::validate() to pass custom validator name or object, like in Marshaller.

### DIFF
--- a/src/Form/Form.php
+++ b/src/Form/Form.php
@@ -140,9 +140,9 @@ class Form implements EventListenerInterface, EventDispatcherInterface, Validato
     /**
      * Set the schema for this form.
      *
+     * @since 4.1.0
      * @param \Cake\Form\Schema $schema The schema to set
      * @return $this
-     * @since 4.1.0
      */
     public function setSchema(Schema $schema)
     {
@@ -158,8 +158,8 @@ class Form implements EventListenerInterface, EventDispatcherInterface, Validato
      * is first built. This hook method lets you configure the
      * schema or load a pre-defined one.
      *
-     * @return \Cake\Form\Schema the schema instance.
      * @since 4.1.0
+     * @return \Cake\Form\Schema the schema instance.
      */
     public function getSchema(): Schema
     {
@@ -177,9 +177,9 @@ class Form implements EventListenerInterface, EventDispatcherInterface, Validato
      * is first built. This hook method lets you configure the
      * schema or load a pre-defined one.
      *
+     * @deprecated 4.1.0 Use {@link setSchema()}/{@link getSchema()} instead.
      * @param \Cake\Form\Schema|null $schema The schema to set, or null.
      * @return \Cake\Form\Schema the schema instance.
-     * @deprecated 4.1.0 Use {@link setSchema()}/{@link getSchema()} instead.
      */
     public function schema(?Schema $schema = null): Schema
     {
--- a/src/Form/Form.php
+++ b/src/Form/Form.php
@@ -210,33 +210,22 @@ class Form implements EventListenerInterface, EventDispatcherInterface, Validato
      * Used to check if $data passes this form's validation.
      *
      * @param array $data The data to check.
-     * @param array $options The validate options
+     * @param string|bool $validator Validator name or `true` for default validator.
      * @return bool Whether or not the data is valid.
      * @throws \RuntimeException If validator is invalid.
      */
-    public function validate(array $data, array $options = []): bool
+    public function validate(array $data, $validator = true): bool
     {
-        $options += [
-            'validate' => true,
-        ];
-
-        $validator = null;
-        if ($options['validate'] === true) {
-            $validator = $this->getValidator();
-        } elseif (is_string($options['validate'])) {
-            $validator = $this->getValidator($options['validate']);
-        } elseif (is_object($options['validate'])) {
-            /** @var \Cake\Validation\Validator $validator */
-            $validator = $options['validate'];
+        if (!$validator) {
+            return true;
         }
 
-        if ($validator === null) {
-            throw new RuntimeException(
-                sprintf('validate must be a boolean, a string or an object. Got %s.', getTypeName($options['validate']))
-            );
+        if ($validator === true) {
+            $validator = null;
         }
 
-        $this->_errors = $validator->validate($data);
+        $this->_errors = $this->getValidator($validator)
+            ->validate($data);
 
         return count($this->_errors) === 0;
     }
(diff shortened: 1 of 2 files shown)


ADmad:
What's your use case which requires switching the validator for a form?

Erwane:
I have 4 pages `index`. Organisers, Bands, Workshops and Events. All have a form to filter this index. 3 have same form and validation, only the Events form index is different and need an extended validation. I can duplicate or extends the `IndexForm` or alter the `Validator` in my controller, but adding this `validate` options allow differents validation and harmonize code and options with the `Marshaller`.

ADmad:
Personally I would just extend the `IndexForm` and modify the validator but if others think this is a useful feature I don't mind it.

That said since `Form::execute()` calls `Form::validate()` in order for this change to be really useful an `$options` argument would need to be added to `Form::execute()` too.

src/Form/Form.php
```diff
@@ -209,11 +210,32 @@ protected function _buildSchema(Schema $schema): Schema
...
+        $options += [
+            'validate' => true,
+        ];
+
+        $validator = null;
+        if ($options['validate'] === true) {
+            $validator = $this->getValidator();
+        } elseif (is_string($options['validate'])) {
+            $validator = $this->getValidator($options['validate']);
+        } elseif (is_object($options['validate'])) {
```
ADmad, line 228: I know that you have copied this code block verbatim from `Marshaller::_validate()` but I don't see the need to support passing a validator object. (Even for the marshaller/table it's not well documented and should be just deprecated/removed but that's a separate issue.)

With the validator name only thing configurable the instead of `$options` name we could just have `?string $validator` as the argument. Then the only change required will be `$validator = $this->getValidator($validator);` instead of `$validator = $this->getValidator();`

Erwane: I'm okay with that. Waiting others comments and i will clean code, or close PR ;)

markstory: I think having the `validate` option is ok. It is not something that will get used often but it also doesn't add 'new' syntax as we have a similar option and usage in ORM tables.

I agree with @ADmad that we don't need to support passing a validator object in, and that we should probably deprecate that behavior from Table as well.

Erwane:
Code was modified to reflect ORM validation in 5.x

ADmad:
As I mentioned earlier `Form::execute()` also needs to be updated to add the `$options` argument, so that `$options['validate']` can be used as 2nd argument when calling `Form::validate()`.

src/Form/Form.php
```diff
@@ -209,12 +209,22 @@ protected function _buildSchema(Schema $schema): Schema
...
      * @return bool Whether or not the data is valid.
+     * @throws \RuntimeException If validator is invalid.
      */
-    public function validate(array $data): bool
+    public function validate(array $data, $validator = true): bool
     {
-        $validator = $this->getValidator();
-        $this->_errors = $validator->validate($data);
+        if (!$validator) {
+            return true;
```
markstory, line 219: I feel like this should return false so that calls with `$validator == false` which also includes `null` (clearly a developer mistake) fail validation instead of passing. Having the method succeed when there is obvious user error seems bad to me.

Another option could be to throw an exception when `$validator === null`. This would enforce the doc block type at runtime helping catch mistakes earlier.

Erwane: $validator = false (or null) mean "no validation", the method should pass.

Erwane:
> As I mentioned earlier `Form::execute()` also needs to be updated to add the `$options` argument, so that `$options['validate']` can be used as 2nd argument when calling `Form::validate()`.

Wait, in the `execute()` the second argument is an array `$options` and in `validate` it is a `string|bool` ?

`execute(array $data, array $options = [])` and `validate(array $data, $validate = true)`

ADmad:
Please check "Allow editing by maintainers" to the PR so that I can push a commit. Things will proceed faster that way :)

ADmad:
```
diff --git a/src/Form/Form.php b/src/Form/Form.php
index a383221827..53374f8008 100644
--- a/src/Form/Form.php
+++ b/src/Form/Form.php
@@ -209,21 +209,13 @@ class Form implements EventListenerInterface, EventDispatcherInterface, Validato
      * Used to check if $data passes this form's validation.
      *
      * @param array $data The data to check.
-     * @param string|bool $validator Validator name or `true` for default validator.
+     * @param string|null $validator Validator name.
      * @return bool Whether or not the data is valid.
      * @throws \RuntimeException If validator is invalid.
      */
-    public function validate(array $data, $validator = true): bool
+    public function validate(array $data, ?string $validator = null): bool
     {
-        if (!$validator) {
-            return true;
-        }
-
-        if ($validator === true) {
-            $validator = null;
-        }
-
-        $this->_errors = $this->getValidator($validator)
+        $this->_errors = $this->getValidator($validator ?: static::DEFAULT_VALIDATOR)
             ->validate($data);
 
         return count($this->_errors) === 0;
@@ -270,20 +262,30 @@ class Form implements EventListenerInterface, EventDispatcherInterface, Validato
      * This hook method can be implemented in subclasses to perform
      * the action of the form. This may be sending email, interacting
      * with a remote API, or anything else you may need.
+     * 
+     * ### Options:
      *
+     * - validate: Set to `false` to disable validation. Can also be a string of the validator ruleset to be applied.
+     *   Defaults to `true`/`'default'`.
+     * 
      * @param array $data Form data.
+     * @param array $options List of options.
      * @return bool False on validation failure, otherwise returns the
      *   result of the `_execute()` method.
      */
-    public function execute(array $data): bool
+    public function execute(array $data, array $options = []): bool
     {
         $this->_data = $data;
+        
+        $options += ['validate' => true];
 
-        if (!$this->validate($data)) {
-            return false;
+        if ($options['validate'] === false) {
+            return $this->_execute($data);
         }
 
-        return $this->_execute($data);
+        $validator = $options['validate'] === true ? static::DEFAULT_VALIDATOR : $options['validate'];
+        
+        return $this->validate($data, $validator) ? $this->_execute($data) : false;
     }
 
     /**
```

Erwane:
> Please check "Allow editing by maintainers" for the PR so that I can push a commit. Things will proceed faster that way :)

It's already checked :(

Erwane:
We can't skip the validation with this patch. Maybe you don't want to allow validation skip ?

ADmad:
> It's already checked :(

Oh sorry, I was trying to update the PR using codespaces and it kept bitching about perms regardless of which repo I tried to push :(

markstory:
Looking goo

othercorey:
Ready for 4.3?

ADmad (approved):

PR #15820, opened 2021-09-03
Document the assoc arrays as per their string key.

dereuromark:
Follow https://github.com/cakephp/cakephp/pull/15811

With this almost all should be done now.

### DIFF
--- a/src/Console/HelpFormatter.php
+++ b/src/Console/HelpFormatter.php
@@ -118,7 +118,7 @@ class HelpFormatter
         }
 
         $options = $parser->options();
-        if (!empty($options)) {
+        if ($options) {
             $max = $this->_getMaxLength($options) + 8;
             $out[] = '<info>Options:</info>';
             $out[] = '';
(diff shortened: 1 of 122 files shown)


src/Console/HelpFormatter.php
```diff
@@ -192,7 +192,7 @@ protected function _generateUsage(): string
     /**
      * Iterate over a collection and find the longest named thing.
      *
-     * @param array $collection The collection to find a max length of.
+     * @param array<\Cake\Console\ConsoleInputOption|\Cake\Console\ConsoleInputArgument|\Cake\Console\ConsoleInputSubcommand> $collection The collection to find a max length of.
```
othercorey, line 195: Can you check this in 5.0 when merging?

othercorey:
This one look ok too.

dereuromark:
How can this be solved?
```
Error: Class Cake\Console\Shell is marked as deprecated (see https://psalm.dev/098)
Error: Class Cake\Console\Shell is marked as deprecated (see https://psalm.dev/098)
Error: Process completed with exit code 2.
```
The --set-baseline didn't work

othercorey:
You have to add an entry to the deprecation section (don't forget to increase occurrences). 
```
<DeprecatedClass occurrences="4">
</DeprecatedClass>
```

The `<code>` should be the full type in the annotation.

ADmad:
> The --set-baseline didn't work

You need to specify the filename, so `--set-baseline=psalm-baseline.xml`.

dereuromark:
Yeah, I figured it out. Psalm could be smart enough to get them from the config, but meeh^^

PR #15823, opened 2021-09-03
cake 5: Add native type hints for Database properties

othercorey:
This skips adding type hints for a few properties such as StatementDecorator.

### DIFF
--- a/src/Database/Expression/AggregateExpression.php
+++ b/src/Database/Expression/AggregateExpression.php
@@ -29,14 +29,14 @@ use Closure;
 class AggregateExpression extends FunctionExpression implements WindowInterface
 {
     /**
-     * @var \Cake\Database\Expression\QueryExpression
+     * @var \Cake\Database\Expression\QueryExpression|null
      */
-    protected $filter;
+    protected ?QueryExpression $filter = null;
 
     /**
-     * @var \Cake\Database\Expression\WindowExpression
+     * @var \Cake\Database\Expression\WindowExpression|null
      */
-    protected $window;
+    protected ?WindowExpression $window = null;
 
     /**
      * Adds conditions to the FILTER clause. The conditions are the same format as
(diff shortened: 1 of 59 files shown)


src/Database/Expression/TupleComparison.php
```diff
@@ -32,7 +32,7 @@ class TupleComparison extends ComparisonExpression
      * @var array<string|null>
      * @psalm-suppress NonInvariantDocblockPropertyType
      */
-    protected $_type;
+    protected array $types;
```
othercorey, line 35: ComparisonExpression::_type is `?string` and cannot be re-declared as array here.

src/Database/Schema/MysqlSchemaDialect.php
```diff
@@ -25,13 +25,6 @@
  */
 class MysqlSchemaDialect extends SchemaDialect
 {
-    /**
-     * The driver instance being used.
-     *
-     * @var \Cake\Database\Driver\Mysql
-     */
-    protected $_driver;
```
othercorey, line 33: We should add any config methods to `DriverInterface`.

src/Database/Expression/AggregateExpression.php
```diff
@@ -87,7 +87,9 @@ public function over(?string $name = null)
     public function partition($partitions)
     {
         $this->over();
-        $this->window->partition($partitions);
+        /** @var \Cake\Database\Expression\WindowExpression $window */
+        $window = $this->window;
+        $window->partition($partitions);
```
othercorey, line 92: This is an annoying workaround for psalm since it can't handle a function initializing a property.

othercorey:
This is ready now.

src/Database/IdentifierQuoter.php
```diff
@@ -32,7 +32,7 @@ class IdentifierQuoter
      *
      * @var \Cake\Database\Driver
      */
-    protected $_driver;
+    protected Driver $_driver;
```
ADmad, line 35: Should this `DriverInterface`?

othercorey: Looks like it can be. We really should rename Driver.

PR #15824, opened 2021-09-03
Add DriverInterface::supports()

othercorey:
We often have the problem of adding new features that are conditionally supported. This requires a new function added to the common interface which is always a BC or awkward for type annotations.

I would like to add `DriverInterface::supports()` which accepts a string with the feature name. This allows supporting arbitrary features in a few drivers without affecting the API for all drivers.

The default implementation in Driver would return false for unknown features which is how you'd expect untested/unsupported features to work.

The core features are defined in `DriverInterface` as constants which does not cause a BC and can be updated without further BC.

Users can define their own feature names without interfering with any core code.

### DIFF
--- a/src/Database/Driver.php
+++ b/src/Database/Driver.php
@@ -75,13 +75,6 @@ abstract class Driver implements DriverInterface
     protected $_autoQuoting = false;
 
     /**
-     * Whether or not the server supports common table expressions.
-     *
-     * @var bool|null
-     */
-    protected $supportsCTEs = null;
-
-    /**
      * The server version
      *
      * @var string|null
@@ -273,17 +266,22 @@ abstract class Driver implements DriverInterface
      */
     public function supportsSavePoints(): bool
     {
-        return true;
+        deprecationWarning('Feature support checks are now implemented by `supports()` with FEATURE_* constants.');
+
+        return $this->supports(static::FEATURE_SAVEPOINT);
     }
 
     /**
      * Returns true if the server supports common table expressions.
      *
      * @return bool
+     * @deprecated 4.3.0 Use `supports(DriverInterface::FEATURE_QUOTE)` instead
      */
     public function supportsCTEs(): bool
     {
-        return $this->supportsCTEs === true;
+        deprecationWarning('Feature support checks are now implemented by `supports()` with FEATURE_* constants.');
+
+        return $this->supports(static::FEATURE_CTE);
     }
 
     /**
@@ -300,12 +298,13 @@ abstract class Driver implements DriverInterface
      * Checks if the driver supports quoting, as PDO_ODBC does not support it.
      *
      * @return bool
+     * @deprecated 4.3.0 Use `supports(DriverInterface::FEATURE_QUOTE)` instead
      */
     public function supportsQuoting(): bool
     {
-        $this->connect();
+        deprecationWarning('Feature support checks are now implemented by `supports()` with FEATURE_* constants.');
 
-        return $this->_connection->getAttribute(PDO::ATTR_DRIVER_NAME) !== 'odbc';
+        return $this->supports(static::FEATURE_QUOTE);
     }
 
     /**
@@ -425,6 +424,25 @@ abstract class Driver implements DriverInterface
     public function isAutoQuotingEnabled(): bool
     {
         return $this->_autoQuoting;
+    }
+
+    /**
+     * Returns whether the driver supports the feature.
+     *
+     * Defaults to true for "quoting" and "savepoints".
+     *
+     * @param string $feature Driver feature name
+     * @return bool
+     */
+    public function supports(string $feature): bool
+    {
+        switch ($feature) {
+            case static::FEATURE_QUOTE:
+            case static::FEATURE_SAVEPOINT:
+                return true;
+        }
+
+        return false;
     }
 
     /**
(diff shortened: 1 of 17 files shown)


src/Database/DriverInterface.php
```diff
@@ -25,9 +25,35 @@
...
  * @method int|null getMaxAliasLength() Returns the maximum alias length allowed.
  * @method int getConnectRetries() Returns the number of connection retry attempts made.
+ * @method bool supports(string $feature) Checks whether a feature is supported by the driver.
  */
 interface DriverInterface
 {
+    /**
+     * @var string
+     */
+    public const FEATURE_QUOTING = 'quoting';
```
markstory, line 35: Wouldn't all drivers support quoting? Or is this more of a 'is quoting enabled?' thing?

markstory (commented):
I like the generalized interface with constants so we retain type safety.

src/Database/DriverInterface.php
othercorey: That's from this existing function:
https://github.com/cakephp/cakephp/blob/4.next/src/Database/Driver.php#L304

markstory: I totally forgot about that.

othercorey:
Added implementation for all drivers. There is a supports function for dynamic constraints. Not sure if we need this since it's only used in the Fixture code.

src/Database/Driver.php
```diff
@@ -303,9 +296,7 @@ public function quote($value, $type = PDO::PARAM_STR): string
      */
     public function supportsQuoting(): bool
     {
-        $this->connect();
-
-        return $this->_connection->getAttribute(PDO::ATTR_DRIVER_NAME) !== 'odbc';
```
markstory, line 308: Should this have a deprecation warning?

src/Database/Driver.php
```diff
@@ -283,7 +276,7 @@ public function supportsSavePoints(): bool
      */
     public function supportsCTEs(): bool
     {
-        return $this->supportsCTEs === true;
```
markstory, line 286: Should this have a deprecation warning?

src/Database/Driver.php
```diff
@@ -273,7 +266,7 @@ public function rollbackTransaction(): bool
      */
     public function supportsSavePoints(): bool
     {
-        return true;
```
markstory, line 276: Should this have a deprecation warning?

src/Database/Driver/Mysql.php
```diff
@@ -305,15 +310,7 @@ public function version(): string
      */
     public function supportsCTEs(): bool
     {
-        if ($this->supportsCTEs === null) {
```
markstory, line 308: Should this have a deprecation warning?

src/Database/Driver/Mysql.php
```diff
@@ -323,15 +320,7 @@ public function supportsCTEs(): bool
      */
     public function supportsNativeJson(): bool
     {
-        if ($this->_supportsNativeJson === null) {
```
markstory, line 326: Should this have a deprecation warning?

src/Database/Driver/Mysql.php
```diff
@@ -341,14 +330,6 @@ public function supportsNativeJson(): bool
      */
     public function supportsWindowFunctions(): bool
     {
-        if ($this->_supportsWindowFunctions === null) {
```
markstory, line 344: Should this have a deprecation warning?

src/Database/Driver/Sqlite.php
```diff
@@ -311,11 +339,7 @@ protected function _transformFunctionExpression(FunctionExpression $expression):
      */
     public function supportsCTEs(): bool
     {
-        if ($this->supportsCTEs === null) {
```
markstory, line 314: Should this have a deprecation warning?

src/Database/Driver/Sqlite.php
```diff
@@ -325,10 +349,6 @@ public function supportsCTEs(): bool
      */
     public function supportsWindowFunctions(): bool
     {
-        if ($this->_supportsWindowFunctions === null) {
```
markstory, line 328: Should this have a deprecation warning?

src/Database/Driver.php
othercorey: Added for all support functions this replaces.

src/Database/DriverInterface.php
```diff
@@ -132,6 +168,14 @@ public function disableForeignKeySQL(): string;
      */
     public function enableForeignKeySQL(): string;
 
+    /**
+     * Returns whether the driver supports the feature.
+     *
+     * @param string $feature Driver feature name
+     * @return bool
+     */
+    public function supports(string $feature): bool;
```
ADmad, line 177: Adding a new method to the interface in a minor release would be a BC break. So for now we just add the `@method` annotation and add the method to the interface in 5.0.

othercorey: This was accidental. I added the annotation then copied this when writing the docblock.

othercorey:
I added test coverage for the mixed collection of feature checks in each of the drivers.

src/Database/Driver.php
```diff
@@ -427,6 +426,25 @@ public function isAutoQuotingEnabled(): bool
         return $this->_autoQuoting;
     }
 
+    /**
+     * Returns whether the driver supports the feature.
+     *
+     * Defaults to true for "quoting" and "savepoints".
```
ADmad, line 432:
```suggestion
     * Defaults to true for "quote" and "savepoints".
```

othercorey: I'll fix this.

PR #15830, opened 2021-09-03
cake 5: Add native type hints for Http properties

othercorey:


src/Http/Client/Response.php
```diff
@@ -89,35 +89,35 @@ class Response extends Message implements ResponseInterface
...
+    protected ?SimpleXMLElement $_xml = null;
 
     /**
      * Cached decoded JSON data.
      *
-     * @var array
+     * @var mixed
      */
-    protected $_json;
+    protected mixed $_json = null;
```
othercorey, line 120: Changed to match the function return types.

ADmad:
Just 1 more failing test which sets data as string which causes type error since `ServerRequest::$data` now has type `array|object|null`.

tests/TestCase/Http/ServerRequestTest.php
```diff
@@ -1228,6 +1228,7 @@ public function testGetDataOnStringData(): void
         $request = new ServerRequest(compact('post'));
         $this->assertNull($request->getData('Model'));
         $this->assertNull($request->getData('Model.field'));
+        $this->assertSame($post, $request->getParsedBody());
```
ADmad, line 1231: Return value of getParsedBody() can't be a string as per the interface https://github.com/php-fig/http-message/blob/efd67d1dc14a7ef4fc4e518e7dee91c271d524e4/src/ServerRequestInterface.php#L164

ADmad: I think we should just throw an exception if the constructor receives as string value as parsed body. We already have the `BodyParserMiddleware` which handles setting the parsed body as array/object from the string body contents https://github.com/cakephp/cakephp/blob/bff920a50e0dd0912e24915a778b0913f57225b2/src/Http/Middleware/BodyParserMiddleware.php#L168.

markstory: Throwing an error if the parsed results are strings makes sense.

ADmad: I have pushed a commit to that effect.

othercorey: This looks good. Can you change the test function description (and name?) just for consistency?

othercorey: n/m, I can do this.

PR #15832, opened 2021-09-04
cake 5: Add native type hints for Log properties

othercorey:


markstory:
There is a type error on this one.

src/Log/Log.php
```diff
@@ -172,37 +172,30 @@ class Log
...
     /**
-     * Initializes registry and configurations
+     * Creates registry if doesn't exist and creates all defined logging
+     * adapters if config isn't loaded.
      *
-     * @return void
+     * @return \Cake\Log\LogEngineRegistry
      */
-    protected static function _init(): void
+    protected static function getRegistry(): LogEngineRegistry
```
othercorey, line 180: psalm cannot follow init functions and gets very angry that everything is uninitialized. I fixed this by adding a getter in other packages so tried here too.

PR #15836, opened 2021-09-04
cake 5: Add native type hints for TestSuite properties

othercorey:
This one is big.


src/TestSuite/Constraint/Response/ContentType.php
```diff
@@ -25,7 +27,7 @@ class ContentType extends ResponseBase
     /**
      * @var \Cake\Http\Response
      */
-    protected $response;
+    protected ResponseInterface $response;
```
othercorey, line 30: Changed these responses to `ResponseInterface`.

PR #15841, opened 2021-09-04
Simplify Marshaller::_validate().

ADmad:

### DIFF
--- a/src/ORM/Marshaller.php
+++ b/src/ORM/Marshaller.php
@@ -25,7 +25,6 @@ use Cake\Datasource\InvalidPropertyInterface;
 use Cake\ORM\Association\BelongsToMany;
 use Cake\Utility\Hash;
 use InvalidArgumentException;
-use RuntimeException;
 
 /**
  * Contains logic to convert array data into entities.
@@ -188,7 +187,7 @@ class Marshaller
                 $entity->setAccess($key, $value);
             }
         }
-        $errors = $this->_validate($data, $options, true);
+        $errors = $this->_validate($data, $options['validate'], true);
 
         $options['isMerge'] = false;
         $propertyMap = $this->_buildPropertyMap($data, $options);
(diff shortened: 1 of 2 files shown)


src/ORM/Marshaller.php
```diff
@@ -240,32 +240,22 @@ public function one(array $data, array $options = []): EntityInterface
...
      *
      * @param array $data The data to validate.
-     * @param array<string, mixed> $options The options passed to this marshaller.
+     * @param bool|string $validator Validator name. `true` to use the default validator.
      * @param bool $isNew Whether it is a new entity or one to be updated.
      * @return array The list of validation errors.
      * @throws \RuntimeException If no validator can be created.
      */
-    protected function _validate(array $data, array $options, bool $isNew): array
+    protected function _validate(array $data, bool|string $validator, bool $isNew): array
```
ADmad, line 248: I would prefer the `Form::validate()` in #15815 also to be simple like this.

src/ORM/Marshaller.php
```diff
@@ -240,32 +240,22 @@ public function one(array $data, array $options = []): EntityInterface
...
 
-        $validator = null;
-        if ($options['validate'] === true) {
-            $validator = $this->_table->getValidator();
-        } elseif (is_string($options['validate'])) {
-            $validator = $this->_table->getValidator($options['validate']);
-        }
-
-        if ($validator === null) {
-            throw new RuntimeException(sprintf(
```
ADmad, line 262: Not passing an array allows using argument types and avoids such checks.

src/ORM/Marshaller.php
```diff
@@ -240,32 +239,22 @@ public function one(array $data, array $options = []): EntityInterface
      * Returns the validation errors for a data set based on the passed options
      *
      * @param array $data The data to validate.
-     * @param array<string, mixed> $options The options passed to this marshaller.
+     * @param bool|string $validator Validator name. `true` to use the default validator.
```
othercorey, line 242:
```suggestion
     * @param string|bool $validator Validator name or `true` for default validator.
```

src/ORM/Marshaller.php
othercorey: Is false supposed to disable it? If not, should we use `null` to indicate default instead?

ADmad: > Is false supposed to disable it?

Yes, one can use `'validate' => false` in `$options` for `Table::newEntity()/patchEntity()` to disable validation.

src/ORM/Marshaller.php
```diff
@@ -240,32 +239,22 @@ public function one(array $data, array $options = []): EntityInterface
...
      *
      * @param array $data The data to validate.
-     * @param array<string, mixed> $options The options passed to this marshaller.
+     * @param string|bool $validator Validator name or `true` for default validator.
      * @param bool $isNew Whether it is a new entity or one to be updated.
      * @return array The list of validation errors.
      * @throws \RuntimeException If no validator can be created.
      */
-    protected function _validate(array $data, array $options, bool $isNew): array
+    protected function _validate(array $data, string|bool $validator, bool $isNew): array
```
markstory, line 247: Nice that we don't have to do the runtime throw as we can use a native type.

PR #15844, opened 2021-09-04
Replace DatabaseSuite with directory-based testsuite

othercorey:
This avoids extending the internal TestSuite class and reduces local testing time.


tests/bootstrap.php
```diff
@@ -104,6 +104,10 @@
 
 ConnectionManager::setConfig('test', ['url' => getenv('DB_URL')]);
 
+if (env('CAKE_TEST_AUTOQUOTE')) {
+    ConnectionManager::get('test')->getDriver()->enableAutoQuoting(true);
+}
```
markstory, line 109: You could also modify the setConfig() call on 105 to have the `quoteIdentifiers` option set. But this works as well.

othercorey: I just copied this from the old code.

PR #15849, opened 2021-09-06
Rename TestConnectionManager to ConnectionHelper

othercorey:
This consolidates multiple copies of test connection aliasing and aligns the name of new fixture helpers.


src/TestSuite/ConnectionHelper.php
```diff
@@ -0,0 +1,71 @@
...
+        }
+    }
+
+    /**
+     * Enables query logging for all database connections.
+     *
+     * @param array<int, string> $connections Connection names or empty for all.
+     * @return void
+     */
+    public function enableQueryLogging(array $connections = []): void
```
othercorey, line 61: This parameter is for testing since I don't want to force enable query logging for all connections and then disable for all connections afterwards.

Not sure if there is a better way.

othercorey:
Should both this and FixtureHelper be static?

markstory (commented):
Looks good to me. We will need to update migrations once this merges as it also uses this class.

PR #15851, opened 2021-09-07
Move SchemaGenerator into SchemaLoader

othercorey:
SchemaGenerator was the equivalent of SchemaLoader (previously SchemaManager) but for internal schema files.

Having two very similar classes was confusing so I merged SchemaGenerator into `SchemaLoader::loadInternalFile()` to help separate the functionality for users. They can clearly see one is an internal loader and one is a public SQL file loader.

This also helps unify the implementation style and naming.


othercorey:
I can't find the extra whitespace phpcs fails on in the unrelated file.

src/TestSuite/Fixture/SchemaLoader.php
```diff
@@ -59,18 +65,20 @@ public function __construct(array $config = [])
...
+     * Load and apply schema sql file, or an array of files.
      *
      * @param array<string>|string $files Schema files to load
      * @param string $connectionName Connection name
      * @param bool $dropTables Drop all tables prior to loading schema files
      * @param bool $truncateTables Truncate all tables after loading schema files
      * @return void
      */
-    public function loadFiles(
+    public function loadSqlFiles(
```
markstory, line 81: We'll need to update the docs for this.

src/TestSuite/Fixture/SchemaLoader.php
```diff
@@ -83,16 +91,15 @@ public function loadFiles(
...
-            $cleaner->dropTables($connectionName);
+            $this->schemaCleaner->dropTables($connectionName);
         }
 
         /** @var \Cake\Database\Connection $connection */
         $connection = ConnectionManager::get($connectionName);
         foreach ($files as $file) {
             if (!file_exists($file)) {
-                throw new InvalidArgumentException("Unable to load schema file `$file`.");
+                throw new InvalidArgumentException("Unable to load sql file `$file`.");
```
markstory, line 102:
```suggestion
                throw new InvalidArgumentException("Unable to load SQL file `$file`.");
```

src/TestSuite/Fixture/SchemaLoader.php
othercorey: Updated the docs with new name.

PR #15855, opened 2021-09-08
cake 5: Add missing native type hints

othercorey:
These are all the native type hints that the phpcs sniff could generate that were skipped.

### DIFF
--- a/src/Database/Statement/PDOStatement.php
+++ b/src/Database/Statement/PDOStatement.php
@@ -33,7 +33,7 @@ class PDOStatement extends StatementDecorator
      *
      * @var \PDOStatement
      */
-    protected $_statement;
+    protected \PDOStatement $_statement;
 
     /**
      * Constructor
--- a/src/ORM/Association.php
+++ b/src/ORM/Association.php
@@ -117,7 +117,7 @@ abstract class Association
      *
      * @var array<string>|string|false
      */
-    protected $_foreignKey;
+    protected array|string|false $_foreignKey;
 
     /**
      * A list of conditions to be always included when fetching records from
@@ -148,14 +148,14 @@ abstract class Association
      *
      * @var \Cake\ORM\Table
      */
-    protected $_sourceTable;
+    protected Table $_sourceTable;
 
     /**
      * Target table instance
      *
      * @var \Cake\ORM\Table
      */
-    protected $_targetTable;
+    protected Table $_targetTable;
 
     /**
      * The type of join to be used when adding the association to a query
@@ -170,7 +170,7 @@ abstract class Association
      *
      * @var string
      */
-    protected $_propertyName;
+    protected string $_propertyName;
 
     /**
      * The strategy name to be used to fetch associated records. Some association
--- a/src/Utility/Xml.php
+++ b/src/Utility/Xml.php
@@ -303,7 +303,7 @@ class Xml
      * @return void
      * @throws \Cake\Utility\Exception\XmlException
      */
-    protected static function _fromArray(DOMDocument $dom, DOMDocument|DOMElement $node, $data, string $format): void
+    protected static function _fromArray(DOMDocument $dom, DOMDocument|DOMElement $node, array $data, string $format): void
     {
         if (empty($data) || !is_array($data)) {
             return;
--- a/src/View/Helper/NumberHelper.php
+++ b/src/View/Helper/NumberHelper.php
@@ -46,7 +46,7 @@ class NumberHelper extends Helper
      *
      * @var \Cake\I18n\Number
      */
-    protected $_engine;
+    protected Number $_engine;
 
     /**
      * Default Constructor
--- a/src/Collection/Iterator/ZipIterator.php
+++ b/src/Collection/Iterator/ZipIterator.php
@@ -91,7 +91,7 @@ class ZipIterator extends MultipleIterator implements CollectionInterface
      * @return array
      */
     #[ReturnTypeWillChange]
-    public function current()
+    public function current(): array
     {
         if ($this->_callback === null) {
             return parent::current();
--- a/src/Command/Helper/ProgressHelper.php
+++ b/src/Command/Helper/ProgressHelper.php
@@ -116,7 +116,7 @@ class ProgressHelper extends Helper
      * @param float|int $num The amount of progress to advance by.
      * @return $this
      */
-    public function increment($num = 1)
+    public function increment(float|int $num = 1)
     {
         $this->_progress = min(max(0, $this->_progress + $num), $this->_total);
 
--- a/src/Console/ConsoleInputOption.php
+++ b/src/Console/ConsoleInputOption.php
@@ -101,7 +101,7 @@ class ConsoleInputOption
         string $short = '',
         string $help = '',
         bool $isBoolean = false,
-        $default = null,
+        mixed $default = null,
         array $choices = [],
         bool $multiple = false,
         bool $required = false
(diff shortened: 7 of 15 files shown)


othercorey:
All of the Cache engine missing type hints are due to `Psr\SimpleCache\CacheInterface`.

src/Console/BaseCommand.php
```diff
@@ -225,7 +225,7 @@ protected function setOutputLevel(Arguments $args, ConsoleIo $io): void
      * @param \Cake\Console\ConsoleIo $io The console io
      * @return int|null|void The exit code or null for success
      */
-    abstract public function execute(Arguments $args, ConsoleIo $io);
+    abstract public function execute(Arguments $args, ConsoleIo $io): int|null|null;
```
othercorey, line 228: This was skipped because we can't declare a type with `|void` and most of these don't return anything.

src/Controller/Controller.php
```diff
@@ -823,7 +823,7 @@ public function isAction(string $action): bool
      * @return \Cake\Http\Response|null|void
      * @link https://book.cakephp.org/4/en/controllers.html#request-life-cycle-callbacks
      */
-    public function beforeFilter(EventInterface $event)
+    public function beforeFilter(EventInterface $event): Response|null|null
```
othercorey, line 826: This was skipped because we can declare a type with `|void`.

src/Database/Driver/Sqlite.php
```diff
@@ -104,7 +104,7 @@ class Sqlite extends Driver
      *
      * @var array
      */
-    protected $featureVersions = [
+    protected array $featureVersions = [
```
othercorey, line 107: 4.next merge.

src/Database/Query.php
```diff
@@ -1941,7 +1941,7 @@ public function func(): FunctionsBuilder
      * @psalm-suppress ImplementedReturnTypeMismatch
      */
     #[ReturnTypeWillChange]
-    public function getIterator()
+    public function getIterator(): StatementInterface
```
othercorey, line 1944: This is not `Traversable`. This needs to be resolved before return-types are non-tentative. 
https://www.php.net/manual/en/class.iteratoraggregate

src/Http/Response.php
```diff
@@ -612,7 +612,7 @@ public function getStatusCode(): int
      * @return static
      * @throws \InvalidArgumentException For invalid status code arguments.
      */
-    public function withStatus($code, $reasonPhrase = '')
+    public function withStatus(int $code, string $reasonPhrase = ''): static
```
othercorey, line 615: `Psr\Http\Message\ResponseInterface`

ADmad:
> All of the Cache engine missing type hints are due to Psr\SimpleCache\CacheInterface.

Hopefully this gets merged soonish https://github.com/php-fig/simple-cache/pull/24.

src/Database/Statement/PDOStatement.php
```diff
@@ -33,7 +33,7 @@ class PDOStatement extends StatementDecorator
      *
      * @var \PDOStatement
      */
-    protected $_statement;
+    protected \PDOStatement $_statement;
```
othercorey, line 36: This is a hot mess.

src/Datasource/ConnectionManager.php
```diff
@@ -194,7 +194,7 @@ public static function dropAlias(string $name): void
      * @throws \Cake\Datasource\Exception\MissingDatasourceConfigException When config
      * data is missing.
      */
-    public static function get(string $name, bool $useAliases = true)
+    public static function get(string $name, bool $useAliases = true): ConnectionInterface
```
othercorey, line 197: This is because apparently connections can be literally anything.

src/TestSuite/Constraint/Response/StatusCodeBase.php
```diff
@@ -36,7 +36,7 @@ abstract class StatusCodeBase extends ResponseBase
      * @return bool
      * @psalm-suppress MoreSpecificImplementedParamType
      */
-    public function matches($other): bool
+    public function matches(array|int $other): bool
```
othercorey, line 39: `matches()` comes from phpunit which is untyped.

src/ORM/Association.php
```diff
@@ -117,7 +117,7 @@ abstract class Association
      *
      * @var array<string>|string|false
      */
-    protected $_foreignKey;
+    protected array|string|false $_foreignKey;
```
othercorey, line 120: This defaults to null but code assumes it's non-null.

src/ORM/Association.php
```diff
@@ -148,14 +148,14 @@ abstract class Association
      *
      * @var \Cake\ORM\Table
      */
-    protected $_sourceTable;
+    protected Table $_sourceTable;
```
othercorey, line 151: This also defaults to null but code assumes it's non-null.

src/Utility/Xml.php
```diff
@@ -303,7 +303,7 @@ public static function fromArray(object|array $input, array $options = []): Simp
      * @return void
      * @throws \Cake\Utility\Exception\XmlException
      */
-    protected static function _fromArray(DOMDocument $dom, DOMDocument|DOMElement $node, $data, string $format): void
+    protected static function _fromArray(DOMDocument $dom, DOMDocument|DOMElement $node, array $data, string $format): void
```
othercorey, line 306: This is not always passed an array.

src/View/Helper/NumberHelper.php
```diff
@@ -46,7 +46,7 @@ class NumberHelper extends Helper
      *
      * @var \Cake\I18n\Number
      */
-    protected $_engine;
+    protected Number $_engine;
```
othercorey, line 49: These accept arbitrary objects as engines.

src/Log/Engine/ArrayLog.php
```diff
@@ -58,7 +58,7 @@ class ArrayLog extends BaseLog
      * @return void success of write.
      * @see Cake\Log\Log::$_levels
      */
-    public function log($level, Stringable|string $message, array $context = []): void
+    public function log(mixed $level, Stringable|string $message, array $context = []): void
```
othercorey, line 61: This is from a PSR interface.

src/Controller/Controller.php
ADmad: I don't think we should add return types to event callback methods. It would be quite annoying to unnecessary add `return null` in most cases.

src/Datasource/ConnectionManager.php
ADmad: Addressed in https://github.com/cakephp/cakephp/pull/15863.

src/Database/Query.php
ADmad: The fact that `StatementInterface` isn't traversable can be fixed by making it extend `Traversable`. But the larger issue is that `Database/Query::getIterator()` returns a `StatementInterface` instance while `ORM\Query::getIterator()` returns a `ResultSetInterface`. This is not reconcilable.

src/Cache/Engine/ArrayEngine.php
```diff
@@ -66,7 +67,7 @@ public function set($key, $value, $ttl = null): bool
      * @return mixed The cached data, or default value if the data doesn't exist, has
      * expired, or if there was an error fetching it.
      */
-    public function get($key, $default = null)
+    public function get(string $key, mixed $default = null): mixed
```
othercorey: I think we can add just the return type for these `get()` methods.

othercorey:
When I have some time, I'll remove the PSR/phpunit methods that we can't change to show what's really left.

PR #15863, opened 2021-09-09
Accept only ConnectionInterface instances.

ADmad:
ConnectionManager and ConnectionRegistry no longer accept arbitrary objects.

### DIFF
--- a/src/Datasource/ConnectionRegistry.php
+++ b/src/Datasource/ConnectionRegistry.php
@@ -19,6 +19,7 @@ namespace Cake\Datasource;
 use Cake\Core\App;
 use Cake\Core\ObjectRegistry;
 use Cake\Datasource\Exception\MissingDatasourceException;
+use Closure;
 
 /**
  * A registry object for connection instances.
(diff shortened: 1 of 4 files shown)


ADmad:
We could also extract most of the `ConnectionInterface` methods which `Connection` implements into a trait.

src/Datasource/ConnectionRegistry.php
```diff
@@ -68,24 +69,26 @@ protected function _throwMissingClassError(string $class, ?string $plugin): void
...
+            unset($config['className']);
 
-        if (is_object($class)) {
-            return $class;
+            /** @var \Cake\Datasource\ConnectionInterface */
+            return new $class($config);
         }
 
-        unset($config['className']);
+        if ($class instanceof Closure) {
```
othercorey, line 86: If we're going to support a hidden Closure object, I think it should be part of the parent `_create()` parameter type hint.

othercorey: Maybe we can implement closure support in the base class and not require every _create() to handle it.

ADmad: > If we're going to support a hidden Closure object, I think it should be part of the parent _create() parameter type hint.

Can't have `Closure|object` as type. It leads to `Fatal error: Type Closure|object|string contains both object and a class type, which is redundant`.

ADmad: I wonder if we can just drop support for `Closure`. `ConnectionRegistry` seems to be the only one supporting it.

I think the idea was to potentially delay creating an expensive connection instance only when needed. Not sure how useful that is in practice.

ADmad: Ugh.. `LogEngine::_create()` also accepts `callable`. In general I am trying to get rid of `callable`s where not necessary.

othercorey: Yea, makes sense that it throws an error. I wasn't 100% sure.

othercorey:
This looks ok to me. The closures are a separate topic if we already support them.

src/Datasource/ConnectionRegistry.php
markstory: > I wonder if we can just drop support for Closure. ConnectionRegistry seems to be the only one supporting it.

I don't think we should be dropping `Closure` support. It can be useful when defining connections/loggers with the DI container, at least until decide on whether or not all those things should be defined in the DI container entirely.

ADmad: > It can be useful when defining connections/loggers with the DI container, at least until decide on whether or not all those things should be defined in the DI container entirely.

Makes sense, I hadn't considered DIC usage. 

This should be good to merge then.

PR #15869, opened 2021-09-10
Remove unneded type checks.

ADmad:
They are covered return types specified for the methods.

### DIFF
--- a/src/Core/ServiceProvider.php
+++ b/src/Core/ServiceProvider.php
@@ -18,7 +18,6 @@ namespace Cake\Core;
 
 use League\Container\ServiceProvider\AbstractServiceProvider;
 use League\Container\ServiceProvider\BootableServiceProviderInterface;
-use RuntimeException;
 
 /**
  * Container ServiceProvider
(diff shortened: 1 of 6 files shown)


src/Core/ServiceProvider.php
```diff
@@ -44,25 +44,11 @@ abstract class ServiceProvider extends AbstractServiceProvider implements Bootab
...
-            $message = sprintf(
-                'Unexpected container type. Expected `%s` got `%s` instead.',
-                ContainerInterface::class,
-                get_debug_type($container)
-            );
-            throw new RuntimeException($message);
-        }
-
-        return $container;
+        return parent::getContainer();
```
othercorey, line 51: How is this set if it's supposed to be cake's ContainerInterface?

ADmad: Because Cake's `ContainerInterface` extends League's `ServiceProviderInterface`.

othercorey: I mean it returns the parent. I assume we set the parent instance somewhere?

ADmad: We don't need to explicitly set it, it's handled automatically:
https://github.com/thephpleague/container/blob/82a57588c630663d2600f046753b23ab6dcda9b5/src/Container.php#L52
https://github.com/thephpleague/container/blob/82a57588c630663d2600f046753b23ab6dcda9b5/src/ServiceProvider/ServiceProviderAggregate.php#L32

src/Core/ServiceProvider.php
```diff
@@ -44,25 +43,13 @@ abstract class ServiceProvider extends AbstractServiceProvider implements Bootab
     /**
      * Get the container.
      *
-     * This method's actual return type and documented return type differ
-     * because PHP 7.2 doesn't support return type narrowing.
-     *
      * @return \Cake\Core\ContainerInterface
+     * @psalm-suppress MoreSpecificReturnType
```
ADmad, line 47: Don't know why psalm whines about covariant return type when it's supported by the language.

othercorey: I think the issue is the method types mismatch. The parent return guarantees only the base class while this return expects the derived class.

The types are allowed but the parent return isn't guaranteed to match.

markstory (approved):
Looks good to me.

PR #15870, opened 2021-09-11
Add deprecation notice for old-fixtures

markstory:
I considered putting this deprecation into the TestFixture class but was concerned about flooding output for non-trivial sized applications. I thought that a single warning was a better option. Upgrading to the new fixtures can be done at the very end of the upgrade cycle or even after all the application code is working, so one message should be enough.


othercorey:
Looks ok. Shouldn't cause any test failures.

othercorey:
Added deprecated annotation to class.

PR #15873, opened 2021-09-11
Update SECURITY.md to current state

LordSimal:
Looking through the files I just noticed, that the SECURITY.md hasn't been updated in like 2 years.

I took the updated info from https://github.com/cakephp/cakephp/wiki but I wonder if we actually need this file at all?


markstory:
> I took the updated info from https://github.com/cakephp/cakephp/wiki but I wonder if we actually need this file at all?

It is useful to have so we have that information in a place that is used commonly by many other projects.

PR #15875, opened 2021-09-11
Fix docblock.

ADmad:

### DIFF
--- a/src/Cache/Cache.php
+++ b/src/Cache/Cache.php
@@ -579,9 +579,8 @@ class Cache
      *   the cache key is empty. Can be any callable type supported by your PHP.
      * @param string $config The cache configuration to use for this operation.
      *   Defaults to default.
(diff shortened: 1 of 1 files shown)


src/Cache/Cache.php
```diff
@@ -579,9 +579,8 @@ public static function enabled(): bool
      *   the cache key is empty. Can be any callable type supported by your PHP.
      * @param string $config The cache configuration to use for this operation.
      *   Defaults to default.
-     * @return mixed If the key is found: the cached data, false if the data
-     *   missing/expired, or an error. If the key is not found: boolean of the
-     *   success of the write
+     * @return mixed If the key is found: the cached data.
+     *   If the key is not found the value returned by the callable.
```
othercorey, line 583: Should we have named it $default instead?

ADmad: I don't think so.

Issue #15382, opened 2021-03-10
PaginatorHelper::limitControl() form escapes query parameters in URL


challgren:
This is a (multiple allowed):

* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 4.2.4
* Platform and Target: MacOS/Ubuntu

### What you did
In my Controller I'm using friendsofcake/Search and have a search applied. Request URL is `/batches?owner=Ch%2A&expected_sku=TST%2A` and the form action generated is `/batches?owner=Ch%2A&amp;expected_sku=TST%2A`

### What happened
The Form action for the `PaginatorHelper::limitControl()` should not escape the URL. But really this boils down to https://github.com/cakephp/cakephp/blob/master/src/View/Helper/FormHelper.php#L413 not having an ['escape' => false] since `UrlHelper` defaults to escaping a URL.

### What you expected to happen
The `PaginatorHelper::limitControl()` form should have an action of `/batches?owner=Ch%2A&expected_sku=TST%2A`

othercorey:
@markstory Does this need to be fixed in 4.3.0 or can it be a bug fix?

markstory:
It can go into either release I think. Once fixed, the output will the same or better.

markstory:
I wasn't able to reproduce this issue, I have a pull request open to add a test for this scenario. Did I miss something to cause it to happen?

challgren:
Well its been so long I'm unable to reproduce this now. But I am noticing that if I have query string parameters set and use the Paginator->limitControl() and choose a value the query string is remove and just the limit is applied.

markstory:
> But I am noticing that if I have query string parameters set and use the Paginator->limitControl() and choose a value the query string is remove and just the limit is applied.

That sounds like a different issue. We have a test covering that scenario (added in #15878) I wonder what could be different about your setup?

Going to close this as the escaping is intentional. If you find a way to reproduce the new problem you're seeing please open an issue :smile:

PR #15878, opened 2021-09-12
Add test for PaginatorHelper::limitControl()

markstory:
I was not able to reproduce the issue described in #15382. The current output does not have double encoded ampersand.

### DIFF
--- a/tests/TestCase/View/Helper/PaginatorHelperTest.php
+++ b/tests/TestCase/View/Helper/PaginatorHelperTest.php
@@ -3382,6 +3382,42 @@ class PaginatorHelperTest extends TestCase
     }
 
     /**
+     * test the limitControl() with a request url and query.
+     *
+     * @return void
+     */
+    public function testLimitControlUrlWithQuery()
+    {
+        $request = new ServerRequest([
+            'url' => '/batches?owner=billy&expected=1',
+            'params' => [
+                'plugin' => null, 'controller' => 'Batches', 'action' => 'index', 'pass' => [],
+            ],
+            'base' => '',
+            'webroot' => '/',
+        ]);
+        Router::setRequest($request);
+        $this->View->setRequest($request);
+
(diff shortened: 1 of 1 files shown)


tests/TestCase/View/Helper/PaginatorHelperTest.php
```diff
@@ -3381,6 +3381,42 @@ public function testLimitControl()
...
+            ],
+            'base' => '',
+            'webroot' => '/',
+        ]);
+        Router::setRequest($request);
+        $this->View->setRequest($request);
+
+        $out = $this->Paginator->limitControl([1 => 1]);
+        $expected = [
+            ['form' => ['method' => 'get', 'accept-charset' => 'utf-8', 'action' => '/batches?owner=billy&amp;expected=1']],
```
challgren, line 3404: We would expect `/batches?owner=billy&expected=1`

markstory: Why? You're in HTML. A bare `&` is not allowed. The [HTML4 spec](https://www.w3.org/TR/html401/charset.html#h-5.3.2) has a section on this. That section is also present in the [HTML 5 specs](https://html.spec.whatwg.org/multipage/syntax.html#attributes-2).

othercorey: I read the standard as saying you can't have a & with identifier-style letters and numbers and then `;`.

markstory: Yes, but there isn't a reasonable way to ensure there is no `;` somewhere in the trailing values. PHP's built-in HTML escaping behaves in a similar way

```php
php > echo htmlspecialchars('foo&bar', ENT_QUOTES);
foo&amp;bar
php > echo htmlspecialchars('foo&bar', ENT_HTML5);
foo&amp;bar
```
Using alternative HTML escaping presents a number of risks an unknown consequences to applications.

PR #15881, opened 2021-09-12
Remove use of ReturnTypeWillChange attribute.

ADmad:

### DIFF
--- a/src/Database/StatementInterface.php
+++ b/src/Database/StatementInterface.php
@@ -16,13 +16,15 @@ declare(strict_types=1);
  */
 namespace Cake\Database;
 
+use Traversable;
+
 /**
  * Represents a database statement. Concrete implementations
  * can either use PDOStatement or a native driver
  *
  * @property-read string $queryString
  */
-interface StatementInterface
+interface StatementInterface extends Traversable
 {
     /**
      * Used to designate that numeric indexes be returned in a result when calling fetch methods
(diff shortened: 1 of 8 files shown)


src/Collection/Iterator/ZipIterator.php
```diff
@@ -90,7 +89,7 @@ public function __construct(array $sets, ?callable $callable = null)
      *
      * @return array
      */
-    #[ReturnTypeWillChange]
+    #[\ReturnTypeWillChange]
```
ADmad: We could enable the `allowFullyQualifiedGlobalClasses` option for this sniff https://github.com/slevomat/coding-standard#slevomatcodingstandardnamespacesreferenceusednamesonly-.

othercorey: Would be nice if we could allow only some classes instead of all `\DateTime` names, but I don't think that's possible.

Wish attributes were recognized better. We could move the CS+SA step to php 8.

ADmad: Yeah, I just reverted to the `use` statement for now.

src/Database/StatementInterface.php
```diff
@@ -16,13 +16,15 @@
...
+use Traversable;
+
 /**
  * Represents a database statement. Concrete implementations
  * can either use PDOStatement or a native driver
  *
  * @property-read string $queryString
  */
-interface StatementInterface
+interface StatementInterface extends Traversable
```
othercorey, line 27: This will get further merged into StatementDecorator which implements IteratorAggregate so this should be fine.

PR #15882, opened 2021-09-12
Include sniff name in error message

ADmad:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -222,7 +222,7 @@ jobs:
       run: composer stan-setup
 
     - name: Run PHP CodeSniffer
-      run: vendor/bin/phpcs --report=checkstyle src/ tests/ | cs2pr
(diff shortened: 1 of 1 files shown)


othercorey:
I remember trying this and found it broke the GA reporting. Maybe I was testing incorrectly though.

ADmad:
> I remember trying this and found it broke the GA reporting. Maybe I was testing incorrectly though.

In that case I'll generate an error to test.

othercorey:
That might have been before cs2pr was added. I don't remember exactly what order I made the changes :(.

ADmad:
So using the `-n` flag doesn't break the GA reporting but doesn't show the sniff name either :(

@markstory Stickler needs looking into, it's passing even though there is an error.

markstory:
> Stickler needs looking into, it's passing even though there is an error.

I've not updated the phpcs version in a bit I'll try that.

othercorey:
We can merge this if you want. I recall this is the issue I had. The checkstyle output seems to omit that info.

PR #15884, opened 2021-09-13
5.0 Consolidate event names

dereuromark:
To normalize with Controller events

            // Controller::implementedEvents()
            'Controller.initialize' => 'beforeFilter',
            'Controller.beforeRender' => 'beforeRender',
            'Controller.beforeRedirect' => 'beforeRedirect',
            'Controller.shutdown' => 'afterFilter',

Components now

            // Component::implementedEvents()
            'Controller.initialize' => 'beforeFilter',
            'Controller.startup' => 'startup',
            'Controller.beforeRender' => 'beforeRender',
            'Controller.beforeRedirect' => 'beforeRedirect',
            'Controller.shutdown' => 'afterFilter',

The alternative would be to rename the controller one to "shutdown", but then the beforeFilter/afterFilter wouldn't match anymore, as well the consistency towards beforeRender/afterRender.
So even though "filter" seems like a weird word here, I guess this is the best way to consolidate.

### DIFF
--- a/src/Controller/Component.php
+++ b/src/Controller/Component.php
@@ -170,7 +170,7 @@ class Component implements EventListenerInterface
             'Controller.startup' => 'startup',
             'Controller.beforeRender' => 'beforeRender',
             'Controller.beforeRedirect' => 'beforeRedirect',
-            'Controller.shutdown' => 'shutdown',
+            'Controller.shutdown' => 'afterFilter',
         ];
         $events = [];
         foreach ($eventMap as $event => $method) {
--- a/src/Controller/Component.php
+++ b/src/Controller/Component.php
@@ -37,13 +37,13 @@ use Cake\Log\LogTrait;
  * cycle. The available callbacks are:
  *
  * - `beforeFilter(EventInterface $event)`
- *   Called before the controller's beforeFilter method by default.
+ *   Called before Controller::beforeFilter() method by default.
  * - `startup(EventInterface $event)`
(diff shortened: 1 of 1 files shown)


ADmad (approved):

src/Controller/Component.php
```diff
@@ -170,7 +170,7 @@ public function implementedEvents(): array
             'Controller.startup' => 'startup',
             'Controller.beforeRender' => 'beforeRender',
             'Controller.beforeRedirect' => 'beforeRedirect',
-            'Controller.shutdown' => 'shutdown',
+            'Controller.shutdown' => 'afterFilter',
```
othercorey, line 173: Deprecating this somehow would be nice. It's a fairly hidden change that users won't understand.

We'd have to implement `afterFilter()` and the default implementation would check for the existence of `shutdown()`.

dereuromark: Yeah, we can deprecate in 4.3 I guess. Or in 5.0 now as you mention.

ADmad: I don't think the callback would be used often. We can add a `Component::shutdown()` in 5.0 itself which throws a message telling using to use `Component::afterFilter()` instead.

dereuromark: True. But https://github.com/cakephp/cakephp/pull/15887 seems cleaner as we then can remove the baggage in 5.0 directly, not having to continue the fallback until 6.x

PR #15886, opened 2021-09-13
Use addHelpers() instead of bool flag param.

dereuromark:
This was also on the list of cleanups
I also added a normalization on the helpers as assoc arrays.
I always had issues with merging them, as they appear twice in the list, once as key and once as value if different levels add helpers with different config.
What do you think?
Could we in general normalize all such configs also for components, behaviors and alike?
This makes it also easier to find out if a helper got added by just checking the key instead of key + in_array etc games.

We will want to deprecate in 4.3 here to allow an upgrade path.

### DIFF
--- a/src/View/ViewBuilder.php
+++ b/src/View/ViewBuilder.php
@@ -310,13 +310,7 @@ class ViewBuilder implements JsonSerializable
      */
     public function addHelper(string $helper, array $options = [])
     {
-        if ($options) {
-            $array = [$helper => $options];
-        } else {
-            $array = [$helper];
-        }
-
-        $this->_helpers = array_merge($this->_helpers, $array);
+        $this->_helpers[$helper] = $options;
 
         return $this;
     }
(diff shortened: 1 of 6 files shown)


src/View/ViewBuilder.php
```diff
@@ -310,27 +310,39 @@ public function getPlugin(): ?string
...
      * Sets the helpers to use.
      *
      * @param array $helpers Helpers to use.
-     * @param bool $merge Whether or not to merge existing data with the new data.
      * @return $this
      */
-    public function setHelpers(array $helpers, bool $merge = true)
+    public function setHelpers(array $helpers)
     {
-        if ($merge) {
```
markstory, line 353: We should deprecate this flag in 4.3

dereuromark: https://github.com/cakephp/cakephp/pull/15898

src/View/ViewBuilder.php
```diff
@@ -310,27 +310,39 @@ public function getPlugin(): ?string
...
+
+        return $this;
+    }
+
+    /**
+     * Adds helpers to use by merging with existing ones.
+     *
+     * @param array $helpers Helpers to use.
+     * @return $this
+     * @since 5.0.0
```
dereuromark: This will be fixed up to 4.3 once 4.next is merged.

dereuromark:
I rebased on top of the backported 4.next PR merged into 5.x.

PR #15887, opened 2021-09-13
4.3: Consolidate event names. Backport

dereuromark:
Backport of https://github.com/cakephp/cakephp/pull/15884

### DIFF
--- a/src/Controller/Component.php
+++ b/src/Controller/Component.php
@@ -37,13 +37,13 @@ use Cake\Log\LogTrait;
  * cycle. The available callbacks are:
  *
  * - `beforeFilter(EventInterface $event)`
- *   Called before the controller's beforeFilter method by default.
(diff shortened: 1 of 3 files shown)


src/Controller/Component.php
```diff
@@ -170,12 +170,19 @@ public function implementedEvents(): array
...
         ];
         $events = [];
         foreach ($eventMap as $event => $method) {
             if (method_exists($this, $method)) {
                 $events[$event] = $method;
+
+                continue;
+            }
+
+            if ($event === 'Controller.shutdown' && method_exists($this, 'shutdown')) {
```
ADmad: The `$eventMap` is static so checking the event name inside the loop is unnecessary. Can instead do the following after the loop:
```
if (!isset($events['Controller.shutdown']) && method_exists($this, 'shutdown'))` {
   ...
}
```

PR #15888, opened 2021-09-13
Change argument/value type from callable to Closure.

ADmad:
It's unlikely these values need elaborate logic to necessitate using a callable
besides Closure. It also helps simplying type checks where the value can be a
string or callable since strings can be valid callbables too.

### DIFF
--- a/src/Http/ServerRequest.php
+++ b/src/Http/ServerRequest.php
@@ -22,6 +22,7 @@ use Cake\Core\Exception\CakeException;
 use Cake\Http\Cookie\CookieCollection;
 use Cake\Http\Exception\MethodNotAllowedException;
 use Cake\Utility\Hash;
+use Closure;
 use InvalidArgumentException;
 use Laminas\Diactoros\PhpInputStream;
 use Laminas\Diactoros\Stream;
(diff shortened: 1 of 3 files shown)


markstory (approved):

src/Http/ServerRequest.php
```diff
@@ -733,17 +734,18 @@ public function isAll(array $types): bool
      * `addDetector('extension', ['param' => '_ext', 'options' => ['pdf', 'csv']]`
      *
      * @param string $name The name of the detector.
-     * @param callable|array $callable A callable or options array for the detector definition.
+     * @param \Closure|array $callable A Closure or options array for the detector definition.
      * @return void
      */
-    public static function addDetector(string $name, callable|array $callable): void
+    public static function addDetector(string $name, Closure|array $callable): void
```
saeideng: the `$callable` is not proper name right now IMO

ADmad: Yes, I am going to change it to `$detector`. But that will go into `4.next` first as that's where we have done all the argument renames.

src/ORM/Behavior/TreeBehavior.php
```diff
@@ -941,14 +943,20 @@ protected function _scope(Query $query): Query
...
         }
-        if (is_callable($scope)) {
+        if ($scope instanceof Closure) {
             return $scope($query);
         }
 
-        return $query;
+        throw new UnexpectedValueException(sprintf(
+            'Scope must be an `array` or `Closure`. Got `%s` instead.',
+            get_debug_type($scope)
```
othercorey: Would like to have tests for exceptions if possible. It says closures aren't tested either?

ADmad: Thinking about it I don't see why any type check is needed here. Scope is just conditions so they can be directly passed to `$query->where()` if non null.

PR #15889, opened 2021-09-13
Fix some grammar topics.

dereuromark:
The IDE reported this as grammar smells.
Turns out the "whether" itself is already a boolean decision point, the or not just not needed.


othercorey:
Looks good.

PR #15891, opened 2021-09-13
Use new string functions instead of strpos().

ADmad:

### DIFF
--- a/src/Console/BaseCommand.php
+++ b/src/Console/BaseCommand.php
@@ -45,7 +45,7 @@ abstract class BaseCommand implements CommandInterface
      */
     public function setName(string $name)
     {
-        if (strpos($name, ' ') < 1) {
+        if (!strpos($name, ' ')) {
             throw new InvalidArgumentException(
                 "The name '{$name}' is missing a space. Names should look like `cake routes`"
             );
--- a/src/Utility/Text.php
+++ b/src/Utility/Text.php
@@ -801,7 +801,7 @@ class Text
             $len = self::_strlen($part, $options);
             if ($offset !== 0 || $totalLength + $len > $length) {
                 if (
-                    strpos($part, '&') === 0
+                    str_starts_with($part, '&')
                     && preg_match($pattern, $part)
                     && $part !== html_entity_decode($part, ENT_HTML5 | ENT_QUOTES, 'UTF-8')
                 ) {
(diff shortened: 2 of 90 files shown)


src/Console/BaseCommand.php
```diff
@@ -45,7 +45,7 @@ abstract class BaseCommand implements CommandInterface
      */
     public function setName(string $name)
     {
-        if (strpos($name, ' ') < 1) {
+        if (!strpos($name, ' ')) {
```
othercorey, line 48: That's an odd check. Might be cleaner to trim it first.

src/Utility/Text.php
```diff
@@ -801,7 +801,7 @@ protected static function _substr(string $text, int $start, ?int $length, array
             $len = self::_strlen($part, $options);
             if ($offset !== 0 || $totalLength + $len > $length) {
                 if (
-                    strpos($part, '&') === 0
+                    mb_strpos($part, '&') === 0
```
othercorey: Should the preg_ functions be multi-byte or unicode?

othercorey:
That is a lot. They all look good. The comments I left are probably separate PRs.

src/Console/BaseCommand.php
markstory, line 48:
```suggestion
        if (!str_contains($name, ' ')) {
```

Isn't this a 'not contains' check?

ADmad: Can't just trim it off, there's a specific test which throws an exception for leading space.

src/Utility/Text.php
markstory: The pattern on 785 is looking for HTML entities and hexadecimal entities. Those patterns shouldn't require unicode support.

ADmad: IDK, I just changed this one as other functions in the method where `mb_*`.

src/Console/BaseCommand.php
othercorey: Apparently, there is a special check for starting with a space for some reason.

ADmad: @markstory `!str_contains()` will pass for string with leading space ` foo` which we don't want to allow. There's also a test for that specific case.

markstory: Oh fun.

PR #15897, opened 2021-09-14
Fix namespaced plugin translation file path

dereuromark:
Fixes backlog item for 5.0

The namespaced plugins should not collide with non-namespaced ones in terms of translation PO file etc.

Refs https://github.com/cakephp/docs/pull/6585


src/I18n/MessagesFileLoader.php
```diff
@@ -172,12 +159,38 @@ public function translationsFolders(): array
...
+     * @param array<string> $folders Folders
+     * @param string $name File name
+     * @param string $ext File extension
+     * @return string|null File if found
+     */
+    protected function translationFile(array $folders, string $name, string $ext): ?string
+    {
+        $file = null;
+
+        $pos = strpos($name, '/');
```
ADmad: I don't think first checking for slash is necessary, can directly use str_replace(). Same in the I18nExtract command too.

tests/test_app/Plugin/PluginNamespaced/MyNamespacedPlugin/src/Plugin.php
```diff
@@ -0,0 +1,10 @@
+<?php
+declare(strict_types=1);
+
+namespace PluginNamespaced\MyNamespacedPlugin;
+
+use Cake\Core\BasePlugin;
+
+class Plugin extends BasePlugin
```
ADmad: Isn't there an already namespaced plugin which can be used instead of creating an additional test plugin?

dereuromark: The company one. but I would rather migrate to a more clearly "named" one if possible.
As it wasnt clear from the start that it already exists as such.

dereuromark: I am also fine reverting to the company one, if that is preferred.

src/I18n/MessagesFileLoader.php
```diff
@@ -55,7 +55,7 @@ class MessagesFileLoader
      * Creates a translation file loader. The file to be loaded corresponds to
      * the following rules:
      *
-     * - The locale is a folder under the `Locale` directory, a fallback will be
+     * - The locale is a folder under the `Locale/` directory, a fallback will be
```
dereuromark: This should be `locales/`, right?

ADmad (approved):

othercorey:
@markstory Can you check and merge?

PR #15902, opened 2021-09-14
Add per-TestCase fixture strategy setup and teardown

othercorey:
This allows support for more specialized strategies. Changing the interface in the future will be difficult.

### DIFF
--- a/src/TestSuite/Fixture/FixtureStrategyInterface.php
+++ b/src/TestSuite/Fixture/FixtureStrategyInterface.php
@@ -22,6 +22,13 @@ namespace Cake\TestSuite\Fixture;
 interface FixtureStrategyInterface
 {
     /**
+     * Called before all tests are run for TestCase.
+     *
+     * @return void
+     */
+    public function setupStrategy(): void;
+
+    /**
      * Called before each test run in each TestCase.
      *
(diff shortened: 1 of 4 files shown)


markstory:
What kind of scenarios do you see the additional hook methods being useful in? I'm not against them but I also can't think of when I would need them either.

othercorey:
> What kind of scenarios do you see the additional hook methods being useful in? I'm not against them but I also can't think of when I would need them either.

The scenario I was trying to keep open is static fixtures that dont need to rebuild each test. 

Might be useful only for large numbers of tests.

othercorey:
I suppose you could setup a temp db with this but I wasn’t planning anything like that.

markstory:
> The scenario I was trying to keep open is static fixtures that dont need to rebuild each test.

Not rebuilding the fixture objects multiple times per test run could help with performance. I don't think we're getting that with this change though as we dereference the fixtures in the teardownafterclass step.

othercorey:
> Not rebuilding the fixture objects multiple times per test run could help with performance. I don't think we're getting that with this change though as we dereference the fixtures in the teardownafterclass step.

This solution would allow static fixtures for a TestCase not an entire test run. 

If we wanted static textures for an entire test run, then we would have to keep a static array of fixtures in the strategy.

PR #10161, opened 2017-02-03
[RFC] Association name setter

robertpustulka:
Following [comments](https://github.com/cakephp/cakephp/pull/10142).

Association names are mutable, but that change isn't reflected in a collection nor the target table. Changing association name without manually reattaching it in a collection makes it unusable. `EagerLoader` [checks if these names match](https://github.com/cakephp/cakephp/blob/3.next/src/ORM/EagerLoader.php#L496) and throws an exception. Is this exception enough or should we disallow changing association name as it would probably break it?

In other words `AssociationCollection` key, `Association` name and target `Table` alias should be identical. Changing one of them would probably break things.

My way to improve this is to check if a new association name matches the target table alias **after** it has been resolved.

There's also an issue with `AssociationCollection` not reflecting association name change, but I'm not sure we can do anything about it without changing the way associations are collected.

PR #15909, opened 2021-09-16
Add missing types for association class properties.

ADmad:

### DIFF
--- a/psalm.xml
+++ b/psalm.xml
@@ -17,6 +17,7 @@
     </projectFiles>
 
     <issueHandlers>
+        <RedundantPropertyInitializationCheck errorLevel="suppress"/>
         <RedundantConditionGivenDocblockType errorLevel="suppress"/>
         <RedundantCastGivenDocblockType errorLevel="suppress"/>
         <DocblockTypeContradiction errorLevel="suppress"/>
--- a/src/ORM/Association.php
+++ b/src/ORM/Association.php
@@ -108,16 +108,16 @@ abstract class Association
     /**
      * The field name in the owning side table that is used to match with the foreignKey
      *
-     * @var array<string>|string|null
+     * @var array<string>|string
      */
-    protected array|string|null $_bindingKey = null;
+    protected array|string $_bindingKey;
 
     /**
      * The name of the field representing the foreign key to the table to load
      *
      * @var array<string>|string|false
      */
-    protected $_foreignKey;
+    protected array|string|false $_foreignKey;
 
     /**
      * A list of conditions to be always included when fetching records from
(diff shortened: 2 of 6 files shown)


src/ORM/Association.php
```diff
@@ -251,7 +251,7 @@ public function __construct(string $alias, array $options = [])
      */
     public function setName(string $name)
     {
-        if ($this->_targetTable !== null) {
+        if (isset($this->_targetTable)) {
```
dereuromark, line 254: Why do we need to use cloaking here?

ADmad: This setter itself doesn't make sense to me :slightly_smiling_face: 

- An association will always have a target table.
- Association name is ~~derived from the target table.~~ set by constructor.
- Based on this check and error message the association name should always match the target table alias.
- So what's the point of this setter??

dereuromark: Sure, that aside I was just wondering about the !== null change
Is that a PHP 8 thing? Due to the typed props?

ADmad: Hmm so here's the thing; association name is set from constructor. The `targetTable` if not specified in constructor option is derived by using the association name as a table alias to get it from the table registry.

So theoretically one can use `setName()` to change association name after object construction until the `targetTable` is derived. But I don't see any practical use case for this.

I suggest simply dropping the setter in 5.0.

ADmad: > Sure, that aside I was just wondering about the !== null change
> Is that a PHP 8 thing? Due to the typed props?

Yes, for typed properties non-initialized vs null are treated differently. If a property is non-nullable you can't use `$this->foo` until it's initialized to an appropriate value first.

ADmad: This is the PR which adds the check for name vs table alias https://github.com/cakephp/cakephp/pull/10161. 
> Association names are mutable, but that change isn't reflected in a collection nor the target table. Changing association name without manually reattaching it in a collection makes it unusable.

So again let's just remove this name setter.

psalm.xml
```diff
@@ -17,6 +17,7 @@
     </projectFiles>
 
     <issueHandlers>
+        <RedundantPropertyInitializationCheck errorLevel="suppress"/>
```
othercorey, line 20: Do we want to suppress this globally? Is this the "should be initialized in constructor" error or a different initialization error?

src/ORM/Association.php
markstory: > So again let's just remove this name setter.

Yes. This method is/was not a good one.

markstory (approved):

psalm.xml
ADmad: It's similar to the "should be initialized in constructor" error. It's thrown for `isset()` checks on the properties which are initialized in the getters instead of constructors.

src/ORM/Association.php
ADmad: > Yes. This method is/was not a good one.

Okay, I'll deprecate it in 4.next first and then remove here.

psalm.xml
othercorey: If we plan to avoid the typed property initialization, then I agree it should just be suppressed. Wasn't sure the policy we wanted to adopt.

ADmad: In general initialising properties in getters instead of constructors is preferred at times. Also the psalm check/error is about redundancy, so I don't think we lose anything by suppressing this message.

PR #15910, opened 2021-09-16
Add param types.

ADmad:

### DIFF
--- a/tests/TestCase/Console/HelpFormatterTest.php
+++ b/tests/TestCase/Console/HelpFormatterTest.php
@@ -183,7 +183,7 @@ txt;
         $parser->addOption('test', ['help' => 'A test option.'])
                ->addOption('number', [
                    'help' => 'The number',
-                   'default' => 2,
+                   'default' => '2',
                ])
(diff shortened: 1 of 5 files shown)


tests/TestCase/Console/HelpFormatterTest.php
```diff
@@ -183,7 +183,7 @@ public function testHelpWithOptions(): void
         $parser->addOption('test', ['help' => 'A test option.'])
                ->addOption('number', [
                    'help' => 'The number',
-                   'default' => 2,
+                   'default' => '2',
```
othercorey, line 186: I wasn't sure about this one. Should we support setting an integer and cast to string? It's true that all values read in would be strings so they user shouldn't falsely expect an integer value.

@markstory

ADmad: Given that all values coming from command line will be strings I don't see any reason for supporting integer here.

markstory: I think only accepting strings makes sense.

PR #15912, opened 2021-09-16
Adjust stack frame for deprecation warnings

othercorey:
These deprecation paths should probably be where deprecationWarning() is called.

### DIFF
--- a/src/Log/Engine/SyslogLog.php
+++ b/src/Log/Engine/SyslogLog.php
@@ -94,7 +94,8 @@ class SyslogLog extends BaseLog
         if (isset($config['format'])) {
             deprecationWarning(
                 '`format` option is now deprecated in favor of custom formatters. ' .
-                'Switching to `LegacySyslogFormatter`.'
+                'Switching to `LegacySyslogFormatter`.',
+                0
             );
             /** @psalm-suppress DeprecatedClass */
             $config['formatter'] = [
--- a/tests/TestCase/Database/Type/DateTimeFractionalTypeTest.php
+++ b/tests/TestCase/Database/Type/DateTimeFractionalTypeTest.php
@@ -48,7 +48,7 @@ class DateTimeFractionalTypeTest extends TestCase
         $this->driver = $this->getMockBuilder('Cake\Database\Driver')->getMock();
 
         Configure::write('Error.ignoredDeprecationPaths', [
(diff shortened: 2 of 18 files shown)


src/Log/Engine/SyslogLog.php
```diff
@@ -94,7 +94,8 @@ public function __construct(array $config = [])
         if (isset($config['format'])) {
             deprecationWarning(
                 '`format` option is now deprecated in favor of custom formatters. ' .
-                'Switching to `LegacySyslogFormatter`.'
+                'Switching to `LegacySyslogFormatter`.',
+                0
```
markstory, line 98: Won't this result in the deprecation file reference now pointing to the constructor of SyslogLog?

tests/TestCase/Database/Type/DateTimeFractionalTypeTest.php
```diff
@@ -48,7 +48,7 @@ public function setUp(): void
         $this->driver = $this->getMockBuilder('Cake\Database\Driver')->getMock();
 
         Configure::write('Error.ignoredDeprecationPaths', [
-            'tests/TestCase/Database/Type/DateTimeFractionalTypeTest.php',
+            'src/I18n/Time.php',
```
markstory, line 51: I think this path is less helpful for *fixing* the deprecation. You no longer know where in the application code the deprecated call is coming from. The thinking behind that was you could start off with an ignore of `src/*` and then get to the point where you have a few entries like `'src/Controller/*', 'src/Model/Table/*'` or you could ignore entire plugins with paths like `vendor/markstory/asset_compress/*`

With this change you can more easily silence a high-volume warning, but you can only see the source in the message so you have a much harder time finding where the deprecations are coming from. so you can fix them.

othercorey: If you look at the other paths from Time and Date, they were in factories and such. I wasn't really trying to hide the db types. I was assuming people would use Time globally or not and wanted to know it's deprecated.

src/Log/Engine/SyslogLog.php
othercorey: Not sure about this one. I don't know if the parent stack frame is relevant or if it's just buried deep in a config factory.

markstory: True, it is likely pointing to a different unhelpful place.

PR #15913, opened 2021-09-17
Deprecate Association::setName().

ADmad:
Changing the association name after object creation isn't reflected in the collection
nor the target table. Hence it's quite limiting and best to deprecate and remove this method.


othercorey:
Looks good.

PR #15914, opened 2021-09-17
cake 5: Remove engine config from NumberHelper and TextHelper

othercorey:
I'm not sure why TextHelperTest used such extensive mocks instead of simply checking an actual call.


src/View/Helper/NumberHelper.php
```diff
@@ -99,7 +53,7 @@ public function __call(string $method, array $params): mixed
      */
     public function precision(string|float $number, int $precision = 3, array $options = []): string
     {
-        return $this->_engine->precision($number, $precision, $options);
+        return Number::precision($number, $precision, $options);
```
ADmad: This proxying would also be covered by `__cal()` isn't it? I think we should remove all such methods which don't do anything extra and just add `@method` annotations instead for IDE autocompletion.

othercorey: Would the one-line comment for the method annotation just be "See Number::method()"?

I'm pretty sure we can specify defaults in annotations, but should check.

ADmad: Yeah such a short comment would be fine. Primarily the method signature is what's useful for IDEs.

dereuromark:
Doesnt this remove the possibility of Auto forwarding to the engine part? For all methods there?
Is it too costly or why removing?

othercorey:
> Doesnt this remove the possibility of Auto forwarding to the engine part? For all methods there?
> Is it too costly or why removing?

The method proxying is still there.

This is due to the lack of "type" for the engine. We have to make all engines extend Number/Text or remove the engine. After discussing, it seemed like customizing the helper was easier than writing an engine that extends Number/Text.

ADmad (approved):

PR #15917, opened 2021-09-17
Make ShadowTable the default strategy for TranslateBehavior.

ADmad:
The EavStrategy which uses one row per field is quite wasteful and can
also be a performance killer for large apps.

### DIFF
--- a/src/ORM/Behavior/TranslateBehavior.php
+++ b/src/ORM/Behavior/TranslateBehavior.php
@@ -18,7 +18,7 @@ namespace Cake\ORM\Behavior;
 
 use Cake\I18n\I18n;
 use Cake\ORM\Behavior;
-use Cake\ORM\Behavior\Translate\EavStrategy;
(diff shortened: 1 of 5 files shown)


src/ORM/Behavior/TranslateBehavior.php
```diff
@@ -70,7 +71,7 @@ class TranslateBehavior extends Behavior implements PropertyMarshalInterface
      * @var string
      * @psalm-var class-string<\Cake\ORM\Behavior\Translate\TranslateStrategyInterface>
      */
-    protected static string $defaultStrategyClass = EavStrategy::class;
+    protected static string $defaultStrategyClass = ShadowTableStrategy::class;
```
markstory, line 74: I'm in favour of this change but I think we'll need to update the translate behavior docs. They currently make a bunch of assumptions that the reader is using the Eav table.

ADmad: I have opened a PR for 4.next to get things started and promote the use of ShadowTable.

PR #15920, opened 2021-09-17
Fix TypeError in CsrfProtectionMiddleware

### LINKED ISSUE #15868: TypeError in CsrfProtectionMiddleware
This is a (multiple allowed):

* [x] bug
* [ ] enhancement

* CakePHP Version: 4.2.8
* Platform and Target: nginx, php-fpm 7.4.23

### What you did
After update from 4.1.7, where issue with malformed/crafted cookies/form-params was manifested by NoticeException **Array to string conversion** on line 300 of vendor/cakephp/cakephp/src/Http/Middleware/CsrfProtectionMiddleware.php

In 4.2.8 it's manifested by TypeError **strlen() expects parameter 1 to be string, bool given** on line 281 of vendor/cakephp/cakephp/src/Http/Middleware/CsrfProtectionMiddleware.php

### What happened
Someone enumerating/bruteforcing endpoints crafted request and provided invalid values where CsrfProtectionMiddleware expected valid, previously framework-set, data.

stacktrace here:
```
#0 /var/www/cakephp.app/vendor/cakephp/cakephp/src/Http/Middleware/CsrfProtectionMiddleware.php(281): strlen()
#1 /var/www/cakephp.app/vendor/cakephp/cakephp/src/Http/Middleware/CsrfProtectionMiddleware.php(153): Cake\Http\Middleware\CsrfProtectionMiddleware->saltToken()
#2 /var/www/cakephp.app/vendor/cakephp/cakephp/src/Http/Runner.php(73): App\Middleware\CsrfUserProtectionMiddleware->process()
#3 /var/www/cakephp.app/vendor/cakephp/cakephp/src/Error/Middleware/ErrorHandlerMiddleware.php(126): Cake\Http\Runner->handle()
#4 /var/www/cakephp.app/vendor/cakephp/cakephp/src/Http/Runner.php(73): Cake\Error\Middleware\ErrorHandlerMiddleware->process()
#5 /var/www/cakephp.app/vendor/cakephp/cakephp/src/Http/Runner.php(73): App\Middleware\TrustProxyMiddleware->process()
#6 /var/www/cakephp.app/vendor/cakephp/cakephp/src/Http/Runner.php(58): Cake\Http\Runner->handle()
#7 /var/www/cakephp.app/vendor/cakephp/cakephp/src/Http/Server.php(90): Cake\Http\Runner->run()
#8 /var/www/cakephp.app/webroot/index.php(40): Cake\Http\Server->run()
#9 {main}
```

Invalid value in cookie/form-params, should be detected and handled, instead of relying on php TypeError faults
Base64_Decode return signature is string|false, and for any kind of invalid value, this should be handled gracefuly, so the client (web user) is not permanently cut out from the app, just because he has invalid cookie data, that prevents him from being able to access the webapp

markstory:
When a cookie contains non-empty data that does not match the old token format and also is not base64 encoded we should assume it is invalid and generate a new CSRF token on GET requests. For non-GET requests this will result in a CSRF validation error.

Fixes #15868

### DIFF
--- a/src/Http/Middleware/CsrfProtectionMiddleware.php
+++ b/src/Http/Middleware/CsrfProtectionMiddleware.php
@@ -149,7 +149,11 @@ class CsrfProtectionMiddleware implements MiddlewareInterface
         $cookieData = Hash::get($cookies, $this->_config['cookieName']);
 
         if (is_string($cookieData) && strlen($cookieData) > 0) {
-            $request = $request->withAttribute('csrfToken', $this->saltToken($cookieData));
+            try {
+                $request = $request->withAttribute('csrfToken', $this->saltToken($cookieData));
+            } catch (RuntimeException $e) {
+                $cookieData = null;
+            }
         }
 
         if ($method === 'GET' && $cookieData === null) {
@@ -277,6 +281,10 @@ class CsrfProtectionMiddleware implements MiddlewareInterface
             return $token;
         }
         $decoded = base64_decode($token, true);
+        if ($decoded === false) {
+            throw new RuntimeException('Invalid token data.');
+        }
+
         $length = strlen($decoded);
         $salt = Security::randomBytes($length);
         $salted = '';
--- a/src/Http/Middleware/CsrfProtectionMiddleware.php
+++ b/src/Http/Middleware/CsrfProtectionMiddleware.php
@@ -23,6 +23,7 @@ use Cake\Http\Exception\InvalidCsrfTokenException;
 use Cake\Http\Response;
 use Cake\Utility\Hash;
 use Cake\Utility\Security;
+use InvalidArgumentException;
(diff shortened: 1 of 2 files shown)


src/Http/Middleware/CsrfProtectionMiddleware.php
```diff
@@ -277,6 +281,10 @@ public function saltToken(string $token): string
             return $token;
         }
         $decoded = base64_decode($token, true);
+        if ($decoded === false) {
+            throw new RuntimeException('Invalid token data.');
```
othercorey: Should this be an InvalidArgumentException? It makes sense as RuntimeException if thrown in process(). Otherwise this looks good.

markstory: Good idea.

PR #15922, opened 2021-09-18
Verify deprecation path for various warnings

othercorey:
Not going to add checks for all warnings, just a few that need to be in the right path.

### DIFF
--- a/tests/TestCase/Controller/ControllerFactoryTest.php
+++ b/tests/TestCase/Controller/ControllerFactoryTest.php
@@ -68,24 +68,22 @@ class ControllerFactoryTest extends TestCase
     /**
      * Test building a prefixed app controller.
      */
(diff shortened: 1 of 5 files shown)


tests/TestCase/Controller/ControllerFactoryTest.php
```diff
@@ -68,24 +68,22 @@ public function testApplicationController(): void
     /**
      * Test building a prefixed app controller.
      */
-    public function testPrefixedAppControllerDeprecated(): void
```
othercorey, line 71: This does not throw a deprecation warning.

markstory:
Having `$this->deprecated()` also assert that a deprecation was created could be a nice way to save LOC in tests. I think @saeideng was interested in looking into that.

Issue #15817, opened 2021-09-03
Sql Server unit tests times have increase 100%


othercorey:
This seems to have started the same time as the new fixture re-work. The timing could be a coincidence and the issue could be on github actions side.

ADmad:
Until we can find a fix maybe we can setup the github action to run the job for Sql Server only on a full moon :trollface:.

PR #15923, opened 2021-09-18
Optimize when constraints are disabled for fixtures

othercorey:
Fixes https://github.com/cakephp/cakephp/issues/15817

This can be simplified in 5.0 when TestFixture doesn't generate a fake schema based on `$fields`. We can rely on the schema returned by `getTableSchema()` to be the reflected schema.

Mysql does not support truncating a table with references to it at all - even if there are no records.

### DIFF
--- a/src/TestSuite/Fixture/FixtureHelper.php
+++ b/src/TestSuite/Fixture/FixtureHelper.php
@@ -17,9 +17,14 @@ declare(strict_types=1);
 namespace Cake\TestSuite\Fixture;
 
 use Cake\Core\Configure;
+use Cake\Database\Connection;
 use Cake\Database\Driver\Postgres;
+use Cake\Database\Driver\Sqlite;
+use Cake\Database\Driver\Sqlserver;
+use Cake\Database\Schema\TableSchema;
 use Cake\Datasource\ConnectionInterface;
 use Cake\Datasource\ConnectionManager;
+use Cake\Datasource\FixtureInterface;
 use Closure;
 use UnexpectedValueException;
 
@@ -130,7 +135,15 @@ class FixtureHelper
     public function insert(array $fixtures): void
     {
         $this->runPerConnection(function (ConnectionInterface $connection, array $groupFixtures) {
-            if ($connection->getDriver() instanceof Postgres) {
+            if ($connection instanceof Connection) {
+                $sortedFixtures = $this->sortByConstraint($connection, $groupFixtures);
+            }
+
+            if (isset($sortedFixtures)) {
+                foreach ($sortedFixtures as $fixture) {
+                    $fixture->insert($connection);
+                }
+            } elseif ($connection->getDriver() instanceof Postgres) {
                 // disabling foreign key constraints is only valid in a transaction
                 $connection->transactional(function () use ($connection, $groupFixtures) {
                     $connection->disableConstraints(function () use ($connection, $groupFixtures) {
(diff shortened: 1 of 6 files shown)


src/TestSuite/Fixture/FixtureHelper.php
```diff
@@ -159,11 +171,71 @@ public function insert(array $fixtures): void
...
+                    }
+                }
+            }
+
+            if (!isset($fixturesWithConstraint[$schema->name()])) {
+                $fixturesWithoutConstraint[$schema->name()] = $fixture;
+            }
+        }
+
+        return array_merge(array_values($fixturesWithoutConstraint), array_values($fixturesWithConstraint));
```
markstory: Couldn't we end up with fixtures in an order that we can't truncate because parent tables come are sorted after the children? Or are we relying on users to define fixtures in the required order? I am ok with people needing to put their fixtures in the correct order to satisfy constraints.  If we have that requirement  (which I think we have always had) do we need to sort fixtures?

othercorey: The fixtures are reversed when truncating.

So, there is a check to see if any fixtures with constraints conflict with each other. If there are any cross-dependencies, then it bails.

However, that check is incomplete. It needs to happen after all the fixtures are categorized.

I did not try to sort cross-dependencies and just revert to disabling constraints.

src/TestSuite/Fixture/FixtureHelper.php
```diff
@@ -159,11 +171,71 @@ public function insert(array $fixtures): void
...
+                    $driver instanceof Sqlite ||
+                    $driver instanceof Sqlserver
+                )
+            ) {
+                foreach (array_reverse($sortedFixtures) as $fixture) {
                     $fixture->truncate($connection);
                 }
-            });
+            } else {
+                $connection->disableConstraints(function () use ($connection, $groupFixtures) {
```
markstory, line 192: I wonder if we should have a wrapper on the `Connection`/`Driver` that can encapsulate these platform differences better. I'm thinking a method like 

```php
$connection->prepareTruncate(function ($connection) use ($groupFixtures) {
  foreach ($groupFixtures as $fixture) {
    $fixture->truncate($connection); 
 }
});
```

Could work. The naming might be awkward, but it lets us encapsulate the driver specific details into a method with a similar name, and keep most of the logic in the driver. This could be helpful if oracle or a future driver needs different semantics around transactions & constraints being disabled.

src/TestSuite/Fixture/FixtureHelper.php
```diff
@@ -159,11 +171,71 @@ public function insert(array $fixtures): void
...
+            $schema = $schemas[$tableName];
+            foreach ($schema->constraints() as $constraintName) {
+                $constraint = $schema->getConstraint($constraintName);
+                if (!$constraint) {
+                    continue;
+                }
+
+                if ($constraint['type'] === TableSchema::CONSTRAINT_FOREIGN) {
+                    $fixturesWithConstraint[$schema->name()] = $fixture;
+                    if (isset($fixturesWithConstraint[$constraint['references'][0]])) {
```
markstory: Might be good to have a comment here around cross referenced fixtures causing an abort. I didn't catch that on the first read.

src/TestSuite/Fixture/FixtureHelper.php
othercorey: Do we need to fix it for 4.3? I'm not sure my first thought will be the cleanest api especially with driver clean up in 5.0.

src/TestSuite/Fixture/FixtureHelper.php
othercorey: I split the foreign references code into a separate function so it's easier to read.

Fixed the cross dependency check to check all references and added a comment explaining why.

PR #15924, opened 2021-09-18
Remove unneeded property overriding.

ADmad:
Parent class has the same value.


ADmad:
Apparently the order of the parts in `$_selectParts` matters.

PR #15925, opened 2021-09-18
Make static analyzers happy.

ADmad:

### DIFF
--- a/src/Database/PostgresCompiler.php
+++ b/src/Database/PostgresCompiler.php
@@ -36,7 +36,9 @@ class PostgresCompiler extends QueryCompiler
     protected bool $_quotedSelectAliases = true;
 
     /**
-     * @inheritDoc
+     * {@inheritDoc}
+     *
+     * @var array<string, string>
      */
     protected array $_templates = [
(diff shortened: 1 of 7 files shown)


src/Database/PostgresCompiler.php
```diff
@@ -36,7 +36,9 @@ class PostgresCompiler extends QueryCompiler
     protected bool $_quotedSelectAliases = true;
 
     /**
-     * @inheritDoc
+     * {@inheritDoc}
+     *
+     * @var array<string, string>
```
ADmad, line 41: This is needed because of a potential bug in psalm. It doesn't seem to inherit the type properly when `@inheritDoc` is used. I have opened an issue for it.

src/Database/Type/DateType.php
```diff
@@ -32,8 +32,11 @@ class DateType extends DateTimeType
     protected string $_format = 'Y-m-d';
 
     /**
-     * @inheritDoc
+     * {@inheritDoc}
+     *
+     * @var array<string>
      */
+
```
dereuromark: One too many

PR #15928, opened 2021-09-19
5.0 Database Driver

ADmad:
- Changed return type of `DriverInterface::connect()` &&  `Connection::connect()` from `bool` to `void`. They always returned `true` on success and threw an exception on failure making a `bool` return type redundant.
- Dropped `DriverInterface::setConnection()`. It was used only for testing.
- Updated signature of `DriverInterface::lastInsertId()` to better correspond to the signature of `PDO::lastInsertId()`.
-  Dropped `DriverInterface::getConnectRetries()`. It didn't make much sense as an interface method and was used only in tests.
- Dropped DriverInterface::newCompiler(). It returns `QueryCompiler` which is marked as internal.
- IMO `DriverInterface::getConnection()` should also be dropped. The `PDO` instance is an internal detail of the driver which doesn't need to be exposed. `MysqlStatement` seems to be the only class which accesses the driver's `PDO` instance: https://github.com/cakephp/cakephp/blob/2918a44d1df727e542e12201434d09f3de6e3949/src/Database/Statement/MysqlStatement.php#L38
We could instead add a wrapper method on the driver class for it.


ADmad:
Can someone look into the failing Sqlserver test please. I am not sure why it's expected to throw an exception and I don't have Sqlserver handy locally.

othercorey:
> Can someone look into the failing Sqlserver test please. I am not sure why it's expected to throw an exception and I don't have Sqlserver handy locally.

The test seems to have nothing to do with persistent = false should simply throw because it's an invalid connection that can't connect.

ADmad:
> The test seems to have nothing to do with persistent = false should simply throw because it's an invalid connection that can't connect.

We should rename the test method. Still I have no idea why the test is failing.

othercorey:
> > The test seems to have nothing to do with persistent = false should simply throw because it's an invalid connection that can't connect.
> 
> We should rename the test method. Still I have no idea why the test is failing.

It looks overly complex for what its trying to do. My guess is they are just checking it doesn’t throw an InvalidArgumemtException like the test below it.

src/Database/DriverInterface.php
```diff
@@ -241,10 +225,9 @@ public function schema(): string;
      * Returns last id generated for a table or sequence in database.
      *
      * @param string|null $table table name or sequence to get last insert value from.
-     * @param string|null $column the name of the column representing the primary key.
-     * @return string|int
+     * @return string
      */
-    public function lastInsertId(?string $table = null, ?string $column = null): string|int;
+    public function lastInsertId(?string $table = null): string;
```
markstory, line 230: Did we come up with an alternate solution to avoiding having to fetch results in postgres? That is the historical context for this parameter as added in 7937b340d59.  I saw that we still use `RETURNING *` in postgres. Without that the ORM will need to provide sequence names to use PDO's lastInsertId method.

ADmad: I don't understand, `Driver::lastInsertId()` uses `PDO::lastInsertId()` which only has one argument. There's no `Postgres::lastInsertId()`. As you can see there are no test failures for postgres after this change.

Perhaps you are referring to `StatementDecorator::lastInsertId()` which still does have (and actually uses) the column name argument https://github.com/cakephp/cakephp/blob/e1c59462a4b17c5b9ef1ba80e6206857bdd82471/src/Database/Statement/StatementDecorator.php#L341-L352

tests/TestCase/Database/ConnectionTest.php
```diff
@@ -123,9 +123,9 @@ public function getMockFormDriver()
     /**
      * Tests connecting to database
      */
-    public function testConnect(): void
+    public function testIsConnect(): void
```
othercorey:
```suggestion
    public function testIsConnected(): void
```

tests/TestCase/Database/ConnectionTest.php
```diff
@@ -215,21 +215,6 @@ public function testWrongCredentials(): void
         $this->assertInstanceOf('PDOException', $e->getPrevious());
     }
 
-    public function testConnectRetry(): void
-    {
-        $this->skipIf(!ConnectionManager::get('test')->getDriver() instanceof Sqlserver);
-
-        $connection = new Connection(['driver' => 'RetryDriver']);
```
othercorey, line 222: This driver triggers a specific failure to ensure the retry-on-failure-code tries again.

src/Database/DriverInterface.php
markstory: You're right. I conflated the two methods together.

ADmad:
Is this good to merge?

othercorey:
Merging since I don't see any objections in the reviews.

Issue #15195, opened 2020-11-28
context parameter of error handler deprecated


othercorey:
https://www.php.net/manual/en/function.set-error-handler.php

> Warning This parameter has been DEPRECATED as of PHP 7.2.0. Relying on it is highly discouraged.

The context parameter of error handler callback is currently used by ErrorHandler.

othercorey:
This was removed in 8.0. We'll have to remove the context data.

> Warning
This parameter has been DEPRECATED as of PHP 7.2.0, and REMOVED as of PHP 8.0.0. If your function defines this parameter without a default, an error of "too few arguments" will be raised when it is called.

othercorey:
This change causes a bug in the deprecations panel in debug kit since the $context was used to find the filename with the correct stack offset.

markstory:
@othercorey The loss of $context leads me to think that the error callback idea we talked about before has some legs. I'll try to get a proposal for that together in the upcoming weeks.

othercorey:
This should be fixed by the new error handler implementation.

PR #15929, opened 2021-09-20
Remove context from error output

othercorey:
Refs https://github.com/cakephp/cakephp/issues/15195

I don't see any test for the Debugger output so this might be breaking some formatting.

The $context parameter to _handleError() is left in for users to manually call, but the expected $context parameter from error handler is removed. This lets users passed in a `_trace_frame_offset` value for custom handlers.

Not sure any other context value should be considered relevant in debug output since it was supposed to be from the error handler.

### DIFF
--- a/src/Error/BaseErrorHandler.php
+++ b/src/Error/BaseErrorHandler.php
@@ -176,7 +176,6 @@ abstract class BaseErrorHandler
                 unset($context['_trace_frame_offset']);
             }
             $data += [
-                'context' => $context,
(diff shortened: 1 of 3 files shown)


markstory:
> This lets users passed in a _trace_frame_offset value for custom handlers.

This key was used by DebugKit to allow the error handler there to show the correct file and line numbers. I'm not sure how we'll be able to handle that in a 5.x world.

othercorey:
We can figure this out some other time

PR #15931, opened 2021-09-20
cake 5: Remove fields and import from TestFixture

othercorey:


tests/TestCase/TestSuite/TestFixtureTest.php
```diff
@@ -220,7 +130,7 @@ public function testInsert(): void
 
         $query->expects($this->once())
             ->method('insert')
-            ->with(['name', 'created'], ['name' => 'string', 'created' => 'datetime'])
+            ->with(['author_id', 'title', 'body', 'published'], ['author_id' => 'integer', 'title' => 'string', 'body' => 'text', 'published' => 'string'])
```
othercorey, line 133: This was changed to ArticlesFixture in tests/Fixture because we don't build schema from $fields and that articles table is the one reflected.

PR #15932, opened 2021-09-21
Disable slash escaping by default in JsonFormatter

othercorey:
This leaves forward slashes alone. This should be preferred for logging format.


othercorey:
The context of this is PHP escapes forward slashes `\/` by default. This seems to be an auto-quoting mechanism for embedding json_encode output into html without injecting html. 

It's not required by JSON by default and looks cleaner without for paths.

ADmad (approved):

PR #15933, opened 2021-09-21
Rename BaseLog::resolve() to interpolate()

othercorey:
Let's use a more consistent name before this is finalized.

Perhaps we can replace it with a general interpolation helper in the future.

### DIFF
--- a/src/Log/Engine/BaseLog.php
+++ b/src/Log/Engine/BaseLog.php
@@ -121,21 +121,21 @@ abstract class BaseLog extends AbstractLogger
      * @param string $message The message to be formatted.
      * @param array $context Additional logging information for the message.
      * @return string
-     * @deprecated 4.3.0 Call `resolve()` directly from your log engine and format the message in a formatter.
+     * @deprecated 4.3.0 Call `interpolate()` directly from your log engine and format the message in a formatter.
      */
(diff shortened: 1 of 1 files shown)


ADmad:
Nice, `resolve()` did sound weird for what it does.

src/Log/Engine/BaseLog.php
```diff
@@ -121,21 +121,21 @@ public function scopes()
      * @param string $message The message to be formatted.
      * @param array $context Additional logging information for the message.
      * @return string
-     * @deprecated 4.3.0 Call `resolve()` directly from your log engine and format the message in a formatter.
+     * @deprecated 4.3.0 Call `interplate()` directly from your log engine and format the message in a formatter.
      */
     protected function _format(string $message, array $context = []): string
     {
-        return $this->resolve($message, $context);
+        return $this->interplate($message, $context);
```
ADmad:
```suggestion
        return $this->interpolate($message, $context);
```

othercorey: yea I saw this too.

othercorey:
> Nice, `resolve()` did sound weird for what it does.

I remember thinking it special for some reason, but re-reading standards, I was wrong.

PR #15935, opened 2021-09-21
improve deprecated method

saeideng:
This PR is a draft yet

### DIFF
--- a/src/Core/functions.php
+++ b/src/Core/functions.php
@@ -311,10 +311,13 @@ if (!function_exists('deprecationWarning')) {
 
         static $errors = [];
         $checksum = md5($message);
-        if (isset($errors[$checksum])) {
+        $duplicate = (bool)Configure::read('Error.enableDuplicatedDeprecations', false);
+        if (isset($errors[$checksum]) && !$duplicate) {
             return;
         }
-        $errors[$checksum] = true;
+        if (!$duplicate) {
+            $errors[$checksum] = true;
+        }
 
         trigger_error($message, E_USER_DEPRECATED);
     }
--- a/src/TestSuite/TestCase.php
+++ b/src/TestSuite/TestCase.php
@@ -203,6 +203,7 @@ abstract class TestCase extends BaseTestCase
      */
     public function deprecated(callable $callable): void
     {
+        /** @var boolean $deprecation */
         $deprecation = false;
         $previousHandler = set_error_handler(
             function ($code, $message, $file, $line, $context = null) use (&$previousHandler, &$deprecation) {
--- a/src/TestSuite/TestCase.php
+++ b/src/TestSuite/TestCase.php
@@ -203,7 +203,7 @@ abstract class TestCase extends BaseTestCase
      */
     public function deprecated(callable $callable): void
     {
-        /** @var boolean $deprecation */
+        /** @var bool $deprecation */
         $deprecation = false;
         $previousHandler = set_error_handler(
             function ($code, $message, $file, $line, $context = null) use (&$previousHandler, &$deprecation) {
--- a/src/TestSuite/TestCase.php
+++ b/src/TestSuite/TestCase.php
@@ -203,8 +203,14 @@ abstract class TestCase extends BaseTestCase
      */
     public function deprecated(callable $callable): void
     {
+        $duplicate = Configure::read('Error.enableDuplicatedDeprecations');
+        Configure::write('Error.enableDuplicatedDeprecations', true);
         /** @var bool $deprecation */
         $deprecation = false;
+
+        /**
+         * @psalm-suppress InvalidArgument
+         */
         $previousHandler = set_error_handler(
             function ($code, $message, $file, $line, $context = null) use (&$previousHandler, &$deprecation) {
                 if ($code == E_USER_DEPRECATED) {
@@ -223,6 +229,9 @@ abstract class TestCase extends BaseTestCase
             $callable();
         } finally {
             restore_error_handler();
+            if ($duplicate !== Configure::read('Error.enableDuplicatedDeprecations')) {
+                Configure::write('Error.enableDuplicatedDeprecations', $duplicate);
+            }
         }
         $this->assertTrue($deprecation, 'Should have at least one deprecation warning');
     }
--- a/src/TestSuite/TestCase.php
+++ b/src/TestSuite/TestCase.php
@@ -203,13 +203,27 @@ abstract class TestCase extends BaseTestCase
      */
     public function deprecated(callable $callable): void
     {
-        $errorLevel = error_reporting();
-        error_reporting($errorLevel & ~E_USER_DEPRECATED);
+        $deprecation = false;
+        $previousHandler = set_error_handler(
+            function ($code, $message, $file, $line, $context = null) use (&$previousHandler, &$deprecation) {
+                if ($code == E_USER_DEPRECATED) {
+                    $deprecation = true;
+
+                    return;
+                }
+                if ($previousHandler) {
+                    return $previousHandler($code, $message, $file, $line, $context);
+                }
+
+                return false;
+            }
+        );
         try {
             $callable();
         } finally {
-            error_reporting($errorLevel);
+            restore_error_handler();
         }
+        $this->assertTrue($deprecation, 'Should have at least one deprecation warning');
     }
 
     /**
--- a/tests/TestCase/Core/FunctionsTest.php
+++ b/tests/TestCase/Core/FunctionsTest.php
@@ -92,6 +92,27 @@ class FunctionsTest extends TestCase
     }
 
     /**
+     * Test deprecation warnings trigger only once
+     */
+    public function testDeprecationWarningTriggerOnlyOnce(): void
+    {
+        $message = 'Test deprecation warnings trigger only once';
+        try {
+            $this->withErrorReporting(E_ALL, function () use ($message): void {
+                deprecationWarning($message);
+            });
+            $this->fail();
+        } catch (\Exception $e) {
+            $this->assertStringContainsString($message, $e->getMessage());
+            $this->assertStringContainsString('TestCase.php', $e->getMessage());
+        }
+
+        $this->withErrorReporting(E_ALL, function () use ($message): void {
+            deprecationWarning($message);
+        });
+    }
+
+    /**
      * Test error messages coming out when deprecated level is on, not setting the stack frame manually
      */
     public function testDeprecationWarningEnabledDefaultFrame(): void
--- a/tests/TestCase/TestSuite/TestCaseTest.php
+++ b/tests/TestCase/TestSuite/TestCaseTest.php
@@ -147,6 +147,32 @@ class TestCaseTest extends TestCase
     }
 
     /**
+     * test deprecated
+     */
+    public function testDeprecated(): void
+    {
+        $this->deprecated(function () {
+            deprecationWarning('test');
+        });
+    }
+
+    /**
+     * test deprecated
+     */
+    public function testDeprecatedWithNoDeprecation(): void
+    {
+        $message = '';
+        try {
+            $this->deprecated(function () {
+            });
+        } catch (\Exception $e) {
+            $message = $e->getMessage();
+        } finally {
+            $this->assertStringStartsWith('Should have at least one deprecation warning', $message);
+        }
+    }
+
+    /**
      * Test that TestCase::setUp() backs up values.
      */
     public function testSetupBackUpValues(): void
--- a/tests/TestCase/TestSuite/TestCaseTest.php
+++ b/tests/TestCase/TestSuite/TestCaseTest.php
@@ -152,7 +152,7 @@ class TestCaseTest extends TestCase
     public function testDeprecated(): void
     {
         $this->deprecated(function () {
-            deprecationWarning('test');
+            trigger_error('deprecation message', E_USER_DEPRECATED);
         });
     }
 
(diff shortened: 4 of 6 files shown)


tests/TestCase/TestSuite/TestCaseTest.php
```diff
@@ -146,6 +146,32 @@ public function testWithErrorReportingWithException(): void
...
+
+    /**
+     * test deprecated
+     */
+    public function testDeprecatedWithNoDeprecation(): void
+    {
+        $message = '';
+        try {
+            $this->deprecated(function () {
+            });
```
saeideng, line 183: conflict when passing ```deprecationWarning('test');``` same as previous test `testDeprecated`

but ```deprecationWarning('other string');``` will stop assert failing in ` $this->assertTrue($deprecation, ...`

src/TestSuite/TestCase.php
```diff
@@ -203,13 +203,27 @@ public function withErrorReporting(int $errorLevel, callable $callable): void
      */
     public function deprecated(callable $callable): void
     {
-        $errorLevel = error_reporting();
-        error_reporting($errorLevel & ~E_USER_DEPRECATED);
+        $deprecation = false;
```
markstory, line 209:
```suggestion
        /** @var boolean $deprecation */
        $deprecation = false;
```

This might help with the error on 226

tests/TestCase/TestSuite/TestCaseTest.php
```diff
@@ -146,6 +146,32 @@ public function testWithErrorReportingWithException(): void
...
+        } catch (\Exception $e) {
+            $message = $e->getMessage();
+        } finally {
+            $this->assertStringStartsWith('Should have at least one deprecation warning', $message);
```
markstory: You could have this assertion in the catch block and skip finally. You can also use `$this->fail()` at the bottom of the try to ensure an exception was raised.

tests/TestCase/TestSuite/TestCaseTest.php
saeideng: this was new feature
Refs: https://github.com/cakephp/cakephp/commit/eef535c1153a9

tests/TestCase/TestSuite/TestCaseTest.php
saeideng: done

othercorey:
All the tests are still failing.

saeideng:
> All the tests are still failing.

yes now shows , some methods don't  have deprecation warning

```

1) Cake\Test\TestCase\Controller\Component\PaginatorComponentTest::testDeprecatedPathMergeOptionsExtraWhitelist
Failed asserting that exception of type "PHPUnit\Framework\Error\Deprecated" is thrown.

2) Cake\Test\TestCase\Controller\ComponentTest::testDeprecatedEventShutdownPath
Failed asserting that exception of type "PHPUnit\Framework\Error\Deprecated" is thrown.
```
these are strange


addressed failures have not `deprecated()` method , but other test case above it  have, with error ... 

and when test these methods  individually no error occurs
```
vendor/bin/phpunit tests/TestCase --filter testDeprecatedEventShutdownPath
vendor/bin/phpunit tests/TestCase --filter testDeprecatedPathMergeOptionsExtraWhitelist
```

I think these wrong errors will fix , when we fix correct methods/testCases

markstory:
> I think these wrong errors will fix , when we fix correct methods/testCases

We could remove the usage of `deprecated()` in those tests as they fail when run in bulk.

othercorey:
> We could remove the usage of `deprecated()` in those tests as they fail when run in bulk.

I think that is hiding some other issue. My guess is the deprecated flag is removed by some other test.

tests/TestCase/TestSuite/TestCaseTest.php
```diff
@@ -146,6 +146,48 @@ public function testWithErrorReportingWithException(): void
...
+
+    /**
+     * test deprecated with assert after trigger warning
+     */
+    public function testDeprecatedWithAssertAfterTriggerWarning(): void
+    {
+        try {
+            $this->deprecated(function () {
+                trigger_error('deprecation message', E_USER_DEPRECATED);
+                $this->assertTrue(false, 'A random message');
```
saeideng, line 167: this testCase tries to show `deprecated()` behave unlike the phpunit's `$this->expectDeprecation();`

saeideng:
> I think that is hiding some other issue. My guess is the deprecated flag is removed by some other test.

it was because of these lines
https://github.com/cakephp/cakephp/blob/219de3bab07ff983662b9f75d9d32fdafaa6d9aa/src/Core/functions.php#L312-L317

tests/TestCase/Validation/ValidatorTest.php
```diff
@@ -1300,28 +1300,6 @@ public function testAllowEmptyDateTimeCondition(): void
...
 
-    /**
-     * test allowEmptyDateTime with deprecated argument order
-     */
-    public function testAllowEmptyDateTimeDeprecated(): void
-    {
-        $validator = new Validator();
-        $this->deprecated(function () use ($validator): void {
-            $validator->allowEmptyDateTime('published', 'datetime required', 'update');
-        });
```
othercorey, line 1311: Is this allowed now?

saeideng: see above testCase , it only is duplicated TestCase but without `deprecated()` right now
not sure what was this testCase before

`$when` as second Argument allowed (with deprecation) in cake 3 (not sure after which version)
but now in cake 4 `$when` as second Argument is not allowed anymore

tests/TestCase/Core/FunctionsTest.php
```diff
@@ -91,6 +91,30 @@ public function testDeprecationWarningEnabled(): void
         });
     }
 
+    /**
+     * Test duplicate deprecations emit in cli environment and when run PHPUnit
```
markstory: Won't emitting duplicates have the drawback of having more noisy output in tests? My thinking was that you might have a method that is called 50+ times and that you don't need a deprecation warning for each time it is called.

tests/TestCase/TestSuite/TestCaseTest.php
```diff
@@ -146,6 +146,48 @@ public function testWithErrorReportingWithException(): void
         }
     }
 
+    /**
+     * test deprecated
+     */
+    public function testDeprecated(): void
+    {
+        $this->deprecated(function () {
+            trigger_error('deprecation message', E_USER_DEPRECATED);
```
markstory, line 155: You could make different messages to get around the duplicate prevention 🤷

tests/TestCase/Core/FunctionsTest.php
saeideng: when we have several testCases each those need to have /catch deprecation warning independently
I had not good Idea to both , catching deprecation with `deprecated()` + not annoying developer

if there is better solution I can revert last commit

tests/TestCase/TestSuite/TestCaseTest.php
saeideng: this commit was before my last commit

saeideng: and we just need a deprecation warning not complex `deprecationwarning()`, and this helps us to maintaining easier
if needed I can put a  `deprecationwarning()`

tests/TestCase/Core/FunctionsTest.php
saeideng: @markstory  there is a way

having other condition in `deprecationWarning()` method
finding deprecated() method in `$trace` and showing all deprecations, otherwise ignoring duplicated warning

I think it is available
if you agree I can do it in next PR

markstory: > finding deprecated() method in $trace and showing all deprecations, otherwise ignoring duplicated warning

I think that is a reasonable solution. Would we only crawl the stack if the message was identified as a duplicate? Walking the stack could be expensive and create measurable overhead in application code at the beginning of their upgrade process.

tests/TestCase/TestSuite/TestCaseTest.php
markstory: I made this suggestion when I thought the removal of the duplicate detection was because it made writing these tests easier. However, that was before I knew about the additional use-cases you outlined in https://github.com/cakephp/cakephp/pull/15935#discussion_r716800070. I no longer think anything needs to change from the current diff.

tests/TestCase/Core/FunctionsTest.php
saeideng: > Would we only crawl the stack if the message was identified as a duplicate?

yes , right now in this PR is like the

```
        if (
                isset($errors[$checksum]) &&
                (
                    // show all deprecations when phpunit runs
                    //conditions
                )
        )
```

saeideng:
about remains failures
```
    public function test1(): void
    {
            $obj->method();
        });
    }

    public function test2(): void
    {
           $this->expectDeprecation();
           $this->expectDeprecationMessage('....');
           $obj->method();
    }
```

now we story checksum `$errors[$checksum] = true;` in first testCase `test1`
both use same method ` $obj->method();` so there are duplicated message 
and `test2` expect to see deprecation


because we only allow to triggering  duplicated warning when using the `deprecated`
and `test2` do not use it , it fails 
```
Failed asserting that exception of type "PHPUnit\Framework\Error\Deprecated" is thrown.
```

src/Core/functions.php
```diff
@@ -308,15 +310,35 @@ function deprecationWarning(string $message, int $stackFrame = 1): void
...
+                    $value['function'] !== 'deprecated'
+                ) {
+                    continue;
+                }
+
+                return true;
+            }
+
+            return false;
+        };
```
markstory: This will work fine, but should we be doing something 'dumber' and use a `Configure` value, or global variable? `TestCase::deprecated()` could set the global state and `deprecationWarning()` could check if it's set. Using a configure variable of giving users an easy way to have all their deprecations display if they wanted that.

saeideng: OK, 
I can remove some current code,
and then adding a static variable to `TestCase` , maybe helps

```
TestCase::showDuplicatedDeprecation = false
```
we can access it from `deprecated()` and `deprecationWarning()`

we can add enabler method in future , depended to feedback
user only needs to enable it when they want to check with `expectDeprecation` , when there are 2 times

also we can keep the `$errors['tests']` to not check the `showDuplicatedDeprecation` in production
let me to try it

saeideng: @markstory I just added global config, and removed the complexity

>Using a configure variable of giving users an easy way to have all their deprecations display if they wanted that.

yes I added `Configure` , user could enable it to show all duplicated deprecations

src/Core/functions.php
```diff
@@ -311,10 +311,13 @@ function deprecationWarning(string $message, int $stackFrame = 1): void
...
         $checksum = md5($message);
-        if (isset($errors[$checksum])) {
+        $duplicate = (bool)Configure::read('Error.enableDuplicatedDeprecations', false);
+        if (isset($errors[$checksum]) && !$duplicate) {
             return;
         }
-        $errors[$checksum] = true;
+        if (!$duplicate) {
+            $errors[$checksum] = true;
+        }
```
saeideng, line 320: this causes to not keep `checksum` if user want to show all deprecations

so if you use `TestSuite::deprecated()` this prevent to save the `checksum` , this allows us to catch all deprecations by `TestSuite::deprecated()`

saeideng: also this preventing helps to use `$this->expectDeprecation()` after using `TestSuite::deprecated()`

but if user had 2 duplicate deprecation + `$this->expectDeprecation()`

he should enable duplicated deprecation

src/Core/functions.php
saeideng: I did not add `TestCase::showDuplicatedDeprecation` because all people not use cake's testSuite
or if someone want to use cake's repo(s) standalone , or even if they want to show all in web (debug time ) 
having global setting is more useful

src/TestSuite/TestCase.php
```diff
@@ -203,13 +203,37 @@ public function withErrorReporting(int $errorLevel, callable $callable): void
      */
     public function deprecated(callable $callable): void
     {
-        $errorLevel = error_reporting();
-        error_reporting($errorLevel & ~E_USER_DEPRECATED);
+        $duplicate = Configure::read('Error.enableDuplicatedDeprecations');
```
markstory, line 206: You shouldn't need to make this variable or restore the value at the end of the test method. The `tearDown()` of `TestCase` will restore all of `Configure` anyways.

markstory (approved):
Looks good to me.

markstory (approved):
Looks good to me.

src/TestSuite/TestCase.php
saeideng: not in this case
1)
```php

    public function testMethod(): void
{
           $this->expectDeprecation();
           $this->expectDeprecationMessage('after deprecated')
                     trigger_error('message within deprecated method', E_USER_DEPRECATED);
                     trigger_error('other message', E_USER_DEPRECATED);
            });

            trigger_error('after deprecated', E_USER_DEPRECATED);
}
```
we want to above testCase passes

```php
    public function testMethod(): void
{
           // section 1
                     // section 2
            });

            // section 3
}
```

by default in `section 1` and `section 3` deprecations will trigger unless if there was duplicate deprecation
(not included warnings from `deprecated()` 

in `section 2` we must trigger all deprecation even duplicated, ( but we should not include them into `checksum` )

`deprecated()`  just works like before this PR,
we dont want to change other sections 

so we have to back to previous Config immediately after finishing the `deprecated()` block


2) by  default we prevent to show duplicated deprecation unless user want to use `deprecated()` 
so we did not change this Config before (in `setUp` method ), so we can not understand what was before to restore it in `tearDown()`

markstory: I did not consider that scenario. Thank you for the great explanation :bow:

PR #15937, opened 2021-09-22
Move php8 only attributes inside docblock.

dereuromark:
Moves those into the attribute "docblock".


src/I18n/Parser/MoFileParser.php
```diff
@@ -30,15 +30,15 @@ class MoFileParser
      * Magic used for validating the format of a MO file as well as
      * detecting if the machine used to create that file was little endian.
      *
-     * @var float
+     * @var int
```
dereuromark, line 33: Or are those actually float values?

ADmad (approved):

Issue #15734, opened 2021-08-05
Set default value for allowMultipleNulls to TRUE


VARGA-Peter:
I am referring to this pull request: https://github.com/cakephp/cakephp/pull/15226

In 4.2 `allowMultipleNulls` has been re-introduced and I think it was a very good decision. Unfortunately I cannot find a proper way(1) how to set it for the whole project to `TRUE` without bothering with rules which aren't - **in this case** - the best solution - or do you have a very simple solution without functions and dozen lines of code?

It should be quite very simple to set `allowMultipleNulls` to `TRUE` for the whole project in one code of line. I think about something like this:

It could be pre-set in the `Datasources` section in `app_local.php` for the whole project. If it isn't set there so it defaults to `FALSE`. This wouldn't break any compatibility for any project but it would be perfect if multiple `NULL` values are valid in a unique index in the underlying database.

(1) For the moment I do `allowMultipleNulls => TRUE`  in line 39 in `cakephp/cakephp/src/ORM/Rule/IsUnique.php`  - I know, this is very stupid and equals to death penalty but for the moment - hoping you implement something fast - I have to live with it.

kevinquinnyo:
One option would be to create your own `RulesChecker` class and override the `isUnique()` method like this:

```php
namespace App\Model;

class RulesChecker extends \Cake\ORM\RulesChecker
{
    public function isUnique(array $fields, $message = null): RuleInvoker
    {
        $options = is_array($message) ? $message : ['message' => $message];
        $options += ['allowMultipleNulls' => true];

        return parent::isUnique($fields, $options);
    }    
}
```
Then your own `AppTable` class and override the `RULES_CHECKER` const like this:

```php
namespace App\Model\Table;

class AppTable extends \Cake\ORM\Table;
{
    public const RULES_CHECKER = \App\Model\RulesChecker::class;
}
```

Then of course you do have to change all of your tables to extend your `AppTable` instead of cakephp's `Table` class...

Maybe someone more knowledgeable with the 4.x framework can chime in with a better solution than this as it is a bit heavy handed.

VARGA-Peter:
@kevinquinnyo
Your code in comparison to my 1 line which works globally. I have only to read my internal to do list when I run composer once a month.

Exactly this is what I meant in my opening post with "this is not the best solution" .

kevinquinnyo:
> @kevinquinnyo
[...]

PR #15939, opened 2021-09-23
cake 5: Set allowMultipleNulls default to true for IsUnique

othercorey:
Closes https://github.com/cakephp/cakephp/issues/15734

This originally defaulted to true in 3.x. The option was dropped in 4.0 for some reason and re-introduced in 4.2. 

This sets the default back to true to mimic SQL behavior.


othercorey:
This is a pretty nasty change over time, but I think this is the correct default. SQL Server does not allow multiple nulls in a unique index, but you can create an index that supports it.

I suspect users have had to set their config manually already by this point.

PR #15943, opened 2021-09-23
cake 5: Add version to deprecationWarning()

othercorey:
This should help document the deprecation. Hopefully, users can find the migration notes for the changes easier if they're upgrading multiple minor versions.

This follows how symfony handles deprecation notices if consistency helps. They include the package name, but not sure we need that.

### DIFF
--- a/src/Core/functions.php
+++ b/src/Core/functions.php
@@ -272,12 +272,13 @@ if (!function_exists('deprecationWarning')) {
     /**
      * Helper method for outputting deprecation warnings
      *
+     * @param string $version The version that added this deprecation warning.
      * @param string $message The message to output as a deprecation warning.
      * @param int $stackFrame The stack frame to include in the error. Defaults to 1
      *   as that should point to application/plugin code.
      * @return void
      */
-    function deprecationWarning(string $message, int $stackFrame = 1): void
+    function deprecationWarning(string $version, string $message, int $stackFrame = 1): void
     {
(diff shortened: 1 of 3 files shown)


src/Core/functions.php
```diff
@@ -272,12 +272,13 @@ function triggerWarning(string $message): void
...
      * Helper method for outputting deprecation warnings
      *
+     * @param string $version The version that added this deprecation warning.
      * @param string $message The message to output as a deprecation warning.
      * @param int $stackFrame The stack frame to include in the error. Defaults to 1
      *   as that should point to application/plugin code.
      * @return void
      */
-    function deprecationWarning(string $message, int $stackFrame = 1): void
+    function deprecationWarning(string $version, string $message, int $stackFrame = 1): void
```
markstory, line 281: Putting the version first is going to make 4.next merging harder. Should we have the parameters be `message, version`?

othercorey: If you think that will help. The merge will need a parameter added either way.

markstory: True. Lets leave it this way then. The deprecation messages will be obviously wrong 😛

dereuromark: You can also add an assert() for the x.y.z format :) If you want to make double sure.

othercorey: Well, it will throw an error on a missing parameter or a TypeError because an int was passed in.

PR #15944, opened 2021-09-23
Non numeric array indexes are always coerced to strings.

ADmad:

### DIFF
--- a/src/ORM/Marshaller.php
+++ b/src/ORM/Marshaller.php
@@ -675,10 +675,8 @@ class Marshaller
             })
             ->toArray();
 
-        /** @psalm-suppress InvalidArrayOffset */
-        $new = $indexed[null] ?? [];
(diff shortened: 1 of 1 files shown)


othercorey:
Looks good.

src/ORM/Marshaller.php
```diff
@@ -675,10 +675,8 @@ public function mergeMany(iterable $entities, array $data, array $options = []):
             })
             ->toArray();
 
-        /** @psalm-suppress InvalidArrayOffset */
-        $new = $indexed[null] ?? [];
-        /** @psalm-suppress InvalidArrayOffset */
-        unset($indexed[null]);
+        $new = $indexed[''] ?? [];
+        unset($indexed['']);
```
othercorey, line 679: Strange this used null index when the above groupBy() clearly uses `''`.

PR #15947, opened 2021-09-23
Add convenience method LocatorAwareTrait::getTable().

ADmad:
This will allow getting rid of `ModelAwareTrait` for following reasons:

- Relying on `Controller::__get()` and `ModelAwareTrait::loadModel()` is not IDE friendly. It requires additional annotation to indicate that the controller property and value returned by `loadModel()` is of type `Table`.
- Setting of dynamic properties is likely to be deprecated in PHP 8.2 https://wiki.php.net/rfc/deprecate_dynamic_properties.
- `$this->getTable()/$this->getTable('Foo')` is clean and convenient and doesn't involve any hidden side effect. It will allow getting rid of `TableRegistry` too finally.

I will finish the PR once there's a consensus.


src/Datasource/ModelAwareTrait.php
```diff
@@ -71,6 +71,10 @@ protected function _setModelClass(string $name): void
         if ($this->modelClass === null) {
             $this->modelClass = $name;
         }
+
+        if ($this->defaultTable === null) {
```
markstory: Should the trait define this property as '' or null to avoid the typing errors?

src/ORM/Locator/LocatorAwareTrait.php
```diff
@@ -58,4 +62,14 @@ public function getTableLocator(): LocatorInterface
         /** @var \Cake\ORM\Locator\LocatorInterface */
         return $this->_tableLocator;
     }
+
+    public function getTable(?string $alias = null, array $options = []): Table
```
markstory: I'm a fan of this method, nice work.

src/Datasource/ModelAwareTrait.php
ADmad: The `LocatorAwareTrait` does have this property. This is a very rough POC. The error will go away once I finish the PR :)

othercorey:
I should request features more often!

src/ORM/Locator/LocatorAwareTrait.php
```diff
@@ -58,4 +67,40 @@ public function getTableLocator(): LocatorInterface
...
+     * Set the default table.
+     *
+     * If the default table is already set it will not be overwritten.
+     *
+     * @param string $name Class name.
+     * @return void
+     */
+    protected function setDefaultTable(string $name): void
+    {
+        if ($this->defaultTable === null) {
```
othercorey: I assume this is for controllers to mimic the default table matching controller name? If so, should we assert that `$defaultTable` is null since users shouldn't be re-configuring.

ADmad: Don't really need this method, I got rid of it.

dereuromark:
> It requires additional annotation to indicate that the controller property and value returned by loadModel() is of type Table.

Sure, with the IdeHelper run this was quite easy to add though, within seconds.

> $this->getTable()/$this->getTable('Foo') is clean and convenient and doesn't involve any hidden side effect.

I do like the method. 
I don't want to be the party pooper, but arent static analyzers gonna have trouble with this new method?
Also some IDEs might. They all understood the annotation (or to be honest: basic property for it added, as that also always worked the same way), but mainly PHPStorm using the meta file can also understand this dynamic resolution.
Not sure about the others. E.g.: A specific Table method then called.

We might want to check on that before we remove any of the other methods that so far exist.

src/Controller/Controller.php
```diff
@@ -282,30 +279,20 @@ public function loadComponent(string $name, array $config = []): Component
...
-        }
-
         $trace = debug_backtrace();
         $parts = explode('\\', static::class);
         trigger_error(
             sprintf(
-                'Undefined property: %s::$%s in %s on line %s',
+                'Undefined property: %s::$%s in %s on line %s.'
+                    . 'Auto loading of default model as property has been removed.'
+                    . ' Use $this->getTable() instead.',
```
othercorey: This message is a little awkward. Maybe `The default model is no longer auto-loaded and assigned to a property.`

ADmad:
@dereuromark The method has return type `Table` so static analyzers / IDEs would be unhappy only if you are accessing a custom table method. In which case one can first assign the value to a variable and annotate it. In general if one finds the needs to add custom methods to the table then they should probably be using services classes instead.

With phpstorm's meta file you can indicate the `getTable('Foo')` returns a value of type `App\Model\Foo` isn't it? So your ide helper plugin can just generate those.

We will have to remove the property assignment if the RFC I linked passes.

ADmad:
We can retain the use of `$this->Foo` for default tables in controllers by making `Controller::_get()` return the value of `getTable()`. It doesn't need to actually set a property.

Edit: Done

dereuromark:
> only if you are accessing a custom table method

But that's kind of the normal case as well, lots of custom methods that are usually called.
So in reality you have to often annotate then every single line, multiple times per file instead of the single one at the top.

I would probably wait if that RFC even passes or not.
If it does, we could still even just add - tooling based - the `public $FooBar` property and then this should also be valid.

Yes, the IDE (at least PHPStorm) will have no issue understanding it, but the tools like PHPStan/Psalm might still.
So here we would need then those inline annotations always, as well.

It is nice to have a convenience method, but maybe we can also find a way to use some factory that can have the necessary stuff autogenerated or annotated in a way that we can further improve the usability?

ADmad:
I have retained the ability to use `$this->Foos` for which you can add `@property` annotation as currently done, doesn't that resolve your concerns? This allows keeping the currently baked controllers exactly the same.

dereuromark:
> I have retained the ability to use $this->Foos

For me it is not so much about the default model as also the custom ones loaded inside controller and other classes using the loadModel() call.
But as long as such a way is still present for now, I dont see any concern with this moving forward.

ADmad:
`$this->loadModel('Foos'); $this->Foos->something()` is awful usage which needs to die.

One can instead have a lot more readable usage which even someone new to Cake can easily understand:
`$this->Foos = $this->getTable('Foos'); $this->Foos->something();`

dereuromark:
Assigning the property at runtime manually seems not useful either though
Then you can just stick to the local instance var IMO.

I guess it all depends in the end on the RFC result. As that determines if the current convenience way of single annotation can stay or not. 

But as I said: That still doesnt mean we cannot move forward with this method, which indeed can replace TableRegistry calls then.

markstory:
> We can retain the use of $this->Foo for default tables in controllers by making Controller::_get() return the value of getTable(). It doesn't need to actually set a property.

We currently use `__get()` to do this for the default model. We currently also use dynamic property assignment for components as well. How would the magic property cascade work? I assume there will be an unfortunate soul who has overlapping model and component names.

ADmad:
> We currently also use dynamic property assignment for components as well. How would the magic property cascade work? 

I'll make a follow up PR to stop setting dynamic properties for components too and update `Controller::_get()` to return component instances. If property names matches the default table alias then that will be returned else it will check for component matching that name.

> I assume there will be an unfortunate soul who has overlapping model and component names.

Overlapping names would create problems currently too. Since `loadComponent()` is called from `initialize()` the property would be set with the component instance, so one wouldn't be able to use the same name to access the default model/table. If someone later uses `loadModel()` with the alias specified then it would overwrite the property which the model instance, which is wack.

So this change will avoid such issues.

ADmad:
This is ready for final review/merge. I have dropped `ModelAwareTrait` and also opened docs PR.

src/Controller/Controller.php
```diff
@@ -282,30 +280,30 @@ public function loadComponent(string $name, array $config = []): Component
...
+            if (!str_contains($this->defaultTable, '\\')) {
+                [, $class] = pluginSplit($this->defaultTable, true);
             } else {
-                $class = App::shortName($this->modelClass, 'Model/Table', 'Table');
+                $class = App::shortName($this->defaultTable, 'Model/Table', 'Table');
             }
 
             if ($class === $name) {
-                return $this->loadModel();
+                return $this->getTable();
```
markstory, line 298: This is something we should make a rector change for. This will be in lots of applications.

src/Datasource/ModelAwareTrait.php
```diff
@@ -1,167 +0,0 @@
...
-     *
-     * @param string|null $modelClass Name of model class to load. Defaults to $this->modelClass.
-     *  The name can be an alias like `'Post'` or FQCN like `App\Model\Table\PostsTable::class`.
-     * @param string|null $modelType The type of repository to load. Defaults to the getModelType() value.
-     * @return \Cake\Datasource\RepositoryInterface The model instance created.
-     * @throws \Cake\Datasource\Exception\MissingModelException If the model class cannot be found.
-     * @throws \UnexpectedValueException If $modelClass argument is not provided
-     *   and ModelAwareTrait::$modelClass property value is empty.
-     */
-    public function loadModel(?string $modelClass = null, ?string $modelType = null): RepositoryInterface
```
markstory, line 95: We will need to add a deprecation for this in 4.3

tests/test_app/TestApp/Controller/ArticlesController.php
```diff
@@ -23,5 +23,5 @@ class ArticlesController extends Controller
     /**
      * @var string
      */
-    protected ?string $modelClass = ArticlesTable::class;
+    protected ?string $defaultTable = ArticlesTable::class;
```
markstory, line 26: I bet there are plenty of applications that are using modelClass. We could emit a deprecation for this during Controller::__construct().

src/Datasource/ModelAwareTrait.php
ADmad: That would require backporting this change to 4.3 which I was hoping to avoid :slightly_smiling_face:. Given how commonly `loadModel()` and `$modelClass` would be used in apps and plugins, the users could be quite annoyed by it's deprecation. People were quite unamused when we started showing deprecation errors for `TableRegistry`'s methods in a minor release.

So I was hoping we could do a clean switch in 5.0 directly and provide rector rules to replace use of `loadModel()` with `getTable()` and `$modelClass` with `$defaultTable`.

markstory: I'm cautious of including a hard break with no deprecation as we can't say 'fix all your deprecations and you are pretty much done' anymore. Could we include `getTable()` in 4.x and update the documentation? Perhaps we deprecate `loadModel()` in a future 4.x release which doesn't have as many deprecations as 4.3 will contain.

ADmad: Sure, we can do. I'll make a PR to add `getTable()` to 4.3.

PR #15948, opened 2021-09-23
Improve code flow as per bail early principle.

dereuromark:
This makes the code on these methods a bit more consistent with the rest of the classes.
With the bail early principle on the falsey cases (guard clauses) to exit early, and keeping the happy case as final return value where possible.

Also removes some nesting and makes it a bit more human readable in some cases.

### DIFF
--- a/src/Database/Type/IntegerType.php
+++ b/src/Database/Type/IntegerType.php
@@ -119,10 +119,10 @@ class IntegerType extends BaseType implements BatchCastingInterface
         if ($value === null || $value === '') {
             return null;
         }
-        if (is_numeric($value)) {
-            return (int)$value;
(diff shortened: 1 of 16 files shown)


src/Database/Type/IntegerType.php
```diff
@@ -119,10 +119,10 @@ public function marshal(mixed $value): ?int
         if ($value === null || $value === '') {
             return null;
         }
-        if (is_numeric($value)) {
-            return (int)$value;
+        if (!is_numeric($value)) {
```
othercorey: This could be folded into the if statement above it now?

othercorey:
Looks good to me.

PR #15949, opened 2021-09-23
Clarify assoc behavior and merging.

dereuromark:
Makes this consistent with
- setOptions() from RuleInvoker
- assoc behavior of such string keyed option arrays

```
There were 2 failures:

1) Cake\Test\TestCase\View\ViewBuilderTest::testArrayPropertyMerge with data set #0 ('options', array('value'))
Should merge
Failed asserting that two arrays are identical.
--- Expected
+++ Actual
@@ @@
 Array &0 (
+    0 => 'Merged'
     'key' => 'value'
-    0 => 'Merged'
 )

/home/runner/work/cakephp/cakephp/tests/TestCase/View/ViewBuilderTest.php:184
/home/runner/work/cakephp/cakephp/src/TestSuite/TestCase.php:182
/home/runner/work/cakephp/cakephp/tests/TestCase/View/ViewBuilderTest.php:188

2) Cake\Test\TestCase\View\ViewBuilderTest::testSetOptionsMultiple
--- Expected
+++ Actual
@@ @@
 Array (
-    0 => 'oldOption'
-    1 => 'newOption'
-    2 => 'anotherOption'
+    0 => 'newOption'
+    1 => 'anotherOption'
 )

/home/runner/work/cakephp/cakephp/tests/TestCase/View/ViewBuilderTest.php:362
```

Do we want to support this behavior in 5.x?
Or shall we just fix up the test cases to match the general expectation of assoc here?

I don't see how the test case demos here show any usefulness of the array_merge() as it was.
As you cannot properly reference the options anyway without a clear key to access them.

### DIFF
--- a/src/Database/Schema/TableSchema.php
+++ b/src/Database/Schema/TableSchema.php
@@ -290,7 +290,7 @@ class TableSchema implements TableSchemaInterface, SqlGeneratorInterface
      * Constructor.
      *
      * @param string $table The table name.
-     * @param array $columns The list of columns for the schema.
+     * @param array<string, array> $columns The list of columns for the schema.
      */
     public function __construct(string $table, array $columns = [])
     {
@@ -674,7 +674,7 @@ class TableSchema implements TableSchemaInterface, SqlGeneratorInterface
      */
     public function setOptions(array $options)
     {
-        $this->_options = array_merge($this->_options, $options);
+        $this->_options = $options + $this->_options;
 
         return $this;
     }
--- a/src/Datasource/SchemaInterface.php
+++ b/src/Datasource/SchemaInterface.php
@@ -53,7 +53,7 @@ interface SchemaInterface
      * - `comment` The comment for the column.
      *
      * @param string $name The name of the column
-     * @param array|string $attrs The attributes for the column or the type name.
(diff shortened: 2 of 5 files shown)


markstory:
The failure in `Cake\Test\TestCase\View\ViewBuilderTest::testSetOptionsMultiple` seems like something that could impact some applications. What benefits are there to introducing that break?

markstory:
The failure in `Cake\Test\TestCase\View\ViewBuilderTest::testSetOptionsMultiple` seems like something that could impact some applications. What benefits are there to introducing that break?

ADmad:
In practice I don't see `setOptions()` being used with arrays with numeric keys. It almost always will be associative arrays. It's complementary method `setOption()` already requires a string $name argument which is used as array key.

Anyone relying on specific ordering of keys for an associative array would be setting themselves up for failure. So I think this change is fine and the test should be updated.

markstory:
> Anyone relying on specific ordering of keys for an associative array would be setting themselves up for failure. So I think this change is fine and the test should be updated.

:100: agree. The ordering matters even less because the main interface for configuration

ADmad:
Can you rebase this onto latest `5.x` please to ensure the static analyzers remain happy.

dereuromark:
Looks all good

PR #15952, opened 2021-09-24
Make CSP nonces opt-in

markstory:
Unfortunately the CSPBuilder library doesn't allow us to read if the policy contains nonce directives. Automatically adding nonces can break applications that use CSP. Making the nonces opt-in allows developers to enable stricter CSP rules when they are ready.


othercorey:
The config code look ok.

PR #15953, opened 2021-09-24
Return empty string when no data for session id

othercorey:
Revert the read failure return type from https://github.com/cakephp/cakephp/pull/15578.

This is causing session_open() to fail when there is no data. Seems to be a documentation problem in php.

https://github.com/php/doc-en/issues/970

### DIFF
--- a/src/Http/Session/DatabaseSession.php
+++ b/src/Http/Session/DatabaseSession.php
@@ -122,14 +122,20 @@ class DatabaseSession implements SessionHandlerInterface
             ->first();
 
         if (empty($result)) {
-            return false;
+            return '';
         }
 
         if (is_string($result['data'])) {
             return $result['data'];
         }
 
-        return stream_get_contents($result['data']);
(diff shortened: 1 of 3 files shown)


src/Http/Session/DatabaseSession.php
```diff
@@ -122,14 +122,20 @@ public function read($id)
...
 
         if (is_string($result['data'])) {
             return $result['data'];
         }
 
-        return stream_get_contents($result['data']);
+        $session = stream_get_contents($result['data']);
+
+        if ($session === false) {
+            return '';
```
othercorey, line 135: This case might need to return false since it's an actual read failure.

markstory: Lets keep the `''`. Failing to start the session won't help the end user. This scenario looks like we found the session record but its data is corrupted, in that scenario we should probably start a new session as blowing up the request won't help the end user move forward at all.

PR #15955, opened 2021-09-24
Allow association binding key to be null during cascadeDelete

### LINKED ISSUE #15334: Missing IS operator in Cake\ORM\Association\DependentDeleteHelper::cascadeDelete()
* [x] bug

* CakePHP Version: 4.2.1

### What you did
Delete Entity with HasOne Association with deleteCascade and null value (The bindingKey and foreignKey are switched between tables).

### What happened
                sprintf('Expression `%s` is missing operator (IS, IS NOT) with `null` value.', $expression)
            );

I was expected the conditions automatic handle the IS operator for null values.
Also would be fine to merge the association conditions into delete conditions.

### My Fix (Yes, I know, it is quite improvable...)

```
<?php
declare(strict_types=1);

/**
 * CakePHP(tm) : Rapid Development Framework (http://cakephp.org)
 * Copyright (c) Cake Software Foundation, Inc. (http://cakefoundation.org)
 *
 * Licensed under The MIT License
 *
 * @copyright     Copyright (c) Cake Software Foundation, Inc. (http://cakefoundation.org)
 * @link          http://cakephp.org CakePHP(tm) Project
 * @since         3.5.0
 * @license       http://www.opensource.org/licenses/mit-license.php MIT License
 */
namespace Cake\ORM\Association;

use Cake\ORM\Association;

/**
 * Helper class for cascading deletes in associations.
 *
 * @internal
 */
class DependentDeleteHelper
{
    /**
     * Cascade a delete to remove dependent records.
     *
     * This method does nothing if the association is not dependent.
     *
     * @param \Cake\ORM\Association $association The association callbacks are being cascaded on.
     * @param \Cake\Datasource\EntityInterface $entity The entity that started the cascaded delete.
     * @param array $options The options for the original delete.
     * @return bool Success.
     */
    public function cascadeDelete(Association $association, EntityInterface $entity, array $options = []): bool
    {
        if (!$association->getDependent()) {
            return true;
        }
        $table = $association->getTarget();
        /** @psalm-suppress InvalidArgument */
        $foreignKey = array_map([$association, 'aliasField'], (array)$association->getForeignKey());
        $bindingKey = (array)$association->getBindingKey();
        $conditions = array_combine($foreignKey, $entity->extract($bindingKey));

        //////////////////////////
        //  START FIX IS NULL   //
        //////////////////////////
        foreach($conditions as $key => $value){
            if (is_null($value)) {
                $conditions["$key IS"] = $value;
                unset($conditions[$key]);
            }
        }
        //////////////////////
        //  END FIX IS NULL //
        //////////////////////
        
        if ($association->getCascadeCallbacks()) {
            foreach ($association->find()->where($conditions)->all()->toList() as $related) {
                $success = $table->delete($related, $options);
                if (!$success) {
                    return false;
                }
            }

            return true;
        }

        $association->deleteAll($conditions);

        return true;
    }
}
```

othercorey:
Fixes https://github.com/cakephp/cakephp/issues/15334

This can occur for non-primary key bindings in the source table.

### DIFF
--- a/src/ORM/Association/DependentDeleteHelper.php
+++ b/src/ORM/Association/DependentDeleteHelper.php
@@ -45,7 +45,11 @@ class DependentDeleteHelper
         /** @psalm-suppress InvalidArgument */
         $foreignKey = array_map([$association, 'aliasField'], (array)$association->getForeignKey());
         $bindingKey = (array)$association->getBindingKey();
-        $conditions = array_combine($foreignKey, $entity->extract($bindingKey));
+        $bindingValue = $entity->extract($bindingKey);
+        if (in_array(null, $bindingValue, true)) {
+            return true;
+        }
+        $conditions = array_combine($foreignKey, $bindingValue);
 
         if ($association->getCascadeCallbacks()) {
             foreach ($association->find()->where($conditions)->all()->toList() as $related) {
--- a/tests/Fixture/NullableAuthorsFixture.php
+++ b/tests/Fixture/NullableAuthorsFixture.php
@@ -1,0 +1,41 @@
+<?php
+/**
+ * CakePHP(tm) : Rapid Development Framework (https://cakephp.org)
+ * Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
+ *
+ * Licensed under The MIT License
+ * For full copyright and license information, please see the LICENSE.txt
+ * Redistributions of files must retain the above copyright notice
+ *
+ * @copyright     Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
+ * @link          https://cakephp.org CakePHP(tm) Project
+ * @since         4.2.10
+ * @license       https://opensource.org/licenses/mit-license.php MIT License
+ */
+namespace Cake\Test\Fixture;
+
+use Cake\TestSuite\Fixture\TestFixture;
+
+class NullableAuthorsFixture extends TestFixture
+{
+    /**
+     * fields property
+     *
+     * @var array
+     */
+    public $fields = [
+        'id' => ['type' => 'integer'],
+        'author_id' => ['type' => 'integer', 'null' => true],
+        '_constraints' => ['primary' => ['type' => 'primary', 'columns' => ['id']]],
+    ];
+
+    /**
+     * records property
+     *
+     * @var array
+     */
+    public $records = [
+        ['author_id' => 3],
+        ['author_id' => null],
+    ];
+}
--- a/tests/TestCase/ORM/Association/HasOneTest.php
+++ b/tests/TestCase/ORM/Association/HasOneTest.php
@@ -33,7 +33,7 @@ class HasOneTest extends TestCase
      *
      * @var array
      */
-    protected $fixtures = ['core.Users', 'core.Profiles'];
+    protected $fixtures = ['core.Articles', 'core.Authors', 'core.NullableAuthors', 'core.Users', 'core.Profiles'];
 
     /**
      * @var bool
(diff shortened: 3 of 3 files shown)


src/ORM/Association/DependentDeleteHelper.php
```diff
@@ -45,7 +45,12 @@ public function cascadeDelete(Association $association, EntityInterface $entity,
         /** @psalm-suppress InvalidArgument */
         $foreignKey = array_map([$association, 'aliasField'], (array)$association->getForeignKey());
         $bindingKey = (array)$association->getBindingKey();
-        $conditions = array_combine($foreignKey, $entity->extract($bindingKey));
+        $conditions = array_combine(
+            array_map(function ($field) {
+                return $field . ' IS';
+            }, $foreignKey),
+            $entity->extract($bindingKey)
+        );
```
othercorey: Should this bail out instead of running the query? This would delete all target table rows with foreign key null which might not be desired either.

src/ORM/Association/DependentDeleteHelper.php
```diff
@@ -45,7 +45,11 @@ public function cascadeDelete(Association $association, EntityInterface $entity,
         /** @psalm-suppress InvalidArgument */
         $foreignKey = array_map([$association, 'aliasField'], (array)$association->getForeignKey());
         $bindingKey = (array)$association->getBindingKey();
-        $conditions = array_combine($foreignKey, $entity->extract($bindingKey));
+        $bindingValue = $entity->extract($bindingKey);
+        if (in_array(null, $bindingValue, true)) {
+            return true;
+        }
```
othercorey, line 51: Should this be if "any" or if "all" are null?

markstory: Should this be an error? The delete cannot be safely cascaded and the intent is to have that done.

othercorey: Well, there is nothing to delete though. If you have an association based on a different field than the entity id, that could be null when deleting the entity.

markstory: Can we delete the rows where the binding field is null then? There could be other uses for rows with non-null values, and the provided entity did have a null so an explicit match would be congruent with how we treat all the other values.

othercorey: I had that originally. You could end up deleting all rows in the target table where foreign key is null even though there was nothing attached to this entity. I wasn't sure which of these solutions was correct.

markstory: I think we need to pick a side and either treat the null as we would all other values or raise an error as the operation is ambiguous and likely a mistake on the developer's behalf. Treating it as an error state is 'safer'. The intent of the current code is to do the delete but because of incorrect logic those deletes aren't happening.

othercorey: My guess is this is a bug and users don't use the feature enough to notice. I can imagine having an entity where the hasOne relationship is empty.

If an association is valid with a null foreign key, then we should delete nulls. If it isn't, and the property on the entity is left empty, then we should ignore it.

markstory: Hrm. The more I think about it I'm wrong and what you have is better. Like you said an optional association will have a null foreign key. This shouldn't cause an error and we can skip attempting the delete as a primary key of null is nonsense.

othercorey: Extended the test to show that a null binding key and null foreign key do not create an association so the delete should not cascade.

PR #15958, opened 2021-09-25
Fix IDE found issues/smells.

dereuromark:
The PHPStorm IDE gets quite powerful in smell detection topics

- `@inheritDoc` used for member without parent members with doc comments 
- null missing on docblock param/return
- missing return values for non void
- return value of void used
- unused local var
- resolve mixed to concrete where possible
- sprintf() without args issue
- duplication removal of if/loops
- duplication in regex removal
- passing non existing args to functions/methods

An actual bug/no-op has also been found and resolved

### DIFF
--- a/src/Command/ServerCommand.php
+++ b/src/Command/ServerCommand.php
@@ -143,7 +143,7 @@ class ServerCommand extends Command
 
         $port = ':' . $this->_port;
         $io->out(sprintf('built-in server is running in http://%s%s/', $this->_host, $port));
-        $io->out(sprintf('You can exit with <info>`CTRL-C`</info>'));
+        $io->out('You can exit with <info>`CTRL-C`</info>');
         system($command);
 
         return static::CODE_SUCCESS;
--- a/src/Console/ShellDispatcher.php
+++ b/src/Console/ShellDispatcher.php
@@ -279,7 +279,6 @@ class ShellDispatcher
 
             $other = static::alias($shell);
             if ($other) {
-                $other = $aliases[$shell];
                 if ($other !== $plugin) {
                     Log::write(
                         'debug',
--- a/src/Database/Expression/QueryExpression.php
+++ b/src/Database/Expression/QueryExpression.php
@@ -736,7 +736,7 @@ class QueryExpression implements ExpressionInterface, Countable
      * @param string $field The value from which the actual field and operator will
      * be extracted.
      * @param mixed $value The value to be bound to a placeholder for the field
-     * @return \Cake\Database\ExpressionInterface|string
+     * @return \Cake\Database\ExpressionInterface
      * @throws \InvalidArgumentException If operator is invalid or missing on NULL usage.
      */
     protected function _parseCondition(string $field, $value)
--- a/src/Datasource/RulesChecker.php
+++ b/src/Datasource/RulesChecker.php
@@ -307,7 +307,7 @@ class RulesChecker
      * Utility method for decorating any callable so that if it returns false, the correct
      * property in the entity is marked as invalid.
      *
-     * @param callable $rule The rule to decorate
+     * @param callable|\Cake\Datasource\RuleInvoker $rule The rule to decorate
      * @param array|string|null $name The alias for a rule or an array of options
      * @param array<string, mixed> $options The options containing the error message and field.
      * @return \Cake\Datasource\RuleInvoker
--- a/src/Http/ServerRequest.php
+++ b/src/Http/ServerRequest.php
@@ -453,7 +453,7 @@ class ServerRequest implements ServerRequestInterface
      *
      * @param string $name The method called
      * @param array $params Array of parameters for the method call
-     * @return mixed
+     * @return bool
      * @throws \BadMethodCallException when an invalid method is called.
      */
     public function __call(string $name, array $params)
--- a/src/TestSuite/Fixture/FixtureHelper.php
+++ b/src/TestSuite/Fixture/FixtureHelper.php
@@ -79,8 +79,6 @@ class FixtureHelper
             } else {
                 /** @psalm-var class-string<\Cake\Datasource\FixtureInterface> */
                 $className = $fixtureName;
-                /** @psalm-suppress PossiblyFalseArgument */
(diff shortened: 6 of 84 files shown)


src/Console/ShellDispatcher.php
```diff
@@ -279,7 +279,6 @@ public function addShortPluginAliases(): array
 
             $other = static::alias($shell);
             if ($other) {
-                $other = $aliases[$shell];
```
dereuromark, line 282: This was the bug - a no-op as $other is being assigned $plugin value and then compared afterwards.

src/View/View.php
```diff
@@ -1169,7 +1172,7 @@ protected function _evaluate(string $templateFile, array $dataForView): string
         ob_start();
 
         try {
-            include func_get_arg(0);
+            include $templateFile;
```
dereuromark: Was there a reason `func_get_arg(0)` was used?

src/TestSuite/Fixture/FixtureHelper.php
```diff
@@ -80,7 +80,7 @@ public function loadFixtures(array $fixtureNames): array
                 /** @psalm-var class-string<\Cake\Datasource\FixtureInterface> */
                 $className = $fixtureName;
                 /** @psalm-suppress PossiblyFalseArgument */
-                $name = preg_replace('/Fixture\z/', '', substr(strrchr($fixtureName, '\\'), 1));
+                //unused $name = preg_replace('/Fixture\z/', '', substr(strrchr($fixtureName, '\\'), 1));
```
dereuromark: This $name was unused, I wonder if it was meant to be $className assignment instead?

src/Auth/ControllerAuthorize.php
```diff
@@ -92,6 +92,6 @@ public function authorize($user, ServerRequest $request): bool
             ));
         }
 
-        return (bool)$this->_Controller->isAuthorized($user);
+        return (bool)$this->_Controller->isAuthorized();
```
markstory: This looks incorrect. The Controller will need the User to do authorization checks.

src/TestSuite/Fixture/FixtureHelper.php
markstory: Probably a left over from refactoring/cleanup changes. We can remove this.

src/View/View.php
markstory: Yes. The view variables could contain `$templateFile` and overwrite the template we include. This is extra dangerous if a controller action sets a `$templateFile` from user data as it opens an arbitrary file access vulnerability. Pretty sure this was added in the 'pre inline comment' era of CakePHP which is why there is no context with it.

markstory: 98e8aaaf682  is when this was switched to func_get_arg(0)

ADmad: We should add a comment now to clarify why the code is so.

src/Command/ServerCommand.php
```diff
@@ -143,7 +143,7 @@ public function execute(Arguments $args, ConsoleIo $io): ?int
 
         $port = ':' . $this->_port;
         $io->out(sprintf('built-in server is running in http://%s%s/', $this->_host, $port));
-        $io->out(sprintf('You can exit with <info>`CTRL-C`</info>'));
+        $io->out(sprintf('You can exit with <info>`%s`</info>', 'CTRL-C'));
```
ADmad: Better to just remove the `sprintf()`?

src/Auth/ControllerAuthorize.php
dereuromark: I fixed up the test file instead to expect this input.

dereuromark:
What just happened regarding deprecations? All builds now fail after my rebase.

dereuromark:
Aha
https://github.com/sebastianbergmann/phpunit/compare/9.5.9...9.5.10#diff-28ac866bc10f444cd73e52c08c5f35319df9a5fea67923a002838f1c58dc97a0R7
What a breaking change in a patch release...^^

src/Database/Expression/QueryExpression.php
```diff
@@ -736,7 +736,7 @@ protected function _addConditions(array $conditions, array $types): void
      * @param string $field The value from which the actual field and operator will
      * be extracted.
      * @param mixed $value The value to be bound to a placeholder for the field
-     * @return \Cake\Database\ExpressionInterface|string
+     * @return \Cake\Database\ExpressionInterface
```
ADmad, line 739: We'll have to fix the actual return types added in 5.x too.

src/Datasource/RulesChecker.php
```diff
@@ -307,7 +307,7 @@ protected function _checkRules(EntityInterface $entity, array $options = [], arr
      * Utility method for decorating any callable so that if it returns false, the correct
      * property in the entity is marked as invalid.
      *
-     * @param callable $rule The rule to decorate
+     * @param callable|\Cake\Datasource\RuleInvoker $rule The rule to decorate
```
ADmad, line 310: Adding `RuleInvoker` is redundant since it's an invokable class and thus sub type of `callable`.

src/Http/ServerRequest.php
```diff
@@ -453,7 +453,7 @@ public function referer(bool $local = true): ?string
      *
      * @param string $name The method called
      * @param array $params Array of parameters for the method call
-     * @return mixed
+     * @return bool
```
ADmad, line 456: Noting for updating type in 5.x

src/Datasource/RulesChecker.php
dereuromark: True, but here it still is helpful to clarify the concrete type expected as then you also have IDE support for the

    $rule->setOptions($options)->setName($name);
    
part. Also feels a bit more clear in terms of communication of input types than just any "callable".

dereuromark: If people feel strongly about this not being explicit, I can also revert it.

ADmad: I dont mind it as long as no tool complains about redundant type.

PR #15959, opened 2021-09-25
Update dev dependency constraint for PHP 8.1 support.

ADmad:


ADmad:
@markstory The test failures seem to be caused by #15808. It's strange though that the testsuite passed for that PR.

ADmad:
Sorry it's a breaking change in latest phpunit release that's causing it.

PR #15960, opened 2021-09-25
Update Stan tools

dereuromark:


dereuromark:
> Child process error (exit code 137): Killed 

That doesn't look good
Maybe we try one version before?

Issue #15956, opened 2021-09-24
Test strategy enhancements


pabloelcolombiano:
This is a (multiple allowed):

* [] bug
* [x] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 4.3.0-RC1

### What you did
Persisted data in a test that did not have test fixtures. 

### What happened
Since there are no fixtures, the test wrapping transaction is not activated and the test DB is not cleaned.

First congratulation on the new test cycle with "strategies", this is working really well and looks really elegant.

Generally, I am implementing the trigger-based strategy that the [test suite light](https://github.com/vierge-noire/cakephp-test-suite-light/issues/49) will be using, and having a tiny issue with the current implementation of the test cycle. I would propose to make the two following enhancements:

1. In `TestCase::setupFixtures()`, I would comment out the return on empty `$fixtureNames`. Since fixtures are not required for managing the schema, it is well possible to create test fixtures with factories and still want to take advantage of the transaction for the cleanup. If you still want to skip the "transactioning" on empty static test fixtures, it would be handier to move that check in the `TransactionStrategy::setUp`, instead of the `TestCase::setupFixtures()`. 

2. `FixtureStrategyInterface::setupTest()` could take  `TestCase $test` as argument, instead of `array $fixtureNames`. This enables to do much more elaborate work in some custom strategy, and on the TransactionStrategy side, it is then still possible to get the fixtures from the test with `$test->getFixtures()`.

3. If you agree on point 2, we could also then have `FixtureStrategyInterface::teardownTest(TestCase $test)`. 

I am open to propose a PR, just wanted to discuss this with you first.

markstory:
> In TestCase::setupFixtures(), I would comment out the return on empty $fixtureNames. Since fixtures are not required for managing the schema, it is well possible to create test fixtures with factories and still want to take advantage of the transaction for the cleanup. 

We do this early return to avoid the runtime cost of starting/stopping transactions doing truncation/transactions on test methods that aren't using the database.

> FixtureStrategyInterface::setupTest() could take TestCase $test as argument, instead of array $fixtureNames. This enables to do much more elaborate work in some custom strategy, and on the TransactionStrategy side, it is then still possible to get the fixtures from the test with $test->getFixtures().

This seems reasonable to me. I have some concerns that the TestCase interface is really wide and the amount of theoretical compatibility breaks that presents. Perhaps @othercorey can weigh in on this?

othercorey:
[...]

PR #15962, opened 2021-09-25
#15956 Enable hybrid test fixtures strategies.

pabloelcolombiano:

### DIFF
--- a/src/TestSuite/Fixture/TransactionStrategy.php
+++ b/src/TestSuite/Fixture/TransactionStrategy.php
@@ -50,6 +50,10 @@ class TransactionStrategy implements FixtureStrategyInterface
      */
     public function setupTest(array $fixtureNames): void
     {
+        if (empty($fixtureNames)) {
+            return;
+        }
+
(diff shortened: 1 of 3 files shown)


src/TestSuite/Fixture/TransactionStrategy.php
```diff
@@ -50,6 +50,10 @@ public function __construct()
      */
     public function setupTest(array $fixtureNames): void
     {
+        if (empty($fixtureNames)) {
```
markstory, line 53: We could have a similar check in TruncateStrategy as well to avoid some method calls.

pabloelcolombiano: In the actual implementation of the TruncateStrategy, an empty $fixtureNames array will lead to no action performed. Would you like to add a check to cover potential future implementation? I don't mind adding it, but it looks here not necessary.

pabloelcolombiano: This is now done. You are right that we want to prevent future development to oversee that.

PR #15963, opened 2021-09-25
Use phpstan level 1 to check basic tests sanity.

dereuromark:
RFC/WIP

I adapted this quite some time ago and I recommend we also do that for core.

It adds very basic sanity checks that helped me in the past
- Valid PSR file structure and class naming
- Basic PHP validity (e.g. reports now undefined class properties, in light of possible 8.2 changes)

E.g.
```
  Line   Console/CommandRunnerTest.php                                       
 ------ -------------------------------------------------------------------- 
  406    Class Cake\Test\TestCase\Console\stdClass not found.                
         💡 Learn more at https://phpstan.org/user-guide/discovering-symbols  
  408    Class Cake\Test\TestCase\Console\stdClass not found.                
         💡 Learn more at https://phpstan.org/user-guide/discovering-symbols  
```
https://github.com/cakephp/cakephp/pull/15963/checks?check_run_id=3710160043

We can adapt our test classes to that standard, we don't however have to go further than level 1 to keep things simple.
30 seconds worth spending in CI runtime.

What do you think?

Given the amount of work to add all props we could start with a subset for now.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -247,3 +247,7 @@ jobs:
     - name: Run phpstan
       if: success() || failure()
       run: vendor/bin/phpstan.phar analyse --error-format=github
+
+    - name: Run phpstan for tests
+      if: success() || failure()
+      run: vendor/bin/phpstan.phar analyse -c tests/phpstan.neon --error-format=github
--- a/composer.json
+++ b/composer.json
@@ -109,6 +109,7 @@
             "@phpstan",
             "@psalm"
         ],
+        "stan-tests": "phpstan.phar analyze -c tests/phpstan.neon",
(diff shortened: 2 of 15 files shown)


markstory:
2600 errors is going to take a while to fix. Did you want to add this build and have it fail for a while we get around to cleaning up/ignoring the problems?

othercorey:
> 2600 errors is going to take a while to fix. Did you want to add this build and have it fail for a while we get around to cleaning up/ignoring the problems?

We just got the build clean again :(

dereuromark:
We could start with subset, e.g. Core

dereuromark:
I added the main folders for now, so it should pass.

We can do the rest as follow up PRs. 522 left :)

othercorey:
Looks ok to me.

PR #15964, opened 2021-09-25
Split Router::extensions() combined setter/getter

dereuromark:
One of the last combined set/get methods in the core.
With this, we should have a clean API now, using the set* part for setting, and the normal method for getting in line with the others we split.

Also consistent with RouteBuilder usage

        $routes->setExtensions(['json', 'xml']);
    });

### DIFF
--- a/src/Routing/Router.php
+++ b/src/Routing/Router.php
@@ -744,7 +744,7 @@ class Router
     public static function extensions($extensions = null, $merge = true): array
     {
         if ($extensions !== null) {
-            deprecationWarning('Setting extensions is deprecated, use setExtensions() instead.');
+            deprecationWarning('Setting extensions with extensions() is deprecated, use setExtensions() instead.');
             static::setExtensions($extensions, $merge);
 
             return static::$_defaultExtensions;
--- a/src/Routing/Router.php
+++ b/src/Routing/Router.php
@@ -720,38 +720,68 @@ class Router
     }
 
     /**
-     * Get or set valid extensions for all routes connected later.
+     * Gets or sets valid extensions for all routes connected later.
      *
      * Instructs the router to parse out file extensions
      * from the URL. For example, http://example.com/posts.rss would yield a file
      * extension of "rss". The file extension itself is made available in the
      * controller as `$this->request->getParam('_ext')`, and is used by the RequestHandler
      * component to automatically switch to alternate layouts and templates, and
-     * load helpers corresponding to the given content, i.e. RssHelper. Switching
+     * load helpers corresponding to the given content, i.e. RSS/CSV/JSON. Switching
      * layouts and helpers requires that the chosen extension has a defined mime type
-     * in `Cake\Http\Response`.
+     * in `\Cake\Http\Response`.
      *
      * A string or an array of valid extensions can be passed to this method.
      * If called without any parameters it will return current list of set extensions.
      *
-     * @param array<string>|string|null $extensions List of extensions to be added.
+     * Deprecated: Setting extensions is deprecated since 4.3.0. Use setExtensions() instead.
+     *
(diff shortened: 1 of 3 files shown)


ADmad:
I would rather deprecate extensions setting using Router's static methods similar to the other methods and use Routebuilder instead. 

The only reason I haven't done so is because RequestHandlerComponent uses `Router::extensions()`. Now that I think about it, it can just use the extension of the route matched for the current request instead.

src/Routing/Router.php
```diff
@@ -720,38 +720,68 @@ public static function normalize($url = '/'): string
...
+     *   Defaults to `true`. Deprecated since 4.3.0.
      * @return array<string> Array of extensions Router is configured to parse.
      */
     public static function extensions($extensions = null, $merge = true): array
     {
-        $collection = static::$_collection;
-        if ($extensions === null) {
-            return array_unique(array_merge(static::$_defaultExtensions, $collection->getExtensions()));
+        if ($extensions !== null) {
+            deprecationWarning('Setting extensions is deprecated, use setExtensions() instead.');
```
markstory:
```suggestion
            deprecationWarning('Setting extensions with extensions() is deprecated, use setExtensions() instead.');
```

ADmad:
> Now that I think about it, it can just use the extension of the route matched for the current request instead.

Actually `RequestHandlerComponent` already check's the currently matched route's extension but if not set it uses the list returned by `Router::extensions()` to decide whether it should set the extension based on the `Accept` header for matching type.

So we could add an option to `RequestHandlerComponent` for specifying the extensions/types it's interested in instead having it use `Router::extensions()`.

ADmad:
TLDR; `Router::extensions()` as a getter is rarely required by users and it's use in `RequestHandler` can be removed, so the method split doesn't provide much value IMO. We either try to deprecated it altogether or keep it as it is and sort it out in `5.x`.

markstory:
> Actually RequestHandlerComponent already check's the currently matched route's extension but if not set it uses the list returned by Router::extensions() to decide whether it should set the extension based on the Accept header for matching type.

I'm not sure this will be 100% backwards compatible. Right now RequestHandler will allow extension types on routes that don't have them enabled due to `Router::extensions()` being used.

Limiting extension detection to only the types that are enabled for the route seems like a good change to me. as we remove types bleeding into actions that shouldn't have extensions enabled, but it is something we would need to document.

ADmad:
Yes, currently the types do unnecessarily bleed into all actions which might not actually support them. Not being able to maintain 100% backwards compatibility or any easy way to update existing code is why I chose to keep `Router::extensions()` and not depreciate like the other methods.

I think it would be best to add the new option to `RequestHandlerComponent` for limiting extensions/types in 5.0, where we don't have the burden of maintaining 100% BC.

So to reiterate, splitting out the setter into a separate method doesn't bring much value if setting/getting extensions globally through router is going to be dropped altogether in 5.0.

markstory:
> So to reiterate, splitting out the setter into a separate method doesn't bring much value if setting/getting extensions globally through router is going to be dropped altogether in 5.0.

I agree. Moving towards the route controlling which extensions are valid is a good path forward.

dereuromark:
I am fine with an alternative PR that deprecates it all, so you can remove in 5.x

markstory:
Part of the replacement for this is in #15971

markstory:
Based on the discussion in #15971 should we close this and revisit in 4.4 when we can come up with a replacement for RequestHandler and fully deprecate the `Router::extensions()` without a static method replacement?

ADmad:
Replacing `RequestHandlerComponent` and not relying on globally set extensions will make `Router::extensions()` redundant. So let's keep it unchanged and deprecate it when we have an alternative to the component ready.

othercorey:
@dereuromark Is it ok to close this do you want to still merge part of it?

PR #15965, opened 2021-09-25
Fix issues found by PHPStan in tests/

dereuromark:
Refs https://github.com/cakephp/cakephp/pull/15963


othercorey:
Don't know why some were cancelled, but looked ok anyway.

PR #15966, opened 2021-09-26
Fix up doc typos or missing class links.

dereuromark:
Using the linking for IDE I found a few invalid class/method links, fixed those along the way.


src/Cache/Cache.php
```diff
@@ -33,7 +33,7 @@
  *
  * ```
  * Cache::config('shared', [
- *    'className' => Cake\Cache\Engine\ApcuEngine::class,
+ *    'className' => \Cake\Cache\Engine\ApcuEngine::class,
```
othercorey: Users would write the original version though, right?

dereuromark: Only in a non-namespaced file it work work.
I guess we could revent this if this is bootstrap file context.

Issue #15514, opened 2021-05-14
Extend primitives support in action arguments


smarek:
This is a (multiple allowed):

* [x] bug
* [x] enhancement
* [ ] feature-discussion (RFC)

* CakePHP Version: 4.2.6
* Platform and Target: nginx, php7.4

Per this code in ControllerFactory https://github.com/cakephp/cakephp/blob/master/src/Controller/ControllerFactory.php#L122

Would it be possible to extend this support for primitives at least for integers? It's usually main navigational param when accessing objects (usually it's database id which is of `int` type)

Currently the only way to support this in implementation is to override the `controllerFactory` protected field within `Application` to subclass of current ControllerFactory and handle the primitives in `invoke`

Previously working was solution I presented in initial report of  #15023 (specifically overriding the `AppController::invokeAction` implementation), which is not with changes introduced in cakephp 4.2.x

markstory:
> Would it be possible to extend this support for primitives at least for integers? It's usually main navigational param when accessing objects (usually it's database id which is of int type)

Possible but we need to solve how invalid data should be handled. If we add `int` we should also add `float`, and `bool` too.

smarek:
@markstory is there anything i can help with ? 
Also please, could the final solution be something, that is easily extendable? I've seen the [DI container](https://book.cakephp.org/4/en/development/dependency-injection.html) work-in-progress, and even tho that could be confusing, i had initially tried to override the primitives support through providing primitives Service, and putting the logic of parsing/casting/matching action arguments in overridable component, would be really appreciated. Thank you

markstory:
> is there anything i can help with ?

If you have time you could extend the string support to also work with int/float/bool. We can start by raising an exception on data that can't be cast to the right type, and go from there.

The DI container is the plan for making action parameters extensible. Primitives need to be an exception though as the DI container uses type hints to resolve parameters and primitives can't be typehint resolved.

smarek:
I'll provide PR with that and tests, not a problem. I'd like to also include setter/getter for `controllerFactory` within `BaseApplication` and externalizing component that handles the casting arguments, so it's extendable. Sounds good to you?

markstory:
>  I'd like to also include setter/getter for controllerFactory within BaseApplication and externalizing component that handles the casting arguments, so it's extendable.

How would a a setter/getter for controllerFactory be used? The application object and just mutate the controllerFactory as required, and there isn't generally anything controlling the application.

[...]

PR #15968, opened 2021-09-26
cake 5: Re-organize action parameter type casting

othercorey:
Refs https://github.com/cakephp/cakephp/issues/15514

This re-organizes the code that determines if a passed parameter type is supported. Supports only `string` primitive types still, but it provides a structure for extending this.

We don't support union types. They will fallback to container lookup which should fail with a null type name.

### DIFF
--- a/src/Controller/ControllerFactory.php
+++ b/src/Controller/ControllerFactory.php
@@ -18,19 +18,16 @@ namespace Cake\Controller;
 
 use Cake\Core\App;
 use Cake\Core\ContainerInterface;
+use Cake\Core\Invoker;
 use Cake\Http\ControllerFactoryInterface;
 use Cake\Http\Exception\MissingControllerException;
 use Cake\Http\MiddlewareQueue;
 use Cake\Http\Runner;
 use Cake\Http\ServerRequest;
-use Closure;
-use InvalidArgumentException;
 use Psr\Http\Message\ResponseInterface;
 use Psr\Http\Message\ServerRequestInterface;
 use Psr\Http\Server\RequestHandlerInterface;
 use ReflectionClass;
-use ReflectionFunction;
-use ReflectionNamedType;
 
 /**
  * Factory method for building controllers for request.
@@ -133,9 +130,11 @@ class ControllerFactory implements ControllerFactoryInterface, RequestHandlerInt
         }
 
         $action = $controller->getAction();
-        $args = $this->getActionArgs(
+        $invoker = new Invoker($this->container);
+        $args = $invoker->resolveArguments(
             $action,
-            array_values((array)$controller->getRequest()->getParam('pass'))
+            (array)$controller->getRequest()->getParam('pass'),
+            ['coerce' => true]
         );
         $controller->invokeAction($action, $args);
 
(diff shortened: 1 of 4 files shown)


src/Controller/ControllerFactory.php
```diff
@@ -210,6 +199,28 @@ protected function getActionArgs(Closure $action, array $passedParams): array
...
+        if (!$typeName) {
+            return $arg;
+        }
+
+        switch ($typeName) {
+            case 'string':
+                return $arg;
+        }
+
+        throw new InvalidArgumentException("Unable to cast action argument to parameter type `$typeName`.");
```
othercorey: This could probably return `null` and the exception handled by invoke(). I don't think null can be passed from the request, can it?

ADmad: > I don't think null can be passed from the request, can it?

Nope.

src/Controller/ControllerFactory.php
```diff
@@ -160,47 +160,36 @@ protected function getActionArgs(Closure $action, array $passedParams): array
...
+                if ($passedParams) {
+                    $args[$position] = $this->castArgumentToType(array_shift($passedParams), $typeName);
                 } elseif ($hasDefault) {
                     $args[$position] = $parameter->getDefaultValue();
                 }
                 continue;
             }
 
-            // Check the container and parameter default value.
+            // Parameters wtih non-primitive types can be found in the container
```
markstory:
```suggestion
            // Parameters with non-primitive types can be found in the container
```

othercorey:
I moved the argument resolution into an Invoker class for a general solution.

src/Core/Invoker.php
```diff
@@ -0,0 +1,152 @@
...
+     * @param string|null $typeName Parameter type name or null for no type
+     * @return string|float|int|bool
+     */
+    protected function coercePrimitiveType(string $argument, ?string $typeName): string|float|int|bool
+    {
+        if ($typeName === null) {
+            return $argument;
+        }
+
+        switch ($typeName) {
```
ADmad, line 145: You'll be adding cases for more types here right?

src/Core/Invoker.php
```diff
@@ -0,0 +1,152 @@
...
+     *
+     * String arguments are coerced to parameter type for
+     * float, int and bool types.
+     *
+     * @param \Closure $closure Function to invoke
+     * @param array<mixed> $arguments Function arguments
+     * @return mixed
+     * @throws \InvalidArgumentException When unable to find argument for parameters.
+     */
+    public function invokeWithCoercion(Closure $closure, array $arguments)
```
markstory, line 65: When would you not want coercion? Seems like we can have a simpler interface if we always coerce.

othercorey: I assumed there was a use case for a generic invoker. Maybe we don't need an Invoker class and just a general way to resolve arguments.

markstory: I don't think we have that use case yet. We could add the methods when we understand those requirements better. As you know, public APIs are hard for us to evolve. You could mark this class as `internal` to escape our backwards compat promises.

othercorey:
Will re-create this PR.

